PR #631, opened 2016-02-04
Multithreading

ThomasWaldmann:
updating the multithreading branch with changes from master.

some tests broken, needs more work.


ThomasWaldmann:
@enkore as you are working on MT stuff, do you think a shall merge this PR although it is not (yet) in the best state?

enkore:
Having this in the main repo in a side-branch would probably make it easier to collaborate on this. We should make sure though that no one gets any wrong ideas and that this code shouldn't be considered for production use any time soon. (Maybe truncate README and put a clear warning there instead?).

ThomasWaldmann:
There's a rather clear warning about non-release versions in README already.

PR #655, opened 2016-02-12
mention kldload

jungle-boogie:
minor update in the docs to mention how to mount with freeBSD


ThomasWaldmann:
I don't think we should spoil the usage section with installation details, that's what the installation section is about. Can you put it there, please? If there is no FreeBSD section yet, create one.

PR #677, opened 2016-02-18
Format code block and match header style to rest of documentation

ibennetch:
Signed-off-by: Isaac Bennetch <bennetch@gmail.com>

I couldn't actually build the documentation locally due to an error `ImportError: No module named _version`, so you might want to double-check before merging...but it looks like the right thing to do.


ThomasWaldmann:
Best is to have everything installed in a virtualenv (as usual for python development).
The _version stuff comes from setuptools-scm, which will be installed then.

I'll check the change, no problem.

ThomasWaldmann:
Can you do the fix at the right place?

ibennetch:
Sure, I'll be happy to take care of that sometime in the not-too-distant future.

Issue #649, opened 2016-02-08
freebsd fuse for mount


jungle-boogie:
Hello,

Trying to mount a backup:
`borg mount pulsebackup::pulse.net-2016-02-06-01-15 gg`
`Loading fuse support failed [ImportError: No module named 'llfuse']`

I manually loaded the kernel module with
`sudo kldload fuse`
and see this is dmesg:
`fuse-freebsd: version 0.4.4, FUSE ABI 7.8`

Fuse was imported into base with this commit from 3 years and 3 months ago:
https://svnweb.freebsd.org/base?view=revision&revision=241519

Any chance of getting `mount` working on freebsd? I tested this on 10.2-release.

ThomasWaldmann:
you need python-llfuse.

jungle-boogie:
> you need python-llfuse.

And that's installed with pip? 

```
% pip3 -V
pip 8.0.2 from /usr/local/lib/python3.4/site-packages (python 3.4)
```

```
% sudo pip3 install llfuse
Collecting llfuse
  Downloading llfuse-0.42.1.tar.bz2 (320kB)
    100% |████████████████████████████████| 323kB 952kB/s
No files/directories in /tmp/pip-build-8wuzxna5/llfuse/pip-egg-info (from PKG-INFO)
```

And yet, I don't see llfuse:

```
% pip3 list
borgbackup (1.0.0rc1)
decorator (4.0.4)
ipython (4.0.0)
ipython-genutils (0.1.0)
msgpack-python (0.4.7)
ngrid (0.1.0)
numpy (1.10.1)
pandas (0.17.0)
pandashells (0.1.5)
path.py (8.1.2)
pexpect (4.0.1)
pickleshare (0.5)
pip (8.0.2)
ptyprocess (0.5)
python-dateutil (2.4.2)
pytz (2015.7)
setuptools (18.5)
simplegeneric (0.8.1)
six (1.10.0)
termsql (0.3)
traitlets (4.0.0)
virtualenv (13.1.2)
wheel (0.26.0)
````

ThomasWaldmann:
0.42.1 does not work due to incompatible changes in llfuse, see my updated first comment.

ThomasWaldmann:
There is also the possibility now to do "pip install borgbackup[fuse]" that will automatically install the correct llfuse version. I updated the docs already about that.

jungle-boogie:
Hello,
I do see the docs were updated:
[...]

PR #685, opened 2016-02-20
 freebsd fuse for mount for #649

jungle-boogie:


docs/installation.rst
```diff
@@ -163,6 +163,23 @@ FUSE for OS X, which is available as a pre-release_.
...
+
+    $ sudo pkg install fusefs-libs
+    $ sudo kldload fuse
+    $ borg mount /mnt/backup::root-2013-08-02 /tmp/mymountpoint
+    $ ls /tmp/mymountpoint
+    bin  boot  etc  lib  lib64  mnt  opt  root  sbin  srv  usr  var
+    $ umount /tmp/mymountpoint
+
+
+
```
ThomasWaldmann: I needed some more stuff to get it installed on freebsd:
https://github.com/borgbackup/borg/blob/master/Vagrantfile#L75

I am not a freebsd user, i just got it working on vagrant for testing.
Some of that stuff is only needed for vagrant or building python and should not be included here.
But some of the other stuff seems missing here (e.g. one does not want to kldload all the time after each reboot).

Also, do not give a mount / umount example here, that is not related to installation.

jungle-boogie: > I needed some more stuff to get it installed on freebsd:

But that's already covered here:
https://github.com/borgbackup/borg/blob/master/docs/installation.rst#dependencies

 > e.g. one does not want to kldload all the time after each reboot).

Okay, I can change that.

> Also, do not give a mount / umount example here, that is not related to installation.

fair enough.

jungle-boogie: What's bash used for?
https://github.com/borgbackup/borg/blob/master/Vagrantfile#L79

a build or run time dependency of what?

ThomasWaldmann: the dependency installation is already described in general (for everybody) and specifically for some systems like debian/ubuntu, fedora, etc.

so, if we want to have more precise instructions for freebsd also, a section like the one seen for debian could be added with specific instructions.

ThomasWaldmann: you can ignore the bash stuff in the Vagrantfile.

I just needed one same shell on every system to have the same scripts work on every platform.

Issue #701, opened 2016-02-29
Drop makedirs for cache and keys on the remote side


manwegit:
Running borg against remote target where home directory is not writeable, borg will exit with
`Remote: PermissionError: [Errno 13] Permission denied: '/home/backup/.borg/keys' `
And same for .borg/cache or .cache/borg/ (depending on version)

Fix with something like:
`        if args.func != self.do_serve:`
to arhiver.py line 1418 (and indents etc) 

Workaround is to pre create said directories or using ` --remote-path=` with wrapper script

ThomasWaldmann:
does it actually use these directories at all on the "server" side?

manwegit:
As far I could test, no:
tried extract, mount, create, check and init

As long as keys and cache dir exist (even when not writable) borg create worked fine.

I did not go through source with a fine comp and this was just running binary test against server.

ThomasWaldmann:
as encryption is done client-side, keys are not expected to ever be on a remote server (or used by remote server).

also, the cache is primarily to accelerate access to remote data, that could be only accessible over a slow connection, so that also is only expected on the client.

ThomasWaldmann:
fixed by:
https://github.com/borgbackup/borg/commit/9c9ac2fd58d3217ae431db96a5dd1196d09be99a

thanks!

PR #702, opened 2016-02-29
Remote end does not need cache nor keys directories, do not make them.

manwegit:
If remote home is readonly for the user, using borg will fail on permission
problem to create cache and keys directories. borg --server does not use nor need
them so drop mkdirs from run()
see #701 for discussion

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -17,7 +17,7 @@ import traceback
 from . import __version__
 from .helpers import Error, location_validator, archivename_validator, format_line, format_time, format_file_size, \
     parse_pattern, PathPrefixPattern, to_localtime, timestamp, safe_timestamp, \
-    get_cache_dir, get_keys_dir, prune_within, prune_split, \
+    get_cache_dir, prune_within, prune_split, \
     Manifest, remove_surrogates, update_excludes, format_archive, check_extension_modules, Statistics, \
     dir_is_tagged, bigint_to_int, ChunkerParams, CompressionSpec, is_slow_msgpack, yes, sysinfo, \
     EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR, log_multi, PatternMatcher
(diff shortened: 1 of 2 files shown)


manwegit:
This needs testing from other people. As far as I could test and read remote code, cache and keys are not needed on the remote side. Remote uses temporary directory for some caching, but not cache dir.

borg/archiver.py
```diff
@@ -1416,20 +1416,21 @@ def run(self, args):
...
+            if not os.path.exists(cache_dir):
+                os.makedirs(cache_dir)
+                os.chmod(cache_dir, stat.S_IRWXU)
+                with open(os.path.join(cache_dir, 'CACHEDIR.TAG'), 'w') as fd:
+                    fd.write(textwrap.dedent("""
+                        Signature: 8a477f597d28d172789f06886806bc55
+                        # This file is a cache directory tag created by Borg.
+                        # For information about cache directory tags, see:
+                        #       http://www.brynosaurus.com/cachedir/
+                        """).lstrip())
```
ThomasWaldmann: well, guess it would work that way.

maybe a bit prettier (and having less code in archiver.py) would be to move the code that creates/inits the dirs into get_cache_dir and get_keys_dir. and remove the calls to get_cache_dir and get_keys_dir from here as the result is not needed here anyway.

manwegit:
Changed directory checks and creation to herlpers.py

ThomasWaldmann:
looks good.

did you check if this fixes the issue from the ticket you filed?

manwegit:
Testing local and ssh setups with both local and remote homes non writeable combinations.
All checks pass and it fixes remote directory problems without being remote specific.

Only problem is if local keys dir is not writeable and keyfile type repo is created (borg init -e keyfile). This problem existed also before this patch. I'll open another issue, because dealing with problem during repo init is a different issue.

manwegit:
Running tests again.. I noticed that get_keys_dir() is not used in archiver.py anymore. Dropping import, running tests and I'll push that change also.

manwegit:
Tests are ok.

Issue #661, opened 2016-02-14
Add bandwidth limit support


chebee7i:
This issue appears on the checkbox list in #5. It is attic issue: https://github.com/jborg/attic/issues/103. I wasn't sure if we should be adding comments there or here.

---

The original issue requests that support be added to limit bandwidth. `trickle` was mentioned as an option. People said it wouldn't/didn't work since borg uses a pipe through ssh, while trickle works with sockets.

I just wanted to report that if you are using borg with sshfs (i.e. backing up to rsync.net), then using trickle will work properly. You must call trickle when performing the ssh mount. So to limit bandwidth to 5 Mbps, you'd use:

```
trickle -u 5000 sshfs username@hostname: /path/to/mountpoint
borg create ...
```

manwegit:
Install pipeviewer (pv) on ubuntu and most likely debian it's simply: `sudo apt-get install pv`

Create a wrapper script: (I'll create it to `/usr/local/bin/pv-wrapper`)
```
#!/bin/bash
     ## -q, --quiet              do not output any transfer information at all
     ## -L, --rate-limit RATE    limit transfer to RATE bytes per second
export RATE=307200
pv -q -L $RATE  | "$@"
```

That done, add  BORG_RSH env
`export BORG_RSH='/usr/local/bin/pv-wrapper.sh ssh ' `

Now borg will be bandwidth limited. Nice thing about pv is that you can change rate-limit on the fly:
`pv -R $(pidof pv) -L 102400`

I've used same setup with rsync where I was moving multi terabyte set of data and had to moderate bandwidth during office hours.

```
# example cron setup:
00 09 * * 1-5 root ( pidof pv > /tmp/.pv-pid && pv -R $(cat /tmp/.pv-pid ) -L  307200)
00 18 * * * root ( pidof pv > /tmp/.pv-pid && pv -R $(cat /tmp/.pv-pid ) -L  819200)

```

Test:
Remote repo was removed before each test

source: 320M    total
Without pipeviewer:
Duration: **5 minutes 8.50 seconds**
Number of files: 60

`## wrapper:`
`pv -q -L 307200  | "$@"`
Duration: **18 minutes 7.00 seconds**

`pv -q -L 102400 | "$@"`
Stopped exection after 40 mins, expected run time around 54 minutes and borg --progress output was inline with the estimate.

Original test with pipeviewer, but no rate-limit:
Duration: **6 minutes 29.38 seconds**
[...]

PR #705, opened 2016-03-01
Faq: how to limit bandwidth

manwegit:
Example to faq documentation how bandwidth limiting can be done with pipeviewer.

see #661 for discussion

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -264,6 +264,32 @@ Since only the files cache is used in the display of files status,
 those files are reported as being added when, really, chunks are
 already used.
 
+
+Is there a way to limit bandwidth with |project_name|?
+------------------------------------------------------
+
+There is no command line option to limit bandwidth with |project_name|, but
+bandwidth limiting can be accomplished with pipeviewer_:
+
+Create a wrapper script:  /usr/local/bin/pv-wrapper  ::
+
+    #!/bin/bash
+        ## -q, --quiet              do not output any transfer information at all
+        ## -L, --rate-limit RATE    limit transfer to RATE bytes per second
+    export RATE=307200
+    pv -q -L $RATE  | "$@"
+
+Add BORG_RSH environment variable to use pipeviewer wrapper script with ssh. ::
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -264,6 +264,32 @@ Since only the files cache is used in the display of files status,
...
+
+    #!/bin/bash
+        ## -q, --quiet              do not output any transfer information at all
+        ## -L, --rate-limit RATE    limit transfer to RATE bytes per second
+    export RATE=307200
+    pv -q -L $RATE  | "$@"
+
+Add BORG_RSH environment variable to use pipeviewer wrapper script with ssh. ::
+
+    export BORG_RSH='/usr/local/bin/pv-wrapper.sh ssh '
```
ThomasWaldmann: is the blank after "ssh" needed?

manwegit: no, it's not needed. Fixed.

docs/faq.rst
```diff
@@ -264,6 +264,32 @@ Since only the files cache is used in the display of files status,
...
+
+    #!/bin/bash
+        ## -q, --quiet              do not output any transfer information at all
+        ## -L, --rate-limit RATE    limit transfer to RATE bytes per second
+    export RATE=307200
+    pv -q -L $RATE  | "$@"
+
+Add BORG_RSH environment variable to use pipeviewer wrapper script with ssh. ::
+
+    export BORG_RSH='/usr/local/bin/pv-wrapper.sh ssh'
```
ThomasWaldmann: hmm, did you try just using it directly, without the wrapper script:

export BORG_RSH='pv --quiet --rate-limit 307200 ssh'

manwegit: pv need pipe, so that would be 'pv --quiet --rate-limit 307200 | ssh' which does not work.
If I remember correctly pv complains about unknow parameter --umask which is borg server parameter.
I don't think you can use pipes in  BORG_RSH env hence the wrapper.

ThomasWaldmann: oops, overlooked the pipe...

PR #710, opened 2016-03-03
Update README.rst to new changelog location

petrcech:
Update README.rst to new changelog location in docs/changes.rst


ThomasWaldmann:
thanks! visiting the other url is really not that useful as github does not follow the symlink...

PR #733, opened 2016-03-10
Unpack file stats before passing to fuse

ipha:
The file stats are packed with 'int_to_bigint' before storing in the archive. Extracting from the archive correctly unpacks these, but not on fuse. This results in a byte string being passed to llfuse if any of the atime/mtime/ctime are unexpectedly large.


ThomasWaldmann:
Good find, thanks!

Issue #729, opened 2016-03-10
Show which distinct versions of a file exist


bwurst:
We do backup with attic (as attic is frozen by now, we think about switching to borg) and there's one thing that really bugs me. I cannot see which distinct versions of a specific file exists. I can mount and extract a daily snapshot and the file is there but I cannot see if this file is the same as the day before.

We operate rdiff-backup in parallel mostly because of this feature. It's really useful if a web app has been hacked for example. Normally, files don't change too much, if some files are changed when no upgrade has been done, we pay stronger attention to this. For this to work, we need to know which file has been changed in which snapshot or, other approach, which versions of a file really exists.

Does borg, internally, know this? Is there some kind of file checksum list that can be used for this? I'd do some coding if the data structure does provide this kind of information but I did not find it by now...

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/31735500-show-which-distinct-versions-of-a-file-exist)

infectormp:
+1
To solve this task we currently developing a webapp that will perform function that you described.

enkore:
You can do this by extracting the file(s) of interest and comparing them. E.g. extract the interesting part of snapshot *n* into `a/`, and *n-1* into `b/`. Diff a, b. If not different, move `b/` to `a/`, extract *n-2* to `b/`. Diff again. Repeat as necessary.

infectormp:
@enkore your workaround good only for non regular operations. If you need to view changes constantly it is uncomfortable. You need have a lot of time to make this operations regularly.

enkore:
Yes. Still, this method would be quite usable for investigating data loss, breaches and the like.

bwurst:
@enkore: this does not work for day to day operation. We back up about 300 GB of user's content per machine. Extracting and comparing this won't be fun.
I thought about building a checksum table daily but it's also a huge waste of resources because I think borg does this anyway. Just now I don't have much time so I'll accept that this seems not to be possible out of the box.

enkore:
Borg doesn't know if a file changed relative to the previous archive, because that is not a concept Borg has (it does not know what archive is considered "the previous archive" by the user).

If using the cache then the output of `--list -v` when creating the archive will show which files have been changed *relative to the file cache used*. E.g. `U some_binary` => unchanged. `A sshd` => someone changed sshd.

If no cache is used then I think every file will always show as `A`.

bwurst:
[...]

Issue #735, opened 2016-03-11
lazy eval for list output?


ThomasWaldmann:
the current `borg list` implementation has some hardcoded optimization to either compute some "standard" output (show some specific attributes in a specific way) or to support user-defined output format (making it possible to show more attributes in a user-defined way).

maybe this code could be unified again using lazy evaluation - to only compute exactly what's needed.

that would also open the door to do some more heavywheight stuff (that we for sure do not want to do in a non-lazy way): e.g.:
- display the sha256 of the file
- display unique chunks of this file(?)

manwegit:
About the newline 'NEWLINE': os.linesep,  # TODO: difference between LF and NEWLINE?
I could not test and I don't if there is differences based on os for newline.
Other than that, some people might be more comfortable with NEWLINE term than LF

Part of the reason for list-format change was to make way for future more complex chunk information prints. I'm not experienced enough python coder, but lazy eval based on keys sounds great.

enkore:
Hm, I guess if someone wants to emulate the output of a different program closely LF vs. CR vs. NEWLINE could be useful, but that seems a bit far-fetched maybe.

Yesterday I profiled the current master branch and a slightly better version of what I made public in my repository so far. The difference in total run time is miniscule (for the exact same output my version takes about 3 % less wall time: performance for the default output format shouldn't be considered a "driving force").

Find two kcachegrind maps attached. As you can see, the runtime is *dominated* by what *has* to be done. Differences in the runtime for the default format thus stem from tiny differences in overhead.

![newlist-profile](https://cloud.githubusercontent.com/assets/1330164/13748536/7b58d7f8-e9fd-11e5-9ad6-2561d19f4d5f.png)
![oldlist-profile](https://cloud.githubusercontent.com/assets/1330164/13748538/7e658522-e9fd-11e5-9d2d-4434b879044f.png)

For the default output safe_timestamp calls can be reduced a bit, but that will also yield only 2-3 % I expect.

btw. I think I found a (valid) optimization of `unpack_many` which speeds up diff, list and extract by 5-15 %. So far it looks like it doesn't change program behaviour otherwise, but I will perform much additional testing before considering pushing it to my public repo.

ThomasWaldmann:
@enkore interesting stuff, but please stay on topic in this ticket (open new / amend existing other topics if it does not fit).

[...]

PR #740, opened 2016-03-12
borg diff: find different files between archives

enkore:
Follow-up to #729 

**TODO**

- [x] ~~Output formatting(?)~~ => later, if needed (compare "machine readable output")
- [x] Tests
- [x] Better implementation of `compare_chunks`? (see line comment)
- [x] Decide how much metadata should be compared. Currently: only mode. Include user and group (yes)? ~~xattrs? ACLs?~~ (open issue if needed)
- [x] ~~Machine readable output~~ (dropped in f5f3e25)
- [x] Detection of chunker params
   - [x] Added chunker params to metadata for new archives
   - [x] Fail-safe by default. Let user override if he's sure (--same-chunker-params).
- [x] investigate: smarter data structure for orphans?

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -145,6 +145,7 @@ class Archive:
         self.numeric_owner = numeric_owner
         if start is None:
             start = datetime.utcnow()
+        self.chunker_params = chunker_params
         self.start = start
         if end is None:
             end = datetime.utcnow()
@@ -261,6 +262,7 @@ Number of files: {0.stats.nfiles}'''.format(
             'username': getuser(),
             'time': start.isoformat(),
             'time_end': end.isoformat(),
+            'chunker_params': self.chunker_params,
         })
         data = msgpack.packb(metadata, unicode_errors='surrogateescape')
         self.id = self.key.id_hash(data)
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1,9 +1,10 @@
 from binascii import hexlify, unhexlify
 from datetime import datetime
-from hashlib import sha256
+from itertools import zip_longest
 from operator import attrgetter
 import argparse
 import functools
+import hashlib
 import inspect
 import io
 import os
@@ -80,6 +81,45 @@ class Archiver:
     def print_file_status(self, status, path):
         if self.output_list and (self.output_filter is None or status in self.output_filter):
             logger.info("%1s %s", status, remove_surrogates(path))
+
+    @staticmethod
+    def compare_chunk_contents(chunks1, chunks2):
+        """Compare two chunk iterators (like returned by :meth:`.DownloadPipeline.fetch_many`)"""
+        end = object()
+        alen = ai = 0
+        blen = bi = 0
+        while True:
+            if not alen - ai:
+                a = next(chunks1, end)
+                if a is end:
+                    return not blen - bi and next(chunks2, end) is end
+                a = memoryview(a)
+                alen = len(a)
+                ai = 0
+            if not blen - bi:
+                b = next(chunks2, end)
+                if b is end:
+                    return not alen - ai and next(chunks1, end) is end
+                b = memoryview(b)
+                blen = len(b)
+                bi = 0
+            slicelen = min(alen - ai, blen - bi)
+            if a[ai:ai + slicelen] != b[bi:bi + slicelen]:
+                return False
+            ai += slicelen
+            bi += slicelen
+
+    @staticmethod
+    def build_matcher(excludes, paths):
+        matcher = PatternMatcher()
+        if excludes:
+            matcher.add(excludes, False)
+        include_patterns = []
+        if paths:
+            include_patterns.extend(parse_pattern(i, PathPrefixPattern) for i in paths)
+            matcher.add(include_patterns, True)
+        matcher.fallback = not include_patterns
+        return matcher, include_patterns
 
     def do_serve(self, args):
         """Start in server mode. This command is usually not used manually.
@@ -304,17 +344,7 @@ class Archiver:
         archive = Archive(repository, key, manifest, args.location.archive,
                           numeric_owner=args.numeric_owner)
 
-        matcher = PatternMatcher()
-        if args.excludes:
-            matcher.add(args.excludes, False)
-
-        include_patterns = []
-
-        if args.paths:
-            include_patterns.extend(parse_pattern(i, PathPrefixPattern) for i in args.paths)
-            matcher.add(include_patterns, True)
-
-        matcher.fallback = not include_patterns
+        matcher, include_patterns = self.build_matcher(args.excludes, args.paths)
 
         output_list = args.output_list
         dry_run = args.dry_run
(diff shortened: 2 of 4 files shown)


borg/archive.py
```diff
@@ -261,6 +262,7 @@ def save(self, name=None, timestamp=None):
             'username': getuser(),
             'time': start.isoformat(),
             'time_end': end.isoformat(),
+            'chunker_params': self.chunker_params,
```
enkore: Is this actually a safe thing to do, or can this break older versions trying to read this?

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+            return format_file_size(count)
+
+        def compare_chunks(lhs, rhs, lhs_archive, rhs_archive):
+            lhs_hash = sha256()
+            rhs_hash = sha256()
+            for chunk in lhs_archive.pipeline.fetch_many(lhs):
+                lhs_hash.update(chunk)
+            for chunk in rhs_archive.pipeline.fetch_many(rhs):
+                rhs_hash.update(chunk)
+            return lhs_hash.digest() == rhs_hash.digest()
```
enkore: So this is simple and all, but inefficient for large files, especially if they changed early in the file. I tried an alternate implementation in Python, basically comparing prefixes of chunks with a moving `index` (i.e. larger chunk starts with smaller chunk, get next chunk, do index starting at offset), but it was overall slower and way more complicated. I also don't think I caught all of the corner cases.

So I opted for this simple implementation where it is easy to see that it does the right thing. I also doubt many people actually change chunker params at all, so this fallback path is rather unimportant I think.

borg/archive.py
ThomasWaldmann: i don't think that can break something, but you have to be careful when reading this as it might not be there.

borg/archiver.py
ThomasWaldmann: interesting, I had expected just doing string compares is faster than sha256.
especially since it is quite likely that you do not need to compare all data to find a difference.

ThomasWaldmann: as left and right has not really a meaning here, can you change it to:
lhs -> chunk_ids1
rhs -> chunk_ids2
lhs_archive -> archive1
rhs_archive -> archive2
?

ThomasWaldmann: also, instead of hardcoding sha256, rather do it like that:
def compare_chunks(..., hash=hashlib.sha512):
    hash1 = hash()
    hash2 = hash()

Note that sha512 is faster than sha256 on 64bit machines.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+
+        def compare_items(path, item, compare_item, deleted=False):
+            """
+            Compare two items with identical paths.
+            :param deleted: Whether one of the items has been deleted
+            """
+            if item[b'mode'] != compare_item[b'mode'] and not deleted:
```
ThomasWaldmann: do the cheap evaluation first: if not deleted and ...

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+        def compare_chunks(lhs, rhs, lhs_archive, rhs_archive):
+            lhs_hash = sha256()
+            rhs_hash = sha256()
+            for chunk in lhs_archive.pipeline.fetch_many(lhs):
+                lhs_hash.update(chunk)
+            for chunk in rhs_archive.pipeline.fetch_many(rhs):
+                rhs_hash.update(chunk)
+            return lhs_hash.digest() == rhs_hash.digest()
+
+        def compare_items(path, item, compare_item, deleted=False):
```
ThomasWaldmann: item1 and item2?
or item and reference_item?

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                else:
+                    print(remove_surrogates(path), "different mode")
+                    print("\t", args.location.archive, stat.filemode(item[b'mode']))
+                    print("\t", args.compare, stat.filemode(compare_item[b'mode']))
+                return
+            if not stat.S_ISREG(item[b'mode']) and not deleted:
+                return
+            if b'chunks' not in item or b'chunks' not in compare_item:
+                # This is a link
+                if item.get(b"source", None) != compare_item.get(b"source", None):
```
ThomasWaldmann: None is the default default for .get().

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                    })
+                else:
+                    print(remove_surrogates(path), "different mode")
+                    print("\t", args.location.archive, stat.filemode(item[b'mode']))
+                    print("\t", args.compare, stat.filemode(compare_item[b'mode']))
+                return
+            if not stat.S_ISREG(item[b'mode']) and not deleted:
+                return
+            if b'chunks' not in item or b'chunks' not in compare_item:
+                # This is a link
```
ThomasWaldmann: if you are sure about this, why do you need .get(b'source', DEFAULT) below?

enkore: It should really say "One of [item, compare_item] is a link".

borg/archiver.py
enkore: Should use one consistent naming scheme. The argparser uses "REFERENCE" and "COMPARE" which is, uh, probably not the best way to name them. All other occurrences should follow that then.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                            "what": "link",
+                            "ref": item.get(b"source", "<regular file>"),
+                            "cmp": compare_item.get(b"source", "<regular file>")
+                        })
+                    else:
+                        print(remove_surrogates(path), "different link")
+                        print("\t", args.location.archive, item.get(b'source', '<regular file>'))
+                        print("\t", args.compare, compare_item.get(b'source', '<regular file>'))
+                return
+            if item[b'chunks'] != compare_item[b'chunks'] or deleted or not can_compare_chunks:
```
ThomasWaldmann: cheap evals first.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                # Contents are different
+                ref_chunks = {c[0] for c in item[b'chunks']}
+                cmp_chunks = {c[0] for c in compare_item[b'chunks']}
+                total_ref = None if not item[b'mode'] else sum(c[1] for c in item[b'chunks'])
+                total_compare = None if not compare_item[b'mode'] else sum(c[1] for c in compare_item[b'chunks'])
```
ThomasWaldmann: why do you check mode here?

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                if (not can_compare_chunks and total_ref == total_compare and not deleted and
+                        compare_chunks(ref_chunks, cmp_chunks, ref, compare)):
+                    return
+                added = sum(c[1] for c in (cmp_chunks - ref_chunks))
+                removed = sum(c[1] for c in (ref_chunks - cmp_chunks))
```
ThomasWaldmann: good :)

borg/archiver.py
enkore: mode = 0 is used as an in-band signal for a deleted file (`deleted=True`). I'll rework that part.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                        print("\t", args.compare, compare_item.get(b'source', '<regular file>'))
+                return
+            if item[b'chunks'] != compare_item[b'chunks'] or deleted or not can_compare_chunks:
+                # Contents are different
+                ref_chunks = {c[0] for c in item[b'chunks']}
+                cmp_chunks = {c[0] for c in compare_item[b'chunks']}
+                total_ref = None if not item[b'mode'] else sum(c[1] for c in item[b'chunks'])
+                total_compare = None if not compare_item[b'mode'] else sum(c[1] for c in compare_item[b'chunks'])
+                if (not can_compare_chunks and total_ref == total_compare and not deleted and
+                        compare_chunks(ref_chunks, cmp_chunks, ref, compare)):
```
ThomasWaldmann: hmm, "not can_compare_chunks" and calling "compare_chunks()" is weird.

enkore: I'll rename that to `can_compare_chunk_ids`. And compare_chunks() to compare_chunk_contents()

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+        manifest, key = Manifest.load(repository)
+        ref = Archive(repository, key, manifest, args.location.archive)
+        compare = Archive(repository, key, manifest, args.compare)
+
+        # Extract and check chunker params
+        cmdline_parser = self.build_parser()
+        ref_chunker_params = ref.metadata.get(b'chunker_params', cmdline_parser.parse_args(
+            ref.metadata[b'cmdline'][1:]).chunker_params)
+        cmp_chunker_params = compare.metadata.get(b'chunker_params', cmdline_parser.parse_args(
+            compare.metadata[b'cmdline'][1:]).chunker_params)
```
ThomasWaldmann: parsing the commandline does not in every case help here.
before 1.0 there was a different internal default than for >= 1.0.
so I guess better remove the cmdline parsing and just use what you get from metadata.

enkore: Hmm.
- The metadata works only for newly created archives, but if it is there, it is always correct.
- The cmdline only shows whether it was changed in the cmdline. So archives upgraded from <1.0 wouldn't be detected.

Maybe a little heuristic would help: Try to find a large file with a couple of chunks. If the average chunk size is somewhere near 64 kB, assume <1.0 chunker params, if it is near ~2 MB, assume >=1.0 chunker params?

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+        can_compare_chunks = ref_chunker_params == cmp_chunker_params
```
ThomasWaldmann: can_compare_chunkids is what you meant.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+        if args.paths:
+            include_patterns.extend(parse_pattern(i, PathPrefixPattern) for i in args.paths)
+            matcher.add(include_patterns, True)
+        matcher.fallback = not include_patterns
+
+        if args.json:
+            print("[")
+        orphan_refs = {}
+        orphan_cmps = {}
+        for refitem, cmpitem in zip_longest(
```
ThomasWaldmann: is zip_longest() useful here?
the sequences will easily be "out of sync" if there is just 1 missing or additionally in one of them.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+            if cmpitem:
+                cmpitem_morphan = orphan_refs.pop(cmpitem[b'path'], None)
+                if cmpitem_morphan:
+                    compare_items(cmpitem[b'path'], cmpitem_morphan, cmpitem)
+                else:
+                    orphan_cmps[cmpitem[b'path']] = cmpitem
+        # At this point orphan_* contain items that had no matching partner in the other archive
+        for added in orphan_cmps.values():
+            compare_items(added[b'path'], {
+                b'mode': 0,
```
ThomasWaldmann: maybe using an out-of-band value like None is better.

borg/archiver.py
```diff
@@ -353,6 +355,155 @@ def do_extract(self, args):
...
+                b'chunks': [],
+            }, added, deleted=True)
+        for deleted in orphan_refs.values():
+            compare_items(deleted[b'path'], deleted, {
+                b'mode': 0,
```
ThomasWaldmann: same here

borg/archiver.py
```diff
@@ -1083,6 +1234,35 @@ def build_parser(self, args=None, prog=None):
...
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='find differences in archive contents')
+        subparser.set_defaults(func=self.do_diff)
+        subparser.add_argument('-e', '--exclude', dest='excludes',
+                               type=parse_pattern, action='append',
+                               metavar="PATTERN", help='exclude paths matching PATTERN')
+        subparser.add_argument('--exclude-from', dest='exclude_files',
+                               type=argparse.FileType('r'), action='append',
+                               metavar='EXCLUDEFILE', help='read exclude patterns from EXCLUDEFILE, one per line')
+        subparser.add_argument('--json', dest='json',
```
ThomasWaldmann: use case?

borg/archiver.py
ThomasWaldmann: keep it simple. use reliable information and if there is none, do it in a safe way.

the issue with old archives being slow then will solve automatically over time. ;)

borg/archiver.py
enkore: Yes, but it's still very fast, because the orphan_* dicts only hold few items (in the test case I am using the peak size is about 6000, of which 5500 persist after this loop, so are never consumed in this loop, because these 5500 where deleted in either archive).

Using a list might be faster in some cases ("cold" end with deleted orphans, "hot" end which is only as large as the difference in the file lists). I take a look at that.

borg/archiver.py
ThomasWaldmann: we have nowhere json output yet, so to start simple, I'ld suggest to drop that from this changeset.

open a ticket about json / machine readable output and if people really need it, it could be added everywhere needed.

enkore: Will do

ThomasWaldmann:
i think comparing user and group (uid and gid) makes sense.
maybe skip acls and xattrs for now.

also tests are important, watch the coverage results that are computed after the travis ci checks.

ThomasWaldmann:
don't forget to add your name to AUTHORS.

Ape:
Just as a reference. [Snapper has a really good and fast ```status``` command](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Storage_Administration_Guide/ch-snapper-status-command.html) that shows which files and directories were added, removed, or modified between any two snapshots.

enkore:
You raise a valid point, the output format is arbitrary. We could instead reuse the code used by `borg list` (after refactoring it a bit), which may also make diff more flexible (#735).

enkore:
That's easier to read and functionally equivalent. Applying

ThomasWaldmann:
aside from the above mentioned 2 minor issues, this seems ready for merge.

enkore:
I'll fix them, also: thank you for your very in-depth review of code, that is some good work!

I think builds were broken in some commits, so I'm going to squash them down into a non-breaking commit (so that git bisect works) if you don't mind.

ThomasWaldmann:
ok, i'll merge it after travis finishes.

jungle-boogie:
Hi @enkore!

Thanks for the commit on this! I think it will be a valuable tool for others to use and find diffs.

Would it be too much to ask for a commit to the docs? If you can give a rough idea of how it works, I can test it out on master and make some docs--if we even need them.

Thanks!

enkore:
In #767 I added it to the sphinx docs, also with an example. I also explained what --same-chunker-params is or why it's needed.

travisdowns:
New feature looks good (trying it on my 1.0.x repositories). Is there any way to sort by the file sizes, or change the +/- size format so it's all bytes or all KB, and not a mix of bytes/kb/mb? My main use case is to see the largest changes by size, but the pretty printing makes that hard (and it's hard to visually scan the list when there are 100s of entries).

PR #745, opened 2016-03-13
provide wrapper for borg mount, fixes #743

### LINKED ISSUE #743: add a mount.borg wrapper
it just needs to do:
```
#!/bin/bash
borg mount "$@"
```

it can be used in fstab like:
```
mount.borg#repo_or_archive mountpoint fuse defaults noauto 0 0
```

nachtgeist:
- add archiver.main_mount()
- update usage of mount subcommand
- run build_usage
- add entry point in setup.py

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1347,6 +1347,13 @@ class Archiver:
         browsing an archive or restoring individual files. Unless the ``--foreground``
         option is given the command will run in the background until the filesystem
         is ``umounted``.
+
+        The command ``borgfs`` provides a wrapper for ``borg mount``. This can also be
+        used in fstab entries:
+        ``/path/to/repo /mnt/point fuse.borgfs defaults,noauto 0 0``
+
+        To allow a regular user to use fstab entries, add the ``user`` option:
+        ``/path/to/repo /mnt/point fuse.borgfs defaults,noauto,user 0 0``
         """)
         subparser = subparsers.add_parser('mount', parents=[common_parser],
                                           description=self.do_mount.__doc__,
@@ -1650,6 +1657,10 @@ def setup_signal_handlers():  # pragma: no cover
 
 
 def main():  # pragma: no cover
+    # provide 'borg mount' behaviour when the main script/executable is named borgfs
+    if os.path.basename(sys.argv[0]) == "borgfs":
+        sys.argv.insert(1, "mount")
+
     # Make sure stdout and stderr have errors='replace') to avoid unicode
     # issues when print()-ing unicode file names
     sys.stdout = io.TextIOWrapper(sys.stdout.buffer, sys.stdout.encoding, 'replace', line_buffering=True)
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -213,7 +213,17 @@ class Statistics:
 
 def get_keys_dir():
     """Determine where to repository keys and cache"""
-    xdg_config = os.environ.get('XDG_CONFIG_HOME', os.path.join(os.path.expanduser('~'), '.config'))
+
+    # os.path.expanduser() behaves differently for '~' and '~someuser' as
+    # parameters: when called with an explicit username, the possibly set
+    # environment variable HOME is no longer respected. So we have to check if
+    # it is set and only expand the user's home directory if HOME is unset.
+    if os.environ.get('HOME', ''):
+        home_dir = os.environ.get('HOME')
+    else:
+        home_dir = os.path.expanduser('~%s' % os.environ.get('USER'))
+
+    xdg_config = os.environ.get('XDG_CONFIG_HOME', os.path.join(home_dir, '.config'))
     keys_dir = os.environ.get('BORG_KEYS_DIR', os.path.join(xdg_config, 'borg', 'keys'))
     if not os.path.exists(keys_dir):
         os.makedirs(keys_dir)
--- a/docs/usage/mount.rst.inc
+++ b/docs/usage/mount.rst.inc
@@ -38,3 +38,10 @@ This command mounts an archive as a FUSE filesystem. This can be useful for
 browsing an archive or restoring individual files. Unless the ``--foreground``
 option is given the command will run in the background until the filesystem
 is ``umounted``.
+
+The command ``borgfs`` provides a wrapper for ``borg mount``. This can also be
+used in fstab entries:
+``/path/to/repo /mnt/point fuse.borgfs defaults,noauto 0 0``
+
+To allow a regular user to use fstab entries, add the ``user`` option:
+``/path/to/repo /mnt/point fuse.borgfs defaults,noauto,user 0 0``
--- a/setup.py
+++ b/setup.py
@@ -262,6 +262,7 @@ setup(
     entry_points={
         'console_scripts': [
             'borg = borg.archiver:main',
+            'borgfs = borg.archiver:main',
         ]
     },
     cmdclass=cmdclass,
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@ Borg Contributors ("The Borg Collective")
 - Michael Hanselmann <public@hansmi.ch>
 - Teemu Toivanen <public@profnetti.fi>
 - Marian Beermann <public@enkore.de>
+- Daniel Reichelt <hacking@nachtgeist.net>
 
 Borg is a fork of Attic.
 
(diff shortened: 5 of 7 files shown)


setup.py
```diff
@@ -262,6 +262,7 @@ def run(self):
     entry_points={
         'console_scripts': [
             'borg = borg.archiver:main',
+            'borgmount = borg.archiver:main_mount',
```
ThomasWaldmann: other specific mount commands are named mount.xxx (e.g. mount.cifs), does it make sense to follow that convention?

infectormp: I think using mount.borg will a good decision. I like it.

nachtgeist: `mount.borg` would suggest that `mount -t borg` works, which doesn't.

With fuse and the PR in its current state it's either `borgmount $arch $mntpt` or `mount -t fuse.borgmount $arch $mntpt` (see `sshfs` for a good example).

So I'd rather suggest renaming the wrapper to `borgfs`.

I'm also going to add the second form of `mount -t` to the docs.

nachtgeist: ...and change the fstab example from `borgmount#` to the `fuse.borgfs` syntax

enkore: > mount.borg would suggest that mount -t borg works, which doesn't.

It worked for me?

nachtgeist: That surprises me.

Well, I still think mount.borg is bad since it breaks out of the "fuse-namespace". None of sshfs, rdiff-backup-fs et all do provide `mount.*`, their binaries are called `sshfs`, `rdiff-backup-fs` and so on. So I guess we should stick to that already somewhat established convention.

enkore: Looks like mount just passes execution over to `mount.<-t>` if it exists[1]. But if there is already an established convention for this thing, we should follow it — I doubt "nice" fstab entries are important to many.

[1]

    $ cat /usr/bin/mount.something 
    #!/bin/sh

    echo mount.something $@
    $ sudo mount -t something a b
    mount.something a b -o rw

nachtgeist:
I did some more testing and this is what I came up with:


* mount.borg
  * fstab: ``/path/to/repo /mnt/repomount/a borg defaults,noauto,user 0 0``
    <pre>
    # mount a
    mount: unknown filesystem type 'borg'
    </pre>


  * fstab: ``/path/to/repo /mnt/repomount/b fuse.borg defaults,noauto,user 0 0``
    <pre>
    # mount b
    usage: borg [-h] [-V] <command> ...
    </pre>
    this is expected, just to show how commands are passed through


  * fstab: ``/path/to/repo /mnt/repomount/c fuse.mount.borg defaults,noauto,user 0 0``
    <pre>
    # mount c
    mount: wrong fs type, bad option, bad superblock on /path/to/repo […]
    </pre>


  * fstab: ``mount.borg#/path/to/repo /mnt/repomount/d fuse defaults,noauto,user 0 0``
    <pre>
    # mount d
    </pre>
    works, but it is ugly and not well documented in fuse itself




* borgfs

  * fstab: ``/path/to/repo /mnt/repomount/e fuse.borgfs defaults,noauto,user 0 0``
    <pre>
    # $ mount e
    </pre>
    works as non-root, as long as the user has access to the repo key and write access to the mount point

    borgfs#/path/to/repo /mnt/repomount/f fuse defaults,noauto,user 0 0
    <pre>
    # $ mount f
    </pre>
    works as non-root, as long as the user has access to the repo key and write access to the mount point

nachtgeist:
update PR for failed helpers.py:test_get_keys_dir()

borg/helpers.py
```diff
@@ -212,7 +212,11 @@ def show_progress(self, item=None, final=False, stream=None, dt=None):
 
 def get_keys_dir():
     """Determine where to repository keys and cache"""
-    xdg_config = os.environ.get('XDG_CONFIG_HOME', os.path.join(os.path.expanduser('~'), '.config'))
+    if os.environ.get('HOME'):
+        home_dir = os.environ.get('HOME')
+    else:
+        home_dir = os.path.expanduser('~%s' % os.environ.get('USER'))
```
ThomasWaldmann: if USER is not in environment, the .get() will return None. So I think this rather should be .get('USER', '').

borg/archiver.py
```diff
@@ -1153,6 +1153,13 @@ def build_parser(self, args=None, prog=None):
         browsing an archive or restoring individual files. Unless the ``--foreground``
         option is given the command will run in the background until the filesystem
         is ``umounted``.
+
+        The command ``borgfs`` provides a wrapper for ``borg mount``. This can also be
+        used in fstab entries:
+        ``/path/to/repo::archive /mnt/point fuse.borgfs defaults,noauto 0 0``
+
+        To allow a regular user to use fstab entries, add the ``user`` option:
+        ``/path/to/repo::archive /mnt/point fuse.borgfs defaults,noauto,user 0 0``
```
ThomasWaldmann: maybe we should rather just have the repo here without archive specification. the archive name changes all the time, so it is unlikely somebody would use that in fstab.

borg/helpers.py
ThomasWaldmann: btw, we have other code that uses the home dir (e.g. cache location), so does that need the same change also?

I am not sure why this needed fixing here, can you explain?

docs/usage/mount.rst.inc
```diff
@@ -38,3 +38,10 @@ This command mounts an archive as a FUSE filesystem. This can be useful for
 browsing an archive or restoring individual files. Unless the ``--foreground``
 option is given the command will run in the background until the filesystem
 is ``umounted``.
+
+The command ``borgfs`` provides a wrapper for ``borg mount``. This can also be
+used in fstab entries:
+``/path/to/repo::archive /mnt/point fuse.borgfs defaults,noauto 0 0``
+
+To allow a regular user to use fstab entries, add the ``user`` option:
+``/path/to/repo::archive /mnt/point fuse.borgfs defaults,noauto,user 0 0``
```
ThomasWaldmann: btw, did you produce this manually? there is a build_usage target that does create the .inc files automatically.

borg/helpers.py
nachtgeist: USER: good catch, thanks

expanduser: see commit msg, I explained it there

ThomasWaldmann:
The procedure needed by borg binary users is still missing.
They either need to create some small script manually or we need to add some code that looks at sys.argv[0] and calls mount_main if that is borgfs.

borg/archiver.py
nachtgeist: agreed, will remove ::archive

docs/usage/mount.rst.inc
nachtgeist: semi-manually; Thanks, I already used build_usage and then removed the leading whitespace since it was like that from the previous commit.

nachtgeist:
k, will check that, too

nachtgeist:
Just updated the PR with all suggestions implemented.

I'll do the separation of HOME handling into a separate function and updating other usages of HOME with it in a separate PR.

borg/archiver.py
```diff
@@ -1458,6 +1465,10 @@ def setup_signal_handlers():  # pragma: no cover
 
 
 def main():  # pragma: no cover
+    # provide borgfs behaviour when the main script/executable is named borgfs
+    if os.path.basename(sys.argv[0]) == "borgfs" and (len(sys.argv)<2 or sys.argv[1] != "mount"):
```
ThomasWaldmann: there's a pep8 error (strange that flake8 did not complain): add spaces around "<".

borg/archiver.py
```diff
@@ -1458,6 +1465,10 @@ def setup_signal_handlers():  # pragma: no cover
...
+	    sys.argv.insert(1, "mount")
```
ThomasWaldmann: is this indented correctly?

nachtgeist: nope. fixed.

borg/archiver.py
```diff
@@ -1458,6 +1465,10 @@ def setup_signal_handlers():  # pragma: no cover
 
 
 def main():  # pragma: no cover
+    # provide borgfs behaviour when the main script/executable is named borgfs
+    if os.path.basename(sys.argv[0]) == "borgfs" and (len(sys.argv) < 2 or sys.argv[1] != "mount"):
```
ThomasWaldmann: do we still need the sys.argv[1] != "mount" check?

nachtgeist: With main_mount() gone again, we indeed do not (works both for entry_point scripts and the binary). Thanks! (And I should be off to bed...)

ThomasWaldmann:
Please add some docs about borgfs when using the binary (you have it mentioned in the comment that there needs to be a symlink, but normal borg users won't read that).

PR #746, opened 2016-03-14
Build api.html reproducibly for ProgressIndicatorEndless/Percent

dannyedel:
Building the API docs for these functions capture the build host's I/O charset, making the build unreproducible.

This PR follows the style of #736, moving the evaluation of `sys.stderr` from the parameter list to the constructor.

### DIFF
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -974,7 +974,7 @@ def yes(msg=None, false_msg=None, true_msg=None, default_msg=None,
 
 
 class ProgressIndicatorPercent:
-    def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=sys.stderr):
+    def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=None):
         """
         Percentage-based progress indicator
 
@@ -989,6 +989,8 @@ class ProgressIndicatorPercent:
         self.total = total
         self.trigger_at = start  # output next percentage value when reaching (at least) this
         self.step = step
+        if file is None:
+            file = sys.stderr
         self.file = file
         self.msg = msg
         self.same_line = same_line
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
:+1:

borg/helpers.py
```diff
@@ -989,6 +989,8 @@ def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=
         self.total = total
         self.trigger_at = start  # output next percentage value when reaching (at least) this
         self.step = step
+        if file is None:
+            file = sys.stderr
```
enkore: You can write `file = file or sys.stderr`

borg/helpers.py
```diff
@@ -989,6 +989,8 @@ def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=
...
         self.file = file
```
enkore: You can write `self.file = file or sys.stderr` instead.

dannyedel: True, although i chose this to match the style used in 7abcaa0d9d34b9629b88b4e9ddf62ca0f76ade6b.

As a bonus, this (longer) form reads clearer, since it's almost an english sentence - "If file is none, file shall be stderr".

If @ThomasWaldmann doesn't object, I'd like to leave it like it is now (mainly for the symmetry reason to `archive.py`)

dannyedel:
BorgBackup in Debian (1.0.0-3 contains both 7abcaa0 and this patch here) now builds [fully reproducible][1], including the API docs.

For reference: Look at [1.0.0-2's diff][2] where `ANSI_X3.​4-​1968` or `UTF-8` was embedded.

[1]: https://tests.reproducible-builds.org/rb-pkg/unstable/amd64/borgbackup.html
[2]: https://tests.reproducible-builds.org/dbd/testing/amd64/borgbackup_1.0.0-2.diffoscope.html

ThomasWaldmann:
It's fine like it is.

I am not even sure I always stick to same style ("if" vs. "or").

PR #751, opened 2016-03-15
ItemFormatter for printing items, borg list redo.

enkore:
Current state:

- compatible except for {formatkeys}, which is gone
  - available keys now right in --help, with additional per-key help if deemed necessary
- supports -e, exclude-from, paths like the others
  (code should/will be shared, see feature-diff branch)
- supports guaranteed_available hashlib hashes
  (to avoid varying functionality depending on environment)
  (also, the other hashes are really obscure, like MD-4)

The two profiles [formerly linked here](https://github.com/borgbackup/borg/issues/735#issuecomment-196355903) were actually a mistake ; I wasn't running the second command inside the current environment and hence still used the old version (which is the reason why these are so similar -- should've been suspicious). The first approach was actually a **lot** slower, due to the many object creations (62 % of total runtime spent inside `builtins.__build_class__`!).

I made another approach, which is more stateful and does not use the class-wrapping approach as before. Together with some optimization of IO, mainly turning line_buffering off, this version and the version from `master` are indistinguishable, +-2 % over multiple runs, and this time I made sure I'm in the right envs ;-)

The overhead of adding additional "lazy" (now it's actually "on demand") keys is very low in this approach.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -16,12 +16,12 @@ import textwrap
 import traceback
 
 from . import __version__
-from .helpers import Error, location_validator, archivename_validator, format_line, format_time, format_file_size, \
-    parse_pattern, PathPrefixPattern, to_localtime, timestamp, safe_timestamp, \
+from .helpers import Error, location_validator, archivename_validator, format_time, format_file_size, \
+    parse_pattern, PathPrefixPattern, to_localtime, timestamp, \
     get_cache_dir, prune_within, prune_split, \
     Manifest, remove_surrogates, update_excludes, format_archive, check_extension_modules, Statistics, \
-    dir_is_tagged, bigint_to_int, ChunkerParams, CompressionSpec, is_slow_msgpack, yes, sysinfo, \
-    EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR, log_multi, PatternMatcher
+    dir_is_tagged, ChunkerParams, CompressionSpec, is_slow_msgpack, yes, sysinfo, \
+    EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR, log_multi, PatternMatcher, ItemFormatter
 from .logger import create_logger, setup_logging
 logger = create_logger()
 from .compress import Compressor, COMPR_BUFFER
@@ -585,79 +585,29 @@ class Archiver:
         repository = self.open_repository(args)
         manifest, key = Manifest.load(repository)
         if args.location.archive:
-            archive = Archive(repository, key, manifest, args.location.archive)
-            """use_user_format flag is used to speed up default listing.
-            When user issues format options, listing is a bit slower, but more keys are available and
-            precalculated.
-            """
-            use_user_format = args.listformat is not None
-            if use_user_format:
-                list_format = args.listformat
-            elif args.short:
-                list_format = "{path}{LF}"
-            else:
-                list_format = "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{LF}"
+            matcher, _ = self.build_matcher(args.excludes, args.paths)
 
-            for item in archive.iter_items():
-                mode = stat.filemode(item[b'mode'])
-                type = mode[0]
-                size = 0
-                if type == '-':
-                    try:
-                        size = sum(size for _, size, _ in item[b'chunks'])
-                    except KeyError:
-                        pass
+            with Cache(repository, key, manifest, lock_wait=self.lock_wait) as cache:
+                archive = Archive(repository, key, manifest, args.location.archive, cache=cache)
 
-                mtime = safe_timestamp(item[b'mtime'])
-                if use_user_format:
-                    atime = safe_timestamp(item.get(b'atime') or item[b'mtime'])
-                    ctime = safe_timestamp(item.get(b'ctime') or item[b'mtime'])
-
-                if b'source' in item:
-                    source = item[b'source']
-                    if type == 'l':
-                        extra = ' -> %s' % item[b'source']
-                    else:
-                        mode = 'h' + mode[1:]
-                        extra = ' link to %s' % item[b'source']
+                if args.format:
+                    format = args.format
+                elif args.short:
+                    format = "{path}{NL}"
                 else:
-                    extra = ''
-                    source = ''
+                    format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{NL}"
+                formatter = ItemFormatter(archive, format)
 
-                item_data = {
-                        'mode': mode,
-                        'user': item[b'user'] or item[b'uid'],
-                        'group': item[b'group'] or item[b'gid'],
-                        'size': size,
-                        'isomtime': format_time(mtime),
-                        'path': remove_surrogates(item[b'path']),
-                        'extra': extra,
-                        'LF': '\n',
-                        }
-                if use_user_format:
-                    item_data_advanced = {
-                        'bmode': item[b'mode'],
-                        'type': type,
-                        'source': source,
-                        'linktarget': source,
-                        'uid': item[b'uid'],
-                        'gid': item[b'gid'],
-                        'mtime': mtime,
-                        'isoctime': format_time(ctime),
-                        'ctime': ctime,
-                        'isoatime': format_time(atime),
-                        'atime': atime,
-                        'archivename': archive.name,
-                        'SPACE': ' ',
-                        'TAB': '\t',
-                        'CR': '\r',
-                        'NEWLINE': os.linesep,
-                        }
-                    item_data.update(item_data_advanced)
-                item_data['formatkeys'] = list(item_data.keys())
-
-                print(format_line(list_format, item_data), end='')
-
+                if not hasattr(sys.stdout, 'buffer'):
+                    # This is a shim for supporting unit tests replacing sys.stdout with e.g. StringIO,
+                    # which doesn't have an underlying buffer (= lower file object).
+                    def write(bytestring):
+                        sys.stdout.write(bytestring.decode('utf-8', errors='replace'))
+                else:
+                    write = sys.stdout.buffer.write
+                for item in archive.iter_items(lambda item: matcher.match(item[b'path'])):
+                    write(formatter.format_item(item).encode('utf-8', errors='surrogateescape'))
+            repository.close()
         else:
             for archive_info in manifest.list_archive_infos(sort_by='ts'):
                 if args.prefix and not archive_info.name.startswith(args.prefix):
@@ -944,12 +894,13 @@ class Archiver:
     def preprocess_args(self, args):
         deprecations = [
             # ('--old', '--new', 'Warning: "--old" has been deprecated. Use "--new" instead.'),
+            ('--list-format', '--format', 'Warning: "--list-format" has been deprecated. Use "--format" instead.'),
         ]
         for i, arg in enumerate(args[:]):
             for old_name, new_name, warning in deprecations:
                 if arg.startswith(old_name):
                     args[i] = arg.replace(old_name, new_name)
-                    print(warning)
+                    self.print_warning(warning)
         return args
 
     def build_parser(self, args=None, prog=None):
@@ -1322,7 +1273,12 @@ class Archiver:
 
         list_epilog = textwrap.dedent("""
         This command lists the contents of a repository or an archive.
-        """)
+
+        See the "borg help patterns" command for more help on exclude patterns.
+
+        The following keys are available for --format:
+
+        """) + ItemFormatter.keys_help()
         subparser = subparsers.add_parser('list', parents=[common_parser],
                                           description=self.do_list.__doc__,
                                           epilog=list_epilog,
@@ -1332,15 +1288,22 @@ class Archiver:
         subparser.add_argument('--short', dest='short',
                                action='store_true', default=False,
                                help='only print file/directory names, nothing else')
-        subparser.add_argument('--list-format', dest='listformat', type=str,
-                               help="""specify format for archive file listing
-                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}")
-                                Special "{formatkeys}" exists to list available keys""")
+        subparser.add_argument('--format', '--list-format', dest='format', type=str,
+                               help="""specify format for file listing
+                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NL}")""")
         subparser.add_argument('-P', '--prefix', dest='prefix', type=str,
                                help='only consider archive names starting with this prefix')
+        subparser.add_argument('-e', '--exclude', dest='excludes',
+                               type=parse_pattern, action='append',
+                               metavar="PATTERN", help='exclude paths matching PATTERN')
+        subparser.add_argument('--exclude-from', dest='exclude_files',
+                               type=argparse.FileType('r'), action='append',
+                               metavar='EXCLUDEFILE', help='read exclude patterns from EXCLUDEFILE, one per line')
         subparser.add_argument('location', metavar='REPOSITORY_OR_ARCHIVE', nargs='?', default='',
                                type=location_validator(),
                                help='repository/archive to list contents of')
+        subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
+                               help='paths to extract; patterns are supported')
 
         mount_epilog = textwrap.dedent("""
         This command mounts an archive as a FUSE filesystem. This can be useful for
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -1,8 +1,9 @@
 import argparse
 from binascii import hexlify
 from collections import namedtuple
-from functools import wraps
+from functools import wraps, partial
 import grp
+import hashlib
 import os
 import stat
 import textwrap
@@ -10,6 +11,7 @@ import pwd
 import re
 from shutil import get_terminal_size
 import sys
+from string import Formatter
 import platform
 import time
 import unicodedata
@@ -548,6 +550,20 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
     return tag_paths
 
 
+def partial_format(format, mapping):
+    """
+    Apply format.format_map(mapping) while preserving unknown keys
+
+    Does not support attribute access, indexing and ![rsa] conversions
+    """
+    for key, value in mapping.items():
+        key = re.escape(key)
+        format = re.sub(r'(?<!\{)((\{%s\})|(\{%s:[^\}]*\}))' % (key, key),
+                        lambda match: match.group(1).format_map(mapping),
+                        format)
+    return format
+
+
 def format_line(format, data):
     # TODO: Filter out unwanted properties of str.format(), because "format" is user provided.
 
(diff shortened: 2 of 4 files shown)


borg/archiver.py
```diff
@@ -1458,10 +1419,8 @@ def setup_signal_handlers():  # pragma: no cover
 
 
 def main():  # pragma: no cover
-    # Make sure stdout and stderr have errors='replace') to avoid unicode
-    # issues when print()-ing unicode file names
-    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, sys.stdout.encoding, 'replace', line_buffering=True)
-    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, sys.stderr.encoding, 'replace', line_buffering=True)
+    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, sys.stdout.encoding, 'surrogateescape', line_buffering=True)
+    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, sys.stderr.encoding, 'surrogateescape', line_buffering=True)
```
enkore: I don't think this is necessary anymore (at all) - maybe for stderr, but I think all places use remove_surrogates when printing to stdout.

borg/archiver.py
```diff
@@ -1140,15 +1093,22 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('--short', dest='short',
                                action='store_true', default=False,
                                help='only print file/directory names, nothing else')
-        subparser.add_argument('--list-format', dest='listformat', type=str,
-                               help="""specify format for archive file listing
-                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}")
-                                Special "{formatkeys}" exists to list available keys""")
+        subparser.add_argument('-f', '--format', '--list-format', dest='format', type=str,
```
ThomasWaldmann: be careful with short options, they easily collide or have different meanings at different places.
also, we might want to use -f for something else in future, like --force?

borg/archiver.py
```diff
@@ -1,9 +1,10 @@
 from binascii import hexlify, unhexlify
 from datetime import datetime
-from hashlib import sha256
 from operator import attrgetter
+from string import Formatter
```
ThomasWaldmann: unused import?

borg/helpers.py
```diff
@@ -546,7 +548,7 @@ def format_line(format, data):
     except (KeyError, ValueError) as e:
         # this should catch format errors
         print('Error in lineformat: "{}" - reason "{}"'.format(format, str(e)))
-    except:
+    except Exception as e:
```
ThomasWaldmann: oops? looks like there was a bug?

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+        formatter = cls(FakeArchive, "")
+        keys = []
+        keys.extend(formatter.call_keys.keys())
+        keys.extend(formatter.get_item_data({
+            b'mode': 0, b'path': '', b'user': '', b'group': '', b'mtime': 0,
+            b'uid': 0, b'gid': 0,
+        }).keys())
+        keys.sort()
+        keys.sort(key=str.isupper)
+        return keys
```
ThomasWaldmann: return sorted(keys, ...)

borg/archiver.py
enkore: Agreed. --format is short enough anyway, --list-format is kinda "strange" however (*what other format would it be?*)

borg/archiver.py
enkore: Yep, remnant of moving the ItemFormatter class from archiver.Archiver to helpers.

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+            'isomtime': partial(self.format_time, b'mtime'),
+            'isoctime': partial(self.format_time, b'ctime'),
+            'isoatime': partial(self.format_time, b'atime'),
+            'mtime': partial(self.time, b'mtime'),
+            'ctime': partial(self.time, b'ctime'),
+            'atime': partial(self.time, b'atime'),
+        }
+        for hash_function in hashlib.algorithms_guaranteed:
+            self.call_keys[hash_function] = partial(self.hash_item, hash_function)
+        self.keys = set(self.call_keys) & self.format_keys
```
ThomasWaldmann: intersection? if you wanted union, that is |.

enkore: Intersection is what we want:  call_keys are alle keys calculated from callables, format_keys are those used by the format string. The intersecting set are those keys used by the format string which are contained in call_keys and need to be sourced from there.

Variable naming should be improved :D

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+            'gid': item[b'gid'],
+
+            'path': remove_surrogates(item[b'path']),
+            'bpath': item[b'path'],
+            'source': source,
+            'linktarget': source,
+            'extra': extra
+        })
+        for key in self.keys:
+            item_data[key] = self.call_keys[key](item)
```
enkore: Note that there is no lazyness here: all requested keys are calculated, unconditionally. The "lazyness" is that we check beforehand which keys need calculating.

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+
+    def format_item(self, item):
+        return self.format.format(**self.get_item_data(item))
+
+    def calculate_size(self, item):
+        if stat.S_ISREG(item[b'mode']):
+            try:
+                return sum(size for _, size, _ in item[b'chunks'])
+            except KeyError:
+                pass
```
ThomasWaldmann: how about:

return sum(chunk[1] for chunk in item.get(b'chunks', []))

The isreg check is maybe rather harmful, borg can backup the contents of devices and then isreg might be false, but chunks still in there.

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+    def hash_item(self, hash_function, item):
+        if b'chunks' not in item:
+            return ""
+        hash = hashlib.new(hash_function)
+        for chunk in self.archive.pipeline.fetch_many([c[0] for c in item[b'chunks']]):
+            hash.update(chunk)
+        return hash.hexdigest()
+
+    def format_time(self, key, item):
+        return format_time(safe_timestamp(item[key]))
```
ThomasWaldmann: ctime and atime is not aways there, should use mtime if not.

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+
+    def time(self, key, item):
+        return safe_timestamp(item[key])
```
ThomasWaldmann: same here.

borg/helpers.py
ThomasWaldmann: ah, right, misread that at first.

borg/helpers.py
enkore: Seems like there was a little bit "lost in translation" :)

I'm going with `timestamp_handling_fn(item.get(key) or item[b'mtime]))` to get eval-once behaviour most of the time. `item.get(key, item[b'mtime])` would be more conventional, but these pieces are literally called a gazillion times...

borg/helpers.py
```diff
@@ -1080,3 +1082,111 @@ def log_multi(*msgs, level=logging.INFO):
...
+        class FakeArchive:
+            name = ""
+
+        formatter = cls(FakeArchive, "")
+        keys = []
+        keys.extend(formatter.call_keys.keys())
+        keys.extend(formatter.get_item_data({
+            b'mode': 0, b'path': '', b'user': '', b'group': '', b'mtime': 0,
+            b'uid': 0, b'gid': 0,
+        }).keys())
```
ThomasWaldmann: fake_item = {...}
keys = formatter.call_keys.keys() + formatter.get_item_data(fake_item).keys()

enkore: I don't think the dict-view types are addable.

ThomasWaldmann:
hmm, travis seems unhappy.

but overall great change - using this we can implement a lot of stuff by just extending the available format keys (like e.g. file compressed size, compression ratio, (un)compressed unique chunks size, count of unique chunks, ...

borg/helpers.py
ThomasWaldmann: oh, I was still thinking in lists, like in python2. yes in py3, that does not work.

that would work, not sure if it helps with readability:

fake_item = {...}
keys = set(formatter.call_keys) | set(formatter.get_item_data(fake_item))

borg/helpers.py
ThomasWaldmann: if the timestamp is 0, that does not work.

borg/helpers.py
```diff
@@ -1107,9 +1114,17 @@ class FakeArchive:
             b'mode': 0, b'path': '', b'user': '', b'group': '', b'mtime': 0,
             b'uid': 0, b'gid': 0,
         }).keys())
-        keys.sort()
-        keys.sort(key=str.isupper)
-        return keys
+        return sorted(sorted(keys), key=str.isupper)
```
ThomasWaldmann: why do you double-sort? is the inner sort doing anything that affects the outer result?

borg/archiver.py
```diff
@@ -438,78 +438,31 @@ def do_list(self, args):
...
-                item_data['formatkeys'] = list(item_data.keys())
-
-                print(format_line(list_format, item_data), end='')
-
+                format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{LF}"
+            formatter = ItemFormatter(archive, format)
+            if not hasattr(sys.stdout, 'buffer'):
+                # This is a shim for supporting unit tests replacing sys.stdout with e.g. StringIO,
+                # which doesn't have an underlying buffer (= lower file object).
+                write = lambda bytestring: sys.stdout.write(bytestring.decode('utf-8', errors='replace'))
```
ThomasWaldmann: there was no major change in the testsuite, so how does it come we suddenly need this?

borg/helpers.py
enkore: first alphabetical sort, then separate keys only composed of uppercase letters to the bottom. This way the "formatting" aids (TAB, NEWLINE, LF, CR, NUL) appear together at the bottom, not mixed between content keys.

borg/archiver.py
enkore: The new code uses sys.stdout.buffer (which is always a binary file) for two reasons

1. to have errors=surrogateescape, so we can print binary strings (=> `{bpath}`). The default stdout we create in main() has the errors=replace policy, which prohibits this.
2. it circumvents the line buffering (…actually looks like "line-fsyncing" for writes) of the stdout TextIOWrapper. This makes IO to stdout significantly faster: with approx identical total run-time `print` in the old code took 12 % of the total run time, while `write` through this object only takes 1.6 % of the total run time.

This is however incompatible with StringIO, which doesn't have the underlying object as a `buffer` attribute, because it isn't a IOWrapper.

borg/helpers.py
ThomasWaldmann: ah, interesting. i somehow had expected that the outer sort might disturb the inner sort result.

enkore:
master branch callmap:

borg list master

this patch (yesterday, today should be slightly better):

borg list lazy

borg/helpers.py
enkore: sort is stable, but I just remembered that tuples also work for keys, so that's simpler. Will include that in the next patch rev.

borg/testsuite/archiver.py
```diff
@@ -898,7 +898,7 @@ def test_list_list_format(self):
         self.cmd('create', test_archive, src_dir)
         output_1 = self.cmd('list', test_archive)
         output_2 = self.cmd('list', '--list-format', '{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}', test_archive)
-        output_3 = self.cmd('list', '--list-format', '{mtime:%s} {path}{NL}', test_archive)
+        output_3 = self.cmd('list', '--list-format', '{mtime:%s} {path}{NEWLINE}', test_archive)
```
enkore: While I still don't know where NL came from in the old code I'm just going to revert this and add NL as an alias for NEWLINE. No one wants to type 10 characters for a new line anyway.

borg/archiver.py
ThomasWaldmann: it somehow seems inappropriate to have a different way of doing output just for borg list.

enkore: I think it's ok, since line buffering for list doesn't make a lot of sense, but for the other commands it does. The errors-policy is another thing, it's probably okay in most other cases to remove non-text chars from paths, but I think for list we should be able to [reliably] support what posix demands.

enkore:
Current performance

Test case: an archive of my /home here, about 1.5 million files. List goes straight into /dev/null. FS caches are hot (2 runs each to warm them up, then average across another three runs).

master: 1m17s (1m15s .. 1m21s)
93a6515: 1m11s +- 1 s

From profiling of this test case I'd say this version is faster for two reasons
* IO / writes to stdout is three times faster (3.5 % vs. 10 % of run time)
* *Slightly* less per-item overhead: no large objects are created per item except the item itself, and we use format_map instead of format, which is about 2 % faster (I made a synthetic benchmark which were closely matches this figure from the profile data).
* No free lunch: the extra calls for size seem to add a little overhead to that block

So in summary, we're getting a little faster, but most importantly, we're *not* slower.

borg/helpers.py
```diff
@@ -1080,3 +1082,146 @@ def log_multi(*msgs, level=logging.INFO):
...
+        'LF': '\n',
+        'SPACE': ' ',
+        'TAB': '\t',
+        'CR': '\r',
+        'NUL': '\0',
+        'NEWLINE': os.linesep,
+        'NL': os.linesep,
+    }
+    KEY_DESCRIPTIONS = {
+        'NL': 'alias of NEWLINE (OS dependent line separator)',
```
ThomasWaldmann: you say that NL is an alias of NEWLINE, but you don't have a description for NEWLINE.

borg/helpers.py
```diff
@@ -1080,3 +1082,146 @@ def log_multi(*msgs, level=logging.INFO):
...
+            'ctime': partial(self.time, b'ctime'),
+            'atime': partial(self.time, b'atime'),
+        }
+        for hash_function in hashlib.algorithms_guaranteed:
+            self.add_key(hash_function, partial(self.hash_item, hash_function))
+        self.used_call_keys = set(self.call_keys) & self.format_keys
+        self.item_data = static_keys
+
+    def partial_format(self, format, mapping):
+        # str.format wishlist: leave_unknown_keys_alone_and_dont_throw_up=True
```
ThomasWaldmann: maybe rather document what this does

enkore: I'll also add a few tests for it, since it's non-trivial

ThomasWaldmann: i just thought how it will behave for keys that are substrings of other keys...

borg/helpers.py
```diff
@@ -540,6 +540,16 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
     return tag_paths
 
 
+def partial_format(format, mapping):
+    """Apply format.format_map(mapping) while preserving unknown keys"""
+    for key, value in mapping.items():
+        def a(x): print(x.groups()); return ""
```
ThomasWaldmann: hm? left over debug code?

enkore: Yep.

enkore:
Some of these commits were breaking, so I will squash them when review is finished.

borg/archiver.py
```diff
@@ -1332,15 +1286,22 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('--short', dest='short',
                                action='store_true', default=False,
                                help='only print file/directory names, nothing else')
-        subparser.add_argument('--list-format', dest='listformat', type=str,
-                               help="""specify format for archive file listing
-                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}")
-                                Special "{formatkeys}" exists to list available keys""")
+        subparser.add_argument('--format', '--list-format', dest='format', type=str,
```
ThomasWaldmann: maybe this should use the cmdline option deprecation mechanism, so we can get rid of the --list-format later.
grep for "deprec" in this file.

ThomasWaldmann:
aside from deprecation handling and the squashing you wanted to do, i think this is ready.

so tell if you are ready with it, so I'll merge it.

enkore:
The deprecation mechanism prints to stdout:
```
    def preprocess_args(self, args):
        deprecations = [
            # ('--old', '--new', 'Warning: "--old" has been deprecated. Use "--new" instead.'),
            ('--list-format', '--format', 'Warning: "--list-format" has been deprecated. Use "--format" instead.'),
        ]
        for i, arg in enumerate(args[:]):
            for old_name, new_name, warning in deprecations:
                if arg.startswith(old_name):
                    args[i] = arg.replace(old_name, new_name)
                    print(warning)
        return args
```

Should I change that to self.print_warning? Seems more logical to me to not mix command output on stdout with these warnings.

ThomasWaldmann:
yes, call print_warning. this will also set warning return code, but i guess that'll help fixing deprecated stuff in scripts.

enkore:
Will do.

The Travis CI fail for 3417906 looks strange, maybe a flaky / one time error?

```
    def test_list_list_format(self):
        test_archive = self.repository_location + '::test'
        self.cmd('create', test_archive, src_dir)
        output = self.cmd('list', '--list-format', '-', test_archive)
        assert output.startswith('Warning: "--list-format" has been deprecated')
        output_1 = self.cmd('list', test_archive)
        output_2 = self.cmd('list', '--format', '{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}', test_archive)
        output_3 = self.cmd('list', '--format', '{mtime:%s} {path}{NL}', test_archive)
>       self.assertEqual(output_1, output_2)
E       AssertionError: 'drwxr-xr-x root   staff         0 Thu, 20[16944 chars]py\n' != 'Exception ignored in: <bound method Repos[17732 chars]py\n'
E       Diff is 17887 characters long. Set self.maxDiff to None to see it.
```

I guess where py.test cut off an exception in some Repository method occured.

ThomasWaldmann:
I restarted the failing test run, let's see if it fails again.

enkore:
I saw it once here now, as well. It's a LockFailed exception downwards of Repository.__del__():

```
Exception ignored in: <bound method Repository.__del__ of <Repository /tmp/tmpjo24yn28/repository>>
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 67, in __del__
    self.close()
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 159, in close
    self.lock.release()
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 293, in release
    with self._lock:
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 117, in __enter__
    return self.acquire()
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 140, in acquire
    raise LockFailed(self.path, str(err)) from None
borg.locking.LockFailed: ('/tmp/tmpjo24yn28/repository/lock.exclusive', "[Errno 2] No such file or directory: '/tmp/tmpjo24yn28/repository/lock.exclusive'")
```

ThomasWaldmann:
yes, there is already a ticket about further locking improvements (which will likely be done by making repository a context manager).

are you finished or do you still need to squash things?

enkore:
I'm done, good to merge, when Travis gives the word.

Issue #628, opened 2016-02-04
consider --ignore-inode option


ThomasWaldmann:
borg compares mtime, size, inode number of backup set files against the files cache to see whether they were modified (qualify for chunking / backup) or are still unmodified (quickly skip them).

the inode part of that can be troublesome for non-unix filesystems (see #624), so we could introduce an option to ignore the inode number. The check is less confident then, so that should be only done if there are no real inode numbers and this causes a malfunction.

ThomasWaldmann:
see also: https://github.com/jborg/attic/issues/89

PR #752, opened 2016-03-15
Feature --ignore--inode

enkore:
Solves the issue with pull backups taking ages, because sshfs does not forward inode numbers from the server[1].

[1] I have not researched why, I guess it is a fundamental problem of sorts - if it were a simple "fix" they would just do it, I suppose.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -964,6 +964,12 @@ class Archiver:
         traversing all paths specified. The archive will consume almost no disk space for
         files or parts of files that have already been stored in other archives.
 
+
+        To speed up pulling backups over sshfs and similar network file systems which do
+        not provide correct inode information the --ignore-inode flag can be used. This
+        potentially decreases reliability of change detection, while avoiding always reading
+        all files on these file systems.
+
         See the output of the "borg help patterns" command for more help on exclude patterns.
         """)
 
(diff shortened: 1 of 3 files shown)


borg/archiver.py
```diff
@@ -1018,6 +1025,9 @@ def build_parser(self, args=None, prog=None):
                                type=ChunkerParams, default=CHUNKER_PARAMS,
                                metavar='CHUNK_MIN_EXP,CHUNK_MAX_EXP,HASH_MASK_BITS,HASH_WINDOW_SIZE',
                                help='specify the chunker parameters. default: %d,%d,%d,%d' % CHUNKER_PARAMS)
+        subparser.add_argument('--ignore-inode', dest='ignore_inode',
+                               action='store_true', default=False,
```
enkore: I think `action='store_true', default=False,` is redundant, but it's always used here. Maybe something for a little cleanup session.

enkore:
stapling #628

piotrjurkiewicz:
This feature is useful also on Virtualbox shared folders (vboxsf filesystem). Inodes numbers are not constant between mounts. It would be good to mention this in help as well.

ThomasWaldmann:
There are likely quite some filesystem that do not have stable inode numbers, so I guess we rather just document our option and what it does than trying to enumerate them all (documenting that a fs does not have stable inodes is in scope of the fs documentation).

enkore:
Q: Also push this into 1.0.1, to "fix" the performance issues with sshfs and others for 1.0.x users?

ThomasWaldmann:
@enkore yes, makes sense.

PR #756, opened 2016-03-16
[Protoype] Feature recompress

enkore:
THIS _WILL_ DESTROY YOUR DATA¹

¹ well it shouldn't but it's your fault if it does shred everything ;)

---

That being said, it seems to work quite fine. **There are TODOs[1][3]. It has no tests.** I tested with an repo containing current Linux sources (about 700 MB or so extracted). First simple create, no compression. Then recompress with LZ4. Repository shrinks to 250 MB.

This feature branch is based on feature-newlist[2], because I used that to generate SHA256 lists and diffed those to verify that Borg successfully reads the recompressed repo.

[1] notably first only decrypting, not decompressing, then detecting what algo the chunk in question uses, and if it is the target algo, skip the chunk and don't _compress, encrypt and store_ it _again_.
[2] The changes are far apart so tracking squishes, rebases and the like should be no problem.
[3] I think it should be Ctrl-C-safe already, however, when I did that it crashed on the next run somewhere in the middle:

    ...............................................................................
    Found manifest - end reached
    Local Exception.
      File "/home/mabe/largehome/Projekte/_oss/borg/borg/archiver.py", line 1510, in main
      File "/home/mabe/largehome/Projekte/_oss/borg/borg/archiver.py", line 1464, in run
        return args.func(args)
      File "/home/mabe/largehome/Projekte/_oss/borg/borg/archiver.py", line 599, in do_recompress
        repository.commit()
      File "/home/mabe/largehome/Projekte/_oss/borg/borg/repository.py", line 167, in commit
        self.write_index()
      File "/home/mabe/largehome/Projekte/_oss/borg/borg/repository.py", line 211, in write_index
        self.index.write(os.path.join(self.path, 'index.tmp'))
    AttributeError: 'NoneType' object has no attribute 'write'

So that needs investigating beside other things.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -451,9 +451,9 @@ class Archiver:
             if args.format:
                 format = args.format
             elif args.short:
-                format = "{path}{LF}"
+                format = "{path}{NL}"
             else:
-                format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{LF}"
+                format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{NL}"
             formatter = ItemFormatter(archive, format)
             if not hasattr(sys.stdout, 'buffer'):
                 # This is a shim for supporting unit tests replacing sys.stdout with e.g. StringIO,
@@ -1101,7 +1101,7 @@ class Archiver:
                                help='only print file/directory names, nothing else')
         subparser.add_argument('--format', '--list-format', dest='format', type=str,
                                help="""specify format for file listing
-                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NEWLINE}")""")
+                                (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NL}")""")
         subparser.add_argument('-P', '--prefix', dest='prefix', type=str,
                                help='only consider archive names starting with this prefix')
         subparser.add_argument('-e', '--exclude', dest='excludes',
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -577,30 +577,49 @@ class Archiver:
 
         from borg.repository import TAG_PUT
 
+        segment_count = sum(1 for _ in repository.io.segment_iterator())
         last_segment_id = repository.io.get_latest_segment()
-        print(last_segment_id, "segments in repository")
+        print(segment_count, "segments in repository")
+        old_size = 0
+        new_size = 0
         for segment_id, segment_filename in repository.io.segment_iterator():
             if segment_id > last_segment_id:
                 print("Reached beginning of recompressed segments")
                 break
+            chunks = {}
             for tag, id_, offset, data in repository.io.iter_objects(segment_id, True):
                 if tag != TAG_PUT:
                     continue
                 if set(id_) == {0}:
                     print("\nFound manifest - end reached")
-                    break
+                    continue
                 # TODO: add decompress kwarg to key.decrypt, so we can detect manually what compressor
                 # TODO: was used here, and skip the chunk if it uses key.compressor
                 # validate, decrypt, decompress, compress, encrypt
+                old_size += len(data)
                 data = key.encrypt(key.decrypt(id_, data))
-                if not dry_run:
-                    repository.put(id_, data)
+                new_size += len(data)
+                chunks[id_] = data
             if not dry_run:
-                repository.commit()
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
cool, features!!! :-D

about [1]: guess if we'll need a --force-recompress because many compression algorithms are not just the algorithm, but support also a compression level - and it should be possible to change that also.

enkore:
There is probably one or the another bug hiding the MT code (as usual). Using concurrent.futures is not an option, the overhead is *enormous* compared to manually managing queues.

Speed up remains insignificant unless the compression op becomes slower than AES, which always runs in the main thread (and I don't think we can effectively change that, at least for encryption, due to the repository-wide CTR?). There is a hard-coded threshold for small chunks which are not send to the compression threads. This is probably a tad high for high LZMA levels.

On my non-AES-NI machine this basically means zlib,>6 or LZMA. zlib,9 gives about 3x speed up with 5 cores for compression on a 6 core CPU. Didn't test LZMA yet. If it releases the GIL properly this should be well suited for it.

Memory usage can be somewhat limited, but we have no influence over the additional buffers used by the compressor. I think for the ~2 MB chunk size we use even the highest LZMA setting shouldn't exceed 50 MB or so.

---

I think I got repository, transaction and segment handling right now, at least the basics. It is not ctrl-c safe[1] in all cases, so something is still a miss. Also, for some reason not deleting the (old) Manifest manually can lead to duplicated manifest in some cases (which is unclean, I guess, but should not cause troubles itself).

[1] clean exit at any team would make it handy as an idle task for backup servers.

ThomasWaldmann:
no multithreading in master branch.

we will have multithreading (in the branch named like that), but not now in master.

ThomasWaldmann:
the reason is that MT code often introduces all sorts of non-trivial bugs, race conditions, etc..

so I suggest to have a non-MT implementation of this in master and have a MT implementation later in multithreading repo branch (which will need much more work and testing until it will be merged into master).

enkore:
I think that the complexity in this operation is much better controllable than in `create`, but I do agree with you that any MT code in an application like Borg requires *rigorous* testing before releasing it.

I'll continue work on the single-threaded branch and be back later with that.

ThomasWaldmann:
thanks! :)

enkore:
What I couldn't wrap my head around was why LZMA performed so poor in a multithreaded environment (just 130 % CPU load). Poking a around in Modules/_lzmamodule.c shows they do this when compressing (lines 512 ff, 148 ff):

1. Make a buffer of BUFSIZ (that's 8096 bytes on my systems)
2. Free GIL
3. Call into LZMA compression
4. Acquire GIL
5. "That 2 MB chunk didn't compress into 8 K after all"
6. Enlarge buffer (that is, realloc) by current size / 8 (so first iteration 8 K, next ~9K, 10 K, …)
7. Goto 2 until done

Using LZMA in a multi-threaded environement basically turns into a dog-fight over the GIL between compressing small chunks of data.

So MT doesn't make any sense to pursue until that gets patched (and if it's just a better guess at the output buffer size than "8K"; min(8K, input_size / 5) sounds reasonable to me).

Anyway, this is a bit OT now. Just wanted to put it here, so if multithreading comes up again we remember this not entirely irrelevant info.

ThomasWaldmann:
sounds like that should be put into python issue tracker, can you do that?

enkore:
I investigate the LZMA issue further and while this buffer handling is part of the issue, I think the main problem is a level further up. A while back I wrote a small benchmark to check how well simple ops can be multithreaded in Python.

When doing lzma with preset=0 it works reasonably ok (25 MB/s for six cores is absolutely lousy, but at least full load), but using any serious (1..9) preset basically *breaks* it. I.e. <<1 MB/s.

For reference, with zlib in the exact same benchmark I have no issues at all and happily achieve >150 MB/s and >550 % CPU at the highest level. When doing adler32 in this benchmark it yields 9-10 GB/s (on a machine that only has ~20 GB/s memory bandwidth).

I conclude that while the lzma module might be thread-safe, it is not fit for use in multi-threaded applications. I'll consider opening an issue in CPython's tracker about this (but I have my doubts about the eventuality of a fix; mt is one of the ugliest corners in all of Python). In the mean time we might consider (that is, if work commences on MT features) building our own (minimal) liblzma binding similar to how we handle LZ4.

ThomasWaldmann:
Yes, I think opening an issue in CPython tracker is advisable. Attach your code there so they can reproduce.

PR #759, opened 2016-03-17
refactor get_home_dir

nachtgeist:


borg/helpers.py
```diff
@@ -211,19 +211,24 @@ def show_progress(self, item=None, final=False, stream=None, dt=None):
...
     # os.path.expanduser() behaves differently for '~' and '~someuser' as
     # parameters: when called with an explicit username, the possibly set
     # environment variable HOME is no longer respected. So we have to check if
     # it is set and only expand the user's home directory if HOME is unset.
     if os.environ.get('HOME', ''):
-        home_dir = os.environ.get('HOME')
+        return os.environ.get('HOME')
     else:
-        home_dir = os.path.expanduser('~%s' % os.environ.get('USER'))
+        return os.path.expanduser('~%s' % os.environ.get('USER'))
```
ThomasWaldmann: if USER is not set, this will fail. 

.get('USER', '') maybe?

nachtgeist: Thanks! Somehow the unfixed version of that assignment made it into the last PR. Sorry about that.

Issue #761, opened 2016-03-17
Extracting archive with hard-linked files and --strip-components fails


nprncbl:
I use borg to backup dovecot maildirs. Dovecot uses hard-links for copied mail to avoid storing messages multiple times if users copy them.

When you restore an archive with a backup of a Dovecot maildir and that maildir contains hard-links, the restore works just fine. Look at this example of a restored mail message that has been copied by the user from the inbox to some other folder:

`root@node4:/tmp# ls -l /tmp/large/vmail/somedomain.com/someuser/.Newsletter.Logistics\ Viewpoints/cur/1457623913.M965480P3612*`

`-rw-r--r-- 2 vmail vmail 59K Mar 10 15:31 /tmp/large/vmail/somedomain/someuser/.Newsletter.Logistics Viewpoints/cur/1457623913.M965480P3612.node4,S=59664,W=60637:2,Sab`

Notice the two links to this file.

The same message can be found in the restored inbox of this user, and it is hard-linked to the same file on disk. Perfect!

Now, let's use borg's great --strip-components feature, as we really do not care for all that prefixed stuff and only want to restore the somedomain part of the archive:

`root@node4:/tmp# time borg extract --verbose --sparse --strip-components 2 /path-to-borg-repo/vmail.borg::vmail.2016-03-17.05-00 large/vmail/somedomain`

The extraction works fine in general, but it complains about missing directories or files when trying to make the hard-links contained in the archive:

`large/vmail/somedomain/someuser/cur/1457623913.M965480P3612.node4,S=59664,W=60637:2,STac: [Errno 2] No such file or directory: '/tmp/large/vmail/somedomain/someuser/.Newsletter.Logistics Viewpoints/cur/1457623913.M965480P3612.node4,S=59664,W=60637:2,Sab' -> '/tmp/somedomain/someuser/cur/1457623913.M965480P3612.node4,S=59664,W=60637:2,STac'`

To me, it seems that the --strip-components option is not taken into account when computing the source and destination for hard-linking files.

Please let me know if you need any more information or clarification.

Thanks for working on borg - it is an awesome piece of backup software!

enkore:
Interesting find.

For symlinks it's probably best to not change anything, they should be stored as absolute paths anyway.

For hardlinks the file should extracted like a regular file if the source lies outside the --strip-components subtree.

enkore:
be51b91 contains a failing test case demonstrating the issue.

[...]

PR #763, opened 2016-03-17
Issue #761: Links fail for extracting subtrees

enkore:
**Issue summary:**

When extracting subtrees ("partial_extract") by using any of (excludes, PATH, strip-components) hard links might fail, because Borg tried to link to non-existing files, if the first appearance of a (inode, device) tuple (which uniquely identifies files and is identical across hard links) was not extracted.

**Implementation:**

* creating a mapping that stores *all* file paths so far and their chunk list or link source. (We need to store all seen files, because we can't tell if a file is hard-linked)
* when extracting a hard link: resolving whether an instance of the link was extracted
  * if it was not, the file contents will be extracted
     * and the entry in our "master file mapping" will be changed to point to this file
  * if it was, a hard link is created

**Drawbacks:**

* a possibly very significant increase in memory usage when extracting subtrees, because the "master file mapping" will essentially grow until all files in the archive were visited, and at that point it will contain all file paths and chunk lists.

However, if no partial extraction is done, the "MFM" is not created, not used and has therefore no impact.

**Won't somebody please think of the ~~children~~ soft links?**

Soft links are always stored as absolute paths and always extracted as such and are not affected by this.

**TODO**

* [x] check whether this handles link metadata correctly / as expected, create test or validate that test exists and indeed covers this additional code path
   * they are the same files, so same U/G ID, permissions, xattrs and ACLs, even on Windows/NTFS.
* [x] Xcode builds broken (some minor thing with output paths or something)
* [x] This "breaks" preloading (now only for "old" archives)
* [x] Memory hit (now only for "old" archives)
   * [x] DONE We could include a nlink / a flag whether nlink > 1 for hard link items. That way the "MFM" would only need to contain these items, and items can be directly filtered again (patternmatcher or nlink), which also fixes preloading (*but only for new archives*). We can't use this for reverse ref-counting, though, because nlink = x doesn't necessarily imply that there are (x-1) items hardlinked to it.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -298,7 +298,19 @@ Number of files: {0.stats.nfiles}'''.format(
         cache.rollback()
         return stats
 
-    def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False):
+    def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False,
+                     hardlink_masters=None, original_path=None):
+        """
+        Extract archive item.
+
+        :param item: the item to extract
+        :param restore_attrs: restore file attributes
+        :param dry_run: do not write any data
+        :param stdout: write extracted data to stdout
+        :param sparse: write sparse files (chunk-granularity, independent of the original being sparse)
+        :param hardlink_masters: maps paths to (chunks, link_target) for extracting subtrees with hardlinks correctly
+        :param original_path: b'path' key as stored in archive
+        """
         if dry_run or stdout:
             if b'chunks' in item:
                 for data in self.pipeline.fetch_many([c[0] for c in item[b'chunks']], is_preloaded=True):
@@ -308,6 +320,7 @@ Number of files: {0.stats.nfiles}'''.format(
                     sys.stdout.buffer.flush()
             return
 
+        original_path = original_path or item[b'path']
         dest = self.cwd
         if item[b'path'].startswith('/') or item[b'path'].startswith('..'):
             raise Exception('Path should be relative and local')
@@ -327,25 +340,36 @@ Number of files: {0.stats.nfiles}'''.format(
         if stat.S_ISREG(mode):
             if not os.path.exists(os.path.dirname(path)):
                 os.makedirs(os.path.dirname(path))
+
             # Hard link?
             if b'source' in item:
                 source = os.path.join(dest, item[b'source'])
                 if os.path.exists(path):
                     os.unlink(path)
-                os.link(source, path)
-            else:
-                with open(path, 'wb') as fd:
-                    ids = [c[0] for c in item[b'chunks']]
-                    for data in self.pipeline.fetch_many(ids, is_preloaded=True):
-                        if sparse and self.zeros.startswith(data):
-                            # all-zero chunk: create a hole in a sparse file
-                            fd.seek(len(data), 1)
-                        else:
-                            fd.write(data)
-                    pos = fd.tell()
-                    fd.truncate(pos)
-                    fd.flush()
-                    self.restore_attrs(path, item, fd=fd.fileno())
+                if not hardlink_masters:
+                    os.link(source, path)
+                    return
+                item[b'chunks'], link_target = hardlink_masters[item[b'source']]
+                if link_target:
+                    # Hard link was extracted previously, just link
+                    os.link(link_target, path)
+                    return
+                # Extract chunks, since the item which had the chunks was not extracted
+            with open(path, 'wb') as fd:
+                ids = [c[0] for c in item[b'chunks']]
+                for data in self.pipeline.fetch_many(ids, is_preloaded=True):
+                    if sparse and self.zeros.startswith(data):
+                        # all-zero chunk: create a hole in a sparse file
+                        fd.seek(len(data), 1)
+                    else:
+                        fd.write(data)
+                pos = fd.tell()
+                fd.truncate(pos)
+                fd.flush()
+                self.restore_attrs(path, item, fd=fd.fileno())
+            if hardlink_masters:
+                # Update master entry with extracted file path, so that following hardlinks don't extract twice.
+                hardlink_masters[item.get(b'source') or original_path] = (None, path)
         elif stat.S_ISDIR(mode):
             if not os.path.exists(path):
                 os.makedirs(path)
@@ -527,7 +551,10 @@ Number of files: {0.stats.nfiles}'''.format(
             source = self.hard_links.get((st.st_ino, st.st_dev))
             if (st.st_ino, st.st_dev) in self.hard_links:
                 item = self.stat_attrs(st, path)
-                item.update({b'path': safe_path, b'source': source})
+                item.update({
+                    b'path': safe_path,
+                    b'source': source,
+                })
                 self.add_item(item)
                 status = 'h'  # regular file, hardlink (to already seen inodes)
                 return status
@@ -549,7 +576,10 @@ Number of files: {0.stats.nfiles}'''.format(
                 status = 'U'  # regular file, unchanged
         else:
             status = 'A'  # regular file, added
-        item = {b'path': safe_path}
+        item = {
+            b'path': safe_path,
+            b'hardlink_master': st.st_nlink > 1,  # item is a hard link and has the chunks
+        }
         # Only chunkify the file if needed
         if chunks is None:
             fh = Archive._open_rb(path)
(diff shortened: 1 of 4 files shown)


borg/archive.py
```diff
@@ -298,7 +298,36 @@ def add_file_chunks(chunks):
...
+            if os.path.exists(linked_to):
+                return True, path, None, linked_to
+
+    def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False, items=None,
+                     original_path=None):
+        """
+        Extract archive item.
+
+        :param item: the item to extract
+        :param restore_attrs: restor file attributes
```
ThomasWaldmann: typo restore

borg/archive.py
```diff
@@ -298,7 +298,36 @@ def add_file_chunks(chunks):
...
+        :param dry_run: do not write any data
+        :param stdout: write extracted data to stdout
+        :param sparse: recreate sparse files (only approximate, no exact hole reproduction)
```
ThomasWaldmann: in fact it is even a "sparsify" of all files that have all-zero chunks, no matter whether they were holes a archive creation time or not.

borg/archive.py
```diff
@@ -327,25 +356,36 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
-                    fd.flush()
-                    self.restore_attrs(path, item, fd=fd.fileno())
+                if not items:
+                    os.link(source, path)
+                    return
+                make_link, entry_path, item[b'chunks'], link_target = self.get_link_chunks(original_path, items)
+                if make_link:
+                    os.link(link_target, path)
+                    return
+                # we will extract the item "by chunks", since the items providing the chunks has not been extracted
```
ThomasWaldmann: typo item

borg/archive.py
```diff
@@ -327,25 +356,36 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
+                # hence we overwrite it's entry with out entry, so that future hardlinks will correctly link
```
ThomasWaldmann: typo its, typo our

borg/archive.py
```diff
@@ -327,25 +356,36 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
-                        if sparse and self.zeros.startswith(data):
-                            # all-zero chunk: create a hole in a sparse file
-                            fd.seek(len(data), 1)
-                        else:
-                            fd.write(data)
-                    pos = fd.tell()
-                    fd.truncate(pos)
-                    fd.flush()
-                    self.restore_attrs(path, item, fd=fd.fileno())
+                if not items:
```
ThomasWaldmann: so, this "if" would trigger if items is None and also if items == {} - is this wanted?

enkore: If items were empty and we're extracting a hard link from a subtree, then archive is corrupted, since the item holding the chunks for the hard link didn't come first.

ThomasWaldmann: ok, so the code should be "if items is None" as the other case is not possible?
maybe also check "if not items" afterwards and raise an error?

enkore: If this were somehow to happen (and I can't see how, assuming that the archive is as-created), then this will blow up and throw an exception anyway when trying to look the b'source' up in it.

borg/archive.py
```diff
@@ -298,35 +298,16 @@ def add_file_chunks(chunks):
         cache.rollback()
```
enkore: "This over avoids storing all files in a mapping and allows proper preloading for new archives."

Oops. Guess I over-pasted part of the commit message :D

borg/archiver.py
```diff
@@ -355,9 +355,10 @@ def do_extract(self, args):
         dirs = []
         partial_extract = not matcher.empty() or strip_components
         items = {} if partial_extract else None
-        for item in archive.iter_items(preload=True):
+        for item in archive.iter_items(preload=True,
+                filter=lambda item: item.get(b'link', True) or matcher.match(item[b'path'])):
             orig_path = item[b'path']
-            if partial_extract and stat.S_ISREG(item[b'mode']):
+            if partial_extract and stat.S_ISREG(item[b'mode']) and item.get(b'link', True):
```
ThomasWaldmann: so for old archives, it will assume all files could be hardlinks (either with data or referring to another hardlink).

borg/archive.py
```diff
@@ -589,7 +572,10 @@ def process_file(self, path, st, cache, ignore_inode=False):
                 status = 'U'  # regular file, unchanged
         else:
             status = 'A'  # regular file, added
-        item = {b'path': safe_path}
+        item = {
+            b'path': safe_path,
+            b'link': st.st_nlink > 1,
```
ThomasWaldmann: so the meaning of "link" is that this item is member of a set of hardlinks (the first one will have data, the other members will refer to the first one). maybe this should be in a comment in the src.

ThomasWaldmann: hm, no. meaning of "link" is rather "this item has the data for a set of hardlinks".

enkore: Not exactly, it means that this is item is a hard link (nlink > 1) and that it has the chunks. That way the "items" mapping only holds items which are actually needed to recreate a hard link.

borg/archiver.py
enkore: Yes, with all the drawbacks mentioned above (a lot higher memory usage, broken/useless preloading for remote repositories, since all items are marked for preloading as if you were extracting the entire archive).

borg/archive.py
```diff
@@ -574,7 +574,7 @@ def process_file(self, path, st, cache, ignore_inode=False):
             status = 'A'  # regular file, added
         item = {
             b'path': safe_path,
-            b'link': st.st_nlink > 1,
+            b'link': st.st_nlink > 1,  # item is a hard link and has the chunks
```
ThomasWaldmann: considering that every regular file is somehow a "hard link", the entry name could just be "multilinks".

borg/archive.py
```diff
@@ -549,7 +549,7 @@ def process_file(self, path, st, cache, ignore_inode=False):
                 item.update({
                     b'path': safe_path,
                     b'source': source,
-                    b'link': False,
+                    b'link': False,  # item is a hard link (and doesn't have the chunks)
```
ThomasWaldmann: hmm, do we need it here? we have already source.

borg/archive.py
enkore: How about b'linkmaster' or b'mainlink'? Makes it clear that this about a link and it's somehow important / different...

borg/archive.py
enkore: No we don't, items with b'source' can never be in `items` (in do_extract).

borg/archive.py
ThomasWaldmann: hardlink_master plus a note that we consider the first hardlink we encounter as master and save it with the data.

ThomasWaldmann:
btw, if a commit comment ends with ", fixes #XXX", github will automatically close the issue with a link to the changeset that closed it.

enkore:
I added another sub-testcase, which wasn't handled well previously (when the hardlink_master was extracted as well and this special-case-code was effective, then it would be extracted twice, even though a hard link to the master was possible).

I think this should cover it. The xcode tests should pass now, their failure were related to the absolute paths of some symlinks, I just removed that code, had nothing to do with this anyway.

The open question is metadata handling, I'll look into that tomorrow.

borg/archiver.py
```diff
@@ -353,8 +353,20 @@ def do_extract(self, args):
         sparse = args.sparse
         strip_components = args.strip_components
         dirs = []
-        for item in archive.iter_items(lambda item: matcher.match(item[b'path']), preload=True):
+        partial_extract = not matcher.empty() or strip_components
+        hardlink_masters = {} if partial_extract else None
+
+        def item_is_hardlink_master(item):
+            return (partial_extract and stat.S_ISREG(item[b'mode']) and
```
ThomasWaldmann: nice name, but not totally true. if partial_extract is False, it returns False even if it is a hardlink_master.

ThomasWaldmann: so, just drop function and inline it?

enkore: It has two call sites, didn't want to duplicate the code. (one in the filter lambda, the other in the if below to only add possible masters to hardlink_masters).

ThomasWaldmann: ah, ok.

dyu:
I still get an error if a hardlink is extracted ```[Errno 2] No such file or directory```

Here's a test script to reproduce
```sh
#!/bin/sh

mkdir test && cd test

#initial data
mkdir data
echo "a" > data/a.txt

#initial backup
mkdir backups && cd backups
BORG_PASSPHRASE=bar! borg init bar
BORG_PASSPHRASE=bar! borg create bar::b1 ../data

# update
echo "bb" > ../data/b.txt
BORG_PASSPHRASE=bar! borg create bar::b2 ../data

#extract
mkdir extracted && cd extracted
BORG_PASSPHRASE=bar! borg extract -v --list ../bar::b2
cd ..

cd ../data
# hardlink
ln b.txt c.txt

cd - > /dev/null
BORG_PASSPHRASE=bar! borg create bar::b3 ../data

mkdir extracted-hardlink && cd extracted-hardlink
# this fails (works fine if data/b.txt is provided as arg)
BORG_PASSPHRASE=bar! borg extract -v --list ../bar::b3 data/c.txt


```

enkore:
Version? This fix has not been released yet.

Issue #757, opened 2016-03-17
Improved output format for 'borg diff'


Ape:
I'd like to thank @enkore and @ThomasWaldmann for implementing ```borg diff``` command. It is really useful for me and it works surprisingly fast. See #177 and #740.

I'd like to have a slightly modified output format for the command. I will present one draft that I think would be very good, but please tell your opinions about the format. We might want to support multiple formats if so desired.

In the current version, each change will take two lines of output. The most important change I'd like to propose is to have only one line per change. In addition, the file paths may be long so anything on the line after the path may be hard to see; especially since the path length typically varies a lot between lines. The format should be also terse and quick to read.

It could look like this:
```
+ path/to/added
- path/to/deleted
[content: +202 B, -215 B] path/to/contents_changed
[user: root -> ape] path/to/user_changed
[mode: -rw-rw-rw- -> -rw-r----] path/to/permissions_changed
[mode: drwxr-xr-x -> lrwxrwxrwx] path/to/folder_to_symlink
[content: +42 kB, -532 kB] [user: ape -> root] [group: users -> root] [mode: -rw-r---- -> -rw-rw-rw] path/to/everything_changed (should be quite rare case)
...
```

enkore:
#751 gives us a flexible `ItemFormatter` that is extensible as well, that can be used for this (e.g. two formatters which are output one after the other with the "old" and "new" item --- whether that results in two output lines is up to the format used).

Ape:
Maybe we could also use terminal colors and bold font. For example deleted files could be red lines, and new files could be green. I'd like to make examining long change lists as easy as possible.

enkore:
Hmm, terminal color handling is quite complicated. If this should be implemented (i have no particular opinion), then it would have to use a soft-dependency, implementing it ourselves is way too much effort. A soft-dependency might be justifiable in this case, since it is no loss of essential functionality if on some systems no colors are printed.

---

Your proposed format looks good to me.

ThomasWaldmann:
no terminal colours for now.

i still have to play with the code and practically try it, but proposal sounds good.

Ape:
I was thinking about this a bit more and realized that the paths should be aligned as possible. Rare cases, like lots of changed metadata, can be an exception to this. So here is something along these ideas:

```
added       path/to/added
removed     path/to/deleted
+842B -215B path/to/contents_changed
+12M  -111K path/to/contents_changed2
[u:root -> u:ape] path/to/user_changed
[...]

PR #767, opened 2016-03-18
Update usage docs (diff and list)

enkore:
- Note in borg diff help explaining --same-chunker-params
- Corrected "PATH" description for borg list saying that it's about extracting
- Ran "setup.py build_usage"
- Added an example for borg diff (we should remember to update that when/if we change the output format)

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1212,7 +1212,8 @@ class Archiver:
             Both archives need to be in the same repository, and a repository location may only
             be specified for ARCHIVE1.
 
-            For archives created with Borg 1.1 or newer only chunks IDs
+            For archives created with Borg 1.1 or newer diff automatically detects whether
+            the archives are created with the same chunker params. If so, only chunk IDs
             are compared, which is very fast.
 
             For archives prior to Borg 1.1 chunk contents are compared by default.
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1212,6 +1212,11 @@ class Archiver:
             Both archives need to be in the same repository, and a repository location may only
             be specified for ARCHIVE1.
 
+            For archives created with Borg 1.1 or newer only chunks IDs
+            are compared, which is very fast. For archives prior to Borg 1.1 chunk contents must be
+            compared, unless you are sure they were created with the same chunker params
+            (pass --same-chunker-params). Note that the chunker params changed from Borg 0.xx to 1.0.
+
             See the output of the "borg help patterns" command for more help on exclude patterns.
             """)
         subparser = subparsers.add_parser('diff', parents=[common_parser],
@@ -1289,7 +1294,7 @@ class Archiver:
 
         See the "borg help patterns" command for more help on exclude patterns.
 
-        The following keys are available for --format:
+        The following keys are available for --format when listing files:
 
         """) + ItemFormatter.keys_help()
         subparser = subparsers.add_parser('list', parents=[common_parser],
@@ -1316,7 +1321,7 @@ class Archiver:
                                type=location_validator(),
                                help='repository/archive to list contents of')
         subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
-                               help='paths to extract; patterns are supported')
+                               help='paths to list; patterns are supported')
 
         mount_epilog = textwrap.dedent("""
         This command mounts an archive as a FUSE filesystem. This can be useful for
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1213,9 +1213,12 @@ class Archiver:
             be specified for ARCHIVE1.
 
             For archives created with Borg 1.1 or newer only chunks IDs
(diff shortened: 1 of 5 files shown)


borg/archiver.py
```diff
@@ -1212,6 +1212,11 @@ def build_parser(self, args=None, prog=None):
             Both archives need to be in the same repository, and a repository location may only
             be specified for ARCHIVE1.
 
+            For archives created with Borg 1.1 or newer only chunks IDs
+            are compared, which is very fast. For archives prior to Borg 1.1 chunk contents must be
```
ThomasWaldmann: ... (if you did not create the archives with different chunker params) ...

docs/usage.rst
```diff
@@ -374,6 +374,48 @@ Examples
...
 
+
+.. include:: usage/diff.rst.inc
+
+Examples
+~~~~~~~~
+::
+
+    $ mkdir testdir
+    $ echo asdf > testdir/file1
```
ThomasWaldmann: if you cd into the testdir, you could avoid repeating it all the time (and give "." to borg create).

docs/usage.rst
```diff
@@ -374,6 +374,48 @@ Examples
...
+    $ borg diff testrepo::archive1 archive2
+    testdir/file1 different mode
+             archive1 -rw-r--r--
+             archive2 -rwxr-xr-x
+    testdir/file2 different contents
+             +28 B, -31 B, 4.19 MB, 4.19 MB
+
+    $ borg diff testrepo::archive2 archive3
+    testdir/file3 different contents
+             +0 B, -0 B, 0 B, <deleted>
```
ThomasWaldmann: hmm, that display is a bit too complicated, it could just say the file was deleted.

docs/usage.rst
```diff
@@ -374,6 +374,48 @@ Examples
...
+Examples
+~~~~~~~~
+::
+
+    $ mkdir testdir
+    $ echo asdf > testdir/file1
+    $ dd if=/dev/urandom bs=1M count=4 > testdir/file2
+    $ touch testdir/file3
+
+    $ borg init testrepo
```
ThomasWaldmann: maybe move the init to the very top of the example, move the create archive1 directly under the touch file3.

docs/usage.rst
ThomasWaldmann: also it seems to compute the difference / abs. size incorrectly.

ThomasWaldmann: we have an example exists -> deleted with that, an example notexists -> created is missing.

docs/usage/list.rst.inc
```diff
@@ -30,15 +31,59 @@ borg list
...
+ - uid
+ - unique_chunks: number of unique chunks in this file
+ - user
+ - CR
+ - LF
+ - NEWLINE: OS dependent line separator
+ - NL: alias of NEWLINE
+ - NUL: NUL character for creating print0 / xargs -0 like ouput, see bpath
+ - SPACE
+ - TAB
```
ThomasWaldmann: linefeed missing.

docs/usage/list.rst.inc
```diff
@@ -30,15 +31,59 @@ borg list
...
     
 Description
 ~~~~~~~~~~~
 
 This command lists the contents of a repository or an archive.
+
+See the "borg help patterns" command for more help on exclude patterns.
+
+The following keys are available for --format when listing files:
+
```
ThomasWaldmann: maybe grouping these by hashes, timestamps, single characters, etc. would make it easier?

docs/usage.rst
enkore: > also it seems to compute the difference / abs. size incorrectly.

dd uses MiB for M. 4 MiB => 4.2 MB

The difference size depends on where the chunker cuts and how exactly something was changed, and how chunks are repeated etc.
(Hm, maybe it should be removed, it could be quite confusing)

ThomasWaldmann: I meant line 408.

enkore: file3 was empty (only touched). +0/-0/0/del is right, but verbose. Better output formatting per se is the scope of #757

ThomasWaldmann: ah, right.

borg/archiver.py
```diff
@@ -1213,9 +1213,12 @@ def build_parser(self, args=None, prog=None):
             be specified for ARCHIVE1.
 
             For archives created with Borg 1.1 or newer only chunks IDs
-            are compared, which is very fast. For archives prior to Borg 1.1 chunk contents must be
-            compared, unless you are sure they were created with the same chunker params
-            (pass --same-chunker-params). Note that the chunker params changed from Borg 0.xx to 1.0.
+            are compared, which is very fast.
```
ThomasWaldmann: you still should say that it is only doing that as long as the chunker params are the same.

ThomasWaldmann:
other than that: shall I merge it or do you want to change anything else?

enkore:
No, I think that's all for this PR

PR #770, opened 2016-03-18
WIP / Feature recompress (2)

enkore:
This is teh announced successor to #756 

I've crunched now a few hundred GB through a few versions of this (including the multi-threaded prototype) and so far no data loss (diff of SHA-512 `borg list`s) and no complaints from `borg check`.

Improvements over older prototypes now include "really clean exit" (i.e. no stale locks, everything nice and committed, it even prints stats if asked to), proper progress indication and reverse processing (see 5c2ea3a for rationale). And now we *really* don't write anything on --dry-runs.

Additional space usage bound: Depends on what the data is stored with and what you're compressing with. You can use recompress to decompress stuff (`-C none`). Generally speaking, it's about 10 * max_segment_size[1].

[1] I'm going with 10 MSS currently, because I'm assuming that commit's are costly (and I think they are, since they do all the clean up of over-written (in the key-value store sense) segments).

Also, additional space use will obviously be much larger if you aren't compressing.

**TODO**

- [ ] Code clean up
- [ ] Tests (some more manual and units)
   - What I really want to do in terms of manual testing is write a small script that calls recompress-list-diff in a loop like before, *but also kills it at random*. (So that we get good real-life coverage of this scenario, something which I'd say wouldn't be practical in unit tests)
   - What about TAG_DELETE?
     - I don't think it matters here. If it does, we could simply add a `id_ not in repository` => `continue` check, to only recompress objects in the index.
- [ ] Docs (Examples, better epilogue-explanation)
- [ ] Possibly change stats output
- [ ] doesn't work for RemoteRepository, because it works *below* they key/value store abstraction. ATM I don't consider this a todo, because it would be quite hard to make it work for RRs.
   - OTOH it could be implemented purely on top of the Repo interface. That would make space usage unpredictable and can/will touch a lot of segments after each other, though. So the performance hit (both time and space) could be quite drastic.

**Feature list**

- [x] recompress stuff
- [x] skip stuff
- [x] option to not skip stuff
- [x] option to skip stuff really fast

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -725,12 +725,29 @@ class Archiver:
             old_size = new_size = 0
             chunks_seen = chunks_recompressed = chunks_skipped = 0
 
+            def seen(self, chunk):
+                self.chunks_seen += 1
+                self.old_size += len(chunk)
+
+            def skipped(self, chunk):
+                self.chunks_skipped += 1
+                self.new_size += len(chunk)
+
+            def recompressed(self, chunk):
+                self.chunks_recompressed += 1
+                self.new_size += len(chunk)
+
         def detect_compressor(compression_header):
             for compressor in COMPRESSOR_LIST:
                 if compressor.detect(compression_header):
                     return compressor
             else:
                 raise ValueError('No decompressor for this data found: %r.', compression_header)
+
+        def recompress_chunk(decompressor, compressed_chunk):
+            decompressed_data = decompressor(buffer=decompress_buffer).decompress(compressed_chunk)
+            key.assert_chunk_id(id_, decompressed_data)
+            return key.encrypt(decompressed_data)
 
         def do_exit_soon(sig_num, stack_frame):
             nonlocal exit_soon
@@ -768,36 +785,24 @@ class Archiver:
         for i, (segment_id, segment_filename) in enumerate(reversed(segments)):
             chunks = {}
             for tag, id_, offset, data in repository.io.iter_objects(segment_id, True):
-                if tag != TAG_PUT or set(id_) == {0}:
+                if tag != TAG_PUT or id_ == Manifest.MANIFEST_ID:
                     continue
-                stats.chunks_seen += 1
-                stats.old_size += len(data)
+                stats.seen(data)
                 decrypted_data = key.decrypt(id_, data, no_decompress=True)
                 compressor = detect_compressor(decrypted_data[:2])
                 if not args.force_recompress and isinstance(key.compressor.compressor, compressor):
-                    stats.chunks_skipped += 1
-                    stats.new_size += len(data)
+                    stats.skipped(data)
                     continue
-                decompressed_data = compressor(buffer=decompress_buffer).decompress(decrypted_data)
-                key.assert_chunk_id(id_, decompressed_data)
-                data = key.encrypt(decompressed_data)
-                stats.new_size += len(data)
-                chunks[id_] = data
+                chunks[id_] = recompress_chunk(compressor, decrypted_data)
             if not dry_run:
-                try:
-                    for id_, data in chunks.items():
-                        repository.put(id_, data)
-                        stats.chunks_recompressed += 1
-                    if exit_soon or repository.io.get_latest_segment() > last_segment_id + 10:
-                        last_segment_id = repository.io.get_latest_segment()
-                        repository.commit()
-                    if exit_soon:
-                        break
-                except Exception as e:  # too broad!
-                    self.print_error("Exception while recompressing, rolling transaction back...")
-                    repository.rollback()
-                    repository.close()
-                    raise
+                for id_, data in chunks.items():
+                    repository.put(id_, data)
+                    stats.recompressed(data)
+                if repository.io.get_latest_segment() > last_segment_id + 10:
+                    last_segment_id = repository.io.get_latest_segment()
+                    repository.commit()
+                if exit_soon:
+                    break
             if args.progress:
                 progress_indicator.show(i)
         if args.progress:
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -725,6 +725,20 @@ class Archiver:
             old_size = new_size = 0
             chunks_seen = chunks_recompressed = chunks_skipped = 0
 
+            def __str__(self):
+                if self.old_size:
+                    ratio = self.new_size / self.old_size * 100
+                else:
+                    ratio = 0
+                return textwrap.dedent("""
+                Chunks seen, recompressed, skipped: {0.chunks_seen}, {0.chunks_recompressed}, {0.chunks_skipped}
+                Old size: {old_size}
+                New size: {new_size}
+                Ratio: {ratio:.0f} %""").format(self,
+                                              old_size=format_file_size(self.old_size),
+                                              new_size=format_file_size(self.new_size),
+                                              ratio=ratio).strip()
+
             def seen(self, chunk):
                 self.chunks_seen += 1
                 self.old_size += len(chunk)
@@ -737,24 +751,6 @@ class Archiver:
                 self.chunks_recompressed += 1
                 self.new_size += len(chunk)
 
-            def print(self):
-                print("Chunks seen, recompressed, skipped: %d, %d, %d" %
-                      (self.chunks_seen, self.chunks_recompressed, self.chunks_skipped))
-                print("Old size:", format_file_size(self.old_size))
-                print("New size:", format_file_size(self.new_size))
-                if self.old_size:
-                    ratio = self.new_size / self.old_size
-                else:
-                    ratio = 0
-                print("Ratio: %d %%" % (ratio * 100))
-
-        def detect_compressor(compression_header):
-            for compressor in COMPRESSOR_LIST:
-                if compressor.detect(compression_header):
-                    return compressor
-            else:
-                raise ValueError('No decompressor for this data found: %r.', compression_header)
-
         def recompress_chunk(decompressor, compressed_chunk):
             decompressed_data = decompressor(buffer=decompress_buffer).decompress(compressed_chunk)
             key.assert_chunk_id(id_, decompressed_data)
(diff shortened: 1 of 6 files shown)


borg/archiver.py
```diff
@@ -719,6 +719,107 @@ def do_upgrade(self, args):
...
+                try:
+                    for id_, data in chunks.items():
+                        repository.put(id_, data)
+                        stats.chunks_recompressed += 1
+                    if exit_soon or repository.io.get_latest_segment() > last_segment_id + 10:
+                        last_segment_id = repository.io.get_latest_segment()
+                        repository.commit()
+                    if exit_soon:
+                        break
+                except Exception as e:  # too broad!
```
enkore: I added this part when evaluating "Ctrl-C safety" the first time (and it caught they KeyboardInterrupt reliably), but I think now that signals are handled explicitly it's unnecessary.

borg/key.py
```diff
@@ -75,7 +75,11 @@ def id_hash(self, data):
     def encrypt(self, data):
         pass
 
-    def decrypt(self, id, data):
+    def decrypt(self, id, data, no_decompress=False):
```
ThomasWaldmann: how about compress=True? ;)

borg/key.py
```diff
@@ -144,14 +154,19 @@ def decrypt(self, id, data):
         if not compare_digest(hmac_computed, hmac_given):
             raise IntegrityError('Encryption envelope checksum mismatch')
         self.dec_cipher.reset(iv=PREFIX + data[33:41])
+        if no_decompress:
+            return self.dec_cipher.decrypt(data[41:])
         data = self.compressor.decompress(self.dec_cipher.decrypt(data[41:]))
```
ThomasWaldmann: same here.

borg/archiver.py
```diff
@@ -719,6 +719,134 @@ def do_upgrade(self, args):
...
+            key.assert_chunk_id(id_, decompressed_data)
+            return key.encrypt(decompressed_data)
+
+        def do_exit_soon(sig_num, stack_frame):
+            nonlocal exit_soon
+            if exit_soon:
+                sys.stderr.write("Received signal, again. I'm not deaf.\n")
+            else:
+                sys.stderr.write("Received signal, will exit cleanly.\n")
+            sys.stderr.flush()
```
ThomasWaldmann: print() has ofile and flush support in py 3.4

borg/key.py
```diff
@@ -99,14 +103,20 @@ def id_hash(self, data):
     def encrypt(self, data):
         return b''.join([self.TYPE_STR, self.compressor.compress(data)])
 
-    def decrypt(self, id, data):
+    def decrypt(self, id, data, no_decompress=False):
         if data[0] != self.TYPE:
             raise IntegrityError('Invalid encryption envelope')
+        if no_decompress:
+            return memoryview(data)[1:]
         data = self.compressor.decompress(memoryview(data)[1:])
```
ThomasWaldmann: put cdata = memoryview(data)[1:] above these 3 lines to remove the duplication.

borg/archiver.py
```diff
@@ -719,6 +719,134 @@ def do_upgrade(self, args):
...
+
+        if not repository.io.is_committed_segment(segments[-1][1]):
+            self.print_error('Last segment in repository is uncomitted. Repository corrupted. Try '
+                             '"borg check --repair".')
+            repository.close()
+            return self.exit_code
+        for i, (segment_id, segment_filename) in enumerate(segments):
+            chunks = {}
+            if segment_id <= segment_pointer:
+                # commited segment files are never appended to, yes?
```
ThomasWaldmann: committed

borg/archiver.py
```diff
@@ -719,6 +719,134 @@ def do_upgrade(self, args):
...
+                else:
+                    ratio = 0
+                print("Ratio: %d %%" % (ratio * 100))
+
+        def detect_compressor(compression_header):
+            for compressor in COMPRESSOR_LIST:
+                if compressor.detect(compression_header):
+                    return compressor
+            else:
+                raise ValueError('No decompressor for this data found: %r.', compression_header)
```
ThomasWaldmann: looks like code that is already used elsewhere, refactor?

borg/archiver.py
```diff
@@ -719,6 +719,134 @@ def do_upgrade(self, args):
...
+
+            def skipped(self, chunk):
+                self.chunks_skipped += 1
+                self.new_size += len(chunk)
+
+            def recompressed(self, chunk):
+                self.chunks_recompressed += 1
+                self.new_size += len(chunk)
+
+            def print(self):
```
ThomasWaldmann: bit unpretty to shadow a builtin

borg/archiver.py
enkore: Thought about it and looked at at the code.

So we have the transaction of the Repository and the one of LoggedIO (the "segment transaction"). So the initial +1 which guarantees that any new transaction will start off in a new segment comes from LoggedIO.cleanup(old_transaction_id), which assigns self.segment = transaction_id + 1.

On Repository.commit() LoggedIO.write_commit() is called - which always closes the current segment. After that compact_segments might squish stuff into less segments than existed before, but because the old segment was closed, a new one is created for these "compacting writes".

borg/key.py
enkore: that's certainly not less_legible :-)

enkore:
So, now we got a "major" problem. csize in chunk id lists. It's a nice book-keeping number, sure. But we never update csize with this approach, because we never touch these lists, nor archives nor the manifest.

We could do that. But now our problems just increased manifold: items -- that is, the item-dict with all the metadata -- are chunked as well, and the list of chunk ids that represents the chunked items is the 'items' key of archive metadata. And this metadata block is just another chunk, with the chunk ID -- the archive's ID -- depending, as usual, on the chunk contents.

Change the csize, changes the chunk IDs of the items and that changes all archive IDs. However, archive IDs must be final and never change (by our definition).

Independent of how we approach recompression, either we

* ignore incorrect csize (just a book keeping number)
* drop / ignore csize entirely
* drop recompression entirely

Dropping csize means that it's harder to find out compressed size of items on disk: while we don't need to actually read it, we'd need to take a peek at the segment holding the chunk, since the segment entry contains the length.

However, `borg create`-time book keeping of how well stuff compressed wouldn't be affected, we could still do that.

ThomasWaldmann:
remote repos not working and the csize issues let this look rather questionable.

enkore:
I agree! Remote repos can be made work, but the csize turns up again as a PITA by principle. So I'm thinking we should solve this problem on a root-cause-level (and the root cause is, that we put csize, which is *unrelated* to the identity of chunks[1], into the chunk lists, which are final/sealed).

What are the clients (in an API sense) of csize?

- borg create: uses it, but can easily be changed *not* to use it.
- borg info: for the compressed stats
- borg list: for {csize}

info requires an up to date Cache. list does as well (it could be lazy about it, though, which might be worth another PR).

So here's my line of thought: new archives get csize=0 (or -1 or some other dummy value) in all chunks. Repository gets a `size_of(id_)` / `size_of_many(ids)` API. When we're writing chunks the Cache gets the real csize.

- [x] no identity unrelated data in chunk lists
- [x] removes the delayer-problem for multithreading (because now the pipeline has no cycles, we can work "forwards-only", not "backwards-passing" of information)
- [x] can recompress etc. easily
- [x] borg info, list etc. continue to work like they did with no additional hit (they use the Cache already)
- But: Cache creation from scratch is more costly

If we decide to solve csize this way, it shouldn't be done in this PR imho, though.

ThomasWaldmann:
yes, csize is a troublemaker. but i guess we need to delay any change of repo api to until later, I'ld like to make a few releases with features and fixes that we can do without breaking repo / repo api compatibility.

enkore:
That's reasonable. Should we postpone / close this one then, or put a disclaimer on it for now basically saying "if you use borg recompress, the compressed numbers of create/info are off. Thy must know yourself whether that matters."

Because I see recompress as something really useful until there is (hopefully) someday a faster, multi-threaded borg create, to (citing commit message)

> In my mind it is a goal in backup systems like Borg to minimize the
"acquisition time", the time needed to store a snapshot of the data.
Using heavy compression will usually increase this time.
> 
> Entrance borg recompress: Use some fast compression while running
borg create (LZ4 is pretty much ideal for any CPU+disk combination),
to minimize acquisition time. Run borg recompress later at any time
to squash space usage down.
> 
> It will compress the chunks of latest archives first.

enkore:
> we can do without breaking repo / repo api compatibility

We might need to revisit #763 then, since the added key would break compatibility with RobustUnpacker of earlier versions (I think it's used by the borg check?).

ThomasWaldmann:
how about a more high-level and more generic approach for now, even if it is more expensive?

as this is likely a rare / one-time / never operation for most users, efficiency is not as important as for the regularly invoked stuff.

e.g. like:
```
for archive in repo: # could also be partial: archive with a specific name, starting with a prefix, ...
    create temp_archive
    for item in archive:
        new_chunks = transform_chunks(item.chunks)
        temp_archive.write_item(item, new_chunks)
    kill archive
    rename temp_archive archive
```
transform_chunks would recompress the chunks as needed. as the chunk ids do not change, space needs could be similar to your approach, but it would work also for remote repos and would not require any api or storage change (AFAICS).

this "recreate archive" approach could even do other stuff: like dropping files according to a new exclude list, applying dynamic compression (depending on filetype) - so it is more like create with data coming from an archive rather than from local disk.

about dropping csize: this would be something useful for multithreading branch and for a major release (2.0, 3.0?). we should keep the thoughts / research results in a ticket about that.

enkore:
That sounds like an idea, would also fulfill feature requests for deleting selectively in old archives (via --exclude, and, I don't know if someone asked for that, but someone probably did).

If we allow processing of single archives the csize will be wrong again for all other archives, though (-- in the general case, see below). And we can't really tell if that's the case: refcount > 1 of a chunk can also mean it's a reference from the same archive. Assuring that it is, is prohibitive in terms of either time or additional memory use.

If this were used in a "progressive recompression" scheme, i.e. after every or every group of `borg create` comes a `borg rewrite` (which is how I'd name it) to squish down space usage, like I described above, then this would indeed give us correct csize; old chunks would already be compressed and wouldn't be recompressed.

Issue #562, opened 2016-01-16
better locking


ThomasWaldmann:
misc ideas to improve locking.

enkore:
Regarding *locking with context managers to avoid stale locks on forgotten .close calls etc.*

An implementation idea would be to stack a decorator on top of a context manager: Practically all do_XYZ() methods of Archiver open a repo, use it all the time, and close (or don't) it at all of the exit points. If we use a context manager, we'll have an additional indent for all these methods over their entire spawn: not pretty. If we wrap the CM in a decorator we avoid that, and can deduplicate some code at the same time.

```
class Archiver:
    def do_blah(self, args):
        repository = self.open_repository(args, exclusive=True)
        manifest, key = Manifest.load(repository)
        # do stuff

    # With a context manager:

    def do_blah(self, args):
        with self.open_repository(args, exclusive=True) as repository:
            manifest, key = Manifest.load(repository)
            # do stuff

    # Now let's get a decorator in the spirit of open_repository (pseudo code)

    def open_repository(function):
        @wraps
        def wrapper(args, **kwargs):
            with (do stuff from open_repository) as repository:
                function(args, repository=repository, **kwargs)
        return wrapper

    @open_repository
    @load_manifest
    def do_blah(self, args, repository, manifest, key):
        # do stuff
```

This kind of decorator stacks (as I tried to express with `@load_manifest`), so even more code dedup.

enkore:
#1049 is another bit for this.

I think the only thing missing now is a repo config item "i_the_user_promise_that_i_wont_access_this_repo_over_sshfs_nfs_or_whatever" that allows removing locks via a PID check on the roster. Off by default.

ThomasWaldmann:
and then another admin of same server moves the repo on nfs...

dr4Ke:
If I understand right, the lock is specified as a couple (hostname, pid). Can't we just check the pid only if this is the local computer. It would solve at least two use cases: local usage, and remote "borg serve" usage. The other ones (access through a remote FS) would leave a lock as (remote_host:remote_pid) and we can also check this pid if the remote_host is the same as the one where is the current user. So the only case where we can't check the pid is if the user is running borg on a remote FS from another machine. Is this correct?

Possible check:
- local use on local FS
- remote use via SSH (`borg serve`)
- local use via remote FS (same hostname as the existing locks)

Impossible check:
- foreign use via remote FS (different hostname than the existing locks)
- local use on local FS with remote lock
[...]

PR #777, opened 2016-03-19
Feature ctxmng: Repository context manager

enkore:
From https://github.com/borgbackup/borg/issues/562#issuecomment-198576577

Not sure if we should do `@drop_unused_args`, avoids some repetition, but could be a bit "opaque".

Looks good to me. I don't see any ignored (raised in __del__) exceptions in ArchiverTestCase anymore, and only one in RemoteArchiver (test_remote_repo_restrict_to_path, I think it might have something to do with that monkey patching). This should essentially eliminate stale locks in practice. (Except for SIGTERM).

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -40,18 +40,56 @@ UMASK_DEFAULT = 0o077
 DASHES = '-' * 78
 
 
-class ToggleAction(argparse.Action):
-    """argparse action to handle "toggle" flags easily
+def argument(args, str_or_bool):
+    """If bool is passed, return it. If str is passed, retrieve named attribute from args."""
+    if isinstance(str_or_bool, str):
+        return getattr(args, str_or_bool)
+    return str_or_bool
 
-    toggle flags are in the form of ``--foo``, ``--no-foo``.
 
-    the ``--no-foo`` argument still needs to be passed to the
-    ``add_argument()`` call, but it simplifies the ``--no``
-    detection.
+def with_repository(fake=False, create=False, lock=True, exclusive=False, manifest=True, cache=False):
     """
-    def __call__(self, parser, ns, values, option):
-        """set the given flag to true unless ``--no`` is passed"""
-        setattr(ns, self.dest, not option.startswith('--no-'))
+    Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, …)
+
+    If a parameter (where allowed) is a str the attribute named of args is used instead.
+    :param fake: (str or bool) use None instead of repository, don't do anything else
+    :param create: create repository
+    :param lock: lock repository
+    :param exclusive: (str or bool) lock repository exclusively (for writing)
+    :param manifest: load manifest and key, pass them as keyword arguments
+    :param cache: open cache, pass it as keyword argument (implies manifest)
+    """
+    def decorator(method):
+        @functools.wraps(method)
+        def wrapper(self, args, **kwargs):
+            location = args.location  # note: 'location' must be always present in args
+            if argument(args, fake):
+                return method(self, args, repository=None, **kwargs)
+            elif location.proto == 'ssh':
+                repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
+            else:
+                repository = Repository(location.path, create=create, exclusive=argument(args, exclusive),
+                                        lock_wait=self.lock_wait, lock=lock)
+            with repository:
+                if manifest or cache:
+                    kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
+                if cache:
+                    with Cache(repository, kwargs['key'], kwargs['manifest'],
+                               do_files=getattr(args, 'cache_files', False), lock_wait=self.lock_wait) as cache_:
+                        return method(self, args, repository=repository, cache=cache_, **kwargs)
+                else:
+                    return method(self, args, repository=repository, **kwargs)
+        return wrapper
+    return decorator
+
+
+def with_archive(method):
+    @functools.wraps(method)
+    def wrapper(self, args, repository, key, manifest, **kwargs):
+        archive = Archive(repository, key, manifest, args.location.archive,
+                          numeric_owner=getattr(args, 'numeric_owner', False), cache=kwargs.get('cache'))
+        return method(self, args, repository=repository, manifest=manifest, key=key, archive=archive, **kwargs)
+    return wrapper
 
 
 class Archiver:
@@ -59,14 +97,6 @@ class Archiver:
     def __init__(self, lock_wait=None):
         self.exit_code = EXIT_SUCCESS
         self.lock_wait = lock_wait
-
-    def open_repository(self, args, create=False, exclusive=False, lock=True):
-        location = args.location  # note: 'location' must be always present in args
-        if location.proto == 'ssh':
-            repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
-        else:
-            repository = Repository(location.path, create=create, exclusive=exclusive, lock_wait=self.lock_wait, lock=lock)
-        return repository
 
     def print_error(self, msg, *args):
         msg = args and msg % args or msg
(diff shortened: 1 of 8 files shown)


borg/archiver.py
```diff
@@ -54,20 +55,69 @@ def __call__(self, parser, ns, values, option):
...
+
+
+def with_cache(method):
+    @functools.wraps(method)
+    def wrapper(self, args, repository, key, manifest, **kwargs):
+        with Cache(repository, key, manifest, do_files=getattr(args, 'cache_files', False),
+                   lock_wait=self.lock_wait) as cache:
+            return method(self, args, repository=repository, key=key, manifest=manifest, cache=cache, **kwargs)
+    return wrapper
+
```
ThomasWaldmann: quite a lot of magic and additional code, just to save 1 indentation level.

borg/remote.py
```diff
@@ -347,7 +360,7 @@ def load_key(self):
     def break_lock(self):
         return self.call('break_lock')
 
-    def close(self):
+    def _close(self):
```
ThomasWaldmann: why did you rename? to point out that one should use "with"?

borg/archiver.py
enkore: I don't see much magic in the with_* decorators, but I agree that `@drop_unused_args` might be too opaque to actually use it outside a Python Obfuscated Code Contest (well, not that bad, but you get the idea).

borg/remote.py
enkore: Yep, _close would now be a private method of Repository, which generally shouldn't be used from client code.

borg/archiver.py
enkore: I dropped @drop_unused_args, @with_cache was also removed.

Rationale: with_cache and with_repository both operated on the same logical level (repository wide), while with_archive operates on another, smaller level (archive level).

enkore:
I reverted the renaming of _close to avoid breaking the RPC protocol. I tested it manually with borg 1.0 using this as the server, and vice versa, for list|extract|create. Is there a way to do automatic RPC compatibility tests between versions that I overlooked? If there isn't, we should devise one, because manually testing this takes a good while and is of course unreliable like all manual testing (and doesn't use the large test suite we have).

I also extended one test case for borg delete slightly, so that Codecov doesn't complain ;)

borg/archiver.py
```diff
@@ -1,5 +1,6 @@
 from binascii import hexlify, unhexlify
 from datetime import datetime
+from inspect import signature
```
ThomasWaldmann: not used any more? (good :)

borg/archiver.py
```diff
@@ -40,18 +41,56 @@
...
+    if isinstance(str_or_bool, str):
+        return getattr(args, str_or_bool)
+    return str_or_bool
 
-    toggle flags are in the form of ``--foo``, ``--no-foo``.
 
-    the ``--no-foo`` argument still needs to be passed to the
-    ``add_argument()`` call, but it simplifies the ``--no``
-    detection.
+def with_repository(fake=False, create=False, lock=True, exclusive=False, manifest=True, cache=False):
```
ThomasWaldmann: is there a specific reason why some items default to True, other to False?

might be easier to say "if you want <item>, you need item=True" (and default to False).

enkore: The defaults are what is used most commonly:

- 8 commands use the defaults (cache=False, manifest=True, exclusive=False)
- 5 have manifest=False (init, check and all do_debug stuff)
- 4 use cache=True
- 4 use exclusive=True

borg/archiver.py
enkore: Yep :)

borg/remote.py
```diff
@@ -164,11 +167,21 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
         self.id = self.call('open', location.path, create, lock_wait, lock)
 
     def __del__(self):
-        self.close()
+        if self.p:
+            self.close()
+            assert False, "cleanup happened in Repository.__del__"
```
ThomasWaldmann: this is just for the case, right - or are you still seeing this?

borg/repository.py
```diff
@@ -59,16 +59,31 @@ def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=Tru
...
+        self.lock_wait = lock_wait
+        self.do_lock = lock
+        self.do_create = create
+        self.exclusive = exclusive
 
     def __del__(self):
-        self.close()
+        if self.lock:
+            self.close()
+            assert False, "cleanup happened in Repository.__del__"
```
ThomasWaldmann: same q. here?

borg/testsuite/archiver.py
```diff
@@ -1122,8 +1129,9 @@ def setUp(self):
 
     def open_archive(self, name):
         repository = Repository(self.repository_path)
-        manifest, key = Manifest.load(repository)
-        archive = Archive(repository, key, manifest, name)
+        with repository:
+            manifest, key = Manifest.load(repository)
+            archive = Archive(repository, key, manifest, name)
         return archive, repository
```
ThomasWaldmann: hmm, if you use "with repository" above, it will return a closed repository here.

borg/testsuite/key.py
```diff
@@ -4,11 +4,14 @@
 import tempfile
 from binascii import hexlify, unhexlify
 
+from borg.logger import setup_logging
 from ..crypto import bytes_to_long, num_aes_blocks
 from ..key import PlaintextKey, PassphraseKey, KeyfileKey
 from ..helpers import Location
 from . import BaseTestCase
 
+setup_logging()
```
ThomasWaldmann: hmm?

borg/remote.py
enkore: For me all appearances of `__del__` cleanup are gone in the testsuite except the one in test_remote_repo_restrict_to_path (which I think is caused by `patch.object` basically overriding the context manager implemented in Repository, i.e. the patched object is bound instead of the Repository).

borg/testsuite/repository.py
```diff
@@ -100,13 +100,15 @@ def test_single_kind_transactions(self):
         self.repository.close()
         # replace
         self.repository = self.open()
-        self.repository.put(b'00000000000000000000000000000000', b'bar')
-        self.repository.commit()
-        self.repository.close()
+        with self.repository:
+            self.repository.put(b'00000000000000000000000000000000', b'bar')
+            self.repository.commit()
+            self.repository.close()
```
ThomasWaldmann: with ... AND .close() ?

borg/testsuite/archiver.py
enkore: True, the later bind would still fix this, but 'open's aren't exactly cheap. I'll patch it.

borg/testsuite/key.py
enkore: Got a few failures due to un-initialized loggers when running this one alone without running the entire suite. When running the test suite the archiver test cases will do this.

Maybe put it in __init__ of the test suite instead, to ensure that the loggers are always there for all tests?

ThomasWaldmann: yes, i'ld rather have it in test init than at module import time.

borg/testsuite/repository.py
enkore: Yeah, that's superfluous. close() is idempotent, so won't blow up either, but it doesn't belong there.

borg/testsuite/key.py
enkore: Yep `__init__` works too, so I'll put it there instead.

borg/remote.py
enkore: Ah, of course, in the tests tearDown performs "hard" tear downs and just rmtree()s the repository. That probably raced with `__del__` and lead to the locks being deleted before del had a change to clean them up.

enkore:
I guess technically codecov isn't quite right that only 85 % hit, it's just that the llfuse tests don't run on travis... :D

ThomasWaldmann:
Hmm, it's a bit hard to review if you immediately collapse your new changes into your old ones.

enkore:
The only thing changed other than the comment-diffs above is the new changes in upgrader.py, which where in a stash I forgot previously. ([diff](https://gist.github.com/enkore/2175f087dd41ec353d92)). Next time I'll use squash-before-merge approach for larger changeset again

PR #783, opened 2016-03-21
Improved diff format

Ape:
This is an implementation for #757. I have tested the implementation with my real snapshots, and I like to use this output format myself. The format is similar to what I described in #757, but I made some changes. I'd like to hear any comments and ideas about this format and implementation details.

The main design goals of the new format:
 - One file takes exactly one line of output
 - The format is easy to read with typical, long list of changes
 - Metadata changes are visible and easy to examine
 - Unuseful information is not shown

~~There are many possible special cases where regular files are replaced with links, or where hard links are replaced by symbolic links etc. In this output format, these modifications are indicated on the file mode. For convenience, I introduce a new file mode notation for hard links. Just like symbolic links have mode ```l```, in this output hard links are indicated by ```h```.~~

- [x]  Basic functionality
- [x] Preserve archive order of files
  - Preserve order inside added files and inside removed files
  - ~~Interleave modified files, added files and removed files~~ (Out of the scope of this PR, see #797)
- [x] Figure out what to do with hard links
  - See #763 for solution
- [x] Figure out what is the correct thing to do with incomparable chunk ids
- ~~Add statistics for the total difference of the archives~~ (Out of the scope of this PR, see #796)
  - ~~Size of the snapshots~~
  - ~~Total bytes added / removed~~
  - ~~Number of chunks changed~~
- [x] Update example output to match the final version

This pull request has the rebased commits ready to be merged. If you want to see chronological iterative commits check [Ape/borg/feature/improved_diff_format](/Ape/borg/commits/feature/improved_diff_format).

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from itertools import zip_longest
 from operator import attrgetter
 import argparse
+import collections
 import functools
 import hashlib
 import inspect
@@ -546,8 +547,8 @@ class Archiver:
                 print("{:<19} {}".format(' '.join(changes), remove_surrogates(path)))
 
         def compare_archives(archive1, archive2, matcher):
-            orphans_archive1 = {}
-            orphans_archive2 = {}
+            orphans_archive1 = collections.OrderedDict()
+            orphans_archive2 = collections.OrderedDict()
             hardlink_masters = {}
             for item1, item2 in zip_longest(
                     archive1.iter_items(lambda item: matcher.match(item[b'path'])),
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -433,15 +433,16 @@ class Archiver:
     @with_archive
     def do_diff(self, args, repository, manifest, key, archive):
         """Diff contents of two archives"""
-        def format_bytes(count):
-            if count is None:
-                return "<deleted>"
-            return format_file_size(count)
-
         def fetch_and_compare_chunks(chunk_ids1, chunk_ids2, archive1, archive2):
             chunks1 = archive1.pipeline.fetch_many(chunk_ids1)
             chunks2 = archive2.pipeline.fetch_many(chunk_ids2)
             return self.compare_chunk_contents(chunks1, chunks2)
+
+        def sum_chunk_size(item):
+            if item.get(b'deleted'):
+                return None
+            else:
+                return sum(c[1] for c in item[b'chunks'])
 
         def get_owner(item):
             if args.numeric_owner:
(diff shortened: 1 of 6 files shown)


enkore:
I like it. Maybe put an explanation of the output format in the --help, especially the `h` notation for hard links (n>2)?

Ape:
The output format could use a comprehensive documentation, but I'm not sure how to describe it briefly enough for ```--help```. Do you think it would be enough to add something like this:
> 'borg diff' output has a custom file mode 'h' to indicate hard links.

Ape:
I'd like to have the output lines sorted by the path. However, it seems that it is not trivial to implement, so it is out of the scope of this pull request. We probably need to add a command line option that first goes through all the files and then sorts the whole output.

Ape:
I pushed updates to the pull request. Now it mentions about the custom file mode on ```--help```. I also added more tests. To see the same changes in chronological order see [Ape/borg/feature/improved_diff_format](/Ape/borg/commits/feature/improved_diff_format).

enkore:
Can you update the output of the example in the Sphinx docs as well? It would be confusing if these were inconsistent to what Borg will do.

ThomasWaldmann:
no sorting please, keep archive order.

Ape:
I pushed more changes rebasing the pull request on the latest master commit. I added more tests and updated the docs.

ThomasWaldmann:
@Ape if you use same status characters as borg create --list uses, you maybe could just refer to their docs.

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
-                return "<deleted>"
-            return format_file_size(count)
+            if count > 0:
+                sign = '+'
+            elif count < 0:
+                sign = '-'
+            else:
+                sign = ''
+
+            return '{}{}'.format(sign, format_file_size(abs(count)))
```
ThomasWaldmann: maybe that functionality could be moved into format_file_size? or a separate helper function.

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
                 return item[b'uid'], item[b'gid']
             else:
                 return item[b'user'], item[b'group']
 
+        def get_mode(item):
+            mode = stat.filemode(item[b'mode'])
+
+            # Special notation for hard links
+            if mode[0] == '-' and b'source' in item:
+                mode = "h" + mode[1:]
```
ThomasWaldmann: hmm, not sure that is helpful.

"hardlinks" are just additional names pointing to same inode. there is no difference in principle between first name and additional names in the filesystem.

in a borg archive, there is a difference: first is with data, 2nd+ is without data, pointing backwards to first one.
but that is only to save storing the data again and also to trigger the hardlinking in the fs.

but we should not display the first name as "normal file" and the 2nd+ as "hardlink", that gives an asymetric impression of the symetric filesystem reality.

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+
+            return mode
+
+        def describe_link(item):
+            if get_mode(item)[0] == 'l':
+                return 'symlink'
+            else:
+                return 'hard link'
```
ThomasWaldmann: maybe similar here, see above.

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+
+            if mode1 != mode2:
+                return '[{} -> {}]'.format(mode1, mode2)
+
         def compare_items(path, item1, item2, deleted=False):
             """
             Compare two items with identical paths.
             :param deleted: Whether one of the items has been deleted
             """
+            line = []
```
ThomasWaldmann: instead of "line", you could use the more meaningful "changes". plural also indicates better that this is a list.

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+                    elif total2 is None:
+                        return 'removed {:>9}'.format(format_file_size(total1))
+                    else:
+                        return '{:>8} {:>8}'.format(format_bytes(added),
+                                                    format_bytes(removed))
+
+        def compare_directory(item1, item2):
+            if not item1.get(b'deleted') and item2.get(b'deleted'):
+                if get_mode(item1)[0] == 'd':
+                    return 'removed directory'
```
ThomasWaldmann: same as:
if item1 == d and not item1 deleted and item2 deleted:

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+            elif not item2.get(b'deleted') and item1.get(b'deleted'):
+                if get_mode(item2)[0] == 'd':
+                    return 'added directory'
```
ThomasWaldmann: similar here (and at other places in this changeset)

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+                chunk_id_set1 = set(chunk_ids1)
+                chunk_id_set2 = set(chunk_ids2)
+                total1 = sum_chunk_size(item1)
+                total2 = sum_chunk_size(item2)
+
+                if (can_compare_chunk_ids or total1 != total2 or
+                    not fetch_and_compare_chunks(chunk_ids1, chunk_ids2,
+                                                    archive1, archive2)):
+                    added = sum(c[1] for c in (chunk_id_set2 - chunk_id_set1))
+                    removed = -sum(c[1] for c in (chunk_id_set1 - chunk_id_set2))
```
ThomasWaldmann: if can_compare_chunk_ids is false (but one of the other terms is true), this is executed and computes set difference if INCOMPARABLE chunk ids.

doesn't look like making sense. :)

borg/archiver.py
```diff
@@ -406,64 +406,127 @@ def item_is_hardlink_master(item):
...
+
+                    if total1 is None:
+                        return 'added {:>11}'.format(format_file_size(total2))
+                    elif total2 is None:
+                        return 'removed {:>9}'.format(format_file_size(total1))
+                    else:
+                        return '{:>8} {:>8}'.format(format_bytes(added),
+                                                    format_bytes(removed))
```
ThomasWaldmann: in fact we have more interesting data in this case:
the file size difference total2 - total1 and the amount of chunks added/removed in the repo.

Ape:
@ThomasWaldmann I kind of agree with your comment about hard links. Do you propose that if there are three links to the same file and the file contents are changed, it should look something like this:

```
   +36 B   -169 B path/to/hardlink1
   +36 B   -169 B path/to/hardlink2
   +36 B   -169 B path/to/hardlink3
```

How should this be implemented? Yes, in the filesystem the situation is completely symmetric, any of those three links are not special compared to the others. However, in borg what I get is one file with changed chunks and two links that are still pointing to the same source, unchanged besides perhaps the modification time.

How should we even detect that the contents are changed when we only have the source file path. As far as I know, we cannot even get evaluate the path and get the file node with the chunks.

borg/archiver.py
Ape: @ThomasWaldmann Hmm, I tried to keep the chunk checking behavior unchanged from the current version in master. I have to admit that the semantics here are not completely clear to me. What should be done if can_compare_chunk_ids is true and what should be done when it's false?

borg/archiver.py
Ape: @ThomasWaldmann True, but I wanted to keep the output lines short so that it is still easy to ready. The size difference can be seen from the ```+``` and ```-``` information. Amount of chunks might be an interesting piece of information, but for me it wouldn't be worth the wasted output space.

enkore:
On 24.03.2016 09:01, Lauri Niskanen wrote:
> @ThomasWaldmann <https://github.com/ThomasWaldmann> I kind of agree with
> your comment about hard links. Do you propose that if there are three
> links to the same file and the file contents are changed, it should look
> something like this:
> 
> |+36 B -169 B path/to/hardlink1 +36 B -169 B path/to/hardlink2 +36 B -169
> B path/to/hardlink3 |
> 
> How should this be implemented? Yes, in the filesystem the situation is
> completely symmetric, any of those three links are not special compared
> to the others. However, in borg what I get is one file with changed
> chunks and two links that are still pointing to the same source,
> unchanged besides perhaps the modification time.
> 
> How should we even detect that the contents are changed when we only
> have the source file path. As far as I know, we cannot even get evaluate
> the path and get the file node with the chunks.
> 

In extract there is a similar issue (see #763), you can use the same
approach here as well.

Ape:
I changed to orphan sets to be instances of OrderedDict. This way the archive order of added and removed files is preserved. I'd still like to preserve the order as a whole and interleave modified files, added files, and removed files. This might be difficult to implement though.

Ape:
I replaced the special hard link notation with hard links just showing up as (duplicate) regular files. Please review.

borg/archiver.py
enkore: (Didn't receive a notification about the mention)

can_compare_chunk_ids tells us whether the chunker params are the same. We end up in this branch in two cases:

1. We can compare chunk IDs, and the chunk ID lists were different
2. We can't compare chunk IDs (*we do not know whether the files are different*)

This second `if` is a continuation of these two branches

1. Either we hit branch (1) above (can compare, were different) then added/remove info is calculated
2. Or we hit branch (2): check if the file sizes are the same. If they aren't something must have changed. If they are the same, compare file contents.

You inverted the logic from the old commit, though. I think this needs an `else: return <nothing changed>`:

* Can't compare chunk ids
* So we hit branch (2) in the first if
* Now we do check 2, if the files are the same this would still print something like `[12345 12345]`

borg/archiver.py
```diff
@@ -405,86 +406,145 @@ def item_is_hardlink_master(item):
...
+
+            if get_mode(item1)[0] == 'l' or get_mode(item2)[0] == 'l':
+                changes += filter(None, [compare_link(item1, item2)])
+
+            if b'chunks' in item1 and b'chunks' in item2:
+                changes += filter(None, [compare_content(path, item1, item2,
+                                                         hardlink_masters)])
+
+            if get_mode(item1)[0] == 'd' or get_mode(item2)[0] == 'd':
+                changes += filter(None, [compare_directory(item1, item2)])
```
enkore: How about `changes.append(compare_XYZ)` and one `filter(None, changes)` in the end?

borg/archiver.py
```diff
@@ -405,86 +406,145 @@ def item_is_hardlink_master(item):
...
-                removed = sum(c[1] for c in (chunk_id_set1 - chunk_id_set2))
-                print(remove_surrogates(path), 'different contents')
-                print('\t +%s, -%s, %s, %s' % (format_bytes(added), format_bytes(removed),
-                                               format_bytes(total1), format_bytes(total2)))
+                changes += filter(None, [compare_owner(item1, item2)])
+                changes += filter(None, [compare_mode(item1, item2)])
+
+            if changes:
+                print("{:<19} {}"
+                      .format(' '.join(changes), remove_surrogates(path)))
```
enkore: e.g.`' '.join(filter(None, changes))`

Ape:
@enkore Thanks for your comments. I think I managed to make the content comparison logic simple and easy to follow, at least compared to the previous version :).

Does anybody have still comments about the pull request?

borg/archiver.py
```diff
@@ -405,86 +406,150 @@ def item_is_hardlink_master(item):
...
+            if can_compare_chunk_ids:
+                return item1[b'chunks'] != item2[b'chunks']
+            else:
+                if sum_chunk_size(item1) != sum_chunk_size(item2):
+                    return True
+                else:
+                    chunk_ids1 = (c[0] for c in item1[b'chunks'])
+                    chunk_ids2 = (c[0] for c in item2[b'chunks'])
+                    return not fetch_and_compare_chunks(chunk_ids1, chunk_ids2,
+                                                        archive1, archive2)
```
enkore: That's much nicer than before! :-)

Ape: Good catch. I'll fix it.

Ape:
I did some minor fixes. I think this PR could be now ready for a merge.

borg/archiver.py
```diff
@@ -405,86 +406,148 @@ def item_is_hardlink_master(item):
...
+            orphans_archive2 = collections.OrderedDict()
+            hardlink_masters = {}
             for item1, item2 in zip_longest(
                     archive1.iter_items(lambda item: matcher.match(item[b'path'])),
                     archive2.iter_items(lambda item: matcher.match(item[b'path'])),
             ):
                 if item1 and item2 and item1[b'path'] == item2[b'path']:
-                    compare_items(item1[b'path'], item1, item2)
+                    compare_items(item1[b'path'], item1, item2,
+                                  hardlink_masters)
```
ThomasWaldmann: hmm, are you doing these linebreaks to obey pep8 80chars limit?
we do not follow that in borg project, having lines with up to 120 chars is fine.

can you reformat the places where you introduced "80chars linebreaks"?

borg/testsuite/archiver.py
```diff
@@ -1225,33 +1225,101 @@ class DiffArchiverTestCase(ArchiverTestCaseBase):
...
+
+            # Directory replaced with a regular file
+            assert '[drwxr-xr-x -> -rwxr-xr-x] input/dir_replaced_with_file' in output
+
+            # Basic directory cases
+            assert 'added directory     input/dir_added' in output
+            assert 'removed directory   input/dir_removed' in output
+
+            # Basic symlink cases
+            assert 'link changed        input/link_changed' in output
```
ThomasWaldmann: 'changed link'? same word order as next 2 lines?

ThomasWaldmann:
@Ape yes, looks pretty good! thanks for improving it.

I found only some cosmetic things, see my comments.

borg/archiver.py
Ape: Yeah, I generally try to obey the 80 character limit. I'm happy to make a bit longer lines if the project style is like that.

Ape:
Thanks for your comments. I made the required changes, and also rebased the commits on the latest master.

PR #789, opened 2016-03-23
Replace stdlib hmac, zero-copy decrypt

enkore:
AESKeyBase.decrypt makes less copies of the data, specifically data[41:],
which copies the payload, is gone. This is possible since this commit
makes our lil' crypto API compatible with simple buffers.

In Cython there is a syntax for creating memoryviews from buffer-enabled
objects, however, it doesn't support read-only buffers. ro_buffer gets
this job done, though, and also does the proper type checking
(by PyBUF_SIMPLE).

Note: msgpack doesn't support memoryviews. Otherwise we could avoid
      copying the en/de-crypted output with the final "return out[:ctl]"

This commit leads to a solid 10-15 % increase in performance of hash-lists
and extract.

--

We could also avoid the outgoing copy in AES, but memoryviews are not handled by msgpack. See msgpack/msgpack-python#126

(Technically the HMAC of the stdlib uses OpenSSL for the hashing via the hashlib module, but this implementation avoids creating new Python objects and doesn't run the HMAC initialization code in Python)

--

GIL-handling: I opted to not unlock the GIL in hmac_sha256 for now. We can add such code when it is needed.

Dependency changes: None

### DIFF
--- a/borg/crypto.pyx
+++ b/borg/crypto.pyx
@@ -3,8 +3,10 @@
 This could be replaced by PyCrypto maybe?
 """
 from libc.stdlib cimport malloc, free
+from cpython.buffer cimport PyBUF_SIMPLE, PyObject_GetBuffer, PyBuffer_Release
 
 API_VERSION = 2
+
 
 cdef extern from "openssl/rand.h":
     int  RAND_bytes(unsigned char *buf, int num)
@@ -35,6 +37,14 @@ cdef extern from "openssl/evp.h":
     int EVP_EncryptFinal_ex(EVP_CIPHER_CTX *ctx, unsigned char *out, int *outl)
     int EVP_DecryptFinal_ex(EVP_CIPHER_CTX *ctx, unsigned char *out, int *outl)
 
+    EVP_MD *EVP_sha256() nogil
+
+
+cdef extern from "openssl/hmac.h":
+    unsigned char *HMAC(const EVP_MD *evp_md,
+                    const void *key, int key_len,
+                    const unsigned char *data, int data_len,
+                    unsigned char *md, unsigned int *md_len) nogil
 
 import struct
 
@@ -44,6 +54,12 @@ _long = struct.Struct('>Q')
 bytes_to_int = lambda x, offset=0: _int.unpack_from(x, offset)[0]
 bytes_to_long = lambda x, offset=0: _long.unpack_from(x, offset)[0]
 long_to_bytes = lambda x: _long.pack(x)
+
+
+cdef Py_buffer ro_buffer(object data) except *:
+    cdef Py_buffer view
+    PyObject_GetBuffer(data, &view, PyBUF_SIMPLE)
+    return view
 
 
 def num_aes_blocks(int length):
@@ -95,6 +111,7 @@ cdef class AES:
         return self.ctx.iv[:16]
 
     def encrypt(self, data):
+        cdef Py_buffer data_buf = ro_buffer(data)
         cdef int inl = len(data)
         cdef int ctl = 0
         cdef int outl = 0
@@ -103,7 +120,7 @@ cdef class AES:
         if not out:
             raise MemoryError
         try:
-            if not EVP_EncryptUpdate(&self.ctx, out, &outl, data, inl):
+            if not EVP_EncryptUpdate(&self.ctx, out, &outl, <const unsigned char*> data_buf.buf, inl):
                 raise Exception('EVP_EncryptUpdate failed')
             ctl = outl
             if not EVP_EncryptFinal_ex(&self.ctx, out+ctl, &outl):
@@ -112,8 +129,10 @@ cdef class AES:
             return out[:ctl]
         finally:
             free(out)
+            PyBuffer_Release(&data_buf)
 
     def decrypt(self, data):
+        cdef Py_buffer data_buf = ro_buffer(data)
         cdef int inl = len(data)
         cdef int ptl = 0
         cdef int outl = 0
@@ -124,7 +143,7 @@ cdef class AES:
         if not out:
             raise MemoryError
         try:
-            if not EVP_DecryptUpdate(&self.ctx, out, &outl, data, inl):
+            if not EVP_DecryptUpdate(&self.ctx, out, &outl, <const unsigned char*> data_buf.buf, inl):
                 raise Exception('EVP_DecryptUpdate failed')
             ptl = outl
             if EVP_DecryptFinal_ex(&self.ctx, out+ptl, &outl) <= 0:
@@ -136,3 +155,20 @@ cdef class AES:
             return out[:ptl]
         finally:
             free(out)
+            PyBuffer_Release(&data_buf)
+
+
+def hmac_sha256(key, data):
+    md = bytes(32)
+    cdef Py_buffer data_buf = ro_buffer(data)
+    cdef const unsigned char *key_ptr = key
+    cdef int key_len = len(key)
+    cdef unsigned char *md_ptr = md
+    try:
+        with nogil:
+            rc = HMAC(EVP_sha256(), key_ptr, key_len, <const unsigned char*> data_buf.buf, data_buf.len, md_ptr, NULL)
+        if rc != md_ptr:
+            raise Exception('HMAC(EVP_sha256) failed')
+    finally:
+        PyBuffer_Release(&data_buf)
+    return md
--- a/borg/key.py
+++ b/borg/key.py
@@ -4,14 +4,14 @@ import getpass
 import os
 import sys
 import textwrap
-from hmac import HMAC, compare_digest
+from hmac import compare_digest
 from hashlib import sha256, pbkdf2_hmac
 
 from .helpers import IntegrityError, get_keys_dir, Error, yes
 from .logger import create_logger
 logger = create_logger()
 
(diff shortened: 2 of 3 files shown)


borg/key.py
```diff
@@ -126,28 +126,29 @@ class AESKeyBase(KeyBase):
     def id_hash(self, data):
         """Return HMAC hash using the "id" HMAC key
         """
-        return HMAC(self.id_key, data, sha256).digest()
+        return hmac_sha256(self.id_key, data)
 
     def encrypt(self, data):
         data = self.compressor.compress(data)
         self.enc_cipher.reset()
         data = b''.join((self.enc_cipher.iv[8:], self.enc_cipher.encrypt(data)))
```
enkore: We could also avoid this copy, since the outgoing buffer is controlled by AES. I.e. we could make the buffer 8 bytes larger in the first place and pass out+8 to EVP_EncryptUpdate.

borg/crypto.pyx
```diff
@@ -124,7 +143,7 @@ cdef class AES:
         if not out:
             raise MemoryError
         try:
-            if not EVP_DecryptUpdate(&self.ctx, out, &outl, data, inl):
+            if not EVP_DecryptUpdate(&self.ctx, out, &outl, <const unsigned char*> data_buf.buf, inl):
```
ThomasWaldmann: So Cython is creating a copy of data before giving it to DecryptUpdate?

borg/crypto.pyx
```diff
@@ -136,3 +155,16 @@ cdef class AES:
...
         finally:
             free(out)
+            PyBuffer_Release(&data_buf)
+
+
+def hmac_sha256(key,  data):
+    cdef Py_buffer data_buf = ro_buffer(data)
+    md = bytes(32)
+    try:
+        if not HMAC(EVP_sha256(), <const unsigned char*> key, len(key),
```
ThomasWaldmann: Isn't the first param just a pointer? ("EVP_sha256", without the "()")?

borg/key.py
```diff
@@ -126,28 +126,29 @@ class AESKeyBase(KeyBase):
     def id_hash(self, data):
         """Return HMAC hash using the "id" HMAC key
         """
-        return HMAC(self.id_key, data, sha256).digest()
+        return hmac_sha256(self.id_key, data)
```
ThomasWaldmann: Seem like less flexible / harder to change than before.

One of my not-yet-merged changes in experimental branch was to make the hash/hmac flexible.
E.g. sha512-256 is faster on 64bit than sha256.

borg/crypto.pyx
enkore: My understanding is that Cython's `<>` operator has two fundamentally different meanings:

1. Python objects are involved, then it will do a Python-like type coercion (which will very likely make a copy)
2. The argument and the target type are C types: Same semantics as a C cast (never copies)

data_buf.buf is a `void*`, which is why a cast is needed here.

borg/key.py
ThomasWaldmann: Yes. But the general question is: how much does all this optimization save us (in percent of total execution time) and cost us (in more code)?

borg/key.py
```diff
@@ -126,28 +126,29 @@ class AESKeyBase(KeyBase):
...
-        hmac_given = memoryview(data)[1:33]
-        hmac_computed = memoryview(HMAC(self.enc_hmac_key, memoryview(data)[33:], sha256).digest())
+        data_view = memoryview(data)
+        hmac_given = data_view[1:33]
+        hmac_computed = memoryview(hmac_sha256(self.enc_hmac_key, data_view[33:]))
         if not compare_digest(hmac_computed, hmac_given):
             raise IntegrityError('Encryption envelope checksum mismatch')
         self.dec_cipher.reset(iv=PREFIX + data[33:41])
-        data = self.compressor.decompress(self.dec_cipher.decrypt(data[41:]))
+        data = self.compressor.decompress(self.dec_cipher.decrypt(data_view[41:]))
```
ThomasWaldmann: ah, good :)

borg/testsuite/crypto.py
```diff
@@ -28,3 +28,29 @@ def test_aes(self):
...
+        key = b'\xaa' * 20
+        data = b'\xdd' * 50
+        hmac = unhexlify('773ea91e36800e46854db8ebd09181a7'
+                         '2959098b3ef8c122d9635514ced565fe')
+        assert hmac_sha256(key, data) == hmac
+        key = unhexlify('0102030405060708090a0b0c0d0e0f10'
+                        '111213141516171819')
+        data = b'\xcd' * 50
+        hmac = unhexlify('82558a389a443c0ea4cc819899f2083a'
+                         '85f0faa3e578f8077a2e3ff46729665b')
```
ThomasWaldmann: i'ld prefer the hex values in one line rather than splitting them (just to obey 80chars limit?).

borg/crypto.pyx
enkore: I haven't quite understood what OpenSSL does there (and the code is very opaque). My understanding from the docs and their examples is, that EVP_sha256() or EVP_aes_whatever_cbc() are just functions that return a pointer to some static structure in the OpenSSL library that just contains further pointers to functions actually implementing these algorithms. I don't think these should be free()d.

Interestingly enough none of these functions are actually defined in the OpenSSL source tree, not even after compilation, so they are probably generated and put somewhere outside the tree.

borg/key.py
enkore: I'd say about one added line of Cython and one line shortened here? Performance difference a few percent at best.

I think the more important question is whether we want to mix AESKeyBase and crypot.AES on this level, as the code is now crypto.AES doesn't know and doesn't care where the IV comes from or how it's stored, this change would intertwine this implementation detail of AESKeyBase with crypto.AES, which is why I only commented instead of committing. (And what I probably should've remarked in the comment as well)

borg/crypto.pyx
ThomasWaldmann: I think we should be sure about that it in fact is a copy otherwise we would "optimize" it for no reason.

borg/testsuite/crypto.py
enkore: I pasted them as they appeared in the RFC, so it's easy to see that these are really the same values as in the RFC.

borg/crypto.pyx
enkore: Here's the C code Cython generates for this:

```
__pyx_t_3 = ((!(EVP_DecryptUpdate((&__pyx_v_self->ctx), __pyx_v_out, (&__pyx_v_outl), ((unsigned char const *)__pyx_v_data_buf.buf), __pyx_v_inl) != 0)) != 0);
```

It's really a straight C-cast of data_buf.buf

borg/key.py
enkore: I don't think that it makes a big difference when changing HMACs whether we'd need to edit all usages of the HMAC as before, or edit/copy hmac_sha256 and edit all occurences as well. In both cases the work surrounding the change would be much larger (upgrader etc.)

borg/crypto.pyx
ThomasWaldmann: I meant the code before your optimization.

enkore: This is the code as generated by this commit for that function call.

`__pyx_v_data_buf` is the buffer as created by ro_buffer, and PyObject_GetBuffer never copies (for any type it should either provide direct access to the underlying memory of the object as specified by the flags etc. or it should raise an exception).

borg/crypto.pyx
```diff
@@ -103,7 +120,7 @@ cdef class AES:
         if not out:
             raise MemoryError
         try:
-            if not EVP_EncryptUpdate(&self.ctx, out, &outl, data, inl):
```
ThomasWaldmann: http://docs.cython.org/src/tutorial/strings.html : 
"""
cdef char* other_c_string = py_string

This is a very fast operation after which other_c_string points to the byte string buffer of the Python string itself.
"""

Ape:
I would be very interested in seeing some benchmark results.

borg/crypto.pyx
enkore: The point of this changeset is that en/decrypt accept memory views now (and bytes, bytearrays etc. like before), so that the stuff in AESKeyBase doesn't need to copy the payload.

enkore:
We don't have a benchmark that's large enough. The numbers above come from tests on an archive of a Linux system with no user data (approx 20 GB and 550k files) and my "largetest" test case (Linux 4.4.2 and 4.5 source trees in two archives). The latter is about 820 MB (uncompressed) and ~250 MB or so with LZ4. "largetest" is already borderline for small optimizations.

It would be nice if we had a benchmark (not only test data but also automatic execution) that's a couple GB worth of data and could be easily shared / compared among machines. Sounds like a good bit of effort though. (unless something like this already exists in a reusable form -- I think most archiver bench marks are too small (a couple hundred MB at most).

ThomasWaldmann:
hmm, why did you "opt to not unlock the gil"? the stdlib function does it.

enkore:
HMAC now unlocks the GIL, perf numbers don't look different (as expected).

ThomasWaldmann:
thanks. sure it does not perform (significantly) different on master.

it might do when merging into multithreading branch.

PR #790, opened 2016-03-24
Ignore SIG_PIPE signal

Ape:
Without this change borg will print errors and stack trace when it's output is interrupted with command line pipes. For example:
```
$ borg diff ::a b | less # then quit with 'q'
```

There error without this was:
```
BrokenPipeError: [Errno 32] Broken pipe
```

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1646,6 +1646,8 @@ def setup_signal_handlers():  # pragma: no cover
     for sig in sigs:
         signal.signal(sig, sig_info_handler)
 
+    signal.signal(signal.SIGPIPE, signal.SIG_DFL)
+
 
 def main():  # pragma: no cover
     # provide 'borg mount' behaviour when the main script/executable is named borgfs


enkore:
I think it would be better if SIGPIPE were handled specifically for commands where it makes sense (e.g. list and diff). For both of them it makes sense to cleanly abort the operation when it is received, instead of ignoring it.

Ape:
You are right. Let's make a new pull request with that behavior.

ThomasWaldmann:
Asked differently: for which commands would this changeset NOT make sense?

Ape:
Let's look at the list of commands:

    serve               start repository server process
    init                initialize empty repository
    check               verify repository
    change-passphrase   change repository passphrase
    migrate-to-repokey  migrate passphrase-mode repository to repokey
    create              create backup
    extract             extract archive contents
    diff                find differences in archive contents
    rename              rename archive
    delete              delete archive
    list                list archive or repository contents
    mount               mount repository
    info                show archive information
    break-lock          break repository and cache locks
    prune               prune archives
    upgrade             upgrade repository format

`diff` and `list` are the ones where this change is really needed. Also I don't see any reason to not ignore the signal on `info`.

-

> E.g. if the output of borg create is piped into an external logger and that logger dies, then this shouldn't be ignored

If both stdout and stderr were redirected to the logger and the logger dies, there is no way to print any error messages or stack traces. If only stdout was redirected, then we can still print the errors, so it might be useful sometimes. However, does the user ever see even these since he was relying on the logger program that died?

We could of course still abort our operation gracefully and clean up things when we get the exception.

Ape:
Actually, I did some testing. This changeset is dangerous even with `diff`. It leaves shared locks reserved if I close the pipe early. So we need a way to clean up things properly. Maybe the same logic should be applied to `Ctrl-C` / `SIGTERM` / `KeyboardInterrupt`.

ThomasWaldmann:
OK, thanks for researching / clarifying.

PR #793, opened 2016-03-25
Installation: Mention GNU Guix package.

lfam:
* docs/installation.rst: Add GNU Guix package to list of distribution packages.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -34,6 +34,7 @@ Distribution Source                                        Command
 ============ ============================================= =======
 Arch Linux   `[community]`_                                ``pacman -S borg``
 Debian       `stretch`_, `unstable/sid`_                   ``apt install borgbackup``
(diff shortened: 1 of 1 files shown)


infectormp:
GNU Guix is a package manager, but this section in docs for os distribution. Maybe need make section **Using GNU Guix**?

lfam:
@infectormp The package is for the GNU Guix package manager, which can be used on any GNU / Linux distro. But, you can also use it on the operating system GNU GuixSD, which includes Guix.

Besides, Homebrew is already in that list, and Homebrew isn't an operating system at all.

infectormp:
@lfam Homebrew strictly for OSX, so it normal have it in OS distribution section.

infectormp:
But this only my opinion.

enkore:
The section is "Installation => Distribution Package". "OS" does not appear, and the title doesn't imply it.

PR #805, opened 2016-03-27
Do not handle intentional aborts as errors

Ape:
KeyboardInterrupts (`Ctrl-C` / `SIGINT`) or BrokenPipeErrors (`SIGPIPE`) are usually done intentionally by the user. With this commit borg aborts silently without printing error messages or stack traces. Command line users should understand what is happening if they abort the process. Intentional aborts are often part of the command usage or scripts. Additional messages or traces would only disturb the user or make output parsing more difficult. This is especially true with read-only commands like info, list, and diff.

It might be argued that BrokenPipeError is sometimes an actual, unexpected error, but in that case there is no way to print any error messages or stack traces anyway.

Borg should abort gracefully so that it doesn't corrupt the repository or otherwise leave it in a state where further operations would be obstructed. This should already be the case; this commit only affects error message printing ~~and the return code~~. If there are problems in the abort behaviour with some commands, then it is a separate issue that needs to be fixed.

This commit is continuation of #790. The previous pull request was problematic since catching the signal like that didn't give Python a change to exit properly. With this new pull request borg will release the repository locks and run any other clean up code correctly.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1730,11 +1730,11 @@ def main():  # pragma: no cover
     except RemoteRepository.RPCError as e:
         msg = '%s\n%s' % (str(e), sysinfo())
         exit_code = EXIT_ERROR
+    except (KeyboardInterrupt, BrokenPipeError):
+        sys.stderr.close()
+        exit_code = EXIT_ERROR
     except Exception:
         msg = 'Local Exception.\n%s\n%s' % (traceback.format_exc(), sysinfo())
-        exit_code = EXIT_ERROR
-    except KeyboardInterrupt:
-        msg = 'Keyboard interrupt.\n%s\n%s' % (traceback.format_exc(), sysinfo())
         exit_code = EXIT_ERROR
     if msg:
         logger.error(msg)


borg/archiver.py
```diff
@@ -1730,12 +1730,12 @@ def main():  # pragma: no cover
     except RemoteRepository.RPCError as e:
         msg = '%s\n%s' % (str(e), sysinfo())
         exit_code = EXIT_ERROR
+    except (KeyboardInterrupt, BrokenPipeError):
+        sys.stderr.close()
+        exit_code = EXIT_SUCCESS
```
ThomasWaldmann: it didn't complete, so it is not success.

enkore: Explicitly closing `sys.stderr` looks strange to me.

Ape: @ThomasWaldmann My logic here is that `borg` didn't really have any errors. The user just decided that he doesn't want run the command further. However, I am okay with non-zero return code, too, if you think it's better that way.

@enkore It is to avoid any further writes. Otherwise TextIOWrapper will try to use it and get an error after this exception handling.

ThomasWaldmann: the current definition of rc codes is:
0 = success, code reached normal end, no issues
1 = warning, code reached normal end of execution, but there were some issues (like issues with single files)
2 = error, code did not reach normal end, but somehow aborted early / unexpectedly

Ape:
@ThomasWaldmann I changed the return code to `EXIT_ERROR`.

ThomasWaldmann:
When borg does a remote backup via ssh, it spawns a ssh process and communicates with it via pipes. If these pipes break (do to some error condition, not due to the user willingly breaking them), wouldn't that also give BrokenPipe exception?

borg's stderr/stout would be connected to the console normally, so it could print an error msg / exception in that case.

Ape:
@ThomasWaldmann That may be an issue. More research is needed.

ThomasWaldmann:
As a general note:

I'ld rather like to see some tracebacks too much than tracebacks missing when one needed them.

attic was a real pain because it "prettified" too many error conditions by suppressing the traceback and replacing it by a 1 line error msg - so bug reports sometimes were pretty useless because they did not contain enough information.

Ape:
Yes, I agree. However, it's also important to show error messages only when there is an error. If usual work flow (without any issues occurring) shows error messages then the actual errors may go unnoticed.

ThomasWaldmann:
this doesn't seem to advance. close?

PR #810, opened 2016-03-28
Flexible compression - pattern matching on path/filename

ThomasWaldmann:
current state:
- --compression-from reads a file with a list compression-spec:matching-pattern that is used to set up CompressionDecider1
- docs/misc/compression.conf example file (currently used to disable compression)
- if no match, compresses like given by --compression
- CompressionDecider2 operates on chunks - it does not do much yet (out of scope of this PR), just either gets the desired compression from Chunk meta or returns the default compression (set by --compression) if there is no meta entry.
- debug log tells filenames and compression
- clean_lines() can be used later to deduplicate similar code at other places

TODO:
- [ ] support --compression-from for recreate, see `TODO` in the code

### DIFF
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -31,6 +31,7 @@ from . import hashindex
 from . import chunker
 from .constants import *  # NOQA
 from . import crypto
+from .compress import COMPR_BUFFER
 from . import shellpattern
 import msgpack
 import msgpack.fallback
@@ -1423,3 +1424,77 @@ except ImportError:
 
 def scandir_inorder(path='.'):
     return sorted(scandir(path), key=lambda dirent: dirent.inode())
+
+
+def clean_lines(lines, lstrip=None, rstrip=None, remove_empty=True, remove_comments=True):
+    """
+    clean lines (usually read from a config file):
+
+    1. strip whitespace (left and right), 2. remove empty lines, 3. remove comments.
+
+    note: only "pure comment lines" are supported, no support for "trailing comments".
+
+    :param lines: input line iterator (e.g. list or open text file) that gives unclean input lines
+    :param lstrip: lstrip call arguments or False, if lstripping is not desired
+    :param rstrip: rstrip call arguments or False, if rstripping is not desired
+    :param remove_comments: remove comment lines (lines starting with "#")
+    :param remove_empty: remove empty lines
+    :return: yields processed lines
+    """
+    for line in lines:
+        if lstrip is not False:
+            line = line.lstrip(lstrip)
+        if rstrip is not False:
+            line = line.rstrip(rstrip)
+        if remove_empty and not line:
+            continue
+        if remove_comments and line.startswith('#'):
+            continue
+        yield line
+
+
+class CompressionDecider1:
+    def __init__(self, compression, compression_files):
+        """
+        Initialize a CompressionDecider instance (and read config files, if needed).
+
+        :param compression: default CompressionSpec (e.g. from --compression option)
+        :param compression_files: list of compression config files (e.g. from --compression-from) or
+                                  a list of other line iterators
+        """
+        self.compression = compression
+        if not compression_files:
+            self.matcher = None
+        else:
+            self.matcher = PatternMatcher(fallback=compression)
+            for file in compression_files:
+                try:
+                    for line in clean_lines(file):
+                        try:
+                            compr_spec, fn_pattern = line.split(':', 1)
+                        except:
+                            continue
+                        self.matcher.add([parse_pattern(fn_pattern)], CompressionSpec(compr_spec))
+                finally:
+                    if hasattr(file, 'close'):
+                        file.close()
+
+    def decide(self, path):
+        if self.matcher is not None:
+            return self.matcher.match(path)
+        return self.compression
+
+
+class CompressionDecider2:
+    def __init__(self, compression):
+        self.compression = compression
+
+    def decide(self, chunk):
+        # nothing fancy here yet: we either use what the metadata says or the default
+        # later, we can decide based on the chunk data also.
+        # if we compress the data here to decide, we can even update the chunk data
+        # and modify the metadata as desired.
+        compr_spec = chunk.meta.get('compress', self.compression)
+        compr_args = dict(buffer=COMPR_BUFFER)
+        compr_args.update(compr_spec)
+        return compr_args, chunk
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -15,13 +15,14 @@ import sys
 import time
 from io import BytesIO
 from . import xattr
-from .compress import Compressor, COMPR_BUFFER
+from .compress import COMPR_BUFFER
 from .constants import *  # NOQA
 from .helpers import Chunk, Error, uid2user, user2uid, gid2group, group2gid, \
     parse_timestamp, to_localtime, format_time, format_timedelta, safe_encode, safe_decode, \
     Manifest, Statistics, decode_dict, make_path_safe, StableDict, int_to_bigint, bigint_to_int, bin_to_hex, \
     ProgressIndicatorPercent, ChunkIteratorFileWrapper, remove_surrogates, log_multi, \
-    PathPrefixPattern, FnmatchPattern, open_item, file_status, format_file_size, consume
+    PathPrefixPattern, FnmatchPattern, open_item, file_status, format_file_size, consume, \
+    CompressionDecider1, CompressionDecider2, CompressionSpec
 from .repository import Repository
 from .platform import acl_get, acl_set
 from .chunker import Chunker
@@ -125,7 +126,7 @@ class Archive:
 
     def __init__(self, repository, key, manifest, name, cache=None, create=False,
                  checkpoint_interval=300, numeric_owner=False, progress=False,
-                 chunker_params=CHUNKER_PARAMS, start=None, end=None):
+                 chunker_params=CHUNKER_PARAMS, start=None, end=None, compression=None, compression_files=None):
         self.cwd = os.getcwd()
         self.key = key
         self.repository = repository
@@ -148,6 +149,9 @@ class Archive:
         if create:
             self.items_buffer = CacheChunkBuffer(self.cache, self.key, self.stats)
             self.chunker = Chunker(self.key.chunk_seed, *chunker_params)
+            self.compression_decider1 = CompressionDecider1(compression or CompressionSpec('none'),
+                                                            compression_files or [])
+            key.compression_decider2 = CompressionDecider2(compression or CompressionSpec('none'))
             if name in manifest.archives:
                 raise self.AlreadyExists(name)
             self.last_checkpoint = time.time()
@@ -592,11 +596,15 @@ Number of files: {0.stats.nfiles}'''.format(
         }
         # Only chunkify the file if needed
         if chunks is None:
+            compress = self.compression_decider1.decide(path)
+            logger.debug('%s -> compression %s', path, compress['name'])
             fh = Archive._open_rb(path)
             with os.fdopen(fh, 'rb') as fd:
                 chunks = []
                 for data in self.chunker.chunkify(fd, fh):
-                    chunks.append(cache.add_chunk(self.key.id_hash(data), Chunk(data), self.stats))
+                    chunks.append(cache.add_chunk(self.key.id_hash(data),
+                                                  Chunk(data, compress=compress),
+                                                  self.stats))
                     if self.show_progress:
                         self.stats.show_progress(item=item, dt=0.2)
             cache.memorize_file(path_hash, st, [c.id for c in chunks])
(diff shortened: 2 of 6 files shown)


borg/compress.pyx
```diff
@@ -178,7 +178,7 @@ class Compressor:
     compresses using a compressor with given name and parameters
     decompresses everything we can handle (autodetect)
     """
-    def __init__(self, name='null', **kwargs):
```
ThomasWaldmann: forgotten in a rename 'null' -> 'none' (CNONE.name == 'none').

borg/testsuite/compress.py
```diff
@@ -25,7 +25,7 @@ def test_get_compressor():
         get_compressor(name='foobar')
 
 
-def test_cnull():
+def test_cnone():
```
ThomasWaldmann: forgotten to rename null -> none.

borg/helpers.py
```diff
@@ -1256,3 +1256,20 @@ def format_time(self, key, item):
 
     def time(self, key, item):
         return safe_timestamp(item.get(key) or item[b'mtime'])
+
+
+COMPRESS_NONE = set("""
+.gz .tgz .bz2 .tbz2 .xz .lzma .lzo .zip .rar .7z
+.mp3 .ogg .oga .flac
+.mp4 .mkv .avi .mpg .mpeg .webm .vob .ts .ogv
+.jpg .jpeg .png .gif
```
ThomasWaldmann: likely rather incomplete, any popular uncompressible stuff missing?

ThomasWaldmann:
there's no way yet to switch the magic off - do we need one?

infectormp:
Would be nice to have option to skip compression for specific path. I have webapp that store attachment in own format, without extension, like this ```-B6oHe96W-G8yJk6ktNPLpCvqSjRNGKA```. So i will happy define path where files will not compressed. 

and sure would be nice have option where we can define own file extension.

borg/helpers.py
```diff
@@ -40,6 +40,13 @@
 EXIT_WARNING = 1  # reached normal end of operation, but there were issues
 EXIT_ERROR = 2  # terminated abruptly, did not reach end of operation
 
+# meta dict, binary data
+Chunk = namedtuple('Chunk', 'meta data')
+
+
+def mkchunk(data, **meta):
+    return Chunk(meta, data)
```
enkore: Hm, an extra function to directly wrap `Chunk.__new__`?

enkore:
A different implementation idea would be to always run LZ4 over every chunk (it's very fast after all), and see how well LZ4 compresses it (just as an indicator of compressibility). If LZ4 doesn't compress a chunk at all or only by 2 % or so, then it's unlikely that zlib or LZMA could do anything about it either.

This would be similar to what compressing file systems do afaik.

Ape:
Using the file extension helps a bit, but it's really not that flexible since we cannot list all the possible extensions, and sometimes the files do not have a detectable extension at all.

Also, it might be possible that there are files with the listed file name extension, but which are still readily compressible. This may be a weird, rare case, but who know what kind of file naming schemes people are using out there.

borg/helpers.py
ThomasWaldmann: it's to enable Chunk(data, compression="zlib", sparse=False) style.

ThomasWaldmann:
@ape i think the biggest issue are files without extension, like some apps generate. the other cases seem rather uncommon.

enkore:
LZ4/Speed: I am seeing 2 to 4 GB/s compression speed for fast mode on a single core. (`lz4 -b some screen shot` (PNG, incompressible, 1.5 MB)).

API-wise interesting: If we give LZ4_compress_limitedOutput a smaller buffer than the input data it will abort early. E.g. we set maxOutputSize to, say, `0.8 * input size` and it should abort early if it can't compress better than 80 %.

LZMA: From the tests I've made so far with the various compressions they all end up in a "similar" ball park. Depending on test data LZ4 gives me 25-50 % ratio, zlib somewhere like 25-35 % and LZMA basically the same as zlib, maybe 20-35 %.

More importantly: When zlib/LZMA give better results, LZ4 does as well, if zlib/LZMA is worse, LZ4 is worse as well. They seem to scale well relative to each other.

Which might be related to the chunk size, I think a compression like LZMA doesn't really work well with these small chunks (<2 MB) of data. So far there were few cases were LZMA,6 and LZMA,9 actually made a tangible difference. For me it's ~10 times slower than zlib,9 and often makes little difference, sometimes even slightly larger output.

---

Gist of how I think of "perfect hole repro": SEEK_DATA/SEEK_HOLE to map data/holes structure to [(False, length), (True, length) etc] put that in `item[b'sparse']`. When writing, consume `item[b'sparse']`, skip `(False, length)`, creating holes. Not the most efficient way (holes will compress very well), but backwards compatible to 1.0 (when extracting)

borg/helpers.py
verygreen: .mov
.m4v and .m4a
.flv (flash videos)
.aac
.ogm
.dmg (compressed disk images from mac at least), I am sure there are others like this on other platforms.
.txz (I think this is used to shorten .tar.xz)
many popular cameras raw image format are compressed, e.g. canon's .CRW

But overall I suspect it's indeed better to gauge compressibility by some other means as mentioned below. - this will allow to catch all those cases where you cannot know from the extension:
no extension
.exe - might be executable or an sfx archive
.pdf - could be compressed or not
.iso .img - same thing
.tiff - could be uncompressed or compressed (some other image formats too)

Combining both approaches might not be a bad idea too to save on cpu:
i.e. (extensible) blacklist for nocompress, (extensible) whitelist for expensive compress like lzma (for text files and other highly compressible data), autodetect with selected compression for the rest. Also ability to specify different compressions for different path components (lzma,9 for /etc, none for ~/Pictures and ~/Video, zlib for the rest of ~ as an example)

FelixSchwarz: I'd also like to exclude rpm files as they are already compressed (likely the same is true for .deb - maybe also .msi?)

FelixSchwarz:
It seems to me that the file extension might be a good indicator for most users but there might be a significant minority which needs more control (e.g. compression based on path names) so it might be nice if borg could read the compression settings from a file if instructed to do so (while defaulting to a rough extension-based mechanism).

borg/helpers.py
enkore: Optional file for compression schemes:
```
lzma,9:<pattern>
zlib,3:<pattern>
```
With the usual support for patterns, which would then easily allow **both** to add extensions, and to add specific paths, with semantics that we already use elsewhere.

The syntax for this compression mode would then be, e.g. `--compression auto,[<number>,]<fallback>`, with `<fallback>` being any normal compression method, that is used on files that don't match any defined pattern.

We can then use LZ4 on a chunk-per-chunk basis to decide whether LZ4 or the `<fallback>` is used. (As I argued below, LZ4 is extremely fast and running it over all chunks that should receive a stronger, much slower compression algorithm will make no tangible difference. Some Googling in the mean time also turned up that LZ4 and some other algorithms are specifically enhanced to be very fast for incompressible inputs).

The optional `<number>` is the threshold used on the LZ4 ratio (default e.g. 75 %).

**Q:** Where to store the file, well-known position (e.g. ~/.config/borg/auto-compression), or repo-dependent / user-supplied
**Q:** Threshold default value?

enkore:
Generally speaking, maybe split this up in one PR that we can do ~pretty much w/o big discussions (the pass-metadata-along-part), then discuss approaches / implementations for flexible/automatic compression elsewhere?

ThomasWaldmann:
@enkore see #934

verygreen:
I think another good idea is to check compressor output and if the result is bigger (or equal) than the original data - discard compressed output and store data as plain-text. Mark it not just as plaintext, but invent a new compressor id "uncompressed-uncompressible" - this is needed for later when people would start recompressing their repositories with more powerful algorithms, to save cpu time on data that just plain does not compress.

ThomasWaldmann:
First commit is already merged, second needs rebasing onto master.

borg/helpers.py
enkore: Above comment takes on two different issues.

1. Making hardcoded list costomizable
2. Automatic compression selection

-- lets keep it simple and build that later if needed.

ZoomRmc:
Regarding files that do not need compression, [Duplicati2](https://github.com/duplicati/duplicati)have a txt file which is read from application directory with quite a comprehensive list (but not perfect: lacks .opus, for example, .tiff files can be uncompressed ) of file extensions: [default_compressed_extensions.txt ](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/default_compressed_extensions.txt)

ThomasWaldmann:
@ZoomRmc ah, interesting - thanks for the pointer.

ZoomRmc:
May be some entropy measurement can be used to predict how well the file/chunk can be compressed? I just naively ran [some Rust implementation](https://github.com/jme/shannon-rust) of Shannon Entropy measurement on a test set and got the following results:
```
# shannon -f test.log

104857600 bytes
shannon entropy:  5.191260284047
Done in: 89.617991 msec
--------------------
# shannon -f testset.wav

91110528 bytes
shannon entropy:  7.614343072426
Done in: 73.439318 msec
--------------------
# shannon -f testset.flac

53651888 bytes
shannon entropy:  7.990729784876
Done in: 42.863125 msec
--------------------
# shannon -f testset.xz

76732004 bytes
shannon entropy:  7.999779254371
Done in: 61.242032 msec

# shannon -f e:\test.jpg

45127926 bytes
shannon entropy:  7.941891449844
Done in: 36.107949  msec
--------------------
# shannon -f nevergonnagiveyouup.mp4

60735907 bytes
shannon entropy:  7.997137165205
Done in: 47.598294  msec
--------------------
# shannon -f nixos-graphical-16.03.581.e409886-x86_64-linux.iso

1052770304 bytes
shannon entropy:  7.993876963475
Done in: 7.579552695 sec
--------------------
```
Entropy is in 1 to 8 range, time measured *including* reading file from HDD. This is fast, as far as I can tell, but I don't know if it's a viable metric. The same .iso from set compressed to .xz with 0.938 rate.

enkore:
For a test like that you really want to vmlock the test files to make sure no cache-trashing program (e.g. Firefox, Thunderbid, IntelliJ) kicks it out.

For reference, on my half-decade old AMD processor LZ4 processes about 2 GB/s of incompressible data.

ThomasWaldmann:
Idea:

Implement this as `--compression-config=FILENAME` (or `--compression-from-file=...`?) as an alternative to `--compression` option.

In the given file, it would have a list like `<compression-args>:<path/filename-pattern>` (like currently hardcoded).

The list would either end with a pattern like "lz4:*" or it would use 'none' as internal fallback (like when not giving `--compression`, then it also uses 'none' by default).

ThomasWaldmann:
In a future PR, compression modes (currently: `none`, `lz4`, `gzip,N`, `lzma,N`) could be amended by a mode `auto,X` and that would mean that a predictor (lz4?) is used to automatically decide about whether to compress a chunk or not. If it decides yes, compression X would be applied.

This future change could be easily integrated with this one, as `auto,X` would be just adding another choice to compression modes.

Ape:
Is there any benefit for using the manually specified file path pattern whitelist / blacklist if `auto,X` works?

ThomasWaldmann:
Yes, not wasting CPU cycles, if you already can say from the path / filename.

enkore:
There are two independent parts here. See https://github.com/borgbackup/borg/pull/810#discussion_r59601501 and the follow-up.

ZoomRmc:
@ThomasWaldmann, that's a great idea. We possibly could even have unique settings for different kinds of data. It would be great if that compression config file was sorted by compression method and was easily editable. Something like this, maybe (sorry for lua-inspired syntax):
```lua
lz4 = {*.vmdk, *.tar, ...}
lzma = {*.txt, *.log, ...}
auto = {
    lzma={*.tif, *.raw, *.dng},
    zlib={*.iso, *.dd},
    ....
  }
none = { *.jpg, *.mp3, *.zip ...}
```
Do I understand correctly, that compression is performed per chunk, and each chunk has a metadata with a mark of a file-type it came from?

Also, there's a paper from IBM researchers specifically on compressibility prediction: [PDF](https://www.usenix.org/system/files/conference/fast13/fast13-final38.pdf), [Slides](https://www.usenix.org/sites/default/files/conference/protected-files/harnik_fast13_slides.pdf). They got pretty good and reliable results. There's a part about real-time analyzing chunks of data, we can utilize some of the ideas.

[Relevant SO question](http://stackoverflow.com/questions/7027022/how-to-efficiently-predict-if-data-is-compressible)

ThomasWaldmann:
@ZoomRmc yes, different compression for different file types is the goal. But I'ld rather re-use existing pattern definition formats we already use elsewhere.

Compression is performed per chunk (and at that time, the filename where it came from is known). But we do not store the file name or type with the chunk. But we store per-file metadata that has a list of content chunk IDs.

Thanks for the links!

dragetd:
Excuse the off-comment, but I'd like to share my humble opinion: I'd prefer a heuristic (see the BTRFS-suggestion on the other chunking issue or a simple idea) or some build-in defaults to more options that blow up documentation and usage-complexity.

ThomasWaldmann:
@dragetd see #1006.

denji:
btrfs/zstd: https://github.com/terrelln/linux/commits/btrfs-zstd

Issue #797, opened 2016-03-25
Interleave modifications, additions, and removals in 'borg diff'


Ape:
I thought about this when implementing #783. The current version of `borg diff` always shows first the file modifications, then file additions, and finally removals. For me it would be easier to follow the list of changes if all of the different change types would be interleaved obeying the archive order.

See the example below. Note that the real-life change lists may be a lot longer with lots of changed in one folder.
```
old:
+234.5 kB -319.7 kB folder1/changed_file
+234.5 kB -319.7 kB folder2/changed_file
+234.5 kB -319.7 kB folder3/changed_file
added     455.61 kB folder1/new_file
added     455.61 kB folder2/new_file
added     455.61 kB folder3/new_file
removed     2.71 kB folder1/removed_file
removed     2.71 kB folder2/removed_file
removed     2.71 kB folder3/removed_file
```

```
new:
+234.5 kB -319.7 kB folder1/changed_file
added     455.61 kB folder1/new_file
removed     2.71 kB folder1/removed_file
+234.5 kB -319.7 kB folder2/changed_file
added     455.61 kB folder2/new_file
removed     2.71 kB folder2/removed_file
+234.5 kB -319.7 kB folder3/changed_file
added     455.61 kB folder3/new_file
removed     2.71 kB folder3/removed_file
```

#783 should be merged before implementing this to avoid conflicts.

ThomasWaldmann:
the problem here might be that borg has no "directory entry list" data in the archive, so while processing items, you do not know in advance what items are coming for a specific directory. Nor can you easily know when you have finished one directory.

there is a hack in the code that assumes a specific traversal method at create time and then detects the directory level change at extract time. this was one of the things breaking after the multithreading changes.

so I'ld prefer not interleaving rather than adding more such hacks that just cause troubles.

Ape:
Perhaps we could process all of the items and only after that start producing output. The diff command runs pretty quickly on my machine even with over 60 GB of data in the snapshots. This feature could be optional with a command line switch.

enkore:
You can work with the assumption that paths are in-order and grouped by parent directory. The approach would look something like this:

```
for item1, item2 in zip(item iters):
  if path1 > path2:
    while path1 > path2:
      output(path1 deleted in path2)
      advance_only_iter1
  # analogous for path1 < path 2
  if path1 == path2:
    # detailed compare
```

[...]

PR #816, opened 2016-03-31
Add '--sort' option for sorting diff command output

Ape:
This is a solution to #797. The commit is quite trivial and should be easy to review.

Previously, on `borg diff`, the output always had first the modifications, then additions, and finally removals. Output may be easier to follow if the various kinds of changes are interleaved. This commit is a simple solution that first collects the output lines and sorts them by file path before printing. This new behavior is optional and disabled by default. It can be enabled with `--sort` command line option.

This option will be especially useful after the planned multi-threading changes arrive. Multi-threading may shuffle the archive order of files making diff output hard to follow without sorting.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -514,7 +514,7 @@ class Archiver:
             if item1[b'mode'] != item2[b'mode']:
                 return '[{} -> {}]'.format(get_mode(item1), get_mode(item2))
 
-        def compare_items(path, item1, item2, hardlink_masters, deleted=False):
+        def compare_items(output, path, item1, item2, hardlink_masters, deleted=False):
             """
             Compare two items with identical paths.
             :param deleted: Whether one of the items has been deleted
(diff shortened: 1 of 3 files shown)


enkore:
If you add a small test case for sorted output you can stop CodeCov from complaining (and get a test case for sorted output for free ;-)

borg/archiver.py
```diff
@@ -544,43 +544,56 @@ def compare_items(path, item1, item2, hardlink_masters, deleted=False):
 
             changes = [x for x in changes if x]
             if changes:
-                print("{:<19} {}".format(' '.join(changes), remove_surrogates(path)))
+                item = (remove_surrogates(path), ''.join(changes))
```
ThomasWaldmann: above, you used a blank to join the changes, now you use an empty string?

borg/archiver.py
```diff
@@ -544,43 +544,56 @@ def compare_items(path, item1, item2, hardlink_masters, deleted=False):
...
                     b'chunks': [],
                 }, added, hardlink_masters, deleted=True)
             for deleted in orphans_archive1.values():
-                compare_items(deleted[b'path'], deleted, {
+                compare_items(items, deleted[b'path'], deleted, {
                     b'deleted': True,
                     b'chunks': [],
                 }, hardlink_masters, deleted=True)
 
+            for item in sorted(items, key=lambda x: x[0]):
```
ThomasWaldmann: is key=... necessary? a tuple sorts by first, second, ... element anyway.

ThomasWaldmann: also, is a "if args.sort" missing here? if it is not sorted, you already printed the output and you don't want to print it again, sorted.

borg/archiver.py
```diff
@@ -513,7 +513,7 @@ def compare_mode(item1, item2):
             if item1[b'mode'] != item2[b'mode']:
                 return '[{} -> {}]'.format(get_mode(item1), get_mode(item2))
 
-        def compare_items(path, item1, item2, hardlink_masters, deleted=False):
+        def compare_items(items, path, item1, item2, hardlink_masters, deleted=False):
```
ThomasWaldmann: about names:

one would somehow expect, that items are a list of ... items - and item1 is one item and item2 is another item. but in fact, items has a different type of elements - it is a list of output data, so maybe rather call it output?

borg/archiver.py
enkore: If args.sort is False, "items" / the list of output lines isn't populated, they are printed directly (lines 549ff)

Ape: Well `items` is empty if `args.sort` was false.

Ape:
@ThomasWaldmann I addressed your comments in this new version.

ThomasWaldmann:
I mentioned that there is a "if args.sort" missing. Well, you ignored that correctly - it is not needed as output list will be empty, so the code there won't do anything in that case.

ThomasWaldmann:
OK, good change. :) Can you write a little test for it?

Ape:
@ThomasWaldmann Added a test.

PR #820, opened 2016-03-31
fix incorrect _ns variables

ipha:
Copy/paste error, shouldn't have used the _ns variables here.


ThomasWaldmann:
Thanks for fixing so quickly! :)

PR #833, opened 2016-04-03
ArchiveChecker: move "orphaned objects check skipped" to INFO log level

### LINKED ISSUE #826: Add --error to only output log messages of error or higher
Back in 2015, the following was added, but refactored out prior to the 1.0.0 release:

https://github.com/borgbackup/borg/commit/25140e8c8218fe19ef9ffc8029e3f53516b2d546#diff-dfaa777fd33eac4e20768ac811e52696R649

That gave the ability to specify the log level for all levels (debug, info, warning, error, critical).  The current default seems to be to only output "warning" or higher, but there is no way to tell borg that you only want to see "error" or higher.

Adding "--error" would allow me to suppress warning messages in the output.

enkore:
Fixes #826

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -879,7 +879,7 @@ class ArchiveChecker:
                 for id_ in unused:
                     self.repository.delete(id_)
         else:
-            logger.warning('Orphaned objects check skipped (needs all archives checked).')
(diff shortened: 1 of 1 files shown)


enkore:
> borgbackup:master

not what i selected, dear github

PR #838, opened 2016-04-03
Remote: don't print tracebacks for Error exceptions handled downstream

enkore:
(GH wanted to put this into master before)

### DIFF
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -93,8 +93,12 @@ class RepositoryServer:  # pragma: no cover
                             f = getattr(self.repository, method)
                         res = f(*args)
                     except BaseException as e:
-                        logging.exception('Borg %s: exception in RPC call:', __version__)
-                        logging.error(sysinfo())
+                        # These exceptions are reconstructed on the client end in RemoteRepository.call_many(),
+                        # and will be handled just like locally raised exceptions. Suppress the remote traceback
+                        # for these, except ErrorWithTraceback, which should always display a traceback.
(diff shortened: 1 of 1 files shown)


borg/remote.py
```diff
@@ -93,8 +94,11 @@ def serve(self):
                             f = getattr(self.repository, method)
                         res = f(*args)
                     except BaseException as e:
-                        logging.exception('Borg %s: exception in RPC call:', __version__)
-                        logging.error(sysinfo())
+                        if not isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, Repository.CheckNeeded,
+                                              IntegrityError, PathNotAllowed, Repository.ObjectNotFound,
+                                              InvalidRPCMethod)):
```
ThomasWaldmann: are these the ones subclassing Error (not ErrorWithTraceback)?

ThomasWaldmann: and if so, could we just check for that?

enkore: No, these are the special-cased exceptions handled [here](https://github.com/enkore/borg/blob/issue/792/borg/remote.py#L234). These are the only Errors that the error-handling code in archiver.py:main will see.

enkore: I'll toss a note in about the relation between these two code blocks

enkore:
I updated this after some more testing and removed some of the exceptions handled by the client, namely all deriving from ErrorWithTraceback, who should (also) print a (remote) traceback.

PR #839, opened 2016-04-03
If BORG_PASSPHRASE is present but wrong, don't prompt for password, f…

### LINKED ISSUE #791: How to check correctness of passphrase in a script?
I'm working on a shellscript that is run interactively which asks for the repo passphrase and then invokes borg multiple times. Is there a simple way to check in a script that the passphrase is correct?

My idea at first was to set `BORG_PASSPHRASE` and then run a simple command (`borg list`) and check the exit code, but unfortunately borg queries the tty for a passphrase if the one in the environment does not match. Right now I'm using _expect_ to get around this, but this is quite hackish -- is there a simpler method?

enkore:
…ail instead.

Leaves PassphraseKey alone, since I cannot test it.

Fixes #791

### DIFF
--- a/borg/key.py
+++ b/borg/key.py
@@ -18,6 +18,10 @@ import msgpack
 PREFIX = b'\0' * 8
 
 
+class PassphraseWrong(Error):
+    """passphrase supplied in BORG_PASSPHRASE is incorrect"""
+
+
 class PasswordRetriesExceeded(Error):
     """exceeded the maximum password retries"""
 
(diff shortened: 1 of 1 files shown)


borg/key.py
```diff
@@ -284,13 +288,19 @@ def detect(cls, repository, manifest_data):
...
+                for retry in range(0, 3):
+                    passphrase = Passphrase.getpass(prompt)
+                    if key.load(target, passphrase):
+                        break
+                else:
+                    raise PasswordRetriesExceeded
         else:
-            raise PasswordRetriesExceeded
+            if not key.load(target, passphrase):
+                raise PasswordWrong
```
ThomasWaldmann: "Password" here vs. "Passphrase" elsewhere.

ThomasWaldmann: hmm, nothing new though.

enkore: I'll make it Passphrase here. Renaming existing output in a bugfix rel. is maybe not a good idea, so we'll do that in master later?

Ape:
The commit subject line generally has a hard limit of 69 characters, but you should aim for 50.

Like for example:
- Fail if BORG_PASSPHRASE is present but wrong
- Do not prompt if BORG_PASSPHRASE is present but wrong

Issue #842, opened 2016-04-05
add a comment to an archive


ThomasWaldmann:
see: https://github.com/jborg/attic/issues/147

Would need 2 things:
- a --comment option to borg create (to immediately specify a comment)
- a borg comment command to add a comment to an archive later

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/32643831-add-a-comment-to-an-archive)

Ape:
We cannot add the comment field to archive metadata since it would affect the hash. Instead we could add a new directory `comments` to the repository, containing a file for each archive. The files could be named with the archive hash and the content would just be the comment text.

A more extensible approach would be to similarly add a new folder but call it `mutables` (we need a better name) and then store a data structure that allows adding more mutable fields besides the comments. Perhaps one of the fields would be a checksum hash for the file itself.

Any comments on this?

enkore:
If "--comment" is part of "create" than it would kinda make sense to have it in "recreate" to add/change it:

@Ape Mutamble/immutable comment, which one do we want?

Ape:
Okay, I agree. The comment field is very similar to the name.

ThomasWaldmann:
@enkore with borg rename, you can change the archive name, so from the user perspective, it is mutable.

enkore:
... changing the archive ID/fingerprint in the process.

infectormp:
@enkore changing the archive ID/fingerprint is not good idea.

Ape:
@infectormp Actually, I think this kind of immutable data is a very good thing. Compare it to how Git stores its data. Things can still be modified, but under the hood everything is clean and immutable (except the thin mutable layer referencing all that immutable data).

enkore:
Yep, that's the point I was making. You can change it, just like e.g. "git rebase" or soon "borg recreate" allows to change committed / archived things. But the commit / archive ID changes, so you can see that it isn't the *same thing* anymore. I think this is highly desirable.

If I wanted to, I could write my archive IDs down, and even when I didn't encrypt them, nobody could change these archives without me noticing -- either you get a different archive ID, or an IntegrityError.

infectormp:
@enkore with your comments i not have any questions more.

PR #847, opened 2016-04-06
Add archive comments

Ape:
This is an implementation for #842. Tests and better documentation will follow.

- Archives now have a new metadata field `comment`
- `borg info` command shows the comment if it's present
- `borg create` command now has option `--comment` for adding comments to archives.
- A new command `borg comment` is added for modifying the comments on existing archives.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -447,17 +447,22 @@ Number of files: {0.stats.nfiles}'''.format(
             except OSError:
                 pass
 
-    def rename(self, name):
-        if name in self.manifest.archives:
-            raise self.AlreadyExists(name)
+    def set_meta(self, key, value):
         metadata = StableDict(self._load_meta(self.id))
-        metadata[b'name'] = name
+        metadata[key] = value
         data = msgpack.packb(metadata, unicode_errors='surrogateescape')
         new_id = self.key.id_hash(data)
         self.cache.add_chunk(new_id, data, self.stats)
-        self.manifest.archives[name] = {'id': new_id, 'time': metadata[b'time']}
+        self.manifest.archives[self.name] = {'id': new_id, 'time': metadata[b'time']}
         self.cache.chunk_decref(self.id, self.stats)
-        del self.manifest.archives[self.name]
+
+    def rename(self, name):
+        if name in self.manifest.archives:
+            raise self.AlreadyExists(name)
+        oldname = self.name
+        self.name = name
+        self.set_meta(b'name', name)
+        del self.manifest.archives[oldname]
 
     def delete(self, stats, progress=False):
         unpacker = msgpack.Unpacker(use_list=False)
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -180,7 +180,7 @@ class Archive:
     def load(self, id):
         self.id = id
         self.metadata = self._load_meta(self.id)
-        decode_dict(self.metadata, (b'name', b'hostname', b'username', b'time', b'time_end'))
+        decode_dict(self.metadata, (b'name', b'comment', b'hostname', b'username', b'time', b'time_end'))
         self.metadata[b'cmdline'] = [arg.decode('utf-8', 'surrogateescape') for arg in self.metadata[b'cmdline']]
         self.name = self.metadata[b'name']
 
@@ -240,7 +240,7 @@ Number of files: {0.stats.nfiles}'''.format(
         del self.manifest.archives[self.checkpoint_name]
         self.cache.chunk_decref(self.id, self.stats)
 
-    def save(self, name=None, timestamp=None):
+    def save(self, name=None, comment=None, timestamp=None):
         name = name or self.name
         if name in self.manifest.archives:
             raise self.AlreadyExists(name)
@@ -256,6 +256,7 @@ Number of files: {0.stats.nfiles}'''.format(
         metadata = StableDict({
             'version': 1,
             'name': name,
+            'comment': comment,
             'items': self.items_buffer.chunks,
             'cmdline': sys.argv,
             'hostname': socket.gethostname(),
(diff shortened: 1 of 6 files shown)


infectormp:
What size limit for this field?

enkore:
Kinda arbitrary. Will be loaded fully into memory. Won't be chunkified (like metadata[b'items], which is quite large). You probably don't want to stuff more than a couple MB in there.

Ape:
@infectormp  I tried how it behaves in practice. It seems to be working with long comments (around 10~100 kB) with various characters including line feeds. After that the issues I get is `zsh: argument list too long: borg`. There are ways to go around that, but I don't recommend using longer comments anyway.

infectormp:
@Ape thank you.

borg/archiver.py
```diff
@@ -625,6 +625,16 @@ def do_rename(self, args, repository, manifest, key, cache, archive):
         return self.exit_code
 
     @with_repository(exclusive=True, cache=True)
+    @with_archive
+    def do_comment(self, args, repository, manifest, key, cache, archive):
+        """Set the archive comment"""
+        archive.set_comment(args.comment)
```
ThomasWaldmann: is this the only call to set_comment?

ThomasWaldmann:
Please add a test for borg comment command.

borg/archiver.py
Ape: @ThomasWaldmann Yes.

Ape:
I added tests and usage examples.

borg/archive.py
```diff
@@ -180,7 +180,7 @@ def _load_meta(self, id):
     def load(self, id):
         self.id = id
         self.metadata = self._load_meta(self.id)
-        decode_dict(self.metadata, (b'name', b'hostname', b'username', b'time', b'time_end'))
+        decode_dict(self.metadata, (b'name', b'comment', b'hostname', b'username', b'time', b'time_end'))
```
ThomasWaldmann: btw, did you locate all the places that have such a tuple and need extension?

borg/archiver.py
ThomasWaldmann: if it is the only call, it could be inlined.

borg/archiver.py
```diff
@@ -731,6 +741,10 @@ def format_cmdline(cmdline):
         stats = archive.calc_stats(cache)
         print('Name:', archive.name)
         print('Fingerprint: %s' % hexlify(archive.id).decode('ascii'))
+
+        if archive.metadata.get(b'comment') is not None:
+            print('Comment:', archive.metadata[b'comment'])
```
ThomasWaldmann: we could also use a default of '' and just always print it.
for old archives without comment feature or new archives with no comment, it would just print "Comment: " then.

docs/usage.rst
```diff
@@ -335,6 +335,26 @@ Examples
     newname                              Mon, 2016-02-15 19:50:19
 
 
+.. include:: usage/comment.rst.inc
+
+Examples
+~~~~~~~~
+::
+
+    $ borg create /mnt/backup::archivename ~ --comment "This is a comment"
```
ThomasWaldmann: most examples have the options before the repo::archive spec.

borg/archive.py
enkore: "Tagging" for some (future) cleanup: Make a METADATA_KEYS (or something like that) as we have one for the item keys (ITEM_KEYS), so that we can de-duplicate all these snippets (=> don't repeat yourself).

ThomasWaldmann: good point. can you create a ticket about this. or maybe even a general one like "put all constants into one place and deduplicate them"?

enkore: -> #856, a checklist should do

borg/archiver.py
enkore: I think it's better to have that implementation detail in the Archive class, like rename (<-- both share a good bit of code, maybe deduplicate that?).

borg/archive.py
Ape: @ThomasWaldmann I'm pretty sure that I updated all the places with this tuple.

borg/archiver.py
Ape: The current implementation has the issue that while the comment can be modified it cannot be removed (set to None). `''` as the default would fix that. However, I think it's better to not show the `Comment:` line when the comment is not set (or is empty in this case).

borg/archive.py
```diff
@@ -459,6 +460,15 @@ def rename(self, name):
...
 
+    def set_comment(self, comment):
+        metadata = StableDict(self._load_meta(self.id))
+        metadata[b'comment'] = comment
+        data = msgpack.packb(metadata, unicode_errors='surrogateescape')
+        new_id = self.key.id_hash(data)
+        self.cache.add_chunk(new_id, data, self.stats)
+        self.manifest.archives[self.name] = {'id': new_id, 'time': metadata[b'time']}
+        self.cache.chunk_decref(self.id, self.stats)
+
```
ThomasWaldmann: yeah, set_comment and rename should be deduplicated.

the common parts could move into a set_meta(key, value) function that is used by rename().
set_comment would be replaced by set_meta which would be called from archiver then.

please do an additional changeset for this and only collapse it after review.

Ape:
@ThomasWaldmann I made the discussed modifications, and pushed the unsquashed branch here for review. Please wait for squashing before actually merging.

borg/archiver.py
```diff
@@ -728,6 +738,10 @@ def do_info(self, args, repository, manifest, key, archive, cache):
         stats = archive.calc_stats(cache)
         print('Name:', archive.name)
         print('Fingerprint: %s' % hexlify(archive.id).decode('ascii'))
+
+        if archive.metadata.get(b'comment'):
+            print('Comment:', archive.metadata[b'comment'])
```
ThomasWaldmann: i'm still thinking whether it is easier if we always print the Comment: entry, even if empty or not present (and default to "" if not present). That might be easier if someone parses this and would not leave questions open whether this borg version can show comments.

enkore: Consistent output is easier to parse, +1

Ape:
@ThomasWaldmann I agree, it's better to show the comment field always. I pushed this change here. I also squashed the pull request for merging. The unsquashed version can be seen on [Ape/borg/archive_comments](https://github.com/Ape/borg/commits/archive_comments). The only change since your last review is 3d3aa94131de93f69efd9ee359e4004988aefd9c.

ThomasWaldmann:
OK, thanks for working on this!

PR #848, opened 2016-04-06
Escape cmdline arguments that have spaces in them

Ape:
`borg info` command shows the command line that was used to create the archive. If the command line has arguments with spaces in them the printed command will not work. This commit will escape these arguments by enclosing it to quotation marks. This is a minor thing.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -725,6 +725,9 @@ class Archiver:
     @with_archive
     def do_info(self, args, repository, manifest, key, archive, cache):
         """Show archive details such as disk space used"""
+        def format_cmdline(cmdline):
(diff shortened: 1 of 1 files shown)


borg/archiver.py
```diff
@@ -725,6 +725,15 @@ def write(bytestring):
     @with_archive
     def do_info(self, args, repository, manifest, key, archive, cache):
         """Show archive details such as disk space used"""
+        def format_argument(argument):
+            if ' ' in argument:
+                return '"{}"'.format(argument)
+            else:
+                return argument
```
enkore: [`shlex.quote()`](https://docs.python.org/3/library/shlex.html#shlex.quote) should be more "solid"

Ape:
@enkore Thanks for the suggestion. I pushed a new version.

PR #858, opened 2016-04-08
Auto-recover from corrupted index/hints file(s)

enkore:
> And don't swallow all OSErrors when creating archives. We need to work on
> that on a more general level....

[One approach](/enkore/borg/tree/map_internal_oserror) (which I don't particularly like. I think it'd be better to carefully inspect all FS operations and add appropriate error handling - but this could also be a lot of additional code). --- I'll open another ticket for discussion about that issue.

### DIFF
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -5,6 +5,8 @@ import os
 
 cimport cython
 from libc.stdint cimport uint32_t, UINT32_MAX, uint64_t
+from libc.errno cimport errno
+from cpython.exc cimport PyErr_SetFromErrnoWithFilename
 
 API_VERSION = 2
 
@@ -26,14 +28,6 @@ cdef extern from "_hashindex.c":
     int hashindex_set(HashIndex *index, void *key, void *value)
     uint32_t _htole32(uint32_t v)
     uint32_t _le32toh(uint32_t v)
-
-
-cdef extern from "errno.h":
-    int errno
-
-
-cdef extern from "string.h":
-    char *strerror(int errnum)
 
 
 cdef _NoDefault = object()
@@ -62,10 +56,6 @@ MAX_VALUE = _MAX_VALUE
 assert _MAX_VALUE % 2 == 1
 
 
-def decoded_strerror(errno):
-    return strerror(errno).decode(locale.getpreferredencoding(), 'surrogateescape')
-
-
 @cython.internal
 cdef class IndexBase:
     cdef HashIndex *index
@@ -78,7 +68,8 @@ cdef class IndexBase:
             self.index = hashindex_read(path)
             if not self.index:
                 if errno:
-                    raise OSError(errno, decoded_strerror(errno), os.fsdecode(path))
+                    PyErr_SetFromErrnoWithFilename(OSError, path)
+                    return
                 raise RuntimeError('hashindex_read failed')
         else:
             self.index = hashindex_init(capacity, self.key_size, self.value_size)
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -63,7 +63,7 @@ cdef class IndexBase:
             path = os.fsencode(path)
             self.index = hashindex_read(path)
             if not self.index:
-                raise Exception('hashindex_read failed')
+                raise RuntimeError('hashindex_read failed')
         else:
             self.index = hashindex_init(capacity, self.key_size, self.value_size)
             if not self.index:
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from collections import namedtuple
+import locale
 import os
 
 cimport cython
@@ -60,6 +61,11 @@ MAX_VALUE = _MAX_VALUE
 
 assert _MAX_VALUE % 2 == 1
 
+
+def decoded_strerror(errno):
+    return strerror(errno).decode(locale.getpreferredencoding(), 'surrogateescape')
+
+
 @cython.internal
 cdef class IndexBase:
     cdef HashIndex *index
@@ -72,7 +78,7 @@ cdef class IndexBase:
             self.index = hashindex_read(path)
             if not self.index:
                 if errno:
-                    raise OSError(errno, strerror(errno), path)
+                    raise OSError(errno, decoded_strerror(errno), os.fsdecode(path))
                 raise RuntimeError('hashindex_read failed')
         else:
             self.index = hashindex_init(capacity, self.key_size, self.value_size)
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -27,6 +27,14 @@ cdef extern from "_hashindex.c":
     uint32_t _le32toh(uint32_t v)
 
 
+cdef extern from "errno.h":
+    int errno
+
+
+cdef extern from "string.h":
+    char *strerror(int errnum)
+
+
 cdef _NoDefault = object()
 
 """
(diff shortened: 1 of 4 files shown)


enkore:
Note to self: *never* rely on chmod for tests.

borg/repository.py
```diff
@@ -188,10 +191,23 @@ def commit(self, save_space=False):
         self.write_index()
         self.rollback()
 
-    def open_index(self, transaction_id):
+    def open_index(self, transaction_id, auto_recover=True):
         if transaction_id is None:
             return NSIndex()
-        return NSIndex.read((os.path.join(self.path, 'index.%d') % transaction_id).encode('utf-8'))
+        index_path = (os.path.join(self.path, 'index.%d') % transaction_id).encode('utf-8')
```
ThomasWaldmann: if you move "% transaction_id" directly right to 'index.%d', you don't need the outer parens.

borg/repository.py
```diff
@@ -188,10 +191,23 @@ def commit(self, save_space=False):
...
+        try:
+            return NSIndex.read(index_path)
+        except RuntimeError as re:
+            assert str(re) == 'hashindex_read failed'  # everything else means we're in *deep* trouble
```
ThomasWaldmann: could this be a permissions issue?

enkore: Good catch. In that case the later unlink() will explode, but that isn't correctly converted to InternalOSError yet, so would be silenced in an upper layer; not good.

borg/repository.py
```diff
@@ -204,15 +220,31 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
+                with open(os.path.join(self.path, 'hints.%d' % transaction_id), 'rb') as fd:
+                    hints = msgpack.unpack(fd)
+            except (msgpack.UnpackException, msgpack.ExtraData, FileNotFoundError) as e:
+                # corrupted or deleted hints file, need to replay segments
+                if not isinstance(e, FileNotFoundError):
+                    os.unlink(os.path.join(self.path, 'hints.%d' % transaction_id))
+                # index must exist at this point
+                os.unlink(os.path.join(self.path, 'index.%d' % transaction_id))
+                self.check_transaction()
+                self.prepare_txn(transaction_id)
```
ThomasWaldmann: could this become an infinite recursion if the problem can not be resolved this way?

enkore: Hm, if there is some reappearing error here, like unreliable storage, then this would recurse until the stack pops. I suppose this is a valid approach, if the underlying system is *that* unreliable...

borg/testsuite/repository.py
```diff
@@ -221,6 +221,53 @@ def segments_in_repository():
...
+        self.repository.commit()
+        self.repository.close()
+
+    def do_commit(self):
+        with self.repository:
+            self.repository.put(b'00000000000000000000000000000000', b'fox')
+            self.repository.commit()
+
+    def test_corrupted_hints(self):
+        with open(os.path.join(self.repository.path, 'hints.0'), 'ab') as fp:
```
ThomasWaldmann: do we use fp at other places? reminds me of file pointer, like in C, but in python, it is just a file (object), so f reads more naturally.

enkore: Looks like 'fd' is usually used in this code base, but 'f' also makes some appearances.

borg/repository.py
```diff
@@ -12,7 +12,7 @@
 from zlib import crc32
 
 import msgpack
-from .helpers import Error, ErrorWithTraceback, IntegrityError, Location, ProgressIndicatorPercent
+from .helpers import Error, ErrorWithTraceback, IntegrityError, Location, ProgressIndicatorPercent, InternalOSError
```
ThomasWaldmann: you could put it after IntegrityError.

borg/repository.py
```diff
@@ -188,10 +191,28 @@ def commit(self, save_space=False):
...
+        except RuntimeError as re:
+            assert str(re) == 'hashindex_read failed'  # everything else means we're in *deep* trouble
+            # corrupted index file, need to replay segments
+            try:
+                os.unlink(os.path.join(self.path, 'hints.%d' % transaction_id))
+                os.unlink(os.path.join(self.path, 'index.%d' % transaction_id))
+            except OSError as e:
+                raise InternalOSError from e
+            if not auto_recover:
+                raise
```
ThomasWaldmann: aren't the unlink commands above already part of the auto recovery? if so, move these 2 lines upwards to above the unlinks?

borg/repository.py
```diff
@@ -204,15 +225,31 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
         if transaction_id is None:
             self.segments = {}  # XXX bad name: usage_count_of_segment_x = self.segments[x]
             self.compact = set()  # XXX bad name: segments_needing_compaction = self.compact
         else:
             if do_cleanup:
                 self.io.cleanup(transaction_id)
-            with open(os.path.join(self.path, 'hints.%d' % transaction_id), 'rb') as fd:
-                hints = msgpack.unpack(fd)
+            try:
+                with open(os.path.join(self.path, 'hints.%d' % transaction_id), 'rb') as fd:
```
ThomasWaldmann: maybe compute hints_path = ... to avoid the duplicated o.p.j? also compute index_path = ... for regularity even if not duplicated.

ThomasWaldmann:
aside from the few cosmetic issues, is this ready for merge?

enkore:
I want to do some more "field-testing" of this before merging. I'll fix the cosmetics and rebase it (just to be sure).

borg/repository.py
```diff
@@ -188,10 +191,28 @@ def commit(self, save_space=False):
...
-    def open_index(self, transaction_id):
+    def open_index(self, transaction_id, auto_recover=True):
         if transaction_id is None:
             return NSIndex()
-        return NSIndex.read((os.path.join(self.path, 'index.%d') % transaction_id).encode('utf-8'))
+        index_path = os.path.join(self.path, 'index.%d' % transaction_id).encode('utf-8')
+        try:
+            return NSIndex.read(index_path)
+        except RuntimeError as re:
+            assert str(re) == 'hashindex_read failed'  # everything else means we're in *deep* trouble
```
ThomasWaldmann: i see why you used "re", but e or err might be less strange.

borg/hashindex.pyx
```diff
@@ -63,7 +71,9 @@ cdef class IndexBase:
             path = os.fsencode(path)
             self.index = hashindex_read(path)
             if not self.index:
-                raise Exception('hashindex_read failed')
+                if errno:
+                    raise OSError(errno, strerror(errno).decode(), os.fsdecode(path))
```
enkore: Hm, I'll need to check whether "strerror(errno).decode()" is correct... probably either locale-dependent encoding or the FS encoding.

enkore:
merge?

borg/hashindex.pyx
```diff
@@ -63,7 +77,9 @@ cdef class IndexBase:
             path = os.fsencode(path)
             self.index = hashindex_read(path)
             if not self.index:
-                raise Exception('hashindex_read failed')
+                if errno:
+                    raise OSError(errno, decoded_strerror(errno), os.fsdecode(path))
```
ThomasWaldmann: hmm, is this going to get us into trouble if there is some crap in path or errno msg, that does not output to the console?

ThomasWaldmann:
looks good, but didn't practically test it. did you?

enkore:
I tested this practically last time I made changes, before a rebase iirc; better retest.

borg/hashindex.pyx
enkore: fsdecode uses surrogateescape (on *nix); shouldn't explode. strerror on the other hand has to be in lc_messages locale....

Issue #824, opened 2016-04-01
add --progress switch to borg check


ThomasWaldmann:
people run borg check from cron script and do not want the "Checking segments ..." output in the log (from borgbackup ML).

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/32498620-add-progress-switch-to-borg-check)

ThomasWaldmann:
Problem: for remote repositories, passing progress=True/False to repo.check() would need a rpc change.

enkore:
faheem (IRC) sparked the idea (kind of) that we could just push the ProgressIndicatorPercent over the normal logging infrastructure and have it use --info level or below. Would not break RPC.

ThomasWaldmann:
@edgewood is this fixed?

edgewood:
Fixed by #865.  The reference must have gotten lost in the squash.

rom1v:
Could it be possible to enable `--progress` by default for `borg check` (and maybe other commands)?

Executing `borg check --verify-data` and getting absolutely no feedback is very confusing. (does it work? is it frozen?)

ThomasWaldmann:
Please no discussions in issues that were closed 5y ago. If you want to file an issue, file a new one.

PR #865, opened 2016-04-09
Print --list output (only) without requiring --info

edgewood:
There are persistent questions why --list output doesn't show up, and borg currently isn't able to show *just* the list output.  The solution is to use a more granular child logger, so that list output only goes to the borg.archiver.list child logger.  That logger can then be configured separately from the parent loggers.

The .list child logger can also be used as a hook in a BORG_LOGGING_CONF config file to log the list output to a separate file (this is my main motivation), or send it to a network socket where some daemon could analyze it.

What I did:
- add an accessor to LazyLogger to call the real logger's getChild()
- log --list output using child logger .list
- if --list is passed, change child logger .list (only) to log at INFO level

This is a proof of concept for adding similar code for all of the options that imply that they produce output, but don't if you don't add --info.  I'll make similar changes for those options, create tests, and update the documentation if this approach is acceptable.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -8,6 +8,7 @@ import functools
 import hashlib
 import inspect
 import io
+import logging
 import os
 import re
 import shlex
@@ -111,7 +112,7 @@ class Archiver:
 
     def print_file_status(self, status, path):
         if self.output_list and (self.output_filter is None or status in self.output_filter):
-            logger.info("%1s %s", status, remove_surrogates(path))
+            logging.getLogger('borg.output.list').info("%1s %s", status, remove_surrogates(path))
 
     @staticmethod
     def compare_chunk_contents(chunks1, chunks2):
@@ -277,7 +278,7 @@ class Archiver:
                               DASHES,
                               str(archive.stats),
                               str(cache),
-                              DASHES)
+                              DASHES, logger=logging.getLogger('borg.output.stats'))
 
         self.output_filter = args.output_filter
         self.output_list = args.output_list
@@ -416,7 +417,7 @@ class Archiver:
                 while dirs and not item[b'path'].startswith(dirs[-1][b'path']):
                     archive.extract_item(dirs.pop(-1), stdout=stdout)
             if output_list:
-                logger.info(remove_surrogates(orig_path))
+                logging.getLogger('borg.output.list').info(remove_surrogates(orig_path))
             try:
                 if dry_run:
                     archive.extract_item(item, dry_run=True)
@@ -673,7 +674,7 @@ class Archiver:
                     log_multi(DASHES,
                               stats.summary.format(label='Deleted data:', stats=stats),
                               str(cache),
-                              DASHES)
+                              DASHES, logger=logging.getLogger('borg.output.stats'))
         else:
             if not args.cache_only:
                 msg = []
@@ -820,18 +821,19 @@ class Archiver:
         to_delete = (set(archives) | checkpoints) - (set(keep) | set(keep_checkpoints))
         stats = Statistics()
         with Cache(repository, key, manifest, do_files=args.cache_files, lock_wait=self.lock_wait) as cache:
+            list_logger = logging.getLogger('borg.output.list')
             for archive in archives_checkpoints:
                 if archive in to_delete:
                     if args.dry_run:
                         if args.output_list:
-                            logger.info('Would prune:     %s' % format_archive(archive))
+                            list_logger.info('Would prune:     %s' % format_archive(archive))
                     else:
                         if args.output_list:
-                            logger.info('Pruning archive: %s' % format_archive(archive))
+                            list_logger.info('Pruning archive: %s' % format_archive(archive))
                         Archive(repository, key, manifest, archive.name, cache).delete(stats)
                 else:
                     if args.output_list:
-                        logger.info('Keeping archive: %s' % format_archive(archive))
+                        list_logger.info('Keeping archive: %s' % format_archive(archive))
             if to_delete and not args.dry_run:
                 manifest.write()
                 repository.commit(save_space=args.save_space)
@@ -840,7 +842,7 @@ class Archiver:
                 log_multi(DASHES,
                           stats.summary.format(label='Deleted data:', stats=stats),
                           str(cache),
-                          DASHES)
+                          DASHES, logger=logging.getLogger('borg.output.stats'))
         return self.exit_code
 
     def do_upgrade(self, args):
@@ -1299,6 +1301,9 @@ class Archiver:
                                help='only check last N archives (Default: all)')
         subparser.add_argument('-P', '--prefix', dest='prefix', type=str,
                                help='only consider archive names starting with this prefix')
+        subparser.add_argument('-p', '--progress', dest='progress',
+                               action='store_true', default=False,
+                               help="""show progress display while checking""")
 
         change_passphrase_epilog = textwrap.dedent("""
         The key files used for repository encryption are optionally passphrase
@@ -2065,12 +2070,27 @@ class Archiver:
         check_extension_modules()
         selftest(logger)
 
+    def _setup_implied_logging(self, args):
+        """ turn on INFO level logging for args that imply that they will produce output """
+        # map of option name to name of logger for that option
+        option_logger = {
+                'output_list': 'borg.output.list',
+                'show_version': 'borg.output.show-version',
+                'show_rc': 'borg.output.show-rc',
+                'stats': 'borg.output.stats',
+                'progress': 'borg.output.progress',
+                }
+        for option, logger_name in option_logger.items():
+            if args.get(option, False):
+                logging.getLogger(logger_name).setLevel('INFO')
+
     def run(self, args):
         os.umask(args.umask)  # early, before opening files
         self.lock_wait = args.lock_wait
         setup_logging(level=args.log_level, is_serve=args.func == self.do_serve)  # do not use loggers before this!
+        self._setup_implied_logging(vars(args))
         if args.show_version:
-            logger.info('borgbackup version %s' % __version__)
+            logging.getLogger('borg.output.show-version').info('borgbackup version %s' % __version__)
         self.prerun_checks(logger)
         if is_slow_msgpack():
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
@@ -2142,15 +2162,16 @@ def main():  # pragma: no cover
     if msg:
         logger.error(msg)
     if args.show_rc:
+        rc_logger = logging.getLogger('borg.output.show-rc')
         exit_msg = 'terminating with %s status, rc %d'
         if exit_code == EXIT_SUCCESS:
-            logger.info(exit_msg % ('success', exit_code))
+            rc_logger.info(exit_msg % ('success', exit_code))
         elif exit_code == EXIT_WARNING:
-            logger.warning(exit_msg % ('warning', exit_code))
+            rc_logger.warning(exit_msg % ('warning', exit_code))
         elif exit_code == EXIT_ERROR:
-            logger.error(exit_msg % ('error', exit_code))
+            rc_logger.error(exit_msg % ('error', exit_code))
         else:
-            logger.error(exit_msg % ('abnormal', exit_code or 666))
+            rc_logger.error(exit_msg % ('abnormal', exit_code or 666))
     sys.exit(exit_code)
 
 
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -1037,7 +1037,7 @@ def yes(msg=None, false_msg=None, true_msg=None, default_msg=None,
 
 
 class ProgressIndicatorPercent:
-    def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=None):
+    def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%"):
         """
         Percentage-based progress indicator
 
@@ -1046,17 +1046,33 @@ class ProgressIndicatorPercent:
         :param start: at which percent value to start
         :param same_line: if True, emit output always on same line
         :param msg: output message, must contain one %f placeholder for the percentage
-        :param file: output file, default: sys.stderr
         """
         self.counter = 0  # 0 .. (total-1)
         self.total = total
         self.trigger_at = start  # output next percentage value when reaching (at least) this
         self.step = step
-        if file is None:
-            file = sys.stderr
-        self.file = file
         self.msg = msg
         self.same_line = same_line
+        self.handler = None
+        self.logger = logging.getLogger('borg.output.progress')
+
+        # If there are no handlers, set one up explicitly because the
+        # terminator and propagation needs to be set.  If there are,
+        # they must have been set up by BORG_LOGGING_CONF: skip setup.
+        if not self.logger.handlers:
+            self.handler = logging.StreamHandler(stream=sys.stderr)
+            self.handler.setLevel(logging.INFO)
+            self.handler.terminator = '\r' if self.same_line else '\n'
+
+            self.logger.addHandler(self.handler)
+            if self.logger.level == logging.NOTSET:
+                self.logger.setLevel(logging.WARN)
+            self.logger.propagate = False
+
+    def __del__(self):
+        if self.handler is not None:
+            self.logger.removeHandler(self.handler)
+            self.handler.close()
 
     def progress(self, current=None):
         if current is not None:
@@ -1073,11 +1089,11 @@ class ProgressIndicatorPercent:
             return self.output(pct)
 
     def output(self, percent):
-        print(self.msg % percent, file=self.file, end='\r' if self.same_line else '\n', flush=True)
+        self.logger.info(self.msg % percent)
 
     def finish(self):
         if self.same_line:
-            print(" " * len(self.msg % 100.0), file=self.file, end='\r')
+            self.logger.info(" " * len(self.msg % 100.0))
 
 
 class ProgressIndicatorEndless:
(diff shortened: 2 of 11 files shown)


edgewood:
Of course I forgot to rerun tox after rebasing on master.  Testing fix...

borg/archiver.py
```diff
@@ -1720,6 +1720,8 @@ def run(self, args):
         setup_logging(level=args.log_level, is_serve=args.func == self.do_serve)  # do not use loggers before this!
         if args.show_version:
             logger.info('borgbackup version %s' % __version__)
+        if 'output_list' in args and args.output_list:
```
ThomasWaldmann: that is a bit strange:
"in" checks for a dict member(?), but then you access an attribute?

ThomasWaldmann:
The whole thing will get more complicated when you consider remote logging (when borg works via ssh). Typically this will be the case if the log output comes from repository code (in case of RemoteRepository).

borg/archiver.py
enkore: Yeah, that's a "speciality" of argparse.Namespace

enkore:
Is it likely that something "list-like" that works on the Repository level will be added?

ThomasWaldmann:
@enkore this is not just about --list, it has to be thought through for all (log) output.

borg/archiver.py
edgewood: Yeah, as the commit message said, it was a quick hack because I only had a few minutes to fix the testcase breaks caused by rebasing.  It looks from the argparse docs that 'vars(args)' is the preferred way to treat an argparse Namespace object like a dict.  I'll update that tonight or tomorrow.

-- 
Ed Blackman

edgewood:
Well, the point of child loggers is that if you log messages to a given child logger, you can change the level on that child without changing output logged to different (parent or sibling) loggers.

That said, I don't want to break things, so if you have any use cases or scenarios that I should test, tell me what they are and I certainly will.

-- 
Ed Blackman

enkore:
The test failure looks like it was caused by removing -v but not updating ArchiveRecreater to use the new child logger

edgewood:
Recreate must output some info level messages that are *NOT* list output, so removing -v removed those messages.  I'm testing a revert to that part of the change, and creating an explicit recreate --list test.  I know it would be better to have explicit tests for all the commands that create --list output, but archiver.py tests are already so slow it seemed wasteful to add explicit tests when I could piggyback on existing tests, and went too far.

borg/testsuite/archiver.py
```diff
@@ -1288,13 +1317,13 @@ def _test_recreate_interrupt(self, change_args, interrupt_early):
...
-            self.cmd('recreate', '-sv', '--list', self.repository_location, 'input/dir2')
+            self.cmd('recreate', '-s', '--list', self.repository_location, 'input/dir2')
         assert 'test.recreate' in self.cmd('list', self.repository_location)
         if change_args:
             with patch.object(sys, 'argv', sys.argv + ['non-forking tests don\'t use sys.argv']):
-                output = self.cmd('recreate', '-sv', '--list', '-pC', 'lz4', self.repository_location, 'input/dir2')
+                output = self.cmd('recreate', '-s', '--list', '-pC', 'lz4', self.repository_location, 'input/dir2')
         else:
-            output = self.cmd('recreate', '-sv', '--list', self.repository_location, 'input/dir2')
+            output = self.cmd('recreate', '-s', '--list', self.repository_location, 'input/dir2')
```
enkore: this one needs reverting

borg/testsuite/archiver.py
```diff
@@ -1288,13 +1317,13 @@ def _test_recreate_interrupt(self, change_args, interrupt_early):
         if interrupt_early:
             process_files = 0
         with patch.object(ArchiveRecreater, 'process_item', self._recreate_interrupt_patch(process_files)):
-            self.cmd('recreate', '-sv', '--list', self.repository_location, 'input/dir2')
+            self.cmd('recreate', '-s', '--list', self.repository_location, 'input/dir2')
         assert 'test.recreate' in self.cmd('list', self.repository_location)
         if change_args:
             with patch.object(sys, 'argv', sys.argv + ['non-forking tests don\'t use sys.argv']):
-                output = self.cmd('recreate', '-sv', '--list', '-pC', 'lz4', self.repository_location, 'input/dir2')
+                output = self.cmd('recreate', '-s', '--list', '-pC', 'lz4', self.repository_location, 'input/dir2')
```
enkore: and this too

borg/testsuite/archiver.py
```diff
@@ -1288,13 +1317,13 @@ def _test_recreate_interrupt(self, change_args, interrupt_early):
         if interrupt_early:
             process_files = 0
         with patch.object(ArchiveRecreater, 'process_item', self._recreate_interrupt_patch(process_files)):
-            self.cmd('recreate', '-sv', '--list', self.repository_location, 'input/dir2')
+            self.cmd('recreate', '-s', '--list', self.repository_location, 'input/dir2')
```
enkore: this one is ok (output isn't used). In fact, you can remove '-s', '--list' entirely.

enkore:
> Recreate must output some info level messages that are NOT list output, so removing -v removed those messages. I'm testing a revert to that part of the change, and creating an explicit recreate --list test.

You could do that at lines 1320 (the last line comment I made), since the output there is not yet used.

edgewood:
I'm going to need some guidance on something used by RemoteRepository.  Unless I'm looking at things wrong, it looks like show-version and show-rc are strictly before/after dealing with the repository, and progress is all to sys.stderr so needs to come off my list.  stats looks similar to list.

Can you point me in the right direction?

ThomasWaldmann:
@edgewood iirc there is some "progress" stuff for some remote repo funcs - you could try to apply your ideas there.

edgewood:
It looks like progress type messages from the remote goes through RemoteRepository.call_many (currently lines 296-302) where data on stderr either has a '$LOG' prefix and goes into the logging system, or gets prefixed on stderr with "Remote:".

Experimentation shows that --progress messages go through the latter path, but the only thing I can get to go through the former path is borg serve stack traces/sysinfo on error, not anything that I'm talking about changing.

So I'm going to go ahead and make the changes for the remaining checklist items, above.  Please let me know if I've missed something and I'll take another look.

edgewood:
Just a quick ping to say I'm still working on this as I have time. I have the code for show-rc, show-version, and stats added. I'm working on the tests.

edgewood:
Code and tests are complete, please review.  Assuming it passes tox (it does locally, but Travis seems to be slow), I plan to squash since there were some failing commits. When it passes review, I'll update the docs.

ThomasWaldmann:
hmm, still nothing from a remote logger. if you introduce a new logging mechanism, the PoC has to show that it also works in the intended way for remote.

E.g. the message "Starting repository check" in repository.py might be emitted in local process (if one uses a locally mounted repo) or in the remote borg process. That message isn't enabled by some special cmdline option, but just imagine it were - how would your code work then?

That applies to whatever logging we want to add in repository.py that would use your new way.

edgewood:
Thanks for the pointer to a particular message.  That really helped narrow down you concern.

There's a link in the commit that explains why this change shouldn't affect anything other than the child logger messages.  Let me know if you would like a breakdown, I know I stared at that flow chart for a couple hours before I really understood what was going on.

ThomasWaldmann:
@edgewood my concern is not that it affects something there, but whether it can be used there.

edgewood:
Right now the remote sends the log level and the message, and the local logs the message at the given level on the root logger. If a remote logs a message on a child logger and it's emitted, the local will log it on the root logger at the given level.  With current defaults, if child loggers were added to dual remote/local code, a message that would have been emitted if the execution were local would not be emitted when executed remotely.

So before child loggers could be used in dual remote/local code, the remote->local log translation layer would have to be enhanced to also pass the logger name on the remote, and log the message on that logger locally.

I'd be willing to open a separate PR to implement that enhancement, but since all of the implied output options are local-only code, I don't think that should block this one.

edgewood:
I had a thought overnight about an easy way to change the remote/local logging adapter to allow child loggers in dual remote/local code.  I didn't realize that it would take me 10 times as long to write the test (mainly due to difficulty doing arbitrary things in the remote)!  Look at the notes in the "Update logging parser" commit for more detail.

enkore:
I'll review this later, but I have to say that your solution sounds very clean and doesn't look like it would make any notable compatibility problems.

edgewood:
Rebased on master and fixed flake8 problems.  Review appreciated.

edgewood:
I saw the comment @enkore made on #824 and realized that the infrastructure I already set up would make the rest of the changes pretty easy.  I'm not super thriilled with "--no-progress", but wanted to wait on feedback.  One possibility might be http://stackoverflow.com/a/9236426, or changing it to --quiet instead, or just making quiet checking the default in 1.1.  Feedback appreciated.

borg/archiver.py
```diff
@@ -1213,6 +1216,8 @@ def build_parser(self, args=None, prog=None):
                                help='only check last N archives (Default: all)')
         subparser.add_argument('-P', '--prefix', dest='prefix', type=str,
                                help='only consider archive names starting with this prefix')
+        subparser.add_argument('--no-progress', dest='no_progress', action='store_false',
```
ThomasWaldmann: I'ld prefer to not having --no-whatever options.

Or, said without double negation: I prefer --whatever options. :)

edgewood: Yeah, I don't like it either. But the current behavior is to print progress, so the choice was to either change the default, or come up with some option name that was descriptive. --silent/--quiet didn't seem right because the option is just to shut off progress, not -v output.  -v --no-progress is a reasonable combination of options, and -v --silent/--quiet seems silly.

Suggestions?
-- 
Ed Blackman


-------- Original Message --------

enkore:
Changing default log output between 1.0 and 1.1 seems ok to me.

borg/archiver.py
enkore: Changing default log output between 1.0 and 1.1 seems ok to me.

edgewood:
What remains after I fix --no-progress for check to be --progress (and thereby change the default check to not emit progress information), and update the doc so that all of the implied output options no longer reference requiring --info/-v?  I would really like some review of the overall change.  I'm happy to answer questions about the approach, etc.

Of course I'll rebase for the next chunk of work, and squash at the end.  Or I could squash now if that would make review easier.

ThomasWaldmann:
rather squash at the very end.

borg/remote.py
```diff
@@ -301,7 +301,13 @@ def fetch_from_cache(args):
                         if line.startswith('$LOG '):
                             _, level, msg = line.split(' ', 2)
                             level = getattr(logging, level, logging.CRITICAL)  # str -> int
-                            logging.log(level, msg.rstrip())
+                            if msg[0:7] == 'Remote:':
```
ThomasWaldmann: .startswith()

borg/archiver.py
```diff
@@ -1919,12 +1925,26 @@ def parse_args(self, args=None):
...
         return args
 
+    def _setup_implied_logging(self, args):
+        """ turn on INFO level logging for args that imply that they will produce output """
+        # map of option name to name of child logger (which matches the option name)
+        option_logger = {
+                'output_list': 'list',
+                'show_version': 'show-version',
+                'show_rc': 'show-rc',
+                'stats': 'stats',
```
ThomasWaldmann: after --(no-)progress is changed, guess it will need to be added here, too.

borg/helpers.py
```diff
@@ -1040,17 +1040,33 @@ def __init__(self, total, step=5, start=0, same_line=False, msg="%3.0f%%", file=
...
+        self.handler = None
+        self.logger = logger.getChild('progress')
+
+        # If there are no handlers, set one up explicitly because the
+        # terminator and propagation needs to be set.  If there are,
+        # they must have been set up by BORG_LOGGING_CONF: skip setup.
+        if not self.logger.handlers:
+            self.handler = logging.StreamHandler(stream=sys.stderr)
+            self.handler.setLevel(logging.INFO)
+            self.handler.terminator = '\r' if self.same_line else '\n'
```
ThomasWaldmann: btw, did you actually try it same line works with the logging system?

there could be (no-)flush-related issues...

edgewood: I did test.  StreamHandler isn't documented to flush after every message, but that's what the implementation does.  One of the tests added earlier looks for the carriage return in stderr after the progress message was sent, but I modified it to send two progress messages between checks.  I can't think of another way to test that behavior in case StreamHandler changes behavior.

I could subclass StreamHandler and add a flush at the end of .emit() after calling super().emit() (which would currently double flush), but that seems excessive to me.

borg/remote.py
edgewood: Fixed.

borg/archiver.py
edgewood: Yes.  I actually had to rework how the alternate logger was selected, since that made using child loggers of the current logger unworkable.  See the d163621 commit message for more detail.

borg/helpers.py
enkore: I think that's okay, "logging" isn't going to change any time soon. And: not flushing after messages doesn't make sense for logging (loses messages when stuff crashes).

edgewood:
So is there anything other than doc updates left for me to do?

borg/archiver.py
```diff
@@ -2033,24 +2032,25 @@ def prerun_checks(self, logger):
...
         """ turn on INFO level logging for args that imply that they will produce output """
-        # map of option name to name of child logger (which matches the option name)
+        # map of option name to name of logger for that option
         option_logger = {
-                'output_list': 'list',
-                'show_version': 'show-version',
-                'show_rc': 'show-rc',
-                'stats': 'stats',
+                'output_list': 'borg.output.list',
+                'show_version': 'borg.output.show-version',
```
ThomasWaldmann: hmm, there are no requirements on the logger name? like being a valid python name or so?

just asking because of that "-" in there and because usually module names are used for this.

borg/archiver.py
```diff
@@ -112,7 +113,7 @@ def print_warning(self, msg, *args):
 
     def print_file_status(self, status, path):
         if self.output_list and (self.output_filter is None or status in self.output_filter):
-            logger.info("%1s %s", status, remove_surrogates(path))
+            logger.getLogger('borg.output.list').info("%1s %s", status, remove_surrogates(path))
```
ThomasWaldmann: hmm, the logger object is a way to lazily do logging.getLogger(modulename).
calling this with another .getLogger(absolute name) seems a bit strange.
can this be done more easily?

borg/logger.py
```diff
@@ -154,6 +154,9 @@ def __logger(self):
         def setLevel(self, *args, **kw):
             return self.__logger.setLevel(*args, **kw)
 
+        def getLogger(self, *args, **kw):
+            return logging.getLogger(*args, **kw)
```
ThomasWaldmann: note: does not use "self" - does it belong in here?

ThomasWaldmann:
@edgewood: I put a bounty on #824 - please visit the bounty page. If you think your changes can solve it, indicate that you are working on it.

borg/archiver.py
edgewood: Python doesn't document any requirements, other than dots make hierarchy.  A glance at the implementation doesn't show any requirements either.

I did test, with this config: https://gist.github.com/edgewood/7e3bd5def843d50012c1b4cee5257303

borg/archiver.py
edgewood: I dropped the logger convenience method and just use logging.getLogger() directly in the latest commit.

borg/archiver.py
ThomasWaldmann: ok, thanks for testing.

edgewood:
Anything else before squash?

ThomasWaldmann:
squash it. maybe update the commit comment(s) and keep what's worth keeping.

ThomasWaldmann:
doing some practical tests with the code now...

ThomasWaldmann:
http://borgbackup.readthedocs.io/en/stable/development.html#output-and-logging

can you update this with a few words about how to handle output now?

ThomasWaldmann:
other than that, it seems to work fine - i did some experiments with local and remote repo (both running your latest code). did not do compat tests with older versions.

edgewood:
Squashed most of the commits down to one.  I left the commit that changes the remote logging interface out in case you wanted to cherry-pick it into 1.0-maint.  I'm also not thrilled with the update to the "output and logging section", so left it out to think about it further.

edgewood:
@ThomasWaldmann I revised the "output and logging" section to provide less of a tutorial and more of a pointer. Let me know if it's what you're looking for.

If you want to incorporate the "remotes pass logger name" into 1.0-maint, let me know.  If not I'll squash it, if so I'll open a new PR.

Short summary of cross-version effects of this PR:
Old borg server, new client -> "borg check" still prints progress without --progress flag
New borg server, old client -> remote messages have logger name in the message: "borg.repository Remote: <message>"
New borg server, new client -> full function

Cherry-picking "remotes pass logger name" into 1.0-maint would allow 1.0.3(?) to be treated as "new" in the list above, but with the side effect of creating the same problems between 1.0.3 and 1.0.[012].  I think it's cleaner to leave it out, but wanted to leave the option open.

ThomasWaldmann:
@LocutusOfBorg ^^ any opinion about that, concerning debian/ubuntu package?

ThomasWaldmann:
dev docs are fine like that.

ThomasWaldmann:
If any change related to this shall get into 1.0.3, I need it today. But maybe we don't need to hurry here, 1.1 release might still take a while, so we could also put that into 1.0.4, if needed.

Also, i think we could merge this into 1.1 / master now and put any 1.0.x-changing-or-not related discussion into a separate ticket.

edgewood:
There's definitely no urgency for the remote logger name patch to go in 1.0.3.

I already cherry-picked that patch to 1.0-maint locally, so you could merge this PR into master as is, or I could squash the final two commits tonight.  Both pass the tests on their own, so it's up to you as a question of style/taste.

I'll open a separate PR tonight on the question of whether to pull the remote logger name patch into 1.0.x.

PR #878, opened 2016-04-10
Fix 'borg diff' when symlink is replacing or being replaced

Ape:
Previously on some cases where symlink replaced something else or was replaced by something else, `borg diff` failed with `KeyError: b'source'`. This commit fixes the issue. There are now also tests that catch this issue.

(I remember testing this case when developing the feature, but I guess I dropped it when refactoring the hardlink behavior.)


ThomasWaldmann:
thanks!

PR #885, opened 2016-04-11
extract: --skip-errors ignores corrupted chunks (w/ log message)

### LINKED ISSUE #840: borg extract: check behaviour on integrity (or other) errors
Assuming there is some issue in the repo storage, like partial corruption, unreadable disk sectors or so - how does borg handle them?

It might be the case that borg extract currently aborts when it hits such a file. More helpful would be to log an error for that file and try to continue. At the end, exit with warning status (1). This is similar to what borg does when creating archives and it encounters problems for single files.

enkore:
Fixes #840 

-- I'm open to suggestions how the extraction loop can be formulated in a less ugly way.


enkore:
Does not work correctly for Repository-level corruption for Remotes. Closing.

PR #888, opened 2016-04-11
Port hashindex_summarize into ChunkIndex.summarize

enkore:
Also adds a test that does

    checking whether hashindex summarize is sane... probably

Generated C code for convenience: https://gist.github.com/enkore/4bdf209d5a16a5deba29d4f101dbf8fd

-- the loop is essentially the same. Brief performance testing [shows](https://gist.github.com/enkore/4bdf209d5a16a5deba29d4f101dbf8fd#file-perfcheck) no difference in time/space usage, as one would expect.

#887 is merged into this for master.

---

*Also* changes int -> int32_t in those places where it's a format dependency. Cython seems to use what's "int" to the underlying C compiler, which is not guaranteed to be 4 bytes (C: >= 2 byte).

### DIFF
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 import os
 
+cimport cython
+from libc.stdint cimport int32_t
+
 API_VERSION = 2
 
 
@@ -19,13 +22,12 @@ cdef extern from "_hashindex.c":
     void *hashindex_next_key(HashIndex *index, void *key)
     int hashindex_delete(HashIndex *index, void *key)
     int hashindex_set(HashIndex *index, void *key, void *value)
-    int _htole32(int v)
-    int _le32toh(int v)
+    int _htole32(int32_t v)
+    int _le32toh(int32_t v)
 
 
 cdef _NoDefault = object()
 
-cimport cython
 
 @cython.internal
 cdef class IndexBase:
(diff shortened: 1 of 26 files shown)


borg/hashindex.pyx
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 import os
 
+cimport cython
+from libc.stdint cimport int32_t
```
ThomasWaldmann: hmm, 1.0-maint?

borg/hashindex.pyx
```diff
@@ -191,11 +188,23 @@ cdef class ChunkIndex(IndexBase):
...
+            key = hashindex_next_key(self.index, key)
+            if not key:
+                break
+            unique_chunks += 1
+            values = <int*> (key + self.key_size)
+            chunks += _le32toh(values[0])
+            unique_size += _le32toh(values[1])
+            unique_csize += _le32toh(values[2])
+            size += <long long> _le32toh(values[1]) * _le32toh(values[0])
+            csize += <long long> _le32toh(values[2]) *  _le32toh(values[0])
```
ThomasWaldmann: long long - can that vary in unwanted ways also?

enkore: It's guaranteed to be >=64 bits; since this is only passed to Python we don't need a fixed width.

enkore:
--- backporting to 1.0-maint

Issue #169, opened 2015-08-29
Chunk refcount limited to 2**31 (4e10)


sourcejedi:
We need to 1) check for overflow and 2) raise the limit.

I think this limit is a bit tight, and we need to add a quick check for this overflow :).  (Note index implementation details may interfere a little. I think you're not allowed to use the bit pattern 0xffffffff or oxfffffffe anyway & there is no check there either).

To fix it I think we only need to change the index format.  In theory the index can be rebuilt if your attic doesn't recognize it, so in some sense we'd already have backwards and forwards compatibility.  (I know attic has some repair code, though I haven't tried just deleting the index ad seeing what happens yet :).

@ThomasWaldmann pointed out that my test results indicate very large chunks (8M, the maximum size?) for the zero data.  (Which also sounds suspicious to me, but hey :).  That should mitigate the problem somewhat.  OTOH I believe the chunker is keyed when encryption is used, so the size should be variable.

> To overflow refcount in the chunk index, you need 2GiB * ~64KiB (of zeros or other consistent pattern, again assuming default chunk size).

> There's not a big margin there, and we need to make sure we have an overflow check. You can include a 500G sparse VM image (mostly zero) in the backup. If it never changes then "incremental" backups are plenty fast, but you overflow after 256 daily backups.

See also: https://github.com/jborg/attic/issues/315

ThomasWaldmann:
True about the chunker table being randomly keyed with encryption.

I don't find it suspicious that lots of zeros give 8MB chunks. The window contents of the rolling has will always be all-zero in such a case and the probability that some specific data has a N-bits-of-zero-ending hash value is about 1 : 2^N.

ThomasWaldmann:
Some other ideas:
- if it is int32_t now, we can check about using uint32_t (not that it helps much, but still).
- overflow/range check: yes, we should have that ASAP.
- users could still continue in a fresh, empty repo of we detect/avoid an overflow (and just abort / rollback then).

ThomasWaldmann:
2^31 * 2^23 = 2^54 bytes = 16 EB of zeros could be a problem.
but i guess if you have that much zeros in your files and you even backup them, then you have other bigger problems. :)

verygreen:
Well, sparse files! (since sparse files are not yet handled in a real way by borg, and even once they are, apparently the same problem might still exist depending on the implementation).

[...]

PR #890, opened 2016-04-11
hashindex_add: convert to native byte-order before add, to LE after

enkore:


borg/_hashindex.c
```diff
@@ -473,7 +473,7 @@ hashindex_add(HashIndex *index, const void *key, int32_t *other_values)
     if(my_values == NULL) {
         hashindex_set(index, key, other_values);
     } else {
-        *my_values += *other_values;
+        my_values[0] = _htole32(_le32toh(my_values[0]) + _le32toh(other_values[0]));
```
verygreen: Hm, will this break existing repos created on bigendian platforms before the change? Or is this purely refcount, so should be fine?

enkore: On BE it would add LE numbers (with a BE-add) and interpret the result as LE. In some cases that works, but as soon as a carry crosses a byte-border it breaks. This might even create smaller results, or overflow earlier. If it makes smaller results = refcounts data might have been deleted by delete/prune.

Now this code should work correctly everywhere. For LE these macros are a no-op. For BE the values (LE) are converted to BE, added with BE-add, and converted back to LE.

(None of this resides in the Repository, this is only used for the Cache. Deleting caches -- they are automatically recreated -- will purge any incorrect refcounts that might linger around. Use extract --dry-run on all archives to verify that no data has been lost).

verygreen:
Another question is you are assuming 32 bit integers all over that code, yet there's value_size in the HashIndex struct which seems to imply that other sizes might be (eventually?) possible?

And another oddity is that if you are using this for a refcount, there's no overflow check, what if I have a hash that's used 4G times? This might sound unlikely, but suppose it's some real common chunk like all zeros, so I have tons of it in every backup and over time my backups are adding a bunch more every time.
Even if you do not want to go to 64 bit counters, you might at least protect against the result of addition getting to 0 (that I imagine might force this chunk removal unexpectedly).

Also apparently the change here should come with a changelog warning of "bigendian users, remove .cache upon upgrading to this version" or detect if the cache was generated with an older version and force rebuild.

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/169

enkore:
TW: Since this is similarly only for ChunkIndex, also port this (_add, _merge) to Cython?

And -1 == "that's one popular chunk"

Also, tests.

enkore:
Oh boi. We should really document hashindex better.

```c

#define BUCKET_IS_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == DELETED)
#define BUCKET_IS_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == EMPTY)

#define EMPTY _htole32(0xffffffff)
#define DELETED _htole32(0xfffffffe)
```

the first 4 bytes of values stored into it better not be EMPTY or DELETED...

enkore:
Superseded by a677612 in #891

enkore:
@verygreen it would be great if you could test #891. The test suite should cover it.

PR #891, opened 2016-04-11
Add overflow checks to refcounting, port stuff to Cython etc

enkore:
(This changeset does not include #890, since that one is already for 1.0-maint. It supersedes merged #887, and unmerged #888 for master)

### DIFF
--- a/borg/_hashindex.c
+++ b/borg/_hashindex.c
@@ -441,32 +441,6 @@ hashindex_get_size(HashIndex *index)
 }
 
 static void
-hashindex_summarize(HashIndex *index, long long *total_size, long long *total_csize,
-                    long long *total_unique_size, long long *total_unique_csize,
-                    long long *total_unique_chunks, long long *total_chunks)
-{
-    int64_t size = 0, csize = 0, unique_size = 0, unique_csize = 0, chunks = 0, unique_chunks = 0;
-    const int32_t *values;
-    void *key = NULL;
-
-    while((key = hashindex_next_key(index, key))) {
-        values = key + index->key_size;
-        unique_chunks++;
-        chunks += _le32toh(values[0]);
-        unique_size += _le32toh(values[1]);
-        unique_csize += _le32toh(values[2]);
-        size += (int64_t) _le32toh(values[0]) * _le32toh(values[1]);
-        csize += (int64_t) _le32toh(values[0]) * _le32toh(values[2]);
-    }
-    *total_size = size;
-    *total_csize = csize;
-    *total_unique_size = unique_size;
-    *total_unique_csize = unique_csize;
-    *total_unique_chunks = unique_chunks;
-    *total_chunks = chunks;
-}
-
-static void
 hashindex_add(HashIndex *index, const void *key, int32_t *other_values)
 {
     int32_t *my_values = (int32_t *)hashindex_get(index, key);
--- a/borg/_hashindex.c
+++ b/borg/_hashindex.c
@@ -439,25 +439,3 @@ hashindex_get_size(HashIndex *index)
 {
     return index->num_entries;
 }
-
-static void
-hashindex_add(HashIndex *index, const void *key, int32_t *other_values)
-{
-    int32_t *my_values = (int32_t *)hashindex_get(index, key);
-    if(my_values == NULL) {
-        hashindex_set(index, key, other_values);
-    } else {
-        *my_values += *other_values;
-    }
-}
-
-static void
-hashindex_merge(HashIndex *index, HashIndex *other)
-{
-    int32_t key_size = index->key_size;
-    void *key = NULL;
-
-    while((key = hashindex_next_key(other, key))) {
-        hashindex_add(index, key, key + key_size);
-    }
-}
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -379,21 +379,19 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
     def chunk_incref(self, id, stats):
         if not self.txn_active:
             self.begin_txn()
-        count, size, csize = self.chunks[id]
-        self.chunks[id] = (count + 1, size, csize)
+        count, size, csize = self.chunks.incref(id)
         stats.update(size, csize, False)
         return id, size, csize
 
     def chunk_decref(self, id, stats):
         if not self.txn_active:
             self.begin_txn()
-        count, size, csize = self.chunks[id]
-        if count == 1:
+        count, size, csize = self.chunks.decref(id)
+        if count == 0:
             del self.chunks[id]
             self.repository.delete(id, wait=False)
             stats.update(-size, -csize, True)
         else:
-            self.chunks[id] = (count - 1, size, csize)
             stats.update(-size, -csize, False)
 
     def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -11,9 +11,6 @@ cdef extern from "_hashindex.c":
     HashIndex *hashindex_read(char *path)
     HashIndex *hashindex_init(int capacity, int key_size, int value_size)
     void hashindex_free(HashIndex *index)
-    void hashindex_summarize(HashIndex *index, long long *total_size, long long *total_csize,
-                             long long *unique_size, long long *unique_csize,
-                             long long *total_unique_chunks, long long *total_chunks)
     void hashindex_merge(HashIndex *index, HashIndex *other)
     void hashindex_add(HashIndex *index, void *key, void *value)
     int hashindex_get_size(HashIndex *index)
@@ -191,11 +188,23 @@ cdef class ChunkIndex(IndexBase):
         return iter
 
     def summarize(self):
-        cdef long long total_size, total_csize, unique_size, unique_csize, total_unique_chunks, total_chunks
-        hashindex_summarize(self.index, &total_size, &total_csize,
-                            &unique_size, &unique_csize,
-                            &total_unique_chunks, &total_chunks)
-        return total_size, total_csize, unique_size, unique_csize, total_unique_chunks, total_chunks
+        cdef long long size = 0, csize = 0, unique_size = 0, unique_csize = 0, chunks = 0, unique_chunks = 0
+        cdef int *values
+        cdef void *key = NULL
+
+        while True:
+            key = hashindex_next_key(self.index, key)
+            if not key:
+                break
+            unique_chunks += 1
+            values = <int*> (key + self.key_size)
+            chunks += _le32toh(values[0])
+            unique_size += _le32toh(values[1])
+            unique_csize += _le32toh(values[2])
+            size += <long long> _le32toh(values[1]) * _le32toh(values[0])
+            csize += <long long> _le32toh(values[2]) *  _le32toh(values[0])
+
+        return size, csize, unique_size, unique_csize, unique_chunks, chunks
 
     def add(self, key, refs, size, csize):
         assert len(key) == self.key_size
--- a/borg/hashindex.pyx
+++ b/borg/hashindex.pyx
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 import os
 
+cimport cython
+from libc.stdint cimport uint32_t, UINT32_MAX, uint64_t
+
 API_VERSION = 2
 
 
@@ -19,13 +22,34 @@ cdef extern from "_hashindex.c":
     void *hashindex_next_key(HashIndex *index, void *key)
     int hashindex_delete(HashIndex *index, void *key)
     int hashindex_set(HashIndex *index, void *key, void *value)
-    int _htole32(int v)
-    int _le32toh(int v)
+    uint32_t _htole32(uint32_t v)
+    uint32_t _le32toh(uint32_t v)
 
 
 cdef _NoDefault = object()
 
-cimport cython
+"""
+The HashIndex is *not* a general purpose data structure. The value size must be at least 4 bytes, and these
+first bytes are used for in-band signalling in the data structure itself.
+
+The constant MAX_VALUE defines the valid range for these 4 bytes when interpreted as an uint32_t from 0
+to MAX_VALUE (inclusive). The following reserved values beyond MAX_VALUE are currently in use
+(byte order is LE)::
+
+    0xffffffff marks empty entries in the hashtable
+    0xfffffffe marks deleted entries in the hashtable
+
+None of the publicly available classes in this module will accept nor return a reserved value;
+AssertionError is raised instead.
+"""
+
+assert UINT32_MAX == 2**32-1
+
+# module-level constant because cdef's in classes can't have default values
+cdef uint32_t _MAX_VALUE = 2**32-1025
+MAX_VALUE = _MAX_VALUE
+
+assert _MAX_VALUE % 2 == 1
 
 @cython.internal
 cdef class IndexBase:
@@ -98,22 +122,30 @@ cdef class NSIndex(IndexBase):
 
     def __getitem__(self, key):
         assert len(key) == self.key_size
-        data = <int *>hashindex_get(self.index, <char *>key)
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
         if not data:
-            raise KeyError
-        return _le32toh(data[0]), _le32toh(data[1])
+            raise KeyError(key)
+        cdef uint32_t segment = _le32toh(data[0])
+        assert segment <= _MAX_VALUE, "maximum number of segments reached"
+        return segment, _le32toh(data[1])
 
     def __setitem__(self, key, value):
         assert len(key) == self.key_size
-        cdef int[2] data
-        data[0] = _htole32(value[0])
+        cdef uint32_t[2] data
+        cdef uint32_t segment = value[0]
+        assert segment <= _MAX_VALUE, "maximum number of segments reached"
+        data[0] = _htole32(segment)
         data[1] = _htole32(value[1])
         if not hashindex_set(self.index, <char *>key, data):
             raise Exception('hashindex_set failed')
 
     def __contains__(self, key):
+        cdef uint32_t segment
         assert len(key) == self.key_size
-        data = <int *>hashindex_get(self.index, <char *>key)
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
+        if data != NULL:
+            segment = _le32toh(data[0])
+            assert segment <= _MAX_VALUE, "maximum number of segments reached"
         return data != NULL
 
     def iteritems(self, marker=None):
@@ -146,25 +178,46 @@ cdef class NSKeyIterator:
         self.key = hashindex_next_key(self.index, <char *>self.key)
         if not self.key:
             raise StopIteration
-        cdef int *value = <int *>(self.key + self.key_size)
-        return (<char *>self.key)[:self.key_size], (_le32toh(value[0]), _le32toh(value[1]))
+        cdef uint32_t *value = <uint32_t *>(self.key + self.key_size)
+        cdef uint32_t segment = _le32toh(value[0])
+        assert segment <= _MAX_VALUE, "maximum number of segments reached"
+        return (<char *>self.key)[:self.key_size], (segment, _le32toh(value[1]))
 
 
 cdef class ChunkIndex(IndexBase):
+    """
+    Mapping of 32 byte keys to (refcount, size, csize), which are all 32-bit unsigned.
+
+    The reference count cannot overflow. If an overflow would occur, the refcount
+    is fixed to MAX_VALUE and will neither increase nor decrease by incref(), decref()
+    or add().
+
+    Prior signed 32-bit overflow is handled correctly for most cases: All values
+    from UINT32_MAX (2**32-1, inclusive) to MAX_VALUE (exclusive) are reserved and either
+    cause silent data loss (-1, -2) or will raise an AssertionError when accessed.
+    Other values are handled correctly. Note that previously the refcount could also reach
+    0 by *increasing* it.
+
+    Assigning refcounts in this reserved range is an invalid operation and raises AssertionError.
+    """
 
     value_size = 12
 
     def __getitem__(self, key):
         assert len(key) == self.key_size
-        data = <int *>hashindex_get(self.index, <char *>key)
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
         if not data:
-            raise KeyError
-        return _le32toh(data[0]), _le32toh(data[1]), _le32toh(data[2])
+            raise KeyError(key)
+        cdef uint32_t refcount = _le32toh(data[0])
+        assert refcount <= _MAX_VALUE
+        return refcount, _le32toh(data[1]), _le32toh(data[2])
 
     def __setitem__(self, key, value):
         assert len(key) == self.key_size
-        cdef int[3] data
-        data[0] = _htole32(value[0])
+        cdef uint32_t[3] data
+        cdef uint32_t refcount = value[0]
+        assert refcount <= _MAX_VALUE, "invalid reference count"
+        data[0] = _htole32(refcount)
         data[1] = _htole32(value[1])
         data[2] = _htole32(value[2])
         if not hashindex_set(self.index, <char *>key, data):
@@ -172,8 +225,37 @@ cdef class ChunkIndex(IndexBase):
 
     def __contains__(self, key):
         assert len(key) == self.key_size
-        data = <int *>hashindex_get(self.index, <char *>key)
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
+        if data != NULL:
+            assert data[0] <= _MAX_VALUE
         return data != NULL
+
+    def incref(self, key):
+        """Increase refcount for 'key', return (refcount, size, csize)"""
+        assert len(key) == self.key_size
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
+        if not data:
+            raise KeyError(key)
+        cdef uint32_t refcount = _le32toh(data[0])
+        assert refcount <= _MAX_VALUE, "invalid reference count"
+        if refcount != _MAX_VALUE:
+            refcount += 1
+        data[0] = _htole32(refcount)
+        return refcount, _le32toh(data[1]), _le32toh(data[2])
+
+    def decref(self, key):
+        """Decrease refcount for 'key', return (refcount, size, csize)"""
+        assert len(key) == self.key_size
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
+        if not data:
+            raise KeyError(key)
+        cdef uint32_t refcount = _le32toh(data[0])
+        # Never decrease a reference count of zero
+        assert 0 < refcount <= _MAX_VALUE, "invalid reference count"
+        if refcount != _MAX_VALUE:
+            refcount -= 1
+        data[0] = _htole32(refcount)
+        return refcount, _le32toh(data[1]), _le32toh(data[2])
 
     def iteritems(self, marker=None):
         cdef const void *key
@@ -188,8 +270,9 @@ cdef class ChunkIndex(IndexBase):
         return iter
 
     def summarize(self):
-        cdef long long size = 0, csize = 0, unique_size = 0, unique_csize = 0, chunks = 0, unique_chunks = 0
-        cdef int *values
+        cdef uint64_t size = 0, csize = 0, unique_size = 0, unique_csize = 0, chunks = 0, unique_chunks = 0
+        cdef uint32_t *values
+        cdef uint32_t refcount
         cdef void *key = NULL
 
         while True:
@@ -197,25 +280,46 @@ cdef class ChunkIndex(IndexBase):
             if not key:
                 break
             unique_chunks += 1
-            values = <int*> (key + self.key_size)
-            chunks += _le32toh(values[0])
+            values = <uint32_t*> (key + self.key_size)
+            refcount = _le32toh(values[0])
+            assert refcount <= MAX_VALUE, "invalid reference count"
+            chunks += refcount
             unique_size += _le32toh(values[1])
             unique_csize += _le32toh(values[2])
-            size += <long long> _le32toh(values[1]) * _le32toh(values[0])
-            csize += <long long> _le32toh(values[2]) *  _le32toh(values[0])
+            size += <uint64_t> _le32toh(values[1]) * _le32toh(values[0])
+            csize += <uint64_t> _le32toh(values[2]) * _le32toh(values[0])
 
         return size, csize, unique_size, unique_csize, unique_chunks, chunks
 
     def add(self, key, refs, size, csize):
         assert len(key) == self.key_size
-        cdef int[3] data
+        cdef uint32_t[3] data
         data[0] = _htole32(refs)
         data[1] = _htole32(size)
         data[2] = _htole32(csize)
-        hashindex_add(self.index, <char *>key, data)
+        self._add(<char*> key, data)
+
+    cdef _add(self, void *key, uint32_t *data):
+        cdef uint64_t refcount1, refcount2, result64
+        values = <uint32_t*> hashindex_get(self.index, key)
+        if values:
+            refcount1 = _le32toh(values[0])
+            refcount2 = _le32toh(data[0])
+            assert refcount1 <= _MAX_VALUE
+            assert refcount2 <= _MAX_VALUE
+            result64 = refcount1 + refcount2
+            values[0] = _htole32(min(result64, _MAX_VALUE))
+        else:
+            hashindex_set(self.index, key, data)
 
     def merge(self, ChunkIndex other):
-        hashindex_merge(self.index, other.index)
+        cdef void *key = NULL
+
+        while True:
+            key = hashindex_next_key(other.index, key)
+            if not key:
+                break
+            self._add(key, <uint32_t*> (key + self.key_size))
 
 
 cdef class ChunkKeyIterator:
@@ -235,5 +339,7 @@ cdef class ChunkKeyIterator:
         self.key = hashindex_next_key(self.index, <char *>self.key)
         if not self.key:
             raise StopIteration
-        cdef int *value = <int *>(self.key + self.key_size)
-        return (<char *>self.key)[:self.key_size], (_le32toh(value[0]), _le32toh(value[1]), _le32toh(value[2]))
+        cdef uint32_t *value = <uint32_t *>(self.key + self.key_size)
+        cdef uint32_t refcount = _le32toh(value[0])
+        assert refcount <= MAX_VALUE, "invalid reference count"
+        return (<char *>self.key)[:self.key_size], (refcount, _le32toh(value[1]), _le32toh(value[2]))
--- a/borg/testsuite/hashindex.py
+++ b/borg/testsuite/hashindex.py
@@ -100,3 +100,17 @@ class HashIndexTestCase(BaseTestCase):
         assert idx1[H(2)] == (7, 200, 200)
         assert idx1[H(3)] == (3, 300, 300)
         assert idx1[H(4)] == (6, 400, 400)
+
+    def test_chunkindex_summarize(self):
+        idx = ChunkIndex()
+        idx[H(1)] = 1, 1000, 100
+        idx[H(2)] = 2, 2000, 200
+        idx[H(3)] = 3, 3000, 300
+
+        size, csize, unique_size, unique_csize, unique_chunks, chunks = idx.summarize()
+        assert size == 1000 + 2 * 2000 + 3 * 3000
+        assert csize == 100 + 2 * 200 + 3 * 300
+        assert unique_size == 1000 + 2000 + 3000
+        assert unique_csize == 100 + 200 + 300
+        assert chunks == 1 + 2 + 3
+        assert unique_chunks == 3
--- a/borg/testsuite/hashindex.py
+++ b/borg/testsuite/hashindex.py
@@ -1,8 +1,13 @@
+import base64
 import hashlib
 import os
+import struct
 import tempfile
+import zlib
 
+import pytest
 from ..hashindex import NSIndex, ChunkIndex
+from .. import hashindex
 from . import BaseTestCase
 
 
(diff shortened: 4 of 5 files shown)


borg/hashindex.pyx
```diff
@@ -22,20 +22,30 @@ cdef extern from "_hashindex.c":
...
-cimport cython
+"""
+The HashIndex is *not* a general purpose data structure. The value size must be at least 4 bytes, and these
+first bytes are used for in-band signalling in the data structure itself::
+
+    (uint32_t) -1 (= all bits set) mark deleted entries.
+    ((uint32_t) -1) - 1 mark empty entries.
+
+This, among other things, limits the number of segments in a repository to 2**32-2. This is a hard limit above
+which data corruption will always occur.
```
enkore: ~10 PB

borg/hashindex.pyx
```diff
@@ -155,12 +166,13 @@ cdef class NSKeyIterator:
 cdef class ChunkIndex(IndexBase):
 
     value_size = 12
+    max_ref = 2**30-1  # to be *extra* safe, from constants like EMPTY and DELETED
```
verygreen: Why not make this 2**31-2 to double the space? Also 2**32-2 should also work it appears since the refcount cannot be negative, so we can always go for the unsigned as the counter, so (unsigned) -1 is deleted, (unsigned)-2 is empty and (unsigned) -3 is "too popular".

This is likely not super critical due to a sheer repetition count, I guess.

borg/hashindex.pyx
```diff
@@ -28,13 +28,24 @@ cdef extern from "_hashindex.c":
 
 cdef _NoDefault = object()
 
+"""
+The HashIndex is *not* a general purpose data structure. The value size must be at least 4 bytes, and these
+first bytes are used for in-band signalling in the data structure itself::
+
+    (uint32_t) -1 (= all bits set) mark deleted entries.
+    ((uint32_t) -1) - 1 mark empty entries.
```
verygreen: This probably should reference byte order right here. -1 is the same everywhere, kind of like 0, but -2 would have a different byte changed on LSB vs MSB.

borg/hashindex.pyx
```diff
@@ -191,11 +188,23 @@ cdef class ChunkIndex(IndexBase):
...
+        cdef long long size = 0, csize = 0, unique_size = 0, unique_csize = 0, chunks = 0, unique_chunks = 0
+        cdef int *values
+        cdef void *key = NULL
+
+        while True:
+            key = hashindex_next_key(self.index, key)
+            if not key:
+                break
+            unique_chunks += 1
+            values = <int*> (key + self.key_size)
```
verygreen: I wonder if it's safe to use int like this, int sizes happen to vary across architectures and platforms, so int32_t is probably more safe?

borg/hashindex.pyx
```diff
@@ -191,11 +188,23 @@ cdef class ChunkIndex(IndexBase):
...
+            chunks += _le32toh(values[0])
+            unique_size += _le32toh(values[1])
+            unique_csize += _le32toh(values[2])
+            size += <long long> _le32toh(values[1]) * _le32toh(values[0])
```
verygreen: Is this cast here really needed?
It's needed in C to avoid integer overflow on multiplication, but for a higher level language like python is this really true? What's the internal representation for those integers I wonder and if it's wrong, would such a cast work the same as in C?

borg/hashindex.pyx
```diff
@@ -212,7 +221,10 @@ cdef class ChunkIndex(IndexBase):
             raise KeyError(key)
         cdef int32_t refcount = _le32toh(data[0])
         if refcount != self.max_ref:
-            refcount -= 1
+            if self.max_ref - refcount < 1:
```
verygreen: it's unclear if this is enough. What's the internal python representation for max_ref when you do not specify type like that in the definition?

I think it's safer to do:
if refcount < 0 || self.max_ref - refcount < 1:

borg/hashindex.pyx
enkore: This is Cython (a strict superset of Python with most elements of C), these statements involve mostly `cdef`s... they are translated almost literally to C. I.e. _le32toh(values[1]) is a int32_t in the generated source code, and the resulting multiplication would be int32_t without an explicit cast.

borg/hashindex.pyx
enkore: The "int"s are replaced with "int32_t" in the next commit.

borg/hashindex.pyx
enkore: Good point. max_ref isn't a cdef and as such is a Python int object (~unlimited resolution). Python's RichCompare (which is also what is used for comparing stuff in Python if you write `><==` etc.) is emitted for this, together with an `int` instanciation from int32_t. So that's at least one round-trip to the allocator, we don't want that.

I'll change it to a cdef, so we get C semantics and no intermediary objects. The test will want a getter for that.

verygreen: note that you'll still need to compare refcount to <= 0 otherwise X - -2 = X+2 anyway and at 2**30 maxref we have plenty of time to still go positive and be bigger than 1.

verygreen:
I tried the current feature branch on a powermac (fedora23), and this is the result:

```
[root@powermac borg]# fakeroot -u tox
GLOB sdist-make: /root/bk/borg/setup.py
py34 create: /root/bk/borg/.tox/py34
py34 installdeps: -rrequirements.d/development.txt, attic
py34 inst: /root/bk/borg/.tox/dist/borgbackup-1.0.2.dev10+ng1f70fbd.zip
py34 installed: You are using pip version 7.1.2, however version 8.1.1 is available.,You should consider upgrading via the 'pip install --upgrade pip' command.,Attic==0.16,borgbackup==1.0.2.dev10+ng1f70fbd,coverage==4.0.3,Cython==0.24,msgpack-python==0.4.7,pluggy==0.3.1,py==1.4.31,pytest==2.9.1,pytest-benchmark==3.0.0,pytest-cov==2.2.1,tox==2.3.1,virtualenv==13.1.2,wheel==0.24.0
py34 runtests: PYTHONHASHSEED='3606867701'
py34 runtests: commands[0] | py.test --cov=borg --cov-config=../.coveragerc --benchmark-skip --pyargs borg.testsuite
============================= test session starts ==============================
platform linux -- Python 3.4.3, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
benchmark: 3.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=5.00us max_time=1.00s calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /root/bk/borg, inifile: setup.cfg
plugins: benchmark-3.0.0, cov-2.2.1
collected 515 items 

py34/lib/python3.4/site-packages/borg/testsuite/archive.py .......
py34/lib/python3.4/site-packages/borg/testsuite/archiver.py .s.....................ss.........X.........ssssssssssssssssssssssssssssssssssssssssss...............s...........ss.........X...........
py34/lib/python3.4/site-packages/borg/testsuite/benchmark.py sssssssssssssssssssssssssssssssssssssssssssssssss
py34/lib/python3.4/site-packages/borg/testsuite/chunker.py ...
py34/lib/python3.4/site-packages/borg/testsuite/compress.py ........
py34/lib/python3.4/site-packages/borg/testsuite/crypto.py ...
py34/lib/python3.4/site-packages/borg/testsuite/hashindex.py ..............
py34/lib/python3.4/site-packages/borg/testsuite/helpers.py .............................................................................................................................................................
py34/lib/python3.4/site-packages/borg/testsuite/key.py ....
py34/lib/python3.4/site-packages/borg/testsuite/locking.py ..............
py34/lib/python3.4/site-packages/borg/testsuite/logger.py ....
py34/lib/python3.4/site-packages/borg/testsuite/lrucache.py ..
py34/lib/python3.4/site-packages/borg/testsuite/platform.py sssss
py34/lib/python3.4/site-packages/borg/testsuite/repository.py ........................................
py34/lib/python3.4/site-packages/borg/testsuite/shellpattern.py .............................................................
py34/lib/python3.4/site-packages/borg/testsuite/upgrader.py .......
py34/lib/python3.4/site-packages/borg/testsuite/xattr.py s
---------------- coverage: platform linux, python 3.4.3-final-0 ----------------
Name                                                    Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------------------------
py34/lib/python3.4/site-packages/borg/archive.py          665     70    285     38    88%
py34/lib/python3.4/site-packages/borg/archiver.py         707    111    233     30    83%
py34/lib/python3.4/site-packages/borg/cache.py            314     23    100     20    89%
py34/lib/python3.4/site-packages/borg/helpers.py          638     84    219     25    86%
py34/lib/python3.4/site-packages/borg/key.py              306     54     72     23    76%
py34/lib/python3.4/site-packages/borg/locking.py          184     12     60      9    91%
py34/lib/python3.4/site-packages/borg/logger.py            80     19     16      4    74%
py34/lib/python3.4/site-packages/borg/lrucache.py          31      0      4      0   100%
py34/lib/python3.4/site-packages/borg/platform.py           5      2      0      0    60%
py34/lib/python3.4/site-packages/borg/remote.py           267     31    117     21    84%
py34/lib/python3.4/site-packages/borg/repository.py       564     37    210     25    91%
py34/lib/python3.4/site-packages/borg/shellpattern.py      36      0     22      0   100%
py34/lib/python3.4/site-packages/borg/upgrader.py         169     55     52     12    61%
py34/lib/python3.4/site-packages/borg/xattr.py            160     87     72     13    41%
-----------------------------------------------------------------------------------------
TOTAL                                                    4126    585   1462    220    83%

============= 410 passed, 103 skipped, 2 xpassed in 782.38 seconds =============
py35 create: /root/bk/borg/.tox/py35
ERROR: InterpreterNotFound: python3.5
flake8 create: /root/bk/borg/.tox/flake8
flake8 installdeps: flake8
flake8 inst: /root/bk/borg/.tox/dist/borgbackup-1.0.2.dev10+ng1f70fbd.zip
flake8 installed: You are using pip version 7.1.2, however version 8.1.1 is available.,You should consider upgrading via the 'pip install --upgrade pip' command.,borgbackup==1.0.2.dev10+ng1f70fbd,flake8==2.5.4,mccabe==0.4.0,msgpack-python==0.4.7,pep8==1.7.0,pyflakes==1.0.0,wheel==0.24.0
flake8 runtests: PYTHONHASHSEED='3606867701'
flake8 runtests: commands[0] | flake8
___________________________________ summary ____________________________________
  py34: commands succeeded
ERROR:   py35: InterpreterNotFound: python3.5
  flake8: commands succeeded
```

enkore:
That looks indeed good. Can you try with 0ac324c whether that still works for BE? Just using INT32_MAX for the ceiling *should* *probably* be no problem, but I will have to think some more about it to be sure...

Naturally all of this breaks for non two's complement, but I doubt that's relevant.

verygreen:
```
[root@powermac borg]# fakeroot -u tox
GLOB sdist-make: /root/bk/borg/setup.py
py34 inst-nodeps: /root/bk/borg/.tox/dist/borgbackup-1.0.2.dev11+ng0ac324c.zip
py34 installed: You are using pip version 7.1.2, however version 8.1.1 is available.,You should consider upgrading via the 'pip install --upgrade pip' command.,Attic==0.16,borgbackup==1.0.2.dev11+ng0ac324c,coverage==4.0.3,Cython==0.24,msgpack-python==0.4.7,pluggy==0.3.1,py==1.4.31,pytest==2.9.1,pytest-benchmark==3.0.0,pytest-cov==2.2.1,tox==2.3.1,virtualenv==13.1.2,wheel==0.24.0
py34 runtests: PYTHONHASHSEED='366175955'
py34 runtests: commands[0] | py.test --cov=borg --cov-config=../.coveragerc --benchmark-skip --pyargs borg.testsuite
============================= test session starts ==============================
platform linux -- Python 3.4.3, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
benchmark: 3.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=5.00us max_time=1.00s calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /root/bk/borg, inifile: setup.cfg
plugins: benchmark-3.0.0, cov-2.2.1
collected 515 items 

py34/lib/python3.4/site-packages/borg/testsuite/archive.py .......
py34/lib/python3.4/site-packages/borg/testsuite/archiver.py .s.....................ss.........X.........ssssssssssssssssssssssssssssssssssssssssss...............s...........ss.........X...........
py34/lib/python3.4/site-packages/borg/testsuite/benchmark.py sssssssssssssssssssssssssssssssssssssssssssssssss
py34/lib/python3.4/site-packages/borg/testsuite/chunker.py ...
py34/lib/python3.4/site-packages/borg/testsuite/compress.py ........
py34/lib/python3.4/site-packages/borg/testsuite/crypto.py ...
py34/lib/python3.4/site-packages/borg/testsuite/hashindex.py ..............
py34/lib/python3.4/site-packages/borg/testsuite/helpers.py .............................................................................................................................................................
py34/lib/python3.4/site-packages/borg/testsuite/key.py ....
py34/lib/python3.4/site-packages/borg/testsuite/locking.py ..............
py34/lib/python3.4/site-packages/borg/testsuite/logger.py ....
py34/lib/python3.4/site-packages/borg/testsuite/lrucache.py ..
py34/lib/python3.4/site-packages/borg/testsuite/platform.py sssss
py34/lib/python3.4/site-packages/borg/testsuite/repository.py ........................................
py34/lib/python3.4/site-packages/borg/testsuite/shellpattern.py .............................................................
py34/lib/python3.4/site-packages/borg/testsuite/upgrader.py .......
py34/lib/python3.4/site-packages/borg/testsuite/xattr.py s
---------------- coverage: platform linux, python 3.4.3-final-0 ----------------
Name                                                    Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------------------------
py34/lib/python3.4/site-packages/borg/archive.py          665     70    285     38    88%
py34/lib/python3.4/site-packages/borg/archiver.py         707    111    233     30    83%
py34/lib/python3.4/site-packages/borg/cache.py            314     23    100     20    89%
py34/lib/python3.4/site-packages/borg/helpers.py          638     86    219     26    85%
py34/lib/python3.4/site-packages/borg/key.py              306     54     72     23    76%
py34/lib/python3.4/site-packages/borg/locking.py          184     12     60      9    91%
py34/lib/python3.4/site-packages/borg/logger.py            80     19     16      4    74%
py34/lib/python3.4/site-packages/borg/lrucache.py          31      0      4      0   100%
py34/lib/python3.4/site-packages/borg/platform.py           5      2      0      0    60%
py34/lib/python3.4/site-packages/borg/remote.py           267     29    117     21    85%
py34/lib/python3.4/site-packages/borg/repository.py       564     37    210     25    91%
py34/lib/python3.4/site-packages/borg/shellpattern.py      36      0     22      0   100%
py34/lib/python3.4/site-packages/borg/upgrader.py         169     55     52     12    61%
py34/lib/python3.4/site-packages/borg/xattr.py            160     87     72     13    41%
-----------------------------------------------------------------------------------------
TOTAL                                                    4126    585   1462    221    83%

============= 410 passed, 103 skipped, 2 xpassed in 782.69 seconds =============
py35 create: /root/bk/borg/.tox/py35
ERROR: InterpreterNotFound: python3.5
flake8 inst-nodeps: /root/bk/borg/.tox/dist/borgbackup-1.0.2.dev11+ng0ac324c.zip
flake8 installed: You are using pip version 7.1.2, however version 8.1.1 is available.,You should consider upgrading via the 'pip install --upgrade pip' command.,borgbackup==1.0.2.dev11+ng0ac324c,flake8==2.5.4,mccabe==0.4.0,msgpack-python==0.4.7,pep8==1.7.0,pyflakes==1.0.0,wheel==0.24.0
flake8 runtests: PYTHONHASHSEED='366175955'
flake8 runtests: commands[0] | flake8
___________________________________ summary ____________________________________
  py34: commands succeeded
ERROR:   py35: InterpreterNotFound: python3.5
  flake8: commands succeeded
```

enkore:
Thanks, looks good. I made some clean-up which don't affect functionality in b7c6504.

Can you copy the cache (~/.borg/cache/<REPO ID>/, the repo id is in the repository/config file) from a LE machine to your BE machine and check whether e.g. borg info gives you the same output (pass `--debug` to make sure that it doesn't rebuild it for some reason).

verygreen:
copied the cache around from Be to LE and from LE to BE and see no problems.

borg/hashindex.pyx
```diff
@@ -175,9 +192,39 @@ cdef class ChunkIndex(IndexBase):
...
 
+    def incref(self, key):
+        """Increase refcount for 'key', return (refcount, size, csize)"""
+        assert len(key) == self.key_size
+        data = <int32_t *>hashindex_get(self.index, <char *>key)
+        if not data:
+            raise KeyError(key)
+        cdef int32_t refcount = _le32toh(data[0])
+        if refcount != INT32_MAX:
+            if INT32_MAX - refcount < 1:
```
ThomasWaldmann: refcount is not equal to INT32_MAX (see previous "if"), thus INT32_MAX - refcount can never be 0.

So the "< 1" (which is equivalent to "<= 0" for integers) is misleading as it can never be 0.

So what you really meant here is "< 0"?

When is that the case?

enkore: You're right, `refcount != INT32_MAX` is already a sufficient check here. It's just the arbitrary additions in _add that need this verbose kind of check.

borg/hashindex.pyx
```diff
@@ -175,9 +192,39 @@ cdef class ChunkIndex(IndexBase):
...
+
+    def decref(self, key):
+        """Decrease refcount for 'key', return (refcount, size, csize)"""
+        assert len(key) == self.key_size
+        data = <int32_t *>hashindex_get(self.index, <char *>key)
+        if not data:
+            raise KeyError(key)
+        cdef int32_t refcount = _le32toh(data[0])
+        if refcount != INT32_MAX:
+            if INT32_MAX - refcount < 1:
```
ThomasWaldmann: same here.

borg/hashindex.pyx
verygreen: The check is in fact correct, it is needed for corrupted repos from before this change.
So INT32_MAX - (int32_t)-2 (or any other negative number, really) is some negative number.
It's either this or we need to make sure that refcount is > 0 (I don't think there's such a check on load?)

borg/testsuite/hashindex.py
```diff
@@ -100,3 +101,94 @@ def test_chunkindex_merge(self):
...
+            assert refcount == self.MAX_REF
+        for i in range(5):
+            refcount, *_ = idx.decref(H(1))
+            assert refcount == self.MAX_REF
+
+    def test_chunkindex_merge_overflow1(self):
+        # Check that it does not overflow at 2**31-2
+        idx1 = ChunkIndex()
+        # n.b. 2**31-1 is odd
+        half = self.MAX_REF // 2
```
ThomasWaldmann: that is 2^30 - 1.

2 * that == 2^31 - 2.

so, why should that "overflow" (see comment)?

borg/hashindex.pyx
enkore: ... but INT32_MAX - (something negative) is only < 1 for INT32_MIN case (because INT32_MIN = -2**31). So if the refcount wrapped around in the past in a cache this wouldn't really work.

borg/testsuite/hashindex.py
enkore: Check that it does **not** overflow at 2**31-2

borg/testsuite/hashindex.py
```diff
@@ -100,3 +101,94 @@ def test_chunkindex_merge(self):
...
+    def test_chunkindex_merge_overflow3(self):
+        # Check that merging maxes out right at 2**31-1
+        idx1 = ChunkIndex()
+        half = self.MAX_REF // 2
+        idx1[H(1)] = half + 1, 1, 2
+        idx2 = ChunkIndex()
+        idx2[H(1)] = half, 1, 2
+        idx1.merge(idx2)
+        refcount, *_ = idx1[H(1)]
+        assert refcount == self.MAX_REF
```
ThomasWaldmann: no overflow here either.

borg/hashindex.pyx
verygreen: I am not sure I get what you mean, this has nothing to do with INT32_MIN other than
INT32_MAX - -1 == INT32_MIN
INT32_MAX - -2 ==  INT32_MIN + 1
...
INT32_MAX - INT32_MIN == 0

All negative numbers (except the last one, but still smaller than 1)

borg/hashindex.pyx
```diff
@@ -191,22 +238,55 @@ cdef class ChunkIndex(IndexBase):
...
+        values = <int32_t*> hashindex_get(self.index, key)
+        if values:
+            refcount1 = _le32toh(values[0])
+            refcount2 = _le32toh(data[0])
+            if INT32_MAX - refcount1 < refcount2:
+                values[0] = _htole32(INT32_MAX)
+            elif INT32_MAX - refcount2 < refcount1:
+                values[0] = _htole32(INT32_MAX)
+            else:
+                values[0] = _htole32(refcount1 + refcount2)
```
ThomasWaldmann: maybe using sum = ... in the if/elif/else and afterwards values[0] = _htole32(sum) would be easier to read.

borg/hashindex.pyx
ThomasWaldmann: ah, add refcount to both sides: INT32_MAX < refcount + 1

add a comment with that?

ThomasWaldmann: still it makes no sense due to "if" above.

borg/hashindex.pyx
ThomasWaldmann: maybe a function "refcount_add" and "refcount_sub" that just has the integer / boundary math would be better here, easier unit tests also.

borg/hashindex.pyx
enkore: @verygreen Ok, I see what you mean, but `INT32_MAX - -1` is the undefined behaviour we want to avoid here. I'm not sure we should consider negative refcounts; if the refcount already went beyond 2**31 then we don't really know how far it went. It could already crossed over zero again (with data loss). (For 1.x the issue is a bit academic, because the only kind of data that would realistically make very high refcounts are large [possibly sparse] disk images with lots of zeroes, but an all zero chunk would always go to MAX_EXP in the chunker [my understanding], so 2**31-1 * 8 MB ~ 20 PB)

verygreen: how does the if above protects us here?

refcount = -2;
if refcount != INT32_MAX then // True
  refcount--;
now refcount is -3, where as it should become INT32_MAX thereby repairing earlier overflow before this code was introduced (or just plain cache corruption of some other sort).

verygreen: @enkore The behavior is undefined, but in reality we all know what the end result is.

The problem statement: due to a prior corruption (1.0.[01] on bigendian architectures, integer overflow on very popular chunks before this change was merged, some other sources of corruption) refcount might become negative (ignoring two validly negative values -1 and -2).

We can assume that "this is incorrect refcount, but we are too lazy to calculate correct one, so set it to "really popular" value" - what the current code is doing.
Alternatively the loading method (hashindex_get, except it does not dereference value so not a good fit?) could check the value to be valid (i.e. positive, 0, -1, or -2) and if not - throw the cache away and rebuild or assign a saner "really popular value" or just throw the hands in the air and abort.

I guess there might be other solutions here. But ignoring this condition seems to be the least safe one.

enkore: *If* we consider this case (and as I wrote above it might be valid to ignore this), then we'd need to do something like

```py
# refcount = -something
if refcount < 0:
  refcount = INT32_MAX
elif refcount != INT32_MAX:
  refcount += / -= 1
```

For _add we'd need to check sign of operands and choose appropiate check from there. Overflowing is UB and can't be used.

> (or just plain cache corruption of some other sort).

There are no integrity checks for the indices, but they should be added.

borg/hashindex.pyx
```diff
@@ -22,13 +22,22 @@ cdef extern from "_hashindex.c":
...
 
 cdef _NoDefault = object()
 
-cimport cython
+"""
+The HashIndex is *not* a general purpose data structure. The value size must be at least 4 bytes, and these
+first bytes are used for in-band signalling in the data structure itself (byte ordering is LE)::
+
+    0xffffffff mark empty entries.
+    0xfffffffe mark deleted entries.
```
ThomasWaldmann: "marks"? 2x

borg/hashindex.pyx
enkore: Note that for the first overflown values (-1, -2) the chunk is lost from the index. So a chunk would only "survive" overflowing by merging, not from incref/decref. Not relevant to the end result.

verygreen: nope.
In fact the first overflow value is INT32_MIN ( = INT32_MAX + 1).
the -2 and -1 are the two last possible values after which we revert to 0.

enkore: Yes.

enkore:
Q: If we just changed all int32_t to uint32_t (and adjust the limit to UINT32_MAX accordingly), then all wrapped around values from previous signed overflows should become valid, and even correct, refcounts.

126, 127, -128, -127 => 126, 127, 128, 129

borg/hashindex.pyx
```diff
@@ -191,22 +237,55 @@ cdef class ChunkIndex(IndexBase):
...
+            key = hashindex_next_key(self.index, key)
+            if not key:
+                break
+            unique_chunks += 1
+            values = <int32_t*> (key + self.key_size)
+            chunks += _le32toh(values[0])
+            unique_size += _le32toh(values[1])
+            unique_csize += _le32toh(values[2])
+            size += <long long> _le32toh(values[1]) * _le32toh(values[0])
+            csize += <long long> _le32toh(values[2]) *  _le32toh(values[0])
```
ThomasWaldmann: 1 blank too much here after *

borg/hashindex.pyx
```diff
@@ -191,22 +237,55 @@ cdef class ChunkIndex(IndexBase):
...
-        hashindex_add(self.index, <char *>key, data)
+        self._add(<char*> key, data)
+
+    cdef _add(self, void *key, int32_t *data):
+        cdef int refcount1, refcount2, result
+        values = <int32_t*> hashindex_get(self.index, key)
+        if values:
+            refcount1 = _le32toh(values[0])
+            refcount2 = _le32toh(data[0])
+            if (refcount1 < 0 or refcount2 < 0 or
```
ThomasWaldmann: when is a refcount < 0?

verygreen:
yes, converting to unsigned like that would work.
Keep in mind that if you do that, you are limiting your future self from adding some additional special index counts should you need them. With current approach you have all negative space for that.

borg/testsuite/hashindex.py
```diff
@@ -100,3 +101,104 @@ def test_chunkindex_merge(self):
...
+
+    def test_decref_limit(self):
+        idx1 = ChunkIndex()
+        idx1[H(1)] = (self.MAX_REF, 6, 7)
+        idx1.decref(H(1))
+        refcount, *_ = idx1[H(1)]
+        assert refcount == self.MAX_REF
+
```
ThomasWaldmann: there could be also tests for a normal incref and decref

enkore:
> ... might consider making this change in master only. OTOH not switching
to unsigned needs more complex overflow checks and it looks like no one
is really comfortable writing those...

> I'll leave this here so we can think about it for a bit. if we decided
what course of action and what should go where then i'll untangle
what's been intertwined in the course of this changeset.

enkore:
as thomas noted independent of what will be applied to 1.0-maint and/or master (which might be different things in the end): this needs very careful testing & review. the chance that somewhere overflows happened is slim and the chance that we're putting extra bugs in here is higher.

borg/hashindex.pyx
```diff
@@ -110,22 +110,23 @@ cdef class NSIndex(IndexBase):
...
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
         if not data:
             raise KeyError(key)
         return _le32toh(data[0]), _le32toh(data[1])
 
     def __setitem__(self, key, value):
         assert len(key) == self.key_size
-        cdef int32_t[2] data
+        cdef uint32_t[2] data
+        assert value[0] <= MAX_REF, "maximum number of segments reached"
```
ThomasWaldmann: in that case, it is a segment number, not a refcount, so MAX_SEGMENT?
or just have a generic MAX_VALUE?

borg/hashindex.pyx
```diff
@@ -158,10 +159,15 @@ cdef class NSKeyIterator:
...
+        cdef uint32_t *value = <uint32_t *>(self.key + self.key_size)
         return (<char *>self.key)[:self.key_size], (_le32toh(value[0]), _le32toh(value[1]))
 
-assert INT32_MAX == 2**31-1
+
+assert UINT32_MAX == 2**32-1
+
+# module-level constant because cdef's in classes can't have default values
+cdef uint32_t _MAX_REF = 2**32-3
+MAX_REF = _MAX_REF
```
ThomasWaldmann: maybe we could reserve a few more "special" values?
you never know when you need them...

borg/hashindex.pyx
```diff
@@ -238,40 +240,39 @@ cdef class ChunkIndex(IndexBase):
...
                 break
             unique_chunks += 1
-            values = <int32_t*> (key + self.key_size)
+            values = <uint32_t*> (key + self.key_size)
             chunks += _le32toh(values[0])
             unique_size += _le32toh(values[1])
             unique_csize += _le32toh(values[2])
             size += <long long> _le32toh(values[1]) * _le32toh(values[0])
-            csize += <long long> _le32toh(values[2]) *  _le32toh(values[0])
+            csize += <long long> _le32toh(values[2]) * _le32toh(values[0])
```
ThomasWaldmann: so if we have uint32 * uint32, that gives a uint64.

borg/testsuite/hashindex.py
```diff
@@ -118,20 +119,22 @@ def test_chunkindex_summarize(self):
 
 
 class HashIndexRefcountingTestCase(BaseTestCase):
-    MAX_REF = 2**31-1
+    def _negative_to_uint(self, number):
+        # yes, a substraction would get the job done as well, but let's be clear about intent
```
ThomasWaldmann: subtraction

borg/testsuite/hashindex.py
```diff
@@ -150,26 +153,26 @@ def merge(refcount1, refcount2):
     def test_chunkindex_merge_limit1(self):
         # Check that it does *not* limit at 2**31-2
         # (2**31-1 is odd)
```
ThomasWaldmann: comment needs update to new range

borg/testsuite/hashindex.py
```diff
@@ -150,26 +153,26 @@ def merge(refcount1, refcount2):
...
-        half = self.MAX_REF // 2
-        assert self._merge(half + 2, half) == self.MAX_REF
-        assert self._merge(half + 1, half + 1) == self.MAX_REF
+        half = hashindex.MAX_REF // 2
+        assert self._merge(half + 2, half) == hashindex.MAX_REF
+        assert self._merge(half + 1, half + 1) == hashindex.MAX_REF
 
     def test_chunkindex_merge_negative(self):
-        assert self._merge(0, -3) == self.MAX_REF
+        assert self._merge(0, self._negative_to_uint(-3)) == hashindex.MAX_REF
```
ThomasWaldmann: sense?

borg/testsuite/hashindex.py
```diff
@@ -180,25 +183,49 @@ def test_chunkindex_add(self):
...
     def test_decref_limit(self):
         idx1 = ChunkIndex()
-        idx1[H(1)] = (self.MAX_REF, 6, 7)
+        idx1[H(1)] = (hashindex.MAX_REF, 6, 7)
         idx1.decref(H(1))
         refcount, *_ = idx1[H(1)]
-        assert refcount == self.MAX_REF
+        assert refcount == hashindex.MAX_REF
 
     def test_negative_refcount(self):
```
ThomasWaldmann: sense?

ThomasWaldmann:
@verygreen btw, I've set up a qemu ppc machine (on a intel i5) with debian 8 and it is faster than expected. :)

ThomasWaldmann:
@enkore the uint stuff looks quite good, just that "negative" tests are confusing somehow.

borg/hashindex.pyx
verygreen: basically yes. or unsigned long long which should be the same thing.

verygreen:
@ThomasWaldmann Yes, ppc64 in qemu is quite good. still slower than the real thing from 10 years ago, but rapidly approaching it. The machines that qemu emulates are quite limited in other ways, though. I think I run into some significant RAM limitations and cpu count too, don't remember exactly.

Considering that you can have dual-ppc powermac g5 from 2005 for like $200 (or so was the price a few years ago) and it takes quite cheap (at the time) DDR2 RAM, it made it a nobrainer to get some for various testing, native compiles and such. And it's also not as slow as old sparcs, alphas and such, takes modern SATA drives (so SSDs are an option too).

borg/hashindex.pyx
```diff
@@ -238,40 +240,39 @@ cdef class ChunkIndex(IndexBase):
...
+    cdef _add(self, void *key, uint32_t *data):
+        cdef uint32_t refcount1, refcount2, result
+        values = <uint32_t*> hashindex_get(self.index, key)
         if values:
             refcount1 = _le32toh(values[0])
             refcount2 = _le32toh(data[0])
-            if (refcount1 < 0 or refcount2 < 0 or
-                    INT32_MAX - refcount1 < refcount2 or INT32_MAX - refcount2 < refcount1):
-                result = INT32_MAX
+            if _MAX_REF - refcount1 < refcount2 or _MAX_REF - refcount2 < refcount1:
```
verygreen: I wonder if a much more readable check here would be:
(unsigned long long) refcount1 + refcount2 > _MAX_REF

This way we nip overflow in the bud during the check and the intent is much more clear too for the reader.

borg/hashindex.pyx
enkore: Let's reserve 1000 or so.

enkore:
> @enkore the uint stuff looks quite good, just that "negative" tests are confusing somehow.

Indeed they are. I'll see how I can clear them up a bit, but I want these test cases in to verify that these cases of prior overflow are handled correctly.

ThomasWaldmann:
maybe they just miss some more comments about what they are actually about and how it can/could happen.

enkore:
hashindex.c: https://gist.github.com/enkore/ef2c256a03adf056aea795b573f74e7c

borg/hashindex.pyx
```diff
@@ -269,16 +269,11 @@ cdef class ChunkIndex(IndexBase):
...
         if values:
-            refcount1 = _le32toh(values[0])
-            refcount2 = _le32toh(data[0])
-            if _MAX_VALUE - refcount1 < refcount2 or _MAX_VALUE - refcount2 < refcount1:
-                result = _MAX_VALUE
-            else:
-                result = refcount1 + refcount2
-            values[0] = _htole32(result)
+            result64 = <uint64_t> _le32toh(values[0]) + _le32toh(data[0])
+            values[0] = _htole32(min(result64, _MAX_VALUE))
```
enkore: This way it's much clearer as @verygreen suggested. Note that if (for some reason) we actually make use of one of the ~1000 reserved values, then this needs changing. But since we don't know what semantics these "maybe" future values have this is okay by me.

enkore:
So with these asserts it's assured to blow up if it has overflown in the past and a refcount is lingering around in our reserved region.

verygreen:
I think this is ok for it to blow up for reserved values, same as if the has count somehow managed to come to (uint32_t)-1 or (uint32_t)-2 which are the two reserved values now.

This is something for borg check --repair to care for I guess.

enkore:
re. changelog. We are addressing ~~two~~ three separate-but-related issues here

1. Fixed wrong output of `borg info` on BE
2. Fixed wrong refcount calculation on BE (that's the BE stuff in _merge), which could have caused data loss (not likely, but possible)
3. Fixed refcount overflowing for very large data sets with specific structures (e.g. billions of identical small files, many archives of very large sparse files). This possibly caused data loss.

N.b. both occurences of "possible data loss" are unlikely in practice, however, a sequence of operations causing data loss is possible.

We should also advise to delete all caches generated on BE systems. And caches for repositories suspected to fall under (3) as well. For anything affected by (2) and (3) we should recommend to run "borg check" **and** "borg extract --dry-run" on all archives.

enkore:
This should be done now. I did some additional testing and did not see any irregularities. There is now also a test doing some sanity checks against a serialized (as written to disk) hash index.

Base is outdated, so needs rebasing; Microcommits will be squashed as well.

ThomasWaldmann:
as you did a byte order fix for big endian, please mention that - i didn't find a commit comment (1st line) about it.

borg/cache.py
```diff
@@ -311,7 +311,7 @@ def create_master_idx(chunk_idx):
             cleanup_outdated(cached_ids - archive_ids)
             if archive_ids:
                 chunk_idx = None
-                for archive_id in archive_ids:
+                for archive_id in sorted(archive_ids):
```
ThomasWaldmann: ?

borg/hashindex.pyx
```diff
@@ -22,13 +22,32 @@ cdef extern from "_hashindex.c":
...
+
+The constant MAX_VALUE is defined as 2**32-1023. As noted above 2**32-1 and -2 are reserved for internal
+use by the HashIndex itself. The values from MAX_VALUE (exclusive) to 2**32-2 (exclusive) are reserved
+as well for future use, should the need for special in-band signalling arise.
+"""
+
+assert UINT32_MAX == 2**32-1
+
+# module-level constant because cdef's in classes can't have default values
+cdef uint32_t _MAX_VALUE = 2**32-1023
```
ThomasWaldmann: if you want to start the reserved values at a "even" (mod 1024) number, the max allowed regular value needs to be 2**32 - 1024.

borg/hashindex.pyx
```diff
@@ -149,24 +169,32 @@ cdef class NSKeyIterator:
...
+        cdef uint32_t *value = <uint32_t *>(self.key + self.key_size)
         return (<char *>self.key)[:self.key_size], (_le32toh(value[0]), _le32toh(value[1]))
 
 
 cdef class ChunkIndex(IndexBase):
+    """
+    Mapping of 32 byte keys to (refcount, size, csize) (all 32 bit signed).
+
+    The reference count cannot overflow. If an overflow would occur, the refcount
+    is fixed to 2**31-1 and will neither increase nor decrease by incref(), decref()
```
ThomasWaldmann: needs to refer to _MAX_VALUE not 2**31-1

enkore:
> as you did a byte order fix for big endian, please mention that - i didn't find a commit comment (1st line) about it.

That was in #887, this changeset only moves that particular fix.

borg/hashindex.pyx
enkore: Indeed!

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,186 @@ def test_chunkindex_merge(self):
...
+
+    def test_chunkindex_merge_negative(self):
+        assert self._merge(0, self._int_to_uint(-1023)) == hashindex.MAX_VALUE
+
+
+class HashIndexDataTestCase(BaseTestCase):
+    # This bytestring was created with 1.0-maint at c2f9533
+    HASHINDEX = b'Gb"0J]a]R$\'Sb#Y:I;r1Ne"V?+Yd"E<ckKQkJHQcG61!$2<RPKGOcMn\\j*@DgUo?TdBsRa4pl.0afm>>Z7O2)/Bb=MbH`\\BZ' \
+                b'7O2)/Bb=MbH`\\BZ7O2)/Bb=MbH`\\BZ7O2)/Bb=MbH`\\BZ7O2)/Bb=MbH`\\BZ7O2)/Bb=MbH`\\BZ7O2)ln@&NF6>#IS_\'+' \
+                b'nR\'Z76]9h.G?%66bQ=X3bAQb=1=dEW%Q=X3bAQb=1=dEW%Q=X3bAQb=1=dEW%Q=X3bAQb=1=dEW%pAaLig9quB'
```
ThomasWaldmann: maybe use hex as in other places? or base64 so that it at least doesn't need that many escapes in the value?

borg/cache.py
enkore: I'll remove that commit in squashing, it was a testing-only thing (no harm in it, just off-topic), to make the chunks cache creation deterministic. This made it much easier to see if there where any changes in the generated files.

borg/_hashindex.c
```diff
@@ -439,51 +439,3 @@ hashindex_get_size(HashIndex *index)
...
-}
-
-static void
-hashindex_add(HashIndex *index, const void *key, int32_t *other_values)
-{
-    int32_t *my_values = (int32_t *)hashindex_get(index, key);
-    if(my_values == NULL) {
-        hashindex_set(index, key, other_values);
-    } else {
-        *my_values += *other_values;
```
ThomasWaldmann: note that this was not fixed for big endian before your changes in this PR.

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,186 @@ def test_chunkindex_merge(self):
...
+        with pytest.raises(KeyError):
+            idx.incref(H(1))
+        with pytest.raises(KeyError):
+            idx.decref(H(1))
+        with pytest.raises(KeyError):
+            idx[H(1)]
+
+    # The following test cases check that refcounts that overflowed when they were handled
+    # as signed integers are handled correctly. Note that the signed integer values -1 and -2
+    # cannot be handled, and that the values -3
```
ThomasWaldmann: sentence is incomplete

borg/testsuite/hashindex.py
enkore: Let's go with b64 (not much difference to a85); hexlify is much longer.

borg/_hashindex.c
enkore: Right! I was thinking about summarize.

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,187 @@ def test_chunkindex_merge(self):
...
+
+    def test_chunkindex_merge_negative(self):
+        assert self._merge(0, self._int_to_uint(-1023)) == hashindex.MAX_VALUE
+
+
+class HashIndexDataTestCase(BaseTestCase):
+    # This bytestring was created with 1.0-maint at c2f9533
+    HASHINDEX = b'eJzt0L0NgmAUhtHLT0LDEI6AuAEhMVYmVnSuYefC7AB3Aj9KNedJbnfyFne6P67P27w0EdG1Eac+Cm1ZybAsy7Isy7Isy7Isy7I' \
+                b'sy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7Isy7LsL9nhc+cqTZ' \
+                b'3XlO2Ys++Du5fX+l1/YFmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVn2/+0O2rYccw=='
```
enkore: `ZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZVmWZ` interesting. frame/chunk length limit of zlib?

borg/hashindex.pyx
```diff
@@ -22,13 +22,35 @@ cdef extern from "_hashindex.c":
...
+    0xffffffff marks empty entries.
+    0xfffffffe marks deleted entries.
+
+This, among other things, limits the number of segments in a repository to 2**32-2. This was a hard limit above
+which data corruption will always occur.
+
+The constant MAX_VALUE is defined as 2**32-1023. As noted above 2**32-1 and -2 are reserved for internal
+use by the HashIndex itself. The values from MAX_VALUE (exclusive) to 2**32-2 (exclusive) are reserved
+as well for future use, should the need for special in-band signalling arise.
+
```
ThomasWaldmann: maybe this could be just summarized to what we have now.
```
0 .. MAX_VALUE = usable for clients
MAX_VALUE+1 .. UINT32_MAX = reserved

current reserved value meanings:
...
```

borg/hashindex.pyx
```diff
@@ -149,24 +178,44 @@ cdef class NSKeyIterator:
         self.key = hashindex_next_key(self.index, <char *>self.key)
         if not self.key:
             raise StopIteration
-        cdef int *value = <int *>(self.key + self.key_size)
-        return (<char *>self.key)[:self.key_size], (_le32toh(value[0]), _le32toh(value[1]))
+        cdef uint32_t *value = <uint32_t *>(self.key + self.key_size)
+        cdef uint32_t segment = _le32toh(value[0])
+        assert <uint32_t> segment <= _MAX_VALUE, "maximum number of segments reached"
```
ThomasWaldmann: is `<uint32_t>` required here (see line above)?
(maybe check this kind of construct at other places also)

borg/hashindex.pyx
```diff
@@ -149,24 +178,44 @@ cdef class NSKeyIterator:
...
     def __getitem__(self, key):
         assert len(key) == self.key_size
-        data = <int *>hashindex_get(self.index, <char *>key)
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
         if not data:
-            raise KeyError
-        return _le32toh(data[0]), _le32toh(data[1]), _le32toh(data[2])
+            raise KeyError(key)
+        cdef uint32_t refcount = _le32toh(data[0])
+        assert refcount <= _MAX_VALUE
```
ThomasWaldmann: add a msg?

borg/hashindex.pyx
```diff
@@ -191,22 +268,53 @@ cdef class ChunkIndex(IndexBase):
...
+    cdef _add(self, void *key, uint32_t *data):
+        cdef uint64_t refcount1, refcount2, result64
+        values = <uint32_t*> hashindex_get(self.index, key)
+        if values:
+            refcount1 = _le32toh(values[0])
+            refcount2 = _le32toh(data[0])
+            assert refcount1 <= _MAX_VALUE
+            assert refcount2 <= _MAX_VALUE
+            result64 = refcount1 + refcount2
+            values[0] = _htole32(min(result64, _MAX_VALUE))
```
ThomasWaldmann: hmm, min() and max() are made to deal with 64bit uint values?

borg/hashindex.pyx
ThomasWaldmann: also, if the unit tests rely on it being a odd number, add a comment and a assert here?

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,187 @@ def test_chunkindex_merge(self):
...
+        # Beyond crossover, result of addition would be 2**31
+        half = hashindex.MAX_VALUE // 2
+        assert self._merge(half + 2, half) == hashindex.MAX_VALUE
+        assert self._merge(half + 1, half + 1) == hashindex.MAX_VALUE
+
+    def test_chunkindex_add(self):
+        idx1 = ChunkIndex()
+        idx1.add(H(1), 5, 6, 7)
+        assert idx1[H(1)] == (5, 6, 7)
+        idx1.add(H(1), 1, 0, 0)
```
ThomasWaldmann: hmm, do we ever call add with a negative refcount?

borg/hashindex.pyx
```diff
@@ -175,9 +224,37 @@ cdef class ChunkIndex(IndexBase):
...
+        return refcount, _le32toh(data[1]), _le32toh(data[2])
+
+    def decref(self, key):
+        """Decrease refcount for 'key', return (refcount, size, csize)"""
+        assert len(key) == self.key_size
+        data = <uint32_t *>hashindex_get(self.index, <char *>key)
+        if not data:
+            raise KeyError(key)
+        cdef uint32_t refcount = _le32toh(data[0])
+        assert refcount <= _MAX_VALUE
```
ThomasWaldmann: assert 1 <= refcount <= _MAX_VALUE

we must not have refcount == 0 and then subtract 1.

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,187 @@ def test_chunkindex_merge(self):
...
+        idx1.incref(H(1))
+        refcount, *_ = idx1[H(1)]
+        assert refcount == hashindex.MAX_VALUE
+
+    def test_decref_limit(self):
+        idx1 = ChunkIndex()
+        idx1[H(1)] = (hashindex.MAX_VALUE, 6, 7)
+        idx1.decref(H(1))
+        refcount, *_ = idx1[H(1)]
+        assert refcount == hashindex.MAX_VALUE
```
ThomasWaldmann: also test if decref() of a already zero refcount raises an AssertionError.

it should not even exist, but you never know...

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,187 @@ def test_chunkindex_merge(self):
...
+
+    def _int_to_uint(self, number):
+        """Return unsigned interpretation of a signed integer."""
+        # yes, a subtraction would get the job done as well, but let's be clear about intent
+        # and do it like it occurs in the C code.
+        return struct.unpack('I', struct.pack('i', number))[0]
+
+    def test_negative_refcount(self):
+        idx1 = ChunkIndex()
+        idx1[H(1)] = (self._int_to_uint(-1023), 6, 7)
```
ThomasWaldmann: maybe we need a constant to compute this -1023 from also.

borg/testsuite/hashindex.py
```diff
@@ -100,3 +105,187 @@ def test_chunkindex_merge(self):
...
+        idx1.decref(H(1))
+        refcount, *_ = idx1[H(1)]
+        assert refcount == hashindex.MAX_VALUE
```
ThomasWaldmann: hmm, what's the sense of this test?

it only works because the -1023 int is the same as the MAX_VALUE uint already.

if a previous int32 overflow would have resulted in any other value, it would not work.

borg/hashindex.pyx
enkore: Yep. (Tests check this also). Implementation detail: Cython is quite smart here and doesn't use Python's min() (which would require object creation), but emits an inlined min() instead; C doesn't have a min() so no risk there.

borg/testsuite/hashindex.py
enkore: Yep, one of the negative tests uses merge, but I'll add another explicit test for it.

borg/hashindex.pyx
enkore: Good idea. It shouldn't be possible by any execution path, but corruption might have lead there.

borg/testsuite/hashindex.py
enkore: Correct, this test is covered by other tests (which was not the case previously).

borg/hashindex.pyx
enkore: I want to keep an account of why/what/the/hell somewhere, but the testsuite is probably a more appropiate place than the module docs for documenting the old behaviours some tests are based on.

borg/testsuite/hashindex.py
ThomasWaldmann: hmm, maybe that is not even possible from the api.
as we have uint32 refcounts, everything that could be a refcount can't be negative.

so, add a .sub() method if we need it?

enkore: Yep, if we decide to make the master index creation smarter then we'll need an extra sub method for that

ThomasWaldmann:
OK, looks good - merge?

enkore:
Did someone run the tests on BE recently? Otherwise looks "good to merge" to me as well (w/ squashed fixups).

ThomasWaldmann:
@enkore i am just booting the ppc qemu. :)

enkore:
Is there a command line that does that? The whole qemukvmvirtvagrant-thing is a bit of a mystery to me ;)

ThomasWaldmann:
https://gmplib.org/~tege/qemu.html

I used ppc (not: ppc64) and debian 8 ooc.

ThomasWaldmann:
```
Linux powerpc 3.16.0-4-powerpc #1 Debian 3.16.7-ckt25-2 (2016-04-08) ppc GNU/Linux
============ 418 passed, 102 skipped, 2 xpassed in 2959.98 seconds =============
```

enkore:
Thank you, my deb8 ppc is still at _Installieren des Grundsystems_ :D

ThomasWaldmann:
\o/

PR #892, opened 2016-04-12
compare_content: fix wrong added/removed file size

ThomasWaldmann:
bug: c[1] was not the size, but the byte at index 1 of the id bytestring!


Ape:
Thanks!

ThomasWaldmann:
@Ape changing tuples to namedtuples discovered this. ;)

PR #912, opened 2016-04-15
decrypt integrity test + some other tests

enkore:
test_symlink_extract doesn't do anything for the coverage, but verifies symlink extraction where there was no test previously.

test_decrypt_integrity tests the various kinds of integrity errors AESKeyBase.decrypt can/should raise; was not covered previously.

These commits are a bit older. Oops.

---

WIP. I'll make this a collective PR for some other tests I want to write. I.e. changeset constrained to borg/testsuite/ _only_.

TODO

- [x] passphrase tests
- [x] rebase
- [x] https://github.com/borgbackup/borg/issues/945#issuecomment-211770238
- [ ] archive, process_stdin has no coverage at all
- [ ] hardlink_master logic in archiverecreater has no coverage

### DIFF
--- a/borg/testsuite/key.py
+++ b/borg/testsuite/key.py
@@ -38,15 +38,10 @@ class TestKey:
         """))
     keyfile2_id = unhexlify('c3fbf14bc001ebcc3cd86e696c13482ed071740927cd7cbe1b01b4bfcee49314')
 
-    @pytest.fixture(autouse=True)
-    def keys_dir(self, request, monkeypatch):
-        tmp_keys_dir = tempfile.mkdtemp()
-
-        def finalize():
-            shutil.rmtree(tmp_keys_dir)
-
-        request.addfinalizer(finalize)
-        monkeypatch.setenv('BORG_KEYS_DIR', tmp_keys_dir)
+    @pytest.fixture
+    def keys_dir(self, request, monkeypatch, tmpdir):
+        monkeypatch.setenv('BORG_KEYS_DIR', tmpdir)
+        return tmpdir
 
     class MockRepository:
         class _Location:
@@ -62,7 +57,7 @@ class TestKey:
         assert hexlify(key.id_hash(chunk.data)) == b'2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae'
         assert chunk == key.decrypt(key.id_hash(chunk.data), key.encrypt(chunk))
 
-    def test_keyfile(self, monkeypatch):
+    def test_keyfile(self, monkeypatch, keys_dir):
         monkeypatch.setenv('BORG_PASSPHRASE', 'test')
         key = KeyfileKey.create(self.MockRepository(), self.MockArgs())
         assert bytes_to_long(key.enc_cipher.iv, 8) == 0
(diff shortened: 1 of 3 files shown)


borg/testsuite/key.py
```diff
@@ -101,3 +101,32 @@ def test_passphrase(self):
         data = b'foo'
         self.assert_equal(hexlify(key.id_hash(data)), b'818217cf07d37efad3860766dcdf1d21e401650fed2d76ed1d797d3aae925990')
         self.assert_equal(data, key2.decrypt(key2.id_hash(data), key.encrypt(data)))
+
+    def test_decrypt_integrity(self):
+        with open(os.path.join(os.environ['BORG_KEYS_DIR'], 'keyfile'), 'w') as fd:
+            fd.write(self.keyfile2_key_file)
+        os.environ['BORG_PASSPHRASE'] = 'passphrase'
```
ThomasWaldmann: iirc there was some environ context manager so one can use "with ...:" and it restores os.environ to previous state on exiting.

enkore: I'm currently moving this entire class to py.test tests + proper fixtures, which also avoids the side-effects of the other tests entirely.

borg/testsuite/key.py
```diff
@@ -101,3 +101,32 @@ def test_passphrase(self):
...
+        with self.assert_raises_regex(IntegrityError, 'Encryption envelope checksum mismatch'):
+            data = bytearray(self.keyfile2_cdata)
+            id = key.id_hash(data)
+            data[36] += 123  # this in the IV/CTR
+            key.decrypt(id, data)
+        with self.assert_raises_regex(IntegrityError, 'Encryption envelope checksum mismatch'):
+            data = bytearray(self.keyfile2_cdata)
+            id = key.id_hash(data)
+            data[50] += 1  # corrupt data
+            key.decrypt(id, data)
```
ThomasWaldmann: for even wider test coverage and as a shorter replacement of all above, you could use a loop to carpet bomb all of the envelope and see if in each case it raises this exception.

borg/testsuite/key.py
```diff
@@ -101,3 +101,32 @@ def test_passphrase(self):
...
+        with self.assert_raises_regex(IntegrityError, 'Chunk id verification failed'):
+            data = bytearray(self.keyfile2_cdata)
+            id = bytearray(key.id_hash(data))  # corrupt chunk id
+            id[12] = 0
+            key.decrypt(id, data)
```
ThomasWaldmann: same here for all of the compressed data.

borg/testsuite/key.py
ThomasWaldmann: ah, cool :)

borg/testsuite/key.py
enkore: Oh, that's a nice idea. A bit like very directed fuzzing.

borg/testsuite/key.py
```diff
@@ -31,14 +38,15 @@ class MockArgs:
...
-    def setUp(self):
-        self.tmppath = tempfile.mkdtemp()
-        os.environ['BORG_KEYS_DIR'] = self.tmppath
-        self.tmppath2 = tempfile.mkdtemp()
+    @pytest.fixture(autouse=True)
+    def keys_dir(self, request, monkeypatch):
+        tmp_keys_dir = tempfile.mkdtemp()
+
+        def finalize():
+            shutil.rmtree(tmp_keys_dir)
```
ThomasWaldmann: isn't there a ready-to-use tmpdir fixture that does this cleanup?

enkore: Yes, there is!

ThomasWaldmann:
one minor thing, see comment, otherwise fine to merge.

PR #914, opened 2016-04-15
Workround a bug in Linux fadvise FADV_DONTNEED, fixes #907

### LINKED ISSUE #907: fadvise DONTNEED needs to be page aligned
Looking at the kernel code, we can see that if we don't align the advise, some of it is disregarded:

```
        case POSIX_FADV_DONTNEED:
                if (!inode_write_congested(mapping->host))
                        __filemap_fdatawrite_range(mapping, offset, endbyte,
                                                   WB_SYNC_NONE);

                /*
                 * First and last FULL page! Partial pages are deliberately
                 * preserved on the expectation that it is better to preserve
                 * needed memory than to discard unneeded memory.
                 */
                start_index = (offset+(PAGE_SIZE-1)) >> PAGE_SHIFT;
                end_index = (endbyte >> PAGE_SHIFT);
```
I guess if you deal with a lot of small files, that might cause unnecessary cache pollution.

verygreen:
Despite what the man page says, Linux does not discard the initial
partial page only. The ending page would be truncated no matter if
it is partial or not.
Page-align the fadvise size to take care of this.
Also while we are at it, roll back initial fadvise offset to the
previous page boundary to actually throw away that page as we
no longer need it having read the second part now and the first
time in the previous call.

This patch has a noticeable impact in my Linux testing when the file
is on the rotating media. The total test runtime decreased by a bit
over 10%, but since over half of that time was actually cpu time,
the actual iowait time decreased around 20%.

### DIFF
--- a/borg/_chunker.c
+++ b/borg/_chunker.c
@@ -41,6 +41,7 @@ static uint32_t table_base[] =
 
 #define BARREL_SHIFT(v, shift) ( ((v) << shift) | ((v) >> (32 - shift)) )
 
+size_t pagemask;
 
 static uint32_t *
 buzhash_init_table(uint32_t seed)
@@ -130,6 +131,7 @@ chunker_fill(Chunker *c)
 {
     ssize_t n;
     off_t offset, length;
+    int overshoot;
     PyObject *data;
     memmove(c->data, c->data + c->last, c->position + c->remaining - c->last);
     c->position -= c->last;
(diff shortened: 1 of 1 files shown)


borg/_chunker.c
```diff
@@ -162,9 +163,16 @@ chunker_fill(Chunker *c)
         // complete cache with data that we only read once and (due to cache
         // size limit) kick out data from the cache that might be still useful
         // for the OS or other processes.
-        if (length > 0) {
-            posix_fadvise(c->fh, offset, length, POSIX_FADV_DONTNEED);
-        }
+        // We rollback the initial offset back to the start of the page,
+        // to avoid it not being truncated as a partial page request.
```
ThomasWaldmann: how about adding here:
mask = getpagesize() - 1

and reusing it at 2 places below?

borg/_chunker.c
```diff
@@ -162,9 +163,16 @@ chunker_fill(Chunker *c)
...
+        if (length > 0)
+            // Linux kernels prior to 4.7 have a bug where they truncate
+            // last partial page of POSIX_FADV_DONTNEED request, so we need
+            // to page-align it ourselves. We'll need the rest of this page
+            // on the next read (assuming this was not EOF)
+            overshoot = (offset + length) & (getpagesize() - 1);
+
+        posix_fadvise(c->fh, offset & (getpagesize() - 1), length - overshoot, POSIX_FADV_DONTNEED);
```
ThomasWaldmann: 2nd arg is wrong.

verygreen: ah, that's right, let me fix it

borg/_chunker.c
```diff
@@ -157,14 +159,33 @@ chunker_fill(Chunker *c)
         }
         length = c->bytes_read - offset;
         #if ( ( _XOPEN_SOURCE >= 600 || _POSIX_C_SOURCE >= 200112L ) && defined(POSIX_FADV_DONTNEED) )
+
+	// Only do it once per run.
+	if (pagemask == 0)
```
ThomasWaldmann: if you do not initialize it, is it guaranteed to be 0?

verygreen: yes. C standard guarantees that uninitialized static variables are zeroed.

PR #918, opened 2016-04-16
better readability and less errors with namedtuples, fixes #823

### LINKED ISSUE #823: use namedtuples

ThomasWaldmann:


borg/cache.py
```diff
@@ -375,12 +378,12 @@ def add_chunk(self, id, data, stats, overwrite=False):
         data = self.key.encrypt(data)
         csize = len(data)
         self.repository.put(id, data, wait=False)
-        self.chunks[id] = (refcount + 1, size, csize)
+        self.chunks[id] = ChunkIndexEntry(refcount + 1, size, csize)
```
enkore: Use `self.chunks.add(id, 1, size, csize)` instead

ThomasWaldmann: chunks.add does not set size and csize.
if we force overwrite, we likely have a reason, e.g. changed compression or encryption.
so the same id chunk may have different csize.

enkore: Change chunks.add to use size/csize from the supplied values and not what's already in the hashindex?

ThomasWaldmann: yes, merging this now.

PR #919, opened 2016-04-16
Fix capabilities extraction on Linux

enkore:
They are extracted correctly, for a little while at least, since chown()
*resets* call capabilities on the chowned file[1]. Which I find curious,
since chown() is a privileged syscall. Probably a safeguard for
sysadmins who are unaware of capabilities.

The solution is to set the xattrs last, after chowingn

[1] Even though I know this now I cannot find *any* documentation indicating this behaviour.


enkore:
Jesus GH, the base is 1.0-maint not master.

PR #920, opened 2016-04-16
Fix capabilities extraction on Linux

enkore:
They are extracted correctly, for a little while at least, since chown()
resets call capabilities on the chowned file[1]. Which I find curious,
since chown() is a privileged syscall. Probably a safeguard for
sysadmins who are unaware of capabilities.

The solution is to set the xattrs last, after chowing

[1] Even though I know this now I cannot find **any** documentation indicating this behaviour.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -372,17 +372,6 @@ Number of files: {0.stats.nfiles}'''.format(
             raise Exception('Unknown archive item type %r' % item[b'mode'])
 
     def restore_attrs(self, path, item, symlink=False, fd=None):
-        xattrs = item.get(b'xattrs', {})
-        for k, v in xattrs.items():
-            try:
-                xattr.setxattr(fd or path, k, v, follow_symlinks=False)
-            except OSError as e:
-                if e.errno not in (errno.ENOTSUP, errno.EACCES, ):
-                    # only raise if the errno is not on our ignore list:
-                    # ENOTSUP == xattrs not supported here
-                    # EACCES == permission denied to set this specific xattr
-                    #           (this may happen related to security.* keys)
-                    raise
         uid = gid = None
         if not self.numeric_owner:
             uid = user2uid(item[b'user'])
@@ -420,6 +409,19 @@ Number of files: {0.stats.nfiles}'''.format(
                 os.lchflags(path, item[b'bsdflags'])
             except OSError:
                 pass
+        # chown removes Linux capabilities, so set the extended attributes at the end, after chown, since they include
+        # the Linux capabilities in the "security.capability" attribute.
+        xattrs = item.get(b'xattrs', {})
+        for k, v in xattrs.items():
+            try:
+                xattr.setxattr(fd or path, k, v, follow_symlinks=False)
+            except OSError as e:
+                if e.errno not in (errno.ENOTSUP, errno.EACCES):
+                    # only raise if the errno is not on our ignore list:
+                    # ENOTSUP == xattrs not supported here
+                    # EACCES == permission denied to set this specific xattr
+                    #           (this may happen related to security.* keys)
+                    raise
 
     def rename(self, name):
         if name in self.manifest.archives:
--- a/borg/xattr.py
+++ b/borg/xattr.py
@@ -2,10 +2,12 @@
 """
 import errno
 import os
+import subprocess
 import sys
 import tempfile
 from ctypes import CDLL, create_string_buffer, c_ssize_t, c_size_t, c_char_p, c_int, c_uint32, get_errno
 from ctypes.util import find_library
+from distutils.version import LooseVersion
 
 from .logger import create_logger
 logger = create_logger()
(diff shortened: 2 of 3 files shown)


borg/archive.py
```diff
@@ -420,6 +409,17 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
                 os.lchflags(path, item[b'bsdflags'])
             except OSError:
                 pass
+        xattrs = item.get(b'xattrs', {})
+        for k, v in xattrs.items():
+            try:
+                xattr.setxattr(fd or path, k, v, follow_symlinks=True)
```
verygreen: Duh? Why the symlinks semantics change all of a sudden? Looks to introduce a potential race condition.

borg/archive.py
```diff
@@ -420,6 +409,17 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
                 os.lchflags(path, item[b'bsdflags'])
             except OSError:
                 pass
+        xattrs = item.get(b'xattrs', {})
```
ThomasWaldmann: 2 things we should have:
- a comment here that says it must be done after chown and why
- a unit test that fails before and succeeds after this cs

borg/archive.py
enkore: My bad.

enkore:
I'll write a test, but it won't fail before this changeset. While fakeroot allows setting capabilities and will retrieve them correctly later, it doesn't model the interaction of chown and capabilities correctly. Compare:

```
(_venv)[root@tiger xyz]# getfattr -d -m- usr/bin/ping 
# file: usr/bin/ping
security.capability=0sAQAAAgAgAAAAAAAAAAAAAAAAAAA=

(_venv)[root@tiger xyz]# chown mabe usr/bin/ping
(_venv)[root@tiger xyz]# getfattr -d -m- usr/bin/ping 
(_venv)[root@tiger xyz]# 
```

```
~/Projekte/_oss/borg> getfattr -d -m- somefile
# file: somefile
security.capability=0sAQAAAgAgAAAAAAAAAAAAAAAAAAA=

~/Projekte/_oss/borg> chown nobody somefile
~/Projekte/_oss/borg> getfattr -d -m- somefile
# file: somefile
security.capability=0sAQAAAgAgAAAAAAAAAAAAAAAAAAA=
```

enkore:
> I'll write a test, but it won't fail before this changeset.

Well, in fact it always fails, since the libc-detection in xattr.py circumvents the LD_PRELOAD set up by fakeroot. So `extract` can't set the caps anyway, since it would try to do so outside of fakeroot. That's also the reason for testsuite/archiver.py line 275:

```py
        if xattr.is_enabled(self.input_path):
            xattr.setxattr(os.path.join(self.input_path, 'file1'), 'user.foo', b'bar')
            # XXX this always fails for me
            # ubuntu 14.04, on a TMP dir filesystem with user_xattr, using fakeroot
            # same for newer ubuntu and centos.
            # if this is supported just on specific platform, platform should be checked first,
            # so that the test setup for all tests using it does not fail here always for others.
            # xattr.setxattr(os.path.join(self.input_path, 'link1'), 'user.foo_symlink', b'bar_symlink', follow_symlinks=False)
```

enkore:
> quite some changes to xattr.py

Less then expected. Now it actually works as a proper regression test (passes iff the change was applied). The change to xattr.py is also active in production. Which is probably a good thing, since it improves compatibility with fakeroot.

borg/archive.py
```diff
@@ -420,6 +409,19 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
                 os.lchflags(path, item[b'bsdflags'])
             except OSError:
                 pass
+        # chown removes call Linux capabilities, so set the extended attributes at the end, since they include
```
ThomasWaldmann: did you mean "chown call removes ..."?

borg/xattr.py
```diff
@@ -46,6 +46,10 @@ def get_all(path, follow_symlinks=True):
         logger.error(msg)
         raise Exception(msg)
 
+LD_PRELOAD = os.environ.get('LD_PRELOAD', '')
```
ThomasWaldmann: add a comment why you do this.

borg/xattr.py
```diff
@@ -46,6 +46,14 @@ def get_all(path, follow_symlinks=True):
         logger.error(msg)
         raise Exception(msg)
 
+# If we are running with fakeroot on Linux, then use the xattr functions of fakeroot. This is needed by
+# the 'test_extract_capabilities' test, but also allows xattrs to work with fakeroot on Linux in normal use.
+# TODO: Check whether fakeroot supports xattrs on all platforms supported below.
+# TODO: If that's the case then we can make Borg fakeroot-xattr-compatible on these as well.
+LD_PRELOAD = os.environ.get('LD_PRELOAD', '')
+if sys.platform.startswith('linux') and 'fakeroot' in LD_PRELOAD:
+    libc_name = LD_PRELOAD
```
ThomasWaldmann: ok, so fakeroot emulates libc by using LD_PRELOAD.

but does everything that is loaded by LD_PRELOAD also emulate libc (or at least the stuff we need in this python module)?

enkore: I checked the xattr tests for coverage, all functions are hit. There are some mentions in 2014 changelog entries that they fixed xattrs. I also checked fakeroot-ng, which does *not* support xattr's at all.

I'll add some checks for this.

enkore: > I'll add some checks for this.

Or not. Anything linking to the libc will have these functions around...

```
>>> import ctypes
>>> so = ctypes.CDLL("/usr/lib/libz.so")
>>> so.malloc
<_FuncPtr object at 0x7f2ae6489688>
>>> so.fsetxattr
<_FuncPtr object at 0x7f2ae6489750>
```

enkore:
Hm, I don't know why these tests fail. Either the fakeroot environment isn't effective, or the fakeroot used there doesn't have xattr support (e.g. fakeroot-ng). As noted above it's kinda hard (impossible?) do detect just with ctypes whether fakeroot re-defined the xattr functions or whether they came from libc...

enkore:
Okay, turns out that the fakeroot in Travis' Ubuntu is just old (0.18.2, released late 2011) and seems to have broken xattr support (changelog mentions fixing it later). So I added a version check of fakeroot to this test (and the using of libfakeroot in xattr.py). I see no way short of either pulling a new fakeroot into Travis (how?) or running the tests as real root (wat) to have the test work in Travis.

enkore:
I still can't find any *specification* of this behaviour, not even after reading the pages in the retracted POSIX draft. But maybe I didn't find the right chapter in those 400-ish pages.

Anyway, I traced the behaviour in the kernel. Here's what happens.

1. Syscall chown
2. Directed to chown_common (fs/open.c:567)
3. That sets ATTR_KILL_SUID, ATTR_KILL_SGID and ATTR_KILL_PRIV
3. and sends out a notify_change on the inode
4. That goes into Linux security (LSM) hooks
5. Finally into cap_inode_need_killpriv (since ATTR_KILL_PRIV is set)
6. That checks whether some caps are set on the inode and returns that
7. If cap_inode_need_killpriv was true, cap_inode_killpriv is called
8. cap_inode_killpriv deletes XATTR_NAME_CAPS == 'security.capabilities' from the inode

Note that this behaviour depends on the used LSM. But this one here is the commoncaps / default behaviour.

PR #925, opened 2016-04-16
Cache: update add_chunk to use ChunkIndex.add

enkore:
Also fixes a small cosmetic bug (would only affect recreate) where with add_chunk(overwrite=True) chunks that aren't unique would count towards the unique/deduplicated size.


ThomasWaldmann:
did you check all callers of ChunkIndex.add whether they want / don't want to update size/csize?

enkore:
The only direct user of this is the Cache. In add_chunk that's what we want, in the index building it's not relevant (in which order they are merged is dictionary order in thus random, hence, if there was any inconsistent csize data in the archives, the effect is random, independent of which side merge/add take the csize from). size is only different for either corruption / logic bugs or hash collisions.

PR #927, opened 2016-04-17
move some constants to new constants module

ThomasWaldmann:


Ape:
Could use something else than `from .constants import *`? Now it's hard to find the constant definitions if you are not familiar with the project structure.

ThomasWaldmann:
@Ape yes, this is a known problem of star imports and usually they are avoided because of this - but then, there's always grep or development docs.

I looked at options like `from . import constants`, but then you always have `constants.SOME_CONSTANT` which is much to type/read and leads to new line breaks.

Then `from . import constants as c` comes to mind, which shortens that to c.SOME_CONSTANT`, but is still somehow more complicated/ugly than just using `SOME_CONSTANT`.

Issue #208, opened 2015-09-19
make a better / real man page


ThomasWaldmann:
README.Debian from the debian package tells about quite some difficulties about creating a man page.

I think this should be solved upstream (== by us), also to make it easier for other linux distributions.

Debian packaging is discussed here:

https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1355849.html

The solution for creating the manpage should be automated.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/26801934-make-a-real-man-page)

anarcat:
more accurate URL for the packaging work is: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=792096

i was able to generate man pages by formatting argparse usage using some shims in monkeysign, initially inspired by:

https://andialbrecht.wordpress.com/2009/03/17/creating-a-man-page-with-distutils-and-optparse/

original source: https://crunchyfrog.googlecode.com/hg/utils/command/build_manpage.py

current implementation: https://sources.debian.net/src/monkeysign/2.0.2/monkeysign/documentation.py/#L13

this would need to be ported to argparse (easy) and eventually to click (harder, see https://github.com/mitsuhiko/click/issues/434)

anarcat:
i gave that my best shot, and argparse is just a nightmare to handle, particularly with our sub-parsers and everything. i would suggest we wait for the click port in #251 before spending any more time on this.

my current work is in my [man-builder branch](https://github.com/anarcat/borg/commits/man-builder) if anyone is masochistic enough to continue this nonsense.

LocutusOfBorg:
what about using rst2man?

rumpelsepp:
In syncthing we do it like this:

* http://docs.syncthing.net/users/syncthing.html
* https://github.com/syncthing/docs/blob/master/conf.py#L251-L291

anarcat:
@LocutusOfBorg would be certainly easier than the curent approach. however we'd need to split the manpages better: right now the `usage.rst.inc` files is all in one file, so we can't use rst2man to generate multiple manpages.

but i think it makes perfect sense, especially considering how messy #384 has become. we could revert all the mock and `have_cython` stuff and simply generate the `rst` files statically. then we'd have a much better control over the content of those files and could generate manpages, which would be readable directly online here (in github) as well.

@ThomasWaldmann would that satisfy the bounty you posted?

anarcat:
[...]

PR #928, opened 2016-04-17
Turn utf-8 string into Unicode string literal

bpereto:
Python2 has broken string handling, so when running sphinx-build under
python2 instead of python3 strings containing UTF-8 characters need to
be Unicode string literals.

Popped up during #208

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -44,7 +44,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'Borg - Deduplicating Archiver'
-copyright = '2010-2014 Jonas Borgström, 2015-2016 The Borg Collective (see AUTHORS file)'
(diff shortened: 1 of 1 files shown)


enkore:
This seems like a kludge to me to ease packaging or work around issues on a specific distribution with missing dependencies. If this is the case I'd say this should not be merged, but kept as a separate patch applied by the packager on the specific distribution needing this.

bpereto:
sure, a seperate patch for packaging is already applied.
and your argument is valid, borgbackup supports only python3.

ThomasWaldmann:
Yes, I agree this would be a bit strange as we have python 3.4 as minimum requirement and it is not needed there. OTOH, we could merge this now, as we dropped 3.2 which would not accept u'...' literals.

ThomasWaldmann:
ok, as you've patched it locally, I am closing this. if one runs sphinx based on python3, there is no such problem.

PR #932, opened 2016-04-17
create: Visit files in inode order

### LINKED ISSUE #905: Order of traversing files in a directory
Moving from #325 in a ticket of its own.

While looking around Archiver::_process I noticed that:
```
            if not dry_run:
                status = archive.process_dir(path, st)
            try:
                entries = os.listdir(path)
            except OSError as e:
                status = 'E'
            else:
                for filename in sorted(entries):
                    entry_path = os.path.normpath(os.path.join(path, filename))
                    self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
                                  keep_tag_files, skip_inodes, entry_path, restrict_dev,
```

I imagine the `sort(entries)` just sorts stuff alphabetically? If so, this is a pretty bad idea on a bunch of filesystems for large directories on spinning media. Affected filesystems include reiserfs, ext4 and probably many more that have static inode allocation.
The problem at hand is when you sort this list in an order different than inode numbers, every lstat that you do seeks the disk to a different inode table block somewhere which is really slow.
Not sorting the output is also not an option since then you get some "random" underlying hash sorting that also might not correspond to creation order. The most optimal strategy here for things with fixed inode tables is to sort by inode number (available in readdir output).
For filesystems with no static inode tables (like reiserfs) I imagine that would also lead to desired effect because it still is a good proxy for create time that leads to ondisk location (earlier created files get closer to start of disk in a chunk if they were written together).
More advanced COW filesystems like zfs and bttrfs - I am less sure, those might be all over the place depending on a bunch of stuff, though. I assume sort ordering would not matter there all that much.

A question was also raised if this makes any practical difference. And it certainly does for large directories (with millions of entries) that are not cached.
Here are some links to ext3 discussions about when htree code became popular and discussion arose. You'll see that performance impact is quite high:
https://bugzilla.kernel.org/show_bug.cgi?id=417
https://lkml.org/lkml/2008/2/18/263

There are tons more and the same issue is present in other filesystems with hashed names. I know about this because I am a Linux fs developer on my day job (and we met this when I was still working on reiserfs).

Additionally as we look outside of Linux (which is really optimized in these areas), other systems caches are (in some cases? in many cases) a lot more suboptimal or even non-exstent (e.g. in BSD where every fs driver used to need to implement their own cache).
I had a big confirmation of this yesterday. So I had a ~2M structure of files in on my Linux server, all on rotating media. find / -perm 0 (to cause stat calls) took 15 seconds there.
Now I also have this macbook I am working on, it has ssd and about 2.5M files on it, same find takes 2m50s. Quite a difference thanks to many-many variables I am sure.

enkore:
TODO

* [x] ~~Try to find a way to comment the changes in do_diff. Preferably human-parseable in polynomnial-deterministic time.~~ 404
* [x] Test on Py3.4
* [x] Test on Py3.4 with scandir
* [x] Test on Py3.5
* [x] Test on Py3.5 w/ fallback
* [x] Verify performance --- *should* be the same as the old patch (since that wasn't really changed), but not *confirmed* yet

Fixes #905

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -25,6 +25,7 @@ from .helpers import Error, location_validator, archivename_validator, format_ti
     log_multi, PatternMatcher, ItemFormatter
 from .logger import create_logger, setup_logging
 logger = create_logger()
+from . import helpers
 from .compress import Compressor, COMPR_BUFFER
 from .upgrader import AtticRepositoryUpgrader, BorgRepositoryUpgrader
 from .repository import Repository
@@ -247,17 +248,18 @@ class Archiver:
                     self.print_file_status(status, path)
                     continue
                 path = os.path.normpath(path)
+                try:
+                    st = os.lstat(path)
+                except OSError as e:
+                    self.print_warning('%s: %s', path, e)
+                    continue
                 if args.one_file_system:
-                    try:
-                        restrict_dev = os.lstat(path).st_dev
-                    except OSError as e:
-                        self.print_warning('%s: %s', path, e)
-                        continue
+                    restrict_dev = st.st_dev
                 else:
                     restrict_dev = None
                 self._process(archive, cache, matcher, args.exclude_caches, args.exclude_if_present,
                               args.keep_tag_files, skip_inodes, path, restrict_dev,
-                              read_special=args.read_special, dry_run=dry_run)
+                              read_special=args.read_special, dry_run=dry_run, st=st)
             if not dry_run:
                 archive.save(comment=args.comment, timestamp=args.timestamp)
                 if args.progress:
@@ -292,16 +294,16 @@ class Archiver:
 
     def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
                  keep_tag_files, skip_inodes, path, restrict_dev,
-                 read_special=False, dry_run=False):
+                 read_special=False, dry_run=False, st=None):
         if not matcher.match(path):
             self.print_file_status('x', path)
             return
-
-        try:
-            st = os.lstat(path)
-        except OSError as e:
-            self.print_warning('%s: %s', path, e)
-            return
+        if st is None:
+            try:
+                st = os.lstat(path)
+            except OSError as e:
+                self.print_warning('%s: %s', path, e)
+                return
         if (st.st_ino, st.st_dev) in skip_inodes:
             return
         # Entering a new filesystem?
@@ -331,15 +333,15 @@ class Archiver:
             if not dry_run:
                 status = archive.process_dir(path, st)
             try:
-                entries = os.listdir(path)
+                entries = helpers.scandir_inorder(path)
             except OSError as e:
                 status = 'E'
                 self.print_warning('%s: %s', path, e)
             else:
-                for filename in sorted(entries):
-                    entry_path = os.path.normpath(os.path.join(path, filename))
+                for dirent in entries:
+                    normpath = os.path.normpath(dirent.path)
                     self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
-                                  keep_tag_files, skip_inodes, entry_path, restrict_dev,
+                                  keep_tag_files, skip_inodes, normpath, restrict_dev,
                                   read_special=read_special, dry_run=dry_run)
         elif stat.S_ISLNK(st.st_mode):
             if not dry_run:
@@ -461,7 +463,7 @@ class Archiver:
                 return [None]
 
         def has_hardlink_master(item, hardlink_masters):
-            return item.get(b'source') in hardlink_masters and get_mode(item)[0] != 'l'
+            return stat.S_ISREG(item[b'mode']) and item.get(b'source') in hardlink_masters
 
         def compare_link(item1, item2):
             # These are the simple link cases. For special cases, e.g. if a
@@ -524,9 +526,6 @@ class Archiver:
             """
             changes = []
 
-            if item1.get(b'hardlink_master') or item2.get(b'hardlink_master'):
-                hardlink_masters[path] = (item1, item2)
-
             if has_hardlink_master(item1, hardlink_masters):
                 item1 = hardlink_masters[item1[b'source']][0]
 
@@ -559,8 +558,26 @@ class Archiver:
             print("{:<19} {}".format(line[1], line[0]))
 
         def compare_archives(archive1, archive2, matcher):
+            def hardlink_master_seen(item):
+                return b'source' not in item or not stat.S_ISREG(item[b'mode']) or item[b'source'] in hardlink_masters
+
+            def is_hardlink_master(item):
+                return item.get(b'hardlink_master', True) and b'source' not in item
+
+            def update_hardlink_masters(item1, item2):
+                if is_hardlink_master(item1) or is_hardlink_master(item2):
+                    hardlink_masters[item1[b'path']] = (item1, item2)
+
+            def compare_or_defer(item1, item2):
+                update_hardlink_masters(item1, item2)
+                if not hardlink_master_seen(item1) or not hardlink_master_seen(item2):
+                    deferred.append((item1, item2))
+                else:
+                    compare_items(output, item1[b'path'], item1, item2, hardlink_masters)
+
             orphans_archive1 = collections.OrderedDict()
             orphans_archive2 = collections.OrderedDict()
+            deferred = []
             hardlink_masters = {}
             output = []
 
@@ -569,31 +586,40 @@ class Archiver:
                     archive2.iter_items(lambda item: matcher.match(item[b'path'])),
             ):
                 if item1 and item2 and item1[b'path'] == item2[b'path']:
-                    compare_items(output, item1[b'path'], item1, item2, hardlink_masters)
+                    compare_or_defer(item1, item2)
                     continue
                 if item1:
                     matching_orphan = orphans_archive2.pop(item1[b'path'], None)
                     if matching_orphan:
-                        compare_items(output, item1[b'path'], item1, matching_orphan, hardlink_masters)
+                        compare_or_defer(item1, matching_orphan)
                     else:
                         orphans_archive1[item1[b'path']] = item1
                 if item2:
                     matching_orphan = orphans_archive1.pop(item2[b'path'], None)
                     if matching_orphan:
-                        compare_items(output, item2[b'path'], matching_orphan, item2, hardlink_masters)
+                        compare_or_defer(matching_orphan, item2)
                     else:
                         orphans_archive2[item2[b'path']] = item2
             # At this point orphans_* contain items that had no matching partner in the other archive
+            deleted_item = {
+                b'deleted': True,
+                b'chunks': [],
+                b'mode': 0,
+            }
             for added in orphans_archive2.values():
-                compare_items(output, added[b'path'], {
-                    b'deleted': True,
-                    b'chunks': [],
-                }, added, hardlink_masters, deleted=True)
+                path = added[b'path']
+                deleted_item[b'path'] = path
+                update_hardlink_masters(deleted_item, added)
+                compare_items(output, path, deleted_item, added, hardlink_masters, deleted=True)
             for deleted in orphans_archive1.values():
-                compare_items(output, deleted[b'path'], deleted, {
-                    b'deleted': True,
-                    b'chunks': [],
-                }, hardlink_masters, deleted=True)
+                path = deleted[b'path']
+                deleted_item[b'path'] = path
+                update_hardlink_masters(deleted, deleted_item)
+                compare_items(output, path, deleted, deleted_item, hardlink_masters, deleted=True)
+            for item1, item2 in deferred:
+                assert hardlink_master_seen(item1)
+                assert hardlink_master_seen(item2)
+                compare_items(output, item1[b'path'], item1, item2, hardlink_masters)
 
             for line in sorted(output):
                 print_output(line)
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -6,6 +6,7 @@ import grp
 import hashlib
 from itertools import islice
 import os
+import os.path
 import stat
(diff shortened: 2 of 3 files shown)


borg/archiver.py
```diff
@@ -294,16 +296,16 @@ def create_inner(archive, cache):
...
         if not matcher.match(path):
             self.print_file_status('x', path)
             return
-
-        try:
-            st = os.lstat(path)
-        except OSError as e:
-            self.print_warning('%s: %s', path, e)
-            return
+        if not st:
```
ThomasWaldmann: if st is None:

ThomasWaldmann: is there a caller that does not provide st?

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
 
     def acl_set(path, item, numeric_owner=False):
         pass
+
+# GenericDirEntry, scandir_generic (c) 2012 Ben Hoyt
+# from the python-scandir package (3-clause BSD license, just like us, so no troubles here)
+# note: simplified version
+
+
+class GenericDirEntry(object):
```
ThomasWaldmann: (object) not needed for us, we only do python3.

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+    __slots__ = ('name', '_stat', '_lstat', '_scandir_path', '_path')
+
+    def __init__(self, scandir_path, name):
+        self._scandir_path = scandir_path
+        self.name = name
+        self._stat = None
+        self._lstat = None
```
ThomasWaldmann: _stat and _lstat unused (yet?).

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        return os.lstat(self.path)
+
+    def is_dir(self, follow_symlinks=True):
+        assert not follow_symlinks
+        try:
+            st = self.stat(False)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+            return False  # Path doesn't exist or is a broken symlink#
```
ThomasWaldmann: typo # at EOL?

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        return stat.S_ISREG(st.st_mode)
+
+    def is_symlink(self):
+        try:
+            st = self.stat(False)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+            return False  # Path doesn't exist or is a broken symlink
+        return stat.S_ISLNK(st.st_mode)
```
ThomasWaldmann: somehow all is_xxx share most of the code?
just implement once and give the stat.S_XXX func as param?

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+
+    def inode(self):
+        st = self.stat(False)
+        return st.st_ino
+
+    def __str__(self):
+        return '<{0}: {1!r}>'.format(self.__class__.__name__, self.name)
```
ThomasWaldmann: no path?

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+    try:
+        # Try python-scandir on Python 3.4
+        from scandir import scandir
+    except ImportError:
+        # If python-scandir is not installed, then use a version that is just as slow as listdir.
+        scandir = scandir_generic
+
+
+def scandir_inorder(path='.'):
+    return sorted(scandir(path), key=lambda dirent: dirent.inode())
```
ThomasWaldmann: all the above has no platform dependency (yet?), so why it is in platform.py?

borg/archiver.py
enkore: `if st is None:` <= stat_results have no special bool() semantics. Either there is an object => True, or None => False.

borg/archiver.py
```diff
@@ -561,8 +563,21 @@ def print_output(line):
             print("{:<19} {}".format(line[1], line[0]))
 
         def compare_archives(archive1, archive2, matcher):
+            def hardlink_master_seen(item):
+                return b'source' not in item or not stat.S_ISREG(item[b'mode']) or item[b'source'] in hardlink_masters
+
+            def defer(item1, item2):
+                deferred.append((item1, item2))
```
ThomasWaldmann: inline this?

enkore:
I simplified GenericDirEntry further. I'll see about the tests. It's the diff test case again... just as I thought I resolved diff's ordering dependency.

borg/testsuite/archiver.py
```diff
@@ -1310,10 +1310,12 @@ def _test_recreate_interrupt(self, change_args, interrupt_early):
         assert 'file1' not in files
 
     def test_recreate_interrupt(self):
-        self._test_recreate_interrupt(False, True)
+        with patch.object(platform, 'scandir_inorder', platform.scandir_generic):
```
ThomasWaldmann: maybe add a src comment here why this patching is needed.

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        return stat.S_ISLNK(st.st_mode)
+
+    def inode(self):
+        st = self.stat(False)
+        return st.st_ino
+
+    def __str__(self):
+        return '<{0}: {1!r}>'.format(self.__class__.__name__, self.name)
+
+    __repr__ = __str__
```
ThomasWaldmann: hmm, iirc there is some automatic fallback from one to the other.

borg/archiver.py
ThomasWaldmann: this can go miserably wrong (maybe not in this case, but e.g. if len(sequence_object) == 0.

so better get used to "is None" if you mean it. ;)

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        self._lstat = None
+        self._path = None
+
+    @property
+    def path(self):
+        if self._path is None:
+            self._path = os.path.join(self._scandir_path, self.name)
+        return self._path
+
+    def stat(self, follow_symlinks=True):
```
verygreen: what's the function of follow_symlinks here?
it's not really used anywhere other than the assertion

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        assert not follow_symlinks
+        return os.lstat(self.path)
+
+    def is_dir(self, follow_symlinks=True):
+        assert not follow_symlinks
+        try:
+            st = self.stat(False)
```
verygreen: same here, also self.stat made me think this would translate into stat, not lstat.

But even more importantly, why have these two methods at all? they would be racy and therefore unsafe.

enkore: This is a shim implementation of the scandir interfaces; can't change the API here. Since we never want to follow symlinks I added the asserts and hardcoded lstat.

borg/archiver.py
enkore: I know, the "restrict_dev" debacle :D

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+            return False  # Path doesn't exist or is a broken symlink
+        return stat.S_ISREG(st.st_mode)
+
+    def is_symlink(self):
+        try:
+            st = self.stat(False)
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+            return False  # Path doesn't exist or is a broken symlink
```
verygreen: Nope. this is lstat inside of self.stat(), so broken symlink is ok.
The only errors possible here are:
path does not exist because it was removed (ENOENT) OR
path does not eist because somebody is playing with us and replaced some intermediate component with a broken symlink or looping symlink (ENOENT or ELOOP) OR
somebody changed permissions along the pathname (I don't think you can change permissions on the symlink itself) (EACCESS) OR
somebody changed the path sucn that some intermediate component is no longer a directory (ENOTDIR) OR
something broke badly and you are getting EIO.

And then on top of all those, there are various resource constraints like ENOMEM

borg/platform.py
```diff
@@ -14,3 +18,86 @@ def acl_get(path, item, st, numeric_owner=False):
...
+        return stat.S_ISLNK(st.st_mode)
+
+    def inode(self):
+        st = self.stat(False)
+        return st.st_ino
```
verygreen: huh? Why not get it from scandir instead?

enkore: > This is a shim implementation of the scandir interfaces; can't change the API here. Since we never want to follow symlinks I added the asserts and hardcoded lstat.

That's the code used if this is running on neither Python 3.5 (which includes scandir) or Python 3.4 with python-scandir installed.

borg/platform.py
enkore: seemed like the best fit. didn't want to add a new module for so few lines of code.

ThomasWaldmann: there is also helpers. ;)
which we will tear into pieces when the time has come.

borg/platform.py
```diff
@@ -1,3 +1,7 @@
+import errno
+import os
+import os.path
+import stat
```
ThomasWaldmann: guess that needs to be moved to helpers, too (and deduped)

enkore:
Performance testing on a nearly full (95 %) XFS file system on an older magnetic drive.

```
$ cat green-gen.c
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>
#include <string.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <linux/limits.h>

/* 1000000 */
#define NUMBER 50000

int main(int argc, char **argv)
{
    int i;
    char filename[PATH_MAX];

    if (argc != 2) {
        printf("Usage: Dir to create files in\n");
        return -1;
    }

    if (chdir(argv[1])) {
        perror("chdir");
        return -2;
    }

    for(i = 0 ; i <= NUMBER ; i++) {
        long rand;
        int fd;

        rand = random();

again:
        sprintf(filename, "%li", rand);
        fd = open(filename, O_CREAT|O_EXCL|O_WRONLY, 00600);
        if ( fd == -1) {
            /* Must be a double hit, increase and try again */
            rand++;
            goto again;
        }

        write(fd, &rand, sizeof(rand) * 2); /* some stack garbage too */
        close(fd);
    }

    return 0;
}
$ gcc green-gen.c -o green-gen
$ mkdir files
$ ./green-gen files
$ borg -V
borg 1.0.2
$ borg init repo --encryption none
$ sync ; echo 3 | sudo tee /proc/sys/vm/drop_caches
$ /bin/time borg create repo::1.0.2 files -svp --no-files-cache
Archive name: 1.0.2
Archive fingerprint: be63e68dd765cc2e561ea5866f0a21021b85aa26b4abe00636803cdae27e54e9
Time (start): Sun, 2016-04-17 23:23:54
Time (end):   Sun, 2016-04-17 23:32:02
Duration: 8 minutes 8.53 seconds
Number of files: 50001
This archive:                8.64 MB              8.79 MB              8.79 MB
All archives:                8.64 MB              8.79 MB              8.79 MB

Chunk index:                   50365                50365
30.90user 3.07system 8:09.30elapsed 6%CPU (0avgtext+0avgdata 43332maxresident)k
475700inputs+62248outputs (68major+15023minor)pagefaults 0swaps
$ borg -V
borg 1.0.3.dev105+ngf13aef5
$ sync ; echo 3 | sudo tee /proc/sys/vm/drop_caches
$ /bin/time borg create repo::inorder files -svp --no-files-cache
Archive name: inorder
Archive fingerprint: d5cd28c67adc50f6a7bc5494865c937f2cecde9a445bdda5c93c9661d9a26f26
Time (start): Sun, 2016-04-17 23:39:24
Time (end):   Sun, 2016-04-17 23:40:37
Duration: 1 minutes 13.03 seconds
Number of files: 50001
This archive:                9.49 MB              9.64 MB              8.69 MB
All archives:               18.13 MB             18.43 MB             17.48 MB

Chunk index:                   50802               100803
16.76user 1.04system 1:14.76elapsed 23%CPU (0avgtext+0avgdata 44452maxresident)k
476012inputs+57312outputs (55major+13563minor)pagefaults 0swaps
```

ThomasWaldmann:
a test on non-synthetic data on a rotating rust filesystem would be great.

enkore:
Kernel source tree? ;)

ThomasWaldmann:
i rather meant a filesystem that is in use for a while and contains a mix of data, naturally grown dirs (with inserts and deletes repeated for a while), etc.

enkore:
Meanwhile, for the 4.4.2 tree it brings about ~20 % improvement. It has ~52k files, but runs in approx 4 minutes (for master), which is a lot faster than 50k files in one dir, confirming the guess that many files in a directory aren't just poor for the FS but also poorly handled by Borg.

I'd say it behaves as expected / suspected. The more data per file / less files per dir the smaller the difference, with the worse case of many small-ish files in one / few dirs.

enkore:
Py 3.4 w/ scandir installed ok.

borg/helpers.py
```diff
@@ -1340,3 +1341,68 @@ def consume(iterator, n=None):
...
+
+    @property
+    def path(self):
+        if self._path is None:
+            self._path = os.path.join(self._scandir_path, self.name)
+        return self._path
+
+    def stat(self, follow_symlinks=True):
+        assert not follow_symlinks
+        return os.lstat(self.path)
```
ThomasWaldmann: is it intended that the stat result is not cached in the instance?

enkore: This fallback-code from scandir (where this comes from) doesn't do that, and we lstat later again (and sometimes, again) anyway.

PR #941, opened 2016-04-18
implement --keep-last N via --keep-secondly N, also --keep-minutely, fixes #537

### LINKED ISSUE #537: prune: keeping last X number of backups
`prune` currently allows one to specify to keep all backups created inside a time interval via `--keep-within` or a more complex regime via `-H -d -w -m -y`. But what is missing (and what I need right now) is the possibility to keep the last X number of backups disregarding any time interval: `--keep-last N`

ThomasWaldmann:
assuming that there is not more than 1 backup archive made in 1s,
--keep-last N and --keep-secondly N are equivalent.


frispete:
Very nice improvement, thanks.

PR #943, opened 2016-04-18
Merge/1.0 maint

enkore:


ThomasWaldmann:
In some minutes i have another one for 1.0-maint.

enkore:
Ok, will update this one then. Then we can fast-forward "windows" to master; don't need to articifally create more merge-conflicts than we'd have anyway... :)

PR #944, opened 2016-04-18
prune: fix bad validator, fixes #942

### LINKED ISSUE #942: mix of neg. & pos. --keep-xxx may trigger validation error
It is valid to give positive or negative numbers X for `--keep-TIMEUNIT X`.

A validator was implemented badly so that if you gave e.g. -1 once and +1 once (or in general: if the sum of them all was zero), it triggered a validation error telling you that you have to give at least one of these options.

Workaround: as any negative value means the same, just give another negative value so that the sum is not 0.

ThomasWaldmann:


enkore:
Regression test case?

ThomasWaldmann:
Well, testing for the fixed bug is a bit pointless as the sum is gone.

enkore:
Isn't that the point of regression testing? Making a test case that failed previously and is fixed by a patch.

ThomasWaldmann:
I assume that there is a test in python that tests that any() works correctly when given all-falsy values, so i don't need to check it here.

enkore:
Ok

ThomasWaldmann:
I am testing this locally right now and will merge when ok. travis seems to be sleeping.

PR #950, opened 2016-04-20
Add BORG_ASKPASS

maz:
Just like SSH has a `SSH_ASKPASS`, this pull request adds a `BORG_ASKPASS` environment variable to allow an external program to ask the user for a password when it is set.

### DIFF
--- a/borg/key.py
+++ b/borg/key.py
@@ -6,6 +6,7 @@ import sys
 import textwrap
 from hmac import compare_digest
 from hashlib import sha256, pbkdf2_hmac
+from subprocess import check_output, CalledProcessError
 
 from .helpers import Chunk, IntegrityError, get_keys_dir, Error, yes, bin_to_hex, CompressionDecider2, CompressionSpec
 from .logger import create_logger
@@ -205,7 +206,20 @@ class Passphrase(str):
 
     @classmethod
     def getpass(cls, prompt):
-        return cls(getpass.getpass(prompt))
+        askpass = os.environ.get('BORG_ASKPASS', None)
+        if askpass is None:
+            return cls(getpass.getpass(prompt))
+        else:
+            try:
+                return cls(
+                    check_output([askpass, prompt])
+                        .decode(sys.getfilesystemencoding(), 'surrogateescape')
+                        .strip()
+                )
+            except CalledProcessError:
+                # This follows what openssl does with non-zero exit codes on SSH_ASKPASS
+                #https://github.com/openssh/openssh-portable/blob/94141b7ade24afceeb6762a3f99e09e47a6c42b6/readpass.c#L152
+                return ""
 
     @classmethod
     def verification(cls, passphrase):
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -75,6 +75,11 @@ General:
         can either leave it away or abbreviate as `::`, if a positional parameter is required.
     BORG_PASSPHRASE
         When set, use the value to answer the passphrase question for encrypted repositories.
+    BORG_ASKPASS
+        When set, behaves like SSH_ASKPASS. The value will be interpreted as the path of a command
+        to execute. The password prompt will be passed as the only command-line argument, and stdout
+        will be interpreted as the password (once stripped of leading and trailing whitespace). The
+        command will be executed in lieu of the ordinary command-line password prompt.
     BORG_DISPLAY_PASSPHRASE
         When set, use the value to answer the "display the passphrase for verification" question when defining a new passphrase for encrypted repositories.
     BORG_LOGGING_CONF


borg/key.py
```diff
@@ -189,7 +190,11 @@ def env_passphrase(cls, default=None):
 
     @classmethod
     def getpass(cls, prompt):
-        return cls(getpass.getpass(prompt))
+        askpass = os.environ.get('BORG_ASKPASS', None)
+        if askpass is None:
+            return cls(getpass.getpass(prompt))
+        else:
+            return cls(check_output([askpass, prompt]).decode('UTF-8').strip())
```
ThomasWaldmann: utf-8 might work often on modern linux systems, but there is no guarantee that this is the correct encoding.

ThomasWaldmann:
can you tell more about the use case?

maz:
I'm making a GUI frontend for borg, and I'd like users to be able to enter their passwords not through the command-line, but through a GUI. While it is possible to pipe passwords in to stdin, and have borg interpret them, detecting when a password is requested would require analysis of stdout, and that solution seems to be overly complicated. Instead, I'm proposing following in the footsteps of ssh, which defines `SSH_ASKPASS`, which shells out to an external program to ask the user to enter a password. This would make implementing a GUI far simpler.

enkore:
You are looking for the BORG_PASSPHRASE environment variable. See http://borgbackup.readthedocs.org/en/stable/usage.html#environment-variables

borg/key.py
enkore: Since BORG_PASSPHRASE comes from os.environ the only consistent choice would be sys.getfilesystemencoding() + surrogateescape imho

maz:
`BORG_PASSPHRASE` serves a different purpose. That is designed for setting the passphrase _before_ borg runs. `BORG_ASKPASS` is designed to provide a different means of user interaction _while_ borg runs.

To quote the quickstart:
> For automated backups the passphrase can be specified using the BORG_PASSPHRASE environment variable.

but my use case is an _interactive_ backup, not an automated backup, and so `BORG_PASSPHRASE` does not seem like the correct solution.

enkore:
I don't quite see the difference here; Borg isn't really an interactive program. It does prompts, but they don't change the course of action, since they only provide information (passphrase) or confirmation *required* to continue with the operation already specified by the (at runtime) immutable command line.

maz:
The main reason why I can't just use `BORG_PASSPHRASE` is that if a user specifies a path to a repository, then the program has no way of knowing whether it should ask the user for a passphrase, since borg needs to look at the repo, before deciding whether a passphrase is needed.

enkore:
(I amended my comment above) -- that would be another valid argument for an askpass-like solution; at least in certain situations (software-structure wise).

ThomasWaldmann:
todo: fix coding issue, add docs and specification.

ThomasWaldmann:
@maz can you finish this one in the near future?

maz:
@ThomasWaldmann I should be able to finish it this Saturday. Sorry I haven't gotten to it sooner. I've been inundated with final exam studying :)

maz:
@ThomasWaldmann Just updated.

borg/key.py
maz: @enkore I've fixed this in the latest updates.

borg/key.py
```diff
@@ -205,7 +206,20 @@ def env_passphrase(cls, default=None):
...
+        askpass = os.environ.get('BORG_ASKPASS', None)
+        if askpass is None:
+            return cls(getpass.getpass(prompt))
+        else:
+            try:
+                return cls(
+                    check_output([askpass, prompt])
+                        .decode(sys.getfilesystemencoding(), 'surrogateescape')
+                        .strip()
+                )
```
ThomasWaldmann: please, there is no bonus for doing all in 1 line - esp. not if you need to break it into 5 due to its length. :-D

borg/key.py
```diff
@@ -205,7 +206,20 @@ def env_passphrase(cls, default=None):
...
+            except CalledProcessError:
+                # This follows what openssl does with non-zero exit codes on SSH_ASKPASS
+                #https://github.com/openssh/openssh-portable/blob/94141b7ade24afceeb6762a3f99e09e47a6c42b6/readpass.c#L152
+                return ""
```
ThomasWaldmann: hmm guess this is ok for create/extract - if there are issues, you'll get an empty password and it will fail to decrypt your key.

but for init - will it protect your key with an empty passphrase without you realizing it?

ThomasWaldmann:
the pep8 checker is also slightly unhappy.

btw, you can run tests locally by using: fakeroot -u tox

ThomasWaldmann:
I am shifting this to 1.2 milestone for now.

I might shift it back to 1.1 if it applies cleanly at the time and all issues are resolved.

enkore:
Would it make more sense to support the pinentry protocol?

ThomasWaldmann:
This is a rather stale PR. @maz are you still interested?

ThomasWaldmann:
closing this one, no progress.

PR #956, opened 2016-04-21
Added Salt example to documentation

icefoxen:
It has more or less the same functionality as the Ansible example, though isn't a line-by-line translation.

It's also tested on a Debian system, but if you want I can rewrite it for Arch; the only real change will be the install lines.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -154,6 +154,7 @@ This is a configuration similar to the one above, configured to be deployed with
 Salt running on a Debian system.
 
 ::
+
   Install borg backup from pip:
     pkg.installed: 
       - pkgs:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Hopefully we do not need to extend that to N different deployment systems...

Opinions about merging this?

edgewood:
Move both Ansible and Salt sections to a new deployment.rst in borgcommunity, and link them there?
-- 
Ed Blackman


-------- Original Message --------

enkore:
+1

ThomasWaldmann:
I am merging this now, but reserve the right to move all different deployment tools howtos out from main docs later maybe. ;)

PR #961, opened 2016-04-22
Fix Fedora installation instructions

scoobadog:
Update installation instructions so that the following error won't occur:

gcc: error: /usr/lib/rpm/redhat/redhat-hardened-cc1: No such file or directory
error: command 'gcc' failed with exit status 1

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -162,7 +162,7 @@ Install the dependencies with development headers::
     sudo dnf install libacl-devel libacl
     sudo dnf install lz4-devel
     sudo dnf install gcc gcc-c++
-    sudo dnf install redhat-rpm-config
+    sudo dnf install redhat-rpm-config                 # not needed in Korora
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@pguth @scoobadog iirc, the korora build worked without that. So, is that just needed for redhat and not korora?

pguth:
We didn't need that step while testing back then, so I guess it's specific to redhats' distribution. But I don't know what the current state is, sorry.

scoobadog:
@ThomasWaldmann @pguth I don't know about Korora, but it didn't seem to work with Fedora 23 without that package so maybe it's Red Hat specific. I'm going to test this with Korora 23 and see whether or not it needs it.

scoobadog:
Korora 23 doesn't need the `redhat-rpm-config` package. Should I add a comment to that line to indicate this?

enkore:
Yes, no need to install it when it's not needed :)

ThomasWaldmann:
thanks for fixing!

PR #965, opened 2016-04-22
borg mount: cache partially read data chunks

enkore:
This fixes performance issues of FUSE mounted archives with programs that do small reads.

An extreme case are the shaXXXsum family of programs which read 32 kB blocks. This leads to chunks being decrypted and decompressed excessively often, the slowdown for these approaches a factor 60. This patch caches partially read chunks in a bounded cache with LRU-eviction (LRUCache), which brings md5sum and the like to the speed of the native "borg list" functionality.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1533,6 +1533,11 @@ class Archiver:
 
         To allow a regular user to use fstab entries, add the ``user`` option:
         ``/path/to/repo /mnt/point fuse.borgfs defaults,noauto,user 0 0``
+
+        The BORG_MOUNT_DATA_CACHE_ENTRIES environment variable is meant for advanced users
+        to tweak the performance. It sets the number of cached data chunks; additional
+        memory usage can be up to ~8 MiB times this number. The default is the number
+        of CPU cores.
         """)
         subparser = subparsers.add_parser('mount', parents=[common_parser], add_help=False,
                                           description=self.do_mount.__doc__,
--- a/borg/fuse.py
+++ b/borg/fuse.py
@@ -8,8 +8,12 @@ import tempfile
 import time
 from .archive import Archive
 from .helpers import daemonize, bigint_to_int
+from .logger import create_logger
+from .lrucache import LRUCache
 from distutils.version import LooseVersion
 import msgpack
+
+logger = create_logger()
 
 # Does this version of llfuse support ns precision?
 have_fuse_xtime_ns = hasattr(llfuse.EntryAttributes, 'st_mtime_ns')
@@ -54,6 +58,9 @@ class FuseOperations(llfuse.Operations):
         self.pending_archives = {}
         self.accounted_chunks = {}
         self.cache = ItemCache()
+        data_cache_capacity = int(os.environ.get('BORG_MOUNT_DATA_CACHE_ENTRIES', os.cpu_count() or 1))
+        logger.debug('mount data cache capacity: %d chunks', data_cache_capacity)
+        self.data_cache = LRUCache(capacity=data_cache_capacity, dispose=lambda _: None)
         if archive:
             self.process_archive(archive)
         else:
@@ -229,7 +236,16 @@ class FuseOperations(llfuse.Operations):
                 offset -= s
                 continue
             n = min(size, s - offset)
-            _, data = self.key.decrypt(id, self.repository.get(id))
+            if id in self.data_cache:
+                data = self.data_cache[id]
+                if offset + n == len(data):
+                    # evict fully read chunk from cache
+                    del self.data_cache[id]
+            else:
+                _, data = self.key.decrypt(id, self.repository.get(id))
+                if offset + n < len(data):
+                    # chunk was only partially read, cache it
+                    self.data_cache[id] = data
             parts.append(data[offset:offset + n])
             offset = 0
             size -= n


enkore:
```

$ dd if=/dev/urandom of=test64M bs=1M count=64 status=progress
62914560 bytes (63 MB, 60 MiB) copied, 3.01193 s, 20.9 MB/s
64+0 records in
64+0 records out
67108864 bytes (67 MB, 64 MiB) copied, 3.21169 s, 20.9 MB/s
$ borg init mountrepo
Enter new passphrase: 
Enter same passphrase again: 
Do you want your passphrase to be displayed for verification? [yN]:
$ borg create mountrepo::archive test64M -svp
Archive name: archive
Archive fingerprint: 3c9af882ef9fd605a1fe345c574e922f2a7908d9b919858346a3b92df8820505
Time (start): Fri, 2016-04-22 22:16:21
Time (end):   Fri, 2016-04-22 22:16:24
Duration: 3.14 seconds
Number of files: 1
This archive:               67.11 MB             67.11 MB             67.11 MB
All archives:               67.11 MB             67.11 MB             67.11 MB

Chunk index:                      28                   28
$ mkdir mount
$ borg -V
borg 1.0.3.dev133+ng282eb69
$ borg mount mountrepo::archive mount
$ /bin/time md5sum mount/test64M
98a368b703375d8f0a12c2de0af57946  mount/test64M
0.33user 0.02system 1:05.88elapsed 0%CPU (0avgtext+0avgdata 2456maxresident)k
131152inputs+0outputs (1major+135minor)pagefaults 0swaps
$ # ~1 MB/s
$ # compare to max-speed Borg could do:
$ /bin/time borg list mountrepo::archive --format '{md5}  {path}{LF}'
98a368b703375d8f0a12c2de0af57946  test64M
2.25user 0.07system 0:02.34elapsed 99%CPU (0avgtext+0avgdata 62048maxresident)k
0inputs+32outputs (0major+19137minor)pagefaults 0swaps

$ # patched version
$ borg -V
1.0.3.dev117+ng53182d8
$ borg mount mountrepo::archive mount
$ /bin/time md5sum mount/test64M
98a368b703375d8f0a12c2de0af57946  mount/test64M
0.15user 0.02system 0:02.01elapsed 8%CPU (0avgtext+0avgdata 2532maxresident)k
131072inputs+0outputs (0major+138minor)pagefaults 0swaps
$ # bwem
```

enkore:
We could also evict cached-then-fully-read chunks from the LRU cache without filling it up. That's a mixed thing, though, it would improve performance when there are multiple parallel "read streams" (since they wouldn't evict the chunks of one another), but would decrease it for internally-deduplicated stuff. I think let's leave it at this.

borg/fuse.py
```diff
@@ -54,6 +55,7 @@ def __init__(self, key, repository, manifest, archive, cached_repo):
         self.pending_archives = {}
         self.accounted_chunks = {}
         self.cache = ItemCache()
+        self.data_cache = LRUCache(capacity=20, dispose=id)
```
ThomasWaldmann: that's up to 160MB (which is significant for smaller devices like rpi).
it sounded like a major speedup could be already achieved with 1 (or few) cached chunks.

enkore: how about number_of_cores for the bound of this cache. and combine it with the explicit cache eviction i described above. would have a very reasonable memory footprint and handle parallel streams w/o troubles.

ThomasWaldmann:
isn't our fuse code single-threaded anyway?

enkore:
Sure, but you can still access the file system with multiple applications concurrently.

borg/archiver.py
```diff
@@ -1533,6 +1533,11 @@ def build_parser(self, args=None, prog=None):
 
         To allow a regular user to use fstab entries, add the ``user`` option:
         ``/path/to/repo /mnt/point fuse.borgfs defaults,noauto,user 0 0``
+
+        The BORG_MOUNT_DCACHE_ENTRIES environment variable is meant for advanced users
+        to tweak the performance. It sets the number of cached data chunks; additional
+        memory usage can be up to ~8 MiB this number. The default is the number of CPU
```
ThomasWaldmann: 8MB times this number?

enkore: ty

borg/archiver.py
```diff
@@ -1533,6 +1533,11 @@ def build_parser(self, args=None, prog=None):
 
         To allow a regular user to use fstab entries, add the ``user`` option:
         ``/path/to/repo /mnt/point fuse.borgfs defaults,noauto,user 0 0``
+
+        The BORG_MOUNT_DCACHE_ENTRIES environment variable is meant for advanced users
```
ThomasWaldmann: isn't dcache entries kernel speak for directory cache entries?

borg/lrucache.py
```diff
@@ -26,7 +30,10 @@ def __delitem__(self, key):
         self._lru.remove(key)
 
     def __contains__(self, key):
-        return key in self._cache
+        contains = key in self._cache
+        self._hits += contains
+        self._misses += not contains
```
ThomasWaldmann: why is this here? usually one would expect that *getting* data is a hit and *not getting* data (from cache) is a miss. but here, simply checking is enough to increase the counters.

ThomasWaldmann:
issue #966

borg/lrucache.py
enkore: I was about to say that client code would prefer `if x in cache: y = cache[x]`, but that's not the case for Repository. I'm just going to stash this, OT anyway.

borg/archiver.py
enkore: Ha, I thought about icache/dcaches of cpus, not the Kernel, but yeah, since this is an FS it's confusing.

borg/fuse.py
```diff
@@ -54,6 +58,9 @@ def __init__(self, key, repository, manifest, archive, cached_repo):
         self.pending_archives = {}
         self.accounted_chunks = {}
         self.cache = ItemCache()
+        data_cache_capacity = int(os.environ.get('BORG_MOUNT_DATA_CACHE_ENTRIES', os.cpu_count() or 1))
+        logger.debug('data cache capacity: %d chunks', data_cache_capacity)
```
ThomasWaldmann: mount data cache capacity:

borg/fuse.py
```diff
@@ -54,6 +58,9 @@ def __init__(self, key, repository, manifest, archive, cached_repo):
...
+        self.data_cache = LRUCache(capacity=data_cache_capacity, dispose=id)
```
ThomasWaldmann: why did you use id here? How about:  `dispose=lambda _: None`

ThomasWaldmann:
other than the 2 minor things i found, this seems ready for merging.

PR #969, opened 2016-04-23
docs/installation: Add FreeBSD Ports tree

rumpelsepp:
https://svnweb.freebsd.org/ports?view=revision&revision=413469

It is not yet in "pkg"; that will automatically happen in the next quarterly period.


ThomasWaldmann:
thanks!

PR #970, opened 2016-04-23
Add power-on self-test

enkore:
These run in ~100 ms here, so even on much slower machines (where also
Python setup will be slower) it shouldn't be noticeable at all. (starting Borg takes ~1.5-2 s)

On the QEMU-emulated ppc machine they take ~3.5 s while starting Borg takes ~15 s.

An explicit SELFTEST_COUNT is kept around to make sure that all the tests were properly discovered. With some of the bugs we have been seeing lately I'm starting to become paranoid :D 

~~--debug is maybe a bit too verbose about this. Yeah, let's make it less verbose; that the successful tests were successful and discovered are asserted in other ways.~~ (done)

Check: does it run for every cmd() in the testsuite? if so, patch that out for the testsuite; don't need to waste time in CI/tests.

### DIFF
--- a/borg/selftest.py
+++ b/borg/selftest.py
@@ -1,0 +1,79 @@
+"""
+Self testing module
+===================
+
+The selftest() function runs a small test suite of relatively fast tests that are meant to discover issues
+with the way Borg was compiled or packaged and also bugs in Borg itself.
+
+Theses tests are a subset of the borg/testsuite and are run with Pythons built-in unittest, hence none of
+the tests used for this can or should be ported to py.test currently.
+
+To assert that self test discovery works correctly the number of tests is kept in the SELFTEST_COUNT
+variable. SELFTEST_COUNT must be updated if new tests are added or removed to or from any of the tests
+used here.
+"""
+
+
+import sys
+import time
+from unittest import TestResult, TestSuite, defaultTestLoader
+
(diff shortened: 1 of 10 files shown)


borg/selftest.py
```diff
@@ -0,0 +1,63 @@
+import sys
+import time
+from unittest import TestResult, TestSuite, defaultTestLoader
```
ThomasWaldmann: hmm, do we want to depend on unittest for this?
i rather thought we would like to get rid of it and replace everything by py.test.
but ofc we would not depend on py.test for production.

enkore: true. CryptoTestCase port to pytest (using pytest fixtures even) is in #912, which would take that off the list (emulating pytest is not an option; needing it at runtime is meh). The others wouldn't really profit from pytest's fixtures, they don't need that. I'd say:

* scratch CryptoTestCase from the list (with #912, not here)
* POST test cases stay with unittest for now

ThomasWaldmann: then these test cases maybe should have a comment that they must stay with unittest because of borg.selftest. and they also need to get moved to a module that does not import py.test or it will fail on users' systems.

enkore: Done. Also a bit of why/what/hu? text.

Issue #35, opened 2015-05-23
pull code from attic gh org merge-all branch


ThomasWaldmann:
this is not yet in borg repo.

maybe create some "experimental" branch for it.

ThomasWaldmann:
it is now in "experimental" branch.

PR #975, opened 2016-04-23
check: support integrity verification

enkore:


borg/archive.py
```diff
@@ -730,6 +742,24 @@ def identify_key(self, repository):
         cdata = repository.get(next(self.chunks.iteritems())[0])
         return key_factory(repository, cdata)
 
+    def integrity_verification(self):
+        logger.info('Starting integrity verification...')
```
ThomasWaldmann: well, a lot what check does is somehow integrity verification, so we could say "chunk data integrity verification".

borg/archive.py
```diff
@@ -730,6 +742,24 @@ def identify_key(self, repository):
...
+                continue
+            encrypted_data = self.repository.get(chunk_id)
+            try:
+                _, data = self.key.decrypt(chunk_id, encrypted_data)
+            except IntegrityError as integrity_error:
+                self.error_found = True
+                logger.error('chunk %s, integrity error: %s', hexlify(chunk_id), integrity_error)
+            count += 1
+        pi.finish()
+        logger.info('Finished integrity verification, verified %d chunks.', count)
```
ThomasWaldmann: you say "verified" here, that could be misunderstood as "successfully verified" and contradict the error msgs above.

borg/archiver.py
```diff
@@ -182,12 +182,16 @@ def do_check(self, args, repository):
...
+            self.print_error("--repository-only and --full contradict each other. Please select one.")
+            return EXIT_ERROR
         if not args.archives_only:
             if not repository.check(repair=args.repair, save_space=args.save_space):
                 return EXIT_WARNING
         if not args.repo_only and not ArchiveChecker().check(
                 repository, repair=args.repair, archive=args.location.archive,
-                last=args.last, prefix=args.prefix, save_space=args.save_space):
+                last=args.last, prefix=args.prefix,integrity_verification=args.verify_data,
+                save_space=args.save_space):
```
ThomasWaldmann: param name could be verify_data also.

borg/archiver.py
```diff
@@ -182,12 +182,16 @@ def do_check(self, args, repository):
             if not yes(msg, false_msg="Aborting.", truish=('YES', ),
                        env_var_override='BORG_CHECK_I_KNOW_WHAT_I_AM_DOING'):
                 return EXIT_ERROR
+        if args.repo_only and args.verify_data:
+            self.print_error("--repository-only and --full contradict each other. Please select one.")
```
ThomasWaldmann: you mean --verify-data.

borg/archive.py
ThomasWaldmann: you could also maybe use .report() instead of logger.error and setting error_found.

borg/archive.py
enkore: Changed to "cryptographic data integrity verification" (even none/plaintext has cryptographic IDs) to be clear about what's doing now.

borg/archive.py
enkore: report() is internal to rebuild_refcounts; it uses a format that explicitly references the chunk position inside an item which does not apply here.

I changed the final information message to include total and errored chunk counts which should avoid any ambiguity.

enkore:
Thanks. I'll update this PR later; I want a proper test for this.

borg/archive.py
```diff
@@ -741,6 +754,26 @@ def identify_key(self, repository):
...
+            encrypted_data = self.repository.get(chunk_id)
+            try:
+                _, data = self.key.decrypt(chunk_id, encrypted_data)
+            except IntegrityError as integrity_error:
+                self.error_found = True
+                errors += 1
+                logger.error('chunk %s, integrity error: %s', hexlify(chunk_id), integrity_error)
+            count += 1
+        pi.finish()
+        logger.info('Finished cryptographic data integrity verification, verified %d chunks with %d integrity errors.',
```
ThomasWaldmann: if errors > 0, this should be on ERROR log level.

borg/archiver.py
```diff
@@ -1229,6 +1245,9 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('--archives-only', dest='archives_only', action='store_true',
                                default=False,
                                help='only perform archives checks')
+        subparser.add_argument('--verify-data', dest='verify_data', action='store_true',
+                               default=False,
+                               help='perform full repository integrity verification (conflicts with --repository-only)')
```
ThomasWaldmann: hmm, this description is a bit conflicting also. :)

shall we rather say "perform full archive data integrity verification"?

enkore: It's now "perform cryptographic archive data integrity verification" (consistent with the log messages)

borg/archive.py
```diff
@@ -771,8 +772,10 @@ def verify_data(self):
                 logger.error('chunk %s, integrity error: %s', hexlify(chunk_id), integrity_error)
             count += 1
         pi.finish()
-        logger.info('Finished cryptographic data integrity verification, verified %d chunks with %d integrity errors.',
-                    count, errors)
+        log = logger.info
+        if errors:
+            log = logger.error
```
ThomasWaldmann: log = logger.error if errors else logger.info

enkore:
I will think again how a simple test for this could be written.

Maybe monkeypatch zlib.crc32 to be ineffective, instead of adding the necessary math code (which is quite a bit, and surprisingly complex) to actually defeat it.

borg/archive.py
```diff
@@ -772,9 +772,7 @@ def verify_data(self):
                 logger.error('chunk %s, integrity error: %s', hexlify(chunk_id), integrity_error)
             count += 1
         pi.finish()
-        log = logger.info
-        if errors:
-            log = logger.error
+        log = logger.error if errors else logger.error
```
ThomasWaldmann: do you want some SPAM or rather some SPAM? :D

enkore:
Test is there. I wasn't really thinking this through; the CRC32 validation of the Repository doesn't matter, I just overwrite the object normally and correctly; the Repository updates the CRC32 for me ;)

borg/testsuite/archiver.py
```diff
@@ -1590,6 +1590,22 @@ def test_extra_chunks(self):
...
+        with repository:
+            for item in archive.iter_items():
+                if item[b'path'].endswith('testsuite/archiver.py'):
+                    chunk = item[b'chunks'][-1]
+                    data = repository.get(chunk.id) + b'1234'
+                    repository.put(chunk.id, data)
+                    break
+            repository.commit()
+        self.cmd('check', self.repository_location, exit_code=0)
+        self.cmd('check', '--verify-data', self.repository_location, exit_code=1)
```
ThomasWaldmann: hmm, if it outputs the filename of the problematic file, the test could check whether it complains about the corrupt file.

enkore: It doesn't give the file name (one chunk is used by potentially many files across many archives). Hmm, print the refcount of corrupted chunks?

ThomasWaldmann: Ah, right. Maybe leave now as is.

It could be later extended to make a pass over all archives and print archives / files that reference the defect chunks.

borg/archive.py
```diff
@@ -741,6 +755,26 @@ def identify_key(self, repository):
...
+            pi.show()
+            if not refcount:
+                continue
+            encrypted_data = self.repository.get(chunk_id)
+            try:
+                _, data = self.key.decrypt(chunk_id, encrypted_data)
+            except IntegrityError as integrity_error:
+                self.error_found = True
+                errors += 1
+                logger.error('chunk %s, integrity error: %s', hexlify(chunk_id), integrity_error)
```
ThomasWaldmann: hmm, hexlify() result is bytes, right? use bin_to_hex()? check other places also.

borg/archiver.py
```diff
@@ -1213,6 +1217,18 @@ def build_parser(self, args=None, prog=None):
           required).
         - The archive checks can be time consuming, they can be skipped using the
           --repository-only option.
+
+        The --verify-data option will perform a full integrity verification (as opposed to
+        checking the CRC32 of the segment) of data, which means reading the data from the
+        repository, decrypting and decompressing it. This is a cryptographically secure
+        verification for encrypted repositories and will detect any manipulation of the
+        data with a probability of ~1.
```
ThomasWaldmann: this leaves somehow the question open what happens with unencrypted repos / what the assurance is for them.

ThomasWaldmann:
a few comments, otherwise fine to merge.

borg/archiver.py
enkore: An unencrypted repo isn't tamper resistant.

It brings the essentially same guarantees about corruption detection to the table, though (SHA-256 for the IDs, which are checked by the Key as usual; might add an explicit test for that case).

ThomasWaldmann: i meant maybe mention that in the text.
it's not all about tampering, more safely detecting accidental damage than with crc32 might also be worth mentioning.

borg/archive.py
enkore: Good catch

Alexander-Shukaev:
How does `check --verify-data` compare to `extract --dry-run` as per #1022?  I see that `borgmatic` has dedicated `extract` check for this as per [#35](https://projects.torsion.org/witten/borgmatic/issues/35), and the requester claims that `extract --dry-run` is

> the only way to check for bit rot without doing a real restore

which appears to be a strange statement in 2017-07, when `check --verify-data` had been supposedly released already.

ThomasWaldmann:
@Alexander-Shukaev yeah, guess there is some overlap when one wants to do a full verification of the content stored in an archive. Both verify all contained content data.

The difference of course is that the check-based option is rather focused on checking (and expecting to detect [and fix, with --repair]) damaged repos/archives while the extract-based option is extremely close to what would happen if you actually extracted an archive (with writing contents to disk), but just does not write contents to disk.

Alexander-Shukaev:
@ThomasWaldmann, now I'm curious, in case of the pure check scenario (without any repair expectations), when one is merely interested to get an answer to the question "Is the lastest archive consistent and restorable?" (which is probably the most widespread and important question for the users, isn't it?), then which of the two `extract --dry-run` applied to lastest archive or `check --verify-data` is faster?  My gut feeling tells me that it should be `extract` because it only considers the last archive.

ThomasWaldmann:
if you use "borg check --verify-data", it will do:
- a low level repo check (crc32 in all segment files, rebuild repo index in memory, ...)
- a high level archives check (default: all archives)

So dry-run extract of last archive can be much faster, depending on how big your repo is and how many archives you have.

PR #976, opened 2016-04-23
fixes Issue 974

ThomasWaldmann:


borg/helpers.py
```diff
@@ -658,7 +662,7 @@ def format_archive(archive):
     return '%-36s %s [%s]' % (
         archive.name,
         format_time(to_localtime(archive.ts)),
-        hexlify(archive.id).decode('ascii'),
+        bin_to_hex(archive.id),
```
enkore: archive.fpr ?

borg/remote.py
```diff
@@ -191,6 +191,10 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.rollback()
         self.close()
 
+    @property
+    def id_str(self):
+        return bin_to_hex(self.id)
```
enkore: Hmmm, maybe just `__str__`? This ID is the only kind of string representation that would make sense for Manifest (and Repository)

ThomasWaldmann: __str__ is used at some places for a much more verbose representation that is used in screen output. archive and stats come to mind.

borg/helpers.py
ThomasWaldmann: i first had that and it failed - because "archive" is in reality an ArchiveInfo instance.

ThomasWaldmann:
merge?

PR #981, opened 2016-04-25
Items refactoring

ThomasWaldmann:
Item class + tests for now, WIP.

This is just the isolated item class + tests for now.
When this is merged, there needs to be a widespread refactoring to actually use this.

### DIFF
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -681,11 +681,15 @@ def posix_acl_use_stored_uid_gid(acl):
 
 def safe_decode(s, coding='utf-8', errors='surrogateescape'):
     """decode bytes to str, with round-tripping "invalid" bytes"""
+    if s is None:
+        return None
     return s.decode(coding, errors)
 
 
 def safe_encode(s, coding='utf-8', errors='surrogateescape'):
     """encode str to bytes, with round-tripping "invalid" bytes"""
+    if s is None:
+        return None
     return s.encode(coding, errors)
 
 
--- a/borg/item.py
+++ b/borg/item.py
@@ -17,6 +17,8 @@ class PropDict:
     """
     VALID_KEYS = None  # override with <set of str> in child class
 
+    __slots__ = ("_dict", )  # avoid setting attributes not supported by properties
+
     def __init__(self, data_dict=None, **kw):
         if data_dict is None:
             data = kw
@@ -39,7 +41,7 @@ class PropDict:
 
     def as_dict(self):
         """return the internal dictionary"""
-        return self._dict  # XXX use StableDict?
+        return StableDict(self._dict)
 
     def _check_key(self, key):
         """make sure key is of type str and known"""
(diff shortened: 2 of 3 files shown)


borg/item.py
```diff
@@ -0,0 +1,137 @@
...
+            if decode is not None:
+                value = decode(value)
+            return value
+
+        def _set(self, value):
+            if not isinstance(value, value_type):
+                raise TypeError(type_error_msg)
+            if encode is not None:
+                value = encode(value)
+            self._dict[key] = value
```
enkore: Do _check_key here to avoid setting invalid attributes?

borg/item.py
```diff
@@ -0,0 +1,137 @@
...
+                raise TypeError("dict keys must be str or bytes, not %r" % k)
+            _dict[k] = v
+        unknown_keys = set(_dict) - self.VALID_KEYS
+        if unknown_keys:
+            raise ValueError("dict contains unknown keys %s" % ','.join(unknown_keys))
+        self._dict = _dict
+
+    def as_dict(self):
+        """return the internal dictionary"""
+        return self._dict  # XXX use StableDict?
```
enkore: Since msgpack doesn't order keys (afaik)... yes? Hash-order in the item-stream would break metadata deduplication.

borg/helpers.py
```diff
@@ -723,11 +723,15 @@ def posix_acl_use_stored_uid_gid(acl):
...
+    if s is None:
+        return None
     return s.decode(coding, errors)
 
 
 def safe_encode(s, coding='utf-8', errors='surrogateescape'):
     """encode str to bytes, with round-tripping "invalid" bytes"""
+    if s is None:
+        return None
     return s.encode(coding, errors)
```
enkore: Why these if-None?

enkore:
I think this should use `__slots__ = ("_dict",)`; not for performance reasons, but to avoid setting invalid attributes that would silently fail to serialize via setattr and it's syntactic sugar.

```py
>>> from borg.item import Item
>>> item = Item()
>>> item.some_invalid_key = 1234
>>> serialized = item.as_dict()
>>> serialized
{}
```

vs

```py
>>> from borg.item import Item
>>> item = Item()
>>> item.some_invalid_key = 1234
AttributeError: 'Item' object has no attribute 'some_invalid_key'
```

borg/helpers.py
ThomasWaldmann: some place of borg sets a string key to a None value.

ThomasWaldmann:
ok, good idea.

borg/item.py
ThomasWaldmann: yes, before msgpack is invoked, existing code always wrapped the existing dict with StableDict().
So, putting this into .as_dict() ...

ThomasWaldmann:
merge? this is isolated, so should be no problem.

enkore:
Just need to remember that we still have to do the actual refactoring :)

Issue #45, opened 2015-06-05
interesting hashes / macs / ciphers / checksums


ThomasWaldmann:
https://github.com/Cyan4973/xxHash - not a cryptographic hash fn, not for HMAC! So, maybe we could use it as a crc32 replacement (if we keep the crc32(header+all_data) approach). **borg uses xxh64 at some places**

siphash - cryptographic hash fn (internally used by python >= 3.4), but: only 64bits return value. a 128bit version is "experimental".

libsodium has some hashes / macs also. but not yet widespread on linux dists.

last but not least: sha512-256 is faster on 64bit CPUs than sha256.

ThomasWaldmann:
note: in experimental branch, there is code for aes-gcm AEAD cipher, ghash / gmac for hashing / maccing.

namelessjon:
If/when you do feel you can use libsodium, that also opens up two things which could be useful:

* scrypt as a PBKDF.  This should be (much) less vulnerable to brute force password guessing
* secretbox with XSalsa20 and Poly-1305.  This would solve worries over counter re-use, as you can randomly generate a 24-byte nonce for each new encryption with negligible chance of reuse.

ThomasWaldmann:
@namelessjon interesting, thanks. about the counter re-use issue: there is also this idea of creating per-backup(-thread) random session keys, start counter from 0, encrypt the keys with the master key and store them with the backup.

namelessjon:
@ThomasWaldmann That sounds less fragile than the current implementation at least (I think).

The secretbox option from libsodium, despite the larger nonce, adds the same 40 byte overhead to the files, because the Poly-1305 MAC is only 16 bytes.  I guess if you're now including a pointer to the encrypted session key with each encrypted blob, secretbox would have less overhead, but that shouldn't be that significant anyway?

ThomasWaldmann:
I had a look at libsodium yesterday, seems pretty nice and it also seem to be in some stable linux distributions now.

It would be useful to get some comparative performance values:
aes256-ctr + hmac-sha256 vs aes256-gcm (hw accel.) vs. xsalsa20 + poly1305
same for sha256 against some faster hash from libsodium.

For interfacing, we have 2 options: either via cython (like we use it for openssl right now) or using some python wrapper for libsodium.

namelessjon:
[...]

PR #985, opened 2016-04-26
Improved LoggedIO write performance

enkore:
Note: the original idea in this PR does not work (well it does, but not the way I wanted it ;). ~~Alternatives are being developed.~~ Alternative developed, needs testing.

Also see https://github.com/borgbackup/borg/issues/45#issuecomment-213715080

Summary

Test set:

*  images + XMP metadata (for some folders)
* average file size ~3.5 MB
* 5388 files
* 19.49 GB

```
                        Throughput     %CPU      total CPU time
1.0.2 / master ........ 24 MB/s ...... 33 % .... 280 s
fastlio none .......... 52 MB/s ...... 62 % .... 237 s
fastlio none64 ........ 54 MB/s ...... 55 % .... 205 s
```

All are I/O-bound, but fastlio is less so. free(1) shows that master branch doesn't
let the kernel use it's buffers (due to fsync after each segment) efficiently, while
fastlio does. Profiling shows that for fastlio fsync() makes no significant wall-clock
appereance, while it is a large factor for master.

This patch doesn't (or shouldn't) trash the page cache, as it immediately tells the kernel after writing
that the data is not needed anymore. The fsync() comes much later, at a point where it seems
the kernel already wrote everything to disk on it's own accord. Hence these fsync()s return
immediately, instead of blocking.

While in my tests none and none64 did not make a tangible difference, none64 would be useful
where fast IO is available but the CPU isn't very fast (=NAS), since the average CPU loads
show it uses much less CPU time.

For machines running encrypted repositories with hardware AES support (AES-NI for example)
the throughput should scale similar. For machines with no hardware AES support the difference
will be less pronounced.

### DIFF
--- a/borg/constants.py
+++ b/borg/constants.py
@@ -12,8 +12,14 @@ UMASK_DEFAULT = 0o077
 CACHE_TAG_NAME = 'CACHEDIR.TAG'
 CACHE_TAG_CONTENTS = b'Signature: 8a477f597d28d172789f06886806bc55'
 
-DEFAULT_MAX_SEGMENT_SIZE = 5 * 1024 * 1024
-DEFAULT_SEGMENTS_PER_DIR = 10000
+# A large, but not unreasonably large segment size. Always less than 2 GiB (for legacy file systems). We choose
+# 500 MiB which means that no indirection from the inode is needed for typical Linux file systems.
+# Note that this is a soft-limit and can be exceeded (worst case) by a full maximum chunk size and some metadata
+# bytes. That's why it's 500 MiB instead of 512 MiB.
+DEFAULT_MAX_SEGMENT_SIZE = 500 * 1024 * 1024
+
+# A few hundred files per directory to go easy on filesystems which don't like too many files per dir (NTFS)
+DEFAULT_SEGMENTS_PER_DIR = 500
 
 CHUNK_MIN_EXP = 19  # 2**19 == 512kiB
 CHUNK_MAX_EXP = 23  # 2**23 == 8MiB
--- a/borg/platform_base.py
+++ b/borg/platform_base.py
@@ -1,0 +1,78 @@
+import os
+
+API_VERSION = 3
+
+fdatasync = getattr(os, 'fdatasync', os.fsync)
+
+
+def acl_get(path, item, st, numeric_owner=False):
+    """
+    Saves ACL Entries
+
+    If `numeric_owner` is True the user/group field is not preserved only uid/gid
+    """
+
+
+def acl_set(path, item, numeric_owner=False):
+    """
+    Restore ACL Entries
+
+    If `numeric_owner` is True the stored uid/gid is used instead
+    of the user/group names
+    """
+
+
+def sync_dir(path):
+    fd = os.open(path, os.O_RDONLY)
+    try:
+        os.fsync(fd)
+    finally:
+        os.close(fd)
+
+
+class SyncFile:
+    """
+    A file class that is supposed to enable write ordering (one way or another) and data durability after close().
+
+    The degree to which either is possible varies with operating system, file system and hardware.
+
+    This fallback implements a naive and slow way of doing this. On some operating systems it can't actually
+    guarantee any of the above, since fsync() doesn't guarantee it. Furthermore it may not be possible at all
+    to satisfy the above guarantees on some hardware or operating systems. In these cases we hope that the thorough
+    checksumming implemented catches any corrupted data due to misordered, delayed or partial writes.
+
+    Note that POSIX doesn't specify *anything* about power failures (or similar failures). A system that
+    routinely loses files or corrupts file on power loss is POSIX compliant.
+
+    TODO: Use F_FULLSYNC on OSX.
+    TODO: A Windows implementation should use CreateFile with FILE_FLAG_WRITE_THROUGH.
+    """
+
+    def __init__(self, path):
+        self.fd = open(path, 'wb')
+        self.fileno = self.fd.fileno()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def write(self, data):
+        self.fd.write(data)
+
+    def sync(self):
+        """
+        Synchronize file contents. Everything written prior to sync() must become durable before anything written
+        after sync().
+        """
+        self.fd.flush()
+        fdatasync(self.fileno)
+        if hasattr(os, 'posix_fadvise'):
+            os.posix_fadvise(self.fileno, 0, 0, os.POSIX_FADV_DONTNEED)
+
+    def close(self):
+        """sync() and close."""
+        self.sync()
+        self.fd.close()
+        sync_dir(os.path.dirname(self.fd.name))
--- a/borg/platform_linux.pyx
+++ b/borg/platform_linux.pyx
@@ -1,13 +1,17 @@
 import os
 import re
+import resource
 from stat import S_ISLNK
 from .helpers import posix_acl_use_stored_uid_gid, user2uid, group2gid, safe_decode, safe_encode
+from .platform_base import SyncFile as BaseSyncFile
+from libc cimport errno
 
-API_VERSION = 2
+API_VERSION = 3
 
 cdef extern from "sys/types.h":
     int ACL_TYPE_ACCESS
     int ACL_TYPE_DEFAULT
+    ctypedef off64_t
 
 cdef extern from "sys/acl.h":
     ctypedef struct _acl_t:
@@ -22,6 +26,12 @@ cdef extern from "sys/acl.h":
 
 cdef extern from "acl/libacl.h":
     int acl_extended_file(const char *path)
+
+cdef extern from "fcntl.h":
+    int sync_file_range(int fd, off64_t offset, off64_t nbytes, unsigned int flags)
+    unsigned int SYNC_FILE_RANGE_WRITE
+    unsigned int SYNC_FILE_RANGE_WAIT_BEFORE
+    unsigned int SYNC_FILE_RANGE_WAIT_AFTER
 
 
 _comment_re = re.compile(' *#.*', re.M)
@@ -77,10 +87,6 @@ cdef acl_numeric_ids(acl):
 
 
 def acl_get(path, item, st, numeric_owner=False):
-    """Saves ACL Entries
-
-    If `numeric_owner` is True the user/group field is not preserved only uid/gid
-    """
     cdef acl_t default_acl = NULL
     cdef acl_t access_acl = NULL
     cdef char *default_text = NULL
@@ -112,11 +118,6 @@ def acl_get(path, item, st, numeric_owner=False):
 
 
 def acl_set(path, item, numeric_owner=False):
-    """Restore ACL Entries
-
-    If `numeric_owner` is True the stored uid/gid is used instead
-    of the user/group names
-    """
     cdef acl_t access_acl = NULL
     cdef acl_t default_acl = NULL
 
@@ -141,3 +142,45 @@ def acl_set(path, item, numeric_owner=False):
                 acl_set_file(p, ACL_TYPE_DEFAULT, default_acl)
         finally:
             acl_free(default_acl)
+
+cdef _sync_file_range(fd, offset, length, flags):
+    assert offset & PAGE_MASK == 0, "offset %d not page-aligned" % offset
+    assert length & PAGE_MASK == 0, "length %d not page-aligned" % length
+    if sync_file_range(fd, offset, length, flags) != 0:
+        raise OSError(errno, os.strerror(errno))
+    os.posix_fadvise(fd, offset, length, os.POSIX_FADV_DONTNEED)
+
+cdef unsigned PAGE_MASK = resource.getpagesize() - 1
+
+
+class SyncFile(BaseSyncFile):
+    """
+    Implemented using sync_file_range for asynchronous write-out and fdatasync for actual durability.
+
+    "write-out" means that dirty pages (= data that was written) are submitted to an I/O queue and will be send to
+    disk in the immediate future.
+    """
+
+    def __init__(self, path):
+        super().__init__(path)
+        self.offset = 0
+        self.write_window = (16 * 1024 ** 2) & ~PAGE_MASK
+        self.last_sync = 0
+        self.pending_sync = None
+
+    def write(self, data):
+        self.offset += self.fd.write(data)
+        offset = self.offset & ~PAGE_MASK
+        if offset >= self.last_sync + self.write_window:
+            self.fd.flush()
+            _sync_file_range(self.fileno, self.last_sync, offset - self.last_sync, SYNC_FILE_RANGE_WRITE)
+            if self.pending_sync is not None:
+                _sync_file_range(self.fileno, self.pending_sync, self.last_sync - self.pending_sync,
+                                 SYNC_FILE_RANGE_WRITE | SYNC_FILE_RANGE_WAIT_BEFORE | SYNC_FILE_RANGE_WAIT_AFTER)
+            self.pending_sync = self.last_sync
+            self.last_sync = offset
+
+    def sync(self):
+        self.fd.flush()
+        os.fdatasync(self.fileno)
+        os.posix_fadvise(self.fileno, 0, 0, os.POSIX_FADV_DONTNEED)
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -17,6 +17,7 @@ from .helpers import Error, ErrorWithTraceback, IntegrityError, Location, Progre
 from .hashindex import NSIndex
 from .locking import UpgradableLock, LockError, LockErrorT
 from .lrucache import LRUCache
+from .platform import SyncFile, sync_dir
 
 MAX_OBJECT_SIZE = 20 * 1024 * 1024
 MAGIC = b'BORG_SEG'
@@ -32,7 +33,7 @@ class Repository:
     On disk layout:
     dir/README
     dir/config
-    dir/data/<X / SEGMENTS_PER_DIR>/<X>
+    dir/data/<X // SEGMENTS_PER_DIR>/<X>
     dir/index.X
     dir/hints.X
     """
@@ -507,7 +508,7 @@ class LoggedIO:
     def __init__(self, path, limit, segments_per_dir, capacity=90):
         self.path = path
         self.fds = LRUCache(capacity,
-                            dispose=lambda fd: fd.close())
+                            dispose=self.close_fd)
         self.segment = 0
         self.limit = limit
         self.segments_per_dir = segments_per_dir
@@ -518,6 +519,11 @@ class LoggedIO:
         self.close_segment()
         self.fds.clear()
         self.fds = None  # Just to make sure we're disabled
+
+    def close_fd(self, fd):
+        if hasattr(os, 'posix_fadvise'):  # only on UNIX
+            os.posix_fadvise(fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
+        fd.close()
 
     def segment_iterator(self, reverse=False):
         data_path = os.path.join(self.path, 'data')
@@ -535,7 +541,7 @@ class LoggedIO:
         return None
 
     def get_segments_transaction_id(self):
-        """Verify that the transaction id is consistent with the index transaction id
+        """Return the last committed segment.
         """
         for segment, filename in self.segment_iterator(reverse=True):
             if self.is_committed_segment(filename):
@@ -578,7 +584,8 @@ class LoggedIO:
                 dirname = os.path.join(self.path, 'data', str(self.segment // self.segments_per_dir))
                 if not os.path.exists(dirname):
                     os.mkdir(dirname)
-            self._write_fd = open(self.segment_filename(self.segment), 'ab')
+                    sync_dir(os.path.join(self.path, 'data'))
+            self._write_fd = SyncFile(self.segment_filename(self.segment))
             self._write_fd.write(MAGIC)
             self.offset = MAGIC_LEN
         return self._write_fd
@@ -590,6 +597,13 @@ class LoggedIO:
             fd = open(self.segment_filename(segment), 'rb')
             self.fds[segment] = fd
             return fd
+
+    def close_segment(self):
+        if self._write_fd:
+            self.segment += 1
+            self.offset = 0
+            self._write_fd.close()
+            self._write_fd = None
 
     def delete_segment(self, segment):
         if segment in self.fds:
@@ -641,7 +655,7 @@ class LoggedIO:
 
     def read(self, segment, offset, id):
         if segment == self.segment and self._write_fd:
-            self._write_fd.flush()
+            self._write_fd.sync()
         fd = self.get_fd(segment)
         fd.seek(offset)
         header = fd.read(self.put_header_fmt.size)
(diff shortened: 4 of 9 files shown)


enkore:
Detailed test log

```
$ borg init -e none64 liotest
$ sudo sync && echo 3 | sudo tee /proc/sys/vm/drop_caches
$ /bin/time borg create liotest::fastlio-none64 ~/largehome/Sync/Bilder -svp --no-files-cache
Archive name: fastlio-none64
Archive fingerprint: 24acf7bd996a25d8041356e1afd2297881e40a3b34026521d1cbaf245c6845ca
Time (start): Tue, 2016-04-26 14:45:48
Time (end):   Tue, 2016-04-26 14:51:53
Duration: 6 minutes 4.99 seconds
Number of files: 5388
This archive:               19.49 GB             19.49 GB             18.08 GB
All archives:               19.49 GB             19.49 GB             18.08 GB

Chunk index:                   11162                12133
186.10user 18.45system 6:05.38elapsed 55%CPU (0avgtext+0avgdata 55384maxresident)k
38102378inputs+35359864outputs (0major+105642minor)pagefaults 0swaps

$ rm -rf liotest
$ borg init -e none liotest
$ sudo sync && echo 3 | sudo tee /proc/sys/vm/drop_caches
$ /bin/time borg create liotest::fastlio-none ~/largehome/Sync/Bilder -svp --no-files-cache
Archive name: fastlio-none
Archive fingerprint: 47ef43ac29025331dec62e515c75de92a55280bb6d94488d486c9eebedd1b8f7
Time (start): Tue, 2016-04-26 14:38:50
Time (end):   Tue, 2016-04-26 14:45:07
Duration: 6 minutes 17.43 seconds
Number of files: 5388
This archive:               19.49 GB             19.49 GB             18.08 GB
All archives:               19.49 GB             19.49 GB             18.08 GB

Chunk index:                   11162                12133
219.13user 17.68system 6:18.18elapsed 62%CPU (0avgtext+0avgdata 56144maxresident)k
38144873inputs+35373736outputs (68major+84764minor)pagefaults 0swaps

$ ./borg-linux64 -V
borg-linux64 1.0.2
$ ./borg-linux64 init -e none liotest-master
$ sudo sync && echo 3 | sudo tee /proc/sys/vm/drop_caches
$ /bin/time ./borg-linux64 create liotest-master::master-none ~/largehome/Sync/Bilder -svp --no-files-cache
Archive name: master-none
Archive fingerprint: b6db9aaee8ff539120fff67bd678d5ca0f55574e2af131757879c4c568a4d08e
Time (start): Tue, 2016-04-26 12:49:03
Time (end):   Tue, 2016-04-26 13:02:55
Duration: 13 minutes 51.93 seconds
Number of files: 5388
This archive:               19.49 GB             19.49 GB             18.08 GB
All archives:               19.49 GB             19.49 GB             18.08 GB

Chunk index:                   11165                12136
253.81user 26.69system 13:52.44elapsed 33%CPU (0avgtext+0avgdata 54912maxresident)k
38136530inputs+35385096outputs (3major+490703minor)pagefaults 0swaps
```

borg/repository.py
```diff
@@ -504,21 +506,30 @@ class SegmentFull(Exception):
...
     def close(self):
-        self.close_segment()
+        self.close_segment(sync=True)
         self.fds.clear()
         self.fds = None  # Just to make sure we're disabled
 
+    def close_fd(self, fd):
+        if hasattr(os, 'posix_fadvise'):  # only on UNIX
+            os.posix_fadvise(fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
+        fd.close()
```
enkore: This is used for cleaning up read FDs. FADV_DONTNEED seems to be a command on Linux, not an advice. I.e. reading a segment would not put it into the page cache on Linux when the segment reaches the end of the LRU cache. This could affect read performance for large repositories with scattered deduplication.

borg/repository.py
```diff
@@ -591,6 +602,28 @@ def get_fd(self, segment):
...
+            self._write_fds.append(self._write_fd)
+            if hasattr(os, 'posix_fadvise'):  # only on UNIX
+                # tell the OS that it does not need to cache what we just wrote,
+                # avoids spoiling the cache for the OS and other processes.
+                os.posix_fadvise(self._write_fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
+            self._write_fd = None
+        if sync:
+            for fd in self._write_fds:
+                self._sync_and_close(fd)
+            self._write_fds.clear()
```
enkore: Note that this logic has the same fsync()iness as before; all segments of a commit are fsync()ed before the committing segment, which is fsync()ed last.

borg/repository.py
ThomasWaldmann: well, it first is in the page cache, but then removed from there. so if we have a LRU cache of size N, it is N * segment size cache space usage (which is well acceptable for relatively small N).

borg/repository.py
ThomasWaldmann: Isn't the required property like this?:
first write and fsync all segments of a commit (but do not write the commit), then write (and fsync) the commit.

ThomasWaldmann:
hmm, fails on OS X.

ThomasWaldmann:
Can you keep the none64 changes in a different PR? Other than they are also tuning speed, they seem unrelated. See also the experimental repo with a lot of similar stuff.

borg/repository.py
enkore: Ah yes. If the commit was written() before the fsync()s then (at least hypothetically) the OS could write the commit to disk first.

borg/repository.py
ThomasWaldmann: hmm, as this is new in your changeset, do i see it right that our previous code spoiled the cache when reading a lot of data from a repo?

borg/repository.py
```diff
@@ -591,6 +602,28 @@ def get_fd(self, segment):
...
+        if len(self._write_fds) == self.MAX_PENDING_WRITE_FDS:
+            self._sync_and_close(self._write_fds.popleft())
+        if self._write_fd:
+            self.segment += 1
+            self.offset = 0
+            self._write_fds.append(self._write_fd)
+            if hasattr(os, 'posix_fadvise'):  # only on UNIX
+                # tell the OS that it does not need to cache what we just wrote,
+                # avoids spoiling the cache for the OS and other processes.
+                os.posix_fadvise(self._write_fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
```
ThomasWaldmann: i have some doubts it works like this.
i either have read or experienced that fadvise-dontneed is a NOP if the page cache is dirty, so one needs to write and sync first.

enkore: So, make max_pending_write_fds much smaller (say, 30 or so)?

Anyway, it's interesting that fadvise doesn't work / is said to not work. From the other stuff that I read lately and what verygreen said I can only conclude that *adivse are utterly broken and misleading APIs.

ThomasWaldmann: yes, 30 seems to be a good number. we shouldn't have too much in write cache if the I/O system can't catch up.

borg/repository.py
```diff
@@ -591,6 +602,28 @@ def get_fd(self, segment):
             self.fds[segment] = fd
             return fd
 
+    def _sync_and_close(self, fd):
+        fd.flush()
+        os.fdatasync(fd.fileno())
```
ThomasWaldmann: uhoh, while searching this i found that OSX doesn't make the guarantee that fsync really waits until all data arrived on disk.

borg/repository.py
```diff
@@ -591,6 +602,28 @@ def get_fd(self, segment):
...
+        fd.close()
+
+    def close_segment(self, sync=False):
+        if len(self._write_fds) == self.MAX_PENDING_WRITE_FDS:
```
ThomasWaldmann: ... == self._write_fds.maxlen:

just to make sure we have the right value. :)

borg/repository.py
```diff
@@ -591,6 +602,28 @@ def get_fd(self, segment):
...
+            self._sync_and_close(self._write_fds.popleft())
```
ThomasWaldmann: also, shouldn't this whole if-block be moved to the line directly above self._write_fds.append()?

borg/repository.py
ThomasWaldmann: this used to be .fsync previously. i read some docs about the difference and .fdatasync seems to be fine for us.

but: i also read enough to think that the sync stuff is a similar mess like the fadvise stuff, so a conservative approach would be to not change this call in this PR, but to open a ticket "fix sync for everybody" - likely needing different changes for linux, osx, windows.

enkore:
I'll leave this here https://lwn.net/Articles/591723/ and return later.

ThomasWaldmann:
non-commit writes: write immediately, queue fsync/fadvise (if queue is full, process first entry to make room).

commit writes: process queue entries (will be quick as kernel has already written most stuff anyway), write commit, fsync commit.

enkore:
None of this works. A short queue (5-30 segments) doesn't make any difference, because the kernel just buffers it and doesn't start to write anything until the fsync()s come in, at which point it's back to stop-and-wait.

I then added (asynchronous) sync_file_range calls when enqueueing segments in the sync queue, which go through, but also doesn't motivate anything; all fdatasync calls still block for eternity, pretty much independent of queue length. Only when the queue length exceeds ~1 GB the performance gets up to where it was above -- with the exact same cache usage as before.

This will need to go in an entirely different direction to improve performance.

ThomasWaldmann:
If we do not want to write and fsync-wait all the time, we just write and use the kernel's write-caching (and related policies / settings of the kernel).

If these settings allow a relatively big age of dirty write buffers (like >>1s), THIS is causing that a lot of dirty write buffers are pending. The more pending buffers, the longer we have to wait if we need to force them all on disk by fsyncing.

So I think the solution is not to allow such a big age of dirty write buffers on a system that writes big amounts of data. Modify the max age of dirty write buffers to a low value.

enkore:
Yes, that should be the purpose of sync_file_range (it doesn't write any metadata, not even the meta necessary to find the data), to limit the amount of dirty pages, so that a later fdatasync completes much quicker. This does not have any effect here though. I am wondering whether this might be due to the small segment size (and the kernel thinking "what? syncing just 5 MB? nah, i don't think so"), I think I'll look into "jumbo-segments" as the next route, maybe that helps.

ThomasWaldmann:
@enkore i wasn't talking about doing exotic calls in borg, but just setting that kernel setting to an appropriate value. That plus https://github.com/borgbackup/borg/pull/985#issuecomment-214938804 should solve it, right?

enkore:
Hm, I didn't really consider those, since they are system-wide and affecting everything. On the other hand, they are more effective than any combination of syncing-queues, jumbo segments and/or sync_file_range I have tried so far. With that in mind, I am going to close this.

By the way, the defaults on my system equalled to "mere" 5 _gigabytes_ of dirty pages. Whoops.

enkore:
Some points from @verygreen 

* Depending on file system fsync()s are not constrained to a single file, but more or less global = expensive
* Ideally just write write write as long as possible
* As few fsync()s as possible
* If the file system flushes it's transaction, then all data and metadata of the txn is consistent
* Segment files should not be too big, so that no extra indirection is needed to access contents. On ext4 this means 512 MiB.

enkore:
Summary:

* Larger segments
* Smarter compaction (hints file is backwards incompatible, version=2)
* Commit mechanism as practised/recommended (double-sync-commit, first write data, then sync, then write commit tag, then sync again)
* Linux: sync_file_range for immediate write-out avoids fdatasync-stall (s_f_r *only* for performance, *not* for durability or anything like that)

borg/platform_linux.pyx
```diff
@@ -141,3 +140,47 @@ def acl_set(path, item, numeric_owner=False):
...
+        self.last_sync = 0
+        self.pending_sync = None
+
+    def write(self, data):
+        self.offset += self.fd.write(data)
+        if self.offset >= self.last_sync + self.write_window:
+            self.fd.flush()
+            request_sync(self.fd, self.last_sync, self.offset - self.last_sync)
+            if self.pending_sync is not None:
+                force_sync(self.fd, self.pending_sync, self.last_sync - self.pending_sync)
```
verygreen: note, that if your data is not page-aligned, the fadv_dontneed in force_sync would flush the whole file content out and the next write would force the OST to read the partial page at the end of file read back in. Is this what you really want here?

enkore: Iirc fadvise should round partial pages at the end of the range down?

enkore: .. is this the same bug as in #1024?

verygreen: Except at the end of file. And you request start=len=0 meaning EVERYTHING in the file should go away, not just the segment you just wrote (where you would need to pass in the actual start and len similar to sync_file_range).
But even then e.g. in Linux there is a bug where it would still discard the page at the end of the range (but not at the start) - see my other patch in 1.0-maint.

enkore: ... yes it is.

I'll add the same page-rounding here as well.

verygreen: I think rounding is best done in force_sync() and possible other sync functions where it makes sense.

borg/platform_base.py
```diff
@@ -0,0 +1,68 @@
...
+    def close(self):
+        """sync() and close."""
+        self.sync()
+        self.fd.close()
+        try:
+            fd = os.open(os.path.dirname(self.fd.name), os.O_RDONLY)
+            os.fsync(fd)
+            os.close(fd)
+        except OSError as os_error:
+            print(os_error)
```
enkore: whoops.

enkore:
Hole punching: Would now be possible with the changes from d6944df. Old versions would choke on it, though.

ThomasWaldmann:
I would not update repo segment size setting automatically, rather document it.

borg/constants.py
```diff
@@ -12,8 +12,14 @@
 CACHE_TAG_NAME = 'CACHEDIR.TAG'
 CACHE_TAG_CONTENTS = b'Signature: 8a477f597d28d172789f06886806bc55'
 
-DEFAULT_MAX_SEGMENT_SIZE = 5 * 1024 * 1024
-DEFAULT_SEGMENTS_PER_DIR = 10000
+# A large, but not unreasonably large segment size. Always less than 2 GiB (for legacy file systems). We choose
+# 500 MiB which means that no indirection from the inode is needed for typical Linux file systems.
+# Note that this is a soft-limit and can be exceeded (worst case) by a full maximum chunk size and some metadata
+# bytes. That's why it's 500 MiB instead of 512 MiB.
+DEFAULT_MAX_SEGMENT_SIZE = 500 * 1024 * 1024
```
ThomasWaldmann: hmm, is it worthwhile to go that big, by default?

for remote operations on segment files over a slow connection, this can mean a huge transfer time per segment. how much would be performance worse if it was just 50MB?

borg/constants.py
```diff
@@ -12,8 +12,14 @@
...
+
+# A few hundred files per directory to go easy on filesystems which don't like too many files per dir (NTFS)
+DEFAULT_SEGMENTS_PER_DIR = 500
```
ThomasWaldmann: otoh, this seems very low.

borg/platform_base.py
```diff
@@ -0,0 +1,68 @@
...
+class SyncFile:
+    """
+    A file class that is supposed to enable write ordering (one way or another) and data durability after close().
+
+    The degree to which either is possible varies with operating system, file system and hardware.
+
+    This fallback implements a naive and slow way of doing this. On some operating systems it can't actually
+    guarantee any of the above, since fsync() doesn't guarantee it. Furthermore it may not be possible at all
+    to satisfy the above guarantees on some hardware or operating systems. In these cases we hope that the thorough
+    checksumming implemented catches any corrupted data due to misordered, delayed and partial writes.
```
ThomasWaldmann: "... or partial writes."?

borg/platform_base.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Note that POSIX doesn't specify *anything* about power failures (or similar failures). It is 100 % A-OK if files
```
ThomasWaldmann: "100%"
and not sure if everybody understands "A-OK".

borg/platform_base.py
ThomasWaldmann: just remove the try/except?

do a open try fsync finally close?

borg/platform_linux.pyx
```diff
@@ -141,3 +140,47 @@ def acl_set(path, item, numeric_owner=False):
...
+    _sync_file_range(fd, offset, length, SYNC_FILE_RANGE_WRITE | SYNC_FILE_RANGE_WAIT_BEFORE)
+    os.posix_fadvise(fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
+
+
+class SyncFile(BaseSyncFile):
+    """
+    Implemented using sync_file_range for asynchronous write-out and fdatasync for actual durability.
+
+    "write-out" means that dirty pages (= data that was written) are submitted to an I/O queue and will be send to
+    disk in the immediate feature.
```
ThomasWaldmann: future

borg/platform_linux.pyx
```diff
@@ -141,3 +140,47 @@ def acl_set(path, item, numeric_owner=False):
...
+    """
+
+    def __init__(self, path):
+        self.fd = open(path, 'wb')
```
ThomasWaldmann: where is it closed? implement a context manager?

borg/platform_base.py
enkore: Yes

borg/platform_linux.pyx
enkore: Open/close is handled by LoggedIO, a context manager doesn't fit the code there. Note that the close() is inherited from platform_base.

borg/constants.py
enkore: Continuous writes to a segment are fast, while the final f[data]sync + directory sync have approximately fixed costs. At a couple hundred MB it's as fast as it gets for me, while it gets progressively slower the lower the segment size.

I see little reason to go for something much smaller; normal remote access doesn't transfer segments. rsync'ing repositories will have to rsync the same amount of data, independent of 1x 500 MB or 100x 5 MB...

borg/constants.py
enkore: With the raised segment size it's actually a bit more data/dir than before (10000 * 5 MB = 50 GB, 500 * 500 MB = 250 GB).

borg/repository.py
```diff
@@ -578,7 +633,7 @@ def get_write_fd(self, no_new=False, raise_full=False):
                 dirname = os.path.join(self.path, 'data', str(self.segment // self.segments_per_dir))
                 if not os.path.exists(dirname):
                     os.mkdir(dirname)
```
enkore: fsync the parent directory here?

ThomasWaldmann: sounds reasonable.

borg/platform_linux.pyx
enkore: It's not a bug it's a future

(said the async developer)

borg/constants.py
ThomasWaldmann: In case we'll ever have to do a segment-converter for upgrading, we'ld have to deal with these huge segment files (and could run out of memory when trying to read them into memory for processing) - that would not be a problem if we just increase to 50 or 100MB.

So the question is whether the speedup of the  50/100 -> 500 MB segment file increase rectifies that potential trouble. I'ld guess the speedup from 5 -> 50 is significant, while the speedup from 50 -> 500 might be negligible.

verygreen: not that I have a strong opinion, but since segments are still made out of chunks, you do not need to keep the whole segment file in RAM, what's to stop you from reading it chunk by chunk, do the conversion and write it out into the other file?

500MB file is not all that big and looks reasonably sized for me, works on all filesystem types too.

On the other hand 50MB files is probably ok too, you'll just have more of them => a bit more of directory overhead, 10x more inodes used and such, but probably nothing fatal.

enkore: 52 MB/s at 500 MB
35 MB/s at 50 MB
25 MB/s at 5 MB

While the attic upgrader does indeed read a segment completly into memory it doesn't need to (`new_segment.write(segment.read())` -> just use `shutil.copyfile*`)

ThomasWaldmann: hmm, unexpected results. can you do one @ 100MB, 200MB?

verygreen: wow, even at 52 MB/sec this is really-really slow for just a write. Or is it doing more than just writes?
strace of the process might be interesting to see what's going on.

enkore: I wouldn't call it unexpected ;)

> Continuous writes to a segment are fast, while the final f[data]sync + directory sync have approximately fixed costs.

ThomasWaldmann: 5MB means 100 fsyncs, 50MB is 10 fsyncs, 500MB is 1 fsync.
So expectation is that 5 -> 50 makes a bigger difference than 50 -> 500.

enkore: 40 MB/s 200 MB

borg/platform_linux.pyx
```diff
@@ -141,3 +140,47 @@ def acl_set(path, item, numeric_owner=False):
                 acl_set_file(p, ACL_TYPE_DEFAULT, default_acl)
         finally:
             acl_free(default_acl)
+
+cdef _sync_file_range(fd, offset, length, flags):
+    if sync_file_range(fd.fileno(), offset, length, flags) != 0:
+        raise OSError(errno, os.strerror(errno))
```
ThomasWaldmann: maybe remove this wrapper? it just adds the raise and the .fileno - and if you did not have the wrapper, you could reuse the fileno for fadvise.

enkore:
> File systems

> We strongly recommend to not use Borg (or any other database-like software) on
> non-journaling file systems like FAT, since it is not possible to assume any
> consistency in case of power failures (or a sudden disconnect of an external drive).

> While Borg uses a data store that is resilient against these failures when used
> on journaling file systems, it is not possible to guarantee this with some hardware.

borg/platform_base.py
```diff
@@ -22,6 +22,14 @@ def acl_set(path, item, numeric_owner=False):
     """
 
 
+def sync_dir(path):
+    fd = os.open(os.path.dirname(path), os.O_RDONLY)
```
ThomasWaldmann: remove the dirname here, you already have it at the calling place.

enkore: yes, path should be dir here.

verygreen:
Note that journaling is not a mandatory requirement, zfs does not use journaling for example.
There must be a better more generic term, but I cannot think of one ATM.

enkore:
"transactional", but that usually means file systems that expose transactional APIs (like TxF) - at least for me. We could say "internally transactional", but that's meaningless to almost everyone.

enkore:
Hmm, if I extend SyncFile to write to a temporary (in the same dir) first and rename(2) on close and sync_dir after that then it would be good as a general-purpose "save file writer" (like QSaveFile). We should then use that for writing e.g. the repository config (which can contain the key which we don't want to lose!)

ThomasWaldmann:
@enkore good idea, but please do that in a future PR and get the existing ones finished first.
that might need more changes, like ignoring such temp files when scanning for valid content.

enkore:
Yes, this one is big and complex enough as it is ;)

docs/usage.rst
```diff
@@ -168,6 +168,22 @@ Network:
...
+sudden disconnect of an external drive or similar failures).
+
+While Borg uses a data store that is resilient against these failures
+when used on journaling file systems, it is not possible to guarantee
+this with some hardware -- independent of the software used. We don't
+know a list of affected hardware.
+
+If you are suspicious whether your Borg repository is still consistent
+and readable after one of the failures mentioned above occured, run ``borg check
+--verify-data`` for certainty.
```
ThomasWaldmann: "... to make sure."?

enkore: "`` to make sure it is [consistent]."?

enkore:
Unstaged all changes unrelated to the IO. Note that this will routinely have worse performance now for smaller runs.

borg/platform_base.py
```diff
@@ -0,0 +1,72 @@
...
+        self.fd.flush()
+        fdatasync(self.fileno)
+        if hasattr(os, 'posix_fadvise'):
+            os.posix_fadvise(self.fileno, 0, 0, os.POSIX_FADV_DONTNEED)
+
+    def close(self):
+        """sync() and close."""
+        self.sync()
+        self.fd.close()
+        sync_dir(os.path.dirname(self.fd.name))
```
ThomasWaldmann: add a context manager?

enkore: I can do that, but it wouldn't be used by LoggedIO, since it doesn't fit the code structure.

ThomasWaldmann: Hmm, iirc you were talking using this for other stuff, too?

enkore: Yes, I'll add a wrapper around SyncFile though for that to avoid cluttering the cleanup/rollback code with additional cleanup code.

PR #990, opened 2016-04-27
borg with-lock REPO CMD ARGS

ThomasWaldmann:
Guess this is ready now for merging. After review, I'll collapse the commits.

manual usage:
```
(borg-env)tw@tux:~/w/borg$ borg with-lock repo ls -d repo/lock.exclusive
repo/lock.exclusive
```

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -12,6 +12,7 @@ import os
 import shlex
 import signal
 import stat
+import subprocess
 import sys
 import textwrap
 import traceback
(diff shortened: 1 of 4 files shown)


edgewood:
Consider using nargs=argparse.REMAINDER to avoid the requirement for quotes and use of shlex.split.
-- 
Ed Blackman

ThomasWaldmann:
@edgewood done, PR updated.

edgewood:
Help text still says "The command needs to be given as a single argument..."

ThomasWaldmann:
Oops, forgot "git add", likely.

ThomasWaldmann:
It seems that the remote repo only creates an exclusive lock when a transaction is started and releases it when the repo is closed. We need to issue a put or delete to start a transaction.

So, shall we add another "well known" and reserved id (like 1, manifest_id is 0) that we can use for a small put?

We could also use get() to fetch some arbitrary id's contents (if any), then put() it back (and delete() it in case it did not exist before).

ThomasWaldmann:
Done. :)

PR #1001, opened 2016-05-01
add BORG_KEY_FILE

ThomasWaldmann:
collapsed all commits into 1, rebased on current master, added tests.

thanks to @legrostdg for implementing this.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -77,8 +77,6 @@ General:
         When set, use the value to answer the passphrase question for encrypted repositories.
     BORG_DISPLAY_PASSPHRASE
         When set, use the value to answer the "display the passphrase for verification" question when defining a new passphrase for encrypted repositories.
-    BORG_KEY_FILE
-        When set, use the given filename as repository key file.
     BORG_LOGGING_CONF
         When set, use the given filename as INI_-style logging configuration.
     BORG_RSH
@@ -103,9 +101,11 @@ Some automatic "answerers" (if set, they automatically answer confirmation quest
     answer or ask you interactively, depending on whether retries are allowed (they by default are
     allowed). So please test your scripts interactively before making them a non-interactive script.
 
-Directories:
+Directories and files:
     BORG_KEYS_DIR
         Default to '~/.config/borg/keys'. This directory contains keys for encrypted repositories.
+    BORG_KEY_FILE
+        When set, use the given filename as repository key file.
     BORG_CACHE_DIR
         Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
         of space for dealing with big repositories).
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -77,6 +77,8 @@ General:
         When set, use the value to answer the passphrase question for encrypted repositories.
     BORG_DISPLAY_PASSPHRASE
         When set, use the value to answer the "display the passphrase for verification" question when defining a new passphrase for encrypted repositories.
+    BORG_KEY_FILE
+        When set, use the given filename as repository key file.
```
enkore: There is a later section "Directories" where BORG_KEYS_DIR is. Rename that section to "Directories and files" and move this there?

ThomasWaldmann: good idea, I'll do that.

legrostdg:
great!

chrisrapson:
Has this been reverted? I can't find any information in the current version of the docs about using an existing keyfile to init a new repository. Does the feature still exist?

When I set BORG_KEY_FILE and BORG_KEYFILE environment variables and ran `borg init --encryption=keyfile /path/to/repo` the first question it asked me was `Enter new passphrase:`. So if the feature is still available, I guess that's not how to use it.

My use case is that I would like to back up my system to multiple external hard drives. I don't need each one to have a unique key, that would just make more work for me to keep track of the keys/passphrases.

ThomasWaldmann:
you must not use 1 key for multiple repos (or duplicate existing repo and use the copies independently), this would be security issue due to counter reuse of AES-CTR mode.

ao, if you have N backup hdds, run borg init N times.

you may use the same passphrase for all of them if you like, each repo would still have a different random AES key as the key is **not** derived from the passphrase.

ThomasWaldmann:
If you think you might lose 1 key, where's the big difference of having a backup of 1 key or of e.g. 3 keys? the keys are encrypted, so if you have a good passphrase, their backups are unproblematic to store.

chrisrapson:
I can make N copies of 1 key, and keep them in safe places separate to the hard drives. I guess I can make N copies of M keys, but N*M is quite a bit more than N. I don't trust my memory of good passphrases either, so I'd have to backup N*M or those too, for a total of N*M*2 vs N*2. You're right, it's not a big difference, but I was hoping to simplify things for myself.

In the meantime I had another read of the docs, and I think I can achieve what I want with "repokey" encryption. Then the authentication is only via passphrase, and the keyfile is stored with the repo. I can duplicate the passphrase as much as I want. I think that's secure enough for my case, and keeps things locked down for users who want "keyfile" encryption.

By the way, I didn't find the explanation of the different encryption modes particularly easy to understand. The information is there, but I didn't pick it up the first couple of times I read it. Would it be helpful if I had a go at a clearer explanation, and made a pull request?

ThomasWaldmann:
You could always use the same passphrase (as you would do if you only had 1 key).

repokey stores the encrypted key in the repo, but you might still want to make a backup of it as suggested in the docs.

A PR with a clearer (correct) explanation is welcome.

PR #1007, opened 2016-05-02
compression heuristics based on lz4, fixes #1006

### LINKED ISSUE #1006: content-based heuristic compression selection
we can choose the compressor based on a heuristic that just looks at the content (not: at the filename or other metadata).

e.g. try using lz4 as a predictor for compressibility - if lz4 can not compress, do not even try something expensive like zlib or lzma. if it can, try expensive compression.

ThomasWaldmann:
also: add some tests that invoke all supported compression algorithms

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1359,6 +1359,8 @@ class Archiver:
                                    type=CompressionSpec, default=dict(name='none'), metavar='COMPRESSION',
                                    help='select compression algorithm (and level):\n'
                                         'none == no compression (default),\n'
+                                        'auto,C[,L] == built-in heuristic decides between none or C[,L] - with C[,L]\n'
+                                        '              being any valid compression algorithm (and optional level),\n'
                                         'lz4 == lz4,\n'
                                         'zlib == zlib (default level 6),\n'
                                         'zlib,0 .. zlib,9 == zlib (with level 0..9),\n'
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
hmm, likely none vs. lz4 is the wrong choice for auto.
using it, -C lz4 and -C auto both give about the same results (for size and execution time).

ThomasWaldmann:
wtf is wrong with codecov? i added a little code, quite some tests and it says coverage is significantly down?

ThomasWaldmann:
Currently, I used "zlib" as auto-compression default, which means "zlib,6" as level 6 is the zlib default.

Maybe we should use a lower level (6 is already relatively slow)?
OTOH, if you search for speed, you'ld use lz4 (and not auto).

enkore:
Since --compresion has no default let's don't have a default here as well?

ThomasWaldmann:
--compression has a default, it is "none". but that doesn't make sense for auto.

still, we could make auto not have a default. for similar reasons 'none' is the default for compression.

enkore:
Yep, that's what I meant.

Travis failure: Maybe something like 763a88ac is also needed for Travis?

ThomasWaldmann:
Done. Anything else other than squash before I merge?

borg/archiver.py
```diff
@@ -1812,6 +1813,7 @@ def build_parser(self, args=None, prog=None):
                                    type=CompressionSpec, default=None, metavar='COMPRESSION',
                                    help='select compression algorithm (and level):\n'
                                         'none == no compression (default),\n'
+                                        'auto,C[,L] == built-in heuristic decides between none or C[,L],\n'
```
enkore: I have only this tiny nit pick: `auto,C[,L]` doesn't explain what C and L is. Maybe just say that C[,L] any other valid compression algo + optional level?

ThomasWaldmann: yes, it is a forward reference somehow.

enkore:
Looks good to me

PR #1008, opened 2016-05-03
prune: kill all checkpoints except the latest

ThomasWaldmann:
prune: only keep latest checkpoint if not superceded by a successful backup

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -783,13 +783,21 @@ class Archiver:
                              '"keep-secondly", "keep-minutely", "keep-hourly", "keep-daily", '
                              '"keep-weekly", "keep-monthly" or "keep-yearly" settings must be specified.')
             return self.exit_code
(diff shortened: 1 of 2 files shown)


jdchristensen:
My only reason for suggesting a command line switch is that some people might backup multiple clients into one repo, and might run a prune command on the server once a week, for example, without a prefix.  But I think most people will arrange that the clients do their own pruning, so I don't feel very strongly about this.

ThomasWaldmann:
@jdchristensen how can that work without prefix? e.g. if you tell it --keep-monthly 6 it will keep the 1 latest monthly archive for each month for 6 months. and if you backup client A and client B into same repo, you will only keep the monthly backup for one of them (and if you always backup first A, then B, it will always keep B and never A).

jdchristensen:
I agree that it's not a likely use case.  The only thing I can imagine is someone using '--keep-within 2y' or something like that.

ThomasWaldmann:
added docs and collapsed commits.

enkore:
Looks good. Docs are nice and clear :)

PR #1013, opened 2016-05-05
faq: fix issue links

ThomasWaldmann:


docs/faq.rst
```diff
@@ -140,8 +140,7 @@ When doing a backup to a remote server (using a ssh: repo URL), it sometimes
 stops after a while (some minutes, hours, ... - not immediately) with
 "connection closed by remote" error message. Why?
 
-That's a good question and we are trying to find a good answer in
-`ticket 636 <https://github.com/borgbackup/borg/issues/636>`_.
+That's a good question and we are trying to find a good answer in :issue:`636`.
```
enkore: RTFD feature? Neat!

ThomasWaldmann: hmm, i thought it was in our global.rst.inc, but it isn't. so yes, seems to be a rtfd feature.

PR #1034, opened 2016-05-09
new crypto code, blackbox, aead internally

ThomasWaldmann:
crypto.pyx now implements a AEAD style API:
- encrypt-then-mac inside encrypt(), optionally mac-ing additional data
- auth-then-decrypt inside decrypt(), optionally auth-ing additional data

Implements:
- aes-ctr-hmac-sha256, aes-gcm
- aes-ocb, chacha20-poly1305 (openssl 1.1)
- simple AES, still needed by the keyfile code
- unencrypted/unauthenticated (so we do not need to specialcase that)

Note about iv/nonce/counter handling:

cipher.encrypt() and .decrypt() always do a full openssl context init, thus it is required for encrypt to set a IV before the call (or give iv as encrypt() param). there is no IV state kept inside the openssl ctx which is used for next encrypt() call.

set_iv sets the iv to use for next encrypt() call, it also resets the blocks counter that is used to reflect how many cipher blocks were encrypted with the last encrypt() call. using that, next_iv() computes the IV you shall use for the next encrypt() call.

this simulates the internal openssl IV state, which is in the opaque ctx since 1.1.

TODO:
- [ ] rename to cipher_key, mac_key and id_key?
- [ ] collapse some fixups

Anything else?

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -25,6 +25,7 @@ from .cache import ChunkListEntry
 from .crypto.key import key_factory
 from .compress import Compressor, CompressionSpec
 from .constants import *  # NOQA
+from .crypto.low_level import IntegrityError as IntegrityErrorBase
 from .hashindex import ChunkIndex, ChunkIndexEntry, CacheSynchronizer
 from .helpers import Manifest
 from .helpers import hardlinkable
@@ -1148,7 +1149,7 @@ class ArchiveChecker:
         else:
             try:
                 self.manifest, _ = Manifest.load(repository, (Manifest.Operation.CHECK,), key=self.key)
-            except IntegrityError as exc:
+            except IntegrityErrorBase as exc:
                 logger.error('Repository manifest is corrupted: %s', exc)
                 self.error_found = True
                 del self.chunks[Manifest.MANIFEST_ID]
@@ -1211,11 +1212,11 @@ class ArchiveChecker:
                 chunk_id = chunk_ids_revd.pop(-1)  # better efficiency
                 try:
                     encrypted_data = next(chunk_data_iter)
-                except (Repository.ObjectNotFound, IntegrityError) as err:
+                except (Repository.ObjectNotFound, IntegrityErrorBase) as err:
                     self.error_found = True
                     errors += 1
                     logger.error('chunk %s: %s', bin_to_hex(chunk_id), err)
-                    if isinstance(err, IntegrityError):
+                    if isinstance(err, IntegrityErrorBase):
                         defect_chunks.append(chunk_id)
                     # as the exception killed our generator, make a new one for remaining chunks:
                     if chunk_ids_revd:
@@ -1225,7 +1226,7 @@ class ArchiveChecker:
                     _chunk_id = None if chunk_id == Manifest.MANIFEST_ID else chunk_id
                     try:
                         self.key.decrypt(_chunk_id, encrypted_data)
-                    except IntegrityError as integrity_error:
+                    except IntegrityErrorBase as integrity_error:
                         self.error_found = True
                         errors += 1
                         logger.error('chunk %s, integrity error: %s', bin_to_hex(chunk_id), integrity_error)
@@ -1254,7 +1255,7 @@ class ArchiveChecker:
                         encrypted_data = self.repository.get(defect_chunk)
                         _chunk_id = None if defect_chunk == Manifest.MANIFEST_ID else defect_chunk
                         self.key.decrypt(_chunk_id, encrypted_data)
-                    except IntegrityError:
+                    except IntegrityErrorBase:
                         # failed twice -> get rid of this chunk
                         del self.chunks[defect_chunk]
                         self.repository.delete(defect_chunk)
@@ -1295,7 +1296,7 @@ class ArchiveChecker:
             cdata = self.repository.get(chunk_id)
             try:
                 data = self.key.decrypt(chunk_id, cdata)
-            except IntegrityError as exc:
+            except IntegrityErrorBase as exc:
                 logger.error('Skipping corrupted chunk: %s', exc)
                 self.error_found = True
                 continue
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -358,8 +358,10 @@ class AESKeyBase(KeyBase):
 
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
-        self.nonce_manager.ensure_reservation(self.cipher.block_count(len(data)))
-        return self.cipher.encrypt(data, header=self.TYPE_STR)
+        next_nonce = int.from_bytes(self.cipher.next_iv(), byteorder='big')
+        next_nonce = self.nonce_manager.ensure_reservation(next_nonce, self.cipher.block_count(len(data)))
+        iv = next_nonce.to_bytes(self.cipher.iv_len, byteorder='big')
+        return self.cipher.encrypt(data, header=self.TYPE_STR, iv=iv)
 
     def decrypt(self, id, data, decompress=True):
         if not (data[0] == self.TYPE or
@@ -401,7 +403,7 @@ class AESKeyBase(KeyBase):
             manifest_blocks = num_cipher_blocks(len(manifest_data))
             nonce = self.cipher.extract_iv(manifest_data) + manifest_blocks
         self.cipher.set_iv(nonce.to_bytes(16, byteorder='big'))
-        self.nonce_manager = NonceManager(self.repository, self.cipher, nonce)
+        self.nonce_manager = NonceManager(self.repository, nonce)
 
 
 class Passphrase(str):
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -359,14 +359,14 @@ class AESKeyBase(KeyBase):
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
         self.nonce_manager.ensure_reservation(self.cipher.block_count(len(data)))
-        return self.cipher.encrypt(data, header=self.TYPE_STR, aad_offset=1)
+        return self.cipher.encrypt(data, header=self.TYPE_STR)
 
     def decrypt(self, id, data, decompress=True):
         if not (data[0] == self.TYPE or
             data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
             id_str = bin_to_hex(id) if id is not None else '(unknown)'
             raise IntegrityError('Chunk %s: Invalid encryption envelope' % id_str)
-        payload = self.cipher.decrypt(data, header_len=1, aad_offset=1)
+        payload = self.cipher.decrypt(data)
         if not decompress:
             return payload
         data = self.decompress(payload)
@@ -385,7 +385,7 @@ class AESKeyBase(KeyBase):
             self.chunk_seed = self.chunk_seed - 0xffffffff - 1
 
     def init_ciphers(self, manifest_data=None):
-        self.cipher = CIPHERSUITE(mac_key=self.enc_hmac_key, enc_key=self.enc_key)
+        self.cipher = CIPHERSUITE(mac_key=self.enc_hmac_key, enc_key=self.enc_key, header_len=1, aad_offset=1)
         if manifest_data is None:
             nonce = 0
         else:
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -11,7 +11,7 @@ from hmac import HMAC, compare_digest
 
 import msgpack
 
-from borg.logger import create_logger
+from ..logger import create_logger
 
 logger = create_logger()
 
@@ -25,10 +25,10 @@ from ..helpers import get_limited_unpacker
 from ..helpers import bin_to_hex
 from ..item import Key, EncryptedKey
 from ..platform import SaveFile
-from .nonces import NonceManager
-from .low_level import AES, bytes_to_long, bytes_to_int, num_aes_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
 
-PREFIX = b'\0' * 8
+from .nonces import NonceManager
+from .low_level import AES, bytes_to_long, long_to_bytes, bytes_to_int, num_aes_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
+from .low_level import AES256_CTR_HMAC_SHA256 as CIPHERSUITE
 
 
 class PassphraseWrong(Error):
@@ -352,35 +352,21 @@ class AESKeyBase(KeyBase):
 
     PAYLOAD_OVERHEAD = 1 + 32 + 8  # TYPE + HMAC + NONCE
 
-    MAC = hmac_sha256
+    MAC = hmac_sha256  # TODO: not used yet
 
     logically_encrypted = True
 
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
         self.nonce_manager.ensure_reservation(num_aes_blocks(len(data)))
-        self.enc_cipher.reset()
-        data = b''.join((self.enc_cipher.iv[8:], self.enc_cipher.encrypt(data)))
-        assert (self.MAC is blake2b_256 and len(self.enc_hmac_key) == 128 or
-                self.MAC is hmac_sha256 and len(self.enc_hmac_key) == 32)
-        hmac = self.MAC(self.enc_hmac_key, data)
-        return b''.join((self.TYPE_STR, hmac, data))
+        return self.enc_cipher.encrypt(data, header=self.TYPE_STR, aad_offset=1)
 
     def decrypt(self, id, data, decompress=True):
         if not (data[0] == self.TYPE or
             data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
             id_str = bin_to_hex(id) if id is not None else '(unknown)'
             raise IntegrityError('Chunk %s: Invalid encryption envelope' % id_str)
-        data_view = memoryview(data)
-        hmac_given = data_view[1:33]
-        assert (self.MAC is blake2b_256 and len(self.enc_hmac_key) == 128 or
-                self.MAC is hmac_sha256 and len(self.enc_hmac_key) == 32)
-        hmac_computed = memoryview(self.MAC(self.enc_hmac_key, data_view[33:]))
-        if not compare_digest(hmac_computed, hmac_given):
-            id_str = bin_to_hex(id) if id is not None else '(unknown)'
-            raise IntegrityError('Chunk %s: Encryption envelope checksum mismatch' % id_str)
-        self.dec_cipher.reset(iv=PREFIX + data[33:41])
-        payload = self.dec_cipher.decrypt(data_view[41:])
+        payload = self.enc_cipher.decrypt(data, header_len=1, aad_offset=1)
         if not decompress:
             return payload
         data = self.decompress(payload)
@@ -406,9 +392,9 @@ class AESKeyBase(KeyBase):
             self.chunk_seed = self.chunk_seed - 0xffffffff - 1
 
     def init_ciphers(self, manifest_nonce=0):
-        self.enc_cipher = AES(is_encrypt=True, key=self.enc_key, iv=manifest_nonce.to_bytes(16, byteorder='big'))
+        self.enc_cipher = CIPHERSUITE(mac_key=self.enc_hmac_key, enc_key=self.enc_key,
+                                      iv=manifest_nonce.to_bytes(16, byteorder='big'))
         self.nonce_manager = NonceManager(self.repository, self.enc_cipher, manifest_nonce)
-        self.dec_cipher = AES(is_encrypt=False, key=self.enc_key)
 
 
 class Passphrase(str):
@@ -772,7 +758,7 @@ class Blake2KeyfileKey(ID_BLAKE2b_256, KeyfileKey):
     STORAGE = KeyBlobStorage.KEYFILE
 
     FILE_ID = 'BORG_KEY'
-    MAC = blake2b_256
+    MAC = blake2b_256  # TODO: not used yet
 
 
 class Blake2RepoKey(ID_BLAKE2b_256, RepoKey):
@@ -781,7 +767,7 @@ class Blake2RepoKey(ID_BLAKE2b_256, RepoKey):
     ARG_NAME = 'repokey-blake2'
     STORAGE = KeyBlobStorage.REPO
 
-    MAC = blake2b_256
+    MAC = blake2b_256  # TODO: not used yet
 
 
 class AuthenticatedKeyBase(RepoKey):
@@ -816,7 +802,8 @@ class AuthenticatedKeyBase(RepoKey):
 
     def decrypt(self, id, data, decompress=True):
         if data[0] != self.TYPE:
-            raise IntegrityError('Chunk %s: Invalid envelope' % bin_to_hex(id))
+            id_str = bin_to_hex(id) if id is not None else '(unknown)'
+            raise IntegrityError('Chunk %s: Invalid envelope' % id_str)
         payload = memoryview(data)[1:]
         if not decompress:
             return payload
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -802,6 +802,10 @@ class AuthenticatedKeyBase(RepoKey):
             raise IntegrityError('Manifest: Invalid encryption envelope')
         return 42
 
+    def init_ciphers(self, manifest_data=None):
+        if manifest_data is not None and manifest_data[0] != self.TYPE:
+            raise IntegrityError('Manifest: Invalid encryption envelope')
+
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
         return b''.join([self.TYPE_STR, data])
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -359,14 +359,14 @@ class AESKeyBase(KeyBase):
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
         self.nonce_manager.ensure_reservation(num_aes_blocks(len(data)))
-        return self.enc_cipher.encrypt(data, header=self.TYPE_STR, aad_offset=1)
+        return self.cipher.encrypt(data, header=self.TYPE_STR, aad_offset=1)
 
     def decrypt(self, id, data, decompress=True):
         if not (data[0] == self.TYPE or
             data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
             id_str = bin_to_hex(id) if id is not None else '(unknown)'
             raise IntegrityError('Chunk %s: Invalid encryption envelope' % id_str)
-        payload = self.enc_cipher.decrypt(data, header_len=1, aad_offset=1)
+        payload = self.cipher.decrypt(data, header_len=1, aad_offset=1)
         if not decompress:
             return payload
         data = self.decompress(payload)
@@ -392,9 +392,9 @@ class AESKeyBase(KeyBase):
             self.chunk_seed = self.chunk_seed - 0xffffffff - 1
 
     def init_ciphers(self, manifest_nonce=0):
-        self.enc_cipher = CIPHERSUITE(mac_key=self.enc_hmac_key, enc_key=self.enc_key,
-                                      iv=manifest_nonce.to_bytes(16, byteorder='big'))
-        self.nonce_manager = NonceManager(self.repository, self.enc_cipher, manifest_nonce)
+        self.cipher = CIPHERSUITE(mac_key=self.enc_hmac_key, enc_key=self.enc_key,
+                                  iv=manifest_nonce.to_bytes(16, byteorder='big'))
+        self.nonce_manager = NonceManager(self.repository, self.cipher, manifest_nonce)
 
 
 class Passphrase(str):
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -27,7 +27,7 @@ from ..item import Key, EncryptedKey
 from ..platform import SaveFile
 
 from .nonces import NonceManager
-from .low_level import AES, bytes_to_long, long_to_bytes, bytes_to_int, num_aes_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
+from .low_level import AES, bytes_to_long, long_to_bytes, bytes_to_int, num_cipher_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
 from .low_level import AES256_CTR_HMAC_SHA256 as CIPHERSUITE
 
 
@@ -514,7 +514,7 @@ class PassphraseKey(ID_HMAC_SHA_256, AESKeyBase):
             key.init(repository, passphrase)
             try:
                 key.decrypt(None, manifest_data)
-                num_blocks = num_aes_blocks(len(manifest_data) - 41)
+                num_blocks = num_cipher_blocks(len(manifest_data) - 41)
                 key.init_ciphers(key.extract_nonce(manifest_data) + num_blocks)
                 key._passphrase = passphrase
                 return key
@@ -554,7 +554,7 @@ class KeyfileKeyBase(AESKeyBase):
         else:
             if not key.load(target, passphrase):
                 raise PassphraseWrong
-        num_blocks = num_aes_blocks(len(manifest_data) - 41)
+        num_blocks = num_cipher_blocks(len(manifest_data) - 41)
         key.init_ciphers(key.extract_nonce(manifest_data) + num_blocks)
         key._passphrase = passphrase
         return key
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -366,7 +366,10 @@ class AESKeyBase(KeyBase):
             data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
             id_str = bin_to_hex(id) if id is not None else '(unknown)'
             raise IntegrityError('Chunk %s: Invalid encryption envelope' % id_str)
-        payload = self.cipher.decrypt(data)
+        try:
+            payload = self.cipher.decrypt(data)
+        except IntegrityError as e:
+            raise IntegrityError("Chunk %s: Could not decrypt [%s]" % (bin_to_hex(id), str(e)))
         if not decompress:
             return payload
         data = self.decompress(payload)
@@ -391,7 +394,7 @@ class AESKeyBase(KeyBase):
         else:
             if not (manifest_data[0] == self.TYPE or
                     manifest_data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
-                raise IntegrityError('Invalid encryption envelope')
+                raise IntegrityError('Manifest: Invalid encryption envelope')
             # manifest_blocks is a safe upper bound on the amount of cipher blocks needed
             # to encrypt the manifest. depending on the ciphersuite and overhead, it might
             # be a bit too high, but that does not matter.
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -590,7 +590,7 @@ class KeyfileKeyBase(AESKeyBase):
         assert enc_key.version == 1
         assert enc_key.algorithm == 'sha256'
         key = passphrase.kdf(enc_key.salt, enc_key.iterations, 32)
-        data = AES(is_encrypt=False, key=key).decrypt(enc_key.data)
+        data = AES(key, b'\0'*16).decrypt(enc_key.data)
         if hmac_sha256(key, data) == enc_key.hash:
             return data
 
@@ -599,7 +599,7 @@ class KeyfileKeyBase(AESKeyBase):
         iterations = PBKDF2_ITERATIONS
         key = passphrase.kdf(salt, iterations, 32)
         hash = hmac_sha256(key, data)
-        cdata = AES(is_encrypt=True, key=key).encrypt(data)
+        cdata = AES(key, b'\0'*16).encrypt(data)
         enc_key = EncryptedKey(
             version=1,
             salt=salt,
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -791,17 +791,6 @@ class AuthenticatedKeyBase(RepoKey):
         super().save(target, passphrase)
         self.logically_encrypted = False
 
-    def extract_nonce(self, payload):
-        # This is called during set-up of the AES ciphers we're not actually using for this
-        # key. Therefore the return value of this method doesn't matter; it's just around
-        # to not have it crash should key identification be run against a very small chunk
-        # by "borg check" when the manifest is lost. (The manifest is always large enough
-        # to have the original method read some garbage from bytes 33-41). (Also, the return
-        # value must be larger than the 41 byte bloat of the original format).
-        if payload[0] != self.TYPE:
-            raise IntegrityError('Manifest: Invalid encryption envelope')
-        return 42
-
     def init_ciphers(self, manifest_data=None):
         if manifest_data is not None and manifest_data[0] != self.TYPE:
             raise IntegrityError('Manifest: Invalid encryption envelope')
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -358,10 +358,9 @@ class AESKeyBase(KeyBase):
 
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
-        next_nonce = int.from_bytes(self.cipher.next_iv(), byteorder='big')
-        next_nonce = self.nonce_manager.ensure_reservation(next_nonce, self.cipher.block_count(len(data)))
-        iv = next_nonce.to_bytes(self.cipher.iv_len, byteorder='big')
-        return self.cipher.encrypt(data, header=self.TYPE_STR, iv=iv)
+        next_iv = self.nonce_manager.ensure_reservation(self.cipher.next_iv(),
+                                                        self.cipher.block_count(len(data)))
+        return self.cipher.encrypt(data, header=self.TYPE_STR, iv=next_iv)
 
     def decrypt(self, id, data, decompress=True):
         if not (data[0] == self.TYPE or
@@ -402,7 +401,7 @@ class AESKeyBase(KeyBase):
             # be a bit too high, but that does not matter.
             manifest_blocks = num_cipher_blocks(len(manifest_data))
             nonce = self.cipher.extract_iv(manifest_data) + manifest_blocks
-        self.cipher.set_iv(nonce.to_bytes(16, byteorder='big'))
+        self.cipher.set_iv(nonce)
         self.nonce_manager = NonceManager(self.repository, nonce)
 
 
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -358,7 +358,7 @@ class AESKeyBase(KeyBase):
 
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
-        self.nonce_manager.ensure_reservation(num_aes_blocks(len(data)))
+        self.nonce_manager.ensure_reservation(self.cipher.block_count(len(data)))
         return self.cipher.encrypt(data, header=self.TYPE_STR, aad_offset=1)
 
     def decrypt(self, id, data, decompress=True):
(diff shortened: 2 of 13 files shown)


enkore:
Separate branch for this?

ThomasWaldmann:
don't think this is necessary. this is mostly a internal interface change.

and I don't want to have a lot of feature creep in this PR, but get it done quickly and merge into master.

enkore:
Yep, that occurred to me as well and I deleted it (you're too fast ;)

borg/crypto.pyx
```diff
@@ -63,93 +74,222 @@ def num_aes_blocks(int length):
...
+        # store only lower 8 bytes, upper 8 bytes are assumed to be 0
+        cdef int i
+        for i in range(8):  # XXX easier way?
+            odata[32+i] = iv[8+i]
+
+
+cdef class AES_GCM_256_GMAC:
+    # Layout: GMAC 16 + IV 16 + CT (borg 1.2)
+    # additionally, each chunk starts with a type byte,
+    # which is not passed to this code or added by this code.
```
enkore: IIRC for AES-GCM a 12 byte nonce/IV is recommended.

borg/crypto.pyx
```diff
@@ -63,93 +74,222 @@ def num_aes_blocks(int length):
...
-            PyBuffer_Release(&data_buf)
+            free(odata)
+            PyBuffer_Release(&idata)
+
+    cdef fetch_iv(self, unsigned char * idata):
+        return idata[16:32]
+
+    cdef store_iv(self, unsigned char * odata, unsigned char * iv):
+        cdef int i
+        for i in range(16):  # XXX easier way?
```
enkore: slice assignment

enkore:
Do we want so split the implementation of cipher suites in two parts ("low level" part in crypto.pyx and "high level" part in key.py), or just keep it all in one place? I think it would be easier to just have it all in one place.

borg/crypto.pyx
ThomasWaldmann: core dumped :P

borg/crypto.pyx
ThomasWaldmann: well, it is the default. but recommended / why not more?

i mainly wanted to avoid the byte-frickling as seen in the legacy class.

borg/crypto.pyx
enkore: Cython bug?

```py
    cdef store_iv(self, unsigned char * odata, unsigned char * iv):
        odata[16:32] = iv
```

=

```
  /* "borg/crypto.pyx":290
 * 
 *     cdef store_iv(self, unsigned char * odata, unsigned char * iv):
 *         odata[16:32] = iv             # <<<<<<<<<<<<<<
 *         #cdef int i
 *         #for i in range(16):  # XXX easier way?
 */
  __pyx_t_1 = __Pyx_PyBytes_FromCString(__pyx_v_iv); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 290, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
  if (__Pyx_PyObject_SetSlice(((PyObject *)__pyx_v_odata), __pyx_t_1, 16, 32, NULL, NULL, NULL, 1, 1, 1) < 0) __PYX_ERR(0, 290, __pyx_L1_error)
  __Pyx_DECREF(__pyx_t_1); __pyx_t_1 = 0;
```

__pyx_v_odata is the unsigned char* as passed in.

Wtf?

enkore: Well then. libc.memcpy

ThomasWaldmann:
I'ld keep encrypt/mac and auth/decrypt in one (low-level) piece - like it is now - they can also deal with reading / writing the bytes from / to the right places into the buffer.

enkore:
Ok. Don't forget to add additional-data/header support though. We'll need that to authenticate the DKID.

ThomasWaldmann:
OK, I'll add support for "external" aad (not just the iv).

enkore:
I read some more about AES-GCM. RFC5084 and others recommend the 12 byte nonce/IV for efficiency. Other sources (e.g. https://www.cryptopp.com/wiki/GCM_Mode) also recommend it for the same reason.

The authentication tag is always calculated as 128 bits wide and only truncated at the end of the operation to the requested size, so there are no performance gains in using a smaller authentication tag. RFC5084 gives no rationale for their recommendation here. In RFC5288 (AES-GCM for TLS 1.2) they use a 128 bit tag and a 96 bit nonce/IV. In RFC4106 the nonce is constructed from a 64 bit IV and a 32 bit salt, so it's 96 bits again, however, the only *required* tag size is 128 bit:

> Implementations MUST support a full-length 16-octet ICV, and MAY
   support 8 or 12 octet ICVs, and MUST NOT support other ICV lengths.

In http://csrc.nist.gov/publications/nistpubs/800-38D/SP-800-38D.pdf they also recommend the 96 bit nonce/IV for "simplicity and efficiency". This paper also says that for IVs >= 96 bits *only* a random IV for at least 96 bits should be used; so they recommend a completely random 96 bit IV (section 8.2).

Also:

> A loss of power to the module shall not cause the repetition of IVs.  If the generation unit 
cannot  recover  from  a  loss  of  power,  then  
the  authenticated  encryption  function  shall  enter a failure state until a fresh key can be established

Which wouldn't be a problem with a good random number generator for the IV.

---------------------

Summary: Stick with 96 bit IV as everyone recommends, use a 128 bit tag for maximum security w/ no drawbacks?

ThomasWaldmann:
Yes, I'll change it. 96bit mac felt crappy anyway.

enkore:
If I understood correctly there is an internal per-message counter in GCM appended to the external IV, hence there is no IV to be retrieved.

Names:

AES_CTR_..._legacy would serve both as the lower abstraction (formerly crypto.AES) for both the old (fixed) crypto code and for a DEK cipher suite? It's not really legacy, remaining a valid choice. (It is strongly unforgable "to 256 bits" instead of "just" 128 bits in AES-GCM).

AES_GCM_256_GMAC: _GMAC is kinda redundant, _GCM already implies authentication using GCM/GMAC.

borg/crypto.pyx
```diff
@@ -63,93 +83,244 @@ def num_aes_blocks(int length):
...
+        EVP_CIPHER_CTX_init(&self.ctx)
+
+    def __dealloc__(self):
+        EVP_CIPHER_CTX_cleanup(&self.ctx)
 
     @property
-    def iv(self):
-        return self.ctx.iv[:16]
+    def current_iv(self):
+        raise NotImplemented  # gcm mode does not maintain/increment the counter in self.ctx.iv
```
enkore: Since there's no interfacing mandating current_iv(): Just remove the method entirely?

Since this is sensitive/crypto code it might be a good idea to explain in a comment why this doesn't make sense with AES-GCM.

ThomasWaldmann: I'll solve this by providing next_iv function, based on self.iv.

ThomasWaldmann: ok, seems like wikipedia was misleading concerning the counter in gcm mode.
i'll update the names.

enkore:
For AES-GCM the message size (number of blocks) is not relevant to the 96 bit IV. When using "AES-GCM-CTR" (so to speak) the 96 bit IV represents the number of messages (EVP_Encrypt... EncryptFinal) encrypted, not the number of blocks.

I think it would be simpler to use a random 96 bit IV for AES-GCM and keep the IV handling as-is for the existing AES-CTR code.

ThomasWaldmann:
Yeah, I realized that after your previous comment. It would not malfunction btw, it would just be wasteful with counter values.

Random IVs is not part of this PR, so I'll just make it work in the "counting way" as we currently do it.

enkore:
> It would not malfunction btw, it would just be wasteful with counter values.

Hmm, I don't see how. The IV is initialized from the manifest encryption envelope, and the manifest timestamp is saved locally to prevent obvious replay attacks.

Suppose the following series of events:

1. IV starts at *x0*
2. Some data blocks *n-1* are written
3. Manifest is written
4. IV of manifest is now x0+n = x1
5. Repository.commit; attacker raises false abort/exception
6. Attacker still has the blocks encrypted with IVs x0 ... x0+n
7. Another session, unclean commit state is detected, rollback is made (as usual)
8. The old manifest is in place with IV x0
9. This is not detected since the commit aborted before the cache was comitted
10. Some data is encrypted, with reused IVs x0 and following

ThomasWaldmann:
I am not fixing existing potential attacks on the IV here, I am just implementing a new internal interface and want to make it work in a similar way as it worked until now.

enkore:
> So, to get IV stuff fixed, I suggest you file a bug about this first and we fix it outside of this PR.

Absolutely :)

> What I was referring to when saying "function correctly, but wasteful" was that incrementing by 1 (0 for empty) is required for correct gcm function, but I was incrementing by num_aes_blocks which is >= 1 (and also 0 for empty). But it was wasting 2^32 / 16 counter values in extreme case.

Ok. I think this is inevitable with AES-GCM and as intended by the GCM spec. In any case, not a problem - independent of CTR or random IVs one never wants to use a significant portion of the possible IV space.

borg/crypto.pyx
```diff
@@ -318,6 +329,7 @@ cdef class AES_GCM_256_GMAC:
             if not EVP_DecryptInit_ex(&self.ctx, EVP_aes_256_gcm(), NULL, NULL, NULL):
                 raise Exception('EVP_DecryptInit_ex failed')
             iv = self.fetch_iv(<unsigned char *> idata.buf)
+            self.set_iv(iv)
```
ThomasWaldmann: needed so that next_iv() works.

we need that after loading / decrypting manifest to know the next IV we may use.

ThomasWaldmann:
encrypt:

has aad support now, but aad is not added to the returned data, so caller would have to prepend/append it, which needs another memcpy / creates string garbage. just always prepend aad in output data?

even with that, we also need the type byte (which is not part of aad in legacy mode), add a "anad" parameter for additional non-authenticated data and just prepend it at the very beginning?

decrypt: nothing special needed, just need to check whether it works with memoryviews for data and aad, so we can give 2 views for them at different offsets / with different lengths on the whole input data.

ThomasWaldmann:
idea for encrypt:

just have a "header" and "aad" param and use it like: out = encrypt(data, header=type_byte+aad, aad=aad)

as type_byte+aad is of negligible size, this doesn't cause lots of overhead or garbage.

enkore:
There's no harm in also authenticating the type byte -- just a header argument that is authenticated and prepended?

enkore:
I agree. Maybe make the API "prefix" (unauthenticated prepended), "header"/"aad" (authenticated prepended) and "data" (auth. enc. data)?

ThomasWaldmann:
attic did not authenticate the type byte. we can do, but not for the legacy support mode.

borg/crypto.pyx
```diff
@@ -56,100 +77,319 @@ cdef Py_buffer ro_buffer(object data) except *:
...
+            if not HMAC_Init_ex(&self.hmac_ctx, self.mac_key, 32, EVP_sha256(), NULL):
+                raise Exception('HMAC_Init_ex failed')
+            if not HMAC_Update(&self.hmac_ctx, aad, alen):
+                raise Exception('HMAC_Update failed')
+            if not HMAC_Update(&self.hmac_ctx, <const unsigned char *> idata.buf+hlen+32, ilen-hlen-32):
+                raise Exception('HMAC_Update failed')
+            if not HMAC_Final(&self.hmac_ctx, hmac_buf, NULL):
+                raise Exception('HMAC_Final failed')
+            if CRYPTO_memcmp(hmac_buf, idata.buf+hlen, 32):
+                raise Exception('Authentication failed')
```
enkore: IntegrityError

borg/crypto.pyx
```diff
@@ -56,100 +77,319 @@ cdef Py_buffer ro_buffer(object data) except *:
...
+            rc = EVP_DecryptUpdate(&self.ctx, odata+offset, &olen, <const unsigned char*> idata.buf+hlen+32+8, ilen-hlen-32-8)
+            if not rc:
+                raise Exception('EVP_DecryptUpdate failed')
+            offset += olen
+            rc = EVP_DecryptFinal_ex(&self.ctx, odata+offset, &olen)
+            if rc <= 0:
+                # this error check is very important for modes with padding or
+                # authentication. for them, a failure here means corrupted data.
+                # CTR mode does not use padding nor authentication.
+                raise Exception('EVP_DecryptFinal_ex failed')
```
enkore: The comment is outdated now (specific class for specific cipher + mode)

borg/crypto.pyx
```diff
@@ -56,100 +77,319 @@ cdef Py_buffer ro_buffer(object data) except *:
...
+            rc = EVP_DecryptFinal_ex(&self.ctx, odata+offset, &olen)
+            if rc <= 0:
                 # this error check is very important for modes with padding or
                 # authentication. for them, a failure here means corrupted data.
-                # CTR mode does not use padding nor authentication.
-                raise Exception('EVP_DecryptFinal failed')
-            ptl += outl
-            return out[:ptl]
+                # for GCM mode, a failure here means corrupted / tampered tag (mac) or data
+                raise Exception('EVP_DecryptFinal_ex failed')
```
enkore: IntegrityError

borg/crypto.pyx
```diff
@@ -74,6 +74,14 @@ bytes_to_long = lambda x, offset=0: _long.unpack_from(x, offset)[0]
 long_to_bytes = lambda x: _long.pack(x)
 
 
+class CryptoError(Exception):
+    """Malfunction in the crypto module."""
+
+
+class IntegrityError(CryptoError):
+    """Integrity checks failed. Corrupted or tampered data."""
```
enkore: `helpers.IntegrityError` <=> `crypto.IntegrityError`?

ThomasWaldmann: i wanted to avoid potentially cyclic imports and not import stuff from higher-level python modules into lower-level cython stuff, but yes, in the end 1 is enough.

ThomasWaldmann:
I am thinking about replacing decrypt's aad param by aad_offset.
We already have header_len, so aad_offset would just say where in the header the aad starts.
Of course that would require that the full aad is included in the header, but it always is, right?

enkore:
So

- `encrypt(data, header, aad_offset=0)` -> `envelope`
- `decrypt(envelope, header_len, aad_offset=0)` -> `data, header`

?

ThomasWaldmann:
Could be done like that, but we'ld need to peek into the header before even knowing which decrypt function we want to call.

enkore:
Yep. I'm thinking if there's a case where that would be useful, but none occur to me, so let's stick with `decrypt() -> data`

ThomasWaldmann:
OK, I am updating api / code to:
```
    encrypt(data, header, aad_offset=0) -> envelope
    decrypt(envelope, header_len, aad_offset=0) -> data
```

ThomasWaldmann:
collapse changesets?

borg/crypto.pyx
```diff
@@ -1,11 +1,42 @@
-"""A thin OpenSSL wrapper"""
+"""An AEAD style OpenSSL wrapper
+
+Note: AES-GCM mode needs OpenSSL >= 1.0.1d due to bug fixes in OpenSSL.
```
enkore: Also in docs/installation.rst ("From Source" section)

ThomasWaldmann: this changeset is still intentionally isolated (just crypto.pyx + test), integration will happen in 1.2.

borg/crypto.pyx
```diff
@@ -44,112 +92,337 @@ import struct
...
+            if not HMAC_Update(&self.hmac_ctx, <const unsigned char *> hdata.buf+aoffset, alen):
+                raise CryptoError('HMAC_Update failed')
+            if not HMAC_Update(&self.hmac_ctx, odata+hlen+32, offset-hlen-32):
+                raise CryptoError('HMAC_Update failed')
+            if not HMAC_Final(&self.hmac_ctx, odata+hlen, NULL):
+                raise CryptoError('HMAC_Final failed')
+            return odata[:offset]
+        finally:
+            free(odata)
+            PyBuffer_Release(&idata)
```
enkore: `PyBuffer_Release(&hdata)` missing (allocated in line 182)

borg/crypto.pyx
```diff
@@ -44,112 +92,337 @@ import struct
...
+        cdef Py_buffer hdata = ro_buffer(header)
+        cdef int ilen = len(data)
+        cdef int hlen = len(header)
+        cdef int aoffset = aad_offset
+        cdef int alen = hlen - aoffset
+        cdef unsigned char *odata = <unsigned char *>malloc(hlen + 32 + 8 + ilen + 16)
+        cdef int olen
+        cdef int offset
+        if not odata:
+            raise MemoryError
```
enkore: In this case idata, hdata are leaked

borg/crypto.pyx
```diff
@@ -44,112 +92,337 @@ import struct
...
+        cdef int ilen = len(envelope)
+        cdef int hlen = header_len
+        cdef int aoffset = aad_offset
+        cdef int alen = hlen - aoffset
+        cdef unsigned char *odata = <unsigned char *>malloc(ilen + 16)
+        cdef int olen
+        cdef int offset
+        cdef unsigned char hmac_buf[32]
+        if not odata:
+            raise MemoryError
```
enkore: Same as in encrypt (idata leak)

borg/crypto.pyx
```diff
@@ -44,112 +92,337 @@ import struct
...
-        cdef int inl = len(data)
-        cdef int ptl = 0
-        cdef int outl = 0
-        # note: modes that use padding, need up to one extra AES block (16b).
-        # This is what the openssl docs say. I am not sure this is correct,
-        # but OTOH it will not cause any harm if our buffer is a little bigger.
-        cdef unsigned char *out = <unsigned char *>malloc(inl+16)
-        if not out:
+            free(odata)
+            PyBuffer_Release(&idata)
```
enkore: Same as in aes_ctr, _Release of hdata is missing

borg/crypto.pyx
```diff
@@ -44,112 +92,337 @@ import struct
...
+        cdef Py_buffer hdata = ro_buffer(header)
+        cdef int ilen = len(data)
+        cdef int hlen = len(header)
+        cdef int aoffset = aad_offset
+        cdef int alen = hlen - aoffset
+        cdef unsigned char *odata = <unsigned char *>malloc(hlen + 16 + 12 + ilen + 16)
+        cdef int olen
+        cdef int offset
+        if not odata:
             raise MemoryError
```
enkore: idata, hdata leak

enkore:
Hm, does AES_GCM update the IV when calling encrypt()? Doesn't look like it, so calling code would have to do that itself (next_iv/set_iv)? AES_CTR still does it via OpenSSL internally (I think).

ThomasWaldmann:
it currently works by doing that externally, always. there is a full init in encrypt().

enkore:
Ah, now I see that in AES_CTR as well (`EVP_EncryptInit_ex(..., self.iv)`)

enkore:
Since we do many small-object allocations here it is probably better to use PyMem_Alloc/Free instead of malloc/free (PyMem has many optimizations for small allocs).

ThomasWaldmann:
I am not completely happy with IV handling yet. So I just documented how it is now, but it might change later, when I have more experience with integrating other ciphers (e.g. from libsodium).

ThomasWaldmann:
about small mallocs: all the mallocs are about len(data), so for us this is usually ~0.5-8MB. Does that still count as "small"?

enkore:
It depends(tm) -- 500 kB probably not, but many files (system files, config/source/text files) are much smaller than that.

ThomasWaldmann:
ah, right, if the file is only 1 chunk...

ThomasWaldmann:
ok, I'll collapse all the changes now.

ThomasWaldmann:
rebased on master (== include openssl 1.1 support).
still have to check whether it actually works on 1.1.

ThomasWaldmann:
The way used for EVP_CIPHER_CTX (see PR #1193) to be compatible to 1.0 and 1.1 does not work for HMAC_CTX, because there is no HMAC_CTX_new/free in 1.0 (only in 1.1).

I submitted a request to openssl project, but acceptance has to be seen.

enkore:
Hmm, we could shim HMAC_CTX_new/free (malloc+init, cleanup+free)... even if they accept a patch for 1.0.2 it would likely take time until distros pick up on it, and those distros that run even older versions picking fixes back into their patchsets would never get it.

ThomasWaldmann:
If they put it in a security&compatibility update like 1.0.2i, it would not take long, likely less long as finishing borgbackup 1.2.

ThomasWaldmann:
I added the HMAC api change, so it works on openssl 1.1 and is consistent with the EVP_CIPHER change.

We can solve for 1.0 either by getting it from an openssl update or by implementing that shim.

src/borg/_crypto_helpers.c
```diff
@@ -0,0 +1,26 @@
...
+#include <openssl/opensslv.h>
+#include <openssl/hmac.h>
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+
+HMAC_CTX *HMAC_CTX_new(void)
+{
+   HMAC_CTX *ctx = OPENSSL_malloc(sizeof(*ctx));
+   if (ctx != NULL) {
+       memset(ctx, 0, sizeof *ctx);
```
enkore:
```
src/borg/_crypto_helpers.c: In function ‘HMAC_CTX_new’:
src/borg/_crypto_helpers.c:12:8: warning: implicit declaration of function ‘memset’ [-Wimplicit-function-declaration]
        memset(ctx, 0, sizeof *ctx);
        ^~~~~~
src/borg/_crypto_helpers.c:12:8: warning: incompatible implicit declaration of built-in function ‘memset’
src/borg/_crypto_helpers.c:12:8: note: include ‘<string.h>’ or provide a declaration of ‘memset’
```

src/borg/crypto.pyx
```diff
@@ -1,11 +1,59 @@
...
+# PyMem_Alloc/Free instead of malloc/free (PyMem has many optimizations for
+# small allocs)."
+#
+# Small mallocs currently happen if the total input file length is small, so
+# the 1 chunk's size is less than what the chunker would produce for big files.
 
 from libc.stdlib cimport malloc, free
 from cpython.buffer cimport PyBUF_SIMPLE, PyObject_GetBuffer, PyBuffer_Release
 
 API_VERSION = 3
```
enkore: Should be bumped to `4`

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
1GB data, i5-4200 cpu
```
4.10s aes256-hmac-sha256
1.28s aes256-gcm
1.15s aes256-ocb
1.52s for chacha20-poly1305
```

src/borg/crypto.pyx
```diff
@@ -480,23 +484,33 @@ cdef class _AEAD_BASE:
...
+
+cdef class _CHACHA_BASE(_AEAD_BASE):
+    def block_count(self, length):
+        return (length + 63) // 64
+
+
+cdef class AES256_GCM(_AES_BASE):
     def __init__(self, mac_key, enc_key, iv=None):
         if OPENSSL_VERSION_NUMBER < 0x10001040:
             raise ValueError('AES GCM requires OpenSSL >= 1.0.1d. Detected: OpenSSL %08x' % OPENSSL_VERSION_NUMBER)
```
enkore: Spin these off into a `@staticmethod` to make checking this easier?

ThomasWaldmann: except the "self" less, how is that easier?

enkore: that the check is here implies that this can vary from system to system. therefore it would be good to be able to check / list what is available, which is easier if that's a static method and doesn't require trial construction of these objects.

src/borg/crypto.pyx
```diff
@@ -320,21 +320,19 @@ cdef class AES256_CTR_HMAC_SHA256:
...
     cdef EVP_CIPHER_CTX *ctx
     cdef unsigned char *enc_key
     cdef unsigned char iv[12]
     cdef long long blocks
-    mode = ''  # override in subclass
 
     def __init__(self, mac_key, enc_key, iv=None):
         assert mac_key is None
         assert isinstance(enc_key, bytes) and len(enc_key) == 32
         assert iv is None or isinstance(iv, bytes) and len(iv) == 12
```
textshell: this (len(iv)==12) is still somewhat AES-GCM (and related) specific.

ThomasWaldmann: ocb and chacha20 have same default iirc.

src/borg/crypto.pyx
```diff
@@ -445,24 +459,24 @@ cdef class _AEAD_BASE:
...
         self.blocks = 0  # number of cipher blocks encrypted with this IV
-        for i in range(12):
+        for i in range(self.iv_len):
             self.iv[i] = iv[i]
 
     def next_iv(self):
         assert self.blocks < 2**32
         # we need 16 bytes for increment_iv:
-        last_iv = b'\0\0\0\0' + self.iv[:12]
+        last_iv = b'\0' * (16 - self.iv_len) + self.iv[:self.iv_len]
```
textshell: This really looks like increment_iv should be more flexible. python nowadays has decent number to big endian conversion with variable bit width.

ThomasWaldmann: yes, maybe. but i had also the idea to somehow standardize on 16 bytes.

textshell: How can we standardize on 16 bytes if some ciphers use 12 bytes? Any why would we want to try?

src/borg/key.py
```diff
@@ -171,22 +171,13 @@ def id_hash(self, data):
...
             data[0] == PassphraseKey.TYPE and isinstance(self, RepoKey)):
             raise IntegrityError('Invalid encryption envelope')
-        data_view = memoryview(data)
-        hmac_given = data_view[1:33]
-        hmac_computed = memoryview(hmac_sha256(self.enc_hmac_key, data_view[33:]))
-        if not compare_digest(hmac_computed, hmac_given):
-            raise IntegrityError('Encryption envelope checksum mismatch')
-        self.dec_cipher.reset(iv=PREFIX + data[33:41])
-        payload = self.dec_cipher.decrypt(data_view[41:])
+        payload = self.enc_cipher.decrypt(data, header_len=1, aad_offset=1)
```
textshell: Catch and reraise integrity errors here with (chunk) id?

src/borg/nonces.py
```diff
@@ -64,24 +64,20 @@ def ensure_reservation(self, nonce_space_needed):
 
         if self.end_of_nonce_reservation:
             # we already got a reservation, if nonce_space_needed still fits everything is ok
-            next_nonce = int.from_bytes(self.enc_cipher.iv, byteorder='big')
+            next_nonce_bytes = self.cipher.next_iv()
+            next_nonce = int.from_bytes(next_nonce_bytes, byteorder='big')
             assert next_nonce <= self.end_of_nonce_reservation
             if next_nonce + nonce_space_needed <= self.end_of_nonce_reservation:
+                self.cipher.set_iv(next_nonce_bytes)
```
textshell: i’m not sure i like setting the iv here. Now that iv is very much more explicit i wonder if next_iv should be a parameter and next_nonce_bytes the return of this method?

src/borg/nonces.py
```diff
@@ -64,24 +64,20 @@ def ensure_reservation(self, nonce_space_needed):
...
         reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
         assert reservation_end < MAX_REPRESENTABLE_NONCE
-        if self.end_of_nonce_reservation is None:
-            # initialization, reset the encryption cipher to the start of the reservation
-            self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
-        else:
-            # expand existing reservation if possible
-            if free_nonce_space != self.end_of_nonce_reservation:
-                # some other client got an interleaved reservation, skip partial space in old reservation to avoid overlap
-                self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
```
textshell: This removes the comments we agreed we wanted to keep. Not sure what to do about those now that there is really no difference.

src/borg/key.py
```diff
@@ -371,7 +371,7 @@ def encrypt_key_file(self, data, passphrase):
         iterations = PBKDF2_ITERATIONS
         key = passphrase.kdf(salt, iterations, 32)
         hash = hmac_sha256(key, data)
-        cdata = AES(is_encrypt=True, key=key).encrypt(data)
+        cdata = AES(key, b'\0'*16).encrypt(data)
```
textshell: Now that this is in plain view. Urgh.
Keyblob changes are really not very safe. This is more of a problem for the repokey case (because the server is more untrusted) but it’s bad in every case. Multiple time pad just is bad.

src/borg/crypto.pyx
```diff
@@ -551,98 +550,102 @@ cdef class CHACHA20_POLY1305(_CHACHA_BASE):
...
+    cdef unsigned char *enc_key
+    cdef int cipher_blk_len
+    cdef int iv_len
+    cdef unsigned char iv[16]
     cdef long long blocks
 
-    def __cinit__(self, is_encrypt, key, iv=None):
+    def __init__(self, enc_key, iv=None):
+        assert isinstance(enc_key, bytes) and len(enc_key) == 32
+        self.enc_key = enc_key
```
textshell: Why enc_key? It’s used for decryption too.

src/borg/crypto.pyx
```diff
@@ -551,98 +550,102 @@ cdef class CHACHA20_POLY1305(_CHACHA_BASE):
...
+        # set_iv needs to be called before each encrypt() call,
+        # because encrypt does a full initialisation of the cipher context.
+        assert isinstance(iv, bytes) and len(iv) == self.iv_len
+        self.blocks = 0  # number of cipher blocks encrypted with this IV
+        for i in range(self.iv_len):
+            self.iv[i] = iv[i]
+
+    def next_iv(self):
+        # call this after encrypt() to get the next iv for the next encrypt() call
+        return increment_iv(self.iv[:self.iv_len], self.blocks)
```
textshell: why slice here? If it really slices it is wrong, loosing parts of the IV. And if it doesn‘t slice it is not needed (increment doesn‘t mutate it‘s argument)

src/borg/key.py
```diff
@@ -171,7 +171,7 @@ def id_hash(self, data):
 
     def encrypt(self, chunk):
         chunk = self.compress(chunk)
-        self.nonce_manager.ensure_reservation(num_aes_blocks(len(chunk.data)))
+        self.nonce_manager.ensure_reservation(self.cipher.block_count(len(chunk.data)))
```
textshell: What we care about here is the amount of nonce space used. Which i think is not block_count in the new generalized setting.

textshell:
I feel the last commits really don‘t fit well into the roadmap. Things like extracting IVs really shouldn’t be needed for any of the new ciphers. But i have to think about the big picture some more.

I think we need a better picture how DEK and CLK will change how the new ciphers are used. And what is the best way to handle the things that the one and only legacy cipher needs.

src/borg/nonces.py
ThomasWaldmann: you mean moving all set_iv calls outside of nonces.py (so the caller does it)?

src/borg/key.py
ThomasWaldmann: I guess it is not in scope of this PR to change anything related to that.
Also, this is done once for repokey/keyfile key generation, so what's the problem?
Via the random salt, the aes key is different for each repo/keyfile key.

src/borg/crypto.pyx
ThomasWaldmann: this is a cython thing. self.iv basically is a pointer and to transform that into python bytes, one needs to slice it to right length.

ThomasWaldmann:
The changeset is just making things work with the new API. There is no DEK etc. yet, we are just counting the counter upwards. As the IV values now are at different offsets depending on the ciphersuite, the old [33:41] code did not work any more (at least not in general).

src/borg/crypto.pyx
ThomasWaldmann: it is enc_key (meaning encryption as well as decryption) vs. mac_key.

src/borg/key.py
ThomasWaldmann: yes, good idea.

src/borg/key.py
ThomasWaldmann: The self.cipher.block_count method is a precise thing, it does the same thing correctly for any cipher in use that num_aes_blocks did correct for AES.

note that the global block_count *function* can be (and is) used at some other places in a more sloppy way, when the cipher block size is not known and we just assume it is not less than 16B.

And there is no future code yet, so this is just making things work correctly for the present code.

src/borg/nonces.py
textshell: Yes, that’s what i was wondering about. But i think we also need to look at how the code will look in the cases without a nonce manager (i.e. modern ciphers)

src/borg/key.py
textshell: Yes, it is not in the scope of this PR. It‘s just that this PR makes the code obvious to see.

Why is this problematic? Maybe it is not in current borg, because we don‘t change the plaintext content. I have to think about what this means for change-passphrase (it reveals the xor of the two key streams at least). In future the keyblob will be editable to add new keys, i think we need to keep in mind that we need to come up with a solution at that time.

src/borg/crypto.pyx
textshell: ok, makes sense.

src/borg/key.py
ThomasWaldmann: https://github.com/borgbackup/borg/issues/1579

src/borg/crypto.pyx
ThomasWaldmann: ok, implementing that...

src/borg/crypto.pyx
ThomasWaldmann: in fact, we have even 3 different keys: enc_key, enc_hmac_key, id_(hmac_)key

enkore: Maybe do a rename sweep later (avoiding more conceptual/naming changes in this PR).

That being said a better naming convention may be cipher_key, mac_key and id_key.

ThomasWaldmann: sounds good. ticket?

src/borg/nonces.py
ThomasWaldmann: i moved the calls, changeset coming after tests...

ThomasWaldmann:
collapse / merge?

src/borg/testsuite/nonces.py
```diff
@@ -79,146 +59,137 @@ def set_cache_nonce(self, nonce):
...
+        next_nonce = manager.ensure_reservation(0x2000, 19)
+        assert next_nonce == 0x2000
 
         assert self.cache_nonce() == "0000000000002033"
         assert self.repository.next_free == 0x2033
 
         # enough space in reservation
-        manager.ensure_reservation(13)
-        cipher.expect_iv_and_advance(0x2013, 0x2000 + 19 + 13)
+        next_nonce = manager.ensure_reservation(0x2013, 13)
```
textshell: i’m not sure i like having 0x2013 here (and in all the similar cases) as number. The previous code had 0x2000 + 19 + ...  which at least still reveals a bit about it‘s construction. But i admit this test is not the nicest in any form. Maybe have a current nonce variable?

src/borg/crypto.pyx
```diff
@@ -385,14 +385,17 @@ cdef class AES256_CTR_HMAC_SHA256:
 
     def set_iv(self, iv):
         # set_iv needs to be called before each encrypt() call
+        if isinstance(iv, int):
+            iv = iv.to_bytes(self.iv_len, byteorder='big')
```
textshell: Do we still need iv as bytes? I wonder if we shouldn’t go the whole way and just have a python `int` as iv in the class and do the to_bytes only in store_iv and in calls to EVP_EncryptInit_ex/EVP_DecryptInit_ex.

Also it seems increment_iv is now unused

src/borg/archive.py
```diff
@@ -21,10 +21,11 @@
 from .chunker import Chunker
 from .compress import Compressor
 from .constants import *  # NOQA
+from .crypto import IntegrityError as IntegrityErrorBase
```
textshell: Do we still need IntegrityErrorBase if key.py wraps low level crypto IntegrityError anyway?

src/borg/nonces.py
textshell: I think we are loosing up to nonce_space_needed-1 of nonce space here, if the current reservation isn’t large enough, but free_nonce_space == self.end_of_nonce_reservation. The old code would not have set a new nonce so used the last part of the old reservation. The new code always returns free_nonce_space.

But the new code is still safe, just a tiny bit wasteful, so we don‘t have to care. Also currently reservations are huge.

src/borg/crypto.pyx
```diff
@@ -90,101 +189,566 @@ cdef Py_buffer ro_buffer(object data) except *:
...
+        self.blocks = 0  # number of cipher blocks encrypted with this IV
+        for i in range(self.iv_len):
+            self.iv[i] = iv[i]
+
+    def next_iv(self):
+        # call this after encrypt() to get the next iv (int) for the next encrypt() call
+        # AES-GCM, AES-OCB, CHACHA20 ciphers all add a internal 32bit counter to the 96bit
+        # (12 byte) IV we provide, thus we only need to increment the IV by 1 (and we must
+        # not encrypt more than 2^32 cipher blocks with same IV):
+        assert self.blocks < 2**32
```
textshell: Can we move this check into encrypt (and maybe decrypt) so it errors out before doing bogus cryptography? We can check this even before we pass any data to openssl (which might have some kind of error handling that we don‘t expect). Also i think this warrants a check that is not disabled in optimized runs (i.e. an if instead of an assert)

textshell:
Sorry for being slow in reviewing this.

i guess if we can agrees on something we can handle my unhappiness about the newer ciphers exposing CS.block_count, CS.extract_iv in their public api in a follow up.  Same for num_cipher_blocks.

This code currently assumes that we always want to store an iv. Which does not hold for all suites still in the discussion for the new cipher suites (e.g. key derivation from chunk id or per chunk random key). But that too could be resolved later.

src/borg/testsuite/nonces.py
ThomasWaldmann: i had a nonce variable in there at some time, but i got confused about where it should be used, so i removd it again.

src/borg/crypto.pyx
ThomasWaldmann: I'll have a look into that...

src/borg/archive.py
ThomasWaldmann: At the time I introduced ...Base, I was having troubles with the 2 different IntegrityError classes and this was an easy way to solve it.

src/borg/crypto.pyx
ThomasWaldmann: ok, sounds good.

src/borg/testsuite/nonces.py
textshell: One way to use it would be to simulate the ciphers iv tracking. i.e.
```
next_nonce = manager.ensure_reservation(nonce, 13)
assert next_nonce == nonce
nonce += 13
```
For where the nonce isn’t changed by the nonce manager. And a similar block with nonce = next_nonce + 13 for when the nonce manager changes it.

textshell:
I think for session key usage we want to bind the authentication to the really used session key. i.e. have the (decrypted) session key be part of the associated data.
This easily avoids key swap attempts. But with the header based api it is inconvenient.

enkore:
I don't think it's easily possible (putting the key for the AEAD in the AD is kinda pointless IMHO), but depending on what's used it might feel safer at least to authenticate just everything, including the session key used (eg CLK1-like scheme).

textshell:
@enkore Why do you think authenticating the session key is pointless? May protocols put "hidden" data into the AD (e.g. implicit message numbers). This especially makes sense when the authentication key is not also a session key, which is unneeded f.e. in HMAC based EtM constructions.

enkore:
No, I specifically said _putting the key for the AEAD in the AD is kinda pointless_.

textshell:
I think AEAD does not imply that every part of the key actually is used in authentication. A AEAD could use a multi component key and differences in the non auth key component(s) would go unnoticed. GCM derives it’s auth key from the main key, but other AEAD ciphers might not.

ThomasWaldmann:
btw, may I suggest that we do not extend the scope of this PR, but rather concentrate on getting the stuff already there into best shape? stuff like adding hidden data to auth (AD) can be added later easily.

ThomasWaldmann:
Rebased on current master to resolve conflicts, partially collapsed changesets.

src/borg/crypto.pyx
```diff
@@ -403,7 +395,84 @@ cdef class AES256_CTR_HMAC_SHA256:
...
+            raise IntegrityError('MAC Authentication failed')
+
+
+cdef class AES256_CTR_BLAKE2b(AES256_CTR_BASE):
+    cdef unsigned char *mac_key
+
+    def __init__(self, mac_key, enc_key, iv=None, header_len=1, aad_offset=1):
+        if len(mac_key) < 128:  # XXX dirty hack to get tests working - remove this when fixed XXX
+            print("warning, mac key too short: %d" % len(mac_key))
+            mac_key += b'\0' * (128 - len(mac_key))
```
ThomasWaldmann: i needed this because the key in the hardcoded keyfile data is only 32B long.

src/borg/testsuite/key.py
```diff
@@ -159,10 +158,9 @@ def test_keyfile2_kfenv(self, tmpdir, monkeypatch):
         key = KeyfileKey.detect(self.MockRepository(), self.keyfile2_cdata)
         assert key.decrypt(self.keyfile2_id, self.keyfile2_cdata).data == b'payload'
 
-    @pytest.mark.skip("temporarily disabled for branch merge")  # TODO
     def test_keyfile_blake2(self, monkeypatch, keys_dir):
         with keys_dir.join('keyfile').open('w') as fd:
-            fd.write(self.keyfile_blake2_key_file)
+            fd.write(self.keyfile_blake2_key_file)  # XXX does this have a 32byte mac key?
```
ThomasWaldmann: looks like it.

ThomasWaldmann:
rebased on current master.

enkore:
A rebased version of this branch is available there: https://github.com/enkore/borg/tree/ThomasWaldmann-crypto-aead

ThomasWaldmann:
@enkore thanks for rebasing!

ThomasWaldmann:
travis is half-unhappy.

also EISDIR again. i start to doubt that the debug flag changed/improved anything. maybe it was just too few test runs and random being a bitch. so maybe that changeset should be reverted.

enkore:
Travis seems a bit confused about branches there.

`$ git clone --depth=50 --branch=crypto-aead https://github.com/borgbackup/borg.git borgbackup/borg` isn't the correct repo.

ThomasWaldmann:
@enkore ok, let's see. I accidentally pushed to borgbackup/borg crypto-aead instead of ThomasWaldmann/borg crypto-aead (and deleted the first soon afterwards again).

ThomasWaldmann:
rebased onto master, finished the "WIP" in last changeset.

ThomasWaldmann:
@enkore @textshell as this can hopefully get merged soon (after 1.1-maint branch creation), maybe do final reviews now / soon. I'll also do a self-review later today.

src/borg/crypto/key.py
```diff
@@ -810,13 +804,18 @@ def extract_nonce(self, payload):
             raise IntegrityError('Manifest: Invalid encryption envelope')
         return 42
 
+    def init_ciphers(self, manifest_data=None):
+        if manifest_data is not None and manifest_data[0] != self.TYPE:
+            raise IntegrityError('Manifest: Invalid encryption envelope')
+
     def encrypt(self, chunk):
         data = self.compressor.compress(chunk)
         return b''.join([self.TYPE_STR, data])
```
ThomasWaldmann: guess we maybe could adapt/reuse lowlevel.UNENCRYPTED "dummy cipher" here and in PlaintextKey.
or kill it if we don't want it for that.

src/borg/algorithms/blake2-libselect.h
```diff
@@ -1,3 +1,8 @@
+/**
+ * Selects between system libb2 or vendored blake2 implementation.
+ * BORG_USE_LIBB2 is defined by the build driver (setup.py) if
+ * libb2 was detected and added to the include path.
+ */
```
ThomasWaldmann: guess that should be in 1.1 also.

enkore: That's already documented somewhere in the build docs

enkore:
Yes, don't add any more nogil things to this PR.

ThomasWaldmann:
ok, rebased again, collapsed some fixup changesets. guess i am finished here.

ThomasWaldmann:
\o/

Issue #908, opened 2016-04-14
Races between readdir/getdents64 and name access.


verygreen:
Currently borg reads the directory content, then iterates the names, does lstat on them and calls a processing function based on a file type.
This is racy as between lstat and open/whatever inside of the `Archive.process_*` method somebody might come in and replace the underlying file (atomically via rename, or via delete + new create).

If the scandir code in #905 is also extended to its logical conclusion and would skip lstat altogether, using file type to determine what process method to call, the race window would extend even more since there might be significant time passing between reading the directory and actually getting into some of the files later in the output.
Additional races are possible even inside of the process methods, some of which are documented in #906 

I suspect the correct way to deal with this would be to pass in inode number into the `process_*` method.
The very first thing such method would then do is to open the name presented (with possible exception of process_symlink, since symlinks cannot be opened).
Code needs to be prepared to get ENOENT on open/readlink. if this happens, this should be handled as if this name was not even in the readdir list (since we lost the race to access it anyway).
Once the open is done, fstat on the file descriptor will tell us if inode number has changed and along with it the file type so that we might need to call another processing function (if we can pass an open fd there, we can totally avoid additional racing there).
The rest of the processing function needs to keep using fd-based accesses since in a live filesystem a file name might start point to a different location at any moment and assuming it's always the same inode is unsafe.

Additional problem is present in `Archiver._process` method for the directory handling.
Upon noticing that we are dealing with a directory, first archive.process_directory is called on the file name, second, we do readdir or scandir on it, bt by the time we get to it, the directory might disappear (which is ok, I guess, we'll just have an entry for an empty dir), or it might be replaced with another file type (also ok, I guess, we still get the empty dir in the archive) or it might be replaced by another dir - this is more problematic as we'd then proceed to insert wrong dir content into the archive which might pose problems starting from just inconsistent data in the archive and then security issues (imagine if the first scan had permissible access, but by the time we got to read the contents - we got one that was a lot more restrictive).
Finally even worse variation of this would be file-replacement - since iteration is happening with the full path every time (yuck), if I have
/path/dir1/file1
/path/dir2/file1
and I rename dir2 to dir1 just at the right time, I am backing up wrong file1 while processing dir1 and I am none the wiser.
The fix here would be to move all the iterating into process_dir().
it would start as usual with open (O_DIRECTORY), then once we have that fd, we are golden, we just need to use fgetdirents64 so that we read this directory at the fd not at some inconclusive path.
[...]

Issue #906, opened 2016-04-14
Racy extended attributes querying.


verygreen:
While doing some unrelated stracing in #905 I noticed that the EA querying is racy and also suboptimal:

```
lstat("/mnt/test/1019429875\n", {st_mode=S_IFREG, st_size=16, ...}) = 0
open("/mnt/test/1019429875\n", O_RDONLY|O_NOATIME|O_CLOEXEC) = 5
fstat(5, {st_mode=S_IFREG, st_size=16, ...}) = 0
fstat(5, {st_mode=S_IFREG, st_size=16, ...}) = 0
ioctl(5, TCGETS, 0x7ffc41602cd0)        = -1 ENOTTY (Inappropriate ioctl for device)
fstat(5, {st_mode=S_IFREG, st_size=16, ...}) = 0
lseek(5, 0, SEEK_CUR)                   = 0
read(5, "\363C\303<\0\0\0\00010194298", 8388608) = 16
fadvise64(5, 0, 16, POSIX_FADV_DONTNEED) = 0
close(5)                                = 0
llistxattr("/mnt/test/1019429875\n", NULL, 0) = 0
getxattr("/mnt/test/1019429875\n", "system.posix_acl_access", 0x0, 0) = -1 ENODATA (No data available)
getxattr("/mnt/test/1019429875\n", "system.posix_acl_default", 0x0, 0) = -1 ENODATA (No data available)
```

llistxattr is racy and should be flistxattr BEFORE we close the file to be 100% sure the data and attributes are from the same file.
getxattr is even more wrong as it would not work correctly on symlinks too, reading the attrs from the symlink target, not symlink itself besides the actual race.

verygreen:
The suboptimallness is in doing llistxattr, and then even despite it returning empty list, still trying to fetch some extended attributes that we now know are not really there. (hm, is this just a platform-dependent thing and on other OSes listxattr would not return ACLs and the like?)

enkore:
Borg has generic xattr support (i.e. it listxattrs all attrs, then saves them), but also specific support for ACLs which are queried separately via acl_XXX functions -- those are probably making the getxattr system.posix_acl_default etc. calls.

enkore:
A related race is found in xattr.get_all: https://github.com/borgbackup/borg/blob/1.0-maint/borg/xattr.py#L28

[...]

PR #1037, opened 2016-05-11
Windows file permissions

Anakonda:
Get and set DACL and owner for files and folders.

Setting owner seems to require administrator.

### DIFF
--- a/borg/constants.py
+++ b/borg/constants.py
@@ -1,7 +1,7 @@
 # this set must be kept complete, otherwise the RobustUnpacker might malfunction:
 ITEM_KEYS = set([b'path', b'source', b'rdev', b'chunks', b'hardlink_master',
                  b'mode', b'user', b'group', b'uid', b'gid', b'mtime', b'atime', b'ctime',
-                 b'xattrs', b'bsdflags', b'acl_nfs4', b'acl_access', b'acl_default', b'acl_extended', ])
+                 b'xattrs', b'bsdflags', b'acl_nfs4', b'acl_access', b'acl_default', b'acl_extended', b'win_dacl'])
 
 ARCHIVE_TEXT_KEYS = (b'name', b'comment', b'hostname', b'username', b'time', b'time_end')
 ITEM_TEXT_KEYS = (b'path', b'source', b'user', b'group')
--- a/borg/platform_windows.pyx
+++ b/borg/platform_windows.pyx
@@ -1,3 +1,5 @@
+#cython: language_level=3
+
 import json
 from libc.stddef cimport wchar_t
 from libc.stdint cimport uint16_t, uint32_t, uint64_t
@@ -9,16 +11,19 @@ from .helpers import safe_decode, safe_encode
 
 API_VERSION = 2
 
+
 cdef extern from 'stdlib.h':
     void free(void* ptr)
     void* malloc(size_t)
     void* calloc(size_t, size_t)
+
 
 cdef extern from 'Python.h':
     wchar_t* PyUnicode_AsWideCharString(object, Py_ssize_t *)
     object PyUnicode_FromWideChar(const wchar_t*, Py_ssize_t)
     void* PyMem_Malloc(int)
     void PyMem_Free(void*)
+
 
 cdef extern from 'windows.h':
     ctypedef int HLOCAL
@@ -36,6 +41,9 @@ cdef extern from 'windows.h':
     DWORD GetLastError();
     void SetLastError(DWORD)
 
+    DWORD FormatMessageW(DWORD, void*, DWORD, DWORD, wchar_t**, DWORD, void*)
+
+
     BOOL InitializeSecurityDescriptor(BYTE*, DWORD)
 
     BOOL LookupAccountNameW(LPCTSTR, LPCTSTR, PSID, LPDWORD, LPCTSTR, LPDWORD, LPDWORD)
@@ -43,6 +51,7 @@ cdef extern from 'windows.h':
 
     cdef extern int ERROR_INSUFFICIENT_BUFFER
     cdef extern int ERROR_INVALID_SID
+    cdef extern int ERROR_NONE_MAPPED
 
     cdef extern int OWNER_SECURITY_INFORMATION
     cdef extern int GROUP_SECURITY_INFORMATION
@@ -58,6 +67,11 @@ cdef extern from 'windows.h':
     cdef extern int PROTECTED_DACL_SECURITY_INFORMATION
 
     cdef extern int SECURITY_DESCRIPTOR_MIN_LENGTH
+
+    cdef extern int FORMAT_MESSAGE_ALLOCATE_BUFFER
+    cdef extern int FORMAT_MESSAGE_FROM_SYSTEM
+    cdef extern int FORMAT_MESSAGE_IGNORE_INSERTS
+
 
 cdef extern from 'accctrl.h':
     ctypedef enum _SE_OBJECT_TYPE:
@@ -80,6 +94,7 @@ cdef extern from 'accctrl.h':
 
     DWORD GetExplicitEntriesFromAclW(_ACL*, uint32_t*, _EXPLICIT_ACCESS_W**)
 
+
 cdef extern from 'Sddl.h':
     ctypedef int* LPBOOL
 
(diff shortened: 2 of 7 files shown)


borg/constants.py
```diff
@@ -1,7 +1,7 @@
 # this set must be kept complete, otherwise the RobustUnpacker might malfunction:
 ITEM_KEYS = set([b'path', b'source', b'rdev', b'chunks', b'hardlink_master',
                  b'mode', b'user', b'group', b'uid', b'gid', b'mtime', b'atime', b'ctime',
-                 b'xattrs', b'bsdflags', b'acl_nfs4', b'acl_access', b'acl_default', b'acl_extended', ])
+                 b'xattrs', b'bsdflags', b'acl_nfs4', b'acl_access', b'acl_default', b'acl_extended', b'dacl'])
```
ThomasWaldmann: does it make sense to name it win_dacl to make clear it is for windows?

borg/platform_windows.pyx
```diff
@@ -0,0 +1,295 @@
+from __future__ import absolute_import
```
ThomasWaldmann: you don't need that, it is mandatory in python3.

borg/platform_windows.pyx
```diff
@@ -0,0 +1,295 @@
...
+    SetNamedSecurityInfoW(cstrPath, SE_FILE_OBJECT, DACL_SECURITY_INFORMATION, NULL, NULL, newDACL, NULL)
+
+    for i in range(0, len(pyDACL)):
+        if pyDACL[i]['user']['name'] == '' or numeric_owner:
+            LocalFree(<HLOCAL>ACEs[i].Trustee.ptstrName)
+        else:
+            PyMem_Free(ACEs[i].Trustee.ptstrName)
+    free(ACEs)
+    PyMem_Free(cstrPath)
+    LocalFree(<HLOCAL>newDACL)
```
ThomasWaldmann: in general: some prints / TODOs left. maybe use a logger?

doesn't such code already exist somewhere? there is a pywin32 package, maybe other 3rd party packages for windows on pypi - did you check them?

enkore: pywin32 is a pretty thin wrapper (unless I missed some high level abstractions); it could still save some code though.

borg/platform_windows.pyx
enkore: No, it should work ok without the future import.

borg/platform_windows.pyx
```diff
@@ -0,0 +1,295 @@
...
+    ConvertSidToStringSidW(sid, &sidstr)
+    ret = PyUnicode_FromWideChar(sidstr, -1)
+    LocalFree(<HLOCAL>sidstr)
+    return ret
+
+def get_owner(path):
+    cdef int request = OWNER_SECURITY_INFORMATION
+    cdef BYTE* sd = _get_file_security(path, request)
+    if sd == NULL:
+        return 'unknown', 'unknown'
```
enkore: The first is human name, the secont the SID? Then the second should be the null SID instead of "unknown" I think.

borg/constants.py
enkore: Yes, "bsdflags" also has a prefix.

enkore:
(A more general remark)

I think in line with #906, #908 and related issues we should try to use file descriptor / handle based operations in favour of path-based operations to avoid various race conditions and inconsistencies when files are replaced during a backup run.

Nothing -yet- for this PR, since the framework around the code isn't ready for it and I'd like those changes to happen in master first (since all platforms benefit). Windows can then use the _get_osfhandle API to retrieve the object backing a file descriptor of the CRTs POSIX-like API (open etc). And that can be used with the usual object-oriented Windows functions (e.g. GetUserObjectSecurity instead of GetFileSecurity)

enkore:
pywin32 won't come to MSYS2 it seems so let's just go down this route.

Anakonda:
Already tried py2exe and cx_freeze to make executable before making my own script.

Leaving MSYS2 means using MSVC. Borg(_hasindex.c) doesn't compile with MSVC and I really don't want to go there.

borg/platform_windows.pyx
```diff
@@ -0,0 +1,295 @@
...
+
+cdef PSID _get_file_security(filename, int request):
+    cdef DWORD length = 0
+    # N.B. This query may fail with ERROR_INVALID_FUNCTION
+    # for some filesystems.
+    cdef wchar_t* wcharfilename = PyUnicode_AsWideCharString(filename, NULL)
+    if wcharfilename == NULL:
+        print 'filename is NULL'
+    GetFileSecurityW(wcharfilename, request, NULL, 0, &length)
+    if GetLastError() == 122:
```
enkore: Use the constant (ERROR_INSUFFICIENT_BUFFER) instead

borg/platform_windows.pyx
```diff
@@ -0,0 +1,295 @@
...
+    cdef BYTE* sd = _get_file_security(path, request)
+    if sd == NULL:
+        return 'unknown', 'unknown'
+    cdef PSID sid = _get_security_descriptor_owner(sd)
+    if sid == NULL:
+        print 'sid is null'
+        return 'unknown', 'unknown'
+    name, domain, sid_type = _look_up_account_sid(sid)
+    free(sd)
+    if domain != '' and domain.lower() != platform.node().lower() and domain != 'BUILTIN':
```
ThomasWaldmann: if domain and ...

borg/platform_windows.pyx
Anakonda: I get `AttributeError: module 'borg.platform' has no attribute 'node'` without it

enkore: Found it. We need to add `language_level=3` to the cython_compiler() call in setup.py to enable PY3 source level by default.

Anakonda:
Did force push with the changes.
Also removed the selftest as it fails.

ThomasWaldmann:
history editing makes review harder  - maybe keep that to the very end, after review.

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+    cdef wchar_t* name = <wchar_t*>malloc((SIZE) * sizeof(wchar_t))
+    cdef wchar_t* domain = <wchar_t*>malloc((SIZE) * sizeof(wchar_t))
+    cdef DWORD cch_name = SIZE
+    cdef DWORD cch_domain = SIZE
+    cdef uint16_t sid_type = <uint16_t>0
+
+    cdef BOOL ret = LookupAccountSidW(NULL, sid, name, &cch_name, domain, &cch_domain, &sid_type)
+    if ret == 0:
+        lasterror = GetLastError()
+        if lasterror == 1332:
```
enkore: Use the constant instead

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+    # for some filesystems.
+    cdef wchar_t* wcharfilename = PyUnicode_AsWideCharString(filename, NULL)
+    if wcharfilename == NULL:
+        print 'filename is NULL'
+    GetFileSecurityW(wcharfilename, request, NULL, 0, &length)
+    if GetLastError() == ERROR_INSUFFICIENT_BUFFER:
+        SetLastError(0)
+    else:
+        print '_get_file_security failed. Windows error:', GetLastError()
+        return NULL
```
enkore: To make error handling easier you could define a function like this:

```py
def raise_error(failed_api='', path=''):
    cdef wchar_t *error_message
    error = GetLastError()
    if not error:
        return
    FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
                   0, error, 0, <wchar_t*> &error_message, 0, 0)
    error_string = PyUnicode_FromWideChar(error_message, -1)
    error_string = (failed_api + ': ' + error_string).strip()
    if path:
        raise OSError(error, error_string, path)
    else:
        raise OSError(error, error_string)
```

(untested)

enkore:
To make error handling easier you could define a function like this:

```py
def raise_error(failed_api, path=''):
    cdef wchar_t *error_message
    error = GetLastError()
    if not error:
        return
    FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
                   0, error, 0, <wchar_t*> &error_message, 0, 0)
    error_string = PyUnicode_FromWideChar(error_message, -1)
    LocalFree(error_message)
    error_string = (failed_api + ': ' + error_string)
    if path:
        raise OSError(error, error_string, path)
    else:
        raise OSError(error, error_string)
```

(untested)

And then you could use it like

```py
SomeApiCall(filename)
raise_error('SomeApiCall', filename)
```

And get exceptions like `OSError [Errno 1234] SomeApiCall: The requested foo did not bar: 'C:/Windows/xyz'`

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+    cdef uint32_t length
+    cdef _EXPLICIT_ACCESS_W* ACEs
+
+    GetExplicitEntriesFromAclW(DACL, &length, &ACEs)
+
+    pyDACL = []
+    cdef PSID newsid
+    cdef uint32_t domainlength
+    cdef uint32_t sid_type
+    for i in range(0, length):
```
enkore: `range(0, x)` = `range(x)`

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+            continue
+        permissions = {'user': {'name': name, 'sid': sidstr}, 'permissions': (ACEs[i].grfAccessPermissions, ACEs[i].grfAccessMode, ACEs[i].grfInheritance)}
+        pyDACL.append(permissions)
+    item[b'win_dacl'] = safe_encode(json.dumps(pyDACL))
+
+    free(SD)
+    LocalFree(<HLOCAL>ACEs)
+
+def acl_set(path, item, numeric_owner=False):
+    if b'dacl' not in item:
```
enkore: `b'win_dacl'`

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+        return
+
+    pyDACL = json.loads(safe_decode(item[b'win_dacl']))
+    cdef _EXPLICIT_ACCESS_W* ACEs = <_EXPLICIT_ACCESS_W*>calloc(sizeof(_EXPLICIT_ACCESS_W), len(pyDACL))
+    cdef wchar_t* temp
+    cdef PSID newsid
+    for i in range(0, len(pyDACL)):
```
enkore: `range(0, x)` = `range(x)`

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+            ACEs[i].Trustee.ptstrName = PyUnicode_AsWideCharString(pyDACL[i]['user']['name'], NULL)
+        ACEs[i].grfAccessPermissions = pyDACL[i]['permissions'][0]
+        ACEs[i].grfAccessMode = pyDACL[i]['permissions'][1]
+        ACEs[i].grfInheritance = pyDACL[i]['permissions'][2]
+    cdef _ACL* newDACL
+    SetEntriesInAclW(len(pyDACL), ACEs, NULL, &newDACL)
+    cdef wchar_t* cstrPath = PyUnicode_AsWideCharString(path, NULL)
+    SetNamedSecurityInfoW(cstrPath, SE_FILE_OBJECT, DACL_SECURITY_INFORMATION, NULL, NULL, newDACL, NULL)
+
+    for i in range(0, len(pyDACL)):
```
enkore: `range(0, x)` = `range(x)`

enkore:
Some nitpicks in the code, and the error handling, otherwise it looks good from the code. Have to try it :)

borg/platform_windows.pyx
```diff
@@ -0,0 +1,291 @@
...
+    DWORD SetEntriesInAclW(unsigned int, _EXPLICIT_ACCESS_W*, PACL, _ACL**)
+
+cdef PSID _get_file_security(filename, int request):
+    cdef DWORD length = 0
+    # N.B. This query may fail with ERROR_INVALID_FUNCTION
+    # for some filesystems.
+    cdef wchar_t* wcharfilename = PyUnicode_AsWideCharString(filename, NULL)
+    if wcharfilename == NULL:
+        print 'filename is NULL'
+    GetFileSecurityW(wcharfilename, request, NULL, 0, &length)
```
enkore: if filename is NULL it should return? Since PyUnicode_AsWideCharString will already raise MemoryError no need to do extra error handling; as soon as control returns to the Python interpreter (after leaving the cdef) the exception will be propagated.

borg/platform_windows.pyx
enkore: Missing a `PyyMem_Free(wcharfilename)` here

ThomasWaldmann:
btw, after review of this is finished and it is merged into windows branch, I guess it would make sense to maybe rebase the windows branch upon then-current master - before we start doing global changes in master (see the open stuff for 1.1 milestone).

borg/platform_windows.pyx
```diff
@@ -190,6 +225,7 @@ def set_owner(path, owner, sidstring = None):
         domainlength = 0
         LookupAccountNameW(NULL, temp, newsid, &length, NULL, &domainlength, &sid_type)
         if GetLastError() != 0:
+            raise_error('LookupAccountNameW', owner)
             PyMem_Free(temp)
             return
```
ThomasWaldmann: it doesn't look like it is ever coming back from raise_error. so, first free the memory?

PR #1041, opened 2016-05-15
Repository: compact v2

enkore:
- Track free space information for each ~~un-compact~~ fragmented segment
- Don't compact large segments with little free space
- Don't read deleted data when compacting

Consider this a WIP, it has not been tested since the last rebase. I will also probably split this into two+ commits for clarity. (Splitting it from #985 and moving it around as a single commit was simply easier)

TODO: 
- [x] Split into multiple atomic commits
- [x] retest 
- [x] Better terminiology 
- [x] ~~better variable names ("self.compact", "self.segment")~~ later/OT
- [x] ~~Possibly add hole-punching compaction (opt-in; also, what are the atomicity guarantees of fallocate? It absolutely cannot corrupt other parts of the segment file in no case!)~~ -- later, if at all in 1.x

Terms:

- Compact segment: contiguous, no holes / deleted data
- Fragmented|Sparse segment: has deleted data in it
- Segment compaction: making fragmented segments compact
- ... and how the heck can we give "deleted data" a better name?! (*terminated data* does not count)
  - "superseded entry"?
    - PUT x, PUT x: first PUT superseded by newer PUT
    - PUT x, DELETE x: first PUT is superseded by DELETE
    - (assuming clean commit state)
  - Let's use that

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -32,15 +32,51 @@ FreeSpace = partial(defaultdict, int)
 
 
 class Repository:
-    """Filesystem based transactional key value store
+    """
+    Filesystem based transactional key value store
+
+    Transactionality is achieved by using a log (aka journal) to record changes. The log is a series of numbered files
+    called segments. Each segment is a series of log entries. The segment number together with the offset of each
+    entry relative to its segment start establishes an ordering of the log entries. This is the "definition" of
+    time for the purposes of the log.
+
+    Log entries are either PUT, DELETE or COMMIT.
+
+    A COMMIT is always the final log entry in a segment and marks all data from the beginning of the log until the
+    segment ending with the COMMIT as committed and consistent. The segment number of a segment ending with a COMMIT
+    is called the transaction ID of that commit, and a segment ending with a COMMIT is called committed.
+
+    When reading from a repository it is first checked whether the last segment is committed. If it is not, then
+    all segments after the last committed segment are deleted; they contain log entries whose consistency is not
+    established by a COMMIT.
+
+    Note that the COMMIT can't establish consistency by itself, but only manages to do so with proper support from
+    the platform (including the hardware). See platform_base.SyncFile for details.
+
+    A PUT inserts a key-value pair. The value is stored in the log entry, hence the repository implements
+    full data logging, meaning that all data is consistent, not just metadata (which is common in file systems).
+
+    A DELETE marks a key as deleted.
+
+    For a given key only the last entry regarding the key, which is called current (all other entries are called
+    superseded), is relevant: If there is no entry or the last entry is a DELETE then the key does not exist.
+    Otherwise the last PUT defines the value of the key.
+
+    By superseding a PUT (with either another PUT or a DELETE) the log entry becomes obsolete. A segment containing
+    such obsolete entries is called sparse, while a segment containing no such entries is called compact.
+
+    Sparse segments can be compacted and thereby disk space freed. This destroys the transaction for which the
+    superseded entries where current.
 
     On disk layout:
+
     dir/README
     dir/config
     dir/data/<X // SEGMENTS_PER_DIR>/<X>
     dir/index.X
     dir/hints.X
     """
+
     class DoesNotExist(Error):
         """Repository {} does not exist."""
 
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -9,6 +9,8 @@ logger = logging.getLogger(__name__)
 import os
 import shutil
 import struct
+from collections import defaultdict
+from functools import partial
 from zlib import crc32
 
 import msgpack
@@ -25,6 +27,8 @@ MAGIC_LEN = len(MAGIC)
 TAG_PUT = 0
 TAG_DELETE = 1
 TAG_COMMIT = 2
+
+FreeSpace = partial(defaultdict, int)
 
 
 class Repository:
@@ -210,21 +214,30 @@ class Repository:
             self.index = self.open_index(transaction_id)
         if transaction_id is None:
             self.segments = {}  # XXX bad name: usage_count_of_segment_x = self.segments[x]
-            self.compact = set()  # XXX bad name: segments_needing_compaction = self.compact
+            self.compact = FreeSpace()  # XXX bad name: freeable_space_of_segment_x = self.compact[x]
         else:
             if do_cleanup:
                 self.io.cleanup(transaction_id)
             with open(os.path.join(self.path, 'hints.%d' % transaction_id), 'rb') as fd:
                 hints = msgpack.unpack(fd)
-            if hints[b'version'] != 1:
-                raise ValueError('Unknown hints file version: %d' % hints['version'])
-            self.segments = hints[b'segments']
-            self.compact = set(hints[b'compact'])
+            if hints[b'version'] == 1:
+                logger.debug('Upgrading from v1 hints.%d', transaction_id)
+                self.segments = hints[b'segments']
+                self.compact = FreeSpace()
+                for segment in sorted(hints[b'compact']):
+                    logger.debug('Rebuilding sparse info for segment %d', segment)
+                    self._rebuild_sparse(segment)
+                logger.debug('Upgrade to v2 hints complete')
+            elif hints[b'version'] != 2:
+                raise ValueError('Unknown hints file version: %d' % hints[b'version'])
+            else:
+                self.segments = hints[b'segments']
+                self.compact = FreeSpace(hints[b'compact'])
 
     def write_index(self):
-        hints = {b'version': 1,
+        hints = {b'version': 2,
                  b'segments': self.segments,
-                 b'compact': list(self.compact)}
+                 b'compact': self.compact}
         transaction_id = self.io.get_segments_transaction_id()
         hints_file = os.path.join(self.path, 'hints.%d' % transaction_id)
         with open(hints_file + '.tmp', 'wb') as fd:
@@ -238,10 +251,10 @@ class Repository:
         if self.append_only:
             with open(os.path.join(self.path, 'transactions'), 'a') as log:
                 print('transaction %d, UTC time %s' % (transaction_id, datetime.utcnow().isoformat()), file=log)
-        # Remove old indices
+        # Remove old auxiliary files
         current = '.%d' % transaction_id
         for name in os.listdir(self.path):
-            if not name.startswith('index.') and not name.startswith('hints.'):
+            if not name.startswith(('index.', 'hints.')):
                 continue
             if name.endswith(current):
                 continue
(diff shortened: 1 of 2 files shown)


borg/repository.py
```diff
@@ -342,7 +367,26 @@ def _update_index(self, segment, objects, report=None):
...
+        self.segments[segment] = 0
+        for tag, key, offset, size in self.io.iter_objects(segment):
+            if tag == TAG_PUT:
+                if self.index.get(key, (-1, -1)) != (segment, offset):
+                    self.compact[segment] += size
+                else:
+                    self.segments[segment] += 1
+            elif tag == TAG_DELETE:
+                s, off = self.index.get(key, (-1, -1))
+                if segment <= s or s == -1:
```
ThomasWaldmann: can you explain why you consider the delete entry as freespace if the key is not in the index?

borg/repository.py
```diff
@@ -342,7 +367,26 @@ def _update_index(self, segment, objects, report=None):
...
+                    self.compact[segment] += size
```
ThomasWaldmann: can you explain why you consider the delete entry as freespace if the key is not in the index? and why is the offset not compared for the segment == s case?

borg/repository.py
```diff
@@ -342,7 +367,26 @@ def _update_index(self, segment, objects, report=None):
...
+                else:
```
ThomasWaldmann: else-branch is executed when negation of if-clause is true:
not (segment <= s or s == -1)
==
segment > s and s != -1

maybe easier to understand if the if/else is turned around?

borg/repository.py
```diff
@@ -653,19 +703,21 @@ def recover_segment(self, segment, filename):
...
         header = fd.read(self.put_header_fmt.size)
-        size, tag, key, data = self._read(fd, self.put_header_fmt, header, segment, offset, (TAG_PUT, ))
+        size, tag, key, data = self._read(fd, self.put_header_fmt, header, segment, offset, (TAG_PUT, ), read_data)
         if id != key:
             raise IntegrityError('Invalid segment entry header, is not for wanted id [segment {}, offset {}]'.format(
                 segment, offset))
-        return data
+        if read_data:
+            return data
+        return size
```
ThomasWaldmann: return data if read_data else size ?

borg/repository.py
```diff
@@ -683,18 +735,32 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags):
...
+            if key is None and tag in (TAG_PUT, TAG_DELETE):
+                key = fd.read(32)
+                length -= 32
+                if len(key) != 32:
+                    raise IntegrityError('Segment entry key short read [segment {}, offset {}]: expected {}, got {} bytes'.format(
+                        segment, offset, 32, len(key)))
+            oldpos = fd.tell()
+            read = fd.seek(length, os.SEEK_CUR) - oldpos
+            if read != length:
+                raise IntegrityError('Segment entry data short read [segment {}, offset {}]: expected {}, got {} bytes'.format(
```
ThomasWaldmann: "short seek"?

borg/repository.py
```diff
@@ -713,7 +779,7 @@ def write_delete(self, id, raise_full=False):
         crc = self.crc_fmt.pack(crc32(id, crc32(header)) & 0xffffffff)
         fd.write(b''.join((crc, header, id)))
         self.offset += self.put_header_fmt.size
-        return self.segment
+        return self.segment, self.put_header_fmt.size
```
ThomasWaldmann: hmm, consistency?

write_put returns segment, offset,
but write_delete returns segment, size?

borg/repository.py
```diff
@@ -683,18 +735,32 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags):
...
+            data = fd.read(length)
+            if len(data) != length:
+                raise IntegrityError('Segment entry data short read [segment {}, offset {}]: expected {}, got {} bytes'.format(
+                    segment, offset, length, len(data)))
+            if crc32(data, crc32(memoryview(header)[4:])) & 0xffffffff != crc:
+                raise IntegrityError('Segment entry checksum mismatch [segment {}, offset {}]'.format(
+                    segment, offset))
+            if key is None and tag in (TAG_PUT, TAG_DELETE):
+                key, data = data[:32], data[32:]
+        else:
```
ThomasWaldmann: in this case, the code does not check the crc32 - so we can't be sure length or key is correct.

enkore: Yes. Meaning that we need to consider how corrupted data will propagate from here; this is used in two instances:

1. Reading length of to-be-superseded PUT. The key "can't be" wrong (seg+offset came from the index). Corrupt length: only used to consider/"unconsider" segments for compaction. Worst case: Segment is compacted when it isn't "ripe" yet, or it's not even though it would be.

Acceptable.

2. The main loop in compact_segments. If the key of a live entry is corrupted it could be silently skipped.

Not acceptable

D: Still use this for superseding entries, don't use for the main loop in compact_segments. For something like "borg delete" or "borg prune" it still significantly reduces the reads done.

ThomasWaldmann: 2. ?

enkore: No second case, I already removed that for this reason but forget I did :D

ThomasWaldmann: i remember some "hole punching" idea for unused entries.

guess for that a wrong length could punch too much.

borg/repository.py
enkore: If the tag_delete is for a key that already doesn't exist then it counts towards free space because it carries no information at all

borg/repository.py
enkore: Yes, the offset needs to be greater than the current offset to adhere to the time series of operations.

borg/repository.py
enkore: write_put returns (segment, offset)

write_delete used to return segment, now (segment, length)

write_commit returns None

Not really consistent either way.

borg/repository.py
enkore: I added a note in LoggedIO.read(read_data=False) that it should not be regarded reliable.

enkore:
The original commit is now two commits. The first one has no changes, the second one adds read_data and already addresses some of this reviews' notes (docs, if-else formatting). The third commit is a new change.

borg/repository.py
```diff
@@ -32,15 +32,53 @@
 
 
 class Repository:
-    """Filesystem based transactional key value store
+    """
+    Filesystem based transactional key value store
+
+    Transactionality is achieved by using a log (aka journal) to record changes. The log is a series of numbered files
+    called segments. Each segment is a series of log entries. The segment number together with the offset of each
+    entry relative to it's segment start establishes an ordering of the log entries. This is the "definition" of
```
ThomasWaldmann: its

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+    time for the purposes of the log.
+
+    Log entries are either PUT, DELETE or COMMIT.
+
+    A COMMIT is always the final log entry in a segment and marks all data from the beginning of the log until the
+    segment ending with the COMMIT as committed and consistent. The segment number of a segment ending with a COMMIT
+    is called the transaction ID of that commit, and a segment ending with a COMMIT is called committed.
+
+    When reading from a repository it is first checked whether the last segment is committed. If it is not, then
+    all segments up to the last committed segment are deleted; they contain log entries whose consistency is not
```
ThomasWaldmann: ... then all segments after the last committed segment are deleted as they ...

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+    established by a COMMIT.
+
+    Note that the COMMIT can't establish consistency by itself, but only manages to do so with proper support from
+    the platform (including the hardware). See platform_base.SyncFile for details.
+
+    A PUT inserts a key-value pair. The value is stored in the log entry, hence the repository implements
+    full data logging, meaning that all data is consistent, not just metadata (which is common in file systems).
+
+    A PUT that is the last PUT for it's key and is not followed by a DELETE is said to be current,
```
ThomasWaldmann: its

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+    while the opposite state is called superseded.
```
ThomasWaldmann: (not limited to being followed by a DEL, it could be also superseded by another PUT)

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+
+    A DELETE marks a key as deleted.
```
ThomasWaldmann: ... if it is the last entry for that id.

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+
+    Due to the ordering of the log entries older entries are superseded by newer entries. For example, a PUT x=y
+    followed by PUT x=z would imply that the current value of the key "x" is "z" (not "y").
+
```
ThomasWaldmann: maybe above paragraph is enough to describe "superseded" and "current" and the previous description could be deleted.

borg/repository.py
```diff
@@ -32,15 +32,53 @@
...
+    By superseding a PUT (with either another PUT or a DELETE) the log entry becomes obsolete. A segment containing
+    such obsolete entries is called sparse, while a segment containing no such entries is called compact.
+
```
ThomasWaldmann: when is DEL superseded?

borg/repository.py
```diff
@@ -32,15 +32,53 @@
 
 
 class Repository:
-    """Filesystem based transactional key value store
+    """
+    Filesystem based transactional key value store
```
ThomasWaldmann: great docs

borg/repository.py
```diff
@@ -210,21 +214,30 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
         else:
             if do_cleanup:
                 self.io.cleanup(transaction_id)
             with open(os.path.join(self.path, 'hints.%d' % transaction_id), 'rb') as fd:
                 hints = msgpack.unpack(fd)
-            if hints[b'version'] != 1:
-                raise ValueError('Unknown hints file version: %d' % hints['version'])
-            self.segments = hints[b'segments']
-            self.compact = set(hints[b'compact'])
+            if hints[b'version'] == 1:
```
ThomasWaldmann: somehow i was asking myself whether we really want such code in the core code or rather in the upgrader code.

borg/repository.py
```diff
@@ -267,32 +280,39 @@ def complete_xfer():
...
-            if self.io.segment_exists(segment):
-                for tag, key, offset, data in self.io.iter_objects(segment, include_data=True):
-                    if tag == TAG_PUT and self.index.get(key, (-1, -1)) == (segment, offset):
+        for segment, freeable_space in sorted(self.compact.items()):
+            if not self.io.segment_exists(segment):
+                del self.compact[segment]
+                continue
+            segment_size = self.io.segment_size(segment)
+            if segment_size > 0.2 * self.max_segment_size and freeable_space < 0.15 * segment_size:
+                logger.debug('saving segment %d for later (not enough potential, %d bytes)', segment, freeable_space)
```
ThomasWaldmann: "saving" might be misleading (like "writing it somewhere").

"keeping"?
"not compacting it now"?

borg/repository.py
```diff
@@ -733,18 +739,32 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags):
...
+                key, data = data[:32], data[32:]
+        else:
+            if key is None and tag in (TAG_PUT, TAG_DELETE):
+                key = fd.read(32)
+                length -= 32
+                if len(key) != 32:
+                    raise IntegrityError('Segment entry key short read [segment {}, offset {}]: expected {}, got {} bytes'.format(
+                        segment, offset, 32, len(key)))
+            oldpos = fd.tell()
+            read = fd.seek(length, os.SEEK_CUR) - oldpos
```
ThomasWaldmann: seeked =

borg/repository.py
```diff
@@ -376,17 +377,14 @@ def _rebuild_usage(self, segment):
         for tag, key, offset, size in self.io.iter_objects(segment):
             if tag == TAG_PUT:
                 if self.index.get(key, (-1, -1)) != (segment, offset):
+                    # This PUT is superseded in a later segment
```
ThomasWaldmann: it could be also superseded at a later offset in this segment, but that would not make a difference for freeable space in this segment.

borg/repository.py
```diff
@@ -376,17 +377,14 @@ def _rebuild_usage(self, segment):
...
             elif tag == TAG_DELETE:
-                s, off = self.index.get(key, (-1, -1))
-                if segment <= s or s == -1:
-                    self.compact[segment] += size
-                else:
-                    self.segments[s] -= 1
-                    size = self.io.read(s, offset, key, read_data=False)
-                    self.compact[s] += size
+                # The outcome of the DELETE has been recorded in the PUT branch already
+                self.compact[segment] += size
```
ThomasWaldmann: hmm, need to look at that later again...

enkore:
_rebuild_usage could be simplified (only build self.compact, not self.segment), because the old self.segment is still valid.

enkore:
(Some stuff still remaining, most prominently, a proper unit test)

PR #1047, opened 2016-05-16
init: move epilogue from docs/usage into help output

enkore:
Also tell which algos are used and that HW accel is supported.

online manual (`borg xyz --help` and `borg help xxx`) has some other spots. Ha, just collect'em here and get that out for 1.1 as well.

- [x] Important epilogue that only was in the www docs -> moved into `borg init --help`
- [ ] --help should say who encrypts and do so clearly somewhere
- [ ] Env vars should have a section in "borg help" [can just move them and include from docs/usage]
- [ ] Should mention "borg help" sections on "borg --help"
- [ ] Move "item flags" to "borg help" [and include from docs/usage]?

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1167,7 +1167,48 @@ class Archiver:
         init_epilog = textwrap.dedent("""
         This command initializes an empty repository. A repository is a filesystem
         directory containing the deduplicated data from zero or more archives.
-        Encryption can be enabled at repository init time.
+
+        Encryption can be enabled at repository init time (the default).
+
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Maybe do less "collection PRs" that bundle stuff together that doesn't require to be together.

You can just concentrate on a single item, solve it, get feedback, merge, done.

Then start work on the next item in the next PR.

ThomasWaldmann:
Change looks fine.

enkore:
I'll do the other stuff in other PRs :)

PR #1049, opened 2016-05-17
Handle SIGTERM

enkore:
kill <pid> is now a clean exit instead of a crashit.

Also, don't print a big and scary traceback for KeyboardInterrupts.

SIGTERM exists everywhere, even on Windows.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1488,6 +1488,14 @@ def sig_info_handler(signum, stack):  # pragma: no cover
             break
 
 
+class SIGTERMReceived(BaseException):
+    pass
+
+
+def sig_term_handler(signum, stack):
+    raise SIGTERMReceived
+
+
 def setup_signal_handlers():  # pragma: no cover
     sigs = []
     if hasattr(signal, 'SIGUSR1'):
@@ -1496,6 +1504,7 @@ def setup_signal_handlers():  # pragma: no cover
         sigs.append(signal.SIGINFO)  # kill -INFO pid (or ctrl-t)
     for sig in sigs:
         signal.signal(sig, sig_info_handler)
+    signal.signal(signal.SIGTERM, sig_term_handler)
 
 
 def main():  # pragma: no cover
@@ -1522,6 +1531,9 @@ def main():  # pragma: no cover
         exit_code = EXIT_ERROR
     except KeyboardInterrupt:
         msg = 'Keyboard interrupt.\n%s\n%s' % (traceback.format_exc(), sysinfo())
+        exit_code = EXIT_ERROR
+    except SIGTERMReceived:
+        msg = 'Received SIGTERM.'
         exit_code = EXIT_ERROR
     if msg:
         logger.error(msg)


borg/archiver.py
```diff
@@ -1488,6 +1488,10 @@ def sig_info_handler(signum, stack):  # pragma: no cover
             break
 
 
+def sig_term_handler(signum, stack):
+    raise KeyboardInterrupt()
```
ThomasWaldmann: dirty. maybe raise a separate exception?

borg/archiver.py
```diff
@@ -1521,7 +1526,7 @@ def main():  # pragma: no cover
         msg = 'Local Exception.\n%s\n%s' % (traceback.format_exc(), sysinfo())
         exit_code = EXIT_ERROR
     except KeyboardInterrupt:
-        msg = 'Keyboard interrupt.\n%s\n%s' % (traceback.format_exc(), sysinfo())
+        msg = 'Keyboard interrupt.'
```
ThomasWaldmann: we are losing information here, at least in some cases.

there are 2 cases when users press Ctrl-C:
- they decided otherwise and just want to interrupt borg
- borg is somehow hanging and they want to terminate it

In the second case, the traceback is interesting to see where it is hanging.

borg/archiver.py
enkore: Other than the name "KeyboardInterrupt" being not entirely accurate I don't see why this would be dirty - it's the same thing Python does for SIGINT raising KeyboardInterrupt (just in C and some more complexity to defer stuff correctly to the Python VM etc)

borg/archiver.py
enkore: Ok, that's a good reason to keep it for ^C. And a good reason to use another exception for SIGTERM :)

borg/archiver.py
ThomasWaldmann: I think we should be able to make a difference between SIGTERM exception and SIGINT exception.

borg/archiver.py
ThomasWaldmann: also, `killall borg` now would say "keyboard interrupt", which is not true.

enkore:
> Merging #1049 into 1.0-maint will decrease coverage by 1.92%

Sure.

ThomasWaldmann:
LOL

borg/archiver.py
```diff
@@ -1523,6 +1532,9 @@ def main():  # pragma: no cover
     except KeyboardInterrupt:
         msg = 'Keyboard interrupt.\n%s\n%s' % (traceback.format_exc(), sysinfo())
         exit_code = EXIT_ERROR
+    except SIGTERMReceived:
+        msg = 'Received SIGTERM.\n'
```
ThomasWaldmann: \n is not needed, rather leave it away.

ThomasWaldmann:
So, the user-visible difference is that sigkill did not clean up locks before but now does?

enkore:
Yes.

I only noticed this after shutting down while a backup was running. systemd (or probably any other service manager) sends SIGTERM first and then after some time a SIGKILL.

ThomasWaldmann:
ehrm, i meant sigterm, not sigkill.
so the previous behaviour on sigterm was to just terminate, without triggering cleanup in context managers? or did it do nothing until a sigkill was received?

enkore:
(I read that as "sigterm" not "sigkill")

SIGKILL can't be trapped in user mode, it will always take down the process hard.

SIGTERM previously used the default Python handler which terminated the interpreter. No Python code ran, and no context manager clean-up happened.

PR #1052, opened 2016-05-18
Repository: fix commit tags being seen in data

enkore:
I have a slightly different fix for master that doesn't destroy performance with larger segments (i.e. put commits in their own segment)

Problem description: is_committed_segment() seeked to the end of a segment file and checked whether there is a COMMIT_TAG there. This can also be the case when a PUT ends with a structure that is a COMMIT_TAG and is thus no valid way to check for the commit state of a segment.

This is a luck-based bug. You'll need to back something up where a chunk ends with a commit tag (another borg repo) and then that has to end up at the end of a segment file and *then* you'll need a kill/crash and *then* you *can theoretically* get an inconsistent repo (with some other conditions applying as well).

Now:

- Early segment magic check by iterator setup in iter_objects
- Sanity check: Do we have a COMMIT_TAG at the end? (<-- this is also the fast-bail-out for practically all uncomitted segments)
- If no: can't be committed
- If yes: iterate objects, check if that's an actual commit tag

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -536,7 +536,7 @@ class LoggedIO:
         """Verify that the transaction id is consistent with the index transaction id
         """
         for segment, filename in self.segment_iterator(reverse=True):
-            if self.is_committed_segment(filename):
+            if self.is_committed_segment(segment):
                 return segment
         return None
 
@@ -550,10 +550,14 @@ class LoggedIO:
             else:
                 break
 
-    def is_committed_segment(self, filename):
+    def is_committed_segment(self, segment):
         """Check if segment ends with a COMMIT_TAG tag
         """
-        with open(filename, 'rb') as fd:
+        try:
+            iterator = self.iter_objects(segment)
+        except IntegrityError:
+            return False
+        with open(self.segment_filename(segment), 'rb') as fd:
             try:
                 fd.seek(-self.header_fmt.size, os.SEEK_END)
             except OSError as e:
@@ -561,7 +565,22 @@ class LoggedIO:
                 if e.errno == errno.EINVAL:
                     return False
                 raise e
-            return fd.read(self.header_fmt.size) == self.COMMIT
+            if fd.read(self.header_fmt.size) != self.COMMIT:
+                return False
+        seen_commit = False
+        while True:
+            try:
+                tag, key, offset = next(iterator)
+            except IntegrityError:
+                return False
+            except StopIteration:
+                break
+            if tag == TAG_COMMIT:
+                seen_commit = True
+                continue
+            if seen_commit:
+                return False
+        return seen_commit
 
     def segment_filename(self, segment):
         return os.path.join(self.path, 'data', str(segment // self.segments_per_dir), str(segment))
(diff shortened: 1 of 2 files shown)


borg/repository.py
```diff
@@ -550,18 +550,37 @@ def cleanup(self, transaction_id):
...
+                tag, key, offset = next(iterator)
+            except IntegrityError:
+                return False
+            except StopIteration:
+                break
+            if tag == TAG_COMMIT:
+                seen_commit = True
+                continue
+            if seen_commit:
+                raise IntegrityError('Commit not located at end of segment [segment {}, offset {}]'.format(segment, offset))
```
ThomasWaldmann: the old code would just return False in this case, not raise.

borg/repository.py
```diff
@@ -582,6 +601,8 @@ def get_write_fd(self, no_new=False, raise_full=False):
         return self._write_fd
 
     def get_fd(self, segment):
+        if segment == self.segment and self._write_fd:
+            self._write_fd.flush()
```
ThomasWaldmann: this seems a unrelated fix.

borg/repository.py
```diff
@@ -550,18 +550,37 @@ def cleanup(self, transaction_id):
...
+        with open(self.segment_filename(segment), 'rb') as fd:
             try:
                 fd.seek(-self.header_fmt.size, os.SEEK_END)
             except OSError as e:
                 # return False if segment file is empty or too small
                 if e.errno == errno.EINVAL:
                     return False
                 raise e
-            return fd.read(self.header_fmt.size) == self.COMMIT
+            if fd.read(self.header_fmt.size) != self.COMMIT:
```
enkore: I wonder why write_commit isn't using this handy constant

borg/testsuite/repository.py
```diff
@@ -268,7 +275,7 @@ def list_objects(self):
         return set(int(key) for key in self.repository.list())
 
     def test_repair_corrupted_segment(self):
-        self.add_objects([[1, 2, 3], [4, 5, 6]])
+        self.add_objects([[1, 2, 3], [4, 5], [6]])
```
ThomasWaldmann: why did you change this?

ThomasWaldmann:
besides unclear test change, code looks good to me.

enkore:
IIRC there is a check later that accessing an object in the last segment raises IntegrityError. This doesn't happen anymore with this code, because the last segment won't be considered committed anymore due to the IntegrityError in it.

ThomasWaldmann:
ok, merge or is something left to do?

enkore:
I will make another PR about this for master after we back-merge 1.0-maint to fix a performance nag (not a bug) this code has for large segments. That involves much more changes to the tests hence nothing for 1.0-maint (which wouldn't be used with large segments).

enkore:
This was also reported a couple years back in attic. At that time it didn't support uncompressed storage, so it wasn't an issue: https://github.com/jborg/attic/issues/80

ThomasWaldmann:
@enkore maybe rather add this at another place than a closed pull request?

PR #1053, opened 2016-05-18
add swidth call, fixes #1051

### LINKED ISSUE #1051: cjk screen output formatting trouble
chinese, japanese, korean "monospace" terminal fonts use 2 cells per glyph, so using len(text) gives an incorrect value if text has cjk unicode code points.

this leads to wrong output in the `--progress` "single line output" - the amount of spaces to "fill" the line and erase previous output is computed wrongly.

there is https://pypi.python.org/pypi/wcwidth to compute the correct width.

3 ideas:
- just require it via setup.py
- bundle it in borg.support
- use a try / except ImportError and use a len(text) dummy if it can't be imported.

Screenshot: borg_jap

ThomasWaldmann:
on posix platforms, this is a small wrapper around wcswidth(3).
on non-supported platforms, it just uses len(s), as before.

### DIFF
--- a/borg/platform_linux.pyx
+++ b/borg/platform_linux.pyx
@@ -5,15 +5,11 @@ import stat
 
 from .helpers import posix_acl_use_stored_uid_gid, user2uid, group2gid, safe_decode, safe_encode
 from .platform_base import SyncFile as BaseSyncFile
+from .platform_posix import swidth
+
 from libc cimport errno
 
 API_VERSION = 3
-
-cdef extern from "wchar.h":
-    cdef int wcswidth(const Py_UNICODE *str, size_t n)
-
-def swidth(s):
-    return wcswidth(s, len(s))
 
 cdef extern from "sys/types.h":
     int ACL_TYPE_ACCESS
--- a/borg/platform_linux.pyx
+++ b/borg/platform_linux.pyx
@@ -9,6 +9,12 @@ from libc cimport errno
 
 API_VERSION = 3
 
+cdef extern from "wchar.h":
+    cdef int wcswidth(const Py_UNICODE *str, size_t n)
+
+def swidth(s):
+    return wcswidth(s, len(s))
+
 cdef extern from "sys/types.h":
     int ACL_TYPE_ACCESS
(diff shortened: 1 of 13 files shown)


enkore:
on windows the CRT doesn't seem to have that function, but I'm not sure whether the MSYS2 Python actually uses the CRT or a glibc port. If it's the CRT then we'd have to resort to Uniscribe I think, if one wanted to implement this for Windows as well.

ThomasWaldmann:
How about adding a platform_posix and moving stuff there that is present on all posix platforms?
Then, in platform_{linux,freebsd,darwin}: from platform_posix import ...

enkore:
Yeah

ThomasWaldmann:
@enkore I'ld like to backport this to 1.0-maint so it gets into 1.0.3. But I think without the platform_posix / platform_base part - so I'ld just copy it to the places where needed.

What do you think?

enkore:
Yes. Just need to be careful when back-merging it ;)

ThomasWaldmann:
Circular import at helpers.py line 35 (from .platform import swidth)

enkore:
Move the Statistics class to archive.py?

ThomasWaldmann:
Yup, doing that right now.

ThomasWaldmann:
Consider the amount of change to fix a cosmetic but, 1.0-maint just got less advisable i think.

ThomasWaldmann:
merge as is?

borg/platform_linux.pyx
```diff
@@ -5,6 +5,8 @@ import stat
 
 from .helpers import posix_acl_use_stored_uid_gid, user2uid, group2gid, safe_decode, safe_encode
 from .platform_base import SyncFile as BaseSyncFile
+from .platform_posix import swidth
+
```
enkore: Do these in platform.py instead? Considering Windows would also only add ~two lines there.

ThomasWaldmann: guess over time it'll become more. also, it avoids the temptation to just duplicate the code. :)

enkore: fair enough

PR #1058, opened 2016-05-18
When probing key files, do binary reads

### LINKED ISSUE #1056: crash related to (non-borg) binary files in keys directory
If i try to list/mount a remote repository i get this error:

`Local Exception.`
`Traceback (most recent call last):`
`  File "borg/archiver.py", line 1498, in main`
`  File "borg/archiver.py", line 1452, in run`
`  File "borg/archiver.py", line 74, in wrapper`
`  File "borg/helpers.py", line 95, in load`
`  File "borg/key.py", line 53, in key_factory`
`  File "borg/key.py", line 289, in detect`
`  File "borg/key.py", line 402, in find_key`
`  File "/home/vagrant/.pyenv/versions/3.5.1/lib/python3.5/codecs.py", line 321, in decode`
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9b in position 12: invalid start byte`

`Platform: Linux xxx 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08) x86_64 `
`Linux: debian 8.4 `
`Borg: 1.0.2  Python: CPython 3.5.1`

I'm using the standalone binary version 1.0.2 on debian 8.4.0 (server and client) with locales configured as en_US.UTF-8.
The generated key file seems to be correctly encoded:
`file -i` outputs "us-ascii" which according to http://stackoverflow.com/a/11303410 is correct.

Has anybody a clue what could be the issue here?

enkore:
Fixes #1056

### DIFF
--- a/borg/key.py
+++ b/borg/key.py
@@ -394,13 +394,15 @@ class KeyfileKey(KeyfileKeyBase):
     FILE_ID = 'BORG_KEY'
 
     def find_key(self):
-        id = hexlify(self.repository.id).decode('ascii')
+        file_id = self.FILE_ID.encode()
+        first_line = file_id + b' ' + hexlify(self.repository.id)
         keys_dir = get_keys_dir()
         for name in os.listdir(keys_dir):
             filename = os.path.join(keys_dir, name)
-            with open(filename, 'r') as fd:
-                line = fd.readline().strip()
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Maybe add this comment:
```
                # we do the magic / id check in binary mode to avoid stumbling over
                # decoding errors if somebody has binary crap in the keys dir.
```

PR #1062, opened 2016-05-19
update CHANGES

ThomasWaldmann:


enkore:
Use the `issue` role to automatically link from RTFD to the github issue in question?

ThomasWaldmann:
I expect that a lot of people will look at the not-rendered plain text file, so I'ld rather not put "strange looking" markup in there.

PR #1075, opened 2016-05-21
Repository: avoid reading large segments for commit state check

enkore:
Forces commit tags into their own micro-segment, which can be checked
quickly by is_committed_segment

Follow-up to #1052 for master.

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -735,8 +735,8 @@ class LoggedIO:
         return self.segment
 
     def write_commit(self):
-        fd = self.get_write_fd(no_new=True)
-        fd.sync()
+        self.close_segment()
+        fd = self.get_write_fd()
         header = self.header_no_crc_fmt.pack(self.header_fmt.size, TAG_COMMIT)
         crc = self.crc_fmt.pack(crc32(header) & 0xffffffff)
         fd.write(b''.join((crc, header)))
--- a/borg/testsuite/repository.py
+++ b/borg/testsuite/repository.py
@@ -216,18 +216,18 @@ class RepositoryAppendOnlyTestCase(RepositoryTestCaseBase):
         self.repository.commit()
 
         self.repository.append_only = False
-        assert segments_in_repository() == 1
+        assert segments_in_repository() == 2
         self.repository.put(b'00000000000000000000000000000000', b'foo')
         self.repository.commit()
         # normal: compact squashes the data together, only one segment
-        assert segments_in_repository() == 1
+        assert segments_in_repository() == 4
 
         self.repository.append_only = True
-        assert segments_in_repository() == 1
+        assert segments_in_repository() == 4
         self.repository.put(b'00000000000000000000000000000000', b'foo')
         self.repository.commit()
         # append only: does not compact, only new segments written
-        assert segments_in_repository() == 2
+        assert segments_in_repository() == 6
 
 
 class RepositoryCheckTestCase(RepositoryTestCaseBase):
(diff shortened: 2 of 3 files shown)


ThomasWaldmann:
tests look rather broken.

enkore:
I knew I forgot something...

borg/testsuite/repository.py
```diff
@@ -216,18 +216,18 @@ def segments_in_repository():
         self.repository.commit()
 
         self.repository.append_only = False
-        assert segments_in_repository() == 1
+        assert segments_in_repository() == 2
         self.repository.put(b'00000000000000000000000000000000', b'foo')
         self.repository.commit()
         # normal: compact squashes the data together, only one segment
-        assert segments_in_repository() == 1
+        assert segments_in_repository() == 4
```
enkore: Hm, that's strange on second look

ThomasWaldmann: why not 2? the change in this changeset seems to only add 1 extra segment for the commit.

enkore: - data of the first transaction, 1 segment
- commit, 1 segment
- [compact doesn't do anything here]
- data superseding older data (the put direct above this assert), 1 segment
- commit, 1 segment
- now a total of four segments
- compact copies live data from segment no. 1 (none) to a new segment (so nothing here). Then writes a commit (1 segment), deletes segment no. 1
- total of four segments

ThomasWaldmann: so the final state is like this?:

segment file: content
---------------------------
(0: deleted)
1: commit
2: put 0 foo
3: commit
4: commit

that looks strange.

enkore: The log was the same before, just distributed differently (all in one segment). In this instance it looks really wasteful, but for transactions containing data it's not a big difference. E.g. `<couple hundred MB> <commit> <couple hundred MB> <commit> <some data copied from sparse segments> <commit>`

ThomasWaldmann: sure it is same as before? I'ld have expected that it removed the old commit with the old segment data when they were in same file, that got deleted in compaction process.

enkore: Yes, the first commit would have been deleted together with the segment. Otherwise it's the same.

ThomasWaldmann: ok, so over time, will the repo accumulate more and more tiny commit segments?

enkore: With this change, yes. Before: depends. With the typical write structure borg does (write manifest, commit) and the old compact algorithm commit segments would practically always be compacted. With the new compact algorithm it would depend on whether there is enough other superseded data in the segment containing a commit (a couple-kilobytes manifest doesn't do the trick).

Plan: merge this as is; amend the compact algorithm after the other compact PR to consider commits as well, so these little files get deleted as well.

ThomasWaldmann: ok, sounds good. open a ticket to not forget?

borg/repository.py
```diff
@@ -735,8 +735,8 @@ def write_delete(self, id, raise_full=False):
         return self.segment
 
     def write_commit(self):
-        fd = self.get_write_fd(no_new=True)
-        fd.sync()
```
ThomasWaldmann: hmm, why did you kill the sync? guess we'ld even need 2x sync now.

enkore: close_segment syncs

ThomasWaldmann: there is just a close, no sync.

enkore: https://github.com/borgbackup/borg/blob/master/borg/platform_base.py#L88 ;)

PR #1110, opened 2016-05-30
windows: Add setup wizard

enkore:
$ python3 deployment/windows/buildwin32.py
...
$

Fire ISS up, build installer. Result is in Output directory.
Should be around ~4 MB [which is actually the smallest binary we have so far]

At current HEAD for example: [Borg Backup 1.1 Setup.zip](https://github.com/borgbackup/borg/files/290028/Borg.Backup.1.1.Setup.zip) (wrapped in a ZIP)

Not sure why it shows a full diff for buildwin32.py, I just renamed it. Line endings or other black magic maybe.


ThomasWaldmann:
instead of collecting all python files, dlls, python lib, having own "bootloader" ... - just use pyinstaller?

enkore:
pyinstaller doesn't work with msys2 because it requires pywin32 at lots of places and that needs MSVC-Python which doesn't work with MSYS-Python.

https://github.com/borgbackup/borg/pull/1037#issuecomment-218579911

ThomasWaldmann:
Note: this PR was merged into the windows branch, not (yet) into master.

PR #1113, opened 2016-05-31
platform_base should use platform-provided implementations when needed

enkore:
The composition issue came up on windows. See https://github.com/borgbackup/borg/blob/windows/src/borg/platform_windows.pyx#L339

Making it a package shouldn't be too much of a change-- uh, unless stuff starts falling apart :D 

Things to look at: ... why does this stuff work on Py35 but not Py34? Was the import mechanism changed? [guess I need to take a peek there tomorrow with the big iron] And what's up with the flake8 toxenv?


src/borg/platform/base.py
```diff
@@ -1,5 +1,19 @@
 import os
 
+from .. import platform
```
ThomasWaldmann: so init imports base, base imports init? looks like a cyclic import.

enkore: It imports just the package name not *from* the pacakge, so it should work imho

enkore: If it really isn't supposed to work then I'd make a lazy intermediary that imports once-on-demand.

ThomasWaldmann: isn't package/__init__.py run when you import package?

ThomasWaldmann:
merge?

PR #1123, opened 2016-06-03
vagrant: fix path to borg package, add hidden import

ThomasWaldmann:


enkore:
Does the hidden-import issue apply to other platforms as well (e.g. linux, freebsd)?

ThomasWaldmann:
@enkore i ran into it on osx, but giving that option only adds that module to the distribution unconditionally (because it is not auto-discovered), so there is no harm if it is auto-discovered.

ThomasWaldmann:
I'm merging it. In case it creates issues on other platforms (which i do not believe), I'll notice it. :)

PR #1129, opened 2016-06-03
FUSE versions view

ThomasWaldmann:
for mixed repos, --prefix would make sense here.
also, not mounting all archives, but just last N would also make sense.
we'll do that after the first/last PR is merged.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -6,12 +6,12 @@ import tempfile
 import time
 from collections import defaultdict
 from distutils.version import LooseVersion
+from zlib import adler32
 
 import llfuse
 import msgpack
 
 from .logger import create_logger
-from .lrucache import LRUCache
 logger = create_logger()
 
 from .archive import Archive
@@ -51,14 +51,18 @@ class ItemCache:
 class FuseOperations(llfuse.Operations):
     """Export archive as a fuse filesystem
     """
-
+    # mount options
     allow_damaged_files = False
+    versions = False
 
     def __init__(self, key, repository, manifest, archive, cached_repo):
         super().__init__()
-        self._inode_count = 0
-        self.key = key
+        self.repository_uncached = repository
         self.repository = cached_repo
+        self.archive = archive
+        self.manifest = manifest
+        self.key = key
+        self._inode_count = 0
         self.items = {}
         self.parent = {}
         self.contents = defaultdict(dict)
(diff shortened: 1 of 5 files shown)


src/borg/fuse.py
```diff
@@ -98,27 +106,48 @@ def process_archive(self, archive, prefix=[]):
...
+                    ident = adler32(chunkid, ident)
+                name = b'%s.%08x' % (name, ident)
+            return name
+
+        if versions_view:
+            parent = self.process_inner(name, parent)
+            name = version_name(name, item)
+        self.process_real_leaf(name, item, parent, prefix)
+
+    def process_real_leaf(self, name, item, parent, prefix):
```
ThomasWaldmann: this is just the old code moved into a method.

src/borg/fuse.py
```diff
@@ -98,27 +106,48 @@ def process_archive(self, archive, prefix=[]):
...
+            item = self.cache.get(inode)
+            item.nlink = item.get('nlink', 1) + 1
+            self.items[inode] = item
+        else:
+            inode = self.cache.add(item)
+        self.parent[inode] = parent
+        if name:
+            self.contents[parent][name] = inode
+
+    def process_inner(self, name, parent):
```
ThomasWaldmann: this is just the old code moved into a method.

src/borg/fuse.py
```diff
@@ -98,27 +106,48 @@ def process_archive(self, archive, prefix=[]):
...
+            if 'chunks' in item:
+                ident = 0
+                for chunkid, _, _ in item.chunks:
+                    ident = adler32(chunkid, ident)
+                name = b'%s.%08x' % (name, ident)
+            return name
+
+        if versions_view:
+            parent = self.process_inner(name, parent)
+            name = version_name(name, item)
```
ThomasWaldmann: these few lines are the whole magic. :)

ThomasWaldmann:
rebased on current master

ThomasWaldmann:
Looks now like this (showing 2 versions of fuse.py coming from 2 archives):
```
$ borg mount repo mnt
$ ls -lR mnt/src/borg/fuse.py
mnt/src/borg/fuse.py:
total 23
-rw-rw-r-- 1 tw tw 12026 Aug 16 02:02 fuse.py.00a40f83
-rw-rw-r-- 1 tw tw 12022 Aug 16 01:48 fuse.py.e3010da6
```
You see there is a directory in place of where usually would be a file.
In that directory, there are the versions of that file.
The filenames are repeated, so it is easier to copy out a file of there and still know its name.
Files have a suffix, computed via adler32(file's chunkids).

ThomasWaldmann:
how about merging this mostly "as is" and declaring it an experimental feature that is subject to change? so we could actually use it and get more feedback from practice.

enkore:
I think the only thing missing to do it would be making `versions_view` a mount option.

ThomasWaldmann:
collapse / merge?

enkore:
lgtm. changelog, collapse, merge.

PR #1131, opened 2016-06-04
llfuse: update version requirement for freebsd

ThomasWaldmann:
as the requirement string is a AND operation, I used the rough "<2.0" requirement
and then excluded the non-working versions.


ThomasWaldmann:
@jungle-boogie I'm merging this into 1.0-maint now, can you test it on freebsd?

PR #1134, opened 2016-06-04
SyncFile: open with O_EXCL (or equivalent)

enkore:
This doesn't avoid "any and all issues", but can limit the damage.

A "backport" (the SyncFile abstraction is new, cherry-picking does not apply) to 1.0-maint makes sense.

Works with sshfs. Linux manpage says it shouldn't work with NFSv3.


ThomasWaldmann:
looks good. and good that we require 3.4+. :)

PR #1149, opened 2016-06-09
item_keys in manifest, validate item keys

### LINKED ISSUE #1147: store valid metadata keys into repository manifest
Some places in borg need the valid metadata keys set of the repository:
- add_item (to validate stuff before it gets stored, used by "borg create")
- RobustUnpacker (to notice when it gets invalid stuff from corrupt repo, used by "borg check")

having a fixed set in the code (see ITEM_KEYS) is slightly problematic, because a future borg could have more keys - and if you'ld want to run "borg check" on such a repo with an older borg version, valid keys could be considered invalid (because not-yet known in ITEM_KEYS of the older borg).

if the repository manifest does not have the valid keys set, it must fallback to a builtin set.

it is essential that borg check can safely and confidently decide whether some metadata dictionary it got from a (potentially) corrupted repo is valid or invalid, because it bases the decision about whether to skip it (== drop the file) or not on that. See also PR #1133.

ThomasWaldmann:
we need a list of valid item metadata keys. using a list stored in the repo manifest
is more future-proof than the hardcoded ITEM_KEYS in the source code.

keys that are in union(item_keys_from_repo, item_keys_from_source) are considered valid.

the 2nd changeset uses the valid item keys set (and a new required keys set) to validate item dicts within borg check.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -591,6 +591,9 @@ ITEM_KEYS = frozenset([b'path', b'source', b'rdev', b'chunks',
                        b'mode', b'user', b'group', b'uid', b'gid', b'mtime', b'atime', b'ctime',
                        b'xattrs', b'bsdflags', b'acl_nfs4', b'acl_access', b'acl_default', b'acl_extended', ])
 
+# this is the set of keys that are always present in items:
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
merge?

ThomasWaldmann:
merge?

PR #1150, opened 2016-06-10
Fix test_sparse_file for solaris

enkore:


src/borg/testsuite/archiver.py
```diff
@@ -408,8 +408,9 @@ def _set_repository_id(self, path, id):
             return repository.id
 
     def test_sparse_file(self):
-        # no sparse file support on Mac OS X
-        sparse_support = sys.platform != 'darwin'
+        # Mac OS X has no sparse file support
+        # Solaris (ZFS) has sparse file support, but is less predictable about it
+        sparse_support = sys.platform not in ['darwin', 'sunos5']
```
ThomasWaldmann: strictly taken, we do assumptions about the fs via checking the OS here.
that might work in some cases, but feels wrong.
esp. on linux, there might be also zfs (or whatever other fs that does not behave like ext4).

enkore: Yes, it's technically not only about the OS but also about the FS (and possibly configuration options of it).

ThomasWaldmann:
yes, merge. good for now, more improvements can be done later.

Issue #1138, opened 2016-06-05
confused I/O Error reporting / handling


ThomasWaldmann:
I am moving this comment by @nesneros which was originally in #1130 to a separate ticket:

Error message seen in log output:
```
/home/jan/projects/spark/core/target/scala-2.11/test-classes/org/apache/spark/rdd/AsyncRDDActionsSuite$$anonfun$1.class: [Errno 5] Input/output error
```

I was thinking more about these I/O errors and disconnection to the target drive. I would think one could experience I/O errors on the target if it suddenly is disconnected. Could the I/O errors be a "reporting bug" in borg? That the I/O error is  related to a file on the target, but is reported for the file being processed on the source? 

I archiver.py around line 328 there is code like:
```
                try:
                    status = archive.process_file(path, st, cache, self.ignore_inode)
                except OSError as e:
                    status = 'E'
                    self.print_warning('%s: %s', path, e)
```
If the process_file function is writing to target I think this could explain the behaviour. My Python skills are basically non-existing so I might be completely off track.

***
As seen in #1130 just ignoring the OSError when writing to the repo and continuing with next file leads to a corrupt, but completed transaction committed to the repo.

If the error would be handled as a fatal error, the transaction would be incomplete and (I guess) easily rolled back.

The bounty is for fixing the wrong log output and the wrong exception handling leading to corruption.

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/34878887-confused-i-o-error-reporting-handling)

ThomasWaldmann:
Reply by @encore:

Yes, the reporting isn't very good there. If you get an EIO when writing to the repository it may seem like the EIO came from reading a backup file.

I'm not sure but it could be possible that the file system you use, smbnetfs, gives EIO if there are connection problems.

ThomasWaldmann:
Note: until fixed, such a log entry has to be interpreted as:

"while processing the file at the left, the error shown at the right occurred." and "processing" does not just include reading that file, but also a lot of other operations, like writing the resulting chunks to the repo.

enkore:
So...

- Wrap try/except OSError around operations on (and only on) the backup source set, remove try/except OSError for process_file etc.
- Or use InternalOSError everywhere else? Like in 1cc644

ThomasWaldmann:
A conservative fix for 1.0-maint would be good.

enkore:
> InternalOSError

[...]

Issue #1116, opened 2016-05-31
BrokenPipeError leading to a crash when using filtering/counting shell utilities


fritschy:
When using e.g. tail or head to limit the number of lines a BrokenPipeError is thrown and leads to an abnormal exit of borg:

```
% sudo borg list BACKUP | head -n1
20160202-121810                      Tue, 2016-02-02 13:03:51
Local Exception.
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1498, in main
    exit_code = archiver.run(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1452, in run
    return args.func(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 80, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 544, in do_list
    print(format_archive(archive_info))
BrokenPipeError: [Errno 32] Broken pipe

Platform: Linux occam 4.6.0-bait.0-wbt #1 SMP PREEMPT Tue May 17 09:32:06 CEST 2016 x86_64 
Linux: debian stretch/sid 
Borg: 1.0.2  Python: CPython 3.5.1+
PID: 21174  CWD: /home/user
sys.argv: ['/usr/bin/borg', 'list', 'BACKUP']
SSH_ORIGINAL_COMMAND: None

Exception ignored in: <_io.TextIOWrapper name='<stdout>' encoding='UTF-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/34723987-brokenpipeerror-leading-to-a-crash-when-using-filtering-counting-shell-utilities)

ThomasWaldmann:
We already had a similar case once and the issue that came up while fixing it was how to differentiate harmless BrokenPipeErrors (like head / tail / more / less) from not harmless ones.

ThomasWaldmann:
Solving this might need to wrap screen output closely by an exception handler:

```
import sys

for i in range(100000):
    try:
        print(i)
    except BrokenPipeError:
        #might need to close stderr to avoid a warning emitted by python:
        #sys.stderr.close()
        sys.exit(0)
```

enkore:
> Solving this might need to wrap screen output closely by an exception handler:

If so, then it should be done after console output was factored out from application logic.

[...]

PR #1151, opened 2016-06-10
Handle BrokenPipeError for list, info and diff commands

andela-kerinoso:
Considering conversations on #1116, #805 and #790, BrokenPipeError message is not desirable in certain cases. As far as I know, cases involving list, info and diff commands seem to be the most important ones when their outputs are piped to filter/count utilities. Thus, this commit takes care of those cases only.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2168,6 +2168,13 @@ def main():  # pragma: no cover
     except RemoteRepository.RPCError as e:
         msg = '%s\n%s' % (str(e), sysinfo())
         exit_code = EXIT_ERROR
+    except BrokenPipeError:
+        if args.func.__name__ in ['do_info', 'do_list', 'do_diff']:
+            sys.stderr.close()
+            sys.exit(0)
+        else:
+            msg = '\n%s\n%s' % (traceback.format_exc(), sysinfo())
+            exit_code = EXIT_ERROR
     except Exception:
         msg = 'Local Exception.\n%s\n%s' % (traceback.format_exc(), sysinfo())
         exit_code = EXIT_ERROR


src/borg/archiver.py
```diff
@@ -2168,6 +2168,13 @@ def main():  # pragma: no cover
     except RemoteRepository.RPCError as e:
         msg = '%s\n%s' % (str(e), sysinfo())
         exit_code = EXIT_ERROR
+    except BrokenPipeError:
+        if args.func.__name__ in ['do_info', 'do_list', 'do_diff']:
```
ThomasWaldmann: ah, interesting approach. so the idea is that for all known "read only" commands, we ignore the BrokenPipeError?

did you try to find out which pipe was broken?

src/borg/archiver.py
```diff
@@ -2168,6 +2168,13 @@ def main():  # pragma: no cover
...
+            sys.stderr.close()
+            sys.exit(0)
```
ThomasWaldmann: why do you exit here and not just set exit_code = EXIT_... and continue?

enkore: if show_rc is True then the code below tries to print to the closed sys.stderr

ThomasWaldmann: well, a flag close_stderr could be set and later stderr be closed (while keeping the --show-rc functionality).

ThomasWaldmann:
The best approach (if doable) could be like:
- find out whether the broken pipe is related to the ssh subprocess for remote repo or not
- if yes, report the broken pipe, exit with rc=2
- if no, suppress the broken pipe exception, exit with rc=0

src/borg/archiver.py
enkore: Wouldn't the BrokenPipeError indicate that further writes to the FD in question would raise?

ThomasWaldmann: If it is the same fd, yes, right. But we need to differentiate cases, see below.

src/borg/archiver.py
andela-kerinoso: Yes, BrokenPipeError is ignored for known read-only commands by this approach.

For the second question, did you mean for which command was the pipe broken? If so, it was only for the info command; it worked fine for both list and diff. I intentionally included the two that worked just because of the comments in previous and current issues/PR related to this PR.

andela-kerinoso:
@ThomasWaldmann: But we would still want to ignore `BrokenPipeError` for remote repo which happens to use ssh for the read-only commands, right? Going by that approach, traceback for `InternalOSError` would still show up for the read-only commands in the case of remote repo.

enkore:
> enclose the place(s) where we deal with the ssh subprocess for remote repo rather closely with an exception handler for BrokenPipeError and reraise it as a InternalOSError. This error is fatal and should exit with rc=2.

That's a good point, but a somewhat different issue. BrokenPipeError is a subclass of OSError, if we resolve #1138 then this is also solved.

ThomasWaldmann:
OK, maybe we better first fix #1138 before fixing this one.

andela-kerinoso:
Alright! I will close this.

enkore:
I think this PR is valid and the correct approach that is just blocked by #1138. Leave open, fix #1138, [rebase if necessary], merge?

andela-kerinoso:
OK. I have started looking into #1138.

ThomasWaldmann:
#1138 is fixed now in 1.0-maint branch.

ThomasWaldmann:
@enkore in #1138 the approach to ignore harmless stuff (having issues with backup input files) is to enclose the ops with harmless failures with the special wrappers you wrote.

So, for BrokenPipeError, the harmless stuff we want to ignore are the print() and logging calls. So they would have to get wrapped and the toplevel error handler would just catch/ignore the HarmlessBrokenPipeError exception. If we then have a non-harmless BrokenPipeError (like when talking to the ssh process or anything else we did not wrap), it would still give a traceback and rc=2.

enkore:
I think since we have central logging setup we could just inject a wrapper doing that there. print() is a bit of a different story but didn't we want to go a full-logging solution anyway?

ThomasWaldmann:
there could be just a print_out() and print_err() wrapper (sys.stdout vs. sys.stderr).

enkore:
Yes, easier/faster to do for now.

ThomasWaldmann:
@andela-kerinoso would you mind if I fix this?

the deadline for 1.0.4 is rather close and we need to get it out soon, due to some critical bugs.

there are lots of other issues (for you, if you like) to fix, we won't run out of them anytime soon. :)

andela-kerinoso:
@ThomasWaldmann sure!

PR #1154, opened 2016-06-11
Adds uuid to archive format tags

qua-bla:
Create time is already stored as meta data. Hence, it often makes more sense to use random names that are potentially globally unique. Other names could lead to confusion if different backup systems exist in parallel etc.

### DIFF
--- a/docs/usage/create.rst.inc
+++ b/docs/usage/create.rst.inc
@@ -79,7 +79,7 @@ The archive name needs to be unique. It must not end in '.checkpoint' or
 checkpoints and treated in special ways.
 
 In the archive name, you may use the following format tags:
-{now}, {utcnow}, {fqdn}, {hostname}, {user}, {pid}
+{now}, {utcnow}, {fqdn}, {hostname}, {user}, {pid}, {uuid}
 
 To speed up pulling backups over sshfs and similar network file systems which do
 not provide correct inode information the --ignore-inode flag can be used. This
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
How about adding all uuid1/3/4/5?

qua-bla:
Thanks for the feedback!

Naming the existing field `uuid4` would be more precise and transparent. Actually, I think I would prefer that. Other types of uuids could be added later.

uuid1: Based on hosts MAC and a very accurate time. However, I don't know of any benefits over uuid4. In general it is problematic since it exposes MAC and precise creation time. People might not be aware of the fact that they are sharing this information, if they share an archives name.

uuid4: Completely random.

uuid3/5: Just an algorithm to hash given values. Nothing random here. Probably you don't want that ;)

ThomasWaldmann:
Right, so only 1 and 4 somehow make sense.

I am not yet very much convinced about the uuid usecase here. why not just servername-datetime as archive name?

qua-bla:
Probably boils down to something like matter of taste. It's just strange to me to label something with time, if time is also available as metadata. Also time contains spaces and colons which is less handy for sysadmins.

(for me uuids are the only sane standard for ids in databases, filesystems etc.)

ThomasWaldmann:
The code change looks fine, I am just asking myself whether it is useful. Merge?

Issue #1157, opened 2016-06-12
make source more pretty


ThomasWaldmann:
like the filesystem object item, which recently got item.Item as a comfortable wrapper to get rid of the item[b'foo'] stuff and rather use item.foo, more such wrappers could be introduced for same reason:

- manifest.archives entries
- Archive metadata dicts
- encryption key dict

ThomasWaldmann:
See #1158 and #1161.

PR #1158, opened 2016-06-12
add Key/EncryptedKey PropDict, see #1157

ThomasWaldmann:


ThomasWaldmann:
merge?

enkore:
Hmm, shouldn't the two PropDict classes be in borg.key, since they aren't and shouldn't be used globally?

ThomasWaldmann:
@enkore yes, I'll move stuff around when I am finished with all the PropDicts. The docstring of them need more work, so I want to have them at same place now.

enkore:
ok

PR #1160, opened 2016-06-12
Fix 1159

ThomasWaldmann:
these changes for archive metadata dicts are somehow almost the same as what has been already done for the item metadata dicts.

note: we are really lucky that neither the item metadata nor the archive metadata dict had grown beyond 15 items.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -595,8 +595,8 @@ ITEM_KEYS = frozenset([b'path', b'source', b'rdev', b'chunks',
 REQUIRED_ITEM_KEYS = frozenset([b'path', b'mtime', ])
 
 
-def valid_msgpacked_item(d, item_keys_serialized):
-    """check if the data <d> looks like a msgpacked item metadata dict"""
+def valid_msgpacked_dict(d, keys_serialized):
+    """check if the data <d> looks like a msgpacked dict"""
     d_len = len(d)
     if d_len == 0:
         return False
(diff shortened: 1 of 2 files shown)


borg/archive.py
```diff
@@ -731,18 +737,24 @@ def rebuild_manifest(self):
 
         Iterates through all objects in the repository looking for archive metadata blocks.
         """
+        def valid_archive(obj):
+            if not isinstance(obj, dict):
+                return False
+            keys = set(obj)
+            return REQUIRED_ARCHIVE_KEYS.issubset(keys)
```
ThomasWaldmann: I left away the "and keys.issubset(ARCHIVE_KEYS)" because this is used in the case of a lost manifest, so we can't rely on getting the valid archive item keys from the manifest.

If it would be added, we could only use ARCHIVE_KEYS from the source and any additional key in the archive dict would make it invalid and it would get skipped.

borg/archive.py
```diff
@@ -731,18 +737,24 @@ def rebuild_manifest(self):
...
+
         logger.info('Rebuilding missing manifest, this might take some time...')
         # as we have lost the manifest, we do not know any more what valid item keys we had.
         # collecting any key we encounter in a damaged repo seems unwise, thus we just use
         # the hardcoded list from the source code. thus, it is not recommended to rebuild a
         # lost manifest on a older borg version than the most recent one that was ever used
         # within this repository (assuming that newer borg versions support more item keys).
         manifest = Manifest(self.key, self.repository)
+        archive_keys_serialized = [msgpack.packb(name) for name in ARCHIVE_KEYS]
```
ThomasWaldmann: note: only keys known to this version of the source.

ThomasWaldmann:
merge?

PR #1161, opened 2016-06-12
make source more pretty

ThomasWaldmann:


ThomasWaldmann:
btw, I'ld like to finish and merge this, but it should be done at a time with not many parallel open PRs to avoid conflicts.

src/borg/archive.py
```diff
@@ -1215,52 +1216,53 @@ def valid_item(obj):
...
-            archive_items = [item for item in self.manifest.archives.items() if item[0] == archive]
-            if not archive_items:
+            info = self.manifest.archives.get(archive)
+            if info is None:
                 logger.error("Archive '%s' not found.", archive)
+                archive_infos = []
+            else:
+                archive_infos = [info]
             num_archives = 1
             end = 1
```
ThomasWaldmann: these "= 1" might be wrong if the archive is not found, need to check this.

ThomasWaldmann:
merge? anything else to clean up that way?

src/borg/archive.py
enkore: the whole section looks like it could use a minor refactoring.

enkore:
should be it

PR #1173, opened 2016-06-15
fix sparse file test, fixes #1170

### LINKED ISSUE #1170: fix the sparse test
as seen in #1150 this test has issues on zfs. and also on xfs as seen in some borg 1.0.3 SuSE build log.

as all these filesystems also exist on linux (or freebsd), disabling by OS is not good enough.

ThomasWaldmann:
found out that xfs is doing stuff behind the scenes: it is pre-allocating 16MB
to prevent fragmentation (in my case, value depends on misc factors).

fixed the test so it just checks that the extracted sparse file uses less (not
necessary much less) space than a non-sparse file would use.

another problem showed up when i tried to verify the holes in the sparse file
via SEEK_HOLE, SEEK_DATA:
after the few bytes of real data in the file, there was another 16MB
preallocated space.
So I ended up checking just the hole at the start of the file.

tested on: ext4, xfs, zfs, btrfs

### DIFF
--- a/borg/testsuite/archiver.py
+++ b/borg/testsuite/archiver.py
@@ -412,8 +412,21 @@ class ArchiverTestCase(ArchiverTestCaseBase):
             self.assert_equal(fd.read(hole_size), b'\0' * hole_size)
         st = os.stat(filename)
         self.assert_equal(st.st_size, total_len)
-        if sparse_support and hasattr(st, 'st_blocks'):
-            self.assert_true(st.st_blocks * 512 < total_len / 9)  # is output sparse?
+        if sparse_support:
+            if hasattr(st, 'st_blocks'):
+                # do only check if it is less, do NOT check if it is much less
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
merge?

enkore:
Tested this on Solaris (because of #1150), found out that /tmp on Solaris is by default tmpfs-y and that doesn't support sparse files. If you move TMPDIR to an actual ZFS partition on Solaris then the test works (and is rather slow).

ThomasWaldmann:
On Linux, http://linux.die.net/man/2/statfs could be used to skip the test if the fs does not support sparse files.
On IRC it was noted that this might be totally broken on OS X. No idea about other OSes.

ThomasWaldmann:
If nobody objects, I'll merge this as is soon.

The statfs change can be done later, after more research / discussion.

Issue #1166, opened 2016-06-14
Trailing and leading whitespace and control characters are allowed as archive names


Abogical:
It is possible for borg archive names to have trailing and leading whitespace and control characters.

```bash
$ borg create "/path/to/repo::archive       " /files/to/backup
$ borg create "/path/to/repo::   archive" /files/to/backup
$ borg create "/path/to/repo::new
> line" /files/to/backup
$ borg create /path/to/repo::$'\1' /files/to/backup
$ borg list /path/to/repo
archive                              Tue, 2016-06-14 02:10:48
   archive                           Tue, 2016-06-14 02:11:20
new
line                             Tue, 2016-06-14 02:11:23
                                    Tue, 2016-06-14 02:12:30
```

Obviously, this creates several issues, escpecially trailing whitespaces.

Trailing whitespace
--
Due to `borg list`'s format **It is impossible to know whether an archive name has trailing spaces, let alone how many spaces it has**.

For example let's create an archive with a normal name.
```bash
$ borg create /path/to/repo::archive /files/to/backup
```
Then create an archive with 1 space after the name, and another with 3 spaces after the name.
```bash
$ borg create "/path/to/repo::archive " /files/to/backup
$ borg create "/path/to/repo::archive   " /files/to/backup
```
Listing the repository gives an awkward result.
```bash
$ borg list /path/to/repo
archive                              Tue, 2016-06-14 03:20:01
archive                              Tue, 2016-06-14 03:24:06
archive                              Tue, 2016-06-14 03:28:13
```
The following problems occur:
- All archives appeares to have the same name, which is impossible.
- It is impossible to know that there are trailing whitespace characters.
- It is impossible to know how many trailing spaces there are.

Control characters and leading whitespaces
--
Control characters and leading whitespaces may make `borg list` harder to read and harder to be parsed by scripts.

Possible solutions
--
- Reject control characters and leading and trailing whitespaces when creating a new archive.
- Truncate leading and trailing whitespace when creating a new archive.
- Change format of `borg list`, or allow to change it via `--list-format`.

*Note: times in the shell's output are not actual times.*

enkore:
> - It is impossible to know that there are trailing whitespace characters.
> - It is impossible to know how many trailing spaces there are.

borg-info gives the verbatim name with no added formatting, so you'd be able to get it from there.

Abogical:
@enkore `borg info` is for a specific archive, The problem comes when listing all archive names in a repository, which is done via `borg list`

[...]

PR #1174, opened 2016-06-15
Reject weird archive names, for #1166

Abogical:
This prevents the creation of new archives to have this 'crap'. However this doesn't do anything for old archives which have those weird names.

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -728,16 +728,16 @@ class Location:
     # the archive names. Thus, we must not accept "/" in archive names.
     ssh_re = re.compile(r'(?P<proto>ssh)://(?:(?P<user>[^@]+)@)?'
                         r'(?P<host>[^:/#]+)(?::(?P<port>\d+))?'
-                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/]+))?$')
+                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s]))?$')
     file_re = re.compile(r'(?P<proto>file)://'
-                         r'(?P<path>[^:]+)(?:::(?P<archive>[^/]+))?$')
+                         r'(?P<path>[^:]+)(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s]))?$')
     scp_re = re.compile(r'((?:(?P<user>[^@]+)@)?(?P<host>[^:/]+):)?'
-                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/]+))?$')
+                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s]))?$')
     # get the repo from BORG_RE env and the optional archive from param.
     # if the syntax requires giving REPOSITORY (see "borg mount"),
     # use "::" to let it use the env var.
     # if REPOSITORY argument is optional, it'll automatically use the env.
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
I have not yet made up my mind about whether I want this change, but I'll provide some feedback about the implementation nevertheless.

src/borg/helpers.py
```diff
@@ -728,16 +728,16 @@ class Location:
...
+                         r'(?P<path>[^:]+)(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s]))?$')
     scp_re = re.compile(r'((?:(?P<user>[^@]+)@)?(?P<host>[^:/]+):)?'
-                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/]+))?$')
+                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s]))?$')
     # get the repo from BORG_RE env and the optional archive from param.
     # if the syntax requires giving REPOSITORY (see "borg mount"),
     # use "::" to let it use the env var.
     # if REPOSITORY argument is optional, it'll automatically use the env.
-    env_re = re.compile(r'(?:::(?P<archive>[^/]+)?)?$')
+    env_re = re.compile(r'(?:::(?P<archive>[^/\s]([^/\s]| )*[^/\s])?)?$')
```
ThomasWaldmann: This is not just related to you change, but more a general thing:
this code / these regexes is somehow way to complex already (and it's not getting better by making the regexes even more complex).
So, I'ld rather like to simplify / deduplicate it somehow.

src/borg/helpers.py
```diff
@@ -773,6 +773,14 @@ def parse(self, text):
...
         self.archive = m.group('archive')
+        return self._is_archive_name_printable()
+
+    # lookout for non-printable characters
+    def _is_archive_name_printable(self):
+        if self.archive:
+            for char in self.archive:
+                if unicodedata.category(char)[0] == 'C':
+                    return False
         return True
```
ThomasWaldmann: this shouldn't be a method, but rather a global utility function.

implementation could looks somehow like:

def printable(s):
    return all(unicodedata.category(c)[0] == 'C' for c in s)

wondering if such a function does not already exist? in stdlib?

src/borg/testsuite/helpers.py
```diff
@@ -104,6 +104,21 @@ def test_no_slashes(self, monkeypatch):
         with pytest.raises(ValueError):
             Location('/some/path/to/repo::archive_name_with/slashes/is_invalid')
 
+    def test_only_printable_characters(self, monkeypatch):
```
ThomasWaldmann: the utility function then could be tested on its own, disconnected from the archivename context.

src/borg/helpers.py
Abogical: I've looked it up, such a function, dissapointedly, doesn't exist. The closest functions would be string.printable and curses.ascii.isprint(), but they only only on ASCII, not unicode.

Abogical: Ok I may be wrong, I think the function would be string.isprintable()

ThomasWaldmann:
we had a short discussion on irc about this and feel that we rather won't take this.

the downside of making the regexes even more complex (and that there is always the option to just not feed crappy archive names into borg) somehow outweighs the good sides of this PR.

also, your other "formatting archive names" PR somehow can solve any such issues in another way.

thanks for engaging anyway. :)

Abogical:
Your welcome :)

PR #1176, opened 2016-06-16
Fix stdlib DLL importing when borg.exe is run from arbitrary cwd()s

enkore:


enkore:
Build failure stems from flake8 2.6.0 introducing F405 which catches on our star imports.

PR #1180, opened 2016-06-16
Fix confused I/O Error reporting/handling, fixes #1138

### LINKED ISSUE #1138: confused I/O Error reporting / handling

andela-kerinoso:
For the other case whereby the processed file is probably not accessible, `e.errno != errno.EIO`. For instance, I manually created a test-case by moving some of the content of the backup source while archive creation was running. For that, `[Errno 2] No such file or directory` equivalent to `errno.ENOENT` occurred. Thus, the warning is still intact.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from hashlib import sha256
 from operator import attrgetter
 import argparse
+import errno
 import functools
 import inspect
 import io
@@ -214,9 +215,10 @@ class Archiver:
                         continue
                 else:
                     restrict_dev = None
-                self._process(archive, cache, matcher, args.exclude_caches, args.exclude_if_present,
+                if self._process(archive, cache, matcher, args.exclude_caches, args.exclude_if_present,
                               args.keep_tag_files, skip_inodes, path, restrict_dev,
-                              read_special=args.read_special, dry_run=dry_run)
+                              read_special=args.read_special, dry_run=dry_run):
+                              return
             if not dry_run:
                 archive.save(timestamp=args.timestamp)
                 if args.progress:
@@ -274,6 +276,8 @@ class Archiver:
                 try:
                     status = archive.process_file(path, st, cache, self.ignore_inode)
                 except OSError as e:
+                    if e.errno == errno.EIO:
+                        return e.errno
                     status = 'E'
                     self.print_warning('%s: %s', path, e)
         elif stat.S_ISDIR(st.st_mode):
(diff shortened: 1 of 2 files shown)


enkore:
As I've already said on [c898979](https://github.com/andela-kerinoso/borg/commit/c89897906aa46e77d75f9c0fd9ad410d03eab8a3#commitcomment-17903316) (also: [here](https://github.com/borgbackup/borg/issues/1138#issuecomment-223809780) and  [there](https://github.com/borgbackup/borg/issues/1138#issuecomment-225430144)) this isn't a correct fix for the issue at hand.

For a correct fix one needs to introduce a OSError wrapper exception (much like InternalOSError in master branch; just that it should be exactly inverse, see the links above) and check basically all syscalls done when an file/directory is processed, wrap those syscalls in a try/except OSError and re-raise the occured exception with the wrapper. This would then allow OSErrors that occur in repository / cache handling to propagate correctly and terminate the program (=fatal error).

ThomasWaldmann:
I agree with @enkore. Additionally, in many cases exceptions are the better way to handle error conditions than using C-style return values.

ThomasWaldmann:
Please run "fakeroot -u tox" locally before commiting.
This also runs the flake8 checker and avoids issues like seen here:
https://travis-ci.org/borgbackup/borg/jobs/138233914

andela-kerinoso:
Thanks for the observation; I have taken the approach of re-raising the exception linked to backup repository/cache. However, I would have to close this PR and raise a new one because this branch was created out of **master** rather than **1.0-maint**; their file/directory layouts are slightly different.

ThomasWaldmann:
Superceded by #1191.

PR #1183, opened 2016-06-18
create subcommand: add new --cap-user option

c4rlo:
The new `--cap-user` option is available when running borg as root (Linux only).

It causes borg to switch to the given user, while retaining the `CAP_DAC_READ_SEARCH` capability, which enables it to read all files. See the `capabilities(7)` manpage on Linux for more information on how this works.

This behaviour is convenient when backup up things like `/etc`, without wanting to run borg with full root privileges.

I've been running a similar setup at home for months now; the difference is that I'm still on attic, which obviously doesn't have this features built in, so I'm currently using a home-grown C++ wrapper that achieves the equivalent.

Note that this introduces a new dependency (on Linux only), namely on [libcap](https://sites.google.com/site/fullycapable/). If this is deemed unacceptable, I could perhaps look into dynamically detecting whether it's installed, thereby making it an optional dependency.

One more thing to note is that this functionality can only be tested by running as root &ndash; fakeroot is not enough. I've therefore added a test which is skipped unless running as non-fake root. I've not taken any steps to automatically run this in the `Vagrantfile` or `.travis/run.sh`, as I wanted to get some feedback first, and I'm also not sufficiently familiar with your Travis setup.

I'm also open to any better ideas for the name of this option; `--cap-user` is admittedly not super-clear.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -170,7 +170,7 @@ class build_usage(Command):
         print('generating usage docs')
         # allows us to build docs without the C modules fully loaded during help generation
         from borg.archiver import Archiver
-        parser = Archiver(prog='borg').parser
+        parser = Archiver(prog='borg', include_all_platform_args=True).parser
         choices = {}
         for action in parser._actions:
             if action.choices is not None:
@@ -310,7 +310,7 @@ if not on_rtd:
         ext_modules.append(Extension('borg.platform.posix', [platform_posix_source]))
 
     if sys.platform == 'linux':
-        ext_modules.append(Extension('borg.platform.linux', [platform_linux_source], libraries=['acl']))
+        ext_modules.append(Extension('borg.platform.linux', [platform_linux_source], libraries=['acl', 'cap']))
     elif sys.platform.startswith('freebsd'):
         ext_modules.append(Extension('borg.platform.freebsd', [platform_freebsd_source]))
     elif sys.platform == 'darwin':
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -28,7 +28,7 @@ from .cache import Cache
 from .constants import *  # NOQA
 from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
 from .helpers import Error, NoManifestError
-from .helpers import location_validator, archivename_validator, ChunkerParams, CompressionSpec
+from .helpers import location_validator, archivename_validator, validate_capuser, ChunkerParams, CompressionSpec
 from .helpers import ItemFormatter, format_time, format_file_size, format_archive
 from .helpers import safe_encode, remove_surrogates, bin_to_hex
 from .helpers import prune_within, prune_split
@@ -41,7 +41,7 @@ from .helpers import log_multi
 from .helpers import parse_pattern, PatternMatcher, PathPrefixPattern
 from .item import Item
 from .key import key_creator, RepoKey, PassphraseKey
-from .platform import get_flags
+from .platform import get_flags, set_capuser, switch_to_user
 from .remote import RepositoryServer, RemoteRepository, cache_if_remote
 from .repository import Repository
 from .selftest import selftest
@@ -102,10 +102,10 @@ def with_archive(method):
 
 class Archiver:
 
-    def __init__(self, lock_wait=None, prog=None):
+    def __init__(self, lock_wait=None, prog=None, include_all_platform_args=False):
         self.exit_code = EXIT_SUCCESS
         self.lock_wait = lock_wait
-        self.parser = self.build_parser(prog)
+        self.parser = self.build_parser(prog, include_all_platform_args)
 
     def print_error(self, msg, *args):
         msg = args and msg % args or msg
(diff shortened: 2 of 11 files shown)


enkore:
> It causes borg to switch to the given user, while retaining the CAP_DAC_READ_SEARCH capability, which enables it to read all files. See the capabilities(7) manpage on Linux for more information on how this works.

Hmm, that will still often be a trivially exploitable capability to become root. Since the same can be achieved by an external wrapper I don't really see a compelling reason (yet?) to have this in borg itself.

c4rlo:
I agree that in principle this is a separate concern that should ideally be provided by some form of wrapper; something like a capabilities-aware sudo. However, I've never come across such a tool. Assuming this is functionality that others would also find useful I'd therefore think it makes sense to have this directly in borg.

c4rlo:
The reason why I went down the `--cap-user` route instead of creating such a tool, by the way, is that when I first looked at capabilities, there was no way to carry them across exec() calls. However, taking another look now, it appears things have improved: there's now a concept of "ambient capabilities" &ndash; the `capabilities(7)` manpage now has:

```
Ambient (since Linux 4.3):
       This is a set of  capabilities  that  are  preserved  across  an
       execve(2)  of  a  program  that  is not privileged.  The ambient
       capability set obeys the invariant that no capability  can  ever
       be ambient if it is not both permitted and inheritable.

       The  ambient  capability  set  can  be  directly  modified using
       prctl(2).  Ambient capabilities  are  automatically  lowered  if
       either  of  the corresponding permitted or inheritable capabili‐
       ties is lowered.

       Executing a program that changes UID or GID due to the set-user-
       ID or set-group-ID bits or executing a program that has any file
       capabilities set will clear the ambient set.  Ambient  capabili‐
       ties  are  added to the permitted set and assigned to the effec‐
       tive set when execve(2) is called.
```

Using this, it should indeed be possible to create a "capabilities-sudo" program. It'll still be less trivially easy than having the feature built into borg though :)

(In my C++ wrapper I solve this issue by linking against the Python .so and calling `Py_Main()`, instead of just exec()ing `/usr/bin/attic`.)

c4rlo:
Actually, there is a wrapper tool one can use: `capsh`. Last time I looked at it, it didn't appear sufficiently capable, but now that I'm having another look, it looks like it does the job:

```
~/tmp $ ls -l foo
-rw-r--r-- 1 root root 4 Jun 18 14:12 foo
~/tmp $ sudo capsh --keep=1 --user=backup --caps='= cap_dac_read_search+eip' -- -c 'echo x >foo'
/bin/bash: foo: Permission denied
~/tmp $ sudo capsh --keep=1 --user=backup --caps='= cap_dac_read_search+eip' -- -c 'cat foo'
foo
```

c4rlo:
Actually, there is a wrapper tool one can use: `capsh`. Last time I looked at it, it didn't appear sufficiently capable, but now that I'm having another look, it looks like the latest version from git does the job:

```
~/tmp $ ls -l foo
-rw------- 1 root root 4 Jun 18 14:12 foo
~/tmp $ sudo ~/src/libcap/progs/capsh --keep=1 --user=backup --inh=cap_dac_read_search --addamb=cap_dac_read_search --drop=all --caps='= cap_dac_read_search+eip' -- -c 'cat foo'
foo
~/tmp $ sudo ~/src/libcap/progs/capsh --keep=1 --user=backup --inh=cap_dac_read_search --addamb=cap_dac_read_search --drop=all --caps='= cap_dac_read_search+eip' -- -c 'echo bar >foo'
/bin/bash: foo: Permission denied
$ sudo ~/src/libcap/progs/capsh --keep=1 --user=backup --inh=cap_dac_read_search --addamb=cap_dac_read_search --drop=all --caps='= cap_dac_read_search+eip' == --print
Current: = cap_dac_read_search+eip
Bounding set =
Ambient set =cap_dac_read_search
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
 secure-no-ambient-raise: no (unlocked)
uid=618(backup)
gid=618(backup)
groups=618(backup)
```

So this is definitely an alternative way of achieving this. It just happens to require the latest unreleased `capsh` tool (which comes with libcap).

ThomasWaldmann:
If there is not good reason any more to move this kind of code into borg (and not having it in a OS-specific wrapper), I guess I'ld prefer keeping it outside.

It is a good thing to drop privileges, but it requires platform-specific code and adds addtl. dependencies.

As a side note: it is also possible to have another user own the repo files by using ssh://thatuser@... for the repo url.

c4rlo:
Okay, that's fine with me.

For now I've fixed my wrapper program to work with borg as well as attic. I may opensource it at some point.

ThomasWaldmann:
@c4rlo shouldn't be such a wrapper rather generic and be contained in "cap-tools" or sudo or so?

enkore:
That'd be "capsh" (see his examples above), though a bit complex to use.

ThomasWaldmann:
I am closing this one, thanks for investigating this.

Maybe we could add something to docs or faq about running borg as non-root, with partial root capabilities?

PR #1188, opened 2016-06-19
Correctly handle multiple LD_PRELOAD entries; fixes #1111

### LINKED ISSUE #1111: failure with eatmydata "tried to call a logger before setup_logging() was called"
As said on Debian bug 825729, with eatmydata borg fails.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=825729
```
==================================== ERRORS ====================================
___ ERROR collecting .pybuild/pythonX.Y_3.5/build/borg/testsuite/__init__.py ___
borg/testsuite/__init__.py:10: in <module>
    from ..xattr import get_all
borg/xattr.py:71: in <module>
    logger.error(msg)
borg/logger.py:173: in error
    return self.__logger.error(*args, **kw)
borg/logger.py:150: in __logger
    raise Exception("tried to call a logger before setup_logging() was called")
E   Exception: tried to call a logger before setup_logging() was called
--------------------------------------------------------------------------------
```

jrtc27:
Issue #1111 is caused by the fact that the entire contents of LD_PRELOAD is used for libc_name. When eatmydata is used, LD_PRELOAD looks something like `libfakeroot-sysv.so:libeatmydata.so`, but if any other library instead of libeatmydata was preloaded, it would also cause this issue. Thus LD_PRELOAD should be properly parsed by splitting it on `[ :]` (both of which are valid separators) and iterating over it to find a libfakeroot entry.

### DIFF
--- a/src/borg/xattr.py
+++ b/src/borg/xattr.py
@@ -2,6 +2,7 @@
 """
 import errno
 import os
+import re
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


src/borg/xattr.py
```diff
@@ -52,17 +53,21 @@ def get_all(path, follow_symlinks=True):
...
-        # 1.20.2 has been confirmed to have xattr support
-        # 1.18.2 has been confirmed not to have xattr support
-        # Versions in-between are unknown
-        libc_name = LD_PRELOAD
-        XATTR_FAKEROOT = True
-
+if sys.platform.startswith('linux'):
+    LD_PRELOAD = os.environ.get('LD_PRELOAD', '')
+    preloads = re.split("[ :]", LD_PRELOAD)
+    fakeroot_preload = None
```
ThomasWaldmann: what is this ^^^ good for?

jrtc27: Oops, nothing! That was from a previous revision... Amended and pushed.

ThomasWaldmann:
@jrtc27 Thanks for fixing this! I assume you have tested this practically and seen that this actually fixes the issue in question?

jrtc27:
Yes, it now builds without issue with fakeroot and eatmydata (not necessarily a good idea for actual backups, but fine for running the test suite) at the same time, and I checked it hit that code path by raising an exception rather than assigning to libc_name.

ThomasWaldmann:
OK, LGTM. Merge?

enkore:
Ditto.

PR #1191, opened 2016-06-21
Fix confused I/O Error reporting/handling, fixes #1138

### LINKED ISSUE #1138: confused I/O Error reporting / handling

andela-kerinoso:
This replaces PR #1180.

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -66,6 +66,8 @@ class Repository:
         self.exclusive = exclusive
 
     def __del__(self):
+        if not os.path.exists(self.path):
+            return
         if self.lock:
             self.close()
             assert False, "cleanup happened in Repository.__del__"
@@ -83,6 +85,9 @@ class Repository:
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None:
             self.rollback()
+            if hasattr(exc_val, 'errno'):
+                if exc_val.errno in [errno.EIO, errno.ENXIO]:
+                    return
         self.close()
 
     def create(self, path):
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from hashlib import sha256
 from operator import attrgetter
 import argparse
+import errno
 import functools
 import inspect
 import io
@@ -266,6 +267,7 @@ class Archiver:
         if restrict_dev is not None and st.st_dev != restrict_dev:
             return
         status = None
+        io_errorcodes = [errno.EIO, errno.ENXIO]
         # Ignore if nodump flag is set
         if has_lchflags and (st.st_flags & stat.UF_NODUMP):
             return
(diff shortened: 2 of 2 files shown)


andela-kerinoso:
I would appreciate feedbacks on this new approach before I spend time on writing test to simulate scenarios that would improve the current coverage.

enkore:
I don't see the difference to your earlier approach?

borg/repository.py
```diff
@@ -83,6 +85,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None:
             self.rollback()
+            if hasattr(exc_val, 'errno'):
+                if exc_val.errno in [errno.EIO, errno.ENXIO]:
+                    return
         self.close()
```
enkore: The context manager's `__exit__` is "on the outside", since the context manager is used like this in borg.archiver:

```
with Repository:
  do_create/delete/...:
     do_something_with_data
        do_something_with_repo
```

Therefore the exceptions that "travel by" this method could've been raised pretty much anywhere.

andela-kerinoso: If what you mean by _I don't see the difference to your earlier approach?_ is the use of return instead of re-raise, note that `__exit__()` methods should not reraise the passed-in exception (Reference: Python Documentation) since the exception will be processed normally upon exit from this method.
Furthermore, return was used to prevent further exceptions to be raised due to the call of `self.close()` in the case of repository/cache related OSError.
It's true that the exceptions that "travel by" this method could've been raised pretty much anywhere. Do you think `errno.EIO` and `errno.ENXIO` could have been caused by other than repository/cache related OSError?

enkore: > If what you mean by _I don't see the difference to your earlier approach?_ ...

I mean that it performs the same error handling as before just written down a different way.

> It's true that the exceptions that "travel by" this method could've been raised pretty much anywhere. Do you think errno.EIO and errno.ENXIO could have been caused by other than repository/cache related OSError?

Exactly. One can't distinguish the two cases (repository/cache-caused or input-caused) by looking at error codes.

andela-kerinoso: But for input-caused, neither `errno.EIO` nor `errno.ENXIO` occurs; rather `errno.ENOENT` occurs.

errno.ENOENT  == No such file or directory
errno.EIO == I/O error
errno.ENXIO == No such device or address

enkore: (Almost) any error code can occur on any file. EIO on input files can and does happen for example (due to drive/media errors).

ThomasWaldmann:
@andela-kerinoso we discussed this issue on IRC and that it maybe makes sense if @enkore fixes it in case it takes you too long to do a correct / best working fix (like @enkore proposed).

The problem with this issue is that it is a rather severe problem that can corrupt data, so we want to fix this rather soon. Also, it blocks fixing some other issues schedule for 1.0.4 (including the one you started with).

andela-kerinoso:
Sorry for the delay. Alright! I will be on the watch for the right fix; I will definitely learn from it.

Issue #1195, opened 2016-06-22
Power-loss during backup corrupts repository index


verygreen:
Apparently my backup host lost power in the middle of backup today,.

Aside from understandably stuck lock the repo itself is corrupted, all operations are failing with:

```
hashindex: /backup1/mordor/index.146096: fread header failed (expected 18, got 0)
```

File index.146096 does indeed exist and is of zero size.

I am trying borg check --repair now to see if that helps, but in general data creating code seems to need an audit to ensure atomicity of transactions at the very least.

The setup is:
Server running borg 1.0.1 ("binary package" from releases page) in serve mode, client is running borg 1.0.3 from Fedora23. x86 Linux on both ends (centos7.2 on the server). ssh is the pipe to communicate.

enkore:
The repository isn't corrupted, it's the index that is truncated. I added some code to master a while back that automatically handles this kind of situation. Delete the offending index+hints file and it should recover with no problems.

verygreen:
If the repository cannot be accessed with an error and to clear the error you need to perform some recovery (manual) steps, then it is corrupted in my book.

verygreen:
It's interestign that if the zero-sized file is removed, it's automatically rebuilt without a hitch even with 1.0.3
So t sounds like detecting a zero-sized file and deleting it should be a good workaround?

On a different note, I use ext4 in ordered journal mode. Looking at the code and at strace for the sequence of events, the truncated file should not be possible at all. I wonder if there's an ext4 bug at play.

verygreen:
So, apparently my understanding of ext4 had a gap - the delayed allocation did not come into the picture, meaning the dirty pages were not part of transaction commit.
Here's a longish commend from Ted Tso  from 2009th explaining the problem:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/317781/comments/45

I guess doing fdatasync on the index file before rename or doing a hackish
"create index.tmp2, create and write index.tmp; rename index.tmp into index.tmp2 ; rename index.tmp2 (to force the allocation) into proper Index.XXX." workaround are too messy for the problem at hand.
Instead it looks like just testing that the index file is zero length and removing it is the most straightforward solution here.

enkore:
I guess we could discuss whether a backport of the changeset I mentioned above (#858) to 1.0-maint for 1.0.4 is worthwhile (or if documenting it is enough).

ThomasWaldmann:
[...]

PR #1199, opened 2016-06-23
Ignore empty index file.

verygreen:
Empty index file is most likely a result from an unclean
shutdown in the middle of write, e.g. on ext4 with delayed
allocation enabled (default).
Ignoring such a file would get it recreated by other parts of code,
where as not ignoring it leads to an exception about
not being able to read enough bytes from the index.

Signed-off-by: Oleg Drokin <green@linuxhacker.ru>

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -129,7 +129,9 @@ class Repository:
         shutil.rmtree(self.path)
 
     def get_index_transaction_id(self):
-        indices = sorted((int(name[6:]) for name in os.listdir(self.path) if name.startswith('index.') and name[6:].isdigit()))
+        indices = sorted(int(fn[6:])
+                         for fn in os.listdir(self.path)
+                         if fn.startswith('index.') and fn[6:].isdigit() and os.stat(os.path.join(self.path, fn)).st_size != 0)
         if indices:
             return indices[-1]
         else:


borkd:
Would it be worthwhile to spit out a warning message that a zero size index was encountered, and then proceed with recreating it?

verygreen:
But what's the value of it?
My logic is simple:
If index file is zero sized, then it's either truncated due to unclean shutdown or there's a deeper problem.
If it's due to unclean shutdown, it just gets rebuilt and all is fine, no need to even mention it.
If it's part of some deeper problem - rebuild would fail too, the file would not be deleted/overwritten and some other exception would kill the whole thing.
Then a human would further analyze the problems.

What would you do with the message/flag if the rebuild was ok?

borg/repository.py
```diff
@@ -129,7 +129,7 @@ def destroy(self):
         shutil.rmtree(self.path)
 
     def get_index_transaction_id(self):
-        indices = sorted((int(name[6:]) for name in os.listdir(self.path) if name.startswith('index.') and name[6:].isdigit()))
+        indices = sorted((int(name[6:]) for name in os.listdir(self.path) if name.startswith('index.') and name[6:].isdigit() and os.stat(os.path.join(self.path, name)).st_size != 0))
```
ThomasWaldmann: how about reformatting this a little and dropping unneeded parens:
```
indices = sorted(int(fn[6:])
                 for fn in os.listdir(self.path)
                 if fn.startswith('index.') and fn[6:].isdigit() and os.stat(os.path.join(self.path, fn)).st_size != 0)
```

enkore:
Logic-wise the patch is ok and following from [Ted Tso's explanation](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/317781/comments/45) linked in #1195 it seems that this is a common case on ext4. It certainly has less impact than back-porting the error handling of `master` back into 1.0-maint.

verygreen:
I refreshed the patch taking the style into account.

verygreen:
btw, current code checks for size == 0, but in fact should we reject all files below a certain size as clearly corrupted and in need of regeneration, or are there cases where this potentially might hide problems we better know about?

I also found that when the index is regenerated, there is a message already:
```
Remote: Replaying segments   0%
```

enkore:
The loader code in _hashindex.c verifies that the file has the correct size (from the header)

verygreen:
Except it does not work because we clearly see it fails with assertion if it cannot read enough bytes for the header.

enkore:
Which I meant with "verifies"...

borkd:
@verygreen: apologies for replying so late. I'm not against automated regeneration, just for an explicit warning/info message when it happens. In general I would argue against assuming that unclean shutdown is the only way an index could become truncated, especially when experimental/cloudy storage backends are in use. 

When your datasets are large enough and you pay for IO, acting on early warnings (failing fast and deciding to work from known good snapshot for example), may be cheaper than ultimately failing and having to restart after lots of disk operations. In such cases automated rebuild may not be the preferred solution.

PR #1211, opened 2016-06-27
Fix incorrect propagation of OSErrors in create code

enkore:
Pulling a thumbdrive during a backup now produces the correct result:

```
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 81, in wrapper
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 247, in do_create
    create_inner(archive, cache)
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 219, in create_inner
    read_special=args.read_special, dry_run=dry_run)
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 275, in _process
  File "/home/mabe/Projekte/_oss/borg/borg/archive.py", line 594, in process_file
    chunks.append(cache.add_chunk(self.key.id_hash(chunk), chunk, self.stats))
  File "/home/mabe/Projekte/_oss/borg/borg/cache.py", line 365, in add_chunk
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 467, in put
    segment, offset = self.io.write_put(id, data)
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 706, in write_put
    fd = self.get_write_fd(raise_full=raise_full)
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 600, in get_write_fd
    self._write_fd = open(self.segment_filename(self.segment), 'ab')
OSError: [Errno 5] Input/output error: '/run/media/mabe/00B524275BF67732/repo/data/0/9'

During handling of the above exception, another exception occurred:

  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 1569, in main
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 1506, in run
    return args.func(args)
  File "/home/mabe/Projekte/_oss/borg/borg/archiver.py", line 81, in wrapper
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 86, in __exit__
    self.close()
  File "/home/mabe/Projekte/_oss/borg/borg/repository.py", line 181, in close
    self.lock.release()
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 291, in release
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 203, in modify
    roster = self.load()
  File "/home/mabe/Projekte/_oss/borg/borg/locking.py", line 182, in load
    data = json.load(f)
  File "/home/mabe/.pyenv/versions/3.5.1/lib/python3.5/json/__init__.py", line 265, in load
    return loads(fp.read(),
OSError: [Errno 5] Input/output error

Platform: Linux tiger 4.5.7-1-ck #1 SMP PREEMPT Wed Jun 8 08:42:43 EDT 2016 x86_64 
Linux: arch  
Borg: 1.0.4.dev46+ngbc0a838.d20160627  Python: CPython 3.5.1
PID: 27050  CWD: /home/mabe/Projekte/_oss/borg/issue1138
sys.argv: ['/home/mabe/.pyenv/versions/3.5.1/bin/borg', 'create', '/run/media/mabe/00B524275BF67732/repo::testbreaking', '...............', '-svp']
SSH_ORIGINAL_COMMAND: None

zsh: exit 2     borg create /run/media/mabe/00B524275BF67732/repo::testbreaking  -svp
```

Tried it a few times, logic seems solid. I don't think I missed a spot, but that's what we do code reviews for, isn't it? ;)

_Rather curious_: ntfs-3g journaling doesn't seem to be very reliable. After a few pulls with no repository corruption I get an EIO on open(2) on a data file and a rather ugly ntfs-3g message:

> Jun 27 20:59:35 tiger ntfs-3g[27314]: Trying to read non-allocated mft records (83 > 82): Nicht erlaubter Seek (_seek not permissible_)
> Jun 27 20:59:35 tiger ntfs-3g[27314]: Trying to read non-allocated mft records (83 > 82): Nicht erlaubter Seek

Even root can't touch that inode anymore, not even rm the dirent. I'll see whether Windows chkdsk can fix that, or whether I have to reformat my thumb drive?!

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -1,4 +1,5 @@
 from binascii import hexlify
+from contextlib import contextmanager
 from datetime import datetime, timezone
 from getpass import getuser
 from itertools import groupby
@@ -43,6 +44,37 @@ has_lchflags = hasattr(os, 'lchflags')
 
 flags_normal = os.O_RDONLY | getattr(os, 'O_BINARY', 0)
 flags_noatime = flags_normal | getattr(os, 'O_NOATIME', 0)
+
+
+class InputOSError(Exception):
+    """Wrapper for OSError raised while accessing input files."""
+    def __init__(self, os_error):
+        self.os_error = os_error
+        self.errno = os_error.errno
+        self.strerror = os_error.strerror
+        self.filename = os_error.filename
+
+    def __str__(self):
+        return str(self.os_error)
+
+
+@contextmanager
+def input_io():
+    """Context manager changing OSError to InputOSError."""
+    try:
+        yield
+    except OSError as os_error:
+        raise InputOSError(os_error) from os_error
+
+
+def input_io_iter(iterator):
+    while True:
+        try:
+            with input_io():
+                item = next(iterator)
+        except StopIteration:
+            return
+        yield item
 
 
 class DownloadPipeline:
@@ -464,12 +496,14 @@ Number of files: {0.stats.nfiles}'''.format(
         }
         if self.numeric_owner:
             item[b'user'] = item[b'group'] = None
-        xattrs = xattr.get_all(path, follow_symlinks=False)
+        with input_io():
+            xattrs = xattr.get_all(path, follow_symlinks=False)
         if xattrs:
             item[b'xattrs'] = StableDict(xattrs)
         if has_lchflags and st.st_flags:
             item[b'bsdflags'] = st.st_flags
-        acl_get(path, item, st, self.numeric_owner)
+        with input_io():
+            acl_get(path, item, st, self.numeric_owner)
         return item
 
     def process_dir(self, path, st):
@@ -504,7 +538,7 @@ Number of files: {0.stats.nfiles}'''.format(
         uid, gid = 0, 0
         fd = sys.stdin.buffer  # binary
         chunks = []
-        for chunk in self.chunker.chunkify(fd):
+        for chunk in input_io_iter(self.chunker.chunkify(fd)):
             chunks.append(cache.add_chunk(self.key.id_hash(chunk), chunk, self.stats))
         self.stats.nfiles += 1
         t = int_to_bigint(int(time.time()) * 1000000000)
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
hmm, how about stat()? And the bsdflags on linux also do OS calls.

enkore:
The stat is done in Archiver._process and try/except'ed directly (l 259).

Linux BSD-Flags is only in master but we'll need to remember to fix that there as well, yes.

ThomasWaldmann:
As a side note: in the changeset I am currently working on (in-file checkpoints), I unified the stdin and file reading code, so the duplication you have seen will be removed there.

ThomasWaldmann:
Changeset looks good. As usual, the traceback isn't pretty / helpful for non-pythonista eyes.

As we can expect such exceptions and it is not our fault / nothing to debug for us, maybe you could subclass your exception class from Error (the one without traceback), so users will get a more pretty error msg.

enkore:
Hmm, not sure about that. The exception causing the traceback is OSError unchanged, so we could catch and remove the traceback in main(), but since not every OSError carries a file name (most don't ... those concerning open(2) normally do) it can be hard to know *where* the error is in the system. There the traceback can help.

ThomasWaldmann:
ah, right, confused that. %-)

borg/archive.py
```diff
@@ -45,6 +46,37 @@
...
+def input_io_iter(iterator):
+    while True:
+        try:
+            with input_io():
+                item = next(iterator)
+        except StopIteration:
+            return
+        yield item
+
+
```
ThomasWaldmann: in case it is likely that we need this outside archive module later: move it to helpers.py?

borg/remote.py
```diff
@@ -287,6 +290,8 @@ def fetch_from_cache(args):
                         type, msgid, error, res = unpacked
                         if msgid in self.ignore_responses:
                             self.ignore_responses.remove(msgid)
+                            if error:
+                                handle_error(error, res)
```
ThomasWaldmann: did you analayse the impact this missing code had? for changelog / commit comment.

enkore: The way I understand it had a similar impact as the original issue. An error (during e.g. a put()) would still be visible in the log, but not cause a program abort and may have stored/corrupted data.

Re. changelog I think it's the same bug in terms of description, that was caused two different ways for local and remote repos each.

ThomasWaldmann:
merge?

PR #1217, opened 2016-06-28
in-file checkpoints

ThomasWaldmann:


src/borg/archive.py
```diff
@@ -1142,7 +1164,7 @@ def item_is_hardlink_master(item):
                     'source' not in item and
                     not matcher.match(item.path))
 
-        for item in archive.iter_items():
+        for item in archive.iter_items(filter=lambda item: 'checkpoint' not in item):
```
enkore: It would make sense to move this check to iter_items() since this conditions is always used.

ThomasWaldmann: I thought about that, but we might add support for listing / extracting these checkpoint archives (e.g. via a cmdline option).

enkore:
It would be a bit cleaner to not store checkpoint-items in the final archive, however that would need to selectively flush ChunksBuffer and I'm not sure whether that's a good idea. Considering the low overhead in terms of complexity and also storage etc. this solution looks quite good.

src/borg/archive.py
enkore: In that case it may be preferable to have something like a `pre_predicate` on the `Archive` that is ANDed with the iter_items predicate. That would make it easy to control this with a cmdline switch via e.g. `with_archive`.

ThomasWaldmann:
updated (yeah, much simpler that way, see last commit).

ThomasWaldmann:
Ignore E731 ("do not assign lambda expressions, use def") globally?

ThomasWaldmann:
merge?

src/borg/archive.py
```diff
@@ -700,22 +713,39 @@ def process_symlink(self, path, st):
...
+                checkpoint_item.path += '.checkpoint_%d' % checkpoint_number
+                checkpoint_item.checkpoint = checkpoint_number
+                checkpoint_number += 1
+                self.add_item(checkpoint_item, show_progress=False)
+                self.write_checkpoint()
+                self.last_checkpoint = time.time()
+                # we have saved the checkpoint file, but we will reference the same
+                # chunks also from the checkpoint or the "real" file we save next
+                for chunk in checkpoint_item.chunks:
+                    cache.chunk_incref(chunk.id, stats)
```
enkore: hmm, is this correct?

suppose

```
chunks a, b, c, ...

abc
refcounts: 1, 1, 1

checkpoint_1

refcounts: 2, 2, 2

Here I'm already starting to wonder. If we abort now, the main archive won't be there, only the checkpoint. Therefore the refcounts should be 1, 1, 1?

def
refcounts: 2, 2, 2, 1, 1, 1

checkpoint_2

refcoutns: 3, 3, 3, 2, 2, 2

checkpoint_1 will be gone now. If we abort now, the main archive is still not there. So the refcounts should be 6× (1,)?

Comment (ThomasWaldmann) [inline src/borg/archive.py]: the incrementing happens after the checkpoint is made / committed, so if it aborts afterwards, the changes will be rolled back (IIRC).

checkpoint_1 won't "be gone" if you write checkpoint_2. all in-file checkpoints will be in the archive.

Comment (enkore) [inline src/borg/archive.py]: checkpoint_n => the checkpoint *archive* is committed. and these are removed again.

Comment (enkore) [inline src/borg/archive.py]: I think it would be correct if there are no additional increments:

chunks a, b, c, ...

abc
refcounts: 1, 1, 1

checkpoint_1 archive committed

Ok, there is only checkpoint_1 archive pointing to these chunks.

def
refcounts: 1, 1, 1, 1, 1, 1

checkpoint_2 archive committed, checkpoint_1 gone.

Ok, there is only checkpoint_2 archive pointing to these chunks.

Done, commit main archive:
removes checkpoint_2

refcounts 6× (1,)

Ok, only the main archive points there.

---

abc
refcounts: 1, 1, 1

checkpoint_1 archive committed

Ok, there is only checkpoint_1 archive pointing to these chunks.

We crash now.

User restarts backup. File is unchanged.

abc
refcounts: 1+1, 1+1, 1+1

othermain.checkpoint_1 archive commited

Ok: the checkpoint_1 archive of the previous run still exists and uses these chunks as well as our current checkpoint_1

def
refcounts: 1+1, 1+1, 1+1, 1, 1, 1

checkpoint_2 archive committed, checkpoint_1 gone.

\<same\>

Done, commit main archive:
removes checkpoint_2

refcounts 2, 2, 2, 1, 1, 1

Ok, main archive points there, and there is still the checkpoint_1 archive of the old, aborted run around.

**Comment by ThomasWaldmann** [comment]
I'll check it again. I added the incref after noticing orphaned chunks.

File: src/borg/archive.py [hunk 3]
```diff
@@ -700,22 +713,39 @@ def process_symlink(self, path, st):
...
+        for data in backup_io_iter(self.chunker.chunkify(fd, fh)):
+            item.chunks.append(cache.add_chunk(self.key.id_hash(data), Chunk(data, **chunk_kw), stats))
+            if self.show_progress:
+                self.stats.show_progress(item=item, dt=0.2)
+            if self.checkpoint_interval and time.time() - self.last_checkpoint > self.checkpoint_interval:
+                checkpoint_item = Item(internal_dict=item.as_dict())
+                checkpoint_item.path += '.checkpoint_%d' % checkpoint_number
+                checkpoint_item.checkpoint = checkpoint_number
+                checkpoint_number += 1
+                self.add_item(checkpoint_item, show_progress=False)
```
ThomasWaldmann: @enkore did you consider that? ^^^

That means that there will be following items in the metadata stream:
```
filename/ITEM    chunks
---------------------------------------------------------
foo.checkpoint_1 abc
foo.checkpoint_2 abcdef
foo              abcdefghi
---------------------------------------------------------
refcount         333222111
```
While the checkpoint ARCHIVES get removed, the item metadata stream is always being appended, so the checkpoint ITEMS stay (the final, finishes archive has all the checkpoint ITEMS).

enkore: Ah, I had a wrong premise above. Yes, now it makes sense. Thanks!

ThomasWaldmann:
TODO: update docs (FAQ)

docs/faq.rst
```diff
@@ -265,27 +265,21 @@ Once your backup has finished successfully, you can delete all
...
-After that, you can remove the parts again and backup the huge file(s) in
-their original form. This will now work a lot faster as a lot of content chunks
-are already in the repository.
-
-After you have successfully backed up the huge original file(s), you can remove
-the temporary archive you made from the parts.
-
-We realize that this is just a better-than-nothing workaround, see :issue:`1198`
-for a potential solution.
+This is not a problem any more, see previous FAQ item.
```
enkore: (since release xy)

ThomasWaldmann: I had the "since..." first, but removed it again - the docs are bound to a release anyway.

enkore: Ok then I'd say remove it for now or change (now/later) to explain the extraction issue / workaround. Because "not a problem any more" is still a bit strange, even though the docs may be tied to a particular release.

src/borg/archive.py
```diff
@@ -719,24 +719,39 @@ def process_symlink(self, path, st):
...
+            item.chunks = item.chunks[from_chunk:]
+            item.path += '.checkpoint_%d' % number
+            item.checkpoint = number
+            number += 1
+            self.add_item(item, show_progress=False)
+            self.write_checkpoint()
+            # we have saved the checkpoint file, but we will reference the same
+            # chunks also from the final, complete file:
+            for chunk in item.chunks:
+                cache.chunk_incref(chunk.id, stats)
```
ThomasWaldmann: hmm, i guess this is wrong since the "parts" change.

before, the duplicate references of the leading chunks happened with every checkpoint commit, as the beginnings of the checkpoint files were overlapping, like abc, abcdef, abcdefghi.
but since working with parts, we do not have the overlap any more while we are doing part files:
abc, def, ghi.
we need to duplicate the references after saving the last part, when we save the final (non-part) full file.

src/borg/archive.py
```diff
@@ -700,22 +718,56 @@ def process_symlink(self, path, st):
...
+                    # if we already have created a part item inside this file, we want to put the final
+                    # chunks (if any) into a part item also (so all parts can be concatenated to get
+                    # the complete file):
+                    from_chunk, part_number = write_part(item, from_chunk, part_number)
+                    self.last_checkpoint = time.time()
+
+                # if we created part files, we have referenced all chunks from the part files,
+                # but we also will reference the same chunks also from the final, complete file:
+                for chunk in item.chunks:
+                    cache.chunk_incref(chunk.id, stats)
```
enkore: This looks correct to me now.

PR #1218, opened 2016-06-28
Vagrantfile: use openbsd 5.9, fixes #716

### LINKED ISSUE #716: vagrant: openbsd testing machine broken
Somehow testing on openbsd doesn't work right now. It worked in the past and it is unclear what's the problem right now.

Issues:
- /etc/resolv.conf has non-working 127.0.1.1 entry (but manually setting it to a working nameserver also does not help)
- "vagrant up" does not install rsync automatically
- even when adding rsync manually with pkg_add rsync, "vagrant rsync openbsd64" does not synchronize the borg directory
- when running vagrant provision, llfuse installation fails with "unable to determine system"

It would be great if some openbsd user could have a look, as I can't test on openbsd until this is fixed.

ThomasWaldmann:
machine already has python3.4.
rsync works now.
fuse does not work on openbsd (and breaks borg install, if tried).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -109,7 +109,6 @@ def packages_openbsd
     pkg_add lz4
     # pkg_add fuse  # does not install, sdl dependency missing
     pkg_add git  # no fakeroot
-    pkg_add python-3.4.2
     pkg_add py3-setuptools
     ln -sf /usr/local/bin/python3.4 /usr/local/bin/python3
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
merge?

enkore:
llfuse not building on OpenBSD is the same state I see on OpenBSD outside Vagrant as well.

PR #1228, opened 2016-06-30
Add --append-only to borg serve [WIP]

### LINKED ISSUE #1168: borg serve --append-only
Extracting the easy part from #1164:

We have this only as repo config option, which is a bit awkward to control (you need to edit the config).

If `borg serve` had this option, it could be used in `.ssh/authorized_keys` like this:
```
command="borg serve --append-only ..." ssh-rsa <key used for not-always-trustable backup clients>
command="borg serve ..." ssh-rsa <key used for backup management>
```

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/35160230-borg-serve-append-only)

PlasmaPower:
Resolves #1168. Not currently well tested, but I'm hoping it will be easier to test by just writing automated tests for it.

TODO: tests

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -163,7 +163,7 @@ class Archiver:
     def do_serve(self, args):
         """Start in server mode. This command is usually not used manually.
         """
-        return RepositoryServer(restrict_to_paths=args.restrict_to_paths).serve()
+        return RepositoryServer(restrict_to_paths=args.restrict_to_paths, append_only=args.append_only).serve()
 
     @with_repository(create=True, exclusive=True, manifest=False)
     def do_init(self, args, repository):
@@ -1220,6 +1220,8 @@ class Archiver:
         subparser.set_defaults(func=self.do_serve)
         subparser.add_argument('--restrict-to-path', dest='restrict_to_paths', action='append',
                                metavar='PATH', help='restrict repository access to PATH')
+        subparser.add_argument('--append-only', dest='append_only', action='store_true',
+                               help='only allow appending to repository segment files')
         init_epilog = textwrap.dedent("""
         This command initializes an empty repository. A repository is a filesystem
         directory containing the deduplicated data from zero or more archives.
@@ -2114,8 +2116,9 @@ class Archiver:
             if result.func != forced_result.func:
                 # someone is trying to execute a different borg subcommand, don't do that!
                 return forced_result
-            # the only thing we take from the forced "borg serve" ssh command is --restrict-to-path
+            # we only take specific options from the forced "borg serve" command:
             result.restrict_to_paths = forced_result.restrict_to_paths
+            result.append_only = forced_result.append_only
         return result
 
     def parse_args(self, args=None):
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -96,7 +96,7 @@ class Repository:
     class ObjectNotFound(ErrorWithTraceback):
         """Object with key {} not found in repository {}."""
 
-    def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=True):
+    def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=True, append_only=False):
         self.path = os.path.abspath(path)
         self._location = Location('file://%s' % self.path)
         self.io = None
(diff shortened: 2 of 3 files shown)


src/borg/repository.py
```diff
@@ -219,7 +220,9 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
             raise self.InvalidRepository(path)
         self.max_segment_size = self.config.getint('repository', 'max_segment_size')
         self.segments_per_dir = self.config.getint('repository', 'segments_per_dir')
-        self.append_only = self.config.getboolean('repository', 'append_only', fallback=False)
+        # append_only can be set in the constructor
+        # it shouldn't be overriden (True -> False) here
+        self.append_only = self.append_only or self.config.getboolean('repository', 'append_only', fallback=False)
```
enkore: +1

src/borg/archiver.py
```diff
@@ -1220,6 +1220,8 @@ def build_parser(self, prog=None):
         subparser.set_defaults(func=self.do_serve)
         subparser.add_argument('--restrict-to-path', dest='restrict_to_paths', action='append',
                                metavar='PATH', help='restrict repository access to PATH')
+        subparser.add_argument('--append-only', dest='append_only', action='store_true',
+                               help='only allow appending to files')
```
ThomasWaldmann: ... to repository segment files?

src/borg/archiver.py
```diff
@@ -2114,8 +2116,8 @@ def get_args(self, argv, cmd):
             if result.func != forced_result.func:
                 # someone is trying to execute a different borg subcommand, don't do that!
                 return forced_result
-            # the only thing we take from the forced "borg serve" ssh command is --restrict-to-path
```
ThomasWaldmann: maybe don't kill the comment completely, but modify it to say:

we only take specific options from the forced "borg serve" command:

src/borg/archiver.py
enkore: Maybe just say "enable append-only mode regardless of repository config". That thing is a bit too complex to describe in a --help synopsis imho.

src/borg/archiver.py
PlasmaPower: Done. Comment isn't marked as outdated for whatever reason.

src/borg/repository.py
```diff
@@ -219,7 +220,9 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
             raise self.InvalidRepository(path)
         self.max_segment_size = self.config.getint('repository', 'max_segment_size')
         self.segments_per_dir = self.config.getint('repository', 'segments_per_dir')
-        self.append_only = self.config.getboolean('repository', 'append_only', fallback=False)
+        # append_only can be set in the constructor
+        # it shouldn't be overriden (True -> False) here
```
ThomasWaldmann: typo: overridden

ThomasWaldmann:
code looks good to me. some tests would be great.

ThomasWaldmann:
1.0-maint?
- new feature (or rather: added interface to existing feature)
+ making better security practically usable (via different ssh keys / force commands in authorized keys)

ThomasWaldmann:
also add docs (when everything else is done / decided).

PlasmaPower:
I don't see any existing tests for `borg serve`, is that correct?

@ThomasWaldmann Do you want me to close this PR and open one to 1.0-maint? If you're asking for my opinion, I'm not the one to ask as this is the first time I've touched this codebase.

ThomasWaldmann:
@PlasmaPower borg serve is tested indirectly via RemoteRepository tests (all the tests in archiver tests run for local repo and for "remote" repo).

The 1.0-maint decision is rather about policy than code base. I am tending to put it into 1.0-maint because the version will be used for years (due to e.g. Ubuntu LTS) and offering practically usable security is a big pro. The alternative would be to somehow edit append_only = 0 in the repo config if one wants to prune/delete and edit it back to 1 afterwards, but I guess not many people will go for that.

ThomasWaldmann:
OK, let's target 1.0-maint (later 1.0-maint will get merged into master, so it will be there automatically).

PlasmaPower:
Replaced by #1229

PR #1229, opened 2016-06-30
Add --append-only to borg serve (1.0-maint)

PlasmaPower:
Replaces #1228 which resolves #1168

PR #1232, opened 2016-06-30
Add --append-only to borg serve (1.0-maint)

### LINKED ISSUE #1168: borg serve --append-only

PlasmaPower:
Fixes #1168

Replaces #1229 in which I accidentally committed `.eggs`.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -115,7 +115,7 @@ class Archiver:
     def do_serve(self, args):
         """Start in server mode. This command is usually not used manually.
         """
-        return RepositoryServer(restrict_to_paths=args.restrict_to_paths).serve()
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
\o/

Issue #1139, opened 2016-06-05
"borg delete" should allow to delete broken archives as well


enkore:
--force for borg delete, so that you can delete broken archives. I.e. it tries to delete normally, but doesn't fail if the metadata is corrupted, and the manifest entry is always removed at least [and the archive superblock deleted from the repo].

ThomasWaldmann:
hmm, guess we could need that for dealing with #1138.

ThomasWaldmann:
fixed by #1239 in 1.0-maint.

Issue #1130, opened 2016-06-04
Error during prune command


nesneros:
Hi, I have an archive that seems to be corrupted. Both prune and check gives an error for the specific archive. I use the --list option for prune and it stops at the same archive every time. 

The repository is on a samba share mounted with smbnetfs.

Error with prune:
```
Traceback (most recent call last):
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 1520, in main
    exit_code = archiver.run(args)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 1465, in run
    return args.func(args)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 81, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 610, in do_prune
    Archive(repository, key, manifest, archive.name, cache).delete(stats)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archive.py", line 448, in delete
    if b'chunks' in item:
TypeError: argument of type 'int' is not iterable

Platform: Linux xps15 4.4.0-22-generic #40-Ubuntu SMP Thu May 12 22:03:46 UTC 2016 x86_64 x86_64
Linux: Ubuntu 16.04 xenial
Borg: 1.0.3  Python: CPython 3.5.1+
PID: 19383  CWD: /home/jan
sys.argv: ['/home/jan/borg-env/bin/borg', 'prune', '-H', '99999', '--list', '-v', '/home/jan/sambadrive']
SSH_ORIGINAL_COMMAND: None
```

Error with check: 
```
Traceback (most recent call last):
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 1520, in main
    exit_code = archiver.run(args)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 1465, in run
    return args.func(args)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 81, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/home/jan/borg-env/lib/python3.5/site-packages/borg/archiver.py", line 148, in do_check
    last=args.last, prefix=args.prefix, save_space=args.save_space):
[...]

PR #1239, opened 2016-07-01
WIP: Forced archive delete

ThomasWaldmann:
See issue #1139.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -24,6 +24,8 @@ from .helpers import Error, uid2user, user2uid, gid2group, group2gid, \
 from .platform import acl_get, acl_set
 from .chunker import Chunker
 from .hashindex import ChunkIndex
+from .repository import Repository
+
 import msgpack
 
 ITEMS_BUFFER = 1024 * 1024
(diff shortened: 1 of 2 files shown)


borg/archive.py
```diff
@@ -467,23 +467,40 @@ def rename(self, name):
...
-                    for chunk_id, size, csize in item[b'chunks']:
-                        self.cache.chunk_decref(chunk_id, stats)
-        if progress:
-            pi.finish()
-        self.cache.chunk_decref(self.id, stats)
+                pi.finish()
+        except Exception:
+            if not forced:
+                raise
+            error = True
```
enkore: I think we can do a bit better and avoid the diaper by safeguarding the unpacker with a try/except msgpack.Xyz and the for chunk_... in item loop as well as the item decref with a separate try/except. So even if the first chunk would somehow get lost the following chunks should still be decref'd / deleted. (While a msgpack error wouldn't try any further).

ThomasWaldmann: Hmm, what's the point?:
- We would still need borg check --repair, right?
- The final result would be the same, right?
- We would risk overlooking some exception. If we do not catch it, borg delete fails and we can't get rid of the archive.

enkore: - Depends on what caused the corruption, but likely.
- Not necessarily. If chunks are missing from the repository then this approach may leave large amounts of unused data (think first few chunks of a large file like a disk image go missing) that wouldn't be removed otherwise
- True, on the other hand over-zealous exception handling led to other bugs in the past.

Also I think it'd be better if the error handling would be more detailed (without tracebacks that can't point to the problem here anyway). If the items stream is corrupted it should tell that the archive metadata was corrupted so the data couldn't be deleted (which seems to be a common case due to #1138). If objects are absent from the Cache that should be fatal, because the cache is corrupted. If objects are absent from the repository that shouldn't be fatal, but it should say that data was already missing.

ThomasWaldmann:
@enkore my impression from the code is that `borg check --repair` removes orphaned chunks.

about error handling: yes, it ignores all the errors that might occur while items metadata is read and reference counters are decreased and just makes sure the archive is gone. wasn't that the point? if we assume that another error happens, like a bug in progress indicator or ssh connection breaks down, it would either manage to get rid of the archive or cause another error when it tries to commit.

enkore:
> if we assume that another error happens, like a bug in progress indicator or ssh connection breaks down, it would either manage to get rid of the archive or cause another error when it tries to commit.

I don't see how. For example, a write error on the repository would be silenced by this and a commit would still be made, further corrupting the repository.

borg/archive.py
```diff
@@ -467,23 +467,40 @@ def rename(self, name):
...
+        # in forced delete mode, we make sure to delete at least the manifest entry,
+        # if possible also the archive superblock, even if processing the items raises
+        # an exception.
+        try:
+            self.cache.chunk_decref(self.id, stats)
+        except Exception:
```
enkore: KeyError, ObjectNotFound

ThomasWaldmann:
OK, trying to refine error handling.

borg/archive.py
```diff
@@ -468,6 +470,23 @@ def rename(self, name):
...
+            try:
+                self.cache.chunk_decref(id, stats)
+            except KeyError as e:
+                # chunkid missing from chunks index - chunks cache corrupted!
+                raise FatalError(str(e))
+            except Repository.ObjectNotFound as e:
+                # object not in repo - strange, but we wanted to delete it anyway.
+                # XXX doesn't this also mean that our chunks cache is corrupted?
+                # XXX obviously it had a refcount of 1 before decref, but the chunk
+                # XXX has already been deleted from the repo.
```
enkore: The repository can also lose objects "on it's own" e.g. corrupted segment and check --repair which then doesn't "rescue" corrupted objects (LoggedIO.recover_segment).

ThomasWaldmann: hmm, yes, guess that might happen if a repo-only check is done.
not if all archives are also checked, then lost-chunk references will get replaced with a zero chunk.
afterwards, the chunks cache would resync and not reference such lost chunks any more.

ThomasWaldmann:
@enkore about tuple unpacking, TypeError, ValueError - can you point to the places?

borg/archive.py
```diff
@@ -467,23 +469,53 @@ def rename(self, name):
...
+            items_ids = self.metadata[b'items']
+            pi = ProgressIndicatorPercent(total=len(items_ids), msg="Decrementing references %3.0f%%", same_line=True)
+            for (i, (items_id, data)) in enumerate(zip(items_ids, self.repository.get_many(items_ids))):
+                if progress:
+                    pi.show(i)
+                unpacker.feed(self.key.decrypt(items_id, data))
+                chunk_decref(items_id, stats)
+                for item in unpacker:
+                    if b'chunks' in item:
+                        for chunk_id, size, csize in item[b'chunks']:
```
enkore: I thought about this one, but I'm unsure whether #1130 / #1138 could have caused this part to be truncated. I think this would only be possible if the item dict became longer than one metadata chunk. I don't know it from the top of the head, but assuming something like 40 bytes per list entry (ID + 2*size + framing), 64K upper chunk bound = some 1600 list entries => approximately 3 GB, so quite easily within reach.

enkore: What *may* be possible is that we don't get the expected types. That would be the long-metadata case, say a long list begins that would span multiple chunks but then one of these are dropped, msgpack tries hard to decode and usually succeds. Then the list elements could be totally different from what'd we expect, for example another item dict.

enkore: I think just wrapping the whole "for item in unpacker:" loop in a try/except ValueError, TypeError should be ok with

```
except ...:
    if not forced: raise
```

ThomasWaldmann:
Hmm, maybe we should use RobustUnpacker and robust_iterator?

borg/archive.py
```diff
@@ -467,23 +469,59 @@ def rename(self, name):
...
+                raise
+            error = True
+        # in forced delete mode, we try hard to delete at least the manifest entry,
+        # if possible also the archive superblock, even if processing the items raises
+        # some harmless exception.
+        chunk_decref(self.id, stats)
         del self.manifest.archives[self.name]
+        if error:
+            logger.warning('forced deletion succeeded, but the deleted archive was corrupted.')
+            logger.warning('borg check --repair is required to clean up the pieces.')
```
enkore: "clean up the pieces" is unspecific. Write `borg check --repair may be required to free all space` instead?

PR #1242, opened 2016-07-02
fixes for --read-special mode

ThomasWaldmann:
please review if the semantics is now right.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -269,7 +269,7 @@ class Archiver:
         # Ignore if nodump flag is set
         if has_lchflags and (st.st_flags & stat.UF_NODUMP):
             return
-        if stat.S_ISREG(st.st_mode) or read_special and not stat.S_ISDIR(st.st_mode):
+        if stat.S_ISREG(st.st_mode):
             if not dry_run:
                 try:
                     status = archive.process_file(path, st, cache, self.ignore_inode)
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -29,7 +29,7 @@ from .upgrader import AtticRepositoryUpgrader, BorgRepositoryUpgrader
 from .repository import Repository
 from .cache import Cache
 from .key import key_creator, RepoKey, PassphraseKey
-from .archive import backup_io, BackupOSError, Archive, ArchiveChecker, CHUNKER_PARAMS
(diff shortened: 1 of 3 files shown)


PlasmaPower:
Can we reliably test this?

borg/archiver.py
```diff
@@ -304,10 +304,16 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
                 status = archive.process_symlink(path, st)
         elif stat.S_ISFIFO(st.st_mode):
             if not dry_run:
-                status = archive.process_fifo(path, st)
+                if not read_special:
+                    status = archive.process_fifo(path, st)
+                else:
+                    status = archive.process_file(path, st, cache)
```
textshell: st.st_mode still says this is a special file. Should that be masked out?

borg/archiver.py
```diff
@@ -304,10 +304,16 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
...
         elif stat.S_ISCHR(st.st_mode) or stat.S_ISBLK(st.st_mode):
             if not dry_run:
-                status = archive.process_dev(path, st)
+                if not read_special:
+                    status = archive.process_dev(path, st)
+                else:
+                    status = archive.process_file(path, st, cache)
```
textshell: st.st_mode still says this is a special file. Should that be masked out?

enkore: process_file could set the mode bits of the item to S_IFREG unconditionally.

textshell:
So other then your TODOs this looks good.

about the symlinks. For many use cases in /dev following symlinks for block special devices is needed i’m not even sure udev/kernel commit to what is a symlink and what the real device node as abi. On the other hand, --read-special is not a good option for follow all symlinks.

ThomasWaldmann:
mode should be ok now. i amended the previous commit with a cleaner implementation of that.

ThomasWaldmann:
ok, added symlink handling.

textshell:
I think we should spell out what --read-special means in the documentation. It certainly not guessable anymore. For the implementation i think we can hardly restrict this anymore without breaking usages.

>       --read-special        open and read block and character special files 
                            and FIFOs as if they were regular files. 
                            Also follows symlink pointing to these kinds files.

ThomasWaldmann:
collapse / merge?

textshell:
seems ready to merge.

enkore:
lgtm, rebase/merge

PR #1246, opened 2016-07-03
WIP: Deleting stale locks

### LINKED ISSUE #562: better locking

verygreen:
If BORG_UNIQ_HOSTNAME shell variable is set, stale locks
in both cache and repository are deleted.
I wanted to add a parameter to the repo config, but alas, we lock
it before the read (makes sense, really), so the only two options
are the shell variable and a command line switch.
Do we want the switch? What would the name be? Any good ways
to enable this permanently for those wishing so if the config is
off-limits?

Currently if the option is set - we would detect if a lock is of
known format and would delete it.
We only update roster if we are to do an exclusive lock. Should we care
to prune stale locks from roster if we are only getting a shared lock,
or can we leave it be until such a time exclusive locker comes around?

Also I guess I need feedback on other named things and also my
weak python-fu ;)

This fixes #562
eventually ;)

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -70,6 +70,7 @@ class Cache:
         self.key = key
         self.manifest = manifest
         self.path = path or os.path.join(get_cache_dir(), repository.id_str)
+        self.unique_hostname = bool(os.environ.get('BORG_UNIQUE_HOSTNAME'))
         self.do_files = do_files
         # Warn user before sending data to a never seen before unencrypted repository
         if not os.path.exists(self.path):
@@ -171,7 +172,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
     def open(self, lock_wait=None):
         if not os.path.isdir(self.path):
             raise Exception('%s Does not look like a Borg cache' % self.path)
-        self.lock = UpgradableLock(os.path.join(self.path, 'lock'), exclusive=True, timeout=lock_wait).acquire()
+        self.lock = UpgradableLock(os.path.join(self.path, 'lock'), exclusive=True, timeout=lock_wait, kill_stale_locks=self.unique_hostname).acquire()
         self.rollback()
 
     def close(self):
--- a/src/borg/locking.py
+++ b/src/borg/locking.py
@@ -1,6 +1,8 @@
+import errno
 import json
 import os
 import socket
+import sys
 import time
 
 from .helpers import Error, ErrorWithTraceback
@@ -17,8 +19,34 @@ _hostname = socket.gethostname()
 
 def get_id():
     """Get identification tuple for 'us'"""
+
+    # If changing the thread_id to ever be non-zero, also revisit the check_lock_stale() below.
     thread_id = 0
     return _hostname, _pid, thread_id
+
+
+def check_lock_stale(host, pid, thread):
+    """Check if the host, pid, thread combination corresponds to a dead process on our local node or not."""
+    if host != _hostname:
+        return False
+
+    if thread != 0:
+        # Currently thread is always 0, if we ever decide to set this to a non-zero value, this code needs to be revisited too to do a sensible thing
+        return False
+
+    try:
+        # This may not work in Windows.
+        # This does not kill anything, 0 means "see if we can send a signal to this process or not".
+        # Possible errors: No such process (== stale lock) or permission denied (not a stale lock)
+        # If the exception is not raised that means such a pid is valid and we can send a signal to it (== not a stale lock too).
+        os.kill(pid, 0)
+        return False
+    except OSError as err:
+        if err.errno != errno.ESRCH:
+            return False
+        pass
+
+    return True
 
 
 class TimeoutTimer:
@@ -107,12 +135,14 @@ class ExclusiveLock:
     This makes sure the lock is released again if the block is left, no
     matter how (e.g. if an exception occurred).
     """
-    def __init__(self, path, timeout=None, sleep=None, id=None):
+    def __init__(self, path, timeout=None, sleep=None, id=None, kill_stale_locks=False):
         self.timeout = timeout
         self.sleep = sleep
         self.path = os.path.abspath(path)
         self.id = id or get_id()
         self.unique_name = os.path.join(self.path, "%s.%d-%x" % self.id)
+        self.ok_to_kill_stale_locks = kill_stale_locks
+        self.stale_warning_printed = False
 
     def __enter__(self):
         return self.acquire()
@@ -135,6 +165,8 @@ class ExclusiveLock:
             except FileExistsError:  # already locked
                 if self.by_me():
                     return self
+                if self.kill_stale_lock():
+                    pass
                 if timer.timed_out_or_sleep():
                     raise LockTimeout(self.path)
             except OSError as err:
@@ -158,6 +190,47 @@ class ExclusiveLock:
     def by_me(self):
         return os.path.exists(self.unique_name)
 
+    def kill_stale_lock(self):
+        for name in os.listdir(self.path):
+
+            try:
+                host_pid, thread_str = name.rsplit('-', 1)
+                host, pid_str = host_pid.rsplit('.', 1)
+                pid = int(pid_str)
+                thread = int(thread_str)
+            except ValueError:
+                # Malformed lock name? Or just some new format we don't understand?
+                # It's safer to just exit
+                return False
+
+            if not check_lock_stale(host, pid, thread):
+                return False
+
+            if not self.ok_to_kill_stale_locks:
+                if not self.stale_warning_printed:
+                    print(("Found stale lock %s, but not deleting because BORG_UNIQUE_HOSTNAME is not set." % name), file=sys.stderr)
+                    self.stale_warning_printed = True
+                return False
+
+            try:
+                os.unlink(os.path.join(self.path, name))
+                print(("Killed stale lock %s." % name), file=sys.stderr)
+            except OSError as err:
+                if not self.stale_warning_printed:
+                    print(("Found stale lock %s, but cannot delete due to %s" % (name, str(err))), file=sys.stderr)
+                    self.stale_warning_printed = True
+                return False
+
+        try:
+            os.rmdir(self.path)
+        except OSError:
+            # Directory is not empty = we lost the race to somebody else
+            # Permission denied = we cannot operate anyway
+            # other error like EIO = we cannot operate and it's unsafe too.
+            return False
+
+        return True
+
     def break_lock(self):
         if self.is_locked():
             for name in os.listdir(self.path):
@@ -172,17 +245,34 @@ class LockRoster:
     Note: you usually should call the methods with an exclusive lock held,
     to avoid conflicting access by multiple threads/processes/machines.
     """
-    def __init__(self, path, id=None):
+    def __init__(self, path, id=None, kill_stale_locks=False):
         self.path = path
         self.id = id or get_id()
+        self.ok_to_kill_zombie_locks = kill_stale_locks
 
     def load(self):
         try:
             with open(self.path) as f:
                 data = json.load(f)
+
+            # Just nuke the stale locks early on load
+            if self.ok_to_kill_zombie_locks:
+                for key in (SHARED, EXCLUSIVE):
+                    elements = set()
+                    try:
+                        for e in data[key]:
+                            (host, pid, thread) = e
+                            if not check_lock_stale(host, pid, thread):
+                                elements.add(tuple(e))
+                            else:
+                                print(("Removed stale %s roster lock for pid %d." % (key, pid)), file=sys.stderr)
+                        data[key] = list(list(e) for e in elements)
+                    except KeyError:
+                        pass
         except (FileNotFoundError, ValueError):
             # no or corrupt/empty roster file?
             data = {}
+
         return data
 
     def save(self, data):
@@ -228,18 +318,18 @@ class UpgradableLock:
     This makes sure the lock is released again if the block is left, no
     matter how (e.g. if an exception occurred).
     """
-    def __init__(self, path, exclusive=False, sleep=None, timeout=None, id=None):
+    def __init__(self, path, exclusive=False, sleep=None, timeout=None, id=None, kill_stale_locks=False):
         self.path = path
         self.is_exclusive = exclusive
         self.sleep = sleep
         self.timeout = timeout
         self.id = id or get_id()
         # globally keeping track of shared and exclusive lockers:
-        self._roster = LockRoster(path + '.roster', id=id)
+        self._roster = LockRoster(path + '.roster', id=id, kill_stale_locks=kill_stale_locks)
         # an exclusive lock, used for:
         # - holding while doing roster queries / updates
         # - holding while the UpgradableLock itself is exclusive
-        self._lock = ExclusiveLock(path + '.exclusive', id=id, timeout=timeout)
+        self._lock = ExclusiveLock(path + '.exclusive', id=id, timeout=timeout, kill_stale_locks=kill_stale_locks)
 
     def __enter__(self):
         return self.acquire()
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -107,6 +107,7 @@ class Repository:
         self.do_lock = lock
         self.do_create = create
         self.exclusive = exclusive
+        self.unique_hostname = bool(os.environ.get('BORG_UNIQUE_HOSTNAME'))
 
     def __del__(self):
         if self.lock:
@@ -210,7 +211,7 @@ class Repository:
         if not os.path.isdir(path):
             raise self.DoesNotExist(path)
         if lock:
-            self.lock = UpgradableLock(os.path.join(path, 'lock'), exclusive, timeout=lock_wait).acquire()
+            self.lock = UpgradableLock(os.path.join(path, 'lock'), exclusive, timeout=lock_wait, kill_stale_locks=self.unique_hostname).acquire()
         else:
             self.lock = None
         self.config = ConfigParser(interpolation=None)


src/borg/cache.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, repository, key, manifest, path=None, sync=True, do_files=Fal
         self.key = key
         self.manifest = manifest
         self.path = path or os.path.join(get_cache_dir(), repository.id_str)
+        if os.environ.get('BORG_UNIQ_HOSTNAME'):
```
ThomasWaldmann: uniq is the unix command, but i guess we rather spell it correctly as unique.

ThomasWaldmann: another thing that came to my mind:
if we create and store a uuid clientside, we could use that and not need the user to assert uniqueness of hostname.

src/borg/cache.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, repository, key, manifest, path=None, sync=True, do_files=Fal
...
+            self.uniq_hostname = True
+        else:
+            self.uniq_hostname = False
```
ThomasWaldmann: just for your python powers: equivalent is this one-liner:

x = bool(os.environment.get(...))

src/borg/cache.py
```diff
@@ -171,7 +175,7 @@ def _do_open(self):
     def open(self, lock_wait=None):
         if not os.path.isdir(self.path):
             raise Exception('%s Does not look like a Borg cache' % self.path)
-        self.lock = UpgradableLock(os.path.join(self.path, 'lock'), exclusive=True, timeout=lock_wait).acquire()
+        self.lock = UpgradableLock(os.path.join(self.path, 'lock'), exclusive=True, timeout=lock_wait, kill_dead_lock=self.uniq_hostname).acquire()
```
ThomasWaldmann: kill STALE lock? if it would be dead, we would not need to kill it. :D

src/borg/locking.py
```diff
@@ -158,6 +161,41 @@ def is_locked(self):
...
+                # Malformed lock name? Or just soem new format we don't understand?
+                # It's safer to just exit
+                return False
+
+            if host != _hostname:
+                return False
+
+            try:
+                # This may not work in Windows.
+                os.kill(int(pid), 0)
```
ThomasWaldmann: maybe move the int(pid) part into above try/except ValueError.
so in case that "pid" is not a string that can be converted to int, you can use the same error handler to bail out early.

src/borg/locking.py
```diff
@@ -107,12 +107,13 @@ class ExclusiveLock:
...
     matter how (e.g. if an exception occurred).
     """
-    def __init__(self, path, timeout=None, sleep=None, id=None):
+    def __init__(self, path, timeout=None, sleep=None, id=None, kill_dead_lock=False):
         self.timeout = timeout
         self.sleep = sleep
         self.path = os.path.abspath(path)
         self.id = id or get_id()
         self.unique_name = os.path.join(self.path, "%s.%d-%x" % self.id)
+        self.ok_to_kill_zombie_locks = kill_dead_lock;
```
ThomasWaldmann: no semicolon

src/borg/locking.py
```diff
@@ -158,6 +161,41 @@ def is_locked(self):
...
+
+            if host != _hostname:
+                return False
+
+            try:
+                # This may not work in Windows.
+                os.kill(int(pid), 0)
+                return False
+            except OSError:
+                pass
```
ThomasWaldmann: what circumstances trigger OSError here?
do we need further checks to make sure some specific circumstances?

src/borg/locking.py
ThomasWaldmann: do we need to wait() until it is actually terminated?

src/borg/locking.py
```diff
@@ -158,6 +161,41 @@ def is_locked(self):
...
+
+            # We should be ok not checking thread no matter what it is, right?
+            # Possible implications on Windows?
+
+            if not self.ok_to_kill_zombie_locks:
+                print(("Found zombie lock %s, but not deleting because repo option uniq_hostname is not set" % name))
```
ThomasWaldmann: you don't need the inner parens. add a dot at the end of the msg.

src/borg/locking.py
```diff
@@ -158,6 +161,41 @@ def is_locked(self):
...
+                return False
+
+            os.unlink(os.path.join(self.path, name))
+
+        # If there's a race and some other locker appeared, we'd get an exception
+        # Do we need to handle it, and if so - how?
+        os.rmdir(self.path)
```
ThomasWaldmann: iirc, the lock is the DIRECTORY. The file inside is just informative.
So, a successful unlink is releasing the lock.

src/borg/locking.py
```diff
@@ -206,6 +244,9 @@ def modify(self, key, op):
         except KeyError:
             elements = set()
         if op == ADD:
+            # If we are adding exclusive lock, that means all other entries here are stale, get rid of them.
+            if key == EXCLUSIVE:
+                elements = set()
```
ThomasWaldmann: have to check this again.

but iirc: an exclusive lock can only be obtained, if all shared locks are gone.

PlasmaPower:
Can we try to read `/etc/machine-id` first, and if that exists use it instead of the hostname and automatically enable this feature?

We could also try to read the mac of the default network interface, which would be cross-platform.

src/borg/locking.py
verygreen: one such circumstance that I was after is that the proess does not exist.

I guess another somewhat feasible one is "permission denied". We probably need to catch it separately, even though we'll fail with exception trying to delete the lock too (do we want to catch that anyway?)
In general I assume it's a bad idea running borg instances from different users on the local box.

src/borg/locking.py
verygreen: the inner parents are there because I had file=sys.stderr before and was not sure if I want to ressurrect that.
The goal here is to make it propagate from server to the client as well.

src/borg/locking.py
verygreen: That does not match my reading of the code.
If the directory is there, but there's nothing inside, the other locker would happily proceed adding their own "informative" file.

src/borg/locking.py
ThomasWaldmann: i guess for the same repo it is usually also the same uid/gid running the borg processes.
otherwise you'ld likely get file permission issues anyway.

src/borg/locking.py
verygreen: What if it takes ages for it to terminate?
I don't think you can wait for a process that you are not a parent of anyway.

```
     The wait4() call provides a more general interface for programs that need
     to wait for certain child processes, that need resource utilization sta-
     tistics accumulated by child processes, or that require options.  The
     other wait functions are implemented using wait4().
```

Note that kill(pid, 0) does not kill anything.
It just checks if the pid is a valid process or not (and that you can signals to it as well).

src/borg/cache.py
verygreen: Whatever uuid you store, it becomes not uniq once you clone a vm or something, right? I thought that was the main reason for not making this the default, otherwise we can just switch to such uuids (generatedand stored in .cache/borg on startup if not there already) and be done with it .

ThomasWaldmann:
Not sure if the MAC addr is a good idea. It is not always unique even for real hardware (although it should be) and also finding out and accessing the default interface might be rather platform specific.

Same goes for machine-id as it seems. It is there on systemd machines. But BSD, OS X, Windows?
So we would rather generate an own unique ID and store it on the client in the config, that works everywhere.

src/borg/locking.py
verygreen: Well, remember there was a discussion on the itrc from two people already that wanted to run their serve process under different gids because some of them were local and some were remote.
There was some trickery with umask and sgid. This is the case that would get caught here - we'll not be able to kill a more privileged process potentially, but we will be able to remove the lock file I imagine, due to umask.

src/borg/cache.py
ThomasWaldmann: the hostname is also the same after cloning, right?
but maybe it is remembered more easily to be changed than some borg uuid. hmm...

src/borg/locking.py
ThomasWaldmann: Oh, I thought kill kills. :)

src/borg/cache.py
verygreen: Frequently enough hostname is NOT something that is cloned with a VM, because it's obtaned via a reverse DNS lookup, for example.

e.g. I have a single subdir that holds an OS instance that I export with NFS, I then run tens of virtual machines out of it (nfsroot), every one of those virtual machines has a different hostname because of different DNS record.

src/borg/locking.py
verygreen: Ha, no, it does not kill anything. it sends a signal, and signal 0 has a special meaning of "don't send anything, just do all the checks" and is frequently used as a way to see if a particular process is there or not. Certainyl beats spawning "ps -ef | grep pid" and other such sily techniques that are also a lot less portable across OSes.

I will add a comment to this effect.

verygreen:
The real trick is how to determine if our local unique ID has become stale and is no longer unique (because a vm was cloned or whatever).

verygreen:
Also another vector is - if we store our own ID - where do we store it? in home dir? but that could be a network share and therefore visible from multiple nodes making it non-unique as well.

PlasmaPower:
> Not sure if the MAC addr is a good idea. It is not always unique even for real hardware (although it should be)

In which cases? I'm talking about the hardware MAC here, not the MAC actually used for networking (can be changed software level).

> finding out and accessing the default interface might be rather platform specific

```py
from uuid import getnode
```

verygreen:
I guess mac (no matter if it was software overridden even) is a good proxy for a unique id, otherwise things would just not work on the network and we only care about all of this for networked access anyway.

enkore:
Obtaining some unique or maybe generated ID depending on various intransparent factors (=uuid.getnode) IMHO is a bad match for this.

Sticking with the hostname is clear and easy to understand + verify.

PlasmaPower:
@enkore If two MACs on the same network are the same, then they've got bigger problems. Also, they'll be much more independent then hostnames. We can specify an override in the config if you'd like.

verygreen:
We cannot specify an override in the config because we lock before we have a chance of reading the repo config.
Unless there's some node-global config somewhere that I missed that could be read before opening a repository.

enkore:
> @enkore If two MACs on the same network are the same, then they've got bigger problems. Also, they'll be much more independent then hostnames. We can specify an override in the config if you'd like.

getnode's documentation is rather unclear on this. I read it as

- may or may not return a hardware-address derived number of a random NIC in the system
- may or may not generate a random number and return that instead
- may or may not store that number somewhere in the FS ("The first time this runs, it may launch a separate program")

PlasmaPower:
@enkore I'd assume the cache is in memory. Also, this is Python's recommended way of doing this specific task, do you have a better way?

src/borg/locking.py
```diff
@@ -172,17 +237,34 @@ class LockRoster:
...
+        self.ok_to_kill_zombie_locks = kill_stale_locks
 
     def load(self):
         try:
             with open(self.path) as f:
                 data = json.load(f)
+
+            # Just nuke the stale locks early on load
+            if self.ok_to_kill_zombie_locks:
+                for key in (SHARED, EXCLUSIVE):
```
verygreen: this is ... um, unelegant.
I wonder if there's an easier way to do these checks.

ThomasWaldmann:
The env var could be BORG_UUID=... so users can put there anything they like and borg would just use that string, if given, as unique id.

verygreen:
I feel like BORG_UUID could be a separate thing altogether with its own patch because the changes are going to be mostly orthogonal to this path other than one extra check for the unique_hostname var.

src/borg/locking.py
```diff
@@ -21,6 +23,29 @@ def get_id():
     return _hostname, _pid, thread_id
 
 
+def check_lock_data(host, pid, thread):
+    """Check if the host, pid, thread combination corresponds to a dead process on our local node or not."""
+    if host != _hostname:
+        return False
+
+    try:
+        # This may not work in Windows.
```
enkore: Should work; on Windows you don't really send signals, but open a process and then call various functions on the process handle. Python shouldn't be different; opening the process for a dead PID fails and should raise OSError. The error number will be different from ESRCH ...

src/borg/locking.py
```diff
@@ -21,6 +23,29 @@ def get_id():
...
+        # If the exception is not raised that means such a pid is valid and we can send a signal to it (== not a stale lock too).
+        os.kill(pid, 0)
+        return False
+    except OSError as err:
+        if err.errno != errno.ESRCH:
+            return False
+        pass
+
+    # We should be ok not checking thread no matter what it is, right?
+    # Possible implications on Windows?
```
enkore: Ok

enkore:
Logic-wise this looks sound to me. A bit of "nitpicking" was already mentioned; I would add that using both "zombie lock" and "stale lock" to refer to the same thing is confusing, use only "stale lock".

src/borg/locking.py
verygreen: I have read that at first kill(pid, 0) used to kill a process in windows for good. And then that was changed and it started to return error at all times.
So this nees to be verified.

enkore: I'd have to look at the implementation to see if/whether signal is checked for it's value. If it's not any signal number may terminate a process, yes...

verygreen:
It's stale lock on the outside, zombie lock on the inside, if you noticed.
But I can change that - no problems.

src/borg/locking.py
verygreen: So, should I instead just ensure that thread is always 0, if it's not 0, something way too new is going on and we return False at the onstart?

src/borg/locking.py
verygreen: Since windows port is in the works, I feel like potentially questionable Windows-related bits should be at least highlighted to be checked as the port is progressing.

enkore: Yes, leave the comments in :)

enkore:
"what to do with the hostname field?"-discussion -> #1253

src/borg/locking.py
```diff
@@ -17,10 +19,36 @@
 
 def get_id():
     """Get identification tuple for 'us'"""
+
+    # If changing the thread_id to ever be non-zero, also revisit the check_lock_stale() below.
     thread_id = 0
     return _hostname, _pid, thread_id
 
 
+def check_lock_stale(host, pid, thread):
```
ThomasWaldmann: is_lock_stale

src/borg/locking.py
```diff
@@ -158,6 +190,47 @@ def is_locked(self):
...
+                # Malformed lock name? Or just some new format we don't understand?
+                # It's safer to just exit
+                return False
+
+            if not check_lock_stale(host, pid, thread):
+                return False
+
+            if not self.ok_to_kill_stale_locks:
+                if not self.stale_warning_printed:
+                    print(("Found stale lock %s, but not deleting because BORG_UNIQUE_HOSTNAME is not set." % name), file=sys.stderr)
```
ThomasWaldmann: use logger.warning?

src/borg/locking.py
```diff
@@ -158,6 +190,47 @@ def is_locked(self):
...
+                    self.stale_warning_printed = True
+                return False
+
+            try:
+                os.unlink(os.path.join(self.path, name))
+                print(("Killed stale lock %s." % name), file=sys.stderr)
+            except OSError as err:
+                if not self.stale_warning_printed:
+                    print(("Found stale lock %s, but cannot delete due to %s" % (name, str(err))), file=sys.stderr)
```
ThomasWaldmann: maybe log the whole path?

src/borg/locking.py
```diff
@@ -172,17 +245,34 @@ class LockRoster:
...
+                    elements = set()
+                    try:
+                        for e in data[key]:
+                            (host, pid, thread) = e
+                            if not check_lock_stale(host, pid, thread):
+                                elements.add(tuple(e))
+                            else:
+                                print(("Removed stale %s roster lock for pid %d." % (key, pid)), file=sys.stderr)
+                        data[key] = list(list(e) for e in elements)
+                    except KeyError:
```
ThomasWaldmann: this is a quite wide-range error handler, it also would catch KeyErrors in check_lock_stale or (when using logger.warning() later), in the logger code.

just use this?:
```
for e in data.get(key, []):
    ...
```

src/borg/locking.py
```diff
@@ -158,6 +190,47 @@ def is_locked(self):
     def by_me(self):
         return os.path.exists(self.unique_name)
 
+    def kill_stale_lock(self):
+        for name in os.listdir(self.path):
```
ThomasWaldmann: somehow this code seems to be undecided whether it should deal with multiple filenames encountered in self.path or not.

you use "for" (== would be able to deal with multiple files), OTOH, you often just "return", so it stops processing.

verygreen: This is as designed:

we loop around in the lock filenames doing stuff with them until we meet one that we cannot kill for whatever reason.
If we do meet one like that - there's no point in processing any further filenames so we just exit with False

We only return True if we completed the whole loop.

src/borg/locking.py
verygreen: I think logger was not really available here, but I guess we can pull it in somehow

src/borg/locking.py
ThomasWaldmann: ah, ok. branch also needs a rebase.

ThomasWaldmann:
BTW, just FYI: i did another round of researching 3rd party lockfile libs. There is still nothing great out there, everything I found has obvious issues.

ThomasWaldmann:
BTW, master or 1.0-maint?

verygreen:
we need both, I think.
But I want a master version first and then we can backport.

enkore:
I'm thinking master first, maybe see it at least in a RC in action before backporting?

enkore:
Superseded by #1674

PR #1248, opened 2016-07-03
Documentation: Fix literate code markup

nomeata:


ThomasWaldmann:
Good find, thanks!

But the stuff needs to get changed in the code. The stuff you see in .rst.inc files is taken from the builtin help of borg, so it has to be changed there (see archiver.py) and the .rst.inc files will automatically get it.

Also, the PR should be against 1.0-maint, so it gets fixed there and also later gets merged into master.

nomeata:
Oh, you have generated files in the git repository? How nasty :-)

nomeata:
I’ll open a new PR against 1.0-maint.

ThomasWaldmann:
It was less nasty than generating them on readthedocs / via setup.py.

PR #1251, opened 2016-07-03
Documentation: Explicate that Fnmatch is default

### LINKED ISSUE #1247: Documentation unclear: What is the default exclude style?
The section http://borgbackup.readthedocs.io/en/stable/usage.html#borg-help-patterns tells me about the various exclusion formats, and says “Explicit style selection is necessary when a non-default style is desired”. But nowhere in the docs I could find what the default style is, and if I can globally change the default style.

Also, currently on that site at least, the examples are not typeset properly, i.e. not as literate code.

nomeata:
This fixes #1247. It also regenerates the usage documentation, so that
styling fixex in that section (as well as other existing changes) make
it into the files in docs/.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -726,26 +726,27 @@ class Archiver:
     helptext = {}
     helptext['patterns'] = textwrap.dedent('''
         Exclusion patterns support four separate styles, fnmatch, shell, regular
-        expressions and path prefixes. If followed by a colon (':') the first two
-        characters of a pattern are used as a style selector. Explicit style
-        selection is necessary when a non-default style is desired or when the
-        desired pattern starts with two alphanumeric characters followed by a colon
-        (i.e. `aa:something/*`).
+        expressions and path prefixes. By default, fnmatch is used. If followed
+        by a colon (':') the first two characters of a pattern are used as a
+        style selector. Explicit style selection is necessary when a
+        non-default style is desired or when the desired pattern starts with
+        two alphanumeric characters followed by a colon (i.e. `aa:something/*`).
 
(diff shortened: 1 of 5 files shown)


borg/archiver.py
```diff
@@ -726,26 +726,27 @@ def do_break_lock(self, args, repository):
...
-            matching any character not specified. For the purpose of these patterns,
-            the path separator ('\\' for Windows and '/' on other systems) is not
-            treated specially. Wrap meta-characters in brackets for a literal match
-            (i.e. `[?]` to match the literal character `?`). For a path to match
-            a pattern, it must completely match from start to end, or must match from
-            the start to just before a path separator. Except for the root path,
-            paths will never end in the path separator when matching is attempted.
-            Thus, if a given pattern ends in a path separator, a '*' is appended
-            before matching is attempted.
+            This is the default style.  These patterns use a variant of shell
```
ThomasWaldmann: this is a bit hard to review. did you change anything else than adding the first sentence?

nomeata: No, I did not. I was not sure whether this might be printed verbatim somewhere, so I ran vim’s `qgg` on it.

ThomasWaldmann:
can you easily split this up into 1 commit with your change and 1 commit with the automatically generated stuff?

nomeata:
Hmm. I squashed them, because other commits in the repo were setting precedent of including both changes. I can try to do some git reset magic, if you want.

ThomasWaldmann:
Oh. Well, just took it as is. Personally, I like to separate automatic stuff from manual stuff, but maybe it was not always done like that.

nomeata:
Thanks

PR #1257, opened 2016-07-04
PoC lockless fuse mount

verygreen:
This  is a proof of concept of the fuse mount that does not hold repository locked.

I cannot actually get it to work because of issues with tox and llfuse not doing well together, but I wanted to show this idea anyway.
In order to highlight the changes needed (not too big) this is lean and simple, but as the result, has some style problems with indentation.

This would also need some extra "revalidation" logic, so that when we e.g. reopen the repo, we check if it's the same one as last time we opened it and if it's not, blow the cache way and repopulate it on demand again (some optimizations are possible here too, but might not be needed in the end).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -726,8 +726,8 @@ class Archiver:
             self.print_error('Loading fuse support failed [ImportError: %s]' % str(e))
             return self.exit_code
 
-        if not os.path.isdir(args.mountpoint) or not os.access(args.mountpoint, os.R_OK | os.W_OK | os.X_OK):
-            self.print_error('%s: Mountpoint must be a writable directory' % args.mountpoint)
+        if not os.path.isdir(args.mountpoint):
+            self.print_error('%s: Mountpoint must be a directory' % args.mountpoint)
             return self.exit_code
 
         with cache_if_remote(repository) as cached_repo:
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -726,8 +726,8 @@ def do_mount(self, args, repository, manifest, key):
             self.print_error('Loading fuse support failed [ImportError: %s]' % str(e))
             return self.exit_code
 
-        if not os.path.isdir(args.mountpoint) or not os.access(args.mountpoint, os.R_OK | os.W_OK | os.X_OK):
-            self.print_error('%s: Mountpoint must be a writable directory' % args.mountpoint)
```
textshell: See fuse(8) for why this check is a good idea to guard from later failures from fusemount. Also mixing unrelated changes into a pull request...

verygreen: I see.
Ok, I did not know fuse was supposed to be run by any user in the fuse group, so they needed to guard against users mounting stuff on top of /etc and so on. It makes sense then.

enkore: But this isn't supposed to be a security check, right? right?

PlasmaPower: It's to prevent future errors I think.

textshell: Yes, the security check is in the suid binary of course. But an error message from borg is better than just whatever fusermount decides to output.</bikeshed>

ThomasWaldmann:
btw, flake8 checker is unhappy.

enkore:
PoC = not intended for merge and no activity = close.

PR #1263, opened 2016-07-05
Repository: cleanup on ENOSPC

### LINKED ISSUE #1262: Document recovery from full disk
I filled my backup partition with borg and now even borg list is crashing.
The documentation should descript how to recover from a full disk. 
[This comment](https://github.com/borgbackup/borg/issues/1114#issuecomment-222656759) suggests:

```
you can try deleting the segment files of the last incomplete backup and then run borg check --repair. After that, use borg delete or prune with the --save-space flag.
```

But i don't know how to safely do this and am afraid of destroying my whole backup.

enkore:
This automates what one would have to do otherwise (find last committed segment, delete everything after that)

Fixes #1262 #1114

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -84,6 +84,12 @@ class Repository:
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None:
+            no_space_left_on_device = exc_type is OSError and exc_val.errno == errno.ENOSPC
+            # The ENOSPC could have originated somewhere else besides the Repository. The cleanup is always safe, unless
+            # EIO or FS corruption ensues, which is why we specifically check for ENOSPC.
+            if self._active_txn and no_space_left_on_device:
+                logger.warning('No space left on device, cleaning up partial transaction to free space.')
+                self.io.cleanup(self.io.get_segments_transaction_id())
             self.rollback()
         self.close()
 
@@ -543,7 +549,7 @@ class LoggedIO:
         return None
 
     def get_segments_transaction_id(self):
-        """Verify that the transaction id is consistent with the index transaction id
+        """Return last committed segment
         """
         for segment, filename in self.segment_iterator(reverse=True):
             if self.is_committed_segment(segment):
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -11,11 +11,15 @@ The next section continues by showing how backups can be automated.
 Important note about free space
 -------------------------------
 
-Before you start creating backups, please make sure that there is **always**
+Before you start creating backups, please make sure that there is *always*
 a good amount of free space on the filesystem that has your backup repository
-(and also on ~/.cache). It is hard to tell how much, maybe 1-5%.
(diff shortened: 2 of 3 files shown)


borg/repository.py
```diff
@@ -84,6 +84,11 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None:
+            no_space_left_on_device = exc_type is OSError and exc_val.errno == errno.ENOSPC
```
ThomasWaldmann: just "no_space_left"?

borg/repository.py
```diff
@@ -84,6 +84,11 @@ def __enter__(self):
...
+            # The ENOSPC could have originated somewhere else besides the Repository. The cleanup is always safe, unless
+            # EIO or FS corruption ensues, which is why we specifically check for ENOSPC.
+            if self._active_txn and no_space_left_on_device:
+                self.io.cleanup(self.io.get_segments_transaction_id())
```
ThomasWaldmann: Hmm, do we emit a "no space left" log msg elsewhere and terminate with rc=2?

borg/repository.py
```diff
@@ -84,6 +84,11 @@ def __enter__(self):
...
             self.rollback()
```
ThomasWaldmann: i think we should log an error here also, like "no space left. rolling back uncommitted segments to free space."

enkore:
I wouldn't print an extra message for this, since this doesn't do anything in some cases¹ (e.g. ENOSPC when writing the index - the journal transaction is already complete, no space to be freed. Or ENOSPC when logging to a file or stuff like that; there is only a weak correlation between the ENOSPC __exit__ see's and the action it takes. As I wrote in the code, this is always safe except in few rare non-ENOSPC cases where the FS may be breaking down on us. Therefore the check for 1.0. We can try to make this the default behaviour for `Repository.rollback`, but not in a maintenance branch.

¹ and because it's an uninteresting internal detail.

ThomasWaldmann:
Well, it might be a little confusing if there is a "no disk space free" exception AND the user sees free disk space when looking afterwards. That's why I think it is a good idea to log that we tried to do something about it.

ThomasWaldmann:
Maybe add some words to the no-space-free section in the docs? + squash / merge

docs/quickstart.rst
```diff
@@ -11,11 +11,15 @@ The next section continues by showing how backups can be automated.
...
 
-Before you start creating backups, please make sure that there is **always**
+Before you start creating backups, please make sure that there is *always*
 a good amount of free space on the filesystem that has your backup repository
-(and also on ~/.cache). It is hard to tell how much, maybe 1-5%.
+(and also on ~/.cache). A few GB should suffice for most hard-drive sized
+repositories. See also :ref:`cache-memory-usage`.
 
-If you run out of disk space, it can be hard or impossible to free space,
+If |project_name| runs out of disk space, it tries to free as much space as it
```
ThomasWaldmann: ... out of disk space on the repository filesystem.

PR #1294, opened 2016-07-08
Fix missing return code check in ChunkIndex._add

enkore:
hashindex_set may resize the hashtable, failing to recognize the return code


ThomasWaldmann:
lgtm.

PR #1301, opened 2016-07-09
warn if a file with damaged chunks is extracted, fixes #1299

### LINKED ISSUE #1299: somehow warn when extracting a damaged file
borg check --repair replaces file chunks that are missing in the repo by all-zero chunks of same length.

but there currently is no indication when extracting such a file, it looks normally, no warning.

so, emit a warning? rename the file in the metadata, e.g. ".damaged"?

when a file is healed later, it should extract normally again.

ThomasWaldmann:
before 1.0.4/1.0.5 we had no means to tell if a file has been "repaired" with all-zero
chunks, but now we can see it if chunks_healthy key is present.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -18,7 +18,7 @@ import time
 from io import BytesIO
 from . import xattr
 from .helpers import Error, uid2user, user2uid, gid2group, group2gid, \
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
merge?

PR #1304, opened 2016-07-09
binary test issues

ThomasWaldmann:
this is not pretty, but I couldn't precisely find out the failure reason, but there is strong evidence that this is not a real bug in borg, but some side-effect of fakeroot usage.

silencing these always failing tests makes vagrant testing easier as "all succeeding" tests is easier to check than having always to check the same 3 failing tests.

### DIFF
--- a/borg/testsuite/archiver.py
+++ b/borg/testsuite/archiver.py
@@ -1108,6 +1108,14 @@ class ArchiverTestCaseBinary(ArchiverTestCase):
     EXE = 'borg.exe'
     FORK_DEFAULT = True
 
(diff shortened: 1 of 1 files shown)


PlasmaPower:
> there is strong evidence that this is not a real bug in borg, but some side-effect of fakeroot usage

I think you're misunderstanding. The problem is that fakeroot is masking the bug with it's custom in RAM fake FS implementation.

PlasmaPower:
Oh, I think I'm the one misunderstanding. Vagrant uses fakeroot. Don't you also use fakeroot for local tests though? I do and I've not had this problem.

PR #1306, opened 2016-07-10
docs issues

ThomasWaldmann:
solves most of the sphinx warnings.


borg/locking.py
```diff
@@ -223,8 +223,8 @@ class UpgradableLock:
     lock (multiple readers are allowed).
 
     If possible, try to use the contextmanager here like:
-    with UpgradableLock(...) as lock:
-        ...
+    with UpgradableLock(...) as lock: ...
+
```
enkore: I think these should be more like

```
like::

    with ... as ..:
        ...
```

to have it formatted as code.

ThomasWaldmann:
collapse fixup / merge?

enkore:
Yeah

PR #1319, opened 2016-07-12
Fixed the issue of absense of vagrant user on xenial64

pankajkgarg:
ubuntu/xenial64 has a default user configured by the name of "ubuntu" instead of "vagrant". This causes problems in running the box because Vagrantfile assumes that the user is "vagrant".

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,12 +12,19 @@ end
 
 def packages_debianoid
   return <<-EOF
+    if id "vagrant" >/dev/null 2>&1; then
+      username='vagrant'
+      home_dir=/home/vagrant
+    else
+      username='ubuntu'
+      home_dir=/home/ubuntu
+    fi
     apt-get update
     # install all the (security and other) updates
     apt-get dist-upgrade -y
     # for building borgbackup and dependencies:
     apt-get install -y libssl-dev libacl1-dev liblz4-dev libfuse-dev fuse pkg-config
-    usermod -a -G fuse vagrant
+    usermod -a -G fuse $username
     apt-get install -y fakeroot build-essential git
     apt-get install -y python3-dev python3-setuptools
     # for building python:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
I also stumbled over that when trying the xenial machines.

An alternative and IMHO better way could be to fix the root cause:
get the creator to make the xenial vagrant image like everybody else is doing vagrant images: using the user vagrant.

pankajkgarg:
Yeah, but in the meantime, this fixes the situation and this issue wasted a lot of my time, because I tried with xenial64 image in the beginning, as this is the latest version of Ubuntu.

ThomasWaldmann:
@pankajkgarg can you contact them? another quick solution might be introducing a "add vagrant user" step for non-complying images before all other steps - easier because we do not need that if/else at misc places.

pankajkgarg:
I have informed them about this issue.
>
>
> another quick solution might be introducing a "add vagrant user" step for
non-complying images before all other steps.
I began with that approach, but adding a separate user doesn't work.
Because after creating a vagrant user and changing user to it, the user
gets changed to default user (in this case 'ubuntu') in the beginning of
each script. And that would mean changes to all the scripts, which results
in a ugly mess.

I can create a separate packages_debianoid script for xenial, but
 'fix_perms' script still needs to use that if/else condition because there
is no way of determining which image is going to be used outside of that
particular image config scope.

Vagrantfile
```diff
@@ -12,12 +12,20 @@ end
 
 def packages_debianoid
   return <<-EOF
+    if id "vagrant" >/dev/null 2>&1; then
+      username='vagrant'
+      home_dir=/home/vagrant
+    else
+      #alternate would be to retrieve username by id 1000
```
ThomasWaldmann: remove the comment for now

pankajkgarg:
I have removed the comment and modified the commit.

ThomasWaldmann:
the PR should be against 1.0-maint (it will also get merged into master then), otherwise it is fine.

maybe check the xenial VM section, at least in master it still refers to trusty.

pankajkgarg:
xenial VM is not defined in 1.0-maint branch

ThomasWaldmann:
yes, because of the troubles you solved. as it is the current LTS, releases should be tested on it. :)

pankajkgarg:
Done! Created a new pull request to 1.0-maint branch.

ThomasWaldmann:
ok, thanks!

PR #1331, opened 2016-07-14
Fixed the issue of absense of vagrant user on xenial64

pankajkgarg:
A fix for absence of 'vagrant' user on ubuntu/xenial64  vm machine.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,12 +12,19 @@ end
 
 def packages_debianoid
   return <<-EOF
+    if id "vagrant" >/dev/null 2>&1; then
+      username='vagrant'
+      home_dir=/home/vagrant
+    else
+      username='ubuntu'
+      home_dir=/home/ubuntu
+    fi
     apt-get update
     # install all the (security and other) updates
     apt-get dist-upgrade -y
     # for building borgbackup and dependencies:
     apt-get install -y libssl-dev libacl1-dev liblz4-dev libfuse-dev fuse pkg-config
-    usermod -a -G fuse vagrant
+    usermod -a -G fuse $username
     apt-get install -y fakeroot build-essential git
     apt-get install -y python3-dev python3-setuptools
     # for building python:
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -338,6 +351,17 @@ Vagrant.configure(2) do |config|
     b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("centos6_64")
   end
 
+  config.vm.define "xenial64" do |b|
+    b.vm.box = "ubuntu/xenial64"
+    b.vm.provider :virtualbox do |v|
+      v.memory = 768
+    end
+    b.vm.provision "packages debianoid", :type => :shell, :inline => packages_debianoid
+    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_sys_venv("trusty64")
```
ThomasWaldmann: "trusty" (that's what I meant in my last comment) - see also below.

pankajkgarg:
Oh, my bad, completely missed that part. Its fixed now.

pankajkgarg:
Just one question, going ahead should I base my future commits from master branch or 1.0-maint ?

ThomasWaldmann:
@pankajkgarg it depends. if it is useful for both branches and is a fix or otherwise advised for 1.0-maint, it often should go there and then automatically gets merged into master also.

ThomasWaldmann:
lgtm, merge?

pankajkgarg:
yupp

PlasmaPower:
I'm running into a lot of problems with the `xenial64` VM, it seems like a mess. For instance, the home folder of "ubuntu" is set to `/root` for whatever reason. Also, since `/etc/hostname` doesn't exist `sudo` complains every time it's executed.

ThomasWaldmann:
@PlasmaPower maybe talk to the creator of the VM. If they would just create their VM as everybody else does, life would be easier. That also applies to using "vagrant" user, not creating a special case just by using "ubuntu".

enkore:
The bugtracker according to ATLAS is https://bugs.launchpad.net/cloud-images/+filebug/+login

PR #1333, opened 2016-07-14
Randomize files cache TTL

enkore:
This avoids* the "spiking backup time" effect, i.e. every 10th backup
takes much longer typically. This distributes this more evenly beyond
the 8th backup.

The mean TTL is bumped slightly to 12 instead of 10.

\* dampens

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -1,5 +1,6 @@
 import configparser
 import os
+import random
 import stat
 import shutil
 from binascii import unhexlify
@@ -217,7 +218,8 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
                     # Discard cached files with the newest mtime to avoid
                     # issues with filesystem snapshots and mtime precision
                     entry = FileCacheEntry(*msgpack.unpackb(item))
-                    if entry.age < 10 and bigint_to_int(entry.mtime) < self._newest_mtime:
+                    age_limit = random.randint(8, 16)
+                    if entry.age < age_limit and bigint_to_int(entry.mtime) < self._newest_mtime:
                         msgpack.pack((path_hash, entry), fd)
         self.config.set('cache', 'manifest', self.manifest.id_str)
         self.config.set('cache', 'timestamp', self.manifest.timestamp)


ThomasWaldmann:
did you see these "spikes" in practice?

enkore:
It depends on how the dataset evolved. If lots of files don't change, then they only get backed up when the TTL runs out, so every 10th backup starting with their initial inclusion, which then is usually the first backup.

Since I rotate a few disks it takes a while to see it, but it's there. These are from the same disk:

```
$ cat logs/2016-03-16.log | grep Duration
Duration: 2 minutes 8.63 seconds
Duration: 5 minutes 5.01 seconds
Duration: 8 minutes 13.47 seconds
$ cat logs/2016-03-18.log | grep Duration
Duration: 19 minutes 21.74 seconds
Duration: 4 hours 30 minutes 10.95 seconds <--!
Duration: 2 hours 22 minutes 51.21 seconds
$ cat logs/2016-04-30.log:
Duration: 2 minutes 25.36 seconds
Duration: 21 minutes 4.47 seconds
Duration: 14 minutes 18.83 seconds
$ cat logs/2016-06-09.log | grep Duration
Duration: 10 minutes 54.01 seconds
Duration: 2 hours 3 minutes 45.05 seconds <--!
Duration: 1 hours 5 minutes 46.84 seconds
$ cat logs/2016-06-11.log | grep Duration
Duration: 2 minutes 11.09 seconds
Duration: 7 minutes 51.20 seconds
Duration: 12 minutes 59.87 seconds
```

I would expect that this patch flattens these "bumps" down considerably.

ThomasWaldmann:
You've seen cache.py:429?
```
        if (entry.size == st.st_size and bigint_to_int(entry.mtime) == st.st_mtime_ns and
                (ignore_inode or entry.inode == st.st_ino)):
            self.files[path_hash] = msgpack.packb(entry._replace(age=0)
```

enkore:
Hmm, this is very strange. I'm going to check more logs here.

ThomasWaldmann:
I wondered already about that code some months ago. ;-)

BTW, a total backup time spoiler is if the latest mtime happens to be a VM image or some other large file, as it is always considered to be potentially modified and thus is chunked.

PR #1334, opened 2016-07-14
faq: 'A' unchanged file; remove ambiguous entry age sentence.

enkore:


ThomasWaldmann:
I would not just remove the other case. I you e.g. do 11 different backups of parts of the same filesystem, you may run into that case also and wonder what's happening.

Issue #1273, opened 2016-07-05
Add a per repo setting of minimum free space to preserve


textshell:
If #1272 is implemented it would be nice to have a setting how much additional space borg should leave free, so on a system with other services these have some minimal amount of free space left to keep working.

Also some filesystems are known to get into badly tested code paths on full filesystems, so that’s another reason to avoid filling the filesystem to much.

enkore:
#1336 merged

Issue #721, opened 2016-03-08
Add user configurable check for remaining diskspace.


manwegit:
I don't know any backup software that does this, but this is a feature that I routinely have to workaround. 

Use case:
Backup target is used by many systems and space usage varies day to day greatly.
Integrity for backup is more important than having finished backup for that particular set.
Currently my automated scripts make disk space checks for backup target before backup starts and report errors / halt backup if 2500 MiB or less space is available. When using remote target this means that I have to make an extra ssh step + checking correct directory for space is PITA. 
Having controlled option in borg to not to do backup if target has less than said value would be nice.

I think limit before backup is enough as everyone needs to adjust the value according what kind of backup they are running, so 2.5 GiB for dedup growth is ok for me.

(If interest for this kind of feature I'll be willing to spend time for a patch + evaluation of said patch)

ThomasWaldmann:
I understand your point, but this is somehow something to be solved outside of borg.

Or how many tools do you know which have builtin free-space checks?

manwegit:
Well that was part of my point, that I don't know backup software that does it. And that missing feature has caused me problems with for example rdiff-backup.

This is a feature that I'd find useful. It might be a feature that others would also appreciate. As such I'd say this is low priority enhancement request.

This missing feature is kind of problem when running backups with authorized_keys lines. You basically don't have any other interface to you backup target at the beginning of backup other than borg. Because remote serve cmd line does not have target folder as a parameter it's not that easy to script with a wrapper.

Another more generic way would be ability to add target path or custom parameters to be passed borg serve. `$SSH_ORIGINAL_COMMAND` would have all options passed that a wrapper script could do what ever with the commands.
Actually I would like this more as it is more generic and leaves diskspace check etc to admin.
(I've done something similar with rsync and modified rrsync script)

The main reason I'd like to have it in borg is that low disk space would not screw up the repo, when there's a minimum buffer requirement for free space (kind of like ext-fs reserved root space).

jungle-boogie:
A noble concern but I'd call this a feature creep.

manwegit:
Adding target to remote serve as cmdline option even if it's not used by borg (so authorized_keys) works, would basically solve all my remote serve problems.

ThomasWaldmann:
Hmm, the (recommended) "borg serve" restriction via authorized_keys is of course in the way if you want to execute something else.

So I guess there are only 2 ways to solve:
[...]

PR #1336, opened 2016-07-14
Check free space before committing

### LINKED ISSUE #1272: Check free space on repository transaction commit
To avoid leaving the filesystem the repository resides on without sufficient free space, borg should check the free space just before commiting a transaction. If the free space is to low an transaction abort will leave the repository in a safe state while preserving a needed free space margin on the file system.

The needed amount of storage for the commit (not counting space freed by compaction) should be fairly easy to calculate.

Thomas suggested to additionally keep 4x index size of free space. Also compaction will need space. I *think* up to the maximum segment size.

enkore:
A failed check looks like this:

> Insufficient free space to complete transaction (required: 1.10 TB, available: 47.58 GB).
> zsh: exit 2     borg create ...

Also logs some --debug output:

> check_free_space: required bytes 26549600, free bytes 47584997376

(in this case: 5 MiB + 20 MiB + 4× some dozen KiB index size)

The check uses f_bavail, which does *not* include root's block reserve. Therefore the block reserve is not written full, even when running as root (yay). For the same reason additional_free_space is around. It can also be used to defeat the check (with a negative value) in desperate situations.

Fixes #1272 #1273

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -441,7 +441,13 @@ hashindex_next_key(HashIndex *index, const void *key)
 }
 
 static int
-hashindex_get_size(HashIndex *index)
+hashindex_len(HashIndex *index)
 {
     return index->num_entries;
+}
+
+static int
+hashindex_size(HashIndex *index)
+{
+    return sizeof(HashHeader) + index->num_buckets * index->bucket_size;
 }
--- a/src/borg/hashindex.pyx
+++ b/src/borg/hashindex.pyx
@@ -8,7 +8,7 @@ from libc.stdint cimport uint32_t, UINT32_MAX, uint64_t
 from libc.errno cimport errno
 from cpython.exc cimport PyErr_SetFromErrnoWithFilename
 
-API_VERSION = 2
+API_VERSION = 3
 
 
 cdef extern from "_hashindex.c":
@@ -18,7 +18,8 @@ cdef extern from "_hashindex.c":
     HashIndex *hashindex_read(char *path)
     HashIndex *hashindex_init(int capacity, int key_size, int value_size)
     void hashindex_free(HashIndex *index)
-    int hashindex_get_size(HashIndex *index)
+    int hashindex_len(HashIndex *index)
+    int hashindex_size(HashIndex *index)
     int hashindex_write(HashIndex *index, char *path)
     void *hashindex_get(HashIndex *index, void *key)
     void *hashindex_next_key(HashIndex *index, void *key)
@@ -119,7 +120,11 @@ cdef class IndexBase:
             raise
 
     def __len__(self):
-        return hashindex_get_size(self.index)
+        return hashindex_len(self.index)
+
+    def size(self):
+        """Return size (bytes) of hash table."""
+        return hashindex_size(self.index)
 
 
 cdef class NSIndex(IndexBase):
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -248,9 +248,10 @@ class Repository:
     def commit(self, save_space=False):
         """Commit transaction
         """
+        # save_space is not used anymore, but stays for RPC/API compatibility.
         self.io.write_commit()
         if not self.append_only:
-            self.compact_segments(save_space=save_space)
+            self.compact_segments()
         self.write_index()
         self.rollback()
 
@@ -348,7 +349,7 @@ class Repository:
             os.unlink(os.path.join(self.path, name))
         self.index = None
 
-    def compact_segments(self, save_space=False):
+    def compact_segments(self):
         """Compact sparse segments by copying data into new segments
         """
         if not self.compact:
@@ -357,12 +358,11 @@ class Repository:
         segments = self.segments
         unused = []  # list of segments, that are not used anymore
 
-        def complete_xfer():
-            # complete the transfer (usually exactly when some target segment
-            # is full, or at the very end when everything is processed)
+        def complete_xfer(intermediate=True):
+            # complete the current transfer (when some target segment is full)
             nonlocal unused
             # commit the new, compact, used segments
-            self.io.write_commit()
+            self.io.write_commit(intermediate=intermediate)
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
                 assert self.segments.pop(segment) == 0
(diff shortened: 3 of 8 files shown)


src/borg/repository.py
```diff
@@ -344,6 +348,28 @@ def write_index(self):
             os.unlink(os.path.join(self.path, name))
         self.index = None
 
+    def check_free_space(self):
+        """Pre-commit check for sufficient free space to actually perform the commit."""
+        # Keep four times the (compact - not necessarily the case) index size free
```
ThomasWaldmann: hmm, iirc i meant 4x the on-disk size.

src/borg/repository.py
```diff
@@ -344,6 +348,28 @@ def write_index(self):
...
+        index_size = len(self.index) * 40 * 4
+        required_free_space = index_size
```
ThomasWaldmann: also, it is rather:
index_size = ...
required_free_space = 4 * index_size

src/borg/repository.py
enkore: Yeah, I guess it'd be better to query the current hashtable size *and* apply a 4× factor here, instead of living with the 33-300 % variance in hashtable size vs. len(hashtable).

src/borg/repository.py
```diff
@@ -344,6 +348,28 @@ def write_index(self):
...
+            # Keep one worst-case full free in non-append-only mode
+            required_free_space += self.max_segment_size + MAX_OBJECT_SIZE
+        try:
+            st_vfs = os.statvfs(self.path)
+        except OSError as os_error:
+            logger.warning('Failed to check free space before committing: ' + str(os_error))
+            return
+        # f_bavail: even as root - don't touch the Federal Block Reserve!
+        free_space = st_vfs.f_bavail * st_vfs.f_bsize
+        logger.debug('check_free_space: required bytes {}, free bytes {}'.format(required_free_space, free_space))
```
ThomasWaldmann: iirc we have a format_size() or so in helpers.

src/borg/repository.py
```diff
@@ -344,6 +348,28 @@ def write_index(self):
...
+    def check_free_space(self):
+        """Pre-commit check for sufficient free space to actually perform the commit."""
+        # Keep four times the (compact - not necessarily the case) index size free
+        index_size = len(self.index) * 40 * 4
+        required_free_space = index_size
+        if not self.append_only:
+            # Keep one worst-case full free in non-append-only mode
+            required_free_space += self.max_segment_size + MAX_OBJECT_SIZE
+        try:
+            st_vfs = os.statvfs(self.path)
```
ThomasWaldmann: if someone mounts a filesystem into repo/data ...

enkore: much easier way to defeat the check is to just run "fallocate" at the right time. there's no way this can be bullet-proof - or even come close to that -, but we can catch the usual case: no strange mount setups, no overlayfs stuff, an FS actually reporting up-to-date numbers, no quotas and (mainly) one writer (our process) to the FS.

src/borg/repository.py
```diff
@@ -353,7 +357,7 @@ def write_index(self):
     def check_free_space(self):
         """Pre-commit check for sufficient free space to actually perform the commit."""
         required_free_space = self.index.size_on_disk() * 4
-        # TODO: hints file size
+        required_free_space += len(msgpack.packb(self.hints()))
```
enkore: probably better to just make a conservative estimate. this copies the entire structure just to get the length.

src/borg/_hashindex.c
```diff
@@ -445,3 +445,9 @@ hashindex_get_size(HashIndex *index)
 {
     return index->num_entries;
 }
+
+static int
+hashindex_get_disk_size(HashIndex *index)
```
ThomasWaldmann: the "get disk size" part sounds a bit strange.
how about just "get size" or "size"?

src/borg/testsuite/hashindex.py
```diff
@@ -121,6 +121,19 @@ def test_chunkindex_summarize(self):
...
+        assert idx.size_on_disk() == 18 + 1031 * (32 + 3 * 4)
+
+    def test_size_on_disk_accurate(self):
+        idx = ChunkIndex()
+        for i in range(1234):
+            idx[H(i)] = i, i**2, i**3
+        with tempfile.NamedTemporaryFile() as file:
+            idx.write(file.name)
+            size = os.path.getsize(file.name)
+        assert idx.size_on_disk() == size
```
ThomasWaldmann: not sure if we want to do that on every borg startup.

enkore: true

src/borg/_hashindex.c
enkore: hashindex_get_size -> hashindex_num_entries

hashindx_get_disk_size -> hashindex_size

src/borg/repository.py
```diff
@@ -343,7 +350,35 @@ def write_index(self):
             os.unlink(os.path.join(self.path, name))
         self.index = None
 
-    def compact_segments(self, save_space=False):
+    def check_free_space(self):
+        """Pre-commit check for sufficient free space to actually perform the commit."""
+        required_free_space = self.index.size() * 4
```
ThomasWaldmann: maybe add the reasoning for 4x also, see ticket.

src/borg/hashindex.pyx
```diff
@@ -119,7 +120,11 @@ cdef class IndexBase:
             raise
 
     def __len__(self):
-        return hashindex_get_size(self.index)
+        return hashindex_len(self.index)
```
ThomasWaldmann: """Return count of used buckets in the hash table."""

ThomasWaldmann:
maybe collapse / amend away the non-ascii commit comment. :)

enkore:
TODO: testing.

enkore:
Stapling #721

enkore:
lgtm. I didn't re-test performance; I don't see c393862 making a big difference, though it will cost a bit.

ThomasWaldmann:
maybe add `__len__` docstring also, merge.

src/borg/hashindex.pyx
enkore: Imho thats less clear than the implied contract of `len(dict-like)`.

enkore:
(rebased)

PR #1343, opened 2016-07-17
windows: merge master

enkore:
TODO: test/fix issues on windows.

Current issue(s):

- uid/gid are SID strings on windows, not integers. --- Better use different Item attribute for that, they're meaningless on !Windows anyway. user_sid/group_sid.


enkore:
+3 failures compared to 7f68acf.

PR #1357, opened 2016-07-22
Add --append-only to borg init

### LINKED ISSUE #1291: add --append-only to 'borg init'
It's a small nit, but I'm automating some borg repo creations and it would be great if append-only mode could be specified at init time instead of having to automate editing the config file.

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/35866629-add-append-only-to-borg-init)

PlasmaPower:
Fixes #1291

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -73,13 +73,16 @@ def with_repository(fake=False, create=False, lock=True, exclusive=False, manife
         @functools.wraps(method)
         def wrapper(self, args, **kwargs):
             location = args.location  # note: 'location' must be always present in args
+            append_only = getattr(args, 'append_only', False)
             if argument(args, fake):
                 return method(self, args, repository=None, **kwargs)
             elif location.proto == 'ssh':
-                repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
+                repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock,
+                                              append_only=append_only, args=args)
             else:
                 repository = Repository(location.path, create=create, exclusive=argument(args, exclusive),
-                                        lock_wait=self.lock_wait, lock=lock)
+                                        lock_wait=self.lock_wait, lock=lock,
+                                        append_only=append_only)
             with repository:
                 if manifest or cache:
                     kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
@@ -1304,6 +1307,8 @@ class Archiver:
         subparser.add_argument('-e', '--encryption', dest='encryption',
                                choices=('none', 'keyfile', 'repokey'), default='repokey',
                                help='select encryption key mode (default: "%(default)s")')
+        subparser.add_argument('-a', '--append-only', dest='append_only', action='store_true',
+                               help='create an append-only mode repository')
 
         check_epilog = textwrap.dedent("""
         The check command verifies the consistency of a repository and the corresponding archives.
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -117,7 +117,7 @@ class RepositoryServer:  # pragma: no cover
     def negotiate(self, versions):
         return RPC_PROTOCOL_VERSION
 
-    def open(self, path, create=False, lock_wait=None, lock=True):
+    def open(self, path, create=False, lock_wait=None, lock=True, append_only=False):
         path = os.fsdecode(path)
         if path.startswith('/~'):
             path = os.path.join(get_home_dir(), path[2:])
@@ -128,7 +128,7 @@ class RepositoryServer:  # pragma: no cover
                     break
             else:
                 raise PathNotAllowed(path)
-        self.repository = Repository(path, create, lock_wait=lock_wait, lock=lock, append_only=self.append_only)
+        self.repository = Repository(path, create, lock_wait=lock_wait, lock=lock, append_only=self.append_only or append_only)
         self.repository.__enter__()  # clean exit handled by serve() method
         return self.repository.id
 
@@ -137,10 +137,14 @@ class RemoteRepository:
     extra_test_args = []
 
     class RPCError(Exception):
-        def __init__(self, name):
+        def __init__(self, name, remote_type):
             self.name = name
+            self.remote_type = remote_type
 
-    def __init__(self, location, create=False, lock_wait=None, lock=True, args=None):
+    class NoAppendOnlyOnServer(Error):
+        """Server does not support --append-only."""
+
+    def __init__(self, location, create=False, lock_wait=None, lock=True, append_only=False, args=None):
         self.location = self._location = location
         self.preload_ids = []
         self.msgid = 0
@@ -176,7 +180,17 @@ class RemoteRepository:
         if version != RPC_PROTOCOL_VERSION:
             raise Exception('Server insisted on using unsupported protocol version %d' % version)
         try:
-            self.id = self.call('open', self.location.path, create, lock_wait, lock)
+            # Because of protocol versions, only send append_only if necessary
+            if append_only:
+                try:
+                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
+                except self.RPCError as err:
+                    if err.remote_type == 'TypeError':
+                        raise self.NoAppendOnlyOnServer() from err
+                    else:
+                        raise
+            else:
+                self.id = self.call('open', self.location.path, create, lock_wait, lock)
         except Exception:
             self.close()
             raise
@@ -272,7 +286,7 @@ class RemoteRepository:
             elif error == b'InvalidRPCMethod':
                 raise InvalidRPCMethod(*res)
             else:
-                raise self.RPCError(res.decode('utf-8'))
+                raise self.RPCError(res.decode('utf-8'), error.decode('utf-8'))
 
         calls = list(calls)
         waiting_for = []
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -154,7 +154,7 @@ class Repository:
         config.set('repository', 'version', '1')
         config.set('repository', 'segments_per_dir', str(DEFAULT_SEGMENTS_PER_DIR))
         config.set('repository', 'max_segment_size', str(DEFAULT_MAX_SEGMENT_SIZE))
-        config.set('repository', 'append_only', '0')
+        config.set('repository', 'append_only', str(int(self.append_only)))
         config.set('repository', 'id', bin_to_hex(os.urandom(32)))
         self.save_config(path, config)
 


src/borg/repository.py
```diff
@@ -154,7 +154,7 @@ def create(self, path):
         config.set('repository', 'version', '1')
         config.set('repository', 'segments_per_dir', str(DEFAULT_SEGMENTS_PER_DIR))
         config.set('repository', 'max_segment_size', str(DEFAULT_MAX_SEGMENT_SIZE))
-        config.set('repository', 'append_only', '0')
+        config.set('repository', 'append_only', str(int(self.append_only)))
```
PlasmaPower: I'm not quite sure if this is the way I should be doing this, but I think it is.

src/borg/archiver.py
```diff
@@ -78,8 +78,10 @@ def wrapper(self, args, **kwargs):
             elif location.proto == 'ssh':
                 repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
             else:
+                if(not hasattr(args, 'append_only')): # most don't have this argument
+                    args.append_only = False
                 repository = Repository(location.path, create=create, exclusive=argument(args, exclusive),
-                                        lock_wait=self.lock_wait, lock=lock)
+                                        lock_wait=self.lock_wait, lock=lock, append_only=args.append_only)
```
ThomasWaldmann: easier / shorter: append_only=getattr(args, 'append_only', False)

src/borg/archiver.py
```diff
@@ -78,8 +78,10 @@ def wrapper(self, args, **kwargs):
             elif location.proto == 'ssh':
                 repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
```
ThomasWaldmann: did you check the remote code also?

if it needs a RPC interface change, leave a comment there, but do not do incompatible RPC interface changes.

src/borg/repository.py
ThomasWaldmann: seems reasonable.

ThomasWaldmann:
@PlasmaPower btw, there is a bounty on that ticket.

src/borg/archiver.py
PlasmaPower: The client currently has no notion of append_only, and I don't think we could implement it without the server's assistance. I'll just add a comment then.

PlasmaPower:
@ThomasWaldmann I noticed :smile:

I think it should be less for some of these easier ones like this, maybe $5 or $10, and put that saved money into the more difficult ones.

src/borg/archiver.py
```diff
@@ -76,10 +76,14 @@ def wrapper(self, args, **kwargs):
             if argument(args, fake):
                 return method(self, args, repository=None, **kwargs)
             elif location.proto == 'ssh':
+                # RemoteRepository has no notion of append_only
+                # It can't be implemented without cooperation with the server
+                # We don't want to break the RPC API though
```
ThomasWaldmann: well, could you prepare everything so it is in place, except actually breaking the RPC api?

PlasmaPower: Wait, would an optional argument break the API? If not, I think I can do this.

PlasmaPower: Oops, of course not (new client -> old server breaks). I could do this without breakage if I messed with the negotiation though, is that fine?

PlasmaPower: Wait, what am I even doing, RemoteServer doesn't even support init!

ThomasWaldmann: open() has a create flag.

PlasmaPower: Oh. I missed that. In that case, mind if I mess with the negotiation to add a downgradable protocol version?

PlasmaPower:
@ThomasWaldmann That negotiation logic look good? IUntested, but I'm working on that now.

PlasmaPower:
Okay I've tested that logic across various client and server borg versions and it works great. Look good? Should the negotiation have automated tests, and if so where should I put them? The logic is very simple but also very important.

src/borg/remote.py
```diff
@@ -173,10 +174,15 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
...
-        if version != RPC_PROTOCOL_VERSION:
+        if version < RPC_PROTOCOL_DOWNGRADABLE_VERSION or version > RPC_PROTOCOL_VERSION:
             raise Exception('Server insisted on using unsupported protocol version %d' % version)
         try:
-            self.id = self.call('open', self.location.path, create, lock_wait, lock)
+            if version >= 3:
+                self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
+            else:
+                if append_only:
+                    raise Exception('Server (protocol version %d) does not support --append-only' % version)
```
PlasmaPower: Raising an Exception means that it generates a very ugly error message. Do we have an alternative? The same thing already exists above for an incompatible version.

src/borg/remote.py
```diff
@@ -17,7 +17,8 @@
 from .helpers import bin_to_hex
 from .repository import Repository
 
-RPC_PROTOCOL_VERSION = 2
+RPC_PROTOCOL_VERSION = 3
+RPC_PROTOCOL_DOWNGRADABLE_VERSION = 2
```
ThomasWaldmann: use a list?

PlasmaPower: I'm assuming you mean a tuple of `(min, max)`? Sure.

src/borg/remote.py
```diff
@@ -173,10 +174,15 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
             version = self.call('negotiate', RPC_PROTOCOL_VERSION)
         except ConnectionClosed:
             raise ConnectionClosedWithHint('Is borg working on the server?') from None
-        if version != RPC_PROTOCOL_VERSION:
+        if version < RPC_PROTOCOL_DOWNGRADABLE_VERSION or version > RPC_PROTOCOL_VERSION:
```
ThomasWaldmann: if version not in (supported set / list)

PlasmaPower: Would there be a situation where, say, 3 and 5 are supported but not 4?

src/borg/remote.py
ThomasWaldmann: no, i meant a set or list. :)

src/borg/remote.py
ThomasWaldmann: maybe not, but we can nevertheless code it in a generic way.

src/borg/remote.py
PlasmaPower: That's harder because the client only supports a single supported version. If that version isn't in the supported set which item in the list do we send back?

PlasmaPower: This might be crazy, but what if we have my version of the negotiate function first and then if that succeeds with `version >= 3` we go to `list_negotiate` which is basically your version but the client sends a list?

enkore:
Couldn't it try append_only=True, if set, and if it doesn't work, ie. TypeError comes back then we just fail with a message - remote borg server doesn't know --append-only

src/borg/remote.py
ThomasWaldmann: maybe rather something simple for now.
we did not want to carry around too much heritage with borg anyway.

PlasmaPower: So do you want a `(min, max)` or a list? If the latter which item do we pull if the given version isn't supported?

PlasmaPower:
@enkore Basically only send append_only if it's set? That's actually a great idea, let me do some testing with that.

PlasmaPower:
Okay, I'm done with all the protocol negotiation stuff due to @enkore's suggestion (thanks!). Only problem is that the remote repo controls the TypeError so I can't catch it. Should I add a note if any exception is generated with append_only set?

enkore:
You should get an RPCError with the name attribute = "TypeError". That condition should re-raise a more specific `Error` subclass in (just this) path.

PlasmaPower:
@enkore The name attribute is `"Remote Exception (see remote log for the traceback)"`.

enkore:
Hm, that looks like an oversight in `handle_error(error, res)`. RPCError should have two attributes: name, the actual exception name, and maybe message or something like that. Changing that needs some minor adjustment here as well -> https://github.com/borgbackup/borg/blob/master/src/borg/archiver.py#L2275

PlasmaPower:
Okay, It will now output "Exception: Server does not support --append-only" but there's a lot of traceback junk around it. Is that fine, or is there something I can do to fix that?

enkore:
Make a new exception type e.g.

```py
class NoAppendOnlyOnServer(Error):
    """Server does not support --append-only."""
```

If you then raise NoAppendOnlyOnServer only that message will be printed and no tracebacks :)

src/borg/remote.py
```diff
@@ -137,10 +137,11 @@ class RemoteRepository:
     extra_test_args = []
 
     class RPCError(Exception):
-        def __init__(self, name):
+        def __init__(self, name, rootCause):
             self.name = name
+            self.rootCause = rootCause
```
enkore: pep8 names please, ie. `root_cause` (or maybe something more descriptive? How about `remote_[exception_]type` or something like that?

src/borg/remote.py
```diff
@@ -176,7 +177,17 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
...
-            self.id = self.call('open', self.location.path, create, lock_wait, lock)
+            # Because of protocol versions, only send append_only if necessary
+            if append_only:
+                try:
+                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
+                except self.RPCError as err:
+                    if err.rootCause == 'TypeError':
+                        raise Exception('Server does not support --append-only') from err
+                    else:
+                        raise err
```
enkore: Just `raise` is enough - it will re-raise the currently handled exception (here in `err`) if no exception is specified. You can also leave the else: away if you like, since the if: block raises, when executed.

PlasmaPower:
Great! It still prints the `Remote:` error message in addition to the NoAppendOnlyOnServer error, but that's fine.

src/borg/archiver.py
```diff
@@ -76,10 +76,12 @@ def wrapper(self, args, **kwargs):
...
                 return method(self, args, repository=None, **kwargs)
             elif location.proto == 'ssh':
-                repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock, args=args)
+                repository = RemoteRepository(location, create=create, lock_wait=self.lock_wait, lock=lock,
+                                              append_only=getattr(args, 'append_only', False), args=args)
             else:
                 repository = Repository(location.path, create=create, exclusive=argument(args, exclusive),
-                                        lock_wait=self.lock_wait, lock=lock)
+                                        lock_wait=self.lock_wait, lock=lock,
+                                        append_only=getattr(args, 'append_only', False))
```
ThomasWaldmann: maybe move append_only=getattr(...) to before the if-block so you don't need to duplicate it.

src/borg/archiver.py
```diff
@@ -1304,6 +1306,8 @@ def build_parser(self, prog=None):
         subparser.add_argument('-e', '--encryption', dest='encryption',
                                choices=('none', 'keyfile', 'repokey'), default='repokey',
                                help='select encryption key mode (default: "%(default)s")')
+        subparser.add_argument('-a', '--append-only', dest='append_only', action='store_true',
+                               help='only allow appending to repository segment files')
```
ThomasWaldmann: not sure if it is better, but we could also write "create repository in append-only mode", because this is the direct effect of this. what you have is the indirect effect what this will have for later operations.

src/borg/remote.py
```diff
@@ -176,7 +180,17 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
...
+            if append_only:
+                try:
+                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
+                except self.RPCError as err:
+                    if err.remote_type == 'TypeError':
+                        raise self.NoAppendOnlyOnServer() from err
+                    else:
+                        raise
+            else:
+                self.id = self.call('open', self.location.path, create, lock_wait, lock)
```
ThomasWaldmann: interesting. but as we still have everything positional, this only works for last param, right?

PlasmaPower: Yes, but it can also work for any number of last params. The number of parameters supported basically represents the `open` RPC API version number.

PlasmaPower:
@ThomasWaldmann Your comments have been fixed BTW.

src/borg/remote.py
```diff
@@ -176,7 +180,17 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, args=None)
         if version != RPC_PROTOCOL_VERSION:
             raise Exception('Server insisted on using unsupported protocol version %d' % version)
         try:
-            self.id = self.call('open', self.location.path, create, lock_wait, lock)
+            # Because of protocol versions, only send append_only if necessary
+            if append_only:
+                try:
+                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
+                except self.RPCError as err:
+                    if err.remote_type == 'TypeError':
```
ThomasWaldmann: could a TypeError happen under different circumstances than what we expect here?

PlasmaPower: Maybe, but I think that if `--append-only` is passed a TypeError on open is extremely likely to be due to this. In addition, to do any more analysis, we'd have to parse the error string.

ThomasWaldmann:
PR against 1.0-maint?

PlasmaPower:
Replaced by #1386 

In the future I'll be sure to PR non-testing PRs against 1.0-maint.

ThomasWaldmann:
@PlasmaPower it's often a not easy decision. but as we already have the append-only feature in 1.0, adding a commandline switch for it seems to be appropriate for there.

PlasmaPower:
@ThomasWaldmann Okay, makes sense.

PR #1358, opened 2016-07-22
Ignore generated platform C files

PlasmaPower:
Seems like every time I make a PR I notice some other autogenerated files that aren't gitignored. At least I didn't commit them this time :smile:.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,10 +8,10 @@ hashindex.c
 chunker.c
 compress.c
 crypto.c
-platform_darwin.c
-platform_freebsd.c
-platform_linux.c
-platform_posix.c
+src/borg/platform/darwin.c
+src/borg/platform/freebsd.c
+src/borg/platform/linux.c
+src/borg/platform/posix.c
 *.egg-info
 *.pyc
 *.pyo


ThomasWaldmann:
guess you could just modify these, they just got moved:
```
 platform_darwin.c
 platform_freebsd.c
 platform_linux.c
 platform_posix.c
```

PlasmaPower:
@ThomasWaldmann Are you saying commit those files? Or hard code them?

ThomasWaldmann:
The files platform_*.c got moved to the platform/ directory, so the .gitignore just needs to get fixed.

PlasmaPower:
Oh, I see. Would a more general `src/borg/platform/platform_*.c` check be better?

PlasmaPower:
Actually, nope. It was moved from `platform_*.c` to `src/borg/platform/*.c`

PlasmaPower:
How about this change? I don't see a reason to hard code the OSs.

ThomasWaldmann:
Yes, that will work as long noone adds hand-edited *.c files there.

PlasmaPower:
Okay, I've hard coded them.

PR #1359, opened 2016-07-22
Test multiple filesystems with Vagrant

### LINKED ISSUE #1289: Test on many file systems automatically
ext4/xfs or tmpfs (which one I am not sure - check that) is already done by Vagrant. We should also test **sshfs**, **cifs/smbfs**, **nfs**, **ntfs-3g** to avoid issues like #1287.

This could be done within the existing wheezy Vagrant machines that also builds the borg binaries.
So we could test from source as well as the freshly built binaries. `testsuite.archiver` is already made so that both binaries and src are covered.

The FS needs to be setup and mounted (for NTFS, use a loop device), then the test suite tempdir ( http://pytest.org/latest/tmpdir.html#base-temporary-directory ) has to point into the FS during a test suite run. This has to be done for each FS.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/35863167-test-on-many-file-systems-automatically)

PlasmaPower:
Fixes #1289 

Both kernel module and FUSE filesystems are supported.

TODO:
- Filesystems requiring a device
  - [x] ext4
  - [x] XFS
  - [x] NTFS
- Filesystems requiring a server
  - [x] CIFS
  - [x] SSHFS
- Filesystems requiring neither
  - [x] tmpfs
- Fix FS specific test issues
  - [x] #1361 - test_non_ascii_acl only failed outside of Vagrant due to übel user missing
  - [x] #1362 - tests failing on SSHFS (fixed with #1373)
  - [x] #1363 - tests failing on ntfs-3g (fixed with #1373 + #1383)
- Platform specific Vagrant testing
  - Linux
    - [x] centos7_64
    - [ ] centos6_32
    - [ ] centos6_64
    - [x] xenial64
    - [x] trusty64
    - [ ] jessie64
    - [ ] wheezy32
    - [ ] wheezy64
  - OS X
    - [x] darwin64
  - BSD
    - [ ] freebsd64
    - [ ] openbsd64
    - [ ] netbsd64

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -29,6 +29,8 @@ def packages_debianoid
     apt-get install -y python3-dev python3-setuptools
     # for building python:
     apt-get install -y zlib1g-dev libbz2-dev libncurses5-dev libreadline-dev liblzma-dev libsqlite3-dev
+    # filesystem drivers
+    apt-get install -y ntfs-3g openssh-server sshfs xfsprogs samba cifs-utils
     # this way it works on older dists (like ubuntu 12.04) also:
     # for python 3.2 on ubuntu 12.04 we need pip<8 and virtualenv<14 as
     # newer versions are not compatible with py 3.2 any more.
@@ -50,6 +52,8 @@ def packages_redhatted
     yum install -y gcc-c++
     # for building python:
     yum install -y zlib-devel bzip2-devel ncurses-devel readline-devel xz xz-devel sqlite-devel
+    # filesystem drivers
+    yum install -y ntfs-3g openssh fuse-sshfs xfsprogs samba cifs-utils ntfsprogs
     #yum install -y python-pip
     #pip install virtualenv
     touch ~vagrant/.bash_profile ; chown vagrant ~vagrant/.bash_profile
@@ -86,6 +90,8 @@ def packages_freebsd
     pkg install -y fakeroot git bash
     # for building python:
     pkg install -y sqlite3
+    # filesystem drivers
+    pkg install -y openssh sshfs samba ntfs-3g
     # make bash default / work:
     chsh -s bash vagrant
     mount -t fdescfs fdesc /dev/fd
@@ -141,6 +147,7 @@ def packages_netbsd
     pkg_add pkg-config  # avoids some "pkg-config missing" error msg, even without fuse
     # pkg_add fuse  # llfuse supports netbsd, but is still buggy.
     # https://bitbucket.org/nikratio/python-llfuse/issues/70/perfuse_open-setsockopt-no-buffer-space
+    pkg_add samba
     pkg_add python34 py34-setuptools
     ln -s /usr/pkg/bin/python3.4 /usr/pkg/bin/python
     ln -s /usr/pkg/bin/python3.4 /usr/pkg/bin/python3
@@ -270,23 +277,7 @@ def build_binary_with_pyinstaller(boxname)
 end
 
 def run_tests(boxname)
-  return <<-EOF
-    . ~/.bash_profile
-    cd /vagrant/borg/borg
-    . ../borg-env/bin/activate
-    if which pyenv > /dev/null; then
-      # for testing, use the earliest point releases of the supported python versions:
-      pyenv global 3.4.0 3.5.0
-    fi
-    # otherwise: just use the system python
-    if which fakeroot > /dev/null; then
-      echo "Running tox WITH fakeroot -u"
-      fakeroot -u tox --skip-missing-interpreters
-    else
-      echo "Running tox WITHOUT fakeroot -u"
-      tox --skip-missing-interpreters
-    fi
-  EOF
+  return "source /vagrant/borg/borg/vagrant-tools/run-tests.sh"
 end
 
 def fix_perms
@@ -309,7 +300,7 @@ Vagrant.configure(2) do |config|
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
   # fix permissions on synced folder
-  config.vm.provision "fix perms", :type => :shell, :inline => fix_perms
+  config.vm.provision "fix perms",                  :type => :shell, :privileged => true,  :inline => fix_perms
 
   config.vm.provider :virtualbox do |v|
     #v.gui = true
@@ -322,22 +313,22 @@ Vagrant.configure(2) do |config|
     b.vm.provider :virtualbox do |v|
       v.memory = 768
     end
-    b.vm.provision "install system packages", :type => :shell, :inline => packages_redhatted
-    b.vm.provision "install pyenv", :type => :shell, :privileged => false, :inline => install_pyenv("centos7_64")
-    b.vm.provision "install pythons", :type => :shell, :privileged => false, :inline => install_pythons("centos7_64")
-    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("centos7_64")
-    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg("centos7_64")
-    b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("centos7_64")
+    b.vm.provision "install system packages",       :type => :shell, :privileged => true, :inline => packages_redhatted
+    b.vm.provision "install pyenv",                 :type => :shell, :privileged => true, :inline => install_pyenv("centos7_64")
+    b.vm.provision "install pythons",               :type => :shell, :privileged => true, :inline => install_pythons("centos7_64")
+    b.vm.provision "build env",                     :type => :shell, :privileged => true, :inline => build_pyenv_venv("centos7_64")
+    b.vm.provision "install borg",                  :type => :shell, :privileged => true, :inline => install_borg("centos7_64")
+    b.vm.provision "run tests",                     :type => :shell, :privileged => true, :inline => run_tests("centos7_64")
   end
 
   config.vm.define "centos6_32" do |b|
     b.vm.box = "centos6-32"
-    b.vm.provision "install system packages", :type => :shell, :inline => packages_redhatted
-    b.vm.provision "install pyenv", :type => :shell, :privileged => false, :inline => install_pyenv("centos6_32")
-    b.vm.provision "install pythons", :type => :shell, :privileged => false, :inline => install_pythons("centos6_32")
-    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("centos6_32")
-    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg_no_fuse("centos6_32")
-    b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("centos6_32")
+    b.vm.provision "install system packages",       :type => :shell, :privileged => true, :inline => packages_redhatted
+    b.vm.provision "install pyenv",                 :type => :shell, :privileged => true, :inline => install_pyenv("centos6_32")
+    b.vm.provision "install pythons",               :type => :shell, :privileged => true, :inline => install_pythons("centos6_32")
+    b.vm.provision "build env",                     :type => :shell, :privileged => true, :inline => build_pyenv_venv("centos6_32")
+    b.vm.provision "install borg",                  :type => :shell, :privileged => true, :inline => install_borg_no_fuse("centos6_32")
+    b.vm.provision "run tests",                     :type => :shell, :privileged => true, :inline => run_tests("centos6_32")
   end
 
   config.vm.define "centos6_64" do |b|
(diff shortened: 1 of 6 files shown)


PlasmaPower:
Okay, first network based FS is implemented (SSHFS). Implementation is a bit hacky (e.g. rootTestingDir hardcoded in sshd_config, couldn't get ChrootDirectory working), but I've made sure to comment it well and it should work fine as-is.

PlasmaPower:
`nfsd` is... interesting. It seems to be implemented as a kernel module. It also has no way to specify a custom config file location from what I can tell, meaning that I can't easily automate it. I'm going to not test NFS, at least for now.

PlasmaPower:
Okay, CIFS is done after a lot of trouble with configuration. That means all the filesystems I was planning to test (except NFS which I couldn't) are done! Any more recommendations?

I'm now going to test if borg actually succeeds in these filesystems (I'm expecting they'll be bugs :/).

PlasmaPower:
Waiting on #1361, #1362, and #1363 to merge (FS specific bugs). Other than that, I think this is good.

PlasmaPower:
Okay, I've added all the drivers I could find to the Vagrantfile, but I might have missed some and some might not exist that I thought existed (or had different names). I'm going to wait on the FS specific issues to be fixed before moving forward though, wouldn't want to have to redo everything.

vagrant-tools/run-tests.sh
```diff
@@ -0,0 +1,99 @@
+. ~/.bash_profile
+cd /vagrant/borg/borg
+. ../borg-env/bin/activate
+if which pyenv > /dev/null; then
+  # for testing, use the earliest point releases of the supported python versions:
+  pyenv global 3.4.0 3.5.0
+fi
+# otherwise: just use the system python
+
+testingPartitionSize=128M
```
ThomasWaldmann: what is the minimum that would safely work?

vagrant-tools/run-tests.sh
```diff
@@ -0,0 +1,99 @@
...
+rootTestingDir="/var/tmp/borg-tests" # Also hardcoded in server configs in this directory
+mkdir -p "$rootTestingDir/rootfs"
+mountBase="/mnt/borg-tests"
+mkdir -p "$mountBase"
+
+rootFS="$(grep -oP '^[^ ]* / \K[^ ]*' /proc/mounts)"
```
ThomasWaldmann: this and some stuff below rather looks like it is only working on linux or few other systems.

vagrant-tools/sshd_config
```diff
@@ -0,0 +1,24 @@
...
+
+AuthorizedKeysFile /var/tmp/borg-tests/ssh_key.pub
+ChallengeResponseAuthentication no
+PasswordAuthentication no
+
+# TODO get ChrootDirectory working
+
+# For some reason, sshd checks /var/tmp's permissions
+# Of course, those are loose
+# The actual key is outputted by ssh-keygen with proper permissions
```
ThomasWaldmann: <nitpick>
maybe rather use "output" than "outputted".
also, finish sentences with a dot.
</nitpick>

vagrant-tools/run-tests.sh
PlasmaPower: I don't know, and I would like to support future tests. The FS minimum size is usually small, around 10M, so I don't think that's a big problem. Maybe we should drop this to 16M? I'll check that

vagrant-tools/run-tests.sh
PlasmaPower: This works on Linux and should work on *BSD from what I can tell. If this fails, it fails with warnings but a blank list of filesystems (except "rootfs" which isn't a mount point at all, just a dir). OSX might not have these /proc and /lib endpoints, but OSX doesn't have stable FS drivers for these either.

vagrant-tools/run-tests.sh
PlasmaPower: Looks like 16M works. Switching to that now.

vagrant-tools/run-tests.sh
PlasmaPower: This has been marked outdated because of an unrelated change, but it's still relevant. Are you fine with the current situation of this or would you like support for OSX?

PlasmaPower:
I think I'm setting some naming conventions here, @ThomasWaldmann do you agree with them? I can change them of course.

vagrant-tools/smb.conf
```diff
@@ -0,0 +1,22 @@
+[global]
+smb ports = 10445
+hosts allow = 127.0.0.1 ::1
+pid directory = /var/tmp/borg-tests
+security = user
+map to guest = Bad User
+
+# Who thought it was a good idea
+# to put printing in a file sharing protocol?
```
enkore: lol :D

vagrant-tools/run-tests.sh
PlasmaPower: 16M fails on NTFS, switching to 32M.

PlasmaPower: I've switched it back to 16M with a custom value of 128M for NTFS because it still runs out at 32M.

ThomasWaldmann:
names should be like in pep8.

PlasmaPower:
@ThomasWaldmann Should the names of files/directories also be snake_case?

ThomasWaldmann:
dashes in file/dirnames are fine imho.

PlasmaPower:
Okay, shell scripts are now use snake_case variable names.

TODO: #1362 (in progress), #1363 (will involve some changes to `run-tests.sh` here), check driver packages, check that this all works in Vagrant

PlasmaPower:
Does `/var/tmp` exist everywhere we're testing? If not, is there another suitable directory? (can't be tmpfs because of space requirements)

enkore:
I think /tmp is fine, if somewhere there is not much memory then there should be no tmpfs there. It is also "without a doubt portable" :)

PlasmaPower:
@enkore Unfortunately it isn't - look at our Vagrantfile configuration:

```rb
      v.memory = 768
    end
```

That means that at maximum, we have 768 megabytes of `/tmp`. That's if all of the memory was devoted to tmpfs though. In reality we'll probably have closer to 256 or 128 megabytes.

PlasmaPower:
I've now added the environment variable `BORG_TESTS_UNRELIABLE_MODES=true` to tests run on `ntfs-3g`. However, due to a lack of builtin support for Bash 4 on OS X (licensing issues) the code for that is hacky. Mind looking over it? (most it's in the directory loop, and there's a comment before the declaration of the directories variable explaining the variable format)

PlasmaPower:
I did a bit of research, and it looks like on OS X `/var` is symlinked to `/private/var` and `/private/var/tmp` is a valid (and not tmpfs) temporary directory.

I've also updated this PR to better handle OS X, it will support SSHFS and ntfs-3g once osxfuse 3.x is released in homebrew.

PlasmaPower:
With this recent push + #1373 + #1383, `vagrant up trusty64` works with all tests passing on all filesystems! :tada:

I'll need to figure out #1384 to test it on more boxes though.

PlasmaPower:
Minor fix, this is now working on OS X.

PlasmaPower:
With this newest push, error handling in the test driver is redone and cleaned up, and CentOS 7 works properly. Now I just need to test (and probably fix) *BSD.

PlasmaPower:
I think I spoke too soon, the test driver was working on CentOS 7 but pyenv wasn't. I've switched around some permissions, and I think it should work but I'm doing some more testing now.

PlasmaPower:
Okay, tests are now passing on CentOS 7. I also cleaned up Vagrantfile provision commands, hopefully that's okay.

ThomasWaldmann:
how about wheezy32 and freebsd64?

PlasmaPower:
@ThomasWaldmann I've added a checklist of all the systems our Vagrantfile supports to the original issue text, and I'm going through them one by one. I've had to make changes for almost all of them so far, so I'm not planning on skipping any.

ThomasWaldmann:
Good. Maybe check these 2 next, they are more important as they are used to build the binaries.

Vagrantfile
```diff
@@ -320,108 +312,108 @@ Vagrant.configure(2) do |config|
...
-    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("centos7_64")
-    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg("centos7_64")
-    b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("centos7_64")
+    b.vm.provision "install system packages",       :type => :shell, :privileged => true,  :inline => packages_redhatted
+    b.vm.provision "install pyenv",                 :type => :shell, :privileged => false, :inline => install_pyenv("centos7_64")
+    b.vm.provision "install pythons",               :type => :shell, :privileged => false, :inline => install_pythons("centos7_64")
+    b.vm.provision "build env",                     :type => :shell, :privileged => false, :inline => build_pyenv_venv("centos7_64")
+    b.vm.provision "install borg",                  :type => :shell, :privileged => false, :inline => install_borg("centos7_64")
+    b.vm.provision "run tests",                     :type => :shell, :privileged => true,  :inline => run_tests("centos7_64")
```
ThomasWaldmann: why did you change to priviledged?

vagrant-tools/run-tests.sh
```diff
@@ -0,0 +1,164 @@
...
+    fuse)
+      fuse_supported=true
+      ;;
+  esac
+done <<< "$filesystems"
+
+if [[ "$fuse_supported" ]]; then
+  if which sshfs > /dev/null; then
+    mkdir "$root_testing_dir/sshfs"
+    ssh-keygen -t rsa -b 4096 -C 'borgbackup@github.com' -N '' -f "$root_testing_dir/ssh_key"
```
ThomasWaldmann: how long does it take to generate a 4096 bit key?
maybe we can use a shorter one to speedup the tests.

PlasmaPower: It only takes about a second, but a 2048 bit key is practically instant so I'll use that.

ThomasWaldmann:
Think we could merge this after these 2 are tested and then fix the rest (if needed).

PlasmaPower:
Should I execute `tox` with fakeroot? I'm not sure why we were in the first place. I was originally executing it as root, but I ran into problems on some distros as pyenv wasn't setup as root. I moved it over to the non-privileged user, but I didn't add fakeroot. Should I?

ThomasWaldmann:
@PlasmaPower some tests want to do e.g. chown and that only works with (fake)root.

PlasmaPower:
Okay, makes sense. Is there a reason we don't just do everything as root?

ThomasWaldmann:
Well, if we are on our own machines (like vagrant) and it works(tm), we could do it.

PlasmaPower:
I'm getting really odd errors on CIFS. I've put the debugging info in [this gist](https://gist.github.com/PlasmaPower/db80e2fce8e4926b53c7695cdc17afd7) because it's a bit long.

PlasmaPower:
I'd appreciate help debugging that error. I'm not really sure what's causing it.

ThomasWaldmann:
Well, seems like a crash in pytest itself.

PlasmaPower:
It does, but I find it odd that it occurs right after a failure in one of our tests.

Maybe I should just disable CIFS?

PlasmaPower:
I can't figure this issue out, so for now at least I'm just disabling CIFS. We already have a network based FS anyway.

ThomasWaldmann:
Pity. Maybe we can enable it again later, CIFS is quite popular, even for internet storage boxes.

PlasmaPower:
Yeah, in my local changeset (will push after testing FreeBSD) it's just continued past with a comment explaining the problem. Hopefully that means it'll be easy to re-enable if the problem gets fixed.

Also, I'm getting some errors on FreeBSD, issue will be up with detailed information soon.

PlasmaPower:
Once #1416 and #1434 have been resolved, I'll rerun the tests, and then this should be good to merge.

PlasmaPower:
Okay, I'm going to revisit the PR to rerun tests now that some of them should have been fixed.

PlasmaPower:
I've been really busy, but I currently have some time to work on this (and I should be relatively free for a while). I'm running tests for wheezy32 now, I'll post logs if anything comes up of course.

It's unfortunate that the whole thing has to be rebuilt and dependencies have to be reinstalled for each filesystem. Is there a workaround that I could use?

ThomasWaldmann:
@PlasmaPower not sure what you mean with rebuilt/reinstalled for each filesystem.

but as a general remark: we'll soon have 1.1.0b3 and likely that will be the last beta, follow by 1+ release candidate. so it would be good to get your PR finished soon, so it can be in 1.1.0 release.

ThomasWaldmann:
another thing: I'ld like to keep the count of toplevel items in the repo low, so how about moving the stuff from vagrant-tools/... to scripts/vagrant/... ?

PlasmaPower:
If anyone else wants to pick up this PR (and claim the bounty), that's fine. The test time is simply too long to try to fix a lot of small issues. Each time tox is run with a different TMPDIR it seems to reinstall dependencies, I haven't looked into preventing this, a symlink in the TMPDIR to a central location might fix it.

PlasmaPower:
And yes, tests are failing on various filesystems, but a lot of the time it's hard to tell where the problem is coming from and if it's the actual tests or just my test driver. Also, it seems to also be somewhat OS dependent.

ThomasWaldmann:
continued in #1820.

PlasmaPower:
Great! I've followed that issue, I'll be happy to help if you're wondering why something's there/what is does.

Issue #1315, opened 2016-07-12
testsuite failure on hurd-i386


LocutusOfBorg:
Hi, following the error:
(note: hurd-i386 is not a release architecture, so you can just close this issue)
https://buildd.debian.org/status/fetch.php?pkg=borgbackup&arch=hurd-i386&ver=1.0.5-2&stamp=1468279201
```
plugins: cov-2.2.1
collected 554 items

borg/testsuite/archive.py ...............................
borg/testsuite/archiver.py .s..F.............s..............X.........sssssssssssssssssssssssssssssssssssssssss........F......s......s..............X...........
borg/testsuite/chunker.py ...
borg/testsuite/compress.py ........
borg/testsuite/crypto.py .....
borg/testsuite/hashindex.py .....................
borg/testsuite/helpers.py ...............................................................................................................................................................
borg/testsuite/key.py ....
borg/testsuite/locking.py ..............
borg/testsuite/logger.py ....
borg/testsuite/lrucache.py ..
borg/testsuite/platform.py sssss
borg/testsuite/repository.py .........................................
borg/testsuite/shellpattern.py .............................................................
borg/testsuite/upgrader.py sssss..
borg/testsuite/xattr.py s

=================================== FAILURES ===================================
_________________________ ArchiverTestCase.test_atime __________________________

self = <borg.testsuite.archiver.ArchiverTestCase testMethod=test_atime>

    def test_atime(self):
        self.create_test_files()
        atime, mtime = 123456780, 234567890
        os.utime('input/file1', (atime, mtime))
        self.cmd('init', self.repository_location)
        self.cmd('create', self.repository_location + '::test', 'input')
        with changedir('output'):
            self.cmd('extract', self.repository_location + '::test')
        sti = os.stat('input/file1')
        sto = os.stat('output/input/file1')
        assert sti.st_mtime_ns == sto.st_mtime_ns == mtime * 1e9
        if hasattr(os, 'O_NOATIME'):
>           assert sti.st_atime_ns == sto.st_atime_ns == atime * 1e9
E           AssertionError: assert 1468278979000000000 == 123456780000000000
E            +  where 1468278979000000000 = os.stat_result(st_mode=35309, st_ino=253216, st_dev=5, st_nlink=2, st_uid=100, st_gid=200, st_size=81920, st_atime=1468278979, st_mtime=234567890, st_ctime=1468278979).st_atime_ns
[...]

Issue #1363, opened 2016-07-23
Tests failing on ntfs-3g


PlasmaPower:
- [ ] No ACLs: `KeyError: 'acl_access'` pops up a bunch - also present on SSHFS (see #1362, fixed in #1373)
- [ ] Funky permissions and modes: (fixed in #1383)
  ```
                  # i have seen cases where the copied tree has world-readable
                  # permissions, which is wrong
>               assert stat_segment(backup).st_mode & UMASK_DEFAULT == 0
E               assert (33279 & 63) == 0

.tox/py35/lib/python3.5/site-packages/borg/testsuite/upgrader.py:180: AssertionError
  ```

  ```
          # Directory replaced with a regular file
>       assert '[drwxr-xr-x -> -rwxr-xr-x] input/dir_replaced_with_file' in output
E       AssertionError: assert '[drwxr-xr-x -> -rwxr-xr-x] input/dir_replaced_with_file' in '    +13 B       0 B input/empty\n    +13 B       0 B input/hardlink_contents_changed\n     +1 B
 -1.0 kB input/file_...moved       512 B input/file_removed2\nremoved directory   input/dir_removed\nremoved link        input/link_removed\n'
  ```

  ```
          mode = os.stat(self.repository_path).st_mode
>       self.assertEqual(stat.S_IMODE(mode), 0o700)
E       AssertionError: 0o777 != 0o700

.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:995: AssertionError
```

- [x] No atime - fixed with #1367:

These should be fixed before #1359 is merged.

**Comment by PlasmaPower** [comment]
Anybody have any idea what's going on with the atime error? It seems like we are accounting for no atime, but the atime is just incorrect. Is that the case, or is the test case broken?

**Comment by enkore** [comment]
Looks like O_NOATIME was just not working. Hmm...

```py
        def has_noatime(some_file):
            atime_before = os.stat(some_file).st_atime_ns
            try:
                os.close(os.open(some_file, flags_noatime))
except PermissionError:
```

Can you try

```py
        def has_noatime(some_file):
            atime_before = os.stat(some_file).st_atime_ns
            try:
                with open(os.open(some_file, flags_noatime)) as file: file.read()
except PermissionError:
```

instead?

PlasmaPower:
@enkore That generates this error:

```python
    def has_noatime(some_file):
        atime_before = os.stat(some_file).st_atime_ns
        try:
[...]

PR #1367, opened 2016-07-23
Fixed noatime detection

PlasmaPower:
Part of #1363 

Thanks to @enkore for the code: https://github.com/borgbackup/borg/issues/1363#issuecomment-234726334


enkore:
Might also have fixed #1315 for good :)

PR #1365, opened 2016-07-23
Allow running tests outside of Vagrant

PlasmaPower:
Not too much too it. Allows running tests with `sudo tox` instead of vagrant or `fakeroot -u tox`. Resolves #1361.

PR #1368, opened 2016-07-23
Catch a ValueError when checking if a user exists

PlasmaPower:
Fixes https://github.com/borgbackup/borg/pull/1365#issuecomment-234734212 as specified here: https://github.com/borgbackup/borg/pull/1365#discussion_r71976549

### DIFF
--- a/src/borg/testsuite/platform.py
+++ b/src/borg/testsuite/platform.py
@@ -40,7 +40,7 @@ def user_exists(username):
     try:
         pwd.getpwnam(username)
         return True
-    except KeyError:
+    except (KeyError, ValueError):
         return False
 
 


enkore:
Fix confirmed, thanks.

src/borg/testsuite/platform.py
```diff
@@ -40,7 +40,7 @@ def user_exists(username):
     try:
         pwd.getpwnam(username)
         return True
-    except KeyError:
+    except (KeyError, ValueError):
```
ThomasWaldmann: so, how can i reproduce the ValueError (UnicodeError)?

PlasmaPower: Test it on *BSD.

PR #1369, opened 2016-07-23
Extends delete command with --oldest & --latest options

funkyfuture:
i added a `--latest` and `--oldest` option to simply delete the newest resp. oldest backup from a repository. eg, it comes handy in bash scripts:

``` bash
while used_space_exceeds_pcent; do
    log 'Used space exceeds ${MAX_DISK_USAGE}%, deleting oldest backup archive.'
    borg delete --oldest --force $BACKUP_PATH & wait
done
```

alternatively, it could be implemented as `--last n` and `--first n` where `n` defaults to 1.


src/borg/archiver.py
```diff
@@ -696,8 +696,25 @@ def do_rename(self, args, repository, manifest, key, cache, archive):
     @with_repository(exclusive=True, manifest=False)
     def do_delete(self, args, repository):
         """Delete an existing repository or archive"""
+        manifest = None
+
+        if args.oldest or args.latest:
+            if args.location.archive:
+                logger.warning('The options --oldest and --latest have no effect on archive targets.')
```
enkore: This should abort to avoid unintentionally deleting stuff.

enkore:
Interesting feature. #128 is kinda related, maybe we can factor this into a more generic version that is easier to apply to other commands (e.g. an `Archiver` method that sets args.location.archive from --latest/--oldest flags). What do you think?

funkyfuture:
sure, more generic is always reasonable. what about these properties / methods?

- `latest` -> an archive name or `None`
- `last(n)` -> a possibly empty list of archive names, returns all archive names of the repo if `n` is greater than the number of archives
- `oldest` (equivalent)
- `first(n)` (equivalent)

funkyfuture:
or should these rather be attributes of `helpers.Manifest`?

funkyfuture:
i followed these thoughts by implementing with this outcome:

the functionality of `Manifest.list_archive_infos` is flexible enough to avoid overhead in the deletion procedure. don't know if they could be usable in some context. so i'd rather add an `Archiver` method that takes `args` and `repository` or `manifest` to figure out a scope for similar uses of these arguments. `info`, `list` and `check` seem to be the only candidates for it.

i ended up with something that is nox exactly what #128 proposes. it's rather a slicing now, so `--latest 1` instead of just `--latest`, but i see no disadvantage here.

ThomasWaldmann:
I also did implement some --last N stuff for borg check already, see the code there.

src/borg/archiver.py
```diff
@@ -696,43 +696,78 @@ def do_rename(self, args, repository, manifest, key, cache, archive):
...
+            n = args.oldest or args.latest
+            archives = manifest.list_archive_infos('ts', reverse=args.latest)[:n]
+            for archive in archives:
+                logger.info('Deleting %s' % archive.name)
+                args.location.archive = archive.name
+                self._delete_archive(args, repository, manifest)
+                if self.exit_code:
+                    break
+            else:
+                logger.error('There are no archives.')
```
ThomasWaldmann: doesn't it always end in here, if there is no exit_code > 0 (warning or error) above?
if we successfully deleted all archives, why does it emit this error? (assuming we wanted to delete all archives)

src/borg/helpers.py
```diff
@@ -158,6 +158,20 @@ def list_archive_infos(self, sort_by=None, reverse=False):
             archives = sorted(archives, key=attrgetter(sort_by), reverse=reverse)
         return archives
 
+    @property
+    def oldest_archive(self):
+        return self.oldest_archives(1) or None
```
ThomasWaldmann: should not return a list, but directly the oldest archive (or None).

src/borg/helpers.py
```diff
@@ -158,6 +158,20 @@ def list_archive_infos(self, sort_by=None, reverse=False):
...
+
+    def oldest_archives(self, n):
+        return self.list_archive_infos('ts')[:n]
+
+    @property
+    def latest_archive(self):
+        return self.latest_archives(1) or None
```
ThomasWaldmann: same here.

src/borg/archiver.py
```diff
@@ -1687,6 +1722,11 @@ def build_parser(self, prog=None):
         subparser.add_argument('location', metavar='TARGET', nargs='?', default='',
                                type=location_validator(),
                                help='archive or repository to delete')
+        group = subparser.add_mutually_exclusive_group()
+        group.add_argument('--oldest', dest='oldest', metavar='n', default=0, type=int,
+                           help='delete n oldest archives')
+        group.add_argument('--latest', dest='latest', metavar='n', default=0, type=int,
```
ThomasWaldmann: as borg check already has --last, maybe use it here also. and --first. both for consistency.

src/borg/archiver.py
funkyfuture: no, this is only reached if `archives` is empty.

src/borg/helpers.py
funkyfuture: yeah, thanks for the spot. though this is still dead code.

funkyfuture: if there's no use for it, i will remove it.

src/borg/archiver.py
```
>>> for i in range(3):
...  print(i)
... else:
...  print("didn't break out of the loop")
... 
0
1
2
didn't break out of the loop
```

funkyfuture:
> If generalizing that, we need to be careful with defaults (if that code includes handling of that).
> For borg check of course it makes sense to have a default of "check all", but not for borg delete.

i think the simplest would be to allow `--last N` and `first N` only on repositories, thus `parse_args` should check that the target is no archive.

though i would prefer a default of 1 for each argument, problems arise as the code can't figure out whether and which flag was actually provided by the user. a clever idea to deal with that would be interesting, but still one would need `--` after an optional argument with an omitted value.

as of now, both flags must be used with an explicit argument, which makes it quiet sane, imo.

this distinguishes it from the `--last` option. hence i wouldn't use the same name.

src/borg/archiver.py
funkyfuture: thanks for reminding me that i need to get rid of that misconception i have about `for ... else`.

funkyfuture:
i tend to move argparsing code to an own module, especially as `build_parser` is hard to maintain. any objections?

ThomasWaldmann:
we already thought about that, but delayed it because we want to maintain multiple branches concurrently and be able to merge changes in archiver.py without troubles. and we shouldn't do big src cleanup in 1.0-maint.

funkyfuture:
okay, i can need some feedback here.

`info` and `list` are now also usable with something i would term slice selection for now, but would prefer something clearer. 

the rationale for using `--oldest` and `--lastest` as argument names is their more specific meaning concerning time, as the antonyms `--first` and `--last` may also be used in another sorting contexts.

i would propose to refactor `check` similarly by moving the selection and dispatching logic to `Archiver` and making the `--last` option an alias of `--latest`.

src/borg/archiver.py
```diff
@@ -707,45 +707,72 @@ def do_rename(self, args, repository, manifest, key, cache, archive):
...
+                          stats.summary.format(label='Deleted data:', stats=stats),
+                          str(cache),
+                          DASHES, logger=logging.getLogger('borg.output.stats'))
+        return self.exit_code
+
+    def _delete_archives(self, args, repository):
+        """Delete multiple archives"""
+        manifest, key = Manifest.load(repository)
+        archives = self._get_archives_slice(args, manifest)
+        for i, archive in enumerate(archives):
```
ThomasWaldmann: ... enumerate(archives, 1) (and remove the +1 below)

src/borg/archiver.py
```diff
@@ -787,65 +814,105 @@ def write(bytestring):
...
+        print('Command line: %s' % format_cmdline(archive.metadata[b'cmdline']))
+        print(DASHES)
+        print(STATS_HEADER)
+        print(str(stats))
+        print(str(cache))
+        return self.exit_code
+
+    def _info_archives(self, args, repository, manifest, key, cache):
+        archives = self._get_archives_slice(args, manifest)
+        for i, archive in enumerate(archives):
```
ThomasWaldmann: same here, for consistency

src/borg/archiver.py
```diff
@@ -787,65 +814,105 @@ def write(bytestring):
...
+            args.location.archive = archive.name
+            self._info_archive(args, repository, manifest, key, cache)
+            if self.exit_code:
+                break
+            if len(archives) - i > 1:
+                print('\n')
```
ThomasWaldmann: this outputs 2 * '\n'.

src/borg/archiver.py
```diff
@@ -2222,6 +2292,14 @@ def build_parser(self, prog=None):
                                help='hex object ID(s) to delete from the repo')
         return parser
 
+    @staticmethod
+    def add_archives_slice_selection_args(subparser):
+        group = subparser.add_mutually_exclusive_group()
+        group.add_argument('--oldest', dest='oldest', metavar='N', default=0, type=int,
+                           help='delete n oldest archives')
```
ThomasWaldmann: N

src/borg/archiver.py
```diff
@@ -2222,6 +2292,14 @@ def build_parser(self, prog=None):
...
+        group.add_argument('--latest', dest='latest', metavar='N', default=0, type=int,
+                           help='delete n latest archives')
```

src/borg/archiver.py
```diff
@@ -2284,6 +2362,18 @@ def run(self, args):
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
         return args.func(args)
 
+    def _get_archives_slice(self, args, manifest):
+        if args.location.archive:
+            logger.error('The options --oldest and --latest must not be used on archive targets.')
```
ThomasWaldmann: hmm, maybe rather say what borg wants: ... must be used on a repository target.

src/borg/archiver.py
```diff
@@ -2284,6 +2362,18 @@ def run(self, args):
...
+            self.exit_code = EXIT_ERROR
+            return []
+        n = args.oldest or args.latest
+        archives = manifest.list_archive_infos('ts', reverse=args.latest)[:n]
```
ThomasWaldmann: maybe use reverse=bool(args.latest) - it is not strictly needed, but clearer that this will be evaluated as a boolean.

src/borg/archiver.py
```diff
@@ -2284,6 +2362,18 @@ def run(self, args):
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
         return args.func(args)
 
+    def _get_archives_slice(self, args, manifest):
+        if args.location.archive:
+            logger.error('The options --oldest and --latest must not be used on archive targets.')
+            self.exit_code = EXIT_ERROR
+            return []
+        n = args.oldest or args.latest
```
ThomasWaldmann: maybe add a "assert n > 0" here - this code should be only called if either oldest or latest is given and > 0.

ThomasWaldmann:
Just an idea: you cleaned up the toplevel dispatch method to be shorter and just call the 3 _xxxx methods for the 3 cases archive, archives, repo. They could be also functions INSIDE the dispatch method, not methods of Archiver. As you prefixed them with underscore, you indicate "don't call me" somehow anyway, so they could be also private functions defined inside the method?

ThomasWaldmann:
Can you rebase on current master?

ThomasWaldmann:
An idea about https://github.com/borgbackup/borg/pull/1369#issuecomment-239619666:

Maybe that is the reason why we rather want to use first and last (not oldest and latest).

We could just document, that sort order is by-timestamp (so first means oldest and last means latest), but later, if we feel the need, we could also sort by other criterial (using a sort-by option) and still use the same first/last option to give the index.

Not sure if we ever need it, currently there is only timestamp and name that somehow qualify for sorting, but selecting the first/last N of a list sorted by name - not sure if it makes sense.

funkyfuture:
> As you prefixed them with underscore, you indicate "don't call me" somehow anyway, so they could be also private functions defined inside the method?

whenever i come across such nested callables, i experience a great loss of readability. flat is better than nested. if you really want it like this, tell me. but i really consider it a bad practice.

> Not sure if we ever need it, currently there is only timestamp and name that somehow qualify for sorting, but selecting the first/last N of a list sorted by name - not sure if it makes sense.

to be consistent i would also add a `--sort-by` defaulting to `time`. this may seem overengineered atm, but as a set of arguments with `--first` and `--last` it makes sense.

enkore:
Another point about nesting callables is that the namespace inside a callable is anonymous for practical purposes, so unreachable for testing.

ThomasWaldmann:
Related: #128

funkyfuture:
aigh't, i added the `--sort-by` argument and grabbed some low hanging fruits on the way.

src/borg/archiver.py
```diff
@@ -862,14 +862,14 @@ def _list_archive(self, args, repository, manifest, key, write):
...
-            write('Contents of {} ({}/{}):'.format(archive.name, i+1, len(archives)))
+        for i, archive in enumerate(archives, 1):
+            write('Contents of {} ({}/{}):'.format(archive.name, i, len(archives)))
             args.location.archive = archive.name
             self._list_archive(args, repository, manifest, key, write)
             if self.exit_code:
                 break
             if len(archives) - i > 1:
-                write('\n')
+                write()
```
ThomasWaldmann: ehrm, guess you need to keep the \n here. write() does not write one by default, like print() does.

src/borg/archiver.py
```diff
@@ -2422,11 +2422,12 @@ def run(self, args):
 
     def _get_archives_slice(self, args, manifest):
         if args.location.archive:
-            logger.error('The options --oldest and --latest must not be used on archive targets.')
+            logger.error('The options --oldest and --latest can only used on repository targets.')
```
ThomasWaldmann: + be

src/borg/archiver.py
```diff
@@ -759,7 +761,7 @@ def _delete_archive(self, args, repository, manifest=None):
 
     def _delete_archives(self, args, repository):
         """Delete multiple archives"""
-        manifest, key = Manifest.load(repository)
+        manifest, _ = Manifest.load(repository)
```
ThomasWaldmann: hmm, this change is unrelated to this changeset, right? don't do that.

src/borg/archiver.py
```diff
@@ -2352,11 +2349,21 @@ def build_parser(self, prog=None):
 
     @staticmethod
     def add_archives_slice_selection_args(subparser):
+        valid_sort_keys = ('timestamp', 'name')
+        sort_by_default = 'timestamp'
+        sort_by_choices = []
+        for r in range(len(valid_sort_keys)):
+            sort_by_choices.extend([','.join(x) for x in permutations(valid_sort_keys, r+1)])
```
ThomasWaldmann: .append(elem)

ThomasWaldmann: you could even make a list comprehension out of this.

ThomasWaldmann: alternatively to using choices here, it could also be just a validated string.

src/borg/archiver.py
```diff
@@ -1537,10 +1537,7 @@ def build_parser(self, prog=None):
         subparser.add_argument('--save-space', dest='save_space', action='store_true',
                                default=False,
                                help='work slower, but using less space')
-        subparser.add_argument('--last', dest='last',
-                               type=int, default=None, metavar='N',
-                               help='only check last N archives (Default: all)')
-        subparser.add_argument('-P', '--prefix', dest='prefix', type=PrefixSpec,
+        subparser.add_argument('-P', '--prefix', dest='prefix', type=prefix_spec, default='',
```
ThomasWaldmann: where is this prefix_spec? it is not in this changeset.

src/borg/repository.py
```diff
@@ -685,8 +685,6 @@ def report_error(msg):
         return not error_found or repair
 
     def rollback(self, cleanup=False):
-        """
-        """
```
ThomasWaldmann: another unrelated change.

suggestion: if you find such stuff, make a note, but do not change it, but stay focussed on the topic of the changeset you want to make.

ThomasWaldmann:
this PR is a bit hard to review. single changesets seem to be incomplete, reviewing the overall changes is way to much.

ThomasWaldmann:
Can you redo your changes and:
- do not do any unrelated changes, no cleanup, no reformatting
- have each changeset self-contained, complete and focussed on some topic that matches the commit comment?
- if you did fixup commits, can you merge them with the commits they fixed?

funkyfuture:
sorry for the mess.

>if you did fixup commits, can you merge them with the commits they fixed?

that's the cause for the mixups in the commits. i'll see to consolidate and give a notice.

funkyfuture:
closing in favor of #1554

Issue #1220, opened 2016-06-29
deadlocks with multiple borg create


erikhydro:
the attached script shows a problem with locking. the script simulates what happens if you have multiple servers where on each of them a cronjob does a backup to the same borg repository on a central server at the same time. because of locking an option --lock-wait 86400 is added to have a maximum waiting time of 24 hours. longer is not good, because then the next cronjob backup is started. so after 24 hours continuous blocking i want to get error output which cron will send me by mail. so far so good. every server should automatically do its backup. one after another because of locking. but it does not work.

the script shows why. you need two computers. on the server you create a borg repo with borg init. on the client you run the script.

it will run 10 backups in parallel.

you will get lots of messages like these:

Remote: Borg 1.0.2: exception in RPC call:
Remote: Traceback (most recent call last):
Remote:   File "/usr/lib/python3/dist-packages/borg/remote.py", line 96, in serve
Remote:     res = f(*args)
Remote:   File "/usr/lib/python3/dist-packages/borg/repository.py", line 455, in put
Remote:     self.prepare_txn(self.get_transaction_id())
Remote:   File "/usr/lib/python3/dist-packages/borg/repository.py", line 199, in prepare_txn
Remote:     self.lock.upgrade()
Remote:   File "/usr/lib/python3/dist-packages/borg/locking.py", line 299, in upgrade
Remote:     self.acquire(exclusive=True, remove=SHARED)
Remote:   File "/usr/lib/python3/dist-packages/borg/locking.py", line 258, in acquire
Remote:     self._wait_for_readers_finishing(remove, sleep)
Remote:   File "/usr/lib/python3/dist-packages/borg/locking.py", line 287, in _wait_for_readers_finishing
Remote:     raise LockTimeout(self.path)
Remote: borg.locking.LockTimeout: /root/foorepo/lock
Remote: Platform: Linux nagios 3.16.0-4-686-pae #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08) i686
Remote: Linux: debian 8.5 
Remote: Borg: 1.0.2  Python: CPython 3.4.2
Remote: PID: 5658  CWD: /root
Remote: sys.argv: ['/usr/bin/borg', 'serve', '--umask=077']
Remote: SSH_ORIGINAL_COMMAND: None

the script won't terminate. there will not be 10 finished backups. if you then kill the script, you will have zero backups.

[borglockchecker.zip](https://github.com/borgbackup/borg/files/338675/borglockchecker.zip)

***

[...]

PR #1371, opened 2016-07-24
fix deadlock

ThomasWaldmann:
see #1220 + commit comments.

The deadlock is fixed already.

But: needed a RPC API change to add the exclusive param. :-(

Also: Thinking about whether we maybe rather want a safe default for exclusive.
Currently it is usually False and you need to give exclusive=True - guess the opposite way would be safer, so one does not forget requesting an exclusive lock.

Nothing really bad would happen if one forgets though, the code has an automatic lock upgrade to an exclusive lock.

The lock upgrade/downgrade code was kept to support old clients.
Current code shall not use it, it is deprecated.

### DIFF
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -114,7 +114,7 @@ class RepositoryServer:  # pragma: no cover
     def negotiate(self, versions):
         return RPC_PROTOCOL_VERSION
 
-    def open(self, path, create=False, lock_wait=None, lock=True, append_only=False):
+    def open(self, path, create=False, lock_wait=None, lock=True, exclusive=False, append_only=False):
         path = os.fsdecode(path)
         if path.startswith('/~'):
             path = path[1:]
@@ -125,7 +125,9 @@ class RepositoryServer:  # pragma: no cover
                     break
             else:
                 raise PathNotAllowed(path)
-        self.repository = Repository(path, create, lock_wait=lock_wait, lock=lock, append_only=self.append_only or append_only)
+        self.repository = Repository(path, create, lock_wait=lock_wait, lock=lock,
+                                     append_only=self.append_only or append_only,
+                                     exclusive=exclusive)
         self.repository.__enter__()  # clean exit handled by serve() method
         return self.repository.id
 
@@ -141,7 +143,7 @@ class RemoteRepository:
     class NoAppendOnlyOnServer(Error):
         """Server does not support --append-only."""
 
-    def __init__(self, location, create=False, lock_wait=None, lock=True, append_only=False, args=None):
+    def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock=True, append_only=False, args=None):
         self.location = self._location = location
         self.preload_ids = []
         self.msgid = 0
@@ -178,16 +180,13 @@ class RemoteRepository:
                 raise ConnectionClosedWithHint('Is borg working on the server?') from None
             if version != RPC_PROTOCOL_VERSION:
                 raise Exception('Server insisted on using unsupported protocol version %d' % version)
-            # Because of protocol versions, only send append_only if necessary
-            if append_only:
-                try:
-                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
-                except self.RPCError as err:
-                    if err.remote_type == 'TypeError':
-                        raise self.NoAppendOnlyOnServer() from err
-                    else:
-                        raise
-            else:
+            try:
+                self.id = self.call('open', self.location.path, create, lock_wait, lock, exclusive, append_only)
+            except self.RPCError as err:
+                if err.remote_type != 'TypeError':
+                    raise
+                if append_only:
+                    raise self.NoAppendOnlyOnServer()
                 self.id = self.call('open', self.location.path, create, lock_wait, lock)
         except Exception:
             self.close()
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -114,7 +114,7 @@ class RepositoryServer:  # pragma: no cover
     def negotiate(self, versions):
         return RPC_PROTOCOL_VERSION
 
-    def open(self, path, create=False, lock_wait=None, lock=True, exclusive=False, append_only=False):
+    def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, append_only=False):
         path = os.fsdecode(path)
         if path.startswith('/~'):
             path = path[1:]
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -79,7 +79,7 @@ class Repository:
         if self.do_create:
             self.do_create = False
             self.create(self.path)
-        self.open(self.path, self.exclusive, lock_wait=self.lock_wait, lock=self.do_lock)
+        self.open(self.path, bool(self.exclusive), lock_wait=self.lock_wait, lock=self.do_lock)
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
@@ -317,6 +317,9 @@ class Repository:
         self.compact = set()
 
     def replay_segments(self, index_transaction_id, segments_transaction_id):
+        # fake an old client, so that in case we do not have an exclusive lock yet, prepare_txn will upgrade the lock:
+        remember_exclusive = self.exclusive
+        self.exclusive = None
         self.prepare_txn(index_transaction_id, do_cleanup=False)
         try:
             segment_count = sum(1 for _ in self.io.segment_iterator())
@@ -332,6 +335,7 @@ class Repository:
             pi.finish()
             self.write_index()
         finally:
+            self.exclusive = remember_exclusive
             self.rollback()
 
     def _update_index(self, segment, objects, report=None):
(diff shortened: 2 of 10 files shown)


ThomasWaldmann:
From IRC discussion:
- need to keep lock upgrade on txn start, for compat with older clients (== if there is no excl. lock already, do an upgrade)
- thus: need to keep lock upgrade/downgrade code in borg 1.0, but deprecate it (emit warning about old client?)
- keep all other changes as is, so new clients based on this code will not use lock upgrade

ThomasWaldmann:
any comments about the `exclusive` default value? change it? in 1.0? in 1.1?

ThomasWaldmann:
please test / review, this is the last essential issue to do for 1.0.7 milestone.

enkore:
I'll review this offline

borg/remote.py
```diff
@@ -176,16 +175,11 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, append_onl
                 raise ConnectionClosedWithHint('Is borg working on the server?') from None
             if version != RPC_PROTOCOL_VERSION:
                 raise Exception('Server insisted on using unsupported protocol version %d' % version)
-            # Because of protocol versions, only send append_only if necessary
-            if append_only:
-                try:
-                    self.id = self.call('open', self.location.path, create, lock_wait, lock, append_only)
-                except self.RPCError as err:
-                    if err.remote_type == 'TypeError':
-                        raise self.NoAppendOnlyOnServer() from err
```
textshell: Why remove this? Silently ignoring Repository settings on create doesn‘t sound nice.

ThomasWaldmann: it's now either both exclusive and append_only params (new api) or neither of them (old api).

textshell: what i meant was that the old api code path should do:

    if append_only:
        raise self.NoAppendOnlyOnServer()

Thus keeping the safe guard from the original PR.

textshell:
For #1374 i either need this in enforcing mode (i.e. situation with a client and server that both have the iv parts needs to never upgrade on put) or i need to add independend locking (or lock upgrading) to #1374

borg/remote.py
ThomasWaldmann: fixed, see new commit.

borg/testsuite/repository.py
```diff
@@ -156,17 +156,6 @@ def test_crash_before_compact_segments(self):
...
-        self.add_keys()
-        for name in os.listdir(self.repository.path):
-            if name.startswith('index.'):
-                os.unlink(os.path.join(self.repository.path, name))
-        with patch.object(UpgradableLock, 'upgrade', side_effect=LockFailed) as upgrade:
-            self.reopen()
-            with self.repository:
-                self.assert_raises(LockFailed, lambda: len(self.repository))
-                upgrade.assert_called_once_with()
-
```
textshell: Why remove this? I think this case is still relevant? i.e. we still need to upgrade, and we still need not to hit any asserts or other problems.

borg/repository.py
```diff
@@ -207,14 +207,18 @@ def open_index(self, transaction_id):
...
-        try:
-            self.lock.upgrade()
-        except (LockError, LockErrorT):
-            # if upgrading the lock to exclusive fails, we do not have an
-            # active transaction. this is important for "serve" mode, where
-            # the repository instance lives on - even if exceptions happened.
-            self._active_txn = False
-            raise
+        if not self.lock.got_exclusive_lock():
+            # this is either an old client talking to a new server (expecting lock upgrade) or
```
textshell: If we want this to be enforcing that upgrades are only used by old code and by index replay we could do a
    if self.exclusive is not None:
        assert("like in your old version")

and change
* the default in remote to be None.
* in __enter__ use True if self.exclusive else False
* change the replay code to disable this check by setting self.exclusive to None

borg/testsuite/repository.py
ThomasWaldmann: I couldn't make much sense of this test.
I first completely removed the lock upgrading code.
Later I re-added it for older clients.
But the current code does not do lock upgrades.

textshell: I think it still does.
* Do a borg list with a remote repo that had it’s last writer crash and no other access since then
* client eventually calls Repository.get
* which calls Repository.get_transaction_id()
* which checks for crash by index_transaction_id != segments_transaction_id
* and calls Repository.replay_segments
* which calls self.prepare_txn
* which needs to upgrade

This is the one legitimate case i can come up where an always read only operation (that correctly uses an non exclusive lock) can get into prepare_txn.

textshell:
Apart from the above this looks good.

Just noticed though that change-passphrase and migrate-to-repokey use Repository.save_key which is not covered by any lock. Not sure we care enough to make a separate ticket for that though.

ThomasWaldmann:
@textshell please verify if the last changeset is what you meant.

textshell:
Yes, that‘s what i meant. Can you please resurrect the test (or similar) for the lock upgrade on replay?
It seems the travis unhappiness is for unrelated reasons.

ThomasWaldmann:
Test resurrected and differentiated into 2 tests for old and new clients.

enkore:
> thomas/fix-deadlock
> with_repository arg changes: ok
> prepare_txn: ok
> locking code: not much of an idea...
> test_replay_of_readonly_repository: hmm, look into test spec, why must be removed

I see the last bit was already addressed :)

ThomasWaldmann:
@enkore guess the most delicate part is the compatibility.

ThomasWaldmann:
about lock upgrade/downgrade: I wanted to deprecate that, due to the obvious deadlock issues, but it seems that even with the new code we need the lock upgrade for somehow unexpectedly needing a write lock, e.g. when borg list finds a crashed repo and needs to do segment replay.

So I think I am removing the deprecation notes and replacing them with a deadlock warning.

borg/remote.py
```diff
@@ -176,16 +178,13 @@ def __init__(self, location, create=False, lock_wait=None, lock=True, append_onl
...
-                    else:
-                        raise
-            else:
+            try:
+                self.id = self.call('open', self.location.path, create, lock_wait, lock, exclusive, append_only)
+            except self.RPCError as err:
+                if err.remote_type != 'TypeError':
+                    raise
+                if append_only:
+                    raise self.NoAppendOnlyOnServer()
```
enkore: k

borg/repository.py
```diff
@@ -207,14 +207,23 @@ def open_index(self, transaction_id):
...
-            # the repository instance lives on - even if exceptions happened.
-            self._active_txn = False
-            raise
+        if not self.lock.got_exclusive_lock():
+            if self.exclusive is not None:
+                # self.exclusive is either True or False, thus a new client is active here.
+                # if it is False and we get here, the caller did not use exclusive=True although
+                # it is needed for a write operation. if it is True and we get here, something else
+                # went very wrong, because we should have a exclusive lock, but we don't.
+                raise AssertionError("bug in code, exclusive lock should exist here")
```
enkore: ok

borg/repository.py
```diff
@@ -308,6 +317,8 @@ def complete_xfer():
         self.compact = set()
 
     def replay_segments(self, index_transaction_id, segments_transaction_id):
+        # fake an old client, so that in case we do not have an exclusive lock yet, prepare_txn will upgrade the lock:
+        self.exclusive = None
```
enkore: Shouldn't this be reset at the end?

borg/repository.py
```diff
@@ -207,14 +207,23 @@ def open_index(self, transaction_id):
...
-            raise
+        if not self.lock.got_exclusive_lock():
+            if self.exclusive is not None:
+                # self.exclusive is either True or False, thus a new client is active here.
+                # if it is False and we get here, the caller did not use exclusive=True although
+                # it is needed for a write operation. if it is True and we get here, something else
+                # went very wrong, because we should have a exclusive lock, but we don't.
+                raise AssertionError("bug in code, exclusive lock should exist here")
+            # if we are here, this is an old client talking to a new server (expecting lock upgrade).
+            # XXX doing lock upgrades to support old clients is deprecated.
```
enkore: overlooked in 8b113aeb163?

borg/testsuite/repository.py
enkore: This is now accounted for in my understanding.

borg/repository.py
ThomasWaldmann: yes, I'll fix.

borg/repository.py
ThomasWaldmann: prepare_txn will upgrade the lock.
either it will succeed, then we have an exclusive lock or it will fail with an exception.

so, for cleanliness we could remember and reset to previous value, but in practice I guess it makes no difference.

enkore: Hm, we do a rollback() in the `finally` clause, so a new transaction could be initiated by the client after replay_segments completes. Not sure if we do that anywhere, I don't think so.

ThomasWaldmann: ok, i added remembering/restoring self.exclusive state.

ThomasWaldmann:
collapse / merge?

enkore:
yes

textshell:
wrt collapse: I would at least keep the first two commits uncollapsed. Renames are alsways better in a separate commit. looks good now

Issue #1362, opened 2016-07-23
Tests failing on SSHFS


PlasmaPower:
All of these issues have been fixed in #1373.

- [ ] No ACLs: `KeyError: 'acl_access'` pops up a bunch - also present on ntfs-3g (see #1363)
- [ ] No FIFOs:
  ```
          # FIFO node
>       os.mkfifo(os.path.join(self.input_path, 'fifo1'))
E       PermissionError: [Errno 1] Operation not permitted

.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:290: PermissionError
  ```

- [ ] No unix sockets:
   ```
            sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
>       sock.bind(os.path.join(self.input_path, 'unix-socket'))
E       PermissionError: [Errno 1] Operation not permitted

.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:377: PermissionError
   ```

- [ ] No hardlinks:
   ```
                self.cmd('extract', self.repository_location + '::test', '--strip-components', '2')
>           assert os.stat('hardlink').st_nlink == 2
E           AssertionError: assert 1 == 2

.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:563: AssertionError
   ```

These should be fixed before #1359 is merged.

PlasmaPower:
To solve these issues, should we check the name of the FS or try to detect which features it supports?

enkore:
Feature detection is preferable imho.

PlasmaPower:
Of course, but it would be substantially harder, especially with #1363 with the permission and mode stuff.

Maybe have flags which we can manually set per FS name?

PlasmaPower:
Maybe we could test for support of ACLs, hard links, unix sockets, and FIFOs, but use the name to determine if permissions and modes are reasonable?

enkore:
Yeah, the errors you got are in create_test_files (or setUp) or so anyway, which also does the have_root detection.

I don't know any simple/specific way (a syscall) to get the FS name, so parsing should be conservative I think.

PlasmaPower:
@enkore I think we could parse `/proc/mounts` which shouldn't be too bad (it's meant to be parsed).

enkore:
Looking at the couple different systems here (solaris, oi, bsds) I'd say neither mount(8) nor /proc/mounts is a feasible idea.

[...]

PR #1373, opened 2016-07-25
Filesystem feature detection and test skipping

PlasmaPower:
So much for the easy tag on #1362.

Resolves all of #1362, and the ACL part of #1363. The rest of ntfs-3g will be fixed using an env variable as described [here](https://github.com/borgbackup/borg/issues/1362#issuecomment-234728068) (will be done in a separate PR).

The only non-testing change here is a try-catch around `os.utime`, as that throws an OSError on SSHFS when used on a symlink. However, if the utime block fails, it's not a big deal and we're fine to just skip over the issue.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -553,10 +553,14 @@ Number of files: {0.stats.nfiles}'''.format(
         else:
             # old archives only had mtime in item metadata
             atime = mtime
-        if fd:
-            os.utime(fd, None, ns=(atime, mtime))
-        else:
-            os.utime(path, None, ns=(atime, mtime), follow_symlinks=False)
+        try:
+            if fd:
+                os.utime(fd, None, ns=(atime, mtime))
+            else:
+                os.utime(path, None, ns=(atime, mtime), follow_symlinks=False)
+        except OSError:
+            # some systems don't support calling utime on a symlink
+            pass
         acl_set(path, item, self.numeric_owner)
         if 'bsdflags' in item:
             try:
--- a/src/borg/testsuite/__init__.py
+++ b/src/borg/testsuite/__init__.py
@@ -1,5 +1,6 @@
 from contextlib import contextmanager
 import filecmp
+import functools
 import os
 import posix
 import stat
@@ -7,6 +8,7 @@ import sys
 import sysconfig
 import tempfile
 import time
+import uuid
 import unittest
 
 from ..xattr import get_all
@@ -52,6 +54,67 @@ else:
 
 if sys.platform.startswith('netbsd'):
     st_mtime_ns_round = -4  # only >1 microsecond resolution here?
+
+
+@contextmanager
+def unopened_tempfile():
+    with tempfile.TemporaryDirectory() as tempdir:
+        yield os.path.join(tempdir, "file")
+
+
+@functools.lru_cache()
+def are_symlinks_supported():
+    with unopened_tempfile() as filepath:
+        try:
+            os.symlink('somewhere', filepath)
+            if os.lstat(filepath) and os.readlink(filepath) == 'somewhere':
+                return True
+        except OSError:
+            pass
+    return False
+
+
+@functools.lru_cache()
+def are_hardlinks_supported():
+    with unopened_tempfile() as file1path, unopened_tempfile() as file2path:
+        open(file1path, 'w').close()
+        try:
+            os.link(file1path, file2path)
+            stat1 = os.stat(file1path)
+            stat2 = os.stat(file2path)
+            if stat1.st_nlink == stat2.st_nlink == 2 and stat1.st_ino == stat2.st_ino:
+                return True
+        except OSError:
+            pass
+    return False
+
+
+@functools.lru_cache()
+def are_fifos_supported():
+    with unopened_tempfile() as filepath:
+        try:
+            os.mkfifo(filepath)
+            return True
+        except OSError:
+            return False
+
+
+@functools.lru_cache()
+def is_utime_fully_supported():
+    with unopened_tempfile() as filepath:
+        # Some filesystems (such as SSHFS) don't support utime on symlinks
+        if are_symlinks_supported():
+            os.symlink('something', filepath)
+        else:
+            open(filepath, 'w').close()
+        try:
+            os.utime(filepath, (1000, 2000), follow_symlinks=False)
+            new_stats = os.lstat(filepath)
+            if new_stats.st_atime == 1000 and new_stats.st_mtime == 2000:
+                return True
+        except OSError as err:
+            pass
+        return False
 
 
 class BaseTestCase(unittest.TestCase):
@@ -103,13 +166,16 @@ class BaseTestCase(unittest.TestCase):
                 d1[4] = None
             if not stat.S_ISCHR(d2[1]) and not stat.S_ISBLK(d2[1]):
                 d2[4] = None
-            # Older versions of llfuse do not support ns precision properly
-            if fuse and not have_fuse_mtime_ns:
-                d1.append(round(s1.st_mtime_ns, -4))
-                d2.append(round(s2.st_mtime_ns, -4))
-            else:
-                d1.append(round(s1.st_mtime_ns, st_mtime_ns_round))
-                d2.append(round(s2.st_mtime_ns, st_mtime_ns_round))
+            # If utime isn't fully supported, borg can't set mtime.
+            # Therefore, we shouldn't test it in that case.
+            if is_utime_fully_supported():
+                # Older versions of llfuse do not support ns precision properly
+                if fuse and not have_fuse_mtime_ns:
+                    d1.append(round(s1.st_mtime_ns, -4))
+                    d2.append(round(s2.st_mtime_ns, -4))
+                else:
+                    d1.append(round(s1.st_mtime_ns, st_mtime_ns_round))
+                    d2.append(round(s2.st_mtime_ns, st_mtime_ns_round))
             d1.append(get_all(path1, follow_symlinks=False))
             d2.append(get_all(path2, follow_symlinks=False))
             self.assert_equal(d1, d2)
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -36,6 +36,7 @@ from ..remote import RemoteRepository, PathNotAllowed
 from ..repository import Repository
 from . import has_lchflags, has_llfuse
 from . import BaseTestCase, changedir, environment_variable
+from . import are_symlinks_supported, are_hardlinks_supported, are_fifos_supported, is_utime_fully_supported
 
 src_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
 
@@ -274,10 +275,12 @@ class ArchiverTestCaseBase(BaseTestCase):
         # File mode
         os.chmod('input/file1', 0o4755)
         # Hard link
-        os.link(os.path.join(self.input_path, 'file1'),
-                os.path.join(self.input_path, 'hardlink'))
+        if are_hardlinks_supported():
+            os.link(os.path.join(self.input_path, 'file1'),
+                    os.path.join(self.input_path, 'hardlink'))
         # Symlink
-        os.symlink('somewhere', os.path.join(self.input_path, 'link1'))
+        if are_symlinks_supported():
+            os.symlink('somewhere', os.path.join(self.input_path, 'link1'))
         if xattr.is_enabled(self.input_path):
             xattr.setxattr(os.path.join(self.input_path, 'file1'), 'user.foo', b'bar')
             # XXX this always fails for me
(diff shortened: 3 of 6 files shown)


PlasmaPower:
@enkore Mind testing this on *BSD? I'm fairly confident that this PR won't break anything, but you never know.

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,83 @@
...
+        filepath = generate_tempfile_path()
+        # Some filesystems (such as SSHFS) don't support utime on symlinks
+        if are_symlinks_supported():
+            os.symlink('something', filepath)
+        else:
+            open(filepath, 'w').close()
+        try:
+            os.utime(filepath, None, ns=(1000, 2000))
+            new_stats = os.lstat(filepath)
+            if new_stats.st_atime == 1000 and new_stats.st_mtime == 2000:
```
ThomasWaldmann: as you gave nanoseconds above, you should also use .st_(a|m)time_ns here.

src/borg/testsuite/archiver.py
```diff
@@ -373,14 +380,21 @@ def filter(output):
...
-
+        try:
+            sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            sock.bind(os.path.join(self.input_path, 'unix-socket'))
+            self.cmd('create', self.repository_location + '::test', 'input')
+            sock.close()
+            with changedir('output'):
+                self.cmd('extract', self.repository_location + '::test')
+                assert not os.path.exists('input/unix-socket')
+        except PermissionError as err:
```
ThomasWaldmann: this is a rather wide-scope handler, includes all code in "borg create" and "borg extract".

src/borg/testsuite/archiver.py
```diff
@@ -373,14 +380,21 @@ def filter(output):
...
+            if err.errno == 1:  # Not permitted
+                pytest.skip('unix sockets disabled or not supported')
+            elif err.errno == 13:  # Permission denied
```
ThomasWaldmann: use errno.E....

ThomasWaldmann:
uh, yeah, that was a bit more than expected. thanks for your work!

src/borg/testsuite/archiver.py
enkore: If L384/385 fail with EPERM or EACESS then the test can be already skipped - while if that works then the rest *should* work as well. I hope :)

enkore:
> @enkore Mind testing this on *BSD? I'm fairly confident that this PR won't break anything, but you never know.

Will do

src/borg/testsuite/archiver.py
PlasmaPower: The reason I put it here is so that if some operation borg performs on the sockets isn't supported, it'll get caught. It's only catching `PermissionError`s after all, those are very unlikely to actually be borg's fault.

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,83 @@
...
 
 
+_symlinks_supported = None
+_fifos_supported = None
+_hardlinks_supported = None
+_utime_supported = None
+
+
+def generate_tempfile_path():
+    return os.path.join(tempfile.gettempdir(), 'borg-test-' + uuid.uuid4().hex)
```
PlasmaPower: Just realized that this could be replaced by `os.tempnam()`. I'll do that in the next push.

PlasmaPower: Nevermind, that was Python 2 only because now they want Python to create the file for you with `tempfile`.

PlasmaPower:
@enkore Could you check if it's still +2 skipped now that I fixed `is_utime_fully_supported`?

PlasmaPower:
Actually, nevermind, I can just test that locally of course.

src/borg/testsuite/archiver.py
PlasmaPower: Should I change this?

ThomasWaldmann: yes, the scope is definitely too wide now.
and nothing is worse than an exception handler covering up exceptions in code it was not meant for.

src/borg/testsuite/platform.py
```diff
@@ -44,6 +46,25 @@ def user_exists(username):
         return False
 
 
+def bad_acl_implementation():
+    global _bad_acl_implementation
```
ThomasWaldmann: could this have a non-negated / non-negative name?
like are_acls_supported/working?

PlasmaPower: Good idea. I'll change that.

src/borg/testsuite/archiver.py
PlasmaPower: Fixed.

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,85 @@
...
+def are_symlinks_supported():
+    global _symlinks_supported
+    if _symlinks_supported is None:
+        _symlinks_supported = False
+        filepath = generate_tempfile_path()
+        try:
+            os.symlink('somewhere', filepath)
+            if os.lstat(filepath) and os.readlink(filepath) == 'somewhere':
+                _symlinks_supported = True
+            os.unlink(filepath)
```
enkore: This should be in the `finally:` branch so that it cleans the file up regardless of symlink() errors?

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,85 @@
...
+
+
+def are_hardlinks_supported():
+    global _hardlinks_supported
+    if _hardlinks_supported is None:
+        _hardlinks_supported = False
+        file1 = tempfile.NamedTemporaryFile()
+        file2path = generate_tempfile_path()
+        try:
+            os.link(file1.name, file2path)
```
enkore: NamedTemporaryFile() ... `.name` may not work here in general (on win32 - can't open the file a second time, not sure about linking).

src/borg/testsuite/platform.py
```diff
@@ -44,6 +46,25 @@ def user_exists(username):
...
 
+def are_acls_working():
+    global _acls_working
+    if _acls_working is None:
+        _acls_working = False
+        try:
+            tmpfile = tempfile.NamedTemporaryFile()
+            access = b'user::rw-\ngroup::r--\nmask::rw-\nother::---\nuser:root:rw-:9999\ngroup:root:rw-:9999\n'
+            acl = {'acl_access': access}
+            acl_set(tmpfile.name, acl)
```
enkore: doesn't matter here, though; this is linux-only anyway..

PlasmaPower: Should I be worried about this for future tests though?

src/borg/testsuite/__init__.py
PlasmaPower: Good catch - fixed.

src/borg/testsuite/platform.py
PlasmaPower: I fixed it, because why not.

PlasmaPower:
Anything holding up this and #1383 from getting merged?

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,86 @@
...
+_fifos_supported = None
+_hardlinks_supported = None
+_utime_supported = None
+
+
+def generate_tempfile_path():
+    return os.path.join(tempfile.gettempdir(), 'borg-test-' + uuid.uuid4().hex)
+
+
+def are_symlinks_supported():
```
enkore: [`@functools.lru_cache()`](file:///usr/share/doc/python/html/library/functools.html) allows you to avoid both the global variables and checking/setting them.

enkore:
Besides that one item I think this is very good. Thanks!

src/borg/archive.py
```diff
@@ -553,10 +553,13 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
...
-            os.utime(fd, None, ns=(atime, mtime))
-        else:
-            os.utime(path, None, ns=(atime, mtime), follow_symlinks=False)
+        try:
+            if fd:
+                os.utime(fd, None, ns=(atime, mtime))
+            else:
+                os.utime(path, None, ns=(atime, mtime), follow_symlinks=False)
+        except OSError:
+            pass
```
ThomasWaldmann: add a short comment here why we have this?

src/borg/testsuite/__init__.py
```diff
@@ -54,6 +55,86 @@
...
 
 
+_symlinks_supported = None
+_fifos_supported = None
+_hardlinks_supported = None
+_utime_supported = None
+
+
+def generate_tempfile_path():
+    return os.path.join(tempfile.gettempdir(), 'borg-test-' + uuid.uuid4().hex)
```
ThomasWaldmann: looks like a home-grown version of tempfile.mktemp.

src/borg/testsuite/__init__.py
```diff
@@ -103,13 +184,16 @@ def _assert_dirs_equal_cmp(self, diff):
...
                 d2[4] = None
-            # Older versions of llfuse do not support ns precision properly
-            if fuse and not have_fuse_mtime_ns:
-                d1.append(round(s1.st_mtime_ns, -4))
-                d2.append(round(s2.st_mtime_ns, -4))
-            else:
-                d1.append(round(s1.st_mtime_ns, st_mtime_ns_round))
-                d2.append(round(s2.st_mtime_ns, st_mtime_ns_round))
+            # If utime isn't fully supported, borg can't set mtime
+            # Therefore, we shouldn't test it in that case
```
ThomasWaldmann: nitpicking: add dots at sentence ends?

src/borg/testsuite/platform.py
```diff
@@ -44,6 +46,27 @@ def user_exists(username):
...
+            acl = {'acl_access': access}
+            acl_set(tmpfile, acl)
+            read_acl = {}
+            acl_get(tmpfile, read_acl, os.stat(tmpfile))
+            read_acl_access = read_acl.get('acl_access', None)
+            if read_acl_access and b'user::rw-' in read_acl_access:
+                _acls_working = True
+            os.unlink(tmpfile)
+        except PermissionError:
+            pass
```
ThomasWaldmann: maybe move tmpfile creation before the "try".
move the tmpfile deletion to "finally".

src/borg/testsuite/platform.py
```diff
@@ -63,6 +86,7 @@ def set_acl(self, path, access=None, default=None, numeric_owner=False):
         item = {'acl_access': access, 'acl_default': default}
         acl_set(path, item, numeric_owner=numeric_owner)
 
+    @unittest.skipIf(not are_acls_working(), 'bad filesystem ACL implementation')
```
ThomasWaldmann: change reason to "ACLs do not work"?

it does not need to be a bad implementation, it can also be just not implemented or mounted without acl support.

ThomasWaldmann: + 3 more times, see below.

ThomasWaldmann:
few minor things, other than that it lgtm.

src/borg/testsuite/__init__.py
PlasmaPower: mktemp also opens the file (creating it). We need to not have a lock on the file, and a lot of tests also use it for non-regular files (unix sockets, hardlinks, symlinks, etc). AFAIK there's no good way to get just a file name, which is why I created this.

src/borg/testsuite/__init__.py
PlasmaPower: That's really helpful, thanks!

src/borg/testsuite/__init__.py
enkore: Q: Does the same caveat that NamedTemporaryFile has apply to mkstemp?

If so, it may be better to just use TemporaryDirectory and create a file with a static name in there instead. As a context manager it also handles cleanup with no extra code.

ThomasWaldmann:
```
>>> from tempfile import mktemp
>>> mktemp()
'/tmp/tmpo9wai7j1'
>>> from os import stat
>>> stat('/tmp/tmpo9wai7j1')
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpo9wai7j1'
```

enkore:
Another bit occured to me, there are quite a bit of conditional skips here. Maybe mention these (are_acls_working, symlinks_supported, fifos_supported, ...) in the py.test header (possibly in an abbreviated format, like `Testing: symlinks, fifos`). The header is [generated here](https://github.com/borgbackup/borg/blob/master/conftest.py#L16).

src/borg/testsuite/__init__.py
PlasmaPower: Oh, nevermind, I was looking at `mkstemp`. `mktemp` is deprecated, should I use it anyway? It does what I'm looking for.

PlasmaPower: I'm going to use `mktemp` because it was deprecated for security reasons which we don't need to worry about for this. Besides, this custom function has the same problems.

enkore: Why not something along the lines of

```py
@contextlib.contextmanager
def reopenable_tempfile():
    with tempfile.TemporaryDirectory() as tempdir:
         yield os.path.join(tempdir, "file")
```

PlasmaPower: That's a great idea, it will clean up after itself too!

ThomasWaldmann: btw, iirc pytest has a fixture for such stuff (with auto-cleanup).

We can only use this in pytest-based tests though.

enkore: These feature-detection functions aren't fixtures (yet?), though

PlasmaPower:
I've fixed all of the issues mentioned, including @enkore's suggestion about the py.test header.

PlasmaPower:
#1383 has been merged into this to include it in the `py.test` header.

PlasmaPower:
Please hold off merging this, I might have broken things.

PlasmaPower:
You can go ahead with merging this, it looks like the CIFS issues were not the result of this PR.

ThomasWaldmann:
thanks for your work! merging...

PR #1374, opened 2016-07-25
Fix AES-CTR nonce reusage

### LINKED ISSUE #22: IVs (nonces) generation
Current borg code uses AES in CTR mode, which requires an IV that never repeats.

That can be either a COUNTER (thus the "CTR" in the mode name) or a RANDOM value.

COUNTER: this is the way the current code works: load a counter from repo manifest, encrypt stuff (and keep incrementing the counter), at the end: save the counter to repo manifest.
Pro:
a) no birthday paradox
b) no collisions / repetition - IF correctly managed

Problems:
a) if a backup crashes and the manifest is not written, the next backup will reuse same counter start value - and that is a encryption security issue.
b) if encryption code runs in parallel (e.g. threads), it needs management of the different counter start values for the different workers.

RANDOM: future way of doing things?

Pro:
a) no "management" of counter start values, no problem for parallel execution
b) no storage of highest counter needed

Problems:
a) needs good randomness source (we also need that for key generation already)
b) source needs to keep up with demand

Needs many random bits to avoid collisions ("birthday problem"), NIST recommends a fixed value + lots of random bits as IV.

Related idea: create a new encryption key per chunk and always start CTR from 0.

Related idea: create a new encryption key per backup (if multithreaded: per crypter worker thread) and start ctr from 0 in each backup (in each worker thread).

https://www.bountysource.com/issues/16559386-ivs-nonces-generation

textshell:
Questions:
- [x] What to do about local locking?
- [x] What to do about repository side locking? Just depend on #1371? -> yes.
- [x] If so, add assert that repo is locked
- [x] What space reservation should be used?
- [x] Should the reservation be reduced on final commit to avoid nonce space wastage? -> No don’t worry about this
- [x] testing
- [x] diagnostic tools?
- [x] Remove commit with debugging hacks
- [x] What is needed with regard to leaking exceptions from file io?

    Reserve nonce space for AES-CTR before using it.
    
    Reusing the nonce totally breaks AES-CTR confidentiality.
    
    This code uses a reservation of nonce space and stores the next nonce available for
    a future reservation on the client and in the repository.
    
    Local storage is needed to protect about evil repositories that try to gain access to
    encrypted data by not saving nonce reservations and aborting the connection or otherwise
    forcing a rollback.
    
    Storage in the repository is needed to protect against another client writing to the
    repository after a transaction was aborted and thus not seeing the last used nonce from
    the manifest.

    With a real counter mode cipher protection for the multiple client case with an actively 
    evil repository is not possible. But this still protects against cases where the attacker can
    not arbitrarily change the repository but can read everything stored and abort connections
    or crash the server.

    Fixes #22

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -269,6 +269,17 @@ def get_keys_dir():
     return keys_dir
 
 
+def get_nonces_dir():
+    """Determine where to store the local nonce high watermark"""
+
+    xdg_config = os.environ.get('XDG_CONFIG_HOME', os.path.join(get_home_dir(), '.config'))
+    nonces_dir = os.environ.get('BORG_NONCES_DIR', os.path.join(xdg_config, 'borg', 'key-nonces'))
+    if not os.path.exists(nonces_dir):
+        os.makedirs(nonces_dir)
+        os.chmod(nonces_dir, stat.S_IRWXU)
+    return nonces_dir
+
+
 def get_cache_dir():
     """Determine where to repository keys and cache"""
     xdg_cache = os.environ.get('XDG_CACHE_HOME', os.path.join(get_home_dir(), '.cache'))
--- a/src/borg/key.py
+++ b/src/borg/key.py
@@ -3,7 +3,7 @@ import getpass
 import os
 import sys
 import textwrap
-from binascii import a2b_base64, b2a_base64, hexlify
+from binascii import a2b_base64, b2a_base64, hexlify, unhexlify
 from hashlib import sha256, pbkdf2_hmac
 from hmac import compare_digest
 
@@ -23,6 +23,7 @@ from .helpers import bin_to_hex
 from .helpers import CompressionDecider2, CompressionSpec
 from .item import Key, EncryptedKey
 from .platform import SaveFile
+from .nonces import NonceManager
 
 
 PREFIX = b'\0' * 8
@@ -169,6 +170,7 @@ class AESKeyBase(KeyBase):
 
     def encrypt(self, chunk):
         chunk = self.compress(chunk)
+        self.nonce_manager.ensure_reservation(num_aes_blocks(len(chunk.data)))
         self.enc_cipher.reset()
         data = b''.join((self.enc_cipher.iv[8:], self.enc_cipher.encrypt(chunk.data)))
         hmac = hmac_sha256(self.enc_hmac_key, data)
@@ -207,8 +209,9 @@ class AESKeyBase(KeyBase):
         if self.chunk_seed & 0x80000000:
             self.chunk_seed = self.chunk_seed - 0xffffffff - 1
 
-    def init_ciphers(self, enc_iv=b''):
-        self.enc_cipher = AES(is_encrypt=True, key=self.enc_key, iv=enc_iv)
+    def init_ciphers(self, manifest_nonce=0):
+        self.enc_cipher = AES(is_encrypt=True, key=self.enc_key, iv=manifest_nonce.to_bytes(16, byteorder='big'))
+        self.nonce_manager = NonceManager(self.repository, self.enc_cipher, manifest_nonce)
         self.dec_cipher = AES(is_encrypt=False, key=self.enc_key)
 
 
@@ -299,7 +302,7 @@ class PassphraseKey(AESKeyBase):
             try:
                 key.decrypt(None, manifest_data)
                 num_blocks = num_aes_blocks(len(manifest_data) - 41)
-                key.init_ciphers(PREFIX + long_to_bytes(key.extract_nonce(manifest_data) + num_blocks))
+                key.init_ciphers(key.extract_nonce(manifest_data) + num_blocks)
                 return key
             except IntegrityError:
                 passphrase = Passphrase.getpass(prompt)
@@ -337,7 +340,7 @@ class KeyfileKeyBase(AESKeyBase):
             if not key.load(target, passphrase):
                 raise PassphraseWrong
         num_blocks = num_aes_blocks(len(manifest_data) - 41)
-        key.init_ciphers(PREFIX + long_to_bytes(key.extract_nonce(manifest_data) + num_blocks))
+        key.init_ciphers(key.extract_nonce(manifest_data) + num_blocks)
         return key
 
     def find_key(self):
--- a/src/borg/nonces.py
+++ b/src/borg/nonces.py
@@ -1,0 +1,87 @@
+import os
+import sys
+from binascii import unhexlify
+
+from .crypto import bytes_to_long, long_to_bytes
+from .helpers import get_nonces_dir
+from .helpers import bin_to_hex
+from .platform import SaveFile
+from .remote import InvalidRPCMethod
+
+
+MAX_REPRESENTABLE_NONCE = 2**64 - 1
+NONCE_SPACE_RESERVATION = 2**28  # This in units of AES blocksize (16 bytes)
+
+
+class NonceManager:
+    def __init__(self, repository, enc_cipher, manifest_nonce):
+        self.repository = repository
+        self.enc_cipher = enc_cipher
+        self.end_of_nonce_reservation = None
+        self.manifest_nonce = manifest_nonce
+        self.nonce_file = os.path.join(get_nonces_dir(), self.repository.id_str)
+
+    def get_local_free_nonce(self):
+        try:
+            with open(self.nonce_file, 'r') as fd:
+                return bytes_to_long(unhexlify(fd.read()))
+        except FileNotFoundError:
+            return None
+
+    def commit_local_nonce_reservation(self, next_unreserved, start_nonce):
+        if self.get_local_free_nonce() != start_nonce:
+            raise Exception("nonce space reservation with mismatched previous state")
+        with SaveFile(self.nonce_file, binary=False) as fd:
+            fd.write(bin_to_hex(long_to_bytes(next_unreserved)))
+
+    def get_repo_free_nonce(self):
+        try:
+            return self.repository.get_free_nonce()
+        except InvalidRPCMethod as error:
+            # old server version, suppress further calls
+            sys.stderr.write("Please upgrade to borg version 1.1+ on the server for safer AES-CTR nonce handling.\n")
+            self.get_repo_free_nonce = lambda: None
+            self.commit_repo_nonce_reservation = lambda next_unreserved, start_nonce: None
+            return None
+
+    def commit_repo_nonce_reservation(self, next_unreserved, start_nonce):
+        self.repository.commit_nonce_reservation(next_unreserved, start_nonce)
+
+    def ensure_reservation(self, nonce_space_needed):
+        # Nonces may never repeat, even if a transaction aborts or the system crashes.
+        # Therefore a part of the nonce space is reserved before any nonce is used for encryption.
+        # As these reservations are commited to permanent storage before any nonce is used, this protects
+        # against nonce reuse in crashes and transaction aborts. In that case the reservation still
+        # persists and the whole reserved space is never reused.
+        #
+        # Local storage on the client is used to protect against an attacker that is able to rollback the
+        # state of the server or can do arbitrary modifications to the repository.
+        # Storage on the server is used for the multi client use case where a transaction on client A is
+        # aborted and later client B writes to the repository.
+        #
+        # This scheme does not protect against attacker who is able to rollback the state of the server
+        # or can do arbitrary modifications to the repository in the multi client usecase.
+
+        if self.end_of_nonce_reservation:
+            # we already got a reservation, if nonce_space_needed still fits everything is ok
+            next_nonce = int.from_bytes(self.enc_cipher.iv, byteorder='big')
+            assert next_nonce <= self.end_of_nonce_reservation
+            if next_nonce + nonce_space_needed <= self.end_of_nonce_reservation:
+                return
+
+        repo_free_nonce = self.get_repo_free_nonce()
+        local_free_nonce = self.get_local_free_nonce()
+        free_nonce_space = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce, self.end_of_nonce_reservation) if x is not None)
+        reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
+        assert reservation_end < MAX_REPRESENTABLE_NONCE
+        if self.end_of_nonce_reservation is None:
+            # initialization, reset the encryption cipher to the start of the reservation
+            self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
+        else:
+            # expand existing reservation if possible
+            if free_nonce_space != self.end_of_nonce_reservation:
+                # some other client got an interleaved reservation, skip partial space in old reservation to avoid overlap
+                self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
+        self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+        self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+        self.end_of_nonce_reservation = reservation_end
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -66,6 +66,8 @@ class RepositoryServer:  # pragma: no cover
         'save_key',
         'load_key',
         'break_lock',
+        'get_free_nonce',
+        'commit_nonce_reservation'
     )
 
     def __init__(self, restrict_to_paths, append_only):
@@ -449,6 +451,12 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
 
     def load_key(self):
         return self.call('load_key')
+
+    def get_free_nonce(self):
+        return self.call('get_free_nonce')
+
+    def commit_nonce_reservation(self, next_unreserved, start_nonce):
+        return self.call('commit_nonce_reservation', next_unreserved, start_nonce)
 
     def break_lock(self):
         return self.call('break_lock')
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -189,6 +189,27 @@ class Repository:
         keydata = self.config.get('repository', 'key')
         return keydata.encode('utf-8')  # remote repo: msgpack issue #99, returning bytes
 
+    def get_free_nonce(self):
+        if not self.lock.got_exclusive_lock():
+            raise AssertionError("bug in code, exclusive lock should exist here")
+
+        nonce_path = os.path.join(self.path, 'nonce')
+        try:
+            with open(nonce_path, 'r') as fd:
+                return int.from_bytes(unhexlify(fd.read()), byteorder='big')
+        except FileNotFoundError:
+            return None
+
+    def commit_nonce_reservation(self, next_unreserved, start_nonce):
+        if not self.lock.got_exclusive_lock():
+            raise AssertionError("bug in code, exclusive lock should exist here")
+
+        if self.get_free_nonce() != start_nonce:
+            raise Exception("nonce space reservation with mismatched previous state")
+        nonce_path = os.path.join(self.path, 'nonce')
+        with SaveFile(nonce_path, binary=False) as fd:
+            fd.write(bin_to_hex(next_unreserved.to_bytes(8, byteorder='big')))
+
     def destroy(self):
         """Destroy the repository at `self.path`
         """
--- a/src/borg/testsuite/key.py
+++ b/src/borg/testsuite/key.py
@@ -1,6 +1,7 @@
 import getpass
 import re
 import tempfile
+import os.path
 from binascii import hexlify, unhexlify
 
 import pytest
@@ -9,6 +10,7 @@ from ..crypto import bytes_to_long, num_aes_blocks
 from ..helpers import Location
 from ..helpers import Chunk
 from ..helpers import IntegrityError
+from ..helpers import get_nonces_dir
 from ..key import PlaintextKey, PassphraseKey, KeyfileKey, Passphrase, PasswordRetriesExceeded, bin_to_hex
 
 
@@ -16,6 +18,11 @@ from ..key import PlaintextKey, PassphraseKey, KeyfileKey, Passphrase, PasswordR
 def clean_env(monkeypatch):
     # Workaround for some tests (testsuite/archiver) polluting the environment
     monkeypatch.delenv('BORG_PASSPHRASE', False)
+
+
+@pytest.fixture(autouse=True)
+def nonce_dir(tmpdir_factory, monkeypatch):
+    monkeypatch.setenv('XDG_CONFIG_HOME', tmpdir_factory.mktemp('xdg-config-home'))
 
 
 class TestKey:
@@ -59,6 +66,12 @@ class TestKey:
         id = bytes(32)
         id_str = bin_to_hex(id)
 
+        def get_free_nonce(self):
+            return None
+
+        def commit_nonce_reservation(self, next_unreserved, start_nonce):
+            pass
+
     def test_plaintext(self):
         key = PlaintextKey.create(None, None)
         chunk = Chunk(b'foo')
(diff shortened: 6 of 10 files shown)


src/borg/key.py
```diff
@@ -154,13 +156,82 @@ class AESKeyBase(KeyBase):
...
+    def get_repo_free_nonce(self):
+        try:
+            repo_free_nonce = self.repository.get_free_nonce()
+        except InvalidRPCMethod as error:
+            # old server version, suppress further calls
+            self.get_repo_free_nonce = lambda: None
+            self.save_repo_end_of_nonce_reservation = lambda next_unreserved, start_nonce: None
+            return None
+        if repo_free_nonce is not None:
+            repo_free_nonce = bytes_to_long(repo_free_nonce)
```
ThomasWaldmann: why not directly return the binary nonce from the repo method? would correspond better to get_local_free_nonce also.

src/borg/key.py
```diff
@@ -154,13 +156,82 @@ class AESKeyBase(KeyBase):
...
+    def ensure_reservation(self, nonce_space_needed):
+        if not self.end_of_nonce_reservation:
+            repo_free_nonce = self.get_repo_free_nonce()
+            local_free_nonce = self.get_local_free_nonce()
+            next_nonce = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce) if x is not None)
+            reservation_end = next_nonce + nonce_space_needed + NONCE_SPACE_RESERVATION
+            self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+            self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+            self.end_of_nonce_reservation = reservation_end
+            if self.enc_cipher.iv != PREFIX + long_to_bytes(next_nonce):
```
ThomasWaldmann: do you really need this read-back of the iv?

src/borg/key.py
```diff
@@ -154,13 +156,82 @@ class AESKeyBase(KeyBase):
...
+                self.enc_cipher.reset(self.enc_key, PREFIX + long_to_bytes(next_nonce))
+        else:
+            next_nonce = bytes_to_long(self.enc_cipher.iv[8:])
```
ThomasWaldmann: hmm, here it is also reading back the iv.

i am asking because reading back the iv was a bit dirty in openssl 1.0 and does not work any more in 1.1.
also, it doesn't work for gcm, iirc.

src/borg/key.py
```diff
@@ -154,13 +156,82 @@ class AESKeyBase(KeyBase):
...
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                else:
+                    # need an entirly new reservation
+                    reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
+                    self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                    self.enc_cipher.reset(self.enc_key, PREFIX + long_to_bytes(next_nonce))
+
```
ThomasWaldmann: somehow thinking whether this whole nonce mgmt should go into a NonceManager class.

src/borg/repository.py
```diff
@@ -173,6 +173,21 @@ def load_key(self):
         keydata = self.config.get('repository', 'key')
         return keydata.encode('utf-8')  # remote repo: msgpack issue #99, returning bytes
 
+    def get_free_nonce(self):
+        nonce_path = os.path.join(self.path, 'nonce')
+        if os.path.exists(nonce_path):
```
ThomasWaldmann: how about instead exists() + open() just catching the exception if it does not open, because file is missing?

src/borg/key.py
textshell: Mostly because i didn’t want to import crypto into the repo class. And borg for some reason seems to not use int.from_bytes. If you prefer i can use int.from_bytes in repository.py and keep an equal interface. I think we still want the wrapper in AESKeyBase/NonceManager for catching InvalidRPCMethod.

src/borg/repository.py
textshell: Yes, that might be cleaner in python.

src/borg/key.py
textshell: Yes, we can just unconditionally reset the cipher, should no harm anything.

src/borg/key.py
textshell: I prefer to use int, as all used need to do arithmetic and having the call sites do the conversion while handling None is messy. Unless you see some problem with sending uint64 sized values as int via msgpack.

ThomasWaldmann: there was some strangeness with msgpack and long ints, have to check how long...

src/borg/key.py
```diff
@@ -18,14 +18,20 @@
...
+from .helpers import get_keys_dir, get_key_nonces_dir
 from .helpers import bin_to_hex
 from .helpers import CompressionDecider2, CompressionSpec
 from .item import Key, EncryptedKey
+from .platform import SaveFile
+from .remote import InvalidRPCMethod
 
 
 PREFIX = b'\0' * 8
+NONCE_SPACE_RESERVATION = 1000
```
ThomasWaldmann: is that for 1000 * 16 bytes data to encrypt?

guess we rather go to 2**28 or so.

textshell: Yes, i think we want this bigger. But going all the way to 4GB seems to be quite much. The way this currently works is that even a delete will consume 4GB of nonce space. Not sure that’s a problem though. 2**64 is a big number. Still gives 2**32 deletes till the nonce space is exhausted.

src/borg/key.py
```diff
@@ -137,6 +143,75 @@ def decrypt(self, id, data):
...
+            repo_free_nonce = self.get_repo_free_nonce()
+            local_free_nonce = self.get_local_free_nonce()
+            next_nonce = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce) if x is not None)
+            reservation_end = next_nonce + nonce_space_needed + NONCE_SPACE_RESERVATION
+            self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+            self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+            self.end_of_nonce_reservation = reservation_end
+            self.enc_cipher.reset(None, PREFIX + long_to_bytes(next_nonce))
+        else:
+            next_nonce = bytes_to_long(self.enc_cipher.iv[8:])
```
ThomasWaldmann: why throw away 8 bytes?

src/borg/key.py
```diff
@@ -137,6 +143,75 @@ def decrypt(self, id, data):
...
+                    self.end_of_nonce_reservation = reservation_end
+                else:
+                    # need an entirly new reservation
+                    print("needed entirly new reservation")
+                    print(repo_free_nonce, local_free_nonce, self.manifest_nonce, self.end_of_nonce_reservation)
+                    reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
+                    self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                    self.enc_cipher.reset(None, PREFIX + long_to_bytes(next_nonce))
```
ThomasWaldmann: see above. we could just do the computations with 16bytes.

src/borg/key.py
textshell: Yes, in that case it would also be easier to assert that the nonce didn‘t grow to large for the chunk format to encode. I copied the usage from other parts where they did make more sense.

But i like to keep the external storage formated as 8 bytes, because that is the usable space for the legacy encryption.

enkore:
> Depends on encore/issue/1060 for SaveFile

testing that right now, if I don't find any issues and CR-wise no problems we can just merge it with this.

ThomasWaldmann:
travis is a bit unhappy currently.

textshell:
I hope the existing tests run now. I still have to add new ones.

src/borg/key.py
```diff
@@ -154,13 +226,17 @@ class AESKeyBase(KeyBase):
 
     PAYLOAD_OVERHEAD = 1 + 32 + 8  # TYPE + HMAC + NONCE
 
+    def __init__(self, repository):
+        super(AESKeyBase, self).__init__(repository)
```
ThomasWaldmann: looks like this is not needed any more.

btw, python3 has argumentless super() for the usual case.

src/borg/remote.py
```diff
@@ -56,6 +56,8 @@ class RepositoryServer:  # pragma: no cover
         'save_key',
         'load_key',
         'break_lock',
+        'get_free_nonce',
+        'commit_nonce_reservation'
```
ThomasWaldmann: hmm, just an idea:

could this be also stored into the manifest (and loaded / saved with the manifest) instead of adding new api functions to the repo?

textshell: No, that won’t work. We need to be able to commit the reservation independently of normal repository commiting. And we need to be able to save the reservation without using the main encryption key.
So don’t think there is any way to include it into the normal chunk storage. And i don‘t think writing it into the config file would be any better (still needs new methods, it really doesn’t belong there). The semantics just don’t match with what we currently have.

src/borg/key.py
textshell: Right, there was a better syntax. I’ll fix that.

src/borg/remote.py

textshell:
I finally got around to rebase to current master and finish tests for this change. This also includes two fixes found while testing.

While the test cases are not as neat as i would like, they should have a fairly good coverage both line wise as well as testing the relevant situations where this protection is needed.

Still open is what kind of locking is needed on the client side. Usually the cache should be open when data is encrypted so that could be ok as lock. As far as i can see all of the usual commands that do encryption have an open cache. Maybe some one time conversion methods don‘t.

We need to finally decide if the iv space reservation size currently used is ok. It‘s quite big so it should be very well performance wise. On the other side a simple delete consumes IV space that otherwise could have been used to store 4GB data to the repository. But still the IV space is good for 68G small operations (i.e. writing less than 4GB), so no real worries there.

I originally thought it would be good to have something to check if a repository had any nonce reusage problems. But i don‘t that is really needed. The realistic case where reuse is possible is with rollbacks, which can‘t be detected from the repository state later anyway.

src/borg/repository.py
```diff
@@ -190,6 +190,9 @@ def load_key(self):
         return keydata.encode('utf-8')  # remote repo: msgpack issue #99, returning bytes
 
     def get_free_nonce(self):
+        if not self.lock.got_exclusive_lock():
+            raise AssertionError("bug in code, exclusive lock should exist here")
```
ThomasWaldmann: hmm, why not: assert got_lock, "bug ..."

src/borg/key.py
```diff
@@ -197,7 +197,7 @@ def ensure_reservation(self, nonce_space_needed):
         else:
             next_nonce = int.from_bytes(self.enc_cipher.iv, byteorder='big')
             assert(next_nonce <= self.end_of_nonce_reservation)
```
ThomasWaldmann: assert is not a function call, but it looks like one here.

src/borg/repository.py
textshell: I copied from prepare_txn, but yes, here the situation is simpler so i’ll just use a normal assert. Do we worry about python removing these checks in optimize mode? That would be the only reason to keep it as explicit raise.

enkore:
Re. reservation size: it could maybe be smaller, but it _doesn't really matter_ the way you implemented it, since either you'll encrypt lots of data, then it doesn't matter, since the nonce space will be actually used, or you don't meaning some maintenance op like delete or prune is done, where normally you'd expect less than one reservation used, so yeah, billions of ops and/or tons of data encrypted. So, I think this is ok.

src/borg/testsuite/key.py
```diff
@@ -77,13 +94,23 @@ def test_keyfile(self, monkeypatch, keys_dir):
...
         assert len({key2.id_key, key2.enc_key, key2.enc_hmac_key}) == 3
         assert key2.chunk_seed != 0
         chunk = Chunk(b'foo')
         assert chunk == key2.decrypt(key.id_hash(chunk.data), key.encrypt(chunk))
 
+    def test_keyfile_nonce_rollback_protection(self, monkeypatch, keys_dir):
+        monkeypatch.setenv('BORG_PASSPHRASE', 'test')
+        repository = self.MockRepository()
+        with open(os.path.join(get_key_nonces_dir(), repository.id_str), "w") as fd:
+            assert fd.write("0000000000002000")
```
enkore: not really meaningful assert

textshell: i’ll remove that.

src/borg/testsuite/key.py
```diff
@@ -232,3 +248,225 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+        id = bytes(32)
+        id_str = bin_to_hex(id)
+
+        def get_free_nonce(self):
+            return self.next_free
+
+        def commit_nonce_reservation(self, next_unreserved, start_nonce):
+            assert start_nonce == self.next_free
+            self.next_free = next_unreserved
+            pass
```
ThomasWaldmann: remove the pass?

src/borg/repository.py
enkore: raise is the proper way to go here imho, it's not a casual "oh btw." assert, but something we wouldn't want to lose just because someone (user, packager, admin, ...) thought that -O globally is a good idea to improve performance by 0.1 % ;D

src/borg/testsuite/key.py
```diff
@@ -232,3 +248,225 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+
+        manager = NonceManager(self.repository, enc_cipher, 0x2000)
+        manager.ensure_reservation(20)
+        enc_cipher.expect_iv_and_advance(0x2000, 0x2000 + 20)
+
+        assert self.cache_nonce() == "0000000000002034"
+        assert self.repository.next_free == 0x2034
+
+    def test_transaction_abort_old_server(self, monkeypatch):
+        monkeypatch.setattr(borg.key, 'NONCE_SPACE_RESERVATION', 32)
```
ThomasWaldmann: the hex / dec mix here is a bit unfortunate. esp. if you use 32 dec and 20 dec, it might confuse more than with other values.

src/borg/testsuite/repository.py
```diff
@@ -390,6 +390,48 @@ def test_additional_free_space(self):
...
+                self.repository.commit_nonce_reservation(200, 15)
+
+            self.repository.commit_nonce_reservation(200, None)
+            with open(os.path.join(self.repository.path, "nonce"), "r") as fd:
+                assert fd.read() == "00000000000000c8"
+
+            with pytest.raises(Exception):
+                self.repository.commit_nonce_reservation(200, 15)
+
+            self.repository.commit_nonce_reservation(512, 200)
```
ThomasWaldmann: similar here: 512 dec and 200 dec.

src/borg/testsuite/key.py
textshell: i can use hex everywhere. Most of the values are readable as hex too. Maybe 20 (dec) was not the greatest choice. But 0x14 isn‘t too bad.

src/borg/repository.py
ThomasWaldmann: ok, keep it.

src/borg/testsuite/key.py
ThomasWaldmann: i don't mean having to convert 20 to 0x14.
if mean using 32 dec (== 20 hex) and 20 dec.

src/borg/key.py
```diff
@@ -147,6 +150,75 @@ def decrypt(self, id, data, decompress=True):
...
+            repo_free_nonce = self.get_repo_free_nonce()
+            local_free_nonce = self.get_local_free_nonce()
+            next_nonce = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce) if x is not None)
+            reservation_end = next_nonce + nonce_space_needed + NONCE_SPACE_RESERVATION
+            assert reservation_end < MAX_REPRESENTABLE_NONCE
+            self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+            self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+            self.end_of_nonce_reservation = reservation_end
+            self.enc_cipher.reset(None, next_nonce.to_bytes(16, byteorder='big'))
+        else:
```
enkore: "if not ...: ... else:" <- un-invert the if and swap the two blocks?

src/borg/key.py
```diff
@@ -147,6 +150,75 @@ def decrypt(self, id, data, decompress=True):
...
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                else:
+                    # need an entirly new reservation
+                    reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
+                    assert reservation_end < MAX_REPRESENTABLE_NONCE
+                    self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                    self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
```
enkore: it looks to me like the three blocks above share quite a bit of logic and code, try to deduplicate that?

re = .., assert, self.crnr, self.clnr, self.eonr look identical, some differences beyond that...

src/borg/key.py
textshell: this order puts initialization before the normal code. But i’m happy to swap it around if you think it does make it more readable.

src/borg/testsuite/key.py
```diff
@@ -9,7 +10,12 @@
 from ..helpers import Location
 from ..helpers import Chunk
 from ..helpers import IntegrityError
+from ..helpers import get_key_nonces_dir
 from ..key import PlaintextKey, PassphraseKey, KeyfileKey, Passphrase, PasswordRetriesExceeded, bin_to_hex
+from ..key import NonceManager
+from ..remote import InvalidRPCMethod
+
+import borg.key  # for monkey patching NONCE_SPACE_RESERVATION
```
enkore: from .. import key?

textshell: For some reason i had problems with relative imports. But your suggesting works.
We could use

    from .. import key as key_module

Importing it as key would be quite confusing in the tests.

src/borg/helpers.py
```diff
@@ -269,6 +269,17 @@ def get_keys_dir():
     return keys_dir
 
 
+def get_key_nonces_dir():
+    """Determine where to store the local nonce high watermark"""
+
+    xdg_config = os.environ.get('XDG_CONFIG_HOME', os.path.join(get_home_dir(), '.config'))
+    nonces_dir = os.environ.get('BORG_KEY_NONCES_DIR', os.path.join(xdg_config, 'borg', 'key-nonces'))
```
ThomasWaldmann: hmm, why "key" nonces? just nonces?

textshell: My thinking was it’s where the nonces for the keyfiles/repokeys are stored. Also key does have the right associations for people who do not even know what a nonce is.

src/borg/key.py
enkore: ah, indeed. in that case it's probably better this way around

src/borg/key.py
```diff
@@ -147,6 +150,75 @@ def decrypt(self, id, data, decompress=True):
...
+            # old server version, suppress further calls
+            self.get_repo_free_nonce = lambda: None
+            self.commit_repo_nonce_reservation = lambda next_unreserved, start_nonce: None
+            return None
+
+    def commit_repo_nonce_reservation(self, next_unreserved, start_nonce):
+        self.repository.commit_nonce_reservation(next_unreserved, start_nonce)
+
+    def ensure_reservation(self, nonce_space_needed):
+        if not self.end_of_nonce_reservation:
```
ThomasWaldmann: ... is None:  # init

src/borg/key.py
textshell: with maximum deduplication it looks like this. It‘s quite a bit reordered, fast path up front.
I’m not quite sure if i like it like this. But maybe even if it‘s not easier to understand it might be easier to maintain.
```
    def ensure_reservation(self, nonce_space_needed):
        if self.end_of_nonce_reservation:
            # we already got a reservation, if nonce_space_needed still fits everything is ok
            next_nonce = int.from_bytes(self.enc_cipher.iv, byteorder='big')
            assert next_nonce <= self.end_of_nonce_reservation
            if next_nonce + nonce_space_needed <= self.end_of_nonce_reservation:
                return

        repo_free_nonce = self.get_repo_free_nonce()
        local_free_nonce = self.get_local_free_nonce()
        free_nonce_space = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce, self.end_of_nonce_reservation) if x is not None)
        reservation_end = free_nonce_space + nonce_space_needed + NONCE_SPACE_RESERVATION
        assert reservation_end < MAX_REPRESENTABLE_NONCE
        if self.end_of_nonce_reservation is None:  # init
            self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
        else:
            # try to expand reservation
            if free_nonce_space != self.end_of_nonce_reservation:
                # some other client got an interleaved reservation, skip partial space in old reservation to avoid overlap
                self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
        self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
        self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
        self.end_of_nonce_reservation = reservation_end
```

textshell: actually if the comments about init and expanding and interleaved are considered to be unimportant it could even be for the whole conditional following `if self.end_of_nonce_reservation is None`
```python
        if free_nonce_space != self.end_of_nonce_reservation:
            self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
```
But i think this it would quite a bit less meaningful. Maybe a big comment could help there, but...

src/borg/key.py
textshell: Yes, that‘t more to the point than the not. If we don‘t go with the one of the fast path first variants i will do this.

src/borg/key.py
```diff
@@ -147,6 +150,75 @@ def decrypt(self, id, data, decompress=True):
...
+                free_nonce_space = max(x for x in (repo_free_nonce, local_free_nonce, self.manifest_nonce, self.end_of_nonce_reservation) if x is not None)
+                if free_nonce_space == self.end_of_nonce_reservation:
+                    # nothing else got an reservation, just extend
+                    reservation_end = self.end_of_nonce_reservation + nonce_space_needed + NONCE_SPACE_RESERVATION
+                    assert reservation_end < MAX_REPRESENTABLE_NONCE
+                    self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+                    self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+                    self.end_of_nonce_reservation = reservation_end
+                else:
+                    # need an entirly new reservation
```
ThomasWaldmann: entirely?

textshell: yes.

ThomasWaldmann:
move the nonce code / tests to own modules?

textshell:
I'll move them to a noncemanager.py module.

src/borg/helpers.py
ThomasWaldmann: but the nonces are used to encrypt repo data, not to encrypt the key.

ThomasWaldmann:
or just nonces.py?

src/borg/helpers.py
textshell: Sure. It was meant as the nonce being a mutable property of the key and thus stored at distinct location but still logically related to the key.
If you prefer BORG_NONCES_DIR i’ll use that.

ThomasWaldmann:
`<nitpick>"its own module" </nitpick>`

src/borg/helpers.py
enkore: KEY_NONCES_DIR is good i think

ThomasWaldmann:
IMHO:
- don't merge into 1.0-maint - it is a rather big change
- merge into master now, so it gets into 1.1.0b1, collapse changesets before merging

textshell:
ok, i‘ve collapsed the changes into one commit and updated the commit message.
I also added a message when used with an older server because the remote code prints to stderr. When using the a new client with a old server it look like this now:

    borg delete ssh://martin@localhost/tmp/borg-test2::a15 
    Remote: RPC method get_free_nonce is not valid
    Please upgrade to borg version 1.1+ on the server for safer AES-CTR nonce handling.

src/borg/nonces.py
```diff
@@ -0,0 +1,87 @@
+import os, sys
```
enkore: pep8: split in two lines

src/borg/nonces.py
```diff
@@ -0,0 +1,87 @@
...
+            self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
+        else:
+            # expand existing reservation if possible
+            if free_nonce_space != self.end_of_nonce_reservation:
+                # some other client got an interleaved reservation, skip partial space in old reservation to avoid overlap
+                self.enc_cipher.reset(None, free_nonce_space.to_bytes(16, byteorder='big'))
+        self.commit_repo_nonce_reservation(reservation_end, repo_free_nonce)
+        self.commit_local_nonce_reservation(reservation_end, local_free_nonce)
+        self.end_of_nonce_reservation = reservation_end
+
```
enkore: pep8: blank line too much

enkore:
travis has two remarks (^), then this can be merged

textshell:
I hope this is fixed. My local flake8 check didn’t complain.

enkore:
Merged! Thanks! :)

textshell:
Thanks very much for helping along the way.

PR #1375, opened 2016-07-26
Don't raise an error if stdout/stderr are closed

### LINKED ISSUE #1116: BrokenPipeError leading to a crash when using filtering/counting shell utilities

PlasmaPower:
Fixes #1116

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1,6 +1,8 @@
 import argparse
+import errno
 import grp
 import hashlib
+import io
 import logging
 import os
 import os.path
@@ -1546,3 +1548,31 @@ def signal_handler(signo, handler):
         yield
     finally:
         signal.signal(signo, old_signal_handler)
+
+
+class ErrorIgnoringTextIOWrapper(io.TextIOWrapper):
+    def read(self, n):
+        if not self.closed:
+            try:
+                return super(ErrorIgnoringTextIOWrapper, self).read(n)
+            except OSError as err:
+                if err.errno != errno.EPIPE:
+                    raise
+                try:
+                    super(ErrorIgnoringTextIOWrapper, self).close()
+                except OSError as err:
+                    pass
+        return ''
+
+    def write(self, s):
+        if not self.closed:
+            try:
+                return super(ErrorIgnoringTextIOWrapper, self).write(s)
+            except OSError as err:
+                if err.errno != errno.EPIPE:
+                    raise
+                try:
+                    super(ErrorIgnoringTextIOWrapper, self).close()
+                except OSError as err:
+                    pass
+        return len(s)
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -41,6 +41,7 @@ from .helpers import dir_is_tagged, is_slow_msgpack, yes, sysinfo
 from .helpers import log_multi
 from .helpers import parse_pattern, PatternMatcher, PathPrefixPattern
 from .helpers import signal_handler
+from .helpers import ErrorIgnoringTextIOWrapper
 from .item import Item
 from .key import key_creator, RepoKey, PassphraseKey
 from .platform import get_flags
@@ -2251,8 +2252,8 @@ def main():  # pragma: no cover
 
     # Make sure stdout and stderr have errors='replace' to avoid unicode
     # issues when print()-ing unicode file names
-    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, sys.stdout.encoding, 'replace', line_buffering=True)
-    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, sys.stderr.encoding, 'replace', line_buffering=True)
+    sys.stdout = ErrorIgnoringTextIOWrapper(sys.stdout.buffer, sys.stdout.encoding, 'replace', line_buffering=True)
+    sys.stderr = ErrorIgnoringTextIOWrapper(sys.stderr.buffer, sys.stderr.encoding, 'replace', line_buffering=True)
     setup_signal_handlers()
     archiver = Archiver()
     msg = None


ThomasWaldmann:
stdout might transport file data (borg extract --stdout), so I don't think we want to silently suppress errors there in general.

PlasmaPower:
@ThomasWaldmann That's the only case though, correct? `borg extract` is also a read-only command, so if it fails it's not a big deal. What if someone wants to do `borg extract --stdout repo::example file.log | tail -n 100`? If it's `borg extract --stdout repo::example file.log > file.log`, and it runs out of disk space, then Bash will output an error saying that it's out of disk space. Whatever's on the other end of the pipe should output the error causing the closing of the pipe if it was closed unintentionally.

PlasmaPower:
Maybe `ErrorFree` is the wrong term, because it can still through a `ValueError` if the stream is already closed. I haven't seen that cause any problems though.

src/borg/helpers.py
```diff
@@ -1546,3 +1547,27 @@ def signal_handler(signo, handler):
...
+        return ''
+
+    def write(self, s):
+        if not self.closed:
+            try:
+                return super(ErrorFreeTextIOWrapper, self).write(s)
+            except IOError:
+                try:
+                    super(ErrorFreeTextIOWrapper, self).close()
+                except IOError:
```
ThomasWaldmann: One minor nitpick, quoting from the python 3.4 docs:
```
The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of OSError.

exception EnvironmentError

exception IOError

exception WindowsError
```

So, please use OSError at all these places.

src/borg/helpers.py
```diff
@@ -1546,3 +1547,27 @@ def signal_handler(signo, handler):
         yield
     finally:
         signal.signal(signo, old_signal_handler)
+
+
+class ErrorFreeTextIOWrapper(io.TextIOWrapper):
```
ThomasWaldmann: Maybe ErrorIgnoring... ?

src/borg/helpers.py
```diff
@@ -1546,3 +1547,27 @@ def signal_handler(signo, handler):
...
+    def read(self, n):
+        if not self.closed:
+            try:
+                return super(ErrorFreeTextIOWrapper, self).read(n)
+            except IOError:
```
ThomasWaldmann: shall we maybe ignore a bit more specifically?
like only the closed pipe error?

PlasmaPower:
So I'm now catching OSError, should I only catch `errno.EPIPE`?

PlasmaPower:
Okay, I'm now only catching `errno.EPIPE` but my code has sort of turned into a mess of conditionals. Not much can be done about it though.

PlasmaPower:
Checking if it's not equal instead simplifies thing a great deal. I think this is ready to merge, unless anyone has any additional comments.

src/borg/helpers.py
```diff
@@ -1546,3 +1548,33 @@ def signal_handler(signo, handler):
...
+        if not self.closed:
+            try:
+                return super(ErrorIgnoringTextIOWrapper, self).read(n)
+            except OSError as err:
+                if err.errno != errno.EPIPE:
+                    raise
+                try:
+                    super(ErrorIgnoringTextIOWrapper, self).close()
+                except OSError as err:
+                    if err.errno != errno.EPIPE:
```
ThomasWaldmann: can this happen?

src/borg/helpers.py
```diff
@@ -1546,3 +1548,33 @@ def signal_handler(signo, handler):
...
+        if not self.closed:
+            try:
+                return super(ErrorIgnoringTextIOWrapper, self).write(s)
+            except OSError as err:
+                if err.errno != errno.EPIPE:
+                    raise
+                try:
+                    super(ErrorIgnoringTextIOWrapper, self).close()
+                except OSError as err:
+                    if err.errno != errno.EPIPE:
```
ThomasWaldmann: same here

src/borg/helpers.py
PlasmaPower: I'm not sure. It's probably be safer just to ignore any OSError when closing anyway, the only reason I'm closing is to clean up. I've changed this.

ThomasWaldmann:
i don't think we have unit tests for this right?
so, did you test the latest code manually?

PlasmaPower:
@ThomasWaldmann Yep I tested `borg info repo::archive` and `borg info repo::archive | head -n 1`.

ThomasWaldmann:
ok, code lgtm - merge?

just one thing: #1116 is in 1.0.7 milestone and as we usually merge 1.0-maint into master anyway, PR against 1.0-maint would be easier to process.

PlasmaPower:
@ThomasWaldmann Sounds good - I'll close then then reapply the patch to 1.0-maint.

PR #1377, opened 2016-07-26
Use atomic file writing and updating for configuration and key files

### LINKED ISSUE #1060: Use atomic file writing and updating for configuration and key files
Proposed sequence

- write to temporary
- fsync(temp)
- rename(temp, real)
- fsync(containing_dir) <-- important for the repository config when we implement DEKs

***

:moneybag: [there is a bounty for this](https://github.com/borgbackup/borg/issues/1060)

enkore:
Renders the issues [described here](https://github.com/borgbackup/borg/issues/1060#issuecomment-229622961) unreproducible for me. Naturally the transaction may be a bit botched and needs recovery (e.g. "hints.tmp", index missing etc), but since the segments are solid so is the transaction.

Fixes #1060

### DIFF
--- a/src/borg/platform/base.py
+++ b/src/borg/platform/base.py
@@ -80,8 +80,11 @@ class SyncFile:
     TODO: A Windows implementation should use CreateFile with FILE_FLAG_WRITE_THROUGH.
     """
 
-    def __init__(self, path):
-        self.fd = open(path, 'xb')
+    def __init__(self, path, binary=True):
+        mode = 'x'
+        if binary:
+            mode += 'b'
+        self.fd = open(path, mode)
         self.fileno = self.fd.fileno()
 
     def __enter__(self):
@@ -110,6 +113,43 @@ class SyncFile:
         self.sync()
         self.fd.close()
         platform.sync_dir(os.path.dirname(self.fd.name))
+
+
+class SaveFile:
+    """
+    Update file contents atomically.
+
+    Must be used as a context manager (defining the scope of the transaction).
+
+    On a journaling file system the file contents are always updated
+    atomically and won't become corrupted, even on pure failures or
+    crashes (for caveats see SyncFile).
+    """
+
+    SUFFIX = '.tmp'
+
+    def __init__(self, path, binary=True):
+        self.binary = binary
+        self.path = path
+        self.tmppath = self.path + self.SUFFIX
+
+    def __enter__(self):
+        from .. import platform
+        try:
+            os.unlink(self.tmppath)
+        except OSError:
+            pass
+        self.fd = platform.SyncFile(self.tmppath, self.binary)
+        return self.fd
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        from .. import platform
+        self.fd.close()
+        if exc_type is not None:
+            os.unlink(self.tmppath)
+            return
+        os.rename(self.tmppath, self.path)
+        platform.sync_dir(os.path.dirname(self.path))
 
 
 def swidth(s):
--- a/src/borg/platform/base.py
+++ b/src/borg/platform/base.py
@@ -148,7 +148,7 @@ class SaveFile:
         if exc_type is not None:
             os.unlink(self.tmppath)
             return
-        os.rename(self.tmppath, self.path)
+        os.replace(self.tmppath, self.path)
         platform.sync_dir(os.path.dirname(self.path))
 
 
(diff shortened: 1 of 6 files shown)


src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
+    Must be used as a context manager (defining the scope of the transaction).
+
+    On a journaling file system the file contents are always updated
+    atomically and won't become corrupted, even on pure failures or
+    crashes (for caveats see SyncFile).
+    """
+
+    SUFFIX = '.tmp'
+
+    def __init__(self, path, binary=True):
```
enkore: I guess the default should be false to match the behaviour of open().

src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
 
 
+class SaveFile:
+    """
+    Update file contents atomically.
+
+    Must be used as a context manager (defining the scope of the transaction).
+
+    On a journaling file system the file contents are always updated
+    atomically and won't become corrupted, even on pure failures or
```
ThomasWaldmann: power

src/borg/platform/base.py
ThomasWaldmann: yes.

src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
+        try:
+            os.unlink(self.tmppath)
+        except OSError:
+            pass
+        self.fd = platform.SyncFile(self.tmppath, self.binary)
+        return self.fd
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        from .. import platform
+        self.fd.close()
```
ThomasWaldmann: in case there was an exception, can close() raise here?

src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
+        if exc_type is not None:
+            os.unlink(self.tmppath)
+            return
+        os.rename(self.tmppath, self.path)
+        platform.sync_dir(os.path.dirname(self.path))
```
ThomasWaldmann: do we need to expect exceptions from close, unlink, rename, sync_dir here?

enkore: Any FS operation can blow up on error, but for example sync_dir is something that shall not blow up just because it's not supported.

src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
+        self.fd = platform.SyncFile(self.tmppath, self.binary)
+        return self.fd
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        from .. import platform
+        self.fd.close()
+        if exc_type is not None:
+            os.unlink(self.tmppath)
+            return
+        os.rename(self.tmppath, self.path)
```
ThomasWaldmann: rename is known for cross-platform issues / different semantics.
e.g. on windows, renaming over an existing file will not work.

enkore: Indeed. os.replace should avoid the issue, but I'll check which API it actually uses.

src/borg/platform/base.py
```diff
@@ -80,8 +80,11 @@ class SyncFile:
     TODO: A Windows implementation should use CreateFile with FILE_FLAG_WRITE_THROUGH.
     """
 
-    def __init__(self, path):
-        self.fd = open(path, 'xb')
+    def __init__(self, path, binary=True):
+        mode = 'x'
+        if binary:
+            mode += 'b'
```
ThomasWaldmann: mode = 'xb' if binary else 'x'

src/borg/platform/base.py
```diff
@@ -112,6 +115,43 @@ def close(self):
...
+        self.binary = binary
+        self.path = path
+        self.tmppath = self.path + self.SUFFIX
+
+    def __enter__(self):
+        from .. import platform
+        try:
+            os.unlink(self.tmppath)
+        except OSError:
+            pass
```
ThomasWaldmann: maybe only ignore it if it does not exist.

src/borg/platform/base.py
ThomasWaldmann: yeah, os.replace is better. good we have 3.4+. :)

enkore:
merge?

ThomasWaldmann:
Yes, but we have to consider first, whether we rather want to have it in 1.0-maint.
Especially since the AES-IV PR depends on this.

enkore:
I think both of these PRs (#1377, #1374) would be better of to be first part of a 1.1 RC, where it is clear that it will be more unstable and may have more issues than an RC or point release in the 1.0 series.

ThomasWaldmann:
ok.

PR #1380, opened 2016-07-26
doc: quickstart on shell pitfalls for env. vars

jzacsh:
continues from chat in IRC

for quick reference, here's a direct link to this branch's rendered view (github-rendered): https://github.com/jzacsh/borg/blob/document-envvars-pitfalls/docs/quickstart.rst#pitfalls-with-shell-variables-and-environment-variables

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -129,6 +129,33 @@ certain number of old archives::
     borg prune -v $REPOSITORY --prefix '{hostname}-' \
         --keep-daily=7 --keep-weekly=4 --keep-monthly=6
 
+Pitfalls with shell variables and environment variables
+-------------------------------------------------------
+
+This applies to all environment variables you want borg to see, not just
+``BORG_PASSPHRASE``. The short explanation is: always ``export`` your variable,
+and use single quotes if you're unsure of the details of your shell's expansion
+behavior. E.g.::
+
+    export BORG_PASSPHRASE='complicated & long'
+
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+Beware of how ``sudo`` interacts with environment variables. For example, you
+may be surprised that the following ``export`` has no effect on your command::
+
+   export BORG_PASSPHRASE='complicated & long'
+   sudo ./yourborgwrapper.sh  # still prompts for password
+
+For more information, see sudo(8) man page. Hint: see ``env_keep`` in
+sudoers(5), or try ``sudo BORG_PASSPHRASE='yourphrase' borg`` syntax.
+
+.. Tip::
+    To debug what your borg process is actually seeing, find its PID
+    (``ps aux|grep borg``) and then look into ``/proc/<PID>/environ``.
+
 .. backup_compression:
 
 Backup compression


docs/quickstart.rst
```diff
@@ -129,6 +129,38 @@ certain number of old archives::
     borg prune -v $REPOSITORY --prefix '{hostname}-' \
         --keep-daily=7 --keep-weekly=4 --keep-monthly=6
 
+Pitfalls with shell variables and environment variables
+-------------------------------------------------------
+
+This applies to all environment variables you want borg to see, not just
+``BORG_PASSPHRASE``. The short explanation is: always ``export`` your variable,
+and use single quotes if you're unsure of the details of your shell's expansion
+behavior. eg::
```
ThomasWaldmann: E.g.::

docs/quickstart.rst
```diff
@@ -129,6 +129,38 @@ certain number of old archives::
...
+
+    export BORG_PASSPHRASE='your complicated & passphrase';
```
ThomasWaldmann: complicated & long

the ; at the end is not needed, is it?

docs/quickstart.rst
```diff
@@ -129,6 +129,38 @@ certain number of old archives::
...
+
+This ensures the borg PID that needs to access this passphrase can do so. Note
```
ThomasWaldmann: the borg process

docs/quickstart.rst
```diff
@@ -129,6 +129,38 @@ certain number of old archives::
...
+and use single quotes if you're unsure of the details of your shell's expansion
+behavior. E.g.::
+
+    export BORG_PASSPHRASE='your complicated & passphrase';
+
+This ensures the borg PID that needs to access this passphrase can do so. Note
+that this is not the same thing as::
+
+  $ BORG_PASSPHRASE='your complicated & passphrase'; # missing "export"!
+  $ ./yourborgjob.sh # will never see the above $BORG_PASSPHRASE value
```
ThomasWaldmann: Maybe rather explain that one needs "export" to have shell variable being exported into the environment of subprocesses - so the borg process can use them. Shell variables are only visible within the script, but not to subprocesses.

docs/quickstart.rst
jzacsh: > complicated & long

Done

> the ; at the end is not needed, is it?

Probably not, but just trying to be complete, for those who aren't proficient with shell (that is: safe for someone who sees the "tl;dr" and copy/pastes)

docs/quickstart.rst
jzacsh: Done.

docs/quickstart.rst
jzacsh: Done.

docs/quickstart.rst
```diff
@@ -129,6 +129,38 @@ certain number of old archives::
...
+
+.. note:: ``sudo`` does not inherit your environment, so ``sudo borg`` will not
+          have access to `BORG_PASSPHRASE` despite running the above example.
+
+To debug what your borg backup is truly seeing, for example if ``BORG_PASSPHRASE``
```
ThomasWaldmann: To debug what your borg process is actually seeing, find its PID (``ps aux|grep borg``) and then look into ``/proc/<PID>/environ``.

(this is not a linux course, but we give some hints)

docs/quickstart.rst
jzacsh: done

docs/quickstart.rst
```diff
@@ -145,18 +145,25 @@ Note that this is not the same thing as::
...
+  $ BORG_PASSPHRASE='complicated & long' # BAD: see explanation above
   $ ./yourborgjob.sh
   ...
   Enter passphrase for key ...:
 
-In another shell::
+Leave that ^ prompt hanging, and in another shell, inspect it's environment::
 
   $ ps aux | grep borg # find the python PID that borg is running as; E.g.: 123
   $ sudo cat /proc/123/environ  # inspect and verify
```
ThomasWaldmann: cut those 10 lines down, see my previous comment. we must not blow up borg documentation too much.

docs/quickstart.rst
```diff
@@ -145,18 +145,25 @@ Note that this is not the same thing as::
   $ BORG_PASSPHRASE='complicated & long'; # missing "export"!
   $ ./yourborgjob.sh # will never see the above $BORG_PASSPHRASE value
 
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page, or the relevant POSIX_ spec.
+
+.. _POSIX: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_22.
+
```
ThomasWaldmann: keep the bash(1) reference, but if someone wants more, he can google the link to the posix spec by himself.

docs/quickstart.rst
jzacsh: Done

docs/quickstart.rst
jzacsh: done.

docs/quickstart.rst
jzacsh: sgtm. i addressed the other comment - is this good now, or cut down further?

docs/quickstart.rst
```diff
@@ -129,6 +129,32 @@ certain number of old archives::
...
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+.. note:: ``sudo`` does not inherit your environment, so ``sudo borg`` will not
+          have access to ``BORG_PASSPHRASE`` despite running the above example.
+
+To debug what your borg process is actually seeing, find its PID
+(``ps aux|grep borg``) and then look into ``/proc/<PID>/environ``.
```
textshell: Do people who need this level of help really can just look into a file with zero terminated records without more hints? environ looks quite messed up for someone not into C programming and remembering that the record separators just get eaten by the terminal. I’m all for giving this hint, and keeping things short, but i’m not sure if this is more confusing like this.

docs/quickstart.rst
```diff
@@ -129,6 +129,32 @@ certain number of old archives::
...
+behavior. E.g.::
+
+    export BORG_PASSPHRASE='complicated & long';
+
+This ensures the borg process that needs to access this passphrase can do so.
+Note that this is not the same thing as::
+
+  $ BORG_PASSPHRASE='complicated & long'; # missing "export"!
+  $ ./yourborgjob.sh # will never see the above $BORG_PASSPHRASE value
+
```
ThomasWaldmann: I'ld say, kill lines 142-147, they are not adding information.

docs/quickstart.rst
jzacsh: it might be more confusing, but i don't think it hurts to just submit, and wait to hear feedback from folks getting confused or not?

Wrt, "level of help" - I knew about `/proc/` yet still didn't think to do this on my own (someone in IRC suggested it). That is: the helpfulness of this hint in the documentation is not necessarily related to being programmer.

docs/quickstart.rst
```diff
@@ -129,6 +129,32 @@ certain number of old archives::
...
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+.. note:: ``sudo`` does not inherit your environment, so ``sudo borg`` will not
+          have access to ``BORG_PASSPHRASE`` despite running the above example.
```
ThomasWaldmann: Maybe a hint how to work around that?

jzacsh: Done.

docs/quickstart.rst
jzacsh: Done.

jzacsh:
@ThomasWaldmann k, i believe i addressed all the comments so far. Feel free to take another look.

docs/quickstart.rst
```diff
@@ -129,6 +129,28 @@ certain number of old archives::
...
+
+    export BORG_PASSPHRASE='complicated & long';
+
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+.. note:: ``sudo`` does not inherit your environment, so ``sudo borg ...`` will not
+          have access to ``BORG_PASSPHRASE`` despite running the above example,
+          but ``sudo export BORG_PASSPHRASE='complicated & long' borg ...``
```
textshell: This doesn‘t seem to work. without the export it seems to work. But sudo is complicated, and calling like this will leave BORG_PASSPHRASE visible to ps in some configurations. sudo is really quite complicated.

Maybe we should rather go

> For options to pass environment variables through sudo consider using the ``env_keep`` option in suders or the ``sudo BORG_PASSPHRASE='complicated & long' borg`` syntax. Values passed as parameters to sudo might be visible to other users.

docs/quickstart.rst
```diff
@@ -129,6 +129,28 @@ certain number of old archives::
...
+and use single quotes if you're unsure of the details of your shell's expansion
+behavior. E.g.::
+
+    export BORG_PASSPHRASE='complicated & long';
+
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+.. note:: ``sudo`` does not inherit your environment, so ``sudo borg ...`` will not
```
jzacsh: after chatting in IRC, it seems I'm touching a gold mine of complexity with my comment on `sudo`, so perhaps the following is better?

> Beware of how `sudo` interacts with environment variables: if testing a wrapper,
> for example, you may be surprised that the following export has no effect:
> ```sh
>    export BORG_PASSPHRASE='...';
>    sudo ./yourborgwrapper.sh # still prompts for password
> ```
> See the sudo(8) man page, and your system's sudo configuration for more.

FabioPedretti:
A couple suggestions for master (with reference to: http://borgbackup.readthedocs.io/en/latest/usage.html):
- [ ] borg list has an example (the last one) for showing diff. This should be removed since there is a dedicated borg list command now;
- [ ] elaborate a bit more on _auto,C[,L]_ heuristic: based on file extension, file type or it tries to compress and use none if the compressed is just X% smaller?
Thanks

ThomasWaldmann:
@FabioPedretti that is out-of-scope of this PR, can you open a new PR or ticket?

FabioPedretti:
Right, reported in #1390. Thanks.

docs/quickstart.rst
jzacsh: Okay, I've _hopefully_ combined both our explanations in a concise format

docs/quickstart.rst
jzacsh: done (same as reply to texshell)

docs/quickstart.rst
```diff
@@ -143,10 +143,14 @@ This is because ``export`` exposes variables to subprocesses, which borg may be
 one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
 bash(1) man page.
 
-.. note:: ``sudo`` does not inherit your environment, so ``sudo borg ...`` will not
-          have access to ``BORG_PASSPHRASE`` despite running the above example,
-          but ``sudo export BORG_PASSPHRASE='complicated & long' borg ...``
-          will.
+Beware of how ``sudo`` interacts with environment variables. ``sudo``.  For
```
ThomasWaldmann: the 2nd sudo seems like a edit leftover.

docs/quickstart.rst
```diff
@@ -143,10 +143,14 @@ This is because ``export`` exposes variables to subprocesses, which borg may be
...
+example, you may be surprised that the following export has no effect::
+
+   export BORG_PASSPHRASE='...'
+   sudo ./yourborgwrapper.sh # still prompts for password
```
ThomasWaldmann: use 2 blanks left of the "#"

docs/quickstart.rst
```diff
@@ -143,10 +143,14 @@ This is because ``export`` exposes variables to subprocesses, which borg may be
...
+
+See the sudo(8) man page, for more (Hint: see ``env_keep`` in sudoers(5), or try
+``sudo BORG_PASSPHRASE='yourphras' borg`` syntax).
```
ThomasWaldmann: ... for more information

'yourphrase'

jzacsh:
@ThomasWaldmann in case I want to revert anything, I'll wait until you give a "LGTM" before I squash all my commits down for a final merge.

docs/quickstart.rst
```diff
@@ -129,6 +129,33 @@ certain number of old archives::
...
 
+Pitfalls with shell variables and environment variables
+-------------------------------------------------------
+
+This applies to all environment variables you want borg to see, not just
+``BORG_PASSPHRASE``. The short explanation is: always ``export`` your variable,
+and use single quotes if you're unsure of the details of your shell's expansion
+behavior. E.g.::
+
+    export BORG_PASSPHRASE='complicated & long';
```
ThomasWaldmann: well, I already noted it a while ago: is the semicolon needed?
I never used one there, so maybe rather not?

docs/quickstart.rst
```diff
@@ -129,6 +129,33 @@ certain number of old archives::
...
+
+This is because ``export`` exposes variables to subprocesses, which borg may be
+one of. More on ``export`` can be found in the "ENVIRONMENT" section of the
+bash(1) man page.
+
+Beware of how ``sudo`` interacts with environment variables. E.g.: you may be
```
ThomasWaldmann: Maybe remove the "E.g.: " here completely.

docs/quickstart.rst
PlasmaPower: I've never used a semicolon, I don't think it's necessary.

ThomasWaldmann:
I left some comments about remaining minor things.
Fix them and then collapse all changesets into 1.

docs/quickstart.rst
jzacsh: Done

docs/quickstart.rst
jzacsh: Done.

docs/quickstart.rst
jzacsh: Done.

docs/quickstart.rst
jzacsh: Done.

docs/quickstart.rst
jzacsh: Done.

jzacsh:
K, squashed into one commit, and addressed each comment.

ThomasWaldmann:
thanks!

adatum:
So finally what is the recommended way of dealing with `sudo`? The documentation does not present a solution.

For example, if backing up `/`, or even just `/etc`, this is a very relevant issue.

ThomasWaldmann:
@adatum in general, the borg docs scope is borg, not trying to document how the world works.

So, if you want to know more about sudo, read the sudo docs.

adatum:
I would suggest it is well within the scope of borg to document best practices on how to properly back up system files or other users' files without creating issues with permissions or security. 

Does a solution for such a common use case need to be re-discovered by each borg user?

ThomasWaldmann:
you can say the same thing about cron or systemd, we also do not document them.

so just read the docs of the tool you want to use.

PR #1383, opened 2016-07-27
Skip mode tests with BORG_TESTS_IGNORE_MODES

### LINKED ISSUE #1363: Tests failing on ntfs-3g

PlasmaPower:
Along with #1373, resolves #1363

If you cherry pick this and #1373 onto #1359, all tests pass! :tada:

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -988,6 +988,7 @@ class ArchiverTestCase(ArchiverTestCaseBase):
             # Restore permissions so shutil.rmtree is able to delete it
             os.system('chmod -R u+w ' + self.repository_path)
 
+    @pytest.mark.skipif('BORG_TESTS_IGNORE_MODES' in os.environ, reason='modes unreliable')
     def test_umask(self):
         self.create_regular_file('file1', size=1024 * 80)
         self.cmd('init', self.repository_location)
@@ -1936,7 +1937,8 @@ class DiffArchiverTestCase(ArchiverTestCaseBase):
             assert 'input/file_unchanged' not in output
 
             # Directory replaced with a regular file
-            assert '[drwxr-xr-x -> -rwxr-xr-x] input/dir_replaced_with_file' in output
+            if 'BORG_TESTS_IGNORE_MODES' not in os.environ:
+                assert '[drwxr-xr-x -> -rwxr-xr-x] input/dir_replaced_with_file' in output
 
             # Basic directory cases
             assert 'added directory     input/dir_added' in output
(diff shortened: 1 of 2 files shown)


enkore:
lgtm

src/borg/testsuite/archiver.py
```diff
@@ -988,6 +988,7 @@ def test_readonly_repository(self):
             # Restore permissions so shutil.rmtree is able to delete it
             os.system('chmod -R u+w ' + self.repository_path)
 
+    @pytest.mark.skipif('BORG_TESTS_UNRELIABLE_MODES' in os.environ, reason='modes unreliable')
```
ThomasWaldmann: that env var name is a bit confusing.
isn't it rather BORG_TESTS_SKIP_UNRELIABLE_MODES?
"borg tests unreliable modes" sounds rather like the opposite.

ThomasWaldmann: also, at least a short comment in the source about why we have this would be useful for the future reader.

ThomasWaldmann:
one minor thing, other than that it lgtm.

src/borg/testsuite/archiver.py
PlasmaPower: I'm trying to indicate that the modes themselves are unreliable, not the tests on them. Maybe `BORG_TESTS_IGNORE_MODES`?

ThomasWaldmann: yes, that is less confusing.

PlasmaPower: Fixed.

PlasmaPower:
I'm going to merge this into #1373 so I can include this in the py.test header. This patch probably belongs there anyway.

PR #1386, opened 2016-07-27
Add --append-only to borg init (1.0-maint)

### LINKED ISSUE #1291: add --append-only to 'borg init'

PlasmaPower:
Replaces #1357 

Fixes #1291


enkore:
lgtm

PR #1403, opened 2016-07-28
FUSE: always create a root dir, fixes #1125

### LINKED ISSUE #1125: borg mount crashes with empty archive
if one tries to mount a completely empty archive, borg mount crashes.

note: usually this never happens as one does not have empty archives.

ThomasWaldmann:
also: refactor / deduplicate directory creation


enkore:
lgtm

ThomasWaldmann:
@enkore I didn't see why the code wasn't that easy before. Do you see the point of the old code?

enkore:
No, I'm not sure why the two paths were not deduplicated. Maybe it was just not noticed?

Issue #1406, opened 2016-07-29
add include patterns


leo-b:
Sometimes the ability to not only exclude patterns but also explicitly specify includes can be useful.

I propose a new optional pattern prefix '+' that may be used with --exclude or in --exclude-from files to define include-patterns. (Just like include/exclude pattern rules in rsync.)

These patterns may be used to explicitly include files in a backup that would otherwise be excluded by a following pattern.

Some use cases:
- exclude docs but backup pdfs:
  
  ```
  --exclude +/data/docs/pdf --exclude /data/docs/
  ```
  
  (maybe `--include /data/docs/pdf` should be introduced to improve readability.)
- use a common exclude list deployed to all backup clients but override it on some specific client:
  
  `global.inclexcl`:
  
  ```
  /tmp
  *.tmp
  /scratch
  ```
  
  `local.inclexcl`:
  
  ```
  +/scratch
  /home/dir_to_exclude
  ```
  
  `$ borg create --exclude-from local.inclexcl --exclude-from global.inclexl backup /`
  
  equivalent:
  
  `global.inclexcl`:
  
  ```
  -sh:/tmp
  -fm:*.tmp
  -fm:/scratch
  ```
  
  `local.inclexcl`:
  
  ```
  +fm:/scratch
  -/home/dir_to_exclude
  ```

(Actually the same approach is already used by the --keep-tag-files option.)

PlasmaPower:
Filenames can start with + though.

leo-b:
> Filenames can start with + though.

Use ``-+file`` or ``pm:+file`` to escape and exclude a file named ``+file``.
It's just like you have to escape files named ``xx:file``:
> Explicit style selection is necessary when a non-default style is desired or when the desired pattern starts with two alphanumeric characters followed by a colon (i.e. aa:something/*).

PlasmaPower:
@leo-b It still would be a breaking change, although probably nobody's excluding a file starting with `+`.

leo-b:
@PlasmaPower I agree it should be noted in the release notes for academic purposes... ;-)

ThomasWaldmann:
@leo-b don't you find giving include patterns with --exclude **weird** somehow?

enkore:
That weirdness is shared with some other programs as well (`.*ignore` files are notorious here).

I'm not sure if this is really needed. It feels kinda rare use case and since both exclude and paths support regexpes the functionality is already there (although more complex to use).

leo-b:
[...]

PR #1407, opened 2016-07-29
add include patterns to --exclude lists

leo-b:
This PR adds include patterns as proposed in #1406.


PlasmaPower:
If we wanted to keep backwards compatibility we could add this as a new option, `--filter-from`. Not sure how important that is though.

ThomasWaldmann:
btw, flake8 checker is unhappy.

leo-b:
Sorry, forgot the last whitespace fixing commit. :-)

enkore:
Superseded by #1971

Issue #1342, opened 2016-07-16
windows: backup and extract ADS


enkore:
Alternate Data Streams (ADS) are the Windows equivalent of resource forks (and/or extended attributes, depending on platform and the moon-phase when you're asking the question ;)

Support for these shall be added to the borg.xattr module.

ThomasWaldmann:
isn't there the big difference that a ADS can contain lots of data? size like the regular file contents?

enkore:
How large xattrs can be depends completely on the system and FS. Linux has a limit of 64K here, I think some file systems on Linux have further restrictions. On the other Hand, OSX resource forks (which are read through the xattr API as well #51) have no length limit, while Solaris has no limits at all (#1337).

Typical examples of ADS aren't very large.

Anakonda:
In theory ADS can store any data that file can - if you open a file it opens ADS without name. eg `filename` is `filename::$DATA`.
I would try storing them as seperate files. Note that when trying to extract `file:stream` created in non-windows `file` would have to be created too/first.

enkore:
Hm, maybe just introduce an Item attribute like "subitems" with a list of items that are encapsulated by the parent item. Then an ADS would just be a regular item in that list, with chunked contents (so no problems with large ADS). This approach would also allow storing what Solaris does.

ThomasWaldmann:
Doesn't that mean that the LIST ("directory") of streams can't be more than 64K, but the stream itself is unbounded?

enkore:
Hm, didn't read it like this, but seems more likely.

> SMB 3.0 supports list **(**of streams**)** less than or equal to 64K.

enkore:
More research

- When using ntfs(-3g) xattr's are mapped to ADS, specifically the `user.` namespace is (everything else is empty)
  - ADS larger than 64K are not accessible on Linux (*Argument list too long*)
  - I.e. the ADS "asdf" becomes the xattr "user.asdf" on Linux
  - An ADS larger than 64K can be listed, but not retrieved.
- NTFS EA are not accessible on Linux (and are only used by Windows 8 and up for some system files apparently)

tk7r:
@enkore : MacOSX xattr is NOT limited to 4K. Quite contrary.
I have tested it on MacOSX 10.7 and could set more than 100 KByte as an attribute. (limited by the command line size limit)
In my tests on MacOSX 10.11 i have found a file "in the wild" with more  than 300KByte as an attribute.
[...]

Issue #983, opened 2016-04-25
Advanced Windows features


enkore:
- [ ] DACLs + SACLs of files / folders
- [ ] NTFS ADS
- [ ] Some attributes, preferably those that can be translated from/to bsdflags (there is a branch doing that for Linux already)
- [ ] --numeric-id should for SIDs vs. usernames like UID vs username on unixy
- [ ] Single redistributable Pyinstaller binary
- [ ] Be able to work with VSS (a user interface as simple as `borg create repo::archive C: --use-vss` would be ideal)
- [ ] Check/ Add support for long paths (NT limit is ~32 kB)

ThomasWaldmann:
About priorities: maybe first concentrate on basic functionality that a non-admin user needs to install borg and backup "normal" files and directories (owned by that user, not system files, not locked/open files), later move on to more special stuff.

Anakonda:
- Doing DACLs atm. I cannot fetch SACL, permission denied even when running as admin - do we even need SACLs.
- Windows own "turn DACL to string" function uses only SID do we really need to use usernames.

- Python does not support XP. In theory Python does support Vista, I don't see point in supporting Vista.

enkore:
- For fetching stuff like the SACL there is an extra security privilege, I'd have to look up which one and whether admin should have it by default (-- UAC...); I tend to agree that the SACL is secondary and the DACL is the primary interest. I split the item.
- I think storing both usernames and SIDs makes sense for the same reasons we store both ID and name on *nix.
- 7 or later it is.

ThomasWaldmann:
http://alphavss.alphaleonis.com/

Anakonda:
Interesting ...

enkore:
COM is actually language independent... although you need *a lot* of glue code in vanilla C/C++ (~Cython). Good news, Python has comtypes, which makes COM very simple:

```py
devmgr = comtypes.CreateObject("WIA.DeviceManager")
for di in devmgr.DeviceInfos:
   pass
```

.... easier than d-bus...

enkore:
What's a bit hairy with DACLs (and SACLs) is that many of them are inherited. So if I e.g. back up something and then extract just a subtree of it then naturally the extracted subtree would only have the ACL entries set in that subtree, making the effective ACL of the extracted subtree and the original subtree different.

[...]

PR #1411, opened 2016-07-30
Windows: alternate data steams

Anakonda:
Backup and extract Windows alternate data steams (#1342 #983 )
ADS are saved as separate items and extracted as such on non-windows.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -51,6 +51,7 @@ from .upgrader import AtticRepositoryUpgrader, BorgRepositoryUpgrader
 
 if sys.platform == 'win32':
     import posixpath
+    from .platform import get_ads
 
 
 def argument(args, str_or_bool):
@@ -409,6 +410,13 @@ class Archiver:
             else:
                 status = '-'  # dry run, item was not backed up
         self.print_file_status(status, path)
+        if sys.platform == 'win32':
+            if ':' not in os.path.basename(path):
+                for stream in get_ads(path):
+                    if stream != '::$DATA':
+                        self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
+                            keep_tag_files, skip_inodes, path + stream[:-6], restrict_dev,
+                            read_special, dry_run, st)
 
     @with_repository()
     @with_archive
--- a/src/borg/platform/windows.pyx
+++ b/src/borg/platform/windows.pyx
@@ -77,6 +77,7 @@ cdef extern from 'windows.h':
     cdef extern int ERROR_INSUFFICIENT_BUFFER
     cdef extern int ERROR_INVALID_SID
     cdef extern int ERROR_NONE_MAPPED
+    cdef extern int ERROR_HANDLE_EOF
 
     cdef extern int OWNER_SECURITY_INFORMATION
     cdef extern int GROUP_SECURITY_INFORMATION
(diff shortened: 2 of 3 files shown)


src/borg/archiver.py
```diff
@@ -409,6 +410,13 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
             else:
                 status = '-'  # dry run, item was not backed up
         self.print_file_status(status, path)
+        if sys.platform == 'win32':
+            if ':' not in os.path.basename(path):
+                for stream in get_ads(path):
+                    if stream != '::$DATA':
+                        self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
+                            keep_tag_files, skip_inodes, path + stream[:-6], restrict_dev,
```
ThomasWaldmann: what's in the last 6 chars?

src/borg/platform/windows.pyx
```diff
@@ -345,3 +359,21 @@ def sync_dir(path):
...
+    cdef HANDLE searchHandle = FindFirstStreamW(cstrPath, FindStreamInfoStandard, <void*>&data, 0)
+    if searchHandle == <HANDLE>INVALID_HANDLE_VALUE:
+        PyMem_Free(cstrPath)
+        return []
+    ret.append(PyUnicode_FromWideChar(data.cStreamName, -1))
+    while FindNextStreamW(searchHandle, <void*>&data) != 0:
+        ret.append(PyUnicode_FromWideChar(data.cStreamName, -1))
+    errno = GetLastError()
+    if errno != 38:
+        print(errno)
```
ThomasWaldmann: print?

src/borg/archiver.py
Anakonda: `:$DATA`

src/borg/platform/windows.pyx
```diff
@@ -345,3 +359,21 @@ def sync_dir(path):
...
+    cdef wchar_t* cstrPath = PyUnicode_AsWideCharString(path, NULL)
+    cdef HANDLE searchHandle = FindFirstStreamW(cstrPath, FindStreamInfoStandard, <void*>&data, 0)
+    if searchHandle == <HANDLE>INVALID_HANDLE_VALUE:
+        PyMem_Free(cstrPath)
+        return []
+    ret.append(PyUnicode_FromWideChar(data.cStreamName, -1))
+    while FindNextStreamW(searchHandle, <void*>&data) != 0:
+        ret.append(PyUnicode_FromWideChar(data.cStreamName, -1))
+    errno = GetLastError()
+    if errno != 38:
```
enkore: ERROR_HANDLE_EOF

ThomasWaldmann:
If I would extract a single filename on windows, it would not extract its ADS, right?
So I somehow would need to lookup first whether the file has ADS and then manually extract them all.

Anakonda:
Correct, I don't think that is problem

enkore:
lgtm

ThomasWaldmann:
@Anakonda well, in the case i have given (someone wants to extract a single file), the resulting file does not have the ADS because they are treated like separate files. If one is aware of that, one can workaround this, but if not ...

enkore:
^- I think this is ok (for now?)

Merge?

Anakonda:
I think merge.

enkore:
Thanks, and sorry this took so long.

PR #1413, opened 2016-07-30
save mountpoint dirs, fixes #1033

### LINKED ISSUE #1033: Store empty mountpoint-directories in archives when staying on single fs
When using borg -x currently the archive would not contain the mointpoints like /proc, /sys and whatever else.
This becomes an inconvenience later on when you want to do a full-system-restore since you need to remember to recreate all of them too.

I think it would be a lot better to store mountpoints as empty directories in the archive instead.

Similarly --exclude might need something like that if the target happens to be a directory, but I am less certain about that and there apparently might be workarounds for that too like specifying slashes at the end of the excluded names.

ThomasWaldmann:


ThomasWaldmann:
hmm, we could slightly simplify processing if we process tagged directories like mountpoints: save the directory (and maybe the tag files), but do not process the dir contents and recurse.

enkore:
Should be. Change lgtm from here

ThomasWaldmann:
Tried to simplify it, but that needs test changes and then the recreater tests also break...

So I'ld say we change that in a separate PR.

ThomasWaldmann:
collapsed changesets

Issue #1391, opened 2016-07-28
document files cache entry age behaviour


ThomasWaldmann:
https://github.com/borgbackup/borg/commit/b9952efd68163a909c05cdd3f2a5408e27dd5424 removed information about entry age behaviour.

Create a new FAQ entry and document it (in a better way).

ThomasWaldmann:
Fixed by #1414.

PR #1414, opened 2016-07-30
implement BORG_FILES_CACHE_TTL, update FAQ

### LINKED ISSUE #1338: Cache persistency
Doing a daily single "borg create" with all required directories is a lot faster (10-50x) than doing many different "borg create" with every directory, still in the same repo.
It looks like the cache get purged of previous files not seen in the current borg run, but that could be seen in a subsequent one.
I propose to add a system of cache persistency, so that it is possible to do many borg runs while finding the files in cache.
An approach could be to keep file caches until at least some time (maybe 1 day as default) is passed without seeing them. This way you could do many daily different borg create.
An improvement could also consider that backups may not be done some time periods (e.g. week end days, or when there is a network failure) and still not purge the cache when the first run of batch is done.

ThomasWaldmann:
raise default ttl to 20 (previously: 10).

fixes #1338 and #1391.

### DIFF
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -193,12 +193,13 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         if not self.txn_active:
             return
         if self.files is not None:
+            ttl = int(os.environ.get('BORG_FILES_CACHE_TTL', 20))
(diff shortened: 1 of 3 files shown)


enkore:
lgtm

PR #1418, opened 2016-07-31
Cleanup prefetch code in RemoteRepository

textshell:
The preloading code pretends to be more general than it is (and is used)

This clarifies what the code can actually do. It also decouples the cache data structure from rpc protocol details opening the way to use something else than tuples as rpc argument encoding.

scaled back version that only changes indexing of the cache and makes sure preload is only used with 'get' method.

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -425,20 +425,21 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                     if calls:
                         if is_preloaded:
                             assert cmd == "get", "is_preload is only supported for 'get'"
-                            if calls[0] in self.cache:
-                                waiting_for.append(fetch_from_cache(calls.pop(0)))
+                            if calls[0][0] in self.cache:
+                                waiting_for.append(fetch_from_cache(calls.pop(0)[0]))
                         else:
                             args = calls.pop(0)
-                            if cmd == 'get' and args in self.cache:
-                                waiting_for.append(fetch_from_cache(args))
+                            if cmd == 'get' and args[0] in self.cache:
+                                waiting_for.append(fetch_from_cache(args[0]))
                             else:
                                 self.msgid += 1
                                 waiting_for.append(self.msgid)
                                 self.to_send = msgpack.packb((1, self.msgid, cmd, args))
                     if not self.to_send and self.preload_ids:
-                        args = (self.preload_ids.pop(0),)
+                        chunk_id = self.preload_ids.pop(0)
+                        args = (chunk_id,)
                         self.msgid += 1
-                        self.cache.setdefault(args, []).append(self.msgid)
+                        self.cache.setdefault(chunk_id, []).append(self.msgid)
                         self.to_send = msgpack.packb((1, self.msgid, 'get', args))
 
                 if self.to_send:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -216,7 +216,7 @@ class RemoteRepository:
         self.preload_ids = []
         self.msgid = 0
         self.to_send = b''
-        self.cache = {}
+        self.chunkid_to_msgids = {}
         self.ignore_responses = set()
         self.responses = {}
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
@@ -350,10 +350,10 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
         if not calls:
             return
 
-        def fetch_from_cache(args):
-            msgid = self.cache[args].pop(0)
-            if not self.cache[args]:
-                del self.cache[args]
+        def pop_preload_msgid(chunkid):
+            msgid = self.chunkid_to_msgids[chunkid].pop(0)
+            if not self.chunkid_to_msgids[chunkid]:
+                del self.chunkid_to_msgids[chunkid]
             return msgid
 
         def handle_error(error, res):
@@ -425,12 +425,12 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                     if calls:
                         if is_preloaded:
                             assert cmd == "get", "is_preload is only supported for 'get'"
-                            if calls[0][0] in self.cache:
-                                waiting_for.append(fetch_from_cache(calls.pop(0)[0]))
+                            if calls[0][0] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(calls.pop(0)[0]))
                         else:
                             args = calls.pop(0)
-                            if cmd == 'get' and args[0] in self.cache:
-                                waiting_for.append(fetch_from_cache(args[0]))
+                            if cmd == 'get' and args[0] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(args[0]))
                             else:
                                 self.msgid += 1
                                 waiting_for.append(self.msgid)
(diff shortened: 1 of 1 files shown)


borg/remote.py
```diff
@@ -316,22 +316,18 @@ def handle_error(error, res):
...
-                                self.to_send = msgpack.packb((1, self.msgid, cmd, args))
+                            self.msgid += 1
+                            waiting_for.append(self.msgid)
+                            self.to_send = msgpack.packb((1, self.msgid, cmd, args))
                     if not self.to_send and self.preload_ids:
                         args = (self.preload_ids.pop(0),)
                         self.msgid += 1
                         self.cache.setdefault(args, []).append(self.msgid)
-                        self.to_send = msgpack.packb((1, self.msgid, cmd, args))
+                        self.to_send = msgpack.packb((1, self.msgid, 'get', args))
```
ThomasWaldmann: This if for:
"""
using anything other that a get or get_many call on the RemoteRepository would have tried to
use the per instance preload list with a wrong command.
"""

Looks correct.

borg/remote.py
```diff
@@ -316,22 +316,18 @@ def handle_error(error, res):
             if w:
                 while not self.to_send and (calls or self.preload_ids) and len(waiting_for) < 100:
                     if calls:
-                        if is_preloaded:
-                            if calls[0] in self.cache:
-                                waiting_for.append(fetch_from_cache(calls.pop(0)))
```
ThomasWaldmann: I have doubts that removing this is beneficial:

The difference between the if and else blocks seems to be that if is_preloaded is True, we know that all the content data chunks have been queued for preloading (borg extract does this). Thus we never do request them. If we do not already have them in the cache, it just means that the preload request was not processed / answered YET (but will be later).

In the "else" branch (is_preloaded is False), we actively request anything we have not in the cache yet because we know that no preload was queued for this chunk, so we need to do the request here.

borg/remote.py
```diff
@@ -317,16 +317,16 @@ def handle_error(error, res):
                 while not self.to_send and (calls or self.preload_ids) and len(waiting_for) < 100:
                     if calls:
                         args = calls.pop(0)
```
ThomasWaldmann: chunkid = args[0]
...

borg/remote.py
```diff
@@ -317,16 +317,16 @@ def handle_error(error, res):
...
-                        if cmd == 'get' and args in self.cache:
-                            waiting_for.append(fetch_from_cache(args))
+                        if cmd == 'get' and args[0] in self.get_cache:
+                            waiting_for.append(fetch_from_cache(args[0]))
                         else:
                             self.msgid += 1
                             waiting_for.append(self.msgid)
                             self.to_send = msgpack.packb((1, self.msgid, cmd, args))
                     if not self.to_send and self.preload_ids:
                         args = (self.preload_ids.pop(0),)
```
ThomasWaldmann: chunkid = self.preload_ids.pop(0)
args = (chunkid, )

borg/remote.py
textshell: I did not quite see that difference. But i think when the result is different the old code was broken. The call sequence is:
* repo.preload(chunk_ids_p1)
* repo.preload(chunk_ids_p2)
* repo.preload(chunk_ids_p3)
* repo.get_many(chunk_ids_f1, is_preload=True)
* repo.get_many(chunk_ids_f2, is_preload=True)

at the moment.
In the old code if chunk_ids_f1 contains chunk_ids that are also in chunk_ids_{p1,p2,p3} then the preloads are not yet in the cache (because that is only filled after requesting the parameters) and nothing is appended to waiting_for. I think that would result in skiping the chunk which get_many may never do because code like this:

        for id_, data in zip(ids, self.repository.get_many(ids, is_preloaded=is_preloaded)):
            yield self.key.decrypt(id_, data)

In my modification the chunks are correctly loaded, but than later are cached and then likely leaked because they are not used from the cache.

borg/testsuite/archiver.py
```diff
@@ -1298,6 +1298,29 @@ def test_fuse(self):
     def test_debug_put_get_delete_obj(self):
         pass
 
+    def test_strip_components_doesnt_leak(self):
```
enkore: Separate this change (--strip-components leaking, test+fix) from the preload changes?

ThomasWaldmann:
btw: flake8 check is unhappy.

ThomasWaldmann:
The leak fix has been cherrypicked into 1.0-maint and will get merged from there into master.

textshell:
thanks for cherrypicking the leak fix.

rebased to current 1.0-maint.

Maybe i should retarget this to master? I still think these cleanups are useful for anybody who needs to understand this code later or change the rpc encoding. But likely nothing currently hits the edge cases in the existing code.

enkore:
Considering that this is hard to test and understand code and that it's also kinda fragile retargeting master may be a better approach here, yes.

borg/remote.py
```diff
@@ -166,7 +166,7 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
         self.preload_ids = []
         self.msgid = 0
         self.to_send = b''
-        self.cache = {}
+        self.get_cache = {}
```
ThomasWaldmann: how about, instead of renaming the attribute, just put a comment to the right of it, explaining that this cache is only used for get?

textshell: I prefer explicit names. But thinking about this, it isn’t really a cache. Maybe it should be preload_msgids. After all it’s a mapping from chunkid to a list of message ids for the get requests. (list because if a chunk is preloaded multiple times it just requests it again each time, because RemoteRepository.responses has no way of "ref counting")

ThomasWaldmann: chunkid_to_msgids?

textshell: ok, it’s chunkid_to_msgids and pop_preload_msgid now.

Also moved the whole thing to be based on master

src/borg/remote.py
```diff
@@ -314,10 +314,10 @@ def call_many(self, cmd, calls, wait=True, is_preloaded=False):
...
             return
 
-        def fetch_from_cache(args):
-            msgid = self.cache[args].pop(0)
-            if not self.cache[args]:
-                del self.cache[args]
+        def pop_preload_msgid(args):
+            msgid = self.chunkid_to_msgids[args].pop(0)
+            if not self.chunkid_to_msgids[args]:
+                del self.chunkid_to_msgids[args]
```
ThomasWaldmann: s/args/chunkid/ ?

src/borg/remote.py
```diff
@@ -388,21 +388,23 @@ def handle_error(error, res):
...
                                 self.to_send = msgpack.packb((1, self.msgid, cmd, args))
                     if not self.to_send and self.preload_ids:
-                        args = (self.preload_ids.pop(0),)
+                        chunk_id = self.preload_ids.pop(0)
+                        args = (chunk_id,)
                         self.msgid += 1
-                        self.cache.setdefault(args, []).append(self.msgid)
-                        self.to_send = msgpack.packb((1, self.msgid, cmd, args))
+                        self.chunkid_to_msgids.setdefault(chunk_id, []).append(self.msgid)
+                        self.to_send = msgpack.packb((1, self.msgid, 'get', args))
```
ThomasWaldmann: as the args tuple is only used here, i'ld put "..., (chunk_id, ))" here and kill 404. (hah!)

ThomasWaldmann:
some minor things found. other than that lgtm. fix, collapse, merge.

src/borg/remote.py
textshell: Fixed

src/borg/remote.py
textshell: I’d rather not, because args is a useful name here. Also this will change in case we change the rpc encoding.

textshell:
rebased to master now. I did leave 3 commits because i think having small steps is a good thing in code like this. Let me know if you prefer to squash it more.

ThomasWaldmann:
@enkore the thumbs up, did it mean "approve"?

PR #1419, opened 2016-07-31
RFC: Make rpc protocol more extensible

textshell:
The current RPC model does not allow to add new parameters to methods without compatibility hacks.

This introduces a new RPC protocol based on named parameters. The clients still starts with the RPC protocol used since 0.29.0 and only upgrades to the new protocol when the server supports it. When in the future this compatibility is no longer needed this use of the old protocol can be removed while still beeing compatible with all versions containing this PR (this can be done by settings RemoteRepository.dictFormat = True in the constructor)

The amount of compatibility code for the old protocol is fairly small. Consisting of a dict which for each method describes the order of the positional arguments in the old protocol, 2 methods 4 lines long each for argument format conversation and 7 if statements in the protocol code.

For the new Protocol:

> Protocol compatibility:
> In general the server is responsible for rejecting too old clients and the client it responsible for rejecting too old servers. This ensures that the knowledge what is compatible is always held by the newer component.
> 
> The server can do checks for the client version in RepositoryServer.negotiate. If the client_data is 2 then then client is in the version range [0.29.0, 1.0.6] inclusive. For newer clients client_data is a dict which contains client_version.
> 
> For the client the return of the negotiate method is either 2 if the server is in the version range [0.29.0, 1.0.6] inclusive, or it is a dict which includes the server version.
> 
> All method calls on the remote repository object must be whitelisted in RepositoryServer.rpc_methods and have api stubs in RemoteRepository. The @api decorator on these stubs is used to set server version requirements.
> 
> Method parameters are identified only by name and never by position. Unknown parameters are ignored by the server side.
> If a new parameter is important and may not be ignored, on the client a parameter specific version requirement needs to be added or the method version requirement needs to be set to the version where this parameter was introduced.
> When parameters are removed, they need to be preserved  as defaulted parameters on the client stubs so that older servers still get the from them still needed parameters.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2753,10 +2753,14 @@ def main():  # pragma: no cover
             tb = "%s\n%s" % (traceback.format_exc(), sysinfo())
             exit_code = e.exit_code
         except RemoteRepository.RPCError as e:
-            msg = "%s %s" % (e.remote_type, e.name)
-            important = e.remote_type not in ('LockTimeout', )
+            important = e.exception_class not in ('LockTimeout', )
             tb_log_level = logging.ERROR if important else logging.DEBUG
-            tb = sysinfo()
+            if important:
+                msg = e.exception_full
+            else:
+                msg = e.get_message()
+            tb = '\n'.join('Borg server: ' + l for l in e.sysinfo.splitlines())
+            tb += "\n" + sysinfo()
             exit_code = EXIT_ERROR
         except Exception:
             msg = 'Local Exception'
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -122,7 +122,8 @@ class RepositoryServer:  # pragma: no cover
         'load_key',
         'break_lock',
         'get_free_nonce',
-        'commit_nonce_reservation'
+        'commit_nonce_reservation',
+        'inject_exception',
     )
 
     def __init__(self, restrict_to_paths, append_only):
@@ -285,6 +286,27 @@ class RepositoryServer:  # pragma: no cover
                                      exclusive=exclusive)
         self.repository.__enter__()  # clean exit handled by serve() method
         return self.repository.id
+
+    def inject_exception(self, kind):
+        kind = kind.decode()
+        s1 = 'test string'
+        s2 = 'test string2'
+        if kind == 'DoesNotExist':
+            raise Repository.DoesNotExist(s1)
+        elif kind == 'AlreadyExists':
+            raise Repository.AlreadyExists(s1)
+        elif kind == 'CheckNeeded':
+            raise Repository.CheckNeeded(s1)
+        elif kind == 'IntegrityError':
+            raise IntegrityError(s1)
+        elif kind == 'PathNotAllowed':
+            raise PathNotAllowed()
+        elif kind == 'ObjectNotFound':
+            raise Repository.ObjectNotFound(s1, s2)
+        elif kind == 'InvalidRPCMethod':
+            raise InvalidRPCMethod(s1)
+        elif kind == 'divide':
+            0 // 0
 
 
 class SleepingBandwidthLimiter:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -175,7 +175,8 @@ class RepositoryServer:  # pragma: no cover
                         args = decode_keys(unpacked[ARGS])
                     elif isinstance(unpacked, tuple) and len(unpacked) == 4:
                         dictFormat = False
-                        type, msgid, method, args = unpacked
+                        # The first field 'type' was always 1 and has always been ignored
+                        _, msgid, method, args = unpacked
                         method = method.decode()
                         args = self.positional_to_named(method, args)
                     else:
@@ -664,7 +665,8 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                         if isinstance(unpacked, dict):
                             msgid = unpacked[MSGID]
                         elif isinstance(unpacked, tuple) and len(unpacked) == 4:
-                            type, msgid, error, res = unpacked
+                            # The first field 'type' was always 1 and has always been ignored
+                            _, msgid, error, res = unpacked
                             if error:
                                 # ignore res, because it is only a fixed string anyway.
                                 unpacked = {MSGID: msgid, b'exception_class': error}
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -112,7 +112,7 @@ class RepositoryServer:  # pragma: no cover
                             self.repository.close()
                         raise UnexpectedRPCDataFormatFromClient(__version__)
                     type, msgid, method, args = unpacked
-                    method = method.decode('ascii')
+                    method = method.decode()
                     try:
                         if method not in self.rpc_methods:
                             raise InvalidRPCMethod(method)
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -371,13 +371,13 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
             elif error == 'CheckNeeded':
                 raise Repository.CheckNeeded(self.location.orig)
             elif error == 'IntegrityError':
-                raise IntegrityError(res)
+                raise IntegrityError('(not available)')
             elif error == 'PathNotAllowed':
-                raise PathNotAllowed(*res)
+                raise PathNotAllowed()
             elif error == 'ObjectNotFound':
-                raise Repository.ObjectNotFound(res[0], self.location.orig)
+                raise Repository.ObjectNotFound('(not available)', self.location.orig)
             elif error == 'InvalidRPCMethod':
-                raise InvalidRPCMethod(*res)
+                raise InvalidRPCMethod('(not available)')
             else:
                 raise self.RPCError(res.decode('utf-8'), error)
 
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -88,7 +88,7 @@ compatMap = {
     '__len__': (),
     'list': ('limit', 'marker', ),
     'put': ('id', 'data', ),
-    'get': ('id_', ),
+    'get': ('id', ),
     'delete': ('id', ),
     'save_key': ('keydata', ),
     'load_key': (),
@@ -690,12 +690,12 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                     if calls:
                         if is_preloaded:
                             assert cmd == 'get', "is_preload is only supported for 'get'"
-                            if calls[0]['id_'] in self.chunkid_to_msgids:
-                                waiting_for.append(pop_preload_msgid(calls.pop(0)['id_']))
+                            if calls[0]['id'] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(calls.pop(0)['id']))
                         else:
                             args = calls.pop(0)
-                            if cmd == 'get' and args['id_'] in self.chunkid_to_msgids:
-                                waiting_for.append(pop_preload_msgid(args['id_']))
+                            if cmd == 'get' and args['id'] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(args['id']))
                             else:
                                 self.msgid += 1
                                 waiting_for.append(self.msgid)
@@ -705,7 +705,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                                     self.to_send = msgpack.packb((1, self.msgid, cmd, self.named_to_positional(cmd, args)))
                     if not self.to_send and self.preload_ids:
                         chunk_id = self.preload_ids.pop(0)
-                        args = {'id_': chunk_id}
+                        args = {'id': chunk_id}
                         self.msgid += 1
                         self.chunkid_to_msgids.setdefault(chunk_id, []).append(self.msgid)
                         if self.dictFormat:
@@ -756,12 +756,12 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
     def scan(self, limit=None, marker=None):
         """actual remoting is done via self.call in the @api decorator"""
 
-    def get(self, id_):
-        for resp in self.get_many([id_]):
+    def get(self, id):
+        for resp in self.get_many([id]):
             return resp
 
     def get_many(self, ids, is_preloaded=False):
-        for resp in self.call_many('get', [{'id_': id_} for id_ in ids], is_preloaded=is_preloaded):
+        for resp in self.call_many('get', [{'id': id} for id in ids], is_preloaded=is_preloaded):
             yield resp
 
     @api(since=parse_version('1.0.0'))
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -102,7 +102,7 @@ class RepositoryServer:  # pragma: no cover
                     if self.repository is not None:
                         self.repository.close()
                     else:
-                        os.write(stderr_fd, "Borg {}: Got connection close before repository was opened.\n"
+                        os.write(stderr_fd, 'Borg {}: Got connection close before repository was opened.\n'
                                  .format(__version__).encode())
                     return
                 unpacker.feed(data)
@@ -137,7 +137,7 @@ class RepositoryServer:  # pragma: no cover
                             tb = '%s\n%s' % (traceback.format_exc(), sysinfo())
                             logging.error(msg)
                             logging.log(tb_log_level, tb)
-                        exc = "Remote Exception (see remote log for the traceback)"
+                        exc = 'Remote Exception (see remote log for the traceback)'
                         os.write(stdout_fd, msgpack.packb((1, msgid, e.__class__.__name__, exc)))
                     else:
                         os.write(stdout_fd, msgpack.packb((1, msgid, None, res)))
@@ -272,7 +272,7 @@ This TypeError is a cosmetic side effect of the compatibility code borg
 clients >= 1.0.7 have to support older borg servers.
 This problem will go away as soon as the server has been upgraded to 1.0.7+.
 """
-                # emit this msg in the same way as the "Remote: ..." lines that show the remote TypeError
+                # emit this msg in the same way as the 'Remote: ...' lines that show the remote TypeError
                 sys.stderr.write(msg)
                 if append_only:
                     raise self.NoAppendOnlyOnServer()
@@ -283,10 +283,10 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
 
     def __del__(self):
         if len(self.responses):
-            logging.debug("still %d cached responses left in RemoteRepository" % (len(self.responses),))
+            logging.debug('still %d cached responses left in RemoteRepository' % (len(self.responses),))
         if self.p:
             self.close()
-            assert False, "cleanup happened in Repository.__del__"
+            assert False, 'cleanup happened in Repository.__del__'
 
     def __repr__(self):
         return '<%s %s>' % (self.__class__.__name__, self.location.canonical_path())
@@ -310,7 +310,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
 
     def borg_cmd(self, args, testing):
         """return a borg serve command line"""
-        # give some args/options to "borg serve" process as they were given to us
+        # give some args/options to 'borg serve' process as they were given to us
         opts = []
         if args is not None:
             opts.append('--umask=%03o' % args.umask)
@@ -430,7 +430,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                 while not self.to_send and (calls or self.preload_ids) and len(waiting_for) < MAX_INFLIGHT:
                     if calls:
                         if is_preloaded:
-                            assert cmd == "get", "is_preload is only supported for 'get'"
+                            assert cmd == 'get', "is_preload is only supported for 'get'"
                             if calls[0][0] in self.chunkid_to_msgids:
                                 waiting_for.append(pop_preload_msgid(calls.pop(0)[0]))
                         else:
@@ -531,7 +531,7 @@ def handle_remote_line(line):
             logname, msg = msg.split(' ', 1)
             logging.getLogger(logname).log(level, msg.rstrip())
     else:
-        sys.stderr.write("Remote: " + line)
+        sys.stderr.write('Remote: ' + line)
 
 
 class RepositoryNoCache:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -345,9 +345,6 @@ class RemoteRepository:
             self.name = name
             self.remote_type = remote_type
 
-    class NoAppendOnlyOnServer(Error):
-        """Server does not support --append-only."""
-
     class RPCServerOutdated(Error):
         """Borg server is too old for {}. Required version {}"""
 
@@ -412,13 +409,20 @@ class RemoteRepository:
             else:
                 raise Exception('Server insisted on using unsupported protocol version %s' % version)
 
-            try:
-                self.id = self.call('open', {'path': self.location.path, 'create': create, 'lock_wait': lock_wait,
-                                              'lock': lock, 'exclusive': exclusive, 'append_only': append_only})
-            except self.RPCError as err:
-                if self.dictFormat or err.remote_type != 'TypeError':
-                    raise
-                msg = """\
+            def do_open():
+                self.id = self.open(path=self.location.path, create=create, lock_wait=lock_wait,
+                                lock=lock, exclusive=exclusive, append_only=append_only)
+
+            if self.dictFormat:
+                do_open()
+            else:
+                # Ugly detection of versions prior to 1.0.7: If open throws it has to be 1.0.6 or lower
+                try:
+                    do_open()
+                except self.RPCError as err:
+                    if err.remote_type != 'TypeError':
+                        raise
+                    msg = """\
 Please note:
 If you see a TypeError complaining about the number of positional arguments
 given to open(), you can ignore it if it comes from a borg version < 1.0.7.
@@ -426,13 +430,12 @@ This TypeError is a cosmetic side effect of the compatibility code borg
 clients >= 1.0.7 have to support older borg servers.
 This problem will go away as soon as the server has been upgraded to 1.0.7+.
 """
-                # emit this msg in the same way as the 'Remote: ...' lines that show the remote TypeError
-                sys.stderr.write(msg)
-                if append_only:
-                    raise self.NoAppendOnlyOnServer()
-                compatMap['open'] = ('path', 'create', 'lock_wait', 'lock', )
-                self.id = self.call('open', {'path': self.location.path, 'create': create, 'lock_wait': lock_wait,
-                                              'lock': lock, 'exclusive': exclusive, 'append_only': append_only})
+                    # emit this msg in the same way as the 'Remote: ...' lines that show the remote TypeError
+                    sys.stderr.write(msg)
+                    self.server_version = parse_version('1.0.6')
+                    compatMap['open'] = ('path', 'create', 'lock_wait', 'lock', ),
+                    # try again with corrected version and compatMap
+                    do_open()
         except Exception:
             self.close()
             raise
@@ -629,6 +632,11 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                         if e.errno != errno.EAGAIN:
                             raise
         self.ignore_responses |= set(waiting_for)
+
+    @api(since=parse_version('1.0.0'),
+         append_only={'since': parse_version('1.0.7'), 'previously': False})
+    def open(self, path, create=False, lock_wait=None, lock=True, exclusive=False, append_only=False):
+        """actual remoting is done via self.call in the @api decorator"""
 
     @api(since=parse_version('1.0.0'))
     def check(self, repair=False, save_space=False):
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -191,25 +191,53 @@ class RepositoryServer:  # pragma: no cover
                         args = self.filter_args(f, args)
                         res = f(**args)
                     except BaseException as e:
-                        if isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, PathNotAllowed)):
-                            # These exceptions are reconstructed on the client end in RemoteRepository.call_many(),
-                            # and will be handled just like locally raised exceptions. Suppress the remote traceback
-                            # for these, except ErrorWithTraceback, which should always display a traceback.
-                            pass
-                        else:
-                            if isinstance(e, Error):
-                                tb_log_level = logging.ERROR if e.traceback else logging.DEBUG
-                                msg = e.get_message()
-                            else:
-                                tb_log_level = logging.ERROR
-                                msg = '%s Exception in RPC call' % e.__class__.__name__
-                            tb = '%s\n%s' % (traceback.format_exc(), sysinfo())
-                            logging.error(msg)
-                            logging.log(tb_log_level, tb)
-                        exc = 'Remote Exception (see remote log for the traceback)'
                         if dictFormat:
-                            os.write(stdout_fd, msgpack.packb({MSGID: msgid, b'exception_class': e.__class__.__name__}))
+                            ex_short = traceback.format_exception_only(e.__class__, e)
+                            ex_full = traceback.format_exception(*sys.exc_info())
+                            if isinstance(e, Error):
+                                ex_short = e.get_message()
+                            if isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, PathNotAllowed)):
+                                # These exceptions are reconstructed on the client end in RemoteRepository.call_many(),
+                                # and will be handled just like locally raised exceptions. Suppress the remote traceback
+                                # for these, except ErrorWithTraceback, which should always display a traceback.
+                                pass
+                            else:
+                                logging.debug('\n'.join(ex_full))
+
+                            try:
+                                msg = msgpack.packb({MSGID: msgid,
+                                                    b'exception_class': e.__class__.__name__,
+                                                    b'exception_args': e.args,
+                                                    b'exception_full': ex_full,
+                                                    b'exception_short': ex_short,
+                                                    b'sysinfo': sysinfo()})
+                            except TypeError:
+                                msg = msgpack.packb({MSGID: msgid,
+                                                    b'exception_class': e.__class__.__name__,
+                                                    b'exception_args': [x if isinstance(x, (str, bytes, int)) else None
+                                                                        for x in e.args],
+                                                    b'exception_full': ex_full,
+                                                    b'exception_short': ex_short,
+                                                    b'sysinfo': sysinfo()})
+
+                            os.write(stdout_fd, msg)
                         else:
+                            if isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, PathNotAllowed)):
+                                # These exceptions are reconstructed on the client end in RemoteRepository.call_many(),
+                                # and will be handled just like locally raised exceptions. Suppress the remote traceback
+                                # for these, except ErrorWithTraceback, which should always display a traceback.
+                                pass
+                            else:
+                                if isinstance(e, Error):
+                                    tb_log_level = logging.ERROR if e.traceback else logging.DEBUG
+                                    msg = e.get_message()
+                                else:
+                                    tb_log_level = logging.ERROR
+                                    msg = '%s Exception in RPC call' % e.__class__.__name__
+                                tb = '%s\n%s' % (traceback.format_exc(), sysinfo())
+                                logging.error(msg)
+                                logging.log(tb_log_level, tb)
+                            exc = 'Remote Exception (see remote log for the traceback)'
                             os.write(stdout_fd, msgpack.packb((1, msgid, e.__class__.__name__, exc)))
                     else:
                         if dictFormat:
@@ -341,9 +369,34 @@ class RemoteRepository:
     extra_test_args = []
 
     class RPCError(Exception):
-        def __init__(self, name, remote_type):
-            self.name = name
-            self.remote_type = remote_type
+        def __init__(self, unpacked):
+            # for borg < 1.1: unpacked only has b'exception_class' as key
+            # for borg 1.1+: unpacked has keys: b'exception_args', b'exception_full', b'exception_short', b'sysinfo'
+            self.unpacked = unpacked
+
+        def get_message(self):
+            if b'exception_short' in self.unpacked:
+                return b'\n'.join(self.unpacked[b'exception_short']).decode()
+            else:
+                return self.exception_class
+
+        @property
+        def exception_class(self):
+            return self.unpacked[b'exception_class'].decode()
+
+        @property
+        def exception_full(self):
+            if b'exception_full' in self.unpacked:
+                return b'\n'.join(self.unpacked[b'exception_full']).decode()
+            else:
+                return self.get_message() + '\nRemote Exception (see remote log for the traceback)'
+
+        @property
+        def sysinfo(self):
+            if b'sysinfo' in self.unpacked:
+                return self.unpacked[b'sysinfo'].decode()
+            else:
+                return ''
 
     class RPCServerOutdated(Error):
         """Borg server is too old for {}. Required version {}"""
@@ -411,7 +464,7 @@ class RemoteRepository:
 
             def do_open():
                 self.id = self.open(path=self.location.path, create=create, lock_wait=lock_wait,
-                                lock=lock, exclusive=exclusive, append_only=append_only)
+                                    lock=lock, exclusive=exclusive, append_only=append_only)
 
             if self.dictFormat:
                 do_open()
@@ -420,7 +473,7 @@ class RemoteRepository:
                 try:
                     do_open()
                 except self.RPCError as err:
-                    if err.remote_type != 'TypeError':
+                    if err.exception_class != 'TypeError':
                         raise
                     msg = """\
 Please note:
@@ -524,8 +577,11 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                 del self.chunkid_to_msgids[chunkid]
             return msgid
 
-        def handle_error(error, res):
-            error = error.decode('utf-8')
+        def handle_error(unpacked):
+            error = unpacked[b'exception_class'].decode()
+            old_server = b'exception_args' not in unpacked
+            args = unpacked.get(b'exception_args')
+
             if error == 'DoesNotExist':
                 raise Repository.DoesNotExist(self.location.orig)
             elif error == 'AlreadyExists':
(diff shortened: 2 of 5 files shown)


textshell:
rebased to master + current version of #1418

src/borg/remote.py
```diff
@@ -21,6 +22,7 @@
 from .repository import Repository
 
 RPC_PROTOCOL_VERSION = 2
+BORG_VERSION = (1, 1, 0)
```
ThomasWaldmann: hmm, this looks like a problem. our usual version is automatically maintained by setuptools-scm (and way more complex than this). mainting a separate tuple here seems not desirable.

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
     """Got unexpected RPC data format from server."""
 
 
+# Protocol compatibility:
+# In general the server is responsible for rejecting too old clients and the client it responsible for rejecting
+# too old servers. This ensures that the knowleadge what is compatible is always held by the newer component.
```
ThomasWaldmann: typo: knowledge

good idea. :)

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+#
+# The server can do checks for the client version in RepositoryServer.negotiate. If the client_data is 2 then then
```
ThomasWaldmann: typo "then then"

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+# client is in the version range [0.29.0, 1.0.6] inclusiv. For newer clients client_data is a dict which contains
```
ThomasWaldmann: typo: inclusive (or just leave it away)

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+# client_version.
+#
+# For the client the return of the negotiate method is either 2 if the server is in the version range [0.29.0, 1.0.6]
+# inclusiv, or it is a dict which includes the server version.
```
ThomasWaldmann: inclusive

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+#
+# All method calls on the remote repository object must be whitelisted in RepositoryServer.rpc_methods and have api
+# stubs in RemoteRepository. The @api decorator on these stubs is used to set server version requirements.
+#
+# Method parameters are identfied only by name and never by position. If not overriden by a version requirement,
```
ThomasWaldmann: identified ... overridden

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+# unknown parameters are ignored by the server side. When parameters are removed, they need to be preserved
+# as defaulted parameters either on the client stubs so that older servers still get the from them still needed parameters.
```
ThomasWaldmann: either ... or ???

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+    'put': ["id_", "data"],
+    'get': ["id_"],
+    'delete': ["id_"],
+    'save_key': ["keydata"],
+    'load_key': [],
+    'break_lock': [],
+    'negotiate': ['client_data'],
+    'open': ['path', 'create', 'lock_wait', 'lock', 'exclusive', 'append_only'],
+    'get_free_nonce': [],
+    'commit_nonce_reservation': ['next_unreserved', 'start_nonce']
```
ThomasWaldmann: please use single-quotes for string literals. do a global search within your changeset.

ThomasWaldmann: tuples rather than lists?

src/borg/remote.py
```diff
@@ -53,6 +55,52 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+}
+
+
+def decodeKeys(d):
+    r = {}
+    for (k, v) in d.items():
+        r[k.decode("utf-8")] = v
+    return r
```
ThomasWaldmann: return {k.decode(): d[k] for k in d}

src/borg/remote.py
```diff
@@ -78,6 +126,22 @@ def __init__(self, restrict_to_paths, append_only):
         self.repository = None
         self.restrict_to_paths = restrict_to_paths
         self.append_only = append_only
+        self.client_version = (1, 1, 0)
```
ThomasWaldmann: another place with a version tuple to maintain.

src/borg/remote.py
```diff
@@ -78,6 +126,22 @@ def __init__(self, restrict_to_paths, append_only):
...
+
+    def positionalToNamed(self, method, argv):
+        kwargs = {}
+        for (pos, name) in enumerate(compatMap[method]):
+            kwargs[name] = argv[pos]
+
```
ThomasWaldmann: remove empty line

src/borg/remote.py
```diff
@@ -78,6 +126,22 @@ def __init__(self, restrict_to_paths, append_only):
...
+        return kwargs
+
+    def filterArgs(self, f, kwargs):
+        filtered = {}
+        known = set(inspect.signature(f).parameters)
+        for (name, value) in kwargs.items():
+            if name in known:
+                filtered[name] = value
+        return filtered
```
ThomasWaldmann: return {name: kwargs[name] for name in kwargs if name in known}

src/borg/remote.py
```diff
@@ -106,20 +170,29 @@ def serve(self):
                     return
                 unpacker.feed(data)
                 for unpacked in unpacker:
-                    if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
+                    if isinstance(unpacked, dict):
+                        dictFormat = True
+                        msgid = unpacked[b"i"]
+                        method = unpacked[b"m"].decode('ascii')
+                        args = decodeKeys(unpacked[b"a"])
```
ThomasWaldmann: ok, i see you're trying to save some bytes on the wire here by using short names. does it make a difference?

src/borg/remote.py
```diff
@@ -78,6 +126,22 @@ def __init__(self, restrict_to_paths, append_only):
...
+        self.client_version = (1, 1, 0)
+
+    def positionalToNamed(self, method, argv):
+        kwargs = {}
+        for (pos, name) in enumerate(compatMap[method]):
+            kwargs[name] = argv[pos]
+
+        return kwargs
+
+    def filterArgs(self, f, kwargs):
```
ThomasWaldmann: docstring?

src/borg/remote.py
```diff
@@ -78,6 +126,22 @@ def __init__(self, restrict_to_paths, append_only):
         self.repository = None
         self.restrict_to_paths = restrict_to_paths
         self.append_only = append_only
+        self.client_version = (1, 1, 0)
+
+    def positionalToNamed(self, method, argv):
```
ThomasWaldmann: docstring?

src/borg/remote.py
```diff
@@ -137,18 +210,29 @@ def serve(self):
...
+                        if dictFormat:
+                            os.write(stdout_fd, msgpack.packb({b'i': msgid, b'exception_class': e.__class__.__name__, b'exception_args': e.args}))
+                        else:
+                            os.write(stdout_fd, msgpack.packb((1, msgid, e.__class__.__name__, exc)))
                     else:
-                        os.write(stdout_fd, msgpack.packb((1, msgid, None, res)))
+                        if dictFormat:
+                            os.write(stdout_fd, msgpack.packb({b'i': msgid, b'r': res}))
+                        else:
+                            os.write(stdout_fd, msgpack.packb((1, msgid, None, res)))
```
ThomasWaldmann: maybe only if/else the arg for packb and only have 1 os.write(stdout_fd, msgpack.packb(arg))?

src/borg/remote.py
```diff
@@ -137,18 +210,29 @@ def serve(self):
...
                 return
 
-    def negotiate(self, versions):
-        return RPC_PROTOCOL_VERSION
+    def negotiate(self, client_data):
+        if client_data == RPC_PROTOCOL_VERSION:
+            return RPC_PROTOCOL_VERSION
+        elif isinstance(client_data, dict):
+            self.client_version = client_data[b"client_version"]
+            return { "server_version": BORG_VERSION }
```
ThomasWaldmann: no in-parens spaces. (please fix globally within your changeset)

src/borg/remote.py
```diff
@@ -200,6 +284,29 @@ def write(self, fd, to_send):
         return written
 
 
+def api(*, since, **kwargs):
+    def decorator(f):
```
ThomasWaldmann: docstring?

src/borg/remote.py
```diff
@@ -200,6 +284,29 @@ def write(self, fd, to_send):
...
+            named = {}
+            for name, param in sig.parameters.items():
+                if name == 'self':
+                    continue
+                if name in bound_args.arguments:
+                    named[name] = bound_args.arguments[name]
+                else:
+                    if param.default is not param.empty:
+                        named[name] = param.default
+
```
ThomasWaldmann: remove empty line?

src/borg/remote.py
```diff
@@ -222,6 +329,8 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
 
         self.unpacker = msgpack.Unpacker(use_list=False)
+        self.dictFormat = False
+        self.server_version = (1, 1, 0)
```
ThomasWaldmann: another place...

src/borg/remote.py
```diff
@@ -342,7 +459,13 @@ def ssh_cmd(self, location):
             args.append('%s' % location.host)
         return args
 
-    def call(self, cmd, *args, **kw):
+    def namedToPositional(self, method, kwargs):
+        argv = []
+        for name in compatMap[method]:
+            argv.append(kwargs[name])
+        return argv
```
ThomasWaldmann: return [kwargs[name] for name in compatMap[method]]

src/borg/remote.py
```diff
@@ -356,36 +479,38 @@ def pop_preload_msgid(chunkid):
                 del self.chunkid_to_msgids[chunkid]
             return msgid
 
-        def handle_error(error, res):
-            error = error.decode('utf-8')
-            if error == 'DoesNotExist':
+        def handle_error(error, args):
+            if error == b'DoesNotExist':
```
ThomasWaldmann: can we rather have a decode() than lots of b'...'?

src/borg/remote.py
```diff
@@ -356,36 +479,38 @@ def pop_preload_msgid(chunkid):
...
+            elif error == b'ObjectNotFound':
+                raise Repository.ObjectNotFound(args[0], self.location.orig)
+            elif error == b'InvalidRPCMethod':
+                raise InvalidRPCMethod(*args)
             else:
-                raise self.RPCError(res.decode('utf-8'), error)
+                if isinstance(args, bytes):
+                    raise self.RPCError(args.decode('utf-8'))
+                else:
+                    raise self.RPCError(args[0].decode('utf-8'))
```
ThomasWaldmann: TIL utf-8 is default for decode() in py3.

src/borg/remote.py
```diff
@@ -425,22 +557,28 @@ def handle_error(error, res):
...
                                 waiting_for.append(self.msgid)
-                                self.to_send = msgpack.packb((1, self.msgid, cmd, args))
+                                if self.dictFormat:
+                                    self.to_send = msgpack.packb({b'i': self.msgid, b'm': cmd, b'a': args})
+                                else:
+                                    self.to_send = msgpack.packb((1, self.msgid, cmd, self.namedToPositional(cmd, args)))
                     if not self.to_send and self.preload_ids:
                         chunk_id = self.preload_ids.pop(0)
-                        args = (chunk_id,)
+                        args = {'id_': chunk_id}
```
ThomasWaldmann: just 'id'? or 'cid'?

src/borg/remote.py
```diff
@@ -425,22 +557,28 @@ def handle_error(error, res):
...
                         self.msgid += 1
                         self.chunkid_to_msgids.setdefault(chunk_id, []).append(self.msgid)
-                        self.to_send = msgpack.packb((1, self.msgid, 'get', args))
+                        if self.dictFormat:
+                            self.to_send = msgpack.packb({b'i': self.msgid, b'm': 'get', b'a': args})
+                        else:
+                            self.to_send = msgpack.packb((1, self.msgid, 'get', self.namedToPositional(cmd, args)))
```
ThomasWaldmann: see above.

src/borg/remote.py
```diff
@@ -452,23 +590,29 @@ def handle_error(error, res):
...
 
+    @api(since=(1, 0, 0))
     def check(self, repair=False, save_space=False):
-        return self.call('check', repair, save_space)
+        pass
 
+    @api(since=(1, 0, 0))
     def commit(self, save_space=False):
-        return self.call('commit', save_space)
+        pass
```
ThomasWaldmann: instead of having a "pass" as placeholder in an otherwise empty block, it is nice to use a docstring.

src/borg/remote.py
```diff
@@ -302,6 +304,13 @@ def do_rpc(self, *args, **kwargs):
 
             if self.server_version < since:
                 raise self.RPCError("Server too old. Need at least: " + ".".join([str(c) for c in since]))
+
+            for name, restriction in kwargs.items():
+                if 'previously' in restriction and named[name] == restriction['previously']:
+                    continue
```
ThomasWaldmann: ???

src/borg/remote.py
```diff
@@ -599,7 +599,8 @@ def handle_error(error, args):
                             raise
         self.ignore_responses |= set(waiting_for)
 
-    @api(since=(1, 0, 0))
+    @api(since=(1, 0, 0),
+         append_only={'since':(1,0,7), 'prevously':False})
```
ThomasWaldmann: add a blank after : and after ,

src/borg/remote.py
textshell: Yes. I need to figure out how to get the setuptools maintained version here.

src/borg/remote.py
textshell: This one actually is the last borg version that used the old protocol. This is duplicated once in the RemoteRepository code so this still needs to be improved. But it‘s not something we need to maintain. But i should add a comment.

src/borg/remote.py
textshell: The problem here is that error sometimes isn‘t even a string, iirc. The error propagation code is somewhat strange in current borg. Maybe this needs a good look again and not just incremental patching.

src/borg/remote.py
textshell: Yes, i really should make use of that more.

textshell:
ok, i answered the easy ones for now. 

If both of you are ok with the general direction of this, i’m happy to start polishing it.

@enkore 
Can you have a look, if this seems good from basic protocol / semantics point of view.

ThomasWaldmann:
@textshell rebase on master again, so the unrelated changesets vanish from this PR?

ThomasWaldmann:
remote.py:385 syntax error
flake8 is rather unhappy (some of the stuff I already noted).

src/borg/remote.py
textshell: Currently this uses 3-tuples for versioning. But setuptools_scm gives strings. Do we need something special for parsing them into nicely comparable tuples? How would we like to handle prereleases?

We could just pass the version, but then we are in trouble if we ever want to change the version format. These tend to be not purely technical in nature.

textshell:
not yet everything done. But at least a life sign.

I removed the version spec for the open call for now because it was mostly an example and the workarounds since used in 1.0.x are not straightforward to fit into this framwork.

ThomasWaldmann:
@enkore some feedback about basic protocol / semantics from you would be useful.

src/borg/remote.py
enkore: #1752 added this.

src/borg/remote.py
```diff
@@ -53,6 +55,50 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+    'put': ('id_', 'data'],
+    'get': ('id_'],
+    'delete': ('id_'],
+    'save_key': ('keydata'],
+    'load_key': (),
+    'break_lock': (),
+    'negotiate': ('client_data'),
+    'open': ('path', 'create', 'lock_wait', 'lock', 'exclusive', 'append_only'),
+    'get_free_nonce': (),
+    'commit_nonce_reservation': ('next_unreserved', 'start_nonce')
```
enkore: ^- some syntax errors there with `(` vs `]`

src/borg/remote.py
```diff
@@ -53,6 +55,50 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+# inclusive, or it is a dict which includes the server version.
+#
+# All method calls on the remote repository object must be whitelisted in RepositoryServer.rpc_methods and have api
+# stubs in RemoteRepository. The @api decorator on these stubs is used to set server version requirements.
+#
+# Method parameters are identfied only by name and never by position. Unknown parameters are ignored by the server side.
+# If a new parameter is important and may not be ignored, on the client a parameter specific version requirement needs
+# to be added.
+# When parameters are removed, they need to be preserved  as defaulted parameters on the client stubs so that older
+# servers still get the from them still needed parameters.
```
enkore: excellent

src/borg/remote.py
```diff
@@ -136,19 +205,30 @@ def serve(self):
                             tb = '%s\n%s' % (traceback.format_exc(), sysinfo())
                             logging.error(msg)
                             logging.log(tb_log_level, tb)
-                        exc = "Remote Exception (see remote log for the traceback)"
-                        os.write(stdout_fd, msgpack.packb((1, msgid, e.__class__.__name__, exc)))
+                        exc = 'Remote Exception (see remote log for the traceback)'
+                        if dictFormat:
+                            os.write(stdout_fd, msgpack.packb({b'i': msgid, b'exception_class': e.__class__.__name__, b'exception_args': e.args}))
+                        else:
+                            os.write(stdout_fd, msgpack.packb((1, msgid, e.__class__.__name__, exc)))
```
enkore: excellent, this then also fixes the rpc-exception-args-transporting issue that I dropped from #1685

src/borg/remote.py
```diff
@@ -1,5 +1,6 @@
 import errno
 import fcntl
+import inspect, functools
```
enkore: also most new functions and variables are camelCase, while snake_case is the usual choice here.

ThomasWaldmann:
@enkore so, your POV is that this change should be merged (after some fixing and fine tuning)? Guess that is what @textshell wants to know before investing more work.

src/borg/remote.py
enkore: The API is documented in Repository, though. This is just another implementation, there shouldn't be any changes, API-wise.

enkore: I guess this could be written as `commit = api(...)`, but it'll be less obvious, especially to IDEs.

enkore:
This would be a very valuable addition to 1.1, since we're trying to get that into LTS of the distros. Having a more flexible RPC protocol, and one that doesn't print spurious tracebacks to the console, would allow to avoid ugly hacks for API changes.

The protocol looks good to me.

The argument mapping is a bit non-trivial, but looks good as well.

The new version negotiation should work much better than the older one, which was just a fake version anyway. The only place where that one might get a bit bumpy is if we have multiple supported release lines with this code and backport features between them (blargh).

src/borg/remote.py
```diff
@@ -53,6 +55,50 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+compatMap = {
+    'check' : ('repair', 'save_space'],
+    'commit': ('save_space'],
+    'rollback': (),
+    'destroy': (),
+    '__len__': (),
+    'list': ('limit', 'marker'],
+    'put': ('id_', 'data'],
+    'get': ('id_'],
+    'delete': ('id_'],
```
enkore: `id_` vs `id` has been inconsistent in the repo API for a while, even base Repository has this (get has id_, put/delete have id). Fix it (also in repostory.py) and always go for `id` (or `id_`, but consistently)?

ThomasWaldmann: id is definitely prettier, but might cause some complaint from linters because it shadows a builtin.

if we want to avoid that, I'ld rather use a more specific word than just appending a underscore. most such "shadows builtin" issues can be fixed by being more specific. So, not "id", but "obj_id" or "oid" or ...

otoh, we could also switch off the linter warning.

enkore: obj(ect)_id would make a clearer distinction when named object APIs are considered.

ThomasWaldmann: @rumpelsepp then you might just get another warning from the pep8 checker. ;)

ThomasWaldmann:
From IRC:
```
textshell$ I think there are 3 open questions i was going to to look at that i can remember.
1) how to get from the borg version string to a nice tuple or if we would work with the string directly.
2) how to fix the 1.0.7 hack about repository opening with fallback and
3) if there is anything reasonable in this timeframe to improve in exception propagation. Exception propagation seem to be somewhat broken.
```

ThomasWaldmann:
@textshell : 
- the version tuple needs a rebase on current master
- see XXX in remote.py / see travis

textshell:
I rebased this to current master.
I included a proposal for handling versions from setuptools_scm. The current code parses the version to a tuple that mostly works with normal python comparison. The other approach would to use a nicer tuple format and a custom comparison function.
I think i got a decent solution for the open hacks for 1.0.6 vs 1.0.7. This also fixes the 'restrictions' handing in the api decorator. But i don‘t have a readily setup 1.0.6 to test against.

Fixing Exception handling is still pending. Also i just noticed i reintroduced double quote, need to fix that again too.

src/borg/remote.py
```diff
@@ -647,31 +647,31 @@ def handle_error(error, args):
 
     @api(since=parse_version('1.0.0'))
     def open(self, path, create=False, lock_wait=None, lock=True, exclusive=False, append_only=False):
-        pass
+        """actual remoting is done via self.call in the @api decorator"""
```
ThomasWaldmann: hmm, not sure this is helpful - especially not if we generate api docs from that.

src/borg/remote.py
```diff
@@ -270,11 +270,17 @@ def serve(self):
                 return
 
     def negotiate(self, client_data):
+        # old format used in 1.0.x
         if client_data == RPC_PROTOCOL_VERSION:
             return RPC_PROTOCOL_VERSION
-        elif isinstance(client_data, dict):
+        # clients since 1.1.0b3 use an dict as client_data
```
ThomasWaldmann: ... a dict ...

src/borg/remote.py
```diff
@@ -338,8 +338,22 @@ def write(self, fd, to_send):
         return written
 
 
-def api(*, since, **kwargs):
-    """TBD"""
+def api(*, since, **kwargs_decorator):
+    """Check version requirements and use self.call to remote method call.
```
ThomasWaldmann: ... to do the remote method call.

src/borg/remote.py
```diff
@@ -338,8 +338,22 @@ def write(self, fd, to_send):
...
+
+    since specifies the version in which borg introduced this method,
```
ThomasWaldmann: <since>

src/borg/remote.py
```diff
@@ -338,8 +338,22 @@ def write(self, fd, to_send):
...
+    calling this method when connected with an older version will fail without transmiting
```
ThomasWaldmann: ... connected to ...

src/borg/remote.py
```diff
@@ -338,8 +338,22 @@ def write(self, fd, to_send):
...
+    anything to the server.
+
+    Further kwargs can be used to encode version specific restrictions.
+    If an previous hardcoded behaviour is parameterized in a version this allows calls that
```
ThomasWaldmann: If a ... parametrized ..., this ...

src/borg/remote.py
```diff
@@ -338,8 +338,22 @@ def write(self, fd, to_send):
...
+    use the previously hardcoded behaviour to pass through and generates an error if another
+    behaviour is requested by the client.
+
+    e.g. when 'append_only' was introduced in 1.0.7 the previous behaviour was what now is append_only=False.
+    Thus @api(..., append_only={'since': parse_version('1.0.7'), 'previously': False}) allows calls
+    with append_only=False for all version but rejects calls using append_only=True on versions older than 1.0.7.
```
ThomasWaldmann: \o/

src/borg/remote.py
textshell: I think it‘s more helpful than pass. But it could also be a comment and a pass. 

I think adding real documentation about the method here would be bad and it would quickly go stale. We might want to document differences or compatibility constraints that differ from the real Repository class though.

src/borg/remote.py
```diff
@@ -77,6 +77,57 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+    version = [int(segment) for segment in parts[:-1]]
+
+    prerelease = re.fullmatch('([0-9]+)(a|b|rc)([0-9]+)', parts[-1])
+    if prerelease:
+        version_type = {'a': -4, 'b': -3, 'rc': -2}[prerelease.group(2)]
+        version += [int(prerelease.group(1)), version_type, int(prerelease.group(3))]
+    else:
+        version += [int(parts[-1]), -1]
+
+    return tuple(version)
```
textshell: ok, i moved them to the new version module.

src/borg/remote.py
textshell: fixed

src/borg/remote.py
textshell: fixed

src/borg/remote.py
textshell: fixed

src/borg/remote.py
textshell: fixed

textshell:
This includes the first stab at doing something sensible with the exceptions.
Basic idea is to move as much as possible to the client. Thus the server only logs on debug (except for problems in negotiate that currently always is done in legacy mode) and sends exception message, full exception incl. traceback, exception class name and sysinfo to the client.
The client reconstructs the whitelisted exception types (but with correct arguments) and wraps everything else into RPCError exception that is printed in archiver.

src/borg/archiver.py
```diff
@@ -2733,10 +2733,16 @@ def main():  # pragma: no cover
...
-            important = e.remote_type not in ('LockTimeout', )
+            important = e.exception_class not in ('LockTimeout', )
             tb_log_level = logging.ERROR if important else logging.DEBUG
-            tb = sysinfo()
+            if important:
+                msg = e.exception_full
+            else:
+                msg = e.get_message()
+            tb = ""
+            for line in e.sysinfo.strip(('\n')).split('\n'):
```
ThomasWaldmann: the parens seem to indicate that you had a tuple of characters here, but with only 1 element and no comma, it is no tuple any more. also .splitlines() maybe?

src/borg/remote.py
```diff
@@ -122,7 +122,8 @@ class RepositoryServer:  # pragma: no cover
         'load_key',
         'break_lock',
         'get_free_nonce',
-        'commit_nonce_reservation'
+        'commit_nonce_reservation',
+        'inject_exception'
```
ThomasWaldmann: it's good style to always have element lines ending with a comma, even if it is the last, so that such changes are only +1 (and not -1 +2).

src/borg/remote.py
```diff
@@ -342,9 +380,44 @@ class RemoteRepository:
...
+                return b'\n'.join(self.unpacked[b'exception_full']).decode()
+            else:
+                return self.get_message() + '\nRemote Exception (see remote log for the traceback)'
+
+        @property
+        def sysinfo(self):
+            if b'sysinfo' in self.unpacked:
+                return self.unpacked[b'sysinfo'].decode()
+            else:
+                return ''
```
ThomasWaldmann: either have all as properties or all as methods (see get_message())?

src/borg/archiver.py
```diff
@@ -2733,10 +2733,16 @@ def main():  # pragma: no cover
...
         except RemoteRepository.RPCError as e:
-            msg = "%s %s" % (e.remote_type, e.name)
-            important = e.remote_type not in ('LockTimeout', )
+            important = e.exception_class not in ('LockTimeout', )
             tb_log_level = logging.ERROR if important else logging.DEBUG
-            tb = sysinfo()
+            if important:
+                msg = e.exception_full
+            else:
+                msg = e.get_message()
```
ThomasWaldmann: this is a bit unpretty due to the inconsistency, one is a method, the other a property...

src/borg/archiver.py
```diff
@@ -2733,10 +2733,16 @@ def main():  # pragma: no cover
...
+            tb = ""
+            for line in e.sysinfo.strip(('\n')).split('\n'):
+                tb += "Borg server: " + line + '\n'
+            tb += "\n" + sysinfo()
```
```
>>> '\n'.join('Borg server: ' + l for l in t.splitlines())
'Borg server: a\nBorg server: b\nBorg server: c\nBorg server: d'
```

src/borg/remote.py
```diff
@@ -53,6 +56,54 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+# inclusive, or it is a dict which includes the server version.
+#
+# All method calls on the remote repository object must be whitelisted in RepositoryServer.rpc_methods and have api
+# stubs in RemoteRepository. The @api decorator on these stubs is used to set server version requirements.
+#
+# Method parameters are identified only by name and never by position. Unknown parameters are ignored by the server side.
+# If a new parameter is important and may not be ignored, on the client a parameter specific version requirement needs
+# to be added.
+# When parameters are removed, they need to be preserved as defaulted parameters on the client stubs so that older
+# servers still get the from them still needed parameters.
```
ThomasWaldmann: if this was a docstring, it could end up in some auto-generated docs.

src/borg/remote.py
```diff
@@ -53,6 +56,54 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+
+compatMap = {
+    'check': ('repair', 'save_space', ),
+    'commit': ('save_space', ),
+    'rollback': (),
+    'destroy': (),
+    '__len__': (),
+    'list': ('limit', 'marker', ),
+    'put': ('id', 'data', ),
+    'get': ('id_', ),
```
ThomasWaldmann: put -> id
get -> id_

make both "id"?

src/borg/remote.py
```diff
@@ -101,54 +163,97 @@ def serve(self):
...
+                        os.write(stderr_fd, 'Borg {}: Got connection close before repository was opened.\n'
                                  .format(__version__).encode())
                     return
                 unpacker.feed(data)
                 for unpacked in unpacker:
-                    if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
+                    if isinstance(unpacked, dict):
+                        dictFormat = True
+                        msgid = unpacked[b'i']
+                        method = unpacked[b'm'].decode('ascii')
```
ThomasWaldmann: could be just .decode() - would be utf-8 then, but that is a superset of ascii, so no problem.

src/borg/remote.py
```diff
@@ -101,54 +163,97 @@ def serve(self):
...
+                        args = decodeKeys(unpacked[b'a'])
+                    elif isinstance(unpacked, tuple) and len(unpacked) == 4:
+                        dictFormat = False
+                        type, msgid, method, args = unpacked
+                        method = method.decode('ascii')
```
ThomasWaldmann: same here (maybe do a search for this).

src/borg/remote.py
```diff
@@ -53,6 +56,54 @@ class UnexpectedRPCDataFormatFromServer(Error):
...
+    'load_key': (),
+    'break_lock': (),
+    'negotiate': ('client_data', ),
+    'open': ('path', 'create', 'lock_wait', 'lock', 'exclusive', 'append_only', ),
+    'get_free_nonce': (),
+    'commit_nonce_reservation': ('next_unreserved', 'start_nonce', ),
+}
+
+
+def decodeKeys(d):
```
ThomasWaldmann: decode_keys?

src/borg/remote.py
```diff
@@ -171,6 +276,27 @@ def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, ap
         self.repository.__enter__()  # clean exit handled by serve() method
         return self.repository.id
 
+    def inject_exception(self, kind):
```
ThomasWaldmann: looks like test code? move to test module?

src/borg/remote.py
```diff
@@ -202,16 +328,96 @@ def write(self, fd, to_send):
         return written
 
 
+def api(*, since, **kwargs_decorator):
+    """Check version requirements and use self.call to do the remote method call.
+
+    since specifies the version in which borg introduced this method,
```
ThomasWaldmann: My past comment that got unreadable was meant like this:
```
<since> specifies the ...
```

src/borg/remote.py
```diff
@@ -358,36 +583,51 @@ def pop_preload_msgid(chunkid):
                 del self.chunkid_to_msgids[chunkid]
             return msgid
 
-        def handle_error(error, res):
-            error = error.decode('utf-8')
+        def handle_error(unpacked):
+            error = unpacked[b'exception_class'].decode()
+            old_server = b'exception_res' in unpacked
+            args = unpacked.get(b'exception_args', None)
```
ThomasWaldmann: None is the default default for .get()

src/borg/remote.py
```diff
@@ -358,36 +583,51 @@ def pop_preload_msgid(chunkid):
...
             if error == 'DoesNotExist':
                 raise Repository.DoesNotExist(self.location.orig)
             elif error == 'AlreadyExists':
                 raise Repository.AlreadyExists(self.location.orig)
             elif error == 'CheckNeeded':
                 raise Repository.CheckNeeded(self.location.orig)
             elif error == 'IntegrityError':
-                raise IntegrityError(res)
+                if old_server:
+                    raise IntegrityError('<not available>')
```
ThomasWaldmann: bad if that msg gets posted to some stupid pastebin that (like github reviews) interprets everything in `<...>` as html or formatting tag and makes everything unknown invisible.

src/borg/remote.py
```diff
@@ -426,23 +673,29 @@ def handle_error(error, res):
                 while not self.to_send and (calls or self.preload_ids) and len(waiting_for) < MAX_INFLIGHT:
                     if calls:
                         if is_preloaded:
-                            assert cmd == "get", "is_preload is only supported for 'get'"
-                            if calls[0][0] in self.chunkid_to_msgids:
-                                waiting_for.append(pop_preload_msgid(calls.pop(0)[0]))
+                            assert cmd == 'get', "is_preload is only supported for 'get'"
+                            if calls[0]['id_'] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(calls.pop(0)['id_']))
```
ThomasWaldmann: maybe at the end check all id_ if they are correct or whether they should be id.

src/borg/archiver.py
textshell: Yes, that’s more concise.

src/borg/archiver.py
textshell: see above.

src/borg/archiver.py
textshell: if your prefer. It‘s a bit dense for my taste.

src/borg/remote.py
textshell: We can do a proper file level doc string in a follow up PR?

src/borg/remote.py
textshell: Yes. I now added a new commit that changes Repository to use `id`. But this commit had only the changes that where already made in Repository.

src/borg/remote.py
textshell: was somehow missed before, fixed.

src/borg/remote.py
textshell: I wonder if we want to be able to test this against older borg versions, it‘s hard to unit test this completely. If not i can have a look how to move this into a subclass in the testsuite.

src/borg/remote.py
textshell: ok. But i don‘t think we use this style consistently yet.

src/borg/remote.py
textshell: fixed and squashed

src/borg/remote.py
textshell: They were. But now Repository is fixed, corrected that in a follow up commit.

src/borg/remote.py
enkore: I think in Borg we mostly used either the olde Sphinx style (those ugly ":param foo: foo is bar" blocks) or the `*foo* is the bar to ...` style.

Personally I very much prefer the "write a complete sentence mentioning the parameters block" method.

textshell:
This is taking shape nicely now.

One thing i noticed in reviewing the code is that currently rollback is dubious, because it previously just discards the cleanup parameter (or rather all parameters). Not sure what this does with the new infrastructure. Is this on purpose?
If so, i need to look at how to model that.

enkore:
Repository.rollback's cleanup parameter is internal and wasn't meant as part of the API when I added it.

textshell:
Ok. Created #1826 to remove the cleanup parameter from Repository.rollback's public api. So remove it from the RemoteRepository too.

textshell:
Ok, this should now be fully featured and from my perspective the rpc wire protocol should be good to go.

One thing i considered but did not do was to use use_bin_type=True / encoding='utf-8' with msgpack. But i thought at that time this would be to radical. But the biggest risk from my current point of view is that that this would make introducing compatibility regressions easier (and possible further delays).

@enkore, @ThomasWaldmann: Are you happy with the new wire format? Because changing it soon after 1.1 again would be a PITA.

Apart from that the exception printing should be checked, if it looks like we want. But we can change that later anyway, as long as we have enough info in the wire format.

I don‘t have a 1.0.6 server handy to test if the fallback code actually works. It would be great if someone could test that and report back.

Also we need to test every rpc method using 1.0.8 and a server from this branch. I might have missed something.

Do we want to check for performance regressions?

src/borg/remote.py
ThomasWaldmann: sure. but maybe fix the last sentence now, it sounds broken.

src/borg/remote.py
```diff
@@ -101,54 +161,97 @@ def serve(self):
...
                 for unpacked in unpacker:
-                    if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
+                    if isinstance(unpacked, dict):
+                        dictFormat = True
+                        msgid = unpacked[b'i']
+                        method = unpacked[b'm'].decode()
+                        args = decode_keys(unpacked[b'a'])
+                    elif isinstance(unpacked, tuple) and len(unpacked) == 4:
+                        dictFormat = False
+                        type, msgid, method, args = unpacked
```
ThomasWaldmann: type is not used, is this correct?
if so, use _ to indicate that?

src/borg/remote.py
```diff
@@ -253,39 +462,54 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
...
+                self.dictFormat = False
+            elif isinstance(version, dict) and b'server_version' in version:
+                self.dictFormat = True
+                self.server_version = version[b'server_version']
+            else:
+                raise Exception('Server insisted on using unsupported protocol version %s' % version)
+
+            def do_open():
+                self.id = self.open(path=self.location.path, create=create, lock_wait=lock_wait,
+                                lock=lock, exclusive=exclusive, append_only=append_only)
```
ThomasWaldmann: use visual indent? (align lock under path)

src/borg/remote.py
```diff
@@ -406,15 +645,23 @@ def handle_error(error, res):
                         raise ConnectionClosed()
                     self.unpacker.feed(data)
                     for unpacked in self.unpacker:
-                        if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
+                        if isinstance(unpacked, dict):
+                            msgid = unpacked[b'i']
+                        elif isinstance(unpacked, tuple) and len(unpacked) == 4:
+                            type, msgid, error, res = unpacked
```
ThomasWaldmann: type not used here either (see other comment).

src/borg/remote.py
```diff
@@ -426,23 +673,29 @@ def handle_error(error, res):
...
-                            if cmd == 'get' and args[0] in self.chunkid_to_msgids:
-                                waiting_for.append(pop_preload_msgid(args[0]))
+                            if cmd == 'get' and args['id'] in self.chunkid_to_msgids:
+                                waiting_for.append(pop_preload_msgid(args['id']))
                             else:
                                 self.msgid += 1
                                 waiting_for.append(self.msgid)
-                                self.to_send = msgpack.packb((1, self.msgid, cmd, args))
+                                if self.dictFormat:
+                                    self.to_send = msgpack.packb({b'i': self.msgid, b'm': cmd, b'a': args})
```
ThomasWaldmann: Hmm, how about defining
MSGID, MSG, ARGS = b'i', b'm', b'a'
and using the constants everywhere?

ThomasWaldmann:
./src/borg/testsuite/repository.py:718:1: E302 expected 2 blank lines, found 1

./src/borg/testsuite/repository.py:742:1: E302 expected 2 blank lines, found 1

src/borg/remote.py
enkore: This actually comes from attic; both client and server send 1s here, but don't do anything with that.

textshell: It’s the only documentation what that field is. But i guess we don‘t really need to remember.

enkore: Let's just stick a comment to it? Something like `# This field was always 1 and has always been ignored`

src/borg/remote.py
```diff
@@ -204,12 +204,23 @@ def serve(self):
...
+                                msg = msgpack.packb({MSGID: msgid,
+                                                    b'exception_class': e.__class__.__name__,
+                                                    b'exception_args': e.args,
+                                                    b'exception_full': ex_full,
+                                                    b'exception_short': ex_short,
+                                                    b'sysinfo': sysinfo()})
+                            except TypeError:
+                                msg = msgpack.packb({MSGID: msgid,
+                                                    b'exception_class': e.__class__.__name__,
+                                                    b'exception_args': [x if isinstance(x, (str, bytes, int)) else None
```
ThomasWaldmann: hmm, `else repr(x)` maybe?

ThomasWaldmann: also, maybe just do that in first try and remove exception handler again?

textshell: I think for filtering on the first try the whitelist has to be better(i.e. support basic lists maybe) which would be more complicated code. Enkore’s first suggestion even was to just drop the args completely if they are troublesome. I’m not sure about the value of using repr instead of None, because these should end up in readable form in exception_full for human debugging anyway. exception_args is for safe automated processing and for recreating exceptions. Both not best served by repr·s.

Anyway this is only a fallback path, i don‘t think we have any exceptions that have information in non msgpackable types yet. And if we add new ones we might want to revisit this anyway.

textshell:
squashed and rebased to master.

ThomasWaldmann:
@enkore anything left from your PoV?

Issue #536, opened 2016-01-09
hashtable local performance breakdowns


ThomasWaldmann:
seems like some circumstance leads to local performance breakdowns depending on the amount N (and kind?) of input and the hashtable size T while usually giving good performance.

e.g.:

```
T = 262147 (initial size, grows when load>0.75)
H1.merge(H2) times:
N=100000 16ms
N=200000 108ms
N=240000 18385ms
```

Having a smaller max load factor seems to help a bit, but also seems to make the problem just less likely / less intense(?), but does not make it go away. Also, we should not go too low with the max load factor to not waste memory / get out-of-memory conditions.

I am wondering a bit, what causes this / what to do against it.

The hash function is taking the upper 32bits of the sha256 and computes modulo T to get the bucket index. T was even prime in one of my experiments.

Source: borg/_hashindex.c

ThomasWaldmann:
http://planetmath.org/goodhashtableprimes interesting, esp. property 3:

    1.    each number in the list is prime
    2.    each number is slightly less than twice the size of the previous
    3.    each number is as far as possible from the nearest two powers of two

ThomasWaldmann:
https://github.com/fragglet/c-algorithms includes src/hash-table.c/h implementing separate chaining with linked lists.

good:
- little memory needs for the hashtable itself (N * pointersize), so we can afford growing early (author grows table with load >0.33!)

bad:
- lots of malloc(), lots of tiny objects
- harder/slower to (un)serialize for disk storage

ThomasWaldmann:
http://blogs.msdn.com/b/dcook/archive/2007/09/09/hashes-and-tables-and-primes-oh-my.aspx anti-prime

ThomasWaldmann:
currently, it seems to me that all the "magic" rules and mechanisms are just there to cope with:
- bad hash functions that don't distribute well (does not apply as we have a cryptographic hash to start from)
- "open addressing" collision handling using linear probing with increment > 1 (does not apply, we use 1)

ThomasWaldmann:
Sounds promising:

http://sebastiansylvan.com/post/robin-hood-hashing-should-be-your-default-hash-table-implementation/

infectormp:
may be out of topic, but maybe we can use jemalloc to improve performance?

ThomasWaldmann:
[...]

PR #1429, opened 2016-08-02
WIP: Robin Hood hashing

rciorba:
This changes the collision handling strategy in _hashindex.c to
robin-hood hashing. Incidentally this should mutually be compatible
with the old version.

This hasn't been properly tested yet, except by the unittests, but
I'm sharing this early to get some feedback. Some testing and a
before and after performance measurement will follow.

Also please note this temporarily disables a couple of tests that
hard coded hashes that now no longer match.


src/borg/_hashindex.c
```diff
@@ -68,7 +68,7 @@ static int hash_sizes[] = {
 };
 
 #define HASH_MIN_LOAD .25
-#define HASH_MAX_LOAD .75  /* don't go higher than 0.75, otherwise performance severely suffers! */
+#define HASH_MAX_LOAD 0.95  /* don't go higher than 0.75, otherwise performance severely suffers! */
```
PlasmaPower: Why is this changed? Also, make sure to update the comment.

rciorba: > Why is this changed? Also, make sure to update the comment.

The point of the Robin Hood hashing is to minimize the worst case for collisions by spreading the pain across all addresses. This should allow high loads in the hash table without performance degrading much. Also I should add the idea for this change isn't mine, @ThomasWaldmann suggested it as something interesting to do at the EuroPython sprints. 

I intentionally didn't update the comments until I run some benchmarks to find an appropriate value.

> Minor nitpicking, don't put a leading zero if the definition above doesn't have it.

Will do. BTW, the code style for C in this project isn't 100% clear to me, so If there are any other style no-no's in my PR, please let me know.

ThomasWaldmann: Thanks for working on this! :)
If you change HASH_MAX_LOAD, do a full text search for it, there is another place depending on its value.

ThomasWaldmann: Also, change the outdated comment right besides it, otherwise it looks wrong.

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
     return BUCKET_ADDR(index, idx) + index->key_size;
 }
 
+inline
+int
+distance(HashIndex  *index, int current_idx, int ideal_idx)
```
ThomasWaldmann: nitpick: only 1 blank before *index.

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
+        bucket_ptr = BUCKET_ADDR(index, idx);
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
+            /* we have a collision */
+            other_offset = distance(
+                index, idx, hashindex_index(index, bucket_ptr));
+            if ( other_offset < offset) {
+                /* Swap the bucket at idx with the current key/value pair.
+                   This is the gist of hobin-hood hashing, we rob from
+                   the key with the lower distance to it's optimal address
+                   by swaping places with it.
```
ThomasWaldmann: typo: ... its ... swapping ...

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
-        ptr = BUCKET_ADDR(index, idx);
-        memcpy(ptr, key, index->key_size);
-        memcpy(ptr + index->key_size, value, index->value_size);
+        memcpy(bucket_ptr, bucket, (index->key_size + index->value_size));
         index->num_entries += 1;
     }
     else
     {
+        /* we already have the key in the index
+           we just need to update it's value */
```
ThomasWaldmann: typo: its

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
 static int
 hashindex_set(HashIndex *index, const void *key, const void *value)
 {
     int idx = hashindex_lookup(index, key);
-    uint8_t *ptr;
+    uint8_t *bucket_ptr;
+    int offset = 0;
+    int other_offset;
+    void *bucket = malloc(index->key_size + index->value_size);
+    void *buffer = malloc(index->key_size + index->value_size);
```
ThomasWaldmann: it would be cool if we could somehow not allocate/free bucket and buffer once per hashindex_set operation.

maybe the index data structure could have 2 pointers to such buffers that are just allocated once?

src/borg/_hashindex.c
```diff
@@ -126,6 +126,7 @@ hashindex_lookup(HashIndex *index, const void *key)
         }
         else if(BUCKET_MATCHES_KEY(index, idx, key)) {
             if (didx != -1) {
+                /* we found a toombstone earlier, so we can move this key on top of it */
```
ThomasWaldmann: thanks a lot for adding all these comments, very helpful!

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
-    uint8_t *ptr;
+    uint8_t *bucket_ptr;
+    int offset = 0;
+    int other_offset;
+    void *bucket = malloc(index->key_size + index->value_size);
+    void *buffer = malloc(index->key_size + index->value_size);
     if(idx < 0)
     {
+        /* we don't have the key in the index
+           we need to find an appropriate address */
```
ThomasWaldmann: I am not sure whether the C code follows some specific coding style, but the max line length for Python code in borg is 120, thus such a comment would be just a single line.

src/borg/_hashindex.c
rciorba: also, hobbin-hood... seems I should actually use my spell checker :)

rciorba:
Thanks for the feedback so far. I'll follow up once I get some performance numbers in as well.

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
         }
         idx = hashindex_index(index, key);
+        memcpy(bucket, key, index->key_size);
+        memcpy(bucket + index->key_size, value, index->value_size);
+        bucket_ptr = BUCKET_ADDR(index, idx);
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
+            /* we have a collision */
+            other_offset = distance(
+                index, idx, hashindex_index(index, bucket_ptr));
+            if ( other_offset < offset) {
```
ThomasWaldmann: nitpick: no blank after (

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
+}
+
 static int
 hashindex_set(HashIndex *index, const void *key, const void *value)
 {
     int idx = hashindex_lookup(index, key);
-    uint8_t *ptr;
+    uint8_t *bucket_ptr;
+    int offset = 0;
+    int other_offset;
```
ThomasWaldmann: maybe just compute this once:
int entry_size = index->key_size + index->value_size;

and use that at all places below.

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
+                index, idx, hashindex_index(index, bucket_ptr));
+            if ( other_offset < offset) {
+                /* Swap the bucket at idx with the current key/value pair.
+                   This is the gist of hobin-hood hashing, we rob from
+                   the key with the lower distance to it's optimal address
+                   by swaping places with it.
+                 */
+                memcpy(buffer, bucket_ptr, (index->key_size + index->value_size));
+                memcpy(bucket_ptr, bucket, (index->key_size + index->value_size));
+                memcpy(bucket , buffer, (index->key_size + index->value_size));
```
ThomasWaldmann: nitpick: no blank after bucket

ThomasWaldmann: move this into a (inline) memxchg(a, b, tmp, size) function?

src/borg/_hashindex.c
```diff
@@ -376,31 +377,69 @@ hashindex_get(HashIndex *index, const void *key)
...
+                offset = other_offset;
+            } else {
+                offset++;
```
ThomasWaldmann: hmm, you increment offset in this case, as you also increment idx and bucket_ptr below.

but shouldn't the offset derived from other_offset (425) also get incremented for same reason?

src/borg/_hashindex.c
ThomasWaldmann: hmm, just noticed: is it robin' hood because he is robbing? :D

now a native speaker must explain!

ThomasWaldmann:
Really looking forward to the effects of this CS. The paper sounded very promising about performance / efficiency.

ThomasWaldmann:
Related: #536

src/borg/_hashindex.c
rciorba: good catch!

ThomasWaldmann:
@rciorba can you do a separate fixup commit with fixes for all the feedback you got?

rciorba:
Sure thing. I'll have some time later today.

src/borg/_hashindex.c
rciorba: > If you change HASH_MAX_LOAD, do a full text search for it, there is another place depending on its value.

Had, a look. All I can see is the comment next to the value and `docs/internals.rst`.
I would update both once I identify a good value for this constant. Let me know if there's any places I've missed.

ThomasWaldmann: search for 1.35 in the source.

src/borg/_hashindex.c
rciorba: i went for memswap, since it felt like a better name

src/borg/_hashindex.c
```diff
@@ -111,7 +111,7 @@ hashindex_index(HashIndex *index, const void *key)
 static int
 hashindex_lookup(HashIndex *index, const void *key)
```
rciorba: This function could also be optimized. Currently for the worst case scenario (not finding a key in the index) we scan all buckets until we find an empty one. At high fill ratios this might get close to O(N). However if we track the maximum offset in the entire hashmap we could bail after at most max_offset iterations.

As the PR is currently, we could just load an old hashmap and start operating on it with the new collision handling code, and it would just work, and also hashmaps created by this would still be usable by olded borg versions. Changing `hashindex_lookup` however would require us to convert the hashmap explicitly, and also change the HashHeader to track this max offset. That would be a bigger deal because it would impact backwards compatibility, so some planning needs to go in to this.

One potential idea would be to use the MAGIC string in the header to also encode a version. For example if we turn BORG_IDX to BORG_I and 2 bytes for versioning, we could determine if this version of the index is fully robin-hood compliant and if not we could convert it on load from disk.

@ThomasWaldmann I'd like to hear your thoughts on this.

src/borg/_hashindex.c
```diff
@@ -377,69 +377,72 @@ hashindex_get(HashIndex *index, const void *key)
...
     /* If the current index is smaller than the ideal index we've wrapped
        around the end of the bucket array and need to compensate for that. */
     return current_idx - ideal_idx + ( (current_idx < ideal_idx) ? index->num_buckets : 0 );
 }
 
+inline void
+memswap(void *a, void *b, void *tmp, size_t entry_size) {
+    memcpy(tmp, a, (entry_size));
+    memcpy(a, b, (entry_size));
+    memcpy(b, tmp, (entry_size));
```
ThomasWaldmann: why so many parens?

rciorba: because search and replace :)

src/borg/_hashindex.c
```diff
@@ -377,69 +377,72 @@ hashindex_get(HashIndex *index, const void *key)
...
-                offset = other_offset;
+            if(other_offset < offset) {
+                /* Swap the bucket at idx with the current entry_to_insert.
+                   This is the gist of robin-hood hashing, we rob from the key with the
+                   lower distance to its optimal address by swapping places with it. */
+                memswap(bucket_ptr, entry_to_insert, tmp_entry, entry_size);
+                offset = other_offset + 1;
             } else {
                 offset++;
             }
```
ThomasWaldmann: instead of the last 4 lines, you could do:
```
    offset = other_offset;
}
offset++;
```

src/borg/_hashindex.c
rciorba: > we could bail after at most max_offset iterations.
Actually if the offset of the key is smaller than the number of buckets we've looked at we can bail.  There's no way the next bucket will contain out desired key, since it would have been swapped on insert.

ThomasWaldmann: scanning a big part of hashindex sounds evil. guess at 95% fill rate, we would run into having to always scan about 20% of all buckets.

maybe that was the real reason for the perf breakdown i was sometimes seeing?

maybe first keep the code compatible until it is accepted / merged, but we'll keep the idea for later.

rciorba: Well, one way to keep it compatible and still speedup `hashindex_lookup` is to always reinsert all items when loading from disk (no more expensive than a resize). I'll do some measurements of performance with and without this implemented.

ThomasWaldmann:
@rciorba can I help here? did you push all you have?

I'ld like to merge this for one of the next betas, so the code gets enough testing before we release 1.1.

rciorba:
Sorry for the delay. Been a bit more busy than usual lately.

Basically I got stuck at some point trying to understand why an expected
performance drop was not happening.

Let me make sure I've pushed the benchmark tests, and then I'll jump on IRC
and maybe you can help me get to the bottom of it.

src/borg/archive.py
```diff
@@ -931,7 +931,7 @@ def init_chunks(self):
         """
         # Explicitly set the initial hash table capacity to avoid performance issues
         # due to hash table "resonance"
-        capacity = int(len(self.repository) * 1.052 + 1)  # > len * 1.0 / HASH_MAX_LOAD (see _hashindex.c)
+        capacity = int(len(self.repository) * (1/0.99) + 1)  # > len * 1.0 / HASH_MAX_LOAD (see _hashindex.c)
```
ThomasWaldmann: good idea, like this it is more clear.

RonnyPfannschmidt: perhaps it should go for getting the constant imported from hashindex always?

src/borg/testsuite/benchmark.py
```diff
@@ -97,3 +99,33 @@ def test_check(benchmark, cmd, archive):
...
+    benchmark.pedantic(do_inserts, rounds=3, setup=setup)
+
+
+def test_chunk_indexer_getitem(benchmark):
+    max_key = 2**17
+    index = ChunkIndex(max_key)
+    key = 0
+    for key in range(max_key):
+        # we want 32 byte keys, since that's what we use day to day
+        key = sha256(bytes(key)).digest()
```
ThomasWaldmann: bytes(n) is n zero bytes - maybe not what you wanted.

there is a H(n) helper function somewhere, use that.

src/borg/testsuite/benchmark.py
```diff
@@ -97,3 +99,33 @@ def test_check(benchmark, cmd, archive):
...
     result, out = benchmark(cmd, 'help')
     assert result == 0
+
+
+def test_chunk_indexer_setitem(benchmark):
+    max_key = 2**17
+    def setup():
+        # return *args, **kwargs for the benchmarked function
+        return [ChunkIndex(max_key), ], dict()
+    key = 0
```
ThomasWaldmann: looks unused

src/borg/testsuite/benchmark.py
```diff
@@ -97,3 +99,33 @@ def test_check(benchmark, cmd, archive):
...
+    def do_inserts(index):
+        for key in range(max_key):
+            # we want 32 byte keys, since that's what we use day to day
+            key = sha256(bytes(key)).digest()
+            index[key] = (0, 0, 0)
+    benchmark.pedantic(do_inserts, rounds=3, setup=setup)
```
ThomasWaldmann: maybe you don't see a big difference because sha256 is slow?

src/borg/testsuite/benchmark.py
```diff
@@ -97,3 +99,33 @@ def test_check(benchmark, cmd, archive):
...
+
+
+def test_chunk_indexer_getitem(benchmark):
+    max_key = 2**17
+    index = ChunkIndex(max_key)
+    key = 0
```
ThomasWaldmann: unused

src/borg/testsuite/benchmark.py
rciorba: indeed, i hadn't considered that

src/borg/testsuite/benchmark.py
ThomasWaldmann: from borg.testsuite.hashindex import H

src/borg/archive.py
ThomasWaldmann: sure. i tried that long ago, but at least it did not work directly. but maybe i tried the wrong way...

src/borg/testsuite/benchmark.py
enkore: I think a list of keys should be populated outside of the benchmark loop (`sha256(n)` gives us live-like key distribution, while `H(n)` does not) and then that should be used for the test.

src/borg/testsuite/benchmark.py
enkore: H is the wrong tool for pef evaluation because it's output distribution is unlike the distribution used in practice (pseudo-random vs. iirc simple number formatting for H()).

src/borg/testsuite/benchmark.py
ThomasWaldmann: @enkore that sounds best, yes.

for even more life-like data, one can read the stuff from a real-life chunks index file.

src/borg/archive.py
enkore: iirc that's a C constant, so not available immediately from Cython or Python

src/borg/testsuite/benchmark.py
rciorba: Indeed, not what i wanted. This seems to be the biggest slowdown. That's probably why I couldn't see a difference.

src/borg/archive.py
ThomasWaldmann: it's even worse: it is a symbol just known to C preprocessor.
but I guess we could fetch it in hashindex.pyx and make it available to python.

this is unrelated to this changeset though.

ThomasWaldmann:
See #1580 about making the max load constant available.

src/borg/testsuite/benchmark.py
rciorba: @enkore thanks for the suggestion! Indeed i stayed away from H because of the non-uniform distribution. This benchmark requires collisions.

ThomasWaldmann: Hmm? if you want collisions, a uniform dist is not the easiest way to get that.

rciorba: Ok, I misspoke. I don't want an artificial worst case number of collisions, i want a distribution similar to the one used in real life usage. I thinks the sha256 is good for that.

rciorba:
Benchmark results are here:
https://gist.githubusercontent.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/8e7e62cdb3a5b20e38611c43f5254a157fdb3b4e/borg-robinhood-bench.md

I settled on 0.93 fill rate since it's about the same performance as master.

ThomasWaldmann:
From that benchmark data, 0.93 seems to be a good value.

You tested with 2^17 values (which is not much when compared to borg practical usage).

Can you run again with 2^23? (I just had a look into my not too large own backup and it has approx. that amount of unique chunks)

rciorba:
Sure.

ThomasWaldmann:
Another good test (I can do that) is to merge 2 existing chunk indexes. This is a heavy hashtable operation that is done when the chunks cache is out of sync and needs to be rebuild from single-archive cached chunk indexes, so performance there is also interesting.

rciorba:
https://gist.githubusercontent.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/ac741960f5332bce46f6abdd0ea982b940c7800c/borg-robinhood-bench.md

Updated for 2^23 keys. A performance drop for set operations can now be seen in comparison with master.

ThomasWaldmann:
Hmm, ok, so, if we do not want worse performance (we do "set" rather often), can we go a bit lower with the load factor?

Of course it is somehow expected that set is more expensive if it does a balancing operation in robin-hood mode, but somehow that should pay out for get performance. So, if we go a bit lower, to e.g. 80% or 85% load factor, does "get" get faster?

If not, we can still use it to get the higher load factor of 93%.

enkore:
Do I misread the chart or does robin hood regress in all of theses tests from master branch performance?

ThomasWaldmann:
Performance is worse, load factor is better. That's why I asked, see above.

ThomasWaldmann:
Also, we may or may not have tested the local performance breakdown issues. As these depend on input data likely together with the collision algorithm and table size, it is hard to test.

ThomasWaldmann:
I did some more benchmarks, using real-life data from existing archives:

https://paste.thinkmo.de/4QsjXxhn#hash_bench.txt

So, speed is definitely an issue still, for all operations, and the "local breakdowns" might be there also, see line 23.

@rciorba can you check whether the algorithm is right or whether something is missing / going wrong?

rciorba:
I'll rerun the benchmarks and update the results later tonight when I get back to my computer but the gist of it is at .86 this branch seems to perform worse than master at .86
In case that is indeed so this pr is obviously not worth merging. I will start to look into potential optimizations. For example for `hashindex_lookup`, if we know the max offset of the hashindex, we can optimize on missing key. That would improve both set and get.

src/borg/_hashindex.c
```diff
@@ -376,29 +379,69 @@ hashindex_get(HashIndex *index, const void *key)
...
+    uint8_t *bucket_ptr;
+    int offset = 0;
+    int other_offset;
+    int entry_size = (index->key_size + index->value_size);
+    static void *entry_to_insert = NULL;
+    static void *tmp_entry = NULL;
+    if (entry_to_insert == NULL) {
+        entry_to_insert = malloc(entry_size * 2);
+        tmp_entry = entry_to_insert + entry_size;
+    }
```
ThomasWaldmann: hmm, is this code correct if used for different kinds of HashIndexes (different entry sizes)?
if it is called first needing entry_size X, it will allocate a 2X sized buffer, but if it is afterwards called needing entry_size Y, it will reuse that buffer...

Also, it only has one buffer, so multithreaded usage of this might lead to problems also (maybe not currently, but in future?).

ThomasWaldmann:
Well, until we found the real issue maybe do not do optimizations which we also do not have in 1.0-maint (and which are not robin-hood specific), they might just hide a bigger problem.

src/borg/_hashindex.c
```diff
@@ -376,29 +379,69 @@ hashindex_get(HashIndex *index, const void *key)
...
-        memcpy(ptr + index->key_size, value, index->value_size);
+        memcpy(bucket_ptr, entry_to_insert, entry_size);
         index->num_entries += 1;
     }
     else
     {
+        /* we already have the key in the index we just need to update its value */
         memcpy(BUCKET_ADDR(index, idx) + index->key_size, value, index->value_size);
     }
     return 1;
```
ThomasWaldmann: i read through the rest of this function and it looks correct.
have to check against the robin hood paper again, whether it is complete.

rciorba: As I mentioned previously `hashindex_lookup` could take advantage of the
maximum offset for the hashindex, by giving up after looking at `maximum
offset` buckets. Since every `hashindex_set` does a lookup first this could
potentially be a great improvement for real world usage. However to track
maximum offset we will add some small overhead.

Deleting would be the tricky bit. If we are deleting a key with maximum
offset we don't know if there is another key with the same offset, or what
the next biggest offset is. But if we simply don't update maximum offset on
delete (basically maximum offset means maximum ever seen), we could still
reap a great benefit from it, and only re-compute it on hash resize.

The only requirement for doing this, as I said in a previous comment, would
be that this requires the hashindex be in robin_hood order already. One way
to achieve this is to re-insert every key when we read from disk. That way
we can keep things mutually compatible with old versions of borg. Or we can
introduce some sort of versioning scheme using the MAGIC string. Anyway, I
should have some time to implement it tonight or tomorrow and we can see if
it's worth further consideration.

src/borg/_hashindex.c
rciorba: Good points about the different entry sizes! Hadn't considered that. 
I'll drop the static in favor of pointers on the struct for thread safety.

ThomasWaldmann:
Hmm, maybe my test is somehow sub optimal: it uses an existing hashindex, iterates over it, feeds the data into the new hashtable - thus the order of the data might be biased.

enkore:
Q

The main problem with the open addressing is tombstoning. There is a sequence of operations that will fill the entire table with either occupied or deleted entries, making every lookup O(n) because the table is scanned from the modulus index to end.

To mitigate it, we define the load factors and resize if entries are deleted/added and the number of items get out of the load factor bound.

But this is an obvious bad proxy for the tombstones, which have nothing to do with the number of items in the table. It still usually works because the load factors are chosen such that there is a sufficient probability of >=1 empty bucket every now and then, under usual conditions.

So why don't we just track (num_items + num_tombstones) and resize according to that? We don't store it, but a simple sequential scan when loading a table gives us the number, and from that we can just track it in memory.

Shouldn't this largely avoid the tombstoning issue?

ThomasWaldmann:
@enkore sounds good. But when looking at my benchmarks: i do not delete anything, so it can't be tombstones.

enkore:
I'm a bit confused here. What does these numbers actually mean? What are the values in parens, and what are the values not in parens? https://paste.thinkmo.de/4QsjXxhn#hash_bench.txt

It would seem that the values in parens are normalized for the first row and then the other rows are relative to that (except stddev, IQR?) [which doesn't seem to make much sense here since every row is a different test]. And the values not in parens are milliseconds per iteration?

If so, then these numbers are even more confusing. Why does RH get so much slower with a load factor of 0.85, for example.

ThomasWaldmann:
No idea about the stuff in parens, might be something like std deviation?
The stuff without parens is the measured time, unfortunately in different unit, see header.

src/borg/_hashindex.c
ThomasWaldmann: The articles' author seems unimpressed by Robin Hood in his first article, but he was able to improve it:
http://codecapsule.com/2013/11/17/robin-hood-hashing-backward-shift-deletion/

http://www.sebastiansylvan.com/post/robin-hood-hashing-should-be-your-default-hash-table-implementation/ has the same trick as in the other article.

rciorba:
On the bench-marking part, I'm not sure the benchmark I initially used is representative of real world usage. The miss ratio on the get operation is 0%, as in my initial test only asked for keys that had been inserted. What would a realistic miss rate be?

Also, I've been playing with some enhancement ideas. Mainly to get `hashindex_lookup` to provide a hint about how many buckets can be skipped when doing a `set`, since the `set` logic is "find the key if it already exists; if it does not find a bucket to insert". I implemented this on the [master-bench](https://github.com/rciorba/borg/commit/6874edfee8fda8ead504237409b1c35fbe43e2dd) branch as well to be able to compare apples to apples.

ThomasWaldmann:
Miss rate is hard to say. If you always backup mostly the same files, most chunks will be already in the index - BUT, as we also look at the files cache and skip files that do not look modified (mtime, inode, size), they will not be chunked and looked up in hashtable anyway and it will somehow only process files that do have changes.

ThomasWaldmann:
We'll soon release 1.1.0b3 - after that we'll work on rc1, so the window to get this in will close soon.

Can we still get it in or shift to 1.2?

rciorba:
I need to re-run the benchmarks, but as far as i recall it looks like it's
not worth merging at the moment. This depends heavily on the hashindex miss
ratio of course, as RH shines when in that case. Playing with the
benchmarks I got good numbers when there was a 30% hashindex lookup for
missing keys, but according to the feedback i got here that's probably not
a realistic value.


I'll cherry pick this enhancement to a separate PR as we should expect an
immediate improvement to hashindex_set:
https://github.com/rciorba/borg/commit/6874edfee8fda8ead504237409b1c35fbe43e2dd

enkore:
Theory suggests that RH should somewhat alleviate #536 / #1729 ... I'll try to make a setup to benchmark this.

ThomasWaldmann:
The extracted smaller fix for existing code was merged and this PR has conflicts, a rebase onto current master is needed.

rciorba:
Hi TW,

Yes, this PR still has some issues that would prevent merging, and the main
one is performance :)
If that's sorted out, I'd be happy to fix the rest of them.

Sadly, my benchmarks don't show improved performance, but quite the
opposite. Only when adding a 33% miss rate to the benchmark for
hashindex_get (which according to discussions on this thread is improbably
high) do i get back to the same ballpark.

Another issue is about the quality of my measurements. I'm not sure how
reliable the benchmarks have been so far since re-running them shows
serious variations in numbers, as you can see here:
https://gist.github.com/rciorba/e3eded290e91a361f74eefdbc84416bb#file-rh-bench-2-md

If anyone has any ideas about how to correctly measure this, I'd love to
hear them.

ThomasWaldmann:
@rciorba can we make it significantly faster by breaking on-disk file compatibility?

enkore:
Maybe there are ways to detect performance breakdowns and deal with them somehow differently? My expectation is that it happens when the hashtable gets clogged by entries and tombstones, so lookups tend to become large linear scans. What if we just tracked some (specially chosen) statistics about scan lengths and triggered a HT rebuilt if one of these tripped a threshold?

src/borg/_hashindex.c
```diff
@@ -110,18 +110,35 @@ hashindex_index(HashIndex *index, const void *key)
...
+    for(int offset = 0;; offset++) {
+        if (skip_hint != NULL) {
+            (*skip_hint) = offset;
+        }
         if(BUCKET_IS_EMPTY(index, idx))
         {
             return -1;
         }
+        if(offset > distance(num_buckets, idx, start)) {
+            return -1;
```
rciorba: @tw Will add. Basically this was an attempt to optimize lookup by taking advantage of the knowledge that the index is in RH order. So if the element you're looking for has larger offset-from-ideal-location than the bucket you're scanning now, you can give up.

This is the breaking change I mentioned.

rciorba:
For some reason my email replies aren't showing up here so adding them again:

@ThomasWaldmann  Already implemented (thus breaking backwords compatibility and deletion). The latest measurements I posted include it.

Maybe we can sync up sometime on some voice chat (google hangouts perhaps?) and we can brainstorm about this. Don't really have a clear next step in improving this, and I'm starting to have doubts about how realistic the benchmark is.

In the meanwhile, after failing to convince my i7 skylake to disable frequency scaling, I'm setting up my raspberry to re-run benchmarks there to hopefully get less variation between runs.


@enkore Well, actually what is implemented on master currently mitigates the issue of tombstone clogging: each hashindex_get, when it encounters a tombstone before your desired key, will move the key over the tombstone. So I don't think clogging is a big issue.

I agree with you on tracking more stats. That would give everyone a better idea of what to optimize for, and allow for some more realistic benchmarks.

ThomasWaldmann:
I remember reading some paper / blog post about RH hashing and that the evaluator was at first disappointed because he didn't get the expected performance, but then tweaked it and it got much better. Did you also find that paper / blog post already?

rciorba:
@ThomasWaldmann i think you're thinking of http://codecapsule.com/2013/11/17/robin-hood-hashing-backward-shift-deletion/

I didn't read it very carefully because the benchmarks ignore deletion entirely. I should add some deletions to the benchmarks, implement backward-shift-deletion and measure.

ThomasWaldmann:
@rciorba yes, that one was it.

rciorba:
@ThomasWaldmann Already implemented (thus breaking backwords compatibility
and deletion). The latest measurements I posted include it.

Maybe we can sync up sometime on some voice chat (google hangouts perhaps?)
and we can brainstorm about this. Don't really have a clear next step in
improving this, and I'm starting to have doubts about how realistic the
benchmark is.

In the meanwhile, after failing to convince my i7 skylake to disable
frequency scaling, I'm setting up my raspberry to re-run benchmarks there
to hopefully get less variation between runs.

rciorba:
Well, actually what is implemented on master currently mitigates the issue
of tombstone clogging: each hashindex_get, when it encounters a tombstone
before your desired key, will move the key over the tombstone. So I don't
think clogging is a big issue.

I agree with you on tracking more stats. That would give everyone a better
idea of what to optimize for, and allow for some more realistic benchmarks.

rciorba:
Hi, 

Some good news, I've had some time over the holidays and managed to track down a bug in my implementation of hashindex_set that caused the RH swapping scheme to never trigger. This was the cause of the very poor performance.

The gist of it, at 93% fill rate, using a slightly older master as baseline, current numbers are:
 * item lookup is 45% of the baseline (faster)
 * inserting is 152% of baseline (slower)
 * updating is 121% of baseline

There is still room for some speedups especially since the delete toombstones should get removed, and that would mean no more checking for toombstones on every lookup.

enkore:
That's great news! :)

> There is still room for some speedups especially since the delete toombstones should get removed, and that would mean no more checking for toombstones on every lookup.

The tombstones from the old hashtable implementation would get flushed out with the first resize. I think we could add a flag to the cache config and the repository hints to signal this situation, so more efficient implementations could be used when we know that there can't be any tombstones.

rciorba:
@enkore Indeed a resize should be is all it takes to go from an older format to a new one. The issue for robin-hood is that it's not just about not having tombstones, but it also needs the buckets to be ordered correctly. The code takes advantage of that order in several places, so if we simply load a hashindex with no tombstones and assume it to be compatible with this code, hashindex_lookup could erroneously claim certain keys aren't in the index.

So yes, probably all this can be avoided by resizing the index at load time if determined to have been created with an older version of borg.

src/borg/testsuite/benchmark.py
```diff
@@ -148,7 +148,7 @@ def do_gets(keys=keys):
 
 def test_chunk_indexer_getitem_c(benchmark):
     import itertools
-    max_key = 2**20
+    max_key = int(1287917 * 0.93 - 10)
```
ThomasWaldmann: this is only "matching" the robin hood hashtable implemention, to compare with the current one, we would need the other load factor here. maybe use load_factor = 0.93 here or even import it from hashindex.pyx.

src/borg/_hashindex.c
```diff
@@ -119,6 +119,13 @@ distance(int num_buckets, int current_idx, int ideal_idx)
 }
 
 
+/* static long unsigned lookups = 0; */
+/* static long unsigned collisions = 0; */
+/* static long unsigned swaps = 0; */
+/* static long unsigned updates = 0; */
+/* static long unsigned inserts = 0; */
+/* static long unsigned shortcuts = 0; */
```
ThomasWaldmann: use `#ifdef DEBUG` or so?
(also at various places below here)

src/borg/_hashindex.c
```diff
@@ -422,8 +434,10 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
     if(idx < 0)
     {
         /* we don't have the key in the index we need to find an appropriate address */
+      /* inserts ++;  //TRACE_PERF */
```
ThomasWaldmann: indentation issue here? tab / space mix? maybe check that globally and also your editor settings.

src/borg/_hashindex.c
```diff
@@ -511,16 +529,56 @@ static void
...
+  /*        shortcuts, */
+  /*        inserts, */
+  /*        index->num_buckets */
+  /*        ); */
+}
+
+static void
+benchmark_setitem(HashIndex *index, char *keys, int key_count)
+{
+  char *key = keys;
```
ThomasWaldmann: indent by 4 spaces, as usual? check globally please.

src/borg/archive.py
```diff
@@ -1010,7 +1010,7 @@ def init_chunks(self):
         """
         # Explicitly set the initial hash table capacity to avoid performance issues
         # due to hash table "resonance"
-        capacity = int(len(self.repository) * (1/0.93) + 1)  # > len * 1.0 / HASH_MAX_LOAD (see _hashindex.c)
+        capacity = int(len(self.repository) * (1/0.99) + 1)  # > len * 1.0 / HASH_MAX_LOAD (see _hashindex.c)
```
ThomasWaldmann: referring to my comment in previous changeset: iirc we made the load factor a global name in hashindex.pyx in master branch, you just need to rebase upon that at some time, no need to re-implement this.

src/borg/hashindex.pyx
```diff
@@ -368,3 +369,6 @@ from cpython.bytes cimport PyBytes_AS_STRING
 
 def bench_getitem(ChunkIndex chunk_index, bytes keys, int key_count):
     benchmark_getitem(chunk_index.index, PyBytes_AS_STRING(keys), key_count)
+
+def bench_setitem(ChunkIndex chunk_index, bytes keys, int key_count):
```
ThomasWaldmann: pep8 (2 empty lines for toplevel defs)

src/borg/testsuite/benchmark.py
```diff
@@ -125,31 +126,31 @@ def test_chunk_indexer_getitem(benchmark):
...
 
     benchmark.pedantic(do_gets, rounds=200)
 
 
 def test_chunk_indexer_getitem_c(benchmark):
-    max_key = int(445649 * 0.93 - 10)
+    max_key = int(445649 * 0.95 - 10)
     index = ChunkIndex(max_key)
-    keys = [sha256(H(k)).digest() for k in range(max_key)]
+    # keys = [b'00'+sha256(H(k)).digest()[2:-2]+b'00'
```
ThomasWaldmann: pep8

src/borg/testsuite/benchmark.py
```diff
@@ -125,31 +126,31 @@ def test_chunk_indexer_getitem(benchmark):
...
+    #  for k in range(max_key)]
+    keys = [sha256(H(k)).digest()
+     for k in range(max_key)]
```
ThomasWaldmann: pep8 / just put it on 1 line?

src/borg/_hashindex.c
```diff
@@ -27,6 +27,17 @@
...
 
+#define DEBUG 0
+
+#define debug_print(fmt, ...)                   \
+  do {                                          \
+    if (DEBUG) {                                \
+      fprintf(stderr, fmt, __VA_ARGS__);        \
+      fflush(NULL);                             \
+    }                                           \
+  } while (0)
```
ThomasWaldmann: hmm, the do..while is so you just get a block?

wouldn't using {..} also work for that purpose?

src/borg/_hashindex.c
```diff
@@ -134,20 +145,21 @@ hashindex_lookup(HashIndex *index, const void *key, int *skip_hint)
     int idx = start;
     int num_buckets = index->num_buckets;
     int offset;
-    for(offset=0; ; offset++) {
-      /* lookups ++;  //TRACE_PERF */
+    for(offset=0; ;offset++) {
+        if (DEBUG) {
+            lookups ++;
+        }
```
ThomasWaldmann: hmm, you seem to rely on the compiler optimizing away this debug code. one could also use #ifdef DEBUG to let the precompiler just remove all the code.

src/borg/_hashindex.c
```diff
@@ -447,19 +461,25 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
...
             if(other_offset < offset) {
                 /* Swap the bucket at idx with the current entry_to_insert.
                    This is the gist of robin-hood hashing, we rob from the key with the
                    lower distance to its optimal address by swapping places with it. */
                 memswap(bucket_ptr, entry_to_insert, tmp_entry, entry_size);
                 offset = other_offset;
-              /* swaps++;  //TRACE_PERF */
+                if (DEBUG) {
+                    swaps++;
+                }
```
ThomasWaldmann: the whole "if debug increment x" could be also INC(x) macro which expands via precompiler to code or nothing, depending on DEBUG value. would save some code lines maybe.

src/borg/_hashindex.c
```diff
@@ -436,14 +447,18 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
...
+    uint8_t tmp_entry[entry_size];
     int num_buckets = index->num_buckets;
-    if (entry_to_insert == NULL) {
-        entry_to_insert = malloc(entry_size * 2);
-        tmp_entry = entry_to_insert + entry_size;
-    }
+    /* if (entry_to_insert == NULL) { */
+    /*     entry_to_insert = malloc(entry_size * 2); */
+    /*     tmp_entry = entry_to_insert + entry_size; */
+    /* } */
```
ThomasWaldmann: guess there should be some rebase -i at the end, collapsing some comment-back-and-forth changes (and have them replaced by precompiler conditionals in the end).

ThomasWaldmann:
https://github.com/borgbackup/borg/pull/1429#issuecomment-269771104 so you compared there:
- older master with 75% LF and
- robin hood PR with 93% LF
?

If so, I guess the slower inserting is not only due to RH, but also caused by the higher LF (and if one wanted a more fair comparison, one would need to adapt LF to be equal).

Hmm, why is updating slower? Isn't that just like a get + setting new values?

src/borg/_hashindex.c
```diff
@@ -108,24 +123,56 @@ hashindex_index(HashIndex *index, const void *key)
...
+            lookups ++;
+        }
+        if (skip_hint != NULL) {
+            (*skip_hint) = offset;
+        }
         if(BUCKET_IS_EMPTY(index, idx))
         {
             return -1;
         }
+        if(offset > distance(num_buckets, idx, hashindex_index(index, BUCKET_ADDR(index, idx)))) {
```
ThomasWaldmann: considering there was a bug here, add a comment that explains this condition?

src/borg/_hashindex.c
```diff
@@ -369,45 +417,90 @@ hashindex_write(HashIndex *index, const char *path)
...
         return NULL;
     }
     return BUCKET_ADDR(index, idx) + index->key_size;
 }
 
+inline void
+memswap(void *a, void *b, void *tmp, size_t entry_size) {
+    memcpy(tmp, a, entry_size);
+    memcpy(a, b, entry_size);
+    memcpy(b, tmp, entry_size);
```
ThomasWaldmann: https://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html

enkore: hashindex and chunker already use memcpy (extensively) though

rciorba:
@ThomasWaldmann https://github.com/borgbackup/borg/pull/1429#issuecomment-269863900
> If so, I guess the slower inserting is not only due to RH, but also caused by the higher LF (and if one wanted a more fair comparison, one would need to adapt LF to be equal).

Both master and robin_hood were compared at 93% fill rate, so the slowness is probably due to all the swapping. There may be room to speed things up.

> Hmm, why is updating slower? Isn't that just like a get + setting new values?

The getitem benchmark included a 33% miss (lookup item not in cache). I'll add a separate benchmark with no misses so we can understand the impact better. 

The gist of it is I need to experiment and measure, but at least now I'm not stuck on non reliable benchmarks and some improvement is visible.

src/borg/_hashindex.c
ThomasWaldmann: it's not a problem as long as we build on wheezy (or on anything else with < 2.14).

ThomasWaldmann:
about timing again, 1.1.0b3 is pending (only 1 todo left), so new hashing likely won't be ready until b3.
but as the main obstacle seems solved now, we could maybe target b4 / rc1.

ThomasWaldmann:
@rciorba nice to see ongoing improvements! :)

if you do a rebase -i, can you reword some commit comments, so it is more clear what's changed? Like:
- WIP
- implement deletion (should be: change tombstone deletion to backshift deletion)

What's the point of not using the stack, but heap?

rciorba:
Some updates.

Implemented delete as well, and added a few benchmarks for that.

Performance was not great in most use cases. RH excels when we have lookups for key that don't exist. In order to speedup other cases I tried small tweaks, like replacing modulo operations with an if statement to wrap-around the index. Since it would have been unfair to only apply those to the RH branch you'll find them backported to [master-bench](https://github.com/rciorba/borg/tree/master-bench) and included in the measurements.

### What is compared:
 * baseline - the current implementation of the hashindex
 * baseline_m - the current implementation but replaces modulo with if statement wherever possible
 * rh - robin hood hashing
 * rh_16 - for getitem only checks the 'distance from ideal address' of the current bucket every 16 buckets
 * rh_8 - same as above, but every 8 buckets

### Rationale for the r_16 and r_8 setups
In order to speedup the lookup for a missing key, we can leverage the knowledge that if the key at the address we're currently inspecting is closer to it's ideal bucket then the offset our searched key, it would have been inserted here. 

This check however seems to be quite expensive (requires we re-compute the ideal address of each bucket as we scan) so I tried to reduce the impact by only running this shortcut check for every N buckets.

### The measurements

https://cdn.rawgit.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/076524eb4a68697deafb7c1217113e65b526ebff/results.html

It's hard for me to judge if merging this would be appropriate since it depends very much on how often lookups for missing keys happen (inserting a new key, getting a key that doesn't exist), but my gut feeling is it's not really worth it especially given the slowdown in updating existing keys (I'm not 100% sure why that is, but my hunch is it's caused by the shortcutting code being ran for every bucket for hashindex_set).

### Other potential speedups

Cache alignment of the current hashindex implementation. This could have a major impact. Currently we have the key(32 bytes)+value(3 bytes) in one contiguous memory location. Most of the operations only look at the key. If the keys were contiguous in memory 2 keys would fit on one cache line on most architectures. I'll give this a try and measure it.

rciorba:
@ThomasWaldmann Sorry, I know the history is a bit of a mess, this whole thing is now very trial and error for me, and I have some doubts about it ever getting merged (simply because the performance isn't there) so my commit messages have gotten sloppy. I will rewrite them.

> What's the point of not using the stack, but heap?

I measured and it's marginally faster.

ThomasWaldmann:
2 ideas:
- do more practical benchmarks, like real backups / chunk re-syncs / archive deletion
- ask author of the (2nd/3rd) RH paper for feedback

ThomasWaldmann:
I just found: http://accidentallyquadratic.tumblr.com/post/153545455987/rust-hash-iteration-reinsertion

ThomasWaldmann:
Just as a vague idea:

It's not like borg is generally slow due to the current hashtable, just sometimes.

So, is it maybe the table resize? That quadratic behaviour described above is due to inserting into the new table in (old) hashtable order. So this affects both the currently used simple hashtable with linear probing as well as robin hood with linear probing.

If that would be true, we maybe would not have a problem at all with the hash table's normal performance being slower - if just the desastrous quadratic case is solved.

Hmm, just noticed: we don't have 2^N table sizes, so maybe the quadratic issue does not apply.

enkore:
I'd say that the actual HT performance is a very minor thing (<1 %), when it works, but as we have observed a couple times there are degenerate cases where >9x % are spent working on the HT. I still think this may be tombstoning issues where lookups turn into long sequential scans.

ThomasWaldmann:
@enkore the backshift deletion does not need tombstones, so would solve that potential issue.

there is one case of a bit more agressive HT use: when borg decides that a file has not changed (mtime/inode/size same), it takes the chunk id list from the files cache **and verifies that they are all still present in the repo** (by looking them up in the hashindex, which is assumed to be correct). So a potentially long list of hashids gets looked up in a burst with no I/O. As for many backups most files don't change, this is the usual case.

rciorba:
@enkore The current HT implementation of hashindex_lookup, when it finds a key, will move said key over the first tombstone it encountered during the scan. This means the impact of tombstones is not that big.
 
A potentially worst case scenario would be deleting a large number of keys would, for a while, make every get also perform a memory swap.

Also, while thinking about an answer to your comment I realized my other PR did something stupid with tombstones. So thanks for your comment and here's a PR to fix that issue: https://github.com/borgbackup/borg/pull/2116

enkore:
> @enkore The current HT implementation of hashindex_lookup, when it finds a key, will move said key over the first tombstone it encountered during the scan. This means the impact of tombstones is not that big.

But doesn't that leave another tombstone in the place where the key was found (since further items with the same key could be located beyond the found item the bucket-chain can't be broken)?

rciorba:
@enkore You're right, that does leave another tombstone, but it's hard to reason about what the impact of them is. The only really bad case I can imagine is having the hash at 75% fill rate, but having the 25% free space be mostly tombstones. And I'm not sure if that's likely to happen, since inserting keys would replace the tombstones with real values.

My reasoning about this:
 * say we're at 75% fill, no tombstones, 
 * delete half the keys 37.5 % fill, 37.5% tombstones (performance is no worse than at 75% fill rate, after the swaps have settled down)
 * insert back up all the way to 75%, X% tombstones, the inserts will have replaces a bunch of them

How about we add a simple tracking of tombstones on an experimental 'telemetry' branch, then we get some actual numbers (basically find out what X is)? 

Also, an eager delete could be implemented for the current master branch as well. I'll have a crack at it tonight/tomorrow if time allows.

enkore:
Yes, my reasoning so far always has been in the direction of tombstones creating very long bucket chains that make look-ups slow, since that would correlate well with the "worst case" issues observed.

> How about we add a simple tracking of tombstones on an experimental 'telemetry' branch, then we get some actual numbers (basically find out what X is)?

Sounds like a good idea. See also https://github.com/borgbackup/borg/pull/1429#issuecomment-246196015 and https://github.com/borgbackup/borg/pull/1429#issuecomment-256139026

ThomasWaldmann:
a rebase would be useful so this can be practically tested.

pczarn:
Having tombstones in hash maps with robin hood hashing makes even less sense than in other hash maps.

I suggest a load factor between 80% and 90%. The difference between 90% and 93% may seem small, but having 7% free space instead of 10% is a big deal. A similar effect slows down hard drives. For a discussion about decreasing load factor, see https://github.com/rust-lang/rust/issues/38003

You may want to remove the use of `distance` from `hashindex_lookup`, since that conditional break optimizes cases where the lookup fails, and you expect a hit rate of 100% in your application. Keep in mind that `hashindex_lookup` is used in `hashindex_set`. (In Rust's hashmap, we don't restart the search like in this PR. Whenever a new element is inserted, we continue at the offset from the failed initial lookup. However, I think the difference is tiny.)

Are keys influenced by foreign input? I'm not sure what is the purpose of this implementation.

src/borg/_hashindex.c
```diff
@@ -369,49 +446,129 @@ hashindex_write(HashIndex *index, const char *path)
...
+        memcpy(entry_to_insert, key, index->key_size);
+        memcpy(entry_to_insert + index->key_size, value, index->value_size);
+        bucket_ptr = BUCKET_ADDR(index, idx);
+        if (DEBUG) {
+            collisions += offset;
         }
-        idx = hashindex_index(index, key);
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
-            idx = (idx + 1) % index->num_buckets;
+            /* we have a collision */
```
pczarn: This loop skips over some number of entries, then continues by displacing other entries. Our inserted entry will take the bucket of the first displaced entry. Instead, you can forward-shift all buckets from the first displaced entry until the end of the chunk, using `memmove`. By increasing the displacement of all these buckets by 1, you keep the invariant of robin hood hashing, which relies on comparing displacements, not on their absolute value (`other_offset < offset`).

Tell me if something in my explanation is unclear. I'm using different names. My word for "distance" is "displacement".

rciorba:
Thanks for the feedback! 

> Having tombstones in hash maps with robin hood hashing makes even less sense than in other hash maps.

There are no tombstones in this implementation. This PR also became a place to discuss the problems of the current implementation, hence the discussion about them.

> 
> I suggest a load factor between 80% and 90%. The difference between 90% and 93% may seem small, but having 7% free space instead of 10% is a big deal. A similar effect slows down hard drives. For a discussion about decreasing load factor, see rust-lang/rust#38003

Thanks. I'll check those out. I did benchmark at different load levels, so the performance impact should be visible there as well.

https://cdn.rawgit.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/076524eb4a68697deafb7c1217113e65b526ebff/results.html

> 
> You may want to remove the use of distance from hashindex_lookup, since that conditional break optimizes cases where the lookup fails, and you expect a hit rate of 100% in your application. 

I'll benchmark without the distance as well, but the periodic run of `distance` (every 16 or every 8 buckets) seems to have a small enough impact to be worth it. (rh8 and rh16 in the linked graphs). 

>Keep in mind that hashindex_lookup is used in hashindex_set. (In Rust's hashmap, we don't restart the search like in this PR. Whenever a new element is inserted, we continue at the offset from the failed initial lookup. However, I think the difference is tiny.)

This PR adds a pointer param so we can 'return' not just the found index but also the address to start the search in case of a miss, so in that respect it should behave the same as Rust's hashmap.

> 
> Are keys influenced by foreign input? 

I don't understand your question, could you please elaborate?

> I'm not sure what is the purpose of this implementation.

I believe the end goal was to use less memory by allowing larger fill rates. Implementing RH was mentioned at a EuroPython sprint for volunteers to pick up, and it sounded interesting, but this is my first contact with their code base. Since I don't know the project that well, I'm not sure what the usage patterns of the hashmap are. I was hoping the maintainers of borg could help guide me with that, hence the multiple variations of the implementation being bench-marked to see what the tradeoffs are for different use cases (like the distance short-cutting in lookup and it's impact on getitem whit no misses, vs getitem with misses).

enkore:
> Are keys influenced by foreign input? I'm not sure what is the purpose of this implementation.

Keys are the first few bits of a (usually keyed) cryptographic hash of input data. In the unkeyed case keys can thus be influenced to some extent; in the keyed case (used whenever the data is stored encrypted) this isn't possible at all.

src/borg/_hashindex.c
rciorba: So basically take the block (the entire contiguous section of buckets up to the first empty bucket) and `memmove` it forward one address, then insert at the ideal location. Oh my, yes, such an elegant solution! Thanks!

> not on their absolute value (other_offset < offset).

Not sure what you mean by absolute value. The offsets are the relative distance from the ideal bucket a key would be in vs which bucket it is now, so I think it's the same as what you call "displacement".

> Tell me if something in my explanation is unclear. I'm using different names. My word for "distance" is "displacement".

Indeed, displacement is a better name for it.


Thanks for the feedback!

rciorba:
@enkore That makes sense. Thank you!

rciorba:
Closing this PR and starting a fresh one since this one has grown to massive proportions and many of the intermediate changesets are no longer relevant to the final form.

PR #1432, opened 2016-08-03
FAQ: Add pull backup stable filesystem names recommendation

robmv:


docs/faq.rst
```diff
@@ -103,6 +103,18 @@ leaving the local machine. We do not mean the transport layer encryption
 by that, but the data/metadata itself. Transport layer encryption (e.g.
 when ssh is used as a transport) applies additionally.
 
+Why are files always transferred when doing a pull backup mounting remote shares?
+---------------------------------------------------------------------------------
+
+Check that the filesystem you are mounting support reading the remote file
+inode numbers and use stable mount points (not random generated temporary
```
ThomasWaldmann: I'ld say split this into 2 FAQ entries:
1) about stable inodes
2) about always using same mountpoint

Both are not necessarily specific to "mounting remote shares".

docs/faq.rst
```diff
@@ -103,6 +103,18 @@ leaving the local machine. We do not mean the transport layer encryption
...
+directories for example). If the directory where you mount the remote host
+filesystem is different for each backup, borgbackup assume it is a different
+file even if it hasn't changed so it will need to check all file segments
+and transfer data over the remote filesystem protocol consumming time and
```
ThomasWaldmann: maybe instead of "check / transfer", talk about reading / chunking / hashing - that's what's happening.

docs/faq.rst
```diff
@@ -103,6 +103,18 @@ leaving the local machine. We do not mean the transport layer encryption
...
+network resources. It will not use more backup disk space since segments
+are there from previous backups.
```
ThomasWaldmann: don't say segments here, that is a term we use in other context. what you mean here are chunks.

(multiple chunks are written into one segment file to reduce the number of files, but that doesn't matter here)

ThomasWaldmann:
also, if you want that in the 1.0.x docs, please file a PR against 1.0-maint.

it will automatically get merged into master from there later.

robmv:
Closing pull request, will create a new one to the 1.0-maint branch

PR #1435, opened 2016-08-03
fix unintended file cache eviction, fixes #1430

### LINKED ISSUE #1430: unintended file cache eviction
When creating archives, the newest file will not be registered in the file cache in order to avoid data loss when working with snapshots (see FAQ). At the end of the backup borg simply removes everything from the file cache with with an mtime the same or newer than the newest mtime seen during the current run. This can result in unintended file cache eviction in the following situations:

1) when interrupting the backup process

2) when a mount point is unavailable during the backup

3) when creating multiple archives from different sources in one repository

In the latter situation the backup is still okay, but a lot of files will be re-hashed over and over again, resulting in long backup times.

The question is... do we really have to forget the newest file when working with nanosecond timestamps? And if so, would it be possible to only evict this single cache entry instead of everything with a newer mtime?

ThomasWaldmann:
thanks much to e477 for diagnosing this and finding the right fix.

### DIFF
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -196,10 +196,13 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
             ttl = int(os.environ.get('BORG_FILES_CACHE_TTL', 20))
             with open(os.path.join(self.path, 'files'), 'wb') as fd:
                 for path_hash, item in self.files.items():
-                    # Discard cached files with the newest mtime to avoid
-                    # issues with filesystem snapshots and mtime precision
+                    # Only keep files seen in this backup that are older than newest mtime seen in this backup -
+                    # this is to avoid issues with filesystem snapshots and mtime granularity.
+                    # Also keep files from older backups that have not reached BORG_FILES_CACHE_TTL yet.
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
merge?

borg/cache.py
```diff
@@ -196,10 +196,12 @@ def commit(self):
...
                 for path_hash, item in self.files.items():
-                    # Discard cached files with the newest mtime to avoid
-                    # issues with filesystem snapshots and mtime precision
+                    # Only keep files seen in this backup that are older than newest mtime seen in this backup -
+                    # this is to avoid issues with filesystem snapshots and mtime granularity.
+                    # Also keep files from older backups that have not reached BORG_FILES_CACHE_TTL yet.
                     item = msgpack.unpackb(item)
-                    if item[0] < ttl and bigint_to_int(item[3]) < self._newest_mtime:
+                    if item[0] == 0 and bigint_to_int(item[3]) < self._newest_mtime or \
+                       item[0] > 0 and item[0] < ttl:
```
enkore: extract `item[0]` as `item_ttl` or so?

PR #1438, opened 2016-08-04
borg debug-dump-repo-objs

ThomasWaldmann:
dump all objects stored in the repository (decrypted and decompressed)

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -681,6 +681,28 @@ class Archiver:
         print('Done.')
         return EXIT_SUCCESS
 
+    @with_repository()
+    def do_debug_dump_repo_objs(self, args, repository, manifest, key):
+        """dump (decrypted, decompressed) repo objects"""
+        marker = None
+        i = 0
+        while True:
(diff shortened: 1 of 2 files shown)


borg/archiver.py
```diff
@@ -681,6 +681,28 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+        while True:
+            result = repository.list(limit=10000, marker=marker)
+            if not result:
+                break
+            marker = result[-1]
+            for id in result:
+                cdata = repository.get(id)
+                give_id = id if id != Manifest.MANIFEST_ID else None
+                data = key.decrypt(give_id, cdata)
+                filename = '%06d_%s.obj' % (i, hexlify(id).decode('ascii'))
```
enkore: `%06d` is maybe a bit short for an entire repository

ThomasWaldmann: yes, noticed that also - but that format is used at other places, too.
nothing bad will happend when it needs more than 6 digits, except being different length.

PR #1447, opened 2016-08-08
Debugging topics and some other minor stuff

enkore:
- --debug-topic sets any logger to DEBUG level, has predefined shortcuts
- Silences some very verbose debug information by default (e.g. which compression for what file)
- Some minor refactors
- borg-extract gains --progress

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -795,7 +795,7 @@ Number of files: {0.stats.nfiles}'''.format(
             # there should be no information in the cache about special files processed in
             # read-special mode, but we better play safe as this was wrong in the past:
             path_hash = ids = None
-        first_run = not cache.files
+        first_run = not cache.files and cache.do_files
         if first_run:
             logger.debug('Processing files ...')
         chunks = None
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -254,6 +254,7 @@ class Archive:
         self.consider_part_files = consider_part_files
         self.pipeline = DownloadPipeline(self.repository, self.key)
         if create:
+            self.file_compression_logger = create_logger('borg.debug.file-compression')
             self.items_buffer = CacheChunkBuffer(self.cache, self.key, self.stats)
             self.chunker = Chunker(self.key.chunk_seed, *chunker_params)
             self.compression_decider1 = CompressionDecider1(compression or CompressionSpec('none'),
@@ -818,7 +819,7 @@ Number of files: {0.stats.nfiles}'''.format(
             item.chunks = chunks
         else:
             compress = self.compression_decider1.decide(path)
-            logger.debug('%s -> compression %s', path, compress['name'])
+            self.file_compression_logger.debug('%s -> compression %s', path, compress['name'])
             with backup_io():
                 fh = Archive._open_rb(path)
             with os.fdopen(fh, 'rb') as fd:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1368,10 +1368,6 @@ class Archiver:
         tamper-resistant as well, unless the attacker has access to the keys.
 
         It is also very slow.
-
-        --verify-data only verifies data used by the archives specified with --last,
-        --prefix or an explicitly named archive. If none of these are passed,
-        all data in the repository is verified.
         """)
         subparser = subparsers.add_parser('check', parents=[common_parser], add_help=False,
                                           description=self.do_check.__doc__,
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1225,6 +1225,10 @@ class Archiver:
         common_group.add_argument('--debug', dest='log_level',
                                   action='store_const', const='debug', default='warning',
                                   help='enable debug output, work on log level DEBUG')
+        common_group.add_argument('--debug-topic', dest='debug_topics',
+                                  action='append', metavar='TOPIC', default=[],
+                                  help='enable TOPIC debugging (can be specified multiple times). '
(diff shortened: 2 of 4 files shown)


src/borg/helpers.py
```diff
@@ -1527,6 +1527,8 @@ def decide(self, path):
 
 
 class CompressionDecider2:
+    logger = create_logger('borg.debug.file-compression')
```
ThomasWaldmann: hmm, this gets executed at import time - move to `__init__`?

src/borg/archiver.py
```diff
@@ -2202,11 +2205,22 @@ def _setup_implied_logging(self, args):
             if args.get(option, False):
                 logging.getLogger(logger_name).setLevel('INFO')
 
+    def _setup_topic_debugging(self, args):
+        """Turn on DEBUG level logging for specified --debug-topics."""
+        debug_topics = {
+            'file-compression': 'borg.debug.file-compression',
+        }
```
ThomasWaldmann: just require the last part to be given on the cli and remove this mapping?

ThomasWaldmann: hmm, can we check what is a valid debug logger name?

src/borg/archiver.py
```diff
@@ -2202,11 +2205,22 @@ def _setup_implied_logging(self, args):
...
+        for topic in args.debug_topics:
+            topic = debug_topics.get(topic, topic)
```
ThomasWaldmann: check for None?

src/borg/archive.py
```diff
@@ -795,7 +795,7 @@ def process_file(self, path, st, cache, ignore_inode=False):
             # there should be no information in the cache about special files processed in
             # read-special mode, but we better play safe as this was wrong in the past:
             path_hash = ids = None
-        first_run = not cache.files
+        first_run = not cache.files and cache.do_files
```
ThomasWaldmann: well, this silences it by making it non-functional when cache.do_files is False.
can we keep it functional?

src/borg/helpers.py
```diff
@@ -597,8 +597,7 @@ def format_time(t):
 def format_timedelta(td):
     """Format timedelta in a human friendly format
     """
-    # Since td.total_seconds() requires python 2.7
```
ThomasWaldmann: now that is embarrassing. :D

src/borg/archiver.py
```diff
@@ -433,6 +435,14 @@ def item_is_hardlink_master(item):
             return (partial_extract and stat.S_ISREG(item.mode) and
                     item.get('hardlink_master', True) and 'source' not in item)
 
+        if progress:
+            extracted_size = 0
+            for item in archive.iter_items(filter=lambda item: matcher.match(item.path)):
+                extracted_size += item.file_size()
+            pi = ProgressIndicatorPercent(total=extracted_size, msg='Extracting files %5.1f%%', step=0.1, same_line=True)
+        else:
+            pi = None
```
ThomasWaldmann: borg currently only does single passes for creation and extraction. this would be the first place that introduces a 2nd pass just for cosmetics.
I'ld suggest the progress indication be like in "create" and not use a 2nd pass.

ThomasWaldmann:
This PR is awful to review as a whole because it mixes lots of different, unrelated topics.
Can you split this up into PRs with each only having related changes?

src/borg/helpers.py
enkore: Isn't that the whole point of LayzLogger and create_logger?

src/borg/archiver.py
enkore: Since this also doubles as an easy way to control just any logger(1) I'd say no extra check is needed.

(1) `debug_topics.get(topic, topic)` -> not in the mapping, just use it directly.

src/borg/helpers.py
ThomasWaldmann: ah, right.

src/borg/archiver.py
ThomasWaldmann: ok, if you wan to control any logger, couldn't this be just:
```
if '.' not in topic:
    topic = 'borg.debug.' + topic
```
That way we would not have to expand the map with an ever growing amount of known debug loggers.

enkore: Nice idea

src/borg/archive.py
enkore: What"s the point of this message anyhow? If it"s to signal the beginning of the archiving op it should be somewhere in Archiver._process instead.

enkore:
Yes, I didn't consider that. I moved the --progress to #1449 and the refactor to #1450

src/borg/archive.py
ThomasWaldmann: iirc, there was some reason to do it here, but i don't remember it.

enkore:
Merge? Aside from "processing files" showing 0 times instead of nfiles times with --no-files-cache I think everything has been addressed. I don't quite see the relevance of that messages, and I still haven't really found a reason why it has to be emitted there and not at a more convenient place.

src/borg/logger.py
```diff
@@ -149,8 +149,13 @@ def __logger(self):
                 if not configured:
                     raise Exception("tried to call a logger before setup_logging() was called")
                 self.__real_logger = logging.getLogger(self.__name)
+                if self.__name.startswith('borg.debug.') and self.__real_logger.level == logging.NOTSET:
+                    self.__real_logger.setLevel('WARNING')
```
ThomasWaldmann: hmm, how is borg.debug.* special so it needs that treatment?
and other borg.something.else does not need this?

enkore: This is so that --debug does not cause these to be enabled by default.

PR #1449, opened 2016-08-08
extract --progress

enkore:


enkore:
> borg currently only does single passes for creation and extraction. this would be the first place that introduces a 2nd pass just for cosmetics.
> I'ld suggest the progress indication be like in "create" and not use a 2nd pass.

(from #1447)

I tested this with a ~20 GB archive and the difference is (locally) practically not measurable, I suspect this may be different for remote repos though (hence the `(may be slower)`). The disadvantage I see in the create-like output is that it's open-ended, just like --list is now (which is imho very hard to judge if you don't exactly now what's in the archive and in what order). Giving a percentage there would make it pretty much the same as this is as-is.

src/borg/archiver.py
```diff
@@ -433,6 +435,14 @@ def item_is_hardlink_master(item):
             return (partial_extract and stat.S_ISREG(item.mode) and
                     item.get('hardlink_master', True) and 'source' not in item)
 
+        if progress:
+            extracted_size = 0
+            for item in archive.iter_items(filter=lambda item: matcher.match(item.path)):
+                extracted_size += item.file_size()
```
ThomasWaldmann: imagine some huge sparse files being in there, like VM disk images.

the progress estimation will be very wrong as the size will look huge, but in the end it will be just a lot of copies of the same all-zero-chunk and (depending on other flags) might just result in a lot of seeks on the output file.

even without sparse files: if a file has a lot of duplicated chunks internally, we could optimize that and make it much faster than without duplication (not sure if we already do that).

so, I am still rather against progress indication that outputs information that can be very wrong and would prefer outputting only information that is correct for sure.

enkore: I somewhat see the point you're trying to make.

> sparse files

We don't extract those by default (--sparse). If that's specified then I think it can be expected that apparent write speed will vary.

> also, it only considers extracted size - what if the bottleneck is the repo connection and thus the compressed size is what counts?

Hm? It still counts correctly. The speed will vary with repo speed and compression ratio (if it varies a lot), which is kinda clear I think.

> so, I am still rather against progress indication that outputs information that can be very wrong and would prefer outputting only information that is correct for sure.

I don't think this is very wrong, since it says "extracting files" so the percentage relates to how much of the files are extracted. That there are many factors contributing (bandwidths on input/output end, latency on both ends, many small vs few large files, compression ratio if any, varying background/other CPU load etc etc) how fast that is.

ThomasWaldmann: hmm, as it is only telling the percentage of extracted size, yes, that is a known value and thus correct.

the whole thing is still questionable though as people will try to correlate percentage with time ("it took 1h for 10%, so it will be finished in 9h") and the speed will vary widely.

enkore: That's the reason why there's no progress *predictor* here (like windows copy file dialog ;)



Stuff like unstable / varying internet speeds (cable internet.........) aside I don't think it's too big of an issue.

src/borg/item.py
```diff
@@ -157,6 +157,14 @@ class Item(PropDict):
...
     part = PropDict._make_property('part', int)
 
+    def file_size(self):
+        if 'chunks' not in self:
+            return 0
+        total_size = 0
+        for chunk_id, size, csize in self.chunks:
+            total_size += size
+        return total_size
+
```
ThomasWaldmann: return sum(c.size for c in self.chunks)

or so.

src/borg/archiver.py
ThomasWaldmann: ok, let's take this. if complaints come up, we can add a faq item about why past speed is not a predictor of future speed.

ThomasWaldmann: Another one:
```
extracted_size = sum(item.file_size()
                     for item in archive.iter_items(filter=lambda item: matcher.match(item.path)))
```

src/borg/archiver.py
```diff
@@ -433,6 +435,14 @@ def item_is_hardlink_master(item):
             return (partial_extract and stat.S_ISREG(item.mode) and
                     item.get('hardlink_master', True) and 'source' not in item)
 
+        if progress:
+            extracted_size = 0
+            for item in archive.iter_items(filter=lambda item: matcher.match(item.path)):
```
textshell: missing strip_components handling.

src/borg/archiver.py
```diff
@@ -433,6 +435,13 @@ def item_is_hardlink_master(item):
             return (partial_extract and stat.S_ISREG(item.mode) and
                     item.get('hardlink_master', True) and 'source' not in item)
 
+        if progress:
+            extracted_size = sum(item.file_size() for item in
+                                 archive.iter_items(filter=lambda item: matcher.match(item.path)))
```
textshell: github hid this, so i’ll try again:
This code is missing strip_components handling, because borg doesn‘t extract files that are cut away completly by stripping components.
I think we should do all filtering in the filter= callable. See #1418 for another reason

enkore: Yes, I think that'd be better.

enkore:
merge?

textshell:
some random thoughts that certainly should not block this PR.
* Maybe we should have a line output "Calculating restore size" so users where is step takes some time actually know what is taking time without a progress bar.
* currently this is just a percentage with 3 digits of precision, might not be enough for large backups
* The same_line parameter has to be the same in all instances of ProgressIndicatorPercent in a process. That is a bit fragile (but certainly not caused by this PR)

enkore:
> Maybe we should have a line output "Calculating restore size" so users where is step takes some time actually know what is taking time without a progress bar.

Hmm, we have ProgressIndicatorEndless, but I don't think it's a good match here (huge variance of archive size -> some will see four dots, some four thousand). If we want progress display here, maybe something like a spinner.

textshell:
Yes, i was just thinking of a static line that just explains what is done. Most refined would be a indefinite style progress bar (right, left bouncing thingy) but i think that would be overkill.
Basically just `logging.getLogger('borg.output.progress').info("Calculating size")` (text would need to be max 18 or so chars long)

enkore:
Re. `same_line`, this can be removed completely I think. When would it be reasonable to not use it? And like you say, it's global, not an actual parameter...

textshell:
Yes, it seems unused at the moment. And i don‘t remember ever wishing for a program not keeping it‘s progress indicator on the same line, so i think it is not useful. Also people can override logging config, they just need to life with the one line full of blanks.

enkore:
Ditto, there are very few programs that do this (only prominent example I can think of would be Windows chkdsk if it's started during the boot - and that's probably a limitation of the console there, not intentional).

src/borg/archiver.py
```diff
@@ -453,6 +455,14 @@ def item_is_hardlink_master(item):
                     item.get('hardlink_master', True) and 'source' not in item)
 
         filter = self.build_filter(matcher, item_is_hardlink_master, strip_components)
+        if progress:
+            logger = logging.getLogger(ProgressIndicatorPercent.LOGGER)
+            logger.info('Calculating size')
+            extracted_size = sum(item.file_size() for item in archive.iter_items(filter))
+            pi = ProgressIndicatorPercent(total=extracted_size, msg='Extracting files %5.1f%%', step=0.1)
+        else:
```
RonnyPfannschmidt: How about a NullProgressIndicator to avoid all the conditionals

enkore: In principle yes, but in this case I don't think it's a good idea. It saves two `if pi:` instances, but changes the parameter into "well uh sometimes it's needed, sometimes it ain't". I could change that but then it's strange semantically the other way around, requiring a parameter that may/may not be used etc.

ThomasWaldmann:
ok.

PR #1454, opened 2016-08-08
repo: do not put objects that we won't get, fixes #1451

### LINKED ISSUE #1451: Repository: MAX_OBJECT_SIZE is not enforced at write-time
E.g.

```py

>>> with borg.repository.Repository("repo", exclusive=True) as r:
...     r.put(b'0' * 32, bytes(50*1024*1024))
...     r.get(b'0' * 32)
  File "<stdin>", line 3, in <module>
  File "/usr/lib/python3.5/site-packages/borg/repository.py", line 474, in get
    return self.io.read(segment, offset, id_)
  File "/usr/lib/python3.5/site-packages/borg/repository.py", line 695, in read
    size, tag, key, data = self._read(fd, self.put_header_fmt, header, segment, offset, (TAG_PUT, ))
  File "/usr/lib/python3.5/site-packages/borg/repository.py", line 717, in _read
    segment, offset))
borg.helpers.IntegrityError: Invalid segment entry size [segment 1, offset 8]
Invalid segment entry size [segment 1, offset 8]
```

ThomasWaldmann:
we will not get() objects that have a segment entry larger than MAX_OBJECT_SIZE.
thus we should never produce such entries.

also: introduce repository.MAX_DATA_SIZE that gives the max payload size.

### DIFF
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -712,9 +712,14 @@ class LoggedIO:
             key = None
         else:
             raise TypeError("_read called with unsupported format")
-        if size > MAX_OBJECT_SIZE or size < fmt.size:
-            raise IntegrityError('Invalid segment entry size [segment {}, offset {}]'.format(
-                segment, offset))
+        if size > MAX_OBJECT_SIZE:
+            # if you get this on an archive made with borg < 1.0.7 and millions of files and
+            # you need to restore it, you can disable this check by using "if False:" above.
+            raise IntegrityError('Invalid segment entry size {} - too big [segment {}, offset {}]'.format(
+                size, segment, offset))
+        if size < fmt.size:
+            raise IntegrityError('Invalid segment entry size {} - too small [segment {}, offset {}]'.format(
+                size, segment, offset))
         length = size - fmt.size
         data = fd.read(length)
         if len(data) != length:
--- a/borg/repository.py
+++ b/borg/repository.py
@@ -731,8 +731,12 @@ class LoggedIO:
         return size, tag, key, data
 
     def write_put(self, id, data, raise_full=False):
+        data_size = len(data)
+        if data_size > MAX_DATA_SIZE:
+            # this would push the segment entry size beyond MAX_OBJECT_SIZE.
+            raise IntegrityError('More than allowed put data [{} > {}]'.format(data_size, MAX_DATA_SIZE))
         fd = self.get_write_fd(raise_full=raise_full)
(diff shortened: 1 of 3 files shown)


borg/repository.py
```diff
@@ -712,8 +712,13 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags):
...
             raise TypeError("_read called with unsupported format")
-        if size > MAX_OBJECT_SIZE or size < fmt.size:
-            raise IntegrityError('Invalid segment entry size [segment {}, offset {}]'.format(
+        if size > MAX_OBJECT_SIZE:
+            # if you get this on an archive made with borg < 1.0.7 and many millions of files and
+            # you need to restore it, you can disable this check by using "if False:" above.
+            raise IntegrityError('Invalid segment entry size - too big [segment {}, offset {}]'.format(
+                segment, offset))
+        if size < fmt.size:
+            raise IntegrityError('Invalid segment entry size - too small [segment {}, offset {}]'.format(
```
enkore: Add fmt.size to both of these error messages?

docs/changes.rst
```diff
@@ -57,6 +57,14 @@ Security fixes:
 
 - fix security issue with remote repository access, #1428
 
+Bug fixes:
+
+- do not write objects to repository that are bigger than the allowed size,
+  borg will reject reading them, #1451.
+  IMPORTANT: if you created archives with many millions of files or
+             directories, please verify if they extract successfully (e.g. try
+             a "borg extract --dry-run").
```
enkore: A `borg list repo::archive in question` or check --archives-only would be sufficient.

borg/repository.py
ThomasWaldmann: how does that help?

enkore: May make it simpler to see what's going on, like segment/offset numbers

docs/changes.rst
ThomasWaldmann: yes, "list" is better (quicker) and sufficient. maybe rather not check, otherwise people might repair and that would just kill the archive due to the integrityerror caused by the huge size.

borg/repository.py
ThomasWaldmann: ah, i misunderstood you.
you did not mean the size of the header, but the total size.
ok, i'll add that.

enkore:
lgtm

ThomasWaldmann:
collapsed and locally tested. travis has issues currently, so I'll just merge it.

PR #1455, opened 2016-08-09
LZ4: dynamically enlarge the (de)compression buffer, fixes #1453

### LINKED ISSUE #1453: COMPR_BUFFER can be too small
https://mail.python.org/pipermail/borgbackup/2016q3/000354.html

```
> Exception: lz4 compress failed
```

crash very likely is because the lz4-compressed data did not fit into the COMPR_BUFFER, which is 1.1 * 2 ** 23 (8.8MB) size currently. the lz4 compressor code checks this and does not overflow the buffer, but fails.

2**23 == 8MB is the chunk buffer size, so normal file content data chunks can never need more buffer (neither for compression nor for decompression).

But: the archive item creation code does not use the chunker, so it can yield much more data, so that the buffer is not enough for the compressed data.

Both compression and decompression might or might not work, depending on whether the result fits into that 8.8MB buffer.

ThomasWaldmann:
the statically allocated COMPR_BUFFER was right size for chunks,
but not for the archive item which could get larger if you have
many millions of files/dirs.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -25,7 +25,7 @@ from .helpers import Error, location_validator, archivename_validator, format_li
     EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR, log_multi, PatternMatcher, ErrorIgnoringTextIOWrapper
 from .logger import create_logger, setup_logging
 logger = create_logger()
-from .compress import Compressor, COMPR_BUFFER
+from .compress import Compressor
 from .upgrader import AtticRepositoryUpgrader, BorgRepositoryUpgrader
 from .repository import Repository
 from .cache import Cache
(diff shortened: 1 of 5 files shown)


enkore:
lgtm

ThomasWaldmann:
At least for decompression (borg extract), the LZ4 class is instantiated once per chunk by the auto-detection code in Compressor. So, while processing a lot of chunks, we produce a lot of 8MB bytes objects that need to be garbage collected again.

So, maybe rather do a manual malloc(), try ... finally: free()?

enkore:
(thread-local) module level buffer as before?

ThomasWaldmann:
merge?

enkore:
yes

PR #1457, opened 2016-08-09
explain the confusing TypeError, fixes #1456

### LINKED ISSUE #1456: confusing msg with >= 1.0.7rc client against <1.0.7 remote repo
e.g. client: borg 1.0.7rc1 + server: borg 1.0.6

there will be always a TypeError traceback in the remote log about too many open() parameters given by the client.

this should be ignored if the server is older than 1.0.7 because it is a cosmetic side effect of the compatibility code we use to support old servers, too:
- we first try calling open() with 7 params (we needed to add 2 for locking and append-only, will work with servers >= 1.0.7)
- if that raises a TypeError on the server side, we call again with 5 params (will work with servers < 1.0.7)

ThomasWaldmann:


textshell:
Shouldn’t this be output when the error is caught in the repo opening code, so it doesn‘t show up on random other problems?

ThomasWaldmann:
@textshell we can only catch it in the new client code, we can not change already deployed, old server code.

and if the server gets updated, it does not happen any more.

textshell:
Sorry. Misread the patch. You are correct of course. I somehow thought this would be in handle_error.

ThomasWaldmann:
Somehow it seems that project coverage is not computed correctly.

ThomasWaldmann:
merge?

PR #1459, opened 2016-08-09
larger item metadata stream chunks, fixes #1452

### LINKED ISSUE #1452: per archive total item count limit
the archive item (which has a list of "item metadata stream" chunks) is currently stored as a single object.

we limit size of repo objects to MAX_OBJ_SIZE (20MB).
note: we did only enfore the limit on get() - not on put().

each list entry takes ~40 bytes (32 sha256 id + 2 * 4 for the size / csize), msgpacked a bit less maybe.

so, that means we can reference ~500.000 item metadata stream chunks per archive.

each item metadata stream chunk is ~16kiB (due to special, more fine grained and hardcoded item metadata chunker settings).

so that means the whole item metadata stream is limited to ~8GB.

the item entry size is determined by:
- file size (number of chunks * 40B)
- path lengths
- amount of ACLs, xattrs
- a constant small amount of file metadata (timestamps, uid/gid, mode, ...)

if the medium size of an item entry is 100B (small size file, no ACLs/xattrs), that means we have a limit of 80M files/directories per archive. if the medium size of an item entry is 2kB (~100MB size files), that means we have a limit of 4M files/directories per archive.

we need a short term change to relax this limit and also a long term change to resolve it completely, this issue is about the short term change.

also, this limit should be documented.

ThomasWaldmann:
increasing the mask (target chunk size) from 14 (16kiB) to 17 (128kiB).
this should reduce the amount of item metadata chunks an archive has to reference to 1/8.
this does not completely fix #1452, but at least enables a 8x larger item metadata stream.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -39,7 +39,7 @@ HASH_MASK_BITS = 21  # results in ~2MiB chunks statistically
 CHUNKER_PARAMS = (CHUNK_MIN_EXP, CHUNK_MAX_EXP, HASH_MASK_BITS, HASH_WINDOW_SIZE)
 
 # chunker params for the items metadata stream, finer granularity
-ITEMS_CHUNKER_PARAMS = (12, 16, 14, HASH_WINDOW_SIZE)
+ITEMS_CHUNKER_PARAMS = (15, 19, 17, HASH_WINDOW_SIZE)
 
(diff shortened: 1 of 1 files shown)


enkore:
Guess this is the best solution for now.

PR #1460, opened 2016-08-10
add transaction_id assertion

ThomasWaldmann:
an acd_cli (amazon cloud drive fuse filesystem) user had "borg init" crash in the line below.

by adding the assertion we tell that we do not expect the transaction_id to be None there,
so it is easier to differentiate from a random coding error.


ThomasWaldmann:
https://paster.li/?ca0b6991a47631f4#wu30RyaaGpgiHqKRLvpMyAq5Z6TcejXBuOLxXcXTffc=

enkore:
The glory of eventual consistency.

PR #1471, opened 2016-08-12
refactor buffer code into helpers.Buffer class

ThomasWaldmann:


enkore:
lgtm

borg/helpers.py
```diff
@@ -655,6 +656,47 @@ def decorated_function(*args):
     return decorated_function
 
 
+class Buffer:
+    """
+    provide a thread-local buffer
+    """
+    def __init__(self, alloc, size=4096, limit=None):
```
enkore: alloc..ator?

ThomasWaldmann:
yes, i'll add tests and rename back to allocator (had that first).

Issue #1442, opened 2016-08-05
Untracked segment


PostalBlab:
If borg enters Repository.compact_segments() with an untracked segment it is possible the segment never gets deleted and stays in REPO/data
Copy paste from my explanation in #borgbackup:
> my repo has 3626 segments and one archive. by deleting the archive, in Archiver.do_delete() with archive.delete(), 2 more segments (3627 and 3628) are created because of all the write_delete() calls. both getting added to Repository.compact. after archive.delete() got called repository.commit() is called. because of the self.io.write_commit() call, segment 3629 gets created and then repository.compact_segments() is called. segment 1-3626 are getting thrown into "unnused" but 3627 and 3628 have all these TAG_DELETE entries which triggers new write_delete() calls and more TAG_DELETE entries. 3629 gets stuffed with TAG_DELETED entries and reaches max_segment_size. 3630 is created and 3629 gets completly untracked.

This happens with 1.0.6

The bug still exist in 1.1.dev and is getting worse.
In Repository.compact_segments(), write_put() and write_delete() are called with raise_full=True. When max_segment_size is reached write_*() throws an LoggedIO.SegmentFull exception, complete_xfer() is called, which leads to closing the current segment and opening a new one and closing it because of the write_commit() call. This leads to more untracked segment files with only 17 byte.

ThomasWaldmann:
Fixed by #1448.

PR #1484, opened 2016-08-16
Track shadowing of log entries

### LINKED ISSUE #1396: borg check --repair does not repair object count mismatch
Tried to repair the broken repo from #1395, but couldn't:

```
(borg-env)tw@tux:~/w/borg$ borg check -v repo
Starting repository check
Completed repository check, no problems found.
Starting archive consistency check...
Analyzing archive 3.checkpoint (3/3)
Analyzing archive 2.checkpoint (2/3)
Analyzing archive 1.checkpoint (1/3)
Archive consistency check complete, no problems found.
```
```
(borg-env)tw@tux:~/w/borg$ borg check -v --repair repo
'check --repair' is an experimental feature that might result in data loss.
Starting repository check
Completed repository check, no problems found.
Starting archive consistency check...
Analyzing archive 3.checkpoint (3/3)
Analyzing archive 2.checkpoint (2/3)
Analyzing archive 1.checkpoint (1/3)
Archive consistency check complete, no problems found.
```

```
(borg-env)tw@tux:~/w/borg$ borg check -v repo
Starting repository check
Index object count mismatch. 92 != 93
Completed repository check, errors found.
```
```
(borg-env)tw@tux:~/w/borg$ borg check -v --repair repo
'check --repair' is an experimental feature that might result in data loss.
Starting repository check
Completed repository check, no problems found.
Starting archive consistency check...
Analyzing archive 3.checkpoint (3/3)
Analyzing archive 2.checkpoint (2/3)
Analyzing archive 1.checkpoint (1/3)
1 orphaned objects found!
Archive consistency check complete, problems found.
```
```
(borg-env)tw@tux:~/w/borg$ borg check -v --repair repo
'check --repair' is an experimental feature that might result in data loss.
Starting repository check
Completed repository check, no problems found.
Starting archive consistency check...
Analyzing archive 3.checkpoint (3/3)
Analyzing archive 2.checkpoint (2/3)
Analyzing archive 1.checkpoint (1/3)
1 orphaned objects found!
Archive consistency check complete, problems found.
```
```
(borg-env)tw@tux:~/w/borg$ borg check -v repo
Starting repository check
Index object count mismatch. 92 != 93
Completed repository check, errors found.
```

enkore:
Fixes #1396 

Note that for append-only repos with unmaterialized DELETEs this makes it behave a bit like #1442 when compaction is run (non-append-only), ie. the unmaterialized DELETEs won't really go away, because they're not in the index. Reinforcing the suggestion to do pruning without append-only.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -22,6 +22,7 @@ from .helpers import Location
 from .helpers import ProgressIndicatorPercent
 from .helpers import bin_to_hex
 from .locking import Lock, LockError, LockErrorT
+from .logger import create_logger
 from .lrucache import LRUCache
 from .platform import SaveFile, SyncFile, sync_dir
 
@@ -413,29 +414,36 @@ class Repository:
         index_transaction_id = self.get_index_transaction_id()
         segments = self.segments
         unused = []  # list of segments, that are not used anymore
+        logger = create_logger('borg.debug.compact_segments')
 
         def complete_xfer(intermediate=True):
             # complete the current transfer (when some target segment is full)
             nonlocal unused
             # commit the new, compact, used segments
             self.io.write_commit(intermediate=intermediate)
+            logger.debug('complete_xfer: wrote %scommit at segment %d', 'intermediate ' if intermediate else '',
+                         self.io.get_latest_segment())
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
+                logger.debug('complete_xfer: deleting unused segment %d', segment)
                 assert self.segments.pop(segment) == 0
                 self.io.delete_segment(segment)
                 del self.compact[segment]
             unused = []
 
+        logger.debug('compaction started.')
         for segment, freeable_space in sorted(self.compact.items()):
             if not self.io.segment_exists(segment):
+                logger.warning('segment %d not found, but listed in compaction data', segment)
                 del self.compact[segment]
                 continue
             segment_size = self.io.segment_size(segment)
             if segment_size > 0.2 * self.max_segment_size and freeable_space < 0.15 * segment_size:
-                logger.debug('not compacting segment %d for later (only %d bytes are sparse)',
-                             segment, freeable_space)
+                logger.debug('not compacting segment %d (only %d bytes are sparse)', segment, freeable_space)
                 continue
             segments.setdefault(segment, 0)
+            logger.debug('compacting segment %d with usage count %d and %d freeable bytes',
+                         segment, segments[segment], freeable_space)
             for tag, key, offset, data in self.io.iter_objects(segment, include_data=True):
                 if tag == TAG_PUT and self.index.get(key, (-1, -1)) == (segment, offset):
                     try:
@@ -492,6 +500,7 @@ class Repository:
             assert segments[segment] == 0
             unused.append(segment)
         complete_xfer(intermediate=False)
+        logger.debug('compaction completed.')
 
     def replay_segments(self, index_transaction_id, segments_transaction_id):
         # fake an old client, so that in case we do not have an exclusive lock yet, prepare_txn will upgrade the lock:
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -111,7 +111,12 @@ class Repository:
         self.io = None
         self.lock = None
         self.index = None
+        # This is an index of shadowed log entries during this transaction. Consider the following sequence:
+        # segment_n PUT A, segment_x DELETE A
+        # After the "DELETE A" in segment_x the shadow index will contain "A -> (n,)".
+        self.shadow_index = defaultdict(list)
         self._active_txn = False
+
         self.lock_wait = lock_wait
         self.do_lock = lock
         self.do_create = create
@@ -306,13 +311,10 @@ class Repository:
             except RuntimeError:
                 self.check_transaction()
                 self.index = self.open_index(transaction_id, False)
-        # This is an index of shadowed log entries during this transaction. Consider the following sequence:
-        # segment_n PUT A, segment_x DELETE A
-        # After the "DELETE A" in segment_x the shadow index will contain "A -> (n,)".
-        self.shadow_index = defaultdict(list)
         if transaction_id is None:
             self.segments = {}  # XXX bad name: usage_count_of_segment_x = self.segments[x]
             self.compact = FreeSpace()  # XXX bad name: freeable_space_of_segment_x = self.compact[x]
+            self.shadow_index.clear()
         else:
             if do_cleanup:
                 self.io.cleanup(transaction_id)
@@ -343,6 +345,11 @@ class Repository:
             else:
                 self.segments = hints[b'segments']
                 self.compact = FreeSpace(hints[b'compact'])
+            # Drop uncommitted segments in the shadow index
+            for key, shadowed_segments in self.shadow_index.items():
+                for segment in list(shadowed_segments):
+                    if segment > transaction_id:
+                        shadowed_segments.remove(segment)
 
     def write_index(self):
         hints = {b'version': 2,
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -306,6 +306,10 @@ class Repository:
             except RuntimeError:
                 self.check_transaction()
                 self.index = self.open_index(transaction_id, False)
+        # This is an index of shadowed log entries during this transaction. Consider the following sequence:
+        # segment_n PUT A, segment_x DELETE A
+        # After the "DELETE A" in segment_x the shadow index will contain "A -> (n,)".
+        self.shadow_index = defaultdict(list)
         if transaction_id is None:
             self.segments = {}  # XXX bad name: usage_count_of_segment_x = self.segments[x]
             self.compact = FreeSpace()  # XXX bad name: freeable_space_of_segment_x = self.compact[x]
@@ -420,9 +424,8 @@ class Repository:
             # complete the current transfer (when some target segment is full)
             nonlocal unused
             # commit the new, compact, used segments
-            self.io.write_commit(intermediate=intermediate)
-            logger.debug('complete_xfer: wrote %scommit at segment %d', 'intermediate ' if intermediate else '',
-                         self.io.get_latest_segment())
+            segment = self.io.write_commit(intermediate=intermediate)
+            logger.debug('complete_xfer: wrote %scommit at segment %d', 'intermediate ' if intermediate else '', segment)
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
                 logger.debug('complete_xfer: deleting unused segment %d', segment)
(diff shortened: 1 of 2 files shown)


src/borg/repository.py
```diff
@@ -22,6 +22,7 @@
 from .helpers import ProgressIndicatorPercent
```
enkore: ^ "all Repository-side logging goes through the pipe" actually not the case for borg.debug. because it's disabled by default,... borg serve may just turn it on always, the client will filter it.

src/borg/repository.py
```diff
@@ -305,6 +306,10 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
             except RuntimeError:
                 self.check_transaction()
                 self.index = self.open_index(transaction_id, False)
+        # This is an index of shadowed log entries during this transaction. Consider the following sequence:
+        # segment_n PUT A, segment_x DELETE A
+        # After the "DELETE A" in segment_x the shadow index will contain "A -> (n,)".
```
enkore: , because ...

src/borg/repository.py
```diff
@@ -413,31 +418,41 @@ def compact_segments(self):
...
         unused = []  # list of segments, that are not used anymore
+        logger = create_logger('borg.debug.compact_segments')
 
         def complete_xfer(intermediate=True):
             # complete the current transfer (when some target segment is full)
             nonlocal unused
             # commit the new, compact, used segments
             self.io.write_commit(intermediate=intermediate)
+            logger.debug('complete_xfer: wrote %scommit at segment %d', 'intermediate ' if intermediate else '',
+                         self.io.get_latest_segment())
```
ThomasWaldmann: this does some fs ops, even if debug logging is not enabled.

src/borg/repository.py
```diff
@@ -413,31 +418,41 @@ def compact_segments(self):
...
+                logger.debug('not compacting segment %d (only %d bytes are sparse)', segment, freeable_space)
                 continue
             segments.setdefault(segment, 0)
+            logger.debug('compacting segment %d with usage count %d and %d freeable bytes',
+                         segment, segments[segment], freeable_space)
             for tag, key, offset, data in self.io.iter_objects(segment, include_data=True):
-                if tag == TAG_PUT and self.index.get(key, (-1, -1)) == (segment, offset):
+                if tag == TAG_COMMIT:
+                    continue
+                in_index = self.index.get(key, (-1, -1)) == (segment, offset)
```
ThomasWaldmann: Use None instead of (-1, -1)?

src/borg/repository.py
```diff
@@ -447,8 +462,23 @@ def complete_xfer(intermediate=True):
...
+                        pass
                 elif tag == TAG_DELETE:
-                    if index_transaction_id is None or segment > index_transaction_id:
+                    # If the shadow index doesn't contain this key, then we can't say if there's a shadowed older tag,
+                    # therefore we do not drop the delete, but write it to a current segment.
+                    shadowed_put_exists = key not in self.shadow_index or any(
+                        # If the key is in the shadow index, then all entries in the shadow index must be newer than this
+                        # tag.
+                        shadowed <= segment and self.io.segment_exists(shadowed)
+                        for shadowed in self.shadow_index[key])
```
ThomasWaldmann: the comment describes a different, but maybe equivalent boolean expression.

src/borg/repository.py
```diff
@@ -705,6 +736,8 @@ def delete(self, id, wait=True):
             segment, offset = self.index.pop(id)
         except KeyError:
             raise self.ObjectNotFound(id, self.path) from None
+        self.shadow_index.setdefault(id, [])
+        self.shadow_index[id].append(segment)
```
ThomasWaldmann: setdefault also works like a get, so you can do the .append(segment) in the first line.

src/borg/testsuite/repository.py
```diff
@@ -293,7 +293,42 @@ def test_moved_deletes_are_tracked(self):
...
+
+        assert self.repository.io.segment_exists(put_segment)
+        assert not self.repository.io.segment_exists(delete_segment)
+
+        # Basic case, since the index survived this must be ok
+        assert H(1) not in self.repository
+        # Nuke index, force replay
+        os.unlink(os.path.join(self.repository.path, 'index.%d' % get_latest_segment()))
+        # Must not reappear
+        assert H(1) not in self.repository
```
ThomasWaldmann: good to have some more comments for this stuff. :)

src/borg/repository.py
```diff
@@ -482,7 +482,7 @@ def complete_xfer(intermediate=True):
                     shadowed_put_exists = key not in self.shadow_index or any(
```
enkore: Since we only track PUTs and we have never "just delete" a segment containing puts.

src/borg/repository.py
```diff
@@ -343,6 +345,12 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
             else:
                 self.segments = hints[b'segments']
                 self.compact = FreeSpace(hints[b'compact'])
+            # Drop uncommitted segments in the shadow index
+            for key, shadowed_segments in self.shadow_index.items():
+                for segment in sorted(shadowed_segments):
+                    if segment <= transaction_id:
+                        continue
+                    shadowed_segments.remove(segment)
```
ThomasWaldmann: reverse condition is easier code.

src/borg/repository.py
```diff
@@ -110,7 +111,12 @@ def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=Tru
         self.io = None
         self.lock = None
         self.index = None
+        # This is an index of shadowed log entries during this transaction. Consider the following sequence:
+        # segment_n PUT A, segment_x DELETE A
+        # After the "DELETE A" in segment_x the shadow index will contain "A -> (n,)".
```
ThomasWaldmann: A -> [n, ]

src/borg/repository.py
```diff
@@ -110,7 +111,12 @@ def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=Tru
...
+        self.shadow_index = {}
         self._active_txn = False
+
```
ThomasWaldmann: remove empty line?

src/borg/repository.py
```diff
@@ -447,8 +468,23 @@ def complete_xfer(intermediate=True):
...
+                        pass
                 elif tag == TAG_DELETE:
-                    if index_transaction_id is None or segment > index_transaction_id:
+                    # If the shadow index doesn't contain this key, then we can't say if there's a shadowed older tag,
+                    # therefore we do not drop the delete, but write it to a current segment.
+                    shadowed_put_exists = key not in self.shadow_index or any(
+                        # If the key is in the shadow index, then all entries in the shadow index must be newer than this
+                        # tag.
+                        shadowed <= segment
+                        for shadowed in self.shadow_index[key])
```
ThomasWaldmann: comments vs code confuses me.

ThomasWaldmann:
when collapsing the stuff at the end, could you create 2 changes - one with logging stuff, one with the rest?

src/borg/repository.py
ThomasWaldmann: maybe update the commit comment?

src/borg/repository.py
```diff
@@ -455,8 +461,22 @@ def complete_xfer(intermediate=True):
...
+                        self.shadow_index[key].remove(segment)
+                    except (KeyError, ValueError):
+                        pass
                 elif tag == TAG_DELETE:
-                    if index_transaction_id is None or segment > index_transaction_id:
+                    # If the shadow index doesn't contain this key, then we can't say if there's a shadowed older tag,
+                    # therefore we do not drop the delete, but write it to a current segment.
+                    shadowed_put_exists = key not in self.shadow_index or any(
+                        # If the key is in the shadow index, then no entry may be older than this tag, otherwise it
+                        # is shadowed.
```
ThomasWaldmann: we are getting closer :)

How about:
If the key is in the shadow index and there is any shadow list entry of an older or of this segment, we have a shadowed put.

Question:
For shadowed segment == this segment, is this always true?
The put could be older or newer than the delete in this case, depending on its offset in the segment.

enkore: That detail is not important, since we are here, so the entire segment will be compacted.

ThomasWaldmann: my comment is just a 1:1 translation of the code.

ThomasWaldmann: if shadowed_put_exists is True, the code below keeps the DELETE (transfers it to new segment).
- same segment (or older segment) PUT older than DELETE: keep the DELETE
- same segment DELETE oder than PUT: keep the DELETE (looks wrong)

enkore: Yes, that's too broad. The correct condition is `any(shadowed < segment)`.

src/borg/repository.py
enkore: Yeah I'll edit it later. That issue is #1491.

src/borg/repository.py
```diff
@@ -447,8 +467,22 @@ def complete_xfer(intermediate=True):
...
+                    # If this is a PUT shadowed by a later tag, then it will be gone when this segment is deleted after
+                    # this loop. Therefore it is removed from the shadow index.
+                    try:
+                        self.shadow_index[key].remove(segment)
+                    except (KeyError, ValueError):
+                        pass
                 elif tag == TAG_DELETE:
-                    if index_transaction_id is None or segment > index_transaction_id:
+                    # If the shadow index doesn't contain this key, then we can't say if there's a shadowed older tag,
+                    # therefore we do not drop the delete, but write it to a current segment.
```
ThomasWaldmann: is this comment will from the time before you added the "or any(...)"?

maybe it would be clearer if both comments are merged into one and updated to reflect the condition that is actually used.

src/borg/repository.py
```diff
@@ -447,8 +467,22 @@ def complete_xfer(intermediate=True):
...
+                    shadowed_put_exists = key not in self.shadow_index or any(
+                        # If the key is in the shadow index and there is any segment with an older PUT of this
+                        # key, we have a shadowed put.
+                        shadowed < segment for shadowed in self.shadow_index[key])
```
ThomasWaldmann: ok, this does not trigger in wrong cases any more, but it does not trigger in all correct cases any more (when the put is in current segment, but at lower offset than the del).

ThomasWaldmann: maybe just put tuples (segment, offset) into the shadow_index value lists?

enkore: > The put could be older or newer than the delete in this case, depending on its offset in the segment.

-

> That detail is not important, since we are here, so the entire segment will be compacted.

ie. order within the segment doesn't matter.

comments: first comment, first condition, second comment, second condition

ThomasWaldmann: ok, i read the code again and think it is correct for the same-segment case.

ThomasWaldmann:
besides unit tests, did this also get some practical testing?

enkore:
Yes; due to sync_dir I also have something in stock for improving on the delete retention. But that is a separate PR.

PR #1492, opened 2016-08-16
(cherry pick) Fix leak in borg-extract --strip-components

enkore:


ThomasWaldmann:
```
./borg/archiver.py:374:9: E731 do not assign a lambda expression, use a def

./borg/archiver.py:376:13: E731 do not assign a lambda expression, use a def

./borg/testsuite/archiver.py:1340:13: E265 block comment should start with '# '
```

borg/archiver.py
```diff
@@ -371,12 +371,13 @@ def do_extract(self, args, repository, manifest, key, archive):
         sparse = args.sparse
         strip_components = args.strip_components
         dirs = []
-        for item in archive.iter_items(lambda item: matcher.match(item[b'path']), preload=True):
+        filter = lambda item: matcher.match(item[b'path'])
+        if strip_components:
+            filter = lambda item: matcher.match(item[b'path']) and os.sep.join(item[b'path'].split(os.sep)[strip_components:])
```
ThomasWaldmann: if-else trigger pattern:
```
x = ...
if cond:
    x = ...
```

borg/remote.py
```diff
@@ -203,6 +203,8 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
             raise
 
     def __del__(self):
+        if len(self.responses):
+            logging.debug("still %d cached responses left in RemoteRepository" % (len(self.responses),))
```
ThomasWaldmann: iirc: pep8 wants a blank after a comma, even in such tuples.

borg/testsuite/archiver.py
```diff
@@ -1327,6 +1327,29 @@ def test_fuse(self):
...
+        self.cmd('init', self.repository_location)
+        self.create_regular_file('dir/file', contents=b"test file contents 123")
+        self.create_regular_file('dir/file2', contents=b"test file contents 345")
+        self.create_regular_file('skipped', contents=b"test file contents 567")
+        self.create_regular_file('skipped2', contents=b"test file contentsasdasd")
+        self.create_regular_file('skipped4', contents=b"sdfdsgdgfhttztu")
+        self.cmd('create', self.repository_location + '::test', 'input')
+        marker = 'cached responses left in RemoteRepository'
+        with changedir('output'):
+            #import rpdb2; rpdb2.start_embedded_debugger("nopass")
```
ThomasWaldmann: want to keep that here?

borg/testsuite/archiver.py
```diff
@@ -1370,3 +1369,26 @@ def test_get_args():
...
+    def test_empty(self):
+        matcher = PatternMatcher(fallback=True)
+        filter = Archiver.build_filter(matcher)
+        assert filter({b'path': 'anything'})
+
+    def test_strip_components(self):
+        matcher = PatternMatcher(fallback=True)
+        filter = Archiver.build_filter(matcher, strip_components=1)
+        assert not filter({b'path': 'shallow'})
+        assert not filter({b'path': 'shallow/'})  # can this even happen? paths are normalized...
```
enkore: (if this *doesn't* happen then there's the obvious optimization (path.count(/) >= strip_co) to be had here

borg/remote.py
enkore: imho this is the conventional way of writing it

ThomasWaldmann:
shall this go into 1.0.7 or rather 1.0.8?

enkore:
Test coverage is good enough that it can go into 1.0.7 I think.

borg/archiver.py
```diff
@@ -371,12 +381,11 @@ def do_extract(self, args, repository, manifest, key, archive):
...
         strip_components = args.strip_components
         dirs = []
-        for item in archive.iter_items(lambda item: matcher.match(item[b'path']), preload=True):
+        filter = self.build_filter(matcher, strip_components)
+        for item in archive.iter_items(filter, preload=True):
             orig_path = item[b'path']
             if strip_components:
                 item[b'path'] = os.sep.join(orig_path.split(os.sep)[strip_components:])
-                if not item[b'path']:
-                    continue
```
ThomasWaldmann: So, the problem this change avoids has its roots in archive.iter_items (or code called from there)?

borg/testsuite/archiver.py
```diff
@@ -1327,6 +1327,28 @@ def test_fuse(self):
     def test_debug_put_get_delete_obj(self):
         pass
 
+    def test_strip_components_doesnt_leak(self):
+        self.cmd('init', self.repository_location)
+        self.create_regular_file('dir/file', contents=b"test file contents 123")
+        self.create_regular_file('dir/file2', contents=b"test file contents 345")
+        self.create_regular_file('skipped', contents=b"test file contents 567")
+        self.create_regular_file('skipped2', contents=b"test file contentsasdasd")
+        self.create_regular_file('skipped4', contents=b"sdfdsgdgfhttztu")
```
ThomasWaldmann: if the contents does not matter, just have to be different, can we just have b"contents 1", b"contents 2", ...?

filenames: skipped1, skipped2, skipped3?

borg/archiver.py
enkore: The issue is that if preload is used for getting items then one *has to* retrieve the chunks of every item that was preloaded ( https://github.com/borgbackup/borg/blob/1.0-maint/borg/archive.py#L101 ). Otherwise RemoteRepo keeps sending get()s for chunks that won't be retrieved; the results of those get()s will then fill up the responses-mapping of RemoteRepo.

PR #1507, opened 2016-08-19
use patched LDLP-preserving pyinstaller

### LINKED ISSUE #1416: borg binary tests failing on wheezy32
I originally found this while testing my multiple filesystem code of course, but I destroyed the box and replicated it on master. I haven't seen these errors on any other distro. Here's the log: https://gist.github.com/c38dd79e6a271bc05ac68b2739626d12
. For ease of reading, here's just the failures section of Python 3.4: https://gist.github.com/832ae39ab55c8d9085ed4986f90030f3. It should be noted that other errors might be occurring on different versions.

ThomasWaldmann:
fixes #1416, see also: https://github.com/pyinstaller/pyinstaller/pull/2148

also: use waf --no-lsb (the bootloader does not build without it on our wheezy vagrant VM)

This fixes the fakeroot issues, when fakeroot can't find its libs.
Also it fixes a lot of tests that get disturbed by the stderr message fakeroot emits due to that.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -248,10 +248,13 @@ def install_pyinstaller_bootloader(boxname)
     . borg-env/bin/activate
     git clone https://github.com/pyinstaller/pyinstaller.git
     cd pyinstaller
-    git checkout v3.1.1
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
merge?

PR #1510, opened 2016-08-20
daemonize(): use os.devnull instead of hardcoded /dev/null

c4rlo:


src/borg/helpers.py
```diff
@@ -1001,7 +1001,7 @@ def daemonize():
     os.close(0)
     os.close(1)
     os.close(2)
-    fd = os.open('/dev/null', os.O_RDWR)
+    fd = os.open(os.devnull, os.O_RDWR)
```
ThomasWaldmann: well, os.fork is unix-only, so the portability advantage of os.devnull doesn't come to play.
otoh, it does no harm either.

merge?

c4rlo: Yeah, there's no real advantage, other than the fact that it's slightly nicer to use a predefined constant rather than hardcoding the string literal.

PR #1511, opened 2016-08-20
helpers.py: replace memoize with functools.lru_cache

c4rlo:


src/borg/helpers.py
```diff
@@ -722,17 +722,7 @@ def format_archive(archive):
...
-
-    def decorated_function(*args):
-        try:
-            return cache[args]
-        except KeyError:
-            val = function(*args)
-            cache[args] = val
-            return val
-    return decorated_function
+memoize = lru_cache(maxsize=None)
```
ThomasWaldmann: maybe go one step further and replace memoize usages by @lrucache?
also the maxsize=None should be checked then, whether it makes sense or can be harmful.

ThomasWaldmann: looks like it is only used in pwd context, so I'ld say replace them, but keep maxsize=None.

ThomasWaldmann:
thanks!

ThomasWaldmann:
https://github.com/borgbackup/borg/tree/0839e6bed39347dc3c46f97e7b3764a2729c101b

that was the change that added memoize. it pre-dates python 3.2. :)

PR #1512, opened 2016-08-20
archiver: minor refactor (DRY)

c4rlo:


ThomasWaldmann:
not sure if we want that. while it is cleaner and avoids having to deal with the args list at 2 places, it increases recursion depth by 2x.

c4rlo:
True. Maybe not a good idea then.

PR #1515, opened 2016-08-20
Stabilize deleted segments on disk, remove 6425d16aa

enkore:
(wip)


enkore:
This PR was too hasty (obviously; was getting late); the aim was to reduce the number of DELETEs we have to save over to new segments by syncing deletes to disk in order to avoid inconsistencies.

However, this doesn't work out with index replaying; if segment 1 had a PUT, commit, index.1 and we deleted that later, commit, delete segment 1, commit, crash (before writing the index). Index replay would go from index.1 and read the new segments. Even though the log itself doesn't contain the PUT anymore, the index would still think it's in segment 1, since no DELETE was seen.

PR #1518, opened 2016-08-21
travis: test fuse-enabled borg

ThomasWaldmann:


PlasmaPower:
Do we also want to test osxfuse? Currently, I think tests will just fail on OS X because FUSE won't be installed.

ThomasWaldmann:
@PlasmaPower i installed osxfuse 3.4 now, works.

ThomasWaldmann:
merge? more coverage! \o/

PR #1522, opened 2016-08-22
WIP: Simplify repo tests

ThomasWaldmann:
rather apply against 1.0-maint to keep src better in sync?


enkore:
> rather apply against 1.0-maint to keep src better in sync?

yes

PR #1525, opened 2016-08-22
borg serve throws exception while trying to close when it gets eof at start

textshell:
While checking if my ssh forced command worked i started an ssh session to the backup server manually and got the following stacktrace:
```
$ echo -n | borg serve
$LOG ERROR borg.archiver Remote: Local Exception.
  File "/home/martin/sw-dev/borg/src/borg/archiver.py", line 2395, in main
  File "/home/martin/sw-dev/borg/src/borg/archiver.py", line 2328, in run
    return args.func(args)
  File "/home/martin/sw-dev/borg/src/borg/archiver.py", line 175, in do_serve
    return RepositoryServer(restrict_to_paths=args.restrict_to_paths, append_only=args.append_only).serve()
  File "/home/martin/sw-dev/borg/src/borg/remote.py", line 91, in serve
    self.repository.close()
AttributeError: 'NoneType' object has no attribute 'close'

Platform: Linux eclipse.s 4.4.1 #1 SMP PREEMPT Mon Feb 1 19:14:30 CET 2016 x86_64 
Linux: debian 8.4 
Borg: 1.0.8.dev442+ng81dd381  Python: CPython 3.4.2
PID: 4544  CWD: /home/martin/sw-dev/borg
sys.argv: ['/home/martin/sw-dev/borg/borg-env/bin/borg', 'serve']
SSH_ORIGINAL_COMMAND: None
```

This PR fixes this to be more friendly:

    $ echo -n | borg serve
    Borg 1.0.8.dev8+nga7c370b: Got connection close before repository was opened

or 

    $ echo something | borg serve
    $LOG ERROR Remote: Borg 1.0.8.dev8+nga7c370b: Got unexpected RPC data format from client.

### DIFF
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -89,6 +89,9 @@ class RepositoryServer:  # pragma: no cover
                 if not data:
                     if self.repository is not None:
                         self.repository.close()
+                    else:
+                        os.write(stderr_fd, "Borg {}: Got connection close before repository was opened.\n"
+                                 .format(__version__).encode())
                     return
                 unpacker.feed(data)
                 for unpacked in unpacker:
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -37,6 +37,14 @@ class InvalidRPCMethod(Error):
     """RPC method {} is not valid"""
 
 
+class UnexpectedRPCDataFormatFromClient(Error):
+    """Borg {}: Got unexpected RPC data format from client."""
+
+
+class UnexpectedRPCDataFormatFromServer(Error):
+    """Got unexpected RPC data format from server."""
+
+
 class RepositoryServer:  # pragma: no cover
     rpc_methods = (
         '__len__',
@@ -87,7 +95,7 @@ class RepositoryServer:  # pragma: no cover
                     if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
                         if self.repository is not None:
                             self.repository.close()
-                        raise Exception("Unexpected RPC data format.")
+                        raise UnexpectedRPCDataFormatFromClient(__version__)
                     type, msgid, method, args = unpacked
                     method = method.decode('ascii')
                     try:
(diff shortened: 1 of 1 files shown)


PlasmaPower:
:+1: - I'd run into this before.

borg/remote.py
```diff
@@ -79,13 +83,18 @@ def serve(self):
             if r:
                 data = os.read(stdin_fd, BUFSIZE)
                 if not data:
-                    self.repository.close()
+                    if self.repository is not None:
+                        self.repository.close()
+                    else:
+                        os.write(stderr_fd, "Borg {}: Got connection close before repository was opened\n"
```
ThomasWaldmann: add a dot at sentence end

borg/remote.py
```diff
@@ -326,7 +335,7 @@ def handle_error(error, res):
                     self.unpacker.feed(data)
                     for unpacked in self.unpacker:
                         if not (isinstance(unpacked, tuple) and len(unpacked) == 4):
-                            raise Exception("Unexpected RPC data format.")
+                            raise Exception("Unexpected RPC data format while reading from server.")
```
ThomasWaldmann: use Error subclass here also?

textshell: Not sure about Error or ErrorWithTraceback here. For the server side code i’m sure that we never need a trace.

ThomasWaldmann: ErrorWithTraceback must be used to catch exceptions that are hard to reproduce. So we have a traceback if sth goes wrong.

Error can be used for other cases. Since my recent change, it emits a traceback on DEBUG level - so if you can reproduce the problem, you can just invoke with --debug and see the traceback.

textshell: So what does your crystal ball say? Are there going to be hard to reproduce data format errors in the future?

ThomasWaldmann: Maybe. OTOH, the traceback is not very useful with the data that was actually received and could not be processed.

ThomasWaldmann: another thing: maybe keep the phrase basically the same, you can still indicate from where the bad data came.

enkore:
lgtm

PR #1526, opened 2016-08-22
extract: When doing a partial restore don't leak prefetched chunks.

textshell:
extract: When doing a partial restore don't leak prefetched chunks.
    
  The filter function passed to iter_items (with preload=True) may never return True for items that
    are not really extracted later because that would leak prefetched items.
    
  For restoring hard linked files the item containing the actual chunks might not be matched
    or implicitly removed from the restore by strip_components. For this reason the chunk list or all
    items that can potentially be used as hardlink target needs to be stored.
    
  To achive both requirements at the same time the filter function needs to store the needed information
    for the hardlinks while not returning True just because it could be a hardlink target.
    
Also contains fixes by encore to fix miscalculation of progress wrt hardlinks and reduces memory usage  by storing hardlink masters only when the file is not going to be extracted.

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -2204,25 +2204,25 @@ def test_compare_chunk_contents():
 
 class TestBuildFilter:
     @staticmethod
-    def item_is_hardlink_master(item):
+    def peek_and_store_hardlink_masters(item):
         return False
 
     def test_basic(self):
         matcher = PatternMatcher()
         matcher.add([parse_pattern('included')], True)
-        filter = Archiver.build_filter(matcher, self.item_is_hardlink_master, 0)
+        filter = Archiver.build_filter(matcher, self.peek_and_store_hardlink_masters, 0)
         assert filter(Item(path='included'))
         assert filter(Item(path='included/file'))
         assert not filter(Item(path='something else'))
 
     def test_empty(self):
         matcher = PatternMatcher(fallback=True)
-        filter = Archiver.build_filter(matcher, self.item_is_hardlink_master, 0)
+        filter = Archiver.build_filter(matcher, self.peek_and_store_hardlink_masters, 0)
         assert filter(Item(path='anything'))
 
     def test_strip_components(self):
         matcher = PatternMatcher(fallback=True)
-        filter = Archiver.build_filter(matcher, self.item_is_hardlink_master, strip_components=1)
+        filter = Archiver.build_filter(matcher, self.peek_and_store_hardlink_masters, strip_components=1)
         assert not filter(Item(path='shallow'))
         assert not filter(Item(path='shallow/'))  # can this even happen? paths are normalized...
         assert filter(Item(path='deep enough/file'))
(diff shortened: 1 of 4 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -2210,14 +2210,14 @@ def item_is_hardlink_master(item):
     def test_basic(self):
         matcher = PatternMatcher()
         matcher.add([parse_pattern('included')], True)
-        filter = Archiver.build_filter(matcher, self.item_is_hardlink_master)
+        filter = Archiver.build_filter(matcher, self.item_is_hardlink_master, 0)
```
ThomasWaldmann: rename the item_is_hardlink_master method?

textshell: yes.

enkore:
> Known problems: When using progress indication the calculated extracted_size now can be smaller
> than the actual extracted size in presence of hard links (master is not restored) instead of
bigger (potential master not used in restore).

Pull branch `c/textshell/preload-hardlink` from https://github.com/enkore/borg.git for fix

Tested it for the cases that came to mind, always produced correct output so far.

enkore:
> Known problems: When using progress indication the calculated extracted_size now can be smaller
> than the actual extracted size in presence of hard links (master is not restored) instead of
bigger (potential master not used in restore).

Pull branch `c/textshell/preload-hardlink` from https://github.com/enkore/borg.git for fix

Tested it for the cases that came to mind, always produced correct output so far.

Other than that this lgtm.

textshell:
Hmm, i just merged your fixed. But maybe that was too fast^^

Anyway, the change i merged looks good. I’m not sure about the part with only storing filtered out items into hardlink masters. I think that makes the progress calculation wrong again. Also i’m not sure if --strip-components and the original hard link code play well together. I have to take a look at that when less tired.

enkore:
Should be done with 2a9b4c8

enkore:
note: build failed because travis OSX infrastructure collapsed again

textshell:
anything else that needs to be done? I think this one doesn‘t need any squashing

enkore:
The OSX builds still lag behind.

Merge?

textshell:
ok with me. I don‘t expect the OSX results to be any different.

src/borg/testsuite/archiver.py
```diff
@@ -2204,25 +2204,25 @@ def ccc(a, b):
 
 class TestBuildFilter:
     @staticmethod
-    def item_is_hardlink_master(item):
-        return False
+    def peek_and_store_hardlink_masters(item, matched):
+        pass
```
ThomasWaldmann: do we have tests that test this function?

ThomasWaldmann:
btw, 1.1b1 is due tomorrow. please decide by assigning a milestone whether this should go in or not.

src/borg/testsuite/archiver.py
enkore: there are tests regarding the whole hardlink_masters affair

enkore:
this should be good for b1, besides the usual unit testing this now also saw quite some practical testing...

PR #1528, opened 2016-08-23
Addjust border color

svx:
This changes the border-color of the search to match with the background-color.

borg-border-color

### DIFF
--- a/docs/borg_theme/css/borg.css
+++ b/docs/borg_theme/css/borg.css
@@ -9,6 +9,11 @@
     background-color: #000000 !important;
 }
 
+.wy-side-nav-search input[type="text"] {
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
thanks for spotting and fixing this!

ThomasWaldmann:
hmm, PR was against master (not 1.0-maint), so this won't update the 1.0.x docs.

svx:
Do you want another PR against 1.0-maint ? Or do you cherry-pick ?

ThomasWaldmann:
If you can easily do another PR with the cherrypick, that would be appreciated.

In case you do more changes, keep branch-choosing in mind. :)

/me needs to document this in the developer docs...

PR #1531, opened 2016-08-23
work around fuse xattr test issue with recent fakeroot

ThomasWaldmann:
fakeroot >= 1.20.2 "supports" xattrs, but this support somehow leads to the fuse tests not seeing
the xattrs in fuse, because the file visible in the fuse mount was not created via fakeroot.

### DIFF
--- a/borg/testsuite/archiver.py
+++ b/borg/testsuite/archiver.py
@@ -276,8 +276,14 @@ class ArchiverTestCase(ArchiverTestCaseBase):
                 os.path.join(self.input_path, 'hardlink'))
         # Symlink
         os.symlink('somewhere', os.path.join(self.input_path, 'link1'))
-        if xattr.is_enabled(self.input_path):
-            xattr.setxattr(os.path.join(self.input_path, 'file1'), 'user.foo', b'bar')
+        self.create_regular_file('fusexattr', size=1)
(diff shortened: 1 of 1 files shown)


borg/testsuite/archiver.py
```diff
@@ -276,8 +276,12 @@ def create_test_files(self):
...
         # Symlink
         os.symlink('somewhere', os.path.join(self.input_path, 'link1'))
-        if xattr.is_enabled(self.input_path):
-            xattr.setxattr(os.path.join(self.input_path, 'file1'), 'user.foo', b'bar')
+        self.create_regular_file('fusexattr', size=1)
+        if not xattr.XATTR_FAKEROOT and xattr.is_enabled(self.input_path):
+            # ironically, due to the way how fakeroot works, comparing fuse
+            # file xattrs to orig file xattrs will FAIL if fakeroot supports
+            # xattrs, thus we only set the xattr if XATTR_FAKEROOT is False.
+            xattr.setxattr(os.path.join(self.input_path, 'fusexattr'), 'user.foo', b'bar')
```
enkore: Because fakeroot with xattr-support does not propagate xattrs of the underlying file into "fakeroot space". Because the xattrs exposed by borgfs *are* these of an underlying file (from fakeroots point of view) they are invisible to the test process inside the fakeroot.

ThomasWaldmann:
I added your comment to the src.

Issue #1533, opened 2016-08-25
Handling FUSE filesystem errors


RealityFork:
Using: Borg 1.0.2, Ubuntu Server 16.04

I am trying to use borg with the cloud storage system Backblae B2 using [b2_fuse](https://github.com/sondree/b2_fuse), mounting a b2 bucket as the 'data' directory inside a borg repo.
When the FUSE filesystem fails mid-backup, borg fails badly and reads some files in the repo when the user resumes a backup. On large backups borg may read a large amount of data, which may cost the user (most cloud storage charges for downloads).

Steps to reproduce:
- Init a borg repo `borg init /backup`
- Copy contents of data folder to b2 bucket
- mount b2 bucket over data folder `python b2fuse.py /backup/data`
- Create a backup and let it finish `borg create /backup::1 /flash`
- Change the contents of `/flash` and start another backup `borg create /backup::2 /flash`
- Kill the fuse filesystem process. The resulting error from borg is this:

```
~ borg create -psv /backup/::2 /flash
Enter passphrase for key /backup:
^CException ignored in: <bound method Repository.__del__ of <Repository /backup>>
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 71, in __del__
    assert False, "cleanup happened in Repository.__del__"
AssertionError: cleanup happened in Repository.__del__
Local Exception.
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 80, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 246, in do_create
    create_inner(archive, cache)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 218, in create_inner
    read_special=args.read_special, dry_run=dry_run)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 300, in _process
    read_special=read_special, dry_run=dry_run)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 274, in _process
    status = archive.process_file(path, st, cache, self.ignore_inode)
  File "/usr/lib/python3/dist-packages/borg/archive.py", line 557, in process_file
    chunks.append(cache.add_chunk(self.key.id_hash(chunk), chunk, self.stats))
  File "/usr/lib/python3/dist-packages/borg/cache.py", line 365, in add_chunk
    self.repository.put(id, data, wait=False)
[...]

Issue #1287, opened 2016-07-08
borg create throws os.fsync error on cifs mounted repo


alraban:
Observed both on Arch linux systems running the binary package and Debian GNU/Linux systems compiling from pip. With borg 1.05, any attempt to create a new archive in a repository that is accessible via cifs begins to run the create and then errors as soon as it reaches the first new file that needs to be written.  Backups to purely local filesystems or remote backups via ssh with the same repositories appear to work perfectly.

The systems in question had no similar problem with any prior versions of borg, and a repository check on the relevant repositories comes back clean.

An example of the errors thrown:

```
Exception ignored in: <bound method Repository.__del__ of <Repository /mnt/backups/Saturnine.borg>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/borg/repository.py", line 72, in __del__
    self.close()
  File "/usr/local/lib/python3.4/dist-packages/borg/repository.py", line 189, in close
    self.io.close()
  File "/usr/local/lib/python3.4/dist-packages/borg/repository.py", line 532, in close
    self.close_segment()
  File "/usr/local/lib/python3.4/dist-packages/borg/repository.py", line 752, in close_segment
    self._write_fd.flush()
ValueError: flush of closed file
Local Exception.
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 81, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 247, in do_create
    create_inner(archive, cache)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 219, in create_inner
    read_special=args.read_special, dry_run=dry_run)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 301, in _process
    read_special=read_special, dry_run=dry_run)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 301, in _process
    read_special=read_special, dry_run=dry_run)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 301, in _process
    read_special=read_special, dry_run=dry_run)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 301, in _process
    read_special=read_special, dry_run=dry_run)
  File "/usr/local/lib/python3.4/dist-packages/borg/archiver.py", line 301, in _process
[...]

PR #1534, opened 2016-08-25
traceback-less messaging for OSErrors

enkore:
- still has the last line of the TB which is un-scary but also the most
  important one
- error line much more prominently (1st line)
- reduced sysinfo() output (borg + python version)

new output would look like this:

```
$ borg list testrepo
I/O error: [Errno 2] No such file or directory: 'doesntexist'

Location:
  File "/home/mabe/Projekte/_oss/borg/src/borg/helpers.py", line 175, in __init__
    open('doesntexist', 'rb')

Borg: 1.0.8.dev445+ng4d214e2 Python: CPython 3.5.2
```

discuss!

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2422,6 +2422,17 @@ def main():  # pragma: no cover
         tb_log_level = logging.ERROR if important else logging.DEBUG
         tb = sysinfo()
         exit_code = EXIT_ERROR
+    except OSError as e:
+        at_frame = traceback.extract_tb(e.__traceback__, limit=-1)[0]
+        msg = textwrap.dedent("""
+        I/O error: {error}
+
+        At: {frame.filename}:{frame.lineno} ({frame.name})
+        {sysinfo}
+        """).strip().format(error=e, frame=at_frame, sysinfo=sysinfo(short=True))
+        tb_log_level = logging.DEBUG
+        tb = '%s\n%s' % (traceback.format_exc(), sysinfo())
+        exit_code = EXIT_ERROR
     except Exception:
         msg = 'Local Exception'
         tb_log_level = logging.ERROR
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1197,12 +1197,14 @@ class ProgressIndicatorEndless:
         print(file=self.file)
 
 
-def sysinfo():
+def sysinfo(short=False):
     info = []
     info.append('Platform: %s' % (' '.join(platform.uname()), ))
     if sys.platform.startswith('linux'):
         info.append('Linux: %s %s %s' % platform.linux_distribution())
     info.append('Borg: %s  Python: %s %s' % (borg_version, platform.python_implementation(), platform.python_version()))
+    if short:
+        return info[-1]
     info.append('PID: %d  CWD: %s' % (os.getpid(), os.getcwd()))
     info.append('sys.argv: %r' % sys.argv)
     info.append('SSH_ORIGINAL_COMMAND: %r' % os.environ.get('SSH_ORIGINAL_COMMAND'))


enkore:
(#1533)

enkore:
discussion/irc:

- If the exception has filename / filename2, then the "Location" bit isn't really needed (better name? "Code:"? "In/At:"? - maybe less traceback-y formatting: `At: /some/file.py:line`) -- if the format were shorter it can be left in always I think
- If the exceptions has no filename (which sadly is the case for most data ops like read/write/seek/tell/sync/...) then that bit can be helpful

enkore:
```
I/O error: [Errno 2] No such file or directory: 'lalalalalalala'

At: /home/mabe/Projekte/_oss/borg/src/borg/helpers.py:174
Borg: 1.0.8.dev445+ng4d214e2  Python: CPython 3.5.2
```

src/borg/archiver.py
```diff
@@ -2422,6 +2422,17 @@ def main():  # pragma: no cover
         tb_log_level = logging.ERROR if important else logging.DEBUG
         tb = sysinfo()
         exit_code = EXIT_ERROR
+    except OSError as e:
+        at_frame = traceback.extract_tb(e.__traceback__, limit=-1)[0]
+        msg = textwrap.dedent("""
+        I/O error: {error}
+
+        At: {frame.filename}:{frame.lineno} ({frame.name})
```
ThomasWaldmann: if the code where it crashes is some low-level routine that is called from a lot of places, this is almost as useless as attic's "AttributeError" (it did not say what/where) message.

as people usually do not run on debug level, we will get bug reports with not enough information.
and if the issue is not easily reproducable, that will be all we have.

enkore: yes, this "RFC" is about how we want to balance the error reporting here. That's a question of gauging probabilities plus some "UX".

- If there are IO errors (actual EIO, or permission issues, or missing files, full disks, ...) the user wants to know where (in the FS / which disk)
  - But only a select few exceptions actually have enough information for telling that (practically only errors on open())
  - The next best thing would be (if somewhat convoluted) telling where (in the code) ("At:" ...)
  - What's kinda bad is throwing a long traceback that's not readily understood and tends to obscure the error (#1287 is a good example, the actual error is somewhere in the middle, followed and preceded by dozens of lines).

- If there are bugs *we* want to know if/how/where they happen in the code. I would argue that in most cases the traceback isn't all that important, because most likely it is incorrect application of API to an object (#1287 is a fine specimen of this), which I would say is usually clear from the location in the code itself. State errors were not OSErrors in the past (writing|flushing|reading... a closed file is a ValueError for example).
- There is also the question assuming that we have a bug that caused this specific error and is *not* reproducible at all (and actually a bug in borg and not a problem in a lower layer), how likely it is that the traceback contains useful information that helps analyse it beyond this. In the past, with other errors, a traceback wasn't sufficient at times, since it didn't have important information (like locals).

  So I'm not sure whether that's a probable case we need to consider.

enkore: I'd love to hear other opinions here as well (as always! :)

enkore:
Bumping this one.

We could also write a detailed trace to some (user-only-readable) file to $TMP, while keeping the normal message concise. Like so:

```
I/O error: [Errno 2] No such file or directory: '/home/mabe/Projekte/_oss/borg/testrepo/data/0/1227'

At: /home/mabe/Projekte/_oss/borg/src/borg/repository.py:887 (get_fd)
Borg: 1.0.8.dev445+ng4d214e2  Python: CPython 3.5.2
In case you suspect that this is a bug in Borg,
detailed diagnostics were written to /tmp/borg-crash-912jkfasd.log
```

ThomasWaldmann:
The problem with dumping logs somewhere is that:
- on /tmp they might be not available any more when requested (tmpfs)
- in ~, they might pile up over time

So, if we really want that we'ld need the persistent location + kind of logrotate there.

enkore:
I don't think we need logrotate for crash logs.

MK-42:
+1 for dumping the stack trace and any other possibly useful information to a file in ~.
If only stack traces are logged there that file is unlikely to get large (hopefully).

The console output should point users here, informing them of the log file and that they should keep it for reference. Is there confidential information in it? If not, they could be advised to attach it right away.

PR #1536, opened 2016-08-26
ProgressIndicatorPercent / extract minor cosmetic enhancement

enkore:
Instead of

```
Calculating Size
Extracting files xx %
```

now

(short) `Calculating Size`, then `Extracting files xx %` in the same line, if no other output was requested, then after successful completion the output collapses completely ie.

```
$ borg extract ...
$
```


ThomasWaldmann:
lgtm

PR #1539, opened 2016-08-27
recreate: fix crash if archive doesn't have chunker_params

enkore:


ThomasWaldmann:
lgtm.

enkore:
don't think we have to wait an hour for travis here...

PR #1547, opened 2016-08-28
doc/faq: warn security issue on passphrase env.

weakish:
---
 docs/faq.rst | 2 ++
 1 file changed, 2 insertions(+)

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -105,6 +105,8 @@ key file based encryption with a blank passphrase. See
           the environment of a process is `accessible only to that
           user
           <https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009>`_.
+          Using the ``env`` command or inline shell scripts may also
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -105,6 +105,8 @@ key file based encryption with a blank passphrase. See
           the environment of a process is `accessible only to that
           user
           <https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009>`_.
+          Using the ``env`` command or inline shell scripts may also
+          cause the passphrase to be recorded in shell history file.
```
ThomasWaldmann: So, I inlined (sourced) a shell script and used env, here is my .bash_history:
```
vim secret.sh
. secret.sh 
env
exit
```
So, what do you really mean?

weakish: Hmm, I misread `inline shell scripts` as `command line shell` such as `env ... borg` directly on command line.

PR #1554, opened 2016-08-30
Archives filters

funkyfuture:
follow-up of #1369

this adds the arguments `--first`, `--last` and `--sort-by`, and groups them with `--prefix` as archives filters that can be applied on repository targets.

usable with `check`, `delete`, `info` and `list`.

use-case in a script:

```bash
while used_space_exceeds_pcent; do
    log 'Used space exceeds ${MAX_DISK_USAGE}%, deleting oldest backup archive.'
    borg delete --first 1 --force $BORG_REPOSITORY & wait
done
```

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -879,34 +879,38 @@ class Archiver:
             write = sys.stdout.buffer.write
 
         if args.location.archive:
-            matcher, _ = self.build_matcher(args.excludes, args.paths)
-            with Cache(repository, key, manifest, lock_wait=self.lock_wait) as cache:
-                archive = Archive(repository, key, manifest, args.location.archive, cache=cache,
-                                  consider_part_files=args.consider_part_files)
-
-                if args.format is not None:
-                    format = args.format
-                elif args.short:
-                    format = "{path}{NL}"
-                else:
-                    format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{NL}"
-                formatter = ItemFormatter(archive, format)
-
-                for item in archive.iter_items(lambda item: matcher.match(item.path)):
-                    write(safe_encode(formatter.format_item(item)))
+            return self._list_archive(args, repository, manifest, key, write)
         else:
+            return self._list_repository(args, manifest, write)
+
+    def _list_archive(self, args, repository, manifest, key, write):
+        matcher, _ = self.build_matcher(args.excludes, args.paths)
+        with Cache(repository, key, manifest, lock_wait=self.lock_wait) as cache:
+            archive = Archive(repository, key, manifest, args.location.archive, cache=cache,
+                              consider_part_files=args.consider_part_files)
             if args.format is not None:
                 format = args.format
             elif args.short:
-                format = "{archive}{NL}"
+                format = "{path}{NL}"
             else:
-                format = "{archive:<36} {time} [{id}]{NL}"
-            formatter = ArchiveFormatter(format)
+                format = "{mode} {user:6} {group:6} {size:8} {isomtime} {path}{extra}{NL}"
+            formatter = ItemFormatter(archive, format)
 
-            for archive_info in manifest.archives.list(sort_by='ts'):
-                if args.prefix and not archive_info.name.startswith(args.prefix):
-                    continue
-                write(safe_encode(formatter.format_item(archive_info)))
+            for item in archive.iter_items(lambda item: matcher.match(item.path)):
+                write(safe_encode(formatter.format_item(item)))
+        return self.exit_code
+
+    def _list_repository(self, args, manifest, write):
+        if args.format is not None:
+            format = args.format
+        elif args.short:
+            format = "{archive}{NL}"
+        else:
+            format = "{archive:<36} {time} [{id}]{NL}"
+        formatter = ArchiveFormatter(format)
+
+        for archive_info in self._get_filtered_archives(args, manifest):
+            write(safe_encode(formatter.format_item(archive_info)))
 
         return self.exit_code
 
@@ -2043,8 +2047,6 @@ class Archiver:
         subparser.add_argument('--format', '--list-format', dest='format', type=str,
                                help="""specify format for file listing
                                 (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NL}")""")
-        subparser.add_argument('-P', '--prefix', dest='prefix', type=prefix_spec,
-                               help='only consider archive names starting with this prefix')
         subparser.add_argument('-e', '--exclude', dest='excludes',
                                type=parse_pattern, action='append',
                                metavar="PATTERN", help='exclude paths matching PATTERN')
@@ -2056,6 +2058,7 @@ class Archiver:
                                help='repository/archive to list contents of')
         subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
                                help='paths to list; patterns are supported')
+        self.add_archives_filters_args(subparser)
 
         mount_epilog = textwrap.dedent("""
         This command mounts an archive as a FUSE filesystem. This can be useful for
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -770,11 +770,31 @@ class Archiver:
 
     @with_repository(exclusive=True, manifest=False)
     def do_delete(self, args, repository):
-        """Delete an existing repository or archive"""
+        """Delete an existing repository or archives"""
+        if any((args.location.archive, args.first, args.last, args.prefix)):
+            return self._delete_archives(args, repository)
+        else:
+            return self._delete_repository(args, repository)
+
+    def _delete_archives(self, args, repository):
+        """Delete archives"""
+        manifest, key = Manifest.load(repository)
+
         if args.location.archive:
-            manifest, key = Manifest.load(repository)
-            with Cache(repository, key, manifest, lock_wait=self.lock_wait) as cache:
-                archive = Archive(repository, key, manifest, args.location.archive, cache=cache)
+            archive_names = (args.location.archive,)
+        else:
+            archive_names = tuple(x.name for x in self._get_filtered_archives(args, manifest))
+            if not archive_names:
+                return self.exit_code
+
+        stats_logger = logging.getLogger('borg.output.stats')
+        if args.stats:
+            log_multi(DASHES, STATS_HEADER, logger=stats_logger)
+
+        with Cache(repository, key, manifest, lock_wait=self.lock_wait) as cache:
+            for i, archive_name in enumerate(archive_names, 1):
+                logger.info('Deleting {} ({}/{}):'.format(archive_name, i, len(archive_names)))
+                archive = Archive(repository, key, manifest, archive_name, cache=cache)
                 stats = Statistics()
                 archive.delete(stats, progress=args.progress, forced=args.forced)
                 manifest.write()
@@ -782,33 +802,41 @@ class Archiver:
                 cache.commit()
                 logger.info("Archive deleted.")
                 if args.stats:
-                    log_multi(DASHES,
-                              STATS_HEADER,
-                              stats.summary.format(label='Deleted data:', stats=stats),
-                              str(cache),
-                              DASHES, logger=logging.getLogger('borg.output.stats'))
-        else:
-            if not args.cache_only:
-                msg = []
-                try:
-                    manifest, key = Manifest.load(repository)
-                except NoManifestError:
-                    msg.append("You requested to completely DELETE the repository *including* all archives it may contain.")
-                    msg.append("This repository seems to have no manifest, so we can't tell anything about its contents.")
-                else:
-                    msg.append("You requested to completely DELETE the repository *including* all archives it contains:")
-                    for archive_info in manifest.archives.list(sort_by='ts'):
-                        msg.append(format_archive(archive_info))
-                msg.append("Type 'YES' if you understand this and want to continue: ")
-                msg = '\n'.join(msg)
-                if not yes(msg, false_msg="Aborting.", invalid_msg='Invalid answer, aborting.', truish=('YES', ),
-                           retry=False, env_var_override='BORG_DELETE_I_KNOW_WHAT_I_AM_DOING'):
-                    self.exit_code = EXIT_ERROR
-                    return self.exit_code
-                repository.destroy()
-                logger.info("Repository deleted.")
-            Cache.destroy(repository)
-            logger.info("Cache deleted.")
+                    log_multi(stats.summary.format(label='Deleted data:', stats=stats),
+                              DASHES, logger=stats_logger)
+                if not args.forced and self.exit_code:
+                    break
+            if args.stats:
+                stats_logger.info(str(cache))
+
+        return self.exit_code
+
+    def _delete_repository(self, args, repository):
+        """Delete a repository"""
+        if not args.cache_only:
+            msg = []
+            try:
+                manifest, key = Manifest.load(repository)
+            except NoManifestError:
+                msg.append("You requested to completely DELETE the repository *including* all archives it may "
+                           "contain.")
+                msg.append("This repository seems to have no manifest, so we can't tell anything about its "
+                           "contents.")
+            else:
+                msg.append("You requested to completely DELETE the repository *including* all archives it "
+                           "contains:")
+                for archive_info in manifest.archives.list(sort_by='ts'):
+                    msg.append(format_archive(archive_info))
+            msg.append("Type 'YES' if you understand this and want to continue: ")
+            msg = '\n'.join(msg)
+            if not yes(msg, false_msg="Aborting.", invalid_msg='Invalid answer, aborting.', truish=('YES',),
+                       retry=False, env_var_override='BORG_DELETE_I_KNOW_WHAT_I_AM_DOING'):
+                self.exit_code = EXIT_ERROR
+                return self.exit_code
+            repository.destroy()
+            logger.info("Repository deleted.")
+        Cache.destroy(repository)
+        logger.info("Cache deleted.")
         return self.exit_code
 
     @with_repository()
@@ -1969,6 +1997,7 @@ class Archiver:
         subparser.add_argument('location', metavar='TARGET', nargs='?', default='',
                                type=location_validator(),
                                help='archive or repository to delete')
+        self.add_archives_filters_args(subparser)
 
         list_epilog = textwrap.dedent("""
         This command lists the contents of a repository or an archive.
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2616,9 +2616,9 @@ class Archiver:
 
         group = filters_group.add_mutually_exclusive_group()
         group.add_argument('--first', dest='first', metavar='N', default=0, type=int,
-                           help='consider first N archives after other filter args were applied')
+                           help='consider first N archives after other filters were applied')
         group.add_argument('--last', dest='last', metavar='N', default=0, type=int,
-                           help='consider last N archives after other filter args were applied')
+                           help='consider last N archives after other filters were applied')
 
     def get_args(self, argv, cmd):
         """usually, just returns argv, except if we deal with a ssh forced command for borg serve."""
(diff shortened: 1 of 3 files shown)


funkyfuture:
strange, locally 42cd9805a47198ee4e8cce9441808f10e9841dc1 appears as second of these commits.

src/borg/archiver.py
```diff
@@ -2281,6 +2368,23 @@ def build_parser(self, prog=None):
...
+        sort_by_default = 'timestamp'
+        filters_group.add_argument('--sort-by', dest='sort_by', type=sort_by_spec, default=sort_by_default,
+                               help='Comma-separated list of sorting keys; valid keys are: {}; default is: {}'
+                               .format(', '.join(HUMAN_SORT_KEYS), sort_by_default))
+
+        group = filters_group.add_mutually_exclusive_group()
+        group.add_argument('--first', dest='first', metavar='N', default=0, type=int,
+                           help='delete N first archives')
+        group.add_argument('--last', dest='last', metavar='N', default=0, type=int,
+                           help='delete N last archives')
```
ThomasWaldmann: it always says "delete", no matter to where you add these filter args.
so, how about "consider N first/last archives".

src/borg/helpers.py
```diff
@@ -143,10 +143,15 @@ def __delitem__(self, name):
         del self._archives[name]
 
     def list(self, sort_by=None, reverse=False):
-        # inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
-        archives = self.values()  # [self[name] for name in self]
+        """ Inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
+
+            :rtype: A :class:`list` of :class:`borg.helpers.ArchiveInfo` instances
+        """
```
ThomasWaldmann: please have a look at the usual docstring style and use the same. do a fixup commit.

src/borg/archiver.py
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
         return args.func(args)
 
+    def _get_filtered_archives(self, args, manifest):
+        if args.location.archive:
+            logger.error('The options --first and --last can only be used on repository targets.')
+            self.exit_code = EXIT_ERROR
+            return []
```
ThomasWaldmann: Hmm, instead of logging an error and **continuing**, you could raise Error('...') which will terminate borg, with error status and show the given error msg.

src/borg/archiver.py
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
...
+
+        archives = manifest.archives.list()
+        if not archives:
+            logger.critical('There are no archives.')
+            self.exit_code = self.exit_code or EXIT_WARNING
+            return []
```
ThomasWaldmann: same here.

src/borg/archiver.py
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
...
 
+    def _get_filtered_archives(self, args, manifest):
+        if args.location.archive:
+            logger.error('The options --first and --last can only be used on repository targets.')
+            self.exit_code = EXIT_ERROR
+            return []
+
+        archives = manifest.archives.list()
+        if not archives:
+            logger.critical('There are no archives.')
```
ThomasWaldmann: why is this critical, why is this an error?

src/borg/archiver.py
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
...
+            return []
+
+        for sortkey in reversed(args.sort_by.split(',')):
+            archives.sort(key=attrgetter(sortkey))
+        if args.last:
+            archives.reverse()
+
+        n = args.first or args.last
+
+        return archives[:n]
```
ThomasWaldmann: nice :)

src/borg/helpers.py
enkore: (Which is basically [PEP 0257](https://www.python.org/dev/peps/pep-0257/) as per PEP 8 - ie. textual, imperative style "Return that-n-that" etc., `:param:` stuff only for many params; no type annotations)

src/borg/archiver.py
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
...
+            logger.critical('There are no archives.')
+            self.exit_code = self.exit_code or EXIT_WARNING
+            return []
+
+        for sortkey in reversed(args.sort_by.split(',')):
+            archives.sort(key=attrgetter(sortkey))
+        if args.last:
+            archives.reverse()
+
+        n = args.first or args.last
```
ThomasWaldmann: slightly better IMHO, but equivalent:
```
        if args.first:
            n = args.first
        else:  # if args.last
            n = args.last
            archives.reverse()
````

File: src/borg/archiver.py [hunk 8]
```diff
@@ -2343,6 +2358,27 @@ def run(self, args):
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
         return args.func(args)
 
+    def _get_filtered_archives(self, args, manifest):
```
ThomasWaldmann: nice method.

src/borg/helpers.py
```diff
@@ -645,6 +650,16 @@ def replace_placeholders(text):
...
+def sort_by_spec(text):
+    for token in text.split(','):
+        if token not in HUMAN_SORT_KEYS:
+            raise ValueError('Invalid sort key: %s' % token)
+    return text.replace('timestamp', 'ts')
+
+
 def safe_timestamp(item_timestamp_ns):
     try:
         return datetime.fromtimestamp(bigint_to_int(item_timestamp_ns) / 1e9)
```
ThomasWaldmann: maybe do all the fixups into 1 fixup commit for this commit, so you can git rebase -i later and squash the fixup commit into the base commit.

src/borg/testsuite/archiver.py
```diff
@@ -985,8 +988,13 @@ def test_delete(self):
         self.cmd('init', self.repository_location)
         self.cmd('create', self.repository_location + '::test', 'input')
         self.cmd('create', self.repository_location + '::test.2', 'input')
+        self.cmd('create', self.repository_location + '::test.3', 'input')
+        self.cmd('create', self.repository_location + '::another_test.1', 'input')
+        self.cmd('create', self.repository_location + '::another_test.2', 'input')
         self.cmd('extract', '--dry-run', self.repository_location + '::test')
         self.cmd('extract', '--dry-run', self.repository_location + '::test.2')
+        self.cmd('delete', '--prefix', 'another_', self.repository_location)
+        self.cmd('delete', '--last', '1', self.repository_location)
```
ThomasWaldmann: this is good for increasing coverage, but you have no assertion here, so it just tests that the code is not crashing.

src/borg/testsuite/archiver.py
```diff
@@ -57,6 +57,9 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, **kw):
         except subprocess.CalledProcessError as e:
             output = e.output
             ret = e.returncode
+        except SystemExit as e:
+            output = ''
+            ret = e.code
```
ThomasWaldmann: how is that change related to this changeset?

src/borg/archiver.py
```diff
@@ -30,7 +30,7 @@
 from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
 from .helpers import Error, NoManifestError
 from .helpers import location_validator, archivename_validator, ChunkerParams, CompressionSpec
-from .helpers import PrefixSpec, sort_by_spec, HUMAN_SORT_KEYS
+from .helpers import prefix_spec, sort_by_spec, HUMAN_SORT_KEYS
```
ThomasWaldmann: hmm, we have also CompressionSpec (and ChunkerParams), so CamelCase has matched that.

src/borg/archiver.py
```diff
@@ -2285,12 +2285,16 @@ def build_parser(self, prog=None):
 
     @staticmethod
     def add_archives_filters_args(subparser):
+        filters_group = subparser.add_argument_group('filters', 'Archive filters can be applied on repository targets.')
```
ThomasWaldmann: applied **to**?

src/borg/archiver.py
```diff
@@ -2285,12 +2285,16 @@ def build_parser(self, prog=None):
...
+        filters_group.add_argument('-P', '--prefix', dest='prefix', type=prefix_spec, default='',
+                               help='only consider archive names starting with this prefix')
```
ThomasWaldmann: at the other places (old code), there is no default, why do you have one here?

please note that a empty string prefix matches all archives, which might be dangerous if the operation is "delete".

src/borg/archiver.py
```diff
@@ -204,18 +204,55 @@ def do_check(self, args, repository):
...
+                if args.location.archive:
+                    archive_info = manifest.archives.get(args.location.archive)
+                    if archive_info is None:
+                        logger.error('Archive %s not found.' % args.location.archive)
+                        return EXIT_ERROR
+                    archives = (archive_info,)
+                elif any((args.first, args.last, args.prefix)):
+                    archives = self._get_filtered_archives(args, manifest)
+                else:
+                    archives = manifest.archives.list(prefix=args.prefix)
```
ThomasWaldmann: that looks strange.
232 will trigger if args.prefix is truish.
but in 235 you give args.prefix as a param (can it be anything else than None or '' here?).

src/borg/archiver.py
funkyfuture: but a use with `--prefix` only would then also require:

```
if not n:
    n = len(archives)
```

src/borg/testsuite/archiver.py
funkyfuture: the assertion is that the command will succeed with exit code 0.

src/borg/testsuite/archiver.py
funkyfuture: the tests for bad filters wouln't work as argparse raises `SystemExit` which would abort pytest.

src/borg/archiver.py
ThomasWaldmann: that would make elif args.last: ... else: ...

src/borg/testsuite/archiver.py
ThomasWaldmann: ah, ok.

src/borg/testsuite/archiver.py
ThomasWaldmann: ok, maybe put just that into a src comment there?

enkore: Hmm, since that only applies to in-source (vs. binary) testing this change should rather go into [exec_cmd](https://github.com/borgbackup/borg/blob/master/src/borg/testsuite/archiver.py#L61)

src/borg/archiver.py
funkyfuture: sure, but uses seven lines instead of one. is simple not better than complex? imo the oneliner is also obvious.

src/borg/helpers.py
funkyfuture: what exactly are you refreing to here?

src/borg/testsuite/archiver.py
funkyfuture: @enkore okay, but it is in that function.

src/borg/archiver.py
funkyfuture: no pep8 compliance?

src/borg/archiver.py
funkyfuture: with `args.prefix == None` [this list comprehension](https://github.com/borgbackup/borg/pull/1554/files#diff-d3eb25b2d47fdf880fd23aae789dd9efR150) would break.

no, it isn't dangerous to use delete without --prefix argument, that is actually what you want when you use one of the three filter arguments. the 'filter code' wouldn't be invoked is only invoked when at least one of these is used.
atm i can't think of an undesired outcome. do you have an example?

src/borg/testsuite/archiver.py
enkore: How embarrassing :) I somehow thought this was main()

funkyfuture:
ai'ght, thanks for the feedback. i changed the commits where the solution was clear to me.

src/borg/archiver.py
ThomasWaldmann: it's seven instead of three.

it's a little less magic, your one-liner relies on short-circuit evaluation.

anyway, take what you like, it was just an option.

src/borg/helpers.py
ThomasWaldmann: i meant if you change something, do the change in the commit where it belongs.
this can be done by first doing a small fixup commit and later merging that small fixup commit into the commit where the to-be-fixed code was introduced.

src/borg/archiver.py
ThomasWaldmann: right, it is somehow consistency (how a validator name looks) vs. pep8 (different style for funcs and classes).

src/borg/archiver.py
```diff
@@ -2281,6 +2368,23 @@ def build_parser(self, prog=None):
...
+        sort_by_default = 'timestamp'
+        filters_group.add_argument('--sort-by', dest='sort_by', type=sort_by_spec, default=sort_by_default,
+                               help='Comma-separated list of sorting keys; valid keys are: {}; default is: {}'
+                               .format(', '.join(HUMAN_SORT_KEYS), sort_by_default))
+
+        group = filters_group.add_mutually_exclusive_group()
+        group.add_argument('--first', dest='first', metavar='N', default=0, type=int,
+                           help='select first N archives')
+        group.add_argument('--last', dest='last', metavar='N', default=0, type=int,
+                           help='delete last N archives')
```
ThomasWaldmann: you call this method from list, delete, check, ... right?
so the help texts are still wrong.
i suggested the wording "consider first/last N archives" in a previous review.

src/borg/archiver.py
```diff
@@ -2281,6 +2368,23 @@ def build_parser(self, prog=None):
                                help='hex object ID(s) to delete from the repo')
         return parser
 
+    @staticmethod
+    def add_archives_filters_args(subparser):
+        filters_group = subparser.add_argument_group('filters', 'Archive filters can be applied to repository targets.')
+        filters_group.add_argument('-P', '--prefix', dest='prefix', type=prefix_spec, default='',
+                                   help='only consider archive names starting with this prefix')
```
ThomasWaldmann: what happens with `borg delete --prefix repo` ?

funkyfuture: if `BORG_REPOSITORY` is set, it would filter all archives with the prefix `repo`, otherwise it would complain about a missing location arg.

ThomasWaldmann: OK, that was a bad example. ;)

src/borg/archiver.py
ThomasWaldmann: ok

funkyfuture:
okay, i did a fix commit concerning the remaining issues. to rebase it is kind of a pita.

ThomasWaldmann:
any idea how to review d61c778869b9f336cb301682b0e01cf5f9f61786 ?

funkyfuture:
maybe with two screens, reading both revisions side-by-side rather than with a diff view

ThomasWaldmann:
Sorry, I can't review https://github.com/borgbackup/borg/commit/d61c778869b9f336cb301682b0e01cf5f9f61786 .

It somehow looks better than the old code, but I can not see whether it still does the same thing and works correctly (except what is covered by the still succeeding unit tests). At the same time, this is critical code with the capability to either not repair your repo or make stuff even worse, if things go wrong.

So, anybody else wants to review this?

Or, remove it from this PR to get the rest adopted more quickly?

In general, I would advise doing less complex/big changesets and especially hold back from refactoring / cleaning up and introducing new features within the same changeset (or even PR) - it just messes up the limited diff view on github and makes things hard to review.

funkyfuture:
since you changed the ArchiveChecker and i was really not keen to rebase onto that, i removed the commit in question.

concerning the review, i have no other idea than (more?!) tests and reading.

enkore:
Can you rebase this on current master? Thanks!

src/borg/archiver.py
```diff
@@ -2290,6 +2344,23 @@ def build_parser(self, prog=None):
...
+
+        sort_by_default = 'timestamp'
+        filters_group.add_argument('--sort-by', dest='sort_by', type=SortBySpec, default=sort_by_default,
+                                   help='Comma-separated list of sorting keys; valid keys are: {}; default is: {}'
+                                   .format(', '.join(HUMAN_SORT_KEYS), sort_by_default))
+
+        group = filters_group.add_mutually_exclusive_group()
+        group.add_argument('--first', dest='first', metavar='N', default=0, type=int,
+                           help='consider first N archives after other filter args were applied')
```
ThomasWaldmann: nitpick: "after other filters were applied"

src/borg/archiver.py
```diff
@@ -2290,6 +2344,23 @@ def build_parser(self, prog=None):
...
+        group.add_argument('--last', dest='last', metavar='N', default=0, type=int,
+                           help='consider last N archives after other filter args were applied')
```

src/borg/helpers.py
```diff
@@ -142,11 +142,15 @@ def __delitem__(self, name):
         name = safe_encode(name)
         del self._archives[name]
 
-    def list(self, sort_by=None, reverse=False):
-        # inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
-        archives = self.values()  # [self[name] for name in self]
+    def list(self, sort_by=None, reverse=False, prefix=''):
+        """ Inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
+            Returns list of borg.helpers.ArchiveInfo instances
+        """
```
ThomasWaldmann: docstring style:
"""
like
this
"""
or
"""like this"""

src/borg/archiver.py
```diff
@@ -2352,6 +2423,25 @@ def run(self, args):
...
 
+    def _get_filtered_archives(self, args, manifest):
+        if args.location.archive:
+            raise Error('The options --first, --last and --prefix can only be used on repository targets.')
+
+        archives = manifest.archives.list(prefix=args.prefix)
+        if not archives:
+            logger.critical('There are no archives.')
+            self.exit_code = self.exit_code or EXIT_WARNING
+            return []
```
ThomasWaldmann: please remove the complete if-block, it is rather harmful than doing anything good.
if archives is empty here, the rest of the code here will still work normally and return a empty list at the end, no special casing needed.

ThomasWaldmann:
@funkyfuture if we can finish this quickly, it can go into 1.1.0b2 still, release is tomorrow.

funkyfuture:
i'll try to get it done this night.

ThomasWaldmann:
Cool, thanks.

funkyfuture:
can you give me a hint what is suddenly causing this:

```
user@host:~/python/borg (archives_filters) $ borg init --encryption=none /tmp/test
self test test_iteritems (borg.testsuite.hashindex.HashIndexTestCase) FAILED:
  File "/home/user/python/borg/src/borg/testsuite/hashindex.py", line 91, in test_iteritems
    self.assert_raises(StopIteration, next, iterator)
  File "/usr/lib/python3.5/site-packages/_pytest/python.py", line 1202, in raises
    pytest.fail(message)
  File "/usr/lib/python3.5/site-packages/_pytest/runner.py", line 537, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: DID NOT RAISE <class 'StopIteration'>

self test failed
This is a bug either in Borg or in the package / distribution you use.
```

enkore:
Recompile the extensions (`python setup.py develop` or `pip install -e .`); it seems we forgot to bumb the API_VERSION there.

funkyfuture:
i incorporated the demanded changes, but did not squash these as the effort and risk to break things are too high.

ThomasWaldmann:
./src/borg/testsuite/archiver.py:1830:5: E303 too many blank lines (2)

ThomasWaldmann:
lgtm modulo test results.

funkyfuture:
got it.

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/1663

PR #1567, opened 2016-09-03
contribution guidelines

ThomasWaldmann:


docs/development.rst
```diff
@@ -10,6 +10,40 @@ This chapter will get you started with |project_name| development.
...
+- choose the branch you base your changesets on wisely:
+
+  - choose x.y-maint for stuff that should go into next x.y release
+    (it usually gets merged into master branch later also)
+  - choose master if that does not apply
+
+- do clean changesets:
+
+  - focus on some topic, resist changing anything else.
+  - do not do style changes mixed with functional changes.
```
ThomasWaldmann: add same for refactorings? like:

- avoid not required refactorings mixed with functional changes

textshell: * try to avoid refactorings mixed with functional changes
maybe?

docs/development.rst
```diff
@@ -10,6 +10,40 @@ This chapter will get you started with |project_name| development.
...
+  - if you need to fix something after commit, do a fixup commit you can
+    merge into the bad commit later.
```
textshell: I’m torn here. In longer commit series it is often good to edit commits instead of piling on fixups. Also when the last review rounds is a bit ago. But we can‘t really capture that in a guideline^^

ThomasWaldmann: if no review is ongoing, editing the commit and fixing it directly is ofc also ok.
what i primarily meant here is not to merge the fix into some arbitrary unrelated commit.

docs/development.rst
ThomasWaldmann: ok

docs/development.rst
textshell: If that’s the important point (i agree this is important) then maybe we can use wording that is more explicit?

I don’t have a good wording at the moment though. Should we recommend to use git "fixup! <prefix of orignial commit>" style commits? (it would be great if github would understand those).

ThomasWaldmann:
collapse / merge?

enkore:
lgtm

docs/development.rst
```diff
@@ -10,6 +10,46 @@ This chapter will get you started with |project_name| development.
...
+    - if there are no ongoing reviews or you did not push the
+      bad commit yet: edit the commit to include your fix or
+      merge the fixup commit before pushing.
+  - have a nice, clear, typo-free commit comment
+  - if you fixed an issue, refer to it in your commit comment
+  - follow the style guide (see below)
+
+- if you write new code, please add tests and docs for it
+
+- run the tests, fix anything that comes up
```
textshell: I wonder if it would be useful to add some tips how to run the tests. Contributors might not know tox and py.test. Also a common way to setup the virtualenv (as subdir of the borg checkout) yields false flake8 warnings when running tox.

ThomasWaldmann: iirc that is documented just some paragraphs below that.

textshell: indeed. Sorry for the noise.

i think collapse + merge is good.

PR #1569, opened 2016-09-03
add a .coafile for coala

ThomasWaldmann:
ignores / disable are set up so that there are not many faults.
we can improve that iteratively.

Note: I made this while playing with coala.

Besides pep8 and pylint it adds some simple generic checkers - and they found some stuff in our source.

Besides finding and complaining about stuff, coala can also assist fixing the stuff: manually, via automated patch, via invoking an editor.

Also, it is somehow nice to see one frontend for all sorts of code analysis develop.


enkore:
0e33294b has a pylintrc that works well (PyLintBear has a pylint_rcfile parameter)

ThomasWaldmann:
OK, this has a stale label now.

Things being stale here:
- we need to decide whether we want coala or not (I'm for it, but I can live without it, too).
- if we go for it, it needs to be rebased on / adapted for master branch, should not go into 1.0-maint.

enkore:
https://www.youtube.com/watch?v=CxK_nA2iVXw

ThomasWaldmann:
see #2592.

PR #1581, opened 2016-09-07
hashindex: export max load factor to Python-space

### LINKED ISSUE #1580: make hashtable max load available to python
_hashindex.c: #define HASH_MAX_LOAD .75

We need that in Python code also, to compute a big enough hashtable size so it does not need to grow:

archive.py: capacity = int(len(self.repository) * 1.35 + 1) # > len * 1.0 / HASH_MAX_LOAD (see _hashindex.c)

So, make this constant available to python code via hashindex.pyx.

enkore:
Fixes #1580

### DIFF
--- a/borg/testsuite/hashindex.py
+++ b/borg/testsuite/hashindex.py
@@ -276,3 +276,8 @@ def test_nsindex_segment_limit():
     assert H(1) not in idx
     idx[H(2)] = hashindex.MAX_VALUE, 0
     assert H(2) in idx
+
+
+def test_max_load_factor():
+    assert NSIndex.MAX_LOAD_FACTOR < 1
+    assert ChunkIndex.MAX_LOAD_FACTOR < 1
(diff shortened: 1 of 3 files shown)


borg/testsuite/hashindex.py
```diff
@@ -266,13 +266,18 @@ def test_read_known_good(self):
...
+        idx[H(1)] = NSIndex.MAX_VALUE + 1, 0
     assert H(1) not in idx
-    idx[H(2)] = hashindex.MAX_VALUE, 0
+    idx[H(2)] = NSIndex.MAX_VALUE, 0
     assert H(2) in idx
+
+
+def test_max_load_factor():
+    assert NSIndex.MAX_LOAD_FACTOR < 1
+    assert ChunkIndex.MAX_LOAD_FACTOR < 1
```
ThomasWaldmann: < 1.0 to make clear it is float?

ThomasWaldmann:
you fix the one place also that started this discussion (see ticket)?

enkore:
sure

borg/testsuite/hashindex.py
ThomasWaldmann: hmm, there is a comment near the load factor definition like "never go above 0.75" - maybe test for that here?

ThomasWaldmann:
lgtm, except the 2 items i already pointed out.

ThomasWaldmann:
ok, lgtm.

PR #1586, opened 2016-09-09
borg check: delete chunks with integrity errors, fixes #1575

### LINKED ISSUE #1575: let check --verify-data remove defect objects?
In `--repair` mode, `borg check --verify-data` could remove objects with IntegrityErrors.

In files using these objects, they would then get replaced by all-zero dummy objects and marked as "repaired" files. Which could later get "healed" if the same chunk gets stored again and borg check is run.

So users would not have to use borg debug commands to manually remove them.

ThomasWaldmann:
so they can be "repaired" immediately and maybe healed later.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1035,6 +1035,7 @@ class ArchiveChecker:
         logger.info('Starting cryptographic data integrity verification...')
         count = len(self.chunks)
         errors = 0
+        defect_chunks = []
         pi = ProgressIndicatorPercent(total=count, msg="Verifying data %6.2f%%", step=0.01)
         for chunk_id, (refcount, *_) in self.chunks.iteritems():
             pi.show()
@@ -1052,7 +1053,39 @@ class ArchiveChecker:
                 self.error_found = True
                 errors += 1
                 logger.error('chunk %s, integrity error: %s', bin_to_hex(chunk_id), integrity_error)
+                defect_chunks.append(chunk_id)
         pi.finish()
+        if defect_chunks:
+            if self.repair:
+                # if we kill the defect chunk here, subsequent actions within this "borg check"
+                # run will find missing chunks and replace them with all-zero replacement
+                # chunks and flag the files as "repaired".
+                # if another backup is done later and the missing chunks get backupped again,
+                # a "borg check" afterwards can heal all files where this chunk was missing.
+                logger.warning('Found defect chunks. They will be deleted now, so affected files can '
+                               'get repaired now and maybe healed later.')
+                for defect_chunk in defect_chunks:
+                    # remote repo (ssh): retry might help for strange network / NIC / RAM errors
+                    # as the chunk will be retransmitted from remote server.
+                    # local repo (fs): as chunks.iteritems loop usually pumps a lot of data through,
+                    # a defect chunk is likely not in the fs cache any more and really gets re-read
+                    # from the underlying media.
+                    encrypted_data = self.repository.get(defect_chunk)
+                    try:
+                        _chunk_id = None if defect_chunk == Manifest.MANIFEST_ID else defect_chunk
+                        self.key.decrypt(_chunk_id, encrypted_data)
+                    except IntegrityError:
+                        # failed twice -> get rid of this chunk
+                        del self.chunks[defect_chunk]
+                        self.repository.delete(defect_chunk)
+                        logger.debug('chunk %s deleted.', bin_to_hex(defect_chunk))
+                    else:
+                        logger.warning('chunk %s not deleted, did not consistently fail.')
+            else:
+                logger.warning('Found defect chunks. With --repair, they would get deleted, so affected '
+                               'files could get repaired then and maybe healed later.')
+                for defect_chunk in defect_chunks:
+                    logger.debug('chunk %s is defect.', bin_to_hex(defect_chunk))
         log = logger.error if errors else logger.info
         log('Finished cryptographic data integrity verification, verified %d chunks with %d integrity errors.', count, errors)
 
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -1956,6 +1956,10 @@ class ArchiverCheckTestCase(ArchiverTestCaseBase):
         self.cmd('check', self.repository_location, exit_code=0)
         output = self.cmd('check', '--verify-data', self.repository_location, exit_code=1)
         assert bin_to_hex(chunk.id) + ', integrity error' in output
+        # repair (heal is tested in another test)
+        output = self.cmd('check', '--repair', '--verify-data', self.repository_location, exit_code=0)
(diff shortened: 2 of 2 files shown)


src/borg/archive.py
```diff
@@ -1052,6 +1052,19 @@ def verify_data(self):
...
                 logger.error('chunk %s, integrity error: %s', bin_to_hex(chunk_id), integrity_error)
+                if self.repair:
+                    # if we kill the defect chunk here, subsequent actions within this "borg check"
+                    # run will find missing chunks and replace them with all-zero replacement
+                    # chunks and flag the files as "repaired".
+                    # if another backup is done later and the missing chunks get backupped again,
+                    # a "borg check" afterwards can heal all files where this chunk was missing.
+                    logger.warning('chunk %s is being deleted, so affected files can get repaired now '
+                                   'and maybe healed later.', bin_to_hex(chunk_id))
+                    del self.chunks[chunk_id]
```
enkore: this is unsafe [modify-collection-that-is-iterated-over], here specifically because self.chunks is the usual hashindex and the iterator is based on the internal modulus index. If the hashtable is resized we may skip stuff.

Solution: Add list of deleted chunk ids, remove them from self.chunks after the loop is done.

ThomasWaldmann: right, will fix that.

src/borg/archive.py
```diff
@@ -1052,6 +1052,19 @@ def verify_data(self):
...
+                    self.repository.delete(chunk_id)
+                else:
+                    logger.warning('chunk %s would be deleted, so affected files can get repaired now '
+                                   'and maybe healed later.', bin_to_hex(chunk_id))
```
enkore: This message is kinda confusing. `--repair would delete the corrupted chunk, so affected files could be repaired (and may be healed later)` or so. Both of these warnings should only be printed once, since there already is a per-chunk message and they just clarify what will happen / would happen.

ThomasWaldmann: don't we have that kind of msgs all over the place?

ThomasWaldmann:
Another thing we might consider:

Imagine, there is some hardware that (rarely) corrupts data in transit (e.g. a USB3 hub, I already had that in practice). So, we'll see an IntegrityError and would kill the chunk although the chunk might be ok on disk and was just corrupted while travelling over USB.

So I think we should do 1 or 2 retries before killing the objects.

Comments?

ThomasWaldmann:
last changeset implements that idea. if ok, I'll dedup some common code.

src/borg/archive.py
```diff
@@ -1065,9 +1065,25 @@ def verify_data(self):
...
+                        try:
+                            _chunk_id = None if defect_chunk == Manifest.MANIFEST_ID else defect_chunk
+                            self.key.decrypt(_chunk_id, encrypted_data)
+                        except IntegrityError:
+                            retries_failed += 1
+                    if retries_failed == DEFECT_CHUNK_RETRIES:
+                        # all retries failed, get rid of this chunk
+                        del self.chunks[defect_chunk]
+                        self.repository.delete(defect_chunk)
+                        logger.warning('chunk %s deleted.', bin_to_hex(defect_chunk))
```
enkore: two thoughts

- this first message is redundant (After the "They will be deleted now" msg "xyz deleted" is no news/expected)
- since this may take a while (consider broken sector on a hard disk, often these are very slow to read because the drive retries internally; or over a network connection) a ProgressIndicatorPercent would make sense

enkore:
> So I think we should do 1 or 2 retries before killing the objects.

I like the idea, if implemented, it should also be added to the repository-side repair_segment code (separate ticket/PR); otherwise it's a bit pointles.

src/borg/archive.py
ThomasWaldmann: the per-chunk log output is mainly to log the chunkid.
the first longer msg is to explain what's happening and why we do that.

the 1s delay only happens for really broken chunks, N-1 retries before have no delay, so it doesn't delay if e.g. the 2nd attemt is successful.

but yes, if there are a lot of errors, a PI might be nice.

ThomasWaldmann:
You mean LoggedIO.recover_segment?

We also have to consider we have different places where corruption can occur: on storage medium (or in general: on levels below repo code) and somewhere above repo code (e.g. for remote repos, somewhere in network card of client or in RAM).

enkore:
Another Q is whether the retry-loop is actually meaningful.

If we read the same object a few times in a row it would be most likely copied from the page cache instead of being read from the drive again (- and even then, if the drive's FEC failed silently before, what are the odds it then starts working again? Also, would this be a condition were we want to leave the chunk where it is?).

So to actually force a re-transfer from the media to help in the situations where it's beneficial (wonky USB cables, everyone loves them, and also in-memory corruption (uh-oh)) we would also need to add some cache-eviction logic to Repository. Which I don't particularly fancy (objects are unaligned etc).

For corruption on the network (which shouldn't be possible *at all* since it's a SSH pipe) it works like it's now.

src/borg/archive.py
enkore: IMHO the usual case is the "really broken chunk", not chunks corrupted by a bad cable or so. Therefore I think it'd be best to log only the exceptional case (chunk recovered, won't be deleted), not the announced one (chunk deleted). Logging the chunk ID isn't necessary, since they were all logged as errors already.

ThomasWaldmann:
Maybe I should split that retry logic stuff into a separate PR.
For fs reads, it could fadvise DONTNEED the data, but that needs more changes elsewhere.

src/borg/archive.py
ThomasWaldmann: OK.

ThomasWaldmann:
Hmm, as the first scanning loop goes over all chunks (lots of data, likely many GBs or even TBs) is it somehow likely that an bad chunk discovered there is not in the cache any more when it is handled in the second loop. Also, there is usually some time in between due to the amount of data.

So, for many cases the caching problem is solved if I just do a single retry in the second (over defects list) loop without requiring any other changes in other parts of the code.

ThomasWaldmann:
The latest push shall fix the tests. Collapse / Merge?

src/borg/archive.py
```diff
@@ -1052,7 +1053,39 @@ def verify_data(self):
...
+                               'get repaired now and maybe healed later.')
+                for defect_chunk in defect_chunks:
+                    # remote repo (ssh): retry might help for strange network / NIC / RAM errors
+                    # as the chunk will be retransmitted from remote server.
+                    # local repo (fs): as chunks.iteritems loop usually pumps a lot of data through,
+                    # a defect chunk is likely not in the fs cache any more and really gets re-read
+                    # from the underlying media.
+                    encrypted_data = self.repository.get(defect_chunk)
+                    try:
+                        _chunk_id = None if defect_chunk == Manifest.MANIFEST_ID else defect_chunk
```
enkore: ah, good catch

enkore:
Yeah, lgtm.

PR #1594, opened 2016-09-13
Handle SIGHUP for proper unlock on lost connection, fixes #1593

### LINKED ISSUE #1593: Does not catch SIGHUP
borg does not catch SIGHUP, which means that the repository might remain locked if the SSH connection between the server and the client is lost in a remote operation. This seems to manifest especially with systemd-logind having KillUserProcesses set to yes, for unknown reasons.

julian-klode:
If the connections hangs up, the borg server needs to clean
up, especially unlock the repository, so a later try will
work again.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1666,6 +1666,14 @@ def sig_term_handler(signum, stack):
     raise SIGTERMReceived
 
 
+class SIGHUPReceived(BaseException):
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
lgtm.

enkore:
Thanks!

PR #1601, opened 2016-09-15
WIP: Backport borg check verify data

ThomasWaldmann:
changesets to do:
- deletion of defect chunks
- speed tuning


ThomasWaldmann:
hmm, I'm getting doubts whether --verify-data should get backported.
otoh, it would be very useful to verify data integrity, but otoh implementing it in a fast way requires quite some changes (Repository.scan(), changes to LoggedIO to start at specific segment/offset). Maybe a bit too much for 1.0-maint.

enkore:
=> #1574

PR #1603, opened 2016-09-15
Fix FUSE hardlinks

ThomasWaldmann:


ThomasWaldmann:
merge as is?

src/borg/fuse.py
```diff
@@ -43,7 +43,9 @@ def add(self, item):
         return pos + self.offset
 
     def get(self, inode):
-        self.fd.seek(inode - self.offset, io.SEEK_SET)
+        offset = inode - self.offset
+        assert offset >= 0, 'ItemCache.get() called with a invalid inode number'
```
enkore: Hm, better use an if...:ValueError?

src/borg/fuse.py
```diff
@@ -133,46 +135,66 @@ def _create_dir(self, parent):
...
         """Build fuse inode hierarchy from archive metadata
         """
+        self.file_versions = {}  # for versions mode: original path -> version
         unpacker = msgpack.Unpacker()
         for key, chunk in zip(archive.metadata.items, self.repository.get_many(archive.metadata.items)):
             _, data = self.key.decrypt(key, chunk)
             unpacker.feed(data)
             for item in unpacker:
                 item = Item(internal_dict=item)
+                path = os.fsencode(os.path.normpath(item.path))
```
enkore: I see that this came from the old code, but is normpath actually necessary? We normalize paths in archives.

ThomasWaldmann: maybe not, but i think this should be done seperately from this changeset (and needs a global + old code review).

src/borg/fuse.py
ThomasWaldmann: ok

src/borg/fuse.py
enkore: I'm ok with that :)

PR #1604, opened 2016-09-15
Add --remote-ratelimit

textshell:
The --remote-ratelimit option adds a very simple rate limit for the
sending data to the remote.

Currently implemented by sleeping if the transmission speed is greater
than the limit.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1322,6 +1322,8 @@ class Archiver:
                                   help='set umask to M (local and remote, default: %(default)04o)')
         common_group.add_argument('--remote-path', dest='remote_path', metavar='PATH',
                                   help='set remote path to executable (default: "borg")')
+        common_group.add_argument('--remote-ratelimit', dest='remote_ratelimit', type=int, metavar='rate',
+                                  help='set remote network upload rate limit in kiByte/s (default: 0=unlimited)')
         common_group.add_argument('--consider-part-files', dest='consider_part_files',
                                   action='store_true', default=False,
                                   help='treat part files like normal files (e.g. to list/extract them)')
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -6,6 +6,7 @@ import select
 import shlex
 import sys
 import tempfile
+import time
 import traceback
 from subprocess import Popen, PIPE
 
@@ -24,6 +25,8 @@ RPC_PROTOCOL_VERSION = 2
 BUFSIZE = 10 * 1024 * 1024
 
 MAX_INFLIGHT = 100
+
+RATELIMIT_PERIOD = 0.1
 
 
 class ConnectionClosed(Error):
@@ -166,6 +169,36 @@ class RepositoryServer:  # pragma: no cover
         return self.repository.id
 
 
+class SleepingBandwidthLimiter:
+    def __init__(self, limit):
+        if limit:
+            self.ratelimit = int(limit * RATELIMIT_PERIOD)
+            self.ratelimit_last = time.monotonic()
+            self.ratelimit_quota = self.ratelimit
+        else:
+            self.ratelimit = None
+
+    def write(self, fd, to_send):
+        if self.ratelimit:
+            now = time.monotonic()
+            if self.ratelimit_last + RATELIMIT_PERIOD <= now:
+                self.ratelimit_quota += self.ratelimit
+                if self.ratelimit_quota > 2 * self.ratelimit:
+                    self.ratelimit_quota = 2 * self.ratelimit
+                self.ratelimit_last = now
+            if self.ratelimit_quota == 0:
+                tosleep = self.ratelimit_last + RATELIMIT_PERIOD - now
+                time.sleep(tosleep)
+                self.ratelimit_quota += self.ratelimit
+                self.ratelimit_last = time.monotonic()
+            if len(to_send) > self.ratelimit_quota:
+                to_send = to_send[:self.ratelimit_quota]
+        written = os.write(fd, to_send)
+        if self.ratelimit:
+            self.ratelimit_quota -= written
+        return written
+
+
 class RemoteRepository:
     extra_test_args = []
 
@@ -185,6 +218,8 @@ class RemoteRepository:
         self.cache = {}
         self.ignore_responses = set()
         self.responses = {}
+        self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
+
         self.unpacker = msgpack.Unpacker(use_list=False)
         self.p = None
         testing = location.host == '__testsuite__'
@@ -406,7 +441,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
 
                 if self.to_send:
                     try:
-                        self.to_send = self.to_send[os.write(self.stdin_fd, self.to_send):]
+                        self.to_send = self.to_send[self.ratelimit.write(self.stdin_fd, self.to_send):]
                     except OSError as e:
                         # io.write might raise EAGAIN even though select indicates
                         # that the fd should be writable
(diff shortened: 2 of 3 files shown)


src/borg/remote.py
```diff
@@ -306,6 +316,26 @@ def ssh_cmd(self, location):
             args.append('%s' % location.host)
         return args
 
+    def write_bandwith_limited(self, to_send):
+        if self.ratelimit:
+            current = time.monotonic()
+            if self.ratelimit_last + RATELIMIT_PERIOD <= current:
+                self.ratelimit_quota += self.ratelimit
+                if self.ratelimit_quota > 2 * self.ratelimit:
+                    self.ratelimit_quota = 2 * self.ratelimit
```
ThomasWaldmann: instead of adding up and then limiting, you could also conditionally add up.
```
if q < 2 * lim:
    q += lim
```
it goes up to almost 3 that way, but guess that doesn't matter, right?

src/borg/remote.py
```diff
@@ -306,6 +316,26 @@ def ssh_cmd(self, location):
             args.append('%s' % location.host)
         return args
 
+    def write_bandwith_limited(self, to_send):
+        if self.ratelimit:
+            current = time.monotonic()
```
textshell: via release notes and linked bugtracker:
python 3.3 Availability: Windows, Mac OS X, Linux, FreeBSD, OpenBSD, Solaris.

Do we care about something else? i.e. does this need better error handling?

src/borg/remote.py
textshell: i like the explicit way. No need to save lines here in my opinion.

src/borg/remote.py
```diff
@@ -306,6 +316,26 @@ def ssh_cmd(self, location):
...
+                self.ratelimit_last = current
+            if self.ratelimit_quota == 0:
+                tosleep = self.ratelimit_last + RATELIMIT_PERIOD - current
+                time.sleep(tosleep)
+                self.ratelimit_quota += self.ratelimit
+                self.ratelimit_last = time.monotonic()
+            if len(to_send) > self.ratelimit_quota:
+                to_send = to_send[:self.ratelimit_quota]
+        written = os.write(self.stdin_fd, to_send)
+        if written > 0 and self.ratelimit:
```
textshell: Yes, i guess python handles are errors via exceptions, i was thinking to much like raw C.

src/borg/remote.py
```diff
@@ -306,6 +316,26 @@ def ssh_cmd(self, location):
...
+    def write_bandwith_limited(self, to_send):
+        if self.ratelimit:
+            current = time.monotonic()
+            if self.ratelimit_last + RATELIMIT_PERIOD <= current:
+                self.ratelimit_quota += self.ratelimit
+                if self.ratelimit_quota > 2 * self.ratelimit:
+                    self.ratelimit_quota = 2 * self.ratelimit
+                self.ratelimit_last = current
+            if self.ratelimit_quota == 0:
+                tosleep = self.ratelimit_last + RATELIMIT_PERIOD - current
```
textshell: This is essentially:

    if self.ratelimit_last + RATELIMIT_PERIOD <= current:
        ...  # sets self.ratelimit_quota to positive value
    if self.ratelimit_quota == 0:
        tosleep = self.ratelimit_last + RATELIMIT_PERIOD - current

so here tosleep is always > 0

src/borg/archiver.py
```diff
@@ -1322,6 +1322,8 @@ def build_parser(self, prog=None):
                                   help='set umask to M (local and remote, default: %(default)04o)')
         common_group.add_argument('--remote-path', dest='remote_path', metavar='PATH',
                                   help='set remote path to executable (default: "borg")')
+        common_group.add_argument('--remote-ratelimit', dest='remote_ratelimit', type=int, metavar='rate',
+                                  help='set remote network rate limit in kbyte/s (default: unlimited)')
```
enkore: doc: 0 is probably = unlimited?

src/borg/remote.py
ThomasWaldmann: i tried it on cygwin and netbsd and py34, both worked.

src/borg/remote.py
ThomasWaldmann: if current (now) makes a bigger than RATELIMIT_PERIOD step due to slow scheduling, tosleep will be negative.

src/borg/archiver.py
textshell: I was going for None, but 0 works better. I can add that to the doc

src/borg/remote.py
textshell: No, in that case `self.ratelimit_last + RATELIMIT_PERIOD <= current` (the if above) and it sets `self.ratelimit_quota += self.ratelimit` (or something close) so `self.ratelimit_quota == 0` will not be true

Or am i confused somehow?

ThomasWaldmann: ah, so: "elif"?

textshell: i thought about it. But it doesn’t quite feel right. But it doesn‘t make a difference, so i can use elif you like that more.

ThomasWaldmann: as you like, just thought it would make the exclusiveness more obvious.

src/borg/archiver.py
ThomasWaldmann: as you compute with 1024, it is "kiByte/s"

ThomasWaldmann: also, you only limit local -> remote traffic, so it is useful for borg create, but not borg extract.

textshell: yes, i should either document that or implement limiting for the other side too. I tend to just document it for now. Because i believe one of the main use cases is asymmetrical (slow) lines and it’s not too clear what would be what users want here. Also i expect the main use case where the user is not actively waiting for borg to be the create case.

ThomasWaldmann:
lgtm. collapse, merge.

ThomasWaldmann:
hmm, test coverage is lowered.

maybe just add a test that executes the code, so we at least know it does not crash?

textshell:
The way things are currently it’s fairly hard to get a consistent coverage. I could add a rate limit in one of the archiver test cases. But that won’t give complete coverage and risks having different coverage depending on timing. Which is really ugly (noise) in coverage difference reports.

I though about doing a test for RemoteRepository but the constructor does so much interaction with the target borg that it would need more code than i want to write for this to even get out of the constructor to be able to test write_bandwith_limited. Especially if 1.2 might change this whole code around (not sure about that)

enkore:
Since write_ratelimited doesn't seem to have much interaction with RemoteRepository (except the FD) you could pull it out into a separate class, allowing unit tests.

ThomasWaldmann:
the seperate class might be also nice for reusability, if later the other direction shall also be supported.

textshell:
Now with artificial coverage. Does anyone know when codecov updates it‘s stats?

src/borg/testsuite/remote.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # only partial write
+        self.expect_write(5, b"123456")
+        it.write(5, b"1234567890")
+
+        # sleeps
+        self.expect_write(5, b"123456")
+        it.write(5, b"123456")
+
+        # long part withouts writes
```
ThomasWaldmann: hm? do you mean "long time without writes"?

textshell: I guess i meant something. I reworded it to make more sense.

src/borg/testsuite/remote.py
```diff
@@ -44,17 +44,17 @@ def test_write(self, monkeypatch):
...
         it.write(5, b"1")
 
-        # long part withouts writes
+        # long time interval between writes, filling up quota
         now += 10
         self.expect_write(5, b"1")
         it.write(5, b"1")
 
-        # long part withouts writes
+        # long time interval between writes, fillung up quota to clip to maximum
```
ThomasWaldmann: typo.

PR #1611, opened 2016-09-16
Fix inconsistency in FAQ

senden9:
The script in the FAQ is named pv-wrapper. But in the variable export pv-wrapper.sh was used.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -399,7 +399,7 @@ Create a wrapper script:  /usr/local/bin/pv-wrapper  ::
 
 Add BORG_RSH environment variable to use pipeviewer wrapper script with ssh. ::
 
-    export BORG_RSH='/usr/local/bin/pv-wrapper.sh ssh'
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -399,7 +399,7 @@ Create a wrapper script:  /usr/local/bin/pv-wrapper  ::
 
 Add BORG_RSH environment variable to use pipeviewer wrapper script with ssh. ::
 
-    export BORG_RSH='/usr/local/bin/pv-wrapper.sh ssh'
+    export BORG_RSH='/usr/local/bin/pv-wrapper ssh'
```
ThomasWaldmann: thanks, good catch!

could you do a PR against 1.0-maint, so it gets into next 1.0.x release? it will automatically be merged into master from there also.

senden9: Done. See PR 1612.

senden9:
Moved to PR 1612. A pull request against 1.0-maint

PR #1615, opened 2016-09-18
Fix #1598

### LINKED ISSUE #1598: Index object count mismatch. 51251 != 51250
I am currently having 18 borg-repos. Backups from 6 different servers, 3 repos per server. My idea is to rotate them for different weekdays (Mon+Thu, TueSat, Wed+Fri+Sun) in case a repo breaks. My way of protecting against the downside of deduplication. Even with three mirrors of the data I am saving lots of space. The backup scripts does a create, prune, check.

I have been running this for 1-2 weeks now, with the last 4 days or so using the 1.1.0-b1 version. After I started to use the beta version I started to get the Index object count mismatch from the check-jobs. I currently have the issue with 4 of my 18 repos. The interesting fact is that this far, all of them has a mismatch of **one**.

At first I only noticed the issue with one host. I deleted a repo to make sure I started over from scratch, and removed the prune job from the backup script, hoping the prune was the issue. This did not help.

Earlier I was using version 1.0.7 and did not get these issues, so I assume it's something with the new code, though I can't be certain.

I assume this isn't a known issue since I couldn't find anything about it. Any ideas what could be the cause? Anything I could do to help finding the issue? Any other ideas?

One of my simplest repos are backed up like this:
${BORG} create --compression lzma,1 borg@${DESTHOST}:${DESTPATH}${BORGREPO}.borg::{now} /etc /usr/local/bin /root /home
Eg, no weird options at all, and it happened to this one early this morning.

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/37897778-index-object-count-mismatch-51251-51250)

enkore:
- A delete should only be preserved if the object is not in the index (line 507)
- Only one delete should be preserved (`processed_keys`).
- Improved check logging
  - mention found transaction IDs (on debug level)
  - index count mismatch tabulates missing objects

Fixes #1598

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -650,18 +650,25 @@ class Repository:
         try:
             transaction_id = self.get_transaction_id()
             current_index = self.open_index(transaction_id)
-        except Exception:
+            logger.debug('Read committed index of transaction %d', transaction_id)
+        except Exception as exc:
             transaction_id = self.io.get_segments_transaction_id()
             current_index = None
+            logger.debug('Failed to read committed index (%s)', exc)
         if transaction_id is None:
+            logger.debug('No segments transaction found')
             transaction_id = self.get_index_transaction_id()
         if transaction_id is None:
+            logger.debug('No index transaction found, trying latest segment')
             transaction_id = self.io.get_latest_segment()
         if repair:
             self.io.cleanup(transaction_id)
         segments_transaction_id = self.io.get_segments_transaction_id()
+        logger.debug('Segment transaction is    %s', segments_transaction_id)
+        logger.debug('Determined transaction is %s', transaction_id)
         self.prepare_txn(None)  # self.index, self.compact, self.segments all empty now!
         segment_count = sum(1 for _ in self.io.segment_iterator())
+        logger.debug('Found %d segments', segment_count)
         pi = ProgressIndicatorPercent(total=segment_count, msg="Checking segments %3.1f%%", step=0.1)
         for i, (segment, filename) in enumerate(self.io.segment_iterator()):
             pi.show(i)
@@ -683,6 +690,7 @@ class Repository:
             report_error('Adding commit tag to segment {}'.format(transaction_id))
             self.io.segment = transaction_id + 1
             self.io.write_commit()
+        logger.info('Starting repository index check')
         if current_index and not repair:
             # current_index = "as found on disk"
             # self.index = "as rebuilt in-memory from segments"
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -481,8 +481,9 @@ class Repository:
             for tag, key, offset, data in self.io.iter_objects(segment, include_data=True):
                 if tag == TAG_COMMIT:
                     continue
-                in_index = self.index.get(key) == (segment, offset)
-                if tag == TAG_PUT and in_index:
+                in_index = self.index.get(key)
+                is_index_object = in_index == (segment, offset)
+                if tag == TAG_PUT and is_index_object:
                     try:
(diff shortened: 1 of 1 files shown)


src/borg/repository.py
```diff
@@ -687,7 +687,23 @@ def report_error(msg):
...
+                for key, value in self.index.iteritems():
+                    current_value = current_index.get(key, not_found)
+                    if current_value != value:
+                        logger.warning(line_format, bin_to_hex(key), value, current_value)
+                for key, current_value in current_index.iteritems():
+                    if key in self.index:
+                        continue
+                    value = self.index.get(key, not_found)
+                    if current_value != value:
+                        logger.warning(line_format, bin_to_hex(key), value, current_value)
```
ThomasWaldmann: much better :)

src/borg/repository.py
```diff
@@ -650,18 +650,25 @@ def report_error(msg):
         try:
             transaction_id = self.get_transaction_id()
             current_index = self.open_index(transaction_id)
-        except Exception:
+            logger.debug('Read committed index of transaction %d', transaction_id)
+        except Exception as exc:
             transaction_id = self.io.get_segments_transaction_id()
             current_index = None
+            logger.debug('Failed to read committed index (%s), using segments transaction %s', exc, transaction_id)
```
ThomasWaldmann: maybe move this to "else" branch of the following "if"?
you could also use %d then for the not-None transaction_id.

ThomasWaldmann: also maybe invert the if/else.

src/borg/repository.py
```diff
@@ -445,7 +445,10 @@ def compact_segments(self):
             return
         index_transaction_id = self.get_index_transaction_id()
         segments = self.segments
-        unused = []  # list of segments, that are not used anymore
+        # list of segments, that are not used anymore, ie. none of the objects are part of the current repository state
+        unused = []
+        # set of keys that were processed, to avoid duplication, ie. the final decision for these keys was implemented
+        processed_keys = set()
```
ThomasWaldmann: will this consume lots of memory?

src/borg/repository.py
ThomasWaldmann: just to be clear:
transaction_id in 659 can be None, so it will say ... "using segments transaction None", which is obviously not true, it won't use that.

src/borg/repository.py
```diff
@@ -650,18 +652,25 @@ def report_error(msg):
...
             current_index = self.open_index(transaction_id)
-        except Exception:
+            logger.debug('Read committed index of transaction %d', transaction_id)
+        except Exception as exc:
             transaction_id = self.io.get_segments_transaction_id()
             current_index = None
+            logger.debug('Failed to read committed index (%s), using segments transaction %s', exc, transaction_id)
         if transaction_id is None:
             transaction_id = self.get_index_transaction_id()
+            logger.debug('No segments transaction found, using index transaction %s', transaction_id)
```
ThomasWaldmann: same issue here: "using index transaction None"

src/borg/repository.py
enkore: Yeah that's bad.

ThomasWaldmann:
looks like only minor stuff is missing, can you finish it, so I can put it into a 1.1.0b2 soon?

src/borg/repository.py
ThomasWaldmann: did you see my comment above?

enkore: This was removed

ThomasWaldmann: oh, right. didn't notice that...

PR #1616, opened 2016-09-18
Implement key import / export

### LINKED ISSUE #1555: key import/export command
In the docs, we of course recommend having a **separate** backup of the encryption key.

For keyfile mode, the key is easily accessible via the local filesystem (if you know / read about where to look for it) - just copy it somewhere safe.

For repokey mode, the (passphrase protected) key is in the repo config. If your repo is local, just copy the config to somewhere safe. But if your repo is remote and your only access to it is via a restricted "borg serve" ssh command (no ssh shell, scp or sshfs access), you can't easily get a copy of your key.

***

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/37428131-key-import-export-command)

textshell:
We recommed that users backup their keys, this adds simple to use commands
to do so.

Supported formats are the keyfile format used by borg internally and
a special format with by line checksums for printed backups.

For this format the import is an interactive process which checks
each line as soon as it is input.

Fixes #1555

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -30,6 +30,7 @@ from .upgrader import AtticRepositoryUpgrader, BorgRepositoryUpgrader
 from .repository import Repository
 from .cache import Cache
 from .key import key_creator, RepoKey, PassphraseKey
+from .keymanager import KeyManager
 from .archive import backup_io, BackupOSError, Archive, ArchiveChecker, CHUNKER_PARAMS, is_special
 from .remote import RepositoryServer, RemoteRepository, cache_if_remote
 
@@ -157,6 +158,39 @@ class Archiver:
     def do_change_passphrase(self, args, repository, manifest, key):
         """Change repository key file passphrase"""
         key.change_passphrase()
+        return EXIT_SUCCESS
+
+    @with_repository(lock=False, exclusive=False, manifest=False, cache=False)
+    def do_key_export(self, args, repository):
+        """Export the repository key for backup"""
+        manager = KeyManager(repository)
+        manager.load_keyblob()
+        if args.paper:
+            manager.export_paperkey(args.path)
+        else:
+            if not args.path:
+                self.print_error("output file to export key to expected")
+                return EXIT_ERROR
+            manager.export(args.path)
+        return EXIT_SUCCESS
+
+    @with_repository(lock=False, exclusive=False, manifest=False, cache=False)
+    def do_key_import(self, args, repository):
+        """Import the repository key from backup"""
+        manager = KeyManager(repository)
+        if args.paper:
+            if args.path:
+                self.print_error("with --paper import from file is not supported")
+                return EXIT_ERROR
+            manager.import_paperkey(args)
+        else:
+            if not args.path:
+                self.print_error("input file to import key from expected")
+                return EXIT_ERROR
+            if not os.path.exists(args.path):
+                self.print_error("input file does not exist: " + args.path)
+                return EXIT_ERROR
+            manager.import_keyfile(args)
         return EXIT_SUCCESS
 
     @with_repository(manifest=False)
@@ -1075,6 +1109,34 @@ class Archiver:
         subparser.set_defaults(func=self.do_change_passphrase)
         subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
                                type=location_validator(archive=False))
+
+        subparser = subparsers.add_parser('key-export', parents=[common_parser],
+                                          description=self.do_key_export.__doc__,
+                                          epilog="",
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='export repository key for backup')
+        subparser.set_defaults(func=self.do_key_export)
+        subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
+                               type=location_validator(archive=False))
+        subparser.add_argument('path', metavar='PATH', nargs='?', type=str,
+                               help='where to store the backup')
+        subparser.add_argument('--paper', dest='paper', action='store_true',
+                               default=False,
+                               help='Create an export suitable for printing and later type-in')
+
+        subparser = subparsers.add_parser('key-import', parents=[common_parser],
+                                          description=self.do_key_import.__doc__,
+                                          epilog="",
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='import repository key from backup')
+        subparser.set_defaults(func=self.do_key_import)
+        subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
+                               type=location_validator(archive=False))
+        subparser.add_argument('path', metavar='PATH', nargs='?', type=str,
+                               help='path to the backup')
+        subparser.add_argument('--paper', dest='paper', action='store_true',
+                               default=False,
+                               help='interactively import from a backup done with --paper')
 
         migrate_to_repokey_epilog = textwrap.dedent("""
         This command migrates a repository from passphrase mode (not supported any
--- a/borg/keymanager.py
+++ b/borg/keymanager.py
@@ -1,0 +1,213 @@
+from binascii import hexlify, unhexlify, a2b_base64, b2a_base64
+import binascii
+import textwrap
+from hashlib import sha256
+
+from .key import KeyfileKey, RepoKey, PassphraseKey, KeyfileNotFoundError, PlaintextKey
+from .helpers import Manifest, NoManifestError, Error, yes
+from .repository import Repository
+
+
+class UnencryptedRepo(Error):
+    """Keymanagement not available for unencrypted repositories."""
+
+
+class UnknownKeyType(Error):
+    """Keytype {0} is unknown."""
+
+
+class RepoIdMismatch(Error):
+    """This key backup seems to be for a different backup repository, aborting."""
+
+
+class NotABorgKeyFile(Error):
+    """This file is not a borg key backup, aborting."""
+
+
+def sha256_truncated(data, num):
+    h = sha256()
+    h.update(data)
+    return h.hexdigest()[:num]
+
+
+KEYBLOB_LOCAL = 'local'
+KEYBLOB_REPO = 'repo'
+
+
+class KeyManager:
+    def __init__(self, repository):
+        self.repository = repository
+        self.keyblob = None
+        self.keyblob_storage = None
+
+        try:
+            cdata = self.repository.get(Manifest.MANIFEST_ID)
+        except Repository.ObjectNotFound:
+            raise NoManifestError
+
+        key_type = cdata[0]
+        if key_type == KeyfileKey.TYPE:
+            self.keyblob_storage = KEYBLOB_LOCAL
+        elif key_type == RepoKey.TYPE or key_type == PassphraseKey.TYPE:
+            self.keyblob_storage = KEYBLOB_REPO
+        elif key_type == PlaintextKey.TYPE:
+            raise UnencryptedRepo()
+        else:
+            raise UnknownKeyType(key_type)
+
+    def load_keyblob(self):
+        if self.keyblob_storage == KEYBLOB_LOCAL:
+            k = KeyfileKey(self.repository)
+            target = k.find_key()
+            with open(target, 'r') as fd:
+                self.keyblob = ''.join(fd.readlines()[1:])
+
+        elif self.keyblob_storage == KEYBLOB_REPO:
+            self.keyblob = self.repository.load_key().decode()
+
+    def store_keyblob(self, args):
+        if self.keyblob_storage == KEYBLOB_LOCAL:
+            k = KeyfileKey(self.repository)
+            try:
+                target = k.find_key()
+            except KeyfileNotFoundError:
+                target = k.get_new_target(args)
+
+            self.store_keyfile(target)
+        elif self.keyblob_storage == KEYBLOB_REPO:
+            self.repository.save_key(self.keyblob.encode('utf-8'))
+
+    def store_keyfile(self, target):
+        with open(target, 'w') as fd:
+            fd.write('%s %s\n' % (KeyfileKey.FILE_ID, hexlify(self.repository.id).decode('ascii')))
+            fd.write(self.keyblob)
+            if not self.keyblob.endswith('\n'):
+                fd.write('\n')
+
+    def export(self, path):
+        self.store_keyfile(path)
+
+    def export_paperkey(self, path):
+        def grouped(s):
+            ret = ''
+            i = 0
+            for ch in s:
+                if i and i % 6 == 0:
+                    ret += ' '
+                ret += ch
+                i += 1
+            return ret
+
+        export = 'To restore key use borg key-import --paper /path/to/repo\n\n'
+
+        binary = a2b_base64(self.keyblob)
+        export += 'BORG PAPER KEY v1\n'
+        lines = (len(binary) + 17) // 18
+        repoid = hexlify(self.repository.id).decode('ascii')[:18]
+        complete_checksum = sha256_truncated(binary, 12)
+        export += 'id: {0:d} / {1} / {2} - {3}\n'.format(lines,
+                                       grouped(repoid),
+                                       grouped(complete_checksum),
+                                       sha256_truncated((str(lines) + '/' + repoid + '/' + complete_checksum).encode('ascii'), 2))
+        idx = 0
+        while len(binary):
+            idx += 1
+            binline = binary[:18]
+            checksum = sha256_truncated(idx.to_bytes(2, byteorder='big') + binline, 2)
+            export += '{0:2d}: {1} - {2}\n'.format(idx, grouped(hexlify(binline).decode('ascii')), checksum)
+            binary = binary[18:]
+
+        if path:
+            with open(path, 'w') as fd:
+                fd.write(export)
+        else:
+            print(export)
+
+    def import_keyfile(self, args):
+        file_id = KeyfileKey.FILE_ID
+        first_line = file_id + ' ' + hexlify(self.repository.id).decode('ascii') + '\n'
+        with open(args.path, 'r') as fd:
+            file_first_line = fd.read(len(first_line))
+            if file_first_line != first_line:
+                if not file_first_line.startswith(file_id):
+                    raise NotABorgKeyFile()
+                else:
+                    raise RepoIdMismatch()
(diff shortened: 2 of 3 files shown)


borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+                if i and i % 6 == 0:
+                    ret += ' '
+                ret += ch
+                i += 1
+            return ret
+
+        export = 'To restore key use borg key-import --paper /path/to/repo\n\n'
+
+        binary = a2b_base64(self.keyblob)
+        export += 'BORG PAPER KEY v1\n'
```
enkore: I like this idea!

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+import textwrap
+from hashlib import sha256
+
+from .key import KeyfileKey, RepoKey, PassphraseKey, KeyfileNotFoundError, PlaintextKey
+from .helpers import Manifest, NoManifestError, Error
+from .repository import Repository
+
+
+class UnencryptedRepo(Error):
+    """Keymanagement not available for unencrypted repositories"""
```
ThomasWaldmann: Nitpick: add dots at end of sentence (also below).

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+            self.store_keyfile(target)
+        elif self.keyblob_storage == KEYBLOB_REPO:
+            self.repository.save_key(self.keyblob)
+
+    def store_keyfile(self, target):
+        with open(target, 'w') as fd:
+            fd.write('%s %s\n' % (KeyfileKey.FILE_ID, hexlify(self.repository.id).decode('ascii')))
+            fd.write(self.keyblob)
+            if not self.keyblob.endswith("\n\n"):
+                fd.write('\n')
```
ThomasWaldmann: I didn't search for it, but don't we have all that code in key.py?

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+                    ret += ' '
+                ret += ch
+                i += 1
+            return ret
+
+        export = 'To restore key use borg key-import --paper /path/to/repo\n\n'
+
+        binary = a2b_base64(self.keyblob)
+        export += 'BORG PAPER KEY v1\n'
+        lines = (len(binary)+17) // 18
```
ThomasWaldmann: pep8

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+        repoid = hexlify(self.repository.id).decode('ascii')[:18]
```
ThomasWaldmann: In master/1.1 we have the simpler bin_to_hex(), hex_to_bin() - please use it everywhere.

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+                if file_first_line.starts_with(file_id):
+                    raise NotABorgKeyFile()
+                else:
+                    raise RepoIdMismatch()
+            self.keyblob = fd.read()
+
+        self.store_keyblob(args)
+
+    def import_paperkey(self, args):
+        # imported here because it has global side effects
```
ThomasWaldmann: like?

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+        import readline
+
+        try:
+            repoid = hexlify(self.repository.id).decode('ascii')[:18]
+
+            while True:
+                idline = input("id: ")
```
ThomasWaldmann: In the rest of the code, I think we use mostly single quotes for strings. You also do it, above.

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+                if idx == lines:
+                    break
+                idx += 1
+
+        except EOFError:
+            print("\n - aborted")
+            return
+
+        if sha256_hexdigest(result)[:12] != id_complete_checksum:
+            print("The complete checksum did not match, aborting")
```
ThomasWaldmann: dot.

borg/keymanager.py
```diff
@@ -0,0 +1,210 @@
...
+
+class RepoIdMismatch(Error):
+    """This key backup seems to be for a different backup repository, aborting."""
+
+
+class NotABorgKeyFile(Error):
+    """This file is not a borg key backup, aborting."""
+
+
+def sha256_hexdigest(data):
```
ThomasWaldmann: as you often only want the first n digits, maybe add n as param?

textshell: ok. It‘s a bit less explicit, but shorter.

borg/keymanager.py
textshell: yes, we have some of this in key.py already. But not readily reusable. Most of the code here just calls into key.py. store_keyfile is the middle  part of KeyfileKey.save. Maybe i could factor that out.

borg/keymanager.py
textshell: with these helpers i’m always unsure. Currently this code is targeting 1.0-maint where they don‘t even exist.

borg/keymanager.py
textshell: yes. I tend to loose them somehow. Should be fixed now.

borg/keymanager.py
textshell: should be fixed.

borg/keymanager.py
ThomasWaldmann: for 1.0 it's maybe even better like it is, not having to refactor code...

but we can do it later in master.

borg/keymanager.py
ThomasWaldmann: yup, i was in the wrong branch mindset also. :)

borg/keymanager.py
```diff
@@ -103,16 +103,16 @@ def grouped(s):
         export += 'BORG PAPER KEY v1\n'
         lines = (len(binary)+17) // 18
         repoid = hexlify(self.repository.id).decode('ascii')[:18]
-        complete_checksum = sha256_hexdigest(binary)[:12]
+        complete_checksum = sha256_hexdigest(binary, 12)
```
ThomasWaldmann: as it is not really a (full) sha256 any more, maybe just call it hexdigest or hash_hex - also shorter to type?

borg/keymanager.py
```diff
@@ -144,26 +144,26 @@ def import_paperkey(self, args):
...
 
             while True:
-                idline = input("id: ")
+                idline = input('id: ')
                 try:
-                    (data, checksum) = idline.replace(" ", "").split("-")
+                    (data, checksum) = idline.replace(' ', '').split('-')
                 except ValueError:
-                    print("each line must contain exactly one '-', try again")
+                    print('each line must contain exactly one '-', try again')
```
ThomasWaldmann: guess that does not work any more

borg/keymanager.py
textshell: flake8 and pep8 agree that this is ok.

I assume you want more spaces?

borg/keymanager.py
textshell: input() uses readline. But i think the details don’t belong into borg, that‘s what the fine manual of python is for.

borg/keymanager.py
textshell: I now use sha256_truncated, it’s not shorter by nicely explicit.

borg/keymanager.py
textshell: fixed.

textshell:
It seems the general idea and the paper key export format are ok? If so i think i need to start adding tests. For now i think i skip import_paperkey, but the main parts of the rest should be easy enough to test.

borg/keymanager.py
ThomasWaldmann: yes, usually there should be space around operators.

ThomasWaldmann:
I didn't practically try it yet. Interesting would be roundtrip tests and also tests that show that a working key can be installed if it had to be installed from the backup.

textshell:
Sure, that something actually work is out of scope for code reviews and automated testing^^
I actually restored a destroyed keyfile with this code at some point. But i have to admin that i currently only do it with cut and paste. Typing that many chars will be quite annoying.

Currently this is only tested for keyfiles, but repokey should also work.

The current output looks like this:

```
# borg key-export --paper borg@server:/path
To restore key use borg key-import --paper /path/to/repo

BORG PAPER KEY v1
id: 19 / 29c739 c94b64 29bbff / f48426 8db397 - cb
 1: 86a961 6c676f 726974 686da6 736861 323536 - 14
...
19: 51e36c b15095 a77665 727369 6f6e01 - 76

```

borg/keymanager.py
enkore: complete checksum ~ overall checksum?

borg/archiver.py
```diff
@@ -1076,6 +1107,34 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
                                type=location_validator(archive=False))
 
+        subparser = subparsers.add_parser('key-export', parents=[common_parser],
```
enkore: General Q: it would seem that we'll see more "key-xyz" operations in the future; make "key" a command group instead?

```
borg key export
borg key import
...
borg key --help
```

textshell: Yes, i expect KeyManager to expose more commands in the future. I did’t look much at borg’s cmdparsing yet. Would we want a group still being dispatched in archiver or can/want we delegate everything into the key manager? Archiver is quite bit and unfocused.

borg/keymanager.py
textshell: yes, that’s better.

borg/archiver.py
ThomasWaldmann: @enkore did you also get the "yes, I expect ..." notification email twice?

textshell: fwiw i get almost every email from these review threads twice. Something broken with github it seems.

ThomasWaldmann: As this stuff should go into 1.0-maint, I'ld prefer not to do stuff differently as we already do there concerning argparsing.

enkore: Since this is new there is nothing to differ from.

(Re. mails, yeah, seems like they became more erratic with the recent update)

ThomasWaldmann: I meant like "do not move argparsing away from archiver.py", for example.

textshell: It seems moving argparsing into keymanager is too intrusive for 1.0-maint for Thomas.

We still should decide if having a "key" command and subcommands for import and exports is worth the style change from the existing debug commands.

borg/archiver.py
```diff
@@ -159,6 +160,36 @@ def do_change_passphrase(self, args, repository, manifest, key):
...
+    def do_key_export(self, args, repository):
+        """Export the repository key for backup"""
+        manager = KeyManager(repository)
+        manager.load_keyblob()
+        if args.paper:
+            manager.export_paperkey(args.path)
+        else:
+            if not args.path:
+                self.print_error("output file to export key to expected")
+                return EXIT_ERROR
```
ThomasWaldmann: as you also need args.path in line 169: move this check to above the if/else block?

borg/keymanager.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class NotABorgKeyFile(Error):
+    """This file is not a borg key backup, aborting."""
+
+
+def sha256_truncated(data, num):
+    h = sha256()
+    h.update(data)
+    return h.hexdigest()[:num]
```
ThomasWaldmann: hmm, for num only even numbers are acceptable, otherwise it can't be converted back to bytes.

borg/archiver.py
textshell: export_paperkey handles None by printing to stdout. i think it is likely that the output is pasted into some program for printing. Maybe even at the other side of an ssh connection, so this seems useful to keep.

borg/keymanager.py
```diff
@@ -0,0 +1,211 @@
...
+            print('\n - aborted')
+            return
+
+        if sha256_truncated(result, 12) != id_complete_checksum:
+            print('The overall checksum did not match, aborting.')
+            return
+
+        self.keyblob = '\n'.join(textwrap.wrap(b2a_base64(result).decode('ascii'))) + '\n'
+
+        self.store_keyblob(args)
```
ThomasWaldmann: there is a lot of vertical spacing here - more than usual in our source?

borg/keymanager.py
textshell: maybe. But this is actually never converted back to bytes. So i’m not sure what to do about it. I do like that i get as many digits as the number specifies. But you certainly can argue that bytes of hash result is more meaningful. On the other hand this all about counting keystrokes in a way.

borg/archiver.py
enkore: Where they are declared is not overly important I think (although adding it directly into keymanager may make maintenance easier in the future).

The part about having a `borg key` subcommand which keymanager-related commands is something we all agree on, yes?

borg/keymanager.py
textshell: yes, it seems that code didn‘t get the mandatory reduce whitespace pass needed for borg^^. Should be better now.

borg/keymanager.py
```diff
@@ -0,0 +1,202 @@
...
+        import readline
+
+        try:
+            repoid = hexlify(self.repository.id).decode('ascii')[:18]
+            while True:
+                idline = input('id: ')
+                try:
+                    (data, checksum) = idline.replace(' ', '').split('-')
+                except ValueError:
+                    print('each line must contain exactly one \'-\', try again')
```
ThomasWaldmann: compare this...

borg/keymanager.py
```diff
@@ -0,0 +1,202 @@
...
+                    continue
+                try:
+                    (id_lines, id_repoid, id_complete_checksum) = data.split('/')
+                except ValueError:
+                    print('the id line contains three \'/\', try again')
```
ThomasWaldmann: ... and that. same thing, different way to say it.

borg/keymanager.py
```diff
@@ -0,0 +1,202 @@
...
+                if idx == lines:
+                    break
+                idx += 1
+
+        except EOFError:
+            print('\n - aborted')
+            return
+
+        if sha256_truncated(result, 12) != id_complete_checksum:
+            print('The overall checksum did not match, aborting.')
```
ThomasWaldmann: as all the stuff is still in the readline buffer, maybe let user re-check?

borg/archiver.py
ThomasWaldmann: @enkore as that would be first subcommand, i do not think it is a good idea for 1.0-maint.
and at the time we introduce subcommands, we could refactor some other stuff also, e.g. debug-* commands.

borg/keymanager.py
textshell: reworded to be more consistent

borg/keymanager.py
textshell: yes, now lets the user retry with readline buffer intact. Also now offers to abort if an empty line was entered.

borg/archiver.py
enkore: We already have `borg help <thing>`. Same semantics...

I'm not really inclined to add something to 1.0-maint that we already know we want to do different in non-compatible ways. Since this is something supported by argparse out-of-the-box I see no technical issues; and the diff between the two solutions will likely be <10 lines.

borg/archiver.py
```diff
@@ -159,6 +160,39 @@ def do_change_passphrase(self, args, repository, manifest, key):
...
+            manager.export(args.path)
+        return EXIT_SUCCESS
+
+    @with_repository(lock=False, exclusive=False, manifest=False, cache=False)
+    def do_key_import(self, args, repository):
+        """Export the repository key for backup"""
```
ThomasWaldmann: Import... from...

borg/keymanager.py
```diff
@@ -0,0 +1,213 @@
...
+            raise UnencryptedRepo()
+        else:
+            raise UnknownKeyType(key_type)
+
+    def load_keyblob(self):
+        if self.keyblob_storage == KEYBLOB_LOCAL:
+            k = KeyfileKey(self.repository)
+            target = k.find_key()
+            with open(target, 'r') as fd:
+                self.keyblob = ''.join(fd.readlines()[1:])
```
ThomasWaldmann: hmm, shouldn't it check the first line in the file?

borg/testsuite/archiver.py
```diff
@@ -1194,6 +1196,108 @@ def test_debug_put_get_delete_obj(self):
...
+        repo_id = 'e294423506da4e1ea76e8dcdf1a3919624ae3ae496fddf905610c351d3f09239'
+
+        export_file = self.output_path + '/exported'
+        self.cmd('init', self.repository_location, '--encryption', 'keyfile')
+        self._set_repository_id(self.repository_path, unhexlify(repo_id))
+
+        key_file = self.keys_path + '/' + os.listdir(self.keys_path)[0]
+
+        with open(key_file, 'w') as fd:
+            # file_id + b' ' + hexlify(self.repository.id)
```
ThomasWaldmann: that comment seems a bit ... redundant.

borg/keymanager.py
textshell: k.find_key() does the needed checking.

borg/archiver.py
textshell: Oh, yes, fixed.

borg/testsuite/archiver.py
textshell: fixed.

ThomasWaldmann:
lgtm. collapse / merge?

about making subcommands: separate PR with general review of what should be a subcommand in 1.0-maint (rarely used debug-* come to mind, other non-breaking changes?)

textshell:
I’m fine with merging.

Yes, we can move subcommands into a separate PR.

borg/testsuite/archiver.py
```diff
@@ -1194,6 +1196,107 @@ def test_debug_put_get_delete_obj(self):
...
+        with open(key_file, 'w') as fd:
+            fd.write(KeyfileKey.FILE_ID + ' ' + repo_id + '\n')
+            fd.write(b2a_base64(b'abcdefghijklmnopqrstu').decode())
+
+        self.cmd('key-export', '--paper', self.repository_location, export_file)
+
+        with open(export_file, 'r') as fd:
+            export_contents = fd.read()
+
+        print(export_contents)
```
enkore: rogue print :)

textshell: fixed.

PR #1631, opened 2016-09-23
Improve signal handling

ThomasWaldmann:
Note: to ease the first review, I did not indent the blocks into the context manager scopes, but just left a TODO there.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -23,6 +23,7 @@ from .helpers import Error, location_validator, archivename_validator, format_li
     Manifest, NoManifestError, remove_surrogates, update_excludes, format_archive, check_extension_modules, Statistics, \
     dir_is_tagged, bigint_to_int, ChunkerParams, CompressionSpec, PrefixSpec, is_slow_msgpack, yes, sysinfo, \
     EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR, log_multi, PatternMatcher, ErrorIgnoringTextIOWrapper
+from .helpers import signal_handler, raising_signal_handler, SigHup, SigTerm
 from .logger import create_logger, setup_logging
 logger = create_logger()
 from .compress import Compressor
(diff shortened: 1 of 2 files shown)


borg/archiver.py
```diff
@@ -1658,44 +1661,23 @@ def sig_info_handler(signum, stack):  # pragma: no cover
...
+    # orderly exit.
+    # SIGHUP is important especially for systemd systems, where logind
+    # sends it when a session exits, in addition to any traditional use.
+    # Output some info if we receive SIGUSR1 or SIGINFO (ctrl-t).
+    with signal_handler('SIGINT', raising_signal_handler(KeyboardInterrupt)), \
+         signal_handler('SIGHUP', raising_signal_handler(SigHup)), \
+         signal_handler('SIGTERM', raising_signal_handler(SigTerm)), \
+         signal_handler('SIGUSR1', sig_info_handler), \
+         signal_handler('SIGINFO', sig_info_handler):
+        pass  # XXX TODO indent everything up to, but without sys.exit() XXX
```
ThomasWaldmann: @textshell suggested on IRC that the sys.exit() call should be made from within the context, so that there is no race condition when signals are received right there.

borg/helpers.py
```diff
@@ -1160,3 +1162,49 @@ def write(self, s):
...
+class SigHup(SignalException):
+    """raised on SIGHUP signal"""
+
+
+class SigTerm(SignalException):
+    """raised on SIGTERM signal"""
+
+
+@contextlib.contextmanager
+def signal_handler(sig, handler):
```
ThomasWaldmann: this is based on code already present in master, but was functionally enhanced to support giving signal NAMES and not failing when a signal name is not supported on some platform.

ThomasWaldmann:
could also go into 1.1.0b2 if we merge that soon.

ThomasWaldmann:
OK, I'll collapse/merge.

PR #1634, opened 2016-09-25
borg help compression, fixes #1582

### LINKED ISSUE #1582: create: better docs for --compression-from
#810 added a configuration example, but the main docs are kinda useless here.

`borg help compressionconfig` or so could be used to explain the format and give some examples. Other help output should refer there. A sentence in borg-(re)create (in the --compression-from help directly I'd say) saying what it's for / what it does would also be good ("read compression patterns" is maybe a bit technical)

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -1313,26 +1313,26 @@ def do_break_lock(self, args, repository):
...
+             # example config file for --compression-from option
+             #
+             # Format of non-comment / non-empty lines:
+             # <compression-spec>:<path/filename pattern>
+             # compression-spec is same format as for --compression option
+             # path/filename pattern is same format as for --exclude option
+             none:*.gz
+             none:*.zip
+             none:*.mp3
+             none:*.ogg
```
ThomasWaldmann: to start, i just copied another borg help section. so if indentation was wrong, it needs to be checked there also.

PR #1636, opened 2016-09-25
Use subcommands for debug-* and key-*

textshell:
This moves the key-export and key-import commands to subcommands of the key command.

For the existing debug-* commands it adds a debug command with subcommands but doesn‘t remove the debug-* commands. This should be done later for 1.1.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1110,7 +1110,15 @@ class Archiver:
         subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
                                type=location_validator(archive=False))
 
-        subparser = subparsers.add_parser('key-export', parents=[common_parser],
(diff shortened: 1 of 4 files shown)


enkore:
I'd say debug-XXX being what they are - and having no mention in the docs beyond "do not use this" - it'd be ok to just move them. No script would, or should, use these. Opinions?

textshell:
I strongly dislike breaking changes in micro versions. The cost of keeping them is minimal. I intend to do a follow up for master that removes the debug-* variants as soon as this merges down into master.

But in this case this is more a "because of principle" thing than anything else. Only plausible thing is that  when debugging something in irc or the bug tracker we can just assume 1.0.x has debug-* if at all.

enkore:
Good point.

PR #1639, opened 2016-09-25
development.rst: Add sphinx_rtd_theme to the sphinx install command

textshell:
It’s used by default, so install it as well.


ThomasWaldmann:
1.0-maint ?

textshell:
sure, if you like that better.

PR #1661, opened 2016-09-29
update CHANGES

ThomasWaldmann:


docs/changes.rst
```diff
@@ -50,6 +50,68 @@ The best check that everything is ok is to run a dry-run extraction::
     borg extract -v --dry-run REPO::ARCHIVE
 
 
+Version 1.0.8rc1 (not released yet)
+-----------------------------------
+
+Bug fixes:
+
+- fix read-from-closed-FD issue, #1551
```
enkore: I'd make this the last item, since it was never (afaik) triggeredin 1.0.x, so it's least important.

PR #1665, opened 2016-09-30
borg extract: add --continue flag

### LINKED ISSUE #1356: resume an extract operation
in master (borg 1.1), borg create has quite nice in-file checkpointing abilities, so you can even backup huge files over instable connections.

but if you want to extract such a huge file over a instable connection, it might break down before completion and there is no resume yet.

there are 2 cases:
a) a bunch of files has already been extracted (each file complete, but not all files you want)
b) extraction of a huge file breaks down in the middle

enkore:
Fixes #1356

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1,4 +1,5 @@
 import errno
+import io
 import os
 import socket
 import stat
@@ -22,7 +23,7 @@ from .chunker import Chunker
 from .compress import Compressor
 from .constants import *  # NOQA
 from .hashindex import ChunkIndex, ChunkIndexEntry
-from .helpers import Manifest
+from .helpers import Manifest, slice_chunks
 from .helpers import Chunk, ChunkIteratorFileWrapper, open_item
 from .helpers import Error, IntegrityError
 from .helpers import uid2user, user2uid, gid2group, group2gid
@@ -428,7 +429,7 @@ Number of files: {0.stats.nfiles}'''.format(
         return stats
 
     def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False,
-                     hardlink_masters=None, stripped_components=0, original_path=None, pi=None):
+                     hardlink_masters=None, complete_partial=False, stripped_components=0, original_path=None, pi=None):
         """
         Extract archive item.
 
@@ -437,6 +438,7 @@ Number of files: {0.stats.nfiles}'''.format(
         :param dry_run: do not write any data
         :param stdout: write extracted data to stdout
         :param sparse: write sparse files (chunk-granularity, independent of the original being sparse)
+        :param complete_partial: False: replace files, True: existing files are completed
         :param hardlink_masters: maps paths to (chunks, link_target) for extracting subtrees with hardlinks correctly
         :param stripped_components: stripped leading path components to correct hard link extraction
         :param original_path: 'path' key as stored in archive
@@ -466,14 +468,15 @@ Number of files: {0.stats.nfiles}'''.format(
         # Attempt to remove existing files, ignore errors on failure
         try:
             st = os.lstat(path)
-            if stat.S_ISDIR(st.st_mode):
-                os.rmdir(path)
-            else:
-                os.unlink(path)
+            if not complete_partial:
+                if stat.S_ISDIR(st.st_mode):
+                    os.rmdir(path)
+                else:
+                    os.unlink(path)
         except UnicodeEncodeError:
             raise self.IncompatibleFilesystemEncodingError(path, sys.getfilesystemencoding()) from None
         except OSError:
-            pass
+            st = None
         mode = item.mode
         if stat.S_ISREG(mode):
             with backup_io():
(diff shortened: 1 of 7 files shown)


src/borg/archive.py
```diff
@@ -509,8 +524,6 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
                         else:
                             fd.write(data)
                 with backup_io():
-                    pos = fd.tell()
-                    fd.truncate(pos)
```
enkore: gnah. mismerge.

src/borg/helpers.py
```diff
@@ -1769,3 +1769,21 @@ def swidth_slice(string, max_width):
...
+
+    Return (sliced_chunks, prefix_length).
+    """
+    def should_drop(chunk):
+        nonlocal current_length
+        dropped = (current_length + chunk.size) <= maximum_length
+        if dropped:
+            current_length += chunk.size
+        return dropped
+    current_length = 0
```
ThomasWaldmann: add an empty line after return?

src/borg/helpers.py
```diff
@@ -1769,3 +1769,21 @@ def swidth_slice(string, max_width):
...
+    sliced_chunks = list(dropwhile(should_drop, chunks))
+    assert current_length >= 0
```
ThomasWaldmann: hm? in case chunk.size gets negative?

src/borg/testsuite/archiver.py
```diff
@@ -736,6 +736,22 @@ def test_extract_with_pattern(self):
...
+        self.create_regular_file('file2', size=1024 * 100)
+        self.cmd('create', self.repository_location + '::test', 'input')
+
+        with changedir('output'):
+            self.cmd('extract', self.repository_location + '::test')
+            with open('input/file1', 'ab') as file:
+                file.truncate(1234)
+            with open('input/file2', 'ab') as file:
+                file.write(b'extra bytes')
+            print(self.cmd('extract', self.repository_location + '::test', '--continue'))
```
ThomasWaldmann: print?

src/borg/testsuite/archiver.py
```diff
@@ -736,6 +736,22 @@ def test_extract_with_pattern(self):
...
+            assert os.path.getsize('input/file1') == 1024 * 80
+            assert os.path.getsize('input/file2') == 1024 * 100
```
ThomasWaldmann: even better than just checking length would be checking that content is as original file.

src/borg/archive.py
```diff
@@ -428,7 +429,7 @@ def add_file_chunks(chunks):
         return stats
 
     def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False,
-                     hardlink_masters=None, stripped_components=0, original_path=None, pi=None):
+                     complete_partial=False, stripped_components=0, hardlink_masters=None, original_path=None, pi=None):
```
ThomasWaldmann: you swapped order for some reason?

src/borg/archive.py
```diff
@@ -496,9 +499,21 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
+                        existing_length = fd.tell()
+                        # Slice chunks by current length of the existing file.
+                        chunks, prefix_length = slice_chunks(chunks, maximum_length=existing_length)
+                        # We don't bother extracting fractional chunks. Just seek to a chunk boundary.
+                        fd.seek(prefix_length)
+                        fd.truncate()
+                    if pi:
+                        pi.show(increase=prefix_length)
+                ids = [c.id for c in chunks]
                 for _, data in self.pipeline.fetch_many(ids, is_preloaded=True):
```
ThomasWaldmann: not sure if this causes an issue with the preloaded chunks:
iirc it preloads all chunks of a file when processing its item, but then it fetches either none (existing, full file) or only some (existing, partial file), before it goes into fetch-all mode (non-existing files).

enkore: Good catch, didn't consider that. I made up RemoteRepository.discard_preload for that, which looks about right to me, but makes it hang - for some reason.

src/borg/archive.py
```diff
@@ -511,6 +511,8 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
                         # We don't bother extracting fractional chunks. Just seek to a chunk boundary.
                         fd.seek(prefix_length)
                         fd.truncate()
+                    discarded_chunk_ids = [c.id for c in item.chunks[:len(chunks)]]
```
ThomasWaldmann:
```
discarded_count = len(item.chunks) - len(chunks)
discarded_chunks_ids =  [c.id for c in item.chunks[:discarded_count]]
```

src/borg/remote.py
```diff
@@ -506,6 +506,22 @@ def close(self):
...
+
+        for discard_id in discard_ids:
+            msgids = self.cache.get((discard_id,), [])
+            for msgid in list(msgids):
+                if msgid in self.responses:
+                    # Discard anything we may have already received
+                    self.responses.pop(msgid)
+                    msgids.remove(msgid)
+                else:
+                    # The call was sent out, but no response yet; if we got one, we'll ignore it
```
ThomasWaldmann: ... if we get one...

src/borg/archive.py
enkore: This is kinda strange. If I set a breakpoint in RR.discard_ids I'll see that neither cache nor responses contain anything, but still removing the IDs from preload_ids makes it hang.

src/borg/archive.py
enkore: Thanks and *ouch* that mistake was actually the cause of the deadlocks M(

src/borg/archive.py
```diff
@@ -476,7 +476,7 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
         except UnicodeEncodeError:
             raise self.IncompatibleFilesystemEncodingError(path, sys.getfilesystemencoding()) from None
         except OSError:
-            pass
+            st = False
```
ThomasWaldmann: None?

src/borg/archive.py
```diff
@@ -499,7 +499,13 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
                     return
                 # Extract chunks, since the item which had the chunks was not extracted
             with backup_io():
-                fd = open(path, 'ab')
+                if complete_partial and st:
```
ThomasWaldmann: ... and st is not None:

enkore:
FYI the reason this was moved to b3 is that with larger files it still hangs in call_many, in the same spot as earlier:

> If I set a breakpoint in RR.discard_ids I see that neither cache nor responses contain anything, but still - removing the IDs from preload_ids makes it hang in call_many, in the while not self.to_send and (calls or self.preload_ids) and len(waiting_for) < MAX_INFLIGHT: loop.

I haven't looked at it again yet.

src/borg/archive.py
```diff
@@ -496,9 +499,30 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
+                    with backup_io():
+                        fd.seek(0, io.SEEK_END)
+                        existing_length = fd.tell()
+                        # Slice chunks by current length of the existing file.
+                        chunks, prefix_length = slice_chunks(chunks, maximum_length=existing_length)
+                        # We don't bother extracting fractional chunks. Just seek to a chunk boundary.
+                        fd.seek(prefix_length)
+                        fd.truncate()
+                    discarded_count = len(item.chunks) - len(chunks)
+                    discarded_chunks_ids = [c.id for c in item.chunks[:discarded_count]]
```
ThomasWaldmann: nitpick: discard_count / discard_chunk_ids (not past tense)

enkore:
"later" and no activity = close.

PR #1666, opened 2016-09-30
add more specific warning about write-access debug commands

### LINKED ISSUE #1587: debug-put and debug-delete do not update chunks index
These commands are quite low-level and simplistic - maybe a bit too simple?

If they do not update the cache, it will be out-of-sync with the repo.

And: borg won't notice that it is out of sync.

ThomasWaldmann:
fixes #1587.


ThomasWaldmann:
anything else?

PR #1672, opened 2016-10-01
Clarify FAQ regarding backup of virtual machines

enkore:
WIP / Needs some polishing.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -12,6 +12,33 @@ Yes, the `deduplication`_ technique used by
 |project_name| makes sure only the modified parts of the file are stored.
 Also, we have optional simple sparse file support for extract.
 
+If you use tools like Borg (or rsync, tar, cp and most other backup tools) to backup
+virtual machines, then these should be turned off for doing so, since Borg doesn't do
+*any* kind of snapshotting on it's own. Backing up live VMs this way can (and will)
+result in corrupted or inconsistent backup contents; a VM image is just a regular file,
+with the same issues as regular files when it comes to concurrent reading at writing from
+the same file.
+
+A better method is to use file system snapshots on the hosts, which establishes
+crash-consistency for the VM images. This means that with most file systems
+(that are journaling) the FS will always be fine in the backup (but may need a
+journal replay to become accessible).
+
+Usually this does not mean that file *contents* are consistent, since file contents
+are normally not journaled. Notable exceptions are ext4 in data=journal mode or ZFS.
+
+Applications designed with crash-consistency in mind (most relational databases
+like postgres, SQLite etc. but also for example Borg repositories) should always
+be able to recover to a consistent state from a backup created with
+crash-consistent snapshots.
+
+Hypervisor snapshots can also be used for backups and can be a better alternative to
+pure file system based snapshots of the VMs disk.
+
+Other applications may require a lot of work to reach that level: this is a broad
+and complex issue that cannot be explained in entirety here. Borg doesn't intend
+to address these issues do their huge complexity and platform dependency.
+
 Can I backup from multiple servers into a single repository?
 ------------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -12,11 +12,11 @@ Yes, the `deduplication`_ technique used by
 |project_name| makes sure only the modified parts of the file are stored.
 Also, we have optional simple sparse file support for extract.
 
-If you use tools like Borg (or rsync, tar, cp and most other backup tools) to backup
-virtual machines, then these should be turned off for doing so, since Borg doesn't do
+If you use tools like Borg (or rsync, tar, cp and most other backup tools) to back
+virtual machines up, then these should be turned off for doing so, since Borg doesn't do
 *any* kind of snapshotting on it's own. Backing up live VMs this way can (and will)
-result in corrupted or inconsistent backup contents; a VM image is just a regular file,
-with the same issues as regular files when it comes to concurrent reading at writing from
+result in corrupted or inconsistent backup contents: a VM image is just a regular file to
+Borg with the same issues as regular files when it comes to concurrent reading and writing from
 the same file.
 
 A better method is to use file system snapshots on the hosts, which establishes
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -12,6 +12,36 @@ Yes, the `deduplication`_ technique used by
 |project_name| makes sure only the modified parts of the file are stored.
 Also, we have optional simple sparse file support for extract.
 
+If you use tools like Borg (or rsync, tar, cp and most other backup tools) to back
+virtual machines up, then these should be turned off for doing so, since Borg doesn't do
```
ThomasWaldmann: ... back up virtual machines, ...

docs/faq.rst
```diff
@@ -12,6 +12,36 @@ Yes, the `deduplication`_ technique used by
...
+*any* kind of snapshotting on it's own. Backing up live VMs this way can (and will)
```
ThomasWaldmann: its

docs/faq.rst
```diff
@@ -12,6 +12,36 @@ Yes, the `deduplication`_ technique used by
...
+result in corrupted or inconsistent backup contents: a VM image is just a regular file to
+Borg with the same issues as regular files when it comes to concurrent reading and writing from
+the same file.
+
+A better method is to use file system snapshots on the hosts, which establishes
```
ThomasWaldmann: hosts is very generic. maybe "VM host"?

docs/faq.rst
```diff
@@ -12,6 +12,36 @@ Yes, the `deduplication`_ technique used by
...
+crash-consistency for the VM images. This means that with most file systems
+(that are journaling) the FS will always be fine in the backup (but may need a
+journal replay to become accessible).
+
+Usually this does not mean that file *contents* are consistent, since file contents
+are normally not journaled. Notable exceptions are ext4 in data=journal mode or ZFS.
+
+Applications designed with crash-consistency in mind (most relational databases
+like postgres, SQLite etc. but also for example Borg repositories) should always
```
ThomasWaldmann: PostgreSQL (postgres is the old stuff)

docs/faq.rst
```diff
@@ -12,6 +12,36 @@ Yes, the `deduplication`_ technique used by
...
+
+Hypervisor snapshots can also be used for backups and can be a better alternative to
+pure file system based snapshots of the VMs disk, since no state is lost. Depending
+on the application this can be the easiest and most reliable way to create application-
+consistent backups.
+
+Other applications may require a lot of work to reach application-consistency:
+It's a broad and complex issue that cannot be explained in entirety here.
+
+Borg doesn't intend to address these issues do their huge complexity and platform/software dependency.
```
ThomasWaldmann: ... due to their ...

docs/faq.rst
```diff
@@ -12,35 +12,36 @@ Yes, the `deduplication`_ technique used by
...
 be able to recover to a consistent state from a backup created with
-crash-consistent snapshots.
+crash-consistent snapshots (even on ext4 with data=writeback or XFS).
 
-Hypervisor snapshots can also be used for backups and can be a better alternative to
-pure file system based snapshots of the VMs disk, since no state is lost. Depending
-on the application this can be the easiest and most reliable way to create application-
-consistent backups.
+Hypervisor snapshots capturing most of the VM's state can also be used for backups
+and can be a better alternative to pure file system based snapshots of the VMs disk,
```
ThomasWaldmann: VM's

docs/faq.rst
```diff
@@ -12,6 +12,39 @@ Yes, the `deduplication`_ technique used by
...
+Borg with the same issues as regular files when it comes to concurrent reading and writing from
+the same file.
+
+A better method is to use file system snapshots on the VM host, which establishes
+crash-consistency for the VM images. This means that with most file systems
+(that are journaling) the FS will always be fine in the backup (but may need a
+journal replay to become accessible).
+
+Usually this does not mean that file *contents* on the VM are consistent, since file
+contents are normally not journaled. Notable exceptions are ext4 in data=journal mode or ZFS.
```
textshell: I believe btrfs is also a cow by default files system that implicitly journals all data writes. Not sure about how barrier/fsync respecting they are, but i would assume they are.

enkore: I'm not sure and couldn't find any solid references, so I left it out. I'll try to make this a formulation that makes it more clear it's not an exhaustive list.

docs/faq.rst
```diff
@@ -12,6 +12,39 @@ Yes, the `deduplication`_ technique used by
 |project_name| makes sure only the modified parts of the file are stored.
 Also, we have optional simple sparse file support for extract.
 
+If you use tools like Borg (or rsync, tar, cp and most other backup tools) to back up
```
textshell: most other backup tools is very vague. But i’m not sure what we could say instead that is not tautological like "other backup tools that don‘t use special file system snapshot features of the operating system". (also way to long)

docs/faq.rst
enkore: I guess one could read it into this: https://btrfs.wiki.kernel.org/index.php/FAQ#Does_Btrfs_have_data.3Dordered_mode_like_Ext3.3F

docs/faq.rst
enkore: `like Borg (and most others)` -- since backup tools that do their own snapshotting one way or another tend to have this as a prominent feature?

docs/faq.rst
textshell: and that they have a special nodatacow file attribute to be used with databases that do their own journaling. I don‘t know is zfs has a similar attribute.
I think we should mention btrfs explicitly because it is getting popular.

docs/faq.rst
textshell: hmm, maybe we can find a nice wording. Maybe just move the key point (not using snapshot features) out of the parenthesis

enkore: If you use non-snapshotting backup tools like Borg ...

textshell: yes. Not sure about if/when. Maybe even all. But i’m not good these details of English.

docs/faq.rst
```diff
@@ -12,6 +12,39 @@ Yes, the `deduplication`_ technique used by
 |project_name| makes sure only the modified parts of the file are stored.
 Also, we have optional simple sparse file support for extract.
 
+If you use non-snapshotting backup tools like Borg to back up virtual machines,
+then these should be turned off for doing so. Backing up live VMs this way can (and will)
+result in corrupted or inconsistent backup contents: a VM image is just a regular file to
+Borg with the same issues as regular files when it comes to concurrent reading and writing from
+the same file.
+
+A better method is to use file system snapshots on the VM host, which establishes
```
textshell: "A better method" than what? This sounds a bit like a better method than using borg. But of course we mean that snapshots are needed in addition to using borg.

textshell:
lgtm

PR #1680, opened 2016-10-03
borg info <repo>: print general repo information

### LINKED ISSUE #1679: show repository ID with other repository information in info
borg-info should show the repository ID and maybe where the local cache is for that repo.

enkore:
```
Repository ID: 6cb322e51621ca7a805692788869fb0cf1556cc250d73b4274cd1dd96fb973ad
Encrypted: Yes (key file)
Key file: /home/mabe/.config/borg/keys/tr_keyfile
Cache: /home/mabe/.cache/borg/6cb322e51621ca7a805692788869fb0cf1556cc250d73b4274cd1dd96fb973ad
All archives:                    0 B                  0 B                  0 B

Chunk index:                       0                    0
```

I think this covers the most important questions:

- What's the repo ID
- Where is that cache
- Where is the key file I need to back up
- Is this stuff even encrypted?

Another interesting question would be "how large is the cache", but with this it can be answered with relative ease. Could still be added later, though.

Fixes #1679

~~PS: Fix for remote repos coming right up.~~ Removed repository path output as a fix; it's literally in the command line, unless the env var is used, so it'll be usually redundant anyway.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -920,7 +920,7 @@ class Archiver:
         if any((args.location.archive, args.first, args.last, args.prefix)):
             return self._info_archives(args, repository, manifest, key, cache)
         else:
-            return self._info_repository(cache)
+            return self._info_repository(repository, key, cache)
 
     def _info_archives(self, args, repository, manifest, key, cache):
         def format_cmdline(cmdline):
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -957,7 +957,17 @@ def format_cmdline(cmdline):
                 print()
         return self.exit_code
 
-    def _info_repository(self, cache):
+    def _info_repository(self, repository, key, cache):
+        print('Repository ID: %s' % bin_to_hex(repository.id))
+        print('Cache: %s' % cache.path)
```
ThomasWaldmann: maybe move "cache: ...", below the key file.
currently, "encrypted: ..." directly follows "cache: ..." so people could think the cache is encrypted.

rest lgtm.

Issue #1610, opened 2016-09-16
repository / rpc api: sequential scan


ThomasWaldmann:
The current `repo.list()` method lists repo objects in repo index order (means: random segment order).
If one just wants the keys, this is no problem, everything is read from the index, which is in memory.

But if you want (all) the values also, this order causes a lot of random access - a real stress test for your disk(s).

If we want to iterate over all repo objects, same-segment offset (1.) and segment (2.) order would be much better. Maybe it could also support just getting the keys and alternatively key/value pairs.

ThomasWaldmann:
Algorithm could be like:
```
for all segment files:
    for all objects in segment file:
        i_pos = index[key]  # seg, offs
        if i_pos == current_pos:
            yield key, data
```

enkore:
I think check is the only op using it anyway. For the reconstruction cache population it doesn't matter how they're ordered, so changing the existing implementation to returning linearly ordered keys makes sense. Since it's purely in-memory it should be plenty fast. Maybe even implement it straight in Cython.

ThomasWaldmann:
borg check and some debug cmd uses it currently.

What I was proposing above is a disk-based algorithm that just checks against the repo index to make sure it is no superseded object, but results in a completely linear ordering for all keys.

So, considering that the repo index itself might already take some hundred MBs .. some GBs of memory, how would your purely-in-memory algorithm look like?

enkore:
Like a nested-loop JOIN, so it'd be ~O(n_segments * n_index), but doesn't do any IO. Hm.

ThomasWaldmann:
OK, so have a low-level op in C / Cython that does a hashtable scan, filtering out all entries for segment N, then sort (much smaller) result by offset.

ThomasWaldmann:
see PR #1607

enkore:
We removed the sequential scan API from #1607 again due to some complications.

What's currently on the table:

- Add new Repository method + RPC call called eg. `scan` or `linear_scan`
- Which returns a number of object IDs (with the usual limit and marker args)
- It works as follows:
  - If `marker`: get (segment, offset) from index
  - Elif not `marker`: segment = first segment, offset = 0
  - iter_objects from (segment, offset) on; ignore everything except PUTs, yield PUTs that are in the index and current
  - Take care not to evict the read segments from the page cache (this bit is important)
  - After yielding `limit` IDs, return.

- This enables
  - Efficient linear scanning
[...]

PR #1682, opened 2016-10-03
WIP: repo.scan - list keys in disk-order

ThomasWaldmann:
Issue #1610.

I kept repo.list as is and just added a new method repo.scan.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -2,7 +2,7 @@ import errno
 import os
 import shutil
 import struct
-from binascii import unhexlify
+from binascii import hexlify, unhexlify
 from collections import defaultdict
 from configparser import ConfigParser
 from datetime import datetime
@@ -750,9 +750,52 @@ class Repository:
         return id in self.index
 
     def list(self, limit=None, marker=None):
+        """
+        list <limit> IDs starting from after id <marker> - in index (pseudo-random) order.
+        """
         if not self.index:
             self.index = self.open_index(self.get_transaction_id())
         return [id_ for id_, _ in islice(self.index.iteritems(marker=marker), limit)]
+
+    def scan(self, limit=None, marker=None):
+        """
+        list <limit> IDs starting from after id <marker> - in on-disk order, so that a client
+        fetching data in this order does linear reads and reuses stuff from disk cache.
+
+        We rely on repository.check() has run already (either now or some time before) and that:
+        - if we are called from a borg check command, self.index is a valid, fresh, in-sync repo index.
+        - if we are called from elsewhere, either self.index or the on-disk index is valid and in-sync.
+        - the repository segments are valid (no CRC errors).
+          if we encounter CRC errors in segment entry headers, rest of segment is skipped.
+        """
+        if limit is not None and limit < 1:
+            raise ValueError('please use limit > 0 or limit = None')
+        if not self.index:
+            transaction_id = self.get_transaction_id()
+            self.index = self.open_index(transaction_id)
+        at_start = marker is None
+        # smallest valid seg is <uint32> 0, smallest valid offs is <uint32> 8
+        marker_segment, marker_offset = (0, 0) if at_start else self.index[marker]
+        result = []
+        for segment, filename in self.io.segment_iterator():
+            if segment < marker_segment:
+                continue
+            obj_iterator = self.io.iter_objects(segment, read_data=False, include_data=False)
+            while True:
+                try:
+                    tag, id, offset, size = next(obj_iterator)
+                except (StopIteration, IntegrityError):
+                    # either end-of-segment or an error - we can not seek to objects at
+                    # higher offsets than one that has an error in the header fields
+                    break
+                if segment == marker_segment and offset <= marker_offset:
+                    continue
+                if tag == TAG_PUT and (segment, offset) == self.index.get(id):
+                    # we have found an existing and current object
+                    result.append(id)
+                    if len(result) == limit:
+                        return result
+        return result
 
     def get(self, id_):
         if not self.index:
--- a/src/borg/testsuite/repository.py
+++ b/src/borg/testsuite/repository.py
@@ -133,6 +133,7 @@ class RepositoryTestCase(RepositoryTestCaseBase):
     def test_list(self):
         for x in range(100):
             self.repository.put(H(x), b'SOMEDATA')
+        self.repository.commit()
         all = self.repository.list()
         self.assert_equal(len(all), 100)
         first_half = self.repository.list(limit=50)
@@ -142,6 +143,23 @@ class RepositoryTestCase(RepositoryTestCaseBase):
         self.assert_equal(len(second_half), 50)
         self.assert_equal(second_half, all[50:])
         self.assert_equal(len(self.repository.list(limit=50)), 50)
+
+    def test_scan(self):
+        for x in range(100):
+            self.repository.put(H(x), b'SOMEDATA')
+        self.repository.commit()
+        all = self.repository.scan()
+        assert len(all) == 100
+        first_half = self.repository.scan(limit=50)
+        assert len(first_half) == 50
+        assert first_half == all[:50]
+        second_half = self.repository.scan(marker=first_half[-1])
+        assert len(second_half) == 50
+        assert second_half == all[50:]
+        assert len(self.repository.scan(limit=50)) == 50
+        # check result order == on-disk order (which is hash order)
+        for x in range(100):
+            assert all[x] == H(x)
 
     def test_max_data_size(self):
         max_data = b'x' * MAX_DATA_SIZE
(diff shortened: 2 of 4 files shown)


src/borg/repository.py
```diff
@@ -750,10 +750,67 @@ def __contains__(self, id):
...
+                except (StopIteration, IntegrityError):
+                    # either end-of-segment or an error - we can not seek to objects at
+                    # higher offsets than one that has an error in the header fields
+                    break
+                if marker_segment is not None and segment == marker_segment and \
+                   marker_offset is not None and offset <= marker_offset:
+                    continue
+                if tag == TAG_PUT:
+                    wanted_pos = self.index.get(id)
+                    if wanted_pos is not None and wanted_pos == (segment, offset):
```
ThomasWaldmann: first term is superfluous as None != (segment, offset) ever.

src/borg/testsuite/repository.py
```diff
@@ -143,6 +144,20 @@ def test_list(self):
...
+        self.repository.commit()
+        all = self.repository.scan()
+        self.assert_equal(len(all), 100)
+        first_half = self.repository.scan(limit=50)
+        self.assert_equal(len(first_half), 50)
+        self.assert_equal(first_half, all[:50])
+        second_half = self.repository.scan(marker=first_half[-1])
+        self.assert_equal(len(second_half), 50)
+        self.assert_equal(second_half, all[50:])
+        self.assert_equal(len(self.repository.scan(limit=50)), 50)
```
ThomasWaldmann: more tests are needed here, for "remaining".

enkore: Use plain asserts for new code?

src/borg/repository.py
```diff
@@ -750,10 +750,66 @@ def __contains__(self, id):
...
+            transaction_id = self.get_transaction_id()
+            self.index = self.open_index(transaction_id)
+        # create a second index object - use it to discover object IDs we can not discover on-disk
+        # (either because they are not there, just in the index, or because they have IntegrityErrors).
+        at_start = marker is None
+        if at_start:
+            transaction_id = self.get_transaction_id()
+            self.remaining = self.open_index(transaction_id)
+        if at_start:
+            marker_segment, marker_offset = None, None
```
enkore: Why not (0, 0)? This would avoid the special casing of None below and still be equivalent:

src/borg/repository.py
```diff
@@ -750,10 +750,66 @@ def __contains__(self, id):
...
+        else:
+            marker_segment, marker_offset = self.index[marker]
+        result = []
+        for segment, filename in self.io.segment_iterator():
+            if marker_segment is not None and segment < marker_segment:
```
enkore: ok: first segment >= 0; but 0<0 is false.

src/borg/repository.py
```diff
@@ -750,10 +750,66 @@ def __contains__(self, id):
...
+            obj_iterator = self.io.iter_objects(segment, read_data=False, include_data=False)
+            while True:
+                try:
+                    tag, id, offset, size = next(obj_iterator)
+                except (StopIteration, IntegrityError):
+                    # either end-of-segment or an error - we can not seek to objects at
+                    # higher offsets than one that has an error in the header fields
+                    break
+                if marker_segment is not None and segment == marker_segment and \
+                   marker_offset is not None and offset <= marker_offset:
```
enkore: ok: smallest offset is MAGIC_LEN > 0.

src/borg/testsuite/repository.py
ThomasWaldmann: ok

src/borg/repository.py
```diff
@@ -750,10 +750,66 @@ def __contains__(self, id):
...
+        list <limit> IDs starting from after id <marker> - in on-disk order, so that a client
+        fetching data in this order does linear reads and reuses stuff from disk cache.
+        """
+        if limit is not None and limit < 1:
+            raise ValueError('please use limit > 0 or limit = None')
+        if not self.index:
+            transaction_id = self.get_transaction_id()
+            self.index = self.open_index(transaction_id)
+        # create a second index object - use it to discover object IDs we can not discover on-disk
+        # (either because they are not there, just in the index, or because they have IntegrityErrors).
```
ThomasWaldmann: if limit is not None, we do not know the complete objects count, the counter will just count up to limit. also,as it is not a generator, we do not keep state...

enkore: Yes, the description wasn't very accurate.

The second index was also (hidden) state, in which case the counter would work. It could also be made explicit state, by making `marker` a tuple (object-id, object-counter) which is returned explicitly instead of implicitly (objects_ids[-1]), or similar. Ofc. with the current client code there is no problem doing this in the client, as it already has a complete copy of the repository index obtained at that point. So as it is now (a3cd5f3) would work and is arguably the simplest/neatest solution.

src/borg/repository.py
```diff
@@ -750,10 +750,57 @@ def __contains__(self, id):
...
+                    tag, id, offset, size = next(obj_iterator)
+                except (StopIteration, IntegrityError):
+                    # either end-of-segment or an error - we can not seek to objects at
+                    # higher offsets than one that has an error in the header fields
+                    break
+                if marker_segment is not None and segment == marker_segment and \
+                   marker_offset is not None and offset <= marker_offset:
+                    continue
+                if tag == TAG_PUT and (segment, offset) == self.index.get(id):
+                    # logger.debug("%05d %08d: %s", segment, offset, bin_to_hex(id))
```
enkore: leave or remove?

src/borg/repository.py
ThomasWaldmann: ok

src/borg/repository.py
```diff
@@ -750,10 +750,54 @@ def __contains__(self, id):
...
+        - if we are called from a borg check command, self.index is a valid, fresh, in-sync repo index.
+        - if we are called from elsewhere, either self.index or the on-disk index is valid and in-sync.
+        - the repository segments are valid (no CRC errors).
+          if we encounter CRC errors in segment entry headers, rest of segment is skipped.
+        """
+        if limit is not None and limit < 1:
+            raise ValueError('please use limit > 0 or limit = None')
+        if not self.index:
+            transaction_id = self.get_transaction_id()
+            self.index = self.open_index(transaction_id)
```
ThomasWaldmann: i'm unsure if this is completely correct.

repository.check does a lot more dance about getting the "right" transaction_id.
if we are in borg check and in repair mode, repo index should be repaired already - but what if we are not in repair mode?

enkore: (non-repair) If check doesn't report an error, then normal repository access must be ok.

Regarding transaction IDs: Repo vs index inconsistencies are fixed-or-aborted in get_transaction_id().

ThomasWaldmann: but if non-repair mode reports an error?

ThomasWaldmann:
I'ld collapse most changesets if you have finished reviewing what's already there.

src/borg/repository.py
ThomasWaldmann: yes, I'll implement it in the client / caller.

src/borg/repository.py
enkore: Then we won't get here.

ThomasWaldmann: ah, right :)

src/borg/repository.py
```diff
@@ -750,10 +750,53 @@ def __contains__(self, id):
...
 
     def list(self, limit=None, marker=None):
+        """
+        list <limit> IDs starting from after id <marker> - in index (pseudo-random) order.
+        """
         if not self.index:
             self.index = self.open_index(self.get_transaction_id())
         return [id_ for id_, _ in islice(self.index.iteritems(marker=marker), limit)]
 
+    def scan(self, limit=None, marker=None):
```
ThomasWaldmann: about list vs. scan:

somehow this is the same thing, so just do both via list() and add a linear=False param to list? in borg 1.0.x we could then do it in the "try newcall / except oldcall" way as we did at other places.

enkore: Wouldn't a separate call be cleaner, because it doesn't log a spurious TypeError we can't suppress?

ThomasWaldmann: hmm, have to try, but it might be just a "unsupported rpc call" log entry instead.

enkore: yeah \o/

ThomasWaldmann: ok, i'll add a in-disk-order test and then merge.

src/borg/testsuite/repository.py
```diff
@@ -143,6 +144,23 @@ def test_list(self):
...
+        first_half = self.repository.scan(limit=50)
+        assert len(first_half) == 50
+        assert first_half == all[:50]
+        second_half = self.repository.scan(marker=first_half[-1])
+        assert len(second_half) == 50
+        assert second_half == all[50:]
+        assert len(self.repository.scan(limit=50)) == 50
+        # check result order == on-disk order (which is hash order)
+        for x in range(100):
+            assert all[x] == H(x)
```
ThomasWaldmann: last change in this PR, merging after travis is happy ^^^

PR #1685, opened 2016-10-05
fix debug get-obj / delete-obj crash if object not found and remote repo 

### LINKED ISSUE #1684: debug get-obj / delete-obj crash if object not found and remote repo

enkore:
Fixes #1684

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -767,7 +767,7 @@ class Archiver:
         else:
             try:
                 data = repository.get(id)
-            except repository.ObjectNotFound:
+            except Repository.ObjectNotFound:
                 print("object %s not found." % hex_id)
             else:
                 with open(args.path, "wb") as f:
@@ -801,7 +801,7 @@ class Archiver:
                     repository.delete(id)
                     modified = True
                     print("object %s deleted." % hex_id)
-                except repository.ObjectNotFound:
+                except Repository.ObjectNotFound:
                     print("object %s not found." % hex_id)
         if modified:
             repository.commit()


borg/remote.py
```diff
@@ -308,7 +314,7 @@ def handle_error(error, res):
             elif error == b'ObjectNotFound':
                 raise Repository.ObjectNotFound(res[0], self.location.orig)
             elif error == b'InvalidRPCMethod':
-                raise InvalidRPCMethod(*res)
+                raise InvalidRPCMethod(res[0].decode())
```
enkore: "Invalid RPC method 82" (unpacking bytes = ints) => "RPC method negotiatea is not valid"

borg/testsuite/repository.py
```diff
@@ -420,7 +420,28 @@ def open(self, create=False):
                                 exclusive=True, create=create)
 
     def test_invalid_rpc(self):
-        self.assert_raises(InvalidRPCMethod, lambda: self.repository.call('__init__', None))
+        try:
+            self.repository.call('__init__', None)
```
ThomasWaldmann: ok, this was already strange before your change, but maybe we could fix it anyway:

``__init__`` rather looks like a valid rpc method. if we want something invalid, how about 'something_invalid'?

borg/archiver.py
```diff
@@ -767,7 +767,7 @@ def do_debug_get_obj(self, args, repository):
         else:
             try:
                 data = repository.get(id)
-            except repository.ObjectNotFound:
+            except Repository.ObjectNotFound:
```
ThomasWaldmann: oops. X-)

borg/remote.py
```diff
@@ -113,10 +113,13 @@ def serve(self):
...
                         # for these, except ErrorWithTraceback, which should always display a traceback.
-                        if not isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, PathNotAllowed)):
+                        if isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, Repository.ObjectNotFound,
+                                          Repository.CheckNeeded, IntegrityError, PathNotAllowed, InvalidRPCMethod)):
+                            exc = e.args
+                        else:
+                            exc = "Remote Exception (see remote log for the traceback)"
                             logging.exception('Borg %s: exception in RPC call:', __version__)
                             logging.error(sysinfo())
-                        exc = "Remote Exception (see remote log for the traceback)"
```
ThomasWaldmann: so, old borg serve code always gave a str for exc (on the remote side received as bytes due to msgpack).

borg/remote.py
```diff
@@ -308,7 +314,11 @@ def handle_error(error, res):
             elif error == b'ObjectNotFound':
                 raise Repository.ObjectNotFound(res[0], self.location.orig)
             elif error == b'InvalidRPCMethod':
-                raise InvalidRPCMethod(*res)
+                if isinstance(res, bytes):
+                    method = res
```
ThomasWaldmann:
```
# support old servers, borg < 1.0.8
```

borg/remote.py
```diff
@@ -308,7 +314,11 @@ def handle_error(error, res):
             elif error == b'ObjectNotFound':
                 raise Repository.ObjectNotFound(res[0], self.location.orig)
```
ThomasWaldmann: did you check if the above exception args are sane?

borg/remote.py
```diff
@@ -113,10 +113,13 @@ def serve(self):
                         # These exceptions are reconstructed on the client end in RemoteRepository.call_many(),
                         # and will be handled just like locally raised exceptions. Suppress the remote traceback
                         # for these, except ErrorWithTraceback, which should always display a traceback.
-                        if not isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, PathNotAllowed)):
+                        if isinstance(e, (Repository.DoesNotExist, Repository.AlreadyExists, Repository.ObjectNotFound,
+                                          Repository.CheckNeeded, IntegrityError, PathNotAllowed, InvalidRPCMethod)):
+                            exc = e.args
```
ThomasWaldmann: will this get (old) client code into trouble as exc used to be a bytes object?

ThomasWaldmann:
As this is in 1.0.8rc1 milestone and one of the last fixes for that, can we finish this one first?

enkore:
I removed the remote stuff from this branch, it literally doesn't matter and testing that change (>100 lines ugly code with old code from 1.0.7 copied in for compat testing, *no thanks*) has absolutely no proportion to the little benefit it may have.

PR #1690, opened 2016-10-06
borg.key: include chunk id in exception msgs

### LINKED ISSUE #1571: more informative IntegrityError
IntegrityError should whenever possible include the chunk id where the problem was encountered.

Currently the error message just says that a specific problem happened on some chunk without nameing the chunk id. That does make debugging and fixing hard. With the chunk id there is at least some chance to drill down and come up with some fix.

ThomasWaldmann:
this is a backport of bcdce91dfb2883c139011322a9e8086059fbe5c2 improvements on the exception msgs.

fixes #1571.

### DIFF
--- a/borg/key.py
+++ b/borg/key.py
@@ -105,10 +105,10 @@ class PlaintextKey(KeyBase):
 
     def decrypt(self, id, data):
         if data[0] != self.TYPE:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
merge?

PR #1703, opened 2016-10-09
Repository.scan(): avoid re-reading same repo object headers repeatedly

### LINKED ISSUE #1610: repository / rpc api: sequential scan

ThomasWaldmann:
the old way had a bad O(N^2) runtime behaviour. for small segments and large objects, it didn't matter much (small N), but for the 500MB large segments and small objects, N^2 can be huge.

btw, the high cpu usage @enkore has seen can not come from crc checks - we do not have a header crc (unfortunately), we just have a header+data crc - and we can't use that if we just read the header and seek over the data.

@enkore can you please repeat your perf test?

Fixes #1610

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -775,12 +775,10 @@ class Repository:
             self.index = self.open_index(transaction_id)
         at_start = marker is None
         # smallest valid seg is <uint32> 0, smallest valid offs is <uint32> 8
-        marker_segment, marker_offset = (0, 0) if at_start else self.index[marker]
+        start_segment, start_offset = (0, 0) if at_start else self.index[marker]
         result = []
-        for segment, filename in self.io.segment_iterator():
-            if segment < marker_segment:
-                continue
-            obj_iterator = self.io.iter_objects(segment, read_data=False, include_data=False)
+        for segment, filename in self.io.segment_iterator(start_segment):
+            obj_iterator = self.io.iter_objects(segment, start_offset, read_data=False, include_data=False)
             while True:
                 try:
                     tag, id, offset, size = next(obj_iterator)
@@ -788,7 +786,11 @@ class Repository:
                     # either end-of-segment or an error - we can not seek to objects at
                     # higher offsets than one that has an error in the header fields
                     break
-                if segment == marker_segment and offset <= marker_offset:
+                if start_offset > 0:
+                    # we are using a marker and the marker points to the last object we have already
+                    # returned in the previous scan() call - thus, we need to skip this one object.
+                    # also, for the next segment, we need to start at offset 0.
+                    start_offset = 0
                     continue
                 if tag == TAG_PUT and (segment, offset) == self.index.get(id):
                     # we have found an existing and current object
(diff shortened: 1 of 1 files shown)


enkore:
Yes, this seems to work much much better with an I/O utilization of 50-70 % (~60-90 MB/s).

src/borg/repository.py
```diff
@@ -776,19 +776,22 @@ def scan(self, limit=None, marker=None):
         at_start = marker is None
         # smallest valid seg is <uint32> 0, smallest valid offs is <uint32> 8
         marker_segment, marker_offset = (0, 0) if at_start else self.index[marker]
+        start_segment, start_offset = marker_segment, marker_offset
```
enkore: Assign s_s, s_o directly in the line above? m_s, m_o are unused now.

enkore: Assign s_s, s_o directly in the line above? m_s, m_o are unused now.

src/borg/repository.py
```diff
@@ -999,7 +1013,7 @@ def segment_exists(self, segment):
     def segment_size(self, segment):
         return os.path.getsize(self.segment_filename(segment))
 
-    def iter_objects(self, segment, include_data=False, read_data=True):
+    def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
```
enkore: API change ok; all call sites use keyword arguments for the existing 2nd and 3rd arguments.

src/borg/repository.py
```diff
@@ -886,14 +889,25 @@ def close_fd(self, fd):
             os.posix_fadvise(fd.fileno(), 0, 0, os.POSIX_FADV_DONTNEED)
         fd.close()
 
-    def segment_iterator(self, reverse=False):
+    def segment_iterator(self, segment=None, reverse=False):
```
enkore: ok; no args used / only kwargs used (reverse) in existing code.

src/borg/repository.py
```diff
@@ -886,14 +889,25 @@ def close_fd(self, fd):
...
             filenames = os.listdir(os.path.join(data_path, dir))
-            sorted_filenames = sorted((filename for filename in filenames
-                                       if filename.isdigit()), key=int, reverse=reverse)
-            for filename in sorted_filenames:
+            if not reverse:
+                filenames = [filename for filename in filenames if filename.isdigit() and int(filename) >= segment]
+            else:
+                filenames = [filename for filename in filenames if filename.isdigit() and int(filename) <= segment]
+            filenames = sorted(filenames, key=int, reverse=reverse)
+            for filename in filenames:
```
enkore: Mmhh, since the case of `segment is not None and reverse` isn't used and thus not tested . . . remove branches, add `if segment is not None and reverse: raise NotImplentedError`?

Suggesting this because I can't really come up with any use case for this combination. Reverse iteration is only used for finding the last commit and there's no reason to off-set that search...

PR #1708, opened 2016-10-10
update changed repo location immediately after acceptance

### LINKED ISSUE #1524: Cache’s previous_location is only updated on cache commit
When the "Warning: The repository at location {} was previously located at" prompt is answered the location should be be updated immediately instead of only on the next commit.

Rational: Automated backup scripts often have no way to answer such prompts. A natural first try is to do a quick operation from the shell to fix that problem like a borg list or borg info. But these commands don’t commit a cache transaction so previous_location is never updated.

ThomasWaldmann:
fixes #1524

before, if a longer backup got interrupted before commit(),
it asked same question again.


enkore:
In master we can do this more efficiently (SaveFile), but in 1.0-maint we'll play it safe.

PR #1711, opened 2016-10-12
implement /./relpath hack

### LINKED ISSUE #1655: ssh: repo urls and arbitrary relative pathes
```
    ssh_re = re.compile(r'(?P<proto>ssh)://(?:(?P<user>[^@]+)@)?'
                        r'(?P<host>[^:/#]+)(?::(?P<port>\d+))?'
                        r'(?P<path>[^:]+)(?:::(?P<archive>[^/]+))?$')
```
As the `/` terminating the host[:port] part is mapped into path, a path must begin with a slash and thus is absolute.

The code has a special hack to support `/~` there to specify a path relative to homedir, but not something to specify a path relative to cwd.

ThomasWaldmann:
Fixes #1655 

TODO:
- remove "WIP" from commit comment

Please help reviewing this so that we have no unexpected effects in stable release update.

### DIFF
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -835,26 +835,32 @@ class Location:
         return True
 
     def _parse(self, text):
+        def normpath_special(p):
+            # avoid that normpath strips away our relative path hack and even makes p absolute
+            relative = p.startswith('/./')
+            p = os.path.normpath(p)
+            return ('/.' + p) if relative else p
+
         m = self.ssh_re.match(text)
         if m:
             self.proto = m.group('proto')
             self.user = m.group('user')
             self.host = m.group('host')
             self.port = m.group('port') and int(m.group('port')) or None
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
             self.archive = m.group('archive')
             return True
         m = self.file_re.match(text)
         if m:
             self.proto = m.group('proto')
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
             self.archive = m.group('archive')
             return True
         m = self.scp_re.match(text)
         if m:
             self.user = m.group('user')
             self.host = m.group('host')
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
             self.archive = m.group('archive')
             self.proto = self.host and 'ssh' or 'file'
             return True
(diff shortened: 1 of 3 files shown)


borg/helpers.py
```diff
@@ -885,9 +891,9 @@ def canonical_path(self):
             return self.path
         else:
             if self.path and self.path.startswith('~'):
-                path = '/' + self.path
+                path = '/' + self.path  # /~/x = path x relative to home dir
             elif self.path and not self.path.startswith('/'):
-                path = '/~/' + self.path
+                path = '/./' + self.path  # /./x = path x relative to cwd
```
ThomasWaldmann: before this change, this was a bit strange: a relative path was transformed to a homedir-relative url.
as we support generic relative urls now, we transform it into such a url.

borg/helpers.py
```diff
@@ -835,26 +835,32 @@ def parse(self, text):
         return True
 
     def _parse(self, text):
+        def normpath_special(p):
+            # avoid that normpath strips away our relative path hack and even makes p absolute
+            relative = p.startswith('/./')
+            p = os.path.normpath(p)
```
ThomasWaldmann: this transforms /./xxx -> /xxx - we lose the /.

borg/helpers.py
```diff
@@ -835,26 +835,32 @@ def parse(self, text):
...
             self.port = m.group('port') and int(m.group('port')) or None
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
             self.archive = m.group('archive')
             return True
         m = self.file_re.match(text)
         if m:
             self.proto = m.group('proto')
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
```
ThomasWaldmann: not strictly necessary, but also should not do any harm.

borg/helpers.py
```diff
@@ -835,26 +835,32 @@ def parse(self, text):
...
             self.archive = m.group('archive')
             return True
         m = self.scp_re.match(text)
         if m:
             self.user = m.group('user')
             self.host = m.group('host')
-            self.path = os.path.normpath(m.group('path'))
+            self.path = normpath_special(m.group('path'))
```
ThomasWaldmann: also not strictly necessary (foo@host:path is relative), but also does not do harm.

borg/helpers.py
ThomasWaldmann: this will trigger a "repo was previously located at ..." warning once if a user uses a relative path, because the url stored in cache before (old style) is compared with different url generated by updated code (new style).

ThomasWaldmann:
I cancelled the OS X tests, they hung at boot.

borg/helpers.py
```diff
@@ -835,26 +835,32 @@ def parse(self, text):
         return True
 
     def _parse(self, text):
+        def normpath_special(p):
+            # avoid that normpath strips away our relative path hack and even makes p absolute
+            relative = p.startswith('/./')
+            p = os.path.normpath(p)
+            return '/.' + p if relative else p
```
ThomasWaldmann: ok

ThomasWaldmann:
I cancelled the OS X tests (after 1h).

PR #1712, opened 2016-10-13
borg mount --first / --last / --sort / --prefix, fixes #1542

### LINKED ISSUE #1542: --first/last/prefix for fuse versions view
if one has a lot of archives in the repo, maybe even a repo with archives from different sources, merging them all together in a versioned view is not desirable.

so, after #1554 is finished and merged, extend it to `borg mount`.

ThomasWaldmann:
also:
- use consider_part_files when dealing with multiple archives / whole repo mount
- refactor / dedup code

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -142,32 +142,29 @@ class Archives(abc.MutableMapping):
         name = safe_encode(name)
         del self._archives[name]
 
-    def list(self, sort_by=None, reverse=False, prefix=''):
+    def list(self, sort_by=(), reverse=False, prefix='', first=None, last=None):
         """
         Inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
please review, I am doing work that is based on these changes.

src/borg/helpers.py
```diff
@@ -142,17 +142,29 @@ def __delitem__(self, name):
...
-        if reverse:
+        for sortkey in reversed(sort_by):
+            archives.sort(key=attrgetter(sortkey))
+        if reverse or last:
             archives.reverse()
-        return archives
+        n = first or last or len(archives)
+        return archives[:n]
+
+    def list_considering(self, args):
```
enkore: At first wanted to suggest a different name, but this is actually quite nice... `archives.list_considering(args)`

ThomasWaldmann:
thanks!

PR #1718, opened 2016-10-13
point XDG_*_HOME to temp dirs for tests, fixes #1714

### LINKED ISSUE #1714: many files in ~/.config/borg/key-nonces dir
I just deleted 14.500 files from there. I suspect they were generated from borg unit tests...

ThomasWaldmann:
otherwise it spoils the user's nonces and cache dirs with lots of files.

also: remove all BORG_* env vars from the outer environment

TODO: check whether some of this should be applied to 1.0-maint first.

### DIFF
--- a/src/borg/testsuite/nonces.py
+++ b/src/borg/testsuite/nonces.py
@@ -10,17 +10,6 @@ from ..remote import InvalidRPCMethod
 from .. import nonces  # for monkey patching NONCE_SPACE_RESERVATION
 
 
-@pytest.fixture(autouse=True)
-def clean_env(monkeypatch):
-    # Workaround for some tests (testsuite/archiver) polluting the environment
-    monkeypatch.delenv('BORG_PASSPHRASE', False)
-
-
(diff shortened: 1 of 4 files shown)


textshell:
nice. lgtm.

Having this in 1.0-maint would be nice, yes.

src/borg/testsuite/nonces.py
```diff
@@ -10,17 +10,6 @@
...
 
-@pytest.fixture(autouse=True)
-def clean_env(monkeypatch):
-    # Workaround for some tests (testsuite/archiver) polluting the environment
-    monkeypatch.delenv('BORG_PASSPHRASE', False)
-
-
-@pytest.fixture(autouse=True)
-def nonce_dir(tmpdir_factory, monkeypatch):
-    monkeypatch.setenv('XDG_CONFIG_HOME', tmpdir_factory.mktemp('xdg-config-home'))
```
enkore: btw one can actually import fixtures like anything else, it works like you'd expect.

PR #1727, opened 2016-10-16
Add standalone build for android. (using glibc)

textshell:
This build uses pyinstaller in directory mode and ships the libc of the build system.
It currently assumes a debian armhf as build system.

Currently tested on android 4.2.2.

Still needs integration into Vagrant and testing on more modern android versions.

Q: Do we need to explicitly manage the ctypes.find_library workaround or is just ignoring the exceptions ok?

### DIFF
--- a/scripts/borg.android.sh
+++ b/scripts/borg.android.sh
@@ -1,0 +1,45 @@
+#! /bin/bash
+
+if [ -z "$1" ] ; then
+    echo "Usage: $0 <path to pyinstaller result>"
+    exit 1
+fi
+
+cd $1
+
+cp -t . \
+    /lib/ld-linux-armhf.so.3 \
+    /lib/arm-linux-gnueabihf/libc.so.6 \
+    /lib/arm-linux-gnueabihf/libm.so.6 \
+    /lib/arm-linux-gnueabihf/libpthread.so.0 \
+    /lib/arm-linux-gnueabihf/librt.so.1 \
+    /lib/arm-linux-gnueabihf/libutil.so.1 \
+    /lib/arm-linux-gnueabihf/libdl.so.2
+
+# keep in mind that this script needs to run with the very minimal
+# support of an standard android build. So no readlink or anything
+# fancy and not even a bash.
+cat > borg <<'EOF'
+#! /system/bin/sh
+
+ORIGDIR="$(pwd)"
+RELPATH="${0%%/borg}"
+if ! [ -d "$RELPATH" ] ; then
+    echo "$0: can't find installation dir. Use absolute path to help."
+    exit 2
+fi
+cd "$RELPATH"
+BORGDIR="$PWD"
+cd "$ORIGDIR"
+
+# jump straigt to pass2 because reexec doesn't work with explicit ld.so invocation.
+export _MEIPASS2="$BORGDIR"
+
+"$BORGDIR/ld-linux-armhf.so.3" --library-path "$BORGDIR" "$BORGDIR/borg.real" "$@"
+EOF
+
+chmod a+x borg
+
+# the bootloader uses the executable name to find it's packaged resources.
+# But it can fallback to the executable name plus .pkg
+cp borg.real ld-linux-armhf.so.3.pkg
--- a/scripts/borg.android.spec
+++ b/scripts/borg.android.spec
@@ -1,0 +1,46 @@
+# -*- mode: python -*-
+# this pyinstaller spec file is used to build borg binaries on posix platforms
+
+import os, sys
+
+#basepath = '/vagrant/borg/borg'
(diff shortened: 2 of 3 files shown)


scripts/borg.android.sh
```diff
@@ -0,0 +1,45 @@
...
+    /lib/ld-linux-armhf.so.3 \
+    /lib/arm-linux-gnueabihf/libc.so.6 \
+    /lib/arm-linux-gnueabihf/libm.so.6 \
+    /lib/arm-linux-gnueabihf/libpthread.so.0 \
+    /lib/arm-linux-gnueabihf/librt.so.1 \
+    /lib/arm-linux-gnueabihf/libutil.so.1 \
+    /lib/arm-linux-gnueabihf/libdl.so.2
+
+# keep in mind that this script needs to run with the very minimal
+# support of an standard android build. So no readlink or anything
```
ThomasWaldmann: ... of a standard ...

scripts/borg.android.sh
```diff
@@ -0,0 +1,45 @@
...
+# fancy and not even a bash.
+cat > borg <<'EOF'
```
ThomasWaldmann: is the quoting around EOF needed?

scripts/borg.android.sh
```diff
@@ -0,0 +1,45 @@
...
+RELPATH="${0%%/borg}"
+if ! [ -d "$RELPATH" ] ; then
+    echo "$0: can't find installation dir. Use absolute path to help."
+    exit 2
+fi
+cd "$RELPATH"
+BORGDIR="$PWD"
+cd "$ORIGDIR"
+
+# jump straigt to pass2 because reexec doesn't work with explicit ld.so invocation.
```
ThomasWaldmann: "straight" (or rather "directly"?)

scripts/borg.android.sh
```diff
@@ -0,0 +1,45 @@
...
+export _MEIPASS2="$BORGDIR"
+
+"$BORGDIR/ld-linux-armhf.so.3" --library-path "$BORGDIR" "$BORGDIR/borg.real" "$@"
+EOF
+
+chmod a+x borg
+
+# the bootloader uses the executable name to find it's packaged resources.
```
ThomasWaldmann: its

scripts/borg.android.sh
```diff
@@ -0,0 +1,45 @@
...
+# But it can fallback to the executable name plus .pkg
```
ThomasWaldmann: ... plus ".pkg" extension.

(all for the sake of ending the sentence with a dot in a non-confusing way)

scripts/borg.android.spec
```diff
@@ -0,0 +1,46 @@
+# -*- mode: python -*-
+# this pyinstaller spec file is used to build borg binaries on posix platforms
```
ThomasWaldmann: considering we already have another (more generic posix) spec file, maybe rather tell this is about "android" than "posix"?

scripts/borg.android.spec
```diff
@@ -0,0 +1,46 @@
...
+
+import os, sys
+
+#basepath = '/vagrant/borg/borg'
+basepath = '/home/borg/borg/borg'
```
ThomasWaldmann: looks like this is specific to your developer system?
so you will change this to the vagrant path before it gets merged?

scripts/borg.android.spec
```diff
@@ -0,0 +1,46 @@
...
+             cipher=block_cipher)
+
+if sys.platform == 'darwin':
+    # do not bundle the osxfuse libraries, so we do not get a version
+    # mismatch to the installed kernel driver of osxfuse.
+    a.binaries = [b for b in a.binaries if 'libosxfuse' not in b[0]]
+
+pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
+
+exe = EXE(pyz,
```
ThomasWaldmann: looks like the whole difference to the linux/bsd/osx borg.exe spec file is below this line.

could we use the main .spec file and just some variables?

scripts/borg.android.spec
```diff
@@ -0,0 +1,46 @@
...
+
+#basepath = '/vagrant/borg/borg'
+basepath = '/home/borg/borg/borg'
+
+block_cipher = None
+
+a = Analysis([os.path.join(basepath, 'src/borg/__main__.py'), ],
+             pathex=[basepath, ],
+             binaries=[],
+             datas=[],
```
ThomasWaldmann: paperkey stuff was added here recently in the main .spec file.

ThomasWaldmann:
considering that this is soon 1y old and no response, I guess I'm going to close this soon due to inactivity.

PR #1739, opened 2016-10-18
borg check --first / --last / --sort / --prefix, fixes #1663

### LINKED ISSUE #1663: adapt borg check like in PR 1554
See #1554.

Also look if other commands can be done / extended using the same mechanisms.

ThomasWaldmann:
@funkyfuture ^

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -2014,6 +2014,12 @@ class ArchiverCheckTestCase(ArchiverTestCaseBase):
         self.assert_in('Starting archive consistency check', output)
         output = self.cmd('check', '-v', '--archives-only', '--prefix=archive2', self.repository_location, exit_code=0)
         self.assert_not_in('archive1', output)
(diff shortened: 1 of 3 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -2014,6 +2014,10 @@ def test_check_usage(self):
         self.assert_in('Starting archive consistency check', output)
         output = self.cmd('check', '-v', '--archives-only', '--prefix=archive2', self.repository_location, exit_code=0)
         self.assert_not_in('archive1', output)
+        output = self.cmd('check', '-v', '--archives-only', '--first=1', self.repository_location, exit_code=0)
+        self.assert_in('archive1', output)
+        output = self.cmd('check', '-v', '--archives-only', '--last=1', self.repository_location, exit_code=0)
+        self.assert_in('archive2', output)
```
enkore: Also assert not in?

ThomasWaldmann: right. fixing...

PR #1742, opened 2016-10-18
avoid previous_location mismatch, fixes #1741

### LINKED ISSUE #1741: deal with canonicalisation change
From IRC:
```
lfam: Hm, I think the SSH relative path changes in 1.0.8rc1 trigger a relocated repository warning
ThomasWaldmann$ lfam: yes, it is due to a change of canonicalisation
ThomasWaldmann$ old: relpath -> /~/relpath, new: relpath -> /./relpath
lfam$ It breaks automation until the user confirms the new location. But it's okay if that's expected
ThomasWaldmann$ guess a better visible changelog entry is needed
```

ThomasWaldmann:
due to the changed canonicalization for relative pathes in PR #1711 (implement /./ relpath hack),
there would be a changed repo location warning and the user would be asked if this is ok.
this would break automation and require manual intervention, which is unwanted.

thus, we introduce a new cache config version and when automatically upgrading to this
new version, we also fix the previous_location config entry, if it only changed in the
expected way, but still means the same location.

### DIFF
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -10,7 +10,7 @@ from .key import PlaintextKey
 from .logger import create_logger
 logger = create_logger()
 from .helpers import Error, get_cache_dir, decode_dict, int_to_bigint, \
-    bigint_to_int, format_file_size, yes, bin_to_hex
+    bigint_to_int, format_file_size, yes, bin_to_hex, Location
 from .locking import Lock
 from .hashindex import ChunkIndex
 
@@ -140,10 +140,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         with open(os.path.join(self.path, 'files'), 'wb') as fd:
             pass  # empty file
 
-    def _do_open(self):
-        self.config = configparser.ConfigParser(interpolation=None)
-        config_path = os.path.join(self.path, 'config')
-        self.config.read(config_path)
+    def _check_upgrade(self, config_path):
         try:
             cache_version = self.config.getint('cache', 'version')
             wanted_version = 1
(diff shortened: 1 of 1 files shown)


borg/cache.py
```diff
@@ -140,20 +142,39 @@ def create(self):
...
+                        repo_loc = self.repository._location.canonical_path()
+                        rl = Location(repo_loc)
+                        cl = Location(cache_loc)
+                        if cl.proto == rl.proto and cl.user == rl.user and cl.host == rl.host and cl.port == rl.port \
+                           and \
+                           cl.path and rl.path and \
+                           cl.path.startswith('/~/') and rl.path.startswith('/./') and cl.path[3:] == rl.path[3:]:
+                            # everything is same except the expected change in relative path canonicalization,
+                            # update previous_location to avoid warning / user query about changed location:
+                            self.config.set('cache', 'previous_location', repo_loc)
```
enkore: ok, logic looks good to me (\n's could be optimized :) 

Why do we need to change the cache version, though? We could just add a flag that this translation happened - not quite as neat, but it wouldn't "break" cache for earlier releases, which expected 1 in the version field. [Or do we really need to signal this at all? I'm not sure, I'll have another look tomorrow]

ThomasWaldmann: i just wanted to be save in case we ever decide cl.path could be None.
the cache version increase is needed to detect whether the previous_path was made from an old version. we still use /~/ in the new version and must not modify that, only needs to be modified when the old version badly mapped a relative path to /~/.

ThomasWaldmann: yes, a flag "fixed that" in the config would also work. or maybe rename previous_location to just location (and use that as flag and at the same time, fix the bad name).

ThomasWaldmann: hmm, thinking about it again, the flag / version change is not needed if we accept the false positive:

the longish if-clause uses the location from the current commandline (repository._location), so it can detect that change without having to check flag / version first.

If the commandline is ever changed from ssh://u@h/~/p to ssh://u@h/./p, this would result in a false positive and the "repo moved" warning would be suppressed.

I have to admit that I didn't investigate / understand yet why this check is done, I just remember that @jborg introduced this change together with the security fix for the exchange-to-unencrypted-empty-repo serverside attack.

ThomasWaldmann: a difference with / without the flag/version-change is that the false positive only can happen once vs. can always happen.

enkore: I don't think that this is a security feature; the server can remap paths however it likes. So this is a check to avoid ... mixups perhaps? So I think it wouldn't be a problem.

ThomasWaldmann: @enkore so, completely remove the cache version change and not use a flag either? I currently don't see a problem with that.

enkore: Yeah

ThomasWaldmann:
updated & collapsed changesets - I also removed the other minor changes (constant for CACHE_VERSION), so it is now the minimal change needed.

borg/cache.py
```diff
@@ -140,10 +140,7 @@ def create(self):
         with open(os.path.join(self.path, 'files'), 'wb') as fd:
             pass  # empty file
 
-    def _do_open(self):
-        self.config = configparser.ConfigParser(interpolation=None)
-        config_path = os.path.join(self.path, 'config')
-        self.config.read(config_path)
```
ThomasWaldmann: a better (minimal) diff tool for github would be really cool. :-|
the same lines are "added" below again.

borg/cache.py
```diff
@@ -154,6 +151,25 @@ def _do_open(self):
         except configparser.NoSectionError:
             self.close()
             raise Exception('%s does not look like a Borg cache.' % config_path) from None
+        # borg < 1.0.8rc1 had different canonicalization for the repo location (see #1655 and #1741).
```
ThomasWaldmann: this is now done independent of the cache version.
it could be also just done only for cache version 1.

PR #1748, opened 2016-10-19
shortcut hashindex_set by having hashindex_lookup hint about address

rciorba:
Performance improvement extracted from the Robin Hood hashing work #1429 

When inserting a new key `hashindex_set` is basically a lookup for existing key to update followed by a lookup for the address to insert at. By having `hashindex_lookup` provide info about where the lookup stopped we can avoid re-examining the same buckets for an appropriate location to insert at.

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -109,12 +109,16 @@ hashindex_index(HashIndex *index, const void *key)
 }
 
 static int
-hashindex_lookup(HashIndex *index, const void *key)
+hashindex_lookup(HashIndex *index, const void *key, int *skip_hint)
 {
     int didx = -1;
     int start = hashindex_index(index, key);
     int idx = start;
-    for(;;) {
+    int offset;
+    for(offset=0;;offset++) {
+        if (skip_hint != NULL) {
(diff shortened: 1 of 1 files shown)


src/borg/_hashindex.c
```diff
@@ -109,12 +109,15 @@ hashindex_index(HashIndex *index, const void *key)
...
 
 static int
-hashindex_lookup(HashIndex *index, const void *key)
+hashindex_lookup(HashIndex *index, const void *key, int *skip_hint)
 {
     int didx = -1;
     int start = hashindex_index(index, key);
     int idx = start;
-    for(;;) {
+    for(int offset=0;;offset++) {
```
enkore:
```
    src/borg/_hashindex.c: In function ‘hashindex_lookup’:
    src/borg/_hashindex.c:117:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
         for(int offset=0;;offset++) {
```

ThomasWaldmann:
looks good, just fix the compiler issue.

ThomasWaldmann:
thanks!

PR #1752, opened 2016-10-21
implement borgmajor/minor/patch placeholders, fixes #1694

### LINKED ISSUE #1694: borgversionN placeholder
borgversion currently expands to e.g. `1.1.0b3.dev6+ng1030660.d20161007`.
that's correct, but for client / server multiple versions support, it is a bit much.

so, add some more placeholders, that expand like:

borgversion1 1
borgversion2 1.1
borgversion3 1.1.0

With this, I guess people would usually use borgversion2 on the client to select the borg version on the server via --remote-path borg-{borgversion2}. On the server, borg-1.1 would just symlink to the latest 1.1.x release.

ThomasWaldmann:


borg/__init__.py
```diff
@@ -1,3 +1,11 @@
...
 
 from ._version import version as __version__
+
+version_re = r'(\d+)\.(\d+)\.(\d+)'
+
+m = re.match(version_re, __version__)
+if m:
+    __version_tuple__ = tuple(map(int, m.groups()))
+else:
+    raise RuntimeError("Can't parse __version__: %r" % __version__)
```
enkore: Use `distutils.version` instead?

```py
>>> c = version.LooseVersion()

>>> c.parse('1.1.0b3.dev54+ng2f4039f.d20161015')

>>> c.version
[1, 1, 0, 'b', 3, 'dev', 54, '+', 'ng', 2, 'f', 4039, 'f', 'd', 20161015]
```

enkore: Good point. I think we're using it in the xattr code somewhere though.

ThomasWaldmann: ok, I switched to LooseVersion. If we use it anyway better have less code.

PR #1756, opened 2016-10-23
Fix freebsd fuse tests

ThomasWaldmann:


borg/testsuite/archiver.py
```diff
@@ -1097,6 +1088,27 @@ def has_noatime(some_file):
...
+                    assert xattr.getxattr(out_fn, 'user.foo') == b'bar'
+                else:
+                    assert xattr.listxattr(out_fn) == []
+                    try:
+                        xattr.getxattr(out_fn, 'user.foo')
+                    except OSError as e:
+                        assert e.errno == llfuse.ENOATTR
+                    else:
+                        assert False, "expected OSError(ENOATTR), but no error was raised"
+            except OSError as err:
```
ThomasWaldmann: i just moved this "list/read xattrs" block to the end and added try/except to ignore the freebsd issue.

ThomasWaldmann:
merge as is?

enkore:
I guess. I looked at it earlier today, I obviously can't judge by looking at code whether that actually fixes the issue, but you probably checked that :)

ThomasWaldmann:
yes, it worked. the only questionable thing is whether this should work on freebsd or not, but i couldn't find much about that.

PR #1757, opened 2016-10-23
docs/deployment.rst: do not use bare variables in ansible snippet

flokli:
The current snippet throws deprecation warnings:

```
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your 
playbooks so that the environment value uses the full variable syntax 
('{{auth_users}}').
This feature will be removed in a future 
release. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```


ThomasWaldmann:
thanks!

so, after that fix, you don't get these warnings any more (means: you have fixed all such places)?

ThomasWaldmann:
oh, and could you please PR against 1.0-maint - this is the branch the 1.0.x release are made from.

ThomasWaldmann:
superceded by #1758 - will be merged into 1.0-maint (and later from there into master).

Issue #1759, opened 2016-10-24
check remote.py "~user/..." path processing


ThomasWaldmann:
compare master against 1.0-maint and see how /~username is processed.

i suspect it is incorrect in master:
https://github.com/borgbackup/borg/blob/0ff2352691ee82ebf3c828092a68d1558a64f872/src/borg/remote.py#L153

1.0-maint:
https://github.com/borgbackup/borg/blob/1.0-maint/borg/remote.py#L136

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/38636201-check-remote-py-user-path-processing)

enkore:
git blames #759

ThomasWaldmann:
@nachtgeist can you have a look?

ThomasWaldmann:
from old irc logs, see also #745 :
```
16:37  nachtgeist$ i'm working on PR #745. in order for a regular user to be able to use an fstab entry with mount, i had to patch 
helpers.py:get_keys_dir(). problem is, /bin/mount has SUID bit set, which messes up get_keys_dir()'s os.path.expanduser('~') which 
by that time points to /root. that path of course is not readable for the user calling mount, so reading keys fails (and keys should be read from the user's home anyways). so i changed it: https://gith
16:42  nachtgeist$ travis output: https://travis-ci.org/borgbackup/borg/jobs/116138703#L382
16:47  nachtgeist$ ah. got it: os.path.expanduser('~%s' % os.environ.get('USER')) evaluates to the user's home while during testing HOME is set differently
17:39  ThomasWaldmann$ hi nachtgeist
17:39  nachtgeist$ hoi
17:40  ThomasWaldmann$ ah, now it's clearer. maybe add a comment to the src?
17:40  nachtgeist$ sure
17:40  ThomasWaldmann$ and likely that should be a separate function, so it can be used for config, cache and whereever the home dir is needed.
17:42  nachtgeist$ get_home_from_env_or_getent? ← for a lack of imagination. i'm open for suggestions ^^
17:42  nachtgeist$ but i'll do that in a separate PR and update all usages of $HOME. agreed?
17:47  ThomasWaldmann$ as long as we don't need another different get_home: just use get_home()?
17:47  nachtgeist$ k
17:48  nachtgeist$ as for os.environ.get('HOME')
17:48  nachtgeist$ i suggest:
17:48  nachtgeist$ home_dir = os.environ.get('HOME', '')
17:48  nachtgeist$ if not len(home_dir):
[...]

PR #1760, opened 2016-10-24
Merge 1.0 maint

ThomasWaldmann:
see also #1759.

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -149,8 +149,10 @@ class RepositoryServer:  # pragma: no cover
 
     def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, append_only=False):
         path = os.fsdecode(path)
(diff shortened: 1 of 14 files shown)


src/borg/remote.py
```diff
@@ -149,8 +149,10 @@ def negotiate(self, versions):
 
     def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, append_only=False):
         path = os.fsdecode(path)
-        if path.startswith('/~'):
-            path = os.path.join(get_home_dir(), path[2:])
+        if path.startswith('/~'):  # /~/x = path x relative to home dir, /~username/x = relative to "user" home dir
+            path = os.path.join(get_home_dir(), path[2:])  # XXX check this (see also 1.0-maint), is it correct for ~u?
```
enkore: It isn't; fix should be  a separate PR.

PR #1764, opened 2016-10-25
add windows virtual machine

ololoru:
EdgeOnWindows10 is an official Microsoft Image with Microsoft Edge configured as the default browser. See [official](https://atlas.hashicorp.com/Microsoft/boxes/EdgeOnWindows10) vagrant image page.

Requirements:
`vagrant-reload` plugin.

Usage:
- Initial `vagrant up` command fails because of missing `/vagrant/borg/` path.
- Run `vagrant provision` after that, it will automatically install required packages and synchronize folders.

Caveats:
- ~~I was not able to make it working with [x86_64 cygwin version](https://www.cygwin.com/setup-x86_64.exe)~~ Works both for x86_64/x86. X86_84 is by default - see https://www.cygwin.com/faq.html section **4.45. How do I fix fork() failures?** why x86_64 is preferred
- One has to enter password when required
- `llfuse` python module is not available on windows
- pyenv fails to build python from sources so default cygwin python3.4 is used

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -155,6 +155,57 @@ def packages_netbsd
   EOF
 end
 
+# Install required cygwin packages and configure environment
+#
+# Microsoft/EdgeOnWindows10 image has MLS-OpenSSH installed by default,
+# which is based on cygwin x86_64 but should not be used together with cygwin.
+# In order to have have cygwin compatible bash 'ImagePath' is replaced with
+# cygrunsrv of newly installed cygwin
+#
+# supported cygwin versions:
+#   x86_64
+#   x86
+def packages_cygwin(version)
+  setup_exe = "setup-#{version}.exe"
+
+  return <<-EOF
+    mkdir -p /cygdrive/c/cygwin
+    powershell -Command '$client = new-object System.Net.WebClient; $client.DownloadFile("https://www.cygwin.com/#{setup_exe}","C:\\cygwin\\#{setup_exe}")'
+    echo '
+    REM --- Change to use different CygWin platform and final install path
+    set CYGSETUP=#{setup_exe}
+    REM --- Install build version of CygWin in a subfolder
+    set OURPATH=%cd%
+	    set CYGBUILD="C:\\cygwin\\CygWin"
+	    set CYGMIRROR=ftp://mirrors.kernel.org/sourceware/cygwin/
+	    set BUILDPKGS=python3,python3-setuptools,binutils,gcc-g++,libopenssl,openssl-devel,git,make,openssh,liblz4-devel,liblz4_1,rsync,curl,python-devel
+    %CYGSETUP% -q -B -o -n -R %CYGBUILD% -L -D -s %CYGMIRROR% -P %BUILDPKGS%
+    cd /d C:\\cygwin\\CygWin\\bin
+    regtool set /HKLM/SYSTEM/CurrentControlSet/Services/OpenSSHd/ImagePath "C:\\cygwin\\CygWin\\bin\\cygrunsrv.exe"
+    bash -c "ssh-host-config --no"
+	    ' > /cygdrive/c/cygwin/install.bat
+    cd /cygdrive/c/cygwin && cmd.exe /c install.bat
+
+    echo "alias mkdir='mkdir -p'" > ~/.profile
+    echo "export CYGWIN_ROOT=/cygdrive/c/cygwin/CygWin" >> ~/.profile
+    echo 'export PATH=$PATH:$CYGWIN_ROOT/bin' >> ~/.profile
+
+    echo '' > ~/.bash_profile
+
+    cmd.exe /c 'setx /m PATH "%PATH%;C:\\cygwin\\CygWin\\bin"'
+    source ~/.profile
+    echo 'db_home: windows' > $CYGWIN_ROOT/etc/nsswitch.conf
+  EOF
+end
+
+def install_cygwin_venv
+  return <<-EOF
+      easy_install-3.4 pip
+      pip install virtualenv
+  EOF
+end
+
+
 def install_pyenv(boxname)
   return <<-EOF
     curl -s -L https://raw.githubusercontent.com/yyuu/pyenv-installer/master/bin/pyenv-installer | bash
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -150,6 +150,64 @@ def packages_netbsd
...
+    /cygdrive/c/cygwin/CygWin/bin/bash.exe -c "
+      cd $HOME
+      . borg-env/bin/activate
+      cd borg
+      rm -f borg/*.so borg/*.cpy*
+      rm -f borg/{chunker,crypto,compress,hashindex,platform_linux}.c
+      rm -rf borg/__pycache__ borg/support/__pycache__ borg/testsuite/__pycache__
+      pip install -r requirements.d/development.txt
+      pip install -e .
+    "
```
ThomasWaldmann: this somehow looks duplicated, except the bash.exe -c "..." part.

can bash.exe somehow be configured as default shell for this, so this is not needed?

Vagrantfile
```diff
@@ -464,4 +522,34 @@ Vagrant.configure(2) do |config|
...
+    b.ssh.insert_key = false
+
+    b.vm.provider :virtualbox do |v|
+      v.memory = 768
+      #v.gui = true
+    end
+
+    #disable unix like mount
+    b.vm.synced_folder ".", "/vagrant/borg/borg", :disabled => true
+    b.vm.synced_folder ".", "/cygdrive/c/Users/IEUser/borg", :type => "rsync", :rsync__args => ["--verbose", "--archive", "--delete", "-z"]
```
ThomasWaldmann: could it maybe linked, so it appears to be at /vagrant/borg/borg (even if it is not really there)?

Vagrantfile
```diff
@@ -464,4 +522,34 @@ Vagrant.configure(2) do |config|
...
+    b.vm.boot_timeout = 600
+    b.vm.graceful_halt_timeout = 600
+
+    b.ssh.shell = "sh -l"
+    b.ssh.username = "IEUser"
+    b.ssh.password = "Passw0rd!"
+    b.ssh.insert_key = false
+
+    b.vm.provider :virtualbox do |v|
+      v.memory = 768
```
ThomasWaldmann: is this enough for windows?

Vagrantfile
```diff
@@ -150,6 +150,64 @@ def packages_netbsd
   EOF
 end
 
+def packages_cygwin
+  return <<-EOF
+    mkdir -p /cygdrive/c/cygwin
+    powershell -Command '$client = new-object System.Net.WebClient; $client.DownloadFile("https://www.cygwin.com/setup-x86.exe","C:\\cygwin\\setup-x86_64.exe")'
+    echo '
+    REM --- Change to use different CygWin platform and final install path
+    set CYGSETUP=setup-x86_64.exe
```
ThomasWaldmann: you said you didn't get it working with this, so, what does this mean precisely?

ThomasWaldmann:
Ah, and thanks for working on this!

Vagrantfile
ololoru: The actual file downloaed from www.cygwin.com is a first argument of ``$client.DownloadFile`` call, I'll rework it to a parametrized method to avoid confusion.

Vagrantfile
ololoru: Vagrant home is missing in this image, but I think /vagrant/borg/ could be linked to /cydrive/c/Users/IEUser using cygwin mount

Vagrantfile
ololoru: ``sh`` vs ``bash`` mess is caused by mls-openssh ssh server installed there, here it is (related)[https://github.com/joefitzgerald/packer-windows/issues/123] discussion in packer-windows project that provides some details. I'll try to find a way to replace that ssh deamon with a cygwin one, otherwise I'll change default shell client on first run

Vagrantfile
ThomasWaldmann: oh, so you download the 32bit setup and store it under the name that usually the 64bit setup has? don't do that, that is confusing.

ololoru: This function is now parametrized

Vagrantfile
ololoru: Existing install_borg_no_fuse is now re-used

Vagrantfile
ololoru: Default memory size is 2048 now

Vagrantfile
ololoru: Now standard path ``/vagrant/borg/`` is used

Vagrantfile
```diff
@@ -150,6 +150,57 @@ def packages_netbsd
...
+  return <<-EOF
+    mkdir -p /cygdrive/c/cygwin
+    powershell -Command '$client = new-object System.Net.WebClient; $client.DownloadFile("https://www.cygwin.com/#{setup_exe}","C:\\cygwin\\#{setup_exe}")'
+    echo '
+    REM --- Change to use different CygWin platform and final install path
+    set CYGSETUP=#{setup_exe}
+    REM --- Install build version of CygWin in a subfolder
+    set OURPATH=%cd%
+	    set CYGBUILD="C:\\cygwin\\CygWin"
+	    set CYGMIRROR=ftp://mirrors.kernel.org/sourceware/cygwin/
```
ThomasWaldmann: is there also a secure site with TLS?

Vagrantfile
```diff
@@ -150,6 +150,57 @@ def packages_netbsd
...
+	    set BUILDPKGS=python3,python3-setuptools,binutils,gcc-g++,libopenssl,openssl-devel,git,make,openssh,liblz4-devel,liblz4_1,rsync,curl,python-devel
+    %CYGSETUP% -q -B -o -n -R %CYGBUILD% -L -D -s %CYGMIRROR% -P %BUILDPKGS%
+    cd /d C:\\cygwin\\CygWin\\bin
+    regtool set /HKLM/SYSTEM/CurrentControlSet/Services/OpenSSHd/ImagePath "C:\\cygwin\\CygWin\\bin\\cygrunsrv.exe"
+    bash -c "ssh-host-config --no"
+	    ' > /cygdrive/c/cygwin/install.bat
+    cd /cygdrive/c/cygwin && cmd.exe /c install.bat
+
+    echo "alias mkdir='mkdir -p'" > ~/.profile
```
ThomasWaldmann: why -p?

Vagrantfile
```diff
@@ -464,4 +515,31 @@ Vagrant.configure(2) do |config|
     b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg_no_fuse("netbsd64")
     b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("netbsd64")
   end
+
+  config.vm.define "windows10" do |b|
+    b.vm.box = "Microsoft/EdgeOnWindows10"
+    b.vm.guest = :windows
+    b.vm.boot_timeout = 600
+    b.vm.graceful_halt_timeout = 600
```
ThomasWaldmann: 10 minutes for both? how about 3..5 for booting and 2..4 for shutdown?

Vagrantfile
```diff
@@ -464,4 +515,31 @@ Vagrant.configure(2) do |config|
...
+    b.ssh.password = "Passw0rd!"
+    b.ssh.insert_key = false
+
+    b.vm.provider :virtualbox do |v|
+      v.memory = 2048
+      #v.gui = true
+    end
+
+    # fix permissions placeholder
+    config.vm.provision "fix perms", :type => :shell,  :privileged => false, :inline => "echo 'fix permission placeholder'"
```
ThomasWaldmann: this is like a NOP, right? can this line be just removed?

ThomasWaldmann:
about the llfuse failure: if it creates bigger trouble, you can use "sed" to kill it, search the 1.0-maint Vagrantfile for "sed" usage.

Vagrantfile
ololoru: This line is required, otherwise vagrant provisioning fails at "fix perms" step

```
vagrant provision windows10
==> windows10: Running provisioner: fix perms (shell)...
    windows10: Running: inline script
==> windows10: sh: sudo: command not found
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

Vagrantfile
ololoru: ``https://mirrors.kernel.org/sourceware/cygwin/`` also works

ololoru:
tests are run but fail on llfuse installation, to me lines https://github.com/borgbackup/borg/commit/b1ac207e67abff73e875bd5711b43b933a01b6ea#diff-23b6f443c01ea2efcb4f36eedfea9089R126-127 should go under ``install_borg_no_fuse`` in order to avoid copypasta in packages_netbsd, packages_openbsd, and it cygwin case

Vagrantfile
ThomasWaldmann: no, only tls would be interesting.

Vagrantfile
ThomasWaldmann: ok, then please change it, we don't want to wait too long if windows has some troubles.

Vagrantfile
ThomasWaldmann: oh, this is not alternative, but overwriting the unixoid behaviour. ok, then keep it.

Vagrantfile
ololoru: To avoid failing when one has to run vagrant provision windows10 several times because `:reload` action is a part of provisioning

ololoru:
@ThomasWaldmann Do you want me to close this PR and create a new one against  1.0-maint ?

ThomasWaldmann:
If you could do that easily, yes!
Since now, I did manual cygwin testing of 1.0.x, but rather infrequently.
With your change, it could become part of release process.

PR #1770, opened 2016-10-28
remote call to logger that were uninitialized

apre:
replace it with a print() call.

### DIFF
--- a/src/borg/xattr.py
+++ b/src/borg/xattr.py
@@ -68,7 +68,7 @@ if libc_name is None:
         libc_name = 'libc.dylib'
     else:
         msg = "Can't find C library. No fallback known. Try installing ldconfig, gcc/cc or objdump."
-        print(msg,file=sys.stderr)  # logger isn't initialized at this stage
+        print(msg, file=sys.stderr)  # logger isn't initialized at this stage
         raise Exception(msg)
 
 # If we are running with fakeroot on Linux, then use the xattr functions of fakeroot. This is needed by
--- a/src/borg/xattr.py
+++ b/src/borg/xattr.py
@@ -68,7 +68,7 @@ if libc_name is None:
         libc_name = 'libc.dylib'
     else:
         msg = "Can't find C library. No fallback known. Try installing ldconfig, gcc/cc or objdump."
-        logger.error(msg)
+        print(msg,file=sys.stderr)  # logger isn't initialized at this stage
         raise Exception(msg)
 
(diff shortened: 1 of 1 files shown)


src/borg/xattr.py
```diff
@@ -68,7 +68,7 @@ def get_all(path, follow_symlinks=True):
         libc_name = 'libc.dylib'
     else:
         msg = "Can't find C library. No fallback known. Try installing ldconfig, gcc/cc or objdump."
-        logger.error(msg)
+        print(msg, file=sys.stderr)  # logger isn't initialized at this stage
```
ThomasWaldmann: oh, oops.

so you found that in a traceback because it could find your C library and then it crashed when trying to log that?

apre: Hi Thomas, 
This is exactly the case. It occured while I was hacking the codebase for own experiments.
Best regards,

Adrien

ThomasWaldmann:
Hmm, does the problem also exist in 1.0-maint branch?

src/borg/xattr.py
ThomasWaldmann: Ehrm, what is your platform on which you triggered that problem?

ThomasWaldmann: And is there some "usual" name for the libc on that platform? (see code above)

apre: Here is my platform.
    Microsoft Windows [version 10.0.14393]
    (c) 2016 Microsoft Corporation. Tous droits réservés.
    D:\borg>python --version
    Python 3.5.2 :: Anaconda 2.5.0 (32-bit)

I am trying to port borg to a _native_ windows version, that's the reason why the 
 _libc_ is missing :)

enkore: There has been some groundwork in the windows branch so far. With native-native you mean Python for Windows (+ MSVC, not MSYS) I guess?

apre: oh, I have missed that branch fully. 
I will have a look into it.
I was trying to get it working with mingw32 compiler ( not so straightforward way, but it is for fun).

Also thinking about using _paramiko_ as replacement of external ssh pipe.

ThomasWaldmann:
It does. So, please: collapse changesets into 1 and rebase onto 1.0-maint, PR against 1.0-maint.

ThomasWaldmann:
@apre can you do that?

apre:
Sure !
I will do that as soon as I find a free timesslice :) !

ThomasWaldmann:
see #1840.

Issue #1443, opened 2016-08-05
Size of segment can exceed MAX_SEGMENT_SIZE


PostalBlab:
Segments can exceed MAX_SEGMENT_SIZE as explained in constants.py but the name is still misleading.

Please rename MAX_SEGMENT_SIZE or make sure a segment can't be bigger than MAX_SEGMENT_SIZE. Maybe it's only necessary to replace 
`def get_write_fd(self, no_new=False, raise_full=False):`
`if not no_new and self.offset and self.offset > self.limit:`

with something like this

`def get_write_fd(self, no_new=False, raise_full=False):`
`if not no_new and self.offset and (self.offset   + len(data) + self.put_header_fmt.size) > self.limit:`

bronwynbiro:
May I work on this?

enkore:
Sure

ThomasWaldmann:
master branch?

Kashyap12:
hi can I take this up? @PostalBlab , new to open source, would  love to help

PostalBlab:
Sure. I'm not working on it.

shreab373:
beginner here, if the PR was merged, why is the issue still open?

enkore:
It wasn't merged, the PR was closed.

jazcap53:
Apologies for the noob question, but:
Running `grep -nr MAX_SEGMENT_SIZE` my borg root directory, I see only instances of DEFAULT_MAX_SEGMENT_SIZE -- none of MAX_SEGMENT_SIZE by itself.
Does this mean the issue has been repaired?

enkore:
The constant ( (DEFAULT_)MAX_SEGMENT_SIZE ) was always only the default value, which is written into the repository config file when a new repository is created. It was always possible to adjust that after the fact by editing the repository config file.

The principal issue hasn't been 'solved' by the rename.

jazcap53:
Thank you enkore for the explanation.

willberriss:
Hi
[...]

PR #1771, opened 2016-10-28
Changed MAX_SEGMENT_SIZE name

Kashyap12:
Hi changed  MAX_SEGMENT_SIZE name to peripherai_segment_size for issue #1443

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -168,7 +168,7 @@ class Repository:
         config.add_section('repository')
         config.set('repository', 'version', '1')
         config.set('repository', 'segments_per_dir', str(DEFAULT_SEGMENTS_PER_DIR))
-        config.set('repository', 'max_segment_size', str(DEFAULT_MAX_SEGMENT_SIZE))
+        config.set('repository', 'peripheral_segment_size', str(DEFAULT_MAX_SEGMENT_SIZE))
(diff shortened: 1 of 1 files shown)


src/borg/repository.py
```diff
@@ -262,14 +262,14 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
         if 'repository' not in self.config.sections() or self.config.getint('repository', 'version') != 1:
             self.close()
             raise self.InvalidRepository(path)
-        self.max_segment_size = self.config.getint('repository', 'max_segment_size')
+        self.peripheral_segment_size = self.config.getint('repository', 'max_segment_size')
```
ThomasWaldmann: here, you read 'max_segment_size' from the config - which works for current configs, but won't work for configs created by line 171, which would have 'peripheral_segment_size'.

So, this has misc. problems:
- it does not work
- names are less consistent now than before
- "peripheral_segment_size" is not really a better name
- still uses DEFAULT_MAX_SEGMENT_SIZE
- you just changed names, not behaviour.

I'ld suggest to first make a solid plan about what to do about this (if anything) in the ticket, agree on it and then work on the code.

ThomasWaldmann:
Closing this due to reasons mentioned above - let's proceed with a plan in the ticket before retrying with a new PR.

PR #1785, opened 2016-10-31
Merge 1.0-maint

ThomasWaldmann:


src/borg/testsuite/__init__.py
```diff
@@ -163,6 +163,8 @@ def _assert_dirs_equal_cmp(self, diff):
             # Assume path2 is on FUSE if st_dev is different
             fuse = s1.st_dev != s2.st_dev
             attrs = ['st_mode', 'st_uid', 'st_gid', 'st_rdev']
+            if has_lchflags and not ignore_bsdflags:
+                attrs.append('st_flags')
```
ThomasWaldmann: this was a merge conflict. 1.0-maint has this code, master did not have it (why?).

and the travis-ci tests now fails, because there is no st_flags attribute in the stat results.

@enkore you worked on this a while ago, do you know what's going wrong and how it should be?

ThomasWaldmann: ah, i found it. need to use get_flags() to be portable.

PR #1819, opened 2016-11-09
Add BLAKE2b key types

### LINKED ISSUE #1800: Add BLAKE2b ID hash functions
- New key types which use BLAKE2b-256 for ID hash (and MAC?).
- `authenticated`: no encryption
- `repokey|keyfile-blake2`: the usual encryption.

From #1044 

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/38938088-add-blake2b-id-hash-functions)

enkore:
The BLAKE2b implementation is triple licensed under the CC0, OpenSSL and Apache 2.0 licenses, all of which are compatible with our BSD 3 clause.

The BLAKE2b implementation files are copied verbatim from git sha1 https://github.com/BLAKE2/BLAKE2/commit/d4418edab4b9e7595973f876f82d9fecf54b9d19 , completely unmodified to make maintenance as easy as possible.

This actually adds very little new code, since almost everything is reused. Note that the encrypt/decrypt code in AuthenticatedKey is copied from PlaintextKey (enc_key, mac_key are unused but still there, no harm done).

#1034 does not apply perfectly cleanly, but all merge conflicts are the simple kind (import lines in two files, one hunk in the crypto tests where git makes up a conflict where both branches added separate tests, the selftest count).

The error message from old borg looks like this:

```
Unsupported payload type 5. A newer version is required to access this repository.
```

Fixes #1800

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -959,7 +959,7 @@ class Archiver:
         else:
             encrypted = 'Yes (%s)' % key.NAME
         print('Encrypted: %s' % encrypted)
-        if key.NAME == 'key file':
+        if key.NAME.startswith('key file'):
             print('Key file: %s' % key.find_key())
         print('Cache: %s' % cache.path)
         print(DASHES)
@@ -1556,6 +1556,7 @@ class Archiver:
                                                     'Access to all sub-directories is granted implicitly; PATH doesn\'t need to directly point to a repository.')
         subparser.add_argument('--append-only', dest='append_only', action='store_true',
                                help='only allow appending to repository segment files')
+
         init_epilog = textwrap.dedent("""
         This command initializes an empty repository. A repository is a filesystem
         directory containing the deduplicated data from zero or more archives.
@@ -1599,8 +1600,21 @@ class Archiver:
         You can change your passphrase for existing repos at any time, it won't affect
         the encryption/decryption key or other secrets.
 
-        When encrypting, AES-CTR-256 is used for encryption, and HMAC-SHA256 for
-        authentication. Hardware acceleration will be used automatically.
+        Encryption modes
+        ++++++++++++++++
+
+        repokey and keyfile use AES-CTR-256 for encryption and HMAC-SHA256 for
+        authentication in an encrypt-then-MAC (EtM) construction. The chunk ID hash
+        is HMAC-SHA256 as well (with a separate key).
+
+        repokey-blake2 and keyfile-blake2 use the same authenticated encryption, but
+        use a keyed BLAKE2b-256 hash for the chunk ID hash.
+
+        "authenticated" mode uses no encryption, but authenticates repository contents
+        through the same keyed BLAKE2b-256 hash as the other blake2 modes.
+        The key is stored like repokey.
+
+        Hardware acceleration will be used automatically.
         """)
         subparser = subparsers.add_parser('init', parents=[common_parser], add_help=False,
                                           description=self.do_init.__doc__, epilog=init_epilog,
@@ -1611,7 +1625,8 @@ class Archiver:
                                type=location_validator(archive=False),
                                help='repository to create')
         subparser.add_argument('-e', '--encryption', dest='encryption',
-                               choices=('none', 'keyfile', 'repokey'), default='repokey',
+                               choices=('none', 'keyfile', 'repokey', 'keyfile-blake2', 'repokey-blake2', 'authenticated'),
+                               default='repokey',
                                help='select encryption key mode (default: "%(default)s")')
         subparser.add_argument('-a', '--append-only', dest='append_only', action='store_true',
                                help='create an append-only mode repository')
--- a/src/borg/crypto.pyx
+++ b/src/borg/crypto.pyx
@@ -231,6 +231,11 @@ def blake2b_256(key, data):
     md = bytes(32)
     cdef unsigned char *md_ptr = md
 
+    # This is secure, because BLAKE2 is not vulnerable to length-extension attacks (unlike SHA-1/2, MD-5 and others).
+    # See the BLAKE2 paper section 2.9 "Keyed hashing (MAC and PRF)" for details.
+    # A nice benefit is that this simpler prefix-MAC mode has less overhead than the more complex HMAC mode.
+    # We don't use the BLAKE2 parameter block (via blake2s_init_key) for this to
+    # avoid incompatibility with the limited API of OpenSSL.
     blake2b_update_from_buffer(&state, key)
     blake2b_update_from_buffer(&state, data)
 
--- a/src/borg/crypto.pyx
+++ b/src/borg/crypto.pyx
@@ -6,7 +6,7 @@ from cpython.buffer cimport PyBUF_SIMPLE, PyObject_GetBuffer, PyBuffer_Release
 API_VERSION = 3
 
 
-cdef extern from "blake2/blake2b-ref.c":
+cdef extern from "blake2-libselect.h":
     ctypedef struct blake2b_state:
         pass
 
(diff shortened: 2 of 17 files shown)


src/borg/testsuite/key.py
```diff
@@ -193,6 +229,13 @@ def test_assert_id(self, key):
         with pytest.raises(IntegrityError):
             key.assert_id(id, plaintext_changed)
 
+    def test_authenticated_encrypt(self, monkeypatch):
+        monkeypatch.setenv('BORG_PASSPHRASE', 'test')
+        key = AuthenticatedKey.create(self.MockRepository(), self.MockArgs())
+        plaintext = Chunk(b'123456789')
+        encrypted = key.encrypt(plaintext)
+        assert encrypted.endswith(plaintext.data)
+
```
ThomasWaldmann: rather use authenticated = ... here as it is not encrypted?

src/borg/crypto.pyx
```diff
@@ -6,6 +6,15 @@ from cpython.buffer cimport PyBUF_SIMPLE, PyObject_GetBuffer, PyBuffer_Release
 API_VERSION = 3
 
 
+cdef extern from "blake2/blake2b-ref.c":
+    ctypedef struct blake2b_state:
+        pass
+
+    int blake2b_init(blake2b_state *S, size_t outlen) nogil
+    int blake2b_update(blake2b_state *S, const void *input, size_t inlen) nogil
+    int blake2b_final(blake2b_state *S, void *out, size_t outlen) nogil
```
ThomasWaldmann: alternatively, openssl 1.1 compatibility?

src/borg/archiver.py
```diff
@@ -959,7 +959,7 @@ def _info_repository(self, repository, key, cache):
         else:
             encrypted = 'Yes (%s)' % key.NAME
         print('Encrypted: %s' % encrypted)
-        if key.NAME == 'key file':
+        if 'key file' in key.NAME:
```
ThomasWaldmann: key.NAME.startswith('key file') ?

src/borg/crypto.pyx
enkore: Yes, a compatible implementation can be written with anything that gives BLAKE2b, no special features (keying, tree hashing, personalization etc.) needed. This makes it compatible with the OpenSSL API (L. 237). Python 3.6 hashlib has the full API and is also compatible


However, to be very sure about that, I'll produce a small test program against OpenSSL 1.1

enkore:
> However, to be very sure about that, I'll produce a small test program against OpenSSL 1.1

btdt, turns out OpenSSL 1.1 only supports specifically BLAKE2b-512 and BLAKE2s-256.

Since Python 3.6 on the other hand has the complete BLAKE2 API, including "digest_size".

Two options:

1. Use BLAKE2b-512 and truncate to 256 bits, to retain the option to change this code to OpenSSL 1.1
2. Leave as is. OpenSSL only becomes usable as they update their API: https://github.com/openssl/openssl/issues/980

enkore:
CPython 3.6 works

```
Input:      037fb9b75b20d623f1d5a568050fccde4a1b7c5f5047432925e941a17c7a2d0d7061796c6f6164
Expected:   a22d4fc81bb61c3846c334a09eaf28d22dd7df08c9a7a41e713ef28d80eebd45
Calculated: a22d4fc81bb61c3846c334a09eaf28d22dd7df08c9a7a41e713ef28d80eebd45
Identical:  True

Input:      616263
Expected:   bddd813c634239723171ef3fee98579b94964e3bb1cb3e427262c8c068d52319
Calculated: bddd813c634239723171ef3fee98579b94964e3bb1cb3e427262c8c068d52319
Identical:  True

Input:      e944973af2256d4d670c12dd75304c319f58f4e40df6fb18ef996cb47e06367631323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930313233343536373839303132333435363738393031323334353637383930
Expected:   97ede832378531dd0f4c668685d166e797da27b47d8cd441e885b60abd5e0cb2
Calculated: 97ede832378531dd0f4c668685d166e797da27b47d8cd441e885b60abd5e0cb2
Identical:  True
```

ThomasWaldmann:
What we also have to consider: linux (and other) package maintainers dislike it if one bundles 3rd party code.

So: is there something like a widespread libblake2b package we could use alternatively?

enkore:
I didn't find any. There is libb2 but it's not commonly packaged (Debian: testing, Arch: No, Fedora: No, ...). It seems plausible that this is also why other software including CPython bundles the references implementation instead of using a library; there just isn't any.

enkore:
I looked at whether the `-march` compiler option might ruin performance for generic distribution builds, which is not the case (somewhat surprisingly the performance of the C implementation does not stem from compiler optimizations).

ThomasWaldmann:
http://packages.ubuntu.com/xenial/libdevel/libb2-dev it's in xenial.
https://github.com/BLAKE2/libb2/issues/5

So, I think we should alternatively link to that, so packagers who feel uncomfortable about bundled stuff can rip it out and link to their libb2 package (which they will have to create in case it is not there yet).

enkore:
Done. Works, and the tests pass, too.

src/borg/blake2/libselect.h
```diff
@@ -0,0 +1,5 @@
+#ifdef BORG_USE_LIBB2
+#include <blake2.h>
+#else
+#include "blake2b-ref.c"
+#endif
```
ThomasWaldmann: add a \n at eof.

ThomasWaldmann: hmm, as this is rather part of borg than of blake2 ref implementation, move it to outside of this dir, so maintainers can just kill the whole blake2/ directory to remove it from the package if they use libb2?

enkore: uh okay

ThomasWaldmann:
https://github.com/borgbackup/borg/pull/1819/commits/4074ea0fd779935f738355debb23659b4826c3d1 that looks strange on github.

enkore:
Accidentally committed empty file; removed in fixup.

ThomasWaldmann:
looks good. collapse some, merge.

PR #1820, opened 2016-11-10
WIP Test multiple filesystems with Vagrant

### LINKED ISSUE #1289: Test on many file systems automatically

ThomasWaldmann:
Fixes #1289, continuation from #1359. 

Both kernel module and FUSE filesystems are supported.

TODO:
- Filesystems requiring a device
  - [x] ext4
  - [x] XFS
  - [x] btrfs
  - [ ] NTFS - implemented, but mode/atime/teardown issues
- Filesystems requiring a server
  - [ ] CIFS - implemented, but badly breaks py.test
  - [ ] SSHFS - implemented, but atime/teardown issues
- Filesystems requiring neither
  - [ ] tmpfs - runs out of memory
- Fix FS specific test issues
  - [x] #1361 - test_non_ascii_acl only failed outside of Vagrant due to übel user missing
  - [x] #1362 - tests failing on SSHFS (fixed with #1373)
  - [x] #1363 - tests failing on ntfs-3g (fixed with #1373 + #1383)
- Platform specific Vagrant testing
  - Linux
    - [x] centos7_64
    - [ ] centos6_32
    - [ ] centos6_64
    - [x] xenial64
    - [x] trusty64
    - [ ] jessie64
    - [ ] wheezy32
    - [ ] wheezy64
  - OS X
    - [x] darwin64
  - BSD
    - [ ] freebsd64
    - [ ] openbsd64
    - [ ] netbsd64

Notes:
- no fakeroot, everything runs as root

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Borg authors ("The Borg Collective")
 - Daniel Reichelt <hacking@nachtgeist.net>
 - Lauri Niskanen <ape@ape3000.com>
 - Abdel-Rahman A. (Abogical)
+- Lee Bousfield <ljbousfield@gmail.com>
 
 Borg is a fork of Attic.
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,7 +36,7 @@ def packages_debianoid
     # for building python:
     apt-get install -y zlib1g-dev libbz2-dev libncurses5-dev libreadline-dev liblzma-dev libsqlite3-dev
     # filesystem drivers
-    apt-get install -y ntfs-3g openssh-server sshfs xfsprogs samba cifs-utils
+    apt-get install -y ntfs-3g openssh-server sshfs xfsprogs btrfs-tools samba cifs-utils
     # this way it works on older dists (like ubuntu 12.04) also:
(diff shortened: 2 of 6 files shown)


enkore:
What's the status on this?

ThomasWaldmann:
I am currently rebasing / fixing the conflicts / testing this.

One fundamental thing with this changeset is that it switches vagrant provisioning to be all-root (privileged = true). We get rid of fakeroot by this, but not sure if we run into new problems (besides never being able to run non-root).

PR #1826, opened 2016-11-10
Repository: rollback's cleanup parameter is internal.

textshell:
The rollback‘s cleanup parameter should not be exposed externally, esp. not via rpc.

Thus remove cleanup from public rollback method and have that one call a internal version that allows finer control and is used internally in Repository when needed.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -147,7 +147,7 @@ class Repository:
                 cleanup = True
             else:
                 cleanup = False
-            self.rollback(cleanup)
+            self._rollback(cleanup=cleanup)
         self.close()
 
     @property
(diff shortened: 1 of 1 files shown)


src/borg/repository.py
```diff
@@ -731,14 +731,17 @@ def report_error(msg):
             logger.info('Completed repository check, no problems found.')
         return not error_found or repair
 
-    def rollback(self, cleanup=False):
+    def _rollback(self, *, cleanup):
```
enkore: Why the star-argument?

textshell: I think it‘s usually a good idea to force boolean arguments to be keyword only.

enkore: Oh! That's a neat trick right there :)

PR #1827, opened 2016-11-10
.gitattributes: Set python diff mode for all *.py files.

textshell:
Some tools like gitk need this for sensible word diffs.


ThomasWaldmann:
1.0-maint?

textshell:
redone for 1.0-maint

PR #1839, opened 2016-11-12
osxfuse non-beta

ThomasWaldmann:


enkore:
Does this also need changes in the Vagrantfile?

ThomasWaldmann:
No, vagrant just downloads directly the release from the osxfuse project (not caskroom).

ThomasWaldmann:
Note: it seems that the correct name as of now is still osxfuse-beta, but their file is somehow unparseable.

I file a bug there, let's see...

I also noted there that it is not beta any more, so maybe they'll change the name?

Issue #909, opened 2016-04-15
file_known_and_unchanged is wasteful and should be converted to inodes


verygreen:
Moved from #908 
I guess another change is to turn cache.file_known_and_unchanged into just inode_known_and_unchanged.
Who cares what's the actual file name, since it's always stored in the archive anyway I believe?
Additional benefit, the "files_cache" huge file in .cache might become smaller as it no longer needs to hold path names.

This poses some difficulties for filesystems without inode numbers like vfat and on Windows, as mentioned by @ThomasWaldmann 

Windows (assuming they don't have inodes - I don't know much about ntfs) - some other unique identifier might be used, I guess. For filesystems that have no symlinks (like vfat) - file path might be unique enough. Still some races with content changing underneath, but at least nothing security critical (also hopefully nobody resurrects umsdos fs) (it's tempting to use filesize + create date (+modification date?) as unique id for vfat to handle renames more optimally there, but that might not be safe).
On nfs there's "file handle" that's unique (no idea if it's easily visible from userspace).
fuse and friends are harder, I guess.
ntfs has symlinks and hardlinks, so they must have have unique file ids similar to inode numbers of some sort.
These strategies could be determined by filesystem type (available from statfs(2) and could be called once we cross a mountpoint boundary assuming this is enabled, also separately checked for every path).
It should be noted that since inodes are not unique outside of a normal filesystem, really inode in the table should be some combination of actual inode number and filesystem id (similar to how nfs creates file handles). Filesystem id might be anything from mount path (unsafe) to underlying block device id (unsafe) to fs uuid (not bullet proof either) to some other options.

Using file path as generic fallback for inode number for cases when inode numbers are ignored for other reasons (commandline switch) is also possible, though it should be marked as "potentially suboptimal".

ThomasWaldmann:
Currently, a mapping is used from key path_hash -> value FileCacheEntry (size, mtime, inode, age, chunk_ids).

key: path_hash = self.key.id_hash(safe_encode(os.path.join(self.cwd, path)))

So, the key hash could be just computed from different input data or get replaced by a tuple.

Algorithm:
```
if has_stable_inodes(path, st):
    fs_id = fs_uuid or blockdev_id or mount_path
    key = (fs_id, st.st_ino)
else:
    key = id_hash(path)  # fallback, see orig. code
```

enkore:
[...]

PR #1842, opened 2016-11-13
WIP: fs type detection for linux, dev/inode cache keys

ThomasWaldmann:
fstype(path) -> determine filesystem type (at least the most important ones with stable inodes)
has_stable_inodes(path) -> True/False

Use for files cache.

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -15,14 +15,17 @@ from .hashindex import ChunkIndex, ChunkIndexEntry
 from .helpers import Location
 from .helpers import Error
 from .helpers import get_cache_dir
-from .helpers import decode_dict, int_to_bigint, bigint_to_int, bin_to_hex
+from .helpers import safe_encode, decode_dict, int_to_bigint, bigint_to_int, bin_to_hex
 from .helpers import format_file_size
 from .helpers import yes
 from .item import Item, ArchiveItem
 from .key import PlaintextKey
 from .locking import Lock
 from .platform import SaveFile
+from .platform import fstype
 from .remote import cache_if_remote
+
+FS_WITH_STABLE_INODES = {'extfs', 'btrfs', 'xfs', 'zfs', }
 
 ChunkListEntry = namedtuple('ChunkListEntry', 'id size csize')
 FileCacheEntry = namedtuple('FileCacheEntry', 'age inode size mtime chunk_ids')
@@ -452,25 +455,34 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         else:
             stats.update(-size, -csize, False)
 
-    def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
+    def file_cache_key(self, hash, path, st, ignore_inode):
+        if ignore_inode or fstype(path) not in FS_WITH_STABLE_INODES:
+            # we don't use the path directly (but its hash) to save memory
+            cache_key = hash(safe_encode(path))
+        else:
+            # we don't use the tuple directly (but packed) to save memory
+            cache_key = msgpack.packb((st.st_dev, st.st_ino))
+        return cache_key
+
+    def file_known_and_unchanged(self, cache_key, st, ignore_inode=False):
         if not (self.do_files and stat.S_ISREG(st.st_mode)):
             return None
         if self.files is None:
             self._read_files()
-        entry = self.files.get(path_hash)
+        entry = self.files.get(cache_key)
         if not entry:
             return None
         entry = FileCacheEntry(*msgpack.unpackb(entry))
         if (entry.size == st.st_size and bigint_to_int(entry.mtime) == st.st_mtime_ns and
                 (ignore_inode or entry.inode == st.st_ino)):
-            self.files[path_hash] = msgpack.packb(entry._replace(age=0))
+            self.files[cache_key] = msgpack.packb(entry._replace(age=0))
             return entry.chunk_ids
         else:
             return None
 
-    def memorize_file(self, path_hash, st, ids):
+    def memorize_file(self, cache_key, st, ids):
         if not (self.do_files and stat.S_ISREG(st.st_mode)):
             return
         entry = FileCacheEntry(age=0, inode=st.st_ino, size=st.st_size, mtime=int_to_bigint(st.st_mtime_ns), chunk_ids=ids)
-        self.files[path_hash] = msgpack.packb(entry)
+        self.files[cache_key] = msgpack.packb(entry)
         self._newest_mtime = max(self._newest_mtime, st.st_mtime_ns)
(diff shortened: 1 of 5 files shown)


enkore:
Hmmm. I'm not sure that this is a good direction, because it isn't the average platform specific code, it's highly platform specific and it's hard to test / maintain. I'd say for #909 it would be much easier to make it a flag or env var. Not only because this seems quite difficult to get right, but also because only a few cases really profit from #909 -- things contained in the same archive are already indexed through their (dev, ino) for hardlink handling, which I'd expect to be a more typical use case than the rsync/hardlink importer.

ThomasWaldmann:
Well, at first I also rather disliked the slightly different APIs, header files, etc. on different platforms. But then I realized that we don't need to support all the platforms / all the filesystems. If a platform is not supported, it will just do it based on the path, as before. That's also the reason why I only whitelisted a few, popular filesystems for linux.

Also, I realized that giving cmdline options to control inode vs. path only works if all source filesystems work in the same way and not if you backup a stable-inode fs that has a unstable-inode fs mounted somewhere.

Yes, the main motivation for me to do this was the rsnapshot/rsync+hardlink importer.

There are some other motivations, though:
- keys: packb((dev, ino)) is likely smaller/faster than sha256(path).
- renamed dirs / differently mounted filesystems: path-based keys can't cope with that, no problem for inode-based. Guess this was the main motivation why #909 was filed back then.

enkore:
Ok let's try it

src/borg/cache.py
```diff
@@ -15,15 +15,18 @@
...
 from .helpers import format_file_size
 from .helpers import yes
 from .item import Item, ArchiveItem
 from .key import PlaintextKey
 from .locking import Lock
 from .platform import SaveFile
+from .platform import fstype
 from .remote import cache_if_remote
 
+FS_WITH_STABLE_INODES = {'extfs', 'btrfs', 'xfs', 'zfs', }
```
ThomasWaldmann: hmm, move this to platform code?

guess it depends on the platform's implementation of a filesystem whether it has stable inodes?

src/borg/cache.py
```diff
@@ -452,25 +455,34 @@ def chunk_decref(self, id, stats):
         else:
             stats.update(-size, -csize, False)
 
-    def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
+    def file_cache_key(self, hash, path, st, ignore_inode):
+        if ignore_inode or fstype(path) not in FS_WITH_STABLE_INODES:
```
ThomasWaldmann: the fstype() call could be done less frequently (only at fs boundaries) and the result kept somewhere.

src/borg/cache.py
enkore: Maybe make the platform API just more specific, eg. `fs_inodes_stable(path)`

src/borg/cache.py
```diff
@@ -452,25 +455,34 @@ def chunk_decref(self, id, stats):
...
+            # we don't use the path directly (but its hash) to safe memory
+            cache_key = hash(safe_encode(path))
```
enkore: So this would be incompatible with existing files caches (incompatible in the sense of rechunking / data is worthless). Since the keys are different it would also ~double the on-disk size until 10 (cache TTL) archives are created.

ThomasWaldmann: ad 1: that's the price. can be avoided by either killing the files cache or using the env var to decrease kept generations.

ad 2: hash is a param of this function, not the builtin.

src/borg/cache.py
```diff
@@ -452,25 +455,34 @@ def chunk_decref(self, id, stats):
         else:
             stats.update(-size, -csize, False)
 
-    def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
+    def file_cache_key(self, hash, path, st, ignore_inode):
+        if ignore_inode or fstype(path) not in FS_WITH_STABLE_INODES:
+            # we don't use the path directly (but its hash) to safe memory
```
ThomasWaldmann: oops.

src/borg/cache.py
ThomasWaldmann: ok.

ThomasWaldmann:
fixed & rebased.

ThomasWaldmann:
the fstype detection could be also interesting to skip some tests when some feature is not supported or known-broken on some filesystem. See e.g. the atime trouble in #1820.

ThomasWaldmann:
any other feedback on current code before I rebase / resolve conflicts?

enkore:
lgtm

ThomasWaldmann:
man statfs -> f_fsid (and read the "below"). seems like made for what we want.

it still sucks as everybody uses a different include file, long vs 2 ints, etc. :(

src/borg/platform/linux.pyx
```diff
@@ -261,6 +261,7 @@ def umount(mountpoint):
 cdef extern from "sys/statfs.h":
     struct statfs_t "statfs":
         long f_type
+        # fsid_t f_fsid
```
ThomasWaldmann: I could not get this working, did always get compiler errors here ...

src/borg/platform/linux.pyx
```diff
@@ -287,8 +288,13 @@ def fstype(path=None, fd=None):
         raise ValueError('path or fd must be given')
     if rc == -1:
         raise OSError(errno.errno, strerror(errno.errno).decode(), os.fsdecode(path))
-    return MAGIC_TO_NAME.get(buf.f_type)
+    return dict(
+        fstype=MAGIC_TO_NAME.get(buf.f_type),
+        #fsid=...,
```
ThomasWaldmann: or there.

textshell:
random idea from irc: Have a external programm to query for maj:min and ship an example / contrib that uses blkid to get the info. Would only work as root. But most backups that care about these things run as root anyway.

That would allow us to just use something that really is able to look into file systems etc and is maintained by people who work with low-level file system aspects. Also it does have a lot more bits, so it should be safer.

On Linux the script could do something like this ($1 beeing the device node):
`blkid -o export $1 | grep -e "^UUID" -e "^TYPE" | sha1sum`

enkore:
blkid probably wants a device path

ThomasWaldmann:
another issue that came up:
assume files cache has key (dev_uuid, inode_no) -> value (mtime_ns, size).
assume a specific disk has a file "foo" sitting on inode X of mtime M and size S and content Cfoo.
now we delete foo and write "bar" with content Cbar != Cfoo.
inode X gets reused, size is S and mtime is M (could happen if "foo" modify, delete and creation of "bar" happens within mtime granularity).

Then the modification check would assume foo == bar, which is not the case.

ThomasWaldmann:
I opened #3946 to refer to this stale / blocked PR, so it can be closed.

PR #1844, opened 2016-11-13
check: catch two kinds of different repos

### LINKED ISSUE #1815: check crashes if repo is valid but completely empty
This is because of

```py
    def identify_key(self, repository):
        try:
            some_chunkid, _ = next(self.chunks.iteritems())
        except StopIteration:
            # repo is completely empty, no chunks
            return None                                             <--
        cdata = repository.get(some_chunkid)
```

It later tries to use `self.key` thereby crashes.

I think this case should either abort and tell to delete repository & start over, or basically do a re-initialize.

enkore:
64a3fa8: no segments, no index. (or no committed segments and no index, if we'd have no committed segments but an index we can usually recover, the stick-a-commit-on-it technique).

2261709: no objects

Fixes #1815

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -833,6 +833,9 @@ class ArchiveChecker:
         self.repair = repair
         self.repository = repository
         self.init_chunks()
+        if not self.chunks:
+            logger.error('Repository contains no apparent data at all, cannot continue check/repair.')
+            return False
         self.key = self.identify_key(repository)
         if Manifest.MANIFEST_ID not in self.chunks:
             logger.error("Repository manifest not found!")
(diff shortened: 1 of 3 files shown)


borg/archive.py
```diff
@@ -833,6 +833,9 @@ def check(self, repository, repair=False, archive=None, last=None, prefix=None,
         self.repair = repair
         self.repository = repository
         self.init_chunks()
+        if not self.chunks:
+            logger.error('Repository contains no apparent data at all, cannot continue check/repair.')
```
ThomasWaldmann: Does this trigger for a freshly initialized repo or a repo that has all archives removed again?
Then it would be nice to also output 'For a freshly initialized repo or a repo without archives, this is not an error.'.

enkore: No, since there would be a manifest.

ThomasWaldmann: ok.

PR #1845, opened 2016-11-13
merge 1.0-maint

enkore:


.travis/install.sh
```diff
@@ -18,7 +18,7 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     brew install xz  # required for python lzma module
     brew outdated pyenv || brew upgrade pyenv
     brew install pkg-config
-    brew install Caskroom/versions/osxfuse-beta
+    brew install Caskroom/versions/osxfuse
```
ThomasWaldmann: while this was a at first unsuccessful attempt to fix the issue, a brew maintainer now said he will remove the -beta and it will be just osxfuse in future. so this is correct in the end.

src/borg/testsuite/archiver.py
```diff
@@ -2154,6 +2154,41 @@ def test_verify_data(self):
...
+                repository.delete(id_)
+            repository.commit()
+        self.cmd('check', self.repository_location, exit_code=1)
+
+    def test_attic013_acl_bug(self):
+        # Attic up to release 0.13 contained a bug where every item unintentionally received
+        # a b'acl'=None key-value pair.
+        # This bug can still live on in Borg repositories (through borg upgrade).
+        class Attic013Item:
+            as_dict = lambda: {
```
ThomasWaldmann: ./src/borg/testsuite/archiver.py:2169:13: E731 do not assign a lambda expression, use a def

Issue #1721, opened 2016-10-14
Review progress displays


enkore:
- [x] No progress on txn initialization (cache.begin_txn), or commit (only compact_segments) -- these could at least say that it's doing _something_
   - [x] Transaction initialization (begin_txn), #1881
   - [x] Transaction commit, #1874
- [x] create -C is a bit shy about itself - apparent progress just stops and then continues after it's done --
    This should be done now with the compact_segments progress and the cache progress.
- [x] Cache sync has custom display that only shows with -v, but is silent with -p, #1892
- [x] Has prune proper output with -p?
- [x] Has check proper output for archives with -p?
- [x] ~~Others?~~ **If someone has specific suggestions, don't hesitate to comment**
- [x] Possibly helpful to generally improve progress display; add a current_something to ProgressPercentIndicator that is displayed after the message, similar to the Stats used by create. This could then be used to indicate current archive in cache sync, current file in extract etc. etc. (#1846)

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/39053475-review-progress-displays)

ThomasWaldmann:
Related: #1107

Abogical:
> Cache sync has custom display that only shows with -v, but is silent with -p

What if progress output is implemented as a logger level? would that be a good idea?

enkore:
The current output is kinda neat for it's detailed timing information, so maybe leave that in, but at debug level. Normal progress could just be the percentage of synced archives + current archive name as the trailing part (like filename).

ThomasWaldmann:
Increased the bounty, quite a lot of work.

Abogical:
> Has check proper output for archives with -p?

I think it does, what's missing?

Abogical:
> - [ ] Everything that does something on the repo should accept --progress, because at the very least the repo has it. So it should just be a common option that you can always give.

After #1912 is merged, the remaining commands that edit the repository and do not accept `--progress` are:
- `init`
- `change-passphrase`
- `rename`

These are cheap commands don't take alot of time, thus implementing `--progress` isn't useful.

Please tell me if I'm missing something.

> - [ ] Others?

Please give examples.

enkore:
I believe this is done now. Great work @Abogical :)

Abogical:
That was fun to work on, thanks for the bounty :)

Abogical:
Also don't forget to respond to my claim. thank you.

PR #1846, opened 2016-11-13
Improve extract progress display, for #1721

Abogical:
This changes the extract progress output to include the current file being extracted.
```bash
$ borg extract rep::archive -p
 42.0% Extracting: /answer/to/life/...verse/and/everything
```
This is made possible by making `ProgressIndicator` more flexible, it can take as many arguments for it's message format.

As suggested in #1721.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -31,7 +31,7 @@ from .helpers import format_time, format_timedelta, format_file_size, file_statu
 from .helpers import safe_encode, safe_decode, make_path_safe, remove_surrogates, swidth_slice
 from .helpers import decode_dict, StableDict
 from .helpers import int_to_bigint, bigint_to_int, bin_to_hex
-from .helpers import ProgressIndicatorPercent, log_multi
+from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
 from .helpers import PathPrefixPattern, FnmatchPattern
 from .helpers import consume, chunkit
 from .helpers import CompressionDecider1, CompressionDecider2, CompressionSpec
@@ -93,11 +93,7 @@ class Statistics:
                     msg = ''
                     space = columns - swidth(msg)
                 if space >= 8:
-                    if space < swidth('...') + swidth(path):
-                        path = '%s...%s' % (swidth_slice(path, space // 2 - swidth('...')),
-                                            swidth_slice(path, -space // 2))
-                    space -= swidth(path)
-                    msg += path + ' ' * space
+                    msg += ellipsis_truncate(path, space)
             else:
                 msg = ' ' * columns
             print(msg, file=stream or sys.stderr, end="\r", flush=True)
@@ -448,7 +444,7 @@ Number of files: {0.stats.nfiles}'''.format(
             if 'chunks' in item:
                 for _, data in self.pipeline.fetch_many([c.id for c in item.chunks], is_preloaded=True):
                     if pi:
-                        pi.show(increase=len(data))
+                        pi.show(increase=len(data), info=[remove_surrogates(item.path)])
                     if stdout:
                         sys.stdout.buffer.write(data)
                 if stdout:
@@ -501,7 +497,7 @@ Number of files: {0.stats.nfiles}'''.format(
                 ids = [c.id for c in item.chunks]
                 for _, data in self.pipeline.fetch_many(ids, is_preloaded=True):
                     if pi:
-                        pi.show(increase=len(data))
+                        pi.show(increase=len(data), info=[remove_surrogates(item.path)])
                     with backup_io():
                         if sparse and self.zeros.startswith(data):
                             # all-zero chunk: create a hole in a sparse file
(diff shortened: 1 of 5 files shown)


src/borg/testsuite/helpers.py
```diff
@@ -904,6 +904,8 @@ def test_yes_env_output(capfd, monkeypatch):
 
 
 def test_progress_percentage_sameline(capfd):
+    os.environ['COLUMNS'] = '4'
+    os.environ['LINES'] = '1'
```
ThomasWaldmann: hmm, why?

src/borg/archive.py
```diff
@@ -93,11 +93,7 @@ def show_progress(self, item=None, final=False, stream=None, dt=None):
                     msg = ''
                     space = columns - swidth(msg)
                 if space >= 8:
-                    if space < swidth('...') + swidth(path):
-                        path = '%s...%s' % (swidth_slice(path, space // 2 - swidth('...')),
-                                            swidth_slice(path, -space // 2))
-                    space -= swidth(path)
-                    msg += path + ' ' * space
+                    msg += ellipsis_truncate(path, space)
```
ThomasWaldmann: nice refactor

src/borg/helpers.py
```diff
@@ -1191,6 +1192,20 @@ def yes(msg=None, false_msg=None, true_msg=None, default_msg=None,
         env_var_override = None
 
 
+def ellipsis_truncate(msg, space):
+    from .platform import swidth
+    """
+    shorten a long string by adding ellipsis between it and return it, example:
+    this_is_a_very_long_string -------> this_is..._string
+    """
```
ThomasWaldmann: please put the docstring right under the def, not into the middle of the function.

was the switdth import circular or why don't you do it on module level?

src/borg/helpers.py
```diff
@@ -1239,14 +1253,23 @@ def progress(self, current=None, increase=1):
             self.trigger_at += self.step
             return pct
 
-    def show(self, current=None, increase=1):
+    def show(self, current=None, increase=1, info=[]):
```
ThomasWaldmann: please use info=None as default and check for "is not None" below.

if you use a mutable datatype (like []) as default, checkers will complain as this can be problematic.

src/borg/helpers.py
```diff
@@ -1239,14 +1253,23 @@ def progress(self, current=None, increase=1):
...
         pct = self.progress(current, increase)
         if pct is not None:
+            # truncate the last argument, if space is available
```
ThomasWaldmann: ?

src/borg/helpers.py
```diff
@@ -1239,14 +1253,23 @@ def progress(self, current=None, increase=1):
...
+                else:
+                    info[-1] = ellipsis_truncate(info[-1], space)
+                return self.output(self.msg % (pct, *info))
+
             return self.output(self.msg % pct)
 
     def output(self, message):
-        self.output_len = max(len(message), self.output_len)
-        message = message.ljust(self.output_len)
+        message = message.ljust(get_terminal_size(fallback=(4, 1))[0])
```
ThomasWaldmann: what the reasoning behind (4, 1)?

src/borg/helpers.py
enkore: Or an empty tuple `(,)`, which might save you an otherwise unneeded if.

ThomasWaldmann:
btw, i see you do this PR from your master branch. It is better for your workflow if you use a feature branch for that stuff, so you don't block your master branch.

src/borg/helpers.py
Abogical: I imported swidth in the same way swidth_slice does in the same file (line 1847)

src/borg/testsuite/helpers.py
ThomasWaldmann: ok, so if what you want is a temporary change of these env vars for this test, look at other places in the src where monkeypatch is used to do that.

src/borg/helpers.py
```diff
@@ -1191,6 +1192,22 @@ def yes(msg=None, false_msg=None, true_msg=None, default_msg=None,
...
+def ellipsis_truncate(msg, space):
+    """
+    shorten a long string by adding ellipsis between it and return it, example:
+    this_is_a_very_long_string -------> this_is..._string
+    """
+    from .platform import swidth
+    ellipsis_width = swidth('...')
+    msg_width = swidth(msg)
+    if space < 8:
+        return '...' + ' ' * (space-ellipsis_width)
```
ThomasWaldmann: pep8 suggests blanks around operators, so it is (space - ellipsis_width).

Abogical:
tox is finally happy now

Abogical:
pep8, wth? this is irrelevant to my changes.

screenshot from 2016-11-14 18-57-26

enkore:
Yes, you can ignore those.

Abogical:
I think I'm done here, I'll look onto the the rest of #1721

ThomasWaldmann:
collapsing some commits first would have been prettier. :)

enkore:
Dang, I wanted to make a squash merge :/

Abogical:
Your welcome :)

PR #1864, opened 2016-11-19
fix determination of newest mtime, fixes #1860

### LINKED ISSUE #1860: mtime cache misbehaving on nested backups
I am seeing borg mtime cache consistently become ineffective during my backups.  I am suspecting it has something to do with how these are structured.  This is just a guess, but I am hoping somebody will recognize the pattern.

Given file hierarchy:
```
/foo
/foo/a
/foo/b/
```

I am backing it up as 3 archives:
* /foo/a 
* /foo/b
* then /foo <- here borg cache becomes ineffective and it rescans all the files

I am running with `export BORG_FILES_CACHE_TTL=10000` 
Borg Version: 1.0.7-0ubuntu1.16.04.1

This seems related to #1430 but the fix for that was already included in 1.0.7

ThomasWaldmann:
bug:
if no files were added/modified, _newest_mtime stayed at its initial 0 value.
when saving the files cache, this killed all age 0 entries.

Now using None as initial value, so we can spot that circumstance.

The 2 ** 63 - 1 value is just so it is MAX_INT on a 64bit platform, for better performance.
It can be easily increased when y2262 is coming.

### DIFF
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -191,7 +191,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
 
     def _read_files(self):
         self.files = {}
-        self._newest_mtime = 0
+        self._newest_mtime = None
         logger.debug('Reading files cache ...')
         with open(os.path.join(self.path, 'files'), 'rb') as fd:
             u = msgpack.Unpacker(use_list=True)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Yes, L457 is correct.

If it is still None, the term gives 0 and max(0, x) is x.
If it is 0 (1970 file), the term also gives 0 ...
If it is t > 0, the term gives t.

gleb:
I backported the fix to my 1.0.7 install.   This fixed reported problem in  production for me.

ThomasWaldmann:
@gleb thanks for feedback.

PR #1876, opened 2016-11-24
Vagrantfile cleanup

ThomasWaldmann:


enkore:
Something left to do here (assignment)?

ThomasWaldmann:
No, it's done.

I am just keeping stuff assigned so one sees in the milestone view what is being worked on (until it is done and merged).

PR #1881, opened 2016-11-26
Add begin_txn progress message

Abogical:
For #1721


src/borg/cache.py
```diff
@@ -233,6 +233,8 @@ def _read_files(self):
 
     def begin_txn(self):
         # Initialize transaction snapshot
+        pi = ProgressIndicatorMessage()
+        pi.output('Beginning cache transaction')
```
enkore: Or make them map 1:1 to the ones in commit or something like that

enkore:
Thanks!

PR #1882, opened 2016-11-26
add more details about resource usage

ThomasWaldmann:
not sure how to write these formulas better, right now they maybe look worse than what's usually happening in practice.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -206,36 +206,79 @@ Resource Usage
 
 |project_name| might use a lot of resources depending on the size of the data set it is dealing with.
 
-CPU:
+If one uses |project_name| in a client/server way (with a ssh: repository),
+the resource usage occurs in part on the client and in another part on the
+server.
+
+If one uses |project_name| as a single process (with a filesystem repo),
+all the resource usage occurs in that one process, so just add up client +
+server to get the approximate resource usage.
+
+CPU client:
+    borg create: does chunking, hashing, compression, crypto (high CPU usage)
+    chunks cache sync: quite heavy on CPU, doing lots of hashtable operations.
+    borg extract: crypto, decompression (medium to high CPU usage)
+    borg check: similar to extract, but depends on options given.
+    borg prune / borg delete archive: low to medium CPU usage
+    borg delete repo: done on the server
     It won't go beyond 100% of 1 core as the code is currently single-threaded.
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -206,44 +206,88 @@ Resource Usage
...
+Repository index (server only):
+    Proportional to the amount of data chunks in your repo. Lots of chunks
+    in your repo imply a big repository index.
+    It is possible to tweak the chunker params (see create options) to
+    influence the amount of chunks being created.
+
+Temporary files (client):
+    Reading data and metadata from a FUSE mounted repository will consume about
+    the same space as the deduplicated chunks used to represent them in the
+    repository.
```
enkore: No, the cache doesn't have all chunks, only small ones.

What it does need is as much free space as the size of the metadata, which can be a couple gigs for large archives...

docs/usage.rst
```diff
@@ -206,44 +206,88 @@ Resource Usage
...
     If your repository is remote, all deduplicated (and optionally compressed/
     encrypted) data of course has to go over the connection (ssh: repo url).
     If you use a locally mounted network filesystem, additionally some copy
     operations used for transaction support also go over the connection. If
     you backup multiple sources to one target repository, additional traffic
     happens for cache resynchronization.
 
-In case you are interested in more details, please read the internals documentation.
+In case you are interested in more details (like formulas), please read the
+internals documentation.
```
enkore: link?

docs/usage.rst
```diff
@@ -206,44 +206,88 @@ Resource Usage
...
+Chunks index (client only):
     Proportional to the amount of data chunks in your repo. Lots of chunks
     in your repo imply a big chunks index.
     It is possible to tweak the chunker params (see create options).
 
-Files index:
-    Proportional to the amount of files in your last backup. Can be switched
+Files index (client only):
+    Proportional to the amount of files in your last backups. Can be switched
     off (see create options), but next backup will be much slower if you do.
```
ThomasWaldmann: How about "... might be much slower" and also documenting that "the speed benefit of using the files cache is proportional to file size.".

enkore: Yeah

ThomasWaldmann:
merging this now as is, if there is something left to do, we can change it later.

ThomasWaldmann:
fsck, should have cleaned up the fixup first.

PR #1883, opened 2016-11-26
implement noatime / noctime, fixes #1853

### LINKED ISSUE #1853: Add option to disable backup of atime and ctime
Hi, I just noticed that borg now backups atime and ctime since version 0.28.0 (issue #317), and this is a problem for me :
- I use rsync daily to update a mirror of my data, and less often create borg archives. Rsync sets atime, so it means that the metadata of my archives can never be deduplicated (and they are close to 400MB for each archive). I have found a patch for rsync to prevent it to set atime, but it has been around there for years and has never been integrated.
- It does not make a lot of sense to me to duplicate metadata because I've done a batch grep on a directory, or an index tool has visited them, or I even just read some files.
- Sometimes I create the borg archive from one of my mirrors, so all the files have different ctime, and again this prevents the metadata from being deduplicated with the archives of another disk, even if the file tree is identical.

Does it make sense to at least add an option to "borg create" to disable backup of atime and ctime ? (in this case they could be set to the same value as mtime).

ThomasWaldmann:


borg/archive.py
```diff
@@ -571,10 +573,15 @@ def stat_attrs(self, st, path):
...
             b'uid': st.st_uid, b'user': uid2user(st.st_uid),
             b'gid': st.st_gid, b'group': gid2group(st.st_gid),
-            b'atime': int_to_bigint(st.st_atime_ns),
-            b'ctime': int_to_bigint(st.st_ctime_ns),
             b'mtime': int_to_bigint(st.st_mtime_ns),
         }
+        # borg can work with archives only having mtime (older attic archives do not have
+        # atime/ctime). it can be useful to omit atime/ctime, if they change without the
+        # file content changing - e.g. to get better metadata deduplication.
+        if not self.noatime:
```
enkore: Normally I'd rather avoid the double negation, but since "noatime" is pracitcally a fixed/well-known word in unix-lands, let's stick with it.

ThomasWaldmann: yup, also my thoughts. :)

PR #1888, opened 2016-11-27
Rename BORG_NONCES_DIR to BORG_SECURITY_DIR and then some

enkore:
Since BORG_NONCES_DIR was only part of beta releases so far I'd say we need no migration in borg itself but can just provide a shell snippet in the changelog that moves the files.

- [x] rename
- [x] add previously discussed additions to the mechanism
- [x] basic tests
- [x] Additional tests (see coverage report)

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -14,7 +14,7 @@ from .constants import CACHE_README
 from .hashindex import ChunkIndex, ChunkIndexEntry
 from .helpers import Location
 from .helpers import Error
-from .helpers import get_cache_dir
+from .helpers import get_cache_dir, get_security_dir
 from .helpers import decode_dict, int_to_bigint, bigint_to_int, bin_to_hex
 from .helpers import format_file_size
 from .helpers import yes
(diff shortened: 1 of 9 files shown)


src/borg/cache.py
```diff
@@ -29,6 +29,111 @@
...
+            fd.write(manifest.timestamp)
+
+    def assert_location_matches(self, cache):
+        # Warn user before sending data to a relocated repository
+        try:
+            with open(self.location_file) as fd:
+                previous_location = fd.read()
+            logger.debug('security: read previous_location %r', previous_location)
+        except OSError as exc:
+            if type(exc) is not FileNotFoundError:
```
ThomasWaldmann: not isinstance(exc, FileNotFoundError)

src/borg/cache.py
```diff
@@ -29,6 +29,111 @@
...
+            cache.begin_txn()
+            cache.commit()
+
+    def assert_no_manifest_replay(self, manifest, key, cache):
+        try:
+            with open(self.manifest_ts_file) as fd:
+                timestamp = fd.read()
+            logger.debug('security: read manifest timestamp %r', timestamp)
+        except OSError as exc:
+            if type(exc) is not FileNotFoundError:
```
ThomasWaldmann: same here

enkore:
Strange that Travis reports the build as failed even if only the "allowed failures" where cancelled and the rest is green.

PR #1900, opened 2016-11-29
partially remove virtualenv/pip version requirement, fixes #1738

### LINKED ISSUE #1738: remove virtualenv version requirement
requirements.d/development.txt:
Vagrantfile:
.travis/install.sh:

```
virtualenv<14.0
```

We did that so it installs a pip that is compatible with python3.2.
IIRC that was to support ubuntu precise (mostly do to us using precise-based travis-ci).
But we use trusty-based travis-ci now, so I guess we could remove that.

ThomasWaldmann:
Is needed only for python 3.2 support.

For normal development, we expect you have py34+ for borg 1.1.

For vagrant, it is still needed because of older VMs like wheezy (py32).

Not needed for Travis-CI any more, we moved to trusty VMs (py34) there.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -31,9 +31,9 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
             ;;
     esac
     pyenv rehash
-    python -m pip install --user 'virtualenv<14.0'
+    python -m pip install --user virtualenv
 else
-    pip install 'virtualenv<14.0'
+    pip install virtualenv
     sudo apt-get update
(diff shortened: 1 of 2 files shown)


enkore:
> For vagrant, it is still needed because of older VMs like wheezy (py32).

Hmm, but Borg doesn't support Python 3.2 (not even 1.0 does)?

ThomasWaldmann:
it is only for initial bootstrap, later, when tox is running, py34+ is used.

PR #1904, opened 2016-11-29
blake2b key modes: use B2B as MAC; longer keys.

### LINKED ISSUE #1896: blake2b as MAC layer for authenticated encryption.
way faster than hmac-sha256, one of borg's biggest cpu hogs.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/39600312-blake2b-as-mac-layer-for-authenticated-encryption)

enkore:
Fixes #1896

### DIFF
--- a/src/borg/key.py
+++ b/src/borg/key.py
@@ -161,6 +161,20 @@ class PlaintextKey(KeyBase):
         return Chunk(data)
 
 
+def random_blake2b_256_key():
+    # This might look a bit curious, but is the same construction used in the keyed mode of BLAKE2b.
+    # Why limit the key to 64 bytes and pad it with 64 nulls nonetheless? The answer is that BLAKE2b
+    # has a 128 byte block size, but only 64 bytes of internal state (this is also referred to as a
+    # "local wide pipe" design, because the compression function transforms (block, state) => state,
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
https://docs.python.org/3.6/library/hashlib-blake2.html#creating-hash-objects

rather have an api like that or at least param lengths doable with that?
then we could get rid of our own blake2b code again once we require python 3.6.

enkore:
Using the Py3.6 or OpenSSL functions would only require a thin wrapper (like hmac_sha256).

ThomasWaldmann:
I was just wondering about key length.

You assemble 64B random + 64B zeros. Python API only takes up to 64B (not 128B).
But I've looked into the python implementation, it seems to do same internally.

Seems you have also checked that hashes are same with that py36-blake2.py script.

enkore:
Correct, the keyed API only takes up to 64B keys for the reasons mentioned in the comment. Note that due to the different parameter blocks between unkeyed and keyed hashes it wouldn't result in the same hash, you'll have to use the unkeyed API and process key and data as indicated. This is also what gives us the option to use OpenSSL.

PR #1911, opened 2016-12-01
Work on metadata handling speed

enkore:
I tested this mostly on a very artificial test case: no data at all in the files, but many files and deep directory structures. So comparisons put a spotlight on how fast Borg processes metadata and ex-aggregates per-file overheads.

There are four changes:

1. Compile borg.item with Cython. Item.update is a very hot function that is called multiple times per file and tends to take a couple percent overall runtime even in more realistic conditions.

    I suspect the slightly increased maintenance burden (have to recompile after code change) is mostly irrelevant, since the file is edited rarely.
2. Skip data IO to empty files entirely - arguably a special case appealing this synthetic benchmark (to great effect). The practical impact of this change will be rather negligible. It might be more interesting to further special-case the single-chunk case (not instantiating chunker and so on).
3. Changed backup_io from a generator-based `@contextmanager` to a ctx mgr class, which is additionally statically instantiated. When directly compared the latter is far faster, for the usual reasons (no allocation, no separate stack for the generator, much less indirection).

    Like point 4.) and 1.) this is called multiple times for each file.
4. Removed "bigint" conversions from timestamps. These support dates beyond the year 2262, and are called many times for each file.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -125,19 +125,22 @@ class BackupOSError(Exception):
         return str(self.os_error)
 
 
-@contextmanager
-def backup_io():
-    """Context manager changing OSError to BackupOSError."""
-    try:
-        yield
-    except OSError as os_error:
-        raise BackupOSError(os_error) from os_error
+class BackupIO:
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type and issubclass(exc_type, OSError):
+            raise BackupOSError(exc_val) from exc_val
+
+
+backup_io = BackupIO()
 
 
 def backup_io_iter(iterator):
     while True:
         try:
-            with backup_io():
+            with backup_io:
                 item = next(iterator)
         except StopIteration:
             return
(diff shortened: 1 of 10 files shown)


ThomasWaldmann:
I think we should drop the empty files specialcasing changeset.
Empty files are not that frequent.

As you noticed single-chunk files (< min_chunk_size) are more interesting - we can optimize that later.

ThomasWaldmann:
There are some tests failing now, see travis.

enkore:
Done. aac61ea is a bit uglier but hey. Depending on test data this branch is now in my testing between 5 and 25 % faster than current master code.

src/borg/archive.py
```diff
@@ -859,7 +859,17 @@ def process_file(self, path, st, cache, ignore_inode=False):
             with backup_io:
                 fh = Archive._open_rb(path)
             with os.fdopen(fh, 'rb') as fd:
-                self.chunk_file(item, cache, self.stats, backup_io_iter(self.chunker.chunkify(fd, fh)), compress=compress)
+                if is_special_file or st.st_size > self.chunker.min_size:
+                    self.chunk_file(item, cache, self.stats, backup_io_iter(self.chunker.chunkify(fd, fh)), compress=compress)
+                else:
+                    data = fd.read()
```
ThomasWaldmann: the problem with avoiding the chunker is that you avoid the posix_fadvise also. so all these small files will blow up the cache.

src/borg/archive.py
```diff
@@ -859,7 +859,17 @@ def process_file(self, path, st, cache, ignore_inode=False):
...
+                    if data:
+                        item.chunks = [cache.add_chunk(self.key.id_hash(data), Chunk(data, compress=compress), self.stats)]
+                    else:
+                        item.chunks = []
+                    if self.checkpoint_interval and time.time() - self.last_checkpoint > self.checkpoint_interval:
+                        self.write_checkpoint()
+                        self.last_checkpoint = time.time()
```
ThomasWaldmann: also, this needed to be copied/duplicated from chunk_file, I guess.

ThomasWaldmann:
flake8 is unhappy.

PR #1912, opened 2016-12-01
Add prune progress display

Abogical:
For #1721

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1016,6 +1016,10 @@ class Archiver:
         stats = Statistics()
         with Cache(repository, key, manifest, do_files=args.cache_files, lock_wait=self.lock_wait) as cache:
             list_logger = logging.getLogger('borg.output.list')
+            if args.output_list:
+                # set up counters for the progress display
+                to_delete_len = len(to_delete)
+                archives_deleted = 0
             for archive in archives_checkpoints:
                 if archive in to_delete:
                     if args.dry_run:
@@ -1023,7 +1027,9 @@ class Archiver:
                             list_logger.info('Would prune:     %s' % format_archive(archive))
                     else:
                         if args.output_list:
-                            list_logger.info('Pruning archive: %s' % format_archive(archive))
+                            archives_deleted += 1
+                            list_logger.info('Pruning archive: %s (%d/%d)' % (format_archive(archive),
+                                                                              archives_deleted, to_delete_len))
                         Archive(repository, key, manifest, archive.name, cache,
                                 progress=args.progress).delete(stats, forced=args.forced)
                 else:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1024,7 +1024,8 @@ class Archiver:
                     else:
                         if args.output_list:
                             list_logger.info('Pruning archive: %s' % format_archive(archive))
-                        Archive(repository, key, manifest, archive.name, cache).delete(stats, forced=args.forced)
+                        Archive(repository, key, manifest, archive.name, cache,
+                                progress=args.progress).delete(stats, forced=args.forced)
                 else:
                     if args.output_list:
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1016,18 +1016,26 @@ def do_prune(self, args, repository, manifest, key):
...
                     else:
-                        if args.output_list:
-                            list_logger.info('Pruning archive: %s' % format_archive(archive))
-                        Archive(repository, key, manifest, archive.name, cache).delete(stats, forced=args.forced)
+                        archives_to_delete.append(archive)
                 else:
                     if args.output_list:
                         list_logger.info('Keeping archive: %s' % format_archive(archive))
+            if args.progress:
+                len_archives_to_delete = len(archives_to_delete)
```
ThomasWaldmann: isn't this the same as len(to_delete)?

if so, can you stay close to the old code to do what you want?

Abogical: Ok that change was dumb. I though maybe not all of the `to_delete` set was in `archive_checkpoints`. Sorry about that, will change it.

src/borg/archiver.py
```diff
@@ -1024,7 +1028,11 @@ def do_prune(self, args, repository, manifest, key):
                     else:
                         if args.output_list:
                             list_logger.info('Pruning archive: %s' % format_archive(archive))
-                        Archive(repository, key, manifest, archive.name, cache).delete(stats, forced=args.forced)
+                        Archive(repository, key, manifest, archive.name, cache,
+                                progress=args.progress).delete(stats, forced=args.forced)
+                        if args.progress:
+                            archives_deleted += 1
+                            logger.warning('(%d/%d) archives pruned' % (archives_deleted, to_delete_len))
```
ThomasWaldmann: this is progress display (information), not a warning.

why didn't you use the ProgressIndicator used elsewhere?

Abogical: Will change warning to info...

src/borg/archiver.py
```diff
@@ -1024,7 +1028,11 @@ def do_prune(self, args, repository, manifest, key):
                     else:
                         if args.output_list:
                             list_logger.info('Pruning archive: %s' % format_archive(archive))
```
ThomasWaldmann: see this ^^^

src/borg/archiver.py
```diff
@@ -1024,7 +1028,11 @@ def do_prune(self, args, repository, manifest, key):
...
-                        Archive(repository, key, manifest, archive.name, cache).delete(stats, forced=args.forced)
+                        Archive(repository, key, manifest, archive.name, cache,
+                                progress=args.progress).delete(stats, forced=args.forced)
+                        if args.progress:
+                            archives_deleted += 1
+                            logger.info('(%d/%d) archives pruned' % (archives_deleted, to_delete_len))
```
ThomasWaldmann: ... and this ^^^.

So, the only difference is that one is output before and the other after deletion of one archive - and the counter / total.

I think one log line is enough for that (and it maybe should be before deleting the archive).

Check how 1.0-maint handles this and also how it is handled at other places (--list vs. --progress).

Also, the progress output happens on the borg.output.progress logger elsewhere.

Abogical: If i understand correctly, Is this commit useless?

ThomasWaldmann: well, it does not list counter/total info (but it could).

Abogical: Good idea, will do that.

Abogical:
I'm not sure if you are waiting for another commit, but I'm done here.

Abogical:
Accidentally closed PR, sorry about that.

Not sure if you're waiting for another commit, but I'm done here.

enkore:
Thanks!

Abogical:
Your welcome, I'm waiting for feedback on #1721.

Issue #1914, opened 2016-12-01
borg info requires repo_or_archive param


ThomasWaldmann:
So, just setting BORG_REPO env var and doing a `borg info` (without the param) does not work.

See how `borg list` does it, it works there.

Found in master branch, check 1.0-maint also.

PeaceOff:
hey i am entering the open-source community, can i take a look?

ThomasWaldmann:
sure, see archiver.py.

PeaceOff:
on it, Thanks!

ThomasWaldmann:
@joaocsf if you end your commit comment fix ", fixes #1914", github will automatically close #1914.

Fixed by #1917. Closing manually...

enkore:
GitHub help: https://help.github.com/articles/closing-issues-via-commit-messages/

joaocsf:
Sorry, I didn't knew about that. Thank you for the information! I'm learning alot with this project!

Issue #1915, opened 2016-12-01
systematically check all commands for BORG_REPO usage


ThomasWaldmann:
See #1914, but check all commands.

1.0-maint and later master.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/39690403-systematically-check-all-commands-for-borg_repo-usage)

ThomasWaldmann:
moved to 1.1 due to new 1.0 policy.

ThomasWaldmann:
master (some commands need `::` placeholder, when other pos. args follow the repo pos. arg):
```
N/A serve            start repository server process

ok  init             initialize empty repository

ok  check            verify repository

ok  key              manage repository key
    ok  borg key export (needs :: placeholder)
    ok  borg key change-passphrase

ok  change-passphrase change repository passphrase

ok  create           create backup
ok  extract          extract archive contents
ok  diff             find differences in archive contents
ok  rename           rename archive

ok  delete           delete archive / repo
    ok delete repo
    ok delete archive

ok  list             list archive or repository contents
    ok list repo
    ok list archive

ok  mount            mount repository (needs :: placeholder)

N/A umount           umount repository

ok  info             show repository or archive information

ok  break-lock       break repository and cache locks

ok  prune            prune archives

ok  upgrade          upgrade repository format

ok  recreate         Re-create archives

ok  with-lock        run user command with lock held (needs :: placeholder)

ok  debug            debugging command (not intended for normal use)
    ok  borg debug info
    ok  borg debug dump-manifest (needs :: placeholder)

    benchmark        benchmark command
    ok borg benchmark crud (needs :: placeholder)
```

ThomasWaldmann:
1.0-maint:
```
N/A serve               start repository server process

ok  init                initialize empty repository

ok  check               verify repository

ok  change-passphrase   change repository passphrase

ok  key                 manage repository key
    ok borg key export :: file
    ok borg key import :: file

ok  migrate-to-repokey  migrate passphrase-mode repository to repokey

ok  create              create backup

ok  extract             extract archive contents

ok  rename              rename archive
ok  delete              delete archive
    ok delete archive
    ok delete repo

ok  list                list archive or repository contents
    ok list repo
    ok list archive

ok  mount               mount repository
ok  umount              umount repository

ok  info                show archive information

ok  break-lock          break repository and cache locks

ok  prune               prune archives

ok  upgrade             upgrade repository format

[...]

PR #1917, opened 2016-12-01
Fixed borg info

joaocsf:
Issue #1914 fixed.
Ran all the tests, everything seems allright.
Probably the same problem in 1.0-maint.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2192,7 +2192,7 @@ class Archiver:
                                           formatter_class=argparse.RawDescriptionHelpFormatter,
                                           help='show repository or archive information')
         subparser.set_defaults(func=self.do_info)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
change looks good, thanks! but the PR should go against 1.0-maint branch IF the problem also exists in 1.0-maint. It will get merged automatically from there into master later.

ThomasWaldmann:
also, please have a good commit comment.

joaocsf:
I have tested the same fix in the 1.0-maint branch and it seems this version doesn't support repository information (only archives). If you specify the archive, it works as expected, even with the BORG_REPO set eg: `borg info ::archive`.

ThomasWaldmann:
@joaocsf ah, right. ok, so master is fine.

ThomasWaldmann:
Thanks for fixing this. Check #1915 also.

PR #1922, opened 2016-12-02
setup.py: fix build_usage not processing all commands

enkore:


enkore:
The debug* files are not processed by Sphinx anyway and thus inaccessible - not even with "deep linking".

PR #1934, opened 2016-12-08
Update link to security contact

musoke:
Information about security contact seems to have moved.


ThomasWaldmann:
Thanks for fixing this!

enkore:
Thanks for the patch, but it's actually not quite correct. Since master is currently not stable the link goes to nil, but as soon as there is a stable release from the master development branch the link becomes correct.

ThomasWaldmann:
@enkore is right - @musoke can you do it like that?

ThomasWaldmann:
@musoke still there?

musoke:
@ThomasWaldmann - sorry, I shouldn't submit pull requests immediately before going AFK for a month.

@enkore, thanks for fixing it.

Issue #1729, opened 2016-10-16
Improve cache sync speed


enkore:
See #235 and #536 

Some findings in https://github.com/borgbackup/borg/issues/235#issuecomment-254010549

So I've started to analyze some of my logs and there are indeed a few observable oddities. For example, some merges will take much longer than others, even when the archive index that is merged is small.

One particularly extreme sample is that merging one archive took almost one minute, while archives with more than ten times as many chunks merged in less than two seconds. When forced to reproduce, it did not happen with the same archive. This might be #536 

We've already found that there's quite a performance regression in master branch here (see third link above), but that one is fixed easily. Some minor difference persists between 1.0.x and master, though; root cause unknown so far.

Further, the process feels quite slow given the data that is processed. I am suspecting that by using msgpack-python we give up a lot of performance, because it creates Python objects for everything in the items stream, most of which doesn't matter here. Quite possible that we could speed it up significantly by using a zero-copy-oriented approach and coding it in C or using msgpack-c in this spot and directly interfacing with the C API of the hash index. (This will likely add either msgpack-c as a dependency -- which shouldn't be an issue with us already requiring msgpack-python anyway -- or a _significant_ chunk of C).

ThomasWaldmann:
guess it rather depends on how many *new* keys are in a index than the amount of all keys in that index. insert is more expensive than update and sometimes might also trigger table growth.

enkore:
Fixed by #2572

ThomasWaldmann:
"will be fixed by #2572" (it is not merged yet)

Issue #235, opened 2015-10-01
significantly reduce cache space needs


anarcat:
after converting an attic repo to borg (using #231), i am having problems running `borg create`. i suspect this is not related to the migration, but with the new caching algorithm that borg uses.

when running `borg create`, it seems to want to regenerate the chunks cache, even though it was converted and is available in `$BORG_CACHE_DIR/<id>/chunks`:

```
# borg create --do-not-cross-mountpoints              --stats                                          /media/anarcat/BOOK/borg::$(hostname)-$(date +%Y-%m-%d)-post-attic       / /boot /usr /var /home                          --exclude-caches                                 --exclude "/home/*/.cache"                       --exclude "*/.Trash-*"                           --exclude "*/[Cc]cache/*"                        --exclude "*/.bitcoin/blocks/*"                  --exclude "/tmp/*"                               --exclude "*/build-area/*"                       --exclude "/proc/*"                              --exclude "/dev/*"                               --exclude "/sys/*"                               --exclude "/var/cache/*"                         --exclude "/var/tmp/*"                      --exclude "/var/log/*" -v -v
Synchronizing chunks cache...
Archives: 284, w/ cached Idx: 9, w/ outdated Idx: 0, w/o cached Idx: 275.
Fetching and building archive index for marcos-2015-08-22 ...
Merging into master chunks index ...
Fetching and building archive index for marcos-2015-04-15 ...
Merging into master chunks index ...
Fetching and building archive index for marcos-2015-05-17 ...
Merging into master chunks index ...
[...]
Fetching and building archive index for marcos-2015-02-19 ...
hashindex: /root/.cache/borg/d31f0c3d6969eb44b6a77eb4edac2ec0c1d257c8fbb3a0e26ae4cede83963a89/chunks.archive.d/28fbf8d98f730e04b9835e2c628fed6f00f057267979ed9b3b7c1497782760ca.tmp: fwrite buckets failed (No space left on device)
```

... you can see how it's quickly running out of space. whereas the cache directory was 1GB in Attic, this ran out of space after 8GB written.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/27100351-significantly-reduce-cache-space-needs)

anarcat:
every chunk archive is 353MB. so if i let this go through, it will take around 100GB:

```
> 353Mibyte * 284 à Gibyte

  353 * mebibyte * 284 = approx. 97.902344 gibibytes
```

[...]

PR #1940, opened 2016-12-09
Cache sync stuff

enkore:


src/borg/cache.py
```diff
@@ -435,10 +434,8 @@ def fetch_and_build_idx(archive_id, repository, key):
                     if not isinstance(item, dict):
                         logger.error('Error: Did not get expected metadata dict - archive corrupted!')
                         continue
-                    item = Item(internal_dict=item)
-                    if 'chunks' in item:
-                        for chunk_id, size, csize in item.chunks:
-                            chunk_idx.add(chunk_id, 1, size, csize)
+                    for chunk_id, size, csize in item.get(b'chunks', []):
+                        chunk_idx.add(chunk_id, 1, size, csize)
```
ThomasWaldmann: Does that make a big difference, did you measure it?

src/borg/cache.py
```diff
@@ -474,21 +470,27 @@ def create_master_idx(chunk_idx):
...
+                            # to avoid growing the hash table from 0 size and also
+                            # to save 1 merge call.
+                            chunk_idx = archive_chunk_idx
+                        else:
+                            chunk_idx.merge(archive_chunk_idx)
                     else:
-                        chunk_idx.merge(archive_chunk_idx)
+                        chunk_idx = chunk_idx or ChunkIndex()
+                        logger.info('Fetching archive index for %s ...' % archive_name)
+                        fetch_and_build_idx(archive_id, repository, self.key, chunk_idx)
```
ThomasWaldmann: ah, nice, that saves the merging.

src/borg/cache.py
enkore: These are both a bit older, this change was looked at previously in #1729 / https://github.com/borgbackup/borg/issues/235#issuecomment-254010549

PR #1941, opened 2016-12-09
Fix some clock problems

### LINKED ISSUE #1936: Incorrect duration if clock jumps backwards during borg-create
Two bugs actually:

1. The duration becomes negative.
2. Negative deltas aren't correctly calculated, either.

Eg.: Starting around 00:52:00 (on the 6th), then crossing over to 23:57:00 (on the 5th) displays something along the lines of "-1 days 5 minutes xy seconds".

(Borg isn't the only one getting it wrong, though. GNU time says: "243user 18system -55:-28.37elapsed" -- note how the absolute time span is still wrong by about 50 minutes!).

IIRC we already store the duration in seconds in the archive metadata, so we'd only have to use a different clock to calculate that (I'd leave the timestamps as-is, no way knowing which one would be correct!), ie. a monotonic one that's good for at least a couple days (large creates and all).

enkore:
Fixes #1936 (at least partially, see ticket)

In 3.4 this is available on "Windows, Mac OS X, Linux, FreeBSD, OpenBSD, Solaris.", but it also works on NetBSD and OI/illumos. Note that it is unconditionally available with 3.5+. On Windows overflows are handled, so it won't roll over after ~50 days.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -304,15 +304,17 @@ Number of files: {0.stats.nfiles}'''.format(
         if name in self.manifest.archives:
             raise self.AlreadyExists(name)
         self.items_buffer.flush(flush=True)
+        duration = timedelta(seconds=time.monotonic() - self.start_monotonic)
         if timestamp is None:
             self.end = datetime.utcnow()
-            self.start = self.end - timedelta(seconds=time.monotonic() - self.start_monotonic)
+            self.start = self.end - duration
             start = self.start
             end = self.end
         else:
             self.end = timestamp
-            start = timestamp
-            end = timestamp  # we only have 1 value
+            self.start = timestamp - duration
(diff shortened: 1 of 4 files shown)


borg/archive.py
```diff
@@ -302,7 +305,7 @@ def save(self, name=None, timestamp=None):
             raise self.AlreadyExists(name)
         self.items_buffer.flush(flush=True)
         if timestamp is None:
-            self.end = datetime.utcnow()
+            self.end = self.start + timedelta(seconds=time.monotonic() - self.start_monotonic)
```
enkore: Good point. Using the likely corrected time has the advantage that it's probably correct, otoh the start time wouldn't match the timestamps of any log entries associated with the job -- but then again they should be linked by the archive name.

Issue #1812, opened 2016-11-07
Export statistics or graph them with borg


lightonflux:
Just a feature suggestion. 

Either a statistics command or a command that exports some statistics in an easily parseable format like json so graphing solutions like grafana can use it. Or a statistics command that draws stats like `vnstat` ([image](http://www.serverninjas.com/wp-content/uploads/2013/04/vnstat-hourly-report.jpg)). 

Statistic items that are (IMHO) worth to graph:

 * number of files
 * time of completion
 * dedup size
 * compressed size
 * orig size
 * chunks  (personally not interested in that)

Here some pseudo UI:

     borg export-stats --format json path-to/repo > wonderful.json

Why is it important to log these infos?

Imagine you backup a data you don't have control over. E.g. a webserver with some software that creates thousands of php session files. This would slow down borg MASSIVLY. Or if you use higher compression and suddenly several gigs of compressed data appear.

It would be nice to see all the stuff in easy to understand graphs. It would allow to improve your retention policy. 

Some of these problems would be addressed by other festures / bug reports e.g.:

 * be faster with many files (think 5 mio 44kb files)
 * limit repo by size  → max retention
 * compression by file type
 * good defaults for compression by file type

This is just a suggestion. And i would like to hear your input about how to make this more useful and if you see a need for this.

ThomasWaldmann:
Nice idea.

I'ld rather object adding graphing stuff to borg and doing that within borg. The interface should be that borg exports such data in easily readable format for external graphing tools - like csv or json.

A first step in that direction could be that borg info archive gives some more info (see above).

enkore:
http://borgbackup.readthedocs.io/en/latest/internals/frontends.html#standard-output

PR #1946, opened 2016-12-11
Added --export-file flag to the info command 

JoseLuisMartins:
#1812 
Added `--export-file` flag to the `borg info` command in order to export repository or archive information encoded as a json file. 

The exporter.py file was added to separate concepts, making it easly in the future to add some new export formats and related stuff, but let me know what you guys think about that.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -373,10 +373,9 @@ Number of files: {0.stats.nfiles}'''.format(
         del self.manifest.archives[self.checkpoint_name]
         self.cache.chunk_decref(self.id, self.stats)
 
-
     def get_summary(self):
-        summary = namedtuple('Summary', ['name','fingerprint', 'comment',
-        'hostname','username','time_start','time_end','duration','command_line'])
+        summary = namedtuple('Summary', ['name', 'fingerprint', 'comment',
+        'hostname', 'username', 'time_start', 'time_end', 'duration', 'command_line'])
         return summary(*self.summarize())._asdict()
 
     def summarize(self):
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -4,6 +4,7 @@ import socket
 import stat
 import sys
 import time
+from collections import namedtuple
 from contextlib import contextmanager
 from datetime import datetime, timezone
 from functools import partial
@@ -60,6 +61,13 @@ class Statistics:
             self.usize += csize
 
     summary = "{label:15} {stats.osize_fmt:>20s} {stats.csize_fmt:>20s} {stats.usize_fmt:>20s}"
+
+    def get_summary(self):
+        summary = namedtuple('Summary', ['original_size', 'compressed_size', 'unique_size', 'number_files'])
+        return summary(*self.summarize())._asdict()
+
+    def summarize(self):
+        return self.osize, self.csize, self.usize, self.nfiles
 
     def __str__(self):
         return self.summary.format(stats=self, label='This archive:')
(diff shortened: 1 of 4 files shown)


src/borg/archiver.py
```diff
@@ -938,20 +939,33 @@ def format_cmdline(cmdline):
...
-            print('Username: %s' % archive.metadata.username)
-            print('Time (start): %s' % format_time(to_localtime(archive.ts)))
-            print('Time (end):   %s' % format_time(to_localtime(archive.ts_end)))
-            print('Duration: %s' % archive.duration_from_meta)
-            print('Number of files: %d' % stats.nfiles)
-            print('Command line: %s' % format_cmdline(archive.metadata.cmdline))
+
+            archiveSummary = archive.get_summary()
+            statsSummary = stats.get_summary()
+            cacheSummary = cache.get_summary()
```
enkore: Nit pick: [PEP8](https://www.python.org/dev/peps/pep-0008/) names would be preferred (eg. `archive_summary` instead of `archiveSummary`)

src/borg/archiver.py
```diff
@@ -976,6 +994,17 @@ def _info_repository(self, repository, key, cache):
         print(DASHES)
         print(STATS_HEADER)
         print(str(cache))
+
+        if export_file:
+            repository_info = {
+                'encripted': encrypted,
```
enkore: `'encrypted'` (typo)

Probably better to make this more machine-readable, eg. like this:

```
'encryption': {
  'type': key.NAME,
  'keyfile': key.find_key(),
}
```

Unencrypted repo -> no 'encryption' dict?

src/borg/cache.py
```diff
@@ -219,6 +219,10 @@ def __str__(self):
 Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         return fmt.format(self.format_tuple())
 
+    def get_summary(self):
+        Summary = namedtuple('Summary', ['total_size', 'total_csize', 'unique_size', 'unique_csize', 'total_unique_chunks', 'total_chunks'])
+        return Summary(*self.chunks.summarize())._asdict()
+
     def format_tuple(self):
         # XXX: this should really be moved down to `hashindex.pyx`
         Summary = namedtuple('Summary', ['total_size', 'total_csize', 'unique_size', 'unique_csize', 'total_unique_chunks', 'total_chunks'])
```
enkore: Some duplication here, can you make format_tuple() use get_summary for the first two lines? They're identical as far as I see.

src/borg/archive.py
```diff
@@ -365,6 +373,16 @@ def write_checkpoint(self):
...
 
+    def get_summary(self):
+        summary = namedtuple('Summary', ['name', 'fingerprint', 'comment',
+        'hostname', 'username', 'time_start', 'time_end', 'duration', 'command_line'])
+        return summary(*self.summarize())._asdict()
+
+    def summarize(self):
+        return [self.name, self.fpr, self.metadata.get('comment', ''),
+        self.metadata.hostname, self.metadata.username, format_time(to_localtime(self.ts)),
+        format_time(to_localtime(self.ts_end)), self.duration_from_meta, self.metadata.cmdline]
```
enkore: I think it would be more clear if you perhaps notated this more directly, eg. the namedtuple isn't really used.

```py
def get_summary(self):
    return {
        'name': self.name,
        'fingerprint': self.fpr,
        # ... rest like you did it
    }
```

What do you think?

enkore:
What do you think about making this more like an display option -- right now --export-json takes a file that is written to, but it also displays the regular output -- like `--json` and it just gives you the JSON on standard out, so you could pipe it where you need it, eg. `jq` to extract some bit of information?

enkore:
Superseded by #2203

Issue #1949, opened 2016-12-13
vagrant testing on Amazon EC2


ThomasWaldmann:
Yesterday I was experimenting with this a bit. Reason is that I usually work on my laptop, but it only has a (dual core) i5-4200U, 8GB RAM (not expandable) and a SATA SSD - it gets really to its limits when I run multiple vagrant machines in parallel, the usual release testing takes hours...

I considered getting a faster machine for myself, but ruled that out (I like light and slim ultrabooks [svp13 == 1kg] and would really dislike a heavy / bulky / loud machine with power I only need now-and-then, but mostly not).

Thus searching for other faster ways:

First try was just to have one really big EC2 instance with lots of cores and memory and run vagrant on that machine. It somehow seemed to work at first (although with some slowness issues and sometimes vagrant VMs being stuck somehow), but then I wanted to get the darwin64 stuff working, needed usb support, installed virtualbox extension pack and current virtualbox and then it stopped working: "running under Xen is not supported". So that approach (nested virtualization) seems wrong and not really supported.

Second try is with vagrant-aws plugin - here you can use a EC2 machine like it normally would use a virtualbox machine. Thus: more and smaller machines, but no nested virtualization issues.

ThomasWaldmann:
BTW, about cost:
- running some few big machines or some more smaller ones for a short time (like some hours, not permanently) on EC2 is rather cheap.
- storing many GBs in their storage (permanently!) not so cheap, due to the time factor.

ThomasWaldmann:
AMIs (amazon machine images) exist for (and could be used as base for our testing machines):
- misc linuxes
- freebsd
- netbsd
- openbsd
- windows (higher machine prices, likely due to license costs, but affordable)
- OS X (darwin 64bits) status / support unknown

ThomasWaldmann:
note: on ec2 it is worse than with the prebuilt vagrant machine concerning having a standard username, so it might be best to just try sudo / su to create the vagrant user if it is not present.

ThomasWaldmann:
see also #864 about parallel test execution.

walkertraylor:
Considerations
- Use Azure provider because it is billed per minute.

ThomasWaldmann:
@walkertraylor ah, interesting, thanks for the hint. it will also depend on other factors though, like image availability for misc. OSes, how hard it is to get them working, etc.

walkertraylor:
[...]

PR #1950, opened 2016-12-13
WIP: vagrant: add AWS (EC2) provider

ThomasWaldmann:
Just a beginning, there are quite some issues left, see #1949.

Also, PR is against master for now, has to be decided whether we want it for 1.0 also.


ThomasWaldmann:
Considering that I've just upgraded my local computing power, I guess I've lost personal interest in this.

So, this PR is free for adoption. ;-)

PR #1964, opened 2016-12-17
Fix processing of remote ~/ and ~user/ paths

### LINKED ISSUE #1759: check remote.py "~user/..." path processing

enkore:
Seems to work now both with ssh://.../~user/ and someone@host:~user (and also someone@host:/~user/) and with plain ~.

Fixes #1759

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -263,14 +263,26 @@ class RepositoryServer:  # pragma: no cover
         # not a known old format, send newest negotiate this version knows
         return {'server_version': BORG_VERSION}
 
-    def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, append_only=False):
+    def _resolve_path(self, path):
         if isinstance(path, bytes):
             path = os.fsdecode(path)
-        if path.startswith('/~'):  # /~/x = path x relative to home dir, /~username/x = relative to "user" home dir
-            path = os.path.join(get_home_dir(), path[2:])  # XXX check this (see also 1.0-maint), is it correct for ~u?
+        # Leading slash is always present with URI (ssh://), but not with short-form (who@host:path).
+        if path.startswith('/~/'):  # /~/x = path x relative to home dir
+            path = os.path.join(get_home_dir(), path[3:])
+        elif path.startswith('~/'):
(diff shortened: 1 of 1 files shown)


src/borg/remote.py
```diff
@@ -263,14 +263,26 @@ def negotiate(self, client_data):
...
+    def _resolve_path(self, path):
         if isinstance(path, bytes):
             path = os.fsdecode(path)
-        if path.startswith('/~'):  # /~/x = path x relative to home dir, /~username/x = relative to "user" home dir
-            path = os.path.join(get_home_dir(), path[2:])  # XXX check this (see also 1.0-maint), is it correct for ~u?
+        # Leading slash is always present with URI (ssh://), but not with short-form (who@host:path).
+        if path.startswith('/~/'):  # /~/x = path x relative to home dir
+            path = os.path.join(get_home_dir(), path[3:])
+        elif path.startswith('~/'):
+            path = os.path.join(get_home_dir(), path[2:])
```
ThomasWaldmann: so, the stuff above is different from 1.0-maint now.

the stuff below corresponds to 1.0-maint.

is that what we want? is 1.0 thus incorrect somehow?

enkore: Either form works in 1.0-maint because all paths are passed through expanduser, and if the path starts with /~ then the leading / is cut off.

So this behaviour is now back to the same behaviour as 1.0 had, just that ~/ is get_home_dir() magics not expanduser.

ThomasWaldmann: OK, and this get_home_dir() magic won't get us into compatibility troubles?
Shall it get backported to 1.0-maint?

ThomasWaldmann: The suid issues / borgfs wrapper is only in master/1.1, so guess it is fine as is.

PR #1966, opened 2016-12-17
Fix subsubparsers for Python <3.4.3

enkore:
This works around http://bugs.python.org/issue9351

Since Debian and Ubuntu ship 3.4.2 and 3.4.0 respectively.

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -1160,7 +1160,7 @@ class Archiver:
                                           help='manage repository key')
 
         key_parsers = subparser.add_subparsers(title='required arguments', metavar='<command>')
-        subparser.set_defaults(func=functools.partial(self.do_subcommand_help, subparser))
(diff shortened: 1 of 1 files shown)


borg/archiver.py
```diff
@@ -1902,11 +1902,12 @@ def parse_args(self, args=None):
     def run(self, args):
         os.umask(args.umask)  # early, before opening files
         self.lock_wait = args.lock_wait
-        setup_logging(level=args.log_level, is_serve=args.func == self.do_serve)  # do not use loggers before this!
+        func = getattr(args, 'func', None) or getattr(args, 'fallback_func')
```
ThomasWaldmann: add a comment with a link to the bug above this line?

ThomasWaldmann:
lgtm, just maybe add that comment, so one does not need to dig git history for the explanation.

ThomasWaldmann:
travis looks good, add comment line & merge.

PR #1970, opened 2016-12-20
fast crc

enkore:
More like a "technology preview" (*chuckle*), still some loose ends to tie up and some compat tests to do. Also have to evaluate improvements (or not) brought on by the zero bypass in practical situations.

CRC32 CLMUL works on all Intel CPUs since Nehalem and AMD desktop CPUs later than K10 (exclusive), but it became a lot faster with Haswell apparently. The zero bypass thing works on all x64.

slice by 8 is a generic/portable technique. It's faster than the zlib implementation, but slower than the folding approach and requires a large table which the folding approach doesn't need.


src/borg/archiver.py
```diff
@@ -1129,6 +1129,64 @@ def do_with_lock(self, args, repository):
...
+                    td += t1 - t0
+                print(func.__name__, 'speed: %.1f MB/s' % (args.trials * size / td / 1000**2))
+            components[func.__name__] = wrapper
+            return wrapper
+
+        @component
+        def chunker(size):
+            from .chunker import Chunker
+            from .constants import CHUNKER_PARAMS
+            import io
```
ThomasWaldmann: so, you are measuring the time needed to import also?

src/borg/archiver.py
```diff
@@ -1129,6 +1129,64 @@ def do_with_lock(self, args, repository):
...
+
+        @component
+        def blake2b(size):
+            from .crypto import blake2b_256
+            blake2b_256(bytes(32), bytes(size))
+
+        @component
+        def crc32_zlib(size):
+            from zlib import crc32
+            crc32(bytes(size))
```
ThomasWaldmann: hmm, is feeding lots of zeros in here triggering a special case that is much faster than the usual case?

enkore: No, there is a difference though due to the way large memory areas are allocated (mapping the same physical page many times), meaning that compared to real data laid out across a memory access will be faster. This reduces throughput (of all implementations), but affects the faster one stronger, of course.

However, since Borg works on chunks that are just processed before they're likely in some cache, so solely looking at the streaming case is probably not quite right, too. Like I wrote above - it's a crude tool, more for telling the ballpark.

enkore:
ftr Travis fail is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56298 so I'll work around that.

src/borg/_chunker.c
```diff
@@ -236,6 +235,96 @@ chunker_fill(Chunker *c)
...
+    /* let's roll */
+    __m128i zerovec;
+    __m128i xmm0, xmm1, xmm2, xmm3;
+    zerovec = _mm_setzero_si128();
+
+    while((length -= 64) >= 0) {
+        xmm0 = _mm_load_si128((__m128i *)data);
+        xmm1 = _mm_load_si128((__m128i *)data + 1);
+        xmm2 = _mm_load_si128((__m128i *)data + 2);
+        xmm3 = _mm_load_si128((__m128i *)data + 3);
```
ThomasWaldmann: how about using data++ 4 times ...

src/borg/_chunker.c
```diff
@@ -236,6 +235,96 @@ chunker_fill(Chunker *c)
...
+        xmm2 = _mm_or_si128(xmm2, xmm3);
+        xmm0 = _mm_or_si128(xmm0, xmm2);
+
+        xmm0 = _mm_cmpeq_epi32(xmm0, zerovec);
+
+        if(_mm_movemask_epi8(xmm0) != 0xffff) {
+            return 0;
+        }
+
+        data += 64;
```
ThomasWaldmann: ... and leaving this away?

src/borg/_chunker.c
```diff
@@ -236,6 +235,96 @@ chunker_fill(Chunker *c)
...
+    }
+
+    if(length < 0) {
+        length += 64;
+    }
+
+partial:
+    /* consume indivisible trailing part */
+    for(; length; data++, length--) {
+        if(*data) {
```
ThomasWaldmann: equivalent, but maybe a little prettier: use *data++ here instead of increment in for statement?

src/borg/_chunker.c
```diff
@@ -236,6 +235,96 @@ chunker_fill(Chunker *c)
...
+
+#else
+
+static const int chunker_fast_zero = 0;
+
+static int
+chunker_is_zero(const uint8_t *data, size_t length)
+{
+    for(; length; data++, length--) {
+        if(*data) {
```
ThomasWaldmann: equivalent, but maybe a little prettier: use *data++ here instead of increment in for statement?

src/borg/_chunker.c
ThomasWaldmann: also: declare __m128i * p = (_m128i *) data once to save 4 times casting here?

src/borg/_chunker.c
```diff
@@ -313,10 +313,20 @@ chunker_is_zero(const uint8_t *data, ssize_t length)
 
 static const int chunker_fast_zero = 0;
 
+/* performance (Haswell, E3-1231v3): in cache about ~16 GB/s, w/ real memory access about 12 GB/s */
+
 static int
 chunker_is_zero(const uint8_t *data, size_t length)
 {
-    for(; length; data++, length--) {
+    // TODO: alignment issues?
```
ThomasWaldmann: some ARM chips might be problematic here

src/borg/_chunker.c
```diff
@@ -313,10 +313,20 @@ chunker_is_zero(const uint8_t *data, ssize_t length)
...
+    while (length >= 8) {
+        if(*((uint64_t *)data)) {
+            return 0;
+        }
+        data += 8;
+        length -= 8;
+    }
+    for (; length; length--, data++) {
         if(*data) {
```
ThomasWaldmann: keep both for loops similar as now, but use *p++ with correct pointer type?

src/borg/_chunker.c
```diff
@@ -335,6 +335,12 @@ chunker_is_zero(const uint8_t *data, size_t length)
 }
 #endif
 
+/*
+ * performance (Haswell, E3-1231v3)
+ * without chunker_is_zero (SSE), about 710-720 MB/s
+ * with    "                "   , about 690-700 MB/s
```
ThomasWaldmann: if it is faster without, why have it?

src/borg/_crc32/clmul.c
```diff
@@ -516,6 +516,11 @@ crc32_clmul(const unsigned char *src, long len, uint32_t initial_crc)
...
     xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_crc1);
 
+    /*
+     * could just as well write xmm_crc3[2], doing a movaps and truncating, but
+     * no real advantage - it's a tiny bit slower per call, while no additional CPUs
+     * would be supported by only requiring SSSE3 and CLMUL instead of SSE4.1 + CLMUL
+     */
     crc = _mm_extract_epi32(xmm_crc3, 2);
     return ~crc;
 }
```
ThomasWaldmann: add a LF

src/borg/archiver.py
```diff
@@ -1137,23 +1137,37 @@ def do_debug_info(self, args):
         return EXIT_SUCCESS
 
     def do_debug_speed(self, args):
+        # This is a really primitive facility, but still useful -- within it's limits.
```
ThomasWaldmann: its

src/borg/_chunker.c
enkore: in the current usage it shouldn't be a problem, since the beginning of a bytes object seems to be always word-aligned (I noticed that small bytes() are consistently aligned on odd 8 bytes on x64; larger ones are aligned to >=16 bytes), but it's something to easily trip over so the code should ultimately just deal with that.

src/borg/_chunker.c
enkore: i did some practical testing and I can't see a sizeable difference in pure dedup scenarios between HEAD and this branch, so it seems that the lenmod change offsets this additional check completely.

There is btw. another enhancing change in that region, by hardcoding lenmod = 31 (4095 & 0x1f) I get to 750-760 MB/s in this branch, but it naturally breaks the tests, which use a window size < 4095. I think the reason might be that x86 has a separate coding for `rol` where the count is immediate, not through a register, which seems to be faster.

src/borg/_chunker.c
enkore: i think writing the index explicitly is easier to parse (xmm0 = ... data, xmm1 = ... data + 1 etc.). Also allows to have the block size (64) explicit, which makes it clear that it has to be the same in all conditions.

enkore:
I removed all changes related to the chunker. I believe the approach is viable and it does indeed provide a useful increase (when fully realized it processes sparse areas at about 3 GB/s which is still pretty poor but much better than 200-300 MB/s at best), however it's not a use case I have and making it sufficiently nice-looking and clear that it does not regress the chunker is not something I felt the need to spend time on.

This leaves only the CRC-related changes.

PR #1971, opened 2016-12-21
new --pattern and --patterns-from options

leo-b:
Second approach to add include patterns via two new options: `--pattern` and `--patterns-from`.


borg/archiver.py
```diff
@@ -1353,6 +1394,13 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('--keep-tag-files', dest='keep_tag_files',
                                action='store_true', default=False,
                                help='keep tag files of excluded caches/directories')
+        subparser.add_argument('--pattern', dest='pattern',
+                               type=parse_inclexcl_pattern, action='append',
+                               metavar="PATTERN", help='include/exclude paths matching PATTERN')
```
enkore: Nice

borg/archiver.py
```diff
@@ -1353,6 +1394,13 @@ def build_parser(self, args=None, prog=None):
...
+        subparser.set_defaults(pattern=[])
```
enkore: Would it make sense to move this default (+line 1482) to the top-level default declaration in line 1137?

(Nitpick: Shouldn't this variable be named `patterns` (plural)?)

borg/helpers.py
```diff
@@ -312,17 +312,37 @@ def load_excludes(fh):
...
+def update_patterns(args):
+    """Merge patterns from exclude- and pattern-files with those on command line."""
+    for file in args.pattern_files:
+        roots, inclexclpatterns = load_patterns(file)
+        args.paths += roots
+        args.pattern += inclexclpatterns
+        file.close()
+    for file in args.exclude_files:
+        args.pattern += load_excludes(file)
+        file.close()
```
enkore: I think it should be documented in the help in which order these are applied, like

> Patterns and excludes from the command line are considered first, then patterns
> from --pattern-from, then exclusion patterns from --exclude-from.

borg/archiver.py
```diff
@@ -121,6 +121,18 @@ def print_file_status(self, status, path):
         if self.output_list and (self.output_filter is None or status in self.output_filter):
             logger.info("%1s %s", status, remove_surrogates(path))
 
+    @staticmethod
+    def build_matcher(inclexcl_patterns, paths):
+        matcher = PatternMatcher()
+        if inclexcl_patterns:
+            matcher.add_inclexcl(inclexcl_patterns)
```
enkore: One thing that's not entirely clear to me here is how/where RootPaths are segregated when given on the command line.

Suppose: `borg create --pattern "R /foo"`, wouldn't that result in an empty args.paths?

borg/archiver.py
```diff
@@ -978,7 +981,42 @@ def do_break_lock(self, args, repository):
...
+            # can be rebuild
+            - /home/*/.cache
+            # they're downloads for a reason
+            - /home/*/Downloads
+            # susan is a nice person
+            # include susans home
+            + /home/susan
+            # don't backup the other home directories
+            - /home/*
+
```
enkore: Good docs

borg/helpers.py
```diff
@@ -506,6 +535,34 @@ def parse_pattern(pattern, fallback=FnmatchPattern):
...
+        '+': True,
+        'R': RootPath,
+        'r': RootPath,
+    }
+    ptype = None
+    if len(pattern) > 1 and pattern[0] in type_prefix_map:
+        (ptype, pattern) = (type_prefix_map[pattern[0]], pattern[1:])
+        pattern = pattern.lstrip()
+    if ptype is None or not pattern:
+        raise argparse.ArgumentTypeError("Unable to parse pattern: {}".format(pattern))
```
enkore: Perhaps writing this in [EAFP](https://docs.python.org/3/glossary.html#term-eafp)-style is easier to read? Decide for yourself.

```py
try:
    ptype = type_prefix_map[pattern[0]]
except (IndexError, KeyError):
    raise argparse.ArgumentTypeError(...)
pattern = pattern[1:].lstrip()  # below the ptype= might be clearer, although this doesn't raise.
```

enkore:
Fixes #1406

borg/archiver.py
leo-b: Oops. That part was missing.
I added a  custom argparse action class now:
https://github.com/leo-b/borg/blob/7d2799107164e1c21b218077fa06dd8b115686c9/borg/helpers.py#L348-L358

borg/archiver.py
leo-b: OK. Changed it.

borg/helpers.py
leo-b: I added a help text:
https://github.com/leo-b/borg/blob/7d2799107164e1c21b218077fa06dd8b115686c9/borg/archiver.py#L1019-L1021

Maybe the inclusion of files (`--patterns-from` and `--exclude-from`) should also be done in order of appearance on the commandline instead of adding them after having processed the commandline options `--patterns` and `--exclude`? (I.e. replace `update_patterns()` with a custom argparser class that processes the file directly when the commandline option is parsed.)
However this would change the current behavior...

borg/helpers.py
leo-b: Changed it.

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+        root path starts with the prefix `R`, followed by a path (a plain path, not a
+        file pattern). An include rule is specified by `+` followed by a pattern.
+        Exclude rules start with a `-`.
+        Inclusion patterns are useful to e.g. exclude the contents of a directory
+        except for some important files in this directory. The first matching pattern
+        is used so if an include pattern matches before an exclude pattern, the file
+        is backed up.
+
+        Note that the default pattern style for `--pattern` and `--patterns-from` is
+        shell style (`sh:`), so those patterns behave like rsync include/exclude patterns.
```
ThomasWaldmann: Not sure we should document "like rsync" here. People could take this as a guarantee that we are and will always be compatible.

borg/archiver.py
ThomasWaldmann: Not sure we want to include BOFH style examples. :D

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+            - /home/*/Downloads
+            # susan is a nice person
+            # include susans home
+            + /home/susan
+            # don't backup the other home directories
+            - /home/*
+
+        Patterns (`--pattern`) and excludes (`--exclude`) from the command line are
+        considered first (in the order of appearance). Then patterns from `--pattern-from`
+        are added. Exclusion patterns from `--exclude-from` files are appended last.
```
ThomasWaldmann: Same here: move the ordering info to the more general description (above)?

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+
+\n\n''')
```
ThomasWaldmann: `\n\n'''` should be the end of the last non-empty line.

borg/archiver.py
```diff
@@ -1353,6 +1398,11 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('--keep-tag-files', dest='keep_tag_files',
                                action='store_true', default=False,
                                help='keep tag files of excluded caches/directories')
+        subparser.add_argument('--pattern', action=ArgparsePatternAction,
+                               metavar="PATTERN", help='include/exclude paths matching PATTERN')
```
ThomasWaldmann: this option might be repeated quite some times, so it would be nice to have a (non-conflicting) short option here.

borg/helpers.py
```diff
@@ -312,17 +312,50 @@ def load_excludes(fh):
     both line ends are ignored.
     """
     patterns = (line for line in (i.strip() for i in fh) if not line.startswith('#'))
-    return [parse_pattern(pattern) for pattern in patterns if pattern]
+    return [parse_exclude_pattern(pattern)
+            for pattern in patterns if pattern]
```
ThomasWaldmann: why did you change?

borg/helpers.py
```diff
@@ -312,17 +312,50 @@ def load_excludes(fh):
...
-        for file in args.exclude_files:
-            args.excludes += load_excludes(file)
-            file.close()
+def load_patterns(fh):
+    """Load and parse include/exclude/root patterns from file object.
+    Lines empty or starting with '#' after stripping whitespace on both line ends are ignored.
+    """
+    patternlines = (line for line in (i.strip() for i in fh) if not line.startswith('#'))
+    roots = []
+    inclexclpatterns = []
```
ThomasWaldmann: inclexcl_patterns for better consistency?

borg/helpers.py
```diff
@@ -312,17 +312,50 @@ def load_excludes(fh):
...
+    for patternline in patternlines:
+        pattern = parse_inclexcl_pattern(patternline)
+        if pattern:
```
ThomasWaldmann: what is this "if" checking? we always get back a 2 element namedtuple, so it is always truish?

borg/helpers.py
```diff
@@ -312,17 +312,50 @@ def load_excludes(fh):
...
+            if pattern.ptype is RootPath:
+                roots.append(pattern.pattern)
+            else:
+                inclexclpatterns.append(pattern)
+    return roots, inclexclpatterns
+
+
+def update_patterns(args):
+    """Merge patterns from exclude- and pattern-files with those on command line."""
```
ThomasWaldmann: ... exclude and pattern files ...

borg/helpers.py
```diff
@@ -312,17 +312,50 @@ def load_excludes(fh):
...
+        file.close()
+
+
+class ArgparsePatternAction(argparse.Action):
+    def __init__(self, nargs=1, **kw):
+        super().__init__(nargs=nargs, **kw)
+
+    def __call__(self, parser, args, values, option_string=None):
+        pattern = parse_inclexcl_pattern(values[0])
+        if pattern:
```
ThomasWaldmann: what is this "if" checking? we always get back a 2 element namedtuple, so it is always truish?

borg/testsuite/helpers.py
```diff
@@ -445,6 +446,93 @@ def evaluate(filename):
...
+      "# comment"], ["/"], 0),
+    (["# comment",
+      "- /data",
+      "R /home"], ["/home"], 1),
+])
+def test_load_patterns_from_file(tmpdir, lines, expected_roots, expected_numpatterns):
+    def evaluate(filename):
+        roots, inclexclpatterns = load_patterns(open(filename, "rt"))
+        return roots, len(inclexclpatterns)
+    patternfile = tmpdir.join("exclude.txt")
```
ThomasWaldmann: patterns.txt? (change all places where it makes sense)

borg/helpers.py
enkore: Yes, that sounds like a better, easier understood behaviour.

I'm not sure it would change current behaviour, since ordering doesn't matter with exclusions?

borg/archiver.py
```diff
@@ -893,21 +894,22 @@ def do_break_lock(self, args, repository):
 
     helptext = collections.OrderedDict()
     helptext['patterns'] = textwrap.dedent('''
-        Exclusion patterns support four separate styles, fnmatch, shell, regular
-        expressions and path prefixes. By default, fnmatch is used. If followed
+        File patterns support four separate styles, fnmatch, shell, regular
```
leo-b: OK.

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
             re:^/home/[^/]\.tmp/
             sh:/home/*/.thumbnails
             EOF
-            $ borg create --exclude-from exclude.txt backup /\n\n''')
+            $ borg create --exclude-from exclude.txt backup /
+
+            # exclude the contents of /data/docs/ but not /data/docs/pdf
+            $ borg create -e +/data/docs/pdf -e /data/docs/ backup /
+            # equivalent:
+            $ borg create -e +pm:/data/docs/pdf -e -pm:/data/docs/ backup /
```
leo-b: Ah - this should be removed. (It doesn't work anyways, only `--pattern` supports the new syntax.)

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+
+
+        A more general way to define filename matching patterns may be passed via
+        `--pattern` and `--patterns-from`. Using these options, you may specify the
```
leo-b: OK.

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+        backup roots (starting points) and patterns for inclusion/exclusion. A
+        root path starts with the prefix `R`, followed by a path (a plain path, not a
+        file pattern). An include rule is specified by `+` followed by a pattern.
```
leo-b: OK.

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+        Exclude rules start with a `-`.
+        Inclusion patterns are useful to e.g. exclude the contents of a directory
+        except for some important files in this directory. The first matching pattern
```
leo-b: OK.

borg/archiver.py
```diff
@@ -978,7 +981,46 @@ def do_break_lock(self, args, repository):
...
+        is used so if an include pattern matches before an exclude pattern, the file
+        is backed up.
```
leo-b: It already comes rather soon after the introduction of `--pattern`. Where would you like it to be placed?

borg/archiver.py
leo-b: I think the rsync syntax is well known among the users of borg and thus the reference to rsync will be useful for borg newbies. Would "similar to" be a better wording?
If it once turns out that borgs syntax diverges from rsync syntax, we can still remove the rsync reference...

borg/archiver.py
leo-b: I think the usage example should be pratice-oriented and meaningful. The `/home/susan` example will easily make sense to every sysadmin, that's why I like it. I don't think the reader will resent it.

Since the ordering rules affect the the mix of `--pattern` and  `--exclude` options, it should be a separate paragraph. But maybe it should be moved above the example?

borg/archiver.py
leo-b: OK.

borg/archiver.py
leo-b: Suggestions? `-p` and `-P` are already in use. Maybe `-M` for *match*?

borg/helpers.py
leo-b: Because exclude patterns are now a special case of general patterns. `parse_pattern()` is now used to parse both `--pattern` and `--exclude` file-patterns and therefore a small wrapper `parse_exclude_pattern()` was needed for the `--exclude` case. (Like `parse_inclexcl_pattern()` is used for `--pattern`.)

borg/helpers.py
leo-b: OK.

borg/helpers.py
leo-b: Maybe I was planning to use it for empty or comment-lines?
But I removed it now.

borg/helpers.py
leo-b: I think the hyphen is important here.
I intentionally used it to build compound words: `exclude-file` and `pattern-file`.
https://en.oxforddictionaries.com/punctuation/hyphen#hyphens_in_compound_words
http://www.grammarbook.com/punctuation/hyphens.asp

borg/helpers.py
leo-b: Removed. (Even though I don't see why this defensive approach could be disturbing in any way.)

borg/testsuite/helpers.py
leo-b: OK.

borg/archiver.py
enkore: Documentation is versioned for a reason; we can't strive to make docs that are eternally valid. I think it's reasonably unlikely that a pattern type is significantly changed after releasing it.

ThomasWaldmann: Except if you are sure matching behaviour is 100% "like rsync": either remove it or say "similar to".

borg/archiver.py
ThomasWaldmann: of course it would be great if we have same words for short and long options.

match and -M sounds good, but then the other options would be "--matches-from"?

"match" could be taken as action and mean positive and negative matching, but --matches-from sounds wrong because matches is a noun there and it does not contain matches, but patterns for matching.

so, guess we did not find best wording yet.

borg/helpers.py
ThomasWaldmann: oh, i did not even see you changed the function name. X-)
what i asked about was the formatting.
and if you would not change that, the real change would be more obvious.

borg/helpers.py
ThomasWaldmann: hmm, i am not a native english speaker but i thought hyphens are used rather rarely in that case, but inserted frequently by native german speakers because it is like that in german.

borg/archiver.py
enkore: If we can't find a verb/noun-pair that makes sense and starts with a free letter we can also just make the abbreviated option non-sensical -- but then it should be really far off, so we don't block of "useful" letters, ie. things like `-w` or `-z`.

enkore:
Subthread hidden by Github, still relevant: https://github.com/borgbackup/borg/pull/1971#discussion_r95541369

leo-b:
The inclusion of files (`--patterns-from` and `--exclude-from`) is now done in order of appearance on the commandline.

borg/archiver.py
```diff
@@ -2030,7 +2026,10 @@ def parse_args(self, args=None):
             args = self.preprocess_args(args)
         parser = self.build_parser(args)
         args = parser.parse_args(args or ['-h'])
-        update_patterns(args)
+        if args.func == self.do_create:
+            # need at least 1 path but args.paths may also be populated from patterns
```
ThomasWaldmann: did you check the requirements of recreate?

borg/helpers.py
```diff
@@ -307,52 +307,60 @@ def parse_timestamp(timestamp):
...
-            roots.append(pattern.pattern)
-        else:
-            inclexcl_patterns.append(pattern)
-    return roots, inclexcl_patterns
+def pattern_file_iter(fileobj):
+    for line in fileobj:
+        line = line.strip()
+        if not line or line.startswith('#'):
+            continue
+        yield line
```
ThomasWaldmann: iirc i have added similar code to master branch, please just backport that utility function, so we are in sync.

borg/archiver.py
leo-b: _recreate_ is not part of 1.0-maint, so it is not relevant here.
However for the recreate subcommand in master, the `paths` arguments are optional. So from the parsers point of view, nothing needs to be changed. However, the resulting args.paths will in future be a mixture of `R /pathname` arguments from pattern-files and `paths` commandline arguments, so the user will need to be careful that the same combination of root paths will be specified at recreate time.

borg/helpers.py
leo-b: done

ThomasWaldmann:
if we could finish this soon and be rather sure about it, we could consider it for 1.0.10rc1.

ThomasWaldmann:
@leo-b do you remember, did we discuss already about whether this should go into 1.0-maint or master?

leo-b:
> @leo-b do you remember, did we discuss already about whether this should go into 1.0-maint or master?

@ThomasWaldmann we didn't. I proactively added it to 1.0-maint. :-)

leo-b:
What are the open issues?
Finding a proper one-letter option for ``--pattern`` is still unsolved...
Are there any other things to do?

ThomasWaldmann:
letters / option name
decide 1.0 or 1.1/master.
if we want it in 1.0, we should be pretty sure we have no incompatible changes (like peoples existing excludes/includes behaving differently now). if unsure, maybe more tests could increase confidence.

leo-b:
> letters / option name

Maybe a non-letter option like ``-%``? :-)

> decide 1.0 or 1.1/master.
> if we want it in 1.0, we should be pretty sure we have no incompatible changes (like peoples existing excludes/includes behaving differently now).

I'd vote for 1.0. Our goal was to design ``--pattern`` options so that ``--exclude`` options won't be affected. And this should be seen through to the finish. :-)

> if unsure, maybe more tests could increase confidence.

Any suggestions for additional tests?
There are already tests for ``--exclude``, for ``--pattern``, for ``--excludes-from``, ``--patterns-from``, for commandlines with explicit roots, with implicit roots via `--pattern R ...`` and for missing root paths.
Maybe the existing tests should be extended by some more corner-case test patterns? However, I think I have a too biased view to contribute more useful additions.. :-)

ThomasWaldmann:
No non-alpha options, could cause troubles.

Do we have tests for order of processing?

leo-b:
> Do we have tests for order of processing?

- for ``--exclude``: No, we don't. Exclude rules are order-independent. However, I've included a simple test.
- for ``--pattern``: I've added one.

borg/testsuite/helpers.py
```diff
@@ -512,9 +517,14 @@ def test_load_invalid_patterns_from_file(tmpdir, lines):
...
       "-/data"],
      ["/data/something00.txt", '/home', '/home/leo', '/home/leo/t', '/home/other']),
+    # include /home/leo and exclude the rest of /home:
     (["+/home/leo",
       "-/home/*"],
      ['/data', '/data/something00.txt', '/data/subdir/something01.txt', '/home', '/home/leo', '/home/leo/t']),
+    # wrong order, /home/leo is already excluded by -/home/*:
+    (["-/home/*",
+      "+/home/leo"],
+     ['/data', '/data/something00.txt', '/data/subdir/something01.txt', '/home']),
```
ThomasWaldmann: yup, like these ones ^^^.

borg/testsuite/helpers.py
```diff
@@ -445,6 +453,104 @@ def evaluate(filename):
...
+    patternfile = tmpdir.join("patterns.txt")
+    with patternfile.open("wt") as fh:
+        fh.write("\n".join(lines))
+    filename = str(patternfile)
+    with pytest.raises(argparse.ArgumentTypeError):
+        roots = []
+        inclexclpatterns = []
+        load_pattern_file(open(filename, "rt"), roots, inclexclpatterns)
+
+
```
ThomasWaldmann: above, you use modes "rt" and "wt" often (which is same as "r" and "w").

maybe adapt the style to the one you see elsewhere, which I suspect is the shorter style.

ThomasWaldmann:
As general style / procedure questions:

We now have `--exclude*` - as this adds include patterns as a possibility in `--pattern*`, do we also want `--include*` for symmetry reasons?

Do we want to deprecate `--exclude*` (and `--include*`) as `--pattern*` somehow supercedes them?

enkore:
> We now have --exclude* - as this adds include patterns as a possibility in --pattern*, do we also want --include* for symmetry reasons?

Hmnah, don't really see the advantage there. If it's little code, so that it's easy to see that it works as expected (which should be the case though), ok...

> Do we want to deprecate --exclude* (and --include*) as --pattern* somehow supercedes them?

No, this is finicky to get right at the script level and we should not force people to revisit this, possibly breaking stuff, maybe even without noticing. <trump>Very Bad</trump> :)

leo-b:
> We now have --exclude* - as this adds include patterns as a possibility in --pattern*, do we also want --include* for symmetry reasons?

No. Includes are only useful in combination with prior excludes so I think the ``--pattern`` syntax will be much clearer as the close relation of those patterns will be perceptible by using the same option name.

> Do we want to deprecate --exclude* (and --include*) as --pattern* somehow supercedes them?

No. ``--exclude`` is internally represented the same way as a ``- ...`` pattern, so keeping the ``--exclude`` option is not very error prone. And ``--exclude`` is currently widely used and is rather self-explanatory.

borg/testsuite/helpers.py
leo-b: I copied that from the test-case just above my one. :-)

ThomasWaldmann:
@leo-b ok, I think we are done here. \o/

Could you practically use this code until we merge it into 1.0.11?

enkore:
I'm just gonna merge this so it doesn't need another rebase/merge here.

We can add a single-char option for --pattern when we come up with a good idea

ThomasWaldmann:
ehrm, so this now would also get into 1.0.10 although not in its milestone.
at least if i do the release (which was planned for tonight) as usual from 1.0-maint tip, as usual.

enkore:
Ew, yeah, I messed that up. Thought we wanted to put it into 1.0.10, not 1.0.11.

ThomasWaldmann:
I will make a release branch anyway, so i could branch off the previous changeset and we merge it with 1.0-maint later.

For the patterns addition, I'ld like to have it in a RC before we release it, but I don't like to delay 1.0.10 (there are some fixes we should get out ASAP), so it will be a 1.0.11 rc.

enkore:
Ok

leo-b:
There was a bug in --patterns-from parsing.
I've fixed it and added a test case:
https://github.com/leo-b/borg/commit/fefe59589e1e048703d186720404f29a6902c866

Should I open a new issue for this?

enkore:
You can create a PR directly; since it was not released no need for an issue.

borg/archiver.py
```diff
@@ -1399,7 +1440,7 @@ def build_parser(self, args=None, prog=None):
         subparser.add_argument('location', metavar='ARCHIVE',
                                type=location_validator(archive=True),
                                help='name of archive to create (must be also a valid directory name)')
-        subparser.add_argument('paths', metavar='PATH', nargs='+', type=str,
+        subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
```
leo-b: Unfortunately changing nargs from '+' to '*' here may trigger a bug in pythons argparse (https://bugs.python.org/issue15112) under special circumstances.
If an optional argument (``--anything``) is placed on the commandline between the positional arguments `location` and `paths`, path-arguments are not populated correctly.

ThomasWaldmann: ok, good to know. not a showstopper I'ld say, it can be just placed differently to get it working. 

Usually I'ld place options to the left, sometimes (when editing a commandline) it is also convenient if they can be appended, but putting them in the middle is a bit unusual anyway.

leo-b: Here is a minimal example that illustrates the problem:

```python
    import argparse
    parser = argparse.ArgumentParser()
    parser.add_argument('--foo', action='store_true')
    parser.add_argument('location')

    # works:
    # parser.add_argument('paths', nargs='+')
    # doesn't work:
    parser.add_argument('paths', nargs='*')
    
    args = parser.parse_args('1 --foo 2'.split())
```
With `nargs='*'`, this raises `unrecognized arguments: 2`

enkore:
```
borg create testrepo::{now}  -vp src
usage: borg [-h] [-V] <command> ...
borg: error: unrecognized arguments: src
```

leo-b: We could:

1. monkey-patch argparse: Unfortunately the bug resides in a closure inside a rather long method: https://hg.python.org/cpython/file/tip/Lib/argparse.py#l1771
We could replace the whole method or just the inner function (consume_positionals()): http://stackoverflow.com/a/27550228

2. use a hack to work around the bug, e.g. something like that:
```python
    import argparse

    class PatchedArgumentParser(argparse.ArgumentParser):
        def _parse_known_args(self, arg_strings, namespace):
            """work around a bug in argparse: https://bugs.python.org/issue15112"""
            arg = getattr(self, '_patch_arg', None)
            if arg:
                saved = arg.nargs, arg.required
                arg.nargs, arg.required = argparse.ONE_OR_MORE, False
            try:
                ret = super()._parse_known_args(arg_strings, namespace)
            finally:
                if arg:
                    arg.nargs, arg.required = saved
            return ret

    if __name__ == '__main__':
        parser = PatchedArgumentParser(prog='PROG')
        parser.add_argument('--foo', action='store_true')

        parser.add_argument('location')
        arg = parser.add_argument('paths', nargs='*')
        parser._patch_arg = arg

        parser.print_help()

        # args = parser.parse_args('1 --foo 2'.split())
        for argstr in ('1 --foo 2',
                       '--foo 1 2',
                       '1 --foo',
                       '--foo'):
            print(argstr)
            args = parser.parse_args(argstr.split())
            print("    {!r}".format(args))
```
3. do nothing and live with the fact that some commandlines have to be re-arranged

enkore: > 3. do nothing and live with the fact that some commandlines have to be re-arranged

i.e. not releasing it in 1.0.x which isn't too unreasonable given that it's a fairly large and complex new feature

leo-b: > i.e. not releasing it in 1.0.x which isn't too unreasonable given that it's a fairly large and complex new feature (also there is a tangible release plan for 1.1 now)

That's not what I ment. :-)
But what about:
- Release the `--pattern` feature in 1.0.x but leave `nargs="+"` in 1.0.x to remain full commandline compatibility. (So at least one path has to be specified on the commandline.)

enkore: That sounds like the best course of action to me

ThomasWaldmann: the use case for not having to specify any paths is if all roots are given e.g. in a patterns file, right?

so, would `borg create --patterns-from-file=patt.txt repo::arch /doesnotexist` be a workaround?
does it emit a warning about that then? is rc == 1 because of that?

enkore: yes/yes/yes

ThomasWaldmann: ok, so not really a great workaround. i guess some people will want to have all in a separate file, so they can use same script on their different hosts - and deal with all roots, all excludes/includes in the patterns file. and still not being unnerved by warnings and rc==1 all the time.

so, just ignore the argparse bug and add a ticket or faq item about it?

enkore: I'm not so sure that writing things like `repo --options paths` is as uncommon as you think. I have for example... `$create::hostname-$date $OPTS -e ... /some /paths` in a script, that / similar patterns are perhaps not as uncommon.

enkore: Hm.

nargs=+ keeps current scripts working (which I think we should really strive for _especially_ in point releases), but means minor inconvenience for users of a new feature.

ThomasWaldmann: @enkore ok, that sounds good.

@leo-b i don't understand why your argparse hack works (arg is not being used in super call, so what does it do?). guess we rather don't want the other approach (monkeypatching it). Didn't yet look into argparse code.

ThomasWaldmann: @leo-b ah, ok, clearer now. @enkore how about that? doesn't sound too bad.

enkore: Ah I see... should have  a few tests.

leo-b: This is another approach to fix the argparse bug.
It uses the proposed fix from Steven Bethard and monkey patches the inner function `consume_positionals()` in a rather defensive way. (Revert all changes if a test fails after patching.)

```python
# https://bugs.python.org/issue15112
import argparse

# Thanks to Martijn Pieters
# http://stackoverflow.com/a/27550237/3456480
def monkeypatch_inner_function(outer, new_inner):
    """Replace a nested function code object used by outer with new_inner

    The replacement new_inner must use the same name and must at most use the
    same closures as the original.

    """
    if hasattr(new_inner, '__code__'):
        # support both functions and code objects
        new_inner = new_inner.__code__
    # find original code object so we can validate the closures match
    ocode = outer.__code__
    function, code = type(outer), type(ocode)
    iname = new_inner.co_name
    orig_inner = next(
        const for const in ocode.co_consts
        if isinstance(const, code) and const.co_name == iname)
    # you can ignore later closures, but since they are matched by position
    # the new sequence must match the start of the old.
    assert (orig_inner.co_freevars[:len(new_inner.co_freevars)] ==
            new_inner.co_freevars), 'New closures must match originals: %s'.format(orig_inner.co_freevars)
    # replace the code object for the inner function
    new_consts = tuple(
        new_inner if const is orig_inner else const
        for const in ocode.co_consts)
    # create a new function object with the new constants
    # note: python 2 code objects don't have co_kwonlyargcount
    c = code(ocode.co_argcount, ocode.co_kwonlyargcount, ocode.co_nlocals,
             ocode.co_stacksize, ocode.co_flags, ocode.co_code, new_consts,
             ocode.co_names, ocode.co_varnames, ocode.co_filename,
             ocode.co_name, ocode.co_firstlineno, ocode.co_lnotab,
             ocode.co_freevars, ocode.co_cellvars)
    return function(
        c,
        outer.__globals__, outer.__name__, outer.__defaults__,
        outer.__closure__)

def create_argparse_consume_positionals():
    # this is a copy of consume_positionals from python 3.6
    # with a proposed fix for bug https://bugs.python.org/issue15112
    # https://hg.python.org/cpython/file/3.6/Lib/argparse.py#l1916    
    # https://bugs.python.org/file26485/15112.patch
    self = None
    arg_strings_pattern = None
    positionals = None
    arg_strings = None
    take_action = None
    def consume_positionals(start_index):
        # match as many Positionals as possible
        match_partial = self._match_arguments_partial
        selected_pattern = arg_strings_pattern[start_index:]
        arg_counts = match_partial(positionals, selected_pattern)

        # if we haven't hit the end of the command line strings,
        # then don't consume any final zero-width arguments yet
        # (we may need to parse some more optionals first)
        if start_index + sum(arg_counts) != len(arg_strings_pattern):
            while arg_counts and arg_counts[-1] == 0:
                arg_counts.pop()

        # slice off the appropriate arg strings for each Positional
        # and add the Positional and its args to the list
        for action, arg_count in zip(positionals, arg_counts):
            args = arg_strings[start_index: start_index + arg_count]
            start_index += arg_count
            take_action(action, args)

        # slice off the Positionals that we just parsed and return the
        # index at which the Positionals' string args stopped
        positionals[:] = positionals[len(arg_counts):]
        return start_index
    return consume_positionals

def argparse_has_positional_bug():
    parser = argparse.ArgumentParser()
    parser.add_argument('--foo', action='store_true')
    parser.add_argument('pos1')
    arg = parser.add_argument('pos2', nargs='*')
    args, unrecognized = parser.parse_known_args("1 --foo 2".split())
    return True if unrecognized == [ '2' ] else False

def maybe_fix_argparse():
    if argparse_has_positional_bug():
        # print("has bug")
        consume_positionals = create_argparse_consume_positionals()
        origmethod = argparse.ArgumentParser._parse_known_args
        try:
            argparse.ArgumentParser._parse_known_args = monkeypatch_inner_function(
                origmethod, consume_positionals)
            if argparse_has_positional_bug():
                # print("still has bug")
                raise Exception()
            else:
                # print("fixed")
                return True
        except Exception:
            # fix didn't work..
            # strategic withdrawal!
            argparse.ArgumentParser._parse_known_args = origmethod
            return False
    else:
        # no need to fix
        return None



if __name__ == '__main__':
    import sys
    result = maybe_fix_argparse()
    print("{!r}".format(result))
```

Issue #1978, opened 2016-12-23
change-passphrase crashes for plaintext repos


enkore:
It doesn't work -- as promised -- but it's kinda ugly as well.

```
AttributeError: 'PlaintextKey' object has no attribute 'change_passphrase'
```

Should be:

```
$repo is not encrypted, cannot change passphrase.
(exit 1 or 2)
```

enkore:
Fixed by #1994

PR #1994, opened 2017-01-03
Fix upgrade --tam, change-passphrase crashes if not encrypted

### LINKED ISSUE #1981: add non-encrypted repo check for --tam repo upgrades (or at least better docs)
Hi, while reading the release notes for the new 1.0.9 version, I saw I need to run a command similar to --tam upgrade or sth like that. It upgrades the repo to make it more secure.

The problem is the docs don't say specifically that the repo needs to have some encryption. For non-encrypted repo-s the upgrade command fails with some python error that looks ugly.

Two simple things could remedy this situation:
1. Add one sentence to docs/release notes that specifies we need to upgrade ONLY repos with encryption.
2. Add a simple if-check for the --tam option that would print "this repo is not encrypted, you don't need to run this upgrade at all. Bye bye..."

Agree?

enkore:
Fixes #1981 #1978

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -182,6 +182,9 @@ class Archiver:
             return EXIT_ERROR
         key.change_passphrase()
         logger.info('Key updated')
+        if hasattr(key, 'find_key'):
+            # print key location to make backing it up easier
+            logger.info('Key location: %s', key.find_key())
         return EXIT_SUCCESS
 
     @with_repository(lock=False, exclusive=False, manifest=False, cache=False)
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -177,6 +177,9 @@ class Archiver:
     @with_repository()
     def do_change_passphrase(self, args, repository, manifest, key):
         """Change repository key file passphrase"""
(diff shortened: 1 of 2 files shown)


borg/archiver.py
```diff
@@ -730,6 +730,10 @@ def do_upgrade(self, args, repository, manifest=None, key=None):
         if args.tam:
             manifest, key = Manifest.load(repository, force_tam_not_required=args.force)
 
+            if not hasattr(key, 'change_passphrase'):
+                print('This repository is not encrypted, cannot enable TAM.')
+                return 2
```
ThomasWaldmann: return EXIT_ERROR

borg/archiver.py
```diff
@@ -177,6 +177,9 @@ def do_check(self, args, repository):
     @with_repository()
     def do_change_passphrase(self, args, repository, manifest, key):
         """Change repository key file passphrase"""
+        if not hasattr(key, 'change_passphrase'):
+            print('This repository is not encrypted, cannot change the passphrase.')
+            return 2
```
ThomasWaldmann: return EXIT_ERROR

PR #1995, opened 2017-01-03
helpers.Buffer: raise OSError subclass if too much memory shall be al…

enkore:


borg/helpers.py
```diff
@@ -722,7 +732,7 @@ def resize(self, size, init=False):
         """
```
ThomasWaldmann: docstring above needs updating.

borg/helpers.py
```diff
@@ -722,7 +732,7 @@ def resize(self, size, init=False):
...
         size = int(size)
         if self.limit is not None and size > self.limit:
-            raise ValueError('Requested buffer size %d is above the limit of %d.' % (size, self.limit))
+            raise Buffer.OutOfMemory(size, self.limit)
```
ThomasWaldmann: is this enough to initialize the OSError, so accessing .errno, .path, .strerror works? (we access them when transforming OSError to our special error class for input file os errors).

ThomasWaldmann:
btw, the exception change in the previous PR could be reverted I guess.

borg/helpers.py
enkore: Yes

borg/helpers.py
```diff
@@ -699,6 +705,10 @@ class Buffer:
     """
     provide a thread-local buffer
     """
+
+    class OutOfMemory(Error, OSError):
```
ThomasWaldmann: btw, this class name is a bit misleading: we are not out of memory, we just reject to go beyond that as we suspect there is some malfunction if we get this far. BufferSizeLimitExceeded?

enkore: my line of thinking was that `Buffer` has a fixed "pool size" so to speak; trying to allocate more memory than that means that there is no more `Buffer` memory available.

ThomasWaldmann: That would be also preferable. In case a user ever sees that in a traceback, he should not think he has not enough RAM in the machine.

PR #2000, opened 2017-01-04
check: fail if single archive does not exist

### LINKED ISSUE #1997: borg check does not exit with failure status when archive is absent
when running a command such as:
borg check -v repo::archive1           (archive1 is not actually part of the repo)
...if archive1 does not exist in the repo, the command should report an error ( that archive "archive1" does not exist ), then exit with a non-zero status.
Instead, it says the archive consistency check is complete(see below), and exits with a 0 status.

Starting repository check
Completed repository check, no problems found.
Starting archive consistency check...
Orphaned objects check skipped (needs all archives checked).
Archive consistency check complete, no problems found.
[root@chlxfs1 trc]# echo $?
0

enkore:
Fixes #1997 

I didn't change the other branch to issue a warning/error if --prefix doesn't match anything or if there are not >=N archives with --last N because this might actually break scripts.

Edit: Made the described cases print a warning but nothing else (esp. not an exit code).

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -1100,8 +1100,12 @@ class ArchiveChecker:
                                    key=lambda name_info: name_info[1][b'time'])
             if prefix is not None:
                 archive_items = [item for item in archive_items if item[0].startswith(prefix)]
+                if not archive_items:
+                    logger.warning('--prefix %s does not match any archives', prefix)
             num_archives = len(archive_items)
             end = None if last is None else min(num_archives, last)
+            if last is not None and end < last:
+                logger.warning('--last %d archives: only found %d archives', last, end)
         else:
             # we only want one specific archive
             archive_items = [item for item in self.manifest.archives.items() if item[0] == archive]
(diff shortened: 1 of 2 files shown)


borg/archive.py
```diff
@@ -1100,13 +1100,21 @@ def valid_item(obj):
...
             archive_items = [item for item in self.manifest.archives.items() if item[0] == archive]
             num_archives = 1
             end = 1
+            if not archive_items:
+                logger.error('Archive %s does not exist', archive)
+                self.error_found = True
+                return
 
         with cache_if_remote(self.repository) as repository:
             for i, (name, info) in enumerate(archive_items[:end]):
```
enkore: btw. why don't we just slice earlier?

ThomasWaldmann: because we would need to do it in both branches of the if/else?

ThomasWaldmann:
can you rebase -i and just use "non-exit-code-changing" directly? having such a special char in a commit comment and an abbreviation noone understands is causing more trouble that having a slightly longer comment.

PR #2006, opened 2017-01-06
Docs: language clarification - VM backup FAQ

sherbang:
Patch speaks for itself.  Just a very minor cleanup of some language I tripped on while reading the FAQ.  The for/before replacement being the most important part.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -13,7 +13,7 @@ Yes, the `deduplication`_ technique used by
 Also, we have optional simple sparse file support for extract.
 
 If you use non-snapshotting backup tools like Borg to back up virtual machines,
-then these should be turned off for doing so. Backing up live VMs this way can (and will)
+then the VMs should be turned off for the duration of the backup. Backing up live VMs can (and will)
 result in corrupted or inconsistent backup contents: a VM image is just a regular file to
 Borg with the same issues as regular files when it comes to concurrent reading and writing from
 the same file.


docs/faq.rst
```diff
@@ -13,7 +13,7 @@ Yes, the `deduplication`_ technique used by
 Also, we have optional simple sparse file support for extract.
 
 If you use non-snapshotting backup tools like Borg to back up virtual machines,
-then these should be turned off for doing so. Backing up live VMs this way can (and will)
+then the VMs should be turned off before doing so. Backing up live VMs can (and will)
```
ThomasWaldmann: Hmm, I am not a native english speaker, but doesn't this change semantics or at least create an ambiguity?

be turned off for doing so == machine is turned off while backup is running.

be turned off before doing so == machine is turned off before backup is started, but unclear if it can be turned on after it is started.

sherbang: As a native (US) English speaker, I found the "turned off for doing so" language difficult to parse.  I eventually assumed "for" was intended to be "before" and then it made sense.

"turned off for the duration of the backup" would be clearer and would get across your point that it must stay off.

BTW: As a long time Attic user, I just discovered Borg yesterday.  Thank you for your efforts to revitalize development.

ThomasWaldmann: Strictly taken one could also shut down, snapshot, turn on, make a backup of the snapshot while machine is already running again, remove snapshot. But without snapshot, the VM should stay off.

Yes, "turned off for the duration of the backup" is clearest, guess we should take that.

About development: you're welcome, we enjoy it. :) If you have old attic archives in the (converted) repo, make sure you use latest borg release 1.0.9 (see changelog).

enkore: You can totally do a block-level snapshot of a running VM, it's just that this gives crash-consistency which can be anywhere from annoying to disastrous when restoring.

ThomasWaldmann: @sherbang can you update wording and PR against 1.0-maint branch, so the docs for the 1.0.x releases get updated (stuff will then get merged into master from there).

ThomasWaldmann:
thanks!

docs/faq.rst
sherbang: @ThomasWaldmann Updated, thanks.

@enkore If you're not doing some kind of snapshotting then the backup tool will be reading blocks in different state of change, as other blocks are being changed in the VM.  This results in a VERY inconsistent image state that's backed-up.  If the backup operation was an atomic operation then it would work fine, but that's what snapshots end up substituting for.

enkore: ... do a block-level snapshot of a running VM, ... just that ...

Issue #1999, opened 2017-01-04
exclude-if-present could check with .exists() instead of .isfile()


costela:
Hi,

Is there a reason why `--exclude-if-present` explicitly checks if its argument is a file? I couldn't find any rationale looking into #395, but maybe I'm missing something.
If there isn't one, then maybe `.exists(...)` would be a more generic check? It would allow doing things like `--exclude-if-present .git`.

Cheers

ThomasWaldmann:
Sounds like basically a good idea, but would have these consequences:
- it would need more general processing, currently it backups the "tag" as a file if "keep tags" mode is active
- for "keep tags" mode it would be somehow strange to keep all the `.git/*` but not the workdir (ok if checkout clean, but not if unclean). it would be also strange to only keep an empty `.git`.

costela:
I'm not sure I understand what you mean.
- if `keep_tag_files` is set, borg currently calls `_process` recursively on the tags, which then get treated appropriately as a file or directory (see [here](https://github.com/borgbackup/borg/blob/902c4230868ebbafc675e1cc3e4e891c45c0aead/src/borg/archiver.py#L415)). Or did I misunderstand your objection?
- as for the "strangeness": I don't think it would be that weird, especially in the case of git, since the .git folder get's backed-up recursively as a folder (as explained above) and not left empty. I.e.: doing a backup with `--keep-tag-files` and `--exclude-if-present .git` would be basically the same as having a backup of the bare git repo, which might make sense, since it might mean a significant size reduction, depending on the repository (in my tests with a small 48MB repo I got a 32MB backup with the options above and 46MB without).

ThomasWaldmann:
Oh, right, it is _process already for the tags, not process_dir - so that's not a problem.

But still it feels like a problematic hack. For .git workdirs this would exclude all **uncommitted/unadded** changes (which are not likely to exist anywhere else - this is different than for added/committed/pushed changes which might also exist on github or whereever).

OTOH, UNIX tools policy is not like standing in the way and making things impossible that the admin wants (and that might make sense in some scenarios), so we could implement that and the admin can feel free to use that for good or shoot himself/his users in the foot.

ThomasWaldmann:
Like this for implementing any filesystem object (files, dirs, ...) as "tag file".

ThomasWaldmann:
Like this for keeping as is: only files can be tag files.

costela:
[...]

PR #2010, opened 2017-01-08
change dir_is_tagged to use os.path.exists()

costela:
Very small commit to make tagging accept folders as tags and allow things like
`--exclude-if-present .git`.
This is being discussed in #1999

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -126,6 +126,18 @@ The best check that everything is ok is to run a dry-run extraction::
 Changelog
 =========
 
+Version 1.1.0b4 (not released yet)
+----------------------------------
+
+New features:
+
+- the --exclude-if-present option now supports tagging a folder with any
+  filesystem object type (file, folder, etc), instead of expecting only files
+  as tags, #1999
+- the --keep-tag-files option has been deprecated in favor of the new 
+  --keep-exclude-tags, to account for the change mentioned above.
+
+
 Version 1.1.0b3 (2017-01-15)
 ----------------------------
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -346,7 +346,7 @@ class Archiver:
                 else:
                     restrict_dev = None
                 self._process(archive, cache, matcher, args.exclude_caches, args.exclude_if_present,
-                              args.keep_tag_files, skip_inodes, path, restrict_dev,
+                              args.keep_exclude_tags, skip_inodes, path, restrict_dev,
                               read_special=args.read_special, dry_run=dry_run, st=st)
             if not dry_run:
                 archive.save(comment=args.comment, timestamp=args.timestamp)
@@ -382,7 +382,7 @@ class Archiver:
         return self.exit_code
 
     def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
-                 keep_tag_files, skip_inodes, path, restrict_dev,
+                 keep_exclude_tags, skip_inodes, path, restrict_dev,
                  read_special=False, dry_run=False, st=None):
         if not matcher.match(path):
             self.print_file_status('x', path)
@@ -419,11 +419,11 @@ class Archiver:
             if recurse:
                 tag_paths = dir_is_tagged(path, exclude_caches, exclude_if_present)
                 if tag_paths:
-                    if keep_tag_files and not dry_run:
+                    if keep_exclude_tags and not dry_run:
                         archive.process_dir(path, st)
                         for tag_path in tag_paths:
                             self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
-                                          keep_tag_files, skip_inodes, tag_path, restrict_dev,
+                                          keep_exclude_tags, skip_inodes, tag_path, restrict_dev,
                                           read_special=read_special, dry_run=dry_run)
                     return
             if not dry_run:
@@ -438,7 +438,7 @@ class Archiver:
                     for dirent in entries:
                         normpath = os.path.normpath(dirent.path)
                         self._process(archive, cache, matcher, exclude_caches, exclude_if_present,
-                                      keep_tag_files, skip_inodes, normpath, restrict_dev,
+                                      keep_exclude_tags, skip_inodes, normpath, restrict_dev,
                                       read_special=read_special, dry_run=dry_run)
         elif stat.S_ISLNK(st.st_mode):
             if not dry_run:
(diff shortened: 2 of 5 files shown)


ThomasWaldmann:
Hmm, 2 tests broken currently.

costela:
Oops, sorry about that. Should be passing now (the Xcode tests are still in the queue).

src/borg/testsuite/archiver.py
```diff
@@ -932,16 +932,20 @@ def _create_test_keep_tagged(self):
         self.create_regular_file('taggedall/%s' % CACHE_TAG_NAME,
                                  contents=CACHE_TAG_CONTENTS + b' extra stuff')
         self.create_regular_file('taggedall/file4', size=1024)
+        self.create_regular_file('taggeddir/.NOBACKUP2/file5', size=1024)
+        self.create_regular_file('taggeddir/file6', size=1024)
```
ThomasWaldmann: as we now consider tags no matter whether they are files or dirs, maybe just modify line925 to cover the dirs case (and drop "taggeddir")?

tagged1..n dirs were meant to cover the n different ways to tag.

taggedall was meant to have all tagging methods present at once, maybe also modify that accordingly.

ThomasWaldmann:
Hmm, 1.0-maint or master?

enkore:
master due to changed behaviour?

ThomasWaldmann:
Guess we must use master (and introduce it in 1.1), because in 1.0 it is documented specifically as acting on **files**:
```
  --exclude-if-present FILENAME
                        exclude directories that contain the specified file
  --keep-tag-files      keep tag files of excluded caches/directories
 ```
Also, these docs need to be updated so they do not just refer to "tag files", "FILENAME", "specified file", ...

**Comment by costela** [comment]
@ThomasWaldmann, sorry for the delayed reaction. Just pushed a new commit with the suggested changes to testing + small corrections to the help texts. Probably not optimal, but still better than nothing! ;)

File: src/borg/archiver.py [hunk 2]
```diff
@@ -1962,7 +1962,7 @@ def build_parser(self, prog=None):
                                    help='exclude directories that contain the specified file')
```
ThomasWaldmann: also change above?

src/borg/archiver.py
```diff
@@ -2550,7 +2550,7 @@ def build_parser(self, prog=None):
...
```
ThomasWaldmann: also change above?

costela:
Last two suggestions incorporated (plus a minor clarification for the `--keep-tag-files` option).
Still a bit unfortunate that the 'files' in the option's name doesn't quite reflect the option's semantics, but I guess it's not worth some API-braking rename or the introduction of a new option.

ThomasWaldmann:
@costela yes, i noticed the option name issue also.

We could improve that like this:
- invent a good, additional option name that triggers the new behaviour
- keep the old option and keep its old behaviour, but deprecate it and issue a deprecation warning, so we can remove it (in 1.1, in 1.2 or whatever future version)

enkore:
Seems like an awful lot of hassle for such a small change.

costela:
@enkore how about a compromise?
Keep the new behavior, but add a new `--keep-tag` option that acts as an alias to `--keep-tag-file`. Deprecate `--keep-tag-file` and remove it some time in the future. This way the slight semantic problem gets fixed eventually, without having to add too much code to this minor change.

enkore:
ok

I mainly don't want to have all the code duplicated.

ThomasWaldmann:
@costela sounds good. there is already some code for dealing with option deprecations, use it.

costela:
Another attempt: I took the liberty of bikeshedding a bit further and using the name `--keep-exclude-tags`, also making the help text a bit more explicit. I also added some testing for the deprecation.

src/borg/archiver.py
```diff
@@ -1959,10 +1960,11 @@ def build_parser(self, prog=None):
                                         'http://www.brynosaurus.com/cachedir/spec.html)')
         exclude_group.add_argument('--exclude-if-present', dest='exclude_if_present',
                                    metavar='FILENAME', action='append', type=str,
```
ThomasWaldmann: this still says FILENAME. maybe rather PATH?

costela: ah, good catch!

docs/changes.rst
```diff
@@ -175,6 +175,9 @@ Other changes:
 - make RPC protocol more extensible: use named parameters.
 - RemoteRepository: misc. code cleanups / refactors
 - clarify cache/repository README file
+- the --keep-tag-files option has been deprecated in favor of --keep-exclude-tags.
+  additionally, both options now also accept directories as tags for backup
+  exclusion, #1999
```
ThomasWaldmann: this should first talk about --exclude-if-present now just checking for the existance of a filesystem object of that name (not requiring the tag to be a file, but also accepting e.g. a directory).
the keeping behaviour should be described afterwards.

docs/usage/create.rst.inc
```diff
@@ -36,10 +36,10 @@ Exclusion options
         | read exclude patterns from EXCLUDEFILE, one per line
     ``--exclude-caches``
         | exclude directories that contain a CACHEDIR.TAG file (http://www.brynosaurus.com/cachedir/spec.html)
-    ``--exclude-if-present FILENAME``
-        | exclude directories that contain the specified file
-    ``--keep-tag-files``
-        | keep tag files of excluded caches/directories
+    ``--exclude-if-present PATH``
```
ThomasWaldmann: while writing above comment, i noticed i was wrong suggesting the usage of "PATH" here, sorry.
it must not be a path (like /a/b/c), just a NAME.

docs/usage/create.rst.inc
```diff
@@ -36,10 +36,10 @@ Exclusion options
...
+        | exclude directories that contain the specified file/directory
```
ThomasWaldmann: ... that are tagged by containing a filesystem object of the given NAME

docs/usage/create.rst.inc
```diff
@@ -36,10 +36,10 @@ Exclusion options
...
+    ``--keep-exclude-tags``
+        | keep tag files/directories (i.e.: arguments to --exclude-if-present) in otherwise excluded caches/directories
```
ThomasWaldmann: keep tag objects...

docs/usage/recreate.rst.inc
```diff
@@ -36,10 +36,10 @@ Exclusion options
...
-        | exclude directories that contain the specified file
-    ``--keep-tag-files``
-        | keep tag files of excluded caches/directories
+    ``--exclude-if-present PATH``
+        | exclude directories that contain the specified file/directory
+    ``--keep-exclude-tags``
+        | keep tag files/directories (i.e.: arguments to --exclude-if-present) in otherwise excluded caches/directories
 
 Archive options
     ``--target TARGET``
```
ThomasWaldmann: maybe do not update the .rst.inc file at all, these files are usually created / updated at release time.

changes in here also often cause merge conflicts e.g. when merging one branch into another.

docs/changes.rst
ThomasWaldmann: also, "both options now also accept directories as tags" is a bit misleading in the context of "--keep-tag-files option" and "--keep-exclude-tags" - these options do not get any NAME as value, but are just booleans.

so, this refers to the --exclude-if-present NAME options rather.

ThomasWaldmann:
oh, the changelog entry should be replaced into b4 section, b3 is already released.

costela:
Incorporated latest suggestions and rebased on latest master. Now let's hope nothing breaks ;)

PR #2011, opened 2017-01-08
Vagrantfile: Split cygwin packaged to be installed into base and project

htgoebel:
The packages which are required for ssh log-in and rsync synchronized
folders to work are separated from those specific for the borg
project. This makes it easier to reuse in other projects.


ThomasWaldmann:
thanks!

Issue #1253, opened 2016-07-03
Extend hostname in lock roster with additional metadata?


enkore:
Recap: hostnames are often not unique. When detecting stale locks this is required however. It would be good to augment the hostname with something which is usually unique.

Earlier discussion:
https://github.com/borgbackup/borg/pull/1246#issuecomment-230149750

Also #562 as the "master issue" regarding locking.

enkore:
Fixed by #2014 (for now? Should be good enough, but if someone has better idea(s)).

PR #2014, opened 2017-01-11
posix: use fully-qualified hostname + node ID

enkore:
The node ID is usually the 48 bit MAC of the primary network
interface.

This pretty much kills off any possibility of colliding hostnames unless one is bent on doing that (same hostnames, spoofing MAC etc.).


ThomasWaldmann:
Fixes #1253, hopefully.

enkore:
I didn't even remember that issue! :)

PR #2017, opened 2017-01-12
Update 1.1.0b3 CHANGES

enkore:


docs/changes.rst
```diff
@@ -129,20 +129,26 @@ Changelog
 Version 1.1.0b3 (not released yet)
 ----------------------------------
 
+Removed features:
```
ThomasWaldmann: use "Compatibility notes:" as in the past?

enkore: yep

PR #2021, opened 2017-01-12
Merge/1.0 maint

enkore:
Many spurious conflicts, wonder what's up with that.


tox.ini
```diff
@@ -9,7 +9,7 @@ deps =
      -rrequirements.d/development.txt
      -rrequirements.d/attic.txt
      -rrequirements.d/fuse.txt
-commands = py.test -rs --cov=borg --cov-config=.coveragerc --benchmark-skip --pyargs {posargs:borg.testsuite}
+commands = py.test -n 8 -rs --cov=borg --cov-config=.coveragerc --benchmark-skip --pyargs {posargs:borg.testsuite}
```
enkore: Could also use "-n auto" here.

ThomasWaldmann:
Seems like we worked in parallel on that.
Yes, merging that was a PITA.
I'll review to see if we came out with the same stuff.

src/borg/archive.py
```diff
@@ -1360,16 +1360,18 @@ def valid_item(obj):
             sort_by = sort_by.split(',')
             if any((first, last, prefix)):
                 archive_infos = self.manifest.archives.list(sort_by=sort_by, prefix=prefix, first=first, last=last)
+                if not archive_infos:
+                    logger.warning('--first/--last/--prefix did not match any archives')
```
ThomasWaldmann: i have a bit more here to handle less-than-expected cases for --first and --last.

tox.ini
ThomasWaldmann: i used 2x corecount to make sure stuff is fully loaded.

src/borg/archive.py
enkore: Picked this and the logger changes from your branch

ThomasWaldmann:
tell me when you are finished taking the good pieces of mine.

PR #2027, opened 2017-01-12
crc: disable CLMUL version due to clang bugs.

### LINKED ISSUE #2026: crc32 on darwin64 compilation issues
See there: https://paste.thinkmo.de/irxJPVVn#darwin64-borg-compile.txt

enkore:
Should fix #2026

Note that the type errors from the paste in #2026 are follow-up errors because the ISA extension is disabled and hence no intrinsics are defined, so it assumes `int ... ()` as the function prototype, which is of course wrong.

### DIFF
--- a/src/borg/_crc32/crc32.c
+++ b/src/borg/_crc32/crc32.c
@@ -3,6 +3,11 @@
 #include "slice_by_8.c"
 
 #ifdef __GNUC__
+/*
+ * clang also has or had GCC bug #56298 explained below, but doesn't support
+ * target attributes or the options stack. So we disable this faster code path for clang.
+ */
+#ifndef __clang__
 #if __x86_64__
 /*
  * Because we don't want a configure script we need compiler-dependent pre-defined macros for detecting this,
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
hmm, did we test the original change on travis OS X? or was it broken at the time?

enkore:
It was broken at the time and didn't get that far. I tested on Linux with clang from Arch, there it worked, but with some warnings (which is why there are some ifdefs that are now removed); so it's not _all_ clang versions, just some, or it might depend on other things as well.

enkore:
Extrapolating from Travis Status we'll have results tomorrow morning :)

ThomasWaldmann:
Lots of IntegrityErrors on OS X.

enkore:
Hm.

enkore:
ARE YOU SHITTING ME?! How the *literal f\*\*\** did f482c3242 fix this?!

enkore:
> won't we need that at some time later for windows port? ^^^

Only if we target MSVC which will take rather many changes anyway (eg. no void* arithmetic which we have tons of).

enkore:
Ah. I have an idea how this might have come about: If Xcode-clang has neither __BYTE_ORDER nor __BIG_ENDIAN then both would be replaced with `0` and `0 == 0` *is* true, obviously. Hence it compiled the code for BE on LE. With our battle-proven byte-order detection from _hashindex.c it then works.

ThomasWaldmann:
OK, that sounds less strange than having __LITTLE_ENDIAN at least.

PR #2032, opened 2017-01-13
borg serve: fix transmission data loss of pipe writes, fixes #1268

### LINKED ISSUE #1268: RemoteArchiverTestCase failures on cygwin (1.0-maint)
Quite some of these tests either hang or fail:

```
py.test -v -k Remote
```

The corresponding local tests work:

```
py.test -v -k "not Remote"    # ... and not bench
```

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/35765312-remotearchivertestcase-failures-on-cygwin-1-0-maint)

ThomasWaldmann:
this problem was found on cygwin/windows due to its small pipe buffer size of 64kiB.

due to that, bigger (like >64kiB) writes are always only partially done and os.write() returns
the amount of data that was actually sent. the code previously did not use that return value
and assumed that always all is sent, which led to a loss of the remainder of transmission data
and usually some "unexpected RPC data format" error on the client side.

On all other operating systems (Linux, *BSD, OS X), this error was not seen.

### DIFF
--- a/borg/remote.py
+++ b/borg/remote.py
@@ -7,6 +7,7 @@ import shlex
 import sys
 import tempfile
 import textwrap
+import time
 from subprocess import Popen, PIPE
 
 from . import __version__
@@ -26,6 +27,23 @@ RPC_PROTOCOL_VERSION = 2
 BUFSIZE = 10 * 1024 * 1024
 
 MAX_INFLIGHT = 100
+
+
+def os_write(fd, data):
+    """os.write wrapper so we do not lose data for partial writes."""
+    # This is happening frequently on cygwin due to its small pipe buffer size of only 64kiB
+    # and also due to its different blocking pipe behaviour compared to Linux/*BSD.
+    # Neither Linux nor *BSD ever do partial writes on blocking pipes, unless interrupted by a
+    # signal, in which case serve() would terminate.
+    amount = remaining = len(data)
+    while remaining:
+        count = os.write(fd, data)
+        remaining -= count
+        if not remaining:
+            break
+        data = data[count:]
+        time.sleep(count * 1e-09)
+    return amount
 
 
 class ConnectionClosed(Error):
@@ -106,7 +124,7 @@ class RepositoryServer:  # pragma: no cover
                     if self.repository is not None:
                         self.repository.close()
                     else:
-                        os.write(stderr_fd, "Borg {}: Got connection close before repository was opened.\n"
+                        os_write(stderr_fd, "Borg {}: Got connection close before repository was opened.\n"
                                  .format(__version__).encode())
                     return
                 unpacker.feed(data)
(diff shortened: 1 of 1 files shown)


enkore:
Neither Linux nor BSD ever do partial writes on blocking pipes, unless interrupted by a signal, in which case `serve` would terminate.

ThomasWaldmann:
@enkore I updated the source and commit comment accordingly.

borg/remote.py
```diff
@@ -28,6 +29,21 @@
...
+def os_write(fd, data):
+    """os.write wrapper so we do not lose data for partial writes."""
+    # This is happening frequently on cygwin due to its small pipe buffer size of only 64kiB
+    # and also due to its different blocking pipe behaviour compared to Linux/*BSD.
+    # Neither Linux nor *BSD ever do partial writes on blocking pipes, unless interrupted by a
+    # signal, in which case serve() would terminate.
+    amount = len(data)
+    while data:
+        count = os.write(fd, data)
+        data = data[count:]
```
enkore: Maybe use a memoryview instead, I believe this would copy the data.

Also, move the "if not data: return amount" before this, ie. "if not count: return amount", to avoid the extra object creation in the normal case.

I think the sleep shouldn't be necessary... or does have cygwin such an interesting notion of blocking, that it continuously returns 0 for a blocking pipe if it's full?

ThomasWaldmann: I thought about memoryview, but considering these 64kiB writes and sleep waiting is rather inefficient anyway I thought I keep it simple.

Using count == 0 as termination criteria: i don't think this is the correct criteria. Also, even if it would never return 0 if there was something to send, it would need an extra, empty write call to return 0.

I added the sleep to avoid too fast looping there. If it does a partial write, it means the buffer is currently full, so it might be rather pointless to just try again immediately without waiting a bit and give it a chance to empty the buffer.

ThomasWaldmann: self.to_send at other places in remote.py is also not a memoryview and used in a similar way.

enkore: > Using count == 0 as termination criteria: i don't think this is the correct criteria. Also, even if it would never return 0 if there was something to send, it would need an extra, empty write call to return 0.

Yes it'd have to be "not amount"; updating amount. to_send uses a somewhat similar construction, it also does copies. Might be an avenue for some minor optimization.

> I added the sleep to avoid too fast looping there. If it does a partial write, it means the buffer is currently full, so it might be rather pointless to just try again immediately without waiting a bit and give it a chance to empty the buffer.

Given that cygwin is quite strange here already this might be reasonable; it *does* interpret "blocking" quite different from all other OSes as you have found, so it might just become non-blocking when the pipe's buffer is full.

(Which I think would be 1-2 cygwin bugs).

ThomasWaldmann: @enkore hmm, do you think the OS scheduler's behaviour makes the sleep() unnecessary?

enkore: My intention here is that the code path taken by OSes actually used (Linux, BSDs+OSX) remains mostly unchanged and doesn't do much extra work like copying things around, but not using them.

ThomasWaldmann: We need amount unchanged to return it in the end, to simular os.write() api for the caller.

ThomasWaldmann: btw, i linked to the ticket and this pr in emails to the cygwin mailing list. so, hopefully someone reads this and fixes any bugs in cygwin.

enkore:
```

def os_write(fd, data):
    """os.write wrapper so we do not lose data for partial writes."""
    # This is happening frequently on cygwin due to its small pipe buffer size of only 64kiB
    # and also due to its different blocking pipe behaviour compared to Linux/*BSD.
    # Neither Linux nor *BSD ever do partial writes on blocking pipes, unless interrupted by a
    # signal, in which case serve() would terminate.
    amount = remaining = len(data)
    while data:
        count = os.write(fd, data)
        remaining -= count
        if not remaining:
            return amount
        data = data[count:]
   return amount
```

ThomasWaldmann: ok, good. I'll just rather do "break" there instead of return. And add the sleep.

PR #2084, opened 2017-01-19
Binaries with py353

ThomasWaldmann:
also: tests / binaries with osxfuse 3.5.4


ThomasWaldmann:
Did a wheezy-based test, it worked.

PR #2085, opened 2017-01-20
vagrant freebsd: some fixes, fixes #2067

### LINKED ISSUE #2067: freebsd: use the 10.3-RELEASE vagrant image
because it can be updated with binary downloads.

don't use the 10.3-STABLE.

ThomasWaldmann:
- use -RELEASE, it can be updated via binaries
- more RAM, otherwise the 4 workers run out of memory.
- do not install / use fakeroot, it seems broken.
- set a hostname, this VM has none

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -83,11 +83,13 @@ end
 
 def packages_freebsd
   return <<-EOF
+    # VM has no hostname set
+    hostname freebsd
     # install all the (security and other) updates, base system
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
```
==> freebsd64:   py34: commands succeeded
==> freebsd64:   py35: commands succeeded
==> freebsd64:   py36: commands succeeded
==> freebsd64:   flake8: commands succeeded
==> freebsd64:   congratulations :)
```

PR #2068, opened 2017-01-15
sdist: remove *.so/*.dll

enkore:
Fixes #2066

PR #2089, opened 2017-01-21
Manifest.in: simplify, also exclude *.so *.dll *.orig

### LINKED ISSUE #2066: binaries in pypi package
Just noticed that when there are binaries (*.so) below src/borg, they will end up in the pypi package when running `python setup.py sdist`.

Is that new? Are we expected to exclude them in MANIFEST.in?

ThomasWaldmann:
fixes #2066, supercedes #2068, PRed against 1.0-maint (merge into master later)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,10 +1,8 @@
 include README.rst AUTHORS LICENSE CHANGES.rst MANIFEST.in
-recursive-include borg *.pyx
-recursive-include docs *
-recursive-exclude docs *.pyc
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
btw, i tested it with looking at the python setup.py sdist result package, it seems ok.

so, merge?

enkore:
Idea for later: add a step to one of the vagrant machines that

- makes an sdist
- looks at it, checks that no +x files are in there
- installs it in a new virtualenv
- does "borg help"

PR #2094, opened 2017-01-22
Fix invalid hard links

### LINKED ISSUE #1297: Race between storing hardlink master and later hardlink references
https://github.com/borgbackup/borg/blob/1.0-maint/borg/archive.py#L625

- inodes are recycled, so an (dev, inode) pair doesn't really identify a file during a backup run
- If there's an error later on, then `self.hard_links[st.st_ino, st.st_dev] = safe_path` would have created a reference to an item that didn't make it into the archive
- I'm not sure, it may be possible to sneak in a reference to file contents with a stat result that doesn't belong to them
  1. have file with inode x, contents y and permission z
  2. have a hardlink there
  3. get rid of the inode (to delete an inode one may not have the necessary permissions to read the contents)
  4. create new file with permissions you like, lodge it so that they have the same inode, create another dirent for the file so that you get nlink > 1
  5. assuming you manage the proper timing:
     - the file (x, y, z) will end up in the archive as usual
    - the new file (x, y, z') has the same contents (due to being recognized as a hard link), the ones of file (x, y, z), but permissions you can control z'
    - This is kinda low-impact though. Not only requires that a very peculiar permissions layout, but also precise timing and the ability to control/predict inodes (which may be not hard or may be very hard)

### LINKED ISSUE #2092: inaccessible hard links are stored
similar to: https://github.com/jborg/attic/issues/393
```
$ touch hl1
$ chmod 000 hl1
$ ln hl1 hl2
$ borg create repo::hl-check .
hl1: [Errno 13] Permission denied: 'hl1'
$ mkdir extract
$ cd extract
$ borg extract ../repo::hl-check
hl2: [Errno 2] No such file or directory: '/home/tw/w/borg/extract/hl1' -> '/home/tw/w/borg/extract/hl2'
$ ls -l hl*
ls: cannot access 'hl*': No such file or directory
$ borg list ../repo::hl-check | grep hl
h--------- tw     tw            0 Sat, 2017-01-21 20:08:04 hl2 link to hl1
```

enkore:
Fixes #2092 

Does not fix #1297 (it's a different race but a race nonetheless)

### DIFF
--- a/borg/fuse.py
+++ b/borg/fuse.py
@@ -128,15 +128,21 @@ class FuseOperations(llfuse.Operations):
                 else:
                     self.items[inode] = item
                     continue
-                segments = prefix + os.fsencode(os.path.normpath(item[b'path'])).split(b'/')
-                del item[b'path']
+                path = item.pop(b'path')
+                segments = prefix + os.fsencode(os.path.normpath(path)).split(b'/')
                 num_segments = len(segments)
                 parent = 1
                 for i, segment in enumerate(segments, 1):
                     # Leaf segment?
                     if i == num_segments:
                         if b'source' in item and stat.S_ISREG(item[b'mode']):
-                            inode = self._find_inode(item[b'source'], prefix)
+                            try:
+                                inode = self._find_inode(item[b'source'], prefix)
+                            except KeyError:
+                                file = path.decode(errors='surrogateescape')
+                                source = item[b'source'].decode(errors='surrogateescape')
+                                logger.warning('Skipping broken hard link: %s -> %s', file, source)
+                                continue
                             item = self.cache.get(inode)
                             item[b'nlink'] = item.get(b'nlink', 1) + 1
                             self.items[inode] = item
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -660,8 +660,6 @@ Number of files: {0.stats.nfiles}'''.format(
                 self.add_item(item)
                 status = 'h'  # regular file, hardlink (to already seen inodes)
                 return status
-            else:
-                self.hard_links[st.st_ino, st.st_dev] = safe_path
         is_special_file = is_special(st.st_mode)
         if not is_special_file:
             path_hash = self.key.id_hash(os.path.join(self.cwd, path).encode('utf-8', 'surrogateescape'))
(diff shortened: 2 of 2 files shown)


borg/fuse.py
```diff
@@ -136,7 +136,13 @@ def process_archive(self, archive, prefix=[]):
                     # Leaf segment?
                     if i == num_segments:
                         if b'source' in item and stat.S_ISREG(item[b'mode']):
-                            inode = self._find_inode(item[b'source'], prefix)
+                            try:
+                                inode = self._find_inode(item[b'source'], prefix)
+                            except KeyError:
```
ThomasWaldmann: that's a rather wide error handler, would catch all KeyErrors in _find_inode also.

borg/fuse.py
```diff
@@ -136,7 +136,13 @@ def process_archive(self, archive, prefix=[]):
...
+                                file = segment.decode(errors='surrogateescape')
```
ThomasWaldmann: segment has only the basename, right?
we could also save item[b'path'] into a local var above 132 (or pop() it) and use it here.

ThomasWaldmann: otherwise the user will not know the full path of the hardlink we had to kill.

ThomasWaldmann:
one osx travis job had strange failures, restarted it...

borg/fuse.py
enkore: fixed

borg/fuse.py
```diff
@@ -129,14 +129,20 @@ def process_archive(self, archive, prefix=[]):
...
                 parent = 1
                 for i, segment in enumerate(segments, 1):
                     # Leaf segment?
                     if i == num_segments:
                         if b'source' in item and stat.S_ISREG(item[b'mode']):
-                            inode = self._find_inode(item[b'source'], prefix)
+                            try:
+                                inode = self._find_inode(item[b'source'], prefix)
+                            except KeyError:
+                                file = path.decode(errors='surrogateescape')
```
ThomasWaldmann: isn't path unicode already? see 131.

enkore: No. fs *encode*.

ThomasWaldmann: ok

borg/fuse.py
```diff
@@ -129,14 +129,20 @@ def process_archive(self, archive, prefix=[]):
                     self.items[inode] = item
                     continue
                 segments = prefix + os.fsencode(os.path.normpath(item[b'path'])).split(b'/')
-                del item[b'path']
+                path = item.pop(b'path')
```
ThomasWaldmann: move to above 131 and use it in 131' already.

PR #2096, opened 2017-01-22
SyncFile: fix use of fd object after close

ThomasWaldmann:
maybe not a problem, but cleaner like that in any case.


ThomasWaldmann:
trivial enough, merging...

PR #2109, opened 2017-01-28
vagrant: improve darwin64 VM settings

ThomasWaldmann:
somehow without these cpuid settings it does not work for everybody.
also nice if we can get away without the extensions pack, which is proprietary.
do not update iTunes, we just want the OS security / bugfix updates


ThomasWaldmann:
the darwin vagrant vm still worked as expected, thus I am merging this.

PR #2111, opened 2017-01-29
Merge 1.0-maint

ThomasWaldmann:
also: refactored posix_fadvise checks / calls into safe_fadvise calls to avoid code duplication.


ThomasWaldmann:
Fixed the issues that showed up on travis / OS X.

After review / approval, I'ld collapse the 3 last commits, so there is only 1 merge commit in the end.

ThomasWaldmann:
can we merge this soon, before it starts to conflict?

PR #2114, opened 2017-01-29
Add 'debug dump-manifest' and 'debug dump-archive' commands.

textshell:
This adds two new commands for low level debugging and maybe analysis. These are debug commands without stability guarantees.

The output is the decrypted contents of the manifest translated from msgpack to json for manifest and similar for archive. For archive the json has a synthetic top level object that combines the metadata from the manifest about the archive with the archive msgpack item and the items stream.

I don‘t think this will duplicate #1946 / #1812 as those are meant to be a nice and higher level interface while this is a debugging and development aid.

```
# borg debug dump-manifest /tmp/borg-repo /dev/stdout
{
    "item_keys": [
        "acl_access",
        "acl_default",
        "acl_extended",
        "acl_nfs4",
        "atime",
        "bsdflags",
        "chunks",
        "chunks_healthy",
        "ctime",
        "gid",
        "group",
        "hardlink_master",
        "mode",
        "mtime",
        "part",
        "path",
        "rdev",
        "source",
        "uid",
        "user",
        "xattrs"
    ],
    "archives": {
        "test1": {
            "time": "2017-01-29T16:36:23.687059",
            "id": "\u007fea93c330cd9782770f0c02c9799e408a8a3b8a182a834fbb34bc4eeecf470466"
        }
    },
    "tam": {
        "type": "HKDF_HMAC_SHA512",
        "hmac": "\u007f8874645345ee77804b78e833a23c91c98086d3f2381036e44c5d7e94947637446cd8166605da454144ad507f8e2b8ea7e4b64e598da3958e67215e62f06fa6d8",
        "salt": "\u007f4ae27b21b7b209664e413ce84c83a772132effd0a7d68fedf77f69dc52b4457e08e91b8aba38494bd9f0921698f92ac5dde8db2a843b6ea4d40747de078e2adf"
    },
    "version": 1,
    "config": {},
    "timestamp": "2017-01-29T16:36:23.728885"
}
```

and

```
# borg debug dump-archive /tmp/borg-repo::test1 /dev/stdout
{
    "id": "\u007fea93c330cd9782770f0c02c9799e408a8a3b8a182a834fbb34bc4eeecf470466",
    "time": "2017-01-29T16:36:23.687059",
    "_name": "test1",
    "_meta": {
        "username": "martin",
        "version": 1,
        "name": "test1",
        "time": "2017-01-29T16:36:23.687059",
        "items": [
            "\u007f8b57530fda213ee55c4ffb7d6ee661977987aa80e0069264efb6e8297a123006"
        ],
        "cmdline": [
            "/home/martin/git/borg/borg-env/bin/borg",
            "create",
            "/tmp/borg-repo::test1",
            "tox.ini"
        ],
        "hostname": "martin",
        "time_end": "2017-01-29T16:36:23.728687",
        "comment": "",
        "chunker_params": [
            19,
            23,
            21,
            4095
        ],
        "tam": {
            "type": "HKDF_HMAC_SHA512",
            "hmac": "\u007fc56cf2a15b44cf6ae410de60a22653597ab2c122dd571039a0955bb518bfc492314ce7dffe69864a374690e77ee97e0db9d585d13a8da3ac1c0b39563c2a8bd4",
            "salt": "\u007fdb57ccdb2acd3a6221e65e13a82f8e6c2c2d4c7dbb3db0fba719898d59381bd8b787d685828bc921228ee22bf6ddfe34eba1646da668357b33354036a9927097"
        }
    },
    "_items": [
        {
            "user": "martin",
            "hardlink_master": false,
            "gid": 1000,
            "atime": 1485706267000000000,
            "uid": 1000,
            "mtime": 1485706177000000000,
            "group": "martin",
            "ctime": 1485706177000000000,
            "path": "tox.ini",
            "chunks": [
                [
                    "\u007f4a4d98a6d4e6e0667af3084190a18823d93e8013d19b11b9c8f352182f15735a",
                    517,
                    520
                ]
            ],
            "mode": 33188
        }
    ]
}
```

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4,6 +4,7 @@ import faulthandler
 import functools
 import hashlib
 import inspect
+import json
 import logging
 import os
 import re
@@ -22,6 +23,8 @@ from itertools import zip_longest
 from .logger import create_logger, setup_logging
 logger = create_logger()
 
+import msgpack
+
 from . import __version__
 from . import helpers
 from .archive import Archive, ArchiveChecker, ArchiveRecreater, Statistics, is_special
@@ -34,11 +37,12 @@ from .helpers import Error, NoManifestError
 from .helpers import location_validator, archivename_validator, ChunkerParams, CompressionSpec
 from .helpers import PrefixSpec, SortBySpec, HUMAN_SORT_KEYS
 from .helpers import BaseFormatter, ItemFormatter, ArchiveFormatter, format_time, format_file_size, format_archive
-from .helpers import safe_encode, remove_surrogates, bin_to_hex
+from .helpers import safe_encode, remove_surrogates, bin_to_hex, prepare_dump_dict
 from .helpers import prune_within, prune_split
 from .helpers import to_localtime, timestamp
 from .helpers import get_cache_dir
 from .helpers import Manifest
+from .helpers import StableDict
 from .helpers import update_excludes, check_extension_modules
 from .helpers import dir_is_tagged, is_slow_msgpack, yes, sysinfo
 from .helpers import log_multi
@@ -1224,6 +1228,74 @@ class Archiver:
             with open(filename, 'wb') as fd:
                 fd.write(data)
         print('Done.')
+        return EXIT_SUCCESS
+
+    @with_repository()
+    def do_debug_dump_archive(self, args, repository, manifest, key):
+        """dump decoded archive metadata (not: data)"""
+
+        try:
+            archive_meta_orig = manifest.archives.get_raw_dict()[safe_encode(args.location.archive)]
+        except KeyError:
+            raise Archive.DoesNotExist(args.location.archive)
+
+        indent = 4
+
+        def do_indent(d):
+            return textwrap.indent(json.dumps(d, indent=indent), prefix=' ' * indent)
+
+        def output(fd):
+            # this outputs megabytes of data for a modest sized archive, so some manual streaming json output
+            fd.write('{\n')
+            fd.write('    "_name": ' + json.dumps(args.location.archive) + ",\n")
+            fd.write('    "_manifest_entry":\n')
+            fd.write(do_indent(prepare_dump_dict(archive_meta_orig)))
+            fd.write(',\n')
+
+            _, data = key.decrypt(archive_meta_orig[b'id'], repository.get(archive_meta_orig[b'id']))
+            archive_org_dict = msgpack.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape')
+
+            fd.write('    "_meta":\n')
+            fd.write(do_indent(prepare_dump_dict(archive_org_dict)))
+            fd.write(',\n')
+            fd.write('    "_items": [\n')
+
+            unpacker = msgpack.Unpacker(use_list=False, object_hook=StableDict)
+            first = True
+            for item_id in archive_org_dict[b'items']:
+                _, data = key.decrypt(item_id, repository.get(item_id))
+                unpacker.feed(data)
+                for item in unpacker:
+                    item = prepare_dump_dict(item)
+                    if first:
+                        first = False
+                    else:
+                        fd.write(',\n')
+                    fd.write(do_indent(item))
+
+            fd.write('\n')
+            fd.write('    ]\n}\n')
+
+        if args.path == '-':
+            output(sys.stdout)
+        else:
+            with open(args.path, 'w') as fd:
+                output(fd)
+        return EXIT_SUCCESS
+
+    @with_repository()
+    def do_debug_dump_manifest(self, args, repository, manifest, key):
+        """dump decoded repository manifest"""
+
+        _, data = key.decrypt(None, repository.get(manifest.MANIFEST_ID))
+
+        meta = prepare_dump_dict(msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
+
+        if args.path == '-':
+            json.dump(meta, sys.stdout, indent=4)
+        else:
+            with open(args.path, 'w') as fd:
+                json.dump(meta, fd, indent=4)
         return EXIT_SUCCESS
 
     @with_repository()
@@ -2715,6 +2787,36 @@ class Archiver:
         subparser.add_argument('location', metavar='ARCHIVE',
                                type=location_validator(archive=True),
                                help='archive to dump')
+
+        debug_dump_archive_epilog = textwrap.dedent("""
+        This command dumps all metadata of an archive in a decoded form to a file.
+        """)
+        subparser = debug_parsers.add_parser('dump-archive', parents=[common_parser], add_help=False,
+                                          description=self.do_debug_dump_archive.__doc__,
+                                          epilog=debug_dump_archive_epilog,
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='dump decoded archive metadata (debug)')
+        subparser.set_defaults(func=self.do_debug_dump_archive)
+        subparser.add_argument('location', metavar='ARCHIVE',
+                               type=location_validator(archive=True),
+                               help='archive to dump')
+        subparser.add_argument('path', metavar='PATH', type=str,
+                               help='file to dump data into')
+
+        debug_dump_manifest_epilog = textwrap.dedent("""
+        This command dumps manifest metadata of a repository in a decoded form to a file.
+        """)
+        subparser = debug_parsers.add_parser('dump-manifest', parents=[common_parser], add_help=False,
+                                          description=self.do_debug_dump_manifest.__doc__,
+                                          epilog=debug_dump_manifest_epilog,
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='dump decoded repository metadata (debug)')
+        subparser.set_defaults(func=self.do_debug_dump_manifest)
+        subparser.add_argument('location', metavar='REPOSITORY',
+                               type=location_validator(archive=False),
+                               help='repository to dump')
+        subparser.add_argument('path', metavar='PATH', type=str,
+                               help='file to dump data into')
 
         debug_dump_repo_objs_epilog = textwrap.dedent("""
         This command dumps raw (but decrypted and decompressed) repo objects to files.
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1,5 +1,6 @@
 import argparse
 import contextlib
+import collections
 import grp
 import hashlib
 import logging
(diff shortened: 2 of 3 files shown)


ThomasWaldmann:
is order of 1st / 2nd level dict keys always the same there?

textshell:
I believe the order will be the same for most objects because the code uses StableDict and OrderedDict. The last one esp. because i wanted the _items to always be last, so a quick inspection with a text viewer shows everything else on the first page.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+    @with_repository()
+    def do_debug_dump_archive(self, args, repository, manifest, key):
+        """dump decoded archive metadata (not: data)"""
+
+        try:
+            archive_meta_orig = manifest.archives.get_raw_dict()[safe_encode(args.location.archive)]
+        except KeyError:
+            raise Archive.DoesNotExist(args.location.archive)
+        archive_meta = self.debug_json(archive_meta_orig)
+        archive_meta['_name'] = args.location.archive
```
ThomasWaldmann: somehow this is making up entries (that are not there in the input data) to document caller parameters / context.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+
+        _, data = key.decrypt(archive_meta_orig[b'id'], repository.get(archive_meta_orig[b'id']))
+        archive_org_dict = msgpack.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape')
+        archive_meta['_meta'] = self.debug_json(archive_org_dict)
```
ThomasWaldmann: same here - should this be a separate command to get the items?

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+        for item_id in archive_org_dict[b'items']:
+            _, data = key.decrypt(item_id, repository.get(item_id))
+            unpacker.feed(data)
+            for item in unpacker:
+                item = self.debug_json(item)
+                items.append(item)
+
+        archive_meta['_items'] = items
+
+        with open(args.path, 'w') as fd:
```
ThomasWaldmann: instead of explicitly giving /dev/stdout, this should be the default if not given.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+    def do_debug_dump_manifest(self, args, repository, manifest, key):
+        """dump decoded repository manifest"""
+
+        _, data = key.decrypt(None, repository.get(manifest.MANIFEST_ID))
+
+        meta = self.debug_json(msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
+
+        with open(args.path, 'w') as fd:
+            json.dump(meta, fd, indent=4)
+        return EXIT_SUCCESS
```
ThomasWaldmann: much more straightforward than above.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+                if isinstance(value, dict):
+                    value = decode(value)
+                elif isinstance(value, tuple) or isinstance(value, list):
+                    value = decode_tuple(value)
+                elif isinstance(value, bytes):
+                    value = decode_bytes(value)
+                if isinstance(key, bytes):
+                    res[key.decode()] = value
+                else:
+                    res[key] = value
```
```
                if isinstance(key, bytes):
                    key = key.decode()
                res[key] = value
```

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+            return res
+
+        return decode(d)
+
+    @with_repository()
+    def do_debug_dump_archive(self, args, repository, manifest, key):
+        """dump decoded archive metadata (not: data)"""
+
+        try:
+            archive_meta_orig = manifest.archives.get_raw_dict()[safe_encode(args.location.archive)]
```
ThomasWaldmann: why use the raw dict?

src/borg/archiver.py
textshell: Not sure if that is a good default for outputting megabytes of data (for the archive use case).

src/borg/archiver.py
textshell: This is supposed to output the content of the data with as little alteration as needed to output as json. So i skipped as much higher level processing in borg as easily possible. Maybe i should do what i do in the manifest command and even load the manifest manually here. But that might not really be needed.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,91 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+        archive_meta['_meta'] = self.debug_json(archive_org_dict)
+
+        unpacker = msgpack.Unpacker(use_list=False, object_hook=StableDict)
+        items = []
+        for item_id in archive_org_dict[b'items']:
+            _, data = key.decrypt(item_id, repository.get(item_id))
+            unpacker.feed(data)
+            for item in unpacker:
+                item = self.debug_json(item)
+                items.append(item)
```
enkore: A principal issue is that this can get really large, storing it all in memory might be problematic. Streaming the output (not possible while maintaining the same form, at least with stdlib `json`, but very possible using "JSONlines") would avoid this problem.

src/borg/archiver.py
textshell: I would prefer not to have many commands. But maybe there is a usecase for just outputting the archive item. The synthetic top level object is a bit quirky. But i don‘t see it as much of a problem. I could imaging a --without-items to be useful though.

src/borg/archiver.py
ThomasWaldmann: if you're dumping lots of stuff, you have to expect lots of stuff. :)
many users might simply not know /dev/stdout. also not sure it is even there on all platforms.
could be also "-" if we do not like it as default.

src/borg/archiver.py
textshell: Yes, it tends to be memory hungry. I actually forgot to include the stats.
I tried this with a 77.56 GB archive. It seemed to peak with 4G memory usage and produced a 594Mb file in the end.
JSONlines would be fairly easy to produce, but i think this is not as widely supported by tools and much harder to read by humans.
I think without 100% purity doing streaming output with python json lib is actually feasible, if we think that it is worth it.

src/borg/archiver.py
textshell: Yes, explicit support for - would be more expected and less likely to leave users with ssh sessions stuck with a megabyte archive dump. I’ll add that.

src/borg/archiver.py
enkore: Right, if we edit the emitted "header" part a bit we can just emit each item separately and emit some finishing braces manually. Should work fine.

src/borg/archiver.py
textshell: fixed

src/borg/archiver.py
ThomasWaldmann: ... or just strip the outer braces, both.

textshell:
That is some ugly code for manual json output. But all repo supplied data is serialized by an actual json serializer and peak memory usage with the same test archive is down to below 37Mb. Maybe that‘s worth is.

enkore:
LGTM, but add a basic test case or two to at least see it doesn't crash? ;)

src/borg/archiver.py
```diff
@@ -1307,8 +1309,11 @@ def do_debug_dump_manifest(self, args, repository, manifest, key):
 
         meta = self.debug_json(msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
 
-        with open(args.path, 'w') as fd:
-            json.dump(meta, fd, indent=4)
+        if args.path == '-':
+            json.dump(meta, sys.stdout, indent=4)
+        else:
+            with open(args.path, 'w') as fd:
```
ThomasWaldmann: maybe we could (later?) refactor this to some dash_open(path, mode) context manager that handles "opening" stdin/out and fs files (and also closing, if needed) to avoid this duplication (likely also found at some other places).

src/borg/archiver.py
```diff
@@ -1222,6 +1226,116 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+                elif isinstance(value, tuple) or isinstance(value, list):
+                    value = decode_tuple(value)
+                elif isinstance(value, bytes):
+                    value = decode_bytes(value)
+                if isinstance(key, bytes):
+                    key = key.decode()
+                res[key] = value
+            return res
+
+        return decode(d)
```
ThomasWaldmann: the above method does not use "self", which is a good indication it should not be a method in the first place.

move it to helpers module?

textshell: I didn‘t want to bloat the helpers module even more with quite specific functions. decode_bytes is not really general purpose and is referenced by this code.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,116 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
...
+                    value = decode_bytes(value)
+                res.append(value)
+            return res
+
+        def decode(d):
+            res = collections.OrderedDict()
+            for key, value in d.items():
+                if isinstance(value, dict):
+                    value = decode(value)
+                elif isinstance(value, tuple) or isinstance(value, list):
```
ThomasWaldmann: isinstance(value, (tuple, list))

src/borg/archiver.py
textshell: Yes, that would be useful in many places. But i wonder if that might also contains decent error reporting etc. That would be quite nice for many simple commands.

src/borg/archiver.py
```diff
@@ -1222,6 +1226,116 @@ def do_debug_dump_archive_items(self, args, repository, manifest, key):
         print('Done.')
         return EXIT_SUCCESS
 
+    def debug_json(self, d):
```
ThomasWaldmann: the code in here is not directly related to json, you just use its result for json generation.

so, just call it prettify_dict()?

src/borg/archiver.py
ThomasWaldmann: could also get some testing then.

src/borg/archiver.py
```diff
@@ -2684,6 +2798,36 @@ def build_parser(self, prog=None):
...
+                                          description=self.do_debug_dump_archive.__doc__,
+                                          epilog=debug_dump_archive_epilog,
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='dump decoded archive metadata (debug)')
+        subparser.set_defaults(func=self.do_debug_dump_archive)
+        subparser.add_argument('location', metavar='ARCHIVE',
+                               type=location_validator(archive=True),
+                               help='archive to dump')
+        subparser.add_argument('path', metavar='PATH', type=str,
+                               help='file to dumped data into')
```
ThomasWaldmann: dump

src/borg/archiver.py
```diff
@@ -2684,6 +2798,36 @@ def build_parser(self, prog=None):
...
+                                          description=self.do_debug_dump_manifest.__doc__,
+                                          epilog=debug_dump_manifest_epilog,
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='dump decoded repository metadata (debug)')
+        subparser.set_defaults(func=self.do_debug_dump_manifest)
+        subparser.add_argument('location', metavar='REPOSITORY',
+                               type=location_validator(archive=False),
+                               help='repository to dump')
+        subparser.add_argument('path', metavar='PATH', type=str,
+                               help='file to dumped data into')
```
ThomasWaldmann: dump

src/borg/archiver.py
textshell: yes

src/borg/archiver.py
textshell: fixed both

src/borg/archiver.py
ThomasWaldmann: bloating archiver.py with utility functions is not preferable IMHO.

textshell:
added some minimal testing (basically that it does not crash, outputs valid json and has the expected top level entries in the dict)
I think i fixed all the nitpicks by thomas also.

textshell:
squashed, anything else to do here?

PR #2115, opened 2017-01-29
Manifest: Make sure manifest timestamp is strictly monotonically increasing.

textshell:
Computer clocks are often not set very accurately set, but borg
assumes manifest timestamps are never going back in time.

Ensure that this is actually the case.

Commands to reproduce:
```
borg init -e repokey /tmp/borg-repo2
borg create /tmp/borg-repo2::test1 tox.ini
borg info /tmp/borg-repo2::test1
faketime -f "+1y" borg create /tmp/borg-repo2::test2 tox.ini
BORG_CACHE_DIR=/tmp/foo BORG_SECURITY_DIR=/tmp/bar borg create /tmp/borg-repo2::test3 tox.ini
borg info /tmp/borg-repo2::test1
```

### DIFF
--- a/borg/helpers.py
+++ b/borg/helpers.py
@@ -110,6 +110,7 @@ class Manifest:
         self.repository = repository
         self.item_keys = frozenset(item_keys) if item_keys is not None else ITEM_KEYS
         self.tam_verified = False
+        self.timestamp = None
 
     @classmethod
     def load(cls, repository, key=None, force_tam_not_required=False):
@@ -151,7 +152,13 @@ class Manifest:
     def write(self):
         if self.key.tam_required:
             self.config[b'tam_required'] = True
-        self.timestamp = datetime.utcnow().isoformat()
+        # self.timestamp needs to be strictly monotonically increasing. Clocks often are not set correctly
+        if self.timestamp is None:
+            self.timestamp = datetime.utcnow().isoformat()
+        else:
+            prev_ts = datetime.strptime(self.timestamp, "%Y-%m-%dT%H:%M:%S.%f")
+            incremented = (prev_ts + timedelta(microseconds=1)).isoformat()
+            self.timestamp = max(incremented, datetime.utcnow().isoformat())
         m = {
             'version': 1,
             'archives': StableDict((name, StableDict(archive)) for name, archive in self.archives.items()),


borg/helpers.py
```diff
@@ -151,7 +152,13 @@ def load(cls, repository, key=None, force_tam_not_required=False):
...
         if self.key.tam_required:
             self.config[b'tam_required'] = True
-        self.timestamp = datetime.utcnow().isoformat()
+        # self.timestamp needs to be strictly monotonically increasing. Clocks often are not set correctly
+        if self.timestamp is None:
+            self.timestamp = datetime.utcnow().isoformat()
+        else:
+            prev_ts = datetime.strptime(self.timestamp, "%Y-%m-%dT%H:%M:%S.%f")
+            incremented = (prev_ts + timedelta(microseconds=1)).isoformat()
+            self.timestamp = max(incremented, datetime.utcnow().isoformat())
```
ThomasWaldmann: how about doing max() on the datetime object and calling isoformat on the result?
think that would be a bit cleaner, although max-of-strings also works.

textshell: I’m happy with both variants. But all other code pathes seem to use the string, so sticking to that seemed less confusing.

enkore: Does it need to actually increase (by 1 µs)? IIRC the check in cache goes like `our_timestamp > manifest_ts: raise PanicUser()`; L159 could go away and L160, too, leaving only `self.timestamp = max(self.timestamp, datetime...isoformat())`

textshell: To actually provide rollback prevention it needs to be higher than the previous one. To just silence the warning it might not.
I somehow thought that the cache also uses the manifest timestamp, but it does use the id which is partly derived from the timestamp (the id beeing the id_hash of the whole manifest). I feel a lot safer if the manifest has a component that will never be the same twice. And we plan to move the TAM out into some kind of envelop so we should‘t leave a footgun laying around depending on the TAM. 

That said, not having to parse the date would be nice. But i don‘t trust that construction.

enkore: > To actually provide rollback prevention it needs to be higher than the previous one. To just silence the warning it might not.

Ok, then I think we should revisit the exact cache condition (perhaps not in this PR - create issue?)

textshell: Not sure what you would like to discuss about the cache resync condition. So if you want to discuss that it would be best if you can open the ticket. Also rollback protection still is an issue without the cache things.

enkore: You say that for rollback prevention the timestamp must actually increase, and I agree, but the cache does not check that:

```
if self.manifest.id != self.manifest_id:
    if self.timestamp and self.timestamp > manifest.timestamp:
```

textshell: Sorry for jumping all around.
Rollback prevention is via
Cache.__init__ -> self.security_manager.assert_secure -> SecurityManager.assert_no_manifest_replay
After that has executed the next line in __init__ is the cache sync detection. It uses self.manifest.id != self.manifest_id which now only has to ensure that borg always rebuilds the cache if anything could possibly changed the repository state (including check, recreate, etc)

ThomasWaldmann: a bit unpretty: once there is a wrong far-future ts in the manifest, it will never ever get correct again, because it will always choose to increment by 1us.

textshell: Yes, that is indeed the case. But the current behavior is worse.
The only secure way i see is to introduce an epoch that is incremented instead of the timestamp or maybe prepended if the time is too much in the future. But of course the device with the bad clock can also be the device doing the current operation so the epoch might get quite high.

I actually thought the rollback protection would use just an number that is incremented but when looking at some other code i was reminded that it is a date. Technically there is no need to have a correct date. I didn‘t find any place where it is actually output or used as a timestamp, so i don‘t think we need to invest too much energy into keeping it the same as the current time.

enkore: > a bit unpretty: once there is a wrong far-future ts in the manifest, it will never ever get correct again, because it will always choose to increment by 1us.

Only cosmetic in the security DB I think, no actual impact.

Merge?

textshell: As Thomas already clicked on the "approve" button i think you can merge this. Or should i merge it?

enkore: Hit it! :)

PR #2116, opened 2017-01-30
fix wrong skip_hint on hashindex_set when encountering tombstones

rciorba:
hashindex_lookup would always hint at skipping whatever it's probe
length had been with no regard for tombstones it had encountered. This
meant new keys would not overwrite first tombstones, but would always
land on empty buckets. 

Ran the benchmarks I used for the Robin Hood hashing so you can 
compare the impact of this patch:
https://cdn.rawgit.com/rciorba/d8990a54be8b778bb3322ac40364de90/raw/90d05773cc377b061f400d3833166bbce4243ba1/results.html

That is a regression I introduced myself as part of: #1748

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -109,19 +109,15 @@ hashindex_index(HashIndex *index, const void *key)
 }
 
 static int
-hashindex_lookup(HashIndex *index, const void *key, int *skip_hint)
+hashindex_lookup(HashIndex *index, const void *key, int *start_idx)
 {
     int didx = -1;
     int start = hashindex_index(index, key);
     int idx = start;
-    int offset;
-    for(offset=0;;offset++) {
-        if (skip_hint != NULL) {
-            (*skip_hint) = offset;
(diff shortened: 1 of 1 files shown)


src/borg/_hashindex.c
```diff
@@ -383,18 +383,18 @@ hashindex_get(HashIndex *index, const void *key)
 static int
 hashindex_set(HashIndex *index, const void *key, const void *value)
 {
-    int offset = 0;
-    int idx = hashindex_lookup(index, key, &offset);
+    int start_addr;
```
ThomasWaldmann: use start_hint? or start_idx?
*_addr sounds rather like a memory address / pointer (see BUCKET_ADDR()).

rciorba: Good point! `start_idx` it is.

ThomasWaldmann:
@enkore can you have a look also?

ThomasWaldmann:
@rciorba thanks!

PR #2120, opened 2017-02-03
docs: use 'restrict' for authorized_keys

ThomasWaldmann:
'restrict' enables all restriction (even future ones), so it is safest.

Thanks to Michael Reed for the patch.


textshell:
`restrict` is nice but it‘s only supported since openssl 7.1p2. So i’m not sure we should remove the way that actually works with older system quite yet. On the other hand we like our docs to be short and on the point.

ThomasWaldmann:
hmm, right. that is not even in debian jessie nor in ubuntu trusty yet. ubuntu xenial has it.

OK, so i guess we rather make a "later" tagged ticket out of this.

PR #2124, opened 2017-02-04
Move migrate-to-repokey, change-passphrase to key group

enkore:


src/borg/archiver.py
```diff
@@ -1886,9 +1878,32 @@ def build_parser(self, prog=None):
                                default=False,
                                help='interactively import from a backup done with --paper')
 
+        change_passphrase_epilog = textwrap.dedent("""
+        The key files used for repository encryption are optionally passphrase
+        protected. This command can be used to change this passphrase.
+        """)
+        subparser = key_parsers.add_parser('change-passphrase', parents=[common_parser], add_help=False,
```
ThomasWaldmann: is parents= correct?

enkore: Yes, `parents` is unrelated to the actual hierarchy of the parsers. The parser inherits options from the parent, while being placed in the hierarchy through `add_parser`.

ThomasWaldmann: ok, was just wondering.

PR #2127, opened 2017-02-05
Man pages

### LINKED ISSUE #208: make a better / real man page

enkore:
Fixes #208 

This isn't the pinnacle of perfection yet (and I don't believe in first-time perfection anyway ;), some things to improve:

- There are no overview pages yet, like `borg(1)`.
- ~~No examples in the pages (but they aren't in --help either, maybe kill two birds with one stone here)~~
- ~~SEE ALSO isn't fleshed out yet, only a stub (borg-common(1)) + the infrastructure to do them. borg-create(1) has some.~~

However, in general they look quite good and read well.

Code wise there is some duplication with the build_usage code, but it's more or less different all over the place so deduping that would make it harder to understand or inflate the code more.


enkore:
Since I don't know a way to display fancy HTML from man pages on or offline, here are two Konsole-grabs in lieu of that: https://pstore.enkore.de/borgman.html

enkore:
A page with more complex layout http://pstore.enkore.de/borgman2.html

setup.py
```diff
@@ -326,6 +329,222 @@ def shipout(text):
...
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        try:
+            import docutils
+        except ImportError:
+            print('Cannot generate man pages -- docutils is not installed.', file=sys.stderr)
```
ThomasWaldmann: this is dangerous, it hides information and might tell wrong conclusions at some times, e.g. if there is an ImportError inside docutils.

i'ld prefer not catching this.

setup.py
```diff
@@ -326,6 +329,222 @@ def shipout(text):
...
+            return
+
+        for command, parser in sorted(choices.items()):
+            if command.startswith('debug') or command == 'help':
+                continue
+
+            man_title = 'borg-' + command.replace(' ', '-')
+            print('building man page', man_title + '(1)', file=sys.stderr)
+
+            if self.generate_level(command + " ", parser, Archiver):
```
ThomasWaldmann: always use single quotes for (short) string literals?

setup.py
```diff
@@ -326,6 +329,222 @@ def shipout(text):
...
+    def write_examples(self, write, command):
+        with open('docs/usage.rst') as fd:
+            usage = fd.read()
+            usage_include = '.. include:: usage/%s.rst.inc' % command
+            begin = usage.find(usage_include)
+            end = usage.find('.. include', begin + 1)
+            examples = usage[begin:end]
+            examples = examples.replace(usage_include, '')
+            examples = examples.replace('Examples\n~~~~~~~~', '')
+            examples = examples.replace('Miscellaneous Help\n------------------', '')
```
ThomasWaldmann: there could be a underlined(text, c) function that returns 'text\ncccc\n'.

setup.py
```diff
@@ -326,6 +329,222 @@ def shipout(text):
...
+        write(', '.join(see_also))
+
+    def gen_man_page(self, name, rst):
+        from docutils.writers import manpage
+        from docutils.core import publish_string
+        man_page = publish_string(source=rst, writer=manpage.Writer())
+        with open('docs/man/%s.1' % name, 'wb') as fd:
+            fd.write(man_page)
+
+    def write_usage(self, parser, fp):
```
ThomasWaldmann: use fd, as elsewhere?

src/borg/archiver.py
```diff
@@ -2024,16 +2023,17 @@ def build_parser(self, prog=None):
...
         archive_group.add_argument('-c', '--checkpoint-interval', dest='checkpoint_interval',
                                    type=int, default=1800, metavar='SECONDS',
                                    help='write checkpoint every SECONDS seconds (Default: 1800)')
         archive_group.add_argument('--chunker-params', dest='chunker_params',
                                    type=ChunkerParams, default=CHUNKER_PARAMS,
-                                   metavar='CHUNK_MIN_EXP,CHUNK_MAX_EXP,HASH_MASK_BITS,HASH_WINDOW_SIZE',
-                                   help='specify the chunker parameters. default: %d,%d,%d,%d' % CHUNKER_PARAMS)
+                                   metavar='PARAMS',
+                                   help='specify the chunker parameters (CHUNK_MIN_EXP, CHUNK_MAX_EXP, '
+                                        'HASH_MASK_BITS, HASH_WINDOW_SIZE). default: %d,%d,%d,%d' % CHUNKER_PARAMS)
```
ThomasWaldmann: add a \n in here, this is longer than 80 chars.

ThomasWaldmann: also check other places for len > 80.

enkore: we use 120 chars line length

setup.py
enkore: Since this removes verbatim text from the usage.rst I'd rather keep it obvious/verbatim.

src/borg/archiver.py
ThomasWaldmann: i didn't mean the src formatting here, i meant the console output.

enkore: I believe parameter help is wrapped by argparse

Issue #1806, opened 2016-11-05
Minimal mandatory feature support for manifest


textshell:
Quite some features require changes in the repository that are not safe when using old client versions. We should have minimal support in place to mark a repository on first use of such an feature as incompatible with older versions. As features can also be "unused" this should be with feature granularity. Also different operations have different requirements. For the start i propose to have mandatory feature lists for "write", "delete" and "check". (E.g. a feature where archives are stored outside of the manifest would be mandatory for delete (and check, but this could be implied)

For this minimal version we should define how these are saved in the manifest in a way that all borg versions will preserve when rewriting the manifest.

Additionally borg should check that all relevant lists are empty before performing an operation.

We need to decide on a format of the features. I propose using short strings.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/38975114-minimal-mandatory-feature-support-for-manifest)

ThomasWaldmann:
Do we really need that?

We could just increment Manifest version to avoid older clients working with the repository.

The RPC mechanism already has some versioning.

textshell:
Yes, i think we need this ;).
Manifest version is a very blunt tool that needs either a lot of planing and predicting the future or doesn’t allow for a graceful upgrading period when combined with optional features.

This is much more fine grained. For example adding a new mandatory feature for check still allows older clients to use the repository. But we are free to add features that would be unsafe when running check on a version that doesn‘t understand new feature. Or maybe it is save for an old version to read the repository but it is not safe to do any changes (i.e. because new parts need to be refcounted).

This also allows to capture certain upgrade steps. I.e. a new archive feature A might start out as optional but we someday want to be able to remove support for archives that lack that feature. The code that adds this feature knows a "a-only" feature for write and if it sees that it unconditionally enables that feature.
So later we add an upgrade step where all archives not yet using this new feature are rewritten and adds the "a-only" mandatory feature to the manifest for write. Now version that don‘t support the feature refuse to create archives. Intermediate version just keep working and new version can drop the old support code.

I think this mechanism will allow smooth upgrading in more scenarios, but i‘m not very creative at the moment. Also this is the standard used in many systems. While we don‘t want to keep compatibility like certain file systems do, we can just more aggressively promote features to mandatory to keep compat code small, but still allow users graceful upgrade pathes with mixed versions.

ThomasWaldmann:
fixed by #2648.

@textshell claim the bounty.

Issue #1681, opened 2016-10-03
extend item data structure to better deal with large xattr-like things?


enkore:
Round-up of #51 #1337 #1342 

- xattrs (Linux, OSX, BSD) are currently stored in-line in the items, as a dictionary.
- NTFS ADS are more file-like and are currently separate items altogether, using the Windows notation of `file:stream`.
- SunOS fsattrs are not supported (but, as a quick summary: they essentially attach a completely independent FS namespace to *every* file).

On Linux xattrs are small by Kernel limitation. On OSX resource forks can be as big as normal files. Ditto for ADS.

There have been a couple ideas how to manage this better:

1. Add "subitems" to items; a chunk ID list for items that are attached to this item, as 

  - streams / resource forks
  - fsattrs

  This approach allows to encode even SunOS extremely advanced structure, and the added indirection avoids inflating the length of the items list unnecessarily.

2. Use in-band signalling in paths. Because no OS accepts NUL in paths, this is the obvious choice for the ISB (and could be thus argued to be out-of-band, but in-stream, in fact).

  This is pretty much the same thing as 1.), with the advantage that older Borg correctly syncs to this and finds all chunks, while having the disadvantage that these items would be in the main items list, enlarging the length of it.

Either way allows for the semantics required and enables support for large xattrs/fsattrs/ADS. They also allow the proper case selection when extracting.

ThomasWaldmann:
@enkore As this is a rather fundamental change, I don't think this fits into the 1.1 (final) milestone.

It should be either in for rc1 or get pushed to 1.2 or 1.3.

infectormp:
just for info, kernel 4.13 get large xattr support https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=e50e5129f384ae282adebfb561189cdb19b81cee

PR #2134, opened 2017-02-06
Add minimal version of in repository mandatory feature flags.

textshell:
This should allow us to make sure older borg versions can be cleanly
prevented from doing operations that are no longer safe because of
repository format evolution. This allows more fine grained control than
just incrementing the manifest version. So for example a change that
still allows new archives to be created but would corrupt the repository
when an old version tries to delete an archive or check the repository
would add the new feature to the check and delete set but leave it out
of the write set.

This is somewhat inspired by ext{2,3,4} which uses sets for
compat (everything except fsck), ro-compat (may only be accessed
read-only by older versions) and features (refuse all access).

This implements #1806

Open questions:

- [x] What granularity is right? I currently use "write" for create, "delete" for delete and prune and "check" for everything else.
- [x] Changes that make delete unsafe likely have consequences for the cache. Should this be handled or should we just wait until the first feature that need this and think about it then?
- [x] Should this also add support for manifest version 2 (just manifest version 1 plus respects the mandatory feature flags), so we have a way to lockout 1.0.x borg versions for changes where this patch already makes it safe with borg 1.1.x?

### DIFF
--- a/borg/archiver.py
+++ b/borg/archiver.py
@@ -56,7 +56,8 @@ def argument(args, str_or_bool):
     return str_or_bool
 
 
-def with_repository(fake=False, invert_fake=False, create=False, lock=True, exclusive=False, manifest=True, cache=False):
+def with_repository(fake=False, invert_fake=False, create=False, lock=True,
+                    exclusive=False, manifest=True, cache=False, compatibility=None):
     """
     Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, …)
 
@@ -67,7 +68,20 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True, excl
     :param exclusive: (str or bool) lock repository exclusively (for writing)
     :param manifest: load manifest and key, pass them as keyword arguments
     :param cache: open cache, pass it as keyword argument (implies manifest)
+    :param compatibility: mandatory if not create and (manifest or cache), specifies mandatory feature categories to check
     """
+
+    if not create and (manifest or cache):
+        if compatibility is None:
+            raise AssertionError("with_repository decorator used without compatibility argument")
+        if type(compatibility) is not tuple:
+            raise AssertionError("with_repository decorator compatibility argument must be of type tuple")
+    else:
+        if compatibility is not None:
+            raise AssertionError("with_repository called with compatibility argument but would not check" + repr(compatibility))
+        if create:
+            compatibility = Manifest.NO_OPERATION_CHECK
+
     def decorator(method):
         @functools.wraps(method)
         def wrapper(self, args, **kwargs):
@@ -84,7 +98,7 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True, excl
                                         append_only=append_only)
             with repository:
                 if manifest or cache:
-                    kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
+                    kwargs['manifest'], kwargs['key'] = Manifest.load(repository, compatibility)
                 if cache:
                     with Cache(repository, kwargs['key'], kwargs['manifest'],
                                do_files=getattr(args, 'cache_files', False), lock_wait=self.lock_wait) as cache_:
@@ -176,7 +190,7 @@ class Archiver:
             return EXIT_WARNING
         return EXIT_SUCCESS
 
-    @with_repository()
+    @with_repository(compatibility=(Manifest.Operation.CHECK,))
     def do_change_passphrase(self, args, repository, manifest, key):
         """Change repository key file passphrase"""
         if not hasattr(key, 'change_passphrase'):
@@ -241,7 +255,7 @@ class Archiver:
         logger.info('Key updated')
         return EXIT_SUCCESS
 
-    @with_repository(fake='dry_run', exclusive=True)
+    @with_repository(fake='dry_run', exclusive=True, compatibility=(Manifest.Operation.WRITE,))
     def do_create(self, args, repository, manifest=None, key=None):
         """Create new archive"""
         matcher = PatternMatcher(fallback=True)
@@ -424,7 +438,7 @@ class Archiver:
                 return matcher.match(item[b'path'])
         return item_filter
 
-    @with_repository()
+    @with_repository(compatibility=(Manifest.Operation.READ,))
     @with_archive
     def do_extract(self, args, repository, manifest, key, archive):
         """Extract archive contents"""
@@ -490,7 +504,7 @@ class Archiver:
                 self.print_warning("Include pattern '%s' never matched.", pattern)
         return self.exit_code
 
-    @with_repository(exclusive=True, cache=True)
+    @with_repository(exclusive=True, cache=True, compatibility=(Manifest.Operation.CHECK,))
     @with_archive
     def do_rename(self, args, repository, manifest, key, cache, archive):
         """Rename an existing archive"""
@@ -506,7 +520,7 @@ class Archiver:
         """Delete an existing repository or archive"""
         if args.location.archive:
             archive_name = args.location.archive
-            manifest, key = Manifest.load(repository)
+            manifest, key = Manifest.load(repository, (Manifest.Operation.DELETE,))
 
             if args.forced == 2:
                 try:
@@ -537,7 +551,7 @@ class Archiver:
             if not args.cache_only:
                 msg = []
                 try:
-                    manifest, key = Manifest.load(repository)
+                    manifest, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK)
                 except NoManifestError:
                     msg.append("You requested to completely DELETE the repository *including* all archives it may contain.")
                     msg.append("This repository seems to have no manifest, so we can't tell anything about its contents.")
@@ -573,7 +587,7 @@ class Archiver:
 
         return self._do_mount(args)
 
-    @with_repository()
+    @with_repository(compatibility=(Manifest.Operation.READ,))
     def _do_mount(self, args, repository, manifest, key):
         from .fuse import FuseOperations
 
@@ -595,7 +609,7 @@ class Archiver:
         """un-mount the FUSE filesystem"""
         return umount(args.mountpoint)
 
-    @with_repository()
+    @with_repository(compatibility=(Manifest.Operation.READ,))
     def do_list(self, args, repository, manifest, key):
         """List archive or repository contents"""
         if args.location.archive:
@@ -681,7 +695,7 @@ class Archiver:
                     print(format_archive(archive_info))
         return self.exit_code
 
-    @with_repository(cache=True)
+    @with_repository(cache=True, compatibility=(Manifest.Operation.READ,))
     @with_archive
     def do_info(self, args, repository, manifest, key, archive, cache):
         """Show archive details such as disk space used"""
@@ -699,7 +713,7 @@ class Archiver:
         print(str(cache))
         return self.exit_code
 
-    @with_repository(exclusive=True)
+    @with_repository(exclusive=True, compatibility=(Manifest.Operation.DELETE,))
     def do_prune(self, args, repository, manifest, key):
         """Prune repository archives according to specified rules"""
         if not any((args.hourly, args.daily,
@@ -759,7 +773,7 @@ class Archiver:
     def do_upgrade(self, args, repository, manifest=None, key=None):
         """upgrade a repository from a previous version"""
         if args.tam:
-            manifest, key = Manifest.load(repository, force_tam_not_required=args.force)
+            manifest, key = Manifest.load(repository, (Manifest.Operation.CHECK,), force_tam_not_required=args.force)
 
             if not hasattr(key, 'change_passphrase'):
                 print('This repository is not encrypted, cannot enable TAM.')
@@ -784,7 +798,7 @@ class Archiver:
                 open(tam_file, 'w').close()
                 print('Updated security database')
         elif args.disable_tam:
-            manifest, key = Manifest.load(repository, force_tam_not_required=True)
+            manifest, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK, force_tam_not_required=True)
             if tam_required(repository):
                 os.unlink(tam_required_file(repository))
             if key.tam_required:
@@ -817,7 +831,7 @@ class Archiver:
         print(sysinfo())
         return EXIT_SUCCESS
 
-    @with_repository()
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
     def do_debug_dump_archive_items(self, args, repository, manifest, key):
         """dump (decrypted, decompressed) archive items metadata (not: data)"""
         archive = Archive(repository, key, manifest, args.location.archive)
@@ -830,7 +844,7 @@ class Archiver:
         print('Done.')
         return EXIT_SUCCESS
 
-    @with_repository()
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
     def do_debug_dump_repo_objs(self, args, repository, manifest, key):
         """dump (decrypted, decompressed) repo objects"""
         marker = None
@@ -904,7 +918,7 @@ class Archiver:
         print('Done.')
         return EXIT_SUCCESS
 
-    @with_repository(manifest=False, exclusive=True, cache=True)
+    @with_repository(manifest=False, exclusive=True, cache=True, compatibility=Manifest.NO_OPERATION_CHECK)
     def do_debug_refcount_obj(self, args, repository, manifest, key, cache):
         """display refcounts for the objects with the given IDs"""
         for hex_id in args.ids:
--- a/borg/cache.py
+++ b/borg/cache.py
@@ -9,8 +9,8 @@ import shutil
 from .key import PlaintextKey
 from .logger import create_logger
 logger = create_logger()
-from .helpers import Error, get_cache_dir, decode_dict, int_to_bigint, \
-    bigint_to_int, format_file_size, yes, bin_to_hex, Location, safe_ns
+from .helpers import Error, Manifest, get_cache_dir, decode_dict, int_to_bigint, \
+    bigint_to_int, format_file_size, yes, bin_to_hex, Location, safe_ns, parse_stringified_list
 from .locking import Lock
 from .hashindex import ChunkIndex
 
@@ -84,6 +84,11 @@ class Cache:
                 self.begin_txn()
                 self.commit()
 
+            if not self.check_cache_compatibility():
+                self.wipe_cache()
+
+            self.update_compatibility()
+
             if sync and self.manifest.id != self.manifest_id:
                 # If repository is older than the cache something fishy is going on
                 if self.timestamp and self.timestamp > manifest.timestamp:
@@ -94,6 +99,7 @@ class Cache:
                 # Make sure an encrypted repository has not been swapped for an unencrypted repository
                 if self.key_type is not None and self.key_type != str(key.TYPE):
                     raise self.EncryptionMethodMismatch()
+
                 self.sync()
                 self.commit()
         except:
@@ -175,6 +181,8 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         self.timestamp = self.config.get('cache', 'timestamp', fallback=None)
         self.key_type = self.config.get('cache', 'key_type', fallback=None)
         self.previous_location = self.config.get('cache', 'previous_location', fallback=None)
+        self.ignored_features = set(parse_stringified_list(self.config.get('cache', 'ignored_features', fallback='')))
+        self.mandatory_features = set(parse_stringified_list(self.config.get('cache', 'mandatory_features', fallback='')))
         self.chunks = ChunkIndex.read(os.path.join(self.path, 'chunks').encode('utf-8'))
         self.files = None
 
@@ -240,6 +248,8 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         self.config.set('cache', 'timestamp', self.manifest.timestamp)
         self.config.set('cache', 'key_type', str(self.key.TYPE))
         self.config.set('cache', 'previous_location', self.repository._location.canonical_path())
+        self.config.set('cache', 'ignored_features', ','.join(self.ignored_features))
+        self.config.set('cache', 'mandatory_features', ','.join(self.mandatory_features))
         with open(os.path.join(self.path, 'config'), 'w') as fd:
             self.config.write(fd)
         self.chunks.write(os.path.join(self.path, 'chunks').encode('utf-8'))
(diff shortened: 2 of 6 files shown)


ThomasWaldmann:
granularity: as you said + read, for completeness.
delete / cache: no idea
manifest v2: not sure how this helps / how you'ld use it.

textshell:
ok, i'll add a read one too. We could add a second one for reading into the archives. Shouldn‘t be much work.
The basic idea about manifest v2 is, that all these mandatory feature flags won‘t stop borg 1.0.9 from possibly corrupting a repo where borg 1.1 would correctly refuse to do disallowed actions. So if we want to prevent older versions we need to change something they check once to make sure mandatory feature support is there. One easy way would be to use the manifest version. We can still do that later of course. I tend to accepting v2 as soon as this is in. (just changing `if not m.get(b'version') == 1:` to `if m.get(b'version') not in (1,2):` in Manifest.load)

textshell:
The cache problem is as follows. Say we want to add a feature that allows larger archives by some means where not all chunk id‘s containing items are stored in the archive chunk directly.
Such a repository would still be safe for adding archives to with an old version. So such a feature should only be mandatory for "delete" and "check". But of course older borg versions would not have the chunks in a large archive be refcounted correctly in the local cache. It i was wondering if we need to store some additional flags with the cache so borg versions can coordinate not to corrupt things. I never much thought about cache changes though.

borg/archiver.py
```diff
@@ -177,6 +178,7 @@ def do_check(self, args, repository):
     @with_repository()
     def do_change_passphrase(self, args, repository, manifest, key):
         """Change repository key file passphrase"""
+        check_repository_compatibility(manifest, REPO_OPERATION_CHECK)
```
ThomasWaldmann: having to add that everywhere to avoid damage seems error prone, it could be forgotten.
so, add it to the decorator maybe?

borg/helpers.py
```diff
@@ -98,6 +101,23 @@ def check_extension_modules():
...
+
+REPO_OPERATION_CHECK = "check"
+REPO_OPERATION_WRITE = "write"
+REPO_OPERATION_DELETE = "delete"
+
+def check_repository_compatibility(manifest, operation):
+    if manifest is None:
+        # various fake options can have manifest is None
+        return
+    if not operation in manifest.config:
```
ThomasWaldmann: "not in" is a operator.

borg/helpers.py
```diff
@@ -98,6 +101,23 @@ def check_extension_modules():
...
+        return
+    requirements = manifest.config[operation]
+    if 'mandatory' in requirements and not set(requirements['mandatory']) <= SUPPORTED_REPO_FEATURES:
+        raise MandatoryFeatureUnsupported(set(requirements['mandatory']) - SUPPORTED_REPO_FEATURES)
```
ThomasWaldmann: maybe always compute the unsupported_features set and raise if not empty? saves a bit of duplication maybe.

borg/archiver.py
textshell: I was unsure how to make that nice.
- There are places where the manifest is loaded only later. So it can not really be universal
- should this be additional in the exiting decorators or would it be stacked?
- I really want this to be in 1.1, so i do consider it ok if it needs to be optimized later.

Maybe we can just load the manifest more often centrally that would certainly make this easier as decorator?

ThomasWaldmann: true, the decorator is not always used. but it would cover at least the places where it is used.

additional in the existing decorator or just an additional call as you have now.

textshell: I made a version using a `compatibility` argument to `with_repository`
Do you like this version better? I made it very explicit about locations where there is no checking.

borg/archiver.py
```diff
@@ -54,7 +56,8 @@ def argument(args, str_or_bool):
     return str_or_bool
 
 
-def with_repository(fake=False, invert_fake=False, create=False, lock=True, exclusive=False, manifest=True, cache=False):
+def with_repository(fake=False, invert_fake=False, create=False, lock=True,
+                    exclusive=False, manifest=True, cache=False, compatibility=None):
```
ThomasWaldmann: compatibility is missing in the docstring below.

borg/archiver.py
```diff
@@ -69,6 +72,11 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True, excl
     def decorator(method):
         @functools.wraps(method)
         def wrapper(self, args, **kwargs):
+            nonlocal compatibility
+            if create:
+                compatibility = REPO_OPERATION_NO_CHECK
+            if compatibility is None:
+                raise Exception("with_repository called without compatibility argument")
```
ThomasWaldmann: the call to check_repository_compatibility() is still missing here (somewhere below), right?

textshell: Yes, i’ll fix that in the next revision.

enkore:
Thoughts

- What is READ actually for? Wouldn't we indicate this rather via version attributes on objects (manifest, archives)?
  - Some of these are not super helpful (`assert` form), but I'd rather fix these and give them better error messages
  - Perhaps to indicate a feature present in the repo that is required to be understood for a cache sync? In this case it would be complimentary to the versions, to make it fail fast, but version attributes would still provide the safety called for.
- WRITE -> CREATE?
  - Does WRITE/CREATE imply READ?
    - Right now it does, but one doesn't need a lot of imagination to see that change: eg. repository-provided bloom filter

borg/archiver.py
enkore: maybe put that into Manifest.load, not here?

borg/helpers.py
```diff
@@ -98,6 +102,28 @@ def check_extension_modules():
...
+    if operation == REPO_OPERATION_NO_CHECK:
+        return
+    if operation not in manifest.config:
+        return
+    requirements = manifest.config[operation]
+    if 'mandatory' in requirements and not set(requirements['mandatory']) <= SUPPORTED_REPO_FEATURES:
+        raise MandatoryFeatureUnsupported(set(requirements['mandatory']) - SUPPORTED_REPO_FEATURES)
+
+
 class Manifest:
```
enkore: v `def load(cls, repository, key=None, operation=None, force_tam_not_required=False):`

enkore:
If operation levels don't have a total ordering then it might make more sense to be able to verify legal operation against a set of operation levels than just one.

ThomasWaldmann:
if this is going to still take a while: shift it to b5?

textshell:
Sorry this got stuck a bit.
I now moved the compatibility checking into the manifest class. Still the majority of the checking is specified by the with_repository decorator, but special cases now use Manifest.load for further checking.

About READ: I originally left read out because i felt that likely something at archive level would be more helpful. So i’m happy with or without it here. I do like the idea to have a REFCOUNT_SYNC operation quite a lot. Maybe instead of READ?

WRITE currently only covers do_create, so maybe we should rename it. In a way WRITE does not really need perfect REFCOUNT_SYNC, it just needs a conservative refcount that doesn‘t have false positives.

Checking more than one might be needed in some situations, esp. if we go with REFCOUNT_SYNC and for now just keep do_create simple by requiring both.

enkore:
> WRITE currently only covers do_create, so maybe we should rename it. In a way WRITE does not really need perfect REFCOUNT_SYNC, it just needs a conservative refcount that doesn‘t have false positives.
>
> Checking more than one might be needed in some situations, esp. if we go with REFCOUNT_SYNC and for now just keep do_create simple by requiring both.

Exactly my thoughts :)

borg/helpers.py
```diff
@@ -146,8 +159,18 @@ def load(cls, repository, key=None, force_tam_not_required=False):
...
         return manifest, key
 
+    def check_repository_compatibility(self, operation):
+        if operation == REPO_OPERATION_NO_CHECK:
+            return
+        if operation not in self.config:
+            return
+        requirements = self.config[operation]
+        if 'mandatory' in requirements and not set(requirements['mandatory']) <= SUPPORTED_REPO_FEATURES:
+            raise MandatoryFeatureUnsupported(set(requirements['mandatory']) - SUPPORTED_REPO_FEATURES)
```
ThomasWaldmann: see https://github.com/borgbackup/borg/pull/2134#discussion_r100644624

borg/helpers.py
```diff
@@ -98,6 +102,15 @@ def check_extension_modules():
         raise ExtensionModuleError
 
 
+SUPPORTED_REPO_FEATURES = frozenset()
+
+REPO_OPERATION_READ = 'read'
+REPO_OPERATION_CHECK = 'check'
+REPO_OPERATION_WRITE = 'write'
+REPO_OPERATION_DELETE = 'delete'
+REPO_OPERATION_NO_CHECK = '-'
```
textshell: Yes, check and no_check seem ambiguous, that needs improving. Replace REPO_OPERATION by OP_ and move in as members of Manifest?

borg/archiver.py
```diff
@@ -65,7 +68,18 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True, excl
     :param exclusive: (str or bool) lock repository exclusively (for writing)
     :param manifest: load manifest and key, pass them as keyword arguments
     :param cache: open cache, pass it as keyword argument (implies manifest)
+    :param compatibility: mandatory if not create and (manifest or cache), specifies mandatory feature category to check
     """
+
+    if not create and (manifest or cache):
+        if compatibility is None:
+            raise Exception("with_repository called without compatibility argument")
```
textshell: The way this is currently written, it raises these on first import of archiver.py. So these are not really exceptions that might need to be caught but more sanity checks done while loading the module. So i don‘t think this matters much.

borg/helpers.py
enkore: Make it a faux enum instead?

```py
class Manifest:
  # possibly as a class of manifest
  class Operation:
    READ = '...
    WRITE = '...
```

borg/helpers.py
```diff
@@ -146,8 +159,18 @@ def load(cls, repository, key=None, force_tam_not_required=False):
...
                 os.unlink(tam_required_file(repository))
+        manifest.check_repository_compatibility(operation)
         return manifest, key
 
+    def check_repository_compatibility(self, operation):
+        if operation == REPO_OPERATION_NO_CHECK:
+            return
+        if operation not in self.config:
+            return
+        requirements = self.config[operation]
```
enkore: ~ maybe use a separate dict in `config` for this, like `feature_flags` or so, to keep the `config` namespace clean.

textshell: currently it is structured
`manifest.config[b'write'] = {b'mandatory': [b'unknown-feature']}
`
But i’m open for other layouts. I think there are some features we want to store in the config that don‘t lockout older versions. So i would want to keep some distinction between mandatory and "as supported" features.

borg/helpers.py
textshell: fixed.

borg/helpers.py
textshell: Now implemented with the stdlib enum class/metaclass. I think requiring 3.4 is ok? Or do we still target 3.3?

borg/helpers.py
```diff
@@ -86,6 +86,10 @@ class PlaceholderError(Error):
     """Formatting Error: "{}".format({}): {}({})"""
 
 
+class MandatoryFeatureUnsupported(Error):
+    """Unsupported repository feature(s) {}. A newer version is required to access this repository."""
```
textshell: now "a newer version of borg"

textshell:
I hope i fixed all the clear cut review comments now. I decided to just include support for manifest version 2, worst case we wasted a version number.

Still open is encore’s question what `READ` actually means and the idea to have a dedicated `REFCOUNT_SYNC`. `REFCOUNT_SYNC` has overlap with `DELETE` because delete always needs correct refcounts.

I’m tempted by a fallback that still allows archive creation without `REFCOUNT_SYNC` but that would be hard to test unless we offered it as a genuine operation mode outside of some obscure fallback code path. `REFCOUNT_SYNC` would solve cache compatibility because that is the only critical cached data that needs handling at the moment it think.

enkore:
It's a bit of a coincidence, since I just finished [a prototype](https://gist.github.com/enkore/fce238674acc3963ae1b1f00be8c08a6) that does exactly what REFCOUNT_SYNC would be needed for, i.e. `borg create` without a cache (sync).

If we separate `REFCOUNT_SYNC` from `READ` then yeah... what is `READ` for?

textshell:
From discussion on irc:

For cache maintenance the basic principles are:
Wipe when
* the client understands a new features that has previously been ignored, OR
* a client doesn't understand a feature that was previously understood.

This can be implement fairly easily:
Each client version has a set of features which contains features that the client knows to handle `client_features`
When accessing a repository collect all mandatory features from the manifest into a `flat_featues` set.
Save in the cache two sets called `ignored_features` and `mandatory_features`.
When updating the cache also update:
`ignored_features` += `flat_featues` - `client_features`
`mandatory_features` += `flat_featues` & `client_features`

when accessing the cache:
if `ignored_features` contains any item from `client_features`: wipe and rebuild cache
if `mandatory_features` contains any item not in `client_features`: wipe and rebuild cache

Possible optimization. A future client could tweak the rules to in cases where it is known to be save in order to reduce rebuilds. But that will always be a case by case decision that can only be made when a new feature will be added, so this PR does not need to consider this in detail.

textshell:
i rebased this to something more recent and implemented the cache wiping strategy from the comment above. This should solve the cache problem.
Possibly we might want skip some of the operations when deciding what feature flags to use for cache compatibility. I did not do this in this patch. But possibly ignoring write would make some sense. This is only useful if we expect features that are mandatory for write but not for delete or check. One kind of feature where this might be useful would be a feature that just prevents regressions in structure after some kind of whole repository upgrades.

I lightly tested this manually. It still needs automated tests. One thing i'm unsure about is how to best test that in all the existing tests the cache wipe is not triggered. That would need some rather global way to make the cache wipe method throw an exception. Maybe a global test fixture would do, but i'm unsure how to disable that in just some tests.

Also still pending is the question what to do with the "READ" operation:
* Move to archive
* Remove
* Keep as is
I thing the first option would be best with a super simple model where each archive can have a feature_flags list of strings that are all mandatory. Not sure how mount of the whole repo would handle this though.

borg/archiver.py
```diff
@@ -537,7 +551,7 @@ def do_delete(self, args, repository):
             if not args.cache_only:
                 msg = []
                 try:
-                    manifest, key = Manifest.load(repository)
+                    manifest, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK)
```
enkore: minor;

given that every(?) time Manifest.load is called explicitly (outside with_repository) NO_OPERATION_CHECK is passed, make it the default?

borg/archiver.py
enkore: The assertion in the else branch is sufficient for both cases; just move the description there?

borg/archiver.py
```diff
@@ -67,7 +68,20 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True, excl
...
+    if not create and (manifest or cache):
+        if compatibility is None:
+            raise Exception("with_repository called without compatibility argument")
+        else:
+            assert type(compatibility) is tuple
+    else:
+        if compatibility is not None:
+            raise Exception("with_repository called with compatibility argument but would not check" + repr(compatibility))
+        if create:
+            compatibility = Manifest.NO_OPERATION_CHECK
```
enkore: ? do_create below specifies compatibility?

borg/helpers.py
```diff
@@ -173,8 +192,34 @@ def load(cls, repository, key=None, force_tam_not_required=False):
...
         return manifest, key
 
+    def check_repository_compatibility(self, operations):
+        for operation in operations:
+            assert isinstance(operation, self.Operation)
+            feature_flags = self.config.get(b'feature_flags', None)
+            if feature_flags is None:
+                return
+            if operation.value.encode() not in feature_flags:
+                return
```
enkore: Shouldn't this be `continue`, so the other operations are still checked?

borg/helpers.py
```diff
@@ -124,8 +129,22 @@ def check_extension_modules():
...
+    @enum.unique
+    class Operation(enum.Enum):
+        READ = 'read'
+        CHECK = 'check'
+        WRITE = 'write'
+        DELETE = 'delete'
+
+    NO_OPERATION_CHECK = tuple()
+
+    ALL_POSSIBLE_FEATURES = set([])
```
enkore: minor; also a frozenset()?

borg/helpers.py
```diff
@@ -124,8 +129,22 @@ def check_extension_modules():
         raise ExtensionModuleError
 
 
+SUPPORTED_REPO_FEATURES = frozenset()
```
enkore: minor; move this constant to Manifest as well, where the other related constants live as well?

borg/cache.py
```diff
@@ -390,6 +400,39 @@ def legacy_cleanup():
...
+        logger.warning("Discarding incompatible cache and forcing a cache rebuild")
+        archive_path = os.path.join(self.path, 'chunks.archive.d')
+        if os.path.isdir(archive_path):
+            shutil.rmtree(os.path.join(self.path, 'chunks.archive.d'))
+            os.makedirs(os.path.join(self.path, 'chunks.archive.d'))
+        self.chunks = ChunkIndex()
+        with open(os.path.join(self.path, 'files'), 'wb') as fd:
+            pass  # empty file
+        self.manifest_id = ''
+        self.config.set('cache', 'manifest', '')
```
enkore: Reuse existing Cache.delete?

enkore:
> I lightly tested this manually. It still needs automated tests. One thing i'm unsure about is how to best test that in all the existing tests the cache wipe is not triggered. That would need some rather global way to make the cache wipe method throw an exception. Maybe a global test fixture would do, but i'm unsure how to disable that in just some tests.

You could patch (unittest.mock.patch) the delete/create method with a mock, then assert it was not called (`not mock.called`).

enkore:
> Also still pending is the question what to do with the "READ" operation:
>
>    Move to archive
>    Remove
>    Keep as is
>
> I thing the first option would be best with a super simple model where each archive can have a feature_flags list of strings that are all mandatory. Not sure how mount of the whole repo would handle this though.

I'm not a huge fan of fine-grained "may work / may not work" TBH. Keeping it repository-wide seems easiest to me and has clear semantics; either a version can read the repository, and then it can read all archives (to at least some extent), or it doesn't and it can't read anything.

textshell:
@ docs:
Yes. This needs docs. Maybe even in the internal part of the rtd docs.

@ read
I'm a bit torn, just having one large archive (with new structure) doesn't need to mean that other archives are unreadable. But of course one simple global flag is easier and sufficient for just avoiding harm.

borg/archiver.py
enkore: I meant the default on `Manifest.load` itself, though. Fair enough either way, you make a solid point.

enkore:
Regarding flag granularity,
it's not just easier from a developer perspective, but like I mentioned above — it's simpler for using it as well; questions such as "Can I restore that backup/archive?" should be easy and confident to answer.

enkore:
Code comments; that's probably the begin/submit review flow offered by Github. It seems to always create new threads, even when replying to another comment. It also makes it easy to not publish comments; I don't use it. The approve/reject/global-comment functionality is accessible without using the other parts of that workflow.

enkore:
Regarding cache flags,
the cache is composed of a few distinct structures (chunks, files and chunks.archive.d). What do you think about giving a distinct set of flags to each of these? For example, we know that chunks.archive.d wastes disk spaces due to its format; or that the files cache format could be changed to better accommodate cache-sync-less operation.
However, in all of these cases backwards-compatible approaches are clearly viable as well. So I'm thinking leave it as is; having too many joints (version fields, feature flags in different places) just increases the chance that they're stuck and thus pointless.

borg/cache.py
```diff
@@ -390,6 +400,39 @@ def legacy_cleanup():
...
+
+    def update_compatibility(self):
+        operation_to_features_map = self.manifest.get_all_mandatory_features()
+        my_features = Manifest.SUPPORTED_REPO_FEATURES
+        repo_features = set()
+        for operation, features in operation_to_features_map.items():
+            repo_features.update(features)
+
+        self.ignored_features.update(repo_features - my_features)
+        self.mandatory_features.update(repo_features & my_features)
```
enkore: Aha, so it turns out I read this kinda wrong :)

So this is not a per se separate set of flags for the cache. Rather, it checks if the set of flags stored in the config is a subset of the supported repository flags.

enkore: Aha. That makes a lot of sense there. This allows to build a cache that might only reflect some subset of the data in the repository, and then rule operations out that would require a complete cache (say, *delete*).

enkore:
To reduce the overall required work (and because this PR is open since before I made the new branching model, I think) I propose to merge this PR into 1.0-maint (as an exception) and cherry pick into master from there.

enkore:
Regarding https://github.com/borgbackup/borg/pull/2134#issuecomment-285946168

So, WRITE does not include synchronizing refcounts. WRITE just means you can update the manifest. If/how you get a list of chunks existing in the repository is NOT covered by WRITE. This would basically make it the usually-lowest level of access.

READ means being able to read ALL archives in the repository, but maybe not every piece of data. Say, we add a new item key with some crazy new FS flags — that would not be a mandatory READ feature. But using new kind of crypto, that would be a mandatory READ feature.

DELETE means being able to establish correct reference counts. Which probably means READ + all additions to referencing chunks (think #1681).

CHECK means being able to do all that DELETE does, but also (?).

We don't need answers every yet. Just a hunch that the category is correct. Which looks good to me so far.

enkore:
Docs are in #2602. You can merge it into this or we keep it independent, doesn't really matter.

enkore:
(?) Idea: Store another key under *config* in the manifest, which gives a short description for each flag. Error messages could then include the descriptions for even better error messages.

borg/cache.py
```diff
@@ -390,6 +400,39 @@ def legacy_cleanup():
             self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
+    def check_cache_compatibility(self):
+        my_features = Manifest.SUPPORTED_REPO_FEATURES
+        if self.ignored_features & my_features:
+            return False
```
enkore: I'm not really sure if I understand this correctly.

ignored_features is the set of features that are mandated for some operations, but not those the cache was originally built for, and these were not supported by the client building / last updating the cache.

So `ignored_features & my_features` would be the set of features that were *not* supported by the previous client. But why wipe the cache here?

enkore: So per https://github.com/borgbackup/borg/pull/2134#issuecomment-303220459 

> So ignored_features & my_features would be the set of features that were not supported by the previous client. But why wipe the cache in this instance?

I.e. the repository has had a feature enabled, which the client just now learned, so as per ^ the cache might need rebuilding. But as per ^^ this can be made a case-to-case decision anyway. Ok. Now I get it :)

ThomasWaldmann: would be cool if the hard to understand stuff ends up in some helpful comments in the src. :)

enkore: Long-form docs are in #2602. Doesn't replace a few well-placed comments, though :)

textshell:
WRITE currently tries to do a normal chunk cache sync. Which i think is ok to try. Because for write it is always safe to assume certain chunks are not in the repository when they in fact are. The cache sync is currently not robust with regard to strange and new archive formats. That is strictly speaking not a correctness problem, but an unsightly "crash".

CHECK is the highest level, assuming total understanding of everything because trying to ensure internal consistency might corrupt things. DELETE in contrast does not change anything valuable, except that it needs to know about all chunks to free and needs a chunk refcount cache that is 100% accurate.

textshell:
Do we still consider this to be a candidate for 1.0-maint? I think it would be nice to have and this branch still targets 1.0-maint as by previous policy.

textshell:
To add about "CHECK". This could also cover various parts that are not worth having separate flags for. E.g. backwards compatible changes in the repokey and various other "special" commands that are too rarely used to be worrying about more precise compatibility checks.

ThomasWaldmann:
flake8 is unhappy:
```
./conftest.py:45:9: E306 expected 1 blank line before a nested definition, found 0
./conftest.py:57:35: W292 no newline at end of file
./borg/testsuite/archiver.py:1017:27: E712 comparison to True should be 'if cond is True:' or 'if cond:'
./borg/testsuite/archiver.py:1020:9: E303 too many blank lines (2)
./borg/testsuite/archiver.py:1028:5: E303 too many blank lines (2)
```

textshell:
I have squashed this changeset down to something more managable. Until e04dad5 this is (modulo fixes for flake8) without any new changes.
The two follow up commits add some additional comments.

I have not yet removed "TMP: debug tune" but i think even if we want to have something like this, it should go in via a separate PR, so i will remove this before merging.

textshell:
Ok, i squashed the last fixups and removed some unneeded print in the tests.

@ThomasWaldmann What do you think?

ThomasWaldmann:
ok. \o/

as discussed on irc we'll first release this via master in a 1.1.0 beta.
depending on how that works out, it could be in 1.0.11 or 1.0.12 later.

enkore:
Who picks this into master?

textshell:
I'm currently trying to rebase this to master.

enkore:
Reroll may be easier / quicker.

PR #2135, opened 2017-02-06
paperkey+qr html+js template

textshell:
Typing in a paper key is quite some effort. This adds an interactive html+js template that allows printing a paperkey that also contains a QR-Code which contains the whole keyfile.

The basic idea to use html+js is that it’s easy to generate reasonable looking printable output and that the user can copy the html file to any sufficiently trusted system to print out the key.

The template allows configuration of the level or error correction, the QR-code size and the text size. Additionally it allows the user to change the help text that is printed along side of the actual key data.

Downside is that this bundles two javascript libraries (one for qr code generation and a sha256 implementation). But i think this is a reasonable trade off because this template works isolated and thus even if there are potential security problems in the embedded code it should not be possible to exploit anything. Also i think having a self contained file that doesn‘t depend on any additional files is important for usability here. Of course we could also have a build step that joins different parts together, but that seems overly complicated in this case.

@LocutusOfBorg Can you think of anything that needs to be done to make this acceptable for debian?


enkore:
Ohh, this is a very nice idea indeed.

I wonder whether we can streamline this a bit; AIUI: find/open template.html, generate key export and put it into there, get QR code.

Perhaps: Separate export command, generates HTML with the necessary data baked in (secure tmp file), open (stdlib `webbrowser` module) it -> done. Finding data files of OS packages is often not so easy.

textshell:
Currently the usage would be to use the documentation (it should be linked from readthedocs and from locally generated documentation) to find the printing template. And use a `borg export key` exported key.

If we can arrange for borg to find the template, it shouldn‘t be hard to bake the key in there and save the resulting html. We didn‘t worry about secure temp files with the other key export options so maybe we don‘t need to worry here. But i expect in may cases the browser will be run on a different system than where borg runs.

enkore:
Ah, I missed that part.

> Currently the usage would be to use the documentation (it should be linked from readthedocs and from locally generated documentation) to find the printing template. And use a borg export key exported key.

textshell:
A usable link to play around with the html file: https://rawgit.com/textshell/borg/feature/paperkey-doc-qr/docs/paperkey.html

I’ve been thinking about the generic html vs generated html with embedded key. I would like to keep the generic one and i think it is nice to have it available from the documentation.
I’m also happy to have a export function that embeds a specific key into a ready made html file. The modification of the html file should be very easy (basically paste it into the textarea). The part i don‘t really know how to solve well is getting access to the html file from the borg code so that is works in all installation scenarios including binary distribution. I could need some help there.

enkore:
I'd just create a separate module with the HTML file as a string embedded ("""" syntax).

Declaring it as a data file should work too, even with PyInstaller, but is a bit more finicky to do, something like this:

```py
setup.py

setup(
...
      package_dir={'': 'src/'},
      package_data={'borg': ['borg/template.html']}, # or without borg/ not sure
...
```

would be different for 1.0-maint.

textshell:
setup.py beats me. The only way i could keep paperkey.html in the doc directory for spinx and have it in the borg package so it can be imported (without installing to random locations when installing borg with pip) which might work with pip install -e . and proper installs is to use a symlink for the install -e case and data_files in setup.py for proper installs.
Of course this mean -e will not really work without symlink support. If anyone has a good idea...

The actual export code works, but is just a 5 minute effort. We need to decide if this mode should be an option to borg key export or it’s own subcommand.

borg/paperkey.html
```diff
@@ -0,0 +1 @@
+../docs/paperkey.html
```
ThomasWaldmann: what's that? a symlink? please try without.

textshell: I did try without, but i didn’t have any luck. But i really don‘t grok setuptools/distutils. If you have any idea how to make this work so that the file can be shared between docs, `pip install -e .` and a real install (pyinstaller, pip install) i’d love to have something cleaner.

textshell: it seems sphinx is easy to get to use paperkey.html from the python package than the other way round. So symlink is not gone. Egg seems to contain the file, but i didn‘t actually test pyinstaller yet.

ThomasWaldmann:
hmm, this is kind of last-minute addition to 1.0-maint before 1.0.10 release (guess I'll work on that tonight).

are we confident enough this works good enough and doesn't cause any regressions?

borg/keymanager.py
```diff
@@ -77,16 +78,27 @@ def store_keyblob(self, args):
...
-                fd.write('\n')
+            fd.write(self.get_keyfile_data())
 
     def export(self, path):
         self.store_keyfile(path)
 
+    def export_qr(self, path):
+        with open(path, 'wb') as fd:
+            key_data = self.get_keyfile_data()
+            html = pkgutil.get_data('borg', 'paperkey.html')
```
ThomasWaldmann: no idea about pkgutil, so I just ask: will this work under all circumstances, no matter how users installed borg?

pip, linux dist package, other dist package, from git, ...

enkore:
The code changes are quite small, apart from the added stuff the only change is cosmetics (get_key_data) and the if guarding the feature.

I can push a simple test later - should help with verifying that it works inside PyInstaller.

borg/keymanager.py
enkore: This works through the standard importlib; since `borg` was already imported... should be hard to break.

ThomasWaldmann: ok

PR #2145, opened 2017-02-12
manpages part two

### LINKED ISSUE #2130: create borg(1) man page, polish man pages a bit

enkore:
Fixes #2130 

I've committed the updated man pages here contrary to what the new docs say to make it easier to see what actually changed in them.


.gitattributes
```diff
@@ -1,3 +1,5 @@
 borg/_version.py export-subst
 
 *.py diff=python
+docs/usage merge=ours
+docs/man merge=ours
```
ThomasWaldmann: git config --global merge.ours.driver true

is that also needed? (found that on SO)

docs/changes.rst
```diff
@@ -232,7 +232,7 @@ Version 1.0.10rc1 (2017-01-29)
 Bug fixes:
 
 - borg serve: fix transmission data loss of pipe writes, #1268
-  This affects only the cygwin platform (not Linux, *BSD, OS X).
+  This affects only the cygwin platform (not Linux, BSD, OS X).
```
ThomasWaldmann: Looks like some changes should go into 1.0-maint, while others can't go there. So it's a bit of a mixup here.

enkore: Meh, I just cherry pick these.

At the moment merging maintenance branches is just-so still viable. I think in the future we'll need to go for cherry picking stuff always, the past few merges were always getting messier and messier...

.gitattributes
enkore: Shouldn't be, it's listed on the man page as a built-in strategy. But another mistake kept it from working; fixed-- the option does not work like a .gitignore, it only works for files, so a `/*` was missing. Seems to work now.

docs/changes.rst
ThomasWaldmann: yeah, had same experience, merging is a pain. not sure that cherrypicking is less pain though.

PR #2148, opened 2017-02-12
 add test for export key --qr-html 

enkore:
Follow up to #2135

### DIFF
--- a/borg/paperkey.html
+++ b/borg/paperkey.html
@@ -2171,8 +2171,11 @@ if (typeof define == 'function' && define.amd) define([], function() { return Sh
         }
     }
 
-
-
+    /* center the QR code on the page */
+    #qr {
+      width: 100%;
+      text-align: center;
+    }
 </style>
 </head>
 <body>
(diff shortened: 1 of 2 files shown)


borg/paperkey.html
```diff
@@ -2171,8 +2171,11 @@
         }
     }
 
-
-
+    /* center the QR code in the page */
```
ThomasWaldmann: center *on* the page?

borg/paperkey.html
```diff
@@ -2217,7 +2220,7 @@
     <div id="printout">
         <div id="title" contenteditable>BorgBackup Printable Key Backup</div>
         <div contenteditable>To restore either scan the QR code below, decode it and import it using
-<pre>borg key-import /path/to/repo scannedfile</pre>
+<pre>borg key import /path/to/repo scannedfile</pre>
```
ThomasWaldmann: hmm, what is the format (mimetype) of scannedfile?

"scannedfile" sounds a bit like .tiff or other bitmap format, but I guess it needs to be text? or binary?

borg/testsuite/archiver.py
```diff
@@ -1318,6 +1318,19 @@ def test_key_export_repokey(self):
...
+        repo_id = self._extract_repository_id(self.repository_path)
+        self.cmd('key', 'export', '--qr-html', self.repository_location, export_file)
+
+        with open(export_file, 'r') as fd:
+            export_contents = fd.read()
+
+        assert bin_to_hex(repo_id) in export_contents
+        assert export_contents.startswith('<!doctype html>')
+        assert export_contents.endswith('</html>')
+
```
ThomasWaldmann: good :)

borg/paperkey.html
enkore: imho clear enough that with "scan the QR code" using a QR scanner is meant not an imaging scanner. In the former case you'll get text / text file which can be directly consumed here. (After copying to PC).

ThomasWaldmann: ok

enkore:
=> #2149

PR #2149, opened 2017-02-12
Release 1.0.10

enkore:


docs/api.rst
```diff
@@ -1,84 +1,95 @@
...
+.. automodule:: borg.repository
     :members:
     :undoc-members:
 
 .. automodule:: borg.hashindex
     :members:
     :undoc-members:
 
-.. automodule:: borg.compress
-    :members: get_compressor, Compressor, CompressorBase
```
ThomasWaldmann: this was lost.

the release todo checklist is in fact not quite complete here, but we have a bug for the build api being problematic at least.

enkore: I'm also going to re-remove the non-linux platform modules.

setup.py
```diff
@@ -230,10 +230,13 @@ def run(self):
...
         with open("docs/api.rst", "w") as doc:
             doc.write("""
+.. IMPORTANT: this file is auto-generated by "setup.py build_api", do not edit!
+
+
 API Documentation
 =================
 """)
-            for mod in glob('borg/*.py') + glob('borg/*.pyx'):
+            for mod in sorted(glob('borg/*.py') + glob('borg/*.pyx')):
```
ThomasWaldmann: ah, cool :)

docs/api.rst
enkore: Hm, checked without the explicit :members:, Sphinx (1.5.2) found them nonetheless.

enkore:
Released

PR #2157, opened 2017-02-14
archived file items: add size metadata

ThomasWaldmann:
if an item has a chunk list, pre-compute the total size and store it into "size" metadata entry.

this speeds up access to item size (e.g. for regular files) and could also be used to verify the validity of the chunks list.

note about hardlinks: size is only stored for hardlink masters (only they have an own chunk list)

- [x] add a size entry to metadata IF item has a chunks list
- [x] keep sum-of-chunk-sizes as fallback for old archives
- [x] optimize borg list archive to directly use metadata file size
- [x] optimize size-related FUSE code
- [ ] do similar pre-computation / optimization for csize as for size? -> No, don't.
- [x] add borg check code that checks metadata file size vs. sum of chunk sizes
- [x] add borg extract code that checks metadata file size vs. sum of chunk sizes (or extracted file size)

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -777,7 +777,7 @@ Utilization of max. archive size: {csize_max:.0%}
         length = len(item.chunks)
         # the item should only have the *additional* chunks we processed after the last partial item:
         item.chunks = item.chunks[from_chunk:]
-        item.file_size(memorize=True)
+        item.get_size(memorize=True)
         item.path += '.borg_part_%d' % number
         item.part = number
         number += 1
@@ -826,7 +826,7 @@ Utilization of max. archive size: {csize_max:.0%}
         )
         fd = sys.stdin.buffer  # binary
         self.chunk_file(item, cache, self.stats, backup_io_iter(self.chunker.chunkify(fd)))
-        item.file_size(memorize=True)
+        item.get_size(memorize=True)
         self.stats.nfiles += 1
         self.add_item(item)
         return 'i'  # stdin
@@ -887,7 +887,7 @@ Utilization of max. archive size: {csize_max:.0%}
                 cache.memorize_file(path_hash, st, [c.id for c in item.chunks])
             status = status or 'M'  # regular file, modified (if not 'A' already)
         item.update(self.stat_attrs(st, path))
-        item.file_size(memorize=True)
+        item.get_size(memorize=True)
         if is_special_file:
             # we processed a special file like a regular file. reflect that in mode,
             # so it can be extracted / accessed in FUSE mount like a regular file:
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1306,6 +1306,13 @@ class ArchiveChecker:
                 logger.info('{}: Completely healed previously damaged file!'.format(item.path))
                 del item.chunks_healthy
             item.chunks = chunk_list
+            if 'size' in item:
+                item_size = item.size
+                item_chunks_size = item.get_size(compressed=False, from_chunks=True)
+                if item_size != item_chunks_size:
+                    # just warn, but keep the inconsistency, so that borg extract can warn about it.
+                    logger.warning('{}: size inconsistency detected: size {}, chunks size {}'.format(
+                                   item.path, item_size, item_chunks_size))
 
         def robust_iterator(archive):
             """Iterates through all archive items
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -777,7 +777,7 @@ Utilization of max. archive size: {csize_max:.0%}
         length = len(item.chunks)
         # the item should only have the *additional* chunks we processed after the last partial item:
         item.chunks = item.chunks[from_chunk:]
-        item.size = sum(chunk.size for chunk in item.chunks)
+        item.file_size(memorize=True)
         item.path += '.borg_part_%d' % number
         item.part = number
         number += 1
@@ -826,7 +826,7 @@ Utilization of max. archive size: {csize_max:.0%}
         )
         fd = sys.stdin.buffer  # binary
         self.chunk_file(item, cache, self.stats, backup_io_iter(self.chunker.chunkify(fd)))
-        item.size = sum(chunk.size for chunk in item.chunks)
+        item.file_size(memorize=True)
         self.stats.nfiles += 1
         self.add_item(item)
         return 'i'  # stdin
@@ -887,7 +887,7 @@ Utilization of max. archive size: {csize_max:.0%}
                 cache.memorize_file(path_hash, st, [c.id for c in item.chunks])
             status = status or 'M'  # regular file, modified (if not 'A' already)
         item.update(self.stat_attrs(st, path))
-        item.size = sum(chunk.size for chunk in item.chunks)
+        item.file_size(memorize=True)
         if is_special_file:
             # we processed a special file like a regular file. reflect that in mode,
             # so it can be extracted / accessed in FUSE mount like a regular file:
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -467,13 +467,20 @@ Utilization of max. archive size: {csize_max:.0%}
         has_damaged_chunks = 'chunks_healthy' in item
         if dry_run or stdout:
             if 'chunks' in item:
+                item_chunks_size = 0
                 for _, data in self.pipeline.fetch_many([c.id for c in item.chunks], is_preloaded=True):
                     if pi:
                         pi.show(increase=len(data), info=[remove_surrogates(item.path)])
                     if stdout:
                         sys.stdout.buffer.write(data)
+                    item_chunks_size += len(data)
                 if stdout:
                     sys.stdout.buffer.flush()
+                if 'size' in item:
+                    item_size = item.size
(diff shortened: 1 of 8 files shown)


src/borg/archive.py
```diff
@@ -774,6 +774,7 @@ def write_part_file(self, item, from_chunk, number):
         length = len(item.chunks)
         # the item should only have the *additional* chunks we processed after the last partial item:
         item.chunks = item.chunks[from_chunk:]
+        item.size = sum(chunk.size for chunk in item.chunks)
```
ThomasWaldmann: this "sum of chunk sizes" function is duplicated at multiple places.
could be moved to helpers.py or item.pyx (can it be sped up with cython?).

src/borg/fuse.py
```diff
@@ -260,6 +260,7 @@ def getattr(self, inode, ctx=None):
         size = 0
         dsize = 0
         if 'chunks' in item:
+            # if we would not need to compute dsize, we could get size quickly from item.size, if present.
```
ThomasWaldmann: shall we just kill that? dsize reflects the space taken up in the repository and is used for st_blocks below.

we could also just say st_blocks = size / blocksize and not consider deduplication and safe the loop below for many cases.

src/borg/archive.py
```diff
@@ -882,6 +884,7 @@ def process_file(self, path, st, cache, ignore_inode=False):
                 cache.memorize_file(path_hash, st, [c.id for c in item.chunks])
             status = status or 'M'  # regular file, modified (if not 'A' already)
         item.update(self.stat_attrs(st, path))
+        item.size = sum(chunk.size for chunk in item.chunks)
```
ThomasWaldmann: could also get size from st.st_size. risky due to race conditions? maybe it is safer like above.

ThomasWaldmann:
note (unrelated to this change): I guess the whole code that deals with file sizes needs a review, whether it works correctly for hardlink slaves.

enkore:
Why not throw st_blocks away entirely? (st_blocks=1). Even now it's not useful.

ThomasWaldmann:
Note for 1.0-maint: fix the fuse st_blocks calculation to use integer division (not float).

ThomasWaldmann:
About "do similar pre-computation / optimization for csize as for size":

Pro:

- quicker borg list csize placeholder
- in general: quicker access to csize without having to sum it up first

Con:

- would increase archive size by some bytes per regular file
- we want to get rid of csize in the chunk entries at some time (but we still have it)

ThomasWaldmann:
Any comments?

enkore:
Wouldn't add something that we kinda want to get rid off already

src/borg/archive.py
enkore: Doesn't have `Item` this one?

src/borg/archive.py
enkore: st_size can be raced, but `item.chunks` is already safe - this way it's correct.

src/borg/archive.py
ThomasWaldmann: ah, good idea, I'll check.

ThomasWaldmann:
rebased on master / added future optimization comment

src/borg/archive.py
```diff
@@ -1289,6 +1292,24 @@ def verify_file_chunks(item):
                 logger.info('{}: Completely healed previously damaged file!'.format(item.path))
                 del item.chunks_healthy
             item.chunks = chunk_list
+            if 'size' in item:
+                item_size = item.size
+                item_chunks_size = item.get_size(compressed=False, from_chunks=True, from_chunks_healthy=False)
```
ThomasWaldmann: get_size() change still TODO

src/borg/archive.py
```diff
@@ -1289,6 +1292,24 @@ def verify_file_chunks(item):
...
+                    if item_size != item_chunks_size:
+                        if has_chunks_healthy:
+                            item_chunks_healthy_size = item.get_size(compressed=False,
+                                                                     from_chunks=False, from_chunks_healthy=True)
+                        else:
+                            item_chunks_healthy_size = None
+                        logger.warning('{}: size inconsistency detected: '
+                                       'size {}, chunks size {}, chunks healthy size {}'.format(
+                                       item.path, item_size, item_chunks_size, item_chunks_healthy_size))
+                        # ok, what now?
```
ThomasWaldmann: ?

src/borg/archive.py
```diff
@@ -1289,6 +1292,24 @@ def verify_file_chunks(item):
...
                 del item.chunks_healthy
             item.chunks = chunk_list
+            if 'size' in item:
+                item_size = item.size
+                item_chunks_size = item.get_size(compressed=False, from_chunks=True, from_chunks_healthy=False)
+                if __name__ == '__main__':
+                    if item_size != item_chunks_size:
+                        if has_chunks_healthy:
+                            item_chunks_healthy_size = item.get_size(compressed=False,
+                                                                     from_chunks=False, from_chunks_healthy=True)
```
ThomasWaldmann: hmm, we can't get anything useful from chunks_healthy list. if a chunkid is in there, but not in chunks list, we have lost that chunk, so we can't determine its size. so, this is not implementable.

src/borg/archive.py
```diff
@@ -1289,6 +1292,24 @@ def verify_file_chunks(item):
                 logger.info('{}: Completely healed previously damaged file!'.format(item.path))
                 del item.chunks_healthy
             item.chunks = chunk_list
+            if 'size' in item:
+                item_size = item.size
+                item_chunks_size = item.get_size(compressed=False, from_chunks=True, from_chunks_healthy=False)
+                if __name__ == '__main__':
```
ThomasWaldmann: wat?

src/borg/archive.py
```diff
@@ -1289,6 +1292,13 @@ def verify_file_chunks(item):
                 logger.info('{}: Completely healed previously damaged file!'.format(item.path))
                 del item.chunks_healthy
             item.chunks = chunk_list
+            if 'size' in item:
+                item_size = item.size
+                item_chunks_size = item.get_size(compressed=False, from_chunks=True)
+                if item_size != item_chunks_size:
+                    logger.warning('{}: size inconsistency detected: size {}, chunks size {}'.format(
+                                   item.path, item_size, item_chunks_size))
+                    # ok, what now? we don't know whether item_size or item_chunks_size is correct.
```
ThomasWaldmann: ?

src/borg/archive.py
enkore: Accidental IDE shortcut?

ThomasWaldmann: guess so...

src/borg/item.pyx
```diff
@@ -156,6 +156,11 @@ class Item(PropDict):
     ctime = PropDict._make_property('ctime', int)
     mtime = PropDict._make_property('mtime', int)
 
+    # size is only present for items with a chunk list and then it is sum(chunk_sizes)
+    # compatibility note: this is a new feature, in old archives size will be missing.
+    size = PropDict._make_property('size', int)
+    # XXX add similar for csize? (as long as we did not get rid of it in the chunks list entries)
```
ThomasWaldmann: todo: remove comment line

src/borg/item.pyx
```diff
@@ -168,13 +173,48 @@ class Item(PropDict):
...
+                    except AttributeError:
+                        # not a hardlink slave, likely a directory or special file w/o chunks
+                        chunks = None
+                    else:
+                        # hardlink slave, try to fetch hardlink master's chunks list
+                        # todo: put precomputed size into hardlink_masters' values and use it, if present
+                        chunks, _ = hardlink_masters.get(master, (None, None))
+                if chunks is None:
+                    return 0
+            size = sum(getattr(chunk, attr) for chunk in chunks)
```
ThomasWaldmann: the getattr doesn't work

ThomasWaldmann:
I'll try to clean up the changesets a bit. Is there anything else?

src/borg/archive.py
ThomasWaldmann: guess there is nothing to be done except emitting the warning.

enkore: print warning, update item_size, because item_chunks_size is all we got.

ThomasWaldmann: how can we be sure that item_size is wrong and not the chunks list?
remember that bug a while ago, where the chunk list contents were lost...

enkore: Yeah but if you'd mount or extract it you'd only get item_chunks_size bytes, no?

ThomasWaldmann: yup. but if we do no patch the inconsistency, extract would warn.

otherwise it would not notice, would not warn.

enkore: Hm okay let's just leave it like now

ThomasWaldmann: ok, i added a comment about that.

ThomasWaldmann:
changesets cleaned:
- some superseded comments removed
- added a comment about why not patching the filesize inconsistency in borg check
- collapsed some fixups

ThomasWaldmann:
merge?

PR #2160, opened 2017-02-14
add "Cryptography in Borg" and "Remote RPC protocol security" sections

enkore:
The former section is a bit older (Nov 2016) and has been the piece
responsible for finding CVE-2016-10099, since while writing it I
wondered how the manifest was authenticated to actually
*be* the manifest. Well. There it is ;)

It has been edited to final form only recently (~mid Jan 2017 and
a few moments ago) and should now be ready for review.

The latter section is new.

---

I think it's sufficient to have this just in the 1.1.x docs.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -1,0 +1,279 @@
+
+.. somewhat surprisingly the "bash" highlighter gives nice results with
+   the pseudo-code notation used in the "Encryption" section.
+
+.. highlight:: bash
+
+========
+Security
+========
+
+Cryptography in Borg
+====================
+
+Attack model
+------------
+
+The attack model of Borg is that the environment of the client process
+(e.g. ``borg create``) is trusted and the repository (server) is not. The
+attacker has any and all access to the repository, including interactive
+manipulation (man-in-the-middle) for remote repositories.
+
+Furthermore the client environment is assumed to be persistent across
+attacks (practically this means that the security database cannot be
+deleted between attacks).
+
+Under these circumstances Borg guarantees that the attacker cannot
+
+1. modify the data of any archive without the client detecting the change
+2. rename, remove or add an archive without the client detecting the change
+3. recover plain-text data
+4. recover definite (heuristics based on access patterns are possible)
+   structural information such as the object graph (which archives
+   refer to what chunks)
+
+The attacker can always impose a denial of service per definition (he could
+forbid connections to the repository, or delete it entirely).
+
+Structural Authentication
+-------------------------
+
+Borg is fundamentally based on an object graph structure (see :ref:`internals`),
+where the root object is called the manifest.
+
+Borg follows the `Horton principle`_, which states that
+not only the message must be authenticated, but also its meaning (often
+expressed through context), because every object used is referenced by a
+parent object through its object ID up to the manifest. The object ID in
+Borg is a MAC of the object's plaintext, therefore this ensures that
+an attacker cannot change the context of an object without forging the MAC.
+
+In other words, the object ID itself only authenticates the plaintext of the
+object and not its context or meaning. The latter is established by a different
+object referring to an object ID, thereby assigning a particular meaning to
+an object. For example, an archive item contains a list of object IDs that
+represent packed file metadata. On their own it's not clear that these objects
+would represent what they do, but by the archive item referring to them
+in a particular part of its own data structure assigns this meaning.
+
+This results in a directed acyclic graph of authentication from the manifest
+to the data chunks of individual files.
+
+.. rubric:: Authenticating the manifest
+
+Since the manifest has a fixed ID (000...000) the aforementioned authentication
+does not apply to it, indeed, cannot apply to it; it is impossible to authenticate
+the root node of a DAG through its edges, since the root node has no incoming edges.
+
+With the scheme as described so far an attacker could easily replace the manifest,
+therefore Borg includes a tertiary authentication mechanism (TAM) that is applied
+to the manifest since version 1.0.9 (see :ref:`tam_vuln`).
+
+TAM works by deriving a separate key through HKDF_ from the other encryption and
+authentication keys and calculating the HMAC of the metadata to authenticate [#]_::
+
+    # RANDOM(n) returns n random bytes
+    salt = RANDOM(64)
+
+    ikm = id_key || enc_key || enc_hmac_key
+    # *context* depends on the operation, for manifest authentication it is
+    # the ASCII string "borg-metadata-authentication-manifest".
+    tam_key = HKDF-SHA-512(ikm, salt, context)
+
+    # *data* is a dict-like structure
+    data[hmac] = zeroes
+    packed = pack(data)
+    data[hmac] = HMAC(tam_key, packed)
+    packed_authenticated = pack(data)
+
+Since an attacker cannot gain access to this key and also cannot make the
+client authenticate arbitrary data using this mechanism, the attacker is unable
+to forge the authentication.
+
+This effectively 'anchors' the manifest to the key, which is controlled by the
+client, thereby anchoring the entire DAG, making it impossible for an attacker
+to add, remove or modify any part of the DAG without Borg being able to detect
+the tampering.
+
+Note that when using BORG_PASSPHRASE the attacker cannot swap the *entire*
+repository against a new repository with e.g. repokey mode and no passphrase,
+because Borg will abort access when BORG_PASSPRHASE is incorrect.
+
+However, interactively a user might not notice this kind of attack
+immediately, if she assumes that the reason for the absent passphrase
+prompt is a set BORG_PASSPHRASE. See issue :issue:`2169` for details.
+
+.. [#] The reason why the authentication tag is stored in the packed
+       data itself is that older Borg versions can still read the
+       manifest this way, while a changed layout would have broken
+       compatibility.
+
+Encryption
+----------
+
+Encryption is currently based on the Encrypt-then-MAC construction,
+which is generally seen as the most robust way to create an authenticated
+encryption scheme from encryption and message authentication primitives.
+
+Every operation (encryption, MAC / authentication, chunk ID derivation)
+uses independent, random keys generated by `os.urandom`_ [#]_.
+
+Borg does not support unauthenticated encryption -- only authenticated encryption
+schemes are supported. No unauthenticated encryption schemes will be added
+in the future.
+
+Depending on the chosen mode (see :ref:`borg_init`) different primitives are used:
+
+- The actual encryption is currently always AES-256 in CTR mode. The
+  counter is added in plaintext, since it is needed for decryption,
+  and is also tracked locally on the client to avoid counter reuse.
+
+- The authentication primitive is either HMAC-SHA-256 or BLAKE2b-256
+  in a keyed mode. HMAC-SHA-256 uses 256 bit keys, while BLAKE2b-256
+  uses 512 bit keys.
+
+  The latter is secure not only because BLAKE2b itself is not
+  susceptible to `length extension`_, but also since it truncates the
+  hash output from 512 bits to 256 bits, which would make the
+  construction safe even if BLAKE2b were broken regarding length
+  extension or similar attacks.
+
+- The primitive used for authentication is always the same primitive
+  that is used for deriving the chunk ID, but they are always
+  used with independent keys.
+
+Encryption::
+
+    id = AUTHENTICATOR(id_key, data)
+    compressed = compress(data)
+
+    iv = reserve_iv()
+    encrypted = AES-256-CTR(enc_key, 8-null-bytes || iv, compressed)
+    authenticated = type-byte || AUTHENTICATOR(enc_hmac_key, encrypted) || iv || encrypted
+
+
+Decryption::
+
+    # Given: input *authenticated* data, possibly a *chunk-id* to assert
+    type-byte, mac, iv, encrypted = SPLIT(authenticated)
+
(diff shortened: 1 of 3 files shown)


docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+We believe that other kinds of attacks, especially critical vulnerabilities
+like remote code execution are inhibited by the design of the protocol:
+
+1. The server cannot send requests to the client on it's own accord,
+   it only can send responses. This avoids "unexpected inversion of control"
+   issues.
+2. msgpack serialization does not allow embedding or referencing code that
+   is automatically executed. Incoming messages are unpacked by the msgpack
+   unpacker into native Python data structures (like tuples and dictionaries),
+   which are then passed to the rest of the program.
```
textshell: maybe mention why borg is not affected by 'object deserialization' bugs like yaml or pickle?

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+to forge the authentication.
+
+This effectively 'anchors' the manifest to the key, which is controlled by the
+client, thereby anchoring the entire DAG, making it impossible for an attacker
+to add, remove or modify any part of the DAG without Borg being able to detect
+the tampering.
+
+Note that when using BORG_PASSPHRASE the attacker cannot swap the *entire*
+repository against a new repository with eg. repokey mode and no passphrase,
+because Borg will abort access when BORG_PASSPRHASE is incorrect.
```
ThomasWaldmann: also it would notice if there is a change in encryption mode.

docs/security.rst
enkore: I was mostly thinking about, yeah, pickle and whatever Jenkins does to get that many remote-code execs in  a REST API.

Not sure how to make the point clearer here.

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+
+However, interactively a user might not notice this kind of attack
+immediately, if she assumes that the reason for the absent passphrase
+prompt is a set BORG_PASSPHRASE.
```
ThomasWaldmann: it would still notice that it previously was encrypted and now is not any more.

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+
+.. [#] The reason that the authentication tag is stored in the packed
```
ThomasWaldmann: The reason why...

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+For offline storage of the encryption keys they are encrypted with a
+user-chosen passphrase. A 256 bit key encryption key (KEK) is derived
+from the passphrase using PBKDF2-HMAC-SHA256 with a random 256 bit
+salt which is then used to Encrypt-then-MAC a packed
+representation of the keys with AES-256-CTR with a constant
+initialization vector of 0. The resulting MAC is stored alongside the
+ciphertext and converted to base64.
+
+This base64 blob (commonly referred to as *keyblob*)
+is then stored in the key file or in the repository (keyfile and
```
ThomasWaldmann: ...in the repository config ...

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+different derived KEK for every encryption due to the salt.
+
+.. _Horton principle: https://en.wikipedia.org/wiki/Horton_Principle
+.. _HKDF: https://tools.ietf.org/html/rfc5869
+.. _length extension: https://en.wikipedia.org/wiki/Length_extension_attack
+
+Remote RPC protocol security
+============================
+
+.. note:: This section could use expansion.
```
ThomasWaldmann: ... could be further expanded / detailled.

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+The assumed worst-case a server can inflict to a client is a
+denial of repository service.
+
+The situation were a server can
+create a general DoS on the client should be avoided, but
+might be possible by eg. forcing the client to allocate large
+amounts of memory to decode large messages. However, since we
+are not aware of any `"zip bomb"`_ equivalent in msgpack this
+would mean that the server has to send a similar amount of data to
+the client, which is not always practical. This is tracked with :issue:`2139`.
```
ThomasWaldmann: ehrm, there is a zip bomb: just say "there is a 2^32-1 sized array of ... coming now".
the client msgpack would allocate the memory first (before maybe noticing that not as much data is coming).
it's not recursive, though (like zip-in-zip).

docs/security.rst
```diff
@@ -0,0 +1,232 @@
...
+
+Borg is fundamentally based on an object graph structure (see :ref:`internals`),
+where the root object is called the manifest.
+
+Borg follows the `Horton principle`_, which states that
+not only the message must be authenticated, but also it's meaning (often
+expressed through context), because every object used is referenced by a
+parent object through it's object ID up to the manifest. The object ID in
+Borg is a MAC of the object's plaintext, therefore this ensures that
+an attacker cannot change the context of an object without forging the MAC.
```
enkore: No, context is meant here. Due to the described mechanism one cannot, say, exchange one item chunk for another.

docs/security.rst
enkore: since the ID of the exchanged repo would be assumed to be different, no, that one doesn't take.

docs/security.rst
enkore: > eg. repokey mode and no passphrase,

docs/security.rst
enkore: Hmm, you're right, even in use_list=False mode it seems to allocate... something... before building the tuple:

```
>>> import msgpack
>>> u = msgpack.Unpacker(use_list=False)
>>> u.feed(b'\xdd\x7f\xff\xff\xff')
>>> next(iter(u))
  File "msgpack/_unpacker.pyx", line 469, in msgpack._unpacker.Unpacker.__next__ (msgpack/_unpacker.cpp:5459)
  File "msgpack/_unpacker.pyx", line 400, in msgpack._unpacker.Unpacker._unpack (msgpack/_unpacker.cpp:4483)
MemoryError
```

I dunno how that even works, perhaps that *is* the actual tuple and it somehow (dirtily?) changes the values of the tuple after-the-fact or something like that.

ThomasWaldmann: we could add limits for everything else than binary data.
unless we can add limits for everything, this is a bit pointless though as an attacker could just choose the binary data type.

docs/security.rst
```diff
@@ -40,19 +40,19 @@ Borg is fundamentally based on an object graph structure (see :ref:`internals`),
...
 Borg is a MAC of the object's plaintext, therefore this ensures that
 an attacker cannot change the context of an object without forging the MAC.
 
 In other words, the object ID itself only authenticates the plaintext of the
-object and not it's context or meaning. The latter is established by a different
+object and not its context or meaning. The latter is established by a different
 object referring to an object ID, thereby assigning a particular meaning to
 an object. For example, an archive item contains a list of object IDs that
-represent packed file metadata. On their own it's not clear that these objects
+represent packed file metadata. On their own its not clear that these objects
```
ThomasWaldmann: it's :) (== it is) - also review / fix that globally.

docs/security.rst
ThomasWaldmann: i think the msgpack stuff is good enough for us, if someone is interested in msgpack, he can specifically look at that project. like he can look at ssh for ssh security properties...

docs/security.rst
enkore: So perhaps we want to make a fix here, that we get the same warning as with unknown unencrypted repos with repos that

- are unknown
- are repokey
- no passphrase supplied
- passphrase is empty

docs/security.rst
```diff
@@ -159,16 +167,20 @@ Borg cannot secure the key material while it is running, because the keys
...
+user-chosen passphrase.
+
+A 256 bit key encryption key (KEK) is derived from the passphrase
+using PBKDF2-HMAC-SHA256 with a random 256 bit salt which is then used
+to Encrypt-then-MAC a packed representation of the keys with
+AES-256-CTR with a constant initialization vector of 0 (this is the
+same construction used for Encryption_ with HMAC-SHA-256).
+
+The resulting MAC is stored alongside the ciphertext, which is
+converted to base64 in it's entirety.
```
ThomasWaldmann: its

docs/security.rst
```diff
@@ -186,9 +186,34 @@ This scheme, and specifically the use of a constant IV with the CTR
...
+- AES-CTR and HMAC-SHA-256 from OpenSSL 1.0 / 1.1 are used,
+  which is also linked into the static binaries we provide.
+  We think this is not an additional risk, since we don't ever
+  use OpenSSL's networking, TLS or X.509 code, but only their
+  primitives implemented in libcrypto.
+- SHA-256 and SHA-512 from Python's hashlib_ standard library module are used
+- HMAC, PBKDF2 and a constant-time comparison from Python's hmac_ standard
+  library module is used.
+- BLAKE2b is either provided by the system's libb2, an official implementation,
+  or a bunbled copy of the BLAKE2 reference implementation (written in C).
```
ThomasWaldmann: bundled

docs/security.rst
```diff
@@ -123,6 +123,15 @@ Depending on the chosen mode (see :ref:`borg_init`) different primitives are use
...
   and is also tracked locally on the client to avoid counter reuse.
+
+  .. rubric:: Future plans
+
+  Adding support for session keys is planned. This would do away with the
+  tricky business of tracking counter usage and avoiding counter reuse and
+  allows to use different encryption algorithms in the same repository,
+  eg. to make encryption feasible on weaker devices or diversify the
+  available algorithms.
+
```
ThomasWaldmann: I'ld only talk about existing code. In the branch where this document is located.

enkore: (from the commit msg)

> since they tie into multi-threading as well i think it's
ok to say this out loud that we plan it.

I think in a document that is (supposed to) go into some detail about security design and practices in the project it's okay and imho even expected to mention some perspectives where we _want_ to improve.

docs/security.rst
```diff
@@ -174,6 +174,13 @@ Decryption::
 
     ASSERT( CONSTANT-TIME-COMPARISON( chunk-id, AUTHENTICATOR(id_key, decompressed) ) )
 
+.. [#] Using the :ref:`borg key migrate-to-repokey <borg_key_migrate-to-repokey>`
+       command a user can convert repositories created using Attic in "passphrase"
+       mode to "repokey" mode. In this case the keys were directly derived from
+       the user's passphrase at some point using PBKDF2.
+
+       Borg does not support "passphrase" mode otherwise.
```
ThomasWaldmann: s/otherwise/any more/ ?

enkore: I'm not sure. Long-winded way of what I mean: Borg have an upgrade path for these repos, but aside from that, no support anymore. So... kinda both?

ThomasWaldmann: ... otherwise any more.

enkore: Nailed it.

docs/security.rst
```diff
@@ -0,0 +1,288 @@
...
+========
+
+Cryptography in Borg
+====================
+
+Attack model
+------------
+
+The attack model of Borg is that the environment of the client process
+(eg. ``borg create``) is trusted and the repository (server) is not. The
```
ThomasWaldmann: https://en.wiktionary.org/wiki/e.g. - so guess it is rather "e.g."

docs/security.rst
```diff
@@ -0,0 +1,288 @@
...
+3. recover plain-text data
+4. recover definite (heuristics based on access patterns are possible)
+   structural information such as the object graph (which archives
+   refer to what chunks)
+
+The attacker can always impose a denial of service per definition (he could
+forbid connections to the repository, or delete it entirely). The attacker
+can gauge how many archives exist in the repository due to the access pattern
+associated with creating archives, and because the manifest size gives this away
+(assuming approximately constant and known length of archive names).
```
ThomasWaldmann: why should we assume the attacker knows the length of archive names?

more than wild guesses (like: borg user maybe using something similar to the example script from the docs) are not possible here, imho.

assuming this here could leave the wrong impression that the attacker could somehow know them.

docs/security.rst
```diff
@@ -0,0 +1,288 @@
...
+the root node of a DAG through its edges, since the root node has no incoming edges.
+
+With the scheme as described so far an attacker could easily replace the manifest,
+therefore Borg includes a tertiary authentication mechanism (TAM) that is applied
+to the manifest since version 1.0.9 (see :ref:`tam_vuln`).
+
+TAM works by deriving a separate key through HKDF_ from the other encryption and
+authentication keys and calculating the HMAC of the metadata to authenticate [#]_::
+
+    salt = random(64)
```
ThomasWaldmann: hmm, is it clear enough that this means 64 Bytes and not 64bit or a random number from range 0..63?

docs/security.rst
ThomasWaldmann: ok.

docs/security.rst
ThomasWaldmann: yes, make an issue. needs investigation.

docs/security.rst
ThomasWaldmann: http://www.dict.cc/deutsch-englisch/K%C3%BCmmere+dich+nicht+um+ungelegte+Eier.html

docs/security.rst
enkore: removing, it's already covered by the preceding

> (heuristics based on access patterns are possible)

enkore:
Think I got everything

PR #2166, opened 2017-02-16
init: mandatory --encryption arg

### LINKED ISSUE #2165: borg init repo -> unencrypted repo
Only applies to master branch (borg 1.1beta).

When blake2b was introduced, we removed the default of aes256-hmac-sha256.

Then intention was that the user does an informed decision between hmac-sha256 and blake2b MACs.

But: if a user does not specify --encryption, we now fall back to no encryption at all (and this was not intended).

enkore:
Fixes #2165 and also showed some minor test issues (missing --encryption, handling of argparse errors "if not FORK_DEFAULT").

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -79,7 +79,13 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, **kw):
                 archiver = Archiver()
             archiver.prerun_checks = lambda *args: None
             archiver.exit_code = EXIT_SUCCESS
-            args = archiver.parse_args(list(args))
+            try:
+                args = archiver.parse_args(list(args))
+                # argparse parsing may raise SystemExit when the command line is bad or
+                # actions that abort early (eg. --help) where given. Catch this and return
+                # the error code as-if we invoked a Borg binary.
+            except SystemExit as e:
+                return e.code, output.getvalue()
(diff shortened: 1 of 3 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -79,7 +79,10 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, **kw):
                 archiver = Archiver()
             archiver.prerun_checks = lambda *args: None
             archiver.exit_code = EXIT_SUCCESS
-            args = archiver.parse_args(list(args))
+            try:
+                args = archiver.parse_args(list(args))
+            except SystemExit as e:
+                return e.code, output.getvalue()
```
ThomasWaldmann: add a comment what this is about?

src/borg/testsuite/archiver.py
```diff
@@ -1777,6 +1769,9 @@ def raise_eof(*args):
             self.cmd('init', '--encryption=repokey', self.repository_location, exit_code=1)
         assert not os.path.exists(self.repository_location)
 
+    def test_init_encryption(self):
+        self.cmd('init', self.repository_location, exit_code=2)
```
ThomasWaldmann: add a comment what this is about?

or rename to test_init_requires_encryption_option to make it clearer.

Issue #1802, opened 2016-11-04
docs table of contents is cluttered


anarcat:
first mentionned here: https://github.com/borgbackup/borg/pull/1798#issuecomment-258266833

> The support link is right before the "appendix"* of the manual (resources, changelog, internals, dev, license), and after the FAQ. Maybe add an FAQ entry pointing to it? That'd be one of the places I'd look. But maybe don't remove the security contact completely - leave a link in the README that says security issues should be reported via [url to docs].
>
> if someone knows how this can be more clearly structured that would be good I think. The ToC is a bit cluttered.

this is an attempt at identifying those issues more clearly and restructure the documentation.

right now, we have this structure overall structure:

 * Installation
 * Quick Start
 * Usage
 * Deployment
 * Frequently asked questions
 * Support
 * Resources
 * Changelog
 * Internals
 * Development
 * Borg Contributors (“The Borg Collective”)
 * License

That list is too long (12 items). It would be better to reduce that to a more reasonable number, like 7-8 items. Note that a bit of restructuring has been done in the PDF document in #1796 but the table of contents there is still way too long.

I would recommend the following structure instead:

* Installation
* Quick Start
* Usage
  * Deployment
  * Manuals
* Support
  * Frequently asked questions
  * Resources
  * Changelog
* Development
  * Internals
  * Borg Contributors (“The Borg Collective”)
  * License

Note how we move Deployment within Usage, the FAQ, Resources and Changelog within Support, and authors, licenses and internals into a development section.

I would also move a bunch of stuff from the quick start into the Usage section, hence the "Manuals" section that would hold the manpages separately.

If you look in the current detailed listing (I removed some subtitles for clarity):

 * Installation
   * Distribution Package
   * Standalone Binary
   * Features & platforms
   * From Source
 * Quick Start
   * A step by step example
   * Important note about free space
   * Automating backups
   * Pitfalls with shell variables and environment variables
   * Backup compression
   * Repository encryption
   * Remote repositories
 * Usage
   * General
   * borg init
   * [...] other manpages
   * Miscellaneous Help
   * Debug Commands
   * Additional Notes
 * Deployment
   * Machines
   * User and group
   * Folders
   * Restrictions
   * Client
   * Ansible
   * Salt
   * Enhancements
   * See also
 * Frequently asked questions
   * Can I backup VM disk images?
   * Can I backup from multiple servers into a single repository?
   [...]
 * Support
   * Issue Tracker
   * Chat (IRC)
[...]

PR #2167, opened 2017-02-16
docs: FAQ by categories as proposed by @anarcat in #1802

enkore:
docs: FAQ by categories as proposed by @anarcat in #1802

Usage & Limitations
Security
Common issues
Miscellaneous (only three items, two fork-related)

I immediately noticed that this makes it *much* easier to find FAQ items -- the list is just too long without any kind of hierarchy / order.

Note: This does not change/invalidate any links to FAQ items.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -5,6 +5,9 @@
 Frequently asked questions
 ==========================
 
+Usage & Limitations
+###################
+
 Can I backup VM disk images?
 ----------------------------
 
@@ -105,7 +108,6 @@ Are there other known limitations?
   An easy workaround is to create multiple archives with less items each.
   See also the :ref:`archive_limitation` and :issue:`1452`.
 
-
 Why is my backup bigger than with attic? Why doesn't |project_name| do compression by default?
 ----------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -93,7 +96,7 @@ Which file types, attributes, etc. are *not* preserved?
       backed up as (deduplicated and compressed) runs of zero bytes.
       Archive extraction has optional support to extract all-zero chunks as
       holes in a sparse file.
-    * filesystem specific attributes, like ext4 immutable bit, see :issue:`618`.
+    * Some filesystem specific attributes, like btrfs NOCOW, see :ref:`platforms`.
```
ThomasWaldmann: what was the problem with giving ext4 immutable bit as example and pointing to the issue?

docs/faq.rst
```diff
@@ -120,6 +122,70 @@ decision about whether you want to use compression, which algorithm
 and which level you want to use. This is why compression defaults to
 none.
 
+If a backup stops mid-way, does the already-backed-up data stay there?
+----------------------------------------------------------------------
+
+Yes, |project_name| supports resuming backups.
```
ThomasWaldmann: Reminds me I need to rephrase that. In fact, we never resume a backup, we always do full backups from the beginning.
It just feels like resuming...

docs/faq.rst
enkore: ext4 immutable has been supported for a while

docs/faq.rst
enkore: It's a bit like "Does Borg do incremental backups?" "Well, uh... it certainly feels like it! But <complex explanation>" :)

docs/faq.rst
ThomasWaldmann: oh, ok. :)

PR #2170, opened 2017-02-16
fuse: fix st_blocks to be an integer (not float) value

ThomasWaldmann:
it's basically ceil(dsize / blocksize) now (without doing floating point ops).

(fallout from #2157)

st_blocks is not expected to be a floating point value.
also: reduces execution time and keeps fuse code better in sync with (future) master.


enkore:
Why not just math.ceil?

ThomasWaldmann:
@enkore because that requires an import and floating point computations.

PR #2171, opened 2017-02-17
borg rpc: use limited msgpack.Unpacker

### LINKED ISSUE #2139: limit msgpack.Unpacker
The Unpacker in the rpc code (borg serve, RemoteRepository, see remote.py) currently operates without limits. Thus a evil client or evil server could send packs that unpack to huge amounts of data and thus cause a DoS via excessive memory consumption.

The danger by this is not very high though:

- attacking a server: borg operates via a ssh and a client usually needs a authorized ssh key or username/password to connect to a server. thus the server knows which credentials are used and could just revoke access for the evil client's account / the key.

- attacking a client: same thing: you know the server who is attacking you because you connected to it. also, if a server were hacked and wants that you can't use borg (to access backups, to create backups), there are easier ways to do that than rpc with evil data.

To fix this, we can set limits on the Unpacker (requires msgpack >= 0.4.6, but we already require that since quite a while anyway).

https://github.com/msgpack/msgpack-python/blob/0.4.6/msgpack/fallback.py#L143

But: currently we have a problem with the archive objects (pointed to via the archiveid): in the past, these could grow bigger than MAX_CHUNK_SIZE and old archive objects could exist without any specific limit in size - thus we can't define a limit for that kind of data. see #1473.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/41861104-limit-msgpack-unpacker)

ThomasWaldmann:
we do not trust the remote, so we are careful unpacking its responses.

the remote could return manipulated msgpack data that announces e.g.
a huge array or map or string. the local would then need to allocate huge
amounts of RAM in expectation of that data (no matter whether really
that much is coming or not).

by using limits in the Unpacker, a ValueError will be raised if unexpected
amounts of data shall get unpacked. memory DoS will be avoided.

Fixes #2139

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -23,7 +23,7 @@ from .helpers import sysinfo
 from .helpers import bin_to_hex
 from .helpers import replace_placeholders
 from .helpers import yes
-from .repository import Repository
+from .repository import Repository, MAX_OBJECT_SIZE, LIST_SCAN_LIMIT
 from .version import parse_version, format_version
 from .logger import create_logger
 
@@ -55,6 +55,27 @@ def os_write(fd, data):
         data = data[count:]
         time.sleep(count * 1e-09)
     return amount
+
+
+def get_limited_unpacker(kind):
+    """return a limited Unpacker because we should not trust msgpack data received from remote"""
+    args = dict(use_list=False,  # return tuples, not lists
+                max_bin_len=0,  # not used
+                max_ext_len=0,  # not used
+                max_buffer_size=3 * max(BUFSIZE, MAX_OBJECT_SIZE),
+                max_str_len=MAX_OBJECT_SIZE,  # a chunk or other repo object
+                )
+    if kind == 'server':
+        args.update(dict(max_array_len=100,  # misc. cmd tuples
+                         max_map_len=100,  # misc. cmd dicts
+                         ))
+    elif kind == 'client':
+        args.update(dict(max_array_len=LIST_SCAN_LIMIT,  # result list from repo.list() / .scan()
+                         max_map_len=100,  # misc. result dicts
+                         ))
+    else:
+        raise ValueError('kind must be "server" or "client"')
+    return msgpack.Unpacker(**args)
 
 
 class ConnectionClosed(Error):
@@ -185,7 +206,7 @@ class RepositoryServer:  # pragma: no cover
         # Make stderr blocking
         fl = fcntl.fcntl(stderr_fd, fcntl.F_GETFL)
         fcntl.fcntl(stderr_fd, fcntl.F_SETFL, fl & ~os.O_NONBLOCK)
-        unpacker = msgpack.Unpacker(use_list=False)
+        unpacker = get_limited_unpacker('server')
         while True:
             r, w, es = select.select([stdin_fd], [], [], 10)
             if r:
@@ -487,8 +508,7 @@ class RemoteRepository:
         self.ignore_responses = set()
         self.responses = {}
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
-
-        self.unpacker = msgpack.Unpacker(use_list=False)
+        self.unpacker = get_limited_unpacker('client')
         self.server_version = parse_version('1.0.8')  # fallback version if server is too old to send version information
         self.p = None
         testing = location.host == '__testsuite__'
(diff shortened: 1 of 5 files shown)


src/borg/remote.py
```diff
@@ -487,8 +487,15 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
         self.ignore_responses = set()
         self.responses = {}
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
-
-        self.unpacker = msgpack.Unpacker(use_list=False)
+        # we do not trust the server, use a limited Unpacker to avoid memory DoS via high element counts:
+        self.unpacker = msgpack.Unpacker(use_list=False,
+                                         max_buffer_size=3 * max(BUFSIZE, MAX_OBJECT_SIZE),
```
ThomasWaldmann: not really sure what a good max_buffer_size value should be.

src/borg/remote.py
```diff
@@ -487,8 +487,15 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
...
+                                         max_str_len=MAX_OBJECT_SIZE, # a chunk or other repo object
+                                         max_bin_len=0,  # not used
+                                         max_array_len=LIST_SCAN_LIMIT,  # repo.list() / .scan() id list result
+                                         max_map_len=100, # misc. result dicts
```
ThomasWaldmann: not sure about this one either.
i let it crash by using a too low value first and then "rounded up".

ThomasWaldmann:
Maybe add another commit for server operators that have to fear evil clients (== limit the server's rpc Unpacker in a similar way)?

src/borg/remote.py
```diff
@@ -57,6 +57,27 @@ def os_write(fd, data):
     return amount
 
 
+def get_limited_unpacker(kind):
+    """return a limited Unpacker because we should not trust msgpack data received from remote"""
+    args = dict(use_list=False,  # return tuples, not lists
+                max_bin_len=0,  # not used
+                max_ext_len=0,  # not used
+                max_buffer_size=3 * max(BUFSIZE, MAX_OBJECT_SIZE),
```
ThomasWaldmann: not sure about best value here.

src/borg/remote.py
```diff
@@ -57,6 +57,27 @@ def os_write(fd, data):
...
+                max_str_len=MAX_OBJECT_SIZE,  # a chunk or other repo object
+                )
+    if kind == 'server':
+        args.update(dict(max_array_len=100,  # misc. cmd tuples
+                         max_map_len=100,  # misc. cmd dicts
+                         ))
+    elif kind == 'client':
+        args.update(dict(max_array_len=LIST_SCAN_LIMIT,  # result list from repo.list() / .scan()
+                         max_map_len=100,  # misc. result dicts
+                         ))
```
ThomasWaldmann: not sure about these "100" either. experimented and "rounded up" what I saw.

ThomasWaldmann:
assigned backport/1.0-maint as it is a DoS protection. not that much interesting if you operate your own clients/servers, but could be interesting for borg backup space providers.

src/borg/remote.py
```diff
@@ -57,6 +57,27 @@ def os_write(fd, data):
...
 
+def get_limited_unpacker(kind):
+    """return a limited Unpacker because we should not trust msgpack data received from remote"""
+    args = dict(use_list=False,  # return tuples, not lists
+                max_bin_len=0,  # not used
+                max_ext_len=0,  # not used
+                max_buffer_size=3 * max(BUFSIZE, MAX_OBJECT_SIZE),
+                max_str_len=MAX_OBJECT_SIZE,  # a chunk or other repo object
+                )
+    if kind == 'server':
```
enkore: Rather make separate args, like so:

```py
def get_limited_unpacker(*, client=False, server=False):
  i = iter((client, server))
  any(i)
  assert not any(i), "Only one arg can be True!"
```

Or if combinatorial notation is preferred:

```py
assert (a ^ b) or not (a or b), "Either client or server can be true, but not both."
```

enkore:
Also update the security doc?

src/borg/remote.py
ThomasWaldmann: way more complex, why not keep it simple?

esp. if other kinds of unpackers are needed, the current approach does not need growing amounts of params and longer and longer assertion expressions.

enkore: didn't notice the else: ValueError before...

ThomasWaldmann:
@enkore done, updated security docs.

src/borg/remote.py
ThomasWaldmann: @enkore yes, I added that for clarity. But even without it, it would return None for unsupported values of "kind" (or typos), which would quickly fail when one tries to call that.

ThomasWaldmann:
ok, anything else in scope of "rpc unpacker"?

PR #2172, opened 2017-02-17
Add dsize and dcsize keys, fixes #2164

### LINKED ISSUE #2164: Add new key "dsize" for "borg list --format"
Hi, in the keys for the --format argument you have "size, csize, num_chunks and unique_chunks". Would it be difficult to have a "dsize" that indicates the amount of deduplicated size (and perhaps a cdsize, the compressed deduplicated size). That way, I know EXACTLY, where the 1,41GB of deduplicated (compressed) size of my archive are.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/42062772-add-new-key-dsize-for-borg-list-format)

Abogical:
These keys shows the amount of deduplicated size and compressed size of each file in the archive. As suggested in #2164.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@ Borg authors ("The Borg Collective")
 - Marian Beermann <public@enkore.de>
 - Daniel Reichelt <hacking@nachtgeist.net>
 - Lauri Niskanen <ape@ape3000.com>
+- Abdel-Rahman A. (Abogical)
 
 Borg is a fork of Attic.
 
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1481,8 +1481,8 @@ class ItemFormatter(BaseFormatter):
         self.call_keys = {
             'size': self.calculate_size,
             'csize': self.calculate_csize,
-            'dsize': self.calculate_dsize,
-            'dcsize': self.calculate_dcsize,
+            'dsize': partial(self.sum_unique_chunks_metadata, lambda chunk: chunk.size),
+            'dcsize': partial(self.sum_unique_chunks_metadata, lambda chunk: chunk.csize),
             'num_chunks': self.calculate_num_chunks,
             'unique_chunks': self.calculate_unique_chunks,
             'isomtime': partial(self.format_time, 'mtime'),
@@ -1531,6 +1531,20 @@ class ItemFormatter(BaseFormatter):
             item_data[key] = self.call_keys[key](item)
         return item_data
 
+    def sum_unique_chunks_metadata(self, metadata_func, item):
+        """
+        sum unique chunks metadata, a unique chunk is a chunk which is referenced globally as often as it is in the
+        item
+
+        item: The item to sum its unique chunks' metadata
+        metadata_func: A function that takes a parameter of type ChunkIndexEntry and returns a number, used to return
+                       the metadata needed from the chunk
+        """
+        chunk_index = self.archive.cache.chunks
+        chunks = item.get('chunks', [])
+        chunks_counter = Counter(c.id for c in chunks)
+        return sum(metadata_func(c) for c in chunks if chunk_index[c.id].refcount == chunks_counter[c.id])
+
     def calculate_num_chunks(self, item):
         return len(item.get('chunks', []))
 
@@ -1543,18 +1557,6 @@ class ItemFormatter(BaseFormatter):
 
     def calculate_csize(self, item):
         return sum(c.csize for c in item.get('chunks', []))
-
-    def calculate_dsize(self, item):
-        chunk_index = self.archive.cache.chunks
-        chunks = item.get('chunks', [])
-        chunks_counter = Counter(c.id for c in chunks)
-        return sum(c.size for c in chunks if chunk_index[c.id].refcount == chunks_counter[c.id])
-
-    def calculate_dcsize(self, item):
-        chunk_index = self.archive.cache.chunks
-        chunks = item.get('chunks', [])
-        chunks_counter = Counter(c.id for c in chunks)
-        return sum(c.csize for c in chunks if chunk_index[c.id].refcount == chunks_counter[c.id])
 
     def hash_item(self, hash_function, item):
         if 'chunks' not in item:
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -19,7 +19,7 @@ import time
 import unicodedata
 import uuid
 from binascii import hexlify
-from collections import namedtuple, deque, abc
+from collections import namedtuple, deque, abc, Counter
 from datetime import datetime, timezone, timedelta
 from fnmatch import translate
 from functools import wraps, partial, lru_cache
@@ -1546,11 +1546,15 @@ class ItemFormatter(BaseFormatter):
 
     def calculate_dsize(self, item):
         chunk_index = self.archive.cache.chunks
-        return sum(c.size for c in item.get('chunks', []) if chunk_index[c.id].refcount == 1)
+        chunks = item.get('chunks', [])
+        chunks_counter = Counter(c.id for c in chunks)
+        return sum(c.size for c in chunks if chunk_index[c.id].refcount == chunks_counter[c.id])
 
     def calculate_dcsize(self, item):
         chunk_index = self.archive.cache.chunks
-        return sum(c.csize for c in item.get('chunks', []) if chunk_index[c.id].refcount == 1)
+        chunks = item.get('chunks', [])
+        chunks_counter = Counter(c.id for c in chunks)
+        return sum(c.csize for c in chunks if chunk_index[c.id].refcount == chunks_counter[c.id])
 
     def hash_item(self, hash_function, item):
         if 'chunks' not in item:
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1484,7 +1484,7 @@ class ItemFormatter(BaseFormatter):
             'dsize': partial(self.sum_unique_chunks_metadata, lambda chunk: chunk.size),
             'dcsize': partial(self.sum_unique_chunks_metadata, lambda chunk: chunk.csize),
             'num_chunks': self.calculate_num_chunks,
-            'unique_chunks': self.calculate_unique_chunks,
+            'unique_chunks': partial(self.sum_unique_chunks_metadata, lambda chunk: 1),
             'isomtime': partial(self.format_time, 'mtime'),
             'isoctime': partial(self.format_time, 'ctime'),
             'isoatime': partial(self.format_time, 'atime'),
@@ -1547,10 +1547,6 @@ class ItemFormatter(BaseFormatter):
 
     def calculate_num_chunks(self, item):
         return len(item.get('chunks', []))
-
-    def calculate_unique_chunks(self, item):
-        chunk_index = self.archive.cache.chunks
-        return sum(1 for c in item.get('chunks', []) if chunk_index[c.id].refcount == 1)
 
     def calculate_size(self, item):
         return sum(c.size for c in item.get('chunks', []))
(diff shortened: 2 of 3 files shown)


Abogical:
I've checked, there isn't any

Abogical:
Unit tests added.

src/borg/testsuite/archiver.py
```diff
@@ -1298,9 +1298,12 @@ def test_list_size(self):
...
         test_archive = self.repository_location + '::test'
         self.cmd('create', '-C', 'lz4', test_archive, 'input')
-        output = self.cmd('list', '--format', '{size} {csize} {path}{NL}', test_archive)
-        size, csize, path = output.split("\n")[1].split(" ")
+        output = self.cmd('list', '--format', '{size} {csize} {dsize} {dcsize} {path}{NL}', test_archive)
+        size, csize, dsize, dcsize, path = output.split("\n")[1].split(" ")
         assert int(csize) < int(size)
+        assert int(dcsize) < int(dsize)
+        assert int(dsize) <= int(size)
+        assert int(dcsize) <= int(csize)
```
ThomasWaldmann: hmm, while thinking about the tests, I've found a problem with the semantics of d(c)size:

this shall indicate a "deduplicated size" (or, in other words, the size that this file adds to the repo via unique chunks).
but there is a problem with just looking at refcount == 1 chunks. there could be also refcount > 1 chunks that are **only** used within **this** file, so if we want to be totally correct, we would also have to add them to dsize.

ThomasWaldmann: the "deduplicated archive size" displayed at another place (not related to this code) needs also to be looked at.

in case this gets too complicated / too expensive to compute, we could also just document this issue.

Abogical: > there could be also refcount > 1 chunks that are only used within this file
I'm not completely familiar with the internals. When does that exactly happen?

ThomasWaldmann: just imagine an archive only has 1 file, a big file with multiple chunks of all zeros.
your code would say dsize == 0 (because there is no chunk with refcount 1).
but in fact dsize should be the size of 1 "all zeros" chunk.

(this is assuming that file size is a precise multiple of the zero chunk size. if that is not the case, the smaller all-zeros chunk at the end would have refcount == 1).

enkore: Psuedo/Python code

```py
counts = {chunk.id: sum(1 for c in chunks if c.id == id) for chunk in item.chunks}
dsize = 0
for chunk in item.chunks:
  if counts[chunk.id] == chunk_index[chunk.id].refcount:
    dsize += chunk.size  # or csize or so
```

ThomasWaldmann: @enkore looks correct, but nested loops are O(N^2).

ThomasWaldmann: btw, `collections.Counter`.

Abogical: > btw, `collections.Counter`

just thought about that...

enkore: > @enkore looks correct, but nested loops are O(N^2).

What nested loop?

enkore: Fixed

ThomasWaldmann: esp. if we want to do this not only for files, but also for archives, this gets expensive.

alejandro-perez:
I was just going to make a PR with the exact same code as Abogical :). Didn't know there was a bounty, thought :(

ThomasWaldmann:
@alejandro-perez oh, pity. double work should be avoided.

sometimes it help if one posts "i am working on this" to a github issues, so other people know.
and / or to announce on bountysource that one is working on the bounty.

src/borg/testsuite/archiver.py
alejandro-perez: So, trying to understand the issue for archives, would be something similar to this (thinking on the worst case scenario):
Have a big file (say 1GB video1.mp4), and for some reason I decide to copy it as video2.mp4 and then create an arvhice (note that the video had never been uploaded before). now, the entire 1GB is completely deduplicated (all the chunks have refcount>1) and thus, it would not be taken into account for the calculation of deduplicated size?

alejandro-perez:
I didn't get the bounty announce :). Nonetheless, glad it is done.

src/borg/testsuite/archiver.py
alejandro-perez: Nope, it does not fail (just try it). I need to re-read your error use case again :)

alejandro-perez: It does not fail either if I "cat video1.mp4 video2.mp4 > video-double.mp4". It calculates correctly the unique chunks (using identical random files of 160MiB)
```
This archive:              335.55 MB            335.55 MB            169.08 MB
All archives:              335.55 MB            335.55 MB            169.08 MB

Chunk index:                      75                  145
```

alejandro-perez: And this is what I get for a 500MiB of /dev/zero:
```
This archive:              524.29 MB            524.29 MB             12.59 MB
All archives:              524.29 MB            524.29 MB             12.59 MB

Chunk index:                       4                   65
```

alejandro-perez: However, as you envisioned, the dsize value (calculated as original @Abogical (and mine) code) fails (--format "{dsize:>10} {size:>10} {path}{NL}")
```
   4194304  524288000 TEST2.img
```

Abogical:
> sometimes it help if one posts "i am working on this" to a github issues, so other people know.
and / or to announce on bountysource that one is working on the bounty.

.... I should have done that :/

> I was just going to make a PR with the exact same code as Abogical :). Didn't know there was a bounty, thought :(

I'm sorry :(
However, if you want to, there are [other new bounties](https://www.bountysource.com/teams/borgbackup) borgbackup has just made that you could try and hunt them down.

enkore:
> I didn't get the bounty announce (and I created the issue). Damm github!
 Nonetheless, glad it is done. 

I believe you need to separately subscribe to notifications on BountySource to get these emails.

src/borg/testsuite/archiver.py
enkore: The two snippets you posted above are from borg create --stats / borg info which correctly consider these cases by design

alejandro-perez: Great then.

alejandro-perez: A question about the tests... is compression working that good always? Isn't there situations where csize is == (or even >) than size? Can you assert that?

Abogical:
> By the way, @Abogical feel free to add yourself to AUTHORS ;)
Thanks 😄

src/borg/testsuite/archiver.py
Abogical: @alejandro-perez Your right, csize may be more than size, but we're testing this on a fixed repository. Where we know csize must be less than size.

Abogical:
Checks complete. You can review and/or merge.

src/borg/helpers.py
```diff
@@ -1540,6 +1544,18 @@ def calculate_size(self, item):
     def calculate_csize(self, item):
         return sum(c.csize for c in item.get('chunks', []))
 
+    def calculate_dsize(self, item):
+        chunk_index = self.archive.cache.chunks
+        chunks = item.get('chunks', [])
+        chunks_counter = Counter(c.id for c in chunks)
```
enkore: Can you also put this new/corrected logic into calculate_unique_chunks above? That way the definition of a unique chunk would be consistent for all of these.

Abogical: Will do, I can also refactor the duplicate code generated.

src/borg/helpers.py
```diff
@@ -1527,12 +1531,21 @@ def get_item_data(self, item):
...
 
-    def calculate_num_chunks(self, item):
-        return len(item.get('chunks', []))
+    def sum_unique_chunks_property(self, property_func, item):
+        """
+        sum unique chunks properties, a unique chunk is a chunk which is referenced globally as often as it is in the
+        item
 
-    def calculate_unique_chunks(self, item):
+        item: The item to sum it's unique chunks
```
ThomasWaldmann: "its"

as this piece of grammar seems to be not too well known:

it is very easy to differentiate between "its" and "it's".

"it's" is a contraction for "it is", so just try if "it is" fits in the sentence.
if not, you wanted "its" (which is a possessive pronoun, like "his", "her").

Related:

http://www.apostrophecatastrophes.com/

http://www.angryflower.com/247.html
(here we see where the confusion comes from: possessive form for nouns is with apostrophe, but the possessive pronoun "its" is without ...)

src/borg/helpers.py
```diff
@@ -1527,12 +1531,21 @@ def get_item_data(self, item):
...
+        property_func: A function that takes a chunk parameter to return the chunk's desired property to sum
```
ThomasWaldmann: rather say chunk id, iirc we also have a Chunk type.

src/borg/helpers.py
ThomasWaldmann: @enkore ^^^ ;-)

src/borg/helpers.py
```diff
@@ -1479,8 +1481,10 @@ def __init__(self, archive, format):
         self.call_keys = {
             'size': self.calculate_size,
             'csize': self.calculate_csize,
+            'dsize': partial(self.sum_unique_chunks_property, lambda chunk: chunk.size),
+            'dcsize': partial(self.sum_unique_chunks_property, lambda chunk: chunk.csize),
             'num_chunks': self.calculate_num_chunks,
-            'unique_chunks': self.calculate_unique_chunks,
+            'unique_chunks': partial(self.sum_unique_chunks_property, lambda chunk: 1),
```
ThomasWaldmann: nice! :)

Abogical:
Comments rewritten

Abogical:
Comments rewritten. Renamed `sum_unique_chunks_property` to `sum_unique_chunks_metadata`, as it seemed more appropriate.

src/borg/helpers.py
enkore: >  (here we see where the confusion comes from: possessive form for nouns is with apostrophe, but the possessive pronoun "its" is without ...)

I wondered why I kept doing this specific mistake -- I don't think I make any other error as frequently as this one, normally I even get their/they're/there right. This is an interesting explanation. Not only does its/it's just not exist in my first language, but it's also kind of inconsistent.

Abogical: tbh putting apostrophe makes the word look better :)

Abogical:
Checks complete. Ready to review

ThomasWaldmann:
hmm, i can't easily see what you changed.

Abogical:
View it commit by commit, idk why the diff turns out to be so bad.

ThomasWaldmann:
There is no commit from 6h ago, guess you merged it already into your old commits.

Abogical:
I've done a rebase

AUTHORS
```diff
@@ -10,6 +10,7 @@ Borg authors ("The Borg Collective")
 - Marian Beermann <public@enkore.de>
 - Daniel Reichelt <hacking@nachtgeist.net>
 - Lauri Niskanen <ape@ape3000.com>
+- Abdel-Rahman A. (Abogical)
```
ThomasWaldmann: I guess your commits contain some email address, would you add that here also?
Do you prefer your family name to be abbreviated?

Abogical: 1. I may chnage my email in the future
2. my family name is my first name, making it an ugly name, so I choose to hide it.

ThomasWaldmann: 1. always true for everyone
2. ah, thanks for explaining.

Abogical:
Merge it, I won't add my email.

enkore:
Don't see a reason why we'd need an email there.

I for only put mine in because it's absolutely impossible to get even more spam on one address ;)

enkore:
Thanks!

Abogical:
Thank you! :)

PR #2182, opened 2017-02-19
WIP borg create: make --compression required, fixes #2179

### LINKED ISSUE #2179: remove compression mode default?
following the same reasoning why we do not have a encryption mode default any more in borg 1.1, we should remove the compression mode default and thus require the user to give `--compression X` for `borg create`.

Having `--compression none` as default (as now) is similarly bad as having `--encryption none` as default (as in early borg/attic).

ThomasWaldmann:
while this strongly points people using borg create to choose whatever compression they want, it might be a bit more uncomfortable than e.g. requiring --encryption for borg init. create is used way more often than init.

otoh, one should create scripts. aliases exist.


enkore:
src/borg/testsuite/archiver.py:1442:27: E241 multiple spaces after ','

ThomasWaldmann:
closing this, change of goals. will open a new one that makes lz4 the default.

PR #2188, opened 2017-02-20
add extra test for the hashindex

rciorba:
Insert a few keys, delete some of them, check we still have
the values we expect, check the deleted ones aren't there.

### DIFF
--- a/src/borg/testsuite/hashindex.py
+++ b/src/borg/testsuite/hashindex.py
@@ -126,28 +126,34 @@ class HashIndexTestCase(BaseTestCase):
 
 
 class HashIndexExtraTestCase(BaseTestCase):
-    """These tests are separate because they should not become part of the selftest
+    """These tests are separate because they should not become part of the selftest.
     """
     def test_chunk_indexer(self):
-        # see _hashindex.c hash_sizes, we want to be close to the max fill rate
-        # because interesting errors happen there
-        max_key = int(65537 * 0.75) - 10
-        index = ChunkIndex(max_key)
-        deleted_keys = [
-            hashlib.sha256(H(k)).digest()
-            for k in range(-1, -(max_key//3), -1)]
-        keys = [hashlib.sha256(H(k)).digest() for k in range(max_key)]
+        # see _hashindex.c hash_sizes, we want to be close to the max. load
+        # because interesting errors happen there.
+        key_count = int(65537 * ChunkIndex.MAX_LOAD_FACTOR) - 10
+        index = ChunkIndex(key_count)
+        all_keys = [hashlib.sha256(H(k)).digest() for k in range(key_count)]
+        # we're gonna delete 1/3 of all_keys, so let's split them 2/3 and 1/3:
+        keys, to_delete_keys = all_keys[0:(2*key_count//3)], all_keys[(2*key_count//3):]
+
         for i, key in enumerate(keys):
             index[key] = (i, i, i)
-        for i, key in enumerate(deleted_keys):
+        for i, key in enumerate(to_delete_keys):
             index[key] = (i, i, i)
 
-        for key in deleted_keys:
+        for key in to_delete_keys:
             del index[key]
         for i, key in enumerate(keys):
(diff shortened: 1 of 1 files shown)


src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,31 @@ def test_chunkindex_summarize(self):
         assert unique_chunks == 3
 
 
+class HashIndexExtraTestCase(BaseTestCase):
+    """These tests are separate because they should not become part of the selftest
+    """
+    def test_chunk_indexer(self):
+        # see _hashindex.c hash_sizes, we want to be close to the max fill rate
+        # because interesting errors happen there
+        max_key = int(65537 * 0.75) - 10
```
ThomasWaldmann: import/use MAX_LOAD_FACTOR (or so)?

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,31 @@ def test_chunkindex_summarize(self):
         assert unique_chunks == 3
 
 
+class HashIndexExtraTestCase(BaseTestCase):
+    """These tests are separate because they should not become part of the selftest
+    """
+    def test_chunk_indexer(self):
+        # see _hashindex.c hash_sizes, we want to be close to the max fill rate
```
ThomasWaldmann: s/max fill rate/max. load/

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,31 @@ def test_chunkindex_summarize(self):
...
+        # because interesting errors happen there
```
ThomasWaldmann: nitpick: add dot at end of sentence. also in 129.

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,31 @@ def test_chunkindex_summarize(self):
...
+        max_key = int(65537 * 0.75) - 10
+        index = ChunkIndex(max_key)
+        deleted_keys = [
+            hashlib.sha256(H(k)).digest()
+            for k in range(-1, -(max_key//3), -1)]
```
ThomasWaldmann: if we are still < 120, put it in 1 line? no 80 chars pep8 limit.

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,31 @@ def test_chunkindex_summarize(self):
...
+        for i, key in enumerate(deleted_keys):
+            index[key] = (i, i, i)
+
+        for key in deleted_keys:
+            del index[key]
+        for i, key in enumerate(keys):
+            assert index[key] == (i, i, i)
+        for key in deleted_keys:
+            assert index.get(key) is None
+
```
ThomasWaldmann: now delete all keys also and make sure all are gone?

src/borg/testsuite/hashindex.py
rciorba: How can I do that? As far as i know it's a C macro, and there is (still) no way to reach it.

src/borg/testsuite/hashindex.py
rciorba: I don't understand what you mean. 

Do you mean add a comment explaining what's going on? I assumed the code is simple and self explanatory.

src/borg/testsuite/hashindex.py
ThomasWaldmann: iirc we made it available via cython and already use it elsewhere.

src/borg/testsuite/hashindex.py
ThomasWaldmann: you are only deleting `deleted_keys` to test insert / delete behaviour with full hashtable / growing hashtable (/ shrinking hashtable).

one could also delete keys from `keys` again, just to have some more tests.

ThomasWaldmann:
@rciorba can we finish this soon to get it off the table? seems like only some simple changes are still todo.

ThomasWaldmann:
thanks for updating, looks good. \o/

I'll wait for travis and merge then.

PR #2192, opened 2017-02-21
Edited + and corrected some stuff in internals+security

enkore:
*docs: make internals.rst an index page* because it's strange in the rtd theme how the sidebar displays that, also results in a strange arrangement in the PDF (first the sub-toctrees, then jumps back to the main doc, with no obvious way of telling that these subheadings don't belong to the preceding docs - very confusing). This way, much nicer, also a couple introductory paragraphs. If we add more pages there (*cough*) it also gives a better overview.


docs/internals.rst
```diff
@@ -88,61 +77,93 @@ identifier for repositories. It will not change if you move the
...
+
+Compaction processes sparse segments from oldest to newest; sparse segments
+which don't contain enough deleted data to justify compaction are skipped. This
+avoids doing e.g. 500 MB of writing current data to a new segment when only
+a couple kB were deleted in a segment.
+
+Segments that are compacted are read in entirety. Current entries are written to
+a new segment, while superseded entries are omitted. After each segment an intermediary
+commit is written to the new segment, data is synced and the old segment is deleted --
+freeing disk space.
```
enkore: save space is ignored

docs/internals.rst
```diff
@@ -1,552 +1,32 @@
...
+specified when the backup was performed.
 
-|project_name| stores its data in a `Repository`. Each repository can
-hold multiple `Archives`, which represent individual backups that
-contain a full archive of the files specified when the backup was
-performed. Deduplication is performed across multiple backups, both on
-data and metadata, using `Chunks` created by the chunker using the Buzhash_
+Deduplication is performed globally across all data in the repository
+(multiple backups and even multiple hosts), both on data and metadata,
+using :ref:`chunks` created by the chunker using the Buzhash_
```
enkore: I guess that's technically correct. But also technically the manifest can't be deduplicated and archives aren't deduplicated since their object ID == archive ID.

I'm not sure whether this has to be encoded into this sentence?

enkore:
Most just moved in e5bbba5

Issue #2185, opened 2017-02-19
network speed regression in 1.1.0b3


jdchristensen:
I started to use borg 1.1.0b3, and noticed that it was unusually slow
when network bound.  So I did some benchmarks comparing various versions
of borg, as well as attic 0.16.

In most cases I did a single backup of a 112MB directory into a fresh
repo.  Sometimes I followed this with a backup that included the same
directory (unchanged) and a second directory of similar size.

Client is a laptop, connected via a solid wifi connection to a home
router, and then over the internet to a remote machine.  Outbound
network can sustain ~500kB/s over ssh.  Tests were repeated multiple times
to rule out network variations, and the results were always consistent.
I can't ping the server, but interactive usage is snappy.

Remote disk is a very fast M.2 SSD raid array.  Remote cpu is very fast,
local is decent.  Local drive is an SSD, but files were in cache
anyways, as shown by the find commands below.

No encryption for any repos.  Default chunker parameters.

I sometimes varied the compression method, as shown.

All borg versions were run using the pre-compiled 64-bit binaries.

First a summary table, and then selected logs.  The table shows the
create time and the deduplicated size.  It some cases, it shows a
followup backup.

```
  borg 1.1.0b3 zlib:    4m55s  76.0MB ; 3m10s 77.7MB  [cpu usage < 10%]
  borg 1.1.0b2 zlib:    2m27s  76.0MB ; 2m30s 77.7MB  [cpu usage < 10%]
  borg 1.0.10 zlib:     2m21s  76.0MB                 [cpu usage < 10%]
  attic 0.16 (zlib):    2m26s  75.6MB ; 2m25s 74.8MB  [cpu usage < 10%]

  borg 1.1.0b3 lzma,6: 10m30s  74.9MB                 [cpu usage 20-60%]
  borg 1.1.0b3 none:    5m35s 102.4MB                 [cpu usage < 10%]
```

Note that attic and the older borg versions fully utilize the network: 76/147 = 0.5MB/s.

I'm happy to test other versions as well.

Dan

```
BORG 1.1.0b3:

$ borg init -e none bu@host:test.borg
$ find dir1 dir2 -type f -exec cat {} \; > /dev/null
$ borg create -C zlib -s bu@host:test.borg::1 dir1
------------------------------------------------------------------------------
Archive name: 1
Archive fingerprint: d27ba2f333f713f21903f02051de77e456f36b5f527e75c870ae14e982842014
[...]

PR #2194, opened 2017-02-22
In api decorator, pass wait argument to RemoteRepository.call; fixes …

jdchristensen:
…#2185

Some more independent timing tests would be good, in case there are other independent regressions.

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -425,12 +425,16 @@ def api(*, since, **kwargs_decorator):
         def do_rpc(self, *args, **kwargs):
             sig = inspect.signature(f)
             bound_args = sig.bind(self, *args, **kwargs)
-            named = {}
+            named = {}  # Arguments for the remote process
+            extra = {}  # Arguments for the local process
             for name, param in sig.parameters.items():
                 if name == 'self':
                     continue
                 if name in bound_args.arguments:
-                    named[name] = bound_args.arguments[name]
+                    if name == 'wait':
+                        extra[name] = bound_args.arguments[name]
+                    else:
+                        named[name] = bound_args.arguments[name]
                 else:
                     if param.default is not param.empty:
                         named[name] = param.default
@@ -447,7 +451,7 @@ def api(*, since, **kwargs_decorator):
                 raise self.RPCServerOutdated("{0} {1}={2!s}".format(f.__name__, name, named[name]),
                                              format_version(restriction['since']))
 
-            return self.call(f.__name__, named)
+            return self.call(f.__name__, named, **extra)
         return do_rpc
     return decorator
 


ThomasWaldmann:
Did 2 measurements, both looked similar to this one:
```
$ git checkout 1.1.0b2
$ . speedtest.sh 

borg 1.1.0b2
Elapsed (wall clock) time (h:mm:ss or m:ss): 1:02.87

$ git checkout test-fix-remote-jdc
$ . speedtest.sh 

borg 1.1.0b4.dev115+ng6d8a538
Elapsed (wall clock) time (h:mm:ss or m:ss): 1:02.90
```
So, looks like @jdchristensen fixed the speed issue. \o/

@textshell can you review the PR please?

textshell:
Nice find jdchristensen.

Sorry, i won’t be able to fully review this very quickly.
But ignoring the wait=False argument used for all put/delete from the cache (i.e. the main data path) is certainly a bug. This PR looks like it fixes that. And the speed measurements seem to agree.
From a quick grep and look this patch seems fine. But as usual with the RPC code more eyes are definitively not an bad idea.

textshell:
One additional thought. It would be nice to have a special mode for some of our tests that notices if the code ignores any of the named parameters. Just to avoid bugs like this in the future. But nothing that needs fixing in this PR.

jdchristensen:
While we are looking at this, are there any other "meta" arguments that are meant for the local code not the remote code?  (That's what is really happening here.)  The patch is designed to make the list easy to extend, if needed, but it's also fragile in the sense that if an api call is added that has `wait` as a valid argument to the remote process, it will break.  If there are others, we could choose a naming scheme that keeps them separate.

enkore:
@ThomasWaldmann 

> $ git checkout test-fix-remote-jdc

Trick 17:

```
$ cat .git/config
... stuff ...
[remote "origin"]
        url = http://github.com/borgbackup/borg
        fetch = +refs/heads/*:refs/remotes/origin/*
        fetch = +refs/pull/*/head:refs/remotes/origin/pr/*
... more stuff ...
```

The second `fetch` line allows access to all PRs in the repository by eg. `git checkout pr/2194`.

jdchristensen:
I added a couple of comments, but no code changes.  I'll squash them later.

ThomasWaldmann:
@enkore nice trick. one first needs to do "git fetch origin" then, but after that it works. \o/

jdchristensen:
I squashed the two commits into one, and think this is ready to merge. If we find that further arguments need the same treatment, it will be easy to extend what I wrote. People touching this code just need to be aware of this issue, and the comments in the new code point out the issue.

enkore:
Thanks!

PR #2197, opened 2017-02-22
info: show utilization of maximum archive size

enkore:
See #1452

"This is 100 % accurate."

Also increases maximum data size by ~41 bytes. Not 100 % side-effect free;
if you manage to exactly land in that area then older Borg would not read
it. OTOH it gives us a nice round number there. No hard opinion here.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -350,10 +350,13 @@ Archive fingerprint: {0.fpr}
 Time (start): {start}
 Time (end):   {end}
 Duration: {0.duration}
-Number of files: {0.stats.nfiles}'''.format(
+Number of files: {0.stats.nfiles}
+Utilization of max. archive size: {csize_max:.0%}
+'''.format(
             self,
             start=format_time(to_localtime(self.start.replace(tzinfo=timezone.utc))),
-            end=format_time(to_localtime(self.end.replace(tzinfo=timezone.utc))))
+            end=format_time(to_localtime(self.end.replace(tzinfo=timezone.utc))),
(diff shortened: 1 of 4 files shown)


src/borg/archive.py
```diff
@@ -350,10 +350,13 @@ def __str__(self):
 Time (start): {start}
 Time (end):   {end}
 Duration: {0.duration}
-Number of files: {0.stats.nfiles}'''.format(
+Number of files: {0.stats.nfiles}
+Utilization of max. archive size: {csize_max:.0%}
```
ThomasWaldmann: Use also a percentage as you do below?

src/borg/archiver.py
```diff
@@ -987,6 +987,7 @@ def format_cmdline(cmdline):
             print('Duration: %s' % archive.duration_from_meta)
             print('Number of files: %d' % stats.nfiles)
             print('Command line: %s' % format_cmdline(archive.metadata.cmdline))
+            print('Utilization of max. archive size: %d %%' % (100 * cache.chunks[archive.id].csize / MAX_DATA_SIZE))
```
ThomasWaldmann: "99%" rather than "99 %"?

src/borg/archive.py
enkore: This is percentage formatting

ThomasWaldmann: oh %-)

src/borg/archiver.py
enkore: fixed

PR #2198, opened 2017-02-22
Handle big extended attributes. Fixes #2161

### LINKED ISSUE #2161: Argument list too long error probably due to xattrs
Reported and discussed this in IRC. Data which was originally on Windows was transferred to OSX, and from there sent to a Linux Ubuntu 14.04 machine with ext4 filesystem where the errors happened upon typing borg extract backup::<repo_name>

See log on pastebin: http://rn0.me/show/QOGX3zmqQDB2f8AMb3dD/

Seems like it has something to do with the icon files; checked the xattrs there. Note this doesn't happen when running it on OSX.

xattrs:
```
ls -l -@ google_drive/Archive/Something/Something_Finance/Monthly\ financials/Icon?
-rw-r--r--@ 1 cluelesscoder  staff  0 Feb 14 12:48 google_drive/Archive/Something/Something_Finance/Monthly financials/Icon?
	com.apple.FinderInfo	32
	com.apple.ResourceFork	340647
```

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/42034922-argument-list-too-long-error-probably-due-to-xattrs)

Abogical:
This will print a warning whenever extended attributes in the archive are too large to extract, and continues the extraction of the item instead of halting. Fixes #2161.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -621,7 +621,10 @@ Number of files: {0.stats.nfiles}'''.format(
             try:
                 xattr.setxattr(fd or path, k, v, follow_symlinks=False)
             except OSError as e:
-                if e.errno not in (errno.ENOTSUP, errno.EACCES):
+                if e.errno == errno.E2BIG:
+                    logger.warning('%s: Value or key of extended attribute %s is too big for this filesystem' %
+                                   (path, k.decode()))
+                elif e.errno not in (errno.ENOTSUP, errno.EACCES):
                     # only raise if the errno is not on our ignore list:
                     # ENOTSUP == xattrs not supported here
                     # EACCES == permission denied to set this specific xattr
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -1029,6 +1029,21 @@ class ArchiverTestCase(ArchiverTestCaseBase):
                 self.cmd('extract', self.repository_location + '::test')
             assert xattr.getxattr('input/file', 'security.capability') == capabilities
 
+    @pytest.mark.skipif(not xattr.XATTR_FAKEROOT, reason='xattr not supported on this system or on this version of'
+                                                         'fakeroot')
+    def test_extract_big_xattrs(self):
+        def patched_setxattr(*args, **kwargs):
+            raise OSError(errno.E2BIG, 'E2BIG')
+        self.create_regular_file('file')
+        xattr.setxattr('input/file', 'attribute', 'value')
+        self.cmd('init', self.repository_location, '-e' 'none')
+        self.cmd('create', self.repository_location + '::test', 'input')
+        with changedir('output'):
+            with patch.object(xattr, 'setxattr', patched_setxattr):
+                out = self.cmd('extract', self.repository_location + '::test', exit_code=EXIT_WARNING)
+                assert out == (os.path.abspath('input/file') + ': Value or key of extended attribute attribute is too big'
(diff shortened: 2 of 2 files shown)


src/borg/archive.py
```diff
@@ -474,6 +475,7 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
             if has_damaged_chunks:
                 logger.warning('File %s has damaged (all-zero) chunks. Try running borg check --repair.' %
                                remove_surrogates(item.path))
+                warning = True
```
ThomasWaldmann: doesn't look like having any effect.

src/borg/archive.py
```diff
@@ -530,21 +532,22 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
...
+                    warning = self.restore_attrs(path, item, fd=fd.fileno()) or warning
             if has_damaged_chunks:
                 logger.warning('File %s has damaged (all-zero) chunks. Try running borg check --repair.' %
                                remove_surrogates(item.path))
+                warning = True
             if hardlink_masters:
                 # Update master entry with extracted file path, so that following hardlinks don't extract twice.
                 hardlink_masters[item.get('source') or original_path] = (None, path)
-            return
+            return warning
```
ThomasWaldmann: check all places that return to actually return the boolean (not None).

src/borg/archiver.py
```diff
@@ -587,6 +590,8 @@ def peek_and_store_hardlink_masters(item, matched):
         if pi:
             # clear progress output
             pi.finish()
+        if warning:
+            return EXIT_WARNING
         return self.exit_code
```
ThomasWaldmann: hmm, looks like there are 3 ways now to deal with warnings/errors/exceptions:
- raise an exception
- set self.exit_code
- return warning boolean

enkore:
If we really wanna go with a warning RC for this we should perhaps come up with a better way?

Abogical:
> There is a `warnings` module in the stdlib, but that feels a bit too big. Add separate logger method to our logging proxy in the logging module, that sets an appropriate flag.

Seems to be the best solution. Since we're solving a different issue, I'll create a seperate PR for this.

ThomasWaldmann:
@enkore what's the problem with a warning rc? We could not extract all data, so a warning rc is appropriate.

src/borg/archive.py
```diff
@@ -621,7 +621,10 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
             try:
                 xattr.setxattr(fd or path, k, v, follow_symlinks=False)
             except OSError as e:
-                if e.errno not in (errno.ENOTSUP, errno.EACCES):
+                if e.errno == errno.E2BIG:
+                    logger.warning('Value or key of extended attribute %s in file %s is too big for this filesystem' %
```
enkore: One last nitpick - then squash + merge.

IIRC most other warnings are formatted like:

```
<filename>: something bad
```

something like -> `%s: [xattr/extended attribute] %s is too big [to extract here / for this filesystem]`

Abogical:
I should've mentioned, I was waiting for #2199 to return a warning code for these situations. The testsuite added is skipped since travis doesn't support xattrs. **That testsuite tests that the return code is EXIT_WARNING, so it will fail for local tests.**

ThomasWaldmann:
ok, so we should finish that one asap. :)

PR #2199, opened 2017-02-23
Add return code functions

Abogical:
Related #2198

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -34,7 +34,7 @@ from .cache import Cache
 from .constants import *  # NOQA
 from .crc32 import crc32
 from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
-from .helpers import Error, NoManifestError
+from .helpers import Error, NoManifestError, set_ec
 from .helpers import location_validator, archivename_validator, ChunkerParams, CompressionSpec
 from .helpers import PrefixSpec, SortBySpec, HUMAN_SORT_KEYS
 from .helpers import BaseFormatter, ItemFormatter, ArchiveFormatter, format_time, format_file_size, format_archive
@@ -3125,7 +3125,7 @@ class Archiver:
         self.prerun_checks(logger)
         if is_slow_msgpack():
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
-        return func(args)
+        return set_ec(func(args))
 
 
 def sig_info_handler(sig_no, stack):  # pragma: no cover
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -51,10 +51,21 @@ def Chunk(data, **meta):
     return _Chunk(meta, data)
 
 
+'''
+The global exit_code variable is used so that modules other than archiver can increase the program exit code if a
+warning or error occured during their operation. This is different from archiver.exit_code, which is only accessible
+from the archiver object.
+'''
 exit_code = EXIT_SUCCESS
 
 
 def set_ec(ec):
+    '''
+    Sets the exit code of the program, if an exit code higher or equal than this is set, this does nothing. This
+    makes EXIT_ERROR override EXIT_WARNING, etc..
+
+    ec: exit code to set
+    '''
     global exit_code
     exit_code = max(exit_code, ec)
     return exit_code
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -51,6 +51,15 @@ def Chunk(data, **meta):
     return _Chunk(meta, data)
 
 
+exit_code = EXIT_SUCCESS
+
+
+def set_ec(ec):
+    global exit_code
+    exit_code = max(exit_code, ec)
+    return exit_code
+
+
 class Error(Exception):
     """Error base class"""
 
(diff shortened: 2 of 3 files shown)


src/borg/logger.py
```diff
@@ -172,6 +175,7 @@ def info(self, *args, **kw):
             return self.__logger.info(*args, **kw)
 
         def warning(self, *args, **kw):
+            warning_printed = True
             return self.__logger.warning(*args, **kw)
```
enkore: We're a bit "abusive" in using `warning` and it's often used just for important (should see by default) messages, but those shouldn't change the return code.

Create a new method that does this, say something like `ẃarn[ing]_with_rc` that does this, and leave warning to be just the "important message" thing?

src/borg/archiver.py
```diff
@@ -3125,7 +3125,10 @@ def run(self, args):
         self.prerun_checks(logger)
         if is_slow_msgpack():
             logger.warning("Using a pure-python msgpack! This will result in lower performance.")
-        return func(args)
+        exit_code = func(args)
+        if is_warning_printed():
+            return EXIT_WARNING
+        return exit_code
```
enkore: if exit_code is already EXIT_ERROR, but a  warning was printed, then it would become EXIT_WARNING.

src/borg/archiver.py
```diff
@@ -20,7 +20,7 @@
 from datetime import datetime
 from itertools import zip_longest
 
-from .logger import create_logger, setup_logging
+from .logger import create_logger, setup_logging, is_warning_printedr
```
enkore: There's an r too much here ;)

Abogical:
- Since setting the return code is irrelevant to logger, I've moved the changes to helpers.py
- I've generalized the function, you could set any exit code other than EXIT_WARNING, the final return code will be the one with the highest set. thus EXIT_ERROR overrides EXIT_WARNING, etc..

src/borg/helpers.py
```diff
@@ -51,6 +51,22 @@ def Chunk(data, **meta):
...
+def set_rc(rc):
+    global return_code
+    return_code = max(return_code, rc)
+    return return_code
+
+
+def init_rc():
+    global return_code
+    return_code = EXIT_SUCCESS
+    return return_code
```
ThomasWaldmann: rather just have it as a global definition on module level?

Abogical: I've tried to simply state:
`return_code = EXIT_SUCCESS`
but this fails when tested with the testsuite. I think this may be because the return_code does not reset after each test.

enkore: Yes, if it's a global definition then you'd have to reset it manually in the test suite, exec_cmd (or so) in testsuite/archiver.py does that for the Archiver.exit_code attribute.

Abogical: Ok, thanks @enkore

ThomasWaldmann: oh, oops. ok, keep it as is and maybe put a comment about that into that init fn.

src/borg/testsuite/archiver.py
```diff
@@ -79,6 +79,7 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, **kw):
                 archiver = Archiver()
             archiver.prerun_checks = lambda *args: None
             archiver.exit_code = EXIT_SUCCESS
+            helpers.return_code = EXIT_SUCCESS
```
ThomasWaldmann: hmm, guess there is no special reason why return_code global should live in helpers module while exit_code lives in archiver module.

so, how about: have both in archiver and make naming clearer (or add comments what they are for).

src/borg/archiver.py
```diff
@@ -34,7 +34,7 @@
 from .constants import *  # NOQA
 from .crc32 import crc32
 from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
-from .helpers import Error, NoManifestError
+from .helpers import Error, NoManifestError, get_rc, set_rc
```
ThomasWaldmann: do you actually need get_rc?

src/borg/helpers.py
```diff
@@ -51,6 +51,19 @@ def Chunk(data, **meta):
...
+
+
+def get_rc():
+    return return_code
+
+
+def set_rc(rc):
+    global return_code
+    return_code = max(return_code, rc)
+    return return_code
```
ThomasWaldmann: if everything else moves to archiver, move set_rc there also.

src/borg/testsuite/archiver.py
Abogical: It can't be in archiver. If you recall at #2198, the reason we did this is that we somehow wanted to tweak the return code from another class method (`Archive.restore_attsr()`). We can't do that if we don't have access to the archiver object. So unless archiver is not global, or is not added to the args of the Archive, this can't be done.

ThomasWaldmann: oh, i didn't notice archiver is an Archiver instance here.

what I meant was archiver module.

Abogical: > make naming clearer
This should've been named exit_code, will do that.

Abogical: > EDIT: also using archiver from the module archive.py will create a cyclic dependency.

Abogical:
Checks passed with a weird macOS failure.

src/borg/helpers.py
```diff
@@ -51,6 +51,15 @@ def Chunk(data, **meta):
     return _Chunk(meta, data)
 
 
+exit_code = EXIT_SUCCESS
```
ThomasWaldmann: can you put a comment above this, explaining about it and archiver.exit_code?

the rest lgtm.

Abogical:
Added comments...

PR #2203, opened 2017-02-23
JSON output for major commands

### LINKED ISSUE #2200: --json for create, info, list
See also #2024 #654

enkore:
Fixes #2200 

This is the data output, going out over stdout. JSON logging as proposed in #2024 would go out over stderr as usual, so no conflicts there.

There is some overlap between info (with --last/--first/--prefix) and list which is more obvious in the JSON output. The difference is that "info" reports more stuff, and takes longer for that, while "list" is fast (only looks at the manifest).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -977,8 +977,19 @@ class Archiver:
             format = "{archive:<36} {time} [{id}]{NL}"
         formatter = ArchiveFormatter(format)
 
+        output_data = []
+
         for archive_info in manifest.archives.list_considering(args):
-            write(safe_encode(formatter.format_item(archive_info)))
+            if args.json:
+                output_data.append(formatter.get_item_data(archive_info))
+            else:
+                write(safe_encode(formatter.format_item(archive_info)))
+
+        if args.json:
+            print_as_json({
+                'repository': manifest.repository,
+                'archives': output_data,
+            })
 
         return self.exit_code
 
@@ -2492,6 +2503,8 @@ class Archiver:
         subparser.add_argument('--format', '--list-format', dest='format', type=str,
                                help="""specify format for file listing
                                 (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NL}")""")
+        subparser.add_argument('--json', action='store_true',
+                               help='format output as JSON')
         subparser.add_argument('location', metavar='REPOSITORY_OR_ARCHIVE', nargs='?', default='',
                                type=location_validator(),
                                help='repository/archive to list contents of')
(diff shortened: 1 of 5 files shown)


src/borg/archiver.py
```diff
@@ -2424,6 +2471,10 @@ def process_epilog(epilog):
         subparser.add_argument('--format', '--list-format', dest='format', type=str,
                                help="""specify format for file listing
                                 (default: "{mode} {user:6} {group:6} {size:8d} {isomtime} {path}{extra}{NL}")""")
+        subparser.add_argument('--json', action='store_true',
+                               help='format output as JSON. The form of --format is ignored, but keys used in it '
+                                    'are added to the JSON output. Some keys are always present. Note: JSON can only '
+                                    'represent text. A "bpath" key is therefore not available.')
```
enkore: Other commands don't have that, so less consistent.

What could make more sense imho is to have --format just list (as in `name,sha256,mtime`) the keys wanted, and then only give back these keys.

src/borg/helpers.py
```diff
@@ -2058,3 +2094,50 @@ def swidth_slice(string, max_width):
     if reverse:
         result.reverse()
     return ''.join(result)
+
+
+class BorgJsonEncoder(json.JSONEncoder):
+    def default(self, o):
```
enkore: o is what the superclass defines. I usually stick with that.

src/borg/archiver.py
```diff
@@ -973,45 +992,71 @@ def format_cmdline(cmdline):
...
-            print(DASHES)
-            print(STATS_HEADER)
-            print(str(stats))
-            print(str(cache))
+            if args.json:
+                output_data.append(archive.info())
+            else:
+                stats = archive.calc_stats(cache)
+                print('Archive name: %s' % archive.name)
+                print('Archive fingerprint: %s' % archive.fpr)
```
enkore: This is a bit more difficult, because human formatting differs from JSON formatting. So if one uses the dict used for JSON, then one has to reformat keys after the fact or not really use the dict for everything, so I didn't.

src/borg/helpers.py
ThomasWaldmann: oh. bad stdlib code. nobody expected the spanish inquisition.

enkore: It's old, bytes where precious :)

enkore:
I'm just gonna merge this for now, only my

> What could make more sense imho is to have --format just list (as in name,sha256,mtime) the keys wanted, and then only give back these keys.

comment is left, but that's actually a bit awkward to do since it pulls the defaults from the cmd parser down to ItemFormatter, which is messier than I'd like here.

Another thing we could do is to try to use any of the faster JSON implementations optionally, there are a few of them (optional dependency).

ThomasWaldmann:
we can still optimize speed later, if there is actual demand for it. rather not add dependencies now.

PR #2207, opened 2017-02-24
UNFINISHED emit clean error msg when encountering attic data, fixes #1933

### LINKED ISSUE #1933: Generic error message when trying to use non-upgraded attic repository
Hi,

while playing with borg (as attic user) i accidentally run borg commands on attic*s repository and i get ugly trackback:
```
  File "/usr/lib64/python3.4/site-packages/borg/archiver.py", line 1928, in main
  File "/usr/lib64/python3.4/site-packages/borg/archiver.py", line 1873, in run
    return args.func(args)
  File "/usr/lib64/python3.4/site-packages/borg/archiver.py", line 81, in wrapper
  File "/usr/lib64/python3.4/site-packages/borg/helpers.py", line 113, in load
    cdata = repository.get(cls.MANIFEST_ID)
  File "/usr/lib64/python3.4/site-packages/borg/repository.py", line 512, in get
  File "/usr/lib64/python3.4/site-packages/borg/repository.py", line 152, in get_transaction_id
    raise self.CheckNeeded(self.path)
borg.repository.CheckNeeded: /srv/backup/system.borg
```
IMO, this exception have to be catched and only error message needs to be reported, to get more better look ;-)

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/39795380-generic-error-message-when-trying-to-use-non-upgraded-attic-repository)

ThomasWaldmann:


ThomasWaldmann:
hmm, this seems a bit more complicated than I thought. When using a different exception (AtticDataError) when encountering data from attic, some other stuff breaks. Even when using same base class at some places it is either "break some stuff" or "have attic cleanly reported". Partly due to some rough error handling (ok vs. not ok) at lower levels.

enkore:
Superseded by #2920

ThomasWaldmann:
Hmm, wondering what go me into troubles back then (but don't precisely remember what the troubles were).

Maybe it was me trying to keep behaviour much as it was before (so attic being detected still behaved like an integrity error, instead of just immediately terminating with an error).

Issue #2178, opened 2017-02-18
document snapshot usage


ThomasWaldmann:
some systems create snapshots at varying pathes.

concerning borg's filecache, this is unfortunate as there will be always a cache miss when going from one snapshot's files backup to the next.

so, we should point out that the snapshot mount point should be always the same (if not otherwise possible, mount --bind could be used).

ThomasWaldmann:
or rather point out that the **absolute** path of src files (and inode number and mtime and size) must not change for speedy files-cache-enabled backups.

enkore:
Fixed by #2218

ThomasWaldmann:
@kmq thanks for the docs update.

btw, did you see the docs bounties?

kmq:
@ThomasWaldmann  just in passing until you mentioned them, [but it seems I might be up for $30](https://github.com/borgbackup/borg/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Akmq%20)

ThomasWaldmann:
@kmq just refer to your PRs in one of the bounty tickets.

PR #2218, opened 2017-02-26
document snapshot usage #2178

kmq:

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -429,7 +429,7 @@ BORG_FILES_CACHE_TTL to at least 26 (or maybe even a small multiple of that),
 it would be much faster.
 
 Another possible reason is that files don't always have the same path, for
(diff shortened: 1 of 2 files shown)


enkore:
Thanks!

docs/faq.rst
```diff
@@ -429,7 +429,7 @@ BORG_FILES_CACHE_TTL to at least 26 (or maybe even a small multiple of that),
 it would be much faster.
 
 Another possible reason is that files don't always have the same path, for
-example if you mount a filesystem without stable mount points for each backup.
+example if you mount a filesystem without stable mount points for each backup or if you are running the backup from a filesystem snapshot whose name is not stable.
```
ThomasWaldmann: in case someone else is wondering about "whose" here, it is correct:

http://dict.leo.org/forum/viewGeneraldiscussion.php?idThread=1086474&idForum=4&lang=de&lp=ende

PR #2219, opened 2017-02-26
document JSON

enkore:


docs/internals/frontends.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Borg does on the other hand provide an API on a command-line level. In other words, a frontend should to
+(for example) create a backup archive just invoke :ref:`borg_create`.
+
+Logging
+-------
+
+Especially for graphical frontends it is important to be able to convey and reformat progress information
+in meaningful ways. The ``--log-json`` option turns the stderr stream of Borg into a stream of JSON lines,
+where each line is a JSON object. The *type* key of the object determines its other contents.
```
ThomasWaldmann: somehow a (non) stability assertion is missing here - is it expected to change?

if so, would a json output of `borg --version` be useful?

enkore: We haven't defined that yet anywhere, although there is a bit of a consensus. We should write that down some time...

enkore:
Merging this for now, defining a somewhat-precise compatibility policy for the CLI and the JSON output (needs to be both, otherwise pointless) is a separate topic.

I still have 1-2 changes to do here (JSON progress indicators), will update docs together with functional changes.

PR #2220, opened 2017-02-26
use lz4 compression by default, fixes #2179

### LINKED ISSUE #2179: remove compression mode default?

ThomasWaldmann:
not for create_src_archive() though as this triggers a bug that has to get fixed outside this PR first.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -108,19 +108,23 @@ Are there other known limitations?
   An easy workaround is to create multiple archives with less items each.
   See also the :ref:`archive_limitation` and :issue:`1452`.
 
-Why is my backup bigger than with attic? Why doesn't |project_name| do compression by default?
-----------------------------------------------------------------------------------------------
+Why is my backup bigger than with attic?
+----------------------------------------
 
 Attic was rather unflexible when it comes to compression, it always
 compressed using zlib level 6 (no way to switch compression off or
 adjust the level or algorithm).
 
-|project_name| offers a lot of different compression algorithms and
-levels. Which of them is the best for you pretty much depends on your
-use case, your data, your hardware -- so you need to do an informed
-decision about whether you want to use compression, which algorithm
-and which level you want to use. This is why compression defaults to
-none.
+The default in Borg is lz4, which is fast enough to not use significant CPU time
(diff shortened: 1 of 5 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -262,7 +262,7 @@ def cmd(self, *args, **kw):
         return output
 
     def create_src_archive(self, name):
-        self.cmd('create', self.repository_location + '::' + name, src_dir)
+        self.cmd('create', '--compression=none', self.repository_location + '::' + name, src_dir)
```
ThomasWaldmann: remove this change to trigger bug.

enkore:
- quickstart.rst Backup compression needs an update (mentions the default)
- example in usage.rst needs update: # use zlib compression (good, but slow) - default is no compression
- FAQ "Why is my backup bigger than with attic? Why doesn’t Borg do compression by default?"

src/borg/testsuite/archiver.py
enkore: #2221

docs/faq.rst
```diff
@@ -119,8 +119,11 @@ adjust the level or algorithm).
 levels. Which of them is the best for you pretty much depends on your
 use case, your data, your hardware -- so you need to do an informed
 decision about whether you want to use compression, which algorithm
-and which level you want to use. This is why compression defaults to
-none.
+and which level you want to use.
```
ThomasWaldmann: the previous sentence is also somehow related to have a "none" default.
instead we should say lz4 is default because it is always better than none,
but that if one needs a better compression, one needs to specify it.

docs/faq.rst
```diff
@@ -119,8 +119,11 @@ adjust the level or algorithm).
...
+
+The default is LZ4, which is fast enough to not use significant CPU time
```
ThomasWaldmann: as our option value is "lz4", maybe rather write it like that. (same below at other places)

PR #2224, opened 2017-02-27
Fix compression exceptions

ThomasWaldmann:


src/borg/testsuite/archiver.py
```diff
@@ -262,7 +262,7 @@ def cmd(self, *args, **kw):
         return output
 
     def create_src_archive(self, name):
-        self.cmd('create', '--compression=none', self.repository_location + '::' + name, src_dir)
+        self.cmd('create', '--compression=lz4', self.repository_location + '::' + name, src_dir)
```
ThomasWaldmann: lz4 is the default. I did the change so I can comment on it:

when I put --compression=zlib here, another strangeness appears: the test intended to corrupt the chunk by appending b'1234' is not effective. likely because the zlib decompressor does not care for appended data and just returns the expected, correct data.

ThomasWaldmann:
no 1.0-maint backport needed, code is new in master/1.1.

ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/builds/205650628 wat?

enkore:
I'm not sure what's up with Travis that we see the race condition or whatever causes it so much more frequently there, because I see that far, far far far less often. Like, once in a month or two, on two very different machines.

src/borg/compress.pyx
```diff
@@ -167,7 +170,10 @@ class ZLIB(CompressorBase):
 
     def decompress(self, data):
         # note: for compatibility no super call, do not strip ID bytes
-        return zlib.decompress(data)
+        try:
+            return zlib.decompress(data)
+        except zlib.error as e:
+            raise DecompressionError(str(e)) from None
```
enkore: This doesn't work correctly, the docstring / "excstring" of DecompressionError needs a placeholder, like so:

`"Failed to decompress chunk: {}"`

src/borg/compress.pyx
```diff
@@ -112,7 +112,7 @@ class LZ4(CompressorBase):
                 break
             if osize > 2 ** 30:
                 # this is insane, get out of here
-                raise Exception('lz4 decompress failed')
+                raise DecompressionError('lz4 decompress failed')
```
enkore: Hmm so in case of corruption this would grow the buffer to >huge before aborting. Related https://github.com/lz4/lz4/issues/156

enkore:
> but maybe we need IntegrityError at other places?

I think some code in 1.0 check is supposed to work around these errors, e.g. regenerating the manifest.

src/borg/compress.pyx
ThomasWaldmann: yes, error signalling from hell. :(

src/borg/compress.pyx
ThomasWaldmann: fixed.

ThomasWaldmann:
ok, anything else besides collapsing / backporting?

PR #2229, opened 2017-02-27
JSON progress indicators

enkore:
This should pretty much kill that area off.

Last area I see left is to handle yes()

### DIFF
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -96,6 +96,39 @@ log_message
     :ref:`msgid <msgid>`
         Message ID, may be *none* or absent
 
+.. rubric:: Examples (reformatted, each object would be on exactly one line)
+
+:ref:`borg_extract` progress::
+
+    {"message": "100.0% Extracting: src/borgbackup.egg-info/entry_points.txt",
+     "current": 13000228, "total": 13004993, "info": ["src/borgbackup.egg-info/entry_points.txt"],
+     "operation": 1, "msgid": "extract", "type": "progress_percent", "finished": false}
+    {"message": "100.0% Extracting: src/borgbackup.egg-info/SOURCES.txt",
+     "current": 13004993, "total": 13004993, "info": ["src/borgbackup.egg-info/SOURCES.txt"],
+     "operation": 1, "msgid": "extract", "type": "progress_percent", "finished": false}
+    {"operation": 1, "msgid": "extract", "type": "progress_percent", "finished": true}
+
+:ref:`borg_create` file listing with progress::
+
+    {"original_size": 0, "compressed_size": 0, "deduplicated_size": 0, "nfiles": 0, "type": "archive_progress", "path": "src"}
+    {"type": "file_status", "status": "U", "path": "src/borgbackup.egg-info/entry_points.txt"}
+    {"type": "file_status", "status": "U", "path": "src/borgbackup.egg-info/SOURCES.txt"}
+    {"type": "file_status", "status": "d", "path": "src/borgbackup.egg-info"}
+    {"type": "file_status", "status": "d", "path": "src"}
+    {"original_size": 13176040, "compressed_size": 11386863, "deduplicated_size": 503, "nfiles": 277, "type": "archive_progress", "path": ""}
+
+Internal transaction progress::
+
+    {"message": "Saving files cache", "operation": 2, "msgid": "cache.commit", "type": "progress_message", "finished": false}
+    {"message": "Saving cache config", "operation": 2, "msgid": "cache.commit", "type": "progress_message", "finished": false}
+    {"message": "Saving chunks cache", "operation": 2, "msgid": "cache.commit", "type": "progress_message", "finished": false}
+    {"operation": 2, "msgid": "cache.commit", "type": "progress_message", "finished": true}
+
+A debug log message::
+
+    {"message": "35 self tests completed in 0.08 seconds",
+     "type": "log_message", "created": 1488278449.5575905, "levelname": "DEBUG", "name": "borg.archiver"}
+
 Standard output
 ---------------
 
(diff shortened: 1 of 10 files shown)


src/borg/helpers.py
```diff
@@ -1404,12 +1417,31 @@ def __init__(self):
             if self.logger.level == logging.NOTSET:
                 self.logger.setLevel(logging.WARN)
             self.logger.propagate = False
+        self.emit = self.logger.getEffectiveLevel() == logging.INFO
```
ThomasWaldmann: can you explain why this is correct?

enkore: the propagation for the progress logger works so:

- if --progress is _not_ set, then the setup_implied_logging code does nothing. So the immediate logger level is NOTSET, this is checked in the line above, and sets the immediate level to WARN
- if --progress _is_ set, then setup_implied_logging code sets INFO immediate level on the progress logger
- note that the progress logger does _not_ propagate, ie. getEffectiveLevel() == .level -- I could've written it directly that way
- progress logging happens at INFO level
- so if the logger level is INFO, then --progress is turned on, otherwise it is not.

enkore: perhaps we should document the logging setup and propagation to various parts of the system (progress, remote, implied logging) better => logger.py docstring.

docs/internals/frontends.rst
```diff
@@ -39,6 +40,39 @@ archive_progress
     path
         Current path
 
+progress_message
+    A message-based progress information with no concrete progress information, just a message
+    saying what is currently worked on.
```
ThomasWaldmann: ... being worked on. ?

docs/internals/frontends.rst
```diff
@@ -54,11 +88,46 @@ log_message
...
         Formatted log message
+    :ref:`msgid <msgid>`
+        Message ID, may be *none* or absent
+
+.. rubric:: Examples (reformatted, each object would be on exactly one line)
+
+:ref:`borg_extract` progress::
+
+    {"message": "100.0% Extracting: src/borgbackup.egg-info/entry_points.txt",
+     "current": 13000228, "total": 13004993, "info": ["src/borgbackup.egg-info/entry_points.txt"],
```
ThomasWaldmann: hmm, if we have the percentage value in the message, but not as a separate value, the client would need to compute and format it on its own. depending on how it does that, the final string could differ from the string in our message (e.g. when using different precision for computation or result formatting).

so, shall we put the percentage value into json, too?

docs/internals/frontends.rst
```diff
@@ -105,19 +174,43 @@ stats
...
 :ref:`borg_list` uses a simpler format that is faster to retrieve. Either return archives in an
 array under the *archives* key, while :ref:`borg_create` returns a single archive object under the
 *archive* key.
 
-Both formats contain a *name* key with the archive name, and the *id* key with the hexadecimal archive ID.
+Both formats contain a *name* key with the archive name, the *id* key with the hexadecimal archive ID,
+and the *start* key with the start timestamp.
 
- info and create further have:
+info and create further have:
```
ThomasWaldmann: either highlight "info" and "create" somehow or say "borg info and borg create"?

docs/internals/frontends.rst
```diff
@@ -152,8 +245,222 @@ username
 comment
     Archive comment, if any
 
+Example of a simple archive listing (``borg list --last 1 --json``)::
+
+    {
+        "archives": [
+            {
+                "archive": "2017-02-27T21:21:51",
```
ThomasWaldmann: this might be a bit confusing if you archive name just happens to be a datetime string.
maybe include example host and/or dataset name to make it clearer.

docs/internals/frontends.rst
```diff
@@ -152,8 +245,222 @@ username
...
+                "barchive": "2017-02-27T21:21:51",
```
ThomasWaldmann: hmm? this is either the same as name or (if it would be non-utf8 binary somehow) you could not include it anyway?

docs/internals/frontends.rst
```diff
@@ -152,8 +245,222 @@ username
...
+                "id": "80cd07219ad725b3c5f665c1dcf119435c4dee1647a560ecac30f8d40221a46a",
+                "name": "2017-02-27T21:21:51",
```
ThomasWaldmann: difference to "archive"?

docs/internals/frontends.rst
```diff
@@ -152,8 +245,222 @@ username
...
+                    "/home/user/repository",
+                    "..."
+                ],
+                "comment": "",
+                "duration": 5.641542,
+                "end": "Mon, 2017-02-27 21:21:58",
+                "hostname": "host",
+                "id": "80cd07219ad725b3c5f665c1dcf119435c4dee1647a560ecac30f8d40221a46a",
+                "limits": {
+                    "max_archive_size": 0.0001330855110409714
```
ThomasWaldmann: hmm, strange names here.

archive_load (see load [factor] of hash tables)?
archive_full (like 100% full == 1.0)?

docs/internals/frontends.rst
```diff
@@ -152,8 +245,222 @@ username
...
+    {
+        "encryption": {
+            "mode": "repokey"
+        },
+        "repository": {
+            "id": "0cbe6166b46627fd26b97f8831e2ca97584280a46714ef84d2b668daf8271a23",
+            "last_modified": "Mon, 2017-02-27 21:21:58",
+            "location": "/home/user/repository"
+        },
+        "files": [
```
ThomasWaldmann: maybe "items" instead of "files". or "fs_items"?

src/borg/helpers.py
ThomasWaldmann: yeah, some more docs would be good. :)

src/borg/helpers.py
```diff
@@ -1404,12 +1417,31 @@ def __init__(self):
...
+        assert self.json
+        if not self.emit:
+            return
+        kwargs.update(dict(
+            operation=self.id,
+            msgid=self.msgid,
+            type=self.JSON_TYPE,
+            finished=finished
+        ))
+        print(json.dumps(kwargs), file=sys.stderr)
```
ThomasWaldmann: hmm, is this making the assumption that output shall always be on stderr?
the logging system can log to many different outputs.

src/borg/helpers.py
```diff
@@ -1585,6 +1622,7 @@ def get_item_data(self, archive):
             'archive': remove_surrogates(archive.name),
             'id': bin_to_hex(archive.id),
             'time': format_time(to_localtime(archive.ts)),
+            'start': format_time(to_localtime(archive.ts)),
```
ThomasWaldmann: wondering about why it needs same value as in "time".
also, if there is start, where is end?

src/borg/logger.py
```diff
@@ -165,24 +165,38 @@ def setLevel(self, *args, **kw):
             return self.__logger.setLevel(*args, **kw)
 
         def log(self, *args, **kw):
+            if 'msgid' in kw:
+                kw.setdefault('extra', {})['msgid'] = kw.pop('msgid')
```
ThomasWaldmann: you duplicated that for performance reasons? if not, a function would be nice.

src/borg/remote.py
```diff
@@ -636,6 +636,12 @@ def borg_cmd(self, args, testing):
                 opts.append('--critical')
             else:
                 raise ValueError('log level missing, fix this code')
+            try:
+                borg_logger = logging.getLogger('borg')
+                if borg_logger.json:
+                    opts.append('--log-json')
```
ThomasWaldmann: so, you give --log-json to the remote borg serve, but in logger:95, you do not use the JsonFormatter if it is in the borg serve process? so, what is it good for?

src/borg/helpers.py
enkore: the docs/spec for JSON output says it always goes to stderr.

docs/internals/frontends.rst
enkore: No, that's what the msgid is for. We _could_ provide the format string, but I think that's less useful, because then the formatting for humans becomes part of the API; ugly. This way we have msgid, which is easy to keep fixed, and *message* which is "don't assume anything" sort of deal: show it verbatim somewhere, or ignore it, but don't try to parse it.

docs/internals/frontends.rst
enkore: ^- these three:

archive/barchive is the usual deal we have with POSIX (like path/bpath), just pointless for JSON since we don't get non-text anyway. We could do something more invasive here, like base64, but I don't see much point for now. Note that archive/barchive/time are not documented, only start is. The former three come from ArchiveFormatter which is used for the regular listings, but we use start here, since that makes the formats supersets of each other: list-archive-format < create-archive-format < info-archive-format

The latter two could be made the same with some more fiddling; the difference only lies in what the caller knows anyway though.

docs/internals/frontends.rst
enkore: archive['limits']['max_archive_size'] -> 0.00...

Seems alright to me

src/borg/logger.py
enkore: Adding a function and pasting one line to all of these versus pasting two lines to all of these -- not much difference imho.

src/borg/remote.py
enkore: It sets the `json` attribute on the Borg logger in the serve process, which would be used to correctly use the progress indicator in check/replay/commit. However, remote --progress was broken some time ago with IIRC implied logging setup. So that has to be revisited anyway.

enkore: Also the entire way of how we configure our logging extensions should maybe redone  -- see comments above regarding `self.emit`.

src/borg/logger.py
ThomasWaldmann: not line-wise, but complexity wise. it would be just f(kw).

enkore: either is pretty simple... just that for f(kw) I'd have to go back and use git-rebase... ;)

PR #2230, opened 2017-02-27
robin hood hashing

rciorba:
supersedes https://github.com/borgbackup/borg/pull/1429

current benchmarks: https://cdn.rawgit.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/7b439639eadc0cbb1c258ba2edbd3cd206cd078b/results.html

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -35,6 +35,17 @@
 #define MAGIC "BORG_IDX"
 #define MAGIC_LEN 8
 
+#define DEBUG 0
+
+#define debug_print(fmt, ...)                   \
+  do {                                          \
+    if (DEBUG) {                                \
+      fprintf(stderr, fmt, __VA_ARGS__);        \
+      fflush(NULL);                             \
+    }                                           \
+  } while (0)
+
+
 typedef struct {
     char magic[MAGIC_LEN];
     int32_t num_entries;
@@ -52,6 +63,7 @@ typedef struct {
     off_t bucket_size;
     int lower_limit;
     int upper_limit;
+    void *tmp_entry;
 } HashIndex;
 
 /* prime (or w/ big prime factors) hash table sizes
@@ -76,7 +88,9 @@ static int hash_sizes[] = {
 };
 
 #define HASH_MIN_LOAD .25
-#define HASH_MAX_LOAD .75  /* don't go higher than 0.75, otherwise performance severely suffers! */
+#define HASH_MAX_LOAD .99  /* use testsuite.benchmark.test_chunk_indexer_* to find
+                              an appropriate value; also don't forget to update this
+                              value in archive.py */
 
 #define MAX(x, y) ((x) > (y) ? (x): (y))
 #define NELEMS(x) (sizeof(x) / sizeof((x)[0]))
@@ -84,7 +98,7 @@ static int hash_sizes[] = {
 #define EMPTY _htole32(0xffffffff)
 #define DELETED _htole32(0xfffffffe)
 
-#define BUCKET_ADDR(index, idx) (index->buckets + (idx * index->bucket_size))
+#define BUCKET_ADDR(index, idx) (index->buckets + ((idx) * index->bucket_size))
 
 #define BUCKET_MATCHES_KEY(index, idx, key) (memcmp(key, BUCKET_ADDR(index, idx), index->key_size) == 0)
 
@@ -92,7 +106,7 @@ static int hash_sizes[] = {
 #define BUCKET_IS_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == EMPTY)
 
 #define BUCKET_MARK_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) = DELETED)
-#define BUCKET_MARK_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) = EMPTY)
+#define BUCKET_MARK_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, (idx)) + index->key_size)) = EMPTY)
 
 #define EPRINTF_MSG(msg, ...) fprintf(stderr, "hashindex: " msg "\n", ##__VA_ARGS__)
 #define EPRINTF_MSG_PATH(path, msg, ...) fprintf(stderr, "hashindex: %s: " msg "\n", path, ##__VA_ARGS__)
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
@rciorba seen that?: https://probablydance.com/2017/02/26/i-wrote-the-fastest-hashtable/

rciorba:
@ThomasWaldmann thanks for the link.
His approach is very interesting, but one expensive computation stands out as being eliminated entirely: distance_from_desired is stored in memory, not recomputed.

Tis means an extra 1 byte used. In Borg's case that would take us from 44 bytes (32 key + 12 val) to 45. The downside is even worse cache alignment (although the current one isn't great either) and more memory used at all sizes.

So for reference resizing to 8410991 from 6007867, means a 40% increase in memory usage, while 157223263 from 144378011 is only a 9% hike. The increase in size by going from 44 bytes to 45 is only 2%, which does suggest memory usage is not a big deal.

rciorba:
On the powers of 2 vs prime numbers, it would make sense for borg to use powers of 2, because we alreasy have uniform distribution of keys, but that limits us to doubling the hash each time. [As you can see here our current size progression](https://cdn.rawgit.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/7bac9a7d18259fce3130c224e38b9fe50b332dae/sizes.svg) our rate of increase is reduced as the hash gets larger and larger, so sadly powers of 2 are probably not a good fit for us (despite the massive savings by ditching the modulo operation).

rciorba:
Some other potential differences between others hashmap usecases and borg's that could explain why we're not seeing massive performance gains, is the size of our buckets. The example in the paper cites 8 byte elements, so that means 64 / 8 = 8 buckets in one cache line. While for borg we can't even guarantee one cache line will hold one element because of the large size and alignment.

This is mostly speculation on my part, since I don't understand very well how much cache pre-fetching helps ether case (because at least access is linear so pre-fetching can do it's job). I did attempt to split out the keys and values in different memory locations to get 2 keys to map neatly in a cache line, but I didn't see an immediate improvement and stopped that line of pursuit due to limited time and the fact the change is very invasive.

rciorba:
Also, on the actual PR: I did find an issue with how skip_hint is reported by hashindex_lookup that will affect the setitem performance, and am working on fixing that.

ThomasWaldmann:
We could even consider:
- add 4 bytes, to keep 32bit alignment for all our hashtables
- use 1 byte for the distance (as he does)
- use 1 byte for flags, currently only using DELETED and FREE (or USED), rest reserved for future
- 2 bytes reserved for future use or as "optional padding"

That would add 9..10% memory usage, but would simplify and generalize the code (e.g. no reserved number space, no in-band signaling for first value, no checks related to that, maybe use a 0-value for FREE, so just zeroing the table is ok for init).

If due to robin hood we can increase load factor from 0.75 to 0.85 or 0.90, we are still not using more memory than in 1.0.

jdchristensen:
About the alignment issue:  the comments to Skarupke's blog post suggest the idea of storing the distance bytes in groups of around 8, to avoid wasted padding.

About TW's last load factor comment:  With the bounded lookups, Skarupke's hash tables tend to resize when they are about 2/3 full, even if the load factor is set higher.  But one could increase the lookup bound from `log2(n)` to something slightly larger, like `1.2*log2(n)`, in order to achieve a higher load factor.

ThomasWaldmann:
@jdchristensen about separating / grouping together distances: seen that, too. but it needs more code and (potentially) could introduce bugs there. there is also the idea to separate values from keys into a separate array, so maybe these 2 can be done together later.

enkore:
> there is also the idea to separate values from keys into a separate array, so maybe these 2 can be done together later.

In case we change the data structure in such an incompatible way I'd _much rather_ have a tree there instead of any kind of hash table.

ThomasWaldmann:
about alignment / changing structure in incompatible way:
- first get this stuff in while staying compatible in layout
- later add upgrade mechanism and extra bytes
- one (some) of the extra bytes could be used for parity / checksum, see #2281.

ThomasWaldmann:
I've looked throught the code, looking for stuff we may want to merge (even if we don't merge the robin hood code):

- debug_print
- (idx)
- benchmarks
- new test

I noticed there is commented "if (DEBUG)" stuff, rather use the preprocessor?

src/borg/_hashindex.c
```diff
@@ -76,23 +88,25 @@ static int hash_sizes[] = {
...
+#define BUCKET_ADDR(index, idx) (index->buckets + ((idx) * index->bucket_size))
 
 #define BUCKET_MATCHES_KEY(index, idx, key) (memcmp(key, BUCKET_ADDR(index, idx), index->key_size) == 0)
 
 #define BUCKET_IS_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == DELETED)
 #define BUCKET_IS_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == EMPTY)
 
 #define BUCKET_MARK_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) = DELETED)
-#define BUCKET_MARK_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) = EMPTY)
+#define BUCKET_MARK_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, (idx)) + index->key_size)) = EMPTY)
```
ThomasWaldmann: hmm, is this change to (idx) needed?

as this is invoking the BUCKET_ADDR and that does internally add parens, shouldn't that be enough?

src/borg/_hashindex.c
```diff
@@ -76,23 +88,25 @@ static int hash_sizes[] = {
...
+                              value in archive.py */
 
 #define MAX(x, y) ((x) > (y) ? (x): (y))
 #define NELEMS(x) (sizeof(x) / sizeof((x)[0]))
 
 #define EMPTY _htole32(0xffffffff)
 #define DELETED _htole32(0xfffffffe)
 
-#define BUCKET_ADDR(index, idx) (index->buckets + (idx * index->bucket_size))
+#define BUCKET_ADDR(index, idx) (index->buckets + ((idx) * index->bucket_size))
```
ThomasWaldmann: I'll review this kind of stuff in a separate PR. This is somehow a pitfall waiting for someone to fall into.

ThomasWaldmann: Note: this was done and merged a while ago.

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,48 @@ def test_chunkindex_summarize(self):
         assert unique_chunks == 3
 
 
+class HashIndexExtraTestCase(BaseTestCase):
```
ThomasWaldmann: looks like we could add this via a separate PR to our testsuite.

src/borg/_hashindex.c
```diff
@@ -35,6 +35,17 @@
...
 
+#define DEBUG 0
+
+#define debug_print(fmt, ...)                   \
+  do {                                          \
+    if (DEBUG) {                                \
+      fprintf(stderr, fmt, __VA_ARGS__);        \
+      fflush(NULL);                             \
+    }                                           \
+  } while (0)
```
ThomasWaldmann: looks like something useful to have, we can add this in a separate PR.

src/borg/testsuite/hashindex.py
```diff
@@ -125,6 +125,48 @@ def test_chunkindex_summarize(self):
...
+    """These tests are separate because they should not become part of the selftest
+    """
+    def test_chunk_indexer(self):
+        # max_key is chosen so that when the deleted_keys get added we're close to max fill rate
+        # but never trigger a resize
+        max_key = int(1031 * (0.99*(2./3.)))
```
ThomasWaldmann: check if this is still correct when applying this separate from the rest.

src/borg/testsuite/hashindex.py
ThomasWaldmann: hmm, we already have a similar test.

ThomasWaldmann:
Closing this for now.

I made a separate PR for the debug_print macro as this is generally useful and just a small change.

I didn't make a separate PR yet for the (bigger) benchmark code yet, though. Can be done later when we need it.

ThomasWaldmann:
Thanks again to @rciorba for his research. \o/

PR #2231, opened 2017-02-28
Fix for PR #1971

leo-b:
This is a fix for the already merged #1971.
 --patterns-from was accessing args.roots instead of args.paths
I also added a test case that parses a command containing --patterns-from.


enkore:
Can you also create a separate PR with this change against the master branch?

PR #2232, opened 2017-02-28
Fix for PR #1971 (master)

leo-b:
This is a fix against master for the already merged #1971.
--patterns-from was accessing args.roots instead of args.paths
add a test case that parses a command containing --patterns-from


enkore:
Thanks!

Issue #2225, opened 2017-02-27
borg delete --force sometimes crashes with ObjectNotFound


ThomasWaldmann:
although that was the use case it was made for. strange is that it only happens sometimes.

```
[gw0] darwin -- Python 3.5.2 /Users/travis/build/borgbackup/borg/.tox/py35/bin/python3.5
self = <borg.testsuite.archiver.RemoteArchiverTestCase testMethod=test_delete_force>
    def test_delete_force(self):
        self.cmd('init', '--encryption=none', self.repository_location)
        self.create_src_archive('test')
        with Repository(self.repository_path, exclusive=True) as repository:
            manifest, key = Manifest.load(repository)
            archive = Archive(repository, key, manifest, 'test')
            for item in archive.iter_items():
                if 'chunks' in item:
                    first_chunk_id = item.chunks[0].id
                    repository.delete(first_chunk_id)
                    repository.commit()
                    break
>       self.cmd('delete', '--force', self.repository_location + '::test')
.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:1208: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:258: in cmd
    ret, output = exec_cmd(*args, fork=fork, exe=self.EXE, archiver=self.archiver, **kw)

.tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py:89: in exec_cmd
    ret = archiver.run(args)

.tox/py35/lib/python3.5/site-packages/borg/archiver.py:3228: in run
    return func(args)

.tox/py35/lib/python3.5/site-packages/borg/archiver.py:114: in wrapper
    return method(self, args, repository=repository, **kwargs)

.tox/py35/lib/python3.5/site-packages/borg/archiver.py:832: in do_delete
    return self._delete_archives(args, repository)

.tox/py35/lib/python3.5/site-packages/borg/archiver.py:877: in _delete_archives
    manifest.write()

.tox/py35/lib/python3.5/site-packages/borg/helpers.py:275: in write
    self.repository.put(self.MANIFEST_ID, self.key.encrypt(Chunk(data, compression={'name': 'none'})))

.tox/py35/lib/python3.5/site-packages/borg/remote.py:454: in do_rpc
[...]

PR #2238, opened 2017-03-01
WIP just a suspicion about #2225

ThomasWaldmann:
currently, I can't reproduce the problem any more.

but here it happened again: https://travis-ci.org/borgbackup/borg/jobs/207708189

if the remote borg serve tries to delete and does not find the object any more,
it will raise ObjectNotFound.

but if we did not wait for it, does just some of the next repo rpc calls issued
get the exception?

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -558,7 +558,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         count, size, csize = self.chunks.decref(id)
         if count == 0:
             del self.chunks[id]
-            self.repository.delete(id, wait=False)
+            self.repository.delete(id, wait=False) # wait=False responsible for #2225?
             stats.update(-size, -csize, True)
         else:
             stats.update(-size, -csize, False)


enkore:
Yes, failures of async (wait=False) calls will just pop up anywhere.

How about storing a trace for each msgid (only in --debug mode) and then attaching that trace in handle_error?

ThomasWaldmann:
sounds good, but maybe attach the trace only for exceptions?

ThomasWaldmann:
I had another look at the code, but currently have no good idea how to fix borg delete --force (or in general: how to sanely deal with async exceptions).

one approach that would fix the issue is to do repo.delete(id, wait=True) - but when doing this for many calls (like deleting all data and metadata chunks related to one archive), performance will likely suffer. 

another idea is to add a delete_many call with wait=True, which could overlap sending out requests with waiting for all responses, so the performance impact would likely be way smaller.

not being able to handle a failing chunk delete at the place where the call is somehow sucks.
the chunk delete's async exception response hitting a (working) manifest put and making it look failing even sucks more.

enkore:
^- Perhaps we rather want to work with e.g. barriers, i.e. explicitly wait for completion of outstanding work.

Suppose:

```
while True:
  try:
    if not repo.barrier(): break
  except ...
```

This would be a no-op for Repository.

ThomasWaldmann:
@enkore as waiting_for is currently local to call_many(), we lose it when using wait=False.

ThomasWaldmann:
see #2255

PR #2248, opened 2017-03-03
rebuild hashtable if we have too little empty buckets, fixes #2246

### LINKED ISSUE #2246: minor fix for borg 1.0 hashtable
Considering https://github.com/borgbackup/borg/issues/2245#issuecomment-283856546 I think we should do:

```
MAX_LOAD = 0.75  # we already have that
MAX_EFF_LOAD = 0.85  # new
```

```
eff_load = (buckets_used + tombstones) / capacity  # new
if eff_load > MAX_EFF_LOAD:
    resize(current_size)  # get rid of tombstones, keep size(*)
```

Equivalent: maintain a free bucket counter, compute min_free value according to MAX_EFF_LOAD and use that.

(*) there could be an addtl. check if the load is already far from what we target and in that case do a real resize to a different size.

ThomasWaldmann:
if there are too many deleted buckets (tombstones), hashtable performance goes down the drain.

in the worst case of 0 empty buckets and lots of tombstones, this results in full table scans for
new / unknown keys.

thus we make sure we always have a good amount of empty buckets.

### DIFF
--- a/borg/_hashindex.c
+++ b/borg/_hashindex.c
@@ -47,11 +47,13 @@ typedef struct {
     void *buckets;
     int num_entries;
     int num_buckets;
+    int num_empty;
     int key_size;
     int value_size;
     off_t bucket_size;
     int lower_limit;
     int upper_limit;
+    int min_empty;
 } HashIndex;
 
 /* prime (or w/ big prime factors) hash table sizes
@@ -77,6 +79,7 @@ static int hash_sizes[] = {
 
 #define HASH_MIN_LOAD .25
 #define HASH_MAX_LOAD .75  /* don't go higher than 0.75, otherwise performance severely suffers! */
+#define HASH_MAX_EFF_LOAD .93
 
 #define MAX(x, y) ((x) > (y) ? (x): (y))
 #define NELEMS(x) (sizeof(x) / sizeof((x)[0]))
@@ -167,8 +170,10 @@ hashindex_resize(HashIndex *index, int capacity)
     free(index->buckets);
     index->buckets = new->buckets;
     index->num_buckets = new->num_buckets;
+    index->num_empty = index->num_buckets - index->num_entries;
     index->lower_limit = new->lower_limit;
     index->upper_limit = new->upper_limit;
+    index->min_empty = new->min_empty;
     free(new);
     return 1;
 }
@@ -185,6 +190,11 @@ int get_upper_limit(int num_buckets){
     if (num_buckets >= max_buckets)
         return num_buckets;
     return (int)(num_buckets * HASH_MAX_LOAD);
+}
+
+int get_min_empty(int num_buckets){
+    /* Differently from load, the effective load also considers tombstones (deleted buckets). */
+    return (int)(num_buckets * (1.0 - HASH_MAX_EFF_LOAD));
 }
 
 int size_idx(int size){
@@ -218,6 +228,19 @@ int shrink_size(int current){
     if (i < 0)
         return hash_sizes[0];
     return hash_sizes[i];
+}
+
+int
+count_empty(HashIndex *index)
+{   /* count empty (never used) buckets. this does NOT include deleted buckets (tombstones).
+     * TODO: if we ever change HashHeader, save the count there so we do not need this function.
+     */
+    int i, count = 0, capacity = index->num_buckets;
+    for(i = 0; i < capacity; i++) {
+        if(BUCKET_IS_EMPTY(index, i))
+            count++;
+    }
+    return count;
 }
 
 /* Public API */
@@ -299,6 +322,17 @@ hashindex_read(const char *path)
     index->bucket_size = index->key_size + index->value_size;
     index->lower_limit = get_lower_limit(index->num_buckets);
     index->upper_limit = get_upper_limit(index->num_buckets);
+    index->min_empty = get_min_empty(index->num_buckets);
+    index->num_empty = count_empty(index);
+    if(index->num_empty < index->min_empty) {
+        /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
+        if(!hashindex_resize(index, index->num_buckets)) {
+            free(index->buckets);
+            free(index);
+            index = NULL;
+            goto fail;
+        }
+    }
 fail:
     if(fclose(fd) < 0) {
         EPRINTF_PATH(path, "fclose failed");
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
decide about branch when finished, also consider robin hood hashing PR (if that goes into master soon, we do not need this PR in master. if not, apply this to master and backport to 1.0).

borg/_hashindex.c
```diff
@@ -47,11 +47,13 @@ typedef struct {
     void *buckets;
     int num_entries;
     int num_buckets;
+    int num_empty;
```
ThomasWaldmann: it seemed to need less code to account empty buckets rather than tombstones, so i did that.

ThomasWaldmann:
@rciorba can you also review please?

borg/_hashindex.c
```diff
@@ -299,6 +322,16 @@ hashindex_read(const char *path)
     index->bucket_size = index->key_size + index->value_size;
     index->lower_limit = get_lower_limit(index->num_buckets);
     index->upper_limit = get_upper_limit(index->num_buckets);
+    index->min_empty = get_min_empty(index->num_buckets);
+    index->num_empty = count_empty(index);
+    if(index->num_empty < index->min_empty) {
+        /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
+        if(!hashindex_resize(index, index->num_buckets)) {
+            free(index);
```
rciorba: i think you need to free `index->buckets` as well since `hashindex_resize` will only return 0 before it calls free on them.

borg/_hashindex.c
```diff
@@ -400,6 +435,15 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
             idx = (idx + 1) % index->num_buckets;
         }
+        if(BUCKET_IS_EMPTY(index, idx)){
+            index->num_empty--;
```
rciorba: don't you want to also `index->num_empty++` in case you're inserting on top of a tombstone?

ThomasWaldmann: empty = never used. deleted = tombstone.

i am only counting empty buckets in num_empty, why should i increase it when inserting on a tombstone? tombstones/deleted get reduced by 1 due to that (but i do not count tombstones), but empty do not change.

borg/_hashindex.c
ThomasWaldmann: I'll have a look...

borg/_hashindex.c
rciorba: Mea culpa, you are right!

borg/_hashindex.c
ThomasWaldmann: yup, fixed it!

borg/_hashindex.c
```diff
@@ -77,6 +79,7 @@ static int hash_sizes[] = {
 
 #define HASH_MIN_LOAD .25
 #define HASH_MAX_LOAD .75  /* don't go higher than 0.75, otherwise performance severely suffers! */
+#define HASH_MAX_EFF_LOAD .85
```
enkore: Reasonably close to [this](https://github.com/leventov/Koloboke/blob/0b4898817f41b0820e0d9a2839fb593f112f9edc/lib/impl/src/main/javaTemplates/net/openhft/koloboke/collect/impl/hash/MutableDHash.java#L31) independent result (0.93)

borg/_hashindex.c
```diff
@@ -220,6 +230,19 @@ int shrink_size(int current){
     return hash_sizes[i];
 }
 
+int
+count_empty(HashIndex *index)
+{   /* count empty (never used) buckets. this does NOT include deleted buckets (tombstones).
+     * TODO: if we ever change HashHeader, save the count there so we do not need this function.
```
enkore: for future reference: mmaping the table thus does not make sense with this code here.

borg/_hashindex.c
ThomasWaldmann: hmm, i read the 0.83 value from a table (iirc from same guy).
if we could even go to 0.93, that would save us some resizes.

borg/_hashindex.c
ThomasWaldmann: now as you mention mmap: maybe having no empty buckets left now and then was one of the reasons why mmapping sucked in performance when jonas tried it - because the full table scan will load everything all the time (plus the memory pressure without fadvise will kick everything out again all the time).

borg/_hashindex.c
enkore: Hm, I didn't mean to say that 0.93 would be more correct, just that 0.85 seems to be the right ballpark. If in doubt, use the more conservative figure (0.85)?

borg/_hashindex.c
enkore: mmaping only makes sense if you advise the OS to actually load the data, otherwise there will be a large warmup period where you'll get a page fault for many accesses and have to block for essentially random-looking IO (since it's a hashtable...).

So, besides some minor advantages w.r.t. memory usage, mmaping can mainly make things faster that *don't* use the data structure much, say, a "borg list" that just retrieves a single item from the repository index.

borg/_hashindex.c
ThomasWaldmann: I'll check if I find again where I got the 0.85 from.
But just from a rough estimation, if we just had some few empty buckets distributed more or less evenly over the hashtable that already help tremendously compare to full table scans with no empty buckets.
0 empty -> full scan
10 empty evenly distr. -> 10% scan
100 -""- -> 1% scan
...

So, gut feeling is that anything <= 99% is already ok.
We aren't searching for the super high performance, good performance is already ok.

ThomasWaldmann: https://github.com/leventov/Koloboke/wiki/Tombstones-purge-from-hashtable:-theory-and-practice#theory-and-practice that was the table. but i guess i read it wrong.

enkore: Mmh but he also mentions that due to lookups happening more often than ins/del (which is also true for us)...

> In real practice, except insertions and removals there are many lookups, making rehash cost relatively less meaningful, therefore optimal _effective loads to rehash at should be lower, than in the 3rd and 4th columns of the table above._

ThomasWaldmann:
```
>>> l=.75
>>> 0.54 + 0.816 * l - 0.363 * l*l
0.9478125000000002
```
So the formula there even points to 0.95.

ThomasWaldmann:
@rciorba how about applying this to master? Guess we could do it for now, it could be easily ripped out again if we switch to robin hood.

ThomasWaldmann:
collapsed changesets.

rciorba:
@ThomasWaldmann Sounds like a good idea.

enkore:
Wouldn't the count-empty-on-load bit still make sense, even after a switch to RH, since we have to deal with older hashtables on disk?

rciorba:
Not really, since the entire load has to be different for RH to be
backwards compatible.
For RH the order of the keys matters, so if you're loading an old index you
always need to reorder and eliminate tombstones (via a resize).

PR #2250, opened 2017-03-04
hashindex: rebuild hashtable if we have too little empty buckets, fixes #2246

### LINKED ISSUE #2246: minor fix for borg 1.0 hashtable

ThomasWaldmann:
if there are too many deleted buckets (tombstones), hashtable performance goes down the drain.
in the worst case of 0 empty buckets and lots of tombstones, this results in full table scans for
new / unknown keys.
thus we make sure we always have a good amount of empty buckets.

same as #2248, for master branch.

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -47,11 +47,13 @@ typedef struct {
     void *buckets;
     int num_entries;
     int num_buckets;
+    int num_empty;
     int key_size;
     int value_size;
     off_t bucket_size;
     int lower_limit;
     int upper_limit;
+    int min_empty;
 } HashIndex;
 
 /* prime (or w/ big prime factors) hash table sizes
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
so the coverage report finds deltas in files i did not even touch. but does not list files i touched. :-(

enkore:
I completely ignore codecov, since the reports are almost always wrong, esp. if one cancels builds or restarts them.

PR #2253, opened 2017-03-04
fix temp filename collision for parallel running code

ThomasWaldmann:
hmm, does that make sense?

or does it just hide issues that we should not have in the first place?

### DIFF
--- a/src/borg/platform/base.py
+++ b/src/borg/platform/base.py
@@ -1,5 +1,6 @@
 import errno
 import os
+import tempfile
 
 """
 platform base module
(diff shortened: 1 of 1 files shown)


enkore:
Each test should have it's own repo etc. in a collision-free tmp path. This change suggests this is not the case? (Since these are internal files of the repo)

ThomasWaldmann:
that's what i meant in top post. i didn't review all tests, it was just a vague suspicion.

ThomasWaldmann:
i am starting this travis build a few times just to see if the isadirectory issue appears.

enkore:
My concern here is that we have no hypothesis why this fixes it / why this should fix it. That this fixes the problem either indicates that there is a bug in the underlying platform or that parallel test cases actually work in the same repository, which raises it's own set of questions (like, what happened to locking?!).

ThomasWaldmann:
ok, guess we better don't apply.

PR #2204, opened 2017-02-24
Test delete --force

ThomasWaldmann:
based on #2184, apply after #2184.

PR #2184, opened 2017-02-19
borg delete --force --force to delete severely corrupted archives, fixes #1975

ThomasWaldmann:

PR #2255, opened 2017-03-05
async exception processing

### LINKED ISSUE #2225: borg delete --force sometimes crashes with ObjectNotFound

ThomasWaldmann:
supercedes #2238, fixes #2225.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -260,7 +260,8 @@ class CacheChunkBuffer(ChunkBuffer):
         self.stats = stats
 
     def write_chunk(self, chunk):
-        id_, _, _ = self.cache.add_chunk(self.key.id_hash(chunk.data), chunk, self.stats)
+        id_, _, _ = self.cache.add_chunk(self.key.id_hash(chunk.data), chunk, self.stats, wait=False)
+        self.cache.repository.async_response(wait=False)
         return id_
 
 
@@ -469,6 +470,8 @@ Utilization of max. archive size: {csize_max:.0%}
         data = self.key.pack_and_authenticate_metadata(metadata.as_dict(), context=b'archive')
         self.id = self.key.id_hash(data)
         self.cache.add_chunk(self.id, Chunk(data), self.stats)
+        while self.repository.async_response(wait=True) is not None:
+            pass
         self.manifest.archives[name] = (self.id, metadata.time)
         self.manifest.write()
         self.repository.commit()
@@ -730,18 +733,27 @@ Utilization of max. archive size: {csize_max:.0%}
         class ChunksIndexError(Error):
             """Chunk ID {} missing from chunks index, corrupted chunks index - aborting transaction."""
 
-        def chunk_decref(id, stats):
-            nonlocal error
+        exception_ignored = object()
+
+        def fetch_async_response(wait=True):
             try:
-                self.cache.chunk_decref(id, stats)
-            except KeyError:
-                cid = bin_to_hex(id)
-                raise ChunksIndexError(cid)
+                return self.repository.async_response(wait=wait)
             except Repository.ObjectNotFound as e:
+                nonlocal error
                 # object not in repo - strange, but we wanted to delete it anyway.
                 if forced == 0:
                     raise
                 error = True
+                return exception_ignored  # must not return None here
+
+        def chunk_decref(id, stats):
+            try:
+                self.cache.chunk_decref(id, stats, wait=False)
+            except KeyError:
+                cid = bin_to_hex(id)
+                raise ChunksIndexError(cid)
+            else:
+                fetch_async_response(wait=False)
 
         error = False
         try:
@@ -778,6 +790,10 @@ Utilization of max. archive size: {csize_max:.0%}
         # some harmless exception.
         chunk_decref(self.id, stats)
         del self.manifest.archives[self.name]
+        while fetch_async_response(wait=True) is not None:
+            # we did async deletes, process outstanding results (== exceptions),
+            # so there is nothing pending when we return and our caller wants to commit.
+            pass
         if error:
             logger.warning('forced deletion succeeded, but the deleted archive was corrupted.')
             logger.warning('borg check --repair is required to free all space.')
@@ -865,7 +881,9 @@ Utilization of max. archive size: {csize_max:.0%}
     def chunk_file(self, item, cache, stats, chunk_iter, chunk_processor=None, **chunk_kw):
         if not chunk_processor:
             def chunk_processor(data):
-                return cache.add_chunk(self.key.id_hash(data), Chunk(data, **chunk_kw), stats)
+                chunk_entry = cache.add_chunk(self.key.id_hash(data), Chunk(data, **chunk_kw), stats, wait=False)
+                self.cache.repository.async_response(wait=False)
+                return chunk_entry
 
         item.chunks = []
         from_chunk = 0
@@ -1654,7 +1672,8 @@ class ArchiveRecreater:
             if Compressor.detect(old_chunk.data).name == compression_spec.name:
                 # Stored chunk has the same compression we wanted
                 overwrite = False
-        chunk_entry = self.cache.add_chunk(chunk_id, chunk, target.stats, overwrite=overwrite)
+        chunk_entry = self.cache.add_chunk(chunk_id, chunk, target.stats, overwrite=overwrite, wait=False)
+        self.cache.repository.async_response(wait=False)
         self.seen_chunks.add(chunk_entry.id)
         return chunk_entry
 
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -524,7 +524,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
             self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
-    def add_chunk(self, id, chunk, stats, overwrite=False):
+    def add_chunk(self, id, chunk, stats, overwrite=False, wait=True):
         if not self.txn_active:
             self.begin_txn()
         size = len(chunk.data)
@@ -533,7 +533,7 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
             return self.chunk_incref(id, stats)
         data = self.key.encrypt(chunk)
         csize = len(data)
-        self.repository.put(id, data, wait=False)
+        self.repository.put(id, data, wait=wait)
         self.chunks.add(id, 1, size, csize)
         stats.update(size, csize, not refcount)
         return ChunkListEntry(id, size, csize)
@@ -554,13 +554,13 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
         stats.update(size, csize, False)
         return ChunkListEntry(id, size, csize)
 
-    def chunk_decref(self, id, stats):
+    def chunk_decref(self, id, stats, wait=True):
         if not self.txn_active:
             self.begin_txn()
         count, size, csize = self.chunks.decref(id)
         if count == 0:
             del self.chunks[id]
-            self.repository.delete(id, wait=False)
+            self.repository.delete(id, wait=wait)
             stats.update(-size, -csize, True)
         else:
             stats.update(-size, -csize, False)
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -514,6 +514,7 @@ class RemoteRepository:
         self.ignore_responses = set()
         self.responses = {}
         self.async_responses = {}
+        self.shutdown_time = None
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
         self.unpacker = get_limited_unpacker('client')
         self.server_version = parse_version('1.0.8')  # fallback version if server is too old to send version information
@@ -605,6 +606,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
     def __exit__(self, exc_type, exc_val, exc_tb):
         try:
             if exc_type is not None:
+                self.shutdown_time = time.monotonic() + 30
                 self.rollback()
         finally:
             # in any case, we want to cleanly close the repo, even if the
@@ -715,6 +717,12 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
         calls = list(calls)
         waiting_for = []
         while wait or calls:
+            if self.shutdown_time and time.monotonic() > self.shutdown_time:
+                # we are shutting this RemoteRepository down already, make sure we do not waste
+                # a lot of time in case a lot of async stuff is coming in or remote is gone or slow.
+                logger.debug('shutdown_time reached, shutting down with %d waiting_for and %d async_responses.',
+                             len(waiting_for), len(self.async_responses))
+                return
             while waiting_for:
                 try:
                     unpacked = self.responses.pop(waiting_for[0])
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -513,6 +513,7 @@ class RemoteRepository:
         self.chunkid_to_msgids = {}
         self.ignore_responses = set()
         self.responses = {}
+        self.async_responses = {}
         self.ratelimit = SleepingBandwidthLimiter(args.remote_ratelimit * 1024 if args and args.remote_ratelimit else 0)
         self.unpacker = get_limited_unpacker('client')
         self.server_version = parse_version('1.0.8')  # fallback version if server is too old to send version information
@@ -670,8 +671,8 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
         for resp in self.call_many(cmd, [args], **kw):
             return resp
 
-    def call_many(self, cmd, calls, wait=True, is_preloaded=False):
-        if not calls:
+    def call_many(self, cmd, calls, wait=True, is_preloaded=False, async_wait=True):
+        if not calls and cmd != 'async_responses':
             return
 
         def pop_preload_msgid(chunkid):
@@ -726,6 +727,22 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                             return
                 except KeyError:
                     break
+            if cmd == 'async_responses':
+                while True:
+                    try:
+                        msgid, unpacked = self.async_responses.popitem()
+                    except KeyError:
+                        # there is nothing left what we already have received
+                        if async_wait and self.ignore_responses:
+                            # but do not return if we shall wait and there is something left to wait for:
+                            break
+                        else:
+                            return
+                    else:
+                        if b'exception_class' in unpacked:
+                            handle_error(unpacked)
+                        else:
+                            yield unpacked[RESULT]
             if self.to_send or ((calls or self.preload_ids) and len(waiting_for) < MAX_INFLIGHT):
                 w_fds = [self.stdin_fd]
             else:
@@ -755,8 +772,14 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                             raise UnexpectedRPCDataFormatFromServer(data)
                         if msgid in self.ignore_responses:
                             self.ignore_responses.remove(msgid)
+                            # async methods never return values, but may raise exceptions.
                             if b'exception_class' in unpacked:
-                                handle_error(unpacked)
+                                self.async_responses[msgid] = unpacked
+                            else:
+                                # we currently do not have async result values except "None",
+                                # so we do not add them into async_responses.
+                                if unpacked[RESULT] is not None:
+                                    self.async_responses[msgid] = unpacked
                         else:
                             self.responses[msgid] = unpacked
                 elif fd is self.stderr_fd:
@@ -805,7 +828,7 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                         # that the fd should be writable
                         if e.errno != errno.EAGAIN:
                             raise
-        self.ignore_responses |= set(waiting_for)
+        self.ignore_responses |= set(waiting_for)  # we lose order here
 
     @api(since=parse_version('1.0.0'),
          append_only={'since': parse_version('1.0.7'), 'previously': False})
@@ -882,6 +905,10 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
             self.p.stdout.close()
             self.p.wait()
             self.p = None
+
+    def async_response(self, wait=True):
+        for resp in self.call_many('async_responses', calls=[], wait=True, async_wait=wait):
+            return resp
 
     def preload(self, ids):
         self.preload_ids += ids
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -862,6 +862,11 @@ class Repository:
             yield self.get(id_)
 
     def put(self, id, data, wait=True):
+        """put a repo object
+
+        Note: when doing calls with wait=False this gets async and caller must
+              deal with async results / exceptions later.
+        """
         if not self._active_txn:
             self.prepare_txn(self.get_transaction_id())
         try:
@@ -881,6 +886,11 @@ class Repository:
         self.index[id] = segment, offset
 
     def delete(self, id, wait=True):
+        """delete a repo object
+
+        Note: when doing calls with wait=False this gets async and caller must
+              deal with async results / exceptions later.
+        """
         if not self._active_txn:
             self.prepare_txn(self.get_transaction_id())
         try:
@@ -894,6 +904,17 @@ class Repository:
         segment, size = self.io.write_delete(id)
         self.compact[segment] += size
         self.segments.setdefault(segment, 0)
+
+    def async_response(self, wait=True):
+        """Get one async result (only applies to remote repositories).
+
+        async commands (== calls with wait=False, e.g. delete and put) have no results,
+        but may raise exceptions. These async exceptions must get collected later via
+        async_response() calls. Repeat the call until it returns None.
+        The previous calls might either return one (non-None) result or raise an exception.
+        If wait=True is given and there are outstanding responses, it will wait for them
+        to arrive. With wait=False, it will only return already received responses.
+        """
 
     def preload(self, ids):
         """Preload objects (only applies to remote repositories)
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -785,6 +785,7 @@ class Repository:
         self._active_txn = False
 
     def rollback(self):
+        # note: when used in remote mode, this is time limited, see RemoteRepository.shutdown_time.
         self._rollback(cleanup=False)
 
     def __len__(self):
--- a/src/borg/testsuite/archive.py
+++ b/src/borg/testsuite/archive.py
@@ -63,10 +63,15 @@ This archive:                   20 B                 10 B                 10 B""
 
 class MockCache:
 
+    class MockRepo:
+        def async_response(self, wait=True):
+            pass
+
     def __init__(self):
         self.objects = {}
+        self.repository = self.MockRepo()
 
-    def add_chunk(self, id, chunk, stats=None):
+    def add_chunk(self, id, chunk, stats=None, wait=True):
         self.objects[id] = chunk.data
         return id, len(chunk.data), len(chunk.data)
 
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -1255,7 +1255,8 @@ class ArchiverTestCase(ArchiverTestCaseBase):
                     repository.delete(first_chunk_id)
                     repository.commit()
                     break
-        self.cmd('delete', '--force', self.repository_location + '::test')
+        output = self.cmd('delete', '--force', self.repository_location + '::test')
+        self.assert_in('deleted archive was corrupted', output)
         self.cmd('check', '--repair', self.repository_location)
         output = self.cmd('list', self.repository_location)
         self.assert_not_in('test', output)


src/borg/archive.py
```diff
@@ -433,6 +433,7 @@ def write_checkpoint(self):
         self.save(self.checkpoint_name)
         del self.manifest.archives[self.checkpoint_name]
         self.cache.chunk_decref(self.id, self.stats)
```
ThomasWaldmann: maybe adding a wait=True default to chunk_decref would be prettier.
so that one only has to do barriers if one actually requested an async delete.

src/borg/archive.py
```diff
@@ -723,15 +725,11 @@ class ChunksIndexError(Error):
         def chunk_decref(id, stats):
             nonlocal error
             try:
+                # note: this might do an async repo.delete(id)
                 self.cache.chunk_decref(id, stats)
```
ThomasWaldmann: maybe adding a wait=True default to chunk_decref would be prettier, so we would need wait=False being given here and the async usage gets explicit without a comment.

src/borg/archive.py
```diff
@@ -723,15 +725,11 @@ class ChunksIndexError(Error):
...
             except KeyError:
                 cid = bin_to_hex(id)
                 raise ChunksIndexError(cid)
-            except Repository.ObjectNotFound as e:
-                # object not in repo - strange, but we wanted to delete it anyway.
-                if forced == 0:
-                    raise
-                error = True
```
ThomasWaldmann: this now never happens as all async exceptions get collected and raised with barrier() calls.

previously, this maybe was not very reliable: some exceptions triggered here, some maybe in later/other calls.

src/borg/remote.py
```diff
@@ -856,6 +872,10 @@ def put(self, id, data, wait=True):
     def delete(self, id, wait=True):
         """actual remoting is done via self.call in the @api decorator"""
 
+    @api(since=parse_version('1.0.99'))
```
ThomasWaldmann: decide about version later

src/borg/repository.py
```diff
@@ -876,6 +886,15 @@ def delete(self, id, wait=True):
         self.compact[segment] += size
         self.segments.setdefault(segment, 0)
 
+    def barrier(self):
+        """Get one async result (only applies to remote repositories).
+
+           async commands (== calls with wait=False, e.g. delete and put) have no results,
+           but may raise exceptions. These async exceptions must get collected later via
+           barrier() calls. Repeat barrier calls until it returns without exception (all
+           previous calls will raise the exceptions we got from the async commands, if any).
```
ThomasWaldmann: need to outdent this

ThomasWaldmann:
naming: s/barrier/async_result/g ?

src/borg/remote.py
enkore: Since it isn't a real RPC call there is no version requirerement

textshell:
How sure are we that recovery from errors actually works correct without a rollback? This seems to make a error -> commit sequence much more likely, so this might be raising sleeping dragons.

Apart from that having a synchronization point to collect the async errors is a good idea in this implementation. Not sure how i would design this in 1.2, which might be different. But of course thats a "future us" problem.

src/borg/remote.py
```diff
@@ -726,6 +727,16 @@ def handle_error(unpacked):
                             return
                 except KeyError:
                     break
+            if cmd == 'barrier' and not self.ignore_responses:
```
textshell: abusing a command for this seems like a kludge, but maybe it still is the way to go if we want to fix this for 1.1

src/borg/remote.py
```diff
@@ -726,6 +727,16 @@ def handle_error(unpacked):
...
+                # start processing barrier command(s) when we have received ALL outstanding async responses.
+                # raise one async exception per barrier command, in arbitrary order:
+                try:
+                    msgid, unpacked = self.async_responses.popitem()
+                except KeyError:
+                    # nothing left, don't raise exception, just return (None)
+                    return
+                else:
+                    handle_error(unpacked)
             if self.to_send or ((calls or self.preload_ids) and len(waiting_for) < MAX_INFLIGHT):
```
textshell: I think the current code triggers this, so w_fds is set and pops one call in the write code path later, that makes this code even have more special cases. One way would be to ensure that calls is empty (by skipping the @api decorator and calling call_many explicitly below) from the start.

enkore:
We need to serialize repository IO in a multi-threaded setting anyway...

src/borg/remote.py
```diff
@@ -755,8 +766,11 @@ def handle_error(unpacked):
                             raise UnexpectedRPCDataFormatFromServer(data)
                         if msgid in self.ignore_responses:
                             self.ignore_responses.remove(msgid)
+                            # async methods never return values, but may raise exceptions.
                             if b'exception_class' in unpacked:
-                                handle_error(unpacked)
```
textshell: Do we really want to arbitrarily delay error handling and grow self.async_responses without small bound here? That seems like a problem in cases where we do not plan to ignore the errors anyway.
Does this need to be parameterised? Do we need to ensure that the errors are only delivered in the barrier() for the repository fixing commands, or is the important part just that the current command is executed and no errors occur after the barrier() call?

Thinking about this... Barrier currently conflates two different parts. It ensures that all outstanding calls are finished and it does retrieve the accumulated errors. Maybe have barrier() just wait and have a lightweight call that retrieves the async errors. So this would allow retrieving errors without delay as soon as they are seen, but does not disturb the current command?

src/borg/remote.py
```diff
@@ -778,6 +792,8 @@ def handle_error(unpacked):
                             args = calls.pop(0)
                             if cmd == 'get' and args['id'] in self.chunkid_to_msgids:
                                 waiting_for.append(pop_preload_msgid(args['id']))
+                            elif cmd == 'barrier':
```
textshell: I think we can get rid of this by calling call_many explicitly.

src/borg/remote.py
textshell: I think we should not use @api here. Because it is not a real rpc api.

src/borg/remote.py
```diff
@@ -856,6 +872,10 @@ def put(self, id, data, wait=True):
     def delete(self, id, wait=True):
         """actual remoting is done via self.call in the @api decorator"""
 
+    @api(since=parse_version('1.0.99'))
+    def barrier(self):
+        """actual remoting is done via self.call in the @api decorator"""
```
textshell: There is no remoting here.
Just do a `self.call_many('barrier', [])` with a command documenting this kludge here explicitly? The empty calls would avoid going into the write path and popping off the call there without a rpc call. This would need a special mention in the guard at the start of call_many for empty calls, but that still seems cleaner than the current dance.

src/borg/remote.py
```diff
@@ -513,6 +513,7 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
         self.chunkid_to_msgids = {}
         self.ignore_responses = set()
         self.responses = {}
+        self.async_responses = {}
```
textshell: call this `async_errors`?

src/borg/remote.py
enkore: Good point, maybe this shouldn't be exposed at all externally, rather, have rollback *act* like a barrier. rollback could get a parameter such as "ignore_previous_errors" or so that discards errors from previous RPC calls. I have to think a bit if that's safe behaviour though.

textshell: the `delete --force` case is explicitly not about rolling back on error.
`delete --force` uses a `on error resume next` policy. Because at that point repository inconsistencies are expected and it should just get rid of anything that is left.
I think the old behavior is ok when combined with rollback (just maybe not giving good error messages).

enkore: Ah right.

src/borg/remote.py
ThomasWaldmann: yes, currently that would be more specific.
but maybe we want later also non-error responses there? so i just used the more general term.
in fact i already had that, when I noticed they do not return anything anyway.

src/borg/remote.py

src/borg/remote.py

src/borg/remote.py

src/borg/remote.py
ThomasWaldmann: yup, did that.

src/borg/cache.py
```diff
@@ -524,6 +524,7 @@ def legacy_cleanup():
             self.chunks = create_master_idx(self.chunks)
 
     def add_chunk(self, id, chunk, stats, overwrite=False):
+        # TODO: this is doing an async put, callers need to call repo.barrier for async exceptions
```
enkore: (or any other synchronous, wait=True, RPC call)

ThomasWaldmann: that will only get you one exception, in a potentially confusing way.

enkore: well yeah, but usually that means the operation is failed, i.e. what I wanted to say here is that this is still safe w.r.t. the transactional model used.

src/borg/remote.py
ThomasWaldmann: ok, implemented the split, now we have async_wait and async_response method.

src/borg/remote.py
```diff
@@ -726,6 +727,22 @@ def handle_error(unpacked):
...
+                    try:
+                        msgid, unpacked = self.async_responses.popitem()
+                    except KeyError:
+                        # nothing left
+                        return
+                    else:
+                        if b'exception_class' in unpacked:
+                            handle_error(unpacked)
+                        else:
+                            yield unpacked[RESULT]
```
ThomasWaldmann: this is capable yielding non-error responses, but we do not have any yet.

src/borg/remote.py
ThomasWaldmann: I've changed it again, so it actually could process non-error async responses.

src/borg/cache.py
```diff
@@ -553,13 +554,13 @@ def chunk_incref(self, id, stats):
...
 
-    def chunk_decref(self, id, stats):
+    def chunk_decref(self, id, stats, wait=True):
         if not self.txn_active:
             self.begin_txn()
         count, size, csize = self.chunks.decref(id)
         if count == 0:
             del self.chunks[id]
-            self.repository.delete(id, wait=False)
+            self.repository.delete(id, wait=wait)
```
ThomasWaldmann: makes some places in the code simpler and errors there clearer.

only if we do mass decrefs, we need async == at one place, in borg delete archive.

src/borg/remote.py
```diff
@@ -883,6 +906,14 @@ def close(self):
...
             self.p = None
 
+    def async_wait(self):
+        for resp in self.call_many('async_wait', calls=[], wait=True):
+            return resp
+
+    def async_response(self):
+        for resp in self.call_many('async_responses', calls=[], wait=True):
+            return resp
+
```
ThomasWaldmann: shall we add a async_response**s** also?

textshell:
I’m not too keen on async responses in the current model. Because they would tend to pile up and might be very hard to dispatch. If borg was using more modern async patterns like futures things might be different. And i don‘t really see much of a use case currently.

I’m sorry for not seeing the coming, but even with the error only responses we have a potential pile up of errors problem. So everything in borg that uses wait=False calls needs to regularly poll async_response which should not be too bad if it‘s really cheap in the happy case. But async_wait currently can pile up quite a bit of errors, can‘t it (i.e. if the first error involves long io waits and there are lots of unprocessed commands)?

Of course we could make wait an enum with semantics of (sync, async and error out, async and keep going on) so most parts of the current code can still keep with the old model. But that doesn‘t sound too great.

ThomasWaldmann:
@textshell well, it piles up as much as the caller lets it pile up.

If you don't want any pile up, you have to do a sync request (wait=True).
Good: if it blows up, it is at the right place with a clearly related exception.
Bad: performance (but no issue for single/infrequent requests).

If you want only max. N pile up, you have to request the async responses every N requests.

As there are no non-failure results in async_responses right now, the good case pile up is zero (no exceptions happened), no matter how many N requests you did.

What I dislike about the old code is that some "old" async exception suddenly raises at a relatively unrelated place (some later, sometimes totally different command, e.g. when writing the manifest).

With the new code, while still being async, it does not do that, it raises whereever async responses are processed, which is usually somewhat close to where the requests were sent. And with the async_wait at the end, we actively make sure it doesn't raise somewhere totally unrelated and lets a successful request look like a failure.

enkore:
Exceptions should be raised as soon as they arrive, I don't see a reason why that should be changed.

ThomasWaldmann:
@enkore then N=1: send request, check async responses, send request. That somehow would be like the old code, except being more explicit. Plus having the async_wait + another check async response to separate from any following different request.

enkore:
Where is the benefit in doing that? Okay, errors show up a bit closer to the request that caused them (has that been an issue in the past?), but on the other hand we have to run through the whole code base and add that stuff everywhere.

ThomasWaldmann:
I didn't look much at PUT yet, but for DELETE it is only 1 place (borg delete archive) that should do it async (for performance reasons).

For PUT, I guess it is just borg create and recreate (and specifically add_chunk).

I first wanted to have an agreement about the fundamentals before looking at PUT, though.

We often would not notice mismatches caused by async processing btw - just in case of manifest writes we know it is ID 0 and if then an error pops up referring to something else than zero, the mismatch becomes obvious (but still it is annoying/confusing). For all other requests, we usually do not know the requested ID, because a traceback only contains the failed ID.

In case we just want to know if everything succeeded or not, that is sufficient though.

ThomasWaldmann:
Reviewed the PUT places, 3 places need an async add_chunk for performance reasons:
- metadata stream writes
- content writes (create and recreate)

The other few places do just a single add_chunk and would be better of with a sync call.

I'ld just add a wait=True to add_chunk, so async gets explicit and sync is the default.

src/borg/archive.py
```diff
@@ -768,6 +764,18 @@ def chunk_decref(id, stats):
...
+            # we did async deletes, process outstanding results (== exceptions),
+            # so there is nothing pending when we return and our caller wants to commit.
+            try:
+                if self.repository.async_response() is None:
+                    break
+            except Repository.ObjectNotFound as e:
+                # object not in repo - strange, but we wanted to delete it anyway.
+                if forced == 0:
+                    raise
+                error = True
```
textshell: We need this after calling chunk_decref too (excluding the wait of course).

src/borg/remote.py
ThomasWaldmann: yes, I considered that, but did not want to move too much stuff around just for that.

ThomasWaldmann:
@textshell "I think we need a async_wait / drain async_responses for create too." -> yup, will add that.

src/borg/remote.py
textshell: I would prefer that call_many monster not growing with avoidable strangeness.

ThomasWaldmann:
Hopefully I've addressed everything now (except moving stuff to other scopes, which I'ld like to do in a separate PR). Tell me when you finished reviewing, so I can clean up commits. Or if there is anything else that needs to be addressed in first PR.

ThomasWaldmann:
And I'ld really like to have clear separation between a request blowing up because it (itself) failed and failures unrelated to that request / to that line in the sourcecode.

wait=False is only used at a few places and they are now addressed, also wait=False is always explicit now and not the default any more.

src/borg/repository.py
```diff
@@ -876,6 +886,19 @@ def delete(self, id, wait=True):
         self.compact[segment] += size
         self.segments.setdefault(segment, 0)
 
+    def async_wait(self):
+        """Wait until all outstanding async commands' responses have been received.
+        """
+
+    def async_response(self):
+        """Get one async result (only applies to remote repositories).
```
enkore: µnitpick: docstring format should either be """oneline"""" or """\n multi\n line\n """, not mixed

enkore:
Aha, that was the rationale I was missing here. Quick look ok, more in-depth review later.

ThomasWaldmann:
collapsed changesets, edited commit comments. no code changes.

ThomasWaldmann:
ok, guess I'm finished with this.

enkore:
We need to backport this if we want to backport #2184 and #2204, don't we?

ThomasWaldmann:
No async changes required for #2184 (it just blows away manifest entries and commits).

#2204 requires the fixed async processing or it will raise ObjectNotFound now and then (as we've seen recently on travis).

enkore:
My understanding is that the errors seen with #2204 are not false positives, so using it would be racy/ may require any number of tries until it works?

ThomasWaldmann:
Yes, double force is no problem, the single force test would often fail without the async fixes.

ThomasWaldmann:
@textshell are you ok with it now?

ThomasWaldmann:
would be good to merge this asap so the borg delete --force test stops failing the builds.

src/borg/archive.py
```diff
@@ -468,6 +470,8 @@ def save(self, name=None, comment=None, timestamp=None, additional_metadata=None
         data = self.key.pack_and_authenticate_metadata(metadata.as_dict(), context=b'archive')
         self.id = self.key.id_hash(data)
         self.cache.add_chunk(self.id, Chunk(data), self.stats)
+        self.repository.async_wait()
```
textshell: I still fear that in async_wait() errors may pile up because with bad timing an slow io error might only return an error when the whole operation is already in the network / ssh pipe.

But maybe that is not really going to happen. Also the pileup is saved in a compact form not as exceptions as i first thought. So maybe just a theoretical problem.

ThomasWaldmann: Guess the only way to avoid that (us sending [many] requests fast and remote taking long, so everything arrives in the async_wait) is to have some sort of window.

E.g. with window=100, call_many would only send requests while the "responses we are waiting for" list length is below 100. Otherwise it would continue looping waiting for a response.

textshell: Yes, a window on responses would do the job. Most code pathes could work with a window of 1 (just enough to detect errors). Maybe make window a non defaulted parameter so this shows up in reviews more easily?
My other idea which was less orthogonal was to instead have a async_wait_and_get_response method that would just be window=1 and combines async_response into it. Less pure, but maybe easier to use.

enkore: I don't think that's a good idea, since it has the same problem the fixed preload window has, i.e. it leads to poor utilization of bandwidth if latency is high.

ThomasWaldmann: hmm, how about wait=True -> window=0 as default? that is equivalent to sync, right?

window>0 would be async.

textshell: async_wait_and_get_response or  windowed async_wait are only used for waiting when all useful work is already queued. So i don‘t see why this could be bad for utilization.
For the work submission phase it should stay operation(..., wait=False) and frequent async_response() polling.

enkore: > async_wait_and_get_response or windowed async_wait are only used for waiting when all useful work is already queued. So i don‘t see why this could be bad for utilization.

vs.

> E.g. with window=100, call_many would only send requests while the "responses we are waiting for" list length is below 100. Otherwise it would continue looping waiting for a response.

I.e. the window size means how many requests can be in flight. If the window is full, then blocking occurs until responses come in. Thus "leads to poor utilization of bandwidth if latency is high."

Or I'm misunderstanding.

textshell: Maybe interesting. in 1.0.10 rollback / uncleanly leaving the contact manager would read a async error and then (via finally in __leave__) semi forceably close the connection (close all fds at least and wait for ssh to finish someway). The current code doesn‘t result in any kind of bailout and will just pile up all errors anyway and wait for a response to rollback.

src/borg/repository.py
```diff
@@ -1123,8 +1107,7 @@ def recover_segment(self, segment, filename):
                 data = data[size:]
 
     def read(self, segment, offset, id, read_data=True):
-        """
-        Read entry from *segment* at *offset* with *id*.
+        """Read entry from *segment* at *offset* with *id*.
 
         If read_data is False the size of the entry is returned instead and integrity checks are skipped.
```
enkore: I thought we did it the other way around, PEP8 style?

```
"""Single line""""
```

```
"""
Multi
line
""""
```

src/borg/archive.py
textshell: Ok, this is getting confusing with the non real time interface in github and edits.

> I think each time the borg server sends an error it will also free space in the other direction so that a command can be send

I was referring to the (byte stream level) windowing that the lower layers (pipe buffers, ssh, tcp) already implement.

> I believe so, with the current state. As we always immediately check for async_responses after sending an async request it should be virtually identical to what the previous code did 

I think the shutdown paths are not covered. Not sure if the new or the old behavior is better though.

Old: With 2 or more async errors:
* random operation is aborted with first error.
* `__leave__` of the repository context manager is called.
* issues a rollback
* is aborted with next async error
* finally does a close which closes all remote file descriptors and waits for ssh to terminate

new: With 2 or more async error and not delete --force:
* async_response raises exception
* `__leave__` of the repository context manager is called.
* issues a rollback
* does not abort. fills RemoeRepository.async_responses while waiting for remote to send response to rollback
* finally does a close which closes all remote file descriptors and waits for ssh to terminate

src/borg/repository.py
ThomasWaldmann: there is no error from pep8 checker, so i won't mix functional changes in this PR with pep8 refactoring. i guess we already have all the styles in our source.

src/borg/archive.py
ThomasWaldmann: i think the only way to avoid too much stuff piling up at some place is to stop processing stuff if we notice we have too much in flight. and that should not be a perf problem as long as window_size * datasize > bandwidth * latency (or so). of course we currently do not know most of these variables....

for the special issue of async_wait waiting a long time to receive all the errors (in case everything is going wrong, but was delayed that much so it was not received before we got into async_wait):
async_wait could get a param fail_fast (default True) and would raise exceptions by itself if fail_fast is True.

ThomasWaldmann: maybe it could be even 1 call again: async_response(barrier=False).
- without barrier, it would just fetch one response.
- with barrier, it would fetch one response and even wait for it, if it is not there yet. or return None if there is nothing left.
(just a vague idea, not fully thought through)

src/borg/repository.py
enkore: Ok, true -- we can clean them up / make them consistent sometime, separate change.

src/borg/archive.py
textshell: `fail_fast` seems to be what i described as async_wait_and_get_response for the current situation that there are no non error async responses. I tend to prefer two distinct method names over a barrier boolean parameter. But that is of course a style detail. Also wait is my preferred color of the bikeshed.

src/borg/repository.py
ThomasWaldmann: ok, i saved the patch for the docstyle fixes and removed it from this PR, so we can apply it later.

src/borg/archive.py
```diff
@@ -468,6 +469,8 @@ def save(self, name=None, comment=None, timestamp=None, additional_metadata=None
         data = self.key.pack_and_authenticate_metadata(metadata.as_dict(), context=b'archive')
         self.id = self.key.id_hash(data)
         self.cache.add_chunk(self.id, Chunk(data), self.stats)
+        while self.repository.async_response(wait=True) is not None:
```
ThomasWaldmann: as there is nothing except exceptions yet in the responses, this will either terminate or raise immediately.

textshell:
It seems that the original comment was incomplete

I worry that the new code does perform differently when a borg command that uses async calls fails because of any exception (that is not caught or causes another exception to be thrown) and there are still async errors either queued in RemoteRepo or later received:

Old: With 2 or more async errors:
* random operation is aborted with first exception.
* __leave__ of the repository context manager is called.
* issues a rollback
* is aborted with next async error (i.e. throws an exception of an unrelated earlier async command)
* `finally` in context manager does a close which closes all remote file descriptors and waits for ssh to terminate
* ssh notices that it‘s output pipe is closed and closes it‘s input pipe on the remote side.
* remote borg is forced to stop sending replies and terminates rolling back 
* client has not to wait long to ssh to exit

new: With 2 or more async error and not delete --force:
* async_response raises exception
* __leave__ of the repository context manager is called.
* issues a rollback
* does not abort. fills RemoteRepository.async_responses while waiting for remote to send response to rollback
* finally does a close which closes all remote file descriptors and waits for ssh to terminate

So the old code would terminate the remote borg a lot faster than the new code when encountering errors. Or am i missing something here?

src/borg/archive.py
ThomasWaldmann: @textshell about https://github.com/borgbackup/borg/pull/2255#discussion_r105570618

I guess old behaviour was also bad somehow. it just sent out rollback (hopefully) and then blew up when it fell over the async exception of some unrelated command.

so it didn't really get the response for rollback (== whether it was done on the remote end). maybe it even killed the whole thing before it was even able to get to remote / being processed by remote?

(copied from irc)

enkore:
It would seem so, but it would only make a difference for cases that seem somewhat rare to me. I see this happening e.g. if there is a repeating error on the remote end, like ENOSPC, so after a certain point all writes fail. Then there would be a window (one RTT) where failures build up. This would have to be drained like you describe in the closing error handling.

Since the general approach seems to be settled now, it may make more sense to special-case this in `__leave__` (when exc_value is set)?

I still have my doubts about 1.0-maint though about the complexity|implications of the change, maybe a simpler / more minimal change should be done there.

ThomasWaldmann:
Hmm, suddently seeing IsADirectoryError more frequently. Coincidence?

ThomasWaldmann:
@enkore @textshell any comments about last 2 commits?

if you have better ideas, please consider pushing your ideas in form of changesets to this branch.

enkore:
That should clear the last blocker

ThomasWaldmann:
@textshell can you have a look now also?

src/borg/remote.py
```diff
@@ -715,6 +717,10 @@ def handle_error(unpacked):
         calls = list(calls)
         waiting_for = []
         while wait or calls:
+            if self.shutdown_time and time.time() > self.shutdown_time:
+                # we are shutting this RemoteRepository down already, make sure we do not waste
+                # a lot of time in case a lot of async stuff is coming in or remote is gone or slow.
+                return
```
textshell: Add a debug log that includes len(waiting_for) and len(self.async_responses) here?

textshell:
I hate how github hides comments when the file is changed in a later commit. 
https://github.com/borgbackup/borg/pull/2255#discussion_r108061534

ThomasWaldmann:
@textshell ok, I did the changes you had suggested in another fixup commit.

If there's nothing else, I'll collapse changsets a bit and merge after travis is happy.

textshell:
sounds good.

ThomasWaldmann:
@enkore remove backport/1.0-maint? it seems the risk of backporting this is somehow higher than the win.

enkore:
Agreed.

PR #2258, opened 2017-03-05
extract: warning RC for unextracted big extended attributes, followup

Abogical:
Previous PR: #2198
This sets a warning exit code whenever an xattr is too big.

Was waiting for #2199 to be merged.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -26,7 +26,7 @@ from .constants import *  # NOQA
 from .hashindex import ChunkIndex, ChunkIndexEntry
 from .helpers import Manifest
 from .helpers import Chunk, ChunkIteratorFileWrapper, open_item
-from .helpers import Error, IntegrityError
+from .helpers import Error, IntegrityError, set_ec
 from .helpers import uid2user, user2uid, gid2group, group2gid
 from .helpers import parse_timestamp, to_localtime
 from .helpers import format_time, format_timedelta, format_file_size, file_status, FileSize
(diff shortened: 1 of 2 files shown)


src/borg/archive.py
```diff
@@ -691,6 +691,7 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
                 if e.errno == errno.E2BIG:
                     logger.warning('%s: Value or key of extended attribute %s is too big for this filesystem' %
                                    (path, k.decode()))
+                    set_ec(EXIT_WARNING)
```
ThomasWaldmann: ok

ThomasWaldmann: not directly related to your previous work, but for consistency: in 695 it basically ignores ENOTSUP and EACCES (sic).

ENOTSUP: if there is not xattr support at all (== even worse than only supporting small to medium xattrs), it stays silent (no warning log entry) and does not set warning rc.

EACCES: not sure about that one but sounds like worth a warning at least too and not silently ignoring it.
We should discuss about that one though, I think that could be maybe borg running as normal user with no permissions to set security related stuff (?). OTOH, we have special ACL support for posix acls. Not sure what the issue was with EACCES and security.*. For normal file mode (st_mode) and user/group, I think we currently also silently ignore them when we do not run as root, but just as a normal user who can not create some mode and files for other users/groups.

Abogical:
Support for other errors have been made, I'll make a testsuite for them soon.

Abogical:
Tests added.

Issue #2256, opened 2017-03-05
Please enable IPv6 addresses for remote repositories


wararjey:
Hello,

I was trying to specify a remote borg repo as something like that
ssh://user@[ipv6_address_string]:/path/to/repo
(and tried with leaving out the ssh:// in front, as well)

borg then somehow tried to read the ipv6 address as a hostname, and the hostname lookup naturally failed.

ssh supports ipv6 address string, by putting the address into a square bracket.

I have a use case where I want to backup to a remote repository via ssh onto a machine by ipv6. However, I do not want to expose machines IPv6 address to DNS, that's why I would like to use an address literal instead of a dns domain name in the remote repo url.

Regards,
wararjey

ThomasWaldmann:
I'ld suggest you make a hosts file entry with a hostname for the host's ipv6 address.

Then you don't need to type complicated/long v6 addresses all the time when you work with borg (or ssh) and you do not need to put it into dns either. "Not exposing to DNS" sounds a bit like security by obscurity btw.

Borg's location URL parsing is already quite complex and ipv6 having ":" and "::" as valid address parts likely would complicate it significantly.

wararjey:
Thanks for the hint with the hosts file, I hadn't thought of that.

I wonder if borg could not let ssh do the hostname/ipaddress parsing for the host part of the url, then you'd get that 'feature' for free...

Regards,
wararjey

ari-s:
I'd like to have this solved: I have a case were adding the entry to /etc/hosts (or ~/.ssh/config) causes significant hassle: 

ansible-generated backup scripts for several remote backup clients that connect to remote servers where the borg repo resides. security model dictates that the borg repos are init'd from the admins machines, so all admins have to change their config. for other reasons, we'd like the backup_servers not in DNS.

ThomasWaldmann:
@ari-s watch progress of #2264 and help testing it when it is ready.

jeanrjc:
Hello, sorry to comment on this issue, but maybe I did not understand something.

The following does not work with borg 1.1.9:

```bash
$ borg check --archives-only backupNC@[2a02:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx]:/media/pi/myCloudDrive/borgBKP --debug
using builtin fallback logging configuration
35 self tests completed in 1.18 seconds
SSH command line: ['ssh', 'backupNC@2a02', 'borg', 'serve', '--umask=077', '--debug']
[...]

PR #2264, opened 2017-03-06
ipv6 address support

ogelpre:
This is a fix for #2256.


ogelpre:
this patch does not work proper. it seems, that i need to dive deeper in to the sources :/

src/borg/helpers.py
```diff
@@ -1068,7 +1068,7 @@ class Location:
     ssh_re = re.compile(r"""
         (?P<proto>ssh)://                                   # ssh://
         """ + optional_user_re + r"""                       # user@  (optional)
-        (?P<host>[^:/]+)(?::(?P<port>\d+))?                 # host or host:port
+        (?P<host>([^:/]+|\[[^/]+\]))(?::(?P<port>\d+))?     # host or host:port or [ipv6] or [ipv6]:port
```
ThomasWaldmann: "[" + "anything except /" + "]" is a rather unspecific match for a v6 address.
could this be more specific easily to avoid it going wrong somehow? like hex digits and colons?
where does the "except /" come from? I would have expected it matching everything until it finds "]"?

src/borg/helpers.py
```diff
@@ -1126,6 +1126,8 @@ def normpath_special(p):
             self.proto = m.group('proto')
             self.user = m.group('user')
             self.host = m.group('host')
+            if isinstance(self.host, str):
```
ThomasWaldmann: what can it be except a str? did you want to check for "is not None"?

src/borg/testsuite/helpers.py
```diff
@@ -45,6 +45,12 @@ def test_ssh(self, monkeypatch):
             "Location(proto='ssh', user='user', host='host', port=1234, path='/some/path', archive=None)"
         assert repr(Location('ssh://user@host/some/path')) == \
             "Location(proto='ssh', user='user', host='host', port=None, path='/some/path', archive=None)"
+        assert repr(Location('ssh://user@[::]:1234/some/path::archive')) == \
+            "Location(proto='ssh', user='user', host='::', port=1234, path='/some/path', archive='archive')"
+        assert repr(Location('ssh://user@[::]:1234/some/path')) == \
+            "Location(proto='ssh', user='user', host='::', port=1234, path='/some/path', archive=None)"
+        assert repr(Location('ssh://user@[::]/some/path')) == \
+            "Location(proto='ssh', user='user', host='::', port=None, path='/some/path', archive=None)"
```
ThomasWaldmann: good that you have some tests for "::" (as this potentially could conflict with our repo::archive separator, but maybe also test v6 addrs that are not just "::".

src/borg/helpers.py
ogelpre: The "except /" comes from just extending the host expression. I'll write a more specific match for an ipv6 address. But how specific should it be? I think this is to specific:
> (([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))

Shouldn the host expression also be more specific? 
> ([a-zA-Z0-9][a-zA-Z0-9-_]{1,61}[a-zA-Z0-9]\.)+([a-zA-Z0-9][a-zA-Z0-9-]{1,61}[a-zA-Z0-9])

src/borg/helpers.py
ogelpre: you're right.

src/borg/helpers.py
ogelpre: Enough jokes. I suggest we should use [0-9a-fA-F:\.]+ for ipv6 addresses and maybe  [a-zA-Z0-9-_\.]+ for hosts.

ThomasWaldmann: sounds reasonable.

ThomasWaldmann:
can you update your PR based on the feedback?

ThomasWaldmann:
@ogelpre can you update your PR based on the feedback?

ThomasWaldmann:
@ogelpre are you still with us? (see above)

ogelpre:
sorry, i was very busy in the past weeks. i'll update the PR in the next view hours.

src/borg/helpers.py
```diff
@@ -1111,7 +1111,7 @@ class Location:
     scp_re = re.compile(r"""
         (
             """ + optional_user_re + r"""                   # user@  (optional)
-            (?P<host>([^:/]+|\[[^/]+\])):                   # host: (don't match / in host to disambiguate from file:)
+            (?P<host>([^:/]+|\[[0-9a-fA-F:.]+\])):          # host: (don't match / or [ipv6] in host to disambiguate from file:)
```
ThomasWaldmann: "don't match / or [ipv6]" sounds wrong.

src/borg/testsuite/helpers.py
```diff
@@ -65,6 +65,24 @@ def test_ssh(self, monkeypatch):
             "Location(proto='ssh', user='user', host='::', port=1234, path='/some/path', archive=None)"
         assert repr(Location('ssh://user@[::]/some/path')) == \
             "Location(proto='ssh', user='user', host='::', port=None, path='/some/path', archive=None)"
+        assert repr(Location('ssh://user@[2001:db8::]:1234/some/path::archive')) == \
+            "Location(proto='ssh', user='user', host='2001:db8::', port=1234, path='/some/path', archive='archive')"
```
ThomasWaldmann: hmm, guess our ssh code reassembles this like `ssh -p port user@host ...`, so if the ipv6 addr has no [] any more, does this actually work?

ogelpre:
At the moment ipv6 connection does not work:
> [...]
>  File "/home/bene/.local/lib64/python3.4/site-packages/borgbackup-1.1.0b5.dev67+ng7513ca3.d20170423-py3.4-linux-x86_64.egg/borg/helpers.py", line 1130, in __init__
>     raise ValueError
> 
> Platform: Linux denkblock 4.9.6-gentoo-r1 #3 SMP PREEMPT Sun Feb 26 12:02:55 CET 2017 x86_64 Intel(R) Core(TM) i7 CPU L 640 @ 2.13GHz
> Linux: Gentoo Base System 2.3 
> Borg: 1.1.0b5.dev67+ng7513ca3.d20170423  Python: CPython 3.4.5
> PID: 26417  CWD: /home/ogelpre/bin
> sys.argv: ['/home/ogelpre/.local/bin/borg', 'prune', '-v', 'ogelpre@[fc00::1]:/home/ogelpre/denkblock', '--keep-daily=28', '--keep-weekly=12', '--keep-monthly=12', '--prefix', 'var-tmp-']
> SSH_ORIGINAL_COMMAND: None

The error is raised on line 1130:
```
         if not self.parse(self.orig):
             raise ValueError
```

I think the error is on line 1137:
`         m = self.env_re.match(text)`

I do not understand the problem at the moment.

src/borg/testsuite/helpers.py
ogelpre: ssh does not work with brackets:
```
$ ssh [2001:db8::]
ssh: Could not resolve hostname [2001:db8::]: Temporary failure in name resolution
$ ssh 2001:db8::
ssh: connect to host 2001:db8:: port 22: Network is unreachable
```
But scp needs brackets because : is used as separator.

src/borg/helpers.py
ogelpre: you're right. i'll remove 'or [ipv6]'.

src/borg/testsuite/helpers.py
ThomasWaldmann: ah, it seems just some of the ssh options want the square brackets (they also need a port), but the ssh target itself does not.

ThomasWaldmann:
doing some tests:
```
(borg-env)tw@tux:~/w/borg$ borg init -e none ssh://tw@[::1]/~/w/borg/repo6
# worked!
```

ThomasWaldmann:
```
diff --git a/src/borg/helpers.py b/src/borg/helpers.py
index 6d50dd2..f938562 100644
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1095,7 +1095,7 @@ class Location:
     def __init__(self, text=''):
         self.orig = text
         if not self.parse(self.orig):
-            raise ValueError
+            raise ValueError('parse failed: %s' % self.orig)
 
     def parse(self, text):
```

ThomasWaldmann:
Last remaining issue in 1.0.11rc1 milestone: https://github.com/borgbackup/borg/milestone/38

ThomasWaldmann:
ping :)

ThomasWaldmann:
continued in #2555

Issue #2268, opened 2017-03-07
Improve documentation for --keep-exclude-tags


edgimar:
The current documentation for the `--keep-exclude-tags` option consists of this one "sentence" in the Usage section of the documentation:

> keep tag objects (i.e.: arguments to –exclude-if-present) in otherwise excluded caches/directories
 
This leaves a lot to be understood (I had to read through the source to figure out what it _actually_ does).  Perhaps a bit more detail on this option should be included in the documentation (along with `--exclude-if-present`) in the [Description](http://borgbackup.readthedocs.io/en/latest/usage.html#id48) section.

ThomasWaldmann:
@edgimar you could make a PR with a better description.

PR #2269, opened 2017-03-07
docs:  improve --exclude-if-present and --keep-exclude-tags

edgimar:
This is to address issue #2268

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2254,6 +2254,15 @@ class Archiver:
         '\*/.bundler/gems' to get the same effect. See ``borg help patterns`` for
         more information.
 
+        In addition to using ``--exclude`` patterns, it is possible to use
+        ``--exclude-if-present`` to specify the name of a filesystem object (e.g. a file
+        or folder name) which, when contained within another folder, will prevent the
+        containing folder from being backed up.  By default, the containing folder and
+        all of its contents will be omitted from the backup.  If, however, you wish to
+        only include the objects specified by ``--exclude-if-present`` in your backup,
+        and not include any other contents of the containing folder, this can be enabled
+        through using the ``--keep-exclude-tags`` option.
+
         Item flags
         ++++++++++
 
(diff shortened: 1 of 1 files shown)


docs/usage/create.rst.inc
```diff
@@ -39,7 +39,7 @@ Exclusion options
     ``--exclude-if-present NAME``
```
enkore: Thanks, but these edits need to go into src/borg/archiver.py --- see note ` .. IMPORTANT: this file is auto-generated from borg's built-in help, do not edit!` at the beginning of these files.

docs/usage/create.rst.inc
```diff
@@ -102,6 +102,8 @@ exclude foo/.bundler/gems. In borg it will not, you need to use --exclude
 '\*/.bundler/gems' to get the same effect. See ``borg help patterns`` for
 more information.
 
+In addition to using ``--exclude`` patterns, it is possible to use ``--exclude-if-present`` to specify the name of a filesystem object (e.g. a file or folder name) which, when contained within another folder, will prevent the containing folder from being backed up.  By default, the containing folder and all of its contents will be omitted from the backup.  If, however, you wish to only include the the objects specified by ``--exclude-if-present`` in your backup, and not include any other contents of the containing folder, this can be enabled through using the ``--keep-exclude-tags`` option.
```
enkore: `you wish to only include the the objects specified by` -> `you wish to include only the objects by` (double "the")

src/borg/archiver.py
```diff
@@ -2254,6 +2254,15 @@ def process_epilog(epilog):
...
         more information.
 
+        In addition to using ``--exclude`` patterns, it is possible to use
+        ``--exclude-if-present`` to specify the name of a filesystem object (e.g. a file
+        or folder name) which, when contained within another folder, will prevent the
+        containing folder from being backed up.  By default, the containing folder and
+        all of its contents will be omitted from the backup.  If, however, you wish to
+        only include the objects specified by ``--exclude-if-present`` in your backup,
+        and not include any other contents of the containing folder, this can be enabled
+        through using the ``--keep-exclude-tags`` option.
```
ThomasWaldmann: well, this is explaining what it does, but not much the motivation to do so.
- for tag files, that is just not to lose the tags (the information about not wanting to back up this specific dir) in case you have to recover from a backup.
- for tag directories with actual content, there is the .git/ usecase: you do not want to back up all your workdir checkouts, but you do want to backup your repos.

enkore: ^- would be more something for the `Examples` section (docs/usage.rst), if one wants to add it, not the main description.

edgimar: Not sure about the examples section, because besides this text (and the short description) there is no documentation at all about these options.  I see this as explaining in detail how the options work, not as examples.

And yes, motivation is not included, and while I agree that it could be nice, I'm not motivated ;) to add it.

PR #2270, opened 2017-03-08
Address SSH batch mode in docs (fixes #2202)

### LINKED ISSUE #2202: Provide support for ssh -oBatchMode=yes
If there is a problem with the ssh key, ssh will fall back to prompting for a password. As this can often be unwanted in scripted backup settings, it would be helpful if borg had an option to call ssh with `-oBatchMode=yes` to disable this fallback. This also could be achieved by putting 
```
Host <backupHost>
  BatchMode yes
```
into `~/.ssh/config`, but having a command line switch for this would be nicer in my opinion.

milkey-mouse:
This adds a snippet in the `Additional notes` section of the docs about why to use the `-oBatchMode=yes` flag in automated backup scripts calling `ssh`.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -991,7 +991,6 @@ SSH batch mode
 ~~~~~~~~~~~~~~
 
 When running |project_name| using an automated script, ``ssh`` might still ask for a password,
-even if there is an SSH key for the target server. To make scripts more robust, set the
-``BORG_RSH`` environment variable in your script with ``-oBatchMode=yes``::
+even if there is an SSH key for the target server. Use this to make scripts more robust::
 
     export BORG_RSH='ssh -oBatchMode=yes'
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -986,3 +986,12 @@ for e.g. regular pruning.
 Further protections can be implemented, but are outside of Borgs scope. For example,
 file system snapshots or wrapping ``borg serve`` to set special permissions or ACLs on
 new data files.
+
+SSH batch mode
+~~~~~~~~~~~~~~
+
+When running |project_name| using an automated script, ``ssh`` might still ask for a password,
+even if there is an SSH key for the target server. To make scripts more robust, set the
+``BORG_RSH`` environment variable in your script with ``-oBatchMode=yes``::
+
+    export BORG_RSH='ssh -oBatchMode=yes'


enkore:
For future PRs please PR against master unless there is a reason not to :) (see http://borgbackup.readthedocs.io/en/latest/development.html#contributions )

docs/usage.rst
```diff
@@ -986,3 +986,12 @@ for e.g. regular pruning.
 Further protections can be implemented, but are outside of Borgs scope. For example,
 file system snapshots or wrapping ``borg serve`` to set special permissions or ACLs on
 new data files.
+
+SSH batch mode
+~~~~~~~~~~~~~~
+
+When running |project_name| using an automated script, ``ssh`` might still ask for a password,
+even if there is an SSH key for the target server. To make scripts more robust, set the
```
ThomasWaldmann: why would it ask for a pw if there is a (valid) key?

docs/usage.rst
```diff
@@ -986,3 +986,12 @@ for e.g. regular pruning.
...
+``BORG_RSH`` environment variable in your script with ``-oBatchMode=yes``::
```
ThomasWaldmann: s/with .../like this:/

docs/usage.rst
enkore: When the key is encrypted.

ThomasWaldmann: yes, but who uses a passphrase-protected key for automated stuff?

milkey-mouse: SSH would also ask for a password when public key access is disabled serverside. The server negotiates the allowed authenticaton methods, so the server admin could enable TOTP 2FA on the server or disable public key access (not sure why one would do that one though) without warning if you don't control the server's ssh settings (i.e. you are not the owner).

It's not that important of an option to use (because it's only used in unattended scripts, and only necessary when the SSH authentication requirements might change/increase) but it's still a good measure in scripts.

milkey-mouse:
@enkore Sorry I didn't PR against `master`, I was accidentally using the 1.0 docs as my reference for contributions, which say to PR against `1.0-maint` unless the change is urgent or a big new feature.

ThomasWaldmann:
We could just keep this PR for 1.0-maint, but we also need to apply it to master.

PR #2271, opened 2017-03-08
Add warning about running build_usage on Python >3.4 (fixes #2123)

### LINKED ISSUE #2123: docs: borg list: guaranteed hashes
`hashlib.algorithms_guaranteed` was expanded with Python 3.6, so we should keep in mind to build the docs with 3.4, since that is our baseline.

```diff
@@ -84,14 +84,22 @@ The following keys are available for --format:
  - isomtime
  - isoctime
  - isoatime
 
+ - blake2b
+ - blake2s
  - md5
  - sha1
  - sha224
  - sha256
  - sha384
+ - sha3_224
+ - sha3_256
+ - sha3_384
+ - sha3_512
  - sha512
+ - shake_128
+ - shake_256
 
  - archiveid
  - archivename
  - extra: prepends {source} with " -> " for soft links and " link to " for hard links
```

milkey-mouse:
Python 3.6 added some new guaranteed hashing algorithms that will show up as available in the docs when they are built with >=3.6 even though the baseline for support is currently Python 3.4. We don't want to include hashing algorithms that might not exist on certain builds of borg in the docs, and running build_usage with Python 3.4 instead will fix this.

I wasn't sure where the diff attached to #2123 came from (which command? API docs?), but I'm pretty sure these changes will suffice.

It might be better to have the link point to the part of the Python 3.6 changelog where it tells about the new guaranteed algorithms instead of to the issue...

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -297,7 +297,10 @@ Checklist:
 - update ``CHANGES.rst``, based on ``git log $PREVIOUS_RELEASE..``
 - check version number of upcoming release in ``CHANGES.rst``
 - verify that ``MANIFEST.in`` and ``setup.py`` are complete
-- ``python setup.py build_api ; python setup.py build_usage ; python setup.py build_man`` and commit
+- ``python setup.py build_api ; python setup.py build_usage ; python
+  setup.py build_man`` and commit (be sure to build with Python 3.4 as
+  Python 3.6 added `more guaranteed hashing algorithms
+  <https://github.com/borgbackup/borg/issues/2123>`_)
 - tag the release::
 
     git tag -s -m "tagged/signed release X.Y.Z" X.Y.Z


ThomasWaldmann:
Maybe we could just overwrite the guaranteed hashes list with whatever we would get from our minimum required python version?

enkore:
Only conditionally (when building docs), otherwise it would disable hashes that are available on the system.

ThomasWaldmann:
@enkore yes, ofc I meant only for building the docs.

milkey-mouse:
@enkore question: where did the diff in #2123 come from? I `grep`ped the entire source and couldn't find the string `The following keys are available for --format`. Sorry if I'm missing something obvious

enkore:
I ran build_usage with Python 3.6 (python setup.py build_usage), the diff in #2123 is the diff between the files generated by that and the tree / master.

docs/development.rst
```diff
@@ -297,7 +297,10 @@ Checklist:
 - update ``CHANGES.rst``, based on ``git log $PREVIOUS_RELEASE..``
 - check version number of upcoming release in ``CHANGES.rst``
 - verify that ``MANIFEST.in`` and ``setup.py`` are complete
-- ``python setup.py build_api ; python setup.py build_usage ; python setup.py build_man`` and commit
+- ``python setup.py build_api ; python setup.py build_usage ; python
+  setup.py build_man`` and commit (be sure to build with Python 3.4 as
+  Python 3.6 added `more guaranteed hashing algorithms
```
ThomasWaldmann: hmm, just noticed: it recommends 3.4 because 3.6 added.... - so, what about 3.5?

enkore: 3.5 didn't change anything there, so 3.5 should be ok, too. Just not 3.6/main/system Python (depending on distro).

PR #2277, opened 2017-03-09
Securely erase config file (fixes #2257)

### LINKED ISSUE #2257: When changing passphrase, is the old one properly erased ?
I didn't find the relevant code, but looking at the `key` entry in my repository's `config` file, I assume that it's the encrypted key of the repository.

If I change the repository's passphrase, will the old value in `config` be securely erased, so that the old one cannot be recovered by reading the unallocated sector on the disk ? Will the config file's rewrite always overwrite the old value ?

I find it very important for security to erase the old key, as changing a too weak or compromised passphrase would be useless if the old version of the encrypted key could be recovered.

Thank you.

milkey-mouse:
This fixes issue #2257 where the old password would still be on the disk after replacing the config file.
Here's how I tested my solution (and maybe a way to implement a test for this, if necessary):

First, create a small loopback image (or partition) to test on. Be sure to use a non-CoW filesystem as they, well, copy on writes:

    $ dd if=/dev/zero of=~/testfs.img bs=1M count=128
    $ mkfs.ext4 ~/testfs.img
    $ sudo mount ~/testfs.img /mnt

Next, initialize a `repokey` encrypted borg repository on the filesystem:

    $ borg init --encryption=repokey /mnt/test-repo
    Your passphrase (between double-quotes): "test1"

Get the first line of the key directly from the filesystem (trimmed for readability):

    $ grep key /mnt/test-repo/config
    key = hqRzYWx... (old key)

Now search for the key in the raw filesystem image:

    $ grep -a -b "key = " ~/testfs.img
    142607523:key = hqRzYWx... (old key)

The current key will obviously always be on the disk. With a little more work it could be extracted.

Next, replace the repo's key with a new one:

    $ borg key change-passphrase /mnt/test-repo
    Your passphrase (between double-quotes): "test2"

Before this change, both keys would have been hidden on disk afterwards:

    $ grep -a -b "key = " ~/testfs.img  # old behavior
    142607523:key = hqRzYWx... (old key)
    943839677:key = hqlhbGd... (new key)

Now that it securely erases the old key, only the new one will show up:

    $ grep -a -b "key = " ~/testfs.img  # new behavior
    943839677:key = hqlhbGd... (new key)

Previously, because of the atomic file writing code (`SaveFile`) it would end up creating a new file and leaving the contents of the old one on disk. Most non-CoW filesystems, however, edit the *contents* of a file in place. By linking the file and filling it with random data after the new one has been successfully written, we preserve the atomicity while still safely deleting the file.


milkey-mouse:
Let me know if this needs an entry in the docs (under the FAQ maybe?). I can't decide whether it's worth the effort to retype it after I messed up Git and deleted my first version.

src/borg/repository.py
```diff
@@ -182,9 +182,29 @@ def create(self, path):
...
     def save_config(self, path, config):
         config_path = os.path.join(path, 'config')
+        old_config_path = os.path.join(path, 'config.old')
+
+        if os.path.isfile(old_config_path):
+            logger.warning("Old config file not securely erased on previous config update")
+            self.secure_erase_config(old_config_path)
+
+        if os.path.isfile(config_path):
+            os.link(config_path, old_config_path)
```
ThomasWaldmann: hmm, does this add a new requirement for the repo filesystem "must support hardlinks" that we did not have before?

ThomasWaldmann: maybe try this on ntfs. from os.link docs it seems that hardlink support was added for windows a while ago. ntfs has junctions, so it might even work. ntfs is quite popular on external usb disks.

but even if it works for ntfs, we could still have a try/except here and track whether making the hardlink worked.
if not, no need to secure erase it later.

src/borg/repository.py
```diff
@@ -182,9 +182,29 @@ def create(self, path):
...
+        if os.path.isfile(old_config_path):
+            self.secure_erase_config(old_config_path)
+
+    def secure_erase_config(self, path):
+        with open(path, 'r+b') as config:
+            length = os.stat(config.fileno()).st_size
+            config.write(os.urandom(length))
+            config.flush()
+            os.fsync(config.fileno())
+        os.unlink(path)
```
ThomasWaldmann: this doesn't use self, which indicates this should not be a method.
maybe we can use it for other stuff later, so rather move this to helpers module.
also, just name if secure_erase, rename "config" to "fd" and add a docstring.

src/borg/repository.py
enkore: I think upgrade requires hardlinks, but that is not the regular usage case.

try/except checking for errno in (EMLINK, EPERM) should be sufficient, and can warn in these cases that the file couldn't be erased.

ThomasWaldmann: if it could not make the link, there is nothing to erase.

src/borg/helpers.py
```diff
@@ -2218,3 +2218,12 @@ def json_dump(obj):
...
     print(json_dump(obj))
+
+def secure_erase(path):
+    """Attempt to securely erase a file by writing random data over a file before deleting it."""
+    with open(path, 'r+b') as fd:
+        length = os.stat(fd.fileno()).st_size
+        fd.write(os.urandom(length))
+        fd.flush()
+        os.fsync(fd.fileno())
+    os.unlink(path)
```
ThomasWaldmann: this file ends with no linefeed, pep8 checkers (and even github code review) complains about this, so please add one (and maybe configure your editor to behave that way).

milkey-mouse: sorry, don't know why my editor didn't add one

ThomasWaldmann:
the pep8 checker found one remaining issue:

src/borg/helpers.py:2260:1: E302 expected 2 blank lines, found 1

ThomasWaldmann:
oh, and please collapse your changesets.

ThomasWaldmann:
I fixed it myself. #2293 supercedes this one.

milkey-mouse:
@ThomasWaldmann when you say "collapse your changesets," do you mean that I should have `squash`ed those commits together?

ThomasWaldmann:
@milkey-mouse yes. squash or fixup.

I usually use git rebase -i master for this and (f)ixup all fixes into their base commit, so only a few fundamental commits are left. That can be 1 or multiple commits.

ThomasWaldmann:
I already did these changes, see the other already accepted PR, so nothing TODO left.

PR #2278, opened 2017-03-09
mount: allow / in archive names with rules that make sense

### LINKED ISSUE #2263: mount: allow / in archive names
Only seems to be forbidden because it would make the `borg mount` code a bit more complex. Parsing is not hard; extra assertion would probably be `posixpath.normpath(archive) == archive` (to rule `../../` & friends out) and it does not start with a /.

Allowing these gives a nicer, more intuitive separation of archives, e.g. `{hostname}/{date...}` is much neater with / as a separator than -_#@ (whichever) for mounting it, but also because / is considered a separator in general.

Btw., currently `.` and `..` are valid archive names; while these don't crash `mount`, they're inaccessible there.

enkore:
- POSIX_normalization(archive name) == archive name

  So foo/bar is valid,  foo/bar/, foo/./bar/, foo/../bar/ are not.

- May not begin with any number of slashes (i.e. not an absolute
  path)

- Archives may not be named "." or ".."

  This is a new restriction and previously broke borg-mount
  (it didn't crash, you just got EIO when reading the mountpoint dir)

  With this change borg-mount just ignores existing archives with
  these names.

- borg-mount is very reasonable. You can have an archive foo/bar,
  and an archive foo, if you mount the repository, you'll find
  foo/bar with the archive but also the contents of foo under
  foo/ (makes sense).

  I don't think this should be restricted. If someone does not
  want that behaviour, just don't name your archives like that.
  I think it can be even quite useful.

  Another question might be, "if I have an archive foo, that
  has a directory bar, but also an archive foo/bar, which one
  will I get?". Answer: --sort-by decides, first come first served.

  Again, this is something that doesn't need restricting. Don't
  want that behaviour (which again can be quite useful) -> don't
  name your archives that way.

Fixes #2263

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -60,28 +60,25 @@ Repository / Archive Locations
 Many commands want either a repository (just give the repo URL, see above) or
 an archive location, which is a repo URL followed by ``::archive_name``.
 
-Borg does not assume a particular meaning or structuring of archive names,
-but for use with :ref:`borg_prune` and other commands it makes sense
-to structure names with prefixes (e.g. ``{hostname}-{date}``) for the
-``--prefix`` option.
+Archive names can be almost any string, but Borg places two important restrictions
+on them:
 
-When mounting an entire repository with :ref:`borg_mount` (without the
-'versions' option) the archive names are used as paths in the
-virtual filesystem.
-
-Archive names can contain ``/`` (slash) characters, but may not begin
-nor end with a slash, and may not contain consecutive slashes.
-Further, the usual path constructs involving ``.`` and ``..`` are not permitted;
-for example, these are all invalid archive names: ``/foo``, ``foo/../bar``, ``foo/.``,
-``//foohost/share``. The archive names ``.`` and ``..`` are invalid as well,
-but otherwise there is no restriction on using dots, e.g. ``foo.example.net....myfiles``
-is valid.
-
-This is to ensure a direct mapping to file system paths in :ref:`borg_mount`.
+1. Since archives can only be filtered by ``--prefix`` (for e.g. :ref:`borg_prune`),
+   grouping needs to be prefix based. For example, when backing up multiple hosts
+   into the same repository ``{hostname}-{date}`` makes sense, while ``{date}-{hostname}``
+   will pose issues for pruning.
+2. When mounting (:ref:`borg_mount`) an entire repository, archive names are interpreted
+   as paths. Therefore archives cannot be named ".", "..", cannot start nor end with a slash,
+   and may not contain two or more consecutive slashes.
 
 Otherwise any string can be used, including whitespace, newlines, Unicode
 and so on, however, to cause less confusion and maximize portability it's
-likely a good idea to avoid special characters.
+likely a good idea to avoid special characters. Similarly backslashes are permitted,
+but should not be used to improve portability.
+
+In summary, these are valid archive names: ``foo.example.net..myfiles``,
+``foo.example.net-system@2017-04-01``, whereas these are invalid: ``/foo``,
+``foo/../bar``, ``foo/.``, ``//foohost/share``.
 
 If you have set BORG_REPO (see above) and an archive location is needed, use
 ``::archive_name`` - the repo URL part is then read from BORG_REPO.
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -60,10 +60,28 @@ Repository / Archive Locations
 Many commands want either a repository (just give the repo URL, see above) or
 an archive location, which is a repo URL followed by ``::archive_name``.
 
-Archive names must not contain the ``/`` (slash) character. For simplicity,
-maybe also avoid blanks or other characters that have special meaning on the
-shell or in a filesystem (borg mount will use the archive name as directory
-name).
+Borg does not assume a particular meaning or structuring of archive names,
+but for use with :ref:`borg_prune` and other commands it makes sense
+to structure names with prefixes (e.g. ``{hostname}-{date}``) for the
+``--prefix`` option.
+
+When mounting an entire repository with :ref:`borg_mount` (without the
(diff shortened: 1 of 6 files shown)


docs/usage_general.rst.inc
```diff
@@ -60,10 +60,28 @@ Repository / Archive Locations
...
-shell or in a filesystem (borg mount will use the archive name as directory
-name).
+Borg does not assume a particular meaning or structuring of archive names,
+but for use with :ref:`borg_prune` and other commands it makes sense
+to structure names with prefixes (e.g. ``{hostname}-{date}``) for the
+``--prefix`` option.
+
+When mounting an entire repository with :ref:`borg_mount` (without the
+'versions' option) the archive names are used as paths in the
+virtual filesystem.
```
ThomasWaldmann: well, this somehow contradicts the first paragraph.
after this proposed change, borg mount assumes that the archive name is a path (without a slash it is still a 1-level path).

src/borg/helpers.py
```diff
@@ -1061,7 +1062,7 @@ class Location:
     optional_archive_re = r"""
         (?:
             ::                                              # "::" as separator
-            (?P<archive>[^/]+)                              # archive name must not contain "/"
+            (?P<archive>[^/]{1}.*)
```
ThomasWaldmann: hmm, what is the point of having `{1}` there?

src/borg/helpers.py
```diff
@@ -29,6 +29,7 @@
 from operator import attrgetter
 from string import Formatter
 from shutil import get_terminal_size
+from posixpath import normpath
```
ThomasWaldmann: rather use import posixpath here to make it clearer below that normpath is from there and not from os.path?

ThomasWaldmann:
pr title should reflect this is a change to how borg mount works.

enkore:
merge?

ThomasWaldmann:
don't merge yet, see discussion in the ticket.

enkore:
I thought we were done with that discussion

ThomasWaldmann:
@textshell ^^

docs/usage_general.rst.inc
```diff
@@ -60,10 +60,25 @@ Repository / Archive Locations
...
-name).
+Archive names can be almost any string, but Borg places two important restrictions
+on them:
+
+1. Since archives can only be filtered by ``--prefix`` (for e.g. :ref:`borg_prune`),
+   grouping needs to be prefix based. For example, when backing up multiple hosts
+   into the same repository ``{hostname}-{date}`` makes sense, while ``{date}-{hostname}``
+   will pose issues for pruning.
+2. When mounting (:ref:`borg_mount`) an entire repository, archive names are interpreted
+   as paths. Therefore archives cannot be named ".", "..", cannot start nor end with a slash,
```
ThomasWaldmann: ... may not end with "." or ".." and can not contain "/./" or "/../", ...

docs/usage_general.rst.inc
```diff
@@ -60,10 +60,25 @@ Repository / Archive Locations
...
+   and may not contain two or more consecutive slashes.
+
+Otherwise any string can be used, including whitespace, newlines, Unicode
+and so on, however, to cause less confusion and maximize portability it's
+likely a good idea to avoid special characters. Similarly backslashes are permitted,
+but should not be used to improve portability.
```
ThomasWaldmann: Could we just disallow backslashes?

src/borg/helpers.py
```diff
@@ -1178,6 +1177,10 @@ def normpath_special(p):
             return True
         return False
 
+    @staticmethod
+    def archive_valid(archive):
+        return not archive or (archive == posixpath.normpath(archive) and archive not in ('..', '.') and '::' not in archive)
```
ThomasWaldmann: handle the "not archive" in the caller, not here.

src/borg/helpers.py
```diff
@@ -1230,8 +1233,8 @@ def validator(text):
 
 def archivename_validator():
     def validator(text):
-        if '/' in text or '::' in text or not text:
-            raise argparse.ArgumentTypeError('Invalid repository name: "%s"' % text)
+        if not Location.archive_valid(text):
```
ThomasWaldmann: archive_valid() currently accepts empty archive name, which was explicitly forbidden before, see 2 lines above.

docs/usage_general.rst.inc
enkore: #2290.

src/borg/helpers.py
enkore: https://docs.python.org/3/library/os.path.html#os.path.normpath

> However, you can also import and use the individual modules if you want to manipulate a path that is always in one of the different formats. They all have the same interface: posixpath for UNIX-style paths

src/borg/fuse.py
```diff
@@ -138,6 +137,23 @@ def _create_dir(self, parent):
...
 
+    def _create_dir_parents(self, path, parent=1):
+        """Create directory tree (equivalent to mkdir --parents)."""
+        segments = path.split(b'/')
+        segments.reverse()
+        inode = parent
+        while segments:
+            try:
+                inode = self.contents[inode][segments[-1]]
+            except KeyError:
```
enkore: NB. specifically this is a common pattern in the entire FUSE code via `try: _find_inode except KeyError:`.

enkore: > What is the right response here if self.contents[inode] throws?

raise Exception, _exit(2).

src/borg/helpers.py
ThomasWaldmann: what I meant is that a falsy archive name is not a valid archive name.

the only case where this is used in a overall-valid context is if we have a repo location without an archive name, but that should be handled outside of an archive_valid function.

ThomasWaldmann:
IIRC the outcome of the discussion was "not wanted without changes" not "not wanted".

PR #2254, opened 2017-03-04
Docs: one link per distro in the installation page

florenthemmi:
BorgBackup 1.0.10 is only available in the experimental distribution for the moment.
Works perfectly for me. So I suggest to add this distribution as well.

http://metadata.ftp-master.debian.org/changelogs/main/b/borgbackup/borgbackup_1.0.10-2_changelog

PR #2284, opened 2017-03-10
picked cherries for 1.0-maint

enkore:
Changes from

#2156
#2159
#2171
#2181
#2187
#2198
#2206
#2224
#2254

-

#2199 
#2258 

-

 #2184

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -542,7 +542,7 @@ Number of files: {0.stats.nfiles}'''.format(
                 raise ChunksIndexError(cid)
             except Repository.ObjectNotFound as e:
                 # object not in repo - strange, but we wanted to delete it anyway.
-                if not forced:
+                if forced == 0:
                     raise
                 error = True
 
@@ -564,14 +564,14 @@ Number of files: {0.stats.nfiles}'''.format(
                 except (TypeError, ValueError):
                     # if items metadata spans multiple chunks and one chunk got dropped somehow,
                     # it could be that unpacker yields bad types
-                    if not forced:
+                    if forced == 0:
                         raise
                     error = True
             if progress:
                 pi.finish()
         except (msgpack.UnpackException, Repository.ObjectNotFound):
             # items metadata corrupted
-            if not forced:
+            if forced == 0:
                 raise
             error = True
         # in forced delete mode, we try hard to delete at least the manifest entry,
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -23,7 +23,7 @@ from .helpers import Error, uid2user, user2uid, gid2group, group2gid, bin_to_hex
 from .platform import acl_get, acl_set
 from .chunker import Chunker
 from .hashindex import ChunkIndex
-from .repository import Repository
+from .repository import Repository, LIST_SCAN_LIMIT
 
 import msgpack
 
@@ -882,7 +882,7 @@ class ArchiveChecker:
         self.chunks = ChunkIndex(capacity)
         marker = None
         while True:
-            result = self.repository.list(limit=10000, marker=marker)
+            result = self.repository.list(limit=LIST_SCAN_LIMIT, marker=marker)
             if not result:
                 break
             marker = result[-1]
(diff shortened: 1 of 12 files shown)


borg/archive.py
```diff
@@ -501,7 +501,10 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
             try:
                 xattr.setxattr(fd or path, k, v, follow_symlinks=False)
             except OSError as e:
-                if e.errno not in (errno.ENOTSUP, errno.EACCES):
+                if e.errno == errno.E2BIG:
+                    logger.warning('%s: Value or key of extended attribute %s is too big for this filesystem' %
+                                   (path, k.decode()))
+                elif e.errno not in (errno.ENOTSUP, errno.EACCES):
```
ThomasWaldmann: this is a bit strange now: warns if we could not extract xattrs (because they were too big), but says nothing if we could not extract at all (because not supported or not permitted). this was fixed in master.

borg/helpers.py
```diff
@@ -87,11 +91,13 @@ class PlaceholderError(Error):
...
 
 def check_extension_modules():
-    from . import platform
+    from . import platform, compress
     if hashindex.API_VERSION != '1.0_01':
         raise ExtensionModuleError
     if chunker.API_VERSION != '1.0_01':
         raise ExtensionModuleError
+    if compress.API_VERSION != '1.0_01':
+        raise ExtensionModuleError
```
ThomasWaldmann: oh? was this missing still in 1.0?

borg/remote.py
```diff
@@ -46,6 +46,27 @@ def os_write(fd, data):
...
+                max_str_len=MAX_OBJECT_SIZE,  # a chunk or other repo object
+                )
+    if kind == 'server':
+        args.update(dict(max_array_len=100,  # misc. cmd tuples
+                         max_map_len=100,  # misc. cmd dicts
+                         ))
+    elif kind == 'client':
+        args.update(dict(max_array_len=LIST_SCAN_LIMIT,  # result list from repo.list() / .scan()
+                         max_map_len=100,  # misc. result dicts
+                         ))
```
ThomasWaldmann: this might be not the tightest limits as 1.0 does not use the dict-based rpc protocol, but should work I guess.

borg/testsuite/archiver.py
```diff
@@ -245,7 +245,7 @@ def cmd(self, *args, **kw):
         return output
 
     def create_src_archive(self, name):
-        self.cmd('create', self.repository_location + '::' + name, src_dir)
+        self.cmd('create', '--compression=lz4', self.repository_location + '::' + name, src_dir)
```
ThomasWaldmann: hmm, do we want that in 1.0?

borg/archive.py
Abogical: You need to cherrypick #2258 for this.

enkore:
Okay, so I'm going to pick up #2199 and #2258 as well

borg/testsuite/archiver.py
enkore: Doesn't make much of a difference

borg/helpers.py
enkore: Seems so

enkore:
And #2184 (but neither #2204 nor #2255, correct)?

ThomasWaldmann:
@enkore yup. let's test the fixed async processing in master / 1.1b first for a while, then we can think whether we want it in 1.0 also.

borg/archiver.py
```diff
@@ -503,9 +503,23 @@ def do_rename(self, args, repository, manifest, key, cache, archive):
...
+                try:
+                    del manifest.archives[archive_name]
+                except KeyError:
+                    raise Archive.DoesNotExist(archive_name)
+                logger.info('Archive deleted.')
+                manifest.write()
+                # note: might crash in compact() after committing the repo
+                repository.commit()
+                logger.info('Done. Run "borg check --repair" to clean up the mess.')
+                return self.exit_code
```
enkore: I changed the error reporting here to match the the singular-behaviour that `borg delete` normally has in 1.0.x (RC=2 if archive not found, no plural messages etc.)

Issue #2279, opened 2017-03-09
Allow specifying default pattern style inside --patterns-from file


edgimar:
Currently the default pattern style used for options like `--patterns-from` is fnmatch (as indicated [here](http://borgbackup.readthedocs.io/en/stable/usage.html#borg-help-patterns)).  Sometimes a different default would be desirable.  It would therefore be nice if it were possible to specify a default to be used within the scope of a particular `--patterns-from` file, similar to how in a mercurial `.hgignore` file you can specify the format you wish to use at the beginning of the file using something like `syntax: regexp`.

ThomasWaldmann:
As we have patterns change in 1.0 also, I added this to next 1.0 milestone, so that we decide about this.

enkore:
#2288.

PR #2288, opened 2017-03-11
support setting the pattern style default in patterns file

ThomasWaldmann:
Issue #2279

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1663,7 +1663,7 @@ class Archiver:
 
         Note that the default pattern style for `--pattern` and `--patterns-from` is
         shell style (`sh:`), so those patterns behave similar to rsync include/exclude
-        patterns.
+        patterns. The pattern style can be set via the `P` prefix.
 
         Patterns (`--pattern`) and excludes (`--exclude`) from the command line are
         considered first (in the order of appearance). Then patterns from `--patterns-from`
@@ -1671,6 +1671,8 @@ class Archiver:
 
         An example `--patterns-from` file could look like that::
 
+            # "sh:" pattern style is the default, so the following line is not needed:
+            P sh
             R /
             # can be rebuild
             - /home/*/.cache
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -391,18 +391,23 @@ def parse_timestamp(timestamp):
         return datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S').replace(tzinfo=timezone.utc)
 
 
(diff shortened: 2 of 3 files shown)


enkore:
Hmm, for the frequent cases single-char "commands" make sense, but I think I'd be better to have more verbose commands here.

E.g.

```
    R /
    p sh
    - /home/*/.cache
    - /home/*/Downloads
    + /home/susan
    - /home/*
```

vs.

```
    root /
    pattern-style sh
    - /home/*/.cache
    - /home/*/Downloads
    + /home/susan
    - /home/*
```

ThomasWaldmann:
@enkore yup, I'll add a changeset here. It is not quite in scope of this PR, but just for the sake of easy backporting.

ThomasWaldmann:
so, anything else (besides docs)?

ThomasWaldmann:
Hmm, the changes only deal with a patterns **file**. Do we need this mechanism for cmdline args also?

@leo-b can you review?

enkore:
That would make command line parsing stateful, not just order-dependent... do we really want that?

leo-b:
BTW: Is there really a need to change the default pattern style _inside a file for just a part_ of the file?

If it is sufficient to specify a per file default, I'd propose this commandline syntax:
``--patterns-from re:/etc/borg/mypatterns.conf``
(postfix uses a similar way to specify the [lookup table style](http://www.postfix.org/DATABASE_README.html)

It is intuitive and clearly defines the scope for this default.

ThomasWaldmann:
@leo-b i think it is useful being able to switch in the file.

- people will like to use sh: syntax mostly, but maybe need re: for more complicated matching
- people will like to have only one such patterns file, so that everything is in one place

ThomasWaldmann:
@leo-b concerning warning about fnmatch: guess that's nothing new and could happen before this change also by giving the bad patterns combinations on the commandline. So: separate PR for that (please help, if you like).

enkore:
I think it's better to set the pattern style in the file itself than on the command line, because otherwise pattern style and the patterns and the files are not synchronized - you have to look in two places to know how they are interpreted.

But maybe it's a good idea to say one style per file is enough (i.e. have to set pattern style at most once before any patterns appear in the file)?

ThomasWaldmann:
@enkore see above. if we restrict to 1 style per file, what happens if you first use sh: and then notice that you can't implement some exclude with it, rewrite the whole thing? that sounds like a rather big inconvenience, considering it did not take much code (in fact: likely none) to implement switching as often as one likes.

enkore:
convinced

ThomasWaldmann:
collapsed changesets, will merge after travis.

PR #2298, opened 2017-03-13
docs/security: reiterate that RPC in Borg does no networking (in Borg itself)

enkore:
This was kinda implied with the previous wording, but it should be said more explicitly.

### DIFF
--- a/docs/internals/security.rst
+++ b/docs/internals/security.rst
@@ -255,9 +255,21 @@ over an encrypted SSH channel (the system's SSH client is used for this
 by piping data from/to it).
 
 This means that the authorization and transport security properties
-are inherited from SSH and the configuration of the SSH client
-and the SSH server. Therefore the remainder of this section
-will focus on the security of the RPC protocol within Borg.
+are inherited from SSH and the configuration of the SSH client and the
+SSH server -- Borg RPC does not contain *any* networking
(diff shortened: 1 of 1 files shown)


docs/internals/security.rst
```diff
@@ -255,9 +255,21 @@ over an encrypted SSH channel (the system's SSH client is used for this
...
-are inherited from SSH and the configuration of the SSH client
-and the SSH server. Therefore the remainder of this section
-will focus on the security of the RPC protocol within Borg.
+are inherited from SSH and the configuration of the SSH client and the
+SSH server -- Borg RPC does not contain *any* networking
+code. Networking is done by the SSH client running in a separate
+process, Borg only communicates over the standard pipes (stdout,
+stderr and stdin) with this process. This also means that Borg doesn't
+have to directly use a SSH client (or SSH at all). For example,
+``sudo`` or ``qrexec`` could be used as an intermediary.
```
ThomasWaldmann: is qrexec widely known enough to serve as a useful example?

enkore: sudo -> isolation in the same OS, qrexec escalates that to isolation across operating systems / Xen Domains which is why I put it in -- to highlight that this mechanism is so basic in concept and implementation that there are a lot of possibilities to separate concerns and privileges.

PR #2309, opened 2017-03-15
clamp (nano)second values to unproblematic range, fixes #2304

### LINKED ISSUE #2304: way-off fs timestamps crash msgpack
https://github.com/msgpack/msgpack-python/issues/221

ThomasWaldmann:
filesystem -> clamp -> archive (create)

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -665,7 +665,7 @@ def timestamp(s):
     """Convert a --timestamp=s argument to a datetime object"""
     try:
         # is it pointing to a file / directory?
-        ts = os.stat(s).st_mtime
+        ts = safe_s(os.stat(s).st_mtime)
         return datetime.utcfromtimestamp(ts)
     except OSError:
         # didn't work, try parsing as timestamp. UTC, no TZ, no microsecs support.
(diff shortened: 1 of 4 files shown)


src/borg/helpers.py
```diff
@@ -818,12 +818,34 @@ def SortBySpec(text):
     return text.replace('timestamp', 'ts')
 
 
+# Not too rarely, we get crappy timestamps from the fs, that overflow some computations.
+# As they are crap anyway, nothing is lost if we just clamp them to the max valid value.
+# msgpack can only pack uint64. datetime is limited to year 9999.
```
ThomasWaldmann: better clamp to (positive) max_int64?

when does python switch from simple into to long int?

enkore: The internal representation is unified (it is always a variable-length integer); the item length is the native size of the platform.

ThomasWaldmann: ok, but assuming we are on a 64bit platform, the native datatype being used is then int64, not uint64.

so, guess it is more efficient if we clamp to 0 .. 2**63 - 1?

enkore:
> archive -> clamp -> filesystem (extract, maybe old broken archives?)

I'm not sure that this part is necessary; we couldn't store these large timestamps before, but filesystems have a large variance in the supported ranges. The os.utime call is already wrapped in OSError, so there's likely nothing needed here.

> archive -> clamp -> FUSE (mount, maybe old broken archives?)

I guess this depends on the size of the st_Xtime and st_Xtime_ns fields. At least the latter ought to be >= 64 bits, so no problems there. If the former can be 32 bits it may be problematic, but evidently llfuseis liberal here with types (or the code path has never been used) -- since we're passing `float`s to these fields.

src/borg/helpers.py
enkore: IIRC the sign is encoded separately from the value itself

ThomasWaldmann:
@enkore ok. i'll remove the unneeded safe_X calls for fuse and extract.

ThomasWaldmann:
assuming tests are ok: collapse, merge?

PR #2312, opened 2017-03-16
clamp (nano)second values to unproblematic range, fixes #2304

### LINKED ISSUE #2304: way-off fs timestamps crash msgpack

ThomasWaldmann:
filesystem -> clamp -> archive (create)

(cherry picked from commit b7a17a6db76283c9d9d6f246831c3010fbb35fe2)

backport of #2309.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -19,7 +19,7 @@ from . import xattr
 from .helpers import Error, uid2user, user2uid, gid2group, group2gid, bin_to_hex, \
     parse_timestamp, to_localtime, format_time, format_timedelta, remove_surrogates, \
     Manifest, Statistics, decode_dict, make_path_safe, StableDict, int_to_bigint, bigint_to_int, \
-    ProgressIndicatorPercent, IntegrityError, set_ec, EXIT_WARNING
+    ProgressIndicatorPercent, IntegrityError, set_ec, EXIT_WARNING, safe_ns
 from .platform import acl_get, acl_set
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
note: int_to_bigint triggers at > 63 bits, safe_X triggers at ~ > 64 bits.

ThomasWaldmann:
also (concerning master code): if we create archive metadata with int_to_bigint in 1.0, but do not have bigint_to_int in master any more, guess it would crash with > 63 bits when encountering old metadata.

ThomasWaldmann:
merge?

enkore:
This isn't a problem in 1.0.x ... and 1.1.x has the compat code anyway?

ThomasWaldmann:
As a note for (near?) future additional work: the int_to_bigint could be changed to only trigger for > 64bits (currently it triggers for > 63 bits, but as msgpack does uint64, I guess that is not needed).

ThomasWaldmann:
Motivation here is not to produce crappy archives (with timestamps > 64bits) as we'll have to live with them in the future.

enkore:
I see. So change behaviour in 1.0.x now to avoid sci-fi timestamps, thus avoiding issues with other code interacting with timestamps (e.g. formatting to ISO format) for new archives. Since this only impacts timestamps that are unlikely to be considered significant...

Issue #2314, opened 2017-03-16
Excluding / doesn't work with --patterns-from


edgimar:
In issue #1406 the patterns file format which was chosen should allow something like the following to work, if I'm not mistaken:

    + /home/foo/bar/**
    + /etc/**
    - /**

This should backup the contents of /home/foo/bar and /etc, and nothing else on the filesystem.  Presently using a patterns file like this results in no files getting selected at all.  Regardless of whether it make sense to have a patterns file like this (some might argue that two roots should be specified on the command-line, but there are use-cases when it is preferable to have just a single root of /), the above should still work.

For the record, I am using borg (1.1.0b4.dev223+ng10d4c97c) from the command-line as follows:

    borg-backup create --exclude-if-present .nobackup --keep-exclude-tags --patterns-from patt_file backupserver:/home/backupuser/backups/borgrepo::myarchive /

Backing up without the `--patterns-from` option works fine.

edgimar:
One other note:  I realize now that I can to some extent work around this problem by just specifying roots within the patterns file instead of on the command-line.  But I still do consider the reported behavior (no files getting backed up) to be a bug.

edgimar:
Oops, I take that back -- it really can't be worked around in a reasonable way using roots -- there is a bug that prevents standard use-cases.  For example, say I have a directory tree (beneath a home directory) that looks like:

    mydir/
    mydir/x
    mydir/x/a
    mydir/x/b
    mydir/y

Now let's say I want to backup everything under `mydir`, except that within the `x` subdirectory, I only want to back up `a`, and no other folders within `x`.  So, I attempt to get this behavior with the following pattern file:

    R /home/user/mydir
    + /home/user/mydir/x/y/**
    - /home/user/mydir/x/**

Unfortunately with this patterns file, and using the command-line below, `mydir/x` does not get backed up at all.  If it were working correctly, `mydir/x/y` should get backed up (and nothing else besides that within `mydir/x`).

Command line used:

    borg-backup create --exclude-if-present .nobackup --keep-exclude-tags --patterns-from patt_file backupserver:/home/backupuser/backups/borgrepo::myarchive

(this is the same as in the original issue, except that it doesn't specify a root on the commandline)

ThomasWaldmann:
I guess for a sane handling of this, we need to push the decision about recursion or not to the user.

So it could be something like:
```
+ include/this
- exclude, but recurse into, maybe finding includes
! exclude, do not recurse into
```

[...]

PR #2320, opened 2017-03-20
bugfix: no directory recursion when it is needed

edgimar:
This is a hack to fix the problem raised by issue #2314.  It isn't ideal
because it ends up needing to fully traverse each root subtree, but this can
be improved by deciding whether to descend into a directory based on whether
there are other entries in the pattern file that refer to children of a given
directory.

The problem occurs when a patterns file excludes a parent directory P later in the
file, and earlier in the file a subdirectory S of P is included.
Because a tree is processed recursively, P is processed before S is, but if P
is excluded, then S used to not even be considered, because we wouldn't
recurse into P.  Now we recurse into P nonetheless, but don't add P (as a
directory entry) to the archive.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -487,13 +487,20 @@ class Archiver:
 
         This should only raise on critical errors. Per-item errors must be handled within this method.
         """
+        if st is None:
+            with backup_io('stat'):
+                st = os.lstat(path)
+
+        dir_excluded = False
         if not matcher.match(path):
             self.print_file_status('x', path)
-            return
+
+            if stat.S_ISDIR(st.st_mode):
+                dir_excluded = True
+            else:
+                return
+
         try:
-            if st is None:
-                with backup_io('stat'):
-                    st = os.lstat(path)
             if (st.st_ino, st.st_dev) in skip_inodes:
                 return
             # if restrict_dev is given, we do not want to recurse into a new filesystem,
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -487,13 +487,20 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
...
+            with backup_io('stat'):
+                st = os.lstat(path)
+
+        dir_nomatch = False
         if not matcher.match(path):
             self.print_file_status('x', path)
-            return
+
+            if stat.S_ISDIR(st.st_mode):
+                dir_nomatch = True
```
ThomasWaldmann: dir_excluded?

src/borg/archiver.py
```diff
@@ -521,7 +528,8 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
                                               read_special=read_special, dry_run=dry_run)
                         return
                 if not dry_run:
-                    status = archive.process_dir(path, st)
+                    if not dir_nomatch:
+                        status = archive.process_dir(path, st)
```
ThomasWaldmann: hmm, didn't you say you wanted to recurse into the excluded dirs to search for presence of maybe included dirs below them?

ThomasWaldmann:
btw, I don't think this can be always optimized by looking just at the patterns.
if there are regex patterns, they could be arbitrarily complex and we can't see whether they would match for anything below the excluded parent dir before actually matching the specific path/filename against the regex.

ThomasWaldmann:
Also, it would be nice to have a test, that:
- fails before your bugfix
- succeeds afterwards

edgimar:
added test -- creating new pull request...

ThomasWaldmann:
(you could have just added another commit in the branch you used for this PR)

edgimar:
Sorry about that -- I wondered if that would be better.  When you merge a PR, do you normally squash its commits?

ThomasWaldmann:
Usually it is good if author squashed commit in a reasonable way. Not always down to 1 commit, but as it is cleanest. After review is finished.

PR #2321, opened 2017-03-21
Revert bigint removal

### LINKED ISSUE #2315: review bigint timestamp change
See there: 8b2e7ec68099fc85ac7298d462db14b5f0ee7486 and b7eaeee26631f145a2b7d73f73f20fb56314c3a8.

I guess the files cache related change is ok (far in the future, no problem).
Also, the msgpack range issue has been fixed differently by introducing safe_(n)s recently.

But the fuse change might be problematic, if an old archive (e.g. of borg 1.0) is read and a file back then has triggered the "int as bytes" branch of int_to_bigint (> 2^63 ns). Guess it would just crash if it tried to use that as timestamp without the bigint_to_int conversion.

So:
- removals in cache.py are somehow ok, but would break loading existing file caches, so better not
- readd bigint functions
- re-add the tests (removals in both changesets)
- revert the item.pyx change, add a comment that bigint is needed for old archives.
- add a test that tests bigint-bytes item metadata -> int timestamp?
- check if bigint needed in safe_timestamp

ThomasWaldmann:
fixes #2315.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -34,6 +34,7 @@ from .helpers import safe_encode, safe_decode, make_path_safe, remove_surrogates
 from .helpers import StableDict
 from .helpers import bin_to_hex
 from .helpers import safe_ns
+from .helpers import int_to_bigint, bigint_to_int
 from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
merge?

PR #2322, opened 2017-03-21
allow excluding parent and including child, fixes #2314

### LINKED ISSUE #2314: Excluding / doesn't work with --patterns-from

edgimar:
This fixes the problem raised by issue #2314 by requiring that each root subtree
be fully traversed.

The problem occurs when a patterns file excludes a parent directory P later in
the file, and earlier in the file a subdirectory S of P is included.  Because a
tree is processed recursively, P is processed before S is, but if P is
excluded, then S used to not even be considered, because we wouldn't recurse
into P.  With this fix, we now recurse into P nonetheless, but don't add P (as
a directory entry) to the archive.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -54,7 +54,7 @@ from .helpers import check_extension_modules
 from .helpers import ArgparsePatternAction, ArgparseExcludeFileAction, ArgparsePatternFileAction, parse_exclude_pattern
 from .helpers import dir_is_tagged, is_slow_msgpack, yes, sysinfo
 from .helpers import log_multi
-from .helpers import parse_pattern, PatternMatcher, PathPrefixPattern
+from .helpers import PatternMatcher
 from .helpers import signal_handler, raising_signal_handler, SigHup, SigTerm
 from .helpers import ErrorIgnoringTextIOWrapper
 from .helpers import ProgressIndicatorPercent
@@ -190,16 +190,11 @@ class Archiver:
             bi += slicelen
 
     @staticmethod
-    def build_matcher(inclexcl_patterns, paths):
+    def build_matcher(inclexcl_patterns, include_paths):
         matcher = PatternMatcher()
-        if inclexcl_patterns:
-            matcher.add_inclexcl(inclexcl_patterns)
-        include_patterns = []
-        if paths:
-            include_patterns.extend(parse_pattern(i, PathPrefixPattern) for i in paths)
-            matcher.add(include_patterns, True)
-        matcher.fallback = not include_patterns
-        return matcher, include_patterns
+        matcher.add_inclexcl(inclexcl_patterns)
+        matcher.add_includepaths(include_paths)
+        return matcher
 
     def do_serve(self, args):
         """Start in server mode. This command is usually not used manually."""
@@ -493,13 +488,20 @@ class Archiver:
 
         This should only raise on critical errors. Per-item errors must be handled within this method.
         """
+        if st is None:
+            with backup_io('stat'):
+                st = os.lstat(path)
+
+        recurse_excluded_dir = False
         if not matcher.match(path):
             self.print_file_status('x', path)
-            return
+
+            if stat.S_ISDIR(st.st_mode) and matcher.recurse_dir:
+                recurse_excluded_dir = True
+            else:
+                return
+
         try:
-            if st is None:
-                with backup_io('stat'):
-                    st = os.lstat(path)
             if (st.st_ino, st.st_dev) in skip_inodes:
                 return
             # if restrict_dev is given, we do not want to recurse into a new filesystem,
@@ -527,7 +529,8 @@ class Archiver:
                                               read_special=read_special, dry_run=dry_run)
                         return
                 if not dry_run:
-                    status = archive.process_dir(path, st)
+                    if not recurse_excluded_dir:
+                        status = archive.process_dir(path, st)
                 if recurse:
                     with backup_io('scandir'):
                         entries = helpers.scandir_inorder(path)
@@ -590,7 +593,9 @@ class Archiver:
                 status = '?'  # need to add a status code somewhere
             else:
                 status = '-'  # dry run, item was not backed up
-        self.print_file_status(status, path)
+
+        if not recurse_excluded_dir:
+            self.print_file_status(status, path)
 
     @staticmethod
     def build_filter(matcher, peek_and_store_hardlink_masters, strip_components):
@@ -616,7 +621,7 @@ class Archiver:
             if sys.platform.startswith(('linux', 'freebsd', 'netbsd', 'openbsd', 'darwin', )):
                 logger.warning('Hint: You likely need to fix your locale setup. E.g. install locales and use: LANG=en_US.UTF-8')
 
-        matcher, include_patterns = self.build_matcher(args.patterns, args.paths)
+        matcher = self.build_matcher(args.patterns, args.paths)
 
         progress = args.progress
         output_list = args.output_list
@@ -681,9 +686,8 @@ class Archiver:
                     archive.extract_item(dir_item)
                 except BackupOSError as e:
                     self.print_warning('%s: %s', remove_surrogates(dir_item.path), e)
-        for pattern in include_patterns:
-            if pattern.match_count == 0:
-                self.print_warning("Include pattern '%s' never matched.", pattern)
+        for pattern in matcher.get_unmatched_include_patterns():
+            self.print_warning("Include pattern '%s' never matched.", pattern)
         if pi:
             # clear progress output
             pi.finish()
@@ -893,13 +897,13 @@ class Archiver:
                                'If you know for certain that they are the same, pass --same-chunker-params '
                                'to override this check.')
 
-        matcher, include_patterns = self.build_matcher(args.patterns, args.paths)
+        matcher = self.build_matcher(args.patterns, args.paths)
 
         compare_archives(archive1, archive2, matcher)
 
-        for pattern in include_patterns:
-            if pattern.match_count == 0:
-                self.print_warning("Include pattern '%s' never matched.", pattern)
+        for pattern in matcher.get_unmatched_include_patterns():
+            self.print_warning("Include pattern '%s' never matched.", pattern)
+
         return self.exit_code
 
     @with_repository(exclusive=True, cache=True)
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
@enkore yup, that's bad. see my comment in the ticket.

edgimar:
Any conclusions on how to move forward with this?  Is TW's suggestion OK, or should something else be done?  His suggestion, from #2314 is:

    + include and recurse into
    - exclude, but recurse into, searching for includes
    ! exclude, do not recurse into

enkore:
See my comments in #2314, i.e. feel free to go ahead implementing it.

Which character is which can be easily changed later, if need be.

edgimar:
Ok, big update -- lots of refactoring, and the "!" feature is now implemented.  Sorry for so many changes, but I'm hoping you'll see their beauty ;)

Here's a summary of the important changes:
 - Use an enum to represent possible commands in --patterns-from files
 - rename various variables to reflect their true meaning / remove ambiguity about their datatype
 - refactor some stuff from `archiver.py` to `PatternMatcher.add_includepaths()`
 - probably other wonderful things I'm forgetting (e.g. the new functionality, the "!" command, though there aren't any other new features)

ThomasWaldmann:
@edgimar could you please use rebase onto master and not merge master? also, rephrase some commit comments so they are more meaningful than "refactoring" (e.g. state what code you refactored, what the goal was, ...).

Maybe create a backup, update your local master from the main repo, switch back to your branch again and then do `git rebase -i master`.

edgimar:
Sorry, I do intend to clean up the repo (rebasing, collapsing commits, good commit messages), but at this point it's not probably worth looking at prior commits since most of them are just taken as "snapshots" of the work in progress and not based on atomic sets of changes.

The best is probably to just look at the overall diff between the latest commit and borgbackup:master.  If any further changes are needed, to the code, let me know and I'll try to update it.  After everything looks OK, I'll deal with rebasing / etc.

edgimar:
Ok, here's a complete list of the changes, hopefully in sufficient detail:

- renamed InclExclPattern named-tuple -> CmdTuple (with names 'val' and 'cmd'), since it is used more generally for commands, and not only for representing patterns.
- represent commands as IECommand enum types (RootPath, PatternStyle, Include, Exclude, ExcludeNoRecurse)
- archiver: Archiver.build_matcher() paths arg renamed -> include_paths to prevent confusion as to whether the list of paths are to be included or excluded.
- helpers: PatternMatcher has recurse_dir attribute that is used to communicate whether an excluded dir should be recursed (used by Archiver._process())
- archiver: Archiver.build_matcher() now only returns a PatternMatcher instance, and not an include_patterns list -- this list is now created and housed within the PatternMatcher instance, and can be accessed from there.
- moved operation of finding unmatched patterns from Archiver to PatternMatcher.get_unmatched_include_patterns()
- added / modified some documentation of code
- renamed helpers:parse_add_pattern() patterns arg -> ie_patterns to clarify that pattern type is *not* a subclass of PatternBase, and is rather a CmdTuple -- in fact, I should rename this further to ie_cmds (and other vars in this function similarly)...
- renamed _PATTERN_STYLES -> _PATTERN_CLASSES since "style" is ambiguous and this helps clarify that the set contains classes and not instances.
- have PatternBase subclass instances store whether excluded dirs are to be recursed.  Because PatternBase objs are created corresponding to each +, -, ! command it is necessary to differentiate - from ! within these objects.

src/borg/archiver.py
```diff
@@ -488,13 +483,20 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
...
+            with backup_io('stat'):
+                st = os.lstat(path)
+
+        recurse_excluded_dir = False
         if not matcher.match(path):
             self.print_file_status('x', path)
-            return
+
+            if stat.S_ISDIR(st.st_mode) and matcher.recurse_dir:
+                recurse_excluded_dir = True
```
edgimar: it doesn't matter much to me what it is named, but I chose recurse (as a command verb) because in my mind the recursing hasn't happened yet at that point, and will only happen later on, based on the value of the flag which is set here.

src/borg/helpers.py
```diff
@@ -391,15 +392,15 @@ def parse_timestamp(timestamp):
...
-def parse_add_pattern(patternstr, roots, patterns, fallback):
+def parse_add_pattern(patternstr, roots, ie_patterns, fallback):
     """Parse a pattern string and add it to roots or patterns depending on the pattern type."""
-    pattern = parse_inclexcl_pattern(patternstr, fallback=fallback)
-    if pattern.ptype is RootPath:
-        roots.append(pattern.pattern)
-    elif pattern.ptype is PatternStyle:
-        fallback = pattern.pattern
+    ie_pattern = parse_inclexcl_command(patternstr, fallback=fallback)
+    if ie_pattern.cmd is IECommand.RootPath:
```
edgimar: Either will work -- but enum values are singletons, so `is` works fine, and I believe it is the "right way" to compare them.

src/borg/helpers.py
```diff
@@ -628,73 +676,93 @@ def _match(self, path):
...
+_PATTERN_CLASS_BY_PREFIX = dict((i.PREFIX, i) for i in _PATTERN_CLASSES)
 
-InclExclPattern = namedtuple('InclExclPattern', 'pattern ptype')
-RootPath = object()
-PatternStyle = object()
+CmdTuple = namedtuple('CmdTuple', 'val cmd')
 
 
-def get_pattern_style(prefix):
+class IECommand(Enum):
```
edgimar: Yes, I agree that IE is a little bit cryptic -- it could simply be InclExcl instead of IE, but I thought that made referencing the values kind of a lot to type.  I don't like "PatternType" because these really are more general commands (in some cases, of what should be done with a particular pattern).  Also, it can lead to confusion about what is meant by "pattern type" -- elsewhere in the code this would mean, for example, whether it is a `re` or `sh` pattern, etc.

src/borg/helpers.py
```diff
@@ -628,73 +676,93 @@ def _match(self, path):
...
+    return CmdTuple(epattern_obj, IECommand.Exclude)
+
+
+def parse_inclexcl_command(cmd_line_str, fallback=ShellPattern):
+    """Read a --patterns-from command from string and return a CmdTuple object."""
+
+    cmd_prefix_map = {
+        '-': IECommand.Exclude,  # False
+        '!': IECommand.ExcludeNoRecurse,
+        '+': IECommand.Include,  # True
```
edgimar: Yes, that's old stuff that should be removed.

edgimar:
Ok, I've updated this PR with what I think should be the final changes (apart from squashing commits).  All the tests pass on my machine, so can someone explain what is causing Travis to fail?

edgimar:
Revisions are now squashed and rebased.  What remains to be done to get this merged?

ThomasWaldmann:
@edgimar TypeError: __init__() missing 1 required positional argument: 'recurse_dir'

https://travis-ci.org/borgbackup/borg/jobs/220795799

edgimar:
@ThomasWaldmann The problems with the tests should now be fixed.

edgimar:
@ThomasWaldmann, @enkore - any further changes needed?

src/borg/archiver.py
```diff
@@ -527,7 +529,8 @@ def _process(self, archive, cache, matcher, exclude_caches, exclude_if_present,
                                               read_special=read_special, dry_run=dry_run)
                         return
                 if not dry_run:
-                    status = archive.process_dir(path, st)
+                    if not recurse_excluded_dir:
+                        status = archive.process_dir(path, st)
```
ThomasWaldmann: there might be a problem here:

if the code recurses into an excluded dir (searching for stuff that is included again in directories deeper below),
it does not archive the intermediate directories, so we won't have their metadata (like uid/gid/mode/...) in the archive.

at extract time, borg will automatically create missing intermediate dirs, but it can not set the correct metadata on them (not even when doing a full extract) as it does not have them in the archive.

edgimar: Hm, does it matter if the intermediate dirs are archived before or after the excluded dir?  It could be possible to add them (at the time an excluded dir is archived) to a list of paths to be archived later.

ThomasWaldmann: permissions might be important to set before filling the dir.
timestamps must be set after filling the dir.

(check the code dealing with dirs elsewhere)

ThomasWaldmann: guess the upper-layer dir entries should precede the lower-layer stuff in the archive.

edgimar: I don't claim to know lots about the details of borg's storage method, but I'm assuming that each backup operation stores file chunks in some manner that doesn't depend on order, but tracks the backup itself in the form of some manifest of the backed-up items.  If this is the case, then it could be solved as simply as just ensuring that the manifest is sorted before storing it such that parent folders come before their children.  This would work in conjunction with the idea of maintaining a list of parents to process after a matching nested folder is found.  Does this sound right or doable?

edgimar: Perhaps just use a heap-queue for the manifest list, and always add item-names to it via heapq.heappush()?

enkore:
Merge as-is. Create new issue for the directory inclusion discussed above.

ThomasWaldmann:
Guess we need a similar change in 1.0-maint, right?

ThomasWaldmann:
@edgimar can you check if we want that in 1.0 also and do a backport to 1.0-maint?

enkore:
I don't think more work should be invested into backporting patterns stuff to 1.0-maint until the other loose ends (cf. activity in the other ticket) are tied up.

ThomasWaldmann:
-> #2463

PR #2326, opened 2017-03-22
update CHANGES (1.0)

ThomasWaldmann:


enkore:
1.0.11(rc1) should be delayed until we figure the include patterns stuff out (cmdline parsing fix, include semantics etc.); but for the 1.1 beta this is not a blocking issue.

ThomasWaldmann:
@enkore yes, see the 1.0.11rc1 milestone. if there is something missing, add tickets.

ThomasWaldmann:
as this might take a while, we could have a CHANGES update now nevertheless.
we need to do that more frequently.

ThomasWaldmann:
so, merge or any changes/additions?

enkore:
We can merge it now and update it when we fix the other stuff mentioned above.

PR #2334, opened 2017-03-24
add PathFullPattern / optimization for it

ThomasWaldmann:
not really a pattern (as in potentially having any variable parts) - it just does a full,
precise match, after the usual normalizations.

this is matched via set membership check within O(1) time.

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -518,6 +518,17 @@ class PatternBase:
         raise NotImplementedError
 
 
+class PathFullPattern(PatternBase):
+    """Full match of a path."""
+    PREFIX = "pf"
+
+    def _prepare(self, pattern):
+        self.pattern = os.path.normpath(pattern)
+
+    def _match(self, path):
+        return path == self.pattern
+
+
 # For PathPrefixPattern, FnmatchPattern and ShellPattern, we require that the pattern either match the whole path
 # or an initial segment of the path up to but not including a path separator. To unify the two cases, we add a path
 # separator to the end of the path before matching.
@@ -600,6 +611,7 @@ class RegexPattern(PatternBase):
 
 _PATTERN_STYLES = set([
     FnmatchPattern,
+    PathFullPattern,
     PathPrefixPattern,
     RegexPattern,
     ShellPattern,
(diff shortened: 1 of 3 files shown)


src/borg/helpers.py
```diff
@@ -528,6 +528,17 @@ def _match(self, path):
         raise NotImplementedError
 
 
+class PathFullPattern(PatternBase):
+    """Full match of a path."""
+    PREFIX = "pf"
```
ThomasWaldmann: naming:

we have PathPrefixPattern pp:, so PathFullPattern pf: was chosen for consistency.

src/borg/helpers.py
```diff
@@ -451,22 +451,45 @@ def __init__(self, fallback=None):
...
-        self._items.extend(patterns)
+        for pattern, pattern_type in patterns:
+            self._add(pattern, pattern_type)
 
     def match(self, path):
+        # do a fast lookup for full path matches:
+        # TODO: path = normalized(path), see normalized_wrapper
+        #       pulling this out of N .match() calls and only doing it once is also a good perf. optimization
+        # XXX we do not count such matches, do we need to?
+        non_existent = object()
```
ThomasWaldmann: this is just to enable the possibility of None values in the dictionary.

src/borg/helpers.py
```diff
@@ -528,6 +551,17 @@ def _match(self, path):
...
 
+class PathFullPattern(PatternBase):
+    """Full match of a path."""
+    PREFIX = "pf"
+
+    def _prepare(self, pattern):
+        self.pattern = os.path.normpath(pattern)
+
+    def _match(self, path):
+        return path == self.pattern
```
ThomasWaldmann: after the optimization, PatternMatcher does not call this any more.

src/borg/helpers.py
```diff
@@ -451,22 +451,45 @@ def __init__(self, fallback=None):
...
         """
-        self._items.extend(patterns)
+        for pattern, pattern_type in patterns:
+            self._add(pattern, pattern_type)
 
     def match(self, path):
+        # do a fast lookup for full path matches:
+        # TODO: path = normalized(path), see normalized_wrapper
+        #       pulling this out of N .match() calls and only doing it once is also a good perf. optimization
+        # XXX we do not count such matches, do we need to?
```
ThomasWaldmann: looks like "no", we only count matches for include patterns coming from args.paths, so we can warn if some never matched, but not for other patterns.

ThomasWaldmann:
adding docs now, but otherwise guess this is finished.

PR #2340, opened 2017-03-25
With --compression auto,C, only use C if lz4 achieves at least 3% compression

jdchristensen:


src/borg/helpers.py
```diff
@@ -2162,13 +2162,13 @@ def heuristic_lz4(self, compr_args, chunk):
         cdata = lz4.compress(data)
         data_len = len(data)
         cdata_len = len(cdata)
-        if cdata_len < data_len:
+        if cdata_len < 0.97*data_len:
```
ThomasWaldmann: let's see if that triggers the pep8 checker. usually, spaces around operators are desired.

ThomasWaldmann: no, it didn't.

jdchristensen: Want me to change it, regardless of what the checker says?

enkore: If you don't mind ;)

jdchristensen: I went ahead and amended the commit to use spaces there.

ThomasWaldmann:
another idea that just came to me:

this change now somehow ignores possible compression ratios in the [0.97 ..1.0) - even if we just have proven that even lz4 can achieve them.

if we consider it is worth it, we could decide on lz4 compression for e.g. the [0.97 .. 0.99] range.

in a later optimization, we can even reuse the lz4 compression results so we do not need to compress again.

enkore:
I think that would tie nicely into making CompressionDecider2 a borg.compress.Compressor implementation.

jdchristensen:
The above ideas are nice, but probably only worth it if the code doesn't become too complicated.  Maybe best to postpone for now?

Are those build errors anything to worry about?

ThomasWaldmann:
@jdchristensen the only change in question / in scope of this PR is adding another 2 lines like (pseudo code):
```
...
elif size < orig * 0.99:
    compr = 'lz4'
...
```

ThomasWaldmann:
travis is happy, ignore codecov.

jdchristensen:
I don't think it's worth redoing the lz4 compression for <= 2%.  It also doesn't follow the current documentation, and I don't think it's worth making the compression documentation even more complicated.

ThomasWaldmann:
OK, thanks for the PR!

Issue #2324, opened 2017-03-21
device/FIFO hardlinks


ThomasWaldmann:
While borg has great support for hardlinked regular files, it does not yet do anything special for hardlinked other filesystem items:

```
# create some devices and a symlink
tw@tux:~/aaa$ sudo mknod b00 b 0 0
tw@tux:~/aaa$ sudo mknod c00 c 0 0
tw@tux:~/aaa$ ln -s /vmlinuz symlink

# create hardlinks to them
tw@tux:~/aaa$ sudo ln b00 b00h
tw@tux:~/aaa$ sudo ln c00 c00h
tw@tux:~/aaa$ sudo ln symlink symlinkh

# at the left there is the inode number:
tw@tux:~/aaa$ ls -li
total 0
4187051 brw-r--r-- 2 root root 0, 0 Mar 21 04:44 b00
4187051 brw-r--r-- 2 root root 0, 0 Mar 21 04:44 b00h
4187052 crw-r--r-- 2 root root 0, 0 Mar 21 04:44 c00
4187052 crw-r--r-- 2 root root 0, 0 Mar 21 04:44 c00h
4187053 lrwxrwxrwx 2 tw   tw      8 Mar 21 04:45 symlink -> /vmlinuz
4187053 lrwxrwxrwx 2 tw   tw      8 Mar 21 04:45 symlinkh -> /vmlinuz

tw@tux:~/aaa$ borg init repo
tw@tux:~/aaa$ borg create repo::arch .

tw@tux:~/aaa$ mkdir extr ; cd extr
tw@tux:~/aaa/extr$ sudo borg extract ../repo::arch

tw@tux:~/aaa/extr$ ls -li
total 0
4187086 brw-r--r-- 1 root root 0, 0 Mar 21 04:44 b00
4187087 brw-r--r-- 1 root root 0, 0 Mar 21 04:44 b00h
4187100 crw-r--r-- 1 root root 0, 0 Mar 21 04:44 c00
4187101 crw-r--r-- 1 root root 0, 0 Mar 21 04:44 c00h
4187080 lrwxrwxrwx 1 tw   tw      8 Mar 21 04:45 symlink -> /vmlinuz
4187083 lrwxrwxrwx 1 tw   tw      8 Mar 21 04:45 symlinkh -> /vmlinuz
```
So, the hardlinked devices/symlink were not restored in a hardlinked way.

borg 1.0.10.

ThomasWaldmann:
```
[...]

PR #2342, opened 2017-03-26
Generic hardlinks

ThomasWaldmann:
see #2324.

maybe rather review the changesets than the overall change (displayed badly on github diff view).

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -927,15 +927,19 @@ Utilization of max. archive size: {csize_max:.0%}
     def process_file(self, path, st, cache, ignore_inode=False):
         status = None
         safe_path = make_path_safe(path)
-        # Is it a hard link?
-        if st.st_nlink > 1:
+        item = Item(path=safe_path)
+        hardlink_master = False
+        hardlinked = st.st_nlink > 1
+        if hardlinked:
             source = self.hard_links.get((st.st_ino, st.st_dev))
             if source is not None:
-                item = Item(path=safe_path, source=source)
+                item.source = source
                 item.update(self.stat_attrs(st, path))
                 self.add_item(item)
                 status = 'h'  # regular file, hardlink (to already seen inodes)
                 return status
+            else:
+                hardlink_master = True
         is_special_file = is_special(st.st_mode)
         if not is_special_file:
             path_hash = self.key.id_hash(safe_encode(os.path.join(self.cwd, path)))
@@ -959,10 +963,7 @@ Utilization of max. archive size: {csize_max:.0%}
                 status = 'U'  # regular file, unchanged
         else:
             status = 'A'  # regular file, added
-        item = Item(
-            path=safe_path,
-            hardlink_master=st.st_nlink > 1,  # item is a hard link and has the chunks
-        )
+        item.hardlink_master = hardlinked
         item.update(self.stat_simple_attrs(st))
         # Only chunkify the file if needed
         if chunks is not None:
@@ -985,7 +986,7 @@ Utilization of max. archive size: {csize_max:.0%}
             item.mode = stat.S_IFREG | stat.S_IMODE(item.mode)
         self.stats.nfiles += 1
         self.add_item(item)
-        if st.st_nlink > 1 and source is None:
+        if hardlinked and hardlink_master:
             # Add the hard link reference *after* the file has been added to the archive.
             self.hard_links[st.st_ino, st.st_dev] = safe_path
         return status
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -835,11 +835,11 @@ Utilization of max. archive size: {csize_max:.0%}
         return attrs
 
     @contextmanager
-    def create_helper(self, path, st, status=None):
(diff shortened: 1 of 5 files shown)


src/borg/archive.py
```diff
@@ -965,15 +962,15 @@ def process_file(self, path, st, cache, ignore_inode=False):
...
         item.update(self.stat_attrs(st, path))
         item.get_size(memorize=True)
         if is_special_file:
             # we processed a special file like a regular file. reflect that in mode,
             # so it can be extracted / accessed in FUSE mount like a regular file:
             item.mode = stat.S_IFREG | stat.S_IMODE(item.mode)
-        self.stats.nfiles += 1
         self.add_item(item)
-        if hardlinked and hardlink_master:
+        if hardlink_master:
```
ThomasWaldmann: when hardlink_master is True, hardlinked is also always True.

src/borg/archive.py
```diff
@@ -965,15 +962,15 @@ def process_file(self, path, st, cache, ignore_inode=False):
...
             # Add the hard link reference *after* the file has been added to the archive.
             self.hard_links[st.st_ino, st.st_dev] = safe_path
```
ThomasWaldmann: hmm, thinking about whether this "*after*" is important.
or whether it could be also done in line 922 else-block.

src/borg/archive.py
```diff
@@ -820,34 +820,55 @@ def stat_attrs(self, st, path):
...
-        self.add_item(item)
-        return 's'  # symlink
+        with self.item_helper(path, st, 's') as (item, status, hardlinked, hardlink_master):  # symlink
+            with backup_io('readlink'):
+                source = os.readlink(path)
+            item.source = source  # XXX this overwrites hardlink slave's usage of item.source
+            if hardlinked:
+                logger.warning('hardlinked symlinks will be extracted as non-hardlinked symlinks!')
+            item.update(self.stat_attrs(st, path))
+            return status
```
enkore: In case someone wonders, yes, one can create hard links of device nodes, symlinks and fifos.

I'm just wondering whether it makes sense to really support that?

src/borg/archive.py
```diff
@@ -820,34 +820,55 @@ def stat_attrs(self, st, path):
         attrs.update(self.stat_ext_attrs(st, path))
         return attrs
 
-    def process_dir(self, path, st):
-        item = Item(path=make_path_safe(path))
-        item.update(self.stat_attrs(st, path))
+    @contextmanager
```
enkore: IIRC @contextmanager is quite slow compared to implementing it directly (see backup_io changes). So maybe check/verify performance impact iff the PR should be merged and is ready for it.

src/borg/archive.py
ThomasWaldmann: I'm wondering also. But can we ever know whether someone is using this feature?

I found support for that in restic, btw. :)

enkore: Ok, but then it should also properly support it for symlinks (second source property?)

ThomasWaldmann:
rebased on master, dropped support for hardlinked symlinks to simplify code.

enkore:
What's the WIP here?

ThomasWaldmann:
extract - i am currently working on that.

src/borg/archive.py
```diff
@@ -505,6 +505,26 @@ def add_file_chunks(chunks):
...
+            if link_target:
+                # Hard link was extracted previously, just link
+                with backup_io('link'):
+                    os.link(link_target, path)
+                    hardlink_set = True
+            elif chunks is not None:
+                # assign chunks to this item, since the item which had the chunks was not extracted
+                item.chunks = chunks
+        yield hardlink_set
+        if not hardlink_set and hardlink_masters:  # 2nd term, is it correct/needed?
```
ThomasWaldmann: any ideas?

enkore: We don't use hardlink_masters if everything is extracted (no excludes, no PATHs), so it's needed.

ThomasWaldmann: ok, so I'll update the term to `hardlink_masters is not None` to make it more clear.

src/borg/archive.py
```diff
@@ -519,7 +519,7 @@ def extract_helper(self, dest, item, path, stripped_components, original_path, h
                 # assign chunks to this item, since the item which had the chunks was not extracted
                 item.chunks = chunks
         yield hardlink_set
-        if not hardlink_set and hardlink_masters:  # 2nd term, is it correct/needed?
```
ThomasWaldmann: btw, it can not be None here, due to https://github.com/borgbackup/borg/pull/2342/commits/155f38c2333b3892527796d96e999149d253616e#diff-a7aee10470e434998489f84fe93dc2bcR541

ThomasWaldmann:
ok, guess I am done.

src/borg/archive.py
enkore: Ah right. So we just have to check here whether it is empty, because if it is indeed empty then there cannot be a master to update and the code path above cannot have extracted a slave as a master.

I created ticket #2388 for documenting these issues.

Issue #2343, opened 2017-03-26
item.source is dual-use


ThomasWaldmann:
item.source is the symlink target for symlink items.
item.source is the hardlink master for hardlink slaves.
thus, we have a conflict for symlink items that are also hardlink slaves.

this should be cleaned up at a breaking time, hardlink slaves shall always use item.master to refer to the hardlink master.

see also: #2325

ThomasWaldmann:
#6663 resolved this.

`item.hlid` is used to related hardlinks. `item.source` is not used any more for hardlinks.

ThomasWaldmann:
this is solved in borg2 branch, in 2.0.0a2 release.

PR #2344, opened 2017-03-26
fix symlink item fs size computation

ThomasWaldmann:
a symlink has a 'source' attribute, so it was confused with a hardlink
slave here. see also issue #2343.

also, a symlink's fs size is defined as the length of the target path.

### DIFF
--- a/src/borg/item.pyx
+++ b/src/borg/item.pyx
@@ -1,3 +1,4 @@
+import stat
 from collections import namedtuple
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
added to b4 milestone because it fixes a regression.

@enkore please have a look.

ThomasWaldmann:
added test, collapsed / rebased on master.

PR #2349, opened 2017-03-28
LICENSE: use canonical formulation

enkore:
("copyright holders and contributors" instead of "author")

This does not change the meaning of the license at all, but allows automatic recognition by licensee ("*Confidence: 95.80 %*"), so it will be more obvious which license is used.

2017-03-28-141057_2560x1440_scrot


ThomasWaldmann:
IANAL, but sounds reasonable.

enkore:
ta-da: https://github.com/borgbackup/borg/blob/master/LICENSE

(Also "BSD-3-Clause" in the project metadata box on the main page)

ThomasWaldmann:
Ah, it would have been a useful piece of information that the actual bsd 3-clause license reads like *after* your change (so we had a modified / outdated one?).

Issue #2313, opened 2017-03-16
borg create without syncing cache


enkore:
code exists, needs testing, integration, but not a priority now

ThomasWaldmann:
code url?

enkore:
https://github.com/borgbackup/borg/compare/master...enkore:f/avoid-cache-sync

ThomasWaldmann:
@enkore that adhoc chunks cache is a cool idea, I like it. \o/

I think this need to be broken down into multiple steps:

1. a refactoring PR that creates the class structure as you have in your branch, but does no functional changes (or additions). So it would only implement the local chunks cache as we have it now. as that new structure seems cleaner / easier, I guess we could do that in any case. and it should be pretty easy to review and get merged.

2. we need to agree about csize (discussing it in detail, all the places where it occurs, all the usages it has, and what to do with it)

3. implement the csize change, if any

4. implement your adhoc chunks cache

ThomasWaldmann:
btw, i guess it would be useful if you do a DONTMERGE PR with your current branch, just to be able to get detailled feedback.

enkore:
The main problem with csize is not so much that the technical issues, but a logical issue:

- If I create an archive with some new chunks, they will have csize set
- If I create another archive, but without the cache, so with csize=0 (which is unambiguous) for these chunks
- If I delete the first archive
- And then sync a cache
- The cache cannot know the csize of these chunks
- So we need to ask the repository, which means that has to do per-chunk I/O, because the csize isn't in the repo index, either (which makes a lot of sense)

But that might be ok, since it's a contrived example. A worst case in the very literal sense. Adding a repository API that allows to only get the length of a chunk (and falling back to len(GET()) in the client if that API is not available) would avoid network cost, and mostly leave the I/O.

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/2313#issuecomment-289770897 I'ld put a bounty on 1. immediately and after discussion of 2. maybe another one on 3. - but needs separate tickets first.

ThomasWaldmann:
Talked with @textshell and we both though that this maybe should be shifted to after 1.1.x due to unclear compatibility / effort / scope.

It's a bit of a pity, it is a super interesting feature, but otoh, we should not delay 1.1 too much.

@enkore what do you think?

enkore:
It doesn't matter, because planning on a per-issue level doesn't work with the dynamics of an open source project. If it works out, good, otherwise, good.

---

[...]

PR #2350, opened 2017-03-28
borg create without a cache – Prototype, Do Not Merge

enkore:
#2313

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -169,7 +169,7 @@ class Cache:
             shutil.rmtree(path)
 
     def __init__(self, repository, key, manifest, path=None, sync=True, do_files=False, warn_if_unencrypted=True,
-                 progress=False, lock_wait=None):
+                 progress=False, avoid_cache_sync=False, lock_wait=None):
         """
         :param do_files: use file metadata cache
         :param warn_if_unencrypted: print warning if accessing unknown unencrypted repository
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -25,6 +25,7 @@ from .key import PlaintextKey
 from .locking import Lock
 from .platform import SaveFile
 from .remote import cache_if_remote
+from .repository import LIST_SCAN_LIMIT
 
 FileCacheEntry = namedtuple('FileCacheEntry', 'age inode size mtime chunk_ids')
 
@@ -302,22 +303,6 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
             self.lock.release()
             self.lock = None
 
-    def _read_files(self):
-        self.files = {}
-        self._newest_mtime = None
-        logger.debug('Reading files cache ...')
-        with open(os.path.join(self.path, 'files'), 'rb') as fd:
-            u = msgpack.Unpacker(use_list=True)
-            while True:
-                data = fd.read(64 * 1024)
-                if not data:
-                    break
-                u.feed(data)
-                for path_hash, item in u:
-                    entry = FileCacheEntry(*item)
-                    # in the end, this takes about 240 Bytes per file
-                    self.files[path_hash] = msgpack.packb(entry._replace(age=entry.age + 1))
-
     def begin_txn(self):
         # Initialize transaction snapshot
         pi = ProgressIndicatorMessage(msgid='cache.begin_transaction')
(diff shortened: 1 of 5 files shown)


src/borg/cache.py
```diff
@@ -494,103 +673,104 @@ def create_master_idx(chunk_idx):
...
+            num_requests += 1
+            result = self.cache.repository.list(limit=LIST_SCAN_LIMIT, marker=marker)
+            if not result:
+                break
+            pi.show(len(result))
+            marker = result[-1]
+            # All chunks from the repository have a refcount of MAX_VALUE, which is sticky,
+            # therefore we can't/won't delete them. Chunks we added ourselves in this transaction
+            # (e.g. checkpoint archives) are tracked correctly.
+            init_entry = ChunkIndexEntry(refcount=ChunkIndex.MAX_VALUE, size=0, csize=ChunkIndex.MAX_VALUE)
```
enkore: csize=0 is another safe choice and has a more compact representation (1 byte vs 5 bytes).

src/borg/cache.py
```diff
@@ -494,103 +673,104 @@ def create_master_idx(chunk_idx):
...
-            return None
-
-    def memorize_file(self, path_hash, st, ids):
-        if not (self.do_files and stat.S_ISREG(st.st_mode)):
-            return
-        entry = FileCacheEntry(age=0, inode=st.st_ino, size=st.st_size, mtime=st.st_mtime_ns, chunk_ids=ids)
-        self.files[path_hash] = msgpack.packb(entry)
-        self._newest_mtime = max(self._newest_mtime or 0, st.st_mtime_ns)
+    def commit(self, config):
+        config.set('cache', 'manifest', 'not in sync')
```
enkore: this must be a hex string

src/borg/cache.py
```diff
@@ -180,6 +182,7 @@ def __init__(self, repository, key, manifest, path=None, sync=True, do_files=Fal
         self.timestamp = None
         self.lock = None
         self.txn_active = False
+        self.txn_set = ['config']
```
ThomasWaldmann: nitpick: naming a list a set is a bit strange. also, I usually avoid having a data type in the name.

txn_files maybe?

src/borg/cache.py
```diff
@@ -189,7 +192,10 @@ def __init__(self, repository, key, manifest, path=None, sync=True, do_files=Fal
         self.hostname_is_unique = yes(env_var_override='BORG_HOSTNAME_IS_UNIQUE', prompt=False, env_msg=None)
         if self.hostname_is_unique:
             logger.info('Enabled removal of stale cache locks')
-        self.do_files = do_files
+        if do_files:
+            self.files = FilesCache(self)
+        else:
+            self.files = DummyFilesCache()
```
ThomasWaldmann: files = FC if do_files else DFC

src/borg/cache.py
```diff
@@ -389,6 +371,203 @@ def rollback(self):
...
+            self.begin_txn()
+        return self.files.memorize_file(path_hash, st, ids)
+
+
+class DummyFilesCache:
+    def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
+        return None
+
+    def memorize_file(self, path_hash, st, ids):
+        return None
```
ThomasWaldmann: has no return value, so just use "pass"? (see same below in commit())

src/borg/cache.py
```diff
@@ -389,6 +371,203 @@ def rollback(self):
...
+
+        Return a ChunkListEntry.
+        """
+
+
+    def seen_chunk(self, id, size=None):
+        """
+        Return whether a chunk with *id* was seen. Optionally verify *size* for
+        enhanced collision resistance.
+        """
```
ThomasWaldmann: Currently, this seems to either return a boolean (seen / not seen) or an int (refcount, >0 = seen, 0 = not seen).
Can we avoid the type confusion?

src/borg/cache.py
```diff
@@ -494,103 +673,104 @@ def create_master_idx(chunk_idx):
...
+        capacity = int(num_chunks / ChunkIndex.MAX_LOAD_FACTOR * 1.1)
+        self.chunks = ChunkIndex(capacity)
+        marker = None
+        pi = ProgressIndicatorPercent(total=num_chunks, msg="Downloading chunk list... %3.0f%%", msgid='cache.download_chunks')
+        while True:
+            num_requests += 1
+            result = self.cache.repository.list(limit=LIST_SCAN_LIMIT, marker=marker)
+            if not result:
+                break
+            pi.show(len(result))
```
ThomasWaldmann: Guess this should be:
```
pi.show(increase=len(result))
```

ThomasWaldmann: btw, if you have rather huge LIST_SCAN_LIMIT, this will be a rather jumpy progress indicator.
ofc, it depends on repo size...

src/borg/cache.py
```diff
@@ -494,103 +673,104 @@ def create_master_idx(chunk_idx):
...
+            for id_ in result:
+                self.chunks[id_] = init_entry
+        pi.finish()
+        duration = perf_counter() - t0
+        logger.debug('AdhocChunksCache: downloaded %d chunks in %.2f s (%d requests)', num_chunks, duration, num_requests)
+
     def add_chunk(self, id, chunk, stats, overwrite=False):
-        if not self.txn_active:
-            self.begin_txn()
+        assert not overwrite, 'Logic Bug'
```
ThomasWaldmann: maybe a few more words than just 'Logic Bug'?

src/borg/cache.py
```diff
@@ -494,103 +673,104 @@ def create_master_idx(chunk_idx):
...
 
     def seen_chunk(self, id, size=None):
-        refcount, stored_size, _ = self.chunks.get(id, ChunkIndexEntry(0, None, None))
-        if size is not None and stored_size is not None and size != stored_size:
-            # we already have a chunk with that id, but different size.
-            # this is either a hash collision (unlikely) or corruption or a bug.
-            raise Exception("chunk has same id [%r], but different size (stored: %d new: %d)!" % (
-                            id, stored_size, size))
-        return refcount
+        return id in self.chunks
```
ThomasWaldmann: bool vs. int, see above.

PR #2358, opened 2017-03-30
extract: small bugfix and refactoring for parent dir creation

ThomasWaldmann:
make_parent(path) helper to reduce code duplication.
also use it for directories although makedirs can also do it.

bugfix: also create parent dir for device files, if needed.

(cherry picked from commit d4e27e2952888d547c54d5b0f4be22892aa9e2f1)

Backported for the bugfix. I kept the small refactor also to keep the code better in sync.

### DIFF
--- a/borg/archive.py
+++ b/borg/archive.py
@@ -403,8 +403,6 @@ Number of files: {0.stats.nfiles}'''.format(
             if b'source' in item:
                 source = os.path.join(dest, item[b'source'])
                 with backup_io():
-                    if os.path.exists(path):
-                        os.unlink(path)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Note: I added the same small duplication fix as in #2371.

ThomasWaldmann:
hmm, is our version tuple code somehow broken?

ThomasWaldmann:
borgbackup==0.1.dev2880+nf921373 wat?

ThomasWaldmann:
ok, after ronny's fix (on which i rebased), it works. merge?

PR #2364, opened 2017-03-31
Refactor compression decision stuff

enkore:
Also some desperately needed docs for this.


src/borg/archiver.py
```diff
@@ -107,6 +108,8 @@ def wrapper(self, args, **kwargs):
             with repository:
                 if manifest or cache:
                     kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
+                    if args.__dict__.get('compression'):
```
ThomasWaldmann: that looks strange, esp. considering that you do an attribute access in the next line.

src/borg/compress.pyx
```diff
@@ -1,9 +1,41 @@
...
+plain chunk, compression is irrelevant to it), and of course before encryption.
+
+Borg has a flexible scheme for deciding which compression to use for chunks.
+
+First, there is a global default set by the --compression command line option,
+which sets the .compressor attribute on the Key.
+
+For chunks that emanate from files CompressionDecider may set a specific
+Compressor based on patterns (this is the --compression-from option). This is stored
+as a Compressor instance in the "compressor" key in the Chunk's meta dictionary.
```
ThomasWaldmann: considering we might soon send this over some socket, I'ld rather like this to be a simple object (specification) than a life Compressor class instance.

src/borg/archive.py
```diff
@@ -1658,21 +1656,20 @@ def process_chunks(self, archive, target, item):
...
             return self.cache.chunk_incref(chunk_id, target.stats)
-        chunk = Chunk(data, compress=compress)
-        compression_spec, chunk = self.key.compression_decider2.decide(chunk)
+        chunk = Chunk(data, compressor=compressor)
         overwrite = self.recompress
         if self.recompress and not self.always_recompress and chunk_id in self.cache.chunks:
             # Check if this chunk is already compressed the way we want it
             old_chunk = self.key.decrypt(None, self.repository.get(chunk_id), decompress=False)
-            if Compressor.detect(old_chunk.data).name == compression_spec.name:
+            if Compressor.detect(old_chunk.data).name == compressor.name:
```
ThomasWaldmann: so, if compressor.name == 'auto', it will not match and always recompress / re-store it, right?

the old code did decision2 first and then it would have matched (if one did not change --compress-from or so).

src/borg/compress.pyx
```diff
@@ -179,12 +211,50 @@ class ZLIB(CompressorBase):
...
+        self.compressor = compressor
+        self.lz4 = get_compressor('lz4')
+        self.none = get_compressor('none')
+
+    def compress(self, data):
+        lz4_data = self.lz4.compress(data)
+        if len(lz4_data) < 0.97 * len(data):
+            return self.compressor.compress(data)
+        elif len(lz4_data) < len(data):
+            return lz4_data
```
ThomasWaldmann: \o/

ThomasWaldmann: even though lz4 decompression is very fast, we could still use a minimum compression here, like:
```
len(lz4_data) < 0.999 * len(data)
or even
len(data) >= 100000 and len(lz4_data) < 0.999 * len(data)
```

src/borg/key.py
```diff
@@ -258,8 +254,8 @@ def id_hash(self, data):
         return sha256(data).digest()
 
     def encrypt(self, chunk):
-        chunk = self.compress(chunk)
```
ThomasWaldmann: the idea here was that chunks carry data and metadata and that both flow together through the whole processing pipeline. so don't strip away the metadata even if we do not use it here right now.

src/borg/testsuite/helpers.py
```diff
@@ -1231,27 +1213,17 @@ def test_compression_decider1():
 none:*.zip
 """.splitlines()
```
ThomasWaldmann: maybe there should be also an auto,... entry.

src/borg/key.py
enkore: IIRC we discussed this many months ago and whether we maybe want to tell the repository which is data and which is metadata but decided against it. Since this is an API that is exclusively used within key.py it has no big impact either way (see diff size of the commit), apart from making the code a bit simpler.

src/borg/compress.pyx
enkore: YAGNI / trying to solve a problem before it exists tends to not work well / at all.

ThomasWaldmann: well, it already was solved, so you are unsolving it rather.

serializing some tuple is no big issue, serializing an instance of a more complex object can be more troublesome.

src/borg/key.py
ThomasWaldmann: just consider chunks being tuple(meta, data) infrastructure we do not want to throw away.
i introduced that concept a while ago for multiple reasons and i don't want to redo that.

enkore:
I'm done here.

src/borg/archiver.py
```diff
@@ -107,6 +108,9 @@ def wrapper(self, args, **kwargs):
             with repository:
                 if manifest or cache:
                     kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
+                    # do_recreate uses args.compression=None as in band signalling for "don't recompress"
+                    if getattr(args, 'compression', False):
+                        kwargs['key'].compressor = args.compression.compressor
```
ThomasWaldmann: even with the comment, it is not quite clear how it works.

key.compressor is by default Compressor('none') if line 113 is not executed.

it is executed for a normal create run and then either lz4 or whatever was specified on the cmdline.

if args.compression is None (IBS) the line is not executed and we still have Compressor('none') there, which is "no/zero compression" but not "don't recompress".

src/borg/compress.pyx
```diff
@@ -222,22 +243,36 @@ class Auto(CompressorBase):
...
+        Decides what to do with *data*. Returns (compressor, lz4_data).
+
+        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
+        """
         lz4_data = self.lz4.compress(data)
-        if len(lz4_data) < 0.97 * len(data):
-            return self.compressor.compress(data)
-        elif len(lz4_data) < len(data):
-            return lz4_data
+        ratio = len(lz4_data) / len(data)
```
enkore: I don't see a way how that could happen - the chunker doesn't emit empty chunks.

ThomasWaldmann: ok. I also couldn't find a case where this blows up, just the theoretical possibility.

PR #2370, opened 2017-04-01
Windows ssh

Anakonda:
Fix ssh not working on windows.
Use putty link (plink) as ssh client if openssh cant be found.

Interactive authenticating does not work so you need to use shh keys

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -2,6 +2,8 @@ import errno
 import sys
 if sys.platform != 'win32':
     import fcntl
+else:
+    import shutil
 import logging
 import os
 import select
@@ -18,6 +20,12 @@ from .helpers import get_home_dir
 from .helpers import sysinfo
 from .helpers import bin_to_hex
 from .repository import Repository
+
+if sys.platform == 'win32':
+    from .platform import select as windowsSelect
+    select.select = windowsSelect
+    class NoSSHClient(Error):
+        """Could not find supported ssh client. Supported  clients are {}."""
 
 RPC_PROTOCOL_VERSION = 2
 
@@ -237,12 +245,27 @@ class RemoteRepository:
     def ssh_cmd(self, location):
         """return a ssh command line that can be prefixed to a borg command line"""
         args = shlex.split(os.environ.get('BORG_RSH', 'ssh'))
-        if location.port:
-            args += ['-p', str(location.port)]
-        if location.user:
-            args.append('%s@%s' % (location.user, location.host))
+        if sys.platform == 'win32' and 'BORG_RSH' not in os.environ:
+            if shutil.which('ssh') is not None:
+                args = ['ssh']
+            elif shutil.which('plink') is not None:
+                args = ['plink', '-ssh', '-batch']
+            else:
+                raise NoSSHClient('ssh and plink')
+        if args[0] == 'plink':
+            if location.port:
+                args += ['-P', str(location.port)]
+            if location.user:
+                args.append('%s@%s' % (location.user, location.host))
+            else:
+                args.append('%s' % location.host)
         else:
-            args.append('%s' % location.host)
+            if location.port:
+                args += ['-p', str(location.port)]
+            if location.user:
+                args.append('%s@%s' % (location.user, location.host))
+            else:
+                args.append('%s' % location.host)
         return args
 
     def call(self, cmd, *args, **kw):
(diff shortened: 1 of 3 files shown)


src/borg/remote.py
```diff
@@ -19,6 +21,29 @@
...
+        retRlist = []
+        retXlist = []
+        for pipe in rlist:
+            size = ctypes.wintypes.DWORD(0)
+            if not PeekNamedPipe(msvcrt.get_osfhandle(pipe), None, 0, None, ctypes.byref(size), None):
+                if size.value == 0 and pipe in xlist:
+                    retXlist.append(pipe)
+            if size.value > 0:
+                retRlist.append(pipe)
+        return retRlist, wlist, retXlist
```
ThomasWaldmann: maybe move the above code to a windows platform module?

src/borg/remote.py
```diff
@@ -19,6 +21,29 @@
...
+    select.select = newSelect
```
ThomasWaldmann: so, does this mean the stdlib select.select is broken on windows?

src/borg/remote.py
```diff
@@ -236,14 +261,26 @@ def borg_cmd(self, args, testing):
 
     def ssh_cmd(self, location):
         """return a ssh command line that can be prefixed to a borg command line"""
-        args = shlex.split(os.environ.get('BORG_RSH', 'ssh'))
-        if location.port:
-            args += ['-p', str(location.port)]
-        if location.user:
-            args.append('%s@%s' % (location.user, location.host))
+        if sys.platform != 'win32' or shutil.which('ssh') is not None:
+            args = shlex.split(os.environ.get('BORG_RSH', 'ssh'))
```
ThomasWaldmann: guess the BORG_RSH must be outside the if/elif/else or it will only work if one is not on windows or has ssh installed. but the point of using this env var is if you do NOT just want to use ssh OR if you want ssh, but with different options.

src/borg/remote.py
enkore: select is part of WSA which only works for sockets and never has / never will work for anything else.

src/borg/remote.py
```diff
@@ -237,12 +243,27 @@ def borg_cmd(self, args, testing):
...
-            args += ['-p', str(location.port)]
-        if location.user:
-            args.append('%s@%s' % (location.user, location.host))
+        if sys.platform == 'win32' and 'BORG_RSH' not in os.environ:
+            if shutil.which('ssh') is not None:
+                args = ['ssh']
+            elif shutil.which('plink') is not None:
+                args = ['plink', '-ssh', '-batch']
+            else:
+                raise Exception('Could not find supported ssh client.')
```
ThomasWaldmann: hmm, can you raise an exception that does not throw a traceback? See "Error" & co classes.

also, to be a bit more helpful, maybe tell what ssh clients are supported (or the command names you are looking for).

src/borg/remote.py
```diff
@@ -237,12 +243,27 @@ def borg_cmd(self, args, testing):
...
+            else:
+                args.append('%s' % location.host)
         else:
-            args.append('%s' % location.host)
+            if location.port:
+                args += ['-p', str(location.port)]
+            if location.user:
+                args.append('%s@%s' % (location.user, location.host))
+            else:
+                args.append('%s' % location.host)
```
ThomasWaldmann: most of the if/else block is the same. are we expecting to have to support more differences than just -P vs -p?

if not, it could be unified and just use a "'-P' if plink else '-p'" expression.

Anakonda: I left it like that on to make adding support for more clients easyer later.

src/borg/remote.py
```diff
@@ -24,6 +24,8 @@
 if sys.platform == 'win32':
     from .platform import select as windowsSelect
     select.select = windowsSelect
+    class NoSSHClient(Error):
+        """Could not find supported ssh client. Supported  clients are {}"""
```
ThomasWaldmann: nitpick: add a dot at the end.

src/borg/remote.py

ThomasWaldmann:
i can't review the windows specific code, but the rest lgtm.

Anakonda:
Should I squash commits?

ThomasWaldmann:
yes, please.

Anakonda:
Squashed all commits.

ThomasWaldmann:
@enkore anything left from your side? if not, I'ld merge this.

PR #2371, opened 2017-04-01
extract: remove duplicate code

ThomasWaldmann:
anything at <path> gets nuked already a few lines above, if possible.

see line 555 / 557.


ThomasWaldmann:
that was quick. :D

PR #2375, opened 2017-04-01
BORG_HOSTNAME_IS_UNIQUE=yes by default.

enkore:
Note b9770c348

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -403,10 +403,16 @@ Examples
 borg serve has special support for ssh forced commands (see ``authorized_keys``
 example below): it will detect that you use such a forced command and extract
 the value of the ``--restrict-to-path`` option(s).
+
 It will then parse the original command that came from the client, makes sure
 that it is also ``borg serve`` and enforce path restriction(s) as given by the
 forced command. That way, other options given by the client (like ``--info`` or
 ``--umask``) are preserved (and are not fixed by the forced command).
+
+Environment variables (such as BORG_HOSTNAME_IS_UNIQUE) contained in the original
+command sent by the client are *not* interpreted, but ignored. If BORG_XXX environment
+variables should be set on the ``borg serve`` side, then these must be set in system-specific
+locations like ``/etc/environment`` or in the forced command itself (example below).
 
 ::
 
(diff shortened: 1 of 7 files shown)


src/borg/remote.py
```diff
@@ -646,8 +645,8 @@ def borg_cmd(self, args, testing):
             except AttributeError:
                 pass
         env_vars = []
-        if yes(env_var_override='BORG_HOSTNAME_IS_UNIQUE', env_msg=None, prompt=False):
-            env_vars.append('BORG_HOSTNAME_IS_UNIQUE=yes')
+        if not hostname_is_unique():
+            env_vars.append('BORG_HOSTNAME_IS_UNIQUE=no')
```
ThomasWaldmann: there is the remaining problem that passing this is not effective with a forced command (see your other PR, you just throw away the env vars).

enkore: That's intentional, because BHIU is local system policy that may not be overridden by an unprivileged remote client.

ThomasWaldmann: Well, guess this needs some docs or users will wonder why switching it off does not work (before your change: why switching it on does not work).

enkore: Just to clarify: The whole SSH_ORIGINAL_COMMAND enchilada only matters iff an SSH forced command is used. Otherwise this directly works, since the client has direct shell access.

docs/usage.rst
```diff
@@ -403,11 +403,15 @@ Examples
...
 example below): it will detect that you use such a forced command and extract
 the value of the ``--restrict-to-path`` option(s).
+
 It will then parse the original command that came from the client, makes sure
 that it is also ``borg serve`` and enforce path restriction(s) as given by the
 forced command. That way, other options given by the client (like ``--info`` or
 ``--umask``) are preserved (and are not fixed by the forced command).
 
+Environment variables (such as BORG_HOSTNAME_IS_UNIQUE) contained in the original
+command sent by the client are *not* interpreted, but ignored.
```
ThomasWaldmann: ... You need to set the environment variable in the forced command.

(does that actually work?)

enkore: - Add it to /etc/environment
- Add it to ~/.ssh/environment
- Add it to other OS-specific places
- Possibly shrc's and /etc/profile, if a shell indirection is invoked along the way (e.g. `command="some_wrapper_around_borg.sh"`)
- Add it the forced command line, e.g. `command="export BORG_SOMETHING=FOO ; borg serve ...",restrict ssh-ed25519 ...`

enkore:
Note that due to b9770c3 it should be very rarely if ever be necessary to disable this. I've tested this on the CI systems and it worked correctly -- only on Oracle Solaris it was different every time, but that is a fail-safe condition, i.e. it keeps it from breaking locks.

docs/usage.rst
ThomasWaldmann: agreed. we don't need to document linux in our docs.

PR #2383, opened 2017-04-04
recreate: add --recompress flag, avoid weirdo use of args.compression

enkore:


docs/usage.rst
```diff
@@ -463,7 +463,7 @@ Examples
     $ borg create /mnt/backup::archive /some/files --compression lz4
     # Then compress it - this might take longer, but the backup has already completed, so no inconsistencies
     # from a long-running backup job.
-    $ borg recreate /mnt/backup::archive --compression zlib,9
+    $ borg recreate /mnt/backup::archive --recompress --compression zlib,9
```
ThomasWaldmann: hmm, so if someone just gives `recreate --compression zlib,9` it won't do anything related to compression?

src/borg/archive.py
```diff
@@ -1557,7 +1557,7 @@ def is_temporary_archive(archive_name):
 
     def __init__(self, repository, manifest, key, cache, matcher,
                  exclude_caches=False, exclude_if_present=None, keep_exclude_tags=False,
-                 chunker_params=None, compression=None, always_recompress=False,
+                 chunker_params=None, compression=None, recompress=False, always_recompress=False,
```
ThomasWaldmann: could be also something like --recompress=never/ifdifferent/always.

src/borg/archiver.py
```diff
@@ -108,14 +108,7 @@ def wrapper(self, args, **kwargs):
...
                     kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
-                    # do_recreate uses args.compression is None as in band signalling for "don't recompress",
-                    # note that it does not look at key.compressor. In this case the default compressor applies
-                    # to new chunks.
-                    #
-                    # We can't use a check like `'compression' in args` (an argparse.Namespace speciality),
-                    # since the compression attribute is set. So we need to see whether it's set to something
-                    # true-ish, like a CompressionSpec instance.
-                    if getattr(args, 'compression', False):
+                    if 'compression' in args:
```
ThomasWaldmann: hmm, isn't compression always there? it has a default (lz4) or it is given explicitly.

src/borg/archiver.py
```diff
@@ -2929,7 +2922,8 @@ def process_epilog(epilog):
         Note that all paths in an archive are relative, therefore absolute patterns/paths
         will *not* match (--exclude, --exclude-from, PATHs).
 
-        --compression: all chunks seen will be stored using the given method.
+        --recompress: all chunks seen will be recompressed using the --compression
+        specified.
```
ThomasWaldmann: unrelated to your change, but I just noticed that this is not quite true:
they are not recompressed if they already have that compression, except with --always-recompress.

src/borg/archiver.py
enkore: Only few commands have a --compression option.

ThomasWaldmann: ah, right %-)

docs/usage.rst
enkore: Still applies to new chunks (metadata, archive item, manifest etc.), but not existing data chunks.

src/borg/archive.py
enkore: Yeah that seems better than having two independent flags. Perhaps argparse can make it so that just `--recompress` means if-different to abbreviate the common case. (Well, as "common" as recreate can be).

PR #2387, opened 2017-04-04
placeholders: deny access to internals and other unspecified stuff

enkore:
Also better error messages.

`Invalid placeholder "horstname" in string: testrepo::{horstname}-{now}`

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1755,6 +1755,10 @@ class Archiver:
 
             The version of borg, only major, minor and patch version, e.g.: 1.0.8
 
+        If literal curly braces need to be used, double them for escaping::
+
+            borg create /path/to/repo::{{literal_text}}
+
         Examples::
 
             borg create /path/to/repo::{hostname}-{user}-{utcnow} ...
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -784,9 +784,10 @@ class DatetimeWrapper:
 
 
 def format_line(format, data):
-    keys = [f[1] for f in Formatter().parse(format)]
-    for key in keys:
-        if '.' in key or '__' in key:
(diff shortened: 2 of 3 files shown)


src/borg/archiver.py
```diff
@@ -1755,6 +1755,10 @@ def do_break_lock(self, args, repository):
 
             The version of borg, only major, minor and patch version, e.g.: 1.0.8
 
+        If literal curly braces need to be used, double them for escaping::
+
+            borg create /path/to/repo::{{literal_text}}
```
ThomasWaldmann: hmm, does a (non-coding) reader understand "literal" and "escaping"?

src/borg/helpers.py
```diff
@@ -780,8 +784,15 @@ def __format__(self, format_spec):
 
 
 def format_line(format, data):
+    for _, key, _, conversion in Formatter().parse(format):
+        if not key:
+            continue
+        if '.' in key or '__' in key or conversion:
```
ThomasWaldmann: instead of blacklisting . and __ we could also use a whitelist.

src/borg/archiver.py
enkore: borg's a shell tool... shell escaping knowledge is required :(

src/borg/helpers.py
```diff
@@ -780,8 +784,15 @@ def __format__(self, format_spec):
...
 def format_line(format, data):
+    for _, key, _, conversion in Formatter().parse(format):
+        if not key:
+            continue
+        if conversion or key not in data:
+            raise InvalidPlaceholder(key, format)
     try:
-        return format.format(**data)
+        return format.format_map(data)
+    except KeyError as ke:
```
ThomasWaldmann: can this KeyError still happen considering 790/791?

PR #2392, opened 2017-04-06
Document repository file system requirements

### LINKED ISSUE #2080: Not documented: Support for NTFS
It is not documented whether borg supports NTFS as source or target (=where the repository is) fs.
- Can I take backup from files on NTFS partition (e.g. USB disk)? And of course be able to restore them back to NTFS partition?
- Can I create a repository on NTFS partition (e.g. USB disk)?
- Are there any problems backing/restoring linux fs (e.g. ext4) files and directories to repository on NTFS disk?

My borgs are running in CentOS / Ubuntu boxes all having ntfs-3g.

Please add this information also to user guide.

wolfpackmars2:
This resolves #2080 by adding the file system requirements to the docs. Information regarding borg requirements are spread throughout the documentation.  Some information shows up in the Installation section, other information shows up in Usage as well as other sections of the documentation. As a new user, I tried to identify a logical place I would go to find this information, and created a new section for borg requirements before the installation.  Going forward, additional requirements and "gotchas" can be added to this section.

No references or information was removed from any other section.


docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+|project_name| Repository Filesystem
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- No special filename requirements exist. Filenames may contain any valid
+  ascii character and be of any length. |project_name| stores data only 3
+  directory levels deep and uses short directory and filenames.
+- No special file metadata requirements exist regarding permissions or extended
+  file attributes. |project_name| only requires read/write permissions within
+  the repository filesystem.
+- Segment files generated by |project_name| are no larger than 5MB by default
```
ThomasWaldmann: 5MiB are not a hard limit, thus rather say "are up to approx. 5MiB".

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+  (Ver. 1.0.x). This can be modified in :ref:`config-file`
```
ThomasWaldmann: add a dot at the end.

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+- There are no special requirements with regard to inode limits. Data is stored
+  within the segment files per the size specified in the config. Storing many
```
ThomasWaldmann: s/Storing/Backing up/

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+  small files will not translate into having an equal or larger amount of files
+  within the repository. Deduplication helps as well. Most importantly, the
```
ThomasWaldmann: Not sure how the "Deduplication helps as well." fits into here.

Same for the "Most importantly,".

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+  number of segments within a directory can be set in the :ref:`config-file`.
+- Mkdir based file locking is used. File system based locking is not required.
```
ThomasWaldmann: mkdir is filesystem based, see the ticket about what's not required.

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+- |project_name| does not create hardlinks, symlinks or other special files
```
ThomasWaldmann: dot at end of sentences.

also, i think we meanwhile temporarily need hardlinks for SaveFile.

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+- Written data must be consistent. **A journaling file system is strongly
```
ThomasWaldmann: not sure if people understand the first sentence.

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+  encouraged.**
+- |project_name| uses the following file system operations:
+  - create, open, read, write, seek, close, rename, delete
```
ThomasWaldmann: link (as in hardlink)

docs/installation.rst
```diff
@@ -20,6 +20,38 @@ There are different ways to install |project_name|:
...
+  - listdir, stat
+  - posix_fadvise (to not flood the operating system's cache)
+  - sync on files and directories (to make sure data has been written)
```
ThomasWaldmann: ... onto storage media

docs/installation.rst
enkore: SaveFile uses rename(2); but an inplace upgrade of an Attic repo uses hardlinks.

ThomasWaldmann: sorry, i meant Repository.save_config. it tries to use a hardlink, but it can live without it, too.

so, no need to mention hardlinks except for inplace upgrade.

docs/installation.rst
enkore: Suggestion: Just recommend a journaling FS; I/O consistency is an extremely complicated concept.

See usage_general line 225 and following ("File systems")

docs/installation.rst
enkore: (500 MiB)

wolfpackmars2: ok I see what you're saying.  I meant "5MB are the default segment size". That was based off this statement:
```no special requirements for the file size (borg creates small to medium sized files). see repo config, default for 1.0.x is ~5MB```
I understand the meaning was that file size limits are not a problem for Borg -- in reference to some file systems for e.g. that cannot have files larger than 2 or 4 gb.
I plan on going through these comments shortly and seeing where I can be more clear.
Appreciate the constructive criticism! :)

File: docs/installation.rst (hunk 4 above)
Comment (wolfpackmars2) [inline docs/installation.rst]: Will review.
Deduplication helps because though you may try to store 10,000 files, you may not end up with that many segments as some files may share segments.  I can either clarify this statement or more likely remove it all together.

I was referencing this statement: "having a lot of small files in the backup source does not cause having the same or even bigger amount of files in the repo." I was trying to give an example of why that could be the case.

Also, I would like a little clarification if I'm mistaken, but it seems that if a file is smaller than a segment size, that files data may be combined with data from another file in the same segment, correct... or not?

File: docs/installation.rst (hunk 5 above)
Comment (wolfpackmars2) [inline docs/installation.rst]: "Borg does not rely on posix, nfs, windows, smb or other file locking calls." I think sounds better. I wasn't confused, but I understand I'm not being clear in my original statement.

I meant "locking mechanisms provided by a particular file system [engine] are not required" but clearly that's not very clear. Clear? :)

File: docs/installation.rst (hunk 1 above)
Comment (ThomasWaldmann) [inline docs/installation.rst]: as your PR is against master branch, it needs to be 500MiB.
5MiB is 1.0-maint branch.

File: docs/installation.rst (hunk 7 above)
Comment (wolfpackmars2) [inline docs/installation.rst]: It was mentioned about symlinks/hardlinks in the original ticket. I considered that it didn't seem very important. However, in my research of backup systems, it seems some people either want hard linking (rsync) or don't want hard linking (don't want to use rsync). Therefore I thought it relevant to point out that Borg does not use hardlinks.

File: docs/installation.rst (hunk 8 above)
Comment (wolfpackmars2) [inline docs/installation.rst]: so I'm clear... when performing an upgrade of an attic repository (specifically, an in-place upgrade), hard links are used?
Briefly, what are the hard links used for?

Comment (ThomasWaldmann) [inline docs/installation.rst]: As a kind of cheap copy.

File: docs/installation.rst (hunk 4 above)
Comment (ThomasWaldmann) [inline docs/installation.rst]: you need to differentiate between source files, chunks and segments:
- source files (what you want to back up)
- chunks (pieces of files) - the result of chunking, small source files usually yield 1 chunk, big files yield multiple chunks with a target size of ~2MiB.
- segment files - files for data storage in the repository, they are ~500MiB (1.0: 5MiB) and contain multiple chunks.

File: docs/installation.rst [hunk 10]
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
...
+.. _installation-requirements:
+
+Pre-Installation Considerations
+-------------------------------
+
+Repository Filesystem
+~~~~~~~~~~~~~~~~~~~~~
+
+- Filenames may contain any valid ascii character and be of any length. |project_name|
+  stores data only 3 directory levels deep and uses short directory and filenames.
```
ThomasWaldmann: ... directory and file names.

docs/installation.rst
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
...
+- |project_name| requires only read/write permissions within the repository filesystem.
```
ThomasWaldmann: that sounds a bit strange. why "only"?
also, I guess we require "x" on the directories.

so, just say "... requires r/w access to the repository filesystem."

docs/installation.rst
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
...
+- Source files are split up into pieces called "chunks". During the chunking
+  process, small files will typically yield one chunk while larger files will
+  be broken up into chunks of ~2MiB in size.
```
ThomasWaldmann: this explains how borg works, but does not really fit into this section.

docs/installation.rst
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
...
+- Chunks are combined into segment files. The number and size of segment files
+  can be set in the :ref:`config-file`.
```
ThomasWaldmann: |project_name| stores backup metadata and data into so-called segment files.
The target size of these files and also the count of these files per directory is set in the :ref:`config-file`.

docs/installation.rst
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
...
+Repository Filesystem
+~~~~~~~~~~~~~~~~~~~~~
+:ref:data-structures-and-file-formats
+- File names may contain any valid ascii character and be of any length. |project_name|
+  stores data only 3 directory levels deep and uses short directory and file names.
+- |project_name| requires read and write permissions on the repository file system.
+- |project_name| stores backup metadata and data into so-called segment files. The 
+  target size of these files and also the count of these files per directory is set 
+  in the :ref:config-file.
+- |project_name| uses Mkdir based `file locking<https://en.wikipedia.org/wiki/File_locking#Lock_files`_.
```
ThomasWaldmann: """
... uses a generic and very portable mkdir-based locking mechanism.
It does **not use**: POSIX locks, NFS locks, windows file locks, lockf(), flock(), hardlinks.
"""
Pointing out what it does not use is useful here because these mechanisms often cause issues (and we have seen quite some of them in attic).

docs/installation.rst
```diff
@@ -20,6 +20,35 @@ There are different ways to install |project_name|:
     have the latest code or use revision control (each release is
     tagged).
 
+.. _installation-requirements:
+
+Pre-Installation Considerations
+-------------------------------
```
wolfpackmars2: I'm open to suggestions ... looking through the existing sections, there doesn't seem to be an existing section where this would better fit.

There isn't currently a requirements section. "Installation" is the first major section of the documentation.  Logically, before installing an app, one should understand whether their system meets the requirements of the app.

I've found that a lot of this information is already available in different places spread out in the docs... but there isn't one location where all this information is pulled together.

It can possibly be shortened and added to the FAQs, but I'm not sure that would have helped in the case of the original bug report.  Maybe #2080 can be resolved with a list of file systems explicitly *NOT* supported?  I think a list of *supported* file systems would be rather long and cumbersome to maintain. Presumably newer file systems will be supported, while a handful of old file systems (e.g. Fat16) may work but aren't recommended?

ThomasWaldmann:
ok, I'll merge it.

Issue #2341, opened 2017-03-26
add FAQ item about borg check --repair


ThomasWaldmann:
Make a FAQ entry, based on this:

```
CharlesA:
Is there any risk to running borgbackup check --repair?
I tried it and saw a warning that it was experimental and it could result in data loss.

ThomasWaldmann:
well, it usually works, but we can't predict all modes of corruption
(or be sure no bugs are left), so it issues this warning.
so, if you can't risk losing the repo, you better have a backup of it
(or run borg check --repair on a copy of it)
when corruption occurs, the most interesting question is always
the root cause.
another reason why it warns might be that if the cause is malfunctioning
hardware, running a repair might even do more damage or malfunctioning
fs or hw drivers.
```

PR #2393, opened 2017-04-06
FAQ to explain warning when running --repair

wolfpackmars2:
In reference to #2341

Added an explanation about the risks associated with running `borg check --repair` and why the program presents a warning to the user.

Also, fixed some _very minor_ grammar.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -496,33 +496,22 @@ issue in the |project_name| issue tracker about that.
 Why does running 'borgbackup check --repair' warn about data loss?
 ------------------------------------------------------------------
 
-Due to the nature of backup, and especially deduplicated backups as used
-by |project_name|, some repair operations may require a roll-back to an
-earlier version of the repository. Some errors, especially where data
-corruption is involved, may require purging corrupted data to bring the
-repository back to a known good state.
-
-Also, we can't predict every possible corruption vector for a repository.
-Some forms of corruption may be undetected, and some attempts to repair
-corruption could break other parts of the repository.
-
-Repair is considered a destructive operation in that it will modify the
-repository in order to automatically repair some known methods of corruption.
+Repair usually works for recovering data in a corrupted archive. However,
+it's impossible to predict all modes of corruption. In some very rare
+instances, such as malfunctioning storage hardware, additional repo
+corruption may occur. If you can't afford to lose the repo, it's strongly
+recommended that you perform repair on a copy of the repo.
 
 In other words, the warning is there to emphasize that |project_name| will:
   - Perform automated routines that will modify your backup repository
   - May not actually fix the problem you are experiencing
   - May further corrupt your repository
-  - May cause other problems which may not be noticable
-  - May cause contents of backed up data to become corrupted or changed
 
-Aside from this, if you find yourself in a position where you believe your
-repository is corrupt, this operation may result in recovering at least
-some of your backup data. You are **STRONGLY ENCOURAGED** to attempt the
-repair on a copy of the repo in case the repository becomes more unusable
-after the repair.
-
-If you are still unsure, seek :ref:`support`!
+In the case of malfunctioning hardware, such as a drive or USB hub
+corrupting data when read or written, it's best to diagnose and fix the
+cause of the initial corruption before attempting to repair the repo. If
+the corruption is caused by a one time event such as a power outage,
+running `borg check --repair` will fix most problems.
 
 Miscellaneous
 #############
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -347,7 +347,7 @@ How can I backup huge file(s) over a unstable connection?
 
 This is not a problem any more, see previous FAQ item.
 
-How can I restore huge file(s) over a unstable connection?
+How can I restore huge file(s) over an unstable connection?
 ----------------------------------------------------------
 
 If you can not manage to extract the whole big file in one go, you can extract
@@ -381,7 +381,7 @@ If you run into that, try this:
 
 .. _a_status_oddity:
 
-I am seeing 'A' (added) status for a unchanged file!?
+I am seeing 'A' (added) status for an unchanged file!?
 -----------------------------------------------------
 
 The files cache is used to determine whether |project_name| already
@@ -492,6 +492,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
 
 If you can reproduce the issue with the proven filesystem, please file an
 issue in the |project_name| issue tracker about that.
+
+Why does running 'borgbackup check --repair' warn about data loss?
+------------------------------------------------------------------
+
+Due to the nature of backup, and especially deduplicated backups as used
+by |project_name|, some repair operations may require a roll-back to an
+earlier version of the repository. Some errors, especially where data
+corruption is involved, may require purging corrupted data to bring the
+repository back to a known good state.
+
+Also, we can't predict every possible corruption vector for a repository.
+Some forms of corruption may be undetected, and some attempts to repair
+corruption could break other parts of the repository.
+
+Repair is considered a destructive operation in that it will modify the
+repository in order to automatically repair some known methods of corruption.
+
+In other words, the warning is there to emphasize that |project_name| will:
+  - Perform automated routines that will modify your backup repository
+  - May not actually fix the problem you are experiencing
+  - May further corrupt your repository
+  - May cause other problems which may not be noticable
+  - May cause contents of backed up data to become corrupted or changed
+
+Aside from this, if you find yourself in a position where you believe your
+repository is corrupt, this operation may result in recovering at least
+some of your backup data. You are **STRONGLY ENCOURAGED** to attempt the
+repair on a copy of the repo in case the repository becomes more unusable
+after the repair.
+
+If you are still unsure, seek :ref:`support`!
 
 Miscellaneous
 #############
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -493,6 +493,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
 If you can reproduce the issue with the proven filesystem, please file an
 issue in the |project_name| issue tracker about that.
 
+Why does running 'borgbackup check --repair' warn about data loss?
+------------------------------------------------------------------
+
+Due to the nature of backup, and especially deduplicated backups as used
+by |project_name|, some repair operations may require a roll-back to an
+earlier version of the repository. Some errors, especially where data
```
ThomasWaldmann: I don't think borg does rollbacks to earlier verisions.
It does rollback to last committed state, but that is not special for repair.

docs/faq.rst
```diff
@@ -493,6 +493,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+corruption is involved, may require purging corrupted data to bring the
+repository back to a known good state.
+
+Also, we can't predict every possible corruption vector for a repository.
+Some forms of corruption may be undetected, and some attempts to repair
```
ThomasWaldmann: What forms may be undetected?

docs/faq.rst
```diff
@@ -493,6 +493,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+corruption could break other parts of the repository.
+
+Repair is considered a destructive operation in that it will modify the
+repository in order to automatically repair some known methods of corruption.
```
ThomasWaldmann: wat?

docs/faq.rst
```diff
@@ -493,6 +493,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+
+In other words, the warning is there to emphasize that |project_name| will:
+  - Perform automated routines that will modify your backup repository
+  - May not actually fix the problem you are experiencing
+  - May further corrupt your repository
+  - May cause other problems which may not be noticable
```
ThomasWaldmann: like?

docs/faq.rst
```diff
@@ -493,6 +493,37 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+  - May cause contents of backed up data to become corrupted or changed
+
+Aside from this, if you find yourself in a position where you believe your
+repository is corrupt, this operation may result in recovering at least
+some of your backup data. You are **STRONGLY ENCOURAGED** to attempt the
+repair on a copy of the repo in case the repository becomes more unusable
+after the repair.
+
+If you are still unsure, seek :ref:`support`!
+
```
ThomasWaldmann: maybe one could specifically mention that if the problems are caused by hardware malfunctioning (e.g. a USB hub flipping bits), doing lots of repo read/write operations may cause further corruption and data loss.

docs/faq.rst
wolfpackmars2: I will review this further. I tried to understand your response to the original ticket. My first thought was that the repair function was relatively new, based on what you said, and that was the cause for the warning. 

However, after reviewing the code and several updates where Repair was implemented, that didn't make sense.  So when I wrote this, I approached the warning as making a statement that --repair will change the repository. Generally, it changes the repository for the better. Sometimes, in some unforeseen instances, it may make the repository worse.  Hence the warning and why I try to make clear that repair should be performed on a repo copy.

From what I have been able to glean from the code, repair will:
- Repair broken indexes (this requires rewriting the index, correct?).
- Attempt to correct corrupted segment files if checksums exist (which would imply modifying such segment files)

docs/faq.rst
wolfpackmars2: Is this not technically correct? Maybe "destructive operation" comes off strongly, but a repair will write changes to the repo if it detects repairable corruption, right?

docs/faq.rst
wolfpackmars2: e.g. a corrupted or modified document segment stored in the repo. Assuming --repair will modify segment files if it detects they may be corrupted.

None of this is likely ... but the basis of this entire FAQ entry is to try to explain why the warning is displayed, so I'm trying to expand on that and give all the real world problems that could possibly be caused by using an automated repair routine...

All of this, I understand, is unlikely. I debated even taking on this task because I hoped the warning would eventually go away once you had more confidence that it had been used and tested more thoroughly. Once I saw that repair was implemented 3 years ago, it changed my perspective of why the warning existed.

docs/faq.rst
ThomasWaldmann: hint: read borg repair --help, I added a quite detailled description there what it does.

docs/faq.rst
wolfpackmars2: I'll review all of this.  I think this FAQ got a little long and definitely comes across negatively, which wasn't my intention.  I wanted to try to make clear why the warning existed, and to explain "What's the worst that could happen if I run --repair?"  The hope being, that a reader would be more likely to take the time to read the part about repairing a repo copy. Especially the last line about seeking support if they need help.

You have put a lot of time into my pull requests - I appreciate it.  My intention was to save you the time and trouble of having to write these things. I hope I haven't wasted your time ... that was not my intention at all.

I did these tasks because it helps me dive into the code. I try not to write things which aren't true, so I try to review the code and existing documentation where possible to create informed documentation.

At any  rate, thank you for your time and attention - hopefully I can make some fixes which will resolve your concerns. :)

docs/faq.rst
ThomasWaldmann: well, you already said that there can be breakage in the paragraph above.
telling it is a "destructive operation" will likely lead to nobody wanting to use it.

docs/faq.rst
ThomasWaldmann: if borg check --repair notices a defect or missing chunk, it will replace it with a all-zero chunk of same length.
so the file in question will extract ok, but will have a range of zeros instead of the corrupt/missing data.
this is noticable (when looking at the file contents) and you also get some error/warning output when that "repair" is done. so I am still not sure what you mean by "not noticable". is this purely hypothetical?

docs/faq.rst
wolfpackmars2: Had a hard drive "flip bits" on me many years ago. By the time I realized what was happening, I lost most of my data. This was before "S.M.A.R.T.". And long before Borg :).

docs/faq.rst
```diff
@@ -493,6 +493,26 @@ maybe open an issue in their issue tracker. Do not file an issue in the
 If you can reproduce the issue with the proven filesystem, please file an
 issue in the |project_name| issue tracker about that.
 
+Why does running 'borgbackup check --repair' warn about data loss?
```
ThomasWaldmann: ehrm, just noticed: the command name is just 'borg', not 'borgbackup'.

docs/faq.rst
```diff
@@ -493,6 +493,26 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+------------------------------------------------------------------
+
+Repair usually works for recovering data in a corrupted archive. However,
+it's impossible to predict all modes of corruption. In some very rare
+instances, such as malfunctioning storage hardware, additional repo
+corruption may occur. If you can't afford to lose the repo, it's strongly
+recommended that you perform repair on a copy of the repo.
+
+In other words, the warning is there to emphasize that |project_name| will:
```
ThomasWaldmann: guess you need to leave one line empty below this one, so the list renders correctly.

docs/faq.rst
```diff
@@ -493,6 +493,26 @@ maybe open an issue in their issue tracker. Do not file an issue in the
...
+  - Perform automated routines that will modify your backup repository
+  - May not actually fix the problem you are experiencing
+  - In very rare cases, may further corrupt your repository
```
ThomasWaldmann: do we need a dot at the end of each of these 3 lines?
is "may" correct or is it "might"?

wolfpackmars2: I think technically they both work equally well but "might" seems more appropriate.

docs/faq.rst
```diff
@@ -493,7 +493,8 @@ maybe open an issue in their issue tracker. Do not file an issue in the
 If you can reproduce the issue with the proven filesystem, please file an
 issue in the |project_name| issue tracker about that.
 
-Why does running 'borgbackup check --repair' warn about data loss?
+
+Why does running 'borg check --repair' warn about data loss?
 ------------------------------------------------------------------
```
ThomasWaldmann: also please adjust the underlining.

docs/faq.rst
```diff
@@ -502,17 +503,18 @@ instances, such as malfunctioning storage hardware, additional repo
...
 recommended that you perform repair on a copy of the repo.
 
-In other words, the warning is there to emphasize that |project_name| will:
-  - Perform automated routines that will modify your backup repository
-  - May not actually fix the problem you are experiencing
-  - In very rare cases, may further corrupt your repository
+In other words, the warning is there to emphasize that |project_name|:
+  - will perform automated routines that modify your backup repository.
+  - Might not actually fix the problem you are experiencing.
+  - might, in very rare cases, further corrupt your repository.
```
ThomasWaldmann: not sure if the (partial) sentences in the list should start with upper or lowercase, but please at least be consistent.

wolfpackmars2: Fixed.

Best I can tell, they should be capitalized and they should not have periods. The subject (Borg) is missing from each bullet point, and adding it to the front of each line would seem repetitive.

https://english.stackexchange.com/a/163667

ThomasWaldmann:
ok, merging this.

PR #2431, opened 2017-04-19
platform.linux: get rid of "resource" module

### LINKED ISSUE #2430: get rid of "resource" module's glibc 2.13 dependency
We currently do in master (1.1) borg/platform/linux.pyx:
```
resource.getpagesize()
````

We could do:
```
>>> os.sysconf_names['SC_PAGESIZE']
30
>>> os.sysconf(30)
4096
```

The later is recommended by `getpagesize(2)`.

### PR DESCRIPTION
**Pull request by enkore**
Fixes #2430

### DIFF
--- a/src/borg/platform/linux.pyx
+++ b/src/borg/platform/linux.pyx
@@ -1,6 +1,5 @@
 import os
 import re
-import resource
 import stat
 import subprocess
 
@@ -53,6 +52,10 @@ cdef extern from "linux/fs.h":
 
 cdef extern from "sys/ioctl.h":
     int ioctl(int fildes, int request, ...)
+
+cdef extern from "unistd.h":
+    int _SC_PAGESIZE
+    long sysconf(int name)
 
 cdef extern from "string.h":
     char *strerror(int errnum)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/borg/platform/linux.pyx [hunk 1]
```diff
@@ -219,7 +222,7 @@ cdef _sync_file_range(fd, offset, length, flags):
         raise OSError(errno.errno, os.strerror(errno.errno))
     safe_fadvise(fd, offset, length, 'DONTNEED')
 
-cdef unsigned PAGE_MASK = resource.getpagesize() - 1
+cdef unsigned PAGE_MASK = sysconf(_SC_PAGESIZE) - 1
```
ThomasWaldmann: does this trigger a warning because you assign a long to an unsigned (int)?

enkore: Hm? You can always do that.

Maybe you're thinking of signed and unsigned comparisons?

ThomasWaldmann: it's getting (potentially) shorter and it's losing its sign, so there should be a warning I guess.

PR #2450, opened 2017-04-27
Fix formatting in pre-installation section

wolfpackmars2:
Some documentation was improperly formatted with #2392. This corrects those errors. Notably, links weren't formatted correct in the source docs.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -27,14 +27,14 @@ Pre-Installation Considerations
 
 Repository File System
 ~~~~~~~~~~~~~~~~~~~~~~
-:ref:data-structures-and-file-formats
+:ref:`data-structures-and-file-formats`
 - |project_name| stores data only 3 directory levels deep and uses short file and
   directory names.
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
can you check whether we need that for 1.0-maint also and do a PR if so?

wolfpackmars2:
1.0-maint is clean. This only affected latest.

Cross references still appear to not be rendering correctly in latest.

enkore:
There are two kinds of references you can make: In the same document any heading can be referenced directly, e.g.
```
`Installation from source`_
```
while references between documents need to use explicitly defined anchors:
```
# document A
.. _my-anchor:

My Heading
========

...

```

```
# document B

See :ref:`my-anchor` for ...
```
The reference is replaced with the heading following the anchor by default. The usual substitution can be done with either syntax:
```
See :ref:`this section <my-anchor>` for...

See `here <My Heading>`_ for ...
```

Issue #2469, opened 2017-05-01
--exclude is broken


enkore:
**Task:** --exclude and --exclude-from were inadvertently changed to recurse during the latest --patterns fixup. Restore correct behaviour.

Reported by @jdchristensen

PR #2471, opened 2017-05-01
Move patterns to module

enkore:
Preparations for #2469


ThomasWaldmann:
the first changeset is just moving stuff without change?

enkore:
Yes.

PR #2472, opened 2017-05-01
docs: disable smartypants

enkore:


ThomasWaldmann:
what's the point?

docs/conf.py
```diff
@@ -148,7 +148,9 @@
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+#
+# This is disabled to avoid mangling --options-that-appear-in-texts.
```
enkore: .

ThomasWaldmann: is this a problem if they are written like this:
```
bla bla ``--option`` bla bla
```
?

enkore: There are a *lot* of occurrences that doesn't use literals for them. There's also the issue of rST syntax in --help output (and "borg help XYZ").

ThomasWaldmann: ah, ok.

PR #2476, opened 2017-05-02
Create borg.algorithms and borg.crypto packages

enkore:
Using the historic low in merge-able pull requests to do some restructuring. PyCharm was so nice to remove some unused imports as well.

algorithms = *implementations* of said entities. That's why blake2/ is there and not in crypto/, which otherwise only has abstractions and management things, no implementations.

Checked sdist, looks complete.


ThomasWaldmann:
I'ld appreciate if you stop moving stuff around as long as the crypto-aead PR is open (= until we start with 1.2).

Recently, I am investing quite some hours rebasing that thing to fix conflicts and that isn't much fun.
It's conflicting again since you extracted the patterns stuff, maybe also due to this one.

Also, "algorithms". What's not an algorithm? Maybe that is a bit too generic?

PR #2477, opened 2017-05-02
add test for preserved intermediate folder permissions

edgimar:
This tests whether the permissions metadata is preserved when a folder is excluded but still recursed into to find a matching file in a subfolder.

To my surprise, it passes, which seems to indicate that the concern raised by @ThomasWaldmann in PR #2322 may not be a problem after all.

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -960,8 +960,8 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         self.assert_not_in('input/x/a', output)
         self.assert_in('A input/y/foo_y', output)
 
-    def test_create_pattern_intermediate_folder_permissions(self):
-        """test for correct metadata when patterns exclude a parent folder but include a child"""
+    def test_create_pattern_intermediate_folders_first(self):
+        """test that intermediate folders appear first when patterns exclude a parent folder but include a child"""
         self.patterns_file_path2 = os.path.join(self.tmpdir, 'patterns2')
         with open(self.patterns_file_path2, 'wb') as fd:
             fd.write(b'+ input/x/a\n+ input/x/b\n- input/x*\n')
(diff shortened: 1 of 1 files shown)


edgimar:
btw, while writing this test I noticed some strange (possibly incorrect) code at https://github.com/edgimar/borg/blob/cf80cc5afe9586291510e74cc2f53bed6085bd17/src/borg/testsuite/archiver.py#L2735 and https://github.com/edgimar/borg/blob/cf80cc5afe9586291510e74cc2f53bed6085bd17/src/borg/testsuite/archiver.py#L2761 -- what is the purpose of ORing the stat values here?

ThomasWaldmann:
@edgimar right, the stat.S_IF* part at both places looks wrong.

stat.S_IF* is for interpreting the type-part of the mode and can not be used to set it (AFAIK), so that should be removed at both places.

Separate PR please.

edgimar:
any thoughts on whether this test is OK?  Is it legitimately testing for metadata preservation?  If so, does it mean that we don't need to do anything to try to ensure that "intermediate" nested folders are explicitly included in the backup?

src/borg/testsuite/archiver.py
```diff
@@ -960,6 +960,31 @@ def test_create_pattern_exclude_folder_no_recurse(self):
         self.assert_not_in('input/x/a', output)
         self.assert_in('A input/y/foo_y', output)
 
+    def test_create_pattern_intermediate_folder_permissions(self):
+        """test for correct metadata when patterns exclude a parent folder but include a child"""
+        self.patterns_file_path2 = os.path.join(self.tmpdir, 'patterns2')
```
ThomasWaldmann: is "self." needed for this?

src/borg/testsuite/archiver.py
```diff
@@ -960,6 +960,31 @@ def test_create_pattern_exclude_folder_no_recurse(self):
...
+        with open(self.patterns_file_path2, 'wb') as fd:
+            fd.write(b'+ input/x/a\n+ input/x/b\n- input/x*\n')
+
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+
+        self.create_regular_file('x/a/foo_a', size=1024 * 80)
+        self.create_regular_file('x/b/foo_b', size=1024 * 80)
+        with changedir('input'):
+            os.chmod('x/a', 0o750)
+            os.chmod('x/b', 0o700)
```
ThomasWaldmann: maybe use 0o600 here so we have a difference in owner x bit, which is usually never masked by umask.

src/borg/testsuite/archiver.py
```diff
@@ -960,6 +960,31 @@ def test_create_pattern_exclude_folder_no_recurse(self):
...
+                     self.repository_location + '::test', '.')
+
+        # extract the archive and verify that the "intermediate" folder
+        # permissions have been preserved.
+        with changedir('output'):
+            self.cmd('extract', self.repository_location + '::test')
+
+            for fname, mode in [('x/a', 0o750), ('x/b', 0o700)]:
+                st = os.stat(fname)
+                self.assert_equal(st.st_mode & 0o777, mode)
```
ThomasWaldmann: hmm, wasn't the point of this test to make an assertion about 'x' (which is excluded) rather than 'x/a' and 'x/b', which both are normally included files?

enkore:
I'd approach this differently. Tests using chmod/chown/chgrp and such frequently fall apart (though it does pass) for various reasons.

Since the test wants to assert that the parent directories are included, I'd just `borg list` and check that the directories are in there before the files.

ThomasWaldmann:
@enkore yup, sounds good and easy.

ThomasWaldmann:
@edgimar can you update your PR?

edgimar:
@ThomasWaldmann done.

enkore:
Thanks!

PR #2478, opened 2017-05-04
Start fakeroot faked in debug mode - fixes EISDIR issues

PlasmaPower:
I have absolutely no idea why, but this seems to reliably fix the EISDIR test issues with no additional verbosity (except 2 messages at startup). I also didn't notice any significant slow downs.

### DIFF
--- a/scripts/faked-debug.sh
+++ b/scripts/faked-debug.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+if which faked; then
+  faked --debug "$@"
+else
+  faked-sysv --debug "$@"
+fi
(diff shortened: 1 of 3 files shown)


PlasmaPower:
On some platforms `fakeroot -f 'faked --debug'` works, but on others the debug flag doesn't get passed on to faked. I also have no idea why that happens, but the small shell script mitigates it.

PlasmaPower:
I had some slight Travis path/binary name issues, but they should be fixed now.

PlasmaPower:
I think the slight coverage difference is unrelated to this PR.

enkore:
Codecov almost never reports meaningful numbers for some reason.

Anyhow, I restarted the build a bunch of times and it never failed, so looks like you found a workaround! :)

scripts/faked-debug.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
ThomasWaldmann: I'ld appreciate some comment here about why this is needed. Otherwise nobody will remember in a year.

PR #2489, opened 2017-05-05
consider repokey w/o passphrase == unencrypted

### LINKED ISSUE #2169: Interactive repository swap
From #2160 

Note that when using BORG_PASSPHRASE the attacker cannot swap the *entire*
repository against a new repository with e.g. repokey mode and no passphrase,
because Borg will abort access when BORG_PASSPRHASE is incorrect.

However, interactively a user might not notice this kind of attack
immediately, if she assumes that the reason for the absent passphrase
prompt is a set BORG_PASSPHRASE.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/42112412-interactive-repository-swap)

note: bounty is for checking /fixing all mentioned "repo swap" issues.

enkore:
Fixes #2169

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -128,6 +128,15 @@ The best check that everything is ok is to run a dry-run extraction::
 Changelog
 =========
 
+Version 1.1.0b6 (unreleased)
+----------------------------
+
+Compatibility notes:
+
+- Repositories in a repokey mode with a blank passphrase are now treated
+  as unencrypted repositories for security checks
+  (e.g. BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK).
+
 Version 1.1.0b5 (2017-04-30)
 ----------------------------
 
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -130,7 +130,7 @@ class SecurityManager:
             self.save(manifest, key, cache)
 
     def assert_access_unknown(self, warn_if_unencrypted, key):
-        if warn_if_unencrypted and isinstance(key, PlaintextKey) and not self.known():
+        if warn_if_unencrypted and not key.passphrase_protected and not self.known():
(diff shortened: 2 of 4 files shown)


docs/changes.rst
```diff
@@ -128,6 +128,15 @@ The best check that everything is ok is to run a dry-run extraction::
...
 
+Version 1.1.0b6 (unreleased)
+----------------------------
+
+Compatibility notes:
+
+- Repositories in a repokey mode with a blank passphrase are now treated
+  as unencrypted repositories for security checks
+  (e.g. BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK).
+
```
ThomasWaldmann: Maybe also list it below "Fixes:" as a security-related fix?

While it is not strictly a borg bug, a user could be tricked easily enough with it.

enkore: _Compatibility Notes_ is usually at the top, even before _Fixes_. Perhaps something to consider for condensed 1.1 released notes?

ThomasWaldmann: My comment was not just about "having it at the top", but about explicitly saying it is security related fix.

enkore: Ah. Perhaps it'd still be better in that case to amend a succinct description of the issue/scenario to the changelog entry, not necessarily duplicating the text.

enkore: What do you think about this?

> - Repository in a  ...
>
> Previously it was possible to be tricked into using a repository (repokey,
> blank passphrase) swapped for an encrypted repository, if one assumed
> that BORG_PASSPHRASE was set when it was not. This does not affect scripts
> etc. using BORG_PASSPHRASE.
>

ThomasWaldmann: "in a ..." - what do you mean?

Hmm, is the order of arguments of "swapped for" correct here?
Maybe use an easier to understand form.

enkore: "Repository in a ..." = Existing changelog entry (see above)

PR #2496, opened 2017-05-09
hashindex: Use Python I/O

### LINKED ISSUE #1809: hashindex size limitations
The size of it is limited to 2**31 entries, probably not important right now (since that would need at least ~180 GB of usable memory). [hash_sizes](https://github.com/borgbackup/borg/blob/master/src/borg/_hashindex.c#L49)

We use normal C I/O currently, which will be limited to 32 bit offsets depending on platform (I think Linux 32 bits and Windows?). This would limit the repository index to ~60 million entries and the chunks cache to ~49 million entries. This is already addressed by the patches in #1688, since Python I/O always uses large offsets if available as far as I'm aware.

enkore:
- Preparation for #1688 / #1101
- Support hash indices >2 GB (fixes #1809)
- Better error reporting

---

Split off from #1688 for easier integration.

The downside is that hashindex.c now includes Python, so one needs to supply the appropriate -lpythonmX.Y etc. to compile it for stand-alone testing. The code is also a bit more complex, because it's (supposed to) correctly clean up references to objects on errors.

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -274,7 +274,7 @@ hashindex_read(PyObject *file_py)
 
     header_bytes = PyObject_CallMethod(file_py, "read", "n", (Py_ssize_t)sizeof(HashHeader));
     if(!header_bytes) {
-        /* Python exception occurred */
+        assert(PyErr_Occurred());
         goto fail;
     }
 
@@ -351,7 +351,7 @@ hashindex_read(PyObject *file_py)
      */
     bucket_bytes = PyObject_CallMethod(file_py, "read", "n", buckets_length);
     if(!bucket_bytes) {
-        /* Python exception occurred; clean up and return */
+        assert(PyErr_Occurred());
         goto fail_release_header_buffer;
     }
     bytes_read = PyBytes_Size(bucket_bytes);
@@ -377,6 +377,7 @@ hashindex_read(PyObject *file_py)
         /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
         if(!hashindex_resize(index, index->num_buckets)) {
             PyErr_Format(PyExc_ValueError, "Failed to rebuild table");
+            goto fail_free_buckets;
         }
     }
 
@@ -385,6 +386,7 @@ hashindex_read(PyObject *file_py)
      * Also note that the buffer in index->buckets_buffer holds a reference to buckets_bytes.
      */
 
+fail_free_buckets:
     if(PyErr_Occurred()) {
         hashindex_free_buckets(index);
     }
@@ -446,7 +448,7 @@ hashindex_free(HashIndex *index)
 static void
 hashindex_write(HashIndex *index, PyObject *file_py)
 {
-    PyObject *length_object, *header_view, *buckets_view;
+    PyObject *length_object, *buckets_view;
     Py_ssize_t length;
     Py_ssize_t buckets_length = (Py_ssize_t)index->num_buckets * index->bucket_size;
     HashHeader header = {
@@ -457,14 +459,7 @@ hashindex_write(HashIndex *index, PyObject *file_py)
         .value_size = index->value_size
     };
 
-    /* Note: explicitly construct view; BuildValue can convert (pointer, length) to Python objects, but copies them for doing so */
-    header_view = PyMemoryView_FromMemory((char*)&header, sizeof(HashHeader), PyBUF_READ);
-    if(!header_view) {
-        /* Exception raised */
-        return;
-    }
-    length_object = PyObject_CallMethod(file_py, "write", "O", header_view);
-    Py_DECREF(header_view);
+    length_object = PyObject_CallMethod(file_py, "write", "y#", &header, (int)sizeof(HashHeader));
     if(PyErr_Occurred()) {
         return;
     }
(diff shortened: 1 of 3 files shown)


src/borg/repository.py
```diff
@@ -324,9 +324,8 @@ def open_index(self, transaction_id, auto_recover=True):
         index_path = os.path.join(self.path, 'index.%d' % transaction_id).encode('utf-8')
         try:
             return NSIndex.read(index_path)
-        except RuntimeError as error:
-            assert str(error) == 'hashindex_read failed'  # everything else means we're in *deep* trouble
-            logger.warning('Repository index missing or corrupted, trying to recover')
+        except (ValueError, OSError) as exc:
```
enkore: The diff also suggests that the previous error handling was incomplete, see above:

```py
-                if errno:
-                    PyErr_SetFromErrnoWithFilename(OSError, path)
```

Which was not handled here.

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
-        }
-        else {
-            EPRINTF_MSG_PATH(path, "fread header failed (expected %ju, got %ju)",
-                             (uintmax_t) sizeof(HashHeader), (uintmax_t) bytes_read);
-        }
-        goto fail;
+    PyObject *header_bytes = PyObject_CallMethod(file, "read", "n", sizeof(HashHeader));
+    if(!header_bytes) {
+        /* Python exception occurred; just return */
+        return 0;
```
ThomasWaldmann: maybe there could be a "goto fail" jumping to the "return index" (which is NULL then)"?

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
+        /* TypeError, not a bytes() object */
+        goto fail_decref_header;
     }
-    if((length = ftell(fd)) < 0) {
-        EPRINTF_PATH(path, "ftell failed");
-        goto fail;
+    if(bytes_read != sizeof(HashHeader)) {
+        /* Truncated file */
+        /* Note: %zd is the format for Py_ssize_t */
+        PyErr_Format(PyExc_ValueError, "Could not read header (expected %zd, but read %zd bytes)", sizeof(HashHeader), bytes_read);
```
ThomasWaldmann: sizeof() result also works with %zd?

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
+        goto fail_decref_header;
     }
-    if(fseek(fd, sizeof(HashHeader), SEEK_SET) < 0) {
-        EPRINTF_PATH(path, "fseek failed");
-        goto fail;
+
+    /* Find length of file */
+    PyObject *length_object = PyObject_CallMethod(file, "seek", "ni", 0, SEEK_END);
```
ThomasWaldmann: maybe it would be a nice convention for *.c to always have _obj suffix if it is a python object.

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
+     * we have them backed by a Python bytes() object instead, and go through
+     * Python I/O.
+     *
+     * Note: Issuing read(buckets_length) is okay here, because buffered readers
+     * will issue multiple underlying reads if necessary. This supports indices
+     * >2 GB on Linux. We also compare lengths later.
+     */
+    PyObject *bucket_bytes = PyObject_CallMethod(file, "read", "n", buckets_length);
+    if(!bucket_bytes) {
+        /* Python exception occured; clean up and return */
```
ThomasWaldmann: occurred

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
+
+    index->num_entries = _le32toh(header->num_entries);
+    index->num_buckets = _le32toh(header->num_buckets);
+    index->key_size = header->key_size;
+    index->value_size = header->value_size;
     index->bucket_size = index->key_size + index->value_size;
     index->lower_limit = get_lower_limit(index->num_buckets);
     index->upper_limit = get_upper_limit(index->num_buckets);
-    index->min_empty = get_min_empty(index->num_buckets);
-    index->num_empty = count_empty(index);
```
ThomasWaldmann: why did you delete these 2 lines? in 370 the values are used.

src/borg/_hashindex.c
```diff
@@ -248,99 +262,130 @@ count_empty(HashIndex *index)
...
+    Py_DECREF(bucket_bytes);
+fail_release_header_buffer:
+    PyBuffer_Release(&header_buffer);
+fail_free_index:
+    if(PyErr_Occurred()) {
+        free(index);
     }
+fail_decref_header:
+    Py_DECREF(header_bytes);
     return index;
```
ThomasWaldmann: "fail:" label above this line?

src/borg/_hashindex.c
```diff
@@ -349,7 +352,7 @@ hashindex_read(PyObject *file)
      */
     PyObject *bucket_bytes = PyObject_CallMethod(file, "read", "n", buckets_length);
     if(!bucket_bytes) {
-        /* Python exception occured; clean up and return */
+        /* Python exception occurred; clean up and return */
```
ThomasWaldmann: another one in 382

src/borg/_hashindex.c
```diff
@@ -248,99 +262,144 @@ count_empty(HashIndex *index)
...
     index->num_empty = count_empty(index);
+
     if(index->num_empty < index->min_empty) {
         /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
         if(!hashindex_resize(index, index->num_buckets)) {
-            free(index->buckets);
-            free(index);
-            index = NULL;
-            goto fail;
+            PyErr_Format(PyExc_ValueError, "Failed to rebuild table");
```
enkore: Good thinking

ThomasWaldmann:
Removing the backport label, see #2498.

PR #2500, opened 2017-05-10
Verify most operations against securitymanager

### LINKED ISSUE #2487: Decouple SecurityManager and Cache and always check with SecurityManager
Currently SecurityManager is only queried if the Cache is used, which elides checks for various commands, e.g. list, key management, extract, mount, ...

The debug commands may be excluded.

enkore:
Fixes #2487 

TODO: Check if some corner case(s) were overlooked. Some functions are now separated and can now be covered with additional tests (the reconanlinzlianoztion thing).

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -32,6 +32,25 @@ FileCacheEntry = namedtuple('FileCacheEntry', 'age inode size mtime chunk_ids')
 
 
 class SecurityManager:
+    """
+    Tracks repositories. Ensures that nothing bad happens (repository swaps,
+    replay attacks, unknown repositories etc.).
+
+    This is complicated by the Cache being initially used for this, while
+    only some commands actually use the Cache, which meant that other commands
+    did not perform these checks.
+
+    Further complications were created by the Cache being a cache, so it
+    could be legitimately deleted, which is annoying because Borg didn't
+    recognize repositories after that.
+
+    Therefore a second location, the security database (see get_security_dir),
+    was introduced which stores this information. However, this means that
+    the code has to deal with a cache existing but no security DB entry,
+    or inconsistencies between the security DB and the cache which have to
+    be reconciled, and also with no cache existing but a security DB entry.
+    """
+
     def __init__(self, repository):
         self.repository = repository
         self.dir = get_security_dir(repository.id_str)
@@ -66,19 +85,19 @@ class SecurityManager:
         with open(self.manifest_ts_file, 'w') as fd:
             fd.write(manifest.timestamp)
 
-    def assert_location_matches(self, cache_config):
+    def assert_location_matches(self, cache_config=None):
         # Warn user before sending data to a relocated repository
         try:
             with open(self.location_file) as fd:
                 previous_location = fd.read()
-            logger.debug('security: read previous_location %r', previous_location)
+            logger.debug('security: read previous location %r', previous_location)
         except FileNotFoundError:
-            logger.debug('security: previous_location file %s not found', self.location_file)
+            logger.debug('security: previous location file %s not found', self.location_file)
             previous_location = None
         except OSError as exc:
             logger.warning('Could not read previous location file: %s', exc)
             previous_location = None
-        if cache_config.previous_location and previous_location != cache_config.previous_location:
+        if cache_config and cache_config.previous_location and previous_location != cache_config.previous_location:
             # Reconcile cache and security dir; we take the cache location.
             previous_location = cache_config.previous_location
             logger.debug('security: using previous_location of cache: %r', previous_location)
@@ -95,9 +114,10 @@ class SecurityManager:
             logger.debug('security: updating location stored in cache and security dir')
             with open(self.location_file, 'w') as fd:
                 fd.write(repository_location)
-            cache_config.save()
+            if cache_config:
+                cache_config.save()
 
-    def assert_no_manifest_replay(self, manifest, key, cache_config):
+    def assert_no_manifest_replay(self, manifest, key, cache_config=None):
         try:
             with open(self.manifest_ts_file) as fd:
                 timestamp = fd.read()
@@ -108,7 +128,8 @@ class SecurityManager:
         except OSError as exc:
             logger.warning('Could not read previous location file: %s', exc)
             timestamp = ''
-        timestamp = max(timestamp, cache_config.timestamp or '')
+        if cache_config:
+            timestamp = max(timestamp, cache_config.timestamp or '')
         logger.debug('security: determined newest manifest timestamp as %s', timestamp)
         # If repository is older than the cache or security dir something fishy is going on
         if timestamp and timestamp > manifest.timestamp:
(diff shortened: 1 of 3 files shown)


src/borg/cache.py
```diff
@@ -77,25 +97,27 @@ def assert_location_matches(self, cache):
...
                 raise Cache.RepositoryAccessAborted()
             # adapt on-disk config immediately if the new location was accepted
             logger.debug('security: updating location stored in cache and security dir')
             with open(self.location_file, 'w') as fd:
-                fd.write(cache.repository._location.canonical_path())
-            cache.begin_txn()
-            cache.commit()
+                fd.write(repository_location)
+            if cache_config:
+                cache_config.save()
```
enkore: = Now only the config is updated, not the entire cache transacted (=chunks cache, files cache copied then rewritten then copies deleted).

src/borg/cache.py
```diff
@@ -115,29 +138,150 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+            and \
+            cl.path and rl.path and \
+            cl.path.startswith('/~/') and rl.path.startswith('/./') and cl.path[3:] == rl.path[3:]:
+        # everything is same except the expected change in relative path canonicalization,
+        # update previous_location to avoid warning / user query about changed location:
+        return repo_location
+    else:
+        return cache_location
+
+
```
enkore: ^^^

This is all really far more complex than this kind of code ought to be, but given the constraints I don't really see many ways to reduce the intrinsic complexity of that stuff. (So yes, pedantically: the code is complex because the requirements are complex).

src/borg/cache.py
```diff
@@ -240,66 +386,27 @@ def create(self):
...
-        self.config = configparser.ConfigParser(interpolation=None)
-        config_path = os.path.join(self.path, 'config')
-        self.config.read(config_path)
-        self._check_upgrade(config_path)
-        self.id = self.config.get('cache', 'repository')
-        self.manifest_id = unhexlify(self.config.get('cache', 'manifest'))
-        self.timestamp = self.config.get('cache', 'timestamp', fallback=None)
-        self.key_type = self.config.get('cache', 'key_type', fallback=None)
-        self.previous_location = self.config.get('cache', 'previous_location', fallback=None)
+        # TODO config open
```
ThomasWaldmann: ^^^

src/borg/cache.py
```diff
@@ -178,25 +322,27 @@ def __init__(self, repository, key, manifest, path=None, sync=True, do_files=Fal
...
+
         # Warn user before sending data to a never seen before unencrypted repository
         if not os.path.exists(self.path):
-            self.security_manager.assert_access_unknown(warn_if_unencrypted, key)
+            self.security_manager.assert_access_unknown(warn_if_unencrypted, manifest, key)
             self.create()
-        self.open(lock_wait=lock_wait)
+
+        self.open()
+        self.security_manager.assert_secure(manifest, key, cache_config=self.cache_config)
```
ThomasWaldmann: as you moved it outside the try/except, an exception in here will not call self.close().

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+        if cache_config and cache_config.key_type is not None and cache_config.key_type != str(key.TYPE):
             raise Cache.EncryptionMethodMismatch()
         if self.known() and not self.key_matches(key):
             raise Cache.EncryptionMethodMismatch()
 
-    def assert_secure(self, manifest, key, cache):
-        self.assert_location_matches(cache)
-        self.assert_key_type(key, cache)
-        self.assert_no_manifest_replay(manifest, key, cache)
+    def assert_secure(self, manifest, key, *, cache_config=None, warn_if_unencrypted=True):
```
ThomasWaldmann: `, *,` ?
so that is for callers giving a cache object there and that is not used, but we rather load the cache_config here?

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+        if cache_config:
+            self._assert_secure(manifest, key, cache_config)
+        else:
+            cache_config = CacheConfig(self.repository)
+            if cache_config.exists():
+                with cache_config:
+                    self._assert_secure(manifest, key, cache_config)
+            else:
+                self._assert_secure(manifest, key)
+        logger.debug('security: all checks ok')
```
ThomasWaldmann: that's a rather wide claim. maybe "repository security"?

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
-            if not yes(msg, false_msg="Aborting.", invalid_msg="Invalid answer, aborting.",
-                       retry=False, env_var_override='BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK'):
+            allow_access = not warn_if_unencrypted or yes(msg, false_msg="Aborting.",
+                invalid_msg="Invalid answer, aborting.",
+                retry=False, env_var_override='BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK')
+            if allow_access:
+                if warn_if_unencrypted:
+                    logger.debug('security: saving state for unknown unencrypted repository (explicitly granted)')
+                else:
+                    logger.debug('security: saving state for unknown unencrypted repository')
```
ThomasWaldmann: how about "security: allowing access to unknown unencrypted repository [(explicitly granted)] and remembering decision"

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def exists(self):
+        return os.path.exists(self.config_path)
+
+    def create(self):
+        assert not os.path.exists(self.config_path)
```
ThomasWaldmann: assert not self.exists()

ThomasWaldmann: or even:
```
if self.exists():
    raise ...
```

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+            config.write(fd)
+
+    def open(self):
+        self.lock = Lock(os.path.join(self.path, 'lock'), exclusive=True, timeout=self.lock_wait,
+                         kill_stale_locks=hostname_is_unique()).acquire()
+        self.load()
+
+    def load(self):
+        self._config = configparser.ConfigParser(interpolation=None)
+        config_path = os.path.join(self.path, 'config')
```
ThomasWaldmann: we already have computed self.config_path.

src/borg/cache.py
```diff
@@ -115,29 +138,152 @@ def assert_no_manifest_replay(self, manifest, key, cache):
...
+            self.previous_location = None
+
+    def save(self, manifest=None, key=None):
+        if manifest:
+            self._config.set('cache', 'manifest', manifest.id_str)
+            self._config.set('cache', 'timestamp', manifest.timestamp)
+        if key:
+            self._config.set('cache', 'key_type', str(key.TYPE))
+        self._config.set('cache', 'previous_location', self.repository._location.canonical_path())
+        with SaveFile(os.path.join(self.path, 'config')) as fd:
```
ThomasWaldmann: self.config_path

src/borg/cache.py
enkore: `, *,` ends the positional argument list. Arguments beyond the `*` marker are only accepted as keyword arguments.

enkore:
Squashed; before squashing: 4ca5892

PR #2501, opened 2017-05-11
serve: ignore --append-only when creating a repository

### LINKED ISSUE #2413: Borg init configuring the repository as read-only when served with "--append-only"
I spent a couple of hours figuring out why my repository size wasn't shrinking despite repeated calls to `prune` before I realised what's going on.

My borg server is configured as `borg serve --append-only`, to protect against backups being deleted by a hacked client. I use a seperate repository for each backup client, so `borg init` is always ran on the client through the SSH connection. This apparently causes the repository to be "permanently" read-only, with `append_only = 1` being written to the repository config. 

This means diskspace cannot be recovered from the repository, even if `borg prune` is ran on the server itself. I would expect `borg init` to create a regular repository, even if the server is running in append-only mode (unless, of course, `-a` is used).

enkore:
Fixes #2413

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -136,6 +136,9 @@ Compatibility notes:
 - Repositories in a repokey mode with a blank passphrase are now treated
   as unencrypted repositories for security checks
   (e.g. BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK).
+- Running "borg init" via a "borg serve --append-only" server will *not* create
+  an append-only repository anymore. Use "borg init --append-only" to initialize
+  an append-only repository.
 
 Version 1.1.0b5 (2017-04-30)
 ----------------------------
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -180,6 +180,10 @@ class RepositoryServer:  # pragma: no cover
     def __init__(self, restrict_to_paths, append_only):
         self.repository = None
         self.restrict_to_paths = restrict_to_paths
+        # This flag is parsed from the serve command line via Archiver.do_serve,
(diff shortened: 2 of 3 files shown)


docs/changes.rst
```diff
@@ -136,6 +136,9 @@ Compatibility notes:
 - Repositories in a repokey mode with a blank passphrase are now treated
   as unencrypted repositories for security checks
   (e.g. BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK).
+- Running borg init`` via a borg serve --append-only server will *not* create
```
ThomasWaldmann: remove remainder of code markup after borg init.

src/borg/remote.py
```diff
@@ -345,8 +345,11 @@ def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, ap
                     break
             else:
                 raise PathNotAllowed(path)
+        # borg init on borg serve --append-only does not create an append only repo,
+        # while borg init --append-only does, regardless of the --append-only flag for serve.
```
ThomasWaldmann: maybe add some double quotes around the borg commands to make sentence easier to parse?

src/borg/remote.py
```diff
@@ -345,8 +345,11 @@ def open(self, path, create=False, lock_wait=None, lock=True, exclusive=None, ap
...
+        append_only = (not create and self.append_only) or append_only
```
ThomasWaldmann: I suspect you have analyzed the precise meaning of self.append_only vs. append_only to make sure this condition is correct, so maybe add some docs to the respective places about what they mean precisely (how they get True).

PR #2502, opened 2017-05-11
Implement IntegrityCheckedFile

enkore:
This is based on much earlier work from October 2016 by me, but is
overall simplified and changed terminology (from "signing" to
hashing and integrity checking).

See #1688 for the full history.

### DIFF
--- a/src/borg/crypto/file_integrity.py
+++ b/src/borg/crypto/file_integrity.py
@@ -54,12 +54,12 @@ class SHA512FileHashingWrapper(FileLikeWrapper):
     are illegal.
     """
 
-    ALGORITHM = 'SHA-512'
+    ALGORITHM = 'SHA512'
 
     def __init__(self, backing_fd, write):
         self.fd = backing_fd
         self.writing = write
-        self.hash = hashlib.sha512()
+        self.hash = hashlib.new(self.ALGORITHM)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is None:
@@ -70,7 +70,7 @@ class SHA512FileHashingWrapper(FileLikeWrapper):
         """
         Write *data* to backing file and update internal state.
         """
-        n = self.fd.write(data)
+        n = super().write(data)
         self.hash.update(data)
         return n
 
@@ -78,17 +78,17 @@ class SHA512FileHashingWrapper(FileLikeWrapper):
         """
         Read *data* from backing file (*n* has the usual meaning) and update internal state.
         """
-        data = self.fd.read(n)
+        data = super().read(n)
         self.hash.update(data)
         return data
 
-    def digest(self):
+    def hexdigest(self):
         """
-        Return current digest bytes.
+        Return current digest bytes as hex-string.
 
         Note: this can be called multiple times.
         """
-        return self.hash.digest()
+        return self.hash.hexdigest()
 
     def update(self, data: bytes):
         self.hash.update(data)
@@ -160,7 +160,7 @@ class IntegrityCheckedFile(FileLikeWrapper):
             return
         self.hasher.update(partname.encode())
         self.hasher.hash_length(seek_to_end=is_final)
-        digest = bin_to_hex(self.hasher.digest())
+        digest = self.hasher.hexdigest()
         if self.writing:
             self.digests[partname] = digest
         elif self.digests and not compare_digest(self.digests.get(partname, ''), digest):
--- a/src/borg/crypto/file_integrity.py
+++ b/src/borg/crypto/file_integrity.py
@@ -1,0 +1,183 @@
+import hashlib
+import io
+import json
+import os
+from hmac import compare_digest
+
+from ..helpers import IntegrityError
+from ..helpers import bin_to_hex
+from ..logger import create_logger
+
+logger = create_logger()
+
+
+class FileLikeWrapper:
+    def __enter__(self):
+        self.fd.__enter__()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.fd.__exit__(exc_type, exc_val, exc_tb)
+
+    def tell(self):
+        return self.fd.tell()
+
+    def seek(self, offset, whence=io.SEEK_SET):
+        return self.fd.seek(offset, whence)
+
+    def write(self, data):
+        return self.fd.write(data)
+
+    def read(self, n=None):
+        return self.fd.read(n)
+
+    def flush(self):
+        self.fd.flush()
+
+    def fileno(self):
+        return self.fd.fileno()
+
+
+class SHA512FileHashingWrapper(FileLikeWrapper):
+    """
+    Wrapper for file-like objects that computes a hash on-the-fly while reading/writing.
+
+    WARNING: Seeks should only be used to query the size of the file, not
+    to skip data, because skipped data isn't read and not hashed into the digest.
+
+    Similarly skipping while writing to create sparse files is also not supported.
+
+    Data has to be read/written in a symmetric fashion, otherwise different
+    digests will be generated.
+
+    Note: When used as a context manager read/write operations outside the enclosed scope
+    are illegal.
+    """
+
+    ALGORITHM = 'SHA-512'
+
+    def __init__(self, backing_fd, write):
+        self.fd = backing_fd
+        self.writing = write
+        self.hash = hashlib.sha512()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is None:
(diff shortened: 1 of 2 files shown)


src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+    Note: When used as a context manager read/write operations outside the enclosed scope
+    are illegal.
+    """
+
+    ALGORITHM = 'SHA-512'
+
+    def __init__(self, backing_fd, write):
+        self.fd = backing_fd
+        self.writing = write
+        self.hash = hashlib.sha512()
```
ThomasWaldmann: hashlib.new(self.ALGORITHM) or so?

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+            # TODO: When we're reading but don't have any digests, i.e. no integrity file existed,
+            # TODO: then we could just short-circuit.
+
+    def hash_filename(self, filename=None):
+        # Hash the name of the file, but only the basename, ie. not the path.
+        # In Borg the name itself encodes the context (eg. index.N, cache, files),
+        # while the path doesn't matter, and moving e.g. a repository or cache directory is supported.
+        # Changing the name however imbues a change of context that is not permissible.
+        filename = os.path.basename(filename or self.path)
+        self.hasher.update(('%10d' % len(filename)).encode())
```
ThomasWaldmann: is using a fixed width format like %10d important here? if so, 10 digits might be not enough, that is only 9.99GB.

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+                if algorithm != hasher.ALGORITHM:
+                    logger.warning('Cannot verify integrity of %s: Unknown algorithm %r', path, algorithm)
+                    return
+                digests = integrity_file['digests']
+                # Require at least presence of the final digest
+                digests['final']
+                return digests
+        except FileNotFoundError:
+            logger.info('No integrity file found for %s', path)
+        except (OSError, ValueError, TypeError, KeyError) as e:
```
ThomasWaldmann: quite a wide scope exc handler, not sure if problematic.

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+            logger.warning('Could not read integrity file for %s: %s', path, e)
```
ThomasWaldmann: maybe say "Could not process ..." as long as catching all sorts of issues.

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+            raise FileIntegrityError(path)
+
+    def hash_part(self, partname, is_final=False):
+        if not self.writing and not self.digests:
+            return
+        self.hasher.update(partname.encode())
+        self.hasher.hash_length(seek_to_end=is_final)
+        digest = bin_to_hex(self.hasher.digest())
```
ThomasWaldmann: hasher.hexdigest?

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.writing:
+            self.digests[partname] = digest
+        elif self.digests and not compare_digest(self.digests.get(partname, ''), digest):
```
ThomasWaldmann: is self.digests term needed? if it is an empty dict, the compare would fail anyway.

src/borg/crypto/file_integrity.py
```diff
@@ -0,0 +1,183 @@
...
+            self.hash_part('final', is_final=True)
+        self.hasher.__exit__(exc_type, exc_val, exc_tb)
+        if exception:
+            return
+        if self.writing:
+            with open(self.integrity_file_path(self.path), 'w') as fd:
+                json.dump({
+                    'algorithm': self.hasher.ALGORITHM,
+                    'digests': self.digests,
+                }, fd)
```
ThomasWaldmann: indent=4, sort_keys=True?

makes some json dicts in the tests multi-line, but easier to read for more than a few short strings.

ThomasWaldmann: I also had the vague idea if we could add a version number for the data structure in the data file here, so we could introduce incompatible changes more easily? The metadata itself could also have a version number.

src/borg/testsuite/file_integrity.py
```diff
@@ -0,0 +1,152 @@
...
+import pytest
+
+from ..crypto.file_integrity import IntegrityCheckedFile, FileIntegrityError
+
+
+class TestReadIntegrityFile:
+    def test_no_integrity(self, tmpdir):
+        protected_file = tmpdir.join('file')
+        protected_file.write('1234')
+        assert not IntegrityCheckedFile.read_integrity_file(str(protected_file), None)
```
ThomasWaldmann: assert ... is None?

src/borg/testsuite/file_integrity.py
```diff
@@ -0,0 +1,152 @@
...
+            IntegrityCheckedFile.read_integrity_file(str(protected_file), None)
+
+    def test_unknown_algorithm(self, tmpdir):
+        class SomeHasher:
+            ALGORITHM = 'HMAC_FOOHASH9000'
+
+        protected_file = tmpdir.join('file')
+        protected_file.write('1234')
+        tmpdir.join('file.integrity').write('{"algorithm": "HMAC_SERIOUSHASH", "digests": "1234"}')
+        assert not IntegrityCheckedFile.read_integrity_file(str(protected_file), SomeHasher())
```
ThomasWaldmann: assert ... is None?

src/borg/testsuite/file_integrity.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class TestIntegrityCheckedFileParts:
+    @pytest.fixture
+    def integrity_protected_file(self, tmpdir):
+        path = str(tmpdir.join('file'))
+        with IntegrityCheckedFile(path, write=True) as fd:
+            fd.write(b'foo and bar')
+            fd.hash_part('foopart')
+            fd.write(b' other data')
```
ThomasWaldmann: nitpick: b' plus other data'

src/borg/testsuite/file_integrity.py
```diff
@@ -0,0 +1,152 @@
...
+    def test_part_independence(self, integrity_protected_file, partial_read):
+        with open(integrity_protected_file, 'ab') as fd:
+            fd.write(b'some extra stuff that does not belong')
+        with pytest.raises(FileIntegrityError):
+            with IntegrityCheckedFile(integrity_protected_file, write=False) as fd:
+                data1 = b'foo and bar'
+                try:
+                    assert fd.read(len(data1)) == data1
+                    fd.hash_part('foopart')
+                except:
```
ThomasWaldmann: diaper pattern alert

src/borg/crypto/file_integrity.py
enkore: `'algorithm'` fulfils this: Changing it is incompatible with older clients, while additional data can just be dumped into the dict anyway.

src/borg/crypto/file_integrity.py
enkore: Checking is optional. If no file was loaded, no error should be raised. (Cf. TODO in `__init__`)

enkore:
Thanks!

src/borg/crypto/file_integrity.py
enkore: Surely enough (tm) for a filename.

It doesn't matter now, just good practice.

src/borg/crypto/file_integrity.py
ThomasWaldmann: that would mix up things in strange ways. you'ld have to change the algorithm to signal a new version even when keeping the hashing algo the same...

src/borg/crypto/file_integrity.py
ThomasWaldmann: oh, i totally misread that (thought it is file length). yeah, it's enough for a filename. :-D

src/borg/crypto/file_integrity.py
ThomasWaldmann: ah, ok.

src/borg/crypto/file_integrity.py
enkore: > have [_exactly_] one joint and keep it well oiled.

enkore:
Integration will be the next PR, concluding the series for #1688 / #1101.

PR #2503, opened 2017-05-11
key: authenticated mode = unencrypted

enkore:


docs/changes.rst
```diff
@@ -133,10 +133,18 @@ Version 1.1.0b6 (unreleased)
...
+  blank passphrase are now treated as unencrypted repositories for security checks
   (e.g. BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK).
 
+  Previously there would be no prompts nor messages if an unknown repository
+  in one of these modes with a blank passphrase was encountered. This would
+  allow an attacker to swap a repository, if one assumed that the lack of
+  password prompts was due to a set BORG_PASSPHRASE.
+
+  Since the "trick" does not work if BORG_PASSPHRASE is set, this does generally
+  not affect scripts.
```
enkore: https://github.com/borgbackup/borg/pull/2489/files#r115141680

docs/changes.rst
```diff
@@ -133,10 +133,18 @@ Version 1.1.0b6 (unreleased)
 
 Compatibility notes:
 
-- Repositories in a repokey mode with a blank passphrase are now treated
-  as unencrypted repositories for security checks
+- Repositories in a repokey mode (including "authenticated" mode) with a
```
ThomasWaldmann: that phrasing is a bit problematic.

i see what you mean with "a repokey mode" as a more general term than specifically the one "-E repokey" mode, but a normal user will not see / know that.

so better just say "Repositories in 'repokey' (or 'authenticated') mode with an empty passphrase ..."

(also note "empty" is maybe more clear than "blank"? see also text below.)

enkore: I'll list them explicitly (repokey, repokey-blake2, authenticated).

src/borg/crypto/key.py
```diff
@@ -749,6 +749,18 @@ class AuthenticatedKey(ID_BLAKE2b_256, RepoKey):
     ARG_NAME = 'authenticated'
     STORAGE = KeyBlobStorage.REPO
 
+    # It's only authenticated, not encrypted.
```
ThomasWaldmann: ehrm. if we do not protect the key AND we have a repo key, how is that "authentication"?

docs/changes.rst
ThomasWaldmann: ok

src/borg/crypto/key.py
enkore: I changed the variable name to something that is perhaps more clear without further context, and also added some documentation (also on some other members).

src/borg/crypto/key.py
```diff
@@ -129,15 +129,31 @@ def tam_required(repository):
...
+
     # Storage type (no key blob storage / keyfile / repo)
     STORAGE = KeyBlobStorage.NO_STORAGE
 
+    # Seed for the buzhash chunker (borg.algorithms.chunker.Chunker)
+    # type: int
+    chunk_seed = None
+
+    # Whether this *particular instance* is encrypted from a practical point of view,
+    # i.e. when it's using encryption with a blank passphrase, then
```
ThomasWaldmann: say "empty passphrase" for consistency.

src/borg/crypto/key.py
```diff
@@ -749,6 +766,29 @@ class AuthenticatedKey(ID_BLAKE2b_256, RepoKey):
...
+    def extract_nonce(self, payload):
+        # This is called during set-up of the AES ciphers we're not actually using for this
+        # key. Therefore the return value of this method doesn't matter; it's just around
+        # to not have it crash should key identification be run against a very small chunk
+        # by "borg check" when the manifest is lost. (The manifest is always large enough
+        # to have the original method read some garbage from bytes 33-41). (Also, the return
+        # value must be larger than the 41 byte bloat of the original format).
+        if payload[0] != self.TYPE:
+            raise IntegrityError('Manifest: Invalid encryption envelope')
+        return 42
```
ThomasWaldmann: I don't see why the nonce (counter) should be larger than the "41 byte bloat".
The counter is counting AES **blocks** (16 bytes), so there should be no direct relationship to a bytes length.

Or did you just mean 42 * 16 > 41 is also true?

enkore: Eh right, I'm too lazy to calculate the precise value now, but the return value gets the number of AES blocks in the detect() chunk *minus* the 41 byte overhead (not) added, so if the manifest is smaller than that that'd get negative which is not well liked.

That'd probably be ~3 (precisely π).

https://github.com/enkore/borg/blob/a16d81271a3b4915c7fe14a3444df6dd0e88cfad/src/borg/crypto/key.py#L509

PR #2508, opened 2017-05-14
Support common options on the main command

### LINKED ISSUE #1163: main command options?
stuff like --remote-path should be an option of main command, not subcommand, so it can be used like:
```
borg --remote-path=... create
```
Currently, it only works like this:
```
borg create --remote-path=...
```
Thus, it is a bit impractical if one wants to define a abbreviation like:
```
BORG="borg --remote-path=x"
```

enkore:
Fixes #1163

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1870,6 +1870,126 @@ class Archiver:
                     print(warning, file=sys.stderr)
         return args
 
+    class CommonOptions:
+        """
+        Support class to allow specifying common options directly after the top-level command.
+
+        Normally options can only be specified on the parser defining them, which means
+        that generally speaking *all* options go after all sub-commands. This is annoying
+        for common options in scripts, e.g. --remote-path or logging options.
+
+        This class allows adding the same set of options to both the top-level parser
+        and the final sub-command parsers (but not intermediary sub-commands, at least for now).
+
+        It does so by giving every option's target name ("dest") a suffix indicating its level
+        -- no two options in the parser hierarchy can have the same target --
+        then, after parsing the command line, multiple definitions are resolved.
+
+        Defaults are handled by only setting them on the top-level parser and setting
+        a sentinel object in all sub-parsers, which then allows to discern which parser
+        supplied the option.
+        """
+
+        def __init__(self, define_common_options, suffix_precedence):
+            """
+            *define_common_options* should be a callable taking one argument, which
+            will be a argparse.Parser.add_argument-like function.
+
+            *define_common_options* will be called multiple times, and should call
+            the passed function to define common options exactly the same way each time.
+
+            *suffix_precedence* should be a tuple of the suffixes that will be used.
+            It is ordered from lowest precedence to highest precedence:
+            An option specified on the parser belonging to index 0 is overridden if the
+            same option is specified on any parser with a higher index.
+            """
+            self.define_common_options = define_common_options
+            self.suffix_precedence = suffix_precedence
+
+            # Maps suffixes to sets of target names.
+            # E.g. common_options["_subcommand"] = {..., "log_level", ...}
+            self.common_options = dict()
+            # Set of options with the 'append' action.
+            self.append_options = set()
+            # This is the sentinel object that replaces all default values in parsers
+            # below the top-level parser.
+            self.default_sentinel = object()
+
+        def add_common_group(self, parser, suffix, provide_defaults=False):
+            """
+            Add common options to *parser*.
+
+            *provide_defaults* must only be True exactly once in a parser hierarchy,
+            at the top level, and False on all lower levels. The default is chosen
+            accordingly.
+
+            *suffix* indicates the suffix to use internally. It also indicates
+            which precedence the *parser* has for common options. See *suffix_precedence*
+            of __init__.
+            """
+            assert suffix in self.suffix_precedence
+
+            def add_argument(*args, **kwargs):
+                if 'dest' in kwargs:
+                    kwargs.setdefault('action', 'store')
+                    assert kwargs['action'] in ('help', 'store_const', 'store_true', 'store_false', 'store', 'append')
+                    is_append = kwargs['action'] == 'append'
+                    if is_append:
+                        self.append_options.add(kwargs['dest'])
+                        assert kwargs['default'] == [], 'The default is explicitly constructed as an empty list in resolve()'
+                    else:
+                        self.common_options.setdefault(suffix, set()).add(kwargs['dest'])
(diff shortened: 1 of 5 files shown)


src/borg/archiver.py
```diff
@@ -1881,58 +1881,170 @@ def process_epilog(epilog):
...
-                                  help='treat part files like normal files (e.g. to list/extract them)')
-
-        parser = argparse.ArgumentParser(prog=self.prog, description='Borg - Deduplicated Backups')
+            # From lowest precedence to highest precedence:
+            # An option specified on the parser belonging to index 0 is overridden if the
+            # same option is specified on any parser with a higher index.
+            SUFFIX_PRECEDENCE = ('_maincommand', '_midcommand', '_subcommand')
+
+            def __init__(self):
+                from collections import defaultdict
```
enkore: XXX

src/borg/archiver.py
```diff
@@ -1881,58 +1881,170 @@ def process_epilog(epilog):
...
+
+                # Maps suffixes to sets of target names.
+                # E.g. common_options["_subcommand"] = {..., "log_level", ...}
+                self.common_options = defaultdict(defaultdict)
```
enkore: XXX

src/borg/archiver.py
```diff
@@ -1881,58 +1881,170 @@ def process_epilog(epilog):
                 epilog = [line for line in epilog if not line.startswith('.. man')]
             return '\n'.join(epilog)
 
-        common_parser = argparse.ArgumentParser(add_help=False, prog=self.prog)
+        class CommonOptions:
```
enkore: Move out of closure to allow proper unit testing.

src/borg/archiver.py
```diff
@@ -1881,58 +1881,170 @@ def process_epilog(epilog):
...
+            Support class to allow specifying common options directly after the top-level command.
+
+            Normally options can only be specified on the parser defining them, which means
+            that generally speaking *all* options go after all sub-commands. This is annoying
+            for common options in scripts, e.g. --remote-path or logging options.
+
+            This class allows adding the same set of options to both the top-level parser
+            and the final sub-command parsers (but not intermediary sub-commands, at least for now).
+
+            It does so by giving every option's target name ("dest") a suffix indicating it's level
```
ThomasWaldmann: its

enkore:
This implements exactly what I described as difficult in the ticket :)

src/borg/archiver.py
```diff
@@ -1870,6 +1870,126 @@ def preprocess_args(self, args):
...
+                        kwargs['help'] = kwargs['help'] % kwargs
+                        if not is_append:
+                            kwargs['default'] = self.default_sentinel
+
+                common_group.add_argument(*args, **kwargs)
+
+            common_group = parser.add_argument_group('Common options')
+            self.define_common_options(add_argument)
+
+        def resolve(self, args: argparse.Namespace):  # Namespace has "in" but otherwise is not like a dict.
```
ThomasWaldmann: i don't think we used type annotations anywhere else in the code yet, so why do we start with it here?

also, how are the compatibility properties of this? is it ok for >= py 3.4 interpreter?
any editor/ide/tool falling over this badly?

src/borg/archiver.py
```diff
@@ -1881,58 +2001,67 @@ def process_epilog(epilog):
...
+                              help='do not load/update the file metadata cache used to detect unchanged files')
+            add_common_option('--umask', dest='umask', type=lambda s: int(s, 8), default=UMASK_DEFAULT, metavar='M',
+                              help='set umask to M (local and remote, default: %(default)04o)')
+            add_common_option('--remote-path', dest='remote_path', metavar='PATH',
+                              help='use PATH as borg executable on the remote (default: "borg")')
+            add_common_option('--remote-ratelimit', dest='remote_ratelimit', type=int, metavar='rate',
+                              help='set remote network upload rate limit in kiByte/s (default: 0=unlimited)')
+            add_common_option('--consider-part-files', dest='consider_part_files',
+                              action='store_true', default=False,
+                              help='treat part files like normal files (e.g. to list/extract them)')
```
ThomasWaldmann: did you change anything in the block above, besides moving it and changing the function name?

src/borg/archiver.py
```diff
@@ -1999,17 +2128,17 @@ def process_epilog(epilog):
         `repokey` and `keyfile` use AES-CTR-256 for encryption and HMAC-SHA256 for
         authentication in an encrypt-then-MAC (EtM) construction. The chunk ID hash
         is HMAC-SHA256 as well (with a separate key).
-        These modes are compatible with borg 1.0.x.
+        These modes are compatible with Borg 1.0.x.
```
ThomasWaldmann: not sure if we had set a policy like "always write 'Borg' as the short form" and also this might be inconsistent at a thousand other places, so maybe this rather needs such a policy and a global cleanup outside of this PR.

src/borg/testsuite/archiver.py
```diff
@@ -1680,6 +1680,12 @@ def test_log_json(self):
         assert log_message['name'].startswith('borg.')
         assert isinstance(log_message['message'], str)
 
+    def test_common_options(self):
+        self.create_test_files()
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        log = self.cmd('--debug', 'create', self.repository_location + '::test', 'input')
+        assert 'security: read previous_location' in log
```
ThomasWaldmann: ok, so this a quite high-level test, like "does it work in production".
maybe have a more argparse-level test also / instead?

src/borg/archiver.py
enkore: Never heard of something like that. I think some more recent code has some of these already.

Generally speaking there is often no need to have them, but in some cases they can clarify things. This is one of them; Namespace is a rather strange object which mixes Python protocols in strange ways. So explicitly telling that `args` is again this strange object here makes sense. Unlike the do_XXX functions it's not instantly clear that it's the same.

src/borg/archiver.py
enkore: No

enkore:
I designed this around the publicly documented argparse API to avoid that, and it doesn't touch argparse objects indecently apart from `parser.common_options = ...`, which I don't expect to break.

src/borg/archiver.py
enkore: Yeah I can unstage these.

From a writing PoV it would make sense to use `Borg` when we're talking about the project (e.g. docs, releases, community), `borg` when specifically the command is meant.

ThomasWaldmann: yup.

PR #2519, opened 2017-05-16
borg export-tar

### LINKED ISSUE #176: Wishlist: Tar file export / import
It would be neat to be able to directly export an archive from borg to a tarball without needing to extract the tree of files to disk. The main use case for this would be offline and fully self-contained backups for security / archival purposes.

The UX for this would be mostly the same as `borg extract` except that instead of writing files to disk (or plainly to stdout), the stdlib `tarfile` module would be used to write to an archive file (or stdout).

The converse, a `borg import` that works like `borg create` but consumes a tar file (or stream) instead of reading files from disk might be simple enough to add as well, although I don't have a compelling use case for this.

enkore:
Fixes #176


src/borg/archiver.py
```diff
@@ -696,6 +698,195 @@ def peek_and_store_hardlink_masters(item, matched):
...
+
+        filter = None
+        if args.tar_filter == 'auto':
+            # Note that filter remains None if tarfile is '-'.
+            if args.tarfile.endswith('.tar.gz'):
+                filter = 'gzip'
+            elif args.tarfile.endswith('.tar.bz2'):
+                filter = 'bzip2'
+            elif args.tarfile.endswith('.tar.xz'):
+                filter = 'xz'
```
ThomasWaldmann: not sure if we discussed already about this, but doesn't tarfile module offer such stuff internally?

src/borg/archiver.py
```diff
@@ -696,6 +698,195 @@ def peek_and_store_hardlink_masters(item, matched):
...
+        tar.close()
+
+        if tarstream_close:
+            tarstream.close()
+
+        if filter:
+            logger.debug('Done creating tar, waiting for filter to die...')
+            rc = filterproc.wait()
+            logger.debug('filter exited with code %d', rc)
+            self.exit_code = max(self.exit_code, rc)
```
ThomasWaldmann: hmm, this assumes "more is worse" or "more is more interesting". is this always true?

we never have negative values here, right?

src/borg/archiver.py
```diff
@@ -696,6 +698,195 @@ def peek_and_store_hardlink_masters(item, matched):
...
+            """
+            Return a file-like object that reads from the chunks of *item*.
+            """
+            chunk_iterator = archive.pipeline.fetch_many([chunk_id for chunk_id, _, _ in item.chunks])
+            if pi:
+                info = [remove_surrogates(item.path)]
+                return ChunkIteratorFileWrapper(chunk_iterator,
+                                                lambda read_bytes: pi.show(increase=len(read_bytes), info=info))
+            else:
+                return ChunkIteratorFileWrapper(chunk_iterator)
```
ThomasWaldmann: hmm, is some of the above duplicated from somewhere else (like "extract")?

src/borg/archiver.py
```diff
@@ -696,6 +698,195 @@ def peek_and_store_hardlink_masters(item, matched):
...
+            tarinfo.name = item.path
+            tarinfo.mtime = item.mtime / 1e9
+            tarinfo.mode = stat.S_IMODE(item.mode)
+            tarinfo.uid = item.uid
+            tarinfo.gid = item.gid
+            tarinfo.uname = item.user
+            tarinfo.gname = item.group
+            # The linkname in tar has the same dual use the 'source' attribute of Borg items,
+            # i.e. for symlinks it means the destination, while for hardlinks it refers to the
+            # file.
```
ThomasWaldmann: hah, now we know where it all started...

src/borg/archiver.py
```diff
@@ -696,6 +698,195 @@ def peek_and_store_hardlink_masters(item, matched):
...
+
+            Return a tuple (tarinfo, stream), where stream may be a file-like object that represents
+            the file contents, if any, and is None otherwise. When *tarinfo* is None, the *item*
+            cannot be represented as a TarInfo object and should be skipped.
+            """
+
+            # In the PAX (POSIX) format we can support most things that aren't possible with classic tar
+            # formats, including GNU tar, such as:
+            # atime, ctime, possibly Linux capabilities (security.* xattrs) and various additions supported
+            # by GNU tar in POSIX mode.
```
ThomasWaldmann: add a TODO marker?

src/borg/archiver.py
```diff
@@ -2470,6 +2661,77 @@ def process_epilog(epilog):
         subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
                                help='paths to extract; patterns are supported')
 
+        export_tar_epilog = process_epilog("""
+        This command creates a tarball from an archive.
+
+        When giving '-' as the output FILE, Borg will write a tar stream to standard output.
+
+        By default (--tar-filter=auto) Borg will detect whether the FILE should be compressed
+        based on its file extension and pipe the tarball through an appropriate filter
```
ThomasWaldmann: yay, "its"!

src/borg/archiver.py
```diff
@@ -2470,6 +2661,77 @@ def process_epilog(epilog):
...
+        subparser.add_argument('--pattern', action=ArgparsePatternAction,
+                               metavar="PATTERN", help='include/exclude paths matching PATTERN')
+        subparser.add_argument('--patterns-from', action=ArgparsePatternFileAction,
+                               metavar='PATTERNFILE', help='read include/exclude patterns from PATTERNFILE, one per line')
+        subparser.add_argument('--strip-components', dest='strip_components',
+                               type=int, default=0, metavar='NUMBER',
+                               help='Remove the specified number of leading path elements. Pathnames with fewer elements will be silently skipped.')
+        subparser.add_argument('location', metavar='ARCHIVE',
+                               type=location_validator(archive=True),
+                               help='archive to extract')
```
ThomasWaldmann: hmm, "archive to transform" or "... to tarball"?

src/borg/archiver.py
enkore: No, extract doesn't use CIFW and recreate generates it's progress display on the other side of it (via chunk_file).

src/borg/archiver.py
enkore: This is more involved => more suited to a ticket.

src/borg/archiver.py
enkore: Guess where I copied the parser? :D

src/borg/archiver.py
ThomasWaldmann: i meant the TODO mainly as a way to signal this is not done yet.
otherwise the comment could be misread like we actually already use PAX.

src/borg/archiver.py
enkore: I believe we did about create-from-tar which uses/d the same design. The reason stays the same, but is even more important (since compressors are slower): That way the compressor inherently runs in a separate thread and does not hog our main thread.

src/borg/archiver.py
enkore: Perhaps a better way would be to check for !=0 and if so, log error and set error RC (=2) — when the filter terminates abnormally it usually means the result is unusable. Also, it tends to crash borg (BrokenPipeError).

src/borg/archiver.py
ThomasWaldmann: ah, there it was. ok, so maybe add a comment about this for future generations / selfs.

src/borg/archiver.py
ThomasWaldmann: gzip:

Exit  status  is normally 0; if an error occurs, exit status is 1. If a  warning occurs, exit status is 2.

bzip2:

       Return values: 0 for a normal exit, 1 for environmental problems  (file
       not found, invalid flags, I/O errors, &c), 2 to indicate a corrupt com‐
       pressed file, 3 for an  internal  consistency  error  (eg,  bug)  which
       caused bzip2 to panic.

xz:

       0      All is good.
       1      An error occurred.
       2      Something  worth  a  warning  occurred,  but  no  actual  errors
              occurred.

src/borg/archiver.py
```diff
@@ -696,6 +698,219 @@ def peek_and_store_hardlink_masters(item, matched):
...
+        # The tarfile module of Python can provide some compression mechanisms
+        # by itself, using the builtin gzip, bz2 and lzma modules (and "tarmodes"
+        # such as "w:xz").
+        #
+        # Doing so would have three major drawbacks:
+        # For one the compressor runs on the same thread as the program using the
+        # tarfile, stealing valuable CPU time from Borg and thus reducing throughput.
+        # Then this limits the available options - what about lz4? Brotli? zstd?
+        # The third issue is that systems can ship more optimized versions than those
+        # built into Python, e.g. pigz or pxz, which can use more then one thread for
```
ThomasWaldmann: more than

src/borg/archiver.py
```diff
@@ -696,6 +698,219 @@ def peek_and_store_hardlink_masters(item, matched):
...
+                tarinfo.linkname = item.source
+            elif modebits == stat.S_IFBLK:
+                tarinfo.type = tarfile.BLKTYPE
+                tarinfo.devmajor = os.major(item.rdev)
+                tarinfo.devminor = os.minor(item.rdev)
+            elif modebits == stat.S_IFCHR:
+                tarinfo.type = tarfile.CHRTYPE
+                tarinfo.devmajor = os.major(item.rdev)
+                tarinfo.devminor = os.minor(item.rdev)
+            elif modebits == stat.S_IFDIR:
```
ThomasWaldmann: maybe move this up a bit, below regular files, as this is likely the 2nd most frequent fs object, more frequent than symlinks, blk/chr devices.

PR #2527, opened 2017-05-17
Fix remote logging and progress

### LINKED ISSUE #2241: --progress for serve/RPC doesn't work
I think this was broken with the implied logging stuff?

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/42549881-progress-for-serve-rpc-doesn-t-work)

enkore:
Fixes #2241 

Remarkably annoying patch to create.

Compatiblity:

```
     Remote |               |               |               |
            |     1.0.x     |   <1.1.0b5    |   master      |
Client      |               |               |   (>1.1.0b6)  |
------------+---------------+---------------+---------------+
1.0.x       |log w/o prefix | sees log messages w/ prefix   |
            |always progress| progress:never| always        |
------------+               +---------------+---------------+
<1.1.0b5    |progress spreads| broken *1    | always progress + spreads     |
------------+               +---------------+---------------+
>1.1.0b6    |progress stays |  log ok,      | superb *3     |
(master)    |in same line   | no progress *2|               |
------------+---------------+---------------+---------------+

*1 --progress does not work on remote
   --log-json only works with new server
*2 bug is in server
*3
  - progress works with old and new server, but can't
    disable with old server (1.0.x)
  - everything works with --log-json
  - debug topics work
  - progress alway stays in the same line
  - behaves virtually the same as a 1.0.x server towards 1.0.x clients
    (unlike <1.1.0b5)
```

~~So 1.1.0b5 is a bit unpleasant to read with this patch applied, but it won't miss any messages, so I consider that level of breakage unfortunate but ok for a beta. One could work around that via client_data in negotiate but it would be code that only serves this minor compatibility issue with beta releases.~~ (Addressed via client_supports_log_v3)

I elected to fix the existing system, but I think more than ever that it'd be good to redesign the propagation of logging, output and progress configuration within a process (RPC is good now). It would also be good to drop BORG_LOGGING_CONF, since --log-json now exists and apart from use cases covered by --log-json there are imho no convincing reasons to have it, but it makes the code significantly more complex, because it forces us to couples a lot of stuff to the logging system *and* consider the possibility of a logging config changing basically everything about it. (Well that and it's largely useless without *extensive* use of internals that aren't stable).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3815,15 +3815,15 @@ class Archiver:
         """ turn on INFO level logging for args that imply that they will produce output """
         # map of option name to name of logger for that option
         option_logger = {
-                'output_list': 'borg.output.list',
-                'show_version': 'borg.output.show-version',
-                'show_rc': 'borg.output.show-rc',
-                'stats': 'borg.output.stats',
-                'progress': 'borg.output.progress',
-                }
+            'output_list': 'borg.output.list',
+            'show_version': 'borg.output.show-version',
+            'show_rc': 'borg.output.show-rc',
+            'stats': 'borg.output.stats',
+            'progress': 'borg.output.progress',
+        }
         for option, logger_name in option_logger.items():
-            if args.get(option, False):
-                logging.getLogger(logger_name).setLevel('INFO')
+            option_set = args.get(option, False)
+            logging.getLogger(logger_name).setLevel('INFO' if option_set else 'WARN')
 
     def _setup_topic_debugging(self, args):
         """Turn on DEBUG level logging for specified --debug-topics."""
@@ -3839,8 +3839,10 @@ class Archiver:
         # This works around http://bugs.python.org/issue9351
         func = getattr(args, 'func', None) or getattr(args, 'fallback_func')
         # do not use loggers before this!
-        setup_logging(level=args.log_level, is_serve=func == self.do_serve, json=args.log_json)
+        is_serve = func == self.do_serve
+        setup_logging(level=args.log_level, is_serve=is_serve, json=args.log_json)
         self.log_json = args.log_json
+        args.progress |= is_serve
         self._setup_implied_logging(vars(args))
         self._setup_topic_debugging(args)
         if args.show_version:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -313,9 +313,10 @@ class RepositoryServer:  # pragma: no cover
         # clients since 1.1.0b3 use a dict as client_data
         if isinstance(client_data, dict):
             self.client_version = client_data[b'client_version']
-            level = logging.getLevelName(logging.getLogger('').level)
-            setup_logging(is_serve=True, json=True, level=level)
-            logger.debug('Initialized loggin system for new protocol')
+            if client_data.get(b'client_supports_log_v3', False):
+                level = logging.getLevelName(logging.getLogger('').level)
+                setup_logging(is_serve=True, json=True, level=level)
+                logger.debug('Initialized logging system for new (v3) protocol')
         else:
             self.client_version = BORG_VERSION  # seems to be newer than current version (no known old format)
 
@@ -559,7 +560,10 @@ class RemoteRepository:
 
         try:
             try:
-                version = self.call('negotiate', {'client_data': {b'client_version': BORG_VERSION}})
+                version = self.call('negotiate', {'client_data': {
+                    b'client_version': BORG_VERSION,
+                    b'client_supports_log_v3': True,
+                }})
             except ConnectionClosed:
                 raise ConnectionClosedWithHint('Is borg working on the server?') from None
             if version == RPC_PROTOCOL_VERSION:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -952,28 +952,32 @@ def handle_remote_line(line):
             assert isinstance(level, int)
             target_logger = logging.getLogger(msg['name'])
             # We manually check whether the log message should be propagated
+            msg['message'] = 'Remote: ' + msg['message']
             if level >= target_logger.getEffectiveLevel() and logging.getLogger('borg').json:
-                sys.stderr.write(line)
+                sys.stderr.write(json.dumps(msg) + '\n')
             else:
                 target_logger.log(level, '%s', msg['message'])
-        elif msg['type'].startswith('progress_') and not msg.get('finished'):
+        elif msg['type'].startswith('progress_'):
             # Progress messages are a bit more complex.
             # First of all, we check whether progress output is enabled. This is signalled
             # through the effective level of the borg.output.progress logger
             # (also see ProgressIndicatorBase in borg.helpers).
             progress_logger = logging.getLogger('borg.output.progress')
             if progress_logger.getEffectiveLevel() == logging.INFO:
-                # When progress output is enabled, then we check whether the client is in
+                # When progress output is enabled, we check whether the client is in
                 # --log-json mode, as signalled by the "json" attribute on the "borg" logger.
                 if logging.getLogger('borg').json:
-                    # In --log-json mode we directly re-emit the progress line as sent by the server.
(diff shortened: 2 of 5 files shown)


src/borg/remote.py
```diff
@@ -930,7 +953,56 @@ def preload(self, ids):
...
+                    # In --log-json mode we directly re-emit the progress line as sent by the server.
+                    sys.stderr.write(line)
+                else:
+                    # In text log mode we write only the message to stderr and termiante with \r
+                    # (carriage return, i.e. move the write cursor back to the beginning of the line)
+                    # so that the next message, progress or not, overwrites it. This mirrors the behaviour
+                    # of local progress displays.
+                    sys.stderr.write(msg['message'] + '\r')
+    elif line.startswith('$LOG '):
+        # This format is used by Borg since 1.1.0b1.
```
ThomasWaldmann: we don't have compatibility guarantees for betas, so just remove this block and put a remark into changelog?

alternatively, if you want it more smooth / give beta testers more time, keep it and add "TODO: remove after b6" or file a ticket.

src/borg/archiver.py
```diff
@@ -3815,15 +3815,17 @@ def _setup_implied_logging(self, args):
...
+            'show_version': 'borg.output.show-version',
+            'show_rc': 'borg.output.show-rc',
+            'stats': 'borg.output.stats',
+            'progress': 'borg.output.progress',
+        }
         for option, logger_name in option_logger.items():
             if args.get(option, False):
                 logging.getLogger(logger_name).setLevel('INFO')
+            else:
+                logging.getLogger(logger_name).setLevel('WARN')
```
ThomasWaldmann: option_set = args.get(option, False)
logging.getLogger(logger_name).setLevel('INFO' if option_set else 'WARN')

src/borg/remote.py
```diff
@@ -646,12 +654,27 @@ def borg_cmd(self, args, testing):
...
+            #
+            # This is not considered a problem, since this is a debugging feature that
+            # should not be used for regular use.
+            for topic in args.debug_topics:
+                # Check whether either the unqualified ...
+                concerns_repository = topic.startswith('repository')
+                if '.' not in topic:
+                    topic = 'borg.debug.' + topic
+                # ... or the qualified topic starts with repository.
+                concerns_repository |= topic.startswith('repository')
```
ThomasWaldmann: if it did not start with 'repository' in line 671, it can not start with it here either.

src/borg/remote.py
```diff
@@ -930,7 +953,56 @@ def preload(self, ids):
...
+            # (also see ProgressIndicatorBase in borg.helpers).
+            progress_logger = logging.getLogger('borg.output.progress')
+            if progress_logger.getEffectiveLevel() == logging.INFO:
+                # When progress output is enabled, then we check whether the client is in
+                # --log-json mode, as signalled by the "json" attribute on the "borg" logger.
+                if logging.getLogger('borg').json:
+                    # In --log-json mode we directly re-emit the progress line as sent by the server.
+                    sys.stderr.write(line)
+                else:
+                    # In text log mode we write only the message to stderr and termiante with \r
```
ThomasWaldmann: terminate

src/borg/remote.py
enkore: It could be `repository.xy`, but it's still wrong, since we'd use something like `borg.repository[xxx]`. So, rather just `repository in topic`.

src/borg/remote.py
```diff
@@ -312,6 +313,10 @@ def negotiate(self, client_data):
         # clients since 1.1.0b3 use a dict as client_data
         if isinstance(client_data, dict):
             self.client_version = client_data[b'client_version']
+            if client_data.get(b'client_supports_log_v3', False):
```
ThomasWaldmann: maybe also add a TODO about removing this client_supports_log_v3 check in 1.1rc?
if we get into this block, we have a borg 1.1 at the other end.

src/borg/remote.py
```diff
@@ -555,7 +560,10 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
 
         try:
             try:
-                version = self.call('negotiate', {'client_data': {b'client_version': BORG_VERSION}})
+                version = self.call('negotiate', {'client_data': {
+                    b'client_version': BORG_VERSION,
+                    b'client_supports_log_v3': True,
```
ThomasWaldmann: remove this line in 1.1rc?

src/borg/remote.py
```diff
@@ -646,12 +654,23 @@ def borg_cmd(self, args, testing):
...
+            #
+            # Needless to say, if you do --debug-topic=repository.compaction, for example,
+            # with a 1.0.x server it won't work, because the server does not recognize the
+            # option.
+            #
+            # This is not considered a problem, since this is a debugging feature that
+            # should not be used for regular use.
+            for topic in args.debug_topics:
+                if '.' not in topic:
+                    topic = 'borg.debug.' + topic
```
ThomasWaldmann: micro-optimization: move this if-block into the if-block below?

src/borg/remote.py
```diff
@@ -930,7 +949,62 @@ def preload(self, ids):
 
 
 def handle_remote_line(line):
-    if line.startswith('$LOG '):
+    """
+    Handle a remote log line.
+
+    This function is remarkably complex because it handles three different wire formats.
```
ThomasWaldmann: s/three/multiple/

otherwise we'll forget to fix that when we remove one of them at 1.1rc.

src/borg/remote.py
```diff
@@ -930,7 +949,62 @@ def preload(self, ids):
...
+        if msg['type'] not in ('progress_message', 'progress_percent', 'log_message'):
+            logger.warning('Dropped remote log message with unknown type %r: %s', msg['type'], line)
+            return
+
+        if msg['type'] == 'log_message':
+            # Re-emit log messages on the same level as the remote to get correct log suppression and verbosity.
+            level = getattr(logging, msg['levelname'], logging.CRITICAL)
+            assert isinstance(level, int)
+            target_logger = logging.getLogger(msg['name'])
+            # We manually check whether the log message should be propagated
```
ThomasWaldmann: add a dot at the end, please.

src/borg/remote.py
```diff
@@ -930,7 +949,62 @@ def preload(self, ids):
...
+            # First of all, we check whether progress output is enabled. This is signalled
+            # through the effective level of the borg.output.progress logger
+            # (also see ProgressIndicatorBase in borg.helpers).
+            progress_logger = logging.getLogger('borg.output.progress')
+            if progress_logger.getEffectiveLevel() == logging.INFO:
+                # When progress output is enabled, we check whether the client is in
+                # --log-json mode, as signalled by the "json" attribute on the "borg" logger.
+                if logging.getLogger('borg').json:
+                    # In --log-json mode we re-emit the progress JSON line as sent by the server, prefixed
+                    # with "Remote: " when it contains a message.
```
ThomasWaldmann: the code below is clear, but the comment above tells something different than what the code does.
it sounds like it would emit "Remote: { ... }".

src/borg/remote.py
```diff
@@ -941,7 +1015,15 @@ def handle_remote_line(line):
...
-        sys.stderr.write('Remote: ' + line)
+        # Plain 1.0.x and older format - re-emit to stderr (mirroring what the 1.0.x
+        # client did) or as a generic log message.
+        # We don't know what priority the line had.
+        if logging.getLogger('borg').json:
+            logging.getLogger('').warning('Remote: ' + line.strip())
+        else:
+            # In non-JSON mode we circumvent logging to preserve carriage returns (\r)
+            # which are generated by remote progress displays.
+            sys.stderr.write('Remote: ' + line)
```
ThomasWaldmann: this is like it ever was, but it can be confusing if the borg client's local ssh process emits an error message on stderr (that is **not** coming from the remote).

enkore:
Thanks, I clarified these in f0b84042.

PR #2528, opened 2017-05-18
document follow_symlinks requirements, check libc, fixes #2507

### LINKED ISSUE #2507: follow_symlinks unavailable on this platform
Restoring any data with (`borg` installed in anaconda via `pip`) results in a crash:

    [...]
    Local Exception.
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archiver.py", line 2089, in main
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archiver.py", line 2025, in run
        return func(args)
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archiver.py", line 91, in wrapper
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archiver.py", line 101, in wrapper
        return method(self, args, repository=repository, manifest=manifest, key=key, archive=archive, **kwargs)
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archiver.py", line 477, in do_extract
        archive.extract_item(item, stdout=stdout, sparse=sparse)
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archive.py", line 452, in extract_item
        raise Exception('Unknown archive item type %r' % item[b'mode'])
    File "/opt/anaconda/lib/python3.5/contextlib.py", line 77, in __exit__
        self.gen.throw(type, value, traceback)
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archive.py", line 79, in backup_io
        yield
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archive.py", line 442, in extract_item
        self.restore_attrs(path, item, symlink=True)
    File "/opt/anaconda/lib/python3.5/site-packages/borg/archive.py", line 489, in restore_attrs
        os.utime(path, None, ns=(atime, mtime), follow_symlinks=False)
    NotImplementedError: utime: follow_symlinks unavailable on this platform

    Platform: Linux webserver 3.13.0-85-generic #129-Ubuntu SMP Thu Mar 17 20:50:15 UTC 2016 x86_64 x86_64
    Linux: debian jessie/sid
    Borg: 1.0.10  Python: CPython 3.5.1
    PID: 4083  CWD: /root/restore
    sys.argv: ['/opt/anaconda/bin/borg', 'extract', 'goerz@backupserver:borg_backup::blub', 'root']
    SSH_ORIGINAL_COMMAND: None

ThomasWaldmann:
first changeset is the fix.

2nd and 3rd rather cleanups (== always use generic call with follow_symlinks=False).

didn't touch lchmod related code, no lchmod on linux.

also didn't touch lchflags, except removing a unused related variable.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -569,7 +569,7 @@ Utilization of max. archive size: {csize_max:.0%}
         path = os.path.join(dest, item.path)
         # Attempt to remove existing files, ignore errors on failure
         try:
-            st = os.lstat(path)
+            st = os.stat(path, follow_symlinks=False)
             if stat.S_ISDIR(st.st_mode):
                 os.rmdir(path)
             else:
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -674,7 +674,7 @@ Utilization of max. archive size: {csize_max:.0%}
             if fd:
                 os.fchown(fd, uid, gid)
             else:
-                os.lchown(path, uid, gid)
+                os.chown(path, uid, gid, follow_symlinks=False)
         except OSError:
             pass
         if fd:
--- a/src/borg/testsuite/__init__.py
+++ b/src/borg/testsuite/__init__.py
@@ -30,13 +30,11 @@ except ImportError:
     raises = None
 
 has_lchflags = hasattr(os, 'lchflags') or sys.platform.startswith('linux')
-no_lchlfags_because = '' if has_lchflags else '(not supported on this platform)'
 try:
     with tempfile.NamedTemporaryFile() as file:
         platform.set_flags(file.name, stat.UF_NODUMP)
 except OSError:
     has_lchflags = False
-    no_lchlfags_because = '(the file system at %s does not support flags)' % tempfile.gettempdir()
 
 try:
     import llfuse
--- a/src/borg/testsuite/__init__.py
+++ b/src/borg/testsuite/__init__.py
@@ -67,7 +67,7 @@ def are_symlinks_supported():
     with unopened_tempfile() as filepath:
         try:
             os.symlink('somewhere', filepath)
-            if os.lstat(filepath) and os.readlink(filepath) == 'somewhere':
+            if os.stat(filepath, follow_symlinks=False) and os.readlink(filepath) == 'somewhere':
                 return True
         except OSError:
             pass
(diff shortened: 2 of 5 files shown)


src/borg/archive.py
```diff
@@ -569,7 +569,7 @@ def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sp
         path = os.path.join(dest, item.path)
         # Attempt to remove existing files, ignore errors on failure
         try:
-            st = os.lstat(path)
+            st = os.stat(path, follow_symlinks=False)
```
enkore: What's the point of changing these? According to the docs they are equivalent.

src/borg/testsuite/__init__.py
```diff
@@ -30,13 +30,13 @@
...
 has_lchflags = hasattr(os, 'lchflags') or sys.platform.startswith('linux')
-no_lchlfags_because = '' if has_lchflags else '(not supported on this platform)'
+no_lchflags_because = '' if has_lchflags else '(not supported on this platform)'
 try:
     with tempfile.NamedTemporaryFile() as file:
         platform.set_flags(file.name, stat.UF_NODUMP)
 except OSError:
     has_lchflags = False
-    no_lchlfags_because = '(the file system at %s does not support flags)' % tempfile.gettempdir()
+    no_lchflags_because = '(the file system at %s does not support flags)' % tempfile.gettempdir()
```
enkore: looks unused -> remove

enkore:
There is also borg.xattr which needs some l-functions such as llistxattr (glibc >2.3?).

src/borg/archive.py
ThomasWaldmann: 2nd and 3rd rather cleanups (== always use generic call with follow_symlinks=False).

consistency. couldn't get all switched though due to lchmod on linux.

ThomasWaldmann:
BTW, there is a alternative way to deal with this, see there:

https://github.com/borgbackup/borg/blob/1.1.0b5/src/borg/archive.py#L681

This is more compatible for setting operations, we can choose not to set symlink attributes if there is no support for that, but still work correctly for other fs object types.

For getting operations and symlinks we would have a problem still, guess it is not accepable to not get or get attributes from symlink target.

src/borg/archive.py
enkore: I can get behind that.

lchmod on Linux is easy, though; it doesn't make sense, so it doesn't exist (Linux ignores permissions on symlinks, like most Unices do).

src/borg/testsuite/__init__.py
ThomasWaldmann: ok

ThomasWaldmann:
OK, removed the unused stuff, partially squashed, rebased.
Not sure what to do with the l* calls from xattr.

enkore:
The l\* xattr functions seem to be very old (glibc 2.3) so probably not an issue in practice.

PR #2530, opened 2017-05-18
Repository compaction docs

enkore:
low-risk part of an unmerged branch (1.3.17)

Previously the auto-upgrade (remember, backwards-compatible to 1.0.4 and later) would seek through all the segments that were sparse, which is not exactly the quickest thing to do. This way the compaction algorithm behaves more like the old one with 1.0.x repos — for the segments created with 1.0.x.

Also increase DSPD a bit since more clobber is created than initially expected.


enkore:
> I don't get what you are doing here - after reading the comments / PR text.

I changed the PR topic to a docs update and removed the change you are probably referring to in accordance with the updated docs ("This process may take some time if a repository ..."). The paragraphs explaining versions and what happens when version=1 is read might clean up the confusion (although I can't know for sure what you didn't understand, since you didn't say).

ThomasWaldmann:
re-read it, this time everything was clear enough. nice pic.

PR #2544, opened 2017-05-21
docs/data structures: add chunk layout diagram

enkore:


enkore:
We do treat the two-byte zlib header as a type word upon decompression. For compression the difference would be that technically `ZLIB (Compressor)` does not explicitly prepend it - from the outside there is no difference.

PR #2553, opened 2017-05-22
docs: rework Attic FAQ

enkore:


docs/conf.py
```diff
@@ -210,6 +204,15 @@
...
 latex_logo = '_static/logo.pdf'
 
+latex_elements = {
+    'papersize': 'a4paper',
+    'pointsize': '10pt',
+    'preamble':
+r"""
+""",
+
+}
```
ThomasWaldmann: above is strangely formatted.

docs/conf.py
```diff
@@ -224,7 +227,12 @@
 #latex_preamble = ''
 
 # Documents to append as an appendix to all manuals.
-#latex_appendices = []
+latex_appendices = [
+    'support',
+    'resources',
+    'changes',
+    'authors',
+]
```
ThomasWaldmann: hmm? so what does this do precisely, which output formats are affected?

enkore: The "all manuals" part in the docstring means all *latex* manuals, since it's possible to generate multiple distinct manuals from the same Sphinx project. (E.g. one manual for the Python stdlib, one for the CPython API, one for the language spec. Or a user's guide and a developer's guide.) We don't do that, though.

enkore: 2017-05-23-100033_3840x1440_scrot

ThomasWaldmann: so, does this mean the appendices do not show up on the website (html output)?

man page?

enkore: Like I said, `latex_` options only affect the latex builder.

ThomasWaldmann: just seen: you only removed them from book.rst, so this only affects output generated from that.

PR #2552, opened 2017-05-22
docs: borg_domain for custom priority index entries

enkore:
This does NOT fix #1485, but it is infrastructure that makes it fixable.

Before:

2017-05-22-194152_3840x1440_scrot

After:

2017-05-22-194159_3840x1440_scrot

PR #2554, opened 2017-05-22
Revert "docs: borg_domain for custom priority index entries"

enkore:
Reverts borgbackup/borg#2552 because it does not work on readthedocs.


ThomasWaldmann:
that sphinx extension could be a bit less borg-y in the source code.
i could imagine that other projects want to use that also.

PR #2555, opened 2017-05-23
WIP ipv6 address support

### LINKED ISSUE #2256: Please enable IPv6 addresses for remote repositories

ThomasWaldmann:
also: Location: more informative exception when parsing fails

continuation of #2264, fixes #2256.

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -769,7 +769,7 @@ def bin_to_hex(binary):
 class Location:
     """Object representing a repository / archive location
     """
-    proto = user = host = port = path = archive = None
+    proto = user = _host = port = path = archive = None
 
     # user must not contain "@", ":" or "/".
     # Quoting adduser error message:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
ipv6 localhost keyfilename is currently like:
```
Enter passphrase for key /home/tw/.config/borg/keys/::1__w_borg_repo: 
```
Are we ok with `:` in filenames?

ThomasWaldmann:
0d406e7bafa8064edba92841235bfee79e191df4 fixes the traceback observed in #2264.

PR #2565, opened 2017-05-25
Implement storage quotas

### LINKED ISSUE #2517: Optional space limit
Server admins using borg should be able to specify a space limit in `borg serve` to prevent users from uploading too much data and thus abusing the service (and e.g. making it unavailable for other users), potentially also somehow DOSing it.

Maybe e.g. `borg server --limit-space 1TB`.

enkore:
Fixes #2517 

I gave #2517 some thought; implementing it in Borg has some advantages:

- Error messages are more helpful
- Our pre-commit checks do not capture FS quotas
- As implemented, the client can always back out of a 100.0 % used quota, which is not possible with (hard) FS quotas

Disadvantages:

- More difficult to deploy compared to FS quotas

  Perhaps a complementary option to --restrict-to-path that matches the path exactly would simplify handling of repo hosting in general?

- The storage quota and the actual bound on disk space use have a somewhat loose relationship.

  (In case anyone feels reminded of file systems and filling them up completely... it's the same structures, the same math — an exercise for the reader: design a transactional data store that has a fixed upper bound X, independent of contents, such that using SPACE(X) the store's contents can still be cleaned up :)

~~TODO: Add tests for quota tracking.~~

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
 (The actual algorithm is more complex to avoid various consistency issues, refer to
 the ``borg.repository`` module for more comments and documentation on these issues.)
 
+.. _internals_storage_quota:
+
+Storage quotas
+~~~~~~~~~~~~~~
+
+Quotas are implemented at the Repository level. The active quota of a repository
+is determined by the ``storage_quota`` `config` entry or a run-time override (via :ref:`borg_serve`).
+The currently used quota is stored in the hints file. Operations (PUT and DELETE) during
+a transaction modify the currently used quota:
+
+- A PUT adds the size of the *log entry* to the quota,
+  i.e. the length of the data plus the 41 byte header.
+- A DELETE subtracts the size of the deleted log entry from the quota,
+  which includes the header.
+
+Thus, PUT and DELETE are symmetric and cancel each other out precisely.
+
+The quota does not track on-disk size overheads (due to conditional compaction
+or append-only mode). In normal operation the inclusion of the log entry headers
+in the quota act as a faithful proxy for index and hints overheads.
+
+By tracking effective content size, the client can *always* recover from a full quota
+by deleting archives. This would not be possible if the quota tracked on-disk size,
+since journaling DELETEs requires extra disk space before space is freed.
+Tracking effective size on the other hand accounts DELETEs immediately as freeing quota.
+
+.. rubric:: Enforcing the quota
+
+The storage quota is meant as a robust mechanism for service providers, therefore
+:ref:`borg_serve` has to enforce it without loopholes (e.g. modified clients).
+
+The quota is enforcible only if *all* :ref:`borg_serve` versions
+accessible to clients support quotas (see next section). Further, quota is
+per repository. Therefore, ensure clients can only access a defined set of repositories
+with their quotas set, using ``--restrict-to-path``.
+
+If the client exceeds the storage quota the ``StorageQuotaExceeded`` exception is
+raised. Normally a client could ignore such an exception and just send a ``commit()``
+command anyway, circumventing the quota. However, when ``StorageQuotaExceeded`` is raised,
+it is stored in the ``transaction_doomed`` attribute of the repository.
+If the transaction is doomed, then commit will re-raise this exception, aborting the commit.
+
+The transaction_doomed indicator is reset on a rollback (which erases the quota-exceeding
+state).
+
+.. rubric:: Compatibility with older servers and enabling quota after-the-fact
+
+If no quota data is stored in the hints file, Borg assumes zero quota is used.
+Thus, if a repository with an enabled quota is written to with an older version
+that does not understand quotas, then the quota usage will be erased.
+
+A similar situation arises when upgrading from a Borg release that did not have quotas.
+Borg will start tracking quota use from the time of the upgrade, starting at zero.
+
+If the quota shall be enforced accurately in these cases, either
+
+- delete the ``index.N`` and ``hints.N`` files, forcing Borg to rebuild both,
+  re-acquiring quota data in the process, or
+- edit the msgpacked ``hints.N`` file (not recommended and thus not
+  documented further).
+
 .. _manifest:
 
 The manifest
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4000,7 +4000,7 @@ def main():  # pragma: no cover
             tb = "%s\n%s" % (traceback.format_exc(), sysinfo())
             exit_code = e.exit_code
         except RemoteRepository.RPCError as e:
-            important = e.exception_class not in ('LockTimeout', )
+            important = e.exception_class not in ('LockTimeout', ) and e.traceback
             msgid = e.exception_class
             tb_log_level = logging.ERROR if important else logging.DEBUG
             if important:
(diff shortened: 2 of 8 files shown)


docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+A similar situation arises when upgrading from a Borg release that did not have quotas.
+Borg will start tracking quota use from the time of the upgrade, starting at zero.
+
+If the quota shall be enforced accurately in these cases, either
+
+- delete the ``index.N`` and ``hints.N`` files, forcing Borg to rebuild both,
+  re-acquiring quota data in the process, or
+- edit the msgpacked ``hints.N`` file (not recommended and thus not
+  documented further).
+
```
enkore: The trade-offs documented here drastically simplify the implementation.

docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+Storage quotas
+~~~~~~~~~~~~~~
+
+Quotas are implemented at the Repository level. The active quota of a repository
+is determined by the ``storage_quota`` `config` entry or a run-time override (via :ref:`borg_serve`).
+The currently used quota is stored in the hints file. Operations (PUT and DELETE) during
+a transaction modify the currently used quota:
+
+- A PUT adds the size of the *log entry* to the quota,
+  i.e. the length of the data plus the 41 byte header.
```
ThomasWaldmann: maybe do not mention the specific size, would get out of sync if we ever change that.
so just "length of the put command header"?

docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+- A DELETE substracts the size of the deleted log entry to the quota,
```
ThomasWaldmann: **subtracts** ... **from** the quota.

docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+  which includes the header.
+
+Thus, PUT and DELETE are symmetric and cancel each other out precisely.
```
ThomasWaldmann: not sure: "cancel out each other"?

docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+
+.. rubric:: Enforcing the quota
+
+The storage quota is meant as a robust mechanism for service providers, therefore
+:ref:`borg_serve` has to enforce it without loopholes (e.g. modified clients).
+
+The quota is enforcible only if *all* :ref:`borg_serve` versions
+accessible to clients support quotas (see next section). Further, quota is
+per repository. Therefore, ensure clients can only access a defined set of repositories
+with their quotas set, using ``--restrict-to-path``.
```
ThomasWaldmann: hmm, iirc ``--restrict-to-path`` was only meant to restrict the path prefix ("base dir"), but not meaning that there is only 1 repo there.

so, would a provider pre-create some repos below the base dir and revoke write permissions to the base dir afterwards for the client so the client can not create any further repos (directories)?

src/borg/archiver.py
```diff
@@ -142,6 +142,13 @@ def wrapper(self, args, repository, key, manifest, **kwargs):
     return wrapper
 
 
+def parse_storage_quota(storage_quota):
+    parsed = parse_file_size(storage_quota)
+    if parsed < parse_file_size('10M'):
+        raise argparse.ArgumentTypeError('quota is too small (%s), must exceed 10M' % storage_quota)
```
ThomasWaldmann: "exceed" (from: excessive) sounds strange here. rather "must be at least 10M"?

also: refactor to not repeat '10M'?

src/borg/repository.py
```diff
@@ -107,10 +107,14 @@ def __init__(self, id, repo):
     class InsufficientFreeSpaceError(Error):
         """Insufficient free space to complete transaction (required: {}, available: {})."""
 
-    def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=True, append_only=False):
+    class StorageQuotaExceeded(Error):
+        """The storage quota ({}) has been exceeded ({}). Try deleting some archives."""
```
ThomasWaldmann: add this to `handle_error` in `remote.py`?

docs/internals/data-structures.rst
enkore: "if ever"

docs/internals/data-structures.rst
enkore: Both forms are used

docs/internals/data-structures.rst
enkore: See PR comment

>  More difficult to deploy compared to FS quotas
>
>    Perhaps a complementary option to --restrict-to-path that matches the path exactly would simplify handling of repo hosting in general?

src/borg/repository.py
enkore: The generic error handling in protocol v2 means that we don't have to do that unless the type of the exception is relevant, i.e. some code expects that exception. For user-facing error messages there is no need.

enkore: Actually that wasn't quite right, since the `.traceback` attributes of Error was not propagated this did not quite work as intended. Fixed below.

src/borg/repository.py
```diff
@@ -209,6 +216,10 @@ def create(self, path):
         config.set('repository', 'segments_per_dir', str(DEFAULT_SEGMENTS_PER_DIR))
         config.set('repository', 'max_segment_size', str(DEFAULT_MAX_SEGMENT_SIZE))
         config.set('repository', 'append_only', str(int(self.append_only)))
+        if self.storage_quota:
+            config.set('repository', 'storage_quota', str(self.storage_quota))
+        else:
+            config.set('repository', 'storage_quota', '0')
```
ThomasWaldmann: set(..., str(self.storage_quota if self.storage_quota else 0))

src/borg/repository.py
```diff
@@ -438,7 +461,8 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
     def write_index(self):
         hints = {b'version': 2,
                  b'segments': self.segments,
-                 b'compact': self.compact}
+                 b'compact': self.compact,
+                 b'storage_quota_use': self.storage_quota_use, }
```
ThomasWaldmann: nitpick: that way of using a trailing comma does not minimize the diff if a new element is added afterwards.

should be rather:
```
{a: 1,
 b: 2,
}
```

docs/internals/data-structures.rst
```diff
@@ -185,6 +185,67 @@ commit logic) showing the principal operation of compaction:
...
+If the transaction is doomed, then commit will re-raise this exception, aborting the commit.
+
+The transaction_doomed indicator is reset on a rollback (which erases the quota-exceeding
+state).
+
+.. rubric:: Compatibility with older servers and enabling quota after-the-fact
+
+If no quota data is stored in the hints file, Borg assumes zero quota is used.
+Thus, if a repository with an enabled quota is written to with an older version
+that does not understand quotas, then the quota usage will be erased.
```
rugk: Does not this allow users to circumvent this feature? Say a server admins wants to enforce a 50 GB quota and puts this limit there. Then the user can just access the repo with an old borg client and voila… the quota is gone.
That's how I understand that sentence.

ThomasWaldmann: see section headline - it is referring to servers here, not clients.

but maybe it makes sense to actually say that in that sentence, just to avoid misunderstandings.

enkore: https://github.com/borgbackup/borg/pull/2588

PR #2566, opened 2017-05-25
Restrict key file names

### LINKED ISSUE #2560: Restrict file names generated by Location.to_keyfile_name
The generated file can currently include special characters like those mentioned in #2290, which should be stripped for compatibility reasons.

From #2555

enkore:
Fixes #2560

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -917,6 +917,11 @@ class Location:
         name = re.sub('[^\w]', '_', self.path).strip('_')
         if self.proto != 'file':
             name = re.sub('[^\w]', '_', self.host) + '__' + name
+        if len(name) > 100:
+            # Limit file names to some reasonable length. Most file systems
+            # limit them to 255 [unit of choice]; due to variations in unicode
+            # handling we truncate to 100 *characters*.
+            name = name[:100]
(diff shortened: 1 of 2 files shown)


src/borg/helpers.py
```diff
@@ -916,7 +916,12 @@ def __str__(self):
     def to_key_filename(self):
         name = re.sub('[^\w]', '_', self.path).strip('_')
         if self.proto != 'file':
-            name = self.host + '__' + name
+            name = re.sub('[^\w]', '_', self.host) + '__' + name
+        if len(name) > 100:
+            # Limit file names to some reasonable length. Most file systems
+            # limit them to 255 [unit of choice]; due to variations in unicode
+            # handling we truncate to 100 *characters*.
+            name = name[:100]
```
ThomasWaldmann: Well, if done like that, it will truncate the path part and if you're unlucky, 2 different repos will map to same key filename (if they reside on same server, same base path, but different final path elements).

enkore: No, there is some code in key.py that ensures no existing key files are overwritten.

ThomasWaldmann: ah, right. ok, then no problem. :)

PR #2569, opened 2017-05-25
setup.py clean to remove compiled files

enkore:


setup.py
```diff
@@ -567,11 +568,30 @@ def is_positional_group(group):
...
+class Clean(clean):
+    def run(self):
+        super().run()
+        for source in cython_sources:
+            genc = source.replace('.pyx', '.c')
+            try:
+                os.unlink(genc)
+                print('rm', genc)
+            except FileNotFoundError:
+                pass
```
ThomasWaldmann: above few lines maybe could be a toplevel utility function so we can call it from everywhere we need it in setup.py.

print is just for debugging?

enkore: The command we're extending here prints what it removes as well.

PR #2572, opened 2017-05-26
Improve cache sync speed

### LINKED ISSUE #1729: Improve cache sync speed

### LINKED ISSUE #2515: Massive disk usage during cache rebuilding
After .cache/borg got corrupted, I removed it, and touched `.cache/borg/.../chunks.archive.d`.
The next backup took a very long time, and downloaded roughly 20G of data into `/var/tmp`.
(I set TMPDIR=/var/tmp.)

Command line was `borg create -s -C zlib -e ... user@remote:foo.borg::foo-full-... /`

What remained are ~700MB `chunks` and ~200MB `files`, which is reasonable.

The repo is quite big, but the temporary data usage should not be 20x of the usual need.

Borg 1.0.10 running on both ends, on Linux x86_64 (source: Void, target: Arch).

enkore:
This branch is approx. ~three times faster than `master`.

- 0be4561 is over one year old and replaces the hacky cache. This change saves about 30 % due to buggy(?) posix_fadvise on Linux tmpfs. Before my CRC32 changes the impact was slightly higher due to slower CRC32. I changed this from LRUCache to a random eviction scheme, because our LRUCache has square runtime, which was highly problematic in this instance.
- c309584 fixes a nested-loops bug (for archive in archive: for archive in archive:). At 1100-1200 archives this is a minor change, but due to strictly n² behaviour it escalates rather quickly, and independent from archive metadata size.
- 17b21fc initializes the master chunk index like we already do for check
- 1f3d59d supplies a replacement for the core loop of the cache sync using msgpack-c. Another viable approach would be to directly use the msgpack unpack state machine, which could avoid a dependency on msgpack-c. The implementation as shown is from earlier this year.
- cda12c4 replaces 1f3d59d with a state machine on top of the unpacking engine
- f33224a allows RepositoryCache to cache decrypted and verified objects, thus saving repeated MAC checking and decompression (slow for LZMA/zlib). The cache is compressed using LZ4 to preserve its effective size.

Fixes #2515 and fixes #1729

~~Another possible optimization is to cache decrypted repository objects. Cache instrumentation suggests that this would yield another 25 %.~~

~~TODOs for integration would be; resolve 1f3d59d, add tests, add fuzzing harness.~~


enkore:
2017-05-26-185044_3840x1440_scrot

^- That's where the idea of directly using the unpacker state (actually stack) machine comes from. msgpack_unpacker_next has to built a complex structure for each Item which we are generally not interested in. The SM interface is basically like a SAX API, so fairly minimal state tracking is sufficient to find the `chunks` and consume them.

Note that the current version as shown here already processes (=cache_sync_feed) metadata at close to 500 MB/s.

Also nicely confirms the angle of decrypted caching.

src/borg/cache.py
```diff
@@ -564,6 +564,9 @@ def create_master_idx(chunk_idx):
             # deallocates old hashindex, creates empty hashindex:
             chunk_idx.clear()
             cleanup_outdated(cached_ids - archive_ids)
+            # Explicitly set the initial hash table capacity to avoid performance issues
+            # due to hash table "resonance".
+            master_index_capacity = int(len(self.repository) / ChunkIndex.MAX_LOAD_FACTOR)
```
ThomasWaldmann:
```
        # Explicitly set the initial hash table capacity to avoid performance issues
        # due to hash table "resonance".
        # Since reconstruction of archive items can add some new chunks, add 10 % headroom
        capacity = int(len(self.repository) / ChunkIndex.MAX_LOAD_FACTOR * 1.1)
```
(from init_chunks())

Guess we need some headroom factor here also.

enkore: Shouldn't the /MAX_LF already account for this? The sync won't add any chunks (unlike check).

enkore: Also; it would probably be nicer if the capacity given would actually be the usable capacity, i.e. if hashindex_init or HashIndex.__init__ already added the necessary load factor margin.

enkore:
Working state-machine version; about 25 % faster in processing the data (@ 630 MB/s), but overall execution time is only slightly reduced, since execution time was already dominated by key.decrypt.

Thus, the next logical step would be to cache decrypted objects instead of encrypted objects. Whether caching completely decompressed or still compressed objects (but only doing ID verification once) depends on the compressor and some other factors, so I don't think I'll do that here. (It might just be that LZ4'ing stuff to fit more into the cache helps, even if you have to invoke LZ4 when the data was not originally compressed with it).

enkore:
Python's take on things

2017-05-27-124240_3840x1440_scrot

msgpack-c

2017-05-27-125330_3840x1440_scrot

FSM (on top of the core msgpack stack machine; I should mention that I didn't remove all unnecessary data movements yet)

2017-05-27-125355_3840x1440_scrot

src/borg/testsuite/cache.py
```diff
@@ -0,0 +1,139 @@
...
+    @pytest.mark.parametrize('data,error', (
+        # Incorrect tuple length
+        ({'chunks': [(bytes(32), 2, 3, 4)]}, 'Invalid chunk list entry length'),
+        ({'chunks': [(bytes(32), 2)]}, 'Invalid chunk list entry length'),
+        # Incorrect types
+        ({'chunks': [(1, 2, 3)]}, 'Unexpected object: integer'),
+        ({'chunks': [(1, bytes(32), 2)]}, 'Unexpected object: integer'),
+        ({'chunks': [(bytes(32), 1.0, 2)]}, 'Unexpected object: double'),
+     ))
+    def test_corrupted_ancillary(self, index, sync, data, error):
```
enkore: src/borg/testsuite/cache.py:134:6: E121 continuation line under-indented for hanging indent

It's indented by four, just like the other parametrizations above. WTH is flake8's problem?

enkore:
For reference purposes, `master` branch

2017-05-27-223721_3840x1440_scrot_1

src/borg/cache.py
ThomasWaldmann: hmm, yeah, as the repo contains all the chunks referenced by archives, the computation is ok.

good idea about giving usable capacity, separate PR just for that.

src/borg/remote.py
```diff
@@ -1045,45 +1047,105 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        target_size = int(0.9 * self.size_limit)
+        while self.size > target_size and self.cache:
+            key = self.cache.pop()
+            file = self.key_filename(key)
+            self.size -= os.stat(file).st_size
+            os.unlink(file)
+            self.evictions += 1
+
+    def add_entry(self, key, data):
+        file = self.key_filename(key)
```
enkore: file is not a builtin anymore, it's just the Github highlighter. It doesn't know that this is Python 3.

textshell:
impressive speedup. But also impressively more code :).

Is the measurement with an enabled but empty archive index cache? Or with disabled archive index cache? Or is it valid for both?

enkore:
A lot of the code isn't new; from the C side of things only unpack.h and cache_sync.c.

I make these measurements with a deleted cache, i.e. a full sync over all archives. The difference between that and a disabled chunks.archive.d is minor in terms of CPU (since hashindex_write doesn't do anything CPU intensive yet, see 1101.integration.cache). The wall-clock difference is also not that big since the writes are not synchronous and also go to a fast SSD.

src/borg/remote.py
ThomasWaldmann: guess we could still just use fname to avoid these kinds of problems.

as it is a "file name", not a "file", "fname" is even more precise.

ThomasWaldmann:
* c309584 -> approved
* 17b21fc -> approved (backport 1.0-maint?)

ThomasWaldmann:
I'll have another look at this later tonight...

enkore:
```
metric     master   this branch  change
======     ======== ============ ======
CPU        1815 s   510 s        -72 %
wall clock 1900 s   457 s        -76 %
network    18.4 GB  7 GB         -62 %
TMP use    3.3 GB   2.15 GB      -35 %

Archives     1372
uniq chunks  951949
tot chunks   499681653
metadata     123.83 GB
             total 770937 chunks
```

Couples notes

- Ratios will differ by repository structure. This is a repository containing frequent system and also data backups.
- master and this branch are up-to-date (with xxh64 and such)
- The network delta confirms that the THRESHOLD in the old cache prevented caching of about ~half the metadata
- Less network transfer even though the cache is smaller
- CPU efficiency is hugely improved, especially evident when comparing more detailed performance data.

ThomasWaldmann:
- 9f8b967a6f45bb7dbbf1f37cf231ea82f149a0f6 approved
- 167875b753290f468751a973f0be719972c265a9 approved

src/borg/remote.py
```diff
@@ -1141,31 +1159,56 @@ def get_many(self, keys, cache=True):
...
 
-def cache_if_remote(repository):
+def cache_if_remote(repository, *, decrypted_cache=False, pack=None, unpack=None, transform=None):
+    """
+    Return a Repository(No)Cache for *repository*.
+
+    If *decrypted_cache* is a key object, then get and get_many will return a tuple
+    (csize, plaintext) instead of the actual data in the repository. The cache will
+    store decrypted data, which increases CPU efficiency (by avoiding repeatedly decrypting
+    and more importantly MAC and ID checking cached objects).
```
ThomasWaldmann: due to this, we lose a validity check. so if there is a bitflip in the cache, we likely won't detect it.

src/borg/remote.py
```diff
@@ -1141,31 +1159,56 @@ def get_many(self, keys, cache=True):
...
+    Internally, objects are compressed with LZ4.
+    """
+    if decrypted_cache and (pack or unpack or transform):
+        raise ValueError('decrypted_cache and pack/unpack/transform are incompatible')
+    elif decrypted_cache:
+        key = decrypted_cache
+        cache_struct = struct.Struct('=I')
+        compressor = LZ4()
+
+        def pack(data):
```
ThomasWaldmann: use some name that indicates the tuple components? just "data" is a bit misleading.

src/borg/remote.py
enkore: I thought about applying XXH64 to these. Note that LZ4 compression is performed by compress.pyx on a block level, not frame level, thus does not include a checksum.

src/borg/remote.py
enkore: These are callables passed to Repository(No)Cache which defines this generic interface. I don't like changing parameter names in what is similar to a derived class.

Perhaps reorder them transform-pack-unpack, so that the relation is more clear. And a comment or so.

src/borg/cache.py
```diff
@@ -612,15 +613,12 @@ def create_master_idx(chunk_idx):
...
-                                                  msgid='cache.sync')
+                pi = ProgressIndicatorPercent(total=len(archive_ids), step=0.1,
+                                              msg='%3.0f%% Syncing chunks cache. Processing archive %s',
+                                              msgid='cache.sync')
                 archive_ids_to_names = get_archive_ids_to_names(archive_ids)
-                for archive_id in archive_ids:
-                    archive_name = archive_ids_to_names.pop(archive_id)
-                    if self.progress:
-                        pi.show(info=[remove_surrogates(archive_name)])
+                for archive_id, archive_name in archive_ids_to_names.items():
```
ThomasWaldmann: guess this processes archives in random order while it was ordered-by-time before?

could look a bit unpretty and also impact performance because when processed in-order, there is more similarity between archives.

src/borg/cache.py
```diff
@@ -612,15 +613,12 @@ def create_master_idx(chunk_idx):
             master_index_capacity = int(len(self.repository) / ChunkIndex.MAX_LOAD_FACTOR)
             if archive_ids:
                 chunk_idx = None
-                if self.progress:
```
ThomasWaldmann: why did you remove the self.progress check?

src/borg/cache.py
enkore: archive_ids is a set, so hash order. This is not changed in Py3.6, but shuffled in <Py3.6. No relevant difference.

src/borg/cache.py
enkore: ProgressHandlerThingy handles that itself. Unless there is some advantage (e.g. like in "borg extract") in not providing progress information, there is no need to litter the code with "if progress: pi..." or "if pi: pi ..." anymore :)

src/borg/remote.py
ThomasWaldmann: sounds good. or use detached integrity checked files inside the cache?

enkore: = b544af2. I removed the automatic recovery part in 4faaa7d though, since these would _usually_ ($TMP could be on a disk, but its rare nowadays) indicate bad RAM. I leave the changesets in though, so if we decide later to have that we can.

enkore:
> do you think you could move them to (an)other PR(s), so we get them off the table first?

Like you say, the large parts are mostly independent, but they're all joined together in cache.py. I don't want to pick that apart and have to re-test from 0 there. Test coverage of that area is poorer than coverage would led one to believe.

src/borg/cache.py
enkore: Actually it's the other way around. Previously it was random, and for <Py3.6 it still is. For Py3.6+ on the other hand, dict() preserves insertion order, and the loop get_archive_ids_to_names operates on the Archives, which are not explicitly sorted. However, when writing the Manifest, StableDict is used, so the entries in the msgpackd-Manifest are sorted by name. This is preserved when loading it. So for Py3.6+ this now processes in sorted-by-name order, which is probably best: `date-backupset` => clustered by date, `backupset-date` (--prefix compliance) => clustered by backupset.

src/borg/remote.py
```diff
@@ -1194,21 +1206,30 @@ def cache_if_remote(repository, *, decrypted_cache=False, pack=None, unpack=None
...
+            return cache_struct.pack(csize, xxh64(compressed)) + compressed
 
         def unpack(data):
-            return cache_struct.unpack(data[:cache_struct.size])[0], compressor.decompress(data[cache_struct.size:])
+            data = memoryview(data)
+            csize, checksum = cache_struct.unpack(data[:cache_struct.size])
+            compressed = data[cache_struct.size:]
+            if checksum != xxh64(compressed):
+                logger.warning('Repository metadata cache: detected corrupted data in cache!')
+                raise RepositoryCache.InvalidateCacheEntry
```
ThomasWaldmann: can we name the exception rather after the exceptional condition encountered (`InvalidCacheEntry`) rather than what we intend to do in the exception handler (invalidate/remove that entry).

enkore: This code is removed in the following commit.

https://github.com/borgbackup/borg/pull/2572/files/5b3667b61760696f5cca985dc556bbb551997aeb..4faaa7d1fa798556a780cb6e9a8f41b8ec106b5a

ThomasWaldmann: yes, just noticed that...

src/borg/remote.py
```diff
@@ -1220,8 +1209,7 @@ def unpack(data):
             csize, checksum = cache_struct.unpack(data[:cache_struct.size])
             compressed = data[cache_struct.size:]
             if checksum != xxh64(compressed):
-                logger.warning('Repository metadata cache: detected corrupted data in cache!')
-                raise RepositoryCache.InvalidateCacheEntry
+                raise IntegrityError('detected corrupted data in metadata cache')
```
ThomasWaldmann: hmm, just as a vague idea: could we get into similar troubles by this as we had with OSError?

i mean: we shouldn't kill a repo chunk after confusing a repo chunk IE with a cache IE... (this somehow fits into the scope of the ticket "review IE, especially when they kill").

ThomasWaldmann:
you removed this from b7 milestone a few days ago, so what's your intention with this?

enkore:
Special-cases aside ("this needs!11 to be fixed for this release"), I consider milestones only for issues, not pull requests.

src/borg/remote.py
enkore: The same problem affects the encrypted cache. Bit flip in cache -> IntegrityError. So a chunk killer that uses a caching wrapper would always need to check whether the IE was caused by the cache.

ThomasWaldmann:
Well, as big new features should not go "last minute" into release candidates, it needs to be in a beta if it shall be in 1.1.0.

enkore:
That's irrelevant, perhaps you misunderstood.

There is little point for me or anyone to do further work based on either the master branch, where the cache is outdated, so whatever is merged first causes time-consuming rebases in my other topic branches (cache-sync-free or whatever),
**or** basing my further work on topic branches forked from _this_ branch, since any changes done on the base again cause time-consuming rebases. Therefore, I don't do any work in these areas. Therefore, delays ensue.

ThomasWaldmann:
That's a somehow long and un-parseable answer to a simple question, so let me make it even simpler:

Shall this PR go into
- [ ] b6
- [ ] b7
- [ ] not into 1.1.0 at all

enkore:
I said one week ago that it's fine for b6.

src/borg/cache_sync/unpack.h
```diff
@@ -0,0 +1,391 @@
...
+    default:
+        if(u->inside_chunks) {
+            SET_LAST_ERROR("Invalid state transition (unexpected array end)");
+            return -1;
+        } else {
+            u->level--;
+            return 0;
+        }
+    }
+
```
ThomasWaldmann: remove empty line (maybe in general try to make whitespace consistent).

src/borg/cache_sync/unpack.h
```diff
@@ -0,0 +1,391 @@
...
+static inline int unpack_callback_ext(unpack_user* u, const char* base, const char* pos,
+                                      unsigned int length)
+{
+    (void)u; (void)base; (void)pos; (void)length;
+    UNEXPECTED("ext");
+    return 0;
+}
+
+#include "unpack_template.h"
```
ThomasWaldmann: single letter variable names are rather frequent in this file, but as long as the functions are rather short, it's no big problem.

additionally, 1 and l are just 1px apart in some fonts.

src/borg/cache_sync/unpack_template.h
```diff
@@ -0,0 +1,360 @@
+/*
+ * MessagePack unpacking routine template
```
ThomasWaldmann: maybe add some words what you changed here.

src/borg/hashindex.pyx
```diff
@@ -375,3 +384,24 @@ cdef class ChunkKeyIterator:
...
+        self.sync = cache_sync_init(self.chunks.index)
+        if not self.sync:
+            raise Exception('cache_sync_init failed')
+
+    def __dealloc__(self):
+        if self.sync:
+            cache_sync_free(self.sync)
+
+    def feed(self, chunk):
+        if not cache_sync_feed(self.sync, <char *>chunk, len(chunk)):
```
ThomasWaldmann: <void *>chunk ?

src/borg/_hashindex.c
```diff
@@ -1,6 +1,6 @@
-#include <Python.h>
```
ThomasWaldmann: how do the defs in Python.h come in here in future?

enkore: From Cython

scripts/fuzz-cache-sync/HOWTO
```diff
@@ -0,0 +1,10 @@
+- Install AFL and the requirements for LLVM mode (see docs)
+- Compile the fuzzing target, e.g.
+
+  AFL_HARDEN=1 afl-clang-fast main.c -o fuzz-target -O3
+
+  (other options, like using ASan or MSan are possible as well)
+- Add additional test cases to testcase_dir
+- Run afl, easiest (but inefficient) way;
+
+  afl-fuzz -i testcase_dir -o findings_dir ./fuzz-target
```
enkore: When writing a parser, fuzzing before shipping is just good practice nowadays.

scripts/fuzz-cache-sync/main.c
```diff
@@ -0,0 +1,30 @@
...
+        ret = cache_sync_feed(ctx, buf, len);
+        if(!ret && cache_sync_error(ctx)) {
+            fprintf(stderr, "error: %s\n", cache_sync_error(ctx));
+            return 1;
+        }
+    }
+    hashindex_free(idx);
+    cache_sync_free(ctx);
+    return 0;
+}
```
enkore: It's a sample input to AFL (testcase_dir)

src/borg/cache_sync/cache_sync.c
```diff
@@ -0,0 +1,105 @@
+
```
enkore: Sure

src/borg/cache_sync/cache_sync.c
```diff
@@ -0,0 +1,105 @@
...
+            memcpy(new_buf, ctx->buf + ctx->head, ctx->tail - ctx->head);
+            free(ctx->buf);
+            ctx->buf = new_buf;
+            ctx->tail -= ctx->head;
+            ctx->head = 0;
+            ctx->size = new_size;
+        }
+    }
+
+    memcpy(ctx->buf + ctx->tail, data, length);
```
enkore: That's news to me — both hashindex and chunker use memcpy, among other things

src/borg/hashindex.pyx
enkore: You raise a good point there. I'll replace this with the usual ro_buffer, so that it shall work with any linear bag of bytes.

src/borg/_hashindex.c
```diff
@@ -56,8 +56,10 @@ typedef struct {
     int lower_limit;
     int upper_limit;
     int min_empty;
+#ifdef Py_PYTHON_H
```
textshell: This looks like abusing a include guard here?

scripts/fuzz-cache-sync/main.c
ThomasWaldmann: maybe mention it in the fuzzing readme or call it test_simple.afl ?

src/borg/cache_sync/cache_sync.c
ThomasWaldmann: well, I remember we had that memcpy issue already, likely somehow related to a pyinstaller binary not working on a older system any more due to glibc requirement.

src/borg/_hashindex.c
enkore: What's your point?

src/borg/cache_sync/cache_sync.c
enkore: Google only turns this up in connection to borg: https://github.com/pyinstaller/pyinstaller/issues/2160#issuecomment-242971327

I still don't see the connection here, the modules mentioned above have been using memcpy forever with no issues.

ThomasWaldmann: the problem is when building a binary on a glibc >= 2.14 system, it won't run on a glibc < 2.14 system, just because of that memcpy madness (and although most other glibc dependencies only require a much older glibc).

enkore: I got that, I just don't see any reason why it should be different between _hashindex.c and cache_sync.c which end up in the same CU. Either memcpy is bad and should not be used, then it needs replacing everywhere, or it's ok.

ThomasWaldmann: hmm, wheezy is glibc 2.13. so while building on that, problem does not manifest.

enkore: Do newer GCC versions still do this? 4.9 (Debian 8) does, but that's now three major releases behind.

Anyway, I'll just open a ticket.

enkore: #2652

scripts/fuzz-cache-sync/main.c
enkore: The directory is mentioned in the HOWTO file

src/borg/cache_sync/unpack.h
```diff
@@ -0,0 +1,390 @@
...
+
+
+static inline int unpack_callback_uint32(unpack_user* u, uint32_t d)
+{
+    return unpack_callback_int64(u, d);
+}
+
+static inline int unpack_callback_uint64(unpack_user* u, uint64_t d)
+{
+    return unpack_callback_int64(u, d);
```
textshell: the msgpack-python doesn't wrap very large values to negative values here.

src/borg/cache_sync/unpack_template.h
```diff
@@ -0,0 +1,365 @@
...
+
+#ifndef USE_CASE_RANGE
+#if !defined(_MSC_VER)
+#define USE_CASE_RANGE
+#endif
+#endif
+
+typedef struct unpack_stack {
+    size_t size;
+    size_t count;
```
textshell: This changes signed types (Py_ssize_t) to unsigned, but i can't see any reason why using unsigned types here could be a problem (the code that sets them uses unsigned types too) -> seems ok

src/borg/cache_sync/unpack_template.h
```diff
@@ -0,0 +1,365 @@
...
+{
+    ctx->cs = CS_HEADER;
+    ctx->trail = 0;
+    ctx->top = 0;
+    unpack_init_user_state(&ctx->user);
+}
+
+#define construct 1
+
+static inline int unpack_execute(unpack_context* ctx, const char* data, size_t len, size_t* off)
```
textshell: This changes signed types (Py_ssize_t) to unsigned, but i can't see any reason why using unsigned types here could be a problem. `*off` is set from a substracton, but if that yields a negative result then the code would have read from a pointer before `data`, which should not be possible.

src/borg/_hashindex.c
textshell: I dislike depending on internal things like that when avoidable. The cleanest would be an explicit define HASHINDEX_WITHOUT_PYTHON (set explicitly in the testing / fuzzing compilation). But picking something from the public api would still be better.

src/borg/fuse.py
```diff
@@ -340,7 +340,7 @@ def read(self, fh, offset, size):
                     # evict fully read chunk from cache
                     del self.data_cache[id]
             else:
-                data = self.key.decrypt(id, self.repository.get(id))
+                data = self.key.decrypt(id, self.repository_uncached.get(id))
```
textshell: this could be a large performance regression for specific usecases. It's unfortunate to hide this in a mildly related pull request.

src/borg/remote.py
```diff
@@ -1058,52 +1068,156 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                with open(file, 'rb') as fd:
+                    self.hits += 1
+                    yield self.unpack(fd.read())
+            else:
                 for key_, data in repository_iterator:
                     if key_ == key:
-                        if len(data) <= self.THRESHOLD:
-                            self.caching_repo.put(key, data)
-                        yield data
+                        transformed = self.add_entry(key, data, cache)
```
textshell: Does this for loop (over repository_iterator) really make sense in this usage?
I know it was like this before, but now cache evictions change the picture a bit. From my reading as soon as a evicted key is fed into this part of the code it just retrieves and discards all remaining chunks in the iterator and then needs to fall back to the fallback path for everything. It seems the for should only be done `if key in unknown_keys:` then we are back to only triggering the strange `key_ != key` case when the repository is confused. Not sure if there is much sense in not just aborting in that case though.

enkore: I don't think the Remote code is re-entrant, so you can't have frames in call_many alive (=have a generator laying around) _and_ do Repository calls at the same time.

A much easier workaround is to simply avoid evicting keys from the current remaining keys set.

src/borg/fuse.py
enkore: The history of the file doesn't make it look to me like that was all that intentional.

Note that repeated reads of the same VFS file are caught by kernel caching. The only case where this can make a (big) difference is e.g. reading deduplicated files in alternation or something like that, where there actually would be any advantage from caching chunks.

src/borg/remote.py
textshell: I fear the remote repo is used in that way all over the place. I do not actually remember all the constraints. That code really needs to be rewritten in 1.2-dev.
But i agree not evicting in one call would solve this too. but that would also mean not caching in that call if the size limit is reached. (which is ok with me, if we free enough space that is rarely happens)

src/borg/remote.py
```diff
@@ -1058,52 +1068,156 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def add_entry(self, key, data, cache):
+        transformed = self.transform(key, data)
+        if not cache:
+            return transformed
+        packed = self.pack(transformed)
+        file = self.key_filename(key)
+        try:
+            with open(file, 'wb') as fd:
+                fd.write(packed)
+        except OSError as os_error:
```
textshell: shouldn't this remove the partially written and unaccounted file?

src/borg/remote.py
```diff
@@ -1058,52 +1068,156 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        self.hits = 0
+        self.misses = 0
+        self.slow_misses = 0
+        self.slow_lat = 0.0
+        self.evictions = 0
+        self.enospc = 0
+
+    def query_size_limit(self):
+        stat_fs = os.statvfs(self.basedir)
+        available_space = stat_fs.f_bsize * stat_fs.f_bavail
```
textshell: maybe refuse outright if available_space is very small?

src/borg/fuse.py
textshell: Yes for small cache sizes (<< memory) the VFS will likely be doing a good job. For cache sizes >> memory there are use cases where this would still be useful. I would expect mounting block device / image backups via a borg mount -> loopback setup would likely be cases where is make a significant difference. We could open a ticket where we can discuss if we care. Because this PR is of course already merged.

textshell: I've opened #2662 for further discussion.

src/borg/cache_sync/unpack.h
enkore: I suppose one could swap u/i here and avoid IB. Wrapping of large/negative values isn't really important, since valid size/csize are <uint32_t. Due to chunk ID verification the authoring client would need to write garbage, and it still can write garbage even if we checked that it didn't wrap.

PR #2573, opened 2017-05-27
Added BORG_PASSCOMMAND environment variable

TuXicc:
The BORG_PASSCOMMAND variable can be used to specify an external program that is asked to supply the passphrase.

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -134,9 +134,16 @@ General:
         It is used when a passphrase is needed to access a encrypted repo as well as when a new
         passphrase should be initially set when initializing an encrypted repo.
         See also BORG_NEW_PASSPHRASE.
+    BORG_PASSCOMMAND
+        When set, use the command to answer the passphrase question for encrypted repositories.
+        It is used when a passphrase is needed to access a encrypted repo as well as when a new
+        passphrase should be initially set when initializing an encrypted repo.
+        If BORG_PASSPHRASE is also set, it takes precedence.
+        See also BORG_NEW_PASSPHRASE.
     BORG_NEW_PASSPHRASE
         When set, use the value to answer the passphrase question when a **new** passphrase is asked for.
-        This variable is checked first. If it is not set, BORG_PASSPHRASE will be checked also.
+        This variable is checked first. If it is not set, BORG_PASSPHRASE and BORG_PASSCOMMAND will be
+        checked also.
         Main usecase for this is to fully automate ``borg change-passphrase``.
     BORG_DISPLAY_PASSPHRASE
         When set, use the value to answer the "display the passphrase for verification" question when defining a new passphrase for encrypted repositories.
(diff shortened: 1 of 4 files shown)


src/borg/crypto/key.py
```diff
@@ -406,18 +410,33 @@ def init_ciphers(self, manifest_nonce=0):
...
+        passphrase = cls._get_env('BORG_PASSPHRASE', default)
+        if passphrase is not None:
+            return passphrase
+        passphrase = cls.env_passcommand()
+        if passphrase is not None:
+            return passphrase
+
+    @classmethod
+    def env_passcommand(cls, default=None):
+        passcommand = cls._get_env('BORG_PASSCOMMAND', default)
```
ThomasWaldmann: that's a bit weird. the returned object is of class `Passphrase` (if not None), but does contain the command, not a passphrase. so i guess you'ld rather call `os.environ.get('BORG_PASSCOMMAND', default)` here.

ThomasWaldmann: ... and when doing it like that, you maybe do not need to rename _env_passphrase method.

src/borg/crypto/key.py
```diff
@@ -406,18 +410,33 @@ def init_ciphers(self, manifest_nonce=0):
...
+        if passcommand is not None:
+            try:
+                passphrase = subprocess.check_output(passcommand, shell=True, universal_newlines=True)
+            except subprocess.CalledProcessError as e:
+                raise PasscommandFailure(e)
+            return Passphrase(passphrase.rstrip('\n'))
```
ThomasWaldmann:
```
return cls(...)
```

docs/usage_general.rst.inc
```diff
@@ -131,12 +131,19 @@ General:
         can either leave it away or abbreviate as `::`, if a positional parameter is required.
     BORG_PASSPHRASE
         When set, use the value to answer the passphrase question for encrypted repositories.
-        It is used when a passphrase is needed to access a encrypted repo as well as when a new
+        It is used when a passphrase is needed to access an encrypted repo as well as when a new
         passphrase should be initially set when initializing an encrypted repo.
         See also BORG_NEW_PASSPHRASE.
+    BORG_PASSCOMMAND
+        When set, use the command to answer the passphrase question for encrypted repositories.
```
ThomasWaldmann: ... via the standard output of the command (trailing newlines are stripped).

src/borg/crypto/key.py
enkore: os.environ.get(..., None); the passed `default` is meant as the default for the passphrase

src/borg/crypto/key.py
```diff
@@ -406,18 +410,33 @@ def init_ciphers(self, manifest_nonce=0):
...
+        passphrase = cls.env_passcommand()
+        if passphrase is not None:
+            return passphrase
+
+    @classmethod
+    def env_passcommand(cls, default=None):
+        passcommand = cls._get_env('BORG_PASSCOMMAND', default)
+        if passcommand is not None:
+            try:
+                passphrase = subprocess.check_output(passcommand, shell=True, universal_newlines=True)
```
enkore: I'd prefer to avoid shell=True

src/borg/crypto/key.py
ThomasWaldmann: right, what @enkore says.

enkore:
Thanks!

enkore:
Stapling #392

PR #2575, opened 2017-05-29
patterns: don't recurse with !/--exclude for pf:, fixes #2509

### LINKED ISSUE #2509: ! pf:/some/dir Does not exclude directory
I have a folder `~/borg-test` containing a file `f0` and directories `d1` and `d2` containing files `f1` and `f2` respectively. When I created an archive of it using the following patterns file:
```
R /home/me/borg-test
! pf:/home/me/borg-test/d1
```
it included all files and directories, and did not exclude `d1`.
Using `-` instead of `!` didn't exclude it either.

This is using the standalone binary borg-linux64 of version 1.1.0b5

edgimar:


src/borg/patterns.py
```diff
@@ -150,9 +150,9 @@ def match(self, path):
 
         if value is not non_existent:
             # we have a full path match!
-            # TODO: get from pattern; don't hard-code
-            self.recurse_dir = True
-            return value
+            self.recurse_dir = (value is IECommand.ExcludeNoRecurse)
```
enkore: This looks like it's inverted, setting recurse_dir to True when value says Exclude*NoRecurse*.

ThomasWaldmann: seen at another place:
```
recurse_dir = cmd not in [IECommand.ExcludeNoRecurse]
```

edgimar: sorry, yes it was inverted - fixed.

src/borg/patterns.py
```diff
@@ -150,9 +150,9 @@ def match(self, path):
 
         if value is not non_existent:
             # we have a full path match!
-            # TODO: get from pattern; don't hard-code
-            self.recurse_dir = True
-            return value
+            self.recurse_dir = (value is not IECommand.ExcludeNoRecurse)
+
```
ThomasWaldmann: hmm, an empty line here is not really needed/helpful, is it?

also, can we do the same check at both places in the same way? (see my previous comment)

ThomasWaldmann:
@edgimar I'ld like to release b6 this weekend, can you finish the PR?

edgimar:
@ThomasWaldmann should be OK now.

src/borg/patterns.py
```diff
@@ -325,6 +324,11 @@ class IECommand(Enum):
     ExcludeNoRecurse = 5
 
 
+def command_recurses_dir(cmd):
+    # TODO?: raise error or return None if *cmd* is RootPath or PatternStyle
```
ThomasWaldmann: guess a ValueError / TypeError might be appropriate.

PR #2568, opened 2017-05-25
1101.integration.cache

enkore:
~~TODO: Corruption tests (manual testing: All good!); Add dedicated test cases for ICF (e: not really necessary).~~

Initially the design provided only the equivalent of "detached signatures", i.e. those .integrity files. However, that caused various design pains, because these files naturally did not use the same transaction mechanisms.

Therefore, I adjusted the design. Detached integrity files are now the exception, not the norm, and can be used where they are apt (archive.chunks.d). Otherwise, integrity data is stored where the surrounding code can accommodate it well. For the cache, this is the `config`. For the repository I will make it `N.ancillary`, since `N.hints` requires integrity checks (N>0 tickets with hints corruption; they're also much larger than the cache config).

Part of our ongoing series to realize #1101 (#1688 #2496 #2502) 

~~NOTE: HashIndex does not use file parts yet. This should probably be changed in this PR as well, to avoid merging a known-unstable format into `master`.~~

✔ Borg 1.0.x interop (with grace!)
✔ Tests
✔ Works in practice
✔ Not tested on animals

PR #2580, opened 2017-06-01
1101.integration.xxh

enkore:
A cryptographic hash is slow for providing features not required here. I don't want to use a CRC since we already know from both theory and practice (use in Repository) that they are not that good at detecting storage errors. So xxh64 it is, specifically designed as a fast storage checksum over large data blocks (used in LZ4, zstd and various media/footage-management products).

(If we ever redo the Repository format, using it as a checksum instead makes sense. The implementation is light on cache, unlike table-based CRC, but as fast as crc32_clmul - and for longer blocks it's a bit faster, without requiring more than plain C compiled to plain x64).

Part of my ongoing series to realize #1101 (#1688 #2496 #2502 #2568)

### DIFF
--- a/src/borg/algorithms/checksums.pyx
+++ b/src/borg/algorithms/checksums.pyx
@@ -25,11 +25,11 @@ cdef extern from "xxh64/xxhash.c":
         XXH_OK,
         XXH_ERROR
 
-    XXH64_hash_t XXH64 (const void* input, size_t length, unsigned long long seed);
+    XXH64_hash_t XXH64(const void* input, size_t length, unsigned long long seed);
 
-    XXH_errorcode XXH64_reset  (XXH64_state_t* statePtr, unsigned long long seed);
-    XXH_errorcode XXH64_update (XXH64_state_t* statePtr, const void* input, size_t length);
-    XXH64_hash_t  XXH64_digest (const XXH64_state_t* statePtr);
+    XXH_errorcode XXH64_reset(XXH64_state_t* statePtr, unsigned long long seed);
+    XXH_errorcode XXH64_update(XXH64_state_t* statePtr, const void* input, size_t length);
+    XXH64_hash_t XXH64_digest(const XXH64_state_t* statePtr);
 
     void XXH64_canonicalFromHash(XXH64_canonical_t* dst, XXH64_hash_t hash);
     XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src);
--- a/src/borg/algorithms/xxh64/xxhash.c
+++ b/src/borg/algorithms/xxh64/xxhash.c
@@ -1,0 +1,615 @@
+/*
+*  xxHash - Fast Hash algorithm
+*  Copyright (C) 2012-2016, Yann Collet
+*
+*  BSD 2-Clause License (http://www.opensource.org/licenses/bsd-license.php)
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions are
+*  met:
+*
+*  * Redistributions of source code must retain the above copyright
+*  notice, this list of conditions and the following disclaimer.
+*  * Redistributions in binary form must reproduce the above
+*  copyright notice, this list of conditions and the following disclaimer
+*  in the documentation and/or other materials provided with the
+*  distribution.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+*  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+*  OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+*  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+*  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+*  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+*  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+*  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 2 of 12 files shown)


ThomasWaldmann:
Hmm, from a "keep code amount we have to care for minimal" perspective:

How about using blake2b? It's also faster than sha512 and we already depend on it anyway (and we already have code taking it from a external lib, if available).

As we use it for the index files only currently, it (due to lower volume, some GBs) is not as speed critical as the checksums for the backup data written into the repository (TBs).

I'ld also like something better than crc32 for the repo, but that is not going to happen soon - so we could delay adding xxhash until we decide to change the repo format (borg 2.0) if we feel the added speed over blake2b is worth it.

ThomasWaldmann:
As a general note: I vaguely remember reading some stuff about that the size of the checksum (relative to the size of the data being checksummed) is important.

So, e.g., it is bad to use a big checksum/hash (e.g. 256 or 512bit) if there is only little data (few bytes, like e.g. a chunk header).

enkore:
chunks.archive.d is large (x-xxx GB) and generation / reading is *already* CPU constrained. Blake2b is a bit faster than SHA-512, but not that much, and more importantly, not universally.

I'm principally open to alternate suggestions, but none of the algorithms we are using so far are a good fit.

> As a general note: I vaguely remember reading some stuff about that the size of the checksum (relative to the size of the data being checksummed) is important.

That's correct. Considering a channel with a defined BER, an optimal length exists for each _individual_ checksum where the system BER (sum of the true positives and false positives due to corrupted checksum) is minimal.

ThomasWaldmann:
@enkore do you have throughput numbers of the sha512 / blake2b / xxhash code as used by us?

ThomasWaldmann:
hmm, is there no libxxh or something for this we could use alternatively?
debian, you know, likes to rip out bundled 3rd party code...

enkore:
With SHA512 merging indices becomes about 75 % (80 s) slower compared
to XXH (45.5 s). I didn't test with BLAKE2b (because I'd have to write
a new streaming wrapper for it to do so); rule of three from the
numbers we know works out to about 55 % slower. Baseline, without
checking, is 44s.

I made another series of tests to look at the difference between base
(no checking) and XXH, this is quicker to do since I can just patch
the DICF usage out of the cache (no need to re-generate the cache),
the difference (with fairly good, i.e. low variance) comes out to 1.7
s extra CPU time or close to -4 %.

That works out to about 12 GB/s inside the application. Simpler
testing suggests 15 GB/s, the same as crc32_clmul; the difference is
likely because during the sync there is a whole lot of other code and
data hit, competing for cache and higher overhead for streaming (which
requires additional logic and buffer management on the function's
part).

The data set is basically the same I used in #2572 and I also merged
#2572 into a temporary branch to create the chunks.archive.d,
otherwise this testing would take a lot longer...

The relative impact on writing chunks.archive.d is of course lower
since that process simply takes much longer overall. On the other
hand, a file in chunks.archive.d is written only once, but read more
than once; for every cache sync until the corresponding archive is
deleted.

enkore:
> hmm, is there no libxxh or something for this we could use alternatively?

Not that I'm aware of. There is a PyPI package which also vendors the code, but that'd be a pip dependency (still no OS packages) and I'd have to review the code to see whether it's bad or not, I don't have to do that here, since I know it's good.

As an side, for many users of XXH it would not make much sense to use a shared library, or, in other words, it would be bad to do so. It's typical for checksumming to be stitched into other processing, e.g. if you're a compressor, you'd stitch the checksum and copying/writing to the output buffer; you want to avoid extra function calls there.

src/borg/algorithms/checksums.pyx
```diff
@@ -0,0 +1,106 @@
...
+        XXH_ERROR
+
+    XXH64_hash_t XXH64 (const void* input, size_t length, unsigned long long seed);
+
+    XXH_errorcode XXH64_reset  (XXH64_state_t* statePtr, unsigned long long seed);
+    XXH_errorcode XXH64_update (XXH64_state_t* statePtr, const void* input, size_t length);
+    XXH64_hash_t  XXH64_digest (const XXH64_state_t* statePtr);
+
+    void XXH64_canonicalFromHash(XXH64_canonical_t* dst, XXH64_hash_t hash);
+    XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src);
```
ThomasWaldmann: can we have consistent style? compare 30..32 and 34..35 (and other places).

src/borg/algorithms/xxh64/xxhash.c
```diff
@@ -0,0 +1,615 @@
...
+    return XXH_readLE32_align(ptr, endian, XXH_unaligned);
+}
+
+/* *************************************
+*  Macros
+***************************************/
+#define XXH_STATIC_ASSERT(c)   { enum { XXH_static_assert = 1/(int)(!!(c)) }; }    /* use only *after* variable declarations */
+XXH_PUBLIC_API unsigned XXH_versionNumber (void) { return XXH_VERSION_NUMBER; }
+
+#ifndef XXH_NO_LONG_LONG
```
ThomasWaldmann: what if that is defined? do we have to care?

src/borg/algorithms/checksums.pyx
enkore: You, uh, mean the parens placement?

src/borg/algorithms/xxh64/xxhash.c
enkore: If we'd define this configure-like option then XXH64 would not be compiled in.

src/borg/algorithms/checksums.pyx
ThomasWaldmann: yes, or more in general: futile attempts to align things.

src/borg/algorithms/xxh64/xxhash.c
ThomasWaldmann: but it is not like we lose XXH64 on some platform / compiler we otherwise support?

enkore: No.

(Note: Borg already uses this code by way of LZ4 which uses xxh32 for its checksum)

enkore: I'll run a test on PPC, I don't expect any surprises though. I don't think I operate x86 anymore though, so can't test that one easily now.

enkore: Ok.

ThomasWaldmann: I'll test on some 32bit OS.

enkore: FUSE tests fail for me in wheezy32, everything else is ok, incl. checksums.

ThomasWaldmann: the problem here is that a reboot is required as the fuse module does not fit the running kernel after security updates.

ThomasWaldmann: ppc BE test was also ok?

enkore: Yes (see above)

PR #2584, opened 2017-06-01
repository: checksum index and hints

enkore:
Part of my ongoing series to realize #1101 (#1688 #2496 #2502 #2568 #2580)

~~Could use a few more tests for specific corruption scenarios.~~

Unlike the cache design this can have false positives whenever an older version rewrites the auxiliary files for a transaction created by a newer version, because that might result in a different index (resizing) or hints file (hash ordering), while not invalidating the integrity file. E.g. using 1.1 on a repository, noticing corruption or something, running 1.0 check --repair, which rewrites the index and hints, then using 1.1 again; 1.1 would complain about index and/or hints corruption and rebuild them. I don't think this is noteworthy enough to be included in the error messages, but is something to mention when I type up design docs for this stuff.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -373,17 +373,18 @@ class Repository:
         self.write_index()
         self.rollback()
 
-    def _read_integrity(self, transaction_id, key=None):
-        integrity_path = os.path.join(self.path, 'integrity.%d' % transaction_id)
+    def _read_integrity(self, transaction_id, key):
+        integrity_file = 'integrity.%d' % transaction_id
+        integrity_path = os.path.join(self.path, integrity_file)
         try:
             with open(integrity_path, 'rb') as fd:
                 integrity = msgpack.unpack(fd)
         except FileNotFoundError:
             return
-        if key:
-            return integrity[key].decode()
-        else:
-            return integrity
+        if integrity.get(b'version') != 2:
+            logger.warning('Unknown integrity data version %r in %s', integrity.get(b'version'), integrity_file)
+            return
+        return integrity[key].decode()
 
     def open_index(self, transaction_id, auto_recover=True):
         if transaction_id is None:
@@ -617,7 +618,7 @@ class Repository:
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
                 logger.debug('complete_xfer: deleting unused segment %d', segment)
-                assert self.segments.pop(segment) == 0
+                assert self.segments.pop(segment) == 0, 'Corrupted segment reference count - corrupted index or hints'
                 self.io.delete_segment(segment)
                 del self.compact[segment]
             unused = []
@@ -711,7 +712,7 @@ class Repository:
                             new_segment, size = self.io.write_delete(key)
                         self.compact[new_segment] += size
                         segments.setdefault(new_segment, 0)
-            assert segments[segment] == 0
+            assert segments[segment] == 0, 'Corrupted segment reference count - corrupted index or hints'
             unused.append(segment)
             pi.show()
         pi.finish()
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -24,6 +24,7 @@ from .logger import create_logger
 from .lrucache import LRUCache
 from .platform import SaveFile, SyncFile, sync_dir, safe_fadvise
 from .algorithms.checksums import crc32
+from .crypto.file_integrity import IntegrityCheckedFile, FileIntegrityError
 
 logger = create_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/borg/repository.py
```diff
@@ -372,13 +373,28 @@ def commit(self, save_space=False):
...
 
+    def _read_integrity(self, transaction_id, key):
+        integrity_file = 'integrity.%d' % transaction_id
+        integrity_path = os.path.join(self.path, integrity_file)
+        try:
+            with open(integrity_path, 'rb') as fd:
+                integrity = msgpack.unpack(fd)
+        except FileNotFoundError:
+            return
+        if integrity.get(b'version') != 2:
```
ThomasWaldmann: some special reason to start at 2 here?

src/borg/repository.py
```diff
@@ -459,28 +476,66 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
+            os.rename(file + '.tmp', file)
+
+        hints = {
+            b'version': 2,
+            b'segments': self.segments,
+            b'compact': self.compact,
+            b'storage_quota_use': self.storage_quota_use,
+        }
+        integrity = {
+            b'version': 2,
```
ThomasWaldmann: same here. did you want to use same version as for hints file?

src/borg/repository.py
```diff
@@ -459,28 +476,66 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
+        hints_file = os.path.join(self.path, hints_name)
+        with IntegrityCheckedFile(hints_file + '.tmp', filename=hints_name, write=True) as fd:
+            msgpack.pack(hints, fd)
+            flush_and_sync(fd)
+        integrity[b'hints'] = fd.integrity_data
+
+        # Write repository index
+        index_name = 'index.%d' % transaction_id
+        index_file = os.path.join(self.path, index_name)
+        with IntegrityCheckedFile(index_file + '.tmp', filename=index_name, write=True) as fd:
```
ThomasWaldmann: previously, the filename was just "index.tmp" (not "index.N.tmp") - does some cleanup code need to be changed?

src/borg/testsuite/repository.py
```diff
@@ -537,13 +544,76 @@ def test_index_outside_transaction(self):
         with self.repository:
             assert len(self.repository) == 1
 
+    def test_index_corrupted(self):
+        self.corrupt(os.path.join(self.repository.path, 'index.1'))
+        with self.repository:
+            assert len(self.repository) == 1
+            assert self.repository.get(H(0)) == b'foo'
```
ThomasWaldmann: what's the point of this test? maybe add a comment.

src/borg/testsuite/repository.py
```diff
@@ -537,13 +544,76 @@ def test_index_outside_transaction(self):
...
+    def test_unknown_integrity_version(self):
+        integrity_path = os.path.join(self.repository.path, 'integrity.1')
+        with open(integrity_path, 'r+b') as fd:
+            msgpack.pack({
+                b'version': 4.7,
+            }, fd)
+            fd.truncate()
+        with self.repository:
+            assert len(self.repository) == 1
+            assert self.repository.get(H(0)) == b'foo'
```
ThomasWaldmann: also add a comment here

src/borg/repository.py
enkore: No, the clean-up code removes all files starting with index./hints. (and now integrity. as well). There seemed to be no particular reason for the inconsistency.

src/borg/repository.py
enkore: Yep

enkore:
> + Coverage   83.64%   83.84%   +0.19%     

Maybe we get to 85% before 1.1? ;)

src/borg/repository.py
```diff
@@ -459,28 +476,68 @@ def prepare_txn(self, transaction_id, do_cleanup=True):
...
+
+        hints = {
+            b'version': 2,
+            b'segments': self.segments,
+            b'compact': self.compact,
+            b'storage_quota_use': self.storage_quota_use,
+        }
+        integrity = {
+            # Integrity version started at 2, the current hints version.
+            # Thus, integrity version == hints version, fow now.
```
ThomasWaldmann: for

src/borg/testsuite/repository.py
```diff
@@ -537,13 +539,109 @@ def test_index_outside_transaction(self):
...
+            fd.truncate()
+
+    def test_subtly_corrupted_hints(self):
+        self._subtly_corrupted_hints_setup()
+        with self.repository:
+            self.repository.append_only = False
+            self.repository.put(H(3), b'1234')
+            # Do a compaction run. Succeeds, since the failed checksum prompted a rebuild of the index+hints.
+            self.repository.commit()
+
```
ThomasWaldmann: remove empty line?

PR #2585, opened 2017-06-02
vagrant: control VM cpus and pytest workers via env

ThomasWaldmann:


Vagrantfile
```diff
@@ -3,6 +3,10 @@
 
 # Automated creation of testing environments / binaries on misc. platforms
 
+$cpus = Integer(ENV.fetch('VMCPUS', '4'))  # create VMs with that many cpus
```
ThomasWaldmann: default maybe appropriate for a dual-core with HT.

Vagrantfile
```diff
@@ -3,6 +3,10 @@
...
+$xdistn = Integer(ENV.fetch('XDISTN', '4'))  # dispatch tests to that many pytest workers
```
ThomasWaldmann: this is a rather low value (also used in tox.ini as default) - uses little memory and causes relatively low load.

Vagrantfile
```diff
@@ -263,10 +269,10 @@ def install_borg(fuse)
     pip install -U wheel  # upgrade wheel, too old for 3.5
     cd borg
     # clean up (wrong/outdated) stuff we likely got via rsync:
-    rm -f borg/*.so borg/*.cpy*
-    rm -f borg/{chunker,crypto,compress,hashindex,platform_linux}.c
-    rm -rf borg/__pycache__ borg/support/__pycache__ borg/testsuite/__pycache__
+    rm -rf __pycache__
+    find src -name '__pycache__' -exec rm -rf {} \;
```
ThomasWaldmann: this did not match the src/borg/... dir structure any more and thus frequently causes troubles.
the rm and find are both needed, did not want to start the find in "." as it then also finds/removes all `__pycache__` in .tox.

Vagrantfile
```diff
@@ -263,10 +269,10 @@ def install_borg(fuse)
...
     pip install -r requirements.d/development.txt
+    python setup.py clean
```
ThomasWaldmann: thanks for that, much better now. :)

Vagrantfile
```diff
@@ -442,6 +448,9 @@ Vagrant.configure(2) do |config|
 
   config.vm.define "wheezy64" do |b|
     b.vm.box = "boxcutter/debian7"
+    b.vm.provider :virtualbox do |v|
+      v.memory = 1024 + $wmem
+    end
```
ThomasWaldmann: this somehow was lost / not added at some time...

Vagrantfile
```diff
@@ -457,6 +466,7 @@ Vagrant.configure(2) do |config|
   config.vm.define "darwin64" do |b|
     b.vm.box = "jhcook/yosemite-clitools"
     b.vm.provider :virtualbox do |v|
+      v.memory = 1536 + $wmem
```
ThomasWaldmann: same here.

Vagrantfile
```diff
@@ -529,7 +540,7 @@ Vagrant.configure(2) do |config|
     b.ssh.insert_key = false
 
     b.vm.provider :virtualbox do |v|
-      v.memory = 2048
+      v.memory = 1536 + $wmem
```
ThomasWaldmann: no support for xdistn yet for windows.

PR #2587, opened 2017-06-02
docs/security: OpenSSL usage

enkore:


enkore:
OpenSSL has a ton of feature flags, I bet disabling a host of stuff we never use could reduce binary size a bit (excluding libssl is ~400k, libcrypto is about 2 MB).

E.g. ASN1, BIO, BN (bignum), CMS, unused algorithms (DES, RSA, elliptic curves, key exchanges), OCSP routines, PEM and X509, PKCS#1-9000, ...

ThomasWaldmann:
Well, we usually do not build openssl, so we can't use these feature flags, I guess.
Also, maybe building openssl has its own traps & pitfalls? Just guessing.

1MB more is maybe preferable over building openssl (and maybe introducing bugs / issue just by doing that).

scripts/borg.exe.spec
```diff
@@ -38,3 +40,16 @@ exe = EXE(pyz,
           strip=False,
           upx=True,
           console=True )
+
+if False:
```
ThomasWaldmann: rather introduce some flag variable set at the top of the file than hardcoding False here.

enkore: I don't agree, having the explanation what this is about is quite sufficient in such a rarely touched file. No extra flag is needed.

PR #2588, opened 2017-06-02
docs: quotas: clarify compatbility; only relevant to serve side

enkore:
also cf. "Enforcing the quota":

    The quota is enforcible only if *all* :ref:`borg_serve` versions
    accessible to clients support quotas

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -233,8 +233,11 @@ state).
 .. rubric:: Compatibility with older servers and enabling quota after-the-fact
 
 If no quota data is stored in the hints file, Borg assumes zero quota is used.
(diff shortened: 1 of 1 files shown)


docs/internals/data-structures.rst
```diff
@@ -233,8 +233,11 @@ state).
 .. rubric:: Compatibility with older servers and enabling quota after-the-fact
 
 If no quota data is stored in the hints file, Borg assumes zero quota is used.
-Thus, if a repository with an enabled quota is written to with an older version
-that does not understand quotas, then the quota usage will be erased.
+Thus, if a repository with an enabled quota is written to with an older ``borg serve``
```
ThomasWaldmann: Same problem happens when a older borg accesses a locally mounted repository (not ssh: repo url).

Issue #1366, opened 2016-07-23
use coala?


ThomasWaldmann:
coala bot that reviews pull requests.

coala detect code duplication.

complexity metrics.

grammar check for docs.

C checks for similar stuff.

Generic stuff checker tabs etc.

enkore:
http://coala-analyzer.org/

enkore:
If we use it, put it in the flake (?) tox env, if that works?

PlasmaPower:
Shouldn't we create another coala environment instead of grouping it with flake8? I'm assuming coala won't totally replace flake8, right?

enkore:
Technically yes, but creating environments is quite slow on travis (because it's "always" the first time on travis), so having a separate environment would make the build run longer by maybe 5 minutes or so.

ThomasWaldmann:
See PR #1569.

sils1297:
> I'm assuming coala won't totally replace flake8, right?

@PlasmaPower if coala isn't able to totally replace flake8 please tell us what's missing :) There was a Flake8Bear in the works which should ease switching in the future as you'll for sure be able to reuse your existing configuration but the contributor didn't continue working on it. (See https://github.com/coala-analyzer/coala-bears/issues/169 ) If that's of interest to you please tell us or consider helping out :)

sils:
Hi, just got this notification as we were deploying the first stable working versionof gitmate.io - you might find that interesting. Just turn it on and it'll use the coafile you already have :)

ThomasWaldmann:
@sils thanks for the info! gitmate looks interesting, but look what I got when clicking "login with github":

![gitmate-access](https://cloud.githubusercontent.com/assets/356103/26754210/90ed9f4e-4876-11e7-9a1f-b6c856127da9.png)

Write access to code repo -> NOPE NOPE NOPE NEVER. Admin access to settings -> NOPE.

So, can you do it with less intrusive permissions (I know github permission system sucks, but did you try?).

sils:
hey, thanks for that feedback :) we're aware that this is an issue and have
already worked out plans to mitigate that before beginning to code.
However, this is an MVP up since a few hours, getting the functionality was
our priority up until now. We're working on using GitHub integrations which
will allow more fine grained permission management and will ultimately
solve that. However to do that we have a long road ahead of us, the "github
integrations" are still in beta even I think. We're also thinking about
this issue on GitLab where they don't have anything like this. Another
[...]

PR #2592, opened 2017-06-03
add a .coafile for coala

ThomasWaldmann:
ignores / disable are set up so that there are not many faults.
we can improve that iteratively.

for trying it: pip install coala coala-bears ; apt install cppcheck


ThomasWaldmann:
fixes #1366.

enkore:
> for trying it: pip install coala coala-bears ; apt install cppcheck

Write that down in the development docs?

Issue #2384, opened 2017-04-04
recreate mixes original and recreated start/end time


enkore:
```
Time (start): Mon, 2017-04-03 21:27:35
Time (end):   Tue, 2017-04-04 18:36:10
Duration: 21 hours 8 minutes 34.32 seconds
```

ThomasWaldmann:
it's not quite clear from the posted output what precisely the bug is.

ThomasWaldmann:
```
(borg-env)tw@tux:~/w/borg$ borg init -e none repo

(borg-env)tw@tux:~/w/borg$ borg create repo::2 .
(borg-env)tw@tux:~/w/borg$ borg info repo::2
Archive name: 2
Archive fingerprint: 6523505b3b85885ff492442cf928e04df01cd2ff78ca86616d80581bb6b30f77
Comment: 
Hostname: tux
Username: tw
Time (start): Sat, 2017-06-03 15:28:28
Time (end): Sat, 2017-06-03 15:28:45
Duration: 16.59 seconds
Number of files: 17047
Command line: /home/tw/w/borg-env/bin/borg create repo::2 .
...

(borg-env)tw@tux:~/w/borg$ borg recreate repo::2 --target 3 --compress none --recompress always
recreate is an experimental feature.
Type 'YES' if you understand this and want to continue: YES
(borg-env)tw@tux:~/w/borg$ borg info repo::3
Archive name: 3
Archive fingerprint: 86ad70485bab8d10bd5027ca3e709370b5561f2170475e62f65f14278babc9b1
Comment: 
Hostname: tux
Username: tw
Time (start): Sat, 2017-06-03 15:28:19
Time (end): Sat, 2017-06-03 15:28:28
Duration: 8.87 seconds
Number of files: 17047
Command line: /home/tw/w/borg-env/bin/borg create repo::2 .
...
```

ThomasWaldmann:
Note (unrelated to this bug): before finding working recreate options, I did this:
```
(borg-env)tw@tux:~/w/borg$ borg recreate repo::2 --target 3
recreate is an experimental feature.
Type 'YES' if you understand this and want to continue: YES
(borg-env)tw@tux:~/w/borg$ borg info repo::3
Archive 3 does not exist

(borg-env)tw@tux:~/w/borg$ borg recreate repo::2 --target 3 --compress none
recreate is an experimental feature.
Type 'YES' if you understand this and want to continue: YES
[...]

PR #2593, opened 2017-06-03
recreate: if single archive is not processed, exit 2

enkore:
#2384

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1598,7 +1598,9 @@ class Archiver:
             if recreater.is_temporary_archive(name):
                 self.print_error('Refusing to work on temporary archive of prior recreate: %s', name)
                 return self.exit_code
-            recreater.recreate(name, args.comment, target)
+            if not recreater.recreate(name, args.comment, target):
+                self.print_error('Nothing to do. Archive was not processed.\n'
+                                 'Specify at least one pattern, PATH, --comment, re-compression or re-chunking option.')
         else:
             if args.target is not None:
                 self.print_error('--target: Need to specify single archive')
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -1598,7 +1598,9 @@ def do_recreate(self, args, repository, manifest, key, cache):
             if recreater.is_temporary_archive(name):
                 self.print_error('Refusing to work on temporary archive of prior recreate: %s', name)
                 return self.exit_code
-            recreater.recreate(name, args.comment, target)
+            if not recreater.recreate(name, args.comment, target):
+                self.print_error('Nothing to do. Archive was not processed.\n'
```
ThomasWaldmann: sounds rather like a warning.

src/borg/archiver.py
```diff
@@ -1608,7 +1610,8 @@ def do_recreate(self, args, repository, manifest, key, cache):
                 if recreater.is_temporary_archive(name):
                     continue
                 print('Processing', name)
-                recreater.recreate(name, args.comment)
+                if not recreater.recreate(name, args.comment):
+                    logger.info('Skipped archive %s: Nothing to do. Archive was not processed.', name)
```
ThomasWaldmann: hmm, maybe also use warning?

src/borg/archiver.py
enkore: No, if you've specified "--target XYZ" or something like that, then the operation would not complete. Thus, its an error.

src/borg/archiver.py
enkore: For mass-recreate this would kinda be expected. E.g. rechunking archives in a mixed repository — it's expected that newer archives would not be processed.

PR #2602, opened 2017-06-04
internals: rewrite manifest & feature flags

enkore:
Docs for #2134

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -339,10 +339,31 @@ the manifest, since an ID check is not possible.
 of Borg preserve its contents (it may have been a better place for *item_keys*,
 which is not preserved by unaware Borg versions).
 
-.. rubric:: Feature flags
+Feature flags
++++++++++++++
 
 Feature flags are used to add features to data structures without causing
-corruption if older versions are used to access or modify them.
+corruption if older versions are used to access or modify them. The main issues
+to consider for a feature flag oriented design are flag granularity,
+flag storage, and cache_ invalidation.
+
+Feature flags are divided in approximately three categories, detailed below.
+Due to the nature of ID-based deduplication, write (i.e. creating archives) and
+read access are not symmetric; it is possible to create archives referencing
+chunks that are not readable with the current feature set. The third
+category are operations that require accurate reference counts, for example
+archive deletion and check.
+
+As the manifest is always updated and always read, it is the ideal place to store
+feature flags, comparable to the super-block of a file system. The only issue problem
+is to recover from a lost manifest, i.e. how is it possible to detect which feature
+flags are enabled, if there is no manifest to tell. This issue is left open at this time,
+but is not expected to be a major hurdle; it doesn't have to be handled efficiently, it just
+needs to be handled.
+
+Lastly, cache_ invalidation is handled by noting which feature
+flags were and were not used to manipulate a cache. This allows to detect whether
+the cache needs to be invalidated, i.e. rebuilt from scratch. See `Cache feature flags`_ below.
 
 The *config* key stores the feature flags enabled on a repository:
 
@@ -396,6 +417,7 @@ Therefore, as soon as any mandatory feature flag is enabled in a repository,
 the manifest version must be switched to version 2 in order to lock out all
 Borg releases unaware of feature flags.
 
+.. _Cache feature flags:
 .. rubric:: Cache feature flags
 
 `The cache`_ does not have its separate flag of feature flags. Instead, Borg stores
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -392,6 +392,36 @@ These can be locked out with manifest version 2. Thus, the only difference betwe
 manifest versions 1 and 2 is that the latter is only accepted by Borg releases
 implementing feature flags.
 
+.. rubric:: Cache feature flags
+
+`The cache`_ does not have its separate flag of feature flags. Instead, Borg stores
+which flags were used to create or modify a cache.
+
+All mandatory manifest features from all operations are gathered in one set.
+Then, two sets of features are computed;
+
+- those features that are supported by the client and mandated by the manifest
+  are added to the *mandatory_features* set,
+- the complement to *mandatory_features*, *ignored_features* comprised
+  of those features mandated by the manifest, but not supported by the client.
+
+Because the client previously checked compliance with the mandatory set of features
+required for the particular operation it is executing, the *mandatory_features* set
+will contain all necessary features required for using the cache safely.
+
+Conversely, the *ignored_features* set contains only those features which were not
+relevant to operating the cache. Otherwise, the client would not pass the feature
+set test against the manifest.
+
+When opening a cache and the *mandatory_features* set is a superset of the features
+supported by the client, the cache is wiped out and rebuilt.
+Since a client not supporting a mandatory feature that the cache was built with
+would be unable to update it correctly.
+
+When opening a cache and the intersection of *ignored_features* and the features
+supported by the client contains any elements, i.e. the client possesses features
+that the previous client did not have, the cache is wiped out and rebuilt.
+
 .. rubric:: Defined feature flags
 
 Currently no feature flags are defined.
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -310,7 +310,7 @@ or modified. It looks like this:
         b'item_keys': [b'acl_access', b'acl_default', ...],
         b'config': {},
         b'archives': {
-            b'archive name': {
+            b'2017-05-05-system-backup': {
                 b'id': b'<32 byte binary object ID>',
                 b'time': b'2017-05-05T12:42:22.942864',
             },
@@ -392,6 +392,10 @@ These can be locked out with manifest version 2. Thus, the only difference betwe
 manifest versions 1 and 2 is that the latter is only accepted by Borg releases
 implementing feature flags.
 
+Therefore, as soon as any mandatory feature flag is enabled in a repository,
+the manifest version must be switched to version 2 in order to lock out all
+Borg releases unaware of feature flags.
+
 .. rubric:: Cache feature flags
 
 `The cache`_ does not have its separate flag of feature flags. Instead, Borg stores
@@ -414,13 +418,18 @@ relevant to operating the cache. Otherwise, the client would not pass the featur
 set test against the manifest.
 
 When opening a cache and the *mandatory_features* set is a superset of the features
-supported by the client, the cache is wiped out and rebuilt.
-Since a client not supporting a mandatory feature that the cache was built with
+supported by the client, the cache is wiped out and rebuilt,
+since a client not supporting a mandatory feature that the cache was built with
 would be unable to update it correctly.
 
 When opening a cache and the intersection of *ignored_features* and the features
 supported by the client contains any elements, i.e. the client possesses features
-that the previous client did not have, the cache is wiped out and rebuilt.
+that the previous client did not have and those new features are enabled in the repository,
+the cache is wiped out and rebuilt.
+
+While the former condition likely requires no tweaks, the latter condition is formulated
+in an especially conservative way to play it safe. It seems likely that specific features
+might be exempted from the latter condition.
 
 .. rubric:: Defined feature flags
 
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -337,8 +337,6 @@ the manifest, since an ID check is not possible.
 of Borg preserve its contents (it may have been a better place for *item_keys*,
 which is not preserved by unaware Borg versions, releases predating 1.0.4).
 
-.. This was implemented in PR#1149, 78121a8 and a7b5165
-
 Feature flags
 +++++++++++++
 
@@ -355,7 +353,7 @@ category are operations that require accurate reference counts, for example
 archive deletion and check.
 
 As the manifest is always updated and always read, it is the ideal place to store
-feature flags, comparable to the super-block of a file system. The only issue problem
+feature flags, comparable to the super-block of a file system. The only problem
 is to recover from a lost manifest, i.e. how is it possible to detect which feature
 flags are enabled, if there is no manifest to tell. This issue is left open at this time,
 but is not expected to be a major hurdle; it doesn't have to be handled efficiently, it just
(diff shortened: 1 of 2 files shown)


docs/internals/data-structures.rst
```diff
@@ -253,49 +278,196 @@ If the quota shall be enforced accurately in these cases, either
...
+The object graph
+----------------
+
+On top of the simple key-value store offered by the Repository_,
+Borg builds a much more sophisticated data structure that is essentially
+a completely encrypted object graph. Objects, such as archives_, are referenced
+by their chunk ID, which is cryptographically derived from their contents.
+More on how this helps security in :ref:`security_structural_auth`.
+
+.. figure:: object-graph.png
```
ThomasWaldmann: can you move the "legend" (that explains the pointer being a reference using a chunk id) a bit more to the left so it does not look like pointing at the manifest?

docs/internals/data-structures.rst
```diff
@@ -253,49 +278,196 @@ If the quota shall be enforced accurately in these cases, either
...
-* timestamp
-* config
+The *tam* key is part of the :ref:`tertiary authentication mechanism <tam_description>`
+(formerly known as "tertiary authentication for metadata") and authenticates
+the manifest, since an ID check is not possible.
 
-Each archive info contains:
+*config* is a general-purpose location for additional metadata. All versions
+of Borg preserve its contents (it may have been a better place for *item_keys*,
+which is not preserved by unaware Borg versions).
```
ThomasWaldmann: open a ticket about this? 2.0?

docs/internals/data-structures.rst
```diff
@@ -253,49 +278,196 @@ If the quota shall be enforced accurately in these cases, either
...
 
-* name
-* id
-* time
+.. rubric:: Feature flags
```
ThomasWaldmann: hmm, this is documenting (in master branch) not-yet-merged stuff (that has a PR against 1.0).

separate PR for this?

docs/internals/data-structures.rst
```diff
@@ -384,13 +556,13 @@ For some more general usage hints see also ``--chunker-params``.
...
-Indexes / Caches
-----------------
+The cache
+---------
 
 The **files cache** is stored in ``cache/files`` and is used at backup time to
 quickly determine whether a given file is unchanged and we have all its chunks.
 
-The files cache is a key -> value mapping and contains:
+The files cache is in memory a key -> value mapping (a Python *dict*) and contains:
```
ThomasWaldmann: In memory, the...

docs/internals/data-structures.rst
```diff
@@ -556,6 +693,35 @@ b) with ``create --chunker-params 19,23,21,4095`` (default):
    You'll save some memory, but it will need to read / chunk all the files as
    it can not skip unmodified files then.
 
+HashIndex
+---------
+
+The chunks cache and the repository index are stored as hash tables, with
+only one slot per bucket, spreading hash collisions to the following
+buckets. As a consequence the hash is just a start position for a linear
+search, and if the element is not in the table the index is linearly crossed
```
ThomasWaldmann: "crossed" sounds a bit like this is completely going over all buckets. rather "searched"?

enkore: "until an empty bucket is found" (this text is older)

But yeah, maybe search is more apt here.

enkore:
I moved some things around, since I thought that cache flags were separate; you can't have a section about cache features with all the repo, index and cache stuff in one section. But that actually wasn't the case, see last commit.

docs/internals/data-structures.rst
```diff
@@ -339,10 +339,31 @@ the manifest, since an ID check is not possible.
...
+
+As the manifest is always updated and always read, it is the ideal place to store
+feature flags, comparable to the super-block of a file system. The only issue problem
+is to recover from a lost manifest, i.e. how is it possible to detect which feature
+flags are enabled, if there is no manifest to tell. This issue is left open at this time,
+but is not expected to be a major hurdle; it doesn't have to be handled efficiently, it just
+needs to be handled.
+
+Lastly, cache_ invalidation is handled by noting which feature
+flags were and were not used to manipulate a cache. This allows to detect whether
```
ThomasWaldmann: just "and" reads strange here. "and also which" maybe?

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,180 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+        b'tam': ...,
+    }
+
+The *version* field can be either 1 or 2. The versions differ in the
+way feature flags are handled, described below.
+
+The *timestamp* field was used to avoid a certain class of replay attack.
+It is still used for that purpose, however, the newer replay protection
+introduced in Borg 1.1 includes all reply attacks. Thus it is not strictly
+necessary any more.
```
textshell: hmm? I seems i'm missing something in my mental picture here

docs/internals/data-structures.rst
textshell: in my descriptions i used 'understood' instead of 'used'. I.e. a borg that understands feature X assumes that the cache is updated with that understanding of the repository. If that was not the case it might infer bad results. Not sure if that is a distinction worth trying to cram in here.

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,180 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+    }
+
+The top-level distinction for feature flags is the operation the client intends
+to perform,
+
+| the *read* operation includes extraction and listing of archives,
+| the *write* operation includes creating new archives,
+| the *delete* (archives) operation,
+| the *check* operation requires full understanding of everything in the repository.
+|
```
textshell: should we document current practice with regard to chunk cache updating?

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,180 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+
+These are weakly set-ordered; *check* will include everything required for *delete*,
+*delete* will likely include *write* and *read*. However, *read* may require more
+features than *write* (due to ID-based deduplication, *write* does not necessarily
+require reading/understanding repository contents).
```
textshell: I'm not 100% sure about these, but i think possible refinement can be done later.
As we don't try to document implementation requirements it doesn't really make sense to add that clients should not depend on ordering relations between these.

docs/internals/data-structures.rst
enkore: nonces manager & security dir. I should reference those directly, but they haven't been documented yet; I'll insert a todo.

enkore:
Thanks!

Seems I mixed up a bit of set theory; I'll straighten that out.

docs/internals/data-structures.rst
textshell: nonces manager only protects against nonce reuse (it does not alert when the nonce is going down, and the nonce value can stay even if everything else is rolled back because it is not encrypted). It does not protect against rollback of the logical repostory state. I think the timestamp is still the only protections against a server that just silently discards all new transactions. (of sometimes does a rollback to remove certain archives)

enkore: Yes, but the timestamp is both in the cache and in the security dir (as well as location, key type, nonce, tam_required).

textshell: yes, so in the *cache* it is no longer needed (superseded by the on in the security dir). In the manifest it is still needed to have something to compare the security directory saved timestamp against.

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,180 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+in all items are a subset of these keys. Thus, an older version of *borg check*
+supporting this mechanism can correctly detect keys introduced in later versions.
+
+The *tam* key is part of the :ref:`tertiary authentication mechanism <tam_description>`
+(formerly known as "tertiary authentication for metadata") and authenticates
+the manifest, since an ID check is not possible.
+
+*config* is a general-purpose location for additional metadata. All versions
+of Borg preserve its contents (it may have been a better place for *item_keys*,
+which is not preserved by unaware Borg versions).
```
enkore: I think it was 1.0.4 "better validation of item metadata dicts, #1130". I'll add that reference.

docs/internals/data-structures.rst
enkore: Ah right!

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+                b'time': b'2017-05-05T12:42:22.942864',
+            },
+        },
+        b'tam': ...,
+    }
+
+The *version* field can be either 1 or 2. The versions differ in the
+way feature flags are handled, described below.
+
+The *timestamp* field was used to avoid logical replay attacks where
```
textshell: s/was used/is used/ ?

enkore: indeed

enkore:
Bookmark of interest regarding "words to coloured pictures, the set theoreticians edition": https://commons.wikimedia.org/wiki/File:Venn0111.svg#Operations_and_relations_in_set_theory_and_logic

docs/internals/data-structures.rst
enkore: I'm not sure what exactly you mean? The cache sync?

Regarding the cache it crossed my mind that what you implemented and what I described below

>Therefore, as soon as any mandatory feature flag is enabled in a repository,
>the manifest version must be switched to version 2 in order to lock out all
>Borg releases unaware of feature flags.

must also be done for the cache config.

textshell: Cache version mismatches are currently fatal. Which if 'safe'. Not sure we want to change this to instead wipe at this time.

What i was thinking about was if we should explicitly document what does or does not prevents cache updating (as this is in a sense a different operation that precedes every write/delete operation.

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+These are weakly set-ordered; *check* will include everything required for *delete*,
+*delete* will likely include *write* and *read*. However, *read* may require more
+features than *write* (due to ID-based deduplication, *write* does not necessarily
+require reading/understanding repository contents).
+
+Each operation can contain several sets of feature flags. Only one set,
+the *mandatory* set is currently defined.
+
+Upon reading the manifest, the Borg client has already determined which operation
+is to be performed. If feature flags are found in the manifest, the set
```
ThomasWaldmann: "is to be performed" sounds strange, is it correct grammar?

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+of feature flags supported by the client is compared to the mandatory set
+found in the manifest. If any unsupported flags are found (i.e. the mandatory set is
+not a subset of the features supported by the Borg client used), the operation
+is aborted with a *MandatoryFeatureUnsupported* error:
+
+    Unsupported repository feature(s) {'some_feature'}. A newer version of borg is required to access this repository.
+
+Older Borg releases do not have this concept and do not perform feature flags checks.
+These can be locked out with manifest version 2. Thus, the only difference between
```
ThomasWaldmann: s/can be locked out with/are locked out by/?

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+implementing feature flags.
+
+Therefore, as soon as any mandatory feature flag is enabled in a repository,
+the manifest version must be switched to version 2 in order to lock out all
+Borg releases unaware of feature flags.
+
+.. _Cache feature flags:
+.. rubric:: Cache feature flags
+
+`The cache`_ does not have its separate flag of feature flags. Instead, Borg stores
```
ThomasWaldmann: flag of feature flags? did you mean "set"?

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+will contain all necessary features required for using the cache safely.
+
+Conversely, the *ignored_features* set contains only those features which were not
+relevant to operating the cache. Otherwise, the client would not pass the feature
+set test against the manifest.
+
+When opening a cache and the *mandatory_features* set is a not a subset of the features
+supported by the client, the cache is wiped out and rebuilt,
+since a client not supporting a mandatory feature that the cache was built with
+would be unable to update it correctly.
```
ThomasWaldmann: well, this assumes that the unsupported feature would somehow reflect in the cache contents.
but as we can't know that without additional information, it is safer to assume that it does.

docs/internals/data-structures.rst
```diff
@@ -439,7 +440,7 @@ Conversely, the *ignored_features* set contains only those features which were n
 relevant to operating the cache. Otherwise, the client would not pass the feature
 set test against the manifest.
 
-When opening a cache and the *mandatory_features* set is a superset of the features
+When opening a cache and the *mandatory_features* set is a not a subset of the features
```
enkore: a not

docs/internals/data-structures.rst
```diff
@@ -294,22 +294,181 @@ More on how this helps security in :ref:`security_structural_auth`.
...
+chunks that are not readable with the current feature set. The third
+category are operations that require accurate reference counts, for example
+archive deletion and check.
+
+As the manifest is always updated and always read, it is the ideal place to store
+feature flags, comparable to the super-block of a file system. The only issue problem
+is to recover from a lost manifest, i.e. how is it possible to detect which feature
+flags are enabled, if there is no manifest to tell. This issue is left open at this time,
+but is not expected to be a major hurdle; it doesn't have to be handled efficiently, it just
+needs to be handled.
```
enkore: - The manifest is small so the overall corruption likelihood is low
- Repository is transacted
- Most features are likely easy to detect
- Enabled flags are in all caches

docs/internals/data-structures.rst
enkore: It is, but *which operation will be performed.* is easier on the eyes.

docs/internals/data-structures.rst
textshell: One solution would be to just add all mandatory flags when recreating an manifest.
That would mean, that only a version with the same or better feature set can use the recovered repository, but that sounds like an ok tradeoff when such an operation is needed.

But i think for most features more precise detection should be possible. But we can always fall back to the just enable everything plan if that seems to be hard or fragile.

docs/internals/data-structures.rst
textshell: Strictly speaking we might use feature flags without bumping the manifest version because we would like to keep compatibility and the harm is not too bad. Currently no code sets manifest version to 2.

ThomasWaldmann: ah, ok.

docs/internals/data-structures.rst
enkore: I think that's an implementation defect, since the cache sync doesn't really obey the correct operations. #2659

PR #2604, opened 2017-06-04
docs/internals: layers image + description

enkore:
Not entirely sure if it wouldn't be better to omit the lowest row.


docs/internals.rst
```diff
@@ -26,6 +26,16 @@ To actually perform the repository-wide deduplication, a hash of each
...
+.. figure:: internals/structure.png
+
+    Layers in Borg. On the very top commands are implemented, using
+    a data access layer provided by the Archive and Item classes.
+    The "key" object provides both compression and authenticated
+    encryption used by the data access layer. The "key" object represents
+    the sole trust boundary in Borg.
+    The lowest layer is the repository, either accessed directly
+    (Repository) or remotely (RemoteRepository).
+
```
ThomasWaldmann: a bit strange in the pic is that it shows segments/hints/index/integrity, but not the clientside cache.

PR #2606, opened 2017-06-04
docs: internals editing

enkore:
Split from #2602


docs/internals/data-structures.rst
```diff
@@ -556,6 +563,46 @@ b) with ``create --chunker-params 19,23,21,4095`` (default):
...
+for an element not in the index stops at the first empty slot), the hash table
+is rebuilt. The maximum *effective* load factor, i.e. including tombstones, is 93%.
+
+Data in a HashIndex is always stored in little-endian format, which increases
+efficiency for almost everyone, since basically no one uses big-endian processors
+any more.
+
+HashIndex does not use a hashing function, because all keys (save manifest) are
+outputs of a cryptographic hash or MAC and thus already have excellent distribution.
+Thus, HashIndex simply uses the first 32 bits of the key as its "hash".
```
enkore: So if I can make many chunks where `the first 32 bits % N` collide a lot I can DoS HashIndex, yeah?

docs/internals/data-structures.rst
```diff
@@ -278,24 +314,32 @@ each time an archive is added, modified or deleted.
...
+
+Instead, it contains a list of chunks which form a msgpacked stream of items_.
+The archive object itself further contains some metadata:
+
+* *version*
+* *name*, which might differ from the name set in the manifest.
+  When :ref:`borg_check` rebuilds the manifest (e.g. if it was corrupted) and finds
+  more than one archive object with the same name, it adds a counter to the name
+  in the manifest, but leaves the *name* field of the archives as it was.
+* *items*, a list of chunk IDs containing item metadata (size: count * ~33B)
```
enkore: Previously it said 40B, but that's actually not correct, because Archive.items is not a ChunkListEntry (id, size, csize), but only the IDs, which are raw8 in msgpack = 1 prefix byte + 1 length byte + 32 bytes = 34B.

PR #2618, opened 2017-06-05
move hashindex to borg.algorithms, fixes #2586

### LINKED ISSUE #2586: move hashindex to algorithms/
it's an own implementation of a well-known algorithm/datastructure, so guess it belongs there.

ThomasWaldmann:
... and also lrucache


ThomasWaldmann:
borg.lrucache -> borg.algorithms.lrucache ?

ThomasWaldmann:
merge?

ThomasWaldmann:
did some minor changes: update .gitignore, rebase, avoid whitespace changes.

will merge after travis is happy.

enkore:
Like I wrote you per email to your question regarding this particular package, and to the question why I did not move hashindex there as well, I explained that I made it to contain standard checksums and hashes and the like (that an application should, ideally, not need to ship on its own). Thus it contains "_functions that calculate a number from a range of bytes_". Whether HashIndex is in good company there, I'll leave up to you.

ThomasWaldmann:
@enkore well, the package name is "algorithms".

A lru cache and a hashtable are well-known algorithms.

What you say rather indicates that you wanted a different package name, but "algorithms" is what we have and I think lrucache and hashindex fits well into there.

Also, you moved the chunker into "algorithms" package also and it does not fit your narrow definition either.

enkore:
Actually, they are data structures.

Perhaps the taxonomically correct approach to this organizational conundrum is to move chunker back to the main package.

ThomasWaldmann:
Yes, and data structures are implemented / operated by algorithms.

You can also move the chunker, I'm fine with that, too.
But in that case please add the narrow definition to algorithms.__doc__.

enkore:
Superseded by #2633

PR #2622, opened 2017-06-06
repository: truncate segments before unlinking

### LINKED ISSUE #2557: Make safe roll-back work on ZFS/Btrfs
In _repository.py_ the code to roll back a transaction because there is no space left on the file system looks like this:

>     def cleanup(self, transaction_id):
>         """Delete segment files left by aborted transactions
>         """
>         self.segment = transaction_id + 1
>         for segment, filename in self.segment_iterator(reverse=True):
>             if segment > transaction_id:
>                 os.unlink(filename)
>             else:
>                 break

Unfortunately this does not work well on 'modern' CoW file systems such as ZFS or Btrfs. If these file systems are full, even a delete as root is not possible any more because of how CoW works. This is a well known problem:

- http://www.surlyjake.com/blog/2010/01/19/zfs-cant-rm-no-space-left-on-device/
- https://serverfault.com/questions/478733/rm-can-not-remove-xxx-no-space-left-on-device-on-btrfs

The trick is to free up space by 'emptying' a file. The above function should therefore do a

>    f = open(filename, "rw")
>    f.truncate();

before the os.unlink(). I find a total of 20 os.unlink() calls in the whole source. To better support the two above file systems I suggest to re-visit each of them, to check if a no space situation is possible and add the truncate code.

https://www.bountysource.com/issues/45481976-make-safe-roll-back-work-on-zfs-btrfs

enkore:
Fixes #2557 

I'd just make similar changes to chunks.archive.d writeout, but that's still locked.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -83,6 +83,30 @@ class Repository:
     dir/data/<X // SEGMENTS_PER_DIR>/<X>
     dir/index.X
     dir/hints.X
+
+    File system interaction
+    -----------------------
+
+    LoggedIO generally tries to rely on common behaviours across transactional file systems.
+
+    Segments that are deleted are truncated first, which avoids problems if the FS needs to
+    allocate space to delete the dirent of the segment. This mostly affects CoW file systems,
+    traditional journaling file systems have a fairly good grip on this problem.
+
+    Note that deletion, i.e. unlink(2), is atomic on every file system that uses inode reference
+    counts, which includes pretty much all of them. To remove a dirent the inodes refcount has
+    to be decreased, but you can't decrease the refcount before removing the dirent nor can you
+    decrease the refcount after removing the dirent. File systems solve this with a lock,
+    and by ensuring it all stays within the same FS transaction.
+
+    Truncation is generally not atomic in itself, and combining truncate(2) and unlink(2) is of
+    course never guaranteed to be atomic. Truncation in a classic extent-based FS is done in
+    roughly two phases, first the extents are removed then the inode is updated. (In practice
+    this is of course way more complex).
+
+    LoggedIO gracefully handles truncate/unlink splits as long as the truncate resulted in
+    a zero length file. Zero length segments are considered to not exist, while LoggedIO.cleanup()
+    will still get rid of them.
     """
 
     class DoesNotExist(Error):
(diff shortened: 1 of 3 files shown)


src/borg/repository.py
```diff
@@ -1207,12 +1236,22 @@ def delete_segment(self, segment):
...
         try:
-            os.unlink(self.segment_filename(segment))
+            filename = self.segment_filename(segment)
+            # Truncate segment files before unlink(). This can help a full file system recover.
+            # In this instance (cf. cleanup()) we need to use r+b (=O_RDWR|O_BINARY) and
+            # issue an explicit truncate() to avoid creating a file
+            # if *segment* did not exist in the first place.
+            with open(filename, 'r+b') as fd:
+                fd.truncate()
+            os.unlink(filename)
```
ThomasWaldmann: if you will need these 2 kinds of deletion code at other places also, it would be nice to have them in a helper function.

maybe they could also be merged into 1 function by providing a sure_exists=False flag additionally to dispatch to the one or the other.

src/borg/repository.py
```diff
@@ -1258,6 +1297,7 @@ def recover_segment(self, segment, filename):
         if segment in self.fds:
             del self.fds[segment]
         with open(filename, 'rb') as fd:
+            # XXX: Rather use mmap.
```
ThomasWaldmann: add why.

src/borg/repository.py
```diff
@@ -1132,7 +1158,8 @@ def cleanup(self, transaction_id):
         self.segment = transaction_id + 1
         for segment, filename in self.segment_iterator(reverse=True):
             if segment > transaction_id:
-                os.unlink(filename)
+                # Truncate segment files before unlink(). This can help a full file system recover.
```
ThomasWaldmann: move comment into function?

src/borg/repository.py
```diff
@@ -1207,12 +1234,20 @@ def delete_segment(self, segment):
         if segment in self.fds:
             del self.fds[segment]
         try:
-            os.unlink(self.segment_filename(segment))
+            filename = self.segment_filename(segment)
+            # Truncate segment files before unlink(). This can help a full file system recover.
+            # In this instance (cf. cleanup()) we need to use r+b (=O_RDWR|O_BINARY) and
+            # issue an explicit truncate() to avoid creating a file
+            # if *segment* did not exist in the first place.
```
ThomasWaldmann: move parts of this comment into function?

PR #2632, opened 2017-06-07
nanorst for --help

enkore:
Strips a small subset of rST syntax from --help output and replaces it with standard vt100 escapes when on a terminal.

This solves the docs-formatting dichotomy between Sphinx and --help.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -64,6 +64,7 @@ from .helpers import basic_json_data, json_print
 from .helpers import replace_placeholders
 from .helpers import ChunkIteratorFileWrapper
 from .helpers import popen_with_error_handling
+from .nanorst import RstToTextLazy, ansi_escapes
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
does that work for:

- windows / windows+cygwin
- other platforms we support
- borg command --help | more  # or less

src/borg/archiver.py
```diff
@@ -2256,6 +2257,18 @@ def resolve(self, args: argparse.Namespace):  # Namespace has "in" but otherwise
                 setattr(args, dest, option_value)
 
     def build_parser(self):
+        if hasattr(sys.stdout, 'isatty') and sys.stdout.isatty() and (sys.platform != 'win32' or 'ANSICON' in os.environ):
```
ThomasWaldmann: ah, ok.

PR #2634, opened 2017-06-08
enable remote tests on cygwin

ThomasWaldmann:
the cygwin issue that caused these tests to break was fixed in cygwin
at least since cygwin 2.8.0 (maybe even since 2.7.0).

also added a comment to our workaround (os_write wrapper, that is needed
still for people running older cygwin versions) that it can be removed
when cygwin 2.8.0 is considered ancient (and everybody has upgraded to
some fixed version).

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -42,6 +42,8 @@ RATELIMIT_PERIOD = 0.1
 
 def os_write(fd, data):
     """os.write wrapper so we do not lose data for partial writes."""
+    # TODO: this issue is fixed in cygwin since at least 2.8.0, remove this
+    #       wrapper / workaround when this version is considered ancient.
     # This is happening frequently on cygwin due to its small pipe buffer size of only 64kiB
     # and also due to its different blocking pipe behaviour compared to Linux/*BSD.
     # Neither Linux nor *BSD ever do partial writes on blocking pipes, unless interrupted by a
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -2835,7 +2835,6 @@ class ManifestAuthenticationTest(ArchiverTestCaseBase):
         assert not self.cmd('list', self.repository_location)
 
 
-@pytest.mark.skipif(sys.platform == 'cygwin', reason='remote is broken on cygwin and hangs')
 class RemoteArchiverTestCase(ArchiverTestCase):
     prefix = '__testsuite__:'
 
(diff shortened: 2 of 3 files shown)


ThomasWaldmann:
Note: changeset works ok on 1.0-maint, but there are lots of test failures on master (on cygwin).

enkore:
Are the endian-related tests passing?

enkore:
Can't debug, the windows10 box is broken (for me).

```
==> windows10: The machine you're rsyncing folders to is configured to use
==> windows10: password-based authentication. Vagrant can't script rsync to automatically
==> windows10: enter this password, so you'll likely be prompted for a password
==> windows10: shortly.
==> windows10: 
==> windows10: If you don't want to have to do this, please enable automatic
==> windows10: key insertion using `config.ssh.insert_key`.
==> windows10: Rsyncing folder: /home/foo/_oss/borg/ => /vagrant/borg/borg
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mkdir '/vagrant/borg/borg'

Stdout from the command:



Stderr from the command:

mkdir: cannot create directory '/vagrant/borg/borg': No such file or directory
```

ThomasWaldmann:
@enkore there's a ticket about the broken win10 vagrant machine. I used the kvm based machine I also used before someone made the win10 vagrant VM.

What's the test name of the "endian test" you mean?

enkore:
HashIndexDataTestCase

ThomasWaldmann:
HashIndexDataTestCase.test_identical_creation fails with MemoryError when calling hashindex_write.

testsuite/checksums.py works ok.

enkore:
hashindex_write doesn't allocate memory through Python apart from the usual number of smaller Python objects. Try some EPRINTF()s to see where the actual issue is in there.

enkore:
Try replacing (int) with (Py_ssize_t) in this line:

```
length_object = PyObject_CallMethod(file_py, "write", "y#", &header, (int)sizeof(HashHeader));
```

enkore:
Regarding FD exception,
clear the exceptfd set and see if and what error is actually communicated through read/write. Like I mentioned in the other ticket, exceptfd should be universally meaningless in this use case. Some systems may additionally report errors through the exceptfd set, but this is neither required nor portable.

ThomasWaldmann:
@enkore any idea about this one?

https://paste.thinkmo.de/EavhUrkb

ThomasWaldmann:
ok, I am merging this one so the remaining bugs (on cygwin) are better visible.

PR #2636, opened 2017-06-08
hashindex: Cython defines PY_SSIZE_T_CLEAN

enkore:
#2634


ThomasWaldmann:
i pulled current master from the cygwin test setup. from endless failures now down to 3.

so, looks like this fixed it.

PR #2639, opened 2017-06-09
docs: highlight experimental features in online docs

enkore:


docs/conf.py
```diff
@@ -74,6 +74,10 @@
 # The reST default role (used for this markup: `text`) to use for all documents.
 #default_role = None
 
+# The Borg doc contain no or very little Python docs.
+# Thus, the primary domain is rst.
+primary_domain = 'rst'
```
enkore: (This allows writing `.. class::`..., for example. Otherwise one would have to write `.. rst-class::`...)

docs/usage/general.rst
```diff
@@ -5,6 +5,13 @@ Borg consists of a number of commands. Each command accepts
 a number of arguments and options and interprets various environment variables.
 The following sections will describe each command in detail.
 
+.. container:: experimental
+
+   Experimental features are marked with red-orange stripes on the sides, like this paragraph.
```
ThomasWaldmann: well, we can do that, but I guess we'll also have `(experimental)` or so as textual marker?

enkore: Where this patch applies, it already is in the text. The visual highlight just makes it more obvious and helps if an experimental feature and a stable feature are described next to each other. I'm sure we'll see that soon.

PR #2642, opened 2017-06-09
docs: deployment: Automated backups to a local hard drive

enkore:
This is based on my gist from early 2016: https://gist.github.com/enkore/a977f21f6d94f3db4891

### DIFF
--- a/docs/deployment/automated-local.rst
+++ b/docs/deployment/automated-local.rst
@@ -1,0 +1,231 @@
+.. include:: ../global.rst.inc
+.. highlight:: none
+
+Automated backups to a local hard drive
+=======================================
+
+This guide shows how to automate backups to a hard drive directly connected
+to your computer. If a backup hard drive is connected, backups are automatically
+started, and the drive shut-down and disconnected when they are done.
+
+This guide is written for a Linux-based operating system and makes use of
+systemd and udev.
+
+Overview
+--------
+
+An udev rule is created to trigger on the addition of block devices. The rule contains a tag
+that triggers systemd to start a oneshot service. The oneshot service executes a script in
+the standard systemd service environment, which automatically captures stdout/stderr and
+logs it to the journal.
+
+The script mounts the added block device, if it is a registered backup drive, and creates
+backups on it. When done, it optionally unmounts the file system and spins the drive down,
+so that it may be physically disconnected.
+
+Configuring the system
+----------------------
+
+First, create the ``/etc/backups`` directory (as root).
+All configuration goes into this directory.
+
+Then, create ``etc/backups/40-backup.rules`` with the following content (all on one line)::
+
+    ACTION=="add", SUBSYSTEM=="bdi", DEVPATH=="/devices/virtual/bdi/*",
+    TAG+="systemd", ENV{SYSTEMD_WANTS}="automatic-backup.service"
+
+.. topic:: Finding a more precise udev rule
+
+    If you always connect the drive(s) to the same physical hardware path, e.g. the same
+    eSATA port, then you can make a more precise udev rule.
+
+    Execute ``udevadm monitor`` and connect a drive to the port you intend to use.
+    You should see a flurry of events, find those regarding the `block` subsystem.
+    Pick the event whose device path ends in something similar to a device file name,
+    typically`sdX/sdXY`. Use the event's device path and replace `sdX/sdXY` after the
+    `/block/` part in the path with a star (\*). For example:
+    `DEVPATH=="/devices/pci0000:00/0000:00:11.0/ata3/host2/target2:0:0/2:0:0:0/block/*"`.
+
+    Reboot a few times to ensure that the hardware path does not change: on some motherboards
+    components of it can be random. In these cases you cannot use a more accurate rule,
+    or need to insert additional stars for matching the path.
+
+The "systemd" tag in conjunction with the SYSTEMD_WANTS environment variable has systemd
+launch the "automatic-backup" service, which we will create next, as the
+``/etc/backups/automatic-backup.service`` file:
+
+.. code-block:: ini
+
+    [Service]
+    Type=oneshot
+    ExecStart=/etc/backups/run.sh
+
+Now, create the main backup script, ``/etc/backups/run.sh``. Below is a template,
+modify it to suit your needs (e.g. more backup sets, dumping databases etc.).
+
+.. code-block:: bash
+
+    #!/bin/bash -ue
+
+    # The udev rule is not terribly accurate and may trigger our service before
+    # the kernel has finished probing partitions. Sleep for a bit to ensure
+    # the kernel is done.
+    #
+    # This can be avoided by using a more precise udev rule, e.g. matching
+    # a specific hardware path and partition.
+    sleep 5
+
+    #
+    # Script configuration
+    #
+
+    # The backup partition is mounted there
+    MOUNTPOINT=/mnt/backup
+
+    # This is the location of the Borg repository
+    TARGET=$MOUNTPOINT/borg-backups/backup.borg
+
+    # Archive name schema
+    DATE=$(date --iso-8601)-$(hostname)
+
+    # This is the file that will later contain UUIDs of registered backup drives
+    DISKS=/etc/backups/backup.disks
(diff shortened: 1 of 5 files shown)


docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,210 @@
...
+	# The backup partition is mounted there
+	MOUNTPOINT=/mnt/backup
+
+	# This is the location of the Borg repository
+	TARGET=$MOUNTPOINT/borg-backups/backup.borg
+
+	# Archive name schema.
+	# If you intend to use prune and have more than one backup set, perhaps
+	# swap hostname and date to be compliant with prune's --prefix option.
+	DATE=$(date --iso-8601)-$(hostname)
```
enkore: No. Date first is easier to browse, and with the capacities of hard drives many won't have reason to use prune. I have used prune **not even once** on my local backups.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+	fi
+
+	echo "Disk $uuid is a backup disk"
+	disk_path=/dev/disk/by-uuid/$uuid
+	# Mount file system if not already done. This assumes that if something is already
+	# mounted at $MOUNTPOINT, it is the backup drive. It won't find the drive if
+	# it was mounted somewhere else.
+	(mount | grep $MOUNTPOINT) || mount $disk_path $MOUNTPOINT
+	drive=$(lsblk -snlpo name $disk_path | head -n1)
+	echo "Drive path: $drive"
```
enkore: Oh, GitHub actually understands enough rST to highlight `.. code-block::` correctly. Neat.

docs/deployment/automated-local.rst
textshell: I think we should recommend naming that is prune compatible. Maybe later the user wants to prune and then it really is annoying to even remove the archives no longer wanted. Later the script adds more stuff to DATE that should also go before the date.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+	  --exclude /root/.cache \
+	  --exclude /var/cache \
+	  --exclude /var/lib/docker/devicemapper \
+	  $TARGET::$DATE-$$-system \
+	  / /boot
+
+	borg create $BORG_OPTS \
+	  --exclude 'sh:/home/*/.cache' \
+	  $TARGET::$DATE-$$-system \
+	  /home/
```
textshell: if this backups /home the previous one needs to exclude home explicitly and not depend on -x doing this.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+	(mount | grep $MOUNTPOINT) || mount $disk_path $MOUNTPOINT
+	drive=$(lsblk -snlpo name $disk_path | head -n1)
+	echo "Drive path: $drive"
+
+	#
+	# Create backups
+	#
+
+	# Options for borg create
+	BORG_OPTS="-svx --compression lz4 --checkpoint-interval 86400"
```
textshell: script should always use the long form of the options for readability.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+
+Create an empty ``/etc/backups/backup.disks`` file, you'll register your backup drives
+there.
+
+The last part is to actually enable the udev rules and services:
+
+.. code-block:: bash
+
+	ln -s /etc/backups/40-backup.rules /etc/udev/rules.d/40-backup.rules
+	ln -s /etc/backups/automatic-backup.service /etc/systemd/system/automatic-backup.service
```
textshell: I don't think linking from these directories to random other places is a good idea. It does keep things in one place though.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+Borg ensures that backups are not created on random drives that "just happen"
+to contain a Borg repository. If an unknown unencrypted repository is encountered,
+then the script aborts (BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK=no).
+
+So backups will only be created on hard drives that have the correct UUID and
+contain a Borg repository that is either known (by ID) to your machine or
+you are using encryption and the passphrases have to match.
+
+Note that generally speaking, physical access to any computer usually equals
+complete compromise of the machine.
```
textshell: This is increasingly untrue with secure boot, tpm and disk encryption. At least for less advanced attackers.

This does open additional attack surface because it mounts filesystems and scans metadata (of course this is not really avoidable and the uuid check reduces the likelihood without specific targeting). But which could cause parts that are not borg to be attacked (fs drivers and software scanning files on all mounted file systems)

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+    Type=oneshot
+    ExecStart=/etc/backups/run.sh
+
+Now, create the main backup script, ``/etc/backups/run.sh``. Below is a template,
+modify it to suit your needs (e.g. more backup sets, dumping databases etc.).
+
+.. code-block:: bash
+
+	#!/bin/bash -ue
+	which python borg > /dev/null
```
textshell: this seems to be designed as guard when needed software is not installed. But it silently terminates the script which is hard to debug.

Also checking for python is likely a bad idea. Because python3 is what borg needs. (I don't think any distribution is that crazy to ship a python3 implementation that is not callable with python3). Also borg might be the one file binary.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+
+	# The udev rule is not terribly accurate and may trigger our service before
+	# the kernel has finished probing partitions. Sleep for a bit to ensure
+	# the kernel is done.
+	#
+	# This can be avoided by using a more precise udev rule, e.g. matching
+	# a specific hardware path and partition.
+	sleep 5
```
textshell: in theory something like udev settle might be more correct. But i don't have experience with this.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+	# Archive name schema.
+	# If you intend to use prune and have more than one backup set, perhaps
+	# swap hostname and date to be compliant with prune's --prefix option.
+	DATE=$(date --iso-8601)-$(hostname)
+
+	# This is the file that will later contain UUIDs of registered backup drives
+	DISKS=/etc/backups/backup.disks
+
+	# Find whether the connected block device is a backup drive
+	for uuid in $(lsblk -nlpo uuid)
```
textshell: --paths does not make any sense here. Also long options in scripts.

docs/deployment/automated-local.rst
```diff
@@ -0,0 +1,227 @@
...
+	  /home/
+
+	echo "Completed backup for $DATE"
+
+	# Just to be completely paranoid
+	sync
+
+	if [ -f /etc/backups/autoeject ]; then
+	        umount $MOUNTPOINT
+	        hdparm -Y $drive
```
textshell: prefer `$disk_path` here?
It has less chance to be misparsed lsblk output.

docs/deployment/automated-local.rst
enkore: This is using udev.

textshell: Yes, what i was thinking about is part of udevadm.
```
udevadm settle [options]
   Watches the udev event queue, and exits if all current events are handled.
```
As i said, i have not actually tried if this works reliably for such use cases.

enkore: It doesn't.

docs/deployment/automated-local.rst
robaato: BTW. Using "which" is a bad idea. It is not a standard utility nor a builtin and it is not always available (at least several distros are missing it by default). See:
http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script/677212#677212

docs/deployment/automated-local.rst
enkore: Prune is now fixed.

PR #2643, opened 2017-06-09
mark --pattern, --patterns-from as experimental

enkore:


docs/usage/help.rst.inc
```diff
@@ -8,16 +8,18 @@ borg help patterns
...
+If followed by a colon (':') the first two characters of a pattern are used as a
 style selector. Explicit style selection is necessary when a
 non-default style is desired or when the desired pattern starts with
 two alphanumeric characters followed by a colon (i.e. `aa:something/*`).
 
 `Fnmatch <https://docs.python.org/3/library/fnmatch.html>`_, selector `fm:`
 
     This is the default style for --exclude and --exclude-from.
-    These patterns use a variant of shell pattern syntax, with '*' matching
+    These patterns use a variant of shell pattern syntax, with '\*' matching
```
ThomasWaldmann: hmm, why do we need that?

bold markup is `**doublestars**`, right?

enkore: `*italic*`. Interestingly enough the rST parser does not pick this one up; I suspect it has a rule that the smallest range has in-line markup applied, i.e. in-line markup seems to be not-greedy in rST. Note that rST does not allow nested in-line markup. The escape is correctly processed, of course.

(Keep in mind: nanorst doesn't have to work for arbitrary stuff, it just needs to work for the rst used in the help texts)

enkore: btw. I don't think rST is _per the spec_ something that can be parsed efficiently without bloppers / hacks.

enkore: btw. I'm surprised how much the little ANSI bits help with readability.

ThomasWaldmann:
tests are failing.

Issue #2647, opened 2017-06-10
pytest-xdist 1.17.0 breaks all testing


enkore:
https://travis-ci.org/borgbackup/borg/jobs/241229446 (master)
https://travis-ci.org/borgbackup/borg/jobs/241460695 (PR)

Reproduces locally, 1.15.0 works, though.

enkore:
This is https://github.com/pytest-dev/pytest-xdist/issues/157

textshell:
can/should we pin this somehow to a working version until this is fixed?

ThomasWaldmann:
https://github.com/pytest-dev/pytest-xdist/pull/158 this is already approved. guess we'll see a fix release soon.

we can pin to a working version until it is released.

enkore:
Pinning doesn't buy much, since it does not help topic branches (PRs), unless they're rebased. Looks like it's going to be resolved real soon now anyway.

nicoddemus:
Yep, I'm preparing the release right now. 👍

enkore:
Thank you!

nicoddemus:
`1.17.1` is now out, sorry guys for the inconvenience.

PR #2648, opened 2017-06-10
Add minimal version of in repository mandatory feature flags. (master)

textshell:
[This is the forward port of #2134
Changes relative to the version committed on 1.0-maint:
mapping of new commands to features:
`borg export-tar` and `borg diff` use Manifest.Operation.READ
`borg recreate` uses Manifest.Operation.CHECK
`borg debug dump-manifest` uses Manifest.NO_OPERATION_CHECK as all other debug commands.
]

This should allow us to make sure older borg versions can be cleanly
prevented from doing operations that are no longer safe because of
repository format evolution. This allows more fine grained control than
just incrementing the manifest version. So for example a change that
still allows new archives to be created but would corrupt the repository
when an old version tries to delete an archive or check the repository
would add the new feature to the check and delete set but leave it out
of the write set.

This is somewhat inspired by ext{2,3,4} which uses sets for
compat (everything except fsck), ro-compat (may only be accessed
read-only by older versions) and features (refuse all access).

This implements #1806

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,13 @@ import os
 
 import pytest
 
+# IMPORTANT keep this above all other borg imports to avoid inconsistent values
+# for `from borg.constants import PBKDF2_ITERATIONS` (or star import) usages before
+# this is executed
+from borg import constants
+# no fixture-based monkey-patching since star-imports are used for the constants module
+constants.PBKDF2_ITERATIONS = 1
+
 # needed to get pretty assertion failures in unit tests:
 if hasattr(pytest, 'register_assert_rewrite'):
     pytest.register_assert_rewrite('borg.testsuite')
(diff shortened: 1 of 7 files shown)


textshell:
for `borg recreate` i could have used (Manifest.Operation.READ, Manifest.Operation.WRITE).
But i didn't consider this used often enough to get a fine grained feature check and there is a expectation that recreate will not loose any information except as specified by command line options.

conftest.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+import borg.cache
```
enkore: This probably breaks `constants.PBKDF2_ITERATIONS = 1` below.

Suggestion:

```py
from borg import constants
constants.PBKDF2_ITERATIONS = 1

# other borg imports
```

enkore:
Travis is currently broken due to #2647

conftest.py
textshell: I'm not entirely sure why this was a `pytest_configure` before, but i'll add a commit moving that to the top without using `pytest_configure`

textshell: If we need to monkey patch this we really should consider boxing it into a mutable object.

For now i added warning comments to borg/__init__.py and conftest.py

enkore:
recreate is a tricky thing to get right and I fully agree with your rationale.

conftest.py
enkore: The reason I originally put it into the plugin function was that this way it's impossible to hit by accident (say, importing something from the conftest in test code, and maybe that is hit during self-test and maybe not immediately noticed, because py.test is installed). Call it ~~defensive~~ paranoid programming ;)

The resolution is ok, perhaps I'll send a patch later that restores the paranoia.

Issue #2458, opened 2017-04-28
Remove superfluous re: examples from docs


enkore:
```
# Backup home directories excluding image thumbnails (i.e. only
# /home/*/.thumbnails is excluded, not /home/*/*/.thumbnails)
$ borg create /path/to/repo::my-files /home \
    --exclude 're:^/home/[^/]+/\.thumbnails/'

# Do the same using a shell-style pattern
$ borg create /path/to/repo::my-files /home \
    --exclude 'sh:/home/*/.thumbnails'
```

Remove former.

```
# The contents of directories in '/home' are not backed up when their name
# ends in '.tmp'
$ borg create --exclude 're:^/home/[^/]+\.tmp/' backup /
```

Rewrite as sh: pattern.

PR #2649, opened 2017-06-10
Remove re docs

rugk:
Fixes https://github.com/borgbackup/borg/issues/2458

The `/home`/`tmp` example has already been removed, so now there is no `re:` example in there, anymore.


enkore:
+ 4cd1cc

Issue #2357, opened 2017-03-30
the new (c)size ticket


ThomasWaldmann:
scope of this ticket
===============
let's concentrate here on the issue of csize (and also size) information in the items' chunks lists, in the chunks and files cache. no crypto or other discussion in here, let's stay focussed.

csize
====
the main issue is that csize is not a direct function of the data, it also depends on compression and encryption (and other overhead) that is applied to the data. as both might change (and thus csize might change) while the chunk still contains the same (plaintext) content and has the same id, it is an annoyance to have csize in the chunks lists of archived items.

size
====
we must have chunk size information in the chunks lists of archived items for the case we lose multiple chunks in the repo - so we can replace them with all-zero chunks of same length. size is a direct function of the data, so no problem here if we change compression/encryption/overhead.

timing of size / csize computation
===========================
- size is computed early, after the chunker has cut the chunks: len(chunk)
- csize is computed late, after compression, after encryption/authentication. note: this can lead to a race (wait) condition in multithreaded processing.

where is chunk size/csize (not) stored?
================================
- repo: the current PUT entry in the segment file contains csize in the length information. no size available here! also, neither size nor csize is in the repo index.
- archive: item.chunks = [(id, size, csize), ...]
- chunks cache: id -> (refcount, size, csize)
- files cache: no size/csize here: path_hash -> (file_size, ino, mtime, chunks=[id, id, ...])

where is size/csize used?
=====================
- size dsize csize dcsize placeholders
- Statistics class + show_progress
- chunk_incref (gets size/csize from chunks cache - important for archiving unchanged files)
- csize: `Archive.info` -> limits -> max_archive_size and `Archive.__str__`
- csize: `Cache.__str__` .chunks_stored_size
- size: do_diff sum_chunk_sizes (to show sum of lengths of added/removed chunks of a file)
- size: borg check size consistency check item.size == sum(chunks size)
- tests

ThomasWaldmann:
Comment by @enkore, moved from #2313:

The main problem with csize is not so much compatibility problems or something like that, but this issue:

- Create an archive with some new chunks, they will have csize set
- Create another archive, but without the cache, so with csize=0 (which is unambiguous) for these chunks
- Delete the first archive
- Sync the cache
- Then the cache cannot know the csize of these chunks
- So we need to ask the repository, which means that has to do per-chunk I/O, because the csize isn't in the repo index, either (which makes a lot of sense)

[...]

PR #2654, opened 2017-06-10
create: --no-cache-sync

### LINKED ISSUE #2313: borg create without syncing cache

enkore:
Fixes #2313 

There are likely some rough edges or smaller inconsistencies when using this, e.g. borg-diff might not quite work...? I don't know.

I'll be upfront; _this PR is not about fixing all these. It's called experimental for a reason._ This PR is about getting the data model, the basics right.

Stuff for the future:

- Add a dedicated RPC call like "get_size[_many]" -- perhaps even taking a list (over RPC) to allow the remote `Repository` to sort by (segment, offset).
- recreate could touch up chunk id lists with csize=0 with their correct csize (depending on mode it'll already do that as a side-effect). check--repair could do that as well.

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -587,7 +587,7 @@ class LocalCache(CacheStatsMixin):
         archive indexes.
         """
         archive_path = os.path.join(self.path, 'chunks.archive.d')
-        # An index of chunks were the size had to be fetched
+        # An index of chunks whose size had to be fetched
         chunks_fetched_size_index = ChunkIndex()
         # Instrumentation
         processed_item_metadata_bytes = 0
@@ -965,7 +965,7 @@ Chunk index:    {0.total_unique_chunks:20d}             unknown"""
     do_files = False
 
     def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
-        pass
+        return None
 
     def memorize_file(self, path_hash, st, ids):
         pass
@@ -986,7 +986,15 @@ Chunk index:    {0.total_unique_chunks:20d}             unknown"""
         return ChunkListEntry(id, size, csize)
 
     def seen_chunk(self, id, size=None):
-        return self.chunks.get(id, ChunkIndexEntry(0, None, None)).refcount
+        if not self._txn_active:
+            self._begin_txn()
+        entry = self.chunks.get(id, ChunkIndexEntry(0, None, None))
+        if entry.refcount and size and not entry.size:
+            # The LocalCache has existing size information and uses *size* to make an effort at detecting collisions.
+            # This is of course not possible for the AdHocCache.
+            # Here *size* is used to update the chunk's size information, which will be zero for existing chunks.
+            self.chunks[id] = entry._replace(size=size)
+        return entry.refcount
 
     def chunk_incref(self, id, stats, size_=None):
         if not self._txn_active:
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -974,7 +974,7 @@ Chunk index:    {0.total_unique_chunks:20d}             unknown"""
         size = len(chunk)
         refcount = self.seen_chunk(id, size)
         if refcount:
-            return self.chunk_incref(id, stats)
+            return self.chunk_incref(id, stats, size_=size)
         data = self.key.encrypt(chunk)
         csize = len(data)
         self.repository.put(id, data, wait=wait)
@@ -985,12 +985,12 @@ Chunk index:    {0.total_unique_chunks:20d}             unknown"""
     def seen_chunk(self, id, size=None):
         return self.chunks.get(id, ChunkIndexEntry(0, None, None)).refcount
 
-    def chunk_incref(self, id, stats):
+    def chunk_incref(self, id, stats, size_=None):
         if not self._txn_active:
             self._begin_txn()
         count, size, csize = self.chunks.incref(id)
-        stats.update(size, csize, False)
-        return ChunkListEntry(id, size, csize)
+        stats.update(size or size_, csize, False)
+        return ChunkListEntry(id, size or size_, csize)
 
     def chunk_decref(self, id, stats, wait=True):
         if not self._txn_active:
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -644,14 +644,17 @@ class LocalCache(CacheStatsMixin):
 
             all_missing_ids = chunk_idx.zero_csize_ids()
             fetch_ids = []
-            for id_ in all_missing_ids:
-                already_fetched_entry = chunks_fetched_size_index.get(id_)
-                if already_fetched_entry:
-                    entry = chunk_idx[id_]._replace(csize=already_fetched_entry.csize)
-                    assert entry.size == already_fetched_entry.size, 'Chunk size mismatch'
-                    chunk_idx[id_] = entry
-                else:
-                    fetch_ids.append(id_)
+            if len(chunks_fetched_size_index):
+                for id_ in all_missing_ids:
+                    already_fetched_entry = chunks_fetched_size_index.get(id_)
+                    if already_fetched_entry:
+                        entry = chunk_idx[id_]._replace(csize=already_fetched_entry.csize)
+                        assert entry.size == already_fetched_entry.size, 'Chunk size mismatch'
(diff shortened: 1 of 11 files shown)


src/borg/cache.py
```diff
@@ -4,9 +4,11 @@
 import stat
 from binascii import unhexlify
 from collections import namedtuple
+from time import perf_counter
 
 import msgpack
 
+from borg.repository import LIST_SCAN_LIMIT
 from .logger import create_logger
```
enkore: abs import

src/borg/cache.py
```diff
@@ -588,6 +632,37 @@ def cleanup_cached_archive(id, cleanup_compact=True):
             except FileNotFoundError:
                 pass
 
+        def fetch_missing_csize(chunk_idx):
+            """
+            Archives create with AdHocCache will have csize=0 in all chunk list entries whose
```
enkore: created

src/borg/cache.py
```diff
@@ -547,10 +587,14 @@ def sync(self):
         archive indexes.
         """
         archive_path = os.path.join(self.path, 'chunks.archive.d')
+        # An index of chunks were the size had to be fetched
```
ThomasWaldmann: did you mean "An index of chunks for which the size had to be fetched"?

src/borg/cache.py
```diff
@@ -838,3 +919,132 @@ def memorize_file(self, path_hash, st, ids):
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+    files = None
+    do_files = False
+
+    def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
+        pass
```
ThomasWaldmann: return False?

ThomasWaldmann:
See also #2357.

ThomasWaldmann:
hmm, if there is no files cache, that means all backup data set files will have to be read, chunked, all chunks id-hashed and looked up, right?

this will have quite an impact for all bigger files (maybe not so much for tiny files where other processing overhead is bigger than the time needed to read/chunk/hash).

enkore:
It would be possible to introduce a new files cache format (`files.v2`?) that addresses it by including full ChunkListEntries, which won't be in this PR.

See https://github.com/borgbackup/borg/issues/2357#issuecomment-292162083

src/borg/cache.py
enkore: return None (equivalent but explicit form of `pass`).

src/borg/cache.py
```diff
@@ -838,3 +919,140 @@ def memorize_file(self, path_hash, st, ids):
...
+        self.repository.put(id, data, wait=wait)
+        self.chunks.add(id, 1, size, csize)
+        stats.update(size, csize, not refcount)
+        return ChunkListEntry(id, size, csize)
+
+    def seen_chunk(self, id, size=None):
+        if not self._txn_active:
+            self._begin_txn()
+        entry = self.chunks.get(id, ChunkIndexEntry(0, None, None))
+        if entry.refcount and size and not entry.size:
```
ThomasWaldmann: ... and size is not None and ...

src/borg/cache.py
```diff
@@ -838,3 +919,140 @@ def memorize_file(self, path_hash, st, ids):
...
+            # Here *size* is used to update the chunk's size information, which will be zero for existing chunks.
+            self.chunks[id] = entry._replace(size=size)
+        return entry.refcount
+
+    def chunk_incref(self, id, stats, size_=None):
+        if not self._txn_active:
+            self._begin_txn()
+        count, size, csize = self.chunks.incref(id)
+        stats.update(size or size_, csize, False)
+        return ChunkListEntry(id, size or size_, csize)
```
ThomasWaldmann: if size is 0 and size_ is None, "size or size_" will evaluate to None.

guess we would not like to have None where and int is required.

so:
```
size = size or _size or 0
```

enkore: That's intentional. If we try to create a ChunkListEntry with size=0, that's a logic bug. Don't do it.

src/borg/cache.py
enkore: size is never 0, the idiomatic way is thus quite alright.

src/borg/cache.py
ThomasWaldmann: ok, then just make it somehow self-explaining.

src/borg/cache.py
```diff
@@ -347,6 +349,69 @@ def destroy(repository, path=None):
             os.remove(config)  # kill config first
             shutil.rmtree(path)
 
+    def __new__(cls, repository, key, manifest, path=None, sync=True, do_files=False, warn_if_unencrypted=True,
+                progress=False, lock_wait=None, permit_adhoc_cache=False):
```
textshell: I would prefer a factory function to overloading __new__.

src/borg/cache.py
```diff
@@ -588,6 +632,37 @@ def cleanup_cached_archive(id, cleanup_compact=True):
...
+                        chunk_idx[id_] = entry
+                    else:
+                        fetch_ids.append(id_)
+            else:
+                fetch_ids = all_missing_ids
+
+            for id_, data in zip(fetch_ids, decrypted_repository.repository.get_many(fetch_ids)):
+                entry = chunk_idx[id_]._replace(csize=len(data))
+                chunk_idx[id_] = entry
+                chunks_fetched_size_index[id_] = entry
```
textshell: this is only needed for the self.do_cache case, isn't it? (not sure how expensive it is to have this duplicate index, but it feels like it could nearly double the memory needed)

textshell:
I think this should at least not memorize broken csize into the archive items. (i.e. Item.get_size should ignore memorize=True if any chunk has a 0 csize.

textshell:
I'm not really happy with something in a release creating archives with somewhat bogus metadata without a clear plan forward. This could mean that we are forced to carry code to resolve missing csize for a long time.
Also i think this really should not disable the files cache. Using the files cache or not should be an orthogonal choice. It feels it shouldn't be hard to make that possible.

src/borg/cache.py
enkore: One step at a time.

enkore:
I'd recommend you read #2313 and #2357.

enkore:
Anything left to do here?

ThomasWaldmann:
did you see my comment about "32" hardcoded?

enkore:
That's in a different patch.

ThomasWaldmann:
I'll have to look it through again.

But my impression was that the discussion with @textshell was not finished, was it?

enkore:
I wrote all the answers done months ago. I don't see anything left to discuss.

enkore:
Just for the sake of closure...

> I think this should at least not memorize broken csize into the archive items. (i.e. Item.get_size should ignore memorize=True if any chunk has a 0 csize.

That never happened. 36fcbd6

> This could mean that we are forced to carry code to resolve missing csize for a long time.

(1) Borg already generates inconsistent statistics (see recreate warnings) (2) it's literally just statistics. I do not see Borg carrying the extra code around to accommodate this, should it be dropped again.

>  Also i think this really should not disable the files cache. Using the files cache or not should be an orthogonal choice. It feels it shouldn't be hard to make that possible.

https://github.com/borgbackup/borg/issues/2313#issuecomment-300438677

src/borg/cache.py
```diff
@@ -588,6 +632,37 @@ def cleanup_cached_archive(id, cleanup_compact=True):
...
+                    if already_fetched_entry:
+                        entry = chunk_idx[id_]._replace(csize=already_fetched_entry.csize)
+                        assert entry.size == already_fetched_entry.size, 'Chunk size mismatch'
+                        chunk_idx[id_] = entry
+                    else:
+                        fetch_ids.append(id_)
+            else:
+                fetch_ids = all_missing_ids
+
+            for id_, data in zip(fetch_ids, decrypted_repository.repository.get_many(fetch_ids)):
```
ThomasWaldmann: add a comment here, that this could get rather expensive, for big fetch_ids lists.

and that the len info would be enough (new repo api call).

enkore:
Rebased

PR #2660, opened 2017-06-11
cache sync: check Operation.READ compatibility with manifest

### LINKED ISSUE #2659: Feature flags: cache does not obey READ operation
```py
@with_repository(fake='dry_run', exclusive=True, compatibility=(Manifest.Operation.WRITE,))
def do_create(self, args, repository, manifest=None, key=None):
```

But the cache may be synced, which needs Operation.READ, which is not checked.

The cache sync thus needs to perform another operation check with READ, independent of what the command opening the cache may have checked.

enkore:
Fixes #2659

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -721,6 +721,11 @@ Chunk index:    {0.total_unique_chunks:20d} {0.total_chunks:20d}"""
             except:
                 pass
 
+        # The cache can be used by a command that e.g. only checks against Manifest.Operation.WRITE,
+        # which does not have to include all flags from Manifest.Operation.READ.
+        # Since the sync will attempt to read archives, check compatibility with Manifest.Operation.READ.
+        self.manifest.check_repository_compatibility((Manifest.Operation.READ, ))
+
         self.begin_txn()
         with cache_if_remote(self.repository, decrypted_cache=self.key) as decrypted_repository:
             legacy_cleanup()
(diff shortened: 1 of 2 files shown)


textshell:
I think your test shows that is is way to hard an condition. Restricting cache deletion because the repository has unknown flags is certainly not needed for safe operation.

Also see my comment in #2659 why i don't think this is the right approach.

I would even go further and just let borg ignore all archives that are not parsable while in degraded (i.e. cache has some ignored features) mode.

enkore:
> Restricting cache deletion because the repository has unknown flags is certainly not needed for safe operation.

That's not in the patch.

textshell:
Sorry, i misread that patch. You are of course correct that delete of the cache is not affected (actually it is tested that it is not) but only used as a preparation step to force cache resync. Sorry about that.

textshell:
I think there are 2 possible approaches here as long as we don't have code that allows creating archives without cache resync:
* What this PR currently does: Block cache resync whenever READ contains unknown flags. That would entangle READ and WRITE and make the practically identical for now.
* Ignore archives that are not parsable. Using WRITE is likely to be done by a `borg create` which ensures that the chunks that the old client cares about are referenced from a archive readable by that older client and thus this only results in useless put chunk operations for new chunks shared with a system that uses the newer borg version once.

ThomasWaldmann:
typo in commit comment.

ThomasWaldmann:
looks correct. can maybe be refined later.

enkore:
I'll just merge this now. As was established in #2659 this is a choice that can be refined later with little effort, but no one seems to want to invest the time to do that right now.

PR #2661, opened 2017-06-11
normalize authenticated key modes

enkore:
rename authenticated to authenticated-blake2, consistent with the other
blake2 key modes

add authenticated mode that fills the blank and is consistent with the
other "unqualified" key modes

TYPE IDs remain of course unchanged and compatible.

### DIFF
--- a/docs/usage/init.rst.inc
+++ b/docs/usage/init.rst.inc
@@ -72,6 +72,26 @@ the encryption/decryption key or other secrets.
 Encryption modes
 ++++++++++++++++
 
++----------+---------------+------------------------+--------------------------+
+| Hash/MAC | Not encrypted | Not encrypted,         | Encrypted (AEAD w/ AES)  |
+|          | no auth       | but authenticated      | and authenticated        |
++----------+---------------+------------------------+--------------------------+
+| SHA-256  | none          | authenticated          | repokey, keyfile         |
++----------+---------------+------------------------+--------------------------+
+| BLAKE2b  | n/a           | authenticated-blake2   | repokey-blake2,          |
+|          |               |                        | keyfile-blake2           |
++----------+---------------+------------------------+--------------------------+
+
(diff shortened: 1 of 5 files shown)


textshell:
LGTM

docs/usage/init.rst.inc
```diff
@@ -80,23 +100,21 @@ These modes are compatible with borg 1.0.x.
...
-hardware-accelerated. BLAKE2b is faster than SHA256 on Intel/AMD 64bit CPUs,
-which makes `authenticated` faster than `none`.
-
-On modern ARM CPUs, NEON provides hardware acceleration for SHA256 making it faster
-than BLAKE2b-256 there.
\ No newline at end of file
+Use it only for new repositories where no encryption is wanted **and** when compatibility
+with 1.0.x is important. If compatibility with 1.0.x is not important, use
+`authenticated-blake2` or `authenticated` instead.
+This mode is compatible with Borg 1.0.x.
```
ThomasWaldmann: diff complains about missing lf.

src/borg/crypto/key.py
```diff
@@ -825,9 +822,22 @@ def decrypt(self, id, data, decompress=True):
...
+    ARG_NAME = 'authenticated-blake2'
+
+
 AVAILABLE_KEY_TYPES = (
     PlaintextKey,
     PassphraseKey,
+    AuthenticatedKey,
     KeyfileKey, RepoKey,
-    Blake2KeyfileKey, Blake2RepoKey, AuthenticatedKey,
+    Blake2KeyfileKey, Blake2RepoKey, Blake2AuthenticatedKey,
```
ThomasWaldmann: move AuthenticatedKey to the end of line 841, so it is similar sequence as 842?

docs/usage/init.rst.inc
enkore: These files all don't seem to end with a new-line.

PR #2668, opened 2017-06-12
changed the date of "day without backup"

philippje:
Changed from 20. December to 19. December for easier comprehension (viewing the calendar.)
The missing 'd' at 20. December is hardly noticeable compared to e.g. the 19. December.


ThomasWaldmann:
thanks!

PR #2671, opened 2017-06-13
chunker: don't do uint32_t >> 32

enkore:
As luck would have it at least x86 already performs this masking. Others probably do as well. GCC knows this and emits the same code.

Found by ubsan. With #2670 and #2671 applied, running the test suite is ubsan-clean.

### DIFF
--- a/src/borg/_chunker.c
+++ b/src/borg/_chunker.c
@@ -63,7 +63,7 @@ static uint32_t table_base[] =
     0xc5ae37bb, 0xa76ce12a, 0x8150d8f3, 0x2ec29218, 0xa35f0984, 0x48c0647e, 0x0b5ff98c, 0x71893f7b
 };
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
https://en.wikipedia.org/wiki/Circular_shift

So, how about:
```
return ((v) << ((shift) & 0x1f)) | ((v) >> ((-(shift)) & 0x1f));
```

enkore:
The macro is compiled down to one rol instruction.

ThomasWaldmann:
... if you have an optimizing compiler.

WP suggests that this is the case for their version, too (and I have just slightly rewritten it).

For non-optimizing compilers, a negation might be quicker than (32 - x).
Also, the left shift value is masked (for symmetry and correctness for shift > 32).

enkore:
The left side is masked outside the macro (imod).

PR #2672, opened 2017-06-13
info: use CacheSynchronizer & HashIndex.stats_against

enkore:
Gets rid of the cache "abuse" (avoiding some I/O, should be nice for `borg info ... --last/--first/-a`) and is also faster. Also, cached archive chunks indices could be used, but that code is touched in the --no-cache-sync PR.

Logically, nothing changes. size and csize are taken from the master (Cache.chunks) index.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -25,7 +25,7 @@ from .cache import ChunkListEntry
 from .crypto.key import key_factory
 from .compress import Compressor, CompressionSpec
 from .constants import *  # NOQA
-from .hashindex import ChunkIndex, ChunkIndexEntry
+from .hashindex import ChunkIndex, ChunkIndexEntry, CacheSynchronizer
 from .helpers import Manifest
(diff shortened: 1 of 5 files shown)


src/borg/archive.py
```diff
@@ -478,30 +478,22 @@ def save(self, name=None, comment=None, timestamp=None, additional_metadata=None
...
-            for item in unpacker:
-                chunks = item.get(b'chunks')
-                if chunks is not None:
-                    stats.nfiles += 1
-                    add_file_chunks(chunks)
-        cache.rollback()
+            sync.feed(data)
+        stats = Statistics()
+        stats.osize, stats.csize, unique_size, stats.usize, unique_chunks, chunks = archive_index.stats_against(cache.chunks)
+        stats.nfiles = sync.num_files
```
enkore: Since we're memorizing size in Items, we could do the same for Archives and the number of files.

PR #2676, opened 2017-06-14
fuse: redo ItemCache

enkore:
~Halves time to load an archive, far less /tmp usage, slightly faster traversal of directories. There are still a lot of low-hanging fruits left in this code. Memory usage increases slightly due to the meta array (for 1 million items it is approx 10 MB - 9.0x bytes on average per Item).

Note: the next stable release of msgpack allows to avoid write_bytes, since a tell() method was added.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -2,6 +2,7 @@ import errno
 import io
 import os
 import stat
+import struct
 import sys
 import tempfile
 import time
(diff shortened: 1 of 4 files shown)


src/borg/fuse.py
```diff
@@ -158,36 +307,31 @@ def process_archive(self, archive_name, prefix=[]):
...
         t0 = time.perf_counter()
-        unpacker = msgpack.Unpacker()
         archive = Archive(self.repository_uncached, self.key, self.manifest, archive_name,
                           consider_part_files=self.args.consider_part_files)
-        for key, chunk in zip(archive.metadata.items, self.repository.get_many(archive.metadata.items)):
-            data = self.key.decrypt(key, chunk)
-            unpacker.feed(data)
-            for item in unpacker:
-                item = Item(internal_dict=item)
-                path = os.fsencode(os.path.normpath(item.path))
```
enkore: I removed the normpaths since it's a rather expensive operation that is already always done by `borg create` anyway.

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
         self.offset = 1000000
 
-    def add(self, item):
-        pos = self.fd.seek(0, io.SEEK_END)
-        self.fd.write(msgpack.packb(item.as_dict()))
-        return pos + self.offset
+        # A temporary file that contains direct items, i.e. items directly cached in this layer.
+        # These are items that span more than one chunk and thus cannot be efficiently cached
+        # by the object cache (self.decrypted_repository), which would require variable-length structures;
+        # possible but not worth the effort, see inode_for_current_item.
```
enkore: update reference

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+
+    indirect_entry_struct = struct.Struct('=cII')
+    assert indirect_entry_struct.size == 9
+
+    def __init__(self, decrypted_repository):
+        self.decrypted_repository = decrypted_repository
+        # self.meta, the "meta-array" is a densely packed array of metadata about where items can be found.
+        # It is indexed by the inode number minus self.offset. (This is in a way eerily similar to how the first
+        # unices did this).
+        # The meta-array contains chunk IDs and item entries (described in inode_for_current_item).
```
enkore: dito

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+
+        # A small LRU cache for chunks requested by ItemCache.get() from the object cache,
+        # this significantly speeds up directory traversal and similar operations which
+        # tend to re-read the same chunks over and over.
+        # The capacity is kept low because increasing it does not provide any significant advantage,
+        # but makes LRUCache's square behaviour noticeable as well as consuming some memory.
+        self.chunks = LRUCache(capacity=10, dispose=lambda _: None)
+
+        # Instrumentation
+        # Count of indirect items, i.e. data is cached in the object cache, in this cache
```
enkore: , not in

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+        self.indirect_items = 0
+        # Count of direct items, i.e. data is in self.fd
+        self.direct_items = 0
 
     def get(self, inode):
         offset = inode - self.offset
         if offset < 0:
             raise ValueError('ItemCache.get() called with an invalid inode number')
```
enkore: Future XXX: Remove hackiness by merging FuseOps.items into ItemCache.

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+        if self.meta[offset] == ord(b'S'):
+            fd_offset = int.from_bytes(self.meta[offset + 1:offset + 9], 'little')
+            self.fd.seek(fd_offset, io.SEEK_SET)
+            return Item(internal_dict=next(msgpack.Unpacker(self.fd, read_size=1024)))
+        else:
+            _, chunk_id_relative_offset, chunk_offset = self.indirect_entry_struct.unpack_from(self.meta, offset)
+            chunk_id_offset = offset - chunk_id_relative_offset
+            # bytearray slices are bytearrays as well, explicitly convert to bytes()
+            chunk_id = bytes(self.meta[chunk_id_offset:chunk_id_offset + 32])
+            chunk_offset = int.from_bytes(self.meta[offset + 5:offset + 9], 'little')
```
enkore: duplicates L100 and is also wrong.

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+
+        write_offset = self.write_offset
+        meta = self.meta
+        pack_indirect_into = self.indirect_entry_struct.pack_into
+
+        def write_bytes(append_msgpacked_bytes):
+            nonlocal msgpacked_bytes
+            nonlocal stream_offset
+            msgpacked_bytes += append_msgpacked_bytes
+            stream_offset += len(append_msgpacked_bytes)
```
enkore: (insert comment about tell() here)

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+                current_item = msgpacked_bytes
+                msgpacked_bytes = b''
+
+                if write_offset + 9 >= len(meta):
+                    self.meta = meta = meta + bytes(self.GROW_META_BY)
+
+                # item entries in the meta-array come in two different flavours, both nine bytes long.
+                # (1) for items that span chunks:
+                #
+                #     'S' + 8 byte offset into the self.fd file, where the msgpacked item starts.
```
enkore: 4 bytes should be easily sufficient, but having both entry types the same length could unify some code later

src/borg/fuse.py
```diff
@@ -35,22 +36,161 @@ def fuse_main():
...
+                    meta[write_offset:write_offset + 9] = b'S' + pos.to_bytes(8, 'little')
+                    write_offset += 9
+                    self.direct_items += 1
+                    inode = write_offset - 9 + self.offset
+                else:
+                    item_offset = stream_offset - current_item_length - chunk_begin
+                    pack_indirect_into(meta, write_offset, b'I', write_offset - current_id_offset, item_offset)
+                    write_offset += 9
+                    self.indirect_items += 1
+                    inode = write_offset - 9 + self.offset
```
enkore: unify identical insns in branches

enkore:
Note: For 64 bit ABIs it's quite straightforward to avoid the indirection and just encode the offsets directly into the (64 bit) inode (with the side effect that indirect items have the same inode in all archives they may appear in). 32 bit ABIs, and also 32 bit applications on a 64 bit system, do not support 64 bit inodes.

src/borg/fuse.py
```diff
@@ -35,22 +36,160 @@ def fuse_main():
...
+            data = memoryview(chunk)[chunk_offset:]
+            unpacker = msgpack.Unpacker()
+            unpacker.feed(data)
+            return Item(internal_dict=next(unpacker))
+
+    def iter_archive_items(self, archive_item_ids):
+        unpacker = msgpack.Unpacker()
+
+        stream_offset = 0
+        chunk_begin = 0
```
ThomasWaldmann: add a comment about what precisely these 2 mean?

src/borg/fuse.py
ThomasWaldmann: was this always the case, even for attic?

src/borg/lrucache.py
```diff
@@ -28,6 +28,14 @@ def __delitem__(self, key):
     def __contains__(self, key):
         return key in self._cache
 
+    def get(self, key, default=None):
+        value = self._cache.get(key, default)
+        if value is default:
+            return value
```
ThomasWaldmann: for detecting that something is not in _cache (and thus needs no further treatment in 35/36), the code needs to use some unique default value that can not exist as a legitimate value.

So, it can't be just None, but must be something like doesnotexistasvalue = object().

src/borg/fuse.py
enkore: Attic 0.8 normalized arguments and joined dirents to them, which results in normalized paths as well.

src/borg/lrucache.py
enkore: That doesn't really make sense (I do know the sentinel pattern), since the LRUCache is specifically meant to cache *something* not *Nonething*.

ThomasWaldmann: sure it makes sense if None could be part of the legitimate values.

if you do not want to allow that, document and assert it at least.

enkore:
...

src/borg/fuse.py
```diff
@@ -35,22 +36,160 @@ def fuse_main():
...
-        return Item(internal_dict=item)
+        if self.meta[offset] == ord(b'S'):
+            fd_offset = int.from_bytes(self.meta[offset + 1:offset + 9], 'little')
+            self.fd.seek(fd_offset, io.SEEK_SET)
+            return Item(internal_dict=next(msgpack.Unpacker(self.fd, read_size=1024)))
+        else:
+            _, chunk_id_relative_offset, chunk_offset = self.indirect_entry_struct.unpack_from(self.meta, offset)
+            chunk_id_offset = offset - chunk_id_relative_offset
+            # bytearray slices are bytearrays as well, explicitly convert to bytes()
+            chunk_id = bytes(self.meta[chunk_id_offset:chunk_id_offset + 32])
```
enkore: I don't feel like checking _again how_ much that impacts things, esp. in the code below (this method can't get around member lookups, the iteration code below _does_).

ThomasWaldmann: well, I just don't like hardcoded numbers much. attic was full of it and it wasn't pretty.

if you like a module level HASH_LEN better, that would also be an improvement already.

about speed: don't think it should matter that much and maybe we should not micro-optimize on that level. esp. if (as you say) there are a lot of low hanging fruits still that likely have a much bigger impact.

enkore: > about speed: don't think it should matter that much and maybe we should not micro-optimize on that level.

No. ItemCache (before and after, sorting movements of data handling) is one of the biggest CPU time eaters when processing archives. In the current revision that's about 60%. These are at a point where you can remove a few VM ops, or replace expensive VM ops (such as member lookups) with cheaper ones (such as local variable lookups), like I did in iter_archive_bla with the meta and offset variables and get a tangible improvement.

Btw. I'd really prefer it if you would not quote me wrong. I referred to memory overhead, not throughput, although either is bad. (Python, and especially CPython, are simply not very good choices to achieve either, but that ship has sailed years ago for Borg of course... although I'll see in 1.2-dev what I can still salvage...).

ThomasWaldmann: """~Halves time to load an archive, far less /tmp usage, slightly faster traversal of directories. There are still a lot of low-hanging fruits left in this code. Memory usage increases slightly due to the meta array (for 1 million items it is approx 10 MB - 9.0x bytes on average per Item)."""

First you talked about speed, then low hanging fruits, then memory.

I did not quote you wrong, you just did not say it precise enough.

ThomasWaldmann: OK, if it makes that much of a difference, we can keep it as is.

PR #2691, opened 2017-06-17
WIP added some comments about IntegrityErrors

ThomasWaldmann:
I just added some comments for now.


src/borg/archive.py
```diff
@@ -1145,10 +1145,10 @@ def check(self, repository, repair=False, archive=None, first=0, last=0, sort_by
         else:
             try:
                 self.manifest, _ = Manifest.load(repository, (Manifest.Operation.CHECK,), key=self.key)
-            except IntegrityError as exc:
+            except IntegrityError as exc:  # KILLING manifest
                 logger.error('Repository manifest is corrupted: %s', exc)
                 self.error_found = True
-                del self.chunks[Manifest.MANIFEST_ID]
+                del self.chunks[Manifest.MANIFEST_ID]  # XXX there is another del in rebuild_refcounts (2 lines below)
```
enkore: Strange, test_manifest_rebuild* should hit this, but don't.

src/borg/repository.py
```diff
@@ -1183,7 +1183,7 @@ def is_committed_segment(self, segment):
         while True:
             try:
                 tag, key, offset, _ = next(iterator)
-            except IntegrityError:
+            except IntegrityError:  # KILLING? because found to be "not committed"?
                 return False
             except StopIteration:
                 break
```
enkore: Yes, these two are an inherent (though I think not very important) issue in a transacted store; it's difficult to tell an aborted transaction (which can of course have partial writes and all that) from a corrupted completed transaction.

This is mitigated in two ways:
- 1.0: only one commit is left alive, and if it is corrupted it will try all segments, failing its way through all data, until the first segment, then abort and tell to run check (which will do the same and in the end stick a commit on it).
- 1.1: commit tags generally survive, so if one is corrupted the previous transaction would be accepted, depending on what exactly was done with some data loss. For compaction this is further mitigated because each segment ends with a commit.

src/borg/archive.py
ThomasWaldmann: seems like this del does not raise a KeyError when it should.

enkore: https://github.com/borgbackup/borg/blob/master/src/borg/hashindex.pyx#L118

Well d'uh

ThomasWaldmann: #2693

enkore:
Status?

ThomasWaldmann:
Did no further work on it yet. The point is not just having comments, but to review all these places.

PR #2693, opened 2017-06-17
implement Hashindex.__delete__ KeyError

ThomasWaldmann:


src/borg/testsuite/hashindex.py
```diff
@@ -48,7 +48,7 @@ def _generic_test(self, cls, make_value, sha):
             del idx[H(x)]
         # Test some keys still in there
         for x in range(50, 100):
-            assert H(x) in idx
+            pass # assert H(x) in idx  # TODO: WTF does this break the hash test?
```
ThomasWaldmann: @enkore any idea?

ThomasWaldmann: solved, see latest commit.

src/borg/testsuite/hashindex.py
```diff
@@ -5,6 +5,8 @@
 import tempfile
 import zlib
 
+import pytest
```
enkore: # Note: these tests are part of the self test, do not use or import py.test functionality here.
 #       See borg.selftest for details. If you add/remove test methods, update SELFTEST_COUNT

src/borg/testsuite/hashindex.py
enkore: Hm? Wouldn't the sha change anyway since the deletes modify the thing?

src/borg/testsuite/hashindex.py
ThomasWaldmann: fixed.

src/borg/testsuite/hashindex.py
ThomasWaldmann: all "del"s I added are for keys that do not exist, so they logically do not change anything.

ThomasWaldmann: also, i tracked it down to the "in" tests, they are changing the sha due to the reason mentioned now in the _hashindex.c comment.

ThomasWaldmann:
collapse / merge?

PR #2694, opened 2017-06-17
switch docs theme

enkore:
no backport of this, but the requirements file (it seems this was undocumented previously) might need to be touched in 1.0-maint, depending on RTD build configuration.

### DIFF
--- a/docs/_templates/logo-text.html
+++ b/docs/_templates/logo-text.html
@@ -1,0 +1,5 @@
+<a href="{{ homepage() }}" class="text-logo">
+  <img src='{{ pathto('_static/logo.svg', 1) }}' width='100%'>
+
+  {{ theme_project_nav_name or shorttitle }}
+</a>
--- a/docs/borg_theme/css/borg.css
+++ b/docs/borg_theme/css/borg.css
@@ -39,12 +39,45 @@ dt code {
     font-weight: normal;
 }
 
-.experimental,
+/* bootstrap has a .container class which clashes with docutils' container class. */
+.docutils.container {
+    width: auto;
+    margin: 0;
+    padding: 0;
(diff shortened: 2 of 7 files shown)


docs/_templates/logo-text.html
```diff
@@ -0,0 +1,5 @@
+<a href="{{ homepage() }}" class="text-logo">
+  <img src='{{ pathto('_static/logo.svg', 1) }}' width='100%'>
```
enkore: [(It's documented that way)](http://www.sphinx-doc.org/en/stable/templating.html#pathto)

docs/borg_theme/css/borg.css
```diff
@@ -39,12 +39,45 @@ dt code {
...
+}
+
+.container.experimental,
 #debugging-facilities,
 #borg-recreate {
     /* don't change text dimensions */
-    margin: 0 -40px; /* padding below + border width */
-    padding: 0 20px; /* 20 px visual margin between edge of text and the border */
+    margin: 0 -30px; /* padding below + border width */
+    padding: 0 10px; /* 10 px visual margin between edge of text and the border */
```
enkore: slightly different margins around the body in this theme

docs/conf.py
```diff
@@ -51,7 +49,8 @@
 # built documents.
 #
 # The short X.Y version.
-version = sw_version.split('-')[0]
+split_char = '+' if '+' in sw_version else '-'
+version = sw_version.split(split_char)[0]
```
enkore: ... probably a remnant of previous scm_setuptools versions ...

enkore:
For reference

2017-06-18-013215_3840x1440_scrot

PR #2698, opened 2017-06-18
docs: neater table borders

enkore:


docs/borg_theme/css/borg.css
```diff
@@ -94,11 +94,39 @@ h1 {
     border-right: 2px solid #4e4a4a;;
 }
 
-table.docutils td,
-table.docutils th {
+table.docutils:not(.footnote) td,
+table.docutils:not(.footnote) th {
```
enkore: Turns out, footnotes are actually tables.

enkore: HTML is a very sophisticated type-setting system, you see...

PR #2715, opened 2017-06-20
delete: remove short option for --cache-only

enkore:
Overlooked in #2692 

This -c also shadows create's -c option. I didn't even knew it existed.


ThomasWaldmann:
@enkore "This -c also shadows create's -c option."

Does it? This "-c" is added as an option to the delete subparser, not to the global options parser.

enkore:
I meant logically - same short option char, very different meaning. I've thought about this change, but I think it's okay from a breakage point of view.

ThomasWaldmann:
ah, ok.

PR #2716, opened 2017-06-20
docs: format metavars more accurately

enkore:


setup.py
```diff
@@ -203,6 +203,20 @@ def detect_libb2(prefixes):
     long_description = re.compile(r'^\.\. highlight:: \w+$', re.M).sub('', long_description)
 
 
+def format_metavar(option):
+    if option.nargs in ('*', '...'):
+        if option.required:
```
enkore: That's a bit of a strange one ... "zero or more arguments but at least one". These instances should probably be replaced by nargs=+.

PR #2731, opened 2017-06-23
Manifest: use limited unpacker

### LINKED ISSUE #2175: Limited unpacker for manifest
The same goes for the unpacker used in `unpack_and_verify_manifest`. (Which would mean that we also need to settle on a maximum number of archives per repo for now -- this is already implicitly limited to IIRC something like 400k by the 20 MB object size limit, but we should really just declare one. N.B. since archive names are kinda arbitrary this might be more difficult).

Split from #2139

ThomasWaldmann:
fixes #2175 

the test is rather slow and now pointless. previously, i tested for going over MAX_ARCHIVES, but that does not work any more due to the assert in Manifest.write(), so I guess the test can be removed.

### DIFF
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -21,10 +21,10 @@ from ..helpers import StableDict
 from ..helpers import Error, IntegrityError
 from ..helpers import yes
 from ..helpers import get_keys_dir, get_security_dir
+from ..helpers import get_limited_unpacker
 from ..helpers import bin_to_hex
 from ..item import Key, EncryptedKey
 from ..platform import SaveFile
-from .. import remote
 from .nonces import NonceManager
 from .low_level import AES, bytes_to_long, bytes_to_int, num_aes_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
 
@@ -217,7 +217,7 @@ class KeyBase:
             logger.warning('Manifest authentication DISABLED.')
             tam_required = False
         data = bytearray(data)
-        unpacker = remote.get_limited_unpacker('manifest')
+        unpacker = get_limited_unpacker('manifest')
         unpacker.feed(data)
         unpacked = unpacker.unpack()
         if b'tam' not in unpacked:
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -24,6 +24,7 @@ from ..helpers import get_keys_dir, get_security_dir
 from ..helpers import bin_to_hex
 from ..item import Key, EncryptedKey
 from ..platform import SaveFile
+from .. import remote
 from .nonces import NonceManager
 from .low_level import AES, bytes_to_long, bytes_to_int, num_aes_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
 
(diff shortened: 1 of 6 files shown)


src/borg/crypto/key.py
```diff
@@ -216,9 +217,9 @@ def unpack_and_verify_manifest(self, data, force_tam_not_required=False):
             logger.warning('Manifest authentication DISABLED.')
             tam_required = False
         data = bytearray(data)
-        # Since we don't trust these bytes we use the slower Python unpacker,
-        # which is assumed to have a lower probability of security issues.
-        unpacked = msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape')
+        unpacker = remote.get_limited_unpacker('manifest')
```
enkore: The rationale for that is a bit flawed since modifying the manifest data and modifying RPC is pretty similar in the attack model, but the slow packer isn't (and shouldn't) be used for RPC.

src/borg/remote.py
```diff
@@ -81,8 +83,19 @@ def get_limited_unpacker(kind):
         args.update(dict(max_array_len=LIST_SCAN_LIMIT,  # result list from repo.list() / .scan()
                          max_map_len=100,  # misc. result dicts
                          ))
+    elif kind == 'manifest':
+        args.update(dict(use_list=True,  # default value
+                         max_array_len=100,  # ITEM_KEYS ~= 22
+                         max_map_len=MAX_ARCHIVES,  # list of archives
+                         max_str_len=1000,  # archive name
```
enkore: This limitation does not exist on the writing side of things.

Also, the limit on archive names would need to be lower for good compatibility (~<250).

src/borg/testsuite/archive.py
```diff
@@ -1,18 +1,44 @@
...
 
 from . import BaseTestCase
+from ..constants import *  # NOQA
 from ..crypto.key import PlaintextKey
 from ..archive import Archive, CacheChunkBuffer, RobustUnpacker, valid_msgpacked_dict, ITEM_KEYS, Statistics
 from ..archive import BackupOSError, backup_io, backup_io_iter
 from ..helpers import Manifest
 from ..item import Item, ArchiveItem
 
+from ..repository import Repository
```
enkore: empty line

src/borg/crypto/key.py
ThomasWaldmann: yeah. i just copied that comment and the usage of the slow unpacker from there.
just using the fast msgpack (C) code is easy.

src/borg/remote.py
ThomasWaldmann: well, the intention of this limitation is to avoid attacks, not to check length for other parts of the code. i'll add some more asserts to write().

ThomasWaldmann:
Tests failing. Expected UnpackError, got ValueError?

Is there an API difference between C and Python?

enkore:
Yes, though it'll be fixed in the next msgpack release. (Compare msgpack/unpacker.pyx, _unpack).

src/borg/remote.py
```diff
@@ -81,8 +83,16 @@ def get_limited_unpacker(kind):
...
+                         max_array_len=100,  # ITEM_KEYS ~= 22
+                         max_map_len=MAX_ARCHIVES,  # list of archives
+                         max_str_len=255,  # archive name
+                         object_hook=StableDict,
+                         unicode_errors='surrogateescape',
+                         ))
     else:
-        raise ValueError('kind must be "server" or "client"')
+        raise ValueError('kind must be "server", "client" or "manifest"')
     return msgpack.Unpacker(**args)
```
ThomasWaldmann: ^ move this func from borg.remote to borg.helpers as it now is used in general, not just by rpc code?

ThomasWaldmann:
partially collapse, merge?

Issue #2121, opened 2017-02-03
docs: use "restrict" option for authorized_keys


ThomasWaldmann:
it is preferable over specifying single restrictions as it will also include all future restrictions.

but: it requires openssh >=7.1, so we have to wait with that docs change until we can expect most people have that.

ThomasWaldmann:
[restrict.diff.txt](https://github.com/borgbackup/borg/files/751735/restrict.diff.txt)

enkore:
removing later labe, @rugk's suggestion is good (use restrict, adding a note about it)

PR #2732, opened 2017-06-23
Simplify ssh authorized_keys file

rugk:
Just using "restrict"; closes https://github.com/borgbackup/borg/issues/2121

Reasoning:
Even Debian 9 (now stable) [uses v7.4](https://packages.debian.org/de/stretch/openssh-server).

AFAIK it is still not possible in Ubuntu Trusty, but hopefully nobody uses that anyway. And waiting for this really nice doc change until April 2019 is not a good idea, IMHO.

Of course, we could also just add another note saying that you need to have support for that. If using it with an older ssh version it likely does not work, so users do not get into an insecure state.
E.g. a line saying:
> If you're using openssh-server < 7.2, you have to explicitly specify the ssh features to restrict and cannot simply use the `restrict` option. Use `,no-port-forwarding,no-X11-forwarding,no-pty,no-agent-forwarding,no-user-rc` in this case, may be a good idea.

Also, this doc change may also *not* be backported, so only recent borg versions have it in their doc, so when installing borg from the package repos one still gets the working doc version.

### DIFF
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -37,5 +37,5 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
     If you're using openssh-server < 7.2, however, you have to explicitly specify
     the ssh features to restrict and cannot simply use the restrict option as it
     has been introduced in v7.2. We recommend to use
-    ``,no-port-forwarding,no-X11-forwarding,no-pty,no-agent-forwarding,no-user-rc``
+    ``no-port-forwarding,no-X11-forwarding,no-pty,no-agent-forwarding,no-user-rc``
     in this case.
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -29,3 +29,13 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
     $ cat ~/.ssh/authorized_keys
     command="export BORG_XXX=value; borg serve [...]",restrict ssh-rsa [...]
 
+.. note::
+    The examples above use the ``restrict`` directive. This does automatically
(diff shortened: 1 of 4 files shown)


enkore:
Yes, such a notice would be good. Perhaps put it into the deployment.rst. `.. note::`...

rugk:
Okay, but I would rather like to put it into the "borg serve" doc. As in the deployment doc it is only mentioned in the ansible/salt part and users not wanting to use ansible/salt likely don't read it. Or, if one only reads the salt part and we have the note in the ansible part that's going to be bad.

I also noticed in [the general hosting "deplyoment" docs](https://borgbackup.readthedocs.io/en/latest/deployment/hosting-repositories.html) you already use `restricted`. Just in a different order (not good to confuse users). I'll change this too.

docs/usage/serve.rst
```diff
@@ -23,9 +23,19 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+    command="borg serve --restrict-to-path /path/to/repo",restrict ssh-rsa AAAAB3[...]
 
     # Set a BORG_XXX environment variable on the "borg serve" side
     $ cat ~/.ssh/authorized_keys
     command="export BORG_XXX=value; borg serve [...]",restrict ssh-rsa [...]
 
+.. note::
+    The examples above use the ``restrict`` directive. This does automatically
+    block potential dangerous ssh features, even when they are added in a future
+    update. Thus, this option should be prefered.
```
ThomasWaldmann: preferred

rugk: https://github.com/borgbackup/borg/pull/2736

Issue #2735, opened 2017-06-23
Link prune example


rugk:
[In the prune doc](https://borgbackup.readthedocs.io/en/latest/usage/prune.html) you state:
> There is also a visualized prune example in docs/misc/prune-example.txt.

Now, average Joe or the man on the Clapham omnibus, thinks: "That's nice, but where can I find this file?"

That's why I suggest to add a link to the file on GitHub. Or do something different to make it easier to find this file. You cannot expect a user to download the source code/clone the repo first. And even for that, they have to know the path relates to the source code and the file is meant to be found there.

enkore:
```
.. include:: misc/prune...
    :literal:
```

(http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment)

In the examples section, or in a separate section following it.

rugk:
So you want to include the file directly? Sure, also possible.

PR #2737, opened 2017-06-23
Include prune example in doc

rugk:
Fixes https://github.com/borgbackup/borg/issues/2735

I also moved it to the bottom as the main examples are more important and this is a large file.

Not tested practically (could not find an easy way to parse these files), I just hope the instructions were correct. :smile:


enkore:
To build the docs locally, see: http://borgbackup.readthedocs.io/en/latest/development.html#building-the-docs-with-sphinx

PR #2748, opened 2017-06-26
archiver: more argparse cleanup, redundant options, missing metavars

enkore:
Lots of instances missed in #2714 

In the future I'd like an entirely consistent layout of option definitions:

```py
add_argument('--option', [metavar='METAVAR',] [dest='...',] [action=...]
             [types, defaults and other rare stuff — defaults go in line above if nothing else is declared]
             help='....')

### CODE REVIEW

File: src/borg/archiver.py [hunk 1]
```diff
@@ -3053,15 +3023,15 @@ def define_common_options(add_common_option):
         exclude_group.add_argument('--exclude-from', action=ArgparseExcludeFileAction,
                                    metavar='EXCLUDEFILE', help='read exclude patterns from EXCLUDEFILE, one per line')
         exclude_group.add_argument('--exclude-caches', dest='exclude_caches',
-                                   action='store_true', default=False,
+                                   action='store_true',
```
ThomasWaldmann: join with previous line?

enkore:
In almost all cases `dest=...` is unnecessary as it appears. It has some (minor) advantage making it more obvious, but also adds noise to these definitions.

ThomasWaldmann:
are these bigger changes you added right now beneficial?
won't it just make make backporting stuff harder?

enkore:
I'd estimate that 98 % of all backport changes don't touch build_parser.

PR #2750, opened 2017-06-26
release 1.0.11rc1

ThomasWaldmann:


docs/changes.rst
```diff
@@ -229,7 +229,9 @@ Other changes:
     to break was fixed in cygwin at least since cygwin 2.8, maybe even since
     2.7.0).
   - remove skipping the noatime tests on GNU/Hurd, #2710
-
+  - fix borg import issue, add comment, #2718
```
enkore: I would've no idea what this means if I didn't occasionally read my github notifications...

docs/changes.rst
```diff
@@ -229,7 +229,9 @@ Other changes:
...
+  - include attic.tar.gz when installing the package
```
enkore: The testsuite-independent-of-attic change wasn't released

docs/changes.rst
enkore: Actually I don't think that this was released, since it was part of the manifest feature flags changeset.

ThomasWaldmann: i committed this recently. it fixes the mentioned ticket that was breaking local tox tests.
and the manifest feature flags were merged into 1.0-maint also.

docs/changes.rst
ThomasWaldmann: https://github.com/borgbackup/borg/blob/1.0-maint/borg/testsuite/attic.tar.gz

docs/changes.rst
enkore: Didn't see that this was the "Other" category; the line above in the diff is "Bug fixes", where only regressions fixed since a release should be noted. idrc about "other".

PR #2758, opened 2017-06-27
Add some missing tests

enkore:


enkore:
```
Coverage    84.3%   85.25%   +0.95%
                    ^^
```

It's of course a flawed metric, but still kinda nice to see that number :)

PR #2761, opened 2017-06-27
archiver: add test for "create -" and "extract --stdout"

enkore:


src/borg/testsuite/archiver.py
```diff
@@ -78,12 +78,18 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, input=b'', **kw):
         except SystemExit as e:  # possibly raised by argparse
             output = ''
             ret = e.code
-        return ret, os.fsdecode(output)
+        if binary_output:
+            return ret, output
+        else:
+            return ret, os.fsdecode(output)
```
ThomasWaldmann: return ret, output if binary_output else os.fsdecode(output)

src/borg/testsuite/archiver.py
```diff
@@ -95,9 +101,11 @@ def exec_cmd(*args, archiver=None, fork=False, exe=None, input=b'', **kw):
                 # actions that abort early (eg. --help) where given. Catch this and return
                 # the error code as-if we invoked a Borg binary.
             except SystemExit as e:
-                return e.code, output.getvalue()
+                output_text.flush()
+                return e.code, output.getvalue() if binary_output else output.getvalue().decode()
             ret = archiver.run(args)
-            return ret, output.getvalue()
+            output_text.flush()
+            return ret, output.getvalue() if binary_output else output.getvalue().decode()
```
ThomasWaldmann: interesting :)

Issue #1293, opened 2016-07-08
Maybe use per-archive cache for delete / prune


textshell:
The per-archive cache already has all chunk ids used in an archive. So it should be possible to use that for quick removal of an already cached archive. On the other hand this will not always be available. Not sure if there are any other problems.

PR #2764, opened 2017-06-27
WIP Split cache sync

### LINKED ISSUE #2866: When merging/pulling chunk index --progress-like output is used despite not passed
When you need to merge the chunk index as you deleted the local chunk index/cache, then you always get this log output: (when piping/directing it to a log file as you usually do in automated systems):
```
Merging chunk index... (0.00%)
Merging chunk index... (1.22%)
Merging chunk index... (3.32%)
[…]
Merging chunk index... (100.00%)
```
(no original log file, just how I remember it)

However the command I executed there (borg create), did not have the `--progress` parameter passed, but just some usual ones (`-v` and `--stats`), whcih do have such a progress bar/indicator. Thus, IMHO, the behaviour of borg is wrong here. The output without `--progress` should be something like `Merging chunk index...` without any percentage.

v1.0.10

enkore:
First commit is mostly refactor; functional changes should only be the removal of a couple redundant log messages.

Related #1293 
Fixes #2866

TODO:

- [ ] This code has no existing tests. Now it can be efficiently tested. Thus, develop tests.

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -780,18 +780,30 @@ class LocalCache(CacheStatsMixin):
             except FileNotFoundError:
                 pass
 
-        def build_index(self, decrypted_repository, chunk_idx, cache_sync_stats):
+        def build_index(self, decrypted_repository, chunk_idx, cache_sync_stats=None, detailed_pi=None):
+            processed_item_metadata_bytes = 0
+            processed_item_metadata_chunks = 0
             csize, data = decrypted_repository.get(self.id)
             chunk_idx.add(self.id, 1, len(data), csize)
             archive = ArchiveItem(internal_dict=msgpack.unpackb(data))
             if archive.version != 1:
                 raise Exception('Unknown archive metadata version')
             sync = CacheSynchronizer(chunk_idx)
+            if detailed_pi:
+                detailed_pi.total = len(archive.items)
             for item_id, (csize, data) in zip(archive.items, decrypted_repository.get_many(archive.items)):
                 chunk_idx.add(item_id, 1, len(data), csize)
-                cache_sync_stats.processed_item_metadata_bytes += len(data)
-                cache_sync_stats.processed_item_metadata_chunks += 1
+                processed_item_metadata_bytes += len(data)
+                processed_item_metadata_chunks += 1
                 sync.feed(data)
+                if detailed_pi:
+                    detailed_pi.show(increase=1)
+            if detailed_pi:
+                detailed_pi.finish()
+            if cache_sync_stats:
+                cache_sync_stats.processed_item_metadata_bytes += processed_item_metadata_bytes
+                cache_sync_stats.processed_item_metadata_chunks += processed_item_metadata_chunks
+            return sync
 
         def write_index(self, chunk_idx, cache_sync_stats):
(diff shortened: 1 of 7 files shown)


ThomasWaldmann:
quite a big diff considering you didn't want a b7.

src/borg/cache.py
```diff
@@ -623,15 +623,13 @@ def create_master_idx(self, decrypted_repository, chunk_idx):
...
                                               msg='%3.0f%% Syncing chunks cache. Processing archive %s',
                                               msgid='cache.sync')
                 chunks_archives = self.get_chunks_archives(archive_ids)
                 for chunks_archive in chunks_archives:
                     pi.show(info=[remove_surrogates(chunks_archive.name)])
-                    if not self.enabled:
-                        chunk_idx = chunk_idx or ChunkIndex(master_index_capacity)
                     archive_chunk_idx = chunks_archive.procure_index(decrypted_repository, cache_sync_stats)
                     if self.chunks_archive_enabled:
                         self.fetch_missing_csize(decrypted_repository, archive_chunk_idx, cache_sync_stats)
```
ThomasWaldmann: hm? above looks rather like an equivalent functionality cleanup for an older optimization than saving 1 merge call by it.

src/borg/constants.py
```diff
@@ -8,7 +8,9 @@
 REQUIRED_ITEM_KEYS = frozenset(['path', 'mtime', ])
 
 # this set must be kept complete, otherwise rebuild_manifest might malfunction:
-ARCHIVE_KEYS = frozenset(['version', 'name', 'items', 'cmdline', 'hostname', 'username', 'time', 'time_end',
+ARCHIVE_KEYS = frozenset(['version', 'name', 'items', 'num_files',
+                          'cmdline', 'hostname', 'username',
+                          'time', 'time_end',
                           'comment', 'chunker_params',
                           'recreate_cmdline', 'recreate_source_id', 'recreate_args'])
```
ThomasWaldmann: 3 recreate* keys

src/borg/item.pyx
```diff
@@ -289,16 +289,14 @@ class ArchiveItem(PropDict):
     If a ArchiveItem shall be serialized, give as_dict() method output to msgpack packer.
     """
 
-    VALID_KEYS = {'version', 'name', 'items', 'cmdline', 'hostname', 'username', 'time', 'time_end',
-                  'comment', 'chunker_params',
-                  'recreate_cmdline', 'recreate_source_id', 'recreate_args', 'recreate_partial_chunks',
```
ThomasWaldmann: 4 recreate* keys

src/borg/constants.py
enkore: This one is correct, the other is wrong.

enkore:
Since this is mostly testability and cleanup, and it seems that the current code is not broken, I likely won't update this for 1.1. Re-roll when I have more time, which probably means 1.2.

enkore:
b39187fd88 could be picked into master in a separate PR I guess, since it'd help in the future with `archive: use chunks archive to avoid reading data from repository` for `borg info`.

ThomasWaldmann:
yup, i agree about b39187f, good idea.

PR #2768, opened 2017-06-30
docs: no third level toc on command usage pages ("Description", "Examples")

enkore:
also fixes missing trailing newlines of the .rst.inc files.

### DIFF
--- a/docs/_templates/globaltoc.html
+++ b/docs/_templates/globaltoc.html
@@ -1,6 +1,16 @@
 <div class="sidebar-block">
   <div class="sidebar-toc">
-    {% set toctree = toctree(maxdepth=3, collapse=True, includehidden=True) %}
+    {# Restrict the sidebar toc depth to two levels while generating command usage pages.
+       This avoids superfluous entries for each "Description" and "Examples" heading. #}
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -281,6 +281,17 @@ def generate_level(self, prefix, parser, Archiver):
...
+                    doc.write(textwrap.dedent("""
+
+                    .. raw:: html
+
+                        <script type='text/javascript'>
+                        $(document).ready(function () {
+                            /* Remove the "Description" and "Examples" level three entries from the global toc */
+                            $('.toctree-l2 ul').remove();
+                        })
+                        </script>
```
ThomasWaldmann: is there no way to avoid them in the first place?

enkore: Turns out there is. Not really a "native" way, but a bit cleaner than adding a bit of JS.

The reason this is so annoying to do is that the maxdepth parameter of the toctree() template function defaults to the maxdepth of the toctree. The *root* toctree, that is, not the *nearest* toctree relative to the current page. And that's two, and the reason why I put three in there. However, Sphinx gives us `pagename` so a little if is all that is needed to fix that up.

PR #2778, opened 2017-07-02
fuse: versions view, linear numbering by archive time

enkore:
The nested directories feel rather unergonomic to me.

This seems to regress about ~10 % compared to master. The effect of keep-exts seems similar.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -9,7 +9,6 @@ import time
 from collections import defaultdict
 from signal import SIGINT
 from distutils.version import LooseVersion
-from zlib import adler32
 
 import llfuse
 import msgpack
@@ -17,7 +16,9 @@ import msgpack
 from .logger import create_logger
 logger = create_logger()
 
+from .crypto.low_level import blake2b_128
 from .archive import Archive
+from .hashindex import FuseVersionsIndex
 from .helpers import daemonize, hardlinkable, signal_handler, format_file_size
 from .item import Item
 from .lrucache import LRUCache
@@ -240,13 +241,14 @@ class FuseOperations(llfuse.Operations):
         if self.args.location.archive:
             self.process_archive(self.args.location.archive)
         else:
+            self.versions_index = FuseVersionsIndex()
             archive_names = (x.name for x in self.manifest.archives.list_considering(self.args))
             for archive_name in archive_names:
                 if self.versions:
                     # process archives immediately
                     self.process_archive(archive_name)
                 else:
-                    # lazy load archives, create archive placeholder inode
+                    # lazily load archives, create archive placeholder inode
                     archive_inode = self._create_dir(parent=1)
                     self.contents[1][os.fsencode(archive_name)] = archive_inode
                     self.pending_archives[archive_inode] = archive_name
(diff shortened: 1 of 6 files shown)


src/borg/fuse.py
```diff
@@ -339,12 +341,19 @@ def process_archive(self, archive_name, prefix=[]):
...
-                for chunkid, _, _ in item.chunks:
-                    ident = adler32(chunkid, ident)
-                return ident
+                file_id = blake2b_128(path)
+                current_version, previous_id = self.versions_index.get(file_id, (0, None))
+
+                chunk_ids = [chunk_id for chunk_id, _, _ in item.chunks]
+                contents_id = blake2b_128(b''.join(chunk_ids))
+
+                if contents_id != previous_id:
```
ThomasWaldmann: hmm, this behaves a bit different than before. same content can appear multiple times.
if a file content changes from A to B to A, you'll have 3 items 0,1,2 in fuse.

can be seen as a feature, depending on what one wants.

src/borg/fuse.py
```diff
@@ -339,12 +341,19 @@ def process_archive(self, archive_name, prefix=[]):
...
             if 'chunks' in item:
-                ident = 0
-                for chunkid, _, _ in item.chunks:
-                    ident = adler32(chunkid, ident)
-                return ident
+                file_id = blake2b_128(path)
+                current_version, previous_id = self.versions_index.get(file_id, (0, None))
+
+                chunk_ids = [chunk_id for chunk_id, _, _ in item.chunks]
+                contents_id = blake2b_128(b''.join(chunk_ids))
```
ThomasWaldmann: how about not creating the chunk_ids list, but just an iterator that is used in the lower line?

ThomasWaldmann:
so, regarding the top post, do you also want to remove the containing directory layer?

guess for a few versions it has better ergonomics, but for many versions (and many files) it can quickly escalate to bad ergonomics and huge directories.

src/borg/fuse.py
enkore: That's strictly the same, because join() will turn any iterable into a list first (otherwise it could not be O(n) for iterators).

ThomasWaldmann:
your --last change should be a separate PR.

src/borg/fuse.py
enkore: Yes, that's intentional.

ThomasWaldmann: yeah, guess it is a feature, because it better reflects change history / order.

ThomasWaldmann: hmm, one issue though: numbers may depend on mount options.

e.g. using borg mount --last N -o versions will use different numbers when N changes.

so, this has same issue as when using numbers for VCS versions compared to stable hashes.

enkore: An alternative could be to just make the names `archive.ext`, since they are in an extra directory anyway.

enkore: And / for archives was already decidedly ruled out.

enkore: However, then every archive will show a separate version (otherwise it again starts to make little sense), even if the file did not actually change.

ThomasWaldmann: putting (potentially long) archive names into the filename seems weird.

but we could build a structure like this:
```
path/filename.ext/filename.N.ext --> ../.../archivename/path/filename.ext
```
- less ugly filenames
- same "issue" with showing all version
- added benefit to be able to find into the corresponding archive via filesystem.

although using a FILE symlink is maybe not most comfortable, so maybe we would have additionally a symlink to the parent directory in the respective archive.

enkore: Thinking about it, anything that doesn't collapse identical instances is pretty much pointless (going from repository view to this view is trivial and will be in most cases just a `**/` away). If identical instances are collapsed, then there is no way to *directly* tell the archive.

ThomasWaldmann: it depends on usecase.

if it is about finding the right version of a single file, then yes, showing all versions is pointless.

if it deals with finding something more complex, like a combination of files, having them all has a point.

enkore: tree: all_versions; https://github.com/enkore/borg/tree/f/fuse-all_versions

Should not be merged, just an unnecessary option.

PR #2779, opened 2017-07-02
docs: mount: show usage of archive filters

enkore:
this is pretty vital, since the versions view is basically
useless without them.

### DIFF
--- a/docs/usage/mount.rst
+++ b/docs/usage/mount.rst
@@ -31,6 +31,11 @@ Examples
     -rw-rw-r-- 1 user group 12204 Aug 26 21:04 doc.fa760f28.txt
     $ borg umount /tmp/mymountpoint
 
+    # Archive filters are supported.
+    # These are especially handy for the "versions view",
+    # which does not support lazy processing of archives.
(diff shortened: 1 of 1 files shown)


docs/usage/mount.rst
```diff
@@ -31,6 +31,11 @@ Examples
     -rw-rw-r-- 1 user group 12204 Aug 26 21:04 doc.fa760f28.txt
     $ borg umount /tmp/mymountpoint
 
+    # Archive filters are supported.
+    # These are especially handy for the "versions view",
+    # which does not support lazy processing of archives.
+    $ borg mount -o versions --glob-archives '*-my-home' --last 10 /path/to/repo /tmp/mymountpoint
```
ThomasWaldmann: maybe use root-* to get it consistent with the above examples (15, 19).

enkore: No, for one thing the root example set up doesn't really make sense to introduce this option (it already showed that there are only two archives, here I intentionally elide that), and it's good to have some diversity in the examples anyway, to show the different options that exist.

ThomasWaldmann: for beginners with a software, consistency is more important than "diversity" (which might rather confuse them).

PR #2780, opened 2017-07-02
fix reversed archive ordering with --last

enkore:

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -219,11 +219,14 @@ class Archives(abc.MutableMapping):
         name = safe_encode(name)
         del self._archives[name]
 
-    def list(self, sort_by=(), reverse=False, glob=None, first=None, last=None):
+    def list(self, *, glob=None, sort_by=(), first=None, last=None, reverse=False):
         """
-        Inexpensive Archive.list_archives replacement if we just need .name, .id, .ts
-        Returns list of borg.helpers.ArchiveInfo instances.
-        sort_by can be a list of sort keys, they are applied in reverse order.
+        Return list of ArchiveInfo instances according to the parameters.
+
+        First match *glob*, then *sort_by*. Apply *first* and *last* filters,
+        and possibly *reverse* the list.
+
+        *sort_by* is a list of sort keys applied in reverse order.
         """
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
we better check whether this global change has unintended effects, e.g. compared to 1.0 or concerning performance of some ops.

src/borg/helpers.py
```diff
@@ -231,10 +231,13 @@ def list(self, sort_by=(), reverse=False, glob=None, first=None, last=None):
...
             archives.sort(key=attrgetter(sortkey))
-        if reverse or last:
+        if reverse:
             archives.reverse()
-        n = first or last or len(archives)
-        return archives[:n]
+        if first:
+            archives = archives[:first]
+        elif last:
+            archives = archives[len(archives) - last:]
```
ThomasWaldmann: if last is bigger than len(archives), this will get negative and will malfunction.

enkore:
1.0 didn't have this globally.

src/borg/helpers.py
enkore: Good point.

ThomasWaldmann:
yes, but for some commands.

enkore:
Only check had that and ordering doesn't matter there for functionality.

src/borg/helpers.py
```diff
@@ -231,10 +231,13 @@ def list(self, sort_by=(), reverse=False, glob=None, first=None, last=None):
...
-        if reverse or last:
+        if reverse:
             archives.reverse()
-        n = first or last or len(archives)
-        return archives[:n]
+        if first:
+            archives = archives[:first]
+        elif last:
+            archives = archives[max(len(archives) - last, 0):]
+        return archives
```
ThomasWaldmann: hmm, reverse() now needs to be done after first/last. otherwise it will turn last into first and vice versa.

enkore: Yeah, the only ordering that actually makes sense is match-sort-firstlast-reverse.

ThomasWaldmann: i shortly considered whether applying both first and last (or last and then first) makes sense, but guess there is no use case for that.

PR #2784, opened 2017-07-02
prune enhancements / fixes

ThomasWaldmann:
- noticed that prune does own globbing, which Archives.list can do.
- also, checkpoint handling with glob option is broken there, because the checkpoints won't be in the list due to shellpatterns doing fullmatches, so the checkpoints are not in the list.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1476,8 +1476,11 @@ class Archiver:
             return self.exit_code
         if args.prefix:
             args.glob_archives = args.prefix + '*'
-        archives_checkpoints = manifest.archives.list(glob=args.glob_archives, sort_by=['ts'], reverse=True)
-        is_checkpoint = re.compile(r'\.checkpoint(\.\d+)?$').search
+        checkpoint_re = r'\.checkpoint(\.\d+)?'
+        archives_checkpoints = manifest.archives.list(glob=args.glob_archives,
+                                                      match_end=r'(%s)?\Z' % checkpoint_re,
+                                                      sort_by=['ts'], reverse=True)
+        is_checkpoint = re.compile(r'(%s)\Z' % checkpoint_re).search
         checkpoints = [arch for arch in archives_checkpoints if is_checkpoint(arch.name)]
         # keep the latest checkpoint, if there is no later non-checkpoint archive
         if archives_checkpoints and checkpoints and archives_checkpoints[0] is checkpoints[0]:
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
note: prune is ok in 1.0-maint (no --glob-archives, just --prefix, matching done with .startswith).
if needed, we could backport the shellpattern change.

src/borg/archiver.py
```diff
@@ -1474,12 +1474,10 @@ def do_prune(self, args, repository, manifest, key):
...
                              '"keep-weekly", "keep-monthly" or "keep-yearly" settings must be specified.')
             return self.exit_code
-        archives_checkpoints = manifest.archives.list(sort_by=['ts'], reverse=True)  # just a ArchiveInfo list
         if args.prefix:
             args.glob_archives = args.prefix + '*'
-        if args.glob_archives:
-            regex = re.compile(shellpattern.translate(args.glob_archives))
-            archives_checkpoints = [arch for arch in archives_checkpoints if regex.match(arch.name) is not None]
+        archives_checkpoints = manifest.archives.list(glob=args.glob_archives, match_end=False,
+                                                      sort_by=['ts'], reverse=True)
```
enkore: I think it'd be better to write this more along the lines of `if name.endswith('.checkpoint'): regex.match(name[:-len('.checkpoint')]) else: regex.match(name)`, because this way it effectively means `...*`, which includes more than just checkpoints.

ThomasWaldmann: Hmm, yeah. Guess I rather had --prefix in mind, not so much the more precise --glob-archives kind of matching. I'll work on it...

ThomasWaldmann:
made the shellpattern match_end more flexible, so it can be used for optional checkpoint ending.

note: at first glance, the code in do_prune now looks like it could be even simplified more, like getting the archives and the checkpoints via 2 calls to list().

but the code needs both together, sorted from new to old, in archives_checkpoints, so it can check whether the newest entry is a checkpoint (which we want to keep if we do not have a new normal archive).

src/borg/shellpattern.py
```diff
@@ -2,14 +2,16 @@
...
-def translate(pat):
+def translate(pat, match_end=r"\Z"):
     """Translate a shell-style pattern to a regular expression.
 
     The pattern may include ``**<sep>`` (<sep> stands for the platform-specific path separator; "/" on POSIX systems) for
     matching zero or more directory levels and "*" for matching zero or more arbitrary characters with the exception of
     any path separator. Wrap meta-characters in brackets for a literal match (i.e. "[?]" to match the literal character
     "?").
 
+    match_end=True will create a regex that has a \Z to match the end of the string. False: won't.
```
enkore: doc out of date

ThomasWaldmann:
merge?

Issue #2738, opened 2017-06-23
review msgpack exceptions


ThomasWaldmann:
seems like at some places in the code, we just catch msgpack.UnpackException, but not ValueError which seems raised by msgpack code in same situation.

ThomasWaldmann:
http://msgpack-python.readthedocs.io/en/latest/api.html#exceptions

Is it just me or are these deprecations backwards?

PR #2786, opened 2017-07-03
WIP: review msgpack exceptions

ThomasWaldmann:
#2738


enkore:
The best way forward would be tests with corrupted data to exercise the error handling paths, from my PoV.

ThomasWaldmann:
If the msgpack author can't be convinced otherwise, we'll have to go down the try/except Exception route. :-(

ThomasWaldmann:
not merging this, will soon update the other PR with cleaned up exception handling.

PR #2790, opened 2017-07-04
use limited unpacker for outer key, fixes #2174

### LINKED ISSUE #2174: Limited unpacker for outer key packing
We should also use a limited unpacker for the outer key packing (`KeyfileKeyBase.decrypt_key_file`), since it is unpacked before verifying the contents. This one is easy since it's a very simple, known, unchanging structure there.

Split from #2139

ThomasWaldmann:


src/borg/helpers.py
```diff
@@ -169,8 +169,16 @@ def get_limited_unpacker(kind):
                          object_hook=StableDict,
                          unicode_errors='surrogateescape',
                          ))
+    elif kind == 'key':
+        args.update(dict(use_list=True,  # default value
+                         max_array_len=0,  # not used
+                         max_map_len=10,  # EncryptedKey dict
+                         max_str_len=1000,  # inner key data
```
enkore: 4K or so (the future awaits)

ThomasWaldmann: done.

ThomasWaldmann:
collapse / merge?

PR #2791, opened 2017-07-04
WIP dash_open

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -1752,11 +1748,8 @@ def do_debug_dump_manifest(self, args, repository, manifest, key):
 
         meta = prepare_dump_dict(msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
 
-        if args.path == '-':
-            json.dump(meta, sys.stdout, indent=4)
-        else:
-            with open(args.path, 'w') as fd:
-                json.dump(meta, fd, indent=4)
+        with dash_open(args.path, 'w') as fd:
+            json.dump(meta, fd, indent=4)
```
ThomasWaldmann: hmm, guess sys.stdout.buffer is closed on CM exit, is that correct behaviour?

src/borg/helpers.py
```diff
@@ -2178,12 +2178,12 @@ def popen_with_error_handling(cmd_line: str, log_prefix='', **kwargs):
...
+def dash_open(path, mode):
     if path == '-':
-        if 'b' in mode:
-            return sys.stdin.buffer
-        else:
-            return sys.stdin
+        stream = sys.stdin if 'r' in mode else sys.stdout
+        # if sys.std(in|out) got replaced by something else, we maybe have no .buffer
+        use_buffer = 'b' in mode and hasattr(stream, 'buffer')
+        return stream.buffer if use_buffer else stream
```
enkore: This looks like it'll be mixing text and binary IO. Rather fix whoever in the testsuite replaced stdin/stdout with incomplete mocks?

src/borg/helpers.py
```diff
@@ -2178,12 +2178,12 @@ def popen_with_error_handling(cmd_line: str, log_prefix='', **kwargs):
         return
 
 
-def open_file_or_stdin(path, mode):
+def dash_open(path, mode):
     if path == '-':
```
enkore: assert '+' not in mode?

src/borg/helpers.py
```diff
@@ -2178,12 +2178,12 @@ def popen_with_error_handling(cmd_line: str, log_prefix='', **kwargs):
         return
 
 
-def open_file_or_stdin(path, mode):
+def dash_open(path, mode):
```
enkore: assert '+' not in mode?

src/borg/archiver.py
enkore: Not if you wanted to do anything with stdout after that...

src/borg/helpers.py
ThomasWaldmann: ok

src/borg/helpers.py
ThomasWaldmann: ok, the alternative is to just let it blow up when it tries to access .buffer.
but guess that is better than silently causing a mix up of binary / text.

enkore: When binary data is written to e.g. StringIO it blows up as well — blowing up early is preferable.

src/borg/archiver.py
ThomasWaldmann: instead of just offering dash_open function, we could have a full context manager that closes real files, but lets the streams open on exit.

ThomasWaldmann: otoh, that difference seems a bit strange. if we would close the real file, why would we keep stdout open and write more stuff to it (that would not be in the file)?

ThomasWaldmann:
merge or get the contextmanager version that allows to customize exit behaviour (close or not close streams)?

PR #2792, opened 2017-07-04
support borg list repo --format {comment}, fixes #2081

### LINKED ISSUE #2081: borg list: make it possible to list archive comments as well
Add optional backup log text into borg create command.

Example

`usage: borg create [-h] [--critical] [--error] [--warning] [--info] [--debug]
                   [--lock-wait N] [--show-rc] [--no-files-cache] [--umask M]
                   [--remote-path PATH] [-s] [-p] [--list]
                   [--filter STATUSCHARS] [-e PATTERN]
                   [--exclude-from EXCLUDEFILE] [--exclude-caches]
                   [--exclude-if-present FILENAME] [--keep-tag-files]
                   [-c SECONDS] [-x] [--numeric-owner] [--noatime] [--noctime]
                   [--timestamp yyyy-mm-ddThh:mm:ss]
                   [--chunker-params CHUNK_MIN_EXP,CHUNK_MAX_EXP,HASH_MASK_BITS,HASH_WINDOW_SIZE]
                   [--ignore-inode] [-C COMPRESSION] [--read-special] [-n]
                   [--log Log text ending with characters --]
                   ARCHIVE PATH [PATH ...]

--- or ---

usage: borg create [-h] [--critical] [--error] [--warning] [--info] [--debug]
                   [--lock-wait N] [--show-rc] [--no-files-cache] [--umask M]
                   [--remote-path PATH] [-s] [-p] [--list]
                   [--filter STATUSCHARS] [-e PATTERN]
                   [--exclude-from EXCLUDEFILE] [--exclude-caches]
                   [--exclude-if-present FILENAME] [--keep-tag-files]
                   [-c SECONDS] [-x] [--numeric-owner] [--noatime] [--noctime]
                   [--timestamp yyyy-mm-ddThh:mm:ss]
                   [--chunker-params CHUNK_MIN_EXP,CHUNK_MAX_EXP,HASH_MASK_BITS,HASH_WINDOW_SIZE]
                   [--ignore-inode] [-C COMPRESSION] [--read-special] [-n]
                   [--log "Log text between quotation marks"]
                   ARCHIVE PATH [PATH ...]
`

Log text is a free text describing the content of the backup. It is shown as last column when running borg list command like this:

borg list /home/foo/bbckpdata/MyBackup
documents--2016-12-21_17:55:50        Wed, 2016-12-21 17:55:54  Initial backup
documents--2016-12-21_18:26:07        Wed, 2016-12-21 18:26:11  Before huge change xxx.
documents--2016-12-21_18:35:08        Wed, 2016-12-21 18:35:08  After adding new content from yyy.
presentations--2016-12-21_19:07:56    Wed, 2016-12-21 19:07:56  Before major change zzz.
presentations--2016-12-21_19:30:07    Wed, 2016-12-21 19:30:07  Daily backup.

Purpose of this log text is to more easily find the correct archive when restore is needed or give some idea what archives can be pruned away.

ThomasWaldmann:
Load full-blown Archive instance if required.

Also supported: {bcomment} for binary and {end}.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -291,7 +291,9 @@ class Archive:
         self.hard_links = {}
         self.stats = Statistics(output_json=log_json)
         self.show_progress = progress
-        self.name = name
+        self.name = name  # overwritten later with name from archive metadata
+        self.name_in_manifest = name  # can differ from .name later (if borg check fixed duplicate archive names)
+        self.comment = None
         self.checkpoint_interval = checkpoint_interval
         self.numeric_owner = numeric_owner
         self.noatime = noatime
@@ -340,6 +342,7 @@ class Archive:
         self.metadata = self._load_meta(self.id)
         self.metadata.cmdline = [safe_decode(arg) for arg in self.metadata.cmdline]
         self.name = self.metadata.name
+        self.comment = self.metadata.get('comment', '')
 
     @property
     def ts(self):
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1323,7 +1323,7 @@ class Archiver:
             if args.json_lines:
                 self.print_error('The --json-lines option is only valid for listing archive contents, not archives.')
                 return self.exit_code
-            return self._list_repository(args, manifest, write)
+            return self._list_repository(args, repository, manifest, key, write)
 
     def _list_archive(self, args, repository, manifest, key, write):
         matcher = self.build_matcher(args.patterns, args.paths)
@@ -1351,14 +1351,14 @@ class Archiver:
 
         return self.exit_code
 
-    def _list_repository(self, args, manifest, write):
+    def _list_repository(self, args, repository, manifest, key, write):
         if args.format is not None:
             format = args.format
         elif args.short:
             format = "{archive}{NL}"
         else:
             format = "{archive:<36} {time} [{id}]{NL}"
-        formatter = ArchiveFormatter(format)
+        formatter = ArchiveFormatter(format, repository, manifest, key, json=args.json)
 
         output_data = []
 
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -1581,28 +1581,107 @@ class BaseFormatter:
 
 
 class ArchiveFormatter(BaseFormatter):
+    KEY_DESCRIPTIONS = {
+        'name': 'archive name interpreted as text (might be missing non-text characters, see barchive)',
+        'archive': 'archive name interpreted as text (might be missing non-text characters, see barchive)',
+        'barchive': 'verbatim archive name, can contain any character except NUL',
+        'comment': 'archive comment interpreted as text (might be missing non-text characters, see bcomment)',
+        'bcomment': 'verbatim archive comment, can contain any character except NUL',
+        'time': 'time (start) of creation of the archive',
+        # *start* is the key used by borg-info for this timestamp, this makes the formats more compatible
+        'start': 'time (start) of creation of the archive',
+        'end': 'time (end) of creation of the archive',
+        'id': 'internal ID of the archive',
+    }
+    KEY_GROUPS = (
+        ('name', 'archive', 'barchive', 'comment', 'bcomment', 'id'),
+        ('time', 'start', 'end'),
+    )
 
-    def __init__(self, format):
-        self.format = partial_format(format, self.FIXED_KEYS)
+    @classmethod
+    def available_keys(cls):
+        fake_archive_info = ArchiveInfo('archivename', b'\1'*32, datetime(1970, 1, 1, tzinfo=timezone.utc))
+        formatter = cls('', None, None, None)
+        keys = []
+        keys.extend(formatter.call_keys.keys())
+        keys.extend(formatter.get_item_data(fake_archive_info).keys())
+        return keys
 
-    def get_item_data(self, archive):
-        return {
-            # *name* is the key used by borg-info for the archive name, this makes the formats more compatible
-            'name': remove_surrogates(archive.name),
-            'barchive': archive.name,
-            'archive': remove_surrogates(archive.name),
-            'id': bin_to_hex(archive.id),
-            'time': format_time(to_localtime(archive.ts)),
-            # *start* is the key used by borg-info for this timestamp, this makes the formats more compatible
-            'start': format_time(to_localtime(archive.ts)),
+    @classmethod
+    def keys_help(cls):
+        help = []
+        keys = cls.available_keys()
+        for key in cls.FIXED_KEYS:
+            keys.remove(key)
+
+        for group in cls.KEY_GROUPS:
+            for key in group:
+                keys.remove(key)
+                text = "- " + key
+                if key in cls.KEY_DESCRIPTIONS:
+                    text += ": " + cls.KEY_DESCRIPTIONS[key]
+                help.append(text)
+            help.append("")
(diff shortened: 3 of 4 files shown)


src/borg/archive.py
```diff
@@ -291,7 +291,9 @@ def __init__(self, repository, key, manifest, name, cache=None, create=False,
         self.hard_links = {}
         self.stats = Statistics(output_json=log_json)
         self.show_progress = progress
-        self.name = name
+        self.name = name  # overwritten later with name from archive metadata
+        self.name_in_manifest = name  # can differ from .name later (if borg check fixed duplicate archive names)
```
ThomasWaldmann: adding this was needed because a test was broken without it.
borg list needs to list names from manifest, not names from archive metadata.

src/borg/archiver.py
```diff
@@ -1363,10 +1363,11 @@ def _list_repository(self, args, manifest, write):
         output_data = []
 
         for archive_info in manifest.archives.list_considering(args):
+            archive = Archive(repository, key, manifest, archive_info.name)
```
ThomasWaldmann: this is a bit more expensive now as it load the Archive metadata, not just the manifest.

only matters for huge numbers of archives and in that case I guess there are bigger performance problems elsewhere.

src/borg/helpers.py
```diff
@@ -1588,9 +1588,11 @@ def __init__(self, format):
     def get_item_data(self, archive):
```
ThomasWaldmann: note: archive is now an Archive instance, but used to be only an "archive info" (manifest entry).

src/borg/archiver.py
enkore: This is problematic, because it is synchronous. So listing 1000 archives now requires 1000 round-trips to the repository. I.e. about a minute for over the Internet.

This needs to be conditional on whether the comment formatters are used. See `format_needs_cache` (`format_keys = {f[1] for f in Formatter().parse(format)}`).

enkore: (Depending on how large the archives are: A couple dozen kB times a few hundred to thousand = a couple dozen MB download for *each* borg list, independent of the RTT issue).

enkore: You could also use the call keys concept from ItemFormatter to fix this; would avoid Archive vs. ArchiveInfo depending on branches.

ThomasWaldmann: i tried the first way, but it gets rather messy as ArchiveInfo is quite different from Archive.

trying call keys concept now...

src/borg/helpers.py
```diff
@@ -1581,32 +1581,104 @@ def keys_help():
...
-            'id': bin_to_hex(archive.id),
-            'time': format_time(to_localtime(archive.ts)),
-            # *start* is the key used by borg-info for this timestamp, this makes the formats more compatible
-            'start': format_time(to_localtime(archive.ts)),
+    @classmethod
+    def format_needs_archive(cls, format):
+        format_keys = {f[1] for f in Formatter().parse(format)}
+        return any(key in cls.KEYS_REQUIRING_ARCHIVE for key in format_keys)
+
+    def __init__(self, format, repository, manifest, key, *, json_lines=False):
```
enkore: json_lines unused

src/borg/helpers.py
```diff
@@ -1581,32 +1581,104 @@ def keys_help():
...
+        self.call_keys = {
+            'comment': partial(self.get_comment, rs=True),
+            'bcomment': partial(self.get_comment, rs=False),
         }
+        self.used_call_keys = set(self.call_keys) & self.format_keys
+        if self.json_lines:
+            self.item_data = {}
+            self.format_item = self.format_item_json
+        else:
+            self.item_data = static_keys
```
enkore: dito

src/borg/helpers.py
```diff
@@ -1581,32 +1581,104 @@ def keys_help():
...
-    @staticmethod
-    def keys_help():
-        return "- archive, name: archive name interpreted as text (might be missing non-text characters, see barchive)\n" \
-               "- barchive: verbatim archive name, can contain any character except NUL\n" \
-               "- comment: archive comment interpreted as text (might be missing non-text characters, see bcomment)\n" \
-               "- bcomment: verbatim archive comment, can contain any character except NUL\n" \
-               "- time: time of creation of the archive\n" \
-               "- id: internal ID of the archive"
+    def format_item_json(self, item):
+        return json.dumps(self.get_item_data(item)) + '\n'
```
enkore: dito

src/borg/helpers.py
ThomasWaldmann: hmm, sounds like --json-lines is only wanted for ItemFormatter, but not for ArchiveFormatter?

ThomasWaldmann: so, what did you mean? remove --json-lines for repos or support it?

enkore: remove

src/borg/archive.py
enkore: This seems to be not used any more; remove it, or leave it for later?

src/borg/helpers.py
```diff
@@ -1581,28 +1581,117 @@ def keys_help():
...
-            'barchive': archive.name,
-            'archive': remove_surrogates(archive.name),
-            'id': bin_to_hex(archive.id),
-            'time': format_time(to_localtime(archive.ts)),
-            # *start* is the key used by borg-info for this timestamp, this makes the formats more compatible
-            'start': format_time(to_localtime(archive.ts)),
+    @classmethod
+    def format_needs_archive(cls, format):
+        format_keys = {f[1] for f in Formatter().parse(format)}
+        return any(key in cls.KEYS_REQUIRING_ARCHIVE for key in format_keys)
```
enkore: unused

src/borg/helpers.py
```diff
@@ -1581,28 +1581,117 @@ def keys_help():
...
+    }
+    KEY_GROUPS = (
+        ('name', 'archive', 'barchive', 'comment', 'bcomment', 'id'),
+        ('time', 'start', 'end'),
+    )
+
+    KEYS_REQUIRING_ARCHIVE = (
+        'comment', 'bcomment',
+        'end',
+    )
```
enkore: unused

src/borg/archive.py
ThomasWaldmann: yes, i noticed also it is not actively used.

i wanted to keep it as it points out the issue with different names in manifest and in archive.

src/borg/helpers.py
ThomasWaldmann: yeah, i also noticed that the lazy loading of archive works without this, automagically.

i'll remove it (and also i'll have a look at ItemFormatter, whether the similar mechanism there can't be replaced in the same way).

ThomasWaldmann:
anything else? otherwise I'ld collapse some changesets and merge...

enkore:
c981d2d is btw. not a fix, strictly speaking.

ThomasWaldmann:
@enkore call it "defensive code": it just works, no matter how the caller uses the returned objects.

I noticed that the way it is used by the caller works also with the pre-changeset way, but when used a bit differently (see previous changesets, when I copied that code to create ArchiveFormatter and used it a bit differently), it breaks and creates a list of N identical objects, because it always used the same dict object and just patched the updated members into it. That might be a bit faster, but is a pitfall somehow.

ThomasWaldmann:
collapsed some changesets, reworded one commit comment, no other changes.

PR #2793, opened 2017-07-04
use commit() in borg with-lock, fixes #1867

### LINKED ISSUE #1867: Does borg do crucial modifications in append-only mode?
I'm planning on doing a backup in borg on a server that is configured to `append-only`.
The backup from this server will then be cloned to a second (off-site) server regularly, using rsync with `ignore-existing`. The aim of this is to keep a copy all files, even if they get changed/deleted on the local backup server.

So, in my scenario, any file will only be synced once and all modifications or deletions won't get synced with the remote server. Will this cause any trouble for borg? I understand that the transaction log won't be synced, but I can live with that if it doesn't cause any problems for backup integrity.

ThomasWaldmann:


ThomasWaldmann:
merge?

enkore:
It'd be neater to fix/nudge rollback instead, but that requires far more review and testing, so lgtm.

Issue #2800, opened 2017-07-05
Confusing advice regarding --ignore-inode command line flag


palant:
The documentation currently says:

> To speed up pulling backups over sshfs and similar network file systems which do not provide correct inode information the –ignore-inode flag can be used. This potentially decreases reliability of change detection, while avoiding always reading all files on these file systems.

I tried to understand how this actually works:

* Inode information is used to avoid reading and hashing files that didn't change since the previous backup.
* This information could be wrong when backing up files from sshfs and similar network file systems, so that some files might be mistakenly considered unchanged, and some probably marked as changed despite not having been changed (is that actually an issue?).
* In order to make change detection work reliably, `--ignore-inode` flag can be specified. This will cause all files to be read in on backup which should improve consistency at the cost of performance.

Did I get this correctly? If so, it doesn't match the documentation at all, I can suggest a better formulation then.

enkore:
> This information could be wrong when backing up files from sshfs and similar network file systems, so that some files might be mistakenly considered unchanged, and some probably marked as changed despite not having been changed (is that actually an issue?).

It's the other way around ("potentially decreases reliability").

I assume the "this" in "This potentially decreases reliability" makes this paragraph very ambiguous? How about "Enabling this flag ..."?

palant:
I looked it up in the source code and apparently I got this wrong. With the `--ignore-inode` flag change detection happens by size and modification time only. So the confusing parts here are "pulling backup over" (I initially associated that with a restore operation, and didn't realize that sshfs is the location of the files being backed up, not the repository) and "potentially decreases reliability" being not very informative. Finally, the context (unchanged files being skipped for incremental backups) is missing.

How about:

> Backup speed is increased by not reprocessing files that are already part of existing archives and weren't modified. Normally, detecting file modifications will take inode information into consideration. This is problematic for files located on sshfs and similar network file systems which do not provide correct inode information, such files will always be considered modified. The –ignore-inode flag can be used to prevent this and improve performance. This flag will reduce reliability of change detection however, with files considered unmodified as long as their size and modification time are unchanged.

ThomasWaldmann:
Guess that's better, just s/correct inode information,/stable inode numbers -/.

palant:
That part was actually copied from the current documentation :smile:

ThomasWaldmann:
Can you do a PR?

palant:
There you go, created #2802.

PR #2802, opened 2017-07-05
#2800 - Better explanation of the --ignore-inode option relevance

palant:
See #2800 for explanation and discussion.

### DIFF
--- a/docs/usage/create.rst.inc
+++ b/docs/usage/create.rst.inc
@@ -156,14 +156,10 @@ checkpoints and treated in special ways.
 In the archive name, you may use the following placeholders:
 {now}, {utcnow}, {fqdn}, {hostname}, {user} and some others.
 
-Backup speed is increased by not reprocessing files that are already part of
-existing archives and weren't modified. Normally, detecting file modifications
-will take inode information into consideration. This is problematic for files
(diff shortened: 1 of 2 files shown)


enkore:
Please do the editing in src/borg/archiver.py; this file is automatically generated ("
.. IMPORTANT: this file is auto-generated from borg's built-in help, do not edit!")

palant:
I assume that I would still have to run docs generation so that `docs/` directory doesn't get out of sync. Am I stupid if I cannot find the mechanism doing that?

enkore:
That's documented [here](http://borgbackup.readthedocs.io/en/latest/development.html#documentation). It is not required to re-generate these files with every edit, since that's always done as part of the [release preparations](http://borgbackup.readthedocs.io/en/latest/development.html#creating-a-new-release).

palant:
> It is not required to re-generate these files with every edit, since that's always done as part of the release preparations.

Wow, so it's duplicated and outdated content being kept in the repository. Sounds like a footgun.

ThomasWaldmann:
@palant The basic problem you address is known and iirc we tried to do it otherwise, but there was some issue which I do not do remember right now.

PR #2804, opened 2017-07-08
update CHANGES (master)

ThomasWaldmann:


docs/changes.rst
```diff
@@ -134,6 +134,76 @@ Changelog
...
+- key import: allow reading from stdin, #2760
+
+Fixes:
+
+- use commit() in borg with-lock, #1867
+- prune: fix checkpoints processing with --glob-archives
+- FUSE: versions view: keep original file extension at end, #2769
+- fix --last, --first: do not accept values <= 0,
+  fix reversed archive ordering with --last
+- include attic.tar.gz when installing the package
```
enkore: include testsuite data (attic.tar.gz) when ...

docs/changes.rst
```diff
@@ -134,6 +134,76 @@ Changelog
...
+- delete: removed short option for --cache-only
+
+New features:
+
+- support borg list repo --format {comment} {bcomment} {end}, #2081
+- key import: allow reading from stdin, #2760
+
+Fixes:
+
+- use commit() in borg with-lock, #1867
```
enkore: with-lock: avoid creating segment files that might be overwritten later, #...

docs/changes.rst
```diff
@@ -134,6 +134,76 @@ Changelog
...
+    - add test for paperkey import, fix bug
+- vagrant:
+
+  - fix set-up of cygwin and its sshd
+  - add Debian 9 "stretch" 64bit
+- travis:
+
+  - don't perform full travis build on docs-only changes, #2531
+  - install fakeroot for Linux (was removed from preinstalled sw stack)
+
```
enkore: A more general question, the changelogs are always quite large, so is it worth it to include things like CI or vagrant changes, or mentioning that a single unit test was added? (Or the many items under "docs:" that basically mean "the formatting changed a bit")

For example, I'd just summarize the `tests:` section with "Added a number of additional unit tests". I doubt anyone cares for more details in the changelog.

Stuff like CI changes are basically unrelated to releases as well.

Issue #2815, opened 2017-07-14
Only one – hyphen in doc in title with parameter?


rugk:
Is it intended that you only use one hyphen (`–`) when a parameter is [mentioned in the title](https://borgbackup.readthedocs.io/en/latest/usage/notes.html#additional-notes) of the doc? E.g.
> ## –umask
> If you use `--umask`, […]

ThomasWaldmann:
Might be a feature gone wrong. Would be easier to understand if the two dashes are visible here.

takluyver:
I have had a go at fixing this in #2818

PR #2818, opened 2017-07-15
Use code markup for command line option headings

takluyver:
Closes gh-2815

The diff here looks big, but it's mostly just indenting the text that was already there.

I've also fixed a Sphinx cross link in the page.

Before:
screenshot from 2017-07-15 11-36-42

After:
screenshot from 2017-07-15 11-36-48

### DIFF
--- a/docs/usage/notes.rst
+++ b/docs/usage/notes.rst
@@ -12,7 +12,7 @@ The chunker params influence how input files are cut into pieces (chunks)
 which are then considered for deduplication. They also have a big impact on
 resource usage (RAM and disk space) as the amount of resources needed is
 (also) determined by the total amount of chunks in the repository (see
-`Indexes / Caches memory usage` for details).
+:ref:`cache-memory-usage` for details).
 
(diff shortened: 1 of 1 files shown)


rugk:
That's more a workaround than a fix. (just moving the `--params` away from headlines…)

takluyver:
True, but it's not a terrible workaround. It means Sphinx knows about them as command line options:

screenshot from 2017-07-15 11-50-46

(Not sure why they show up twice)

I tried code markup in headings - that works if you prefer, but it looks weird to me, because the headings become much smaller than normal.

rugk:
Interesting. What is the search term you used to find them?

takluyver:
I put in `--chunker-params`

enkore:
IIRC inline markup is permitted in section titles, so writing

```
``--...``
---
```

might be an easier workaround.

takluyver:
OK, redone with code markup in the headings

ThomasWaldmann:
Thanks!

Issue #2816, opened 2017-07-14
Doc: Prefer $()


rugk:
In the shell script [examples](https://borgbackup.readthedocs.io/en/latest/quickstart.html#automating-backups) I often see:
```
`command`
```
being used for executing commands inline/in subshells. However, IMHO/as far as I heard, this is not really recommend and one should use
```
$( command )
```
for this as it is more readable, explicit and does not use strange characters.

PR #2820, opened 2017-07-15
Fix bashism in shell scripts

rugk:
This makes the scripts fully POSIX-compatible and fixes some other small errors.
Used shellcheck.

Also fixes https://github.com/borgbackup/borg/issues/2816

Some errors, which were in the script:
* https://github.com/koalaman/shellcheck/wiki/SC1090
* https://github.com/koalaman/shellcheck/wiki/SC2076
* https://github.com/koalaman/shellcheck/wiki/SC2064

The modified parts of the quickstart script have been tested and it has exactly the same behaviour, i.e. "info" starts with a newline character and ends with ~one~ two, so that the line is separated by an empty line above and below.
Also this line has an alternative way with the same behaviour, I included there too. Choose what you like best. :smiley:

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -91,7 +91,6 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
 
     # some helpers and error handling:
     info() { printf "\n%s %s\n\n" "$( date )" "$*" >&2; }
-    # OR: info() { echo; echo "$( date ) $*" >&2; echo; }
     trap 'echo $( date ) Backup interrupted >&2; exit 2' INT TERM
 
     info "Starting backup"
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -90,8 +90,9 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     export BORG_PASSCOMMAND='pass show backup'
 
     # some helpers and error handling:
-    function info  () { echo -e "\n"`date` $@"\n" >&2; }
-    trap "echo `date` Backup interrupted >&2; exit 2" SIGINT SIGTERM
+    info() { printf "\n%s %s\n\n" "$( date )" "$*" >&2; }
+    # OR: info() { echo; echo "$( date ) $*" >&2; echo; }
+    trap 'echo $( date ) Backup interrupted >&2; exit 2' INT TERM
 
     info "Starting backup"
 
@@ -135,7 +136,8 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
 
     prune_exit=$?
 
-    global_exit=$(( ${backup_exit} >  ${prune_exit} ? ${backup_exit} : ${prune_exit} ))
+    # use highest exit code as global exit code
+    global_exit=$(( backup_exit > prune_exit ? backup_exit : prune_exit ))
 
     if [ ${global_exit} -eq 1 ];
     then
--- a/.travis/upload_coverage.sh
+++ b/.travis/upload_coverage.sh
@@ -4,7 +4,7 @@ set -e
 set -x
 
 NO_COVERAGE_TOXENVS=(pep8)
-if ! [[ "${NO_COVERAGE_TOXENVS[*]}" =~ "${TOXENV}" ]]; then
+if ! [[ "${NO_COVERAGE_TOXENVS[*]}" =~ ${TOXENV} ]]; then
     source ~/.venv/bin/activate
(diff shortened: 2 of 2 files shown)


docs/quickstart.rst
```diff
@@ -90,8 +90,9 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     export BORG_PASSCOMMAND='pass show backup'
 
     # some helpers and error handling:
-    function info  () { echo -e "\n"`date` $@"\n" >&2; }
-    trap "echo `date` Backup interrupted >&2; exit 2" SIGINT SIGTERM
+    info() { printf "\n%s %s\n\n" "$( date )" "$*" >&2; }
```
ThomasWaldmann: so "$*" is same as "$@"?
"function" is not needed?

docs/quickstart.rst
```diff
@@ -90,8 +90,9 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
...
+    # OR: info() { echo; echo "$( date ) $*" >&2; echo; }
+    trap 'echo $( date ) Backup interrupted >&2; exit 2' INT TERM
```
ThomasWaldmann: short form of signals is more compatible / not causing issues somewhere?

docs/quickstart.rst
```diff
@@ -90,8 +90,9 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     export BORG_PASSCOMMAND='pass show backup'
 
     # some helpers and error handling:
-    function info  () { echo -e "\n"`date` $@"\n" >&2; }
-    trap "echo `date` Backup interrupted >&2; exit 2" SIGINT SIGTERM
+    info() { printf "\n%s %s\n\n" "$( date )" "$*" >&2; }
+    # OR: info() { echo; echo "$( date ) $*" >&2; echo; }
```
ThomasWaldmann: I'm happy with anything that works reliably anywhere.
So, just keep whatever you like better considering that and remove the other version.

docs/quickstart.rst
rugk: Not directly. "$@" returns an array, which is troublesome here when multiple parameters are passed as the second (third, …) parameter would be on new lines. "$*" is a string only.

docs/quickstart.rst
rugk: It works both and both POSIX-compatible. What do you consider to look "better"/"nicer"?

docs/quickstart.rst
rugk: Shellcheck said that in POSIX signals are not prefixed with "SIG" and thus I removed it. The [ubuntu man page](http://manpages.ubuntu.com/manpages/trusty/man1/trap.1posix.html) is a bit confusing to me on this topic:
> Implementations may also accept the names with the SIG prefix; no known historical shell does so. 

Don't know what a historical shell is for them. For full compatibility you can only still use the signal numbers directly. (when signint and sigterm are the same on all systems…)

docs/quickstart.rst
ThomasWaldmann: oh, ok.

I am not much into shell scripting. If it is more than some trivial lines, I usually tend to rather use Python.

docs/quickstart.rst
ThomasWaldmann: ok, let's take printf.

docs/quickstart.rst
ThomasWaldmann: ok. historical likely means stuff like the bourne shell (sh, not bash), maybe also some other older shells.

rugk: So?

ThomasWaldmann: well, if it is posix, it should be fine as you have it now.

rugk: Great, then we have everything fixed. You can approve this review. :wink: 
And you can, BTW, also close this per-line discussions AFAIK.

ThomasWaldmann: github usually automatically folds review comments when they are outdated.

milkey-mouse:
Remember to close #2816 as well (or edit your commit message when you collapse the changesets so it auto-closes).

enkore:
> Also fixes #2816

milkey-mouse:
Does that also work in the body of a PR? The more you know.

rugk:
> Does that also work in the body of a PR?

AFAIK it actually only works in the body of the PR. In the title it may be ignored.

PR #2828, opened 2017-07-16
Reference list of status chars in borg recreate --filter description

### LINKED ISSUE #2730: Document valid status chars for --filter
The supported status chars should be listed with the help for `borg create --filter` (`--filter STATUSCHARS`).
This could just list the available choices (and validate them), and then link to some more specifc documentation/list, where each of them is explained / listed with a short desc, similar to `borg help pattern`?!

milkey-mouse:
Fixes #2730

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3469,8 +3469,8 @@ class Archiver:
         subparser.set_defaults(func=self.do_recreate)
         subparser.add_argument('--list', dest='output_list', action='store_true',
                                help='output verbose list of items (files, dirs, ...)')
-        subparser.add_argument('--filter', dest='output_filter', metavar='STATUSCHARS',
-                               help='only display items with the given status characters')
+        subparser.add_argument('--filter', metavar='STATUSCHARS', dest='output_filter',
+                               help='only display items with the given status characters (listed in borg create --help)')
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change anything')
         subparser.add_argument('-s', '--stats', dest='stats', action='store_true',


milkey-mouse:
On second thought I'm not sure if it's better to make the list of status chars the metavar like I did or just put them at the end of the helptext...

src/borg/archiver.py
```diff
@@ -2857,7 +2857,7 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                                help='print statistics for the created archive')
         subparser.add_argument('--list', dest='output_list', action='store_true',
                                help='output verbose list of items (files, dirs, ...)')
-        subparser.add_argument('--filter', metavar='STATUSCHARS', dest='output_filter',
+        subparser.add_argument('--filter', metavar='AMUEdbchsfi-x', dest='output_filter',
```
ThomasWaldmann: hmm, is this really an improvement or just another place that would need to get updated and could easily be forgotten?

the stuff is already described in 2819+.

src/borg/archiver.py
```diff
@@ -3469,8 +3469,8 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         subparser.set_defaults(func=self.do_recreate)
         subparser.add_argument('--list', dest='output_list', action='store_true',
                                help='output verbose list of items (files, dirs, ...)')
-        subparser.add_argument('--filter', dest='output_filter', metavar='STATUSCHARS',
-                               help='only display items with the given status characters')
+        subparser.add_argument('--filter', metavar='AMUEdbchsfi-x', dest='output_filter',
```

src/borg/archiver.py
milkey-mouse: that was my concern, should I just list them at the end of the help text?

milkey-mouse: Or should I consolidate both texts, like with a dict with each status char & its meaning that could be pretty printed in the help text or displayed in short form with `"".join(statuschars.keys())`?

enkore: I think just saying in the --filter help that this is described below is sufficient.

src/borg/archiver.py
```diff
@@ -2858,7 +2858,7 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         subparser.add_argument('--list', dest='output_list', action='store_true',
                                help='output verbose list of items (files, dirs, ...)')
         subparser.add_argument('--filter', metavar='STATUSCHARS', dest='output_filter',
-                               help='only display items with the given status characters (see description)')
+                               help='only display items with the given status characters (listed in borg create --help)')
```
ThomasWaldmann: as this is the borg create --help text, just keep the "see description" here?

ThomasWaldmann: also: collapse the changesets into 1.

ThomasWaldmann:
thanks!

PR #2832, opened 2017-07-16
replace modulo with if to check for wraparound in hashmap

rciorba:
Integer division is slow, and this improves the speed of all operations on the hashmap.

Benchmarks:
https://cdn.rawgit.com/rciorba/e3eded290e91a361f74eefdbc84416bb/raw/9e5d61e03c5842712d629e3e5567da3a512f1d79/results.html

Benchmarked this patch on the [rciorba/master-bench](https://github.com/borgbackup/borg/compare/master...rciorba:master-bench), where I applied the robin-hood benchmarks on master, if anyone cares to check the benchmark code

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -146,7 +146,10 @@ hashindex_lookup(HashIndex *index, const void *key, int *start_idx)
             }
             return idx;
         }
-        idx = (idx + 1) % index->num_buckets;
+        idx++;
+        if (idx >= index->num_buckets) {
+            idx -= index->num_buckets;
+        }
         if(idx == start) {
             break;
         }
@@ -547,7 +550,10 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
         }
         idx = start_idx;
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
-            idx = (idx + 1) % index->num_buckets;
+            idx++;
+            if (idx >= index->num_buckets){
+                idx -= index->num_buckets;
+            }
         }
         if(BUCKET_IS_EMPTY(index, idx)){
             index->num_empty--;


src/borg/_hashindex.c
```diff
@@ -146,7 +146,10 @@ hashindex_lookup(HashIndex *index, const void *key, int *start_idx)
             }
             return idx;
         }
-        idx = (idx + 1) % index->num_buckets;
+        idx ++;
```
enkore: nitpick: `idx++` (no space around unary operators)

ThomasWaldmann:
Hmm, actually the improvements on the left side might be similarly big, just less good visible due to scale.

@rciorba when you look at raw measurement values, how much is the improvement at 50% fill ratio?

rciorba:
@ThomasWaldmann the raw numbers are visible if you hover over the columns in the SVG

src/borg/_hashindex.c
rciorba: Fixed.

ThomasWaldmann:
So at 50% fill ratio it is 3..4% for most benchmarks, setitem even being 7% faster.

Big win for such a small change.

src/borg/_hashindex.c
```diff
@@ -146,7 +146,10 @@ hashindex_lookup(HashIndex *index, const void *key, int *start_idx)
             }
             return idx;
         }
-        idx = (idx + 1) % index->num_buckets;
+        idx++;
+        if (idx >= index->num_buckets) {
+            idx = 0;
```
ThomasWaldmann: hmm, this is a bit logically strange, without causing an actual malfunction.

i think it either should be `idx == index->num_buckets`, because only in that case `idx = 0` as action is correct AND we can be sure to hit that index as we are always incrementing by 1.

OR

keep `idx >= index->num_buckets`, but then it needs to be `idx -= index->num_buckets`.

src/borg/_hashindex.c
```diff
@@ -547,7 +550,10 @@ hashindex_set(HashIndex *index, const void *key, const void *value)
         }
         idx = start_idx;
         while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
-            idx = (idx + 1) % index->num_buckets;
+            idx++;
+            if (idx >= index->num_buckets){
+                idx = idx - index->num_buckets;
```
ThomasWaldmann: interesting, here it is correct already. except the `-=` syntactical abbreviation...

ThomasWaldmann:
adding backport-1.0 label.

little code change, easy to review, relatively big effect.

rciorba:
@ThomasWaldmann changes done. I notice the python 3.5.2 tests failed, are they flaky somehow? It would be strange if I had actually broken them but not 3.4 or 3.6, and the error seems unrelated.

enkore:
All tests are flaky on Travis due to fakeroot bugs.

ThomasWaldmann:
@rciorba thanks! yes, if you see "is a directory" error, it is likely due to a bug in fakeroot. and it is not just on travis, i also see it a lot on centos in vagrant machines.

ThomasWaldmann:
thanks!

PR #2836, opened 2017-07-16
Cleanup installation

enkore:


docs/usage_general.rst.inc
```diff
@@ -236,6 +236,15 @@ If you are suspicious whether your Borg repository is still consistent
 and readable after one of the failures mentioned above occurred, run
 ``borg check --verify-data`` to make sure it is consistent.
 
+.. rubric:: Requirements for Borg repository file systems
+
+- Long file names
+- At least three directory levels with short names
+- File sizes up to a few hundred MB
```
ThomasWaldmann: repo index could be much larger.

ThomasWaldmann: maybe we should specifically refer to "segment files" for this an the next line.

enkore: No, internal specificities shouldn't go in here. Rather, "- Typically, file sizes up to a few hundred MBs. Large repositories may require large files (>2 GB)." (i.e. not FAT32 for a big repo)

docs/usage_general.rst.inc
```diff
@@ -236,6 +236,16 @@ If you are suspicious whether your Borg repository is still consistent
 and readable after one of the failures mentioned above occurred, run
 ``borg check --verify-data`` to make sure it is consistent.
 
+.. rubric:: Requirements for Borg repository file systems
+
+- Long file names
+- At least three directory levels with short names
+- Typically, file sizes up to a few hundred MB.
+  Large repositories may require large files (>2 GB).
+- Up to 10000 files per directory
```
ThomasWaldmann: hmm, the new default in master is 1000 (was: 10000 in 1.0-maint). so, maybe mention these defaults/versions?

if we just say 10000 here, it might trigger work for some people who want to use some of that limited-to-3000 cloud stores until they find out that the current default is already below that.

ThomasWaldmann: fix that one sentence so we have it off the table and in rc1?

PR #2837, opened 2017-07-16
Detail how to use macOS/GNOME/KDE keyrings for repo passwords (fixes #392)

### LINKED ISSUE #392: support keyring for passphrase storage

milkey-mouse:
I haven't tested the macOS instructions yet, they were made with a careful reading of the security man pages. I also haven't made the equivalent tutorial for the KDE keyring equivalent (KWallet) yet.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -274,14 +274,15 @@ Security
 How can I specify the encryption passphrase programmatically?
 -------------------------------------------------------------
 
-The encryption passphrase or a command to retrieve the passphrase can be
-specified programmatically using the `BORG_PASSPHRASE` or `BORG_PASSCOMMAND`
-environment variables. This is convenient when setting up automated encrypted
-backups. Another option is to use key file based encryption with a blank passphrase.
-See :ref:`encrypted_repos` for more details.
+There are several ways to specify a passphrase without human intervention:
 
-.. _password_env:
-.. note:: Be careful how you set the environment; using the ``env``
+Setting ``BORG_PASSPHRASE``
+  The passphrase can be specified using the ``BORG_PASSPHRASE`` enviroment variable.
+  This is often the simplest option, but can be insecure if the script that sets it
+  is world-readable.
+
+  .. _password_env:
+  .. note:: Be careful how you set the environment; using the ``env``
           command, a ``system()`` call or using inline shell scripts
           (e.g. ``BORG_PASSPHRASE=hunter12 borg ...``)
           might expose the credentials in the process list directly
@@ -290,6 +291,83 @@ See :ref:`encrypted_repos` for more details.
           the environment of a process is `accessible only to that
           user
           <https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009>`_.
+
+Using ``BORG_PASSCOMMAND`` with a properly permissioned file
+  Another option is to create a file with a password in it in your home
+  directory and use permissions to keep anyone else from reading it. For
+  example, first create a key::
+
+    head -c 1024 /dev/urandom | base64 > ~/.borg-passphrase
+    chmod 400 ~/.borg-passphrase
+
+  Then in an automated script one can put::
+
+    export BORG_PASSCOMMAND="cat ~/.borg-passphrase"
+
+  and Borg will automatically use that passphrase.
+
+Using keyfile-based encryption with a blank passphrase
+  It is possible to encrypt your repository in ``keyfile`` mode instead of the default
+  ``repokey`` mode and use a blank passphrase for the key file. See :ref:`encrypted_repos`
+  for more details.
+
+Using ``BORG_PASSCOMMAND`` with MacOS Keychain
+  MacOS has a native manager for secrets (such as passphrases) which is safer
+  than just using a file as it is encrypted at rest and unlocked manually
+  (fortunately, the login keyring automatically unlocks when you login). With
+  the built-in ``security`` command, you can access it from the command line,
+  making it useful for ``BORG_PASSCOMMAND``.
+
+  First generate a passphrase and use ``security`` to save it to your login
+  (default) keychain::
+
+    security add-generic-password -D secret -U -a $USER -s borg-passphrase -w $(head -c 1024 /dev/urandom | base64)
+
+  In your backup script retrieve it in the ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="security find-generic-password -a $USER -s borg-passphrase"
+
+Using ``BORG_PASSCOMMAND`` with GNOME Keyring
+  GNOME also has a keyring daemon that can be used to store a Borg passphrase.
+  First ensure ``libsecret-tools``, ``gnome-keyring`` and ``libpam-gnome-keyring``
+  are installed. If ``libpam-gnome-keyring`` wasn't already installed, ensure it
+  runs on login::
+
+    sudo sh -c "echo session optional pam_gnome_keyring.so auto_start >> /etc/pam.d/login"
+    sudo sh -c "echo password optional pam_gnome_keyring.so >> /etc/pam.d/passwd"
+    # you may need to relogin afterwards to activate the login keyring
+
+  Then add a secret to the login keyring::
+
+    head -c 1024 /dev/urandom | base64 | secret-tool store borg-repository repo-name --label="Borg Passphrase"
+
+  If a dialog box pops up prompting you to pick a password for a new keychain, use your
+  login password. If there is a checkbox for automatically unlocking on login, check it
+  to allow backups without any user intervention whatsoever.
+
+  Once the secret is saved, retrieve it in a backup script using ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="secret-tool lookup borg-repository repo-name"
+
+  .. note:: For this to automatically unlock the keychain it must be run
+    in the ``dbus`` session of an unlocked terminal; for example, running a backup
+    script as a ``cron`` job might not work unless you also ``export DISPLAY=:0``
+    so ``secret-tool`` can pick up your open session. `It gets even more complicated`__
+    when you are running the tool as a different user (e.g. running a backup as root
+    with the password stored in the user keyring).
+
+__ https://github.com/borgbackup/borg/pull/2837#discussion_r127641330
+
+Using ``BORG_PASSCOMMAND`` with KWallet
+  KDE also has a keychain feature in the form of KWallet. The command-line tool
+  ``kwalletcli`` can be used to store and retrieve secrets. Ensure ``kwalletcli``
+  is installed, generate a passphrase, and store it in your "wallet"::
+
+    head -c 1024 /dev/urandom | base64 | kwalletcli -Pe borg-passphrase -f Passwords
+
+  Once the secret is saved, retrieve it in a backup script using ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="kwalletcli -e borg-passphrase -f Passwords"
 
 When backing up to remote encrypted repos, is encryption done locally?
 ----------------------------------------------------------------------
@@ -675,7 +753,7 @@ Here's a (incomplete) list of some major changes:
 * uses fadvise to not spoil / blow up the fs cache
 * better error messages / exception handling
 * better logging, screen output, progress indication
-* tested on misc. Linux systems, 32 and 64bit, FreeBSD, OpenBSD, NetBSD, Mac OS X
+* tested on misc. Linux systems, 32 and 64bit, FreeBSD, OpenBSD, NetBSD, MacOS
 
 Please read the :ref:`changelog` (or ``docs/changes.rst`` in the source distribution) for more
 information.


milkey-mouse:
If anyone reading this has a Mac and is willing to test the Mac instructions that would help (otherwise I'll have to go down to the Apple store and hope none of their Geniuses notice me 😉)

docs/faq.rst
```diff
@@ -291,6 +292,70 @@ See :ref:`encrypted_repos` for more details.
...
+  login password. If there is a checkbox for automatically unlocking on login, check it.
+
+  Once the secret is saved, retrieve it in a backup script using ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="secret-tool lookup borg-repository repo-name"
+
+  .. note:: For this to automatically unlock the keychain it must be run
+    in the ``dbus`` session of an unlocked terminal; for example, running a backup
+    script as a ``cron`` job might not work unless you also ``export DISPLAY=:0``
+    so ``secret-tool`` can pick up your open session.
```
enkore: Note: accessing the session bus _of a different user_, even as root, was made more complex some time ago.

```
$ cat /etc/dbus-1/session-local.conf 
<busconfig>
<policy context="default">
<allow user="root"/>  <!-- clearly, root needs explicit permission
                           to do anything nowadays -->
</policy>
</busconfig>
```

env vars: DISPLAY, XAUTHORITY=/home/.../.Xauthority, DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus

docs/faq.rst
```diff
@@ -291,6 +292,70 @@ See :ref:`encrypted_repos` for more details.
           user
           <https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009>`_.
 
+Using ``BORG_PASSCOMMAND`` with a properly permissioned file
+  Another option is to create a file with a password in it in your home
+  directory and use permissions to keep anyone else from reading it. For
+  example, first create a key::
+
+    dd if=/dev/urandom of=~/.borg-passphrase bs=1024 count=1
```
enkore: Perhaps use hex here as well, to avoid non-printables in the passphrase. Also, any NUL terminates the passphrase, so there is a relatively good chance that this passphrase is only a few characters long.

docs/faq.rst
```diff
@@ -291,6 +292,70 @@ See :ref:`encrypted_repos` for more details.
...
+    chmod 400 ~/.borg-passphrase
+
+  Then in an automated script one can put::
+
+    export BORG_PASSCOMMAND="cat ~/borg-passphrase"
```
enkore: ~/**.**b

docs/faq.rst
milkey-mouse: The note that's there already was copied pretty much verbatim from discussion in #392. Should I add another note about this/replace the existing one?

Why would you need to access the session bus of a different user for a backup script anyway? If you had `borg create` in root's crontab?

docs/faq.rst
milkey-mouse: Are all those `|project_name|`s mostly remnants from the attic fork? I was just using what I saw in the rest of the docs.

docs/faq.rst
enkore: Between a rock and a hard place: Document largely undocumented things about a platform which changes on a whim? Or don't document it, but someone(?) who attempts it will find it difficult to do so?

> If you had borg create in root's crontab?

Yes; it can also be used to send notifications to the desktop [of a specific user, Linux is not a multi-user OS].

docs/faq.rst
enkore: I think it was once part of the sphinx quickstart.

lbeltrame:
Note for docs: KWallet users can (assuming it's the KF5 based KWallet) use `kwallet-query --read-password <entry name> <wallet name>`. Unfortunately at the moment it outputs JSON, so it needs additional processing (via stdout) to extract the password.

milkey-mouse:
`kwallet-query` ended up not working on my Kubuntu test VM(?) but `kwalletcli` works perfectly and should also have backwards compatibility with KDE 4 and earlier, so I ended up just using that.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+  (default) keychain::
+
+    security add-generic-password -D secret -U -a $USER -s borg-passphrase -w $(head -c 30 /dev/urandom | xxd -p)
+
+  In your backup script retrieve it in the ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="security find-generic-password -a $USER -s borg-passphrase"
+
+Using ``BORG_PASSCOMMAND`` with GNOME Keyring (Ubuntu, Debian, Fedora)
+  GNOME has a keyring daemon similar to that of MacOS. However, it requires a bit more setup
```
rugk: Sounds biased… :wink: 

I.e. do not compare the keyrings and do not evaluate them.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+  First generate a passphrase and use ``security`` to save it to your login
+  (default) keychain::
+
+    security add-generic-password -D secret -U -a $USER -s borg-passphrase -w $(head -c 30 /dev/urandom | xxd -p)
+
+  In your backup script retrieve it in the ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="security find-generic-password -a $USER -s borg-passphrase"
+
+Using ``BORG_PASSCOMMAND`` with GNOME Keyring (Ubuntu, Debian, Fedora)
```
rugk: The desktop environment does not depend on the distro used, so rather leave the "()" list out. Add "(Linux)" if you want as AFAIK GNOME only runs on Linux.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+  GNOME has a keyring daemon similar to that of MacOS. However, it requires a bit more setup
+  to get the login keyring to work. First ensure ``libpam-gnome-keyring`` is installed and
+  running on login (skip this step on Fedora, as it is installed by default)::
```
rugk: Mhh, I doubt Fedora is the only distro where this is the case. Maybe delete that sentence or make it more general ("In some ditros this is done by default.")

Or just change the wording in general: "If XY is not installed, then you have to install it first."

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+    sudo apt install gnome-keyring libpam-gnome-keyring
+    # the following commands may also be necessary if the previous command installed new packages
+    sudo sh -c "echo session optional pam_gnome_keyring.so auto_start >> /etc/pam.d/login"
+    sudo sh -c "echo password optional pam_gnome_keyring.so >> /etc/pam.d/passwd"
+    # you may need to relogin afterwards to activate the login keyring
+
+  Then add a secret to the login keyring::
+
+    sudo apt install libsecret-tools  # Ubuntu, Debian
+    sudo yum install libsecret-tools  # Fedora, RHEL
```
rugk: New RHEL/Fedora-based distros should use "dnf". Please change it, the syntax stays the same.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
           user
           <https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009>`_.
 
+Using ``BORG_PASSCOMMAND`` with a properly permissioned file
+  Another option is to create a file with a password in it in your home
+  directory and use permissions to keep anyone else from reading it. For
+  example, first create a key::
+
+    head -c 1024 /dev/urandom | xxd -p > ~/.borg-passphrase
```
rugk: Do you rally want to use binary stuff as your passphrase? IMHO the output should be filtered, so only ASCII (or similar) characters are used. At some day you may want to enter the passphrase manually.

If xxd is the thing doing this, then let's just say me that it is not installed by default for me. (Fedora 26 here) Better use a compatible command.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+    # the following commands may also be necessary if the previous command installed new packages
+    sudo sh -c "echo session optional pam_gnome_keyring.so auto_start >> /etc/pam.d/login"
+    sudo sh -c "echo password optional pam_gnome_keyring.so >> /etc/pam.d/passwd"
+    # you may need to relogin afterwards to activate the login keyring
+
+  Then add a secret to the login keyring::
+
+    sudo apt install libsecret-tools  # Ubuntu, Debian
+    sudo yum install libsecret-tools  # Fedora, RHEL
+    head -c 30 /dev/urandom | xxd -p | secret-tool store borg-repository repo-name --label="Borg Passphrase"
```
rugk: So why do you only use 30 bytes here in contras to 1024 above?

Also same "xxd" issue as above.

rugk: You may also mention that of course you can also do it with graphical tools (Seahorse in GNOME).

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+
+  If a dialog box pops up prompting you to pick a password for a new keychain, use your
+  login password. If there is a checkbox for automatically unlocking on login, check it.
```
rugk: Maybe: "check it, if you ant that to be the case." OR "check it, if you want automatic updates in the background without interrupting you." or something similar.

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+    in the ``dbus`` session of an unlocked terminal; for example, running a backup
+    script as a ``cron`` job might not work unless you also ``export DISPLAY=:0``
+    so ``secret-tool`` can pick up your open session. `It gets even more complicated`__
+    when you are running the tool as a different user (e.g. running a backup as root
+    with the password stored in the user keyring).
+
+__ https://github.com/borgbackup/borg/pull/2837#discussion_r127641330
+
+Using ``BORG_PASSCOMMAND`` with KWallet (Kubuntu, other KDE systems)
+  KDE also has a similar keychain feature in a program called KWallet. The command-line
```
rugk: Again this comparison… Just don't talk of "similar" here. (I could probably accept "also".) That does not matter for the user anyway – they do not want to know how/whether these things are similar, :smiley:

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+  tool ``kwalletcli`` can be used to store and retrieve secrets. First generate a
+  passphrase and store it in your "wallet"::
+
+    sudo apt install kwalletcli  # might not be installed by default
+    head -c 30 /dev/urandom | xxd -p | kwalletcli -Pe borg-passphrase -f Passwords
```
rugk: xxd…

docs/faq.rst
```diff
@@ -291,6 +292,86 @@ See :ref:`encrypted_repos` for more details.
...
+
+  Once the secret is saved, retrieve it in a backup script using ``BORG_PASSCOMMAND``::
+
+    export BORG_PASSCOMMAND="kwalletcli -e borg-passphrase -f Passwords"
+
+Using keyfile-based encryption with a blank passphrase
```
rugk: I'd say this suggestion should be mentioned before all the keychain stuff as it is simple, general and something a user would appreciate/should think about before potentially getting into more difficult stuff.

docs/faq.rst
enkore: Consider removing installation instructions for other software entirely - it's always going to change slightly and there will always be distros left out. Just saying what is needed should be sufficient to install it.

docs/faq.rst
milkey-mouse: `xxd` should be converting it to hex here. I thought (mistakenly) that it was in coreutils and I could assume people would have it installed.

As for why I was using 30 chars, it was the boundary for a line break in `xxd`, and I was averse to having newlines in the password. Of course, I could have fixed it with a simple `tr`, but that would have complicated the instructions. In hindsight, I don't see why I was so dead-set on not having newlines in the password anyway...

I'll change that to use `base64`, which is *actually* in GNU coreutils (and seems to exist on Mac, too).

docs/faq.rst
milkey-mouse: Same reason for the 30 bytes thing as above, I'll probably change these to use ~1024 bytes.

milkey-mouse: It's harder to write tutorials for how to do it in a GUI IMO. Also if they copy-paste these instructions they can't accidentally use the wrong folder or anything (it will put them in the defaults for `secret-tool`).

docs/faq.rst
milkey-mouse: I had the file-based option at the very beginning, which IMO is less hacky than the keyfile-based encryption, but on second thought it would make sense to have the keyfile option before the keychain stuff.

docs/faq.rst
milkey-mouse: I put that there in the first place because it was kind of a PITA to get it working/get working instructions on Debian and Ubuntu (which don't have it installed by default), so I wanted to save the users some time. If I'm dropping all mentions of which distros have what DEs and what package managers they use I might as well drop this too, though; the remedy if it's not installed is only two commands.

milkey-mouse:
collapsed changesets, ready to merge

ThomasWaldmann:
@milkey-mouse it is good to collapse some changesets at the end (after all review is finished and at least one core dev approved), but if you do it in the middle of review, you require reviewers to re-read all the stuff just to find out whether the feedback got actually into the changeset. It easier to review incremental diffs, fixup changesets.

docs/faq.rst
rugk: Yeah, but you could just mention that it is _alos_ possible with the GUI, so users not liking the terminal don't get angry. :wink: 

Seriously, they should just know that it is possible. Mention the application name(s) as I made in my comment, e.g., if you want to give a hint how to do it. That's all.

enkore:
Thanks!

docs/faq.rst
jkirk: Not sure if this is the best place to comment, but I hope my this helps someone who stumbles of this and point me to the right place where it could be added.

This is how I use GNOME Keyring with borg in a user cron job (as I need a `unlocked  terminal` (which I read as unlocked GNOME session) anyway:
```
export BORG_PASSCOMMAND='sudo -u $USERNAME DISPLAY=:0 DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus secret-tool lookup borg-repository repo-name'
sudo -E borg create /path/to/repo:mybackup-{now} /path/to/backup 
````

The secret key was added as described as user `$USERNAME`:
```
secret-tool store borg-repository repo-name --label="Borg Passphrase"
```

PR #2838, opened 2017-07-17
Use archive creation time as mtime for FUSE mount (fixes #2834)

### LINKED ISSUE #2834: Modification time in fuse mount should show creation time of archive
STR: (tested with 1.1.0b6)
``` 
# create some backups
borg mount :: /tmp/mount
ls -la /tmp/mount
total 0
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 .
drwxrwxrwt. 27 root root 660 Jul 16 18:58 ..
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 backup-block-device
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 backup1
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 backup2
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 backup3
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 rugk-2017-07-16T18:51:34
drwxr-xr-x.  1 rugk rugk   0 Jul 16 18:58 rugk-2017-07-16T18:52:19
```

They are all set to the current time. It would be better if the directories show the creation time of the archives as their "modification time".

milkey-mouse:
This is already stored in the `ArchiveInfo` from the manifest so it can still lazy-load the archive itself.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -242,16 +242,15 @@ class FuseOperations(llfuse.Operations):
             self.process_archive(self.args.location.archive)
         else:
             self.versions_index = FuseVersionsIndex()
-            archive_names = (x.name for x in self.manifest.archives.list_considering(self.args))
-            for archive_name in archive_names:
+            for archive in self.manifest.archives.list_considering(self.args):
                 if self.versions:
                     # process archives immediately
-                    self.process_archive(archive_name)
+                    self.process_archive(archive.name)
                 else:
                     # lazily load archives, create archive placeholder inode
-                    archive_inode = self._create_dir(parent=1)
-                    self.contents[1][os.fsencode(archive_name)] = archive_inode
-                    self.pending_archives[archive_inode] = archive_name
+                    archive_inode = self._create_dir(parent=1, mtime=int(archive.ts.timestamp() * 1e9))
+                    self.contents[1][os.fsencode(archive.name)] = archive_inode
+                    self.pending_archives[archive_inode] = archive.name
 
     def sig_info_handler(self, sig_no, stack):
         logger.debug('fuse: %d synth inodes, %d edges (%s)',
@@ -303,11 +302,15 @@ class FuseOperations(llfuse.Operations):
         finally:
             llfuse.close(umount)
 
-    def _create_dir(self, parent):
+    def _create_dir(self, parent, mtime=None):
         """Create directory
         """
         ino = self.allocate_inode()
-        self.items[ino] = self.default_dir
+        if mtime is not None:
+            self.items[ino] = Item(**self.default_dir.as_dict())
+            self.items[ino].mtime = mtime
+        else:
+            self.items[ino] = self.default_dir
         self.parent[ino] = parent
         return ino
 
@@ -375,7 +378,7 @@ class FuseOperations(llfuse.Operations):
                 self.file_versions[path] = version
 
         path = item.path
-        del item.path  # safe some space
+        del item.path  # save some space
         if 'source' in item and hardlinkable(item.mode):
             # a hardlink, no contents, <source> is the hardlink master
             source = os.fsencode(item.source)


src/borg/fuse.py
```diff
@@ -242,16 +243,20 @@ def _create_filesystem(self):
...
                     # process archives immediately
-                    self.process_archive(archive_name)
+                    self.process_archive(archive.name)
                 else:
                     # lazily load archives, create archive placeholder inode
-                    archive_inode = self._create_dir(parent=1)
-                    self.contents[1][os.fsencode(archive_name)] = archive_inode
-                    self.pending_archives[archive_inode] = archive_name
+                    archive_inode = self.allocate_inode()
+                    self.items[archive_inode] = deepcopy(self.default_dir)
```
enkore: `Item(**self.default_dir.as_dict())` is perhaps less magic

src/borg/fuse.py
```diff
@@ -242,16 +243,20 @@ def _create_filesystem(self):
...
+                    created_ts = int(archive.ts.timestamp() * 1e9)
+                    self.items[archive_inode].mtime = created_ts
```
enkore: It is a bit more complicated, search this file for `have_fuse_xtime_ns` too see how integer-nanoseconds are exposed to FUSE.

milkey-mouse: Does that need to be accounted for here? The `have_fuse_xtime_ns` bit seems to always use `item.mtime` in microseconds.

src/borg/fuse.py
```diff
@@ -242,16 +242,18 @@ def _create_filesystem(self):
...
+                    self.process_archive(archive.name)
                 else:
                     # lazily load archives, create archive placeholder inode
-                    archive_inode = self._create_dir(parent=1)
-                    self.contents[1][os.fsencode(archive_name)] = archive_inode
-                    self.pending_archives[archive_inode] = archive_name
+                    archive_inode = self.allocate_inode()
+                    self.items[archive_inode] = Item(**self.default_dir.as_dict())
+                    self.items[archive_inode].mtime = int(archive.ts.timestamp() * 1e9)
+                    self.parent[archive_inode] = 1
```
ThomasWaldmann: hmm, maybe add a mtime=None param to _create_dir and move the code there?

src/borg/fuse.py
```diff
@@ -242,16 +242,15 @@ def _create_filesystem(self):
...
                 if self.versions:
                     # process archives immediately
-                    self.process_archive(archive_name)
+                    self.process_archive(archive.name)
                 else:
                     # lazily load archives, create archive placeholder inode
-                    archive_inode = self._create_dir(parent=1)
-                    self.contents[1][os.fsencode(archive_name)] = archive_inode
-                    self.pending_archives[archive_inode] = archive_name
+                    archive_inode = self._create_dir(parent=1, mtime=int(archive.ts.timestamp() * 1e9))
```
ThomasWaldmann: move the computation into _create_dir, give the float timestamp as param (or rename mtime param to mtime_ns and give the ns integer value).

there, you need to consider the value of have_fuse_xtime_ns to compute the correct value and choose the correct attribute name to assign that value. see the other place where have_fuse_xtime_ns is used.

milkey-mouse: I see the `have_fuse_xtime_ns` code, but doesn't that run for the `EntryAttributes` object and convert it there? Why do I need to convert it if it's not necessary for [`self.default_dir`](https://github.com/borgbackup/borg/blob/master/src/borg/fuse.py#L232) here?

ThomasWaldmann: I have a look right now...

ThomasWaldmann: You're right. we are constructing an Item here, which has mtime in nanoseconds always.

Only the entry constructed by FUSE's getattr needs to consider the have_fuse_xtime_ns flag.

OK, so we keep it as is.

src/borg/fuse.py
```diff
@@ -9,6 +9,7 @@
 from collections import defaultdict
 from signal import SIGINT
 from distutils.version import LooseVersion
+from copy import deepcopy
```
ThomasWaldmann: not used any more?

milkey-mouse:
sure

milkey-mouse:
oops...

PR #2839, opened 2017-07-17
Enable pylint check E0211 (method has no argument) in .coafile

Alexander-N:
Adapt Attic013Item in test_attic013_acl_bug so that as_dict takes `self` as
first argument. This was the only case where E0211 was violated.


ThomasWaldmann:
thanks!

Issue #2475, opened 2017-05-02
Add diff functionality to mount


imsodin:
This is a proposal for a new function, that does the following:  
Add an option to the mount command to mount only files/directories that changed in between two archives.  
It would be even more awesome, if files could be further filtered by the categories given by the diff command (i.e. added, removed, ...). For the filter command itself this could also be viable, but less so, as it can already be done by grepping its output.

Why do I want this: I use the following structure with rsync based backups to quickly find/retrieve changes on different platforms (i.e. nfs/smb mounted):  

<pre>
archive/
  full/
  incr/ (only changes to previous backup)
  del/ (only files deleted since previous backup)
    dates/
       backup/ (actual original directory structure)
</pre>

I am aware of the new merged mount option in 1.1.0, which will allow to quickly visualize changes, but not really restore - so these additions are still worthwhile. Actually they would play together perfectly.

Thanks once more for all your work - borg is awesome!

imsodin:
I would like to implement this feature. I already spent some time reading the relevant code and have a general idea of how to approach it. Before going ahead and spending time on implementation, I want to get a "sounds reasonable" from someone experienced with the borg code.

The general idea:
Introduce a "--diff ARCHIVE2" option to the mount command. Only mount items that differ compared to ARCHIVE2.
I would also like to introduce an additional option to both diff and mount that allows to specify which "types of differences" should be considered. That might be trough additional options (e.g. --diff-del, diff-mod, diff-meta, ...) or one option with a string to be parsed (e.g. --diff-types=del,mod,meta). This would probably be a followup PR though, but it could influence some design choices.

I want to refactor the code regarding the diff feature to separate comparing two archives from generating the pretty command line output. The first part will return some data structure, containing all changes (e.g. a dictionary with change types as keys and list of items as values).

In the fuse code I want to use the first part of the diff feature to filter out items that did not change compared to the other archive. I see two options to do this: Either add a condition in `process_archive` dedicated to this purpose or introduce a `filter` keyword argument in the same way as there is in archive.py. The first option is the least intrusive one, the second is more generic (i.e. could be used for other purposes as well) and consistent with already existing code.

ThomasWaldmann:
Sounds reasonable.

Maybe it should be `borg mount NEW mnt --diff OLD` so you can visualize added / changed files in NEW (by showing them below mnt).

[...]

PR #2842, opened 2017-07-18
Refactor diff functionality

imsodin:
This is the first step for #2475. It dos not change any functionality, it just moves logic around and provides a way to get changes between two archives in a "programmatically reusable" form.

Do you usually squash merge or do I need to rebase instead of merging master?

### DIFF
--- a/src/borg/item.pyx
+++ b/src/borg/item.pyx
@@ -365,9 +365,9 @@ class ManifestItem(PropDict):
         # indicated in compare_mode instead.
         if item1.get('deleted'):
             return 'added link'
-        elif item2.get('deleted'):
+        if item2.get('deleted'):
             return 'removed link'
-        elif 'source' in item1 and 'source' in item2 and item1.source != item2.source:
+        if 'source' in item1 and 'source' in item2 and item1.source != item2.source:
             return 'changed link'
 
 def compare_content(path, item1, item2):
@@ -390,7 +390,7 @@ def compare_content(path, item1, item2):
     def compare_directory(item1, item2):
         if item2.get('deleted') and not item1.get('deleted'):
             return 'removed directory'
-        elif item1.get('deleted') and not item2.get('deleted'):
+        if item1.get('deleted') and not item2.get('deleted'):
             return 'added directory'
 
     def compare_owner(item1, item2):
@@ -406,13 +406,11 @@ def compare_content(path, item1, item2):
     def contents_changed(item1, item2):
         if can_compare_chunk_ids:
             return item1.chunks != item2.chunks
-        else:
-            if sum_chunk_size(item1) != sum_chunk_size(item2):
-                return True
-            else:
-                chunk_ids1 = [c.id for c in item1.chunks]
-                chunk_ids2 = [c.id for c in item2.chunks]
-                return not fetch_and_compare_chunks(chunk_ids1, chunk_ids2, archive1, archive2)
+        if sum_chunk_size(item1) != sum_chunk_size(item2):
+            return True
+        chunk_ids1 = [c.id for c in item1.chunks]
+        chunk_ids2 = [c.id for c in item2.chunks]
+        return not fetch_and_compare_chunks(chunk_ids1, chunk_ids2, archive1, archive2)
 
     @staticmethod
     def compare_chunk_contents(chunks1, chunks2):
--- a/src/borg/item.pyx
+++ b/src/borg/item.pyx
@@ -5,6 +5,7 @@ from .constants import ITEM_KEYS
 from .helpers import safe_encode, safe_decode
 from .helpers import bigint_to_int, int_to_bigint
 from .helpers import StableDict
+from .helpers import format_file_size
 
 cdef extern from "_item.c":
     object _object_to_optr(object obj)
@@ -184,19 +185,22 @@ class Item(PropDict):
 
     part = PropDict._make_property('part', int)
 
-    def get_size(self, hardlink_masters=None, memorize=False, compressed=False, from_chunks=False):
+    def get_size(self, hardlink_masters=None, memorize=False, compressed=False, from_chunks=False, consider_ids=None):
         """
         Determine the (uncompressed or compressed) size of this item.
 
-        For hardlink slaves, the size is computed via the hardlink master's
-        chunk list, if available (otherwise size will be returned as 0).
-
-        If memorize is True, the computed size value will be stored into the item.
+        :param hardlink_masters: If given, the size of hardlink slaves is computed via the hardlink master's chunk list,
+        otherwise size will be returned as 0.
+        :param memorize: Whether the computed size value will be stored into the item.
+        :param compressed: Whether the compressed or uncompressed size will be returned.
+        :param from_chunks: If true, size is computed from chunks even if a precomputed value is available.
+        :param consider_ids: Returns the size of the given ids only.
         """
         attr = 'csize' if compressed else 'size'
         assert not (compressed and memorize), 'Item does not have a csize field.'
+        assert not (consider_ids is not None and memorize), "Can't store size when considering only certain ids"
         try:
-            if from_chunks:
+            if from_chunks or consider_ids is not None:
                 raise AttributeError
             size = getattr(self, attr)
         except AttributeError:
@@ -226,7 +230,10 @@ class Item(PropDict):
                         chunks, _ = hardlink_masters.get(master, (None, None))
                 if chunks is None:
                     return 0
-            size = sum(getattr(ChunkListEntry(*chunk), attr) for chunk in chunks)
+            if consider_ids is not None:
+                size = sum(getattr(ChunkListEntry(*chunk), attr) for chunk in chunks if chunk.id in consider_ids)
+            else:
+                size = sum(getattr(ChunkListEntry(*chunk), attr) for chunk in chunks)
             # if requested, memorize the precomputed (c)size for items that have an own chunks list:
             if memorize and having_chunks:
                 setattr(self, attr, size)
@@ -251,6 +258,21 @@ class Item(PropDict):
     def from_optr(self, optr):
         return _optr_to_object(optr)
 
+    @classmethod
+    def create_deleted(cls, path):
+        return cls(deleted=True, chunks=[], mode=0, path=path)
+
+    def is_link(self):
+        return self._is_type(stat.S_ISLNK)
+
+    def is_dir(self):
+        return self._is_type(stat.S_ISDIR)
+
+    def _is_type(self, typetest):
+        try:
+            return typetest(self.mode)
+        except AttributeError:
+            return False
 
 
 class EncryptedKey(PropDict):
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
at the end of review, I usually like to (partially) collapse the changesets (so trivial fixes are merged into the changeset they fix, but stuff that somehow implements a bigger separate idea is kept separate). if there is a merge conflict showing up, rebasing it after review makes sense.

imsodin:
> just as a general comment:
> moving stuff around and at the same time refactoring it makes review a pain, because one can not easily see the changes.

I create two preliminary commits, whose only purpose is to make the diffs more readable. The first only moves and indents code, the second does some style changes. Both are not functional, so they need to be squashed before committing.

ThomasWaldmann:
ok, thanks, I'll have another look later tonight.

imsodin:
I changed how changes are obtained from archives: Previously the changes were given as a dictionary, now as a generator. So this should now not use any more memory than it did previously.

imsodin:
Any feedback?

src/borg/item.pyx
```diff
@@ -184,19 +185,22 @@ class Item(PropDict):
...
+        :param compressed: Whether the compressed or uncompressed size will be returned.
+        :param from_chunks: If true, size is computed from chunks even if a precomputed value is available.
+        :param consider_ids: Returns the size of the given ids only.
         """
         attr = 'csize' if compressed else 'size'
         assert not (compressed and memorize), 'Item does not have a csize field.'
+        assert not (consider_ids and memorize), "Can't store size when considering only certain ids"
         try:
-            if from_chunks:
+            if from_chunks or consider_ids:
```
ThomasWaldmann: - None means: consider all ids
- `[]` means: consider no ids
- as both are falsy, you need the mentioned more precise check.

src/borg/item.pyx
```diff
@@ -251,6 +258,17 @@ class Item(PropDict):
...
+        return self._is_type('l')
+
+    def is_dir(self):
+        return self._is_type('d')
+
+    def _is_type(self, typestr):
+        try:
+            return stat.filemode(self.mode)[0] == typestr
+        except AttributeError:
+            return False
```
ThomasWaldmann: this is like the code previously did it, we can change this later to directly check self.mode, without the stat.filemode call.

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
+
+        if not (self.item1.get('deleted') or self.item2.get('deleted')):
+            changes.append(self._owner_string())
+            changes.append(self._mode_string())
+
+        return ' '.join((x for x in changes if x))
+
+    def _equal(self, chunk_iterator1, chunk_iterator2, can_compare_chunk_ids):
+        # if both are deleted, there is nothing at path regardless of what was deleted
+        if self.item1.get('deleted') and self.item2.get('deleted'):
```
ThomasWaldmann: hmm? how can that happen? if it would be deleted in both archives, how would we know of it?

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
+            return True
+
+        attr_list = ['deleted', 'mode', 'source']
+        if self._numeric_owner:
+            attr_list += ['uid', 'gid']
+        else:
+            attr_list += ['user', 'group']
```
ThomasWaldmann: attr_list += A if self._numeric_owner else B

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
-        chunk_ids1 = {c.id for c in item1.chunks}
-        chunk_ids2 = {c.id for c in item2.chunks}
+        return ''
+
+    def _content_string(self):
+        if self.item1.get('deleted'):
+            return ('added {:>13}'.format(format_file_size(self.item2.get_size())))
+        if self.item2.get('deleted'):
+            return ('removed {:>11}'.format(format_file_size(self.item1.get_size())))
+        # How is working with ids correct if we cannot compare ids (i.e. can_compare_chunk_ids == False)?
```
ThomasWaldmann: good catch. it is not correct.

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
 
-    def compare_directory(item1, item2):
-        if item2.get('deleted') and not item1.get('deleted'):
+    def _dir_string(self):
+        if self.item2.get('deleted') and not self.item1.get('deleted'):
             return 'removed directory'
-        if item1.get('deleted') and not item2.get('deleted'):
+        if self.item1.get('deleted') and not self.item2.get('deleted'):
             return 'added directory'
+        return ''
```
ThomasWaldmann: as you filter out all falsy values, the implicit `return None` would also work.

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
-            return '[{}:{} -> {}:{}]'.format(user1, group1, user2, group2)
+    def _owner_string(self):
+        if self._numeric_owner:
+             if self.item1.get('uid') != self.item2.get('uid') or self.item1.get('gid') != self.item2.get('gid'):
+                 return '[{}:{} -> {}:{}]'.format(self.item1.get('uid'), self.item1.get('gid'),
+                                                  self.item2.get('uid'), self.item2.get('gid'))
+        else:
+            if self.item1.get('user') != self.item2.get('user') or self.item1.get('group') != self.item2.get('group'):
+                 return '[{}:{} -> {}:{}]'.format(self.item1.get('user'), self.item1.get('group'),
+                                                  self.item2.get('user'), self.item2.get('group'))
```
ThomasWaldmann: guess that would read better like this:
```
u_attr, g_attr = ('uid', 'gid') if self._numeric_owner else ('user', 'group')
u1, g1 = self.item1.get(u_attr), self.item1.get(g_attr)
u2, g2 = self.item2.get(u_attr), self.item2.get(g_attr)
if (u1, g1) != (u2, g2):
    return '[{}:{} -> {}:{}]'.format(u1, g1, u2, g2)
```

src/borg/item.pyx
```diff
@@ -359,60 +377,127 @@ class ManifestItem(PropDict):
...
+    def _chunk_contents_equal(chunks1, chunks2):
+        """
+        Compare chunk contents and return True if they are identical.
+
+        The chunks must be given as chunk iterators (like returned by :meth:`.DownloadPipeline.fetch_many`).
+        """
+
         end = object()
         alen = ai = 0
         blen = bi = 0
```
ThomasWaldmann: btw, did you check how good the test coverage for all this stuff is?

src/borg/item.pyx
imsodin: Don't think into it, that was me taking the "save" route as I am not sure about the role of the "deleted" attribute. An item can either have this attribute or just be missing from one archive ("orphans"). The two theories I came up with both don't work out: "deleted" is used when a file was deleted in relation to the last archive - however AFAIK borg doesn't have a concept of "consecutive" archives, it "just" deduplicates. The other option is that this is written once all archives referencing this item are deleted/pruned - however in this case I would never have to care about 'deleted', as it can never come up. The documentation talks about this, but more about how it is handled in the database, not about what causes elements/objects to be deleted.

src/borg/item.pyx
```diff
@@ -358,3 +358,84 @@ class ManifestItem(PropDict):
...
+                if b is end:
+                    return not alen - ai and next(chunks1, end) is end
+                b = memoryview(b)
+                blen = len(b)
+                bi = 0
+            slicelen = min(alen - ai, blen - bi)
+            if a[ai:ai + slicelen] != b[bi:bi + slicelen]:
+                return False
+            ai += slicelen
+            bi += slicelen
```
imsodin: Forgot to remove it from archiver.py - thanks for catching this!

src/borg/item.pyx
imsodin: When comparing chunks, the length of a memoryview of a chunk is its size in bytes, right?
I could then fix this, but probably better in a separate PR, as it is unrelated.

src/borg/item.pyx
imsodin: Codecov doesn't seem to include item.pyx into its report?

src/borg/item.pyx
ThomasWaldmann: IIRC, "deleted" was only an attribute added by the diffing code, not something stored in the archive.

But I have to check again...

src/borg/item.pyx
ThomasWaldmann: if this was just taken from the previous code: yes, do it after the refactor. maybe file an issue so it is not forgotten.

src/borg/item.pyx
enkore: I believe so. This is also something that we'll want to get a better grip on for multithreading, because some extra attributes that shall not be stored need to be carried around.

src/borg/item.pyx
ThomasWaldmann: pyx is compiled to C which is compiled to binary, so i don't think coverage.py can work with that.

src/borg/item.pyx
```diff
@@ -198,9 +198,9 @@ class Item(PropDict):
         """
         attr = 'csize' if compressed else 'size'
         assert not (compressed and memorize), 'Item does not have a csize field.'
-        assert not (consider_ids and memorize), "Can't store size when considering only certain ids"
+        assert not (consider_ids is not None and memorize), "Can't store size when considering only certain ids"
         try:
-            if from_chunks or consider_ids:
+            if from_chunks or consider_ids is None:
```
ThomasWaldmann: is not None?

src/borg/item.pyx
enkore: It's of course possible to have code coverage for C, its just not configured in this repository.

Cython on the other hand is more difficult because you'd need something like sourcemaps and I never heard of any tooling like that for Cython.

src/borg/item.pyx
ThomasWaldmann: much better. maybe could be also done at some other places.

src/borg/item.pyx
ThomasWaldmann: I just checked it:
```
deleted_item = Item(deleted=True, chunks=[], mode=0)
```
This is artificially introduced to have a comparison counterpart for orphaned items in archive1 or archive2 (== items that miss a counterpart). Then the usual comparison code can work.

ThomasWaldmann: BTW, there might be a bug in the original code as the one deleted_item object is just patched with new pathes (but it is always the same one object). That of course does not work if references to it are stored for other purposes / for longer than it has the correct path.

imsodin: Thanks for looking into this. I think there was no problem in the original code as the constructed item is only used to generate the comparison string. ItemDiff does not care about the path, but I think it is still a problem in waiting to store references.
It seems pointless to add an attr to `Item` just for this single purpose. The artificially created deleted item doesn't carry any relevant information except that it was deleted. This could easily be done by passing `None` to the ItemDiff constructor to signify a deleted item. Good idea?

ThomasWaldmann: it can be solved in other ways, but i guess the point was to have an easy comparison without introducing a lot of special cases in the comparison function.

e.g. the deleted_item has chunks = [], so if it is compared to a normal item, all chunks show up as added/removed naturally. Also, size increase/decrease will also just work. Same thing can be done for all sorts of metadata, just put an empty/zero value in deleted_item.

src/borg/item.pyx
imsodin: It is also possible to have line coverage for Cython like this
http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html
https://medium.com/@dfdeshom/better-test-coverage-workflow-for-cython-modules-631615eb197a
However I don't get the hole fakeroot/tox/setup process of borg, so I couldn't figure out quickly how to integrate this. Would certainly be useful though.

src/borg/item.pyx
```diff
@@ -398,87 +402,90 @@ class ItemDiff:
...
-        if 'mode' in self.item1:     # mode of item1 and item2 is equal
-            filemode = stat.filemode(self.item1.mode)
-            if filemode[0] == 'l' and 'source' in self.item1 and 'source' in self.item2 and self.item1.source != self.item2.source:
-                return False
+        if 'mode' in self._item1:     # mode of item1 and item2 is equal
+            filemode = stat.filemode(self._item1.mode)
+            try:
+                if self._item1.is_link() and self._item1.source != self._item2.source:
+                    return False
+            except AttributeError:
```
ThomasWaldmann: guess you mean to catch AttributeErrors from `.source` here, but it would also catch them from `._itemN` or `.is_link` or from inside of the `.is_link` method.

maybe we can do it as in the original code. i think you micro-optimized here, but as is_link() is most often false, it does not make a big difference, I guess.

imsodin: Yes I do want to catch the errors from `.source`. I don't really understand the problem, as I think there is nothing to catch for `._itemN` (always set on initialization) and `.is_link()` (already excepts `AttributeError` in the function).

ThomasWaldmann: the point was just about a style issue with a too wide scope and rather general exception handler.
so in some cases it could catch not the exception you intended to catch.
you might be lucky here that it can never happen, but such code is problematic in general.

ThomasWaldmann:
can we in the end collapse the "review" changesets, so they don't appear in history?

imsodin:
@ThomasWaldmann Should I also collapse the first two non-functional commits? These don't build.

ThomasWaldmann:
It is easier for review to keep the first two separate.

But if it is too much hassle to apply the review-fixups to the respective places, collapsing them all is still an option.

src/borg/item.pyx
```diff
@@ -358,3 +380,140 @@ class ManifestItem(PropDict):
...
+
+        attr_list = ['deleted', 'mode', 'source']
+        attr_list += ['uid', 'gid'] if self._numeric_owner else ['user', 'group']
+        for attr in attr_list:
+            if self._item1.get(attr) != self._item2.get(attr):
+                return False
+
+        if 'mode' in self._item1:     # mode of item1 and item2 is equal
+            if (self._item1.is_link() and 'source' in self._item1 and 'source' in self._item2
+                and self._item1.source != self._item2.source):
```
ThomasWaldmann: btw, up to 120 chars is fine if it keeps formatting simpler, no strict 80 chars limit.

imsodin: I saw that in the docs, I adjusted my editor to it. This line would be 140 chars long, hence the break.

ThomasWaldmann: oh, ok then.

imsodin:
> It is easier for review to keep the first two separate.
> 
> But if it is too much hassle to apply the review-fixups to the respective places, collapsing them all is still an option.

No hassle at all, I just wanted to be sure to be in line with your policies. There is some projects that have the requirement that every commit in master should be functional and this will introduce two commits that are definitely not functional (but result in nicer diffs).

imsodin:
No idea why 3.6-dev fails (I don't see any difference to 3.6 which passed) - can you trigger a retest please.

ThomasWaldmann:
restarted.

PR #2846, opened 2017-07-20
Document consequences of changing the segments_per_dir value

wzyboy:
After a user changes `segments_per_dir`, the path returned by
`segment_filename()` is incorrect because it calculates the path with the
new `segments_per_dir` value.

In this case, we try to find the segment path manually.  If we still
can't find the segment, there could be two reasons:

- The segment to be read is missing on disk;
- The segment to be written is not created yet.

For either reason, we return the path calculated with new
`segments_per_dir` value as-is and let the caller decide what to do.

If everything goes well, after some period of time (new data get backed
up while old ones get pruned), segments will eventually be in their
right place.

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -60,8 +60,8 @@ and looks like this::
 
     [repository]
     version = 1
-    segments_per_dir = 10000
-    max_segment_size = 5242880
+    segments_per_dir = 1000
+    max_segment_size = 524288000
     id = 57d6c1d52ce76a836b532b0e42e677dec6af9fca3673db511279358828a21ed6
 
 This is where the ``repository.id`` is stored. It is a unique
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -88,7 +88,10 @@ Segments
 ~~~~~~~~
 
 Objects referenced by a key are stored inline in files (`segments`) of approx.
-500 MB size in numbered subdirectories of ``repo/data``.
+500 MB size in numbered subdirectories of ``repo/data``. The number of segments
+per directory is controlled by the value of ``segments_per_dir``. If you change
+this value in a non-empty repository, you may also need to relocate the segment
+files manually.
 
 A segment starts with a magic number (``BORG_SEG`` as an eight byte ASCII string),
 followed by a number of log entries. Each log entry consists of:


enkore:
Full disclosure: I don't really like this change. For one thing the code in repository.py is very complex and it is often difficult to exactly understand how things are impacted by changes. [1] It is one of the areas where any change has a high probability to introduce bug(s). You'll understand that I'm hesitant to apply changes that further increase the complexity of this code (that includes invariants and expectations, e.g. "s_p_d is accurate"). There was no upfront discussion about this, and changing segments_per_dir can be accomplished right now with some manual work (if any) — when someone manually edits config files with undocumented settings, this has to be expected.

[1] For example, this change suddenly performs IO in a function that previously did not. Thus, it will start to raises errors which may not be expected by other code, or even mishandled.

wzyboy:
Is `segments_per_dir` recorded somewhere else? If so, maybe we could write a helper function to relocate the segment files if the value in config file differs.

ThomasWaldmann:
I agree with @enkore, this is a bit much effort to support a rare case.
One can easily adjust the s_p_d value in a fresh repo, just not in one that has lots of stuff.

A utility function sounds better at first glance, but guess that would need a rpc change, so it is also a bit much...

So, how about just documenting that in case one wants/needs to adjust s_p_d, it is easy to do on a fresh repo, but not so easy (and unsupported by borg code) on a repo that already has multiple segment dirs.

wzyboy:
Thanks for the suggestions, they are what I have not thought of. I have reverted the code change and made this PR a doc-change-only one.

docs/internals/data-structures.rst
```diff
@@ -60,8 +60,8 @@ and looks like this::
 
     [repository]
     version = 1
-    segments_per_dir = 10000
-    max_segment_size = 5242880
+    segments_per_dir = 1000
+    max_segment_size = 524288000
```
ThomasWaldmann: ah, good catch. :)

docs/internals/data-structures.rst
```diff
@@ -88,7 +88,10 @@ Segments
 ~~~~~~~~
 
 Objects referenced by a key are stored inline in files (`segments`) of approx.
-500 MB size in numbered subdirectories of ``repo/data``.
+500 MB size in numbered subdirectories of ``repo/data``. The number of segments
+per directory is controlled by the value of ``segments_per_dir``. If you change
+this value in a non-empty repository, you may also need to relocate the segment
+files manually.
```
ThomasWaldmann: good update.

otoh, considering the default is now only 1000, I guess we are not likely to run into any problems with that (unlike with the 10000 it was before in 1.0).

so, we can add it or not, I am kind of indifferent about it.

wzyboy: I did not know about 1.1.x new features until yesterday :-) My Linux distribution (Arch) has only the stable 1.0.10 version. My repositories was created with 10000 s_p_d. When I tried to change s_p_d from 10000 down to 1000, I encountered FileNotFound. That's why I opened this PR in the first place.

wzyboy:
@ThomasWaldmann Rebased :-)

ThomasWaldmann:
thanks!

klausenbusk:
FWIW if anyone else stumble on this "issue". I ended up using the following script to relocate the segments and change `segments_per_dir`:
```
#!/bin/bash
set -o nounset -o errexit
REPO="${1}"
segments_per_dir="${2}"

sed "s/\(segments_per_dir = \).*/\1${segments_per_dir}/" -i "${REPO}/config"
for s in ${REPO}/data/*/*; do
	sn="$(basename "$s")"
	dir="$(echo ${sn} / ${segments_per_dir} | bc)"
	mkdir -p "${REPO}/data/${dir}"
	if [ "${s}" != "${REPO}/data/${dir}/${sn}" ]; then
		mv "${s}" "${REPO}/data/${dir}/${sn}"
	fi
done
rmdir ${REPO}/data/*
```
Run it as `./foo.sh <path to repo> <new segments_per_dir>`.

wzyboy:
@klausenbusk That's a really usefuly script! Thanks!

My repo was created with borg 1.0.x so s_p_d was 10000. This makes listing the directory contents very slow over mergerfs + nfs (I do not know where the bottleneck is). After changing the s_p_d to 1000 and relocate the segments with your script, the directory listing process is much faster now.

PR #2852, opened 2017-07-22
lgtm fixes

ThomasWaldmann:
https://lgtm.com/projects/g/borgbackup/borg/alerts/

just fixed the easy stuff. any ideas about the rest?

### DIFF
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -34,7 +34,7 @@ MAX_DATA_SIZE = 20971479
 # MAX_OBJECT_SIZE = <20 MiB (MAX_DATA_SIZE) + 41 bytes for a Repository PUT header, which consists of
 # a 1 byte tag ID, 4 byte CRC, 4 byte size and 32 bytes for the ID.
 MAX_OBJECT_SIZE = MAX_DATA_SIZE + 41  # see LoggedIO.put_header_fmt.size assertion in repository module
-assert MAX_OBJECT_SIZE == 20971520 == 20 * 1024 * 1024
+assert MAX_OBJECT_SIZE == 20 * 1024 * 1024
 
 # borg.remote read() buffer size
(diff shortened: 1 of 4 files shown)


src/borg/constants.py
```diff
@@ -34,7 +34,7 @@
 # MAX_OBJECT_SIZE = <20 MiB (MAX_DATA_SIZE) + 41 bytes for a Repository PUT header, which consists of
 # a 1 byte tag ID, 4 byte CRC, 4 byte size and 32 bytes for the ID.
 MAX_OBJECT_SIZE = MAX_DATA_SIZE + 41  # see LoggedIO.put_header_fmt.size assertion in repository module
-assert MAX_OBJECT_SIZE == 20971520 == 20 * 1024 * 1024
+assert MAX_OBJECT_SIZE == 20 * 1024 * 1024  # == 20971520
```
enkore: -1 to this.

src/borg/repository.py
```diff
@@ -642,7 +642,8 @@ def complete_xfer(intermediate=True):
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
                 logger.debug('complete_xfer: deleting unused segment %d', segment)
-                assert self.segments.pop(segment) == 0, 'Corrupted segment reference count - corrupted index or hints'
+                count = self.segments.pop(segment)
+                assert count == 0, 'Corrupted segment reference count - corrupted index or hints'
                 self.io.delete_segment(segment)
```
enkore: good

src/borg/constants.py
ThomasWaldmann: well, if we keep as is, lgtm.com will have that warning in every report...

enkore: Then rather remove the explicit number entirely. (Though I doubt the usefulness of any static analysis unable to suppress false positives.)

PR #2855, opened 2017-07-22
Add .mailmap

lfos:
Coalesce together commits by the same person in git-shortlog(1).

For people with different aliases, prefer full names over abbreviations
or nick names. For people with different email addresses, use the most
recent email address.


ThomasWaldmann:
thanks!

PR #2873, opened 2017-07-24
with-lock, info docs

### LINKED ISSUE #2868: 1.1.0b6: Meaning of "max. archive size" unclear
`borg info` has a new entry:
```
Utilization of max. archive size: 0%
```

So what does this mean?
I assume this is https://github.com/borgbackup/borg/pull/2565, so it is a storage quota, but does not it apply to the whole repo, not only to one archive?
So does not it have to say this?
```
Utilization of max. repo size: 0%
```

Or do you mean?
```
How much size this archive utilises in the whole repo: 0%
```

~~Then I would suggest something like this (it's in a more readable shorter way):~~
~~Part of max. repo size: 0%~~

Well… no, it basically means the same as "Utilization of max. repo size". I think that wording is okay.

### LINKED ISSUE #2869: fix with-lock docs
see there: https://github.com/borgbackup/borg/pull/2867

enkore:
Fixes #2869 
Fixes #2868

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3525,8 +3525,9 @@ class Archiver:
         .. note::
 
             If you copy a repository with the lock held, the lock will be present in
-            the copy, obviously. Thus, before using borg on the copy, you need to
-            use "borg break-lock" on it.
+            the copy. Thus, before using borg on the copy from a different host,
+            you need to use "borg break-lock" on the copied repository, because
+            Borg is cautious and does not automatically remove stale locks made by a different host.
         """)
         subparser = subparsers.add_parser('with-lock', parents=[common_parser], add_help=False,
                                           description=self.do_with_lock.__doc__,
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1415,7 +1415,7 @@ class Archiver:
                 Duration: {duration}
                 Number of files: {stats[nfiles]}
                 Command line: {command_line}
-                Utilization of max. archive size: {limits[max_archive_size]:.0%}
+                Utilization of maximum supported archive size: {limits[max_archive_size]:.0%}
                 ------------------------------------------------------------------------------
                                        Original size      Compressed size    Deduplicated size
                 This archive:   {stats[original_size]:>20s} {stats[compressed_size]:>20s} {stats[deduplicated_size]:>20s}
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1415,7 +1415,7 @@ def format_cmdline(cmdline):
                 Duration: {duration}
                 Number of files: {stats[nfiles]}
                 Command line: {command_line}
-                Utilization of max. archive size: {limits[max_archive_size]:.0%}
+                Utilization of maximum possible archive size: {limits[max_archive_size]:.0%}
```
ThomasWaldmann: s/possible/supported/ ?

src/borg/archiver.py
```diff
@@ -3211,9 +3211,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         are meaning different things:
 
         This archive / deduplicated size = amount of data stored ONLY for this archive
-                                         = unique chunks of this archive.
+        = unique chunks of this archive.
         All archives / deduplicated size = amount of data stored in the repo
-                                         = all chunks in the repository.
+        = all chunks in the repository.
+
+        Borg only supports archives up to a certain size. The size of an archive
```
ThomasWaldmann: well, this is correct, but can be misunderstood by users who do not know that content data lives outside of the archive. so maybe rather say "... archive metadata streams up to ...".

src/borg/archiver.py
```diff
@@ -3211,9 +3211,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+        relative to this limit depends on a number of factors, mainly the number
+        of files, the lengths of paths and other metadata stored for files.
+        This is shown as *utilization of maximum possible archive size*.
```
ThomasWaldmann: also update here if you change above.

src/borg/archiver.py
```diff
@@ -3525,8 +3530,9 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         .. note::
 
             If you copy a repository with the lock held, the lock will be present in
-            the copy, obviously. Thus, before using borg on the copy, you need to
-            use "borg break-lock" on it.
+            the copy. Thus, before using borg on the copy from or on a different host,
```
ThomasWaldmann: not sure about what "from or on" shall mean here.

if you run the client "from" a machine, it means the same as if you run the client "on" a machine.

src/borg/archiver.py
```diff
@@ -3525,8 +3530,9 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+            you need to use "borg break-lock" on it.
+            (Borg is cautious and does not automatically remove stale locks made by a different host)
```
ThomasWaldmann: remove parens, add dot?
or move it to before the final dot of previous sentence?

src/borg/archiver.py
enkore: Yes, that's more clear.

PR #2875, opened 2017-07-24
umount: try fusermount, then try umount

### LINKED ISSUE #2863: Can mount, but not umount
v1.0.9

STR (working as a non-root user by default):
1. mkdir mnt
2. setting repo/passphrase
3. ```shell
    $ borg mount :: ./mnt
    fuse: failed to exec fusermount: No such file or directory
    ``` 
4. Retried with `sudo borg mount :: ./mnt` works. (`borg mount :: ./mnt  4,81s user 0,75s system 27% cpu 19,883 total`) on a btrfs filesystem BTW. (Also set repo etc. again, so it works.)
5. browse in it
(interestingly for non-root users no username/password were shown and they could not access it)
6. Unmounting fails:
```shell 
borg umount ./mnt
Local Exception.
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 2052, in main
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1997, in run
    return func(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 564, in do_umount
    return umount(args.mountpoint)
  File "borg/platform_linux.pyx", line 148, in borg.platform_linux.umount (borg/platform_linux.c:4217)
  File "/usr/lib/python3.5/subprocess.py", line 247, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.5/subprocess.py", line 676, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1282, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'fusermount'

Platform: XXXXXXX #2 SMP Sat May 6 19:26:50 UTC 2017 armv7l 
Linux: debian 9.0 
Borg: 1.0.9  Python: CPython 3.5.3
PID: 9277  CWD XXXXX
sys.argv: ['/usr/bin/borg', 'umount', './mnt']
SSH_ORIGINAL_COMMAND: None
```

With `umount ./mnt` I could finally unmount the repo anyway…

enkore:
Fixes #2863

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -2277,3 +2277,10 @@ def dash_open(path, mode):
 
 def is_terminal(fd=sys.stdout):
     return hasattr(fd, 'isatty') and fd.isatty() and (sys.platform != 'win32' or 'ANSICON' in os.environ)
+
+
+def umount(mountpoint):
+    try:
+        return subprocess.call(['fusermount', '-u', mountpoint])
+    except FileNotFoundError:
+        return subprocess.call(['umount', mountpoint])
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -66,11 +66,12 @@ from .helpers import replace_placeholders
 from .helpers import ChunkIteratorFileWrapper
 from .helpers import popen_with_error_handling
 from .helpers import dash_open
+from .helpers import umount
 from .nanorst import rst_to_terminal
 from .patterns import ArgparsePatternAction, ArgparseExcludeFileAction, ArgparsePatternFileAction, parse_exclude_pattern
 from .patterns import PatternMatcher
 from .item import Item
-from .platform import get_flags, umount, get_process_id, SyncFile
+from .platform import get_flags, get_process_id, SyncFile
 from .remote import RepositoryServer, RemoteRepository, cache_if_remote
 from .repository import Repository, LIST_SCAN_LIMIT
 from .selftest import selftest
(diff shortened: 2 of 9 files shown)


src/borg/helpers.py
```diff
@@ -2277,3 +2277,10 @@ def dash_open(path, mode):
 
 def is_terminal(fd=sys.stdout):
     return hasattr(fd, 'isatty') and fd.isatty() and (sys.platform != 'win32' or 'ANSICON' in os.environ)
+
+
+def umount(mountpoint):
+    try:
+        return subprocess.call(['fusermount', '-u', mountpoint])
```
ThomasWaldmann: is there `fusermount` on anything else than linux?

would work, but maybe a bit unpretty. if we ever have windows support with fuse, we also would likely have to move it back to platform package.

enkore: For a single-use four-line function I consider it "prettier" to just define it in one place, rather than spreading it out across 6+ modules.

ThomasWaldmann: why? -> because I suspect windows will have its own way (as usual) rather than "umount" or "fusermount".

agreed that it's early to think about that considering no native windows code is merged yet.
but considering we already have the platform code, we don't need to remove it if we can already see it might be useful later (and even now, for fusermount linux special case).

enkore: Guess we'll see it when we get there

PR #2880, opened 2017-07-25
hashindex: use except NULL

enkore:


src/borg/hashindex.pyx
```diff
@@ -22,7 +22,7 @@ cdef extern from "_hashindex.c":
         uint32_t version
         char hash[16]
 
-    HashIndex *hashindex_read(object file_py, int permit_compact) except *
+    HashIndex *hashindex_read(object file_py, int permit_compact) except NULL
```
ThomasWaldmann: http://cython.readthedocs.io/en/latest/src/reference/language_basics.html#checking-return-values-for-non-cython-functions

that was the first thing that showed up when i searched for "cython except NULL".

ThomasWaldmann: ah, purpose is to really propagate exceptions from there (as hashindex_read calls python c api).

src/borg/hashindex.pyx
```diff
@@ -92,7 +92,6 @@ cdef class IndexBase:
                     self.index = hashindex_read(fd, permit_compact)
             else:
                 self.index = hashindex_read(path, permit_compact)
-            assert self.index, 'hashindex_read() returned NULL with no exception set'
```
ThomasWaldmann: maybe rather keep that?

enkore: I was going to say that probably Cython generates code to handle this case, but in fact the /*except:*/ code relies on an exception being set and the branch to it only depends on ret==NULL. Btw., this patch produces virtually the same code as before, including the temporary, so it won't fix the false positive of that tool.

PR #2887, opened 2017-07-29
multithreading: add pyzmq dep, chunker GIL

enkore:
The GIL change actually was one of the last changes I made on the prototype, but it's rather straightforward, and with some GIL stuff already scattered across the code base I don't see a good reason to not just add it.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -142,6 +142,7 @@ following dependencies first:
 * OpenSSL_ >= 1.0.0, plus development headers.
 * libacl_ (which depends on libattr_), both plus development headers.
 * liblz4_, plus development headers.
+* ZeroMQ_ >= 4.0.0, plus development headers.
 * some Python dependencies, pip will automatically install them for you
 * optionally, the llfuse_ Python package is required if you wish to mount an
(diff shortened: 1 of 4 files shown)


docs/installation.rst
```diff
@@ -142,6 +142,7 @@ following dependencies first:
 * OpenSSL_ >= 1.0.0, plus development headers.
 * libacl_ (which depends on libattr_), both plus development headers.
 * liblz4_, plus development headers.
+* ZeroMQ_ >= 4.0.0, plus development headers.
```
ThomasWaldmann: i noticed that binary builds of the pyzmq package contain libzmq, so in that case, it is not needed to install them separately (at least as we do not need it for our own code).

enkore: I expected it would still be required to build pyzmq. There seems to be a debian package for pyzmq, which depends on libzmq. Not sure what to make of that, for now I'd say it's sufficient to note that in one form or another this is required.

ThomasWaldmann: ok course debian has cleanly split packages, but if you list pyzmq pypi package files, you'll find the lib zeromq .so.

Issue #1031, opened 2016-05-07
add flexible encryption chunk header


ThomasWaldmann:
I've looked at the experimental branch, thinking about what to get from there and how.
## Compression

The flexible compression first done there was meanwhile implemented in borg - a bit differently. The idea here was adding a 2-byte header in front of the compressed data that tells the (de)compression algorithm. Plus the small hack that we can keep existing zlib chunks without header, because we can determine zlib from first 2 zlib-output bytes also. We just must not use zlib-looking 2 type bytes for all other compression algorithms.

```
Old: N bytes zlib,6 output (always, attic heritage)

New: 2 bytes compression type (except for zlib) + N bytes compressor output
```

This is nice for layering, we just give the decompressor the whole bytestring and it can make up everything from that (choosing the right algo internally). The compressor just prepends the right 2 bytes when returning its output.
## Encryption

I'ld like to add flexible encryption in a similar way - see link in next post.

enkore:
Mmmhh, as I mentioned in [that thing](https://gist.github.com/enkore/d16849b9e2eecdab0903bcd37bd0ee27) it imho makes more sense to think in terms of cipher suites that provide AEAD in a black-box fashion, without depending on having separable MAC and ciphers.

ThomasWaldmann:
the meta_tuple would be "AD" (additional data), also protected by the mac.

enkore:
Yes, so why not push all that out of the chunk format into blackbox-ciphersuite-do-it-yourself? ;)

ThomasWaldmann:
cipher, hmac type could be stored with the key, right, but the IV needs to be stored with the chunk as it is different per chunk (either a counter or random).

enkore:
> but the IV needs to be stored with the chunk as it is 

Absolutely. I'm just saying that that should be the responsibility of the ciphersuite, not some metadata prepended to the encrypted payload (output of the ciphersuite).

@ new key per chunk: would need lots of high quality randomness though.

enkore:
I updated the link above with some ideas from here.

- Use msgpack + structured data for keys instead of passwd-like stuff
- 16 byte key IDs instead of 4 byte --> maybe just use UUIDv4 format.
- Added an API proposal
- Removed "user interface" section

ThomasWaldmann:
hmm, not sure if i like msgpack for the keys - something readable feels better there. json?

[...]

PR #2888, opened 2017-07-29
crypto: fixes & remove AES-GCM

enkore:
Fixups for #1034

### DIFF
--- a/src/borg/crypto/low_level.pyx
+++ b/src/borg/crypto/low_level.pyx
@@ -222,8 +222,8 @@ cdef class AES256_CTR_BASE:
     cdef unsigned char iv[16]
     cdef long long blocks
 
-    @staticmethod
-    def requirements_check():
+    @classmethod
+    def requirements_check(cls):
         if OPENSSL_VERSION_NUMBER < 0x10000000:
             raise ValueError('AES CTR requires OpenSSL >= 1.0.0. Detected: OpenSSL %08x' % OPENSSL_VERSION_NUMBER)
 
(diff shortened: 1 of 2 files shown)


src/borg/crypto/low_level.pyx
```diff
@@ -67,7 +67,6 @@ cdef extern from "openssl/evp.h":
         pass
 
     const EVP_CIPHER *EVP_aes_256_ctr()
-    const EVP_CIPHER *EVP_aes_256_gcm()
```
ThomasWaldmann: i intentionally did not remove gcm although some people have bad feelings about it. other people have also bad feelings about patents (even with the patent license for FOSS and openssl), so there would be gcm as option for them.

technically, I'ld also choose ocb over gcm.

src/borg/crypto/low_level.pyx
```diff
@@ -223,8 +222,8 @@ cdef class AES256_CTR_BASE:
     cdef unsigned char iv[16]
     cdef long long blocks
 
-    @staticmethod
-    def requirements_check():
+    @classmethod
+    def requirements_check(cls):
```
ThomasWaldmann: interesting, i didn't notice malfunction here. did you file a bug at cython?

enkore: It's not a malfunction, just slightly incorrect code.

```
foo something(); // this is not a prototype for the function below

foo
something() { // this function was not declared above!
 return ...;
}
```

src/borg/crypto/low_level.pyx
enkore: chapoly > either, like I argued previously. With no performance impact.

enkore: See https://github.com/borgbackup/borg/issues/1031#issuecomment-304049587

enkore: So, remove OCB as well.?

ThomasWaldmann: no, keep it, it is hw accelerated and fastest.

PR #2889, opened 2017-07-29
multithreading: item.to_optr(), Item.from_optr()

enkore:
Attached to Item() because that's the only expected usage for now.

### DIFF
--- a/src/borg/_item.c
+++ b/src/borg/_item.c
@@ -1,0 +1,41 @@
+#include "Python.h"
+
+/*
+ * This is not quite as dark magic as it looks. We just convert the address of (pointer to)
+ * a PyObject into a bytes object in _wrap_object, and convert these bytes back to the
+ * pointer to the original object.
+ *
+ * This mainly looks a bit confusing due to our mental special-casing of "char*" from other
+ * pointers.
+ *
+ * The big upside to this is that this neither does *any* serialization (beyond creating tiny
+ * bytes objects as "stand-ins"), nor has to copy the entire object that's passed around.
+ */
+
+static PyObject *
+_object_to_optr(PyObject *obj)
+{
+    /*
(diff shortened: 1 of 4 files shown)


src/borg/_item.c
```diff
@@ -0,0 +1,41 @@
...
+        PyErr_SetString(PyExc_TypeError, "Cannot unwrap non-bytes object");
+        return NULL;
+    }
+    if(PyBytes_Size(bytes) != sizeof(void*)) {
+        PyErr_SetString(PyExc_TypeError, "Invalid length of bytes object");
+        return NULL;
+    }
+    PyObject *object = * (PyObject **) PyBytes_AsString(bytes);
+    return object;
+}
```
ThomasWaldmann: maybe rename `*wrap*` - this is not about wrapping, rather about getting a pointer / addr / ref.

src/borg/item.pyx
```diff
@@ -227,6 +232,26 @@ class Item(PropDict):
...
+        that can be passed exactly once to Item.from_optr to get this
+        object back.
+
+        to_optr/from_optr must be used symmetrically,
+        don't call from_optr multiple times.
+
+        This object can't be deallocated after a call to to_optr()
+        until from_optr() is called.
+        """
+        return _wrap_object(self)
```
ThomasWaldmann: _to_optr maybe?

src/borg/item.pyx
```diff
@@ -227,6 +232,26 @@ class Item(PropDict):
...
+
+        to_optr/from_optr must be used symmetrically,
+        don't call from_optr multiple times.
+
+        This object can't be deallocated after a call to to_optr()
+        until from_optr() is called.
+        """
+        return _object_to_optr(self)
+
+    @classmethod
```
imsodin: For my education: What is the reason this is not a `@staticmethod`?

PR #2900, opened 2017-07-30
Only compare contents when chunker params match (fixes #2899)

### LINKED ISSUE #2899: Comparing modified items incorrectly assumes comparable chunk ids
This came up in #2842 (see during https://github.com/borgbackup/borg/pull/2842/files#diff-0c1790c20ac3974e84f97b2ded80507bR449) but the problem was there before https://github.com/borgbackup/borg/blob/master/src/borg/archiver.py#L1028.
The current solution of comparing just chunk ids is fine as long as the two archives were created with the same chunker params, but doesn't work otherwise.
I am not sure what would be the cheapest way to get matching contents from two chunk iterators. Maybe creating new chunks from the chunks with the same params?

imsodin:
See issue #2899 for a description of the problem.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1021,18 +1021,19 @@ class Archiver:
         def compare_content(path, item1, item2):
             if contents_changed(item1, item2):
                 if item1.get('deleted'):
-                    return ('added {:>13}'.format(format_file_size(sum_chunk_size(item2))))
-                elif item2.get('deleted'):
-                    return ('removed {:>11}'.format(format_file_size(sum_chunk_size(item1))))
-                else:
-                    chunk_ids1 = {c.id for c in item1.chunks}
-                    chunk_ids2 = {c.id for c in item2.chunks}
-                    added_ids = chunk_ids2 - chunk_ids1
-                    removed_ids = chunk_ids1 - chunk_ids2
-                    added = sum_chunk_size(item2, added_ids)
-                    removed = sum_chunk_size(item1, removed_ids)
-                    return ('{:>9} {:>9}'.format(format_file_size(added, precision=1, sign=True),
-                                                 format_file_size(-removed, precision=1, sign=True)))
+                    return 'added {:>13}'.format(format_file_size(sum_chunk_size(item2)))
+                if item2.get('deleted'):
+                    return 'removed {:>11}'.format(format_file_size(sum_chunk_size(item1)))
+                if not can_compare_chunk_ids:
+                    return 'modified'
+                chunk_ids1 = {c.id for c in item1.chunks}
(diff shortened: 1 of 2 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -3307,9 +3308,10 @@ def do_asserts(output, archive):
             # The inode has two links and the file contents changed. Borg
             # should notice the changes in both links. However, the symlink
             # pointing to the file is not changed.
-            assert '0 B input/empty' in output
+            change = '{:<19}'.format('modified') if archive == '1b' else '0 B'
```
ThomasWaldmann: somehow that "modified", "1b", "0 B" is rather confusing.

src/borg/archiver.py
```diff
@@ -1022,17 +1022,18 @@ def compare_content(path, item1, item2):
...
-                    chunk_ids1 = {c.id for c in item1.chunks}
-                    chunk_ids2 = {c.id for c in item2.chunks}
-                    added_ids = chunk_ids2 - chunk_ids1
-                    removed_ids = chunk_ids1 - chunk_ids2
-                    added = sum_chunk_size(item2, added_ids)
-                    removed = sum_chunk_size(item1, removed_ids)
-                    return ('{:>9} {:>9}'.format(format_file_size(added, precision=1, sign=True),
-                                                 format_file_size(-removed, precision=1, sign=True)))
+                if not can_compare_chunk_ids:
+                    return ('modified')
```
imsodin: We know that it was modified in this case as well due to https://github.com/borgbackup/borg/pull/2900/files/a724f2fab3609eb99be4530212d06324c846e309#diff-48f2b9140a35ffd45a46ccfbe63931a6R1022

imsodin:
The test failure is unrelated to the changes - should be retested (didn't see an option to trigger a retest myself).

src/borg/archiver.py
ThomasWaldmann: link does not work

imsodin: https://github.com/borgbackup/borg/pull/2900/files#diff-48f2b9140a35ffd45a46ccfbe63931a6R1022
Basically this entire part is behind `if contents_changed(item1, item2):` which actually checks chunk contents if ids are not comparable.

imsodin:
@ThomasWaldmann Yes, nothing to do from my side.

ThomasWaldmann:
can you do the 1.1-maint backport also?
and afterwards rebase your other PR onto current master?

ThomasWaldmann:
```
git cherry-pick <changesethash> --strategy recursive -X rename-threshold=5% -x
```
That's useful for backporting, esp. if stuff was moved around.

PR #2902, opened 2017-07-30
mt-2a

enkore:


enkore:
- Hardlinks are timing-dependent for now, so a bunch of tests don't work any more (I'm surprised how many tests rely on this). To fix this, something like an "anticipated" HM master list is needed, and "anticipated" HM referrers are deferred until their anticipated HM is processed or failed.
- Didn't do error handling yet.
- In two tests there is an mtime issue with "dir2", couldn't reproduce it outside the tests.
- Because it's kinda impossible to interrupt threads, any uncaught exception signals. This is kinda annoying for the test suite, doubly so since most of the time the output is suppressed. This has been a big issue previously as well, so it should be fixed...
- RepositoryService.API isn't implemented yet, so nonce reservations will probably break stuff at least with remote repos.

src/borg/helpers.py
```diff
@@ -801,7 +801,7 @@ def format_archive(archive):
     )
 
 
-class Buffer:
+class Buffer(threading.local):
```
ThomasWaldmann: s/TLS/thread local storage/ just to not confuse people.

enkore:
Looks like it throws/asserts on macos. Yay!

src/borg/threading/create.py
```diff
@@ -624,6 +624,13 @@ def pack_and_add_item(self):
         self.num_items += 1
         if 'chunks' in item:
             self.archive.stats.nfiles += 1
+        try:
+            # XXX This is the sledgehammer approach of getting rid of these, see comment in Item
+            del item.original_path
+            del item.status
+            del item.num_chunks
+        except AttributeError:
+            pass
```
ThomasWaldmann: hmm, you seem to expect that any of them could be missing (thus the scope around all 3).
but the first one missing will raise and the remaining ones won't be killed.
so, 3x try/except?

enkore: These are added when an item was processed by the pipeline, then all three should be present; if the item was not processed by the pipeline, none should be present.

I want an overall nicer way for this, though, that is safer to use and makes it more clear which attributes are runtime only and which are actually stored (see item.pyx).

src/borg/archive.py
```diff
@@ -469,10 +476,17 @@ def save(self, name=None, comment=None, timestamp=None, additional_metadata=None
         metadata.update(additional_metadata or {})
         metadata = ArchiveItem(metadata)
         data = self.key.pack_and_authenticate_metadata(metadata.as_dict(), context=b'archive')
+        return data
+
+    def save(self, name=None, comment=None, timestamp=None, additional_metadata=None):
+        name = self._check_name(name)
+        self.items_buffer.flush(flush=True)
+        data = self.pack_archive_item(name, comment, timestamp, additional_metadata)
```
ThomasWaldmann: hmm, so _check_name is called twice (once by pack_archive_item)?

if intended, maybe add a comment about why.

src/borg/testsuite/archiver.py
```diff
@@ -743,6 +744,7 @@ def _extract_hardlinks_setup(self):
     requires_hardlinks = pytest.mark.skipif(not are_hardlinks_supported(), reason='hardlinks not supported')
 
     @requires_hardlinks
+    @pytest.mark.skip('Relies on strict hardlink semantics')
```
ThomasWaldmann: maybe explain that once in the source.

.travis/install.sh
```diff
@@ -4,8 +4,6 @@ set -e
 set -x
 
 if [[ "$(uname -s)" == 'Darwin' ]]; then
-    brew update || brew update
```
ThomasWaldmann: hmm, i also wondered whether travis is running some brew update in the background?
sometimes seen a message about it already running.
but how are tests supposed to be stable if there is a upgrade running in the background?

ThomasWaldmann: hmm, how is that related to mt?

enkore: Its not, but I finally snapped and removed it after it failed yet another couple of my macos builds.

enkore:
Status:

- macOS test failures down to one mtime issue and a FD issue ("Too many open files"). The latter doesn't seem to happen on Linux, and lsof does not show any leaked or excessive use of FDs on Linux. I can also run the entire suite in one process, which would exacerbate the problem, but nothing happens. So this one might be actually difficult to debug.
- Some added flakiness.
  - Sometimes a test failure seems to trigger a pytest-xdist bug which causes it to assert like this:

    ```
    INTERNALERROR>   File "/home/travis/build/borgbackup/borg/.tox/py35/lib/python3.5/site-packages/xdist/dsession.py", line 596, in slave_slavefinished
    INTERNALERROR>     assert not crashitem, (crashitem, node)
    INTERNALERROR> AssertionError: ('.tox/py35/lib/python3.5/site-packages/borg/testsuite/archive.py::ArchiveTimestampTestCase::test_with_microseconds', <SlaveController gw1>)
    ```

    This seems to cause an inconsistent system state in py.test which makes it think that a number of tests were not collected or something like that - it tends to exit green, with a low number of passed tests, if no other tests fail.

  - For the first time I now saw a bad interaction between pytest-timeout and something else:

    ```
    INTERNALERROR>     signal.signal(signal.SIGALRM, handler)
    INTERNALERROR>   File "/opt/python/3.5.3/lib/python3.5/signal.py", line 47, in signal
    INTERNALERROR>     handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
    INTERNALERROR> ValueError: signal only works in main thread
    ```
  
  - test_compression_XYZ_(un)compressible is rarely flaky, looks like a genuine MT bug, so I'll probably look into that more closely.

enkore:
My plan going forward is to mainline as much as possible of this in separate patches and clean as much up as possible before adding major new stuff.

ThomasWaldmann:
Why did you close it?

I'ld actually have some time to review after 1.1.1.

PR #2905, opened 2017-07-31
Refactor helpers

### LINKED ISSUE #2904: split helpers module, create package
the all-stuff-that-didnt-fit-elsewhere `borg.helpers` module is rather big and should be split into modules within a `helpers` package that have some specific topic.

Old: `helpers.py`

New:
```
helpers/
  __init__.py   # imports stuff from other modules to stay compatible at first
  foo.py
  bar.py
  ...
```

ThomasWaldmann:
fixes #2904

no code changes except moving stuff around and fixing imports.
also did some micro changes to docstring style.

suggestions about remaining "misc" module contents welcome, so "misc" won't become the new monster.

todo: testsuite/helpers.py is still one big monster, refactor that after package layout of tested code is finalized.


ThomasWaldmann:
@enkore merge this before you do more changes to helpers?

src/borg/helpers/datastruct.py
```diff
@@ -0,0 +1,56 @@
+import threading
+from collections import namedtuple
+
+from .errors import Error
+
+
+ArchiveInfo = namedtuple('ArchiveInfo', 'name id ts')
```
enkore: Move to .manifest?

src/borg/helpers/parseformat.py
```diff
@@ -0,0 +1,917 @@
+import argparse
```
enkore: Merge .convert into this?

src/borg/helpers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from .fs import *  # NOQA
+from .manifest import *  # NOQA
+from .misc import *  # NOQA
+from .msgpack import *  # NOQA
+from .parseformat import *  # NOQA
+from .process import *  # NOQA
+from .progress import *  # NOQA
+from .time import *  # NOQA
+from .usergroup import *  # NOQA
+from .yes import *  # NOQA
```
enkore: Add `__all__`.

src/borg/helpers/datastruct.py
ThomasWaldmann: it initially was there and i moved to avoid some cyclic import while doing the refactoring.
i can try again if it works now.

src/borg/helpers/__init__.py
ThomasWaldmann: why?

ThomasWaldmann: all names from the modules are imported into helpers.

with `__all__` and a `from borg.helpers import *` one would get borg.progress.*, which is different.

ThomasWaldmann:
@enkore is `__all__` now like you meant? it's maybe not useful in the current state of the code, but if we ever directly import from helpers.foo, it could be useful.

anything else besides collapsing changesets?

src/borg/helpers/__init__.py
enkore: (Overlooked this in my inbox) I'm mistaken here, __all__ isn't useful since there is no `from helpers import *` anywhere. __all__ might have some minor use in the individual helpers.XY modules, to avoid surprises with shadowed names, but it's not anything important, probably not worth it.

ThomasWaldmann:
ok, removed `__all__` again and collapsed changesets - no other changes.

enkore:
Rebase needed

ThomasWaldmann:
crap, merging stuff in wrong order now caused a huge conflict. :(

PR #2915, opened 2017-08-05
Only compare contents when chunker params match (fixes #2899)

### LINKED ISSUE #2899: Comparing modified items incorrectly assumes comparable chunk ids

imsodin:
cherry pick of #2900

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1013,18 +1013,19 @@ class Archiver:
         def compare_content(path, item1, item2):
             if contents_changed(item1, item2):
                 if item1.get('deleted'):
-                    return ('added {:>13}'.format(format_file_size(sum_chunk_size(item2))))
-                elif item2.get('deleted'):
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/jobs/261423443 wat?

imsodin:
Nope.

ThomasWaldmann:
thanks!

Issue #2376, opened 2017-04-01
JSON: Use a more standard, timezone-aware format date time format


enkore:
Cf. #2368

Can be implemented similar to the output-type dependent file-size formatting.

enkore:
This is a fairly simple change, but I think it should really be made for 1.1.

ThomasWaldmann:
@enkore borg currently does a lot using localtime, but has no real timezone support (likely would require pytz). So I am not totally sure that even now all timestamps shown are totally correct (esp. regarding tzoffset and dst).

So how's your idea about this change being "fairly simple"?

enkore:
This issue would just be about using a ISO8601 format (and output in UTC), not true TZ support, which would probably be bigger.

ThomasWaldmann:
Ah, OK, so I'll try that...

ThomasWaldmann:
A first step could be to use the yyyy-mm-ddThh:mm:ss format without tz (so, no weekday and the T separator would be the difference to the human-readable iso-like format we use).

ThomasWaldmann:
partially addressed by #2925 (no tz yet), nothing to do left for 1.1 milestone.

sophie-h:
Would it maybe be an option for 1.2 to begin with storing an additional datetime with timezone for archive `start` and `end` times? And maybe also output them in a (possibly unstable/transient `x-start-tz`) json keys, if a datetime with timezone is available as well?

sophie-h:
@ThomasWaldmann Maybe 2.0 is a good chance for doing everything properly with timezones?

ThomasWaldmann:
Yeah, would be good timing. But already a ton of other todo, do you want to work on it?

ThomasWaldmann:
I started working on it...

ThomasWaldmann:
Python (C?) wart: seems like strftime and strptime disagree on tz being with/without ":" between HH and MM.

PR #2916, opened 2017-08-06
json: use isoformat

ThomasWaldmann:
a first step for #2376, see commit comments.

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -717,9 +717,15 @@ def safe_timestamp(item_timestamp_ns):
 
 
 def format_time(t):
-    """use ISO-8601 date and time format
+    """use ISO-8601-like date and time format (human readable, with wkday and blank date/time separator)
     """
     return t.strftime('%a, %Y-%m-%d %H:%M:%S')
+
+
+def isoformat_time(t):
+    """use ISO-8601 date and time format (machine readable, no wkday, no microseconds either)
+    """
+    return t.strftime('%Y-%m-%dT%H:%M:%S')  # note: first make all datetime objects tz aware before adding %z here.
 
 
 def format_timedelta(td):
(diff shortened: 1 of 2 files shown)


src/borg/helpers.py
```diff
@@ -1781,11 +1802,6 @@ def __init__(self, archive, format, *, json_lines=False):
         for hash_function in hashlib.algorithms_guaranteed:
             self.add_key(hash_function, partial(self.hash_item, hash_function))
         self.used_call_keys = set(self.call_keys) & self.format_keys
-        if self.json_lines:
-            self.item_data = {}
-            self.format_item = self.format_item_json
-        else:
-            self.item_data = static_keys
```
ThomasWaldmann: needed to move this upwards, so the time formatting gets set up before it is used.

ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/jobs/261426845 wat?

src/borg/helpers.py
```diff
@@ -717,11 +717,17 @@ def safe_timestamp(item_timestamp_ns):
...
-    """use ISO-8601 date and time format
+    """use ISO-8601-like date and time format (human readable, with wkday and blank date/time separator)
     """
     return t.strftime('%a, %Y-%m-%d %H:%M:%S')
 
 
+def isoformat_time(t):
+    """use ISO-8601 date and time format (machine readable, no wkday, no microseconds either)
+    """
+    return t.strftime('%Y-%m-%dT%H:%M:%S')  # note: first make all datetime objects tz aware before adding %z here.
```
enkore: Could t.isoformat() be used here, or does it something wrong (with TZ or so)?

ThomasWaldmann:
added tests, collapsed a typo fix. merge?

ThomasWaldmann:
about datetime.isoformat: the new spec param is only available since 3.6 and I guess before it sometimes added the microseconds OR NOT (if == 0). I wanted a specific format, not one depending on the value, thus I used strftime.

ThomasWaldmann:
note: i reset master to the previous cs of this merge. so a new rebase pr of this will need to get merged again.

ThomasWaldmann:
see #2928.

Issue #2444, opened 2017-04-25
review IntegrityErrors, especially when they kill


ThomasWaldmann:
at some places (borg check --repair) we kill repository objects when we encounter IntegrityErrors on them.

for many cases this is the right thing to do - if an object is corrupt, it is not useful any more.

but when reading the LoggedIO code, i got the impression that some of the IEs raised there could be due to:
- corrupt repository index just pointing at some bad segment/offset where no object starts
- corrupt repository index pointing at a valid object, but for a different ID (maybe extremely unlikely, but who knows?)
- too large objects, https://github.com/borgbackup/borg/blob/1.1.0b4/src/borg/repository.py#L1175
- other cases?

https://www.bountysource.com/issues/44509028-review-integrityerrors-especially-when-they-kill

ThomasWaldmann:
closing due to #5230.

ThomasWaldmann:
reopening, see #5230.

PR #2920, opened 2017-08-06
Detect non-upgraded Attic repositories

### LINKED ISSUE #1933: Generic error message when trying to use non-upgraded attic repository

lfos:
When coming accross an invalid segment, explicitly check for the Attic
segment MAGIC. If an Attic segment is detected, raise a dedicated
exception, telling the user to upgrade the repository first.

Fixes #1933.

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -726,6 +726,11 @@ This problem will go away as soon as the server has been upgraded to 1.0.7+.
                     raise IntegrityError('(not available)')
                 else:
                     raise IntegrityError(args[0].decode())
+            elif error == 'AtticRepository':
+                if old_server:
+                    raise Repository.AtticRepository('(not available)')
+                else:
+                    raise Repository.AtticRepository(args[0].decode())
             elif error == 'PathNotAllowed':
                 if old_server:
                     raise PathNotAllowed('(unknown)')
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -30,6 +30,8 @@ logger = create_logger(__name__)
 
 MAGIC = b'BORG_SEG'
 MAGIC_LEN = len(MAGIC)
+ATTIC_MAGIC = b'ATTICSEG'
+assert len(ATTIC_MAGIC) == MAGIC_LEN
 TAG_PUT = 0
 TAG_DELETE = 1
 TAG_COMMIT = 2
@@ -116,6 +118,9 @@ class Repository:
     class InvalidRepository(Error):
         """{} is not a valid repository. Check repo config."""
 
+    class AtticRepository(Error):
+        """Attic repository detected. Please run "borg upgrade {}"."""
+
     class CheckNeeded(ErrorWithTraceback):
         """Inconsistency detected. Please run "borg check {}"."""
 
@@ -134,7 +139,7 @@ class Repository:
         """The storage quota ({}) has been exceeded ({}). Try deleting some archives."""
 
     def __init__(self, path, create=False, exclusive=False, lock_wait=None, lock=True,
-                 append_only=False, storage_quota=None):
+                 append_only=False, storage_quota=None, check_segment_magic=True):
         self.path = os.path.abspath(path)
         self._location = Location('file://%s' % self.path)
         self.io = None  # type: LoggedIO
@@ -154,6 +159,7 @@ class Repository:
         self.storage_quota = storage_quota
         self.storage_quota_use = 0
         self.transaction_doomed = None
+        self.check_segment_magic = check_segment_magic
 
     def __del__(self):
         if self.lock:
@@ -370,6 +376,12 @@ class Repository:
             self.storage_quota = self.config.getint('repository', 'storage_quota', fallback=0)
         self.id = unhexlify(self.config.get('repository', 'id').strip())
         self.io = LoggedIO(self.path, self.max_segment_size, self.segments_per_dir)
+        if self.check_segment_magic:
+            # read a segment and check whether we are dealing with a non-upgraded Attic repository
+            segment = self.io.get_latest_segment()
+            if segment is not None and self.io.get_segment_magic(segment) == ATTIC_MAGIC:
+                self.close()
+                raise self.AtticRepository(path)
 
     def close(self):
         if self.lock:
@@ -1244,6 +1256,11 @@ class LoggedIO:
 
     def segment_size(self, segment):
         return os.path.getsize(self.segment_filename(segment))
+
+    def get_segment_magic(self, segment):
+        fd = self.get_fd(segment)
+        fd.seek(0)
+        return fd.read(MAGIC_LEN)
 
     def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
         """
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -55,6 +55,7 @@ from . import has_lchflags, has_llfuse
 from . import BaseTestCase, changedir, environment_variable, no_selinux
 from . import are_symlinks_supported, are_hardlinks_supported, are_fifos_supported, is_utime_fully_supported
 from .platform import fakeroot_detected
+from .upgrader import attic_repo
 from . import key
 
 
@@ -2724,6 +2725,27 @@ id: 2 / e29442 3506da 4e1ea7 / 25f62a 5a3d41 - 02
             assert os.stat('input/dir1/subdir/hardlink').st_nlink == 2
             assert os.stat('input/dir1/aaaa').st_nlink == 2
             assert os.stat('input/dir1/source2').st_nlink == 2
+
+    def test_detect_attic_repo(self):
+        path = attic_repo(self.repository_path)
+        cmds = [
+            ['create', path + '::test', self.tmpdir],
+            ['extract', path + '::test'],
+            ['check', path],
+            ['rename', path + '::test', 'newname'],
+            ['list', path],
+            ['delete', path],
+            ['prune', path],
+            ['info', path + '::test'],
+            ['mount', path, self.tmpdir],
+            ['key', 'export', path, 'exported'],
+            ['key', 'import', path, 'import'],
+            ['change-passphrase', path],
+            ['break-lock', path],
+        ]
+        for args in cmds:
+            output = self.cmd(*args, fork=True, exit_code=2)
+            assert 'Attic repository detected.' in output
 
 
 @unittest.skipUnless('binary' in BORG_EXES, 'no borg.exe available')
--- a/src/borg/testsuite/upgrader.py
+++ b/src/borg/testsuite/upgrader.py
@@ -85,8 +85,8 @@ def test_convert_segments(attic_repo, inplace):
     :param attic_repo: a populated attic repository (fixture)
     """
     repo_path = attic_repo
-    # check should fail because of magic number
-    assert not repo_valid(repo_path)
+    with pytest.raises(Repository.AtticRepository):
+        repo_valid(repo_path)
     repository = AtticRepositoryUpgrader(repo_path, create=False)
     with repository:
         segments = [filename for i, filename in repository.io.segment_iterator()]
@@ -149,8 +149,8 @@ def test_convert_all(attic_repo, attic_key_file, inplace):
     """
     repo_path = attic_repo
 
-    # check should fail because of magic number
-    assert not repo_valid(repo_path)
+    with pytest.raises(Repository.AtticRepository):
+        repo_valid(repo_path)
 
     def stat_segment(path):
         return os.stat(os.path.join(path, 'data', '0', '0'))
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -499,6 +499,8 @@ Errors
         Insufficient free space to complete transaction (required: {}, available: {}).
     Repository.InvalidRepository
         {} is not a valid repository. Check repo config.
+    Repository.AtticRepository
+        Attic repository detected. Please run "borg upgrade {}".
     Repository.ObjectNotFound
         Object with key {} not found in repository {}.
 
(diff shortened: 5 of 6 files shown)


src/borg/repository.py
```diff
@@ -872,7 +879,7 @@ def report_error(msg):
                 continue
             try:
                 objects = list(self.io.iter_objects(segment))
-            except IntegrityError as err:
+            except (IntegrityError, LoggedIO.AtticSegment) as err:
                 report_error(str(err))
                 objects = []
                 if repair:
```
ThomasWaldmann: hmm, is it correct if it would attempt to repair here?

lfos: If am not 100% sure. Looking at the code, it seems like a repair would convert the Attic segments into proper borg segments. So maybe, we would end up with a semi-converted repository? Or does `borg check` also take care of other places where things need to be converted and perform a full conversion? Either way, this is just backwards-compatible behavior. If this should be fixed, maybe in another commit -- see #2444?

lfos:
Pushed a new version that addresses your remarks. I also fixed the base class of `AtticSegment` (should be `Error` instead of `Exception`). Thanks for the review!

src/borg/repository.py
ThomasWaldmann: hmm...

i see 2 options:

- catch the AtticSegment exception separately and re-raise it as AtticRepository exception (== "check --repair" is not "upgrade from attic")
- actually change "check --repair" to be the "upgrade from attic" - there is a ticket open about the upgrade only being able to work locally, but if we would do it via "borg check" in the repository code, it could work remotely also.

to not have scope-creep in here, maybe do the first here and we can discuss doing the second later.

@enkore ^

ThomasWaldmann:
added to 1.1 backports. guess we really need this in 1.1, maybe even in 1.0.

in 1.2 it will be somehow pointless as release time of 1.2 is quite a bit in the future - maybe no attic repos left by then?

src/borg/repository.py
lfos: I just realized that we will actually never catch `AtticSegment` here. We already call `get_segments_transaction_id()` much earlier in the `check()` function. So if we decide to catch that error during `check`, it needs to be done much earlier. I tend to remove the extra check for `AtticSegment` here (leave this part of code untouched) and add proper Attic repository handling to `check` instead.

lfos:
Added Attic repository detection to `check` which is now also covered in the test case.

ThomasWaldmann:
Btw, when doing changes, rather add changeset than amending the existing one - easier to review and see what's changed. At the end, after review is finished, changesets can be collapsed to 1 or few.

lfos:
Good point. I am rather used to the ML-based patch submission process (as used for the Linux kernel or Git development itself) where it is more common to resubmit amended patch series. It is kind of annoying that there seems to be no easy way to display inter-diffs in GitHub... In the future, I will submit changes on top and squash only when the whole series is ready to be merged.

src/borg/repository.py
```diff
@@ -321,7 +325,10 @@ def get_index_transaction_id(self):
 
     def check_transaction(self):
         index_transaction_id = self.get_index_transaction_id()
-        segments_transaction_id = self.io.get_segments_transaction_id()
+        try:
+            segments_transaction_id = self.io.get_segments_transaction_id()
+        except LoggedIO.AtticSegment:
+            raise self.AtticRepository(self.path)
```
ThomasWaldmann: more clear tracebacks using this:
```
except ... as e:
    raise ... from e
```

src/borg/repository.py
```diff
@@ -843,6 +850,8 @@ def report_error(msg):
             transaction_id = self.get_transaction_id()
             current_index = self.open_index(transaction_id)
             logger.debug('Read committed index of transaction %d', transaction_id)
+        except self.AtticRepository as exc:
+            raise
```
ThomasWaldmann: same here

src/borg/repository.py
```diff
@@ -1084,6 +1093,9 @@ class LoggedIO:
     class SegmentFull(Exception):
         """raised when a segment is full, before opening next"""
 
+    class AtticSegment(Error):
+        """Attic segment detected [segment {}]"""
```
ThomasWaldmann: 'Attic segment magic detected [segment {}, offset {}]'.format(segment, 0)

src/borg/testsuite/repository.py
```diff
@@ -714,6 +714,14 @@ def test_repair_missing_segment(self):
         self.check(repair=True, status=True)
         self.assert_equal(set([1, 2, 3]), self.list_objects())
 
+    def test_detect_attic_segment(self):
+        self.add_objects([[1, 2, 3]])
+        with open(os.path.join(self.tmppath, 'repository', 'data', '0', '1'), 'r+b') as fd:
+            fd.seek(0)
+            fd.write(ATTIC_MAGIC)
+        self.assert_raises(Repository.AtticRepository, lambda: self.get_objects(4))
+        self.assert_raises(Repository.AtticRepository, lambda: self.check())
```
ThomasWaldmann: looks ok. another way would be to unpack the attic.tgz we have in the testsuite (which contains a repo made by attic).

src/borg/remote.py
```diff
@@ -370,6 +370,8 @@ def inject_exception(self, kind):
             raise Repository.CheckNeeded(s1)
         elif kind == 'IntegrityError':
             raise IntegrityError(s1)
+        elif kind == 'AtticRepository':
+            raise Repository.AtticRepository(s1)
```
lfos: I had the impression that we want to be able to test any exception with this function but that does not seem to be the case. Will remove this hunk.

src/borg/repository.py
```diff
@@ -29,6 +29,7 @@
 logger = create_logger(__name__)
 
 MAGIC = b'BORG_SEG'
+ATTIC_MAGIC = b'ATTICSEG'
 MAGIC_LEN = len(MAGIC)
```
lfos: Looks good, will do.

src/borg/repository.py
lfos: Not sure about this one. The whole purpose of this patch is to get rid of the tracebacks and show a short and meaningful error message when running borg commands on Attic repositories...

src/borg/repository.py
lfos: This is even more unclear to me. What do you want this to look like? Anyway, even if we decide to keep the simple `raise` here, the `as exc` part can go so this hunk will change in a fix-up commit.

src/borg/repository.py
ThomasWaldmann: well, the point was to have nice traceback information.
whether we then show it or not is decided elsewhere.

src/borg/repository.py
ThomasWaldmann: yes, just remove the "as exc". the plain "raise" re-raises the existing exception.

src/borg/repository.py
lfos: Huh? I am pretty sure you cannot do this in a doc string... Do you mean `"""Attic segment detected [segment {}, offset 0]"""` instead?

ThomasWaldmann: ehrm, yes, the .format() happens elsewhere.

ThomasWaldmann: the idea was just that it looks more consistent with the other magic related exception message.

lfos: The Attic MAGIC will always be at offset 0, so I think it is fine to hardcode the value here.

ThomasWaldmann: the other magic related message also specifically mentions offset 0.
this is to be consistent with other error messages always giving segment and offset.

ThomasWaldmann:
btw, another thing i noticed is that the attic magic is also defined in upgrader.py.

lfos:
True. We could import it from there but then we would lose the new assertion. Maybe both magic values should be defined in a central location (where we also have the length assertion)? Seems like this is going a bit out of the scope of this patch... But sounds like a nice idea for a follow-up patch!

ThomasWaldmann:
To avoid cyclic imports we maybe rather move it to repository.py. constants.py might be also an option.
Sure, can be in another patch.

ThomasWaldmann:
Hmm, btw, did you check that this handles all use cases when a user runs borg against a attic repo (no matter which command is used)?

lfos:
Yes, I checked that and no, we do not obtain an error message on every command that is run against an Attic repository. This only affects commands actually reading segments. Running a command like `borg break-lock` against an Attic repository works fine. Should we really error out there as well? If we do, we might as well scrap this patch (keep the `IntegrityError`) and add a check that is run every time before borg accesses a repository... What do you think?

src/borg/testsuite/repository.py
ThomasWaldmann: hmm, maybe copy the more highlevel test from my older try to fix this?

https://github.com/borgbackup/borg/pull/2207/files#diff-a74dadd5d28b08a8516d608b345d6f00R2269

ThomasWaldmann:
@enkore any feedback from you? would like to merge / backport this soon.

ThomasWaldmann:
@lfos no, i don't think we need to invest extra work on special stuff like `break-lock` or `debug ...`.

It's fine if the usual commands fail with a nice error message (e.g. create, extract, list, delete, prune).

src/borg/repository.py
enkore: What would be the point in re-raising? Since it's an `Error` the traceback is not normally shown, so there's little point in attaching a new traceback?

ThomasWaldmann: the point of catching/reraising is so that is not catched by the `except Exception` below.

src/borg/testsuite/repository.py
```diff
@@ -714,6 +714,14 @@ def test_repair_missing_segment(self):
...
+
     def test_repair_missing_commit_segment(self):
```
enkore: (No need to change it, just mentioning it) - the test suite was developed on unittest, but new code is allowed and encouraged to use pytest features like `with pytest.raises(SomeException): ...`

enkore:
LGTM

ThomasWaldmann:
@lfos still working on it or finished?

lfos:
I noticed some issues. Almost done with a new implementation. Will provide more details and an updated patch in a couple of minutes...

lfos:
I won't be able to finish the patch today. Have something that works more or less but I still need to rewrite the tests. The problem is that some operations (such as `borg create`) read segments only after making modifications which results in the Attic error message not being shown for some operations and being shown only after things have been modified for others. The new implementation always tries to read the latest segment (and check its MAGIC) when opening a repository, unless this check is explicitly disabled by an optional parameter (which is currently only used for `borg upgrade`).

lfos:
Pushed the rewrite (decided to rebase since the new commit almost has nothing in common with the original implementation). It also contains a "high-level test case" to make sure the usual commands are covered.

src/borg/testsuite/archiver.py
```diff
@@ -2725,6 +2726,27 @@ def test_extract_hardlinks(self):
             assert os.stat('input/dir1/aaaa').st_nlink == 2
             assert os.stat('input/dir1/source2').st_nlink == 2
 
+    def test_detect_attic_repo(self):
+        path = attic_repo(self.repository_path)
+        cmds = [
+            ['create', path + '::test', self.tmpdir],
+            ['extract', path + '::test', 'output.tar'],
```
ThomasWaldmann: output.tar?

src/borg/testsuite/upgrader.py
```diff
@@ -149,8 +149,8 @@ def test_convert_all(attic_repo, attic_key_file, inplace):
     """
     repo_path = attic_repo
 
-    # check should fail because of magic number
-    assert not repo_valid(repo_path)
+    with pytest.raises(Repository.AtticRepository):
+        repo_valid(repo_path)
```
lfos: Nowhere. It is defined in `src/borg/testsuite/upgrader.py` and does not have any other call sites.

src/borg/testsuite/archiver.py
lfos: Good catch. Should be dropped. I will wait for further comments/reviews before submitting a fix, though...

src/borg/testsuite/upgrader.py
ThomasWaldmann: ok, i've looked it up now.
it is a bit of strange now as it can raise AtticRepository or return True/False for ok / not ok.
but i guess that is ok to deal with this special case, so keep it.

lfos:
Pushed an updated patch. Should be ready to be merged.

enkore:
Thanks!

ThomasWaldmann:
@lfos can you do the backport to 1.1-maint branch?

lfos:
Not sure what to do there, the commit applies fine against `1.1-maint`, so there isn't anything to do...

ThomasWaldmann:
Well, then it is just the PR.

PR #2924, opened 2017-08-07
migrate locks to child PID when daemonize is used

### LINKED ISSUE #2923: lock.roster left behind after umount in 1.1.0 rc1
I have been testing v1.1.0 rc1 and found the following issue:
Before doing a backup I check if a backup is neccessary by mounting the repo and comparing the latest archive with the source.
Then the repo is unmounted.
At the beginning of the next backup I get a message that lock.roster files are removed for PID ...
A workaround is doing a borg break-lock after each borg umount.

ThomasWaldmann:
fixes #2923 

the pid change due to daemonize didn't work well with the stale lock detection (as the initial/parent process was gone). thus we better transfer the locks to the new pid.

this is only needed for client-side (local) repos because the daemonize happens there.
remote repos are locked by borg serve and the pid problem does not happen there.

### DIFF
--- a/src/borg/locking.py
+++ b/src/borg/locking.py
@@ -204,13 +204,13 @@ class ExclusiveLock:
 
     def migrate_lock(self, old_id, new_id):
         """migrate the lock ownership from old_id to new_id"""
-        if self.id == old_id:
-            new_unique_name = os.path.join(self.path, "%s.%d-%x" % new_id)
-            if self.is_locked() and self.by_me():
-                with open(new_unique_name, "wb"):
-                    pass
-                os.unlink(self.unique_name)
-            self.id, self.unique_name = new_id, new_unique_name
+        assert self.id == old_id
+        new_unique_name = os.path.join(self.path, "%s.%d-%x" % new_id)
+        if self.is_locked() and self.by_me():
+            with open(new_unique_name, "wb"):
+                pass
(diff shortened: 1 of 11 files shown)


ThomasWaldmann:
rebased, added tests, collapsed.

ThomasWaldmann:
note: this fix (== having the right PID in the locks) somehow also applies to 1.0, but no issues manifests there as it does not do stale lock removal.

ThomasWaldmann:
@enkore anything else? merge?

src/borg/locking.py
```diff
@@ -202,6 +202,16 @@ def break_lock(self):
                 os.unlink(os.path.join(self.path, name))
             os.rmdir(self.path)
 
+    def migrate_lock(self, old_id, new_id):
+        """migrate the lock ownership from old_id to new_id"""
+        if self.id == old_id:
```
enkore: I don't quite see how this could be false? Perhaps make it an `assert self.id == old_id` instead?

ThomasWaldmann: well, it could be wrong if the caller does something wrong.
yes, could be also an assertion.

ThomasWaldmann:
if there's nothing else, i'll merge this after travis.

ThomasWaldmann:
crap, forgot to collapse. doing the 1.1 backport now.

PR #2925, opened 2017-08-07
Datetime formatting

enkore:
#2376

### DIFF
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -209,9 +209,9 @@ Standard output
 *stdout* is different and more command-dependent than logging. Commands like :ref:`borg_info`, :ref:`borg_create`
 and :ref:`borg_list` implement a ``--json`` option which turns their regular output into a single JSON object.
 
-Dates are formatted according to ISO-8601 in local time. Neither an explicit time zone nor microseconds
-are specified *at this time* (subject to change). The equivalent strftime format string is '%Y-%m-%dT%H:%M:%S',
-e.g. 2017-08-07T12:27:20.
+Dates are formatted according to ISO 8601 in local time. No explicit time zone is specified *at this time*
+(subject to change). The equivalent strftime format string is '%Y-%m-%dT%H:%M:%S.%f',
+e.g. ``2017-08-07T12:27:20.123456``.
 
 The root object at least contains a *repository* key with an object containing:
 
@@ -268,7 +268,7 @@ Example *borg info* output::
         },
         "repository": {
             "id": "0cbe6166b46627fd26b97f8831e2ca97584280a46714ef84d2b668daf8271a23",
-            "last_modified": "2017-08-07T12:27:20",
+            "last_modified": "2017-08-07T12:27:20.789123",
             "location": "/home/user/testrepo"
         },
         "security_dir": "/home/user/.config/borg/security/0cbe6166b46627fd26b97f8831e2ca97584280a46714ef84d2b668daf8271a23",
@@ -329,7 +329,7 @@ Example of a simple archive listing (``borg list --last 1 --json``)::
             {
                 "id": "80cd07219ad725b3c5f665c1dcf119435c4dee1647a560ecac30f8d40221a46a",
                 "name": "host-system-backup-2017-02-27",
-                "start": "2017-08-07T12:27:20"
+                "start": "2017-08-07T12:27:20.789123"
             }
         ],
         "encryption": {
@@ -337,7 +337,7 @@ Example of a simple archive listing (``borg list --last 1 --json``)::
         },
         "repository": {
             "id": "0cbe6166b46627fd26b97f8831e2ca97584280a46714ef84d2b668daf8271a23",
-            "last_modified": "2017-08-07T12:27:20",
+            "last_modified": "2017-08-07T12:27:20.789123",
             "location": "/home/user/repository"
         }
     }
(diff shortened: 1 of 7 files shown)


ThomasWaldmann:
later, first merge helpers split.

ThomasWaldmann:
guess I made a bit of a mess here by force pushing to master, removing the merge of the json timestamp changes, sorry. it caused a big, unclear conflict, so i rather wanted to first have the helpers split in master and then rebase the json timestamp stuff on that (rather than the other way).

src/borg/helpers.py
```diff
@@ -756,6 +756,22 @@ def __format__(self, format_spec):
...
+        self.ts = ts
+
+    def __format__(self, format_spec):
+        return format_time(self.ts)
+
+    def __str__(self):
+        return '{}'.format(self)
+
+    def to_json(self):
+        return isoformat_time(self.ts)
```
ThomasWaldmann: somehow slightly misleading name. you use it for json output, but it is not json by itself.
in fact, the only difference is that it is rather optimized to be machine readable (and really iso 8601) and the other format is more human readable and only parts of it are formatted like in iso.

src/borg/helpers.py
```diff
@@ -1879,15 +1888,10 @@ def hash_item(self, hash_function, item):
...
-
-    def format_time_json(self, key, item):
-        t = self.time(key, item)
-        return isoformat_time(t)
+        return OutputTimestamp(safe_timestamp(item.get(key) or item.mtime))
 
-    def time(self, key, item):
-        return safe_timestamp(item.get(key) or item.mtime)
+    def format_iso_time(self, key, item):
+        return self.format_time(key, item).to_json()
```
ThomasWaldmann: strange: format_time is returning a formatter instance while format_iso_time returns a string.

docs/internals/frontends.rst
```diff
@@ -209,8 +209,10 @@ Standard output
 *stdout* is different and more command-dependent than logging. Commands like :ref:`borg_info`, :ref:`borg_create`
 and :ref:`borg_list` implement a ``--json`` option which turns their regular output into a single JSON object.
 
-Dates are formatted according to ISO-8601 with the strftime format string '%a, %Y-%m-%d %H:%M:%S',
-e.g. *Sat, 2016-02-25 23:50:06*.
+Dates are formatted according to ISO-8601 in local time. Neither an explicit time zone nor microseconds
+are specified *at this time* (subject to change). The equivalent strftime format string is '%Y-%m-%dT%H:%M:%S',
+e.g. 2017-08-07T12:27:20. Standard formatting is used for microseconds, if available. Only :ref:`borg_list`
```
ThomasWaldmann: "Standard formatting is used for microseconds, if available."

That reads a bit strange, after saying "Neither an explicit time zone nor microseconds are specified *at this time* (subject to change)."

So, what shall it mean?

docs/internals/frontends.rst
```diff
@@ -405,8 +407,8 @@ Refer to the *borg list* documentation for the available keys and their meaning.
 
 Example (excerpt) of ``borg list --json-lines``::
 
-    {"type": "d", "mode": "drwxr-xr-x", "user": "user", "group": "user", "uid": 1000, "gid": 1000, "path": "linux", "healthy": true, "source": "", "linktarget": "", "flags": null, "isomtime": "Sat, 2016-05-07 19:46:01", "size": 0}
-    {"type": "d", "mode": "drwxr-xr-x", "user": "user", "group": "user", "uid": 1000, "gid": 1000, "path": "linux/baz", "healthy": true, "source": "", "linktarget": "", "flags": null, "isomtime": "Sat, 2016-05-07 19:46:01", "size": 0}
+    {"type": "d", "mode": "drwxr-xr-x", "user": "user", "group": "user", "uid": 1000, "gid": 1000, "path": "linux", "healthy": true, "source": "", "linktarget": "", "flags": null, "mtime": "2017-02-27T12:27:20.023407", "size": 0}
+    {"type": "d", "mode": "drwxr-xr-x", "user": "user", "group": "user", "uid": 1000, "gid": 1000, "path": "linux/baz", "healthy": true, "source": "", "linktarget": "", "flags": null, "mtime": "2017-02-27T12:27:20.585407", "size": 0}
```
ThomasWaldmann: here there are usecs vs you documented "no usecs at this time" above.

src/borg/helpers.py
```diff
@@ -717,11 +717,18 @@ def safe_timestamp(item_timestamp_ns):
...
+    """use ISO-8601-like date and time format (human readable, with wkday and blank date/time separator)
     """
     return t.strftime('%a, %Y-%m-%d %H:%M:%S')
 
 
+def isoformat_time(t, microseconds=False):
+    """use ISO-8601 date and time format (machine readable, no wkday, no microseconds either)
+    """
+    # note: first make all datetime objects tz aware before adding %z here.
+    return t.strftime('%Y-%m-%dT%H:%M:%S' + ('.%f' if microseconds else ''))
```
ThomasWaldmann: docstring

src/borg/helpers.py
```diff
@@ -1863,10 +1890,10 @@ def hash_item(self, hash_function, item):
         return hash.hexdigest()
 
     def format_time(self, key, item):
-        return format_time(safe_timestamp(item.get(key) or item.mtime))
+        return OutputTimestamp(safe_timestamp(item.get(key) or item.mtime), json_microseconds=True)
```
ThomasWaldmann: so, if we want usecs here in json, maybe simpler to just always/everywhere use usecs for json?

ThomasWaldmann:
@enkore can you finish this or shall I help (rebase, fix conflicts, ...)?

Guess it would make sense to have this in rc2 (so we won't need another rc).

enkore:
Free time is available starting on the 16th Aug.

enkore:
Merge?

docs/internals/frontends.rst
ThomasWaldmann: hmm, isn't that confusing now in other ways?
key is "mtime", but it does not correspond to "mtime" format but rather to "isomtime" format?

src/borg/testsuite/archiver.py
```diff
@@ -1862,7 +1864,7 @@ def test_list_json(self):
         file1 = items[1]
         assert file1['path'] == 'input/file1'
         assert file1['size'] == 81920
-        assert datetime.strptime(file1['isomtime'], ISO_FORMAT)  # must not raise
+        assert datetime.strptime(file1['mtime'], ISO_FORMAT)  # must not raise
```
ThomasWaldmann: same here: key -> no iso, but parsing iso format.

docs/internals/frontends.rst
enkore: I don't think it's very confusing, because the docs say dates are formatted according to ISO 8601 in JSON mode, so that should be the default. I see your point regarding the minor inconsistency between isoXY and XY, but I don't think its a big problem.

PR #2927, opened 2017-08-07
Revert "json: use isoformat"

ThomasWaldmann:
Reverts borgbackup/borg#2916


ThomasWaldmann:
not needed, just reset master to previous cs.

PR #2930, opened 2017-08-13
adjust logo colour to a more pleasant green

enkore:
Based on user  feedback, ultra-pure intense green is not a pleasant colour. I picked #44d02e as a somewhat brighter relative of RAL 6038 ("Leuchtgrün"), but open to suggestions.

Also: Document-by-action how the PNG and PDFs are created, some minor CSS cleanup.
Because the SVG is now used as the master, the new PNG is not 100 % the same as the old version (created in GIMP, per git history).


ThomasWaldmann:
maybe it depends on the screen and brightness settings (or eyes or ...), but the previous one looks better for me.

enkore:
FWIW I don't like the intense/pure green from before as well (kinda stingy), but I'm not great with colors [1]. So I think it should be changed, but I'm just proposing a rough direction and completely open to alternate suggestions.

[1] So colorschemedesigner will have to make do to generate accent colours for the website from the new primary.

ThomasWaldmann:
or just leave colour as it was before for now and just commit the other changes?

PR #2932, opened 2017-08-13
migrate locks to child PID when daemonize is used

ThomasWaldmann:
(cherry picked from commit 09e3a02fbc5edd64125ac437cb3ec6fb20ce730a)

### DIFF
--- a/borg/locking.py
+++ b/borg/locking.py
@@ -8,17 +8,20 @@ from borg.helpers import Error, ErrorWithTraceback
 ADD, REMOVE = 'add', 'remove'
 SHARED, EXCLUSIVE = 'shared', 'exclusive'
 
-# only determine the PID and hostname once.
-# for FUSE mounts, we fork a child process that needs to release
-# the lock made by the parent, so it needs to use the same PID for that.
-_pid = os.getpid()
+# for performance reasons, only determine the hostname once.
 _hostname = socket.gethostname()
 
 
 def get_id():
-    """Get identification tuple for 'us'"""
+    """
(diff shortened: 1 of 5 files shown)


borg/locking.py
```diff
@@ -8,17 +8,22 @@
...
 SHARED, EXCLUSIVE = 'shared', 'exclusive'
 
-# only determine the PID and hostname once.
-# for FUSE mounts, we fork a child process that needs to release
-# the lock made by the parent, so it needs to use the same PID for that.
-_pid = os.getpid()
+# for performance reasons, only determine the hostname once.
+## XXX this sometimes requires live internet access for issuing a DNS query in the background.
+##_hostname = '%s@%s' % (socket.getfqdn(), uuid.getnode())
 _hostname = socket.gethostname()
```
ThomasWaldmann: 12: 1.1 comment
13: 1.1 way
14: 1.0 way

so, keep the 1.0 way or change it to 1.1 behaviour? we do not have stale detection in 1.0.

enkore: gethostname should only look in uname(2) or equivalent, so it should never require Internet access. Keep it as is for 1.0.

(Which reminds me that there's a sentence in the 1.1 docs saying something like "Borg doesn't do networking" which could be considered incorrect in certain setups due to the getfqdn() call involving a resolver).

ThomasWaldmann: ok, i keep it as is.

networking: well, it's rather the stdlib doing it then, so it is indirect.

PR #2944, opened 2017-08-18
development.rst: Miscellaneous improvements

lfos:
* Fix a couple of grammar mistakes and improve wording.
* Use proper Git terminology for operations on patches.
* Unify capitalization of "GitHub".
* Add proper formatting to branch names.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -17,7 +17,7 @@ Contributions
 
 Some guidance for contributors:
 
-- discuss about changes on github issue tracker, IRC or mailing list
+- discuss changes on the GitHub issue tracker, IRC or on the mailing list
 
 - make your PRs on the ``master`` branch (see `Branching Model`_ for details)
 
@@ -29,9 +29,9 @@ Some guidance for contributors:
   - if you need to fix something after commit/push:
 
     - if there are ongoing reviews: do a fixup commit you can
-      merge into the bad commit later.
+      squash into the bad commit later.
     - if there are no ongoing reviews or you did not push the
-      bad commit yet: edit the commit to include your fix or
+      bad commit yet: amend the commit to include your fix or
       merge the fixup commit before pushing.
   - have a nice, clear, typo-free commit comment
   - if you fixed an issue, refer to it in your commit comment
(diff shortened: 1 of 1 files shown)


docs/development.rst
```diff
@@ -17,7 +17,7 @@ Contributions
 
 Some guidance for contributors:
 
-- discuss about changes on github issue tracker, IRC or mailing list
+- discuss changes on the GitHub issue tracker, IRC or on the mailing list
```
ThomasWaldmann: "mailing list" - "on the" is repeated for that, but not for "IRC".

so, guess we do not need to repeat it for "mailing list" either.

docs/development.rst
```diff
@@ -29,19 +29,19 @@ Some guidance for contributors:
...
+      bad commit yet: amend the commit to include your fix or
       merge the fixup commit before pushing.
   - have a nice, clear, typo-free commit comment
   - if you fixed an issue, refer to it in your commit comment
   - follow the style guide (see below)
 
 - if you write new code, please add tests and docs for it
 
-- run the tests, fix anything that comes up
+- run the tests, fix any errors that come up
```
ThomasWaldmann: no. really anything that comes up was meant.

test errors, test failures, pep8 style issues, stuff that you will notice by having a second look, ...

docs/development.rst
lfos: I would rather also say "on IRC" here... But I don't really feel strongly about it :)

docs/development.rst
lfos: Well, "anything" sounds quite strange here. There are "things" (such as the summary) coming up in any case, even if there are no errors. Maybe "fix any issues" instead? I know that this is just nitpicking. But then again, the whole patch is just minor tweaks...

ThomasWaldmann: yup, change it to "fix any issues that come up".

docs/development.rst
ThomasWaldmann: ok, so "on the GitHub issuetracker, on IRC or on the mailing list".

ThomasWaldmann:
thanks!

PR #2951, opened 2017-08-20
Buffer: fix wrong thread-local storage use

enkore:
Instead of backporting, just remove threading-related things from older branches.

### DIFF
--- a/src/borg/helpers/datastruct.py
+++ b/src/borg/helpers/datastruct.py
@@ -10,8 +10,10 @@ class StableDict(dict):
         return sorted(super().items())
 
 
-class Buffer:
-    """provide a thread-local buffer"""
+class Buffer(threading.local):
+    """
+    provide a thread-local buffer
+    """
 
     class MemoryLimitExceeded(Error, OSError):
         """Requested buffer size {} is above the limit of {}."""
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
can you file a bug for this please with more details?

enkore:
Well its not a bug unless multiple threads try to use the same buffer instance (eventually the MT-support of Buffer should be removed, since evidently it implies that some object is used concurrently by multiple threads, something that shall be avoided).

Anyhow, the issue with thread_local is that it's not magic. If you assign an attribute to it, then only the current thread assigns that attribute to its thread-local backing dict. Conversely, that attribute does not exist in other threads. Therefore, if a Buffer is initialized in one thread, it has a  thread_local with a buffer attribute in exactly that thread. If another thread uses that Buffer, then the object is effectively not in the state expected, because the buffer attribute is missing.

The fix is to derive from thread_local instead, which has a number of implications (none of this is documented in the Python docs, you need to read the source docstrings in the cpython source dist for this), including a complete change of `__init__` semantics. A thread_local's `__init__` is  called from every Python thread that (a) already exists or (b) is created during the lifetime of an instance. This way, each thread has a consistently initialized Buffer.

Now its I think straightforward to go from this point of complexity to the conclusion that Buffer should cease to support usage from multiple threads in the future entirely — perhaps before we even add any threading to Borg in the first place.

enkore:
That being said, I'm just going to remove that, so as to never use this sort of hack. Thread-locals in Python are a neat foot gun, as outlined above.

ThomasWaldmann:
"Instead of backporting, just remove threading-related things from older branches."

ok, so this meanwhile changed to "just backport as is"?

ThomasWaldmann:
strange, no threadlocal stuff in 1.1, but in 1.0.

PR #2952, opened 2017-08-20
delete various nogil and threading related lines

enkore:


enkore:
Yes, but they're just declarations, while `with nogil` actually does something.

PR #2953, opened 2017-08-20
create: document exclusion through nodump

### LINKED ISSUE #2949: OS X: unexpected "excluded" directories
On IRC today:

```
09:46  qwebirc80461$
           hi. I have an issue using borg 1.1.0rc1 when trying to backup
           some macOS folders like Documents or Desktop.
           I am running the create command like this:
           borg create -v --list test::d Documents
           It backs up the files in the directory, but every subfolder is excluded
           (--list prepends them with an 'x').
           I am not specifying exclude patterns.
          Why are the folders excluded?
10:27  thw[m]$ is there anything special about these folders?
10:30  qwebirc80461$
           not that i know of.
           they all have the same permissions as the parent folders.
           also if i move one of the 'excluded' subfolders into 
           another directory, eg. foo. The subfolder gets backed up correctly. oO
10:31  thw[m]$
           if you just added a new normal folder with some normal files,
           does it back up that one?
10:32  qwebirc80461$ yes. that works
10:33  thw[m]$ And if you then do another backup, does it still work?
10:34  qwebirc80461$ yes. but the files show always as 'A' (via --list)
10:35  thw[m]$
           And the stuff listed with x - is it really not in the archive
           if you do a borg list?
10:35  qwebirc80461$ yeap. not there
10:36  thw[m]$ File system type?
10:36  qwebirc80461$ HFS+
10:36  qwebirc80461$ Journaled. Not case sensitive
10:36  thw[m]$ Ok, somehow that sounds like a bug.
10:37  thw[m]$
          Can you file a github issue including how to reproduce
          and  your terminal logs?
10:38  qwebirc80461$ ok, thx
10:38  thw[m]$ Are there nonascii chars in the folder names?
10:38  qwebirc80461$ No
10:40  thw[m]$ I'll try to reproduce asap but maybe not today.
10:40  qwebirc80461$ I am using Filevault but I don't think that makes a difference?
10:42  qwebirc80461$ == Full Disk encryption
10:43  thw[m]$ Well i'ld expect that to be fully transparent for apps?
10:46  thw[m]$ Mention it in the bug report. Os version. Fs. Encr.
                        How borg was installed. Borg version.  Python version.
10:46  qwebirc80461$ ok
11:21  qwebirc80461$
           The subfolders which were ignored were copied using an
           SMB network connection to my computer. I just deleted
            them all and copied them to my disk using via external
            usb drive. Now the backup works as expected. Though the
            subfolders have the same permissions as before... I wonder 
            what made the difference.
11:23  thw[m]$ not sure that is the cause
12:53  qwebirc80461$
           I can reproduce it. Create a file share on macOS.
           It can then be mounted via smb or afp protocol.
           If folders are copied using smb they are excluded in borg.
           Copied via afp they are included. How strange.
13:06  ThomasWaldmann$
           would be great if you coud try to find the difference,
           maybe using a terminal session as a normal user
           dot use finder for that, it seems to have special permissions
13:30  qwebirc80461$
           I copied (using finder) one example folder with smb and afp.
           diff -rq shows no difference. There are also no extended attributes 
           set on the folders. ls -lR also shows no difference. Folder copied
           with afp works. With smb it is ignored
13:49  ThomasWaldmann$ ACLs maybe?
13:58  qwebirc80461$
            ls -leR shows no difference
            (and no ACLs; the -e displays the ACL information)
```

enkore:
#2949

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2815,6 +2815,9 @@ class Archiver:
         and not include any other contents of the containing folder, this can be enabled
         through using the ``--keep-exclude-tags`` option.
 
+        Borg respects the nodump flag. Files flagged nodump will be marked as excluded (x)
+        in ``--list`` output.
+
         Item flags
         ++++++++++
 


ThomasWaldmann:
Well, this rather fixes our documentation obligation about nodump flag usage by borg than #2949.

OS X users would be still unaware of this strange behaviour and run into it.
But maybe we can only do more once the root cause / reason of this is known.

PR #2996, opened 2017-09-03
Improve macOS faq entry

mgee:
The macOS keychain integration guide is missing a `-w` in the `BORG_PASSCOMMAND` environment variable. This flag causes only the password to be written to stdout. Without this flag the integration does not work.


enkore:
Thanks!

PR #2998, opened 2017-09-05
fix .isoformat() issues

### LINKED ISSUE #2994: timestamp microseconds parsing issue
Spurious failure seen in a unrelated 1.1-maint PR on py 3.4:
```
.tox/py34/lib/python3.4/site-packages/borg/archiver.py:1232: in _delete_archives
    manifest.write()
.tox/py34/lib/python3.4/site-packages/borg/helpers.py:393: in write
    prev_ts = self.last_timestamp
.tox/py34/lib/python3.4/site-packages/borg/helpers.py:321: in last_timestamp
    return datetime.strptime(self.timestamp, "%Y-%m-%dT%H:%M:%S.%f")

E           ValueError: time data '2017-09-02T16:16:34' does not match format '%Y-%m-%dT%H:%M:%S.%f'
```

Suspicion: crappy standard formatting behaviour of stdlib to only add `.uuu` for the microsecs if microsecs != 0, which then causes complications when trying to parse that.

ThomasWaldmann:
fixes #2994

### DIFF
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -66,6 +66,12 @@ EXIT_SUCCESS = 0  # everything done, no problems
 EXIT_WARNING = 1  # reached normal end of operation, but there were issues
 EXIT_ERROR = 2  # terminated abruptly, did not reach end of operation
 
+# never use datetime.isoformat(), it is evil. always use one of these:
+# datetime.strftime(ISO_FORMAT)  # output always includes .microseconds
+# datetime.strftime(ISO_FORMAT_NO_USECS)  # output never includes microseconds
+ISO_FORMAT_NO_USECS = '%Y-%m-%dT%H:%M:%S'
+ISO_FORMAT = ISO_FORMAT_NO_USECS + '.%f'
+
 DASHES = '-' * 78
 
 PBKDF2_ITERATIONS = 100000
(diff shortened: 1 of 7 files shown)


src/borg/constants.py
```diff
@@ -66,6 +66,12 @@
 EXIT_WARNING = 1  # reached normal end of operation, but there were issues
 EXIT_ERROR = 2  # terminated abruptly, did not reach end of operation
 
+# never use datetime.isoformat(), it is evil. always use one of these:
```
enkore: Is there a linter that can check stuff like this automatically?

src/borg/repository.py
```diff
@@ -540,7 +540,8 @@ def rename_tmp(file):
         # Log transaction in append-only mode
         if self.append_only:
             with open(os.path.join(self.path, 'transactions'), 'a') as log:
-                print('transaction %d, UTC time %s' % (transaction_id, datetime.utcnow().isoformat()), file=log)
+                print('transaction %d, UTC time %s' % (
+                    transaction_id, datetime.utcnow().strftime(ISO_FORMAT)), file=log)
```
enkore: Indentation is off-by-one

src/borg/constants.py
ThomasWaldmann: i don't know.

src/borg/repository.py
ThomasWaldmann: it's indented by 4, but yes, prettier would be to indent to in-parens level.

PR #3006, opened 2017-09-07
with-lock: close segment file before invoking subprocess (1.1-maint)

ThomasWaldmann:


ThomasWaldmann:
no backport to 1.0 needed, no with-lock in 1.0.

PR #3024, opened 2017-09-11
implement files cache mode control, fixes #911

### LINKED ISSUE #911: Files cache: use ctime instead of mtime
ctime can't be set via userspace, while mtime is easily manipulated.

For most files (e.g. not extracted from tarballs, or installed by a package manager) these two are the same, which limits the impact of the change on existing file caches.
- Works with sshfs? -> (YES) (see below)

@verygreen noted that in the Windows world ctime usually means "creation" not "change" time
- Works with windows+cygwin? -> YES
- Works with native windows? -> NO

ThomasWaldmann:
You can now control the files cache mode using this option:

--files-cache={ctime,mtime,size,inode,rechunk,disabled}*

(only some combinations are supported)

Previously, only these modes were supported:
- mtime,size,inode (default of borg < 1.1.0rc4)
- mtime,size (by using --ignore-inode)
- disabled (by using --no-files-cache)

Now, you additionally get:
- ctime alternatively to mtime (more safe), e.g.:
  ctime,size,inode (this is the new default of borg >= 1.1.0rc4)
- rechunk (consider all files as changed, rechunk them)

Deprecated:
- --ignore-inodes (use modes without "inode")
- --no-files-cache (use "disabled" mode)

The tests needed some changes:
- previously, we use os.utime() to set a files mtime (atime) to specific
  values, but that does not work for ctime.
- now use time.sleep() to create the "latest file" that usually does
  not end up in the files cache (see FAQ)

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -738,7 +738,7 @@ b) with ``create --chunker-params 19,23,21,4095`` (default):
 
   mem_usage  =  0.31GiB
 
-.. note:: There is also the ``--no-files-cache`` option to switch off the files cache.
+.. note:: There is also the ``--files-cache=disabled`` option to disable the files cache.
    You'll save some memory, but it will need to read / chunk all the files as
    it can not skip unmodified files then.
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -45,7 +45,7 @@ from .crypto.keymanager import KeyManager
 from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
 from .helpers import Error, NoManifestError, set_ec
 from .helpers import positive_int_validator, location_validator, archivename_validator, ChunkerParams
-from .helpers import PrefixSpec, SortBySpec
+from .helpers import PrefixSpec, SortBySpec, FilesCacheMode
 from .helpers import BaseFormatter, ItemFormatter, ArchiveFormatter
 from .helpers import format_timedelta, format_file_size, parse_file_size, format_archive
 from .helpers import safe_encode, remove_surrogates, bin_to_hex, prepare_dump_dict
@@ -346,7 +346,7 @@ class Archiver:
             compression = '--compression=none'
             # measure create perf (without files cache to always have it chunking)
             t_start = time.monotonic()
-            rc = self.do_create(self.parse_args(['create', compression, '--no-files-cache', archive + '1', path]))
+            rc = self.do_create(self.parse_args(['create', compression, '--files-cache=disabled', archive + '1', path]))
             t_end = time.monotonic()
             dt_create = t_end - t_start
             assert rc == 0
@@ -485,6 +485,7 @@ class Archiver:
         self.output_filter = args.output_filter
         self.output_list = args.output_list
         self.ignore_inode = args.ignore_inode
+        self.files_cache_mode = args.files_cache_mode
         dry_run = args.dry_run
         t0 = datetime.utcnow()
         t0_monotonic = time.monotonic()
@@ -548,7 +549,7 @@ class Archiver:
                     return
             if stat.S_ISREG(st.st_mode):
                 if not dry_run:
-                    status = fso.process_file(path, st, cache, self.ignore_inode)
+                    status = fso.process_file(path, st, cache, self.ignore_inode, self.files_cache_mode)
             elif stat.S_ISDIR(st.st_mode):
                 if recurse:
                     tag_paths = dir_is_tagged(path, exclude_caches, exclude_if_present)
@@ -1960,14 +1961,17 @@ class Archiver:
 
     def preprocess_args(self, args):
         deprecations = [
-            # ('--old', '--new', 'Warning: "--old" has been deprecated. Use "--new" instead.'),
+            # ('--old', '--new' or None, 'Warning: "--old" has been deprecated. Use "--new" instead.'),
             ('--list-format', '--format', 'Warning: "--list-format" has been deprecated. Use "--format" instead.'),
             ('--keep-tag-files', '--keep-exclude-tags', 'Warning: "--keep-tag-files" has been deprecated. Use "--keep-exclude-tags" instead.'),
+            ('--ignore-inode', None, 'Warning: "--ignore-inode" has been deprecated. Use "--files-cache=ctime,size" or "...=mtime,size" instead.'),
+            ('--no-files-cache', None, 'Warning: "--no-files-cache" has been deprecated. Use "--files-cache=disabled" instead.'),
         ]
         for i, arg in enumerate(args[:]):
             for old_name, new_name, warning in deprecations:
                 if arg.startswith(old_name):
-                    args[i] = arg.replace(old_name, new_name)
+                    if new_name is not None:
+                        args[i] = arg.replace(old_name, new_name)
                     print(warning, file=sys.stderr)
         return args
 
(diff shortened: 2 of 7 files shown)


src/borg/archiver.py
```diff
@@ -2594,6 +2595,8 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         ``--ignore-inode`` flag can be used to prevent this and improve performance.
         This flag will reduce reliability of change detection however, with files
         considered unmodified as long as their size and modification time are unchanged.
+        By default, file ctime is used for the "file not modified" detection, but the
+        ``--use-mtime`` flag can be used to use file mtime instead.
```
ThomasWaldmann: hmm, s/used/considered/ for the first and last instance.

ThomasWaldmann:
any feedback?

enkore:
I don't like extra options for such obscure details. I think it would be better to conclude either possibility and have no new option.

ThomasWaldmann:
I don't like them either, but there is no clear choice here.

A lot of tools just use mtime and mtime is faster.

ctime is safer because userspace software can't do weird stuff with it, but slower if someone does a chmod/chown/ln over bigger amounts of data.

So the rationale of this PR is "do it safe by default, leave option to do it quick (if admin can assert no strange stuff is going on with mtime)".

src/borg/archiver.py
```diff
@@ -2699,6 +2702,8 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                               help='do not store ctime into archive')
         fs_group.add_argument('--ignore-inode', dest='ignore_inode', action='store_true',
                               help='ignore inode data in the file metadata cache used to detect unchanged files.')
+        fs_group.add_argument('--use-mtime', dest='use_mtime', action='store_true',
```
ThomasWaldmann: use `--cache-mtime` and args.cache_mtime to make it more clear that this is cache related?

we also have `--noctime` and `--noatime`, but they are completely unrelated.

src/borg/cache.py
```diff
@@ -911,7 +912,8 @@ def file_known_and_unchanged(self, path_hash, st, ignore_inode=False):
         if not entry:
             return None
         entry = FileCacheEntry(*msgpack.unpackb(entry))
-        if (entry.size == st.st_size and bigint_to_int(entry.mtime) == st.st_mtime_ns and
+        cmtime = st.st_mtime_ns if use_mtime else st.st_ctime_ns
+        if (entry.size == st.st_size and bigint_to_int(entry.cmtime) == cmtime and
```
ThomasWaldmann: interesting detail:

as we use `==`, it does not matter whether the fs timestamp is newer or older than the cached timestamp.

guess that means switching from mtime to ctime or vice versa does not create any harmful effect (besides hashing some stuff again), so we do not need `borg delete --cache-only` when switching.

ThomasWaldmann:
about using `max(mtime, ctime)` instead of just `ctime`:

good:
- if a fs would return a completely broken ctime (like always 0), our cache would use mtime and still work

bad:
- more complex behaviour

ThomasWaldmann:
I'm going to merge this soon as I don't see any better alternative.

Whether using max(mtime, ctime) is any good / is really needed is still an open question.

Guess if we supported native windows, using max() would be good as there ctime is creation time (not change time), thus not helpful for what we need.

enkore:
I think max(mtime, ctime) is kinda like the worst combination of both. It still re-chunks on metadata-only changes, but doesn't have the idiot-safety of ctime.

ThomasWaldmann:
Hm? It would keep the safety of ctime IF ctime is supported by the OS / fs in a UNIX-like way.

The problem with mtime is that it can be faked to stay same, even though the file was changed.
ctime would change to a bigger numerical value and can't be faked, so max(ctime, mtime) would result in the ctime for such a case.

The only cases when max(ctime, mtime) result in mtime is when ctime does not work at all (like always 0, not sure if that can be seen out there) or when ctime means creation time and thus is always smaller than mtime.

enkore:
Ah right.

enkore:
Right, but I could e.g. change the mtime to some very large value (mtime > ctime), modify the file, re-set mtime to the large value etc.

ThomasWaldmann:
Yup, that would escape the change detection if there was a backup after the first far-future mtime change, but before the file modification. Crap.

ThomasWaldmann:
ctime << 64 + mtime?

src/borg/archiver.py
```diff
@@ -2699,6 +2703,8 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                               help='do not store ctime into archive')
         fs_group.add_argument('--ignore-inode', dest='ignore_inode', action='store_true',
                               help='ignore inode data in the file metadata cache used to detect unchanged files.')
+        fs_group.add_argument('--cache-mtime', dest='cache_mtime', action='store_true',
+                              help='use mtime instead of ctime in the file metadata cache to detect unchanged files.')
```
ThomasWaldmann: hmm, in case there are other ideas in future, shall this rather work like:

--cache-timestamp='c' (ctime, default), 'm' (mtime)

or with long names: 'ctime', 'mtime'

enkore: Perhaps use an environment variable as is done for other rarely-if-ever modified settings.

ThomasWaldmann:
Note: i kept it as a cmdline option rather than env var.

We have --ignore-inode and --no-files-cache now, but they can be superceded by just using the respective --cache-mode options. So, 1 option less long term than before this changeset.

For local UNIX file systems and when valuing safety over speed, no option must be used.
If ctime or inode numbers are not working as expected, options can be used to work around that.

Using env vars for this feels somehow more clunky, esp. when the source filesystems need different cache mode values.

ThomasWaldmann:
src/borg/archiver.py:2614:1: W293 blank line contains whitespace

https://travis-ci.org/borgbackup/borg/jobs/280267622

There is no whitespace.

enkore:
Ok. I think it would be better to use less cryptic option names, consider `--cache-mode=ignore-inode,mtime`.

src/borg/constants.py
```diff
@@ -60,6 +60,10 @@
 # chunker params for the items metadata stream, finer granularity
 ITEMS_CHUNKER_PARAMS = (15, 19, 17, HASH_WINDOW_SIZE)
 
+# operating mode of the files cache (for fast skipping of unchanged files)
+DEFAULT_CACHE_MODE_UI = 'ctime,size,inode'
+DEFAULT_CACHE_MODE = 'cis'  # == CacheMode(DEFAULT_CACHE_MODE_UI)
```
ThomasWaldmann: didn't want to import CacheMode from helpers into constants to avoid cyclic imports.

that these 2 actually correspond is asserted elsewhere.

src/borg/helpers/parseformat.py
```diff
@@ -117,6 +117,22 @@ def ChunkerParams(s):
     return int(chunk_min), int(chunk_max), int(chunk_mask), int(window_size)
 
 
+def CacheMode(s):
+    ENTRIES_MAP = dict(ctime='c', mtime='m', size='s', inode='i', rechunk='r', disabled='d')
+    VALID_MODES = ('cis', 'ims', 'cs', 'ms', 'cr', 'mr', 'd')  # letters in alpha order
```
ThomasWaldmann: wanted to keep the short representation for performance reasons.

docs/internals/data-structures.rst
```diff
@@ -738,7 +738,7 @@ b) with ``create --chunker-params 19,23,21,4095`` (default):
 
   mem_usage  =  0.31GiB
 
-.. note:: There is also the ``--no-files-cache`` option to switch off the files cache.
+.. note:: There is also the ``--cache-mode=disabled`` option to disable the files cache.
```
ThomasWaldmann: hmm, use `--files-cache` for this option?

is more specific about which cache is meant.

ThomasWaldmann:
@enkore can you do a final review before I merge this?

Issue #2637, opened 2017-06-08
truncated json


ThomasWaldmann:
```
_______________ RemoteArchiverTestCase.test_corrupted_repository _______________
[gw0] cygwin -- Python 3.6.1 /home/admin/w/borg/.tox/py36/bin/python3.6
self = <borg.testsuite.archiver.RemoteArchiverTestCase testMethod=test_corrupted_repository>

    def test_corrupted_repository(self):
        self.cmd('init', '--encryption=repokey', self.repository_location)
        self.create_src_archive('test')
        self.cmd('extract', '--dry-run', self.repository_location + '::test')
        output = self.cmd('check', '--show-version', self.repository_location)
        self.assert_in('borgbackup version', output)  # implied output even without --info given
        self.assert_not_in('Starting repository check', output)  # --info not given for root logger

        name = sorted(os.listdir(os.path.join(self.tmpdir, 'repository', 'data', '0')), reverse=True)[1]
        with open(os.path.join(self.tmpdir, 'repository', 'data', '0', name), 'r+b') as fd:
            fd.seek(100)
            fd.write(b'XXXX')
>       output = self.cmd('check', '--info', self.repository_location, exit_code=1)

.tox/py36/lib/python3.6/site-packages/borg/testsuite/archiver.py:1392:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.tox/py36/lib/python3.6/site-packages/borg/testsuite/archiver.py:280: in cmd
    ret, output = exec_cmd(*args, fork=fork, exe=self.EXE, archiver=self.archiver, **kw)
.tox/py36/lib/python3.6/site-packages/borg/testsuite/archiver.py:99: in exec_cmd
    ret = archiver.run(args)
.tox/py36/lib/python3.6/site-packages/borg/archiver.py:3951: in run
    return set_ec(func(args))
.tox/py36/lib/python3.6/site-packages/borg/archiver.py:131: in wrapper
    return method(self, args, repository=repository, **kwargs)
.tox/py36/lib/python3.6/site-packages/borg/archiver.py:270: in do_check
    if not repository.check(repair=args.repair, save_space=args.save_space):
.tox/py36/lib/python3.6/site-packages/borg/remote.py:480: in do_rpc
    return self.call(f.__name__, named, **extra)
[...]

PR #3032, opened 2017-09-18
flush json mode progress stderr output

ThomasWaldmann:
if borg stderr is not connected to a tty, but to ssh (when using
borg client/server), sys.stderr is block buffered (tty: line buffered).

thus we better flush explicitly after emitting a line as the receiving
side can not handle partial json at the end of the block.

also, it might solve some delays, when output didn't arrive at
receiving side in time.

hopefully a partial fix for #2637 (progress indication in json mode)


ThomasWaldmann:
I'll merge and backport this so it can be tested more easily.

PR #3042, opened 2017-09-21
fuse: Separate creation of filesystem from implementation of llfuse funcs

imsodin:
Still on the way to #2475. This again does not alter any functionality, it just does what @enkore proposed in a comment on the linked issue:

> (Edit: In fact, it seems like a fairly good idea in general to separate the generation of items from the actual FuseOps implementation).

This is a straight separation without changing the concepts. Probably when actually working with it to create the diff mounting functionality, there will be reason for "real" change.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -202,27 +202,21 @@ class ItemCache:
         self.write_offset = write_offset
 
 
-class FuseOperations(llfuse.Operations):
-    """Export archive as a FUSE filesystem
+class FuseBackend(object):
+    """Virtual filesystem based on archive(s) to provide information to fuse
     """
-    # mount options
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
```
src/borg/fuse.py:257:55: F821 undefined name 'old_id'
src/borg/fuse.py:257:63: F821 undefined name 'new_id'
src/borg/fuse.py:385:1: E302 expected 2 blank lines, found 1
```

ThomasWaldmann:
hard to review, would have been easier when first restructuring, then renaming attributes/methods in separate cs.

but I didn't see anything wrong and I assume that most stuff was just moved.

ThomasWaldmann:
thanks!

PR #3051, opened 2017-09-23
fix subprocess environments

ThomasWaldmann:
DONE (same potential issue):
- [x] --tar-filter
- [x] fusermount/umount
- [x] with-lock subprocess
- [x] xattr module fakeroot version check?


ThomasWaldmann:
```
$ ldd `which fakeroot`
	not a dynamic executable
```
Hmm, seems not to be dynamically linked on Ubuntu Linux.

But guess we can not rely on that and need to prepare the env for that subprocess also?

PR #3061, opened 2017-09-25
Simplified rate limiting wrapper in FAQ

Self-Perfection:
Exporting $RATE as environment variable is not need in this case.

And example does not use any bash specific features.
It should use default system shell instead.


ThomasWaldmann:
@Self-Perfection looks good, thanks for the simplification.

PR #3084, opened 2017-10-03
Quickstart guide minor corrections

TBurchfield:
Minor grammar correction and added consistency with substitution of 'Borg' in quickstart guide.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -28,7 +28,7 @@ reserve some space in |project_name| itself just to be safe by adjusting the
 ``config`` file. A good starting point is ``2G``.
 
 If |project_name| runs out of disk space, it tries to free as much space as it
-can while aborting the current operation safely, which allows it to free more space
+can while aborting the current operation safely, which allows the user to free more space
 by deleting/pruning archives. This mechanism is not bullet-proof in some
 circumstances [1]_.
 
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -21,14 +21,14 @@ a good amount of free space on the filesystem that has your backup repository
...
 on file systems which do not support this mechanism (e.g. XFS) we recommend to
-reserve some space in Borg itself just to be safe by adjusting the
+reserve some space in |project_name| itself just to be safe by adjusting the
 ``additional_free_space`` setting in the ``[repository]`` section of a repositories
 ``config`` file. A good starting point is ``2G``.
 
 If |project_name| runs out of disk space, it tries to free as much space as it
-can while aborting the current operation safely, which allows to free more space
+can while aborting the current operation safely, which allows it to free more space
 by deleting/pruning archives. This mechanism is not bullet-proof in some
```
ThomasWaldmann: the "allows to free" refers to the admin who can choose to run delete/prune then if he/she likes.

borg won't do that by itself / automatically, so no "it", IMHO.

jdchristensen: "allows to free" is not correct grammar, so it has to be "allows one to free more space", "allows the user to free more space" or "allows more space to be freed".

ThomasWaldmann: ah, thanks!

PR #3085, opened 2017-10-03
auto compression: make sure expensive compression is actually better

ThomasWaldmann:
if it is not significantly better compressed, we just store lz4 compressed data (which we already have computed anyway), because that at least decompresses super fast.

### DIFF
--- a/src/borg/testsuite/compress.py
+++ b/src/borg/testsuite/compress.py
@@ -110,12 +110,18 @@ def test_compressor():
 
 
 def test_auto():
-    compressor = CompressionSpec('auto,zlib,9').compressor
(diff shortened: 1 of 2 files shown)


src/borg/testsuite/compress.py
```diff
@@ -110,12 +110,18 @@ def test_compressor():
 
 
 def test_auto():
-    compressor = CompressionSpec('auto,zlib,9').compressor
-
-    compressed = compressor.compress(bytes(500))
-    assert Compressor.detect(compressed) == ZLIB
```
ThomasWaldmann: this test failed. looks like ZLIB compressed this data worse than LZ4 did.

PR #3088, opened 2017-10-06
List help topics when invalid topic is requested

n-st:
Currently, the available help topics are only listed manually, scattered throughout different parts of the documentation.  
It would be helpful to have an option for `borg help` that lists all available topics (e.g. `borg help --list-topics`).

Until that's implemented, this commit adds a list of available topics to the "topic not found" error message.  
This might also remain useful even after the separate flag has been added. Alternatively, we could replace the list with a mention of the new `--list-topics` once that exists.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1951,7 +1951,12 @@ class Archiver:
             else:
                 commands[args.topic].print_help()
         else:
-            parser.error('No help available on %s' % (args.topic,))
+            msg_lines = []
+            msg_lines += ['No help available on %s.' % args.topic]
+            msg_lines += ['Try one of the following:']
+            msg_lines += ['    Commands: %s' % ', '.join(sorted(commands.keys()))]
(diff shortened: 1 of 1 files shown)


enkore:
Topics should probably be listed in `borg [--]help`, just like commands. This may be hacky, though, with argparse.

ThomasWaldmann:
I agree with @enkore, that would be a better place (and also less strange than having to request an invalid topic just to see the list of topics). @n-st can you try to get it in there?

ThomasWaldmann:
@n-st did you read @enkore's and my feedback?

n-st:
Apologies for not replying earlier.  
I did read your feedback, but didn't have time to pursue this beyond the quick'n'dirty fix I already pushed.  
Not sure when I'll find time to work on this, so this improvement is effectively up for grabs by anyone who would like to implement it.

ThomasWaldmann:
I've looked at it again.

We need both:
- the changes in this changeset (to be more helpful than "no help available on")
- the change @enkore suggested

ThomasWaldmann:
thanks!

PR #3112, opened 2017-10-09
use --format rather than --list-format in examples, the latter is deprecated

FabioPedretti:
Should be pushed to master and 1.1-maint.


ThomasWaldmann:
thanks!

Issue #3115, opened 2017-10-09
can't use --chunker-params to achieve bigger chunks


willyvmm:
I would like to use bigger chunks up to 32Mb (and maybe bigger), but it does not work, what I'm doing wrong?

```shell
willy@browar: ~/borg # borg create -s --chunker-params 19,25,21,4095 /tmp/borgtest::test1 /root/borg/                                                                                                 
usage: borg create [-h] [--critical] [--error] [--warning] [--info] [--debug]                      
                   [--debug-topic TOPIC] [-p] [--log-json]                                         
                   [--lock-wait SECONDS] [--show-version] [--show-rc]                              
                   [--no-files-cache] [--umask M] [--remote-path PATH]                             
                   [--remote-ratelimit RATE] [--consider-part-files]                               
                   [--debug-profile FILE] [-n] [-s] [--list]                                       
                   [--filter STATUSCHARS] [--json] [--no-cache-sync]                               
                   [-e PATTERN] [--exclude-from EXCLUDEFILE]                                       
                   [--pattern PATTERN] [--patterns-from PATTERNFILE]                               
                   [--exclude-caches] [--exclude-if-present NAME]                                  
                   [--keep-exclude-tags] [-x] [--numeric-owner] [--noatime]                        
                   [--noctime] [--ignore-inode] [--files-cache MODE]                               
                   [--read-special] [--comment COMMENT]                                            
                   [--timestamp TIMESTAMP] [-c SECONDS]                                            
                   [--chunker-params PARAMS] [-C COMPRESSION]                                      
                   ARCHIVE [PATH [PATH ...]]     
borg create: error: argument --chunker-params: invalid ChunkerParams value: '19,25,21,4095'        
willy@browar: ~/borg # borg create -s --chunker-params 19,23,21,4095 /tmp/borgtest::test1 /root/borg/                                                                                                 
------------------------------------------------------------------------------
Archive name: test1
Archive fingerprint: 8e660562d23f25035b2e309301f02b74810968f82d51d56a79b1125aafeac331
Time (start): Mon, 2017-10-09 19:14:47
Time (end):   Mon, 2017-10-09 19:14:48
Duration: 0.35 seconds
Number of files: 370
Utilization of max. archive size: 0%
------------------------------------------------------------------------------
                       Original size      Compressed size    Deduplicated size
This archive:               24.99 MB             19.06 MB             19.06 MB
All archives:               24.99 MB             19.06 MB             19.06 MB

[...]

PR #3116, opened 2017-10-09
Added zstandard compression support with mutithreading support.

willyvmm:
Added support for Zstandard - Fast real-time compression algorithm http://www.zstd.net
https://github.com/facebook/zstd

prerequisites:
python-zstandard library https://github.com/indygreg/python-zstandard
pip3 install zstandard

There is also python-zstd implementation. It will not work, as this offer only basic support

    Features:
         - compression levels from 1 to 22 supported
            default value is 5

         - multithreaded compression support
            0: maximum cpu count
            -1..-99: maximum cpu count minus 1..99
            1..99: defined cpu usage
            default value is 1

     Usage: borg create -C zstd,level,threads ...

    Warning: This is NOT THREAD SAFE IMPLEMENTATION
      It allow only ONE python context to be created.
      It should work flawless as long as borg will call ONLY ONE compression job at time.

--
I'am not too familiar with Python/git, so please review my code. Thanks.

### DIFF
--- a/src/borg/compress.pyx
+++ b/src/borg/compress.pyx
@@ -22,6 +22,21 @@ try:
 except ImportError:
     lzma = None
 
+import multiprocessing as mp
+
+"""
+python-zstandard
+https://github.com/indygreg/python-zstandard
+pip3 install zstandard
+There is also python-zstd implementation. It will not work as this offer only basic support
+"""
+
+try:
+    import zstd 
+except ImportError:
+    zstd = None
+
+
 from .helpers import Buffer, DecompressionError
 
 API_VERSION = '1.1_03'
@@ -126,7 +141,8 @@ class LZ4(CompressorBase):
         osize = LZ4_compressBound(isize)
         buf = buffer.get(osize)
         dest = <char *> buf
-        osize = LZ4_compress_limitedOutput(source, dest, isize, osize)
+        with nogil:
+            osize = LZ4_compress_limitedOutput(source, dest, isize, osize)
         if not osize:
             raise Exception('lz4 compress failed')
         return super().compress(dest[:osize])
@@ -149,7 +165,8 @@ class LZ4(CompressorBase):
             except MemoryError:
                 raise DecompressionError('MemoryError')
             dest = <char *> buf
-            rsize = LZ4_decompress_safe(source, dest, isize, osize)
+            with nogil:
+                rsize = LZ4_decompress_safe(source, dest, isize, osize)
             if rsize >= 0:
                 break
             if osize > 2 ** 27:  # 128MiB (should be enough, considering max. repo obj size and very good compression)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks for the PR, but did you read our ticket / policy about adding compression algorithms?

ThomasWaldmann:
Hmm, seems like this can work as a optional compression, so it is not required to have the additional python / binary dependencies on every platform / distribution, so it might be compatible with the policy.

src/borg/compress.pyx
```diff
@@ -22,6 +22,21 @@ try:
 except ImportError:
     lzma = None
 
+import multiprocessing as mp
```
ThomasWaldmann: without "as mp" please. just search for "mp" to see why.

src/borg/compress.pyx
```diff
@@ -22,6 +22,21 @@ try:
...
+
+"""
+python-zstandard
+https://github.com/indygreg/python-zstandard
+pip3 install zstandard
+There is also python-zstd implementation. It will not work as this offer only basic support
+"""
```
ThomasWaldmann: this needs to go into setup.py as a optional requirement.

src/borg/compress.pyx
```diff
@@ -126,7 +141,8 @@ class LZ4(CompressorBase):
         osize = LZ4_compressBound(isize)
         buf = buffer.get(osize)
         dest = <char *> buf
-        osize = LZ4_compress_limitedOutput(source, dest, isize, osize)
+        with nogil:
+            osize = LZ4_compress_limitedOutput(source, dest, isize, osize)
```
ThomasWaldmann: why did you change this?

src/borg/compress.pyx
```diff
@@ -149,7 +165,8 @@ class LZ4(CompressorBase):
             except MemoryError:
                 raise DecompressionError('MemoryError')
             dest = <char *> buf
-            rsize = LZ4_decompress_safe(source, dest, isize, osize)
+            with nogil:
+                rsize = LZ4_decompress_safe(source, dest, isize, osize)
```
ThomasWaldmann: why did you change this?

src/borg/compress.pyx
```diff
@@ -217,6 +234,50 @@ class ZLIB(CompressorBase):
...
+    zstd compression / decompression binding
+
+    Features:
+         - compression levels from 1 to 22 supported
+            default value is 5
+         - multithreaded compression support
+            0: maximum cpu count
+            -1..-99: maximum cpu count minus 1..99
+            1..99: defined cpu usage
+            default value is 1
```
ThomasWaldmann: we'll have to discuss about multithreading after the basic issues are fixed.

src/borg/compress.pyx
```diff
@@ -217,6 +234,50 @@ class ZLIB(CompressorBase):
...
+    def __init__(self, level=5, threads=1, **kwargs):
+        super().__init__(**kwargs)
+        self.level = level
+        self.threads = threads
+        if zstd is None:
+            raise ValueError('No zstd support found.')
+
+    def compress(self, data):
+        cctx = zstd.ZstdCompressor(level=self.level, threads=self.threads, write_content_size=True)
+        data = cctx.compress(bytes(data)) # not sure about that typecast but it works.
```
ThomasWaldmann: pep8: 2 blanks to the left of # and one to the right.

src/borg/compress.pyx
```diff
@@ -344,6 +406,36 @@ class CompressionSpec:
...
+            elif count == 3:
+                maxcpu = mp.cpu_count() #get cpu count
+                level = int(values[1])
+                if not 0 <= level <= 22:
+                    raise ValueError
+
+                threads = int(values[2])
+
+                if threads == 0:
+                    threads=maxcpu # use maximum avaliable cpu's
```
ThomasWaldmann: pep8: blanks around operators, so it is " = ". also fix typos: "available" and "CPUs"

src/borg/compress.pyx
```diff
@@ -344,6 +406,36 @@ class CompressionSpec:
...
+                elif threads < 0:
+                    threads = maxcpu + threads #threads is NEGATIVE ... remember !!
```
ThomasWaldmann: pep8

src/borg/compress.pyx
```diff
@@ -344,6 +406,36 @@ class CompressionSpec:
...
+                    if threads < 1: # too less cpu's
```
ThomasWaldmann: not enough CPUs

src/borg/compress.pyx
```diff
@@ -344,6 +406,36 @@ class CompressionSpec:
...
+                        raise ValueError
+                else:
+                    if threads > maxcpu: # too many cpu's
```
ThomasWaldmann: CPUs

enkore:
Multithreaded compressors generally only make sense for larger chunks of data as the compression ratio tends to suffer and little speedup is attained for small chunks of data and multiple threads.

willyvmm:
Multithreaded compression is related to this ticket: https://github.com/borgbackup/borg/issues/3115 but i can see it will not work as i imagined. 

I did PR to 1.1 fork because i need it to work :)

Thanks for all suggestions, i will rewrite the code and take all suggestions, and will  remove multithreading support.
I did some tests, and it seems that multiple threads was never spawned, zstd logic is using much bigger window and do not spawn more threads when it is working with standard size chunks. 
 
I think this ticket can be closed for now.

Best Regards 
Willy.

ThomasWaldmann:
@willyvmm thanks!

Next PR against master, we'll discuss then whether we could adopt/backport that to 1.1.x.

FabioPedretti:
@willyvmm any news? :)

ThomasWaldmann:
@willyvmm did you do more work on this?

now as zstd made it even into the linux kernel, I guess we also could take it. :)

it would be perfect if it could be an optional dependency (so borg can still work, even if there is no libzstd [or however it is called] is available).

ThomasWaldmann:
Hmm, I just started on this. Based on @willyvmm's stuff, addressing my own comments from above.

PR will come soon.

PR #3124, opened 2017-10-10
Added required -e switch to init command

jest:
Not sure if `-e none` should be recommended in README, maybe `repokey` should be used instead. But then, probably following examples should also be modified to accommodate for that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,14 +89,19 @@ Main features
 Easy to use
 ~~~~~~~~~~~
 
-Initialize a new backup repository and create a backup archive::
+Initialize a new, password-protected backup repository and create a backup archive::
 
-    $ borg init -e none /path/to/repo
+    $ borg init -e repokey-blake2 /path/to/repo
+    Enter new passphrase:
+    Enter same passphrase again:
+    Do you want your passphrase to be displayed for verification? [yN]: N
     $ borg create /path/to/repo::Saturday1 ~/Documents
+    Enter passphrase for key /path/to/repo:
 
 Now doing another backup, just to show off the great deduplication::
 
     $ borg create -v --stats /path/to/repo::Saturday2 ~/Documents
+    Enter passphrase for key /path/to/repo:
     -----------------------------------------------------------------------------
     Archive name: Saturday2
     Archive fingerprint: 622b7c53c...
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
user should do an informed decision. yup, do not show -e none.

jest:
Then, what to show? As it stands, the example is incorrect, `-e` is required; claiming "using Borg is as simple as ... [`borg init /repo/dir` follows]" is not true: it's harder, to the extent of adding a dash and letter E :-)

ThomasWaldmann:
borg 1.0 defaulted to repokey.
otoh, 1.1 has faster repokey with blake2b authentication.
so, repokey is backwards compat, while the other is faster if that is not needed.
that's why i said informed decision... :)

jest:
OK, but this is README, for people who use the tools for the first time, such as me. I have no feeling to what expect, only suggest that some `-e` option be given. Can be `blake2b` as well, it doesn't matter IMO in this context (read: just let me run this beast and play with its CLI).

ThomasWaldmann:
how about showing repokey in the command and putting a comment above, that the b2 variant is faster, but not compatible to borg 1.0?

jest:
Sound good to me

ThomasWaldmann:
@jest can you update the PR?

jest:
Done

README.rst
```diff
@@ -89,14 +89,19 @@ Main features
 Easy to use
 ~~~~~~~~~~~
 
-Initialize a new backup repository and create a backup archive::
+Initialize a new, password-protected backup repository and create a backup archive::
 
-    $ borg init /path/to/repo
+    $ borg init -e repokey-blake2 /path/to/repo
```
ThomasWaldmann: maybe do what we talked about?

README.rst
```diff
@@ -89,14 +89,19 @@ Main features
...
+    Enter new passphrase:
+    Enter same passphrase again:
+    Do you want your passphrase to be displayed for verification? [yN]: N
```
ThomasWaldmann: well, maybe we can omit the passphrase "dialogue" for brevity here (and also below).

this is the readme, it should stay relatively short.

jest: OK, do whatever you want. I'm done

ThomasWaldmann:
see #3150.

PR #3147, opened 2017-10-12
move files-cache related options to borg create, fixes #3146

### LINKED ISSUE #3146: borg: error: unrecognized arguments: --files-cache=disabled
$ borg -V
borg 1.1.0

```
borg prune ... -v -s --list --no-files-cache ...
...
Warning: "--no-files-cache" has been deprecated. Use "--files-cache=disabled" instead.
```
and

```bash
borg prune ... -v -s --list --files-cache=disabled ...
...
borg: error: unrecognized arguments: --files-cache=disabled
```

ThomasWaldmann:
while the files cache is only really used by borg create, the value
of this argument is used for the Cache context manager at more places,
e.g. also for borg prune.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2154,6 +2154,9 @@ class Archiver:
                               help='show/log the return code (rc)')
             add_common_option('--no-files-cache', dest='cache_files', action='store_false',
                               help='do not load/update the file metadata cache used to detect unchanged files')
+            add_common_option('--files-cache', metavar='MODE', dest='files_cache_mode',
+                              type=FilesCacheMode, default=DEFAULT_FILES_CACHE_MODE_UI,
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1326,7 +1326,7 @@ def do_prune(self, args, repository, manifest, key):
             keep += prune_split(archives, '%Y', args.yearly, keep)
         to_delete = (set(archives) | checkpoints) - (set(keep) | set(keep_checkpoints))
         stats = Statistics()
-        with Cache(repository, key, manifest, do_files=args.cache_files, lock_wait=self.lock_wait) as cache:
+        with Cache(repository, key, manifest, do_files=False, lock_wait=self.lock_wait) as cache:
```
ThomasWaldmann: for prune, we only use the chunks cache (as expected), so do_files can be just always False.

Issue #3160, opened 2017-10-13
switch off nodump behaviour or bsdflags


ThomasWaldmann:
we already had some issues with that:
- unraid showing uninitialized/random bsdflags #3127
- macOS shares showing nodump in general #2949
- the linux bsdflags support needs to open the fs item.
  - slowdown due to the extra open/ioctl?
  - also there was a bug with opening non-present device files. #3130

so, add a `--ignore-nodump`? (double negation, not nice, rather `--honor-nodump` for the opposite?

or even a `--no-bsdflags` to skip reading / archiving bsdflags in general?

both?

ThomasWaldmann:
about `--honor-nodump`:
- borg < 1.1.0: did not care about nodump, this just backed up everything
- borg 1.1.0: honored nodump, so did not back up stuff with nodump set (no matter whether this was intentional or just a bug, see top post)
- borg 1.1.1: we could go back to the < 1.1.0 behaviour by adding a `--honor-nodump` flag and not honoring nodump by default.

enkore:
> borg < 1.1.0: did not care about nodump, this just backed up everything

Only on Linux, since BSD flags have been supported since forever with the BSD API.

PR #3164, opened 2017-10-14
implement --nobsdflags and --exclude-nodump

ThomasWaldmann:
see #3160 

could be also `--no-bsdflags`, but we already have `--noatime` (similar for names in the code).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -154,7 +154,9 @@ def with_archive(method):
     @functools.wraps(method)
     def wrapper(self, args, repository, key, manifest, **kwargs):
         archive = Archive(repository, key, manifest, args.location.archive,
-                          numeric_owner=getattr(args, 'numeric_owner', False), cache=kwargs.get('cache'),
+                          numeric_owner=getattr(args, 'numeric_owner', False),
+                          nobsdflags=getattr(args, 'nobsdflags', False),
+                          cache=kwargs.get('cache'),
                           consider_part_files=args.consider_part_files, log_json=args.log_json)
         return method(self, args, repository=repository, manifest=manifest, key=key, archive=archive, **kwargs)
     return wrapper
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
any feedback?

enkore:
> In general it is a good idea to look at exactly what is slow and what is not. For Borg, almost always Borg is slow, and not the syscalls. Only if genuinely the syscalls cause a problem an option should be added; otherwise, the code should be fixed.

That being said, it may be reasonable to remove NODUMP-based exclusion of files entirely for 1.1, even if that call would be a bit late.

ThomasWaldmann:
OK, I'll add --exclude-nodump and it'll be off by default.

ThomasWaldmann:
OK, so now there is --exclude-nodump (default: false, to avoid confusing/unexpected default behaviour) for people who want NODUMP-flagged files not being backed up.

--nobsdflags is for these cases:
- borg create: avoid backing up crappy bsdflags from source filesystems not supporting them or accidentally setting strange bsdflags (like that case with a macOS share having NODUMP everywhere).
- borg extract: if you have already backed up crappy bsdflags, you for sure do not want to have them restored.

src/borg/archiver.py
```diff
@@ -2738,6 +2743,10 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                               help='do not store atime into archive')
         fs_group.add_argument('--noctime', dest='noctime', action='store_true',
                               help='do not store ctime into archive')
+        fs_group.add_argument('--nobsdflags', dest='nobsdflags', action='store_true',
+                              help='do not read and store bsdflags (e.g. NODUMP, IMMUTABLE) into archive')
+        fs_group.add_argument('--exclude-nodump', dest='exclude_nodump', action='store_true',
+                              help='honour the NODUMP flag: do not store flagged files into archive')
```
enkore: exclude files flagged NODUMP

enkore: I think this option should go into the exclusion group.

ThomasWaldmann:
@enkore if there's nothing else, I'ld collapse all changesets and merge.

PR #3165, opened 2017-10-14
update CHANGES (1.1-maint)

ThomasWaldmann:


docs/changes.rst
```diff
@@ -131,6 +131,39 @@ The best check that everything is ok is to run a dry-run extraction::
 Changelog
 =========
 
+Version 1.1.1rc1 (not released yet)
```
ThomasWaldmann: maybe we should even directly go to 1.1.1 to make the fixes available more widely/quickly.

enkore: Release candidates only make sense if a tangible risk of regressions is expected.

https://github.com/borgbackup/borg/compare/1.1.0...1.1-maint

Is it?

PR #3170, opened 2017-10-16
Documentation consistency.

TBurchfield:
Not all instances of "Borg" or "borg" outside of code were represented as "|project_name|".  I've corrected all that I found.  A few minor instances of rephrasing outside to reflect this most naturally.

I imported the global.rst.inc when necessary.


enkore:
I'd rather go the other way around, since |project_name| makes the plain text much harder to read and doesn't work in anything GitHub displays. It also makes it harder to see whether sentences are grammatically sound.

TBurchfield:
Sounds good.  I'll make a new PR when I get some time removing |project_name|

ThomasWaldmann:
When I forked borg from attic, I found it a bit tedious to replace all the "attic" in the docs by "borg", this is why there is this variable now.

PR #3177, opened 2017-10-17
remove warning about hardlinked symlinks, update docs, fixes #3175

### LINKED ISSUE #3175: hardlinked symlinks warning
see there: https://github.com/borgbackup/borg-import/issues/23

maybe the warning should include the path, so we have that information independently of --list.

ThomasWaldmann:


ThomasWaldmann:
Hmm, maybe it would be better to have this either
- not at all (no warning, like in 1.0)
- a one-time-per-backup-only warning (warning module)

PR #3194, opened 2017-10-20
Add Nix/NixOS install command

Baughn:
It may be more common to use declarative means, but this will tell NixOS users the package name and it's a valid install command.


ThomasWaldmann:
thanks!

PR #3204, opened 2017-10-22
release 1.1.1

ThomasWaldmann:


src/borg/logger.py
```diff
@@ -91,7 +91,9 @@ def setup_logging(stream=None, conf_fname=None, env_var='BORG_LOGGING_CONF', lev
     logger = logging.getLogger('')
     handler = logging.StreamHandler(stream)
     if is_serve and not json:
-        fmt = '$LOG %(levelname)s %(name)s Remote: %(message)s'
+        # if borg serve is not using json, we likely are talking to an old
+        # borg client that does not support json (like 1.0.x).
+        fmt = '$LOG %(levelname)s Remote: %(message)s'
```
ThomasWaldmann: otoh, progress output like 1.1 server "compacting segments" still ends up on 1.0 client's screen.

unclear how to fix that best, maybe it needs keeping the logger name here and a change in 1.0-maint to support the format and not output progress msgs. wouldn't help for past 1.0 releases...

@enkore ^^

PR #3212, opened 2017-10-23
don't do stat() when not recursing into excluded dir, fixes #3209

### LINKED ISSUE #3209: excluded directory still accessed, causing backup to fail on missing permissions
This is a regression in borg 1.1.0.

I'm running backups as `root` while regular users have some directories mounted via FUSE. `root` is not allowed to access those directories. Therefore I exclude them from being backed up with `--exclude /path/to/fuse-mount-point`.

In borg 1.0.9 this used to work fine. The directories were not backed up, borg did not emit an error message, and barring any other errors borg exited with code 0 signalling success.

In borg 1.1.0 it still tries to access that mount point. That fails, and borg emits an error message "Permission denied". The whole borg run exits with error code 1 in that case signalling an error.

Here's how to reproduce:

1. Create a repository
2. Set up a tree to back up with some files and a directory to be used as the FUSE mount point
3. Configure FUSE not to allow access to mounted directories by other users (in `/etc/fuse.conf`: option `user_allow_other` must not be active)
4. Mount a file system as a regular user on said mount point, e.g. with `sshfs`
5. Try to create a borg backup as root while excluding said mount point

Here's an example I've just run:

``` sh
[0 mbunkus@chai-latte ~] mkdir -p tmp/test/{tree,repo}
[0 mbunkus@chai-latte ~] cd ~/tmp/test/tree
[0 mbunkus@chai-latte ~/tmp/test/tree] borg init -e none ~/tmp/test/repo
[0 mbunkus@chai-latte ~/tmp/test/tree] echo chunky > bacon.txt
[0 mbunkus@chai-latte ~/tmp/test/tree] echo foo > bar.txt
[0 mbunkus@chai-latte ~/tmp/test/tree] mkdir fuse
[0 mbunkus@chai-latte ~/tmp/test/tree] sshfs sweet-chili: fuse
[0 mbunkus@chai-latte ~/tmp/test/tree] ls -l
total 12
-rw-rw-r-- 1 mbunkus Domain Users    7 Oct 23 12:41 bacon.txt
-rw-rw-r-- 1 mbunkus Domain Users    4 Oct 23 12:43 bar.txt
drwxr-xr-x 1    1000        10002 2468 Oct 23 10:38 fuse/
[2 mbunkus@chai-latte ~/tmp/test/tree] sudo borg create --exclude ~/tmp/test/tree/bar.txt --exclude ~/tmp/test/tree/fuse $HOME/tmp/test/repo::test ~/tmp/test/tree
Warning: Attempting to access a previously unknown unencrypted repository!
Do you want to continue? [yN] y
/home/mbunkus/tmp/test/tree: stat: [Errno 13] Permission denied: '/home/mbunkus/tmp/test/tree/fuse'
[1 mbunkus@chai-latte ~/tmp/test/tree] sudo borg list $HOME/tmp/test/repo::test
drwxrwxr-x mbunkus Domain Users        0 Mon, 2017-10-23 12:43:30 home/mbunkus/tmp/test/tree
-rw-rw-r-- mbunkus Domain Users        7 Mon, 2017-10-23 12:41:55 home/mbunkus/tmp/test/tree/bacon.txt
[0 mbunkus@chai-latte ~/tmp/test/tree]
```

The test shows that the `--exclude` statements are correct. The first exclusion of the file `bar.txt` is handled correctly by borg; the file isn't present in the archive. The second exclusion isn't handled correctly as the error message shows.

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -537,17 +537,24 @@ def _process(self, fso, cache, matcher, exclude_caches, exclude_if_present,
...
+            if  not matcher.recurse_dir:
+                return
+            if st is None:
+                with backup_io('stat'):
+                    st = os.stat(path, follow_symlinks=False)
+            recurse_excluded_dir = stat.S_ISDIR(st.st_mode)
+            if not recurse_excluded_dir:
                 return
 
         try:
```
ThomasWaldmann: in 1.0, the stat() call was inside this try/except block.

PR #3252, opened 2017-11-03
borg list: fix broken pipe handling, fixes #3245

### LINKED ISSUE #3245: 1.1.1 gives a BrokenPipeError when sending output to | less
With 1.1.1 when I run `borg list | less` I get this exception. It works fine on 1.0.11.

> Local Exception
> Traceback (most recent call last):
>   File "borg/archiver.py", line 4049, in main
>   File "borg/archiver.py", line 3977, in run
>   File "borg/archiver.py", line 148, in wrapper
>   File "borg/archiver.py", line 1356, in do_list
>   File "borg/archiver.py", line 1399, in _list_repository
> BrokenPipeError: [Errno 32] Broken pipe
> 
> Platform: Linux borg 4.9.0-4-amd64 #1 SMP Debian 4.9.51-1 (2017-09-28) x86_64
> Linux: debian 9.2
> Borg: 1.1.1  Python: CPython 3.5.4
> PID: 1956  CWD: /root/borg-script/1.1
> sys.argv: ['./borg-1.1.1', 'list']
> SSH_ORIGINAL_COMMAND: None
> 
> Exception ignored in: <_io.TextIOWrapper name='<stdout>' encoding='UTF-8'>
> BrokenPipeError: [Errno 32] Broken pipe

ThomasWaldmann:
always use sys.stdout.write as in main() we have replaced that with a
handler that deals with BrokenPipe exceptions.

also simplify (remove) encode/decode dance as format_item (get_item_data) already removes surrogates.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1148,26 +1148,18 @@ class Archiver:
     @with_repository(compatibility=(Manifest.Operation.READ,))
     def do_list(self, args, repository, manifest, key):
         """List archive or repository contents"""
-        if not hasattr(sys.stdout, 'buffer'):
-            # This is a shim for supporting unit tests replacing sys.stdout with e.g. StringIO,
-            # which doesn't have an underlying buffer (= lower file object).
-            def write(bytestring):
-                sys.stdout.write(bytestring.decode('utf-8', errors='replace'))
-        else:
-            write = sys.stdout.buffer.write
-
         if args.location.archive:
             if args.json:
                 self.print_error('The --json option is only valid for listing archives, not archive contents.')
                 return self.exit_code
-            return self._list_archive(args, repository, manifest, key, write)
+            return self._list_archive(args, repository, manifest, key)
         else:
             if args.json_lines:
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1147,13 +1147,8 @@ def do_umount(self, args):
...
         """List archive or repository contents"""
-        if not hasattr(sys.stdout, 'buffer'):
-            # This is a shim for supporting unit tests replacing sys.stdout with e.g. StringIO,
-            # which doesn't have an underlying buffer (= lower file object).
-            def write(bytestring):
-                sys.stdout.write(bytestring.decode('utf-8', errors='replace'))
-        else:
-            write = sys.stdout.buffer.write
+        def write(bytestring):
+            sys.stdout.write(bytestring.decode('utf-8', errors='replace'))
```
ThomasWaldmann: so, in the end it first calls safe_encode (utf-8 encode, surrogate-escape) and then decode (utf-8, replace).

guess we could just call remove_surrogates (utf-8 encode, replace + utf-8 decode) - like at a lot of other places.

Issue #2862, opened 2017-07-22
Officially maintain/include autocompletion


rugk:
IMHO autocompletion for mayor shells (zsh, bash) is a feature each good linux tool should have. And it should be maintained officially in this repo and included in it, because then packagers rather see it and can be assured that support for it won't be dropped easily.

Maybe put it in some doc or misc dir.

This issue means include 
https://github.com/mrkmg/borgbackup-zsh-completion into this repo (& update it) and include https://github.com/borgbackup/borg/issues/1782 here.

ThomasWaldmann:
@rugk there are quite some different shells out there, so having this packaged/maintained separately seems to be easier for me.

rugk:
Mhh, then maybe at least maintain/include them in the official [BorgBackup organisation](https://github.com/borgbackup) in contrast to "just some community scripts/things". Maybe make a "shell-autocompletion" repo, where all these can be maintained.

ThomasWaldmann:
I can do that if at least 2 maintainers of shell completion show interest in that.
Alternatively, they could also get linked from the community repo's link list.

rugk:
If @mrkmg and maybe someone solving https://github.com/borgbackup/borg/issues/1782 show interest we have already two people.

mrkmg:
I am willing to maintain the zsh auto-completion. I use borg daily, although I would probably need to be alerted to flag changes as my use is fairly regular and consistent.

As for bash, I do not use bash nor do I have a good understanding of how to setup auto-completion for it.

rugk:
So I've suggested the Debian package maintainer to include the borg autocompletion scripts in the package. (bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=869381) and the reply was:
> I really prefer to see them merged upstream before applying them.
> please upstream them and ping back, thanks!

So I think that's another argument for "merging" – or at least moving them to the borgbackup organisation. So, at least, currently, the zsh auto-completion could be maintained…

bpereto:
There is also a  request on the fedora side from @rugk :
https://bugzilla.redhat.com/show_bug.cgi?id=1473973

I would package and maintain a seperate package for fedora (or in the same if its merged to the main repo), but I think it should be maintained by one maintainer/project, preferable the borgbackup project, instead of each distro reinventing the autocompletion-wheel.

ThomasWaldmann:
[...]

PR #3268, opened 2017-11-04
Add bash and fish shell auto-completions.

### LINKED ISSUE #3086: [question] Fish completions to maintain here or at fish-shell?
I wrote tab completion definitions for the fish-shell.
https://github.com/SanskritFritz/fish-shell/blob/patch-1/share/completions/borg.fish
My question is, would you like to include it into the borg distribution or should I merge it to fish? Fish does have a history of including completions for foreign packages, so that would not be a problem.
However, if borg includes the completions, changes in borg would be reflected sooner in the completion definitons. I'm willing to maintain it, even over at fish-shell.

SanskritFritz:
As mentioned in #2862, this PR adds bash and fish shell completions. Zsh will be added separately by @mrkmg. This PR fixes #3086.

### DIFF
--- a/scripts/shell_completions/bash/borg
+++ b/scripts/shell_completions/bash/borg
@@ -1,0 +1,165 @@
+#!/bin/bash
+
+# Completions for borg version 1.1.1
+# https://www.borgbackup.org/
+# Note:
+# Listing archives works on password protected repositories only if $BORG_PASSPHRASE is set.
+# Install:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
thanks!

ThomasWaldmann:
can you do another PR against 1.1-maint?

SanskritFritz:
Sure.

PR #3271, opened 2017-11-04
Add bash and fish tab completions.

SanskritFritz:
As requested in #3268.


SanskritFritz:
Oops, this is now what we want. Closing...

PR #3285, opened 2017-11-05
Fix build with LibreSSL

olesalscheider:


src/borg/crypto/_crypto_helpers.c
```diff
@@ -4,7 +4,7 @@
 #include <openssl/opensslv.h>
 #include <openssl/hmac.h>
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
```
ThomasWaldmann: hmm, you don't actually check the libressl version number.

guess that's because there is no openssl 1.1 compatible libressl yet (not even planned?).

as we plan to maybe require openssl >= 1.1 in the future to support AES-OCB and chacha20-poly1305, the question is whether it makes sense to add compatibility for libressl now if it would then break again.

can you tell where you see the benefits of supporting libressl?

olesalscheider: I'm not aware of any plans to make the libressl API compatible with the newer openssl versions. I guess that they will support AES-OCB and chacha20-poly1305 but I don't know about that.

The benefit of supporting libressl is that some distributions switched to it and others (e. g. Exherbo) allow the user to choose. If borg does not support libressl it cannot be used on these distributions...

ThomasWaldmann: hmm, looks like chacha20-poly1305 is already there, but not AES-OCB.

guess you can always use the binary of borg, which comes with openssl (and everything else) included.

olesalscheider: Sure, that should work. But of course that won't be packaged by the distributions and the users have to get it themselves.

If this fix is an issue in the long term, ok. But right now it seems trivial...
And my assumption is that libressl will provide all the functionality soon.

ThomasWaldmann:
Guess whether this make sense long term depends on whether somebody wants to maintain the libressl compatibility code needed.

As of now (and as long as we support openssl 1.0.x), this is quite easy likely.

But if we require openssl 1.1 at some time (because we need some stuff that does not come with openssl 1.0) and if libressl can not provide same functionality, more code / docs will be needed and in the worst case, we'ld have to drop libressl again if nobody can / will make it work.

olesalscheider:
Would it really be a problem to accept it for now and drop it if supporting libressl really becomes infeasible in the future?
I mean for most projects it is not a big delta and libressl is also continuously improved. So I think that there is a good chance that it will support everything needed in the future.

ThomasWaldmann:
OK, merged. Can you do a PR against 1.1-maint also?

olesalscheider:
I think a PR against 1.1-maint is not necessary. It builds fine here with libressl and it does not contain the code that I ifdef-ed.

ThomasWaldmann:
ah, ok. might be due to a change in the huge crypto-aead PR that only went into master. thanks for checking!

Issue #3216, opened 2017-10-25
borgfs is missing a man page


ThomasWaldmann:
and therefore, it is listed there:

https://qa.debian.org/man-pages.html

PR #3290, opened 2017-11-05
Generate usage & man page for borgfs

### LINKED ISSUE #3287: borgfs still shows 'mount' in options
When running `borgfs --help`, `borgfs mount`, `borgfs`, etc. the short usage still shows that you should put `mount` after borgfs:

    usage: borgfs mount [-h] [--critical] [--error] [--warning] [--info] [--debug]
                        [--debug-topic TOPIC] [-p] [--log-json]
                        [--lock-wait SECONDS] [--show-version] [--show-rc]
                        [--umask M] [--remote-path PATH] [--remote-ratelimit RATE]
                        [--consider-part-files] [--debug-profile FILE] [-f]
                        [-o OPTIONS] [-P PREFIX | -a GLOB] [--sort-by KEYS]
                        [--first N | --last N]
                        REPOSITORY_OR_ARCHIVE MOUNTPOINT

Whereas the first line should just be:

    usage: borgfs [-h] [--critical] [--error] [--warning] [--info] [--debug]

milkey-mouse:
I also fixed the borgfs help text so it no longer says it needs the `mount` subcommand as well.
Fixes #3287 and #3216.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -242,10 +242,11 @@ class build_usage(Command):
         # allows us to build docs without the C modules fully loaded during help generation
         from borg.archiver import Archiver
         parser = Archiver(prog='borg').build_parser()
+        borgfs_parser = Archiver(prog='borgfs').build_parser()
 
-        self.generate_level("", parser, Archiver)
+        self.generate_level("", parser, Archiver, {'borgfs': borgfs_parser})
 
-    def generate_level(self, prefix, parser, Archiver):
+    def generate_level(self, prefix, parser, Archiver, extra_choices=None):
         is_subcommand = False
         choices = {}
         for action in parser._actions:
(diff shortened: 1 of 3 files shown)


milkey-mouse:
This has some weird test errors on `py35` but I don't think they're my fault...

setup.py
```diff
@@ -266,7 +269,7 @@ def generate_level(self, prefix, parser, Archiver):
             if self.generate_level(command + " ", parser, Archiver):
                 continue
 
-            with open('docs/usage/%s.rst.inc' % command.replace(" ", "_"), 'w') as doc:
+            with open('docs/usage/%s.rst%s' % (command.replace(" ", "_"), ".inc" if command != "borgfs" else ""), 'w') as doc:
```
ThomasWaldmann: is that needed? guess we could also have usage/borgfs.rst including the .inc it would generate without that change.

milkey-mouse: ok, will try that

milkey-mouse:
yes, --help works on every level

ThomasWaldmann:
@enkore any feedback from you?

ThomasWaldmann:
Assuming you have looked at the man page generated and verified it is ok, I guess this is finished.

Can you collapse the fixup commit with the commit it is fixing?

Issue #14, opened 2015-05-15
advanced sparse file support


ThomasWaldmann:
See there for the basics: https://github.com/jborg/attic/issues/256

The current state in borg is that it has simple sparse file support (meaning that it does nothing special on "create", but offers the option to deal with all-zero chunks in 2 ways at "extract" time: a) write zeros to disk (default) b) just "seek" in the output file, creating a hole in a sparse file (--sparse).

See the original attic ticket: while this always works correctly from a logical file content perspective, it is _not_ extracting the data and hole sections of files exactly as they were when the archive was made.

Precise reproduction would need saving the type (data vs hole), length and in case of data, the binary data for each section in the file. SEEK_DATA and SEEK_HOLE support this.

Attic (and current Borg) just has a simple stream of binary file contents and as there is no type/length yet, it can't be added in a compatible way on the file contents level.

Update: I put a bounty on this. It is for implementing sparse file support at archive creation time (seeking over holes [not reading them as zeros], storing which parts of a file are holes / contain data) and reproducing sparse files hole/data layout precisely at archive extraction time.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/15075076-advanced-sparse-file-support)

ThomasWaldmann:
Just to keep a vague idea:

If the different layers in borg would communicate a bit more (like the compression method being flexibly given by upper layers and the decompression result being able to tell whether its result is data that needs to be written to disk or just a sparse hole that needs to be "seeked"), the adv. sparse file support could maybe be just a special kind of "compression/decompression".

sparse compression would just store: 2 type bytes for "sparse", 8 bytes length of hole
sparse decompression would result in: hole=True, hole length
other decompression would result in: hole=False, data

ThomasWaldmann:
Advanced sparse file support would also save some time for archive creation as it does not have to read all these zeros in the sparse wholes (simple sparse file support just reads them all).

enkore:
The less involved feature, not having to chunk-HMAC hole contents is handled in #1354

ThomasWaldmann:
@milkey-mouse bs just told me you have started on this. maybe first evaluate how much changes this needs - if it is a bigger change, it might be not applicable to 1.1-maint (and also might conflict with 1.2 goals).

milkey-mouse:
[...]

PR #3292, opened 2017-11-06
Add instructions for using ntfsclone

### LINKED ISSUE #81: try / document borg + ntfsclone
could be useful to clone windows systems (with ntfs filesystems)?
due to ntfsclone, would only save allocated blocks.

try it and document your results.

### LINKED ISSUE #3259: update docs bounty (11)
there are quite some docs issues: https://github.com/borgbackup/borg/labels/documentation

grab some doc update issues, create a PR against master (and, after discussion, also to 1.1-maint and 1.0-maint, if applicable, see [here](http://borgbackup.readthedocs.io/en/latest/development.html#branching-model), also linked from PR submission template)

the bounty is for fixing 2 small issues or 1 bigger issue.
docs issues that already have a bounty on them do not qualify.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/51130572-update-docs-bounty-11)

milkey-mouse:
Fixes #81 and fixes #3259.


docs/faq.rst
```diff
@@ -48,6 +48,44 @@ and platform/software dependency. Combining Borg with the mechanisms provided
 by the platform (snapshots, hypervisor features) will be the best approach
 to start tackling them.
 
+How can I decrease the size of disk image backups?
+--------------------------------------------------
+
+Full disk images are as large as the full disk when uncompressed and don't get much
+smaller post-deduplication after heavy use. This is because virtually all file systems
```
ThomasWaldmann: guess one can't tell the dedup reduction that general (the user might have tons of duplicate data **in** existing files).
but it gets clearer what you mean in the following text, so maybe fix phrasing above like "and might not get...".

docs/faq.rst
```diff
@@ -48,6 +48,44 @@ and platform/software dependency. Combining Borg with the mechanisms provided
...
+space it takes up when compressed and deduplicated. Depending on the filesystem of the
+VM (or physical computer, if for some reason a normal filesystem backup can't be taken),
+there are several ways to decrease the size of a full image:
+
+Using ntfsclone (NTFS, i.e. Windows VMs)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+ntfsclone can only operate on filesystems with the journal cleared (i.e. turned-off
+machines) which somewhat limits its utility in the case of VM snapshots. However,
+when it can be used, its special image format is even more efficient than just zeroing
```
ThomasWaldmann: is it still more efficient over time, like when you daily use the VM and backup it with borg?

docs/faq.rst
```diff
@@ -48,6 +48,44 @@ and platform/software dependency. Combining Borg with the mechanisms provided
...
+and deduplicating. For backup, save the MBR and the contents of each partition::
+
+    sfdisk -d /dev/sdX > sfdisk.txt
+    dd if=/dev/sdX of=mbr_gap count=2048  # all sectors until 1st partiton
+    borg create repo::hostname-partinfo-mbr-gap sfdisk.txt mbr_gap
+    sfdisk -l -o Device,Type /dev/sdX | grep NTFS | cut -d' ' -f1 | while read x; do
+        ntfsclone -s -o - $x | borg create repo::hostname-part$(echo $x | grep -Eo "[0-9]+$") -
+    done
```
ThomasWaldmann: hmm, this reads a bit strange as this seems to be about a physical disk / machine and last thing you mentioned above was a VM.

maybe add a "check this!" in the comment of the dd, it depends on layout. there is no problem if we save a bit more than needed, but there might be a problem if we do not save the whole gap (if it is larger).

you practically tested the script?

docs/faq.rst
```diff
@@ -48,6 +48,44 @@ and platform/software dependency. Combining Borg with the mechanisms provided
...
+
+Restoration is essentially the same process done in reverse::
+
+    borg extract repo::hostname-partinfo-mbr-gap
+    sfdisk /dev/sdX < sfdisk.txt
+    dd if=mbr_gap of=/dev/sdX bs=1M
```
ThomasWaldmann: the dd is writing over the mbr, which was just edited by sfdisk.
also maybe rather use count=2048, like above.

docs/faq.rst
```diff
@@ -48,6 +48,44 @@ and platform/software dependency. Combining Borg with the mechanisms provided
...
+    sfdisk -l -o Device,Type /dev/sdX | grep NTFS | cut -d' ' -f1 | while read x; do
+        borg extract --stdout repo::hostname-part$(echo $x | grep -Eo "[0-9]+$") | ntfsclone -r -O $x -
+        ntfsclone -s -o - $x | borg create repo::hostname-part -
```
ThomasWaldmann: what is this line for?

docs/faq.rst
milkey-mouse: I would need to write a test script (edit some files in NTFS partitons & create archives with zeroing vs. ntfsclone format, rinse & repeat, compare final sizes) but the ntfsclone man page recommends always using the special format if you can control both backup and restore (as we do here). ntfsclone can clone to sparse files, so maybe once #14 is implemented that could simplify these instructions.

docs/faq.rst
milkey-mouse: I tested it on a MBR-formatted loopback device with a few NTFS partitions, I am testing it with a full windows installation right now.

On second thought I think saving the MBR with `sfdisk` is kind of redundant, when copying all space leading up to the first partition should do virtually the same thing. I can parse that out of the `sfdisk` output (it seems to always be 2048 bytes on MBR disks). Maybe instead I should use `dd` to copy up to the first partition and then use `partprobe` to notify the kernel of the new partition layout instead of using `sfdisk` at all.

Should these instructions back up every partition, NTFS or not, using ntfsclone for ntfs ones? At that point this is becoming more full bash script territory instead of just a snippet.

docs/faq.rst
milkey-mouse: It shouldn't be there, I copy-pasted the wrong line from my test script

docs/faq.rst
milkey-mouse: I just tested it on a Windows VM, which obviously used the MBR format. I added several partitions and checked restore, all the files were in the same places on the secondary NTFS & fat32 partitions and Windows booted (obviously I didn't meticulously check every windows system file but if it booted I'm assuming they stayed in place).

docs/faq.rst
ThomasWaldmann: guess the stuff we need to compare for efficiency are these:
- zeroing free space + running borg on the block device with --read-special (reading all the zeros will take a while, but they will dedup / compress to almost nothing. also, everything else stays in place and unchanged sectors will likely dedup ok with future block device images)
- using ntfsclone in special format (maybe dedups badly with future dumps due to special format?)

milkey-mouse:
shall I collapse the ntfsclone fixup?

ThomasWaldmann:
yes.

maybe we should even move the whole stuff into docs/misc/disk-imaging.txt or so? gets rather long for a faq entry and includes quite some scripting...

milkey-mouse:
I agree, especially if we add a section with benchmarks for ntfsclone special/ntfsclone zeroed/zerofree/raw --read-special, we will need a separate docs page.

milkey-mouse:
When you say to move everything into a `disk-imaging.txt`, do you mean something like putting both FAQ entries in the txt and reference it from its former place on the FAQ page?

ThomasWaldmann:
yes. alternatively, it could be an own chapter about disk imaging. not sure what's better, a docs section is maybe easier discoverable, so it would not need a link from faq.

milkey-mouse:
The new page could go under the Usage section, then I can keep the rst formatting as well

milkey-mouse:
If existing content is being moved to a new place, we don't want links to 1.1 docs to break, should this still be backported?

ThomasWaldmann:
if someone wants permanent links, a link to a specific version can be used (like 1.1.2).

PR #3297, opened 2017-11-06
Clarify encryption of key for borg key export

### LINKED ISSUE #3296: check docs about key export
i remember multiple people asking about whether the data exported by `borg key export` contains the key in encrypted or unencrypted form.

this should be in the docs, also some related statement about the passphrase.

milkey-mouse:
Fixes #3296. Also explicitly documented the `--qr-html` option in the `borg key export` help.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2498,10 +2498,11 @@ class Archiver:
         If repository encryption is used, the repository is inaccessible
         without the key. This command allows to backup this essential key.
 
-        There are two backup formats. The normal backup format is suitable for
+        There are three backup formats. The normal backup format is suitable for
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2497,11 +2497,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         key_export_epilog = process_epilog("""
         If repository encryption is used, the repository is inaccessible
         without the key. This command allows to backup this essential key.
+        Note that the backup produced does not include the passphrase itself
+        (i.e. the exported key stays encrypted).
```
ThomasWaldmann: maybe we should point out here that one needs both key and passphrase.

ThomasWaldmann:
thanks!

PR #3300, opened 2017-11-07
borg mount: support exclusion group options and paths, fixes #2138

### LINKED ISSUE #2138: borg mount only a selected PATH (as in extract)
Given the fairly amount of CPU power needed for the "versions" option (which I found utterly useful), I'd like to ask for restrict the mount to just a specific PATH. IMO, the use of the versions option typically means you are looking for the versions of a (some) specific (set of) file(s).

E.g. OMG! Someone, at some time, modified "Documents/Important/analysis.odt". I need to recover the previous version (whenever it was). I don't want to mount all the (potentially huge) directory tree with all the versions. Just that file/directory.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/41830614-borg-mount-only-a-selected-path-as-in-extract)

ThomasWaldmann:
borg mount [options] repo_or_archive mountpoint path [paths...]

paths: you can just give some "root paths" (like for borg extract) to
only partially populate the FUSE filesystem.

Similar for these exclusion group options:
--exclude
--exclude-from
--pattern
--patterns-from
--strip-components

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2325,6 +2325,9 @@ class Archiver:
         subparser.add_argument('-o', dest='options', type=str,
                             help='Extra mount options')
         define_archive_filters_group(subparser)
+        subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
+                               help='paths to extract; patterns are supported')
+        define_exclusion_group(subparser, strip_components=True)
         if parser.prog == 'borgfs':
             return parser
 
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -119,7 +119,7 @@ class ItemCache:
         else:
             raise ValueError('Invalid entry type in self.meta')
 
-    def iter_archive_items(self, archive_item_ids, filter=None):
+    def iter_archive_items(self, archive_item_ids, filter=None, consider_part_files=False):
         unpacker = msgpack.Unpacker()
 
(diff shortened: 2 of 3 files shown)


alejandro-perez:
This is beautiful!
I tested it, and the results are amazing. Just mounted my Dropbox folder (yes I backup that one too, trust none), which has 2GB and 5650 files, and it took 1m22s and only 92m RAM, as opposed to the 4m1s and 1100m RAM of mounting the whole thing. And all that in less than 30 lines of code, astonishing.

alejandro-perez:
Another example. Mounting my LXC repository takes 10 minutes and 4,6GB RAM, but if I just select one container, it takes only 3m:30s and just 200MB RAM.
Savings in time are good, but savings in memory can be really huge, from making it unusable (if you have a really big repository and not that much RAM) to make it completely usable. So thanks again.

ThomasWaldmann:
rebased on master, resolved conflicts. ^^^

ThomasWaldmann:
Updated again, now it should work correctly for hardlinks, even if the hardlink master (first hardlink) is excluded.

ThomasWaldmann:
rebased again, minor cosmetic fixes.

if someone wants to try / review, feel free.

ThomasWaldmann:
I'll do some manual testing now, then merge and backport to 1.1 ...

PR #3305, opened 2017-11-10
Show an error message when --dry-run & --stats are used simultaneously

### LINKED ISSUE #3298: Dry Run Statistics, please implement or document
Hi,

this is a follow-up to #265.

I don't agree with the with the way it was closed. And I was tricked by the same assumption as shown there, and there is STILL no documentation nor implementation of this functionality.

Seriously, I do expect that "--stats -C none ..." would give me some useful information. That is what I get from rsync when used for similar purpose ("rsync -a --stats --checksum ...") and a such limitation in borg seems to be weird, frankly speaking.

milkey-mouse:
Fixes #3298. `borg create --dry-run` doesn't actually compress/dedupe data, so it can't display stats on the created archive. Previously, running `borg create --dry-run --stats` would fail silently (by not showing stats).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2739,10 +2739,10 @@ class Archiver:
                                           help='create backup')
         subparser.set_defaults(func=self.do_create)
 
-        group = subparser.add_mutually_exclusive_group()
-        group.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
+        dryrun_group = subparser.add_mutually_exclusive_group()
+        dryrun_group.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not create a backup archive')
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2738,10 +2739,12 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                                           help='create backup')
         subparser.set_defaults(func=self.do_create)
 
-        subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
+        group = subparser.add_mutually_exclusive_group()
```
ThomasWaldmann: as dryrun is not doing quite some stuff, maybe name that after it: dryrun_group

ThomasWaldmann:
thanks!

PR #3312, opened 2017-11-11
Add borg config command

### LINKED ISSUE #3304: 'borg config' command
Similar to [`git config`](https://git-scm.com/docs/git-config), it would be nice to have a convenience `borg config` command that just opens the repo config and changes an option accordingly.

milkey-mouse:
Fixes #3304. Similar to [`git config`](https://git-scm.com/docs/git-config), this command parses repo and cache configs to get, set, and delete values. It only works on local repos so a malicious client can't do something like override their `storage_quota` or reset `append_only`. It's a small addition, but the docs mention editing the config file often enough that I think it makes sense to have a command for it.

I still need to:
- [x] write usage docs
- [x] write unit tests
- [x] change all mentions of manually viewing/editing config files in docs to use `borg config`

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -502,8 +502,7 @@ space for chunks.archive.d (see :issue:`235` for details):
 ::
 
     # this assumes you are working with the same user as the backup.
-    # you can get the REPOID from the "config" file inside the repository.
-    cd ~/.cache/borg/<REPOID>
+    cd ~/.cache/borg/$(borg config /path/to/repo id)
     rm -rf chunks.archive.d ; touch chunks.archive.d
 
 This deletes all the cached archive chunk indexes and replaces the directory
@@ -808,7 +807,7 @@ There are some caveats:
 - If the repository is in "keyfile" encryption mode, the keyfile must
   exist locally or it must be manually moved after performing the upgrade:
 
-  1. Locate the repository ID, contained in the ``config`` file in the repository.
+  1. Get the repository ID with ``borg config /path/to/repo id``.
   2. Locate the attic key file at ``~/.attic/keys/``. The correct key for the
      repository starts with the line ``ATTIC_KEY <repository id>``.
   3. Copy the attic key file to ``~/.config/borg/keys/``
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1,5 +1,6 @@
 import argparse
 import collections
+import configparser
 import faulthandler
 import functools
 import hashlib
@@ -1516,6 +1517,41 @@ class Archiver:
             # any other mechanism relying on existing segment data not changing).
             # see issue #1867.
             repository.commit()
+
+    @with_repository(exclusive=True, cache=True, compatibility=(Manifest.Operation.WRITE,))
+    def do_config(self, args, repository, manifest, key, cache):
+        """get, set, and delete values in a repository or cache config file"""
+        try:
+            section, name = args.name.split('.')
+        except ValueError:
+            section = args.cache and "cache" or "repository"
+            name = args.name
+
+        if args.cache:
+            cache.cache_config.load()
+            config = cache.cache_config._config
+            save = cache.cache_config.save
+        else:
+            config = repository.config
+            save = lambda: repository.save_config(repository.path, repository.config)
+
+        if args.delete:
+            config.remove_option(section, name)
+            if len(config.options(section)) == 0:
+                config.remove_section(section)
+            save()
+        elif args.value:
+            if section not in config.sections():
+                config.add_section(section)
+            config.set(section, name, args.value)
+            save()
+        else:
+            try:
+                print(config.get(section, name))
+            except (configparser.NoOptionError, configparser.NoSectionError) as e:
+                print(e, file=sys.stderr)
+                return EXIT_WARNING
+        return EXIT_SUCCESS
 
     def do_debug_info(self, args):
         """display system information for debugging / bug reports"""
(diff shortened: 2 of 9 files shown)


src/borg/archiver.py
```diff
@@ -1517,6 +1518,41 @@ def do_with_lock(self, args, repository):
...
+            section, name = args.name.split('.')
+        except ValueError:
+            section = args.cache and "cache" or "repository"
+            name = args.name
+
+        if args.cache:
+            cache.cache_config.load()
+            config = cache.cache_config._config
+            save = cache.cache_config.save
+        elif:
```
ThomasWaldmann: syntax error

ThomasWaldmann:
all tests failing, guess this should not be run for "Remote" tests.

also, the help message it emits when it fails looks strange.

milkey-mouse:
I forgot to commit the test skip fix; it was sitting in my git stash :P Let me re-run the tests to make sure it works and I will add a fixup

ThomasWaldmann:
```
src/borg/archiver.py:1536:13: E731 do not assign a lambda expression, use a def
src/borg/testsuite/archiver.py:2750:17: F402 import 'key' from line 60 shadowed by loop variable
```
Stuff from you or just a new flake8?

src/borg/archiver.py
```diff
@@ -1533,7 +1533,9 @@ def do_config(self, args, repository, manifest, key, cache):
             save = cache.cache_config.save
         else:
             config = repository.config
-            save = lambda: repository.save_config(repository.path, repository.config)
+
+            def save():
+                repository.save_config(repository.path, repository.config)
```
ThomasWaldmann: now as i see it (especially when also considering 1533), i like the lambda version more - for consistency.

so, shall we rather disable the flake8 warning?

milkey-mouse: yeah, I prefer the old one as well...

docs/faq.rst
```diff
@@ -808,7 +807,7 @@ There are some caveats:
 - If the repository is in "keyfile" encryption mode, the keyfile must
   exist locally or it must be manually moved after performing the upgrade:
 
-  1. Locate the repository ID, contained in the ``config`` file in the repository.
+  1. Get the repository ID with ``borg config /path/to/repo id``.
```
ThomasWaldmann: hmm, we are dealing with an attic repo here, so guess we rather don't use `borg config` on it.

ThomasWaldmann: hmm, no, it is after upgrade.

src/borg/archiver.py
```diff
@@ -1517,6 +1518,41 @@ def do_with_lock(self, args, repository):
...
             repository.commit()
 
+    @with_repository(exclusive=True, cache=True, compatibility=(Manifest.Operation.WRITE,))
+    def do_config(self, args, repository, manifest, key, cache):
+        """get, set, and delete values in a repository or cache config file"""
+        try:
+            section, name = args.name.split('.')
+        except ValueError:
+            section = args.cache and "cache" or "repository"
+            name = args.name
```
ThomasWaldmann: the above could be a argparse validator and it could be more strict, always requiring section.name to avoid any ambiguities and future problems. it would not need a `--cache` option then, because it is cache.*.

hmm, while thinking about it, isn't it rather configtype:sectionname:key?

configtype = repository or cache
sectionname = [section] from ini-like file

src/borg/archiver.py
```diff
@@ -3469,6 +3505,37 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         subparser.add_argument('args', metavar='ARGS', nargs=argparse.REMAINDER,
                                help='command arguments')
 
+        config_epilog = process_epilog("""
+        This command gets and sets options in a local repository or cache config file.
```
ThomasWaldmann: s/options/configuration values/

src/borg/archiver.py
```diff
@@ -3469,6 +3505,37 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+        For security reasons, this command only works on local repositories.
+
+        To delete a config value entirely, use ``--delete``. To get an existing key, pass
+        only the key name. To set a key, pass both the key name and the new value. Keys
+        can be specified in the format "section.name" or simply "name"; the section will
+        default to "repository" and "cache" for the repo and cache configs, respectively.
```
ThomasWaldmann: maybe just give some examples?

src/borg/archiver.py
```diff
@@ -3469,6 +3505,37 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+
+        By default, borg config manipulates the repository config file. Using ``--cache``
+        edits the repository cache's config file instead.
+        """)
+        subparser = subparsers.add_parser('config', parents=[common_parser], add_help=False,
+                                          description=self.do_config.__doc__,
+                                          epilog=config_epilog,
+                                          formatter_class=argparse.RawDescriptionHelpFormatter,
+                                          help='get and set repository config options')
```
ThomasWaldmann: get and set configuration values

src/borg/archiver.py
```diff
@@ -3469,6 +3505,37 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+        subparser.add_argument('-c', '--cache', dest='cache', action='store_true',
+                               help='get and set values from the repo cache')
+        subparser.add_argument('-d', '--delete', dest='delete', action='store_true',
+                               help='delete the key from the config file')
+
+        subparser.add_argument('location', metavar='REPOSITORY',
+                               type=location_validator(archive=False, proto='file'),
+                               help='repository to configure')
+        subparser.add_argument('name', metavar='NAME',
+                               help='name of config key')
```
ThomasWaldmann: could use type=some_validator here.

src/borg/archiver.py
milkey-mouse: 90% of the stuff in the config file is in the main section (it would be rare that someone would want to edit e.g. `integrity`), it simplifies it to have the default.

src/borg/archiver.py
milkey-mouse: Instead of the description? None of the other commands have examples in the help section.

src/borg/archiver.py
ThomasWaldmann: there's also the issue that section and configtype are not kept separately.

milkey-mouse: The "default" section (the one with the most important keys in it) is`cache` for the cache config and `repository` for the repo config.
Most commands with a similar purpose I know of (`git config`, `npm config`, etc.) use the same syntax as here, with a --flag for the non-default location.

src/borg/helpers/parseformat.py
```diff
@@ -457,6 +457,11 @@ def validator(text):
             raise argparse.ArgumentTypeError('"%s": No archive specified' % text)
         elif archive is False and loc.archive:
             raise argparse.ArgumentTypeError('"%s" No archive can be specified' % text)
+        if proto is not None and loc.proto != proto:
+            if proto == 'file':
+                raise argparse.ArgumentTypeError('"%s" must be a local repository' % (text))
+            else:
+                raise argparse.ArgumentTypeError('"%s" must be a remote repository' % (text))
```
ThomasWaldmann: maybe keep same style as in 457 and 459, to be consistent.

(foo) is pointless anyway as it is not a 1-tuple, for that it would need (foo, ).

ThomasWaldmann:
ok, some minor fixes and collapsing some changesets (maybe collapse all borg-config-related changeset, but keep the pep-checker change in an extra commit) and it's done.

src/borg/archiver.py
ThomasWaldmann: ok, keep it as is.

PR #3336, opened 2017-11-20
Highlight that information is obtained from security dir

aidantwoods:
Deleting the cache will not bypass this error in the event the user knows this is a legitimate repo.

Change points the user in the right direction in the event that they obtain an old copy of a repo and want to resume backups to it.


ThomasWaldmann:
thanks!

aidantwoods:
No problem!

Just dropping a note here also to say thanks to you and the other maintainers for all the great work on this project, much appreciated! 😃

PR #3340, opened 2017-11-20
Mention break-lock in checkpointing FAQ entry

### LINKED ISSUE #3328: FAQ should mention break-lock
http://borgbackup.readthedocs.io/en/stable/faq.html says "If a backup was interrupted, you do not need to do any special considerations, just invoke borg create as you always do. ". This is not (always) true. If locks are still there it'll fail. "borg break-lock" is then needed to be able to continue.

milkey-mouse:
Fixes #3328, also streamlined some of the language for the rest of the paragraph

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -132,8 +132,10 @@ completed is useful because it references all the transmitted chunks up
 to the checkpoint. This means that in case of an interruption, you only need to
 retransfer the data since the last checkpoint.
 
-If a backup was interrupted, you do not need to do any special considerations,
-just invoke ``borg create`` as you always do. You may use the same archive name
+If a backup was interrupted, you normally do not need to do anything special,
+just invoke ``borg create`` as you always do. (If locks are still on a repository,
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -132,8 +132,10 @@ completed is useful because it references all the transmitted chunks up
 to the checkpoint. This means that in case of an interruption, you only need to
 retransfer the data since the last checkpoint.
 
-If a backup was interrupted, you do not need to do any special considerations,
-just invoke ``borg create`` as you always do. You may use the same archive name
+If a backup was interrupted, you normally do not need to do anything special,
+just invoke ``borg create`` as you always do. (If locks are still on a repository,
+such as if the client suffered a sudden poweroff during a backup, you may need to
+run ``borg break-lock`` before the next backup.) You may use the same archive name
```
ThomasWaldmann: How about: "In case the repository is still locked, you may need to ..."

no outer parens. and we also do not need to speculate about the reasons why it is still locked.

PR #3361, opened 2017-11-25
Moved Support to .org (#3360)

MichaelRupertDev:
docs: moved support options to www.borgbackup.org
(cherry picked from commit 71fbac94fbd0b15d3e302c1ed6d967331037bd4e)

PR #3379, opened 2017-11-26
fixed support links

MichaelRupertDev:
Fixed support links as seen in a comment here: https://github.com/borgbackup/borg/pull/3361

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,7 +141,7 @@ Links
 * `Web-Chat (IRC) <http://webchat.freenode.net/?randomnick=1&channels=%23borgbackup&uio=MTY9dHJ1ZSY5PXRydWUa8>`_ and
   `Mailing List <https://mail.python.org/mailman/listinfo/borgbackup>`_
 * `License <https://borgbackup.readthedocs.org/en/stable/authors.html#license>`_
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -141,7 +141,7 @@ Links
 * `Web-Chat (IRC) <http://webchat.freenode.net/?randomnick=1&channels=%23borgbackup&uio=MTY9dHJ1ZSY5PXRydWUa8>`_ and
   `Mailing List <https://mail.python.org/mailman/listinfo/borgbackup>`_
 * `License <https://borgbackup.readthedocs.org/en/stable/authors.html#license>`_
-* `Security contact <https://borgbackup.readthedocs.io/en/latest/support.html#security-contact>`_
+* `Security contact <https://borgbackup.readthedocs.io/en/latest/free.html#security-contact>`_
```
ThomasWaldmann: guess this should stay?
the file in the docs is still named support.html, isn't it?
just the file on the homepage is free.html.

MichaelRupertDev: That is correct. I searched for all usages of support.html, and those are the only ones still around.

ThomasWaldmann: I mean the html docs file name is still support.html (so do not change that to free.html).

ThomasWaldmann:
thanks! hopefully, it's all in sync now. :)

Issue #3390, opened 2017-11-27
micro opt potential, also clarity


ThomasWaldmann:
```
   def build_filter(matcher, peek_and_store_hardlink_masters, strip_components):
        if strip_components:
            def item_filter(item):
                matched = matcher.match(item.path) and os.sep.join(item.path.split(os.sep)[strip_components:])
                peek_and_store_hardlink_masters(item, matched)
                return matched
        else:
            def item_filter(item):
                matched = matcher.match(item.path)
                peek_and_store_hardlink_masters(item, matched)
                return matched
        return item_filter
```
the first matched expression looks like we want a path in matched, but when looking at the second and also at the name "matched", it seems we are rather just interested in a boolean.

then this could be replaced:
```
os.sep.join(item.path.split(os.sep)[strip_components:])
|
v
len(item.path.split(os.sep)) > strip_components
```

ThomasWaldmann:
See #3406 (not merged, closed as there was no activity on it).

ThomasWaldmann:
Here's the diff from the PR in case it goes away:
```
     def build_filter(matcher, peek_and_store_hardlink_masters, strip_components):
         if strip_components:
             def item_filter(item):
-                matched = matcher.match(item.path) and os.sep.join(item.path.split(os.sep)[strip_components:])
+                matched = matcher.match(item.path) and len(item.path.split(os.sep)) > strip_components
                 peek_and_store_hardlink_masters(item, matched)
                 return matched
         else:
```

PR #3406, opened 2017-12-01
Don't reformat path after finding match in item_filter

milkey-mouse:
Implements the micro-opt described in #3390.


ThomasWaldmann:
can you rebase on current master?

ThomasWaldmann:
```
assert not filter(Item(path='shallow/'))  # can this even happen? paths are normalized...
```

Good question. Guess that needs a code review and maybe additional tests.

ThomasWaldmann:
I referred to this via #3390 and #3944 so we do not need to keep this open in pull requests forever.
Once #3944 is done, I guess we could apply this.

PR #3411, opened 2017-12-02
add zstd compression

ThomasWaldmann:
based on willyvmm's work in PR #3116, but some changes:

- removed all multithreading changes, multithreading will be addressed in borg later and differently.
- tests
- fix: minimum level is 1 (not 0)
- default compression level = 3
- only-if-needed bytes() conversion
- other cosmetic fixes
- removed python-zstandard external dependency
- using built-in cython-based wrapper (like we do for lz4 also)
- bundling zstd code into borg, so no new external dependency here either
- if there is a system shared libzstd, it will build against this instead of the bundled zstd code

### DIFF
--- a/src/borg/compress.pyx
+++ b/src/borg/compress.pyx
@@ -22,11 +22,6 @@ try:
 except ImportError:
     lzma = None
 
-try:
-    import zstd
-except ImportError:
-    zstd = None
-
 
 from .helpers import Buffer, DecompressionError
 
@@ -36,6 +31,17 @@ cdef extern from "lz4.h":
     int LZ4_compress_limitedOutput(const char* source, char* dest, int inputSize, int maxOutputSize) nogil
     int LZ4_decompress_safe(const char* source, char* dest, int inputSize, int maxOutputSize) nogil
     int LZ4_compressBound(int inputSize) nogil
+
+
+cdef extern from "algorithms/zstd-libselect.h":
+    size_t ZSTD_compress(void* dst, size_t dstCapacity, const void* src, size_t srcSize, int  compressionLevel) nogil
+    size_t ZSTD_decompress(void* dst, size_t dstCapacity, const void* src, size_t compressedSize) nogil
+    size_t ZSTD_compressBound(size_t srcSize) nogil
+    unsigned long long ZSTD_CONTENTSIZE_UNKNOWN
+    unsigned long long ZSTD_CONTENTSIZE_ERROR
+    unsigned long long ZSTD_getFrameContentSize(const void *src, size_t srcSize) nogil
+    unsigned ZSTD_isError(size_t code) nogil
+    const char* ZSTD_getErrorName(size_t code) nogil
 
 
 buffer = Buffer(bytearray, size=0)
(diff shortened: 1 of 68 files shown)


src/borg/compress.pyx
```diff
@@ -220,6 +226,35 @@ class ZLIB(CompressorBase):
             raise DecompressionError(str(e)) from None
 
 
+class ZSTD(CompressorBase):
+    """zstd compression / decompression (pypi: zstandard, gh: python-zstandard)"""
+    # This is a NOT THREAD SAFE implementation.
+    # Only ONE python context must to be created at a time.
+    # It should work flawlessly as long as borg will call ONLY ONE compression job at time.
+    ID = b'\x10\x00'
```
ThomasWaldmann: taken "as is" from @willyvmm's PR.

next free one would be \x03\x00, so maybe we rather use that.

src/borg/compress.pyx
```diff
@@ -220,6 +226,35 @@ class ZLIB(CompressorBase):
...
+    def compress(self, data):
+        cctx = zstd.ZstdCompressor(level=self.level, write_content_size=True)
+        data = cctx.compress(bytes(data))  # not sure about that typecast but it works.
+        return super().compress(data)
+
+    def decompress(self, data):
+        dctx = zstd.ZstdDecompressor()
+        data = super().decompress(data)
+        try:
+            return dctx.decompress(bytes(data))  # not sure about taht typecast but it works.
```
ThomasWaldmann: typo

src/borg/compress.pyx
```diff
@@ -346,6 +382,16 @@ class CompressionSpec:
             else:
                 raise ValueError
             self.level = level
+        elif self.name in ('zstd', ):
+            if count < 2:
+                level = 5  # default compression level for zstd
```
ThomasWaldmann: taken "as is", but it seems like the default level in python-zstandard is rather 3.

so, take 3 or use a different level better for typical borg usecase?

src/borg/testsuite/compress.py
```diff
@@ -104,6 +119,12 @@ def test_compressor():
             dict(name='lzma', level=6),
             # we do not test lzma on level 9 because of the huge memory needs
         ]
+    if zstd:
+        params_list += [
+            dict(name='zstd', level=1),
+            dict(name='zstd', level=5),
+            # also avoiding high zstd levels, memory needs unclear
```
ThomasWaldmann: research this.

src/borg/compress.pyx
ThomasWaldmann: and check whether that bytes() is needed.

ThomasWaldmann:
The amount and kind of changes in this PR seem relatively safe, so maybe we can consider this even for 1.1-maint (discuss it with package maintainers first).

But even if some dist does not / can not have python-zstandard for some reason, this is on the same level somehow as not having lzma in python. In both cases, borg will just say it's missing, but everything else will work.

ThomasWaldmann:
https://github.com/indygreg/python-zstandard/blob/master/NEWS.rst some changes ahead in 0.9.

Maybe we should just vendor zstandard (>= 0.9 when it is ready) as we can't expect that to be in dists (except some rolling release dists).

src/borg/compress.pyx
ThomasWaldmann: see zstandard "NEWS" @ 0.9.0.

enkore:
What's the rationale for using (thus, requiring) a Python wrapper if you're calling it from C[ython] anyway?

ThomasWaldmann:
"Getting it working quickly" :-D

I am aware that we interfaced to the other stuff more directly from cython.
We can also do it for this, but it'll require some more work.

OTOH, if we could get rid of the python wrapper requirement and its not-yet-fully-stable api.

ThomasWaldmann:
Hmm, in fact it is only more direct for lz4. ZLIB and LZMA call the python wrapper in python stdlib.

ThomasWaldmann:
OK, did some fix ups.

I'ld work on the low-level stuff after discussing whether / how it can be adopted best into 1.1-maint and linux distributions.

Like @enkore pointed out, adapting to zstd C code via cython would remove the `zstandard` python wrapper requirement.

If zstd support can go into 1.1-maint, we would have to vendor (bundle) zstd code as we can not expect currently stable linux (or other) distributions to have it / have it as a recent enough library. Also the package maintainer would not need to change the external package requirements.

Additionally (maybe later), we could offer building against a external zstd also.

ThomasWaldmann:
@LocutusOfBorg ^

ThomasWaldmann:
@bpereto ^

LocutusOfBorg:
This?
https://pypi.python.org/pypi/zstandard 
or this one?
https://pypi.python.org/pypi/zstd

ThomasWaldmann:
@LocutusOfBorg: yes (current state of this PR), but this likely will change to the Cython based wrapper.

so zstd either comes from external libzstd or from bundled zstd code. the cython based wrapper will be part of borg, that's just a few lines of Cython.

the code changes are mostly separate from existing stuff, so little potential for regressions (i mean for stable dists which package 1.1, if any).

problem is a bit that zstd tends to be a bit outdated in released dists, but I'ld rather want something recent, like 1.3.2. so guess I'll need to bundle the zstd C code with borg and offer an option so it can optionally be built against external libzstd also, if the version requirement can be met (like it is already done for libb2 / blake2 code).

(updated from IRC discussion, to have stuff in 1 place)

ThomasWaldmann:
I pushed the code so it can use external libzstd.

No builtin zstd right now, so build will fail if there is no external AND recent libzstd(-dev).

LocutusOfBorg:
nice, thanks! less bindings and wrappers is nice

ThomasWaldmann:
@enkore ^ see recent changes.

FabioPedretti:
Just curious, why didn't you bundle 1.3.2 rather than 1.3.1?

ThomasWaldmann:
2 reasons:
- python-zstandard (where i took the initial version of setup_zstd.py from) bundles 1.3.1 (and iirc the .c/.h files are others/more in 1.3.2)
- ubuntu 17.10 has 1.3.1 in the distribution, so maybe it is good if we test it with that.

I'll look a bit more at the differences between 1.3.1 and 1.3.2 later, maybe I'll update the bundled stuff to 1.3.2.

ThomasWaldmann:
- rebased on current master
- added zstd license
- collapsed some changesets

enkore:
> If zstd support can go into 1.1-maint, we would have to vendor (bundle) zstd code as we can not expect currently stable linux (or other) distributions to have it / have it as a recent enough library.

I can sort of see the point here, though it should be considered that older distros that won't have zstd would seem unlikely to adapt borg or upgrade their packages to 1.1 from 1.0. I would suggest bundling this quite large library only if really necessary and not on a "if-maybe" premise (I don't know how far you looked into this).

Though I would say that it should be avoided to add a new, mandatory dependency in 1.1.x. Maybe only bundle it there?

ThomasWaldmann:
@enkore I don't expect stable/released dists upgrading from 1.0 to 1.1, nor would I suggest/recommend that.

So this is only for stable/released dists which already include some 1.1.x version.

I agree that we shouldn't keep the bundled zstd in borg long-term (nor any other bigger stuff that can be easily provided by an external library), but currently 1.3.x is too new, so we should keep it for now (so developers and testers can run master code without stumbling over zstd).

When we are approaching 1.2 RC we should evaluate zstd 1.3.x shared lib availability and remove it when it is widely available (same for blake2).

ThomasWaldmann:
Some users will also build 1.1.x from source on a dist that does not have zstd-1.3.x yet.
So bundling it with 1.1 will just let them do that without the complication of having to first provide a locally/custom install of a recent libzstd.

ThomasWaldmann:
I'll do some practical tests of the code now and will merge the stuff soon if the outcome is positive.

ThomasWaldmann:
Practical tests successful (pics, .tox, a VM).  2 small new changesets with other stuff I found.

If noone is holding me back, will collapse & merge this later today.

ThomasWaldmann:
collapsed & rebased.

PR #3429, opened 2017-12-11
Add placeholder for fqdn in reverse notation

pngwjpgh:
This allows for having lexicographic sorting coincide with domain structure without having to hardcode fqdns in automated scripts.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -180,10 +180,11 @@ def format_line(format, data):
 def replace_placeholders(text):
     """Replace placeholders in text with their values."""
     current_time = datetime.now()
+    fqdn = socket.getfqdn()
     data = {
         'pid': os.getpid(),
-        'fqdn': socket.getfqdn(),
-        'reverse-fqdn': '.'.join(reversed(socket.getfqdn().split('.'))),
+        'fqdn': fqdn,
+        'reverse-fqdn': '.'.join(reversed(fqdn.split('.'))),
         'hostname': socket.gethostname(),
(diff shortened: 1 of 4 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -183,6 +183,7 @@ def replace_placeholders(text):
     data = {
         'pid': os.getpid(),
         'fqdn': socket.getfqdn(),
+        'reverse-fqdn': '.'.join(reversed(socket.getfqdn().split('.'))),
```
ThomasWaldmann: as getfqdn can take rather long, please refactor this so it gets only called once and the result is used for both placeholders.

pngwjpgh: I'll push a commit to call `getfqdn()` once per invocation of `replace_placeholders()` – if we were going to cache things more globally (i.e. across invocations of `replace_placeholders()`) we should probably do the same for `getpid()`, `gethostname()`, and `getuid()`, shouldn't we?

pngwjpgh:
I pushed a commit to call `getfqdn()` once per invocation of `replace_placeholders()` – if we were going to cache things more globally (i.e. across invocations of `replace_placeholders()`) we should probably do the same for `getpid()`, `gethostname()`, and `getuid()`, shouldn't we?

ThomasWaldmann:
maybe rather don't touch pid / uid / gid stuff.

but hostname / fqdn looks like good "do it just once" candidates (for a separate PR).

ThomasWaldmann:
thanks!

PR #3437, opened 2017-12-13
Document interaction of ``--timestamp`` and time placeholders

pngwjpgh:
I was surprised to see that placeholders do not honor ``--timestamp``.

At least documenting that interaction seems like a good idea.

---

Is this expected behavior?


ThomasWaldmann:
guess this should be rather a bugreport + a fix for that.

pngwjpgh:
Document it now and fix it whenever someone finds the time?

ThomasWaldmann:
There is no code implementing the feature "take --timestamp time for placeholders", so it's a missing (but somehow expected) feature, not a bug. We do not need to document which features we do not have.

pngwjpgh:
Okay :-)

Issue #3415, opened 2017-12-08
use same setup / extension-building mechanism as for zstd


ThomasWaldmann:
when I integrated zstd in #3411, I modified the code to be a bit more modular:
- have a setup_zstd.py with:
  - the prefix-discovery code to find a system-provided library and headers
  - kwargs-amending code to build the arguments for the Extension() initialization
  - having lists of relative pathes for bundled source code
- call code from setup_zstd.py from setup.py
- have a boolean to easily choose whether to prefer the system library over the bundled code (or not)

we could do the same or similar for:
- blake2 (less modular code currently)
- lz4 (no bundled code currently)

ThomasWaldmann:
done.

PR #3443, opened 2017-12-15
compress ext building (lz4 part)

ThomasWaldmann:
Bundle lz4 and do it like the zstd part, for code consistency. See #3415.

We can finally get rid of the deprecation warning and raise lz4 requirement to >= 1.7.0 / r129.

readthedocs build: checked source, should be still fine.


ThomasWaldmann:
@enkore will merge this soon and continue with libb2 (same thing).

Issue #3444, opened 2017-12-15
borg check --repair runs very slow (and corrupts archives)


ThomasWaldmann:
from @bebehei on IRC (IRC log shortened):

borg 1.1.3 on client and server, running borg check --repair on a corrupt repository hangs for >5h after emitting a msg about missing metadata.

```
23:11  bebehei$ Now it happened: Oh, the metadata chunk is missing.
23:12  bebehei$ strace output slowed down immediately
23:16  bebehei$ item metadata chunk missing [chunk: 001056_bdee87d...a3e50d]
23:16  bebehei$ But the borg serve process is dead.
23:22  ThomasWaldmann$ archive.py:1497 is the place where that msg gets emitted
```

Bounty: https://www.bountysource.com/issues/52774936-borg-check-repair-hangs-infinitely

ThomasWaldmann:
The bounty is for the bug being found and fixed.

To prove that it is this bug and it is fixed, maybe a unit test is needed - reproducing the bug first (hanging) and then succeeding after the fix.

ThomasWaldmann:
Hmm, I suspect this bug is even worse.

As the "hang" is not really a infinite hang (but just borg being very slow) and the bug in the validator basically lets borg check "think" every item (even good ones) is invalid once it had to switch into resync-mode, it will destroy the rest of the archive after a missing chunk in the archive metadata stream. :-(

So, if you don't kill borg check (and thus, abort the transaction), but wait patiently for borg check to finish (and the transaction gets committed), all affected archives will be severely damaged.

ThomasWaldmann:
Fix is coming, see #3445.

PR #3445, opened 2017-12-16
fix for borg check --repair malfunction, #3444 (1.1-maint)

ThomasWaldmann:
due to a malfunctioning validator, resyncing to the next item metadata dict in the archive metadata stream after a lost / zero-replaced chunk did not work.

this is a borg 1.1 regression, code in 1.0 is fine.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1457,7 +1457,7 @@ class ArchiveChecker:
             """
             item_keys = frozenset(key.encode() for key in self.manifest.item_keys)
             required_item_keys = frozenset(key.encode() for key in REQUIRED_ITEM_KEYS)
-            unpacker = RobustUnpacker(lambda item: isinstance(item, dict) and 'path' in item,
+            unpacker = RobustUnpacker(lambda item: isinstance(item, StableDict) and b'path' in item,
                                       self.manifest.item_keys)
             _state = 0
 
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -3006,7 +3006,7 @@ class ArchiverCheckTestCase(ArchiverTestCaseBase):
     def test_missing_archive_item_chunk(self):
         archive, repository = self.open_archive('archive1')
         with repository:
-            repository.delete(archive.metadata.items[-5])
+            repository.delete(archive.metadata.items[0])
             repository.commit()
         self.cmd('check', self.repository_location, exit_code=1)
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
Some practical tests were done on @bebehei's defect repository.

With the fix, borg was able to repair the repo and borg check / list / extract afterwards worked as expected.

aradis:
The release notes about this bug don't explain what exactly the effect is. If we experienced this bug when doing a `borg check --repair` (as I did), what precisely was corrupted? A single archive, many archives, the entire repository? If only specific archives, is there a way to identify which ones?

The notes also don't say what actions an affected user should take after upgrading to 1.1.4. Will running `borg check --repair` with 1.1.4 fix the damage that the bug did? If not, what steps can be taken to minimize the damage done and continue to use the repository and as many archives as possible?

ThomasWaldmann:
@aradis The trigger for this bug is a defect/missing chunk in the archive items metadata stream (where all metadata for fs items like files/dirs/... is stored. name, timestamps, user/group/mode, ...). Depending how many such defect/missing chunks and where precisely they are located, none, a few or many archives can be affected. Missing file content data chunks won't trigger the bug.

If you ran borg check --repair (with a borg 1.1.x version before the fix) on a repo with such missing metadata chunks (meaning the repo already had some data loss / corruption issues before you ran borg check --repair), borg would not manage to resynchronize onto the metadata stream after the missing chunk and thus consider all remaining metadata in the affected archive as invalid and discard that metadata. Potentially large parts of that archive would be effectively gone (you could see that in the archive content listing, a lot of stuff missing that should be there).

Luckily, the bug and not being able to successfully resynchronize onto the metadata stream also meant that borg check --repair was very slow in such cases. The bug report we got about this was not about borg eating data, but borg being super slow / "hanging", so the user who found this bug interrupted it. If one did not let it finish, one had no data loss due to this. If one let it finish, the affected archive metadata is irrecoverably gone, affected archives will be incomplete afterwards. borg check --repair does not help with that.

What you can do is borg list repo::archive for all your archives (in repos that had borg check --repair runs with unfixed 1.1.x borg AND missing/corrupted chunks) and "look" at the output - are all expected dirs/files there? or does the list prematurely end, missing (potentially lots of) stuff that should be in there? then decide if you want to rather discard incomplete archives or mark them somehow.

Other data structures in the repo / unaffected archives will be unaffected / fine by this bug.

PR #3450, opened 2017-12-16
also delete security dir when deleting a repo, fixes #3427

### LINKED ISSUE #3427: borg delete <repo> doesn’t delete BORG_SECURITY_DIR/<repodir>
I just started using borgbackup and tested a few things with throwaway repositories. I noticed that deleting a repository (using `borg delete`) does not delete its subdirectory below BORG_SECURITY_DIR. Is that intentional?

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -1108,6 +1108,7 @@ def _delete_repository(self, args, repository):
                 return self.exit_code
             repository.destroy()
             logger.info("Repository deleted.")
+            SecurityManager.destroy(repository)
```
ThomasWaldmann: like this, the security dir is only killed when the repository is removed.

outdenting this one level would also kill the security dir when using `--cache-only` - do we want that?

PR #3467, opened 2017-12-21
added BORG_BASE_DIR, s/get_home_dir/get_base_dir/g, docs, fixes #3338

### LINKED ISSUE #3338: environment variable to set borg base location
In some use cases it is necessary to tell borg to use a home location that is not the user $HOME. Actually I do that do start all borg commands with modificated $HOME variable. But that feature should be included into borg! That makes it also possible to set another variables relativ to it and leads to cleaner code.

So, a variable like **BORG_BASE_DIR** or **BORG_HOME_DIR** should be introduced. e.g
```bash
# new shell variable, default value is $HOME
BORG_BASE_DIR=/yxz/abc

# possible default values of offical variables, if they are not set explicitly
BORG_CONFIG_DIR=$BORG_BASE_DIR/.config/borg
BORG_CACHE_DIR=$BORG_BASE_DIR/.cache/borg
BORG_KEYS_DIR=$BORG_CONFIG_DIR/keys
BORG_SECURITY_DIR=$BORG_CONFIG_DIR/security
# ...
```

ThomasWaldmann:
also: sorted env var docs by level

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -203,19 +203,22 @@ Some automatic "answerers" (if set, they automatically answer confirmation quest
     allowed). So please test your scripts interactively before making them a non-interactive script.
 
 Directories and files:
-    BORG_KEYS_DIR
-        Default to '~/.config/borg/keys'. This directory contains keys for encrypted repositories.
-    BORG_KEY_FILE
-        When set, use the given filename as repository key file.
+    BORG_BASE_DIR
+        Default to '$HOME', '~$USER', '~' (in that order)'.
+        If we refer to ~ below, we in fact mean BORG_BASE_DIR.
+    BORG_CONFIG_DIR
+        Default to '~/.config/borg'. This directory contains the whole config directories.
+    BORG_CACHE_DIR
+        Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
+        of space for dealing with big repositories.
     BORG_SECURITY_DIR
         Default to '~/.config/borg/security'. This directory contains information borg uses to
         track its usage of NONCES ("numbers used once" - usually in encryption context) and other
         security relevant data.
-    BORG_CACHE_DIR
-        Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
-        of space for dealing with big repositories).
-    BORG_CONFIG_DIR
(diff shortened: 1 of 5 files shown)


docs/usage_general.rst.inc
```diff
@@ -203,19 +203,22 @@ Some automatic "answerers" (if set, they automatically answer confirmation quest
...
-    BORG_KEY_FILE
-        When set, use the given filename as repository key file.
+    BORG_BASE_DIR
+        Default to '$HOME', '~$USER', '~' (in that order)'.
+        If we refer to ~ below, we in fact mean BORG_BASE_DIR.
+    BORG_CONFIG_DIR
+        Default to '~/.config/borg'. This directory contains the whole config directories.
+    BORG_CACHE_DIR
+        Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
+        of space for dealing with big repositories).
```
witten: Predated your changeset, but the ")" appears to be a stray.

src/borg/helpers/fs.py
```diff
@@ -12,18 +12,22 @@
...
     # parameters: when called with an explicit username, the possibly set
     # environment variable HOME is no longer respected. So we have to check if
     # it is set and only expand the user's home directory if HOME is unset.
-    if os.environ.get('HOME', ''):
-        return os.environ.get('HOME')
-    else:
-        return os.path.expanduser('~%s' % os.environ.get('USER', ''))
+    if not base_dir:
+        base_dir = os.path.expanduser('~%s' % os.environ.get('USER', ''))
+    return base_dir
```
witten: Might be nice to have some test coverage for this function, to go with the other `get_*_dir()` tests: https://github.com/borgbackup/borg/blob/master/src/borg/testsuite/helpers.py#L20 .. You could assert that given a particular `os.environ` dict, the returned base dir is as expected.

docs/usage_general.rst.inc
ThomasWaldmann: good catch, fixed.

src/borg/helpers/fs.py
ThomasWaldmann: done.

ThomasWaldmann:
added tests, fixed typo, rebased on current master.

PR #3473, opened 2017-12-22
Properly quote repo path in `borg init` output

anowlcalledjosh:
This is a really unimportant issue, but when the message about the manifest spoofing check is printed by `borg init`, if your repository path has a quote in it, then the command Borg prints won't be valid. This fixes that by passing the path through [`shlex.quote`].

[`shlex.quote`]: https://docs.python.org/3/library/shlex.html#shlex.quote

example: this invalid command

    borg upgrade --disable-tam '/path/test'repo'

is now printed as

    borg upgrade --disable-tam '/path/test'"'"'repo'


ThomasWaldmann:
thanks!

PR #3475, opened 2017-12-22
borg init: use given --storage-quota for local repo, fixes #3470

### LINKED ISSUE #3470: `borg init --storage-quota` not working properly?
I ran `borg init --encryption keyfile --storage-quota 300G --progress --show-version --show-rc /repo`, but the `/repo/config` file has `storage_quota = 0`. Is this expected? If not, what format should the number in the config file be in? (looks like bytes from reading the source, but I'd like to check)

ThomasWaldmann:
writing the storage_quota into the repo config at `borg init` time worked for remote repos, but not for local ones.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -124,6 +124,7 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
         def wrapper(self, args, **kwargs):
             location = args.location  # note: 'location' must be always present in args
             append_only = getattr(args, 'append_only', False)
+            storage_quota = getattr(args, 'storage_quota', None)
             if argument(args, fake) ^ invert_fake:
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -124,15 +124,16 @@ def decorator(method):
         def wrapper(self, args, **kwargs):
             location = args.location  # note: 'location' must be always present in args
             append_only = getattr(args, 'append_only', False)
+            storage_quota = getattr(args, 'storage_quota', None)
```
ThomasWaldmann: getattr must be used here, because not every borg command has the `--storage-quota` argument.

anowlcalledjosh:
Looks fine to me :)

PR #3485, opened 2017-12-24
Correct some confusing error messages from `borg init`

### LINKED ISSUE #3465: Confusing error message from borg init
If user doesn't have access to write to `/path`, this happens:
```
$ borg init -e keyfile /path
A repository already exists at /path.
```
borg 1.1.3

anowlcalledjosh:
Fixes #3465.

Examples:

```
$ borg init -e none /tmp/borg-test/
Files not created by Borg already exist at /tmp/borg-test.
$ borg init -e none /tmp/borg-test/foo
/tmp/borg-test/foo is not a directory.
```

whereas previously this would have happened:

```
$ borg init -e none /tmp/borg-test/
A repository already exists at /tmp/borg-test.
$ borg init -e none /tmp/borg-test/foo 
A repository already exists at /tmp/borg-test/foo.
```

Notes:

- this presumably needs tests, I have no idea how to even start doing that
- I'm open to suggestions on the exact wording of the new error messages

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -195,6 +195,21 @@ class Repository:
     def id_str(self):
         return bin_to_hex(self.id)
 
+    @staticmethod
+    def is_repository(path):
+        """Check whether there is already a Borg repository at *path*."""
+        try:
+            # Use binary mode to avoid troubles if a README contains some stuff not in our locale
+            with open(os.path.join(path, 'README'), 'rb') as fd:
+                # Read only the first ~100 bytes (if any), in case some README file we stumble upon is large.
+                readme_head = fd.read(100)
+                # The first comparison captures our current variant (REPOSITORY_README), the second comparison
+                # is an older variant of the README file (used by 1.0.x).
+                return b'Borg Backup repository' in readme_head or b'Borg repository' in readme_head
+        except OSError:
+            # Ignore FileNotFound, PermissionError, ...
+            return False
+
     def check_can_create_repository(self, path):
         """
         Raise self.AlreadyExists if a repository already exists at *path* or any parent directory.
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
> Files not created by Borg already exist at /tmp/borg-test.

Does it matter who/what created it? Borg should reject any stuff at the target location.

> /tmp/borg-test/foo is not a directory.

That somehow suggests it would be fine if it was a directory.

But what we really expect and usually have is just nothing at the target location, so borg can create its repo directory there.

src/borg/repository.py
```diff
@@ -207,8 +228,13 @@ def check_can_create_repository(self, path):
...
         """
-        if os.path.exists(path) and (not os.path.isdir(path) or os.listdir(path)):
-            raise self.AlreadyExists(path)
+        if os.path.exists(path):
+            if not os.path.isdir(path):
+                raise self.NotADirectory(path)
+            elif self.is_repository(path):
+                raise self.AlreadyExists(path)
+            elif os.listdir(path):
+                raise self.OtherFilesExist(path)
```
ThomasWaldmann: strictly taken, you don't need "ELif" if you raise in every block.

src/borg/repository.py
```diff
@@ -218,18 +244,8 @@ def check_can_create_repository(self, path):
...
-                    readme_head = fd.read(100)
-                    # The first comparison captures our current variant (REPOSITORY_README), the second comparison
-                    # is an older variant of the README file (used by 1.0.x).
-                    if b'Borg Backup repository' in readme_head or b'Borg repository' in readme_head:
-                        raise self.AlreadyExists(path)
-            except OSError:
-                # Ignore FileNotFound, PermissionError, ...
-                pass
+            if self.is_repository(path):
+                raise self.AlreadyExists(path)
```
ThomasWaldmann: good refactor.

ThomasWaldmann:
See the messages I added in my comment above, guess they cover all cases of "something there".

We do not need to point out that borg can cope with an empty dir, I've seen some scripts that do mkdir first, but that is pointless as borg does that anyway.

anowlcalledjosh:
Ok, sounds good, I will push those changes once the tests have run :)

before I do that, it seems a bit weird to have an `OtherFilesExist` exception on the `Repository` class, since it doesn't really have anything to do with the repository. Is there a better place it should go?

src/borg/repository.py
anowlcalledjosh: 👍 changed to plain `if`s

ThomasWaldmann:
thanks, looks good! i collapsed the changesets and merged it.
can you do the backport PR against 1.1-maint?

PR #3503, opened 2018-01-01
Happy new year!

Abogical:
Should this step be automated?

### DIFF
--- a/src/borg/shellpattern.py
+++ b/src/borg/shellpattern.py
@@ -15,7 +15,7 @@ def translate(pat, match_end=r"\Z"):
 
     This function is derived from the "fnmatch" module distributed with the Python standard library.
 
-    Copyright (C) 2001-2017 Python Software Foundation. All rights reserved.
+    Copyright (C) 2001-2016 Python Software Foundation. All rights reserved.
 
     TODO: support {alt1,alt2} shell-style alternatives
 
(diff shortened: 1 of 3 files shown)


Partheo:
Happy new year!

src/borg/shellpattern.py
```diff
@@ -15,7 +15,7 @@ def translate(pat, match_end=r"\Z"):
 
     This function is derived from the "fnmatch" module distributed with the Python standard library.
 
-    Copyright (C) 2001-2017 Python Software Foundation. All rights reserved.
+    Copyright (C) 2001-2018 Python Software Foundation. All rights reserved.
```
ThomasWaldmann: maybe we should not update that one, it was copied in 2017 from the py stdlib.

Abogical: Actually it was copied back in 2016 at #1989, but I modified that line last year for similar reasons. I'm sorry for this. Should we change it back to 2016?

ThomasWaldmann: yes, maybe change it back.

Abogical:
Reverted `shellpattern.sh` change.

ThomasWaldmann:
@Abogical can you do same against 1.1-maint and 1.0-maint?

PR #3510, opened 2018-01-04
borg shell completion: remove shebang, useless because scripts are so…

LocutusOfBorg:
remove shebang, to fix lintian issue:
https://lintian.debian.org/tags/script-not-executable.html


SanskritFritz:
I approve this change, completion works without the shebang.

ThomasWaldmann:
thanks! same change needed for 1.1-maint also.

PR #3518, opened 2018-01-09
DONTMERGE: give use_bin_type=False, see #3517

### LINKED ISSUE #3517: FutureWarning about use_bin_type from msgpack 0.5
https://travis-ci.org/borgbackup/borg/jobs/326101482

see the complaints about use_bin_type.

ThomasWaldmann:
since 0.5.0, msgpack always wants use_bin_type.

if not given, it emits a FutureWarning about the intention to change the default value later.

Note: code is getting uglier.

I am thinking about introducing a wrapper around msgpack to deal with such stuff. But likely, that will cost a bit of performance, not sure how much.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -234,7 +234,7 @@ class ChunkBuffer:
 
     def __init__(self, key, chunker_params=ITEMS_CHUNKER_PARAMS):
         self.buffer = BytesIO()
-        self.packer = msgpack.Packer(unicode_errors='surrogateescape')
+        self.packer = msgpack.Packer(unicode_errors='surrogateescape', use_bin_type=False)
         self.chunks = []
         self.key = key
         self.chunker = Chunker(self.key.chunk_seed, *chunker_params)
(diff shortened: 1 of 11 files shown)


ThomasWaldmann:
an alternative approach could be just requiring msgpack-python >=0.4.6, <0.5.0 - especially as more api changes are planned every 0.1 step.

ThomasWaldmann:
the problem with the FutureWarning of msgpack 0.5.0 was "fixed" by forbidding msgpack 0.5.0.
0.5.1 does not have the warning any more.

but I'ld still like to discuss the idea of a wrapper, for better consistency and a central place to deal with future msgpack changes. @enkore, what do you think?

enkore:
I don't know about that but I think some code in borg.fuse would be able to use msgpack 0.5 features.

ThomasWaldmann:
@enkore yes, I've seen that. It is just unfortunate that msgpack dropped py34 in 0.5+ and we still support it in borg 1.1, so we can't do that in 1.1.

ThomasWaldmann:
closing this, I'll do a PR with a less ugly solution soon.

PR #3520, opened 2018-01-09
Chunk seed is generated per /repository/

pngwjpgh:
Storage in the keyfile implies to me that the chunk seed is unique per /repository/ not archive; otherwise deduplication would hardly work.

Is this correct?


ThomasWaldmann:
yes, of course "once per repo" was meant. good catch! :)

PR #3521, opened 2018-01-09
require msgpack >= 0.4.6 and < 0.6.0, exclude 0.5.0

ThomasWaldmann:
maybe this is the easiest way for us to deal with msgpack compatibility.


ThomasWaldmann:
backports need to be: `>=0.4.6,<0.5.0` because py34 was dropped at 0.5.0.

PR #3526, opened 2018-01-10
sec docs: explicitly note what happens OUTSIDE the attack model

enkore:
I thought I included this from the start, but obviously I didn't.

### DIFF
--- a/docs/internals/security.rst
+++ b/docs/internals/security.rst
@@ -37,6 +37,10 @@ Under these circumstances Borg guarantees that the attacker cannot
 The attacker can always impose a denial of service per definition (he could
 forbid connections to the repository, or delete it entirely).
 
+When the above attack model is extended to include multiple clients
+independently updating the same repository, then Borg fails to provide
+confidentiality (i.e. guarantees 3) and 4) do not apply any more).
+
 .. _security_structural_auth:
 
 Structural Authentication


docs/internals/security.rst
```diff
@@ -37,6 +37,9 @@ Under these circumstances Borg guarantees that the attacker cannot
 The attacker can always impose a denial of service per definition (he could
 forbid connections to the repository, or delete it entirely).
 
+In an attack model where multiple clients independently access the same repository,
+Borg fails to provide confidentiality (i.e. guarantees 3) and 4) are essentially void).
```
ThomasWaldmann: In an extended attack model (same as above, but with multiple clients independently updating the same repository), ...

docs/internals/security.rst
```diff
@@ -37,6 +37,10 @@ Under these circumstances Borg guarantees that the attacker cannot
 The attacker can always impose a denial of service per definition (he could
 forbid connections to the repository, or delete it entirely).
 
+When the above attack model is extended to include multiple clients
+independently updating the same repository, then Borg fails to provide
+confidentiality (i.e. guarantees 3) and 4) do not apply any more).
+
```
ThomasWaldmann: Hmm, maybe we should add a note to the normal user docs also. Who reads internals docs anyway? :)

ThomasWaldmann: @enkore did you see my last comment?

enkore: I didn't, but I don't feel like puzzling over were to put that in the user docs, either.

PR #3546, opened 2018-01-15
use timezone-aware datetime objects

BlessJah:
Calls to `datetime.utcnow()` were replaced with `datetime.now(timezone.utc)`.
The `utcnow()` return naive datetime object, one unaware of timezones. The
`datetime.now(timezone.utc)` results in exact same datetime value, but
object aware of timezones[1].

The change should  allow use of `%z` and `%Z` in `{utcnow}` format string[2].

Minor change: `current_time` was renamed to `frozen_datetime`.

1. https://docs.python.org/3.6/library/datetime.html#strftime-and-strptime-behavior
2. https://docs.python.org/3.6/library/datetime.html#datetime.datetime.utcnow

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -180,14 +180,14 @@ def format_line(format, data):
 def replace_placeholders(text):
     """Replace placeholders in text with their values."""
     from ..platform import fqdn
-    current_time = datetime.now()
+    frozen_datetime = datetime.now()
     data = {
         'pid': os.getpid(),
         'fqdn': fqdn,
         'reverse-fqdn': '.'.join(reversed(fqdn.split('.'))),
         'hostname': socket.gethostname(),
-        'now': DatetimeWrapper(current_time.now()),
-        'utcnow': DatetimeWrapper(current_time.utcnow()),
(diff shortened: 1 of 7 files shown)


BlessJah:
I've renamed `current_time` to `frozen_datetime` because the object ceases to be *current* next moment after it is created. Furthermore new name is grep-friendly for those seeking all `datetime` uses.

src/borg/helpers/parseformat.py
```diff
@@ -180,14 +180,14 @@ def format_line(format, data):
 def replace_placeholders(text):
     """Replace placeholders in text with their values."""
     from ..platform import fqdn
-    current_time = datetime.now()
+    frozen_datetime = datetime.now()
```
ThomasWaldmann: why did you rename this?

BlessJah: See my previous comment.

BlessJah:
It seems that this patch needs more work. Indeed my version mixes `naive` and `aware` objects which may rise issues in future. It seems that code uses `naive` objects.

ThomasWaldmann:
do you plan to finish this or shall we close it?

ThomasWaldmann:
closing this for now, if you still work on it, please do a new PR.

Issue #3431, opened 2017-12-13
create --progress --stats is defect in master branch


ThomasWaldmann:
```
0 B O 0 B C 0 B D 0 N home/tw/Downloads/gluon-ffs-1.0+2017-04-10-g.84dd03a-s.5cab37c-tp-link-tl-wr841n-nd-v9-sysupgrade.bin            
```

(that was NOT the first file, it's always 0 0 0 0)

Maiestas28:
Hey, Im new to Python and this is my first time on Git. Can you guide me through what exactly is required here ? If this is too advanced you can let me know, I'l look for something else.

ThomasWaldmann:
@Maiestas28 maybe first get more familiar with git and python (and python debugging), there are tutorials for both out there (and github also has good help IIRC).

Maiestas28:
Got it. Thank you.

Gu1nness:
Hi,
Can I pick this one ?

ThomasWaldmann:
sure.

PR #3557, opened 2018-01-18
Fix --progress option.

Gu1nness:
When called with --progress, the progress was always showing
0 B O 0 B C 0 B D 0 N.

This is fixed by this commit.

Fixes https://github.com/borgbackup/borg/issues/3431

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -443,7 +443,7 @@ Utilization of max. archive size: {csize_max:.0%}
     def add_item(self, item, show_progress=True, stats=None):
         if show_progress and self.show_progress:
             if stats is None:
-                stats= self.stats
+                stats = self.stats
             stats.show_progress(item=item, dt=0.2)
         self.items_buffer.add(item)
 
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
https://github.com/borgbackup/borg/pull/3557/files#diff-48f2b9140a35ffd45a46ccfbe63931a6R487

have a look there ^ - this looks wrong now as you use the archive.stats instance all the time already.

maybe do a validity check of the numbers you see in --progress and --stats, whether they are as expected (just compare to borg 1.1).

ThomasWaldmann:
still there?

ThomasWaldmann:
can you rebase on master, so only your changes show up here?

ThomasWaldmann:
@enkore besides the minor problem I have pointed out above, is this the right way to fix?

Gu1nness:
Hi, sorry for the long delay.
I have just pushed a new and patched version.
The tricky thing was about bounded class method (`add_item`)
Hope this patch will be okay for you

enkore:
I guess so

ThomasWaldmann:
@Gu1nness thanks for the PR. it needed a small fix, see #3681.

Gu1nness:
My bad, I had missed this one.
Sorry for the small error.

ThomasWaldmann:
No problem. :)

Issue #3545, opened 2018-01-15
stat() errors in scandir_generic() cause the whole directory to be skipped


leo-b:
When backing up a directory, ``scandir_inorder()`` is used to read the directory entries ordered by inode number:
https://github.com/borgbackup/borg/blob/1.1.4/src/borg/archiver.py#L613-L614
https://github.com/borgbackup/borg/blob/1.1.4/src/borg/helpers.py#L2104-L2105

 If the ``scandir_generic()`` implementation is used (python3.4 without the scandir package installed), the implicit `stat()` that is done on ``dirent.inode()`` will raise `Permission denied` e.g. on fuse mountpoints without `allow_other`. As a result, the whole directory will be skipped just because of one failing stat.
https://github.com/borgbackup/borg/blob/1.1.4/src/borg/helpers.py#L2046-L2090

1.) A failing ``stat()`` should only skip the corresponding directory entry.
2.) If the ``stat()`` fails on a directory entry that is going to be skipped anyway, the error message should be suppressed.

ThomasWaldmann:
Good analysis. Would you like to do a PR to fix that issue?

ThomasWaldmann:
@leo-b sounds good, just do a PR.

ThomasWaldmann:
fixed by #3558.

PR #3558, opened 2018-01-18
ignore exceptions in scandir_inorder() caused by an implicit stat()

leo-b:
Proposed fix for #3545.

### DIFF
--- a/src/borg/helpers.py
+++ b/src/borg/helpers.py
@@ -2101,8 +2101,21 @@ except ImportError:
         scandir = scandir_generic
 
 
+def scandir_keyfunc(dirent):
+    try:
+        return (0, dirent.inode())
+    except OSError as e:
+        # maybe a permission denied error while doing a stat() on the dirent
+        logger.debug('scandir_inorder: Unable to stat %s: %s', dirent.path, e)
+        # order this dirent after all the others lexically by file name
+        # we may not break the whole scandir just because of an exception in one dirent
+        # ignore the exception for now, since another stat will be done later anyways
+        # (or the entry will be skipped by an exclude pattern)
+        return (1, dirent.name)
+
+
 def scandir_inorder(path='.'):
-    return sorted(scandir(path), key=lambda dirent: dirent.inode())
+    return sorted(scandir(path), key=scandir_keyfunc)
 
 
 def clean_lines(lines, lstrip=None, rstrip=None, remove_empty=True, remove_comments=True):


leo-b:
Yes. I tested it with Python 3.4 and the borg implementation of scandir as root using a directory containing a fuse mountpoint.
The debug message is output when the mount-point is hit by scandir_keyfunc(). The directory contents are correctly returned and processed. If the fuse mountpoint isn't excluded, an error is logged later when it is hit by stat() in Archiver._process().
If the mountpoint _is_ excluded by a pattern, only the debug message in the key-function is logged.
Thus it looks good from my point of view.

ThomasWaldmann:
thanks!

iirc, this should be 1.1-only as master is python 3.5+ and 1.0 does not yet have scandir.

can you check / confirm?

leo-b:
> iirc, this should be 1.1-only as master is python 3.5+ and 1.0 does not yet have scandir.

It is unfortunate that borg-1.2 will be python-3.5 only. At our university have many RHEL6 and 7 boxes which will need to be supported and backed up almost up to their EOL (2020-11-30 for EL6 and 2024-06-30). It will be much harder to provide a python-3.5 version of borg as RPM for those distros, as there are much more missing dependencies to handle.
It would certainly be beneficial for a wide dissemination of borg, if it was easily accessible via distribution packages. And for spreading in large institutions, supporting multiple distro versions are especially important.

Python 3.5 makes borgs ``scandir_generic()`` obsolete but still uses ``scandir_inorder()``, which does a ``dirent.inode()`` call. The documentation states:
https://docs.python.org/3/library/os.html#os.scandir
> All os.DirEntry methods may perform a system call, but is_dir() and is_file() usually only require a system call for symbolic links; os.DirEntry.stat() always requires a system call on Unix but only requires one for symbolic links on Windows.

So I guess it is OS and filesystem dependent, whether the ``dirent.inode()`` call might raise an ``OSError``.
Thus I'd recommend to include my patch in master too.

ThomasWaldmann:
We offer the standalone binaries for some of the systems where otherwise the installation would be a bigger effort / pain (they include python and all other needed stuff).

Also, it is very likely that 1.2 development will take quite a while and that 1.1 will be supported/updated in parallel to 1.2 for a while. Also, for such old systems, you can just continue using an older (== not bleeding edge) borg, so you will not really run into a problem.

Python 3.5 has some nice features and also less issues/quirks than 3.4.

Other software is also dropping 3.4 support, like e.g. msgpack just did with their 0.5.0 release.

ThomasWaldmann:
A PR against master would be nice. :)

PR #3569, opened 2018-01-20
ignore exceptions in scandir_inorder() caused by an implicit stat() (against master)

### LINKED ISSUE #3545: stat() errors in scandir_generic() cause the whole directory to be skipped

leo-b:
This is the PR against master that fixes #3545.


leo-b:
I'll do another one...

PR #3580, opened 2018-01-27
do .h file content checks in binary mode, fixes #3544

### LINKED ISSUE #3544: ascii decoding of header files can lead to UnicodeDecodeError
https://github.com/borgbackup/borg/blob/1.1.4/setup_lz4.py#L27 e.g. there, but also some other places for other header files.

https://github.com/openbsd/ports/commit/1232bc2b8d672eb5b4fa1873e0402f2d94da3101

ThomasWaldmann:
we can't know the encoding header file authors will use, so using binary for our simple checks is the safest way.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -149,8 +149,8 @@ def detect_openssl(prefixes):
     for prefix in prefixes:
         filename = os.path.join(prefix, 'include', 'openssl', 'evp.h')
         if os.path.exists(filename):
-            with open(filename, 'r') as fd:
-                if 'PKCS5_PBKDF2_HMAC(' in fd.read():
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
I tagged this for 1.0-maint, TODO:
- check whether it partially applies to 1.0 (I guess the openssl and lz4 header part applies, but not zstd and blake2).
- the checks are all in setup.py

This is to prevent build issues with non-ascii header files and pure-ascii environment on some platforms.

ThomasWaldmann:
@miek ^^^ while you're at it... :-)

Issue #3562, opened 2018-01-19
allow a combination of BORG_LOGGING_CONF and commandline log-level


leo-b:
Currently the commandline options that set the log-level (``--verbose``, ``--debug`` etc.) are ignored, if a logging config is specified via ``BORG_LOGGING_CONF``.

I believe it would make sense to be able to override the log level of the file-based config with the commandline-options.

The easiest way to obtain that would be to change the argparse default to ``None`` and set the level of the root logger in the "config file branch" if level is different from None. (And default to `warning` otherwise.)
https://github.com/borgbackup/borg/blob/1.1.4/src/borg/logger.py#L72

However maybe it would be better to only change the level of the _stream logger_ if a logging config is used? That way one could define a logging config like below, that logs both to a logfile and stderr and the  commandline log-level would only affect the stderr output...

```
[loggers]
keys=root

[logger_root]
handlers=stdout,file
level=NOTSET

[handlers]
keys=file,stdout

[handler_file]
class=FileHandler
level=INFO
formatter=file
args=('/var/log/borgbackup.log',)

[handler_stdout]
class=StreamHandler
formatter=stream
level=NOTSET
args=(sys.stdout,)

[formatters]
keys=stream,file

# name: module, e.g. borg.archiver
[formatter_stream]
format=%(levelname).1s %(message)s

[formatter_file]
format=%(asctime)s %(levelname).1s %(message)s
```

ThomasWaldmann:
Somehow this does not sound like making things simpler.

Maybe it makes one use case simpler, but overall the interaction is more complex (esp. if the commandline logging args only affect SOME loggers, as you described).

ThomasWaldmann:
BTW, if this gets merged and docs updated with the logging config shown above, it should not use `sys.stdout` stream as that obviously mixes up with `borg extract --stdout`.

leo-b:
I've added  a new [PR](#3583) against master.
This one also includes tests for file based logging.

leo-b:
My use case:

I am running borg from a cron job with an ini-file based config. File logging goes to /var/log/borg.log at a fixed level of INFO. On stderr only warnings are output by default to keep the cron-job rather quiet.

However, sometimes more verbose output is needed when doing tests on the commandline. This patch allows to do a test run at stderr level DEBUG while still logging to the file at the "normal" level.

ThomasWaldmann:
[...]

PR #3583, opened 2018-01-28
allow to override the log level of stderr logging

leo-b:
PR against master for #3562

### DIFF
--- a/src/borg/testsuite/logger.py
+++ b/src/borg/testsuite/logger.py
@@ -5,21 +5,8 @@ import os
 import pytest
 
 from ..logger import find_parent_module, create_logger, setup_logging
-from .. import logger as _logger
+from .. import logger as logger_module
 logger = create_logger()
-
-
-def reset_logging():
-    global logger
-    import importlib
-    # otherwise we'd need to reset the root logger and delete all other loggerss
-    importlib.reload(logging)
-    _logger.configured = False
-    logger = create_logger()
-
-
-def teardown_function(function):
-    reset_logging()
 
 
 @pytest.fixture()
@@ -29,6 +16,17 @@ def io_logger():
     handler.setFormatter(logging.Formatter('%(name)s: %(message)s'))
     logger.setLevel(logging.DEBUG)
     return io
+
+
+@pytest.fixture()
+def reset_logging():
+    # this should only be used in tests marked as ownprocess for testing logging setup
+    global logger
+    import importlib
+    # otherwise we'd need to reset the root logger and delete all other loggers
+    importlib.reload(logging)
+    logger_module.configured = False
+    logger = create_logger()
 
 
 @pytest.fixture(scope='function')
@@ -67,7 +65,7 @@ format=%%(levelname)s: %%(message)s
 format=FILE %%(levelname)s: %%(message)s
 """
     logfile = tmpdir.join('borg.log')
-    # touch the file
+    # create an empty logfile
     logfile.open(mode='w')
 
     class context:
@@ -90,7 +88,6 @@ format=FILE %%(levelname)s: %%(message)s
             return open(cls.cfg['logfile_path']).read()
     os.environ['BORG_LOGGING_CONF'] = str(context.logging_config_file)
     yield context
-    del os.environ['BORG_LOGGING_CONF']
 
 
 def test_setup_logging(io_logger):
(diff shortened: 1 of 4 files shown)


src/borg/testsuite/logger.py
```diff
@@ -1,12 +1,27 @@
...
 from ..logger import find_parent_module, create_logger, setup_logging
+from .. import logger as _logger
 logger = create_logger()
 
 
+def reset_logging():
+    global logger
+    import importlib
+    # otherwise we'd need to reset the root logger and delete all other loggerss
+    importlib.reload(logging)
```
ThomasWaldmann: iirc, reload() has all sorts of issues. does it really work here?

typo: loggerss

src/borg/testsuite/logger.py
```diff
@@ -16,11 +31,120 @@ def io_logger():
...
+# name: module, e.g. borg.archiver
+[formatter_stream]
+format=%%(levelname)s: %%(message)s
+
+[formatter_file]
+format=FILE %%(levelname)s: %%(message)s
+"""
+    logfile = tmpdir.join('borg.log')
+    # touch the file
+    logfile.open(mode='w')
```
ThomasWaldmann: this looks like truncate / create-new rather than touch.

src/borg/testsuite/logger.py
```diff
@@ -39,7 +163,7 @@ def test_parent_module():
     assert find_parent_module() == __name__
 
 
-def test_lazy_logger():
+def test_lazy_logger(io_logger):
```
ThomasWaldmann: hm?

src/borg/testsuite/logger.py
```diff
@@ -16,11 +31,120 @@ def io_logger():
...
+
+        @classmethod
+        def write_logging_conf(cls):
+            data = logging_conf % cls.cfg
+            cls.logging_config_file.write(data)
+
+        @classmethod
+        def logfile_contents(cls):
+            return open(cls.cfg['logfile_path']).read()
+    os.environ['BORG_LOGGING_CONF'] = str(context.logging_config_file)
```
ThomasWaldmann: an already present value will be lost

src/borg/testsuite/logger.py
```diff
@@ -16,11 +31,120 @@ def io_logger():
...
+    yield context
+    del os.environ['BORG_LOGGING_CONF']
```
ThomasWaldmann: if meanwhile the key was already removed, this will crash

src/borg/testsuite/logger.py
leo-b: In the meantime I came to the conclusion that re-setting the logging config shouldn't be our method of choice. See my argumentation below...

src/borg/testsuite/logger.py
leo-b: All of borgs testing code relies on a global logging setup defined in conftest.py. But when testing the logging setup itself, it makes sense to reset it after every test function. My new teardown_function() does that.
Thus ``test_lazy_logger()`` doesn't have a logging setup anymore.
An alternative would be to call ``setup_logging()`` at the beginning.

leo-b:
Resetting the initial state of pythons logging is really hard. Removing all loggers isn't enough. A reference to the root logger is kept in several places inside the python logging module and the whole hierarchic logging configuration is kept inside the logging module.

All borg tests rely on a logging configuration set up in conftest.py. Other previously executed borg test modules have already initialized most of borgs modules, which themselves keep their own reference to their module level logger.

That's why I just came to the conclusion that destroying and re-building the existing logging config won't work.

However, testing the logging setup itself and especially testing a file based logging config requires a fresh environment. I am not a pytest expert... Is there a way to force running the logging tests in separate processes for proper isolation? I guess pytest-xdist would be the way to go...

ThomasWaldmann:
pytest-xdist is usually used for parallelization. but i don't think we should require it for correct execution of tests.

leo-b:
Hmm..
No [py.test --boxed](https://github.com/pytest-dev/pytest-xdist#running-tests-in-a-boxed-subprocess)?
Any other idea?
I think the only way to isolate logging setup properly is to run those tests in separate processes.
If not, the tests could always interfere with each other or damage the global logging setup that all other tests rely on.
Unfortunately I don't think I'll be able to contribute an implementation for that...

ThomasWaldmann:
Well, not really. Except the usual "if something can't be done in a reasonable way, don't do it".

leo-b:
Eventually I digged into pytest and managed to find a way to run particular tests in a separate forked process.
This is based on the [_pytest-forked_ plugin](https://github.com/pytest-dev/pytest-forked), which is also required by _pytest-xdist_, so it shouldn't add any new requirements.
If _pytest-xdist_ isn't available, the corresponding tests are simply skipped with an appropriate message.

PR #3596, opened 2018-02-05
Fixed tab completion problem in Bash

balvik:
Fixed tab completion problem where a space is always added after path even when it shouldn't

When using tab completion, an unwanted space was always added after the path, e.g. borg info /path/to/repo<space>, which meant a backspace always had to be pressed in between directories when traversing them. This commit fixes that.

### DIFF
--- a/scripts/shell_completions/bash/borg
+++ b/scripts/shell_completions/bash/borg
@@ -7,6 +7,7 @@
 
 _borg()
 {
+	compopt -o default
 	COMPREPLY=()
 	local cur="${COMP_WORDS[COMP_CWORD]}"
 	local prev="${COMP_WORDS[COMP_CWORD-1]}"
@@ -18,7 +19,8 @@ _borg()
 	if [[ ${COMP_CWORD} == 1 ]] ; then
 		local borg_commands="init create extract check rename list diff delete prune info mount umount key serve upgrade recreate export-tar with-lock break-lock benchmark config"
 		COMPREPLY=( $(compgen -W "${borg_commands}" -- ${cur}) )
-		return 0
+		compopt +o default
+        	return 0
 	fi
 
 	case "${prev}" in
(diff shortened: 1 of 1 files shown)


scripts/shell_completions/bash/borg
```diff
@@ -18,7 +19,8 @@ _borg()
 	if [[ ${COMP_CWORD} == 1 ]] ; then
 		local borg_commands="init create extract check rename list diff delete prune info mount umount key serve upgrade recreate export-tar with-lock break-lock benchmark config"
 		COMPREPLY=( $(compgen -W "${borg_commands}" -- ${cur}) )
-		return 0
+		compopt +o default
+        	return 0
```
SanskritFritz: Why is this indented?

SanskritFritz: Ah I see, spaces are used instead of tabs. Now I don't know what the preferred method is in borg, but mixing tabs and spaces is not good.

SanskritFritz:
I tested this change and approve it, works perfectly. Thanks!

scripts/shell_completions/bash/borg
balvik: Sorry about the mixed indenting, using a new editor which made a little mess of the indenting. This one spilled by me. Do you want me to fix it in the request or will you adjust it when pulling? (I think it's just that one line).

ThomasWaldmann: for python code we use 4 spaces per indentation level, no tabs.

and, if nothing speaks against that, I'ld prefer the same for other stuff, too.

SanskritFritz: Yes, just checked it, and it would be nice if @balvik could convert the file please.

SanskritFritz: I'll do it gladly after this PR is accepted, and will convert fish and zsh completions as well, so no worries.

SanskritFritz:
Will do it probably tonight.

PR #3600, opened 2018-02-06
Using spaces instead of tabs for indentation

SanskritFritz:
... in bash and fish completions. Zsh was already using spaces.


ThomasWaldmann:
thanks!

PR #3633, opened 2018-02-23
fix borg create never showing M status

ThomasWaldmann:
the problem was that the upper layer code did not have enough information
about the file, whether it is known or not - and thus, could not decide
correctly whether status should be M)odified or A)dded.

now, file_known_and_unchanged method returns an additional "known"
boolean to fix this.

another small fix is to not load the files cache in cache_mode='r' -
in this rechunk mode, files are always considered unknown and get
rechunked, so the files cache is not needed.


ThomasWaldmann:
http://borgbackup.readthedocs.io/en/stable/faq.html#i-am-seeing-a-added-status-for-an-unchanged-file

Hmm, likely that is now `M status`.

PR #3636, opened 2018-02-26
[Docs] Update quickstart.rst to fix typo

jesferman:

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -264,7 +264,7 @@ For automated backups the passphrase can be specified using the
     You can make backups using :ref:`borg_key_export` subcommand.
 
     If you want to print a backup of your key to paper use the ``--paper``
-    option of this command and print the result, or this print `template`_
+    option of this command and print the result, or print this `template`_
     if you need a version with QR-Code.
 
     A backup inside of the backup that is encrypted with that key/passphrase


docs/quickstart.rst
```diff
@@ -264,7 +264,7 @@ For automated backups the passphrase can be specified using the
     You can make backups using :ref:`borg_key_export` subcommand.
 
     If you want to print a backup of your key to paper use the ``--paper``
-    option of this command and print the result, or this print `template`_
+    option of this command and print the result, or print this `template`_
```
ThomasWaldmann: shouldn't we rather talk about --qr-html option here as alternative to --paper rather than linking to an html file?

@textshell  ^

textshell: I like having the template as alternative, because the key is feasible to copy per clipboard and the output of `--qr-html` is not. An at least for servers having a local browser and clipboard possibly over some remote desktop is often the only useful way to print something.

Now if we had a link somewhere else (i think this is the only link to that template) we might want to switch this place to `--qr-html`. But I'm not sure, because in my experience the clipboard is usually more convenient than scp.

ThomasWaldmann: ok, let's deal with that later.

ThomasWaldmann:
thanks!

PR #3652, opened 2018-03-01
Backport 3596 and 3600 to 1.1-maint

miek:
This backports #3596 and #3600 to 1.1-maint as requested in #3645.
I've combined these two into one PR as the second one was a clean-up commit dependant on the first.


ThomasWaldmann:
thanks!

PR #3665, opened 2018-03-05
Improve formatting of GPG output in documentation

anowlcalledjosh:
Previously the GPG output in support.rst was formatted as a blockquote, so the linebreaks would not be preserved (at least when viewing directly on GitHub):

image

Formatting it as a literal block preserves the linebreaks:

image


ThomasWaldmann:
thanks!

ThomasWaldmann:
hmm, guess this is also broken in 1.1-maint and 1.0-maint?

anowlcalledjosh:
yep, looks like it, I'll create a PR for each

Issue #3394, opened 2017-11-27
Reading files cache.... forever


tobecwb:
So, I was trying to use BorgBackup to do a backup of 15TB of data.

This backup is huge, and has several millions of small files (more than 8 millions of files in total).

I tried to create the first backup of all files...
Create a shell script to do the backup of some directories on same repository... with default compression
like:

`borg create --stats --progress /mnt/repo/borg::FTP-{now:%Y-%m-%dT%H:%M:%S} /mnt/files/FTP`
`borg create --stats --progress /mnt/repo/borg::TEMPLATES-{now:%Y-%m-%dT%H:%M:%S} /mnt/files/TEMPLATES`
`borg create --stats --progress /mnt/repo/borg::PUBLIC-{now:%Y-%m-%dT%H:%M:%S} /mnt/files/PUBLIC`
....

Until now, taking 4 days to backup (files on a raid 6 - 80tb -, on same machine, cache on a SSD drive on same machine too, repository on same raid - yes, I know this is wrong) and not finished yet.
Then, I have to cancel the backup, and later tried to resume the backup...

when I start the borg, I receive the following message:
`Reading files cache ....` and this is taking forever...
until now, this message was on screen for more than 8 hours.

I already tried to use the `--debug` to discover if something is wrong, but everything seens to be ok, only taking forever to read the cache...

This is a normal behaviour? 
I known that to do backup of several small files can be problematic and slow, and that it's better to do a image backup, but there's a way to know if something is wrong?

Why I don't to a image backup: The server was configured by another person in this way (all 80tb is on the same partition, and all files on a single partition). I have only 30tb on a NAS, and already has some backups of this data on NAS.
If I wanna to do a image backup, I must delete all the backup on NAS to free space and to a image backup.

ThomasWaldmann:
Check memory usage of borg. Is it using more than you have as physical memory?

If you have a gazillion of files, the files cache might use big amounts of memory. You can run borg without files cache, but it will be slower then for backups of unmodified data.

1.0: --no-files-cache
1.1: --files-cache=disabled

tobecwb:
No, it's using almost nothing of memory...
something like 500mb only (the server has 16gb)

what's I don't get its taking like forever for a directory with few files!
I already backup eight directories in same repository.... some are big, some are small... (in size and number of files)...
now I'm trying to backup a directory with only 127 files, but with 1.5tb of data... this is taking more time than others

ThomasWaldmann:
[...]

PR #3672, opened 2018-03-08
non-lazy files cache

ThomasWaldmann:
read files cache early, if at all - not lazy loading it.

this fixes checkpoint timing as the cache loading happens before the timer is initialized (see #3394).

also cleaned up: do_files and ignore_inode (used by legacy options) now mapped onto cache_mode.

needs forward port to master.


ThomasWaldmann:
@enkore can you review this, please?

ThomasWaldmann:
I am going to merge this in some hours, so if someone wants to review, do it now.

PR #3673, opened 2018-03-08
Mark up Unicode string

tschwinge:
Running Sphinx v1.4.8
    WARNING: the config value 'copyright' is set to a string with non-ASCII characters; this can lead to Unicode errors occurring. Please use Unicode strings, e.g. u'Content'.
    [...]
    Encoding error:
    'ascii' codec can't decode byte 0xc3 in position 23: ordinal not in range(128)

Running `make epub` on a Ubuntu 16.10 system.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,7 +42,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'Borg - Deduplicating Archiver'
-copyright = '2010-2014 Jonas Borgström, 2015-2018 The Borg Collective (see AUTHORS file)'
+copyright = u'2010-2014 Jonas Borgström, 2015-2018 The Borg Collective (see AUTHORS file)'
 
 # The version info for the project you're documenting, acts as replacement for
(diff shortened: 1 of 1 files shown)


tschwinge:
I see this has been discussed in #226 before, but couldn't get resolved then, waiting for #65, which has since been closed.  So, can this change be reconsidered now?

ThomasWaldmann:
can't you just run sphinx with python3?

i guess i personally never had an issue with that because i usually work from a python3 virtual env, so "python" is python 3 there.

tschwinge:
Well, I guess I could ;-) -- but what happened is that I wanted to build an EPUB, on the Ubuntu 16.10 system that I had available during travel.  (... which I then did read, and now looking forward to actually start using borg!)

Anyway, please merge or close, as you prefer; no big deal for me either way.

ThomasWaldmann:
OK, let's try it. Hopefully it won't blow up rtfd now.

PR #3676, opened 2018-03-09
recreate --recompress: add metavar, clarify description

ThomasWaldmann:
@enkore ^^^

borg recreate --help looks now like this:

```
...
  --recompress [MODE]   recompress data chunks according to ``--compression``.
                        MODE `if-different`: recompress if current compression
                        is with a different compression algorithm (the level
                        is not considered). MODE `always`: recompress even if
                        current compression is with the same compression
                        algorithm (use this to change the compression level).
                        MODE `never` (default): do not recompress.
...
```

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3541,12 +3541,17 @@ class Archiver:
                                    type=CompressionSpec, default=CompressionSpec('lz4'),
                                    help='select compression algorithm, see the output of the '
                                         '"borg help compression" command for details.')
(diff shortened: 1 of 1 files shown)


DrupaListo-com:
I like the text above, now it's much better than before.
I'd just reorder thing: start with the defautl=never, then if-different and finally "always". But anyways now it's ok.

ThomasWaldmann:
@DrupaListo-com why mention the option ("never", default) you'll never use first?

PR #3682, opened 2018-03-11
Set cache previous_location on load instead of save (fixes #3304)

### LINKED ISSUE #3304: 'borg config' command

milkey-mouse:
This caused a really stupid bug with borg config --cache, see https://github.com/borgbackup/borg/issues/3304#issuecomment-371896766

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -290,6 +290,7 @@ class CacheConfig:
             self.previous_location = recanonicalize_relative_location(previous_location, self.repository)
         else:
             self.previous_location = None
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@textshell @enkore ^

ThomasWaldmann:
@textshell can you review this? last thing open for 1.1.5.

Issue #3689, opened 2018-03-13
getpass fails when invoking borg without a terminal


baldurmen:
Hi!

I've written a [nagios/icinga2 check](https://gitlab.com/baldurmen/check_borg) to test the multiple borg archives I have, and when running it from a terminal, I have not problems whatsoever.

Sadly, when invoking it through icinga2 as an actual check, getpass fails since borg is not called from an actual terminal.

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 2052, in main exit_code = archiver.run(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1997, in run return func(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 84, in wrapper kwargs['manifest'], kwargs['key'] = Manifest.load(repository)
  File "/usr/lib/python3/dist-packages/borg/helpers.py", line 118, in load key = key_factory(repository, cdata)
  File "/usr/lib/python3/dist-packages/borg/key.py", line 87, in key_factory return RepoKey.detect(repository, manifest_data)
  File "/usr/lib/python3/dist-packages/borg/key.py", line 421, in detect passphrase = Passphrase.getpass(prompt)
  File "/usr/lib/python3/dist-packages/borg/key.py", line 311, in getpass return cls(getpass.getpass(prompt))
  File "/usr/lib/python3.5/getpass.py", line 92, in unix_getpass passwd = fallback_getpass(prompt, stream)
  File "/usr/lib/python3.5/getpass.py", line 127, in fallback_getpass return _raw_input(prompt, stream)
  File "/usr/lib/python3.5/getpass.py", line 149, in _raw_input raise EOFError
EOFError
```

This issue is referenced in the [getpass documentation](https://pymotw.com/2/getpass/#using-getpass-without-a-terminal).

I'm not sure I understand completely how borg deals with passwords, but if it helps, [my check exports BORG_PASSPHRASE](https://gitlab.com/baldurmen/check_borg/blob/master/check_borg#L90-93) as recommended by the documentation.

Do you think you could try to mitigate this by using `isatty` and `readline` as described in the getpass doc?

```
import getpass
import sys

if sys.stdin.isatty():
    p = getpass.getpass('Using getpass: ')
else:
    print 'Using readline'
    p = sys.stdin.readline().rstrip()

print 'Read: ', p
```

ThomasWaldmann:
[...]

PR #3704, opened 2018-03-20
improve getpass user experience, see #3689

ThomasWaldmann:
if interactive passphrase query fails and the env vars are not set,
show a clear error message about this.

users often do 'BORG_PASSPHRASE=secret', forgetting the 'export'.
or they use sudo (and not sudo -E).
in both cases, the env vars won't be available for the borg process.


ThomasWaldmann:
can not acquire a passphrase: BORG_PASSPHRASE is not set. BORG_PASSCOMMAND is not set. Interactive password query failed.

Issue #331, opened 2015-10-24
locking issue with FUSE mount


ThomasWaldmann:
Reproduce like this:
```
borg mount repo mnt
fusermount -u mnt
borg delete repo
# will hang here waiting for exclusive lock
```

ThomasWaldmann:
Idea: PID of locking process is different from the unlocking process' PID due to the fork. That likely creates a mismatch there and might make the unlock fail (or think that there is not "his" lock).

PR #3710, opened 2018-03-23
security: describe chunk size / proximity issue, see #331

ThomasWaldmann:
@enkore @nadalle is this ok for a short doc about it? i didn't want to get too much into the details...


nadalle:
Looks great.

PR #3712, opened 2018-03-23
Vagrant: move cleanup steps to fs_init, fixes #3706

### LINKED ISSUE #3706: Vagrantfile: do more cleanup
Sometimes the state rsynced from the host causes issues.

Removing `.tox/` and all `__pycache__/` fixes it, add it to the script.

ThomasWaldmann:
so after a vagrant up or vagrant reload (which both rsync), we
can just call the fs_init and run_tests provisioning steps and do
not need to call install_borg again.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -264,9 +264,6 @@ def install_borg(fuse)
     . borg-env/bin/activate
     pip install -U wheel  # upgrade wheel, too old for 3.5
     cd borg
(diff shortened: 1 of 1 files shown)


milkey-mouse:
Like #3713, this doesn't quite apply to 1.0 since there's no `fs_init`. If this needs a backport so does 732eab8.

PR #3726, opened 2018-03-25
non-lazy files cache (1.2 forward-port)

milkey-mouse:
See #3672.


ThomasWaldmann:
looks good. did you have conflicts?

milkey-mouse:
Just some stuff in e4125b4 relating to instances of `ignore_inode` that had already been replaced IIRC.  On one hand, it caused it to not apply automatically, on the other hand, the parts that didn't apply were just work that was done for me already.

PR #3731, opened 2018-03-27
Added example of `borg create` using hardlinks to place content in desired location

cpritcha:

### DIFF
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -67,3 +67,17 @@ Examples
     $ cd /home/user/Documents
     # The root directory of the archive will be "projectA"
     $ borg create /path/to/repo::daily-projectA-{now:%Y-%m-%d} projectA
+
+    # Backing up desired paths to different locations within the archive
+    # Create a temporary directory
+    mkdir tmp
+    cd tmp
+    # Hardlink folders and files to the locations you want them to be in borg archive
+    ln /shared/library library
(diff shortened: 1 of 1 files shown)


docs/usage/create.rst
```diff
@@ -67,3 +67,17 @@ Examples
...
+    mkdir tmp
+    cd tmp
+    # Hardlink folders and files to the locations you want them to be in borg archive
+    ln /shared/library library
+    ln /shared/backups/latest/dump.sql backup.sql
+    # Create the archive
+    borg create /path/to/repo::{now} library backup.sql
+    # Optionally delete the temporary directory
+    cd ..
+    rm -rf tmp
```
ThomasWaldmann: thanks for your PR!

not sure we want that in the docs though, because it is stuff outside of borg scope.

basically you just first arrange what you actually want to have backed up and afterwards clean up.

milkey-mouse:
Another way to do it that would be kind of neat (especially if you're creating archives programatically like in [backup-vm](https://github.com/milkey-mouse/backup-vm)) would be to create & stream a tar archive on the fly to `borg import-tar`.

cpritcha:
Maybe there should be a section in usage > general or in the FAQ about the options for programmatically creating a borg archive instead (renaming files and folders, using hardlinks or in the future using `borg import-tar`?

PR #3734, opened 2018-03-27
Use same datetime object for {now} and {utcnow} (fixes #3548)

### LINKED ISSUE #3548: Placeholders {now} and {utcnow} are replaced with timestamps microseconds apart
Both `{now}` and `{utcnow}` should point to exact same moment, save for timezone adjustment. They are not, because current time is queried twice when replacing placeholders:

Borg version: 1.0.11, but master is also affected.

https://github.com/borgbackup/borg/blob/26f1cd148603e2c2762c5f27ef56bf5b9ca17d18/src/borg/helpers/parseformat.py#L189

```
$ borg create \
  --info \
  --stats \
  ':: utcnow: {utcnow:%H:%M:%S.%f}, now: {now:%H:%M:%S.%f}' \
  /dev/null
Archive name:  utcnow: 21:16:01.582465, now: 22:16:01.582458
Archive fingerprint: ...
Time (start): Mon, 2018-01-15 22:16:03
Time (end):   Mon, 2018-01-15 22:16:03
```

milkey-mouse:
{now} and {utcnow} should point to the same exact moment, but they don't because .now() and .utcnow() create two different objects at different times. Although the difference will be on the order of microseconds on all but the slowest machines, this bug still tickles my inner pedant...

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -180,14 +180,14 @@ def format_line(format, data):
 def replace_placeholders(text):
     """Replace placeholders in text with their values."""
     from ..platform import fqdn
-    current_time = datetime.now()
+    current_time = datetime.now(timezone.utc)
     data = {
         'pid': os.getpid(),
         'fqdn': fqdn,
         'reverse-fqdn': '.'.join(reversed(fqdn.split('.'))),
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
thanks!

PR #3739, opened 2018-03-29
run platform.* calls within a cleaned env, fixes #3732

### LINKED ISSUE #3732: ncurses version mismatch in release 1.1.4
I never worked with Vagrantfiles before so I don't see an obvious solution.  I downloaded `borg-linux64` from the `1.1.4` release and it seems to work, but if I try to create an already-existing name, in addition to failing as expected I get a dynamic linker error:

```text
$ borg create --stats -x -C lz4 /tmp/testborg::iodocker1 io-docker
Enter passphrase for key /tmp/testborg: 
/bin/sh: symbol lookup error: /lib/x86_64-linux-gnu/libncurses.so.5: undefined symbol: _nc_putchar
Archive iodocker1 already exists
```

Considering that the binary does not in fact link with my system's ncurses:

```text
$ ldd bin/borg
	linux-vdso.so.1 (0x00007ffcad78a000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f65f6ec1000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f65f6ca6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f65f68fa000)
	/lib64/ld-linux-x86-64.so.2 (0x000055f8a1b34000)
```

…I can reasonably conclude that the problem lies inside the virtual environment created with Vagrant.  Something, maybe `/bin/sh`, needs another version of ncurses than what's bundled. :(

ThomasWaldmann:


src/borg/helpers.py
```diff
@@ -1602,11 +1602,29 @@ def finish(self):
 
 
 def sysinfo():
+    orig_env = os.environ
+    cleaned_env = prepare_subprocess_env(system=True)
+    linux_distribution = None
+    try:
+        # some platform calls internally run /bin/sh, so run them within a cleaned env,
+        # so that sh doesn't use pyinstaller env libs, but system libs, see #3732
+        os.environ = cleaned_env
```
ThomasWaldmann: not sure whether this is good enough, so it actually ends up in internally created subprocesses.

PR #3740, opened 2018-03-29
tox/travis: test on python 3.7-dev

ThomasWaldmann:
some rolling release linux dists might have 3.7 later this year, check if we are compatible.


ThomasWaldmann:
Looks like xenial VMs do not work at all. And the py37 on trusty is stoneage.

Deferring it to later again...

ThomasWaldmann:
guess we rather test on / fix for 3.7 in master / vagrant first.

PR #3768, opened 2018-04-12
Zstd update

ThomasWaldmann:


jungle-boogie:
Nice work, thanks!

Looking forward to zstd with borgbackup.

ThomasWaldmann:
@jungle-boogie we have zstd since borg 1.1.4, this is just updating to the newer zstd release which is a bit faster.

It only matters if the bundled zstd code is used. Linux maintainers likely disable this for the sake of better package maintenance via a separate zstd library package.

jungle-boogie:
hi @ThomasWaldmann,

https://www.borgbackup.org

`Compression: LZ4, zlib, LZMA.`

Is zlib zstd?

ThomasWaldmann:
No, zlib == .gz (and rather old).

zstd = fancy new compression stuff.

ThomasWaldmann:
TODO: platform testing

PR #3771, opened 2018-04-16
docs: describe what BORG_LIBZSTD_PREFIX does

Moredread:


docs/usage_general.rst.inc
```diff
@@ -243,6 +243,9 @@ Building:
     BORG_LIBB2_PREFIX
         Adds given prefix directory to the default locations. If a 'include/blake2.h' is found Borg
         will be linked against the system libb2 instead of a bundled implementation. (setup.py)
+    BORG_LIBZSTD_PREFIX
+        Adds given prefix directory to the default locations. If a 'include/zstd.h' is found Borg
+        will be linked against the system libzstd instead of a bundled implementation. (setup.py)
```
ThomasWaldmann: can you check how this looks in 1.1-maint branch? if it is already described there, forward port it to master rather than writing new docs. if it's not, we need a backport of your change to 1.1-maint.

Moredread: It is missing there, but this PR commit can be cherry-picked for that branch without problems. Can you do this when merging this PR, or should I create another one for that branch?

ThomasWaldmann: If you can create another one, please do. Otherwise I'll cherrypick later.

PR #3774, opened 2018-04-17
docs: Remove outdated default encryption

lumbric:
According to changelog, there is no default encryption since 1.1.x.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -260,7 +260,8 @@ Repository encryption
 ---------------------
 
 Repository encryption can be enabled or disabled at repository creation time
-(the default is enabled, with `repokey` method)::
+(for 1.0.x the default value is `repokey`, for >= 1.1.x setting `--encryption` 
+explicitly is mandatory)::
 
     $ borg init --encryption=none|repokey|keyfile PATH
 


docs/quickstart.rst
```diff
@@ -260,7 +260,8 @@ Repository encryption
 ---------------------
 
 Repository encryption can be enabled or disabled at repository creation time
-(the default is enabled, with `repokey` method)::
+(for 1.0.x the default value is `repokey`, for >= 1.1.x setting `--encryption` 
+explicitly is mandatory)::
```
ThomasWaldmann: good catch! but we need to rephrase that, it's not mandatory, only if you want encryption.

how about:

(for borg 1.0.x, `--encryption repokey` was the default. for borg >= 1.1, you need to give `--encryption MODE` explicitly to enable encryption.)

lumbric: I think this is not correct. borg 1.1.5 says without `-e`:

`borg init: error: the following arguments are required: -e/--encryption`

From [changelog](https://borgbackup.readthedocs.io/en/stable/changes.html#version-1-1-0b4-2017-03-27): "init: the –encryption argument is mandatory now"

lumbric:
> I've just seen that more changes are needed here.

Hm yes, it might be worth mentioning the [new encryption modes in 1.1.x](http://borgbackup.readthedocs.io/en/stable/usage/init.html#encryption-modes). I didn't know about them when I made the pull request an hour ago.

docs/quickstart.rst
ThomasWaldmann: hmm, right. encryption is not mandatory, but `--encryption` is.

ThomasWaldmann:
see #3776.

PR #3780, opened 2018-04-19
docs: Improve borg with-lock

infectormp:
* move note about repo copy to FAQ
* add link to FAQ

### DIFF
--- a/docs/usage/with-lock.rst.inc
+++ b/docs/usage/with-lock.rst.inc
@@ -57,5 +57,5 @@ running in the repo), then execute the given command as a subprocess and wait
 for its termination, release the lock and return the user command's return
 code as borg's return code.
 
-As example you can use "borg with-lock" for copy repo to another location
-(see :ref:`can_i_copy_or_synchronize_my_repo_to_another_location` in the FAQ)
+As example you can use "borg with-lock" to copy a repo to another location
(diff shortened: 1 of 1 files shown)


docs/usage/with-lock.rst.inc
```diff
@@ -57,9 +57,5 @@ running in the repo), then execute the given command as a subprocess and wait
...
 code as borg's return code.
 
-.. note::
-
-    If you copy a repository with the lock held, the lock will be present in
-    the copy. Thus, before using borg on the copy from a different host,
-    you need to use "borg break-lock" on the copied repository, because
-    Borg is cautious and does not automatically remove stale locks made by a different host.
\ No newline at end of file
+As example you can use "borg with-lock" for copy repo to another location
```
ThomasWaldmann: I'm not a native speaker, but isn't it rather "... to copy a repo to another location".

docs/usage/with-lock.rst.inc
```diff
@@ -57,9 +57,5 @@ running in the repo), then execute the given command as a subprocess and wait
...
+(see :ref:`can_i_copy_or_synchronize_my_repo_to_another_location` in the FAQ)
```
ThomasWaldmann: Also, add a dot at the end of the sentence. </nitpick>

docs/usage/with-lock.rst.inc
infectormp: you right

Issue #3612, opened 2018-02-12
List of all repository options and their values


seschwar:
`borg config` allows to get and set options for a repository, provided you know the option's name. Unfortunately there's no way to get a list of available options on the command line.  Even in the documentation they aren't listed in a single place.  There is no way besides reading through all of the documentation to find out the possible repository options.

Therefore there should be an option to `borg config` that lists all repository options and their values similar to `git config --list`, for example.

SanskritFritz:
Shell tab completions could also benefit from this, so 👍

bhaveshpoddar94:
First Timer here. Could I try and solve this?

SanskritFritz:
This is open source. If you have the time and the knowledge, do it, and send a pull request :smile:

milkey-mouse:
I closed a duplicate of this issue, see #3671 for more discussion.

Gu1nness:
Since argparse is kind of stupid (patched in python 3.7 but not backward compatible)
the only solution I see is to allow
`borg config --list <repo>` and `borg config --delete <repo> <name>` but not `borg config <repo> --delete <name>`
This is because argparse does not handles well mixed optional and positional arguments.
Thus, it would need to modify a little the test suite.

Is it a problem if I do so ?
I have a patch ready for this.

Gu1nness:
Ok perfect.
Then, I'll change [this test](https://github.com/borgbackup/borg/blob/master/src/borg/testsuite/archiver.py#L2794) also in order to be compliant with the documentation.

PR #3784, opened 2018-04-22
Cat configuration file

Gu1nness:
Add a `borg config --list <repo>` option that dumps the configuration file.

Refers and https://github.com/borgbackup/borg/issues/3612.

I have patched a test that was not compliant with [the documentation](http://borgbackup.readthedocs.io/en/stable/usage/general.html#positional-arguments-and-options-order-matters), and added a test that checks that `--delete` and `--list` cannot be used in the same time.


src/borg/archiver.py
```diff
@@ -3637,13 +3641,17 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         subparser.set_defaults(func=self.do_config)
         subparser.add_argument('-c', '--cache', dest='cache', action='store_true',
                                help='get and set values from the repo cache')
-        subparser.add_argument('-d', '--delete', dest='delete', action='store_true',
+
+        group = subparser.add_mutually_exclusive_group()
+        group.add_argument('-d', '--delete', dest='delete', action='store_true',
                                help='delete the key from the config file')
+        group.add_argument('-l', '--list', action='store_true',
```
ThomasWaldmann: maybe give dest='list' here to be consistent in style with 2 lines above.

src/borg/archiver.py
```diff
@@ -3637,13 +3641,17 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+                               help='list the configuration options if the repo')
```
ThomasWaldmann: of the repo

src/borg/testsuite/archiver.py
```diff
@@ -2793,6 +2793,7 @@ def test_config(self):
             assert output == cfg_value + '\n'
             self.cmd('config', '--delete', self.repository_location, cfg_key)
             self.cmd('config', self.repository_location, cfg_key, exit_code=1)
+            self.cmd('config', '--list', '--delete', self.repository_location, exit_code=2)
```
ThomasWaldmann: doesn't look like there is a test whether --list works as expected.

Gu1nness: I'll have a look at it.
The problem I have is that some values change at each new repo (repo id for example). I'll try to find a way to check that it works as expected.

src/borg/archiver.py
Gu1nness: Indeed, my bad, I'll fix that.

src/borg/archiver.py
Gu1nness: Ok, no problem :)

Gu1nness:
Concerning the two differents possibilities: I think that listing all the keys and their values (set or default) may be the best idea.
Maybe we can also add something that explicitely shows that it is a default value (I'll have a look at the code to find where all these values are stored and everything)
As a side question, for now, is there a resource that lists all possible keys somewhere ? If not, I think it can be a good idea to add this in the documentation.

src/borg/archiver.py
```diff
@@ -1567,11 +1568,32 @@ def cache_validate(section, name, value=None, check_value=True):
...
-            section, name = args.name.split('.')
-        except ValueError:
-            section = args.cache and "cache" or "repository"
-            name = args.name
+        def list_config(config):
+            default_values = {
+                'version': '1',
+                'segments_per_dir': str(DEFAULT_SEGMENTS_PER_DIR),
+                'max_segment_size': str(MAX_SEGMENT_SIZE_LIMIT),
+                'additionnal_free_space': '0',
```
ThomasWaldmann: typo: only 1 "n".

src/borg/archiver.py
```diff
@@ -1567,11 +1568,32 @@ def cache_validate(section, name, value=None, check_value=True):
...
+            print('[repository]')
+            for key in ['version', 'segments_per_dir', 'max_segment_size',
+                        'storage_quota', 'additional_free_space', 'append_only',
+                        'id']:
+                value = config.get('repository', key, fallback=False)
+                if value:
+                    print('%s = %s' % (key, value))
+                else:
+                    assert key != 'id'
+                    print('%s = %s' % (key, default_values[key]))
```
ThomasWaldmann: How about this:
```
if value is None:
    value = default_values.get(key)
    if value is None:
        raise Error('The repository config is missing the %s key which has no default value.' % key)
print('%s = %s' % (key, value))
```

src/borg/archiver.py
```diff
@@ -3621,10 +3645,13 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         This command gets and sets options in a local repository or cache config file.
         For security reasons, this command only works on local repositories.
 
-        To delete a config value entirely, use ``--delete``. To get an existing key, pass
-        only the key name. To set a key, pass both the key name and the new value. Keys
-        can be specified in the format "section.name" or simply "name"; the section will
-        default to "repository" and "cache" for the repo and cache configs, respectively.
+        To delete a config value entirely, use ``--delete``. To list the values
+        stored in the configuration file, use ``--list``.  To get an existing
```
ThomasWaldmann: well, in fact it lists the value from the config file OR the default.

src/borg/archiver.py
```diff
@@ -3637,13 +3664,17 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
         subparser.set_defaults(func=self.do_config)
         subparser.add_argument('-c', '--cache', dest='cache', action='store_true',
                                help='get and set values from the repo cache')
-        subparser.add_argument('-d', '--delete', dest='delete', action='store_true',
+
+        group = subparser.add_mutually_exclusive_group()
+        group.add_argument('-d', '--delete', dest='delete', action='store_true',
                                help='delete the key from the config file')
+        group.add_argument('-l', '--list', dest='list', action='store_true',
+                               help='list the configuration options of the repo')
```
ThomasWaldmann: i'ld remove "options" here. options sound rather like the stuff that could be configured, but the code lists the actual current configuration (and not only the keys, but also the values).

src/borg/testsuite/archiver.py
```diff
@@ -57,6 +57,7 @@
 from . import are_symlinks_supported, are_hardlinks_supported, are_fifos_supported, is_utime_fully_supported, is_birthtime_fully_supported
 from .platform import fakeroot_detected
 from .upgrader import attic_repo
+
```
ThomasWaldmann: don't do unwanted changes.

src/borg/testsuite/archiver.py
```diff
@@ -2791,8 +2800,9 @@ def test_config(self):
             self.cmd('config', self.repository_location, cfg_key, cfg_value)
             output = self.cmd('config', self.repository_location, cfg_key)
             assert output == cfg_value + '\n'
-            self.cmd('config', self.repository_location, '--delete', cfg_key)
+            self.cmd('config', '--delete', self.repository_location, cfg_key)
```
ThomasWaldmann: good. order matters. :)

src/borg/archiver.py
```diff
@@ -1,4 +1,5 @@
 import argparse
+import sys
```
ThomasWaldmann: i don't think you used that in the end, right?

ThomasWaldmann:
Also it looks like the stuff needs some `git rebase -i master` and resolving of conflicts.

ThomasWaldmann:
btw, i'll release 1.1.6 in the near future. if the remaining minor issues could be resolved until then, we could merge this before the release.

PR #3803, opened 2018-05-05
make "borg config repo" possible without key

ThomasWaldmann:
the `with_repository` context manager was setting up repository and cache (because the cache config can also get changed via `borg config`).

but, setting up the cache needed to load the manifest because it did some checks against that - and that needs the key / passphrase to unlock the key (which is unneeded if we just want to use borg config on the repo config).

to solve, I only set up the repository with `with_repository` CM and only set up the cache (and key/manifest) when a cache configuration change is desired.

the individual, minimal commits are easier to review than the overall change.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1774,38 +1774,39 @@ class Archiver:
             assert_secure(repository, manifest)
             cache = Cache(repository, key, manifest, lock_wait=self.lock_wait)
 
-        if args.cache:
-            cache.cache_config.load()
-            config = cache.cache_config._config
-            save = cache.cache_config.save
-            validate = cache_validate
-        else:
-            config = repository.config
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1774,38 +1774,39 @@ def cache_validate(section, name, value=None, check_value=True):
...
+            else:
+                try:
+                    print(config.get(section, name))
+                except (configparser.NoOptionError, configparser.NoSectionError) as e:
+                    print(e, file=sys.stderr)
+                    return EXIT_WARNING
+            return EXIT_SUCCESS
+        finally:
+            if args.cache:
+                cache.close()
```
ThomasWaldmann: only the try/finally is new, the rest just got indented / moved to the right place.

PR #3852, opened 2018-05-22
borg delete: only commit once, fixes #3823

### LINKED ISSUE #3823: borg delete (archives): commit only once
see do_prune: it also deletes multiple archives, but commits only once.

if we can do that for borg prune, we can also do it for borg delete (and be quicker).

ThomasWaldmann:
borg delete was slower than required when it deleted multiple archives
as it committed repo and cache once per archive.

borg prune (doing a similar job) is faster as it commits only once for
all deleted archives.

borg delete now also only commits once and also (similar to borg prune)
only outputs a overall statistics for all deleted archives.

log output of borg delete was made similar to borg prune.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1262,27 +1262,21 @@ class Archiver:
                 logger.warning('Aborted.')
             return self.exit_code
 
-        stats_logger = logging.getLogger('borg.output.stats')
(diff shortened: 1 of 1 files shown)


nemaniarjun:
Looks good!

Issue #3753, opened 2018-04-03
msgpack(-python), document versions and issues


ThomasWaldmann:
The 0.5+ releases of msgpack(-python) were a bit of a PITA:
- rename of the package and bad side effects of that (the fix from msgpack project was to make them also available under the old name)
- TypeErrors due to api changes in 0.5.3 (fixed in later versions)
- introducing that FutureWarning, which made borg output somehow unpretty / annoying (the fix from msgpack was to remove the FutureWarning again at some version)
- lack of py34 testing introducing some uncertainty for borg, which still supports py34 (no fix)

Likely there are some working 0.5+ versions if one has py35+ or does own testing on py34, but one needs to check if there are FutureWarnings in borg output.

So the `msgpack-python<0.5` is basically a "we know that the given versions work, but when using newer versions you need to do own tests".

There are currently no plans to drop py34 compatibility in borg within the 1.1.x stable release series, so using any 0.5+ version is problematic due to the lack of testing within the msgpack project.

Also, it is not planned to change the name from `msgpack-python` to `msgpack` in our requirements within the 1.1.x stable release series, because the required versions are available under the old name and we do not want to break any distribution which uses the old name still.

If borg gets packaged for distribution based on py35 or py36, the lack of py34 testing is no issue, so newer msgpack-python versions can be used, but should be tested first. You can also change the package name to `msgpack` when making a distribution package, but you need to test if the result actually works.

LocutusOfBorg:
```
   * Add debian/python{3,}-msgpack.pydist to help transition to the new upstream
     egg name.
```
In Debian we are providing the old and new namespace with the same source package, so we don't care anymore about the renaming.

bpereto:
In recent fedora Version (rawhide, f28) there is no python3-msgpack lower than 0.5.0+.

So, basically I must use 0.5.0+ but don't know if its working?

ThomasWaldmann:
@bpereto I'll do some tests...

ThomasWaldmann:
0.4.8: heavily tested, works ok, last version with official support for python 3.4

ThomasWaldmann:
pip install msgpack-python==0.5.0

broken, installs msgpack-python==0.5.0 AND msgpack==0.5.6 !?

ThomasWaldmann:
0.5.1: pip install msgpack-python==0.5.1

works better, installs msgpack-python==0.5.1, as expected.

ThomasWaldmann:
[...]

PR #3891, opened 2018-06-11
Fix msgpack version constraints using proper setuptools metadata.

eli-schwartz:
Embrace environment markers (a.k.a. PEP 508 compliance)! They let you do sane handling of different requirements for different versions of python. There's no need to bend over backwards to try to ensure usage of python34-compatible dependencies on python36.

Trying to handle <=0.4 just because python34 needs to be able to match on that, results in a very awkward list. It also means that Arch Linux has to periodically patch this out, and then re-patch it when our previous patch does not apply, then get more bugs opened once it fails a third time.

It's really not worth it to support ancient PyPI names and versions when you can just use a separate branch for python34.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,8 @@ install_requires = [
     # we are rather picky about msgpack versions, because a good working msgpack is
     # very important for borg, see https://github.com/borgbackup/borg/issues/3753
     # best versions seem to be 0.4.6, 0.4.7, 0.4.8 and 0.5.6:
-    'msgpack-python >=0.4.6, <=0.5.6, !=0.5.0, !=0.5.1, !=0.5.2, !=0.5.3, !=0.5.4, !=0.5.5',
+    'msgpack-python <0.5;python_version=="3.4"',
+    'msgpack >=0.5.6;python_version >="3.5"',
     # if you can't satisfy the above requirement, these are versions that might
     # also work ok, IF you make sure to use the COMPILED version of msgpack-python,
     # NOT the PURE PYTHON fallback implementation: ==0.5.1, ==0.5.4
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ install_requires = [
     # we are rather picky about msgpack versions, because a good working msgpack is
     # very important for borg, see https://github.com/borgbackup/borg/issues/3753
     # best versions seem to be 0.4.6, 0.4.7, 0.4.8 and 0.5.6:
-    'msgpack-python <0.5;python_version=="3.4"',
+    'msgpack-python >=0.4.6, <0.5;python_version=="3.4"',
     'msgpack >=0.5.6;python_version >="3.5"',
     # if you can't satisfy the above requirement, these are versions that might
     # also work ok, IF you make sure to use the COMPILED version of msgpack-python,


setup.py
```diff
@@ -39,7 +39,8 @@
     # we are rather picky about msgpack versions, because a good working msgpack is
     # very important for borg, see https://github.com/borgbackup/borg/issues/3753
     # best versions seem to be 0.4.6, 0.4.7, 0.4.8 and 0.5.6:
-    'msgpack-python >=0.4.6, <=0.5.6, !=0.5.0, !=0.5.1, !=0.5.2, !=0.5.3, !=0.5.4, !=0.5.5',
+    'msgpack-python <0.5;python_version=="3.4"',
+    'msgpack >=0.5.6;python_version >="3.5"',
```
ThomasWaldmann: the problem is just that this is not what we want:
- we do not want to accept almost any early msgpack version on 3.4
- we do not want to require a very recent on >=3.5

ThomasWaldmann:
About "ancient pypi names": the old name was in use when we released borg 1.0 and 1.1, so it is not ancient. Not going to break package dependencies all over the place by updating the package name within a borg patch release. But I opened an issue to do it in borg 1.2.

setup.py
ThomasWaldmann: Also: i tested 0.5.6 on py34, it was ok (see that big msgpack issue on our issue tracker). So it can be used on py34.

ThomasWaldmann: Also: msgpack made a big mess in 0.5.x (see same issue), so I won't trust in future releases "just working" without checking first - so no way for ">=0.5.6".

ThomasWaldmann: Also note how on github only 0.4.6 is marked as "release" and everything newer is "pre-release"...

eli-schwartz: > we do not want to require a very recent on >=3.5

Why not? Having to deal with two different naming schemes on >=3.5 (msgpack-python is not guaranteed to exist, it emphatically does not exist on Arch Linux) does not seem very reasonable. Moreover, having to fight with pip to ensure that the dummy package name actually installs the thing it's supposed to install, is just asking for trouble!

Meanwhile, anyone who is *updating* to new versions of borg, should have no problem *updating* msgpack too. I'm not sure I understand the motivation to support old stuff when there's compelling reasons to not do so, and *zero* motivation to do so.

This would work flawlessly for people using pip. By default they get the latest version of msgpack and the latest version of borg.

If this is solely about distro support for old versions of msgpack, then any distro which does not want to upgrade msgpack can be assumed to not want to upgrade borg either... holding the project hostage to old distros and requiring the *newer* distros to be the ones to patch borg in order to remain functional, seems entirely backwards.

> So it can be used on py34.

I did not see that anywhere. BTW are you planning on continually testing newer msgpack releases on python 3.4 before adding it for python>=3.5 at all? If it stops working on python 3.4 will you just freeze the pinned msgpack version for every python release?

> Also: msgpack made a big mess in 0.5.x (see same issue), so I won't trust in future releases "just working" without checking first - so no way for ">=0.5.6".

I don't see why "it's slow, but only if you build from git without cython installed" or "it prints warnings which aren't wrong", or "it removed private APIs which are not permitted to be used, and broke bad code as a result", are reasons to fundamentally distrust it. The biggest issue I see is that the dummy package causes issues for people who do not migrate to the new name...

Plus one regression fixed later the same day.

And I'd guess this is why Pipfile.lock exists anyway.

> Also note how on github only 0.4.6 is marked as "release" and everything newer is "pre-release"...

I also note how *you're* the one who opened msgpack/msgpack-python#298 and discovered that these tags have no meaning. I mean, I guess I don't know why they're marked as anything at all, maybe AppVeyor does this?

It's a bit disingenuous to then bring this as a naysaying argument here...

eli-schwartz:
> Not going to break package dependencies all over the place by updating the package name within a borg patch release.

In which situations does it break package dependencies?

kbabioch:
@ThomasWaldmann The whole situation is a mess right now and basically everybody is patching around those issues (for instance [1] [2]). Please, try do not make it unnecessarily hard for people trying to stay on current versions, using current versions of Python.

[1]: https://build.opensuse.org/package/show/Archiving:Backup/borgbackup
[2]: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/borg

ThomasWaldmann:
closing this for the already mentioned reasons.

not going to break packages in a patch release.

i will change to "msgpack" name, but it'll have to wait until 1.2.0.

ThomasWaldmann:
The easy fix for dists who already have switched to "msgpack" name: s/msgpack-python/msgpack/ for borg's setup.py. done.

eli-schwartz:
Gentoo's solution is to simply remove all mention of msgpack from setup.py: https://gitweb.gentoo.org/repo/gentoo.git/tree/app-backup/borgbackup/borgbackup-1.1.5.ebuild?id=49d54aea50ae504fa087e4cf25e30ec8b9502d5f#n44

Debian only provides borgbackup > 1.0.9 in Debian testing, where they provide msgpack 0.5.6 and patch setup.py yet again, but this time to allow 0.5.x at all. 

Fedora never patched borgbackup, but they did mark it as depending exclusively on msgpack < 0.5.0

eli-schwartz:
> The easy fix for dists who already have switched to "msgpack" name:
> 
> s/msgpack-python/msgpack/ in borg's setup.py.

We tried that. It didn't work, since we also had to un-pin the dependency on <0.5.0. Then the sed line we used broke because the match no longer worked after 5be756a0c6d526fc513fa0ad08767b127b941e90.

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/3753 some more details on why the msgpack-python requirements are as they are.

ThomasWaldmann:
It's not due to your prodding... EOD for me, this is not productive.

PR #3892, opened 2018-06-12
Embrace environment markers (a.k.a. PEP 508 compliance)!

eli-schwartz:
Failure to use environment markers means the dependencies are unconditionally added at build time based on the host instead of being always present and evaluated at runtime on the target; e.g. wheels have the wrong information.

Also use python_requires. This teaches pip how to natively comprehend when the current version of python is not supported by borg.

Returns errors in the format:
borgbackup requires Python '>=3.5' but the running Python is $oldver

...

Doesn't look like borg officially provides wheels, and since it includes binary extensions anyway the wheels are both OS and python version dependent, but it's the principle of the matter. :D

Environment markers are nicer. They've also been around since late 2015, so they should be getting used more in projects.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,13 +38,6 @@ prefer_system_libzstd = True
 # True: use the shared libb2 from the system, False: use the bundled blake2 code
 prefer_system_libb2 = True
 
-min_python = (3, 5)
-my_python = sys.version_info
-
-if my_python < min_python:
-    print("Borg requires Python %d.%d or later" % min_python)
-    sys.exit(1)
-
 cpu_threads = multiprocessing.cpu_count() if multiprocessing else 1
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
looks good.

not sure how good this works on platforms that do not have the latest stuff, though, so I guess we can try this in master, but not backport to 1.1 or 1.0.

eli-schwartz:
It works anywhere you have a version of setuptools newer than May 2016: https://setuptools.readthedocs.io/en/latest/history.html#v20-6-8

Or pip >=6 (December 2014) : https://pip.pypa.io/en/stable/news/#id36

This is pretty stable...

ThomasWaldmann:
thanks!

PR #3918, opened 2018-06-21
xattr: move to platform pkg, use cython, use bytes, use fd

### LINKED ISSUE #2495: Integrate borg.xattr into platform modules
Native code (Cythonized or C) avoids the whole "let's go find a libc" mess (including the "let's go and parse LD_PRELOAD and then parse the output of some command")

ThomasWaldmann:
this code used to live in borg.xattr and used ctypes
(and was the only ctypes-using code in borg).

the low level code now was converted to cython and
the platform code moved to platform package.

got rid of the code that tried to find the libc.


ThomasWaldmann:
The code is only tested on linux yet, travis will do tests on OS X also.

Please review.

ThomasWaldmann:
tried freebsd12:
- with fakeroot: lots of EISDIR fails all over the place (not xattr related)
- without fakeroot: "invalid argument" for listxattr('input/fusexattr')

ThomasWaldmann:
@tim-seoss yeah, having this on the radar. doesn't help with xattrs/acls though and still needs a while until everybody (on linux) has that.

src/borg/platform/linux.pyx
```diff
@@ -8,11 +8,25 @@ from ..helpers import user2uid, group2gid
...
+cdef extern from "attr/xattr.h":
+    ssize_t c_listxattr "listxattr" (const char *path, char *list, size_t size)
+    ssize_t c_llistxattr "llistxattr" (const char *path, char *list, size_t size)
+    ssize_t c_flistxattr "flistxattr" (int filedes, char *list, size_t size)
+
+    ssize_t c_getxattr "getxattr" (const char *path, const char *name, void *value, size_t size)
+    ssize_t c_lgetxattr "lgetxattr" (const char *path, const char *name, void *value, size_t size)
+    ssize_t c_fgetxattr "fgetxattr" (int filedes, const char *name, void *value, size_t size)
+
+    int c_setxattr "setxattr" (const char *path, const char *name, const void *value, size_t size, int flags)
```
ThomasWaldmann: the declaration for listxattr, getxattr, setxattr is missing completely in the ctypes code:

https://github.com/borgbackup/borg/pull/3918/files#diff-87630d54442e9151262050bf3d4bad49L192

ThomasWaldmann:
about freebsd12 issues:
- the EISDIR issues seen with fakeroot are nothing new. freebsd10 testing in 1.1-maint just did not have fakeroot installed.
- the issues seen in fuse xattr tests seem to be a bug in freebsd12 fuse code, normal native fs works ok with xattrs. so, not a borg issue.

general issues:
- the currently hardcoded XATTR_FAKEROOT flag influences which tests are executed and how, likely this should get reviewed / improved / removed.

Issue #2850, opened 2017-07-21
recreate leaves emtpy segments behind


fsironman:
After running the recreate command with borg-linux64_1.1.0b6 there are a lot of emtpy segment files left in the data dir

ls -lh borg/data/0
[...]
`-rw-rw-r-x 1 myuser localsamba   17 Jul 21 12:23 4338`
`-rw-rw-r-x 1 myuser localsamba   17 Jul 21 12:24 4340`
`-rw-rw-r-x 1 myuser localsamba   17 Jul 21 12:24 4342`
`-rw-rw-r-x 1 myuser localsamba   17 Jul 21 12:24 4344`
`-rw-rw-r-x 1 myuser localsamba   17 Jul 21 12:24 4346`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 13:31 4348`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 13:32 4356`
`-rw-rw-r-x 1 myuser localsamba  31M Jul 21 13:32 4357`
[...]

Full paste:
https://pastebin.com/6y0Gxy0X

They stay there even after a new delete or create into the repository.
The only way for me to fix this was was to run borg check --repair
At the end of the repository check the "empty"  17 byte segments were deleted.

ls -lh    

`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 10:55 4032`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 10:55 4033`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 10:55 4034`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 13:31 4348`
`-rw-rw-r-x 1 myuser localsamba 201M Jul 21 13:32 4356`
`-rw-rw-r-x 1 myuser localsamba  31M Jul 21 13:44 4360`

I changed the max_segment_size = 209715200 in the config
not sure if this might be causing the problem

ThomasWaldmann:
iirc, they are not really empty, but contain a commit tag.

ThomasWaldmann:
@enkore ^

enkore:
That's a limitation in the current compaction algorithm which doesn't track whether it has any data to commit. If a segment is completely superseded (as recreate may do), then a commit-only segment shows up.

fsironman:
A followup on this one: I tested borg 1.1 RC1 and even normal backup leave empty segment files behind:

```
[...]

Issue #2195, opened 2017-02-22
separate compaction


ThomasWaldmann:
Some (all?) repo-writing borg commands call compact() after their original work is done, to optimize the repository (remove non-compact segments).

I am thinking about having a separate `borg compact` command (or having it as part of a `borg optimize` command) because:
- not compacting after borg delete will likely accelerate delete a lot, good for interactive usage
- doing compact only once after multiple repo-writing commands (e.g. multiple delete) is faster (considering total time needed)
- some commands might be less prone to crash (e.g. borg delete --force (--force)). while trying to write tests for these, I had often borg crash in compact() when it touched corrupted repo parts. this had the slightly strange effect that the archive I wanted to delete was gone although borg crashed, raising questions about transactional behaviour.
- like borg prune (delete), it could be run by a different user for append-only repos

level323:
FWIW I think it's a good idea.

enkore:
`borg debug compact-repository`

ThomasWaldmann:
@enkore the idea was rather to remove the compact_segments() from most/all commands and have a separate command just for that (which would be a normal command then, not a debug command).

rugk:
Could not one already trigger this behavior (for now) by adding some random (fake, not existent) `borg debug put-obj` and `borg debug del-obj` or would that be a bit dangerous (when done in production archives?)?

ThomasWaldmann:
The writing borg commands would still call compact_segments() to not change behaviour there.

We could add a `--no-compact` option to skip that, so delete and prune would work much faster, delaying some of the work until compact is called.

PR #3925, opened 2018-06-24
compact commits / separate compaction

ThomasWaldmann:
this is trying to get @enkore's work on #2850 finished.

TODO:
- client/server compatibility?
- update docs about major change in space-freeing behaviour

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -426,7 +426,9 @@ class Repository:
             raise exception
         self.check_free_space()
         self.log_storage_quota()
-        self.io.write_commit()
+        segment = self.io.write_commit()
+        self.segments.setdefault(segment, 0)
+        self.compact[segment] += LoggedIO.header_fmt.size
         if compact and not self.append_only:
             self.compact_segments()
         self.write_index()
@@ -676,6 +678,8 @@ class Repository:
             nonlocal unused
             # commit the new, compact, used segments
             segment = self.io.write_commit(intermediate=intermediate)
+            self.segments.setdefault(segment, 0)
+            self.compact[segment] += LoggedIO.header_fmt.size
             logger.debug('complete_xfer: wrote %scommit at segment %d', 'intermediate ' if intermediate else '', segment)
             # get rid of the old, sparse, unused segments. free space.
             for segment in unused:
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -21,6 +21,7 @@ from .helpers import ProgressIndicatorPercent
 from .helpers import bin_to_hex
 from .helpers import hostname_is_unique
 from .helpers import secure_erase, truncate_and_unlink
+from .helpers import Manifest
 from .locking import Lock, LockError, LockErrorT
 from .logger import create_logger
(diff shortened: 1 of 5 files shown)


src/borg/testsuite/repository.py
```diff
@@ -728,20 +728,21 @@ def test_repair_missing_segment(self):
...
-        self.delete_segment(3)
+        self.delete_segment(5)  # XXX problem: if the last and only commit is killed, we have to diagnose "beyond repair"
         self.assert_raises(Repository.ObjectNotFound, lambda: self.get_objects(4))
         self.assert_equal(set([1, 2, 3]), self.list_objects())
 
     def test_repair_corrupted_commit_segment(self):
         self.add_objects([[1, 2, 3], [4, 5, 6]])
-        with open(os.path.join(self.tmppath, 'repository', 'data', '0', '3'), 'r+b') as fd:
+        with open(os.path.join(self.tmppath, 'repository', 'data', '0', '5'), 'r+b') as fd:
+            # XXX problem: if the last and only commit is corrupted, we have to diagnose "beyond repair"
```
ThomasWaldmann: if borg check does not find any commit tag, it currently tells "this is beyond repair".

src/borg/repository.py
```diff
@@ -675,6 +677,8 @@ def complete_xfer(intermediate=True):
             nonlocal unused
             # commit the new, compact, used segments
             segment = self.io.write_commit(intermediate=intermediate)
+            self.segments.setdefault(segment, 0)
+            self.compact[segment] += LoggedIO.header_fmt.size
```
ThomasWaldmann: how about only doing this for intermediate commits and not registering final commits for compaction (neither here, nor in commit() method)?

that might avoid the "not any commit found, beyond repair" issue the tests are currently stumbling over.
and also it would make the repo more robust, while reducing leftover commit segment clutter.

ThomasWaldmann: ^ that wouldn't help against the "lots of 17-byte files issue" as the intermediate commits are within the current segment file and only the final ones are separate.

ThomasWaldmann:
Decided to combine this with the "separate compaction" ticket #2195 as this make sense for 2 reasons:

- it needs similar adjustments to the repo tests, so they can be just done once

- if compacting after write ops is not the default behaviour any more AND we track the commits for compaction (to avoid all those 17byte COMMIT-only segment files), we have a more robust repo state as long as `borg compact` has not run:

`xxxxxxxx C yyyyyyyy C zzzzzz C` this is how repo roughly looks after 3 tiny backups: xxx... is the data written for the first backup, C is a commit-only segment. similar for yyy... C and zzz... C.

In case the last commit gets lost or corrupted, we still have the previous commits. This would not be the case if we always compacted immediately, then we would have no commit at all in that case.

- after `borg compact` has run, it would look like this:

`xxxxxxxxyyyyyyyyzzzzzz C`

ThomasWaldmann:
@enkore @textshell any comments?

ThomasWaldmann:
superseded by #3970 (simpler and the extra segment for manifest also works from within compact_segments).

PR #3926, opened 2018-06-24
docs: add link to tempfile module

lumbric:
Linking tempfile docs helps to figure out, which default value might be
used, if TMPDIR is not set.

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -202,7 +202,7 @@ General:
         exceptions is not shown.
         Please only use for good reasons as it makes issues harder to analyze.
     TMPDIR
-        where temporary files are stored (might need a lot of temporary space for some operations)
+        where temporary files are stored (might need a lot of temporary space for some operations), see tempfile_ for details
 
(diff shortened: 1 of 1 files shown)


docs/usage_general.rst.inc
```diff
@@ -261,6 +261,8 @@ Please note:
 
 .. _INI: https://docs.python.org/3/library/logging.config.html#configuration-file-format
 
+.. _tempfile: https://docs.python.org/2/library/tempfile.html
+
```
ThomasWaldmann: please use py 3 docs and if it is specifically about the platform docs, maybe say so and link there?

https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

lumbric: Uh yes, sorry sloppiness on my side. I didn't link to the anchor on purpose, because I saw so many paragraphs speaking about TMPDIR, but actually the link you provided is the right paragraph (and the others are not).

ThomasWaldmann:
thanks!

Issue #3494, opened 2017-12-28
DiffArchiverTestCase.test_basic_functionality randomly fails?


sersorrel:
I've now run the tests locally a couple of times, and this test has failed twice for no apparent reason – if I run the tests again afterwards, it passes. It's now also happened on Travis, [here](https://travis-ci.org/borgbackup/borg/jobs/322675024).

I'm not familiar with pytest, but it looks like the failure occurs on [this](https://github.com/borgbackup/borg/blob/f83e5b7f80e631f69667195cbff50db284d7cf67/src/borg/testsuite/archiver.py#L3502) line of `DiffArchiverTestCase.test_basic_functionality`:

```python
do_asserts(self.cmd('diff', self.repository_location + '::test0', 'test1a'), True)
```

where specifically [this](https://github.com/borgbackup/borg/blob/f83e5b7f80e631f69667195cbff50db284d7cf67/src/borg/testsuite/archiver.py#L3443) line in `do_asserts` is failing:

```python
assert '{} input/file_replaced'.format(change) in output
```

Here's what looks like the important bit of the output from pytest:

```
output = '    +13 B       0 B input/empty\n    +13 B       0 B input/hardlink_contents_changed\n[drwxr-xr-x -> -rwxr-xr-x] inpu...emoved directory   input/dir_removed\nremoved link        input/link_removed\n     +1 B   -1.0 kB input/link_removed\n'
can_compare_ids = True                                                                            
                                                                                                                                       
    def do_asserts(output, can_compare_ids):                                                                                           
        # File contents changed (deleted and replaced with a new file) 
        change = 'B' if can_compare_ids else '{:<19}'.format('modified')    
>       assert '{} input/file_replaced'.format(change) in output                         
E       AssertionError: assert 'B input/file_replaced' in '    +13 B       0 B input/empty\n    +13 B       0 B input/hardlink_contents
_changed\n[drwxr-xr-x -> -rwxr-xr-x] inpu...emoved directory   input/dir_removed\nremoved link        input/link_removed\n     +1 B   -1.0 kB input/link_removed\n'                                               
E        +  where 'B input/file_replaced' = <built-in method format of str object at 0x7fda6ad3aa98>('B')
E        +    where <built-in method format of str object at 0x7fda6ad3aa98> = '{} input/file_replaced'.format
```

I'm running the tests with `fakeroot -u tox`; the last time I had this failure, the tests all passed on Python 3.5, but this one failed on Python 3.6.

---

[...]

PR #3942, opened 2018-06-30
tests: trying to debug test fails on travis, #3494

ThomasWaldmann:
- reordered code a little, so files are touched in different order

let's see if this changes anything.

- create_regular_file: giving size AND contents is not supported

the code did not create a 4096 bytes long file, but just a 1 byte
long file. adding an assertion to avoid unsupported usage.

(cherry picked from commit 2d490aa8c88a56872afb4e173ee552f8d51510a8)


ThomasWaldmann:
obviously it did not fix the sporadic test failure.

i'm merging it nevertheless, to keep master consistent with 1.1-maint.

PR #3943, opened 2018-07-01
wrap msgpack to avoid api troubles

### LINKED ISSUE #3632: msgpack: add use_bin_type everywhere
```
Remote: /usr/lib/python3.6/site-packages/msgpack/__init__.py:47:
FutureWarning: use_bin_type option is not specified.
Default value of the option will be changed in future version.
```

borg 1.1 requires msgpack-python < 0.5.0 now, so is not affected by this future issue (and that requirement is also necessary because msgpack 0.5.0 dropped support for py34 which is still supported by borg 1.1.x).

borg master (1.2) will need the fix for the future and does not support py34 any more.

ThomasWaldmann:
wrap msgpack to avoid future upstream api changes making troubles
or that we would have to globally spoil our code with extra params.

make sure the packing is always with use_bin_type=False,
thus generating "old" msgpack format (as borg always did) from
bytes objects.

make sure the unpacking is always with raw=True,
thus generating bytes objects.

note:

safe unicode encoding/decoding for some kinds of data types is done in Item
class (see item.pyx), so it is enough if we care for bytes objects on the
msgpack level.

also wrap exception handling, so borg code can catch msgpack specific
exceptions even if the upstream msgpack code raises way too generic
exceptions typed Exception, TypeError or ValueError.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -13,8 +13,6 @@ from io import BytesIO
 from itertools import groupby, zip_longest
 from shutil import get_terminal_size
 
-import msgpack
-
 from .logger import create_logger
 
 logger = create_logger()
@@ -39,6 +37,7 @@ from .helpers import StableDict
 from .helpers import bin_to_hex
 from .helpers import safe_ns
 from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
+from .helpers import msgpack
 from .patterns import PathPrefixPattern, FnmatchPattern, IECommand
 from .item import Item, ArchiveItem, ItemDiff
 from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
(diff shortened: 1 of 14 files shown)


src/borg/archiver.py
```diff
@@ -1726,7 +1725,7 @@ def do_debug_dump_manifest(self, args, repository, manifest, key):
 
         data = key.decrypt(None, repository.get(manifest.MANIFEST_ID))
 
-        meta = prepare_dump_dict(msgpack.fallback.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
+        meta = prepare_dump_dict(msgpack.unpackb(data, object_hook=StableDict, unicode_errors='surrogateescape'))
```
ThomasWaldmann: does someone remember why this was explicitly using the fallback implementation?

src/borg/archive.py
```diff
@@ -1474,9 +1462,8 @@ def valid_archive(obj):
                 continue
             try:
                 archive = msgpack.unpackb(data)
-            # Ignore exceptions that might be raised when feeding
-            # msgpack with invalid data
-            except (TypeError, ValueError, StopIteration):
+            # Ignore exceptions that might be raised when feeding msgpack with invalid data
+            except msgpack.UnpackException:
```
methane: `except Exception`

PR #3965, opened 2018-07-07
Update installation instructions for macOS

russelldavis:
- Updated `OS X` to the new `macOS` name
- Fuse 3.x is no longer just a prerelease
- Added instructions to install fuse via homebrew


ThomasWaldmann:
@russelldavis thanks! can you backport the PR to 1.1-maint and 1.0-maint branch?

PR #3976, opened 2018-07-10
xattr: use sys/xattr.h as per setxattr(2)

intelfx:


src/borg/platform/linux.pyx
```diff
@@ -15,7 +15,7 @@ from libc.stdint cimport int64_t
 
 API_VERSION = '1.2_02'
 
-cdef extern from "attr/xattr.h":
+cdef extern from "sys/xattr.h":
```
ThomasWaldmann: for ubuntu, it is only in `attr/xattr.h`.

ThomasWaldmann: ubuntu man page:
```
SETXATTR(2)                                     System calls                                     SETXATTR(2)

NAME
       setxattr, lsetxattr, fsetxattr - set an extended attribute value

SYNOPSIS
       #include <sys/types.h>
       #include <attr/xattr.h>

       int setxattr (const char *path, const char *name,
                       const void *value, size_t size, int flags);
       int lsetxattr (const char *path, const char *name,
                       const void *value, size_t size, int flags);
       int fsetxattr (int filedes, const char *name,
                       const void *value, size_t size, int flags);

DESCRIPTION
       Extended  attributes are name:value pairs associated ...
```

PR #3981, opened 2018-07-15
cache lock: use lock_wait everywhere to fix infinite wait

ThomasWaldmann:
in the absence of a specific lock_wait / timeout value (int, to timeout after waiting n seconds), the default of lock_wait / timeout is None (meaning to never timeout and wait forever).


ThomasWaldmann:
todo: check if this needs a backport to 1.0 and if so, do it.

ThomasWaldmann:
the bug / bugfix does not apply to 1.0, but i backported the improved docstrings.

PR #3982, opened 2018-07-15
Locking: better logging, some asserts (1.1)

ThomasWaldmann:


src/borg/locking.py
```diff
@@ -356,21 +356,14 @@ def acquire(self, exclusive=None, remove=None, sleep=None):
...
                     return  # we are the only one and we keep the lock!
                 # restore the roster state as before (undo the roster change):
                 if remove is not None:
                     self._roster.modify(remove, ADD)
-            except:
-                # avoid orphan lock when an exception happens here, e.g. Ctrl-C!
-                self._lock.release()
-                raise
-            else:
-                self._lock.release()
```
ThomasWaldmann: somehow I thought that the above change is equivalent.

but suddenly a lot of test failures!?

ThomasWaldmann: https://github.com/borgbackup/borg/pull/3982/commits/d74a040a58f5e313281f0e54bf201e4cf5f7d137#diff-63888eb46f44c0967d13c3158d5f8302R112

see there for self._lock (ExclusiveLock) context manager.

src/borg/locking.py
```diff
@@ -356,21 +356,14 @@ def acquire(self, exclusive=None, remove=None, sleep=None):
     def _wait_for_readers_finishing(self, remove, sleep):
         timer = TimeoutTimer(self.timeout, sleep).start()
         while True:
-            self._lock.acquire()
-            try:
+            with self._lock:
```
ThomasWaldmann: this and ...

src/borg/locking.py
ThomasWaldmann: ... that.

somehow I thought that the above change is equivalent.

but suddenly a lot of test failures!?

PR #3987, opened 2018-07-18
Note regarding borg version

margau:
Added a note about the version >=1.1 to FAQ, regarding topic on Mailing List: https://mail.python.org/pipermail/borgbackup/2018q3/001110.html

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -104,6 +104,7 @@ you may need to run ``borg break-lock`` before the next backup. You may use the
 same archive name as in previous attempt or a different one (e.g. if you always
 include the current datetime), it does not matter.
 
+
 Borg always does full single-pass backups, so it will start again
 from the beginning - but it will be much faster, because some of the data was
 already stored into the repo (and is still referenced by the checkpoint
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -122,6 +123,10 @@ really desperate (e.g. if you have no completed backup of that file and you'ld
 rather get a partial file extracted than nothing). You do **not** want to give
 that option under any normal circumstances.
 
+Note that checkpoints inside files are created only since version 1.1,
```
ThomasWaldmann: Not sure if we want to include this into the faq. There are quite some features added or changing between versions and our changelog lists them. There is also a shorter changelog linked from the website.

margau: I think it would be good to have it in FAQ, at least as long as there are outdated versions shipped with Debian etc. - if others have the same problem as I had, probably it's much faster than looking into changelogs.

ThomasWaldmann: ok, let's add it to faq.

2 wishes:
- please remove the empty line you accidentally added.
- can you also do a backport of this change to 1.1-maint branch (that's where 1.1.7 docs will come from, master will get into 1.2, but that might still take a while).

thanks!

PR #4050, opened 2018-08-17
docs: improve BORG_CACHE_DIR documentation

vifo:
Improve wording in documentation of BORG_CACHE_DIR. Include hint about putting cache on fastest device as well as size warning.

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -234,8 +234,10 @@ Directories and files:
     BORG_CONFIG_DIR
         Default to '~/.config/borg'. This directory contains the whole config directories.
     BORG_CACHE_DIR
-        Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
-        of space for dealing with big repositories.
+        Defaults to '~/.cache/borg'. This directory contains data cached during
+        backups and helps speed up subsequent backups. Note that when dealing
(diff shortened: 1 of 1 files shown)


docs/usage_general.rst.inc
```diff
@@ -234,8 +234,10 @@ Directories and files:
     BORG_CONFIG_DIR
         Default to '~/.config/borg'. This directory contains the whole config directories.
     BORG_CACHE_DIR
-        Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
-        of space for dealing with big repositories.
+        Defaults to '~/.cache/borg'. This directory contains data cached during
+        backups and helps speed up subsequent backups. Note that when dealing
+        with large repositories this cache will use a lot of disk space. For
+        best performance consider moving this directory to your fastest device.
```
ThomasWaldmann: Well, this describes the current state, but not the general usage of an application cache dir.

While maybe we currently only use it for storing data created while doing a backup, this does not need to be true in future.

Also, the stuff there is read sequentially, so even if you have your default cache location on HDD, I don't think it is advisable to "optimize" it by putting it on "fastest device" (SSD) - especially not because it can use a lot of disk space and your SSD is maybe much smaller than your HDD.

vifo: Well, I have to agree to all of the above. Request withdrawn.

PR #4063, opened 2018-08-29
Forward the `format_spec` to `datetime` class

audeoudh:
If specified, the argument `format_spec` is forwarded from the method
`OutputTimestamp.__format__` to the method `format_time` method.  Thus,
we are able to specify the format, in a way supported by the class
`datetime`.

The default behaviour of `format_time` is not affected.  The other parts
of code that does not provide a `format_spec` are not affected by the
change.

Also refer audeoudh/borg@5957d713 for history.

### DIFF
--- a/src/borg/helpers/time.py
+++ b/src/borg/helpers/time.py
@@ -89,11 +89,11 @@ def safe_timestamp(item_timestamp_ns):
     return datetime.fromtimestamp(t_ns / 1e9)
 
 
-def format_time(ts: datetime):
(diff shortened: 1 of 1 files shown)


infectormp:
Solve https://github.com/borgbackup/borg/issues/4055

ThomasWaldmann:
thanks! can you also do a backport and PR against 1.1-maint branch? that way it'll get into 1.1.8 later.

audeoudh:
Done at #4067

PR #4067, opened 2018-08-30
Forward the `format_spec` to `datetime` class

audeoudh:
If specified, the argument `format_spec` is forwarded from the method
`OutputTimestamp.__format__` to the method `format_time` method.  Thus,
we are able to specify the format, in a way supported by the class
`datetime`.

The default behaviour of `format_time` is not affected.  The other parts
of code that does not provide a `format_spec` are not affected by the
change.

Refer borgbackup/borg#4063


audeoudh:
Solve #4055

ThomasWaldmann:
thanks!

PR #4080, opened 2018-09-20
read a passphrase from a file descriptor

steelman:
Read a passpharase from a file descriptor specified in the
BORG_PASSPHRASE_FD environment variable.

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -174,6 +174,13 @@ General:
         passphrase should be initially set when initializing an encrypted repo.
         If BORG_PASSPHRASE is also set, it takes precedence.
         See also BORG_NEW_PASSPHRASE.
+    BORG_PASSPHRASE_FD
+        When set, specifies a file descriptor to read a passphrase
+        from. Programs starting borg may choose to open an anonymous pipe
+        and use it to pass a passphrase. This is safer than passing via
+        BORG_PASSPHRASE, because on some systems (e.g. Linux) environment
+        can be examined by other processes.
+        If BORG_PASSPHRASE or BORG_PASSCOMMAND are also set, they take precedence.
     BORG_NEW_PASSPHRASE
         When set, use the value to answer the passphrase question when a **new** passphrase is asked for.
         This variable is checked first. If it is not set, BORG_PASSPHRASE and BORG_PASSCOMMAND will also
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -36,7 +36,7 @@ class NoPassphraseFailure(Error):
 
 
 class PassphraseWrong(Error):
-    """passphrase supplied in BORG_PASSPHRASE or by BORG_PASSCOMMAND is incorrect."""
+    """passphrase supplied in BORG_PASSPHRASE, by BORG_PASSCOMMAND or via BORG_PASSPHRASE_FD is incorrect."""
 
 
 class PasscommandFailure(Error):
@@ -429,6 +429,9 @@ class Passphrase(str):
         passphrase = cls.env_passcommand()
         if passphrase is not None:
             return passphrase
+        passphrase = cls.fd_passphrase()
+        if passphrase is not None:
+            return passphrase
 
     @classmethod
     def env_passcommand(cls, default=None):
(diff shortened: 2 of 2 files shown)


src/borg/crypto/key.py
```diff
@@ -442,6 +445,18 @@ def env_passcommand(cls, default=None):
                 raise PasscommandFailure(e)
             return cls(passphrase.rstrip('\n'))
 
+    @classmethod
+    def fd_passphrase(cls):
+        try:
+            fd = int(os.environ.get('BORG_PASSPHRASE_FD', None))
```
ThomasWaldmann: `None` is the default for the 2nd `.get()` argument, so you do not need to give it.

src/borg/crypto/key.py
```diff
@@ -442,6 +445,18 @@ def env_passcommand(cls, default=None):
...
+        except (ValueError, TypeError):
+            return None
+        with os.fdopen(fd) as f:
+            passphrase = f.read()
+        if passphrase is not None:
```
ThomasWaldmann: passphrase can never be `None` here.

either the `fdopen()` / `read()` blows up with an exception and it'll never get here. (does it make sense to catch it?)

or it works and password will be a `str` (empty or non-empty, we would have to consider both as valid).

src/borg/crypto/key.py
```diff
@@ -442,6 +445,18 @@ def env_passcommand(cls, default=None):
                 raise PasscommandFailure(e)
             return cls(passphrase.rstrip('\n'))
 
+    @classmethod
+    def fd_passphrase(cls):
+        try:
+            fd = int(os.environ.get('BORG_PASSPHRASE_FD', None))
+        except (ValueError, TypeError):
+            return None
+        with os.fdopen(fd) as f:
```
ThomasWaldmann: it is the default, but maybe it would be clearer here to give mode='r' - just to point out we're using text mode (resulting in `str`) and not binary mode (resulting in `bytes`).

docs/man/borg.1
```diff
@@ -358,6 +358,9 @@ passphrase should be initially set when initializing an encrypted repo.
 If BORG_PASSPHRASE is also set, it takes precedence.
 See also BORG_NEW_PASSPHRASE.
 .TP
+.B BORG_PASSPHRASE_FD
+When set, specifies a file descriptor to read a passphrase from. Programs starting borg may choose to open an anonymous pipe and use it to pass a passphrase. This is safer than passing via BORG_PASSPHRASE, because on some systems (e.g. Linux) environment can be examined by other processes.
+.TP
```
ThomasWaldmann: if this was automatically generated, you do not need to include it in the PR.

but you need to include the change you (hopefully) did to the `.rst` or `.py` file which is the stuff one needs to edit.

all docs / man pages / etc. will be autogenerated as part of the release process.

src/borg/crypto/key.py
steelman: Leaving out `if'.

If I set BORG_PASSPHRASE_FD and set up the fd, I don't mind borg blowing up if my setup is invalid. At this point it's more UX than anything else.

src/borg/crypto/key.py
steelman: Done.

src/borg/crypto/key.py
steelman: Done.

docs/man/borg.1
steelman: Moved to docs/usage_general.rst.inc

steelman:
I've tested it by running bash command (bash opens fd number 5 for borg and echos the string into it)
```
BORG_PASSPHRASE_FD=5 borg list ./backup 5<<<"veryveryverysecretphrase"
```
I am not sure how to test it from within the testing framework. `test_passphrase` looks very complicated and there is no test for BORG_PASSCOMMAND. If you give me some hints how to do it, I think I will be able to do it.

docs/man/borg.1
```diff
@@ -357,7 +357,6 @@ It is used when a passphrase is needed to access an encrypted repo as well as wh
 passphrase should be initially set when initializing an encrypted repo.
 If BORG_PASSPHRASE is also set, it takes precedence.
 See also BORG_NEW_PASSPHRASE.
-.TP
```
ThomasWaldmann: there is a unwanted change...

docs/usage_general.rst.inc
```diff
@@ -174,6 +174,13 @@ General:
         passphrase should be initially set when initializing an encrypted repo.
         If BORG_PASSPHRASE is also set, it takes precedence.
         See also BORG_NEW_PASSPHRASE.
+    BORG_PASSPHRASE_FD
+        When set, specifies a file descriptor to read a passphrase
+        from. Programs starting borg may choose to open an anonymous pipe
+        and use it to pass a passphrase. This is safer than passing via
+        BORG_PASSPHRASE, because on some systems (e.g. Linux) environment
+        can be examined by other processes.
+        If BORG_PASSPHRASE BORG_PASSCOMMAND are also set, they take precedence.
```
ThomasWaldmann: if ... or ... are ...

ThomasWaldmann:
thanks!

if you want it to be available in next 1.1.x release, you could do a backport to 1.1-maint branch.

master branch will become 1.2.x at some time in the future.

PR #4083, opened 2018-09-22
read a passphrase from a file descriptor

steelman:
Read a passpharase from a file descriptor specified in the
BORG_PASSPHRASE_FD environment variable.


ThomasWaldmann:
thanks!

PR #4091, opened 2018-10-02
Add argument storage_quota to forced_result. Avoids override by client.

m3nu:
Without this, the client will always override the `storage_quota` set at the server.


ThomasWaldmann:
Oops.

Can you file a bug for this, so it can be referenced from the change log?

ThomasWaldmann:
Thanks!

ThomasWaldmann:
can you also do a PR for 1.1-maint, please?

ThomasWaldmann:
does not apply to 1.0-maint (no storage quotas there).

m3nu:
OK. Adding a PR for 1.1-maint.

Issue #4093, opened 2018-10-02
Client can override server storage-quota


m3nu:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).
1.1.7

#### Operating system (distribution) and version.
Debian 9

#### Hardware / network configuration, and filesystems used.
Ext4

#### How much data is handled by borg?
10MB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
- command run on client: `borg init -e none ssh://$REPO_HOST --storage-quota 1000M`
- forced ssh command on server:`borg serve --restrict-to-repository /srv/repos/nz71ss19/repo --storage-quota 10M --append-only` (server)
- result: backup repo on server is initialized with 1000M instead of 10M. Client was able to override the server quota.

## Describe the problem you're observing.
The repo-config file always gets the quota specified by the client. Quota-enforcement is also as per client-set quota.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.
Run client and server commands (set in ssh `command=`)

#### Include any warning/errors/backtraces from the system logs
No errors

Possible fix by extending `forced_result` as done in #4091

ThomasWaldmann:
For borg hosting providers, this might be a security issue, because users could use unlimited resources, even if the provider set a quota on `borg serve` side (e.g. via forced command in `.ssh/authorized_keys`).

lfam:
Distro packagers should consider cherry-picking this commit for their packages, right?

ThomasWaldmann:
@lfam Yes, this is for 1.1.x: https://github.com/borgbackup/borg/pull/4095/commits/975cc33206e0e3644626fb7204c34d2157715b61

ThomasWaldmann:
Fixed by #4091 and #4095, thanks to @m3nu!

knutov:
@ThomasWaldmann Current 1.1.8 milestone stuck for 15+ days. Is it possible to release binaries with this patch with version named like '1.1.7.1' or '1.1.8' now or near future?

ThomasWaldmann:
I'll release 1.1.8 soon.

knutov:
@ThomasWaldmann How soon is soon? )

PR #4095, opened 2018-10-03
Add storage_quota for forced_result if set by serve-command.

m3nu:
See #4093 and #4091

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4276,6 +4276,7 @@ class Archiver:
             result.restrict_to_paths = forced_result.restrict_to_paths
             result.restrict_to_repositories = forced_result.restrict_to_repositories
             result.append_only = forced_result.append_only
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -4276,6 +4276,7 @@ def get_args(self, argv, cmd):
             result.restrict_to_paths = forced_result.restrict_to_paths
             result.restrict_to_repositories = forced_result.restrict_to_repositories
             result.append_only = forced_result.append_only
+            result.storage_quota = forced_result.storage_quota
```
enkore: Fail open vs. fail closed -- instead of specifying what options are blacklisted (i.e. not overriden), consider whitelisting instead.

step21: Sounds great. I am just not sure yet what is the closed and what the open part in fail close/open with regard to this. As only expected options are explicitly specified, isn't this already whitelisting?

Issue #3561, opened 2018-01-19
add repo name to borg create output


roadrunnerjb:
Hi,

I'm using borg to backup my systems via batch script to more than one repository and I'm missing the repo name in the output to be able to identify what repo returns an error. It should be possible to add the repo name passed as parameter to the output of the create statement.

Regards
Joachim

ThomasWaldmann:
Did you try something like this in your script?
```
echo Using repo $BORG_REPO 1>&2
```

roadrunnerjb:
Hello Thomas,

thank you for the fast reply.

Your suggestion was my workaround, but why do it in a script... So please treat it as a feature request.

Regards Joachim

ThomasWaldmann:
OK, let's discuss this:

shall we add something like --show-repo to make borg print the repo url at the start of backup log output?

roadrunnerjb:
It's up to you but why adding another cmdline switch ? borg create reports so many things, why not adding the repo name too ?

Regards Joachim

infectormp:
for future diagnostic i add repo name to log using `echo` so +1 for adding to standard output

lightonflux:
When pushing logs to a central server it get quite confusing. So having the repo name is a good idea. 

I too prefer having it as a default without additional/needed cli switches.

BTW: @ThomasWaldmann do you have a vision how borgs OOTB experience should be. Is it more ready to run, or more a toolkit to make your own backup script? I prefer the former, as backups should always as easy as possible.

ThomasWaldmann:
@lightonflux the idea is that borg can do everything that is useful and in its scope of tasks and can be easily implemented for the supported platforms. otoh, we try to avoid "in-tool scripting" because you'll usually need a script that calls borg anyway, so we do not / should not implement stuff that can be just called in such a script before/after borg runs.

jikkujj:
First timer here. Has this been worked on? If not I would like to pick this up

ThomasWaldmann:
Sure, feel free.

Maybe first have some thoughts about what makes sense / is most useful and how it should be done.

Then do a PR to discuss the code.

patelsv:
Another first timer. Since this is still open, I'd like to work on this. 

After a quick check, I think it'd make sense to have a stderr logging output where the repository is being created in remote.py and repository.py files.

If someone can corroborate that it is an appropriate approach then I can do a PR against the master branch to discuss further. Thank you!

ThomasWaldmann:
[...]

PR #4105, opened 2018-10-09
Added output for create function. Solves #3561

JoshuaRM:
Solved issue #3561.
The create function now logs the name of the repository to std.err
The output logs the message "Repository x has been created successfully". Where x is the repo name

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -534,7 +534,7 @@ class Archiver:
                     process_file_chunks=cp.process_file_chunks, add_item=archive.add_item,
                     chunker_params=args.chunker_params, show_progress=args.progress)
                 create_inner(archive, cache, fso)
-                logger.error("Directory %s has been created successfully" % archive.name)
+                logger.error("Repository %s has been created successfully" % archive.name)
         else:
             create_inner(None, None, None)
         return self.exit_code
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -534,6 +534,7 @@ def create_inner(archive, cache, fso):
                     process_file_chunks=cp.process_file_chunks, add_item=archive.add_item,
                     chunker_params=args.chunker_params, show_progress=args.progress)
                 create_inner(archive, cache, fso)
+                logger.error("Repository %s has been created successfully" % archive.name)
```
ThomasWaldmann: Why output it on error level? Check the available logging levels to find something more appropriate.

Also, you are saying "Repository %s", but what you fill into the placeholder is the archive's name, not the repository URL.

src/borg/archiver.py
```diff
@@ -534,6 +534,7 @@ def create_inner(archive, cache, fso):
...
         else:
             create_inner(None, None, None)
```
ThomasWaldmann: in this branch, there will be no output with your modification. but guess there should be.

dry-run is to simulate a backup run without really writing something, the messages can be the same though.

also maybe move it to before the recursion is started and say something like "Creating archive ... in repository ....".
then we will have that output even in case the backup is aborted later. the final success state is available via --show-rc.

ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/builds/439192463?utm_source=github_status&utm_medium=notification

there is some breakage in the tests, see above.

PR #4120, opened 2018-10-22
Allow borg with-lock children to grab locks on same repository.

vctgross:
This PR adds a `--shared-lock` option to `borg with-lock`, and allows borg children of `borg with-lock --shared-lock` to grab an exclusive or shared lock.
  
I have a server that hosts all my borg repositories, and the only access allowed is SSH with `borg server --append-only`. I want to run a daily cron on this server that runs borg prune only if no archive has been deleted since the last borg prune; in other words, the borg prune cron is the only way to delete archives. It could be easy enough to do :

`export BORG_REPO`
`borg with-lock $BORG_REPO /bin/bash -c 'is_subset "$(< $REPO_ARCHIVE_REFFILE)" "$(borg list)" && borg prune && borg list > $REPO_ARCHIVE_REFFILE'`

except that borg with-lock takes an exclusive lock on the repository.

This PR address this issue by allowing to downgrade borg with-lock to a shared lock, and by allowing a borg process to take an exclusive lock on top of a shared one. This is subject to restrictions : there can be at most one process holding the shared lock on the repository, and this process' PID must be equal to our PGID.
Because forked process inherit their parent PGID, and because shells with job control (set -m) enabled spawns jobs ("a set of processes, comprising a shell pipline" [Open Group Base Definitions 3.202]) in their own process groups, this means that, assuming the shared lock holder PGID equals its PID, without additional actions (i.e. setpgid(2)) only children of the shared lock holder will meet the conditions.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1545,7 +1545,7 @@ class Archiver:
                 self.print_error('Running borg with-lock using shared locks is supported only for local repositories.')
                 return self.exit_code
         env = prepare_subprocess_env(system=True)
-        os.setpgid(0,0)
+        os.setpgid(0, 0)
         # we exit with the return code we get from the subprocess
         return subprocess.call([args.command] + args.args, env=env)
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1530,18 +1530,24 @@ class Archiver:
         # we can only do this for local repositories (with .io), though:
         if hasattr(repository, 'io'):
             repository.io.close_segment()
+        # we need to commit the "no change" operation we did to the manifest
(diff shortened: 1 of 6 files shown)


textshell:
I think the restriction to local repositories is strange and likely an artifact of your specific requirements.
For determining which processes may access i would prefer a environment variable with some kind of token so more complex usage scenarios can be supported.
Should the borg commands that share the lock require a "--with-shared-lock" option or is that overkill?

ThomasWaldmann:
There were quite big changes in #8332 to the repository and locking code, so this does conflict a lot and in any case does not apply any more, because the locking code is now completely different.

A lot of borg commands do not take an exclusive lock anymore, exceptions are check and compact.

Issue #2677, opened 2017-06-14
hashindex: clean void* arithmetic up


enkore:

ThomasWaldmann:
this ticket is not quite clear, what / how would you want to "clean up"?

enkore:
void* arithmetic is illegal C. User char* arithmetic instead.

enkore:
Fixed by #4127

PR #4127, opened 2018-10-24
c code portability fixes: issue #2677 and 2 more

motwok:
Hello,

- As already discussed in issue #2677 pointer aritmetic with "void*" is undefined. Many c compilers reject this code.
solved issue(s):
-- hashindex: clean void* arithmetic up  #2677
-- lowlevel: cleanup void* arithmetic (not eported yet)
I used "unsigned char*" instead of "char*" because it is used many times in lowlevel.c.

- "false" is not standard C, the other code uses 0 or 1
solved issue(s):
-- unpack: replace non standard "false" with 0 (not reported yet)

This is one step to make it possible to build borg also on non GCC platforms.
(My  impulse is to make it work with native Python on Windows, this needs a few more changes to make it compile and many, many more to make it work)

Emmo


src/borg/_hashindex.c
```diff
@@ -355,7 +355,7 @@ hashindex_read(PyObject *file_py, int permit_compact)
     index->num_buckets = _le32toh(header->num_buckets);
     index->key_size = header->key_size;
     index->value_size = header->value_size;
-    index->bucket_size = index->key_size + index->value_size;
+    index->bucket_size = index->key_size + index->value_size;
```
ThomasWaldmann: accidental change

ThomasWaldmann:
If you like, you could also rebase on current master, so the flake8 issue goes away.

ThomasWaldmann:
If you'ld like this to get merged today, we could also merge "as is" (but please fix the accidental change) and do any further work in a new PR.

motwok:
This can be merged now (accidental change reversed).
There should be some oberhauling of the hashindex parts, see #4145.

src/borg/_hashindex.c
motwok: Fixed

ThomasWaldmann:
thanks!

PR #4141, opened 2018-10-31
Moving the flags of the compiled regex to the front to fix #4137

### LINKED ISSUE #4137: Flags not at the start of the expression
```
.tox/py36/lib/python3.6/site-packages/borg/patterns.py:281
  /home/travis/build/borgbackup/borg/.tox/py36/lib/python3.6/site-packages/borg/patterns.py:281: DeprecationWarning: Flags not at the start of the expression 'b\\á\\/(?:[^\\/]*\\/)*[^' (truncated)
    self.regex = re.compile(shellpattern.translate(self.pattern))
```

step21:
As the title says, after setting up my environment and the tests, I think I fixed this one. All tests seem to pass. So unless there is some regex magic that I missed it should be fine.
As far as I can tell, I complied with all the contribution guidelines.

### DIFF
--- a/src/borg/shellpattern.py
+++ b/src/borg/shellpattern.py
@@ -62,4 +62,4 @@ def translate(pat, match_end=r"\Z"):
         else:
             res += re.escape(c)
 
-    return res + match_end + "(?ms)"
+    return "(?ms)" + res + match_end


step21:
Oh yeah, except there is this weird failure in python 3.6, but probably that was not me?

ThomasWaldmann:
Thanks for the PR, looks good! Did you check whether our 1.1-maint branch also needs that fix?

ThomasWaldmann:
about the travis test fails: sadly, the tests sometimes fail due to still unknown reasons (we have a ticket about it, some fails might be related to fakeroot). restarting the test will make it succeed, sooner or later...

step21:
No, I didn't check yet whether the 1.1 branch needs this as well. I can though.
Thanks for the info about the tests.

step21:
Yes, the 1.1-maint branch also needs this fix. What is the best procedure for that?

ThomasWaldmann:
checkout the 1.1-maint branch and cherrypick your change into there, then make a PR against 1.1-maint.

Issue #4136, opened 2018-10-29
Environment variable ... converted to str implicitly


ThomasWaldmann:
```
.tox/py36/lib/python3.6/site-packages/_pytest/monkeypatch.py:234
  /home/travis/build/borgbackup/borg/.tox/py36/lib/python3.6/site-packages/_pytest/monkeypatch.py:234:
    PytestWarning: Environment variable value local('/tmp/pytest-of-travis/pytest-0/popen-gw2/xdg-config-home0') should be str, converted to str implicitly
    value
```
a flood of these is seen on travis-ci sometimes.

step21:
The first one I found, but any specific command to see where exactly these variables originate? Somehow I find them hard to find in the code.

ThomasWaldmann:
I currently only have what I posted above. Seen a flood of these in the test runs. From the sheer amount, I suspect it might be the test setup code (iirc in conftest.py).

ThomasWaldmann:
maybe it helps to temporarily modify `monkeypatch` to just raise an Exception there instead of a warning, so one gets more information about where it goes wrong.

ThomasWaldmann:
fixed by #4142.

btw, ending your PR or commit comment with ", fixes #nnnn" would close the issue nnnn automatically.

step21:
Yeah, sorry, I forgot to add the #id here.

PR #4142, opened 2018-10-31
Explicitly convert environment variable to str to address #4136

step21:
After running the tests again locally, it seems I had already found all relevant places. Assuming tests pass.


step21:
1.1-maint branch could also use this fix.

ThomasWaldmann:
I also had a quick look into the src, looks like you've found all the places missing the str().

Thanks!

For 1.1, please make a separate PR.

Issue #4140, opened 2018-10-30
coala fails on broken link to cachedir spec


step21:
Would like to work on some 'easy' issues and trying to get all the tests running. Tox fails with some permission problems on ~/.cache but that is owned by my user (which might be a local problem) and/but coala gives 3 errors at the start with not being able to connect to `Broken link - unable to connect to http://www.brynosaurus.com/cachedir/spec.html (HTTP Error: 503)` When you check that link, it says something about yale university network having blocked this for malware or something like that.

Is this normal, or should I not worry about this for now?

ThomasWaldmann:
If you like you could talk to the yale IT guys, i think this is an obvious error - if somebody publishes a specification for a public standard, readers should not be blocked for nonsense reasons.

razalot:
I also encountered the dead link (http://www.brynosaurus.com/cachedir/spec.html) in the borg documentation. It's under https://borgbackup.readthedocs.io/en/stable/usage/create.html where the option --exclude-caches is explained.

After some googling I found:
http://www.bford.info/cachedir/spec.html
http://www.bford.info/cachedir/

These websites are from the same author.
He probably has moved his personal homepage http://www.brynosaurus.com/ from Yale to EPF Lausanne.

ThomasWaldmann:
OK, so a PR for updating the links would be nice. Be aware that some of the docs get auto-generated from borg code (archiver.py).

PR #4143, opened 2018-10-31
 Fixed links to cachedir spec in docs and code to address #4140 

step21:
As far as I am aware, all other instances of the link should be derived/created for doc/man/etc.


ThomasWaldmann:
thanks! guess a PR for 1.1-maint is also needed, same issue there.

Issue #4146, opened 2018-11-02
Basic MSC Compatibility


motwok:
Use __MSC makros and others to make c Code MSC compatible for native Windows compile.
Most Code is compatible for the compiler, but not all.

After this borg compiles on Windows.
There are more changes needed to run on native Windows.

ThomasWaldmann:
fixed by #4147.

PR #4147, opened 2018-11-02
Basic MSC Compatibility

motwok:
see #4146

### DIFF
--- a/src/borg/_endian.h
+++ b/src/borg/_endian.h
@@ -4,22 +4,22 @@
 #include <stdlib.h>
 #include <stdint.h>
 
- #if defined (__SVR4) && defined (__sun)
- #include <sys/isa_defs.h>
- #endif
- 
- #if (defined(BYTE_ORDER) && defined(BIG_ENDIAN) && (BYTE_ORDER == BIG_ENDIAN)) ||  \
-     (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)) || \
-     (defined(_BIG_ENDIAN) && defined(__SVR4) && defined(__sun))
- #define BORG_BIG_ENDIAN 1
- #elif (defined(BYTE_ORDER) && defined(LITTLE_ENDIAN) && (BYTE_ORDER == LITTLE_ENDIAN)) || \
-       (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)) || \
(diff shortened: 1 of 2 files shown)


src/borg/_endian.h
```diff
@@ -1,22 +1,25 @@
...
+     (defined(_BIG_ENDIAN) && defined(__SVR4) && defined(__sun))
+ #define BORG_BIG_ENDIAN 1
+ #elif (defined(BYTE_ORDER) && defined(LITTLE_ENDIAN) && (BYTE_ORDER == LITTLE_ENDIAN)) || \
+       (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)) || \
+       (defined(_LITTLE_ENDIAN) && defined(__SVR4) && defined(__sun)) || \
+       (defined(_MSC_VER) && (defined(_M_AMD64) || defined(_M_IX86)))
+ #define BORG_BIG_ENDIAN 0
+ #else
+ #error Unknown byte order
+ #endif
```
ThomasWaldmann: why is all this indented?

motwok: Looks like diff tool takes to much, should look like this (thats the changed/added code):
--       (defined(_LITTLE_ENDIAN) && defined(__SVR4) && defined(__sun))
++     (defined(_LITTLE_ENDIAN) && defined(__SVR4) && defined(__sun)) || \ 
++     (defined(_MSC_VER) && (defined(_M_AMD64) || defined(_M_IX86)))

ThomasWaldmann: isn't there an additional blank or tab at the start of these lines?

motwok: Of corse, there are blanks… but on purpose.
This is adapted from from existing code.
removing the 2 blanks in the rows following the #elif would honor a "4 Blanks" Rule but make the code unreadable.

If you think that is a problem, i can remove this 3 Lines (and change 2 Additional lines).

ThomasWaldmann: I don't understand.

What I mean is that it looks like e.g. old line 5 (new line 7) was additionally indented by your changeset.
Same for ALL the lines showing up in red/green above.

If all that stuff was nested in something else I would understand that change, but there is no such nesting.

Ehrm, did you accidentally change these lines from lf-only to crlf?

motwok: Now is see… i have no clue where this blanks came from, and i didn't see them until now
about LF vs CRLF: Strange enougth, when i open completly untouched files binary, i have CRLF, may git converts those automatically.
May you are right that there is some issue with the conversion because VS told me about mixed line endings, i ignored this to NOT change more than needed.

hope this is fixed now

ThomasWaldmann:
thanks!

Issue #4148, opened 2018-11-02
Platform independency, redirect imports from platform.posix to platform


motwok:
All direct imports of `platform.posix` (or other modules) violating platform independency.

These imports have to be rerouted via platform module.

Remarks:

Methods in `posix.pyx` are not all posix... for example windows is also posix... maybe "posix" should be renamed to "linux" (also wrong because of darwin and cygwin, maybe we find a better name).

PR #4153, opened 2018-11-05
Plattform independency, redirect imports from plattform.posix to platform #4148

motwok:
fixes issue #4148 partly. (one small step)


motwok:
looks like a rebase fail… will try again

PR #4157, opened 2018-11-06
Added PATH::archive output for creation function on info log level.

### LINKED ISSUE #3561: add repo name to borg create output

Ashmodei:
Fixes #3561
"PATH::archive" output is added onto info log level into do_create() function.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -446,6 +446,8 @@ class Archiver:
         matcher = PatternMatcher(fallback=True)
         matcher.add_inclexcl(args.patterns)
 
+        logger.info('Creating archive at "%s"' % args.location.orig)
+
         def create_inner(archive, cache, fso):
             # Add cache dir to inode_skip list
             skip_inodes = set()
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -446,6 +446,8 @@ def do_create(self, args, repository, manifest=None, key=None):
         matcher = PatternMatcher(fallback=True)
         matcher.add_inclexcl(args.patterns)
 
+        logger.info('Creating archive at "%s"' % args.location.orig)
```
ThomasWaldmann: would work like this, but maybe better move it to before `if not dry_run:` in line ~520.

Ashmodei:
I've changed this to the way you proposed, it really looks more the same like other logger's invocations.

ThomasWaldmann:
thanks! can you apply the same change to 1.1-maint with another PR?

PR #4164, opened 2018-11-10
remove posix issues and add fixups for unsupported methodes. (ex. #4154)

motwok:
Part of #4148
Please doublecheck on review


src/borg/platformflags.py
```diff
@@ -0,0 +1,12 @@
+"""
+Flags for Platform-specific APIs.
+
+Use this Flags instead of sys.platform.startswith('<OS>') and try except
```
ThomasWaldmann: ... `these` ... or try`/`except`.`

src/borg/platform/posix.pyx
```diff
@@ -61,3 +64,55 @@ def local_pid_alive(pid):
...
+            if len(fields) == 4:
+                entries.append(':'.join([fields[0], fields[3], fields[2]]))
+            else:
+                entries.append(entry)
+    return safe_encode('\n'.join(entries))
+
+
+def getosusername():
+    """Return the os user name."""
+    return uid2user(os.getuid())
```
ThomasWaldmann: was: `uid2user(os.getuid(), os.getuid())`

guess this was an accidental change, please use same semantics as before, but optimize:
```
    uid = os.getuid()
    return uid2user(uid, uid)
```

src/borg/platformflags.py
motwok: fixed with "recommit"

src/borg/platform/posix.pyx
motwok: fixed with new commit (rebase into b997d5ba5bffa0cf4a34efcd902c218b88fbd4d1 will destroy commit order in PR)

PR #4171, opened 2018-11-13
fix EISDIR issues

ThomasWaldmann:


ThomasWaldmann:
i restarted travis for about 10 times now and EISDIR did not happen, so i guess getting rid of fakeroot is the solution (at least on travis).

PR #4226, opened 2018-12-20
Docs: Improve logging in example automation script

greggrossmeier:
* Add an info entry for the compacting step
* Add a final info entry for when no errors occurred

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -230,19 +230,12 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     global_exit=$(( backup_exit > prune_exit ? backup_exit : prune_exit ))
     global_exit=$(( compact_exit > global_exit ? compact_exit : global_exit ))
 
-    if [ ${global_exit} -eq 1 ];
-    then
-        info "Backup, Prune and/or Compact finished with a warning"
-    fi
-
-    if [ ${global_exit} -gt 1 ];
-    then
-        info "Backup, Prune and/or Compact finished with an error"
-    fi
-    
-    if [ ${global_exit} -eq 0 ];
-    then
-        info "Backup, Prune, and Compact finished without error"
+    if [ ${global_exit} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -237,7 +239,12 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     then
         info "Backup, Prune and/or Compact finished with an error"
     fi
-
+    
+    if [ ${global_exit} -lt 1 ];
```
ThomasWaldmann: guess this should rather check for rc == 0.

greggrossmeier: Good point.

docs/quickstart.rst
```diff
@@ -237,7 +239,12 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
     then
         info "Backup, Prune and/or Compact finished with an error"
     fi
-
+    
+    if [ ${global_exit} -eq 0 ];
+    then
+        info "Backup, Prune, and Compact finished without error"
+    fi
+
```
ThomasWaldmann: sorry to bother again, but i guess this whole check (not only your new code) should be done like this (please translate into shell syntax and also please test it):

```
if rc == 0:
    ... finished successfully   # do not use the word error here
else if rc == 1:
    ... finished with warnings.  # plural for generic case: 1 or more
else:
    ... finished with errors.  # plural for generic case: 1 or more
```

ThomasWaldmann: btw, this pull request is against master branch, so it will get into 1.2 when it will be released in the future.

if you'ld like to get it also into 1.1.9, can you please also file a PR against 1.1-maint branch (after the master PR is merged)?

greggrossmeier:
Thanks for the thoughtful review and guidance. Updated PR. Tested locally by running a normal run of the script (successful), then adding in /etc while running as a non-privileged user (so warnings), then removing all PATHs (causing a borg error). Worked as expected.

ThomasWaldmann:
thanks!

Issue #4234, opened 2018-12-31
Clarify the necessity to separately backup key in repokey mode


cornop:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Question / request to improve documentation

## Describe the problem you're observing.

`borg init --encryption=repokey-blake2` states:

```
IMPORTANT: you will need both KEY AND PASSPHRASE to access this repo!
Use "borg key export" to export the key, optionally in printable format.
Write down the passphrase. Store both at safe place(s).
```

while `borg init --help` says:

```
If you want "passphrase-only" security, use one of the repokey modes. The
key will be stored inside the repository (in its "config" file). In above
mentioned attack scenario, the attacker will have the key (but not the
passphrase).

[...]

Make a backup copy of the key file (keyfile mode) or repo config file
(repokey mode) and keep it at a safe place, so you still have the key in
case it gets corrupted or lost.
```

I find these contradicting. Do I need to export and separately store the key, or do I only need it in case the config gets corrupted?

ThomasWaldmann:
What exactly do you find contradicting?

You want to backup the key, no matter whether you have it in the repo directory or in your home directory.

cornop:
I know that I *want* to backup the key. The questions is whether I need the key to restore my backup (as opposed to needing the key only in case of corruption). `borg init` says I do, while the manual says I don't.

ThomasWaldmann:
That you "need the key" for about every operation on an encrypted repo does not mean you explicitly have to give it to borg (like e.g. by specifying its path/filename).

It usually is discovered from either:
- the repository config file (`repokey` modes): the encrypted key is in `repository_dir/config` file
- a key file (`keyfile` modes): the encrypted key is (usually) in `~/.config/borg/keys/*` file

In both cases, the encrypted key is automatically read from there and gets decrypted using the passphrase (which is either queried from you interactively or read from environment or read from the output of some command or via a file descriptor).

ThomasWaldmann:
So I'm not sure how to formulate the docs less confusing, maybe make a pull request to improve it.

PR #4237, opened 2019-01-04
Clarify the necessity to separately backup key in repokey mode

cornop:
See https://github.com/borgbackup/borg/issues/4234.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -262,8 +262,9 @@ class Archiver:
                 logger.warning(
                     '\n'
                     'IMPORTANT: you will need both KEY AND PASSPHRASE to access this repo!\n'
-                    'The key is already included in the repository config.\n'
-                    'Write down the passphrase. Store it at safe place(s).\n')
+                    'The key is included in the repository config, but should be backed up in case the repository gets corrupted.\n'
+                    'Use "borg key export" to export the key, optionally in printable format.\n'
+                    'Write down the passphrase. Store both at safe place(s).\n')
             else:
                 logger.warning(
                     '\n'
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -258,11 +258,18 @@ def do_init(self, args, repository):
...
-            logger.warning(
-                '\n'
-                'IMPORTANT: you will need both KEY AND PASSPHRASE to access this repo!\n'
-                'Use "borg key export" to export the key, optionally in printable format.\n'
-                'Write down the passphrase. Store both at safe place(s).\n')
+            if 'repokey' in key.NAME:
+                logger.warning(
+                    '\n'
+                    'IMPORTANT: you will need both KEY AND PASSPHRASE to access this repo!\n'
+                    'The key is already included in the repository config.\n'
```
ThomasWaldmann: as long as nothing goes wrong in the repo config, the key is there.

but it would be somehow ridiculous, if the repo config gets corrupted (hw or sw error), and you lose access to all these years of backups you've made, just because you did not make a backup of the key.

so, i do not see it as an improvement NOT to recommend "borg key export" here.

cornop: Makes sense.

Just a thought: That's something that could be improved on the side of borg (for example, by storing the key multiple times in the repo).

ThomasWaldmann:
thanks. this is in master branch now (and at some time in 1.2 release).

would you like to backport it to 1.1-maint branch also?

cornop:
> would you like to backport it to 1.1-maint branch also?

Nope, this being in in the future works for me.

PR #4238, opened 2019-01-04
implement a fixed block size chunker, fixes #1086

### LINKED ISSUE #1086: fixed record size chunker?
Maybe it is useful to add a fixed record size chunker additionally to the content-defined chunker to get better deduplication for that kind of data.

E.g. if a database file has fixed record size RS and an optional header with size HS, we could chunk like this:

```
HS | RS | RS | RS | ...
```

Of course one needs to be careful to not generate too many chunks or resource usage to manage them will go through the roof (so instead of chunking every RS, a multiple of RS might save some resources).

ThomasWaldmann:
also:
- bump chunker API_VERSION
- prepend algorithm specification to chunker_params
- make ChunkerParams parsing more flexible

compatibility:
- chunker_params stored in archive metadata now start with the
  algorithm, this makes them incompatible with older borg versions
  even if the standard buzhash algorithm is used.
- chunker_params given on the commandline are still understood in
  the old format, internally defaulting to buzhash then.
- chunker_params in old archives are understood for borg recreate, avoiding rechunking if unnecessary

### DIFF
--- a/src/borg/chunker.pyx
+++ b/src/borg/chunker.pyx
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
-API_VERSION = '1.1_02'
+API_VERSION = '1.1_03'
+
+import os
 
 from libc.stdlib cimport free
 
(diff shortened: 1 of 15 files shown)


src/borg/chunker.pyx
```diff
@@ -17,7 +19,77 @@ cdef extern from "_chunker.c":
...
+        offset = 0
+        use_fh = fh >= 0
+
+        if use_fh:
+            def read(size):
+                nonlocal offset
+                data = os.read(fh, size)
+                amount = len(data)
+                if hasattr(os, 'posix_fadvise'):
+                    # UNIX only and requires bug fixed linux kernel > 4.6.0, see comment in _chunker.c.
```
ThomasWaldmann: The other, standard chunker has a workaround in its C implementation.

Not sure if we need to make this implementation more complicated also or just document that we need linux kernel > 4.6.0 for this fixed blocksize chunker.

For some applications, we would use some 2^N (with a not too small N) block_size and 0 header_size with this, so maybe that fadvise bug does not even bite here.

src/borg/item.pyx
```diff
@@ -325,6 +325,21 @@ class Key(PropDict):
...
 
 
+def mft_encode(t):
+    """encode a manifest tuple"""
+    return t
+
+
+def mft_decode(t):
+    """decode a manifest tuple"""
+    # we get bytes objects from msgpack.unpack, but want str
```
ThomasWaldmann: msgpack legacy not roundtripping py3 str correctly requires this.

ThomasWaldmann: without this, json mode output encode falls over the bytes object returned by msgpack.

ThomasWaldmann:
Looking at the recreate test issue with the different archive-id now... (2 failing tests).

Issue #4094, opened 2018-10-03
1.0-maint testing broken on travis-ci


ThomasWaldmann:
see there: https://github.com/borgbackup/borg/pull/4084

https://travis-ci.org/borgbackup/borg/builds/431821410?utm_source=github_status&utm_medium=notification

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/64227624-1-0-maint-testing-broken-on-travis-ci)

ThomasWaldmann:
i'm working on this.

ThomasWaldmann:
fixed by #4099.

TODO: talk to tox developers about the tox regression / incompatibility - i didn't see anything in the changelog.

ThomasWaldmann:
more experiments: #4253

PR #4253, opened 2019-01-13
experiments related to #4094

ThomasWaldmann:
#4094 was worked around by requiring `tox<3.2.0`, because back then, newer versions broke the tests.

trying again with the latest tox version: import issues still there, see linux tests on travis.


ThomasWaldmann:
considering that this is 1.0-maint only and I am EOLing 1.0-maint soon anyway (see #4352) and we have a workaround for now, I am closing this.

PR #4269, opened 2019-01-26
Clarify explanation of encryption options

chrisrapson:
Text has been replaced, added or re-organised to try to improve the explanation of encryption options.

Some changes also corrected what appeared to be errors, e.g. "none", "repokey" and "keyfile" were marked as **non**-backwards-compatible modes in places.

I think I still do not understand authentication without encryption. What is the purpose of it, and how does it work? I have not added any information regarding this option, but it would probably be worthwhile adding something in a future update.

See [issue #1001](https://github.com/borgbackup/borg/pull/1001) for preliminary discussion leading to this PR.

Sorry about the merge commit in my fork. I can remove that if you want, but I assume it's easier for you to squash it on your end.

### DIFF
--- a/docs/man/borg-init.1
+++ b/docs/man/borg-init.1
@@ -43,17 +43,12 @@ Encryption can be enabled at repository init time. It cannot be changed later.
 It is not recommended to work without encryption. Repository encryption protects
 you e.g. against the case that an attacker has access to your backup repository.
 .sp
-But be careful with the key / the passphrase:
-.sp
-If you want "passphrase\-only" security, use one of the repokey modes. The
-key will be stored inside the repository (in its "config" file). In above
-mentioned attack scenario, the attacker will have the key (but not the
-passphrase).
-.sp
-If you want "passphrase and having\-the\-key" security, use one of the keyfile
-modes. The key will be stored in your home directory (in .config/borg/keys).
-In the attack scenario, the attacker who has just access to your repo won\(aqt
-have the key (and also not the passphrase).
+The repository is encrypted with a key, and the key is protected with a passphrase. 
+By setting the encryption mode, you can choose whether to store the key inside the
+repository ("passphrase only" or "repokey" mode) or to store the key separately in 
+.config/borg/keys ("passphrase and having\-the\-key" or "keyfile" mode). In repokey 
+mode, an attacker with access to your repository needs to know (or guess) the 
+passphrase. In keyfile mode, they will need the keyfile \fBand\fP the passphrase.
 .sp
 Make a backup copy of the key file (keyfile mode) or repo config file
 (repokey mode) and keep it at a safe place, so you still have the key in
@@ -81,6 +76,14 @@ a different keyboard layout.
 You can change your passphrase for existing repos at any time, it won\(aqt affect
 the encryption/decryption key or other secrets.
 .SS Encryption modes
+.sp
+For each repo, the user may select the degree of encryption, and the Hash/MAC 
+algorithm. The table below shows what string to enter, to obtain each combination
+of encryption degree and Hash/MAC algorithm.
+.sp
+Not all modes are backwards compatible with Borg1.0.x. New modes which are not 
+backwards compatible are \fImarked like this\fP 
+.sp
 .\" nanorst: inline-fill
 .
 .TS
@@ -126,46 +129,46 @@ _
 .\" nanorst: inline-replace
 .
 .sp
-\fIMarked modes\fP are new in Borg 1.1 and are not backwards\-compatible with Borg 1.0.x.
-.sp
+How should a user choose between SHA-256 and Blake2b? The main difference is execution speed. 
 On modern Intel/AMD CPUs (except very cheap ones), AES is usually
 hardware\-accelerated.
 BLAKE2b is faster than SHA256 on Intel/AMD 64\-bit CPUs
 (except AMD Ryzen and future CPUs with SHA extensions),
-which makes \fIauthenticated\-blake2\fP faster than \fInone\fP and \fIauthenticated\fP\&.
+which makes \fIauthenticated\-blake2\fP faster than none and \fIauthenticated\fP\&.
 .sp
 On modern ARM CPUs, NEON provides hardware acceleration for SHA256 making it faster
 than BLAKE2b\-256 there. NEON accelerates AES as well.
 .sp
 Hardware acceleration is always used automatically when available.
 .sp
-\fIrepokey\fP and \fIkeyfile\fP use AES\-CTR\-256 for encryption and HMAC\-SHA256 for
+repokey and keyfile use AES\-CTR\-256 for encryption and HMAC\-SHA256 for
 authentication in an encrypt\-then\-MAC (EtM) construction. The chunk ID hash
 is HMAC\-SHA256 as well (with a separate key).
-These modes are compatible with Borg 1.0.x.
 .sp
 \fIrepokey\-blake2\fP and \fIkeyfile\-blake2\fP are also authenticated encryption modes,
 but use BLAKE2b\-256 instead of HMAC\-SHA256 for authentication. The chunk ID
 hash is a keyed BLAKE2b\-256 hash.
-These modes are new and \fInot\fP compatible with Borg 1.0.x.
 .sp
 \fIauthenticated\fP mode uses no encryption, but authenticates repository contents
 through the same HMAC\-SHA256 hash as the \fIrepokey\fP and \fIkeyfile\fP modes (it uses it
 as the chunk ID hash). The key is stored like \fIrepokey\fP\&.
-This mode is new and \fInot\fP compatible with Borg 1.0.x.
 .sp
 \fIauthenticated\-blake2\fP is like \fIauthenticated\fP, but uses the keyed BLAKE2b\-256 hash
 from the other blake2 modes.
-This mode is new and \fInot\fP compatible with Borg 1.0.x.
 .sp
-\fInone\fP mode uses no encryption and no authentication. It uses SHA256 as chunk
-ID hash. Not recommended, rather consider using an authenticated or
+none mode uses no encryption and no authentication. It uses SHA256 as chunk
+ID hash. This mode is not recommended, users should consider using an authenticated or
 authenticated/encrypted mode. This mode has possible denial\-of\-service issues
 when running \fBborg create\fP on contents controlled by an attacker.
 Use it only for new repositories where no encryption is wanted \fBand\fP when compatibility
 with 1.0.x is important. If compatibility with 1.0.x is not important, use
 \fIauthenticated\-blake2\fP or \fIauthenticated\fP instead.
-This mode is compatible with Borg 1.0.x.
+.sp
+Backwards compatible modes are: none, repokey and keyfile. \fIauthenticated\fP, 
+\fIauthenticated\-blake2\fP, \fIrepokey\-blake2\fP and \fIkeyfile\-blake2\fP are all NOT 
+backwards compatible with Borg1.0.x. Repos created with earler versions of borg can be 
+upgraded with `borg upgrade`
+.sp
 .SH OPTIONS
 .sp
 See \fIborg\-common(1)\fP for common options of Borg commands.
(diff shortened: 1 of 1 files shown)


chrisrapson:
I just found that some of the text is duplicated in `borg/docs/usage/init.rst`. Should I make changes there as well? Do you know any other places where the text might be duplicated? Would it be good practice to try to avoid those duplicates?

ThomasWaldmann:
That stuff is auto-generated from `archiver.py`, please **only** edit the original text there.

chrisrapson:
So I should edit `archiver.py` to match the changes I have made to docs/man/borg-init.1 ?

ThomasWaldmann:
I can first review your changes, but in the end they need to get into `archiver.py` **only**.

docs/man/borg-init.1
```diff
@@ -43,17 +43,12 @@ Encryption can be enabled at repository init time. It cannot be changed later.
...
-If you want "passphrase\-only" security, use one of the repokey modes. The
-key will be stored inside the repository (in its "config" file). In above
-mentioned attack scenario, the attacker will have the key (but not the
-passphrase).
-.sp
-If you want "passphrase and having\-the\-key" security, use one of the keyfile
-modes. The key will be stored in your home directory (in .config/borg/keys).
-In the attack scenario, the attacker who has just access to your repo won\(aqt
-have the key (and also not the passphrase).
+The repository is encrypted with a key, and the key is protected with a passphrase.
```
ThomasWaldmann: Borg relies on randomly generated key material and uses that for chunking, id generation, encryption and authentication. The key material is encrypted using the passphrase you give before it is stored on-disk.

docs/man/borg-init.1
```diff
@@ -43,17 +43,12 @@ Encryption can be enabled at repository init time. It cannot be changed later.
...
-mentioned attack scenario, the attacker will have the key (but not the
-passphrase).
-.sp
-If you want "passphrase and having\-the\-key" security, use one of the keyfile
-modes. The key will be stored in your home directory (in .config/borg/keys).
-In the attack scenario, the attacker who has just access to your repo won\(aqt
-have the key (and also not the passphrase).
+The repository is encrypted with a key, and the key is protected with a passphrase. 
+By setting the encryption mode, you can choose whether to store the key inside the
+repository ("passphrase only" or "repokey" mode) or to store the key separately in
```
ThomasWaldmann: this will confuse people about the mode name.

docs/man/borg-init.1
```diff
@@ -43,17 +43,12 @@ Encryption can be enabled at repository init time. It cannot be changed later.
...
-passphrase).
-.sp
-If you want "passphrase and having\-the\-key" security, use one of the keyfile
-modes. The key will be stored in your home directory (in .config/borg/keys).
-In the attack scenario, the attacker who has just access to your repo won\(aqt
-have the key (and also not the passphrase).
+The repository is encrypted with a key, and the key is protected with a passphrase. 
+By setting the encryption mode, you can choose whether to store the key inside the
+repository ("passphrase only" or "repokey" mode) or to store the key separately in 
+.config/borg/keys ("passphrase and having\-the\-key" or "keyfile" mode). In repokey
```
ThomasWaldmann: same here.

docs/man/borg-init.1
```diff
@@ -81,6 +76,14 @@ a different keyboard layout.
 You can change your passphrase for existing repos at any time, it won\(aqt affect
 the encryption/decryption key or other secrets.
 .SS Encryption modes
+.sp
+For each repo, the user may select the degree of encryption, and the Hash/MAC
```
ThomasWaldmann: degree of encryption?

docs/man/borg-init.1
```diff
@@ -81,6 +76,14 @@ a different keyboard layout.
 You can change your passphrase for existing repos at any time, it won\(aqt affect
 the encryption/decryption key or other secrets.
 .SS Encryption modes
+.sp
+For each repo, the user may select the degree of encryption, and the Hash/MAC 
+algorithm. The table below shows what string to enter, to obtain each combination
+of encryption degree and Hash/MAC algorithm.
```
ThomasWaldmann: usually one would say encryption algorithm.

sadly, in borg's case this is currently rather "enc. algo plus where the key is stored".

docs/man/borg-init.1
```diff
@@ -81,6 +76,14 @@ a different keyboard layout.
...
+.sp
+Not all modes are backwards compatible with Borg1.0.x. New modes which are not 
+backwards compatible are \fImarked like this\fP
```
ThomasWaldmann: why is this better than the old phrasing?

docs/man/borg-init.1
```diff
@@ -126,46 +129,46 @@ _
...
 .sp
-\fIMarked modes\fP are new in Borg 1.1 and are not backwards\-compatible with Borg 1.0.x.
-.sp
+How should a user choose between SHA-256 and Blake2b? The main difference is execution speed. 
 On modern Intel/AMD CPUs (except very cheap ones), AES is usually
 hardware\-accelerated.
 BLAKE2b is faster than SHA256 on Intel/AMD 64\-bit CPUs
 (except AMD Ryzen and future CPUs with SHA extensions),
-which makes \fIauthenticated\-blake2\fP faster than \fInone\fP and \fIauthenticated\fP\&.
+which makes \fIauthenticated\-blake2\fP faster than none and \fIauthenticated\fP\&.
```
ThomasWaldmann: i guess "none" here meant the mode with that name. if you remove the markup, that is not clear any more.

docs/man/borg-init.1
```diff
@@ -126,46 +129,46 @@ _
...
 .sp
 On modern ARM CPUs, NEON provides hardware acceleration for SHA256 making it faster
 than BLAKE2b\-256 there. NEON accelerates AES as well.
 .sp
 Hardware acceleration is always used automatically when available.
 .sp
-\fIrepokey\fP and \fIkeyfile\fP use AES\-CTR\-256 for encryption and HMAC\-SHA256 for
+repokey and keyfile use AES\-CTR\-256 for encryption and HMAC\-SHA256 for
```
ThomasWaldmann: keep markup, for consistency.

docs/man/borg-init.1
```diff
@@ -126,46 +129,46 @@ _
...
 as the chunk ID hash). The key is stored like \fIrepokey\fP\&.
-This mode is new and \fInot\fP compatible with Borg 1.0.x.
 .sp
 \fIauthenticated\-blake2\fP is like \fIauthenticated\fP, but uses the keyed BLAKE2b\-256 hash
 from the other blake2 modes.
-This mode is new and \fInot\fP compatible with Borg 1.0.x.
 .sp
-\fInone\fP mode uses no encryption and no authentication. It uses SHA256 as chunk
-ID hash. Not recommended, rather consider using an authenticated or
+none mode uses no encryption and no authentication. It uses SHA256 as chunk
```
ThomasWaldmann: keep markup

docs/man/borg-init.1
```diff
@@ -126,46 +129,46 @@ _
...
 when running \fBborg create\fP on contents controlled by an attacker.
 Use it only for new repositories where no encryption is wanted \fBand\fP when compatibility
 with 1.0.x is important. If compatibility with 1.0.x is not important, use
 \fIauthenticated\-blake2\fP or \fIauthenticated\fP instead.
-This mode is compatible with Borg 1.0.x.
+.sp
+Backwards compatible modes are: none, repokey and keyfile. \fIauthenticated\fP, 
+\fIauthenticated\-blake2\fP, \fIrepokey\-blake2\fP and \fIkeyfile\-blake2\fP are all NOT 
+backwards compatible with Borg1.0.x. Repos created with earler versions of borg can be 
+upgraded with `borg upgrade`
```
ThomasWaldmann: remove mention of borg upgrade here, it is rather misleading.

docs/man/borg-init.1
chrisrapson: That sounds like useful information. I will update the PR to add those sentences.

docs/man/borg-init.1
chrisrapson: Maybe you're right. I re-used the terminology from the existing manpage. Perhaps it would be better to just say "repokey" mode. i.e. don't mention "passphrase only" at all.

docs/man/borg-init.1
chrisrapson: same here, only call it "keyfile mode" and don't say "passphrase and having-the-key"

docs/man/borg-init.1
chrisrapson: Yes, my mistake, "degree of encryption" is not the correct terminology. I wanted to make it clear that the user has two choices: 
1. none vs authentication vs repokey vs keyfile
2. SHA-256 vs blake2b

How about:
For each repo, the user may choose between four encryption options:
1. none
2. authentication
3. repokey (store key with the repo, only passphrase needed for access)
4. keyfile (store key in $HOME/.config/borg/keys, passphrase and key needed for access)
If encryption is used, the user can select the algorithm from
1. SHA-256
2. Blake2b
The strings to select combinations of each of these options are shown in the following table.

docs/man/borg-init.1
chrisrapson: When I read "Marked modes are new in Borg 1.1" (with "Marked modes" underlined) I didn't realise at first that the underlining was how they were marked. I assume that whether the text is underlined, bold or italics can depend on the renderer, so you can't say "The underlined modes". Anyway, to me it is clearer to state "marked like this".

docs/man/borg-init.1
chrisrapson: My understanding was that the markup indicates which modes are not backwards compatible with Borg1.0.x. Surely "none" is backwards compatible?

docs/man/borg-init.1
chrisrapson: My understanding was that the markup indicates which modes are not backwards compatible with Borg1.0.x. Surely "repokey" and "keyfile" are backwards compatible?

docs/man/borg-init.1
chrisrapson: see above

docs/man/borg-init.1
chrisrapson: Yes, you're probably right. I will remove it.

docs/man/borg-init.1
ThomasWaldmann: AES-CTR is the encryption algorithm borg currently uses.
HMAC-SHA256 or BLAKE2B is the authentication algorithm used.

docs/man/borg-init.1
ThomasWaldmann: "none" (when referring to a mode) must not look like normal text.

so:
- `"none"`
- **none**
- *none*

Just be consistent.

ThomasWaldmann:
btw, it would be nice not to have a merge commit inside a PR.

just one or few normal commits with some specific "topic".

docs/man/borg-init.1
chrisrapson: Ok, I'll try and include that information too:
When initiating each repo, the user has several options which are indicated by inputs to the `--encryption` argument. Firstly, the user may choose to have no encryption, authentication only, or to authentication and encryption. If the user chooses to encrypt the repository, they can choose where to store the key:
1. repokey (store key with the repo, only passphrase needed for access)
2. keyfile (store key in $HOME/.config/borg/keys, passphrase and key needed for access)
If the user chooses to use authentication (including authentication and encryption), they can choose between two different authentication algorithms:
1. SHA-256
2. Blake2b
The strings to select combinations of each of these options are shown in the following table.

docs/man/borg-init.1
chrisrapson: ok. I think the `code` markup makes most sense when talking about command line arguments. I will use that consistently throughout the document.

docs/man/borg-init.1
ThomasWaldmann: The key is also needed for the "authenticated" mode.

wording in 1 is misleading. you always need to have key AND passphrase to access your data.

but if you store the key in the repo, a potential attacker with repo access already has it, so the security offered by this mode is equivalent to "passphrase only" (because this is what the attacker does not have yet).

Authentication is HMAC-SHA-256 (not SHA-256).

ThomasWaldmann:
any news here?

ThomasWaldmann:
close this without merging?

chrisrapson:
Sorry, I had too many other things on. (And, if I'm completely honest, I found the tone of some of the counterpoints more abrasive than constructive. But I'm still very appreciative of the time and effort that you put into this open source project, so I'll assume you're just economising on those scarce resources.)

Anyway, looking back, it seems to me like we reached some sort of consensus on all of the issues you raised. Should I make the changes in the proper location (archiver.py), close this PR and submit a new one?

ThomasWaldmann:
Re-done this, see #5115.

PR #4278, opened 2019-01-30
Update BORG_PASSCOMMAND example and clarify variable expansion.

m3nu:
Updates the `BORG_PASSCOMMAND` example in the FAQ to use `$HOME` instead of `~`, with the latter not being expanded by `cat`.

Next it adds a note regarind the lack of a full shell when executing the `BORG_PASSCOMMAND` in HTML- and man-docs.


ThomasWaldmann:
thanks! can you do a PR against 1.1-maint also?

PR #4280, opened 2019-01-30
Fix '\' not appearing in ReadTheDocs output

strugee:
See https://borgbackup.readthedocs.io/en/stable/usage/help.html for where this is missing.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1984,7 +1984,7 @@ class Archiver:
             any number of characters, '?' matching any single character, '[...]'
             matching any single character specified, including ranges, and '[!...]'
             matching any character not specified. For the purpose of these patterns,
-            the path separator ('\\' for Windows and '/' on other systems) is not
+            the path separator ('\\\\' for Windows and '/' on other systems) is not
             treated specially. Wrap meta-characters in brackets for a literal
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1984,7 +1984,7 @@ def do_break_lock(self, args, repository):
             any number of characters, '?' matching any single character, '[...]'
             matching any single character specified, including ranges, and '[!...]'
             matching any character not specified. For the purpose of these patterns,
-            the path separator ('\\' for Windows and '/' on other systems) is not
+            the path separator ('\\\\' for Windows and '/' on other systems) is not
```
ThomasWaldmann: does it also look correct in `borg help` on the shell after that?

strugee: I don't know; I just sent this from the GitHub web UI. I can try locally if you really want? But it would be a hassle to set up a development environment.

ThomasWaldmann: yes, please try.

it would not help fixing readthedocs if it breaks the cli help.

ThomasWaldmann: update: as i suspected, this fix will render "\\" on the shell.

so thanks for finding this, but i'll close this PR as this fix just moves the problem to the cli help.

PR #4282, opened 2019-01-30
Add new FAQ regarding change of compression settings. Fixes #4222

### LINKED ISSUE #4222: Can I safely change the compression level or algorithms?
Hi there, this is a **question**.

I just read the following in the borg manual about the `lzma` compression alg:

```
... Giving levels above 6 is pointless and counterproductive because it does not compress better due to the buffer size used by borg - but it wastes lots of CPU cycles and RAM.
```

I have some questions regarding that:

1) I've been using 9 as the level. Now I'd like to reduce to 6 but keep using `lzma`. Can I safely do this without affecting my backup archives? Is the new archive going to take more time or take more space because of it? Are there any other consequences of doing this?

2) Also, what if I want a more radical change and decide to use another compression algorithm, what are the consequences in the repo (besides the final compression output varying) if any?

3) Can I also add the `auto` compression parameter later without further consequences?

Thanks in advance!

m3nu:


docs/faq.rst
```diff
@@ -295,6 +295,15 @@ the :ref:`borg_recreate` command to rewrite all archives with a
 different ``--exclude`` pattern. See the examples in the
 :ref:`borg_recreate` manpage for more information.
 
+Can I safely change the compression level or algorithms?
```
ThomasWaldmann: "algorithm" (for consistency with "level").

docs/faq.rst
```diff
@@ -295,6 +295,15 @@ the :ref:`borg_recreate` command to rewrite all archives with a
...
+--------------------------------------------------------
+
+The compression level and algorithm don't affect deduplication. File
+hashes are calculated *before* compression. New compression settings
+will only be applied to new data, not existing archives. So it's safe
+to change them.
```
ThomasWaldmann: Maybe we rather should refer to "chunks" here, as that is being processed (not: file, archive).

So:
- "Chunk ID hashes"
- "new chunks"
- "existing chunks"

Also we could mention that borg recreate can be used to recompress.

m3nu:
Updated wording as per review.

ThomasWaldmann:
thanks!

PR #4318, opened 2019-02-05
work around the backslash issue, see #4280

ThomasWaldmann:
supercedes #4280 - thanks to AJ Jordan @strugee for finding this.

while using "\\" gives a single backslash when using "borg help pattern"
on the shell, our rendered html docs did not show the single backslash
at all.

i was lazy do dig into that issue, so i just used the word "backslash".

:-)


strugee:
Much appreciated :-)

So sorry for the delayed response on #4280; I've been swamped with college stuff (and getting sick too).

ThomasWaldmann:
@strugee no problem, wish you get well soon!

PR #4331, opened 2019-02-07
borg mount / borgfs security fix (1.1-maint)

ThomasWaldmann:
platform testing will be done as part of release process.


ThomasWaldmann:
fix is in 1.1.9 release.

result of platform testing was at first that same fakeroot-caused test failures as on macOS in the past are now also seen on linux (due to the on-by-default permissions check). fixed all platforms by ignoring permissions checking for this test.

PR #4349, opened 2019-02-11
Update bundled lz4 code to 1.8.3, fixes #4209

### LINKED ISSUE #4209: update bundled lz4 code
we have 1.8.2 src bundled, current release see there: https://github.com/lz4/lz4/releases

there is a corruption fix in 1.8.3 and it does not affect borg, see https://github.com/lz4/lz4/issues/560.

also, borg's default is to use a system lz4 library (not the bundled src), so the question with that is whether misc. dists did already fix their liblz4.

donnyward:
Tested on Ubuntu 16.04.2 LTS.


ThomasWaldmann:
thanks!

PR #4372, opened 2019-02-15
Adds docker image.

pierreozoux:
Relates to #4364

This is the first step, and then once you detail what is the release process, we can add that later.

It always build latest, but if you want we can add an ARG with the version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM debian:stretch
+
+ENV LANG=C.UTF-8
+
+COPY ./borg-linux64 /usr/bin/borg
+COPY ./scripts/docker/borgbackup /usr/bin/
+COPY ./scripts/docker/borgserver /usr/bin/
+
+ENTRYPOINT ["/usr/bin/borg"]
--- a/scripts/docker/CLIENT.md
+++ b/scripts/docker/CLIENT.md
@@ -1,0 +1,177 @@
+## CLient side
+
+The state of this document and the associated script is beta, please use with caution.
+
+### QuickStart
+
+Create and run the borg backup container.
+In this quick start, the `/etc` and `/home` directories
+from the host are bind mounted to the container as read only. These are
+the directories which will be backed up. The backed up data will be stored
+in the `borg-repo` Docker volume, and the data will be encrypted with a key 
+protected with the `my-secret-pw` password. If the host is using SELinux, the
+`--security-opt label:disable` flag must be used, because we don't want
+to relabel the `/etc` and `/home` directories while we want the container
+to have access to them. After the backup is done, data will be pruned
+according to the default policy and checked for errors. Borg is running
+in a verbose mode within the container, so the detailed output from backup
+will be printed. 
+
+```
+docker run \
+  -e BORG_REPO=/borg/repo \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e EXCLUDE='*/.cache*;*.tmp;/borg/data/etc/shadow' \
+  -e COMPRESSION=lz4 \
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v borg-repo:/borg/repo \
+  -v /etc:/borg/data/etc:ro \
+  -v /home:/borg/data/home:ro \
+  --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+### More examples
+
+Backup docker volumes to remote location (Borg must be running in server mode at that remote location):
+
+```
+docker run \
+  -e BORG_REPO='user@hostname:/path/to/repo' \
+  -e ARCHIVE=wordpress-{now:%Y-%m-%d} \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e BACKUP_DIRS=/borg/data \
+  -e COMPRESSION=lz4 \
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v mariadb-data:/borg/data/mariadb:ro \
+  -v wordpress-data:/borg/data/wordpress:ro \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+Using sshfs (in case when the Borg is not installed on the remote location):
+
+```
+docker run \
+  -e SSHFS='user@hostname:/path/to/repo' \
+  -e SSHFS_PASSWORD=my-ssh-password \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e BACKUP_DIRS=/borg/data \
+  -e COMPRESSION=lz4 \
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v mariadb-data:/borg/data/mariadb:ro \
+  -v wordpress-data:/borg/data/wordpress:ro \
+  --cap-add SYS_ADMIN --device /dev/fuse --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+Using sshfs with ssh key authentication:
+
+```
+docker run \
+  -e SSHFS='user@hostname:/path/to/repo' \
+  -e SSHFS_IDENTITY_FILE=/root/ssh-key/key \
+  -e SSHFS_GEN_IDENTITY_FILE=1 \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e BACKUP_DIRS=/borg/data \
+  -e COMPRESSION=lz4 \
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v borg-ssh-key:/root/ssh-key \
+  -v mariadb-data:/borg/data/mariadb:ro \
+  -v wordpress-data:/borg/data/wordpress:ro \
+  --cap-add SYS_ADMIN --device /dev/fuse --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+Restoring files from specific day to folder on host:
+
+```
+docker run \
+  -e BORG_REPO='user@hostname:/path/to/repo' \
+  -e ARCHIVE=wordpress-2016-05-25 \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e EXTRACT_TO=/borg/restore \
+  -e EXTRACT_WHAT=only/this/file \
+  -v borg-cache:/root/.cache/borg \
+  -v /opt/restore:/borg/restore \
+  --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+Running custom borg command:
+
+```
+docker run \
+  -e BORG_REPO='user@hostname:/path/to/repo' \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e BORG_PARAMS='list ::2016-05-26' \
+  -v borg-cache:/root/.cache/borg \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+## Environment variables
+
+This docker image accepts [environment variables from borg](https://borgbackup.readthedocs.io/en/stable/usage/general.html#environment-variables), and some more described here:
+
+### Core variables
+
+**ARCHIVE** - archive parameter for Borg repository. If empty, defaults to `"{hostname}_{now:%Y-%m-%d}"`. For more info see [Borg documentation](https://borgbackup.readthedocs.io/en/stable/usage.html)
+
+**BACKUP_DIRS** - directories to back up
+
+**EXCLUDE** - paths/patterns to exclude from backup. Paths must be separated by `;`. For example: `-e EXCLUDE='/my path/one;/path two;*.tmp'`
+
+**BORG_PARAMS** - run custom borg command inside of the container. If this variable is set, default commands are not executed, only the one specified in *BORG_PARAMS*. For example `list` or `list ::2016-05-26`. In both examples, repo is not specified, because borg understands the `BORG_REPO` env var and uses it by default
+
+**BORG_SKIP_CHECK** - set to `1` if you want to skip the `borg check` command at the end of the backup
+
+**SSH_KEY** - set the ssh key when you start this image in server mode
+
+### Compression
+
+**COMPRESSION** - compression to use. Defaults to lz4. [More info](https://borgbackup.readthedocs.io/en/stable/usage.html#borg-create)
+
+### Extracting (restoring) files
+
+**EXTRACT_TO** - directory where to extract (restore) borg archive. If this variable is set, default commands are not executed, only the extraction is done. Repo and archive are specified with *BORG_REPO* and *ARCHIVE* variables. [More info](https://borgbackup.readthedocs.io/en/stable/usage.html#borg-extract)
+
+**EXTRACT_WHAT** - subset of files and directories which should be extracted
+
+### Pruning
+
+**PRUNE** - if set, prune the repository after backup. Empty by default. [More info](https://borgbackup.readthedocs.io/en/stable/usage.html#borg-prune)
+
+**PRUNE_PREFIX** - filter data to prune by prefix of the archive. Empty by default - prune all data
+
+**KEEP_DAILY** - keep specified number of daily backups. Defaults to 7
+
+**KEEP_WEEKLY** - keep specified number of weekly backups. Defaults to 4
+
+**KEEP_MONTHLY** - keep specified number of monthly backups. Defaults to 6
+
+### SSHFS
+
+**SSHFS** - sshfs destination in form of `user@host:/path`. When using sshfs, container needs special permissions: `--cap-add SYS_ADMIN --device /dev/fuse` and if using SELinux: `--security-opt label:disable` or apparmor: `--security-opt apparmor:unconfined`
+
+**SSHFS_PASSWORD** - password for ssh authentication
+
+**SSHFS_IDENTITY_FILE** - path to ssh key
+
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM debian:stretch
```
ThomasWaldmann: https://github.com/docker-library/official-images/issues/1516

ends with:

```
So for most of us there is no way to run verified trusted docker
images in production. Very unfortunate.
```

I am not familiar with docker, but sounds pretty bad?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+ENV LANG=C.UTF-8
+
+RUN set -x \
+  && apt-get update \
```
ThomasWaldmann: with recent apt, you can just run "apt" instead of "apt-get".

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    libacl1 \
+    libacl1-dev \
+    libssl-dev \
+    liblz4-dev \
+    liblz4-1 \
+    python3-pip \
+  && pip3 install borgbackup \
+  && apt-get remove -y --purge build-essential libssl-dev liblz4-dev libacl1-dev \
+  && apt-get autoremove -y --purge \
+  && rm -rf /var/lib/apt/lists/* \
```
ThomasWaldmann: do apt clean (first)?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN set -x \
+  && apt-get update \
+  && apt-get install -y \
+    build-essential \
+    libacl1 \
+    libacl1-dev \
+    libssl-dev \
+    liblz4-dev \
+    liblz4-1 \
```
ThomasWaldmann: just noticed that our docs need improving here. also check if you are missing something (that might be present, but better do it explicitly):

https://borgbackup.readthedocs.io/en/stable/installation.html#debian-ubuntu

we shall get libzstd(-dev) installed from dist also (prefer that over our builtin zstd code).

ThomasWaldmann:
one general thing: the whole docker image building should run inside the stretch64 vagrant vm.

Dockerfile
DarrienG: debian/debian based distros recommend not using `apt` in scripting and it won't throw errors, but will throw a bunch of text to `stderr` if you use `apt` instead of `apt-get`. Feel free to try in a small bash script if you're interested. I recommend keeping `apt-get`.

Dockerfile
DarrienG: You're not wrong, but in order for an attack like this to happen a malicious user would have to:
- Steal the Docker corporation's login credentials
- Point the Docker library repo from the official debian repo to somewhere else (or compromise the Debian repo)
- Put it up for review and hope nobody notices the alarm bells that will be going off everywhere
- Wait for the automated build to push the official image up (yet again that would throw a fit)
- Clean any cached images on a user's disks
- Re-pull and re-run images

All content (official and not) served from Docker is done so via HTTPS/TLS so stealing credentials should be the only vector of attack.

DarrienG:
I vote for versioned building. We're asking for trouble always building the latest.

Dockerfile
ThomasWaldmann: interesting, didn't know that recommendation.

Dockerfile
ThomasWaldmann: well, my idea of security is not to rely on some infrastructure being secure and nothing going wrong, but to just check a gpg signature. it doesn't matter how or from where you get the stuff as long as you can verify the sig.

ThomasWaldmann:
Maybe it could be also moved to `scripts/Dockerfile`? There are already too many files in `.`.

Dockerfile
MarkusTeufelberger: You could/should just use a 2 stage container build instead (one for the build dependencies, one with only runtime dependencies and the output of the build).

Dockerfile
MarkusTeufelberger: You could pin to a sha256 hash of the base image instead of a tag, but then you'd have the maintenance burden of verifying every new version manually and updating this hash.

Gpg is no magic security fairy dust either, in Debian you also rely on the infrastructure of whoever created and distributes the signature. Unfortunately Docker images are not created in a deterministic way themselves (and this Dockerfile also adds additional nondeterminism on top by using whatever package version is current at build time).

ThomasWaldmann: @MarkusTeufelberger if you have the correct public key of the release maintainer, you can verify all releases with their signature, no matter how you get the release or the release signature. you can not do that just with sha256.

ThomasWaldmann:
@pierreozoux are you still working on this?

Dockerfile
pierreozoux: There is a new feature Docker content trust, I'm not familiar with it yet, and this is an "enterprise feature" (at least in docker)... But other container runtime use it also (like cri-o) and other docker registry too (like quay.io). I propose that we investigate that in a future iteration.

pierreozoux:
Sorry for the delay, holiday.. ;)

I put some brain juice on it. And about your questions, I think this PR is the best possible way to tackle that.

    Don't install third party in the VM

[docker is a third party repo in debian](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/debian/).

Copying from VM to local machine would have been a hassle, so I though it is just pain easier to build it from your dev machine. (as anyway you'll need the client for pushing to the hub).

    use `scripts` folder for Dockerfile

I think I like that the Dockerfile is at the root. As a docker user, in a quick look, I can see docker support of a repo, but no strong opinion there.
If you run the `build_and_publish` script from the root of the repo, all should work.
You'd need a docker hub account thought. I [created the org already](https://cloud.docker.com/u/borgbackup/repository/docker/borgbackup/borg/tags), once you have your docker handle, give it to me, and I make you admin (and then you can remove me, I'm fine).

    pip vs setup.py vs pyinstaller

I could have choosen another method, but I'm not really familiar with python. And so I though that the fat binary was equivalent to any other method. But I have no strong opinion there. If you have a strong opinion, we can change that.

    borgbackup and borgserver

It is just an attempt, if it creates too much debate, I'd prefer to not ship it with this PR, and open a separate PR

scripts/docker/README.md
```diff
@@ -0,0 +1,190 @@
...
+  -e BORG_REPO=/borg/repo \
+  -e BORG_PASSPHRASE=my-secret-pw \
+  -e BACKUP_DIRS=/borg/data \
+  -e EXCLUDE='*/.cache*;*.tmp;/borg/data/etc/shadow' \
+  -e COMPRESSION=lz4 \
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v borg-repo:/borg/repo \
+  -v /etc:/borg/data/etc:ro \
+  -v /home:/borg/data/home:ro \
```
ThomasWaldmann: i somehow dislike there is a whole new layer of wrapping / options / env vars (requiring a whole new layer of docs), just to get the usual borg command lines over into the container.

can't this be done way easier?

scripts/docker/README.md
```diff
@@ -0,0 +1,190 @@
...
+  -v borg-cache:/root/.cache/borg \
+  -v mariadb-data:/borg/data/mariadb:ro \
+  -v worpdress-data:/borg/data/wordpress:ro \
+  --cap-add SYS_ADMIN --device /dev/fuse --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+```
+
+Using sshfs with ssh key authentification:
```
ThomasWaldmann: authentication

scripts/docker/README.md
```diff
@@ -0,0 +1,190 @@
...
+  -e PRUNE=1 \
+  -v borg-cache:/root/.cache/borg \
+  -v borg-repo:/borg/repo \
+  -v /etc:/borg/data/etc:ro \
+  -v /home:/borg/data/home:ro \
+  --security-opt label:disable \
+  --entrypoint=/usr/bin/borgbackup \
+  --name borg-backup \
+  borgbackup/borg
+docker rm borg-backup
```
DarrienG: It probably makes sense to just run with `docker run --rm` so you don't have to run `docker rm` afterwards. This sample image doesn't save data in the container, so it's best to be a good citizen and remove it afterwards.

scripts/docker/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Quick start
+
+First, pull the image to keep it up to date. Then create and run the borg backup container. In this quick start, the `/etc` and `/home` directories from the host are bind mounted to the container as read only. These are the directories which will be backed up. The backed up data will be stored in the `borg-repo` Docker volume, and the data will be protected with the `my-secret-pw` password. If the host is using SELinux, the `--security-opt label:disable` flag must be used, because we don't want to relabel the `/etc` and `/home` directories while we want the container to have access to them. After the backup is done, data will be pruned according to the default policy and checked for errors. Borg is running in a verbose mode within the container, so the detailed output from backup will be printed. At the end, the container is deleted. This is done by separate `docker rm` command, because the `--rm` option to the `docker run` would remove also the Docker volumes, and we don't want that. Deleting the container and pulling the image from registry every time keeps the container fresh every time the backup is run.
+
+```
+docker pull borgbackup/borg
+docker run \
```
DarrienG: As long as you use json syntax (which we are) for the entrypoint, you don't need to do anything weird to pass arguments to the container.

You could use `docker run --rm borgbackup/borg -v $YOUR_MOUNTS $NORMAL_BORG_OPTIONS as if it were a binary.

scripts/docker/README.md
DarrienG: I don't think there should be a whole set of docs, especially for the client side version of borg. Docker makes running commands dead simple as if you were working with a binary. As long as we say that the entrypoint for the container is borg, the rest can be run like a binary.

scripts/docker/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+**SSH_KEY** - set the ssh key when you start this image in server mode
+
+### Compression
+
+**COMPRESSION** - compression to use. Defaults to none. [More info](https://borgbackup.readthedocs.io/en/stable/usage.html#borg-create)
+
+### Encryption
+
+**BORG_PASSPHRASE** - `repokey` mode password to encrypt the backed up data. Defaults to none. Only the `repokey` mode encryption is supported by this Docker image. [More info](https://borgbackup.readthedocs.io/en/stable/usage.html#borg-init)
```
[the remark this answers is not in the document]
DarrienG: Agree - there's no need to duplicate info. Docker is just supposed to be a container environment. This doc should only have Docker specific options.

dargmuesli:
Let me add a quick **thank you** for working on this guys! :)

scripts/docker/README.md
pierreozoux: it is using docker volumes, so it would delete the associated volume.

scripts/docker/README.md
pierreozoux: I separated the binary part from the convenience scripts for more clarity.

pierreozoux:
Ok, I separated the documentation from the binary and the convenience scripts (borgbackup and borgserver), I clearly tagged them as beta.
I think we are now good to get started like this, and improve in an iterative manner.
(I'll also use the convenience script in my production, once I use them, I'll probably add some fix and remove the beta flag :) )

scripts/docker/CLIENT.md
```diff
@@ -0,0 +1,177 @@
+## CLient side
```
dargmuesli: `Client`

pierreozoux:
@ThomasWaldmann if I remove all the client and server scripts, would you feel better about it?

I can spend a bit of time to rework this to have a basic docker image at least.

ThomasWaldmann:
Yeah, would feel better with a very basic thing. But as I don't use docker, I can't comment much about how that should look like.

But there were some other reviewers here, maybe you can discuss about how a basic docker image for borg should look like, so that it is generally useful.

Nielio:
Any updates on this? Would be nice to have an official containerd borg !

pascaldornfeld:
I am looking for an official image too

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM debian:stretch
```
williamdes, line 1: No sure this program needs Debian
You should use Alpine

Docker content trust works fine, you can read how to automate or use it here: https://github.com/sudo-bot/action-docker-sign

scripts/docker/borgserver
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -eux
+if [ -n "${SSH_KEY:-}" ]; then
+  dpkg-reconfigure openssh-server
+  sed -i \
+    -e 's/^#PasswordAuthentication yes$/PasswordAuthentication no/g' \
+    -e 's/^PermitRootLogin without-password$/PermitRootLogin no/g' \
+    /etc/ssh/sshd_config
+  if [[ $(ls /etc/ssh-keys/ssh_host_*) ]]; then
```
rugk, line 8: AFAIK one should not use [ls in scripts/parse it](https://unix.stackexchange.com/a/129120/146739). Okay, maybe there is not much parsing happening, but I you may anyway get crazy edge-cases/it is just not clean when you e.g. have spaces or so in there…
What I wanna say: Maybe there is a better solution? :slightly_smiling_face: 

Also, were these scripts tested with [shellcheck](https://www.shellcheck.net/)? (Just pasted this script and it did not complain to my surprise, so yeah,hmm…)

scripts/docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+  -v /path/to/host/repo:/path/to/repo \
+  borgbackup/borg \
+  create /path/to/repo::Tuesday /host/src /host/Documents
+```
+
+Please read the [quick start](https://borgbackup.readthedocs.io/en/stable/quickstart.html)
+from the borg command line to continue this Quick Start.
+
+Note that this example is using the `latest` version of borg, and we recommend you
+to pin the version by appending version e.g. `borgbackup/borg:1.0.7` to use `1.0.7`.
```
rugk, line 47: Hmm why would you recommend it?

I mean, maybe to not break setups the major version could be pinned aka `borgbackup/borg:1`, but semantic versioning says everything else should not break the API…

scripts/docker/SERVER.md
```diff
@@ -0,0 +1,20 @@
+## Server side
+
+The state of this document and the associated script is beta, please use with caution.
+
+To start a borg server:
+ - generate an ssh-key
+   - `ssh-keygen -f ./id_rsa -N '' -t rsa`
```
rugk, line 7: It's 2021, not sure whether recommending RSA keys is still what we need. ED25519 could be used just as that, is shorter and at least as secure…

scripts/docker/borgserver
```diff
@@ -0,0 +1,23 @@
...
+    echo "Already initialized."
+  else
+    echo "Initializing ssh-keys folder..."
+    cp /etc/ssh/ssh_host_* /etc/ssh-keys
+  fi
+  chmod 400 /etc/ssh-keys/ssh_host_*
+  sed -i 's|/ssh/|/ssh-keys/|' /etc/ssh/sshd_config
+  sed -i '/^#HostKey/s/^#//' /etc/ssh/sshd_config
+  sed -e "s#SSH_KEY#${SSH_KEY}#g" /home/borg/authorized_keys.sample > /home/borg/.ssh/authorized_keys
```
rugk, line 17: Hmm, where does this sample file come from? Debian I guess?
'
Okay, then it is empty.

Anyway, you should really suggest to save this file/dir in some (persistent) volume or prefill it (and not only `SSHFS_IDENTITY_FILE`), because otherwise you can get client connection errors…
Or oh well, this is the server part, nevermind I comment above…

scripts/docker/borgbackup
```diff
@@ -0,0 +1,133 @@
...
+    exit "$1"
+  fi
+
+  exit 0
+}
+
+if [ -n "${SSHFS:-}" ]; then
+  if [ -n "${SSHFS_IDENTITY_FILE:-}" ]; then
+    if [ ! -f "$SSHFS_IDENTITY_FILE" ] && [ -n "${SSHFS_GEN_IDENTITY_FILE:-}" ]; then
+      ssh-keygen -t rsa -b 4096 -N '' -f "$SSHFS_IDENTITY_FILE"
```
rugk, line 24: Again in 2021, we maybe should use ed25519 instead of RSA…

scripts/docker/borgbackup
```diff
@@ -0,0 +1,133 @@
...
+  fi
+  if [ -n "${SSHFS_PASSWORD:-}" ]; then
+    SSHFS_PASSWORD="echo ${SSHFS_PASSWORD} |"
+    SSHFS_PASSWORD_OPT='-o password_stdin'
+  else
+    SSHFS_PASSWORD=''
+    SSHFS_PASSWORD_OPT=''
+  fi
+  mkdir -p /mnt/sshfs
+  eval "${SSHFS_PASSWORD} sshfs -o StrictHostKeyChecking=no ${SSHFS} /mnt/sshfs ${SSHFS_IDENTITY_FILE} ${SSHFS_PASSWORD_OPT}"
```
rugk, line 40: `StrictHostKeyChecking=no`? Please, by all means, do not do this. This image is likely intended to be used as a client with remote servers, so not checking the server key could allow MITM attacks…

So better pass in the host key either as a variable or just use SSH's default [TOFU](https://en.wikipedia.org/wiki/Trust_on_first_use) security scheme, at least – with allowing the user to confirm the step/check the key manually. After all, they will run this manually one time at least, so that should work.

scripts/docker/borgserver
rugk: But should not this be configurable via a command-line, too? Otherwise, clients cannot connect if they are not authorized.

Also you'd like to restrict the commands to `borg serve` for security reasons.
See https://borgbackup.readthedocs.io/en/stable/deployment/central-backup-server.html#restrictions and https://borgbackup.readthedocs.io/en/stable/deployment/hosting-repositories.html

The script here could do that too/configure it properly…

rugk (changes requested):

pierreozoux:
Sorry, I'll close this PR, feel free to make your own PR based on this code or not.
(I'm not using borg any more)

It was a really nice software, I really enjoyed it! We are now using s3 and minio, using versionning and mirroring of s3, basically our requirements changed from let's have our backups encrypted to let's have a mirror site that can start fast in case of emergency.

PR #4378, opened 2019-02-20
Emit user-friendly error if repo key is exported to a directory

### LINKED ISSUE #4348: borg key export confusing error msg
from IRC:
```
borg key export REPO PATH
```

if the second argument is a directory it will throw an error.

I understand it needs a file, but the error it throws may be confusing to other users. :)

milkey-mouse:
Fixes #4348.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -325,10 +325,14 @@ class Archiver:
             if not args.path:
                 self.print_error("output file to export key to expected")
                 return EXIT_ERROR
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
```
src/borg/testsuite/archiver.py:2647:9: F841 local variable 'repo_id' is assigned to but never used
```

milkey-mouse:
That's what I get for only running the `archiver` tests to save time.

PR #4382, opened 2019-02-20
add --list flag for borg delete, fixes #4359

### LINKED ISSUE #4359: Feature Request: --list for delete command
While cleaning up my repo to remove some misnamed archives (from a script snafu on my part) I had occasion to use the `delete` command. While using delete to remove some archives with a specific starting string, I found the command to be lacking a `--list` flag like the `prune` command has. I would suggest that it would be good UI for delete to have this too, so that Users can use the `--dry-run` flag and see what they will be matching.

The current STDOUT results of both `borg delete --dry-run -P "some_string"` (repo set by ENV VAR) and `borg delete -P "some_string"` is NULL currently.

joey4:

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1048,6 +1048,8 @@ class Archiver:
 
         manifest, key = Manifest.load(repository, (Manifest.Operation.DELETE,))
 
+        list_logger = logging.getLogger('borg.output.list')
+
         if args.location.archive or args.archives:
             archives = list(args.archives)
             if args.location.archive:
@@ -1061,6 +1063,7 @@ class Archiver:
         if args.forced == 2:
             deleted = False
             for i, archive_name in enumerate(archive_names, 1):
+                current_archive = manifest.archives[archive_name]
                 try:
                     del manifest.archives[archive_name]
                 except KeyError:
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -1070,6 +1073,11 @@ def _delete_archives(self, args, repository):
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
                     logger.info(msg.format(archive_name, i, len(archive_names)))
+                    list_msg = 'Would delete archive:' if dry_run else 'Deleted archive:'
+                    if args.output_list:
+                        list_logger.info("{message:<40} {archive}".format(
+                            message=list_msg, archive=format_archive(current_archive)
+                        ))
```
ThomasWaldmann: hmm, are we duplicating functionality we already have?

it looks like giving `--info` (aka `-v`), see code in 1074/1075, does already what you added for `--list`.

(Sorry, was not aware of that when commenting on the ticket.)

src/borg/archiver.py
```diff
@@ -1084,11 +1092,18 @@ def _delete_archives(self, args, repository):
...
                 msg = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
+                list_msg = 'Would delete archive:' if dry_run else 'Deleting archive:'
                 logger.info(msg.format(archive_name, i, len(archive_names)))
                 if not dry_run:
                     Archive(repository, key, manifest, archive_name, cache=cache).delete(
                         stats, progress=args.progress, forced=args.forced)
+                if args.output_list:
+                    list_logger.info("{message:<40} {archive}".format(
+                        message=list_msg, archive=format_archive(current_archive)
+                ))
```
ThomasWaldmann: same here, see 1098.

src/borg/archiver.py
joey4: --info shows only the name of the archive. Since the timestamp of the archives is also printed when deleting the entire repository, I think it would be more consistent to do this for parts of the repository as well.

ThomasWaldmann: ok, the timestamp might be useful indeed (at least for people who do not include it into the archive name anyway).

but still, the functionality is now almost duplicate and i think one way should be enough.
if that way shall include the timestamp, then just add that.

joey4: okay, i'll do that and then make a new PR. :)

Issue #2251, opened 2017-03-04
Audit an append-only mode repo to make sure the client was well behaved


nadalle:
This idea is an attempt to detect/mitigate the following attack scenario involving append-only mode:

Say you're using append-only mode to protect your repo from an untrusted and hacked client deleting / changing your old archives.  In addition, you periodically connect to the repo from a special trusted client in normal mode, and issue some prune commands.

Currently, if you do this, the trusted client will happily compact the repo as soon as it tries to write, happily applying all the untrusted client's bad changes permanently. So, the administrator would have to qualify the repo contents before compaction happens.

How could an automated qualification work?

The repo itself is a log-like structure, that means it is a sequence of operations like:

put(1, A), commit, put(1, B), commit

The current value of object 1 is now B, but we still can see it was A before. Until compaction happens and removes the superceded first put().

Untrusted clients are forced into append-only mode by the server, so we see their log of operations.
The trusted admin client is not append-only and removes the log as soon as compaction runs.

PUT to id 0: these are manifest updates, we can check if archives are appended and previously present archives are still same (name, date, id). Special case: overwriting a .checkpoint archive of the same name.

PUT to other ids:
- id is new: likely legitimate, done by "borg create"
- id is not new: this means we overwrite already present data with new (different?) data. "borg recreate" legitimately does this when recompressing chunks, but other than that it should not happen. existing-id overwrite maybe could also be denied by server in append-only mode.

DEL: should not happen, only exceptions are .checkpoint archives (? CHECK THIS). A well behaved borg client should reject high level "borg delete" and "borg prune" commands in append-only mode.

This stuff could be a "borg qualify" ("borg audit"?) command and it would exit with success if all looks ok or emit warnings/errors and exit with warning/error status if not. If not, it should also give the last known-good transaction number.

Additionally, the trusted client likely wants to run "borg check" to assert the integrity of the archive (like discovering bad sectors or other corruption / inconsistency).

Optionally, the trusted client could make added archives pass a set of tests (like checking for expected content: is it there, can it be extracted, can it be used?).

This way, a user who worries about their untrusted clients being hacked can feel safe running a prune command automatically on their trusted client with a cron job: they'll still have old backups from before the hack available, so long as they detect it in time.

(by tw based on nadelle's post)

ThomasWaldmann:
@nadalle hmm, some stuff above is a bit confusing (and partially not quite how I meant it). shall I edit your text and fix it / remove all the confusing or not needed stuff?

[...]

Issue #1772, opened 2016-10-28
Implement a "safe" append-/readonly-mode


MichaelHierweck:
The docs claim append-mode can be used to used to prevent hacked clients from _permanently_ altering existing archives. This can be achieved be granting only append-mode access to the client. Then changes to the repository are appended to the transaction log/journal and can be reverted by removing the lastet transactions from the journal.

First, this kind of manual roleback is not state-of-the-art. ;)

Second, disk space is not infinite. Sooner or later a trusted client (or the server) itself will need to free disk space. This requires "true" write access to repository and is done by prune. However archives that have been marked as (to-be-)deleted in append-mode will be wiped out by prune even if the retention policy specified along with the prune invokation should have preserved them.

See:  #1689 and #1744

Therefore the trusted client the invokes prune on the repository is responsible for checking the integrity of the repository. But how could the be achieved? When a trusted client runs prune at a time when a hack of a client was not detected yet the prune action will apply any malicious trancations permanently. Then even archives might be purges or compromised that have been created before the hack and should not have been purged according to the retention policy. This would make desaster recovery from borgbackup based backups impossible.

I would like to suggest the implementation of a (new) safe append-, readonly-, worm-mode or whatever-mode that restricts clients to add new archives and rejects any action that would delete or change existing archives. Prohibited actions should be rejected immediately and therefore should not go into the journal at all.

ThomasWaldmann:
Yes, currently one has to be sure about having a "valid" (untampered) repo state before writing to it with append-mode=0.

borg list repo, borg list archive, borg extract --dry-run archive can help here, but making really really sure might be difficult.

We could have something better if we could disallow delete tags within a no-delete mode.

ThomasWaldmann:
I reviewed the code where `repository.delete(id)` is used:
- by `Archive.delete()` (via `chunk_decref()`)
- by `borg debug delete-obj`
- by `borg check --verify-data` in `verify_data()` to remove corrupt objects (so that they will be replaced by non-corrupt ones by later backups, hopefully)
- by `borg check --repair` in `orphan_chunks_check()`
- by `Archive.write_checkpoint()` to remove the checkpoint archive item again **after** it has been saved/committed (so that the next checkpoint [or final] save/commit will replace it without creating unreferenced stuff)

The first ones are more or less expected and unproblematic (we just need to fail them early if there is no delete capability) - they don't need to be done from a not-that-much-trusted client (but can be done from a more trusted machine).

[...]

PR #4384, opened 2019-02-20
Elaborate on append-only mode docs

milkey-mouse:
I explained the importance of verifying repository integrity before writing in append-only mode, and some simple ways to do so.

Related to #3504. If you find these docs adequately explain the situation with append-only mode it can be closed. I agree it is kind of confusing; the ideal solution (aside from completely replacing append-only mode) would be to implement something like #2251.

### DIFF
--- a/docs/usage/notes.rst
+++ b/docs/usage/notes.rst
@@ -208,6 +208,9 @@ To activate append-only mode, set ``append_only`` to 1 in the repository config:
 
     borg config /path/to/repo append_only 1
 
+Note that you can go back-and-forth between normal and append-only operation with
+``borg config``; it's not a "one way trip."
+
 In append-only mode Borg will create a transaction log in the ``transactions`` file,
 where each line is a transaction and a UTC timestamp.
 
@@ -278,10 +281,19 @@ won't free disk space, they merely tag data as deleted in a new transaction.
 Be aware that as soon as you write to the repo in non-append-only mode (e.g. prune,
 delete or create archives from an admin machine), it will remove the deleted objects
 permanently (including the ones that were already marked as deleted, but not removed,
-in append-only mode).
+in append-only mode). Automated edits to the repository (such as a cron job running
+``borg prune``) will render append-only mode moot if data is deleted.
 
-Note that you can go back-and-forth between normal and append-only operation by editing
-the configuration file, it's not a "one way trip".
+Even if an archive appears to be available, it is possible an attacker could delete
+just a few chunks from an archive and silently corrupt its data. While in append-only
+mode, this is reversible, but ``borg check`` should be run before a writing/pruning
+operation on an append-only repository to catch accidental or malicious corruption::
+
+    # run without append-only mode
+    borg check --verify-data repo && borg compact repo
+
+Aside from checking repository & archive integrity you may want to also manually check
+backups to ensure their content seems correct.
 
 Further considerations
 ++++++++++++++++++++++


docs/usage/notes.rst
```diff
@@ -208,6 +208,9 @@ To activate append-only mode, set ``append_only`` to 1 in the repository config:
 
     borg config /path/to/repo append_only 1
 
+Note that you can go back-and-forth between normal and append-only operation by editing
+the configuration file, it's not a "one way trip".
```
ThomasWaldmann: maybe don't say "by editing the configuration file" (that sounds like `vim config`) just below showing the `borg config` command that does that.

docs/usage/notes.rst
```diff
@@ -278,10 +281,19 @@ won't free disk space, they merely tag data as deleted in a new transaction.
 Be aware that as soon as you write to the repo in non-append-only mode (e.g. prune,
 delete or create archives from an admin machine), it will remove the deleted objects
 permanently (including the ones that were already marked as deleted, but not removed,
-in append-only mode).
+in append-only mode). Automated edits to the repository (such as a cron job running
+``borg prune``) will render append-only mode moot if data is deleted.
```
ThomasWaldmann: not sure about what new information this is telling me?

ThomasWaldmann: edits to the repository -> write operations to the repository?
if data is deleted -> if data was maliciously/accidentally deleted?

is that what you meant?

docs/usage/notes.rst
```diff
@@ -278,10 +281,19 @@ won't free disk space, they merely tag data as deleted in a new transaction.
...
 
-Note that you can go back-and-forth between normal and append-only operation by editing
-the configuration file, it's not a "one way trip".
+Even if an archive appears to be available, it is possible an attacker could delete
+just a few chunks from an archive and silently corrupt its data. While in append-only
```
ThomasWaldmann: good point

docs/usage/notes.rst
```diff
@@ -278,10 +281,19 @@ won't free disk space, they merely tag data as deleted in a new transaction.
...
+mode, this is reversible, but ``borg check`` should be run before a writing/pruning
+operation on an append-only repository to catch accidental or malicious corruption::
+
+    # run on the repository host, without append-only mode
+    borg check --verify-data repo && borg compact repo
```
ThomasWaldmann: repo host = machine that runs borg serve?

one often does not have the key there, so you can't run borg check for archives there.

docs/usage/notes.rst
```diff
@@ -278,10 +281,19 @@ won't free disk space, they merely tag data as deleted in a new transaction.
...
+
+Aside from checking repository & archive integrity you may want to also manually check
+backups to ensure their content seems correct.
```
ThomasWaldmann: not sure that is practical, but yet we can not recommend something better.

docs/usage/notes.rst
milkey-mouse: That sentence was already in the docs, I just moved it to a more logical place. I could change it to say "with `borg config`"?

docs/usage/notes.rst
milkey-mouse: That's what I meant. I saw several issues where people were confused about aspects of append-only mode mentioned in the docs. I thought spelling out "a cron job that writes to the repository will make append-only mode useless" would make it more clear.

docs/usage/notes.rst
milkey-mouse: I forgot the terminology aside from "host," I'll change that.

I didn't consider that `borg check` would need to decrypt—I forgot the "A" of "AEAD" requires decryption. I guess I'll just say "run without append-only mode".

docs/usage/notes.rst
milkey-mouse: I greatly prefer the solutions in #2251, #1772, and one other issue I'm forgetting the ID for which described a command kind of like `git filter-branch` which would run a shell script on each backup to check the data, but all of these are out of scope of this issue.

docs/usage/notes.rst
ThomasWaldmann: ok, please do.

docs/usage/notes.rst
ThomasWaldmann: ok.

docs/usage/notes.rst
ThomasWaldmann: it really needs to decrypt, so it is also the "E".

so just say "run without append-only mode" or "... in non-append-only mode".

PR #4389, opened 2019-02-23
timestamp for borg delete --info added, fixes #4359

### LINKED ISSUE #4359: Feature Request: --list for delete command

joey4:

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1079,14 +1079,14 @@ class Archiver:
             deleted = False
             for i, archive_name in enumerate(archive_names, 1):
                 try:
-                    del manifest.archives[archive_name]
+                    current_archive = format_archive(manifest.archives.pop(archive_name))
                 except KeyError:
                     self.exit_code = EXIT_WARNING
                     logger.warning('Archive {} not found ({}/{}).'.format(archive_name, i, len(archive_names)))
                 else:
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    logger.info(msg.format(archive_name, i, len(archive_names)))
+                    logger.info(msg.format(current_archive, i, len(archive_names)))
             if dry_run:
                 logger.info('Finished dry-run.')
             elif deleted:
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -1079,14 +1079,14 @@ def _delete_archives(self, args, repository):
...
-                    del manifest.archives[archive_name]
+                    current_archive = format_archive(manifest.archives.pop(archive_name))
                 except KeyError:
                     self.exit_code = EXIT_WARNING
                     logger.warning('Archive {} not found ({}/{}).'.format(archive_name, i, len(archive_names)))
                 else:
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    logger.info(msg.format(archive_name, i, len(archive_names)))
+                    logger.info(msg.format(current_archive, i, len(archive_names)))
```
ThomasWaldmann: better do the format_archive() call here - also for consistency (see code below).

src/borg/archiver.py
```diff
@@ -1079,14 +1079,14 @@ def _delete_archives(self, args, repository):
             deleted = False
             for i, archive_name in enumerate(archive_names, 1):
                 try:
-                    del manifest.archives[archive_name]
+                    current_archive = format_archive(manifest.archives.pop(archive_name))
```
ThomasWaldmann: better do the format_archive() call later (see below).

when doing it here, the exception handler here covers more code than wanted and it would be bad if a KeyError would be raised inside format_archive code and you catch it here also.

ThomasWaldmann:
a hint about workflow:

i see you do your PRs from your master branch. better do them from a feature/bugfix branch you just create for your PR (e.g. `git checkout -b fix4359`).

that way you can update your master at any time (from borgbackup/borg master) and use it as a base for other work or to rebase stuff onto it or ...

src/borg/archiver.py
joey4: okay, i'll do that and thank you for the tip with the feature branch. :)

PR #4409, opened 2019-02-28
crypto: LibreSSL has HMAC_CTX_free and *HMAC_CTX_new

bket:
While here improve spacing a bit (indentation of 4 spaces instead of 3).


ThomasWaldmann:
thanks!

PR #4422, opened 2019-03-04
incremental repo check

### LINKED ISSUE #1657: Incremental repository check
E.g. `borg check [--repository-only] --incremental`. Checks all data since last borg check --incremental invocation. Saves last checked segment into the repository config, continues from there when called again.

Guess this could be extended to archives as well by storing the last --incremental timestamp in the manifest.

bounty: https://www.bountysource.com/issues/38004309-incremental-repository-check

ThomasWaldmann:
fixes #1657.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -290,8 +290,11 @@ class Archiver:
         if args.repo_only and any((args.verify_data, args.first, args.last, args.prefix)):
             self.print_error("--repository-only contradicts --first, --last, --prefix and --verify-data arguments.")
             return EXIT_ERROR
+        if args.repair and args.max_duration:
+            self.print_error("--repair does not allow --max-duration argument.")
+            return EXIT_ERROR
         if not args.archives_only:
-            if not repository.check(repair=args.repair, save_space=args.save_space):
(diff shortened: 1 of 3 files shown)


textshell:
From a short look this looks good.

I'd prefer not to write to config for non critical operations. But that would complicate the code a lot. But if users are going to check a repo there might actually be some storage breakage which this could make even worse. (one way to do that would be to have an auxiliary file that is like config and that stores all keys outside of a manually maintained critical set)

ThomasWaldmann:
True.

But otoh the only really critical thing in the config is the key (in repokey modes) and we advise to have a key backup.

Also, the config is not directly overwritten, but first written to a tmp name and then renamed over the previous config (plus some syncs).

textshell:
Yes they are both true. But i think i explicitly ruled out using config for nounce because i wanted to avoid having frequently changing information in config.

And yes, that key is primarily what we want to protect. And maybe the structure so borg still recognizes the directory are repo.

ThomasWaldmann:
I'm merging this as is now and will open a ticket about separating config into 2.

ThomasWaldmann:
see #4428.

PR #4435, opened 2019-03-11
docs: fix spelling errors

akozlins:
some spelling fixes:
situation were -> situation where
kiB -> KiB (see https://en.wikipedia.org/wiki/Kibibyte)
posess -> possess


ThomasWaldmann:
thanks. learned something, uppercase "K" in "KiB". :)

i just assumed consistency with the small "k" like in "kg".

ThomasWaldmann:
guess there might be other places in borg docs and code with the lowercase "k"?

PR #4443, opened 2019-03-15
Shell completions: borg diff second archive

SanskritFritz:
Added shell completions for
```borg diff repo::archive1 archive2```


ThomasWaldmann:
same here :)

PR #4450, opened 2019-03-18
freebsd xattr namespace fixes

ThomasWaldmann:
see commit comments.

### DIFF
--- a/src/borg/helpers/checks.py
+++ b/src/borg/helpers/checks.py
@@ -35,5 +35,5 @@ def check_extension_modules():
         raise ExtensionModuleError
     if item.API_VERSION != '1.2_01':
         raise ExtensionModuleError
-    if platform.API_VERSION != platform.OS_API_VERSION or platform.API_VERSION != '1.2_04':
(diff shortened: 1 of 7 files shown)


ThomasWaldmann:
remove freebsd from might-be-broken list in exception handler.
at least in master branch, which tests on freebsd 12, it is not broken.

ThomasWaldmann:
tagging this for a backport to 1.1-maint branch, as this is a bug fix.

it's a bit unpretty that this will change xattr storage format in 1.1.x, but the code is able to deal with the not-namespace-prefixed xattr keys of older backups also.

asomers:
You may want to look at something like https://github.com/xattr/xattr for a more portable API.

ThomasWaldmann:
@asomers i know it, see the ticket we have about it.

ThomasWaldmann:
note: do not backport d79d505d89aac07d6daeefa0f5ffafc3bba75008 as long as 1.1-maint branch vagrant testing is based on freebsd 10.3.

PR #4461, opened 2019-03-20
give invalid repo error msg if repo config not found, fixes #4411

### LINKED ISSUE #4411: Compatibility issue (minor), error message in case of invalid repo.
Compatibility issue (minor), error message in case of invalid repo.

Old version

	jk@paju:~$ borg  --version
	borg 1.1.8

Repo dir is not valid repo:

	jk@paju:~$ borg create ~/tmp::foo ~/tmp
	/home/jk/tmp is not a valid repository. Check repo config.

Repo dir does not exist:

	jk@paju:~$ borg create ~/tmp/foo::foo ~/tmp
	Repository /home/jk/tmp/foo does not exist.

Latest version

	jk@tiira:~$ borg --version
	borg 1.1.9

Repo dir is not valid repo:

	jk@tiira:~$ borg create ~/tmp::foo ~/tmp
	Exception ignored in: <bound method Repository.__del__ of <Repository /home/jk/tmp>>
	  File "/usr/lib/python3/dist-packages/borg/repository.py", line 179, in __del__
	AssertionError: cleanup happened in Repository.__del__
	Local Exception
	  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4455, in main
	  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4387, in run
    	return set_ec(func(args))
	  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 139, in wrapper
    	with repository:
	  File "/usr/lib/python3/dist-packages/borg/repository.py", line 189, in __enter__
	    self.open(self.path, bool(self.exclusive), lock_wait=self.lock_wait, lock=self.do_lock)
	  File "/usr/lib/python3/dist-packages/borg/repository.py", line 396, in open
	    with open(os.path.join(self.path, 'config')) as fd:
	FileNotFoundError: [Errno 2] No such file or directory: '/home/jk/tmp/config'
	
	Platform: Linux tiira 4.4.0-142-generic #168-Ubuntu SMP Wed Jan 16 21:00:45 UTC 2019 x86_64
	Linux: Ubuntu 16.04 xenial
	Borg: 1.1.9  Python: CPython 3.5.2
	PID: 23536  CWD: /home/jk
	sys.argv: ['/usr/bin/borg', 'create', '/home/jk/tmp::foo', '/home/jk/tmp']
	SSH_ORIGINAL_COMMAND: None

Repo dir does not exist:
	
	jk@tiira:~$ borg create ~/tmp/foo::foo ~/tmp
	Repository /home/jk/tmp/foo does not exist.

Please restore the v1.1.8 type of error message in case of invalid repo dir.

user062:
this isn't much different than it was before 1.1.9,i think the simplest solution is to make it like before (as in 1.1.8).

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -393,8 +393,12 @@ class Repository:
         else:
             self.lock = None
         self.config = ConfigParser(interpolation=None)
-        with open(os.path.join(self.path, 'config')) as fd:
-            self.config.read_file(fd)
+        try:
+            with open(os.path.join(self.path, 'config')) as fd:
+                self.config.read_file(fd)
+        except FileNotFoundError:
+            self.close()
+            raise self.InvalidRepository(self.path)
         if 'repository' not in self.config.sections() or self.config.getint('repository', 'version') != 1:
             self.close()
             raise self.InvalidRepository(path)


ThomasWaldmann:
> this isn't much different than it was before 1.1.9,

it is. the traceback now shows the real reason why it fails.

if you catch the exception and replace it by a way more generic error 
message, you might hide the real cause of the issue.

user062:
i agree,but from that perspective i think this is not a bug.

user062:
@ThomasWaldmann 

How about printing the real error

```python
except IOError as e:
    print(e) 
    raise self.InvalidRepository(self.path)

```

ThomasWaldmann:
@user062 this is a bit about pretty error msg vs. informative traceback.

our policy has been to rather show the informative traceback, esp. if the cause is not quite clear.

in this case it could be:
- file not there
- permission issues
- a hardware or fs error while reading the file
- ... (?)

i agree that tracebacks might look unpretty to a normal user's eyes, but this is less bad than a developer having to wonder about what happened, where and why.

ThomasWaldmann:
I looked at the ticket again and for fixing the "compatibility" (concerning past and current error messages), we could catch `FileNotFoundError` first and in that specific case emit the same error msg as previously (invalid repo, check repo config). If the repo config file is not there, the hint to check the repo config is helpful.

For other errors, like permission issues, etc. I suggest to just let it blow up and print the traceback.

user062:
```
python

        try:
            with open(os.path.join(self.path, 'config')) as fd:
                self.config.read_file(fd)
        except FileNotFoundError:
            self.close()

```

this addresses the issue in case of FileNotFoundError.

it passes all tests for me.

do you think i should commit it?

(btw sorry for not replaying no the other pull request)

ThomasWaldmann:
@user062 yup, looks good.

ThomasWaldmann:
force pushed:
- rebase on current master
- squashed commits together
- reworded commit comment

PR #4470, opened 2019-03-22
FIX #4373

### LINKED ISSUE #4373: Incorrect borg prune --stats output with --dry-run
With `borg 1.1.7` on `ubuntu 18.10`:
```
borg prune <borg_dir> --dry-run --list --stats --keep-last 1 
...
Deleted data:                    0 B                  0 B                  0 B
All archives:              981.25 GB            793.40 GB            112.32 GB

Chunk index:                  250054              2164226
```
```
borg prune <borg_dir> --list --stats --keep-last 1 
...
Deleted data:             -820.45 GB           -669.72 GB             -9.60 GB
All archives:              160.80 GB            123.69 GB            102.72 GB

Chunk index:                  237044               371408
```
Pruning worked as I wanted it to, but the dry run stats do not give an accurate preview.

user062:
couldn't find a better place to put the code other than `archiver.run`,
 because `--dry-state` and `--state` are both arguments of more than one command.


src/borg/archiver.py
```diff
@@ -4230,6 +4230,9 @@ def run(self, args):
         args.progress |= is_serve
         self._setup_implied_logging(vars(args))
         self._setup_topic_debugging(args)
+        if all((hasattr(args, x) for x in ('stats', 'dry_run'))) and all((args.stats, args.dry_run)):
```
ThomasWaldmann: you could combine `hasattr` and the actual access of `args.*` like:
```
if getattr(args, 'stats', False) and getattr(args, 'dry_run', False): ...
```

src/borg/archiver.py
```diff
@@ -4230,6 +4230,9 @@ def run(self, args):
...
+            logger.error("--stats and --dry-run shouldn't be used together")
```
ThomasWaldmann: guess users trying that will then ask themselves (or us): **why?**

you could also just tell "--stats does not work with --dry-run".

src/borg/testsuite/archiver.py
```diff
@@ -1870,10 +1870,9 @@ def test_prune_repository_save_space(self):
         self.cmd('init', '--encryption=repokey', self.repository_location)
         self.cmd('create', self.repository_location + '::test1', src_dir)
         self.cmd('create', self.repository_location + '::test2', src_dir)
-        output = self.cmd('prune', '--list', '--stats', '--dry-run', self.repository_location, '--keep-daily=2')
+        output = self.cmd('prune', '--list', '--dry-run', self.repository_location, '--keep-daily=2')
```
ThomasWaldmann: oops. good catch.

src/borg/archiver.py
ThomasWaldmann: having this code here means that `--stats` is *never* supported when using `--dry-run` (which might be true, did you check all places?).

src/borg/archiver.py
user062: right,actually i just copied the error message from your comment.

src/borg/archiver.py
user062: thanks for refactoring.
i checked `archiver.py`,`docs`,`tests`
tests have nothing but the one i changed.
docs have them for create and delete.
archiver.py `do_create` and `do_delete` and i think `--stats` is never supported when using `--dry-run`

user062:
CI failed for bug #3494

src/borg/archiver.py
user062: is there is a possibility for adding a command in the future which runs both `--stats` and `--dry-run` together? if yes, then i think we should add the restriction per command.
maybe define a function  in `archiver`,and call it on `do_create`, `do_prune`, `do_recreate` (and on any new new method where the restriction applies)
```python
def check_stats_and_dry_run(self, command):
    self.print_error("--stats does not work with --dry-run with %s.", command)
    return set_ec(self.exit_code)

```

ThomasWaldmann: well, that of course depends on the meaning of the options for that command.

for the ones you checked (what we currently have), it does not make sense.

so it is ok for now and in case that will change in future, we need to change the code for that.
but that will be rather obvious, so no problem, keep it as it is (except doing the simplification I suggested).

user062: Done.
by the way i edited the comment above to clarify what i meant because i forgot a line in the function.

ThomasWaldmann:
@user062 thanks for the PR! it could get a backport to 1.1-maint, do you want to do that?

Issue #4471, opened 2019-03-23
borg help should print <command> list in sorted order


fantasya-pbem:
When calling "borg help", in section "required arguments: <command>" it lists the available commands unsorted. They should be sorted alphabetically.

@ThomasWaldmann: I'm not a Python programmer, but I looked into archiver.py and think I can fix that by myself. I consider this issue a good first one for me.

ThomasWaldmann:
sounds good :) just make a pull request.

PR #4480, opened 2019-03-29
Fix for #4471 - borg help should print sorted command list

fantasya-pbem:
Sorted the help parsers in 15 steps for easier review.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3033,6 +3033,236 @@ class Archiver:
         subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
                                help='paths to archive')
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
all commits seem fine, but that one still is too much and creates a diff display mess:

https://github.com/borgbackup/borg/pull/4480/commits/a5e98125db96d5b932e659079b44127e08108059

also, some commits need rewording of the commit comment - you tell they are about "borg debug x", but they are really about "borg x".

fantasya-pbem:
Thank you for feedback. In my IDE that diff was fine... I will do some Git Rebase magic and make a new PR.

PR #4481, opened 2019-03-29
experiments related to tests hanging at the end

ThomasWaldmann:


ThomasWaldmann:
Merged this. Does not fix the issue, but due to -v and no parallelism, it makes it more clear where problems happen (and excludes parallelism as a source of problems).

PR #4487, opened 2019-04-02
do not check python/libc for borg serve, fixes #4483

### LINKED ISSUE #4483: only do glibc checks on client?
Copying my reply to an issue posted on the mailing list to here:

```
     $ borg init --e none ./borgtest
     Remote: FATAL: this Python was compiled for a too old (g)libc and
     misses required functionality.
```
This is the code which triggers this exception / error message:

```
    required_funcs = {os.stat, os.utime, os.chown}
    if not os.supports_follow_symlinks.issuperset(required_funcs):
        raise PythonLibcTooOld
```

So, as you see, quite basic functionality like not following symlinks is not supported for all function where borg needs that.

So far for the bad news.

The good news are that these functions might be only required for using borg as a client, but not for "borg serve" (repo server part in client/server mode, if repo access is done via ssh:).

borg currently always does this check, but I guess it could be maybe omitted for the "borg serve" server part.

ThomasWaldmann:
the check checks whether follow_symlinks=False is supported, which
requires that the glibc is recent enough / python was compiled
for a recent enough glibc. follow_symlinks=False is only used for borg
create and extract, but not needed for borg serve.

thus, this makes it possible to run "borg serve" even on a bit older
servers. be careful, due to the presence of this check on the server
side until now, older server systems are not really tested.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4322,8 +4322,10 @@ class Archiver:
             parser.error('Need at least one PATH argument.')
         return args
 
-    def prerun_checks(self, logger):
-        check_python()
+    def prerun_checks(self, logger, is_serve):
+        if not is_serve:
+            # this is the borg *client*, we need to check the python:
+            check_python()
         check_extension_modules()
         selftest(logger)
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
this needs practical testing. if you did, please post in ticket #4483.

ThomasWaldmann:
master forward port: see #4528

Issue #4485, opened 2019-04-01
How to gracefully kill borg on remote machine


marcpope:
## Have you checked borgbackup docs, FAQ, and open Github issues?
Yes

## Is this a BUG / ISSUE report or a QUESTION?
Question

## System information. For client/server mode post info for both machines.
Backing up from a client machine to a server over network 

#### Your borg version (borg -V). 
1.1.3  (I know I need to upgrade).

#### Operating system (distribution) and version. 
CentOS 7.6

#### Hardware / network configuration, and filesystems used.
Local Network, GB Connection, CentOS 6.7, default LVM type filesystems

#### How much data is handled by borg?
500-600gb in this instance, about 6.7million files, doesn't seem to matter in this case.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
borg create --list someuser@hostname:reponame::e9gKdQDJ-tjU7ZKwvebfq2mLWNkOy /chroot/home
    (note: private info redacted)

## Describe the problem you're observing.
I don't know if this is necessarily a problem, or a feature. Let's say a backup is in progress. The client is backing up to the server and the client crashes or restarts. The server leaves the process open, for a period of time. (Eventually after a few hours, roughly, it seems to timeout). Would there be a safe way to kill this process and free the locks? Should anything be done to make sure the next backup is clean or can another backup just continue?

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.
Yes, everytime.

#### Include any warning/errors/backtraces from the system logs
n/a

ThomasWaldmann:
you can maybe solve this on ssh level.
if the borg client dies, this means the ssh client is dying also.
if the ssh server checks the connection regularly, it should notice the client and the connection has died and terminate itself also, also terminating the server side "borg serve" process.

marcpope:
so if the server can see that the client is no longer running the borg process, is it safe to kill the process? then run borg break-locks on the repo? I have found that the server eventually terminates at some predefined period on it's own, but it seems to be quite a while (maybe an hour or more).

ThomasWaldmann:
If there is no ongoing activity, yes it should be safe to kill the process.

Preferably, send it a SIGTERM.

There is some code in borg that should clean up locks when it terminates, so you maybe do not even have to remove them afterwards.

mikenye:
See issue #3988.

[...]

Issue #3988, opened 2018-07-18
Broken Pipe (SSH): Break Lock


margau:
Hello,
sometimes my SSH-Connection breaks, but the lock on the server is still active:
```
Remote: packet_write_wait: Connection to 84.x.x.x port 22: Broken pipe
Connection closed by remote host
```
If I try to restart, borg can't acquire the lock, I have to manually break it on the server.

My idea: When the Server (borg serve) notices the broken pipe, it automatically breaks the lock, because the client has obviously gone away.

Best regards 
margau

ThomasWaldmann:
Before you break the lock, check if the borg serve process is still alive (and has the lock).

Also add your borg version to the top post.

margau:
Borg Version is 1.1.6 - in my case, the SSH-Pipe breaks sometimes, but borg serve is still alive.
Something to automatically "heal" this case would be great.

margau:
I have an hourly cronjob, it ran into problems until I manually ran break-lock. But since I updated to 1.1.6 (had the Debian version before) this didn't happen again.

ThomasWaldmann:
@margau is the server-side ssh process also still running in such a case?

margau:
Yes, that's what I meant with borg-serve. The client process exits in such a case with "broken pipe".

ThomasWaldmann:
Well, borg serve's pipes (stdin/stdout/stderr) are connected to the sshd process on the server side. So if the sshd is still running, there is no broken pipe from borg's perspective.

margau:
Would it be an Option to kill borg-serve if no data was received for 15 minutes?

ThomasWaldmann:
Well, if the client is a bit on the slow side and needs a long time for some local computation without server communication, the server would kick it - not nice.

How about just using ssh's features for that kind of stuff?

knutov:
It's not the solution when plug is off for some time, it's for the case when tcp connect was broken somehow but immediately new connect is possible to create.

Works perfect for sshfs for example.

margau:
Hello, from Systemd I see the user session is gone:
`Jul 26 16:16:04 x systemd[1]: Stopping User Manager for UID 1003...
Jul 26 16:16:04 x systemd[64347]: Stopped target Default.
Jul 26 16:16:04 x systemd[64347]: Stopped target Basic System.
Jul 26 16:16:04 x systemd[64347]: Stopped target Sockets.
[...]

Issue #636, opened 2016-02-06
connection closed by remote


ThomasWaldmann:
sometimes a borg backup gets interrupted with "connection closed by remote".

it is unclear why this happens or whether this is a problem in borg, in the ssh configuration or on the network layer.

Suspicions:
- ssh / tcp keepalive configuration (this was pointed out on the atticmatic site, but it did not help in a experiment a borg user did), so it either does not help or is sometimes not sufficient (from IRC: Azelphur$  FYI, TCPKeepAlive yes, ServerAliveInterval 60, ServerAliveCountMax 5 didn't work :( )
  See also: http://www.cyberciti.biz/tips/open-ssh-server-connection-drops-out-after-few-or-n-minutes-of-inactivity.html
- borg: EINTR and in general exception handling maybe could be improved in remote.py (though it is unclear what would cause EINTR in our case and would not terminate the process anyway). See there for details: http://250bpm.com/blog:12
- there might be longer times with no or only little data transfer (e.g. if borg is reading over a lot of zeros in a sparse file or if all the chunks are already stored in the repo - e.g. when backing up VMs)
- a router might be configured in its security/DoS settings to do something special in case a lot of packets arrive - could this be triggered by borg?
- also (trivial), it might be just a unstable network connection that breaks down

ThomasWaldmann:
Idea for maybe getting more infos from the server side: instead of relying on the log messages that come over the wire (or rather don't come if the connection gets killed), configure the borg on the server to use logging to a local file.

logging.conf: see pastebin (url coming soon)

export BORG_LOGGING_CONF=/path/to/that/logging.conf
^ that needs to be put into the environment ssh users (like the borg client) get.
(not sure whether we need a borg wrapper script here or a sshd config change)

ThomasWaldmann:
TODO: update the FAQ and point to here.

Azelphur:
The borg log sadly doesn't actually contain any useful information, it just says
```
2016-02-08 19:14:54,477 borg.logger DEBUG using logging configuration read from "/home/azelphur/logging.conf"
```
Here is the log from the client side, running with BORG_RSH="ssh -v"

```
# BORG_RSH="ssh -v" time /home/azelphur/.borg-backup
Backing up to ssh://azelphur@backup.azelphur.com:23/home/azelphur/darth-vader.borg
Remote: OpenSSH_6.9p1 Ubuntu-2ubuntu0.1, OpenSSL 1.0.2d 9 Jul 2015
[...]

Issue #448, opened 2015-11-26
support request: borg cache dir: is it possible to move cache dir once it is created?


DrupaListo-com:
1. is it possible to move cache dir once it is created?
2. Possible to use some --cache-dir-location type of switch? I saw its used in 'borg delete' - the "-c" switch and it works - deletes the cache in $HOME/.cache/borg

So should I symlink $HOME/.cache/borg to the real location I desire or it IS possible or easy to make it possible in future release to give the cache dir on 'borg init' or 'borg create'

Making sense ? @ThomasWaldmann This one should be relatively easy topic to think about - the location of the cache dir...

ThomasWaldmann:
1. yes (including its content)
2. http://borgbackup.readthedocs.org/en/stable/usage.html#environment-variables or just use a symlink

AnnoyingTechnology:
updated link https://borgbackup.readthedocs.io/en/stable/usage/general.html#environment-variables

PR #4493, opened 2019-04-04
Add "SSH Configuration" section

mikenye:
Add "SSH Configuration" section to "borg serve" documentation page. This new section documents the additional ssh configuration necessary to prevent the borg serve process from keeping a lock on a repo in the event the ssh connection is abnormally disconnected (leaving borg serve running until sshd times out, which can take upwards of an hour in some cases). This pull request is in response to issues #3988, #636 and #4485 (and probably others).

### DIFF
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -55,7 +55,7 @@ In the client side's ``~/.ssh/config`` file:
 
 Replacing ``backupserver`` with the hostname, FQDN or IP address of the borg server.
 
-This will cause the client to send a keepalive to the server every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the borg process to terminate gracefully.
+This will cause the client to send a keepalive to the server every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the ssh client process will be terminated, causing the borg process to terminate gracefully.
 
 On the server side's ``sshd`` configuration file (typically ``/etc/ssh/sshd_config``):
 ::
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -46,7 +46,7 @@ SSH Configuration
 
 In order to avoid this, it is recommended to perform the following additional SSH configuration:
 
-In the client side's ``~/.ssh/config`` file:
+Either in the client side's ``~/.ssh/config`` file, or in the client's ``/etc/ssh/ssh_config`` file:
 ::
 
     Host backupserver
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
     has been introduced in v7.2. We recommend to use
     ``no-port-forwarding,no-X11-forwarding,no-pty,no-agent-forwarding,no-user-rc``
     in this case.
+    
+SSH Configuration
+~~~~~~~~~~~~~~~~~
+``borg serve``'s pipes (``stdin``/``stdout``/``stderr``) are connected to the ``sshd`` process on the server side. In the event that the SSH connection between ``borg serve`` and the client is disconnected abnormally (for example, due to a network outage), it can take a long time for ``sshd`` to notice the client is disconnected. In the meantime, ``sshd`` continues running, and as a result so does the ``borg serve`` process holding the lock on the repository. This can cause subsequent ``borg`` operations on the remote repository to fail with the error: ``Failed to create/acquire the lock``.
+
+In order to avoid this, it is recommended to perform the following additional SSH configuration:
+
+In the client side's ``~/.ssh/config`` file:
+::
+
+    Host backupserver
+            ServerAliveInterval 10
+            ServerAliveCountMax 30
+
+Replacing ``backupserver`` with the hostname or FQDN of of the borg server.
+
+This will cause the client to send a keepalive to the server every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the borg process to terminate gracefully.
+
+On the server side's ``sshd`` configuration file (typically ``/etc/ssh/sshd_config``):
+::
+
+    ClientAliveInterval 10
+    ClientAliveCountMax 30
+
+This will cause the server to send a keep alive to the client every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the ``borg serve`` process to terminate gracefully and release the lock on the repository.
+
+If you then run borg commands with ``--lock-wait 600``, this gives sufficient time for the borg serve processes to terminate after the SSH connection is torn down after the 300 second wait for the keepalives to fail.
+
+You may, of course, modify the timeout values demonstrated above to values that suit your environment and use case.
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -42,7 +42,7 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
     
 SSH Configuration
 ~~~~~~~~~~~~~~~~~
-``borg serve``'s pipes (``stdin``/``stdout``/``stderr``) are connected to the ``sshd`` process on the server side. In the event that the SSH connection between ``borg serve`` and the client is disconnected abnormally (for example, due to a network outage), it can take a long time for ``sshd`` to notice the client is disconnected. In the meantime, ``sshd`` continues running, and as a result so does the ``borg serve`` process holding the lock on the repository. This can cause subsequent ``borg`` operations on the remote repository to fail with the error: ``Failed to create/acquire the lock``.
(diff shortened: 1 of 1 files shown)


mikenye:
Typo. Issue #448 is incorrect. It should reference issue #4485. My apologies.

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+::
+
+    ClientAliveInterval 10
+    ClientAliveCountMax 30
+
+This will cause the server to send a keep alive to the client every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the ``borg serve`` process to terminate gracefully and release the lock on the repository.
+
+If you then run borg commands with ``--lock-wait 600``, this gives sufficient time for the borg serve processes to terminate after the SSH connection is torn down after the 300 second wait for the keepalives to fail.
+
+You may, of course, modify the timeout values demonstrated above to values that suit your environment and use case.
```
ThomasWaldmann: guess that is somehow clear, the only question remaining is what the safe minimum would be.

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
     has been introduced in v7.2. We recommend to use
     ``no-port-forwarding,no-X11-forwarding,no-pty,no-agent-forwarding,no-user-rc``
     in this case.
+    
+SSH Configuration
+~~~~~~~~~~~~~~~~~
+``borg serve``'s pipes (``stdin``/``stdout``/``stderr``) are connected to the ``sshd`` process on the server side. In the event that the SSH connection between ``borg serve`` and the client is disconnected abnormally (for example, due to a network outage), it can take a long time for ``sshd`` to notice the client is disconnected. In the meantime, ``sshd`` continues running, and as a result so does the ``borg serve`` process holding the lock on the repository. This can cause subsequent ``borg`` operations on the remote repository to fail with the error: ``Failed to create/acquire the lock``.
```
ThomasWaldmann: ... is disconnected or stuck abnormally ...

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+
+In order to avoid this, it is recommended to perform the following additional SSH configuration:
+
+In the client side's ``~/.ssh/config`` file:
```
ThomasWaldmann: on the server side, you do the config change globally, but here only for the backupserver host.

any special reason for this? it might count as defensive, but on the other hand: which scenario could there be that does not want to terminate the connection after it has been non-working for 5 minutes?

mikenye:
force-push to fix typo in initial commit

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+::
+
+    Host backupserver
+            ServerAliveInterval 10
+            ServerAliveCountMax 30
+
+Replacing ``backupserver`` with the hostname or FQDN of of the borg server.
```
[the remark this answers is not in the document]
mikenye: Fixed in commit dd238bc2

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+
+This will cause the client to send a keepalive to the server every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the borg process to terminate gracefully.
```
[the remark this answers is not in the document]
mikenye: Adjusted wording in commit ec1b45be.

With regards to lower counts, no I didn't. I tried to pick numbers that would work in most people's situations. In the event someone is backing up over a congested link with packet loss, I didn't want the connection to terminate too soon. Happy to change this if you'd prefer shorter timeouts in the documentation. :-)

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+
+On the server side's ``sshd`` configuration file (typically ``/etc/ssh/sshd_config``):
+::
+
+    ClientAliveInterval 10
+    ClientAliveCountMax 30
```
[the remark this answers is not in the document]
mikenye: I did try this, however it didn't seem to make any difference. I can add it in the documentation if you'd like?

docs/usage/serve.rst
```diff
@@ -39,3 +39,32 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
...
+
+This will cause the server to send a keep alive to the client every 10 seconds. If 30 consecutive keepalives are sent without a response (a time of 300 seconds), the connection will be terminated, causing the ``borg serve`` process to terminate gracefully and release the lock on the repository.
```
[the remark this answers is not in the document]
mikenye: Wording adjusted in commit 045df0b0

docs/usage/serve.rst
mikenye: Perhaps I should change the wording to:

```If you then run borg commands with ``--lock-wait 600``, this should give sufficient time for ssh and borg processes to gracefully terminate in the event processes are left running due to a network issue.```

Comment (mikenye) [reply docs/usage/serve.rst]: With regards to a safe minimum, I've found that borg serve needs a little bit of time to "tidy up" once the connection fails. In my environment, I simply doubled it and this works for me, however I'm not sure what a safe minimum would be. What internal tasks is borg serve doing when terminating on connection loss, and what is a safe maximum on how long should these take?

File: docs/usage/serve.rst (hunk 2 above)
Comment (mikenye) [reply docs/usage/serve.rst]: Wording changed in commit a52af99d

File: docs/usage/serve.rst (hunk 6 above)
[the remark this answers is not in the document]
Comment (mikenye) [reply docs/usage/serve.rst]: Some additional information on this. I'm not sure we should be recommending TCPKeepAlive, as the sshd manpage seems to suggest it may be vulnerable to spoofing.

> It is important to note that the use of client alive messages is very different from TCPKeepAlive (below). The client alive messages are sent through the encrypted channel and therefore will not be spoofable. The TCP keepalive option enabled by TCPKeepAlive is spoofable.

File: docs/usage/serve.rst (hunk 3 above)
Comment (mikenye) [reply docs/usage/serve.rst]: This is deliberate. When I add the `ClientAliveInterval` and `ClientAliveCountMax` within a match block, eg:

```
Match User backupuser
    AllowTCPForwarding no
    X11Forwarding no
    ClientAliveInterval 10
    ClientAliveCountMax 30
```

...when sshd is reloaded, it terminates the the following errors in syslog:

```
Apr  5 10:47:10 frac systemd[1]: Reloading OpenBSD Secure Shell server.
Apr  5 10:47:10 frac systemd[1]: Reloaded OpenBSD Secure Shell server.
Apr  5 10:47:10 frac sshd[14143]: /etc/ssh/sshd_config line 98: Directive 'ClientAliveInterval' is not allowed within a Match block
Apr  5 10:47:10 frac systemd[1]: ssh.service: Main process exited, code=exited, status=255/n/a
Apr  5 10:47:10 frac systemd[1]: ssh.service: Unit entered failed state.
Apr  5 10:47:10 frac systemd[1]: ssh.service: Failed with result 'exit-code'.
```

Thus, it looks like these settings need to be made globally on the server side.

On the client side, there is no issues adding these to the user's `~/.ssh/config` file as-per what I've written.

mikenye: If you'd like, I can add some text letting users know the server side configuration items need to be global, to prevent them from trying it in a match block and breaking their sshd and locking themselves out of their server. :-)

docs/usage/serve.rst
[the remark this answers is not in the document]
mikenye: Also thanks for letting me know how to re-word commits - I didn't know this was possible (still learning my way around git).

docs/usage/serve.rst
mikenye: > it might count as defensive, but on the other hand: which scenario could there be that does not want to terminate the connection after it has been non-working for 5 minutes?

I've given this some thought, and I honestly can't think of any... Possibly on highly latent links (satellite) where there may be long periods of inactivity? Hence my "adjust to suit your environment" comment. :-)

docs/usage/serve.rst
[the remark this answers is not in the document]
ThomasWaldmann: git rebase -i is a really nice functionality. i use it often for all sorts of cleanups, like reordering commits, collapsing fixup commits into the original commit, squashing commits together, completely dropping some commits, ...

docs/usage/serve.rst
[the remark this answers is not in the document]
ThomasWaldmann: considering it didn't work for you, we don't need to mention it.

docs/usage/serve.rst
ThomasWaldmann: maybe i phrased it badly. i didn't mean to reduce lock-wait, but i also meant the 300s here.

lock-wait is just the maximum waiting time. if everything works fine, the lock will go away and it will wait less than that.

docs/usage/serve.rst
ThomasWaldmann: what i meant is "why not make it global in /etc/ssh/ssh_config?".

mikenye: Yep, we could do that. I'll adjust the wording.

mikenye: Wording updated in commit f9d011be

mikenye:
I've updated the serve.rst as-per our discussions. Are you happy with how it reads?

ThomasWaldmann:
very happy! thanks! :)

i merged it into master - do you want to try a backport to 1.1-maint, so it'll get into the next 1.1.x release?

mikenye:
Yep, I'd love to. I'll do it today. :-)

PR #4497, opened 2019-04-06
create: only run stat_simple_attrs() once

ThomasWaldmann:
the second call was done in stat_attrs().

this increases backup with lots of unchanged files performance by ~ 5%.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1195,6 +1195,7 @@ class FilesystemObjectProcessors:
             with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags, noatime=True) as fd:
                 with backup_io('fstat'):
                     st = stat_update_check(st, os.fstat(fd))
+                item.update(self.metadata_collector.stat_simple_attrs(st))
                 is_special_file = is_special(st.st_mode)
                 if not hardlinked or hardlink_master:
                     if not is_special_file:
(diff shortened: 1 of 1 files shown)


jdchristensen:
Maybe I'm misreading the code, but I don't see any fewer calls to stat_simple_attrs.  One call got moved out of an if, so it will happen more often, not less often.  And another call got slightly changed to avoid a temporary variable.

jdchristensen:
Ah, nevermind, I see now that the second change calls stat_ext_attrs instead of stat_attrs, and that's probably more efficient.

ThomasWaldmann:
@jdchristensen stat_attrs just calls simple and ext. so just calling ext is 1 simple less.

ThomasWaldmann:
this was also backported to 1.1-maint now in #4530.

FabioPedretti:
Hi, just some feedback here. I did some test with 1.1.10 (which should include this fix) vs 1.1.9 and I fail see any improvement in speed.
I tested both with a benchmark script posted elsewhere (20000 empty files) and my own production backup (millions of files, few changes between runs).
Actually, after 9 benchmark and production runs, it looks 1.1.10 is a bit slower than 1.1.9.
But as I said elsewhere my tests are not very reproducible due to my systems being used by other stuff.

ThomasWaldmann:
@FabioPedretti the reason why you don't see an improvement is likely due to OS level caching, so the 2nd stat() call is handled in-memory from cache and does not cause real I/O. this is especially likely for small files.

PR #4528, opened 2019-04-27
2 forward ports

shawnl:


ThomasWaldmann:
For #4483, I first wanted to wait until somebody has tested this on a older system.

While the borg change looks harmless, due to the presence of that check, it borg has never been tested on older systems than the ones succeeding the test, so there might be other issue.

But until now, it seems like noone has tested it and I guess I rather won't merge this change until someone did.

PR #4530, opened 2019-04-27
2 back ports

shawnl:


ThomasWaldmann:
https://github.com/borgbackup/borg/pull/4497/files please check this again, it seems you have not backported the lower part.

also please check that in all possible code paths it will call "simple" once and "ext" once.

shawnl:
the lower part did nothing,and shouldn't have been in a patch to be backported.
```
                md = self.metadata_collector.stat_attrs(st, path, fd=fd)
                item.update(md)
```
```
                item.update(self.metadata_collector.stat_ext_attrs(st, path, fd=fd))
```
Also it isn't applicable, as there is no metadata collector in the old code.

ThomasWaldmann:
Look again, the lower part of the master patch was the main thing.

shawnl:
oh wow! yeah, and ext and simple are both called for every code path (it all goes to `return status`, and those are both at the same code level.)

ThomasWaldmann:
ok, looks good now! thanks for helping!

PR #4531, opened 2019-04-27
use whitelist approach for borg serve, fixes #4097

### LINKED ISSUE #4097: specify options NOT to enforce/override via borg serve
== whitelist options determined client-side (and everything else is enforced server-side) - this avoids such security issues.

See there: https://github.com/borgbackup/borg/pull/4095#pullrequestreview-162266367

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/64379886-specify-options-not-to-enforce-override-via-borg-serve)

ThomasWaldmann:
for borgbackup's client/server mode security, it is essential that some stuff can be enforced server-side (e.g. quota, access only to specific repos).

the way this is done in the code should be resistant against "forgetting" to enforce if new stuff is added.
this was not the case until this PR, but should be fixed now.

please review / give feedback, see #4097.

a test for this (`test_get_args`) already existed before and still succeeds.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4214,20 +4214,43 @@ class Archiver:
         """usually, just returns argv, except if we deal with a ssh forced command for borg serve."""
         result = self.parse_args(argv[1:])
         if cmd is not None and result.func == self.do_serve:
-            forced_result = result
-            argv = shlex.split(cmd)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
consider backporting this to 1.1-maint later.

PR #4543, opened 2019-05-06
delete / prune archive(s): consider part files correctly for stats (1.1-maint)

ThomasWaldmann:


ThomasWaldmann:
hmm, this also needs #4391 (which also adds new stuff to archive metadata).

guess this is a bit too much for a backport...

ThomasWaldmann:
see above.

PR #4544, opened 2019-05-06
small borg compact improvements, fixes #4522

### LINKED ISSUE #4522: borg compact -v could output some info
I tried `borg compact -v` in 1.2.0a6 and noticed it shows no output. I tried `borg compact --debug` and indeed it shows debug output.

`-v` usually shows some useful info/stats with other commands, it would be nice if borg compact could also output some basic stats (space freed or something else meaningful).

Also a couple tips related to compact (maybe I should open a new issue):
- the web page (https://borgbackup.readthedocs.io/en/latest/usage/compact.html) should details a bit more the description of --cleanup-commits , e.g. something like `"cleanup commit-only 17-byte segment files generated by borg 1.1. Use this one time after upgrading borg (server) to 1.2+"`, removing the same info from the Example;
- the docs should also clarify when it is useful to run compact, and also which commands may need it. I suppose it is only useful after a delete/prune. If I just run creates (and never delete/prune) is it still needed? BTW in 1.1 in which commands was executed?

ThomasWaldmann:


FabioPedretti:
This is what I had in mind, thanks!

Issue #4540, opened 2019-05-05
Need ability to specify tmpdir for remote


Endareth:
## Have you checked borgbackup docs, FAQ, and open Github issues?
Yes

## Is this a BUG / ISSUE report or a QUESTION?
Issue

#### Your borg version (borg -V).
borg 1.1.5/borg 1.1.9

#### Operating system (distribution) and version.
Ubuntu 18.04 (local), QNAP 4.3.6.0923

#### Hardware / network configuration, and filesystems used.
Various, n/a to issue

#### How much data is handled by borg?
Various, n/a to issue

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
Any use of remote repo

## Describe the problem you're observing.
Temp directory on remote full. The remote in question is a QNAP NAS, and configured with a tiny /tmp directory (64MB!).

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.
Running multiple backups results in multiple /tmp/_M* directories on remote. While some calls seem to cleanup after themselves, I just re-checked and saw multiple copies of that those _M tmp directories containing all the borg binaries/libraries.

#### Include any warning/errors/backtraces from the system logs
`Failed to write all bytes for _bisect.so`
From running a borg command remotely once the /tmp drive is full.

So to summarise, due to the QNAP NAS having a way too small /tmp directory (only 64MB allocated to it), being able to specify the tmp directory for a remote, either via env variable or command line option, would be a huge help.

fantasya-pbem:
According to doc, BorgBackup should check temp dir env variables. See here:
https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

ThomasWaldmann:
https://borgbackup.readthedocs.io/en/stable/usage/general.html#environment-variables

see last item there.

ThomasWaldmann:
For a "borg serve" server, you need to have this in the environment of the "borg serve" process, of course.

Endareth:
Thanks for the tips. Looks like QNAP doesn't set any of the temp variables in a standard sshd session. I was able to finally get it working by setting `PermitUserEnvironment yes` in the sshd_config, then adding `environment="TEMP=/root/tmp"` to the start of each of the SSH public key entries in authorized_hosts to use a different location. Rather clunky, but at least it works. Would still be nice to have an safer way of setting this for the borg remote, to avoid having to script changes to my sshd_config and authorized_keys file.

[...]

PR #4545, opened 2019-05-07
Updated doc to cover use of temp directory on remote

Endareth:
PR raised as per request in https://github.com/borgbackup/borg/issues/4540
Added some doco regarding use of temp directory on remote, in particular cases where default tmp directory is not appropriate for use.

### DIFF
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -404,7 +404,16 @@ Temporary files (client):
     won't be locally cached.
 
 Temporary files (server):
-    None.
+    A non-trivial (potentially 40-50 MB) amount of data will be stored on the
+    remote temp directory for each client that connects to it. For some remotes,
+    in particular QNAP NAS systems, this can fill the default (tiny!) temporary
+    directory at /tmp. This can be remediated by ensuring the $TMPDIR, $TEMP,
+    or $TMP environment variable is properly set for the sshd process. For some
+    OSes, this can be done just by setting the correct value in the .bashrc (or
+    equivalent login config file for other shells), however in other cases it
+    may be neccessary to first enable ``PermitUserEnvironment yes`` in your
+    ``sshd_config`` file, then add ``environment="TMPDIR=/my/big/tmpdir"`` at
+    the start of the public key to be used in the ``authorized_hosts`` file.
 
 Cache files (client only):
     Contains the chunks index and files index (plus a collection of single-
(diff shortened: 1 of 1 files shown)


jdchristensen:
I haven't tested this, but maybe it would be easier to specify something like

    --remote-path='/usr/bin/env TMPDIR=/my/big/tmpdir borg'

on the client borg command line if you want to set an environment variable for borg.  Then you don't need to edit sshd_config, which requires root privileges.

ThomasWaldmann:
@jdchristensen yeah, if the repo has just one client, that indeed looks simpler (just needs to be tested).

for a repo server with multiple / many clients, it might be easier though to do the change server-side.

@Endareth did you try whether setting the env var somewhere else also works (so one does not have to do it per authorized_keys entry, but just once for the "borg" user - assuming that all backups go to borg@backupserver)?

Endareth:
If setting the TMPDIR via ``--remote-path`` works, that'd definitely be a good option. As for setting it via other methods, that varies by OS. sshd on some systems will still respect the .bashrc config file, some will use ``~/.ssh/environment``. Unfortunately QNAP doesn't do either, and that's what I was working with in this instance. I'll test the ``--remote-path`` method and update the doco with the results.

docs/usage_general.rst.inc
```diff
@@ -404,7 +404,16 @@ Temporary files (client):
     won't be locally cached.
 
 Temporary files (server):
-    None.
+    A non-trivial (potentially 40-50 MB) amount of data will be stored on the
```
ThomasWaldmann: does it make sense to give a specific amount here?
guess it could vary widely, depending on repo size and what you do?

docs/usage_general.rst.inc
```diff
@@ -404,7 +404,16 @@ Temporary files (client):
...
+    remote temp directory for each client that connects to it. For some remotes,
+    in particular QNAP NAS systems, this can fill the default (tiny!) temporary
```
ThomasWaldmann: i don't think it is necessary to refer to a specific brand here.

likely all systems that mount /tmp on a tmpfs and don't have huge amounts of RAM (and a huge tmpfs size) are having this issue.

similar if /tmp is just part of the on-disk root filesystem and that one wasn't made with too much free space.

ThomasWaldmann:
@Endareth I want to release 1.1.10 soon, if you could finish testing and update the PR, it could be merged before release.

ThomasWaldmann:
ping.

ThomasWaldmann:
Fixed it.

PR #4560, opened 2019-05-12
Bundle msgpack

ThomasWaldmann:
trying to bundle msgpack 0.5.6 into borg, based on patches @FelixSchwarz used for Fedora (thanks!).

TODO:
- tests
- try to make it switchable, similar to how compression libs are handled (update: thinking about that, it might be unneccessary for borg 1.1.x because the msgpack 0.5.6 we will have bundled is the latest and greatest that is supported by borg anyway, so there is not much point in taking it from the system).
- check master branch msgpack adapter for the path structure, so it could be maybe adopted for master branch also.


FelixSchwarz:
> try to make it switchable, similar to how compression libs are handled (update: thinking about that, it might be unneccessary for borg 1.1.x because the msgpack 0.5.6 we will have bundled is the latest and greatest that is supported by borg anyway, so there is not much point in taking it from the system).

Generally I'd prefer easy unbundling of bundled code. For Fedora this probably doesn't matter too much in the context of borg 1.1.x as we will need to bundle msgpack anyway (and Fedora's rules about bundled software are not as strict as they used to be) but still we prefer unbundled software.
(Also it makes licensing more complicated - if you bundle msgpack this means borgbackup is now available under 3-clause BSD AND Apache v2.)

ThomasWaldmann:
2 problems found:
- the flake8 test is unhappy: it did not require cython yet, but the new setup.py now depends on cython
- litte/big endian is determined when the cpp is generated and short afterwards compiled
- different handling of cpp based build than for c

the way borg's setup.py deals with C extensions and cython is that it supports building the C files from cython at release time and the generated C files get bundled into the pypi package - so users installing the pypi package do not need cython.

it seems that this is not yet supported for msgpack's CPP sources yet.

ThomasWaldmann:
@FelixSchwarz Guess I could solve all the remaining issues. Can you review?

FelixSchwarz:
I'll do that tomorrow (evening). Thank you.

ThomasWaldmann:
Merged! \o/ And will be in 1.1.10 release soon.

FelixSchwarz:
Thank you. How soon is "soon"? A few days/couple of weeks? Today?

ThomasWaldmann:
I hope to get it out before Saturday (will be busy otherwise at Sat/Sun).

On May 15, 2019 9:22:27 PM GMT+02:00, Felix Schwarz <notifications@github.com> wrote:
>Thank you. How soon is "soon"? A few days/couple of weeks? Today?
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/borgbackup/borg/pull/4560#issuecomment-492789640

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #4570, opened 2019-05-15
rename test_mount_hardlinks to test_fuse_mount_hardlinks (master)

ThomasWaldmann:
so both can be excluded by "not test_fuse"

(fedora does this for testing)


FelixSchwarz:
thanks.

PR #4575, opened 2019-05-16
AUTHORS: mention copyright+license for bundled msgpack

FelixSchwarz:
I think msgpack should be present here as well.


ThomasWaldmann:
Good catch!

But the PR needs to be against 1.1-maint, msgpack is not bundled (yet?) into master branch code.

FelixSchwarz:
sorry, I had also pull requests configured in my git config and so "git log" tricked me into thinking that this was also done for the master branch.

PR #4600, opened 2019-06-02
Add example of exclude-norecurse rule in help patterns

d7415:
I haven't generated the new docs - I wasn't sure from the contributions section whether they should be generated here or wait for a release, so I kept it minimal.

There may be a better example to use, but this is one that came up in #borgbackup today.


ThomasWaldmann:
looks good, can you do a PR for 1.1-maint branch also?

d7415:
> looks good, can you do a PR for 1.1-maint branch also?

Sure

Issue #4592, opened 2019-05-25
ability to unbundle xxhash?


FelixSchwarz:
I noticed that borgbackup bundles (parts of) the xxHash library since commit 47894072ad5ed6d28375bdb7d6316b6bc8661284 (borgbackup 1.1.0). It seems as if you copied xxHash 0.6.2 and stripped all non 64-bit functionality.

Do you think it is possible to use a system-provided xxHash library? Is there a technical reason why the xxHash library is not suitable for borg?
I noticed a [comment in the original pull request](https://github.com/borgbackup/borg/pull/2580#issuecomment-305579398) by @enkore. He mentions performance benefits of not having additional function calls. Is there a place which is most likely affected by this?

This is a request mandated by [Fedora's packaging policy](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling):
> In addition to indicating bundling in this manner, packages whose upstreams have no mechanism to build against system libraries must be contacted publicly about a path to supporting system libraries. If upstream refuses, this must be recorded in the spec file, (...)

ThomasWaldmann:
I didn't work with that code yet, just reviewed it back then.

I did a quick check about libxxhash: seems to be rather new in dists, for ubuntu one needs >= cosmic, for debian >= buster.

So, for borg 1.1.x I'ld prefer things staying as is for now, for master branch we could do it like we do for the other libs (see setup_xxx.py code).

FelixSchwarz:
> So, for borg 1.1.x I'ld prefer things staying as is for now, for master branch we could do it like we do for the other libs (see setup_xxx.py code).

That is sensible. I'll see if I can take a stab at this.

enkore:
IIRC it was new back then and no one of relevance had it packaged.

FelixSchwarz:
Just a quick update: I have some code which unbundles xxhash which builds. There is some more work to do because just putting borg 1.2.0a6 in Fedora's spec file (+ unbundling patch) caused test failures. However that might be also a side-effect of the 1.2.0a6 upgrade (I never even tried to build the unmodified borg alpha version).

If someone is interested in working on that I'm happy to push my work in progress to a branch.

Btw: I noticed borgbackup right now ships *two* copies of xxhash already :-)

ThomasWaldmann:
You could just open a "WIP: ..." PR.

About the 2 copies: I guess one is the directly bundled xxhash code, the other is indirect via some other code we bundle (like zstd).

[...]

PR #4602, opened 2019-06-03
ability to unbundle xxhash

FelixSchwarz:
This should fix issue #4592 . I tested this (briefly) on Fedora 30 and the test suite passes for me.

### DIFF
--- a/src/borg/algorithms/checksums.pyx
+++ b/src/borg/algorithms/checksums.pyx
@@ -12,7 +12,7 @@ cdef extern from "crc32_dispatch.c":
     int _have_clmul "have_clmul"()
 
 
-cdef extern from "xxh64/xxhash.c":
+cdef extern from "xxhash-libselect.h":
     ctypedef struct XXH64_canonical_t:
         char digest[8]
 
--- a/src/borg/algorithms/checksums.pyx
+++ b/src/borg/algorithms/checksums.pyx
@@ -25,6 +25,8 @@ cdef extern from "xxh64/xxhash.c":
         XXH_OK,
         XXH_ERROR
 
+    XXH64_state_t* XXH64_createState();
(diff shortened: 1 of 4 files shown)


src/borg/algorithms/checksums.pyx
```diff
@@ -80,25 +82,29 @@ def xxh64(data, seed=0):
 
 
 cdef class StreamingXXH64:
-    cdef XXH64_state_t state
+    cdef XXH64_state_t* state
 
     def __cinit__(self, seed=0):
+        self.state = XXH64_createState();
```
ThomasWaldmann: no semicolon.

ThomasWaldmann: also: can this call go wrong somehow, do we need an error handler?

FelixSchwarz: `XXH64_createState()` is basically just a renamed `malloc()` ([source](https://github.com/Cyan4973/xxHash/blob/701423eeda62b151a437444290a2669f46324f13/xxhash.c#L454)).

I could check for `self.state is None` (assuming cython would map `NULL` to `None`) and raise an exception afterwards but I'm not sure that would work in an actual out-of-memory situation.

I can of course implement something like this if you prefer (also other suggestions welcome).

ThomasWaldmann: hmm, i had a look at other code parts, we do not check for allocation errors of tiny structs there either. and yes, guess python could then be in a rather dysfunctional state anyway.

so keep as is and just remove the semicolon (and collapse the fixup changeset).

FelixSchwarz:
ok, I force-pushed new commits which do not include the semicolon. Anything I missed?

ThomasWaldmann:
thanks!

PR #4603, opened 2019-06-03
setup.py: fix detection of missing Cython

FelixSchwarz:
borgbackup's setup.py tries to display a meaningful error message if the user needs Cython but the library is not available. This pull request fixes the check (which was broken sometime in 1.2.x as it works in in 1.1.10.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,9 +115,8 @@ else:
         def __init__(self, *args, **kwargs):
             raise Exception('Cython is required to run sdist')
 
-    if not on_rtd and not all(os.path.exists(path) for path in [
-        compress_source, crypto_ll_source, chunker_source, hashindex_source, item_source, checksums_source,
-        platform_posix_source, platform_linux_source, platform_freebsd_source, platform_darwin_source]):
+    cython_c_files = [fn.replace('.pyx', '.c') for fn in cython_sources]
+    if not on_rtd and not all(os.path.exists(path) for path in cython_c_files):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -115,9 +115,8 @@ class Sdist(sdist):
         def __init__(self, *args, **kwargs):
             raise Exception('Cython is required to run sdist')
 
-    if not on_rtd and not all(os.path.exists(path) for path in [
-        compress_source, crypto_ll_source, chunker_source, hashindex_source, item_source, checksums_source,
-        platform_posix_source, platform_linux_source, platform_freebsd_source, platform_darwin_source]):
+    cython_c_files = map(lambda fn: fn.replace('.pyx', '.c'), cython_sources)
```
ThomasWaldmann: Guess a list comprehension is easier here than map/lambda:
```
cython_c_files = [fn.replace('.pyx', '.c') for fn in cython_sources]
```

It even could be put into the following line by just adding the `.replace(...)` there.

FelixSchwarz: I'll change that in a minute though I have a slight preference for two lines (I think the code is more expressive if readers see easily that this is a about "cython files (names) in C". Also I had in mind that we might have to do this for "cpp" if msgpack needs to be bundled in 1.2.0.

ThomasWaldmann: yeah, 2 lines is more clear.

FelixSchwarz:
updated code pushed :-)

ThomasWaldmann:
thanks!

PR #4613, opened 2019-06-09
add a push-style example to borg-create(1)

majewsky:
When I browsed the readthedocs and manpages while setting up Borg for myself for the first time, I saw multiple hints that pushing backups via SSH is supported (e.g. the existence of the `borg serve` command and the `--rsh` option), but I never found a clear example. I only found <https://borgbackup.readthedocs.io/en/stable/usage/general.html#repository-urls> after being pointed to it by @ThomasWaldmann.

Therefore, an explicit example in the `borg create` documentation seems appropriate. I put it next to the pull-style example using sshfs which serves a similar usecase.

### DIFF
--- a/docs/man/borg-create.1
+++ b/docs/man/borg-create.1
@@ -251,6 +251,9 @@ $ borg create /path/to/repo::my\-files /home \e
 # use zlib compression (good, but slow) \- default is lz4 (fast, low compression ratio)
 $ borg create \-C zlib,6 \-\-one\-file\-system /path/to/repo::root\-{now:%Y\-%m\-%d} /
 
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
https://borgbackup.readthedocs.io/en/stable/quickstart.html#remote-repositories there is a little bit there.

also, if we want to add more to the borg-create man page, I'ld suggest rather using the `ssh://...` syntax because that syntax also supports non-standard ssh ports.

ThomasWaldmann:
2 changes needed here:
- see my previous comment
- do not edit/update the man page, this is automatically generated at release time

ThomasWaldmann:
Fixed it.

Issue #4211, opened 2018-12-14
OSError: [Errno 5] Input/output error in "msgpack/_unpacker.pyx", line 159


finefoot:
## Have you checked borgbackup docs, FAQ, and open Github issues?

I have read the docs, FAQ and GitHub issue search results but I haven't browsed through *every* GitHub issue. Sorry, if I should have missed something. Might be interesting in the light of https://github.com/borgbackup/borg/issues/4059

## Is this a BUG / ISSUE report or a QUESTION?

Bug

#### Your borg version (borg -V).

    $ borg -V
    borg 1.0.11

#### Operating system (distribution) and version.

Lubuntu 16.04

#### Hardware / network configuration, and filesystems used.

CIFS mounted with

    sudo mount -t cifs -o rw,user=[redacted] //192.168[redacted] /media/[redacted]/backup

#### How much data is handled by borg?

Sorry, what exactly do you want to know here? The line with the progress is partly overwritten with `Local Exception`:

    Local Exception.31 GB C 237.61 GB D 40173 N [redacted]

I'm trying to backup approximately 2 TB of data, approximately 230 GB had been written so far.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

    borg create --verbose --progress --stats /media/[redacted]/backup::Monday /media/[redacted]/archive

## Describe the problem you're observing.

I initiated the bord repo with

    borg init --encryption=repokey /media/[redacted]/backup

Then ran `borg create` as described above. After approximately 230 GB had been written, I got the error below.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

I ran the exact `borg create` line afterwards again which resulted in the same error.

#### Include any warning/errors/backtraces from the system logs

Pasted below are two `stdout`/`stderr` output logs. First one from the initial attempt, which failed after approximately 230 GB had been written and took quite some time. Second one from the subsequent attempt with the same command. This failed almost immediately.

Initial attempt:

```
$ borg create --verbose --progress --stats /media/[redacted]/backup::Monday /media/[redacted]/archive

Enter passphrase for key /media/[redacted]/backup: 
Local Exception.31 GB C 237.61 GB D 40173 N [redacted]
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 2168, in main
    exit_code = archiver.run(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 2104, in run
[...]

Issue #4086, opened 2018-09-24
Crash when creating new repository on CIFS mounted volume


fuxoft:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes. My bug is possibly related to bug #4056 but I am not sure.

## Is this a BUG / ISSUE report or a QUESTION?

Bug

#### Your borg version (borg -V).

1.1.5

#### Operating system (distribution) and version.

Xubuntu 18.04.1 LTS

#### Hardware / network configuration, and filesystems used.

CIFS mounted using the following fstab line:
//192.168.69.97/sklad /media/sklad cifs vers=1.0,guest,uid=1000 0 0

#### How much data is handled by borg?

none

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

borgbackup init -v --encryption=repokey-blake2 /media/sklad/backup/frantisek_borg/fukafutokyu_orig

## Describe the problem you're observing.

I am trying to create new repository on CIFS mounted backup drive using the command above. After entering and confirming the passphrase, the following crash occurs:

```
Local Exception
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4253, in main
    exit_code = archiver.run(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4185, in run
    return set_ec(func(args))
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 152, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 266, in do_init
    key = key_creator(repository, args)
  File "/usr/lib/python3/dist-packages/borg/crypto/key.py", line 111, in key_creator
    return key.create(repository, args)
  File "/usr/lib/python3/dist-packages/borg/crypto/key.py", line 678, in create
    key.save(target, passphrase)
  File "/usr/lib/python3/dist-packages/borg/crypto/key.py", line 788, in save
    target.save_key(key_data)
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 299, in save_key
    self.save_config(self.path, self.config)
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 281, in save_config
    os.link(config_path, old_config_path)
[...]

Issue #4059, opened 2018-08-26
check repository on smbfs / cifs


ThomasWaldmann:
the question is whether #4056 was just a broken samba server (or smbfs/cifs client) or whether we have a problem in general with that.

So, try if that works:
```
# have a smbfs/cifs filesystem mounted on /mnt/smb, then:
borg init -e repokey /mnt/smb/repo
borg create /mnt/smb/repo::archive somefiles
```

If you tested it, post the outcome and also add OS/platform and smb client and server version info.

matt-h:
I just did a test on a samba server and it worked just fine.

```
$ borg init -e repokey /mnt/smb/testing/testborg
nter new passphrase:
Enter same passphrase again:
Do you want your passphrase to be displayed for verification? [yN]: y
Your passphrase (between double-quotes): "test123"
Make sure the passphrase displayed above is exactly what you wanted.

By default repositories initialized with this version will produce security
errors if written to with an older version (up to and including Borg 1.0.8).

If you want to use these older versions, you can disable the check by running:
borg upgrade --disable-tam /mnt/smb/testing/testborg

See https://borgbackup.readthedocs.io/en/stable/changes.html#pre-1-0-9-manifest-spoofing-vulnerability for details about the security implications.
```
```
$ borg create /mnt/smb/testing/testborg::testarchive git/testfiles
Enter passphrase for key /mnt/smb/testing/testborg:
```
```
$ borg list /mnt/smb/testing/testborg
Enter passphrase for key /mnt/smb/testing/testborg:
testarchive                          Mon, 2018-08-27 11:45:02 [c1f48620712130ddcd47e7c16a2c973599f2651125d62999f216210287b08c82]
```

### Client
```
$ uname -rv
4.18.3-arch1-1-ARCH #1 SMP PREEMPT Sat Aug 18 09:22:54 UTC 2018
```
```
$ borg --version
borg 1.1.6
```
Mounted using cifs

cifs-utils version 6.8

```
$ mount | grep smb
//servername/sambashare on /mnt/smb type cifs (rw,relatime,vers=default,cache=strict,username=matt,domain=,uid=1000,forceuid,gid=100,forcegid,addr=10.10.10.2,file_mode=0755,dir_mode=0755,soft,nounix,serverino,mapposix,rsize=1048576,wsize=1048576,echo_interval=60,actimeo=1)
```

### Server
```
$ uname -rv
[...]

Issue #4056, opened 2018-08-21
backup to samba


ape364:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Question

#### Your borg version (borg -V).

borg 1.1.6

#### Operating system (distribution) and version.

`Linux MiBook 4.18.3-arch1-1-ARCH #1 SMP PREEMPT Sat Aug 18 09:22:54 UTC 2018 x86_64 GNU/Linux`

#### Hardware / network configuration, and filesystems used.

```
 ~  cat /etc/systemd/system/mnt-wdsmb.mount 
[Unit]
Description=Mount WD Share at boot
Requires=systemd-networkd.service
After=network-online.target
Wants=network-online.target

[Mount]
What=//192.168.88.1/wd_share
Where=/mnt/wdsmb
Options=credentials=/home/ape364/.smb_cred,iocharset=utf8,file_mode=0777,dir_mode=0777,defaults,uid=1000,gid=985,ip=192.168.88.1,vers=2.0,rw,user,x-systemd.automount
Type=cifs
TimeoutSec=10

[Install]
WantedBy=multi-user.target

```

#### How much data is handled by borg?

none

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
 ~  ls -la /mnt/wdsmb/ | grep arch_backup
drwxrwxrwx 2 ape364 users       4096 Aug 22 00:44 arch_backup
 ~  whoami
ape364
 ~  borg init -e repokey /mnt/wdsmb/arch_backup
Enter new passphrase: 
Enter same passphrase again: 
Do you want your passphrase to be displayed for verification? [yN]: y 
Your passphrase (between double-quotes): "qwe"
Make sure the passphrase displayed above is exactly what you wanted.
Failed to securely erase old repository config file (hardlinks not supported>). Old repokey data, if any, might persist on physical storage.

By default repositories initialized with this version will produce security
errors if written to with an older version (up to and including Borg 1.0.8).

If you want to use these older versions, you can disable the check by running:
borg upgrade --disable-tam /mnt/wdsmb/arch_backup

See https://borgbackup.readthedocs.io/en/stable/changes.html#pre-1-0-9-manifest-spoofing-vulnerability for details about the security implications.
 ~  ls -la /home/ape364/backup_test/
total 12
drwxr-xr-x  2 ape364 users 4096 Aug 22 00:47 ./
[...]

PR #4630, opened 2019-06-19
Improve FAQ section about issues with network filesystems

finefoot:
I did read the FAQ before using Borg, yet I still ran into the mentioned issue with cifs - as did many other people before me: #4056, #4059, #4086, ... So, I think it makes sense to improve the respective FAQ section as proposed in #4211. Especially since at first, the filesystem seems to work fine and the error only appears later, randomly along the backup process.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -692,7 +692,8 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
 I am having troubles with some network/FUSE/special filesystem, why?
 --------------------------------------------------------------------
 
-Borg is doing nothing special in the filesystem, it only uses very
+In general, Borg just needs a well-working filesystem: It is doing
+nothing special in the filesystem, it only uses very
 common and compatible operations (even the locking is just "mkdir").
 
 So, if you are encountering issues like slowness, corruption or malfunction
@@ -705,8 +706,18 @@ recommended that you talk to the developers / support of the network fs and
 maybe open an issue in their issue tracker. Do not file an issue in the
 Borg issue tracker.
 
-If you can reproduce the issue with the proven filesystem, please file an
-issue in the Borg issue tracker about that.
+A common issue is ``OSError: [Errno 5] Input/output error`` which might
+randomly appear during the backup process while writing to a cifs-mounted
+volume. See issues
+`#4056 <https://github.com/borgbackup/borg/issues/4056>`_,
+`#4059 <https://github.com/borgbackup/borg/issues/4059>`_,
+`#4086 <https://github.com/borgbackup/borg/issues/4086>`_ or
+`#4211 <https://github.com/borgbackup/borg/issues/4211>`_ for example,
+where you can find further information and some useful mount options which
+might resolve the issue.
+
+However, if you *can* reproduce the issue with the proven filesystem, please
+do file an issue in the Borg issue tracker about that.
 
 
 Why does running 'borg check --repair' warn about data loss?


finefoot:
@ThomasWaldmann 

> I don't think linking to a whole bunch of issues is the best way to solve this. As I already noted, someone having the problem rather should find out which mount option(s) make a difference.

1. I'm sorry if you feel like I ignored your comment. On the contrary, I do agree that there shouldn't be an explicit recommendation for cifs mount options. Additionally, however, I don't think there's ever going to be one, true, best set of cifs mount options. Every user has to find suitable mount options for their use case.

2. So why the addition to the FAQ at all? Because, obviously, I did have an issue with the filesystem, yet I didn't realize that at the moment. Everything appeared to be working, at first. Also, I've never had any problems with the filesystem with other apps. Only hours into the backup process, I got the error. This lead me to - mistakenly - believe the error lies with Borg. So, despite reading the FAQ and especially the section about faulty filesystems, I did unnecessarily file an issue. And I'm not the only one as you can see from the listing of linked issues.

That's why I think the section should be improved. And those improvements don't concern explicit mount options for cifs. Did I misunderstand you and you did already address this?

ThomasWaldmann:
> I don't think there's ever going to be one, true, best set of cifs mount options.
> Every user has to find suitable mount options for their use case.

Yes and no. :) Of course things can change over time, but our docs can be updated also.

So what we need now is to specifically find out what mount option makes this specific problem go away and document it (preferably together with some info about smbd / client kernel / cifs mount version).

This is quite easy to find if done systematically, but sadly no quick thing if it only happens in the middle of a long backup.

We can also add some generic docs like "if you see an I/O error that is not borg's fault, but likely your filesystem or storage device misbehaving".

finefoot:
@ThomasWaldmann 

> Yes and no. :) Of course things can change over time, but our docs can be updated also.
> 
> So what we need now is to specifically find out what mount option makes this specific problem go away and document it (preferably together with some info about smbd / client kernel / cifs mount version).

I was trying to say that the working mount option will likely depend on so many different factors that it's not possible to extend the docs with that "one, true, best set of cifs mount options" I was talking about. (Factors like OS, versions for both client as well as server - in my case a rather old NAS where I don't even know if I can find out which cifs version it's running...)

> We can also add some generic docs like "if you see an I/O error that is not borg's fault, but likely your filesystem or storage device misbehaving".

Yes! :) That's exactly what I was trying to do with this PR. I tried to keep the added sentences very generic towards the I/O error, but apparently not generic enough for your taste. ;)

finefoot:
@ThomasWaldmann Waiting for your response. Not trying to bother you, so if you just didn't get a chance to reply, that's fine of course. :) Your previous review requested changes but at this point I'm not sure if that's still true? Do you agree that it's impossible to provide a perfect set of mount options which will work for everyone? (That's why I just linked some of the issues, instead.) So I'm not sure if "changes requested" still stands with regard to the decision to not include mount options. (But to add a general hint and links about known issues - not with borg, but with using remote/network file systems in combination with borg.)

ThomasWaldmann:
Well, I still think this is not in the best state. If you are not willing to invest the work to find out which of the bunch of mount options you added really made a difference **for you**, that is fine - but without some precise information there is no point in adding lots of links to the docs.

As I already pointed out, we can also add some generic docs like "if you see an I/O error that is not borg's fault, but likely your filesystem or storage device misbehaving".

finefoot:
> but without some precise information there is no point in adding lots of links to the docs

Why do you say that? I honestly don't understand and respectfully have to disagree: The links don't clutter the paragraph up and serve as citations for the previously made claim that often-times issues with the backup process turn out to not be issues with borg but with the (network) filesystem. Why would you oppose those citations? The links are not in any way presented as links to a general solution as there can never be a general solution. (Different setups need different mount options.) But reading those issues will help the majority of users stumbling upon that section about network issues in the docs and prevent further duplicate issues on the topic. (That's the goal, right?!)

> we can also add some generic docs like "if you see an I/O error that is not borg's fault, but likely your filesystem or storage device misbehaving"

I'm sorry, I genuinely don't get where this statement is different from the current version. I'm afraid I won't be able to make any improvements without further clarification. Feel free to make the changes or close this PR altogether. No hard feelings, though. :)

ThomasWaldmann:
I don't think this is going anywhere and I also think I have made my point clear enough already.

PR #4635, opened 2019-06-22
first ctrl-c: checkpoint and abort, fixes #4606

### LINKED ISSUE #4606: use ctrl-c to trigger checkpoint&abort

Yes

Issue, really a feature request.

1.1.9

Ubuntu 16.04, Ubuntu 18.04, CentOS 7, OS X 10.13.6.

Various.

Varies, kBytes to GBytes.

    borg create --progress ...

Borg makes checkpoints at intervals. Suppose we want to interrupt a backup (maybe it's taking too long and we'll try again later). If waiting 5 minutes longer would allow a checkpoint to be created, it's advisable to wait. But Borg doesn't tell us when a checkpoint is coming or when one has been done. So we have no idea when is the right time to hit ^C.

A good solution would be for Borg to print when a checkpoint has been done if the --progress option has been given; and, if possible, also print how much longer to the next checkpoint.

In a more fanciful scenario, I can imagine a progress bar with marks showing where checkpoints will occur or have occurred. Like how some video services show you where the ads are. (May be hard to implement as Borg probably does not know how much data will be transmitted in the future.)

Currently, --progress causes a line of output to be shown at the bottom of the screen. The documentation doesn't mention that any part of this output tells us when a checkpoint occurs. 

Admittedly, I have not downloaded and tested version 1.1.10 to see if it prints anything different.

A search for the word "checkpoint" in issues, docs, and faq didn't turn up anything relevant.

Not applicable.

Not applicable.

ThomasWaldmann:

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -40,6 +40,7 @@ from .helpers import safe_ns
 from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
 from .helpers import os_open, flags_normal
 from .helpers import msgpack
+from .helpers import sig_int
 from .patterns import PathPrefixPattern, FnmatchPattern, IECommand
 from .item import Item, ArchiveItem, ItemDiff
 from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
As there was no negative feedback at least, I am merging this now so it gets into next alpha release.

If there are problems, we can still revert it later.

PR #4636, opened 2019-06-23
Enable placeholder usage in all extra archive arguments

elho:
This fixes the issue, that with `borg diff REPO::ARCHIVE1 ARCHIVE2`, placeholders only work in `ARCHIVE1`, not in `ARCHIVE2`.

The `REPO::ARCHIVE1` argument uses `location_validator` as its type, which does replace placeholders, the `ARCHIVE2` argument uses  `archivename_validator`, which prior to this change did not.

This other two commands using arguments of type `archivename_validator`, namely `rename` and `recreate`, did do replacing of placeholders in the extra archive arguments on their own. With the change to `archivename_validator()` this duplicate code could be removed.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -498,6 +498,7 @@ def location_validator(archive=None, proto=None):
 
 def archivename_validator():
     def validator(text):
+        text = replace_placeholders(text)
         if '/' in text or '::' in text or not text:
             raise argparse.ArgumentTypeError('Invalid repository name: "%s"' % text)
         return text
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1071,8 +1071,7 @@ class Archiver:
     @with_archive
     def do_rename(self, args, repository, manifest, key, cache, archive):
         """Rename an existing archive"""
-        name = replace_placeholders(args.name)
-        archive.rename(name)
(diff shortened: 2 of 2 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -498,6 +498,7 @@ def validator(text):
 
 def archivename_validator():
     def validator(text):
+        text = replace_placeholders(text)
         if '/' in text or '::' in text or not text:
             raise argparse.ArgumentTypeError('Invalid repository name: "%s"' % text)
```
ThomasWaldmann: unrelated to your change, but just noticed: the error message seems to be wrong, should say "Invalid archive name".

elho:
> Did you check all places where archivename_validator is called, whether they are ok with this change?

Yes, its only used 3 times in the entire source, as type for the respective arguments of the commands `diff`, `recreate` and `rename`. I didn't run the test suite locally, but did successfully test all 3 commands with an archive name containing `{hostname}` placeholder.

ThomasWaldmann:
Fine. :) Do you want to do a PR with this against 1.1-maint branch also?

elho:
> Fine. :) Do you want to do a PR with this against 1.1-maint branch also?

Yes, done in PR #4637.

PR #4637, opened 2019-06-23
Enable placeholder usage in all extra archive arguments

elho:
PR #4636 rebased on 1.1-maint and adapted for non-split `helpers.py`.


ThomasWaldmann:
Thanks!

PR #4638, opened 2019-06-23
support xxh64 checksum in addition to the hashlib hashes in borg list

elho:
Sopport xxh64 checksum in `borg list` besides the existing hashes provided by hashlib, using the built-in xxh64 implementation.

This e.g. allows to generate an archive listing that could be used with `xxh64sum` to quickly compare (verify) actual archive contents with existing files. While there may not be a noticeable speedup in borg's operation on a current system, it does lower CPU usage and may help  systems with slow CPUs.
It also gives people that happen to already have `xxh64sum` generated checksum files for their data another option.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -648,8 +648,6 @@ class ArchiveFormatter(BaseFormatter):
 
 
 class ItemFormatter(BaseFormatter):
-    from ..algorithms.checksums import StreamingXXH64
-    xxh64 = StreamingXXH64
     hash_algorithms = hashlib.algorithms_guaranteed.union({'xxh64'})
     KEY_DESCRIPTIONS = {
         'bpath': 'verbatim POSIX path, can contain any character except NUL',
@@ -714,6 +712,8 @@ class ItemFormatter(BaseFormatter):
         return any(key in cls.KEYS_REQUIRING_CACHE for key in format_keys)
 
     def __init__(self, archive, format, *, json_lines=False):
+        from ..algorithms.checksums import StreamingXXH64
+        self.xxh64 = StreamingXXH64
         self.archive = archive
         self.json_lines = json_lines
         static_keys = {
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -741,7 +741,7 @@ class ItemFormatter(BaseFormatter):
             'ctime': partial(self.format_time, 'ctime'),
             'atime': partial(self.format_time, 'atime'),
         }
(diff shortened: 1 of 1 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -648,6 +648,9 @@ def format_time(self, ts):
 
 
 class ItemFormatter(BaseFormatter):
+    from .algorithms.checksums import StreamingXXH64
```
ThomasWaldmann: hmm, this worked, but putting the same line in global module scope did not?

src/borg/helpers/parseformat.py
```diff
@@ -738,7 +741,7 @@ def __init__(self, archive, format, *, json_lines=False):
             'ctime': partial(self.format_time, 'ctime'),
             'atime': partial(self.format_time, 'atime'),
         }
-        for hash_function in hashlib.algorithms_guaranteed:
+        for hash_function in self.__class__.hash_algorithms:
```
ThomasWaldmann: self.hash_algorithms didn't work?

src/borg/helpers/parseformat.py
```diff
@@ -812,7 +815,10 @@ def calculate_csize(self, item):
     def hash_item(self, hash_function, item):
         if 'chunks' not in item:
             return ""
-        hash = hashlib.new(hash_function)
+        if hash_function in hashlib.algorithms_guaranteed:
+            hash = hashlib.new(hash_function)
+        elif hash_function == 'xxh64':
+            hash = self.__class__.xxh64()
```
ThomasWaldmann: self.xxh64() ?

ThomasWaldmann:
Guess there should be also a mention in the docs and it should point out, that xxh64 is a **non-cryptographic** hash function.

src/borg/helpers/parseformat.py
elho: Yes, this works and neither that same import nor any other variant I came up with did, including a plain `import` without `from` - guess it would have to be that in both modules to make cyclic import work, or maybe Cython is a special case here...

src/borg/helpers/parseformat.py
elho: That would work as well, going to update the PR.

src/borg/helpers/parseformat.py
elho: Correction, this does work on 1.1, , not with 1.2 though :thinking:

ThomasWaldmann:
thanks! is it enough if this will be in borg 1.2 later or do you want to update 1.1-maint also?

PR #4640, opened 2019-06-24
Fix osxfuse github link in installation docs

russelldavis:
It was broken because there's already a different reference named `github` in `docs/global.rst.inc`. Fixed by using an anonymous hyperlink reference, per https://stackoverflow.com/a/14067756/278488


ThomasWaldmann:
thanks! could you update 1.1-maint branch also with the same change?

russelldavis:
sure: https://github.com/borgbackup/borg/pull/4642

PR #4643, opened 2019-06-25
Clarify docs on patterns

russelldavis:
The docs for the `pp:` selector were misleading to me -- they seemed to imply that matching whole sub-directories isn't already straightforward with the default styles (`fm:` and `sh:`). It also wasn't clear to me why `pp:` needed to exist at all given the similar behavior of `pf:`, until I thought about the behavior without exclude-norecurse (which isn't introduced until later in the document).

Thinking about this also gave me an idea for an optimization, which I added a comment for in the code.

Happy to update things if I misunderstood anything here.

### DIFF
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -58,8 +58,16 @@ Regular expressions, selector `re:`
     the re module <https://docs.python.org/3/library/re.html>`_.
 
 Path prefix, selector `pp:`
-    This pattern style is useful to match whole sub-directories. The pattern
-    `pp:root/somedir` matches `root/somedir` and everything therein.
+    This pattern style is useful to match whole sub-directories without worrying
+    about escaping special characters. Otherwise, you can just use the default
+    `fm:` or `sh:` styles, which also match whole sub-directories. Unlike those
+    styles, a trailing slash here has no effect on what gets matched. The pattern
+    `pp:root/somedir` (and `pp:root/somedir/`) matches `root/somedir` and
+    everything therein.
+
+    When using exclude-norecurse rules, this style is equivalent to `pf:`,
+    except that `pf:` won't match patterns with trailing slashes (and `pp:`
+    won't be as efficient, though that may be optimized in the future).
 
 Path full-match, selector `pf:`
     This pattern style is (only) useful to match full paths.
@@ -138,7 +146,8 @@ Examples::
     path. The first matching pattern is used so if an include pattern matches before
     an exclude pattern, the file is backed up. If an exclude-norecurse pattern matches
     a directory, it won't recurse into it and won't discover any potential matches for
-    include rules below that directory.
+    include rules below that directory. Note that this is the behavior that gets applied
+    to all patterns from `--exclude` and `--exclude-from`.
 
     Note that the default pattern style for ``--pattern`` and ``--patterns-from`` is
     shell style (`sh:`), so those patterns behave similar to rsync include/exclude
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -58,16 +58,8 @@ Regular expressions, selector `re:`
     the re module <https://docs.python.org/3/library/re.html>`_.
 
 Path prefix, selector `pp:`
-    This pattern style is useful to match whole sub-directories without worrying
-    about escaping special characters. Otherwise, you can just use the default
-    `fm:` or `sh:` styles, which also match whole sub-directories. Unlike those
-    styles, a trailing slash here has no effect on what gets matched. The pattern
-    `pp:root/somedir` (and `pp:root/somedir/`) matches `root/somedir` and
-    everything therein.
-
-    When using exclude-norecurse rules, this style is equivalent to `pf:`,
-    except that `pf:` won't match patterns with trailing slashes (and `pp:`
-    won't be as efficient, though that may be optimized in the future).
+    This pattern style is useful to match whole sub-directories. The pattern
+    `pp:root/somedir` matches `root/somedir` and everything therein.
 
 Path full-match, selector `pf:`
     This pattern style is (only) useful to match full paths.
@@ -146,8 +138,7 @@ Examples::
     path. The first matching pattern is used so if an include pattern matches before
     an exclude pattern, the file is backed up. If an exclude-norecurse pattern matches
     a directory, it won't recurse into it and won't discover any potential matches for
-    include rules below that directory. Note that this is the behavior that gets applied
-    to all patterns from `--exclude` and `--exclude-from`.
+    include rules below that directory.
 
     Note that the default pattern style for ``--pattern`` and ``--patterns-from`` is
     shell style (`sh:`), so those patterns behave similar to rsync include/exclude
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2067,14 +2067,21 @@ class Archiver:
             the re module <https://docs.python.org/3/library/re.html>`_.
 
         Path prefix, selector `pp:`
-            This pattern style is useful to match whole sub-directories. The pattern
-            `pp:root/somedir` matches `root/somedir` and everything therein.
+            This pattern style is useful to match whole sub-directories without worrying
+            about escaping special characters. Otherwise, you can just use the default
+            `fm:` or `sh:` styles, which also match whole sub-directories. Unlike those
+            styles, a trailing slash here makes no difference. The pattern `pp:root/somedir`
+            (and `pp:root/somedir/`) matches `root/somedir` and everything therein.
+
+            When used with exclude-norecurse rules, this style is equivalent to `pf:`
+            (except not as efficient).
 
         Path full-match, selector `pf:`
             This pattern style is (only) useful to match full paths.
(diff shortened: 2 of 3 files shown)


src/borg/patterns.py
```diff
@@ -71,6 +71,7 @@ class PatternMatcher:
     *fallback* is a boolean value that *match()* returns if no matching patterns are found.
 
     """
+
```
ThomasWaldmann: accidental change?

src/borg/patterns.py
```diff
@@ -340,6 +341,10 @@ def parse_pattern(pattern, fallback=FnmatchPattern, recurse_dir=True):
     """Read pattern from string and return an instance of the appropriate implementation class.
 
     """
+    # Potential optimization for fm, sh, and pf patterns: if recurse_dir is False and the pattern
+    # doesn't do any special matching, it could be converted to a PathFullPattern (we would also
+    # need to remove any trailing slashes from pf patterns, and exclude fm and sh patterns with
+    # trailing slashes from the optimization).
```
ThomasWaldmann: why should we do any "optimization" or transformation?

we can rather keep it simple and just do what the user tells borg to do.

docs/usage/help.rst.inc
```diff
@@ -58,8 +58,16 @@ Regular expressions, selector `re:`
     the re module <https://docs.python.org/3/library/re.html>`_.
 
 Path prefix, selector `pp:`
-    This pattern style is useful to match whole sub-directories. The pattern
-    `pp:root/somedir` matches `root/somedir` and everything therein.
+    This pattern style is useful to match whole sub-directories without worrying
+    about escaping special characters. Otherwise, you can just use the default
```
ThomasWaldmann: I'ld rather remove the "not worrying" here and add a "be careful to escape special chars so borg gets to see them and they are not expanded by your shell" to the patterns that frequently have a `*` or so in them.

docs/usage/help.rst.inc
```diff
@@ -58,8 +58,16 @@ Regular expressions, selector `re:`
...
+    `fm:` or `sh:` styles, which also match whole sub-directories. Unlike those
+    styles, a trailing slash here has no effect on what gets matched. The pattern
+    `pp:root/somedir` (and `pp:root/somedir/`) matches `root/somedir` and
+    everything therein.
```
ThomasWaldmann: `A trailing slash here has no effect.`

docs/usage/help.rst.inc
```diff
@@ -58,8 +58,16 @@ Regular expressions, selector `re:`
...
+
+    When using exclude-norecurse rules, this style is equivalent to `pf:`,
+    except that `pf:` won't match patterns with trailing slashes (and `pp:`
```
ThomasWaldmann: similar here, please describe the specialities of `pf` in the pf description, see below.

docs/usage/help.rst.inc
ThomasWaldmann: (please check if we already have a warning about shell wildcard expansion and quoting somewhere)

russelldavis: How about, "This pattern style is useful for matching whole sub-directories without having to escape special characters (beyond any escaping needed when using it within a shell)".

> (please check if we already have a warning about shell wildcard expansion and quoting somewhere)

Yeah, there's a line later on that says "When used from within a shell the patterns should be quoted to protect them from expansion."

docs/usage/help.rst.inc
russelldavis: It's comparing `pf` and `pp`, so it seems like either way, one of the sections will be referring to the other.

src/borg/patterns.py
russelldavis: yeah, will fix

src/borg/patterns.py
russelldavis: The docs talk about how `pf` patterns perform better. It's implied that if you have a lot of patterns, it'd be better to use this style. It would be nice if users didn't have to worry about that, knowing the patterns they write will automatically optimized where possible.

Not a big deal though. Happy to add the context above to the comment, move it elsewhere, it remove it entirely.

ThomasWaldmann: guess it is enough if the docs are clear about performance.

pf > pp > all else (speed wise).

should only matter if you have lots of patterns, so likely no issue for most users.

docs/usage/help.rst.inc
ThomasWaldmann: ok, that should be enough. no change needed.

it should be clear that if you don't use wildcards, that you don't need to quote any, of course.

docs/usage/help.rst.inc
ThomasWaldmann: that pf: needs to be used without trailing slashes (when specifying dirs) should be documented in the pf section.

ThomasWaldmann: (and we could change the code to strip any trailing slashes for pf:)

russelldavis:
Ok, I updated the PR. A few additional notes:
- I had originally edited the autogenerated `help.rst.inc` file instead of the docstring it is based on. The changes are now in the docstring, and I reverted that file. It will get updated on the next run of `build_usages` (I kept that out of this PR since there are a bunch of other unrelated changes it would pull in.)
- My update about "`pf:` won't match patterns with trailing slashes" was actually incorrect. It was based on a bad assumption from seeing `rstrip(os.path.sep)` in the other pattern classes and not in PathFullPattern. I updated the docs to reflect this.

src/borg/archiver.py
```diff
@@ -2147,7 +2154,8 @@ def do_break_lock(self, args, repository):
             path. The first matching pattern is used so if an include pattern matches before
             an exclude pattern, the file is backed up. If an exclude-norecurse pattern matches
             a directory, it won't recurse into it and won't discover any potential matches for
-            include rules below that directory.
+            include rules below that directory. Note that this is the behavior that gets applied
```
russelldavis: I thought about whether this should be mentioned higher up in the section about `--exclude`, but I think it makes more sense here, since the difference in behavior is only relevant if you are using the experimental `--pattern` selectors in this section.

russelldavis:
@ThomasWaldmann anything I missed here, or any other changes you'd like?

src/borg/archiver.py
```diff
@@ -2067,14 +2067,21 @@ def do_break_lock(self, args, repository):
...
-            This pattern style is useful to match whole sub-directories. The pattern
-            `pp:root/somedir` matches `root/somedir` and everything therein.
+            This pattern style is useful to match whole sub-directories without worrying
+            about escaping special characters. Otherwise, you can just use the default
+            `fm:` or `sh:` styles, which also match whole sub-directories. Unlike those
+            styles, a trailing slash here makes no difference. The pattern `pp:root/somedir`
+            (and `pp:root/somedir/`) matches `root/somedir` and everything therein.
+
+            When used with exclude-norecurse rules, this style is equivalent to `pf:`
+            (except not as efficient).
```
ThomasWaldmann: For the reasons I already told (scope, not talking about not relevant stuff), I think this just should be:
```
This pattern style is useful to match whole sub-directories
(a trailing slash makes no difference).
```

This is telling all that is in the scope of `pp:`.

I thought about transforming your last sentence to a hint to rather use pf: for exclude-norecurse rules for performance reasons. But then, considering that one usually does not have very many of them, I thought it does not really matter.

The performance point of pf comes when having very many of them (like feeding a full list of all files to borg).

src/borg/archiver.py
```diff
@@ -2067,14 +2067,21 @@ def do_break_lock(self, args, repository):
...
 
         Path full-match, selector `pf:`
             This pattern style is (only) useful to match full paths.
             This is kind of a pseudo pattern as it can not have any variable or
-            unspecified parts - the full path must be given.
-            `pf:root/file.ext` matches `root/file.txt` only.
+            unspecified parts - the full path must be given. Trailing slashes make no
+            difference. `pf:root/file.ext` (and `pp:root/file.ext/`) matches `root/file.txt`
+            only.
```
ThomasWaldmann: I guess it does not need the last sentence with an example for what a trailing slash is, especially not a weird one that rather looks like a quirk.

src/borg/archiver.py
ThomasWaldmann: A hint about the potential need to escape or quote when using "special characters" (that would otherwise be interpreted/expanded by the shell) should be added to a general section (not to a specific pattern style), if we do not already have that (please check).

russelldavis:
> For the reasons I already told (scope, not talking about not relevant stuff), I think this just should be:
> ```
> This pattern style is useful to match whole sub-directories
> (a trailing slash makes no difference).
> ```
> This is telling all that is in the scope of pp:.

The sentences I added comparing `pp` to the other styles are specifically to clarify the confusion explained in the PR description. So, I do think the extra explanation is relevant and helpful. Maybe they could go in a new section?

Ultimately, you know best how you want this structured and worded, so rather than going back and forth, I'm going to close this PR and leave it to you to make any changes you see fit. Of course feel free to use or not use any of the changes here. Thanks for the discussion.

ThomasWaldmann:
OK, I'll have a look...

PR #4645, opened 2019-06-25
Fix broken test that relied on improper zlib assumptions

thiagomacieira:
Fix broken test that relied on improper zlib assumptions

This test assumed that zlib would always produce the same output for a
given input, and that assumption has proven incorrect. It rested on
zlib never improving. The zlib shipped in Clear Linux has algorithm
improvements, which caused this test to break.

This patch instead decompresses the expected value and compares that.

### DIFF
--- a/src/borg/testsuite/hashindex.py
+++ b/src/borg/testsuite/hashindex.py
@@ -321,7 +321,8 @@ class HashIndexDataTestCase(BaseTestCase):
         idx1[H(2)] = 2**31 - 1, 0, 0
         idx1[H(3)] = 4294962296, 0, 0  # 4294962296 is -5000 interpreted as an uint32_t
 
-        assert self._serialize_hashindex(idx1) == self.HASHINDEX
+        serialized = self._serialize_hashindex(idx1)
+        assert self._unpack(serialized) == self._unpack(self.HASHINDEX)
 
     def test_read_known_good(self):
         idx1 = self._deserialize_hashindex(self.HASHINDEX)


src/borg/testsuite/hashindex.py
```diff
@@ -300,7 +300,7 @@ def _serialize_hashindex(self, idx):
             file = os.path.join(tempdir, 'idx')
             idx.write(file)
             with open(file, 'rb') as f:
-                return self._pack(f.read())
+                return f.read()
```
ThomasWaldmann: no, this needs to stay complementary to the method below, so do not change it.

src/borg/testsuite/hashindex.py
```diff
@@ -321,7 +321,8 @@ def test_identical_creation(self):
         idx1[H(2)] = 2**31 - 1, 0, 0
         idx1[H(3)] = 4294962296, 0, 0  # 4294962296 is -5000 interpreted as an uint32_t
 
-        assert self._serialize_hashindex(idx1) == self.HASHINDEX
+        x = self._serialize_hashindex(idx1)
```
ThomasWaldmann: just wrap the rhs in self._unpack().

thiagomacieira:
GitHub says there's 1 change requested, but all request are closed. Sorry, I don't know how to mark that the changes have been implemented.

ThomasWaldmann:
OK, merged!

Can you also do the same PR against 1.1-maint branch?

thiagomacieira:
Deal. Done in #4650

PR #4646, opened 2019-06-25
SecurityManager.known(): check all files, fixes #4614

### LINKED ISSUE #4614: Repo security_dir without location file
I found that below `~/.config/borg/security/`, there is one directory that does not have a `location` file. It does contain these files: `key-type` `manifest-timestamp` `nonce` `tam_required`

I have no idea to which repository this may have belonged to, I was looking at all the `location` files wanting to clean up after doing countless tests with borg the last few weeks, creating and using many temporary repositories for that, most of which have been deleted since.

This is on Debian Buster, using its borgbackup 1.1.9-2 package.

ThomasWaldmann:
Before this fix, because known() only checked the key-type file, the
location file could go missing without being automatically regenerated.

Now a save() is triggered if not all files are present.

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -73,7 +73,8 @@ class SecurityManager:
             shutil.rmtree(path)
 
     def known(self):
-        return os.path.exists(self.key_type_file)
+        return all(os.path.exists(f)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@enkore please review. ^^^

PR #4650, opened 2019-06-28
Fiz zlib bad assumption

thiagomacieira:
This is the cherry-pick of the fix in #4645 for 1.1-maint.

>     This test assumed that zlib would always produce the same output for a
>     given input, and that assumption has proven incorrect. It rested on
>     zlib never improving. The zlib shipped in Clear Linux has algorithm
>     improvements, which caused this test to break.
>     
>     This patch instead decompresses the expected value and compares that.


ThomasWaldmann:
Thanks!

PR #4656, opened 2019-07-01
Fix invalid archive error message

emichael:
I found this while trying to use the `--target` option on `borg recreate`.


ThomasWaldmann:
thanks! can you check/fix 1.1-maint also?

emichael:
Done.

PR #4679, opened 2019-07-09
RST syntax fix in docs

nomeata:


ThomasWaldmann:
thanks! can you check whether 1.1-maint branch needs the same fix and if so, make a PR against it?

Issue #4516, opened 2019-04-21
Compression can yield bigger size - should rather use none


FabioPedretti:
This was already discussed in the past (#385 and elsewhere), but was not implemented.

When a file is already compressed, compressing it again may increase its size. For example using the default borg compressor, and standard (already compressed on my system) debian manpages:

```
mkdir /tmp/test
cd /usr/share/man/man1
cp *gz /tmp/test
cd /tmp/test
lz4 -v -m * > output.txt 2>&1
```

Have a look at the output.txt file:

```
*** LZ4 command line interface 64-bits v1.8.3, by Yann Collet ***
Compressed 1721 bytes into 1740 bytes ==> 101.10%                              
Compressed 563 bytes into 582 bytes ==> 103.37%                                
Compressed 592 bytes into 611 bytes ==> 103.21%                                
Compressed 5411 bytes into 5430 bytes ==> 100.35%                              
Compressed 2867 bytes into 2886 bytes ==> 100.66%                              
Compressed 2363 bytes into 2382 bytes ==> 100.80%                              
Compressed 2678 bytes into 2697 bytes ==> 100.71%                              
Compressed 3847 bytes into 3866 bytes ==> 100.49%                              
Compressed 1208 bytes into 1227 bytes ==> 101.57%                              
Compressed 577 bytes into 596 bytes ==> 103.29%
Compressed 1033 bytes into 1052 bytes ==> 101.84%
Compressed 7141 bytes into 7160 bytes ==> 100.27%
Compressed 154 bytes into 173 bytes ==> 112.34%
...
```
Most files are bigger then their original (actually by few bytes, but increasing to some KB whith bigger files). Actually they are all bigger than the original:
```
grep "1..\...%" output.txt|wc
grep -v "1..\...%" output.txt | grep % | wc
```
The last 2 commands give 2038 files with >= 100.02% size and 0 smaller than input files.

So it would be nice to have borg reverting to none compression when the compressed version is not smaller than the original. I know there is already auto compression, but while auto is mostly intended to avoid compressing with a slower algo this would still have 2 advantages even when not using auto:
- smaller repo (usually by a tiny fraction)
- faster file access (extract or mount) - this one is the most meaningful advantage!

ThomasWaldmann:
Fixed by #4683.

Issue #385, opened 2015-11-08
check if compressed data is larger than original and deal with it


ThomasWaldmann:
Unclear:
- do compression algorithms already internally deal with that?
- lz4?
- zlib?
- lzma?

If not, we could easily add a check to borg code and store chunks uncompressed if compression was found to be ineffective.

See also #383.

aklaver:
From the IRC conversation earlier and so the information persists:

lz4

On the already compressed front, for lz4-https://github.com/Cyan4973/lz4/blob/master/lz4_Frame_format.md . See Data Blocks.
The legacy frame did not do this, according to the next section, but that has been deprecated.
 A test on my old machine.
time lz4 postgresql-9.4.4.tar.bz2 
Compressed filename will be : postgresql-9.4.4.tar.bz2.lz4 
Compressed 17616272 bytes into 17388193 bytes ==> 98.71% 
real 0m0.052s
user 0m0.021s
sys 0m0.030s
l postgresql-9.4.4.tar.bz2*
-rw-r--r-- 1 aklaver users 17616272 Nov 17 15:33 postgresql-9.4.4.tar.bz2
-rw-r--r-- 1 aklaver users 17388193 Nov 17 15:34 postgresql-9.4.4.tar.bz2.lz4

lzma

Hmm, for lzma I have not found anything conclusive yet, though the numbers seem to indicate it still tries to compress.
time lzma postgresql-9.4.4.tar.bz2 
real 0m6.894s
user 0m6.777s
sys 0m0.109s
l postgresql-9.4.4.tar.bz2*
-rw-r--r-- 1 aklaver users 17388193 Nov 17 15:34 postgresql-9.4.4.tar.bz2.lz4
-rw-r--r-- 1 aklaver users 17384834 Nov 17 15:33 postgresql-9.4.4.tar.bz2.lzma

aklaver:
zlib

http://zlib.net/zlib_tech.html
[...]

PR #4683, opened 2019-07-14
only store compressed data if the result actually is smaller

elho:
- add decide() function to LZ4, LZMA and ZSTD classes that
  falls back to NONE if the compressed data turns out bigger than
  the original data
- leave ZLIB as is, as it does not use an ID header and compatibility
  being the only reason to use it instead of above algorithms anyway
- adapt testsuite to cope with incompressible test data not being
  compressed
- add tests to verify that incompressible data is not compressed to
  larger size compressed data

Watching the progress output in one of my initial backups using borg I did notice compressed showing a larger number than original at some point. Unlike done in issues #4516 and #385, I actually tested with borg to verify that indeed compressed chunks being stored that were larger than an uncompressed chunk would have been.
Based on this I addressed that problem in this PR.

In addition to the test suite, I also ran a test on real world data with this change applied to Debian's borg 1.1.9 package from the same snapshot to two newly created, otherwise empty repositories (different chunking due to different hashing due to different key adds extra differences).

Using LZ4 and the original 1.1.9:
```
This archive:              304.99 GB            124.05 GB             92.03 GB
All archives:              314.41 GB            130.24 GB             92.03 GB

Chunk index:                  493930               559552
```

Using LZ4 and a patched 1.1.9 with this change:
```
This archive:              304.99 GB            123.87 GB             91.86 GB
All archives:              314.41 GB            130.05 GB             91.86 GB

Chunk index:                  493644               558659
```

The actual repositories on disk on the backup server:
```
# du -s comp-*
89921264	comp-orig
89760448	comp-patch
```

Oh, and a `borg list --format '{sha256}  {path}{NL}' ` lead to identical results (and matching on disk files) for both, too. :wink:

### DIFF
--- a/src/borg/compress.pyx
+++ b/src/borg/compress.pyx
@@ -72,7 +72,8 @@ cdef class CompressorBase:
         using Auto compression it needs to determine the _actual_ target compression of a chunk
         in order to detect whether it should be recompressed.
 
-        For all Compressors that are not Auto this always returns *self*.
+        Any compressor may return a compressor other than *self*, like e.g. the CNONE compressor,
+        and should actually do so if *data* would become larger on compression.
         """
         return self
 
@@ -95,6 +96,50 @@ cdef class CompressorBase:
         # strip ID bytes
         return data[2:]
 
+cdef class DecidingCompressor(CompressorBase):
+    """
+    base class for (de)compression classes that (based on an internal _decide
+    method) decide whether and how to compress data.
+    """
+    name = 'decidebaseclass'
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _decide(self, data):
+        """
+        Decides what to do with *data*. Returns (compressor, compressed_data).
+
+        *compressed_data* can be the result of *data* being processed by *compressor*,
+        if that is generated as a side-effect of the decision process, or None otherwise.
+
+        This private method allows for more efficient implementation of compress()
+        and decide_compress() making use of *compressed_data*, if already generated.
+        """
+        raise NotImplementedError
+
+    def decide(self, data):
+        return self._decide(data)[0]
+
+    def decide_compress(self, data):
+        """
+        Decides what to do with *data* and handle accordingly. Returns (compressor, compressed_data).
+
+        *compressed_data* is the result of *data* being processed by *compressor*.
+        """
+        compressor, compressed_data = self._decide(data)
+
+        if compressed_data is None:
+            compressed_data = compressor.compress(data)
+
+        if compressor is self:
+            # call super class to add ID bytes
+            return self, super().compress(compressed_data)
+
+        return compressor, compressed_data
+
+    def compress(self, data):
+        return self.decide_compress(data)[1]
 
 class CNONE(CompressorBase):
     """
@@ -113,7 +158,7 @@ class CNONE(CompressorBase):
         return data
 
 
-class LZ4(CompressorBase):
+class LZ4(DecidingCompressor):
     """
     raw LZ4 compression / decompression (liblz4).
 
@@ -128,7 +173,12 @@ class LZ4(CompressorBase):
     def __init__(self, **kwargs):
         pass
 
-    def compress(self, idata):
+    def _decide(self, idata):
+        """
+        Decides what to do with *data*. Returns (compressor, lz4_data).
+
+        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
+        """
         if not isinstance(idata, bytes):
             idata = bytes(idata)  # code below does not work with memoryview
         cdef int isize = len(idata)
@@ -142,7 +192,11 @@ class LZ4(CompressorBase):
             osize = LZ4_compress_default(source, dest, isize, osize)
         if not osize:
             raise Exception('lz4 compress failed')
-        return super().compress(dest[:osize])
+        # only compress if the result actually is smaller
+        if osize < isize:
+            return self, dest[:osize]
+        else:
+            return NONE_COMPRESSOR, None
 
     def decompress(self, idata):
         if not isinstance(idata, bytes):
@@ -174,7 +228,7 @@ class LZ4(CompressorBase):
         return dest[:rsize]
 
 
-class LZMA(CompressorBase):
+class LZMA(DecidingCompressor):
     """
     lzma compression / decompression
     """
@@ -187,10 +241,18 @@ class LZMA(CompressorBase):
         if lzma is None:
             raise ValueError('No lzma support found.')
 
-    def compress(self, data):
+    def _decide(self, data):
+        """
+        Decides what to do with *data*. Returns (compressor, lzma_data).
+
+        *lzma_data* is the LZMA result if *compressor* is LZMA as well, otherwise it is None.
+        """
         # we do not need integrity checks in lzma, we do that already
-        data = lzma.compress(data, preset=self.level, check=lzma.CHECK_NONE)
-        return super().compress(data)
+        lzma_data = lzma.compress(data, preset=self.level, check=lzma.CHECK_NONE)
+        if len(lzma_data) < len(data):
+            return self, lzma_data
+        else:
+            return NONE_COMPRESSOR, None
 
     def decompress(self, data):
         data = super().decompress(data)
@@ -200,7 +262,7 @@ class LZMA(CompressorBase):
             raise DecompressionError(str(e)) from None
 
 
-class ZSTD(CompressorBase):
+class ZSTD(DecidingCompressor):
     """zstd compression / decompression (pypi: zstandard, gh: python-zstandard)"""
     # This is a NOT THREAD SAFE implementation.
     # Only ONE python context must to be created at a time.
@@ -212,7 +274,12 @@ class ZSTD(CompressorBase):
         super().__init__(**kwargs)
         self.level = level
 
-    def compress(self, idata):
+    def _decide(self, idata):
+        """
+        Decides what to do with *data*. Returns (compressor, zstd_data).
+
+        *zstd_data* is the ZSTD result if *compressor* is ZSTD as well, otherwise it is None.
+        """
         if not isinstance(idata, bytes):
             idata = bytes(idata)  # code below does not work with memoryview
         cdef int isize = len(idata)
@@ -227,7 +294,11 @@ class ZSTD(CompressorBase):
             osize = ZSTD_compress(dest, osize, source, isize, level)
         if ZSTD_isError(osize):
             raise Exception('zstd compress failed: %s' % ZSTD_getErrorName(osize))
-        return super().compress(dest[:osize])
+        # only compress if the result actually is smaller
+        if osize < isize:
+            return self, dest[:osize]
+        else:
+            return NONE_COMPRESSOR, None
 
     def decompress(self, idata):
         if not isinstance(idata, bytes):
@@ -306,44 +377,54 @@ class Auto(CompressorBase):
 
     def _decide(self, data):
         """
-        Decides what to do with *data*. Returns (compressor, lz4_data).
+        Decides what to do with *data*. Returns (compressor, compressed_data).
 
-        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
+        *compressor* is the compressor that is decided to be best suited to compress
+        *data*, *compressed_data* is the result of *data* being compressed by a
+        compressor, which may or may not be *compressor*!
+
+        There are three possible outcomes of the decision process:
+        * *data* compresses well enough for trying the more expensive compressor
+          set on instantiation to make sense.
+          In this case, (expensive_compressor_class, lz4_compressed_data) is
+          returned.
+        * *data* compresses only slightly using the LZ4 compressor, thus trying
+          the more expensive compressor for potentially little gain does not
+          make sense.
+          In this case, (LZ4_COMPRESSOR, lz4_compressed_data) is returned.
+        * *data* does not compress at all using LZ4, in this case
+          (NONE_COMPRESSOR, none_compressed_data) is returned.
+
+        Note: While it makes no sense, the expensive compressor may well be set
+        to the LZ4 compressor.
         """
-        lz4_data = LZ4_COMPRESSOR.compress(data)
-        # lz4_data includes the compression type header, while data does not yet
-        ratio = len(lz4_data) / (len(data) + 2)
+        compressor, compressed_data = LZ4_COMPRESSOR.decide_compress(data)
+        # compressed_data includes the compression type header, while data does not yet
+        ratio = len(compressed_data) / (len(data) + 2)
         if ratio < 0.97:
-            return self.compressor, lz4_data
-        elif ratio < 1:
-            return LZ4_COMPRESSOR, lz4_data
+            return self.compressor, compressed_data
         else:
-            return NONE_COMPRESSOR, None
+            return compressor, compressed_data
 
     def decide(self, data):
         return self._decide(data)[0]
 
     def compress(self, data):
-        compressor, lz4_data = self._decide(data)
-        if compressor is LZ4_COMPRESSOR:
+        compressor, cheap_compressed_data = self._decide(data)
+        if compressor in (LZ4_COMPRESSOR, NONE_COMPRESSOR):
             # we know that trying to compress with expensive compressor is likely pointless,
-            # but lz4 managed to at least squeeze the data a bit.
-            return lz4_data
-        if compressor is NONE_COMPRESSOR:
-            # we know that trying to compress with expensive compressor is likely pointless
-            # and also lz4 did not manage to squeeze the data (not even a bit).
-            uncompressed_data = compressor.compress(data)
-            return uncompressed_data
+            # so we fallback to return the cheap compressed data.
+            return cheap_compressed_data
         # if we get here, the decider decided to try the expensive compressor.
-        # we also know that lz4_data is smaller than uncompressed data.
-        exp_compressed_data = compressor.compress(data)
-        ratio = len(exp_compressed_data) / len(lz4_data)
+        # we also know that the compressed data returned by the decider is lz4 compressed.
+        expensive_compressed_data = compressor.compress(data)
+        ratio = len(expensive_compressed_data) / len(cheap_compressed_data)
         if ratio < 0.99:
             # the expensive compressor managed to squeeze the data significantly better than lz4.
-            return exp_compressed_data
+            return expensive_compressed_data
         else:
             # otherwise let's just store the lz4 data, which decompresses extremely fast.
-            return lz4_data
+            return cheap_compressed_data
 
     def decompress(self, data):
         raise NotImplementedError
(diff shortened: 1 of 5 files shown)


src/borg/testsuite/compress.py
```diff
@@ -43,14 +43,16 @@ def test_lz4():
...
 
 
-def test_lz4_buffer_allocation():
+def test_lz4_buffer_allocation(monkeypatch):
+    # disable fallback to no compression on incompressible data
+    monkeypatch.setattr(LZ4, 'decide', lambda always_compress: LZ4)
     # test with a rather huge data object to see if buffer allocation / resizing works
     data = os.urandom(5 * 2**20) * 10  # 50MiB badly compressible data
     assert len(data) == 50 * 2**20
     c = get_compressor(name='lz4')
```
ThomasWaldmann: change this line to
```
c = Compressor('lz4')
```
and keep the line below as it was?

src/borg/compress.pyx
```diff
@@ -126,9 +126,14 @@ class LZ4(CompressorBase):
     name = 'lz4'
 
     def __init__(self, **kwargs):
-        pass
+        self.none = get_compressor('none')
```
ThomasWaldmann: Hmm, thinking about a slight micro opt and also reduction of code duplication here:

Could it just be one global NONE_COMPRESSOR object, initialized only once instead of each compressor class constructor doing it every time?

ThomasWaldmann:
@elho did you see my feedback?

elho:
> @elho did you see my feedback?

Yes, sorry for the delay. You gave the feedback shortly after I left for summer vacation and I didn't have the time to work on it yet since back. It's defenitely on the TODO list. :smile:

ThomasWaldmann:
ping :-)

src/borg/testsuite/archiver.py
```diff
@@ -2124,6 +2124,7 @@ def test_compression_lz4_compressible(self):
     def test_compression_lz4_uncompressible(self):
         size, csize = self._get_sizes('lz4', compressible=False)
         assert csize >= size
+        assert csize == size + 3  # same as compression 'none'
```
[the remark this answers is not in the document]
elho: While, of course, the first test is pointless given the second, reason I left it that way is that it is exactly the code as used by [`test_compression_none_compressible()`](https://github.com/borgbackup/borg/blob/d85c4f9e5dbabe77f93b50f7b6053b41436b352e/src/borg/testsuite/archiver.py#L2091) and [`test_compression_none_uncompressible()`](https://github.com/borgbackup/borg/blob/d85c4f9e5dbabe77f93b50f7b6053b41436b352e/src/borg/testsuite/archiver.py#L2096).
So I guess the way to go is a separate PR cleaning those existing cases up plus changing this PR...

src/borg/compress.pyx
elho: That also should then be independendly done to the auto compressor as well, which I did not touch at all and which [in its init function does the same thing for both lz4 and none compressors](https://github.com/borgbackup/borg/blob/d85c4f9e5dbabe77f93b50f7b6053b41436b352e/src/borg/compress.pyx#L364)...

src/borg/testsuite/compress.py
elho: Yeah, that is indeed more straight forward.

src/borg/testsuite/archiver.py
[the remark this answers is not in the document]
elho: Created that separate PR #5042.

src/borg/compress.pyx
elho: Created that separate PR #5043.

elho:
Feedback adressed, rebased onto current master and fixed a typo in a previous commit message (thus the force-push).

ThomasWaldmann:
I noticed some bits that maybe could be improved, see XXX below:

```
    # XXX from class Auto
    def _decide(self, data):
        """
        Decides what to do with *data*. Returns (compressor, lz4_data).
        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
        """
        # XXX note that lz4_data includes the compression type header
        lz4_data = LZ4_COMPRESSOR.compress(data)

        # XXX so guess we need to adjust here a litte (+/- 2)
        ratio = len(lz4_data) / len(data)

        if ratio < 0.97:
            return self.compressor, lz4_data

        # XXX below stuff was already done in LZ4.compress, it
        #         might return LZ4-compressed or NONE-compressed data
        elif ratio < 1:
            return LZ4_COMPRESSOR, lz4_data
        else:
            return NONE_COMPRESSOR, None

    def decide(self, data):
        return self._decide(data)[0]

    def compress(self, data):
        compressor, lz4_data = self._decide(data)

        # XXX below 2 if-blocks also need to be adapted (see above)
        if compressor is LZ4_COMPRESSOR:
            # we know that trying to compress with expensive compressor is likely pointless,
            # but lz4 managed to at least squeeze the data a bit.
            return lz4_data
        if compressor is NONE_COMPRESSOR:
            # we know that trying to compress with expensive compressor is likely pointless
            # and also lz4 did not manage to squeeze the data (not even a bit).
            uncompressed_data = compressor.compress(data)
            return uncompressed_data

        # if we get here, the decider decided to try the expensive compressor.
        # we also know that lz4_data is smaller than uncompressed data.
        exp_compressed_data = compressor.compress(data)

        # XXX note: below line is correct, because both are with compression-type header
        ratio = len(exp_compressed_data) / len(lz4_data)

        if ratio < 0.99:
            # the expensive compressor managed to squeeze the data significantly better than lz4.
            return exp_compressed_data
        else:
            # otherwise let's just store the lz4 data, which decompresses extremely fast.
            return lz4_data
```

elho:
> ```
>     # XXX from class Auto
>     def _decide(self, data):
>         # XXX note that lz4_data includes the compression type header
>         lz4_data = LZ4_COMPRESSOR.compress(data)
> 
>         # XXX so guess we need to adjust here a litte (+/- 2)
>         ratio = len(lz4_data) / len(data)

>         # XXX note: below line is correct, because both are with compression-type header
>         ratio = len(exp_compressed_data) / len(lz4_data)```

Good catch! But as this particular issue is present in the existing code that is not touched by this PR, I fixed this part in separate PR#5076.

elho:
> ```
>     # XXX from class Auto
>     def _decide(self, data):
>         """
>         Decides what to do with *data*. Returns (compressor, lz4_data).
>         *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
>         """
> ```

That doc string is inaccurate, too. lz4_data will also be the LZ4 result, if the expensive compressor is chasen, which may or may not be LZ4 as well. `compress()` just knows that und does not use lz4_data in that case.
So rephrasing the doc string to whatever is accurate after fixing the other issues is on the to-do list, too.

src/borg/compress.pyx
```diff
@@ -306,44 +381,40 @@ class Auto(CompressorBase):
 
     def _decide(self, data):
         """
-        Decides what to do with *data*. Returns (compressor, lz4_data).
+        Decides what to do with *data*. Returns (compressor, compressed_data).
 
-        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
+        *compressed_data* is the LZ4 result unless *compressor* is NONE, in which
+        case it is the uncompressed NONE compressor result.
```
ThomasWaldmann, line 387: 2x NONE -> NONE_COMPRESSOR?

ThomasWaldmann: Also, I guess it might be less confusing to just say:

There are 2 outcomes of this decision method:
- the data looks compressible (according to a more or less successful lz4 compression attempt), in that case we return (expensive_compressor_class, lz4_compressed_data)
- the data looks incompressible (according to a failed lz4 compression attempt), in that case we return (NONE_COMPRESSOR, none_compressed_data)

src/borg/compress.pyx
```diff
@@ -306,44 +381,40 @@ class Auto(CompressorBase):
...
-            return NONE_COMPRESSOR, None
+            return compressor, compressed_data
 
     def decide(self, data):
         return self._decide(data)[0]
 
     def compress(self, data):
-        compressor, lz4_data = self._decide(data)
-        if compressor is LZ4_COMPRESSOR:
+        compressor, chp_compressed_data = self._decide(data)
```
ThomasWaldmann, line 401: maybe write cheap if you mean cheap. :)

src/borg/compress.pyx
```diff
@@ -306,44 +381,40 @@ class Auto(CompressorBase):
...
+        # the cheap compressed data can either be LZ4 compressed or, if that turned out pointless,
+        # NONE compressor type uncompressed data.
```
ThomasWaldmann, line 403: guess this comment would not be needed with the improved docstring, see above.

src/borg/compress.pyx
ThomasWaldmann: plus:

note: we do not forbid to set the expensive_compressor to LZ4 (although it doesn't make sense).

src/borg/compress.pyx
```diff
@@ -142,7 +147,38 @@ class LZ4(CompressorBase):
...
+        """
+        compressor, compressed_data = self._decide(data)
+
+        if compressor is self:
+            return self, super().compress(compressed_data)
+
+        if compressed_data is None:
+            compressed_data = compressor.compress(data)
+
+        return compressor, compressed_data
```
ThomasWaldmann, line 173: considering what `self._decide` does, this could be just a simple `if ... return ... else return ...`.

src/borg/compress.pyx
```diff
@@ -142,7 +147,38 @@ class LZ4(CompressorBase):
...
+
+    def compress(self, data):
+        compressor, lz4_data = self._decide(data)
+
+        if compressor is self:
+            return super().compress(lz4_data)
+        else:
+            return compressor.compress(data)
```
ThomasWaldmann, line 181: considering that this is basically `decide_compress(...)[0]`, could we move both to `BaseCompressor` class to avoid code duplication?

src/borg/compress.pyx
elho: I used chp_ solely to mathch the existing exp_ of the expensive one, I'll change both to the long form then. ;)

ThomasWaldmann: oh, ok :)

src/borg/compress.pyx
elho: I'd rather go with the class name based "CNONE compressor" there rather than what happens to just be the global variable for it, matching the existing part of this comment and others talking about LZ4 and not LZ4_COMPRESSOR either, More consistent and less cumbersome to read. :smile: 
I'll clarify the doc string some more. You just proved it to be necessary: There actually is three outcomes, not two (thus the "unless compressor is NONE"). :smile:  Besides the two you desribed, the data may have been compressible with LZ4 yielding a ratio for which 0.97<ratio<1 holds true, in which case (LZ4_COMPRESSOR, lz4_compressed_data) is returned. That is one of the original cases, the case of NONE is new with my change and happens if LZ4 compression does not lead to a smaller result than NONE.

ThomasWaldmann: CNONE is also fine with me. NONE is a bit too close to None though.

ThomasWaldmann: ah, right, overlooked that case.

src/borg/compress.pyx
elho: I did opt for the robust implementation depending less on how `_decide()` is actually implemented, covering the known special cases (compressor benig self and no compressed data being returned).

I see why you are leaning to a micro-optimization here instead, but do you still regard that the better choice if moving it to the `BaseCompressor`class, where a derived class could have its own `_decide()` that, e.g. similar to the one in auto compressor, may have more possible outcomes than the two cases of `(self, self_compressed_data)` and `(NONE_COMPRESSOR, None)`?

ThomasWaldmann: yeah, you're right. let's keep the safer variant.

src/borg/compress.pyx
elho: Need to rethink this. Sounds good at first, but then either one also relies on the internal "decide and maybe compress" `_decide()` method, that would have to be present in `BaseCompressor` as well then. Also the current `super().compress()` idiom for "add compressor type header" would go away.
The most simple compressor CNONE uses these and would then have to optimize things by replacing those methods with simpler ones.
So would any potential new compressor that had a way to implement `decide()` without ever having to actually compress the data. The latter may be far fetched, but what I'm saying is, is that it doesn't feel right to give up the truely general nature of the base class and cause new code duplication to reduce another.

Maybe have one more layer of classes, ie. one like the current `BaseCompressor` that is general enough for a non-compressor like CNONE to be based on it without overhead, and another (like `SimpleCompressor` or some better name?) that adds the `decide()` based compression methods that are duplicate in the "normal" compressors (LZ4, LZMA, ZSD, any new algorithm)?

ThomasWaldmann: hmm, sounds more complicated than estimated...

maybe try your suggestion in the last paragraph, but it's also fine if it doesn't work out to be worthwhile.

elho: Yeah, I will try it in a sub-branch of this one, then we can just ignore it at any point it turns out not to be worth the effort.

Meanwhile, I will push the other changes, plus another comment that needed updating due to the change, here to have them out the door.

ThomasWaldmann:
Still working on this?

elho:
> Still working on this?

Yes, paused it for the easter weekend, but will pick it up this weekend. :slightly_smiling_face:

ThomasWaldmann:
Last thing to do for 1.2a8.

src/borg/compress.pyx
elho: Splitting out the deuplicate code into a separate class did turn out quite straight forward, so its fine to go straight into this PR's branch. :slightly_smiling_face:

src/borg/compress.pyx
```diff
@@ -96,6 +99,51 @@ cdef class CompressorBase:
...
+        """
+        Decides what to do with *data*. Returns (compressor, compressed_data).
+
+        *compressed_data* can be the result of *data* being processed by *compressor*,
+        if that is generated as a side-effect of the decision process, or None otherwise.
+
+        This private method allows for more efficient implementation of compress()
+        and decide_compress() making use of *compressed_data*, if already generated.
+        """
+        return self, data
```
ThomasWaldmann, line 123: hmm, isn't that something that **must** be implemented by derived classes?

so, `raise NotImplementedError` rather?

elho: It would work if it wasn't (it does return a compressor and data and due to the compressor being self, decide_compress() passes things up in hierarchy to eventually have an ID added), but there wouldn't be any compression, so it does not make sense to not implement it.
 (In other words, a FooCompressor derivedfrom DecidingCompressor that does not implement any methods of its own (and only sets ID) would be an overcomplicated NONE compressor).
Still go for `NotImplementedError`?

src/borg/compress.pyx
```diff
@@ -96,6 +99,51 @@ cdef class CompressorBase:
         # strip ID bytes
         return data[2:]
 
+cdef class DecidingCompressor(CompressorBase):
+    """
+    base class for (de)compression classes that ,
+    handles decide and compression based on an internal
```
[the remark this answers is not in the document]
elho: What I had on mind when writing compression was that compress() completes always, but the question is whether actual compression happens or not.
So other options would be "decision and compression" or maybe "decide() and resulting compression)" :thinking:

src/borg/compress.pyx
```diff
@@ -72,10 +72,14 @@ cdef class CompressorBase:
...
         in order to detect whether it should be recompressed.
 
-        For all Compressors that are not Auto this always returns *self*.
+        Any compressor may return a compressor other than *self*, like e.g. the CNONE compressor,
+        and should actually do so if *data* would become larger on compresson.
         """
         return self
 
+    def decide_compress(self, data):
+        raise NotImplementedError
```
ThomasWaldmann, line 81: remove `decide` and `decide_compress` from `CompressorBase` and only have them in `DecidingCompressor` base class?

src/borg/compress.pyx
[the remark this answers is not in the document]
ThomasWaldmann: how about:

base class for (de)compression classes that (based on internal _decide method) decides whether and how to compress data.

src/borg/compress.pyx
ThomasWaldmann: Well, if it doesn't make sense to not override it, i think it is better to require overriding it (by raising NotImplementedError).

src/borg/compress.pyx
elho: Makes sense I guess, no guarantee for it to be available and thus not requiring it and leaving it up to its users to only use it with compressors supporting it.

ThomasWaldmann: did you forget to remove the 2nd method or do you think it is needed?

elho: Both actually. :smile:  I did miss it reading your comment, but I also do think it is needed, it at least was in the base class (and thus any compressor) before any of my changes, and IIRC recreate does need to call decide() when recompressing, I'll check that.

elho: Yes, [`chunk_processor()`](https://github.com/borgbackup/borg/blob/27a41f5c1f8a8187c9ae129d1ec1de73e6a8712d/src/borg/archive.py#L1964) in class `ArchiveRecreater` does use `decide()` unconditionally, so it is needed.

ThomasWaldmann: ah, ok.

src/borg/compress.pyx
```diff
@@ -95,6 +96,51 @@ cdef class CompressorBase:
         # strip ID bytes
         return data[2:]
 
+cdef class DecidingCompressor(CompressorBase):
+    """
+    base class for (de)compression classes that (based on an internal _decide
+    method) decide whether and how to compress data.
+    .
```
ThomasWaldmann, line 103: there's a stray `.`.

src/borg/compress.pyx
```diff
@@ -72,7 +72,8 @@ cdef class CompressorBase:
         using Auto compression it needs to determine the _actual_ target compression of a chunk
         in order to detect whether it should be recompressed.
 
-        For all Compressors that are not Auto this always returns *self*.
+        Any compressor may return a compressor other than *self*, like e.g. the CNONE compressor,
+        and should actually do so if *data* would become larger on compresson.
```
ThomasWaldmann, line 76: typo: compress`i`on

src/borg/compress.pyx
```diff
@@ -306,44 +378,54 @@ class Auto(CompressorBase):
...
-        Decides what to do with *data*. Returns (compressor, lz4_data).
-
-        *lz4_data* is the LZ4 result if *compressor* is LZ4 as well, otherwise it is None.
+        Decides what to do with *data*. Returns (compressor, compressed_data).
+
+        *compressor* is the compressor that is decided to be best suited to compress
+        *data*, *compressed_data* is the result of *data* being compressed by a
+        compressor, which may or may not be *compressor*!
+
+        There is three possible outcomes of the decision process:
```
ThomasWaldmann, line 387: There `are` ...

elho:
With all that cleanups and refactoring alongside the main compression issue, I'll individually squash them. :smile:

elho:
Squashed it all into one, merging  the commit messages to reflect the overall changes.

ThomasWaldmann:
@elho thanks a lot!

PR #4687, opened 2019-07-17
Update documentation of borg create

tgdnt:
borg create examples within documentation mention 1.1 as a future version of borg.


ThomasWaldmann:
Thanks!

Issue #4692, opened 2019-07-23
Disable smartquotes in the sphinx configuration


cydave:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

QUESTION

## The Question

Regarding the sphinx documentation and it's configuration,
I was wondering if it wouldn't be a better idea to disable  
smart quoting mode (which is enabled by default).
The motivation behind my question/suggestion is that it might
improve the overall "copy-pasteability" of the documentation.
Currently there are various places in the rendered documentation
on readthedocs which prevent easy copy pasting of commands
due to the conversion of the double dashes to a unicode dash.

Wouldn't it be a better idea to disable the [smartquotes](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes) option
inside `conf.py` so people who are working on the documentation
don't have to escape each and every occurrence of a command-line
flag with two dashes?

ThomasWaldmann:
Can you link to a place in the docs that can't be copy and pasted due to this?

And make a PR afterwards?

cydave:
Sorry for my late reply,

https://borgbackup.readthedocs.io/en/master/quickstart.html#backup-compression contains
a blockquote with a command inside which was "victim" of a such conversion. I'm not quite sure yet
if this is due to the smartquotes configuration or if the "en dash" character ended up like this
due to an editor configuration. I'll try to dig a little bit further into this :)

ThomasWaldmann:
Yeah, this one example got rendered badly, when smartquotes are enabled. But the problem are not the smartquotes, but that the example is not made using the correct markup. In the fragment shown below, the first example is rendered badly (causing the c&p problem), the second example does not have this problem:

```
zstd is a modern compression algorithm and might be preferable over zlib and
lzma, except if you need compatibility to older borg versions (< 1.1.4) that
did not yet offer zstd.

    $ borg create --compression zstd,N /path/to/repo::arch ~

Other options are:

If you have a fast repo storage and you want minimum CPU usage, no compression::

    $ borg create --compression none /path/to/repo::arch ~
```

Note the `::` introducing the 2nd example, this should be used for the 1st one also.

So, I'ld rather prefer the markup being fixed (at all places) than the configuration being changed.

cydave:
Good eye, I guess we can close this issue then :)
Thanks for looking into it!

ThomasWaldmann:
[...]

PR #4703, opened 2019-08-01
Add smartquotes = False to sphinx conf.py

cydave:


docs/conf.py
```diff
@@ -35,6 +35,10 @@
 # The encoding of source files.
 #source_encoding = 'utf-8-sig'
 
+# Disable the use of smartquotes, which converts quotes and dashes to
+# typographically correct entities.
```
ThomasWaldmann: Maybe don't explain the pros of enabling it, but rather why it should be disabled.

ThomasWaldmann:
See also my comments in #4692 about the root cause of this issue (wrong markup).

ThomasWaldmann:
as this needs a different fix, i am closing this PR unmerged.

ThomasWaldmann:
thanks for pointing to the issue!

PR #4708, opened 2019-08-05
Fix various code blocks in the docs

### LINKED ISSUE #4692: Disable smartquotes in the sphinx configuration

ntova:
fixes #4692 

These were all occurences I could find, please comment if I missed one :)

### DIFF
--- a/docs/deployment/central-backup-server.rst
+++ b/docs/deployment/central-backup-server.rst
@@ -89,17 +89,23 @@ or relative remote repository path and can directly access the repositories at
 Client
 ------
 
-The client needs to initialize the `pictures` repository like this: ::
+The client needs to initialize the `pictures` repository like this:
+
+::
 
  borg init backup@backup01.srv.local:pictures
 
 Or with the full path (should actually never be used, as only for demonstrational purposes).
-The server should automatically change the current working directory to the `<client fqdn>` folder. ::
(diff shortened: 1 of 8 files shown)


ThomasWaldmann:
*.rst: can you check the style for these and maybe do them all in the same way?

Guess this style is the best / most explicit:
```
some text in the leading paragraph:

::

    code code code
```

*.py: here the situation is a bit different as the python code might output it more or less "as is" (please test) and also it is extracted to generate part of the html and man page docs.

ThomasWaldmann:
Fixed it. :)

PR #4751, opened 2019-09-06
create: make --noatime the default, deprecate --noatime, fixes #4673

### LINKED ISSUE #4673: Consider making `--no-atime` the default
The [docs already mention](https://borgbackup.readthedocs.io/en/stable/usage/notes.html#noatime-noctime) the problem that storing `atime` can create:

> it might also affect metadata stream deduplication: if only this timestamp changes between backups and is stored into the metadata stream, the metadata stream chunks won’t deduplicate just because of that.

What the docs don't mention is that on systems that don't support `O_NOATIME` (e.g. macOS), every single file will have a modified `atime` by virtue of borg scanning it. This means that the metadata stream will get almost no deduplication at all, resulting in large archive sizes even when ~no data has changed.

Additionally, the stored `atime` values won't really be useful, since they'll just reflect the time of the backup.

So, I think it makes sense for `--noatime` to be the default (with a `--atime` flag to opt-in), at least on systems where `O_NOATIME` is not available. (But I'd lean towards making it the default everywhere, since it's unclear that the value of storing it outweighs the downside for the average user.)

There of course could be suitable compatibility measures, like keeping the old default for existing repos.

ThomasWaldmann:
also: add --atime option to enable storing files' atime.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -192,6 +192,8 @@ Compatibility notes:
   which does not need an external Python interpreter.
   Maybe this requirement will be raised to Python 3.6 later.
 - freeing repository space only happens when "borg compact" is invoked.
+- borg create --noatime is deprecated. Not storing atime is the default behaviour
+  now (use --atime if you want to store the atime).
 - list: corrected mix-up of "isomtime" and "mtime" formats. Previously,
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
note: no backport of this.

PR #4761, opened 2019-09-21
Typo fix (neccessary -> necessary).

cherouvim:


ThomasWaldmann:
Thanks!

Just a note in case you find more stuff: you don't need to regenerate man pages or other automatically created files - this will be done at release time.

PR #4763, opened 2019-09-22
Add birthtime to FUSE entries

lamyj:
This PR fixes the `st_birthtime` (used e.g. in macOS) missing from the entries when using `borg mount` (note that `borg extract` does not exhibit this error).

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -556,10 +556,12 @@ class FuseOperations(llfuse.Operations, FuseBackend):
             entry.st_mtime_ns = mtime_ns
             entry.st_atime_ns = item.get('atime', mtime_ns)
             entry.st_ctime_ns = item.get('ctime', mtime_ns)
+            entry.st_birthtime_ns = item.get('birthtime', mtime_ns)
         else:
             entry.st_mtime = mtime_ns / 1e9
             entry.st_atime = item.get('atime', mtime_ns) / 1e9
             entry.st_ctime = item.get('ctime', mtime_ns) / 1e9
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Good catch!

Did you check if 1.1-maint branch also needs this? Do we have half-done birthtime support there also?

lamyj:
The 1.1-maint branch has the same error and the patch only differs by the line numbers:
```patch
diff --git a/src/borg/fuse.py b/src/borg/fuse.py
index e84d8ff5..5b2072a5 100644
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -541,10 +541,12 @@ def getattr(self, inode, ctx=None):
             entry.st_mtime_ns = mtime_ns
             entry.st_atime_ns = item.get('atime', mtime_ns)
             entry.st_ctime_ns = item.get('ctime', mtime_ns)
+            entry.st_birthtime_ns = item.get('birthtime', mtime_ns)
         else:
             entry.st_mtime = mtime_ns / 1e9
             entry.st_atime = item.get('atime', mtime_ns) / 1e9
             entry.st_ctime = item.get('ctime', mtime_ns) / 1e9
+            entry.st_birthtime = item.get('birthtime', mtime_ns) / 1e9
         return entry
 
     def listxattr(self, inode, ctx=None):
```

ThomasWaldmann:
Can you make a PR against 1.1-maint please?

lamyj:
I wasn't sure that this was worthy of a PR, but it's done (#4767).

PR #4767, opened 2019-09-27
Backport #4763 to 1.1-maint

lamyj:
Same as #4763 in the 1.1-maint branch: this PR fixes the st_birthtime (used e.g. in macOS) missing from the entries when using borg mount (note that borg extract does not exhibit this error).


ThomasWaldmann:
Thanks!

Issue #4764, opened 2019-09-23
borg info tracebacks on empty directory


witten:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes, all three. Specifically, read through usage init/info docs, FAQ, and searched open and closed issues.

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).

1.1.10

#### Operating system (distribution) and version.

Manjaro Linux 18.1-rc10 with Linux kernel 4.19.69

#### Hardware / network configuration, and filesystems used.

Lenovo ThinkPad X1 Carbon. No network used with Borg. ext4 filesystem.

#### How much data is handled by borg?

None.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
mkdir empty-dir ; borg info empty-dir
```

## Describe the problem you're observing.

Borg experiences a traceback and exits unceremoniously with an error exit code when `borg info` is run on an empty directory. Instead, I would expect Borg *not* to traceback and exit gracefully (with an error exit code). Side note: This directory is empty because it is a [Docker volume](https://docs.docker.com/storage/volumes/) mount, which when mounted for the first time (if it doesn't exist on the host previously), manifests as an empty directory.

My use case, for context: Within a [Borg wrapper script](https://torsion.org/borgmatic/), I'm trying to use `borg info` to probe for whether a Borg repository already exists, so I know whether I need to run `borg init` or not. I can't just run `borg init` unconditionally, because I can't tell from its exit code whether the repository already exists. (And I can't capture stderr and try to parse the "A repository already exists" message, because I need stderr to pass through to the user in case there are interactive repository creation messages on stderr.)

So, I'm hoping that I can use `borg info` or a similar Borg sub-command to determine whether a repository exists.. Even if it's on an empty directory. The existing traceback is [confusing to my users](https://projects.torsion.org/witten/borgmatic/issues/201#issuecomment-1813), so it would be great to have a traceback-free way to determine whether a repository exists in a directory.

I originally thought of just probing for the existence of the `<repository-directory-name>/config` path without using Borg, but that doesn't handle the remote repository case.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes. Just run `borg info empty-dir` again on the empty directory.

#### Include any warning/errors/backtraces from the system logs

```
Traceback (most recent call last):
[...]

PR #4778, opened 2019-10-05
exit instead of traceback when directory is no borg repository

tehtbl:
Hi!

here is a fix for #4764 :)

Please don't hesitate to contact me in case of open questions or advise for improvements.

Thanks,
Thomas

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -396,8 +396,8 @@ class Repository:
         try:
             st = os.stat(path)
         except FileNotFoundError:
-            logger.warning("%s is not a valid borgbackup directory." % (path))
-            os.exit(2)
+            self.close()
+            raise self.InvalidRepository(self.path)
         if not stat.S_ISDIR(st.st_mode):
             raise self.InvalidRepository(path)
         if lock:
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -396,7 +396,8 @@ class Repository:
         try:
             st = os.stat(path)
         except FileNotFoundError:
-            raise self.DoesNotExist(path)
+            logger.warning("%s is not a valid borgbackup directory." % (path))
+            os.exit(2)
         if not stat.S_ISDIR(st.st_mode):
             raise self.InvalidRepository(path)
         if lock:


witten:
Hooray!

src/borg/repository.py
```diff
@@ -396,7 +396,8 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
         try:
             st = os.stat(path)
         except FileNotFoundError:
-            raise self.DoesNotExist(path)
+            logger.warning("%s is not a valid borgbackup directory." % (path))
+            os.exit(2)
```
ThomasWaldmann: this does not seem to fix the issue you referred to, but it changes the behaviour if there is no repository directory at all.

not sure the behaviour change is the right way either, see the other places below where it handles misc. error conditions also by raising a specific exception, so why should it be different here?

in general, you need to reproduce the issue without any change, then fix it and try to reproduce it again (or even better: add a test, so we can automatically test that).

for repository code changes, you also need to test with a local repo and a ssh: repo (borg in client/server mode) and make sure both scenarious work ok.

tehtbl:
Hi @ThomasWaldmann 

thank you for your suggestions! I've updated the PR accordingly. 

I have reproduced the issue as reported in the issue itself via `mkdir empty-dir ; borg info empty-dir` while being in the Vagrant environment. There, I've also did my tests.

I am not fluent in writing tests yet. But if you want any test case for this PR please let me know and I will create one.

thanks,
Thomas

src/borg/repository.py
```diff
@@ -396,8 +396,8 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
         try:
             st = os.stat(path)
         except FileNotFoundError:
-            logger.warning("%s is not a valid borgbackup directory." % (path))
-            os.exit(2)
+            self.close()
```
ThomasWaldmann: did you look at what `close()` is doing and whether it is needed here?

src/borg/repository.py
```diff
@@ -396,7 +396,8 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
         try:
             st = os.stat(path)
         except FileNotFoundError:
-            raise self.DoesNotExist(path)
+            self.close()
+            raise self.InvalidRepository(self.path)
```
ThomasWaldmann: overall, you changed the "does not exist" (which is true) into a "invalid" (which is questionable) exception.

what's the point?

ThomasWaldmann: also, please do not merge master into your branch if there is no reason to do so.

tehtbl:
I re-checked all about the issue again. Long story short, the PR #4461 already fixed the issue. The Bug is still valid in version 1.1.10 but no longer in the current master branch (1.2.0a8.dev9+g73ab1f1d.d20191007).

Sorry for taking your time :/

ThomasWaldmann:
@tehtbl oh, interesting find - thanks for noting this.

i'll have a look whether we want to backport #4461.

Issue #4752, opened 2019-09-06
master: python setup.py build_man is broken


ThomasWaldmann:
somehow it doesn't process `borgfs` correctly:

```
$ python setup.py build_man
Detected OpenSSL [via pkg-config]
Using bundled BLAKE2
Detected and preferring liblz4 [via pkg-config]
Using bundled ZSTD
Using bundled xxhash
running build_man
building man pages (in docs/man)
building man page borg-benchmark(1)
building man page borg-benchmark-crud(1)
building man page borgfs(1)
building man page borg-borgfs-benchmark(1)
building man page borg-borgfs-benchmark-crud(1)
error: [Errno 2] No such file or directory: 'docs/usage/borgfs_benchmark_crud.rst'
```

ThomasWaldmann:
Because of that, the related man pages will be missing in 1.2.0a7.

Gu1nness:
Hum,
I'm interested by this one.
However I'm not sure if it is just a matter of creating the good files for the borgfs commands, or some python code in the setup_docs.py file (which reads the files) in order to create them automatically.

ThomasWaldmann:
The goal is to fix `python setup.py build_man` (or the `argparse` related code in `borg.archiver`?), so that build_man works without an issue.

`borg` is the usual cli command to invoke borgbackup, `borgfs` is a special cli command using by `fstab` / `mount` only in context of mounting a borg repo / archive.

`borg benchmark crud` exists as a command, but stuff like `borgfs benchmark crud` doesn't make sense nor does it need docs to be generated.

Gu1nness:
Ok right,
I'll have a look at it then!
Thanks for your answer!

ThomasWaldmann:
BTW, this is likely a regression compared to older borg releases - I only noticed this brokenness recently.
So maybe checking older releases whether they work ok and comparing new to older code could give hints about what is broken.

Gu1nness:
Sure!
I'll have a look at the former releases!

Gu1nness:
As I said in the commit message, the issue #4471 generated way more problems than what was reported here.
It actually created all the commands that existed for borg but for borgfs, creating a duplicate command of borg.
This might be something you want to release soon in order to have everything patched ASAP.

PR #4779, opened 2019-10-09
Fixes the man pages creation

Gu1nness:
The issue https://github.com/borgbackup/borg/issues/4471 solution
created many commands for borgfs which shouldn't exist in addition to
creating issue https://github.com/borgbackup/borg/issues/4752.

Fixes https://github.com/borgbackup/borg/issues/4752 .

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2604,6 +2604,22 @@ class Archiver:
                 group.add_argument('--last', metavar='N', dest='last', default=0, type=positive_int_validator,
                                    help='consider last N archives after other filters were applied')
 
+        def define_borg_mount(subparser):
+            subparser.set_defaults(func=self.do_mount)
+            subparser.add_argument('location', metavar='REPOSITORY_OR_ARCHIVE', type=location_validator(),
+                                help='repository or archive to mount')
+            subparser.add_argument('mountpoint', metavar='MOUNTPOINT', type=str,
+                                help='where to mount filesystem')
+            subparser.add_argument('-f', '--foreground', dest='foreground',
+                                action='store_true',
+                                help='stay in foreground, do not daemonize')
+            subparser.add_argument('-o', dest='options', type=str,
+                                help='Extra mount options')
+            define_archive_filters_group(subparser)
+            subparser.add_argument('paths', metavar='PATH', nargs='*', type=str,
+                                   help='paths to extract; patterns are supported')
+            define_exclusion_group(subparser, strip_components=True)
+
         parser = argparse.ArgumentParser(prog=self.prog, description='Borg - Deduplicated Backups',
                                          add_help=False)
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2604,6 +2604,22 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
                 group.add_argument('--last', metavar='N', dest='last', default=0, type=positive_int_validator,
                                    help='consider last N archives after other filters were applied')
 
+        def define_borg_mount(subparser):
```
ThomasWaldmann: guess this should be `parser`.

that the parser can be in fact a subparser sometimes is a detail not needed here (and also is not true for borgfs).

src/borg/archiver.py
```diff
@@ -2623,6 +2639,55 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+        signals or crashes do not.
+        """)
+
+        if parser.prog == 'borgfs':
+            parser.description = self.do_mount.__doc__
+            parser.epilog = mount_epilog
+            parser.formatter_class = argparse.RawDescriptionHelpFormatter
+            parser.help = 'mount repository'
+            subparser = parser
+            define_borg_mount(subparser)
```
ThomasWaldmann: replace these 2 lines by:
```
define_borg_mount(parser)
```

The introduction of the `subparser` name is a relict from the old code structure - but as you have it as a function paramenter now, it is not needed any more.

fantasya-pbem:
Oh dear, guess that is my fault, the fix for issue #4471 was mine...
I can confirm that build_man now builds successful.

Gu1nness:
Indeed most of the patch is moving the code around.
I patched your remarks, I was tired when I wrote this it seems :)

ThomasWaldmann:
Thanks!

Is this change only needed for master or also for 1.1-maint?

ThomasWaldmann:
Also, see #4789, there might be another issue.

Gu1nness:
It does not seem to be necessary for 1.1-maint the man pages build correctly.
For #4789 the man page does not seem to list the commands especially (except in the "SEE ALSO" section, in which they are alphabetically unordered (but logically ordered : 
```man
borg-init(1), borg-create(1), borg-mount(1),  borg-extract(1),  borg-list(1),  borg-info(1),  borg-delete(1),  borg-prune(1), borg-recreate(1
```

ThomasWaldmann:
1.1-maint: ok, good.

what i meant with #4789 issue is that the borg.1 manpage is not updated.

Gu1nness:
Oh yeah indeed.
Having a look at the code, I can see [here](https://github.com/borgbackup/borg/blob/master/docs/man_intro.rst) that the date is hardcoded in the man_intro.
Let me patch this on my fork and I'll give you the corresponding commit :)

Issue #4532, opened 2019-04-28
borg backup and btrfs 


manfredlotz:
This is Fedora 30 Beta and borg is version 1.1.9

When I do a backup of a `btrfs` file system I first do a readonly snapshot, then back up the readonly snapshot.

Example
```
btrfs sub snap -r /home /.snapshots/home
borg create -v -stats /mnt/borg_repos::home.20190428 /.snapshots/home
btrfs sub del /.snapshots/home
```
The problem with this approach is that instead of paths starting with `/home` I have now `/.snapshots/home`.

Anything, I can do here.

ThomasWaldmann:
```
cd /.snapshots
borg create -v -stats /mnt/borg_repos::home.20190428 home
```

alternatively: do it as before and strip away the `.snapshots` at extraction time, see `borg extract --help`.

ThomasWaldmann:
TODO: check FAQ and docs whether this is already covered. if not, a PR for this is welcome!

manfredlotz:
`cd /.snapshots` aso is a possible solution for `/home`.

But I have other mount points like for instance: `/data/images` where it doesn't fit nicely. Sure, I could workaround by naming the snapshot: `data.images`.

It would be great to have a mapping parameter in the `borg create` subcommand.

manfredlotz:
I think I have to deal with it when doing the backup because otherwise the paths look too weird when mounting the repository.

ThomasWaldmann:
IIRC you can also strip for `borg mount`.

fantasya-pbem:
I take this ToDO and make a PR with a new FAQ, mentioning `extract` and `mount` PATH.

PR #4781, opened 2019-10-13
[DOC] #4532 - Add FAQ for restoring from BTRFS snapshots.

fantasya-pbem:
I created one new FAQ entry that shows usage of PATH feature in extract and mount, in addition to alternative of creating the snapshot archive from within the snapshots directory.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -702,6 +702,36 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
 .. _pipeviewer: http://www.ivarch.com/programs/pv.shtml
 
 
+How can I restore from a BTRFS snapshot to the original path?
+-------------------------------------------------------------
+
+When I do a backup of a BTRFS file system I first do a readonly snapshot and
+then back it up:
+
+::
+
+    btrfs subvolume snapshot -r /home /.snapshots/home
+    borg create /mnt/borg::home123 /.snapshots/home
+    btrfs subvolume delete /.snapshots/home
+
+The problem with this approach is that in a restore instead of paths starting with /home I
+have now /.snapshots/home.
+
+Solution 1: Run ``borg_create`` from within the snapshot directory.
+
+::
+
+    cd /.snapshots
+    borg create /mnt/borg::home123 home
+
+Solution 2: Use partial ``extract`` / ``mount`` by specifying a PATH.
+
+::
+
+    borg extract /mnt/borg::home123 /.snapshots/home
+    borg mount /mnt/borg::home123 /mnt/restore /.snapshots/home
+
+
 I am having troubles with some network/FUSE/special filesystem, why?
 --------------------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -711,25 +711,18 @@ then back it up:
 ::
 
     btrfs subvolume snapshot -r /home /.snapshots/home
-    borg create /mnt/borg::home123 /.snapshots/home
+    borg create /path/to/repo::home_backup /.snapshots/home
     btrfs subvolume delete /.snapshots/home
 
 The problem with this approach is that in a restore instead of paths starting with /home I
 have now /.snapshots/home.
 
-Solution 1: Run ``borg_create`` from within the snapshot directory.
+Solution: Run ``borg_create`` from within the snapshot directory.
 
 ::
 
     cd /.snapshots
-    borg create /mnt/borg::home123 home
-
-Solution 2: Use partial ``extract`` / ``mount`` by specifying a PATH.
-
-::
-
-    borg extract /mnt/borg::home123 /.snapshots/home
-    borg mount /mnt/borg::home123 /mnt/restore /.snapshots/home
+    borg create /path/to/repo::home_backup
 
 
 I am having troubles with some network/FUSE/special filesystem, why?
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -702,6 +702,36 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+
+    cd /.snapshots
+    borg create /mnt/borg::home123 home
+
+Solution 2: Use partial ``extract`` / ``mount`` by specifying a PATH.
+
+::
+
+    borg extract /mnt/borg::home123 /.snapshots/home
+    borg mount /mnt/borg::home123 /mnt/restore /.snapshots/home
```
ThomasWaldmann: how is that helpful? if you have created the archive like shown above, there is nothing else in there anyway.

maybe for this case one could use --strip.

in general, I think documenting 1 good solution is better than documenting alternatives and making the docs unneccesarily long, so maybe just document the `cd` solution which avoids getting the unwanted path prefix into the archive.

fantasya-pbem: You're right. I did that before my other PR where I understood what PATH really does. So this is complete crap, I will try again.

docs/faq.rst
```diff
@@ -711,25 +711,18 @@ then back it up:
...
     cd /.snapshots
-    borg create /mnt/borg::home123 home
-
-Solution 2: Use partial ``extract`` / ``mount`` by specifying a PATH.
-
-::
-
-    borg extract /mnt/borg::home123 /.snapshots/home
-    borg mount /mnt/borg::home123 /mnt/restore /.snapshots/home
+    borg create /path/to/repo::home_backup
```
ThomasWaldmann: I just noticed that this is the solution to a way more general problem.

It is not just when snapshotting a btrfs subvolume (which mounts the snapshot), but basically when mounting any filesystem that is usually mounted elsewhere in the directory hierarchy. That could be also a bind-mount or a mount of a remote filesystem.

Also, I don't think it is needed to first show the commands for the wrong way, then point to the problem and then show the commands for the right way - just state the general problem and the solution.

fantasya-pbem: Made it so!

docs/faq.rst
```diff
@@ -702,6 +702,29 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
 .. _pipeviewer: http://www.ivarch.com/programs/pv.shtml
 
 
+How can I restore from a BTRFS snapshot to the original path?
```
ThomasWaldmann: still sounds btrfs specific.

how about "How can I avoid unwanted base directories getting stored into archives?"

docs/faq.rst
```diff
@@ -702,6 +702,29 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+-------------------------------------------------------------
+
+When I do a backup of a BTRFS file system I first do a readonly snapshot and
+then back it up:
```
ThomasWaldmann: You could replace this with a list of use cases:
- btrfs snapshots
- mounting a filesystem to another place than usual

docs/faq.rst
```diff
@@ -702,6 +702,29 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+
+::
+
+    btrfs subvolume snapshot -r /home /.snapshots/home
+    cd /.snapshots
+    borg create /path/to/repo::home_backup
+    btrfs subvolume delete /.snapshots/home
```
ThomasWaldmann: This would be a bit more generic and not limited to btrfs users:

mount --bind / /rootfs
cd /rootfs
borg create /path/to/repo::root_backup .

ThomasWaldmann: Note that `.` at the end.

docs/faq.rst
```diff
@@ -702,6 +702,29 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+
+Note that ``borg_create`` should be run from within the snapshot directory,
+because in the backup you don't want to have paths starting with the
+``.snapshots`` prefix; that might cause trouble when restoring from the backup.
```
ThomasWaldmann: Rephrase to more generic case, not referring to btrfs snapshots.

docs/faq.rst
```diff
@@ -702,6 +702,29 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+
+::
+
+    cd /.snapshots
+    borg create /path/to/repo::home_backup
```
ThomasWaldmann: guess this is not needed any more.

ThomasWaldmann:
btw, if you fix an issue, end your commit comment with ", fixes #1234567" then github will auto-close the issue. also, this is the format i need for docs/changes.rst.

docs/faq.rst
fantasya-pbem: Yes, forgot to delete it.

docs/faq.rst
fantasya-pbem: It started as a BTRFS-specific FAQ. But I will make it more generic.

docs/faq.rst
fantasya-pbem: OK

docs/faq.rst
fantasya-pbem: OK

fantasya-pbem: > Note that `.` at the end.

Why is this? Can't it be ommitted?

ThomasWaldmann: borg create needs at least 1 PATH argument.

fantasya-pbem: Good to know, I will have to fix my pull mode PR then, too.

Issue #3866, opened 2018-05-29
Borg extract hangs after about an hour. (strace: Timeout)


rvanlaar:
## Have you checked borgbackup docs, FAQ, and open Github issues?

yes

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE

#### Your borg version (borg -V).

borg 1.1.5

#### Operating system (distribution) and version.

Ubuntu 18.04 

#### Hardware / network configuration, and filesystems used.

Hardware: UX501VW laptop, 
network config: 250Mbit network connection
filesystems used: ext4

#### How much data is handled by borg?

90 GB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

borg -v -p --remote-path=borg1 extract USER@HOST.rsync.net:REPONAME::2018-05-28

## Describe the problem you're observing.

After a while all activity stops, no cpu usage, no downloads 
Strace shows:

select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

I can reproduce the issue. The issues shows up after about an hour. 
I tried restarting the export. This gave the same problem.

I would like two things for borg to support.
A better retry after the timeout and borg export to be able to resume after a failure.
Restoring an 80 GB repo isn't fun this way. 

#### Include any warning/errors/backtraces from the system logs

```
strace: Process 5656 attached
select(8, [5 7], [], [4 5 7], {tv_sec=0, tv_usec=783801}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
select(8, [5 7], [], [4 5 7], {tv_sec=1, tv_usec=0}) = 0 (Timeout)
[...]

PR #4782, opened 2019-10-13
[DOC] #3866 - Add FAQ for preventing SSH timeout in extract.

fantasya-pbem:
A new FAQ entry mentioning various ways to work around or prevent SSH timeout when extracting.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -509,6 +509,31 @@ Please disclose security issues responsibly.
 Common issues
 #############
 
+Why does Borg extract hang after some time?
+-------------------------------------------
+
+When I do a ``borg extract``, after a while all activity stops, no cpu usage,
+no downloads.
+
+This may happen when SSH terminates the connection on server side. You can
+configure SSH on client side to prevent this, for example in ~/.ssh/config:
+
+::
+
+    Host borg.example.com
+        ServerAliveInterval 300
+        ServerAliveCountMax 3
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -509,6 +509,31 @@ Please disclose security issues responsibly.
 Common issues
 #############
 
+Why does Borg extract hang after some time?
```
ThomasWaldmann: i think this should be split into 2 sections:

1. avoiding hangs by configuring ssh/sshd alive interval/count for both server and client side. this makes sure from both sides that the connection gets terminated and does not just hang for a longer time.

2. how to deal with unstable connections (doing smaller ops, using rsync)

docs/faq.rst
```diff
@@ -516,15 +516,30 @@ When I do a ``borg extract``, after a while all activity stops, no cpu usage,
 no downloads.
 
 This may happen when SSH terminates the connection on server side. You can
```
ThomasWaldmann: Hmm, not sure that it is really the case when "SSH terminates the connection on the server side".

If the connection is really terminated, I guess the client would notice that?

So, shall we maybe say a bit more generically "This may happen when the ssh connection is stuck somehow."?

fantasya-pbem: OK

ThomasWaldmann:
Thanks! Is this also needed for 1.1-maint branch?

Issue #4118, opened 2018-10-22
Pattern from file are not excluded `--exclude-from`


aggsol:
## Have you checked borgbackup docs, FAQ, and open Github issues?

YES and `borg help patterns`

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).

`borg 1.1.7`

#### Operating system (distribution) and version.

Ubuntu

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```sh
$ borg create --verbose --filter AME --list --stats --compression auto,lzma,6 --exclude-caches --exclude-from /home/ubuntu/workspace/bash-scripts/cubus-backup/ignoretest '/home/ubuntu/BorgRepo/::backup-{now}' /home/ubuntu/tmp
```

## Describe the problem you're observing.

The pattern provided in the file `ignoretest` as parameter for `--exclude-from` is ignored. File is backuped anyways. The file contains of the following lines:

```
# exclude file for borg backup
- /home/ubuntu/tmp/ignoreme/*
```

#### Can you reproduce the problem?

1. Use the given exclude file from above. 
2. Create the folder from the exclude
3. Create a file `touch testdata.data` in there
4. Execute `borg create ...`
5. See the file added with status `A`

aggsol:
The leading `- ` is not a valid in an exclude file even though the pattern example from `borg help patterns` even though the documentation claims that an `EXCLUDEFILE` should contain one pattern per line.

ThomasWaldmann:
hmm, iirc, exclude files just have the path/filenames and pattern files have the additional qualifiers (as they can contain root, include, exclude patterns, while exclude files only have excludes).

aggsol:
@ThomasWaldmann I guess so. The help text for patterns and the manual are not clear about the difference about `EXCLUDEFILE` and `PATTERNFILE` even though both contain patterns. Rephrasing it might be in order or an example usage of the `--exclude-from`  parameter.

fantasya-pbem:
I will try to improve the borg help patterns docs.

fantasya-pbem:
Not really a big addition, the current doc is quite good I think. Please comment.

aggsol:
I still think an example for `--exclude-from` would be the clearest solution.

PR #4783, opened 2019-10-13
[DOC] #4118 - Explain difference between --exclude and --pattern.

fantasya-pbem:
In my opinion the borg help pattern is already quite good in explaining the --exclude and --pattern features, so I just added a note that explicitly states the additional prefix in --pattern lines.
If you read this help carefully it should be sufficient.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2159,9 +2159,8 @@ class Archiver:
             .. note::
 
                 Via ``--pattern`` or ``--patterns-from`` you can define BOTH inclusion and exclusion
-                of files using pattern prefixes ``+`` and ``-``, so the PATH argument of the command
-                can be omitted. With ``--exclude`` and ``--exlude-from`` ONLY excludes are defined,
-                and the PATH argument defines the includes.
+                of files using pattern prefixes ``+`` and ``-``. With ``--exclude`` and
+                ``--exlude-from`` ONLY excludes are defined.
 
             Inclusion patterns are useful to include paths that are contained in an excluded
             path. The first matching pattern is used so if an include pattern matches before
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2159,8 +2159,9 @@ class Archiver:
             .. note::
 
                 Via ``--pattern`` or ``--patterns-from`` you can define BOTH inclusion and exclusion
-                of files using pattern prefixes ``+`` and ``-``. With ``--exclude`` and
-                ``--exlude-file`` ONLY excludes are defined.
+                of files using pattern prefixes ``+`` and ``-``, so the PATH argument of the command
+                can be omitted. With ``--exclude`` and ``--exlude-from`` ONLY excludes are defined,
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2155,6 +2155,13 @@ def do_break_lock(self, args, repository):
             A root path starts with the prefix `R`, followed by a path (a plain path, not a
             file pattern). An include rule starts with the prefix +, an exclude rule starts
             with the prefix -, an exclude-norecurse rule starts with !, all followed by a pattern.
+
+            .. note::
+
+                Via ``--pattern`` or ``--patterns-from`` you can define BOTH inclusion and exclusion
+                of files using pattern prefixes ``+`` and ``-``. With ``--exclude`` and
+                ``--exlude-file`` ONLY excludes are defined.
```
ThomasWaldmann: it is not `--exlude-file`, but `--exclude-from`.

one could also continue the last sentence with "... and you need to use ``--include`` ... to define includes ONLY.

fantasya-pbem: Stupid me...

But I'm not sure what you mean with --include. There is no --include mentioned in e.g. borg create, is this a new undocumented option?

src/borg/archiver.py
```diff
@@ -2155,6 +2155,14 @@ def do_break_lock(self, args, repository):
             A root path starts with the prefix `R`, followed by a path (a plain path, not a
             file pattern). An include rule starts with the prefix +, an exclude rule starts
             with the prefix -, an exclude-norecurse rule starts with !, all followed by a pattern.
+
+            .. note::
+
+                Via ``--pattern`` or ``--patterns-from`` you can define BOTH inclusion and exclusion
+                of files using pattern prefixes ``+`` and ``-``, so the PATH argument of the command
+                can be omitted. With ``--exclude`` and ``--exlude-from`` ONLY excludes are defined,
+                and the PATH argument defines the includes.
```
fantasya-pbem: Better now?

src/borg/archiver.py
```diff
@@ -2159,8 +2159,9 @@ def do_break_lock(self, args, repository):
             .. note::
 
                 Via ``--pattern`` or ``--patterns-from`` you can define BOTH inclusion and exclusion
-                of files using pattern prefixes ``+`` and ``-``. With ``--exclude`` and
-                ``--exlude-file`` ONLY excludes are defined.
+                of files using pattern prefixes ``+`` and ``-``, so the PATH argument of the command
+                can be omitted. With ``--exclude`` and ``--exlude-from`` ONLY excludes are defined,
+                and the PATH argument defines the includes.
```
ThomasWaldmann: remove ", so the PATH argument of the command can be omitted".
also remove ", and the PATH argument defines the includes".

the PATH arguments rather define the recursion roots / starting points (so it is rather like the `R` given in a pattern file).

fantasya-pbem: Done.

ThomasWaldmann:
Do we need same change for 1.1-maint?h

Issue #1552, opened 2016-08-29
push restore howto


ThomasWaldmann:
borgbackup push restore with correct uid:gid restore for file and folder ownership

Assume you have a pull backup system set up with borg, where a backup server pulls the data from the target via sshfs.

In case of a restore there is no direct way for borg to know the correct uid:gid of files, because it uses the /etc/passwd and /etc/group of the backup server. Using the archived uids/gids (no name mapping) might lead to wrong results, because after a reinstall the name-to-id mapping might have changed.

The workaround is chrooting into an sshfs mounted directory:

At the backup server do as root:

```
> mkdir /tmp/sshfs
> sshfs root@TARGETIP:/ /tmp/sshfs
> mkdir /tmp/sshfs/borgrepo
> mount --bind PATHTOREPO /tmp/sshfs/borgrepo
> cp /usr/local/bin/borg /tmp/sshfs/usr/local/bin/borg
> cd /tmp/sshfs
> for i in dev proc sys; do mount --bind /$i $i; done
> chroot /tmp/sshfs
```

Now we are on the backup system but inside a chroot with the target's rootfs.

```
> borg extract /borgrepo::ARCHIVENAME 
> exit
```

Now we left the chroot again, umount all the stuff:

```
> cd /tmp/sshfs
> for i in dev proc sys borgrepo; do umount ./$i; done
> cd /root
> fusermount -uz /tmp/sshfs
```

Thanks to secuser on IRC for this howto.

enkore:
Or simply `borg export-tar | ssh 'tar xf -'`, if the loss of detail is acceptable.

fantasya-pbem:
@enkore A simple ``borg export-tar`` seems not a good short alternative for a full restore. I'd not mention it is a push restore howto solution. Maybe only to make the results clear when uid/guid changes have happened.

fantasya-pbem:
- Created a new PR.
- A "-" was missing in export-tar (FILE parameter).
- Changed -e to --exclude for better readability.
- Tested everything, works as expected.

Issue #4780, opened 2019-10-12
Feature request: Borgbackup in pull data mode


ledufakademy:
I know that this is not an issue BUT ... push data to a repository can be, in certain circumstance, a very big security issue.
So is it possible to implement a features in order to backup a remote file system and store it in a local (or distant why not) borgbackup repository.
i already do that with sshFS ..so technically it's fully doingable.

Thank you for reading me.

ThomasWaldmann:
IIRC there are already other tickets talking about pull mode ideas, so closing this one as duplicate.

PR #4784, opened 2019-10-13
[DOC] #1552 - Add a pull backup / push restore howto.

fantasya-pbem:
Converted the recipe into a new Deployment article.

### DIFF
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -7,16 +7,22 @@ Backing up in pull mode
 Assuming you have a pull backup system set up with borg, where a backup server
 pulls the data from the target via sshfs.
 
-In this mode, the backup client's file system is mounted via sshfs on the backup
+In this mode, the backup client's file system is mounted remotely on the backup
 server. Pull mode is even possible if the SSH connection must be established by
-the client via a remote tunnel.
+the client via a remote tunnel. Other network file systems like NFS or SMB could
+be used as well, but sshfs is very simple to set up and probably the most secure
+one.
 
-There are some restrictions caused by sshfs. Unless you define UID and GID
-mappings when mounting via sshfs, owners and groups of the mounted file system
-will probably change, and you may not have access to those files if BorgBackup
-is not run with root privileges.
+There are some restrictions caused by sshfs. For example, unless you define UID
+and GID mappings when mounting via sshfs, owners and groups of the mounted file
+system will probably change, and you may not have access to those files if
+BorgBackup is not run with root privileges.
 
-[ADD MORE RESTRICTIONS HERE: filename encoding, xattrs, ACLs, bsdflags...]
+The sshfs is a FUSE file system and uses the SFTP protocol, so there may be also
+other unsupported features that the actual implementations of sshfs, libfuse and
+sftp on the backup server do not support, like file name encodings, ACLs, xattrs
+or bsdflags. So there is no guarantee that you are able to restore a system
+completely in every aspect from such a backup.
 
 To mount the client's root file system you will need root access to the client.
 This contradicts to the usual threat model of BorgBackup, where clients don't
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -24,17 +24,19 @@ sftp on the backup server do not support, like file name encodings, ACLs, xattrs
 or bsdflags. So there is no guarantee that you are able to restore a system
 completely in every aspect from such a backup.
 
-To mount the client's root file system you will need root access to the client.
-This contradicts to the usual threat model of BorgBackup, where clients don't
-need to trust the backup server (data is encrypted). In pull mode the server
-(when logged in as root) could cause unlimited damage to the client. Therefore,
-pull mode should be used only from servers you do fully trust!
+.. warning::
+
+    To mount the client's root file system you will need root access to the client.
+    This contradicts to the usual threat model of BorgBackup, where clients don't
+    need to trust the backup server (data is encrypted). In pull mode the server
+    (when logged in as root) could cause unlimited damage to the client. Therefore,
+    pull mode should be used only from servers you do fully trust!
 
 Creating a backup
 -----------------
 
 In this approach the client file system is simply mounted and then backed up.
-Note that the backup is created from within the mount point, so that all files
+Note that the backup is created from within the mount point so that all files
 in the archive have their original paths (otherwise they would be backed up with
 the mount point prefix, e.g. /mnt/sshfs/bin/bash instead of /bin/bash).
 
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -5,67 +5,46 @@ Backing up in pull mode
 =======================
 
 Assuming you have a pull backup system set up with borg, where a backup server
-pulls the data from the target via sshfs. In this mode, the backup client's file
+pulls the data from the target via SSHFS. In this mode, the backup client's file
 system is mounted remotely on the backup server. Pull mode is even possible if
 the SSH connection must be established by the client via a remote tunnel. Other
-network file systems like NFS or SMB could be used as well, but sshfs is very
+network file systems like NFS or SMB could be used as well, but SSHFS is very
 simple to set up and probably the most secure one.
 
-There are some restrictions caused by sshfs. For example, unless you define UID
-and GID mappings when mounting via sshfs, owners and groups of the mounted file
-system will probably change, and you may not have access to those files if
+There are some restrictions caused by SSHFS. For example, unless you define UID
+and GID mappings when mounting via ``sshfs``, owners and groups of the mounted
+file system will probably change, and you may not have access to those files if
 BorgBackup is not run with root privileges.
 
-The sshfs is a FUSE file system and uses the SFTP protocol, so there may be also
-other unsupported features that the actual implementations of sshfs, libfuse and
+SSHFS is a FUSE file system and uses the SFTP protocol, so there may be also
+other unsupported features that the actual implementations of ssfs, libfuse and
 sftp on the backup server do not support, like file name encodings, ACLs, xattrs
 or bsdflags. So there is no guarantee that you are able to restore a system
 completely in every aspect from such a backup.
 
 .. warning::
 
-    To mount the client's root file system you will need root access to the client.
-    This contradicts to the usual threat model of BorgBackup, where clients don't
-    need to trust the backup server (data is encrypted). In pull mode the server
-    (when logged in as root) could cause unlimited damage to the client. Therefore,
-    pull mode should be used only from servers you do fully trust!
+    To mount the client's root file system you will need root access to the
+    client. This contradicts to the usual threat model of BorgBackup, where
+    clients don't need to trust the backup server (data is encrypted). In pull
+    mode the server (when logged in as root) could cause unlimited damage to the
+    client. Therefore, pull mode should be used only from servers you do fully
+    trust!
 
 Creating a backup
 -----------------
 
-In this approach the client file system is simply mounted and then backed up.
-Note that the backup is created from within the mount point so that all files
-in the archive have their original paths (otherwise they would be backed up with
-the mount point prefix, e.g. /mnt/sshfs/bin/bash instead of /bin/bash).
+Generally, in a pull backup situation there is no direct way for borg to know
+the client's original UID:GID name mapping of files, because Borg would use
+``/etc/passwd`` and ``/etc/group`` of the backup server to map the names. To
+derive the right names, Borg needs to have access to the client's passwd and
+group files and use them in the backup process.
 
-::
-
-    sshfs root@host:/ /mnt/sshfs
-    cd /mnt/sshfs
-    borg create /path/to/repo::archive . # note the dot!
-    cd ~
-    umount /mnt/sshfs
-
-Restore methods
----------------
-
-The counterpart of a pull backup is a push restore. When restoring from a
-backup, you might have to take a closer look on the user and group IDs and names
-of the backup content. Depending on the type of restore – full restore or
-partial restore – there are different methods to make sure the correct IDs are
-restored. Generally, there is no direct way for borg to know the correct UID:GID
-of files, because it uses the ``/etc/passwd`` and the ``/etc/group`` of the
-backup server.
-
-Partial restore
-~~~~~~~~~~~~~~~
-
-In case of a partial restore, using the archived UIDs/GIDs might lead to wrong
-results, because after a reinstall the name-to-ID mapping might have changed.
-The workaround is chrooting into an sshfs mounted directory. In this example the
-whole client root file system is mounted. We use the stand-alone BorgBackup
-executable and copy it into the mounted file system to make Borg available after
-entering chroot; this can be skipped if Borg is already installed on the client.
+The solution to this problem is chrooting into an sshfs mounted directory. In
+this example the whole client root file system is mounted. We use the
+stand-alone BorgBackup executable and copy it into the mounted file system to
+make Borg available after entering chroot; this can be skipped if Borg is
+already installed on the client.
 
 ::
 
@@ -82,17 +61,19 @@ entering chroot; this can be skipped if Borg is already installed on the client.
     for i in dev proc sys; do mount --bind /$i $i; done
     chroot /tmp/sshfs
 
-Now we are on the backup system but inside a chroot with the target's root file
+Now we are on the backup system but inside a chroot with the client's root file
 system. We have a copy of Borg binary in ``/usr/local/bin`` and the repository
-in ``/borgrepo``. Now Borg is able to map the user/group names of the backup
-files to the actual IDs on the client, and we can run
+in ``/borgrepo``. Borg will back up the client's user/group names, and we can
+create the backup, retaining the original paths, excluding the repository:
 
 ::
 
-    borg extract /borgrepo::archive PATH
+    borg create -e /borgrepo /borgrepo::archive /
 
-to partially restore whatever we like. Finally we need to exit chroot, unmount
-all the stuff and clean up:
+(For the sake of simplicity only ``/borgrepo`` is excluded here. You may want to
+set up an exclude file with additional files and folders to be excluded.)
+
+Finally, we need to exit chroot, unmount all the stuff and clean up:
 
 ::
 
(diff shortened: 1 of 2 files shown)


infectormp:
Link to similar issue https://github.com/borgbackup/borg/issues/4780

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
+.. include:: ../global.rst.inc
+.. highlight:: none
+
+Backing up in pull mode
+=======================
+
+Assume you have a pull backup system set up with borg, where a backup server
```
ThomasWaldmann: "Assuming ..." ?

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+pulls the data from the target via sshfs.
+
+In this mode, the backup client's file system is mounted via sshfs on the backup
+server. Pull mode is even possible if the SSH connection must be established by
+the client via remote tunnel.
```
ThomasWaldmann: "via a ..."

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Creating backup
```
ThomasWaldmann: "Creating a ..."

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    borg create /REPOSITORY::archive /mnt/sshfs
+
+Note that this will back up every file with the ``/mnt/sshfs`` prefix. You can
+get rid of this prefix by changing the working directory before creating the
+backup:
+
+::
+
+    cd /mnt/sshfs
+    borg create /REPOSITORY::archive
```
ThomasWaldmann: Why not just document the latter way if it is more appropriate?

But if you cd into there, you also need to cd out of there again.

ThomasWaldmann: Also, there should be a note here, that if you backup server has root access to your clients, you need to have a lot of trust in the backup server (which is quite the opposite of borg's usualy threat model where the backup server is relatively untrusted) and a compromise of it would be a quite "global" desaster on potentially all the clients it has root access to.

ThomasWaldmann: Also, as everything goes over sshfs, borg can only backup stuff supported by sshfs, which might be less than when directly access a local filesystem.

Check: filenames (encoding?), xattrs, ACLs, bsdflags, ...

ThomasWaldmann: Also, if there is a UID/username GID/groupname mapping issue, it should be already mentioned here.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Creating backup
+---------------
+
+In this approach the client file system is simply mounted and then backed up.
+
+::
+
+    mkdir /mnt/sshfs
+    sshfs root@TARGET_IP:/ /mnt/sshfs
```
ThomasWaldmann: TARGET might be a bit misleading here maybe?

Isn't target usually rather meaning destination than source?

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    borg create /REPOSITORY::archive /mnt/sshfs
```
ThomasWaldmann: iirc we have some convention for a repo path and iirc it is not `/REPOSITORY`, please use the same you see elsewhere in the docs.

this also applies for other stuff like hosts, etc.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+
+The workaround is chrooting into an sshfs mounted directory.
+
+::
+
+    mkdir /tmp/sshfs
+    sshfs root@TARGET_IP:/ /tmp/sshfs
+    mkdir /tmp/sshfs/borgrepo
+    mount --bind /REPOSITORY /tmp/sshfs/borgrepo
+    cp /usr/local/bin/borg /tmp/sshfs/usr/local/bin/borg
```
ThomasWaldmann: this might only work for the borg fat binary - if one is using this from that location.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    for i in dev proc sys; do mount --bind /$i $i; done
+    chroot /tmp/sshfs
+
+Now we are on the backup system but inside a chroot with the target's root file
+system. We have a copy of BorgBackup binary in ``/usr/local/bin`` and the
+repository in ``/borgrepo``, and can run a simple
+
+::
+
+    borg extract /borgrepo::archive /mnt/sshfs
```
ThomasWaldmann: why do you need `/mnt/sshfs` here? 
i guess this won't work due to the leading slash.
also, if you use the `cd` way, it is not there anyway.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    exit
+
+in case all files have the ``/mnt/sshfs`` prefix) or just
```
ThomasWaldmann: there is no leading slash in paths in borg archives.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+
+::
+
+    borg extract /borgrepo::archive
```
ThomasWaldmann: also use "typical" borg repo path here.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    chroot /tmp/sshfs
+
+Now we are on the backup system but inside a chroot with the target's root file
+system. We have a copy of BorgBackup binary in ``/usr/local/bin`` and the
+repository in ``/borgrepo``, and can run a simple
+
+::
+
+    borg extract /borgrepo::archive /mnt/sshfs
+    exit
```
ThomasWaldmann: add ` # exit chroot` comment to clarify.

docs/deployment/pull-backup.rst
ThomasWaldmann: here same thing as above applies: if we do avoid the unwanted prefix, we also only need to show one way here for extraction.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    borg extract /borgrepo::archive
+    exit
+
+otherwise. We have now left chroot again and have to unmount all the stuff:
+
+::
+
+    cd /tmp/sshfs
+    for i in dev proc sys borgrepo; do umount ./$i; done
+    cd /root
```
ThomasWaldmann: `cd /` or just `cd`.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+    fusermount -uz /tmp/sshfs
```
ThomasWaldmann: what's -z doing?

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Thanks to secuser on IRC for this howto.
+
+Simple full restore
+-------------------
+
+If you don't care for the correct UIDs/GIDs or can guarantee that they have not
```
ThomasWaldmann: s/for/about/?

"... or you can ..."

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,92 @@
...
+changed, a simple full restore can be achieved with::
+
+    borg export-tar /REPOSITORY::archive | ssh root@TARGET_IP 'tar xf -'
+
+(let Borg export the archive in tar format, upload it to the client and directly
+extract it in root file system). Again, don't forget the ``/mnt/sshfs`` prefix
+if used.
```
ThomasWaldmann: can we just get rid of that prefix stuff so we don't always need to show 2 alternative ways?

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: "lazy unmount" - just took it from the recipe. I assume if you use -z the command immediately returns instead of waiting for sync. But right - we don't need it here.

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: I'll have a look...

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: I'll add a sentence about the stand-alone binary used here for example purposes.

docs/deployment/pull-backup.rst
fantasya-pbem: Good points. I'll add a section discussing unsupported features when using sshfs and the security aspects.

docs/deployment/pull-backup.rst
fantasya-pbem: Would ``mnt/ssfs`` work here? I'll interprete it as "extract the contents of mnt/sshfs", am I getting it right?

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: OK

docs/deployment/pull-backup.rst
fantasya-pbem: What about adding a short note mentioning the path thing, but in the examples use the cd way?

docs/deployment/pull-backup.rst
fantasya-pbem: I did not. Just made some experiments with extract and mount & path. The full path is always extracted, so it seems always better to change working directory before borg create.

docs/deployment/pull-backup.rst
fantasya-pbem: Yes, seems best to get rid of the prefix.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,108 @@
...
+In this mode, the backup client's file system is mounted via sshfs on the backup
+server. Pull mode is even possible if the SSH connection must be established by
+the client via a remote tunnel.
+
+There are some restrictions caused by sshfs. Unless you define UID and GID
+mappings when mounting via sshfs, owners and groups of the mounted file system
+will probably change, and you may not have access to those files if BorgBackup
+is not run with root privileges.
+
+[ADD MORE RESTRICTIONS HERE: filename encoding, xattrs, ACLs, bsdflags...]
```
fantasya-pbem: I could not find useful information about restrictions, do you have reliable info? A sentence like "There may be more restictions: ..." seems not appropriate to me.

ThomasWaldmann: Depending on the filesystem used (sshfs here, but could be anything), you need to read about its limitations in its own docs, that is out of scope of borg docs.

ThomasWaldmann: So, the uid/username gid/groupname mapping issue is in general worth noting.
For other issues I'ld just put some sentence like:

Please make sure the filesystem you use implements all the features you need (e.g. check if it implements xattrs and ACLs, if bsdflags work, if long non-ascii filenames work).

fantasya-pbem: OK. My thoughts were "this recipe uses sshfs explicitly, so I should note what restrictions it brings". I haven't found any sshfs-specific info about restrictions yet. I started looking into FUSE docs (libfuse), but there is also nothing mentioned. I will check the FUSE kernel module docs, too. If I cannot find any useful info, I will insert your proposed sentence.

fantasya-pbem: Did some research, sshfs uses SFTP protocol, which is in IETF draft status. The definition generally does support xattrs and ACLs, but implementations may not have support. So I'll add some explanations regarding this.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+installed on the client.
+
+::
+
+    # Mount client root file system.
+    mkdir /tmp/sshfs
+    sshfs root@host:/ /tmp/sshfs
+    # Mount BorgBackup repository inside it.
+    mkdir /tmp/sshfs/borgrepo
+    mount --bind /REPOSITORY /tmp/sshfs/borgrepo
```
ThomasWaldmann: /path/to/repo?

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+::
+
+    borg extract /borgrepo::archive
+
+to restore the whole archive. Finally we need to exit chroot and unmount all the
+stuff:
+
+::
+
+    exit #ext chroot
```
ThomasWaldmann: typo.
also put a space after the `#`.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+
+The sshfs is a FUSE file system and uses the SFTP protocol, so there may be also
+other unsupported features that the actual implementations of sshfs, libfuse and
+sftp on the backup server do not support, like file name encodings, ACLs, xattrs
+or bsdflags. So there is no guarantee that you are able to restore a system
+completely in every aspect from such a backup.
+
+To mount the client's root file system you will need root access to the client.
+This contradicts to the usual threat model of BorgBackup, where clients don't
+need to trust the backup server can encrypt their data. In pull mode the server
```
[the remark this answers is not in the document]
fantasya-pbem: OK

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+(when logged in as root) could cause unlimited damage to the client. Therefore,
+pull mode should be used only from servers you own or trust yourself!
```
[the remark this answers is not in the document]
fantasya-pbem: OK

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+
+    # Mount client root file system.
+    mkdir /mnt/sshfs
+    sshfs root@host:/ /mnt/sshfs
+    # Change into mount dir and back up.
+    cd /mnt/sshfs
+    borg create /borg/to/repo::archive
+    # Unmount client.
+    cd ~
+    fusermount -u /mnt/sshfs
```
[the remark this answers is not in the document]
fantasya-pbem: OK

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Restoring a full backup
+-----------------------
+
+In case of a restore there is no direct way for borg to know the correct UID:GID
+of files, because it uses the ``/etc/passwd`` and ``/etc/group`` of the backup
+server. Using the archived UIDs/GIDs might lead to wrong results, because after
```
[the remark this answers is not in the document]
fantasya-pbem: Let's assume I have backed up ``/home/fantasya``, UID is 1000. After a crash I reinstalled, but with UID 1001. So restoring would set my homedir with UID 1000 - bad. This will be the case with --numeric-owner. But without --numeric-owner? I read that Borg also saves owner name. Would ``extract`` try to set owner to ``fantasya`` if it is available?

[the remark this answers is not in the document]
fantasya-pbem: This question could be a new FAQ: "Does Borg restore owner/group by ID or by name?"
I assume Borg first checks if a name exists in /etc/passwd or /etc/group, then sets "chown <name> <file>" or so. If not found, then sets "chown <uid> <file>". Is that correct?

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,116 @@
...
+
+::
+
+    # Mount client root file system.
+    mkdir /mnt/sshfs
+    sshfs root@host:/ /mnt/sshfs
+    # Change into mount dir and back up.
+    cd /mnt/sshfs
+    borg create /path/to/repo::archive .
+    # Unmount client.
```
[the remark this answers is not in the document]
ThomasWaldmann: we could also consider removing all these comment lines as it is rather obvious what is happening here.

the only "trick" that is worth noting is the use of a relative path `.` from within the mountpoint to avoid getting unwanted base dirs into the archived paths.

docs/deployment/pull-backup.rst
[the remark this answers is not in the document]
ThomasWaldmann: my comment was based on the idea of a full system backup and a full restore.
in that case, everything will be consistent if one just restores the numeric uids/gids as present in the archive.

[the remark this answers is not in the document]
ThomasWaldmann: even in the other case (not doing a full backup/restore), I guess, at backup time, one wants to avoid looking up user/group names on the backup server for the uids/gids seen on the client files - there's a quite big chance of mismatches.

at restore time, it depends.
- full backup / full restore will be automatically consistent when using only the numeric ids.
- if the original /etc/passwd and groups is lost and a new different one is effective now, one of course needs to use name-based mapping and all user/group names need to be at least present.

another problem is you can't mix these. if you did not backup the names, you can't use them at restore time.

i'm a fan of full backups. :-)

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,116 @@
...
+-------------------
+
+If you don't care about the correct UIDs/GIDs, or you can guarantee that they
+have not changed, a simple full restore can be achieved with
+
+::
+
+    borg export-tar /path/to/repo::archive | ssh root@host 'tar xf -'
+
+(let Borg export the archive in tar format, upload it to the client and directly
```
ThomasWaldmann: s/upload it/stream it/

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,116 @@
...
+extract it in root file system).
```
ThomasWaldmann: guess it would be extracted into root's home directory as that would be the cwd when invoking tar?

fantasya-pbem: Right, it should be / - I'll fix that.

docs/deployment/pull-backup.rst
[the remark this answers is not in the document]
fantasya-pbem: I think you are right, but I'm a bit confused now. When we do a full restore with --numeric-owner, then why do we have to create a chroot? It should be sufficient to mount to /mnt/sshfs, then cd /mnt/sshfs, then borg extract. Am I missing something?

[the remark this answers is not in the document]
fantasya-pbem: Thank you. I have tried to add this explanations and made a separation between partial and full restores. Also added a note for the ``export-tar`` variant (loss of features).

docs/deployment/pull-backup.rst
```diff
@@ -89,16 +85,19 @@ repository in ``/borgrepo``, and can run a simple
...
-to restore the whole archive. Finally we need to exit chroot and unmount all the
-stuff:
+to restore the whole archive. Finally we need to exit chroot, unmount all the
+stuff and clean up:
 
 ::
 
     exit # exit chroot
     cd /tmp/sshfs
+    rm usr/local/bin/borg
```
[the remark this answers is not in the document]
fantasya-pbem: fixed

docs/deployment/pull-backup.rst
```diff
@@ -48,19 +46,26 @@ the mount point prefix, e.g. /mnt/sshfs/bin/bash instead of /bin/bash).
...
 
-In case of a restore there is no direct way for borg to know the correct UID:GID
-of files, because it uses the ``/etc/passwd`` and ``/etc/group`` of the backup
-server. Using the archived UIDs/GIDs might lead to wrong results, because after
-a reinstall the name-to-ID mapping might have changed.
+Partial restore
+~~~~~~~~~~~~~~~
 
+In case of a partial restore, using the archived UIDs/GIDs might lead to wrong
+results, because after a reinstall the name-to-ID mapping might have changed.
```
ThomasWaldmann: ... wrong results if the name-to-ID mapping on the target system has changed compared to backup time (might be the case e.g. for a fresh OS install).

fantasya-pbem:
I have changed the "Alternative full restore" heading to make it clear that export-tar is not a real "alternative" but a lossy method.

docs/deployment/pull-backup.rst
```diff
@@ -78,15 +83,16 @@ installed on the client.
...
+files to the actual IDs on the client, and we can run
 
 ::
 
-    borg extract /borgrepo::archive
+    borg extract /borgrepo::archive PATH
 
-to restore the whole archive. Finally we need to exit chroot, unmount all the
-stuff and clean up:
+to partially restore whatever we like. Finally we need to exit chroot, unmount
```
ThomasWaldmann: "Finally, "?

(not sure)

docs/deployment/pull-backup.rst
```diff
@@ -99,17 +105,33 @@ stuff and clean up:
     umount /tmp/sshfs
     rmdir /tmp/sshfs
 
-Thanks to secuser on IRC for this howto.
+Thanks to secuser on IRC for this how-to!
+
+Full restore
+~~~~~~~~~~~~
+
+In a full restore differences between backup and client do not matter, so there
```
ThomasWaldmann: Don't say what is not the case, say what is the case, like:

When doing a full restore, you restore all files (including the ones containing the ID-to-name mapping, /etc/passwd and /etc/group). So everything will be consistent automatically if you restore the numeric IDs stored in the archive.

docs/deployment/pull-backup.rst
```diff
@@ -0,0 +1,137 @@
...
+In this approach the client file system is simply mounted and then backed up.
+Note that the backup is created from within the mount point so that all files
+in the archive have their original paths (otherwise they would be backed up with
+the mount point prefix, e.g. /mnt/sshfs/bin/bash instead of /bin/bash).
+
+::
+
+    sshfs root@host:/ /mnt/sshfs
+    cd /mnt/sshfs
+    borg create /path/to/repo::archive . # note the dot!
```
ThomasWaldmann: this is still storing wrong user/group names into the archive.

either: do not store names (just numeric ids)

or: work in a way so the id-to-name mapping of the backupped system is active.

fantasya-pbem: You're right, that's a point I didn't consider yet. So again a chroot might be the best solution here? Only storing numeric IDs would make the partial restore with mapping impossible.

ThomasWaldmann: yeah, guess it is chroot time again.

ThomasWaldmann:
Looks good to me.

Can you practically try it? backup, restore, check results - on a backup system with mismatching uids.

fantasya-pbem:
OK. I will create a new PR for this final version after I've tested it (did you notice that my last commit message refers to the wrong issue? ;-) Until then I leave this PR open.

Issue #4591, opened 2019-05-25
The .borg-passphrase created by following the FAQ can be hard to use.


ryanjaeb:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE / SUGGESTION

#### Your borg version (borg -V).

1.1.9

#### Operating system (distribution) and version.

CentOS Linux release 7.6.1810 (Core)

## Describe the problem you're observing.

The `.borg-passphrase` file created by following the [FAQ](https://borgbackup.readthedocs.io/en/stable/faq.html#password-env) has line breaks every 76 characters because of the [base64](https://linux.die.net/man/1/base64) command:

> -w, --wrap=COLS
>    Wrap encoded lines after COLS character (default 76). Use 0 to disable line wrapping. 

I created a passphrase with the suggested command:

```sh
head -c 1024 /dev/urandom | base64 > ~/.borg-passphrase
chmod 400 ~/.borg-passphrase
```

This creates a passphrase that's difficult to enter without using the suggested environment variable:
```
BORG_PASSCOMMAND="cat $HOME/.borg-passphrase"
```

When I attempt to copy the passphrase in my `.borg-passphrase` file and enter it via the command line (when prompted), I can't get it to work.  There's some encoding or escaping going on somewhere that I don't know how to reproduce.

I think this could be an issue when restoring because the passphrase might have to be entered via the `BORG_PASSCOMMAND` to get it to work.  I had the opposite happen.  I initialized my repo with a passphrase that had no line breaks and it took me a while to figure out it wasn't working because I had line breaks in my `.borg-passphrase` file.  I was copy / pasting from my password manager.  I somehow got the line breaks stripped out (by accident) before adding the passphrase to my password manager.

I think it might be better to suggest creating the passphrase file without line breaks:

```sh
head -c 1024 /dev/urandom | base64 --wrap=0 > ~/.borg-passphrase
```

IMO, the above is easier to store in a password manager and enter via the command line prompt.

ThomasWaldmann:
Also the 1024 random bytes is a bit much. Guess we don't need more than 256 bits (AES key length) or an attacker could just brute force the key directly rather than the passphrase,

So, how about `head -c 32 /dev/urandom | base64 > ~/.borg-passphrase`?

ryanjaeb:
@ThomasWaldmann That seems reasonable to me.

ThomasWaldmann:
[...]

PR #4787, opened 2019-10-18
[DOC] #4591 - Improve password FAQ

fantasya-pbem:
Improve password FAQ (decrease pw length, add -w 0 option to base64 to prevent line wrap).
I thought about adding a note regarding why the -w option may be needed, but finally didn't do that to keep the FAQ short.


ThomasWaldmann:
Thanks! Do 1.1-maint docs need the same change?

PR #4789, opened 2019-10-19
ran python setup.py build_man

ThomasWaldmann:
noticeable:
- does not badly malfunction any more
- it updates now the `borgfs.1` man page (good)
- it updates now the `borg.1` man page (good)

### DIFF
--- a/docs/man/borg-benchmark-crud.1
+++ b/docs/man/borg-benchmark-crud.1
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH BORG-BENCHMARK-CRUD 1 "2019-09-06" "" "borg backup tool"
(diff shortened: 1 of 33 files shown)


Gu1nness:
There was indeed a bug in the borg.1 man page creation.
This is fixed by this commit (on a dedicated branch in my fork): https://github.com/Gu1nness/borg/commit/5bdf998a2134e30e7af82eedf28a279077106a8b

ThomasWaldmann:
@Gu1nness ok, can you make a PR from that commit?

If the problem is only that (for otherwise unchanged content) the date is not regenerated as in all other generated manpages, it is less serious than I though (I thought it is not generated at all, never).

But still, guess we want to behaviour to be consistent - and if only to not have to wonder about it again.

PR #4790, opened 2019-10-19
[DOC] #4591 - Improve password FAQ

fantasya-pbem:
Backport from master: Decrease pw length, add -w 0 option to base64 to prevent line wrap).


ThomasWaldmann:
Thanks!

PR #4794, opened 2019-10-19
Doc/4532 faq retain original paths

### LINKED ISSUE #4532: borg backup and btrfs

fantasya-pbem:
fixes #4532

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -702,6 +702,23 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
 .. _pipeviewer: http://www.ivarch.com/programs/pv.shtml
 
 
+How can I avoid unwanted base directories getting stored into archives?
+-----------------------------------------------------------------------
+
+Possible use cases:
+
+* Another file system is mounted and you want to backup it with original paths.
+* You have created a BTRFS snapshot in a ``/.snapshots`` directory for backup.
+
+To achieve this, run ``borg_create`` within the mountpoint/snapshot directory:
+
+::
+
+    # Example: Some file system mounted in /mnt/rootfs.
+    cd /mnt/rootfs
+    borg create /path/to/repo::rootfs_backup .
+
+
 I am having troubles with some network/FUSE/special filesystem, why?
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -702,6 +702,23 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
 
+How can I avoid unwanted base directories getting stored into archives?
+-----------------------------------------------------------------------
+
+Possible use cases:
+
+- Another file system is mounted and you want to backup it with original paths.
+- You have created a BTRFS snapshot in a ``/.snapshots`` directory for backup.
+
+To achieve this, run ``borg_create`` within the mountpoint/snapshot directory:
```
ThomasWaldmann: Hmm, how is this rendered? Is this a link to the borg create docs?

Or should it rather be ```borg create``` (with blank, not underscore)?

docs/faq.rst
```diff
@@ -702,6 +702,23 @@ Now Borg will be bandwidth limited. Nice thing about pv is that you can change r
...
+
+::
+
+    # Example: Some file system mounted in /mnt/rootfs.
+    cd /mnt/rootfs
+    borg create /path/to/repo::rootfs_backup .
```
ThomasWaldmann: yeah, even better. we do not need to show the mount.

docs/faq.rst
fantasya-pbem: That should be a space, not underscore. Fixed.

ThomasWaldmann:
same for 1.1-maint?

PR #4798, opened 2019-10-20
Feature/4674 compact threshold

fantasya-pbem:


ThomasWaldmann:
tests fail, see travis output.

src/borg/archiver.py
```diff
@@ -1616,7 +1616,8 @@ def do_compact(self, args, repository):
         # see the comment in do_with_lock about why we do it like this:
         data = repository.get(Manifest.MANIFEST_ID)
         repository.put(Manifest.MANIFEST_ID, data)
-        repository.commit(compact=True, cleanup_commits=args.cleanup_commits)
+        percent = args.threshold / 100
```
ThomasWaldmann: percent is a misnomer / misleading.

guess args.threshold is a percent value (e.g. 5 would mean 5%).
but if you divide that by 100, it is not a percent value any more.

so, just `threshold`?

src/borg/remote.py
```diff
@@ -901,7 +901,7 @@ def check(self, repair=False, save_space=False, max_duration=0):
     @api(since=parse_version('1.0.0'),
          compact={'since': parse_version('1.2.0a0'), 'previously': True, 'dontcare': True},
          cleanup_commits={'since': parse_version('1.2.0a0'), 'previously': False, 'dontcare': True})
-    def commit(self, save_space=False, compact=True, cleanup_commits=False):
+    def commit(self, save_space=False, compact=True, threshold=0.1, cleanup_commits=False):
```
ThomasWaldmann: you need to extend the @api decoration in line 901+ also.

check the decorator code for the meaning of the misc. values.

src/borg/remote.py
```diff
@@ -900,6 +900,7 @@ def check(self, repair=False, save_space=False, max_duration=0):
 
     @api(since=parse_version('1.0.0'),
          compact={'since': parse_version('1.2.0a0'), 'previously': True, 'dontcare': True},
+         threshold={'since': parse_version('1.2.0a0'), 'previously': 0, 'dontcare': True},
```
ThomasWaldmann: I guess "previously" should be 0.1 (the value resulting in the behaviour before introducing this parameter).

fantasya-pbem: I'm new to Python. I came quite far, but didn't yet find where the @api decorators are defined. Also didn't see why the test failed. If you point to where @api docs are/should be, I'll try to improve it.

ThomasWaldmann: search for "def api".

ThomasWaldmann: (but that is for a separate PR please)

ThomasWaldmann:
```
    def test_crash_before_compact(self):
        self.repository.put(H(0), b'data')
        self.repository.put(H(0), b'data2')
        # Simulate a crash before compact
        with patch.object(Repository, 'compact_segments') as compact:
            self.repository.commit(compact=True)
>           compact.assert_called_once_with()

.tox/py35/lib/python3.5/site-packages/borg/testsuite/repository.py:788: 
```

You need to supply the `0.1` to `compact.assert_called_once_with()`.

src/borg/remote.py
fantasya-pbem: OK, tried to add doc in #4801

ThomasWaldmann:
thanks!

PR #4804, opened 2019-10-24
#1552 Deployment Pull Backup

fantasya-pbem:


docs/deployment/pull-backup.rst
```diff
@@ -11,6 +11,13 @@ the SSH connection must be established by the client via a remote tunnel. Other
 network file systems like NFS or SMB could be used as well, but SSHFS is very
 simple to set up and probably the most secure one.
 
+.. note::
+
+    There are user reports that SSHFS can be extremely slow over high latency
+    networks. Using SSHFS also means Borg does all the checksumming over the
+    network which is also very slow. You should benchmark your connection's
+    throughput first before deploying a huge setup.
```
ThomasWaldmann: hmm, this slightly misses the point.

the "checksumming" (chunking / hashing / lookup of the file contents) happens only if the files cache lookup considers the file as potentially changed (or new).

if that check works correctly (which might need fine tuning using --files-cache), it should work quite reliably. if a file is really changed, we need to read (chunk, ...) it anyway. if read speed in general is slow over sshfs, that is not a borg specific problem.

so, the reading (chunking, ...) is not the remarkable part that should be documented here, but that one needs to get the files cache working (what you already did below).

docs/deployment/pull-backup.rst
```diff
@@ -68,10 +75,13 @@ create the backup, retaining the original paths, excluding the repository:
...
 
-    borg create --exclude /borgrepo /borgrepo::archive /
+    borg create --exclude /borgrepo --files-cache ctime,size /borgrepo::archive /
 
-(For the sake of simplicity only ``/borgrepo`` is excluded here. You may want to
-set up an exclude file with additional files and folders to be excluded.)
+For the sake of simplicity only ``/borgrepo`` is excluded here. You may want to
+set up an exclude file with additional files and folders to be excluded. Also
+note that we have to modify Borg's file change detection behaviour – SSHFS does
+not support real inode numbers, so we have to supply the ``--files-cache``
```
ThomasWaldmann: it doesn't matter whether they are "real", but rather whether they are stable.

so, they are not stable?

ThomasWaldmann:
OK, thanks for writing this, guess this is merge ready?

Did you practically try this?

fantasya-pbem:
Yes, I tested it on two VirtualBox VMs, everythings works as expected.

ThomasWaldmann:
Merged! Thanks for your patience!

PR #4805, opened 2019-10-24
[DOC] #1552 - Add a pull backup / push restore how-to

fantasya-pbem:
Backport from master.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -13,3 +13,4 @@ This chapter details deployment strategies for the following scenarios.
    deployment/hosting-repositories
    deployment/automated-local
    deployment/image-backup
+   deployment/pull-backup
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -1,0 +1,173 @@
+.. include:: ../global.rst.inc
+.. highlight:: none
+
+Backing up in pull mode
+=======================
+
+Assuming you have a pull backup system set up with borg, where a backup server
+pulls the data from the target via SSHFS. In this mode, the backup client's file
+system is mounted remotely on the backup server. Pull mode is even possible if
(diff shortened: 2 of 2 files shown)


felinira:
Please add a note that SSHFS can be extremely slow over high latency network (and even moderate latency network). This also means borg does all the checksumming over the network which is also very slow. I tried a setup like this and it took multiple days to backup a server over 100 Mbit LAN (~40GB) and over a week to backup 200GB over internet (only 16Mbit DSL but it should only take 1-2 days). The network link saturation was very low most of the time.

This is definitely a working solution for pull style backup but it does not seem very feasible to do over the internet.

fantasya-pbem:
Thank your for this information.

> SSHFS can be extremely slow over high latency network

Can you give a source where this is explained?

ThomasWaldmann:
That reminds me of something that needs to be tried and results documented in the "pull via sshfs" docs:

the borg files cache speeds up backup a lot if it can work correctly, by default this means for a file to be considered as unchanged, these conditions must be met:
- size of file unchanged (we can assume this is ok for sshfs)
- ctime of file unchanged (does sshfs support this?)
- inode number of file unchanged and long term stable (does sshfs support this?)

If that is not fully supported, tweaking via `--files-cache=...` is possible.

fantasya-pbem:
IETF has a draft of SSH file transfer: https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13
Afaik there is no better doc, apart from looking into the SSH code directly.
- ctime is available
- inode number: ? (not mentioned in doc)

If you can specify the procedure that should be tried, I will do it and document the results.

fantasya-pbem:
See https://borgbackup.readthedocs.io/en/stable/usage/create.html for ``--files-cache``:
In SSHFS ``--files-cache ctime,size`` seems to be the right setting.

PR #4815, opened 2019-11-02
Borg recreate timestamp is a no op

### LINKED ISSUE #4745: borg recreate --timestamp is a no-op

Yes

BUG (either docs or borg-recreate)

```
$ borg -V
borg 1.1.10
```

4.19.0-5-amd64, Debian testing

tmpfs

few MB

```
$ borg recreate --timestamp '2019-01-01T00:00:00' --comment 're-created for fun' testrepo::archive1
```

Yes, 100%. Assuming you have some stuff to play with in ``./testcontent``.

```console
$ borg init -e none testrepo
$ borg create testrepo::archive1 testcontent
$ borg info testrepo::archive1
Archive name: archive1
Archive fingerprint: fc4f5feb66c314073d124acd10c856b153f077e5865fd655b23423d21b6227ac
Comment:
Hostname: sinistra
Username: horazont
Time (start): Sun, 2019-09-01 17:02:18
Time (end): Sun, 2019-09-01 17:02:18
Duration: 0.03 seconds
Number of files: 2
Command line: /usr/bin/borg create testrepo::archive1 testcontent
Utilization of maximum supported archive size: 0%
This archive:                3.75 MB              3.76 MB              3.76 MB
All archives:                3.75 MB              3.76 MB              3.76 MB

Chunk index:                       5                    5
$ borg recreate --timestamp '2019-01-01T00:00:00' --comment 're-created for fun' 
$ borg info testrepo::archive1
Archive name: archive1
Archive fingerprint: f317eeb7a04e2fa1f06b3d0e3aa919042cb6730fc1d913d348f08e768d800314
Comment: re-created for fun
Hostname: sinistra
Username: horazont
Time (start): Sun, 2019-09-01 17:02:18
Time (end): Sun, 2019-09-01 17:02:18
Duration: 0.03 seconds
Number of files: 2
Command line: /usr/bin/borg create testrepo::archive1 testcontent
Utilization of maximum supported archive size: 0%
This archive:                3.75 MB              3.76 MB              3.76 MB
All archives:                3.75 MB              3.76 MB              3.76 MB

Chunk index:                       5                    5
```

I expect the timestamp in the second borg info call to be changed after borg recreate, but that is not happening.

Bonus:
```console
$ borg recreate --timestamp '2019-01-01T00:00:00' testrepo::archive1
Nothing to do. Archive was not processed.
Specify at least one pattern, PATH, --comment, re-compression or re-chunking option.
```

It appears the code knows about this, too.

I noticed that args.timestamp is not used at all here: https://github.com/borgbackup/borg/blob/master/src/borg/archiver.py#L1538

Gu1nness:
I also added a test to ensure that no regression happens on this feature.

Fixes #4745.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1977,7 +1977,6 @@ class ArchiveRecreater:
             return
         if comment is None:
             comment = archive.metadata.get('comment', '')
-        duration = timedelta(seconds=time.monotonic() - target.start_monotonic)
 
         # Keep for the statistics if necessary
         if self.stats:
@@ -1986,21 +1985,22 @@ class ArchiveRecreater:
         if self.timestamp is None:
             end = archive.metadata.get('time_end') or archive.metadata.time
             start = archive.metadata.time
+            additional_metadata = {
+                'time': start,
+                'time_end': end,
+                'cmdline': archive.metadata.cmdline,
+                # but also remember recreate metadata:
+                'recreate_cmdline': sys.argv,
+            }
         else:
-            end = (self.timestamp + duration).strftime(ISO_FORMAT)
-            start = self.timestamp.strftime(ISO_FORMAT)
+            additional_metadata = {
+                'cmdline': archive.metadata.cmdline,
+                # but also remember recreate metadata:
+                'recreate_cmdline': sys.argv,
+            }
 
-        target.start = start
-        target.end = end
-
-        target.save(comment=comment, timestamp=self.timestamp, stats=target.stats, additional_metadata={
-            # keep some metadata as in original archive:
-            'time': start,
-            'time_end': end,
-            'cmdline': archive.metadata.cmdline,
-            # but also remember recreate metadata:
-            'recreate_cmdline': sys.argv,
-        })
+        target.save(comment=comment, timestamp=self.timestamp,
+                    stats=target.stats, additional_metadata=additional_metadata)
         if replace_original:
             archive.delete(Statistics(), progress=self.progress)
             target.rename(archive.name)
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1852,7 +1852,7 @@ class ArchiveRecreater:
                  exclude_caches=False, exclude_if_present=None, keep_exclude_tags=False,
                  chunker_params=None, compression=None, recompress=False, always_recompress=False,
                  dry_run=False, stats=False, progress=False, file_status_printer=None,
-                 checkpoint_interval=1800):
+                 timestamp=None, checkpoint_interval=1800):
         self.repository = repository
         self.key = key
         self.manifest = manifest
@@ -1872,6 +1872,7 @@ class ArchiveRecreater:
         self.compression = compression or CompressionSpec('none')
         self.seen_chunks = set()
 
+        self.timestamp = timestamp
         self.dry_run = dry_run
         self.stats = stats
         self.progress = progress
@@ -1976,10 +1977,27 @@ class ArchiveRecreater:
             return
         if comment is None:
             comment = archive.metadata.get('comment', '')
-        target.save(comment=comment, stats=target.stats, additional_metadata={
+        duration = timedelta(seconds=time.monotonic() - target.start_monotonic)
+
+        # Keep for the statistics if necessary
+        if self.stats:
+            _start = target.start
+
+        if self.timestamp is None:
+            end = archive.metadata.get('time_end') or archive.metadata.time
+            start = archive.metadata.time
+            logger.info(archive.metadata.time)
+        else:
+            end = (self.timestamp + duration).strftime(ISO_FORMAT)
+            start = self.timestamp.strftime(ISO_FORMAT)
+
+        target.start = start
+        target.end = end
+
+        target.save(comment=comment, timestamp=self.timestamp, stats=target.stats, additional_metadata={
             # keep some metadata as in original archive:
-            'time': archive.metadata.time,
-            'time_end': archive.metadata.get('time_end') or archive.metadata.time,
+            'time': start,
+            'time_end': end,
             'cmdline': archive.metadata.cmdline,
             # but also remember recreate metadata:
             'recreate_cmdline': sys.argv,
@@ -1988,6 +2006,7 @@ class ArchiveRecreater:
             archive.delete(Statistics(), progress=self.progress)
             target.rename(archive.name)
         if self.stats:
+            target.start = _start
             target.end = datetime.utcnow()
             log_multi(DASHES,
                       str(target),
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1986,7 +1986,6 @@ class ArchiveRecreater:
         if self.timestamp is None:
             end = archive.metadata.get('time_end') or archive.metadata.time
             start = archive.metadata.time
-            logger.info(archive.metadata.time)
         else:
             end = (self.timestamp + duration).strftime(ISO_FORMAT)
             start = self.timestamp.strftime(ISO_FORMAT)
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1983,11 +1983,9 @@ class ArchiveRecreater:
             _start = target.start
 
         if self.timestamp is None:
-            end = archive.metadata.get('time_end') or archive.metadata.time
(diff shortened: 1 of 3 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -2551,6 +2551,16 @@ def test_recreate_recompress(self):
...
 
+    def test_recreate_timestamp(self):
+        self.create_test_files()
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        archive = self.repository_location + '::test0'
+        self.cmd('create', archive, 'input')
+        self.cmd('recreate', '--timestamp', "1970-01-01T00:00:00", '--comment',
+                 'test', archive)
+        info = self.cmd('info', archive).splitlines()
+        assert any([re.search(r'Time \(start\).+ 1970-01-01', item) for item in info])
```
ThomasWaldmann: in case that timestamp is considered local time, please use something later to avoid an underflow in utc.

src/borg/testsuite/archiver.py
```diff
@@ -2561,6 +2571,7 @@ def test_recreate_dry_run(self):
         archives_after = self.cmd('list', self.repository_location + '::test')
         assert archives_after == archives_before
 
+
```
ThomasWaldmann: remove accidental change, please.

src/borg/archive.py
```diff
@@ -1976,10 +1977,27 @@ def save(self, archive, target, comment=None, replace_original=True):
...
+        duration = timedelta(seconds=time.monotonic() - target.start_monotonic)
+
+        # Keep for the statistics if necessary
+        if self.stats:
+            _start = target.start
+
+        if self.timestamp is None:
+            end = archive.metadata.get('time_end') or archive.metadata.time
+            start = archive.metadata.time
+            logger.info(archive.metadata.time)
```
ThomasWaldmann: guess you added this for debugging.

src/borg/archive.py
```diff
@@ -1988,6 +2006,7 @@ def save(self, archive, target, comment=None, replace_original=True):
             archive.delete(Statistics(), progress=self.progress)
             target.rename(archive.name)
         if self.stats:
+            target.start = _start
             target.end = datetime.utcnow()
```
ThomasWaldmann: so, if --timestamp is used, we'll have the desired timestamp as start value, but the current time as end value?

this might look like a backup taking a year, if the start time is a year in the past...

use case: having a set of old backups and re-creating them with borg.

ThomasWaldmann: guess there could be 2 solutions for this:
- use start + duration as end time (this might be a bit fake though as the duration comes from the recreate operation, not from the create operation)
- use end == start (indicating "we don't know the end time" somehow)

src/borg/testsuite/archiver.py
ThomasWaldmann: this is why the travis flake8 test is failing currently.

src/borg/archive.py
Gu1nness: Hum I forgot to remove the logger yes.
For what is left, actually no:
Since we need to keep the good metadata for the statistics, we want to have either the old metadata (L1987 and 1988). 

We need to keep `_start` (which is the date of creation of the new `Archive`) for the statistics: I guess we want to have the statistics of the current run, or are they supposed to be the one with the recreation date ?
If so indeed my code is wrong, and I can remove this `_start`.

src/borg/archive.py
Gu1nness: Oh, it highlighted an edge case I forgot indeed.
I think that we can give the statistics of the recreate command: how much time did it take to recreate the archive?

src/borg/testsuite/archiver.py
Gu1nness: Working on it to have a cleaner test.
Actually, the timestamp is considered localtime after some checks (I'm UTC+1 and I can see
`Time start: 1970-01-01 01:00:00`.
The code is nearly ready, tests are running locally to be sure.

src/borg/testsuite/archiver.py
Gu1nness: Sure, it is done.
My bad, I forgot to re-run this test in the end.

src/borg/testsuite/archiver.py
ThomasWaldmann: I'ld just get at least 24h away from time 0.

src/borg/archive.py
ThomasWaldmann: ok, guess i misunderstood it.

it is fine to have the recreate duration (start, end) for the --stats output as this refers to the current operation.

if no --timestamp is used, the created archive has the same start/end as it had before. so the duration (end-start) refers to the time needed for the original backup.

if --timestamp is used, the duration should follow the same semantics I guess or it will be inconsistent to the other case.
for the created archive, we want to have the given --timestamp in the metadata,

src/borg/testsuite/archiver.py
```diff
@@ -2551,6 +2551,16 @@ def test_recreate_recompress(self):
...
 
+    def test_recreate_timestamp(self):
+        self.create_test_files()
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        archive = self.repository_location + '::test0'
+        self.cmd('create', archive, 'input')
+        self.cmd('recreate', '--timestamp', "1970-01-02T00:00:00", '--comment',
+                 'test', archive)
+        info = self.cmd('info', archive).splitlines()
+        assert any([re.search(r'Time \(start\).+ 1970-01-02', item) for item in info])
```
ThomasWaldmann: can you check the end time also?

src/borg/archive.py
```diff
@@ -1976,10 +1977,26 @@ def save(self, archive, target, comment=None, replace_original=True):
...
+
+        if self.timestamp is None:
+            end = archive.metadata.get('time_end') or archive.metadata.time
+            start = archive.metadata.time
+        else:
+            end = (self.timestamp + duration).strftime(ISO_FORMAT)
+            start = self.timestamp.strftime(ISO_FORMAT)
+
+        target.start = start
+        target.end = end
```
ThomasWaldmann: `target.save` method always overwrites this.

src/borg/archive.py
```diff
@@ -1976,10 +1977,26 @@ def save(self, archive, target, comment=None, replace_original=True):
...
+
+        target.save(comment=comment, timestamp=self.timestamp, stats=target.stats, additional_metadata={
             # keep some metadata as in original archive:
```
ThomasWaldmann: looks like comment does not match the code any more now.

src/borg/archive.py
```diff
@@ -1976,10 +1977,26 @@ def save(self, archive, target, comment=None, replace_original=True):
...
+        # Keep for the statistics if necessary
+        if self.stats:
+            _start = target.start
+
+        if self.timestamp is None:
+            end = archive.metadata.get('time_end') or archive.metadata.time
+            start = archive.metadata.time
+        else:
+            end = (self.timestamp + duration).strftime(ISO_FORMAT)
+            start = self.timestamp.strftime(ISO_FORMAT)
```
ThomasWaldmann: looks like `target.save` method is doing just that when it was given a timestamp != None as parameter.

so guess we only need to prepare a different `additional_metadata` dict depending on timestamp being None or not:
- None: dict with start/end info of original archive (see that comment) to override what is normally computed
- not None: dict without start/end overrides

src/borg/archive.py
```diff
@@ -1976,18 +1977,35 @@ def save(self, archive, target, comment=None, replace_original=True):
...
+        # Keep for the statistics if necessary
+        if self.stats:
+            _start = target.start
+
+        if self.timestamp is None:
+            end = archive.metadata.get('time_end') or archive.metadata.time
+            start = archive.metadata.time
+            additional_metadata = {
+                'time': start,
+                'time_end': end,
```
ThomasWaldmann: as start and end are only used here, maybe directly assign without using these variables, like in the old code?

ThomasWaldmann:
Thanks for working on this! Does 1.1-maint branch need the same change?

Gu1nness:
Yes it needs the same changes. Do you want me to do it ?

ThomasWaldmann:
yes, please!

PR #4818, opened 2019-11-03
Add a try catch when formatting the info string.

### LINKED ISSUE #4793: traceback when trying to delete non-existent archive
From the borgbackup mailing list:

I am testing out borg, and accidentally did:

```
borg create borgrepo::test4 ~
```

after 3 previous test archives were created using an --exclude-if-present .nobackup option.

When I noticed the missing --exclude-if-present in the above create (because it was taking too
long), I ^C'ed it, then tried:

```
borg delete borgrepo::test4
```

and got this failure (<...> obfuscations added):

```
Local Exception
  File "/usr/lib/python3.7/site-packages/borg/archiver.py", line 4501, in main
  File "/usr/lib/python3.7/site-packages/borg/archiver.py", line 4433, in run
    return set_ec(func(args))
  File "/usr/lib/python3.7/site-packages/borg/archiver.py", line 166, in wrapper
  File "/usr/lib/python3.7/site-packages/borg/archiver.py", line 1257, in do_delete
    return self._delete_archives(args, repository)
  File "/usr/lib/python3.7/site-packages/borg/archiver.py", line 1304, in _delete_archives
    logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
  File "/usr/lib/python3.7/site-packages/borg/helpers.py", line 244, in __getitem__
    raise KeyError
KeyError

Platform: Linux <...> 5.3.6-1-MANJARO #1 SMP PREEMPT Sat Oct 12 09:30:05
UTC 2019 x86_64 Linux: arch Manjaro Linux
Borg: 1.1.10  Python: CPython 3.7.4 msgpack: 0.5.6
PID: 10547  CWD: <...>
sys.argv: ['/usr/bin/borg', 'delete', 'borgrepo::test4']
SSH_ORIGINAL_COMMAND: None
```

Is this borgrepo now fried?  If not, how can I recover it?

Gu1nness:
This way, if the archive does not exist, we print a pretty message and
abort.

Fixes #4793.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1145,12 +1145,16 @@ class Archiver:
         stats = Statistics()
         with Cache(repository, key, manifest, progress=args.progress, lock_wait=self.lock_wait) as cache:
             for i, archive_name in enumerate(archive_names, 1):
-                msg = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
-                logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
-                if not dry_run:
-                    archive = Archive(repository, key, manifest, archive_name, cache=cache,
-                                      consider_part_files=args.consider_part_files)
-                    archive.delete(stats, progress=args.progress, forced=args.forced)
+                try:
+                    msg = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
+                    logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
+                    if not dry_run:
+                        archive = Archive(repository, key, manifest, archive_name, cache=cache,
+                                          consider_part_files=args.consider_part_files)
+                        archive.delete(stats, progress=args.progress, forced=args.forced)
+                except KeyError:
+                    logger.warning('Archive {} not found ({}/{}).'.format(archive_name, i, len(archive_names)))
+
             if not dry_run:
                 manifest.write()
                 repository.commit(compact=False, save_space=args.save_space)
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1153,9 +1153,9 @@ class Archiver:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
                     logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
-                    archive = Archive(repository, key, manifest, archive_name, cache=cache,
-                                      consider_part_files=args.consider_part_files)
                     if not dry_run:
+                        archive = Archive(repository, key, manifest, archive_name, cache=cache,
+                                          consider_part_files=args.consider_part_files)
                         archive.delete(stats, progress=args.progress, forced=args.forced)
 
             if not dry_run:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -45,7 +45,7 @@ try:
     from .crypto.key import key_creator, key_argument_names, tam_required_file, tam_required, RepoKey, PassphraseKey
     from .crypto.keymanager import KeyManager
     from .helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_ERROR
-    from .helpers import Error, NoManifestError, set_ec
+    from .helpers import Error, NoManifestError, NoArchiveError, set_ec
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
There's a ton of code being executed within this exception handler and KeyError is rather generic.

src/borg/archiver.py
```diff
@@ -1145,12 +1145,16 @@ def _delete_archives(self, args, repository):
...
-                logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
-                if not dry_run:
-                    archive = Archive(repository, key, manifest, archive_name, cache=cache,
-                                      consider_part_files=args.consider_part_files)
-                    archive.delete(stats, progress=args.progress, forced=args.forced)
+                try:
+                    msg = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
+                    logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
+                    if not dry_run:
+                        archive = Archive(repository, key, manifest, archive_name, cache=cache,
```
ThomasWaldmann: if we want similar behavior as do_list for a unknown archive, we could create the Archive instance first, like:
```
arch = Archive(...)  # if it does not exist, that will raise an Exception leading to a clean error message (as in do_list)
msg = ...
logger.info(...)
if not dry_run:
    arch.delete(...)

Comment (Gu1nness) [reply src/borg/archiver.py]: True it is.
Sorry for being away this week, I was sick and away from almost everything :/
Patch coming soon.

File: src/borg/helpers/manifest.py [hunk 2]
```diff
@@ -53,7 +56,7 @@ def __getitem__(self, name):
         _name = safe_encode(name)
         values = self._archives.get(_name)
         if values is None:
-            raise KeyError
+            raise NoArchiveError
```
ThomasWaldmann: this is a good idea to support wide-scope catching, but not needed here, see above.

src/borg/archiver.py
```diff
@@ -1145,12 +1145,16 @@ def _delete_archives(self, args, repository):
...
+                archive = Archive(repository, key, manifest, archive_name, cache=cache,
+                                  consider_part_files=args.consider_part_files)
+                try:
+                    msg = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
+                    logger.info(msg.format(format_archive(manifest.archives[archive_name]), i, len(archive_names)))
+                    if not dry_run:
+                        archive.delete(stats, progress=args.progress, forced=args.forced)
+                except NoArchiveError:
+                    logger.warning('Archive {} not found ({}/{}).'.format(archive_name, i, len(archive_names)))
+
```
ThomasWaldmann: well, this is now a mix up of my idea how to solve the "too wide scope for just catching KeyError" and your idea.

mine was to rely on Archive class bailing out when the archive name is not known.

yours was to make a more specific subclass of KeyError, which would somehow solve the problem of wide-scope catching a KeyError that could be caused by a lot of code (while not killing the loop).

but in the end, we do not even need the more specific subclass, but we need narrow-scope KeyError catching:

```
msg_delete = ...
msg_not_found = ...
for ...
    try:
        archive_info = manifest.archives[archive_name]
    except KeyError:  # note the very close scope, KeyError can only happen for what we really mean
        logger.warning(msg_not_found...)
    else:
        logger.info(msg_delete...)
        ...
```

src/borg/archiver.py
```diff
@@ -1144,13 +1144,20 @@ def _delete_archives(self, args, repository):
...
+                    archive_info = manifest.archives[archive_name]
+                except KeyError:
+                    logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
+                else:
+                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
                     archive = Archive(repository, key, manifest, archive_name, cache=cache,
                                       consider_part_files=args.consider_part_files)
-                    archive.delete(stats, progress=args.progress, forced=args.forced)
+                    if not dry_run:
+                        archive.delete(stats, progress=args.progress, forced=args.forced)
```
ThomasWaldmann: archive is only used within the `if`, right?

so move line 1156 directly above the delete.

ThomasWaldmann:
Thanks for implementing this, it's merged into master now. Does 1.1-maint branch need the same change?

Gu1nness:
It needs the same change. Should I do it for you ?

ThomasWaldmann:
yes, please!

Issue #1021, opened 2016-05-06
chunker performance


ThomasWaldmann:
the chunker seems to perform at 400-500MB/s on a in-memory file (BytesIO or on ramdisk), including reading, chunk buffer management and the chunking itself.
 
just reading (fd.read()) the file performs at ~6000MB/s.

can we make the chunker significantly faster without making the code more complex?

enkore:
Force-inlining buzhash and buzhash_update seemed to make a difference when I tried it. If you have a good, deterministic benchmark it's worth a shot.

ThomasWaldmann:
Hmm, i changed `static` to `static inline` for buzhash/buzhash_update, but did not see a difference in speed.

enkore:
iirc `__attribute__((always_inline))`

ThomasWaldmann:
gcc does inline small functions with -O2 (as it is run by default by cythonize call).

ThomasWaldmann:
note: replacing .read() by .readinto() (and thus avoiding the memcpy) in the python-file part of the chunker doesn't give a noticeable speedup.

ThomasWaldmann:
btw, i had a look into the assembler code and gcc -O2 seems clever enough to translate that "BARREL_SHIFT" code into a single "rol" x86 instruction. \o/

rakajack:
Replacing lenmod in a buzhash_update with a constant (because window is always 4095) gives 30 mb/s increase in chunker speed.

ThomasWaldmann:
@ah, good find! it could be also computed outside the loop and passed in instead the length.

rakajack:
A chunk have a minimum size (min_size). Are we skipping (min_size - window_size) bytes after finding a boundary?

ThomasWaldmann:
No. Also, we are not yet skipping files that are too small.

ThomasWaldmann:
See PR #1077.

ThomasWaldmann:
new chunker
-----------------
```
    length   count      dt throughput chunks_count
1000000000       1   2.271 440.392    120
 100000000      10   2.310 432.956     12
  10000000     100   2.200 454.530      2
   1000000    1000   1.100 908.739      1
    100000   10000   0.104 9609.142     1
     10000  100000   0.225 4434.996     1
      1000 1000000   1.750 571.296      1
       100 1000000   1.732 57.728       1
        10 1000000   1.719 5.816        1
         1 1000000   1.821 0.549        1
```

master chunker
--------------------
```
[...]

PR #4833, opened 2019-11-09
chunker speedup

ncleaton:
I'm a bit late to the https://github.com/borgbackup/borg/issues/1021 party, buy here's a chunker speedup by rearranging the loop a bit. Testing throughput on my slow old laptop:

Before:

```(borg-env) lenovo_0$ borg-env/bin/python3  ~/borgbench
    length   count      dt throughput chunks_count
1000000000       1 5.442 183.741    120
 100000000      10 5.531 180.795     12
  10000000     100 5.225 191.399      2
   1000000    1000 2.764 361.778      1
    100000   10000 0.172 5826.037      1
     10000  100000 0.993 1006.679      1
      1000 1000000 8.999 111.122      1
       100 1000000 8.776 11.394      1
        10 1000000 8.763 1.141      1
         1 1000000 8.768 0.114      1
```

After:
```(borg-env) lenovo_0$ borg-env/bin/python3  ~/borgbench
    length   count      dt throughput chunks_count
1000000000       1 3.891 256.990    120
 100000000      10 3.989 250.687     12
  10000000     100 3.736 267.687      2
   1000000    1000 2.022 494.470      1
    100000   10000 0.172 5827.721      1
     10000  100000 0.984 1016.437      1
      1000 1000000 8.931 111.970      1
       100 1000000 8.682 11.518      1
        10 1000000 8.649 1.156      1
         1 1000000 8.677 0.115      1
```

### DIFF
--- a/src/borg/_chunker.c
+++ b/src/borg/_chunker.c
@@ -247,7 +247,7 @@ static PyObject *
 chunker_process(Chunker *c)
 {
     uint32_t sum, chunk_mask = c->chunk_mask;
-    size_t n = 0, old_last, min_size = c->min_size, window_size = c->window_size;
+    size_t n, old_last, min_size = c->min_size, window_size = c->window_size;
 
     if(c->done) {
         if(c->bytes_read == c->bytes_yielded)
@@ -282,15 +282,19 @@ chunker_process(Chunker *c)
      */
     c->position += min_size;
     c->remaining -= min_size;
-    n += min_size;
     sum = buzhash(c->data + c->position, window_size, c->table);
     while(c->remaining > c->window_size && (sum & chunk_mask)) {
-        sum = buzhash_update(sum, c->data[c->position],
-                             c->data[c->position + window_size],
-                             window_size, c->table);
-        c->position++;
-        c->remaining--;
-        n++;
+        uint8_t *p = c->data + c->position;
+        uint8_t *stop_at = p + c->remaining - window_size;
+        size_t did_bytes;
+        while (p < stop_at && (sum & chunk_mask)) {
+            sum = buzhash_update(sum, p[0], p[window_size],
+                                 window_size, c->table);
+            p++;
+        }
+        did_bytes = p - (c->data + c->position);
+        c->position += did_bytes;
+        c->remaining -= did_bytes;
         if(c->remaining <= window_size) {
             if(!chunker_fill(c)) {
                 return NULL;
--- a/src/borg/testsuite/chunker_slow.py
+++ b/src/borg/testsuite/chunker_slow.py
@@ -1,0 +1,39 @@
+from io import BytesIO
+from binascii import unhexlify
+
+from ..chunker import Chunker
+from ..crypto.low_level import blake2b_256
+from ..constants import *  # NOQA
+from . import BaseTestCase
+
+
+class ChunkerRegressionTestCase(BaseTestCase):
+
+    def test_chunkpoints_unchanged(self):
+        def twist(size):
+            x = 1
+            a = bytearray(size)
+            for i in range(size):
+                x = (x * 1103515245 + 12345) & 0x7FFFFFFF
+                a[i] = x & 0xFF
+            return a
+
+        data = twist(100000)
+
+        runs = []
+        for winsize in (65, 129, HASH_WINDOW_SIZE, 7351):
+            for minexp in (4, 6, 7, 11, 12):
+                for maxexp in (15, 17):
+                    if minexp >= maxexp:
+                        continue
+                    for maskbits in (4, 7, 10, 12):
+                        for seed in (1849058162, 1234567653):
+                            fh = BytesIO(data)
+                            chunker = Chunker(seed, minexp, maxexp, maskbits, winsize)
+                            chunks = [blake2b_256(b'', c) for c in chunker.chunkify(fh, -1)]
+                            runs.append(blake2b_256(b'', b''.join(chunks)))
+
+        # The "correct" hash below matches the existing chunker behavior.
+        # Future chunker optimisations must not change this, or existing repos will bloat.
+        overall_hash = blake2b_256(b'', b''.join(runs))
+        self.assert_equal(overall_hash, unhexlify("b559b0ac8df8daaa221201d018815114241ea5c6609d98913cd2246a702af4e3"))


src/borg/_chunker.c
```diff
@@ -282,15 +282,19 @@ chunker_process(Chunker *c)
...
-        sum = buzhash_update(sum, c->data[c->position],
-                             c->data[c->position + window_size],
-                             window_size, c->table);
-        c->position++;
-        c->remaining--;
-        n++;
+        uint8_t *p = c->data + c->position;
+        uint8_t *stop_at = p + c->remaining - window_size;
+        size_t did_bytes;
+        while (p < stop_at && (sum & chunk_mask)) {
```
ThomasWaldmann: that looks a bit strange - why do we have 2 nested while-loops now?

ThomasWaldmann:
BTW, an important test after finishing the optimization work would be to make sure that the new code cuts precisely the same chunks as the old code. This is important for deduplication in existing repos.

src/borg/_chunker.c
ncleaton: The inner loop is over every byte of input, the outer loop iterates every time we need to read more inpit. This allows some housekeeping to be done once per chunker_fill rather than once per input byte.

ncleaton:
> BTW, an important test after finishing the optimization work would be to make sure that the new code cuts precisely the same chunks as the old code. This is important for deduplication in existing repos.

Good point. Should a script to perform this test go into the borg repo, either as part of the test suite (perhaps normally disabled) or as something in the scripts directory ?

ThomasWaldmann:
Yes, a unit test would be fine.

ncleaton:
> Yes, a unit test would be fine.

Added. I've also run the same test code over each of the files under /boot on my system with old and new chunker code, and the results are identical.

src/borg/testsuite/chunker.py
```diff
@@ -59,3 +60,30 @@ def read(self, nbytes):
...
+                for maxexp in (15, 17, CHUNK_MAX_EXP):
+                    if minexp >= maxexp:
+                        continue
+                    for maskbits in (4, 7, 10, 12, HASH_MASK_BITS):
+                        for seed in (1849058162, 1234567653):
+                            fh = BytesIO(data)
+                            chunker = Chunker(seed, minexp, maxexp, maskbits, winsize)
+                            for c in chunker.chunkify(fh, -1):
+                                h.update(sha256(c).hexdigest().encode())
+                            h.update(b"\n")
```
ThomasWaldmann: for what is that needed?

src/borg/testsuite/chunker.py
```diff
@@ -59,3 +60,30 @@ def read(self, nbytes):
...
+            for minexp in (4, 6, 7, 11, 12, CHUNK_MIN_EXP):
+                for maxexp in (15, 17, CHUNK_MAX_EXP):
+                    if minexp >= maxexp:
+                        continue
+                    for maskbits in (4, 7, 10, 12, HASH_MASK_BITS):
+                        for seed in (1849058162, 1234567653):
+                            fh = BytesIO(data)
+                            chunker = Chunker(seed, minexp, maxexp, maskbits, winsize)
+                            for c in chunker.chunkify(fh, -1):
+                                h.update(sha256(c).hexdigest().encode())
```
ThomasWaldmann: just use `h.update(sha256(c).digest())` ?

also, in general and if you want to use some old-fashioned hash algo from stdlib: please rather use sha512, it is faster on 64bit cpus.

otherwise rather use blake2b, it is much faster than that.

src/borg/testsuite/chunker.py
```diff
@@ -59,3 +60,30 @@ def read(self, nbytes):
...
+                            h.update(b"\n")
+
+        self.assert_equal(h.hexdigest(), "ef6ba5fbf60391af0b45692c991fa2490b0fd5865afb337f52498a677a2bd670")
```
ThomasWaldmann: there should be a comment above this line about how this hash was computed.

src/borg/_chunker.c
```diff
@@ -282,15 +282,19 @@ chunker_process(Chunker *c)
...
+            sum = buzhash_update(sum, p[0], p[window_size],
+                                 window_size, c->table);
```
ThomasWaldmann: a small micro opt that could be still done is to use just `table` here and pull the pointer deref to outside of the loop.

ncleaton:
Good feedback, thanks.

I'm not sure if I should amend the commits on my branch and force push, or just add a new commit with the feedback fixes. Do you have a preference ?

ThomasWaldmann:
Additional commits are easier to review on github.

src/borg/testsuite/chunker.py
ncleaton: switching it blake2b, thanks for the tip

src/borg/testsuite/chunker.py
ncleaton: I was using a single digest over all of the chunk digests of all the test runs, and this was to mark the end of one run and the start of the next.
With the switch to binary digests rather than hex, I'm now building a distinct digest for each run and then finally building a digest of those digests.

src/borg/testsuite/chunker.py
```diff
@@ -59,3 +60,30 @@ def read(self, nbytes):
...
+    def test_random_input(self):
+        def twist(size):
+            x = 1
+            a = bytearray(size)
+            for i in range(size):
+                x = (x * 1103515245 + 12345) & 0x7FFFFFFF
+                a[i] = x & 0xFF
+            return a
+
+        data = twist(100000)
```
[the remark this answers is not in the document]
ncleaton: Good point, I'll take some of those out

src/borg/testsuite/chunker.py
```diff
@@ -59,3 +60,30 @@ def read(self, nbytes):
         chunker = get_chunker(*CHUNKER_PARAMS, seed=0)
         reconstructed = b''.join(chunker.chunkify(SmallReadFile()))
         assert reconstructed == b'a' * 20
+
+    def test_random_input(self):
```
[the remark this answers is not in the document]
ncleaton: Now with blake2b and the test cases trimmed down a little, 1.2 seconds on my old laptop with the chunking performance pasted in the first comment.

src/borg/testsuite/chunker.py
ncleaton: adding one

src/borg/_chunker.c
ncleaton: I tried that, it made no measurable difference for me.

ThomasWaldmann:
travis is unhappy, the blake2b import does not work.

src/borg/testsuite/chunker.py
```diff
@@ -82,11 +83,10 @@ def twist(size):
...
                             fh = BytesIO(data)
                             chunker = Chunker(seed, minexp, maxexp, maskbits, winsize)
-                            runhash = blake2b()
-                            for c in chunker.chunkify(fh, -1):
-                                runhash.update(blake2b(c).digest())
-                            overall_hash.update(runhash.digest())
+                            chunks = [blake2b_256(b'', c) for c in chunker.chunkify(fh, -1)]
+                            runs.append(blake2b_256(b'', b''.join(chunks)))
 
         # The "correct" hash below matches the existing chunker behavior.
```
ThomasWaldmann: Just to make sure: the known-good hash was computed by running the test using the chunker code from *before* your change?

ncleaton: I've lost track, but I just ran the test against the current master branch and it passed with the same good-hash, so it's good.

ThomasWaldmann: ok, great!

ThomasWaldmann:
there is a "self test count" mismatch now, this is why most tests now fail.

can you search for this phrase and adjust it to 38 (+1), please?

ncleaton:
Fixed now, the 2 travis failures on the latest run were both out of time in archiver.py before it got to my new test.

ThomasWaldmann:
Restarted the 2 macOS jobs, somehow that stuff was extremely slow (but usually is only slow).

ThomasWaldmann:
Hmm, I recently noticed that we run into timeout issues with the macOS based tests on travis-ci.

That is totally unrelated to this PR as it is not merged yet.

But while thinking about it, I noticed that if that new chunker test is included in the self-tests (which are executed at **each** borg start), it will make the tests even slower (and not just 1 test time slower, but N test times slower).

So, I guess while we want to have that test in our testsuite, we do not want to have it in the self-test tests.

ThomasWaldmann:
Looks like the travis-ci macOS tests have a performance issue (in general) as they time out all the time currently.

Looks like this is ready to merge (I guess I'ld squash all commits together into one)?

ncleaton:
commits squashed

ThomasWaldmann:
Yay, we have a faster chunker now! Merged, thanks for your work!

Issue #4788, opened 2019-10-19
borg list without checkpoint archives


witten:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yup.

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE / feature request.

#### Your borg version (borg -V).

1.1.10

#### Operating system (distribution) and version.

Manjaro Linux 18.1.1

#### Hardware / network configuration, and filesystems used.

Lenovo X1 Carbon ext4

#### How much data is handled by borg?

Very little in this contrived example: ~164 MB.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

First, interrupt `borg create` with ctrl-C such that a checkpoint archive is created. (It's easier to do this by lowering the checkpoint interval artificially.)

Then:

```
# borg list test.borg
...
flux-2019-10-16T10:00:42.004841      Wed, 2019-10-16 10:00:42 [8f68292443590242781686b9d1c4e1a702903292dce146193a9133e4a5647985]
flux-2019-10-16T10:01:13.113527.checkpoint Wed, 2019-10-16 10:01:13 [d4fceca2de82f9629edbfa151e4ff2ced15cc6b80d81b571a4d3bbe5d4403b7d]
...
```

## Describe the problem you're observing.

It's not a problem so much as a feature request. I'd like to be able to `borg list` while having Borg omit checkpoint archives. My motivation is that I'm trying to programmatically find completed (non-checkpoint) archives, so that I can report on / monitor whether Borg is successfully creating archives.

It's possible to do this checkpoint filtering outside of Borg via `grep` or similar, but as discussed in #4047, that would prevent me from using this checkpoint filtering in conjunction with other `borg list` filters like `--last`, and/or in conjunction with `--json`.

So it would be really great if Borg had an option like `--consider-checkpoints`, similar to the existing `--consider-part-files`. Note that this implies inverting the default behavior, and omitting checkpoints from archive listing by default.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

I can reproduce `borg list` showing checkpoint archives, as that is the default behavior in current versions of Borg.

#### Include any warning/errors/backtraces from the system logs

N/A

ThomasWaldmann:
A good time to introduce this will be borg 1.2 (== master branch now).

natea123:
Eager to try and take a crack at this if its still available.

ThomasWaldmann:
[...]

PR #4835, opened 2019-11-11
Issue #4788 added consider checkpoints

atkailash:
For issue https://github.com/borgbackup/borg/issues/4788

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -109,7 +109,7 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
                     exclusive=False, manifest=True, cache=False, secure=True,
                     compatibility=None):
     """
-    Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, …)
+    Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, &)
 
     If a parameter (where allowed) is a str the attribute named of args is used instead.
     :param fake: (str or bool) use None instead of repository, don't do anything else
@@ -2492,7 +2492,7 @@ class Archiver:
         # e.g. through "borg ... --help", define a substitution for the reference here.
         # It will replace the entire :ref:`foo` verbatim.
         rst_plain_text_references = {
-            'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
+            'a_status_oddity': '"I am seeing A (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
         }
 
@@ -3824,6 +3824,8 @@ class Archiver:
                                           formatter_class=argparse.RawDescriptionHelpFormatter,
                                           help='list archive or repository contents')
         subparser.set_defaults(func=self.do_list)
+        subparser.add_argument('--consider-checkpoints', action='store_true', dest='consider_checkpoints',
+                               help='Show checkpoints in the repository contents. (Hidden by default)')
         subparser.add_argument('--short', dest='short', action='store_true',
                                help='only print file/directory names, nothing else')
         subparser.add_argument('--format', '--list-format', metavar='FORMAT', dest='format',
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -109,7 +109,7 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
                     exclusive=False, manifest=True, cache=False, secure=True,
                     compatibility=None):
     """
-    Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, &)
+    Method decorator for subcommand-handling methods: do_XYZ(self, args, repository, …)
 
     If a parameter (where allowed) is a str the attribute named of args is used instead.
     :param fake: (str or bool) use None instead of repository, don't do anything else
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -3824,6 +3824,8 @@ def define_borg_mount(parser):
                                           formatter_class=argparse.RawDescriptionHelpFormatter,
                                           help='list archive or repository contents')
         subparser.set_defaults(func=self.do_list)
+        subparser.add_argument('--consider-checkpoints', action='store_true', dest='consider_checkpoints',
+                help='Show checkpoints in the repository contents. (Hidden by default)')
```
ThomasWaldmann: Show checkpoint archives in the repository contents list (default: hidden).

src/borg/helpers/manifest.py
```diff
@@ -86,7 +86,10 @@ def list(self, *, glob=None, match_end=r'\Z', sort_by=(), first=None, last=None,
         if isinstance(sort_by, (str, bytes)):
             raise TypeError('sort_by must be a sequence of str')
         regex = re.compile(shellpattern.translate(glob or '*', match_end=match_end))
-        archives = [x for x in self.values() if regex.match(x.name) is not None]
+        if consider_checkpoints:
+            archives = [x for x in self.values() if regex.match(x.name) is not None]
+        else:
+            archives = [x for x in self.values() if ((regex.match(x.name) is not None) and ('checkpoint' not in x.name))]
```
ThomasWaldmann: well, guess it is a bit a matter of taste, but I'ld rather avoid duplicating the code that would build the full list and just remove the checkpoints from that list. it might be a little slower, but usually we do not have that many archives and the code is simpler that way.

so how about:
```
archives = [x for x in self.values() if regex.match(x.name) is not None]
if not consider_checkpoints:
    archives = [x for x in archives if 'checkpoint' not in x.name]
```

Also please check other places in the code whether `'checkpoint' not in x.name` is the correct way to check for a checkpoint archive name.

src/borg/helpers/manifest.py
```diff
@@ -99,13 +102,13 @@ def list(self, *, glob=None, match_end=r'\Z', sort_by=(), first=None, last=None,
 
     def list_considering(self, args):
         """
-        get a list of archives, considering --first/last/prefix/glob-archives/sort cmdline args
+        get a list of archives, considering --first/last/prefix/glob-archives/sort/consider-checkpoints cmdline args
         """
         if args.location.archive:
-            raise Error('The options --first, --last, --prefix and --glob-archives can only be used on repository targets.')
+            raise Error('The options --first, --last, --prefix, --glob-archives, and --consider-checkpoints can only be used on repository targets.')
```
ThomasWaldmann: is ", and" grammatically correct?

src/borg/helpers/manifest.py
atkailash: I did it this way for speed considerations, but I'm happy to modify it to just remove them instead of effectively duplicating if that's preferred. It should be fairly quick in this case I guess so the speed hits wouldn't be significant.

src/borg/helpers/manifest.py
atkailash: It's an Oxford comma (https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/). But I'm happy to remove it if it's not preferred.

src/borg/archiver.py
atkailash: Will make this change.

atkailash:
Kept the Oxford comma, but made modifications to the manifest.py and subparser for help text. Will also check other source for where it looks for checkpoints.

src/borg/helpers/manifest.py
```diff
@@ -86,7 +86,9 @@ def list(self, *, glob=None, match_end=r'\Z', sort_by=(), first=None, last=None,
         if isinstance(sort_by, (str, bytes)):
             raise TypeError('sort_by must be a sequence of str')
         regex = re.compile(shellpattern.translate(glob or '*', match_end=match_end))
-        archives = [x for x in self.values() if regex.match(x.name) is not None]
+        archives = [x for x in self.values() if ((regex.match(x.name) is not None) and ('checkpoint' not in x.name))]
+        if not consider_checkpoints:
+            archives = [x for x in self.values() if regex.match(x.name) is not None]
```
ThomasWaldmann: 1st line should build the full list into `archives` (and only do the regex match).

3rd line should only remove the checkpoint archives from the already computed `archives` list (not do the regex stuff again).

src/borg/helpers/manifest.py
ThomasWaldmann: ok, as long as it is not wrong, we can keep it.

src/borg/helpers/manifest.py
ThomasWaldmann: The way how you did it is slower as you did not use if/else, but you do build the list twice if consider_checkpoints is False.

with if/else, it would be faster than what I suggested, but I prefer to reduce code duplication.

ThomasWaldmann:
Please check the travis-ci output, lots of test fails.

ThomasWaldmann:
ping.

atkailash:
Sorry, been busy with job interviews and such.

I noticed those errors as well, but they also seem to have been occurring for PRs prior to this one as well, so I didn't think much of it. I'll take a closer look at it to ensure it's not from my changes while I work on the changes suggested.

ThomasWaldmann:
ping.

mirobertod:
#4918

ThomasWaldmann:
superseded by #4918.

Issue #4853, opened 2019-11-24
test recreate_timestamp FAILED 


jjgalvez:
borg/testsuite/archiver.py::ArchiverTestCase::test_recreate_timestamp FAILED 

## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).
N/A Trying to build from git

#### Operating system (distribution) and version.
Arch fully updated this morning

#### Hardware / network configuration, and filesystems used.
N/A

#### How much data is handled by borg?
N/A

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
N/A

## Describe the problem you're observing.
Trying to build the latest 1.2.0a7 tag from git, and the test suite fails with 2 errors

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.
yes the issue is reproducible, the test suite fails every time I try to build it

#### Include any warning/errors/backtraces from the system logs
```
========================================================= FAILURES =========================================================
_________________________________________ ArchiverTestCase.test_recreate_timestamp _________________________________________

self = <borg.testsuite.archiver.ArchiverTestCase testMethod=test_recreate_timestamp>

    def test_recreate_timestamp(self):
        self.create_test_files()
        self.cmd('init', '--encryption=repokey', self.repository_location)
        archive = self.repository_location + '::test0'
        self.cmd('create', archive, 'input')
        self.cmd('recreate', '--timestamp', "1970-01-02T00:00:00", '--comment',
                 'test', archive)
        info = self.cmd('info', archive).splitlines()
>       assert any([re.search(r'Time \(start\).+ 1970-01-02', item) for item in info])
E       AssertionError: assert False
E        +  where False = any([None, None, None, None, None, None, ...])

borg/testsuite/archiver.py:2562: AssertionError
______________________________________ RemoteArchiverTestCase.test_recreate_timestamp ______________________________________

self = <borg.testsuite.archiver.RemoteArchiverTestCase testMethod=test_recreate_timestamp>

    def test_recreate_timestamp(self):
        self.create_test_files()
        self.cmd('init', '--encryption=repokey', self.repository_location)
        archive = self.repository_location + '::test0'
        self.cmd('create', archive, 'input')
        self.cmd('recreate', '--timestamp', "1970-01-02T00:00:00", '--comment',
                 'test', archive)
[...]

PR #4861, opened 2019-11-28
Fix the test failing due to timestamp localisation

Gu1nness:
Fixes issue #4853

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -18,6 +18,7 @@ import unittest
 from binascii import unhexlify, b2a_base64
 from configparser import ConfigParser
 from datetime import datetime
+from datetime import timezone
 from datetime import timedelta
 from hashlib import sha256
 from io import BytesIO, StringIO
@@ -2552,6 +2553,10 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         assert sha256_before == sha256_after
 
     def test_recreate_timestamp(self):
+        if sys.version_info >= (3, 6):
+            local_timezone = datetime.now(timezone(timedelta(0))).astimezone().tzinfo
+        else:
+            local_timezone = datetime.now(timezone.utc).astimezone().tzinfo
         self.create_test_files()
         self.cmd('init', '--encryption=repokey', self.repository_location)
         archive = self.repository_location + '::test0'
@@ -2559,8 +2564,9 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         self.cmd('recreate', '--timestamp', "1970-01-02T00:00:00", '--comment',
                  'test', archive)
         info = self.cmd('info', archive).splitlines()
-        assert any([re.search(r'Time \(start\).+ 1970-01-02', item) for item in info])
-        assert any([re.search(r'Time \(end\).+ 1970-01-02', item) for item in info])
+        dtime = datetime(1970, 1, 2) + local_timezone.utcoffset(None)
+        assert any([re.search(r'Time \(start\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
+        assert any([re.search(r'Time \(end\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
 
     def test_recreate_dry_run(self):
         self.create_regular_file('compressible', size=10000)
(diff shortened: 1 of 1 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -2552,15 +2553,20 @@ def test_recreate_recompress(self):
...
         archive = self.repository_location + '::test0'
         self.cmd('create', archive, 'input')
         self.cmd('recreate', '--timestamp', "1970-01-02T00:00:00", '--comment',
                  'test', archive)
         info = self.cmd('info', archive).splitlines()
-        assert any([re.search(r'Time \(start\).+ 1970-01-02', item) for item in info])
-        assert any([re.search(r'Time \(end\).+ 1970-01-02', item) for item in info])
+        dtime = datetime(1970, 1, 2) + local_timezone.utcoffset(None)
+        assert any([re.search(r'Time \(start\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
+        assert any([re.search(r'Time \(end\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
```
ThomasWaldmann: see my comment in the ticket.

src/borg/testsuite/archiver.py
```diff
@@ -2552,15 +2553,20 @@ def test_recreate_recompress(self):
         assert sha256_before == sha256_after
 
     def test_recreate_timestamp(self):
+        if sys.version_info >= (3, 6):
+            local_timezone = datetime.now(timezone(timedelta(0))).astimezone().tzinfo
+        else:
+            local_timezone = datetime.now(timezone.utc).astimezone().tzinfo
```
ThomasWaldmann: is there a way to do this without checking the python version?

src/borg/testsuite/archiver.py
ThomasWaldmann: can you move the strftime call to outside the loops, only computing it once after line 2567?

src/borg/testsuite/archiver.py
```diff
@@ -2553,10 +2553,7 @@ def test_recreate_recompress(self):
         assert sha256_before == sha256_after
 
     def test_recreate_timestamp(self):
-        if sys.version_info >= (3, 6):
-            local_timezone = datetime.now(timezone(timedelta(0))).astimezone().tzinfo
-        else:
-            local_timezone = datetime.now(timezone.utc).astimezone().tzinfo
+        local_timezone = datetime.now(timezone(timedelta(0))).astimezone().tzinfo
```
ThomasWaldmann: so, this also works for py35?

src/borg/testsuite/archiver.py
```diff
@@ -2565,8 +2562,9 @@ def test_recreate_timestamp(self):
                  'test', archive)
         info = self.cmd('info', archive).splitlines()
         dtime = datetime(1970, 1, 2) + local_timezone.utcoffset(None)
-        assert any([re.search(r'Time \(start\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
-        assert any([re.search(r'Time \(end\).+ %s' % dtime.strftime("%Y-%m-%d"), item) for item in info])
+        s_time = dtime.strftime("%Y-%m-%d")
+        assert any([re.search(r'Time \(start\).+ %s' % s_time , item) for item in info])
```
ThomasWaldmann: pep8 checker is unhappy:

src/borg/testsuite/archiver.py:2566:62: E203 whitespace before ','

src/borg/testsuite/archiver.py
Gu1nness: According to my tests and Travis, yes :)

src/borg/testsuite/archiver.py
ThomasWaldmann: fixed now.

ThomasWaldmann:
i played a bit with datetime, tried your code and thought about the problem again.

from https://borgbackup.readthedocs.io/en/stable/usage/general.html#date-and-time
```
Unless otherwise noted, we display local date and time.
Internally, we store and process date and time as UTC.
```
OK, that does not directly say something about `--timestamp`, but as the cli is also part of the user interface (as output is), this might mean that maybe not the test needs fixing, but our processing of that option.

**But, the docs also say**:
```
--timestamp TIMESTAMP
manually specify the archive creation date/time (UTC,  yyyy-mm-ddThh:mm:ss format).
Alternatively, give a reference  file/directory.
```

So, using UTC there is a documented feature, although it is different from output using localtime.

So, the test fix in this PR is correct according to the docs. \o/

ThomasWaldmann:
thanks! can you check if 1.1-maint has the same issue?

Gu1nness:
Yes, see #4873 for the commit fixing this on 1.1-maint

PR #4866, opened 2019-12-04
Elaborate on "Calculating size"

BirkhoffLee:
It actually took me several days to figure out that `--progress` was the culprit to Borg constantly being SIGKILL'd by the kernel, because Borg ate all of the memory, and because I enabled progress option. This elaborates on what's the calculation for.

This is not intended to be a "garbage" PR, but if you think this is useless you can close this.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -778,7 +778,7 @@ class Archiver:
         filter = self.build_filter(matcher, peek_and_store_hardlink_masters, strip_components)
         if progress:
             pi = ProgressIndicatorPercent(msg='%5.1f%% Extracting: %s', step=0.1, msgid='extract')
-            pi.output('Calculating total archive size for the progress indicator (might take long if it\'s a large one)')
+            pi.output('Calculating total archive size for the progress indicator (might take long for large archives)')
             extracted_size = sum(item.get_size(hardlink_masters) for item in archive.iter_items(filter))
             pi.total = extracted_size
         else:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
OK, I see that the hint that it might take long is useful (especially as it ADDS time to whatever it will need anyway).

But it is unclear to me how this is related to borg getting killed by the OOM killer - can you clarify?

BirkhoffLee:
@ThomasWaldmann Oh sorry. I was extracting a backup and used `--progress` option. With that option enabled the `Calculating size` takes a few hours, and after I came back borg gets killed. I check system log and found out it was because of OOM. Wasn't really sure which part of the extraction procedure was causing OOM though

ThomasWaldmann:
It would be interesting if you could compare extract with progress vs. extract without and watch the memory consumption. If there is a significant difference, file a bug with that data.

ThomasWaldmann:
Also, can we have timing with/without progress option?

src/borg/archiver.py
```diff
@@ -778,7 +778,7 @@ def peek_and_store_hardlink_masters(item, matched):
         filter = self.build_filter(matcher, peek_and_store_hardlink_masters, strip_components)
         if progress:
             pi = ProgressIndicatorPercent(msg='%5.1f%% Extracting: %s', step=0.1, msgid='extract')
-            pi.output('Calculating size')
+            pi.output('Calculating total archive size for the progress indicator (might take long if it\'s a large one)')
```
ThomasWaldmann: `... (might take long for large archives)`

avoids the ugly backslash and also clearer.

BirkhoffLee:
> Also, can we have timing with/without progress option?

Will do when I got time. We can probably merge this already and I'll file an issue later if the bug actually exists, I want to make sure it wasn't an issue on my end first.

ThomasWaldmann:
Merged. Guess it is useful to know for what the additional runtime is needed.

How this is related to OOM is still unclear.

The PR was against master, would you like to make the same change to 1.1-maint branch?

PR #4873, opened 2019-12-12
The test recreate_timestamp is fixed

Gu1nness:
The test was failing due to localisation.
This is fixed by this commit.


ThomasWaldmann:
thanks!

PR #4885, opened 2019-12-18
Elaborate on "Calculating size" message

BirkhoffLee:
This is the same change as https://github.com/borgbackup/borg/pull/4866
@ThomasWaldmann


ThomasWaldmann:
Thanks!

Issue #4899, opened 2019-12-23
How much of a cryptographic vulnerability would a leaked cache pose?


finefoot:
There is a short exchange about this on the mailing list: https://mail.python.org/pipermail/borgbackup/2019q1/001267.html

The question was if it's okay to put the (unencrypted) cache with the (encrypted) repository. While it has been answered to "not do this" due to the cache containing metadata information, there is no deeper explanation.

Assuming one doesn't care about metadata and puts the cache with the repo anyway or in case the cache has inadvertently been leaked: How much of a cryptographic vulnerability would one be facing?

(I'm not planning to leak the cache on purpose, like putting it with the encrypted repository - I'm just wondering about how much I should worry about keeping the cache secure.)

ThomasWaldmann:
Fixed by #4900, backport to 1.1-maint still required.

finefoot:
@ThomasWaldmann 

> Fixed by #4900

It is? My understanding is, that the changes introduced by #4900 clarify that it's of importance to put the cache at a secure location as well as the related essential steps to help prevent leaking of the cache.

However, this issue poses the question about cryptographic vulnerability of the borg repo/archive *in case the cache has already been leaked*.

finefoot:
I mean, maybe the answer is simply: There is no cryptographic vulnerability from a leaked cache. You'll only expose unencrypted metadata information.

Is that what you mean?

ThomasWaldmann:
The cache is not encrypted, thus even the title of this issue is unclear.

You are not supposed to "leak" the cache and the default location is considered to be secure.

The files cache contents are H(filepath) -> (size, timestamp, inode number, chunkid list).
The chunks cache contents are chunkid -> (size, csize, refcount).

So, yes, it is metadata. Whether it is "only" metadata (in the sense that metadata does not matter that much) pretty much depends on your threat model.

finefoot:
@ThomasWaldmann 

> The cache is not encrypted, thus even the title of this issue is unclear.

See my previous comment: "this issue poses the question about cryptographic vulnerability of the borg repo/archive in case the cache has already been leaked". Take the following scenario for example:

- My drive, and therefore my home folder, is fully encrypted.
- However, I forgot to lock my laptop and an attacker was able to copy my `.cache/borg` folder.
- My borg repo/archive is of course still encrypted with a password.
- Will the copy of the cache, which is now in the possession of the attacker, help with attacking the borg repo?
- Or is the only use of that copy the gained information about metadata?

[...]

Issue #4894, opened 2019-12-23
Why does XDG_CONFIG_HOME have precedence over BORG_BASE_DIR?


finefoot:
I want to store the complete borg app data (config, cache, etc.) together with the backup data (in a different folder) on my external hard drive. The docs at https://borgbackup.readthedocs.io/en/stable/usage/general.html#environment-variables say:

> **BORG_BASE_DIR**
>     Default to `$HOME`, `~$USER`, `~` (in that order). If we refer to ~ below, we in fact mean BORG_BASE_DIR.
> **BORG_CONFIG_DIR**
>     Default to ‘~/.config/borg’. This directory contains the whole config directories.

This appears to imply that I won't have to set BORG_CONFIG_DIR, BORG_CACHE_DIR etc. separately but all is done with setting BORG_BASE_DIR. (In my view, this is not only convenient but actually makes sense in case another path variable will be added in the future for some feature xyz like `~/.xyz/borg` with BORG_XYZ_DIR, which I wouldn't have to explicitly set since I have already set BORG_BASE_DIR.)

However, this does not work if XDG_CONFIG_HOME is set, as it has precedence over BORG_BASE_DIR when the other paths are being assembled. What's the reason behind this?

I'm currently using borg-linux64 1.1.10, but the issue is relevant to both `master` and `1.1-maint`: See https://github.com/borgbackup/borg/blob/master/src/borg/helpers/fs.py#L80 and https://github.com/borgbackup/borg/blob/1.1-maint/src/borg/helpers.py#L567, respectively.

I would have expected something like this

```python
config_home = os.path.join(get_base_dir(), '.config')
if not os.environ.get('BORG_BASE_DIR'):
    config_home = os.environ.get('XDG_CONFIG_HOME', config_home)
config_dir = os.environ.get('BORG_CONFIG_DIR', os.path.join(config_home, 'borg'))
```

or this

```python
config_dir = os.environ.get('BORG_CONFIG_DIR')
if not config_dir:
    config_home = os.path.join(get_base_dir(), '.config')
    if not os.environ.get('BORG_BASE_DIR'):
        config_home = os.environ.get('XDG_CONFIG_HOME', config_home)
    config_dir = os.path.join(config_home, 'borg')
```

as well as XDG_CONFIG_HOME being mentioned in the docs.

ThomasWaldmann:
The cache data is not encrypted, so if you have your repo encrypted (as it is by default), it is not a good idea to store the cache files with the repo files.

[...]

PR #4900, opened 2019-12-23
Add paragraph regarding cache security assumptions

finefoot:
I think it might be a good idea to add this information to the FAQs. While it has been answered on the mailing list, it was missing from the docs, as far as I can tell: If this assumption of "cache is at the same place as original files" is being made, it should be noted somewhere.

Could be extended further depending on #4899.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -352,6 +352,17 @@ to change them.
 Security
 ########
 
+Do I need to take security precautions regarding the cache?
+-----------------------------------------------------------
+
+The cache contains a lot of metadata information about the files in
+your repositories and it is not encrypted.
+
+However, the assumption is that the cache is being stored on the very
+same system which also contains the original files which are being
+backed up. So someone with access to the cache files would also have
+access the the original files anyway.
+
 How can I specify the encryption passphrase programmatically?
 -------------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -352,6 +352,8 @@ to change them.
 Security
 ########
 
+.. _cache_security:
+
 Do I need to take security precautions regarding the cache?
 -----------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -363,6 +363,9 @@ same system which also contains the original files which are being
 backed up. So someone with access to the cache files would also have
 access the the original files anyway.
 
+If you ever need to move the cache to a different location, this can
+be achieved by using the appropriate :ref:`env_vars`.
+
 How can I specify the encryption passphrase programmatically?
 -------------------------------------------------------------
 
--- a/docs/usage_general.rst.inc
+++ b/docs/usage_general.rst.inc
@@ -257,7 +257,8 @@ Directories and files:
         Default to '~/.config/borg'. This directory contains the whole config directories.
     BORG_CACHE_DIR
         Default to '~/.cache/borg'. This directory contains the local cache and might need a lot
-        of space for dealing with big repositories.
+        of space for dealing with big repositories. Make sure you're aware of the associated
+        security aspects of the cache location: :ref:`cache_security`
     BORG_SECURITY_DIR
         Default to '~/.config/borg/security'. This directory contains information borg uses to
         track its usage of NONCES ("numbers used once" - usually in encryption context) and other


fantasya-pbem:
I'd also mention BORG_CACHE_DIR, and when setting it to a remote file system one should think about adding custom encryption.

finefoot:
> I'd also mention BORG_CACHE_DIR, and when setting it to a remote file system one should think about adding custom encryption.

Agreed, good idea. :+1: After all, I ended up doing exactly that and only looked into these cache security assumptions after trying to set a custom path for the cache and config folders. (By the way, I wrongly assumed this could be accomplished by simply setting BORG_BASE_DIR. See #4894 if you don't mind weighing in there, too.)

I'll add a commit with a sentence on BORG_CACHE_DIR and encryption shortly.

ThomasWaldmann:
BORG_CACHE_DIR should **not** ever be set to a remote directory.

It is a cache, it is supposed to be local and fast.

finefoot:
> BORG_CACHE_DIR should **not** ever be set to a remote directory.

Could you elaborate on that? That's not universally true but advised for best security/performance, right?

(Also, an external hard drive would count as local in this case, I guess?)

ThomasWaldmann:
Yes, caches are for performance and thus it should be local (not over slow or higher latency network).

docs/faq.rst
```diff
@@ -352,6 +352,17 @@ to change them.
...
+-----------------------------------------------------------
+
+The cache contains a lot of metadata information about the files in
+your repositories and it is not encrypted.
+
+However, the assumption is that the cache is being stored on the very
+same system which also contains the original files which are being
+backed up. So someone with access to the cache files would also have
+access the the original files anyway.
+
```
ThomasWaldmann: Sounds good. Maybe it could point to the description of the env vars which can be used to configure the cache location.

finefoot:
> Yes, caches are for performance and thus it should be local (not over slow or higher latency network).

Yeah, sure, that makes sense. I was just asking because [your previous comment](https://github.com/borgbackup/borg/pull/4900#issuecomment-571354205) was worded in such absolute fashion:

> BORG_CACHE_DIR should **not** ever be set to a remote directory.

If the "remote directory" is still fast, why not put the cache there?

---

In any case, any further information about requirements for moving the cache to a different location (if deemed necessary) like read/write speeds, latency and encryption should probably go to the env vars section (i.e. with the description how to change the location, in the first place) or a whole other new FAQ section, but not into the security assumptions section of this pull request, right?

docs/faq.rst
finefoot: :+1:

ThomasWaldmann:
Speed, latency, space and security are the considerations one needs to do.

I thought about it again and maybe putting that rather close to the env vars instead of into the faq would be better.

The cache is at a good place by default (home dir of the user running borg). So moving it away from there needs using those env vars, so the stuff one needs to consider belong to these env vars docs.

finefoot:
> Speed, latency, space and security are the considerations one needs to do.

Yes, sure, I'm well aware of that. Like I said, I was just taken aback by the way you worded your comment, saying cache should "not ever" be located in a remote directory. ;) It's not the remote directory which is the problem but the possible loss of speed and latency.

> I thought about it again and maybe putting that rather close to the env vars instead of into the faq would be better.
> 
> The cache is at a good place by default (home dir of the user running borg). So moving it away from there needs using those env vars, so the stuff one needs to consider belong to these env vars docs.

It goes both ways: For example, I was backing up data from an encrypted external hard drive, initially using a laptop whose home folder (and therefore the default location of the cache folder) was *not* encrypted. Yes, I think, borg's assumption to put the cache into the home folder is fair. At the same time, this assumption should be documented in the FAQs and not only in the env vars section, because it applies to every borg user, not only those who already look into changing the cache location. Therefore, the paragraph seems to be most appropriately located in the FAQs.

Of course, it also makes sense to have a link to the cache security informations for people who look into the env vars with the intention to change its location. Therefore, the env vars section should also have a link to the cache paragraph in the FAQs.

See most recent commit. What do you think?

ThomasWaldmann:
Thanks! Could you do a backport to 1.1-maint?

finefoot:
@ThomasWaldmann  I'm glad to see that you were satisfied with the changes. :)

> Could you do a backport to 1.1-maint?

Sure, what's the procedure here? Just open a new pull request against 1.1-maint branch?

By the way, just curious, shouldn't the changes be visible at https://borgbackup.readthedocs.io/en/master/?

ThomasWaldmann:
Yes, open a new PR against 1.1-maint branch.

ThomasWaldmann:
I manually triggered a build at readthedocs.io, now it shows up. Strange that it did not do that automatically, though.

finefoot:
> Yes, open a new PR against 1.1-maint branch.

Thanks! :+1: Done, see #4930

> I manually triggered a build at readthedocs.io, now it shows up. Strange that it did not do that automatically, though.

I didn't have a closer look at the deployment scripts, but it might have happened because the merge commit had failed tests: https://github.com/borgbackup/borg/commit/61b8234704b83ba946df20bf129d74f8ab7dd699

PR #4903, opened 2019-12-30
support platforms with no os.link

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -744,8 +744,17 @@ def do_extract(self, args, repository, manifest, key, archive):
         hardlink_masters = {} if partial_extract else None
```
willsALMANJ: This needs to be something like:

```
         hardlink_masters = {} if partial_extract or not has_link else None 
```

ThomasWaldmann: I'll fix this, thanks!

ThomasWaldmann: fixed.

do you think it is ready to merge now?

ThomasWaldmann: update: reread the ticket, answer seems "yes".

PR #4918, opened 2020-01-11
Issue #4788 added consider checkpoints and relative test

mirobertod:
Took a cue from #4835 

For issue #4788

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1115,6 +1115,7 @@ class Archiver:
                 archives.insert(0, args.location.archive)
             archive_names = tuple(archives)
         else:
+            args.consider_checkpoints = True
             archive_names = tuple(x.name for x in manifest.archives.list_considering(args))
             if not archive_names:
                 return self.exit_code
@@ -1321,6 +1322,7 @@ class Archiver:
         if args.location.archive:
             archive_names = (args.location.archive,)
         else:
+            args.consider_checkpoints = True
             archive_names = tuple(x.name for x in manifest.archives.list_considering(args))
             if not archive_names:
                 return self.exit_code
@@ -1409,6 +1411,7 @@ class Archiver:
             args.glob_archives = args.prefix + '*'
         checkpoint_re = r'\.checkpoint(\.\d+)?'
         archives_checkpoints = manifest.archives.list(glob=args.glob_archives,
+                                                      consider_checkpoints=True,
                                                       match_end=r'(%s)?\Z' % checkpoint_re,
                                                       sort_by=['ts'], reverse=True)
         is_checkpoint = re.compile(r'(%s)\Z' % checkpoint_re).search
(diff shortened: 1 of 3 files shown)


src/borg/helpers/manifest.py
```diff
@@ -86,7 +87,9 @@ def list(self, *, glob=None, match_end=r'\Z', sort_by=(), first=None, last=None,
         if isinstance(sort_by, (str, bytes)):
             raise TypeError('sort_by must be a sequence of str')
         regex = re.compile(shellpattern.translate(glob or '*', match_end=match_end))
-        archives = [x for x in self.values() if regex.match(x.name) is not None]
+        archives = [x for x in self.values() if ((regex.match(x.name) is not None) and ('.checkpoint' not in x.name))]
+        if consider_checkpoints:
+            archives = [x for x in self.values() if regex.match(x.name) is not None]
```
ThomasWaldmann: see my comments in the previous PR.

src/borg/archiver.py
```diff
@@ -3370,6 +3373,8 @@ def define_borg_mount(parser):
         subparser.set_defaults(func=self.do_delete)
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
+        subparser.add_argument('--consider-checkpoints', action='store_true', dest='consider_checkpoints',
```
ThomasWaldmann: why is that useful for borg delete?

src/borg/archiver.py
```diff
@@ -3565,6 +3570,8 @@ def define_borg_mount(parser):
         subparser.add_argument('location', metavar='REPOSITORY_OR_ARCHIVE', nargs='?', default='',
                                type=location_validator(),
                                help='repository or archive to display information about')
+        subparser.add_argument('--consider-checkpoints', action='store_true', dest='consider_checkpoints',
```
ThomasWaldmann: why is that useful for borg info?

src/borg/helpers/manifest.py
```diff
@@ -38,6 +38,7 @@ class Archives(abc.MutableMapping):
     and we can deal with str keys (and it internally encodes to byte keys) and either
     str timestamps or datetime timestamps.
     """
+
```
ThomasWaldmann: guess this is an accidental change? please remove.

mirobertod: yes, removed

mirobertod: it actually was the "format on save" setting of vscode

src/borg/archiver.py
mirobertod: delete and info command use the list command, so that's why I added that arg. but as you suggest it's not useful, now I forced the arg "consider-checkpoints" to true inside the two commands. I don't know if it's the right way to do that

src/borg/helpers/manifest.py
mirobertod: ok, fixed

src/borg/helpers/manifest.py
```diff
@@ -87,9 +86,9 @@ def list(self, *, glob=None, match_end=r'\Z', sort_by=(), consider_checkpoints=F
         if isinstance(sort_by, (str, bytes)):
             raise TypeError('sort_by must be a sequence of str')
         regex = re.compile(shellpattern.translate(glob or '*', match_end=match_end))
-        archives = [x for x in self.values() if ((regex.match(x.name) is not None) and ('.checkpoint' not in x.name))]
-        if consider_checkpoints:
-            archives = [x for x in self.values() if regex.match(x.name) is not None]
+        archives = [x for x in self.values() if (regex.match(x.name) is not None)]
```
ThomasWaldmann: you do not need the `()` around `regex.match(x.name) is not None`.

ThomasWaldmann:
looks good now (except the amount of fixfixfix commits. did you have a look at `git rebase -i master`?

Issue #4923, opened 2020-01-16
Borg repository locking is not safe


alfredo08154711:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes. And I discussed it in borgbackup@python.org (&ldquo;Isn't locking broken, because stale lock removal doesn't comply with the locking protocol?&rdquo;) with the confirmation that the issue most probably exists.

## Is this a BUG&nbsp;/ ISSUE report or a QUESTION?

BUG.

I did not (yet) observe any damage event resulting from it, and I don't know if anybody did. However, the risk of repository damage definitely exists.

Furthermore, the requirement for the filesystem on which the repository resides,<sup><a name="cite-1"/>[1](#user-content-footnote-1)</sup>

> mkdir(2) should be atomic, since it is used for locking

is useless because of this bug.

## System information. For client/server mode post info for both machines.

The bug is basical and affects all platforms and file systems.

However, what I observed in order to confirm it is how the implementation deals with the file system in order to acquire and to release a lock and to kill stale locks. The information below relates to these observations.

#### Your borg version (borg -V).

1.1.7 and&nbsp;1.1.10

#### Operating system (distribution) and version.

Linux&nbsp;4.15.0-65-generic, Ubuntu&nbsp;18.04.3&nbsp;LTS

#### Hardware&nbsp;/ network configuration, and file systems used.

ext4 (mounted rw,relatime), tmpfs (mounted rw,nosuid,nodev,relatime)

#### How much data is handled by borg?

empty and almost empty repository

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

The exact commandline doesn't matter. I tried the commands `borg list` and `borg mount` as processes creating a shared lock, and `borg with-lock` and `borg create` as processes creating an exclusive lock. See also [Can you reproduce the problem?&nbsp;&hellip;](#reproduce).

## Describe the problem you're observing.

The removal of stale locks violates the locking procedure as described in the documentation,<sup><a name="cite-2"/>[2](#user-content-footnote-2)</sup> <sup><a name="cite-3"/>[3](#user-content-footnote-3)</sup>

> If the process can create the lock.exclusive directory for a resource, it has the lock for it. If creation fails (because the directory has already been created by some other process), lock acquisition fails.

Let's assume two concurrent borg processes, _A_ and _B_. _A_ gets the lock. _B_ sees the lock.exclusive directory, thinking it be stale, removes it and recreates it, now thinking it owns the lock, too.

[...]

PR #4928, opened 2020-01-22
Fixed locking, issue #4923

alfredo08154711:
See #4923  .

### DIFF
--- a/src/borg/locking.py
+++ b/src/borg/locking.py
@@ -150,9 +150,17 @@ class ExclusiveLock:
                     temp_path = None  # see finally:-block below
                     return self
         finally:
-            if temp_path is not None:  # Renaming failed for any reason, so temp_dir still exists and should be cleaned up anyway.
-                os.unlink(temp_unique_name)
-                os.rmdir(temp_path)
+            if temp_path is not None:
+                # Renaming failed for some reason, so temp_dir still exists and
+                # should be cleaned up anyway. Try to clean up, but don't crash.
+                try:
+                    os.unlink(temp_unique_name)
+                except:
+                    pass
+                try:
+                    os.rmdir(temp_path)
+                except:
+                    pass
 
     def release(self):
         if not self.is_locked():
--- a/src/borg/locking.py
+++ b/src/borg/locking.py
@@ -147,7 +147,7 @@ class ExclusiveLock:
                     if timer.timed_out_or_sleep():
                         raise LockTimeout(self.path) from None
                 else:
-                    temp_path = None
+                    temp_path = None  # see finally:-block below
                     return self
         finally:
             if temp_path is not None:  # Renaming failed for any reason, so temp_dir still exists and should be cleaned up anyway.
--- a/src/borg/locking.py
+++ b/src/borg/locking.py
@@ -163,8 +163,8 @@ class ExclusiveLock:
         try:
             os.rmdir(self.path)
         except OSError as err:
-            if err.errno == errno.ENOTEMPTY:
-                # Directory is not empty = we lost the race to somebody else--which is ok.
+            if err.errno == errno.ENOTEMPTY or err.errno == errno.ENOENT:
+                # Directory is not empty or doesn't exist any more = we lost the race to somebody else--which is ok.
                 pass
             else:
                 # EACCES or EIO or ... = we cannot operate anyway, so re-throw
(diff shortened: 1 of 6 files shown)


alfredo08154711:
Not yet finished&mdash;I'm still preparing additional tests.

alfredo08154711:
Added TestExclusiveLock.test_race_condition() which fails because of issue #4923 and which passes with this PR.

So I think this PR is ready for review.

alfredo08154711:
In `TestExclusiveLock.test_race_condition()` I used `threading.Thread`. This is fine to reveal the bug in borgbackup.

However, I strongly guess that this is not suited to detect issues with the file system, i.e. if `mkdir(2)` or `rename(2)`, resp., doesn't work as specified. I think so, because Python threads are said to share I/O scheduling. So I fear that the file system calls are pretty serialized. In order to also detect the reliability under race condition of the file system, another test function should be written using `multiprocessing` instead of threading, maybe with the same logic.

alfredo08154711:
In the meantime I have also got an idea how to fix the problem with the lock migration, also for exclusive locks.

If there are no objections, I will add the fix to this PR.

ThomasWaldmann:
Yes, borg mount should also keep working.

But please do **not** extend the scope of this PR, locking behaviour must not change (e.g. the kinds of locks must not change).

ThomasWaldmann:
btw, i am currently rather busy otherwise, so reviewing this might still take a while...

alfredo08154711:
O.k., in order not to extend the scope of this PR, we should leave it as it is now. The daemonization and the lock migration together act in a way which also makes locking unsafe. However, this is actually an independent issue and has nothing to do with the problem which is solved by this PR&mdash;except that it makes locking unsafe.

So I will open another issue.

ThomasWaldmann:
I rebased this on current master and also solved the merged conflicts in the docs.

No content changes besides that.

src/borg/locking.py
```diff
@@ -124,30 +125,50 @@ def acquire(self, timeout=None, sleep=None):
...
+            timer = TimeoutTimer(timeout, sleep).start()
+            while True:
+                try:
+                    os.rename(temp_path, self.path)
+                except OSError:  # already locked
+                    if self.by_me():
+                        return self
+                    self.kill_stale_lock()
+                    if timer.timed_out_or_sleep():
+                        raise LockTimeout(self.path)
```
ThomasWaldmann, line 148: guess this should be `... from None` as we raise this from an exception handler block (but the original exception is not interesting in the traceback). the original code also has this cosmetic issue.

src/borg/locking.py
```diff
@@ -124,30 +125,50 @@ def acquire(self, timeout=None, sleep=None):
...
+                else:
+                    temp_path = None
```
ThomasWaldmann, line 150: maybe add a comment `# see finally: block` to make clear why this is useful.

src/borg/locking.py
```diff
@@ -124,30 +125,50 @@ def acquire(self, timeout=None, sleep=None):
...
-            except OSError as err:
-                raise LockFailed(self.path, str(err)) from None
-            else:
-                with open(self.unique_name, "wb"):
-                    pass
-                return self
+        finally:
+            if temp_path is not None:  # Renaming failed for any reason, so temp_dir still exists and should be cleaned up anyway.
+                os.unlink(temp_unique_name)
+                os.rmdir(temp_path)
```
ThomasWaldmann, line 155: if the open-wb fails, there is no temp_unique_name file, so guess we better put the first line into a `try: ... except: pass`

src/borg/locking.py
```diff
@@ -124,30 +125,50 @@ def acquire(self, timeout=None, sleep=None):
...
-        os.rmdir(self.path)
+        try:
+            os.rmdir(self.path)
+        except OSError as err:
+            if err.errno == errno.ENOTEMPTY or err.errno == errno.ENOENT:
+                # Directory is not empty or doesn't exist any more = we lost the race to somebody else--which is ok.
+                pass
+            else:
+                # EACCES or EIO or ... = we cannot operate anyway, so re-throw
+                raise err
```
ThomasWaldmann, line 171: can we invert the condition to save the `pass`?

like:
```
            if err.errno not in (errno.ENOTEMPTY, errno.ENOENT, ):
                # EACCES or EIO or ... = we cannot operate anyway, so re-throw
                raise err
            # else: directory is not empty or doesn't exist any more = we lost the race to somebody else (which is ok).
```

src/borg/locking.py
```diff
@@ -156,42 +177,46 @@ def by_me(self):
...
+                    logger.warning('Killed stale lock %s.', name)
+                except OSError as err:
+                    if not self.stale_warning_printed:
+                        # This error will bubble up and likely result in locking failure
+                        logger.error('Found stale lock %s, but cannot delete due to %s', name, str(err))
+                        self.stale_warning_printed = True
+                    return False
+
+        except FileNotFoundError:  # Another process did our job in the meantime.
+            pass
```
ThomasWaldmann, line 213: this is a quite wide scope for this exception handler, including `os.listdir`, `os.unlink` and some function calls.

could this catch and hide exceptions which we did not want to catch/hide?

src/borg/testsuite/locking.py
```diff
@@ -90,6 +94,80 @@ def test_migrate_lock(self, lockpath):
...
+                    return self.count
+
+            def decr(self):
+                with self.lock:
+                    self.count -= 1
+                    return self.count
+
+        def acquire_release_loop(id, timeout, thread_id, lock_owner_counter, exception_counter, print_lock, last_thread=None):
+            with print_lock:
+                print("Thread %2d: Starting acquire_release_loop(id=%s, timeout=%d); lockpath=%s" % (thread_id, id, timeout, lockpath))
```
ThomasWaldmann, line 128: define a print_locked function that does this?

src/borg/testsuite/locking.py
```diff
@@ -90,6 +94,80 @@ def test_migrate_lock(self, lockpath):
...
+        thread = None
+        for thread_id in range(RACE_TEST_NUM_THREADS):
+            thread = Thread(target=acquire_release_loop, args=(('foo', thread_id, 0), RACE_TEST_DURATION, thread_id, lock_owner_counter, exception_counter, print_lock, thread))
+            thread.start()
+        thread.join()  # joining the last thread
+
+        assert lock_owner_counter.maxvalue() > 0, 'Never gained the lock? Something went wrong here...'
+        assert lock_owner_counter.maxvalue() <= 1, "Maximal number of concurrent lock holders was %d. So exclusivity is broken." % (lock_owner_counter.maxvalue())
+        assert exception_counter.value() == 0, "ExclusiveLock threw %d exceptions due to unclean concurrency handling." % (exception_counter.value())
+
```
ThomasWaldmann, line 170: nice test! :-)

did you have this running / failing with the original code?

src/borg/testsuite/locking.py
```diff
@@ -10,6 +12,8 @@
 
 ID1 = "foo", 1, 1
 ID2 = "bar", 2, 2
+RACE_TEST_NUM_THREADS = 40
+RACE_TEST_DURATION = 0.4  # seconds
```
ThomasWaldmann, line 16: was this enough to see test fails with the original code?

src/borg/locking.py
ThomasWaldmann: ... and maybe also the second line into a separate handler (or do we ever want to fail while trying to clean up?)

ThomasWaldmann:
hmm, guess @alfredo08154711 is busy otherwise.

So I'll work a bit on the desired changes and push some more changesets to this PR...

src/borg/locking.py
ThomasWaldmann: done

src/borg/locking.py
ThomasWaldmann: done

src/borg/locking.py
ThomasWaldmann: done

src/borg/locking.py
ThomasWaldmann: done

src/borg/testsuite/locking.py
ThomasWaldmann: done

src/borg/locking.py
ThomasWaldmann: fixed (please review my change).

src/borg/testsuite/locking.py
alfredo08154711: Yes. On my laptop, at least, which was a fast machine...6 years ago. :-) We can increase that, if you deem it necessary, currently, I have no access to a fast machine. The probability should be very close to 1 to reveal the problem, so maybe there should be about 10 cases with more than two threads "thinking" they own the lock.

src/borg/testsuite/locking.py
alfredo08154711: Thanks :-)

ThomasWaldmann:
@alfredo08154711 can you review my changes? if ok, i guess we could merge this soon.

src/borg/locking.py
```diff
@@ -124,30 +125,50 @@ def acquire(self, timeout=None, sleep=None):
...
+            with open(temp_unique_name, "wb"):
+                pass
+        except OSError as err:
+            raise LockFailed(self.path, str(err)) from None
+        else:
+            timer = TimeoutTimer(timeout, sleep).start()
+            while True:
+                try:
+                    os.rename(temp_path, self.path)
+                except OSError:  # already locked
```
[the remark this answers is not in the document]
alfredo08154711: Yes.

src/borg/locking.py
alfredo08154711: I leave this to your judgement.

src/borg/locking.py
alfredo08154711: ok

src/borg/locking.py
alfredo08154711: Probably you're right. I remember I thought about it, and in my opinion it wasn't necessary, because if such kind of things happen, it would have already happened before. But basically, everything can happen during I/O at any time...

src/borg/locking.py
alfredo08154711: Yeah. You see, I'm a bloody Python beginner, I wasn't thinking in Python...

src/borg/locking.py
alfredo08154711: I agree.

src/borg/testsuite/locking.py
alfredo08154711: yeah, looks better now.

alfredo08154711:
All changes are o.k. from my side. Please go ahead. Thank you.

ThomasWaldmann:
\o/ thanks for fixing this!

Issue #3509, opened 2018-01-03
borg check crashing


ThomasWaldmann:
from irc user mlazze:

arch linux running inside virtualbox on a win10 host accessing borg repo via vboxsf (which is known-problematic). repo is on host ntfs filesystem.

this issue is not about vboxsf being unreliable, but rather about checking if we can make borg check more robust to not crash here:

```
~ borg check --debug -p /mnt/z/borg/borg.repo                                                                                                                               
using builtin fallback logging configuration
35 self tests completed in 0.06 seconds
Starting repository check
Upgrading from v1 hints.31692
Rebuilding sparse info for segment 28722
Failed to read committed index ([Errno 2] No such file or directory: '/mnt/z/borg/borg.repo/data/2/28722')
Segment transaction is    31693
Determined transaction is 31693
Found 26881 segments
Local Exceptionts 93.8%
Traceback (most recent call last):
  File "/home/doc/.local/lib/python3.6/site-packages/borg/archiver.py", line 4175, in main
    exit_code = archiver.run(args)
  File "/home/doc/.local/lib/python3.6/site-packages/borg/archiver.py", line 4107, in run
    return set_ec(func(args))
  File "/home/doc/.local/lib/python3.6/site-packages/borg/archiver.py", line 150, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/home/doc/.local/lib/python3.6/site-packages/borg/archiver.py", line 304, in do_check
    if not repository.check(repair=args.repair, save_space=args.save_space):
  File "/home/doc/.local/lib/python3.6/site-packages/borg/repository.py", line 904, in check
    objects = list(self.io.iter_objects(segment))
  File "/home/doc/.local/lib/python3.6/site-packages/borg/repository.py", line 1317, in iter_objects
    read_data=read_data)
  File "/home/doc/.local/lib/python3.6/site-packages/borg/repository.py", line 1398, in _read
    data = fd.read(length)
OSError: [Errno 71] Protocol error

Platform: Linux staskij 4.14.10-1-ARCH #1 SMP PREEMPT Fri Dec 29 20:17:35 UTC 2017 x86_64
Linux: arch
Borg: 1.1.4  Python: CPython 3.6.4
PID: 785  CWD: /home/doc
sys.argv: ['/home/doc/.local/bin/borg', 'check', '--debug', '-p', '/mnt/z/borg/borg.repo']
SSH_ORIGINAL_COMMAND: None
```

[...]

PR #4940, opened 2020-02-09
borg check: Handle io errors

funkyfuture:
ciao, here's a patch that handles IOErrors when checking a repository. in my case the error is happening when checking a borg repo on a non-raid btrfs volume where's a checksum error upon reading a segment file. the patch also incorporates what my manual interventions would be: find the filepath from the inode that os reported in the kernel log, delete it, and restart the borg check.

i think there was a specific issue addressing this a few months back, but i can't find it. here's a comment where someone encountered the same exception: https://github.com/borgbackup/borg/issues/3509#issuecomment-411107848

i propose to port the patch to the 1.1 maintenance branch to make my life easier.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -5,7 +5,7 @@ import shutil
 import stat
 import struct
 import time
-from binascii import hexlify, unhexlify
+from binascii import unhexlify
 from collections import defaultdict
 from configparser import ConfigParser
 from datetime import datetime
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1526,7 +1526,15 @@ class LoggedIO:
                 size, segment, offset))
         length = size - fmt.size
         if read_data:
-            data = fd.read(length)
+            try:
+                data = fd.read(length)
+            except OSError as err:
+                if err.errno == errno.EIO:
+                    raise IntegrityError(
+                        'I/O error while reading segment file [segment {}, offset {}] : {}'
+                        .format(segment, offset, fd.name)
+                    )
+                raise err
             if len(data) != length:
                 raise IntegrityError('Segment entry data short read [segment {}, offset {}]: expected {}, got {} bytes'.format(
                     segment, offset, length, len(data)))


src/borg/repository.py
```diff
@@ -973,8 +973,16 @@ def report_error(msg):
...
+                    self.io.recover_segment(segment, filepath)
                     objects = list(self.io.iter_objects(segment))
+            except OSError as err:
+                if err.errno == errno.EIO:
+                    report_error("I/O error when reading segment file {}".format(filepath))
+                    objects = []
+                    if repair:
+                        os.remove(filepath)
+                else:
+                    raise
```
ThomasWaldmann: Hmm, some questions / issues:

1. report_error() is called differently here than in the exception handler directly above.
the information from `err` **must** be included in the output message to not lose information.
that might or not include the information which you have in your message, please check.
and both exception handlers must behave / output in a consistent way.

2. the integrityerror exception handler tries to recover as much as possible from the segment file that had such an error inside. there are ususally multiple objects in a segment file and only one of them might have issues and we can rescue all other ones (see recover_segments).

your new OSError handler now assumes that one EIO is enough to just kill the whole segment file.
are there cases where an I/O error might only affect a file partially and we maybe could recover other parts of same file?

funkyfuture: > the information from err must be included in the output message to not lose information.

please see the example traceback that shows less information than the proposed message. afaik the exception instance contains the message, the error number and the filepath, which are all included in the message.

could i use `logger.exception` before calling `report_error`?

> the integrityerror exception handler tries to recover as much as possible from the segment file that had such an error inside.

> your new OSError handler now assumes that one EIO is enough to just kill the whole segment file.

my understanding is that in my case the segment file can't be read at all by the os, because the btrfs kernel module detects a checksum mismatch of an inode. hence i would assume parts of it can't be recovered.

> are there cases where an I/O error might only affect a file partially and we maybe could recover other parts of same file?

oh wow, i just realized how big these segment files are, so yes. i guess that would be w/ `LoggedIO.recover_segment`, right? i'll look into that, maybe even throwing an `IntegrityError` further down the code path. but i would avoid creating further execution frames within loops.

funkyfuture:
i'm pasting the exception that is thrown w/o this patch as refrence:

```
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4501, in main
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4433, in run
    return set_ec(func(args))
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 166, in wrapper
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 327, in do_check
    if not repository.check(repair=args.repair, save_space=args.save_space):
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 923, in check
    objects = list(self.io.iter_objects(segment))
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 1378, in iter_objects
    read_data=read_data)
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 1470, in _read
    data = fd.read(length)
OSError: [Errno 5] Input/output error
```

ThomasWaldmann:
how about str(err) + filepath?

funkyfuture:
i figured to handle the `OSError` in `LoggedIO._read` and escalate an `IntegrityError`. i'm checking it w/ a real-world error.

funkyfuture:
i pushed that take and will report back w/ the outcome of a run. is there a way to tell how much of a segment file was recovered from the terminal output?

src/borg/repository.py
```diff
@@ -1526,7 +1526,15 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags, read_data=Tru
...
         if read_data:
-            data = fd.read(length)
+            try:
+                data = fd.read(length)
+            except OSError as err:
+                if err.errno == errno.EIO:
+                    raise IntegrityError(
+                        'I/O error while reading segment file [segment {}, offset {}] : {}'
+                        .format(segment, offset, fd.name)
+                    )
```
ThomasWaldmann: again: i would like not to lose the information present in str(err).

src/borg/repository.py
```diff
@@ -1526,7 +1526,15 @@ def _read(self, fd, fmt, header, segment, offset, acceptable_tags, read_data=Tru
...
+                raise err
```
ThomasWaldmann: just "raise"?

src/borg/repository.py
funkyfuture: that would be `OSError: [Errno 5] Input/output error`, right? could i just use `logger.exception()` before raising?

src/borg/repository.py
funkyfuture: i assumed that you'd prefer the explicit variant and wasn't sure whether Python 2.7 already knows this syntax.

src/borg/repository.py
ThomasWaldmann: It would be better to have it in one log line.

src/borg/repository.py
ThomasWaldmann: borg is py3 only.

src/borg/repository.py
funkyfuture: i think it would be more effective if you proposed the contents of the message.

ThomasWaldmann: well, you can have a look at other places where error get catched and logged or re-raised.

but in general, using str(err) + " (whatever you want to add)" is a good approach.

hiding error information, telling something that sometimes but not always is a correct error message or suppressing errors completely is bad.

just think of "unknown error" as the extreme or a completely wrong/misleading error msg as worst examples of that.

funkyfuture:
unfortunately that approach (to try to recover the faulty segment file) led to a situation that can't be handled at all. also unfortunately, i have only have a bitmap representation of the output:

borg_check_repair

1. the first line that starts with `Data integrity error …` is an unrelated issue.
1. the second line that starts with these chars is where the file is read, this happened:
    1. `OSError` is caught in `LoggedIO._read`, an `IntegrityError` is raised
    1. that exception is handled in `Repository.check`, the message is printed
1. `LoggedIO.recover_segment` is called from within that last exception handler
1. when `data[4:size]` is evaluated where `data` is a `memoryview` on the faulty file, the interpreter goes bang kaputt

the kernel logs correspondingly `2i` at 100069 and `4` at 100073:

kernel_log

now the question is whether each possible situation that leads to an IOError would fail in recovery attempts in the same way. if so, my proposal would still be to remove faulty segment files. yet, i don't know what the pursued requirements for the repair functionality are and whether my assumption that a deleted segment file's contents would re-appear in subsequently created archives is valid.

haroflow:
We also had problems with `_read` throwing exceptions during borg info, list and check.
Borg version 1.1.11

```
Local Exception
  File "borg/archiver.py", line 4529, in main
  File "borg/archiver.py", line 4461, in run
  File "borg/archiver.py", line 153, in wrapper
  File "borg/helpers.py", line 371, in load
  File "borg/repository.py", line 1100, in get
  File "borg/repository.py", line 387, in get_transaction_id
  File "borg/repository.py", line 384, in check_transaction
  File "borg/repository.py", line 837, in replay_segments
  File "borg/repository.py", line 847, in _update_index
  File "borg/repository.py", line 1408, in iter_objects
  File "borg/repository.py", line 1500, in _read
OSError: raw readinto() returned invalid length 4294967291 (should have been between 0 and 1048576)
```

I wonder if this pull request could help in this case... would borg try to recover the segment instead of crashing?

funkyfuture:
what a year, i'd have thought this pr is much older and i don't remember it well.

beside some information of the underlying error (which fs type, related kernel log entries) it'd be interesting how the patch behaves in your case.

though these are/were actually two or three different proposals. i'm not sure what my last post relates to, my current guess is that was related to code that i didn't push and removed from disk b/c i needed something that works.

also, would removing the corrupted segment file help you? see the first outdated snippet at the beginning of this thread. i did that manually for what the extended error message is helpful. iirc the design question is here whether that removal would be considered as a repair measure and whether it is actually feasible in every situation where an IOError is occuring.

ThomasWaldmann:
please move your problem to a ticket to discuss it.

haroflow:
@funkyfuture Thanks for the response!
I compiled borg with a try/catch around fd._read() just to know what the "corrupted" segments are, and apparently almost all of them showed as problems.
Considering this, I don't think that removing segments would be an option... it's a production backup, with total size around 1.5TB.

The funny thing is, the last backup (2-3 hours earlier) seemed to finish sucessfully, even "borg check" reported no inconsistencies... and that makes me think it might be a hardware problem...

@ThomasWaldmann I'm gathering more info on this, i'll be sure to open a ticket. Thanks!

haroflow:
Sorry for the delay, everything seems ok for now...

Our automated backup was using mount with the `nounix` option, which was missing on our manual mount... That's why we were receiving `OSError: raw readinto() returned invalid length`. Using nounix fixed it.

Besides that, we had hardware bottlenecks, and a missing index.*** file made borg info/list take a very long time, but we were able to show the repository info after that.

Thanks!

ThomasWaldmann:
@haroflow if you want to keep that information, move all your stuff to a ticket. a PR is not the right place for this.

Issue #4953, opened 2020-02-17
Locking is not safe while borg mount (local repo) going to background


alfredo08154711:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes.

## Is this a BUG&nbsp;/ ISSUE report or a QUESTION?

BUG.

## System information. For client/server mode post info for both machines.

All platforms are affected on which `borg mount` works, at latest since the stale lock removal was introduced (version **1.1.0b3**).

I noticed the bug when I looked at the code. The basic problem is known, because it was already fixed by commit `6f94949a3 "migrate locks to child PID when daemonize is used"`, and the fix was released in version **1.1.0rc2** (2017-08-28).

## Describe the problem you're observing.

When `borg mount` (a local repository) is going to background by a double `os.fork()` (`borg.helpers.process.daemonize()`), its process id changes. Since the lock (a shared `borg.locking.Lock` created on the repo to be mounted in this case) identifies its holder by its host and pid, this means that a concurrent borg process would consider the lock stale as soon as the original (parent) borg mount process, which created the lock, dies in favor of the background clone. The fix `6f94949a3 "migrate locks to child PID when daemonize is used"` introduced a lock migration which changes the process identifier to the background pid.

The problem is that this migration is performed when the parent process is already dead. So there is still a little chance that a concurrent borg process deems the lock stale, removes it and acquires another lock before or during the migration.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

The bug can be seen by `strace` with a `borg mount` command. I used a binary installation of borg 1.1.10 under Linux&nbsp;4.15.0-65-generic, Ubuntu&nbsp;18.04.3&nbsp;LTS, with a repo on an ext4 (mounted rw,relatime) file system:

```
$ export BORG_HOST_ID=hostname@1234567890
$ borg init -e none testrepo
$ mkdir testrepo.mnt
$ strace -e trace=%file -f -o borg-mount.log borg mount testrepo testrepo.mnt
```

In the `strace` output (borg-mount.log) one finds the following system calls dealing with file system I/O:

```
01  ...
02  17250 mkdir("/home/thomas/testrepo/lock.exclusive", 0777) = 0
03  17250 openat(AT_FDCWD, "/home/thomas/testrepo/lock.exclusive/hostname@1234567890.17250-0", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 4
[...]

Issue #4425, opened 2019-03-06
with-lock reader instead of exclusive


jarondl:
## Have you checked borgbackup docs, FAQ, and open Github issues?
Yes. This is somewhat related to #3779
## Is this a BUG / ISSUE report or a QUESTION?

Feature request

`borg with-lock` creates an exclusive lock. It also creates an empty commit on the repo (for backwards compatibility). But for my use case (remote replication), I don't need an exclusive lock, I just need a reader's lock. A readers lock will allow me to run consistency checks together with remote replication.

I propose to add a `"--reader"` flag to `with-lock`, that will create a reader lock instead of an exclusive lock. It will also avoid creating the empty commit.

I could try to implement it if the maintainers accept the idea.

infectormp:
related to https://github.com/borgbackup/borg/issues/1035

ThomasWaldmann:
we had different kinds of locks for a while, but after noticing some fundamental issues, there is currently always an exclusive lock.

jarondl:
Can you shed some light on the fundamental issues? I thought there were still two kinds, it's just that with-lock uses the exclusive one?

The naive implementation for the requested feature (without updating documentation) is ff3f2f90b9087abd76e3b4f6c1330d3f806a73c9

ThomasWaldmann:
One problem was that "upgrading" a lock could deadlock.

ThomasWaldmann:
closing this, currently not doable.

alfredo08154711:
I would also like to have exactly this feature. (Yes, I know about the downside of copying repos around: counter reuse in CTR mode. But this is an issue of its own and could or should be improved independently, see #4883, for example.) Aside from that, I would like to have a comprising understanding of the locking problem after I have already prepared two locking related PRs. Maybe I could help.

If you mean the lock upgrade deadlock problem as described in #1220 and (almost) fixed since 1.0.7 (specifying and applying the lock type explicitly and up-front in Repository.open()): I don't see how this could prevent the feature from being implemented in a usable way.

First of all, the upgrade deadlock problem is not really a deadlock, because we use a timeout during lock acquisition. As far as I understand, this was an issue in the described situation, bot not a bug. Since version 1.0.7 the uprade is only done in special cases.

[...]

PR #4954, opened 2020-02-19
Fixed lock migration, issue #4953

alfredo08154711:
See #4953, also #4923.

### DIFF
--- a/src/borg/helpers/process.py
+++ b/src/borg/helpers/process.py
@@ -6,6 +6,8 @@ import shlex
 import signal
 import subprocess
 import sys
+import time
+import traceback
 
 from .. import __version__
 
@@ -13,17 +15,23 @@ from ..platformflags import is_win32, is_linux, is_freebsd, is_darwin
 from ..logger import create_logger
 logger = create_logger()
 
+from ..helpers import EXIT_SUCCESS, EXIT_WARNING, EXIT_SIGNAL_BASE
 
-def daemonize():
-    """Detach process from controlling terminal and run in background
 
-    Returns: old and new get_process_id tuples
-    """
+@contextlib.contextmanager
+def _daemonize():
     from ..platform import get_process_id
     old_id = get_process_id()
     pid = os.fork()
     if pid:
-        os._exit(0)
+        exit_code = EXIT_SUCCESS
+        try:
+            yield old_id, None
+        except _ExitCodeException as e:
+            exit_code = e.exit_code
+        finally:
+            logger.debug('Daemonizing: Foreground process (%s, %s, %s) is now dying.' % old_id)
+            os._exit(exit_code)
     os.setsid()
     pid = os.fork()
     if pid:
@@ -31,13 +39,101 @@ def daemonize():
     os.chdir('/')
     os.close(0)
     os.close(1)
-    os.close(2)
     fd = os.open(os.devnull, os.O_RDWR)
     os.dup2(fd, 0)
     os.dup2(fd, 1)
-    os.dup2(fd, 2)
     new_id = get_process_id()
-    return old_id, new_id
+    try:
+        yield old_id, new_id
+    finally:
+        # Close / redirect stderr to /dev/null only now
+        # for the case that we want to log something before yield returns.
+        os.close(2)
+        os.dup2(fd, 2)
+
+
+def daemonize():
+    """Detach process from controlling terminal and run in background
+
+    Returns: old and new get_process_id tuples
+    """
+    with _daemonize() as (old_id, new_id):
+        return old_id, new_id
+
+
+@contextlib.contextmanager
+def daemonizing(*, timeout=5):
+    """Like daemonize(), but as context manager.
+
+    The with-body is executed in the background process,
+    while the foreground process survives until the body is left
+    or the given timeout is exceeded. In the latter case a warning is
+    reported by the foreground.
+    Context variable is (old_id, new_id) get_process_id tuples.
+    An exception raised in the body is reported by the foreground
+    as a warning as well as propagated outside the body in the background.
+    In case of a warning, the foreground exits with exit code EXIT_WARNING
+    instead of EXIT_SUCCESS.
+    """
+    with _daemonize() as (old_id, new_id):
+        if new_id is None:
+            # The original / parent process, waiting for a signal to die.
+            logger.debug('Daemonizing: Foreground process (%s, %s, %s) is waiting for background process...' % old_id)
+            exit_code = EXIT_SUCCESS
+            # Indeed, SIGHUP and SIGTERM handlers should have been set on archiver.run(). Just in case...
+            with signal_handler('SIGINT', raising_signal_handler(KeyboardInterrupt)), \
+                 signal_handler('SIGHUP', raising_signal_handler(SigHup)), \
+                 signal_handler('SIGTERM', raising_signal_handler(SigTerm)):
+                try:
+                    if timeout > 0:
+                        time.sleep(timeout)
+                except SigTerm:
+                    # Normal termination; expected from grandchild, see 'os.kill()' below
+                    pass
+                except SigHup:
+                    # Background wants to indicate a problem; see 'os.kill()' below,
+                    # log message will come from grandchild.
+                    exit_code = EXIT_WARNING
+                except KeyboardInterrupt:
+                    # Manual termination.
+                    logger.debug('Daemonizing: Foreground process (%s, %s, %s) received SIGINT.' % old_id)
+                    exit_code = EXIT_SIGNAL_BASE + 2
+                except BaseException as e:
+                    # Just in case...
+                    logger.warning('Daemonizing: Foreground process received an exception while waiting:\n' +
+                                   ''.join(traceback.format_exception(e.__class__, e, e.__traceback__)))
+                    exit_code = EXIT_WARNING
+                else:
+                    logger.warning('Daemonizing: Background process did not respond (timeout). Is it alive?')
+                    exit_code = EXIT_WARNING
+                finally:
+                    # Don't call with-body, but die immediately!
+                    # return would be sufficient, but we want to pass the exit code.
+                    raise _ExitCodeException(exit_code)
+
+        # The background / grandchild process.
+        sig_to_foreground = signal.SIGTERM
+        logger.debug('Daemonizing: Background process (%s, %s, %s) is starting...' % new_id)
+        try:
+            yield old_id, new_id
+        except BaseException as e:
+            sig_to_foreground = signal.SIGHUP
+            logger.warning('Daemonizing: Background process raised an exception while starting:\n' +
+                           ''.join(traceback.format_exception(e.__class__, e, e.__traceback__)))
+            raise e
+        else:
+            logger.debug('Daemonizing: Background process (%s, %s, %s) has started.' % new_id)
+        finally:
+            try:
+                os.kill(old_id[1], sig_to_foreground)
+            except BaseException as e:
+                logger.error('Daemonizing: Trying to kill the foreground process raised an exception:\n' +
+                             ''.join(traceback.format_exception(e.__class__, e, e.__traceback__)))
+
+
+class _ExitCodeException(BaseException):
+    def __init__(self, exit_code):
+        self.exit_code = exit_code
 
 
 class SignalException(BaseException):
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Rebased on current master, fixed pep8 issues, no other changes.

ThomasWaldmann:
I'll rebase this again onto master now...

alfredo08154711:
I'll look at the conflicts soon.

ThomasWaldmann:
@alfredo08154711 Please first update your local master branch, so it includes your now merged locking fixes, then rebase this PR on top of that (and try to resolve the conflicts).

alfredo08154711:
In the rebase I have combined the logic of the conflicting changes in testsuite.BaseTestCase.fuse_mount() and re-run the tests. Please go ahead.

src/borg/testsuite/archiver.py
```diff
@@ -2503,6 +2503,94 @@ def test_fuse_mount_options(self):
...
+                    'exception': None,
+                    'exception.extr_tb': None,
+                    'after': {
+                        'old_id_alive': None,
+                        'new_id_alive': None}}
+                try:
+                    with open(assert_data_file, 'wb') as _out:
+                        pickle.dump(assert_data, _out)
+                except:
+                    pass
```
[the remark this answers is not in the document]
alfredo08154711: We are writing to a file, as we already said: during I/O, everything can happen at any time. For example, disk full during testing&mdash;very unlikely, but possible.

If it happens (I tried it), you won't see the stack trace, because it happens in the borg mount process daemonized in testsuite.fuse_mount(). What you'll see then is the timeout raised by testsuite.wait_for_mount() (to be continued...)

src/borg/testsuite/archiver.py
```diff
@@ -2503,6 +2503,94 @@ def test_fuse_mount_options(self):
...
+                    assert_data['exception.extr_tb'] = traceback.extract_tb(e.__traceback__)
+                finally:
+                    assert_data['after'].update({
+                        'old_id_alive': platform.process_alive(*old_id),
+                        'new_id_alive': platform.process_alive(*new_id)})
+                    try:
+                        with open(assert_data_file, 'wb') as _out:
+                            pickle.dump(assert_data, _out)
+                    except:
+                        pass
```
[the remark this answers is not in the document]
alfredo08154711: See above.

src/borg/testsuite/archiver.py
```diff
@@ -2503,6 +2503,94 @@ def test_fuse_mount_options(self):
...
+                        'new_id_alive': None}}
+                try:
+                    with open(assert_data_file, 'wb') as _out:
+                        pickle.dump(assert_data, _out)
+                except:
+                    pass
+                try:
+                    return migrate_lock(self, old_id, new_id)
+                except BaseException as e:
+                    assert_data['exception'] = e
```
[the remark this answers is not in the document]
alfredo08154711: Yes. If you replace `return migrate_lock(self, old_id, new_id)` with `raise Exception('Dummy exception')`, you'll get
```
Lock.migrate_lock(): assert_data = {'num_calls': 1, 'old_id': ('lubutom@181261324266631', 8350, 0), 'new_id': ('lubutom@181261324266631', 8354, 0), 'before': {'old_id_alive': False, 'new_id_alive': True}, 'exception': Exception('Dummy exception',), 'exception.extr_tb': [<FrameSummary file /home/thomas/Development/borgbackup/git-myfork/src/borg/testsuite/archiver.py, line 2375 in wrapper>], 'after': {'old_id_alive': False, 'new_id_alive': True}}.
Lock.migrate_lock() raised an exception:
  File "/home/thomas/Development/borgbackup/git-myfork/src/borg/testsuite/archiver.py", line 2375, in wrapper
    raise Exception('Dummy exception')
Exception: Dummy exception
```

src/borg/testsuite/archiver.py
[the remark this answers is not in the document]
ThomasWaldmann: ok, thanks!

src/borg/testsuite/archiver.py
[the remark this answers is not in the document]
ThomasWaldmann: ok

src/borg/helpers/process.py
```diff
@@ -6,38 +6,132 @@
...
+                        time.sleep(timeout)
+                except SigTerm:
+                    # Normal termination.
+                    pass
+                except KeyboardInterrupt:
+                    # Manual termination.
+                    logger.debug('Daemonizing: Foreground process (%s, %s, %s) received SIGINT.' % old_id)
+                except SigHup:
+                    # Background wants to indicate a problem.
+                    exit_code = EXIT_WARNING
```
ThomasWaldmann, line 98: i recently fixed exit codes (#5161) with https://github.com/borgbackup/borg/pull/5242.

guess this needs to be considered here also.

src/borg/helpers/process.py
```diff
@@ -6,38 +6,132 @@
...
+
+    Returns: old and new get_process_id tuples
+    """
+    with _daemonize() as (old_id, new_id):
+        return old_id, new_id
+
+
+@contextlib.contextmanager
+def daemonizing(*, timeout=5):
+    """Like daemonize(), but as context manager.
```
ThomasWaldmann, line 66: this is doing quite a bit more than previously.

can you edit the commit comment and write a bit about why this is needed, what problem is fixed by the new `daemonizing`?

src/borg/testsuite/__init__.py
```diff
@@ -238,30 +238,55 @@ def _assert_dirs_equal_cmp(self, diff, ignore_flags=False, ignore_xattrs=False,
...
+                    os._exit(0)
+                # The grandchild process.
+                try:
+                    self.cmd(*args, fork=False, **kwargs)  # borg mount not spawning.
+                finally:
+                    # This should never be reached, since it daemonizes,
+                    # and the grandchild process exits before cmd() returns.
+                    # However, just in case...
+                    print('Fatal: borg mount did not daemonize properly. Force exiting.',
+                            file=sys.stderr, flush=True)
```
ThomasWaldmann, line 278: looks like the 2nd line is not indented properly. strange that the pep8 checker did not complain.

guess the `f` should be under the `'`.

ThomasWaldmann: also, is print() the right way here, not logging.error()?

src/borg/testsuite/archiver.py
```diff
@@ -2503,6 +2503,94 @@ def test_fuse_mount_options(self):
...
+            self.cmd('init', '--encryption=none', self.repository_location)
+            self.create_src_archive('arch')
+            mountpoint = os.path.join(self.tmpdir, 'mountpoint')
+            # In order that the decoration is kept for the borg mount process, we must not spawn, but actually fork;
+            # not to be confused with the forking in borg.helpers.daemonize() which is done as well.
+            with self.fuse_mount(self.repository_location, mountpoint, os_fork=True):
+                pass
+            with open(assert_data_file, 'rb') as _in:
+                assert_data = pickle.load(_in)
+            print('\nLock.migrate_lock(): assert_data = %s.' % (assert_data), file=sys.stderr, flush=True)
```
ThomasWaldmann, line 2569: the % operator needs a tuple as argument and I guess with `(assert_data)` you wanted to give one.

but that is not a 1-tuple, for that it would need to be `(assert_data, )`.

also, you could use `%r` instead of `%s` because you want a repr of that.

src/borg/testsuite/archiver.py
```diff
@@ -2503,6 +2503,94 @@ def test_fuse_mount_options(self):
...
+            assert assert_data['num_calls'] == 1, "Lock.migrate_lock() must be called exactly once."
+            assert exception is None, "Lock.migrate_lock() may not raise an exception."
+
+            assert_data_before = assert_data['before']
+            assert assert_data_before['old_id_alive'], "old_id must be alive (=may not be stale) when calling Lock.migrate_lock()."
+            assert assert_data_before['new_id_alive'], "new_id must be alive (=may not be stale) when calling Lock.migrate_lock()."
+
+            assert_data_after = assert_data['after']
+            assert assert_data_after['old_id_alive'], "old_id must be alive (=may not be stale) when Lock.migrate_lock() has returned."
+            assert assert_data_after['new_id_alive'], "new_id must be alive (=may not be stale) when Lock.migrate_lock() has returned."
```
ThomasWaldmann, line 2589: s/may not/must not/g ?

alfredo08154711:
O.k., thank you for your feedback. Again, it will take some hours. It's half a year ago that I coded this&hellip;

ThomasWaldmann:
@alfredo08154711 thanks for working on this!

Idea for long term:

It is a bit embarassing that we have to write our own locking code and in fact, even write it 3 times:
- first attempt was by attic author (using posix locking, which had issues iirc on windows)
- second attempt was by me (using mkdir, but you found a race condition with that)
- third attempt by you (this PR, using directory rename)

So the idea is to move this to a separate locking library (back then, when I did the second attempt, I looked at at least 2 or 3 existing locking libraries and none of them was usable for some reason or the other).

This is independent of whether we will later bundle/vendor that libraries' code into the borg codebase or not.

src/borg/testsuite/__init__.py
alfredo08154711: Indentation: I didn't know. As you can see (the other pep8 issues as well), I indented by 8 spaces in these cases. This is used by many style guides and code formatters in the Java world. :-)

I used print() as in the other testcases.&mdash;To be honest, I tried to understand the logger stuff, but without much success. I only tried to do it the way I found in the code.

In other words, I should do it as it should be done; I have no opinion on it.

src/borg/helpers/process.py
alfredo08154711: I agree in the KeyboardInterrupt case, but not in the SigHup case. SigTerm and SigHup are used here as communication / are expected from the background process. I will add this in the comments, and I will amend the KeyboardInterrupt case.

src/borg/helpers/process.py
alfredo08154711: I'll try.

src/borg/testsuite/archiver.py
alfredo08154711: Apparently, I didn't fully understand this tuple stuff.

src/borg/testsuite/archiver.py
alfredo08154711: Yes.

alfredo08154711:
I hope it's better now. Please have a look.

"Idea for long term": Your mkdir solution was perfectly working. It was only broken by inception of the automatic stale lock killing, because this literally violates the old locking "protocol." So in my opinion it is not that embarassing as it feels. "Shit happens"&mdash;this is life. However, there is nothing to be said against your idea.

I would also talk about an idea. It is why I looked at locking at all. Please see my comment on #4425 (written on 24 Feb).

src/borg/helpers/process.py
ThomasWaldmann: correct! :-)

src/borg/testsuite/__init__.py
ThomasWaldmann: yeah, guess the checker accepts indentation by 4n as well as "same" indentation within the parens.

oh, seems i missed this is test code. yeah, guess print() is better then.

src/borg/testsuite/archiver.py
ThomasWaldmann: (x) is the same as x

(x, ) is a 1-tuple and x its first element.

notable: not the `()` makes a tuple, but the `,` is the tuple operator.

ThomasWaldmann:
Looks good. Do you want to squash some of the changesets before merging?

alfredo08154711:
I have no opinion on that. It should be done according to the customs of the project. I am not so well versed in the use of git; rebasing causes me headaches every time, so feel free to make any changes you deem appropriate. The main commit comment is that of 66d4577, of course.

src/borg/testsuite/archiver.py
alfredo08154711: Thanks.

ThomasWaldmann:
OK, I'll squash and merge. Thanks!

ThomasWaldmann:
\o/

PR #4955, opened 2020-02-21
Allow osx builds to fail

finefoot:
At this time, OSX builds don't seem to work well on Travis CI. This results in failed tests for each pull request or merge commit. Wouldn't it make sense to use `fast_finish` and `allow_failures` then to specifically allow those builds to fail if they're currently being ignored anyway? See https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail for documentation. Then, if a justified failed test happens, it would show up faster and also wouldn't be disguised by the in any case failing OSX builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
         - $HOME/.cache/pip
 
 matrix:
+    fast_finish: true
     include:
         - python: 3.5
           os: linux
@@ -34,6 +35,15 @@ matrix:
           os: linux
           dist: xenial
           env: TOXENV=flake8
+        - language: generic
+          os: osx
+          osx_image: xcode8.3
+          env: TOXENV=py35
+        - language: generic
+          os: osx
+          osx_image: xcode8.3
+          env: TOXENV=py37
+    allow_failures:
         - language: generic
           os: osx
           osx_image: xcode8.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,14 +35,6 @@ matrix:
           os: linux
           dist: xenial
           env: TOXENV=flake8
-        - language: generic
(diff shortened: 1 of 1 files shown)


finefoot:
By the way, these changes are already in effect for this pull request as you can see by looking at its Travis CI log: https://travis-ci.org/borgbackup/borg/builds/653412916

.travis.yml
```diff
@@ -42,6 +43,15 @@ matrix:
...
           env: TOXENV=py37
+    allow_failures:
+        - language: generic
+          os: osx
+          osx_image: xcode8.3
+          env: TOXENV=py35
+        - language: generic
+          os: osx
+          osx_image: xcode8.3
+          env: TOXENV=py37
```
ThomasWaldmann: isn't this duplicating the entries we already have above?

finefoot: Yes, that's how I remembered it. See 9c4bbdd1329e866f5517e35150ea02501e87de37 where I removed the entries from above, then Travis CI doesn't do the builds from `allow_failures` at all: https://travis-ci.org/borgbackup/borg/builds/653877191

However, I agree, it's not nice to have the osx entries both under `include` as well as under `allow_failures`, so with a44c545278000d3b3ba3ac72b315fde2f95f17a7 I just universally marked all builds with `os: osx` as allowed failures. Now the jobs run as expected again: https://travis-ci.org/borgbackup/borg/builds/653879285 And since `os: osx` is also the reason why the builds even fail in the first place, I think universally marking `os: osx` should be fine, too. (At least for the time being.)

ThomasWaldmann:
thanks! can you backport it to 1.1-maint?

Issue #4905, opened 2020-01-02
platforms without sync_file_range


ThomasWaldmann:
Moved from a comment in #4901 by @willsALMANJ:

When I try to run `borg init`, I get the error `ImportError: dlopen failed: cannot locate symbol "sync_file_range" refereced by ...borg/platform/linux.cpython-38.so`. Looking in `usr/include/fcntl.h`, I found

```
#if __ANDROID_API__ >= 26
int sync_file_range(...);
#endif
```

My phone uses Android 7 which corresponds to __ANDROID_API__ level 24, so it seems that I need something with at least Android 8 to test. I could try testing in an Android VM if I can't find a newer phone to test with and someone else doesn't test it.

By the way, I get the same error when I use the prebuilt borgbackup Termux package.

To do the test, I did `pkg install liblz4 zstd openssl libacl pkg-config clang` and then `pip install pkgconfig cython` in a virtual environment before running `python setup.py install`. That all appeared to build fine. I discovered that for borg 1.10 it is necessary to set the paths for the required shared libraries with the environment variables like `BORG_OPENSSL_PREFIX` like https://github.com/termux/termux-packages/pull/4644 does, but on master the new usage of `pkgconfig` in `setup.py` works. One other note -- `clang` was needed above for `cython`; in the version of borg packaged for Termux a released version of borg is used which does not need `cython`, so that is why `cython` and `clang` are not referenced in the Termux package's `build.sh`.

ThomasWaldmann:
We also had a sync_file_range issue on Windows Subsystem for Linux (WSL).

Maybe the way master branch deals with that (env var) should be also offered on 1.1-maint (but keeping the automated handler for WSL), so it could be also used on old android.

ThomasWaldmann:
@willsALMANJ how does `uname -r` look like on your android phone?

is there an easy way to check the ANDROID_API level at runtime?

willsALMANJ:
> We also had a sync_file_range issue on Windows Subsystem for Linux (WSL).
Ah, I hadn't noticed this. So for the testing I was doing based on master, I can use `BORG_WORKAROUNDS=basesyncfile` and it should work.

[...]

PR #4965, opened 2020-03-01
Move sync_file_range to its own extension

willsALMANJ:
Here is one way of addressing #4905. I moved the `cdef extern` import of `sync_file_range` into its own extension and imported it in a try block. If the import fails, borg falls back to using the BaseSyncFile class like it does with BORG_WORKAROUNDS=basesyncfile. This way BORG_WORKAROUNDS is not required when there is no sync_file_range support. An alternative would be to still require BORG_WORKAROUNDS.


ThomasWaldmann:
Thanks!

ThomasWaldmann:
Can you backport this to 1.1-maint branch and do a PR for that?

Issue #4901, opened 2019-12-27
AttributeError on borg init when run with Python without hard-link support


willsALMANJ:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V)

1.1.10

#### Operating system (distribution) and version.

Android / Termux

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

`borg init`

## Describe the problem you're observing.

Thanks to https://github.com/termux/termux-packages/pull/4644, borg is now available in Termux. One issue discovered (and worked around with a patch) might be worth addressing in borg. `borg init` choked trying to do `os.link` because the Termux Python is built without hard-link support. Rather than an `OSError` that `repository.py` has been set up to handle, an `AttributeError` is generated because the `link` function does not even exist in the `os` module.

I think this could be addressed by making the `except` clause in `repository.py` slightly more complex. The benefit would be that `pip install borgbackup` should then also work if all of the required dependencies are installed.

#### Include any warning/errors/backtraces from the system logs

This is copied from the PR linked above:

```
Local Exception
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.10-py3.8-linux-x86_64.egg/borg/archiver.py", line 4501, in main
    exit_code = archiver.run(args)
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.10-py3.8-linux-x86_64.egg/borg/archiver.py", line 4433, in run
    return set_ec(func(args))
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.10-py3.8-linux-x86_64.egg/borg/archiver.py", line 166, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.10-py3.8-linux-x86_64.egg/borg/archiver.py", line 280, in do_init
    key = key_creator(repository, args)
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.10-py3.8-linux-x86_64.egg/borg/crypto/key.py", line 110, in key_creator
    return key.create(repository, args)
[...]

PR #4970, opened 2020-03-01
Move sync_file_range to its own extension

willsALMANJ:
This is the equivalent of #4965 targeted to 1.1-maint instead of master.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ item_source = 'src/borg/item.pyx'
 checksums_source = 'src/borg/algorithms/checksums.pyx'
 platform_posix_source = 'src/borg/platform/posix.pyx'
 platform_linux_source = 'src/borg/platform/linux.pyx'
+platform_syncfilerange_source = 'src/borg/platform/syncfilerange.pyx'
 platform_darwin_source = 'src/borg/platform/darwin.pyx'
 platform_freebsd_source = 'src/borg/platform/freebsd.pyx'
 msgpack_packer_source = 'src/borg/algorithms/msgpack/_packer.pyx'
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
didn't work on linux py34 and on macOS py35, see travis-ci logs.

willsALMANJ:
Very strange errrors. py34 flake8 failed complaining about Cython but py34 test suite run passed? For macOS it complains about not being able to import "ContextManager" -- I'm not sure where this capitalization of ContextManager is coming from.

I have now run the test suite on this branch on my phone. Most tests pass, but a few fail. Mostly they seem harmless, but I haven't had a chance to study all of them. I will try to summarize them later.

willsALMANJ:
A bunch of the archiver tests fail with a non-zero return code with the message "Permission denied when setting extended attribute security.selinux". One is example is [this point in test_return_codes](https://github.com/willsALMANJ/borg/blob/4327922bdd46b4c11b34a6b8b3e5aabcb41879c5/src/borg/testsuite/archiver.py#L158) where it is trying to extract a repo. We had encountered this issue before [here](https://github.com/borgbackup/borg/issues/4901#issuecomment-572409069). The problem seems to be that Termux/Android uses selinux but my user does not have permission to modify selinux attributes. Is there a workaround to ignore errors setting selinux attributes that I could use to see if the tests otherwise pass?

ThomasWaldmann:
Merged this. The test failures seemed to be unrelated to the changes in this PR.

willsALMANJ:
Just an update -- Termux updated to borg 1.1.11 this week and now with with this change and #4903 `apt install borgbackup` gives me a working borg command on Android 7.

PR #4979, opened 2020-03-03
fix setup.py for syncfilerange

ThomasWaldmann:


ThomasWaldmann:
this needs checking whether it needs to be forward ported to master.

ThomasWaldmann:
checking this now...

ThomasWaldmann:
Hmm, the building in master is rather different and I don't currently see any changes needed for master. 

If somebody else wants to have a look, this would be very helpful.

ThomasWaldmann:
@willsALMANJ can you have a look?

willsALMANJ:
Hmm, sorry that I missed these things the first time (I guess the only affect building a distribution, and I only tested with `pip install -e .`?). setup.py is pretty different for master (I am looking forward to it being released because I think "pip install borg" will work in Termux without the BORG_*_PREFIX variables). [This line](https://github.com/borgbackup/borg/blame/master/setup.py#L239) looks funny to me. It is missing syncfilerange_ext but it also seems redundant with the next line which does include syncfilerange_ext.

ThomasWaldmann:
Yeah, looks redundant. Guess we'll try without. :)

ThomasWaldmann:
see #5003.

PR #4985, opened 2020-03-06
xxhash>=0.7.3 provides a pkgconfig file

bket:
https://github.com/Cyan4973/xxHash/pull/324

As advertised, xxhash is picked up by py-pkgconfig.


ThomasWaldmann:
Great that you added pkgconfig to xxhash, thanks!
And they even did a release directly afterwards, so we have that now. \o/

Guess I need to update the builtin bundled code here again from 0.7.2 to 0.7.3.

PR #4987, opened 2020-03-06
Adding comments and explanations to Travis config and install script, improving OS X builds

finefoot:
As suggested on IRC, simply going ahead and posting this as a pull request to make it easier to discuss possible improvements.

Intrigued by the exceeding timeouts of OS X builds and wanting to learn a little bit more about the build process, I've been playing around with the CI setup and made a few changes to `.travis.yml` and `.travis/install.sh`. Mostly tidying up, writing comments on what I learned, keeping it POSIX-compliant and verified with [ShellCheck](https://www.shellcheck.net/).

Turns out, OS X builds now consistently pass (within about half an hour), meeting the Travis CI maximum build time and not timing out. :tada:  (I hope I didn't break something without noticing.)

Possibly, the reasons for the previous timeouts have been some time-consuming remnants from using xcode6.4 which didn't get removed when switching to xcode8.3 with https://github.com/borgbackup/borg/commit/c7503f651853ba1eac577338a0999fdee390b9dd. I'm not sure though. 🤷

However, using images later than xcode8.3 results in tests regarding FUSE simplify freezing (`testsuite/archiver.py::ArchiverTestCase::test_fuse`, for example) and the checks fail with 10 minutes of no output received. This happens due to `osxfuse` being a kernel extension and therefore requiring explicit activation due to SIP (System Integrity Protection) in recent versions of OS X - affected versions being used for image > xcode8.3. Travis CI seems to be aware of this issue, but no changes have been made, yet. Here is one of the most recent posts: https://travis-ci.community/t/allow-disabling-sip-on-macos-10-13/3253 (But you'll find many more.)

This means, at the moment, if you want to run the full test suite on OS X, you need to use xcode8.3. Excluding FUSE tests by disabling the installation of FUSE requirements with pip (namely `llfuse`) via `SKIPFUSE=true` is a way around this and shows that everything else works just fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,19 +7,19 @@ cache:
 matrix:
     fast_finish: true
     include:
-        - python: 3.5
+        - python: "3.5"
           os: linux
           dist: trusty
           env: TOXENV=py35
-        - python: 3.6
+        - python: "3.6"
           os: linux
           dist: trusty
           env: TOXENV=py36
-        - python: 3.7
+        - python: "3.7"
           os: linux
           dist: xenial
           env: TOXENV=py37
-        - python: 3.8
+        - python: "3.8"
           os: linux
           dist: xenial
           env: TOXENV=py38
@@ -31,22 +31,22 @@ matrix:
           os: linux
           dist: xenial
           env: TOXENV=py38
-        - python: 3.5
+        - python: "3.5"
           os: linux
           dist: xenial
           env: TOXENV=flake8
         - language: generic
           os: osx
-          osx_image: xcode8.3
+          osx_image: xcode8.3  # This is the latest working xcode image with osxfuse compatibility; later images come with an OS X version which doesn't allow kernel extensions
           env: TOXENV=py35
         - language: generic
           os: osx
-          osx_image: xcode8.3
-          env: TOXENV=py37
+          osx_image: xcode11.3
+          env: TOXENV=py37 SKIPFUSE=true
     allow_failures:
-        - os: osx
+        - os: osx  # OS X builds often take too long and time out, even though tests don't actually fail
 
-before_install:
+before_install: # Abort installation and don't run tests for pull requests if commit only changed the docs
 - |
     test $TRAVIS_EVENT_TYPE != "pull_request" || {
         echo Checking whether $TRAVIS_COMMIT_RANGE changed only docs
@@ -56,15 +56,11 @@ before_install:
         }
     }
 
-install:
-    - git fetch --unshallow --tags
-    - ./.travis/install.sh
+install: ./.travis/install.sh
 
-script:
-    - ./.travis/run.sh
+script: ./.travis/run.sh
 
-after_success:
-    - ./.travis/upload_coverage.sh
+after_success: ./.travis/upload_coverage.sh
 
 notifications:
     irc:
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -3,54 +3,73 @@
 set -e
 set -x
 
-if [[ "$(uname -s)" == 'Darwin' ]]; then
-    export HOMEBREW_NO_AUTO_UPDATE=1
-    export HOMEBREW_LOGS=~/brew-logs
-    export HOMEBREW_TEMP=~/brew-temp
-    mkdir $HOMEBREW_LOGS
-    mkdir $HOMEBREW_TEMP
-    brew update > /dev/null
-    brew cleanup > /dev/null  # do this here, so it won't automatically trigger in the middle of other stuff
-    brew outdated pkg-config || brew upgrade pkg-config
-    # do NOT update openssl 1.0.x, brew will also update a lot of dependent pkgs (and their dependencies) then!
-    #brew outdated openssl || brew upgrade openssl
-    export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig:$PKG_CONFIG_PATH"
-    brew install readline
-    export PKG_CONFIG_PATH="/usr/local/opt/readline/lib/pkgconfig:$PKG_CONFIG_PATH"
-    brew install zstd
-    brew install lz4
-    brew install xz  # required for python lzma module
-    brew install Caskroom/cask/osxfuse
+# Travis clones with depth=50 which might be too shallow for git describe, see https://github.com/pypa/setuptools_scm/issues/93
+git fetch --unshallow --tags
 
-    brew outdated pyenv || brew upgrade pyenv
-    if which pyenv > /dev/null; then
-        eval "$(pyenv init -)"
+if [ "${TRAVIS_OS_NAME}" = "osx" ]
+then
+
+    # Update brew itself
+    export HOMEBREW_NO_AUTO_UPDATE=1  # Auto-updating everything would take too much time
+    brew update
+    brew cleanup  # Preempt possible scheduled clean-up so it doesn't clutter the log later
+
+    # Install and/or upgrade dependencies
+    #brew install zstd || brew upgrade zstd  # Installation takes very long due to CMake dependency and isn't necessary for the tests as borg comes bundled with zstd anyway
+    brew install lz4 || brew upgrade lz4
+    brew install xz || brew upgrade xz  # Required for Python lzma module
+    brew install Caskroom/cask/osxfuse || brew upgrade Caskroom/cask/osxfuse  # Required for Python llfuse module
+    brew install pyenv || brew upgrade pyenv
+
+    # Configure pkg-config to use OpenSSL 1.1 from Homebrew
+    export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig:${PKG_CONFIG_PATH}"
+
+    # Configure pyenv with Python version according to TOXENV
+    eval "$(pyenv init -)"
+    if [ "${TOXENV}" = "py35" ]
+    then
+        pyenv install 3.5.3  # Minimum version for OpenSSL 1.1.x
+        pyenv global 3.5.3
+    elif [ "${TOXENV}" = "py37" ]
+    then
+        pyenv install 3.7.0
+        pyenv global 3.7.0
+    else
+        printf '%s\n' "Unexpected value for TOXENV environment variable"
+        exit 1
     fi
-
-    case "${TOXENV}" in
-        py35)
-            pyenv install 3.5.3  # minimum for openssl 1.1.x
-            pyenv global 3.5.3
-            ;;
-        py37)
-            pyenv install 3.7.0
-            pyenv global 3.7.0
-            ;;
-    esac
     pyenv rehash
-    python -m pip install --user virtualenv
-else
-    pip install virtualenv
+
+elif [ "${TRAVIS_OS_NAME}" = "linux" ]
+then
+
+    # Install dependencies
     sudo apt-get update
     sudo apt-get install -y pkg-config fakeroot
-    #sudo apt-get install -y liblz4-dev  # too old on trusty and xenial
-    #sudo apt-get install -y libzstd-dev  # too old on trusty and xenial
+    #sudo apt-get install -y liblz4-dev  # Too old on trusty and xenial, but might be useful in future versions
+    #sudo apt-get install -y libzstd-dev  # Too old on trusty and xenial, but might be useful in future versions
     sudo apt-get install -y libacl1-dev
-    sudo apt-get install -y libfuse-dev fuse  # optional, for FUSE support
+    sudo apt-get install -y libfuse-dev fuse  # Required for Python llfuse module
+
+else
+
+    printf '%s\n' "Unexpected value for TRAVIS_OS_NAME environment variable"
+    exit 1
+
 fi
 
+# Setup and activate virtual environment
+python -m pip install virtualenv
 python -m virtualenv ~/.venv
 source ~/.venv/bin/activate
+
+# Recent versions of OS X don't allow kernel extensions which makes the osxfuse tests fail; those versions are marked with SKIPFUSE=true in .travis.yml
+if [ "${SKIPFUSE}" = "true" ]
+then
+    truncate -s 0 requirements.d/fuse.txt
+fi
+
+# Install requirements
 pip install -r requirements.d/development.txt
 pip install codecov
 python setup.py --version


ThomasWaldmann:
Cool, let's see how all the xcode versions will behave.

Will take a while because we only have few parallel macOS travis jobs.

.travis/install.sh
```diff
@@ -3,54 +3,80 @@
...
-    brew install zstd
+# Abort installation and don't run tests for pull requests if commit only changed the docs
+if [ "${TRAVIS_EVENT_TYPE}" = "pull_request" ]
+then
+    printf '%s\n' "Checking whether ${TRAVIS_COMMIT_RANGE} changed only docs"
+    git diff --name-only "${TRAVIS_COMMIT_RANGE}" | grep --quiet --invert-match --extended-regexp '(AUTHORS|README\.rst|^(docs)/)' || {
+        printf '%s\n' "Only docs were updated, stopping build process."
+        exit 0
+    }
+fi
```
ThomasWaldmann: does this actually work when invoked from here?

.travis/install.sh
```diff
@@ -3,54 +3,80 @@
...
+    brew install xz  # Required for Python lzma module
+    brew install Caskroom/cask/osxfuse  # Required for Python llfuse module
+    brew install pyenv || brew upgrade pyenv  # Make sure to use most recent pyenv
 
-    brew outdated pyenv || brew upgrade pyenv
-    if which pyenv > /dev/null; then
-        eval "$(pyenv init -)"
-    fi
+    # Configure pkg-config to use OpenSSL 1.1 from Homebrew
+    export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig:${PKG_CONFIG_PATH}"
```
ThomasWaldmann: shouldn't we install pkg-config at first, to make sure we have it?

then readline (and adjust PKG_CONFIG_PATH), then openssl (and adjust PKG_CONFIG_PATH), then pyenv?

also make sure we have zstd.

.travis/install.sh
```diff
@@ -3,54 +3,80 @@
...
-    pip install virtualenv
+
+elif [ "${TRAVIS_OS_NAME}" = "linux" ]
+then
+
+    # Install dependencies
     sudo apt-get update
     sudo apt-get install -y pkg-config fakeroot
-    #sudo apt-get install -y liblz4-dev  # too old on trusty and xenial
-    #sudo apt-get install -y libzstd-dev  # too old on trusty and xenial
```
ThomasWaldmann: maybe keep at least the information in a comment.

.travis.yml
```diff
@@ -31,40 +31,50 @@ matrix:
...
+          osx_image: xcode10.3
+          env: TOXENV=py37 SKIPFUSE=true
+        - language: generic
+          os: osx
+          osx_image: xcode11.3
+          env: TOXENV=py35 SKIPFUSE=true
+        - language: generic
+          os: osx
+          osx_image: xcode11.3
+          env: TOXENV=py37 SKIPFUSE=true
```
ThomasWaldmann: and we take this to test "new world".

we only have 2 parallel macOS builds, so guess we ONLY take the old world and new world test runs and remove all the intermediate ones again.

.travis/install.sh
finefoot: I'm curious, why do you even install zstd if it seemingly doesn't matter for the tests? Same for readline and its PKG_CONFIG_PATH. Didn't look deeper into it, though.

.travis/install.sh
finefoot: I don't know why it wouldn't? :thinking: What are you concerned about? I can run an isolated test in a fresh repo if you'd like to be sure.

.travis.yml
finefoot: I just added them because with `fast_finish: true` it doesn't really matter if the `os: osx` tests take 3 hours. So you'd like to keep only `xcode8.3` and the most recent one?

.travis/install.sh
finefoot: Can you explain why the comments are there in the first place?

.travis.yml
```diff
@@ -31,40 +31,50 @@ matrix:
...
-        - python: 3.5
+        - python: "3.5"
           os: linux
           dist: xenial
           env: TOXENV=flake8
         - language: generic
           os: osx
-          osx_image: xcode8.3
+          osx_image: xcode8.3  # This is the latest working xcode image with osxfuse compatibility; later images come with an OS X version which doesn't allow kernel extensions
           env: TOXENV=py35
```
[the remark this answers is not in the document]
finefoot: Yes, unfortunately it seems like there isn't much response from Travis regarding the issue. Alternatively, maybe other free CI/CD services offer enabling KEXT in SIP? I don't know how important fuse tests are, but the xcode8.3 has been deprecated for a while now...

.travis/install.sh
ThomasWaldmann: is the amount of tests the same with and without zstd?
maybe we don't use zstd in the tests then.
assuming that zstd has own tests that make sure it functions correctly, the only thing we need to test is our interfacing to zstd.

IIRC, readline is a requirement for python. if it is not there, pyenv will download and install it (for every python it compiles?). That can maybe be avoided by offering it before pyenv is installed/used.

.travis/install.sh
ThomasWaldmann: we can use them if they start not being too old in the dist, maybe with bionic.

.travis.yml
ThomasWaldmann: Yes, because I don't want to put too much load onto the macOS infrastructure - it is slow as hell already anyway. Also I don't want to wait ages until it shows results.

.travis/install.sh
ThomasWaldmann: Doesn't that just terminate the install step, but would still do the test step?

.travis/install.sh
finefoot: Well, `brew install zstd` seems to be the main culprit regarding the timeout since it depends on CMake which takes approximately 15 minutes to build:

> 🍺  /usr/local/Cellar/cmake/3.16.5: 6,015 files, 56.8MB, built in 14 minutes 53 seconds

There is no way to directly tell if exactly the same tests run because, well, the timeout happens and not all tests run. However I can say that *without* `brew install zstd` until the timeout happens at about 43% all tests are the same. When omitting `brew install zstd`, there is a test called `test_zstd` which passes:

> [gw0] [ 52%] PASSED .tox/py35/lib/python3.5/site-packages/borg/testsuite/compress.py::test_zstd

Also, I when comparing a reduced testsuite without `archiver.py` and `benchmark.py` (to speed up the testing) runs exactly the same tests with the same results independent from `brew install zstd`. So I think it doesn't make a difference to the testing if we install it and can save that time.

Similar with `brew install readline`: I removed it from `.travis/install.sh` because it wasn't necessary. If pyenv needs it and can't find it, it will just install it then. Some xcode images seem to already provide a compatible version. So I thought it would be better to keep a clear `install.sh` but I can try if it makes a time difference.

.travis/install.sh
finefoot: But what advantage do those packages even provide? Are currently tests being skipped? :thinking:

.travis.yml
finefoot: Fair enough. Will change it.

.travis/install.sh
finefoot: You mean that it doesn't immediately quit but just keeps running? No, I don't think so, as far as I know that's only true for `script`. But I'll test it to make sure. :)

finefoot: Oh, I see what you mean, now. `exit` from inside the script only terminates the script itself and Travis just goes on to continue with `run.sh`. That's what you meant, right? I think there are three ways to go about this:

- Move the test for docs-only change back into `.travis.yml`. But I don't think that's a good place to have code - that's why I moved it out of there in the first place.
- Elaborate way to relay the "termination" signal from the exit in `install.sh` back to `.travis.yml`. I think that's way too much effort for such a small issue. I'd say having the code inside `.travis.yml` would be preferable to this.
- Merge `install.sh` and `run.sh` into one `script.sh` which then fully terminates with the one `exit` call. `run.sh` effectively only contains one line of code anyway, so there is not much separation lost other than the fold in the Travis log.

What do you think? I'll see if I can come up with other ideas, too.

ThomasWaldmann: guess moving it back to .travis.yml is the simplest and least-change solution.
otherwise we you also have to rename stuff, which is not nice for git history.

.travis/install.sh
ThomasWaldmann: ehrm, it must not run benchmarks on travis-ci, these are only intended for local usage under controlled conditions.

i didn't know zstd is a that big effort to install, we should live without it on travis macOS because of that.

omitting the archiver tests is not an option, they are the most important high-level tests.

the questions about readline is how often it gets downloaded and installed. if pyenv installs multiple times and we only do it once (and avoid the multiple installs), we could save some time.

.travis/install.sh
ThomasWaldmann: don't remember, could be. installing them on linux is almost zero effort, unlike to brew installation.

.travis/install.sh
ThomasWaldmann: and please put a comment in the src about zstd brew effort and that we don't want it.

.travis/install.sh
finefoot: :+1:

.travis/install.sh
finefoot: > and please put a comment in the src about zstd brew effort and that we don't want it.

:+1: 

> omitting the archiver tests is not an option, they are the most important high-level tests.

That's a misunderstanding ;) I didn't mean to suggest permanently remove the archiver tests - I merely tried to temporarily remove roughly the first half of the tests to see how the other half is doing (since I wouldn't see that second half due to the timeout).

> the questions about readline is how often it gets downloaded and installed

But there is only 1 installation per job? 1x pyenv which then 1x installs one specific version of Python.

.travis/install.sh
finefoot: :+1:

.travis/install.sh
ThomasWaldmann: ah, ok (archiver tests).

readline installations: hmm, maybe i confused it with my vagrant based tests where i build many pythons at once.

ThomasWaldmann:
the xcode 11 based tests stumble over xz already being installed.

i restarted the xcode 8 based tests, these fails looked like maybe a temporary problem.

.travis/install.sh
finefoot: :+1:

finefoot:
> the xcode 11 based tests stumble over xz already being installed.

Yes, apparently `brew install` quits with non-zero exit code if the package is already installed and a newer version is available. This issue only appeared now because if no newer version is available, `brew install` only throws a warning.

.travis/install.sh
```diff
@@ -3,54 +3,73 @@
...
+    export HOMEBREW_NO_AUTO_UPDATE=1  # Auto-updating everything would take too much time
+    brew update
+    brew cleanup  # Preempt possible scheduled clean-up so it doesn't clutter the log later
+
+    # Install and/or upgrade dependencies
+    #brew install zstd || brew upgrade zstd  # Installation takes very long due to CMake dependency and isn't necessary for the tests as borg comes bundled with zstd anyway
+    brew install lz4 || brew upgrade lz4
+    brew install xz || brew upgrade xz  # Required for Python lzma module
+    brew install Caskroom/cask/osxfuse || brew upgrade Caskroom/cask/osxfuse  # Required for Python llfuse module
+    brew install pyenv || brew upgrade pyenv
```
finefoot: @ThomasWaldmann Or would you rather use `brew bundle` at this point or `for FORMULA in lz4 xz Caskroom/cask/osxfuse pyenv`?

ThomasWaldmann: i've no idea about brew, not being a macOS user except borg testing.

finefoot: I think there are three possibilities:

- `brew install lz4 || brew upgrade lz4` for each formula (see current code, introduced with 25d3ff783b9d989dcc495ba7212d20d0c803056e)
- `for FORMULA in lz4 xz Caskroom/cask/osxfuse pyenv` with `brew install $FORMULA || brew upgrade $FORMULA`
- `brew bundle` which is like `pip -r` where requirements.txt is called Brewfile to specify formulae but can also be read from stdin

finefoot: But that's just a question of code style. The current code now works again for xcode11.3 https://travis-ci.org/github/borgbackup/borg/builds/666803536

ThomasWaldmann: for now keep the current style, more tuning can be done in a future PR. and I guess I prefer having it in-file, not in an external file in this case.

finefoot: I agree. :+1:

finefoot:
@ThomasWaldmann I don't know why some of the checks don't appear here on GitHub. It only shows the AppVeyor build. Even though Travis CI doesn't show up, the jobs did run and pass: https://travis-ci.org/github/borgbackup/borg/builds/667623581 Also, it doesn't seem to be an issue with the changes from this pull request as it happens in other pull requests, too.

.travis.yml
```diff
@@ -31,22 +31,30 @@ matrix:
...
         - language: generic
           os: osx
           osx_image: xcode8.3
           env: TOXENV=py37
+        - language: generic
+          os: osx
+          osx_image: xcode11.3
+          env: TOXENV=py35 SKIPFUSE=true
```
ThomasWaldmann, line 49: can we remove this, so we only have 2 macOS jobs in total?

.travis/install.sh
```diff
@@ -3,54 +3,73 @@
...
+    # Update brew itself
+    export HOMEBREW_NO_AUTO_UPDATE=1  # Auto-updating everything would take too much time
+    brew update
+    brew cleanup  # Preempt possible scheduled clean-up so it doesn't clutter the log later
+
+    # Install and/or upgrade dependencies
+    #brew install zstd || brew upgrade zstd  # Installation takes very long due to CMake dependency and isn't necessary for the tests as borg comes bundled with zstd anyway
+    brew install lz4 || brew upgrade lz4
+    brew install xz || brew upgrade xz  # Required for Python lzma module
+    brew install Caskroom/cask/osxfuse || brew upgrade Caskroom/cask/osxfuse  # Required for Python llfuse module
```
ThomasWaldmann, line 21: check SKIPFUSE here?

.travis/install.sh
```diff
@@ -3,54 +3,73 @@
...
+    #sudo apt-get install -y libzstd-dev  # Too old on trusty and xenial, but might be useful in future versions
     sudo apt-get install -y libacl1-dev
-    sudo apt-get install -y libfuse-dev fuse  # optional, for FUSE support
+    sudo apt-get install -y libfuse-dev fuse  # Required for Python llfuse module
+
+else
+
+    printf '%s\n' "Unexpected value for TRAVIS_OS_NAME environment variable"
+    exit 1
+
```
ThomasWaldmann, line 58: remove the blank lines please (consistency, see above).

.travis.yml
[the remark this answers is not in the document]
ThomasWaldmann: i don't know any better. also, it might be quite a huge effort to move anyway...

.travis.yml
finefoot: I'm sorry, I don't understand. Currently, there are 4 jobs with OS X, testing xcode8.3/xcode11.3 and py35/py37, respectively. Which ones do you want to remove to arrive at 2 OS X jobs in total?

.travis/install.sh
finefoot: My interpretation when I introduced the SKIPFUSE flag was to indicate if borg testsuite skips fuse tests, not to skip everything fuse-related. Additionally, installing osxfuse doesn't take much time at all. So I just left it there to avoid unnecessarily cluttering the lines. Hopefully, in the future, we'll be able to use osxfuse in xcode11.3 anyway. Do you still want to add a check for SKIPFUSE?

.travis/install.sh
finefoot: I'm confused... but the blank lines are consistent with the rest? There are blank lines around each platform-specific block: around linux and around osx - so I added blank lines around the else block for the unrecognized platform, too. Maybe that wasn't visible in the review and looking at the whole file clears up what I mean. Or do you want me to remove them all?

finefoot:
Thanks for restarting the Travis CI jobs. Now the checks show up. :+1:

> guess it is ready to merge then?

Yes, I'm not working on anything regarding this pull request.

.travis/install.sh
ThomasWaldmann: i mean 55 and 58 (see the other else branch).

.travis/install.sh
ThomasWaldmann: ok, if it doesn't take lots of time, we can keep it.

.travis.yml
ThomasWaldmann: - old osx / xcode, old python
- new osx / xcode, new python

.travis/install.sh
finefoot: Yes, I was talking about lines 55 and 58, too. The respective blank lines on the same level, for the `if` statements are lines 11, 42, 45, 53. The `if`, `else` from lines 29 to 40 doesn't have any blank lines because it's inside the indented block. (Another level.) Should I add blank lines there, too?

ThomasWaldmann: oh, sorry, i confused it with the other indentation level.

so, keep as is.

.travis.yml
finefoot: :+1:

finefoot:
Sorry, maybe I shouldn't have amended the change for easier reviewing. Last push only removed the two jobs xcode8.3/py37 and xcode11.3/py35 so there are now two jobs in total.

ThomasWaldmann:
OK, guess we want to backport this to 1.1-maint (because it makes the tests work more reliably).

Guess we could watch the PRs on master for a bit to see whether there is anything special and then backport this.

finefoot:
Agreed. :+1: I think I'll still just go ahead and add a pull request for the backport to see if anything fails.

PR #5003, opened 2020-03-08
remove redundant cythonize call

ThomasWaldmann:
the call in the following line did basically the same thing, just better as it was built according to the platform.

also, it includes syncfilerange_ext, which was missing from the removed line.


ThomasWaldmann:
meh, wrong: it is not redundant, closing.

PR #5023, opened 2020-03-11
Suppress Sphinx lexer warnings in frontends.rst

milkey-mouse:
Because some of the JSON blocks in frontends.rst included non-JSON contents (user replies, etc.) the blocks didn't parse/highlight and `make html` complained:

    borg/docs/internals/frontends.rst:28: WARNING: Could not lex literal_block as "json".

There's no easy way to enable and disable highlighting just for specific lines, but individually unsetting the highlight language per block suppresses the warnings.

### DIFF
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -1,5 +1,5 @@
 .. include:: ../global.rst.inc
-.. highlight:: json
+.. highlight:: none
 
 .. _json_output:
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
thanks! 1.1-maint also affected?

milkey-mouse:
I think so, I'll backport if so.

PR #5025, opened 2020-03-11
Reflow some really long lines in the docs

milkey-mouse:
I used `grep -Pnr '.{100}' *.rst` to find lines >100 characters long and reflowed them where reasonable (not in code snippets, otherwise well-formatted lists, etc.). Line length in the docs isn't too important (it doesn't matter once they're compiled), but it's nice not to have super-long lines in one's editor.


ThomasWaldmann:
thanks!

Issue #3867, opened 2018-06-01
efficiently transfer an archive from one repository to another


quickslyver:
- I have checked borgbackup docs, FAQ, and open Github issues
- It is a QUESTION

## Use case (context)

1. I have a local backup repository where there are multiple archives (dailies, weeklies and monthlies archives).
2. I would like to have a much smaller offsite backup repository where only the most recent daily archive is copied

## Problem

Currently it seems there is no command available in borg to copy archives between repository. (true ?)

## Workaround

I've found the following workarounds that all have drawbacks:
1. mount the archive i want to copy on the local backup repository (fuse module) and run a borg create on the offsite backup repository. But the fuse module seems not to be designed for that kind of "heavy duty access" (slow)
2. "borg extract" the archive on the local backup server filesystem  , "borg create" the archive on the offsite server. But it wastes a lot of storage on the local backup server.
3. backup directly from the source servers (the backed-up servers) to the offsite server: then the backed-up need to have access directly to the offsite server and they need to be backed-up twice (for the local and the offsite server)

The perfect solution would be a direct copy of the archive from one repository to another using a borg backup command

## Questions
Two questions:
1. Is there a better workaround for my use case ?
2. Is adding a command to copy an archive between two different borg repository feasible ? (I can help)

quickslyver:
> Archive transfer from one repo to other is neither trivial nor especially fast due to security reasons.

yes, I understand that the algorithm would be something near that of "borg recreate with re-chunking", rechunking is needed due to different encryption key.

wolfpackmars2:
Couldn't you just use two backup repositories? The local copy which has multiple backups and the remote backup, which presumably you could auto-prune to the last backup at the end of a backup operation. Yes you would be running borg backup twice, but it's probably ultimately less overhead than trying to keep two separate backups in sync.

quickslyver:
Yes that's the workaround number 3 !

But with my real use case it's a little cumbersome.

Currently:

* I backup (using rsync) every day tens of servers to a local backup server (currently on ZFS)
* I make snapshots every day a keep a fixed number of daily/weekly/monthly (classic...)
* every week I **physically** externalize the most recent backup to a 8TB HDD using borg backup (and I can't do this using a network connection)
* I'm trying to switch to borg backup instead of ZFS snapshot for the local backup (which triggers the problem of transferring the  most recent borg backup of one repo to another)

[...]

PR #5032, opened 2020-03-15
Fix FUSE low linear read speed on large files

### LINKED ISSUE #5067: FUSE mount is slow for linear reads of big files
As discovered and fixed by @lexa-a in #5032, linearly reading a big file from a borg fuse mount is slower than it needs to be.

From the PR comment:

This patch is a quick fix for an inefficient seek algorithm used in read() function:
- On each read call chunk list for given file is regenerated.
- Finding needed chunk takes a lot of time for large offsets.

For 64 GB files read speed drops to 16 MB/s, for 128 GB - to 8 MB/s, and so on.

After applying this patch read speed becomes 270 MB/s, and does not depend on read position inside file.

lexa-a:
Fixes #5067.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -38,7 +38,10 @@ else:
         return None
 
 # size of some LRUCaches (1 element per simultaneously open file)
-FILES = 100
+# note: _inode_cache might have rather large elements - Item.chunks can be large!
+#       also, simultaneously reading too many files should be avoided anyway.
+#       thus, do not set FILES to high values.
+FILES = 4
 
 
 class ItemCache:
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -37,6 +37,9 @@ else:
         llfuse.main(single=True)
         return None
 
+# size of some LRUCaches (1 element per simultaneously open file)
+FILES = 100
+
 
 class ItemCache:
     """
@@ -223,7 +226,8 @@ class FuseBackend(object):
         # made up and are not contained in the archives. For example archive directories or intermediate directories
         # not contained in archives.
         self._items = {}
-        self._read_cache = LRUCache(capacity=1, dispose=lambda _: None)
+        # cache up to <FILES> Items
+        self._inode_cache = LRUCache(capacity=FILES, dispose=lambda _: None)
         # _inode_count is the current count of synthetic inodes, i.e. those in self._items
         self.inode_count = 0
         # Maps inode numbers to the inode number of the parent
@@ -260,16 +264,16 @@ class FuseBackend(object):
                     self.pending_archives[archive_inode] = archive.name
 
     def get_item(self, inode):
-        item = self._read_cache.get(inode)
-        if item:
+        item = self._inode_cache.get(inode)
+        if item is not None:
             return item
         try:
-            item = self._items[inode]
-            self._read_cache[inode] = item
-            return item
+            # this is a cheap get-from-dictionary operation, no need to cache the result.
+            return self._items[inode]
         except KeyError:
+            # while self.cache does some internal caching, it has still quite some overhead, so we cache the result.
             item = self.cache.get(inode)
-            self._read_cache[inode] = item
+            self._inode_cache[inode] = item
             return item
 
     def check_pending_archive(self, inode):
@@ -438,7 +442,7 @@ class FuseOperations(llfuse.Operations, FuseBackend):
         data_cache_capacity = int(os.environ.get('BORG_MOUNT_DATA_CACHE_ENTRIES', os.cpu_count() or 1))
         logger.debug('mount data cache capacity: %d chunks', data_cache_capacity)
         self.data_cache = LRUCache(capacity=data_cache_capacity, dispose=lambda _: None)
-        self._last_pos = LRUCache(capacity=1, dispose=lambda _: None)
+        self._last_pos = LRUCache(capacity=FILES, dispose=lambda _: None)
 
     def sig_info_handler(self, sig_no, stack):
         logger.debug('fuse: %d synth inodes, %d edges (%s)',
@@ -615,21 +619,20 @@ class FuseOperations(llfuse.Operations, FuseBackend):
     def read(self, fh, offset, size):
         parts = []
         item = self.get_item(fh)
-        last_pos = self._last_pos.get(fh)
 
-        if not last_pos:
-            last_pos = (0, 0)
-        if last_pos[1] > offset:
-            last_pos = (0, 0)
+        # optimize for linear reads:
+        # we cache the chunk number and the in-file offset of the chunk in _last_pos[fh]
+        chunk_no, chunk_offset = self._last_pos.get(fh, (0, 0))
+        if chunk_offset > offset:
+            # this is not a linear read, so we lost track and need to start from beginning again...
+            chunk_no, chunk_offset = (0, 0)
 
-        pos, poffset = last_pos
-        offset -= poffset
-        npos = pos
-        for id, s, csize in item.chunks[pos:]:
+        offset -= chunk_offset
+        for id, s, csize in item.chunks[chunk_no:]:
             if s < offset:
                 offset -= s
-                poffset += s
-                npos += 1
+                chunk_offset += s
+                chunk_no += 1
                 continue
             n = min(size, s - offset)
             if id in self.data_cache:
@@ -647,9 +650,9 @@ class FuseOperations(llfuse.Operations, FuseBackend):
             size -= n
             if not size:
                 if fh in self._last_pos:
-                    self._last_pos.upd(fh, (npos, poffset))
+                    self._last_pos.upd(fh, (chunk_no, chunk_offset))
                 else:
-                    self._last_pos[fh] = (npos, poffset)
+                    self._last_pos[fh] = (chunk_no, chunk_offset)
                 break
         return b''.join(parts)
 
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -628,7 +628,10 @@ class FuseOperations(llfuse.Operations, FuseBackend):
             chunk_no, chunk_offset = (0, 0)
 
         offset -= chunk_offset
-        for id, s, csize in item.chunks[chunk_no:]:
+        chunks = item.chunks
+        # note: using index iteration to avoid frequently copying big (sub)lists by slicing
+        for idx in range(chunk_no, len(chunks)):
+            id, s, csize = chunks[idx]
             if s < offset:
                 offset -= s
                 chunk_offset += s
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -262,14 +262,14 @@ class FuseBackend(object):
     def get_item(self, inode):
         item = self._read_cache.get(inode)
         if item:
-            return item        
+            return item
         try:
             item = self._items[inode]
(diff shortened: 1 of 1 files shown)


lexa-a:
And here is backport for 1.1 branch, if needed.
[fuse-for-1.1-branch.zip](https://github.com/borgbackup/borg/files/4334294/fuse-for-1.1-branch.zip)

src/borg/fuse.py
```diff
@@ -606,9 +615,21 @@ def opendir(self, inode, ctx=None):
     def read(self, fh, offset, size):
         parts = []
         item = self.get_item(fh)
-        for id, s, csize in item.chunks:
+        last_pos = self._last_pos.get(fh)
+
+        if not last_pos:
```
ThomasWaldmann: general remark: if you use .get() and want afterwards to test for the "not found" case, always use `is None`.

Just using `if` or `if not` can easily introduce bugs if there can be falsy values in the cache dict.

This of course assumes that `None` is not in the valid values stored in the cache dict, otherwise one needs to use a different singleton as default and in the `is defaultobject` test.

src/borg/fuse.py
```diff
@@ -259,10 +260,17 @@ def _create_filesystem(self):
...
         try:
-            return self._items[inode]
+            item = self._items[inode]
+            self._read_cache[inode] = item
+            return item
         except KeyError:
-            return self.cache.get(inode)
+            item = self.cache.get(inode)
+            self._read_cache[inode] = item 
+            return item
```
ThomasWaldmann: I didn't read the full code right now, but that looks almost like putting a cache onto a cache (or something that could be a cache).

src/borg/fuse.py
```diff
@@ -606,9 +615,21 @@ def opendir(self, inode, ctx=None):
...
+        last_pos = self._last_pos.get(fh)
+
+        if not last_pos:
+            last_pos = (0, 0)
+        if last_pos[1] > offset:
+            last_pos = (0, 0)
+                    
+        pos, poffset = last_pos
+        offset -= poffset
+        npos = pos
```
ThomasWaldmann: can you use more self-explaining names?

src/borg/fuse.py
```diff
@@ -223,6 +223,7 @@ def __init__(self, key, manifest, repository, args, decrypted_repository):
         # made up and are not contained in the archives. For example archive directories or intermediate directories
         # not contained in archives.
         self._items = {}
+        self._read_cache = LRUCache(capacity=1, dispose=lambda _: None)
```
ThomasWaldmann: guess we'll need capacity > 1 if parallel reading of more than a file should perform better?

src/borg/fuse.py
```diff
@@ -430,6 +438,7 @@ def __init__(self, key, repository, manifest, args, decrypted_repository):
         data_cache_capacity = int(os.environ.get('BORG_MOUNT_DATA_CACHE_ENTRIES', os.cpu_count() or 1))
         logger.debug('mount data cache capacity: %d chunks', data_cache_capacity)
         self.data_cache = LRUCache(capacity=data_cache_capacity, dispose=lambda _: None)
+        self._last_pos = LRUCache(capacity=1, dispose=lambda _: None)
```
ThomasWaldmann: guess we'll need capacity > 1 if parallel reading of more than a file should perform better?

ThomasWaldmann:
About the backport: not sure what is in the .zip, but let's first finish this PR and then decide whether to backport.

src/borg/fuse.py
lexa-a: Check comment for  self._items = {} :
        # Maps inode numbers to Item instances. This is used for synthetic inodes, i.e. file-system objects that are made up and are not contained in the archives. For example archive directories or intermediate directories not contained in archives.

File objects are never placed to self._items cache. So _read_cache is just a workaround for this problem, not permanent solution.

src/borg/fuse.py
lexa-a: Yes

src/borg/fuse.py
lexa-a: Yes, equal to _read_cache capacity

lexa-a:
> thanks for this PR. do we have an issue for this?
> if not, please file an issue first with all the details and then refer to this issue with "fixes #XXXX" from this PR.

We tried to implement repository synchronization without data extraction, but using fuse mount.
Problem with fuse mount read speed is described in this ticket:
https://github.com/borgbackup/borg/issues/3867
"But the fuse module seems not to be designed for that kind of "heavy duty access" (slow)"
"I also tried the mount method described by the op (workaround 1) but it was incredibly slow to do it this way, but that's what I'll probably continue to use for now."

src/borg/fuse.py
lexa-a: Sorry, my knowledge of python is very limited, I asked my friend to help me write this fix. I suppose it would be easy for project maintainers to just rewrite whole fix to be consistent with the project.

lexa-a:
fuse

Attaching python profiler data for unpached borg while reading files via FUSE.
See ncalls for read, get_item, get, ...., Unpascker.__next__  are all ~62000. This shows that caching does now work.

ThomasWaldmann:
Some issues found by travis-ci / pep8 checker:
```
src/borg/fuse.py:265:24: W291 trailing whitespace

src/borg/fuse.py:272:43: W291 trailing whitespace

src/borg/fuse.py:624:1: W293 blank line contains whitespace

src/borg/fuse.py:628:44: E203 whitespace before ':'

src/borg/fuse.py:628:46: E202 whitespace before ']'

src/borg/fuse.py:653:1: W293 blank line contains whitespace
```

ThomasWaldmann:
ping.

ThomasWaldmann:
I'll work a bit on this one ...

lexa-a:
Today i have found another problem with speed after applying this patch.
It works good on default chunker params, but using smaller chunks i get lower speeds (testing using dd of=/dev/null bs=1M)
1) (16,17,16,4095) -> 23 MB/s on 46Gb file, 170 MB/s on 3 Gb file
2) (17,18,17,4095) -> 46 MB/s on 46Gb file, 187 MB/s on 3 Gb file
3) (17,21,19,4095) -> 140 MB/s on 46Gb file, 180 MB/s on 3 Gb file

I did not profile this problem so far.

src/borg/fuse.py
```diff
@@ -259,10 +260,17 @@ def _create_filesystem(self):
                     self.pending_archives[archive_inode] = archive.name
 
     def get_item(self, inode):
+        item = self._read_cache.get(inode)
+        if item:
+            return item        
         try:
-            return self._items[inode]
+            item = self._items[inode]
+            self._read_cache[inode] = item
```
ThomasWaldmann, line 268: this is not needed, it is "caching" just a pure-dictionary get operation.

this is rather counterproductive as a dict-get is fast and caching it takes away a LRU cache entry that could be used better.

src/borg/fuse.py
```diff
@@ -259,10 +260,17 @@ def _create_filesystem(self):
...
+            return item
         except KeyError:
-            return self.cache.get(inode)
+            item = self.cache.get(inode)
+            self._read_cache[inode] = item
```
ThomasWaldmann, line 272: otoh, this makes sense as there is some overhead in self.cache.get() - although it internally also uses a cache for chunks.

ThomasWaldmann:
OK, I resolved the stuff I gave in my first review.

@lexa-a please review my 2 fixup changesets.

ThomasWaldmann:
did another micro opt to my code.

@lexa-a can you benchmark the current code (all changesets) against your initial code (first changeset)?

lexa-a:
> @lexa-a can you benchmark the current code (all changesets) against your initial code (first changeset)?

Thanks for fixing, I will try to benchmark tomorrow.

lexa-a:
I have tested new fix, it gives +3% speed to orignal fix (330 mb/s new, 320 mb/s - old, tested using dd of=/dev/null bs=1M) on 46 Gb file with default chunker parameters. 

Testing on chunker (16,17,16,4095) gives same low speed - 23 MB/s  on 46 Gb files. and 300 mb/s on 3 Gb file.

I am attaching two borg profiles for reading 46 Gb and 3Gb file from (16,17,16,4095) archive . It is visible that read() function itself becomes slow on large number of chunks  (compare tottime of read() to openssl_sha256()) .

46Gb file read:
46Gb-profile

3Gb file read:
3Gb-profile

Tokyx:
Hi,

Come from #5063, I see this PR solving a reading problem with a lot amount of chunks when reading from fuse.
On our side, we have the same problem, but we don't use FUSE. We use SSH. Is it possible that this problem is also present in the reading function through SSH? Is it fixable ?

ThomasWaldmann:
@lexa-a did you really use this PR's code to make the new profiles? they show fuse.read at line 606, but it is at line 619.

@Tokyx the borg mount code is rather different from borg extract. A common issue is of course that the more chunks you have, the more overhead you get.

Tokyx:
@ThomasWaldmann thank you for the answer.

What seems surprising to me is that the performance problem only present when the extraction is done by ssh.

borg extract on local repo: 50-60MB / s
borg extract on local repo via SSH (ssh 127.0.0.1): 4-5MB / s.
In borg create, regardless of the use of SSH or not, we have performances at 50-60MB / s.

Is it a common issue ?

With the fix of @lexa-a, it would be interesting to test the perfs with borg mount in SSHFS and borg extract.

lexa-a:
> @lexa-a did you really use this PR's code to make the new profiles? they show fuse.read at line 606, but it is at line 619.

Here is correct profile for 46Gb file with small chunks - same situation.
(previous profiles was captured with my initial fix for borg 1.1 by mistake, but speed measurements were correct)

46gb-2

ThomasWaldmann:
I had a look at `read()`, but didn't find anything obvious to me why it is slow for smaller chunks.

But the current PR has a clear scope and definitely fixes one issue, so just merge at is and search for more later?

lexa-a:
> But the current PR has a clear scope and definitely fixes one issue, so just merge at is and search for more later?
It's up to you, maybe it's better to create a new issue. (But later also may become never)
Function read should be profiled inside to find the cause, but I do not know how to do that, I do not know python well.

ThomasWaldmann:
@lexa-a good idea. i somehow had thought that the construct was efficient, but some research seems to indicate it is not (but doing lots of temporary sub-list creation).

I tried to optimize it in my last fixup commit, can you check whether this performs better for the many small chunks case?

lexa-a:
Thanks for fix, now it works well:
Deduplicated size=31.62 GB, Unique chunks=367413
(48 GB) copied, 131.732 s, 367 MB/s

lexa-a:
>FILES = 100
Last concern is how much memory will take cache for 100 files, 300000 chunks each?
There seems to be no explicit eviction of elements, so cache will remain full after reading 100 files?
Maybe it's better to set lower default value?

ThomasWaldmann:
Yeah, guess FILES should be lower. I was rather thinking about small / medium files, when the chunks list has negligible size. But if one has a bunch of large files, like VM images AND small chunk size, that's not the case ...

ThomasWaldmann:
I'll squash and merge after tests have finished.

ThomasWaldmann:
Thanks for the PR, big improvement!

I'll backport it now to 1.1-maint...

Issue #4073, opened 2018-09-06
PlaceholderError not printed as json


sophie-h:
borg 1.1.7

    $ borg create --progress --log-json dest::{x} src
    Invalid placeholder "x" in string: dest::{x}

    $ borg create --progress --log-json dest::x src
    {"type": "log_message", "name": "borg.archiver", "msgid": "Repository.DoesNotExist", …

ThomasWaldmann:
The first one is a arg parsing issue.

arg parsing happens before logging (including the json logger) is set up, so in case of exceptions, it does a rather simple `print` to output stuff.

as the logging setup depends on arguments being parsed already (e.g. for logging options like `--debug`, `-v`, whether or not to setup the json logger), I guess we have a chicken and egg problem here.

sophie-h:
I think this could be solved in principle (ENV + catching argparse exception), but that pretty hackish and the non-json error is just a minor inconvenience imo.

Maybe, it would still be worth to document it in the JSON section?

ThomasWaldmann:
Yeah, could be mentioned - can you do a PR?

fantasya-pbem:
I have added a short warning, see my PR.

fantasya-pbem:
This issue can be close now.

ThomasWaldmann:
Thanks for fixing!

PR #5033, opened 2020-03-15
[DOCS] #4073 – PlaceholderError not printed as JSON

fantasya-pbem:
Add a warning for --log-json option that parsing errors will always be printed as plain text.


ThomasWaldmann:
thanks!

Issue #3428, opened 2017-12-11
Borg repository restore/reinstall instructions are needed


jk-1:
In case of HW failure or media corruption you need to restore your borg repos from a backup media. Because the restored repo is almost always older than damaged repo you need to to something with the client cache to get the restored repo working.

I tried following scenarions to restore repo after an imaginary HW failure:

This is the original working repo soon to be damaged:
```
~/tmp/RepoTest$ borg list repo01 --format="{archive} {TAB} {time} {TAB}{comment} {NEWLINE}"
b01 	 Mon, 2017-12-11 10:32:37 	"Initial-backup" 
b02 	 Mon, 2017-12-11 10:32:38 	"Second-backup" 
b-mod-bckp-01 	 Mon, 2017-12-11 10:32:42 	"This-is-mod-bckp-01" 
b-mod-bckp-02 	 Mon, 2017-12-11 10:32:43 	"This-is-mod-bckp-02" 
b-mod-bckp-03 	 Mon, 2017-12-11 10:32:44 	"This-is-mod-bckp-03" 
```
Repo is periodically backed up to another "secure" location, e.g. removable disk, tape etc.
```
~/tmp/RepoTest$ cp -avr repo01 repo01.1
```
Repo is in use, backups are stored into it.
```
borg create repo01::B03 dir01 --comment "B03"

~/tmp/RepoTest$ cp -avr repo01 repo01.2

borg create repo01::B04 dir02 --comment "B04"

~/tmp/RepoTest$ cp -avr repo01 repo01.3

borg create repo01::B05 dir02 --comment "B05"
```
Situation before crash:
```
~/tmp/RepoTest$ borg list repo01 --format="{archive} {TAB} {time} {TAB}{comment} {NEWLINE}"
b01 	 Mon, 2017-12-11 10:32:37 	"Initial-backup" 
b02 	 Mon, 2017-12-11 10:32:38 	"Second-backup" 
b-mod-bckp-01 	 Mon, 2017-12-11 10:32:42 	"This-is-mod-bckp-01" 
b-mod-bckp-02 	 Mon, 2017-12-11 10:32:43 	"This-is-mod-bckp-02" 
b-mod-bckp-03 	 Mon, 2017-12-11 10:32:44 	"This-is-mod-bckp-03" 
B03 	 Mon, 2017-12-11 10:37:38 	B03 
B04 	 Mon, 2017-12-11 10:38:36 	B04 
[...]

PR #5034, opened 2020-03-15
[DOCS] #3428 –  Borg repo restore instructions needed

fantasya-pbem:
Add new FAQ: A repo is corrupt and must be replaced with an older repo.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -202,9 +202,10 @@ the working repository to the same location:
 A plain delete command would remove the security info in
 ``~/.config/borg/security``, including the nonce value. In BorgBackup
 :ref:`security_encryption` is AES-CTR, where the nonce is a counter. When the
-working repo was used later for creating new archives, Borg would initialize a
-fresh nonce, which would be bad for security reasons. To prevent this, the
-``keep-security-info`` option is applied so that the nonce counter is kept.
+working repo was used later for creating new archives, Borg would re-use nonce
+values due to starting from a lower counter value given by the older copy of the
+repository. To prevent this, the ``keep-security-info`` option is applied so
+that the client-side nonce counter is kept.
 
 Can Borg add redundancy to the backup data to deal with hardware malfunction?
 -----------------------------------------------------------------------------
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -187,6 +187,38 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
+My repository is corrupt, how can I restore from an older but working repository?
+---------------------------------------------------------------------------------
+
+If the working repo has the same ID as the corrupt one, the recommended method
+is to delete the corrupted repository, and then copy the working repository to
+the same location. The delete command will completely remove the corrupt repo
+and delete the corresponding cache and security subdirectory in
+``~/.config/borg/security``, including the nonce value (if encryption is used).
+When the working repo is used later for creating new archives, Borg would
+initialize a fresh nonce, which would be bad for security reasons (nonce values
+should never be reused). To prevent this, the security subdirectory should be
+saved before deleting, and later moved back into place.
+
+Example:
+
+::
+
+    # Get the repo ID from repo config.
+    REPO_ID=$(borg config /path/to/repo-good id)
+
+    # Rename the repo security dir so Borg won't delete it.
+    cd ~/.config/borg/security
+    mv $REPO_ID $REPO_ID.backup
+
+    # Now delete and rename the security dir back.
+    borg delete /path/to/repo
+    mv $REPO_ID.backup $REPO_ID
+
+    # Finally copy the good repo to the original place.
+    rsync -avH /path/to/repo-good /path/to/repo
+
+
 Can Borg add redundancy to the backup data to deal with hardware malfunction?
 -----------------------------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -187,37 +187,24 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
-My repository is corrupt, how can I restore from an older but working repository?
----------------------------------------------------------------------------------
+My repository is corrupt, how can I restore from an older copy of it?
+---------------------------------------------------------------------
 
-If the working repo has the same ID as the corrupt one, the recommended method
-is to delete the corrupted repository, and then copy the working repository to
-the same location. The delete command will completely remove the corrupt repo
-and delete the corresponding cache and security subdirectory in
-``~/.config/borg/security``, including the nonce value (if encryption is used).
-When the working repo is used later for creating new archives, Borg would
-initialize a fresh nonce, which would be bad for security reasons (nonce values
-should never be reused). To prevent this, the security subdirectory should be
-saved before deleting, and later moved back into place.
-
-Example:
+If your repositories are encrypted and have the same ID, the recommended method
+is to delete the corrupted repository, but keep its security info, and then copy
+the working repository to the same location:
 
 ::
 
-    # Get the repo ID from repo config.
-    REPO_ID=$(borg config /path/to/repo-good id)
+    borg delete --keep-security-info /path/to/repo
+    rsync -aH /path/to/repo-working/ /path/to/repo  # Note the trailing slash.
 
-    # Rename the repo security dir so Borg won't delete it.
-    cd ~/.config/borg/security
-    mv $REPO_ID $REPO_ID.backup
-
-    # Now delete and rename the security dir back.
-    borg delete /path/to/repo
-    mv $REPO_ID.backup $REPO_ID
-
-    # Finally copy the good repo to the original place.
-    rsync -avH /path/to/repo-good /path/to/repo
-
+A plain delete command would remove the security info in
+``~/.config/borg/security``, including the nonce value. In BorgBackup
+:ref:`security_encryption` is AES-CTR, where the nonce is a counter. When the
+working repo was used later for creating new archives, Borg would initialize a
+fresh nonce, which would be bad for security reasons. To prevent this, the
+``keep-security-info`` option is applied so that the nonce counter is kept.
 
 Can Borg add redundancy to the backup data to deal with hardware malfunction?
 -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


docs/faq.rst
```diff
@@ -187,6 +187,20 @@ all the part files and manually concatenate them together.
...
+My repository is corrupted, how can I restore from an older but working copy?
+-----------------------------------------------------------------------------
+
+The recommended method is to delete the corrupted repository, and then copy the
+working repo to the same location:
+
+::
+
+    borg delete /path/to/repo
+    cp -avr /path/to/repo-good /path/to/repo
```
ThomasWaldmann: There is a nonce-reuse security issue with that.

fantasya-pbem: How do you mean that? Isn't the local .config/borg/security/.../nonce deleted by ``borg delete``?

ThomasWaldmann: Guess so. And that is (IIRC) the problem. If there is no local information about the highest already used counter (nonce), it will rely on the counter determined from the repo.
And if you use an older copy of the repo, the counter will be lower and thus it will reuse counter values.

fantasya-pbem: So this should be considered an issue in ``borg delete`` and as a work-around until it is fixed, we could add a ``rm -r ~/.config/borg/security/<ID>`` (or just <ID>/nonce?)? If you agree, I'd open an issue for that and add this to the PR.

ThomasWaldmann: not sure you understood my previous comment correctly.

it would need rather the opposite of `rm` to work around, I guess.

fantasya-pbem: What is bad about reusing older nonces?

ThomasWaldmann: it breaks the security of AES-CTR encryption.

fantasya-pbem: OK, so we have to use the higher nonce of the deleted repo with the older repo. How to achieve that? Maybe rename the security/<ID> directory so that ``borg delete`` doesn't remove it? And rename it back finally?

ThomasWaldmann: yeah, could be. but first try that practically.

fantasya-pbem: OK, I will improve it and run a test.

docs/faq.rst
```diff
@@ -187,6 +187,38 @@ all the part files and manually concatenate them together.
...
+nonce which is bad for security reasons (nonce values should never be reused).
+To prevent this, the security subdirectory should be saved before deleting, and
+later moved back into place.
+
+Example:
+
+::
+
+    # Get the repo ID from repo config.
+    REPO_ID=$(grep 'id = ' /path/to/repo-good/config | cut -d ' ' -f 3)
```
ThomasWaldmann: don't we have some borg command showing the id? that would also work for remote ssh: repos.

docs/faq.rst
```diff
@@ -187,6 +187,38 @@ all the part files and manually concatenate them together.
...
+
+    # Rename the repo security dir so Borg won't delete it.
+    cd ~/.config/borg/security
+    mv $REPO_ID $REPO_ID.backup
+
+    # Now delete and rename the security dir back.
+    borg delete /path/to/repo
```
ThomasWaldmann: maybe borg delete should have some --keep-security-infos option.

docs/faq.rst
```diff
@@ -187,6 +187,38 @@ all the part files and manually concatenate them together.
...
+    mv $REPO_ID.backup $REPO_ID
```
ThomasWaldmann: this depends. if the intact repo is really a COPY of the repo (same id), this works.

if it is just another independent repo with the same contents, we need a different procedure.

docs/faq.rst
```diff
@@ -187,6 +187,38 @@ all the part files and manually concatenate them together.
...
+
+    # Finally copy the good repo to the original place.
+    cp -avr /path/to/repo-good /path/to/repo
```
ThomasWaldmann: `rsync -avH` is a better tool here. it's better than cp regarding copying as much metadata as possible and also can work remotely. also if it gets interrupted, it does not start from the beginning again.

docs/faq.rst
fantasya-pbem: Oh yes, it's ``borg config /path/to/repo id``. I will use that.

docs/faq.rst
fantasya-pbem: If the good repo G has another ID, the security dir of G won't get removed by ``borg delete /path/to/repo``, so we don't need to save it. This FAQ is for a real copy only, otherwise the whole procedure is unneccessary. I will add an additional note.

docs/faq.rst
fantasya-pbem: Good point, I'll use that.

docs/faq.rst
fantasya-pbem: Good idea! Let me see if I can add such an option.

Shall I make a separate PR for this?

fantasya-pbem:
Somehow the TravisCI check is missing here. But it exists:
https://travis-ci.org/github/borgbackup/borg/builds/667127892

src/borg/archiver.py
```diff
@@ -3383,10 +3388,11 @@ def define_borg_mount(parser):
                                help='print statistics for the deleted archive')
         subparser.add_argument('--cache-only', dest='cache_only', action='store_true',
                                help='delete only the local cache for the given repository')
-        subparser.add_argument('--force', dest='forced',
-                               action='count', default=0,
+        subparser.add_argument('--force', dest='forced', action='count', default=0,
                                help='force deletion of corrupted archives, '
                                     'use ``--force --force`` in case ``--force`` does not work.')
+        subparser.add_argument('--keep-security-info', dest='keep_security_info', action='store_true',
+                               help='keep the security info when deleting a repository')
```
ThomasWaldmann: ... **local** security info ... (see help text of --cache-only).

src/borg/archiver.py
```diff
@@ -1207,9 +1208,12 @@ def _delete_repository(self, args, repository):
             if not dry_run:
                 repository.destroy()
                 logger.info("Repository deleted.")
-                SecurityManager.destroy(repository)
+                if not keep_security_info:
+                    SecurityManager.destroy(repository)
             else:
                 logger.info("Would delete repository.")
+                if keep_security_info:
+                    logger.info("Would keep security info.")
```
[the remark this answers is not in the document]
fantasya-pbem: done

docs/faq.rst
```diff
@@ -187,6 +187,26 @@ all the part files and manually concatenate them together.
...
+
+::
+
+    borg delete --keep-security-info /path/to/repo
+    rsync -aH /path/to/repo-working/ /path/to/repo  # Note the trailing slash.
+
+A plain delete command would remove the security info in
+``~/.config/borg/security``, including the nonce value. When the working repo
+was used later for creating new archives, Borg would initialize a fresh nonce,
+which would be bad for security reasons (nonce values should never be reused).
```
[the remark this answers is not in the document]
fantasya-pbem: done

docs/faq.rst
```diff
@@ -187,6 +187,26 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
+My repository is corrupt, how can I restore from an older but working repository?
```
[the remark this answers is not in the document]
fantasya-pbem: done

src/borg/archiver.py
fantasya-pbem: done

src/borg/archiver.py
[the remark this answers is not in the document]
fantasya-pbem: done

fantasya-pbem:
Yes, I tested it for every combination of -n / --keep-security-info and checked that --cache-only works. Tested also the rsync command.

fantasya-pbem:
Why are the TravisCI checks not linked here anymore?

ThomasWaldmann:
travis-ci: ehrm, good question. didn't change anything.

fantasya-pbem:
Is this ready to merge? Is something missing yet?

src/borg/archiver.py
```diff
@@ -1207,9 +1208,14 @@ def _delete_repository(self, args, repository):
...
+                else:
+                    logger.info("Would delete security info.")
```
ThomasWaldmann, line 1218:
```
logger.info("Would %s security info." % ("keep" if keep_security_info else "delete"))
```

src/borg/archiver.py
```diff
@@ -3352,9 +3358,10 @@ def define_borg_mount(parser):
         Important: When deleting archives, repository disk space is **not** freed until
         you run ``borg compact``.
 
-        If you delete the complete repository, the local cache for it (if any) is
-        also deleted. Alternatively, you can delete just the local cache with the
-        ``--cache-only`` option.
+        When you delete a complete repository, the security info and local cache for it
+        (if any) is also deleted. Alternatively, you can delete just the local cache
```
ThomasWaldmann, line 3362: ... are also deleted.

docs/faq.rst
```diff
@@ -187,6 +187,25 @@ all the part files and manually concatenate them together.
...
+::
+
+    borg delete --keep-security-info /path/to/repo
+    rsync -aH /path/to/repo-working/ /path/to/repo  # Note the trailing slash.
+
+A plain delete command would remove the security info in
+``~/.config/borg/security``, including the nonce value. In BorgBackup
+:ref:`security_encryption` is AES-CTR, where the nonce is a counter. When the
+working repo was used later for creating new archives, Borg would initialize a
+fresh nonce, which would be bad for security reasons. To prevent this, the
```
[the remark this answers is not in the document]
fantasya-pbem: fixed

docs/faq.rst
```diff
@@ -187,6 +187,25 @@ all the part files and manually concatenate them together.
...
+``keep-security-info`` option is applied so that the nonce counter is kept.
```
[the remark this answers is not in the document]
fantasya-pbem: fixed

src/borg/archiver.py
fantasya-pbem: nice!

src/borg/archiver.py
fantasya-pbem: fixed

fantasya-pbem:
Merge?

ThomasWaldmann:
Thanks!

PR #5042, opened 2020-03-17
drop pointless asserts covered by other stricter asserts already

elho:
Based on the comments in PR #4683, remove asserts that will always be true if the assert following on he next line is true as well.


ThomasWaldmann:
The asserts were not completely pointless as it is still interesting if the removed ones pass if the still there ones fail.

OTOH, if we have failing tests, one needs to look at it anyway and can also add debug prints or whatever to find out more details.

PR #5043, opened 2020-03-17
Optimize compressor instantiation

elho:
Based on the comments in PR #4683, as a micro-optimization, do not instantiate LZ4 and none compressors on each instantiation of the auto compressor, but use global instances instead.


ThomasWaldmann:
Thanks!

PR #5046, opened 2020-03-18
[FEATURE] #4489 – Deprecate --nobsdflags option

### LINKED ISSUE #4489: Clarify --nobsdflags behavior
It is unclear what file flags and attributes will be skipped from the backup when `--nobsdflags` is used. Does this option only affect BSD like systems, or is it about a certain subset of flags that are present on all filesystems?

fantasya-pbem:
Replaced by --noflags. In internal data structure the key 'bsdflags' is kept for backwards compatibility.

fixes #4489

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -375,7 +375,7 @@ class Archive:
         """Failed to encode filename "{}" into file system encoding "{}". Consider configuring the LANG environment variable."""
 
     def __init__(self, repository, key, manifest, name, cache=None, create=False,
-                 checkpoint_interval=1800, numeric_owner=False, noatime=False, noctime=False, nobsdflags=False,
+                 checkpoint_interval=1800, numeric_owner=False, noatime=False, noctime=False, noflags=False,
                  progress=False, chunker_params=CHUNKER_PARAMS, start=None, start_monotonic=None, end=None,
                  consider_part_files=False, log_json=False):
         self.cwd = os.getcwd()
@@ -393,7 +393,7 @@ class Archive:
         self.numeric_owner = numeric_owner
         self.noatime = noatime
         self.noctime = noctime
-        self.nobsdflags = nobsdflags
+        self.noflags = noflags
         assert (start is None) == (start_monotonic is None), 'Logic error: if start is given, start_monotonic must be given as well and vice versa.'
         if start is None:
             start = datetime.utcnow()
@@ -846,7 +846,7 @@ Utilization of max. archive size: {csize_max:.0%}
             if warning:
                 set_ec(EXIT_WARNING)
             # bsdflags include the immutable flag and need to be set last:
-            if not self.nobsdflags and 'bsdflags' in item:
+            if not self.noflags and 'bsdflags' in item:
                 try:
                     set_flags(path, item.bsdflags, fd=fd)
                 except OSError:
@@ -1025,11 +1025,11 @@ Utilization of max. archive size: {csize_max:.0%}
 
 
 class MetadataCollector:
-    def __init__(self, *, noatime, noctime, numeric_owner, nobsdflags, nobirthtime):
+    def __init__(self, *, noatime, noctime, numeric_owner, noflags, nobirthtime):
         self.noatime = noatime
         self.noctime = noctime
         self.numeric_owner = numeric_owner
-        self.nobsdflags = nobsdflags
+        self.noflags = noflags
         self.nobirthtime = nobirthtime
 
     def stat_simple_attrs(self, st):
@@ -1058,16 +1058,16 @@ class MetadataCollector:
 
     def stat_ext_attrs(self, st, path, fd=None):
         attrs = {}
-        bsdflags = 0
+        flags = 0
         with backup_io('extended stat'):
-            if not self.nobsdflags:
-                bsdflags = get_flags(path, st, fd=fd)
+            if not self.noflags:
+                flags = get_flags(path, st, fd=fd)
             xattrs = xattr.get_all(fd or path, follow_symlinks=False)
             acl_get(path, attrs, st, self.numeric_owner, fd=fd)
         if xattrs:
             attrs['xattrs'] = StableDict(xattrs)
-        if bsdflags:
-            attrs['bsdflags'] = bsdflags
+        if flags:
+            attrs['bsdflags'] = flags
         return attrs
 
     def stat_attrs(self, st, path, fd=None):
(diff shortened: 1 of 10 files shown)


src/borg/archive.py
```diff
@@ -846,7 +846,7 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
             if warning:
                 set_ec(EXIT_WARNING)
             # bsdflags include the immutable flag and need to be set last:
-            if not self.nobsdflags and 'bsdflags' in item:
+            if not self.noflags and 'bsdflags' in item:
```
ThomasWaldmann: hmm, while we have the less misleading "flags / noflags" on the UI / in the docs, we still have "bsdflags" in the items.

not sure whether we should change this or not, maybe needs some thoughts, esp. about compatibility.

src/borg/archiver.py
```diff
@@ -175,7 +175,7 @@ def with_archive(method):
     def wrapper(self, args, repository, key, manifest, **kwargs):
         archive = Archive(repository, key, manifest, args.location.archive,
                           numeric_owner=getattr(args, 'numeric_owner', False),
-                          nobsdflags=getattr(args, 'nobsdflags', False),
+                          noflags=(getattr(args, 'nobsdflags', False) or getattr(args, 'noflags', False)),
```
ThomasWaldmann: outer parens not needed.

src/borg/archiver.py
```diff
@@ -557,7 +557,7 @@ def create_inner(archive, cache, fso):
 
         self.output_filter = args.output_filter
         self.output_list = args.output_list
-        self.nobsdflags = args.nobsdflags
+        self.noflags = (args.nobsdflags or args.noflags)
```
ThomasWaldmann: outer parens not needed.hh

src/borg/archiver.py
```diff
@@ -574,7 +574,7 @@ def create_inner(archive, cache, fso):
                                   chunker_params=args.chunker_params, start=t0, start_monotonic=t0_monotonic,
                                   log_json=args.log_json)
                 metadata_collector = MetadataCollector(noatime=not args.atime, noctime=args.noctime,
-                    nobsdflags=args.nobsdflags, numeric_owner=args.numeric_owner, nobirthtime=args.nobirthtime)
+                    noflags=(args.nobsdflags or args.noflags), numeric_owner=args.numeric_owner, nobirthtime=args.nobirthtime)
```
ThomasWaldmann: outer parens not needed.

src/borg/archiver.py
```diff
@@ -2370,7 +2370,8 @@ def do_subcommand_help(self, parser, args):
     def preprocess_args(self, args):
         deprecations = [
             # ('--old', '--new' or None, 'Warning: "--old" has been deprecated. Use "--new" instead.'),
-            ('--noatime', None, 'Warning: "--noatime" has been deprecated because it is the default now.')
+            ('--noatime', None, 'Warning: "--noatime" has been deprecated because it is the default now.'),
+            ('--nobsdflags', None, 'Warning: "--nobsdflags" has been deprecated because it is misleading. Use --noflags instead.')
```
ThomasWaldmann: remove the "because...".

it might not be true for linux users who have read the docs.
also it is for sure not be true for bsd users.

src/borg/archiver.py
```diff
@@ -3525,8 +3530,12 @@ def define_borg_mount(parser):
                                help='do not actually change any files')
         subparser.add_argument('--numeric-owner', dest='numeric_owner', action='store_true',
                                help='only obey numeric user and group identifiers')
+        # --nobsdflags is misleading and deprecated.
+        # use --noflags as a replacement.
         subparser.add_argument('--nobsdflags', dest='nobsdflags', action='store_true',
-                               help='do not extract/set bsdflags (e.g. NODUMP, IMMUTABLE)')
+                               help='DEPRECATED: do not extract/set bsdflags (e.g. NODUMP, IMMUTABLE)')
```
ThomasWaldmann: instead of a source comment, rather point to --noflags from here, so it is visible on the UI.

src/borg/archiver.py
```diff
@@ -3132,8 +3133,12 @@ def define_borg_mount(parser):
                               help='do not store ctime into archive')
         fs_group.add_argument('--nobirthtime', dest='nobirthtime', action='store_true',
                               help='do not store birthtime (creation date) into archive')
+        # --nobsdflags is misleading and deprecated.
+        # use --noflags as a replacement.
         fs_group.add_argument('--nobsdflags', dest='nobsdflags', action='store_true',
-                              help='do not read and store bsdflags (e.g. NODUMP, IMMUTABLE) into archive')
+                              help='DEPRECATED: do not read and store bsdflags (e.g. NODUMP, IMMUTABLE) into archive')
```
ThomasWaldmann: instead of a source comment, rather point to --noflags from here, so it is visible on the UI.

src/borg/archive.py
fantasya-pbem: Yes, I intentionally left it untouched to not break compatibility.

src/borg/archiver.py
fantasya-pbem: Yes. I saw the --noatime comments few lines above and added some comments, too.

src/borg/archiver.py
```diff
@@ -3132,8 +3133,12 @@ def define_borg_mount(parser):
                               help='do not store ctime into archive')
         fs_group.add_argument('--nobirthtime', dest='nobirthtime', action='store_true',
                               help='do not store birthtime (creation date) into archive')
+        # --nobsdflags is misleading and deprecated.
+        # use --noflags as a replacement.
```
ThomasWaldmann: as this is now in help text, maybe we do not need the comment at all.

src/borg/archiver.py
```diff
@@ -3525,8 +3530,12 @@ def define_borg_mount(parser):
                                help='do not actually change any files')
         subparser.add_argument('--numeric-owner', dest='numeric_owner', action='store_true',
                                help='only obey numeric user and group identifiers')
+        # --nobsdflags is misleading and deprecated.
+        # use --noflags as a replacement.
```
ThomasWaldmann: as this is now in help text, maybe we do not need the comment at all.

src/borg/archiver.py
fantasya-pbem: Removed and rebased everything into one commit.

src/borg/archiver.py
fantasya-pbem: Removed, too.

ThomasWaldmann:
Thanks!

Issue #1035, opened 2016-05-09
Allow operations on read-only filesystems


lxp:
I think it makes sense to allow repository checks on read-only filesystems.
I would already be happy, if I could force borg to not try to lock the repository during a check (with all it's implications).
Especially, in case of a check without repair, the worst thing that could happen should be that the check fails. Or am I wrong?

```
$ borg check /srv/repository
Failed to create/acquire the lock /srv/repository/lock.exclusive ([Errno 30] Read-only file system: '/srv/repository/lock.exclusive').
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1498, in main
    exit_code = archiver.run(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 1452, in run
    return args.func(args)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 72, in wrapper
    with repository:
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 80, in __enter__
    self.open(self.path, self.exclusive, lock_wait=self.lock_wait, lock=self.do_lock)
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 161, in open
    self.lock = UpgradableLock(os.path.join(path, 'lock'), exclusive, timeout=lock_wait).acquire()
  File "/usr/lib/python3/dist-packages/borg/locking.py", line 261, in acquire
    with self._lock:
  File "/usr/lib/python3/dist-packages/borg/locking.py", line 118, in __enter__
    return self.acquire()
  File "/usr/lib/python3/dist-packages/borg/locking.py", line 141, in acquire
    raise LockFailed(self.path, str(err)) from None
borg.locking.LockFailed: ('/srv/repository/lock.exclusive', "[Errno 30] Read-only file system: '/srv/repository/lock.exclusive'")

Platform: Linux debian 4.5.0-0.bpo.1-amd64 #1 SMP Debian 4.5.1-1~bpo8+1 (2016-04-20) x86_64 
Linux: debian 8.4 
Borg: 1.0.2  Python: CPython 3.4.2
PID: 8075  CWD: /root
sys.argv: ['/usr/bin/borg', 'check', '/srv/repository']
SSH_ORIGINAL_COMMAND: None
```

ThomasWaldmann:
Currently not possible. A writer wait until there are no readers, readers wait until there is no writer.

lxp:
[...]

PR #5048, opened 2020-03-19
Option to bypass locking to use sensible borg commands with read-only repositories

finefoot:
Linked to #1035. It makes sense wanting to mount read-only repos - and alternative solutions like copying to temp first or an overlay mount are unnecessarily tedious or complex.

Suggested explanation in the docs for `--bypass-lock`:

> Option ``--bypass-lock`` allows you to access the repository while bypassing borg's locking mechanism. This is necessary if your repository is on a read-only storage where you don't have write permissions or capabilities and therefore cannot acquire a lock. Examples are repositories stored on a Bluray disc or a read-only network storage. Avoid this option if you are able to use locks as it's the safer way; see the warning below.
> 
> **Note:** If you do use ``--bypass-lock``, you are responsible to ensure that no other borg instances have write access to the repository. Otherwise, you might experience errors and read broken data if changes to that repository are being made at the same time.

### DIFF
--- a/docs/usage/general.rst
+++ b/docs/usage/general.rst
@@ -31,6 +31,20 @@ All Borg commands share these options:
 
 .. include:: common-options.rst.inc
 
+Option ``--bypass-lock`` allows you to access the repository while bypassing
+borg's locking mechanism. This is necessary if your repository is on a read-only
+storage where you don't have write permissions or capabilities and therefore
+cannot create a lock. Examples are repositories stored on a Bluray disc or a
+read-only network storage. Avoid this option if you are able to use locks as
+that is the safer way; see the warning below.
+
+.. warning::
+
+    If you do use ``--bypass-lock``, you are responsible to ensure that no other
+    borg instances have write access to the repository. Otherwise, you might
+    experience errors and read broken data if changes to that repository are
+    being made at the same time.
+
 Examples
 ~~~~~~~~
 ::
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -134,9 +134,19 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
         if create:
             compatibility = Manifest.NO_OPERATION_CHECK
 
+    # To process the `--bypass-lock` option if specified, we need to
+    # modify `lock` inside `wrapper`. Therefore we cannot use the
+    # `nonlocal` statement to access `lock` as modifications would also
+    # affect the scope outside of `wrapper`. Subsequent calls would
+    # only see the overwritten value of `lock`, not the original one.
+    # The solution is to define a place holder variable `_lock` to
+    # propagate the value into `wrapper`.
+    _lock = lock
+
     def decorator(method):
         @functools.wraps(method)
         def wrapper(self, args, **kwargs):
+            lock = getattr(args, 'lock', _lock)
             location = args.location  # note: 'location' must be always present in args
             append_only = getattr(args, 'append_only', False)
             storage_quota = getattr(args, 'storage_quota', None)
@@ -2554,6 +2564,9 @@ class Archiver:
                               help='Output one JSON object per log line instead of formatted text.')
             add_common_option('--lock-wait', metavar='SECONDS', dest='lock_wait', type=int, default=1,
                               help='wait at most SECONDS for acquiring a repository/cache lock (default: %(default)d).')
+            add_common_option('--bypass-lock', dest='lock', action='store_false',
+                              default=argparse.SUPPRESS,  # only create args attribute if option is specified
+                              help='Bypass locking mechanism')
             add_common_option('--show-version', dest='show_version', action='store_true',
                               help='show/log the borg version')
             add_common_option('--show-rc', dest='show_rc', action='store_true',
@@ -4328,6 +4341,12 @@ class Archiver:
         if func == self.do_create and not args.paths:
             # need at least 1 path but args.paths may also be populated from patterns
             parser.error('Need at least one PATH argument.')
+        if not getattr(args, 'lock', True):  # Option --bypass-lock sets args.lock = False
+            bypass_allowed = {self.do_check, self.do_config, self.do_diff,
+                              self.do_export_tar, self.do_extract, self.do_info,
+                              self.do_list, self.do_mount, self.do_umount}
+            if func not in bypass_allowed:
+                raise Error('Not allowed to bypass locking mechanism for chosen command')
         return args
 
     def prerun_checks(self, logger, is_serve):
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -308,8 +308,15 @@ class Repository:
                 # some python ports have no os.link, see #4901
                 logger.warning(link_error_msg)
 
-        with SaveFile(config_path) as fd:
-            config.write(fd)
+        try:
+            with SaveFile(config_path) as fd:
+                config.write(fd)
+        except PermissionError as e:
+            # error is only a problem if we even had a lock
+            if self.do_lock:
+                raise
+            logger.warning("%s: Failed writing to '%s'. This is expected when working on "
+                           "read-only repositories." % (e.strerror, e.filename))
 
         if os.path.isfile(old_config_path):
             secure_erase(old_config_path)
@@ -325,7 +332,7 @@ class Repository:
         return keydata.encode('utf-8')  # remote repo: msgpack issue #99, returning bytes
 
     def get_free_nonce(self):
-        if not self.lock.got_exclusive_lock():
+        if self.do_lock and not self.lock.got_exclusive_lock():
             raise AssertionError("bug in code, exclusive lock should exist here")
 
         nonce_path = os.path.join(self.path, 'nonce')
@@ -336,14 +343,21 @@ class Repository:
             return None
 
     def commit_nonce_reservation(self, next_unreserved, start_nonce):
-        if not self.lock.got_exclusive_lock():
+        if self.do_lock and not self.lock.got_exclusive_lock():
             raise AssertionError("bug in code, exclusive lock should exist here")
 
         if self.get_free_nonce() != start_nonce:
             raise Exception("nonce space reservation with mismatched previous state")
         nonce_path = os.path.join(self.path, 'nonce')
-        with SaveFile(nonce_path, binary=False) as fd:
-            fd.write(bin_to_hex(next_unreserved.to_bytes(8, byteorder='big')))
+        try:
+            with SaveFile(nonce_path, binary=False) as fd:
+                fd.write(bin_to_hex(next_unreserved.to_bytes(8, byteorder='big')))
+        except PermissionError as e:
+            # error is only a problem if we even had a lock
+            if self.do_lock:
+                raise
+            logger.warning("%s: Failed writing to '%s'. This is expected when working on "
+                           "read-only repositories." % (e.strerror, e.filename))
 
     def destroy(self):
         """Destroy the repository at `self.path`
@@ -504,7 +518,7 @@ class Repository:
 
     def prepare_txn(self, transaction_id, do_cleanup=True):
         self._active_txn = True
-        if not self.lock.got_exclusive_lock():
+        if self.do_lock and not self.lock.got_exclusive_lock():
             if self.exclusive is not None:
                 # self.exclusive is either True or False, thus a new client is active here.
                 # if it is False and we get here, the caller did not use exclusive=True although
--- a/src/borg/testsuite/__init__.py
+++ b/src/borg/testsuite/__init__.py
@@ -235,26 +235,66 @@ class BaseTestCase(unittest.TestCase):
             self._assert_dirs_equal_cmp(sub_diff, ignore_bsdflags=ignore_bsdflags, ignore_xattrs=ignore_xattrs, ignore_ns=ignore_ns)
 
     @contextmanager
-    def fuse_mount(self, location, mountpoint, *options):
-        os.mkdir(mountpoint)
-        args = ['mount', location, mountpoint] + list(options)
-        self.cmd(*args, fork=True)
-        self.wait_for_mount(mountpoint)
+    def fuse_mount(self, location, mountpoint=None, *options, **kwargs):
+        if mountpoint is None:
+            mountpoint = tempfile.mkdtemp()
+        else:
+            os.mkdir(mountpoint)
+        if 'fork' not in kwargs:
+            # For a successful mount, `fork = True` is required for
+            # the borg mount daemon to work properly or the tests
+            # will just freeze. Therefore, if argument `fork` is not
+            # specified, the default value is `True`, regardless of
+            # `FORK_DEFAULT`. However, leaving the possibilty to run
+            # the command with `fork = False` is still necessary for
+            # testing for mount failures, for example attempting to
+            # mount a read-only repo.
+            kwargs['fork'] = True
+        self.cmd('mount', location, mountpoint, *options, **kwargs)
+        self.wait_for_mountstate(mountpoint, mounted=True)
         yield
         umount(mountpoint)
+        self.wait_for_mountstate(mountpoint, mounted=False)
         os.rmdir(mountpoint)
         # Give the daemon some time to exit
-        time.sleep(.2)
+        time.sleep(0.2)
 
-    def wait_for_mount(self, path, timeout=5):
-        """Wait until a filesystem is mounted on `path`
-        """
+    def wait_for_mountstate(self, mountpoint, *, mounted, timeout=5):
+        """Wait until a path meets specified mount point status"""
         timeout += time.time()
         while timeout > time.time():
-            if os.path.ismount(path):
+            if os.path.ismount(mountpoint) == mounted:
                 return
-            time.sleep(.1)
-        raise Exception('wait_for_mount(%s) timeout' % path)
+            time.sleep(0.1)
+        message = 'Waiting for %s of %s' % ('mount' if mounted else 'umount', mountpoint)
+        raise TimeoutError(message)
+
+    @contextmanager
+    def read_only(self, path):
+        """Some paths need to be made read-only for testing
+
+        Using chmod to remove write permissions is not enough due to
+        the tests running with root privileges. Instead, the folder is
+        rendered immutable with chattr or chflags, respectively.
+        """
+        if sys.platform.startswith('linux'):
+            cmd_immutable = 'chattr +i "%s"' % path
+            cmd_mutable = 'chattr -i "%s"' % path
+        elif sys.platform.startswith(('darwin', 'freebsd', 'netbsd', 'openbsd')):
+            cmd_immutable = 'chflags uchg "%s"' % path
+            cmd_mutable = 'chflags nouchg "%s"' % path
+        elif sys.platform.startswith('sunos'):  # openindiana
+            cmd_immutable = 'chmod S+vimmutable "%s"' % path
+            cmd_mutable = 'chmod S-vimmutable "%s"' % path
+        else:
+            message = 'Testing read-only repos is not supported on platform %s' % sys.platform
+            self.skipTest(message)
+        try:
+            os.system(cmd_immutable)
+            yield
+        finally:
+            # Restore permissions to ensure clean-up doesn't fail
+            os.system(cmd_mutable)
 
 
 class changedir:
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -51,6 +51,7 @@ from ..helpers import flags_noatime, flags_normal
 from ..nanorst import RstToTextLazy, rst_to_terminal
 from ..patterns import IECommand, PatternMatcher, parse_pattern
 from ..item import Item, ItemDiff
+from ..locking import LockFailed
 from ..logger import setup_logging
 from ..remote import RemoteRepository, PathNotAllowed
 from ..repository import Repository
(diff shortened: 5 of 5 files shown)


finefoot:
@ThomasWaldmann What's your opinion on something like this? :) I think it would be really useful to have. Could be extended to `borg list`, too.

ThomasWaldmann:
If the repo is not locked by `borg mount`, another borg can do operations in the repo (like creating, deleting archives, compacting segment files, etc.).

The `borg mount` (fuse fs) code needs to be reviewed whether it can work correctly with such operations happening in parallel. E.g. look for cache coherency issues.

finefoot:
> The borg mount (fuse fs) code needs to be reviewed whether it can work correctly with such operations happening in parallel. E.g. look for cache coherency issues.

That is covered and explicitly excluded by:

> If you do use this option, you are responsible to ensure that no other borg instances have write access to the repository. Otherwise, you might experience errors and read broken data from your mounted repository if changes to the repository are being made at the same time.

The idea is not to provide an elaborate way to work *with* the locking mechanism but to willfully and knowingly *ignore* and bypass it if (and only if) there is no chance of coherency issues due to parallel operations. See the example to mount a repo from a Bluray disc. You wouldn't be able to modify that repo in parallel anyway.

finefoot:
@ThomasWaldmann By the way, I only marked this pull request as WIP because I didn't change the docs yet. But the introduced changes by 1f316da21782cf678bcf9331847192eb501061da are a working proof of concept: Checks are passing and with those changes I am able to successfully mount and read a read-only repo.

ThomasWaldmann:
Reviewed the change, looks fine for this use case. Please add it to the docs.

People having this use case (r/o media) will likely want to use all sorts of other borg read operations on their repos also, but now it only works for `borg mount`.

finefoot:
> People having this use case (r/o media) will likely want to use all sorts of other borg read operations on their repos also, but now it only works for borg mount.

:+1: I'll look into other commands that were requested in the mentioned issue thread to see if `--bypass-lock` could be applicable there, too.

> Please add it to the docs.

Will do. :+1: Will also add a regression test, no?

ThomasWaldmann:
Not sure how to test this use case, but try. :-)

finefoot:
> Not sure how to test this use case, but try. :-)

@ThomasWaldmann Indeed, you were right, it definitely wasn't straightforward - but I think I made it work. (At least for Linux, for now.) There was a similar test called `test_readonly_repository`. That test didn't work, though: Even if you'd remove the `@pytest.mark.xfail`, it would pass and not fail as intended. I believe that's because the test hasn't been updated when at some point testing switched to using root privileges? Doesn't matter, though. Anyway, due to root privileges, `chmod` is not enough to make the repository folder read-only, so the test is now using `chattr +i` to make the folder immutable. (Maybe bindfs is an platform-independent alternative, haven't tried that, yet.) I have moved the code for this into a separate context manager so it can be used by multiple test functions for the various borg commands in combination with `--bypass-lock`. Instead of fixing the original faulty `test_readonly_repository`, I have removed it altogether because it is superseded by the individual tests for `--bypass-lock`, currently `test_readonly_list` and `test_readonly_mount`.

Regarding the docs: I have pasted the same description for `--bypass-lock` into both entries `borg list` and `borg mount`. Maybe - especially since further borg commands should follow - it would make more sense to store the description in a constant and insert the constant where necessary? Where would be a good place for this? Or does it make sense to use `--bypass-lock` as a common option?

Option `--bypass-lock` works for `borg list` and `borg mount` so far. For now, I'll wait for your feedback.

src/borg/archiver.py
```diff
@@ -134,9 +134,19 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
...
             compatibility = Manifest.NO_OPERATION_CHECK
 
+    # To process the `--bypass-lock` option if specified, we need to
+    # modify `lock` inside `wrapper`. Therefore we cannot use the
+    # `nonlocal` statement to access `lock` as modifications would also
+    # affect the scope outside of `wrapper`. Subsequent calls would
+    # only see the overwritten value of `lock`, not the original one.
+    # The solution is to define a place holder variable `_lock` to
+    # propagate the value into `wrapper`.
+    _lock = lock
```
ThomasWaldmann: good explanation how you found this solution.

src/borg/archiver.py
```diff
@@ -2643,6 +2653,10 @@ def define_borg_mount(parser):
                                 help='stay in foreground, do not daemonize')
             parser.add_argument('-o', dest='options', type=str,
                                 help='Extra mount options')
+            parser.add_argument('--bypass-lock', dest='lock',
+                                action='store_false',
+                                default=argparse.SUPPRESS,
```
ThomasWaldmann: is this the only place where we use SUPPRESS? had to look it up in argparse docs.

maybe add a comment like "only create args.lock attribute if this option is specified".

src/borg/archiver.py
```diff
@@ -2694,6 +2708,20 @@ def define_borg_mount(parser):
           option is internally enforced by borg. "ignore_permissions" can be given to
           not enforce "default_permissions".
 
+        Option ``--bypass-lock`` allows you to access the repository while bypassing
+        borg's locking mechanism. This is necessary if your repository is on a read-only
+        storage where you don't have write permissions or capabilities and therefore
+        cannot acquire a lock. Examples are repositories stored on a Bluray disc or a
```
ThomasWaldmann: maybe `can't create a lock` because the word create makes it clearer why write access is needed.

src/borg/archiver.py
```diff
@@ -2694,6 +2708,20 @@ def define_borg_mount(parser):
...
+        read-only network storage. Avoid this option if you are able to use locks as
+        it's the safer way; see the warning below.
```
ThomasWaldmann: ... as **that** is the safer way ...

src/borg/archiver.py
```diff
@@ -2694,6 +2708,20 @@ def define_borg_mount(parser):
...
+
+        .. note::
+
```
ThomasWaldmann: is there also **warning** instead **note**?

src/borg/testsuite/__init__.py
```diff
@@ -256,6 +256,17 @@ def wait_for_mount(self, path, timeout=5):
             time.sleep(.1)
         raise Exception('wait_for_mount(%s) timeout' % path)
 
+    @contextmanager
+    def read_only(self, path):
+        """Some paths need to be made read-only for testing"""
+        try:
+            # Using chmod is not enough due to root privileges
+            os.system('chattr +i "%s"' % path)
```
ThomasWaldmann: this only works if OS and FS supports this.

maybe check result code of the call?

src/borg/testsuite/archiver.py
```diff
@@ -1528,17 +1529,44 @@ def test_corrupted_repository(self):
...
+                else:
+                    with pytest.raises((LockFailed, RemoteRepository.RPCError)) as excinfo:
+                        self.cmd('mount', self.repository_location, mountpoint)
+                    if isinstance(excinfo.value, RemoteRepository.RPCError):
+                        assert excinfo.value.exception_class == 'LockFailed'
+            finally:
+                os.rmdir(mountpoint)
+            # verify that mount works with read-only repo when using --bypass-lock
+            with self.fuse_mount(self.repository_location, mountpoint, '--bypass-lock'):
+                pass
```
ThomasWaldmann: this is somehow different way than in the block above. also, block above kills mountpoint in finally block.

src/borg/archiver.py
finefoot: :+1:

src/borg/archiver.py
finefoot: Indeed, it is. Didn't expect that, I'm using `SUPPRESS` quite often. Will add your suggested comment, can't hurt. :+1:

src/borg/archiver.py
finefoot: :+1:

src/borg/archiver.py
finefoot: :+1:

src/borg/archiver.py
finefoot: Yes, Sphinx/RST has `.. warning::` I only initially used `.. note::` because most other "warnings" are notes, too, like https://github.com/borgbackup/borg/blob/master/docs/usage/config.rst Changed it. :+1:

src/borg/testsuite/__init__.py
finefoot: > maybe check result code of the call?

:+1: Added an `NotImplementedError`.

src/borg/testsuite/archiver.py
finefoot: `self.fuse_mount` is the "complete" test including mkdir, mount, umount, rmdir. (Reusing that functiong from other tests.) The block above already fails at mount, so that's what is explicitly being tested. Do you want me to rewrite `fuse_mount` so it works with testing a failed mount?

src/borg/archiver.py
```diff
@@ -3823,6 +3851,21 @@ def define_borg_mount(parser):
...
+        that is the safer way; see the warning below.
+
+        .. warning::
+
+            If you do use ``--bypass-lock``, you are responsible to ensure that no other
+            borg instances have write access to the repository. Otherwise, you might
+            experience errors and read broken data if changes to that repository are
+            being made at the same time.
+
+
```
ThomasWaldmann: I guess we do not want to copy that text to all borg subcommands that should support --bypass-lock.
Guess we should make it a global options even if not all subcommands will support it.

src/borg/testsuite/archiver.py
ThomasWaldmann: i just wondered why it is not calling via self.cmd.

src/borg/archiver.py
finefoot: Do you mean you don't want to insert the text itself multiple times into the code? (We could define `BYPASS_LOCK_HELP_TEXT` somewhere and add that to the epilog/help of each command.) Or do you not want to display the text for each command as it would clutter the help page with that big warning which only concers a few people?

src/borg/testsuite/archiver.py
finefoot: It is calling via `self.cmd` but by using `self.fuse_mount` ;) So that is resolved? Or you do suggest changes to clarify?

src/borg/archiver.py
ThomasWaldmann: well, primarily i do not want n-fold duplication within the source.

but also, if something is needed frequently, that's also a reason for a global option, even if we can't/won't support it for all commands.

ThomasWaldmann: needed frequently -> supported by a lot of commands

finefoot: Yup, agreed. I'll take a look how to move it to the common options.

src/borg/testsuite/__init__.py
```diff
@@ -256,6 +256,30 @@ def wait_for_mount(self, path, timeout=5):
...
+        Using chmod to remove write permissions is not enough due to
+        the tests running with root privileges. Instead, the folder is
+        rendered immutable with chattr or chflags, respectively.
+        """
+        if sys.platform.startswith('linux'):
+            cmd_immutable = 'chattr +i "%s"' % path
+            cmd_mutable = 'chattr -i "%s"' % path
+        elif sys.platform.startswith('darwin'):
+            cmd_immutable = 'chflags schg "%s"' % path
+            cmd_mutable = 'chflags noschg "%s"' % path
```
finefoot: BSD has `chflags` to make folders immutable so tests should now be working on OS X, too.

finefoot: Yup, (even though the job eventually times out) the relevant tests pass:

>     .tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py::ArchiverTestCase::test_readonly_list 
>     [gw0] [ 10%] PASSED .tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py::ArchiverTestCase::test_readonly_list 
>     .tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py::ArchiverTestCase::test_readonly_mount 
>     [gw0] [ 11%] PASSED .tox/py35/lib/python3.5/site-packages/borg/testsuite/archiver.py::ArchiverTestCase::test_readonly_mount 

https://travis-ci.org/github/borgbackup/borg/jobs/667056832

src/borg/testsuite/__init__.py
```diff
@@ -235,26 +235,63 @@ def _assert_dirs_equal_cmp(self, diff, ignore_bsdflags=False, ignore_xattrs=Fals
...
+        if mountpoint is None:
+            mountpoint = tempfile.mkdtemp()
+        else:
+            os.mkdir(mountpoint)
+        if 'fork' not in kwargs:
+            # For a successful mount, `fork = True` is required. The
+            # option to run the command with `fork = False` is still
+            # necessary to test mount failures like attempting to mount
+            # a read-only repo.
+            kwargs['fork'] = True
```
finefoot: I guess it has something to do with running the daemon that `fork=True` is required?

src/borg/testsuite/archiver.py
finefoot: > i just wondered why it is not calling via self.cmd.

Reflecting on that, I think you're right. It's unnecessarily complex to use different calls. I have modified `self.fuse_mount` to make it accept keyword arguments which are then being passed on to `self.cmd`. This works and make the `test_readonly_mount` look cleaner. :+1:

src/borg/testsuite/__init__.py
```diff
@@ -235,26 +235,63 @@ def _assert_dirs_equal_cmp(self, diff, ignore_bsdflags=False, ignore_xattrs=Fals
...
         timeout += time.time()
         while timeout > time.time():
-            if os.path.ismount(path):
+            if os.path.ismount(mountpoint) != umount:
                 return
-            time.sleep(.1)
-        raise Exception('wait_for_mount(%s) timeout' % path)
+            time.sleep(0.1)
+        message = 'Waiting for %s of %s' % ('umount' if umount else 'mount', mountpoint)
+        raise TimeoutError(message)
```
finefoot: I had to modify `wait_for_mount` a little bit, too. It's now possible to wait for mounting as well as unmounting. I also replaced the generic `Exception` with `TimeoutError` which seems more suitable.

finefoot:
- Currently, `--bypass-lock` is allowed for: `borg extract`, `borg list`, `borg mount` and `borg umount`. I still need to add a test for `borg extract`, but that should be straightforward, now.
- I will add `borg info`, which seems useful. Already tested locally, don't expect issues there.
- The following commands are not sensible to use with `--bypass-lock` as their intention is to write changes to the repo: `borg benchmark crud`, `borg break-lock`, `borg create`, `borg delete`, `borg init`, `borg key change-passphrase`, `borg key import`, `borg prune`, `borg rename`, `borg upgrade`, `with-lock`
- Still have to look into `borg check`, `borg diff`, `borg key export`, `borg export-tar`, `borg config`

finefoot:
Also, same issue here with the missing link to the (passing) Travis CI checks for the most recent commit: https://travis-ci.org/github/borgbackup/borg/builds/667516614

> - Still have to look into `borg check`, `borg config`, `borg diff`, `borg export-tar`, `borg key export`.

@ThomasWaldmann Do you have an opinion regarding those commands? I think they're not essential to use with a read-only repo.

finefoot:
Regarding [my previous comment](https://github.com/borgbackup/borg/pull/5048#issuecomment-604930588): Yeah, I think this pull request shouldn't include support for `borg check`, `borg config`, `borg diff`, `borg export-tar`, `borg key export`, for now. I think `borg list`, `borg info`, `borg extract` and `borg mount` cover the essential actions with read-only repos. Do you agree? We can always add that feature in another pull request, later.

@ThomasWaldmann Do you want to review the current improvements on `fuse_mount` and `wait_for_mount` or should I finish `borg extract` and `borg info`, first, so you can do everything at once?

finefoot:
Okay, I've had a little bit of spare time and finished what was left to do here. Option `--bypass-lock` works for read-only commands `borg check`, `borg config`, `borg diff`, `borg export-tar`, `borg extract`, `borg info`, `borg list`, `borg mount` and `borg umount`. So I think this pull request is now implementing everything that has been requested by #1035.

Travis CI checks are passing, too, so I'll wait for your feedback now. Just let me know if I should do something like squash the commits before your review. Thanks :)

docs/usage/general.rst
```diff
@@ -31,6 +31,20 @@ All Borg commands share these options:
 
 .. include:: common-options.rst.inc
 
+Option ``--bypass-lock`` allows you to access the repository while bypassing
```
ThomasWaldmann, line 34: i thought about the option name while reading through the changes.

i rather think it should be called `--read-only-repo` or so, because it does not only suppress lock creation but also all kinds of other write ops to the repo dir.

finefoot: Hmm, yes, I understand where you're coming from. However, two counter points to consider:

- I went with `--bypass-lock` because I thought that describes most accurately what happens: it actually ignores and bypasses the locking mechanism. If you'd use `borg with-lock` to indefinitely keep a lock on the repo, you could still access it with `--bypass-lock`. Using this "bypassing feature" is just a tool to also provide access a read-only repo.
- > i rather think it should be called --read-only-repo or so, because it does not only suppress lock creation but also all kinds of other write ops to the repo dir

    The option (and the changeset it introduces) actually doesn't suppress other writing operations at all. It only suppresses errors if there are permission conflicts. If the repo is not read-only, there is effectively no difference between using `--bypass-lock` or not. (Apart from not using a lock for the operations.)

finefoot: Oh, you meant "suppress write operations" because of the manually raised exception?

finefoot: Do you mind clarifying what you meant?

I deliberately chose `NotImplementedError` in [those respective lines](https://github.com/borgbackup/borg/pull/5048/files#diff-48f2b9140a35ffd45a46ccfbe63931a6R4344-R4350) which check for "not allowed commands". Would you rather omit that check altogether and have the user simply get the inevitable `PermissionError` if they're using `--bypass-lock` with a write command like `borg create`? (Note, it won't be `LockFailed`, but `PermissionError`.)

src/borg/testsuite/__init__.py
ThomasWaldmann: somehow a signature like this would be more self-explaining:
```
wait_for_mountstate(mountpoint, *, mounted, timeout=5)
```

calls:
```
wait_for_mountstate('/mnt', mounted=True)
wait_for_mountstate('/mnt', mounted=False)
```

docs/usage/general.rst
ThomasWaldmann: looking at this again now...

src/borg/testsuite/__init__.py
```diff
@@ -235,26 +235,63 @@ def _assert_dirs_equal_cmp(self, diff, ignore_bsdflags=False, ignore_xattrs=Fals
...
+            # For a successful mount, `fork = True` is required. The
+            # option to run the command with `fork = False` is still
+            # necessary to test mount failures like attempting to mount
+            # a read-only repo.
+            kwargs['fork'] = True
+        self.cmd('mount', location, mountpoint, *options, **kwargs)
         self.wait_for_mount(mountpoint)
         yield
         umount(mountpoint)
+        self.wait_for_mount(mountpoint, umount=True)
```
ThomasWaldmann, line 253: hmm, is this a bugfix?

isn't `umount()` expected to return **after** the fs is actually umounted?

docs/usage/general.rst
ThomasWaldmann: well, i rather meant "silently fail" than "suppress" (e.g. in save_config).

i am wondering a bit about the nonce and the start transaction stuff - is that used by any of the supported commands? sounds more like for repo-writing ops.

ThomasWaldmann: About the NotImplementedError exception in case of an unsupported command: we should exit cleanly in such a case. no long traceback as this is expected and we do not need to analyze a bug here.

src/borg/testsuite/__init__.py
ThomasWaldmann: guess so.

src/borg/testsuite/__init__.py
```diff
@@ -235,26 +235,63 @@ def _assert_dirs_equal_cmp(self, diff, ignore_bsdflags=False, ignore_xattrs=Fals
...
+        """
+        if sys.platform.startswith('linux'):
+            cmd_immutable = 'chattr +i "%s"' % path
+            cmd_mutable = 'chattr -i "%s"' % path
+        elif sys.platform.startswith('darwin'):
+            cmd_immutable = 'chflags schg "%s"' % path
+            cmd_mutable = 'chflags noschg "%s"' % path
+        else:
+            message = 'Testing read-only repos is currently not supported on your platform'
+            raise NotImplementedError(message)
```
ThomasWaldmann, line 288: i will run those tests on misc. BSDs and openindiana, so does that mean my tests fill fail?

they rather should be skipped (or xfailed?) if we can not support this way of testing.

docs/usage/general.rst
finefoot: > i am wondering a bit about the nonce and the start transaction stuff - is that used by any of the supported commands?

Yes, `borg check` uses those. There is a call to `save_config` [here](https://github.com/borgbackup/borg/blob/master/src/borg/repository.py#L979) for the repository check. And a call to `commit_nonce_reservation` happens during `rebuild_refcounts` [here](https://github.com/borgbackup/borg/blob/master/src/borg/archive.py#L1427) for the archives check.

finefoot: > About the NotImplementedError exception in case of an unsupported command: we should exit cleanly in such a case. no long traceback as this is expected and we do not need to analyze a bug here.

What does that mean, effectively? :)

src/borg/testsuite/__init__.py
finefoot: I agree with your comment. Did you mark this as resolved by accident?

src/borg/testsuite/__init__.py
finefoot: Surely, those platforms have a similar command? Wouldn't that make more sense? We'll just need their [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) values.

src/borg/testsuite/__init__.py
finefoot: Since [`helpers.fs.umount`](https://github.com/borgbackup/borg/blob/master/src/borg/helpers/fs.py#L286) simply calls `fusermount` or `umount` and doesn't check the result, I thought it wouldn't hurt to add that. I'm not familiar with the internals of either of those commands, so I can't say if the unmount has always succeeded (or failed) when they return. So a "timeout" might not be necessary, but the check for mount state of the mountpoint seemed useful either way.

finefoot:
Thanks for reviewing :)

src/borg/testsuite/__init__.py
ThomasWaldmann: ok, i'll try to do some research for the other platforms.

src/borg/testsuite/__init__.py
finefoot: :+1: I have improved the explanation a little bit. This should now clear up any confusion, why the `fork` argument is handled so particulary.

src/borg/testsuite/__init__.py
finefoot: Like I commented above, `chflags` is a BSD command so that should not only work for OS X but for FreeBSD etc., too. However, for OpenIndiana/Solaris, I don't know. Is this the first time that there is a platform-depending issue? Instead of `NotImplementedError`, we could do `pytest.skip`, of course, as a quick fix, if you don't want to include that test for some platforms.

src/borg/testsuite/__init__.py
ThomasWaldmann: it doesn't show as resolved here.

finefoot: Only because I clicked on "unresolve" before commenting. :wink:  (I wasn't sure if you'd get a notification for a comment in a resolved thread.)

I'll change the function. :+1:

src/borg/testsuite/__init__.py
ThomasWaldmann: openindiana:

sys.platform == 'sunos5'

chmod S+vimmutable file / chmod S-vimmutable file

ThomasWaldmann: openbsd:

sys.platform == 'openbsd6'

`chflags schg` is a one-way ticket and only for root, so we can't use that.
`chflags (no)uchg` works.

ThomasWaldmann: netbsd:

sys.platform == 'netbsd7'

schg: similar to openbsd, except for the one-way ticket.
(no)uchg works.

ThomasWaldmann: freebsd:

sys.platform == 'freebsd10'

same as netbsd.

ThomasWaldmann: for all other platforms, skip the test.

docs/usage/general.rst
ThomasWaldmann: try: `raise Error("error message for user")`

src/borg/testsuite/__init__.py
ThomasWaldmann: ok, so guess we at least would notice clearly if the umount did not really work. good.

docs/usage/general.rst
ThomasWaldmann: ok, the not working save_config (for read-only case) means that incremental checks won't work, which is expected if we can't remember how far we got.

the rebuild_refcounts stuff is only commit in repair mode.

finefoot: `helpers.Error` calls `format` but `__doc__` doesn't contain any `{}` :thinking:  https://github.com/borgbackup/borg/blob/master/src/borg/helpers/errors.py#L6

src/borg/testsuite/__init__.py
finefoot: :+1:

finefoot: https://github.com/borgbackup/borg/pull/5048/commits/5b52bcd0ae3e5ab067842d4600d1d164e97d1a1d Checks are passing, too

src/borg/repository.py
```diff
@@ -308,8 +308,13 @@ def save_config(self, path, config):
...
 
-        with SaveFile(config_path) as fd:
-            config.write(fd)
+        try:
+            with SaveFile(config_path) as fd:
+                config.write(fd)
+        except PermissionError:
+            # error is only a problem if we even had a lock
+            if self.do_lock:
+                raise
```
finefoot, line 317: Also, what do you think about logging the silently failing PermissionError as a warning?

```
except PermissionError as e:
    # error is only a problem if we even had a lock
    if self.do_lock:
        raise
    logger.warning('%s This is expected when working on read-only repositories.' % e)
```

finefoot: There might be a warning during borg check anyway: "Failed to securely erase old repository config file (hardlinks not supported>). Old repokey data, if any, might persist on physical storage."

ThomasWaldmann: logging a warning sounds good. is there a `.` at the end of the exception message so this looks correct?

docs/usage/general.rst
ThomasWaldmann: guess it is because we only used subclasses of this yet (or did not give a msg).

src/borg/repository.py
finefoot: The filename in single quotes:

> [Errno 13] Permission denied: '/home/dummy/test/config.tmp' This is expected when working on read-only repositories.
> [Errno 13] Permission denied: '/home/dummy/test/nonce.tmp' This is expected when working on read-only repositories.

Maybe the `[Errno 13]` is confusing. Since PermissionError is an OSError, we could also assemble a custom message from `e.strerror` and `e.filename` https://docs.python.org/3/library/exceptions.html#OSError

docs/usage/general.rst
finefoot: Two occurrences:

- https://github.com/borgbackup/borg/blob/master/src/borg/archiver.py#L1707
- https://github.com/borgbackup/borg/blob/master/src/borg/helpers/manifest.py#L107

src/borg/repository.py
finefoot: With https://github.com/borgbackup/borg/pull/5048/commits/69a985ab14f459140b7aad12e72f360e9b29d60f, output would look something like this:

>     $ borg check --verify-data --bypass-lock --debug test
>     using builtin fallback logging configuration
>     37 self tests completed in 0.10 seconds
>     Starting repository check
>     Verified integrity of /home/dummy/test/index.12
>     Read committed index of transaction 12
>     Segment transaction is    12
>     Determined transaction is 12
>     Found 13 segments
>     Failed to securely erase old repository config file (hardlinks not supported>). Old repokey data, if any, might persist on physical storage.
>     Permission denied: Failed writing to '/home/dummy/test/config.tmp'. This is expected when working on read-only repositories.
>     finished segment check at segment 12
>     Failed to securely erase old repository config file (hardlinks not supported>). Old repokey data, if any, might persist on physical storage.
>     Permission denied: Failed writing to '/home/dummy/test/config.tmp'. This is expected when working on read-only repositories.
>     Starting repository index check
>     Index object count match.
>     Finished full repository check, no problems found.
>     Starting archive consistency check...
>     Verified integrity of /home/dummy/test/index.12
>     Starting cryptographic data integrity verification...
>     Finished cryptographic data integrity verification, verified 8 chunks with 0 integrity errors.
>     TAM-verified manifest
>     Analyzing archive one (1/2)
>     Permission denied: Failed writing to '/home/dummy/test/nonce.tmp'. This is expected when working on read-only repositories.
>     Analyzing archive two (2/2)
>     Archive consistency check complete, no problems found.

docs/usage/general.rst
finefoot: Those two occurrences basically never happen - still shouldn't be like this, no? Concerns ErrorWithTraceback, too:

```
>>> raise ErrorWithTraceback("custom message")
__main__.ErrorWithTraceback: like Error, but show a traceback also
```

Opened a pull request to fix it. #5100

ThomasWaldmann:
OK, ready to merge?

finefoot:
I still need to turn NotImplementedError into Error, now that #5100 is merged.

finefoot:
:+1: Checks passed and commits squashed - looks like it's ready to merge.

I appreciate the feedback and reviewing process during the course of this pull request. :)

finefoot:
Oh, my bad. One of the OS X checks randomly failed. Do you want to restart it to be on the safe side?

ThomasWaldmann:
Restarted, let's see.

ThomasWaldmann:
Thanks!

Issue #2929, opened 2017-08-11
Provide a way to completely delete data from the repository


realityexists:
Sometimes data needs to be securely deleted. If it was backed up with borg there needs to be a way to also delete it from any backups. This data may be an entire file (e.g. a private key file) or a small part of a large file (e.g. a password inadvertently included in source code and later edited out).

Currently, the user can run "borg delete" or "borg delete" to delete any archives containing the deleted data. The data is then logically deleted (it would not be restored by any borg restore commands), but it is unclear whether it still exists in any borg segment files. (Apparently borg 1.1 introduces an optimisation where small changes don't cause segments to be rewritten, for example.)

Please provide and document a way to ensure that deleted data is truly gone from the repository. This may be a new "garbage collection" command that forces all segments containing deleted data to be rewritten or some combination of existing commands/options.

ThomasWaldmann:
See also: #2195 

Having some params for the compaction command is easier for ad-hoc usage than e.g. editing of parameters in the repo config and then having to artificially trigger some write op just to have it call compact_segments() also.

enkore:
Secure deletion often doesn't really work [1]. This is aggravated by Borg's deduplicating nature, since you have to delete all archives that reference the data in question before it is deleted. This may be non-trivial in itself (and Borg has no real reverse mapping to help you with this) [2]. Only when the data is logically deleted, secure physical deletion becomes relevant.

[1] Encrypting data and throwing the key away to dispose of it is much simpler and more secure.
[2] A Borg Expert's way to do this is to hack `borg debug dump-archive` to dump chunk IDs (reasonable to merge into master) and then delete those chunks manually, running `check --repair` afterwards to mask the errors.

realityexists:
Absolutely, these are important points that should be documented, but I don't think they're reasons to not implement this.

Of course, the data should only be deleted when there are no archives left containing this data. That should be fairly obvious to the user (though still worth documenting). It's not always hard to know which archives contain it - maybe I know I just added it yesterday, so only the last daily archive needs to be deleted? Worst case scenario: assume it's in all archives until now, create one new archive without the data and delete all previous archives. This is still much easier and faster than deleting and re-creating the whole repository.

[...]

Issue #4674, opened 2019-07-03
Allow option to control compacting threshold


russelldavis:
Some other users have mentioned this would be useful:
- https://github.com/borgbackup/borg/issues/2985#issuecomment-375391451
- https://mail.python.org/pipermail/borgbackup/2016q4/000501.html

I would find it useful for the same reasons (to minimize churn when rcloning a Borg repo to S3).

fantasya-pbem:
@ThomasWaldmann I could try to add an option. But what is expected? A new option for ``borg compact`` that has an effect when this command is run?

ThomasWaldmann:
Yeah, guess it needs a new compact option.

Also, please note that `RemoteRepository.commit` also needs to be extended so it will also work for remote repos.

fantasya-pbem:
Yeah, my first-ever Python programming experience merged!

PR #5049, opened 2020-03-19
[DOCS] #2929 – Document how to completely delete data

fantasya-pbem:
Add new FAQ entry that lists two ways how to really delete files from archives.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -335,11 +335,19 @@ also supports the ``--glob-archives`` parameter.
 How do I remove files from an existing backup?
 ----------------------------------------------
 
-Say you now want to remove old logfiles because you changed your
-backup policy as described above. The only way to do this is to use
-the :ref:`borg_recreate` command to rewrite all archives with a
-different ``--exclude`` pattern. See the examples in the
-:ref:`borg_recreate` manpage for more information.
+A file is only removed from a BorgBackup repository if all archives that contain
+the file are deleted and the corresponding data chunks are removed from the
+repository There are two ways how to remove files from a repository.
+
+1. Use :ref:`borg_delete` to remove all archives that contain the files. This
+will of course delete everything in the archive, not only some files.
+
+2. If you really want to remove only some specific files, you can run the
+:ref:`borg_recreate` command to rewrite all archives with a different
+``--exclude`` pattern. See the examples in the manpage for more information.
+
+Finally, run :ref:`borg_compact` with the ``--threshold 0`` option to delete the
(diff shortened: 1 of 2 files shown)


docs/faq.rst
```diff
@@ -335,11 +335,18 @@ also supports the ``--glob-archives`` parameter.
...
 ----------------------------------------------
 
-Say you now want to remove old logfiles because you changed your
-backup policy as described above. The only way to do this is to use
-the :ref:`borg_recreate` command to rewrite all archives with a
-different ``--exclude`` pattern. See the examples in the
-:ref:`borg_recreate` manpage for more information.
+A file is only removed from a BorgBackup repository if all archives that contain
+the file are deleted and the corresponding data chunks are removed from the
+repository There are two ways how to remove files from a repository:
```
ThomasWaldmann: add a dot after repository.

docs/faq.rst
```diff
@@ -335,11 +335,18 @@ also supports the ``--glob-archives`` parameter.
...
+
+1. Use :ref:`borg_delete` to remove all archives that contain the files. This
+will of course delete everything in the archive, not only some files. Finally,
+run :ref:`borg_compact` to delete the archives' chunks from the repository.
+
+2. If you really want to remove only some specific files, you can run the
+:ref:`borg_recreate` command to rewrite all archives with a different
+``--exclude`` pattern. See the examples in the :ref:`borg_recreate` manpage for
+more information.
```
ThomasWaldmann: Hmm, these methods are not 100% safe (regarding the removal of the information from the repo's segment files).

They will of course logically remove it, but there is some chance that compact won't remove the unused chunks of the removed file because the respective segment file is under the compaction threshold. This can happen especially if the removed file is rather small.

ThomasWaldmann: to solve that, we need options to `borg compact` so we can force an arbitrary threshold (in this case: a 0 threshold).

fantasya-pbem: Do we have such an issue (compact with threshold) already? If not, I will create such an issue.

Of course, there is a chance that a chunk is not compacted/removed. I will add a note about this.

ThomasWaldmann: not sure whether we have an issue or not, please check and create one if needed.

fantasya-pbem: Oh dear, I am very forgetful... We had #4674 which I solved introducing the threshold parameter... I will update the PR accordingly.

fantasya-pbem:
I think I messed it up. Will make a new PR...

Issue #5000, opened 2020-03-08
add rendering docs to release todo


ThomasWaldmann:
avoids markup issues, e.g. in changelog / bug advisories.

milkey-mouse:
Happy 5000th issue!

milkey-mouse:
On a more serious note, isn't this already in the release checklist?

https://github.com/borgbackup/borg/blame/master/docs/development.rst#L322-L323

ThomasWaldmann:
that's not what i meant.

what i mean is rendering changes.rst to html (sphinx: make html) and looking at it, to avoid markup errors like in 1.1.11 release (in the security advisory part at the top of the file).

fantasya-pbem:
So you want to add something like "render changes.rst to html and check for markup errors" here: https://borgbackup.readthedocs.io/en/stable/development.html#creating-a-new-release
Did I understand correctly? Or do you want "render documentation with sphinx and check", which would be quite a lot to do.

PR #5056, opened 2020-03-20
[DOCS] #5000 – Add rendering docs to release todo

fantasya-pbem:
Backport from master, see #5045, including various other consistency changes.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -17,33 +17,33 @@ Contributions
 
 Some guidance for contributors:
 
-- discuss about changes on github issue tracker, IRC or mailing list
+- Discuss changes on the GitHub issue tracker, on IRC or on the mailing list.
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
The diff doesn't look like a clean backport.

fantasya-pbem:
I took over all the clean-up changes that happend in this file in master, but didn't touch the python versions.

docs/development.rst
```diff
@@ -5,9 +5,9 @@
 Development
 ===========
 
-This chapter will get you started with |project_name| development.
+This chapter will get you started with Borg development.
```
ThomasWaldmann: are this kind of changes present in other 1.1-maint docs files? if not, it should not be here either.

fantasya-pbem: You're right, the |project_name| thing is used everywhere. I will undo that change.

ThomasWaldmann:
Thanks!

PR #5066, opened 2020-04-01
Update prevalence of env vars to set config and cache paths

ThomasWaldmann:
re-applied #4951 to current master.

PR #4951, opened 2020-02-16
Update prevalence of env vars to set config and cache paths

finefoot:
See #4894

PR #5068, opened 2020-04-01
Backport of #5066 Update prevalence of env vars to set config and cache paths

finefoot:
See #5066, #4951 and #4894


ThomasWaldmann:
Thanks!

Issue #3687, opened 2018-03-13
Small file size and directory order information leakage


nadalle:
After talking to Thomas Waldmann on IRC, I'm opening a ticket mostly to document this issue.

Currently, borg's model for storing chunks in the repository leaks information about the size and (potentially) directory structure of small files. Size and directory structure information can often be used to determine information about the content of the files themselves, through various fingerprint / watermark attacks.

Files in borg are chunked, compressed, and then the chunks are stored in the repository using PUT commands. These PUT commands necessarily contain the size of the chunk being stored, and additionally the segment files in the repository contain these sizes in easily parsed form.

For small files, the size of the chunks prior to compression will typically just be the size of the file itself. Additionally, a great many modern file formats are not compressible, and in any event an attacker can simply compress the data too. In this way, the chunk sizes can leak information about the size of small files.

Additionally, chunks will tend to be stored in temporal order in their associated segment files. Since the order borg processed the files in will be some directory ordering, the temporal order in the repository should typically match a directory sort order, which leaks additional information.

As a hypothetical example of how size and directory structure information can be used as a data fingerprint, consider the common method of fingerprinting CD's based on their track length. Even if a CD was encoded to mp3, it's typically possible to convert file size back to track length by assuming a standard bitrate.

It should then be evident that a database of CD track lengths is sufficient to show that a directory of encrypted files contains the CD tracks with high confidence, just by comparing the ratios of the file sizes to the ratios of track lengths. This can be done without reading any of the data, just the file sizes.

Many other more advanced sorts of watermarking attacks are possible.

There are various potential ways to solve this information leak. A few ideas:
- Pad small files up to the chunk size.
- Bundle multiple files into a chunk somehow.
- More generally, differentiate between segment records (known to the remote server) and chunks (known to the client). Make the records all the same size.

Regardless of whether this is fixed any time soon, it seemed worth documenting better. The current documentation states "[...] all your files/dirs data and metadata are stored in their encrypted form into the repository" so a leak of file size and directory order information seems notable.

nadalle:
As an additional note, storing chunks post-compression leaks some information on large files too. For example, imagine I store a compressible file such as a VM image. After it gets chunked and compressed, an attacker can see that a sequence of blocks, each of which was compressible by a certain amount. If the attacker has a copy of the file, this ought to be sufficient to show that borg stored that large file.

ThomasWaldmann:
The small file fingerprinting is a somehow known problem.

[...]

PR #5070, opened 2020-04-02
Chunker permutation

ncleaton:
Make the buzhash chunker more resistant to fingerprinting by introducing a permutation of the buzhash table, as discussed at https://github.com/borgbackup/borg/issues/3687


ThomasWaldmann:
It would be useful, if someone else would also review this.

textshell:
On first look this seems to silently diverge on the chunking when this key is used in older borg versions? That sounds problematic.

src/borg/crypto/key.py
```diff
@@ -345,6 +350,35 @@ def id_hash(self, data):
...
+        pool >>= 4  # 211 bytes is 1688 bits, 4 bits more than we want
+        perm = list(range(256))
+        for i in range(256):
+            pool, offset = divmod(pool, 256-i)
+            j = i + offset
+            perm[i], perm[j] = perm[j], perm[i]
+
+        if pool == 0:
+            # the pool value was less than 256!, we have an unbiased choice
+            return bytes(perm)
```
ThomasWaldmann, line 376: could be also just `break`.

ThomasWaldmann (commented):
found one cosmetic thing while doing another review.

ThomasWaldmann:
@textshell looks to me like:
- if repo was initialized with old (e.g. 1.1.14), the PR code will use the null permutation and behave same, compute same chunks, no dedup issue, even if borg is switched between old and new versions.
- if repo was initialized with new PR code, the key material will be longer (132 bytes instead of 100 bytes), the new code would compute different chunks than the old code. So there would be a dedup issue, but we could just document that in changelog.

I have still to check how the old code deals with a key that is longer than expected. I guess it would just silently ignore that.

Do you think it is problematic if we just document "don't use borg 1.1.x on repos created with borg 1.2"?

ThomasWaldmann:
We could also set item.Key.version to 2 (was: 1), then borg 1.1.x code will reject working with a key made by borg 1.2.

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
I think we can close this:

- sorry that it did not get merged, but i didn't touch the chunker for a long time (especially since i did not feel very comfortable working with that big piece of C).
- since borg2 will be a breaking release and also since the 2 CDC chunker security papers, I felt that now is a good time to work on this. people will need to transfer their repos from borg1 to borg2 anyway and i recently added the code so it can re-chunk while doing that. so a chunker change will not "cost" much extra and the new chunker also does not need to be compatible with the old one.
- i refactored the chunker code, broke it into smaller submodules of a package, separated the reading/sparse/fmap handing into separate classes, translated the buzhash C to Cython.
- after that, copied buzhash to buzhash64 so the hash width is 64bits now.
- also changed the table init: removed the hardcoded table, the XOR seed and replaced it with some cryptographic code generating a table from scratch, deriving the chunker key from the ID key.

So, guess we won't change how the buzhash code works, but rather transition to buzhash64.

See: #8903

PR #5082, opened 2020-04-05
[DOCS] #4883 – Doc downplays severity of Nonce reuse issue

### LINKED ISSUE #4883: Borg documentation downplays severity of Nonce reuse issue
The attack scheme itself is known to Borg community: repository controlled by malicious party may provide clients with up-to-date copies of the repository, thus effectively forcing Nonce reuse, which breaks encryption in [CTR mode](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Counter_(CTR)) (to complete the attack malicious party also needs some plain text material from the client, but the latter is relatively easy to guess as each client stores some bits of the Internet, and vice versa; for instance it is known that the client has a certain Python program named "borg").

However documentation doesn't bring the issue upfront, making it harder for average users to fully understand the level of security they get.

As of this writing, section "[Can I backup from multiple servers into a single repository?](https://borgbackup.readthedocs.io/en/latest/faq.html#can-i-backup-from-multiple-servers-into-a-single-repository)" answers

> Yes, but [long description of security-unrelated matters]...

whereas the correct answer would be

> No, unless one of the following is true:
> - repository is trusted
> - privacy/confidentiality is not a concern (rarely the case)
> - you run backup process via a wrapper that follows the protocol (using secure communication channel):
>   0: among all clients sharing same repo select a master (this step is performed only once)
>   1: on master acquire _client-side_ exclusive lock that will prevent parallel execution of backup (read-only executions are OK); if master is unreachable then abort
>   2: from _each_ client sharing same repo fetch client's security database; if any host is unreachable then release master lock and abort (if master host became unreachable be extremely careful when removing the lock manually later - failing to ensure that no backup is running is a security breach)
>   3: across all security databases (fetched ones and one's own) find the maximum used Nonce value
>   4: update own security database with the next unused Nonce value
>   5: perform the backup
>   6: release master lock; if master is unreachable see warning in 2.
> 
> However [long description of security-unrelated matters]...

"The protocol is too complex, will it be enough if we just never run two backups in parallel?" - No.

Actually, the very [next FAQ question](https://borgbackup.readthedocs.io/en/latest/faq.html#can-i-copy-or-synchronize-my-repo-to-another-location) talks about dangers of repo copies, however average user may not see the connection with multi-client case, so solid "Yes" to the first question is misleading at best.

Regarding Nonce synchronization the end of section "Internals **(!)**/Security/[Encryption](https://borgbackup.readthedocs.io/en/latest/internals/security.html#encryption)" mentiones:

> If the repository is untrusted, but a trusted synchronization channel exists between clients, the security database could be synchronized between them over said trusted channel.  _This is not part of Borg’s functionality._

(I italicized the invitation to put your dirty hands into Borg's internal data structures which is required by steps 3-4 of the protocol above) whereas it should say:

> If the repository is untrusted, there _must_ exists trusted channel between clients and security databases _must be_ synchronized with the outlined protocol.

To understand protocol requirements think of malicious repo providing two clients with separate up-to-date copies of the initially shared repo.  We need to synchronize Nonce allocations to avoid value reuse (steps 3-4 above).  For this we need to fetch security databases from all peers (step 2) and be sure there's no race that will outdate fetched copies (step 1; since each client observes its own copy of the repo repo-side locks won't work, so we need client-side locking).  To avoid [split brain](https://en.wikipedia.org/wiki/Split-brain_(computing)) situation all peers have to be reachable, thus single master lock is sufficient (step 0).

Finally, it's worth pointing out that not only multi-client setups are affected.  Section [Repository encryption](https://borgbackup.readthedocs.io/en/latest/quickstart.html#repository-encryption) warns:

> Make a backup copy of the key file (keyfile mode) or repo config file (repokey mode) and keep it at a safe place, so you still have the key in case it gets corrupted or lost.

(BTW, it's kinda of a recursive joke, isn't it?) It should also say that

> If you loose security database and there are no other clients sharing same repo to synchronize it with, and the repo is not trusted, then, after restoring key file or repo config file you may use the repo in question _only in read-only mode_, i.e. restoring from it is safe, but for new backups create another repo.

Attack scheme here is to provide _older_ copy of the repo (missing everything but keys we can't validate that the repo is up-to-date), thus effectively unrolling latest Nonce allocations and forcing their reuse (and if you ask here, "how the attacker will know I lost my security database?" - there are ways (for instance you could recently perform massive restore from another repo), but the question itself is invalid: you always have to assume that attacker is smarter than you and sees more ways to exploit than you can think of).

fantasya-pbem:
Improve the consistency of multi-client issues with same repository.

fixes #4883

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -32,7 +32,7 @@ Yes, this is *possible* from the technical standpoint, but it is
 built upon a defined :ref:`attack_model` that cannot provide its
 guarantees for multiple clients using the same repository.
 
-In order for the deduplication used by Borg to work, it
+Also, in order for the deduplication used by Borg to work, it
 needs to keep a local cache containing checksums of all file
 chunks already stored in the repository. This cache is stored in
 ``~/.cache/borg/``.  If Borg detects that a repository has been
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -27,7 +27,12 @@ which is slower.
 Can I backup from multiple servers into a single repository?
 ------------------------------------------------------------
 
-Yes, but in order for the deduplication used by Borg to work, it
+Yes, this is *possible* from the technical standpoint, but it is
+*not recommended* from the security perspective. BorgBackup is
+built upon a defined :ref:`attack_model` that cannot provide its
+guarantees for multiple clients using the same repository.
+
+In order for the deduplication used by Borg to work, it
 needs to keep a local cache containing checksums of all file
 chunks already stored in the repository. This cache is stored in
 ``~/.cache/borg/``.  If Borg detects that a repository has been
@@ -85,9 +90,7 @@ Also, you must not run borg against multiple instances of the same repo
 - Encryption security issues if you would update repo and copy-of-repo
   independently, due to AES counter reuse.
 
-There is also a similar encryption security issue for the disaster case:
-If you lose repo and the borg client-side config/cache and you restore
-the repo from an older copy-of-repo, you also run into AES counter reuse.
+See also: :ref:`faq_corrupt_repo`
 
 Which file types, attributes, etc. are *not* preserved?
 -------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
Thanks!

Issue #2295, opened 2017-03-13
One cause of orphaned chunks


enkore:
When a file is processed and a file-critical error occurs, e.g. EIO somewhere in the file, then all chunks that were already add_chunk()d will become orphans.

This seems like an O.K. trade-off to me, since while fixing it in the current codebase is doable ("just" tricky), it'd be a significant complexity increase in any multi-threaded structure I can come up with.

If one wants to fix it, then a possible solution could be to delay the incref when processing files to the point where add_item() is called, and iterating over the chunks cache, deleting chunks with refcount==0.

fantasya-pbem:
I have thought about documenting this, but I'm not sure yet where to place that. FAQ? Is it a "common issue" or rather a "limitation"? Could be like "Borg check finds orphaned objects, why?", but that should explain more than just one cause.

Or we could add a note to ``borg create`` usage (if file read error, then orphan), but that again is just one of several things that could go wrong.

ThomasWaldmann:
Guess it could be in the `borg check` docs.

fantasya-pbem:
At least we have this documented now.

ThomasWaldmann:
Guess documenting it is good enough, closing.

PR #5097, opened 2020-04-10
[DOC] Document one cause of orphaned chunks in check command, #2295

fantasya-pbem:
Clean up the whole check usage paragraph, use consistent wording.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2812,58 +2812,59 @@ class Archiver:
 
         First, the underlying repository data files are checked:
 
-        - For all segments the segment magic (header) is checked
-        - For all objects stored in the segments, all metadata (e.g. crc and size) and
+        - For all segments, the segment magic header is checked.
+        - For all objects stored in the segments, all metadata (e.g. CRC and size) and
           all data is read. The read data is checked by size and CRC. Bit rot and other
           types of accidental damage can be detected this way.
-        - If we are in repair mode and a integrity error is detected for a segment,
-          we try to recover as many objects from the segment as possible.
-        - In repair mode, it makes sure that the index is consistent with the data
-          stored in the segments.
-        - If you use a remote repo server via ssh:, the repo check is executed on the
-          repo server without causing significant network traffic.
+        - In repair mode, if an integrity error is detected in a segment, try to recover
+          as many objects from the segment as possible.
+        - In repair mode, make sure that the index is consistent with the data stored in
+          the segments.
+        - If checking a remote repo via ``ssh:``, the repo check is executed on the server
+          without causing significant network traffic.
         - The repository check can be skipped using the ``--archives-only`` option.
-        - A repository check can be time consuming. Partial checks are possible with the ``--max-duration`` option.
+        - A repository check can be time consuming. Partial checks are possible with the
+          ``--max-duration`` option.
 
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2812,58 +2812,59 @@ def define_borg_mount(parser):
...
-          we try to recover as many objects from the segment as possible.
-        - In repair mode, it makes sure that the index is consistent with the data
-          stored in the segments.
-        - If you use a remote repo server via ssh:, the repo check is executed on the
-          repo server without causing significant network traffic.
+        - In repair mode, if an integrity error is detected in a segment, try to recover
+          as many objects from the segment as possible.
+        - In repair mode, make sure that the index is consistent with the data stored in
+          the segments.
+        - If checking a remote repo via ``ssh:``, the check is executed on the server
```
ThomasWaldmann, line 2823: ... the **repo check** ...

src/borg/archiver.py
```diff
@@ -2812,58 +2812,59 @@ def define_borg_mount(parser):
...
-          required).
-        - The archive checks can be time consuming, they can be skipped using the
+          files, check if chunk is present. In repair mode, if a chunk is not present,
+          replace it with a same-size replacement chunk of zeroes. If a previously lost
+          chunk reappears (e.g. via a later backup), in repair mode the all-zero replacement
+          chunk will be replaced by the correct chunk. This requires reading of archive and
+          file metadata, but not data.
+        - In repair mode, when all the archives were checked, orphaned chunks are deleted
+          from the repo. One cause for orphaned chunks are read errors in the archive
+          creation process.
```
ThomasWaldmann, line 2842: ... are **input file related errors** in ... (or **source file**?)

(not sure if these are just read errors, i guess all sorts of non-recoverable file-scope errors happening somewhere not right at the beginning lead to that)

ThomasWaldmann: is it "one cause for" or "one cause of"?

src/borg/archiver.py
```diff
@@ -2812,58 +2812,59 @@ def define_borg_mount(parser):
...
+        - If checking a remote repo via ``ssh:``, the archive check is executed on the
+          client machine. (Because if encryption is enabled, the checks will require
+          decryption, and this is always done client-side as key access is required.)
```
ThomasWaldmann, line 2845: i am not a native english speaker, but this way of putting the parens looks strange to me.

```
foo (bar).
```
vs.
```
foo. (bar.)
```

fantasya-pbem: Well, it's not just "bar" in parentheses, but a full sentence that ends with a full stop. It's a stylistic thing in my opinion. I will find a better phrase.

src/borg/archiver.py
fantasya-pbem: Yes, "one cause of" is better.

BTW, shall I resolve conversations after fixing things, or do you want to this yourself?

src/borg/archiver.py
fantasya-pbem: fixed

src/borg/archiver.py
ThomasWaldmann: doesn't matter, only that it is fixed or discussed.

ThomasWaldmann:
Thanks!

Issue #4029, opened 2018-08-10
borg extract and absolute pathes


ThomasWaldmann:
when using `borg create`, the leading slash of absolute pathes is automatically removed from items when they are written into an archive.

otoh, when using `borg extract repo::archive /some/path`, borg extract tries to match exactly what was given on the command line. And because the archive has no absolute paths in it, it fails with "nothing matched".

Check whether we should automatically remove leading slashes for `borg extract` (and some other commands that do path matching) also.

ThomasWaldmann:
see also #2763

milkey-mouse:
Crazy idea: `borg info` shows the original command line arguments that created the backup. We could try to re-parse those args at least well enough to get the original path and see if it started with a slash, then use that to inform `borg extract` & friends.

PR #5099, opened 2020-04-10
[FEATURE] Borg commands accept absolute pathes, #4029

fantasya-pbem:
Remove all leading slashes in patterns of all sorts but re: style.

This change will break some tests, but as far as I could determine it has no other side effects. I left out changing the re: patterns because removing the leading slash can invalidate all non-trivial patterns (like /?some/dir) or is complicated (like ^/some/dir).

I checked ``extract`` and ``mount`` to verify that all --exclude pattern styles work and the pathes args are recognized.

What's your opinion? Acceptable, or too esoteric?

### DIFF
--- a/src/borg/patterns.py
+++ b/src/borg/patterns.py
@@ -215,7 +215,7 @@ class PathFullPattern(PatternBase):
     PREFIX = "pf"
 
     def _prepare(self, pattern):
-        self.pattern = os.path.normpath(pattern)
+        self.pattern = os.path.normpath(pattern).lstrip(os.path.sep)  # sep at beginning is removed
 
     def _match(self, path):
         return path == self.pattern
@@ -235,7 +235,9 @@ class PathPrefixPattern(PatternBase):
     PREFIX = "pp"
 
     def _prepare(self, pattern):
-        self.pattern = os.path.normpath(pattern).rstrip(os.path.sep) + os.path.sep
+        sep = os.path.sep
+
+        self.pattern = (os.path.normpath(pattern).rstrip(sep) + sep).lstrip(sep)  # sep at beginning is removed
 
     def _match(self, path):
         return (path + os.path.sep).startswith(self.pattern)
@@ -253,7 +255,7 @@ class FnmatchPattern(PatternBase):
         else:
             pattern = os.path.normpath(pattern) + os.path.sep + '*'
 
-        self.pattern = pattern
+        self.pattern = pattern.lstrip(os.path.sep)  # sep at beginning is removed
 
         # fnmatch and re.match both cache compiled regular expressions.
         # Nevertheless, this is about 10 times faster.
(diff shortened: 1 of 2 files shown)


fantasya-pbem:
In a second step I fixed most tests.

ThomasWaldmann:
there are some formatting issues which are hopefully also found by the flake8 checker:

https://travis-ci.org/github/borgbackup/borg/jobs/673750516

ThomasWaldmann:
Did you think about how this relates to native windows support?

fantasya-pbem:
> Did you think about how this relates to native windows support?

No, and I cannot run tests or try the changes on Windows. I guess Borg also saves only relative paths, just that they have a backslash in them? ``\Programs\MyTool`` becomes ``Programs\MyTool``, and ``C:\some\file`` becomes ``some\file``? So this fix would work when not using drive letters I guess, isn't it?

ThomasWaldmann:
I didn't review your changes yet. If you only deal with pathes as seen in archives, there is never a leading /. But on pathes that somehow come from the user (cli args or other) might have leading / (or \).
See also the other PRs.

ThomasWaldmann:
```
flake8 run-test: commands[0] | flake8 src scripts conftest.py
src/borg/patterns.py:218:69: E703 statement ends with a semicolon
src/borg/patterns.py:218:70: E261 at least two spaces before inline comment
src/borg/patterns.py:258:51: E261 at least two spaces before inline comment
src/borg/patterns.py:282:43: E261 at least two spaces before inline comment
src/borg/patterns.py:295:31: E261 at least two spaces before inline comment
```

src/borg/patterns.py
```diff
@@ -235,7 +235,9 @@ class PathPrefixPattern(PatternBase):
     PREFIX = "pp"
 
     def _prepare(self, pattern):
-        self.pattern = os.path.normpath(pattern).rstrip(os.path.sep) + os.path.sep
+        sep = os.path.sep
+
+        self.pattern = (os.path.normpath(pattern).rstrip(sep) + sep).lstrip(sep)  # sep at beginning is removed
```
ThomasWaldmann, line 240: how about:
- normpath
- lstrip
- rstrip
- add sep

would not require parens like seen above.

fantasya-pbem:
> did you check whether we test all the use cases where the matcher is used, like create, extract, mount, ...?

Yes, I did a search for the PatternMatcher class and found that these changes here should cover all occurrences of paths and patterns. As the matcher classes are not used for other stuff, I think these few changed lines are the minimum solution.

src/borg/patterns.py
ThomasWaldmann: ah, ok, I overlooked that case.

ThomasWaldmann:
Thanks!

PR #5103, opened 2020-04-11
[DOC] Document one cause of orphaned chunks in check command, #2295

fantasya-pbem:
Backport from master: Clean up the whole check usage paragraph.


ThomasWaldmann:
Thanks!

PR #5108, opened 2020-04-12
fuse: set f_namemax in statfs result, fixes #2684

### LINKED ISSUE #2684: `pathconf(..., _PC_NAME_MAX) == 0` in borg mount
When I use `pathconf(..., _PC_NAME_MAX)` to allocate space for `struct dirent` as described in [`man 3 readdir`](https://linux.die.net/man/3/readdir), I get back 0. [`man 3 pathconf`](https://linux.die.net/man/3/pathconf) says it should return -1 on error.

Original bindfs bug: https://github.com/mpartel/bindfs/issues/54 by @mengelmann

ThomasWaldmann:
setting it to 255 for now (as seen on Linux / ext4), better than the default 0.

the attribute is present since llfuse 1.3.0, which is the minimum requirement currently anyway.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -540,6 +540,7 @@ class FuseOperations(llfuse.Operations, FuseBackend):
         stat_.f_files = 0
         stat_.f_ffree = 0
         stat_.f_favail = 0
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Note: the backport needs to check whether the attribute is actually present (to support llfuse < 1.3.0).

PR #5133, opened 2020-04-18
[BUGFIX] Borg commands accept absolute pathes, #4029

fantasya-pbem:
Strip leading path separator before trying to match a path.

I just found that I forgot this. In my tests I only checked if ``borg create ::archive relative/path`` works, but not with ``/absolute/path``.


ThomasWaldmann:
Thanks!

Issue #4982, opened 2020-03-05
1.0-maint testing broken on macOS / pyenv openssl issue


ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/jobs/658694432?utm_medium=notification&utm_source=github_status

ThomasWaldmann:
also here:
https://travis-ci.org/borgbackup/borg/jobs/658694433?utm_medium=notification&utm_source=github_status

ThomasWaldmann:
this worked:
https://travis-ci.org/borgbackup/borg/jobs/658694434?utm_medium=notification&utm_source=github_status

ThomasWaldmann:
minimum python version when using openssl 1.1 maybe?

borg 1.0.x was meant to be used with openssl 1.0.x, not sure if we backported the changes needed for openssl 1.1.x.

ThomasWaldmann:
fixed by #5135.

PR #5135, opened 2020-04-19
Fix MacOS build for 1.0-maint

nijave:
Fixes https://github.com/borgbackup/borg/issues/4982

Homebrew updated default openssl to 1.1 but not all versions of Python support that version (i.e. older versions of Python). It looks like the pyenv code assumes `openssl` formula is 1.0 and `openssl@1.1` is 1.1 when in fact they're now the same. This instructs pyenv to use `openssl@1.0` formula from rbenv tap

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -9,10 +9,8 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     export HOMEBREW_TEMP=~/brew-temp
     mkdir $HOMEBREW_LOGS
     mkdir $HOMEBREW_TEMP
-    # brew update
-    if [[ "${OPENSSL}" != "0.9.8" ]]; then
-        brew outdated openssl || brew upgrade openssl
-    fi
+
+    brew install rbenv/tap/openssl@1.0
 
     if which pyenv > /dev/null; then
         eval "$(pyenv init -)"
@@ -25,11 +23,11 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
 
     case "${TOXENV}" in
         py34)
-            pyenv install 3.4.5
+            PYTHON_BUILD_HOMEBREW_OPENSSL_FORMULA=openssl@1.0 pyenv install 3.4.5
             pyenv global 3.4.5
             ;;
         py35)
-            pyenv install 3.5.2
+            PYTHON_BUILD_HOMEBREW_OPENSSL_FORMULA=openssl@1.0 pyenv install 3.5.2
             pyenv global 3.5.2
             ;;
         py36)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@nijave thanks for the PR.

Did you check whether the openssl 1.1 changes inside borg were backported to 1.0 as I mentioned there?:

https://github.com/borgbackup/borg/issues/4982#issuecomment-595256258

nijave:
> @nijave thanks for the PR.
> 
> Did you check whether the openssl 1.1 changes inside borg were backported to 1.0 as I mentioned there?:
> 
> [#4982 (comment)](https://github.com/borgbackup/borg/issues/4982#issuecomment-595256258)

There's this? https://github.com/borgbackup/borg/pull/1193

Not sure if there's anything else required--looks like it passed the tests. I've since found out py3.4 outright only supports 1.0 so I may try the first commit (rbenv's forked version of the openssl 1.0 brew formula)

I don't really care either way--just looked like an easy first issue to pass the time

ThomasWaldmann:
Thanks for digging into that. Yeah, looks like it was backported.

The python 3.4.10 on macOS didn't work on travis, though.

ThomasWaldmann:
Guess there are 2 ways out of this:
- drop py34 testing on macOS (py34 does not support openssl 1.1.x)
- install openssl 1.0.x so py34 will work

ThomasWaldmann:
3.5.3, 3.6.2, 3.7.0 are the version working with openssl 1.1 (from Vagrantfile).

nijave:
> Guess there are 2 ways out of this:
> 
>     * drop py34 testing on macOS (py34 does not support openssl 1.1.x)
> 
>     * install openssl 1.0.x so py34 will work

I started a build with an attempt at 1.0. py 3.4 is EOL https://devguide.python.org/devcycle/#end-of-life-branches

.travis/install.sh
```diff
@@ -9,10 +9,8 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     export HOMEBREW_TEMP=~/brew-temp
     mkdir $HOMEBREW_LOGS
     mkdir $HOMEBREW_TEMP
-    # brew update
-    if [[ "${OPENSSL}" != "0.9.8" ]]; then
-        brew outdated openssl || brew upgrade openssl
-    fi
+
+    brew install rbenv/tap/openssl@1.0
```
ThomasWaldmann, line 13: hmm, can you add a comment why this is like it is (compared to using something like `openssl@1.0`?

nijave: Added a note. More details in the tap docs https://docs.brew.sh/Taps#the-brew-tap-command

tldr; rbenv/tap becomes github.com/rbenv/homebrew-tap and openssl Formula is inside that repo

ThomasWaldmann: thanks for the note, will merge after tests.

ThomasWaldmann:
Thanks a lot for making this work!

Issue #4160, opened 2018-11-08
recreate deletes all files using --patterns-from parameter


kcciti:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug

## System information. For client/server mode post info for both machines.

Dell Laptop

#### Your borg version (borg -V).

borg 1.1.7

#### Operating system (distribution) and version.

Linux Solus v3.99999

#### Hardware / network configuration, and filesystems used.

ext4

#### How much data is handled by borg?

2.3M (test case)

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
borg create --stat --verbose --list .::full ~/R_script
borg create --stat --verbose --list --patterns-from test.lst .::patterns
borg recreate --stat --verbose --list --patterns-from test.lst --target full.recreated .::full

CONTENTS: R_Script
R_Script/test.R
R_Script/.Rproj.user/7038483/pcs/files-pane.pper (and many other files under .Rproj.user)
R_Script/Jags-Ydich-XnomSsubj-MbernBeta-Diagtheta[1].eps
R_Script/Jags-Ydich-XnomSsubj-MbernBeta-Diagtheta[2].eps
R_Script/Jags-Ydich-XnomSsubj-MbernBeta-Diagtheta[3].eps
R_Script/Jags-Ydich-XnomSsubj-MbernBeta-Diagtheta[4].eps
R_Script/R_Script/Jags-Ydich-XnomSsubj-MbernBeta-Post.eps

CONTENT:S test.lst
```
P re
R /home/user/R_script
+ .*Jags.*(1).*
- .*\.eps.*
- .*\.Rproj.*
```

## Describe the problem you're observing.

[...]

PR #5138, opened 2020-04-19
[DOCS] Add note about patterns and stored paths, #4160

fantasya-pbem:
Add a section that explains storage of relative paths and how patterns can be used arbitrarily since Borg 1.2.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2066,17 +2066,27 @@ class Archiver:
         when invoking borg and these can be either relative or absolute paths.
 
         So, when you give `relative/` as root, the paths going into the matcher
-        will look like `relative/.../file.ext`. When you give `/absolute/` as root,
-        they will look like `/absolute/.../file.ext`. This is meant when we talk
-        about "full path" below.
+        will look like `relative/.../file.ext`. When you give `/absolute/` as
+        root, they will look like `/absolute/.../file.ext`.
+
+        File paths in Borg archives are always stored normalized and relative.
+        This means that e.g. ``borg create /path/to/repo ../some/path`` will
+        store all files as `some/path/.../file.ext` and ``borg create
+        /path/to/repo /home/user`` will store all files as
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2164,7 +2176,7 @@ def do_break_lock(self, args, repository):
             $ borg create -e '/home/*/junk' backup /
 
             # Exclude the contents of '/home/user/cache' but not the directory itself:
-            $ borg create -e /home/user/cache/ backup /
+            $ borg create -e home/user/cache/ backup /
```
ThomasWaldmann, line 2179: hmm, this (`/home/...`) never matched before your changes, right?

src/borg/archiver.py
```diff
@@ -2179,8 +2191,8 @@ def do_break_lock(self, args, repository):
             /home/*/junk
             *.tmp
             fm:aa:something/*
-            re:^/home/[^/]\\.tmp/
-            sh:/home/*/.thumbnails
+            re:^home/[^/]\\.tmp/
+            sh:home/*/.thumbnails
```
ThomasWaldmann, line 2195: same here!?

src/borg/archiver.py
fantasya-pbem: Before, in ``create`` it matched only as `/home`.

fantasya-pbem: And: Before, in ``extract`` it matched only as `home`.

src/borg/archiver.py
fantasya-pbem: I removed the / to show that it can be used or be left out.

src/borg/archiver.py
ThomasWaldmann: uh, right.

src/borg/archiver.py
ThomasWaldmann: ok

Issue #1726, opened 2016-10-15
Can't mount on ubuntu. Missing dependencies ?


jsalatiel:
Trying to mount a backup archive on ubuntu 16.04 fails with:
`# borg mount ~/.backup/ /tmp/pqp/`
`Loading fuse support failed [ImportError: No module named 'llfuse']`

I have installed borgbackup from ubuntu repositories. Is there any missing dependencies on ubuntu packages ?

> 
> `# dpkg -l | grep -i fuse`
> ii  exfat-fuse                                  1.2.3-1                                    amd64        read and write exFAT driver for FUSE
> ii  fuse                                        2.9.4-1ubuntu3.1                           amd64        Filesystem in Userspace
> ii  gvfs-fuse                                   1.28.2-1ubuntu1~16.04.1                    amd64        userspace virtual filesystem - fuse server
> ii  ifuse                                       1.1.2-0.1build3                            amd64        FUSE module for iPhone and iPod Touch devices
> ii  libfuse2:amd64                              2.9.4-1ubuntu3.1                           amd64        Filesystem in Userspace (library)
> ii  ntfs-3g                                     1:2015.3.14AR.1-1build1                    amd64        read/write NTFS driver for FUSE

ThomasWaldmann:
borg mount needs llfuse.

http://packages.ubuntu.com/xenial-updates/borgbackup

it recommends python3-llfuse, but does not require it.

apt-get install python3-llfuse

jsalatiel:
Thanks ! It worked ! Before closing this, shouldn't this package be a dependency in ubuntu repositories ? I dont know who mantains the package, but it should be warned.

jsalatiel:
IMHO, if one of borg features depends on this package, it should be a dependency, not a suggestion =)

ThomasWaldmann:
Well, you can use borg without using borg mount, so it is somehow optional.
As long as llfuse is available, one of course wants to have it. But llfuse is a bit of a fragile thing concerning platform support.

ThomasWaldmann:
@LocutusOfBorg ^^^

LocutusOfBorg:
eh, I'm stick to what upstream suggests
```
# msgpack pure python data corruption was fixed in 0.4.6.
# Also, we might use some rather recent API features.
install_requires = ['msgpack-python>=0.4.6', ]

extras_require = {
    # llfuse 0.40 (tested, proven, ok), needs FUSE version >= 2.8.0
[...]

PR #5139, opened 2020-04-19
borg mount dependency: add python3-llfuse

Wikinaut:
Source for this change: https://github.com/borgbackup/borg/issues/1726#issuecomment-254014151

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -186,7 +186,7 @@ Install the dependencies with development headers::
     liblz4-dev libzstd-dev \
     build-essential \
     pkg-config python3-pkgconfig
-    sudo apt-get install libfuse-dev fuse    # optional, for FUSE support
+    sudo apt-get install libfuse-dev fuse python3-llfuse # optional, for FUSE support
 
 In case you get complaints about permission denied on ``/etc/fuse.conf``: on
 Ubuntu this means your user is not in the ``fuse`` group. Add yourself to that


docs/installation.rst
```diff
@@ -186,7 +186,7 @@ Install the dependencies with development headers::
     liblz4-dev libzstd-dev \
     build-essential \
     pkg-config python3-pkgconfig
-    sudo apt-get install libfuse-dev fuse    # optional, for FUSE support
+    sudo apt-get install libfuse-dev fuse python3-llfuse # optional, for FUSE support
```
ThomasWaldmann, line 189: I don't think this is a valid change. But you have likely found some other issue.

In the docs we already have:
```
some other Python dependencies, pip will automatically install them for you.
```
That includes llfuse and this is the reason why your change is not valid.

But, we also have that:
```
optionally, the llfuse_ Python package is required if you wish to mount an archive as a FUSE filesystem. See setup.py about the version requirements.
```
This is a bit superfluous (as that is already included in what I mentioned above) and also maybe misleading.

Wikinaut: > I don't think this is a valid change. But you have likely found some other issue.
> 
> In the docs we already have:
> 
> ```
> some other Python dependencies, pip will automatically install them for you.
> ```
> 
> That includes llfuse and this is the reason why your change is not valid.
> 
> But, we also have that:
> 
> ```
> optionally, the llfuse_ Python package is required if you wish to mount an archive as a FUSE filesystem. See setup.py about the version requirements.
> ```
> 
> This is a bit superfluous (as that is already included in what I mentioned above) and also maybe misleading.

Thomas, in my systems, that dependency was _not_ installed - I followed your documentation. You are working with borg since years - I try to install on fresh machines. (here: Linux Mint 19.3, many instances, was lacking that dependency.

Arguendo you were right with your Behauptungen, the llfuse was missing nevertheless.  I followed strictly the installation rules in the borg documentation.

ThomasWaldmann: from stable docs:
```
# install Borg + Python dependencies into virtualenv
pip install borgbackup
# or alternatively (if you want FUSE support):
pip install borgbackup[fuse]                          <------- that installs llfuse
```

ThomasWaldmann: and the reason why this is not done by default is that fuse support is not present / unproblematic on all platforms.

Wikinaut:
I don't believe, that my reasoned change - the addition of ```python3-llfuse``` - harms. If you think, it is superfluous, then the whole existing line
```sudo apt-get install libfuse-dev fuse # optional, for FUSE support```
is superfluous in the same way.

ThomasWaldmann:
It does not harm, but it is not needed because pip will install it, IF you request it (see above).
It is a python package and it is handled by setup.py, so it is reasonable to do it that way (you also might get a more recent version than provided by your linux distribution).

OTOH, libfuse-dev and fuse are NOT python packages, but development headers and binaries that need to be provided by the OS. Of course you do not need them if you do not want fuse support. For operating systems providing good working fuse support, it makes sense to install them. If the OS does not have that, it makes sense to NOT install them.

Wikinaut:
Thomas: my use case was that one, perhaps you understand then:

I used borg without FUSE support for a long time.

Now I wanted to "borg mount" and got
grafik

I consulted the documentation and found the line
```sudo apt-get install libfuse-dev fuse # optional, for FUSE support```
**and** your previous post https://github.com/borgbackup/borg/issues/1726#issuecomment-254014151 .


Combing the two information pieces and without your developer background of knowing all the details, I found it reasonable to apply your pip command https://github.com/borgbackup/borg/issues/1726#issuecomment-254014151 . 

Now I wanted to improve your documentation from my point of view **as a dumb user**. Perhaps you can make a Perspektivwechsel and will understand why changing the line makes (may make) sense for future users.

ThomasWaldmann:
In your situation, you can solve it like that.

Nevertheless, that solution is not consistent with the docs (you'ld end up with 2 llfuse installs then, 1 on OS-leve and one in virtualenv).

PR #5155, opened 2020-04-22
fix "--recompress option is counter intuitive", fixes #5154

### LINKED ISSUE #5154: --recompress option is counter intuitive
Have you checked borgbackup docs, FAQ, and open Github issues?: **Yes**
Is this a BUG / ISSUE report or a QUESTION?: **Issue/Question**
Borg version: **1.1.11**
(Other information omitted since not relevant to issue)

I appreciate that `recreate` is an experimental feature and is likely to undergo heavy development until full release (I'm looking forward to the proposed changes in #3631 in particular!). However, the following issue is a UI issue which is perhaps best fixed whist `recreate` is still marked "experimental" and backwards compatibility is not as important.

Currently the default `MODE` when `--recompress`ing is `never`. This is is so counter intuitive that _even the documentation_ gets it wrong:

In the documentation (`man borg-recreate`) under `Examples` it currently states that
```
Create a backup with little but fast compression
$ borg create /mnt/backup::archive /some/files --compression lz4
# Then compress it - this might take longer, but the backup has already completed,
# so no inconsistencies from a long-running backup job.
$ borg recreate /mnt/backup::archive --recompress --compression zlib,9
```
however, according to discussions in #3617, and the merge #3676 (which reflects current documentation), I believe the second borg command would *not* recompress the data (since the default `MODE` is `never`)? I think the example *should* be
```
$ borg recreate /mnt/backup::archive --recompress if-different --compression zlib,9
```
If I have understood correctly, there is no difference between `if-different` and `always` in this context because the previous compression was `lz4` and now it is `zlib`?

To mitigate this, I would suggest that borg should throws an error if no `MODE` is specified.

In fact, it might be worth going one step further, and removing the `never` `MODE` entirely. I'm not sure in which context
```
$ borg recreate $REPO::$ARCHIVE --recompress never [...]
```
makes sense? What is the difference between this, and
```
$ borg recreate $REPO::$ARCHIVE [...]
```
? Perhaps it to underline the fact that if an archive is `recreate`d without `--compression`, then the previous compression algorithms and levels are used? If so, I think this should be documented more clearly.

However, I appreciate that, for backwards compatibility (even as an experimental feauture), this may not be feasible. As a compromise, I would create a warning in the style of the experimental feature warning:
```
WARNING: Currently borg will not perform any recompression even though '--compression' was passed since no MODE was given.
```
The rationale for the proposed changes is that --- before changes in #3631 are implemented -- it could be a very time-wasting endeavor to `recreate --compress --compression L[,C]` only to find out ~24 hrs later that no compression took place.

Further, I would stress the importance of the `MODE` in the first line of the description. That is, change
```
recompress data chunks according to --compression
```
to
```
recompress data chunks according to MODE and --compression
```

Thank you for your hard work on this project.

(Edited for clarification and wording)

velleto:
These changes would implement some of the changes I suggested in #5154. Namely that

1. The documentation stresses the importance of the `MODE`
2. The documentation example is corrected
3. A warning is thrown when the `--recompress` option is used without specifying a `MODE`.

However, since `args.recompress` is set to `never` by default, we cannot check later if it is explicitly set by the user. In this case

    $ borg recreate --recompress --compression L,C $REPO
    $ borg recreate --recompress never --compression L,C $REPO

both present a warning. Perhaps this could be solved by changing the argument parser

https://github.com/borgbackup/borg/blob/914a52ac156f5970529afa02b32db0d3577c082e/src/borg/archiver.py#L4105

to make the default `None` and then later replacing

https://github.com/borgbackup/borg/blob/914a52ac156f5970529afa02b32db0d3577c082e/src/borg/archiver.py#L1578

with

    recompress = args.recompress != 'never' and args.recompress != None

However, I'm not sure what impact this would have on the remainder of the code.


src/borg/archiver.py
```diff
@@ -4104,15 +4109,16 @@ def define_borg_mount(parser):
...
+                                        'do not recompress.'
                                         'MODE `if-different`: '
                                         'recompress if current compression is with a different compression algorithm '
                                         '(the level is not considered). '
                                         'MODE `always`: '
                                         'recompress even if current compression is with the same compression algorithm '
                                         '(use this to change the compression level). '
-                                        'MODE `never` (default): '
-                                        'do not recompress.')
+                                        )
```
ThomasWaldmann, line 4121: src/borg/archiver.py:4121:41: E124 closing bracket does not match visual indentation

src/borg/archiver.py
```diff
@@ -1566,8 +1566,13 @@ def do_upgrade(self, args, repository, manifest=None, key=None):
     @with_repository(cache=True, exclusive=True, compatibility=(Manifest.Operation.CHECK,))
     def do_recreate(self, args, repository, manifest, key, cache):
         """Re-create archives"""
-        msg = ("recreate is an experimental feature.\n"
-               "Type 'YES' if you understand this and want to continue: ")
+        if args.recompress == 'never':
+            msg = ("recreate is an experimental feature.\n"
+                   "WARNING: Currently borg will not perform any recompression.\n"
```
ThomasWaldmann, line 1571: AFAICS, it would output this, if `--recompress` was not given on the commandline. So somehow telling the obvious.

It also would output this, if `--recompress never` is given, same thing.

src/borg/archiver.py
```diff
@@ -4104,15 +4109,16 @@ def define_borg_mount(parser):
...
+                                        '(default) MODE `never`: '
+                                        'do not recompress.'
                                         'MODE `if-different`: '
                                         'recompress if current compression is with a different compression algorithm '
                                         '(the level is not considered). '
                                         'MODE `always`: '
                                         'recompress even if current compression is with the same compression algorithm '
                                         '(use this to change the compression level). '
-                                        'MODE `never` (default): '
-                                        'do not recompress.')
```
ThomasWaldmann, line 4115: this could be clarified by just giving the 4(!) possible cases and their outcome.

docs/usage/recreate.rst
```diff
@@ -13,7 +13,7 @@ Examples
     $ borg create /mnt/backup::archive /some/files --compression lz4
     # Then compress it - this might take longer, but the backup has already completed,
     # so no inconsistencies from a long-running backup job.
-    $ borg recreate /mnt/backup::archive --recompress --compression zlib,9
+    $ borg recreate /mnt/backup::archive --recompress if-different --compression zlib,9
```
ThomasWaldmann, line 16: this does not change the behaviour as this is the same as the `const` value, see argparse docs.

src/borg/archiver.py
velleto: You are right. That seems like the clearest solution.

velleto:
I've created about 5 commits for something that should be about 2 at most. I'm not sure what the best course of action is for this, since I have already pushed them. I am happy to reset, rebase, `push --force` and create a new pull request.

src/borg/archiver.py
```diff
@@ -4104,15 +4104,16 @@ def define_borg_mount(parser):
...
+                                   help='recompress data chunks according to `MODE` and ``--compression``. '
+                                        'Possible modes are '
+                                        '`if-different`: recompress if current compression is with a different '
+                                        'compression algorithm (the level is not considered); '
+                                        '`always`: recompress even if current compression is with the same '
+                                        'compression algorithm (use this to change the compression level); and '
+                                        '`never`: do not recompress (use this option to explicitly prevent '
+                                        'recompression). '
+                                        'If no MODE is given, `if-different` will be used. '
+                                        'Not passing --recompress is equivalent "--recompress never".')
```
ThomasWaldmann, line 4116: ... is equivalent **to** ...

Yeah, much better now. That should not leave any unclarities.

ThomasWaldmann:
Note: I can squash this into 1 commit when merging the PR.

src/borg/archiver.py
velleto: Well spotted. Fixed

ThomasWaldmann:
Thanks!

PR #5160, opened 2020-04-25
[BUGFIX] Borg config --list does not show last_segment_checked, #5159

### LINKED ISSUE #5159: Borg config --list does not show last_segment_checked

Yes

BUG

1.2.0a8 / also master

`borg config --list /path/to/repo`

If a partial check has started and has been stopped due to timeout, the config file contains the `last_segment_checked` key with an int value.
You can verify that with `borg config /path/to/repo last_segment_checked`, which prints the segment number.

Now try `borg config --list /path/to/repo`: The last_segment_checked is missing.

Reason: In source code of archiver.py `last_segment_checked` is not read when the list subcommand is executed.

I will create a PR for this.

fantasya-pbem:
Add a for loop for "dynamic" config keys like last_segment_checked, skip printing if they are not found in the config.

Also allow last_segment_checked to be set with a int value.

Added tests.

fixes #5159

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1659,7 +1659,7 @@ class Archiver:
         def repo_validate(section, name, value=None, check_value=True):
             if section not in ['repository', ]:
                 raise ValueError('Invalid section')
-            if name in ['segments_per_dir', 'max_segment_size', 'storage_quota', ]:
+            if name in ['segments_per_dir', 'max_segment_size', 'storage_quota', 'last_segment_checked', ]:
                 if check_value:
                     try:
                         int(value)
@@ -1715,6 +1715,11 @@ class Archiver:
                     value = default_values.get(key)
                     if value is None:
                         raise Error('The repository config is missing the %s key which has no default value' % key)
+                print('%s = %s' % (key, value))
+            for key in ['last_segment_checked', ]:
+                value = config.get('repository', key, fallback=None)
+                if value is None:
+                    continue
                 print('%s = %s' % (key, value))
 
         if not args.list:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Well, I was not sure whether this should be changed, considering last_segment_checked is a rather internal thing that is maybe not interesting to read and maybe even bad if set to a wrong value.

OTOH, we can expect admins to be able to deal with it and they are allowed to shoot themselves in the feet. :-)

src/borg/archiver.py
```diff
@@ -1716,6 +1716,11 @@ def list_config(config):
                     if value is None:
                         raise Error('The repository config is missing the %s key which has no default value' % key)
                 print('%s = %s' % (key, value))
+            for key in ['last_segment_checked']:
```
ThomasWaldmann, line 1719: can you add a comma and a blank after the list entry? see 1662.

ThomasWaldmann: hmm, the default value of last_segment_checked is (IIRC) 0 (or 1?).

if it is not there, it starts at the beginning, with the first segment.

ThomasWaldmann: so the question is, whether this value could be treated in same way as the other values and not need special-casing with own code at all.

fantasya-pbem: I'd prefer to treat it specially. It should not be printed if it is not there. If we'd return last_segment_checked = 0 that means that a partial check is running, which is wrong.

ThomasWaldmann: ok

ThomasWaldmann:
Thanks!

Issue #5170, opened 2020-05-02
Borg check -- ask for passphrase immediately


kdmurray91:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes, apologies if I missed something though

## Is this a BUG / ISSUE report or a QUESTION?

Question/enhancement/feature request

#### Your borg version (borg -V).

`borg 1.1.9` from the debian sid package

#### Operating system (distribution) and version.

Debian sid

#### Hardware / network configuration, and filesystems used.

xfs on a raid 6 array

#### How much data is handled by borg?

~1.5TB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
export BORG_REPO='/local/path/to/images.repo'
export BORG_PASSCOMMAND='pass borg/imagerepo'
borg check --prefix images --last 3 --progress
```

## Describe the problem you're observing.

`borg check` first does a repo scan which doesn't require a passphrase, and later asks for a passphrase when it starts checking the archives.

This causes some inconvenience, as i'll often run a backup and then leave `borg check` to run overnight. As this repo is quite large, the repo check will take a couple of hours, which means the archive check hangs until i can put the password in the next morning. If `borg check` asked for the passphrase immediately (via borg_passcommand, which in this case is [password-store](https://www.passwordstore.org/)), the scan over archives could complete overnight too. 

Taking a quick look at the code, perhaps the [`ArchiveChecker` instance](https://github.com/borgbackup/borg/blob/9019b48fd5982802af44a510b9decc05fa672c33/src/borg/archiver.py#L335) could be instantiated before the [`repository.check()` call](https://github.com/borgbackup/borg/blob/9019b48fd5982802af44a510b9decc05fa672c33/src/borg/archiver.py#L331) (only if `not args.repo_only`), and `ArchiveChecker`'s `__init__()` be changed to unlock the repo on instantitation? I assume `self.key = self.identify_key(repository)` is what gets the key? It seems that would require a bit of a rejig of `ArchiveChecker` to split initialisation code out from the `ArchiveChecker.check()` method. Are there places where an `ArchiveChecker` instance is used to check more than one repo, and therefore can't be one instance for one repo? If this logic seems sensible, i can code up a PR later.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

N/A, see above comandline.

#### Include any warning/errors/backtraces from the system logs

All commands successful, and logs are normal.

[...]

PR #5171, opened 2020-05-03
Implement an early-decrypting check

kdm9:
Hello,

(NB: this is definitely a draft)

This is my quick attempt at implementing the early-decrypting behaviour asked for in #5170. I'm not sure if it behaves exactly as I wish, as i haven't trusted it to run on a real repo until it passes a review in case i've screwed something up. The tests that relate to `borg check` or archive checking seem to pass fine, but there are a couple of tests failing for reasons I can't easily fathom (hoping these failures go away in CI).

Cheers,
Kevin

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1386,13 +1386,9 @@ class RobustUnpacker:
 
 class ArchiveChecker:
 
-    def __init__(self):
-        self.error_found = False
-        self.possibly_superseded = set()
-
-    def check(self, repository, repair=False, archive=None, first=0, last=0, sort_by='', glob=None,
+    def __init__(self, repository, repair=False, archive=None, first=0, last=0, sort_by='', glob=None,
               verify_data=False, save_space=False):
-        """Perform a set of checks on 'repository'
+        """A checker of repostiory archives
 
         :param repair: enable repair mode, write updated or corrected data into repository
         :param archive: only check this archive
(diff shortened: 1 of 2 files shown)


src/borg/archive.py
```diff
@@ -1386,13 +1386,9 @@ def __next__(self):
...
 
-    def __init__(self):
-        self.error_found = False
-        self.possibly_superseded = set()
-
-    def check(self, repository, repair=False, archive=None, first=0, last=0, sort_by='', glob=None,
+    def __init__(self, repository, repair=False, archive=None, first=0, last=0, sort_by='', glob=None,
               verify_data=False, save_space=False):
-        """Perform a set of checks on 'repository'
+        """A checker of repostiory archives
```
ThomasWaldmann, line 1391: typo

src/borg/archive.py
```diff
@@ -1459,7 +1468,7 @@ def identify_key(self, repository):
         cdata = repository.get(some_chunkid)
         return key_factory(repository, cdata)
 
-    def verify_data(self):
+    def do_verify_data(self):
```
ThomasWaldmann, line 1471: why did you rename this?

src/borg/archiver.py
```diff
@@ -327,15 +327,21 @@ def do_check(self, args, repository):
...
+        if not args.repo_only:
+            try:
+                checker = ArchiveChecker(repository, repair=args.repair, archive=args.location.archive, first=args.first,
+                                         last=args.last, sort_by=args.sort_by or 'ts', glob=args.glob_archives,
+                                         verify_data=args.verify_data, save_space=args.save_space)
+            except Exception as exc:
+                self.print_error(str(exc))
+                return EXIT_WARNING
         if not args.archives_only:
             if not repository.check(repair=args.repair, save_space=args.save_space, max_duration=args.max_duration):
```
ThomasWaldmann, line 342: the potential fundamental problem with first accessing the repo inside `ArchiveChecker.__init__` before doing `repository.check` is that we do not know whether we are working on valid data.

The repo check / repair makes sure that the fundamental data is ok (in memory and in repair mode also on disk).
The archives check / repair builds upon that (and assumes that lower level structures are ok).

So, guess there needs to be pretty much review before touching repo data before the repo has been checked.

src/borg/archive.py
```diff
@@ -1401,32 +1397,45 @@ def check(self, repository, repair=False, archive=None, first=0, last=0, sort_by
...
         self.repository = repository
+        self.archive = archive
+        self.first = first
+        self.last = last
+        self.glob = glob
+        self.check_all = archive is None and not any((first, last, glob))
+        self.sort_by = sort_by
+        self.save_space = save_space
         self.init_chunks()
+        self.key = self.identify_key(repository)
```
ThomasWaldmann, line 1413: guess these 2 need checking.

do these have any unwanted consequences when run before the repo check has happened?

src/borg/archive.py
kdmurray91: as we now have a boolean flag we need to save to `self`:

```
def __init__(self, ..., verify_data=False):
    ...
    self.verify_data = verify_data
```

src/borg/archiver.py
kdmurray91: OK, and that's what I feared. Is there some way to verify only the chunks that contain the metadata needed to decrypt in `ArchiveChecker.__init__()`, so we know that the crypto & manifest & whatever other metadata is kosher, but don't spend 5 hours doing all of `repo/data/*`

ThomasWaldmann: yes, a "miniature version" of the borg repo check, that only makes sure that 1 chunk is ok (which could be the manifest (id 0) or any other chunk). And then detect crypto based on that chunk.

that might work, needs checking...

ThomasWaldmann:
Hmm, what are we doing with this PR?

kdmurray91:
> Hmm, what are we doing with this PR?

@ThomasWaldmann I don't think I have the confidence to verify that my changes here have no terrible side effects, so this PR has somewhat stalled. As far as I can tell it's OK, but I don't have anywhere near the knowledge of Borg's internals I'd need to be confident about that. I guess I'd need someone with good knowledge of Borg's innards to check my change and see that I'm not doing anything stupid.

ThomasWaldmann:
The borg2 code in master branche changed significantly meanwhile.

I recently changed the check code and fixed #5170.

PR #5174, opened 2020-05-04
Allow creating archives using stdout of given command

pgerber:
In addition to allowing:

```
some-command --param value | borg create REPO::ARCH -
```

also allow:

```
borg create --content-from-command create REPO::ARCH -- some-command --param value
```

The difference is that the latter approach deals with errors properly.
In the former example, an archive is created no matter what. Even, if
`some-command` aborts and the output is truncated, Borg won't realize.
In the latter example, the status code is checked and archive creation
is aborted properly when appropriate.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3030,7 +3030,7 @@ class Archiver:
         When giving '-' as path, borg will read data from standard input and create a
         file 'stdin' in the created archive from that data. In some cases it's more
         appropriate to use --content-from-command, however. See section *Read from
-        Stdin* below for details.
+        stdin* below for details.
 
         The archive will consume almost no disk space for files or parts of files that
         have already been stored in other archives.
@@ -3151,7 +3151,7 @@ class Archiver:
         - 'x' = excluded, item was *not* backed up
         - '?' = missing status code (if you see this, please file a bug report!)
 
-        Reading from Stdin
+        Reading from stdin
         ++++++++++++++++++
 
         There are two methods to read from stdin. Either specify ``-`` as path and
@@ -3174,7 +3174,7 @@ class Archiver:
         **one can end up with truncated output being backed up**. Using
         ``--content-from-command``, in contrast, borg is guaranteed to fail without
         creating an archive should the command fail. The command is considered failed
-        when it returned an non-zero exit code.
+        when it returned a non-zero exit code.
 
         By default, the content read from stdin is stored in a file called 'stdin'.
         Use ``--stdin-name`` to change the name.
@@ -3205,7 +3205,7 @@ class Archiver:
                                help='use NAME in archive for stdin data (default: %(default)r)')
         subparser.add_argument('--content-from-command', action='store_true',
                                help='interpret PATH as command and store its stdout. See also section Read from'
-                                    ' Stdin below.')
+                                    ' stdin below.')
 
         exclude_group = define_exclusion_group(subparser, tag_files=True)
         exclude_group.add_argument('--exclude-nodump', dest='exclude_nodump', action='store_true',
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3173,8 +3173,8 @@ class Archiver:
         archive even if the command piping to borg exits with a failure. In this case,
         **one can end up with truncated output being backed up**. Using
         ``--content-from-command``, in contrast, borg is guaranteed to fail without
-        creating an archive should the command fail. A command is considered to have
-        failed when it returns an non-zero exit code.
+        creating an archive should the command fail. The command is considered failed
+        when it returned an non-zero exit code.
 
         By default, the content read from stdin is stored in a file called 'stdin'.
         Use ``--stdin-name`` to change the name.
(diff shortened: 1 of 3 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -1044,6 +1044,25 @@ def test_create_stdin(self):
         extracted_data = self.cmd('extract', '--stdout', self.repository_location + '::test', binary_output=True)
         assert extracted_data == input_data
 
+    def test_create_content_from_command(self):
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        input_data = 'some test content'
+        name = 'a/b/c'
+        self.cmd('create', '--stdin-name', name, '--content-from-command',
+                 self.repository_location + '::test', 'echo', input_data)
```
ThomasWaldmann, line 1052: no '--' here separating borg and echo command?

src/borg/archiver.py
```diff
@@ -507,39 +507,60 @@ def create_inner(archive, cache, fso):
...
-                        status = '-'
-                    self.print_file_status(status, path)
-                    continue
-                path = os.path.normpath(path)
-                parent_dir = os.path.dirname(path) or '.'
-                name = os.path.basename(path)
-                # note: for path == '/':  name == '' and parent_dir == '/'.
-                # the empty name will trigger a fall-back to path-based processing in os_stat and os_open.
-                with OsOpen(path=parent_dir, flags=flags_root, noatime=True, op='open_root') as parent_fd:
+            if args.content_from_command:
```
ThomasWaldmann, line 510: the overall diff is not very readable, it looks like more change than it really is (i guess).

could you change the if-condition to `if not args.content_from_command` and swap the if/else blocks to see if the github diff display works better then?

src/borg/testsuite/archiver.py
ThomasWaldmann: even if it is not strictly needed, i guess we better test if it also works with `--`.

src/borg/archiver.py
pgerber: Yes, it looks like far more than it is. However, swapping the branches isn't much of an improvement [(swapped version)](https://github.com/borgbackup/borg/compare/9019b48fd5982802af44a510b9decc05fa672c33...pgerber:swapped#diff-48f2b9140a35ffd45a46ccfbe63931a6L509).

Kind of sad that after all this years GitHub's diff view has improved so little.

src/borg/testsuite/archiver.py
pgerber: agreed

src/borg/archiver.py
```diff
@@ -3151,7 +3202,10 @@ def define_borg_mount(parser):
         subparser.add_argument('--no-cache-sync', dest='no_cache_sync', action='store_true',
                                help='experimental: do not synchronize the cache. Implies not using the files cache.')
         subparser.add_argument('--stdin-name', metavar='NAME', dest='stdin_name', default='stdin',
-                               help='use NAME in archive for stdin data (default: "stdin")')
+                               help='use NAME in archive for stdin data (default: %(default)r)')
+        subparser.add_argument('--content-from-command', action='store_true',
+                               help='interpret PATH as command and store its stdout. See also section Read from'
```
ThomasWaldmann, line 3207: **Reading from stdin**

src/borg/archiver.py
```diff
@@ -3127,6 +3150,34 @@ def define_borg_mount(parser):
         - '-' = dry run, item was *not* backed up
         - 'x' = excluded, item was *not* backed up
         - '?' = missing status code (if you see this, please file a bug report!)
+
+        Reading from Stdin
```
ThomasWaldmann, line 3154: As stdin is a fixed symbol name, maybe do not capitalize it?

src/borg/archiver.py
ThomasWaldmann: ok, I try to review "as is".

src/borg/archiver.py
```diff
@@ -3007,7 +3028,9 @@ def define_borg_mount(parser):
         directory.
 
         When giving '-' as path, borg will read data from standard input and create a
-        file 'stdin' in the created archive from that data.
+        file 'stdin' in the created archive from that data. In some cases it's more
+        appropriate to use --content-from-command, however. See section *Read from
+        Stdin* below for details.
```
ThomasWaldmann, line 3033: **Reading from stdin**

src/borg/archiver.py
```diff
@@ -3127,6 +3150,34 @@ def define_borg_mount(parser):
...
+
+        ``--`` is used to ensure ``--id`` and ``--stdout`` are **not** considered
+        arguments to ``borg`` but rather ``backup-vm``.
+
+        The difference between the two approaches is that piping to borg creates an
+        archive even if the command piping to borg exits with a failure. In this case,
+        **one can end up with truncated output being backed up**. Using
+        ``--content-from-command``, in contrast, borg is guaranteed to fail without
+        creating an archive should the command fail. The command is considered failed
+        when it returned an non-zero exit code.
```
ThomasWaldmann, line 3177: ... **a** non-zero ...

PR #5186, opened 2020-05-12
WIP: Support copying objects between repos

akvadrako:
This implements dump and load commands that allow directly copying objects between repos, like requested in #4337. It works but I wanted to get a little feedback before adding docs, tests and so forth.

- The **dump** command outputs a msgpack stream of decrypted objects (id/data pairs) from the repo.
- The **load** command takes objects from stdin, encrypts them and stores them in the repo, except the manifest. It adds any archives it encounters to the existing manifest.
- The **recreate** command gets a new option, `--new-ids` which forces rechunking while ignoring invalid IDs, because the new objects won't match the new encryption key.

I was thinking that a cleaner way to handle this would be to allow multiple keys for one repo. So you could move the old key along with the data and the recryption and **recreate** steps would be optional.

I also would like to make the **dump** command recognize the archives, so **load** doesn't need to guess.

Thoughts?

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1859,6 +1859,7 @@ class ArchiveRecreater:
     def __init__(self, repository, manifest, key, cache, matcher,
                  exclude_caches=False, exclude_if_present=None, keep_exclude_tags=False,
                  chunker_params=None, compression=None, recompress=False, always_recompress=False,
+                 always_recreate=False,
                  dry_run=False, stats=False, progress=False, file_status_printer=None,
                  timestamp=None, checkpoint_interval=1800):
         self.repository = repository
@@ -1877,6 +1878,7 @@ class ArchiveRecreater:
         self.chunker_params = chunker_params or CHUNKER_PARAMS
         self.recompress = recompress
         self.always_recompress = always_recompress
+        self.always_recreate = always_recreate
         self.compression = compression or CompressionSpec('none')
         self.seen_chunks = set()
 
@@ -1893,6 +1895,7 @@ class ArchiveRecreater:
         target = self.create_target(archive, target_name)
         if self.exclude_if_present or self.exclude_caches:
             self.matcher_add_tagged_dirs(archive)
+        print('RECHUNK',target.recreate_rechunkify )
         if self.matcher.empty() and not self.recompress and not target.recreate_rechunkify and comment is None:
             return False
         self.process_items(archive, target)
@@ -2062,7 +2065,11 @@ class ArchiveRecreater:
         if len(source_chunker_params) == 4 and isinstance(source_chunker_params[0], int):
             # this is a borg < 1.2 chunker_params tuple, no chunker algo specified, but we only had buzhash:
             source_chunker_params = (CH_BUZHASH, ) + source_chunker_params
-        target.recreate_rechunkify = self.rechunkify and source_chunker_params != target.chunker_params
+
+        if self.always_recreate:
+            target.recreate_rechunkify = True
+        else:
+            target.recreate_rechunkify = self.rechunkify and source_chunker_params != target.chunker_params
         if target.recreate_rechunkify:
             logger.debug('Rechunking archive from %s to %s', source_chunker_params or '(unknown)', target.chunker_params)
         target.process_file_chunks = ChunksProcessor(
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -36,7 +36,7 @@ try:
     from . import __version__
     from . import helpers
     from .algorithms.checksums import crc32
-    from .archive import Archive, ArchiveChecker, ArchiveRecreater, Statistics, is_special
+    from .archive import Archive, ArchiveChecker, ArchiveItem, ArchiveRecreater, Statistics, is_special
     from .archive import BackupError, BackupOSError, backup_io, OsOpen, stat_update_check
     from .archive import FilesystemObjectProcessors, MetadataCollector, ChunksProcessor
     from .archive import has_link
@@ -168,6 +168,10 @@ def with_repository(fake=False, invert_fake=False, create=False, lock=True,
                         kwargs['key'].compressor = args.compression.compressor
                     if secure:
                         assert_secure(repository, kwargs['manifest'], self.lock_wait)
+                            
+                    if 'new_ids' in args and args.new_ids:
+                        kwargs['key'].skip_assert_id = True
+        
                 if cache:
                     with Cache(repository, kwargs['key'], kwargs['manifest'],
                                progress=getattr(args, 'progress', False), lock_wait=self.lock_wait,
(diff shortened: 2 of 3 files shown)


ThomasWaldmann:
please have a look at the travis-ci output, there are issues with pep8 and py35.

akvadrako:
At this point I'm looking more for comments on the whole design. If it needs to be majorly refactored I don't see how minor issues like style are relevant yet.

src/borg/crypto/key.py
```diff
@@ -170,6 +170,10 @@ class KeyBase:
     # was supplied, and if an empty passphrase works, then Borg won't ask for one.
     logically_encrypted = False
 
+    # skip the check for valid IDs.
+    # This is required after loading data from a repo with an alternative key.
```
ThomasWaldmann, line 174: ... after "borg load" because data is stored to IDs which are valid at origin repo, but not here.

src/borg/crypto/key.py
```diff
@@ -170,6 +170,10 @@ class KeyBase:
...
+    skip_assert_id = False
```
ThomasWaldmann, line 175: maybe avoid double negation and use `id_check = True`.

src/borg/archiver.py
```diff
@@ -1582,6 +1693,7 @@ def do_recreate(self, args, repository, manifest, key, cache):
                                      exclude_caches=args.exclude_caches, exclude_if_present=args.exclude_if_present,
                                      keep_exclude_tags=args.keep_exclude_tags, chunker_params=args.chunker_params,
                                      compression=args.compression, recompress=recompress, always_recompress=always_recompress,
+                                     always_recreate=args.new_ids,
```
ThomasWaldmann, line 1696: naming issue

src/borg/archive.py
```diff
@@ -2062,7 +2065,11 @@ def create_target(self, archive, target_name=None):
         if len(source_chunker_params) == 4 and isinstance(source_chunker_params[0], int):
             # this is a borg < 1.2 chunker_params tuple, no chunker algo specified, but we only had buzhash:
             source_chunker_params = (CH_BUZHASH, ) + source_chunker_params
-        target.recreate_rechunkify = self.rechunkify and source_chunker_params != target.chunker_params
+
+        if self.always_recreate:
+            target.recreate_rechunkify = True
```
ThomasWaldmann, line 2070: naming issue

src/borg/archiver.py
```diff
@@ -1099,6 +1103,113 @@ def print_output(diff, path):
 
         return self.exit_code
 
+    @with_repository(compatibility=(Manifest.Operation.READ,))
+    def do_dump(self, args, repository, manifest, key):
```
ThomasWaldmann, line 1107: hmm, iirc there is some `borg debug dump` already. maybe a source of confusion if this is named the same?

src/borg/archiver.py
```diff
@@ -1099,6 +1103,113 @@ def print_output(diff, path):
...
+            if item[b'id'] == Manifest.MANIFEST_ID:
+                print('SKIP MANIFEST')
+                continue
+
+            id_ = bin_to_hex(key.id_hash(item[b'data']))
+            print(id_[:10], bin_to_hex(item[b'id'])[:10])
+            cdata = key.encrypt(item[b'data'])
+            repository.put(item[b'id'], cdata)
+
+            add_archive(item[b'id'], item[b'data'])
```
ThomasWaldmann, line 1200: a less magic approach would be:
- at the origin repo (when dumping), transmit a set of all archive IDs (or just the full manifest?) **first**
- at the destination repo (when loading) check if the current ID is in that set and then do sth like in add_archive.

src/borg/archiver.py
```diff
@@ -1099,6 +1103,113 @@ def print_output(diff, path):
...
+        # cdata = repository.get(ids[0])
+        # key = key_factory(repository, cdata)
+        
+        def add_archive(id, data):
+            try:
+                raw = msgpack.unpackb(data)
+            except msgpack.UnpackException:
+                return
+
+            if isinstance(raw, dict) and b'name' in data and b'version' in data:
```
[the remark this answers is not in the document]
ThomasWaldmann: but this can be avoided, see my "less magic" comment below.

src/borg/archiver.py
```diff
@@ -1099,6 +1103,113 @@ def print_output(diff, path):
...
+        while True:
+            try:
+                item = unpacker.unpack()
+            except msgpack.OutOfData:
+                print('OUT OF DATA')
+                break
+            
+            if item[b'id'] == Manifest.MANIFEST_ID:
+                print('SKIP MANIFEST')
+                continue
```
ThomasWaldmann, line 1193: you are throwing away precious information (IDs of archives) here which you then need to determine again.

the problem is just that scan() very likely encounters the manifest **last** and you need it **first**. But as the manifest ID is known, that should be easy to special case.

src/borg/archiver.py
```diff
@@ -168,6 +168,10 @@ def wrapper(self, args, **kwargs):
                         kwargs['key'].compressor = args.compression.compressor
                     if secure:
                         assert_secure(repository, kwargs['manifest'], self.lock_wait)
+                            
+                    if 'new_ids' in args and args.new_ids:
+                        kwargs['key'].skip_assert_id = True
```
ThomasWaldmann, line 173: naming issue

ThomasWaldmann:
btw, the issue you linked is talking about tar import/export, so maybe it is not the best fit.

also, if you are transferring archives between repos, maybe rather talk about transferring archives. (not "objects").

ThomasWaldmann:
Considering multiple keys: such stuff is planned for the crypto milestone, I'ld rather avoid it for now or this PR might get way out of scope.

ThomasWaldmann:
about performance / efficiency:
- dump: read, authenticate, decrypt, **decompress**
- deduplicated transfer over some pipe, with optional **pipe compression** outside of borg
- load: **compress**, encrypt, authenticate, write (but storing to bad IDs)
- recreate: read, authenticate, decrypt, decompress, rechunk, compress, encrypt, authenticate, write

the stuff marked in bold maybe could be avoided for better performance (in a future PR).

src/borg/archiver.py
akvadrako: Indeed. Perhaps "save" is better ?

src/borg/archiver.py
akvadrako: I agree that it would be better for the origin to identify the archives, though instead of a list I was thinking to tag the archive packets.

akvadrako:
Thanks for the thorough review. I agree with your concerns. My use cases for this was to combine two repos and change the encryption method. This was a dumb and safe way to do it, but it has problems.

1. I can't actually run recreate on my backup host (probably too little RAM). It's anyway very slow. 
2. The skip-ids thing is insecure. I guess that's fine if soon we get multiple keys.
3. I don't think it'll interact well with deduplication.
4. No way to transfer selected archives.

In light of these limitations, maybe it's better to go with a smarter approach, and actually walk through archives, keeping track of seen files (or chunks?). For me, avoiding duplicated data is a must but this will require state. I guess the smartest way to do this would be to use the index of the target repo for deduplication during the "dump" stage.

What do you think?

akvadrako:
> about performance / efficiency:
> - dump: **decompress**
> - load: **compress**

I did not even realize I was decompressing stuff :/ I have to understand the chunking a little better to go for a deeper approach.

infectormp:
>     1. I can't actually run recreate on my backup host (probably too little RAM). 
Maybe this https://github.com/borgbackup/borg/issues/5202 fix this problem

akvadrako:
> > 1. I can't actually run recreate on my backup host (probably too little RAM). 
> 
> Maybe this #5202 fix your problem

Unfortunately not. I tried with commit ebd83a5688f2807d088d51cf262d2701fc444b39 but it still got OOM-killed after using 1GB of RAM.

ThomasWaldmann:
If one wants to copy selected archives only, I guess one would need to make `borg recreate` (and borg in general) capable of dealing with multiple repos (multiple encryption keys, caches, ...) at once.

Then `recreate` could read from origin repo and write to destination repo. If the local chunks cache of the destination repos works as usual, it would not transfer or compress duplicate content again, as usual.

IIRC, we might have an open issue about generalizing `recreate` in that way.

ThomasWaldmann:
@akvadrako btw, you you can't run recreate on your backup host, just run it on some client having access to that backup repo. that will resolve your memory/cpu issue, but of course will do a lot of network data transfer.

enkore:
My two cents:

I wouldn't add something like --disable-id-check, because it normalizes broken data being committed to the repo. Instead I would recommend to keep the contents of a repository clean and correct, so that any broken data is a clear signal of something being wrong.

The overall approach seems sound, but would need just some tweaks in my opinion.

- Instead of iterating the repository directly, traverse the objects+chunks of archives, and include that metadata in the output (manifest or manifest-like).
- Dump: process the objects to fix up IDs to one of the encryption-less methods. This requires one extra pass of hash over all data for a dump, which seems reasonable (decrypt is significantly more expensive).
- Load: Encrypt to the current key, which requires ID fixups, rechunking is probably not necessary, add declared archives.

An alternative design would be more like the export-tar/import-tar pair, just using the proprietary Borg "Item" structure. This might actually be less code and less difficult code, because you don't need to worry about IDs at all, you can just pack everything into one huge msgpack object that you stream, and rebuild an Archive on the other side with the usual approach.

akvadrako:
Hi @enkore, I have been thinking along the same lines, which has kept me unmotivated from working on this more. I think the design of borg just doesn't take into account moving chunks between repos efficiently.

> * Dump: process the objects to fix up IDs to one of the encryption-less methods. This requires one extra pass of hash over all data for a dump, which seems reasonable (decrypt is significantly more expensive).
> 
> * Load: Encrypt to the current key, which requires ID fixups, rechunking is probably not necessary, add declared archives.

What I don't like is you would have to keep track of every converted ID during the conversion process.

> An alternative design would be more like the export-tar/import-tar pair, just using the proprietary Borg "Item" structure. This might actually be less code and less difficult code, because you don't need to worry about IDs at all, you can just pack everything into one huge msgpack object that you stream, and rebuild an Archive on the other side with the usual approach.

I'm thinking this is probably the way to go. It's not so efficient, but it avoids the risk of corruption. The one additional thing I would really like is archive diffs - omitting chunks that already exist is previously synced archives.

enkore:
Okay, so to deduplicate the transfer itself you'll need interactivity
(and persistence to be more reasonable), so we're talking more about
a pair like "borg send" / "borg receive" rather than "borg import/export".

Here's how that could work.

"borg send" keeps a translation hashtable in `$CACHE/send/<remote repository id>`.
This maps local object IDs to remote object IDs. They're different because both
repositories are on different keys.

"borg send" builds a stream like I described above, but for every file chunk
it looks the ID up in the translation table, if it finds it in there, it
just inserts a reference like `{"local": local-id, "id": remote-id"}`.
The reason we want to communicate our local ID would be that the chunk could
have been deleted in the remote repository, and we can't search the translation
hashtable in reverse easily.

If it doesn't find a match in the translation table, then it inserts the
chunk like this: `{"local": local-id", "data": data}`. The reason we include the
local ID here is that the "receive" side can directly tell us their ID for this
chunk later.

"borg receive" then unpacks that stream. For a reference it checks if it has that
object, and if it does, it can directly add a reference to it. If it doesn't
it sends a request back to the "send" side asking for `{"local": local-id, "id": remote-id"}`.
"send" obliges. Same flow. If it gets a straight chunk, then it computes the ID
as usual, and sends it to the "send" side, besides processing it as usual.

The interface could look like this:

```
borg send repo::archive | ssh remote-server borg receive
borg send repo | ssh remote-server borg receive
-> Sending all archives that don't exist in the remote yet (by name, of course)
```

There are some obvious, and some not so obvious, security implications
to doing this.

Note: Avoid rechunking (like above), since it's slow, annoying and gets everywhere.
That might cause some duplication if both repositories are used independently, but
I suspect for typical send/receive use cases this isn't so important, since I'd wager
these mostly follow a "primary -> backup" scheme.

Overall this doesn't _sound_ too complicated, but it's still a lot of work, since it'll need
a separate RPC protocol (and you wouldn't necessarily want to reuse the existing RPC code,
although using the core itself may be reasonable) and probably handling for a bunch of corner
cases, which means lots of testing to go around.

ThomasWaldmann:
This is WIP/Draft for rather long. Close?

ThomasWaldmann:
Maybe it would be useful to condense the above comments / insights into a ticket, for next try. :-)

PR #5188, opened 2020-05-17
docs: explain hash collision

### LINKED ISSUE #4884: Borg may restore corrupt files without noticing
**Please close this report if I am wrong in the following assumption**: after restoring some file Borg doesn't compute checksum (say, SHA256) over the whole file to validate its integrity, instead such integrity is _derived_ from individual chunk integrity and the integrity of Borg internal data structures.

------

I assume so because otherwise first `read()` operation of any byte via BorgFS would require restoring whole file and computing its SHA.  I also grepped the code and had an impression that my assumption is correct.

If so then suppose we backup some file whose chunk representation is `C1 C2 C3`.  We compute `Key1 = MAC(C1), Key2 = MAC(C2), Key2 = MAC(C3)`.  `Key2` is found in cache, because there was some _other_ chunk `C42` that happens to have `MAC(C42) == Key2`, so on restore we'll get `C1 C42 C3` (such restoration won't trigger [IntegrityError](https://borgbackup.readthedocs.io/en/latest/faq.html#i-get-an-integrityerror-or-similar-what-now), nor would the problem be found with [`check`](https://borgbackup.readthedocs.io/en/latest/faq.html#can-borg-verify-data-integrity-of-a-backup-archive) operation).  Also note than once there, such corruption may penetrate any number of time slices and any number of files with affected `C2` chunk. 

Yes, we all know that while theoretical collisions are [_guaranteed_](https://en.wikipedia.org/wiki/Pigeonhole_principle), the probability of such a collision to actually happen is [_extremely low_](https://en.wikipedia.org/wiki/Birthday_problem).  However here are some popular but wrong assumptions:

- it won't happen in foreseeable future
- there are billions of people on Earth, it won't happen to me
- "You were right, it happened yesterday to George!  So the _next_ collision is far in the future!"

Here are some correct assuptions:

- it can happen tomorrow
- it can happen to me
- _it could have happen to me already, I just didn't notice it yet_ (there are much more backup operations then there are restore operations, and users trust Borg and do not verify integrity by external means)

Unless Borg truly follows [Horton Principle](https://en.wikipedia.org/wiki/Horton_Principle) and authenticates/verifies file as a whole, someone out there may end up restoring executable with a random code injection, or simply loosing their precious selfies.

My humble estimation is that data integrity is way more important than performance of BorgFS, so SHA must be validated on restored files.  Until such verification is implemented please update the docs to clearly warn users of potential risks.

ImUrX:
Trying to fix #4884 

I'm listening to any problem that you guys may have on the doc, but please limit yourselves.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -345,14 +345,19 @@ resolution from that.
 Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
 and we have a "perfect" hash algorithm, we can think that the probability of collision 
 would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
-we have a billion chunks (:math:`p=10^9`) (a billion chunks would be about 2.1ZB).
-The probability would be around to 4.3e-60.
+we have a 1,000 million chunks (:math:`p=10^9`) (a 1,000 million chunks would be about 2,100TB).
+The probability would be around to **4.3e-60**.
+
+According to the `US Weather Service <https://www.weather.gov/safety/lightning-odds>`_ you
+have 1:1,222,000 odds of being struck in a given year (in USA obviously). This is **54** 
+orders of magnitude more probable than a SHA-256 collision.
 
 Still, the real question is if Borg tries to not make this happen?
 
 Well... it used to not check anything but there was a feature added which saves the size
 of the chunks too, so the size of the chunks is compared to the size that you got with the
-hash, this doesn't save us from everything but reduces the chances of corrupting the file.
+hash and if the check says there is a mismatch it will raise an exception instead of corrupting
+the file. This doesn't save us from everything but reduces the chances of corruption.
 There are other ways of trying to escape this but it would affect performance so much that
 it wouldn't be worth it and it would contradict Borg's design, so if you don't want this to
 happen, simply don't use Borg.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -345,8 +345,8 @@ resolution from that.
 Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
 and we have a "perfect" hash algorithm, we can think that the probability of collision 
 would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
-we have :math:`10^9` chunks (:math:`p=10^9`) (:math:`10^9` chunks would be about 2100TB).
-The probability would be around to `4.3e-60 <https://www.wolframalpha.com/input/?i=4.3e-60>`_.
+we have 1000 million chunks (:math:`p=10^9`) (1000 million chunks would be about 2100TB).
+The probability would be around to 0.0000000000000000000000000000000000000000000000000000000000043.
 
 A mass-murderer space rock happens about once every 30 million years on average.
 This leads to a probability of such an event occurring in the next second to about :math:`10^{-15}`.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -351,7 +351,8 @@ The probability would be around to **4.3e-60**.
 A mass-murderer space rock happens about once every 30 million years on average.
 This leads to a probability of such an event occurring in the next second to about :math:`10^{-15}`.
 That's **45** orders of magnitude more probable than the SHA-256 collision. Briefly stated,
-if you find SHA-256 collisions scary then your priorities are wrong.
+if you find SHA-256 collisions scary then your priorities are wrong. This example was grabbed from
+`this SO answer <https://stackoverflow.com/a/4014407/13359375>`_, it's great honestly.
 
 Still, the real question is if Borg tries to not make this happen?
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -330,6 +330,33 @@ needs to be ascertained and fixed.
 issues. We recommend to first run without ``--repair`` to assess the situation.
 If the found issues and proposed repairs seem right, re-run "check" with ``--repair`` enabled.
 
+How probable is it to get a hash collision problem?
+---------------------------------------------------
+
+If you noticed, there are some issues (:issue:`170` (**warning: hell**) and :issue:`4884`)
+about the probability of a chunk having the same hash as another chunk, making the file
+corrupted because it grabbed the wrong chunk. This is called the `Birthday Problem
+<https://en.wikipedia.org/wiki/Birthday_problem>`_.
+
+There is a lot of probability in here so, I can give you my interpretation of
+such math but it's honestly better that you read it yourself and grab your own
+resolution from that.
+
+Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
+and we have a "perfect" hash algorithm, we can think that the probability of collision 
+would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -330,6 +330,33 @@ needs to be ascertained and fixed.
...
+<https://en.wikipedia.org/wiki/Birthday_problem>`_.
+
+There is a lot of probability in here so, I can give you my interpretation of
+such math but it's honestly better that you read it yourself and grab your own
+resolution from that.
+
+Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
+and we have a "perfect" hash algorithm, we can think that the probability of collision 
+would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
+we have a billion chunks (:math:`p=10^9`) (a billion chunks would be about 2.1ZB).
```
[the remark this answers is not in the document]
ThomasWaldmann: Also, "ZB" is an amount / a unit of data most people are not familiar with. Rather express it in TB?

docs/faq.rst
```diff
@@ -330,6 +330,33 @@ needs to be ascertained and fixed.
...
+The probability would be around to 4.3e-60.
```
ThomasWaldmann, line 349: How many times is that getting hit by lightning? :-)

docs/faq.rst
```diff
@@ -330,6 +330,33 @@ needs to be ascertained and fixed.
...
+
+Still, the real question is if Borg tries to not make this happen?
+
+Well... it used to not check anything but there was a feature added which saves the size
+of the chunks too, so the size of the chunks is compared to the size that you got with the
+hash, this doesn't save us from everything but reduces the chances of corrupting the file.
```
ThomasWaldmann, line 355: hmm, iirc it just raises an exception. so it will just not silently corrupt a file, but still not result in a correct file, right?

ImUrX:
woah, thanks for the suggestions, will finish somo stuff and update the pr :P

docs/faq.rst
```diff
@@ -330,6 +330,33 @@ needs to be ascertained and fixed.
...
+corrupted because it grabbed the wrong chunk. This is called the `Birthday Problem
+<https://en.wikipedia.org/wiki/Birthday_problem>`_.
+
+There is a lot of probability in here so, I can give you my interpretation of
+such math but it's honestly better that you read it yourself and grab your own
+resolution from that.
+
+Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
+and we have a "perfect" hash algorithm, we can think that the probability of collision 
+would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
```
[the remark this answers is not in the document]
ImUrX: sorry but the math would be 
```
p^2
-------
2^(n+1)
```
from what i understood. im basing myself on the [stackexchange answer](https://stackoverflow.com/a/4014407/13359375) that says 
> if we have a "perfect" hash function with output size n, and we have p messages to hash (individual message length is not important), then probability of collision is about p^2/2^(n+1) (this is an approximation which is valid for "small" p, i.e. substantially smaller than 2^(n/2))

docs/faq.rst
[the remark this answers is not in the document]
ImUrX: yeah i understand the billion one, its the same on spanish..,

docs/faq.rst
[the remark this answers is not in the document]
ImUrX: the other way of expressing it is ``2^(-n - 1) p^2`` but that doesnt seem very nice

docs/faq.rst
```diff
@@ -330,6 +330,38 @@ needs to be ascertained and fixed.
...
+we have a 1,000 million chunks (:math:`p=10^9`) (a 1,000 million chunks would be about 2,100TB).
```
ThomasWaldmann, line 348: we have another language / interpretation issue here:
x,yyy (en) == x.yyy (de) - thousands separator comma/point
x,yyy (de) == x.yyy (en) - decimal comma/point

so i'ld suggest better not using a thousands separator comma.

docs/faq.rst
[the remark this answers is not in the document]
ThomasWaldmann: oh, sorry, got the formula wrong. ignore.

docs/faq.rst
```diff
@@ -330,6 +330,38 @@ needs to be ascertained and fixed.
...
+The probability would be around to **4.3e-60**.
+
+According to the `US Weather Service <https://www.weather.gov/safety/lightning-odds>`_ you
+have 1:1,222,000 odds of being struck in a given year (in USA obviously). This is **54** 
+orders of magnitude more probable than a SHA-256 collision.
```
ThomasWaldmann, line 353: ... of being struck by lightning ...

ThomasWaldmann: good example. thanks for researching. but i now realized that people will say: yeah, but people **do** get struck by lightning and i want to have good backups. so maybe the killer space rock example from SO is even better.

ImUrX: yeah i thought the same too

docs/faq.rst
ImUrX: im replacing the 1000000 to exp notation

ImUrX:
Fast Fix

docs/faq.rst
```diff
@@ -345,12 +345,13 @@ resolution from that.
 Assuming that all your chunks have a size of :math:`2^{21}` bytes (approximately 2.1 MB)
 and we have a "perfect" hash algorithm, we can think that the probability of collision 
 would be of :math:`p^2/2^{n+1}` then, using SHA-256 (:math:`n=256`) and for example
-we have a 1,000 million chunks (:math:`p=10^9`) (a 1,000 million chunks would be about 2,100TB).
+we have 1e9 chunks (:math:`p=10^9`) (1e9 chunks would be about 2100TB).
```
ThomasWaldmann: you already have the p=10^9 and e-notation is no good for mathematically less inclined people. 

so just say `1000 million` and everybody knows what it means.

docs/faq.rst
```diff
@@ -345,12 +345,13 @@ resolution from that.
...
 The probability would be around to **4.3e-60**.
 
-According to the `US Weather Service <https://www.weather.gov/safety/lightning-odds>`_ you
-have 1:1,222,000 odds of being struck in a given year (in USA obviously). This is **54** 
-orders of magnitude more probable than a SHA-256 collision.
+A mass-murderer space rock happens about once every 30 million years on average.
+This leads to a probability of such an event occurring in the next second to about :math:`10^{-15}`.
+That's **45** orders of magnitude more probable than the SHA-256 collision. Briefly stated,
+if you find SHA-256 collisions scary then your priorities are wrong.
```
ThomasWaldmann: great :)

ImUrX:
What do you think now? I even linked the answer to wolfram-alpha which shows you the exact form

ThomasWaldmann:
Guess I'ld have rather copy&pasted that long number from there and avoided an external link.

Also the 10^9 is still there...

ImUrX:
sorry i misunderstood lol

ThomasWaldmann:
Thanks! Would you like to do a backport to 1.1-maint, so it gets into next 1.1 release (soon!)?

ImUrX:
@ThomasWaldmann sorry, things happened after that. I think its kind of late for that now, no?

ThomasWaldmann:
@ImUrX you can do it now, so it will be in next 1.1.x release.

PR #5199, opened 2020-05-26
Improve sizeof_fmt()

darkdragon-001:
- Use built-in sign handling
- Format integer correctly when using biggest unit
- Use consistent format inside and outside of loop

I came here from [StackOverflow](https://stackoverflow.com/questions/1094841/reusable-library-to-get-human-readable-version-of-file-size#comment68302428_1094933) and used the opportunity to improve the code even further.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -271,16 +271,13 @@ def parse_file_size(s):
 
 
 def sizeof_fmt(num, suffix='B', units=None, power=None, sep='', precision=2, sign=False):
-    prefix = '+' if sign and num > 0 else ''
-
+    sign = '+' if sign else ''
+    fmt = '{0:{1}d}{3}{4}{5}' if isinstance(num, int) else '{0:{1}.{2}f}{3}{4}{5}'
     for unit in units[:-1]:
         if abs(round(num, precision)) < power:
-            if isinstance(num, int):
-                return "{}{}{}{}{}".format(prefix, num, sep, unit, suffix)
-            else:
-                return "{}{:3.{}f}{}{}{}".format(prefix, num, precision, sep, unit, suffix)
+            return fmt.format(num, sign, precision, sep, unit, suffix)
         num /= float(power)
-    return "{}{:.{}f}{}{}{}".format(prefix, num, precision, sep, units[-1], suffix)
+    return fmt.format(num, sign, precision, sep, units[-1], suffix)
 
 
 def sizeof_fmt_iec(num, suffix='B', sep='', precision=2, sign=False):
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -272,12 +272,14 @@ def parse_file_size(s):
 
 def sizeof_fmt(num, suffix='B', units=None, power=None, sep='', precision=2, sign=False):
     sign = '+' if sign else ''
-    fmt = '{0:{1}d}{3}{4}{5}' if isinstance(num, int) else '{0:{1}.{2}f}{3}{4}{5}'
+    fmt = '{0:{1}.{2}f}{3}{4}{5}'
+    prec = 0
     for unit in units[:-1]:
         if abs(round(num, precision)) < power:
-            return fmt.format(num, sign, precision, sep, unit, suffix)
+            return fmt.format(num, sign, prec, sep, unit, suffix)
         num /= float(power)
-    return fmt.format(num, sign, precision, sep, units[-1], suffix)
+        prec = precision
+    return fmt.format(num, sign, prec, sep, units[-1], suffix)
 
 
 def sizeof_fmt_iec(num, suffix='B', sep='', precision=2, sign=False):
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
thanks for trying to improve this, but there is an issue now.

from travis-ci logs:
```
    def sizeof_fmt(num, suffix='B', units=None, power=None, sep='', precision=2, sign=False):
        sign = '+' if sign else ''
        fmt = '{0:{1}d}{3}{4}{5}' if isinstance(num, int) else '{0:{1}.{2}f}{3}{4}{5}'
        for unit in units[:-1]:
            if abs(round(num, precision)) < power:
>               return fmt.format(num, sign, precision, sep, unit, suffix)
E               ValueError: Unknown format code 'd' for object of type 'float'

.tox/py35/lib/python3.5/site-packages/borg/helpers/parseformat.py:278: ValueError
```

darkdragon-001:
@ThomasWaldmann What is the desired behavior:
- Should integers stay integers (rounded division)?
- Or should integers be converted to floats when division requires?

If I see it correctly, before it stayed integer only on the first loop run and was then converted to float because of the division. Is this the desired behavior? Or should we even force this? First run is bytes and should be without precision and consecutive runs should be should always show precision?

darkdragon-001:
I think this is good now:
- first run (without divisor, e.g. Bytes) does not allow fractions
- all consecutive runs show the specified precision

darkdragon-001:
I modified a test such that '0' with `sign=True` will display as `+0 B`. Sign enforcing should go for all numbers!

src/borg/testsuite/helpers.py
```diff
@@ -579,7 +579,7 @@ def test_file_size_precision():
 
 def test_file_size_sign():
     si_size_map = {
-        0: '0 B',
+        0: '+0 B',
```
ThomasWaldmann, line 582: hmm, when used in a difference context (like: how much bytes do we have additionally / less?) having a "neutral" 0 would be nice.

darkdragon-001: Why would you force signs then? Minus sign is always displayed. I changed to the default Python formater sign handling.

src/borg/helpers/parseformat.py
```diff
@@ -271,16 +271,15 @@ def parse_file_size(s):
...
         if abs(round(num, precision)) < power:
-            if isinstance(num, int):
-                return "{}{}{}{}{}".format(prefix, num, sep, unit, suffix)
-            else:
-                return "{}{:3.{}f}{}{}{}".format(prefix, num, precision, sep, unit, suffix)
+            return fmt.format(num, sign, prec, sep, unit, suffix)
         num /= float(power)
-    return "{}{:.{}f}{}{}{}".format(prefix, num, precision, sep, units[-1], suffix)
+        prec = precision
+    return fmt.format(num, sign, prec, sep, units[-1], suffix)
```
ThomasWaldmann, line 282: to avoid repeating the format code and also to have a single point where the function returns:
```
    fmt = '{0:{1}.{2}f}{3}{4}{5}'
    prec = 0
    for unit in units[:-1]:
        if abs(round(num, precision)) < power:
            break
        num /= float(power)
        prec = precision
    else:
        unit = units[-1]
    return fmt.format(num, sign, prec, sep, unit, suffix)
```

src/borg/testsuite/helpers.py
ThomasWaldmann: so, the test should not be changed.

src/borg/helpers/parseformat.py
```diff
@@ -271,16 +271,15 @@ def parse_file_size(s):
 
 
 def sizeof_fmt(num, suffix='B', units=None, power=None, sep='', precision=2, sign=False):
-    prefix = '+' if sign and num > 0 else ''
-
+    sign = '+' if sign else ''
```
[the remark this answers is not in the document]
ThomasWaldmann: (the point of it is to **not** include num == 0)

src/borg/testsuite/helpers.py
darkdragon-001: How can one force that any sign is always displayed then? This would not be possible. IMHO, it makes more sense to have the following two possibilities:
```
-1
0
1
```
and
```
-1
+0
+1
```

src/borg/helpers/parseformat.py
darkdragon-001: Didn't know about for/else. Nice tweak for a single format statement!

src/borg/testsuite/helpers.py
ThomasWaldmann: Obviously, the sign=True param was never intended to display a sign for 0, see the unit test.

And I already explained why: this is about difference.

E.g. if something got bigger, we want to display +123MB.
If something got small, we want to display -456MB.
But if it is the same, we want just a plain 0 - no plus, no minus in that case.

ThomasWaldmann: see my other comment.

darkdragon-001: I do not intend to deviate from the default python and mathematical convention. You can finish the PR according to your needs.

ThomasWaldmann: OK, I'll finish it. Thanks!

ThomasWaldmann:
superseded by #5246

Issue #5202, opened 2020-05-30
borg uses too much memory (with misc. operations)


antlarr:
## Is this a BUG / ISSUE report or a QUESTION?

ISSUE

#### Your borg version (borg -V).

client: 1.1.11 (openSUSE TW system package)
server: 1.1.11 (built locally from debian package)

#### Operating system (distribution) and version.

client: openSUSE Tumbleweed (20200516)
server: osmc (2020.03-1).  Debian-based distribution for mediacenters.

#### Hardware / network configuration, and filesystems used.

client: x86_64
server: aarch64 (a vero 4k mediacenter with external usb hard disk)

#### How much data is handled by borg?

Original size of all archives: 2.11 TB
Each archive original size is between 31 and 60 GB
There's a total of 43 archives.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

`borg recreate -e 'home/user/large-directory' -e 'home/user/large-file' ssh://borgbackup@server/home/borgbackup/repositories/hostname::hostname-2019-09-30T00:01:57`

## Describe the problem you're observing.

Recreating the archive consumes a lot of memory. I checked it kept growing during the execution of the recreate command peaking just at the end at a VmSize of 34155864 k (~34GB).

After some debugging, I could observe the problem lies in the self.responses dictionary in remote.py, which keeps all the messages received during the processing.

In the middle of a debugging session I got this:
```
(Pdb) p total_size(self.responses)
25272256788
(Pdb) p len(self.responses.keys())
89057
```
(total_size is the function from https://code.activestate.com/recipes/577504/ which returns the total size in memory of an object and its contents)

It seems the only way for an item to be removed from the responses dictionary is by first getting its msgid into the waiting_for list. I checked (in a different debugging session than the commands above) and waiting_for doesn't seem to change at all during the execution. It stayed with the same single content every time I checked:

```
(Pdb) p waiting_for               
[45112]
(Pdb) p max(self.responses.keys())
24628
```

Of course, the max msgid value of self.responses.keys() kept growing every time I checked, just as its length.

I don't know where waiting_for got that value from, but I'd say maybe waiting_for is waiting for the last msgid coming from the server, so every response will just be stored in self.responses and never discarded.

Note that all this happens inside the `self.repository.commit()` call in Archive.save . This is the backtrace that can be seen by stopping the execution at mostly any time during the recreate (after a few seconds after starting):

```
Traceback (most recent call last):
[...]

PR #5206, opened 2020-06-01
rename local preload() function to not overwrite keyword argument of same name

elho:
The locally defined preload() function overwrites the preload boolean keyword
argument, always evaluating to true, so preloading is done, even when not
requested by the caller, causing a memory leak.
Also move its definition outside of the loop.

This issue was found by Antonio Larrosa in borg issue #5202.

PR #5207, opened 2020-06-02
rename local preload() function to not overwrite keyword argument of same name

elho:
The locally defined preload() function overwrites the preload boolean keyword
argument, always evaluating to true, so preloading is done, even when not
requested by the caller, causing a memory leak.
Also move its definition outside of the loop.

This issue was found by Antonio Larrosa in borg issue #5202.


elho:
1.1.x backport of PR #5206.

ThomasWaldmann:
thanks!

PR #5226, opened 2020-06-14
--umask is for the local borg process only, fixes #4947

### LINKED ISSUE #4947: --umask not honored in ssh authorized_keys forced command
I'm setting up a remote borg server (v1.1.10 on both sides), and I tried setting the `--umask` option in the `authorized_keys` file like so:

```restrict,command="borg serve --umask 007" ssh-ed25519 AAAA....```

Unfortunately, repositories and archives created using this ssh key nonetheless use the default 077 umask. Specifying `--umask 007` explicitly on the client side with each operation (init, create, ...) results in the correct permissions on the server, but the whole idea of the ssh forced command approach is for the server to be able to enforce (or at least set a default for) some options without the client having to worry about it every time.

There is another open issue (#346) for repo-level default values which would also meet my need, but I'm filing this as a technically separate issue.

There is also a closed issue (#1168) for supporting `borg serve --append-only` which might be a good template for adding similar support for `borg serve --umask`?

### PR DESCRIPTION
**Pull request by ThomasWaldmann**
The umask value is NOT transmitted from client to server any more,
so the borg client can not influence the borg server umask any more.

If one wants to have a specific umask on the server side, one needs to
use a ssh forced command in .ssh/authorized_keys file.

OTOH, as the default value is 077 (in general, for client as well as for
the server) and the server does not take the value from the client any more,
there usually should be no need to give it on the server side, IF you are
happy with the default value.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2617,7 +2617,7 @@ class Archiver:
             add_common_option('--show-rc', dest='show_rc', action='store_true',
                               help='show/log the return code (rc)')
             add_common_option('--umask', metavar='M', dest='umask', type=lambda s: int(s, 8), default=UMASK_DEFAULT,
-                              help='set umask to M (local and remote, default: %(default)04o)')
+                              help='set umask to M (local only, default: %(default)04o)')
             add_common_option('--remote-path', metavar='PATH', dest='remote_path',
                               help='use PATH as borg executable on the remote (default: "borg")')
             add_common_option('--remote-ratelimit', metavar='RATE', dest='remote_ratelimit', type=int,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ThomasWaldmann** [comment]
Please review and comment whether you think we should have this in 1.1-maint also.

File: src/borg/archiver.py [hunk 1]
```diff
@@ -4413,12 +4413,12 @@ def get_args(self, argv, cmd):
                     'restrict_to_repositories',
                     'append_only',
                     'storage_quota',
+                    'umask',
                 }
                 whitelist = {
                     'debug_topics',
                     'lock_wait',
                     'log_level',
-                    'umask',
```
ThomasWaldmann, line 4421: in case an old client is sending `--umask=XXX`, the server (running this code) would just ignore it and use either the default umask or whatever is given via a ssh forced command.

src/borg/remote.py
```diff
@@ -647,7 +647,6 @@ def borg_cmd(self, args, testing):
         # give some args/options to 'borg serve' process as they were given to us
         opts = []
         if args is not None:
-            opts.append('--umask=%03o' % args.umask)
```
ThomasWaldmann, line 650: a new client (using this code) would not send `--umask=XXX` to the server.

an old server would just use the default umask (077) in that case.

PR #5238, opened 2020-06-24
docs: clarify --exclude-if-present in recreate, #5193

### LINKED ISSUE #5193: borg recreate and --exclude-if-present flag behavior isn't very clear

Yes

Documentation Bug, IMO

Linux home-desktop 5.3.0-53-generic #47~18.04.1-Ubuntu SMP Thu May 7 13:10:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Distro - KDE Neon

borg 1.1.11

ext4, local disk

12Gig repo

borg recreate --exclude-if-present .nobackup /data/borg-backup

So I figured that there was a lot of cruft built up and my overall backup size had grown a lot (12G),
Mounting the backup, it was clear that a lot of thing in my home folder were the problem
for instance: `/home/raghu/.arduino15` isn't a folder I need. Ditto for `/home/raghu/code/go`
For my nightly backups, I have `borg create <archive> --exclude-if-present .nobackup /home /etc` and my preference is to exclude folders with a `touch /path/to/exclude/.nobackup` 

So, before running recreate, I dropped a lot of `.nobackup`  files into folders that I needed to delete. In my mind, I figure that these folders will be excluded by virtue of me dropping `.nobackup` files into the folders. It doesn't work of course. 

Obviously (well, not obvious at all - took me more than a day), the penny finally drops and I realize that recreate ONLY looks at the files/folders already IN the archive - and won't touch the fs at all.

So now, I've reverted to using `--exclude-from-file` - but the behavior of  `--exclude-if-present` should really be highlighted in the documentation on the `recreate` page.

fantasya-pbem:
fixes #5193

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4092,9 +4092,10 @@ class Archiver:
 
         Important: Repository disk space is **not** freed until you run ``borg compact``.
 
-        ``--exclude``, ``--exclude-from``, ``--exclude-if-present``, ``--keep-exclude-tags``, and PATH
-        have the exact same semantics as in "borg create". If PATHs are specified the
-        resulting archive will only contain files from these PATHs.
+        ``--exclude``, ``--exclude-from``, ``--exclude-if-present``, ``--keep-exclude-tags``
+        and PATH have the exact same semantics as in "borg create", but they only check
+        for files in the archives and not in the local file system. If PATHs are specified,
+        the resulting archives will only contain files from these PATHs.
 
         Note that all paths in an archive are relative, therefore absolute patterns/paths
         will *not* match (``--exclude``, ``--exclude-from``, PATHs).


src/borg/archiver.py
```diff
@@ -4096,6 +4096,10 @@ def define_borg_mount(parser):
         have the exact same semantics as in "borg create". If PATHs are specified the
         resulting archive will only contain files from these PATHs.
 
+        ``--exclude-if-present`` just checks for files that are present in the archive,
+        it will not recognize new tag files in the local file system that have been
+        added after your archives were created.
```
ThomasWaldmann, line 4101: true, but isn't that a bit too specific?

recreate always only looks at the stuff in the input archive and never into the filesystem.
so **all** options and arguments refer to the input archive and not to the filesystem.

ThomasWaldmann:
thanks!

PR #5241, opened 2020-06-29
rephrase some warnings, fixes #5164

### LINKED ISSUE #5164: recreate: drop the "experimental feature" warning?

Yes

More of a feature request, really :)

```
backup ~ # borg -V
borg 1.1.9
```

```
backup ~ # lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
backup ~ # cat /etc/debian_version
10.3
```

```
borg recreate --dry-run --verbose --list -e 'pp:home/some.user' '::archive...'
```

`borg recreate` currently prints a warning indicating that the `recreate` feature is still experimental:

```
Type 'YES' if you understand this and want to continue:
```

[This warning has been in the code for 4 years](https://github.com/borgbackup/borg/blame/08a12cc40523b58bdfe92be9e9d937179287e805/src/borg/archiver.py#L1569) (which was the point where `recreate` was originally implemented). As far as I can tell from the currently open bug reports, there are currently no serious issues regarding `recreate`, and looking at the closed pull requests, most of the work on it happened back in 2017.

Is this warning still relevant today, or can it safely be removed?

ThomasWaldmann:
borg check --repair and borg recreate are now present in the code
since rather long, so they are not experimental any more.

but, both bear some risk of data loss:
- there could be an undiscovered bug in the code
- the user might use them wrong (e.g. accidentally excluding everything /
  not matching anything when recreating an archive)
- there might be kinds of corruption borg check --repair can not fix
  and it might make things even worse while trying to fix.

so, if your repo / archive(s) are important, be careful.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -307,9 +307,8 @@ class Archiver:
         """Check repository consistency"""
         if args.repair:
             msg = ("This is a potentially dangerous function.\n"
-                   "check --repair might lead to data loss (in case of bugs or for kinds of\n"
-                   "corruption it is not capable of dealing with).\n"
-                   "BE VERY CAREFUL!\n"
+                   "check --repair might lead to data loss (for kinds of corruption it is not\n"
+                   "capable of dealing with). BE VERY CAREFUL!\n"
                    "\n"
                    "Type 'YES' if you understand this and want to continue: ")
             if not yes(msg, false_msg="Aborting.", invalid_msg="Invalid answer, aborting.",
@@ -1591,8 +1590,7 @@ class Archiver:
     def do_recreate(self, args, repository, manifest, key, cache):
         """Re-create archives"""
         msg = ("This is a potentially dangerous function.\n"
-               "recreate might lead to data loss (if used wrongly or in case of bugs).\n"
-               "BE VERY CAREFUL!\n"
+               "recreate might lead to data loss (if used wrongly). BE VERY CAREFUL!\n"
                "\n"
                "Type 'YES' if you understand this and want to continue: ")
         if not yes(msg, false_msg="Aborting.", truish=('YES',),
@@ -2868,8 +2866,7 @@ class Archiver:
         The check command verifies the consistency of a repository and the corresponding archives.
 
         check --repair is a potentially dangerous function and might lead to data loss
-        (in case of bugs or for kinds of corruption it is not capable of dealing with).
-        BE VERY CAREFUL!
+        (for kinds of corruption it is not capable of dealing with). BE VERY CAREFUL!
 
         First, the underlying repository data files are checked:
 
@@ -4099,8 +4096,7 @@ class Archiver:
         Recreate the contents of existing archives.
 
         recreate is a potentially dangerous function and might lead to data loss
-        (if used wrongly or in case of bugs).
-        BE VERY CAREFUL!
+        (if used wrongly). BE VERY CAREFUL!
 
         Important: Repository disk space is **not** freed until you run ``borg compact``.
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -306,7 +306,7 @@ class Archiver:
     def do_check(self, args, repository):
         """Check repository consistency"""
         if args.repair:
-            msg = ("This is a potentially dangerous functionality.\n"
+            msg = ("This is a potentially dangerous function.\n"
                    "check --repair might lead to data loss (in case of bugs or for kinds of\n"
                    "corruption it is not capable of dealing with).\n"
                    "BE VERY CAREFUL!\n"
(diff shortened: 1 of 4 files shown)


src/borg/archiver.py
```diff
@@ -1587,7 +1590,10 @@ def do_upgrade(self, args, repository, manifest=None, key=None):
     @with_repository(cache=True, exclusive=True, compatibility=(Manifest.Operation.CHECK,))
     def do_recreate(self, args, repository, manifest, key, cache):
         """Re-create archives"""
-        msg = ("recreate is an experimental feature.\n"
+        msg = ("This is a potentially dangerous functionality.\n"
+               "recreate might lead to data loss (if used wrongly or in case of bugs).\n"
```
horazont, line 1594: s/data loss/data loss in subtle ways/?

Because it seems obvious that using a borg command incorrectly can lead to data loss (e.g. borg delete). The key thing about recreate is that it can subtly lead to data loss if you accidentally exclude all the things.

src/borg/archiver.py
```diff
@@ -4088,7 +4098,9 @@ def define_borg_mount(parser):
         recreate_epilog = process_epilog("""
         Recreate the contents of existing archives.
 
-        This is an *experimental* feature. Do *not* use this on your only backup.
+        recreate is a potentially dangerous functionality and might lead to data loss
```
horazont, line 4101: Dito.

src/borg/archiver.py
ThomasWaldmann: hmm, as a non-native english speaker, i guess i would not really get more informed when reading "in subtle ways".

not sure if it is due to my english skills or just not being more information there. :-)

horazont: Me not being a native speaker either probably doesn’t help.

Adding the hint makes the warning much less "well, that’s kinda obvious" to me. Hopefully we can get some natives in!

src/borg/archiver.py
```diff
@@ -306,8 +306,11 @@ def do_init(self, args, repository):
     def do_check(self, args, repository):
         """Check repository consistency"""
         if args.repair:
-            msg = ("'check --repair' is an experimental feature that might result in data loss." +
-                   "\n" +
+            msg = ("This is a potentially dangerous functionality.\n"
```
d7415, line 309: I'd say `This is a potentially dangerous function.`
```suggestion
            msg = ("This is a potentially dangerous function.\n"
```

src/borg/archiver.py
elho: That "in subtle ways" makes the warning sound rather strange. Plus I would not call an `--exclude` that excludes everything subtle. The fact that `--exclude` excludes stuff is pretty obvious, its just that your pattern may not work as expected, much like with `prune`.

elho:
>     * there could be an undiscovered bug in the code

Well, either the verdict is, that there is a higher likeliness of bugs in that particular code, which would be an argument to keep the experimental warning, or that there just like in every other place of the code could be hidden bugs (like e.g. the pre-1.1.11 index corruption), which is something that always applies to any sofware and does not need any waring, ie. no "or in case of bugs".

>     * the user might use them wrong (e.g. accidentally excluding everything /
>       not matching anything when recreating an archive)

The user may also `delete` the wrong archive or `prune` everything, still they do not generally warn about them being used at all. `delete` does warn when trying to dele the whole repository. That both is good.

So I'd rather have recreate not warn, or only warn if used with `--exclude` (but not `--target`). There is no reason to warn when just adding/chaning an archive comment, recompressing or any other operation that is foolproof to use.

DurvalMenezes:
> The user may also delete the wrong archive or prune everything, still they do not generally warn about them being used at all. delete does warn when trying to dele the whole repository. That both is good.
> 
> So I'd rather have recreate not warn, or only warn if used with --exclude (but not --target). There is no reason to warn when just adding/cha[i]ning an archive comment, recompressing or any other operation that is foolproof to use.

FWIW, I agree with @elho: it's not (or it should not be) the job of `borg` to protect users from themselves. And the documentation does a good enough job of that already, quoting from https://borgbackup.readthedocs.io/en/stable/usage/recreate.html:
'''
USE WITH CAUTION. Depending on the PATHs and patterns given, recreate can be used to permanently delete files from archives. When in doubt, use --dry-run --verbose --list to see how patterns/PATHS are interpreted.
'''

src/borg/archiver.py
ThomasWaldmann: fixed

src/borg/archiver.py
ThomasWaldmann: done

ThomasWaldmann:
OK, I removed the talk about bugs.

I'ld like to keep the "dangerous" warning for `borg check --repair`. It's because here the user expectation definitely is that things will improve (it's a "repair" command, not a "kill" command). But we can not assure that, there might be corruptions where check --repair makes things worse. If the user is warned and the repo is important, it is the user to decide whether to make a repo copy first before trying to repair it.

About the "dangerous" warning for `borg recreate`: there is only the danger of bad usage left (which is of course the users fault). For prune there is a similar problem and we only warn about it in the docs, but not at runtime and also do not require an extra confirmation. We could do it in the same way for recreate, which would mean also to remove the I_KNOW_WHAT_I_AM_DOING override env var for recreate.

ThomasWaldmann:
squash / merge?

PR #5247, opened 2020-07-07
use allow/deny list wording

ThomasWaldmann:
... and get rid of problematic blacklist/whitelist wording.


ThomasWaldmann:
It might make sense to backport this to keep 1.1-maint and master in sync (to ease backporting future changes in that part of the code).

PR #5263, opened 2020-07-25
persist shadow_index in between borg runs, fixes #4830

### LINKED ISSUE #4830: master branch: check shadow_index for separate compact_segments()
Seems like the 1.1-maint code builds the `shadow_index` while doing a (borg create) transaction and then uses that in `compact_segments()` (which is done in the same borg run in 1.1).

In master, compact_segments was split into a separate command and usually not done within borg create, but later in a separate borg invoaction. Check what happens with the shadow_index data - is it lost?

ThomasWaldmann:
in borg 1.1, compact_segments() was always run directly after some repo writing operation (in same borg process).

but now, only "borg compact" is used to compact segments and it is a separate borg invocation (new process), so we need to persist the shadow_index so we do not lose that information.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -160,6 +160,7 @@ class Repository:
         # This is an index of shadowed log entries during this transaction. Consider the following sequence:
         # segment_n PUT A, segment_x DELETE A
         # After the "DELETE A" in segment_x the shadow index will contain "A -> [n]".
+        # .delete() is updating this index, it is persisted into "hints" file and is later used by .compact_segments().
         self.shadow_index = {}
         self._active_txn = False
         self.lock_wait = lock_wait
@@ -570,6 +571,7 @@ class Repository:
                 self.segments = hints[b'segments']
                 self.compact = FreeSpace()
                 self.storage_quota_use = 0
+                self.shadow_index = {}
                 for segment in sorted(hints[b'compact']):
                     logger.debug('Rebuilding sparse info for segment %d', segment)
                     self._rebuild_sparse(segment)
@@ -580,6 +582,7 @@ class Repository:
                 self.segments = hints[b'segments']
                 self.compact = FreeSpace(hints[b'compact'])
                 self.storage_quota_use = hints.get(b'storage_quota_use', 0)
+                self.shadow_index = hints.get(b'shadow_index', {})
             self.log_storage_quota()
             # Drop uncommitted segments in the shadow index
             for key, shadowed_segments in self.shadow_index.items():
@@ -600,6 +603,7 @@ class Repository:
             b'segments': self.segments,
             b'compact': self.compact,
             b'storage_quota_use': self.storage_quota_use,
+            b'shadow_index': self.shadow_index,
         }
         integrity = {
             # Integrity version started at 2, the current hints version.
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
will check if i can add a test for that somehow, but at least i hope that i've analyzed #4830 correctly.

@enkore can you have a look. your code looked a bit like it was already trying to use what I have added right now, see the end of `prepare_txn` (i wondered where shadow_index comes from when it is used there).

ThomasWaldmann:
Some tests fail (not the new ones, but some we had for long):
```
___________ RepositoryCommitTestCase.test_moved_deletes_are_tracked ____________
[gw0] linux -- Python 3.6.3 /home/travis/build/borgbackup/borg/.tox/py36/bin/python
self = <borg.testsuite.repository.RepositoryCommitTestCase testMethod=test_moved_deletes_are_tracked>
    def test_moved_deletes_are_tracked(self):
        self.repository.put(H(1), b'1')
        self.repository.put(H(2), b'2')
        self.repository.commit(compact=False)
        self.repository.delete(H(1))
        self.repository.commit(compact=True)
        last_segment = self.repository.io.get_latest_segment() - 1
        num_deletes = 0
        for tag, key, offset, size in self.repository.io.iter_objects(last_segment):
            if tag == TAG_DELETE:
                assert key == H(1)
                num_deletes += 1
        assert num_deletes == 1
        assert last_segment in self.repository.compact
        self.repository.put(H(3), b'3')
        self.repository.commit(compact=True)
        assert last_segment not in self.repository.compact
        assert not self.repository.io.segment_exists(last_segment)
        for segment, _ in self.repository.io.segment_iterator():
            for tag, key, offset, size in self.repository.io.iter_objects(segment):
>               assert tag != TAG_DELETE
E               AssertionError: assert 1 != 1
.tox/py36/lib/python3.6/site-packages/borg/testsuite/repository.py:334: AssertionError
------------------------------ Captured log call -------------------------------
INFO     borg.repository:logger.py:193 compaction freed about 101 B repository space.
INFO     borg.repository:logger.py:193 compaction freed about 59 B repository space.
_____________ RepositoryCommitTestCase.test_shadow_index_rollback ______________
[gw0] linux -- Python 3.6.3 /home/travis/build/borgbackup/borg/.tox/py36/bin/python
self = <borg.testsuite.repository.RepositoryCommitTestCase testMethod=test_shadow_index_rollback>
    def test_shadow_index_rollback(self):
        self.repository.put(H(1), b'1')
        self.repository.delete(H(1))
        assert self.repository.shadow_index[H(1)] == [0]
        self.repository.commit(compact=True)
        # note how an empty list means that nothing is shadowed for sure
>       assert self.repository.shadow_index[H(1)] == []
E       KeyError: b'00000000000000000000000000000001'
.tox/py36/lib/python3.6/site-packages/borg/testsuite/repository.py:376: KeyError
```

ThomasWaldmann:
The lower one is because empty segment lists are now removed from shadow_index.

ThomasWaldmann:
I reverted removing the empty segments lists from shadow_index because the code relies on them being there.

Guess they should be removed at some time though and should not stay in there forever?

enkore:
Overall this looks correct to me.

> Guess they should be removed at some time though and should not stay in there forever?

Yeah. In general the repository unit tests are all super-low-level and kind of a headache to deal with; I pretty much just grabbed a pen and paper half the time to convince myself that the new behaviour, which caused a test failure, is correct.

PR #5267, opened 2020-07-25
docs: add a note to create from stdin regarding files cache, fixes #5180

### LINKED ISSUE #5180: [Question] borg create from stdin and caching
I was not able to found clarification or any description of this scenarion in documentation:

How metadata caching is working in case of creating snapshot from stdin pipe? Example:

`mysqldump ... | borg create $REPOSITORY::mysql-$now -`

Is there any sense to add option `--files-cache=ctime,size` in this case? It's not file, so may be better to add `--no-cache-sync` or `--no-files-cache`? Will deduplication be working in this case?

btw, does `--files-cache=ctime,size` doing the same as `--ignore-inode` ?

fantasya-pbem:


ThomasWaldmann:
thanks!

PR #5280, opened 2020-08-02
[Docs] Notes about borg cache workaround

viq:
Add a note about cache ballooning up on cache recreation, and that the workaround may be removed once cache is already in place.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -734,6 +734,17 @@ connections and release the lock).
 The borg cache eats way too much disk space, what can I do?
 -----------------------------------------------------------
 
+This may especially happen if borg needs to rebuild the local "chunks" index -
+either because it was removed, or because it was not coherent with the
+repository state any more (e.g. because another borg instance changed the
+repository).
+
+To optimize this rebuild process, borg caches per-archive information in the
+``chunks.archive.d/`` directory. It won't help the first time it happens, but it
+will make the subsequent rebuilds faster (because it needs to transfer less data
+from the repository). While being faster, the cache needs quite some disk space,
+which might be unwanted.
+
 There is a temporary (but maybe long lived) hack to avoid using lots of disk
 space for chunks.archive.d (see :issue:`235` for details):
 


docs/faq.rst
```diff
@@ -758,6 +763,9 @@ This has some pros and cons, though:
   machine (if you share same backup repo between multiple machines) or if
   your local chunks cache was lost somehow.
 
+Above workaround is necessary only on the run when cache is being recreated,
+and once it completes may be reverted, with borg cache behaving as previously.
+
```
ThomasWaldmann, line 768: Not quite.

If you do that workaround at any time, it afterwards will never use a lot of disk space because it will never create the chunks.archives.d/ cache.

viq: I made an experiment, and I blew away cache of all three repositories I have for one of the hosts, and also removed the contents of one of those repositories. I created the removed one anew, for the two remaining repos I started a borg operation that would create the cache (I think it was a `check`), stopped it with `^C`, applied the workaround in the cache directory by removing the `chunks.archive.d` and making it a file as described. I then proceeded to use all the repositories as normal. After a while, in both cache directories, I removed the file `chunks.archive.d` and made it into a directory. With normal usage of borg that directory remained empty.

That's what I'm trying to say by "workaround is necessary when cache is recreated and can be removed afterwards". 

I don't know in what situations `chunks.archive.d/` directory may be used, thus me trying to put it in simple terms that would relate my experience.

viq: I've updated my merge request to talk more about this.

docs/faq.rst
```diff
@@ -734,6 +734,12 @@ connections and release the lock).
...
 -----------------------------------------------------------
 
+This may especially happen if borg needs to recreate the local cache based 
+on repository data, either because local cache was removed, or because of multiple
+machines writing to same repository. Borg uses a smaller cache if the repo is never accessed
+from a different cache, but on rebuild switches to a "format" that is easier
+to update but much bigger.
+
 There is a temporary (but maybe long lived) hack to avoid using lots of disk
 space for chunks.archive.d (see :issue:`235` for details):
```
ThomasWaldmann, line 744: did you read issue 235?

viq: Read, yes, but I don't think I was able to understand it properly :(

ThomasWaldmann: How about:

This may especially happen if borg needs to rebuild the local "chunks" index - either because it was removed, or because it was not coherent with the repository state any more (e.g. because another borg changed the repository).

To optimize this rebuild process, borg caches per-archive information in the "chunks.archive.d/" directory.
It won't be faster the first time that happens, but on subsequent rebuilds (because it needs to transfer less data from the repository). While being faster, the cache needs quite some disk space, which might be unwanted.

docs/faq.rst
```diff
@@ -758,6 +764,13 @@ This has some pros and cons, though:
   machine (if you share same backup repo between multiple machines) or if
   your local chunks cache was lost somehow.
 
+Above workaround is necessary only on the run when cache is being recreated,
+and once it completes may be reverted, with borg cache behaving as previously.
+Which means that if you don't want this behaviour, it's enough to have it in place
+only on first run if the cache got removed, but you'll need to keep it there in case
+of multiple machines writing to same repo, since in that scenario cache will be
+constantly getting invalidated.
```
ThomasWaldmann, line 772: and just remove that text again, i guess it is rather confusing than helping.

either one is wanting more speed and accepts high space usage.

or one does not, then one needs to apply the described hack at any time (and needs to keep it).

docs/faq.rst
viq: Mmm, nice. How about slightly reworded?

This may especially happen if borg needs to rebuild the local "chunks" index -
either because it was removed, or because it was not coherent with the
repository state any more (e.g. because another borg instance changed the
repository).

To optimize this rebuild process, borg caches per-archive information in the
``chunks.archive.d/`` directory. It won't help the first time it happens, but it
will make the subsequent rebuilds faster (because it needs to transfer less data
from the repository). While being faster, the cache needs quite some disk space,
which might be unwanted.

docs/faq.rst
viq: Yeah, with the above explanation it's clear when it's created and what it's used for, so this is superfluous.

docs/faq.rst
ThomasWaldmann: ok

viq:
I've squashed commits to include just the updated text.

ThomasWaldmann:
Thanks!

ThomasWaldmann:
Do you want to make a PR for a 1.1-maint backport?

viq:
I can try, though I've never really done that before.

viq:
https://github.com/borgbackup/borg/pull/5283 is my attempt at it.

PR #5283, opened 2020-08-04
Better description of how cache and rebuilds of it work

viq:
and how the workaround applies to that.

https://github.com/borgbackup/borg/pull/5280 for `1.1-maint` branch


ThomasWaldmann:
Thanks!

PR #5290, opened 2020-08-09
docs: Add FAQ discussing choice of encryption mode, fixes #5285

### LINKED ISSUE #5285: Encryption advice: Safety repokey vs. keyfile / reuse key or key password
This is a question. I'm planning a backup strategy based on borg. I have to backup two different machines to three different locations. As I consider the destinations untrusted, I want to encrypt my backup. This comes with the downside of having to backup the passwords / keys / location information in a seperate location, as these might get lost when my machine breaks. For backing up these information, it would be conventient to have as little as possible "pieces" to backup. This is where my questions arise. I did not find good answers to these questions (if my search strategy failed, I apologize).

1. I can choose between repokey, keyfile with password and keyfile without password. From an encryption perspective, can any of these be considered "safer" (when choosing and adequately safe password)?
2. To reduce the amount of passwords I have to store, one option could be reusing the same password for multiple keys (keyfile or repokey mode), or even across different machines (i.e. one password for all 6 keys (2 machines x 3 locations) up to 6 passwords for 6 keys). I feel that this option leads to security issues, but did not find a reliable source for this. Can you comment on this?

Yes

Question

Client: MacBook
Server: Synology, Hetzner Storage Box

1.1.13

macOS

fantasya-pbem:
This is a first draft, change requests welcome.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -606,6 +606,43 @@ into the repository.
 Yes, as an attacker with access to the remote server could delete (or
 otherwise make unavailable) all your backups.
 
+Which encryption mode is the safest / should I choose?
+------------------------------------------------------
+
+This question is impossible to answer in a simple way, because there is a bunch
+of different aspects that need to be considered. Availability, confidentiality
+and integrity take influence on how complex or comfortable your backup process
+can be designed.
+
+1. Do you want full automation?
+
+If you choose password protection for encryption keys, you will have to store
+the plain password somewhere for fully automated backups. If you have several
+backup repositories, it is generally safer to have individual passwords.
+
+2. Do you trust the backup server?
+
+In repokey mode the key is stored in the repo and could be stolen by an
+attacker. In keyfile mode the key never leaves the backup client.
+
+3. Do you trust the backup client?
+
+In keyfile mode the key is stored on the client and could be stolen by other
+users. Password protection makes stealing the key useless.
+
+4. Can you afford to loose the backup repo?
+
+Can you restore a corrupted or lost repo? If not, there is no point in storing
+the keyfile on the client separate from the repo.
+
+5. Do you mind securing your keyfile?
+
+If you loose your keyfile, the backup repo is lost. You may not want to store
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -606,6 +606,43 @@ into the repository.
...
+users. Password protection makes stealing the key useless.
+
+4. Can you afford to loose the backup repo?
+
+Can you restore a corrupted or lost repo? If not, there is no point in storing
+the keyfile on the client separate from the repo.
+
+5. Do you mind securing your keyfile?
+
+If you loose your keyfile, the backup repo is lost. You may not want to store
```
ThomasWaldmann, line 640: s/lost/unusable/

docs/faq.rst
```diff
@@ -606,6 +606,43 @@ into the repository.
...
+copies of the key at other places and choose repokey mode.
```
ThomasWaldmann, line 641: I'ld rather rephrase that:

A repokey is safer than keyfile in "total loss" scenarios:
- repokey is stored inside the repo. if you totally lose your repo, you won't need the key for it any more also.
- keyfile is stored on the client. if you totally lose your client's filesystem (and you do not have a key backup), you lose access to your repo just at the time when you need it most!

There are other failure scenarios besides "total loss" of course, so having a backup of your keys is always strongly recommended.

docs/faq.rst
```diff
@@ -606,6 +606,43 @@ into the repository.
...
+
+The safest choice may be password-protected keyfile mode running interactively
+and storing the keyfile on USB stick.
```
ThomasWaldmann, line 644: I'ld never ever recommend usb sticks.

People lose them, mechanically break them, ESD destroy them, they die by themselves also, sometimes have rather low quality, get put in all sorts of own and foreign computers, ...

So rather recommend "store a key backup at a safe place" (guess we already have that recommendation at another place in the docs).

ThomasWaldmann:
@deermo `borg init` always tells you to back up your keys.

PR #5302, opened 2020-08-16
Build directory-based binary for macOS to avoid Gatekeeper delays.

### LINKED ISSUE #5185: Brew cask running really slow on OSX
hello there, I installed borgbackup via brew cask and its running really slow.

`borg help  0.88s user 0.25s system 8% cpu 13.565 total`

I am running OSX 10.15.4 Catalina. I also tried installing from source and pip3 which also leads to the same problem. Is there an easy fix to this and is anybody else experiencing similar issues?

Thanks a lot!

m3nu:
Directory-style binaries avoid known PyInstaller issues with newer macOS versions: When using a single-file PyInstaller binary on macOS, the required libraries are extracted and individually checked by macOS. This can take 10 to 20 seconds.

A directory-style binary avoids these issues, but isn't as user-friendly, since the user gets to see all the additional library files.

This PR adjusts the PyInstaller spec file to build a dir-style bin for macOS and compresses it as tar file. Users would need to download this file and link `borg.exe` to their `PATH`.

Fixes #5185

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -86,23 +86,15 @@ def packages_darwin
     # install all the (security and other) updates
     sudo softwareupdate --ignore iTunesX
     sudo softwareupdate --ignore iTunes
+    sudo softwareupdate --ignore Safari
     sudo softwareupdate --ignore "Install macOS High Sierra"
     sudo softwareupdate --install --all
-    # get osxfuse 3.x release code from github:
-    curl -s -L https://github.com/osxfuse/osxfuse/releases/download/osxfuse-3.10.4/osxfuse-3.10.4.dmg >osxfuse.dmg
-    MOUNTDIR=$(echo `hdiutil mount osxfuse.dmg | tail -1 | awk '{$1="" ; print $0}'` | xargs -0 echo) \
-    && sudo installer -pkg "${MOUNTDIR}/Extras/FUSE for macOS 3.10.4.pkg" -target /
-    sudo chown -R vagrant /usr/local  # brew must be able to create stuff here
-    which brew || ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    which brew || CI=1 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
     brew update > /dev/null
-    brew install pkg-config || brew upgrade pkg-config
-    brew install readline || brew upgrade readline
-    brew install openssl@1.1 || brew upgrade openssl@1.1
-    brew install zstd || brew upgrade zstd
-    brew install lz4 || brew upgrade lz4
-    brew install xz || brew upgrade xz  # required for python lzma module
-    brew install fakeroot || brew upgrade fakeroot
-    brew install git || brew upgrade git
+    brew install pkg-config readline openssl@1.1 zstd lz4 xz fakeroot git
+    brew tap homebrew/cask
+    brew cask install osxfuse
+    brew upgrade  # upgrade everything
     echo 'export PKG_CONFIG_PATH=/usr/local/opt/openssl@1.1/lib/pkgconfig' >> ~vagrant/.bash_profile
   EOF
 end
@@ -210,6 +202,7 @@ def build_binary_with_pyinstaller(boxname)
     cd borg
     pyinstaller --clean --distpath=/vagrant/borg scripts/borg.exe.spec
     echo 'export PATH="/vagrant/borg:$PATH"' >> ~/.bash_profile
+    cd .. && tar -czvf borg.tgz borg-dir
   EOF
 end
 
@@ -251,7 +244,7 @@ end
 
 Vagrant.configure(2) do |config|
   # use rsync to copy content to the folder
-  config.vm.synced_folder ".", "/vagrant/borg/borg", :type => "rsync", :rsync__args => ["--verbose", "--archive", "--delete", "-z"], :rsync__chown => false
+  config.vm.synced_folder ".", "/vagrant/borg/borg", :type => "rsync", :rsync__args => ["--verbose", "--archive", "--delete", "--exclude", ".python-version"], :rsync__chown => false
   # do not let the VM access . on the host machine via the default shared folder!
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
@@ -362,8 +355,9 @@ Vagrant.configure(2) do |config|
     b.vm.box = "macos-sierra"
     b.vm.provider :virtualbox do |v|
       v.memory = 2048 + $wmem
-      v.customize ['modifyvm', :id, '--ostype', 'MacOS1012_64']
+      v.customize ['modifyvm', :id, '--ostype', 'MacOS_64']
       v.customize ['modifyvm', :id, '--paravirtprovider', 'default']
+      v.customize ['modifyvm', :id, '--nested-hw-virt', 'on']
       # Adjust CPU settings according to
       # https://github.com/geerlingguy/macos-virtualbox-vm
       v.customize ['modifyvm', :id, '--cpuidset',
--- a/scripts/borg.exe.spec
+++ b/scripts/borg.exe.spec
@@ -51,15 +51,24 @@ exe = EXE(pyz,
           upx=True,
           console=True)
 
-if False:
-    # Enable this block to build a directory-based binary instead of
-    # a packed single file. This allows one to easily look at all included
-    # files (e.g. without having to strace or halt the built binary
-    # and introspect /tmp).
-    coll = COLLECT(exe,
-                   a.binaries,
-                   a.zipfiles,
-                   a.datas,
-                   strip=False,
-                   upx=True,
-                   name='borg-dir')
+# Build a directory-based binary in addition to a packed
+# single file. This allows one to easily look at all included
+# files (e.g. without having to strace or halt the built binary
+# and introspect /tmp). Also avoids unpacking all libs when
+# running the app, which is better for app signing on various OS.
+slim_exe = EXE(pyz,
+            a.scripts,
+            exclude_binaries=True,
+            name='borg.exe',
+            debug=False,
+            strip=False,
+            upx=False,
+            console=True)
+
+coll = COLLECT(slim_exe,
+                a.binaries,
+                a.zipfiles,
+                a.datas,
+                strip=False,
+                upx=False,
+                name='borg-dir')


scripts/borg.exe.spec
```diff
@@ -33,33 +33,53 @@ a = Analysis([os.path.join(basepath, 'src', 'borg', '__main__.py'), ],
              win_private_assemblies=False,
              cipher=block_cipher)
 
+pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
+
+# For macOS we build a directory-based binary to get around repeated
+# lib validations. See borgbackup/borg#5185 and
```
ThomasWaldmann, line 39: and ... ?

scripts/borg.exe.spec
```diff
@@ -33,33 +33,53 @@ a = Analysis([os.path.join(basepath, 'src', 'borg', '__main__.py'), ],
...
-                   a.datas,
-                   strip=False,
-                   upx=True,
-                   name='borg-dir')
+                a.binaries,
+                a.zipfiles,
+                a.datas,
+                strip=False,
+                upx=False,
+                name='borg-macosx64')
```
ThomasWaldmann, line 60: guess the name could be computed in advance for all platforms using sys.platform.
also we could do the single-file and the directory approach for all platforms.

the pyinstaller run is relatively quick, considering the platform testing runs for hours on some platforms.

ThomasWaldmann: or, maybe even simpler, use `borg-dir` here also (see non-darwin64 code).

Vagrantfile
```diff
@@ -209,6 +209,7 @@ def build_binary_with_pyinstaller(boxname)
     . borg-env/bin/activate
     cd borg
     pyinstaller --clean --distpath=/vagrant/borg scripts/borg.exe.spec
+    tar -czvf borg-macosx64.tar.gz --strip-components=2 dist/borg-macosx64
```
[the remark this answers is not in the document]
ThomasWaldmann: maybe use `borg-dir` here, so it works for all platforms? see also comments below.

Vagrantfile
```diff
@@ -86,23 +86,15 @@ def packages_darwin
...
     brew update > /dev/null
-    brew install pkg-config || brew upgrade pkg-config
-    brew install readline || brew upgrade readline
-    brew install openssl@1.1 || brew upgrade openssl@1.1
-    brew install zstd || brew upgrade zstd
-    brew install lz4 || brew upgrade lz4
-    brew install xz || brew upgrade xz  # required for python lzma module
-    brew install fakeroot || brew upgrade fakeroot
-    brew install git || brew upgrade git
+    brew install pkg-config readline openssl@1.1 zstd lz4 xz fakeroot git
```
ThomasWaldmann, line 94: doesn't this fail in case some of them are already installed and thus the form `brew install x | brew upgrade x` is needed?

Vagrantfile
```diff
@@ -86,23 +86,15 @@ def packages_darwin
...
+    brew tap homebrew/cask
+    brew cask install osxfuse
```
ThomasWaldmann, line 96: guess we lose a bit of control here compared to directly installing some specific osxfuse version.

otoh, as long as it works, it does maybe not matter. i won't be able to tell in the README for the binaries "was built using osxfuse X.X" then.

Vagrantfile
```diff
@@ -86,23 +86,15 @@ def packages_darwin
     # install all the (security and other) updates
     sudo softwareupdate --ignore iTunesX
     sudo softwareupdate --ignore iTunes
-    sudo softwareupdate --ignore "Install macOS High Sierra"
-    sudo softwareupdate --install --all
-    # get osxfuse 3.x release code from github:
-    curl -s -L https://github.com/osxfuse/osxfuse/releases/download/osxfuse-3.10.4/osxfuse-3.10.4.dmg >osxfuse.dmg
-    MOUNTDIR=$(echo `hdiutil mount osxfuse.dmg | tail -1 | awk '{$1="" ; print $0}'` | xargs -0 echo) \
-    && sudo installer -pkg "${MOUNTDIR}/Extras/FUSE for macOS 3.10.4.pkg" -target /
-    sudo chown -R vagrant /usr/local  # brew must be able to create stuff here
```
ThomasWaldmann, line 95: so dir owner / perms were fixed at another place?

Vagrantfile
```diff
@@ -86,23 +86,15 @@ def packages_darwin
...
-    which brew || ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    sudo softwareupdate --ignore Safari
+    #sudo softwareupdate --ignore "Install macOS Catalina"
+    #sudo softwareupdate --install --all
+    which brew || CI=1 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
ThomasWaldmann, line 92: CI=1 ?

Vagrantfile
```diff
@@ -360,11 +351,12 @@ Vagrant.configure(2) do |config|
...
 
   config.vm.define "darwin64" do |b|
-    b.vm.box = "jhcook/macos-sierra"
+    b.vm.box = "yzgyyang/macOS-10.14"
     b.vm.provider :virtualbox do |v|
       v.memory = 2048 + $wmem
-      v.customize ['modifyvm', :id, '--ostype', 'MacOS1012_64']
+      v.customize ['modifyvm', :id, '--ostype', 'MacOS_64']
       v.customize ['modifyvm', :id, '--paravirtprovider', 'default']
+      v.customize ['modifyvm', :id, '--nested-hw-virt', 'on']
```
ThomasWaldmann, line 359: is this needed? for what?

does that make problems on older cpus?

Vagrantfile
```diff
@@ -380,7 +372,7 @@ Vagrant.configure(2) do |config|
     b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("darwin64")
     b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg(true)
     b.vm.provision "install pyinstaller", :type => :shell, :privileged => false, :inline => install_pyinstaller()
-    b.vm.provision "build binary with pyinstaller", :type => :shell, :privileged => false, :inline => build_binary_with_pyinstaller("darwin64")
+    b.vm.provision "build binary with pyinstaller", :type => :shell, :run =>"always", :privileged => false, :inline => build_binary_with_pyinstaller("darwin64")
```
ThomasWaldmann, line 375: why?

scripts/borg.exe.spec
```diff
@@ -57,29 +72,4 @@ if sys.platform == 'darwin':
...
-        # and introspect /tmp).
-        coll = COLLECT(exe,
-                    a.binaries,
-                    a.zipfiles,
-                    a.datas,
-                    strip=False,
-                    upx=True,
-                    name='borg-dir')
-
+                name='borg-dir')
```
ThomasWaldmann, line 75: nitpick: add a \n here so diff does not complain

scripts/borg.exe.spec
```diff
@@ -51,15 +51,25 @@ exe = EXE(pyz,
           upx=True,
           console=True)
 
-if False:
+if os.environ.get('PYINSTALLER_DIR_BIN', False):
```
ThomasWaldmann, line 54: can we just always build the exe and the dir binary?

iirc the tests check for presence of borg.exe and if it is there, the tests will be invoked using that exe.

so guess we should always have the exe (and maybe just hope that the dir-based binary will also work if the exe works).

m3nu:
The VM provided by @ThomasWaldmann works well. Not sure what's the issue with public Vagrant boxes.

I'll proceed to clean up this PR to:

- [ ] Simplify some Homebrew stuff that changed over time
- [ ] Build a single-file and single-dir release for macOS to avoid Gatekeeper delays. The latter is a bit complicated to install, so Homebrew is probably better, but I'll use it as official fallback in Vorta.
- [ ] Let's call your Vagrant VM `waldmann/macos-sierra`, so it's clear where it came from?

Anything else that would need fixing, testing or improving in this context, @ThomasWaldmann ?

ThomasWaldmann:
Don't change the VM name please.

m3nu:
Alright. Will see to finish the rest then.

Vagrantfile
m3nu: Not an issue. Just causes a warning, but the rest is installed.

`Warning: git 2.28.0 is already installed and up-to-date`

m3nu:
Ran the VM from start to finish. Looks OK. Comparison of single-file vs. single-dir:

First run: ~15 seconds for both
```shell
$ time ./borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg.exe --version  0.49s user 0.22s system 4% cpu 15.599 total
$ time ./borg-dir/borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg-dir/borg.exe --version  0.41s user 0.08s system 3% cpu 15.078 total
```

Subsequent runs: near-zero for single-dir and 15 to 30 seconds for single-file
```shell
$ time ./borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg.exe --version  0.45s user 0.21s system 4% cpu 14.045 total
$ time ./borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg.exe --version  0.51s user 0.29s system 2% cpu 27.026 total

$ time ./borg-dir/borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg-dir/borg.exe --version  0.42s user 0.09s system 83% cpu 0.615 total
$ time ./borg-dir/borg.exe --version
borg.exe 1.2.0a9.dev68+gce2470a2.d20200829
./borg-dir/borg.exe --version  0.44s user 0.10s system 61% cpu 0.865 total
```

Vagrantfile
m3nu: I assume Homebrew takes care of it. The permissions are correct in the end.

```
$ ls -l /usr/ | grep local
drwxr-xr-x    15 vagrant  wheel    510 Apr 19 13:24 local
```

Vagrantfile
m3nu: This should reduce interaction and output by pretending to be a CI bot, according to https://discourse.brew.sh/t/silent-automated-homebrew-install/3180

Vagrantfile
m3nu: We could install a specific version by pointing it to an older Homebrew commit. But I'd rather avoid that.

Vagrantfile
m3nu: `MacOS_64` worked for me for all future versions as well. The version-specific OS settings in Virtualbox seem to stop after 10.13. Since `MacOS_64` works with old *and* new versions, I'd say it's more future-proof and no settings need to be changed if the VM is updated at some point.

Vagrantfile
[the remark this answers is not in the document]
m3nu: Removed this again, so it can be done outside the VM.

> maybe use borg-dir here, so it works for all platforms? see also comments below.
It's not needed for any other platform and it's very confusing for the user to get a directory with 20 library files. I wouldn't publish it for every platform.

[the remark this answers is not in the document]
ThomasWaldmann: iirc we had also cases on other platforms. iirc when wanting to use capabilities on linux, one can only work with an executable, not a script. and some people also complained about unpacking time on linux.

so i think we should offer this for all platforms and let people choose what they want.

best would be a single tar file that includes both builds, so I do not need to sign and upload more files than currently.

Vagrantfile
ThomasWaldmann: ah, ok

Vagrantfile
ThomasWaldmann: and does "brew install" also upgrade if an old version is installed?

Vagrantfile
ThomasWaldmann: ok

Vagrantfile
ThomasWaldmann: ok

Vagrantfile
m3nu: You're right, it wouldn't upgrade it. Question is if that's relevant, since the VM would be fresh and the install only runs once. Or how do you usually re-run Vagrant for new releases?

scripts/borg.exe.spec
m3nu: And what? More intro text?

Vagrantfile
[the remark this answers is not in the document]
m3nu: Then we build both for everyone. Just needs a small adjustment to the PyInstaller spec file.

Vagrantfile
m3nu: Forgot. This line further down will upgrade everything. So the `brew install xx || brew upgrade xx` isn't needed.

`brew upgrade  # upgrade everything`

m3nu:
Adressed the remaining issues. You can now do

- `vagrant scp boxname:/vagrant/borg.exe borg-exe-OSNAME`
- `vagrant scp boxname:/vagrant/borg.tgz borg-dir-OSNAME.tgz`

for all systems.

scripts/borg.exe.spec
ThomasWaldmann: I mean that line 39 ending with "and" - there seems to be something missing.

scripts/borg.exe.spec
```diff
@@ -51,15 +51,24 @@ exe = EXE(pyz,
...
-                   a.datas,
-                   strip=False,
-                   upx=True,
-                   name='borg-dir')
+# Build a directory-based binary in addition to a packed
+# single file. This allows one to easily look at all included
+# files (e.g. without having to strace or halt the built binary
+# and introspect /tmp). Also avoids unpacking all libs when
+# running the app, which is better for app signing on various OS.
+slim_exe = EXE(pyz,
```
ThomasWaldmann, line 59: duplicate of line 43+?

ThomasWaldmann:
@m3nu did you see my last comments?

m3nu:
Only now. 😬

scripts/borg.exe.spec
m3nu: No. The binary in line 43 includes all libs and data. This one doesn't.

m3nu:
All responded to. I'll also check my Github notification settings.

ThomasWaldmann:
@m3nu My impression is that github notifications are not really reliable.

scripts/borg.exe.spec
ThomasWaldmann: looks like that change is not present any more in this PR, so it is resolved.

scripts/borg.exe.spec
ThomasWaldmann: thanks for fixing.

ThomasWaldmann:
Merged. \o/

I'll give this a test run on my vagrant host later today

PR #5304, opened 2020-08-18
Update Homebrew install instructions. See #5185

m3nu:
With the new Homebrew [formula](https://formulae.brew.sh/formula/borgbackup) merged, some minor updates to the install instructions were needed.


ThomasWaldmann:
Thanks! Guess we need this in 1.1-maint also.

m3nu:
Sure. Will cherry-pick it into 1.1-maint.

They just removed the previous Cask. So the old commands won't work any more starting today.

PR #5309, opened 2020-08-25
Update prune docs to mention double force

voron:
SSIA


ThomasWaldmann:
forcing (and double forcing) was added as a `borg delete` feature.

But it sounds like you tried it and it also worked for `borg prune`?

Can well be, as prune is basically just a wrapper calling delete internally.

voron:
>But it sounds like you tried it and it also worked for borg prune?

Correct, here is the output I got from borg 1.1.13 `prune --force --force`
```
borg -v -p prune --list  --keep-last 25 --force --force ssh://...
Using a pure-python msgpack! This will result in lower performance.
Keeping archive: 2020-08-25_06:39                     Tue, 2020-08-25 07:04:41 [14b00a52d8d08882bf457d5c045c1f331a54c7431ed38838816728da2fa97c43]
...
Keeping archive: 2020-08-01_06:39                     Sat, 2020-08-01 07:06:08 [c07f5c1d7bb86f0c7e6c0d582bb75d77c3d0fa479d90685f95a7c98a704835fb]
Pruning archive: 2020-07-31_06:39                     Fri, 2020-07-31 07:04:05 [99a48ff4917bf8c75b40ab80f6c4342afc6413650a095e5b4b9dc5dc357e412e] (1/8)
...
Pruning archive: 2020-07-05_06:39                     Sun, 2020-07-05 06:57:48 [f8f3212b2b568d81e7f8d9d051aa22595f4250213b904b9a9edecc487dba94b0] (8/8)
forced deletion succeeded, but the deleted archive was corrupted.
borg check --repair is required to free all space.
```
During `prune` with single `--force` I got a fatal error
```
Chunk ID 6c0a8d34dd4b688b17dcf397438e274e65f6f0f9234f8f0f48a45746cdf95ec7 missing from chunks index, corrupted chunks index - aborting transaction.
```

ThomasWaldmann:
Looks like there are multiple problems:

- your repo is corrupted. tried `borg check --repair`? also read the top of the changelog, there are some notes about corruption issues and how to deal with them (if they were caused by borg bugs, not be defective hw or filesystem or ...).

- it complains about pure-python msgpack, so something went wrong when installing that. it should work, but it is slower.

ThomasWaldmann:
guess 1.1-maint branch should get the same change.

ThomasWaldmann:
thanks for the docs update!

voron:
>tried borg check --repair? 

Seems, corruption affects outdated (already removed) archives only. `borg check --last 2` is present in backup scripts with fail propagation. I will look into this repo more carefully, thank you.

>it complains about pure-python msgpack, so something went wrong when installing that. it should work, but it is slower.

Binary file from github on centos 7 with glibc-2.17, seems to be known issue. Let's see the actual slowdown, I have output logged for ~1 month of this backup with specified by `borg --stats` duration.

ThomasWaldmann:
```
.. IMPORTANT: this file is auto-generated from borg's built-in help, do not edit!
```

Looks like this was overlooked...

I'll fix it at the right place...

PR #5322, opened 2020-09-03
Changing rephrase explanation

Nono-m0le:
Somehow, I find it more clear when the commands are on several lines (instead of one).
Also, exporting the "BORG_NEW_PASSPHRASE" to 'old' isn't very useful, rather set the BORG_PASSPHRASE to 'old' (in order to access the REPO normally)

### DIFF
--- a/docs/usage/key.rst
+++ b/docs/usage/key.rst
@@ -35,9 +35,12 @@ Fully automated using environment variables:
 
 ::
 
-    $ BORG_NEW_PASSPHRASE=old borg init -e=repokey repo
+    $ export BORG_PASSPHRASE='old'
+    $ borg init -e=repokey repo
     # now "old" is the current passphrase.
-    $ BORG_PASSPHRASE=old BORG_NEW_PASSPHRASE=new borg key change-passphrase repo
+    $ export BORG_PASSPHRASE='old'
+    $ export BORG_NEW_PASSPHRASE='new' 
+    $ borg key change-passphrase repo
     # now "new" is the current passphrase.
 
 
--- a/docs/usage/key.rst
+++ b/docs/usage/key.rst
@@ -35,11 +35,10 @@ Fully automated using environment variables:
 
 ::
 
-    $ export BORG_PASSPHRASE='old'
+    $ BORG_PASSPHRASE='old'
     $ borg init -e=repokey repo
     # now "old" is the current passphrase.
-    $ export BORG_PASSPHRASE='old'
-    $ export BORG_NEW_PASSPHRASE='new' 
+    $ BORG_NEW_PASSPHRASE='new' 
     $ borg key change-passphrase repo
     # now "new" is the current passphrase.
 


docs/usage/key.rst
```diff
@@ -35,9 +35,12 @@ Fully automated using environment variables:
...
 
-    $ BORG_NEW_PASSPHRASE=old borg init -e=repokey repo
+    $ export BORG_PASSPHRASE='old'
+    $ borg init -e=repokey repo
     # now "old" is the current passphrase.
-    $ BORG_PASSPHRASE=old BORG_NEW_PASSPHRASE=new borg key change-passphrase repo
+    $ export BORG_PASSPHRASE='old'
+    $ export BORG_NEW_PASSPHRASE='new' 
+    $ borg key change-passphrase repo
     # now "new" is the current passphrase.
```
ThomasWaldmann, line 44: some thoughts:

did you try whether your new version works (esp. the borg init part)?

new line 38 and 41 are now duplicate, so 41 is not really needed and could be remove.

but, due to now the env vars being exported, it is not as clear as before any more, what the command is actually making use of.

so, if the bugfix part of this actually works and the original version does not, I'ld prefer a PR that **only** changes this and does not refactor to multiline and usage of export.

Nono-m0le: I had the feeling that the previous version didn't work at all, aka this :
`BORG_NEW_PASSPHRASE=old borg init -e=repokey repo `
beeing executed without quote, would just ignore the "borg" command. So I directly try MY version and it works well.

Regarding the duplicate 38/41 : II would indeed, keep only the line 38. The duplicate make it more clear IMO

ThomasWaldmann:
```
$ FOO='bar' bash -c 'echo $FOO'
bar
user@remote:~$ FOO=bar bash -c 'echo $FOO'
bar
```

Maybe something special about your shell?

ThomasWaldmann: I really do not want the exports there. Besides the scope issue in the example, there can also be scope issue when people use it.

After doing `$ export BORG_PASSPHRASE='old'` and `$ export BORG_NEW_PASSPHRASE='new'` these env vars will be active until the shell terminates. If the user later runs some other "borg key change-passphrase", e.g. on another repo, it will just use these env vars and the user might wonder why borg did not ask.

So, no `export`, but the bugfix part is welcome!

docs/usage/key.rst
```diff
@@ -35,9 +35,11 @@ Fully automated using environment variables:
 
 ::
 
-    $ BORG_NEW_PASSPHRASE=old borg init -e=repokey repo
+    $ BORG_PASSPHRASE='old'
+    $ borg init -e=repokey repo
```
ThomasWaldmann, line 39: i bet this does not work (without "export" this does not stay in the environment).

but stuff like `KEY=value command` works and temporarily puts this into the env of the command process.

docs/usage/key.rst
```diff
@@ -35,9 +35,11 @@ Fully automated using environment variables:
...
     # now "old" is the current passphrase.
-    $ BORG_PASSPHRASE=old BORG_NEW_PASSPHRASE=new borg key change-passphrase repo
+    $ BORG_NEW_PASSPHRASE='new' 
+    $ borg key change-passphrase repo
```
ThomasWaldmann, line 42: same here.

ThomasWaldmann:
@Nono-m0le why did you close?

That bugfix part (name of env var) was likely something we want to have (assuming that the documented way does not work as expected, which seems likely).

PR #5326, opened 2020-09-06
check --repair: fix potential data loss, fixes #5325

### LINKED ISSUE #5325: Aborting check --repair in recover_segment() may leave segment partially recovered
I have not tested this, so may be missing something, but looking at the code:

[`recover_segment`](https://github.com/borgbackup/borg/blob/85e6c0afd82f52182c56bfcc0b795aef33adc037/src/borg/repository.py#L1485) does rename a segment adding the [`.beforerecover`](https://github.com/borgbackup/borg/blob/85e6c0afd82f52182c56bfcc0b795aef33adc037/src/borg/repository.py#L1489) extension and then creates a new file with the original name, [**slowly**](https://github.com/borgbackup/borg/issues/4243#issuecomment-451702272) iterates over the renamed original segment, copying any parts with correct CRC over.

This code is the only occurrence of the string `beforerecover` within the source, so I conclude, that such a renamed original segment is never considered for repair again.

Thus, when a long running `borg check --repair` is interrupted for some reason and that interruption happens within above process, the original (renamed to `.beforerecover`) segment as well as the new segment, which however only has parts of the original data recovered yet, are left behind and a new run of `borg check --repair` would check than new segment only and skip on to the next, if its fine in itself (or if not (or is the new partial one always self-consistent?), maybe even overwrite the original by again renaming the partial new one to `.beforerecover`?).

Either way, with no new errors showing up, the user will assume the `borg check --repair` to have resumed just fine, when it eventually skipped over what worst case could be a large part of a segment containing data that could have been recovered.

[This comment](https://github.com/borgbackup/borg/issues/4243#issuecomment-451702455) seems to have observed that happening in practice, but seems to have went unnoticed within that question.

ThomasWaldmann:
We already have used SaveFile context manager since long at other places.
By using it, the original segment file stays in place until recovery of it
is completed (writing/syncing into *.tmp).
On successful completion, .tmp is renamed over original + dir syncing.
If aborted by some exception, including Ctrl-C, the original file is unmodified.

Fixes #5325.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1438,23 +1438,21 @@ class LoggedIO:
         logger.info('attempting to recover ' + filename)
         if segment in self.fds:
             del self.fds[segment]
-        backup_filename = filename + '.beforerecover'
-        os.rename(filename, backup_filename)
-        if os.path.getsize(backup_filename) < MAGIC_LEN + self.header_fmt.size:
+        if os.path.getsize(filename) < MAGIC_LEN + self.header_fmt.size:
             # this is either a zero-byte file (which would crash mmap() below) or otherwise
             # just too small to be a valid non-empty segment file, so do a shortcut here:
-            with open(filename, 'wb') as fd:
+            with SaveFile(filename, binary=True) as fd:
                 fd.write(MAGIC)
             return
-        with open(backup_filename, 'rb') as backup_fd:
+        with open(filename, 'rb') as src_fd:
             # note: file must not be 0 size or mmap() will crash.
-            with mmap.mmap(backup_fd.fileno(), 0, access=mmap.ACCESS_READ) as mm:
+            with mmap.mmap(src_fd.fileno(), 0, access=mmap.ACCESS_READ) as mm:
                 # memoryview context manager is problematic, see https://bugs.python.org/issue35686
                 data = memoryview(mm)
                 d = data
                 try:
-                    with open(filename, 'wb') as fd:
-                        fd.write(MAGIC)
+                    with SaveFile(filename, binary=True) as dst_fd:
+                        dst_fd.write(MAGIC)
                         while len(d) >= self.header_fmt.size:
                             crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
                             if size < self.header_fmt.size or size > len(d):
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Note: there is no `.beforerecover` any more.

On the one hand, keeping a `.beforerecover` is a bit more safe (in case `recover_segment` would make things worse).
OTOH, the next `borg --repair` would overwrite the first one (original file) with the next one (potentially worse repair try1).

To work around this, we would need to keep all "original states" of segment files before repairing, but that could pile up over time.

Also not sure if users really could use the "before" / original state(s) file(s) for something useful: obviously the original file(s) has some problem (otherwise we would not try to recover it).

ThomasWaldmann:
@elho have a look! ^^^

src/borg/repository.py
```diff
@@ -1463,7 +1461,7 @@ def recover_segment(self, segment, filename):
                             if crc32(d[4:size]) & 0xffffffff != crc:
                                 d = d[1:]
                                 continue
-                            fd.write(d[:size])
+                            dst_fd.write(d[:size])
                             d = d[size:]
                 finally:
                     del d
```
ThomasWaldmann, line 1467: Hmm, guess I better invert the nesting of the context managers, so that the rename at exit of SaveFile does not happen before the mmap is gone and the src_fd is closed.

luke-jr:
>To work around this, we would need to keep all "original states" of segment files before repairing, but that could pile up over time.

Maybe something like an xdelta3 file of the parts removed?

elho:
Looks good. :+1:
And thinknig about it, I doo agree there is little need keeping the original files lingering in there. I and anyone else not willing to put its data at avoidable risk will make a filesystem snapshot or other sort of temporary backup before attempting a repair, which is much easier to use for another attempt or a deeper analysis and debugging in case the repair goes wrong.

ThomasWaldmann:
@luke-jr I feed that putting a lot of effort in optimizing the managing of original files would be a bit over-engineering things. In the end, all these "original files" would have issues and it would be difficult to do anything useful with them for the user anyway.

As @elho said, making a snapshot or copy first is easier to do and easier to use in the end.

PR #5327, opened 2020-09-06
Skip corrupt chunks fast

luke-jr:
When recovery encounters a chunk that the CRC fails on, check if the size is correct by looking at the subsequent chunk offset by that size.

If the next chunk validates, assume the size was correct for the corrupt chunk, and avoid iterating through it byte-by-byte looking for a valid chunk.

Possible drawbacks:
- If it happens to have a corrupt size that coincidentally hits a valid chunk, we might fail to recover valid chunks in between.
- Each time the size is corrupt, we verify the CRC of more data than we needed to, which could significantly slow down cases where the size was corrupt. ~~Since the fallback is iterating byte-by-byte, we effectively double the already-slow process.~~ (Edit: No longer doing that)

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -37,6 +37,7 @@ assert len(ATTIC_MAGIC) == MAGIC_LEN
 TAG_PUT = 0
 TAG_DELETE = 1
 TAG_COMMIT = 2
+all_tags = {TAG_PUT, TAG_DELETE, TAG_COMMIT}
 
 FreeSpace = partial(defaultdict, int)
 
@@ -1482,18 +1483,21 @@ class LoggedIO:
             fd.seek(offset)
             header = fd.read(self.header_fmt.size)
 
-    def recover_segment_chunk(self, d, try_next_chunk=True):
+    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
         """
         Returns a tuple of chunk data (or None) and length of bytes to skip forward
 
         If try_next_chunk is True, then when the CRC mismatches, it will seek past the purported size and check the CRC of the next chunk. If that is correct, or exactly EOF, it will be returned with a length of both the corrupt chunk and the next one being returned.
+        If pedantic is True (always for the next-chunk check), chunk will only be accepted if the tag is valid.
         """
         crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
+        if pedantic and tag not in all_tags:
+            return None, 1
         if size < self.header_fmt.size or size > len(d):
             return None, 1
         if crc32(d[4:size]) & 0xffffffff != crc:
             if try_next_chunk and len(d) >= size + self.header_fmt.size:
-                next_chunk, next_chunk_size = self.recover_segment_chunk(d[size:], try_next_chunk=False)
+                next_chunk, next_chunk_size = self.recover_segment_chunk(d[size:], try_next_chunk=False, pedantic=True)
                 if next_chunk is not None:
                     # If the next chunk is valid, we assume the size of this chunk was correct, and just skip over it without retrying for every byte of the data the CRC fails to match
                     return next_chunk, next_chunk_size + size
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1482,6 +1482,17 @@ class LoggedIO:
             fd.seek(offset)
             header = fd.read(self.header_fmt.size)
 
+    def recover_segment_chunk(self, d):
+        """
+        Returns a tuple of chunk data (or None) and length of bytes to skip forward
+        """
+        crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
+        if size < self.header_fmt.size or size > len(d):
+            return None, 1
+        if crc32(d[4:size]) & 0xffffffff != crc:
+            return None, 1
+        return d[:size], size
+
     def recover_segment(self, segment, filename):
         logger.info('attempting to recover ' + filename)
         if segment in self.fds:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Good idea, but this will likely conflict with #5326, so better work on that after that is merged.

ThomasWaldmann (commented):
Looks like way more efficient than what we had, cool!

Assuming that a chunk is 2MB, then the old code did about 2 Million CRC32 calls due to advancing by 1 byte per call to find the next valid chunk.

But, as this is critical code, it would be good to have some unit tests to make sure it really does what it should, can you add these?

ThomasWaldmann:
Oh, ånd the existing tests fail, please check why. vvv

src/borg/repository.py
```diff
@@ -1482,6 +1483,22 @@ def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
             fd.seek(offset)
             header = fd.read(self.header_fmt.size)
 
+    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
```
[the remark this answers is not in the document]
elho: Yes, especially as on the one hand the second parameter actually is the offset to skip over the returned chunk and not its size (as using the size variable for its values leads one to think at first glance) and on the other hand it is used for the hack to return 1 to make the caller resort to search one byte at a time...

elho:
For now I'm undecided what to think about this. The following aspect scares me quite a bit:
>     * If it happens to have a corrupt size that coincidentally hits a valid chunk, we might fail to recover valid chunks in between.

On the other hand, the original approach of iterating one byte at a time could in theory find what qualifies to it as a chunk within garbled part of chunk and go on from there with a bad offset as well.

So I guess, to evalute this properly, possible corruption scenarios should be systematically listed as well possible issues the current as well as this new code could face with each as well as taking an educated guess at how likely each case would be.
Corruption cases would be:

- a singe bit (or up to a byte) being corrupted
  -   affecting the data
  -   affecting the `crc` header field
  -   affecting  the `size` header field
  -   affecting  the `tag` header field
- multiple bytes up to disk block missing
- multiple bytes up to disk block benig corrupted
- larger amounts of data missing/being corrupted
- malicious data within the segment (crafted to look like an ordinary storage error, but intended to make a repair attempt wreak havoc)

luke-jr:
>On the other hand, the original approach of iterating one byte at a time could in theory find what qualifies to it as a chunk within garbled part of chunk and go on from there with a bad offset as well.

Much less likely (by accident), because the CRC would need to match too.

ThomasWaldmann:
I've looked at the failing test:

```
self = <borg.repository.LoggedIO object at 0x7fcbbac02400>
d = <memory at 0x7fcbbabc3108>, try_next_chunk = True, pedantic = False

    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
        crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
        if pedantic:
            if tag not in all_tags:
                return (None, 1)
        if size < self.header_fmt.size or size > len(d):
            return (None, 1)
        if crc32(d[4:size]) & 0xffffffff != crc:
            if try_next_chunk:
>               next_chunk, next_chunk_size = self.recover_segment_chunk(d[size:], try_next_chunk=False, pedantic=True)
.tox/py35/lib/python3.5/site-packages/borg/repository.py:1495: 

self = <borg.repository.LoggedIO object at 0x7fcbbac02400>
d = <memory at 0x7fcbbabc3348>, try_next_chunk = False, pedantic = True

    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
>       crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
E       struct.error: unpack requires a bytes object of length 9
.tox/py35/lib/python3.5/site-packages/borg/repository.py:1487: error
```

Guess this happens at the end of the segment file, when there is nothing left, but it still tries to look into the next chunk.

src/borg/repository.py
```diff
@@ -1500,20 +1517,21 @@ def recover_segment(self, segment, filename):
                 # memoryview context manager is problematic, see https://bugs.python.org/issue35686
                 data = memoryview(mm)
                 d = data
+                chunk = None
+                alignment_normal = True
                 try:
                     with open(filename, 'wb') as fd:
                         fd.write(MAGIC)
                         while len(d) >= self.header_fmt.size:
-                            crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
```
ThomasWaldmann, line 1507: you see here: it never tries to unpack if there is not enough (>= self.header_fmt.size) data.

src/borg/repository.py
```diff
@@ -1482,6 +1483,22 @@ def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
...
+    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
+        crc, size, tag = self.header_fmt.unpack(d[:self.header_fmt.size])
+        if pedantic:
+            if tag not in all_tags:
+                return (None, 1)
+        if size < self.header_fmt.size or size > len(d):
+            return (None, 1)
+        if crc32(d[4:size]) & 0xffffffff != crc:
+            if try_next_chunk:
+                next_chunk, next_chunk_size = self.recover_segment_chunk(d[size:], try_next_chunk=False, pedantic=True)
```
ThomasWaldmann, line 1495: but here is the check missing, whether there is enough data left.

ThomasWaldmann:
Please rebase this on current master (as expected, this conflicts now).

luke-jr:
Rebased and addressed everything except missing tests. I believe someone familiar with the test suite can likely add those much more efficiently than I could.

src/borg/repository.py
```diff
@@ -1482,6 +1483,29 @@ def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
             fd.seek(offset)
             header = fd.read(self.header_fmt.size)
 
+    def recover_segment_chunk(self, d, try_next_chunk=True, pedantic=False):
+        """
+        Returns a tuple of chunk data (or None) and length of bytes to skip forward
+
+        If try_next_chunk is True, then when the CRC mismatches, it will seek past the purported size and check the CRC of the next chunk. If that is correct, or exactly EOF, it will be returned with a length of both the corrupt chunk and the next one being returned.
```
ThomasWaldmann, line 1490: this line is way too long and triggers the pep8 checker, see travis-ci output.

luke-jr: PEP8 is stupid, and the other comments don't follow this rule, so I'd rather not bother tbh...

ThomasWaldmann: our pep8 checker is configured to ignore some stuff and has high limits on line length, but that line is beyond with no good reason.

ThomasWaldmann: I've fixed that for you.

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
I am closing this, see there:

https://github.com/borgbackup/borg/pull/6022#issuecomment-952378712

@luke-jr Thanks for this PR, which preceeded #6022 and maybe even was an inspiration for it.

When merging #6022 I also looked back at your PR here, just making sure we do not miss anything of major importance.

In the end, I went with #6022 which was just a smaller, more conservative change.

PR #5332, opened 2020-09-10
Keep oldest when retention target not met

amikula:
Borg only considers the last archive in a given time period for retention under a given rule. This code makes a slight modification to that behavior: If the retention target is not met for a rule, it preserves the oldest available archive.

Consider the following retention rules with backups created once per day:

1. keep-daily 7
2. keep-monthly 4
3. keep-yearly 2

In this scenario, an archive created on January 1st in a given year would be expired on January 8th. Archives would continue to be pruned until February 7th, when the January 31st archive would become the first monthly backup to be preserved. In June, the January 31st archive would be expired, and so on. Sometime in the middle of the following year, the December 31st backup would start being preserved as the first yearly backup.

It's reasonable to expect that the initial backup would be preserved until _all_ retention targets are met. This code change causes the oldest backup to be preserved when any retention target is not met.

Pull request is structured in 3 commits:

1. Add test for new behavior
2. Add new behavior
3. Fix tests which were dependent on the old behavior

I'm happy to debate the merits of this feature, restructure the code, add additional tests as necessary, etc.


amikula:
One more thing: I plan to update the documentation once any details are ironed out, but decided to send now for feedback on the behavior, in case it ends up affecting the documentation.

jdchristensen:
> In this scenario, an archive created on January 1st in a given year would be expired on January 8th.

I don't believe that this is correct.  The Jan 1 backup should count as the monthly backup when the Jan 2 to Jan 8 backups are the 7 daily backups.  However, I believe that on Jan 9th, the Jan 1 backup would be deleted.

I'm not sure I like the magic that this patch proposes.  Instead, I'd suggest a `--keep-first N` switch (similar to the existing `--keep-last N`.  Then `--keep-first 1` could be used to preserve the oldest backup.

amikula:
> > In this scenario, an archive created on January 1st in a given year would be expired on January 8th.
> 
> I don't believe that this is correct. The Jan 1 backup should count as the monthly backup when the Jan 2 to Jan 8 backups are the 7 daily backups. However, I believe that on Jan 9th, the Jan 1 backup would be deleted.

That's what I expected too based on the documentation, but instead, in prune_split, only the [last archive of a period](https://github.com/borgbackup/borg/blob/master/src/borg/helpers/misc.py#L54) is considered, and if it's [already in kept_because](https://github.com/borgbackup/borg/blob/master/src/borg/helpers/misc.py#L56), it is not returned as a kept archive. The Jan 1 backup is never considered for a monthly backup target. This matches exactly the behavior I'm seeing on my system.

> I'm not sure I like the magic that this patch proposes. Instead, I'd suggest a `--keep-first N` switch (similar to the existing `--keep-last N`. Then `--keep-first 1` could be used to preserve the oldest backup.

Think of it this way: If I tell borg that I want to keep 6 monthly backups, and it only identifies 5 monthly backups because the 6th would be earlier in the same month, it retains the 6th (the oldest backup) because that's the closest it has. This enables the oldest backup to continue aging unpruned until it is superceded by an actual yearly/monthly/daily backup that meets the normal behavior, while still following the general thinning out behavior that borg already has.

If I go from 6 monthly backups to 6 monthly backups and 2 yearly backups, then as soon as I get a 7th monthly backup, that oldest backup starts getting retained in lieu of my first yearly backup, resulting in me immediately keeping an older backup, instead of potentially having to wait 6 months to get an effective change in behavior.

It's possible that the current behavior of the code does not match the intended behavior (as implied by your expectation), but I do believe this is actually a very reasonable behavior that would match people's expectations pretty well. I can certainly say that it's surprised me to see my oldest backups pruned repeatedly, and I'd be shocked if I ran borg for a full year starting around January 1st and found at the end of that year that my oldest backup was only about 7 months old.

`--keep-first 1` would always keep the first archive and never expire it, which is not what I think most people would want.

jdchristensen:
I think that's a bug in the code.  I touched that code many years ago and wrote one version of the documentation, so I'm at least partly to blame.  The line checking if `a.id` is in `kept_because` should probably come as the first line of the loop, something like:

```python
    for a in sorted(archives, key=attrgetter('ts'), reverse=True):
        if a.id in kept_because:  continue
        period = to_localtime(a.ts).strftime(pattern)
        if period != last:
            last = period
            keep.append(a)
            kept_because[a.id] = (rule, len(keep))
            if len(keep) == n:
                break
```

Untested, and I'm swamped with other things right now, but this should probably make it match the docs.

Even with that change, it won't fix the problem you are mentioning, except to delay the problem.  On Jan 10, I think the Jan 3 backup would be saved as a monthly and the Jan 2 as a yearly, but the Jan 1 would still be deleted.

So I'm not opposed to the change.  If I understand correctly, the docs would need to be updated to say something like:

If after considering all archives, there are rules that didn't fully match, the oldest archive is also kept.

amikula:
Ah, I think you're right, that would probably make the code match the documentation. I'm afraid it might have some odd side effects in a few edge cases though, so I'd like to take some time to test it out a bit.

Any suggestions on setting up test archives with a given timestamp? Looks like so far everyone has sidestepped that in `testsuite/archiver.py`.

I was thinking of setting up a helper that creates an archive, loads it, and updates the timestamp in the metadata. Would that even work?

Maybe the best thing would be to directly write tests for `do_prune` in `archiver.py`?

amikula:
Also, should I address the fact that there's an apparent side effect of 0.43% code coverage loss even though code I've added is covered? I think there was some incidental coverage that was lost when I updated the existing test cases, but I haven't had a chance to dig into it yet.

jdchristensen:
The MockArchive allows a timestamp, but maybe you mean something else?

amikula:
Well, the current tests that use `MockArchive` are only directly testing the `prune_split` helper method. The behaviors we're discussing are dependent on more complex behavior dependent on `Archiver::do_prune`, which is currently only tested with command calls. I think I'll try testing `do_prune` with mock objects. Hopefully that won't be too hard.

As for the modified code you've proposed, I've thought about it a bit, and if I've got the impact right, I prefer the current behavior.

With the proposed modification, again if I've got it right, If the user specifies `--keep-daily 7 --keep-yearly 1`, there is literally no difference in behavior from `--keep-daily 8`. As soon as there's an 8th daily archive, it will be selected as the first yearly archive instead of being pruned, and the next older archive will get pruned. This behavior compounds, such that `--keep-daily 7 --keep-weekly 1 --keep-monthly 1 --keep-yearly 1` will behave identically to `--keep-daily 10`. Sure, the behavior changes with values greater than 1, but I think this behavior is counterintuitive, and certainly a significant departure from the current behavior.

jdchristensen:
I think your analysis is correct, and I have a slightly preference the behaviour that my change would produce.  For one, it seems easier to document (and is already documented).  Moreover, if all I had done were 8 daily backups in the same year and I asked borg to keep 7 daily backups and 1 yearly, I would expect it to keep all 8.  (Of course, with your other proposed change, this would happen too, but a similar situation arises even with that change.)  OTOH, if my change is made, people will suddenly lose some of their oldest backups, so it would need to be announced broadly.

If we keep the current behaviour, we'd need to modify the docs.  Do you think this can be done without making them even more confusing?  People already have a lot of trouble understanding pruning...  Oh, looking some more, I see that the current behaviour is documented in the detailed example at the end of [this page](https://borgbackup.readthedocs.io/en/stable/usage/prune.html).  So maybe it would actually be better to keep the current behaviour and update the "borg help prune" to match this.

To be honest, I think one of the reasons these subtleties hadn't been noticed is that borg is so space efficient that you can design your pruning rules to only very gradually prune things, and have things like infinite yearly retention.

amikula:
I did a diff between `docs/usage/prune.rst.inc`, `docs/man/borg-prune.1`, and `src/borg/archiver.py`, and they're all in sync both for master and for 1.1.13. Looks like there's at least partial automation to sync these docs.

The docs do say "--keep-daily 7 means to keep the latest backup on each day" for an example, but they don't state that only the latest backup is ever considered for the daily backup, and if it's already chosen, a _previous_ day will be chosen as opposed to a previous backup from the same day. I think that's where the confusion comes in, and may be worth documenting explicitly.

As for why this hasn't come up, we're definitely talking about edge cases here. Most people probably won't prune automatically and watch its behavior, and most people won't find themselves needing that one backup from the day they started backing up their system.

src/borg/testsuite/helpers.py
```diff
@@ -400,6 +400,38 @@ def subset(lst, ids):
...
+    assert kept_because[3][0] == "yearly"
+    assert kept_because[4][0] == "yearly"
+
+    # Otherwise, prune it
+    kept_because = {}
+    keep = prune_split(archives, "yearly", 2, kept_because)
+
+    assert set(keep) == subset(archives, [3, 4])
+    for item in keep:
+        assert kept_because[item.id][0] == "yearly"
```
ThomasWaldmann, line 432: maybe it's easier if that part of the test is implemented in the same way as above (line 422..424), without a loop.

ThomasWaldmann (commented):
if we implement this, guess this will be master-only (borg 1.2.x only).

all the test changes - are they expected / correct according to new behaviour?

https://github.com/borgbackup/borg/blob/master/docs/misc/prune-example.txt needs checking.

also, docs need an update that this behaviour is new compared to borg 1.1.x and earlier.

ThomasWaldmann:
also a test that has more complex / multi-interval scenario (many people use all of daily, monthly, yearly), but that is still manually easy to verify would be good.

amikula:
> if we implement this, guess this will be master-only (borg 1.2.x only).

Yes, I understand this to be consistent with the project's branching policy. Out of curiosity, when is that release planned? This pull request still needs a bit of polishing, and I don't want to miss the release.


> all the test changes - are they expected / correct according to new behaviour?

Yes, I believe so. The tests all depended on the behavior of dropping a previous backup in the same time period as the latest backup when 2 archives were requested to be kept. That behavior didn't seem relevant the names of the tests, so I assume it was incidental to the tests. I wouldn't mind if someone double-checked me there, though.


> https://github.com/borgbackup/borg/blob/master/docs/misc/prune-example.txt needs checking.

The example doesn't include any cases where an archive is dropped at all, and all retention targets are met, so this code doesn't apply to the example as currently written.


> also, docs need an update that this behaviour is new compared to borg 1.1.x and earlier.

Ok.


> also a test that has more complex / multi-interval scenario (many people use all of daily, monthly, yearly), but that is still manually easy to verify would be good.

Agreed, multi-interval scenarios are completely missing from the tests, and at a minimum it would be nice to have one that matches the example cited above. This doesn't look easy to add given the way tests on `archiver.py` are currently implemented, but I'll give it a go.

jdchristensen:
If the behaviour is changed to keep the oldest backup when the keep rules aren't all satisfied, then I think the detailed prune-example.txt should also illustrate this behaviour, e.g. by adding a --keep-yearly 1 rule and some explanation.  (I'm also not convinced that displaying that whole calendar is helpful, but that's another issue.)

amikula:
By the way, one aspect of the behavior we haven't discussed:

The patch adds "[oldest]" after the name of the rule that caused retention, in this special case. Is that really desirable or necessary? Does it help, or will it cause confusion?

amikula:
> If the behaviour is changed to keep the oldest backup when the keep rules aren't all satisfied, then I think the detailed prune-example.txt should also illustrate this behaviour, e.g. by adding a --keep-yearly 1 rule and some explanation. (I'm also not convinced that displaying that whole calendar is helpful, but that's another issue.)

I agree, I think this would be useful to include in the example.

amikula:
So here's my to-do list at this point. If nobody corrects this list, I'll assume they're necessary and most probably sufficient to have this pull request accepted (not looking for a commitment that this will be accepted based solely on the following work, but also don't want to do the work if it's not likely to be accepted anyway):

1. Correct the docs to add a statement about this behavior, calling out the fact that it's new in 1.2.0.
2. Add this new case to the example.
3. Add tests that verify pruning behavior with multiple --keep-X intervals, including one matching the example.


Possibly for a separate pull request that could be backported to 1.1.X branch (ie, not impacted by the behavior in this branch):

1. Add a statement to the docs clarifying that _only_ the latest backup in a time period is considered for retention under that time period's rule, and if it has already been selected, a backup from an _earlier time period_ will be selected, rather than an earlier backup in the same time period.

ThomasWaldmann:
borg 1.2 (master branch) is still in alpha releases, so that will still take a while...

docs/misc/prune-example.txt
```diff
@@ -53,16 +57,16 @@ Mo Tu We Th Fr Sa Su  Mo Tu We Th Fr Sa Su  Mo Tu We Th Fr Sa Su
 List view
 ---------
 
---keep-daily 14     --keep-monthly 6
+--keep-daily 14     --keep-monthly 6     --keep-yearly 6
```
ThomasWaldmann, line 60: did you mean --keep-yearly 1 (as in the above example)?

docs/misc/prune-example.txt
```diff
@@ -76,18 +80,23 @@ Notes
...
-first), not due to the --keep-monthly rule.
+first), not due to the --keep-monthly or --keep-yearly rule.
 
-Because of that, the --keep-monthly 6 rule keeps Nov, Oct, Sep, Aug, Jul and
-Jun. December is not considered for this rule, because that backup was already
-kept because of the daily rule.
+The --keep-yearly 1 rule does not consider the December 31st backup because it
+has already been kept due to the daily rule. There are no backups available
+from previous years, so the --keep-yearly target of 1 backup is not satisfied.
+Because of this, the 2015-01-01 archive (the oldest archive available) is kept.
```
ThomasWaldmann, line 88: What would happen for --keep-yearly 2? Would it keep 2015-01-01 and 2015-01-02?

docs/misc/prune-example.txt
amikula: Whoops, copypasta error. Fixed.

docs/misc/prune-example.txt
amikula: No, it would only keep 2015-01-01, because that is the oldest archive available. (In the example, the user started backups on 2015-01-01).

The only change from the previous behavior is this: If any rule can't match its retention target (in your example, 2 yearly archives), it will retain the oldest archive if no previous rule has retained it. In this same example, without this patch, borg would prune that archive, even with the `--keep-yearly 1` rule.

This will generally mean that the first archive the user ever made with borg is retained until it is expired because it has been replaced by an archive that matches the rules given.

src/borg/testsuite/helpers.py
amikula: Done.

ThomasWaldmann:
Had time to take a deeper look: I think the existing code in this PR is correct and the behavior change is desirable.

Of course it makes the behavior a bit more complex and some more people might think "prune is not working correctly" (we already had that quite often), but guess that's life and we can still point them at the docs / at the prune example.

Super cool would be to have 1 test that matches the example and proves that the example is actually correctly showing what the code does.

So, thanks for the PR, go on and tell us when finished for a final review.

ThomasWaldmann (approved):

amikula:
I've added the test matching the prune example and a test for daily archive and prune confirming that the oldest is kept and eventually expired. Let me know if you'd like to see any additional tests or changes to the tests.

One thing I don't like about the tests is that they're somewhat **dependent on the developer's time zone**, because command line timestamps are in UTC but prune interprets the timestamps in local time. I've compensated for that by getting the current timezone and offsetting from UTC accordingly in the tests. Also note that **DST changes are not accounted for**. I've put the test timestamps at midday so there won't be problems with this, but it's a kludge that may bite someone later on down the road. The alternative of calculating the correct local offset for a particular day doesn't seem to be handled by Python core libraries, so I think the tests would need to **bring in a new library to handle localtime correctly for dates in the past**. Even then, the offset for the tests will be different depending on the environment's timezone, so it's probably best to **bring in a testing library** (eg. freezegun) to **make the tests independent of the test environment's timezone**.

Please let me know how you'd like to see this addressed.

amikula:
Force pushed a minor change to the test code (`_to_utc_timestamp` helper method) to correctly map to UTC before creating the timestamp.

amikula:
Force pushed an update to move dateutil import to match alphabetical order of other imports in `src/borg/testsuite/archiver.py`

src/borg/testsuite/archiver.py
```diff
@@ -2023,17 +2024,108 @@ def test_prune_repository(self):
...
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 23, 0, 0, 0), self.repository_location + '::test13', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 24, 0, 0, 0), self.repository_location + '::test14', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 25, 0, 0, 0), self.repository_location + '::test15', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 26, 0, 0, 0), self.repository_location + '::test16', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 27, 0, 0, 0), self.repository_location + '::test17', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 28, 0, 0, 0), self.repository_location + '::test18', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 29, 0, 0, 0), self.repository_location + '::test19', src_dir)
+        self.cmd('create', '--timestamp', self._to_utc_timestamp(2015, 12, 30, 0, 0, 0), self.repository_location + '::test20', src_dir)
```
ThomasWaldmann, line 2059: idea how to make this code easier / less duplicate:

```
def create_archive_ts(name, y, m, d, H=0, M=0, S=0):
    loc = self.repository_location + '::' + name
    self.cmd('create', '--timestamp', self._to_utc_timestamp(y, m, d, H, M, S), loc, src_dir)
```

... and then just call that.

ThomasWaldmann:
About #5535 : fixing that globally is not in scope of this PR (like by introducing new test tooling / frameworks).

So, if the current solution here works for the prune tests, we just use that for now.

amikula:
Makes sense. I'll reduce the duplication as you've suggested above, and I think this is ready to go.

If I understand the dev guidelines correctly, I should squash together the commits in this branch into a cleaner changeset.

Would you like to see it squashed down to the same structure I started with:

1. Add test for new behavior
2. Add new behavior
3. Fix tests which were dependent on the old behavior

ThomasWaldmann:
That kind of squashing would be perfect (if it is not too much effort).

On the github UI i can just have all separate or all squashed into 1 commit...

src/borg/testsuite/archiver.py
ThomasWaldmann: thanks for fixing!

ThomasWaldmann:
@amikula are you working on the commits still or shall I merge it as is?

amikula:
I think I'll have time to reorganize the commits today. Spending some time on it right now.

amikula:
Ok, just force pushed the reorganized code. I also rebased against master.

I ended up with a 4th commit with the documentation changes and associated tests.

Feel free to still squash into a single commit if you see fit, or to request any other changes.

Cheers,

-Alf

ThomasWaldmann:
Thanks! Merged, will be in next 1.2 pre-release!

amikula:
Great! I'm glad to be able to make this small contribution to the project. I hope it works out well!

If bugs are reported against this functionality, will they be connected to this pull request in some way?

ThomasWaldmann:
Only if the PR number is mentioned. But guess this is rather unlikely because people tend to just report "my prune does not work like i think it should".

PR #5341, opened 2020-09-23
Fix a few issues in zsh completion

oxiedi:
Namely:

  - add tags to all untagged contexts;
  - don't leak $debug_commands;
  - plus some cosmetic changes.

### DIFF
--- a/scripts/shell_completions/zsh/_borg
+++ b/scripts/shell_completions/zsh/_borg
@@ -5,7 +5,7 @@
 # Recommended _borg specific settings:
 #
 #   zstyle -e ':completion:*:*:borg-*:argument-rest:*' tag-order \
-#     '[[ $words[CURRENT] == -* ]] && reply=( "!archives archive-files" "-" )'
+#     '[[ $words[CURRENT] == -* ]] && reply=( "! archives archive-files" "-" )'
 #   zstyle ':completion:*:*:(borg|-value-,BORG_)*' sort false
 #   zstyle ':completion:*:*:borg-config:argument-2:keys' list-grouped false
 #   zstyle ':completion:*:*:borg-*:*' gain-privileges true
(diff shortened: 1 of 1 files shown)


oxiedi:
According to milestones, you are about to release `1.1.14` and `1.2.0a9`. It might be a good time to update shell completions, correct?

ThomasWaldmann:
@oxiedi yes, 1.1.14 asap, afterwards 1.2.0a9.

i can't review zsh completion changes, can you look out for another reviewer of this PR?

oxiedi:
> @oxiedi yes, 1.1.14 asap, afterwards 1.2.0a9.

OK, I'll issue PRs after merging of this one.

> i can't review zsh completion changes, can you look out for another reviewer of this PR?

Well, maybe @SanskritFritz can give us a hand here. The tags part simplifies configuration for a user. E.g., in order to complete archives in a less verbose manner, before this change, one had to use:

    zstyle ':completion:*' verbose true
    zstyle ':completion:*:*:borg-(list|info|...):argument-1:' verbose false
    zstyle ':completion:*:*:borg-(diff|delete|...):argument-[1-9]:' verbose false

After:

    zstyle ':completion:*' verbose true
    zstyle ':completion:*:*:borg-*:*:archives' verbose false
    
It should be obvious what other changes do.

ThomasWaldmann:
@SanskritFritz can you have a look?

SanskritFritz:
I reviewed the changes, they look good to me. However, I don't use zsh, so my testing is not extensive. But looking at the changes I see lots of bugfixes (how did the '#q' slip in earlier? :smile: ), the functionality has not been changed.

ThomasWaldmann:
@SanskritFritz thanks for reviewing.

Not sure I understand "how did the '#q' slip in earlier" - as it was added by this PR.

oxiedi:
Depending on the state of the `bareglobqual` option, `#q` may be necessary to distinguish glob qualifiers from a pattern. However, the completion system has its own set of options (stored in `$_comp_options`) in effect, which includes `bareglobqual`. So this is just a matter of preference, whether to use `#q` here or not. I hope this makes any sense to you.

PR #5356, opened 2020-09-27
Update shell completions

oxiedi:
CC @SanskritFritz


SanskritFritz:
Fish and bash completion changes are perfect, thanks @oxiedi 
The zsh completions changes look OK to me, but that is not really my area.

PR #5366, opened 2020-09-29
upgrade bundled xxhash to 0.7.4

ThomasWaldmann:


ThomasWaldmann:
Hmm, there are compile issues. Master branch did compile. wtf?

https://travis-ci.org/github/borgbackup/borg/builds/731442394

ThomasWaldmann:
I remember that I already had such issues due to the xxh3 code, so I just left that part of the code away in earlier borg releases.

But now, xxh3 is integrated into xxhash.c/.h, so I can not easily leave it away any more.

ThomasWaldmann:
Another option, if the compile issue can't be easily fixed, is to just leave this as is and only have this upgrade in master branch.

elho:
> I remember that I already had such issues due to the xxh3 code, so I just left that part of the code away in earlier borg releases.

Yes, that error message shows up if functions of extended intructions sets such as SSE and AVX are used, but the according -mXXX option to generate such code is not set. However, for borg it makes no sense to bulid e.g. SSE only, given the speed of xxHash compared to the rest of borg, the benefit would be hardly noticeable, if at all, and of course, being able to backup machines that do not have SSE instructions is a requirement anyway.

The xxHash [README](https://github.com/Cyan4973/xxHash/blob/dev/README.md) discusses the `XXH_VECTOR` macro. Maybe setting it to `XXH_SCALAR` scalar is enough to either bulit non-SSE2 variants of XXH3 and XXH128 or not built those. If that does not work, that would be a an issue on the xxHash side.

Cyan4973:
Indeed, `XXH_VECTOR=XXH_SCALAR` can be used to force the code to only employ regular instructions.
In general, it results in lower performance for `XXH3` (but it doesn't impact `XXH64`).
The resulting binary _might_ still be auto-vectorized, but that's now the responsibility of the compiler.

The one thing that surprises me though is that the compilation error message seems to choke at some `SSE2` instructions.
> `emmintrin.h:1029:1: error: inlining failed in call to always_inline ‘_mm_add_epi64’: target specific option mismatch`

But `SSE2` is not some kind of "optional" extension : on `x64` platforms, it is mandatory, plain and simple. There is no `x64` cpu without `SSE2`. Consequently, all compilers I'm aware of (`gcc`, `clang`, `visual`, `icc`) default to `SSE2` enabled when compiling for `x64`. And at least some compilers (tested on `clang`) are unable to generate `x64` binaries if one tries to disable `SSE2`.

`SSE2` is not guaranteed on 32-bit `x86`, so it makes more sense to specifically disable it for this platform. But my understanding is that compilation tests on `travisCI` target `x86-64`, so that's not the use case. Moreover, no special action should be needed for 32-bit `x86` : `XXH3` source code  should determine the outcome automatically, and disable vector code by default.

ThomasWaldmann:
@Cyan4973 @elho thanks for the insights. Still strange that in master branch it just worked.

ThomasWaldmann:
Guess this might be an issue with the 1.1-maint "build system", master is cleaner there.

I'll updated this PR to just use the latest pre-0.8.0 code, where I can just remove the problematic xxh3.h (we don't use any of the new stuff yet anyway).

PR #5377, opened 2020-10-03
docs: how to approach borg speed issues, fixes #5371

### LINKED ISSUE #5371: check the docs + faq about files cache
there are frequent questions like "why is borg slow for me, considering all my files modified" when the underlying filesystem has issues with:
- stable inode numbers
- when people do not always have the same full/absolute paths
- when they have ctime issues (like cron scripts running chown -R/chmod -R)

there are quite some related posts on this issue tracker (also see closed issues) and on the borg mailing list.

the usual way to debug / fix this is using `borg create --list ...` to see file status and whether a lot of files are considered modified although they should not be plus using the `stat` command on these files to see why borg considers them changed or has a files cache miss (if not obvious, e.g. due to full path change).

fixing the issue is sometimes out of borg scope (e.g. mounting always to same mountpoint, not doing chown/chmod by cron) and sometimes tweaking borg's `--files-cache` (e.g. if the source fs does not have stable inode numbers).

check our docs (esp. borg create and faq) and make them more complete / better discoverable, so less people ask these questions.

ThomasWaldmann:
also: give a speed example for unchanged files

also: explicitly mention chmod/chown changing ctime

also: reformat one paragraph

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -736,6 +736,67 @@ If you run into that, try this:
   the parent directory (or even everything)
 - mount the repo using FUSE and use some file manager
 
+.. _expected_performance:
+
+What's the expected backup performance?
+---------------------------------------
+
+A first backup will usually be somehow "slow" because there is a lot of data
+to process. Performance here depends on a lot of factors, so it is hard to
+give specific numbers.
+
+Subsequent backups are usually very fast if most files are unchanged and only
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
@fantasya-pbem can you review, please?

ThomasWaldmann:
merging. if changes are needed, they can be applied in a future PR also.

PR #5379, opened 2020-10-04
try llfuse on py39

ThomasWaldmann:
there was a fixed llfuse release today, so lets try it!


ThomasWaldmann:
Works! (I killed some of the test runs, py38 and py39 worked!)

PR #5391, opened 2020-10-06
Clarify '--one-file-system' for btrfs

eike-fokken:
This addresses #4009. The documentation now explicitly mentions
btrfs subvolumes.

I'm happy to change the wording if the PR is welcome but the exact wording is not.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3111,6 +3111,16 @@ class Archiver:
         and not include any other contents of the containing folder, this can be enabled
         through using the ``--keep-exclude-tags`` option.
 
+        The ``-x`` or ``--one-file-system`` option excludes directories, that are mountpoints (and everything in them).
+        It detects mountpoints by comparing the device number from the output of ``stat`` of the directory and its
+        parent. Be aware that in Linux there are directories with device number different from their parent, which the
+        kernel does not consider a mountpoint and also the other way around. Examples are bind mounts (possibly same
+        device number, but always a mountpoint) and ALL subvolumes of a btrfs (different device number from parent but
+        not necessarily a mountpoint). Therefore in Linux one should make doubly sure that the backup works as intended
+        especially when using btrfs. This is even more important, if the btrfs layout was created by someone else, e.g.
+        the distribution installer.
+
+
         .. _list_item_flags:
 
         Item flags
@@ -3220,7 +3230,7 @@ class Archiver:
 
         fs_group = subparser.add_argument_group('Filesystem options')
         fs_group.add_argument('-x', '--one-file-system', dest='one_file_system', action='store_true',
-                              help='stay in the same file system and do not store mount points of other file systems. Be aware that this treats every btrfs subvolume as a separate file system and will not back them up! Make doubly sure to check your backup does what you want when using btrfs.')
+                              help='stay in the same file system and do not store mount points of other file systems.  This might behave different from your expectations, see the docs.')
         fs_group.add_argument('--numeric-owner', dest='numeric_owner', action='store_true',
                               help='only store numeric user and group identifiers')
         # --noatime is the default now and the flag is deprecated. args.noatime is not used any more.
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -3220,7 +3220,7 @@ def define_borg_mount(parser):
 
         fs_group = subparser.add_argument_group('Filesystem options')
         fs_group.add_argument('-x', '--one-file-system', dest='one_file_system', action='store_true',
-                              help='stay in the same file system and do not store mount points of other file systems')
+                              help='stay in the same file system and do not store mount points of other file systems. Be aware that this treats every btrfs subvolume as a separate file system and will not back them up! Make doubly sure to check your backup does what you want when using btrfs.')
```
ThomasWaldmann, line 3223: this gets rather (too) long for here.

also, i don't want to get into filesystem specifics here (and potentially point out every filesystem that has such specific stuff).

rather than adding it to the (short) help here, one could check whether it is documented how `-x` works (see the check in the code - iirc it is checking for the device number [not] changing) and if that is not documented, add it there.

after talking about how borg -x works, one could also point out that some filesystems trigger this change detection:
- btrfs for subvolumes
- (more to come?)

eike-fokken:
> rather than adding it to the (short) help here, one could check whether it is documented how `-x` works (see the check in the code - iirc it is checking for the device number [not] changing) and if that is not documented, add it there.

I don't understand what you mean by "check in the code". I found the code relating to `--one-file-system` in archiver.py.
To my understanding it explicitely checks the device number from stat.
But is there further documentation on this?

> after talking about how borg -x works, one could also point out that some filesystems trigger this change detection:
> 
> * btrfs for subvolumes
> * (more to come?)

I think, borg create is the only command accepting `--one-file-system`.
So, do you mean, I should add a paragraph in the text part of the borg create page?

eike-fokken:
> rather than adding it to the (short) help here, one could check
whether it is documented how `-x` works (see the check in the code -
iirc it is checking for the device number [not] changing) and if that
is not documented, add it there.

I don't understand what you mean by "check in the code". I found the
code relating to `--one-file-system` in archiver.py.
To my understanding it explicitely checks the device number from stat.
But is there further documentation on this?

> after talking about how borg -x works, one could also point out that
some filesystems trigger this change detection:
> 
> * btrfs for subvolumes
> * (more to come?)

I think, borg create is the only command accepting `--one-file-system`.
So, do you mean, I should add a paragraph in the text part of the borg
create page?

ThomasWaldmann:
Yes, that is what I meant, add a paragraph to user docs.

Just say what it does so that it is clear how "one filesystem" is defined.

ThomasWaldmann:
This PR is changing way to many files.

ThomasWaldmann:
Do not run build_usage / build_man (or at least do not commit the result, so it is more clear what the PR is actually changing) - this is done at release time.

eike-fokken:
Oh, sorry, I'm not familiar with github pull requests. I meant to make a differen pull request, but it got mashed into this one. I will have look later on or tomorrow.

eike-fokken:
I now inserted a paragraph in the prose part of the docs.

ThomasWaldmann:
Looks good.

Can you get some other btrfs user from #4009 to review this also (I am unfamiliar with btrfs specifics)?

eike-fokken:
I will.
What does it mean, that the codecov/project check fails?

ThomasWaldmann:
@eike-fokken it means the overall test coverage could be improved. Unrelated to this PR.

eike-fokken:
I got replies from two btrfs users and included their comments, see #4009.

ThomasWaldmann:
Looks good, are you finished?

eike-fokken:
I am!

That was a pleasant experience, thanks for your time!

ThomasWaldmann:
Thanks for working on this. Guess we better do a backport to 1.1-maint (as the 1.1.x docs are what people usually read).

eike-fokken:
Seems like a good idea, my system's borg is 1.1.13 and I am on Debian testing, so rather up-to-date.

PR #5427, opened 2020-10-17
export-tar: set tar format to GNU_FORMAT explicitly

### LINKED ISSUE #5274: Tar export on Ubuntu 20.04 outputs @paxheader files
Yes

ISSUE

borg 1.1.13
Ubuntu 20.04.1 LTS

borg export-tar --tar-filter="gzip -3" --strip-components 6 /var/borg/$App::$DownloadName $DownloadName.tar.gz

Gzip/tar file contains lots of @PaxHeader files and missing file extensions.

Is there a way to force GNU export over POSIX from tar in borg? Tar is set to GNU by default on the system but I haven't found how to change this within the borg files, there seems to be something overriding this. I have borg installed and exporting ok on older Ubuntu LTS versions, and can only replicate this issue on 20.04

enkore:
Python 3.8 changed the default format to PAX, but GNU format is documented for Borg.

Fixes #5274

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -933,7 +933,7 @@ class Archiver:
 
         # The | (pipe) symbol instructs tarfile to use a streaming mode of operation
         # where it never seeks on the passed fileobj.
-        tar = tarfile.open(fileobj=tarstream, mode='w|')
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks!

PR #5429, opened 2020-10-18
Remove Python3.5 from tox.ini as it is not supported anymore

### LINKED ISSUE #5426: tox.ini is outdated compared to setup.py

Yes

ISSUE

master branch

## Problem : 

The [setup file](https://github.com/borgbackup/borg/blob/master/setup.py#L297) 
specifies that the python version should be >=3.6 but the tox.ini still lists py35 as a possible test version.
py35 should be removed from tox.ini.

I can do this if you're OK with it.

Gu1nness:
tox.ini still had py35 in it's configuration while the setup.py [specifies](https://github.com/borgbackup/borg/blob/master/setup.py#L297) that the minimum supported version is py36.
This commits fixes this issue.

Fixes #5426

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 # fakeroot -u tox --recreate
 
 [tox]
-envlist = py{35,36,37,38,39},flake8
+envlist = py{36,37,38,39},flake8
 
 [testenv]
 deps =


ThomasWaldmann:
Thanks!

PR #5433, opened 2020-10-21
Allow appending the files cache filename with BORG_FILES_CACHE_SUFFIX env var

luke-jr:


ThomasWaldmann:
So we have a PR now, but there is no issue what this is "fixing"?

luke-jr:
It's more of a feature than a fix, but if you want to know the motivation, it's because my files cache is 2.2 GB and it doesn't make sense to load the whole thing for each archive. Instead, now I can load only the (much smaller) cache for the partition I'm actually backing up.

ThomasWaldmann:
The pep8 checker is failing:
```
src/borg/cache.py:228:1: E302 expected 2 blank lines, found 1
```

This time (and in future), please fix such stuff on your own. You can run the test locally by invoking `tox`.

src/borg/cache.py
```diff
@@ -225,6 +225,9 @@ def recanonicalize_relative_location(cache_location, repository):
 def cache_dir(repository, path=None):
     return path or os.path.join(get_cache_dir(), repository.id_str)
 
+def files_cache_name():
+    return 'files' + os.environ.get('BORG_FILES_CACHE_SUFFIX', '')
```
ThomasWaldmann, line 229: if you set `BORG_FILES_CACHE_SUFFIX=foo`, this will be `filesfoo`.

works, but is a bit ugly. how about adding a dot in between IF the env var is not empty?

ThomasWaldmann (commented):
Looks good, just found a minor nitpick.

To be useful, the docs need some words about this feature.

There is a list of supported env vars in `docs/usage/.../environment.rst.inc`.

src/borg/cache.py
fantasya-pbem: Why not replacing 'files' completely with a env variable `BORG_FILES_CACHE_NAME=files`?

ThomasWaldmann: guess that could be harmful, just imagine someone setting it to `chunks` (or some other already taken name, or empty).

besides that, the docs talk about the "files cache", so it should be recognizable as that.

src/borg/cache.py
```diff
@@ -497,7 +500,7 @@ def create(self):
         self.cache_config.create()
         ChunkIndex().write(os.path.join(self.path, 'chunks'))
         os.makedirs(os.path.join(self.path, 'chunks.archive.d'))
-        with SaveFile(os.path.join(self.path, 'files'), binary=True):
```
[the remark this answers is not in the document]
luke-jr: Yes, I searched for `[/"']files`

src/borg/cache.py
luke-jr: Personally, I am using a hyphen as my lead character for `BORG_FILES_CACHE_SUFFIX`. A dot historically implies a file type. Does a hyphen seem acceptable to you?

luke-jr:
Don't really have any interest in auditing/installing `tox`, especially not with something as convoluted as `virtualenv`... Perhaps Travis is enough?

luke-jr:
Added an implicit hyphen (pending agreement) and documentation.

src/borg/cache.py
ThomasWaldmann: ### examples from borg cache:

`chunks` vs. `chunks.archive.d` - uses a `.` to separate different parts of file name.

### examples from borg repo:

`hints.NN`, `index.NN`, `integrity.NN` (with NN being a number) - also uses a dot to separate parts of filename. `NN` is not the file type here, but the segments transaction id.

### examples from borg config:

`key-type`, `manifest-timestamp` - a dash `-` is just used for readability, not as a separator to append a scope.

so guess using a dot is more consistent with what we already have.

ThomasWaldmann:
About tox / flake8: yes, of course you can also just rely on travis-ci result.

Important is just that there are no new pep8 related failures blocking the merging of this PR.

docs/usage/general/environment.rst.inc
```diff
@@ -56,6 +56,10 @@ General:
     BORG_REMOTE_PATH
         When set, use the given path as borg executable on the remote (defaults to "borg" if unset).
         Using ``--remote-path PATH`` commandline option overrides the environment variable.
+    BORG_FILES_CACHE_SUFFIX
+        When set to a value at least one character long, instructs borg to use a specifically named
+        (based on the suffix) alternative files cache. This can be used to avoid loading and saving
+        cache entries for backup sources other than the current target.
```
ThomasWaldmann, line 62: sources and target (usually rather meaning the opposite than the same thing) are a bit confusing here?

so, rather say `... other than the current sources.` ?

src/borg/cache.py
```diff
@@ -226,6 +226,13 @@ def cache_dir(repository, path=None):
     return path or os.path.join(get_cache_dir(), repository.id_str)
 
 
+def files_cache_name():
+    suffix = os.environ.get('BORG_FILES_CACHE_SUFFIX', '')
+    if suffix:
+        suffix = '-' + suffix
+    return 'files' + suffix
```
ThomasWaldmann, line 233:
```
return 'files-' + suffix if suffix else 'files'
```

ThomasWaldmann:
ping

luke-jr:
Changes made

ThomasWaldmann:
Thanks! Do you want to do a 1.1-maint backport of this?

luke-jr:
It's a new feature, so is it even eligible?

ThomasWaldmann:
That depends on a lot of factors. As this is rather small and simple, off-by-default, useful-when-you-need-it, we could consider it.

FabioPedretti:
I tried this with 1.1.15 and I can say it's a huge scalability improvement over the old way to manage multiple backup sets with BORG_FILES_CACHE_TTL. This is especially true when having a huge repo along many tiny repos. I added a couple of followup suggestions in #5602.

PR #5436, opened 2020-10-23
Change documentation inaccuracy on chunk size.

### LINKED ISSUE #5336: The meaning of "statistical medium".

Yes.

A request for clarification of a term in the documentation.

## Details

This appears at a few places in the documentation:
"HASH_MASK_BITS = 21 (**statistical medium** chunk size ~= 2^21 B = 2 MiB)"

At first I thought that by 'medi **_um_**' it meant 'medi **_an_**', but in another place it says
"producing chunks of 2^HASH_MASK_BITS Bytes on **_average_**",
and average usually means 'arithmetic mean'.

Can you say which measure of '[central tendency](https://en.wikipedia.org/wiki/Central_tendency)' is the intended/correct meaning of 'statistical medium'?

When I know which it means, will I be able to choose my chunker params any better? Not sure about that ... ;)

Gu1nness:
We know use only "target chunk size" when speaking of the chunk size
that is expected to happen most of the time. This removes statistical
and mathematical inacurracies that could be troublesome for mathematical
people.

Fixes #5336

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -608,8 +608,8 @@ default is not to have a differently sized header chunk).
 "buzhash" chunker
 +++++++++++++++++
 
-The buzhash chunker triggers (chunks) when the last HASH_MASK_BITS bits of
-the hash are zero, producing chunks of 2^HASH_MASK_BITS Bytes on average.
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks! Guess we want to backport that to 1.1-maint also, because that is the most-read documentation.

Issue #5333, opened 2020-09-12
Unexpected enforced file permissions when the archive is created from STDIN


schors:
I create the archive from STDIN:
`borg create --umask 0022 --compression lz4 --stdin-name dbname.sql '::mysql-dbname-{now:%Y-%m-%dT%H:%M:%S}' -
`
File in the archive has 660 permissions:
```
borg list ::mysql-dbname-2020-09-11T06:25:38
-rw-rw---- root   root    8667267 Fri, 2020-09-11 06:25:41 dbname.sql
```

I tried to set different umask, nothing changed. I created the archive without umask, nothing changed. I did new clean init every time

```
borg -V
borg 1.1.13
```

knutov:
Full steps to reproduce:

```
borg init stdin-test-repo --umask 0022 -e none

echo '123' | borg create --umask 0022 stdin-test-repo::test-001 -

borg list stdin-test-repo::test-001
-rw-rw---- root   root          4 Sat, 2020-09-12 03:46:34 stdin
```

ThomasWaldmann:
If you create from stdin, there is no real file metadata, just (content) data.

But because borg usually stores metadata for each file, it just **makes them up** in case of stdin.

umask only affects file creation and the borg `--umask` option only affects files created by borg in the filesystem (e.g. new segment files in the repo, index files, caches, etc.) - for extracted files, borg always sets the permissions from what it finds in the archive.

So, this is not a bug, but expected behaviour.

If you want to have more control, you can also extract to stdout and then do whatever you like with the stream.

knutov:
@ThomasWaldmann the real problem is reading this file under some non root user in case of `borg mount` with `allow_other` option made under root.

May be it is beter to add option to set file rights with backup from stdin?

schors:
"umask" is not a point. We use "umask" just in case. Permissions is the point. 0660 isn't expected behavior. Why is 0660? It's a weather on the Mars for any cases

The problem is sharing repo to users. By `borg mount` for example

ThomasWaldmann:
0o660 was chosen because it does not include "read for everybody" (which can always be seen as a security issue).

And root/root (uid/gid 0/0) was chosen because that is the only / most simple user that is present on all UNIX-like systems (and also to not cause security issues).

Options you have:
- borg extract to stdout and then redirect to whereever you want
- borg mount with --uid / --gid option?

ThomasWaldmann:
One metadata related option `--stdin-name` already exists.

[...]

PR #5437, opened 2020-10-25
Set file permissions when the archive is created from STDIN

schors:
`--stdin-mode`, `--stdin-user` and `--stdin-group` added
File mode forces to regular file with other values from the option `--stdin-mode`. Default mode is 0o100660.
Error if given user or group do not exists. Default user and group values are taken from system uid/gid 0

PR #5462, opened 2020-11-01
create: implement --stdin-mode, --stdin-user and --stdin-group, #5333

schors:
--stdin-mode, --stdin-user and --stdin-group options added
File mode forces to regular file with other values from the option --stdin-mode. Default mode is 0o100660.
Error occurs if given user or group do not exists. Default user and group values are taken from system uid/gid 0


schors:
Previous PR https://github.com/borgbackup/borg/pull/5437

ThomasWaldmann:
Thanks for the PR! Looks like non-critical to backport to 1.1-maint branch, do you want to do that?

schors:
Yes I do

PR #5463, opened 2020-11-01
Allow EIO with warning when trying to hardlink

d5h:
Hello! Thanks for Borg, it's great!

I ran into an `[Errno 5] Input/output error` when doing `borg init`. It occurs when Borg tries to hardlink `config` to `config.old`. Since the code already ignores (with warning) a few errors from `os.link`, I just added `EIO` (errno 5).

For some more context, I'm running Borg in Docker on Mac. Docker for Mac recently changed the default bind mount filesystem from `osxfs` to `gRPC-FUSE`. It's `gRPC-FUSE` that has this hardlink error. I can workaround by switching Docker back to `osxfs`, but that's slower. So it seems preferable to be able to run Borg on the newer filesystem.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -301,7 +301,7 @@ class Repository:
             try:
                 os.link(config_path, old_config_path)
             except OSError as e:
-                if e.errno in (errno.EMLINK, errno.ENOSYS, errno.EPERM, errno.EACCES, errno.ENOTSUP):
+                if e.errno in (errno.EMLINK, errno.ENOSYS, errno.EPERM, errno.EACCES, errno.ENOTSUP, errno.EIO):
                     logger.warning(link_error_msg)
                 else:
                     raise


ThomasWaldmann:
And you're sure this is not some kind of media error, like it usually is for I/O error?

d5h:
> And you're sure this is not some kind of media error, like it usually is for I/O error?

The media is a FAT32 formatted HDD. I'm sure the HDD works fine, it's just that FAT32 doesn't support links.

I can switch Docker back to using `osxfs` to bind mount this, and I will still get an error when trying to hard link, but in that case the error is `ENOTSUP`.

I'm not sure if you're a Docker for Mac user so maybe the extra filesystems are a bit confusing. I'll give a quick background, and apologize if you already know this. When you bind mount a host volume in Docker for Mac, it introduces its own filesystem on top of the actual host filesystem. This is because Docker for Mac runs Linux in a VM and needs some extra magic, like to pass inotify events from the host to the VM and vice versa. In any case, inside the container, the filesystem is `osxfs` (old Docker) or `gRPC-FUSE` (new Docker), even though the actual filesystem of the media is FAT32.

So basically:

* The actual HDD hardware works fine.
* The HDD filesystem (FAT32) doesn't support links.
* The old Docker filesystem (`osxfs`) gives an errno of `ENOTSUP` when trying to link.
* The new Docker filesystem (`gRPC-FUSE`) gives `EIO`.

Returning `EIO` when trying to hard link is arguably a Docker for Mac bug, so I could open an issue with them. It just seemed to me that if Borg can't create a hard link, it probably shouldn't be a fatal error in any case.

ThomasWaldmann:
Yes, guess you're right. I think you should open an issue with them additional, it is ridiculous if everybody uses a different error code for this. But, as we have to live with that obviously, your changes seems useful.

ThomasWaldmann:
Would you like to backport this to 1.1-maint branch, so it also gets into next borg 1.1.x release?

d5h:
Yes, it would be great to backport this. Thanks!

I opened an issue with Docker for Mac this morning: https://github.com/docker/for-mac/issues/5036. It’s already assigned and they seem to be very active on gRPC-FUSE, so hopefully it will be fixed soon.

ThomasWaldmann:
are you working on the backport still?

d5h:
Sorry, I misunderstood the backport process and didn’t realize I needed to follow up. I can do this later today though. Thanks for the reminder.

PR #5464, opened 2020-11-02
Doc / Deployment with ansible: generic install

Samuel-BF:
I replaced the [pacman module](https://docs.ansible.com/ansible/latest/collections/community/general/pacman_module.html) with the [package module](https://docs.ansible.com/ansible/2.3/package_module.html), so the snippet work for most linuxes, and not only arch.

I also switched state from `latest` to `present` : it is meant to preserve idempotency. This is a [debattable change](https://github.com/ansible/ansible-lint/issues/479).

### DIFF
--- a/docs/deployment/central-backup-server.rst
+++ b/docs/deployment/central-backup-server.rst
@@ -120,11 +120,7 @@ Ansible
 -------
 
 Ansible takes care of all the system-specific commands to add the user, create the
-folder. Even when the configuration is changed the repository server configuration is
-satisfied and reproducible.
-
-Automate setting up an repository server with the user, group, folders and
-permissions a Ansible playbook could be used.
+folder, install and configure software.
 
 ::
 
--- a/docs/deployment/central-backup-server.rst
+++ b/docs/deployment/central-backup-server.rst
@@ -124,9 +124,7 @@ folder. Even when the configuration is changed the repository server configurati
 satisfied and reproducible.
 
 Automate setting up an repository server with the user, group, folders and
-permissions a Ansible playbook could be used. Keep in mind the playbook
-uses the Arch Linux `pacman <https://www.archlinux.org/pacman/pacman.8.html>`_
-package manager to install and keep borg up-to-date.
+permissions a Ansible playbook could be used.
 
 ::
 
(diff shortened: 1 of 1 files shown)


docs/deployment/central-backup-server.rst
```diff
@@ -124,9 +124,7 @@ folder. Even when the configuration is changed the repository server configurati
 satisfied and reproducible.
 
 Automate setting up an repository server with the user, group, folders and
-permissions a Ansible playbook could be used. Keep in mind the playbook
-uses the Arch Linux `pacman <https://www.archlinux.org/pacman/pacman.8.html>`_
-package manager to install and keep borg up-to-date.
+permissions a Ansible playbook could be used.
```
ThomasWaldmann, line 127: not related to your change, but while we're at it:

parse error. can you fix so this to understandable, correct english?

ThomasWaldmann (commented):
can you please change the commit comment, so it contains the reasoning about this change?

maybe also a short summary of that "debatable change" link and why you think your change is an improvement.

Samuel-BF:
Done. The commit comment is now :

>    Doc / Deployment with ansible: pacman -> package
>    
>    Using "package" ansible module make the ansible playbook able to run
>    against most unix OSes. Pacman module only works with Arch and derivatives.
>    
>    Also : changing state from "latest" to "present". Ansible should not be
>    a way to keep your system up-to-date : it's a configuration management
>    system and, as such, should not change anything if neither the playbook
>    nor the machine changed its state (idempotency).
>    
>    Reference : https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/PackageIsNotLatestRule.py#L24

About :
>parse error. can you fix so this to understandable, correct english?

What part was this comment about ? I shortened the ansible introduction, guessing it was about this part (second commit). I'm not a native english speaker/writer, so I'm less accurate in finding this kind of errors :) .

docs/deployment/central-backup-server.rst
ThomasWaldmann: Is maybe this what is meant?:

**To** automate setting up **a** borg repository server **an** Ansible playbook could be used.

ThomasWaldmann:
Thanks! Can you check if we need a backport to 1.1-maint branch?

PR #5465, opened 2020-11-02
create: backport implementation --stdin-mode, --stdin-user and --stdin-group, #5333

schors:
--stdin-mode, --stdin-user and --stdin-group options added
File mode forces to regular file with other values from the option --stdin-mode. Default mode is 0o100660.
Error occurs if given user or group do not exists. Default user and group values are taken from system uid/gid 0


schors:
Ooops. I didn't check integrity

PR #5484, opened 2020-11-04
try macos 11

ThomasWaldmann:


ThomasWaldmann:
looks like macos-11 is always hanging and gets shot after 6h of hanging.

Issue #5439, opened 2020-10-27
Race condition with get_cache_dir()


sophie-h:
My integration tests for borg bindings fail from time to time because there is a race condition here

```py
def get_cache_dir():
…
    # Create path if it doesn't exist yet
    if not os.path.exists(cache_dir):
        os.makedirs(cache_dir)
```

ThomasWaldmann:
What does "fail" mean precisely? Do you get a traceback and if you, can you add it here please?

enkore:
I suppose we could just add exist_ok=True there, after reviewing that nothing around that snippet assumes exclusive access (shouldn't be the case, we got explicit locks for that).

sophie-h:
I thought the race condition was obvious from the code above:

> File \"/app/lib/python3.8/site-packages/borg/cache.py\", line 226, in cache_dir\n    return path or os.path.join(get_cache_dir(), repository.id_str)\n  File \"/app/lib/python3.8/site-packages/borg/helpers.py\", line 561, in get_cache_dir\n    os.makedirs(cache_dir)\n  File \"/usr/lib/python3.8/os.py\", line 223, in makedirs\n    mkdir(name, mode)\nFileExistsError: [Errno 17] File exists: \'/build/.cache/borg\'\n\nPlatform: Linux runner-6yewjmgw-project-11121-concurrent-0 3.10.0-1127.19.1.el7.x86_64 #1 SMP Tue Aug 25 17:23:54 UTC 2020 x86_64\nLinux: Unknown Linux  \nBorg: 1.1.14  Python: CPython 3.8.5 msgpack: 0.5.6\nPID: 7348  CWD: /run/build/pika-backup\nsys.argv: [\'/app/bin/borg\', \'init\', \'--rsh\', \'ssh -o BatchMode=yes\', \'--encryption=repokey\', \'--log-json\', \'--\', \'/tmp/0ddbaa5d-8179-40dc-a71f-b1a5b3cdf048\']

It also happens for config

[...]

PR #5485, opened 2020-11-04
Fixes BORG_CACHE_DIR crashing borg if empty

### LINKED ISSUE #5216: Setting BORG_CACHE_DIR to an empty string crashes borg
Hello,

I've just hit something strange (from the user's point of view) when adding functionality for a helper-script to using borg within backupninja. The feature I was adding was the possibility of displacing the cache dir. For this I set the BORG_CACHE_DIR env var to what users set in the configuration file.

Now the issue that happened was that if somehow the variable BORG_CACHE_DIR ends up empty (and exported as such to the environment), borg uses this empty string as the path to the cache dir and obviously fails, which is confusing for users:

~~~
$ export BORG_CACHE_DIR=
$ borg create ./testborg::blah somesourcedir 
Local Exception
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4529, in main
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4461, in run
    return set_ec(func(args))
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 157, in wrapper
    assert_secure(repository, kwargs['manifest'], self.lock_wait)
  File "/usr/lib/python3/dist-packages/borg/cache.py", line 205, in assert_secure
    sm = SecurityManager(repository)
  File "/usr/lib/python3/dist-packages/borg/cache.py", line 63, in __init__
    self.cache_dir = cache_dir(repository)
  File "/usr/lib/python3/dist-packages/borg/cache.py", line 226, in cache_dir
    return path or os.path.join(get_cache_dir(), repository.id_str)
  File "/usr/lib/python3/dist-packages/borg/helpers.py", line 553, in get_cache_dir
    os.makedirs(cache_dir)
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: ''

Platform: Linux meevyl 5.6.0-1-amd64 #1 SMP Debian 5.6.7-1 (2020-04-29) x86_64
Linux: Unknown Linux  
Borg: 1.1.11  Python: CPython 3.8.3 msgpack: 0.5.6
PID: 990872  CWD: /home/gabster/dev
sys.argv: ['/usr/bin/borg', 'create', './testborg::blah', 'backupninja']
SSH_ORIGINAL_COMMAND: None
~~~

If the env var BORG_CACHE_DIR exists, I would argue that borg should probably add a check to make sure that it is non-empty. if it is, then it should be discarded and the default cache dir placement should be used.

Bounty: https://www.bountysource.com/issues/91660210-setting-borg_cache_dir-to-an-empty-string-crashes-borg

Gu1nness:
This patch introduces a new function that atomically checks if the given
directory exists or not, and if not, tries to fix it.
The error is reraised if necessary with a nice catching.

Fixes #5216.

### DIFF
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -21,7 +21,7 @@ logger = create_logger()
 py_37_plus = sys.version_info >= (3, 7)
 
 
-def ensure_dir(path, name=None, mode=0o777, pretty_deadly=True):
+def ensure_dir(path, mode=stat.S_IRWXU, pretty_deadly=True):
     """
     Ensures that the dir exists with the right permissions.
     1) Make sure the directory exists in a race-free operation
@@ -64,9 +64,8 @@ def get_keys_dir():
     """Determine where to repository keys and cache"""
 
     keys_dir = os.environ.get('BORG_KEYS_DIR', os.path.join(get_config_dir(), 'keys'))
-    if ensure_dir(keys_dir, "key directory", stat.S_IRWXU):
-        logger.debug("Created keys directory : %s" % keys_dir)
-    return keys_dir
+    if ensure_dir(keys_dir):
+        return keys_dir
 
 
 def get_security_dir(repository_id=None):
@@ -74,9 +73,8 @@ def get_security_dir(repository_id=None):
     security_dir = os.environ.get('BORG_SECURITY_DIR', os.path.join(get_config_dir(), 'security'))
     if repository_id:
         security_dir = os.path.join(security_dir, repository_id)
-    if ensure_dir(security_dir, "security directory", stat.S_IRWXU):
-        logger.debug("Created security directory : %s" % security_dir)
-    return security_dir
+    if ensure_dir(security_dir):
+        return security_dir
 
 
 def get_cache_dir():
@@ -89,16 +87,15 @@ def get_cache_dir():
     # Use BORG_CACHE_DIR if set, otherwise assemble final path from cache home path
     cache_dir = os.environ.get('BORG_CACHE_DIR', os.path.join(cache_home, 'borg'))
     # Create path if it doesn't exist yet
-    if ensure_dir(cache_dir, "cache directory", stat.S_IRWXU):
-        logger.debug("Created cache directory : %s" % cache_dir)
-    with open(os.path.join(cache_dir, CACHE_TAG_NAME), 'wb') as fd:
-        fd.write(CACHE_TAG_CONTENTS)
-        fd.write(textwrap.dedent("""
-        # This file is a cache directory tag created by Borg.
-        # For information about cache directory tags, see:
-        #       http://www.bford.info/cachedir/spec.html
-        """).encode('ascii'))
-    return cache_dir
+    if ensure_dir(cache_dir):
+        with open(os.path.join(cache_dir, CACHE_TAG_NAME), 'wb') as fd:
+            fd.write(CACHE_TAG_CONTENTS)
+            fd.write(textwrap.dedent("""
+            # This file is a cache directory tag created by Borg.
+            # For information about cache directory tags, see:
+            #       http://www.bford.info/cachedir/spec.html
+            """).encode('ascii'))
+        return cache_dir
 
 
 def get_config_dir():
@@ -111,9 +108,8 @@ def get_config_dir():
     # Use BORG_CONFIG_DIR if set, otherwise assemble final path from config home path
     config_dir = os.environ.get('BORG_CONFIG_DIR', os.path.join(config_home, 'borg'))
     # Create path if it doesn't exist yet
-    if ensure_dir(config_dir, "config directory", stat.S_IRWXU):
-        logger.debug("Created config directory : %s" % config_dir)
-    return config_dir
+    if ensure_dir(config_dir):
+        return config_dir
 
 
 def dir_is_cachedir(path):
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -7,6 +7,8 @@ import subprocess
 import sys
 import textwrap
 
+from .errors import Error
+
 from .process import prepare_subprocess_env
 from ..platformflags import is_win32
 
@@ -17,6 +19,33 @@ logger = create_logger()
 
 
 py_37_plus = sys.version_info >= (3, 7)
+
+
+def ensure_dir(path, name=None, mode=None, reraise=False):
+    """
+    Ensures that the dir exists with the right permissions.
+    1) Make sure the directory exists in a race-free condition
+    2) If mode is not None and the directory has been created, give the right
+    permissions
+    3) If reraise is True, catch fatal exceptions, reraise them with a pretty
+    message and the full exception output.
+    Returns True if the file has been created and has the right permissions,
+    False otherwise
+    """
+    mode = 0o777 if mode is None else mode
+    try:
+        os.makedirs(path, exist_ok=False)
+    except FileExistsError:
+        return False
+    except FileNotFoundError as e:
+        print("Error : Could not create directory '%s'\n%s" % (name, e))
+        if reraise:
+            raise Error(e)
+    except OSError as e:
+        print("Error : Could not create directory '%s'\n%s" % (name, e))
+        if reraise:
+            raise Error(e)
+    return True
 
 
 def get_base_dir():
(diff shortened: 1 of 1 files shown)


src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
 py_37_plus = sys.version_info >= (3, 7)
 
 
+def ensure_dir(path, name=None, mode=None, reraise=False):
+    """
+    Ensures that the dir exists with the right permissions.
+    1) Make sure the directory exists in a race-free condition
```
ThomasWaldmann, line 27: s/in a race-free condition/using a (race-free|atomic) operation/

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+    2) If mode is not None and the directory has been created, give the right
+    permissions
+    3) If reraise is True, catch fatal exceptions, reraise them with a pretty
+    message and the full exception output.
```
ThomasWaldmann, line 31: "full exception output" sounds a bit like including the traceback, which is not the case.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+    Returns True if the file has been created and has the right permissions,
+    False otherwise
+    """
+    mode = 0o777 if mode is None else mode
```
ThomasWaldmann, line 35: just use an argument default.

the method you used here is only advisable when using mutable data types, like a dict or a list.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+    try:
+        os.makedirs(path, exist_ok=False)
+    except FileExistsError:
+        return False
+    except FileNotFoundError as e:
+        logger.error("Could not create directory '%s'\n%s" % (name, e))
```
ThomasWaldmann, line 41: guess we do not need logging here.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+        if reraise:
+            raise Error(e.args[1])
+    except OSError as e:
+        logger.error("Error : Could not create directory '%s'\n%s" % (name, e))
```
ThomasWaldmann, line 45: guess we do not need logging here.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+    3) If reraise is True, catch fatal exceptions, reraise them with a pretty
+    message and the full exception output.
+    Returns True if the file has been created and has the right permissions,
+    False otherwise
+    """
+    mode = 0o777 if mode is None else mode
+    try:
+        os.makedirs(path, exist_ok=False)
+    except FileExistsError:
+        return False
```
ThomasWaldmann, line 39: in what cases is this used?

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
...
+def ensure_dir(path, name=None, mode=None, reraise=False):
+    """
+    Ensures that the dir exists with the right permissions.
+    1) Make sure the directory exists in a race-free condition
+    2) If mode is not None and the directory has been created, give the right
+    permissions
+    3) If reraise is True, catch fatal exceptions, reraise them with a pretty
+    message and the full exception output.
+    Returns True if the file has been created and has the right permissions,
+    False otherwise
```
ThomasWaldmann, line 33: Maybe we just want to use exceptions (== either not catch them or re-raise them) if something bad happened and no exception if it worked.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,33 @@
 py_37_plus = sys.version_info >= (3, 7)
 
 
+def ensure_dir(path, name=None, mode=None, reraise=False):
```
ThomasWaldmann, line 24: maybe rename the `reraise` to `pretty_deadly=True`.

the main difference of `Error` exception is that it does not include a traceback (thus it is "pretty") and also it usually terminates the application (thus it is "deadly").

Then, in the exception handlers:
```
if pretty_deadly:
    raise Error(...)
else:
    raise
```

src/borg/helpers/fs.py
ThomasWaldmann: also: mode is not used in the code.

ThomasWaldmann (commented):
thanks for the PR!

i mainly commented on the new function, let's first improve that one.

Gu1nness:
Thanks for all these comments, I will make the changes you requested ASAP :)

src/borg/helpers/fs.py
Gu1nness: Hum, I guess if you set BORG_CACHE_DIR to an existing directory.

ThomasWaldmann: Guess rather not. Also, that is the usual case except for the very first call (also: shouldn't exist_ok be True?).

But it could be that some path element happens to be a **file** and thus is in the way.

ThomasWaldmann: https://github.com/borgbackup/borg/issues/5439#issuecomment-717481619

Gu1nness: Indeed exist_true shoud be True

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,37 @@
...
+    except FileNotFoundError as e:
+        if pretty_deadly:
+            raise Error(e.args[1])
+        else:
+            raise
+    except OSError as e:
+        if pretty_deadly:
+            raise Error(e.args[1])
+        else:
+            raise
```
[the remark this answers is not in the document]
Gu1nness: Indeed they are.
Committing this now.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,27 @@
 py_37_plus = sys.version_info >= (3, 7)
 
 
+def ensure_dir(path, mode=stat.S_IRWXU, pretty_deadly=True):
+    """
+    Ensures that the dir exists with the right permissions.
+    1) Make sure the directory exists in a race-free operation
+    2) If mode is not None and the directory has been created, give the right
```
ThomasWaldmann, line 28: according to the python docs, it only assures that the **leaf directory** will get these permissions.

src/borg/helpers/fs.py
```diff
@@ -19,6 +21,37 @@
...
+        if pretty_deadly:
+            raise Error(e.args[1])
+        else:
+            raise
+    except OSError as e:
+        if pretty_deadly:
+            raise Error(e.args[1])
+        else:
+            raise
+    return True
```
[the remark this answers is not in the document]
ThomasWaldmann: seen this?

src/borg/helpers/fs.py
```diff
@@ -41,21 +64,17 @@ def get_keys_dir():
     """Determine where to repository keys and cache"""
 
     keys_dir = os.environ.get('BORG_KEYS_DIR', os.path.join(get_config_dir(), 'keys'))
-    if not os.path.exists(keys_dir):
-        os.makedirs(keys_dir)
-        os.chmod(keys_dir, stat.S_IRWXU)
-    return keys_dir
+    if ensure_dir(keys_dir):
+        return keys_dir
```
ThomasWaldmann, line 68: you don't need the "if". just call ensure_dir and return the keys_dir.

it it does not work, it will blow up with some more or less pretty exception (and by default it is the more pretty one).

same applies to all other calls.

src/borg/helpers/fs.py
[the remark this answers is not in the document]
ThomasWaldmann: btw, the current code is a mixup of 2 fundamentally different ways to deal with error handling:

- C / Go style: return some 0 / 1 / negative / positive / True / False value to signal success / failure. In the worse case, this is even in-band signalling (e.g. most values meaning a measured value and some specific values, like 0, meaning an error).

This is bad and leads to unpretty code structure (tons of if/else and bubbling up of error codes, bugs if the "if" is "forgotten") and in case of in-band signalling also to other problems.

- Python / Java(?) style: do **not** return error codes, but throw meaningful exceptions. If there is e.g. some measurement value, return that (if the measurement succeeded). Exceptions can be catched at any level in the code (or not, if fatal).

Gu1nness:
Sorry I quite forgot about this issue, I will fix the last few things tonight.

src/borg/helpers/fs.py
```diff
@@ -39,7 +39,7 @@ def ensure_dir(path, mode=stat.S_IRWXU, pretty_deadly=True):
             raise Error(e.args[1])
         else:
             raise
-    return True
+    return
```
ThomasWaldmann, line 42: you don't need to have an empty return at the end of a function. if there is nothing left, it'll return anyway.

ThomasWaldmann (commented):
we're almost there!

ThomasWaldmann:
Thanks! Do you want to backport to 1.1-maint?

Gu1nness:
Sure, I'll backport it!

PR #5472, opened 2020-11-03
Use --timestamp in placeholders if given

milkey-mouse:
Fixes #5189. `{now}` and `{utcnow}` will take the value of `--timestamp` if given. I also added a test for this functionality.

Issue #5189, opened 2020-05-17
--timestamp-aware {now}/{utcnow} placeholders


martinvonwittich:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Feature request

#### Your borg version (borg -V).

```
martin@martin ~ % borg -V
borg 1.1.5
```

#### Operating system (distribution) and version.

```
martin@martin ~ % lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic
```

#### Hardware / network configuration, and filesystems used.

shouldn't matter

#### How much data is handled by borg?

shouldn't matter

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
borg create -s --timestamp /path/to/old/backup -C auto,zstd --exclude-caches ./::"{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}" /path/to/old/backup
```

## Describe the problem you're observing.

`{now}`/`{utcnow}` currently refer to the current date and time. This is fine when creating normal backups, but it's not really useful when using the `--timestamp` option to manually set the archive creation date, e.g. when importing old backups:

```
--timestamp TIMESTAMP
       manually specify the archive creation date/time
       (UTC, yyyy-mm-ddThh:mm:ss format).
       Alternatively, give a reference file/directory.
```

For `--timestamp`, it would be useful if `{now}`/`{utcnow}` would either refer to the timestamp as specified by `--timestamp`, or if new placeholders like `{timestamp}` and `{utctimestamp}` were introduced. The new placeholders could then refer the the current time if `--timestamp` wasn't used, and to the referred timestamp otherwise.

Bounty: https://www.bountysource.com/issues/91451358-timestamp-aware-now-utcnow-placeholders

martinvonwittich:
My current workaround is:

```
cd /path/to/old
for i in *; do echo "$i"; borg create --progress -s --timestamp "$i" -C auto,zstd --exclude-caches /path/to/repo/::"import-$i-$(date -r "$i" +"%Y-%m-%dT%H:%M:%S")" "$i"; done
```

ThomasWaldmann:
Guess it would make sense to assign `now` and/or `utcnow` from `--timestamp`.

If we want to do both, guess we'ld need the timezone from the current system.

[...]

PR #5488, opened 2020-11-04
Use --timestamp in placeholders if given (1.1 backport)

milkey-mouse:
See #5189, #5472.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4457,6 +4457,8 @@ class Archiver:
                               self.do_list, self.do_mount, self.do_umount}
             if func not in bypass_allowed:
                 raise Error('Not allowed to bypass locking mechanism for chosen command')
+        if getattr(args, 'timestamp', None):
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
py35 tests failing.

milkey-mouse:
Is there a good reason --timestamp is limited to UTC in the first place? Currently the timestamp, whether given directly or via a file, is a naive datetime. IMO the "best" way to solve this issue is to make them timezone aware.

ThomasWaldmann:
I can't help with that, please find out in the code yourself.

I just want it to be consistent, so implement dealing with timezones in the same way as the code is doing it right now.

milkey-mouse:
I was mostly thinking out loud. TL;DR it works correctly, such that {now} is different from {utcnow} with --timestamp (if you're not in UTC timezone).

ThomasWaldmann:
Thanks for the PR. For remaining issues / ideas, please open a ticket.

PR #5489, opened 2020-11-05
Timestamp fixes

milkey-mouse:
This is a forward-port from my [back-port](https://github.com/borgbackup/borg/pull/5488) of #5472. Making the timestamp value timezone-aware makes `{now}` different from `{utcnow}`, which upon another look I think is what was [originally intended](https://github.com/borgbackup/borg/issues/5189#issuecomment-629865397).

Also, I got rid of a duplicate function that was accidentally added during a refactor.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -88,26 +88,6 @@ def interval(s):
     return hours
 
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
Thanks!

PR #5490, opened 2020-11-07
minor rewrite to make it less confusing (get rid of "Avoid to create a mixup")

matkoniecz:
Original wording was confusing
> Avoid to create a mixup of users and permissions in your repository (or cache).

It is not clear, what should be avoided?

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -71,12 +71,12 @@ Also helpful:
 Important note about permissions
 --------------------------------
 
+To avoid permissions issues (in your borg repository or borg cache), always
+access the repository using the same user account.
+
 Using root likely will be required if you want to backup files of other users
 or the operating system. If you only back up your own files, you neither need
 nor want to use root.
-
-To avoid permissions issues (in your borg repository or borg cache), always
-access the repository using the same user account.
 
 Of course, a non-root user will have no permission to work with the files
 created by root (or another user) and borg operations will just fail with
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -75,9 +75,8 @@ Using root likely will be required if you want to backup files of other users
 or the operating system. If you only back up your own files, you neither need
 nor want to use root.
 
-Avoid to create a mixup of users and permissions in your repository (or cache).
-
-This can easily happen if you run borg using different user accounts (e.g. your
+Mixup of users and permissions in your repository (or cache) is unwanted.
+It will easily happen if you run borg using different user accounts (e.g. your
 non-privileged user and root) while accessing the same repo.
 
 Of course, a non-root user will have no permission to work with the files
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -71,24 +71,20 @@ Also helpful:
 Important note about permissions
 --------------------------------
 
-To avoid permissions issues (in your borg repository or borg cache), always
-access the repository using the same user account.
+To avoid permissions issues (in your borg repository or borg cache), **always
+access the repository using the same user account**.
 
-Using root likely will be required if you want to backup files of other users
-or the operating system. If you only back up your own files, you neither need
-nor want to use root.
+If you want to backup files of other users or the operating system, running
+borg as root likely will be required (otherwise you'ld get `Permission denied`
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -75,9 +75,8 @@ Using root likely will be required if you want to backup files of other users
 or the operating system. If you only back up your own files, you neither need
 nor want to use root.
 
-Avoid to create a mixup of users and permissions in your repository (or cache).
-
-This can easily happen if you run borg using different user accounts (e.g. your
+Mixup of users and permissions in your repository (or cache) is unwanted.
+It will easily happen if you run borg using different user accounts (e.g. your
 non-privileged user and root) while accessing the same repo.
```
ThomasWaldmann, line 80: I don't see the improvement here, can you explain? Are you a native english speaker?

Also: "will happen" is too strong as it also depends on some circumstances.

matkoniecz: > Avoid to create a mixup of users and permissions in your repository (or cache).

is unclear as it is not clear what is supposed to be avoided.

And in addition "Avoid to create a mixup of" is weird, as avoiding this will prevent mixup - not cause it.

"Are you a native english speaker?" no

ThomasWaldmann: "Avoid to do something" is a recommendation not to do something, so it is clear (IMHO) what to avoid.

How exactly this happens is explained in the next sentence, referring back using "This".

ThomasWaldmann: As we are both not native speakers, it would be good if a native speaker could improve the phrasing here, if necessary.

matkoniecz: > "Avoid to do something" is a recommendation not to do something, so it is clear (IMHO) what to avoid.

I think that "Avoid creating a mixup of users" would be correct if such structure was intended (but I am not a native speaker and so on).

ThomasWaldmann: Not sure if that really expresses that (or if you are just "reading between the lines" here).

Maybe one could also say "Be careful not to create a mixup..." and that would express better that it could happen accidentally if one is not careful.

jdchristensen: "Avoid to create" is not correct English.   I also don't like the "a mixup of" part.  Looking at the FAQ item more broadly, it seems too long winded.  The current sentence could say "To avoid permissions issues, always access the repository using the same user account."  And this sentence should probably be the first one in the FAQ answer, followed by information about which account to use (root vs user) and how to achieve it (local vs ssh).

Another comment is that having every sentence as its own paragraph is unusual.  Some could be merged.

ThomasWaldmann:
@matkoniecz did you see feedback of @jdchristensen ? can you update the PR?

matkoniecz:
@ThomasWaldmann Done!

ThomasWaldmann:
@matkoniecz ping.

matkoniecz:
> but if you want to backup the whole system, you need to run borg as root. thus, writing to a local repo would also happen as root. that might be fine, but if another owner is desired for the repo files, there is also the ssh://someuser@localhost/... trick to make all repo files owned by someuser.

Is it impossible to have separate local repos for full system backup and some additional local backups? I understood this that encouraging such separation is the entire part of this documentation section?

(I may be really confused and sentence above may be an utter gibberish. But if that is impossible - how user should setup it?)

ThomasWaldmann:
Sure you can have as many repos as you like. The point is that one should not create a mixup of owners/permissions within a specific repo. See the suggestions of @jdchristensen about how to make this more clear.

matkoniecz:
Sadly, I have no ide ahow to express it better. Feel free to close it or reject or wite own version.

I want just to register that this part of docs is confusing and I apparently do not understand it and it is unclear what remains to be changed.

ThomasWaldmann:
OK, I'll try to fix this PR soon. Thanks for your effort so far!

docs/quickstart.rst
ThomasWaldmann: done

ThomasWaldmann:
I tried to reorder / rephrase as @jdchristensen suggested. Also shortened it a bit without losing important information.

Issue #4102, opened 2018-10-07
add stat-based include/exclude


ThomasWaldmann:
there are tickets about doing file size based exclusion: #902, jborg/attic#330

file size is a stat result attribute, so this is a special case of a stat-based rule.

also in stat result:
- timestamps: atime, ctime, mtime
- type and mode
- uid / gid

so we could add a mechanism to define inclusion / exclusion rules not only based on the file's path/name (as we already have), but also based on comparing stat attributes to given values.

ThomasWaldmann:
If somebody is finding this searching for a solution/workaround for borg 1.0/1.1:
- you can exclude "known-big" files by a name-based pattern, like `*.iso` (or their directory, like `.../Virtualbox VMs/*`).
- you can use `find` unix tool to create a list for `--exclude-from` borg option

Temporarily excluding big files is especially useful for a initial backup(s), which might take a while.

ThomasWaldmann:
Note: the first implementation could just limit the scope to size-based include/exclude (but when writing the code, do it in a way that e.g. timestamp-based can be easily done also).

n-st:
> you can use find unix tool to create a list for --exclude-from borg option

Beware of race conditions, though (i.e. large files appearing after you've generated the list).

russelldavis:
I have a proposal for how this could be implemented. Rather than a global CLI flag like `exclude-by-size`, it could be added as a special [borg-patterns](https://borgbackup.readthedocs.io/en/stable/usage/help.html#borg-patterns) prefix that is applied at the individual pattern level. This would make it quite flexible -- you could apply the rule to certain files/directories only, use it with include patterns, etc.

There's already logic in place to handle prefixes (for `R` and `P`) so adding another one should be simple and backwards compatible. I propose calling it `F` for "filter". The prefix would be followed by a filter-type specifier, any arguments needed for the filter, and finally the pattern to apply the filter to.

So, to exclude files over 100M from Downloads folders, you would write:

```
F size > 100M -/Users/*/Downloads
```

To exclude files over 1G everywhere, you could add this to the command line:
```
borg ... --pattern='F size > 1G -**'
```

For other stat filters, just replace `size` with `mtime`, `mode`, etc.

I have more thoughts, including how to combine multiple filters together, but wanted to put this out there first. What do you think of the proposal? (If it's well received I may take a stab at implementing it.)

ThomasWaldmann:
In the end, guess this will need boolean expressions.

[...]

PR #5492, opened 2020-11-08
Outsource recursive directory walking

Lapin0t:
As discussed on IRC, i'd like to do somewhat complex file matching rules (in particular contextual stuff like: "if directory contains `Cargo.toml` exclude `target`"). Other requests for complex matching are #4102 (stat-based filtering). The proposition is to add a lower-level mode for archive creation where borg would read on stdin for paths which it would add to the archive (still checking if stuff has changed but not recursing into directories). Then arbitrarily complex matching can then be scripted by the user as eg `find ... | borg create-nowalk`.

I'm starting the PR with only a single commit to gather comments, i intend to push more stuff before merge. I refactored `Archiver._process` into `Archiver._rec_walk` (which is functionally just a renaming of `Archiver._process`) and `Archiver._process_any` which does "single path import" logic. All tests seem to pass locally on py38.

An alternative implementation would be making `_rec_walk` an iterator of `(path, parent_fd, name, st)`, which would arguably be cleaner. But maintaining the exact error-handling behavior would've been too complicated: the divergence would've been that if an I/O error happens in `scandir` (`with backup_io('scandir'):` block in `archiver.py`), then only a warning could be printed and not an `E <path>` status line. The reason is that printing status lines would be done from `_process_any` but that error would originate in `_rec_walk`. Having an error handler higher-up wouldn't be possible since raising from the iterator would cancel the full walk. If this behavior change is ok i'd be glad to change my patch (i didn't completely finalize the iterator version so i didn't look if it triggered some tests).

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1203,12 +1203,12 @@ class FilesystemObjectProcessors:
     def process_dir(self, *, path, parent_fd, name, st):
         with self.create_helper(path, st, 'd', hardlinkable=False) as (item, status, hardlinked, hardlink_master):
             with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags_dir,
-                        noatime=True, op='dir_open') as child_fd:
-                # child_fd is None for directories on windows, in that case a race condition check is not possible.
-                if child_fd is not None:
+                        noatime=True, op='dir_open') as fd:
+                # fd is None for directories on windows, in that case a race condition check is not possible.
+                if fd is not None:
                     with backup_io('fstat'):
-                        st = stat_update_check(st, os.fstat(child_fd))
-                item.update(self.metadata_collector.stat_attrs(st, path, fd=child_fd))
+                        st = stat_update_check(st, os.fstat(fd))
+                item.update(self.metadata_collector.stat_attrs(st, path, fd=fd))
                 return status
 
     def process_fifo(self, *, path, parent_fd, name, st):
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1200,6 +1200,11 @@ class FilesystemObjectProcessors:
         if hardlink_master:
             self.hard_links[(st.st_ino, st.st_dev)] = safe_path
 
+    def process_dir_with_fd(self, *, path, fd, st):
+        with self.create_helper(path, st, 'd', hardlinkable=False) as (item, status, hardlinked, hardlink_master):
+            item.update(self.metadata_collector.stat_attrs(st, path, fd=fd))
+            return status
+
     def process_dir(self, *, path, parent_fd, name, st):
         with self.create_helper(path, st, 'd', hardlinkable=False) as (item, status, hardlinked, hardlink_master):
             with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags_dir,
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -38,7 +38,7 @@ from .helpers import StableDict
 from .helpers import bin_to_hex
 from .helpers import safe_ns
 from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
-from .helpers import os_open, flags_normal
+from .helpers import os_open, flags_normal, flags_dir
 from .helpers import msgpack
 from .helpers import sig_int
 from .patterns import PathPrefixPattern, FnmatchPattern, IECommand
@@ -1200,10 +1200,16 @@ class FilesystemObjectProcessors:
         if hardlink_master:
             self.hard_links[(st.st_ino, st.st_dev)] = safe_path
 
-    def process_dir(self, *, path, fd, st):
+    def process_dir(self, *, path, parent_fd, name, st):
         with self.create_helper(path, st, 'd', hardlinkable=False) as (item, status, hardlinked, hardlink_master):
-            item.update(self.metadata_collector.stat_attrs(st, path, fd=fd))
-            return status
+            with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags_dir,
+                        noatime=True, op='dir_open') as child_fd:
+                # child_fd is None for directories on windows, in that case a race condition check is not possible.
+                if child_fd is not None:
+                    with backup_io('fstat'):
+                        st = stat_update_check(st, os.fstat(child_fd))
+                item.update(self.metadata_collector.stat_attrs(st, path, fd=child_fd))
+                return status
 
     def process_fifo(self, *, path, parent_fd, name, st):
         with self.create_helper(path, st, 'f') as (item, status, hardlinked, hardlink_master):  # fifo
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -720,7 +720,7 @@ class Archiver:
             # if restrict_dev is given, we do not want to recurse into a new filesystem,
             # but we WILL save the mountpoint directory (or more precise: the root
             # directory of the mounted filesystem that shadows the mountpoint dir).
-            right_dev = restrict_dev is None or st.st_dev == restrict_dev
+            recurse = restrict_dev is None or st.st_dev == restrict_dev
 
             if self.exclude_nodump:
                 # Ignore if nodump flag is set
@@ -737,31 +737,30 @@ class Archiver:
                         with backup_io('fstat'):
                             st = stat_update_check(st, os.fstat(child_fd))
 
-                    if right_dev:
+                    if recurse:
                         tag_names = dir_is_tagged(path, exclude_caches, exclude_if_present)
                         if tag_names:
                             # if we are already recursing in an excluded dir, we do not need to do anything else than
                             # returning (we do not need to archive or recurse into tagged directories), see #3991:
-                            if recurse_excluded_dir:
-                                return
-                            if keep_exclude_tags:
-                                status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
-                                                      cache=cache, read_special=read_special, dry_run=dry_run)
-                                for tag_name in tag_names:
-                                    tag_path = os.path.join(path, tag_name)
-                                    self._rec_walk(
-                                            path=tag_path, parent_fd=child_fd, name=tag_name, fso=fso, cache=cache,
-                                            matcher=matcher, exclude_caches=exclude_caches, exclude_if_present=exclude_if_present,
-                                            keep_exclude_tags=keep_exclude_tags, skip_inodes=skip_inodes,
-                                            restrict_dev=restrict_dev, read_special=read_special, dry_run=dry_run)
-                            self.print_file_status('x', path)
+                            if not recurse_excluded_dir:
+                                if keep_exclude_tags:
+                                    status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
+                                                          cache=cache, read_special=read_special, dry_run=dry_run)
+                                    for tag_name in tag_names:
+                                        tag_path = os.path.join(path, tag_name)
+                                        self._rec_walk(
+                                                path=tag_path, parent_fd=child_fd, name=tag_name, fso=fso, cache=cache,
+                                                matcher=matcher, exclude_caches=exclude_caches, exclude_if_present=exclude_if_present,
+                                                keep_exclude_tags=keep_exclude_tags, skip_inodes=skip_inodes,
+                                                restrict_dev=restrict_dev, read_special=read_special, dry_run=dry_run)
+                                self.print_file_status('x', path)
                             return
 
                     if not recurse_excluded_dir:
                         status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
                                                    cache=cache, read_special=read_special, dry_run=dry_run)
 
-                    if right_dev:
+                    if recurse:
                         with backup_io('scandir'):
                             entries = helpers.scandir_inorder(path=path, fd=child_fd)
                         for dirent in entries:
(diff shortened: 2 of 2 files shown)


src/borg/archive.py
```diff
@@ -1200,10 +1200,16 @@ def create_helper(self, path, st, status=None, hardlinkable=True):
...
-            item.update(self.metadata_collector.stat_attrs(st, path, fd=fd))
-            return status
+            with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags_dir,
+                        noatime=True, op='dir_open') as child_fd:
+                # child_fd is None for directories on windows, in that case a race condition check is not possible.
+                if child_fd is not None:
+                    with backup_io('fstat'):
+                        st = stat_update_check(st, os.fstat(child_fd))
+                item.update(self.metadata_collector.stat_attrs(st, path, fd=child_fd))
+                return status
```
ThomasWaldmann, line 1212: `... as fd:` - consistency with other handlers, see below.

ThomasWaldmann (commented):
some first feedback, more in next round.

please, for now, do additional commits and do not squash - that is easier for review.

src/borg/archiver.py
```diff
@@ -652,126 +709,80 @@ def _process(self, *, path, parent_fd=None, name=None,
...
-                                    for tag_name in tag_names:
-                                        tag_path = os.path.join(path, tag_name)
-                                        self._process(path=tag_path, parent_fd=child_fd, name=tag_name,
-                                                      fso=fso, cache=cache, matcher=matcher,
-                                                      exclude_caches=exclude_caches, exclude_if_present=exclude_if_present,
-                                                      keep_exclude_tags=keep_exclude_tags, skip_inodes=skip_inodes,
-                                                      restrict_dev=restrict_dev, read_special=read_special, dry_run=dry_run)
-                                self.print_file_status('x', path)
+
+                    tag_names = dir_is_tagged(path, exclude_caches, exclude_if_present)
```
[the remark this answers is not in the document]
Lapin0t: Nice catch, i even forgot to do the check on the recursive scandir block a couple lines lower.

src/borg/archiver.py
```diff
@@ -652,126 +709,80 @@ def _process(self, *, path, parent_fd=None, name=None,
...
         if status == 'C':
             self.print_warning('%s: file changed while we backed it up', path)
-        # Status output
         if status is None:
-            if not dry_run:
-                status = '?'  # need to add a status code somewhere
-            else:
-                status = '-'  # dry run, item was not backed up
-
+            status = '?'
```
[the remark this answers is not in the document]
Lapin0t: agreed

src/borg/archive.py
Lapin0t: agreed

src/borg/archiver.py
```diff
@@ -652,126 +709,82 @@ def _process(self, *, path, parent_fd=None, name=None,
...
+                                status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
+                                                      cache=cache, read_special=read_special, dry_run=dry_run)
+                                for tag_name in tag_names:
+                                    tag_path = os.path.join(path, tag_name)
+                                    self._rec_walk(
+                                            path=tag_path, parent_fd=child_fd, name=tag_name, fso=fso, cache=cache,
+                                            matcher=matcher, exclude_caches=exclude_caches, exclude_if_present=exclude_if_present,
+                                            keep_exclude_tags=keep_exclude_tags, skip_inodes=skip_inodes,
+                                            restrict_dev=restrict_dev, read_special=read_special, dry_run=dry_run)
+                            self.print_file_status('x', path)
```
ThomasWaldmann, line 757: did you just do 745/746 to get out more quickly?


if so and there is no functional change, please keep the old way as it is easier to review for now (smaller diff).

src/borg/archiver.py
```diff
@@ -652,126 +709,82 @@ def _process(self, *, path, parent_fd=None, name=None,
...
-                return
-            else:
-                self.print_warning('Unknown file type: %s', path)
-                return
+                            self._rec_walk(
+                                    path=normpath, parent_fd=child_fd, name=dirent.name, fso=fso, cache=cache, matcher=matcher,
+                                    exclude_caches=exclude_caches, exclude_if_present=exclude_if_present,
+                                    keep_exclude_tags=keep_exclude_tags, skip_inodes=skip_inodes, restrict_dev=restrict_dev,
+                                    read_special=read_special, dry_run=dry_run)
+            elif right_dev:
```
ThomasWaldmann, line 774: as this is never a directory, doesn't have imply it is always the right dev (or we will not even get here)?

a directory change is always at the directory mounted over the mountpoint.

src/borg/archiver.py
```diff
@@ -652,126 +709,82 @@ def _process(self, *, path, parent_fd=None, name=None,
...
 
             if (st.st_ino, st.st_dev) in skip_inodes:
                 return
             # if restrict_dev is given, we do not want to recurse into a new filesystem,
             # but we WILL save the mountpoint directory (or more precise: the root
             # directory of the mounted filesystem that shadows the mountpoint dir).
-            recurse = restrict_dev is None or st.st_dev == restrict_dev
-            status = None
+            right_dev = restrict_dev is None or st.st_dev == restrict_dev
+
```
ThomasWaldmann, line 724: maybe do not rename variables for now, create a smaller, easier to review diff.

if it is important, this can be done afterwards also.

src/borg/archiver.py
```diff
@@ -652,126 +709,82 @@ def _process(self, *, path, parent_fd=None, name=None,
...
-                    if not dry_run:
-                        if not recurse_excluded_dir:
-                            status = fso.process_dir(path=path, fd=child_fd, st=st)
-                    if recurse:
+
+                    if not recurse_excluded_dir:
+                        status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
+                                                   cache=cache, read_special=read_special, dry_run=dry_run)
+
+                    if right_dev:
```
ThomasWaldmann, line 764: at this place, i find recurse an even better name than right_dev.

if it is true, we scan the directory (== recurse). wanting or not wanting to recurse could also have other reasons than it being the right device or not.

ThomasWaldmann (commented):
another round of feedback, more later :)

src/borg/archiver.py
Lapin0t: I think i did it mainly to avoid deep nesting. It's definitely style, i'll revert that.

src/borg/archiver.py
Lapin0t: Well i just checked and linux `mount --bind` supports mounting non-directories. I'll revert it and prepare another clean patch because i believe there's an issue to be addressed there (we should probably save mount points in general, eg save only the `stat`s).

src/borg/archiver.py
Lapin0t: i'll revert

src/borg/archiver.py
Lapin0t: For my defense: `recurse` isn't really explicit about what makes that variable be true/false, whereas the fact that it's used to decide on recursion should be obvious from the surrounding of it's use. Also i started using the value of `right_dev` with an interpretation unrelated to recursing or not (like line 764, arguable line )... Anyway you're gonna read this code more than me, i'll do whatever you say!

src/borg/archiver.py
```diff
@@ -664,114 +721,69 @@ def _process(self, *, path, parent_fd=None, name=None,
...
-            elif stat.S_ISDIR(st.st_mode):
+
+            if stat.S_ISDIR(st.st_mode):
                 with OsOpen(path=path, parent_fd=parent_fd, name=name, flags=flags_dir,
                             noatime=True, op='dir_open') as child_fd:
                     # child_fd is None for directories on windows, in that case a race condition check is not possible.
                     if child_fd is not None:
                         with backup_io('fstat'):
                             st = stat_update_check(st, os.fstat(child_fd))
+
```
ThomasWaldmann, line 739: if you added the empty line accidentally, please remove it again.

ThomasWaldmann (commented):
much better already! :-)

as this is a critical change and we want to avoid any regression, an easy-to-review diff is important here.

src/borg/archiver.py
ThomasWaldmann: thanks

src/borg/archiver.py
ThomasWaldmann: are you sure?

it didn't work for me (i used `/rw/config/rc.local` just because it is a regular file on a different filesystem):
```
$ sudo mount --bind /rw/config/rc.local foobar
mount: mount point foobar does not exist
$ mkdir foobar
$ sudo mount --bind /rw/config/rc.local foobar
mount: mount(2) failed: /home/user/foobar: Not a directory
```

src/borg/archiver.py
```diff
@@ -664,114 +721,69 @@ def _process(self, *, path, parent_fd=None, name=None,
...
+                                    read_special=read_special, dry_run=dry_run)
             else:
-                self.print_warning('Unknown file type: %s', path)
-                return
+                # everything non-directory goes is this branch
+                # note that directories have to be handled specifically just because they can be
+                # excluded based on tag files they might contain
+                status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso, cache=cache,
+                                           read_special=read_special, dry_run=dry_run)
+
```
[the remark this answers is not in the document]
Lapin0t: Yes, the diff would be nicer, i was kind of bothered by that but before the reviews this was an if/elif and not an if/else so inversion would've made some checking duplication, so i didn't do it

src/borg/archiver.py
```diff
@@ -652,10 +709,10 @@ def _process(self, *, path, parent_fd=None, name=None,
                 # could trigger an error, e.g. if access is forbidden, see #3209.
                 if not matcher.recurse_dir:
                     return
+                recurse_excluded_dir = True
                 with backup_io('stat'):
                     st = os_stat(path=path, parent_fd=parent_fd, name=name, follow_symlinks=False)
-                recurse_excluded_dir = stat.S_ISDIR(st.st_mode)
-                if not recurse_excluded_dir:
+                if not stat.S_ISDIR(st.st_mode):
                     return
```
ThomasWaldmann, line 716: note for other potential reviewers: this is just a style fix, functionality is the same.

src/borg/archiver.py
```diff
@@ -687,91 +747,41 @@ def _process(self, *, path, parent_fd=None, name=None,
                             # if we are already recursing in an excluded dir, we do not need to do anything else than
                             # returning (we do not need to archive or recurse into tagged directories), see #3991:
                             if not recurse_excluded_dir:
-                                if keep_exclude_tags and not dry_run:
-                                    fso.process_dir(path=path, fd=child_fd, st=st)
+                                if keep_exclude_tags:
+                                    status = self._process_any(path=path, parent_fd=parent_fd, name=name, st=st, fso=fso,
+                                                          cache=cache, read_special=read_special, dry_run=dry_run)
```
ThomasWaldmann, line 752: notable change:

we have locally opened (see `OsOpen` call here) the directory and have `child_fd` for it. the old code used that `child_fd` to process the directory, while the new code re-opens the directory again (see `OsOpen` call in `process_dir`).

guess we could also pass `fd=child_fd` to `process_any` and finally to `process_dir` to
- avoid a race condition here
- maybe also being useful for processing of stdin?

ThomasWaldmann:
@enkore @textshell can you review this, too, please?

I guess it is a very good change we need for multiple reasons, we just want to be very careful and avoid any regression.

src/borg/archiver.py
Lapin0t: Actually there's no real reason for this, i should revert this snippet (adding a `fso.process_dir_with_fd` function that is the same as the old `fso.process_dir`). It would've only been worth if there wasn't the tag-file exclusion thing (then `self._process_any` could've been called only once without special-casing directories).

Lapin0t: I pushed the revert without waiting for answer, will undo if you think it was better before. It does duplicate process_dir, but having it that way is more orthogonal and will be used for my "read-paths-from-stdin".

Lapin0t:
Now that the hard refactor part seems to be mostly OK i finalized my feature patch. I'm not adding it here until you ask me to, to not clutter the review, but it's [there](https://github.com/borgbackup/borg/commit/b725b0e17165705d63f1707769fb19fa7f23e1dc).

ThomasWaldmann (approved):
looks correct to me!

ThomasWaldmann:
The py36 CI job hanged for ~2h, the other succeeded.

I restarted them all (on can not restart a single job).

src/borg/archiver.py
ThomasWaldmann: looks good.

the reading from stdin needs some special casing anyway because it is not a fs file, so we can't read attrs / xattrs / etc. from the filesystem

src/borg/archiver.py
Lapin0t: You have to mount it onto a file. Seems kinda arbitrary: I didn't try too hard but apparently everything other than a directory can be bind-mounted onto a file. I guess the only way to know for sure would be to look at the kernel source, it's not gonna be documented anywhere.

ThomasWaldmann:
Added this to 1.1.15 milestone:

If we can be sure enough that this does not break anything, I guess getting in this + the followup "read paths from stdin" would be a super useful feature for power users.

enkore (approved):

ThomasWaldmann:
ok, guess @textshell has no time currently, so i'll merge this.

ThomasWaldmann:
Thanks for the PR, merged! Can you make a PR with your changes that are based on this?

PR #5501, opened 2020-11-13
Update the documentation on hacked backup client.

### LINKED ISSUE #5480: FAQ points to closed issue for pull backups
https://borgbackup.readthedocs.io/en/latest/faq.html#how-can-i-protect-against-a-hacked-backup-client says:

> # How can I protect against a hacked backup client?
> 
> Assume you backup your backup client machine C to the backup server S and C gets hacked. In a simple push setup, the attacker could then use borg on C to delete all backups residing on S.
> 
> These are your options to protect against that:
> 
> - Do not allow to permanently delete data from the repo, see Append-only mode (forbid compaction).
> - Use a pull-mode setup using ssh -R, see #900.
> - Mount C’s filesystem on another machine and then create a backup of it.
> - Do not give C filesystem-level access to S.

That's great except #900 is a kind of long issue that was closed because BountySource and chaos. Maybe it would be best to point to some other documentation for pull-mode? Maybe something about #5288 which I believe made https://borgbackup.readthedocs.io/en/latest/deployment/pull-backup.html possible?

Gu1nness:
Fixes #5480.

### DIFF
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -1,5 +1,6 @@
 .. include:: ../global.rst.inc
 .. highlight:: none
+.. _pull_backup:
 
 =======================
 Backing up in pull mode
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Thanks!

anarcat:
thanks!

PR #5503, opened 2020-11-13
Add a --list switch to borg delete

### LINKED ISSUE #5116: Inconsistent switch to display changes during 'dry-run' operation
There are two operations I use, so far, to destructively modify a repository:
_delete_ and _recreate_

They allow wildcards, patterns, etc.  Obviously this is a dangerous operation and therefore the '--dry-run' switch exists for both to check your work before execution.

I have found an inconsistency in the method of displaying the proposed (dry-run) changes in order to review them however.

When I use _recreate_, the switches are "--dry-run --list" which will output what is going to happen.

However when using _delete_  the same process of "--dry-run --list" results in error:
_borg: error: unrecognized arguments: --list_

When using _delete_ , instead (through trial and error) I found out that instead I have to use:
"--dry-run --verbose"

My issues are:
1) this is not consistent switch usage for the listing of changes among operations that are relatively similar/destructive/deleting/changing; I think '--list' should function for both; (addendum: '--list' is also used with _create_ to likewise show added files and directories)

2) '--verbose' is generally viewed in a debugging light and should not be used in this context to display the desired output of a '--dry-run'; (It can still do that, but I think the 'what will change with this command' should be primarily on the '--list' command. verbose might include other info as well);

3) if '--verbose' is kept as the command to review results of a _delete_ dry-run it is not listed on the documentation page, (whereas the _recreate_ documentation page has '--list' right next to the '--dry-run' switch so a user can immediately grasp what to do.)  I suggest adding it under --dry-run switch to clarify how to see proposed deletions:
"_--verbose_      output list of items deleted (or to delete with --dry-run)"

Gu1nness:
Also add some documentation on this.

Fixes #5116.

### DIFF
--- a/docs/usage/delete.rst
+++ b/docs/usage/delete.rst
@@ -12,7 +12,6 @@ Examples
     # delete all archives whose names begin with the machine's hostname followed by "-"
     $ borg delete --prefix '{hostname}-' /path/to/repo
 
-
     # delete all archives whose names contain "-2012-"
     $ borg delete --glob-archives '*-2012-*' /path/to/repo
 
--- a/docs/usage/delete.rst
+++ b/docs/usage/delete.rst
@@ -16,7 +16,6 @@ Examples
     $ borg delete --glob-archives '*-2012-*' /path/to/repo
 
     # see what would be deleted if delete was run without --dry-run
-    $ borg delete -v --dry-run -a '*-May-*' /path/to/repo
     $ borg delete --list --dry-run -a '*-May-*' /path/to/repo
 
     # delete the whole repository and the related local cache:
--- a/docs/usage/delete.rst
+++ b/docs/usage/delete.rst
@@ -12,11 +12,13 @@ Examples
     # delete all archives whose names begin with the machine's hostname followed by "-"
     $ borg delete --prefix '{hostname}-' /path/to/repo
 
+
     # delete all archives whose names contain "-2012-"
     $ borg delete --glob-archives '*-2012-*' /path/to/repo
 
     # see what would be deleted if delete was run without --dry-run
     $ borg delete -v --dry-run -a '*-May-*' /path/to/repo
+    $ borg delete --list --dry-run -a '*-May-*' /path/to/repo
 
     # delete the whole repository and the related local cache:
     $ borg delete /path/to/repo
--- a/docs/usage/delete.rst.inc
+++ b/docs/usage/delete.rst.inc
@@ -68,7 +68,6 @@ borg delete
 
     optional arguments
         -n, --dry-run    do not change repository
-        --list          list the archives/files which would be deleted
         -s, --stats     print statistics for the deleted archive
         --cache-only    delete only the local cache for the given repository
         --force         force deletion of corrupted archives, use ``--force --force`` in case ``--force`` does not work.
--- a/docs/usage/delete.rst.inc
+++ b/docs/usage/delete.rst.inc
@@ -113,4 +113,3 @@ two options are mutually exclusive.
 To avoid accidentally deleting archives, especially when using glob patterns,
 it might be helpful to use the ``--dry-run`` to test out the command without
 actually making any changes to the repository.
-
--- a/docs/usage/delete.rst.inc
+++ b/docs/usage/delete.rst.inc
@@ -68,6 +68,7 @@ borg delete
 
     optional arguments
         -n, --dry-run    do not change repository
+        --list          list the archives/files which would be deleted
         -s, --stats     print statistics for the deleted archive
         --cache-only    delete only the local cache for the given repository
         --force         force deletion of corrupted archives, use ``--force --force`` in case ``--force`` does not work.
@@ -113,3 +114,4 @@ two options are mutually exclusive.
 To avoid accidentally deleting archives, especially when using glob patterns,
 it might be helpful to use the ``--dry-run`` to test out the command without
 actually making any changes to the repository.
+
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1245,7 +1245,7 @@ class Archiver:
             else:
                 if self.output_list:
                     msg.append("You requested to completely DELETE the repository *including* all archives it "
-                           "contains:")
+                               "contains:")
                     for archive_info in manifest.archives.list(sort_by=['ts']):
                         msg.append(format_archive(archive_info))
                 else:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3523,7 +3523,7 @@ class Archiver:
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
         subparser.add_argument('--list', dest='output_list', action='store_true',
-                               help='output verbose list of archives or print repository')
+                               help='output verbose list of archives')
         subparser.add_argument('-s', '--stats', dest='stats', action='store_true',
                                help='print statistics for the deleted archive')
         subparser.add_argument('--cache-only', dest='cache_only', action='store_true',
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1172,7 +1172,7 @@ class Archiver:
 
         if args.forced == 2:
             deleted = False
-            _logger = logging.getLogger('borg.output.list') if self.output_list else logger
+            logger_list = logging.getLogger('borg.output.list')
             for i, archive_name in enumerate(archive_names, 1):
                 try:
                     current_archive = manifest.archives.pop(archive_name)
@@ -1182,7 +1182,11 @@ class Archiver:
                 else:
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    _logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
+                    if output_list:
+                        logger_list.info(msg.format(format_archive(current_archive),
+                                                    i, len(archive_names)))
+                    logger.info(msg.format(format_archive(current_archive),
+                                                    i, len(archive_names)))
 
             if dry_run:
                 logger.info('Finished dry-run.')
@@ -1199,14 +1203,16 @@ class Archiver:
         with Cache(repository, key, manifest, progress=args.progress, lock_wait=self.lock_wait) as cache:
             msg_delete = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
             msg_not_found = 'Archive {} not found ({}/{}).'
-            _logger = logging.getLogger('borg.output.list') if self.output_list else logger
+            logger_list = logging.getLogger('borg.output.list')
             for i, archive_name in enumerate(archive_names, 1):
                 try:
                     archive_info = manifest.archives[archive_name]
                 except KeyError:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
-                    _logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                if self.output_list:
+                    logger_list.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
 
                     if not dry_run:
                         archive = Archive(repository, key, manifest, archive_name, cache=cache,
@@ -1241,13 +1247,13 @@ class Archiver:
                 msg.append("This repository seems to have no manifest, so we can't tell anything about its "
                            "contents.")
             else:
-                msg.append("You requested to completely DELETE the repository *including* all archives it "
-                           "contains:")
                 if self.output_list:
+                    msg.append("You requested to completely DELETE the repository *including* all archives it "
+                           "contains:")
                     for archive_info in manifest.archives.list(sort_by=['ts']):
                         msg.append(format_archive(archive_info))
                 else:
-                    msg.append("%d archives would be deleted" % len(manifest.archives))
+                    msg.append("%d archives are going to be deleted." % len(manifest.archives))
             msg.append("Type 'YES' if you understand this and want to continue: ")
             msg = '\n'.join(msg)
             if not yes(msg, false_msg="Aborting.", invalid_msg='Invalid answer, aborting.', truish=('YES',),
@@ -1260,8 +1266,8 @@ class Archiver:
                 if not keep_security_info:
                     SecurityManager.destroy(repository)
             else:
-                logger.info("would delete repository.")
-                logger.info("would %s security info." % ("keep" if keep_security_info else "delete"))
+                logger.info("Would delete repository.")
+                logger.info("Would %s security info." % ("keep" if keep_security_info else "delete"))
         if not dry_run:
             Cache.destroy(repository)
             logger.info("Cache deleted.")
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1144,6 +1144,7 @@ class Archiver:
         """Delete an existing repository or archives"""
         archive_filter_specified = any((args.first, args.last, args.prefix is not None, args.glob_archives))
         explicit_archives_specified = args.location.archive or args.archives
+        self.output_list = args.output_list
         if archive_filter_specified and explicit_archives_specified:
             self.print_error('Mixing archive filters and explicitly named archives is not supported.')
             return self.exit_code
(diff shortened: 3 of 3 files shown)


docs/usage/delete.rst
```diff
@@ -12,11 +12,13 @@ Examples
     # delete all archives whose names begin with the machine's hostname followed by "-"
     $ borg delete --prefix '{hostname}-' /path/to/repo
 
+
```
ThomasWaldmann, line 15: please remove accidentally added whitespace.

docs/usage/delete.rst.inc
```diff
@@ -68,6 +68,7 @@ borg delete
 
     optional arguments
         -n, --dry-run    do not change repository
+        --list          list the archives/files which would be deleted
```
ThomasWaldmann, line 71: read line 1 of this file - no need to commit this and make the PR review more work, all these files will be auto-generated before a release.

src/borg/archiver.py
```diff
@@ -1180,7 +1181,10 @@ def _delete_archives(self, args, repository):
                 else:
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
+                    if self.output_list:
+                        logging.getLogger('borg.output.list').info(msg.format(format_archive(current_archive), i, len(archive_names)))
+                    else:
+                        logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
```
ThomasWaldmann, line 1187: if only the logger is different, how about:
```
_logger = logging.getLogger('borg.output.list') if self.output_list else logger
_logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
```

src/borg/archiver.py
```diff
@@ -1202,7 +1206,10 @@ def _delete_archives(self, args, repository):
                 except KeyError:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
-                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    if self.output_list:
+                        logging.getLogger('borg.output.list').info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    else:
+                        logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
```
ThomasWaldmann, line 1212: same

src/borg/archiver.py
```diff
@@ -3509,6 +3523,8 @@ def define_borg_mount(parser):
         subparser.set_defaults(func=self.do_delete)
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
+        subparser.add_argument('--list', dest='output_list', action='store_true',
+                               help='output verbose list of items (files, dirs, ...)')
```
ThomasWaldmann, line 3527: in this case, it is rather a list of archives, right?

src/borg/archiver.py
```diff
@@ -3486,6 +3497,9 @@ def define_borg_mount(parser):
         with the ``--cache-only`` option, or keep the security info with the
         ``--keep-security-info`` option.
 
+        When in doubt, use ``--dry-run --verbose --list`` to see how patterns/PATHS are
+        interpreted. See :ref:`list_item_flags` in ``borg create`` for details.
```
ThomasWaldmann, line 3501: when --list is used, does it really also require --verbose?

please check for delete and also for other borg subcommands and do it in a consistent way.

src/borg/archiver.py
```diff
@@ -1252,8 +1259,12 @@ def _delete_repository(self, args, repository):
...
                     SecurityManager.destroy(repository)
             else:
-                logger.info("Would delete repository.")
-                logger.info("Would %s security info." % ("keep" if keep_security_info else "delete"))
+                if self.output_list:
+                    logging.getLogger('borg.output.list').info("Would delete repository.")
+                    logging.getLogger('borg.output.list').info("Would %s security info." % ("keep" if keep_security_info else "delete"))
+                else:
+                    logger.info("Would delete repository.")
+                    logger.info("Would %s security info." % ("keep" if keep_security_info else "delete"))
```
ThomasWaldmann, line 1267: guess this is a bit questionable regarding whether this counts as "list".

as borg always operates on 1 repo, generic hints about this repo and the related cache are not "listing" anything.

only when it outputs all archives in that repo, that would be "listing".

src/borg/archiver.py
ThomasWaldmann: also, the item_flags are not relevant for borg delete (ARCHIVE|REPO).

docs/usage/delete.rst.inc
```diff
@@ -112,4 +112,5 @@ two options are mutually exclusive.
 
 To avoid accidentally deleting archives, especially when using glob patterns,
 it might be helpful to use the ``--dry-run`` to test out the command without
-actually making any changes to the repository.
\ No newline at end of file
+actually making any changes to the repository.
```
ThomasWaldmann: maybe do git checkout thatfile to revert it to the original state so it disappears from the PR.

src/borg/archiver.py
```diff
@@ -3509,6 +3521,8 @@ def define_borg_mount(parser):
         subparser.set_defaults(func=self.do_delete)
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
+        subparser.add_argument('--list', dest='output_list', action='store_true',
+                               help='output verbose list of archives or repositories')
```
ThomasWaldmann, line 3525: guess we do not have the repositories (plural) case, right? and "listing" just 1 does not make much sense, as this is obviously already known to the user.

src/borg/archiver.py
Gu1nness: Hum, then I don't know which logger to use.
If I use the info level, there is no difference between the --list and nothing for the repositories (which can be a consistent way of acting). To be sure that something is written I could use the warn level.

ThomasWaldmann:
I have reviewed the current master branch code about how `--list` is currently used:
- it is typically used inside a loop, to output file names or archive names
- it typically generates **additional** output that is not generated when `--list` is not given

Typical use:
```
            if output_list:
                logging.getLogger('borg.output.list').info(...)
```
Note:
- there is no `else`.
- there is no switching between different loggers
- no non-loop output

src/borg/archiver.py
Gu1nness: No I don't think it requires --verbose, however it is present by default in all subparsers due to `define_common_options` but I can silently forget about it if --list is present.
On `borg create`, `--verbose` does nothing interesting, so this might be what we want.

src/borg/archiver.py
Gu1nness: @ThomasWaldmann What's your idea on this?
Should I use the warn level to be sure to print something?

src/borg/archiver.py
```diff
@@ -1202,7 +1205,9 @@ def _delete_archives(self, args, repository):
                 except KeyError:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
-                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    _logger = logging.getLogger('borg.output.list') if self.output_list else logger
+                    _logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
```
ThomasWaldmann, line 1209: same here with 1202.

src/borg/archiver.py
ThomasWaldmann: this is not a list, so no change is needed. giving --list or not should not make a difference here.

1244-1247 is the "listy" stuff here:
- with --list: like now
- without --list: just tell that "N archives are going to get deleted (use --list for details)"

src/borg/archiver.py
```diff
@@ -3486,6 +3495,9 @@ def define_borg_mount(parser):
         with the ``--cache-only`` option, or keep the security info with the
         ``--keep-security-info`` option.
 
+        When in doubt, use ``--dry-run --list`` to see how patterns/PATHS are
```
ThomasWaldmann, line 3498: what is patterns/PATHs referring to?

src/borg/archiver.py
```diff
@@ -3509,6 +3521,8 @@ def define_borg_mount(parser):
         subparser.set_defaults(func=self.do_delete)
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
+        subparser.add_argument('--list', dest='output_list', action='store_true',
+                               help='output verbose list of archives or print repository')
```
ThomasWaldmann, line 3525: fix this when the rest is finished.

docs/usage/delete.rst
```diff
@@ -17,6 +17,7 @@ Examples
 
     # see what would be deleted if delete was run without --dry-run
     $ borg delete -v --dry-run -a '*-May-*' /path/to/repo
+    $ borg delete --list --dry-run -a '*-May-*' /path/to/repo
```
ThomasWaldmann, line 20: fix this when the rest is finished.

src/borg/archiver.py
```diff
@@ -1243,8 +1243,11 @@ def _delete_repository(self, args, repository):
             else:
                 msg.append("You requested to completely DELETE the repository *including* all archives it "
                            "contains:")
-                for archive_info in manifest.archives.list(sort_by=['ts']):
-                    msg.append(format_archive(archive_info))
+                if self.output_list:
+                    for archive_info in manifest.archives.list(sort_by=['ts']):
+                        msg.append(format_archive(archive_info))
+                else:
+                    msg.append("%d archives would be deleted" % len(manifest.archives))
```
ThomasWaldmann, line 1250: The "else" msg now does not match the 1244 msg nicely any more.

src/borg/archiver.py
```diff
@@ -1171,6 +1172,7 @@ def _delete_archives(self, args, repository):
 
         if args.forced == 2:
             deleted = False
+            _logger = logging.getLogger('borg.output.list') if self.output_list else logger
```
ThomasWaldmann, line 1175: well, i recommended this if/else some reviews ago, but meanwhile i noticed that the --list output is just *additional* output going to that logger *only* (and no output in the other case).

see other places where it uses borg.output.list.

src/borg/archiver.py
```diff
@@ -1196,13 +1199,15 @@ def _delete_archives(self, args, repository):
         with Cache(repository, key, manifest, progress=args.progress, lock_wait=self.lock_wait) as cache:
             msg_delete = 'Would delete archive: {} ({}/{})' if dry_run else 'Deleting archive: {} ({}/{})'
             msg_not_found = 'Archive {} not found ({}/{}).'
+            _logger = logging.getLogger('borg.output.list') if self.output_list else logger
```
ThomasWaldmann, line 1202: same here.

src/borg/archiver.py
```diff
@@ -1252,8 +1260,8 @@ def _delete_repository(self, args, repository):
                 if not keep_security_info:
                     SecurityManager.destroy(repository)
             else:
-                logger.info("Would delete repository.")
-                logger.info("Would %s security info." % ("keep" if keep_security_info else "delete"))
+                logger.info("would delete repository.")
+                logger.info("would %s security info." % ("keep" if keep_security_info else "delete"))
```
ThomasWaldmann, line 1264: did you unintentionally lowercase the "w" of "Would"?

src/borg/archiver.py
```diff
@@ -1182,7 +1182,11 @@ def _delete_archives(self, args, repository):
                 else:
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    _logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
+                    if output_list:
+                        logger_list.info(msg.format(format_archive(current_archive),
+                                                    i, len(archive_names)))
+                    logger.info(msg.format(format_archive(current_archive),
+                                                    i, len(archive_names)))
```
ThomasWaldmann, line 1189: you need to kill the 2nd log call or it will be duplicate if you give `--list`.

src/borg/archiver.py
```diff
@@ -1199,14 +1203,16 @@ def _delete_archives(self, args, repository):
...
             for i, archive_name in enumerate(archive_names, 1):
                 try:
                     archive_info = manifest.archives[archive_name]
                 except KeyError:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
-                    _logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                if self.output_list:
+                    logger_list.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
```
ThomasWaldmann: you need to remove the 1st log call and indent the 2nd call (so it is within else block).

src/borg/archiver.py
```diff
@@ -1241,13 +1247,13 @@ def _delete_repository(self, args, repository):
...
-                msg.append("You requested to completely DELETE the repository *including* all archives it "
-                           "contains:")
                 if self.output_list:
+                    msg.append("You requested to completely DELETE the repository *including* all archives it "
+                           "contains:")
                     for archive_info in manifest.archives.list(sort_by=['ts']):
                         msg.append(format_archive(archive_info))
                 else:
-                    msg.append("%d archives would be deleted" % len(manifest.archives))
+                    msg.append("%d archives are going to be deleted." % len(manifest.archives))
```
ThomasWaldmann, line 1256: How about:
```
You requested to completely DELETE the repository *including* %d archives it contains."
```

src/borg/archiver.py
```diff
@@ -1180,7 +1182,13 @@ def _delete_archives(self, args, repository):
...
                     deleted = True
                     msg = 'Would delete: {} ({}/{})' if dry_run else 'Deleted archive: {} ({}/{})'
-                    logger.info(msg.format(format_archive(current_archive), i, len(archive_names)))
+                    if output_list:
+                        logger_list.info(msg.format(format_archive(current_archive),
+                                                    i, len(archive_names)))
+                    else:
+                        logger.info(msg.format(format_archive(current_archive),
+                                               i, len(archive_names)))
+
```
ThomasWaldmann, line 1191: this is still outputting the (same) list even when not giving `--list`.

and while you're at it: if `msg` is only used in the if-block, move it there.

ThomasWaldmann: also: is it `self.output_list` ?

src/borg/archiver.py
```diff
@@ -1196,13 +1204,15 @@ def _delete_archives(self, args, repository):
...
             msg_not_found = 'Archive {} not found ({}/{}).'
+            logger_list = logging.getLogger('borg.output.list')
             for i, archive_name in enumerate(archive_names, 1):
                 try:
                     archive_info = manifest.archives[archive_name]
                 except KeyError:
                     logger.warning(msg_not_found.format(archive_name, i, len(archive_names)))
                 else:
-                    logger.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
+                    logger_list.info(msg_delete.format(format_archive(archive_info), i, len(archive_names)))
```
ThomasWaldmann, line 1214:
```
if self.output_list:
    logger_list...
```

src/borg/archiver.py
```diff
@@ -1236,10 +1246,15 @@ def _delete_repository(self, args, repository):
...
+                    msg.append("You requested to completely DELETE the repository *including* all archives it "
                            "contains:")
-                for archive_info in manifest.archives.list(sort_by=['ts']):
-                    msg.append(format_archive(archive_info))
+                    for archive_info in manifest.archives.list(sort_by=['ts']):
+                        msg.append(format_archive(archive_info))
+                else:
+                    msg.append("You requested to completely DELETE the"
+                               "repository *including* %d archives it"
+                               "contains." % len(manifest.archives))
```
ThomasWaldmann, line 1257: nitpick: we do not obey 80 chars/line, but (iirc) 120 (see flake8 configuration in setup.cfg or so).

so, you could put this into 1 or 2 lines, not 3.

src/borg/archiver.py
ThomasWaldmann: yup, the tests also found that ^^^.

docs/usage/delete.rst
ThomasWaldmann, line 20: can you check whether we still need the example in 19 (and not just the one in 20)?

src/borg/archiver.py
```diff
@@ -3509,6 +3522,8 @@ def define_borg_mount(parser):
         subparser.set_defaults(func=self.do_delete)
         subparser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
                                help='do not change repository')
+        subparser.add_argument('--list', dest='output_list', action='store_true',
+                               help='output verbose list of archives or print repository')
```
ThomasWaldmann, line 3526: guess we need to remove the "or print repository" now.

the effect is only seeing the list of archives.

ThomasWaldmann:
guess this also needs a rebase onto current master to get the (macOS) tests working again.

src/borg/archiver.py
```diff
@@ -1236,10 +1243,14 @@ def _delete_repository(self, args, repository):
                 msg.append("This repository seems to have no manifest, so we can't tell anything about its "
                            "contents.")
             else:
-                msg.append("You requested to completely DELETE the repository *including* all archives it "
+                if self.output_list:
+                    msg.append("You requested to completely DELETE the repository *including* all archives it "
                            "contains:")
```
ThomasWaldmann, line 1248: please fix indentation of 1248.

ThomasWaldmann:
Thanks!

PR #5511, opened 2020-11-17
[1.1] Allow appending the files cache filename with BORG_FILES_CACHE_SUFFIX env var

luke-jr:
Backport of #5433


ThomasWaldmann:
thanks!

PR #5512, opened 2020-11-17
Add a hint on sleeping computer in the FAQ

### LINKED ISSUE #5301: macOS sleep causes Broken pipe error

Yes

Issue.

1.1.13 (using Vorta GUI)

Client: macOS 10.15.6
Server: Hetzner Storage Box

Network: Wifi

300-400 GB

`/usr/local/bin/borg create --list --info --log-json --json --filter=AM -C zstd,8 --exclude-from ... --exclude-if-present .nobackup ssh://PATH /PATH`

I observed that my backups to a network share were frequently interrupted. The errors from the log file were always
```
- vorta.borg.borg_thread - WARNING - Remote: client_loop: send disconnect: Broken pipe
- vorta.borg.borg_thread - ERROR - Connection closed by remote host
```

I thus digged into these issues, and my client SSH config now reads
```
Host XYZ
    ServerAliveInterval 30
    ServerAliveCountMax 240
    IPQoS=throughput
```
On the server, I've also set the ClientAliveInterval parameter. 
(Thanks to [this](https://medium.com/@rogelio.flores/how-to-fix-a-broken-pipe-3eb57ddeec86), [BorgBase/Vorta FAQ](https://docs.borgbase.com/faq/) and #636. The IPQoS option is not mentioned already and could help too?)

Altough I felt this improved things, the issue continued. I could finally reproduce the same error message when I forced my MacBook to sleep early and had the "Prevent computer from sleeping automatically when the display is off" (orange rectangle) unchecked (thus NOT preventing the computer from sleeping, i.e. keeping the computer awake although display is off). 
grafik

Using a tool like Amphetamine to force macOS to "stay awake" also resolved the issue. 

- Set Display Sleep to 1 Minute, uncheck the option highlighted in the graphic ("Prevent computer from sleeping automatically when the display is off")
- Start a borg backup
- Wait (it takes a few more minutes than the sleep interval)

## Discussion
This issue raised a few aspects for me (somehow related, somehow seperate), which I want to share:
- Although this behaviour does not seem to be technically a borg bug, I had a hard time tracing the bug. This is partially caused by the error message, so I wonder if it is possible to change the reporting, so that the user notices that the macOS sleep is the issue and not the SSH session (on macOS, it is sometimes hard to tell if the display is just turned off or if the computer is sleeping)
- It could be an option to stop the mac from sleeping while backing up (alternatively, this could be a Vorta thing?). The [caffeinate](https://ss64.com/osx/caffeinate.html) command could be an easy option here. 
- I am unsure if these SSH options help. The `IPQoS=throughput` is not mentioned in #636 and could help? 
- It might help other users to update the [Borg](https://borgbackup.readthedocs.io/en/stable/faq.html#why-do-i-get-connection-closed-by-remote-after-a-while) and [Vorta](https://docs.borgbase.com/faq/#my-ssh-connection-breaks-after-a-long-backup-or-prune-operation) FAQ at and include the hint that the macOS hibernation might be the cause of a "SSH error" message

Gu1nness:
Now the FAQ warns that a sleeping computer can lead to a broken pipe
when backing up over ssh.

Fixes #5301

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -183,6 +183,12 @@ This is not a problem anymore.
 
 For more details, see :ref:`checkpoints_parts`.
 
+My machine goes to sleep causing `Broken pipe`
+----------------------------------------------
+
+When backing up your data over the network, your machine should not go to sleep.
+On macOS you can use `caffeinate` to avoid that.
+
 How can I restore huge file(s) over an unstable connection?
 -----------------------------------------------------------
 


ThomasWaldmann:
Thanks!

PR #5516, opened 2020-11-21
fix --patterns-from example

### LINKED ISSUE #5310: How to back up only certain files (e.g. *.jpg)

Yes

Question

borg 1.1.13

Ubuntu 18.04.4

Intel(R) Core(TM)2 Duo CPU E8400 @ 3.00GHz
16 GB RAM
ZFS 5 x 3-way mirrors, 22 TiB usable

15 TiB

## Question

Is this possible? I can't find a reference to a command-line flag that will do this, or related FAQ, doc, or issue.

The reason I'm doing this, is to back up my most important files first, which I can get most optimally "close enough" with wildcard matching.

I understand I can generate a separate "includes" file but if there's a way to keep it native, that would be better for my use case.

Many thanks.

Update: I realize there actually is no `--files-from` flag or similar functionality (and `--patterns-from` is marked experiments - and seems a little dicey to include millions of files).

lorsanta:
Fixes #5310

I wrote a new example trying to explain the following:
- usage of leading / in patterns - bug in docs or ignored by the code?    
- ability of using a pattern type prefix directly in a pattern line      
- scope of a P definition line
- explicitly saying that R means recursion root (starting point of recursion)    
- extend the example pattern file to cover more of this

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2272,18 +2272,34 @@ class Archiver:
 
                 # "sh:" pattern style is the default, so the following line is not needed:
                 P sh
-                R /
-                # can be rebuild
-                - /home/*/.cache
-                # they're downloads for a reason
-                - /home/*/Downloads
-                # susan is a nice person
-                # include susans home
-                + /home/susan
-                # don't backup the other home directories
-                - /home/*
-                # don't even look in /proc
-                ! /proc\n\n''')
+                # borg will consider only the files within a recursion root
+                # defined in this way: "R path/to/recursion_root"
+                R /home
+
+                # this won't be included because nowhere in the file
+                # there isn't a line like "R /" or "R /proc"
+                + /proc
+
+                # dir won't be included because of '!'
+                + /home/*/Downloads/dir
+                ! /home/*/Downloads
+
+                # from now on "pp:" pattern style will be applied
+                P pp
+
+                # This is correct
+                # the position of this line in the file is irrelevant
+                R /home/*/Pictures
+
+                # it is possible to use a pattern type prefix
+                #  directly in a pattern line
+                + fn:/home/*/file1
+
+                # this will be included
+                + ./home/*/Documents/dir
+                - ./home/*/Documents/
+                # this won't
+                + ./home/*/Documents/dir2\n\n''')
     helptext['placeholders'] = textwrap.dedent('''
         Repository (or Archive) URLs, ``--prefix``, ``--glob-archives``, ``--comment``
         and ``--remote-path`` values support these placeholders:


src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
-                # susan is a nice person
-                # include susans home
-                + /home/susan
-                # don't backup the other home directories
-                - /home/*
-                # don't even look in /proc
-                ! /proc\n\n''')
+                # borg will consider only the files within a recursion root
+                # defined in this way: "R path/to/recursion_root"
+                R /home
```
ThomasWaldmann, line 2277: no leading slash in the text, but in the example.

ThomasWaldmann: also, maybe better to understand: borg starts the recursion (discovering files to back up) from such a recursion root.

"consider ... within ..." sounds similar like a include pattern.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+
+                # this won't be included because nowhere in the file
+                # there isn't a line like "R /" or "R /proc"
+                + /proc
+
```
ThomasWaldmann, line 2282: this is true, but a bit silly.

if one makes clear that we are starting recursion from /home, of course it doesn't come by /proc.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+                # dir won't be included because of '!'
+                + /home/*/Downloads/dir
+                ! /home/*/Downloads
```
ThomasWaldmann, line 2285: also true, but also a bit silly: if you forbid entering Downloads, then of course it will not enter that directory.
and as it does not enter / list that directory, it can never match the .../dir pattern.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+
+                # from now on "pp:" pattern style will be applied
+                P pp
+
+                # This is correct
```
ThomasWaldmann, line 2290: it would be good to only give correct examples.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+                # the position of this line in the file is irrelevant
+                R /home/*/Pictures
```
ThomasWaldmann, line 2292: yes, because you just define recursion roots like that.

it is similar to calling borg like `borg create repo::archive root1 root2 root3`

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+
+                # it is possible to use a pattern type prefix
+                #  directly in a pattern line
```
ThomasWaldmann, line 2295: remove the addtl. blank.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+                + fn:/home/*/file1
+
+                # this will be included
+                + ./home/*/Documents/dir
+                - ./home/*/Documents/
```
ThomasWaldmann, line 2300: if this example would be closer to the "!" one, it would be more obvious what's the point here.

src/borg/archiver.py
```diff
@@ -2272,18 +2272,34 @@ def do_break_lock(self, args, repository):
...
+                # this won't
+                + ./home/*/Documents/dir2\n\n''')
```
ThomasWaldmann, line 2302: ..., because order matters and first match counts.

ThomasWaldmann (commented):
if anything i said in my comments here is not documented, then the docs need to be fixed / clarified, not the examples.

lorsanta:
The person who opened the issue looked directly at the example, instead of reading the entire borg help patterns , that's why I thought that it would be better to rewrite it. If I'm wrong about this, maybe it would be better if I just close the PR.

src/borg/archiver.py
lorsanta: I know, I thought it was better to explicitly say that adding those lines

src/borg/archiver.py
lorsanta: But it's not obvious, for this reason I add that

src/borg/archiver.py
lorsanta: Yeah it is a bit silly, but isn't more clear to add those lines? In this way someone can immediately understand what it does.

ThomasWaldmann: understanding how stuff works needs to be in the docs.

the examples should just show correct usage of what is learned by the docs.

if you put incorrect stuff into examples, people will just copy and paste it and wonder (even if you try to put helpful comments above - people who do not read the docs will not read these comments either).

PR #5519, opened 2020-11-23
Complete documentation on append-only remote repos

### LINKED ISSUE #5497: Running borg init via a borg serve --append-only server will not create an append-only repository.

Yes

QUESTION

What exactly does the following mean?

> Running borg init via a borg serve --append-only server will not create an append-only repository.

My understanding is that if the server is configured using the following `authorized_keys`, when initializing a repo using `borg init --encryption=keyfile-blake2 "borg@server:backup"`, `--append-only` on the server is ignored.

```shell
command="borg serve --append-only --restrict-to-repository /home/borg/backup --storage-quota 10G",restrict ssh-rsa ...
```

This doesn't feel logical to me as the server config should supersede client (for security reasons). Also, I would like to use two SSH keys, one with `--append-only` and one without (using a key that requires a YubiKey for increased security).

```shell
command="borg serve --append-only --restrict-to-repository /home/borg/backup --storage-quota 10G",restrict ssh-rsa ...
```

vs

```shell
command="borg serve --restrict-to-repository /home/borg/backup --storage-quota 10G",restrict ssh-rsa ...
```

Can someone please confirm how this works?

Gu1nness:
I have updated both the FAQ and the https://github.com/borgbackup/borg/blob/master/docs/deployment/hosting-repositories.rst files.

Fixes #5497

### DIFF
--- a/docs/deployment/hosting-repositories.rst
+++ b/docs/deployment/hosting-repositories.rst
@@ -68,16 +68,17 @@ can be filled to the specified quota.
 If storage quotas are used, ensure that all deployed Borg releases
 support storage quotas.
 
-**Append-only repositorues**
+Refer to :ref:`internals_storage_quota` for more details on storage quotas.
 
-Running ``borg init`` via a ``borg serve --append-only`` server will not create
-a repository that is configured to be append-only by its repository config.
+**Append-only repositories**
 
-But, ``--append-only`` arguments in authorized_keys will override the
+Running ``borg init`` via a ``borg serve --append-only`` server will **not**
+create a repository that is configured to be append-only by its repository
+config.
+
+But, ``--append-only`` arguments in ``authorized_keys`` will override the
 repository config, therefore append-only mode can be enabled on a key by key
 basis.
-
-Refer to :ref:`internals_storage_quota` for more details on storage quotas.
 
 Refer to the `sshd(8) <http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man8/sshd.8>`_
 man page for more details on SSH options.
--- a/docs/deployment/hosting-repositories.rst
+++ b/docs/deployment/hosting-repositories.rst
@@ -70,7 +70,7 @@ support storage quotas.
 
 Refer to :ref:`internals_storage_quota` for more details on storage quotas.
 
-**Append-only repositories**
+**Specificities: Append-only repositories**
 
 Running ``borg init`` via a ``borg serve --append-only`` server will **not**
 create a repository that is configured to be append-only by its repository
--- a/docs/deployment/hosting-repositories.rst
+++ b/docs/deployment/hosting-repositories.rst
@@ -68,6 +68,15 @@ can be filled to the specified quota.
 If storage quotas are used, ensure that all deployed Borg releases
 support storage quotas.
 
+**Append-only repositorues**
+
+Running ``borg init`` via a ``borg serve --append-only`` server will not create
+a repository that is configured to be append-only by its repository config.
+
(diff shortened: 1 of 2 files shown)


docs/deployment/hosting-repositories.rst
```diff
@@ -68,6 +68,15 @@ can be filled to the specified quota.
 If storage quotas are used, ensure that all deployed Borg releases
 support storage quotas.
 
+**Append-only repositorues**
```
ThomasWaldmann, line 71: typo

ThomasWaldmann: also, compared to the paragraph above, which has some introduction into usage, this one is only pointing out the special stuff.

docs/deployment/hosting-repositories.rst
```diff
@@ -68,6 +68,15 @@ can be filled to the specified quota.
...
+
+Running ``borg init`` via a ``borg serve --append-only`` server will not create
```
ThomasWaldmann, line 73: guess the **not** should be emphasized or bold.

docs/faq.rst
```diff
@@ -183,6 +183,14 @@ This is not a problem anymore.
 
 For more details, see :ref:`checkpoints_parts`.
 
+Do we have to run borg config "backup" append_only 0 and borg config "backup" append_only 1 on the sever to switch append only mode
```
ThomasWaldmann, line 186: "backup"?

did you mean REPO?

also: add a question mark.

ThomasWaldmann: or rephrase to "How can I switch append-only on and off?"

docs/faq.rst
```diff
@@ -183,6 +183,14 @@ This is not a problem anymore.
...
+-----------------------------------------------------------------------------------------------------------------------------------
+
+You could do that, but using different ssh keys and different entries in
+authorized_keys is much easier and also maybe has less potential of thing going
```
ThomasWaldmann, line 190: add markup so that `authorized_keys` looks like a filename.

docs/faq.rst
```diff
@@ -183,12 +183,12 @@ This is not a problem anymore.
 
 For more details, see :ref:`checkpoints_parts`.
 
-Do we have to run borg config "backup" append_only 0 and borg config "backup" append_only 1 on the sever to switch append only mode
+How can I switch append-only mode on and off?
 -----------------------------------------------------------------------------------------------------------------------------------
 
 You could do that, but using different ssh keys and different entries in
```
ThomasWaldmann, line 189: You could do that (via `borg config REPO append_only 0/1`), but ...

docs/faq.rst
```diff
@@ -183,12 +183,12 @@ This is not a problem anymore.
...
-authorized_keys is much easier and also maybe has less potential of thing going
-wrong somehow.
+``authorized_keys`` is much easier and also maybe has less potential of thing
```
ThomasWaldmann, line 190: things

docs/deployment/hosting-repositories.rst
ThomasWaldmann: the 2nd item is still todo.

Gu1nness: What do you mean exactly ?
The former paragraph is about storage quotas, which seems also a special stuff, no?

ThomasWaldmann: the paragraph about quotas first tells general stuff, like how to use them.

this paragraph does not, it just says what does not work like some people could maybe expect.

ThomasWaldmann: i'll change the remaining stuff later.

PR #5520, opened 2020-11-23
Complements the documentation for pattern files and exclude files

### LINKED ISSUE #5236: Example path for --exclude-from

Yes

ISSUE

I found it quite hard to find out how to use/allow spaces in --exclude-from file, one extra example in manual/help might help a lot:

# No need to escape spaces 
/mnt/c/System Volume Information

(not having to escape spaces is quite counterintuitive :)

Gu1nness:
Fixes #5236

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2227,9 +2227,9 @@ class Archiver:
             fm:aa:something/*
             re:^home/[^/]\\.tmp/
             sh:home/*/.thumbnails
-            EOF
             # Example with spaces, no need to escape as it is processed by borg
             some file with spaces.txt
+            EOF
             $ borg create --exclude-from exclude.txt backup /
 
         .. container:: experimental
@@ -3619,6 +3619,8 @@ class Archiver:
         can be selected by passing a list of ``PATHs`` as arguments.
         The file selection can further be restricted by using the ``--exclude`` option.
 
+        See the output of the "borg help patterns" command for more help on exclude patterns.
+
         ``--progress`` can be slower than no progress display, since it makes one additional
         pass over the archive metadata.
         """)
@@ -3649,6 +3651,7 @@ class Archiver:
         be restricted by using the ``--exclude`` option.
 
         See the output of the "borg help patterns" command for more help on exclude patterns.
+
         By using ``--dry-run``, you can do all extraction steps except actually writing the
         output data: reading metadata and data chunks from the repo, checking the hash/hmac,
         decrypting, decompressing.
@@ -3990,6 +3993,8 @@ class Archiver:
         # borg list
         list_epilog = process_epilog("""
         This command lists the contents of a repository or an archive.
+
+        See the "borg help patterns" command for more help on exclude patterns.
 
         .. man NOTES
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2228,6 +2228,8 @@ class Archiver:
             re:^home/[^/]\\.tmp/
             sh:home/*/.thumbnails
             EOF
+            # Example with spaces, no need to escape as it is processed by borg
+            some file with spaces.txt
             $ borg create --exclude-from exclude.txt backup /
 
         .. container:: experimental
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2644,11 +2644,13 @@ def define_exclude_and_patterns(add_option, *, tag_files=False, strip_components
...
             add_option('--exclude-from', metavar='EXCLUDEFILE', action=ArgparseExcludeFileAction,
-                       help='read exclude patterns from EXCLUDEFILE, one per line')
+                       help='read exclude patterns from EXCLUDEFILE, one per'
+                       'line. These entries are processed by borg and not the shell')
             add_option('--pattern', metavar='PATTERN', action=ArgparsePatternAction,
                        help='experimental: include/exclude paths matching PATTERN')
             add_option('--patterns-from', metavar='PATTERNFILE', action=ArgparsePatternFileAction,
-                       help='experimental: read include/exclude patterns from PATTERNFILE, one per line')
+                       help='experimental: read include/exclude patterns from PATTERNFILE, one per line'
+                       'These entries are processed by borg and not the shell')
```
ThomasWaldmann, line 2653: maybe this is a) too long fo here and b) not really helpful as the user would need to be already aware of what the specialities of shell processing are.

so guess we better add this to the (long) docs and also give an example there that shows that spaces are not escaped in pattern files.

Gu1nness: Sure I'll do this now.

src/borg/archiver.py
```diff
@@ -3106,6 +3106,18 @@ def define_borg_mount(parser):
...
+        $ cat exclude-file.txt
+        # Comment line
+        /home/*/junk
+        *.tmp
+        One file with spaces
+        ```
+        Then the command borg ``create --exclude-from exclude-file.txt`` will
+        delete all files excluding /home/*/junk, *.tmp and the file named
+        ``One file with spaces``
+
```
ThomasWaldmann, line 3120: be careful when using markup in archiver.py:

while the rst / sphinx docs (-> online, html, pdf) support rather powerful markup, the stuff that `borg help` outputs should also look ok (and guess we only support a rather limited subset of the markup for that).

for each command, there is also a separate .rst.inc (.inc.rst?) file with examples, that could also be used.

one could also check that "borg help patterns" text, considering it is already mentioned a line above and that we need to refer to such an example for multiple borg commands, at least create, recreate and mount - so putting it just into the create help is maybe not the best solution and also causes duplication.

Gu1nness: I'll put it in the borg help patterns then :)

src/borg/archiver.py
```diff
@@ -2228,6 +2228,8 @@ def do_break_lock(self, args, repository):
             re:^home/[^/]\\.tmp/
             sh:home/*/.thumbnails
             EOF
+            # Example with spaces, no need to escape as it is processed by borg
+            some file with spaces.txt
```
ThomasWaldmann, line 2232: you need to put that above the EOF line so it becomes part of that file.

src/borg/archiver.py
```diff
@@ -2644,7 +2646,7 @@ def define_exclude_and_patterns(add_option, *, tag_files=False, strip_components
                        type=parse_exclude_pattern, action='append',
                        help='exclude paths matching PATTERN')
             add_option('--exclude-from', metavar='EXCLUDEFILE', action=ArgparseExcludeFileAction,
-                       help='read exclude patterns from EXCLUDEFILE, one per line')
+                       help='read exclude patterns from EXCLUDEFILE, one per line.')
```
ThomasWaldmann, line 2649: please remove accidental change.

src/borg/archiver.py
```diff
@@ -3605,8 +3619,6 @@ def define_borg_mount(parser):
         can be selected by passing a list of ``PATHs`` as arguments.
         The file selection can further be restricted by using the ``--exclude`` option.
 
-        See the output of the "borg help patterns" command for more help on exclude patterns.
-
```
ThomasWaldmann, line 3609: accidental change?

src/borg/archiver.py
```diff
@@ -3637,7 +3649,6 @@ def define_borg_mount(parser):
         be restricted by using the ``--exclude`` option.
 
         See the output of the "borg help patterns" command for more help on exclude patterns.
-
```
ThomasWaldmann, line 3640: accidental change?

src/borg/archiver.py
```diff
@@ -3980,8 +3991,6 @@ def define_borg_mount(parser):
         list_epilog = process_epilog("""
         This command lists the contents of a repository or an archive.
 
-        See the "borg help patterns" command for more help on exclude patterns.
-
```
ThomasWaldmann, line 3984: accidental change?

ThomasWaldmann (commented):
would be good if you (in general) looked at the overall diff on github, so you'll notice (and fix) accidental changes.

Gu1nness:
Sorry for all the accidental removals, I checked on my laptop but it did not appear to my eyes that these lines had disappeared…

src/borg/archiver.py
ThomasWaldmann: todo

src/borg/archiver.py
ThomasWaldmann: is this still needed?

ThomasWaldmann:
Thanks!

PR #5538, opened 2020-12-03
create: implement --paths-from-stdin and --paths-from-command

Lapin0t:
WIP. Follow-up of #5492. Implement two switches which shortcut the filesystem-walking logic completely: borg will read the paths and backup without doing any recursion (directories can --and should-- be passed). Should honor any relevant option (common options and all filesystem option but --one-file-system).

Slight change from my earlier proposal: now if either `--paths-from-stdin` or `--paths-from-command` are given, then it won't process any normal path from `args.path`. In `--path-from-command` it wouldn't make sense anyway since `args.path` is being used for the command. Technically, for `--paths-from-stdin` we could still handle `args.path` as usual but it's probably safe to assume that a user don't want to mix both things. Besides it simplifies the code and the top-level behavior of `create`.

WIP: Just wrote, didn't execute! :) I should verify if it works correctly and add unit tests to the PR.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -546,7 +546,7 @@ class Archiver:
                     pipe_bin = proc.stdout
                 else:  # args.paths_from_stdin == True
                     pipe_bin = sys.stdin.buffer
-                pipe = TextIOWrapper(pipe_bin, encoding='utf-8', errors='surrogateescape')
+                pipe = TextIOWrapper(pipe_bin, errors='surrogateescape')
                 for path in iter_separated(pipe, paths_sep):
                     try:
                         with backup_io('stat'):
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3312,10 +3312,9 @@ class Archiver:
                                     ' stdin below.')
         subparser.add_argument('--paths-from-stdin', action='store_true',
                                help='read DELIM-separated list of paths to backup from stdin. Will not '
-                                    'recurse in directories')
+                                    'recurse into directories.')
         subparser.add_argument('--paths-from-command', action='store_true',
-                               help='interpret PATH as command and treat its output as ``--paths-from-stdin``'
-                                    'recurse in directories')
+                               help='interpret PATH as command and treat its output as ``--paths-from-stdin``')
         subparser.add_argument('--paths-delimiter', metavar='DELIM',
                                help='set path delimiter for ``--paths-from-stdin`` and ``--paths-from-command`` (default: \\n) ')
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -51,7 +51,7 @@ try:
     from .helpers import PrefixSpec, GlobSpec, CommentSpec, SortBySpec, FilesCacheMode
     from .helpers import BaseFormatter, ItemFormatter, ArchiveFormatter
     from .helpers import format_timedelta, format_file_size, parse_file_size, format_archive
-    from .helpers import safe_encode, remove_surrogates, bin_to_hex, prepare_dump_dict
+    from .helpers import safe_encode, remove_surrogates, bin_to_hex, prepare_dump_dict, eval_escapes
     from .helpers import interval, prune_within, prune_split, PRUNING_PATTERNS
     from .helpers import timestamp
     from .helpers import get_cache_dir, os_stat
@@ -533,6 +533,36 @@ class Archiver:
                 else:
                     status = '-'
                 self.print_file_status(status, path)
+            elif args.paths_from_command or args.paths_from_stdin:
+                paths_sep = eval_escapes(args.paths_delimiter) if args.paths_delimiter is not None else '\n'
+                if args.paths_from_command:
+                    try:
+                        proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE)
+                    except (FileNotFoundError, PermissionError) as e:
+                        self.print_error('Failed to execute command: %s', e)
+                        return self.exit_code
+                    pipe = proc.stdout
+                else:  # args.paths_from_stdin == True
+                    pipe = sys.stdin
+                for path in iter_separated(pipe, paths_sep):
+                    try:
+                        with backup_io('stat'):
+                            st = os_stat(path=path, parent_fd=None, name=None, follow_symlinks=False)
+                        status = self._process_any(path=path, parent_fd=None, name=None, st=st, fso=fso,
+                                                   cache=cache, read_special=args.read_special, dry_run=dry_run)
+                    except (BackupOSError, BackupError) as e:
+                        self.print_warning('%s: %s', path, e)
+                        status = 'E'
+                    if status == 'C':
+                        self.print_warning('%s: file changed while we backed it up', path)
+                    if status is None:
+                        status = '?'
+                    self.print_file_status(status, path)
+                if args.paths_from_command:
+                    rc = proc.wait()
+                    if rc != 0:
+                        self.print_error('Command %r exited with status %d', args.paths[0], rc)
+                        return self.exit_code
             else:
                 for path in args.paths:
                     if path == '-':  # stdin
@@ -3277,6 +3307,14 @@ class Archiver:
         subparser.add_argument('--content-from-command', action='store_true',
                                help='interpret PATH as command and store its stdout. See also section Reading from'
                                     ' stdin below.')
+        subparser.add_argument('--paths-from-stdin', action='store_true',
+                               help='read DELIM-separated list of paths to backup from stdin. Will not '
+                                    'recurse in directories')
+        subparser.add_argument('--paths-from-command', action='store_true',
+                               help='interpret PATH as command and treat its output as ``--paths-from-stdin``'
+                                    'recurse in directories')
+        subparser.add_argument('--paths-delimiter', metavar='DELIM',
+                               help='set path delimiter for ``--paths-from-stdin`` and ``--paths-from-command`` (default: \\n) ')
 
         exclude_group = define_exclusion_group(subparser, tag_files=True)
         exclude_group.add_argument('--exclude-nodump', dest='exclude_nodump', action='store_true',
@@ -4522,10 +4560,12 @@ class Archiver:
         args = parser.parse_args(args or ['-h'])
         parser.common_options.resolve(args)
         func = get_func(args)
+        if func == self.do_create and args.paths and args.paths_from_stdin:
+            parser.error('Must not pass PATH with ``--paths-from-stdin``.')
         if func == self.do_create and not args.paths:
-            if args.content_from_command:
+            if args.content_from_command or args.paths_from_command:
                 parser.error('No command given.')
-            else:
+            elif not args.paths_from_stdin:
                 # need at least 1 path but args.paths may also be populated from patterns
                 parser.error('Need at least one PATH argument.')
         if not getattr(args, 'lock', True):  # Option --bypass-lock sets args.lock = False
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -27,6 +27,7 @@ try:
     from binascii import unhexlify
     from contextlib import contextmanager
     from datetime import datetime, timedelta
+    from io import TextIOWrapper
 
     from .logger import create_logger, setup_logging
 
@@ -73,6 +74,7 @@ try:
     from .helpers import flags_root, flags_dir, flags_special_follow, flags_special
     from .helpers import msgpack
     from .helpers import sig_int
+    from .helpers import iter_separated
     from .nanorst import rst_to_terminal
     from .patterns import ArgparsePatternAction, ArgparseExcludeFileAction, ArgparsePatternFileAction, parse_exclude_pattern
     from .patterns import PatternMatcher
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
```
src/borg/archiver.py:548:22: E261 at least two spaces before inline comment
src/borg/helpers/misc.py:217:1: E302 expected 2 blank lines, found 1
```
after you fix that, the real tests will run.

ThomasWaldmann:
`--path-from-command` <-- there seems to be an "s" missing.

can you fix the commit comment, the PR top post, etc., please?

src/borg/archiver.py
```diff
@@ -533,6 +533,39 @@ def create_inner(archive, cache, fso):
                 else:
                     status = '-'
                 self.print_file_status(status, path)
+            elif args.paths_from_command or args.paths_from_stdin:
+                paths_sep = '\n'
+                if args.paths_delimiter is not None:
+                    # evaluate escape sequences in the separator, in case the shell doesn't replace eg '\\0'
+                    paths_sep = escape_decode(args.paths_delimiter.encode('utf-8'))[0].decode('utf-8')
```
ThomasWaldmann, line 540: maybe a helper function for this would be nice, guess we might also need that at other places.

when done, these 4 lines could be like:
```
paths_sep = thathelper(args.path_delimiter) if args.path_delimiter is not None else '\n'
```

src/borg/archiver.py
```diff
@@ -533,6 +533,39 @@ def create_inner(archive, cache, fso):
...
+                if args.paths_from_command:
+                    try:
+                        proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE)
+                    except (FileNotFoundError, PermissionError) as e:
+                        self.print_error('Failed to execute command: %s', e)
+                        return self.exit_code
+                    pipe = proc.stdout
+                else: # args.paths_from_stdin == True
+                    pipe = sys.stdin
```
ThomasWaldmann, line 549: guess for normal text, this would be fine.

but if there is \0 in there or byte sequences that are not clean unicode: use the binary stream rather?

src/borg/archiver.py
```diff
@@ -3277,6 +3310,14 @@ def define_borg_mount(parser):
         subparser.add_argument('--content-from-command', action='store_true',
                                help='interpret PATH as command and store its stdout. See also section Reading from'
                                     ' stdin below.')
+        subparser.add_argument('--paths-from-stdin', action='store_true',
+                               help='read DELIM-separated list of paths to backup from stdin. Will not '
+                                    'recurse in directories')
+        subparser.add_argument('--paths-from-command', action='store_true',
+                               help='interptet PATH as command and treat its output as ``--paths-from-stdin``'
```
ThomasWaldmann, line 3317: typo "interpret"

ThomasWaldmann (commented):
thanks for the PR, nice powerful functionality!

ThomasWaldmann:
would be cool to get this into 1.2.0b1 (soon).

Lapin0t:
Sorry for the force-push, i needed to edit the commit message...

src/borg/archiver.py
Lapin0t: I put it in `helpers/parseformat.py`, seems like there is a bunch of encoding stuff there already. Btw i changed the escape-evaluation mechanism again. `codecs.escape_decode` wasn't really public api (some old python2 leftover apparently) and now i understand `unicode_escape` better.

ThomasWaldmann: ok, great!

ThomasWaldmann:
No problem, there is no other way than to force-push if one edits history.

Shall I merge it like this for now and we check the bytes vs. text paths later?

Lapin0t:
No wait a bit, i'm testing it and there is a weird bug, it's really not doing what it should.

ThomasWaldmann:
guess a test would be fine also.

Lapin0t:
Ok, this should be fine now. The tests are mostly copy&paste from content-from-command.. not really exhaustive in any way neither. About the text/binary thing, i realized `proc.stdout` was already in binary mode by default, so i just changed `sys.stdin.buffer` and wrapped the whole thing in textmode with `surrogateescape` which seems to be the right thing to do when handling paths (i saw there are some helpers already doing that kind of encoding so maybe you know more about it).

src/borg/helpers/misc.py
```diff
@@ -223,6 +223,7 @@ def iter_separated(fd, sep='\n', read_size=1024):
         part2, *items = buf.split(sep)
         *full, part = (part + part2, *items)
         yield from full
+        buf = fd.read(read_size)
```
Lapin0t, line 226: this was the culprit

src/borg/archiver.py
```diff
@@ -541,9 +543,10 @@ def create_inner(archive, cache, fso):
                     except (FileNotFoundError, PermissionError) as e:
                         self.print_error('Failed to execute command: %s', e)
                         return self.exit_code
-                    pipe = proc.stdout
+                    pipe_bin = proc.stdout
                 else:  # args.paths_from_stdin == True
-                    pipe = sys.stdin
+                    pipe_bin = sys.stdin.buffer
+                pipe = TextIOWrapper(pipe_bin, encoding='utf-8', errors='surrogateescape')
```
ThomasWaldmann, line 549: utf-8? do we know?

how about keeping it as bytes?

maybe also check what the internal recursion is working with, iirc it is using byte paths, not text paths.

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,23 @@ def test_dash_open():
...
+    fd = StringIO('foo\nbar/baz\n나윤a선나윤선나윤선나윤선나윤선\n')
+    assert list(iter_separated(fd)) == ['foo', 'bar/baz', '나윤a선나윤선나윤선나윤선나윤선']
+
+    # null
+    fd = StringIO('foo/bar\0baz\0spam')
+    assert list(iter_separated(fd, sep='\0')) == ['foo/bar', 'baz', 'spam']
+
+    # multichar
+    fd = StringIO('foo/barSEPbazSEPspam')
+    assert list(iter_separated(fd, sep='SEP')) == ['foo/bar', 'baz', 'spam']
```
ThomasWaldmann, line 1041: ByteIO?

src/borg/archiver.py
Lapin0t: I checked and it's using strings (i traced the type of `path` and `name` arguments to `Archiver._process_any`). But in any case `surrogateescape` makes it safe: if there is any non-utf8 byte in the input it will be encoded as a lone surrogate (a python implementation trick) and then this will be reverted whenever interacting with the filesystem. This is all documented in [pep 383](https://www.python.org/dev/peps/pep-0383/).

Maybe i shouldn't have passed the encoding tho (to use the default system-guessed encoding). I'm removing that.

src/borg/testsuite/helpers.py
Lapin0t: If we agree strings should be kept in thread above then this is fine. But `iter_separated` should be agnostic of bytes vs str so adding the test can't hurt.

src/borg/archiver.py
ThomasWaldmann: ah, ok.

src/borg/helpers/misc.py
```diff
@@ -213,3 +213,23 @@ def write(self, s):
                 except OSError:
                     pass
         return len(s)
+
+
+def iter_separated(fd, sep=None, read_size=1024):
+    """Iter over chunks of open file ``fd`` delimited by ``sep``. Doesn't trim."""
+    buf = fd.read(read_size)
```
ThomasWaldmann, line 220: would move that line to right above the `while`.

src/borg/helpers/misc.py
```diff
@@ -213,3 +213,23 @@ def write(self, s):
...
+    if type(buf) is str:
+        part = ''
+        sep = sep or '\n'
+    else:
+        part = b''
+        sep = sep or b'\n'
```
ThomasWaldmann, line 226: how about:
```
using_str = isinstance(buf, str)
part = '' if using_str else b''
sep = (sep or '\n') if using_str else (sep or b'\n')
```

src/borg/helpers/misc.py
```diff
@@ -213,3 +213,23 @@ def write(self, s):
                 except OSError:
                     pass
         return len(s)
+
+
+def iter_separated(fd, sep=None, read_size=1024):
```
ThomasWaldmann, line 218: hmm, maybe read_size=4096 (typical fs block size and also less partial processing due to less block boundaries)?

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,24 @@ def test_dash_open():
     assert dash_open('-', 'w') is sys.stdout
     assert dash_open('-', 'rb') is sys.stdin.buffer
     assert dash_open('-', 'wb') is sys.stdout.buffer
+
+
+def test_iter_separated():
+    # newline and utf-8
+    fd = StringIO('foo\nbar/baz\n나윤a선나윤선나윤선나윤선나윤선\n')
+    assert list(iter_separated(fd)) == ['foo', 'bar/baz', '나윤a선나윤선나윤선나윤선나윤선']
```
ThomasWaldmann, line 1033: how about:
```
sep, items = '\n', ['foo', 'bar', ...]
f = StringIO(sep.join(items))
assert list(iter_separated(f, sep)) == items
```

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,24 @@ def test_dash_open():
...
+    # null
+    fd = StringIO('foo/bar\0baz\0spam')
+    assert list(iter_separated(fd, sep='\0')) == ['foo/bar', 'baz', 'spam']
+    # multichar
+    fd = StringIO('foo/barSEPbazSEPspam')
+    assert list(iter_separated(fd, sep='SEP')) == ['foo/bar', 'baz', 'spam']
+    # bytes
+    fd = BytesIO(b'foo\nblop\t\n')
+    assert list(iter_separated(fd, sep=b'\n')) == [b'foo', b'blop\t']
```
ThomasWaldmann, line 1042: interesting would be to have e.g. a latin1 byte sequence with a german umlaut (so sth. that can't be decoded as utf-8).

src/borg/helpers/misc.py
Lapin0t: Definitely! I'm ashamed i forgot about isinstance :O

src/borg/helpers/misc.py
Lapin0t: Page size makes sense, i wasn't too sure about this number.

src/borg/helpers/misc.py
Lapin0t: I cannot since subsequent `isinstance` test depends of `buf`. Although this is possible if you prefer:
```
part = fd.read(0)
sep = sep or ('\n' if isinstance(part, str) else b'\n')
buf = fd.read(read_size)
```

ThomasWaldmann: oh, overlooked that. keep as is.

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,28 @@ def test_dash_open():
     assert dash_open('-', 'w') is sys.stdout
     assert dash_open('-', 'rb') is sys.stdin.buffer
     assert dash_open('-', 'wb') is sys.stdout.buffer
+
+
+def test_iter_separated():
+    # newline and utf-8
+    sep, items = '\n', ['foo', 'bar/baz', '나윤a선나윤선나윤선나윤선나윤선']
```
ThomasWaldmann, line 1032: my font does not really display characters here, just squares with numbers.

maybe use more usual unicode chars? äöü?

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,28 @@ def test_dash_open():
...
+    assert list(iter_separated(fd, sep=sep)) == items
+    # bytes
+    sep, items = b'\n', [b'foo', b'blop\t', b'gr\xe4ezi']
+    fd = BytesIO(sep.join(items))
+    assert list(iter_separated(fd)) == items
+
+
+def test_eval_escapes():
+    assert eval_escapes('\\n\\0\\x23') == '\n\0#'
+    assert eval_escapes('ä捃\\n') == 'ä捃\n'
```
ThomasWaldmann, line 1051: similar here, if it just shall be unicode, maybe something more usual?

src/borg/testsuite/helpers.py
Lapin0t: You don't have a CJK font installed? :) Ok i changed some, i just left an α character since else we only have accentuated characters, which might not be very representative.

src/borg/testsuite/helpers.py
```diff
@@ -1022,3 +1025,28 @@ def test_dash_open():
...
+    # bytes
+    sep, items = b'\n', [b'foo', b'blop\t', b'gr\xe4ezi']
+    fd = BytesIO(sep.join(items))
+    assert list(iter_separated(fd)) == items
+
+
+def test_eval_escapes():
+    assert eval_escapes('\\n\\0\\x23') == '\n\0#'
+    assert eval_escapes('äç\\n') == 'äç\n'
+    assert eval_escapes('aoeu\u1234') == 'aoeu\u1234'
```
ThomasWaldmann, line 1052: did you intend only one backslash on the left hand side?

Lapin0t: Yes, i wanted to check that it didn't touch it. Actually now i realize this is duplicate with the `äç` thing, this test is probably stupid.

ThomasWaldmann: yes, it is just testing one non-ascii unicode char (same as line before, just different notation in py src).
the \\ stuff is already tested 2 lines before that, so guess you could remove 1052.

ThomasWaldmann:
Thanks! Merged!

I created a related issue #5551.

PR #5561, opened 2020-12-11
sparse map / file map support for fixed size chunker

### LINKED ISSUE #5565: improve the fixed chunker for better sparse file support
we have a fixed blocksize chunker since a while in master branch that is good for reading raw disk files (e.g. virtual machine disks). also, it is rather simple Cython code (not hard to change/maintain C code as the variable size buzhash chunker).

often, these disk files are sparse, so unused blocks are not stored on disk.

for sparse files, there is a `os.lseek(..., SEEK_HOLE/SEEK_DATA)` api to discover the ranges that actually have data (are stored on-disk) and the ranges that are holes and have no data (the fs would just generate zeros when reading them).

the chunker currently reads input files completely. this could be optimized so it does not read holes from the fs, saving some all-zeros data shuffling from the fs code (kernel) to the borg code (userspace).

with the chunker being adapted to read only some ranges from the file (which have data we want) and not other ranges (which do not have data we want), this could be also used for other purposes in future (e.g. if we get a "changed blocks" list from some CBT "changed blocks tracking" mechanism outside of borg).

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/95154303-improve-sparse-file-support)

ThomasWaldmann:
Fixes #5565. 

- create a map of a file: (offset, size, is_data)  - is_data==False means hole (zeros only)
- use the map in the fixed chunker to read the data / seek over the sparse ranges
- use the same code with a "there is no hole" map if sparse processing is unwanted or not possible
- either create the map internally inside `chunkify` or give a ready-made map to chunkify

IDEAS:
- generalize so that such "maps" can also be used for slightly different purpose, e.g. reading the **changed** parts of a VM disk image (the other parts are not necessarily sparse then, just unchanged compared to a already known state)
- for now, it is just avoiding **reading** the sparse zeros via the fs, but it is still generating the zeros internally
- the fixed blocksize chunker is usually generating the one same all-zero chunk of the desired blocksize (except last chunk of a range being a different size likely)


ThomasWaldmann:
@enkore @infectormp @milkey-mouse would you like to review this?

ThomasWaldmann:
going to merge this soon, so if somebody wants to review, do it soon, please.

ThomasWaldmann:
The macOS ci is failing, looks like the filesystem used for tests there does not support sparse files.

ThomasWaldmann:
freebsd64 and openindiana64:

seems like /tmp has no sparse support. similar failure as macOS on github ci.


py36 on openbsd:
```
    openbsd64: >   whence = os.SEEK_HOLE                                                                                 
    openbsd64: E   AttributeError: module 'os' has no attribute 'SEEK_HOLE'     
```
Is this a Python bug? From the docs it seemed like these names are always there.

Issue #5568, opened 2020-12-15
memory leak in borg export-tar with remote (ssh:) repo


Sephtex:
Archive size: 400-500G (20~30 archive dates, expected restore size of ~90% of the total archive size)

After restoring around 60-70G, RAM usage was going up exponentially (very fast: few minutes?), going from 100-200mb to a few gigs (10-16G) till killed by OOM while the system became unstable as also other things where killed by OOM (when no RAM or SWAP was free anymore).

When RAM went up, it seemed that nothing, or hardly anything was saved to the disk. I had the feeling that nothing was written to disk but the data was kept in memory. (but this last can be a wrong assumption, and might point to a memory leak or a recursion,..)

example vars:
```
backfrom=user@192.168.2.123
path=/mount/backups/borgarchivename
archive_backup_date=2020-05-01
restore_path=/mount/restored
```

command used with using the internal ssh of borg:
```
borg export-tar "${backfrom}:${path}::${archive_backup_date}" - | tar -C $restore_path -x
```

But using borg over ssh resolved that issue, and only ~200mb was used on the system that had the backup (where we connect to):
```
ssh ${backfrom} "borg export-tar --tar-filter='gzip -2' \"${path}::${archive_backup_date}\" -" | tar -C $restore_path -xz
```

The filesystem on source and target is ZFS.
Network transfer was at least 100Mbit, could have been 1Gbit. (transfer speed was for site ~10mb/s, maybe it was 2-3 times faster, but can't remember)

Checked and tested with 1.1.13.

infectormp:
See also https://github.com/borgbackup/borg/issues/4852

ThomasWaldmann:
The issue title says `borg extract`, but the text says `borg export-tar` - these are 2 different functions, so please fix whatever is wrong.

Also, if you can reproduce the issue, can you also try with 1.1.14 and 1.1.15-rc?

Sephtex:
@ThomasWaldmann Fixed the title.

If I recall correctly, I tested with the latest version that was available 2 months ago when I was testing this. If I'm correct it was 1.1.14, 1.1.15-rc isn't yet available on my system. (I didn't mentioned it since I wasn't sure about it, only that I tried a more recent version that was available at that time.)

ThomasWaldmann:
OK, this looks interesting. Especially the difference between:
- borg repo access via ssh (and the tar stream over local pipe)
- borg repo access locally (and the tar stream over ssh pipe)

[...]

PR #5569, opened 2020-12-15
fix memory leak in export-tar with ssh: repo

ThomasWaldmann:
found the leak, see last commit.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -922,8 +922,8 @@ class Archiver:
         hardlink_masters = {} if partial_extract else None
 
         def peek_and_store_hardlink_masters(item, matched):
-            if (partial_extract and not matched and hardlinkable(item.mode) and
-                    item.get('hardlink_master', True) and 'source' not in item):
+            if ((partial_extract and not matched and hardlinkable(item.mode)) and
+                    (item.get('hardlink_master', True) and 'source' not in item)):
                 hardlink_masters[item.get('path')] = (item.get('chunks'), None)
 
         filter = self.build_filter(matcher, peek_and_store_hardlink_masters, strip_components)
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1024,7 +1024,8 @@ class Archiver:
                 return None, stream
             return tarinfo, stream
 
-        for item in archive.iter_items(filter, preload=True, hardlink_masters=hardlink_masters):
+        for item in archive.iter_items(filter, partial_extract=partial_extract,
+                                       preload=True, hardlink_masters=hardlink_masters):
             orig_path = item.path
             if strip_components:
                 item.path = os.sep.join(orig_path.split(os.sep)[strip_components:])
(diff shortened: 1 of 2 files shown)


src/borg/archiver.py
```diff
@@ -1024,7 +1024,8 @@ def item_to_tarinfo(item, original_path):
                 return None, stream
             return tarinfo, stream
 
-        for item in archive.iter_items(filter, preload=True, hardlink_masters=hardlink_masters):
+        for item in archive.iter_items(filter, partial_extract=partial_extract,
+                                       preload=True, hardlink_masters=hardlink_masters):
```
ThomasWaldmann, line 1028: From above:
```
partial_extract = not matcher.empty() or strip_components
```

So this fix is only relevant for cases when this gets True.

Like: giving any options that lead to partial extraction of an archive, including/excluding stuff, stripping directory layers.

src/borg/archiver.py
```diff
@@ -922,8 +922,8 @@ def _export_tar(self, args, archive, tar):
         hardlink_masters = {} if partial_extract else None
 
         def peek_and_store_hardlink_masters(item, matched):
-            if (partial_extract and not matched and hardlinkable(item.mode) and
-                    item.get('hardlink_master', True) and 'source' not in item):
+            if ((partial_extract and not matched and hardlinkable(item.mode)) and
+                    (item.get('hardlink_master', True) and 'source' not in item)):
```
enkore, line 926: Since these parentheses don't change the semantics of this expression, what do they convey?

src/borg/archiver.py
enkore: Good catch.

ThomasWaldmann: yeah, but the main fix is in last commit.

src/borg/archiver.py
ThomasWaldmann: no they don't, but in borg extract there is some explanation explaining the 2 toplevel terms.
i just copied same parens structure from there.

src/borg/archiver.py
```diff
@@ -940,7 +940,8 @@ def item_content_stream(item):
             """
             Return a file-like object that reads from the chunks of *item*.
             """
-            chunk_iterator = archive.pipeline.fetch_many([chunk_id for chunk_id, _, _ in item.chunks])
+            chunk_iterator = archive.pipeline.fetch_many([chunk_id for chunk_id, _, _ in item.chunks],
+                                                         is_preloaded=True)
```
ThomasWaldmann, line 944: this is the fix for what was reported in #5568 - full archive extraction was used there.

enkore: Ah I see.

Sephtex: With this small change, RAM usage was much less (half of what it was before) and the file where the issue was before got through. Thanks!

ThomasWaldmann: ok, great, so I'll merge this now.

PR #5571, opened 2020-12-16
Add some examples to pp and pf patterns

### LINKED ISSUE #5300: What is the difference of pattern types pf and pp
Hello,

  from reading borg-patterns(1), it is not clear to me what the difference is between the pattern types pp and pf.

Sticking with the example - say I wanted to exclude root/somedir. Both pp:root/somedir and pf:root/somedir seem to have exactly the same effect, the directory and all its contents are not part of the backup.

What is the difference?

Gu1nness:
Fixes #5300

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2161,12 +2161,14 @@ class Archiver:
             This pattern style is useful to match whole sub-directories. The pattern
             `pp:root/somedir` matches `root/somedir` and everything therein. A leading
             path separator is always removed.
+            Another example : `pp:foo` will match `foobar`
 
         Path full-match, selector `pf:`
             This pattern style is (only) useful to match full paths.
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2161,12 +2161,14 @@ def do_break_lock(self, args, repository):
             This pattern style is useful to match whole sub-directories. The pattern
             `pp:root/somedir` matches `root/somedir` and everything therein. A leading
             path separator is always removed.
+            Another example : `pp:foo` will match `foobar`
```
ThomasWaldmann, line 2164: can you please test examples before adding them?

Gu1nness: Sorry, I thought I had tested them but seems to be a hallucination (I am schizophrenic…)

PR #5574, opened 2020-12-18
Add remote upload buffer (--remote-buffer)

axapaxa:
This PR adds a new option --remote-buffer and allows archiver to_send buffer to grow to a chosen size. To_send is then used to fill the pipe, even on non-sending commands (like async_response).
**Why is this useful?**
When sending data, it is common to have longer runs of data (over single chunk) followed by a big number of unchanged data. Instead of sending all chunks first, then checking all unchanged data, it is better to instead send data and look for the next changed chunk in parallel. This PR more or less allows just that for remote repositories. The maximum size of the buffer is configurable by --remote-buffer specified in MiB. This should shorten the backup time for repositories with large gaps of unchanged data and slow remotes, at the cost of some memory.

As for drawbacks, this option seemingly stops borg when creating checkpoint or finishing `create`. This is expected, as up to the whole buffer has to be sent before the remote will accept any new commands.

I tried my best to update docs, but perhaps I made mistakes. I looked at tests but didn't find any for call_many, so I gave up (improvements welcome). I didn't fully run tests because `fakeroot -u tox` was only at 30% after 2h of waiting (on very high end machine). No idea what's wrong here.

I'll be unable to make any changes until new year (no access to main dev machine), so improvements are welcome.


ThomasWaldmann:
Did you do some practical benchmark that shows that this speeds up a backup?

axapaxa:
@ThomasWaldmann Yes, I did some practical tests that shower rather small improvement (from 11min down to 10min) on real environment (so true test, but noisy). My data was: 500mb of same /dev/urandom + 40mb of always random /dev/urandom copied 20x. So only first 500mb was random, rest was deduplicated, but 40mb chunks were always "new". Chunker was set to 4MiB blocks. From my measurement it also showed that data continued to be transmitted during those 500mb random chunks (because async_response is always called so it always filled buffer). But keep in mind that this was only on ~12Gb.  Also, this was over WAN internet, so 11min vs 10min was not measurable enough for my liking :). I'll probably follow with some more tests (but likely only after new year), but sadly `--remote-ratelimit` will probably not reflect real world in this case, as it does `sleep` etc. which is not true to "real" pipe behavior.
I also did `create` + `extract` test to make sure something didn't go wrong like chunks missing etc. - it went fine, returned file matched.
If anyone else wants to test and share their result, feel free, just remember to avoid using `--remote-ratelimit` as this will not reflect potential gains of this method. And to tests this on rather large files, where most of the file didn't change, only small parts. Ideally, harddisk images should be nearly perfect for this improvement.

axapaxa:
I was able to do some experiments and I discovered that such large `to_send` buffer was causing major slowdown. So I created chunked collection to hold data, and here are results for 80gb sparse file (no --sparse argument) with about 2gb of changing data spread out in 80mb chunks):
Without fix: 8m28s, 8m22s
With fix (`--remote-buffer 160`): 5m2s, 5m14s

Monitoring internet usage during that period revealed that "without fix" version only sent data during changing bits, not using connection in between when seeking to next chunk. "With fix" version sent data continuously, but transfer speed dropped when changing chunks were encountered: this is fairly easy to explain, as then repository was called less frequently (as reading, encryption etc. takes time). So my fix is by no means perfect.
@ThomasWaldmann I have three ways to proceed I see. I can either try to write threaded code that uploads data in mostly blocking manner (hence avoiding transfer drop when main thread is busy), polish up my current version which is faster but not perfect, or I can leave it up to someone more knowledgeable in borg than me to fix. Ideas? :)

By the way: As a side effect I noticed that large chunks are sent over network actually slower than smaller chunks, in my experimentation ~1mb chunks were able to perfectly fill my upload transfer, but ~4mb less so. This is because `to_send` is immutable and every socket write it is copied in memory cutting some data from front. This in turn leads to less perfect filling of buffer, slowing down whole process.

ThomasWaldmann:
I'ld prefer a minimal fix, definitely not adding multithreading.

axapaxa:
I have added `EfficientBytesQueue` to reduce cutting performance slowdown when `to_send` buffer gets large. Tests included. If there is anything else I need to do, let me know :).  In long run, sending thread might be beneficial, but I understand that multithreading is not a simple thing.

src/borg/testsuite/efficient_bytes_queue.py
```diff
@@ -0,0 +1,40 @@
...
+class TestEfficientQueue:
+    def test_base_usage(self):
+        sut = EfficientBytesQueue(100)
+        assert not sut.peek_front()
+        sut.push_back(b'1234')
+        assert sut.peek_front() == b'1234'
+        assert len(sut) == 4
+        assert sut
+        sut.pop_front(4)
+        assert not sut.peek_front()
```
ThomasWaldmann, line 17: rather `assert sut.peek_front() == b''` (if the result shall always be the same data type, here `bytes`).

alternatively, `None` could also be an option.

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
         if size is not None:
             self.resize(size, init)
         return self.buffer
+
+
+class EfficientBytesQueue:
```
ThomasWaldmann, line 54: with a little change, i guess this could also be generalized to have any sequence type as elements (not just bytes).

not likely to be used by borg though, so guess this is optional...

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+
+    def __init__(self, split_size):
+        self.buffers = []
+        self.size = 0
+        self.split_size = split_size
+
+    def peek_front(self):
+        if not self.buffers:
+            return []
+        buffer = self.buffers[0]
```
ThomasWaldmann, line 70: this is a bit inconsistent: returns an empty `list` for empty case, otherwise returns an `element (bytes)`.

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+
+    class SizeUnderflow(Error):
+        """Could not pop_front first {} elements, collection only has {} elements.."""
+
+    def __init__(self, split_size):
+        self.buffers = []
+        self.size = 0
+        self.split_size = split_size
+
+    def peek_front(self):
```
ThomasWaldmann, line 67: add docstrings, please (here and other methods).

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+            buffer = buffer + data[:to_add]
+            data = data[to_add:]
+            self.buffers[-1] = buffer
+            self.size += to_add
+
+    def __len__(self):
+        return self.size
+
+    def __bool__(self):
+        return len(self) != 0
```
ThomasWaldmann, line 106: as you directly use `self.size` above also, you could also directly use it here.

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+        if size > self.size:
+            raise EfficientBytesQueue.SizeUnderflow(size, self.size)
+        while size > 0:
+            buffer = self.buffers[0]
+            to_remove = min(size, len(buffer))
+            buffer = buffer[to_remove:]
+            if buffer:
+                self.buffers[0] = buffer
+            else:
+                self.buffers = self.buffers[1:]
```
ThomasWaldmann, line 83: `del self.buffers[0]`

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+        if not self.buffers:
+            self.buffers = [b'']
+        while data:
+            buffer = self.buffers[-1]
+            if len(buffer) >= self.split_size:
+                buffer = b''
+                self.buffers = self.buffers + [buffer]
+
+            to_add = min(len(data), self.split_size - len(buffer))
+            buffer = buffer + data[:to_add]
```
ThomasWaldmann, line 97: here, it looks like the last element can not grow bigger than split_size.

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,58 @@ def get(self, size=None, init=False):
...
+            self.size -= to_remove
+
+    def push_back(self, data):
+        if not self.buffers:
+            self.buffers = [b'']
+        while data:
+            buffer = self.buffers[-1]
+            if len(buffer) >= self.split_size:
+                buffer = b''
+                self.buffers = self.buffers + [buffer]
```
ThomasWaldmann, line 94: is this needed? see next comment.

src/borg/helpers/datastruct.py
ThomasWaldmann: also: `buffer += data[:to_add]`

src/borg/remote.py
```diff
@@ -864,29 +881,20 @@ def handle_error(unpacked):
...
-                    try:
-                        written = self.ratelimit.write(self.stdin_fd, self.to_send)
-                        self.tx_bytes += written
-                        self.to_send = self.to_send[written:]
-                    except OSError as e:
-                        # io.write might raise EAGAIN even though select indicates
-                        # that the fd should be writable
-                        if e.errno != errno.EAGAIN:
-                            raise
+                fill_send_buffer()
```
ThomasWaldmann, line 897: please keep this inline to make the diff simpler to review.

src/borg/testsuite/efficient_bytes_queue.py
```diff
@@ -0,0 +1,40 @@
+from tempfile import TemporaryFile
```
ThomasWaldmann, line 1: not used?

src/borg/testsuite/efficient_bytes_queue.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+
+from ..helpers.datastruct import EfficientBytesQueue
+
+
+class TestEfficientQueue:
+    def test_base_usage(self):
+        sut = EfficientBytesQueue(100)
```
ThomasWaldmann, line 10: i had to think a bit until i found that sut is "subject under test" likely. :-)

is this a frequent name in unit tests?

ThomasWaldmann (changes requested):

src/borg/remote.py
axapaxa: I added another use, that's why I moved it to separate function. I could bring it back, but then I see no other choice than code duplication? Forgive me if as python noob I miss something :).

src/borg/testsuite/efficient_bytes_queue.py
axapaxa: Yes it means "subject under test". Frequent depends on where :). I use it quite often where I work, but I can adjust to whatever is your preference. Seemingly you don't use any convention so I can rename it `queue` if you don't like `sut`. I don't have hard preferences.

src/borg/remote.py
```diff
@@ -711,6 +713,19 @@ def call_many(self, cmd, calls, wait=True, is_preloaded=False, async_wait=True):
         if not calls and cmd != 'async_responses':
             return
 
+        def fill_send_buffer():
```
ThomasWaldmann, line 716: why is this called **fill**_send_buffer? looks like it is rather sending (and emptying) the send buffer.

src/borg/remote.py
```diff
@@ -760,6 +775,8 @@ def handle_error(unpacked):
 
         calls = list(calls)
         waiting_for = []
+        maximum_to_send = 0 if wait else self.upload_buffer_size_limit
+        fill_send_buffer()
```
ThomasWaldmann, line 779: please add a comment why this line is needed.

src/borg/remote.py
ThomasWaldmann: ah, ok, then keep it as is. didn't notice that addtl. call.

src/borg/remote.py
axapaxa: It is filling pipe send buffer from internal send buffer... But you are right, from here it looks like we are emptying buffer, I'll just rename it `send_data`?

src/borg/testsuite/efficient_bytes_queue.py
ThomasWaldmann: yeah, queue might be easier.

src/borg/remote.py
ThomasWaldmann: yeah, send_data please.

src/borg/helpers/datastruct.py
axapaxa: Yes, last element will not grow bigger than split_size. Instead in next loop iteration new chunk will be created to accommodate remaining elements. This is by design, as queue guarantees that chunk will have between 1 and split_size elements.

src/borg/helpers/datastruct.py
axapaxa: If previous chunk is full, we have to make splitting point and grow "buffers". Otherwise we will always have only single buffer.
This is rationale for whole "efficiency". If we make single huge array then remove few elements from front, whole array has to be copied without first few elements. This is a problem if that array is hundreds of mebibytes (easily possible with `--remote-buffer 100`). So instead I create multiple arrays each at most 1mebibyte in length. Then if we have to make this copying, we never have to move more than 1mebibyte worth of data.

ThomasWaldmann: ah, i was irritated by the `>= self.split_size` because it always will be `<=`, so only `==` will trigger this block.

but please keep it as is, it is more defensive like that.

the code in the block can be optimized to:
```
self.buffers.append(b'')  # start a new buffer when previous is full
```

src/borg/remote.py
```diff
@@ -535,14 +536,15 @@ def __init__(self, location, create=False, exclusive=False, lock_wait=None, lock
         self.msgid = 0
         self.rx_bytes = 0
         self.tx_bytes = 0
-        self.to_send = b''
+        self.to_send = EfficientCollectionQueue(1024 * 1024, lambda: b'')
```
ThomasWaldmann, line 539: instead of that lambda, you could give just `member_type=bytes`.

usually, calling `member_type()` will just result in an empty instance.

src/borg/helpers/datastruct.py
```diff
@@ -49,3 +49,83 @@ def get(self, size=None, init=False):
...
+
+class EfficientCollectionQueue:
+    """
+    An efficient FIFO queue that splits received elements into chunks.
+    """
+
+    class SizeUnderflow(Error):
+        """Could not pop_front first {} elements, collection only has {} elements.."""
+
+    def __init__(self, split_size, empty_collection_lambda):
```
ThomasWaldmann, line 62: s/empty_collection_lambda/member_type/g

src/borg/remote.py
```diff
@@ -711,6 +713,19 @@ def call_many(self, cmd, calls, wait=True, is_preloaded=False, async_wait=True):
...
 
+        def send_buffer():
+            if self.to_send:
+                try:
+                    written = self.ratelimit.write(self.stdin_fd, self.to_send.peek_front())
+                    self.tx_bytes += written
+                    self.to_send.pop_front(written)
+                except OSError as e:
+                    # io.write might raise EAGAIN even though select indicates
+                    # that the fd should be writable
```
ThomasWaldmann, line 724: nitpick: add a dot at end of sentence.

src/borg/remote.py
```diff
@@ -711,6 +713,19 @@ def call_many(self, cmd, calls, wait=True, is_preloaded=False, async_wait=True):
...
+                    # EWOULDBLOCK is added for defensive programming sake.
+                    if (e.errno != errno.EAGAIN) and (e.errno != errno.EWOULDBLOCK):
```
ThomasWaldmann, line 726:
```
if e.errno not in (errno.EAGAIN, errno.EWOULDBLOCK): ...
```

src/borg/remote.py
```diff
@@ -760,6 +775,8 @@ def handle_error(unpacked):
 
         calls = list(calls)
         waiting_for = []
+        maximum_to_send = 0 if wait else self.upload_buffer_size_limit
+        send_buffer()  # Try to send data, as some cases (async_response) will never try to send data otherwise.
```
ThomasWaldmann, line 779: thanks!

docs/usage/common-options.rst.inc
```diff
@@ -14,6 +14,7 @@
 --umask M          set umask to M (local only, default: 0077)
 --remote-path PATH    use PATH as borg executable on the remote (default: "borg")
 --remote-ratelimit RATE    set remote network upload rate limit in kiByte/s (default: 0=unlimited)
+--remote-buffer UPLOAD_BUFFER   set remote upload buffer size in miByte. (default: 0=no buffer)
```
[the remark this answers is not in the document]
axapaxa: perhaps I should change both --remote-ratelimit and --remote-buffer to "set borg-over-ssh network upload rate limit" and "set borg-over-ssh network upload buffer size"? Then we are clear that this is network not parameter, and this only relates to "borg-over-ssh" usecase, NOT remote filesystems like sshfs/samba etc. But given that I'm usually great at confusing end users this probably also does more harm than good :D.
Also I wanted to be consistent with ratelimit, hence probably `MiByte` would be closer. So perhaps we should either change `MiB` and `KiB` or `MiByte` and `KiByte` in both options?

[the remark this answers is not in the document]
ThomasWaldmann: we can change the option names in a separate PR later. as `--remote-ratelimit` is already released (IIRC), we can only deprecate it and add a better option name for future use.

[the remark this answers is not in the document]
ThomasWaldmann: guess `MiBytes` if you choose the long form.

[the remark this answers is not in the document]
axapaxa: I actually  meant to change description, not argument name itself (which I know is difficult to rename as it's already released). But sure, it can be done later.

src/borg/archiver.py
```diff
@@ -2708,8 +2708,8 @@ def define_common_options(add_common_option):
                               help='use PATH as borg executable on the remote (default: "borg")')
             add_common_option('--remote-ratelimit', metavar='RATE', dest='remote_ratelimit', type=int,
                               help='set remote network upload rate limit in kiByte/s (default: 0=unlimited)')
-            add_common_option('--remote-buffer', metavar='UPLOAD_BUFFER', dest='remote_buffer', type=int,
-                              help='set remote upload buffer size in miByte. (default: 0=no buffer)')
+            add_common_option('--remote-buffer', metavar='UPLOAD_BUFFER', dest='remote_buffDer', type=int,
```
ThomasWaldmann, line 2711: typo

ThomasWaldmann (approved):
LGTM.

Guess this will need a bit more practical testing, but this can be done while testing 1.2 beta/rc.

ThomasWaldmann:
Thanks for the PR, merged!

PR #5576, opened 2020-12-19
Use HTTPS everywhere (mechanical edit using util from https-everywhere)

ypid:
Ref: https://github.com/EFForg/https-everywhere/tree/master/utils/rewriter

```Shell
~/src/EFForg/https-everywhere/utils/rewriter/rewriter.js .
```

A few changes were reset/fixed manually before the commit.

PR #5584, opened 2020-12-22
Use HTTPS everywhere (mechanical edit using util from https-everywhere)

ypid:
Backport of: #5576

> Do you want to backport this to 1.1-maint branch?

The patch applies cleanly so why not :)


ThomasWaldmann:
Thanks!

PR #5587, opened 2020-12-24
update shell completions

oxiedi:


ThomasWaldmann:
Ah, great, just in time! :-) Want to do 1.1.15 soon...

PR #5600, opened 2020-12-28
fix cythonize crash on macOS/py39, fixes #5599

### LINKED ISSUE #5599: cythonize crashes on macOS with py39
Found this issue on the macOS vagrant machine after upgrading python to 3.9.1:

https://github.com/cython/cython/issues/3262

ThomasWaldmann:


ThomasWaldmann:
1.1-maint needs checking if the fix is also needed for there (if multiprocessing/threads are used for cython there).

PR #5614, opened 2021-01-06
remove empty shadowed_segments lists, fixes #5275

### LINKED ISSUE #5275: check empty segments lists in shadow_index
while doing #5263, i tried to remove empty segments lists from the shadow_index, but then tests failed.

as they maybe shouldn't remain in there forever, this needs checking again

ThomasWaldmann:
also:
- add test for removed empty shadowed_segments list
- add some comments
- add repo_dump test debug tool

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -794,6 +794,8 @@ class Repository:
                     try:
                         self.shadow_index[key].remove(segment)
                     except (KeyError, ValueError):
+                        # do not remove entry with empty shadowed_segments list here,
+                        # it is needed for shadowed_put_exists code (see below)!
                         pass
                 elif tag == TAG_DELETE and not in_index:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
@enkore can you please try & review?

timmc:
I notice that `Repository.put` doesn't add to `shadow_index` -- is that intentional? Not sure if there's an impact on this PR.

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
@timmc good catch! i did a little refactor and added a comment (see 41800bc).

we will have to check whether there is a bug or that has to be like that, not sure yet.

ThomasWaldmann:
@timmc @enkore see discussion above and commits 41800bc and c7f0aac .

ThomasWaldmann:
I'll merge this soon, but it would be good if there was some more independent review.

PR #5615, opened 2021-01-07
Change docs/install instruction: "python-virtualenv" -> "python3-virtualenv"

jedie:


ThomasWaldmann:
Good catch. Can you also do a backport to 1.1-maint branch?

PR #5636, opened 2021-01-18
Fix "put updates shadow index"

ThomasWaldmann:
@timmc found:

"I notice that Repository.put doesn't add to shadow_index -- is that intentional? Not sure if there's an impact on this PR."

See commit comments.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1216,11 +1216,11 @@ class Repository:
             segment, offset = self.index.pop(id)
         except KeyError:
             raise self.ObjectNotFound(id, self.path) from None
-        self.shadow_index.setdefault(id, []).append(segment)
         self._delete(id, segment, offset)
 
     def _delete(self, id, segment, offset):
         # common code used by put and delete
+        self.shadow_index.setdefault(id, []).append(segment)
         self.segments[segment] -= 1
         size = self.io.read(segment, offset, id, read_data=False)
         self.storage_quota_use -= size
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1191,13 +1191,9 @@ class Repository:
         except KeyError:
             pass
         else:
-            self.segments[segment] -= 1
-            size = self.io.read(segment, offset, id, read_data=False)
-            self.storage_quota_use -= size
-            self.compact[segment] += size
-            segment, size = self.io.write_delete(id)
-            self.compact[segment] += size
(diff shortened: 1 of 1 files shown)


src/borg/repository.py
```diff
@@ -1213,6 +1207,10 @@ def delete(self, id, wait=True):
             segment, offset = self.index.pop(id)
         except KeyError:
             raise self.ObjectNotFound(id, self.path) from None
+        self._delete(id, segment, offset)
+
+    def _delete(self, id, segment, offset):
+        # common code used by put and delete
         self.shadow_index.setdefault(id, []).append(segment)
```
ThomasWaldmann, line 1214: ^ this line was missing in original put code.

ThomasWaldmann:
OK, so looks like we had a bug here for quite a while.

But borg does usually not `put` something it already has in the repo - usually this would be pointless, except for:
- chunkid 0 (manifest): updates always get put to id 0
- borg recreate recompressing chunks: id is based on plaintext, so recompressed content gets put to same id

ThomasWaldmann:
@enkore please review.

src/borg/repository.py
```diff
@@ -1184,13 +1184,7 @@ def put(self, id, data, wait=True):
...
             pass
         else:
-            self.segments[segment] -= 1
-            size = self.io.read(segment, offset, id, read_data=False)
-            self.storage_quota_use -= size
-            self.compact[segment] += size
-            segment, size = self.io.write_delete(id)
-            self.compact[segment] += size
-            self.segments.setdefault(segment, 0)
+            self._delete(id, segment, offset)
```
timmc, line 1187: It might be worth including a comment here indicating that the shadow_index will not get updated without performing an explicit delete here.

I'm not sure what to make of how borg's repository layer allows a PUT to shadow another PUT, but currently never *does* that (instead using DELETE + PUT). I'm not sure if `PUT A, PUT A` should be considered a sequence in need of repair.

timmc:
Would it be sufficient to add to `borg check`'s capabilities and add all but the final manifest to the shadow index?

src/borg/repository.py
ThomasWaldmann: added a comment.

seems like put del put is also needed for other bookkeeping.
e.g. imagine borg recreate with recompress - data size will usually change here, but chunkid will stay same.

ThomasWaldmann:
let's discuss consequences and borg check in #5661.

PR #5642, opened 2021-01-22
fix grammar in faq

### LINKED ISSUE #5632: fix grammar in docs
Current docs:
```
It does intentionally not contain files that have a timestamp same
as the newest timestamp in the created archive.
```

From @jdchristensen : 

This is grammatically incorrect and should be rephrased to "It intentionally does not contain ..." or, better, "It intentionally excludes files that have a timestamp that matches the newest timestamp in the created archive".

rpolley:
Closes #5632


ThomasWaldmann:
Thanks! This maybe needs a backport to 1.1-maint.

Issue #5504, opened 2020-11-13
update docs on dev env setup


laur89:
Feel like the documentation needs some updating on preparing developing environment.
Could we confirm whether following is agreeable.

### 1
Following is said [here](https://borgbackup.readthedocs.io/en/stable/development.html#building-a-development-environment):

> First, just install borg into a virtual env as described before.

It's unclear what `before` is referring to, as it's not on the same page. Would it be reasonable to link `described before` to https://borgbackup.readthedocs.io/en/stable/installation.html#from-source? Specifically to the "Using pip" and/or "using git" subsections?

### 2
[Installation/using git](https://borgbackup.readthedocs.io/en/stable/installation.html#using-git) has following

> pip install -r requirements.d/fuse.txt  # optional, for FUSE support

[requirements.d/](https://github.com/borgbackup/borg/tree/master/requirements.d) has no such file. Instructions on the Installation docs need to be updated.

### 3
It would be nicer to provide a Docker-based dev environment as opposed to vagrant. Eg I'm stuck on a Windows computer and do all development from Virtualbox VM, which excludes vagrant. 

### 4
Supported python versions need to be documented.
I'm running Debian testing release, and at the time of writing have versions `3.8.6` & `3.9.0+` installed. Eg running tox produces following errors:

    ERROR:  py36-fuse2: InterpreterNotFound: python3.6                                                                                                                                                                                             
    ERROR:  py36-fuse3: InterpreterNotFound: python3.6                                                                                                                                                                                             
    ERROR:  py37-fuse2: InterpreterNotFound: python3.7                                                                                                                                                                                             
    ERROR:  py37-fuse3: InterpreterNotFound: python3.7

Or maybe add note for adding `-e py3x` option to tox whenever needed?

ThomasWaldmann:
1 yes, the borg install instructions are meant. of course a link would be an improvement.

2 master and 1.1-maint might differ, please check. `pip install -e .[fuse]` or `pip install borgbackup[fuse]`

3 there is a PR about docker and if multiple docker users would review that and agree on something simple, it could be merged.

i don't use docker, so i won't merge anything complicated that might need much / complex maintenance.

4 setup.py specifies supported py versions and that stuff also gets on pypi.org.

you can call tox with --skip-missing-interpreters (or so) to skip any missing interpreters.
i'ld rather not use -e pyXX because that will not run the flake8 test and people will push stuff with pep8 issues.

rpolley:
[...]

PR #5643, opened 2021-01-22
Update docs for dev environment installation instructions

rpolley:
See #5504 for motivation. The current installation instructions are confusing and result in failing tests on a typical setup, this adds a link back to an earlier section that's mentioned in the dev environment setup and changes the installation instructions to recommend only running the tests for installed versions of python.


ThomasWaldmann:
Thanks! Maybe this also needs a backport to 1.1-maint.

PR #5644, opened 2021-01-22
add examples for --paths-from-stdin, --paths-from-command, --paths-separator

### LINKED ISSUE #5551: examples for --paths-from-stdin --paths-from-command --paths-separator
we could add some useful examples for these, esp. combining it with the `find` tool:

- backup small (<1MiB) files first
- backup files from user joe first

rpolley:
Closes #5551
Added some examples for these options to the `borg create` documentation, one that showcases each of them.

### DIFF
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -81,7 +81,7 @@ Examples
     # Use --paths-from-command with find to only backup files from a given user
     $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
     # Use --paths-from-stdin with --paths-delimiter to backup files listed in a csv file
-    $ cat ~files-to-backup.csv | borg create \
+    $ cat ~/files-to-backup.csv | borg create \
         --paths-from-stdin \
         --paths-delimiter "," \
         /path/to/repo::csv-specified-files
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -80,8 +80,8 @@ Examples
     $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
     # Use --paths-from-command with find to only backup files from a given user
     $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
-    # Use --paths-from-stdin with --paths-delimiter to backup files listed in a csv file
-    $ cat ~/files-to-backup.csv | borg create \
+    # Use --paths-from-stdin with --paths-delimiter (for example, for files with newlines in them)
+    $ find ~ -size -1000k -print0 | borg create \
         --paths-from-stdin \
-        --paths-delimiter "," \
-        /path/to/repo::csv-specified-files
+        --paths-delimiter "\0" \
+        /path/to/repo::smallfiles-handle-newline
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -79,7 +79,7 @@ Examples
     # Use --paths-from-stdin with find to only backup files less than 1MB in size
     $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
     # Use --paths-from-command with find to only backup files from a given user
-    $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
+    $ borg create --paths-from-command /path/to/repo::joes-files -- find /srv/samba/shared -user joe
     # Use --paths-from-stdin with --paths-delimiter (for example, for files with newlines in them)
     $ find ~ -size -1000k -print0 | borg create \
         --paths-from-stdin \
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -74,3 +74,14 @@ Examples
     $ cd /home/user/Documents
     # The root directory of the archive will be "projectA"
     $ borg create /path/to/repo::daily-projectA-{now:%Y-%m-%d} projectA
+
+    # Use external command to determine files to archive
+    # Use --paths-from-stdin with find to only backup files less than 1MB in size
+    $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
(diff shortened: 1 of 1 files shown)


docs/usage/create.rst
```diff
@@ -74,3 +74,14 @@ Examples
     $ cd /home/user/Documents
     # The root directory of the archive will be "projectA"
     $ borg create /path/to/repo::daily-projectA-{now:%Y-%m-%d} projectA
+
+    # Use external command to determine files to archive
+    # Use --paths-from-stdin with find to only backup files less than 1MB in size
+    $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
+    # Use --paths-from-command with find to only backup files from a given user
+    $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
```
ThomasWaldmann, line 82: is that correct? would it also work like:
```
borg create --paths-from-command='find ~ -user joe' /path/to/repo::joes-files
```

docs/usage/create.rst
```diff
@@ -74,3 +74,14 @@ Examples
...
+    # Use --paths-from-stdin with --paths-separator to backup files listed in a csv file
+    $ echo files-to-backup.csv | borg create \
+        --paths-from-stdin \
+        --paths-separator "," \
+        /path/to/repo::csv-specified-files
```
ThomasWaldmann, line 87: maybe csv is a bad example as this would expect a single row of comma separated filenames.
also, echo obviously does not work - please **actually try** the stuff you are adding as example.

maybe using NUL chars (0x00) as separator would be a good example (together with find which iirc has an option to output using that).

ThomasWaldmann (changes requested):

docs/usage/create.rst
rpolley: Sorry about that. Guess that's what I get for not testing the *entire* example command. The find with NUL separator idea seems almost more contrived because you can just not use that option on find and not specify a separator with borg. Although I'm not sure CSV is a good example either, since when I tested it just now vim added a newline to the end that tripped things up.

rpolley: Honestly it's fairly hard to find a use case for that option that isn't somewhat contrived.

ThomasWaldmann: From `find` man page:
```
       -print0
              True; print the full file name on the standard output,
              followed by a null character (instead of the newline
              character that -print uses).  This allows file names that
              contain newlines or other types of white space to be
              correctly interpreted by programs that process the find
              output.  This option corresponds to the -0 option of
              xargs.
```
This is kind of a safety feature: the \0 character is used as string terminator in C, so we can be 100% sure it is not part of a path/filename. For \n, it usually works, but we cannot be 100% sure.

docs/usage/create.rst
```diff
@@ -74,3 +74,14 @@ Examples
...
     # The root directory of the archive will be "projectA"
     $ borg create /path/to/repo::daily-projectA-{now:%Y-%m-%d} projectA
+
+    # Use external command to determine files to archive
+    # Use --paths-from-stdin with find to only backup files less than 1MB in size
+    $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
+    # Use --paths-from-command with find to only backup files from a given user
+    $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
+    # Use --paths-from-stdin with --paths-delimiter to backup files listed in a csv file
+    $ cat ~files-to-backup.csv | borg create \
```
ThomasWaldmann, line 84: it needs to have a `/` after the `*`.

also, there is still the "one row" issue. when thinking of csv, on usually has items distributed over rows, not over columns (which are rather for other attributes of same item).

rpolley: Read up on the one find option and now I see why someone would want to do that.

docs/usage/create.rst
rpolley: yeah that's correct I tested it

ThomasWaldmann: note: the command i suggested does **not** work, strange. i'll have to check whether this is consistent with similar commands.

but i have checked this, it works and looks a bit more easy maybe?
```
borg create --paths-from-command /path/to/repo::joes-files -- find ~ -user joe
```

note: the special argument "--" forces the end of option-scanning (borg shall option-scan the borg command, but not the find command).

another issue with that command is that it makes little sense usually: as the starting point for find is the home directory of the current user (who might be joe or not joe), usually all or no files will be owned by joe.

so, if we want to demonstrate `-user`, I suggest to use /srv/samba/shared as the starting point (which is a more usual place to be shared by different users). or we use something other than `-user`.

docs/usage/create.rst
```diff
@@ -74,3 +74,14 @@ Examples
     $ cd /home/user/Documents
     # The root directory of the archive will be "projectA"
     $ borg create /path/to/repo::daily-projectA-{now:%Y-%m-%d} projectA
+
+    # Use external command to determine files to archive
+    # Use --paths-from-stdin with find to only backup files less than 1MB in size
+    $ find ~ -size -1000k | borg create --paths-from-stdin /path/to/repo::small-files-only
+    # Use --paths-from-command with find to only backup files from a given user
+    $ borg create --paths-from-command -- /path/to/repo::joes-files find ~ -user joe
+    # Use --paths-from-stdin with --paths-delimiter (for example, for files with newlines in them)
```
ThomasWaldmann, line 83: ... with newlines in their filenames

ThomasWaldmann: "newlines in them" can be misunderstood to mean the file contents.

ThomasWaldmann: alternatively:

"for example, for filenames with newlines in them"

rpolley: fixed

ThomasWaldmann (approved):
LGTM

PR #5647, opened 2021-01-24
use blake2 from py36

ThomasWaldmann:
we can stop using it from the bundled code or from libb2 library because python3.6+ now has it for us.


src/borg/crypto/low_level.pyx
```diff
@@ -439,22 +431,9 @@ cdef class AES256_CTR_BLAKE2b(AES256_CTR_BASE):
...
-                if rc != -1:
-                    rc = blake2b_update(&state, data2, data2_len)
-        if rc == -1:
-            raise Exception('blake2b_update() failed')
-        rc = blake2b_final(&state, mac_buf, self.mac_len)
-        if rc == -1:
-            raise Exception('blake2b_final() failed')
+        data = bytes(self.mac_key[:128]) + bytes(data1[:data1_len]) + bytes(data2[:data2_len])
+        mac = hashlib.blake2b(data, digest_size=self.mac_len).digest()
+        mac_buf[:self.mac_len] = mac[:self.mac_len]
```
ThomasWaldmann, line 436: the tests are crashing, likely here. does somebody see what's wrong?

ThomasWaldmann: e9c4085 fixes this. is this a cython bug or was i expecting too much from its magic? @scoder

scoder: See https://github.com/cython/cython/issues/2544
It's a bug that is specific to `char*` pointers (which are common enough to make this annoying).

ThomasWaldmann: ah, at least it is known. \o/

ThomasWaldmann: so, is a loop and byte-wise assignment the best way for now (for a few bytes)?

scoder: If it's literally just a few bytes, then you probably won't notice a difference, regardless of how you do it. At some point, `memcpy()` would certainly beat it, but it can have its own bit of overhead. Non-ancient x86 CPUs have their own instruction for this, though. Can't say under which the C compiler decides to use it.

PR #5669, opened 2021-02-03
misc. crash fixes (master)

ThomasWaldmann:


ThomasWaldmann:
TODO: check whether parts of this also apply to 1.1-maint.

PR #5670, opened 2021-02-04
revert incorrect fix for put updating shadow_index, fixes #5661

### LINKED ISSUE #5661: consequences of borg put shadow_index bug?
see #5636 for the bugfix.

as this bug was present for a while in borg, let's think about its consequences here (and what we can / must do about it for existing repos).

ThomasWaldmann:
A) the compaction code needs the shadow index only for this case:

segment A: PUT x, segment B: DEL x, with A < B  (DEL shadows the PUT).

B) for the following case, we have no shadowing DEL (or rather: it does not matter,
because there is a PUT right after the DEL) and x is in the repo index,
thus the shadow_index is not needed for the special case in the compaction code:

segment A: PUT x, segment B: DEL x PUT x

see also PR #5636.

reverts f079a83fed91c1221951303f758187987be0ac47
and clarifies the code by more comments.

we keep the code deduplication of 5f32b5666af961f5256221660a7356d467f66605
and just add a update_shadow_index param to make it not look like there was
something accidentally forgotten, which was the whole reason for the reverted
"fix".

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1191,9 +1191,11 @@ class Repository:
         except KeyError:
             pass
         else:
-            # note: doing a delete first will do some bookkeeping,
-            # like updating the shadow_index, quota, ...
-            self._delete(id, segment, offset)
+            # note: doing a delete first will do some bookkeeping.
+            # we do not want to update the shadow_index here, because
+            # we know already that we will PUT to this id, so it will
+            # be in the repo index (and we won't need it in the shadow_index).
+            self._delete(id, segment, offset, update_shadow_index=False)
         segment, offset = self.io.write_put(id, data)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@timmc @enkore i basically reverted the "fix" (which was not really needed), hopefully this is correct now.

i kept the dedup refactor though and added some comments. please check.

timmc (commented):
So the shadow index is only for entries that get shadowed by a DEL or PUT _not_ in the same segment? If so, this looks good to me.

ThomasWaldmann:
Yes, only in that case the crash with bad write ordering issue described in the compact_segments code becomes a problem.

ThomasWaldmann:
Update: this was wrong. :-(

See reopened #5661.

Issue #5622, opened 2021-01-08
borg-macosx64.tgz errors with code signature in (Python) not valid for use in process using Library Validation: library load disallowed by system policy


Mehuge:
macOS Catalina 10.15.7

Downloaded https://github.com/borgbackup/borg/releases/download/1.1.15/borg-macosx64.tgz and extracted, which creates `borg-dir`. 

The first issue, is a typical and expected one.

In terminal, tried to run borg.exe and get the expected **borg.exe cannot be opened because the developer cannot be verified** message. So, using finder, navigated into borg-dir, found borg.exe and right clicked borg.exe and clicked open while holding CMD to get the alternative dialog **macOs cannot verify the developer of "borg.exe" are you sure you want to open it?** and click the Open button.

Having clicked open, because I use Parallels VM to run a windows machine, parallels tries to open borg.exe in the VM which ofc fails with **This app cannot run on your PC**. Again, not unexpected. So I renamed `borg.exe` to `borg-exe` (for want of a better name) to avoid this.

Running `./borg-exe` from the command line now gives a new dialog **"Python" cannot be opened because the developer cannot be verified**, with the following output in the terminal
```
[9278] Error loading Python lib '/Users/adf/Downloads/borg-dir/Python': dlopen: dlopen(/Users/adf/Downloads/borg-dir/Python, 10): no suitable image found.  Did find:
	/Users/adf/Downloads/borg-dir/Python: code signature in (/Users/adf/Downloads/borg-dir/Python) not valid for use in process using Library Validation: library load disallowed by system policy
```
This was the point I gave up because I had since discovered `brew install borgbackup/tap/borgbackup-fuse`.

However, as I am running through the process again to write the issue, I continued

In Finder, going back into `borg-dir`, right click Python and click open while holding CMD and select open. Get the are you sure dialog and click open. This opens a terminal window which immediately finishes (need to manually close).

Back in terminal window, I again try `./borg-exe -v` and am presented with

![image](https://user-images.githubusercontent.com/99895/103999143-78e1b680-5194-11eb-858d-50f989a0329c.png)

Clicking cancel (my options are move to bin, or cancel) I get two subsequent similar dialogs for other things its trying to load, the output in the terminal is
```
9278] Error loading Python lib '/Users/adf/Downloads/borg-dir/Python': dlopen: dlopen(/Users/adf/Downloads/borg-dir/Python, 10): no suitable image found.  Did find:
[...]

PR #5682, opened 2021-02-10
Recommend download via CLI on macOS

m3nu:
See https://github.com/borgbackup/borg/issues/5622#issuecomment-776235155


ThomasWaldmann:
thanks! backport to 1.1-maint?

Issue #383, opened 2015-11-08
deduplicated archive bigger than source


ClipperS87:
borg create  -s -p -v -C lzma,9 repo.borg::backupset source

Start time: Wed Nov  4 06:21:18 2015
End time: Sat Nov  7 20:56:20 2015
Duration: 3 days 14 hours 35 minutes 1.63 seconds
Number of files: 10

                       Original size      Compressed size    Deduplicated size
This archive:              225.65 GB            225.96 GB            225.96 GB
All archives:              225.65 GB            225.96 GB            225.96 GB

                       Unique chunks         Total chunks
Chunk index:                 3390929              3390929
-----------------------------------------------------------------------------
Filesystem shows for source (225.500.871.191 Bytes) and for repo.borg (226.355.757.056 Bytes)
, plus 854.885.865 Byte or 815,3 MB

Q1
I assume the the compression algorithm switch to "store only" as soon the result of the compression is not smaller then the source
Q2
The overhead of about 252 byte/chunk is ok and the total number of chunks can be reduced by setting different chunker-params ?

ThomasWaldmann:
A1: no, it does not (at least not in the borg code, not sure about what lzma does internally).

A2: yes

What kind of data was that? video?

ClipperS87:
3 VMWare .vmdk files in preallocated format

ThomasWaldmann:
@ClipperS87 that's unexpected then, i would expect quite some compression.

can you try lz4, zlib and lzma with default level, just:
```
borg create --compression lz4
borg create --compression zlib
borg create --compression lzma
```
lzma,9 is quite resource-intensive stuff, maybe it had a problem with getting these resources and fell back to not compressing?

ClipperS87:
other threads on the same machine did fine (as expected) so the file seem simply not to be compressed - i tried with latest .rar and maximum settings - similar results

ThomasWaldmann:
OK, seems like .vmdk might be already compressed. So further compression is unexpected.

PR #5684, opened 2021-02-11
msgpack: drop support for old buffer protocol

FelixSchwarz:
This is my attempt to solve #5683.

We should also update the msgpack version as [requested by @ThomasWaldmann](https://github.com/borgbackup/borg/issues/5683#issuecomment-777781315) and [@RonnyPfannschmidt](https://github.com/borgbackup/borg/issues/5683#issuecomment-777794854). The version is currently `version = (0, 5, 6)`.

Updating to 0.5.7 is easy though I also like the idea of using a more descriptive version number. As far as I can see nothing in borg uses the msgpack version number so we could also change to `version = (0, 5, 6, '+borg')` easily. Comments?

----
Commit message

The old buffer protocol was deprecated since Python 3.0 and will be removed entirely in Python 3.10.

This is a backport of

    commit 9ae43709e42092c7f6a4e990d696d9005fa1623d
    Author: Inada Naoki <songofacandy@gmail.com>
    Date:   Thu Dec 5 20:20:53 2019 +0900

        Drop old buffer protocol support (#383)

Backporting by me, any errors are likely due to backporting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,9 @@ jobs:
             - os: ubuntu-20.04
               python-version: 3.9
               toxenv: py39
+            - os: ubuntu-20.04
+              python-version: 3.10
+              toxenv: py310
             - os: macos-latest
               # note: it seems that 3.8 and 3.9 are currently broken,
               # neverending RuntimeError crashes...
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,27 +52,27 @@ jobs:
       matrix:
         include:
             - os: ubuntu-20.04
-              python-version: 3.5
+              python-version: '3.5'
               toxenv: py35
             - os: ubuntu-20.04
-              python-version: 3.6
+              python-version: '3.6'
               toxenv: py36
             - os: ubuntu-20.04
-              python-version: 3.7
+              python-version: '3.7'
               toxenv: py37
             - os: ubuntu-20.04
-              python-version: 3.8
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
I'ld suggest `+borg1` as the local part of the version number, so it is clear how to proceed in case we need to patch again.

ThomasWaldmann:
There is some code in borg that check for supported msgpack versions, it has to be adapted maybe (we'll see if tests fail).

ThomasWaldmann:
@FelixSchwarz are you still working on adding the version change or shall i do that?

ThomasWaldmann:
I plan to make a 1.1.16 release because of this and also because of the pyinstaller regression.

ThomasWaldmann:
Looks like fixing this would need Cython 0.29.22, which is not released yet:
```
E   ImportError: /home/runner/work/borg/borg/.tox/py310/lib/python3.10/site-packages/borg/item.cpython-310-x86_64-linux-gnu.so: undefined symbol: _PyGen_Send
```

ThomasWaldmann:
Working on the versioning stuff  now...

ThomasWaldmann:
@FelixSchwarz ^^^ merged! Thanks for backporting! \o/

ThomasWaldmann:
Note: some parts of this PR need to be forward ported to master:
- quoting the python versions
- declaring support and testing for py310

FelixSchwarz:
Thank you. As you can see I'm really swamped atm so I failed to respond in time.

PR #5687, opened 2021-02-12
docs: remove /var/cache exclusion in example commands

### LINKED ISSUE #5625: do not exclude /var/cache
got an email from nicolas, noting:

```
The example scripts in quickstart.rst and deployment/automated-local.rst 
exclude /var/cache from the backup.

According to https://unix.stackexchange.com/questions/150487/how-should-i-restore-var-cache 
it is probably a good idea not to blindly exclude /var/cache.

I'd suggest not using it as an example of something to exclude.
```

https://unix.stackexchange.com/questions/150487/how-should-i-restore-var-cache

mtrower:
This is generally a poor idea and shouldn't be promoted through examples.

Fixes #5625


ThomasWaldmann:
Thanks! Can you backport to 1.1-maint also?

Issue #5660, opened 2021-01-28
macOS: test borg mount with macFUSE 4.x


ThomasWaldmann:
macFUSE 4.x is the successor of OSXfuse 3.x - does it work as expected?

m3nu:
Works fine with `llfuse`, but not `pyfuse3`, since macFUSE only supports fuse v2 API, not v3:

> pyfuse3 requires a libfuse 3.x compatible API. OSXFUSE only supports the libfuse 2.9 API. (from [here](https://github.com/libfuse/pyfuse3/issues/32))

Steps done to test:

1. Install macFUSE
2. Restart
3. `$ BORG_OPENSSL_PREFIX=/usr/local/Cellar/openssl@1.1/1.1.1i pip install '.[llfuse]'`
4. `$ borg mount /Users/manu/Downloads/test-repo::1 ~/Downloads/mnt`
5. Archive is mounted as expected

m3nu:
So this is also a docs issue to mention that Borg on macOS only works with `llfuse`?

ThomasWaldmann:
Well, yes, but we need to be careful because that could change over time, when macFUSE switches to fuse3 some day.

m3nu:
So anything left to do here? Is it worth adding a note on using `llfuse` only in the docs here: https://borgbackup.readthedocs.io/en/master/installation.html#macos ?

ThomasWaldmann:
https://borgbackup.readthedocs.io/en/master/installation.html#dependencies guess we should just tell there what fuse versions llfuse and pyfuse3 needs, this will apply to all operating systems.

ThomasWaldmann:
About BORG_OPENSSL_PREFIX - isn't openssl discovered via pkg-config in master branch?
Did you:
- brew install pkg-config
- pip install pkgconfig
- then pip install borgbackup[llfuse]
?

m3nu:
No. By default the Homebrew openssl formula is NOT linked. Adding it to pkg-config is mentioned as optional step.

```
$ brew info openssl
openssl@1.1: stable 1.1.1i (bottled) [keg-only]
Cryptography and SSL/TLS Toolkit
https://openssl.org/
...
openssl@1.1 is keg-only, which means it was not symlinked into /usr/local,
because macOS provides LibreSSL.

If you need to have openssl@1.1 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> ~/.zshrc

[...]

PR #5691, opened 2021-02-15
Add short note on pointing setup.py to openssl on macOS

m3nu:
See comment https://github.com/borgbackup/borg/issues/5660#issuecomment-778704981

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -50,7 +50,7 @@ Fedora/RHEL  `Fedora official repository`_                 ``dnf install borgbac
 FreeBSD      `FreeBSD ports`_                              ``cd /usr/ports/archivers/py-borgbackup && make install clean``
 macOS        `Homebrew`_                                   | ``brew install borgbackup`` (official formula, **no** FUSE support)
                                                            | **or**
-                                                           | ``brew install --cask osxfuse`` (`private Tap`_, FUSE support) 
+                                                           | ``brew install --cask macfuse`` (`private Tap`_, FUSE support) 
                                                            | ``brew install borgbackup/tap/borgbackup-fuse``
 Mageia       `cauldron`_                                   ``urpmi borgbackup``
 NetBSD       `pkgsrc`_                                     ``pkg_add py-borgbackup``
@@ -258,7 +258,7 @@ For FUSE support to mount the backup archives, you need at least version 3.0 of
 macFUSE, which is available via `github
 <https://github.com/osxfuse/osxfuse/releases/latest>`__, or Homebrew::
 
-    brew install --cask osxfuse
+    brew install --cask macfuse
 
 When installing Borg via ``pip``, be sure to install the ``llfuse`` extra,
 since macFUSE only supports FUSE API v2. Also, since Homebrew won't link
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -260,11 +260,11 @@ macFUSE, which is available via `github
 
     brew install --cask osxfuse
 
-When installing Borg via ``pip``, be sure to use ``pip install borgbackup[llfuse]``,
-since macFUSE only supports FUSE API v2.
+When installing Borg via ``pip``, be sure to install the ``llfuse`` extra,
+since macFUSE only supports FUSE API v2. Also, since Homebrew won't link
+the installed ``openssl`` formula, point pkg-config to the correct path::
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -260,6 +260,12 @@ macFUSE, which is available via `github
 
     brew install --cask osxfuse
 
+When installing Borg via ``pip``, be sure to use ``pip install borgbackup[llfuse]``,
+since macFUSE only supports FUSE API v2.
+
+Since Homebrew won't link the installed ``openssl`` formula, set the ``BORG_OPENSSL_PREFIX`` 
+environment variable to ``/usr/local/Cellar/openssl`` before running ``pip install``.
```
ThomasWaldmann, line 267: is there no `@1.1` for this case?

m3nu: It works without, but I'm not sure why.

Make it `@X.X` to cover future version changes?

ThomasWaldmann: I don't know any specifics, i just remembered from other occasions that there often used to be a `@1.1` appended to the openssl path. not sure what the path without points to, if there is such a path.

can you check on your system whether the path exists / whether there are symlinks?

m3nu: Already checked. Only the `@1.1` path is there. No symlink. Inside it is a more specific version folder.

So there is already some flexibility finding the folder. Maybe that's why it works. since this line in `setup_crypto.py` is at least one folder level off too: `lib_dir = os.path.join(system_prefix, 'lib')`

```
$ ls -l /usr/local/Cellar/ | grep openssl
drwxr-xr-x  3 manu  staff  96 Jan 13 15:33 openssl@1.1
```

```
$ tree -L 2 /usr/local/Cellar/openssl@1.1
/usr/local/Cellar/openssl@1.1
└── 1.1.1i
    ├── AUTHORS
    ├── CHANGES
    ├── INSTALL_RECEIPT.json
    ├── LICENSE
    ├── NEWS
    ├── README
    ├── bin
    ├── include
    ├── lib
    └── share

5 directories, 6 files
```

---

On the other hand, the pkg-config folder without version is  real symlink. Maybe we use that instead of `BORG_OPENSSL_PREFIX `?

```
$ ls -l /usr/local/opt/ | grep openssl
lrwxr-xr-x  1 manu  admin  28 Jan 13 11:48 openssl -> ../Cellar/openssl@1.1/1.1.1i
lrwxr-xr-x  1 manu  admin  28 Jan 13 11:48 openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1i
```

Would be this: `PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig`

ThomasWaldmann: guess if we want 1.1, we should use the @1.1 version.
Some people might have 1.0 still available and there might be some newer openssl version in the future.

if the pkgconfig way can be made working, i guess we should use that, pkgconfig's job is to make it easy to find headers/libs/etc.

ThomasWaldmann:
guess we'll also want to check if there is a similar issue in 1.1-maint.
the setup.py stuff there works quite differently though.

should be checked before 1.1.16 release.

docs/installation.rst
m3nu: In that case, we should point it to the pkgconfig path of openssl 1.1 specifically:

Screen Shot 2021-02-17 at 08 44 25

Any style preference on how to write this? Show it together with the `pip` command?

m3nu:
`1.1-maint` will just find openssl due to the hardcoded possible prefix `/usr/local/opt/openssl` in `setup.py:171`. So no change or env var needed to install it.

docs/installation.rst
ThomasWaldmann: How about:
```
When installing Borg via ``pip``, be sure to install the ``llfuse`` extra,
since macFUSE only supports FUSE API v2. Also, since Homebrew won't link
the installed ``openssl`` formula, point pkg-config to the correct path:

    PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig" pip install borgbackup[llfuse]
```

m3nu: Works too and it's all in one. 😄

m3nu: Updated PR

ThomasWaldmann:
Thanks!

PR #5703, opened 2021-02-19
add mypy, add to tox/CI

ThomasWaldmann:
some troublesome places have mypy silenced, by either of 2 methods:
- added some packages/modules to setup.cfg that do not support mypy yet
- added some inline "# type: ignore" comments

also: refactor borg.platform init as i could not get rid
of the redefinition errors, so make sure all symbols are defined
(or imported) exactly once and not redefined.
got rid of borg.platform.base.{process_alive,local_pid_alive} as
they were always redefined and never used from there.


ThomasWaldmann:
superceded by #6862

Issue #3765, opened 2018-04-11
borg diff: add json output


phdoerfler:
Right now borg diff shows something like this:

```
+27.4 kB  -27.4 kB var/vmail/example.com/foobar/Auto/Steam und Co/dovecot-uidlist
+490.4 kB -489.3 kB var/vmail/example.com/foobar/Auto/Steam und Co/dovecot.index.cache
 +29.6 kB  -29.3 kB var/vmail/example.com/foobar/Auto/Steam und Co/dovecot.index.log
added      62.77 kB var/vmail/example.com/foobar/Auto/Steam und Co/new/1520542202.M174402P24533.turtle,S=62768,W=63599
  +4.5 kB   -4.5 kB var/vmail/example.com/foobar/maildirsize
```

And that's great. Except these human readable size informations are a bit tough for a machine to parse. Ideally there would be an option to output machine readable sizes (aka just the bytes, without any prefix).
Speaking of prefix: Are these kB as in 1000 Bytes or KiB as in 1024 Bytes? While at it maybe this could be clarified, too. Thanks!

ThomasWaldmann:
Did you check whether we have json output support for that?

phdoerfler:
I did indeed check the `borg help diff` for that. The only thing related to JSON I found is this:

```
  --log-json            Output one JSON object per log line instead of
                        formatted text.
```

So no luck there I am afraid.

ThomasWaldmann:
OK, if we do not have that yet, adding json output seems to be a good idea.

JonasKvarnstrom:
In that case I would also like to request including both the uncompressed and compressed size of each change.  As far as I can understand the Borg source code, the current output refers to the uncompressed size.

Ashmodei:
I propose to add an option `--json-output` for diff.
And a diff output should be something like that:
```
{
 added: [
        {path: '/path/to/file', change: '27.4 kB'}
        ],
 modified: [
        {path: '/path/to/file', change: '+490.4 kB -489.3 kB'}
        ],
 deleted: [
        {path: '/path/to/file', change: 'directory'}
        ]
}
```
Given this way, users will get easy way to parse the output, cause they will have separate groups and fields.
So, I'm going to add inner function `print_output_json` to `do_diff`, which will produce the output like above from diffs generator. 

Also for not-human-readable format we can add an option `--bytes` and show file size without units. I think it would be nice to have this option and for standard output too.
[...]

PR #5708, opened 2021-02-24
add --json-lines option to diff command

rblenis:
fix issue #3765

### DIFF
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -231,11 +231,16 @@ Standard output
 *stdout* is different and more command-dependent than logging. Commands like :ref:`borg_info`, :ref:`borg_create`
 and :ref:`borg_list` implement a ``--json`` option which turns their regular output into a single JSON object.
 
+Some commands, like :ref:`borg_list` and :ref:`borg_diff`, can produce *a lot* of JSON. Since many JSON implementations
+don't support a streaming mode of operation, which is pretty much required to deal with this amount of JSON, these
+commands implement a ``--json-lines`` option which generates output in the `JSON lines <http://jsonlines.org/>`_ format,
+which is simply a number of JSON objects separated by new lines.
+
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
Please tell me when this is in sync with the already merged 1.1-maint PR #5710.

ThomasWaldmann:
So, this is just a forward-port of the merged 1.1-maint changeset?

rblenis:
yes. all the diff engine stuff moved to item.pyx, so comparing to 1.1.x is a pain. Also, just remembered - it looks like a number of the asserts in testsuite/archiver.py are changed to self.assert_in(), ...  (or are new tests). Looks like to give better failure messages.  Not all of the asserts in that file have been changed. Should I make use of the 'better' (newer) assert plumbing in the diff tests? And if so, should I add a task of perhaps cleaning up all the test code to use the new plumbing?

rblenis:
Also, when I submit a PR, should I also include changes in the changelog, or is the changelog updated by whoever starts the release build?

ThomasWaldmann:
- using `self.assert_xxx` methods is unittest style - this should be only used for specific tests that are run within borg's self-test. usually we use the more comfortable pytest style which is just `assert ...`.
- no need to update changes.rst, i do that when releasing. but doing good commit comments is helpful, so i can use them for the changelog.

rblenis:
When are borg's self-tests run? Is there some documentation on this?  Should selftest(unittest) be mixed in the same file (or even same class) as pytest? - cause that's happening in testsuite/archiver.py, yet some comments in other files seem to indicate not to mix in the same file. Should unittest be organized in a different folder? A lot of questions - I need to investigate a little more.

ThomasWaldmann:
For the self tests, there's a list and a count somewhere. They always run when borg starts.

ThomasWaldmann:
So, is there anything left to do here?

rblenis:
all done.

ThomasWaldmann:
Thanks!

PR #5714, opened 2021-02-28
add support and testing for python 3.10

ThomasWaldmann:
fwd port of the py310 testing part of #5684.


FelixSchwarz:
btw: Cython seems to be problematic with Python 3.10a6. Fedora/Red Hat developers are tackling the problem upstream but just so you are aware of the problem: https://github.com/cython/cython/issues/4025#issuecomment-800967701

Issue #5295, opened 2020-08-12
ACL code does a lot of user/group<->uid/gid lookups


biji:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Issue

#### Your borg version (borg -V).

client: borg 1.1.13

server: borg 1.1.13

#### Operating system (distribution) and version.

client: Centos 6

server: Centos 7

#### Hardware / network configuration, and filesystems used.

Backup via internet

#### How much data is handled by borg?

450GB

302366 files

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

borg create  -s -p  -C lzma --list --filter=AME --exclude-from=/root/exclude-backup.txt --files-cache=ctime,size borg@server-ip:backup/pdc::pdc-{now} /home/data/

## Describe the problem you're observing.

I experience slow 2nd+ backup with winbind as nss provider after a after successful 1st backup.

First backup takes 23 hours, next backups still takes >20 hours

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

I have tried to remove /root/.cache , but still slow

Thanks

#### Include any warning/errors/backtraces from the system logs

none

ThomasWaldmann:
Easiest explanation is that something frequently/regularly changes the `ctime` of many fs items, e.g. `chmod/chown -R ...`.
Or, that the full path is changing (does not look like, you wrote that you use `/home/data/`).

Of course, after removing borg's cache, next backup is expected to be slow, but the one after that should then be faster again.

Another reason why you maybe do not see a huge speedup could be that you have lots of small files, which are in general less efficient to deal with. In that case 20h < 23h is maybe all the speedup you'll get.

biji:
Yes, there are lots of small files: roaming storage for windows users. Any suggestion for speeding up backup lots of small files?
I'll try again using mtime,size

Thanks

biji:
Hi, i've tried 3 backups with clear borg's cache, using files-cache=ctime,size
Excluding some files, now files around 130.000

Time required for first backup is: 23 hours , next backups is 17 hours , and new files is not many

Is there any way to make next backup faster than 17 hours? 

```
Archive name: pdc-2020-08-13
Archive fingerprint: 9c72ae92713cf9313f85393f57d49edd7dd78156007666fca081c2b7161eee54
Comment:
Hostname: pdc
Username: root
[...]

PR #5718, opened 2021-03-01
release 1.1.16

ThomasWaldmann:


detoxhby:
Do we have any estimate on the possible release date? (issues in production getting worse lately by #5295 as storage size expands, we have to make a decision by the end of the month to drop borg as a backup solution; sadly we can't wait more due to angry clients)

Thanks ✋🏻 Your precious work is appreciated!

ThomasWaldmann:
@detoxhby I initially wanted to support py310 with that release, but it's not really clear yet how to do that best.

so I'll push that to a future release and release borg 1.1.16 soon (without py310 support, py310 is still alpha anyway).

ThomasWaldmann:
@detoxhby https://github.com/borgbackup/borg/releases/tag/1.1.16

PR #5726, opened 2021-03-10
cleanup code style checks

rblenis:
Update flake8 config to enforce desired 120 char line length and fix non-conforming files.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -15,16 +15,15 @@ if hasattr(pytest, 'register_assert_rewrite'):
     pytest.register_assert_rewrite('borg.testsuite')
 
 
-import borg.cache
-from borg.logger import setup_logging
+import borg.cache  # noqa: E402
+from borg.logger import setup_logging  # noqa: E402
 
 # Ensure that the loggers exist for all tests
 setup_logging()
 
-from borg.testsuite import has_lchflags, has_llfuse, has_pyfuse3
-from borg.testsuite import are_symlinks_supported, are_hardlinks_supported, is_utime_fully_supported
-from borg.testsuite.platform import fakeroot_detected, are_acls_working
-from borg import xattr
+from borg.testsuite import has_lchflags, has_llfuse, has_pyfuse3  # noqa: E402
+from borg.testsuite import are_symlinks_supported, are_hardlinks_supported, is_utime_fully_supported  # noqa: E402
+from borg.testsuite.platform import fakeroot_detected  # noqa: E402
 
 
 @pytest.fixture(autouse=True)
@@ -67,7 +66,8 @@ class DefaultPatches:
         self.org_cache_wipe_cache = borg.cache.LocalCache.wipe_cache
 
         def wipe_should_not_be_called(*a, **kw):
-            raise AssertionError("Cache wipe was triggered, if this is part of the test add @pytest.mark.allow_cache_wipe")
+            raise AssertionError("Cache wipe was triggered, if this is part of the test add "
+                                 "@pytest.mark.allow_cache_wipe")
         if 'allow_cache_wipe' not in request.keywords:
             borg.cache.LocalCache.wipe_cache = wipe_should_not_be_called
         request.addfinalizer(self.undo)
--- a/scripts/errorlist.py
+++ b/scripts/errorlist.py
@@ -2,7 +2,7 @@
 
 from textwrap import indent
 
-import borg.archiver
+import borg.archiver  # noqa: F401 - need import to get Error and ErrorWithTraceback subclasses.
 from borg.helpers import Error, ErrorWithTraceback
 
 classes = Error.__subclasses__() + ErrorWithTraceback.__subclasses__()
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,13 +4,92 @@ markers =
     allow_cache_wipe
 
 [flake8]
+# for reference ...
+#   E121 continuation line under-indented for hanging indent
+#   E122 continuation line missing indentation or outdented
+#   E123 closing bracket does not match indentation of opening bracket's line
+#   E125 continuation line with same indent as next logical line
+#   E126 continuation line over-indented for hanging indent
+#   E127 continuation line over-indented for visual indent
+#   E128 continuation line under-indented for visual indent
+#   E221 multiple spaces before operator
+#   E226 missing whitespace around arithmetic operator
+#   E261 at least two spaces before inline comment
+#   E265 block comment should start with '# '
+#   E301 expected 1 blank line
+#   E305 expected 2 blank lines after class or function definition
+#   E401 multiple imports on one line
+#   E402 module level import not at top
+#   E501 line too long
+#   E722 do not use bare except
+#   E731 do not assign a lambda expression, use def
+#   E741 ambiguous variable name
+#   F401 import unsused
+#   F403 from ... import * used, unable to detect undefined names
+#   F405 undefined or defined from star imports
+#   F811 redef of unused var
+#   F821 undefined name
+#   W391 blank line at end of file
+#   #### Pick either W503, or W504 - latest recommendation from pep8 is to ignore W503
+#   W503 line break before binary operator
+#   W504 line break after binary operator
+
+# borg code style guidelines:
+ignore = E226, W503
+
+# Code style violation exceptions:
 # please note that the values are adjusted so that they do not cause failures
 # with existing code. if you want to change them, you should first fix all
 # flake8 failures that appear with your change.
-ignore = E122,E123,E125,E126,E127,E128,E226,E402,E722,E731,E741,
-         F401,F405,F811,
-         W504
-# line length long term target: 120
-max-line-length = 255
-exclude = build,dist,.git,.idea,.cache,.tox,docs/conf.py
+per_file_ignores =
+    conftest.py:E402,E501,F401
+    docs/conf.py:E121,E126,E265,E305,E401,E402
+    scripts/errorlist.py:F401
+    src/borg/archive.py:E122,E125,E127,E402,E501,F401,F405,W504
+    src/borg/archiver.py:E126,E127,E128,E501,E722,E731,E741,F401,F405,W504
+    src/borg/cache.py:E127,E128,E402,E501,E722,W504
+    src/borg/fuse.py:E402,E501,E722,W504
+    src/borg/fuse_impl.py:F811
+    src/borg/locking.py:E128,E501,E722
+    src/borg/remote.py:E128,E501,F405
+    src/borg/repository.py:E126,E128,E501,F401,F405,W504
+    src/borg/shellpattern.py:E501
+    src/borg/upgrader.py:E501
+    src/borg/xattr.py:E402
+    src/borg/crypto/key.py:E125,E128,E402,E501,F401,F405,W504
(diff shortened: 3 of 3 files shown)


ThomasWaldmann:
Thanks for investing some work on this.

But I have to admit that after looking at the diff, I don't really like it due to the issues it is introducing:
- some stuff that is used as a string changed contents (due to using a multi-line string with indentation)
- some stuff that was just a little over the limit now has 1 long line and only a few chars on the next (new) line, which looks more ugly than before. of course this is a rather fundamental problem when limiting the line length, except that for higher limits it happens less frequently.
- if we apply this change only to master branch, it will make backporting changes to 1.1-maint branch more difficult (guess they might be rejected just due to different formatting) - would you backport the whitespace change to 1.1-maint to avoid that?
- if we apply this change before other currently open / active PRs, it might cause annoying conflicts

rblenis:
I would be happy to backport. I also didn't like some of the formatting changes, and have gone through and cleaned up some and added line specific noqa in places to restore the text string where it may have caused issues. I did this prior to seeing your comments, so I will look over it again.

rblenis:
Take a look at the changes in setup.cfg. This can at least contain the disabled checks to specific files, and then possibly tell people if the work on a file, fix the violations in the file and remove them from the exceptions in setup.cfg

setup.cfg
```diff
@@ -7,10 +7,38 @@ markers =
 # please note that the values are adjusted so that they do not cause failures
 # with existing code. if you want to change them, you should first fix all
 # flake8 failures that appear with your change.
-ignore = E122,E123,E125,E126,E127,E128,E226,E402,E722,E731,E741,
-         F401,F405,F811,
-         W504
+ignore =
+    E122,   # continuation line missing indentation or outdented
+    E123,   # closing vracket does not match indentation of opening bracket's line
```
ThomasWaldmann, line 12: bracket

src/borg/archiver.py
```diff
@@ -527,7 +530,8 @@ def create_inner(archive, cache, fso):
                         except (FileNotFoundError, PermissionError) as e:
                             self.print_error('Failed to execute command: %s', e)
                             return self.exit_code
-                        status = fso.process_pipe(path=path, cache=cache, fd=proc.stdout, mode=mode, user=user, group=group)
+                        status = fso.process_pipe(path=path, cache=cache, fd=proc.stdout, mode=mode, user=user,
+                                                  group=group)
```
ThomasWaldmann, line 534: mode, user, group somehow below together, maybe move them all to the 2nd line.

src/borg/archiver.py
```diff
@@ -578,7 +582,8 @@ def create_inner(archive, cache, fso):
                         group = args.stdin_group
                         if not dry_run:
                             try:
-                                status = fso.process_pipe(path=path, cache=cache, fd=sys.stdin.buffer, mode=mode, user=user, group=group)
+                                status = fso.process_pipe(path=path, cache=cache, fd=sys.stdin.buffer, mode=mode,
+                                                          user=user, group=group)
```
ThomasWaldmann, line 586: same here.

src/borg/archiver.py
```diff
@@ -844,9 +851,11 @@ def do_extract(self, args, repository, manifest, key, archive):
         """Extract archive contents"""
         # be restrictive when restoring files, restore permissions later
         if sys.getfilesystemencoding() == 'ascii':
-            logger.warning('Warning: File system encoding is "ascii", extracting non-ascii filenames will not be supported.')
+            logger.warning('Warning: File system encoding is "ascii", extracting non-ascii filenames will not be '
+                           'supported.')
             if sys.platform.startswith(('linux', 'freebsd', 'netbsd', 'openbsd', 'darwin', )):
-                logger.warning('Hint: You likely need to fix your locale setup. E.g. install locales and use: LANG=en_US.UTF-8')
+                logger.warning('Hint: You likely need to fix your locale setup. E.g. install locales and use: '
+                               'LANG=en_US.UTF-8')
```
ThomasWaldmann, line 858: here: indent into parens...

src/borg/archiver.py
```diff
@@ -1121,7 +1130,8 @@ def item_to_tarinfo(item, original_path):
             elif modebits == stat.S_IFIFO:
                 tarinfo.type = tarfile.FIFOTYPE
             else:
-                self.print_warning('%s: unsupported file type %o for tar export', remove_surrogates(item.path), modebits)
+                self.print_warning('%s: unsupported file type %o for tar export', remove_surrogates(item.path),
+                    modebits)
```
ThomasWaldmann, line 1134: here: different style.

src/borg/archiver.py
```diff
@@ -2797,10 +2808,10 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
             if sort_by:
                 sort_by_default = 'timestamp'
-                filters_group.add_argument('--sort-by', metavar='KEYS', dest='sort_by',
-                                           type=SortBySpec, default=sort_by_default,
-                                           help='Comma-separated list of sorting keys; valid keys are: {}; default is: {}'
-                                           .format(', '.join(AI_HUMAN_SORT_KEYS), sort_by_default))
+                filters_group.add_argument('--sort-by', metavar='KEYS', dest='sort_by', type=SortBySpec,
+                    default=sort_by_default,
+                    help='Comma-separated list of sorting keys; valid keys are: {}; default is: {}'
+                         .format(', '.join(AI_HUMAN_SORT_KEYS), sort_by_default))
```
ThomasWaldmann, line 2814: shall we just increase to 160 or so to not have to reformat such stuff?

src/borg/cache.py
```diff
@@ -277,7 +277,8 @@ def load(self):
         self.timestamp = self._config.get('cache', 'timestamp', fallback=None)
         self.key_type = self._config.get('cache', 'key_type', fallback=None)
         self.ignored_features = set(parse_stringified_list(self._config.get('cache', 'ignored_features', fallback='')))
-        self.mandatory_features = set(parse_stringified_list(self._config.get('cache', 'mandatory_features', fallback='')))
+        self.mandatory_features = set(parse_stringified_list(self._config.get('cache', 'mandatory_features',
+            fallback='')))
```
ThomasWaldmann, line 281: more than 120 would also avoid such formatting, which is worse IMHO.

src/borg/archiver.py
rblenis: Before I made these changes there were lines longer than 200! The problem, at least in my setup, is that past 120, you can't do side-by-side diffs without scrolling back and forth to see the whole line. Even the github diffs on this page (shown top/bottom, not side-by-side) only show about 96 characters.  I think we can still make it pretty at 120, and in cases where it really is best to be longer, we can always '# noqa' or '# noqa: E501' the line or file.

src/borg/cache.py
rblenis: In situations like this with 3+ levels of nested function calls, would it be better to use a temp variable, such as this:
`
  \# use a temp, easy to understand and comment.
  cfg = self._config.get('cache', 'mandatory_features', fallback='')   # get configuration
  self.mandatory_features = set(parse_stringified_list(cfg))   # parse config string and set as mandatory feature...

  \# vs 3 level deep nested function calls can be a pain to grok
  self.mandatory_features = set(parse_stringified_list(self._config.get('cache', 'mandatory_features',
                                      fallback='')))
`

src/borg/archiver.py
ThomasWaldmann: ok

setup.cfg
```diff
@@ -7,10 +7,48 @@ markers =
...
+    src/borg/testsuite/crypto.py:E126,E741
+    src/borg/testsuite/helpers.py:E126,E127
+    src/borg/testsuite/locking.py:E126,E722
+    src/borg/testsuite/patterns.py:E123
+    src/borg/testsuite/platform.py:F811
+    src/borg/testsuite/shellpattern.py:E123
+
 # line length long term target: 120
-max-line-length = 255
+max-line-length = 120
```
ThomasWaldmann, line 52: btw, the tests output a warning for this, telling that it should be `max_line_length`.

rblenis: ok. fixed this and finished adding all the specific files for flake8 disabled checks.  Looks like from 1.1.x to 1.2 there are 12 more offending files, so getting at least the setup.cfg change in will have minimal impact on active PRs and prevent more code from violating coding styles.

rblenis:
It looks like the wrapping of lines work actually violates a lot of the other coding styles that are currently globally disabled in setup.cfg.  I'm beginning to think that the better approach is to use the tightened up setup.cfg and then start addressing issues file-by file so that there are not tons of changes across 30+ files. Perhaps only a few files at a time, if there are minor changes, and single files at a time when lots of changes to get conforming code.  How much do we want to meet pep8?  Should I spend some time looking at code formatter like autopep8, so that we don't really worry about code style issues during review, since it should be handled by lint and a code formatter tool?

rblenis:
I renamed this to just cleaning up the codestyle checks, and backed off from changing so many files at once. This way, this doesn't touch existing code files, just narrows the enforcement so that new files will be checked more strictly. I still have all the changes made for line length in another branch, but now have it structured as individual file patches, that fix all the various code style issues in that file, not just line length. We can then merge them in, in smaller batches to reduce conflicts, if you like.

rblenis:
@ThomasWaldmann, ready for review. Includes code style fixes in 2 files, as examples of how to fix the file and remove it from the list of  exceptions.

conftest.py
```diff
@@ -15,16 +15,15 @@
...
 # Ensure that the loggers exist for all tests
 setup_logging()
 
-from borg.testsuite import has_lchflags, has_llfuse, has_pyfuse3
-from borg.testsuite import are_symlinks_supported, are_hardlinks_supported, is_utime_fully_supported
-from borg.testsuite.platform import fakeroot_detected, are_acls_working
-from borg import xattr
+from borg.testsuite import has_lchflags, has_llfuse, has_pyfuse3    # noqa: E402
+from borg.testsuite import are_symlinks_supported, are_hardlinks_supported, is_utime_fully_supported   # noqa: E402
+from borg.testsuite.platform import fakeroot_detected               # noqa: E402
```
ThomasWaldmann, line 26: if you add a inline `# ...` comment, please have exactly 2 spaces between that and the code.

trying to align comments is a noble goal and looks pretty at first, but doesn't work in general (as you see in 25) and also leads to a lot of readjusting over time, when other lines get comments and then a lot of pre-existing comments need to be re-adjusted. so, it is easier and more regular to not try to align comments.

scripts/errorlist.py
```diff
@@ -2,7 +2,7 @@
 
 from textwrap import indent
 
-import borg.archiver
+import borg.archiver    # noqa: F401 - need import to get Error and ErrorWithTraceback subclasses.
```
ThomasWaldmann, line 5: same here.

ThomasWaldmann (commented):
much better approach now, less likely to conflict or cause other issues.

one minor nitpick to fix and then we could merge this to master / backport to 1.1-maint.

other improvements / more work can be done in later PRs.

rblenis:
oops, just rebased before seeing these comments. will go make appropriate changes.

rblenis:
made the requested changes for this PR.

PR #5728, opened 2021-03-12
1.1-cleanup code style checks

rblenis:
@ThomasWaldmann, how about I start with this to grandfather in specific files that violate some checks. This will make sure that all other files meet the requirements and allow us to fix smaller sets of files, reducing conflicts for active PRs.


rblenis:
I added 2 commits as examples of the changes to make files code style compliant and removing them from the exception list.

PR #5743, opened 2021-03-21
Add option to express sizes in powers of 1024 (KiB, MiB…)

rom1v:
A (minor) source of confusion for me are the sizes expressed in powers of 1000 instead of 1024 ([1KB = 1000B vs 1KiB = 1024B](https://en.wikipedia.org/wiki/Binary_prefix)), which is not consistent with the units used by many commands (`ls -h`, `df -h`, `du -h`, `dd`…).

To make it consistent,  add a common option (`--iec`).

For now, it's implemented for:
- info
- create --stats

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -54,8 +54,9 @@ has_link = hasattr(os, 'link')
 
 class Statistics:
 
-    def __init__(self, output_json=False):
+    def __init__(self, output_json=False, iec=False):
         self.output_json = output_json
+        self.iec = iec
         self.osize = self.csize = self.usize = self.nfiles = 0
         self.osize_parts = self.csize_parts = self.usize_parts = self.nfiles_parts = 0
         self.last_progress = 0  # timestamp when last progress was shown
@@ -75,7 +76,7 @@ class Statistics:
     def __add__(self, other):
         if not isinstance(other, Statistics):
             raise TypeError('can only add Statistics objects')
-        stats = Statistics(self.output_json)
+        stats = Statistics(self.output_json, self.iec)
         stats.osize = self.osize + other.osize
         stats.csize = self.csize + other.csize
         stats.usize = self.usize + other.usize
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
Thanks for the PR. Yeah, guess we want some option like that. Not sure about the option name, I don't really like `--iec`, but I don't have a better suggestion currently.

There is one big issue seen in the tests:
```
E       TypeError: Object of type FileSize is not JSON serializable
```
Guess it is because you removed the subclassing from `int`.

ThomasWaldmann:
https://en.wikipedia.org/wiki/Gigabyte (see "Definition")

So, `--iec` seems appropriate if we do not find something better.

ThomasWaldmann:
@rom1v did you see my feedback?

rom1v:
Hi @ThomasWaldmann,

Thank you for your feedback. Yes, indeed there are still some fixes/work to do on this PR.

I'm not sure I will have the time to work on it soon though, as I have other priorities (it was more a feature request with some experiments). Feel free to take ownership/close :wink:

Keep up the good work, borg is awesome :+1:

ThomasWaldmann:
@rom1v OK, thanks for the PR, I'll try to finish it.

ThomasWaldmann:
OK, I rebased it on current master and also tried to fix the subclassing from `int`. Let's see if it works now...

ThomasWaldmann:
@rom1v looks like the fix worked, have a look...

ThomasWaldmann:
@rom1v thanks for the PR - finished and merged!

rom1v:
Oh, thank you :+1: I wanted to test this evening or later this week when I noticed your work-on-progress commits, but then I'll test the master branch directly :wink:

rom1v:
@ThomasWaldmann I just tested with `borg info`, it works fine :+1:

```console
$ borg info --iec ::myarchive
...
This archive:               9.27 GiB             9.07 GiB             4.01 MiB
All archives:               3.44 TiB             3.39 TiB           530.99 GiB
```

However, if i pass `--iec` before `info`, it "fails" silently:

```console
$ borg --iec info ::myarchive
...
This archive:                9.95 GB              9.74 GB              4.21 MB
All archives:                3.79 TB              3.73 TB            570.14 GB
```

ThomasWaldmann:
Can you check how other supposed-to-be-common options behave?

rom1v:
For example, `--debug` works both in `borg --debug list` and `borg list --debug`.

ThomasWaldmann:
@rom1v see #5789.

PR #5745, opened 2021-03-21
docs: explain hash collision

ImUrX:
backport from master

hello, im back after 9 months! remembering this was fun. good luck and keep doing what you guys like to do!


ThomasWaldmann:
Thanks and welcome back!

PR #5766, opened 2021-04-16
--remote-* option names deprecation

### LINKED ISSUE #5611: check / deprecate --remote-xxx options
naming is misleading:

- --remote-buffer does **not** set a **remote** buffer, but a send buffer used to send to remote repos
- --remote-ratelimit does **not** set a **remote** ratelimit, but a sending ratelimit used for sending to remote repos

otoh:

- --remote-path really sets a path to be used on the remote side

todo:

- add alias for --remote-ratelimit, deprecate --remote-ratelimit
- just change --remote-buffer, no deprecation  needed, this is brand new
- keep --remote-path as is

ThomasWaldmann:
Fixes #5611.


ThomasWaldmann:
see also #5767.

PR #5809, opened 2021-05-14
add progress indicator for archive check

elho:
Depending on the number of archives in a repository, the archive check part
of the check operation can take some time, so it should have a progress
indicator as well.

I do recall at least one person reporting on IRC that his `borg check --progress` "got stuck doing nothing after it completed to 100%" - completed the segment check, that was. :wink: 

Result looks like this:
```
$ borg check --progress
Checking archives 9.6%                                                                                                                            
```

For the `msgid` parameter I followed to what seems to be a `classname.callingfunctionname` convention in most places, but if it is rather supposed to be something more functionality based, I'll happily change it to something like e.g. `archive.check`.

I built and tested the change as-is on borg 1.1.16 for my own use, so happy to do a PR for 1.1-maint. :smiley:

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1877,8 +1877,11 @@ class ArchiveChecker:
                 return
         num_archives = len(archive_infos)
 
+        pi = ProgressIndicatorPercent(total=num_archives, msg='Checking archives %3.1f%%', step=0.1,
+                                      msgid='check.rebuild_refcounts')
         with cache_if_remote(self.repository) as repository:
             for i, info in enumerate(archive_infos):
+                pi.show(i)
                 logger.info('Analyzing archive {} ({}/{})'.format(info.name, i + 1, num_archives))
                 archive_id = info.id
                 if archive_id not in self.chunks:
@@ -1908,6 +1911,7 @@ class ArchiveChecker:
                 cdata = self.key.encrypt(data)
                 add_reference(new_archive_id, len(data), len(cdata), cdata)
                 self.manifest.archives[info.name] = (new_archive_id, info.ts)
+            pi.finish()
 
     def orphan_chunks_check(self):
         if self.check_all:


src/borg/archive.py
```diff
@@ -1877,8 +1877,11 @@ def valid_item(obj):
                 return
         num_archives = len(archive_infos)
 
+        pi = ProgressIndicatorPercent(total=num_archives, msg='Checking archives %3.1f%%', step=0.1,
+                                      msgid='archive.robust_iterator')
```
ThomasWaldmann, line 1881: 0.1 looks suspicious as num_archives is an integer.

src/borg/archive.py
```diff
@@ -1877,8 +1877,11 @@ def valid_item(obj):
...
         with cache_if_remote(self.repository) as repository:
             for i, info in enumerate(archive_infos):
+                pi.show(i)
```
ThomasWaldmann, line 1884: usually showing the progress indicator depends on `--progress` being given, see other places where it is used.

ThomasWaldmann (commented):
thanks for the PR!

... also please fix the typo in the commit comment and PR title.

src/borg/archive.py
elho: 0.1 is what is used in other places with integer "input" counts and while the documentation of the `step` parameter in `ProgressIndicatorPercent` is rather vague, it does state it to be in prcent and the code does treat it as steps of the percentage output:
https://github.com/borgbackup/borg/blob/7a872bbcddc1e4eccf9cfabaefa2a76d57c68582/src/borg/helpers/progress.py#L116
https://github.com/borgbackup/borg/blob/7a872bbcddc1e4eccf9cfabaefa2a76d57c68582/src/borg/helpers/progress.py#L134

So in this case, output is displayed, whenever the progress percentage increased by 0.1% or more since the last invocation of `show()`. This does not hurt with few archives either, because updates happen no more than one archive processed (and `show()` invoked) anyway.

src/borg/archive.py
elho: While I indeed have not fully analyzed the workings of this, I did observe that the equivalent code for the "Checknig segments" progress indicator does not have any such conditional either, the code does work as desired (showing progress output when `--progress` is given and not showing any output if not) and I do see that the `ProgressIndicatorBase` class uses the `borg.output.progress` logger, so I would assume that logger is only configured to output things if `--progress` was given. :slightly_smiling_face:

elho:
> ... also please fix the typo in the commit comment and PR title.

Oops, fixed.

src/borg/archive.py
ThomasWaldmann: Right, makes sense.

src/borg/archive.py
ThomasWaldmann: Right. Must have had some code from the past in my head.

Strange though, that some places in the code still check `if ... progress:` before calling `.show()`.

src/borg/archive.py
ThomasWaldmann, line 1881: Maybe easier to make sense of:
```
msgid='check.rebuild_refcounts'
```

ThomasWaldmann: Other than that, guess we are fine.

ThomasWaldmann: A PR for 1.1-maint would be welcome, thanks!

ThomasWaldmann:
Have to admit that there is not really a msgid "convention" or it is maybe not really strictly followed in the code base.

It's often rather something close to class / method names, but also something that is easy to make sense from.

src/borg/archive.py
elho: > A PR for 1.1-maint would be welcome, thanks!

Done: #5810

ThomasWaldmann:
thanks!

PR #5831, opened 2021-06-11
Allow --files-cache=size

### LINKED ISSUE #5686: Allow --files-cache=size

Yes

Feature suggestion

There are use cases when using size only for file change detection make sense.

rsync supports this feature (--size-only), and it would be convenient to have it in borg too (--files-cache=size)

Swanand01:
I have changed the VALID_MODES in the FilesCacheMode function in parseformat.py and added 's' for size only mode.

Fixes #5686.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -130,7 +130,7 @@ def ChunkerParams(s):
 
 def FilesCacheMode(s):
     ENTRIES_MAP = dict(ctime='c', mtime='m', size='s', inode='i', rechunk='r', disabled='d')
-    VALID_MODES = ('cis', 'ims', 'cs', 'ms', 'cr', 'mr', 'd')  # letters in alpha order
+    VALID_MODES = ('cis', 'ims', 'cs', 'ms', 'cr', 'mr', 'd', 's')  # letters in alpha order
     entries = set(s.strip().split(','))
     if not entries <= set(ENTRIES_MAP):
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
OK, change looks good!

How did you test it?

ThomasWaldmann:
@Swanand01 did you see my question / can you do some manual tests?

Swanand01:
> @Swanand01 did you see my question / can you do some manual tests?

Very sorry I missed your previous message. I don't actually know how to perform tests, could you please suggest what is to be done?

ThomasWaldmann:
- do a backup with default options
- touch a input file, see how it is detected as `M` due to its mtime change in the next backup
- touch input file again, see how it is **not** detected as `M` (but `U` like unchanged) if you run next backup with `--files-cache=size`

Swanand01:
`Exception: Could not find OpenSSL lib/headers, please set BORG_OPENSSL_PREFIX`

I am getting this error while installing from setup.py. I am on a windows based machine

ThomasWaldmann:
@Swanand01 borg does not work on native windows, you can try cygwin or WSL or Linux, BSD, macOS.

Usually there is some package containing the openssl header files, called libssl-dev or so.

ThomasWaldmann:
I merged this now without waiting for the practical test.

--files-cache=s is not used by default and also not advisable for most use cases.

Somebody brave enough to think that s is enough will test this. :-)

Issue #5839, opened 2021-06-15
do back/forward ports (55)


ThomasWaldmann:
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Abackport%2F1.1-maint
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Aforwardport%2Fmaster

do 2 simple/small or 1 mid-sized/mid-complex forward/backport from there.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/99262958-do-back-forward-ports-55)

braewoods:
What exactly are you asking for here? Taking some of these changes and adapting them to an older release?

ThomasWaldmann:
backport: take a change from master to 1.1-maint branch
forward port: take a change from 1.1-maint to master
(just klick on the links in top post to get a list of PRs that are marked for porting)

sometimes it is as simple as "git cherry-pick" + making sure the tests work.

sometimes there are conflicts that need to get resolved, sometimes stuff has to be done differently.

braewoods:
Ok. I'll see what I can do.

braewoods:
Ok. I await review to see if there's more required.

braewoods:
This one appears to be all done.

Issue #5822, opened 2021-06-05
borg check --repair --verify-data: inconsistent/sporadic chunk id verification failure


jurgenhaas:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE

#### Your borg version (borg -V).

1.1.16

#### Operating system (distribution) and version.

Ubuntu 20.04

#### Hardware / network configuration, and filesystems used.

Intel NUC with ect4

#### How much data is handled by borg?

400GB on that host, 274GB in the problematic repository

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
borg check --repair --verify-data --verbose /var/backups/borg/nextcloud_privat
```

## Describe the problem you're observing.

My monthly check reported a problem with the integrity of some chunks, so I started to analyse the repository. The command above logs the following statement:

```
Starting repository check
Starting repository index check
Completed repository check, no problems found.
Starting archive consistency check...
Enter passphrase for key /var/backups/borg/nextcloud_privat:
Starting cryptographic data integrity verification...
chunk f856bf715978bb1a4acc9d2f4b23bfc319a2452058541ae5c8ee086b75279ff9, integrity error: Data integrity error: Chunk f856bf715978bb1a4acc9d2f4b23bfc319a2452058541ae5c8ee086b75279ff9: id verification failed
Found defect chunks. They will be deleted now, so affected files can get repaired now and maybe healed later.
chunk %s not deleted, did not consistently fail.
Finished cryptographic data integrity verification, verified 298463 chunks with 1 integrity errors.
Analyzing archive ps1-2020-05-31T04:07:42.711754 (1/28)
Analyzing archive ps1-2020-06-30T04:12:33.529670 (2/28)
Analyzing archive ps1-2020-07-31T04:13:08.699011 (3/28)


Analyzing archive ps1-2021-06-03T04:33:13.815852 (27/28)
Analyzing archive ps1-2021-06-04T05:15:33.195727 (28/28)
Writing Manifest.
Committing repo (may take a while, due to compact_segments)...
Finished committing repo.
Archive consistency check complete, problems found.
```

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes, no matter how often I run this. It continues reporting that integrity error but still doesn't delete the reported chunk.

ThomasWaldmann:
```
[...]

PR #5843, opened 2021-06-16
fix missing parameter in "did not consistently fail" msg, see #5822

braewoods:
This is a forward-port attempt of #5823. It is intended to be part of issue #5839.


ThomasWaldmann (approved):
LGTM

Issue #5838, opened 2021-06-15
do back/forward ports (54)


ThomasWaldmann:
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Abackport%2F1.1-maint
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Aforwardport%2Fmaster

do 2 simple/small or 1 mid-sized/mid-complex forward/backport from there.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/99262889-do-back-forward-ports-54)

braewoods:
Let me know if you need something more for this to be closed.

ThomasWaldmann:
The first one was invalid (sorry, forgot to remove porting label).

Also, maybe write a comment like https://github.com/borgbackup/borg/issues/5772#issuecomment-822740629 here also - when claiming the bounty, bountysource will ask you for a url to your solution and the comment url can be used for this.

braewoods:
Ok. I'll see what I can substitute instead.

braewoods:
@ThomasWaldmann Ok. I prepared something else to take its place. Hopefully this one will suffice for this issue.

braewoods:
This should be all done now. 2 small ports completed.

PR #5844, opened 2021-06-16
docs: add unsafe workaround to use an old repo copy, fixes #5722

### LINKED ISSUE #5722: using a backup of a respository results in improper/unclear result and errors

Yes

Issue/Suggestion

n/a

1.1.13

n/a
n/a
n/a
using a backup (older) repository  "borg create .....", etc.

I have a borg repository that I experimented with and didn't like my changes.  So I got a backup copy of the repository from a few weeks prior and resumed using it normally.  Upon attempted use (conducting a regular file backup "borg create") a message/error came back:

"Cache, or information obtained from the security directory is newer than repository - this is either an attack or unsafe (multiple repos with same ID)"

I searched docs, issues, net, etc and tried to figure out the issue and/or fix. (I knew of course it was because I had restored a repository, but I couldn't figure out how to fix it or tell borg things were OK and to proceed.  I deleted cache.  Did "borg check", etc. Ultimately (after frustration) I figured out that i needed to manually dig into a hidden directory and delete some files associated with the (newer/deleted/scrapped) repository before borg would then run correctly (with my restored backup). (I think it was deleting cache/security, but dont' remember exactly at this point.)

COMMENTS AND SUGGESTIONS:
1) Restoring a backup of a repository (after corrupting or losing or testing the orginal) is not an unusual thing to do.  Throwing this error that doesn't mention that situation is not user friendly, generates fear, and is unhelpful. 

2) I suggest the message be changed, such as to: "Cache, or information obtained from the security directory is newer than repository - this is either an attack or unsafe (multiple repos with same ID) UNLESS YOU ARE USING AN OLD/BACKUP VERSION OF A REPOSITORY THAT HAS BEEN RESTORED.  IF THAT IS THE CASE THEN YOU NEED TO REMOVE/RESET SECURITY PARAMETERS USING "borg delete --securitycache-only"

3) Since use of backup files in computer world is not abnormal, not only should the error message mention it, but there should also be a command that resets or fixes that security information.  There is already a flag to delete a repository cache.  I therefore suggest addding a new flag to the "borg delete" command which will purge/reset/remove/fix the necessary security information (I dont know the specific files...whatever needs to be done so a person can use an older backup.)

4) documentation should mention this flag in the "delete" section, of course, but I also suggest a general section in the "additional notes" or other appropriate doc area that covers the use of an older backup.  I was unable to find an such discussion.  (ALTHOUGH if the command line error message is rewritten as in ##2 above then the message itself will then point the user to "delete --securitycache-only" whereby the documention there will say something like "If you are using an old version of a respository and know that you do not have a hack or security situation, then you can reset the security parameters with this flag", etc, etc.

I believe these steps would make a repository restore situation clear as to what is happening and how to address Borg suspecting a hack.

n/a

n/a

braewoods:
This is a forward-port attempt of #5732. It is intended to be part of issue #5838.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -118,6 +118,31 @@ you could delete the manifest-timestamp and the local cache:
 
 This is an unsafe and unsupported way to use borg, you have been warned.
 
+"this is either an attack or unsafe" warning
+--------------------------------------------
+
+About the warning:
+
+  Cache, or information obtained from the security directory is newer than
+  repository - this is either an attack or unsafe (multiple repos with same ID)
+
+"unsafe": If not following the advice from the previous section, you can easily
+run into this by yourself by restoring an older copy of your repository.
+
+"attack": maybe an attacker has replaced your repo by an older copy, trying to
+trick you into AES counter reuse, trying to break your repo encryption.
+
+If you'ld decide to ignore this and accept unsafe operation for this repository,
+you could delete the manifest-timestamp and the local cache:
+
+::
+
+  borg config repo id   # shows the REPO_ID
+  rm ~/.config/borg/REPO_ID/manifest-timestamp
+  borg delete --cache-only REPO
+
+This is an unsafe and unsupported way to use borg, you have been warned.
+
 Which file types, attributes, etc. are *not* preserved?
 -------------------------------------------------------
 


ThomasWaldmann:
If you look at the context of your change, you see that this already was ported to master.
So your change is not needed and would rather duplicate information that is already there.

So likely I just forgot to remove the porting label from the older PR.

PR #5846, opened 2021-06-16
Complete documentation on append-only remote repos (#5519)

braewoods:
This is a back-port attempt of #5519. It is intended to be part of issue #5838.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -210,6 +210,14 @@ This is not a problem anymore.
 
 For more details, see :ref:`checkpoints_parts`.
 
+How can I switch append-only mode on and off?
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
**Specificities: Append-only repositories** <- that section is missing in your changeset - did you check if it is already there?

braewoods:
It appears to be so. See the whole file here: https://github.com/braewoods/borg/blob/typos_dec/docs/deployment/hosting-repositories.rst

braewoods:
Apparently the version already in the 1.1-maint branch already has that "Specificities: Append-only repositories" section header.

ThomasWaldmann:
OK, thanks for checking!

Issue #5837, opened 2021-06-15
do back/forward ports (53)


ThomasWaldmann:
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Abackport%2F1.1-maint
https://github.com/borgbackup/borg/issues?utf8=%E2%9C%93&q=label%3Aforwardport%2Fmaster

do 2 simple/small or 1 mid-sized/mid-complex forward/backport from there.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/99262890-do-back-forward-ports-53)

braewoods:
I'll work on some more here.

braewoods:
After this I'll see about the typo backport. Your notes suggested these two should be backported first though which is why I've waited.

PR #5850, opened 2021-06-16
Complements the documentation for pattern files and exclude files (#5520)

braewoods:
This is a back-port attempt of #5520. It is intended to be part of issue #5837.


ThomasWaldmann:
hah! :)

Issue #5129, opened 2020-04-18
update docs bounty (37)


ThomasWaldmann:
there are quite some docs issues: https://github.com/borgbackup/borg/labels/documentation

grab some doc update issues, create a PR against master (and, after discussion, also to 1.1-maint, if applicable, see here, also linked from PR submission template)

the bounty is for fixing 2 small issues or 1 bigger issue.
docs issues that already have a bounty on them do not qualify.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/91114924-update-docs-bounty-37)

braewoods:
Is this still available?

ThomasWaldmann:
@braewoods sure!

braewoods:
@ThomasWaldmann I noticed the source code includes a copy of the manpage documentation. How is that intended to be updated?

ThomasWaldmann:
man pages and parts of the docs are generated from archiver.py.

braewoods:
Ok, since those 2 are merged now and back-ported, the work for this issue should be all done. Can you close it please?

ThomasWaldmann:
thanks!

ThomasWaldmann:
@braewoods you can claim the bounty now.

braewoods:
@ThomasWaldmann Thanks. I've filed a claim. Now if I could just get a cash out...

ThomasWaldmann:
You've just to request that from BS. IIRC they process cash outs once a week.

braewoods:
That's what the official sources say but the reality is they've stopped processing them that I can tell. Their website repository is littered with people complaining about it and I've been waiting over 2 weeks for my cash out request.

braewoods:
See here: https://github.com/bountysource/core/issues/1539

ThomasWaldmann:
Thanks for informing me about this. Let's hope they get their stuff fixed soon.

PR #5855, opened 2021-06-17
add notice about defective hardware to check documentation (#5753)

### LINKED ISSUE #5753: Add a notice to man borg-check about defective hardware?

Yes.

Neither, it's a suggestion to improve the documentation.

borg 1.1.9 (from Debian buster, borgbackup=1.1.9-2+deb10u1 (both client and server)

Debian 10 (buster) (both client and server)

irrelevant

irrelevant

```
borg check --progress --verbose /path/to/repo
```

I encountered my first case of a failing `borg check` yesterday that wasn't actually caused by a damaged repository, but by defective memory. This is actually documented in the [FAQ](https://borgbackup.readthedocs.io/en/stable/faq.html#i-get-an-integrityerror-or-similar-what-now):

> Checking memory
>    Intermittent issues, such as borg check finding errors inconsistently between runs, are frequently caused by bad memory.
>    Run memtest86+ (or an equivalent memory tester) to verify that the memory subsystem is operating correctly.

But due to the fact that I had several damaged repositories in the past (due to power failures) that I was always able to to repair with `borg check --repair`, I didn't actually bother to look through the FAQ, and directly proceeded with `borg check --repair`. The timeline looked something like this:

- A manual attempt to run the backup failed with `File failed integrity check: /path/to/repo/cache/9ce198549ec83582155f288b853891e6cb1d33f41547d52748d4e2c9fb5ada1d/chunks`
- `borg check` found errors (`Segment entry checksum mismatch`), so I figured that the repository was damaged.
- `borg check --repair` suddenly no longer found any errors.
- Another `borg check` again found errors.
- Another `borg check --repair` again didn't find any errors.
- At this point I was very confused and wondered if I was doing something wrong or if I had encountered a very strange bug, because it looked to me like I had some kind of repository corruption that is visible to `borg check`, but not to `borg check --repair`. I asked for help on IRC, and was advised to use the most recent borg version as the one provided by Debian is rather old.
- I downloaded a static build of the most recent borg release (1.1.16), and ran a `check` with this version. Now this also didn't find any errors during `borg check`.
- I ran borg 1.1.9 check again to see if the old version was still able to see the corruption, and again it would report `Segment entry checksum mismatch`.

Only at this point I actually compared the output of my multiple `borg check` runs and noticed that the segment numbers were always different:

```
Starting repository check
Data integrity error: Segment entry checksum mismatch [segment 2061, offset 448940545]
```

```
Starting repository check                                                                          
Data integrity error: Segment entry checksum mismatch [segment 1479, offset 393404125]             
```

```
Starting repository check
Data integrity error: Segment entry checksum mismatch [segment 1007, offset 37742236]                                                                                                                                                       
Data integrity error: Segment entry checksum mismatch [segment 2129, offset 116527082]                                                                                                                                                      
```

At this point it was rapidly becoming clear to me that the hardware might be the actual cause, and a `memtester` run confirmed it:

```
server ~ # memtester 6G
memtester version 4.3.0 (64-bit)
Copyright (C) 2001-2012 Charles Cazabon.
Licensed under the GNU General Public License version 2 (only).

pagesize is 4096
pagesizemask is 0xfffffffffffff000
want 6144MB (6442450944 bytes)
got  6144MB (6442450944 bytes), trying mlock ...locked.
Loop 1:
  Stuck Address       : ok
  Random Value        : ok
  Compare XOR         : ok
  Compare SUB         : ok
  Compare MUL         : ok
  Compare DIV         : ok
  Compare OR          : ok
  Compare AND         : ok
  Sequential Increment: ok
  Solid Bits          : ok
  Block Sequential    : testing   4FAILURE: 0x404040404040404 != 0x404040404040400 at offset 0x4a7f0fb0.
  Checkerboard        : ok
  Bit Spread          : testing   2FAILURE: 0x00000014 != 0x00000010 at offset 0x4a7f0fb0.
  Bit Flip            : testing  17FAILURE: 0x00000004 != 0x00000000 at offset 0x4a7f0fb0.
  Walking Ones        : ok
  Walking Zeroes      : testing 125FAILURE: 0x00000004 != 0x00000000 at offset 0x4a7f0fb0.
  8-bit Writes        : ok
  16-bit Writes       : |FAILURE: 0x77cea9b8efef8c2c != 0x77cea9b8efef8c28 at offset 0x4a7f0fb0.
```

I think mentioning this potential problem in `man borg-check` would be worthwhile, to increase the likelihood that people who use `borg check` are aware of it. E.g. with the following passage:

> Note that `borg check` can report spurious errors when running on defective hardware. If you're seeing errors during `borg check` but not in a subsequent `borg check --repair`, run multiple checks and compare the defective segment IDs. If the defective segment IDs vary between checks, check your hardware e.g. with memtest86+ or memtester.

irrelevant

irrelevant

braewoods:
This should close #5753 and count towards completion of #5129.

### DIFF
--- a/docs/usage/check.rst.inc
+++ b/docs/usage/check.rst.inc
@@ -90,6 +90,11 @@ The check command verifies the consistency of a repository and the corresponding
 check --repair is a potentially dangerous function and might lead to data loss
 (for kinds of corruption it is not capable of dealing with). BE VERY CAREFUL!
 
+Pursuant to the previous warning it is also highly recommended to test the
+reliability of the hardware running this software with stress testing software
+such as memory testers. Unreliable hardware can also lead to data loss especially
+when this command is run in repair mode.
+
 First, the underlying repository data files are checked:
 
 - For all segments, the segment magic header is checked.


docs/usage/check.rst.inc
```diff
@@ -90,6 +90,11 @@ The check command verifies the consistency of a repository and the corresponding
 check --repair is a potentially dangerous function and might lead to data loss
 (for kinds of corruption it is not capable of dealing with). BE VERY CAREFUL!
 
+Also note that the check command can report spurious errors when the underlying
+hardware is defective. If you are getting inconsistent error reports then we
+advise you to run hardware testing software such as memory testers to determine
+the exact cause of the problem.
+
```
ThomasWaldmann, line 97: Yes, think we should add something like this, just not sure about the phrasing.

can report -> may report?

inconsistent errors -> check hardware - not sure about the "inconsistent" here. e.g. if one had some bit flipping in RAM and that data was saved to disk, it might as well become permanent and consistently report the same issue again and again.

braewoods: Yea, that's the problem with hardware problems. It's impossible to truly predict how it will influence program behavior.

braewoods:
Ok. I revamped the text using a different approach that basically accomplishes the same thing. A warning about the reliability of check when the underlying hardware is unreliable.

ThomasWaldmann:
thanks!

PR #5861, opened 2021-06-18
add paragraph added in #5855 to utility documentation source

braewoods:
I noticed the changes I did were to the generated files and not the original source. This should correct that.


ThomasWaldmann:
oops, missed that in review also. :-)

glad you noticed it.

Issue #5857, opened 2021-06-18
leading slash required for include/exclude matching of full paths


Alexander-Shukaev:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE

#### Your borg version (borg -V).

```
borg 1.1.16
```

#### Operating system (distribution) and version.

Arch Linux

#### Hardware / network configuration, and filesystems used.

EXT4

## Describe the problem you're observing.

I tried

```
sh:home/*/Projects/**/*.log
```

and it appears that this pattern does not match.  I suspect that it's because double asterisk (star) is only supported in the beginning of the pattern.  Is that true?  Is there any technical limitation that prevents them from being present anywhere?

Alexander-Shukaev:
I figured this out.  It's an issue with documentation, maybe amend it whenever you have time.  Run

```bash
borg help patterns
```

and review this part:

```bash
Examples:

    # Exclude '/home/user/file.o' but not '/home/user/file.odt':
    $ borg create -e '*.o' backup /

    # Exclude '/home/user/junk' and '/home/user/subdir/junk' but
    # not '/home/user/importantjunk' or '/etc/junk':
    $ borg create -e '/home/*/junk' backup /

    # Exclude the contents of '/home/user/cache' but not the directory itself:
    $ borg create -e /home/user/cache/ backup /

    # The file '/home/user/cache/important' is *not* backed up:
    $ borg create -e /home/user/cache/ backup / /home/user/cache/important

    # The contents of directories in '/home' are not backed up when their name
    # ends in '.tmp'
    $ borg create --exclude 're:^/home/[^/]+\.tmp/' backup /

    # Load exclusions from file
    $ cat >exclude.txt <<EOF
    # Comment line
    /home/*/junk
    *.tmp
    fm:aa:something/*
    re:^home/[^/]\.tmp/
    sh:home/*/.thumbnails
     # Example with spaces, no need to escape as it is processed by borg
    some file with spaces.txt
    EOF
    $ borg create --exclude-from exclude.txt backup /
```

namely it should be (missing characters are highlighted with `^`):

```bash
    re:^/home/[^/]+\.tmp/
        ^         ^
    sh:/home/*/.thumbnails
       ^
```

ThomasWaldmann:
Hmm, yeah guess that's wrong.

All the paths inside an archive have the leading `/` removed, but guess the matcher still sees it in the fs path.

PR #5869, opened 2021-06-22
docs: add missing leading slashes in help patterns

remyabel:
Fixes: https://github.com/borgbackup/borg/issues/5857

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2348,8 +2348,8 @@ class Archiver:
             /home/*/junk
             *.tmp
             fm:aa:something/*
-            re:^home/[^/]\\.tmp/
-            sh:home/*/.thumbnails
+            re:^/home/[^/]+\\.tmp/
+            sh:/home/*/.thumbnails
             # Example with spaces, no need to escape as it is processed by borg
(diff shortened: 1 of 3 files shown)


src/borg/archiver.py
```diff
@@ -2348,8 +2348,8 @@ def do_break_lock(self, args, repository):
             /home/*/junk
             *.tmp
             fm:aa:something/*
-            re:^home/[^/]\\.tmp/
-            sh:home/*/.thumbnails
+            re:^/home/[^/]+\\.tmp/
+            sh:/home/*/.thumbnails
```
ThomasWaldmann, line 2352: btw, for this PR it is sufficient to change these 2 lines.

did you do the other changes manually or did you run python setup.py build_man/build_usage?
it is not needed to do that, i run these at release time.

remyabel: I did it manually. Would you like me to revert the changes?

ThomasWaldmann: yeah, reverting the 2 unneeded changes would be good. thanks!

ThomasWaldmann:
thanks!

PR #5880, opened 2021-07-02
docs: mention tar --compare

adept:
Hi!

I think that `tar --compare` is not very widely known, and it would be nice to mention this trick in the FAQ

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -230,13 +230,14 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
-How can I compare contents of the archive to my local filesystem?
+How can I compare contents of an archive to my local filesystem?
 -----------------------------------------------------------------
 
-You can instruct ``export-tar`` to send tar archive to the stdout, and
+You can instruct ``export-tar`` to send a tar stream to the stdout, and
 then use ``tar`` to perform the comparison:
 
 ::
+
     borg export-tar /path/to/repo::archive-name - | tar --compare -f - -C /path/to/compare/to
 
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -230,6 +230,16 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
+How can I compare contents of the archive to my local filesystem?
+-----------------------------------------------------------------
+
+You can instruct ``export-tar`` to send tar archive to the stdout, and
+then use ``tar`` to perform the comparison:
+
+::
```
ThomasWaldmann, line 239: can you add an empty line below the `::` - iirc it will not format correctly otherwise, see line 114.

docs/faq.rst
```diff
@@ -230,6 +230,16 @@ all the part files and manually concatenate them together.
 
 For more details, see :ref:`checkpoints_parts`.
 
+How can I compare contents of the archive to my local filesystem?
```
ThomasWaldmann, line 233: Not sure, but does "... of **an** archive ..." sound better?

docs/faq.rst
```diff
@@ -230,6 +230,16 @@ all the part files and manually concatenate them together.
...
+-----------------------------------------------------------------
+
+You can instruct ``export-tar`` to send tar archive to the stdout, and
```
ThomasWaldmann, line 236: "... to send **a** tar archive ..." or "... to send **a** tar **stream** ..."

ThomasWaldmann (changes requested):
good idea, thanks for the PR!

adept:
Thank you! Fixed spelling and formatting

ThomasWaldmann (approved):
Thanks!

PR #5901, opened 2021-07-10
Adding try/finally block to make sure cleanup happens

### LINKED ISSUE #5630: [doc] typo / inaccuracy on 'borg benchmark' usage page
[EDIT - originally I misunderstood the docs. They are correct but could benefit from maybe adding an example command ? ]

When using wrong arguments or with permission issues, `borg benchmark crud` can exit without deleting the `borg-test-data/` it created, leading to some head scratching. It would be a nice addition to cleanup that folder systematically on exit (clean or error), unless there's some use case for the current bheavior ?

brunobehnken:
In issue #5630 it was specified that when `borg benchmark` is called with wrong arguments or with permission issues, the folder `borg-test-data` is not deleted.
During tests I have also discovered that the same happens when borg receives a `keyboard interrupt` while executing the benchmark.

To solve this problem, a `try/finally` block was added in method `test_files`, to make sure the command that removes the folder will be executed (in the `finally` statement) even if the code in the `try` statement fails.

In my tests, this have solved the issue on the specified test cases and in the case of the keyboard interrupt.

This PR resolves #5630

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -456,16 +456,18 @@ class Archiver:
 
         @contextmanager
         def test_files(path, count, size, random):
-            path = os.path.join(path, 'borg-test-data')
-            os.makedirs(path)
-            z_buff = None if random else memoryview(zeros)[:size] if size <= len(zeros) else b'\0' * size
-            for i in range(count):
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
LGTM, thanks for the PR!

PR #5911, opened 2021-07-11
Forwardport pull request #5902

KN4CK3R:


ThomasWaldmann:
Could you do the port in a way so that the original commit comments show in history?

Otherwise it is hard to generate a meaningful changelog, if there are only references to PR numbers.

ThomasWaldmann:
@KN4CK3R did you see my feedback?

KN4CK3R:
Sorry, missed your feedback.
The original commit message is still there or what do you mean?
grafik

ThomasWaldmann:
Ah, it's in the long comment, which I did not see, just looked at the short (first line) commit comment.

Please don't backport the merge commits in future, but the commit that implements the change, so the comment is better visible.

PR #5923, opened 2021-07-12
Update shell completions to 1.1.17

oxiedi:


ThomasWaldmann:
too late for 1.1.17, but we'll have a good preparation for .18 now.

ThomasWaldmann:
needs review also (not by me).

oxiedi:
Completion for Zsh could be maintained at https://github.com/zsh-users/zsh-completions (a third-party repository with completion functions) or Zsh's repository.

Pros:
- can be updated after a release;
- no "required review without a reviewer" dilemma.

Cons:
- user has to explicitly install zsh-completions or
- Zsh's release cycle slower than Borg's which would result in outdated completion for Borg.

What do you think?

ThomasWaldmann:
Hmm, I don't have strong opinions on this one. Maybe some borg package maintainers can comment?

In any case, changing the way we deal with it could cause some work for them.

ThomasWaldmann:
Can somebody please review this?

ThomasWaldmann:
Thanks for the PR / for review, merged!

Guess we need a forward port to master now.

Issue #5926, opened 2021-07-14
duplicate faq entry 1.1-maint


ThomasWaldmann:
```
How can I switch append-only mode on and off?
You could do that (via borg config REPO append_only 0/1), but using different ssh keys and different entries in authorized_keys is much easier and also maybe has less potential of things going wrong somehow.

My machine goes to sleep causing Broken pipe
When backing up your data over the network, your machine should not go to sleep. On macOS you can use caffeinate to avoid that.

How can I switch append-only mode on and off?
You could do that (via borg config REPO append_only 0/1), but using different ssh keys and different entries in authorized_keys is much easier and also maybe has less potential of things going wrong somehow.
```

ThomasWaldmann:
Also duplicate: My machine goes to sleep causing Broken pipe

PR #5931, opened 2021-07-18
docs: remove duplicate faq entries

remyabel:
Fixes: #5926


ThomasWaldmann:
thanks! can you also please check if it is duplicate in master, too?

remyabel:
> thanks! can you also please check if it is duplicate in master, too?

Doesn't look like it.

PR #5937, opened 2021-07-20
fix xxh64 related build

ThomasWaldmann:
@LocutusOfBorg this is inspired by your debian patch.

https://salsa.debian.org/debian/borgbackup/-/blob/master/debian/patches/bundled-xxhash.patch

changes:
- the ".../xxh64" part of the path was missing in your patch.
- xxh64 belongs to checksums module (not crypto)
- getting it to compile (xxh64, zstd with bundled code or with system lib) was a bit of a pain, but I hope it works now

also, i gave xxh64 0.8.1 a try and it worked, so i updated the bundled code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -222,6 +222,7 @@ libxxhash_prefix = setup_xxhash.xxhash_system_prefix(possible_libxxhash_prefixes
 if prefer_system_libxxhash and libxxhash_prefix:
     print('Detected and preferring libxxhash over bundled XXHASH')
     define_macros.append(('BORG_USE_LIBXXHASH', 'YES'))
+    define_macros.append(('XXH_PRIVATE_API', 'YES'))  # do not use this for bundled zstd or xxh64, breaks the build
     libxxhash_system = True
 else:
     libxxhash_system = False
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
Hmm, looks good, but does not compile. :-( @LocutusOfBorg any idea?

I also tried adding that XXH_PRIVATE_API, but that doesn't help either.

I am also wondering: what was the reason for your patch, I did not see a bug report, neither did I have a build issue myself with the 1.1.17 code.

ThomasWaldmann:
https://salsa.debian.org/debian/borgbackup/-/blob/master/debian/patches/bundled-xxhash.patch

ThomasWaldmann:
@LocutusOfBorg please review!

ThomasWaldmann:
Forward-port notes:
- in master we have xxhash 0.8.0 still
- some stuff done here might also apply to master, but not everything, the setup is rather different.

ThomasWaldmann:
This needs some testing for the "system lib" as well as for the "use bundled code" case, see upper part of setup.py.

LocutusOfBorg:
Test cases performed:
- 1.1.17 + this patch, no xxhash installed on the system (build OK)
- 1.1.17 + this patch, xxhash 0.8.1 installed on the system (build OK)
thanks a lot for fixing this issue!

ThomasWaldmann:
@LocutusOfBorg thanks for testing, will merge after test CI has completed.

Issue #4634, opened 2019-06-22
borg diff should have --format option


elho:
The `diff` sub-command should have a `--format` option like e.g. the `list` sub-command has.

Placeholders should include (but not be limited to), `path`, `type` (directory entry type, same as with `list`), `change` (giving the type of change, ie. whether added, deleted, modified (content change), owner, group, mode, etc. - speaking long or like AME, whatever is deemed preferable) and whatever fields to represent what else diff has at its hands that it can output, e.g. `oldsize`, `newsize`, `deltasize`, etc.
If several aspects of a directory entry were changed, multiple lines, each with the according `change` value would be printed. This does not sound nice at first, bit fits the format-string style output (as opposed to more template style with conditionals etc.) and suits working on the output with shell tools like `grep`.

This would allow to do stuff like `borg diff --format "{change}{type}:{path}{NUL}" | sed --null-data --quiet --expression='s/(A|M)-://p' | xargs --null dosomethingwithchangedfiles`, thus being able to workaround/solve e.g. #3514 and  #4191.

palao:
I'd like to implement this. Is it ok?

elho:
> I'd like to implement this. Is it ok?

Great, go ahead. :+1:

Just note that it is not as trivial as it sounds, the current output is not generated by the diff function but deeper in the code where the actual changes between items are detected.

So based on what I remember from back then, when I did briefly look into the code to see whether I could just fix it quickly instead of opening the bug, I would imagine it to involve these steps:
1. Redo that internal part to return/communicate changes using some defined proper data structure (e.g. a nested dict of the attributes changed, each with old and new maybe?) instead of strings.
2. Make the existing diff function build the current output based on that data structure gathered from the internal diffing process.
3. Make any other part of the code that eventually uses those strings to use the new data structure.
4. Extent the  existing diff function generate `--format` output based on those very same data structures

Further additions, such as some sort of JSON output would then be quite trivial.

palao:
One question: what are the expected list of possible placeholders?

diivi:
I would like to work on this. Please assign this issue to me, I'll ask for help when needed on the #borgbackup IRC channel.

ThomasWaldmann:
@diivi you have it, thanks for helping!

diivi:
[...]

PR #5944, opened 2021-07-31
Initial functional tests for diff --format (Issue #4634)

palao:
Could you please tell me if this test is in the right direction and what else would you like to see there?

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3776,7 +3776,7 @@ class Archiver:
                                help='Sort the output lines by file path.')
         subparser.add_argument('--json-lines', action='store_true',
                                help='Format output as JSON Lines. ')
-        subparser.add_argument('--format', '--diff-format', metavar='FORMAT', dest='format',
+        subparser.add_argument('--format', metavar='FORMAT', dest='format',
                                help=(
                                    'specify format for file differences (default: '
                                    '"{newsize:8d} B {oldsize:8d} B {path}{NL}")'
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -28,6 +28,7 @@ try:
     from contextlib import contextmanager
     from datetime import datetime, timedelta
     from io import TextIOWrapper
+    from collections import defaultdict
 
     from .logger import create_logger, setup_logging
 
@@ -1101,11 +1102,48 @@ class Archiver:
     def do_diff(self, args, repository, manifest, key, archive):
         """Diff contents of two archives"""
 
-        def print_json_output(diff, path):
+        if args.format is not None:
+            format = args.format
+        else:
+            format = "{change:<19} {path}{NL}"
+        def print_json_output(diff, path, *rest):
             print(json.dumps({"path": path, "changes": [j for j, str in diff]}))
 
-        def print_text_output(diff, path):
-            print("{:<19} {}".format(' '.join([str for j, str in diff]), path))
+        def print_text_output(diff, path, baseitem):
+            diffs_dict = defaultdict(str)
+            diffs_dict.update(
+                {
+                    "NL": os.linesep,
+                    "change": ' '.join([str for j, str in diff]),
+                    "path": path,
+                    "size": 0,
+                    "oldsize": 0,
+                    "newsize": 0,
+                    "deltasize": 0,
+                }
+            )
+            for attr in ("size", "mtime", "ctime", "atime"):
+                try:
+                    value = getattr(baseitem, attr)
+                except AttributeError:
+                    pass
+                else:
+                    diffs_dict[attr] = value
+            for change, _ in diff:
+                if change == "type":
+                    continue
+                diffs_dict.update(change)
+            diffs_dict["changetype"] = diffs_dict["type"]
+            diffs_dict["newsize"] = diffs_dict["size"]
+            if diffs_dict["deltasize"] == 0:
+                if "added" in diffs_dict:
+                    diffs_dict["deltasize"] = diffs_dict["added"]
+                if "removed" in diffs_dict:
+                    diffs_dict["deltasize"] -= diffs_dict["removed"]
+            if diffs_dict["deltasize"] != 0:
+                if diffs_dict["oldsize"] == 0:
+                    diffs_dict["oldsize"] = diffs_dict["size"]-diffs_dict["deltasize"]
+            print(format.format(**diffs_dict))
 
         print_output = print_json_output if args.json_lines else print_text_output
 
@@ -1124,13 +1162,13 @@ class Archiver:
 
         diffs = Archive.compare_archives_iter(archive1, archive2, matcher, can_compare_chunk_ids=can_compare_chunk_ids)
         # Conversion to string and filtering for diff.equal to save memory if sorting
-        diffs = ((path, diff.changes()) for path, diff in diffs if not diff.equal)
+        diffs = ((path, diff.changes(), diff._item2) for path, diff in diffs if not diff.equal)
 
         if args.sort:
             diffs = sorted(diffs)
 
-        for path, diff in diffs:
-            print_output(diff, path)
+        for path, diff, item2 in diffs:
+            print_output(diff, path, item2)
 
         for pattern in matcher.get_unmatched_include_patterns():
             self.print_warning("Include pattern '%s' never matched.", pattern)
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3776,6 +3776,11 @@ class Archiver:
                                help='Sort the output lines by file path.')
         subparser.add_argument('--json-lines', action='store_true',
                                help='Format output as JSON Lines. ')
+        subparser.add_argument('--format', '--diff-format', metavar='FORMAT', dest='format',
+                               help=(
+                                   'specify format for file differences (default: '
+                                   '"{newsize:8d} B {oldsize:8d} B {path}{NL}")'
+                                ))
         subparser.add_argument('location', metavar='REPO::ARCHIVE1',
                                type=location_validator(archive=True),
                                help='repository location and ARCHIVE1 name')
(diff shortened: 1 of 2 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -4095,6 +4095,42 @@ def test_sort_option(self):
...
+        os.utime('input/a_file', (atime, mtime))
+
+        test_archive1 = self.repository_location + '::test1'
+        self.cmd('create', test_archive1, self.input_path)
+
+        output_file_mod = self.cmd('diff', test_archive0, "test1")
+        output2 = self.cmd(
+            'diff', '--format',
+            '{newsize:8d} B {oldsize:8d} B {path}{NEWLINE}',
+            test_archive0, "test1")
```
ThomasWaldmann, line 4117: as long as you can stay < 120 chars, you can keep it on one line.

src/borg/testsuite/archiver.py
```diff
@@ -4095,6 +4095,42 @@ def test_sort_option(self):
...
+        output3 = self.cmd(
+            'diff', '--format',
+            '{change} ({deltasize:8d} B) {mtime:%s} {path}{NL}',
+            test_archive0, "test1")
+        self.assert_equal(output_file_mod, output2)
+        self.assert_not_equal(output_file_mod, output3)
```
ThomasWaldmann, line 4123: the latter is a rather weak test :-)

src/borg/testsuite/archiver.py
```diff
@@ -4095,6 +4095,42 @@ def test_sort_option(self):
...
+        expected2 = [
+            "     +25 B       -8 B input/a_file\n"
+        ]
+        expected3 = [
+            "modified (     +25 B) 234900290 input/a_file\n"
+        ]
+        assert all(_ in line for _, line in zip(expected2, output2))
+        assert all(_ in line for _, line in zip(expected3, output3))
```
ThomasWaldmann, line 4131: content in line for ...

ThomasWaldmann (commented):
looks pretty good.

maybe i'ld keep the asserts closer to the commands, likely that is better to deal with when the amount of commands/asserts grows.

also names maybe like output_default, output_format1, excpected_default, expected_format1, ...

src/borg/testsuite/archiver.py
ThomasWaldmann: also add a comma after list elements, even if it is just 1. better for future diffs and copy&paste.

src/borg/testsuite/archiver.py
palao: I made those changes. Only the long lines I kept them splitted since they are > 120 chars plus it's more comfortable in my environment to have not too long lines, if that's not an issue for you

src/borg/testsuite/archiver.py
palao: I have removed it

src/borg/testsuite/archiver.py
palao: I changed the assert lines; they seemed wrong to me

palao:
Do I close this pull request and make a new one when ready, or do we use this one?

src/borg/testsuite/archiver.py
ThomasWaldmann: _ is usually rather used for some tuple elements that are NOT used.

so maybe "content" or "s" or ... rather?

src/borg/testsuite/archiver.py
```diff
@@ -4204,36 +4204,36 @@ def test_format_option(self):
...
-        output3 = self.cmd(
+        expected_default = [
+            "    +33 B      -8 B input/a_file"
+        ]
+        self.assert_equal(output_default, output_default_explicit_fmt)
+        self.assert_equal(expected_default, output_default_explicit_fmt.splitlines())
+
+        output_nontrivial_fmt1 = self.cmd(
             'diff', '--format',
             '{change} ({deltasize:8d} B) {mtime:%s} {path}{NL}',
```
ThomasWaldmann, line 4230: 8d - hmm, is this really an integer that gets formatted?

iirc it could be a string, see that "+" in there, for a normal int, that is not present.

ThomasWaldmann (commented):
parts of your last changeset look like unintended changes, please check.

ThomasWaldmann:
@palao did you see my feedback?

you can just continue within this pull request, just do some fixup commits.

also, considering age of this PR, maybe updating your local master branch and then doing a `git rebase -i master` within this PR branch would be a good idea.

palao:
@ThomasWaldmann thank you for the reminder. I'll do that.

ThomasWaldmann:
Considering this PR is a bit older already, maybe you could:
- update your local master branch
- switch to your PR branch
- `git rebase -i master`  # ... and carefully resolve conflicts, if any
- `git push -f`  # update PR

We recently had some fixes for testing, so CI will work better after that.

src/borg/testsuite/archiver.py
```diff
@@ -4199,6 +4199,42 @@ def test_sort_option(self):
 
         assert all(x in line for x, line in zip(expected, output.splitlines()))
 
+    def test_format_option(self):
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+
+        self.create_regular_file('a_file', size=8)
+        test_archive0 = self.repository_location + '::test0'
+        self.cmd('create', test_archive0, "input")#self.input_path)
```
ThomasWaldmann, line 4207: ^ looks strange

src/borg/testsuite/archiver.py
```diff
@@ -4199,6 +4199,42 @@ def test_sort_option(self):
...
+
+        os.unlink('input/a_file')
+        self.create_regular_file('a_file', size=33)
+        atime, mtime = 123457520, 234900290
+        os.utime('input/a_file', (atime, mtime))
+
+        test_archive1 = self.repository_location + '::test1'
+        self.cmd('create', test_archive1, "input")#self.input_path)
```
ThomasWaldmann, line 4215: same here

src/borg/testsuite/archiver.py
ThomasWaldmann: can you check?

ThomasWaldmann (changes requested):

src/borg/testsuite/archiver.py
palao: Well, in the test case it is an integer. The idea is to allow the user to provide a formatting, right?
This is a case that should fail without this feature and should pass once the feature is implemented.
Do I get it wrong? How houls I test this?

src/borg/testsuite/archiver.py
palao: Do you mean the `#self.input_path` part? I wasn't sure about what to use here and I left this comment there on purpose to get feedback...

src/borg/testsuite/archiver.py
palao: same as before...

src/borg/archiver.py
```diff
@@ -3776,6 +3776,11 @@ def define_borg_mount(parser):
                                help='Sort the output lines by file path.')
         subparser.add_argument('--json-lines', action='store_true',
                                help='Format output as JSON Lines. ')
+        subparser.add_argument('--format', '--diff-format', metavar='FORMAT', dest='format',
```
ThomasWaldmann, line 3779: iirc we had `--list-format` for borg list, but just deprecated it / got rid of it, only using `--format` in future.

so, guess we also only want `--format` here.

src/borg/testsuite/archiver.py
ThomasWaldmann: ah! just search the file you edited, there are lots of other such tests there.

src/borg/archiver.py
palao: Ok, fine. Thank you for the comment.

src/borg/testsuite/archiver.py
ThomasWaldmann: :-)

src/borg/testsuite/archiver.py
palao: It is a long file... I see cases where one solution seems to be used, and cases where the other is preferred (unless I got confused).
Maybe I should call the `create_test_files` aux method? Do you think it is a good idea, instead of creating files by hand?

src/borg/testsuite/archiver.py
ThomasWaldmann: but just "8d" won't generate output like in line 4223 (the normal integer formatting does not output a "+" for positive numbers).

but the implementation for `--format` is missing still, so you'll notice that later anyway.

palao: Yes, the implementation is missing: my bad. Sorry.
But you are right I would have notice it later.
And yes, the test is wrong: it should read `{deltasize:8d}` or change the next line to `modified (      33 B) 234900290 input/a_file`...

src/borg/testsuite/archiver.py
ThomasWaldmann: create_test_files / create_regular_file add some convenience, so use them if nothing speaks against that.

palao:
As the code is now, the tests pass. The code is asking for a refactor, though. I'm aware of that. The implementation is too simple and ugly.
But before refactoring, I think one needs to consider what are the real needs: what are the necessary placeholders? That will dictate how to modify the tests, and how to implement the functionality in a better way (`ItemDiffFormatter`?)

derekkaup (approved):
Gp

ThomasWaldmann:
@derekkaup what did you want to tell us here?

palao:
FWIW and just to repeat, or to make more clear my intentions: this PR's motivation is to get feedback to be able to follow a direction. The PR itself, as it is, is incomplete.

RF-Tar-Railt:
#7534

ThomasWaldmann:
@RF-Tar-Railt can you review / check this?

RF-Tar-Railt:
> @RF-Tar-Railt can you review / check this?

format part can be ignored, but the part of test can check

Issue #5942, opened 2021-07-26
some docs still reference freebsd 10.3 although we use 12.1


ThomasWaldmann:
master and 1.1-maint branch build on a fbsd 12.1 box meanwhile.

10.3 references:
```
22:04  tao$ https://borgbackup.readthedocs.io/en/stable/installation.html#pyinstaller-binary
```
maybe more, do a full search...

PR #5945, opened 2021-07-31
docs: update referenced freebsd version

remyabel:
Fixes: https://github.com/borgbackup/borg/issues/5942

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -110,10 +110,10 @@ are available on the releases_ page for the following platforms:
 
 * **Linux**: glibc >= 2.13 (ok for most supported Linux releases).
   Older glibc releases are untested and may not work.
-* **MacOS**: 10.10 or newer (To avoid signing issues download the file via
+* **MacOS**: 10.12 or newer (To avoid signing issues download the file via
   command line **or** remove the ``quarantine`` attribute after downloading:
   ``$ xattr -dr com.apple.quarantine borg-macosx64.tgz``)
-* **FreeBSD**: |freebsd_version| (unknown whether it works for older releases)
+* **FreeBSD**: 12.1 (unknown whether it works for older releases)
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
no, we can't do it like that:
- changes.rst should be readable in source form also, so there is no placeholder expansion
- entries in changes.rst (esp. older ones) are not supposed to change (when updating the placeholder definition)

remyabel:
Hmm OK, I'll redo it.

remyabel:
OK if I understood correctly, all I needed to do was revert the changes to the changelog. Couldn't find any other instances that needed to be updated.

docs/installation.rst
```diff
@@ -113,7 +113,7 @@ are available on the releases_ page for the following platforms:
 * **MacOS**: 10.10 or newer (To avoid signing issues download the file via
   command line **or** remove the ``quarantine`` attribute after downloading:
   ``$ xattr -dr com.apple.quarantine borg-macosx64.tgz``)
-* **FreeBSD**: 10.2 (unknown whether it works for older releases)
+* **FreeBSD**: |freebsd_version| (unknown whether it works for older releases)
```
ThomasWaldmann, line 116: if this is only used at a single place, guess it is easier to just have it literally there, not using a variable.

docs/installation.rst
```diff
@@ -113,7 +113,7 @@ are available on the releases_ page for the following platforms:
 * **MacOS**: 10.10 or newer (To avoid signing issues download the file via
```
ThomasWaldmann, line 113: this also needs fixing, it is macos sierra 10.12 now.

ThomasWaldmann:
Thanks!

ThomasWaldmann:
Can you check same for 1.1-maint (binary build OS versions see Vagrantfile there)?

PR #5951, opened 2021-08-07
docs: clarify pp pattern type

remyabel:
Fixes: #5300

This incorporates the comments from `patterns.py` into the documentation
to clarify the difference between `pp` and `pf`.


ThomasWaldmann:
thanks!

PR #5957, opened 2021-08-14
Use an exception catching to avoid a stack trace

Gu1nness:
When exclude files are used, this can trigger a stack trace if the file
does not exist. Instead exit nicely without the stacktrace.


ThomasWaldmann:
Thanks!

Issue #5723, opened 2021-03-04
borg mount: Permissions errors despite seemingly correct mount options


electricboogie:
Please fill in as much of the template as possible.
-->

## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG/ISSUE

#### Your borg version (borg -V).

borg 1.1.15

#### Operating system (distribution) and version.

MacOS 11.2.2
Ubuntu 20.04.2

#### How much data is handled by borg?

------------------------------------------------------------------------------
                       Original size      Compressed size    Deduplicated size
All archives:               13.39 GB              8.54 GB              7.43 GB

                       Unique chunks         Total chunks
Chunk index:                   50058                80196

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

borg mount --verbose /path/to/repo ~/mnt

## Describe the problem you're observing.

borg mount of MacOS backup mounted locally on a Linux remote repo machine mounts with the following default mount options:  "ro,nosuid,nodev,relatime,user_id=1000,group_id=1000,default_permissions" causes permissions errors like "ls: cannot open directory '.': Permission denied" and "bash: cd: ..: Permission denied."  Although user_id=1000, group_id=1000 are specified, borg seems to mount with the permissions of the MacOS machine (id=501, group=20) 

Remounting with following manual options:  "ro,nosuid,nodev,noatime,uid=1000,gid=1000,default_permissions" seems to fix this issue.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes.

ThomasWaldmann:
You are saying these are the **default** mount permissions.

What do you mean by that precisely? How do you know that? Provide full terminal output to show what you mean. Also show  the uid and gid of your current user.

https://www.man7.org/linux/man-pages/man8/fuse.8.html

It doesn't sound like `user_id` and `group_id` is something you shall use.

You can give `uid` and `gid` **if you want to override user and group of all files**.

Borg only parses the options it actually supports (like `uid` and `gid`) and doesn't do anything for stuff it does not support (like `user_id` and `group_id`.

electricboogie:
I mean when I pass "borg mount" with no options these are the options with which the borgfs is mounted.  See:

borg mount --verbose path/to/repo ~/mnt
mount | grep borgfs
[...]

PR #5958, opened 2021-08-18
docs: clarify user_id vs uid for fuse

remyabel:
Fixes: #5723

### DIFF
--- a/docs/usage/mount.rst.inc
+++ b/docs/usage/mount.rst.inc
@@ -124,6 +124,16 @@ To allow a regular user to use fstab entries, add the ``user`` option:
 
 For FUSE configuration and mount options, see the mount.fuse(8) manual page.
 
+The man page references ``user_id`` and ``uid``, which are not the same thing.
+``user_id`` is set by libfuse (or the filesystem) and contains the effective
+user id of the mount owner. ``uid`` on the other hand refers to the actual
+owner of the file. While you can pass ``uid`` as a mount option to borg to
+change the owner of all of the files (i.e, ``borg mount -o uid=1000,gid=1000``)
+it is recommended you use the ``--numeric-ids`` switch instead.
+
+``numeric-ids`` will use the ``uid``s and ``gid``s from the archive instead of
+attempting to map username and groupnames to local ids.
+
 Additional mount options supported by borg:
 
 - versions: when used with a repository mount, this gives a merged, versioned


docs/usage/mount.rst.inc
```diff
@@ -124,6 +124,16 @@ To allow a regular user to use fstab entries, add the ``user`` option:
...
+The man page references ``user_id`` and ``uid``, which are not the same thing.
+``user_id`` is set by libfuse (or the filesystem) and contains the effective
+user id of the mount owner. ``uid`` on the other hand refers to the actual
+owner of the file. While you can pass ``uid`` as a mount option to borg to
+change the owner of all of the files (i.e, ``borg mount -o uid=1000,gid=1000``)
+it is recommended you use the ``--numeric-ids`` switch instead.
+
+``numeric-ids`` will use the ``uid``s and ``gid``s from the archive instead of
+attempting to map username and groupnames to local ids.
+
```
ThomasWaldmann, line 136: same here: this is autogenerated, edit archiver.py.

also, I don't think we should recommend anything here (depends very much on circumstances/desires), but just describe the effects of the options.

remyabel: OK done. Also removed the word recommended.

src/borg/archiver.py
```diff
@@ -2916,6 +2916,17 @@ def define_borg_mount(parser):
...
+        thing. ``user_id`` is set by libfuse (or the filesystem) and contains
+        the effective user id of the mount owner. ``uid`` on the other hand
+        refers to the actual owner of the file. While you can pass ``uid`` as a
+        mount option to borg to change the owner of all of the files (i.e,
+        ``borg mount -o uid=1000,gid=1000``) you can use the ``--numeric-ids``
+        switch instead.
+
+        ``numeric-ids`` will use the ``uid``s and ``gid``s from the archive
+        instead of attempting to map username and groupnames to local ids.
+
```
ThomasWaldmann, line 2929: user/group: the default behaviour is to use the archived user/group names of each file and map them via the system's name-to-id mapping to the respective user/group ids. Alternatively, you can use ``numeric-ids`` (XXX check whether there is only --numeric-ids cli cmd option or also numeric-ids fstab option XXX) to use the archived user/group ids and not try to map from names.

The mount.fuse man page talks about ``user_id`` and ``group_id`` mount options (implemented by libfuse) which specify the user/group id of the mount owner (XXX what does that even mean? XXX). They are set automatically, you do not need to specify them (XXX could one specify them differently? XXX).

The ``uid`` and ``gid`` mount options (implemented by borg) are something different: you can use them to override user/group id of all files instead of using the archived user/group id.

ThomasWaldmann: Quite some unclarity (just wrote what I had in mind, didn't look up all the stuff / try it).

The key here is just telling what option causes what behaviour precisely. In order of usefulness / frequency of use.

remyabel: Alright I implemented your suggestions. According to https://github.com/libfuse/libfuse/blob/bdd2d4110fbc6d2059eb699efad2cca4a7eacccb/doc/kernel.txt mount owner just means "user who does the mounting". There is no definition for "filesystem owner", but extrapolating just means something similar. Looked at fstab options, don't really think there is a direct correspondence to `numeric-ids`.

remyabel:
PTAL.

src/borg/archiver.py
```diff
@@ -2916,6 +2916,20 @@ def define_borg_mount(parser):
 
         For FUSE configuration and mount options, see the mount.fuse(8) manual page.
 
+        Borg's default behavior is use the archived user and group names of each file
```
ThomasWaldmann, line 2919: ... is **to** use ...

src/borg/archiver.py
```diff
@@ -2916,6 +2916,20 @@ def define_borg_mount(parser):
...
+        and map them to the system's respective user and group ids. Alternatively, using
+        ``numeric-ids`` will instead used the archived user and group ids without any
```
ThomasWaldmann, line 2921: ... will ... **use** ...

src/borg/archiver.py
```diff
@@ -2916,6 +2916,20 @@ def define_borg_mount(parser):
...
+        mapping.
+
+        The ``uid`` and ``gid`` mount options (implemented by Borg) can be used to
+        override the user and group ids of all files (i.e, ``borg mount -o
```
ThomasWaldmann, line 2925: **i.e.**

src/borg/archiver.py
```diff
@@ -2916,6 +2916,20 @@ def define_borg_mount(parser):
...
+        uid=1000,gid=1000``).
+
+        The man page references ``user_id`` and ``group_id`` mount options (implemented
+        by fuse) which specify the user and group id of the mount owner (aka, the user
+        who does the mounting). It is set automatically by libfuse (or the filesystem if
+        libfuse is not used). However, you should not specify these manually.
```
ThomasWaldmann, line 2931: We need to add what's the precise effect of these options.

ThomasWaldmann (changes requested):
some grammar nitpicks and one thing that is still unclear.

Issue #5176, opened 2020-05-05
borg list patterns doc / example


akvadrako:
I was trying to use `borg list` with path patterns but the documentation was lacking. I didn't see how to do it until I found a email thread with an example. The docs only say:

>   | PATH | paths to list; patterns are supported

My understanding is that unlike with `--exclude`, patterns are not recognized by default, but require a prefix like `re:`. It would be nice to have at least one example and maybe a line that says something like:

> if `path` starts with a pattern prefix, it will be treated like a pattern. Otherwise it must be an exact match."

flolilo:
If I understand you correctly, this site might be what you were looking for:
https://borgbackup.readthedocs.io/en/stable/usage/help.html#borg-patterns

IMHO, this could be a little easier to find in the docs - I think I found it when looking through all search results of "exclude" in the RTD-search.

akvadrako:
Not quite. It’s easy to find the documentation for patterns, and the exclude options are fairly clear.

It’s not clear that paths are only patterns if they are prefixed. Or maybe more clearly, the default prefix for path is “pp”.

vjkr:
As it is labelled recently. I will give a try, Thanks Thomas

remyabel:
I created a PR that clarifies that the pattern in command paths is distinct from `--exclude` and `--pattern` and is path prefix by default like akvadrako said. Let me know if that's sufficient.

PR #5959, opened 2021-08-18
docs: clarify pattern usage with commands

remyabel:
Fixes: #5176

### DIFF
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -23,8 +23,7 @@ store all files as `some/path/.../file.ext` and ``borg create
 File patterns support these styles: fnmatch, shell, regular expressions,
 path prefixes and path full-matches. By default, fnmatch is used for
 ``--exclude`` patterns and shell-style is used for the experimental
-``--pattern`` option. For commands that support patterns in their ``PATH``
-argument like (``borg list``), the default pattern is path prefix.
+``--pattern`` option.
 
 Starting with Borg 1.2, for all but regular expression pattern matching
 styles, all paths are treated as relative, meaning that a leading path
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -23,7 +23,8 @@ store all files as `some/path/.../file.ext` and ``borg create
 File patterns support these styles: fnmatch, shell, regular expressions,
 path prefixes and path full-matches. By default, fnmatch is used for
 ``--exclude`` patterns and shell-style is used for the experimental
-``--pattern`` option.
+``--pattern`` option. For commands that support patterns in their ``PATH``
+argument like (``borg list``), the default pattern is path prefix.
 
 Starting with Borg 1.2, for all but regular expression pattern matching
 styles, all paths are treated as relative, meaning that a leading path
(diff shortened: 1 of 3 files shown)


docs/usage/help.rst.inc
```diff
@@ -23,7 +23,8 @@ store all files as `some/path/.../file.ext` and ``borg create
 File patterns support these styles: fnmatch, shell, regular expressions,
 path prefixes and path full-matches. By default, fnmatch is used for
 ``--exclude`` patterns and shell-style is used for the experimental
-``--pattern`` option.
+``--pattern`` option. For commands that support patterns in their ``PATH``
+argument like (``borg list``), the default pattern is path prefix.
```
ThomasWaldmann, line 27: see 1st line of this file: this is autogenerated.

so you need to revert this change and do the respective change in `archiver.py`.

docs/usage/list.rst
```diff
@@ -32,3 +32,12 @@ Examples
...
     -rw-rw-r-- user   user    1416192 Sun, 2015-02-01 11:00:00 code/myproject/file.ext
     ...
+
+    $ borg list /path/to/repo::archiveA 'code/myproject'
+    drwxrwxr-x user   user          0 Sun, 2015-02-01 11:00:00 code/myproject
+    -rw-rw-r-- user   user    1416192 Sun, 2015-02-01 11:00:00 code/myproject/file.ext
+    ...
+
+    $ borg list /path/to/repo::archiveA 'pf:code/myproject'
+    drwxrwxr-x user   user          0 Sun, 2015-02-01 11:00:00 code/myproject
```
ThomasWaldmann, line 42: pf is maybe not the best example here - who would use this in practice like that?

maybe rather `re:....`?

ThomasWaldmann: also, please try whether this actually works. :-)

ThomasWaldmann (changes requested):
thanks for the PR, found some issues...

docs/usage/help.rst.inc
remyabel: Sorry, forgot.

docs/usage/list.rst
remyabel: I tried the pattern on my existing repo. Let me know if it's still wrong.

docs/usage/list.rst
```diff
@@ -32,3 +32,7 @@ Examples
     drwxrwxr-x user   user          0 Sun, 2015-02-01 11:00:00 code/myproject
     -rw-rw-r-- user   user    1416192 Sun, 2015-02-01 11:00:00 code/myproject/file.ext
     ...
+
+    $ borg list /path/to/repo/::archiveA --pattern "re:ext$"
```
ThomasWaldmann, line 36: `'re:\.ext$'` would make sense:

- shows how to escape a dot (please check whether \ or \\ is needed)
- uses single quotes, so the shell does not expand anything
- makes sure only .ext extension matches and not .text also

remyabel: Should be good now. As far as I can tell both `\` and `\\` have the same effect.

ThomasWaldmann (approved):
LGTM

Issue #5770, opened 2021-04-19
Borg erroneously thinks my files have changed


auanasgheps:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Issue

#### Your borg version (borg -V).
borg 1.1.15

#### Operating system (distribution) and version.
Debian 10 / OMV5

#### Hardware / network configuration, and filesystems used.
Source: Debian 10 server, ext4, borg 1.1.15
Destination: remote Synology NAS running borg 1.1.13

#### How much data is handled by borg?

4 TB (uncompressed, deduplicated)

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

Please note I am using borgmatic to automate. The actual executed command is:

`borg create --exclude-from /tmp/tmp410uy2bb --compression zstd --remote-ratelimit 7300 --remote-path /usr/local/bin/borg --list --filter AME- --info MYUSER@MYSERVER:backup::{hostname}-{now:%Y-%m-%dT%H:%M:%S} /srv/dev-disk-by-label-nvme /srv/fb4ab0da-2b79-4740-834a-7435c0123e28`

## Describe the problem you're observing.
Hi, I've been happily running backups with borg every week for last 4 months from my local server to a remote Synology NAS.

Last run (on Saturday) did not end when I expected, and I had to stop it. Most of my data does not change every week, and even if I do big stuff in 3 to 4 hours it's done.

When enabling bormatic file ouput, I discovered that borg thinks my files have been modified: there's an M at the beginning of every entry.

They have not.

In fact I also run a rsync job every week on a USB hard drive, and the transfer was quite short, since almost nothing changed from the week before. I believe rsync uses date/time to check if a file has changed. Either way, I'm 100% those files haven't been touched.

This is the first time I'm seeing the issue since I've setup borg. 

Since I had this issue the backup has been running for about 20 hours, but hasn't completed yet. It's not a problem of bandwidth: most of the time is spent reading non-changed local files that borg thinks have changed.

After the upload of the actual changed files is dove, borg is only uploading these files:

![immagine](https://user-images.githubusercontent.com/20586878/115275789-8965dd80-a142-11eb-9929-acb26e57323b.png)

I can't afford to wait for borg to read **ALL**  of my data again!

I haven't done any other change to borg client, borg server or my data.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

[...]

PR #5965, opened 2021-08-22
docs: document behavior for filesystems with inconsistent inodes

Jubjub:
Fixes issue #5770


docs/faq.rst
```diff
@@ -987,6 +987,12 @@ stable. If the directory where you mount a filesystem is different every time,
 Borg assumes they are different files. This is true even if you backup these
 files with relative pathnames - borg uses full pathnames in files cache regardless.
 
+It is possible for some filesystems, such as mergerfs_,
+to return inconsistent inode numbers across runs, causing borg to consider them
+changed. A workaround is to set the option ``--ignore-inode`` so that files with
```
ThomasWaldmann, line 992: iirc we deprecated --ignore-inode in favour of --files-cache (at least for master branch, please also check 1.1-maint, see docs/changes.rst.

ThomasWaldmann: also it could be noted that unstable inodes happen also on network filesystems.

docs/faq.rst
```diff
@@ -987,6 +987,12 @@ stable. If the directory where you mount a filesystem is different every time,
...
+different inodes won't be treated as modified.
+
+.. _mergerfs: https://github.com/trapexit/mergerfs
```
ThomasWaldmann, line 995: i don't think we need to link to mergerfs.

docs/faq.rst
Jubjub: I've updated that, --ignore-inode was indeed deprecated back in 1.1.0rc4 so the doc changes should apply to the maintenance branch as well.

docs/faq.rst
Jubjub: I've removed the link.

docs/faq.rst
```diff
@@ -987,6 +987,12 @@ stable. If the directory where you mount a filesystem is different every time,
 Borg assumes they are different files. This is true even if you backup these
 files with relative pathnames - borg uses full pathnames in files cache regardless.
 
+It is possible for some filesystems, such as ``mergerfs`` or network-mounted
```
ThomasWaldmann, line 990: maybe rather say "network filesystems". the mount is on a local directory, but the filesystem is via the network.

docs/faq.rst
```diff
@@ -987,6 +987,12 @@ stable. If the directory where you mount a filesystem is different every time,
...
+filesystems, to return inconsistent inode numbers across runs, causing borg to
+consider them changed. A workaround is to set the option ``--files-cache=ctime,size``
+to exclude the inode from the default cache calculation so that files with different
```
ThomasWaldmann, line 993: ... to exclude the ``inode`` number comparison from the files cache check so that files with different inode numbers ...

ThomasWaldmann (commented):
Thanks for updating this, I found 2 more little nitpicks.

PR #5973, opened 2021-08-29
Add an option --extract-to-source-path to 'borg extract' command

Mark-Joy:
Add an option `--extract-to-source-path` to `borg extract` command: extract the content to its source path where it was backup from. `borg extract` default behaviour always writes to current directory. This option overrides that default behaviour.

Changes in source code:
1. Add attribute `source_path` to class `Item`. This attr holds the real source path (without being modified by `make_path_safe`) when `borg create` an archive.
2. Add an option `--extract-to-source-path` to `borg extract` command which uses `Item.source_path` instead of `Item.path` for extracting.
3. Since `Item.source_path` is newly added and only be used with `--extract-to-source-path`, these changes will NOT affect the default behaviour.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -694,7 +694,7 @@ Utilization of max. archive size: {csize_max:.0%}
                 # In this case, we *want* to extract twice, because there is no other way.
                 pass
 
-    def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, sparse=False,
+    def extract_item(self, item, restore_attrs=True, dry_run=False, stdout=False, ex2src=False, sparse=False,
                      hardlink_masters=None, stripped_components=0, original_path=None, pi=None):
         """
         Extract archive item.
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Well, guess this needs some discussion / community feedback first.

Using this option obviously could lead to mix ups or overwriting of data (if some data is already present at the restore target path). It helps that this is not the default behaviour, but still there is "shooting in its own foot" potential.

A minor thing I don't like is that most of the path is now duplicate in the metadata. Guess this could be fixed by splitting into prefix and path, so that abs_path == os.path.join(prefix, path).

Mark-Joy:
It is true that it may lead to overwriting of data, but it also depends on how users use this option. I am sure that it is like any other option, he/she must understand how it works before using it.
If you have suggestion regarding the switch's name and its description, I'll happily fix it. About other points:
1. sure
2. an exception shall be raised for old archives, explain about the new option
3. To be honest, my skill is mediocre. So I would rather take someone help for test cases. Although, I have tested it myself with some basic silly test cases. And I am coding on my phone only (PC broken, covid lockdown)
After your feedback, I'll fix the code and update the PR in case you want to use it.

elho:
I agree with the criticism.  Extraction should be a conscious process and storing the backup root(s) for each item is hideous.
It would be of zero use to anyone doing his backups properly from read-only and/or temporary filesystem snapshots, too.

If you have a specific use case where you are unable to remember what you backup, have your backup script add the backup root of each archive to the archive comment and the extract script grab them from there. Or, if wanting multiple disjunctive backup rotos in one archive, just go backing up from the fs root as single backup root with according include/exclude patterns restricting to the wanted multiple paths. That you then can extract into / if you want to extract in place.

PR #5987, opened 2021-09-15
require python 3.8+, fixes #5975

### LINKED ISSUE #5975: require py38 (master branch /  borg 1.2)
considering that there is widespread support for 3.8+ meanwhile and creating the next release (1.2) is still todo, i guess we could also drop support for py37 and require py38 as the minimum for borg >= 1.2.

anyone seeing a problem with this?

everybody who is still on <=py37 seems to prefer older / stable stuff, so can as well just use borg 1.1.x.

ThomasWaldmann:
if you do not have py38+ yet, just use borg 1.1.x.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,8 @@ jobs:
       matrix:
         include:
             - os: ubuntu-20.04
-              python-version: '3.7'
-              toxenv: py37-fuse2
-            - os: ubuntu-20.04
               python-version: '3.8'
-              toxenv: py38-fuse3
+              toxenv: py38-fuse2
             - os: ubuntu-20.04
               python-version: '3.9'
               toxenv: py39-fuse3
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -66,8 +63,8 @@ jobs:
             - os: macos-latest
               # note: it seems that 3.8 and 3.9 are currently broken,
               # neverending RuntimeError crashes...
-              python-version: '3.7'
-              toxenv: py37-fuse2
+              python-version: '3.8'
+              toxenv: py38-fuse2
```
ThomasWaldmann, line 67: let's see if it is still broken...

Vagrantfile
```diff
@@ -53,7 +53,7 @@ def packages_freebsd
     pkg install -y fusefs-libs3 || true
     pkg install -y git bash  # fakeroot causes lots of troubles on freebsd
     # for building python:
-    pkg install -y python37 py37-sqlite3 py37-virtualenv py37-pip
+    pkg install -y python37 py37-sqlite3 py37-virtualenv py37-pip  # TODO -> py38?
```
ThomasWaldmann, line 56: TODO

.github/workflows/ci.yml
ThomasWaldmann: works!

Issue #6009, opened 2021-10-13
Passing device file paths via `--paths-from-stdin` triggers a TypeError exception


jimparis:
<!--
Thank you for reporting an issue.

*IMPORTANT* -  *before* creating a new issue please look around:
 - Borgbackup documentation: http://borgbackup.readthedocs.io/en/stable/index.html
 - FAQ: https://borgbackup.readthedocs.io/en/stable/faq.html
 and
 - open issues in Github tracker: https://github.com/borgbackup/borg/issues
  
If you cannot find a similar problem, then create a new issue.

Please fill in as much of the template as possible.
-->

## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug

## System information. For client/server mode post info for both machines.

#### Your borg version (borg -V).

borg 1.2.0b4.dev83+g93e0d0f614ae

#### Operating system (distribution) and version.

Debian bullseye

#### Hardware / network configuration, and filesystems used.

ext4

#### How much data is handled by borg?

n/a

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

see below

## Describe the problem you're observing.

Passing device file paths via `--paths-from-stdin` triggers a TypeError exception.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

```bash
#!/bin/bash                                                                                                                                                                                                             
set -e
export BORG_PASSPHRASE=asdf
borg init --encryption=repokey /tmp/test-repo
# this works                                                                                                                                                                                                            
borg create --verbose --stats /tmp/test-repo::test1 /dev/zero
# this fails                                                                                                                                                                                                            
echo /dev/zero | borg create --verbose --stats --paths-from-stdin /tmp/test-repo::test2
```

#### Include any warning/errors/backtraces from the system logs

```
Creating archive at "/tmp/test-repo::test2"
Local Exception
Traceback (most recent call last):
  File "/home/jim/git/borg/src/borg/archiver.py", line 4984, in main
    exit_code = archiver.run(args)
  File "/home/jim/git/borg/src/borg/archiver.py", line 4915, in run
    return set_ec(func(args))
  File "/home/jim/git/borg/src/borg/archiver.py", line 183, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File "/home/jim/git/borg/src/borg/archiver.py", line 663, in do_create
    create_inner(archive, cache, fso)
[...]

PR #6011, opened 2021-10-14
create: fix passing device nodes and symlinks to --paths-from-stdin

jimparis:
Paths that come from --paths-from-stdin or --paths-from-command don't
have a parent_fd or name, so we need to use the os_stat helper that
falls back on the full path if those are missing.

Fixes borgbackup/borg#6009


jimparis:
1.1 doesn't support `--paths-from-stdin`/`--paths-from-command` so I don't think so

PR #6016, opened 2021-10-18
Fix manifest-timestamp path in docs

strugee:


strugee:
Sure, thanks for merging! Followup filed.

Issue #5995, opened 2021-10-01
Remote repository seems to get corrupted easily, and then borg check --repair hangs indefinitely


SeanDS:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug / issue report

#### Your borg version (borg -V).

Server: borg 1.1.15
Client: borg 1.1.17

#### Operating system (distribution) and version.

Server: Ubuntu 20.04.3 LTS
Client: Ubuntu 20.04.3 LTS

#### Hardware / network configuration, and filesystems used.

Backup over SSH using [borgmatic](https://torsion.org/borgmatic/). Runs as root on the client.

Client uses ext4 on top of an LVM, server uses just ext4.

#### How much data is handled by borg?

Roughly 200 GB.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

Error on already-corrupt repository triggered by `borg create --exclude-from /etc/borgmatic/excludes --exclude-caches --compression zstd --remote-ratelimit 500 --info user@server:/path/to/backup/::{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f} /data/media /etc /home /srv /var` on client.

Hang while trying to repair triggered by `borg check --repair --progress --debug /path/to/backup` on server.

## Describe the problem you're observing.

I have a nightly backup scheduled between client and server over SSH. On at least three occasions, after the initial 200 GB backup is made, after a period of days, weeks, or months of successful incremental backups, the remote repository will develop what appears to be corruption, with subsequent backups reporting the error:

```
Data integrity error: Segment entry checksum mismatch [segment 2321, offset 483648553]
```

On each occasion, after some temporary fixes (see below) have stopped working, I have nuked the repository and started again with a full backup, but eventually the corruption returns (days, weeks or months later).

After this error starts to appear, running a `borg check --repair` on the server sometimes fixes the issue for a few days before it returns. However, eventually `borg check --repair` seems to enter a state where it hangs indefinitely after printing something like:

```
using builtin fallback logging configuration
35 self tests completed in 0.14 seconds
This is a potentially dangerous function.
check --repair might lead to data loss (for kinds of corruption it is not
capable of dealing with). BE VERY CAREFUL!

Type 'YES' if you understand this and want to continue: YES
Starting repository check
Verified integrity of /path/to/backup/index.3017
Read committed index of transaction 3017
[...]

PR #6022, opened 2021-10-25
Speed up search for next valid object in segment in --repair mode

pgerber:
When an object is corrupted, the start position of the next object
will not be known as the size field belonging to the corrupted
object may be corrupted as well. In order to find the next object
within the segment, the remainder is scanned for the next valid
object, byte-by-byte. An object is considered valid if the CRC
checksum matches the content. However, doing so the scan accepted
any object size that fit within the remainder of the segment. As a
result, in particular when the corruption occurred near the start
of a segment, CRC checksums were calculated for large objects,
often hundreds of megabytes in size, despite the size being limited
to 20 MiB. This change makes it so that CRC calculation is skipped
when the object header indicates an impossible size, thereby,
greatly reducing the number of CPU cycles used for CRC calculations.
In my case, this brought down the time for repair from hours to mere
minutes.

This has also the additional benefit that there is some verification
in addition to the CRC checksum. The 4-bytes checksum is rather
short considering the amount of data that might be in an archive.

Likely fixes the hanging --repair in #5995 also.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -38,6 +38,15 @@ TAG_PUT = 0
 TAG_DELETE = 1
 TAG_COMMIT = 2
 
+# Highest ID usable as TAG_* value
+#
+# Code may expect not to find any tags exceeding this value. In particular,
+# in order to speed up `borg check --repair`, any tag greater than MAX_TAG_ID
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@pgerber thanks for fixing this! could you do the 1.1-maint backport also?

Issue #6014, opened 2021-10-17
borg create: bad error msg when repo not found


ThomasWaldmann:
borg 1.1.16 on debian bullseye:

```
Repository ssh://borg@backup:22/mnt/backup/server::server-home-2021-10-16T23:50:01 does not exist.
```

Guess the backup disk was not mounted at that moment, so the repo at that path was not found.

It's a bit bad though that the error message also mentions the **archive name** which it intended to create.

Of course, the full location specified on the command line for borg create included the to-be-created archive name, but it should not be shown in that error msg, but only the repo part.

PR #6023, opened 2021-10-27
do not show archive name in error msgs referring to repository

sw9719:
Fixes [issue](https://github.com/borgbackup/borg/issues/6014) where the error message would display the archive when the Respository does not exist by removing the archive name from the error string when `Repository.DoesNotExist`
is raised.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -493,8 +493,9 @@ class Location:
         })
 
     def omit_archive(self):
-        #self.orig = self.orig.split("::")[0]
-        return self
+        loc = Location(self.orig)
+        loc.orig = loc.orig.split("::")[0]
+        return loc
 
 
 def location_validator(archive=None, proto=None):
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -526,8 +526,6 @@ def archivename_validator():
     return validator
 
 
-
-
 class BaseFormatter:
     FIXED_KEYS = {
         # Formatting aids
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -494,6 +494,7 @@ class Location:
 
     def omit_archive(self):
         loc = Location(self.orig)
+        loc.archive = None
         loc.orig = loc.orig.split("::")[0]
         return loc
 
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -492,6 +492,10 @@ class Location:
             'utcnow': DatetimeWrapper(timestamp),
         })
 
+    def omit_archive(self):
+        #self.orig = self.orig.split("::")[0]
+        return self
+
 
 def location_validator(archive=None, proto=None):
     def validator(text):
@@ -519,6 +523,8 @@ def archivename_validator():
             raise argparse.ArgumentTypeError('Invalid archive name: "%s"' % text)
         return text
     return validator
+
+
 
 
 class BaseFormatter:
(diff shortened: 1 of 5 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -492,6 +492,10 @@ def with_timestamp(self, timestamp):
             'utcnow': DatetimeWrapper(timestamp),
         })
 
+    def omit_archive(self):
+        #self.orig = self.orig.split("::")[0]
+        return self
```
ThomasWaldmann, line 497: This must be something like:
```
loc = Location(...)  # create new instance
... # modify instance as needed
return loc
```

sw9719:
I did modify why the logic as mentioned. I hope this works.

src/borg/helpers/parseformat.py
```diff
@@ -493,8 +493,9 @@ def with_timestamp(self, timestamp):
         })
 
     def omit_archive(self):
-        #self.orig = self.orig.split("::")[0]
-        return self
+        loc = Location(self.orig)
+        loc.orig = loc.orig.split("::")[0]
+        return loc
```
ThomasWaldmann: can you please add a unit test that checks whether this works as expected?

also check the value of `loc.archive` within that test.

sw9719: I am trying to figure out where to write the new test. Does testsuite/helpers.py sound good?

src/borg/testsuite/repository.py
```diff
@@ -856,7 +856,7 @@ class RemoteRepositoryTestCase(RepositoryTestCase):
     repository = None  # type: RemoteRepository
 
     def open(self, create=False):
-        return RemoteRepository(Location('__testsuite__:' + os.path.join(self.tmppath, 'repository')),
+        return RemoteRepository(Location('__testsuite__:' + os.path.join(self.tmppath, 'repository')+"::archive"),
```
ThomasWaldmann, line 859: why is this needed?

sw9719: Sorry, this was left over from something I was testing. Ill be more careful about this.

sw9719:
I added tests for when repo does not contain archive name

src/borg/testsuite/helpers.py
```diff
@@ -234,6 +234,12 @@ def test_bad_syntax(self):
             # this is invalid due to the 2nd colon, correct: 'ssh://user@host/path'
             Location('ssh://user@host:/path')
 
+    def test_omit_archive(self):
+        loc = Location('ssh://user@host:1234/some/path::archive')
+        loc_without_archive = loc.omit_archive()
+        assert loc_without_archive.archive == "archive"
```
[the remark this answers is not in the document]
ThomasWaldmann: did you see this ^^^?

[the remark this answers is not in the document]
sw9719: So I checked this. When '::<archive'-name> is present in the Location string, `Locatoin.archive` attribute will contain the string after '::".

With 'Location('ssh://user@host:1234/some/path')', i.e. without '::<archive-name', `Location.archive` attributed is `None.`

So I added tests for both test cases: When archive name is present in string and when it's not.

Since only `Location.orig` is being used to raise appropriate exceptions, I only modified that attribute. If needed I can change the value of `Location.archive` when returning from `omit_archive()` method, but I hope it does not break anything.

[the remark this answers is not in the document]
ThomasWaldmann: OK, so please set `.archive` to None within `omit_archive`.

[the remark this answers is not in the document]
sw9719: Done

ThomasWaldmann:
Thanks! Can you check if the same change applies to 1.1-maint branch?

sw9719:
> Thanks! Can you check if the same change applies to 1.1-maint branch?

I don't see the new test in 1.1-maint branch. Also, `src/borg/helpers/` dir which contained changes in `Location` class in `parseformat.py`.

ThomasWaldmann:
`borg.helpers` used to be a module (`helpers.py`) in 1.1 and was split up into a package (directory) in master.

hexagonrecursion:
I'll work on backporting this

Issue #6024, opened 2021-10-27
EROFS due to O_NOATIME with volume shadow copy in WSL1


cuevavirus:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug

#### Your borg version (borg -V).

borg 1.1.17

#### Operating system (distribution) and version.

Windows 10 21h1

#### Hardware / network configuration, and filesystems used.

WSL1 drvfs

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

A volume shadow copy is made available in WSL1 with these steps:

1. Create a symlink in Windows with the shadow copy's UNC path.
    `mklink /d c \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy1\`
2. Map symlink to a drive letter.
    `subst z: c`
3. Mount the drive in WSL1.
    `mount -t drvfs -o ro z: c`

When borg encounters a file in this mounted volume owned by the invoking user, the file is opened with `O_NOATIME` among other flags, which produces this error `open: [Errno 30] Read-only file system`. When the file is not owned by the invoking user, `O_NOATIME` is not used and it works fine.

Can be confirmed with this patch

```diff
diff --git a/src/borg/archive.py b/src/borg/archive.py
index 93385c9d..4c08f9a7 100644
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -47,7 +47,7 @@
 has_link = hasattr(os, 'link')

 flags_normal = os.O_RDONLY | getattr(os, 'O_BINARY', 0)
-flags_noatime = flags_normal | getattr(os, 'O_NOATIME', 0)
+flags_noatime = flags_normal


 class Statistics:
```

PR #6027, opened 2021-10-29
Workaround for volume shadow copy in WSL1

cuevavirus:
Fixes https://github.com/borgbackup/borg/issues/6024

src/borg/archive.py changed too much in master for this patch to apply. If master is affected by the same problem a different workaround will have to be used.

### DIFF
--- a/docs/usage/general/environment.rst.inc
+++ b/docs/usage/general/environment.rst.inc
@@ -94,6 +94,11 @@ General:
             Using this does not affect data safety, but might result in a more bursty
             write to disk behaviour (not continuously streaming to disk).
 
+        retry_erofs
+            Retry opening a file without O_NOATIME if opening a file with O_NOATIME
+            caused EROFS. You will need this to make archives from volume shadow copies
+            in WSL1 (Windows Subsystem for Linux 1).
+
 Some automatic "answerers" (if set, they automatically answer confirmation questions):
     BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK=no (or =yes)
         For "Warning: Attempting to access a previously unknown unencrypted repository"
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -37,6 +37,7 @@ from .helpers import bin_to_hex
 from .helpers import safe_ns
 from .helpers import ellipsis_truncate, ProgressIndicatorPercent, log_multi
 from .helpers import msgpack
+from .helpers import workarounds
 from .patterns import PathPrefixPattern, FnmatchPattern, IECommand
 from .item import Item, ArchiveItem
 from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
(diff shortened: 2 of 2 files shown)


src/borg/archive.py
```diff
@@ -1115,6 +1115,10 @@ def _open_rb(path):
                 raise
             # Was this EPERM due to the O_NOATIME flag? Try again without it:
             return os.open(path, flags_normal)
+        except OSError as exc:
+            # O_NOATIME causes EROFS when accessing a volume shadow copy in WSL1
+            if exc.errno == errno.EROFS:
+                return os.open(path, flags_normal)
```
ThomasWaldmann, line 1121: Hmm, this matches the strange behaviour you have seen on WSL1, but I think:
- this is a strange bug in WSL1
- this should not be the standard reaction of borg if it encounters a EROFS in general

So, please extend BORG_WORKAROUNDS and only implement this behaviour if specifically requested by a specific value in the workarounds list.

cuevavirus:
Extended BORG_WORKAROUNDS with retryreadonly.

docs/man/borg.1
```diff
@@ -484,6 +484,11 @@ You might need this to run borg on WSL (Windows Subsystem for Linux) or
 in systemd.nspawn containers on some architectures (e.g. ARM).
 Using this does not affect data safety, but might result in a more bursty
 write to disk behaviour (not continuously streaming to disk).
+.TP
+.B retryreadonly
+Retry opening a file without O_NOATIME if opening a file with O_NOATIME
+caused EROFS. You will need this to make archives from volume shadow copies
+in WSL1 (Windows Subsystem for Linux 1).
 .UNINDENT
```
ThomasWaldmann, line 492: this is a generated file, do not edit / update / commit it (see line 1).

(all these files will be updated automatically at release time)

docs/usage/general/environment.rst.inc
```diff
@@ -94,6 +94,11 @@ General:
             Using this does not affect data safety, but might result in a more bursty
             write to disk behaviour (not continuously streaming to disk).
 
+        retryreadonly
+            Retry opening a file without O_NOATIME if opening a file with O_NOATIME
+            caused EROFS. You will need this to make archives from volume shadow copies
+            in WSL1 (Windows Subsystem for Linux 1).
+
```
ThomasWaldmann, line 101: Yeah, this seems to be the right place to edit.

`retry_erofs` maybe?

src/borg/archive.py
```diff
@@ -1115,6 +1116,12 @@ def _open_rb(path):
                 raise
             # Was this EPERM due to the O_NOATIME flag? Try again without it:
             return os.open(path, flags_normal)
+        except OSError as exc:
+            # O_NOATIME causes EROFS when accessing a volume shadow copy in WSL1
+            if 'retryreadonly' in workarounds and exc.errno == errno.EROFS:
+                return os.open(path, flags_normal)
+            else:
+                raise exc
```
ThomasWaldmann, line 1124: guess this is just `raise`, see line 1116.

also, check the code in 1114..1116 - if we do not have NOATIME at all the initial open did not use it, so there is no need to try it again with same flags.

cuevavirus:
Made changes according to comments.

ThomasWaldmann:
Looks good! You tested it on WSL1?

cuevavirus:
Yes tested it on WSL1.

ThomasWaldmann:
OK, merged. Can you check if master branch has the same issue / needs a similar fix?

Issue #6036, opened 2021-11-07
Borg key mismatch


Killamus:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug/Issue

#### Your borg version (borg -V).
borg 1.1.17 (Client and server match)

#### Operating system (distribution) and version.
Archlinux/Centos/TrueNAS(BSD-based)

#### Hardware / network configuration, and filesystems used.
Central NAS uploading to Rsync.net nightly. ZFS filesystem on client and server.

#### How much data is handled by borg?
250ish-GB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
Unknown - backup command is: 

```
export BORG_KEY_FILE=/root/borg.key
export BORG_RSH='ssh -i /root/.ssh/rsync'
/usr/local/bin/borg create --list --stats --compression lzma --exclude-caches \
--filter AME \
--exclude-if-present .borg-ignore \
--remote-path 'borg1' \
-p ssh://USERNAME@HOST.rsync.net/data1/home/USERNAME/borg::'{hostname}-{now}' /mnt/Media
```

## Describe the problem you're observing.
I've been using BORG for about 2-ish years now on the same backup. Recently I migrated to TrueNAS, and was able to continue my backups. However, last night while experimenting, I did... something? And now I'm getting a key mismatch on my main backup. Things that I did last night:

1) Regular backup
2) Borg mount
2.1) Successfully mounted and restored a backed-up file
Note: Below here, I was testing how BORG creates new files for the purposes of determining the cost of a nightly sync of a 2TB repo to AWS Deep-Glacier
3) borg init --encryption=keyfile /mnt/Temp/backup
3.1) After this, my repokey in /root/borg.key DID change, at exactly the same time that I did the borg init. This was unexpected, I would have thought that this would create a new repository with the same key.
4) borg create --list --stats /mnt/Temp/backup::first /mnt/Media/Camera (12GB folder)
4.1) Backup created successfully
5) cp /mnt/Temp/backup /mnt/Temp/backup0
6) cp /mnt/Media/Secrets /mnt/Media/Camera (1GB folder to simulate added, non-duplicated data)
7) borg create --list --stats /mnt/Temp/backup::second /mnt/Media/Camera 
8) diff /mnt/Temp/backup /mnt/Temp/backup0
9) borg create --list --stats /mnt/Temp/backup::third /mnt/Media/Camera 
[...]

PR #6046, opened 2021-11-12
Disallow overwriting of existing keyfiles on init

Killamus:
This commit aims to check to ensure that keys are not overwritten on keyfile repositories by the use of the BORG_KEY_FILE variable. Pretty basic, but I wasn't sure best practice to exit immediately, which is why it throws an exception. If Borg has a better way to handle that, let me know. Took inspiration from https://github.com/borgbackup/borg/blob/450ab38e4227773db83ad45edf5d5e8df7d4e1b8/src/borg/crypto/key.py#L112

Moved passphrase below so people don't put in a passphrase then get told the key is bad.

Issue that spawned this MR:
https://github.com/borgbackup/borg/issues/6036

### DIFF
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -684,7 +684,7 @@ class KeyfileKeyBase(AESKeyBase):
         logger.info('Keep this key safe. Your data will be inaccessible without it.')
         return key
 
-    def save(self, target, passphrase):
+    def save(self, target, passphrase, create=False):
         raise NotImplementedError
 
     def get_new_target(self, args):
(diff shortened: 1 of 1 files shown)


src/borg/crypto/key.py
```diff
@@ -673,12 +673,14 @@ def change_passphrase(self, passphrase=None):
...
     @classmethod
     def create(cls, repository, args):
-        passphrase = Passphrase.new(allow_empty=True)
         key = cls(repository)
         key.repository_id = repository.id
         key.init_from_random_data()
         key.init_ciphers()
         target = key.get_new_target(args)
+        if os.path.isfile(target):
+            raise ValueError('Key in "%s" already exists. Not creating repository. Please move or rename this key first.' % target)
```
ThomasWaldmann, line 682: This will generate a traceback (which is not very useful for expected errors with known root cause).

You can use `Error` or a subclass of it to avoid the traceback, give a prettier error msg (but otherwise behave similarly).

ThomasWaldmann: Also, `target` seems to be a repository instance, this is why tests are failing currently.

ThomasWaldmann:
@Killamus did you see my feedback?

src/borg/crypto/key.py
ThomasWaldmann: guess you need to move this check to `KeyfileKey.save` - there it is already clear that we deal with some specific path and a local file (not a `RepoKey`).

ThomasWaldmann (changes requested):
see the inline comments.

ThomasWaldmann:
maybe I'll fix this myself, continuing the work in this PR (not much time left before 1.2.0rc1 though).

src/borg/crypto/key.py
ThomasWaldmann: done.

ThomasWaldmann:
implemented my suggestions, let's see what the tests say...

ThomasWaldmann (approved):
LGTM :-)

ThomasWaldmann:
@Killamus done! have a look!

Issue #6028, opened 2021-10-31
race condition, when calling borg init first time for two repositories simultaneously


hmb:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

## System information. For client/server mode post info for both machines.

borg 1.1.15

Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal

#### Hardware / network configuration, and filesystems used.

It's on a cloud system, using ext4. The bug is probably unrelated to the hardware.

#### How much data is handled by borg?

None, bug happens at `init` time.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

    BORG_PASSCOMMAND="cat borgpasswd-client1.pwd" borg init --encryption=keyfile ssh://backupstore-key/./client1

`backupstore-key` is a ssh configuration in `.ssh/config`:

```
Host backupstore-key
  Hostname u0000.server.de
  Port 23
  User u0000
  IdentityFile ~/.ssh/backup-host00
  IdentitiesOnly yes
```

## Describe the problem you're observing.

The host has multiple client VMs using LXC/LXD, which are all backuped with borg running from the host. When initializing the host with *Ansible* the above `borg init` is run for every client. Due to the scripted nature of Ansible this happens - more or less - simultaneously. Sometimes Ansible stops with an exception:

    FileExistsError: [Errno 17] File exists: '/root/.config/borg'

This is on a freshly installed system that is setup from scratch with two client VMs. My sight on what happens is, as the two `borg init` commands are issued shortly after another, both find the `.config/borg/` directory missing, so both try to create it. This obvoiusly leads to one process creating the directory first and the second trying to create it again with the already existing directory. Apparently python does not like that. It crashes with the followin stack trace:

```
Local Exception
Traceback (most recent call last):
  File \"/usr/lib/python3/dist-packages/borg/archiver.py\", line 4621, in main
    exit_code = archiver.run(args)
  File \"/usr/lib/python3/dist-packages/borg/archiver.py\", line 4553, in run
    return set_ec(func(args))
  File \"/usr/lib/python3/dist-packages/borg/archiver.py\", line 177, in wrapper
    return method(self, args, repository=repository, **kwargs)
  File \"/usr/lib/python3/dist-packages/borg/archiver.py\", line 291, in do_init
    key = key_creator(repository, args)
[...]

PR #6056, opened 2021-11-16
atomically create the CACHE_TAG file, see #6028

ThomasWaldmann:


src/borg/helpers/fs.py
```diff
@@ -92,15 +92,15 @@ def get_cache_dir():
...
-            """).encode('ascii'))
+    cache_tag_fn = os.path.join(cache_dir, CACHE_TAG_NAME)
+    cache_tag_contents = CACHE_TAG_CONTENTS + textwrap.dedent("""
+    # This file is a cache directory tag created by Borg.
+    # For information about cache directory tags, see:
+    #       http://www.bford.info/cachedir/spec.html
+    """).encode('ascii')
+    from ..platform import SaveFile
+    with SaveFile(cache_tag_fn, binary=True) as fd:
+        fd.write(cache_tag_contents)
```
ThomasWaldmann, line 103: Hmm, nope.

That would create, write and rename the tag file for each call of `get_cache_dir()`, which is expensive / bad in other ways.

So guess we still want to use the "not exists" condition, even if slightly racy, and combine it with the code seen above.

ThomasWaldmann: solved, see commits

ThomasWaldmann:
@enkore @textshell can you review please?

ThomasWaldmann:
BTW, I thought about using a random SUFFIX for the tmp file (tempfile stdlib module style or so).

But guess that would lose the nice .tmp file cleanup property of the current code (as the current code always kills the always-same-suffix .tmp file first, if any is present).

hexagonrecursion:
I can backport this. Should I wait until this is merged?

ThomasWaldmann:
Yes, please wait until merged. Thanks!

enkore (approved):

ThomasWaldmann:
Thanks @enkore for having a look. Ready for backporting to 1.1-maint now.

hexagonrecursion:
Sorry. Life. Busy. I don't know when I will be able to work on backporting this.

PR #6058, opened 2021-11-18
Backport of #6022: Speed up search for next valid object in segment in --repair mode

pgerber:
Backport of #6022 as requested by @ThomasWaldmann.


ThomasWaldmann:
Thanks!

Issue #5680, opened 2021-02-09
Ownership and permissions of parent directory differ from source if archive is mounted or extracted


Woi:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Question, maybe a bug

#### Your borg version (borg -V).
1.1.15

#### Operating system (distribution) and version.
Fedora 32

#### Hardware / network configuration, and filesystems used.
No network, local filesystem only: btrfs, fuse, tmpfs

#### How much data is handled by borg?
~1GB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

Let's check ownership and file permissions  of home directories:
```
ls -l /home/
drwx------. 62 user1 user1  4096  9. Feb 21:15 user1
drwx------. 16 user2 user2  4096  8. Feb 22:19 user2
```

As root:
Create a repository and backup /home/user2 and /home/user1/subdir
```
borg init --encryption=repokey-blake2 --storage-quota=10G --make-parent-dirs /tmp/borgtest/{hostname}
borg create --progress --stats  /tmp/borgtest/{hostname}::{now} /home/user1/subdir /home/user2
```

Mount the backup to allow users to restore their own files:
```
mkdir /tmp/borgtest/mount
borg mount -o allow_other /tmp/borgtest/{hostname} /tmp/borgtest/mount
chmod 755 /tmp/borgtest
```

As user1:
```
cd /tmp/borgtest/mount/2021-02-09T20:46:05/home
ls -l
drwxr-xr-x. 1 root  root  0  9. Feb 20:47 user1
drwx------. 1 user2 user2 0  8. Feb 22:19 user2
```
Homedirectory of user1 is owned by root and permissions are set to 755!?

As root again:
Let's check what borg extract does, to see if this behaviour is fuse specific:
```
borg umount /tmp/borgtest/mount
mkdir /tmp/borgtest/restore
cd /tmp/borgtest/restore/
borg extract --progress --strip-components=1 /tmp/borgtest/{hostname}::2021-02-09T20:46:05
ls -l
drwx------.  3 root  root   60  9. Feb 20:51 user1
drwx------. 16 user2 user2 600  8. Feb 22:19 user2
```
Ownership is still root/root, but permissions looks okay. So this differs from source and borg mount.

## Describe the problem you're observing.
Directory ownership differs from source if only a subdirectory is backed up. 
[...]

Issue #1751, opened 2016-10-21
Directory-permissions broken after extraction


AxelRHD:
Hello Borg-Team,

I have problems with the extract-function. Had to backup with sudo because of file permissions (f. e. backing up the /usr/local/bin-folder). The directory-permissions changed from `drwx r-x r-x` to `drwx --- ---`. I used the command `sudo borg extract repo::archive usr/local/bin/file` to restore a single file.

Is there something wrong in my usage of the extraction-tool or is this a bug?

ThomasWaldmann:
that's a known issue and we already have a ticket for it.

ThomasWaldmann:
hmm, or maybe not?

Are you saying you restored directly to the existing /usr/local/bin directory?
== you were root, and cwd was / when invoking the borg command?

AxelRHD:
No, I tried to restore the file to another directory. I got from this destination the complete sub-directories usr/local/bin/file (for example in my home-directory).

ThomasWaldmann:
OK, then see the attic issue I mentioned. The point is you told borg to only extract one file (not directories), so it had only the data of that one file. It created the directories without having data about them, except the names (which are stored with the file).

ThomasWaldmann:
I reopened this to locally track it, so users can better find it. Will check the attic issue from the list instead.

SanskritFritz:
> It created the directories on-the-fly without having metadata for them, except the names (which are stored with the file).

How about having an option to istruct Borg not to recreate the path but only restore the files where the permissions are known? This might be important because if I restrore a single file for another user into their home directory, they still would not see their file since the path would have been created with root:root 700 permissions.

ThomasWaldmann:
@SanskritFritz Imagine you backup /home/foo/x and /home/bar/x and not extract the whole path, then you'ld get a x vs x collision.

SanskritFritz:
Yes, that is why I thought it should be optional.

ygoe:
Wait, borg backup cannot properly restore directory permissions for 3 years now? With no signs of activity towards actually fixing it? How come so many are recommending this software for whole-system backup? I haven't tested restoring a complete system yet, just started with a Postgres database. All went quickly but the directory permissions are unusable. I have to manually check them down to some levels. This will be a nightmare if it happens to a complete Linux system!

SanskritFritz:
@ygoe If you backup a whole linux system, you will likely backup it from the root dir. Then all directories will be backed up correctly with permissions and etc.
[...]

PR #6062, opened 2021-11-25
docs: excluded parent dir's metadata can't restore.

cyfdecyf:
Refer to discussion in #5680.

### DIFF
--- a/docs/usage/extract.rst.inc
+++ b/docs/usage/extract.rst.inc
@@ -121,5 +121,5 @@ pass over the archive metadata.
     so make sure you ``cd`` to the right place before calling ``borg extract``.
 
     When parent directories are not extracted (because of using file/directory selection
-    or any other reasons), borg can not restore those directories' metadata, e.g. owner,
+    or any other reason), borg can not restore parent directories' metadata, e.g. owner,
     group, permission, etc.
--- a/docs/usage/extract.rst.inc
+++ b/docs/usage/extract.rst.inc
@@ -119,3 +119,7 @@ pass over the archive metadata.
 
     Currently, extract always writes into the current working directory ("."),
     so make sure you ``cd`` to the right place before calling ``borg extract``.
+
+    When using file and directory selection, if the parent directories of selected
+    targets are not selected, borg can not restore those directories' metadata, e.g.
+    owner, group, permission, etc.
--- a/docs/usage/extract.rst.inc
+++ b/docs/usage/extract.rst.inc
@@ -120,6 +120,6 @@ pass over the archive metadata.
     Currently, extract always writes into the current working directory ("."),
     so make sure you ``cd`` to the right place before calling ``borg extract``.
 
-    When using file and directory selection, if the parent directories of selected
-    targets are not selected, borg can not restore those directories' metadata, e.g.
-    owner, group, permission, etc.
+    When parent directories are not extracted (because of using file/directory selection
+    or any other reasons), borg can not restore those directories' metadata, e.g. owner,
+    group, permission, etc.
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -170,7 +170,7 @@ Examples::
     .. note::
 
         It's possible that a sub-directory/file is matched while parent directories are not.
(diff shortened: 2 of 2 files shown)


docs/usage/help.rst.inc
```diff
@@ -167,6 +167,16 @@ Examples::
...
+    .. note::
+
+        It's possible that a sub-directory/file is matched while parent directories are not.
+        In that case, parent directories are not backed up thus their user, group, permission,=
+        etc. can not be restored.
+
+        For example, with `--pattern=+/home/user/subdir --pattern=-/home/user/*`,
+        `/home/user/subdir` can be restored with correct metadata, while `\home`and
+        `/home/user` can not.
+
```
ThomasWaldmann, line 179: guess that is a rather special scenario.

the more general scenario is that a deep fs object is extracted while the parent dir(s) are not extracted (either because they were not chosen to get extracted or they are not even present in the archive because they were not backed up).

and borg just can't set metadata for stuff it does not extract from an archive entry, because it does not know that metadata in that case.

cyfdecyf: While googling about this directory metadata not get restored problem, I saw issue #1751. I'm perfectly fine with current behavior, just guess it's better to mention this in the doc.

Maybe mention this also in `borg extract` doc?

ThomasWaldmann: yeah, i agree it should be in the docs. borg extract looks like a good place.

cyfdecyf: Added some comments for the behavior in `borg extract` doc. I'm not native English speaker, please feel free to change the description if it's not clear.

ThomasWaldmann (changes requested):
crap, i forgot to submit this review when i made it a while ago.. X-)

so, here it is now.

docs/usage/extract.rst.inc
```diff
@@ -118,4 +118,8 @@ pass over the archive metadata.
 .. note::
 
     Currently, extract always writes into the current working directory ("."),
-    so make sure you ``cd`` to the right place before calling ``borg extract``.
\ No newline at end of file
+    so make sure you ``cd`` to the right place before calling ``borg extract``.
+
+    When using file and directory selection, if the parent directories of selected
+    targets are not selected, borg can not restore those directories' metadata, e.g.
+    owner, group, permission, etc.
```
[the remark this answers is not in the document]
cyfdecyf: Just pushed my change as suggested.

docs/usage/help.rst.inc
```diff
@@ -167,6 +167,16 @@ Examples::
...
 
+    .. note::
+
+        It's possible that a sub-directory/file is matched while parent directories are not.
+        In that case, parent directories are not backed up thus their user, group, permission,
+        etc. can not be restored.
+
+        For example, with `--pattern=+/home/user/subdir --pattern=-/home/user/*`,
+        `/home/user/subdir` can be restored with correct metadata, while `\home`and
+        `/home/user` can not.
```
ThomasWaldmann, line 178: can this be simplified?

also: no backslashes.

cyfdecyf: I didn't get that we don't need to mention the special case in parent docs in another comment. Just removed this example to avoid complicating the doc.

ThomasWaldmann: ok

ThomasWaldmann:
Thanks! Could you backport this to 1.1-maint branch?

hexagonrecursion:
Warning: you made the change in the wrong place: docs/usage/extract.rst.inc and docs/usage/help.rst.inc are automatically generated from the builtin help:

https://github.com/borgbackup/borg/blob/74c087c76a894700236ea36d232df3853150ff8a/docs/usage/extract.rst.inc#L1

I'll fix this.

cyfdecyf:
> Warning: you made the change in the wrong place: docs/usage/extract.rst.inc and docs/usage/help.rst.inc are automatically generated from the builtin help:
> 
> https://github.com/borgbackup/borg/blob/74c087c76a894700236ea36d232df3853150ff8a/docs/usage/extract.rst.inc#L1
> 
> I'll fix this.

Thanks, didn't notice it's auto generated.

Issue #6063, opened 2021-11-26
xattrs not supported warning during extract fails due to % in filename


petenewcomb:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).

1.1.15

#### Operating system (distribution) and version.

Ubuntu 20.04

#### Hardware / network configuration, and filesystems used.

Local repo, Ext4

#### How much data is handled by borg?

4.6TiB of Borg repositories total, archive in question is 776GiB but extract fails at 57GiB due to this bug.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

borg extract /backups/oldeyelow/repo::fluffy-2020-08-22T23:13:37.716949

## Describe the problem you're observing.

Restoring a file with `%p` in the name that includes xattrs (from Windows Subsystem for Linux) onto an ext4 partition causes an exception because the filename gets expanded into the LHS of the `%` formatting operator.  Excerpted log and full stack trace below.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes, same results every time I attempted to extract the archive.  I tried using `--errors` to suppress, but the faulty logic still gets executed.  The version I'm using doesn't yet support `--noxattrs` as now exists in the `1.1-maint` branch, so that wasn't an option.  I therefore edited `/usr/lib/python3/dist-packages/borg/archive.py` to change the logic so that the LHS is always a constant string.  This allowed the archive extract completely while still generating the intended warnings.

#### Include any warning/errors/backtraces from the system logs

```
[... many more "xattrs not supported" warnings like the following line ...]
/data/oldbackups/flufftest/mnt/c/Users/isaac/AppData/Roaming/Mozilla/Firefox/Profiles/2416qrmh.default-release/storage/default/https+++mail.google.com/idb/953658429glmaaviyle-ks-w.files: when setting extended attribute system.wsl_case_sensitive: xattrs not supported on this filesystem
Local Exception
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/borg/archive.py", line 771, in restore_attrs
    xattr.setxattr(fd or path, k, v, follow_symlinks=False)
  File "/usr/lib/python3/dist-packages/borg/xattr.py", line 249, in setxattr
    _setxattr_inner(func, path, name, value)
[...]

PR #6064, opened 2021-11-26
avoid expanding path into LHS of formatting operation, see #6063

petenewcomb:
Submitted on `1.1-maint` because the code in question has been moved, removed, or refactored on `master`.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -792,7 +792,7 @@ Utilization of max. archive size: {csize_max:.0%}
                         # EACCES: permission denied to set this specific xattr (this may happen related to security.* keys)
                         # EPERM: operation not permitted
                         err_str = os.strerror(e.errno)
-                    logger.warning('%s: when setting extended attribute %s: %s' % (path, k.decode(), err_str))
+                    logger.warning('%s: when setting extended attribute %s: %s', path, k.decode(), err_str)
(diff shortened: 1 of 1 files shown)


src/borg/archive.py
```diff
@@ -793,7 +792,7 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
                         # EACCES: permission denied to set this specific xattr (this may happen related to security.* keys)
                         # EPERM: operation not permitted
                         err_str = os.strerror(e.errno)
-                    logger.warning(msg_format % err_str)
+                    logger.warning('%s: when setting extended attribute %s: %s' % (path, k.decode(), err_str))
```
ThomasWaldmann, line 795: thanks for discovering and fixing this.

the above code is a bit better, but still problematic, because the first arg of `logger.warning` is interpreted as a format string, see:

https://docs.python.org/3/library/logging.html#logging.Logger.debug

ThomasWaldmann: so guess it needs to be:
```
logger.warning(format, path, k.decode(), err_str)
```

ThomasWaldmann: and also guess we need to look for other such places in the code.

ThomasWaldmann:
Thanks!

hexagonrecursion:
I'll look into forwardporting this

PR #6071, opened 2021-11-30
Backport #6023: do not show archive name in error msgs referring to repository

hexagonrecursion:
Backport #6023 (do not show archive name in error msgs referring to repository)

The original PR did not have a test to prove that the bug is fixed so I took the liberty of adding one. After this is merged I'll forward port the test to master.

PR #6076, opened 2021-12-02
Forward port: Test that the archive name is not mentioned

hexagonrecursion:
This test was originally added in #6071


ThomasWaldmann:
Thanks!

PR #6079, opened 2021-12-07
Remove --list-format from borg list

hexagonrecursion:
While reading the docs I noticed that in `borg list` the options --list-format and --format do the same thing. Using `git log -S` I have uncovered that --list-format used to be deprecated and was supposed to be removed in c87393cab736d30a3c138f5f694c8cc0c8ca6ccc, but you overlooked it and undeprecated it instead. *What should we do now? Just remove it or deprecate it again?*


ThomasWaldmann:
Good catch! Thanks!

PR #6085, opened 2021-12-10
Fix exceptions thrown by `setup.py build_man`

hexagonrecursion:
While doing some doc updates I needed a way to test them - to build
the documentation and inspect the output. I ran into an issue:
running python setup.py build_man was throwing exceptions:

1. The import-tar parser had a None description causing:

    File "/home/user/borg/setup_docs.py", line 451, in write_heading
      write(char * len(header))
    TypeError: object of type 'NoneType' has no len()

2. There was no docs/usage/import-tar.rst causing an exception too

### DIFF
--- a/docs/usage/import-tar.rst
+++ b/docs/usage/import-tar.rst
@@ -1,0 +1,1 @@
+.. include:: import-tar.rst.inc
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Thanks for fixing. Strange, can't remember noticing this bug when doing the last beta releases!?

PR #6093, opened 2021-12-16
key export: print key if path is '-' or not given, fixes #6092

### LINKED ISSUE #6092: Confusing CLI message "output file to export key to expected"

Yes

ISSUE

borg 1.1.17

Pop!_OS 21.10

None yet.

None yet.

`borg key export /media/duane/JOHNSONVID2/backup.borg`

The error message is confusing:
`output file to export key to expected`

Something like this would have been clearer to me:
`please specify the destination file where you'd like the key to be exported`

(Nice to have): In addition, it would be nice if the absence of a file just dumped the key to stdout.

grahamstockton:
Implemented #6092. "borg key export [REPOSITORY]" now prints out the key

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -364,8 +364,9 @@ class Archiver:
             manager.export_paperkey(args.path)
         else:
             if not args.path:
-                self.print_error("output file to export key to expected")
-                return EXIT_ERROR
+                print("No output path provided. Printing key:")
+                manager.export_paperkey(None)
+                return EXIT_SUCCESS
             try:
                 if args.qr:
                     manager.export_qr(args.path)
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -364,8 +364,7 @@ class Archiver:
             manager.export_paperkey(args.path)
         else:
             if not args.path:
-                print("No output path provided. Printing key:")
-                manager.export_paperkey(None)
+                print(manager.get_keyfile_data())
                 return EXIT_SUCCESS
             try:
                 if args.qr:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -363,9 +363,6 @@ class Archiver:
         if args.paper:
             manager.export_paperkey(args.path)
         else:
-            if not args.path:
-                print(manager.get_keyfile_data())
-                return EXIT_SUCCESS
             try:
                 if args.qr:
                     manager.export_qr(args.path)
--- a/src/borg/crypto/keymanager.py
+++ b/src/borg/crypto/keymanager.py
@@ -75,13 +75,19 @@ class KeyManager:
         return data
 
     def store_keyfile(self, target):
-        with open(target, 'w') as fd:
+        with dash_open(target, 'w') as fd:
             fd.write(self.get_keyfile_data())
 
     def export(self, path):
+        if path is None:
+            path = '-'
+
         self.store_keyfile(path)
 
     def export_qr(self, path):
+        if path is None:
+            path = '-'
+
         with open(path, 'wb') as fd:
             key_data = self.get_keyfile_data()
             html = pkgutil.get_data('borg', 'paperkey.html')
@@ -89,6 +95,9 @@ class KeyManager:
             fd.write(html)
 
     def export_paperkey(self, path):
+        if path is None:
+            path = '-'
+
         def grouped(s):
             ret = ''
             i = 0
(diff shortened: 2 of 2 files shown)


grahamstockton:
Let me know if this is what was intended. I can fix/do more if necessary.

src/borg/archiver.py
```diff
@@ -364,8 +364,9 @@ def do_key_export(self, args, repository):
             manager.export_paperkey(args.path)
         else:
             if not args.path:
-                self.print_error("output file to export key to expected")
-                return EXIT_ERROR
+                print("No output path provided. Printing key:")
+                manager.export_paperkey(None)
```
ThomasWaldmann, line 368: maybe it would be even better without that "print", because then the stdout output is exactly the same as borg would write to the on-disk file when a file path has been provided.

ThomasWaldmann: But there is another big issue:

364 is dealing with the `--paper` part and is happy to work with args.path having any value (including the default, which is None I guess?). So that functionality is already implemented there (with or without a path).

366+ block is the "not paper" branch, so I guess it must not call `export_paperkey` there, but `export_qr` or `export` as seen in 370+. So I guess there 2 have to be chacked whether they can somehow work without a on-disk file as output (but stdout).

ThomasWaldmann (changes requested):

ThomasWaldmann:
@grahamstockton did you see the feedback?

grahamstockton:
Happy holidays. Yes, I am working on it now and will be available to work more frequently in the future. I just changed it to call the KeyManager.get_keyfile_data() method, the same one that is called internally in export. Let me know what you think.

grahamstockton:
Both the export and export_qr required a path. Options other than what I did just now include adding an if statement to the export function for when there is no path given, or creating a new method of KeyManager that just wraps the print statement that I have in de470ef.

src/borg/archiver.py
```diff
@@ -364,8 +364,8 @@ def do_key_export(self, args, repository):
             manager.export_paperkey(args.path)
         else:
             if not args.path:
-                self.print_error("output file to export key to expected")
-                return EXIT_ERROR
+                print(manager.get_keyfile_data())
+                return EXIT_SUCCESS
```
ThomasWaldmann, line 368: we are getting closer, but still, we are not there yet...

now we have the strange situation that the code in 367/368 implements just another behaviour (based on absence of a path, but it does not care about whether qr.args is given or not) in parallel to the behaviour in 370..373 (which is different based on whether args.qr is given or not).

i really think the block in 366..368 needs to get completely removed.

and as a replacement, `KeyManager.export` and `.export_qr` need to be made tolerant of getting path=None or path='-' and just use sys.stdout in that case.

ThomasWaldmann: so, `store_keyfile` and `export_qr` could just call `dash_open` instead of plain `open`.

with that, everything left to do is to use `'-'` (the dash) in case args.path is None (which can be also handled right at the beginning of `export` and `export_qr`).

export_paperkey should then also get a little refactored and the `if path` block at its end being replaced by `dash_open` also.

ThomasWaldmann: with that, we get:
- consistent behaviour no matter which options are given
- with no path or with path == '-', it uses stdout
- less code, reusing the dash_open code

ThomasWaldmann (changes requested):

grahamstockton:
Hopefully this is what you had in mind. Let me know what you think.

src/borg/crypto/keymanager.py
```diff
@@ -75,20 +75,29 @@ def get_keyfile_data(self):
...
+        if path is None:
+            path = '-'
+
         self.store_keyfile(path)
 
     def export_qr(self, path):
+        if path is None:
+            path = '-'
+
         with open(path, 'wb') as fd:
```
ThomasWaldmann, line 91: dash_open?

ThomasWaldmann: other than that, yeah, that's what i meant. guess creates consistent behaviour.

ThomasWaldmann (approved):
LGTM! Thanks!

grahamstockton:
Great! Thank you for the pointers, I really appreciate them.

ThomasWaldmann:
If we want this in 1.1.x releases also, it could be backported to the 1.1-maint branch.

PR #6096, opened 2021-12-24
Workaround for volume shadow copy in WSL1

### LINKED ISSUE #6024: EROFS due to O_NOATIME with volume shadow copy in WSL1

cuevavirus:
Fixes #6024 for master. Similar to patch #6027 for maint-1.1 submitted previously.

### DIFF
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -286,6 +286,13 @@ def os_open(*, flags, path=None, parent_fd=None, name=None, noatime=False):
                 raise
             # Was this EPERM due to the O_NOATIME flag? Try again without it:
             fd = os.open(fname, _flags_normal, dir_fd=parent_fd)
+        except OSError as exc:
+            # O_NOATIME causes EROFS when accessing a volume shadow copy in WSL1
+            from . import workarounds
+            if 'retry_erofs' in workarounds and exc.errno == errno.EROFS and _flags_noatime != _flags_normal:
+                fd = os.open(fname, _flags_normal, dir_fd=parent_fd)
+            else:
+                raise
     else:
         fd = os.open(fname, _flags_normal, dir_fd=parent_fd)
     return fd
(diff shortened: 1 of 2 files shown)


src/borg/helpers/fs.py
```diff
@@ -286,6 +286,13 @@ def os_open(*, flags, path=None, parent_fd=None, name=None, noatime=False):
                 raise
             # Was this EPERM due to the O_NOATIME flag? Try again without it:
             fd = os.open(fname, _flags_normal, dir_fd=parent_fd)
+        except OSError as exc:
+            # O_NOATIME causes EROFS when accessing a volume shadow copy in WSL1
+            from . import workarounds
```
ThomasWaldmann, line 291: did you try to do this import on global level instead of here?

cuevavirus: Yes in that case import fails due to circular import.

https://github.com/borgbackup/borg/blob/bf392367f159fd81494a9d76bb646e7f3ab37132/src/borg/helpers/__init__.py#L12

ThomasWaldmann: ok, i had a similar effect recently, too.

ThomasWaldmann:
Thanks!

PR #6108, opened 2022-01-05
list: Add command_line to format keys

sophie-h:
Analogous to #4184

Do I have to rebuild docs manually?

### DIFF
--- a/scripts/shell_completions/zsh/_borg
+++ b/scripts/shell_completions/zsh/_borg
@@ -1148,7 +1148,7 @@ _borg_format_keys() {
   local repo_or_arch=${(Q)1}
 
   local -a keys=( NEWLINE NL NUL SPACE TAB CR LF )
-  local -a repository_keys=( archive name barchive comment bcomment id start time end hostname username )
+  local -a repository_keys=( archive name barchive comment bcomment id start time end command_line hostname username )
   local -a archive_keys=( type mode uid gid user group path bpath source linktarget flags size csize dsize dcsize
     num_chunks unique_chunks mtime ctime atime isomtime isoctime isoatime blake2b blake2s md5 sha1 sha224 sha256 sha384
     sha3_224 sha3_256 sha3_384 sha3_512 sha512 shake_128 shake_256 archiveid archivename extra health )
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -4,6 +4,7 @@ import json
 import os
 import os.path
 import re
+import shlex
 import socket
 import stat
 import uuid
@@ -567,13 +568,14 @@ class ArchiveFormatter(BaseFormatter):
         'start': 'time (start) of creation of the archive',
         'time': 'alias of "start"',
         'end': 'time (end) of creation of the archive',
+        'command_line': 'command line which was used to create the archive',
         'id': 'internal ID of the archive',
         'hostname': 'hostname of host on which this archive was created',
         'username': 'username of user who created this archive',
     }
     KEY_GROUPS = (
         ('archive', 'name', 'barchive', 'comment', 'bcomment', 'id'),
-        ('start', 'time', 'end'),
+        ('start', 'time', 'end', 'command_line'),
         ('hostname', 'username'),
     )
 
@@ -624,6 +626,7 @@ class ArchiveFormatter(BaseFormatter):
             'comment': partial(self.get_meta, 'comment', rs=True),
             'bcomment': partial(self.get_meta, 'comment', rs=False),
             'end': self.get_ts_end,
+            'command_line': self.get_cmdline,
         }
         self.used_call_keys = set(self.call_keys) & self.format_keys
         if self.json:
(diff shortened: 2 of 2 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -624,6 +625,7 @@ def __init__(self, format, repository, manifest, key, *, json=False, iec=False):
             'comment': partial(self.get_meta, 'comment', rs=True),
             'bcomment': partial(self.get_meta, 'comment', rs=False),
             'end': self.get_ts_end,
+            'command_line': partial(self.get_meta, 'cmdline', rs=False),
```
ThomasWaldmann, line 628: maybe just stick with `cmdline` to keep the names in sync?

also, rs=False should be rather `bcmdline`, while `cmdline` would use rs=True.

scripts/shell_completions/zsh/_borg
```diff
@@ -1148,7 +1148,7 @@ _borg_format_keys() {
   local repo_or_arch=${(Q)1}
 
   local -a keys=( NEWLINE NL NUL SPACE TAB CR LF )
-  local -a repository_keys=( archive name barchive comment bcomment id start time end hostname username )
+  local -a repository_keys=( archive name barchive comment bcomment id start time end command_line hostname username )
```
ThomasWaldmann, line 1151: ok, so you fixed zsh - are there other completions to fix also?

ThomasWaldmann (commented):
thanks for the PR!

maybe check that IF one rebuilds the docs, documentation about this new stuff will be present (not sure, but guess we might auto-generate these docs).

but please do NOT add/commit stuff that is auto-generated, that will happen at release time anyway.

scripts/shell_completions/zsh/_borg
sophie-h: That was not manual. It was some magic and I trusted it.

src/borg/helpers/parseformat.py
sophie-h: The naming is in sync with the existing `borg info` field. So maybe it's more consistent this way?

The output of `info` might not be binary though.

ThomasWaldmann: The borg info output is rather a textual description (like "Command line" or "Archive name"), not the short identifier style like used for the placeholders.

So please use cmdline and/or bcmdline here (with the corresponding rs param).

scripts/shell_completions/zsh/_borg
ThomasWaldmann: Oh? OK...

src/borg/helpers/parseformat.py
sophie-h: > The borg info output is rather a textual description

I meant the `--json` output key which is `command_line` for `info`. So far `info` and `list` JSON keys have been in sync.

sophie-h: Looks like `borg info --json` outputs a `command_line` key without calling `remove_surrogates` on it. That's only done for the [none `--json` output.](https://github.com/borgbackup/borg/blob/master/src/borg/archiver.py#L1410)

Renaming the `borg info` key is probably not an option because of backward compatibility? Should I just add `command_line` and `bcommand_line` to both `list` and `info` and make the slight backward-incompatible change to map `remove_surrogates` over `command_line` in `borg info --json`?

src/borg/helpers/parseformat.py
```diff
@@ -662,7 +664,13 @@ def archive(self):
 
     def get_meta(self, key, rs):
         value = self.archive.metadata.get(key, '')
-        return remove_surrogates(value) if rs else value
+        if rs:
+            if isinstance(value, list):
+                return list(map(remove_surrogates, value))
```
ThomasWaldmann, line 669: doesn't the code expect a str rather than a list?

sophie-h: I think `json.dumps` should get the list while we want a neat string without JSON. It's implemented now.

src/borg/helpers/parseformat.py
```diff
@@ -662,14 +665,26 @@ def archive(self):
 
     def get_meta(self, key, rs):
         value = self.archive.metadata.get(key, '')
-        return remove_surrogates(value) if rs else value
+        if rs:
+            if isinstance(value, list):
+                return self.format_cmdline(map(remove_surrogates, value))
```
ThomasWaldmann, line 670: this assumes that every list is a command line / has to get treated like a command line.

true now, but likely not forever / in general.

src/borg/helpers/parseformat.py
```diff
@@ -662,14 +665,26 @@ def archive(self):
...
+            else:
+                return remove_surrogates(value)
+        else:
+            return value
```
ThomasWaldmann, line 674: also, the else branch just passes through the list.

works now, but not in general if we want to support lists.

src/borg/helpers/parseformat.py
sophie-h: And `get_meta` defaults to an empty string if the key is not available. All bad. New attempt.

ThomasWaldmann:
OK, looks good. Guess there are no tests for this, did you try it manually?

sophie-h:
> OK, looks good. Guess there are no tests for this, did you try it manually?

Yes, tried it manually

src/borg/helpers/parseformat.py
ThomasWaldmann: done.

ThomasWaldmann:
Guess we'ld need a backport to 1.1-maint - would you like to do that?

PR #6119, opened 2022-01-08
repository: quota computation fixes for delayed compaction

ThomasWaldmann:
see commit comments.

for master branch, these fixes are required, because we have delayed compaction (separate `borg compact` command).

for 1.1-maint, they could be also applied, but are not that important because compact_segments is always called after writing.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -806,6 +806,7 @@ class Repository:
                         # do not remove entry with empty shadowed_segments list here,
                         # it is needed for shadowed_put_exists code (see below)!
                         pass
+                    self.storage_quota_use -= len(data) + self.io.put_header_fmt.size
                 elif tag == TAG_DELETE and not in_index:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
@enkore @textshell can you have a look?

src/borg/repository.py
```diff
@@ -806,6 +806,7 @@ def complete_xfer(intermediate=True):
                         # do not remove entry with empty shadowed_segments list here,
                         # it is needed for shadowed_put_exists code (see below)!
                         pass
+                    self.storage_quota_use -= len(data) + self.io.put_header_fmt.size
```
ThomasWaldmann, line 809: this is in `compact_segments()`, where space is really freed.

enkore (approved):

ThomasWaldmann:
@enkore thanks!

Issue #6099, opened 2021-12-26
test_fuse on macOS hangs


ThomasWaldmann:
Since some time, the macOS CI run via github actions usually gets killed before it is finished.

ThomasWaldmann:
Overall setup until execution of tests starts takes about 3-4 minutes so guess that is not the problem.

But whatever comes after this seems to take long or blocks:
```
[gw0] [  8%] PASSED .tox/py38-fuse2/lib/python3.8/site-packages/borg/testsuite/archiver.py::ArchiverTestCase::test_file_status_rc_cache_mode 
```

ThomasWaldmann:
That comes next: borg/testsuite/archiver.py::ArchiverTestCase::test_fuse 

So we have a FUSE issue here...

ThomasWaldmann:
From the log:
```
`osxfuse` has been succeeded by `macfuse` as of version 4.0.0.
To update to a newer version, do:
  brew uninstall osxfuse
  brew install macfuse

osxfuse requires a kernel extension to work.

If the installation fails, retry after you enable it in:

  System Preferences → Security & Privacy → General

For more information, refer to vendor documentation or this Apple Technical Note:

  https://developer.apple.com/library/content/technotes/tn2459/_index.html

You must reboot for the installation of osxfuse to take effect.
```

ThomasWaldmann:
Tried more recent macfuse instead of older osxfuse, but still hangs at same spot.

ThomasWaldmann:
Well, that PR rather was a workaround than a fix.

It still has to be investigated what's up with FUSE / test_fuse.

ThomasWaldmann:
Note: fuse tests on macos sierra 10.12.6 vagrant vm (master branch) work.

ThomasWaldmann:
Maybe "full disk access" permission missing?

ThomasWaldmann:
On my laptop: giving full disk access (to pycharm in this case) fixed the issue.

strangely, there are now some "python quit unexpectedly" popups when i run `pytest -k test_fuse`.

ThomasWaldmann:
See comments in #6196, closing.

PR #6125, opened 2022-01-16
CI: downgrade to macOS 10.15, work around #6099

ThomasWaldmann:


.github/workflows/ci.yml
```diff
@@ -107,7 +107,7 @@ jobs:
         brew install zstd || brew upgrade zstd
         brew install lz4 || brew upgrade lz4
         brew install openssl@1.1 || brew upgrade openssl@1.1
-        #brew install homebrew/cask/osxfuse || brew upgrade homebrew/cask/osxfuse  # Required for Python llfuse module
+        brew install --cask macfuse || brew upgrade --cask macfuse  # Required for Python llfuse module
```
hexagonrecursion, line 110: I do not fully understand this patch. Why did you switch from osxfuse to macfuse?

ThomasWaldmann: strictly taken, the switch was not required, it also worked with osxfuse after the macos downgrade.

but considering a later potential upgrade of macos to some more recent version again (after the hanging issue is analyzed and fixed), i wanted to switch to the more recent macfuse code to be prepared for that.

hexagonrecursion:
Hello. You forgot to remove the "backport" label

PR #6132, opened 2022-01-18
clarify who starts the remote agent

Tavlor:
It was unclear that the user _only_ needs to have borg installed on a remote system to use client/server mode. Hopefully this change makes it apparent that the user doesn't start anything on the remote system themselves.
If the language is awkward or even more unclear, I'm happy to make changes


ThomasWaldmann:
Thanks! Can you do the backport to 1.1-maint also?

PR #6150, opened 2022-01-22
vagrant fixes (master)

ThomasWaldmann:


ThomasWaldmann:
backporting note for 1.1-maint branch:
- no clean2 there
- no such or different end-of-file comment about python versions there

Issue #28, opened 2015-05-22
use py.test for testing, refactor tests


ThomasWaldmann:
pytest is way superior (and also more comfortable) compared to unittest.

less work to write tests, prettier / easier readable tests, more useful output (esp. on failure).

the existing tests can be taken "as is" as a starting point, pytest is able to run them.

later, more pytest specific features can be used to further improve.

**be careful!**

`borg.selftest` is special because it must not depend on pytest. it runs every time `borg` is invoked by a user.

see there: https://github.com/borgbackup/borg/pull/6157#issuecomment-1024996349

These must **not** be converted to `unittest`:
```
from .testsuite.hashindex import HashIndexDataTestCase, HashIndexRefcountingTestCase, HashIndexTestCase
from .testsuite.crypto import CryptoTestCase
from .testsuite.chunker import ChunkerTestCase

SELFTEST_CASES = [
    HashIndexDataTestCase,
    HashIndexRefcountingTestCase,
    HashIndexTestCase,
    CryptoTestCase,
    ChunkerTestCase,
]
```

ThomasWaldmann:
I made the necessary changes so py.test is used via tox (local testing) and travis ci.

Keeping this issue open as a reminder to actually migrate the tests to "py.test style" - but that should be done a bit later as currently too much unmerged code is still waiting to get merged and global changes would be a pain for that...

hexagonrecursion:
You currently have a hierarchy of test classes with unittest.TestCase as the root. Several very nice features of pytest are unavailable in those tests. Once that stings especially badly is pytest.mark.parametrize.

https://docs.pytest.org/en/6.2.x/unittest.html?highlight=unittest#pytest-features-in-unittest-testcase-subclasses

I am going to start slowly purging the traces of unittest from your codebase.

ThomasWaldmann:
There's progress: all archiver tests were recently migrated to pytest! Thanks to @bigtedde!

ThomasWaldmann:
The helpers tests also use pytest now, so this is finished.

PR #6157, opened 2022-01-22
Refactor: remove assert_true (master)

hexagonrecursion:
Work toward https://github.com/borgbackup/borg/issues/28

This PR removes the usage of assert_true in the master branch

### DIFF
--- a/src/borg/testsuite/__init__.py
+++ b/src/borg/testsuite/__init__.py
@@ -164,7 +164,6 @@ class BaseTestCase(unittest.TestCase):
     assert_not_in = unittest.TestCase.assertNotIn
     assert_equal = unittest.TestCase.assertEqual
     assert_not_equal = unittest.TestCase.assertNotEqual
-    assert_true = unittest.TestCase.assertTrue
 
     if raises:
         assert_raises = staticmethod(raises)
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
did you see the comment in `src/borg/testsuite/__init__.py` line 24?

some of these tests are run as "borg selftest" every time borg is invoked, so they really run with the stdlib `unittest` to avoid a dependency on pytest. not sure how this change affects that, maybe better try it out by triggering some test failure in the self tests.

ThomasWaldmann:
Update: tried that myself:

E.g. `class HashIndexDataTestCase(BaseTestCase)` is part of `borg.selftest` and (before your changes) already uses a plain `assert` instead of `self.assert_true()`.

I uninstalled pytest to avoid any interference (it does a lot of magic, including with assert), then ran `borg list repo` after artifically tweaking one of the test methods there to fail the assertion.

Well, it worked and the output was practically almost identical (except the missing "False is not true", which is not helpful anyway).

Thus, we can merge your PR after review.

ThomasWaldmann:
Thanks!

hexagonrecursion:
Theoretically, assert can be a no-op:
https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

PR #6178, opened 2022-01-25
Backport upgrade pythons (#6116)

KN4CK3R:
Backport of #6116.

I did not port the lines from
https://github.com/borgbackup/borg/pull/6116/files#diff-1ea02434f746b6d1522cdef3d4a56e57cf2d5a72c5273927746a71eaf06d74fdR182-R183
because this branch still uses 3.7

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -220,7 +220,7 @@ def install_pythons(boxname)
     pyenv install 3.7.11  # binary build, use latest 3.7.x release
     pyenv install 3.8.0  # tests
     pyenv install 3.9.0  # tests
-    pyenv install 3.10-dev  # tests
+    pyenv install 3.10.0  # tests
     pyenv rehash
   EOF
 end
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -300,9 +300,9 @@ def run_tests(boxname)
     . ../borg-env/bin/activate
     if which pyenv 2> /dev/null; then
       # for testing, use the earliest point releases of the supported python versions.
-      # on some dists, 3.10-dev does not compile, so if pyenv fails due to this, try without 3.10.
-      pyenv global 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0 3.10-dev || pyenv global 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0
-      pyenv local 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0 3.10-dev || pyenv local 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0
+      # on some dists, 3.10 does not compile, so if pyenv fails due to this, try without 3.10.
+      pyenv global 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0 3.10.0 || pyenv global 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0
+      pyenv local 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0 3.10.0 || pyenv local 3.5.3 3.6.2 3.7.11 3.8.0 3.9.0
```
ThomasWaldmann, line 305: you missed the 3.9.9 change here.

ThomasWaldmann: ehrm, no, sorry, wrong comment.

3.9.0 should not be changed in this branch.

Vagrantfile
```diff
@@ -219,8 +219,8 @@ def install_pythons(boxname)
     pyenv install 3.6.2  # tests
     pyenv install 3.7.11  # binary build, use latest 3.7.x release
     pyenv install 3.8.0  # tests
-    pyenv install 3.9.0  # tests
-    pyenv install 3.10-dev  # tests
+    pyenv install 3.9.9  # tests
```
ThomasWaldmann, line 222: no, this should stay at 3.9.0.

we usually test at the oldest patch-release that works (usually .0).
the only exception is the python release for which we build binaries (and thus do not use .0, but .latest).

ThomasWaldmann: in this case, we use 3.7.11 for the binaries.

ThomasWaldmann:
Thanks!

PR #6181, opened 2022-01-26
Backport repository: quota computation fixes for delayed compaction (#6119)

KN4CK3R:
Backport of #6119.

I adjusted the tests a little bit because the commit method is different.


ThomasWaldmann:
Thanks!

PR #6184, opened 2022-01-27
Doc: borg init: explain the encryption modes better

hexagonrecursion:
The documentation for borg init was not structured logically:
1. The topic is switched from the general discussion of `borg init`
   to the discussion of encryption modes without a title.
2. Obscure technical details (chunking, id generation etc) were
   above the high-level overview and other key information.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2733,7 +2733,7 @@ class Archiver:
             'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
-            'key_files': 'Internals -> Data structures and file formats -> Key files in the online documentation',
+            'key_files': 'Internals -> Data structures and file formats -> Key files',
             'borg_key_export': 'borg key export --help'
         }
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3987,7 +3987,7 @@ class Archiver:
         Borg will:
 
         1. Ask you to come up with a passphrase.
-        2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
+        2. Create a borg key (which contains 3 random secrets. See :ref:`key_files`).
         3. Encrypt the key with your passphrase.
         4. Store the encrypted key alongside the repository in ``/path/to/repo``.
            This is why it is essential to use a secure passphrase.
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3999,7 +3999,7 @@ class Archiver:
            never sees your passphrase, your unencrypted key or your unencrypted files.
            Chunking and id generation are also based on your key to improve
            your privacy (See https://github.com/borgbackup/borg/pull/6184#issuecomment-1028105848).
-        6. Use the key when extracting files to verify that the contents of
+        6. Use the key when extracting files to decrypt them and to verify that the contents of
            the backups have not been accidentally or maliciously altered.
 
         Picking a passphrase
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3976,12 +3976,16 @@ class Archiver:
         you can neither configure it on a per-archive basis nor change the
         encryption mode of an existing repository.
 
-        Use::
+        Use ``repokey``::
 
             borg init --encryption repokey /path/to/repo
 
+        Or ``repokey-blake2`` depending on which is faster on your client machines (see below)::
+
+            borg init --encryption repokey-blake2 /path/to/repo
+
         Borg will:
-        
+
         1. Ask you to come up with a passphrase.
         2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
         3. Encrypt the key with your passphrase.
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3990,7 +3990,7 @@ class Archiver:
         2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
         3. Encrypt the key with your passphrase.
         4. Store the encrypted key alongside the repository in ``/path/to/repo``.
-           This is why it is vital to use a secure passphrase.
+           This is why it is essential to use a secure passphrase.
         5. Encrypt your backups to prevent anyone from reading them unless they
            have the key and know the passphrase. Make sure to keep a backup of
            your key **outside** the repository - do not lock yourself out by
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3994,7 +3994,7 @@ class Archiver:
            For remote backups the encryption is done locally - the remote machine
            never sees your passphrase, your unencrypted key or your unencrypted files.
            Chunking and id generation are also based on your key to improve
-           your privacy (See ).
+           your privacy (See https://github.com/borgbackup/borg/pull/6184#issuecomment-1028105848).
         6. Use the key when extracting files to verify that the contents of
            the backups have not been accidentally or maliciously altered.
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3998,7 +3998,7 @@ class Archiver:
            For remote backups the encryption is done locally - the remote machine
            never sees your passphrase, your unencrypted key or your unencrypted files.
            Chunking and id generation are also based on your key to improve
-           your privacy (See https://github.com/borgbackup/borg/pull/6184#issuecomment-1028105848).
+           your privacy.
         6. Use the key when extracting files to decrypt them and to verify that the contents of
            the backups have not been accidentally or maliciously altered.
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3993,6 +3993,8 @@ class Archiver:
            "leaving your keys inside your car" (see :ref:`borg_key_export`).
            For remote backups the encryption is done locally - the remote machine
            never sees your passphrase, your unencrypted key or your unencrypted files.
+           Chunking and id generation are also based on your key to improve
+           your privacy (See ).
         6. Use the key when extracting files to verify that the contents of
            the backups have not been accidentally or maliciously altered.
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2734,7 +2734,7 @@ class Archiver:
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
             'key_files': 'Internals -> Data structures and file formats -> Key files',
-            'borg_key_export': 'borg key export --help'
+            'borg_key_export': 'borg key export --help',
         }
 
         def process_epilog(epilog):
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2733,7 +2733,7 @@ class Archiver:
             'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
-            'key_files': 'https://borgbackup.readthedocs.io/en/master/internals/data-structures.html#key-files',
+            'key_files': 'Internals -> Data structures and file formats -> Key files in the online documentation',
             'borg_key_export': 'borg key export --help'
         }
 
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2733,6 +2733,8 @@ def build_parser(self):
             'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
+            'key_files': 'https://borgbackup.readthedocs.io/en/master/internals/data-structures.html#key-files',
```
ThomasWaldmann, line 2736: why is this a full https url? the other stuff here looks rather like some link text on the rhs.

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
+        you can neither configure it on a per-archive basis nor change the
+        encryption mode of an existing repository.
 
-        Borg relies on randomly generated key material and uses that for chunking, id
-        generation, encryption and authentication. The key material is encrypted using
-        the passphrase you give before it is stored on-disk.
+        Use::
 
-        You need to be careful with the key / the passphrase:
+            borg init --encryption repokey /path/to/repo
```
ThomasWaldmann, line 3981: well, this was intentionally not just recommended/documented like that.

`repokey` uses hmac-sha256 which is slower than `repokey-blake2` usually - except if your platform happens to have hw accelerated sha256.

so this needs an informed decision especially considering you can't change it later.

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
 
-        If you want "passphrase and having-the-key" security, use one of the keyfile
-        modes. The key will be stored in your home directory (in .config/borg/keys).
-        In the attack scenario, the attacker who has just access to your repo won't
-        have the key (and also not the passphrase).
+        1. Ask you to come up with a passphrase.
+        2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
+        3. Encrypt the key with your passphrase.
+        4. Store the encrypted key alongside the repository in ``/path/to/repo``.
+           This is why it is vital to use a secure passphrase.
```
ThomasWaldmann, line 3989: s/vital/essential/ ?

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
-        mentioned attack scenario, the attacker will have the key (but not the
-        passphrase).
+        Borg will:
 
-        If you want "passphrase and having-the-key" security, use one of the keyfile
-        modes. The key will be stored in your home directory (in .config/borg/keys).
-        In the attack scenario, the attacker who has just access to your repo won't
-        have the key (and also not the passphrase).
+        1. Ask you to come up with a passphrase.
+        2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
```
ThomasWaldmann, line 3986: 2. Create a borg key (which contains 3 random secrets ...

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
+        3. Encrypt the key with your passphrase.
+        4. Store the encrypted key alongside the repository in ``/path/to/repo``.
```
ThomasWaldmann, line 3988: 4. Store the encrypted borg key inside the repository directory (in the repo config).

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
+           This is why it is vital to use a secure passphrase.
+        5. Encrypt your backups to prevent anyone from reading them unless they
+           have the key and know the passphrase. Make sure to keep a backup of
+           your key **outside** the repository - do not lock yourself out by
+           "leaving your keys inside your car" (see :ref:`borg_key_export`).
+           For remote backups the encryption is done locally - the remote machine
+           never sees your passphrase, your unencrypted key or your unencrypted files.
+        6. Use the key when extracting files to verify that the contents of
+           the backups have not been accidentally or maliciously altered.
```
ThomasWaldmann, line 3997: ... and decrypt the contents.

src/borg/archiver.py
```diff
@@ -4011,14 +4017,22 @@ def define_borg_mount(parser):
...
-        basis. The mode determines encryption algorithm, hash/MAC algorithm and also the
-        key storage location.
+        Only use ``--encryption none`` if you are OK with anyone who has access to
+        your repository being able to read your backups and tamper with their
+        contents without you noticing.
 
-        Example: `borg init --encryption repokey ...`
+        If you want "passphrase and having-the-key" security, use ``--encryption keyfile``.
+        The key will be stored in your home directory (in ``~/.config/borg/keys``).
+        TODO: where is the key stored on Windows?
```
ThomasWaldmann, line 4029: remove the TODO, adding this will be the task of some borg windows developer.

src/borg/archiver.py
```diff
@@ -3967,31 +3969,35 @@ def define_borg_mount(parser):
...
+        2. Generate a random key (3 keys to be more precise. See :ref:`key_files`).
+        3. Encrypt the key with your passphrase.
+        4. Store the encrypted key alongside the repository in ``/path/to/repo``.
+           This is why it is vital to use a secure passphrase.
+        5. Encrypt your backups to prevent anyone from reading them unless they
+           have the key and know the passphrase. Make sure to keep a backup of
+           your key **outside** the repository - do not lock yourself out by
+           "leaving your keys inside your car" (see :ref:`borg_key_export`).
+           For remote backups the encryption is done locally - the remote machine
+           never sees your passphrase, your unencrypted key or your unencrypted files.
```
ThomasWaldmann, line 3995: It's not just encryption, but also authentication - could be mentioned here, too, not only in the next item.

ThomasWaldmann (commented):
the chunking and id generation being influenced by a random secret is missing now.

both might be interesting for some people though, considering privacy and fingerprinting attacks.

hexagonrecursion:
Thank you for your review, Thomas. I expect I'll have the time to address your suggestions tomorrow.

hexagonrecursion:
> the chunking and id generation being influenced by a random secret is missing now.
> 
> both might be interesting for some people though, considering privacy and fingerprinting attacks.

I wanted to link to the doc page describing in detail *how* the use of secrets for chunking and id generation improves user privacy, but I can't find it.

ThomasWaldmann:
Maybe there is something in the internals docs, not sure.

But in short:

- the chunker secret determines the cutting places of the chunker (and thus: the length of the chunks cut). So an attacker can not determine the set of chunk lengths for a known set of big files, because the attacker would use a different chunker secret than the attacked borg setup.

- id generation: if we would just use a simple hash (like sha256) as chunk id, it would be bad, because then an attacker could backup a known set of small files (1 chunk each) and the chunkids would directly show the hashes of these files, so the attacker could see which of these known files are in the attacked borg setup. but we use hmac-sha256 and thus the secret random MAC key also influences the id, so that this fingerprinting attack is not possible.

src/borg/archiver.py
hexagonrecursion: IMHO links here would be more helpful. Example:
```sh
borg help compact | tail --lines=1
See "Separate compaction" in Additional Notes for more details.
```
This raises a question: where to find "Additional Notes"?

I suppose it could read "See in Usage -> Additional Notes -> Separate compaction in the online documentation for more details.", but a URL that just take you directly there is better.

I'll remove the URL for now for consistency.

src/borg/archiver.py
```diff
@@ -2733,6 +2733,8 @@ def build_parser(self):
             'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
+            'key_files': 'Internals -> Data structures and file formats -> Key files in the online documentation',
```
ThomasWaldmann, line 2736: guess i would drop the " in the online documentation" part.

the reader is already reading the docs somewhere (doesn't matter whether online or not), so a reference naturally refers to the same docs.

src/borg/archiver.py
```diff
@@ -3967,31 +3969,41 @@ def define_borg_mount(parser):
...
-        case it gets corrupted or lost. Also keep the passphrase at a safe place.
-        The backup that is encrypted with that key won't help you with that, of course.
+        Borg will:
+
+        1. Ask you to come up with a passphrase.
+        2. Create a borg key (which contains 3 random secrets. See :ref:`key_files`).
+        3. Encrypt the key with your passphrase.
+        4. Store the encrypted borg key inside the repository directory (in the repo config).
+           This is why it is essential to use a secure passphrase.
+        5. Encrypt sign your backups to prevent anyone from reading or forging them unless they
```
ThomasWaldmann, line 3994: Encrypt and sign ...?
Encrypt and authenticate ...?

src/borg/archiver.py
```diff
@@ -3967,31 +3969,41 @@ def define_borg_mount(parser):
...
+           have the key and know the passphrase. Make sure to keep a backup of
+           your key **outside** the repository - do not lock yourself out by
+           "leaving your keys inside your car" (see :ref:`borg_key_export`).
+           For remote backups the encryption is done locally - the remote machine
+           never sees your passphrase, your unencrypted key or your unencrypted files.
+           Chunking and id generation are also based on your key to improve
+           your privacy (See https://github.com/borgbackup/borg/pull/6184#issuecomment-1028105848).
```
ThomasWaldmann, line 4001: We rather should have this inside the docs somewhere than on github. Maybe it already is there.

src/borg/archiver.py
```diff
@@ -3967,31 +3969,41 @@ def define_borg_mount(parser):
...
+        6. Use the key when extracting files to decrypt them and to verify that the contents of
+           the backups have not been accidentally or maliciously altered.
```
ThomasWaldmann, line 4003: great!

src/borg/archiver.py
hexagonrecursion: As far as I can tell:
1. This placeholder is _only_ printed by `borg help init`
2. The `borg help` only supports help for specific commands - only a fraction of the borg docs

In this situation I do not see a better solution than referring to the more complete online docs

src/borg/archiver.py
hexagonrecursion: Typo. Sorry. Good catch! Fixed.

src/borg/archiver.py
ThomasWaldmann: But this is nothing new, right? so it also applies to the lines above and we should stay consistent.

hexagonrecursion: I see your point. Will fix in the next commit

src/borg/archiver.py
hexagonrecursion: I have searched the docs and still can't find anything worth lining to. There are a couple places that mention this in passing, but nothing detailed. I could expand on this. Should this be a part of this PR or a separate one?

src/borg/archiver.py
```diff
@@ -2733,6 +2733,8 @@ def build_parser(self):
             'a_status_oddity': '"I am seeing ‘A’ (added) status for a unchanged file!?"',
             'separate_compaction': '"Separate compaction"',
             'list_item_flags': '"Item flags"',
+            'key_files': 'Internals -> Data structures and file formats -> Key files',
+            'borg_key_export': 'borg key export --help'
```
ThomasWaldmann, line 2737: if one adds a trailing comma after the last element, future changes are easier / diffs are prettier.

ThomasWaldmann: see your own change above: you did not need to edit line 2735, because the comma was already there.

src/borg/archiver.py
ThomasWaldmann: ok, then I guess we can't link it for now and need to improve that later.

i don't want links to comments in pull requests in the help/docs.

ThomasWaldmann (commented):
a few minor nitpicks, but guess we are done after them.

src/borg/archiver.py
hexagonrecursion: Done.

src/borg/archiver.py
ThomasWaldmann: I'ld like to merge this soon. Can you just remove the part in parentheses for now?

hexagonrecursion: done

ThomasWaldmann:
Thanks!

PR #6185, opened 2022-01-27
Doc: ~/.config/borg/keys is not used w/ repokey

### LINKED ISSUE #6107: Incomplete FAQ item about keys

Yes, and that's why I'm opening this issue.

It's a documentation issue, combinded with a question.

The FAQ at [How important is the $HOME/.config/borg directory?](https://borgbackup.readthedocs.io/en/master/faq.html#how-important-is-the-home-config-borg-directory) state about the `$HOME/.config/borg/keys` directory, that the keys in there are essential and cannot be recovered. Losing them means losing access to the repository.

We do not have that directory on any of our hosts and that's confusing in relation to this FAQ statement. We are using SSH private keys to access the repositories and those are stored in `$HOME/.ssh` which works just fine. I guess, it's those keys you're referring to, right?

If that's correct, then the FAQ should probably be updated.

And also, if we lost that SSH key on a host but still had access to the UI of the repository over https://www.borgbase.com, we could still assign a new SSH key to an existing repository and recover access to it that way.

Is that true too? If so, that would be nice to be added to the FAQ too.

hexagonrecursion:
Fixes #6107

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -514,7 +514,8 @@ The Borg config directory has content that you should take care of:
   recovered.
 
 ``keys`` subdirectory
-  In this directory all your repository keyfiles are stored. You MUST make sure to have an
+  In this directory all your repository keyfiles are stored. This does **not** apply if you
+  use the repokey mode because your key is stored alongside the the repository. You MUST make sure to have an
   independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
   them. You also MUST keep these files secret; everyone who gains access to your repository and has
   the corresponding keyfile (and the key passphrase) can extract it.


docs/faq.rst
```diff
@@ -514,7 +514,8 @@ The Borg config directory has content that you should take care of:
   recovered.
 
 ``keys`` subdirectory
-  In this directory all your repository keyfiles are stored. You MUST make sure to have an
+  In this directory all your repository keyfiles are stored. This does **not** apply if you
+  use the repokey mode because your key is stored alongside the the repository. You MUST make sure to have an
```
ThomasWaldmann, line 518: almost. :-)

I'ld rather phrase it like "This does not apply to repokey keys because for them the key is stored in the repository."

Note: There still can be other keys in that keys subdirectory.

docs/faq.rst
```diff
@@ -514,8 +514,8 @@ The Borg config directory has content that you should take care of:
   recovered.
 
 ``keys`` subdirectory
-  In this directory all your repository keyfiles are stored. This does **not** apply if you
-  use the repokey mode because your key is stored alongside the the repository. You MUST make sure to have an
+  In this directory all your repository keyfiles are stored. This does **not** apply to repokey keys because they
+  are stored alongside the the repository rather than in ``$HOME/.config/borg/keys``. You MUST make sure to have an
```
ThomasWaldmann, line 518: sorry to bother again, but noticed something:

the text talks about **repository keyfiles**, which can easily be confused that it means **repokey** specifically or all keys including repokeys (which is both wrong). so guess it better should read:

```
All your borg keyfile keys are stored in this directory.
```

The next sentence then should read (shorter and easier):
```
Please note that borg repokey keys are stored inside the repository.
```

hexagonrecursion: > inside the repository


I do not understand why you insist on saying that repokey keys are stored *inside* the repository, but you are the boss. My concern is that users might think that repokey keys are stored inside the backup even though they are merely stored in the same directory as the repository.

ThomasWaldmann: well, "repository" is all inside the repository directory (e.g. "repo/") and the repokey is in "repo/config", so it is inside (the directory). one could also say that it is in the repository config, but not everybody will know that the repository config is inside the repository directory.

we could check if the docs say this as clear as that, but maybe that is another place in the docs, this place is specifically about `~/.config/borg/keys`.

hexagonrecursion: ok

ThomasWaldmann:
Thanks!

PR #6186, opened 2022-01-27
Refactor: move a test to ArchiverTestCase

### LINKED ISSUE #6168: test methods in wrong test class
check for 1.1-maint and master branch:

it seems that there are some test methods in `RemoteArchiverTestCase` that should be rather in the `ArchiverTestCase` class (which is inherited by `RemoteArchiverTestCase`).

idea:
- most tests should be in the base class `ArchiverTestCase`
- the remote tests class should contain tests that only make sense for remote repos
- the remote tests class might contain some test overrides for tests that do not work remotely, but would normally be inherited from the base class
-

hexagonrecursion:
Fixes #6168

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -3499,6 +3499,13 @@ id: 2 / e29442 3506da 4e1ea7 / 25f62a 5a3d41 - 02
             with patch.object(xattr, 'setxattr', patched_setxattr_EACCES):
                 self.cmd('extract', self.repository_location + '::test', exit_code=EXIT_WARNING)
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
test_strip_components_doesnt_leak <- i think that one is also at the wrong place (run the tests after moving it).

ThomasWaldmann:
could you rename the strip test a little, so that there is remote_repo in the test name.

if one does not fully read the test (like me), one would assume that a strip-related test also applies to local repos, but in this case, the condition checked only applies to remote repos.

hexagonrecursion:
Done

ThomasWaldmann:
Thanks!

PR #6196, opened 2022-01-29
WIP: fix fuse tests hanging

ThomasWaldmann:


.github/workflows/ci.yml
```diff
@@ -121,12 +86,17 @@ jobs:
...
+        pip install llfuse
+        echo "AllFiles vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv"
+        sudo sqlite3 /Library/Application\ Support/com.apple.TCC/TCC.db 'select * from access' | grep AllFiles
+        # kTCCServiceSystemPolicyAllFiles|/bin/bash|1|2|0|1||||UNUSED||0|1583997993
+        which borg
+        # vvv only available from revovery OS
+        #sudo csrutil disable
+        #sudo csrutil status
+        sudo tccutil --service kTCCServiceSystemPolicyAllFiles --insert /bin/sh
+        sudo tccutil --enable /bin/sh
```
ThomasWaldmann, line 99: doesn't work, likely SIP is enabled.

ThomasWaldmann:
@m3nu any idea how to get test_fuse not hanging forever?

m3nu:
Installing and running FUSE in an automated way is ambitious. Doesn't it need confirmation of the module in the *Security Preferences*? If so, there is probably a notification popup blocking somewhere.

I can check for that by taking screenshots while Github Actions is running, but may not have a solution for the actual error. Similar to doing things with the Keychain. One needs to be very very careful to avoid triggering a confirmation message.

m3nu:
Will try this today. We shouldn't disable SIP or any other feature.

m3nu:
Here where it gets stuck.

The workarounds in the script are being blocked. Looking at other options to install this headless.

test

m3nu:
Looked into this a bit more (https://github.com/osxfuse/osxfuse/issues/828#issuecomment-1033292345) and don't think we should spend time to override a security feature here. Better to mock the MacFUSE parts and focus on the code Borg runs.

m3nu:
Only alternative I see is Vagrant or something local, which allows disabling SIP.

ThomasWaldmann:
@m3nu thanks for investigating! ok, so according to your results / advice, we'll have:

- github CI running on old macOS that doesn't need problematic tweaks (as long as that OS is offered / works otherwise for us)
- local vagrant based testing (currently also on old macOS, in future maybe some newer OS with a locally prepared VM)

So I am closing this PR, won't work like this.

ThomasWaldmann:
well, github is deprecating macOS 10 based testing. today a test failed with a note telling so.

so guess we have the same problem again on github as we are now forced to move to macOS 11 or 12.

PR #6202, opened 2022-01-30
delete: don't commit if nothing was deleted, avoid cache sync, fixes #6060

### LINKED ISSUE #6060: borg delete with a non existing archive name leaves a outdated local cache
When trying to delete an archive that does not exist from a repository borg leaves the local cache in an outdated state (repo manifest time stamp does not match cache manifest timestamp). This causes the next access to the repository to do a costly cache rebuild.

I think this is caused by _delete_archives forcing a manifest write in all cases but when no archive was deleted the local cache instance has no active transaction and thus does not update the cache configuration when _delete_archives triggers a commit of the cache.

Maybe LocalCache.commit should check the manifest timestamp even if there is no active transaction and if it differs trigger an update for the local cache config anyway. If that is seen as to risky we could add code in _delete_archives to force a transaction in LocalCache and this avoid leaving an seemingly outdated cache. Or maybe not write the manifest if nothing changed.

I've encountered this problem on a system using a fairly old borg version (1.10+git1+0d9e88d) but source inspection says that this is likely still true in master. I didn't look at latest 1.1.x but this is like there also.

ThomasWaldmann:


textshell (approved):
Looks good.

Issue #6204, opened 2022-01-30
Documentation contradiction: required backup of keyfile


rolandu:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Documentation BUG (I think)

## Affected documentation

(1) https://borgbackup.readthedocs.io/en/stable/quickstart.html#encrypted-repos

> If the repository is encrypted, you will also need the key and the passphrase (which is protecting the key).
> 
> The key can be located:
> in the repository (repokey mode). Easy, this will usually “just work”.
> in the home directory of the user who did the backup (keyfile mode). This may cause a bit more effort: (...)

(2) https://borgbackup.readthedocs.io/en/stable/usage/init.html

> If you want “passphrase-only” security, use one of the repokey modes. The key will be stored inside the repository (in its “config” file). In above mentioned attack scenario, the attacker will have the key (but not the passphrase).
> 
> If you want “passphrase and having-the-key” security, use one of the keyfile modes. The key will be stored in your home directory (in .config/borg/keys). In the attack scenario, the attacker who has just access to your repo won’t have the key (and also not the passphrase).

(3) https://borgbackup.readthedocs.io/en/stable/faq.html#how-can-i-specify-the-encryption-passphrase-programmatically

> Using keyfile-based encryption with a blank passphrase

(4) [How important is the $HOME/.config/borg directory?](https://borgbackup.readthedocs.io/en/stable/faq.html#how-important-is-the-home-config-borg-directory)

> keys subdirectory
> In this directory all your repository keyfiles are stored. You MUST make sure to have an independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose them. You also MUST keep these files secret; everyone who gains access to your repository and has the corresponding keyfile (and the key passphrase) can extract it.

## Description of problem

- 1 and 2 tells me that there are different encryption modes: repokey will store the key together with the repo, keyfile will not.
- 3 tells me that there is a special case of keyfile encryption where I don't set a passphrase. Only in this special case will the keyfile be unencrypted. It will be unencryped and only locally stored in that case.
- 4 says that the "keys" subdirectory contains keyfiles and it sounds as if they are *always required* for disaster recovery and *always unencrypted*... which contradicts the above in my opinion.
[...]

PR #6207, opened 2022-01-30
Clarification of keyfile backup

rolandu:
Proposed clarification in response to my confusion in https://github.com/borgbackup/borg/issues/6204.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -514,10 +514,12 @@ The Borg config directory has content that you should take care of:
   recovered.
 
 ``keys`` subdirectory
-  All your borg keyfile keys are stored in this directory. Please note that
-  borg repokey keys are stored inside the repository. You MUST make sure to have an
-  independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
-  them. You also MUST keep these files secret; everyone who gains access to your repository and has
+  All your borg keyfile keys are stored in this directory. Please note that borg repokey keys are stored inside the repository. 
+  
+  If you use keyfile encryption, you MUST make sure to have an independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
+  them. If you use repokey encryption it is also highly recommended that you backup the keyfiles in case the originals (stored within the repository) get damaged. The best way to create a backup is using the ``borg key export`` command. 
+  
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -514,10 +514,12 @@ The Borg config directory has content that you should take care of:
...
 
 ``keys`` subdirectory
-  All your borg keyfile keys are stored in this directory. Please note that
-  borg repokey keys are stored inside the repository. You MUST make sure to have an
-  independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
-  them. You also MUST keep these files secret; everyone who gains access to your repository and has
+  All your borg keyfile keys are stored in this directory. Please note that borg repokey keys are stored inside the repository. 
+  
+  If you use keyfile encryption, you MUST make sure to have an independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
+  them. If you use repokey encryption it is also highly recommended that you backup the keyfiles in case the originals (stored within the repository) get damaged. The best way to create a backup is using the ``borg key export`` command.
```
ThomasWaldmann, line 520: Thinking about this: can't we just simply say that you must have a backup of the keys in any case and use borg key export for this?

In the end, it doesn't really matter where they are stored and how you lose them, but that for both options, it is possible to lose them.

rolandu: True, something like "backup the key in case you need it" is probably the shortest version for the end user.

My confusion really came from the fact that this said disaster recovery without a backup of the keyfile was impossible but I did remember doing just that last year (which was possible because I had repokey and not keyfile enc).

ThomasWaldmann: Or: By all means, backup the key! And keep the backup and your passphrase at a safe place.

ThomasWaldmann:
Hmm, I just noticed we should document that elsewhere: In the `borg key export` docs.

Currently, we have 2 places talking about keys and key backup:
- faq (== this PR)
- quickstart (the warning section below "repository encryption")

I think both places should only have one sentence about that where they link to `borg key export` docs.
Like "You **must** make a backup of your borg key, see --->borg key export how to do that.".

And everything else should get refactored and deduplicated into the `borg key export` docs.

ThomasWaldmann:
Superceded by #9287.

PR #6223, opened 2022-02-02
do not use broken setuptools version

ThomasWaldmann:
see there: https://github.com/pypa/setuptools/issues/3063

### DIFF
--- a/requirements.d/development.txt
+++ b/requirements.d/development.txt
@@ -1,4 +1,4 @@
-setuptools>=45
+setuptools >=45, !=60.6.0
 setuptools_scm
 pip
 virtualenv


hexagonrecursion:
Is this still required?

https://github.com/pycalphad/pycalphad/pull/393#issue-1122147830
> pypa/setuptools#3070 is fixes pypa/setuptools#3063 in setuptools 60.6.0 and 60.7.0 is available now.

hexagonrecursion:
If so, wouldn't a better fix be `setuptools >=60.6.0` ?

hexagonrecursion:
https://github.com/pypa/setuptools/releases

I see:
1. v60.5.0 released Jan 8
2. The bug was introduced Jan 14 https://github.com/pypa/setuptools/commit/dba50bf3180cc61e4691dbaf64dd7437da27f92f
3. The only broken version (v60.6.0) was released Jan 31
4. The fix (v60.7.0) was released Feb 2 10:32 UTC

ThomasWaldmann:
Our change is not required any more for fresh virtual envs considering the new version is released now.
For existing virtual envs that have the broken version, we better force upgrading it by requirement.

Using >= would be wrong as we also work with many other versions, except the broken one.

We can remove this later again, but guess we better keep it for a while.

hexagonrecursion:
The backport is merged. The backport/1.1-maint label can be removed.

PR #6177, opened 2022-01-25
Backport key export: print key if path is '-' or not given, fixes #6092 (#6093)

KN4CK3R:
Backport of #6093

key export: print key if path is '-' or not given, fixes #6092

"borg key export [REPOSITORY]" now displays the key.

PR #6227, opened 2022-02-03
Doc: borg key export: add examples

hexagonrecursion:
https://github.com/borgbackup/borg/issues/6204#issuecomment-1027150308

The examples I provided should work in the next 1.2 release and
the next 1.1 release (see #6092 and #6177). Alternatively I could give
the following example that should work today:
```sh
borg key export --qr-html /path/to/repo encrypted-key-backup.html
```

@ThomasWaldmann, you have added "documentation" label after a comment by https://github.com/rolandu. Is this the fix you wanted/expected?

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4131,6 +4131,13 @@ class Archiver:
         repository in the config file. A backup is thus not strictly needed,
         but guards against the repository becoming inaccessible if the file
         is damaged for some reason.
+
+        Examples::
+
+            borg key export /path/to/repo > encrypted-key-backup
+            borg key export /path/to/repo --paper > encrypted-key-backup.txt
+            borg key export /path/to/repo --qr-html > encrypted-key-backup.html
+
         """)
         subparser = key_parsers.add_parser('export', parents=[common_parser], add_help=False,
                                           description=self.do_key_export.__doc__,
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
maybe additionally give the same examples, but instead of output redirection, give the output filename as parameter.

the docs are versioned, so people can always read the correct docs for the release they use.

hexagonrecursion:
> maybe additionally give the same examples, but instead of output redirection, give the output filename as parameter.

My rationale for only including the redirection was: (1) less is more, (2) knowing one way to export a key is enough, (3) output redirection is more idiomatic on a unix-like OS, but you are the boss. Adding 3 more examples now.

ThomasWaldmann:
@hexagonrecursion i had similar thoughts, but user preferences and use cases are different. Also, if somebody uses an old borg version and reads the latest docs, they will still find a working command even if their borg does not yet support the stdout-by-default method.

hexagonrecursion:
@ThomasWaldmann, we are out of "do back/forward ports" bounties again. BTW If you decide to create more, I suggest you use one search link instead of two: https://github.com/borgbackup/borg/issues?q=label%3Aforwardport%2Fmaster%2Cbackport%2F1.1-maint+ - it's easier to check if there are any back/forward ports available with one link.

ThomasWaldmann:
Yeah, I will create some today. OK, I'll use the better link, thanks!

hexagonrecursion:
The backport is merged. The backport/1.1-maint label can be removed.

Issue #6231, opened 2022-02-03
Patternfile doesn't seem to work with relative paths


bobthebadguy:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Question

## System information. For client/server mode post info for both machines.

Arch linux
#### Your borg version (borg -V).
1.1.17
#### Hardware / network configuration, and filesystems used.
Zfs
#### How much data is handled by borg?
None
#### Full borg commandline that lead to the problem (leave away excludes and passwords)

`borg create --list --dry-run --patterns-from patternfile ~/borgrepo::test`

Patternfile:
```
R /home/user/borgtest
- dir1
+ dir2/includeme
- dir2
```
Directory structure:
```
|-- dir1
|   |-- dir1-a
|   `-- dir1-b
|-- dir2
|   |-- excludeme
|   `-- includeme
```
## Describe the problem you're observing.

In the patternfile, after specifying a root with `R`, exclusion rules still require an absolute path.
With the setup described above, I would expect dir1 not to be backed up, and only includeme to be backed up, and not excludeme.

What actually happens is that everything is backed up:
```
- /home/user/borgtest/dir1/dir1-b
- /home/user/borgtest/dir1/dir1-a
- /home/user/borgtest/dir1
- /home/user/borgtest/dir2/includeme
- /home/user/borgtest/dir2/excludeme
```

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.
The patternfile only works if all the paths are absolute. This becomes a problem, since some of my paths are very long.

The docs seem to suggest that using `R` sets a root from which rules apply from, which doesn't seem to be happening.

ThomasWaldmann:
- check if docs need to be more clear
- check if we can easily implement relative patterns

bobthebadguy:
After re-reading the docs a few times, I think I understand.

`R` is just a replacement for specifying paths on the command line and it doesn't allow relative paths:
```
borg create ~/borgrepo::test ~/backmeup
    This here can be replaced by R ↑
```
Is that correct? Also, I assume that specifying multiple roots works the same way.

The docs don't seem really clear on that specific part.

ThomasWaldmann:
Yeah, the paths you give on the cmdline are the same as the recursion roots (R).

ThomasWaldmann:
Docs fixed by #6242 and #6265.

jefft:
[...]

PR #6242, opened 2022-02-05
Clarify usage of patternfile roots

bobthebadguy:
See #6231. Clarifies what patternfile roots actually do with examples.


docs/usage/help.rst.inc
```diff
@@ -200,6 +200,18 @@ Examples::
...
+        borg create --exclude /home/bobby/junk repo::arch /home/bobby /home/susan
+
+    Or you could do that with a patternfile like so, replacing roots specified on the command line with ones in the file::
+
+        R /home/bobby
+        R /home/susan
+
+        # Note that excludes use fm: by default and patternfiles use sh: by default.
+        - /home/bobby/junk
+
```
ThomasWaldmann, line 214: maybe add the borg command that uses this pattern file.

bobthebadguy:
I still feel like this could be a bit clearer

docs/usage/help.rst.inc
```diff
@@ -149,7 +149,7 @@ Examples::
 
     A more general and easier to use way to define filename matching patterns exists
     with the experimental ``--pattern`` and ``--patterns-from`` options. Using these, you
-    may specify the backup roots (starting points) and patterns for inclusion/exclusion.
+    may specify the backup roots (starting points), pattern files, and patterns for inclusion/exclusion.
```
ThomasWaldmann, line 152: this paragraph is talking about the patterns specifically (not the command line in general), so we are already "in" the pattern file, so you can't specify a pattern file there.

so I'ld suggest to just remove this change.

docs/usage/help.rst.inc
```diff
@@ -195,11 +195,30 @@ Examples::
...
+        borg create --patterns-from patternfile.lst repo::arch
+
+    The patternfile::
+
+        R /home/bobby
+        R /home/susan
+
+        # note that excludes use fm: by default and patternfiles use sh: by default.
+        - /home/bobby/junk
+
```
ThomasWaldmann, line 219: you say "same thing", so people might expect this is precisely doing the same thing as the previous example, which it does not.

maybe adjust it so it really does exactly the same thing?

ThomasWaldmann (changes requested):
most important: read the comment at the top of the file you are editing.

just use full-text search the right place, likely it is in `src/borg/archiver.py`.

bobthebadguy:
Did I mess up the indentation? 
Sorry about that, should have read more closely

src/borg/archiver.py
```diff
@@ -2412,10 +2412,31 @@ def do_break_lock(self, args, repository):
...
                 + /home/susan
+                # also back up this exact file
+                + pf:/home/bobby/specialfile.txt
                 # don't backup the other home directories
                 - /home/*
                 # don't even look in /proc
-                ! /proc\n\n''')
+                ! /proc
+
+            As another example, you could do the same thing in either of these ways::
```
ThomasWaldmann, line 2422: "same" might get misunderstood as a backreference to the previous example (lines before 2420).

maybe rather say:

Here is another example that shows how recursion roots and excludes can be given either directly on the command line or in a patternfile.

(with that, you could drop the comments from 2424 and 2426)

src/borg/archiver.py
```diff
@@ -2412,10 +2412,31 @@ def do_break_lock(self, args, repository):
...
+
+                # note that excludes use fm: by default and patternfiles use sh: by default.
+                # therefore, we need to specify fm: to have the same exact behavior.
+                P fm
+                R /home/bobby
+                R /home/susan
+
+                - /home/bobby/junk
+
+            This allows you to share the same patterns between multiple repositories without needing to specify them on the command line.\n\n''')
```
ThomasWaldmann, line 2439: borg can output this help directly on the terminal.
if a line is longer than 80 chars, rather continue on next line.

ThomasWaldmann:
there seems to be a conflict now with other master branch changes.

and because you have committed to directly to your master branch, this is a little tricky to fix now...

maybe just tell when you've finished your changes and i'll try to fix it.

ThomasWaldmann:
now indentation is wrong again. most stuff will appear inside a code block now instead as normal text.

ThomasWaldmann:
```
cd docs ; make clean ; make html
```
then look at the local html files with a browser.

bobthebadguy:
It looks like that fixed it?

Thanks for all your help with this and sorry for the trouble

ThomasWaldmann (approved):
LGTM

ThomasWaldmann:
Thanks!

hexagonrecursion:
I plan to backport this today

PR #6243, opened 2022-02-05
Test: borg init refuses to overwrite keyfile

hexagonrecursion:
https://github.com/borgbackup/borg/pull/6046 neglected to add a regression test


ThomasWaldmann:
thanks, but the diff here looks strange.

hexagonrecursion:
> the diff here looks strange

fixed

Issue #6224, opened 2022-02-02
update development.lock.txt


ThomasWaldmann:
1.1-maint: check versions and either use all latest working versions or just update the critical stuff.

master: create such a file.

ThomasWaldmann:
master: fixed by #6235 
1.1-maint: todo (== backport of #6235)

PR #6235, opened 2022-02-04
add locked requirements (master)

ThomasWaldmann:

PR #6253, opened 2022-02-05
Backport: add locked requirements

hexagonrecursion:
This will be a backport of https://github.com/borgbackup/borg/pull/6235
Fixes: #6224

The first step is to test the updated requirements on github ci. I'll add the rest once the CI is green.

https://github.com/borgbackup/borg/pull/6235#issuecomment-1030312419
> try the locked requirements on github ci first (by placing them in development.txt)


hexagonrecursion:
```
ERROR: Could not find a version that satisfies the requirement pip==22.0.3
```
I'll debug this tomorrow morning

hexagonrecursion:
https://pip.pypa.io/en/stable/news/#id13

pip 22.0 dropped support for Python 3.6. What should we do? Should we lock an old version of pip? Should we drop Python 3.6?

It might be possible to make the lock depend on the python version.

ThomasWaldmann:
Lock pip to a working version. In 1.1-maint, we should not drop old python versions except if there is no other possibility or if they are out of support for a while.

hexagonrecursion:
AttributeError: module 'collections' has no attribute 'MutableMapping

I'll debug this later today

requirements.d/development.txt
```diff
@@ -1,11 +1,11 @@
...
+setuptools-scm==5.0.2
+pip==20.3.4
+virtualenv==20.13.1
+tox==3.24.5
+pytest==6.1.2
+pytest-xdist==2.2.1
+pytest-cov==2.12.1
+pytest-benchmark==3.4.1
+Cython==0.29.27
+twine==1.15.0
```
ThomasWaldmann, line 11: for reference, this is what the locked requirements in 1.1-maint used previously:
```
setuptools==50.3.0
setuptools-scm==4.1.2
pip==20.2.3
virtualenv==20.0.33
importlib-metadata==1.7.0
pluggy==0.13.1
tox==3.20.0
pytest==6.1.1
pytest-xdist==2.1.0
pytest-cov==2.10.1
pytest-benchmark==3.2.3
Cython==0.29.23
twine==1.15.0
```

hexagonrecursion: Sorry. setuptools==28.8.0 is a result of some strange interaction between conda and venv.

hexagonrecursion: Cython==0.29.27 is the latest Cython (unless you count pre-releases). pytest==6.1.2 is the last pytest supporting python3.5 (see https://docs.pytest.org/en/stable/changelog.html#id97)

ThomasWaldmann: to fix the current issue maybe keep these pluggy and importlib-metadata lines at that place.

hexagonrecursion:
```
TypeError: required field "lineno" missing from alias
```
Possibly relevant: https://github.com/pytest-dev/pytest/issues/8539 https://github.com/pytest-dev/pytest/issues/8546

hexagonrecursion:
I have tried the existing requirements.d/development.lock.txt and it fails on python3.10 too. There appears to be no version of pytest that simultaneously supports both py3.5 ad 3.10:
1. [pytest 6.2.x supports python3.6+ only](https://docs.pytest.org/en/stable/changelog.html#id97)
2. https://github.com/pytest-dev/pytest/pull/8540 was backported to pytest6.2.x, but not to pytest6.1.x

I see two paths forward:
1. Theoretically https://github.com/pytest-dev/pytest/pull/8540 could be backported to pytest6.1.x
2. Use environment markers

ThomasWaldmann:
just noticed that i used the locked requirements for the "main" borg build in vagrant (see Vagrantfile).
guess that build is not done with py310.

but then, when tox runs in the vagrant machine, it uses the unlocked requirements (see tox.ini).

so, we'll just ignore the py310 issue and make the locked requirements as good as it gets for < 3.10.

hexagonrecursion:
> note: the borg binaries are built with python 3.7.11 so it is most important to have everything up-to-date that is used for that.

This appears to be at odds with the goal of supporting 3.5:
```sh
$ python -V
Python 3.7.11
$ pip list --outdated
Package        Version Latest Type
-------------- ------- ------ -----
pip            20.3.4  22.0.3 wheel
pluggy         0.13.1  1.0.0  wheel
pytest         6.1.2   7.0.0  wheel
pytest-cov     2.12.1  3.0.0  wheel
pytest-xdist   2.2.1   2.5.0  wheel
setuptools     50.3.2  60.8.1 wheel
setuptools-scm 5.0.2   6.4.2  wheel
twine          1.15.0  3.8.0  wheel

```

ThomasWaldmann:
When I said "up-to-date" I meant also "as good as it gets" considering our other requirements (thus not necessarily using the latest and greatest, but something more recent than we had before).

ThomasWaldmann:
BTW, considering your recent activity, feel free to add yourself to AUTHORS file if you like.

hexagonrecursion:
Should be ready now. CI is happy with the locks: https://github.com/borgbackup/borg/runs/5085767052

PR #6255, opened 2022-02-05
docs: indentation fixes (master)

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -2237,7 +2237,7 @@ def do_break_lock(self, args, repository):
         File patterns support these styles: fnmatch, shell, regular expressions,
         path prefixes and path full-matches. By default, fnmatch is used for
         ``--exclude`` patterns and shell-style is used for the ``--pattern``
-         option. For commands that support patterns in their ``PATH`` argument
```
hexagonrecursion, line 2240: Regression introduced in 5d4dda75d36d4bdaff9fc4a38a234383aef4b733 - nothing to backport

src/borg/archiver.py
```diff
@@ -2358,64 +2358,64 @@ def do_break_lock(self, args, repository):
             EOF
             $ borg create --exclude-from exclude.txt backup /
```
hexagonrecursion, line 2360: The below regression was introduced in 5d4dda75d36d4bdaff9fc4a38a234383aef4b733 - nothing to backport

src/borg/archiver.py
ThomasWaldmann: yup.

src/borg/archiver.py
ThomasWaldmann: do you mean the bigger block with bad indentation? i don't see it there.

hexagonrecursion: Sorry. My comment got stuck because of the github review feature and I did not notice.

ThomasWaldmann: Thanks for the hint!

ThomasWaldmann: removed backport label, nothing to do here...

PR #6263, opened 2022-02-07
improve linking inside docs

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -3328,16 +3330,15 @@ def define_borg_mount(parser):
...
-        See the output of the "borg help placeholders" command for more help on placeholders.
+        For more help on placeholders, see the :ref:`borg_placeholders` command output.
 
         .. man NOTES
 
         The ``--exclude`` patterns are not like tar. In tar ``--exclude`` .bundler/gems will
         exclude foo/.bundler/gems. In borg it will not, you need to use ``--exclude``
-        '\\*/.bundler/gems' to get the same effect. See ``borg help patterns`` for
-        more information.
+        '\\*/.bundler/gems' to get the same effect.
```
ThomasWaldmann, line 3341: Removed that because it was just written a few lines above this...

ThomasWaldmann:
merge after #6184 (conflicts).

ThomasWaldmann:
when backporting, do NOT backport the py 3.5 -> 3.8 docs link change, 1.1-maint is still using 3.5+.

hexagonrecursion:
Hello. I have noticed that the exact same code is sometimes in a different order between 1.1 and master. This is causing many conflicts when using `git cherry-pick`. How did this happen?

Workaround:
```sh
git show "$COMMIT" | git apply -
```

ThomasWaldmann:
don't remember. maybe some cleanup?

ThomasWaldmann:
but i hope we can soon do a bit less backporting.

after 1.2.x is stable enough so everybody who wants "latest" can use it, we can limit backports to critical stuff.

otoh, backports might become more difficult when we run black and some "python upgrader" on the source.

ThomasWaldmann:
also, `borg.archiver` is a candidate for splitting into a package of modules (same as happened for `borg.helpers` a while ago).

PR #6290, opened 2022-02-12
compact segments: improve freeable / freed space log output

ThomasWaldmann:


src/borg/repository.py
```diff
@@ -778,7 +778,6 @@ def complete_xfer(intermediate=True):
                              segment, freeable_ratio * 100.0, freeable_space)
                 pi.show()
                 continue
-            freed_space += freeable_space  # this is what we THINK we can free
```
ThomasWaldmann, line 781: see comment, we just got that from hints and that says nothing about whether we really can free that amount.

hexagonrecursion:
Hello. I think you forgot to remove the backport/1.1-maint label.

PR #6295, opened 2022-02-13
diff: reduce memory consumption, fix is_hardlink_master

taandrighetti:
I was trying the `diff` command with somewhat big archives (60 GB of a Linux root filesystem) but with little differences (the second backup created right after the first one), and it used a lot of memory compared to `export-tar`.

Adding some prints in the code, I figured that `hardlink_masters` grew up to 300k items (the partition has only 10k hardlinks), but most of them were directories.

Compared to the other implementation of [item_is_hardlink_master](https://github.com/borgbackup/borg/blob/master/src/borg/archive.py#L2055), I conclude that storing the directories is a bug.

Comparing version 1.1.17 to `1.1-maint` with this change, it decreases the memory used (measured with `/usr/bin/time -f %M`) from 1.1 GB to 110 MB for the diff between the archives of the Linux partition.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -990,7 +990,7 @@ Utilization of max. archive size: {csize_max:.0%}
             return 'source' not in item or not hardlinkable(item.mode) or item.source in hardlink_masters
 
         def is_hardlink_master(item):
-            return item.get('hardlink_master', True) and 'source' not in item
+            return item.get('hardlink_master', True) and 'source' not in item and hardlinkable(item.mode)
 
         def update_hardlink_masters(item1, item2):
             if is_hardlink_master(item1) or is_hardlink_master(item2):


ThomasWaldmann:
Thanks for the PR! Guess we need fewer is_hardlink_master implementations to have less bugs. ;-)

ThomasWaldmann:
@imsodin can you have a look?

imsodin:
Wow, that was some time ago :)  
My PR that last touched this code actually just copied over an existing function from a different place. The huge class and nested functions make it very hard to spot such issues. Looks like one could indeed extract the base implementation between the two functions. Not sure if a suitable class for "item utility functions" exists already, I don't remember the codebase well anymore.

taandrighetti:
I was going to create a PR against 1.1-maint but then saw on the development guide that PRs should be against master. Yeah, the same `is_hardlink_master` is present in that branch, before [#2842](https://github.com/borgbackup/borg/pull/2842/files#diff-5a95fd5391f550ba0f2f4a482061ddda0734670c408663a739dc5e821a72188bL1096)

Yes, a refactor would be better. Should I close this PR? Consider it a bug report instead of a proper fix :)

imsodin:
> Yes, a refactor would be better. Should I close this PR? Consider it a bug report instead of a proper fix :)

In my opinion: Rather merge. Better have a fix with the existing nastyness of partial duplication, than having the existing nastyness and no fix. However I am not a regular contributor to this project, so this isn't for me to call the shot here.

ThomasWaldmann:
OK, I have looked at the stuff and this fix looks correct, so I'll merge this.

A backport of the fix to 1.1-maint would be welcome!

ThomasWaldmann:
I opened issue #6298 about the code deduplication.

PR #6296, opened 2022-02-13
info: use a pre12-meta cache to accelerate stats for borg < 1.2 archives

ThomasWaldmann:
first time borg info is invoked on a borg 1.1 repo, it can take
a rather long time computing and caching some stats values for
1.1 archives, which borg 1.2 archives have in their archive
metadata structure. be patient, esp. if you have lots of old
archives.

following invocations are much faster.


ThomasWaldmann:
The cache management is rather simple, it only **adds** archive infos if the archive is made with borg < 1.2.
It never removes anything from the cache. But considering it's only about 300b per archive and we likely only will see decreasing amounts of 1.1 archives, I don't think this is a problem.

Cleanup could be done with `rm pre12-meta`, if desired (forcing a cache rebuild next time).

fantasya-pbem:
I could live with this solution. What would it take to have Borg "convert" a <1.2 repo to 1.2? Could this be done on server only, or would a full decrypt - encrypt cycle be necessary?

m3nu:
I can confirm that this fully resolves the issue I saw with 1.2rc1 and earlier versions. Happy to throw away the Vorta workaround I prepared for it. "Converting" would be interesting too, if it's not risky or something.

ThomasWaldmann:
@fantasya-pbem see there: https://github.com/borgbackup/borg/discussions/6166#discussioncomment-2172971

PR #6297, opened 2022-02-14
docs: explain the difference between a path that ends with or without a slash

### LINKED ISSUE #6258: Always getting "Permission denied" on a folder that has been excluded

Yes

BUG

My borg command is as follow:

```
borg create \
    --exclude-from borg_excluded.txt \
    --progress \
    --one-file-system \
    --compression auto,zlib,6 \
    $BACKUP_DRIVE::$DATE \
    /Users/USERNAME/ \
    /Library
```

And my exclude file as follow:

```
...
/Library/Application Support/ApplePushService/
...
```

So the important part is that the "ApplePushService" folder is excluded. However, whenever I run the backup I'm getting the error:

> /Library/Application Support/ApplePushService: scandir: [Errno 13] Permission denied: '/Library/Application Support/ApplePushService'

So it seems Borg is still trying to access the folder even though I've excluded it. To be sure I've also tried excluding "**/ApplePushService" but that didn't make any difference and I still got the same error.

Any idea what might be the issue? Is that possibly a bug in Borg, or am I using the wrong exclude pattern?

laurent22:
Resolves #6258: explains the difference between a path that ends with or without a slash.

In the doc I specifically mentioned the difference when used as an exclusion pattern, because if I'm not mistaken there's no difference for inclusion patterns (`some/path/` and `some/path` would result in the same backup set). But let me know if I should tweak this.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2234,6 +2234,11 @@ class Archiver:
         /path/to/repo /home/user`` will store all files as
         `home/user/.../file.ext`.
 
+        A folder exclusion pattern can end either with or without a slash ('/').
+        If it ends with a slash, such as `some/path/`, the directory will be
+        included but not its content. If it does not end with a slash, such as
+        `some/path`, both the directory and content will be excluded.
+
         File patterns support these styles: fnmatch, shell, regular expressions,
         path prefixes and path full-matches. By default, fnmatch is used for
         ``--exclude`` patterns and shell-style is used for the ``--pattern``
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2234,10 +2234,10 @@ class Archiver:
         /path/to/repo /home/user`` will store all files as
         `home/user/.../file.ext`.
 
-        A folder exclusion pattern can end either with or without a slash ('/').
-        If it ends with a slash, such as `some/path/`, the directory will be
-        included but not its content. If it does not end with a slash, such as
-        `some/path`, both the directory and content will be excluded.
+        A directory exclusion pattern can end either with or without a slash
+        ('/'). If it ends with a slash, such as `some/path/`, the directory will
+        be included but not its content. If it does not end with a slash, such
+        as `some/path`, both the directory and content will be excluded.
 
         File patterns support these styles: fnmatch, shell, regular expressions,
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -2234,6 +2234,11 @@ def do_break_lock(self, args, repository):
         /path/to/repo /home/user`` will store all files as
         `home/user/.../file.ext`.
 
+        A folder exclusion pattern can end either with or without a slash ('/').
```
ThomasWaldmann, line 2237: s/folder/directory/ please.

i checked the patterns help, it nowhere else uses "folder", but always talks about "directory".

The paths mention here are usually included in backticks, like:
```
... without a slash (`/`).
```

Maybe check if that renders ok here, too.

laurent22: Should it actually be backticks? I don't mind changing it but I've used single quotes here because that's how it is in several instances below (also for ':', '*', '?'). It's not consistent as there are also a few instances of backticks for single characters, but less.

Ok for folder/directory.

ThomasWaldmann: Hmm, right, when talking about single chars, it uses normal ticks. OK, so we keep that one as is.

src/borg/archiver.py
```diff
@@ -2234,10 +2234,10 @@ def do_break_lock(self, args, repository):
         /path/to/repo /home/user`` will store all files as
         `home/user/.../file.ext`.
 
-        A folder exclusion pattern can end either with or without a slash ('/').
-        If it ends with a slash, such as `some/path/`, the directory will be
-        included but not its content. If it does not end with a slash, such as
-        `some/path`, both the directory and content will be excluded.
+        A directory exclusion pattern can end either with or without a slash
+        ('/'). If it ends with a slash, such as `some/path/`, the directory will
```
ThomasWaldmann, line 2238: the upper line is only 68 chars long, so it would look better (and also not cause issues) if you kept the end of the sentence on same line.

ThomasWaldmann: (for html, it of course makes no difference, but it does when read on the terminal, like `borg help patterns`)

laurent22: Sure, I've updated it now.

ThomasWaldmann:
Thanks!

PR #6306, opened 2022-02-16
SaveFile: fix race conditions

ThomasWaldmann:
Thanks to Andrey Bienkowski (@hexagonrecursion) for reporting this and writing reproducer code.

Changes:
- use different, randomly (but recognizably) named temp files while writing (securely made by os.mkstemp())
- make sure temp files are cleaned up in normal and error conditions
- SyncFile can now get corresponding pair of path + open os-level fd
- cleaned up: fd now means os-level fd, f means python-file-like object
- fixed a caller of SaveFile


ThomasWaldmann (commented):
some comments

ThomasWaldmann:
@hexagonrecursion ^^^ (see also last email)

hexagonrecursion:
Confirmed: I can't find any more race conditions inside SaveFile. As long as callers are prepared for the last-to-close-wins semantics we should be good.

PR #6310, opened 2022-02-19
Backport: docs: explain the difference between a path that ends with or without a slash

hexagonrecursion:
This is a backport of https://github.com/borgbackup/borg/pull/6297


ThomasWaldmann:
Thanks!

PR #6301, opened 2022-02-15
safe_unlink (was: truncate_and_unlink)

ThomasWaldmann:
the previous implementation caused collateral damage on hardlink-copies of a repository,
see: https://github.com/borgbackup/borg/discussions/6286

PR #6311, opened 2022-02-19
Write a test for safe_unlink

hexagonrecursion:
I am going to backport #6301, but first I'll gift you a test for it.

### DIFF
--- a/src/borg/testsuite/helpers.py
+++ b/src/borg/testsuite/helpers.py
@@ -1,3 +1,4 @@
+import errno
 import hashlib
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
I have to say it: the top two reasons I :heart: borg:
1. When tests fail it is usually because **I** broke something rather than a flaky test or something that got merged even though it broke the tests.
2. :moneybag:

src/borg/testsuite/helpers.py
```diff
@@ -1133,3 +1134,15 @@ def test_iter_separated():
...
+def test_safe_unlink_is_safe(tmpdir):
+    contents = b"Hello, world\n"
+    victim = tmpdir / 'victim'
+    victim.write_binary(contents)
+    hard_link = tmpdir / 'hardlink'
+    hard_link.mklinkto(victim)
+
+    safe_unlink(hard_link)
+
+    assert victim.read_binary() == contents
```
ThomasWaldmann, line 1148: well, it is ok that we have a test, but we also must face that this only tests the usual behaviour (which is the rather not-so-complicated part of `safe_unlink` and basically just a call to `os.unlink`) and not the more complicated part of it.

ThomasWaldmann: hmm, can you check whether mocking `os.unlink` to raise ENOSPC works?

ThomasWaldmann:
Thanks!

PR #6345, opened 2022-02-23
Documented the needs of a Secret Key

### LINKED ISSUE #6268: Doc: explain why chunking and id generation depend on a secret key
Borg uses a secret key for chunking and id generation. This is a feature that is supposed to enhance privacy. Currently there appears to be virtually no information about this feature in the documentation - its existence is mentioned in a couple of places, but that is it.

The most comprehensive explanation of the feature that I was able to find is this:
https://github.com/borgbackup/borg/pull/6184#issuecomment-1028105848

> Maybe there is something in the internals docs, not sure.
> 
> But in short:
> 
>     * the chunker secret determines the cutting places of the chunker (and thus: the length of the chunks cut). So an attacker can not determine the set of chunk lengths for a known set of big files, because the attacker would use a different chunker secret than the attacked borg setup.
> 
>     * id generation: if we would just use a simple hash (like sha256) as chunk id, it would be bad, because then an attacker could backup a known set of small files (1 chunk each) and the chunkids would directly show the hashes of these files, so the attacker could see which of these known files are in the attacked borg setup. but we use hmac-sha256 and thus the secret random MAC key also influences the id, so that this fingerprinting attack is not possible.

We should add a comprehensive explanation of **why** this feature is important to <https://borgbackup.readthedocs.io/en/latest/internals/security.html>. What is the attacker model? What would happen if we did not implement this? How does this feature thwart the attacker?

ReethuVinta:
Fixes #6268


ReethuVinta:
For #6278

docs/internals/security.rst
```diff
@@ -445,6 +445,18 @@ To summarize, this is making size-based fingerprinting difficult:
...
+
+Borg uses secret key for Chunking and ID generation. This key is used to enhance privacy and to protect 
+from an attacker who has access to a borg setup and encrypted chunks. The chunk's cutting positions are 
+determined by using the secret key and thereby the length of the chunks cut. Because the 
+attacker would use a different chunker secret than the borg setup being attacked, they would not be able 
+to determine the set of chunk lengths for a known set of files. For the chunk ID generation, simply using 
+the hased values of these files is insecure. Secret key is combined with chunk's hashed value to generate chunk ID.
+So, the attacker can't use known small files' hashes to get chunk IDs and thereby the contents of the borg setup.
+Thus, using SHA256 and a random MAC key makes fingerprinting attack impossible.
+
```
ThomasWaldmann, line 459: can you proofread this again, please?

ReethuVinta: Done. Can you please confirm and close this issue?

docs/internals/security.rst
```diff
@@ -445,6 +445,19 @@ To summarize, this is making size-based fingerprinting difficult:
...
+Borg uses the secret key for Chunking and ID generation. This key is used to enhance privacy and 
+to protect from an attacker who has access to a borg setup and encrypted chunks. The chunk's cutting 
+positions are determined by using the secret key and thereby the length of the chunks cut. Because 
+the attacker would use a different chunker secret than the borg setup being attacked, they would not 
+be able to determine the set of chunk lengths for a known set of files. For the chunk ID generation, 
+simply using the hashed values of these files is insecure. The secret key is combined with the chunk's 
+hashed value to generate chunk ID. So, the attacker can't use known small files' hashes to get chunk IDs 
+and thereby the contents of the borg setup. Thus, using SHA256 and a random MAC key makes fingerprinting 
+attacks impossible.
```
ThomasWaldmann, line 459: I'll add some fixes in a separate commit.

ThomasWaldmann: rebased onto current master.

ThomasWaldmann (approved):

ThomasWaldmann:
@ReethuVinta did you have a look? i'ld merge this if there is nothing to do left.

ReethuVinta:
Yeah, it's good to go.

PR #6350, opened 2022-02-24
docs: do not transform --/--- to unicode dashes (master)

hexagonrecursion:
* This is a forward port of #6348


ThomasWaldmann:
Thanks!

PR #6355, opened 2022-02-25
Add pyproject.toml

hexagonrecursion:
This blocks #6264 because black configuration is stored in pyproject.toml.
See <https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format>
I copied this from <https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use>


hexagonrecursion:
> Permission denied: '/usr/local/lib/python3.9

I am working on this.

hexagonrecursion:
https://github.com/pypa/setuptools/issues/3019

setup.py
```diff
@@ -19,6 +19,7 @@
 except ImportError:
     cythonize = None
 
+sys.path += [os.path.dirname(__file__)]
```
ThomasWaldmann, line 22: so I assume it didn't work without that even if you invoked stuff from the toplevel dir?

usually one should rather try not having to modify sys.path.

ThomasWaldmann:
btw, for now please no bigger / fundamental changes to setup.py / setup.cfg.

setup.py
hexagonrecursion: Neither the CWD nor the parent of setup.py are on sys.path if we use setuptools.build_meta. This is, apparently, a feature. Either we append `os.path.dirname(__file__)` or we use `setuptools.build_meta:__legacy__`, which will do the same for us.

https://github.com/pypa/setuptools/discussions/3134#discussioncomment-2249652

> The default backend (setuptools.build_meta) will not add the project folder to the path on purpose.
> That is only supported in `setuptools.build_meta:__legacy__`

ThomasWaldmann: ok, so keep it with line 22. hopefully this won't shoot ourselves in the knee at some time.

hexagonrecursion:
Another workaround would be to `pip install .` instead of `pip install -e .`

ThomasWaldmann (approved):
LGTM! Thanks!

PR #6356, opened 2022-02-25
s/apt/apt-get/

hexagonrecursion:
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.


ThomasWaldmann:
Thanks.

hexagonrecursion:
There is nothing to backport to 1.1 - there is no file named `.github/workflows/codeql-analysis.yml` in 1.1

PR #6367, opened 2022-02-26
[1.2-maint] src/borg/compress.pyx: fix compiler warning, closes #6365

### LINKED ISSUE #6365: compress.pyx compiler warning
I found this compiler warning from the generated code:
```
src/borg/compress.c: In function ‘__pyx_pf_4borg_8compress_4ZSTD_2_decide’:
src/borg/compress.c:6017:31: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
 6017 |   __pyx_t_2 = ((__pyx_v_osize < __pyx_v_isize) != 0);
      |                               ^
```
I was reviewing the source code and I noted that the size integers in the ZSTD section are all over the place, differing not just in signedness but also size. It appears there may be more going on here than just a signedness mismatch. Do you have any preferences for how this is resolved?

braewoods:
This resolves a compiler warning from the generated code that
resulted from a comparison of two local variables of different
signedness. The issue is resolved by changing the type of both
to int since this seems like the safest choice available.

### DIFF
--- a/src/borg/compress.pyx
+++ b/src/borg/compress.pyx
@@ -286,7 +286,7 @@ class ZSTD(DecidingCompressor):
         if not isinstance(idata, bytes):
             idata = bytes(idata)  # code below does not work with memoryview
         cdef int isize = len(idata)
-        cdef size_t osize
+        cdef int osize
         cdef char *source = idata
         cdef char *dest
         cdef int level = self.level


ThomasWaldmann:
Thanks!

PR #6368, opened 2022-02-26
fix multiple warnings related to _hashindex.c

### LINKED ISSUE #6333: _hashindex.c compiler warnings
mostly harmless, but still annoying.

check if we can get rid of these by adding some casts.

braewoods:
This should close #6333 as it silences all the warnings. I put it in 2 commits because there were effectively 2 sets of warnings.

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -118,7 +118,7 @@ static void hashindex_write(HashIndex *index, PyObject *file_py);
 static uint64_t hashindex_compact(HashIndex *index);
 static HashIndex *hashindex_init(int capacity, int key_size, int value_size);
 static const unsigned char *hashindex_get(HashIndex *index, const unsigned char *key);
-static int hashindex_set(HashIndex *index, const unsigned char *key, const unsigned char *value);
+static int hashindex_set(HashIndex *index, const unsigned char *key, const void *value);
 static int hashindex_delete(HashIndex *index, const unsigned char *key);
 static unsigned char *hashindex_next_key(HashIndex *index, const unsigned char *key);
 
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
thanks!

braewoods:
It's available for 1.2-maint.

hexagonrecursion:
* _hashindex.c warnings were introduced in #4127, but it was never backported. Maybe it should be.

Void pointer arithmetic:
https://github.com/borgbackup/borg/blob/a146f112282baf08c76ba3af5c348087806ae29b/src/borg/_hashindex.c#L163-L174

hexagonrecursion:
src/borg/cache_sync/unpack.h fix is not necessary for 1.1 because `hashindex_*` functions operate on `const void *key`

https://github.com/borgbackup/borg/blob/a146f112282baf08c76ba3af5c348087806ae29b/src/borg/_hashindex.c#L96-L99

ThomasWaldmann:
The problem seems to be that void pointer arithmetic is not standardized and some compilers don't do it.
gcc does it, so we won't notice the problem on linux.

So while we can use `void *` in the api, we need to convert to `char *` or `unsigned char *` before doing arithmetic on the pointer, because `sizeof(char) == 1` everywhere.

braewoods:
You are correct, you're not supposed to be able to perform pointer arithmetic on type void because it has no defined size.

hexagonrecursion:
Should I backport the relevant parts of <https://github.com/borgbackup/borg/pull/4127>?

ThomasWaldmann:
We don't need windows / MSC compatibility in 1.1-maint, but if we could reduce compiler warnings, a careful backport of misc. PRs related to that would be nice.

hexagonrecursion:
> reduce compiler warnings

As far as I can tell 1.1-maint already builds without warnings. I think it is not necessary to backport this to 1.1 for reasons I have outlined above.

ThomasWaldmann:
OK, fine, thanks for checking.

Issue #6316, opened 2022-02-19
get rid of most bundled 3rd party code


ThomasWaldmann:
considering that our version requirements aren't that high, i guess we can remove the bundled code and require that respective libs / headers are present on the system.

see `src/borg/algorithms`
- lz4
- zstd
- xxhash
- not sure about the crc32 stuff there, can we get it from somewhere and not bundle it?

this will also enable us to simplify `setup*.py` significantly.

FelixSchwarz:
just wanted to mention that Fedora probably needs a way to bundle msgpack. In Fedora packages often get minor version updates even within a stable release and borg is pretty picky about usable msgpack versions.

If the situation arise again (like the one we had with msgpack 1.0 vs 0.5.6) we might have to resort to bundling msgpack again. Of course we can carry the patches ourself but it would be helpful if the code structure to do so would be left in place. (yes, I know that borgbackup 1.2/master never bundled msgpack)

ThomasWaldmann:
As we have `borg.helpers.msgpack` now in borg 1.2 as a wrapper around msgpack, we are a bit more flexible and can adapt to some degree to msgpack changes (as long as it is not totally broken).

The msgpack-bundling we have in 1.1-maint is not present in 1.2 and I would prefer to keep it like that.

Of course, if nothing else helped, we could bundle it again, but guess that would be the last option.

It would be nice if distributions would not "upgrade" to broken msgpack versions and cause unnecessary issues.
Maybe check borg testsuite before upgrading msgpack? :-)

FelixSchwarz:
> The msgpack-bundling we have in 1.1-maint is not present in 1.2 and I would prefer to keep it like that.

me too.

> It would be nice if distributions would not "upgrade" to broken msgpack versions and cause unnecessary issues. Maybe check borg testsuite before upgrading msgpack? :-)

Currently maintenance of the msgpack package done by other packagers who were not as cooperative as I'd wished (though that was also before I was main packager for borgbackup and the previous borg packager was unresponsive). However I should probably talk to the current maintainers, maybe I can stay in the loop for msgpack updates in Fedora :-)

ThomasWaldmann:
crc32 stuff:
- libcrcutil - hw accelerated and fast sw crc32/crc64 implementations
- python3-crc32c - also has hw accelerated and pure sw implementations

FelixSchwarz:
[...]

PR #6378, opened 2022-02-27
suppress -Wimplicit-fallthrough warning

braewoods:
These instances of implicit switch case fallthrough appear to be
intentional. Add comments that the compiler understands to suppress
the false positive warning.

### DIFF
--- a/src/borg/cache_sync/unpack_template.h
+++ b/src/borg/cache_sync/unpack_template.h
@@ -196,7 +196,7 @@ static inline int unpack_execute(unpack_context* ctx, const char* data, size_t l
 
 
         _fixed_trail_again:
-            ++p;
+            ++p; // fallthrough (suppresses -Wimplicit-fallthrough)
 
         default:
             if((size_t)(pe - p) < trail) { goto _out; }


ThomasWaldmann:
Hmm, guess this is 3rd party code what we have there?

While we are able to "fix" it locally, it might be worth it checking if there is a more recent / fixed upstream version.

In #6316 i looked for alternatives, but for the crc32 stuff, didn't find any yet.

braewoods:
What crc32 algorithm is Borg currently using? I know zlib bundles a crc32 algorithm we could use and it's already a common dependency.

braewoods:
I have removed the third party code modifications to my commit since it sounds like you want to solve it via replacing the code entirely.

braewoods:
After reviewing the third party code, I now realize it is doing something special with SIMD to achieve faster crc32 than generic C code alone can achieve. Zlib's implementation is just a portable C crc32 with no efforts to enhance it. It may work the same but it won't perform the same.

ThomasWaldmann:
Yeah, just using zlib.crc32 is no option, we could even have that via python stdlib.

borg repository code is doing a ton of crc32 computation (== over all data it stores / verifies), so we shouldn't make that slower.

ThomasWaldmann:
Thanks for the PR! Does it also apply to 1.2-maint and 1.1-maint?

ThomasWaldmann:
Another comment about crc32: due to my recent experiments I noticed that on macOS zlib.crc32 is rather fast, but on linux it is super slow. So for master, I'll try to replace borg's crc32 code by zlib and deflate lib...

PR #6388, opened 2022-03-02
borg benchmark cpu: display benchmarks of cpu bound stuff

### LINKED ISSUE #6233: implement borg benchmark cpu

Yes

ISSUE

# Describe the problem you're observing

https://borgbackup.readthedocs.io/en/master/usage/init.html says that `repokey-blake2` my be faster than `repokey` on some clients, but slower on others.

Instead of trying to give users hints on which mode might be faster we should show users how to compare the two modes on **their** hardware. Trusting advice given in the docs instead of doing your own experiments is a cargo cult (and we should discourage that!).

It looks like we already have [`borg benchmark crud`](https://borgbackup.readthedocs.io/en/master/usage/benchmark.html) that could be used for this purpose - someone just needs to figure out how and document that.

ThomasWaldmann:
fixes #6233.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -537,6 +537,102 @@ class Archiver:
 
         return 0
 
+    def do_benchmark_cpu(self, args):
+        """Benchmark CPU bound operations."""
+        from timeit import timeit
+        random_10M = os.urandom(10*1000*1000)
+        key_256 = os.urandom(32)
+        key_128 = os.urandom(16)
+        key_96 = os.urandom(12)
+
+        import io
+        from borg.chunker import get_chunker
+        print("Chunkers =======================================================")
+        size = "1GB"
+
+        def chunkit(chunker_name, *args, **kwargs):
+            with io.BytesIO(random_10M) as data_file:
+                ch = get_chunker(chunker_name, *args, **kwargs)
+                for _ in ch.chunkify(fd=data_file):
+                    pass
+
+        for spec, func in [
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
```
$ borg benchmark cpu

Non-cryptographic checksums / hashes ===========================
crc32        1GB        0.722s
xxh64        1GB        0.122s
Cryptographic hashes / MACs ====================================
hmac-sha256  1GB        0.680s
blake2b-256  1GB        2.421s
Compression ====================================================
lz4          0.1GB      0.010s
zstd,1       0.1GB      0.027s
zstd,3       0.1GB      0.031s
zlib,0       0.1GB      0.026s
Chunkers =======================================================
buzhash,19,23,21,4095  1GB        1.539s
fixed,1048576          1GB        0.036s
Encryption =====================================================
aes-256-ctr-hmac-sha256 1GB        1.143s
aes-256-ctr-blake2b    1GB        2.883s
aes-256-ocb            1GB        0.949s
```

FabioPedretti:
Great!
Rather than using a fixed size of 1GB what about running every test for a fixed amount of time?
It would provide better precision for tests using a very small amount of time.

ThomasWaldmann:
@FabioPedretti Guess it was easier like that. :-)

ThomasWaldmann:
Just want to add an EPYC benchmark result:
```
aes-256-ocb              1GB        0.185s          ==     >5GB/s!
chacha20-poly1305        1GB        0.434s
```

PR #6428, opened 2022-03-09
Allow extra compiler flags for every extension build

braewoods:
This is mainly intended for explicit warnings but it can be
used for other flags as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,13 @@ extras_require = {
     'nofuse': [],
 }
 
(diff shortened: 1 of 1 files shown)


braewoods:
As previously discussed this enables compiler warnings for all extensions in the setup script itself. This includes `-Wpointer-arith`.

ThomasWaldmann:
Thanks!

hexagonrecursion:
1.1 backport will be harder. I'll do it tomorrow.

PR #6455, opened 2022-03-14
delete: add repository id and location to prompt

Whissi:
This commit will change prompt of `borg delete` command from
```
# borg delete
You requested to completely DELETE the repository *including* all archives it contains:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
```
to
```
# borg delete
You requested to completely DELETE the repository with ID 2c91b9eee2c718f766dcec336fa4c25af0568714110df36b5972b24a551a2310 located in ssh://username@borgbackup.example.net/path/to/repo *including* all archives it contains:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
Type 'YES' if you understand this and want to continue: 
```

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1392,23 +1392,38 @@ class Archiver:
 
         if not args.cache_only:
             if args.forced == 0:  # without --force, we let the user see the archives list and confirm.
+                id = bin_to_hex(repository.id)
+                location = repository._location.canonical_path()
                 msg = []
                 try:
                     manifest, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK)
+                    n_archives = len(manifest.archives)
+                    msg.append(f"You requested to completely DELETE the following repository "
+                               f"*including* {n_archives} archives it contains:")
                 except NoManifestError:
-                    msg.append("You requested to completely DELETE the repository *including* all archives it may "
-                               "contain.")
-                    msg.append("This repository seems to have no manifest, so we can't tell anything about its "
-                               "contents.")
-                else:
-                    if self.output_list:
-                        msg.append("You requested to completely DELETE the repository *including* all archives it "
-                                   "contains:")
-                        for archive_info in manifest.archives.list(sort_by=['ts']):
-                            msg.append(format_archive(archive_info))
+                    n_archives = None
+                    msg.append("You requested to completely DELETE the following repository "
+                               "*including* all archives it may contain:")
+
+                msg.append(DASHES)
+                msg.append(f"Repository ID: {id}")
+                msg.append(f"Location: {location}")
+
+                if self.output_list:
+                    msg.append("")
+                    msg.append("Archives:")
+
+                    if n_archives is not None:
+                        if n_archives > 0:
+                            for archive_info in manifest.archives.list(sort_by=['ts']):
+                                msg.append(format_archive(archive_info))
+                        else:
+                            msg.append("This repository seems to not have any archives.")
                     else:
-                        msg.append("You requested to completely DELETE the repository *including* %d archives it contains."
-                                   % len(manifest.archives))
+                        msg.append("This repository seems to have no manifest, so we can't "
+                                   "tell anything about its contents.")
+
+                msg.append(DASHES)
                 msg.append("Type 'YES' if you understand this and want to continue: ")
                 msg = '\n'.join(msg)
                 if not yes(msg, false_msg="Aborting.", invalid_msg='Invalid answer, aborting.', truish=('YES',),


ThomasWaldmann (commented):
Thanks for the PR!

As we're in master branch here (requiring py39+), you could use python f-strings (the code was recently upgraded to that globally).

Also, maybe first mention the location (which is usually recognized better by human users) and the ID after that.

Whissi:
OK, what do you prefer:

1. > You requested to completely DELETE the repository with ID 2c91b9eee2c718f766dcec336fa4c25af0568714110df36b5972b24a551a2310 located in ssh://username@borgbackup.example.net/path/to/repo *including* all archives it contains:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
Type 'YES' if you understand this and want to continue: 
2. > You requested to completely DELETE the repository located in ssh://username@borgbackup.example.net/path/to/repo with ID 2c91b9eee2c718f766dcec336fa4c25af0568714110df36b5972b24a551a2310 *including* all archives it contains:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
Type 'YES' if you understand this and want to continue: 
3. > You requested to completely DELETE the repository (ssh://username@borgbackup.example.net/path/to/repo) [2c91b9eee2c718f766dcec336fa4c25af0568714110df36b5972b24a551a2310] *including* all archives it contains:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
Type 'YES' if you understand this and want to continue: 

I prefer 3. But maybe you have an even better idea?

ThomasWaldmann:
Could be also more formatted / less in one sentence:
```
You requested to completely DELETE this repository including all archives it contains.
Location: ssh://username@borgbackup.example.net/path/to/repo
ID: 2c91b9eee2c718f766dcec336fa4c25af0568714110df36b5972b24a551a2310
Archives:
host1.example.net-2022-03-14T14:55:31 Mon, 2022-03-14 15:55:31 [ecb4e323c38e5d560452192456073f284aa3acaf59b53e24f584ae33e3025519]
Type 'YES' if you understand this and want to continue:
```

(for the upper part, use same format as seen in borg info)

Whissi:
OK, I also included dashes and an empty line for a cleaner look. Are you OK with
```
# borg delete
You requested to completely DELETE the following repository *including* all archives it contains:
Repository ID: 793cb3842d5d59f3a10277d33edb26a8378b2f408dcae4effde707476d216f98
Location: ssh://user@borgbackup.example.net/path/to/repo

Archives:
host1.example.net-2022-03-14T15:32:24 Mon, 2022-03-14 16:32:26 [74a40e23e64a5d9bdc015688f2c3a5b12f1243527823a6a5faf02159126733e4]
host1.example.net-2022-03-14T23:54:48 Tue, 2022-03-15 00:54:51 [602e9d6bffa03a84fb0a824c7a0cebc2674b665ecd853a02550fd4cab9c1d330]
Type 'YES' if you understand this and want to continue:
```

ThomasWaldmann:
Ah, that looks pretty! Just a small nitpick: 1 dash more so it at least matches the (fixed) line length of line with the repo id.

Whissi:
Mh, the code is using the [DASHES variable](https://github.com/borgbackup/borg/blob/c641e31e82548798688c58f620f20a07c28dc5ba/src/borg/constants.py#L102). Should we increase value in general? Just add one additional dash to this specific output? Or leave it as is?

ThomasWaldmann:
ok, guess that would work, but there is a lot of duplication in the code now.

Whissi:
Yes, I can refactor the function a little bit to reduce code duplication.

src/borg/archiver.py
```diff
@@ -1392,23 +1392,37 @@ def _delete_repository(self, args, repository):
...
             if args.forced == 0:  # without --force, we let the user see the archives list and confirm.
+                id = bin_to_hex(repository.id)
+                location = repository._location.canonical_path()
                 msg = []
                 try:
                     manifest, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK)
+                    n_archives = len(manifest.archives)
+                    msg.append(f"You requested to completely DELETE the following repository "
+                               f"*including* {n_archives} archives it contains:")
                 except NoManifestError:
```
ThomasWaldmann, line 1403: if `Manifest.load()` raises this exception, `manifest` will be undefined.

src/borg/archiver.py
```diff
@@ -1392,23 +1392,37 @@ def _delete_repository(self, args, repository):
...
+
+                msg.append(DASHES)
+                msg.append(f"Repository ID: {id}")
+                msg.append(f"Location: {location}")
+
+                if self.output_list:
+                    msg.append("")
+                    msg.append("Archives:")
+
+                    if manifest:
```
ThomasWaldmann, line 1415: and if it is undefined, it will crash here.

src/borg/archiver.py
Whissi: Yes, if that will fail, we will land in exception handling, i.e.
```python
                    n_archives = len(manifest.archives)
                    msg.append(f"You requested to completely DELETE the following repository "
                               f"*including* {n_archives} archives it contains:")
```
won't get executed anymore. Am I wrong?

src/borg/archiver.py
Whissi: ...so if manifest is undefined, that if clause should get skipped, not?

Whissi:
OK, I now initialized `n_archives` as `None` and check this var for `None` later instead of using `manifest` variable. Works for me when I add `raise NoManifestError` for testing before `Manifest.load()`. Is this OK for you? Another idea would be using `if 'manifest' in locals():` but I think it isn't good practice in Python to check for variable existence.

src/borg/archiver.py
ThomasWaldmann: correct.

you could also move that `n_archives = None` to the exception handler block.

src/borg/archiver.py
```diff
@@ -1392,23 +1392,38 @@ def _delete_repository(self, args, repository):
...
+                msg.append(DASHES)
+                msg.append(f"Repository ID: {id}")
+                msg.append(f"Location: {location}")
+
+                if self.output_list:
+                    msg.append("")
+                    msg.append("Archives:")
+
+                    if n_archives is not None:
+                        if manifest.archives:
```
ThomasWaldmann, line 1417: as you've determined `n_archives` already, you could also use it here, like `if n_archives > 0:`.

ThomasWaldmann (commented):
some minor nitpicks, but guess we can merge after these.

src/borg/archiver.py
Whissi: I think this is not possible because we are checking 3 states here:

- If manifest is not available, we tell user that we were unable to check.
- If we have manifest and therefore `n_archives >= 0` we want the output with `Archives:`
- ...but in case there aren't any archives, display a friendly message indicating that the missing listing is not an application error -- there just aren't any archives to list.

ThomasWaldmann:
Thanks! Merged.

Guess we want this in 1.2-maint also (maybe even 1.1-maint), would you like to do the backports also @Whissi?

PR #6463, opened 2022-03-17
new AEAD crypto with session keys

### LINKED ISSUE #1031: add flexible encryption chunk header

### LINKED ISSUE #1044: Crypto roadmap
Coordination and high-level discussion ticket
  - [ ] "integrate AEAD-style crypto code" (this also adds AES-GCM crypto code, which would be unused at first). At first, only the legacy-compatible mode of the code would be used.
  - [ ] Implement _key-agility_ and _algorithm-agility_ via DEKs
    - [x] Discussion in #1031 and linked spec draft
    - [x] Low-level AEAD-style abstractions in #1034 (with AES-CTR compatible to older borg repos and AES-GCM)
  - [x] require OpenSSL >= 1.1.1 (via #1034)
    - [x] AES-256-OCB? 25-50 % faster than AES-GCM (with _and_ without AES-NI). Limited data/key is not a problem since we'll have key agility at this point (also, GCM has the exact same problem and is pretty much the same code-wise; just share the code!) [License: Borg is eligible, no explicit licensing needed, since we're OSI] - Ie. OCB is pretty much the same thing for us as GCM, just faster.
    - [x] Chacha20-Poly1305 - is great because it's AEAD crypto that's faster than any AES-AEAD mode in software.

Note: edited / updated 2022-02 by TW

### LINKED ISSUE #3814: Use randomized key derivation instead of relying on stateful nonces

Yes

Bug

N/A

Borg currently uses the same AES key for all encryption.  This is an anti-pattern and leads to Borg’s weaknesses when it comes to nonces.

A much better solution is to generate a fresh encryption key every time as (say) `blake2b(my_personality_string, random_number, key)`.  This has the following advantages:

* There is no reliance on storing state on the server, since keys are generated at random each time.  We can instead use a randomly generated nonce.
* We can use AEAD ciphers, such as AES-GCM or ChaCha20-Poly1305, which have formal security reductions to the security of the underlying cipher.

ThomasWaldmann:
- AEAD crypto (AES-OCB and chacha20-poly1305) fixes and updates
- new keytypes
  - for OCB/chacha * keyfile/repokey * hmac-sha256/blake2b
  - a new random 192bit sessionid is generated for each borg invokation
  - to encrypt new data, a new session key is derived from static key material and sessionid, 48bit IVs start at 0 (1 to be precise) for each session.
  - thus, no global or persistent IV counter management needed for these modes
  - all data in the output is authenticated now (complete header + chunkid + encrypted payload) via a 128bit auth tag from the AEAD cipher
  - AES-OCB is super fast and single-pass on hw with acceleration
  - chacha20-poly1305 is a very fast pure software cipher
- misc. code cleanups 
- docs updates

fixes #3814, fixes #1031, fixes #1044. also related to #45.

borg-crypto


src/borg/crypto/low_level.pyx
```diff
@@ -549,29 +551,24 @@ cdef class _AEAD_BASE:
...
+                                   <const unsigned char*> idata.buf+hlen+self.mac_len,
+                                   ilen-hlen-self.mac_len)
             if not rc:
                 raise CryptoError('EVP_DecryptUpdate failed')
             offset += olen
             rc = EVP_DecryptFinal_ex(self.ctx, odata+offset, &olen)
-            if rc <= 0:
+            if not rc:
                 # a failure here means corrupted or tampered tag (mac) or data.
                 raise IntegrityError('Authentication / EVP_DecryptFinal_ex failed')
```
ThomasWaldmann, line 573:
```
borg init -e repokey-aes-ocb repo
borg list repo
```
crashes here - does somebody see why?

ThomasWaldmann: (it is not due to that change, crashes with both conditions)

ThomasWaldmann: see commit "avoid losing the key"

src/borg/constants.py
```diff
@@ -123,6 +123,11 @@ class KeyType:
     BLAKE2REPO = 0x05
     BLAKE2AUTHENTICATED = 0x06
     AUTHENTICATED = 0x07
+    # new crypto starts at 0xa0
+    AESOCBKEYFILE = 0xa0
+    AESOCBREPO = 0xa1
+    CHPOKEYFILE = 0xa2
+    CHPOREPO = 0xa3
```
ThomasWaldmann, line 130: I dislike how this grows like `n * m * o` (and I did not yet implement blake2b id hashing here, for that i would've needed 8 keytypes already).
n == amount of key storage places (currently 2)
m == amount of id hashing algorithms (currently 2, hmac-sha256 and blake2b)
o == amount of aead ciphersuites (now 3, aes-ctr-hmac-sha256 and 2 new aes-ocb and chacha20-poly1305)

in the future, guess m could grow due to blake3.

src/borg/crypto/key.py
```diff
@@ -704,8 +706,131 @@ class Blake2AuthenticatedKey(ID_BLAKE2b_256, AuthenticatedKeyBase):
...
+    CIPHERSUITE = CHACHA20_POLY1305
+
+
+class CHPORepoKey(ID_HMAC_SHA_256, AEADKeyBase, FlexiKey):
+    TYPES_ACCEPTABLE = {KeyType.CHPOKEYFILE, KeyType.CHPOREPO}
+    TYPE = KeyType.CHPOREPO
+    NAME = 'repokey ChaCha20-Poly1305'
+    ARG_NAME = 'repokey-chacha20-poly1305'
+    STORAGE = KeyBlobStorage.REPO
+    CIPHERSUITE = CHACHA20_POLY1305
```
ThomasWaldmann, line 826: see my growth comment. here: 4 new classes, 4 (blake2b) still missing.

src/borg/crypto/low_level.pyx
```diff
@@ -184,10 +184,10 @@ class UNENCRYPTED:
 
 
 cdef class AES256_CTR_BASE:
-    # Layout: HEADER + MAC 32 + IV 8 + CT (same as attic / borg < 1.2 IF HEADER = TYPE_BYTE, no AAD)
+    # Layout: HEADER + MAC 32 + IV 8 + CT (same as attic / borg < 1.3 IF HEADER = TYPE_BYTE, no AAD)
 
     cdef EVP_CIPHER_CTX *ctx
-    cdef unsigned char *enc_key
+    cdef unsigned char enc_key[32]
```
ThomasWaldmann, line 190: we were lucky that worked all the time, see "avoid losing the key" commit.

src/borg/crypto/low_level.pyx
```diff
@@ -360,7 +359,7 @@ cdef class AES256_CTR_BASE:
 
 
 cdef class AES256_CTR_HMAC_SHA256(AES256_CTR_BASE):
-    cdef unsigned char *mac_key
+    cdef unsigned char mac_key[32]
```
ThomasWaldmann, line 362: more luck.

src/borg/crypto/low_level.pyx
```diff
@@ -390,7 +389,7 @@ cdef class AES256_CTR_HMAC_SHA256(AES256_CTR_BASE):
 
 
 cdef class AES256_CTR_BLAKE2b(AES256_CTR_BASE):
-    cdef unsigned char *mac_key
+    cdef unsigned char mac_key[128]
```
ThomasWaldmann, line 392: more luck.

src/borg/crypto/low_level.pyx
```diff
@@ -644,17 +627,17 @@ cdef class CHACHA20_POLY1305(_CHACHA_BASE):
...
-        super().__init__(mac_key, enc_key, iv=iv, header_len=header_len, aad_offset=aad_offset)
+        super().__init__(key, iv=iv, header_len=header_len, aad_offset=aad_offset)
 
 
 cdef class AES:
     """A thin wrapper around the OpenSSL EVP cipher API - for legacy code, like key file encryption"""
     cdef CIPHER cipher
     cdef EVP_CIPHER_CTX *ctx
-    cdef unsigned char *enc_key
+    cdef unsigned char enc_key[32]
```
ThomasWaldmann, line 640: more luck.

ThomasWaldmann:
@enkore @textshell "it works", thus ready for some early feedback, if you have some time.

src/borg/crypto/low_level.pyx
ThomasWaldmann: i filed an issue to fix this also in the other branches, so i can close this here.

src/borg/constants.py
```diff
@@ -123,6 +123,15 @@ class KeyType:
...
+    BLAKE2AESOCBKEYFILE = 0xa4
+    BLAKE2AESOCBREPO = 0xa5
+    BLAKE2CHPOKEYFILE = 0xa6
+    BLAKE2CHPOREPO = 0xa7
```
ThomasWaldmann, line 134: I dislike how this grows like `n * m * o` (and I did not yet implement blake2b id hashing here, for that i would've needed 8 keytypes already).
n == amount of key storage places (currently 2)
m == amount of id hashing algorithms (currently 2, hmac-sha256 and blake2b)
o == amount of aead ciphersuites (now 3, aes-ctr-hmac-sha256 and 2 new aes-ocb and chacha20-poly1305)

in the future, guess m could grow due to blake3.

src/borg/testsuite/crypto.py
```diff
@@ -111,35 +110,34 @@ def test_AE(self):
...
     def test_AEAD(self):
         # test with aad
-        mac_key = None
-        enc_key = b'X' * 32
-        iv = 0
+        key = b'X' * 32
+        iv_int = 0
         data = b'foo' * 10
-        header = b'\x12\x34\x56'
+        header = b'\x12\x34\x56' + iv_int.to_bytes(12, 'big')
```
ThomasWaldmann, line 140: the old code used to append the IV to the header (and fed it all into the auth tag computation).

to keep our test vectors working, the caller now has to append the iv to the header as the code does not do that any more.

ThomasWaldmann:
@enkore I guess I will update docs / diagram to remove that 4bit:4bit split ciphersuite and keytype again.

The current code and also the code in this PR does not work like that and I also see a quite near-future "running out of bits" in the keytype when done that way.

The current code does not have ciphersuite and just dispatches a 8bit keytype to old and new Key classes.
The stuff starting at keytype $00 is the old stuff (pre borg 1.3) and the borg 1.3 stuff currently starts at $a0.
Without that split, we have 96 keytypes ($a0 .. $ff) there, with the 4:4 split we would just have 16.

Note: I chose $a0 vs. $00 to have a 2bit difference in the upper bits (thinking about bitflips). Not sure if that is really necessary. As we likely will not add anything to the old stuff, we could also start the new stuff at e.g. $10 or so.

ThomasWaldmann:
Alternatively, keep the ciphersuite:4 keytype:4 and do this:

ciphersuite 0 == legacy modes, "as is" (in the past, we used $00 .. $07, so this fits)
ciphersuite 1  == AES-OCB + hmac-sha256 based keytypes (currently 2 storage types, 14 free)
ciphersuite 2  == AES-OCB + blake2b based keytypes (currently 2 storage types, 14 free)
ciphersuite 3 == chacha20-poly1305 + hmac-sha256 based keytypes (currently 2 storage types, 14 free)
ciphersuite 4 == chacha20-poly1305 + blake2b based keytypes (currently 2 storage types, 14 free)
ciphersuite 5/6 == * + blake3 (later) 
...

ThomasWaldmann:
Looks like there is no ongoing review (although I asked around), so I'll merge this after tests finish.

If you find something after the merge, just file an issue on the issue tracker.

ThomasWaldmann:
Note: i did not change the dispatching yet, so it still works using the full type byte (not the 4bits suite + 4 bits keytype). guess the dispatching can be changed more easily after we completely remove the legacy crypto.

ThomasWaldmann:
@hexagonrecursion merging this will likely cause some merge conflicts for your ongoing PR - sorry about that.

i can offer increasing the bounty for the additional work or helping rebasing what you have now (whatever you prefer).

hexagonrecursion:
I prefer to rebase my PR myself.

py0xc3:
As proposed, my analysis, which is focused on the use of ChaCha20-Poly1305.

Within one session, are all chunks encrypted using the same invocation of the AEAD? Or is there a separated invocation of AEAD for each chunk?

If it is the case assumed in the first question (1 session : 1 AEAD invocation), the CTR seems to not fulfill a cryptographic purpose as the salt of the HKDF already ensures that the AEAD key input changes on each invocation (if 1 HKDF output : 1 AEAD invocation). It is not realistic that a 192 bit salt repeats (if there is a proper cryptographic PRNG). However, the CTR does not create an issue.

If it is the second case (separated invocation per chunk / multiple AEAD invocations per session), I have a somewhat ambivalent view of the CTR. I assume 2^48 counters with at least 512K chunk each seems sufficient for a long time for average users, being able to encrypt 134.217.728 TB before the counter starts again at 0. 

But in the second case, it might be an easy-to-achieve goal to consider every potential use case in Borg's design, so that it can be easily forked/adjusted to, e.g., enterprise use cases. Storing amounts like 134.217.728 TB is already today the dimension of some multinational corporations, and the amounts keep increasing exponentially (a related [2016 paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/44830.pdf); and when a new repo is created, one session includes all initial data at once :). The goal should be to ensure that no architect in any use case will need to consider the nonce (and the dangerous security issues it causes when it repeats with the same key) in the foreseeable future. Here the design could be a bit more future-proof and (user-)fault-tolerant.

My suggestion would be to increase the counter to at least 64 bit for the "normal" ChaCha20 ("normal" in order to separate it from the modification XChaCha, see below). [RFC 8439](https://datatracker.ietf.org/doc/html/rfc8439#section-2.3) suggests a 96 bit nonce but 32 bit are sender-specific and only 64 bit of the nonce are a counter (additional to the block counter, which is separated from the nonce); but the RFC also considers other use cases that might not be relevant here: 64 bit (as a 64 bit counter) for the nonce is appropriate if the "normal" ChaCha20 has to be used. 

But using a 96 bit counter as nonce (unlike the RFC), if it does not make a difference in the implementation, might be the best solution for the "normal" ChaCha20. The then reduced 32 bit block counter is more than sufficient if chunk sizes are >256GB (assuming 1 AEAD invocation : 1 chunk). However, it might be noted that XChaCha12/20 (as in Adiantum in the Linux kernel for disk encryption) is often preferred for use cases that have to be prepared for future-proof large data amounts. Nevertheless, I think the "normal" ChaCha20 is sufficient as well for the foreseeable future if the counter size is adjusted because the chunk sizes are much larger than block sizes of Linux's disk encryption with XChaCha.

Alternatively to the increased counter size, a new session could be automatically invoked if the counter is at its end. This should be implemented explicitly, and would split a very huge backup automatically into >1 increments due to >1 session.

If OFB needs a 48 bit IV, I expect a truncation might not impact the performance in a noteworthy manner.

Feel free to repeat my calculations to make sure I didn't make an error.

I hope this is somewhat helpful for this great project.

ThomasWaldmann:
The HKDF is used once per session to derive the session key (one invocation of e.g. `borg create` is one session).

The AEAD crypto is invoked once per chunk and usually there are many chunks processed in a session. The default target size for a chunk is 2MiB, but if the input file size is way smaller than that, it will also result in a separate chunk.

Due to the deduplication and for any non-first backup of some data set, the usual amount of **new** chunks (and only these will need to get encrypted) is relatively small though, because most chunks will be duplicate because they already exist in the repository.

ThomasWaldmann:
About counter size:

- I'ld expect that enterprise users do not put all their stuff in one repository, but use many of them. It's just way easier to deal with if a `borg check` is expected to finish within the admin's lifetime. :-) Also, the repo is usually locked while doing that.
- There are other limitations to consider. E.g. borg uses a in-memory chunks index and that needs about 50 bytes RAM per chunk. So if you have 10G chunks, you'ld already need 500GB of RAM. 10G chunks * 10KB per chunk = 100 TB. So, if your first backup would run over 100TB (which would take a really long time), it would produce 10G new chunks and 10G AEAD invocations. 10G fits in a 34 bit counter. 48 bit is 16384 times that (run time and also memory needs).

ThomasWaldmann:
Note: we use AES-OCB, not OFB.

ThomasWaldmann:
BTW, thanks for your review!

py0xc3:
> * I'ld expect that enterprise users do not put all their stuff in one repository, but use many of them. It's just way easier to deal with if a `borg check` is expected to finish within the admin's lifetime. :-) Also, the repo is usually locked while doing that.

I have seen worse solutions, even in larger use cases :) Think about how many easy-to-fix / legacy vulnerabilities have been used in the past to successfully attack large organizations/enterprises. Unfortunately, large organizations do not imply best practices or proper design stages before implementation, especially in non-IT organizations. Expectations can be dangerous, and failure in consequent/preceding stages of a component/system should be anticipated and balanced if possible (avoid single points of failure and such). In short, apply "secure by design" not just to code but also to the social/organizational environment in which it evolves/deploys :-). But you are right, my statement assumes a "worst practice" scenario. If an admin designs a backup system properly, my scenario is obsolete anyway.
 
> * There are other limitations to consider. E.g. borg uses a in-memory chunks index and that needs about 50 bytes RAM per chunk. So if you have 10G chunks, you'ld already need 500GB of RAM. 10G chunks * 10KB per chunk = 100 TB. So, if your first backup would run over 100TB (which would take a really long time), it would produce 10G new chunks and 10G AEAD invocations. 10G fits in a 34 bit counter. 48 bit is 16384 times that (run time and also memory needs).

As mentioned, my statement only intended to address large scale and long term scenarios, just in case Borg wants to consider them already now. I like to have design stages always consider every practically-implementable and foreseeable case. However, given the other limitations you mentioned, my statement is obviously obsolete in Borg's use case, and in the **overall architecture of Borg, there is consequently no incentive to adjust the counter** at all: the RAM usage you indicated is not realistic / not feasible for 2^48 chunks (>12,5 Petabyte RAM).

However, one question that goes beyond the scope of this PR but that is related to its security: what is/are the origin(s) of the secret(s) in HKDF's input (so, enc_key:256 & hmac_key:256)?

> Note: we use AES-OCB, not OFB.

My bad. Thinking error during writing; I recently had to read some stuff about the other one (seems to have impacted me more than expected :-)

ThomasWaldmann:
the ikm (enc_key + hmac_key) is coming from the borg repokey or keyfile. when creating that key, borg uses `os.urandom` to create the key material. initial they were used for the old crypto, thus the 2 separate keys for encryption and authentication.

docs/internals/security.rst
```diff
@@ -124,7 +124,88 @@ prompt is a set BORG_PASSPHRASE. See issue :issue:`2169` for details.
...
+
+Notable:
+
+- More modern and often faster AEAD ciphers instead of self-assembled stuff.
+- Due to the usage of session keys, IVs (nonces) do not need special care here as
+  they did for the legacy encryption modes.
+- The id is now also input into the authentication tag computation.
+  This strongly associates the id with the written data (== associates the key with
+  the value). When later reading the data for some id, authentication will only
+  succeed if what we get was really written by us for that id.
```
real-or-random, line 196: (Curious user here who happens to be a cryptographer.) 

What gains do you expect from this?

So if I read this correctly, then the chunk id is an additional MAC over the uncompressed and unencrypted data. But the AEAD already authenticates the unencrypted data (no matter whether it's encrypted or not), so I fail to see the purpose of adding the id to the AAD.

 Does the id have a purpose in other parts of borg or is this just for the encryption?

I may be able to provide a more detailed review but I realize I'm quite late to the party... Do you think this would still be helpful?

real-or-random: Ok, this makes a lot of sense now! Adding the chunkid as AAD binds the key (as in "key-value store") to the value.

> Thus, we do not need to compute the MAC to verify if the chunk is really for the chunkid we wanted - saves some CPU time.

Indeed yes, I see that this would save CPU time. But then I'm confused whether it's really implemented like this? AFAIU the code still recomputes the MAC:
https://github.com/borgbackup/borg/blob/b52781a9991be6f6c73411092e940627356cef36/src/borg/crypto/key.py#L897

> thanks for looking at our crypto, review there is very welcome (and not too late, borg2 is still in alpha, soon beta, so we could still change things)!

Thanks for the warm welcome. I'll keep looking at the PR and docs here. I'll probably continue to ask some very basic questions, so please bear with me. :)

real-or-random:
A more high-level question: Is there a reason why deterministic encryption / SIV mode wasn't used?

Deterministic encryption sounds scary on first glance but it fits the deduplication model of borg. We'd anyway never store the same chunk twice. Or in other words, even if we were to create multiple different ciphertexts for the same chunk, the attacker would notice as they're index by the chunk ID if I understand correctly.  

I see that has been discussed a few times, e.g., https://github.com/borgbackup/borg/issues/1044#issuecomment-733026989  and @ozppupbg reinvented it in https://github.com/borgbackup/borg/issues/3814#issuecomment-637669547. @demimarie commented that compression could be an issue  https://github.com/borgbackup/borg/issues/3814#issuecomment-646375499 but I'm pretty sure this could be solved by including the compression parameters in the AAD. (Actually, are they currently authenticated somewhere?)

(Do we want me to move the discussion to somewhere else, e.g., one these issues?)

ThomasWaldmann:
Guess the main spoiler was that SIV is not in openssl. Did that change?

Issue #747, opened 2016-03-14
Using a more modern KDF than PBKDF2


enkore:
With repokey the encrypted keys are stored in the repo itself. So in almost any attack scenario the attacker has access to that (And if we store a repo in essentially untrusted storage, e.g. "the cloud", we must assume that the repokey blob is essentially 'public knowledge'). Which makes me think that PBKDF2 (with it's susceptibility to GPU/FPGA and ASIC based attacks) might not be the best choice here.

Upgrading the repokey can be done safely and transparently to the user when accessing a repo. Increasing just the iterations of PBKDF2 would not make the problem go away.

There are some recent developments into key derivation functions which are very hard to speed up with GPUs or dedicated hardware, e.g. scrypt, argon2d/argon2i. These can be tuned to derive a key in a reasonable time frame on commodity hardware (0.1 s < t < 1 s) while off-line attacks remain essentially inefficient.

I think we are in a very favourable position here, since it is not an issue if the KDF takes one second per key.

EDIT: This of course also affects the other key storage methods, repokey just being the most handy of them.

ThomasWaldmann:
There was already a discussion about pbkdf2, iterations, etc. in another ticket.

One paper linked from there pointed out that increasing iterations is not best method to increase security, but one should increase passphrase length.

So, do you still think we need another kdf even if people choose long passphrases?

The problem with these other kdfs is that they often add another dependency. If we add a dependency that is not present in some linux distribution, borg can't be packaged for it (or is way more effort to package all the stuff needed). Even worse, if we add a conflicting requirement, it might even block packaging.

enkore:
> another ticket

I should have searched before creating this one: #77 

Yes, the dependency hell is an issue, especially since soft-dependencies are IMHO not an option for borg core functionality. Vendoring might be an option, since they're small, but meh. That's not a clean solution either.

I don't think that there is "immediate need for action". There is however a point to being "truly and completely paranoid safe than ~~sorry~~ just a little paranoid safe". Of course a very long, high entropy pass phrase is an absolute must, independent of KDF.

Specifically re. argon there are even two bindings floating around, both of which are maintained. Huh?

In summary: maybe increase them a little, or use calibration (on `borg create` / KeyfileKeyBase.create), but either one is more "nice to have". Other KDFs are more a long-term thing until they have hit common libraries and distros.

ThomasWaldmann:
Yes, tagging it "later". Maybe edit ticket title to "use more modern kdf?"?

[...]

PR #6468, opened 2022-03-19
Argon2 the first part: Implement key derivation (was: part 0)

hexagonrecursion:
This is the first part of #747. I have added a method to `helpers.passphrase.Passphrase` to derive keys via argon2.

I have chosen to add a new method instead of changing `helpers.passphrase.Passphrase.kdf because`:
1. argon2 has more tunable parameters
2. decrypt_key_file needs to be able to select pbkdf2 or argon2 based on key version
3. I have heard that it is usually more pythonic to define two functions than to define one that does two different things based on a flag

PS. After reading the code I have reversed my original decision and started implementing this bottom-up. I'll do [decrypt_key_file](https://github.com/borgbackup/borg/blob/dfd7ea8171a947b3fe8730ac8d2c189e803985e9/src/borg/crypto/key.py#L431) next
PPS. Sorry for the delay. I am back in your galaxy and ready to work on this today.

### DIFF
--- a/requirements.d/development.txt
+++ b/requirements.d/development.txt
@@ -11,3 +11,4 @@ pytest-benchmark
 Cython!=0.27
 twine
 python-dateutil
+argon2-cffi
--- a/requirements.d/development.txt
+++ b/requirements.d/development.txt
@@ -11,4 +11,3 @@ pytest-benchmark
 Cython!=0.27
 twine
 python-dateutil
-argon2-cffi
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -139,3 +139,6 @@ class Passphrase(str):
 
     def kdf(self, salt, iterations, length):
         return pbkdf2_hmac('sha256', self.encode('utf-8'), salt, iterations, length)
+
+    def argon2(*, salt, time_cost, memory_cost, parallelism, type):
+        return b'', b''
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -144,7 +144,7 @@ class Passphrase(str):
 
     def argon2(
         self,
-        *,
+        output_len_in_bytes: int,
         salt: bytes,
         time_cost,
         memory_cost,
@@ -153,11 +153,11 @@ class Passphrase(str):
     ) -> (bytes, bytes):
         key = argon2.low_level.hash_secret_raw(
             secret=self.encode("utf-8"),
-            hash_len=64,  # hash_len is in bytes
+            hash_len=output_len_in_bytes,
             salt=salt,
             time_cost=time_cost,
             memory_cost=memory_cost,
             parallelism=parallelism,
             type=type,
         )
-        return key[:32], key[32:]
+        return key
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -12,6 +12,8 @@ from . import prepare_subprocess_env
 
 from ..logger import create_logger
 
+import argon2.low_level
+
 logger = create_logger()
 
 
@@ -140,5 +142,22 @@ class Passphrase(str):
     def kdf(self, salt, iterations, length):
         return pbkdf2_hmac('sha256', self.encode('utf-8'), salt, iterations, length)
 
-    def argon2(*, salt, time_cost, memory_cost, parallelism, type):
-        return b'', b''
+    def argon2(
+        self,
+        *,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: argon2.low_level.Type
+    ) -> (bytes, bytes):
+        key = argon2.low_level.hash_secret_raw(
+            secret=self.encode("utf-8"),
+            hash_len=64,  # hash_len is in bytes
+            salt=salt,
+            time_cost=time_cost,
+            memory_cost=memory_cost,
+            parallelism=parallelism,
+            type=type,
+        )
+        return key[:32], key[32:]
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -4,6 +4,7 @@ import shlex
 import subprocess
 import sys
 from hashlib import pbkdf2_hmac
+from typing import Literal
 
 from . import bin_to_hex
 from . import Error
@@ -149,8 +150,13 @@ class Passphrase(str):
         time_cost,
         memory_cost,
         parallelism,
-        type: argon2.low_level.Type
+        type: Literal[b'argon2i', b'argon2d', b'argon2id']
     ) -> (bytes, bytes):
+        type_map = {
+            b'argon2i': argon2.low_level.Type.I,
+            b'argon2d': argon2.low_level.Type.D,
+            b'argon2id': argon2.low_level.Type.ID,
+        }
         key = argon2.low_level.hash_secret_raw(
             secret=self.encode("utf-8"),
             hash_len=output_len_in_bytes,
@@ -158,6 +164,6 @@ class Passphrase(str):
             time_cost=time_cost,
             memory_cost=memory_cost,
             parallelism=parallelism,
-            type=type,
+            type=type_map[type],
         )
         return key
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -151,7 +151,7 @@ class Passphrase(str):
         memory_cost,
         parallelism,
         type: Literal[b'argon2i', b'argon2d', b'argon2id']
-    ) -> (bytes, bytes):
+    ) -> bytes:
         type_map = {
             b'argon2i': argon2.low_level.Type.I,
             b'argon2d': argon2.low_level.Type.D,
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -150,12 +150,12 @@ class Passphrase(str):
         time_cost,
         memory_cost,
         parallelism,
-        type: Literal[b'argon2i', b'argon2d', b'argon2id']
+        type: Literal['i', 'd', 'id']
     ) -> bytes:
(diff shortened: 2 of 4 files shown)


requirements.d/development.txt
```diff
@@ -11,3 +11,4 @@ pytest-benchmark
 Cython!=0.27
 twine
 python-dateutil
+argon2-cffi
```
ThomasWaldmann, line 14: guess this rather should be in setup.py.

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +141,23 @@ def __repr__(self):
...
+        *,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: argon2.low_level.Type
+    ) -> (bytes, bytes):
+        key = argon2.low_level.hash_secret_raw(
+            secret=self.encode("utf-8"),
+            hash_len=64,  # hash_len is in bytes
```
ThomasWaldmann, line 156: kdf has a variable length parameter - do we want that here also?

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +141,23 @@ def __repr__(self):
...
+            salt=salt,
+            time_cost=time_cost,
+            memory_cost=memory_cost,
+            parallelism=parallelism,
+            type=type,
+        )
+        return key[:32], key[32:]
```
ThomasWaldmann, line 163: seems like this could be done by the caller with more flexibility.

src/borg/testsuite/helpers.py
```diff
@@ -1207,3 +1210,25 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+            type=argon2.low_level.Type.I
+        )
+
+        # Assert
+        assert len(enc_key)*8 == 256
+        assert len(mac_key)*8 == 256
+        # echo -n "hello, pass phrase" | argon2 saltsaltsaltsaltsaltsaltsaltsaltsaltsaltsaltsaltsaltsaltsaltsalt -i -t 1 -k 4096 -p 1 -l 64 -r
+        expected = 'b668a71dded9a9772686ab4c2da2a06c491087e9e27c14f9e798ad8b929506b43ef3878e4bb03055b6efbecddcdbfc58bfc1617a0e1790a46c1a9991c3269b34'
+        expected = b16decode(expected.upper())
+        enc_key + mac_key == expected
```
ThomasWaldmann, line 1234: besides the length checks, this is basically testing THEIR code.
can you check if they have own tests so we do not need to test it here?

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +141,23 @@ def __repr__(self):
...
+    def argon2(
+        self,
+        *,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: argon2.low_level.Type
+    ) -> (bytes, bytes):
+        key = argon2.low_level.hash_secret_raw(
```
ThomasWaldmann, line 154: dinosaurs with laser guns! :-)

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +141,23 @@ def __repr__(self):
...
         return pbkdf2_hmac('sha256', self.encode('utf-8'), salt, iterations, length)
+
+    def argon2(
+        self,
+        *,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: argon2.low_level.Type
```
ThomasWaldmann, line 152: maybe change this to str, so that the caller does not need to import from argon2.low_level?

otoh, don't we always want "ID" anyway? so we could just hardcode that here and remove the param?

src/borg/helpers/passphrase.py
hexagonrecursion: The high level API of argon2 is designed for _password hashing_. Using it as a _key derivation function_ would be awkward and have no benefit as far as I can tell.

src/borg/helpers/passphrase.py
hexagonrecursion: I am sorry, but I can not find the place where you describe the design rationale for choosing argon2id.

I admit that I have almost no knowledge of crypto, but this quite from the [argon2 paper](https://raw.githubusercontent.com/P-H-C/phc-winner-argon2/master/argon2-specs.pdf) suggests we should use Argon2i

> 
> Argon2i is optimized for more dangerous settings, where the adversary possibly can access the same machine,
> use its CPU or mount cold-boot attacks. We use three passes to get rid entirely of the password in the memory.
> We suggest the following settings:
> • Key derivation for hard-drive encryption, that takes 3 seconds on a 2 GHz CPU using 2 cores — Argon2i
> with 4 lanes and 6 GB of RAM;

And this one too:

> 9 Recommended parameters
> We recommend the following procedure to select the type and the parameters for practical use of Argon2:
> 1. Select the type y. If you do not know the difference between them or you consider side-channel attacks
> as viable threat, choose Argon2i.

requirements.d/development.txt
hexagonrecursion: Note to self: CI did not fail. It should have.

src/borg/testsuite/helpers.py
hexagonrecursion: [It looks like they do have their own regression tests](https://github.com/hynek/argon2-cffi/blob/0c7b6ec66798b983a1966cbf8b802f8e9d68f1ad/tests/test_low_level.py#L130-L146). I'll remove this than.

src/borg/helpers/passphrase.py
ThomasWaldmann: yes, I've seen that the HL api is not usable for us.

requirements.d/development.txt
ThomasWaldmann: on CI, we have a development/testing setup.

but normal borg users will just have what's pulled via setup.py (or pyproject.toml).

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +142,28 @@ def __repr__(self):
...
+
+    def argon2(
+        self,
+        output_len_in_bytes: int,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: Literal[b'argon2i', b'argon2d', b'argon2id']
+    ) -> (bytes, bytes):
```
ThomasWaldmann, line 154: only bytes now.

src/borg/helpers/passphrase.py
hexagonrecursion: From your link (emphasis mine):

> ID = 2[¶](https://argon2-cffi.readthedocs.io/en/stable/api.html#argon2.low_level.Type.ID)
> 
>     Argon2id is a hybrid of Argon2i and Argon2d, using a combination of data-depending and data-independent memory accesses, which gives some of Argon2i’s resistance to side-channel cache timing attacks and much of Argon2d’s resistance to GPU cracking attacks.
> 
>     That makes it the preferred type for password hashing and **password-based key derivation**.

I would prefer if the author of the binding linked to their source for this bit of wisdom.

src/borg/helpers/passphrase.py
```diff
@@ -139,3 +142,28 @@ def __repr__(self):
...
         return pbkdf2_hmac('sha256', self.encode('utf-8'), salt, iterations, length)
+
+    def argon2(
+        self,
+        output_len_in_bytes: int,
+        salt: bytes,
+        time_cost,
+        memory_cost,
+        parallelism,
+        type: Literal[b'argon2i', b'argon2d', b'argon2id']
```
ThomasWaldmann, line 153: hmm, thinking about it again, guess i'ld like str here better than bytes. it's text.

do we need to repeat the "argon2" part in the types?

src/borg/helpers/passphrase.py
hexagonrecursion: At the very least their own advice is inconsistent:

> https://argon2-cffi.readthedocs.io/en/stable/argon2.html#what-is-argon2
> Argon2i    uses data-independent memory access, which is preferred for password hashing and **password-based key derivation**. Argon2i is slower as it makes more passes over the memory to protect from tradeoff attacks.

ThomasWaldmann: file an issue in their issue tracker?

ThomasWaldmann:
https://www.ietf.org/archive/id/draft-irtf-cfrg-argon2-03.txt

```
Select the type y.
If you do not know the difference between them or you consider side-channel attacks as viable threat, choose Argon2id.
```

src/borg/helpers/passphrase.py
hexagonrecursion: Ref: https://github.com/hynek/argon2-cffi/issues/118

src/borg/helpers/passphrase.py
hexagonrecursion: > do we need to repeat the "argon2" part in the types?

Are you saying the set of values should be `{'i', 'd', 'id'}`?

hexagonrecursion:
ci.appveyor.com:

> 
> Uploading artifacts...
> Error uploading artifact to the storage: The request was aborted: Could not create SSL/TLS secure channel.

src/borg/helpers/passphrase.py
ThomasWaldmann: yes :)

ThomasWaldmann:
seems like they have issues at appveyor (but we can just ignore that for now).

ThomasWaldmann:
btw, if this shall get merged asap (is that why you split into multiple PRs?), change the PR title so it matches the PR contents.

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
OK, ready for merging?

ThomasWaldmann:
I'ld squash the commits together...

hexagonrecursion:
> OK, ready for merging?

ready

PR #6469, opened 2022-03-19
Argon2 the second part: implement key encryption / decryption

hexagonrecursion:
~~Note: this PR is not yet complete. I am submitting it to try running CI~~

This is the second part of https://github.com/borgbackup/borg/issues/747

~~ - [x] Add code in decrypt_key_file necessary to be able to decrypt v2 keys ~~
~~ - [x] Implement encryption of keys in the new format ~~
See [new TODO](https://github.com/borgbackup/borg/pull/6469#issuecomment-1079898226)

~~This PR includes #6468~~

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -36,6 +36,8 @@ def clean_env(tmpdir_factory, monkeypatch):
             if key.startswith('BORG_') and key not in ('BORG_FUSE_IMPL', )]
     for key in keys:
         monkeypatch.delenv(key, raising=False)
+    # Speed up tests: skip key derivation
+    monkeypatch.setenv("BORG_TESTONLY_MOCK_KDF", "1")
 
 
 def pytest_report_header(config, startdir):
--- a/conftest.py
+++ b/conftest.py
@@ -2,11 +2,6 @@ import os
 
 import pytest
 
-# IMPORTANT keep this above all other borg imports to avoid inconsistent values
-# for `from borg.constants import PBKDF2_ITERATIONS` (or star import) usages before
-# this is executed
-from borg import constants
-
 # needed to get pretty assertion failures in unit tests:
 if hasattr(pytest, 'register_assert_rewrite'):
     pytest.register_assert_rewrite('borg.testsuite')
@@ -33,8 +28,8 @@ def clean_env(tmpdir_factory, monkeypatch):
             if key.startswith('BORG_') and key not in ('BORG_FUSE_IMPL', )]
     for key in keys:
         monkeypatch.delenv(key, raising=False)
-    # Speed up tests: skip key derivation
-    monkeypatch.setenv("BORG_TESTONLY_MOCK_KDF", "1")
+    # Speed up tests
+    monkeypatch.setenv("BORG_TESTONLY_WEAKEN_KDF", "1")
 
 
 def pytest_report_header(config, startdir):
--- a/conftest.py
+++ b/conftest.py
@@ -6,9 +6,6 @@ import pytest
 # for `from borg.constants import PBKDF2_ITERATIONS` (or star import) usages before
 # this is executed
 from borg import constants
-# no fixture-based monkey-patching since star-imports are used for the constants module
-constants.PBKDF2_ITERATIONS = 1
-
 
 # needed to get pretty assertion failures in unit tests:
 if hasattr(pytest, 'register_assert_rewrite'):
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -398,7 +398,8 @@ class Archiver:
             setattr(key_new, name, value)
 
         key_new.target = key_new.get_new_target(args)
-        key_new.save(key_new.target, key._passphrase, create=True)  # save with same passphrase
+        algorithm = 'sha256'  # TODO: copy from `key`
+        key_new.save(key_new.target, key._passphrase, create=True, algorithm=algorithm)  # save with same passphrase
 
         # rewrite the manifest with the new key, so that the key-type byte of the manifest changes
         manifest.key = key_new
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -398,8 +398,8 @@ class Archiver:
             setattr(key_new, name, value)
 
         key_new.target = key_new.get_new_target(args)
-        algorithm = 'sha256'  # TODO: copy from `key`
-        key_new.save(key_new.target, key._passphrase, create=True, algorithm=algorithm)  # save with same passphrase
+        # save with same passphrase and algorithm
+        key_new.save(key_new.target, key._passphrase, create=True, algorithm=key._encrypted_key_algorithm)
 
         # rewrite the manifest with the new key, so that the key-type byte of the manifest changes
         manifest.key = key_new
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4311,8 +4311,7 @@ class Archiver:
                                help='Set storage quota of the new repository (e.g. 5G, 1.5T). Default: no quota.')
         subparser.add_argument('--make-parent-dirs', dest='make_parent_dirs', action='store_true',
                                help='create the parent directories of the repository directory, if they are missing.')
-        subparser.add_argument('--key-algorithm', dest='key_algorithm', default='argon2 aes256-ctr hmac-sha256',
-                               choices=['pbkdf2-sha256 aes256-ctr hmac-sha256', 'argon2 aes256-ctr hmac-sha256'])
+        subparser.add_argument('--key-algorithm', dest='key_algorithm', default='argon2', choices=list(KEY_ALGORITHMS))
 
         # borg key
         subparser = subparsers.add_parser('key', parents=[mid_common_parser], add_help=False,
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4311,6 +4311,8 @@ class Archiver:
                                help='Set storage quota of the new repository (e.g. 5G, 1.5T). Default: no quota.')
         subparser.add_argument('--make-parent-dirs', dest='make_parent_dirs', action='store_true',
                                help='create the parent directories of the repository directory, if they are missing.')
+        subparser.add_argument('--key-algorithm', dest='key_algorithm', default='argon2 aes256-ctr hmac-sha256',
+                               choices=['pbkdf2-sha256 aes256-ctr hmac-sha256', 'argon2 aes256-ctr hmac-sha256'])
 
         # borg key
         subparser = subparsers.add_parser('key', parents=[mid_common_parser], add_help=False,
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -439,6 +439,7 @@ class FlexiKey:
         if encrypted_key.version != 1:
             raise UnsupportedKeyFormatError()
         else:
+            self._encrypted_key_algorithm = encrypted_key.algorithm
             if encrypted_key.algorithm == 'sha256':
                 return self.decrypt_key_file_pbkdf2(encrypted_key, passphrase)
             elif encrypted_key.algorithm == 'argon2 aes256-ctr hmac-sha256':
@@ -548,8 +549,7 @@ class FlexiKey:
     def change_passphrase(self, passphrase=None):
         if passphrase is None:
             passphrase = Passphrase.new(allow_empty=True)
-        algorithm = 'sha256'  # TODO: copy from `self`
-        self.save(self.target, passphrase, algorithm=algorithm)
+        self.save(self.target, passphrase, algorithm=self._encrypted_key_algorithm)
 
     @classmethod
     def create(cls, repository, args):
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -499,20 +499,11 @@ class FlexiKey:
         return msgpack.packb(enc_key.as_dict())
 
     def encrypt_key_file_argon2(self, data, passphrase):
-        # https://www.rfc-editor.org/rfc/rfc9106.html#section-4-6.2
-        time_cost = 3
-        memory_cost = 2**16
-        parallelism = 4
-        type_ = 'id'
-
-        salt = os.urandom(16)
+        salt = os.urandom(ARGON2_SALT_BYTES)
         key = passphrase.argon2(
             output_len_in_bytes=64,
             salt=salt,
-            time_cost=time_cost,
-            memory_cost=memory_cost,
-            parallelism=parallelism,
-            type=type_,
+            **ARGON2_ARGS,
         )
         enc_key, mac_key = key[:32], key[32:]
         ae_cipher = AES256_CTR_HMAC_SHA256(
@@ -524,11 +515,8 @@ class FlexiKey:
             version=1,
             algorithm='argon2 aes256-ctr hmac-sha256',
             salt=salt,
-            argon2_time_cost=time_cost,
-            argon2_memory_cost=memory_cost,
-            argon2_parallelism=parallelism,
-            argon2_type=type_,
             data=ae_cipher.encrypt(data),
+            **{'argon2_' + k: v for k, v in ARGON2_ARGS.items()},
         )
         return msgpack.packb(encrypted_key.as_dict())
 
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -612,7 +612,7 @@ class FlexiKey:
             self.target = target
         return success
 
-    def save(self, target, passphrase, create=False):
+    def save(self, target, passphrase, algorithm='TODO', create=False):
         key_data = self._save(passphrase)
         if self.STORAGE == KeyBlobStorage.KEYFILE:
             if create and os.path.isfile(target):
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -502,7 +502,8 @@ class FlexiKey:
     def change_passphrase(self, passphrase=None):
         if passphrase is None:
             passphrase = Passphrase.new(allow_empty=True)
-        self.save(self.target, passphrase)
+        algorithm = 'sha256'  # TODO: copy from `self`
+        self.save(self.target, passphrase, algorithm=algorithm)
 
     @classmethod
     def create(cls, repository, args):
@@ -512,7 +513,8 @@ class FlexiKey:
         key.init_from_random_data()
         key.init_ciphers()
         target = key.get_new_target(args)
-        key.save(target, passphrase, create=True)
+        algorithm = 'sha256'  # TODO: initialize based on `args`
+        key.save(target, passphrase, create=True, algorithm=algorithm)
         logger.info('Key in "%s" created.' % target)
         logger.info('Keep this key safe. Your data will be inaccessible without it.')
         return key
@@ -612,7 +614,7 @@ class FlexiKey:
             self.target = target
         return success
 
-    def save(self, target, passphrase, algorithm='TODO', create=False):
+    def save(self, target, passphrase, algorithm, create=False):
         key_data = self._save(passphrase)
         if self.STORAGE == KeyBlobStorage.KEYFILE:
             if create and os.path.isfile(target):
@@ -688,8 +690,8 @@ class AuthenticatedKeyBase(AESKeyBase, FlexiKey):
         self.logically_encrypted = False
         return success
 
-    def save(self, target, passphrase, create=False):
-        super().save(target, passphrase, create=create)
+    def save(self, target, passphrase, algorithm, create=False):
+        super().save(target, passphrase, algorithm, create=create)
         self.logically_encrypted = False
 
     def init_ciphers(self, manifest_data=None):
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -547,8 +547,7 @@ class FlexiKey:
         key.init_from_random_data()
         key.init_ciphers()
         target = key.get_new_target(args)
-        algorithm = 'sha256' if args.key_algorithm == 'pbkdf2-sha256 aes256-ctr hmac-sha256' else args.key_algorithm
-        key.save(target, passphrase, create=True, algorithm=algorithm)
+        key.save(target, passphrase, create=True, algorithm=KEY_ALGORITHMS[args.key_algorithm])
         logger.info('Key in "%s" created.' % target)
         logger.info('Keep this key safe. Your data will be inaccessible without it.')
         return key
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -25,6 +25,7 @@ from ..platform import SaveFile
 from .nonces import NonceManager
 from .low_level import AES, bytes_to_long, long_to_bytes, bytes_to_int, num_cipher_blocks, hmac_sha256, blake2b_256, hkdf_hmac_sha512
 from .low_level import AES256_CTR_HMAC_SHA256, AES256_CTR_BLAKE2b, AES256_OCB, CHACHA20_POLY1305
+from . import low_level
 
 
 class UnsupportedPayloadError(Error):
@@ -467,7 +468,10 @@ class FlexiKey:
             enc_key=enc_key,
             mac_key=mac_key,
         )
-        return ae_cipher.decrypt(encrypted_key.data)
+        try:
+            return ae_cipher.decrypt(encrypted_key.data)
+        except low_level.IntegrityError:
+            return None
 
     def encrypt_key_file(self, data, passphrase, algorithm):
         if algorithm == 'sha256':
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -472,7 +472,7 @@ class FlexiKey:
     def encrypt_key_file(self, data, passphrase, algorithm):
         if algorithm == 'sha256':
             return self.encrypt_key_file_pbkdf2(data, passphrase)
-        elif algorithm == 'argon2 aes256_ctr hmac_sha256':
+        elif algorithm == 'argon2 aes256-ctr hmac-sha256':
             return self.encrypt_key_file_argon2(data, passphrase)
         else:
             raise ValueError(f'Unexpected algorithm: {algorithm}')
@@ -555,7 +555,7 @@ class FlexiKey:
         key.init_from_random_data()
         key.init_ciphers()
         target = key.get_new_target(args)
-        algorithm = 'sha256'  # TODO: initialize based on `args`
+        algorithm = 'sha256' if args.key_algorithm == 'pbkdf2-sha256 aes256-ctr hmac-sha256' else args.key_algorithm
         key.save(target, passphrase, create=True, algorithm=algorithm)
         logger.info('Key in "%s" created.' % target)
         logger.info('Keep this key safe. Your data will be inaccessible without it.')
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -51,6 +51,10 @@ class RepoKeyNotFoundError(Error):
     """No key entry found in the config of repository {}."""
 
 
+class UnsupportedKeyFormatError(Error):
+    """Your borg key is stored in an unsupported format. Try using a newer version of borg."""
+
+
 class TAMRequiredError(IntegrityError):
     __doc__ = textwrap.dedent("""
     Manifest is unauthenticated, but it is required for this repository.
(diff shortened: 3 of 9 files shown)


hexagonrecursion:
After refreshing my memory of Encrypt-then-MAC vs Encrypt-and-MAC vs MAC-then-Encrypt I feel like I would prefer to use a high level API that just does the right thing. Do you have one? If not I'll try to Encrypt-then-MAC. How hard can it be to remember to MAC all the bits in the message that affect the decryption?

hexagonrecursion:
Done. I will mark this as ready for review after #6468 is merged.

hexagonrecursion:
Note: a malicious key file might be able to perform denial of service e.g. by causing us to allocate too much RAM

src/borg/crypto/key.py
```diff
@@ -432,13 +432,33 @@ def decrypt_key_file(self, data, passphrase):
...
+    def decrypt_key_file_v1(self, encrypted_key, passphrase):
+        assert encrypted_key.algorithm == 'sha256'
+        key = passphrase.kdf(encrypted_key.salt, encrypted_key.iterations, 32)
+        data = AES(key, b'\0'*16).decrypt(encrypted_key.data)
+        if hmac_sha256(key, data) == encrypted_key.hash:
             return data
+        return None
+
+    def decrypt_key_file_v2(self, encrypted_key, passphrase):
+        enc_key, mac_key = passphrase.argon2(
```
ThomasWaldmann, line 452: maybe also add a algorithm check above this like in v1 in case we ever want to change?

src/borg/crypto/key.py
```diff
@@ -432,13 +432,33 @@ def decrypt_key_file(self, data, passphrase):
...
-        key = passphrase.kdf(enc_key.salt, enc_key.iterations, 32)
-        data = AES(key, b'\0'*16).decrypt(enc_key.data)
-        if hmac_sha256(key, data) == enc_key.hash:
+        encrypted_key = EncryptedKey(internal_dict=data)
+        if encrypted_key.version == 1:
+            return self.decrypt_key_file_v1(encrypted_key, passphrase)
+        elif encrypted_key.version == 2:
+            return self.decrypt_key_file_v2(encrypted_key, passphrase)
+        else:
+            raise Exception('Your borg key is stored in an unsupported format. Try using a newer version of borg')
```
ThomasWaldmann, line 441: Please add a dot at the end of a sentence.

src/borg/crypto/key.py
```diff
@@ -432,13 +432,33 @@ def decrypt_key_file(self, data, passphrase):
...
+        enc_key, mac_key = passphrase.argon2(
+            salt=encrypted_key.salt,
+            time_cost=encrypted_key.argon2_time_cost,
+            memory_cost=encrypted_key.argon2_memory_cost,
+            parallelism=encrypted_key.argon2_parallelism,
+            type=encrypted_key.argon2_type,
+        )
+        if not hmac.compare_digest(encrypted_key.hash, hmac_sha256(mac_key, encrypted_key.data)):
+            return None
+        return AES(enc_key, b'\0'*16).decrypt(encrypted_key.data)
```
ThomasWaldmann, line 461: Just had the idea if we maybe want to use chacha20-poly1305 here instead of self-made authenticated encryption.

ThomasWaldmann: otoh, openbsd with libressl does not have it, so guess we better don't. :-(

src/borg/testsuite/crypto.py
```diff
@@ -228,3 +233,57 @@ def test_hkdf_hmac_sha512_5(self):
 
         okm = hkdf_hmac_sha512(ikm, salt, info, l)
         assert okm == bytes.fromhex('1407d46013d98bc6decefcfee55f0f90b0c7f63d68eb1a80eaf07e953cfc0a3a5240a155d6e4daa965bb')
+
+
+def test_decrypt_key_file_v2(monkeypatch):
+    # Arrange
+    plain = b'hello'
+    enc_key = bytes.fromhex('b668a71dded9a9772686ab4c2da2a06c491087e9e27c14f9e798ad8b929506b4')
+    cypher = AES(enc_key, b'\0'*16).encrypt(plain)
```
ThomasWaldmann, line 242: cipher? please check the src which way we usually write it.

src/borg/testsuite/crypto.py
```diff
@@ -228,3 +233,57 @@ def test_hkdf_hmac_sha512_5(self):
...
+        'algorithm': 'sha256',
+        'hash': b'E'*32,
+        'data': data
+    })
+
+    # Act
+    result = key.decrypt_key_file(encrypted, passphrase)
+
+    # Assert
+    assert result is None
```
ThomasWaldmann, line 289: hmm, do you think arrange / act / assert comments are giving enough value to have them?
it's always that order and also the assert is rather obvious anyway.

src/borg/item.pyx
```diff
@@ -304,6 +307,15 @@ class EncryptedKey(PropDict):
...
     data = PropDict._make_property('data', bytes)
+    argon2_time_cost = PropDict._make_property('argon2_time_cost', int)
+    argon2_memory_cost = PropDict._make_property('argon2_memory_cost', int)
+    argon2_parallelism = PropDict._make_property('argon2_parallelism', int)
+    argon2_type = PropDict._make_property(
+        'argon2_type',
+        argon2.low_level.Type,
+        encode=lambda t: t.name.encode(),
+        decode=lambda b: argon2.low_level.Type[b.decode()]
+    )
```
ThomasWaldmann, line 318: quite some effort.

but considering key files will last for long, we can not accept any library-side change of `argon2.low_level.Type` anyway (because then our existing keyfiles would not match any more and stuff would likely blow up).

so just have str / bytes or just always use `ID`?

src/borg/crypto/key.py
ThomasWaldmann: `borg.crypto.low_level.AES256_CTR_HMAC_SHA256` could be used.

that way we maybe could get rid of the legacy `AES` class at some time in the future.

ThomasWaldmann:
About DoS via expensive argon2 parameters in key file: guess if someone wants to annoy you and has write access to your key files, they can do worse than that, like e.g. just delete the key files.

src/borg/crypto/key.py
```diff
@@ -432,13 +436,37 @@ def decrypt_key_file(self, data, passphrase):
...
+        assert encrypted_key.algorithm == 'sha256'
+        key = passphrase.kdf(encrypted_key.salt, encrypted_key.iterations, 32)
+        data = AES(key, b'\0'*16).decrypt(encrypted_key.data)
+        if hmac_sha256(key, data) == encrypted_key.hash:
             return data
+        return None
+
+    def decrypt_key_file_v2(self, encrypted_key, passphrase):
+        if encrypted_key.algorithm != "argon2 aes256-ctr hmac-sha256":
+            raise UnsupportedKeyFormatError()
```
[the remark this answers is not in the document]
ThomasWaldmann: or, as you've touched/moved the old code anyway: do the same as here at line 448.

src/borg/crypto/key.py
hexagonrecursion: I'll switch from `borg.crypto.low_level.AES` to `borg.crypto.low_level.AES256_CTR_HMAC_SHA256` - it's not that hard

hexagonrecursion:
This should be ready now.

src/borg/crypto/key.py
ThomasWaldmann: ok, incremental improvement. \o/ and we can kill the AES class once nobody is using v1 keys any more.

ThomasWaldmann:
maybe also do the "encrypt key file" here?

src/borg/testsuite/crypto.py
```diff
@@ -228,3 +233,60 @@ def test_hkdf_hmac_sha512_5(self):
...
+
+    envelope = algorithm.encrypt(plain)
+
+    encrypted = msgpack.packb({
+        'version': 2,
+        'salt': b'salt'*16,
+        'argon2_time_cost': 1,
+        'argon2_memory_cost': 4096,
+        'argon2_parallelism': 1,
+        'argon2_type': b'i',
```
ThomasWaldmann, line 256: didn't we end up with b'id' being the best type?

src/borg/crypto/key.py
```diff
@@ -432,13 +436,41 @@ def decrypt_key_file(self, data, passphrase):
...
+            parallelism=encrypted_key.argon2_parallelism,
+            type=encrypted_key.argon2_type,
+        )
+        enc_key, mac_key = key[:32], key[32:]
+        algorithm = AES256_CTR_HMAC_SHA256(
+            iv=0, header_len=0, aad_offset=0,
+            enc_key=enc_key,
+            mac_key=mac_key,
+        )
+        return algorithm.decrypt(encrypted_key.data)
```
ThomasWaldmann, line 473: 473 has different behaviour (throws exception on failure) than 452..454 (returns None on failure).

src/borg/crypto/key.py
```diff
@@ -432,13 +436,41 @@ def decrypt_key_file(self, data, passphrase):
...
+        key = passphrase.argon2(
+            output_len_in_bytes=64,
+            salt=encrypted_key.salt,
+            time_cost=encrypted_key.argon2_time_cost,
+            memory_cost=encrypted_key.argon2_memory_cost,
+            parallelism=encrypted_key.argon2_parallelism,
+            type=encrypted_key.argon2_type,
+        )
+        enc_key, mac_key = key[:32], key[32:]
+        algorithm = AES256_CTR_HMAC_SHA256(
```
ThomasWaldmann, line 468: ae_cipher is maybe a more speaking name.

hexagonrecursion:
> maybe also do the "encrypt key file" here?

back to draft then

hexagonrecursion:
Things left to do before this PR is ready:
- [ ] Fix passphrase retry bug
- [ ] borg init - defalt to argon2
- [ ] borg key change-passphrase - keep key algorithm the same
- [ ] borg key change-location - keep key algorithm the same

src/borg/crypto/key.py
ThomasWaldmann: did you see this ^^^?

ThomasWaldmann:
looks good so far. TBD yet whether we should rather use the recommended or 2nd recommended profile for argon2.

problem is that we have quite widespread use cases, some people use borg on their old raspi or on their mobile phone, other use it on server / workstation class hardware with lots of memory.

one way to solve that would be that memory limited users keep using pbkdf2 while we use the high-memory profile for argon2 for everybody else. another way would be to offer both argon2 profiles, so people with low memory do not need to use the old crap. :-)

src/borg/crypto/key.py
hexagonrecursion: I did. This is causing [the passphrase retry bug I mentioned](https://github.com/borgbackup/borg/pull/6469#issuecomment-1079898226). I'll fix this.

hexagonrecursion:
> TBD yet whether we should rather use the recommended or 2nd recommended profile for argon2

I went with the 2nd recommended for a very selfish reason - I did not want to raise the memory cap for the virtual machine:
```py
>>> argon2.low_level.hash_secret_raw(secret=b'hello', hash_len=64, salt=b'S'*16, time_cost=1, parallelism=4, memory_cost=2**21, type=argon2.low_level.Type.ID)
  File "/home/user/borg-master/.direnv/python-3.9.10/lib64/python3.9/site-packages/argon2/low_level.py", line 172, in hash_secret_raw
    raise HashingError(error_to_str(rv))
argon2.exceptions.HashingError: Memory allocation error
```

hexagonrecursion:
Ready.

Plan for the next two PRs:
* third part: `borg key change-algorithm`
* forth part: changelog and usage docs

src/borg/archiver.py
```diff
@@ -4311,6 +4311,8 @@ def define_borg_mount(parser):
                                help='Set storage quota of the new repository (e.g. 5G, 1.5T). Default: no quota.')
         subparser.add_argument('--make-parent-dirs', dest='make_parent_dirs', action='store_true',
                                help='create the parent directories of the repository directory, if they are missing.')
+        subparser.add_argument('--key-algorithm', dest='key_algorithm', default='argon2 aes256-ctr hmac-sha256',
+                               choices=['pbkdf2-sha256 aes256-ctr hmac-sha256', 'argon2 aes256-ctr hmac-sha256'])
```
ThomasWaldmann, line 4315: i guess we should remove the blanks from the algorithm names so users do not need to quote the names.

also guess people will hate the long names, same issue as for -E.

src/borg/crypto/key.py
```diff
@@ -555,7 +555,7 @@ def create(cls, repository, args):
         key.init_from_random_data()
         key.init_ciphers()
         target = key.get_new_target(args)
-        algorithm = 'sha256'  # TODO: initialize based on `args`
+        algorithm = 'sha256' if args.key_algorithm == 'pbkdf2-sha256 aes256-ctr hmac-sha256' else args.key_algorithm
```
ThomasWaldmann, line 562: guess these algo names deserve some constants in `borg.constants`.

src/borg/crypto/key.py
```diff
@@ -467,7 +468,10 @@ def decrypt_key_file_argon2(self, encrypted_key, passphrase):
             enc_key=enc_key,
             mac_key=mac_key,
         )
-        return ae_cipher.decrypt(encrypted_key.data)
+        try:
+            return ae_cipher.decrypt(encrypted_key.data)
+        except low_level.IntegrityError:
+            return None
```
ThomasWaldmann, line 475: exactly!

src/borg/testsuite/crypto.py
```diff
@@ -377,5 +369,6 @@ def test_repo_key_detect_does_not_raise_integrity_error(getpass, monkeypatch):
     RepoKey.create(repository, args=MagicMock(key_algorithm='argon2 aes256-ctr hmac-sha256'))
     repository.load_key.return_value = repository.save_key.call_args.args[0]
 
-    # This used to raise integrity errors due to a bug
+    # 1. detect() tries an empty passphrase first before prompting the user
+    # 2. load() was thorwing integrity errors instead of returning None due to a bug
```
ThomasWaldmann, line 373: typo

src/borg/crypto/key.py
```diff
@@ -454,7 +498,41 @@ def encrypt_key_file(self, data, passphrase):
         )
         return msgpack.packb(enc_key.as_dict())
 
-    def _save(self, passphrase):
+    def encrypt_key_file_argon2(self, data, passphrase):
+        # https://www.rfc-editor.org/rfc/rfc9106.html#section-4-6.2
+        time_cost = 3
+        memory_cost = 2**16
+        parallelism = 4
+        type_ = 'id'
```
ThomasWaldmann, line 506: for usage in `borg benchmark cpu`, it would be nice to have a constant with this, like we have for PBKDF2_ITERATIONS - so benchmark and real usage stay in sync.

src/borg/testsuite/archiver.py
```diff
@@ -3603,6 +3603,39 @@ def test_recovery_from_deleted_repo_nonce(self):
         self.cmd('create', self.repository_location + '::test2', 'input')
         assert os.path.exists(nonce)
 
+    def test_init_defaults_to_argon2(self):
```
ThomasWaldmann, line 3606: hmm, we "borg init" quite a lot in `borg.testsuite.archiver`.

If we slow that down via very secure argon2 parameters, our tests will run much slower.
guess we should have some way to accelerate that and using different params for testing and real world.

src/borg/testsuite/crypto.py
```diff
@@ -248,3 +257,118 @@ def test_hkdf_hmac_sha512_5(self):
...
+    })
+    key = KeyfileKey(None)
+
+    decrypted = key.decrypt_key_file(encrypted, passphrase)
+
+    assert decrypted == plain
+
+
+def test_decrypt_key_file_unsupported_algorithm(monkeypatch):
+    '''We will add more algorithms in the future. We should raise a helpful error.'''
```
ThomasWaldmann, line 319: can we consistently use triple-doublequotes, please?

src/borg/archiver.py
hexagonrecursion: Good point! How about this?

CLI  | internal
-|-
--key-algorithm pbkdf2 | sha256
--key-algorithm argon2 | argon2 aes256-ctr hmac-sha256

src/borg/testsuite/crypto.py
```diff
@@ -342,22 +342,25 @@ def test_decrypt_key_file_v2_is_unsupported(monkeypatch):
         key.decrypt_key_file(encrypted, passphrase)
 
 
-@pytest.mark.parametrize('key_algorithm', ('argon2 aes256-ctr hmac-sha256', 'sha256'))
-def test_key_file_roundtrip(monkeypatch, key_algorithm):
+@pytest.mark.parametrize('CLI_argument, expect_algorithm', (
+    ('argon2', b'argon2 aes256-ctr hmac-sha256'),
+    ('pbkdf2', b'sha256'),
```
ThomasWaldmann, line 347: can you use `KEY_ALGORITHMS.items()`?

ThomasWaldmann: or even without `.items()`?

ThomasWaldmann: and maybe keep the uppercase names for classes, `cli_argument` is fine, too.

ThomasWaldmann: `expected_algorithm` reads a bit nicer.

ThomasWaldmann (commented):
good change, better usability.

src/borg/testsuite/crypto.py
hexagonrecursion: > or even without `.items()`?

It does not work without `.items()` - I need both keys and values

src/borg/testsuite/archiver.py
hexagonrecursion: I am working on a solution for this. The first step is to measure the potential savings.

ThomasWaldmann: that problem isn't new btw. - maybe pbkdf2 also slowed down our tests already before your PR, not sure how much.

so a good comparison would be the default kdf against some  tuned-for-tests-only kdf.

hexagonrecursion: Look what I found:
https://github.com/borgbackup/borg/blob/22fc6d1bddc91c1bea163d906ce40f85c8744655/conftest.py#L8-L10

hexagonrecursion: I have performed an experiment: we can expect savings of about 220 +-20 seconds in a full test run. (note: I skipped test files that do not call the KDF because I was worried my benchmark would exceed github max timeout)

https://github.com/hexagonrecursion/borg/runs/5804738142?check_suite_focus=true

- [ ] TODO: patch conftest.py

ThomasWaldmann: definitely worth saving that time / resources.

ThomasWaldmann:
About BORG_TESTONLY_MOCK_KDF:
- makes the "kdf" super fast :-)
- this env variable would be also available for production use (unlike the PBKDF2_ITERATIONS patching we did in conftest.py). could this be intentionally abused / misused more easily?

conftest.py
```diff
@@ -36,6 +36,8 @@ def clean_env(tmpdir_factory, monkeypatch):
             if key.startswith('BORG_') and key not in ('BORG_FUSE_IMPL', )]
     for key in keys:
         monkeypatch.delenv(key, raising=False)
+    # Speed up tests: skip key derivation
+    monkeypatch.setenv("BORG_TESTONLY_MOCK_KDF", "1")
```
ThomasWaldmann, line 40: IF we keep this: remove line 9-10?

src/borg/testsuite/crypto.py
```diff
@@ -248,3 +258,120 @@ def test_hkdf_hmac_sha512_5(self):
...
+    passphrase = Passphrase.new()
+    key = KeyfileKey(None)
+
+    decrypted = key.decrypt_key_file(encrypted, passphrase)
+
+    assert decrypted == plain
+
+
+def test_decrypt_key_file_pbkdf2_sha256_aes256_ctr_hmac_sha256(monkeypatch):
+    monkeypatch.delenv('BORG_TESTONLY_MOCK_KDF')
```
ThomasWaldmann, line 297: if we just tweaked the algo **params** for the tests like we did for pbkdf2 in the past, this line would not be needed and we still would test the real code of the algo (just not in its production strength).

hexagonrecursion:
> * PBKDF2_ITERATIONS patching we did in conftest.py). could this be intentionally abused / misused more easily?

I do not see how this could be used maliciously. If I recall correctly, part of our attacker model is that we trust the environment. I assume this includes the environment variables.

I chose this because ArchiverTestCaseBinary spawns a child - our monkeypatching would not affect it.

hexagonrecursion:
Should we run testsuite/benchmark.py at full KDF strength? I feel that weakening KDF may defeat the point of a benchmark.

ThomasWaldmann:
Not sure about the benchmarks, but I guess:
- we do not intend to performance optimise the kdf code, so having it at full strength would not add value to the benchmark results
- it would increase resource usage and slow down the benchmarks
- especially this could become an issue if we got for argon2 rfc profile 1 in the end, which has much higher resource usage

hexagonrecursion:
I am starting one more experiment to see the difference between skipping KDF entirely and weakening its parameters

ThomasWaldmann:
Skip and weaken are often rather close. What are the difference between benchmark 0..3?

hexagonrecursion:
My data tells me that TESTONLY_MOCK_KDF=weaken and TESTONLY_MOCK_KDF=skip performed essentially identically.

Which approach would you prefer?
1. Skip the KDF entirely
2. Weaken the KDF

* A. Use the environment variable
* B. Patch a python variable in contest.py

ThomasWaldmann:
- weaken (because it basically still executes kdf code normally, just with other params)
- env var (because, as you discovered, only that works for ArchiverTestCaseBinary)

hexagonrecursion:
> What are the differences between benchmark 0..3 configuration?

None. I wanted to be confident in my methodology so I repeat the same test 4 times. I got consistent results as expected.

I have submitted one final benchmark - to verify that the performance improvement matches our expectation https://github.com/hexagonrecursion/borg/actions/runs/2094039179

My plan is to start working on `borg key change-algorithm` after this is merged. After `borg key change-algorithm` is merged I'll work on adding an argon2 benchmark to `borg benchmark cpu` and after that I'll work on the changelog and usage docs.

ThomasWaldmann:
The benchmark run you linked above did not work, rc == 1.

hexagonrecursion:
Thanks. I have started a run with --show-output to see what failed https://github.com/hexagonrecursion/borg/actions/runs/2096410352

hexagonrecursion:
Mission success https://github.com/hexagonrecursion/borg/actions/runs/2097120013

hexagonrecursion:
Are we ready merge this now or are we going to expand the scope of this PR again?

src/borg/helpers/passphrase.py
```diff
@@ -141,8 +141,8 @@ def __repr__(self):
         return '<Passphrase "***hidden***">'
 
     def kdf(self, salt, iterations, length):
-        if os.environ.get("BORG_TESTONLY_MOCK_KDF") is not None:
-            return b'X' * length
+        if os.environ.get("BORG_TESTONLY_WEAKEN_KDF") is not None:
+            iterations = 1
```
ThomasWaldmann, line 145: could the code from `yes` module be used here?

as it is, there is some danger that people set this to 0 or 1.

src/borg/helpers/passphrase.py
```diff
@@ -154,8 +154,11 @@ def argon2(
         parallelism,
         type: Literal['i', 'd', 'id']
     ) -> bytes:
-        if os.environ.get("BORG_TESTONLY_MOCK_KDF") is not None:
-            return b'X' * output_len_in_bytes
+        if os.environ.get("BORG_TESTONLY_WEAKEN_KDF") is not None:
+            time_cost = 1
+            parallelism = 1
+            # 8 is the smallest value that aviods the "Memory cost is too small" exception
```
ThomasWaldmann, line 160: typo

src/borg/helpers/passphrase.py
ThomasWaldmann: if not:
```
os.environ.get("BORG_TESTONLY_WEAKEN_KDF", "0") == "1"
```

hexagonrecursion: > if not:
> 
> ```
> os.environ.get("BORG_TESTONLY_WEAKEN_KDF", "0") == "1"
> ```

Good idea! This have not occurred to me

src/borg/testsuite/archiver.py
```diff
@@ -3603,6 +3603,39 @@ def test_recovery_from_deleted_repo_nonce(self):
...
+        """https://github.com/borgbackup/borg/issues/747#issuecomment-1076160401"""
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        with Repository(self.repository_path) as repository:
+            key = msgpack.unpackb(a2b_base64(repository.load_key()))
+        assert key[b'algorithm'] == b'argon2 aes256-ctr hmac-sha256'
+
+    def test_init_with_explicit_key_algorithm(self):
+        """https://github.com/borgbackup/borg/issues/747#issuecomment-1076160401"""
+        self.cmd('init', '--encryption=repokey', '--key-algorithm', 'pbkdf2', self.repository_location)
```
ThomasWaldmann, line 3615: just for consistency: can you please give `--key-algorithm` in the same way as `--encryption`?

please fix all places with this.

src/borg/testsuite/archiver.py
```diff
@@ -3603,6 +3603,39 @@ def test_recovery_from_deleted_repo_nonce(self):
...
+
+    def test_change_passphrase_does_not_change_algorithm(self):
+        self.cmd('init', '--encryption=repokey', '--key-algorithm', 'argon2', self.repository_location)
+        os.environ['BORG_NEW_PASSPHRASE'] = 'newpassphrase'
+
+        self.cmd('key', 'change-passphrase', self.repository_location)
+
+        with Repository(self.repository_path) as repository:
+            key = msgpack.unpackb(a2b_base64(repository.load_key()))
+            assert key[b'algorithm'] == b'argon2 aes256-ctr hmac-sha256'
```
ThomasWaldmann, line 3628: well, guess this test should start from pbkdf2 key, because:
- this is what people usually have now
- we would change algorithm pbkdf2 -> argon2 (but not argon2 -> pbkdf2) **if** we would do automated upgrades on change passphrase. so if you want to make sure here that we do **not** upgrade, you need to use the constellation that would actually get upgraded.

src/borg/testsuite/archiver.py
```diff
@@ -3603,6 +3603,39 @@ def test_recovery_from_deleted_repo_nonce(self):
...
+
+    def test_change_location_does_not_change_algorithm(self):
+        self.cmd('init', '--encryption=keyfile', '--key-algorithm', 'argon2', self.repository_location)
+
+        self.cmd('key', 'change-location', self.repository_location, 'repokey')
+
+        with Repository(self.repository_path) as repository:
+            key = msgpack.unpackb(a2b_base64(repository.load_key()))
+            assert key[b'algorithm'] == b'argon2 aes256-ctr hmac-sha256'
+
```
ThomasWaldmann, line 3638: same here, needs to start from pbkdf2.

src/borg/testsuite/crypto.py
```diff
@@ -248,3 +258,116 @@ def test_hkdf_hmac_sha512_5(self):
...
+    )
+
+    envelope = ae_cipher.encrypt(plain)
+
+    encrypted = msgpack.packb({
+        'version': 1,
+        'salt': b'salt'*4,
+        'argon2_time_cost': 3,
+        'argon2_memory_cost': 2**16,
+        'argon2_parallelism': 4,
```
ThomasWaldmann, line 280: this does not seem to match params seen in 265.

ThomasWaldmann: guess this can succeed because you are force-weakening the params in the tests even if the stored params disagree.

src/borg/testsuite/crypto.py
```diff
@@ -248,3 +258,116 @@ def test_hkdf_hmac_sha512_5(self):
...
+    salt = b'salt'*4
+    monkeypatch.setenv('BORG_PASSPHRASE', "hello, pass phrase")
+    passphrase = Passphrase.new()
+    key = passphrase.kdf(salt, iterations=1, length=32)
+    hash = hmac_sha256(key, plain)
+    data = AES(key, b'\0'*16).encrypt(plain)
+    encrypted = msgpack.packb({
+        'version': 1,
+        'algorithm': 'sha256',
+        'iterations': 1,
```
ThomasWaldmann, line 305: but here the weakened params are stored in the key also, so this is not consistent procedure with the above.

src/borg/testsuite/crypto.py
```diff
@@ -248,3 +258,116 @@ def test_hkdf_hmac_sha512_5(self):
...
+def test_repo_key_detect_does_not_raise_integrity_error(getpass, monkeypatch):
+    """https://github.com/borgbackup/borg/pull/6469#discussion_r832670411"""
+    repository = MagicMock(id=b'repository_id')
+    getpass.return_value = "hello, pass phrase"
+    monkeypatch.setenv('BORG_DISPLAY_PASSPHRASE', 'no')
+    RepoKey.create(repository, args=MagicMock(key_algorithm='argon2'))
+    repository.load_key.return_value = repository.save_key.call_args.args[0]
+
+    # 1. detect() tries an empty passphrase first before prompting the user
+    # 2. load() was throwing integrity errors instead of returning None due to a bug
```
ThomasWaldmann, line 372: can you check this test again? it somehow unclear what it is doing / what it is checking. Also the comment is not very clear as it refers to functions you are not calling (directly).

ThomasWaldmann (commented):
did a full review again and found some minor stuff.

src/borg/testsuite/crypto.py
hexagonrecursion: Good catch!

hexagonrecursion:
@ThomasWaldmann I have addressed your comments.

PR #6473, opened 2022-03-20
remove AEAD ciphers AES-OCB and CHACHA20-POLY1305, fixes #6472

### LINKED ISSUE #6472: remove low_level AEAD crypto from 1.2-maint
we have the AES-OCB and CHACHA20-POLY1305 in there since quite a while, but borg 1.2.x does not use these.

se we can also just remove these, esp. considering they got some changes / fixes in master branch - so we do not need to backport them.

ThomasWaldmann:
not used in 1.2.x, we'll start using them in 1.3.

### DIFF
--- a/src/borg/crypto/_crypto_helpers.c
+++ b/src/borg/crypto/_crypto_helpers.c
@@ -23,14 +23,3 @@ void HMAC_CTX_free(HMAC_CTX *ctx)
     }
 }
 #endif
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
@bket ^fewer workarounds needed for libressl/openbsd now (for borg 1.2.x).

PR #6475, opened 2022-03-21
1.2 - Remove unused code

bket:
#6473 Removed AEAD ciphers AES-OCB and CHACHA20-POLY1305 from 1.2. As a
result some additional code can be removed.

Passes regression tests.


ThomasWaldmann:
Thanks, also for testing!

PR #6476, opened 2022-03-22
Delete pointless assert

hexagonrecursion:
Maybe I am missing some subtle detail, but this assert looks redundant


hexagonrecursion:
Sorry. Deleted wrong line

PR #6477, opened 2022-03-22
Delete pointless assert

hexagonrecursion:
Maybe I am missing some subtle detail, but this assert looks redundant


ThomasWaldmann:
hah! yeah, that's pretty pointless.

as this likely would cause a conflict with my big crypto PR, I'll rather close yours and just apply your fix in my PR.

thanks for finding/fixing.

ThomasWaldmann:
13a7d594144f2e343df1a6e365fd6b293733955e

PR #6482, opened 2022-03-23
Simplify Libs Setup

### LINKED ISSUE #6392: simplify libs setup
in master branch, we recently got rid of most of the code in `setup_*.py`.

guess we now could put all the remaining code into a `setup_libs.py` or maybe after a little refactoring and deduplication of code maybe even directly into `setup.py`.

m3nu:
My first take on refactoring the setup script. What I did:

- Move some metadata to `setup.cfg` and only keep actual code in `setup.py`
- Common function `lib_ext_kwargs` to discover all libs

Open issues:
- There was a special condition for openssl for win32, but not the other libs. Is this leftover code or will other libs just work on Windows as-is?
- The minimum Python version in the metadata is 3.9. Shouldn't this be 3.7 or 3.8 (oldest version in Tox)?

Fixes #6392


ThomasWaldmann:
Before we started adding libdeflate and argon2-cffi, stuff has worked on windows.

But the current windows development effort seem to be stuck, see PRs / tickets.

Minimum Python shall be 3.9, so tox might need an update (master only).

setup.cfg
```diff
@@ -1,3 +1,74 @@
+[metadata]
+name = borgbackup
+author = The Borg Collective (see AUTHORS file)
+description = Deduplicated, encrypted, authenticated and compressed backups
+url = https://github.com/borgbase/vorta
```
ThomasWaldmann, line 5: nope :)

setup.cfg
```diff
@@ -1,3 +1,74 @@
...
+    Operating System :: MacOS :: MacOS X
+    Operating System :: POSIX :: Linux
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Topic :: Security :: Cryptography
+    Topic :: System :: Archiving :: Backup
+platforms = Linux, MacOS X, FreeBSD, OpenBSD, NetBSD
+long_description = file: README.rst
```
ThomasWaldmann, line 27: IIRC, README processing was a bit more complicated than that.

setup.cfg
```diff
@@ -1,3 +1,74 @@
...
+[options.extras_require]
+llfuse = llfuse >= 1.3.8
+pyfuse3 = pyfuse3 >= 3.1.1
+nofuse = 
+
+[options.exclude_package_data]
+* =
+    *.c
+    *.h
+    *.pyx
```
ThomasWaldmann, line 70: Maybe keep some info from the comments in 293-296.

setup.py
```diff
@@ -182,23 +140,37 @@ def members_appended(*ds):
...
+        if pc and pc.installed(lib_pkg_name, pc_version):
+            print(f'Detected and preferring {lib_pkg_name} [via pkg-config]')
+            return pc.parse(lib_pkg_name)
+        raise Exception(f'Could not find {lib_name} lib/headers, please set {prefix_env_var}')
+
+
     crypto_ext_kwargs = members_appended(
         dict(sources=[crypto_ll_source, crypto_helpers]),
-        setup_crypto.crypto_ext_kwargs(pc, system_prefix_openssl),
+        lib_ext_kwargs(pc, 'BORG_OPENSSL_PREFIX', 'crypto', 'libcrypto', '>=1.1.1'),
```
ThomasWaldmann, line 159: Please note that we also support libressl and i suspect its libcrypto won't have a openssl compatible version number?

That is why the orginal code didn't check for the version, but just for existence.

We have checks in the borg code that makes sure the crypto is not too old.

setup.py
```diff
@@ -182,23 +140,37 @@ def members_appended(*ds):
...
+        lib_ext_kwargs(pc, 'BORG_LIBLZ4_PREFIX', 'lz4', 'liblz4', '>= 1.7.0'),
+        lib_ext_kwargs(pc, 'BORG_LIBZSTD_PREFIX', 'zstd', 'libzstd', '>= 1.3.0'),
         dict(extra_compile_args=cflags),
     )
 
     checksums_ext_kwargs = members_appended(
         dict(sources=[checksums_source]),
-        setup_checksums.xxhash_ext_kwargs(pc, system_prefix_libxxhash),
-        setup_checksums.deflate_ext_kwargs(pc, system_prefix_libdeflate),
+        lib_ext_kwargs(pc, 'BORG_LIBXXHASH_PREFIX', 'xxhash', 'libxxhash', '>= 0.7.3'),
```
ThomasWaldmann, line 172: 0.8.1 was usually required for pkg-config support.

setup.py
```diff
@@ -165,6 +113,16 @@ def run(self):
     'clean2': Clean,
 }
 
+
+# How the build process finds the system libs:
+#
+# 1. if BORG_LIBXXX_PREFIX is set, it will use headers and libs from there.
```
ThomasWaldmann, line 119: Maybe also add BORG_OPENSSL_PREFIX here.

setup.cfg
```diff
@@ -1,3 +1,74 @@
...
+    Documentation = https://borgbackup.readthedocs.io
+    Source Code = https://github.com/borgbackup/borg
+
+[options]
+packages = find:
+package_dir =
+    =src
+# See also the MANIFEST.in file.
+# We want to install all the files in the package directories...
+include_package_data = true
```
ThomasWaldmann, line 41: this used to be right above exclude_package_data, so comments could be understood together.

setup.py
ThomasWaldmann: Maybe some always true version check could also work, like `>=0` or so?

setup_crypto.py
```diff
@@ -1,27 +0,0 @@
...
-
-def crypto_ext_kwargs(pc, system_prefix):
-    if system_prefix:
-        print('Detected OpenSSL [via BORG_OPENSSL_PREFIX]')
-        if is_win32:
-            lib_dir = system_prefix
-            lib_name = 'libcrypto'
-        else:
-            lib_dir = os.path.join(system_prefix, 'lib')
-            lib_name = 'crypto'
```
ThomasWaldmann, line 17: That needs checking on windows.

ThomasWaldmann (commented):
Nice improvement, much less duplication. Some comments...

setup.cfg
m3nu: Oops. Copied too much.

setup.py
m3nu: I got some test errors with 0.8.1. Currently it's 0.7.3 in code, but 0.8.1 in the comment. Will try to use 0.8.1 everywhere.

https://github.com/borgbackup/borg/blob/master/setup_checksums.py#L13

setup.py
m3nu: Just [checked](http://www.libressl.org/releases.html) and LibreSSL started at version 2.0, so will always be higher than openssl.

setup.py
m3nu: 0.8.1 fails in CI. Maybe their version is older. Will keep 0.7.3 for now.

```
  File "/home/runner/work/borg/borg/setup.py", line 177, in <module>
    lib_ext_kwargs(pc, 'BORG_LIBXXHASH_PREFIX', 'xxhash', 'libxxhash', '>= 0.8.1'),
...
Exception: Could not find xxhash lib/headers, please set BORG_LIBXXHASH_PREFIX
Error: Process completed with exit code 1.
```

setup_crypto.py
m3nu: Added this, but not pretty. If more Windows adjustments are needed, we may want a more scalable way.

setup.cfg
m3nu: True. Since we still need `setup_docs` anyways, I'm keeping the old function.

setup.py
ThomasWaldmann: ah right, it should of course work on the CI we have now.

setup.py
```diff
@@ -165,6 +113,16 @@ def run(self):
...
+
+# How the build process finds the system libs:
+#
+# 1. if BORG_{LIBXXX,OPENSSL}_PREFIX is set, it will use headers and libs from there.
+# 2. if not and pkg-config can locate the lib, the lib located by
+#    pkg-config will be used. We use the pkg-config tool via the pkgconfig
+#    python package, which must be installed before invoking setup.py.
+#    if pkgconfig is not installed, this step is skipped.
+# 3. otherwise raise a fatal error.
+
```
ThomasWaldmann, line 125: Maybe we should move this quite a bit upwards, because if people run into issues with `setup.py`, they likely should read this first.

setup.py
```diff
@@ -182,23 +140,42 @@ def members_appended(*ds):
...
+        if pc and pc.installed(lib_pkg_name, pc_version):
+            print(f'Detected and preferring {lib_pkg_name} [via pkg-config]')
+            return pc.parse(lib_pkg_name)
+        raise Exception(f'Could not find {lib_name} lib/headers, please set {prefix_env_var}')
+
+    if is_win32:
+        crypto_ext_lib = lib_ext_kwargs(
+            pc, 'BORG_OPENSSL_PREFIX', 'libcrypto', 'libcrypto', '>=1.1.1', lib_subdir='')
+    else:
+        crypto_ext_lib = lib_ext_kwargs(pc, 'BORG_OPENSSL_PREFIX', 'crypto', 'libcrypto', '>=1.1.1')
```
ThomasWaldmann, line 160: format like 157/158?

setup.py
```diff
@@ -182,23 +140,42 @@ def members_appended(*ds):
...
+        if system_prefix:
+            print(f'Detected and preferring {lib_pkg_name} [via {prefix_env_var}]')
+            return dict(include_dirs=[os.path.join(system_prefix, 'include')],
+                        library_dirs=[os.path.join(system_prefix, lib_subdir)],
+                        libraries=[lib_name])
+
+        if pc and pc.installed(lib_pkg_name, pc_version):
+            print(f'Detected and preferring {lib_pkg_name} [via pkg-config]')
+            return pc.parse(lib_pkg_name)
+        raise Exception(f'Could not find {lib_name} lib/headers, please set {prefix_env_var}')
```
ThomasWaldmann, line 154: While we are at it:

setting the `BORG_*_PREFIX` env var is only one way.

maybe we should also suggest that maybe `PKG_CONFIG_PATH` path needs an additional entry to also find the `pkg-config` information (`*.pc` file) for the lib that was not found.

setup.py
m3nu: OK. Moved it higher up after imports.

setup.py
m3nu: Good idea. We can reuse the `lib_pkg_name` variable already used by pkg-config. So for openssl it would say:

`Exception: Could not find crypto lib/headers, please set BORG_OPENSSL_PREFIX or ensure libcrypto.pc is in PKG_CONFIG_PATH.`

ThomasWaldmann (approved):
LGTM!

PR #6508, opened 2022-03-30
1.1-maint backport: delete: add repository id and location to prompt

aspargas2:
This is a backport of #6455.

The uses of f-strings in the original PR were changed to .format, as 1.1-maint does not seem to use f-strings anywhere.


ThomasWaldmann:
Thanks!

PR #6511, opened 2022-03-31
repo code cleanup

ThomasWaldmann:
nothing major, just simplifying some stuff so that future changes are easier.


src/borg/repository.py
```diff
@@ -807,7 +807,7 @@ def complete_xfer(intermediate=True):
                         # do not remove entry with empty shadowed_segments list here,
                         # it is needed for shadowed_put_exists code (see below)!
                         pass
-                    self.storage_quota_use -= len(data) + self.io.put_header_fmt.size
+                    self.storage_quota_use -= len(data) + self.io.HEADER_ID_SIZE
```
ThomasWaldmann, line 810: is not only used for PUT, but also for DEL.

Issue #5830, opened 2021-06-10
tar import/export: add --format option, pax-borg metadata


ThomasWaldmann:
Formats:
- GNU (borg 1.1 default)
- PAX
- BORG

The special `BORG` format would use pax format with some pax headers:
- `BORG.item.version` -> `1`
- `BORG.item.meta` -> `complete borg item metadata, serialised`

Intended application of the `borg` format are pipes like `borg export-tar ..| borg import-tar ...`. If import-tar discovers the `BORG.item` keys, it would just get all metadata from the corresponding value and would ignore the normal tarinfo members.

The borg format should provide a lossless transfer of all borg item metadata and data.

Bounty: https://app.bountysource.com/issues/99224362-tar-import-export-add-format-option-pax-borg-metadata

ThomasWaldmann:
Outputs a script that copies all archives from repo1 to repo2:
```
for A T in `borg list --format='{archive} {time:%Y-%m-%dT%H:%M:%S}{LF}' repo1`
do
    echo "borg export-tar --tar-format=BORG repo1::$A - | borg import-tar --timestamp=$T repo2::$A -"
done
```

PR #6522, opened 2022-04-02
import/export-tar: PAX format, ctime and atime support

ThomasWaldmann:
see #5830.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1445,9 +1445,20 @@ class TarfileObjectProcessors:
 
     @contextmanager
     def create_helper(self, tarinfo, status=None, type=None):
+        def s_to_ns(s):
+            return safe_ns(int(float(s) * 1e9))
+
         item = Item(path=make_path_safe(tarinfo.name), mode=tarinfo.mode | type,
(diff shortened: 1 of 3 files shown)


callegar:
Please, can we have tar-format=pax and maybe even tar-format=borg in 1.2.2? I see that 1.2.1 still does not have any tar-format selection. Would love to start cleaning up some entangled backups.

ThomasWaldmann:
@callegar how does this help to clean up "entangled backups"?

The PAX tar-format in master currently only adds the better timestamps, but still does not have xattrs/acls/flags.

The BORG tar-format in master adds all, but is rather close to the internals of borg. It was planned as a way for the borg 1.2 to 2.0 archive conversion, but I found out that is way to expensive (considering there is no deduplication) and have given up implementing this in favor of the way better "borg transfer". It might be still useful for other purposes / low volume transfers, but I won't add borg 1.2 to 2.0 conversion code to that, so it could be only used within the same borg version.

callegar:
I am more in the case of the few small archives, I think. If I can move a backup every few hours, then making that work through a few nights I'm done.

PR #6523, opened 2022-04-02
import/export-tar: --tar-format=BORG: roundtrip ALL item metadata

### LINKED ISSUE #5830: tar import/export: add --format option, pax-borg metadata

ThomasWaldmann:
fixes #5830.

export-tar: just msgpack and b64encode all item metadata and put that into a BORG specific PAX header. this is *additional* to the standard tar metadata.

import-tar: when detecting the BORG specific PAX header, just get all metadata from there (and ignore the standard tar metadata).

### DIFF
--- a/docs/usage/tar.rst
+++ b/docs/usage/tar.rst
@@ -1,6 +1,6 @@
-.. include:: import-tar.rst.inc
-
 .. include:: export-tar.rst.inc
+
+.. include:: import-tar.rst.inc
 
 Examples
 ~~~~~~~~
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
if noone objects or has review feedback, will merge this soon.

callegar:
Thanks for this addition! I see that this item is in the docs for 1.3 alpha. Is it planned to land in the 1.2 series as well?

ThomasWaldmann:
Well, the idea is that this is the enabler for crossing even a breaking release. For that to work, it needs to be available on both old and new side.

The code from this PR basically already solves that, but a remaining problem here is efficiency: just imagine you have 100 archives referencing 1TB data each. That would mean it would read 100TB from source repo and pump 100TB over the tar pipe and that would take quite a while. See my current PR which tries to optimize this.

callegar:
The PR you mention is that about the chunked BORG-C tar format, right? I have looked at it... there are a couple of things that are not completely clear to me...

1. Passing through the `tar` format is nice because the tar format is a sort of *lingua franca*. Yes, if you have multiple archives to export and re-import it can get quite inefficient. But it is something that could provide interoperability with other backup solutions. You could have backup software `foo`, `tar-export` with that and then `tar-import` in borg. In fact, having `borg` support import/export in `tar` format could well push users to expect the same functionality from other backup tools.  Using the *standard* gnu PAX format is also well in line with this.

2. Using `tar` with a minimum borg-specific PAX (as in `tar-format=borg`) assuring a correct roundtrip of all borg metadata can also be quite useful even if it remains not very efficient. First, it can be a way for moving archives from one repo to another until an efficient solution is developed. Secondly, it can also be a tool for interoperability between backup solutions. For instance another backup solution could *pretend* to be `borg` and make its best effort to export in a `tar` format with the borg pax extensions.  

3. However, for maximum efficiency, I miss something. I do not understand why passing through a `tar` derived format at all, given that it would anyway become so borg-specific to be more or less unsupportable by anything else.  Wouldn't the highest efficiency be achievable with `import-borg` that gets what it needs about an archive directly from another borg repo? And because there is a way of serving a repo over `ssh` with `serve` I expect that it could work also over the network.  And I believe it could be made to cross breaking releases.  An objection could be that a borg repo is not a single file. But once you have an `import-from-borg` command, it gets easy to build single-archive repos that can be placed inside a single `tar` or a `zip` file to simplify their moving around.

Am I missing something?  

Thanks in advance in case you can provide any bit that I am missing . Out of my curiosity, I will really appreciate that. But if you can't I will well understand, since I do not want to abuse of your time!

PR #6524, opened 2022-04-03
(docs) Recommend umask for passphrase file perms

bcat:
The previous sample for creating a ~/.borg-passphrase file creates it first and then chmod's it to 400 permissions. That's probably fine in practice, but means there's a tiny window where the passphrase file is sitting with default permissions (likely world readable, depending on the system umask).

It seems safer to first change the umask to remove all group & world bits (0077) _before_ creating the file. To be polite and avoid messing with the user's previous umask, we do this in a subshell. (Note that umask 0077 leads to a mode of 600 rather than the previous 400, because removing the owner write bit doesn't seem to buy much since the owner can just chmod the file anyway.)

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -569,8 +569,7 @@ Using ``BORG_PASSCOMMAND`` with a properly permissioned file
   directory and use permissions to keep anyone else from reading it. For
   example, first create a key::
 
-    head -c 32 /dev/urandom | base64 -w 0 > ~/.borg-passphrase
-    chmod 400 ~/.borg-passphrase
+    (umask 0077; head -c 32 /dev/urandom | base64 -w 0 > ~/.borg-passphrase)
 
   Then in an automated script one can put::
 


ThomasWaldmann:
Thanks! Do you also want to apply this to 1.2-maint / 1.1-maint branches?

bcat:
I suppose that might make sense, since these docs are independent of Borg version.

ThomasWaldmann:
The docs get released together with the borg releases. So master branch will get into borg 1.3.x later, while the other branches get into production releases sooner.

bcat:
Should I make PRs against merging into those other branches then? (I'm not super familiar with the workflow, sorry.)

ThomasWaldmann:
Yeah, if you like just do PRs against the other branches. With some luck, you can cherry-pick the changeset from master.

bcat:
Sorry for the delay, created those PRs:

* 1.2-maint: https://github.com/borgbackup/borg/pull/6558
* 1.1-maint: https://github.com/borgbackup/borg/pull/6559

Issue #2055, opened 2017-01-14
openbsd symlink mode test failure


ThomasWaldmann:
```

self = <borg.testsuite.archiver.ArchiverTestCase testMethod=test_basic_functionality>

    def test_basic_functionality(self):
        have_root = self.create_test_files()
        # fork required to test show-rc output
        output = self.cmd('init', '--encryption=repokey', '--show-version', '--show-rc', self.repository_location, fork=True)
        self.assert_in('borgbackup version', output)
        self.assert_in('terminating with success status, rc 0', output)
        self.cmd('create', self.repository_location + '::test', 'input')
        output = self.cmd('create', '--stats', self.repository_location + '::test.2', 'input')
        self.assert_in('Archive name: test.2', output)
        self.assert_in('This archive: ', output)
        with changedir('output'):
            self.cmd('extract', self.repository_location + '::test')
        list_output = self.cmd('list', '--short', self.repository_location)
        self.assert_in('test', list_output)
        self.assert_in('test.2', list_output)
        expected = [
            'input',
            'input/bdev',
            'input/cdev',
            'input/dir2',
            'input/dir2/file2',
            'input/empty',
            'input/file1',
            'input/flagfile',
        ]
        if are_fifos_supported():
            expected.append('input/fifo1')
        if are_symlinks_supported():
            expected.append('input/link1')
        if are_hardlinks_supported():
            expected.append('input/hardlink')
        if not have_root:
            # we could not create these device files without (fake)root
            expected.remove('input/bdev')
            expected.remove('input/cdev')
        if has_lchflags:
            # remove the file we did not backup, so input and output become equal
            expected.remove('input/flagfile')  # this file is UF_NODUMP
            os.remove(os.path.join('input', 'flagfile'))
        list_output = self.cmd('list', '--short', self.repository_location + '::test')
        for name in expected:
            self.assert_in(name, list_output)
>       self.assert_dirs_equal('input', 'output/input')

src/borg/testsuite/archiver.py:385: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/borg/testsuite/__init__.py:151: in assert_dirs_equal
    self._assert_dirs_equal_cmp(diff, **kwargs)
src/borg/testsuite/__init__.py:192: in _assert_dirs_equal_cmp
    self.assert_equal(d1, d2)
[...]

PR #6537, opened 2022-04-04
1.1 - Fix OpenBSD symlink mode test failure (#2055)

bket:
OpenBSD does not have `lchmod()` causing `os.lchmod` to be unavailable
on this platform. As a result ArchiverTestCase::test_basic_functionality
fails when run manually (#2055).

OpenBSD does have `fchmodat()`, which has a flag that makes it behave
like `lchmod()`. In Python this can be used via `os.chmod(path, mode,
follow_symlinks=False)`.

As of Python 3.3 `os.lchmod(path, mode)` is equivalent to
`os.chmod(path, mode, follow_symlinks=False)`. As such, switching to the
latter is preferred as it enables more platforms to do the right thing.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -44,7 +44,6 @@ from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
 from .remote import cache_if_remote
 from .repository import Repository, LIST_SCAN_LIMIT
 
-has_lchmod = hasattr(os, 'lchmod')
 has_link = hasattr(os, 'link')
 
 flags_normal = os.O_RDONLY | getattr(os, 'O_BINARY', 0)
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Looks like there is an issue with at least py35.

src/borg/archive.py
```diff
@@ -738,10 +737,18 @@ def restore_attrs(self, path, item, symlink=False, fd=None):
...
+        else:
+            # To check whether a particular function in the os module accepts False for its
+            # follow_symlinks parameter, the in operator on supports_follow_symlinks should be
+            # used. However, os.chmod is special as some platforms without a working lchmod() do
+            # have fchmodat(), which has a flag that makes it behave like lchmod(). fchmodat()
+            # is ignored when deciding whether or not os.chmod should be set in
+            # os.supports_follow_symlinks. Work around this by using try/except.
+            try:
+                os.chmod(path, item.mode, follow_symlinks=False)
+            except Exception:
```
ThomasWaldmann, line 749: Uuh. :-)

Can we catch that a bit less broad?

ThomasWaldmann:
Thanks!

PR #6545, opened 2022-04-07
add pyproject.toml, fix sys.path, fixes #6466

### LINKED ISSUE #6466: add pyproject.toml to 1.2-maint branch?
We have it already in master and it seems to solve the build requirements setup (like installing `pkgconfig` before `setup.py` runs).

1.2.0 release did not have `pyproject.toml` yet, shall we add it to `1.2.1`?

ThomasWaldmann:
Using pyproject.toml, we can avoid people running into issues when they did not install pkgconfig and cython first.

Also:

setup.py: add parent to sys.path
When using pyproject.toml the parent of setup.py is not on sys.path by default.
See <pypa/setuptools#3134>.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "pkgconfig", "Cython"]
+build-backend = "setuptools.build_meta"
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ try:
 except ImportError:
     cythonize = None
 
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
Note: we have quite some more setup.py / setup.cfg / pyproject.toml related changes in master, but the idea for 1.2 is to only do relatively small, conservative changes. We do not want to break anything in 1.2.1 that worked in 1.2.0.

PR #6547, opened 2022-04-07
Make a test cleanup after itself

hexagonrecursion:
The test was modifying os.environ without undoing the changes. Now it undoes them.


ThomasWaldmann:
- don't we have other places doing the same thing?
- isn't clean_env dealing with this?

hexagonrecursion:
> * 
> 
> * isn't clean_env dealing with this?

Indeed it is. Sorry

PR #6548, opened 2022-04-07
cleanup fs api calls, fixes #6531

### LINKED ISSUE #6531: cleanup fs api calls
https://docs.python.org/3/library/os.html#os.chown

https://docs.python.org/3/library/os.html#os.chmod

https://docs.python.org/3/library/os.html#files-and-directories

https://docs.python.org/3/library/os.html#os.supports_fd

as we noticed when working on #2055, some calls have more modern variants that:
- have a `follow_symlinks` parameter (so we do not need the `l*` variant of the api call any more)
- take a file descriptor instead of path (so we do not need the `f*` variant of the api call any more)

ThomasWaldmann:
try without the f* variants by just using "fd or path" as path parameter.

Let's try if this just works, even without first checking os.supports_fd and os.supports_follow_symlinks.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -868,26 +868,20 @@ Utilization of max. archive size: {csize_max:.0%}
         # This code is a bit of a mess due to os specific differences
         if not is_win32:
             try:
-                if fd:
-                    os.fchown(fd, uid, gid)
-                else:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
crap. closing, see #6531.

PR #6549, opened 2022-04-08
Argon2 the third part: `borg key change-algorithm`

hexagonrecursion:
[As promised](https://github.com/borgbackup/borg/pull/6469#issuecomment-1088264147) I have implemented `key change-algorithm`.

# My plan after this is merged:
* The fourth part: add an argon2 benchmark to borg benchmark cpu
* The fifth part: changelog and usage docs.


src/borg/archiver.py
```diff
@@ -4455,6 +4464,30 @@ def define_borg_mount(parser):
...
                                help='keep the key also at the current location (default: remove it)')
 
+        change_algorithm_epilog = process_epilog("""
+        Change the algorithm we use to encrypt and authenticate the borg key.
+
+        Your repository is encrypted and authenticated with a key that is randomly generated by ``borg init``.
+        The key is encrypted and authenticated with your passphrase.
+
+        We currently support two choices:
+        1. argon2 - recommended. This algorithm is used by default when creating a repository.
```
ThomasWaldmann, line 4474: maybe it is less confusing to newbies to say "... when initialising a ..." (borg init, not borg create).

src/borg/archiver.py
```diff
@@ -4455,6 +4464,30 @@ def define_borg_mount(parser):
...
+           The key encryption key is derived from your passphrase via argon2-id.
+           Argon2 is considered more modern and secure than pbkdf2.
+        1. pbkdf2 - the legacy algorithm. Use this if you want to access your repo via old versions of borg.
+           The key encryption key is derived from your passphrase via PBKDF2-HMAK-SHA256.
```
ThomasWaldmann, line 4478: typo HMAC

ThomasWaldmann (changes requested):
some minor nitpicks, rest LGTM.

src/borg/testsuite/archiver.py
```diff
@@ -3630,6 +3630,15 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
 
+    def test_key_change_algorithm(self):
+        self.cmd('init', '--encryption=repokey', '--key-algorithm=pbkdf2', self.repository_location)
+
+        self.cmd('key', 'change-algorithm', self.repository_location, 'argon2')
+
+        with Repository(self.repository_path) as repository:
+            _, key = Manifest.load(repository, Manifest.NO_OPERATION_CHECK)
+        assert key._encrypted_key_algorithm == 'argon2 aes256-ctr hmac-sha256'
+
```
ThomasWaldmann, line 3641: maybe you could do a `borg info` here to actually use the new key / check whether it works.

ThomasWaldmann (commented):
one last thing

ThomasWaldmann:
btw, there is already an argon2 / kdf cpu benchmark, but it should be updated so both benchmarks use the same parameters as borg in production.

src/borg/testsuite/archiver.py
hexagonrecursion: Good idea! Done.

hexagonrecursion:
> one last thing

What "last thing"?

ThomasWaldmann (approved):

ThomasWaldmann:
Thanks, merged!

Issue #6449, opened 2022-03-14
move kdf code


ThomasWaldmann:
`borg.helpers.passphrase` module has a `Passphrase.kdf` method.

- that's crypto code, so it should rather be below `borg.crypto` package
- all the other code there is just rather simple input / output / env var processing

@hexagonrecursion ^^^

(master branch only)

PR #6552, opened 2022-04-09
Argon2 fourth part: borg benchmark cpu: use ARGON2_ARGS

hexagonrecursion:
I have updated the cpu benchmark to use the same argon2 configuration as we use normally.

After this is merged I'll work on the last PR of #747 - the changelog and documentation updates

### DIFF
--- a/src/borg/__init__.py
+++ b/src/borg/__init__.py
@@ -1,7 +1,5 @@
 from packaging.version import parse as parse_version
 
-# IMPORTANT keep imports from borg here to a minimum because our testsuite depends on
-# being able to import borg.constants and then monkey patching borg.constants.PBKDF2_ITERATIONS
 from ._version import version as __version__
 
 
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -626,7 +626,7 @@ class Archiver:
         count = 5
         for spec, func in [
             ("pbkdf2", lambda: Passphrase('mypassphrase').kdf(b'salt'*8, PBKDF2_ITERATIONS, 32)),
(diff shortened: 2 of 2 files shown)


ThomasWaldmann (approved):
LGTM.

can you check that, guess it does not apply any more?
https://github.com/borgbackup/borg/blob/76063ab219cd84229f0ec077cf774d3b600b945d/src/borg/__init__.py#L3

hexagonrecursion:
> can you check that, guess it does not apply any more?

I think you are right
```sh
fgrep -r --exclude-dir=.tox --exclude-dir=.git --exclude-dir=__pycache__ PBKDF2_ITERATIONS
src/borg/archiver.py:            ("pbkdf2", lambda: Passphrase('mypassphrase').kdf(b'salt'*8, PBKDF2_ITERATIONS, 32)),
src/borg/crypto/key.py:        iterations = PBKDF2_ITERATIONS
src/borg/__init__.py:# being able to import borg.constants and then monkey patching borg.constants.PBKDF2_ITERATIONS
src/borg/constants.py:PBKDF2_ITERATIONS = 100000
```

ThomasWaldmann:
BTW, if you like, we could change the plan a bit:
- remove docs and changelog from THIS bounty, but add #6449 
- do docs and changelog for kdf-related changes within a "update docs" bounty

What do you think?

hexagonrecursion:
> BTW, if you like, we could change the plan a bit:
> 
>     * remove docs and changelog from THIS bounty, but add [move kdf code #6449](https://github.com/borgbackup/borg/issues/6449)
> 
>     * do docs and changelog for kdf-related changes within a "update docs" bounty
> 
> 
> What do you think?

Sure

PR #6557, opened 2022-04-10
fix build_man: missing benchmark_cpu.rst and key_change-location.rst

hexagonrecursion:
```
$ python setup.py build_usage && python setup.py build_man
...
building man page borg-benchmark-cpu(1)
error: [Errno 2] No such file or directory: 'docs/usage/benchmark_cpu.rst'
```

I get a similar error due to missing key_change-location.rst and key_change-algorithm.rst. I will submit key_change-algorithm.rst separately - I want to add a couple examples.


hexagonrecursion:
Sorry. This is the wrong patch. I'll submit the correct one soon.

PR #6565, opened 2022-04-11
Fix selftest

hexagonrecursion:
selftest imports testsuite.crypto
I did not realise this and imported pytest from testsuite.crypto
This broke the selftest.

Solution: move the tests that depend on pytest to testsuite.key.
All three affected tests are tests for the Key classes, so
this is probably a better plase for them anyway.


ThomasWaldmann (approved):
LGTM

PR #6593, opened 2022-04-14
openbsd: use openssl, remove libressl support, fixes #6474

### LINKED ISSUE #6474: libressl's AEAD api
it seems that they don't support these ciphers via the same EVP api as openssl:
- AES-OCB
- CHACHA20-POLY1305

For borg 1.3+ it seems like a good idea if we would deprecate the crypto modes based on AES-CTR and recommend the new AEAD modes.

So, how's the situation on OpenBSD? Are there other platforms requiring LibreSSL support?

- Do we need to support LibreSSL to run on OpenBSD or could we use OpenSSL there also?
- If we can neither use OpenSSL there nor some compatible API in LibreSSL, do we want to add some code for LibreSSL to our `borg.crypto.low_level` module that uses LibreSSL's EVP_AEAD api (which does not exist in OpenSSL)? https://man.openbsd.org/EVP_AEAD_CTX_init.3
- seems like we only get AES-GCM and chacha20-poly1305 via that api. there were objections against gcm, so we went for ocb rather. so guess that would mean with LibreSSL it would only be chacha20-poly1305 (which is fine as it is quite fast).

ThomasWaldmann:


setup.py
```diff
@@ -156,17 +157,23 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
...
 
+    crypto_cflags = list(cflags)  # copy the default cflags
     if is_win32:
         crypto_ext_lib = lib_ext_kwargs(
             pc, 'BORG_OPENSSL_PREFIX', 'libcrypto', 'libcrypto', '>=1.1.1', lib_subdir='')
+    elif is_openbsd:
+        # use openssl (not libressl) because we need AES-OCB and CHACHA20-POLY1305 via EVP api
+        crypto_ext_lib = lib_ext_kwargs(
+            pc, 'BORG_OPENSSL_PREFIX', 'crypto', 'libecrypto11', '>=1.1.1')
+        crypto_cflags += ['-Wl,-rpath,/usr/local/lib/eopenssl11']
```
ThomasWaldmann, line 168: somehow this looks strange.

would `'-Wl,-rpath=/usr/local/lib/eopenssl11'` work also?

setup.py
```diff
@@ -156,17 +157,23 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
             f"or ensure {lib_pkg_name}.pc is in PKG_CONFIG_PATH."
         )
 
+    crypto_cflags = list(cflags)  # copy the default cflags
```
ThomasWaldmann, line 160: see line 44:
```
cflags = [
    '-Wall',
    '-Wextra',
    '-Wpointer-arith',
]
```

setup.py
bket: yes, that works.

setup.py
bket: It does not make sense to pass linker flags via CFLAGS. Linker seems to ignore them, and compiler complains.

```
cc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -O2 -pipe -g -fPIC -O2 -pipe -g -O2 -pipe -g -O2 -pipe -fPIC -Isrc/borg/crypto -I/usr/local/include/eopenssl11 -I/usr/local/include/python3.9 -c src/borg/crypto/_crypto_helpers.c -o /tmp/ports/pobj/borgbackup-1.3.0.pre20220414/borg-3ef355a98f882f525a4ac7fa58dd561003e0e70a/temp.openbsd-7.1-amd64-3.9/src/borg/crypto/_crypto_helpers.o -Wall -Wextra -Wpointer-arith -Wl,-rpath,/usr/local/lib/eopenssl11
cc: warning: -Wl,-rpath,/usr/local/lib/eopenssl11: 'linker' input unused [-Wunused-command-line-argument]
```
What do you think of the diff below?

```
--- setup.py
+++ setup.py
@@ -23,6 +23,7 @@
 import setup_docs
 
 is_win32 = sys.platform.startswith('win32')
+is_openbsd = sys.platform.startswith('openbsd')
 
 # Number of threads to use for cythonize, not used on windows
 cpu_threads = multiprocessing.cpu_count() if multiprocessing and multiprocessing.get_start_method() != 'spawn' else None
@@ -156,9 +157,15 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
             f"or ensure {lib_pkg_name}.pc is in PKG_CONFIG_PATH."
         )
 
+    crypto_ldflags = []
     if is_win32:
         crypto_ext_lib = lib_ext_kwargs(
             pc, 'BORG_OPENSSL_PREFIX', 'libcrypto', 'libcrypto', '>=1.1.1', lib_subdir='')
+    elif is_openbsd:
+        # use openssl (not libressl) because we need AES-OCB and CHACHA20-POLY1305 via EVP api
+        crypto_ext_lib = lib_ext_kwargs(
+            pc, 'BORG_OPENSSL_PREFIX', 'crypto', 'libecrypto11', '>=1.1.1')
+        crypto_ldflags += ['-Wl,-rpath=/usr/local/lib/eopenssl11']
     else:
         crypto_ext_lib = lib_ext_kwargs(
             pc, 'BORG_OPENSSL_PREFIX', 'crypto', 'libcrypto', '>=1.1.1')
@@ -167,6 +174,7 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
         dict(sources=[crypto_ll_source, crypto_helpers]),
         crypto_ext_lib,
+        dict(extra_link_args=crypto_ldflags),
     )
 
     compress_ext_kwargs = members_appended(
```

ThomasWaldmann: yeah, sorry, i overlooked this wasn't meant for cflags, but ldflags.

src/borg/crypto/low_level.pyx
```diff
@@ -47,6 +47,8 @@ API_VERSION = '1.3_01'
 cdef extern from "openssl/crypto.h":
     int CRYPTO_memcmp(const void *a, const void *b, size_t len)
 
+cdef extern from "openssl/opensslv.h":
+    long OPENSSL_VERSION_NUMBER
```
ThomasWaldmann, line 51: currently not used, but guess it could be useful at some time.

ThomasWaldmann:
tests on openbsd current in vagrant also worked. thanks @bket!

PR #6604, opened 2022-04-16
argon2 key: use chacha20-poly1305 instead of aes256-ctr+hmac-sha256

### LINKED ISSUE #6601: switch argon2 keys to chacha20-poly1305 cipher
recently we had chosen to use aes-ctr-hmac-sha256 ciphersuite for the new argon2 keys.

the reasoning for this was:

1. be able to get rid of the low_level.AES class long term
2. we could not use AES-OCB or CHACHA20-POLY1305 because LibreSSL does not have that and we needed LibreSSL on OpenBSD.

now, as we moved to building borg with OpenSSL even on OpenBSD, we better do:

1. still applies
2. with OpenSSL, we can use CHACHA20-POLY1305 now! this is better than the AES-CTR based stuff and also we need it long-term anyway because we'll encrypt backup data chunks with it.
3. if we use CH-PO, we do not need AES-CTR any more for the borg key (and even better: we maybe could get completely rid of it long term)

ThomasWaldmann:
fixes #6601

so we can completely get rid of aes-ctr some day.

### DIFF
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -112,8 +112,8 @@ ARGON2_SALT_BYTES = 16
 KEY_ALGORITHMS = {
     # encrypt-and-MAC, kdf: PBKDF2(HMAC−SHA256), encryption: AES256-CTR, authentication: HMAC-SHA256
     'pbkdf2': 'sha256',
-    # encrypt-then-MAC, kdf: argon2, encryption: AES256-CTR, authentication: HMAC-SHA256
-    'argon2': 'argon2 aes256-ctr hmac-sha256',
+    # encrypt-then-MAC, kdf: argon2, encryption: chacha20, authentication: poly1305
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
@hexagonrecursion ^^^

hexagonrecursion:
I will try to get around to reviewing this today

hexagonrecursion (approved):
:+1: 
1. I support the decision to remove the ability to read 'argon2 aes256-ctr hmac-sha256' keys because we have never made a release that would create them
2. To be honest I do not know what makes CHACHA20_POLY1305 preferable to AES256_CTR_HMAC_SHA256

ThomasWaldmann:
The whole point is to be able to drop aes-ctr completely from the codebase. For the file content chunk encryption we have the more modern ocb and chacha AEAD modes, so we can drop it there also if we make a breaking release.

PR #6613, opened 2022-04-17
pre12-meta cache: do not use the cache if want_unique is True, fixes #6612

### LINKED ISSUE #6612: Deduplicated Size shows 0 bytes with borg 1.2.0
I ran into this issue where borg info shows deduplicated size of 0 B when it should have something for this value.

If I run borg info with 1.1.17, it returns the expected value. This appears to occur for multiple archives in this particular repo. Other repos return the expected value with Borg 1.2

OLD - Borg 1.2.0

```
Archive name: 2022-03-12
Archive fingerprint: 40321bc138667a51ccb444b04a43e5f198ca887fc8aa6d0f03a4ee58056c71b0
Comment:
Hostname: hostname
Username: user
Time (start): Sat, 2022-03-12 17:42:44
Time (end): Sat, 2022-03-12 19:24:08
Duration: 1 hours 41 minutes 23.77 seconds
Number of files: 530936
Command line: /usr/local/bin/borg create --show-rc --info --progress --stats --compression auto,lz4 user@host.example.com:/mnt/backup/bluray/::2022-03-12 /array/bluray
Utilization of maximum supported archive size: 1%
This archive:               44.85 TB             43.48 TB                  0 B
All archives:                1.27 PB              1.23 PB             42.83 TB

Chunk index:                17284323            511548486
```

OLD - Borg 1.1.17

```
Archive name: 2022-03-12
Archive fingerprint: 40321bc138667a51ccb444b04a43e5f198ca887fc8aa6d0f03a4ee58056c71b0
Comment:
Hostname: hostname
Username: user
Time (start): Sat, 2022-03-12 17:42:44
Time (end): Sat, 2022-03-12 19:24:08
Duration: 1 hours 41 minutes 23.77 seconds
Number of files: 530936
Command line: /usr/local/bin/borg create --show-rc --info --progress --stats --compression auto,lz4 user@host.example.com:/mnt/backup/bluray/::2022-03-12 /array/bluray
Utilization of maximum supported archive size: 1%
This archive:               44.85 TB             43.48 TB            157.25 GB
All archives:                1.28 PB              1.24 PB             42.83 TB

Chunk index:                17284323            511548486
```

ThomasWaldmann:


elho:
> depending on want_unique, the cached result is different, thus we need to include that into the cache key.

I wonder, is it worthwhile to store both the partial (without unique count) and full stats and actually recalculate the partial ones when the full ones would be available already, instead of just biting the bullet and when not cached yet just calculating and storing in the cache the full stats?
In total faster (as the partial cache never needs to be built, only one instead of 2 operations per archive are slower than cached), less disk usage for the cache and simpler code.

ThomasWaldmann:
My previous attempt of fixing this was wrong.

The correct fix is that we can not  / must not / do not cache `usize` (unique size of archive).
This is not a static property of an archive, but changes dynamically when chunks get reused by other archives.
So, if we `want_unique`, we must always dynamically calculate it.

PR #6660, opened 2022-04-30
init --other-location=OTHER_REPO: reuse key material from OTHER_REPO

### LINKED ISSUE #6554: initialise a borg repo reusing existing key material
with the new AEAD session key based modes, we do not have counter management issues / nonce reuse issues any more.

at the start of a backup (or other) session, the session key is derived from a random sessionid and the master key material (from borg key). no problem if 2 or more repos use the same master key material in parallel.

one reason to actually use that is replication of archives/chunks from one repo into another (or in general, if one might want to copy or move an archive from one to the other) - that can be done very efficiently, keeping the encrypted chunks "as is", if both repos share the same master key material (including the id key and chunker secret).

so, what we need now is something like:
```
borg init --key-from existing_repo new_repo
```

Keep the same passphrase. If desired, it can be changed afterwards via `borg key change-passphrase`.

Bounty: https://app.bountysource.com/issues/106877923-initialise-a-borg-repo-reusing-existing-key-material (a bit less than usual, but no more funds available)

ThomasWaldmann:
fixes #6554.

it potentially will ask for the passphrase for the key of OTHERREPO.
for the newly created repo, it will use the same passphrase.

it will copy: enc_key, enc_hmac_key, id_key, chunker_seed.

keeping the id_key (and id algorithm) and the chunker seed (and chunker
algorithm and parameters) is desirable for deduplication.
the id algorithm is usually either HMAC-SHA256 or BLAKE2b.

keeping the enc_key / enc_hmac_key must be implemented carefully:
A) AES-CTR -> AES-CTR is INSECURE due to nonce reuse, thus not allowed.
B) AES-CTR -> AEAD with session keys is secure.
C) AEAD with session keys -> AEAD with session keys is secure.

AEAD modes with session keys: AES-OCB and CHACHA20-POLY1305.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -112,6 +112,35 @@ def argument(args, str_or_bool):
     return str_or_bool
 
 
+def get_repository(location, *, create, exclusive, lock_wait, lock, append_only,
+                   make_parent_dirs, storage_quota, args):
+    if location.proto == 'ssh':
+        repository = RemoteRepository(location.omit_archive(), create=create, exclusive=exclusive,
+                                      lock_wait=lock_wait, lock=lock, append_only=append_only,
+                                      make_parent_dirs=make_parent_dirs, args=args)
+
+    else:
+        repository = Repository(location.path, create=create, exclusive=exclusive,
+                                lock_wait=lock_wait, lock=lock, append_only=append_only,
+                                make_parent_dirs=make_parent_dirs, storage_quota=storage_quota)
+    return repository
+
+
+def compat_check(*, create, manifest, key, cache, compatibility, decorator_name):
+    if not create and (manifest or key or cache):
+        if compatibility is None:
+            raise AssertionError(f"{decorator_name} decorator used without compatibility argument")
+        if type(compatibility) is not tuple:
+            raise AssertionError(f"{decorator_name} decorator compatibility argument must be of type tuple")
+    else:
+        if compatibility is not None:
+            raise AssertionError(f"{decorator_name} called with compatibility argument, "
+                                 f"but would not check {compatibility!r}")
+        if create:
+            compatibility = Manifest.NO_OPERATION_CHECK
+    return compatibility
+
+
 def with_repository(fake=False, invert_fake=False, create=False, lock=True,
                     exclusive=False, manifest=True, cache=False, secure=True,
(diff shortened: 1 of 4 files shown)


src/borg/crypto/key.py
```diff
@@ -851,15 +894,27 @@ def decrypt(self, id, data, decompress=True):
...
         # Convert to signed int32
-        if self.chunk_seed & 0x80000000:
-            self.chunk_seed = self.chunk_seed - 0xffffffff - 1
+        if chunk_seed & 0x80000000:
+            chunk_seed = chunk_seed - 0xffffffff - 1
+        self.init_from_given_data(
+            enc_key=data[0:32],
+            enc_hmac_key=data[32:64],
+            id_key=data[64:96],
+            chunk_seed=chunk_seed)
```
ThomasWaldmann, line 917: this is just copy & paste from the other class.

it was duplicated before and still is.

src/borg/archiver.py
```diff
@@ -4361,6 +4392,9 @@ def define_borg_mount(parser):
         subparser.add_argument('location', metavar='REPOSITORY', nargs='?', default='',
                                type=location_validator(archive=False),
                                help='repository to create')
+        subparser.add_argument('--other-repo', metavar='OTHER_REPOSITORY',
+                               type=location_validator(archive=False),
+                               help='reuse the key material from the other repository')
```
ThomasWaldmann, line 4397: guess we'll reuse this "other repository" mechanism for other purposes also, thus the name is rather generic.

src/borg/archiver.py
```diff
@@ -154,20 +179,26 @@ def decorator(method):
...
-            with repository:
+                                        make_parent_dirs=make_parent_dirs, storage_quota=storage_quota,
+                                        args=args)
+
+            other_repository = get_repository(other_location, create=False, exclusive=argument(args, exclusive),
+                                              lock_wait=self.lock_wait, lock=True, append_only=False,
+                                              make_parent_dirs=False, storage_quota=None,
+                                              args=args)
+            if not isinstance(other_repository, DummyRepo):
+                kwargs['other_repository'] = other_repository
```
ThomasWaldmann, line 199: this can also be used for other purposes where we need another repo.

Issue #6666, opened 2022-05-04
Changes in docs to facilitate appropriate crypto use


py0xc3:
At first, thanks for maintaining this great tool!

## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes.

## Is this a BUG / ISSUE report or a QUESTION?

Issue report about content in the docs.

## System information. For client/server mode post info for both machines.

Not relevant. Issue is only related to the content of the docs.

#### Your borg version (borg -V).

Latest Borg docs.

#### Operating system (distribution) and version.

Not relevant. The issue is general.

#### Hardware / network configuration, and filesystems used.

Not relevant. Issue is only related to the content of the docs.

#### How much data is handled by borg?

Not relevant. Issue is only related to the content of the docs.

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

Not relevant. Issue is only related to the content of the docs.

## Describe the problem you're observing.

Issue is not critical / low priority.

I suggest two minor changes in the docs:

1) The docs page about [Security -> Cryptography in Borg](https://borgbackup.readthedocs.io/en/stable/internals/security.html#cryptography-in-borg) contains the following two comments:

```
... HMAC-SHA-256 uses 256 bit keys, while BLAKE2b-256 uses 512 bit keys.

The latter is secure not only because BLAKE2b itself is not susceptible to length extension, but also since it truncates the hash output from 512 bits to 256 bits, which would make the construction safe even if BLAKE2b were broken regarding length extension or similar attacks.
```

Although this is correct, it is highly suggestive: When encryption/authentication is enabled, Borg uses HMAC-SHA256 and not only SHA256. As long as it is used in conjunction with HMAC, there is no risk concerning length extension attacks against SHA256: with HMAC, SHA256 runs twice sequentially with different input each. The length extension resistance created by the first run and the concatenation before the 2nd is comparable to Blake2b's truncation (if truncation will become necessary for Blake2, which I think is not realistic).

[...]

PR #6667, opened 2022-05-05
docs: authentication primitives: improved security and performance infos (master)

py0xc3:
See https://github.com/borgbackup/borg/issues/6666 
1st draft of adjusted security.rst (no changes to init.rst.inc yet).
Shift focus and help to choose the best authentication primitive.


enkore (approved):

docs/internals/security.rst
```diff
@@ -223,14 +223,17 @@ Depending on the chosen mode (see :ref:`borg_init`) different primitives are use
...
-  construction safe even if BLAKE2b were broken regarding length
-  extension or similar attacks.
+  in a keyed mode. 
+
+  Both HMAC-SHA-256 and BLAKE2b have undergone extensive cryptanalysis  
+  and have proven secure against known attacks. The known vulnerability
+  of SHA-256 against length extension attacks does not apply to HMAC-SHA-256.
+  
+  The authentication primitive should be chosen based upon SHA hardware support:
+  AMD Ryzen and new 12th+ Generation Intel Core processors (beginning with
```
ThomasWaldmann, line 233: I checked the CPU support and according to these sources, we should change this a bit:

https://en.wikichip.org/wiki/intel/microarchitectures/ice_lake_(client) -> "10th+ generation / Ice Lake"

https://en.wikichip.org/wiki/amd/microarchitectures/zen -> "Zen/Ryzen/EPYC"

docs/internals/security.rst
```diff
@@ -223,14 +223,17 @@ Depending on the chosen mode (see :ref:`borg_init`) different primitives are use
...
+  in a keyed mode. 
+
+  Both HMAC-SHA-256 and BLAKE2b have undergone extensive cryptanalysis  
+  and have proven secure against known attacks. The known vulnerability
+  of SHA-256 against length extension attacks does not apply to HMAC-SHA-256.
+  
+  The authentication primitive should be chosen based upon SHA hardware support:
+  AMD Ryzen and new 12th+ Generation Intel Core processors (beginning with 
+  the Alder Lake architecture) support SHA extensions and are likely to perform
+  best with HMAC-SHA-256. Among ARM processors, support for SHA extensions varies.
```
ThomasWaldmann, line 235: Looking at the benchmarks, I guess we can tell that Apple M1 (and likely also later) CPUs have SHA acceleration.

This is notable, considering that Apple completely abandons x64 architecture, so that in a few years  all Apple devices will have these CPUs.

ThomasWaldmann (changes requested):
Thanks for your PR. I reviewed the more practical part.

py0xc3:
Interesting phenomenon. Ice Lake has indeed already the extensions, but I cannot find information on Intel.com that its successor Tiger Lake has it as well. But I think we can assume it contains them, too. Apple's processors are also worth to be named on themselves given the widespread use. Thanks for the note.

However, I think your approach of benchmark makes more sense on the long term: the user shall make a benchmark and use what is fastest. `lscpu` would be an alternative to identify the extensions, although this works not on all OS. 

But maybe it makes sense to enable a benchmark that does not need the creation of a repo in advance, and then replace my whole adjustment about the architectures and their extensions by one or two sentences about how and why to make the benchmark? So, facilitating an initial benchmark as first step before creating the first borg repo on a system.

I can write an adjustment in the next few days.

docs/internals/security.rst
enkore: Note that ice lake was limited to mobile, so it's not just the generation number, but also which kind of SKU, which determines uarch and thus ISA extensions. For desktops 12th+ is correct.

enkore: *actually it seems rocket lake already included it - so 10th+ for mobile, 11th+ for desktop should be right.

enkore:
There's "borg benchmark cpu" specifically for doing those tests. I don't know if it does, but a nice thing there would be to just plainly tell you which mode is quickest on your system, and associated caveats if any.

py0xc3:
> There's "borg benchmark cpu" specifically for doing those tests. I don't know if it does, but a nice thing there would be to just plainly tell you which mode is quickest on your system, and associated caveats if any.

My point was that "borg benchmark *" seems to [need a repo that already exists](https://borgbackup.readthedocs.io/en/stable/usage/benchmark.html) before it can run. So, a user would need to create a repo, then do the benchmark, and then potentially delete the repo again to create a new one with a different primitive. Maybe this can be simplified in future versions? It would be definitely the better approach to just let the user know how to do the benchmark and then chose the primitive that is fastest.

"borg benchmark cpu" leads to the error "borg benchmark: error: argument <command>: invalid choice: 'cpu' (choose from 'crud')" on my system, but I expect this is because I have currently no borg repo.

docs/internals/security.rst
py0xc3: updated.

docs/internals/security.rst
py0xc3: updated.

ThomasWaldmann:
borg benchmark cpu is only in master branch and does not require a repo. it does pure algorithmic tests without I/O.

py0xc3:
> borg benchmark cpu is only in master branch and does not require a repo. it does pure algorithmic tests without I/O.

I was not aware of this :) In this case, it makes sense to already prepare the docs in master branch for that. I will delete all the architecture stuff and simply refer to how to use and interpret "borg benchmark cpu".

ThomasWaldmann:
You could keep the architectural stuff for a 1.2-maint PR.

Yes, for master, you can just refer to `borg benchmark cpu`.

py0xc3:
security.rst is updated & "borg benchmark cpu" is introduced. You may verify my syntax with commands (such as "borg benchmark cpu") because my experience with readthedocs is limited.

ThomasWaldmann (approved):
LGTM

ThomasWaldmann:
thanks!

PR #6663, opened 2022-05-03
transfer archives from another repo

ThomasWaldmann:
transfers and also cleans up legacy stuff / converts to future, see commit comments.

developed this as an alternative approach to the advanced tar-pipe stuff.

guess this is the better approach.

PR #6668, opened 2022-05-05
cleanup msgpack related str/bytes mess

ThomasWaldmann:
based on #6663, see the commits not prefixed with `transfer:`.


ThomasWaldmann:
will redo this PR against borg2 branch.

PR #6685, opened 2022-05-08
fix create_filter_process exception handler, fixes #6681

### LINKED ISSUE #6681: crashing with UnboundLocalError in create_filter_process
```
        try:
            yield stream
    
        except Exception:
            # something went wrong with processing the stream by borg
            logger.debug('Exception, killing the filter...')
>           proc.kill()
E           UnboundLocalError: local variable 'proc' referenced before assignment

src/borg/helpers/process.py:335: UnboundLocalError
```

Guess cmd can be None and there should be a `if cmd:`.

ThomasWaldmann:
if cmd was falsy (e.g. None), there is no proc.
then, if "yield stream" raises an exception, the exception handler crashed at "proc.kill()".

### DIFF
--- a/src/borg/helpers/process.py
+++ b/src/borg/helpers/process.py
@@ -332,7 +332,8 @@ def create_filter_process(cmd, stream, stream_close, inbound=True):
     except Exception:
         # something went wrong with processing the stream by borg
         logger.debug('Exception, killing the filter...')
-        proc.kill()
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
note: no such code in 1.1-maint, thus only a 1.2-maint backport.

PR #6689, opened 2022-05-12
repoindex improvements

ThomasWaldmann:
improvements:
- NSIndex: adds the payload size (aka `csize`) and `extra` to the repo index
- NSIndex1: keeps the repo index code for old borg
- open_index peeks into the hashindex header to find out what's old and what's new.

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -13,7 +13,7 @@ from functools import partial
 from itertools import islice
 
 from .constants import *  # NOQA
-from .hashindex import NSIndex
+from .hashindex import NSIndexEntry, NSIndex, NSIndex1, hashindex_variant
 from .helpers import Error, ErrorWithTraceback, IntegrityError, format_file_size, parse_file_size
 from .helpers import Location
 from .helpers import ProgressIndicatorPercent
@@ -50,6 +50,18 @@ TAG_PUT2 = 3
 MAX_TAG_ID = 15
 
 FreeSpace = partial(defaultdict, int)
+
+
+def header_size(tag):
+    if tag == TAG_PUT2:
+        size = LoggedIO.HEADER_ID_SIZE + LoggedIO.ENTRY_HASH_SIZE
+    elif tag == TAG_PUT or tag == TAG_DELETE:
+        size = LoggedIO.HEADER_ID_SIZE
+    elif tag == TAG_COMMIT:
+        size = LoggedIO.header_fmt.size
+    else:
+        raise ValueError(f"unsupported tag: {tag!r}")
+    return size
 
 
 class Repository:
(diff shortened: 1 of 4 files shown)


src/borg/repository.py
```diff
@@ -1504,8 +1511,9 @@ def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
             # we are touching this segment for the first time, check the MAGIC.
             # Repository.scan() calls us with segment > 0 when it continues an ongoing iteration
             # from a marker position - but then we have checked the magic before already.
-            if fd.read(MAGIC_LEN) != MAGIC:
-                raise IntegrityError(f'Invalid segment magic [segment {segment}, offset {0}]')
+            mr = fd.read(MAGIC_LEN)
+            if mr != MAGIC:
+                raise IntegrityError(f'Invalid segment magic [segment {segment}, offset {0}, magic {mr!r}]')
```
ThomasWaldmann, line 1516: one test fails here with `mr == b''`.

src/borg/repository.py
```diff
@@ -1515,7 +1523,8 @@ def iter_objects(self, segment, offset=0, include_data=False, read_data=True):
             if include_data:
                 yield tag, key, offset, data
             else:
-                yield tag, key, offset, size
+                yield tag, key, offset, size - header_size(tag)  # corresponds to len(data)
+            assert size >= 0
```
ThomasWaldmann, line 1527: i changed this so that it either returns the `data` (so caller could do `len(data)`) or it does not return the data, but precisely `len(data)` without any header overheads.

ThomasWaldmann: if callers now want the size **with** overhead, they add `header_size(tag)` again.

src/borg/repository.py
ThomasWaldmann: does anybody see why?

ThomasWaldmann: https://github.com/borgbackup/borg/runs/6403364221?check_suite_focus=true#step:9:3258

ThomasWaldmann: guess i found it, see commits.

src/borg/repository.py
```diff
@@ -1448,6 +1455,9 @@ def clean_old():
                         del self.fds[k]
 
         clean_old()
+        if self._write_fd is not None:
+            # without this, we have test failure now
+            self._write_fd.sync()
```
ThomasWaldmann, line 1460: this was needed now to avoid a test failure.
otherwise it could not read the MAGIC back from a recently created segment file.

ThomasWaldmann: This is done a bit more than necessary, sufficient would be:
```
self._write_fd.f.flush()
```

But this is not available yet via SyncFile api.

Issue #6691, opened 2022-05-13
remove scp syntax


ThomasWaldmann:
borg uses regexes to parse the location argument given to it on the commandline.

these regexes are a rather complex and at least the scp-style location regex also tends to mismatches, because it is used for:
- local repos (just giving a fs path as the repo)
- remote repos (which can be also given via the `ssh://` url syntax, which is less limited and clearly a remote repo)

that the user often can also give an archive name within a LOCATION specifier makes it even a bit more complex, but that is the topic of #948.

one more bad thing about scp-style is that one can not give a ssh port within it. so we also need docs / FAQ entries telling the user to not use that style if the port is not 22. also we need examples for both styles...

so, the plan is to just remove scp-style syntax. less code, less docs, less hard to debug/fix regex issues.

obviously this is a breaking change as users who use that style will have to edit their scripts / configs to use ssh url style.

bounty: https://app.bountysource.com/issues/108445136-remove-scp-syntax

ThomasWaldmann:
btw, in case somebody is wondering about the USD 41.85 bounty amount: that was just the remainder before the organisation funds went to zero.

ThomasWaldmann:
Crap, I was a little fast with merging this.

The idea was to merge this IF the other breaking changes are also merged.
But anyway, either we do that or we'll revert this change later again.

PR #6697, opened 2022-05-15
Remove scp syntax for locations

emilmgeorge:
For #6691.
I have removed the scp part of the regex while keeping the local file part. Also removed the scp syntax from testsuite, docs.
Commits are not yet squashed for easier review.

### DIFF
--- a/docs/deployment/central-backup-server.rst
+++ b/docs/deployment/central-backup-server.rst
@@ -81,7 +81,7 @@ The options which are added to the key will perform the following:
 Due to the ``cd`` command we use, the server automatically changes the current
 working directory. Then client doesn't need to have knowledge of the absolute
 or relative remote repository path and can directly access the repositories at
-``<user>@<host>:<repo>``.
+``ssh://<user>@<host>/<repo>``.
 
 .. note:: The setup above ignores all client given commandline parameters
           which are normally appended to the `borg serve` command.
@@ -93,21 +93,21 @@ The client needs to initialize the `pictures` repository like this:
 
 ::
 
- borg init backup@backup01.srv.local:pictures
+ borg init ssh://backup@backup01.srv.local/./pictures
 
 Or with the full path (should actually never be used, as only for demonstrational purposes).
 The server should automatically change the current working directory to the `<client fqdn>` folder.
 
 ::
 
-  borg init backup@backup01.srv.local:/home/backup/repos/johndoe.clnt.local/pictures
+  borg init ssh://backup@backup01.srv.local/home/backup/repos/johndoe.clnt.local/pictures
 
 When `johndoe.clnt.local` tries to access a not restricted path the following error is raised.
 John Doe tries to backup into the Web 01 path:
 
 ::
 
-  borg init backup@backup01.srv.local:/home/backup/repos/web01.srv.local/pictures
+  borg init ssh://backup@backup01.srv.local/home/backup/repos/web01.srv.local/pictures
 
 ::
 
--- a/docs/deployment/central-backup-server.rst
+++ b/docs/deployment/central-backup-server.rst
@@ -81,7 +81,7 @@ The options which are added to the key will perform the following:
 Due to the ``cd`` command we use, the server automatically changes the current
 working directory. Then client doesn't need to have knowledge of the absolute
 or relative remote repository path and can directly access the repositories at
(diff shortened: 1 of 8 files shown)


ThomasWaldmann (changes requested):
Thanks for the PR!

Looks quite good, but guess you missed some of the docs showing examples in scp style, these need to get fixed also.:
- docs/deployment/central-backup-server.rst
- docs/quickstart.rst
- docs/usage/init.rst

ThomasWaldmann:
src/borg/helpers/parseformat.py:365:13: E126 continuation line over-indented for hanging indent

emilmgeorge:
Thanks, fixed!

ThomasWaldmann:
While you're at it, could you add this:
```
Converting from scp-style repo URLs
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Borg does not support scp-style repo URLs any more.

Here is how you can convert to URL style:

::

    user@host:path/to/repo             # relative to cwd
    -->
    ssh://user@host:22/./path/to/repo  # relative to cwd
    or (usually the cwd is the home dir after ssh login)
    ssh://user@host:22/~/path/to/repo  # relative to home dir

    user@host:/path/to/repo             # absolute repo path
    -->
    ssh://user@host:22/path/to/repo     # absolute repo path

Notes:

Port 22 is the default, so you can omit the ``:22`` if you like.

If you used some hack to use a non-standard port (which was not directly
supported by the scp-style repo specification), you can now do it that way
with the ``ssh:`` URL and remove the hack (usually in ssh configuration or
given via ``--rsh`` borg option).
```

docs/deployment/central-backup-server.rst
```diff
@@ -81,7 +81,7 @@ The options which are added to the key will perform the following:
 Due to the ``cd`` command we use, the server automatically changes the current
 working directory. Then client doesn't need to have knowledge of the absolute
 or relative remote repository path and can directly access the repositories at
-``<user>@<host>:<repo>``.
+``ssh://<user>@<host>/<repo>``.
```
ThomasWaldmann, line 84: Hmm, considering the sentence in line 82/83, this needs to stay a relative-to-cwd path, just use the ssh url syntax.

docs/usage/general/repository-urls.rst.inc
```diff
@@ -52,3 +36,31 @@ to use the default - it will be read from BORG_REPO then.
...
+
+    user@host:path/to/repo             # relative to cwd
+    -->
+    ssh://user@host:22/./path/to/repo  # relative to cwd
+    or (usually the cwd is the home dir after ssh login)
+    ssh://user@host:22/~/path/to/repo  # relative to home dir
+
+    user@host:/path/to/repo             # absolute repo path
+    -->
+    ssh://user@host:22/path/to/repo     # absolute repo path
```
ThomasWaldmann, line 57: can you check/fix column of the `#` being the same for these lines?

docs/usage/init.rst
```diff
@@ -15,8 +15,8 @@ Examples
 
     # Remote repository (accesses a remote borg via ssh)
     # repokey: stores the (encrypted) key into <REPO_DIR>/config
-    $ borg init --encryption=repokey-aes-ocb user@hostname:backup
+    $ borg init --encryption=repokey-aes-ocb ssh://user@hostname/backup
 
     # Remote repository (accesses a remote borg via ssh)
     # keyfile: stores the (encrypted) key into ~/.config/borg/keys/
-    $ borg init --encryption=keyfile-aes-ocb user@hostname:backup
+    $ borg init --encryption=keyfile-aes-ocb ssh://user@hostname/backup
```
ThomasWaldmann, line 22: relative-to-cwd needed

ThomasWaldmann (changes requested):
discovered some more issues. but guess we're done after fixing these.

docs/deployment/central-backup-server.rst
emilmgeorge: Done

ThomasWaldmann (approved):
LGTM

ThomasWaldmann:
Thanks!

ThomasWaldmann:
Crap, I was a little fast with merging this.

The idea was to merge this IF the other breaking changes are also merged.
But anyway, either we do that or we'll revert this change later again.

PR #6703, opened 2022-05-18
borg2: "borg transfer" cmd (and also getting rid of legacy)

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -338,6 +339,137 @@ def do_serve(self, args):
...
+
+        args.consider_checkpoints = True
+        archive_names = tuple(x.name for x in other_manifest.archives.list_considering(args))
+        if not archive_names:
+            return EXIT_SUCCESS
+
+        for name in archive_names:
+            transfer_size = 0
+            present_size = 0
+            if name in manifest.archives and not dry_run:
```
elho, line 420: I think that message should still be printed in a dry run, no?
I would expecpt to learn which archives would be skipped due to already being present during a dry-run and that not behave differently (going into the `else` branch below) than the normal run.

ThomasWaldmann:
i intentionally let it process all archives.

it is nice to see that really all chunks are present, kind of verification.

src/borg/archiver.py
ThomasWaldmann: i intentionally let it process all archives.

it is nice to see that really all chunks are present, kind of a reassuring verification and not just a skip based on the name.

PR #6705, opened 2022-05-18
Repoindex

ThomasWaldmann:


ThomasWaldmann:
Hmm, instead of giving a bit number and a 0/1 value, we could also give a bit mask and a value to put there / to expect there. That would enable a very efficient iteration over items that have a specific flags combination.

Also, it would not be limited to 1-bit fields.

OTOH, harder to use.

Issue #6407, opened 2022-03-05
borg 1.2 regular expression pattern behavior change


julian-klode:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Seems like a bug as it's not announced behaviour change, and dangerous as it suddenly includes files (or no longer includes files) it has included befor.

#### Your borg version (borg -V).

1.2 locally, no idea remote

#### Operating system (distribution) and version.

Ubuntu 22.04

#### Hardware / network configuration, and filesystems used.

-

#### How much data is handled by borg?

-

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

`borg create --progress --one-file-system --exclude-caches --info --stats --exclude-from=/home/jak/.config/borg/exclude.conf -C zlib,9 $repostory::$archive /home/jak`

## Describe the problem you're observing.

`re:` patterns no longer apply to the full filename, but to the relative one in the archive. Meaning I suddenly had all my regex excluded files pop up in my backups, such as files in `/home/jak/mail/.notmuch` which are huge. I had to remove the leading `/` to get the pattern to match again, as can be seen in https://github.com/julian-klode/dotfiles/commit/7a6f1e0437fe96311c59a38cbf2c743043941e31 and in this excerpt:

```diff
-re:/home/jak/.*/\.notmuch
+re:home/jak/.*/\.notmuch
```

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

N/A

julian-klode:
Smaller reproducer:

```
jak@jak-t480s:/tmp$ mkdir in out
jak@jak-t480s:/tmp$ borg init -e none out
jak@jak-t480s:/tmp$ touch in/foo
jak@jak-t480s:/tmp$ touch in/fooz
jak@jak-t480s:/tmp$ touch in/bar
jak@jak-t480s:/tmp$ borg create --exclude "re:/tmp/in/foo" out::slash /tmp/in/
jak@jak-t480s:/tmp$ borg create --exclude "re:tmp/in/foo" out::noslash /tmp/in/
jak@jak-t480s:/tmp$ borg list out::slash
drwxrwxr-x jak    jak           0 Sat, 2022-03-05 19:41:53 tmp/in
-rw-rw-r-- jak    jak           0 Sat, 2022-03-05 19:41:51 tmp/in/foo
[...]

PR #6718, opened 2022-05-27
#6407 - Document Borg 1.2 pattern behavior change

fantasya-pbem:
Backport from master: Make clear that absolute paths always go into the matcher as if they are relative (without leading slash). Adapt all examples accordingly.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2244,15 +2244,12 @@ class Archiver:
         currently active recursion root. You usually give the recursion root(s)
         when invoking borg and these can be either relative or absolute paths.
 
-        So, when you give `relative/` as root, the paths going into the matcher
-        will look like `relative/.../file.ext`. When you give `/absolute/` as
-        root, they will look like `/absolute/.../file.ext`.
-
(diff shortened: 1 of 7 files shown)


src/borg/archiver.py
```diff
@@ -2431,34 +2429,34 @@ def do_break_lock(self, args, repository):
...
 
         The patternfile::
 
             # note that excludes use fm: by default and patternfiles use sh: by default.
             # therefore, we need to specify fm: to have the same exact behavior.
             P fm
-            R /home/bobby
-            R /home/susan
+            R home/bobby
+            R home/susan
```
ThomasWaldmann, line 2457: oops. guess i missed that in the last review. this is not a pattern, but a recursion root, guess it must not be changed.

fantasya-pbem: Shall I fix that in master branch, too?

ThomasWaldmann: yes, please.

fantasya-pbem: Fixed & new PR for master.

ThomasWaldmann (approved):

PR #5777, opened 2021-04-19
do not load files cache for commands not using it, fixes #5673

ThomasWaldmann:

PR #6724, opened 2022-05-28
make constants for files cache mode more clear

elho:
Traditionally, DEFAULT_FILES_CACHE_MODE_UI and DEFAULT_FILES_CACHE_MODE
were - as the naming scheme implies - the same setting, one being the UI
representation as given to the --files-cache command line option and the
other being the same default value in the internal representation.

It happended that the actual value used in borg create always comes from
DEFAULT_FILES_CACHE_MODE_UI (because that does have the --files-cache
option) whereas for all other commands (that do not use the files cache) it
comes from DEFAULT_FILES_CACHE_MODE.

PR #5777 then abused this fact to implement the optimisation to skip loading
of the files cache in those other commands by changing the value of
DEFAULT_FILES_CACHE_MODE to disabled.
This however also changes the meaning of that variable and thus redesignates
it to something not matching the original naming anymore.

Anyone not aware of this change and the intention behind it looking at the
code would have a hard time figuring this out and be easily mislead.

This does away with the confusion making the code more maintainable by
renaming DEFAULT_FILES_CACHE_MODE to FILES_CACHE_MODE_DISABLED, making the
new intention of that internal default clear.

Thank you for contributing code to Borg, your help is appreciated!

Please, before you submit a pull request, make sure it complies with the
guidelines given in our documentation:

https://borgbackup.readthedocs.io/en/latest/development.html#contributions

**Please remove all above text before submitting your pull request.**

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3799,8 +3799,8 @@ class Archiver:
         fs_group.add_argument('--sparse', dest='sparse', action='store_true',
                                help='detect sparse holes in input (supported only by fixed chunker)')
         fs_group.add_argument('--files-cache', metavar='MODE', dest='files_cache_mode', action=Highlander,
-                              type=FilesCacheMode, default=DEFAULT_FILES_CACHE_MODE_UI,
-                              help='operate files cache in MODE. default: %s' % DEFAULT_FILES_CACHE_MODE_UI)
+                              type=FilesCacheMode, default=FILES_CACHE_MODE_UI_DEFAULT,
+                              help='operate files cache in MODE. default: %s' % FILES_CACHE_MODE_UI_DEFAULT)
         fs_group.add_argument('--read-special', dest='read_special', action='store_true',
                               help='open and read block and char device files as well as FIFOs as if they were '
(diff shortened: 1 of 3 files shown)


elho:
This is the "minimal" version, it may make sense to extend it by also renaming `DEFAULT_FILES_CACHE_MODE_UI` by sneaking a `CREATE` in there, e.g. `FILES_CACHE_MODE_UI_CREATE_DEFAULT` to make that one even more clear, too. Thoughts?

ThomasWaldmann:
FILES_CACHE_MODE_UI_CREATE_DEFAULT - good idea, but as we currently **only** use it for create and also we do not know yet if / how it would be used elsewhere, i guess we can leave away the "CREATE" for now.

ThomasWaldmann:
@elho are you finished with this or do you want to make changes?

elho:
> @elho are you finished with this or do you want to make changes?

Oh, I took your previous comment as "leave DEFAULT_FILES_CACHE_MODE_UI alone", but I'll happily still bring it in line with the hierarchical scheme just leaving the CERATE part out...

ThomasWaldmann (approved):

PR #6726, opened 2022-05-28
Use relative imports

elho:
Make unnecessary absolute imports relative instead, both to make code style more consistent and reduce the code parts a potential future module rename, e.g. for a breaking borg2, would have to touch.


ThomasWaldmann:
linter is unhappy.

ThomasWaldmann:
Thanks!

PR #6741, opened 2022-06-02
[DOCS] #5310 - Overhaul borg help patterns

### LINKED ISSUE #5310: How to back up only certain files (e.g. *.jpg)

fantasya-pbem:
My first shot to improve the patterns help a bit. I tried to keep it short.
- Start with explaination where patterns can be used.
- Emphasize that leading separator is trimmed.
- List patterns file prefixes one by one to make it more readable.
- Add the tip for testing patterns.
- Keep examples short.

I haven't changed existing examples a lot, but I tested all aspects with different pattern styles.

fixes #5310

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2485,40 +2485,35 @@ class Archiver:
 
     helptext = collections.OrderedDict()
     helptext['patterns'] = textwrap.dedent('''
-        The path/filenames used as input for the pattern matching start from the
-        currently active recursion root. You usually give the recursion root(s)
-        when invoking borg and these can be either relative or absolute paths.
+        When specifying one or more file paths in a Borg command that supports
+        patterns for the respective option or argument, you can apply the
+        patterns described here to include only desired files and/or exclude
+        unwanted ones. Patterns can be used
 
-        If you give `/absolute/` as root, the paths going into the matcher will
-        look relative like `absolute/.../file.ext`, because file paths in Borg
-        archives are always stored normalized and relative. This means that e.g.
-        ``borg create /path/to/repo ../some/path`` will store all files as
-        `some/path/.../file.ext` and ``borg create /path/to/repo /home/user``
-        will store all files as `home/user/.../file.ext`.
+        - for ``--exclude`` option,
+        - in the file given with ``--exclude-from`` option,
+        - for ``--pattern`` option,
+        - in the file given with ``--patterns-from`` option and
+        - for ``PATH`` arguments that explicitly support them.
 
-        A directory exclusion pattern can end either with or without a slash ('/').
-        If it ends with a slash, such as `some/path/`, the directory will be
-        included but not its content. If it does not end with a slash, such as
-        `some/path`, both the directory and content will be excluded.
+        Borg always stores all file paths normalized and relative to the
+        current recursion root. The recursion root is also named ``PATH`` in
+        Borg commands like `borg create` that do a file discovery, so do not
+        confuse the root with the ``PATH`` argument of e.g. `borg extract`.
 
-        File patterns support these styles: fnmatch, shell, regular expressions,
-        path prefixes and path full-matches. By default, fnmatch is used for
-        ``--exclude`` patterns and shell-style is used for the ``--pattern``
-        option. For commands that support patterns in their ``PATH`` argument
-        like (``borg list``), the default pattern is path prefix.
+        Starting with Borg 1.2, paths that are matched against patterns always
+        appear relative. If you give ``/absolute/`` as root, the paths going
+        into the matcher will look relative like ``absolute/.../file.ext``.
+        If you give ``../some/path`` as root, the paths will look like
+        ``some/path/.../file.ext``.
 
-        Starting with Borg 1.2, discovered fs paths are normalised, have leading
-        slashes removed and then are matched against your patterns.
-        Note: You need to review your include / exclude patterns and make
-        sure they do not expect leading slashes. Borg can only deal with this
-        for some very simple patterns by removing leading slashes there also.
+        File patterns support five different styles. If followed by a colon ':',
+        the first two characters of a pattern are used as a style selector.
+        Explicit style selection is necessary if a non-default style is desired
+        or when the desired pattern starts with two alphanumeric characters
+        followed by a colon (i.e. ``aa:something/*``).
 
-        If followed by a colon (':') the first two characters of a pattern are
-        used as a style selector. Explicit style selection is necessary when a
-        non-default style is desired or when the desired pattern starts with
-        two alphanumeric characters followed by a colon (i.e. `aa:something/*`).
-
-        `Fnmatch <https://docs.python.org/3/library/fnmatch.html>`_, selector `fm:`
+        `Fnmatch <https://docs.python.org/3/library/fnmatch.html>`_, selector ``fm:``
             This is the default style for ``--exclude`` and ``--exclude-from``.
             These patterns use a variant of shell pattern syntax, with '\\*' matching
             any number of characters, '?' matching any single character, '[...]'
@@ -2526,7 +2521,7 @@ class Archiver:
             matching any character not specified. For the purpose of these patterns,
             the path separator (backslash for Windows and '/' on other systems) is not
             treated specially. Wrap meta-characters in brackets for a literal
-            match (i.e. `[?]` to match the literal character `?`). For a path
+            match (i.e. ``[?]`` to match the literal character '?'). For a path
             to match a pattern, the full path must match, or it must match
             from the start of the full path to just before a path separator. Except
             for the root path, paths will never end in the path separator when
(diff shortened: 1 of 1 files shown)


fantasya-pbem:
Of course I checked "build_usage", "build_man" and "make html"...

src/borg/archiver.py
```diff
@@ -2626,34 +2626,56 @@ def do_break_lock(self, args, repository):
...
+        A more general and easier to use way to define filename matching patterns
+        exists with the ``--pattern`` and ``--patterns-from`` options. Using
+        these, you may specify the backup roots, default pattern styles and
+        patterns for inclusion/exclusion.
 
-        .. note::
+        Root path prefix ``R``
+            A root path starts with the prefix `R`, followed by a path (a plain
+            path, not a file pattern). Use this prefix to move root paths from
+            command line arguments to a patterns file.
```
ThomasWaldmann, line 2637: maybe it should use "recursion root paths" at least in the explanation.

src/borg/archiver.py
```diff
@@ -2626,34 +2626,56 @@ def do_break_lock(self, args, repository):
...
+        Exclude no-recurse pattern prefix ``!``
+            Use the prefix `!`, followed by a pattern, to define an exclusion
+            that does not recurse into subdirectories.
+
+        Inclusion patterns are useful to include paths that are contained in an
+        excluded path. The first matching pattern is used. so if an include
+        pattern matches before an exclude pattern, the file is backed up. If an
+        exclude-norecurse pattern matches a directory, it won't recurse into it
+        (which saves time), but won't discover any potential matches for include
+        rules below that directory.
```
ThomasWaldmann, line 2664: change order:

1. exclude
2. exclude no-recurse
3. include

and then explain why include is useful (maybe even within the include explanation).

src/borg/archiver.py
```diff
@@ -2626,34 +2626,56 @@ def do_break_lock(self, args, repository):
...
-        patterns. The pattern style can be set via the `P` prefix.
+        This will print the considered files one per line, prefixed with a
+        character that indicates the action (e.g. 'x' for excluding, see
+        **Item flags** in `borg create` usage docs).
+
+        .. note::
+
+            It's possible that a sub-directory/file is matched while parent
+            directories are not. In that case, parent directories are not backed
+            up thus their user, group, permission, etc. can not be restored.
```
ThomasWaldmann, line 2678: ... backed up **and** thus ... ?

src/borg/archiver.py
```diff
@@ -2665,42 +2687,40 @@ def do_break_lock(self, args, repository):
             # note: using = is essential to avoid cmdline argument parsing issues.
             borg create --pattern=+pics/2018/good --pattern=-pics/2018 repo::arch pics
 
+            # backup only JPG/JPEG files (case insensitive) in all home directories:
+            borg create --pattern 're:\\.jpe?g(?i)' repo::arch /home
+
```
ThomasWaldmann, line 2692: maybe a more realistic example would be to exclude all *.ISO (they usually are downloads **and** big).

guess most people do not **only** want to backup their jpegs, but also other files in their homes.

src/borg/archiver.py
```diff
@@ -2665,42 +2687,40 @@ def do_break_lock(self, args, repository):
...
             # use a file with patterns:
-            borg create --patterns-from patterns.lst repo::arch
+            borg create --patterns-from patterns.lst repo::arch /
```
ThomasWaldmann, line 2694: is the / required? the recursion root can also be given in the patterns.lst.

src/borg/archiver.py
```diff
@@ -2665,42 +2687,40 @@ def do_break_lock(self, args, repository):
...
 
         The patterns.lst file could look like that::
 
-            # "sh:" pattern style is the default, so the following line is not needed:
+            # sh: pattern style is the default, so the following line is not needed:
```
ThomasWaldmann, line 2698: is that an improvement? i guess the quotes were there intentionally because the colon is not punctuation in that sentence.

src/borg/archiver.py
```diff
@@ -2665,42 +2687,40 @@ def do_break_lock(self, args, repository):
...
             # include susans home
             + home/susan
             # also back up this exact file
             + pf:home/bobby/specialfile.txt
             # don't backup the other home directories
             - home/*
-            # don't even look in /proc
-            ! proc
+            # don't even look in /dev, /proc, /run, /sys, /tmp
+            ! re:^(dev|proc|run|sys|tmp)/
```
ThomasWaldmann, line 2709: if there is a slash at the end, borg might try to look "into" these directories, because the directory was not matched by the expression (yet), I guess?

src/borg/archiver.py
```diff
@@ -2626,34 +2626,56 @@ def do_break_lock(self, args, repository):
...
 
-            It's possible that a sub-directory/file is matched while parent directories are not.
-            In that case, parent directories are not backed up thus their user, group, permission,
-            etc. can not be restored.
+            $ borg create --dry-run --list --patterns-from patterns.txt /repo::arch
 
-        Note that the default pattern style for ``--pattern`` and ``--patterns-from`` is
-        shell style (`sh:`), so those patterns behave similar to rsync include/exclude
-        patterns. The pattern style can be set via the `P` prefix.
+        This will print the considered files one per line, prefixed with a
```
ThomasWaldmann, line 2670: s/print/list/

ThomasWaldmann (commented):
thanks for the PR, some feedback.

src/borg/archiver.py
```diff
@@ -2485,35 +2485,36 @@ def do_break_lock(self, args, repository):
...
-        Starting with Borg 1.2, discovered fs paths are normalised, have leading
-        slashes removed and then are matched against your patterns.
-        Note: You need to review your include / exclude patterns and make
-        sure they do not expect leading slashes. Borg can only deal with this
-        for some very simple patterns by removing leading slashes there also.
-
-        If followed by a colon (':') the first two characters of a pattern are
+        When specifying one or more file paths in a Borg command that supports
+        patterns for the respective option or argument, you can apply the
+        patterns described here to filter out desired files. Patterns can be
```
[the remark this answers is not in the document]
fantasya-pbem: I'll rephrase this to "...patterns described here to include only desired files and/or exclude unwanted ones."

src/borg/archiver.py
```diff
@@ -2485,35 +2485,36 @@ def do_break_lock(self, args, repository):
...
+
+        - for ``--exclude`` option,
+        - in the file given with ``--exclude-from`` option,
+        - for ``--pattern`` option,
+        - in the file given with ``--patterns-from`` option and
+        - for ``PATH`` arguments that explicitly support them.
+
+        Borg always stores all file paths normalized and relative to the
+        current recursion root. The recursion root is also named ``PATH`` in
+        Borg commands that do a file search like `borg create`, so don't
```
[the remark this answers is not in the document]
fantasya-pbem: Discovery sounds good.

There has always been a mix of double/single backticks. I'm not sure if we should use only one or the other.

src/borg/archiver.py
```diff
@@ -2485,35 +2485,36 @@ def do_break_lock(self, args, repository):
...
+        Starting with Borg 1.2, paths that are matched against patterns always
+        appear relative. If you give ``/absolute/`` as root, the paths going
+        into the matcher will look relative like `absolute/.../file.ext`. If you
+        give ``../some/path`` as root, the paths will look like
+        `some/path/.../file.ext`.
+
+        File patterns support these styles: **fnmatch**, **shell**, **regular**
+        **expressions**, **path prefixes** and **path full-matches**. By default,
+        fnmatch is used for ``--exclude`` patterns and shell style is used for
+        the ``--pattern`` option. For commands that support patterns in their
```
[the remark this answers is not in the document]
fantasya-pbem: Yes, this can be shortened a bit.

src/borg/archiver.py
```diff
@@ -2540,16 +2541,15 @@ def do_break_lock(self, args, repository):
             is that the pattern may include `**/` for matching zero or more directory
             levels, `*` for matching zero or more arbitrary characters with the
             exception of any path separator. A leading path separator is always removed.
+            Shell-style patterns behave similar to rsync include/exclude patterns.
```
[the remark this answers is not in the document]
fantasya-pbem: I agree, we should not reference to rsync. It does not clarify anything here.

src/borg/archiver.py
```diff
@@ -2540,16 +2541,15 @@ def do_break_lock(self, args, repository):
...
 
-        Regular expressions, selector `re:`
+        `Regular expressions <https://docs.python.org/3/library/re.html>`_, selector `re:`
             Regular expressions similar to those found in Perl are supported. Unlike
```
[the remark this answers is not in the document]
fantasya-pbem: ok

src/borg/archiver.py
fantasya-pbem: I didn't find the right words. Nice!

src/borg/archiver.py
fantasya-pbem: ok

src/borg/archiver.py
fantasya-pbem: ok

src/borg/archiver.py
fantasya-pbem: ok

src/borg/archiver.py
fantasya-pbem: I just realize that there is a missing '+' in 're:'...
I wanted to include a simple example for the use case of #5310 question (back up files of distinct extension).
Of course, we can add more common useful examples. I'll add something...

src/borg/archiver.py
fantasya-pbem: It's not required, but showcases a patterns file without R prefix. The other example below features roots in the file only.

src/borg/archiver.py
fantasya-pbem: ok, reverting

src/borg/archiver.py
fantasya-pbem: That's right, but without / it would exclude also files like "/device". Not very likely, but...
I remove the / and add a comment with a warning.

fantasya-pbem:
Please wait until I re-checked the built HTML and verified examples again.

src/borg/archiver.py
fantasya-pbem: I added a comment to clarify this.

src/borg/archiver.py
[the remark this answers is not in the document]
fantasya-pbem: I now use double backticks everywhere except for borg commands, these appear underlined and look like links then.

fantasya-pbem:
Ok, please continue your review.

src/borg/archiver.py
```diff
@@ -2624,36 +2617,58 @@ def do_break_lock(self, args, repository):
...
 
-        Inclusion patterns are useful to include paths that are contained in an excluded
-        path. The first matching pattern is used so if an include pattern matches before
-        an exclude pattern, the file is backed up. If an exclude-norecurse pattern matches
-        a directory, it won't recurse into it and won't discover any potential matches for
-        include rules below that directory.
+        Exclude pattern prefix ``-``
+            Use the prefix ``-``, followed by a pattern, to define an exclusion
+            for ``--patterns`` / ``--patterns-from`` and ``--exclude`` /
+            ``--exclude-from``.
```
ThomasWaldmann, line 2641: i don't think we support prefixing a minus sign for `--exclude*`.

src/borg/archiver.py
```diff
@@ -2624,36 +2617,58 @@ def do_break_lock(self, args, repository):
...
 
-        .. note::
+        Exclude no-recurse pattern prefix ``!``
+            Use the prefix ``!``, followed by a pattern, to define an exclusion
+            that does not recurse into subdirectories. This saves time, but
+            prevents include patterns to match any files in subdirectories.
+
+        Include pattern prefix ``+``
+            Use the prefix ``+``, followed by a pattern, to define include files.
```
ThomasWaldmann, line 2649: "include files" have a special meaning for software developers.

so i'ld rather say "inclusions" or "files to include".

src/borg/archiver.py
fantasya-pbem: Yes, that makes no sense.

src/borg/archiver.py
fantasya-pbem: Rephrased!

ThomasWaldmann (approved):

ThomasWaldmann:
thanks!

PR #6747, opened 2022-06-05
[DOCS] #4776 - FAQ: How to swap backup media

### LINKED ISSUE #4776: Backups to alternating targets

Yes (and that's what this issue is partly about)

QUESTION / ISSUE

Debian 10  (current)

Current standard backup procedure is to rsync to a (number of USB) disc(s). Only one of them is connected at all times, the others are offline (and possibly offsite), rotating in a multi-generation scheme (daily/weekly/monthly).

As far as I could read from the docs this might be an issue, as when replacing the current backup disc with one from the safe that has not seen updates in a month.  AFAIU the local index (on the client) expects due to the cache that the blocks written in the last month to be on the disc. 

question 1 - potential data loss:
Does Borg automatically update the missing pieces on the older target?

question 2 - loss of snapshot:
If I have to restore a snapshot from 2 weeks ago (but not older) - that would be on the currently offline disc. Or does Borg maintain a local (client-side) copy of the snapshot indices?
Can Borg access snapshot from an old disc after it already worked with the new one?

### possible workaround

I guess one could work around that issue by using the same target mount point but different target names. Or even label the discs differently, so the mount points differ.

fantasya-pbem:
fixes #4776

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -45,6 +45,12 @@ repository is only modified from one place. Also keep in mind that
 Borg will keep an exclusive lock on the repository while creating
 or deleting archives, which may make *simultaneous* backups fail.
 
+Can I back up to multiple, swapped backup targets?
+--------------------------------------------------
+
+It is possible to swap your backup disks if each backup medium is assigned its
+own repository by creating a new one with :ref:`borg_init`.
+
 Can I copy or synchronize my repo to another location?
 ------------------------------------------------------
 


docs/faq.rst
```diff
@@ -45,6 +45,23 @@ repository is only modified from one place. Also keep in mind that
...
+--------------------------------------------------
+
+To protect my backups against data loss by malware or catastrophes like a fire I
+want to back up to weekly swapped-out USB disks. Then I place the media
+currently not in use offline and off-site to keep them safe. The swapping out
+can be as simple as switching between two disks regularly, or in more elaborate
+schemes like
+`Grandfather-Father-Son <https://en.wikipedia.org/wiki/Backup_rotation_scheme>`_.
+
+It is possible to swap your backup disks if each backup medium/set is assigned
```
ThomasWaldmann, line 58: what shall "backup set" mean here?

docs/faq.rst
```diff
@@ -45,6 +45,23 @@ repository is only modified from one place. Also keep in mind that
...
+its own repository by creating a new one with :ref:`borg_init`. Borg needs to
+keep a local cache containing checksums of all file chunks already stored in the
+repositories. These caches are stored in subdirectories of ``~/.cache/borg/``.
+If Borg detects that a repository has been modified since the local cache was
+updated, it will rebuild the cache which can be quite time consuming.
```
ThomasWaldmann, line 63: hmm, do we need to talk about cache / syncing in this context?

docs/faq.rst
```diff
@@ -45,6 +45,23 @@ repository is only modified from one place. Also keep in mind that
...
 
+Can I back up to multiple, swapped backup targets?
+--------------------------------------------------
+
+To protect my backups against data loss by malware or catastrophes like a fire I
+want to back up to weekly swapped-out USB disks. Then I place the media
+currently not in use offline and off-site to keep them safe. The swapping out
+can be as simple as switching between two disks regularly, or in more elaborate
+schemes like
+`Grandfather-Father-Son <https://en.wikipedia.org/wiki/Backup_rotation_scheme>`_.
```
ThomasWaldmann, line 56: somehow the reader expects an answer to the question in the headline here.

but this reads like a situation description of the user asking the question. i think this should be rephrased so it rather looks like an answer.

docs/faq.rst
fantasya-pbem: Caching seems to be the main reason, so yes.

docs/faq.rst
fantasya-pbem: Rephrased.

docs/faq.rst
fantasya-pbem: set removed

docs/faq.rst
fantasya-pbem: Of course we could simply answer this faq with "Create one repo per media.", without any explaination.

ThomasWaldmann: The latter was what I was thinking. :-)

That "if borg detects ..." is a general behaviour of borg and not related to swapping disks, but to updating the repo from another client.

docs/faq.rst
```diff
@@ -45,6 +45,18 @@ repository is only modified from one place. Also keep in mind that
...
 or deleting archives, which may make *simultaneous* backups fail.
 
+Can I back up to multiple, swapped backup targets?
+--------------------------------------------------
+
+To protect my backups against data loss, I want to back up to weekly swapped-out
+USB disks and place the media currently not in use offline and off-site to keep
+them safe. How can I apply this or other more elaborate
+`backup schemes <https://en.wikipedia.org/wiki/Backup_rotation_scheme>`_ with
+BorgBackup?
```
ThomasWaldmann, line 55: this is still the "i want ..." point of view.

but it should be rather an answer to the question in the headline.

so just rephrase to "it is possible to ..." and merge with next paragraph.

ThomasWaldmann:
Much better. No one likes to read long docs aka TL;DR. :-)

Issue #5960, opened 2021-08-18
What happens when the quota size is reached?


ddofborg:
I cannot find any information on what happens when the quota size on the repo is reached while backing up? Or when the free space runs out.

Can anyone shine some light on this (and maybe add something about this to the docs, `borg init` and `borg create` maybe)?

Also it would be nice if `borg create` would get some options to control what happens when quota is reached. Like prunning older backups with some --keep-XXX actions OR raising an error?

ThomasWaldmann:
IIRC it will just abort in the middle of the backup that exceeds the quota / free space.

There is no automated resolution of that "full" state and I also would rather dislike making that built-in into `borg create`, making the command's code and command line interface even more complex than it already is (especially not by mixing in functionality for which we already have a separate `borg prune` command).

You can just use `borg prune`with reasonable values to make your repo space not overflow usually.

Also you must use some kind of monitoring anyway to make sure your backups actually work and if monitoring shows failed backups you can decide how to resolve that (usually it is by adjusting prune parameters, excluding some unintended stuff from the backup source or increasing quota / available space for the repo).

ddofborg:
I do agree this would make things more complex, but I don't think current _abort_ method is the best user experience. Especially not with such an amazing tool as `borg`.

Having a way to example remove oldest backups first would be very useful. 

From what I see there is no way to do a `dry-run` and see if the backup would fit. That could be a less complex solution; basically do dry-run before actually uploading the backup. If it fits, then just upload the files? If it fails, remove the oldest backup (or _N_ oldest backups with total size of amount of needed space), then run the backup. If that fails, then abort. Or if backup is larger than total quota, abort too.

What do you think about this feature?

ThomasWaldmann:
Even "oldest" is not simply defined nor is it clear whether that is what the user wants.
See borg prune options and how they must be used, esp. if one backs up multiple data sets into same repo.

Yes, due to dedup, there is no simple way to determine whether something fits.
And it also doesn't make sense to make some "advanced dry-run" if that would take a similar effort as really doing a backup.

Also due to dedup, it is not easy to see how much space is freed by removing N backups for N>1.

A relatively simple thing we could do is have more distinct error codes / return codes (see other ticket about that) instead of just 0/1/2. Then some borg wrapper (e.g. your backup script) could detect the "out of quota / disk full" and invoke borg prune / borg delete and do whatever you want.

ddofborg:
[...]

PR #6755, opened 2022-06-06
[DOCS] #5960 - FAQ: Full quota / full disk

fantasya-pbem:


docs/faq.rst
```diff
@@ -1255,6 +1276,20 @@ It may be useful to set ``BORG_RELOCATED_REPO_ACCESS_IS_OK=yes`` to avoid the
 prompts when renaming multiple repositories or in a non-interactive context
 such as a script. See :doc:`deployment` for an example.
 
+My backup disk is full, what can I do?
+--------------------------------------
+
+First try to run :ref:`borg_check` to find out if there is any error in your
+backup repository. If something is wrong, first try to repair it before running
+any other Borg command.
```
ThomasWaldmann, line 1284: If the disk is full, guess borg check can't work.

ThomasWaldmann: So guess the first suggestion would be "free some disk space".

docs/faq.rst
```diff
@@ -1255,6 +1276,20 @@ It may be useful to set ``BORG_RELOCATED_REPO_ACCESS_IS_OK=yes`` to avoid the
...
+
+Second, if you ran a backup that has stopped due to a full disk, there may be a
+checkpoint archive that you can delete and get disk space back running
+:ref:`borg_delete` and :ref:`borg_compact` afterwards. Run :ref:`borg_list` with
+``--consider-checkpoints`` to show checkpoint archives.
```
ThomasWaldmann, line 1289: Even deleting objects in the repo needs disk space (because it appends "DEL chunkid" entries and a commit), so this only works if there is still some free space.

docs/faq.rst
```diff
@@ -1255,6 +1255,43 @@ It may be useful to set ``BORG_RELOCATED_REPO_ACCESS_IS_OK=yes`` to avoid the
...
+--------------------------------------
+
+Borg cannot work if you really have zero free space on the backup disk, so the
+first thing you must do is deleting some files to regain free disk space. See
+:ref:`about_free_space` for further details.
+
+Next you should run :ref:`borg_check` to find out if there is any error in your
+backup repository. If something is wrong, try to repair it first before running
+any other Borg command.
+
```
ThomasWaldmann, line 1289: Is there a special reason to do that in this situation?

Of course one can do that at any time to check the state of the repo, but I don't see a special need to do this after a disk full problem. borg should be able to roll back any uncommitted segments to return to a clean state.

fantasya-pbem: Yes, it should. I would definitely check my repo after disk full, before I use it again for backups.

fantasya-pbem: I have added "first thing you must do is deleting some files to regain free disk space" before, to make sure `borg check` works. I cannot test this in a realistic setting (like an encrypted repo on 8 or 16 GB USB stick) at the moment, but it is definitely interesting to see if our theoretical recipes work in reality...

fantasya-pbem: I did some disk-full tests yesterday. IMO there are two relevant conditions / cases:
1. disk already full when Borg starts - Borg almost immediately stops working, nothing bad happens
2. disk gets full when Borg saves chunks - a rollback happens and at least the new segment file is deleted, so that the disk is not full anymore

In case 2 my `borg check` didn't find any issues, so the rollback was successfull). Because of this I will change the "full disk faq" to note these two cases and that nothing special must be done regarding Borg.

PR #6762, opened 2022-06-10
[DOCS] Remove unnecessary/dangerous line continuation

Maltimore:
The line continuation is unnecessary for the last line of the command.
It could even be dangerous if a user modifies the script and puts something in the line directly below the command, not realizing that it will be mashed together with the previous command.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -189,7 +189,7 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
         /etc                            \
         /home                           \
         /root                           \
-        /var                            \
+        /var
 
     backup_exit=$?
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Well, I understand what you mean, but this was intentionally done like this to simplify copy & paste and modify (and avoid that the backslash is missing in some in-the-middle line).

So, the question is what's more likely to go wrong...

Maltimore:
Ah I see what the intention was :)
So I don't care too much either way and would be fine closing this PR.
I can just say from my own experience: reading through the docs and seeing the line continuation characters, I asked myself whether I had a misunderstanding about how line continuation works. I actually  made a `tmp.sh` file and experimented ;). So in some sense, I would say the way it is now is "wrong". It doesn't do any harm as is, but it's not how line continuation should be. So based on that I would go for the more "correct" way and delete the last line continuation character.

ThomasWaldmann:
OK, let's fix it.

PR #6763, opened 2022-06-11
borg2: there is no csize

ThomasWaldmann:
"what would we lose/win if we would just not track csize (the compressed size of a chunk)", see #2357.
- do not have csize in the item.chunks list entries
- do not have csize in the ChunkIndex entries
- do not track csize in Statistics
- but: my other PR https://github.com/borgbackup/borg/pull/6705 would make borg track csize in the repo index

What we lose (some stuff could be just done differently):
- archive info: compressed size and deduplicated-compressed size stats (the latter is based on csize and refcount)
- archive list: placeholders for csize / dcsize

What we win:
- get rid of some ugly / complicated code
- repo-scope recompression would be easily doable, because there is no csize to "fix" in the archives (and in the chunks index)
- no super expensive `fetch_missing_csize` and no `zero_csize_ids` any more for the (still experimental) `AdHocCache`
- 10% less memory needs for the chunks index and the item.chunks list (in memory and in the stored archive)

Also:
- I reimplemented the deduplicate size info, but now based on the uncompressed sizes.

### DIFF
--- a/src/borg/cache_sync/unpack.h
+++ b/src/borg/cache_sync/unpack.h
@@ -279,7 +279,6 @@ static inline int unpack_callback_array_end(unpack_user* u)
             /* refcount, size */
             cache_values[0] = _htole32(1);
             cache_values[1] = _htole32(u->current.size);
-            cache_values[2] = _htole32(0);  /* fake csize for now */
             if(!hashindex_set(u->chunks, u->current.key, cache_values)) {
                 SET_LAST_ERROR("hashindex_set failed");
                 return -1;
--- a/src/borg/cache_sync/unpack.h
+++ b/src/borg/cache_sync/unpack.h
@@ -86,14 +86,12 @@ typedef struct unpack_user {
 
         /*
          * processing ChunkListEntry tuple:
-         * expect_key, expect_size, expect_csize, expect_entry_end
+         * expect_key, expect_size, expect_entry_end
          */
         /* next thing must be the key (raw, l=32) */
         expect_key,
         /* next thing must be the size (int) */
         expect_size,
-        /* next thing must be the csize (int) */
-        expect_csize,
         /* next thing must be the end of the CLE (array_end) */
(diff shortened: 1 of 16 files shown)


src/borg/cache_sync/unpack.h
```diff
@@ -283,18 +276,16 @@ static inline int unpack_callback_array_end(unpack_user* u)
             refcount += 1;
             cache_entry[0] = _htole32(MIN(refcount, _MAX_VALUE));
         } else {
-            /* refcount, size, csize */
+            /* refcount, size */
             cache_values[0] = _htole32(1);
             cache_values[1] = _htole32(u->current.size);
-            cache_values[2] = _htole32(u->current.csize);
+            cache_values[2] = _htole32(0);  /* fake csize for now */
```
ThomasWaldmann, line 282: check this

ThomasWaldmann: fixed

sophie-h:
Pika Backup currently repots an 'additional backup space used' during backup. It reports the deduplicated size, I'm not even sure how the compression is counted here.

I don't care that much about compression here but the deduplicated size is was always a good indicator for me during a backup: If it's suddenly uploading a lot of data but you would have expected everything to be covered by deduplication, you might have missed excluding something.

However, this all does not matter during scheduled backups.

ThomasWaldmann:
@sophie-h I re-added the "deduplicated size", just changed its meaning:

It is now based on the uncompressed size (this is because we do not have the compressed size any more). So this can still be used as an indicator for sudden unexpected backup growth (because some exclude is missing or some crap unintentionally got backed up).

ThomasWaldmann:
What's currently missing is an indication of how good the compression worked.

But guess we could re-add this later in a different way. After #6689 is merged, we have the "csize" (or something closely related) in the repo index, so after adding some api method, we could compute some statistics based on that.

PR #6769, opened 2022-06-13
secure_erase: avoid collateral damage, fixes #6768

### LINKED ISSUE #6768: borg check causes corruption in hardlinked config file

Yes

Bug

borg 1.2.0

Tested on Ubuntu 20.04 and 21.10.

ext4

Doesn't matter.

```
mkdir /tmp/borgtest
cd /tmp/borgtest
borg init -e none test.repo
cp -al test.repo copy.repo
borg check -v test.repo
cat copy.repo/config  [binary data]
```

After making a hardlink copy of a repo, running check on the original causes corruption of the config file in the copy.  It happens with archives in the repo too, but I omitted this to make the script as short as possible.

Reproducible using the steps above.

None.

ThomasWaldmann:
if a hardlink copy of a repo was made and a new repo config
shall be saved, do NOT fill in random garbage before deleting
the previous repo config, because that would damage the hardlink
copy.

### DIFF
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -189,13 +189,23 @@ def scandir_inorder(*, path, fd=None):
     return sorted(os.scandir(arg), key=scandir_keyfunc)
 
 
-def secure_erase(path):
-    """Attempt to securely erase a file by writing random data over it before deleting it."""
+def secure_erase(path, *, avoid_collateral_damage):
+    """Attempt to securely erase a file by writing random data over it before deleting it.
+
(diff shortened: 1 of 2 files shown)


jdchristensen:
The logic looks correct to me.

jdchristensen:
By the way, why is borg writing out a new version of the config file?  Maybe it would be most safe to avoid doing that unless it really has changed...

ThomasWaldmann:
Guess the config writing is since I added support for partial, time-limited repo checks (it tracks up to where it came in the repo config).

Issue #6875, opened 2022-07-18
_version.py format change


ThomasWaldmann:
there seems to be a problem in debian or ubuntu because the file format changed slightly between setuptools_scm 6.4.2 (what I used) and 7.x (what they use), so a file changes unexpectedly.

guess i'll just release borg 1.2.2 using the 7.x version.

LocutusOfBorg:
TEMPLATES is the way to go?

ThomasWaldmann:
Fixed, see PRs above.

PR #6876, opened 2022-07-18
 Fix pyproject.toml to create a fixed _version.py file, compatible wi…

LocutusOfBorg:
…th both old and new setuptools_scm version (see: #6875)


ThomasWaldmann:
Thanks, needs another PR against 1.2-maint, can you do it?

LocutusOfBorg:
#6877 :)

PR #6880, opened 2022-07-20
new crypto does not need to call ._assert_id()

ThomasWaldmann:
See: https://github.com/borgbackup/borg/pull/6463#discussion_r925436156

### DIFF
--- a/docs/internals/security.rst
+++ b/docs/internals/security.rst
@@ -180,8 +180,6 @@ Decryption::
 
     decompressed = decompress(decrypted)
 
-    ASSERT( CONSTANT-TIME-COMPARISON( chunk-id, MAC(id_key, decompressed) ) )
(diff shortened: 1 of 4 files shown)


real-or-random:
The call to `assert_id` is also in the pseudocode in the docs, so it should be removed there, too.

Do you think `borg check --verify-data` should still recompute the MAC? I don't think so, just checking.

ThomasWaldmann:
Fixed docs, thanks for the hint.

`ArchiveChecker.verify_data` only calls `key.decrypt()` so it is already fixed by my change in the decrypt method, right?

ThomasWaldmann:
But we could give `decompress=False` to save some cpu time. Only works for new repos, but we do not support checking old repos anyway.

PR #6884, opened 2022-07-22
Fix typos and shellcheck warnings

a1346054:


ThomasWaldmann:
Thanks for the PR! Can you do the same changes to 1.2-maint also?

PR #6895, opened 2022-07-27
CI: test on macOS 12 without fuse / fuse tests

ThomasWaldmann:


ThomasWaldmann:
@m3nu guess from now on we need to have an eye on the fuse test results when testing locally.

i can do the M1 based testing and some intel macOS users could do the intel based testing.

as this is due to github kicking out old macOS, it will affect all branches, not just master.

PR #6911, opened 2022-07-29
docs(1.2): provide more details on object layout

horazont:
While writing my own out-of-band decoder, I had a hard time figuring out
how to unpack the manifest. From the description, I was only able to
read that the manifest is msgpack'd, but I had not been able to figure
out that it's also going through the same encryption+compression logic
as all other things do.

This should make it a little clearer and provide the necessary
information to understand how the compression works.

---

I also have the same change cherry-picked against master, but as master is now Borg 2.0, I'm not sure that it's correct. Any guidance appreciated (second PR with cherry-pick? PR against master only? ..?), my custom extractor is only tested against 1.2 :-).

### DIFF
--- a/docs/internals/data-structures.rst
+++ b/docs/internals/data-structures.rst
@@ -121,6 +121,14 @@ partial/uncommitted transaction.
 The size of individual segments is limited to 4 GiB, since the offset of entries
 within segments is stored in a 32-bit unsigned integer in the repository index.
 
+Objects
+~~~~~~~
+
+All objects (the manifest, archives, archive item streams chunks and file data
+chunks) are encrypted and/or compressed. See :ref:`data-encryption` for a
+graphic outlining the anatomy of an object in Borg. The `type` for compression
+is explained in :ref:`data-compression`.
+
 Index, hints and integrity
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -860,6 +868,8 @@ HashIndex is implemented in C and wrapped with Cython in a class-based interface
 The Cython wrapper checks every passed value against these reserved values and
 raises an AssertionError if they are used.
 
+.. _data-encryption:
+
 Encryption
 ----------
 
@@ -961,18 +971,23 @@ key file, wrapped using the standard ``textwrap`` module with a header.
 The header is a single line with a MAGIC string, a space and a hexadecimal
 representation of the repository id.
 
+.. _data-compression:
+
 Compression
 -----------
 
-Borg supports the following compression methods:
+Borg supports the following compression methods, each identified by two bytes:
 
-- none (no compression, pass through data 1:1)
-- lz4 (low compression, but super fast)
+- none (no compression, pass through data 1:1), identified by ``\x00\x00``
+- lz4 (low compression, but super fast), identified by ``\x01\x00``
 - zstd (level 1-22 offering a wide range: level 1 is lower compression and high
-  speed, level 22 is higher compression and lower speed) - since borg 1.1.4
+  speed, level 22 is higher compression and lower speed) - since borg 1.1.4,
+  identified by ``\x03\x00``
 - zlib (level 0-9, level 0 is no compression [but still adding zlib overhead],
-  level 1 is low, level 9 is high compression)
-- lzma (level 0-9, level 0 is low, level 9 is high compression).
+  level 1 is low, level 9 is high compression), identified by a zlib header
+  (``\x.8\x..``)
+- lzma (level 0-9, level 0 is low, level 9 is high compression), identified
+  by ``\x02\x00``.
 
 Speed:  none > lz4 > zlib > lzma, lz4 > zstd
 Compression: lzma > zlib > lz4 > none, zstd > lz4


docs/internals/data-structures.rst
```diff
@@ -961,18 +971,24 @@ key file, wrapped using the standard ``textwrap`` module with a header.
...
+
 Compression
 -----------
 
 Borg supports the following compression methods:
 
-- none (no compression, pass through data 1:1)
-- lz4 (low compression, but super fast)
+- none (no compression, pass through data 1:1), identified by ``\x00\x00``
+  (two NUL bytes)
```
ThomasWaldmann, line 982: maybe leave away the "(two NUL bytes)", it doesn't add useful information.

docs/internals/data-structures.rst
```diff
@@ -961,18 +971,24 @@ key file, wrapped using the standard ``textwrap`` module with a header.
 The header is a single line with a MAGIC string, a space and a hexadecimal
 representation of the repository id.
 
+.. _data-compression:
+
 Compression
 -----------
 
 Borg supports the following compression methods:
```
ThomasWaldmann, line 979: "(each identified by two type bytes)"

docs/internals/data-structures.rst
```diff
@@ -961,18 +971,24 @@ key file, wrapped using the standard ``textwrap`` module with a header.
...
+- lz4 (low compression, but super fast), idenified by ``\x01\x00``
 - zstd (level 1-22 offering a wide range: level 1 is lower compression and high
-  speed, level 22 is higher compression and lower speed) - since borg 1.1.4
+  speed, level 22 is higher compression and lower speed) - since borg 1.1.4,
+  identified by ``\x03\x00``
 - zlib (level 0-9, level 0 is no compression [but still adding zlib overhead],
-  level 1 is low, level 9 is high compression)
-- lzma (level 0-9, level 0 is low, level 9 is high compression).
+  level 1 is low, level 9 is high compression), identified by a zlib header
+  (``\x.8\x..``)
```
ThomasWaldmann, line 989: this (ZLIB_legacy) will go away in borg 2.0 and will be replaced by ZLIB that has a normal 2 known/constant bytes header as the others.

but we do not need to talk about it in 1.2.x docs.

docs/internals/data-structures.rst
ThomasWaldmann: this will change in borg 2.0:

one byte for compression type, one byte for compression level.

but no need to talk about it in 1.2 docs.

ThomasWaldmann (commented):
thanks for the PR!

docs/internals/data-structures.rst
horazont: My reason for having that was to make it fully unambiguous what is meant by `\x00`. In such specifications/documents, I do prefer to be more precise by stating things in two different ways than leaving something ambiguous.

But I can change it if you disagree despite the rationale :).

docs/internals/data-structures.rst
horazont: Ah, I didn't see the first comment in this thread at first. With that, the `(two NUL bytes)` is not needed, so I'll just change that up.

docs/internals/data-structures.rst
ThomasWaldmann: well, there is still the inconsistency that you explain it only for the none compression, but not for the others.

for the others, if you'ld stay using ascii control char abbreviations, would you like to use SOH as explanation for \x01 and STX for \x02?

i really don't think this is helpful. it "explains" something easy / familiar by using something more complicated / unfamiliar.

that's why i moved the info that there are 2 type bytes to the header (which applies to all) and suggested to drop the ascii char stuff from the none description.

horazont: Yup, fixed that in the most recent update :-).

(I thought I had fixed it in the update before already, but I somehow messed that up)

docs/internals/data-structures.rst
```diff
@@ -961,18 +971,23 @@ key file, wrapped using the standard ``textwrap`` module with a header.
...
 Compression
 -----------
 
-Borg supports the following compression methods:
+Borg supports the following compression methods, each identified by two bytes:
 
-- none (no compression, pass through data 1:1)
-- lz4 (low compression, but super fast)
+- none (no compression, pass through data 1:1), identified by ``\x00\x00``
+- lz4 (low compression, but super fast), idenified by ``\x01\x00``
```
ThomasWaldmann, line 982: typo: indentified

ThomasWaldmann (commented):
thanks for the update, spotted a typo, rest is fine.

docs/internals/data-structures.rst
horazont: Fixed thanks!

ThomasWaldmann:
I'm curious: would you like to tell for what do you write that OOB decoder?

horazont:
@ThomasWaldmann It is a learning exercise, as I like to understand how my backups work exactly to be able to recover stuff on my own if need be. I came from an rsnapshot-like setup to borg a few years back, so I lacked that kind of "safety" since then :). 

The current state is at [a repo of mine](https://github.com/horazont/borgrdr), if you like to take a look (it's not pretty. It's also very slow).

ThomasWaldmann:
Well, it's of course a lot more effort than with the rsync-based stuff to do it without the original tool because in borg's case the result is not just a copy of the original fs.

The good news is that it will get a bit less complicated / less legacy at some places with borg2.

ThomasWaldmann:
If you like, a PR against master would be great.

Just have a look at the compress code from master, the 2 bytes are used a bit differently.

PR #6915, opened 2022-07-30
docs(master): provide more details on object layout

horazont:
Same as #6911, with the following additions:

- Accounted for Zlib_legacy vs. ZLIB
- Updated the graphic which visualises all this to split the formerly two-byte type field into type and level.

I had to convert the drawing to odg because I don't have Visio. I can strip that commit off if you don't like that :-).


docs/internals/data-structures.rst
```diff
@@ -998,18 +1008,32 @@ key file, wrapped using the standard ``textwrap`` module with a header.
...
+  by 0x02.
+
+The type byte is followed by a byte indicating the compression level.
+
+.. note::
+
+    Borg 1.x (and attic) used to write zlib chunks without dedicated type
+    and level bytes. That means that in archives created using these tools,
+    such legacy zlib chunks may exist. These chunks are identified by the
+    lowest nibble of the first byte being 0x08.
```
ThomasWaldmann, line 1036: Correct, but we do not need this note in borg2 docs. Stuff from legacy repos will get transferred into new repos using borg2's  `transfer` command, which also will convert ZLIB_legacy to ZLIB on the fly.

Shorter docs are also a feature, thus I prefer not to describe stuff that does not apply any more (I also removed the AES-CTR related docs, AES-CTR is gone, except when borg transfer reads from an old repo).

Hmm, it seems I overlooked these encryption.* files - are they still references from the text?

horazont: Aha, I was not aware that `borg transfer` removes that. In that case, fair enough, I dropped that just now.

`encryption.png` is used in the `Legacy modes` section. I found that graphic quite useful to understand the layering of Borg objects (encryption -> compression -> data). It would be good to have an updated version for the AEAD mode.

ThomasWaldmann: I added a new diagram a while ago.

horazont: Ah, I hadn't seen that because I obviously looked at the stable docs for my experiments :-).

Would you like to get rid of the Legacy modes section altogether anyway? In that case I'll just drop the commit fixing the graphic.

ThomasWaldmann: Yes, guess we can get rid of the legacy modes section. This also means slight changes elsewhere (like "both modes" and "aead modes").

ThomasWaldmann: I'll merge it now and clean up afterwards.

PR #6941, opened 2022-08-05
massively increase archive metadata stream size limit, fixes #1473

### LINKED ISSUE #1473: per archive total item count limit
See #1452 for the basics.

While we have a short time fix for this, raising the limits, we also might need a long term fix, solving the limit problem completely.

ThomasWaldmann:
implemented by introducing one level of indirection, the limit is now very high, so it is not practically relevant any more.

we always use the indirection (storing the metadata stream chunk ids list not directly into the archive item, but into some repo objects referenced by the new ArchiveItem.item_ptrs list).

thus, the code behaves the same for all archive sizes.


ThomasWaldmann:
https://github.com/borgbackup/borg/pull/6941/files#diff-f6bfbe19a29b170971b1d015b7c07c31dba57c9412e536beb62cce2cfa8cc5c8L535

@m3nu @sophie-h that will go away, do you use it?

sophie-h:
I don't think so

Issue #6928, opened 2022-08-03
convert .vsd to .odg


ThomasWaldmann:
we have some diagrams made with visio in the docs, the should be transformed to an open format, like libreoffice draw.

ThomasWaldmann:
note: i tried loading the stuff into libreoffice on macOS, but it looked rather weird / incorrect.

mh4ckt3mh4ckt1c4s:
It's opening quite well for me with libreoffice on Linux. Do you want to recreate near-exact copies or can we be lenient regarding the look ?

ThomasWaldmann:
We can be lenient. It should look ok and transport the same information.

PR #6942, opened 2022-08-06
Update diagram to odg format (fix of #6928)

mh4ckt3mh4ckt1c4s:
I started to write the fix of #6928, only one doc out of 4, tell me if the format is fine for you. The only main change is the color of the arrows because the distinction between the black and dark violet was tough on the previous format.


ThomasWaldmann:
Good changes, thanks for working on this.

Discovered one issue: "Cut into chunks" is duplicate.

mh4ckt3mh4ckt1c4s:
Thanks for the report, the issue is fixed. I'm working on the others diagrams.

ThomasWaldmann:
@mh4ckt3mh4ckt1c4s if you work on the object-graph, can you add a "(simplified)" there, please?

reason: In v2 archives, the archive does not directly point to the item metadata stream chunks, but there is a layer of indirection in between, see #1473.

mh4ckt3mh4ckt1c4s:
Added a mention of "simplified schema" on top of it, tell me if it's okay for you.

mh4ckt3mh4ckt1c4s:
Hello @ThomasWaldmann, do you know if these things on the "compaction" schema mean anything specific ? Because they are a pain to reproduce on LibreOffice and if I can replace them by something else I'll do it, just tell me what suits you most. If it does mean something specify I'll figure out a way to reproduce it accurately.

image

ThomasWaldmann:
Well, it basically means that there can be more and only a part of it is shown. Could be also something like `...`.

ThomasWaldmann:
BTW, did you intentionally change the size of the PNGs?

mh4ckt3mh4ckt1c4s:
I just exported the odg to PNGs without paying intention to the PNGs size. If it is an issue I can just re import them with a specific size and make a commit to fix this.

ThomasWaldmann:
I did not try viewing them in the rendered docs (please do that to check the result of your work).

If the old size is a problem, please use some unproblematic size.

If the old size is ok, please use same or similar size, so the diff display in this PR works better.

mh4ckt3mh4ckt1c4s:
Hello, I finished to change all the docs to the odg format. Feel free to review and tell me if something is wrong so I can fix it before merging. I proofread myself but I can have forgotten some details. 

Regarding the size of the PNGs, I built the doc locally and checked that the sizes used now are correct.

ThomasWaldmann:
I still see very different png sizes compared to before (in pixels), see github diff view, try to use the "onion" slider.

In compaction, there used the be different shades of the colors to point out what's relevant to the descriptive text on the right side, but in the new pictures there is only full-intensity shade.

BTW, did you have to draw that all from scratch or did the import from .vsd work somehow?

mh4ckt3mh4ckt1c4s:
Oh yeah, I didn't know about this feature. I'm gonna take the old sizes and generates the new PNGs accordingly.

Indeed, I tried to use a transparency effect on LibreOffice but it wasn't working well. I'm gonna try and play with the colors to get a look similar to the old picture.

I tried to open the vsd files on LibreOffice, some were correct, some other completely bugged, so I figured out I'll lose less time to do everything from scratch than trying to salvage things that maybe wouldn't open well on other versions of LibreOffice or OpenOffice. So everything I created is from scratch, yes. Made me brush up my old libreoffice drawing skills 😄

mh4ckt3mh4ckt1c4s:
@ThomasWaldmann, I need your opinion. As it appears there is no full transparency option for objects in LibreOffice (or I'm not good enough to find it...). The transparency option exists only for backgrounds. So the easiest way to apply transparency effect on the `compaction` schema is to use this option, but the look is far from the previous one (especially because the font is not affected by transparency, menaning that nothing seems to change for the white rectangles), and the distinction between what matters or not is not really visible.

Here is an example of that : 

image

Another way I found is to create empty rectangles on top on what I want to make transparent. It gives a way better look, but it's hacky and on edit, one would have to move these transparent layers away in order to edit the schema, then put the transparent layers back. However, if we suppose that this schema won't be edited much during the following years, that might do. 

Here is an example of this, with one transparent rectangle selected so you can see where it is : 

image
 
What do you think ?

ThomasWaldmann:
Well, the lower one looks better, but the method used for creating this sounds like a pain.

So, using the upper one's approach, could one:
- make the colours of the unimportant parts look even more whitish / dim?
- use a bold font for the important parts to make them look more different from the non-bold in the unimportant parts

?

mh4ckt3mh4ckt1c4s:
It's not too much pain adding it, but it will make further edits more complicated...

I can dim the colors even more, see below. 

Some of the important things were in bold on the old doc but not all, but I can switch everything important in bold, but it's not really visible : 

image

I can also do bold + increase font size by 1 to make it more obvious : 

image

ThomasWaldmann:
The lower one looks quite good. Thanks for all your efforts!

mh4ckt3mh4ckt1c4s:
Okay, put the transparency in place and also fixed the size of the images. I think everything is good, just need to review a last time !

mh4ckt3mh4ckt1c4s:
Checked the newly generated docs, images are displaying correctly... But maybe we went too far on transparency 😆 

image

ThomasWaldmann:
Yeah, hard to see the colours.

mh4ckt3mh4ckt1c4s:
Fixed it.

image

ThomasWaldmann:
Thanks a lot! Would you like to claim the bounty?

If so I would first increase it a bit considering this was more work than expected.

mh4ckt3mh4ckt1c4s:
No problem, happy to help with this project ! Yes, I would like to claim the bounty. I will fill the request on bountysource.

PR #6944, opened 2022-08-06
ctrl-c must not kill important subprocesses, fixes #6912

### LINKED ISSUE #6912: ctrl-c does not work as intended when using a remote repo
## Feature
borg has "ctrl-c/SIGINT support" for some operations since a while.

the idea is that if a user presses ctrl-c **once**, borg will intercept that, write some sort of "checkpoint" (so it does not have to start all over from the beginning, but saves the progress it has made) and then terminate.

## The issue
While working on #6887 i noticed that this seems not to work with a remote (`ssh://`) repo:
- ctrl-c key combo gets detected by the terminal and it sends a SIGINT to all processes in the process group (borg and ssh client invoked by borg)
- borg receives the SIGINT, catches it and likely would want to do a checkpoint...
- ssh also receives the SIGINT and terminates, breaking the pipe to borg, breaking the connection to the repo
- borg notices that (broken pipe / connection closed by remote) and terminates due to the connection being closed
- borg does not get to doing the checkpoint

## Workaround
When using `kill -INT <BORG_PID>` to send a SIGINT to the borg process **only** (and not to ssh), it works as intended.

ThomasWaldmann:


ThomasWaldmann:
Did a quick test running a `borg create` against a ssh: remote repo: it behaves now as expected:
- no "broken pipe"
- successfully writes checkpoint

ThomasWaldmann:
There are some other places with subprocesses:
- `--content-from-command`
- `--paths-from-command`
- `import-tar / export-tar` (for the (de)compression filter process)

PR #6967, opened 2022-08-13
_version.py: remove trailing blank, add LF at EOF

ThomasWaldmann:
avoid triggering the pep8 checker.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -264,7 +264,7 @@ setup(
     name='borgbackup',
     use_scm_version={
         'write_to': 'src/borg/_version.py',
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@LocutusOfBorg can you check if that still work for what you've fixed it originally for?

ThomasWaldmann:
we'll just try. as the template now is the same as i've seen in the older release, i hope it doesn't cause any regressions.

LocutusOfBorg:
I guess it won't make much difference! Thanks for the fixup I'll test and report on next release!

PR #6979, opened 2022-08-18
compression header: include uncompressed size

### LINKED ISSUE #6702: prefix compressed chunks with decompressed size?
we don't have that yet, but guess it would be good for:
- decompression buffer allocation
- not only knowing the csize, but also the size without actually reading and decompressing it

as this is encrypted, we do not disclose anything by this.

borg can then find out:
- csize: by looking at the PUT header, by looking at the length of the read data, via the new repo index
- size: by reading a few bytes from the chunk, decrypting them reveals: compression type, level, size

borg 1.1/1.2 needs to read all archive metadata streams to find out all chunk sizes.
the new way would "only" need to read a few bytes from each chunk exactly once.

ThomasWaldmann:
so it is now: ID, level, length

refactored the compression classes a bit:
- ID is now an int
- no super() call to add/remove the header
- new make_hdr, parse_hdr, get_payload calls
- for the header, use Struct.pack/unpack
- simplify lz4 decompression: we now *know* osize

fixes #6702


ThomasWaldmann:
we can't do a partial AEAD decryption, thus this change can't achieve what we initially wanted (to just read some few bytes, decrypt) to get the chunk (decompressed/plaintext) size.

Issue #2209, opened 2017-02-25
Add instrumentation to borg create?


textshell:
Just a thought from an irc discussion about finding backup performance problems. If we had some option that outputs more detailed statistics about potentially performance relevant details it might be a lot easier to diagnose performance problems.

Ideas:
* number of files cache hits
* number of files cache indicated changed files
* number of files not in files cache at all
* time spent in chunking
* time spent in id hashing
* time spent in crypto (encryption + integrity hash)
* bytes sent to remote
* bytes read from remote
* can we do an reasonable approximation of how long borg was blocked by remote communications?

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/42441165-add-a-profiling-summary-option-for-borg-create)

enkore:
> bytes send to remote
bytes read from remote

I actually wrote that a few days ago. #2210

=> Separate borg.output. logger + cmdline option for these?

textshell:
Magic ;)

I was thinking like this would be an new option similar to --summary but with performance focused information. Still more like a report than just some logging lines. Just using a logger might be easier though. I expect some of the data collection to potentially have a performance impact though. (or maybe not? e.g. querying the current time on linux is just a memory read)

enkore:
Direct tracing (via `sys.setprofile`) is really quite expensive, adding ~two extra function calls to encrypt+decrypt+id_hash has maybe not such a big impact, but perhaps it would better to implement it directly (not via @decorator magic or extra methods).

textshell:
I didn‘t consider `sys.setprofile` a valid path for lightweight stats. I agree that such performance code should be kept lean and not too magic. But using an extra method would ease implementation:
Something like this would be fairly maintainable and easy to check for correct usage.
```
    performance_stats.set_section("chunking")
    ...
    performance_stats.set_section("id_hash")
```

ThomasWaldmann:
another idea might be to add some benchmark, so we can see perf independently of user data:
- write some known content / size files to src folder, like:

  - big, all zero
  - big, all random
  - lots of small files, all zero
  - lots of small files, all random
- create backups of these into dst repo
- don't use compression (we know that lz4 is fast and lzma is slow)
- test with / without encryption depending on how dst repo was initialized

textshell:
More synthetic tests are covered by #1788 i think.

enkore:
```
Resource Usage (relative): 5% CPU, 0% input, 94% network
[...]

PR #6991, opened 2022-08-24
Adding performance statistics to borg create

francoo98:
Issue #2209
Up until now, I have worked on:

- number of files cache hits (b9d713b)
- number of files cache indicated changed files (b9d713b)
- number of files not in files cache at all (b9d713b)
- time spent in chunking (3e7326e)
- time spent in id hashing (c17f6f1)

The changes are pretty short and simple, I don't think a description is necessary.
Let me know what you think and if there is anything that needs to be changed.


src/borg/archive.py
```diff
@@ -1289,6 +1289,7 @@ def __init__(
 
         self.hlm = HardLinkManager(id_type=tuple, info_type=(list, type(None)))  # (dev, ino) -> chunks or None
         self.stats = Statistics(output_json=log_json, iec=iec)  # threading: done by cache (including progress)
+        self.stats.files_stats = {"A": 0, "M": 0, "U": 0}
```
ThomasWaldmann, line 1292: such stuff belongs into `Statistics.__init__` I guess.

src/borg/archive.py
```diff
@@ -1401,7 +1402,9 @@ def process_file(self, *, path, parent_fd, name, st, cache, flags=flags_normal):
                 else:  # normal case, no "2nd+" hardlink
                     if not is_special_file:
                         hashed_path = safe_encode(os.path.join(self.cwd, path))
+                        started_hashing = time.time()
                         path_hash = self.key.id_hash(hashed_path)
+                        self.stats.hashing_time += time.time() - started_hashing
```
ThomasWaldmann, line 1407: guess you need the monotonic timer here.

src/borg/archive.py
```diff
@@ -1424,19 +1427,22 @@ def process_file(self, *, path, parent_fd, name, st, cache, flags=flags_normal):
...
                         with backup_io("read"):
+                            started_chunking = time.time()
                             self.process_file_chunks(
                                 item,
                                 cache,
                                 self.stats,
                                 self.show_progress,
                                 backup_io_iter(self.chunker.chunkify(None, fd)),
                             )
+                            self.stats.chunking_time += time.time() - started_chunking
```
ThomasWaldmann, line 1445: monotonic timer

src/borg/archiver/create_cmd.py
```diff
@@ -196,6 +196,9 @@ def create_inner(archive, cache, fso):
                             json_print(basic_json_data(manifest, cache=cache, extra={"archive": archive}))
                         else:
                             log_multi(str(archive), str(archive.stats), logger=logging.getLogger("borg.output.stats"))
+                        print(f"Files status: {fso.stats.files_stats}")
```
ThomasWaldmann, line 199: looks like this would render a dict?

src/borg/archiver/create_cmd.py
```diff
@@ -266,6 +269,8 @@ def create_inner(archive, cache, fso):
                     iec=args.iec,
                     file_status_printer=self.print_file_status,
                 )
+                fso.stats.chunking_time = 0
+                fso.stats.hashing_time = 0
```
ThomasWaldmann, line 273: also move this into __init__?

ThomasWaldmann (changes requested):
thanks for the PR, I added some feedback.

ThomasWaldmann:
It would be good to have a test in `borg.testsuite.archiver` that checks if the new code works correctly.
See the other tests you already see there about how this could work.

src/borg/archiver/create_cmd.py
ThomasWaldmann: see the few lines of code right above what you added: the output needs to be either json or human readable text.

francoo98:
Should the time be printed in seconds? Are decimals needed?

> It would be good to have a test in `borg.testsuite.archiver` that checks if the new code works correctly. See the other tests you already see there about how this could work.

I am thinking about creating a test that would create a few archives and then check that the `files_stats` variable is correct (e.g: if one file was added "A" should be 2), and check that the printing is correct.
I don't know how to test time measures.

ThomasWaldmann:
IIRC `borg create --stats` already outputs some timing stats, do the new stats in a similar way concerning units and precision.

Testings the timing stats of course can not be done precisely, but maybe you could check it is `> 0.0`?

ThomasWaldmann:
BTW, you are working in your master branch.

Usually you should first create a feature branch for the feature you are working on, that works much better in the end.

You can leave it like it is for this time, but remember it next time.

src/borg/archiver/create_cmd.py
```diff
@@ -192,6 +194,8 @@ def create_inner(archive, cache, fso):
                     archive.save(comment=args.comment, timestamp=args.timestamp, stats=archive.stats)
                     args.stats |= args.json
                     if args.stats:
+                        archive.stats.chunking_time = format_timedelta(timedelta(seconds=archive.stats.chunking_time))
+                        archive.stats.hashing_time = format_timedelta(timedelta(seconds=archive.stats.hashing_time))
                         if args.json:
                             json_print(basic_json_data(manifest, cache=cache, extra={"archive": archive}))
```
ThomasWaldmann, line 200: maybe looks how basic_json_data() works and what str(archive.stats) invokes.

for json, check if and how other values are formatted.

for str, check how other values are formatted and maybe do it there in the same way.

src/borg/archive.py
```diff
@@ -80,13 +83,19 @@ def __add__(self, other):
         stats.osize_parts = self.osize_parts + other.osize_parts
         stats.usize_parts = self.usize_parts + other.usize_parts
         stats.nfiles_parts = self.nfiles_parts + other.nfiles_parts
+        stats.files_stats = other.files_stats
+        stats.chunking_time = other.chunking_time
+        stats.hashing_time = other.hashing_time
```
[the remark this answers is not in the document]
francoo98: I realized it, yes. They are needed so `archive.stats += fso.stats` (create_cmd.py line 188) moves those variables from object to another.
If this is done, it is not needed to modify the calls to `json_print` and `log_multi`.

src/borg/archive.py
```diff
@@ -80,13 +83,19 @@ def __add__(self, other):
...
         return stats
 
     def __str__(self):
         return """\
 Number of files: {stats.nfiles}
 Original size: {stats.osize_fmt}
 Deduplicated size: {stats.usize_fmt}
+Files status: {stats.files_stats}
```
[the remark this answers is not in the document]
francoo98: Can I use f-strings for this? I am having problems printing the dictionary because `format` can't work with them, so I think the only solution is adding an argument to the call like this `format(stats=self, **self.files_stats)`.
Furthermore, I neither can't call functions so the formatting of `hashing_time` and `chunking_time` would have to be done before the return clause.

Without f-strings it would look something like this:
``` python
def __str__(self):
        hashing_time = format_timedelta(timedelta(seconds=self.hashing_time))
        chunking_time = format_timedelta(timedelta(seconds=self.chunking_time))
        return """\
Number of files: {stats.nfiles}
Original size: {stats.osize_fmt}
Deduplicated size: {stats.usize_fmt}
Files status: {{'A': {A}, 'U': {U}, 'M': {M}}}
Time spent in hashing: {hashing_time}
Time spent in chunking: {chunking_time}
""".format(
            stats=self,
            hashing_time=hashing_time,
            chunking_time=chunking_time,
            **self.files_stats
        )
```
This is the `__str__` method of `Statistics` by the way.

src/borg/archive.py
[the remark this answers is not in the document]
ThomasWaldmann: What I meant is that you are implementing an **add** operation here, so just assigning the stuff is not the correct operation. Even if the only use case was adding it to zero base values, you still need to do the addition.

src/borg/archive.py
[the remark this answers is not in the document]
ThomasWaldmann: Yes, minimum python requirement since borg 1.2 (and we won't backport this to 1.1.x) is python 3.8, so f-strings are available.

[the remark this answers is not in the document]
ThomasWaldmann: But a smaller change would be to just pick the 3 individual value from the dict and give them to `.format`.

I think a smaller change is better here. Refactoring `.format` to f-strings is not in scope of this PR (but could be done separately later).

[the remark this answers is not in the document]
francoo98: So you want it to look something like this?
```python
def __str__(self):
        hashing_time = format_timedelta(timedelta(seconds=self.hashing_time))
        chunking_time = format_timedelta(timedelta(seconds=self.chunking_time))
        return """\
Number of files: {stats.nfiles}
Original size: {stats.osize_fmt}
Deduplicated size: {stats.usize_fmt}
Files status: {{'A': {added_files}, 'U': {unchanged_files}, 'M': {modified_files}}}
Time spent in hashing: {hashing_time}
Time spent in chunking: {chunking_time}
""".format(
            stats=self,
            hashing_time=hashing_time,
            chunking_time=chunking_time,
            added_files=self.files_stats["A"]
            unchanged_files=self.files_stats["U"]
            modified_files=self.files_stats["M"]
        )
```

src/borg/archive.py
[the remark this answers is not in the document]
francoo98: Ok, I understand what you mean. There is nothing to add here, so I'll have to move those lines somewhere else.
They could be moved to `create_cmd.py` below line 188
```python
    archive.stats += fso.stats
    archive.stats.files_stats = fso.stats.files_stats
    archive.stats.chunking_time = fso.stats.chunking_time
    archive.stats.hashing_time = fso.stats.hashing_time
```
And then it will go on and print the statistics.

src/borg/archive.py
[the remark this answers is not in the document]
ThomasWaldmann: Guess this would be simpler:
```
Added files:  {added_files}
Modified files: {modified_files}
Unchanged files: {unchanged_files}
```

src/borg/archive.py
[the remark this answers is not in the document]
ThomasWaldmann: If the values are initializes to 0 or 0.0 (as they should be), adding them should be no problem?

src/borg/archive.py
```diff
@@ -83,21 +83,33 @@ def __add__(self, other):
...
         stats.usize_parts = self.usize_parts + other.usize_parts
         stats.nfiles_parts = self.nfiles_parts + other.nfiles_parts
-        stats.files_stats = other.files_stats
-        stats.chunking_time = other.chunking_time
-        stats.hashing_time = other.hashing_time
+        stats.chunking_time = self.chunking_time + other.chunking_time
+        stats.hashing_time = self.hashing_time + other.hashing_time
+        stats.files_stats["A"] = self.files_stats["A"] + other.files_stats["A"]
+        stats.files_stats["U"] = self.files_stats["U"] + other.files_stats["U"]
+        stats.files_stats["M"] = self.files_stats["M"] + other.files_stats["M"]
```
ThomasWaldmann, line 90: btw, i just noticed that in case there is no "added file", such code will crash, because then there will be no "A" key in the dict.

maybe it is simpler to not have that dict and just some `stats.*` attributes likes files_added, files_modified, files_unchanged.

francoo98: I think it won't because the dict is initialized with the three keys in `Statistics.__init__`.

src/borg/archive.py
```diff
@@ -1424,19 +1450,22 @@ def process_file(self, *, path, parent_fd, name, st, cache, flags=flags_normal):
                     else:
                         status = "M" if known else "A"  # regular file, modified or added
                     self.print_file_status(status, path)
+                    self.stats.files_stats[status] += 1
```
ThomasWaldmann, line 1453: Just realized we have some more status keys:
- misc. for non-regular files, like d for dirs, s for symlinks, b/c for block/char devices, ...
- E for files borg could not read

I guess we should extend stats to some of them and ignore the less interesting ones:
- E we should definitely track
- maybe check the borg code and the docs if you find other interesting ones.

src/borg/testsuite/archive.py
```diff
@@ -60,6 +60,11 @@ def test_stats_format(stats):
 Number of files: 1
 Original size: 20 B
 Deduplicated size: 20 B
+Time spent in hashing: 0.00 seconds
+Time spent in chunking: 0.00 seconds
+Added files: 0
+Unchanged files: 0
+Modified files: 0
```
ThomasWaldmann, line 67: that's good to test the initialisation, but guess we want to also test if it increments the counters correctly.

francoo98 (commented):
I'm not sure what I'm doing here tbh, I've never written a test before 😅

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,23 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+        """Test file status counters in the stats of `borg create --stats`"""
+        # Test case set up: create a repository
+        self.cmd(f"--repo={self.repository_location}", "rcreate", KF_ENCRYPTION)
+        # Create an archive
+        result = self.cmd(f"--repo={self.repository_location}", "create", "--stats", "test_archive", "./")
+        result = result.split("\n")
+        result = [l.split(":") for l in result]
+        result = [l for l in result if l[0] in ("Added files", "Unchanged files", "Modified files")]
+        # The line above left the value to the right of the `:` with a leading space.
+        # The line below removes the space and converts the value to int.
```
ThomasWaldmann, line 3778: `int(" 123")` just works.

ThomasWaldmann: Something like this should also work:
```
result = result.split("\n")
result = [line.split(":", 1) for line in result]
result = {
    key: int(value)
    for key, value in result
    if key in ("Added files", "Unchanged files", "Modified files")}
```

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,23 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+        result = [[l[0], int(l[1][1:])] for l in result]
+        result = {l[0]:l[1] for l in result}
+        assert result["Added files"] == 3
+        assert result["Unchanged files"] == 0
+        assert result["Modified files"] == 0
+
```
ThomasWaldmann, line 3784: ok, this looks quite good if that test already works.

after this initial backup, you could just add 1 more files and modifiy one existing file and check if the values behave accordingly.

src/borg/testsuite/archive.py
```diff
@@ -96,6 +96,11 @@ def test_stats_progress_json(stats):
     assert "original_size" not in result
     assert "nfiles" not in result
 
+def test_hashing_time(stats):
+    assert stats.hashing_time >= 0.0
+
+def test_chunking_time(stats):
+    assert stats.chunking_time >= 0.0
```
ThomasWaldmann, line 103: guess this is not really testing something useful yet.

the `stats` you have as argument here comes from the fixture in line 19+.

ThomasWaldmann:
@francoo98 writing tests seems a bit tedious at first, esp. if you are new to it, but long term it can save countless of hours of manual testing (and/or countless of bugs). Check "TDD / test driven development" and "pytest".

`pytest` is a quite powerful testing framework, makes writing tests easier than other ones (after you got used to it).

src/borg/testsuite/archive.py
francoo98: Should I remove these? https://github.com/borgbackup/borg/pull/6991/commits/3ef8052ff8486d09ac8b4cac38399d161b04f65c Replaces `test_hashing_time`, I'm gonna add one for chunking time too.

ThomasWaldmann: yes, remove

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,64 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+            return hashing_time
+
+        # Test case set up: create a repository and a file
+        self.cmd(f"--repo={self.repository_location}", "rcreate", KF_ENCRYPTION)
+        self.create_regular_file("testfile1", contents=b"test1")
+        # Archive
+        result = self.cmd(f"--repo={self.repository_location}", "create", "--stats", "test_archive", self.input_path)
+        hashing_time = extract_hashing_time(result)
+
+        assert hashing_time >= 0.0
```
ThomasWaldmann, line 3824: this is a quite weak test considering you initialise the value to 0.0.

of course we can not know how long it takes, so guess the best we can do is check `> 0.0`.

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,64 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+        assert result["Added files"] == 1
+        assert result["Unchanged files"] == 1
+        assert result["Modified files"] == 0
+
+    def test_hashing_time(self):
+        """Tests hashing_time is positive or zero"""
+
+        def extract_hashing_time(borg_create_output: str) -> float:
+            borg_create_output = borg_create_output.split("\n")
+            borg_create_output.pop()  # The last line needs to be removed because its an empty string
```
ThomasWaldmann, line 3810: the empty line comes from trailing whitespace, so you could get rid of that early:
```
output = output.strip().splitlines()
```

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,64 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+            borg_create_output = [line.split(":", 1) for line in borg_create_output]
+            hashing_time = [line for line in borg_create_output if line[0] == "Time spent in hashing"].pop()
+            hashing_time = hashing_time[1]
+            hashing_time = float(hashing_time[1][0 : len(hashing_time) - 7])
```
ThomasWaldmann, line 3814: that looks strange...

francoo98: Borg create outputs time like `<two decimal float> seconds`. With the last line I'm removing `seconds` from the string.

ThomasWaldmann:
```
>>> float("12.34 seconds".rstrip(" seconds"))
12.34
```

francoo98 (commented):
I think there is a bug in this test. The set up in the last scenario is 1 modified file (`testfile2`) and 1 unmodified (`testfile`), so `result["Added files"]` should store a 1, but it stores a 2 for some reason.

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,65 @@ def test_change_location_does_not_change_algorithm_argon2(self):
             key = msgpack.unpackb(a2b_base64(repository.load_key()))
             assert key["algorithm"] == "argon2 chacha20-poly1305"
 
+    def test_file_status_counters(self):
+        """Test file status counters in the stats of `borg create --stats`"""
+
+        def to_dict(borg_create_output: str) -> dict:
+            borg_create_output = borg_create_output.split("\n")
+            borg_create_output.pop()  # The last line needs to be removed because its an empty string
```
ThomasWaldmann, line 3773: same strip / splitlines as below.

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,65 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+        assert result["Unchanged files"] == 0
+        assert result["Modified files"] == 0
+        # Archive a dir with 1 unmodified file and 1 modified
+        with open(self.input_path + "/testfile2", "a") as file:
+            file.write("data")
+        result = self.cmd(f"--repo={self.repository_location}", "create", "--stats", "test_archive3", self.input_path)
+        result = to_dict(result)
+        assert result["Added files"] == 1
+        assert result["Unchanged files"] == 1
+        assert result["Modified files"] == 0
```
ThomasWaldmann, line 3805: the expected values here should be 0 / 1 / 1.

src/borg/testsuite/archiver.py
```diff
@@ -3765,6 +3765,65 @@ def test_change_location_does_not_change_algorithm_argon2(self):
...
+
+    def test_hashing_time(self):
+        """Tests hashing_time is positive or zero"""
```
ThomasWaldmann, line 3808: maybe remove the comment.

src/borg/testsuite/archiver.py
francoo98: I think it should be 1/1/0. Added should be one because of https://borgbackup.readthedocs.io/en/stable/faq.html#i-am-seeing-a-added-status-for-an-unchanged-file, in this case the added file and the modified file are the same file.

ThomasWaldmann:
Looks like there is some conflicts with master branch, please resolve.

src/borg/archive.py
francoo98: I'm working on this right now, but I don't know how to purposefully damage a file

ThomasWaldmann: Yeah, testing this is a bit tricky. But we already have tests that **mock** erroneous behaviour. See e.g. the xattr related tests.

Guess we could also just initialize the "E" counter to 0 and (considering it is the same code that also increments the "M" counter) just assume that it will also work.

ThomasWaldmann:
the linter is still unhappy, see github checks.

you can run `black .`and `tox -e flake8` locally to detect and fix such stuff early.

src/borg/archive.py
francoo98: I took a look at the docs and I think maybe 'x' is important. Should I add it?

ThomasWaldmann: maybe let's first focus on getting the other stuff working and the tests succeed.

`x` can be a bit difficult, depending on whether `-` or `!` is used to exclude, the counts there would be very different.

src/borg/archive.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, output_json=False, iec=False):
         self.osize = self.usize = self.nfiles = 0
         self.osize_parts = self.usize_parts = self.nfiles_parts = 0
         self.last_progress = 0  # timestamp when last progress was shown
+        self.files_stats = {"A": 0, "U": 0, "M": 0, "E": 0}
```
ThomasWaldmann, line 63: Hmm, maybe using a `defaultdict(int)` would be better here:

https://docs.python.org/3.9/library/collections.html#collections.defaultdict

ThomasWaldmann (commented):
the mypy checker is also unhappy.

src/borg/archive.py
```diff
@@ -81,15 +84,36 @@ def __add__(self, other):
         stats.osize_parts = self.osize_parts + other.osize_parts
         stats.usize_parts = self.usize_parts + other.usize_parts
         stats.nfiles_parts = self.nfiles_parts + other.nfiles_parts
+        stats.chunking_time = self.chunking_time + other.chunking_time
+        stats.hashing_time = self.hashing_time + other.hashing_time
+        stats.files_stats["A"] = self.files_stats["A"] + other.files_stats["A"]
+        stats.files_stats["U"] = self.files_stats["U"] + other.files_stats["U"]
+        stats.files_stats["M"] = self.files_stats["M"] + other.files_stats["M"]
+        stats.files_stats["E"] = self.files_stats["E"] + other.files_stats["E"]
```
ThomasWaldmann, line 92: considering the code now automatically initialises values to 0 for arbitrary new keys due to the defaultdict, you could use a loop here to iterate over all keys.

guess "all" would need to be the union of the key sets of self.files_stats and other.files_stats.

src/borg/archiver/create_cmd.py
```diff
@@ -187,6 +187,11 @@ def create_inner(archive, cache, fso):
                 if args.progress:
                     archive.stats.show_progress(final=True)
                 archive.stats += fso.stats
+                try:
+                    archive.stats.rx_bytes = repository.rx_bytes
+                    archive.stats.tx_bytes = repository.tx_bytes
+                except AttributeError:
+                    pass  # We are working with a local repo
```
ThomasWaldmann, line 194: yeah, would work like that.

you could also use `getattr(repository, 'rx_bytes', 0)`.

francoo98:
I have some issues that I don't know how to solve.

1.  Where should I put the test for the `tx_bytes` and `rx_bytes`?
2. `test_hashing_time` fails because the execution is too fast, how many files should I create in the test to make it slower?
3. The original issue (#2209) tells to add time spent in encryption, but I don't if this means encryption of repository contents (i.e: `-e` set to something different than `none`) or encryption of network traffic (because of a `RemoteRepository`)

ThomasWaldmann:
Ehrm, that code is wrong. I meant `.removesuffix()`, not `.rstrip()`:
```
hashing_time = float(hashing_time.rstrip(" seconds"))
```

Please fix all similar places in your PR.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -731,6 +771,44 @@ def test_common_options(self):
...
+            borg_create_output = [line.split(":", 1) for line in borg_create_output]
+            hashing_time = [line for line in borg_create_output if line[0] == "Time spent in hashing"].pop()
+            hashing_time = hashing_time[1]
+            hashing_time = float(hashing_time.rstrip(" seconds"))
+            return hashing_time
+
+        # Test case set up: create a repository and a 500 files
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        for i in range(0, 500):
+            self.create_regular_file(f"testfile{i}")
```
ThomasWaldmann, line 786: how long will that file be?

francoo98: Zero bytes long. Still, I'm not sure what to do here. In my computer if a create a single file like this: `self.create_regular_file("testfile", size=50000000)` it fails the test.

ThomasWaldmann: if the file is 0 bytes long, there is nothing to hash, so you can't expect the hashing time to be more than 0.0 seconds.

find out, why the test fails. if an assertion fails, you can usually see the output of any `print()` you did before.

after debugging the root cause, please use less than 50MB for the test. the tests shall run rather quickly, so do not produce/use more data than needed.

francoo98: I just noticed that the function `id_hash` method is given a file path as its `data` argument, not the contents of the file.
That is why it fails with a single file 50MB long and it passes with 1000 files

ThomasWaldmann: id_hash is used for multiple purposes. for the main purpose, it gets file content chunks.

but how about just not giving it an empty file?

ThomasWaldmann: btw, due to deduplication, avoid giving it a lot of duplicate data (or it will do less than you might think).

you can avoid dedup by giving it random data.

francoo98: Ok, I understand where I'm wrong now. Up to this point, the only hashing I measure in my code is in `archive.py` line 1441. This is what I mentioned in my previous comment about only file path being hashed.
So, I am missing the measurement of the hashing time done to process file chunks right? This is done in the method `ChunksProcessor.process_fille_chunks`

ThomasWaldmann: yes. for that, the file does need content (> 0 bytes length). Also, if you must produce more than a single chunk, avoid using all-zero bytes, but use random data to make borg really compute hashes (and not do some clever optimisation).

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -780,9 +781,19 @@ def extract_hashing_time(borg_create_output):
...
+            filename = os.path.join(self.input_path, "test_file")
+            with open(filename, "wb") as file:
+                for byte in data:
+                    file.write(bytes(byte))
+
         # Test case set up: create a repository and a file
         self.cmd(f"--repo={self.repository_location}", "rcreate", "--encryption=none")
-        self.create_regular_file("testfile", size=6000000)
+        create_test_file()
+        # self.create_regular_file("testfile", size=3000000, contents=create_random_bytes())
```
ThomasWaldmann, line 796: please do not reinvent the wheel.

you can call `self.create_regular_file("testfile", content=...)`.

francoo98: I didn't notice.
Turns out there is a bug in my `create_test_file` function, it wasn't creating a 50KB file, it actually was much longer than that because `bytes()` create an array of zeroes when you pass it an int (what I am doing in line 791). So, I was actually creating a file filled with zeroes.
Now, I modified the code to use `create_regular_file` as you said, but it now requires 6MB of random bytes to pass the test.

src/borg/archiver/create_cmd.py
```diff
@@ -185,6 +187,8 @@ def create_inner(archive, cache, fso):
                 if args.progress:
                     archive.stats.show_progress(final=True)
                 archive.stats += fso.stats
+                archive.stats.rx_bytes = getattr(repository, "rx_bytes", 0)
+                archive.stats.tx_bytes = getattr(repository, "tx_bytes", 0)
```
ThomasWaldmann, line 191: shouldn't the rx/tx_bytes also use `+=`?

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -731,6 +772,42 @@ def test_common_options(self):
...
+            borg_create_output = borg_create_output.strip().splitlines()
+            borg_create_output = [line.split(":", 1) for line in borg_create_output]
+            hashing_time = [line for line in borg_create_output if line[0] == "Time spent in hashing"].pop()
+            hashing_time = hashing_time[1]
+            hashing_time = float(hashing_time.removesuffix(" seconds"))
+            return hashing_time
+
+        # Test case set up: create a repository and a file
+        self.cmd(f"--repo={self.repository_location}", "rcreate", "--encryption=none")
+        self.create_regular_file("testfile", size=6000000, contents=randbytes(6000000))
```
ThomasWaldmann, line 786: you don't need to give size if you directly give the contents.

ThomasWaldmann: btw, guess the root cause for that you need a rather large file to detect `> 0.0` is in `format_timedelta` which needs at least 10ms for that:

https://github.com/borgbackup/borg/blob/3715e327a6ea1306b7fbd3c37f5bdec38b0439bb/src/borg/helpers/time.py#L102

src/borg/archiver/create_cmd.py
francoo98: No because those are the only lines where `rx_bytes` and `tx_bytes` are manipulated so they are both zero at that moment

ThomasWaldmann: ok

src/borg/testsuite/archiver/create_cmd.py
francoo98: What should I do about it?

src/borg/archive.py
```diff
@@ -81,15 +86,36 @@ def __add__(self, other):
         stats.osize_parts = self.osize_parts + other.osize_parts
         stats.usize_parts = self.usize_parts + other.usize_parts
         stats.nfiles_parts = self.nfiles_parts + other.nfiles_parts
+        stats.chunking_time = self.chunking_time + other.chunking_time
+        stats.hashing_time = self.hashing_time + other.hashing_time
+        for key in other.files_stats.keys():
```
ThomasWaldmann, line 91: cosmetic: the code as you wrote it works, but even more elegant is to use the default behaviour:
```
for key in some_dict:
```
That automatically iterates over the keys in some_dict by default.

src/borg/testsuite/archiver/create_cmd.py
francoo98: Btw, you are right. I modified the value of txt to `"%.3f seconds" % s` and the test passes with 1MB of data.
But I don't think I should push that modification right?

src/borg/archive.py
```diff
@@ -1431,19 +1464,22 @@ def process_file(self, *, path, parent_fd, name, st, cache, flags=flags_normal):
...
                         with backup_io("read"):
+                            started_chunking = time.monotonic()
                             self.process_file_chunks(
                                 item,
                                 cache,
                                 self.stats,
                                 self.show_progress,
                                 backup_io_iter(self.chunker.chunkify(None, fd)),
                             )
+                            self.stats.chunking_time += time.monotonic() - started_chunking
```
ThomasWaldmann, line 1482: Hmm, guess this measures way more than just the chunking time:

it also measures the time to do **all processing** for each chunk, right?

guess we should rather only measure **reading and chunking** (== cutting into pieces) of the input data for this stats value (separating reading from chunking might be hard, the buzhash chunker is C code and does reading via chunking buffer refills, so guess we just measure both together for simplicity).

but after chunks are cut, they get:
- hashed (id_hash) - you already have that (and this is interesting because it might be slow on some CPUs)
- compressed (very interesting, this can be very fast or very slow, depending on algorithm and level)
- encrypted (very interesting, can be very fast or very slow, depending on cpu and algorithm)
- (transmitted and) written to repo (also very interesting, can be slow for slow remote repos)

ThomasWaldmann: for this PR, maybe you could fix the chunking_time to really only measure chunking/reading.

but then I guess it would be good to merge this PR before working on the other items to not widen the scope too much, we already have quite a lot of commits here.

src/borg/testsuite/archiver/create_cmd.py
ThomasWaldmann: Yeah, guess we should do that modification.

It could well be that we run into problems again after doing the fix to only measure real read/chunk time (without all the processing) and would need to increase the amount of data if we don't do that modification.

Considering that the next unit below s is ms, having 3 digits there even makes some sense.

src/borg/archive.py
francoo98: So, do I have to measure only the call of `chunkify`? Line 1481 there.

ThomasWaldmann: That's not a simple call, but that returns an iterator. Have a look into `chunker.pyx -> ChunkerFixed`.

ThomasWaldmann: `*.pyx` is cython code and needs recompiling after changes, easiest is something like:
```
python setup.py clean clean2 ; pip install -e .
```

francoo98: As I understand, the chunking is done lazily at `Chunker.__next__` and `ChunkerMixed.chunkify` (a generator), so I should measure these methods.
This is easy to do, but these methods don't have access to the Statistics object.
I think I could add the time it took to read the chunk as an attribute to the Chunk class, so the chunker methods could yield/return like this: `yield Chunk(data, size=got, allocation=allocation, chunking_time=measured_time)`.
And the process_file_chunks would look something like this:
```python
for chunk in chunk_iter:
            stats.chunking_time += chunk.chunking_time
            item.chunks.append(chunk_processor(chunk))
...
```

ThomasWaldmann: Guess you could either give stats to the chunker init or have the timing in the chunker instance and pick it from there from the calling code.

We're only interested in the overall chunking time, so better do not put it into the Chunk instance, but into the Chunker instance.

francoo98: Am I gonna need to modify pyx and c files to do that?

ThomasWaldmann: Likely the chunker.pyx, just have a look.

ThomasWaldmann (commented):
BTW, there is a `time.sleep(1)` in the test you added, is it needed / why?

francoo98:
> BTW, there is a `time.sleep(1)` in the test you added, is it needed / why?
It is needed because the second file needs newer timestamp than the first file.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -626,6 +627,46 @@ def test_file_status_excluded(self):
...
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        # Archive an empty dir
+        result = self.cmd(f"--repo={self.repository_location}", "create", "--stats", "test_archive", self.input_path)
+        result = to_dict(result)
+        assert result["Added files"] == 0
+        assert result["Unchanged files"] == 0
+        assert result["Modified files"] == 0
+        # Archive a dir with two added files
+        self.create_regular_file("testfile1", contents=b"test1")
+        time.sleep(1)  # testfile2 must have newer timestamps than testfile1
```
ThomasWaldmann, line 653: can you try keeping the comment, but removing the "sleep" or making it much shorter?

there are only few fs with very rough granularity (e.g. FAT and similar) and we do not use them for testing.
most other fs nowadays have rather fine granularity, like us or ns.

a test run executes > 1000 automated tests and there should be only a sleep if really needed.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -650,7 +650,7 @@ def to_dict(borg_create_output):
         assert result["Modified files"] == 0
         # Archive a dir with two added files
         self.create_regular_file("testfile1", contents=b"test1")
-        time.sleep(1)  # testfile2 must have newer timestamps than testfile1
+        time.sleep(0.004)  # testfile2 must have newer timestamps than testfile1
```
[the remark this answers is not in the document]
francoo98: 0.004 is the smallest that worked on my machine. Smaller numbers made the test pass randomly (sometimes it passes and sometimes it doesn't)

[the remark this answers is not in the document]
ThomasWaldmann: did you find out why?

as i noted, modern fs should have timestamp with microseconds resolution or better.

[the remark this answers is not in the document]
francoo98: I checked the timestamps using the `stat` command in bash and the two files have the same timestamp down to the smallest decimal.
Maybe there is an OS level optimization that delays the creation of the files?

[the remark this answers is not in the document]
ThomasWaldmann: @francoo98 can you check if an fsync and no delay solves it also?

@jdchristensen otoh, in case of the fsync taking >= 0.004s, we do not know why it worked (because of what it fundamentally does or because of how long that took).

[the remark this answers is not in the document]
jdchristensen: I did some tests, and using os.sync() instead of the time command works in practice.  @ThomasWaldmann, it's true that I can't be sure *why* it works, but it's still likely better than a hardcoded number that might fail on a slow system.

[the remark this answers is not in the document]
jdchristensen: (I also confirmed on my system that time.sleep(0.003) often produces files with identical time stamps.  This is suspicious, since ext4 should have better resolution than that.)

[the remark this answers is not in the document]
jdchristensen: Ah, there is an explanation here: https://stackoverflow.com/questions/70539589/file-seems-to-be-timestamped-before-it-is-created

It says that the time used for filesystem operations is only updated roughly every 10ms.  I just tried a loop that created 1000 files using `create_regular_file`, and there were only three different timestamps, almost exactly 4ms apart:
```
 Birth: 2022-10-16 17:15:15.092035097 -0400
 Birth: 2022-10-16 17:15:15.096035113 -0400
 Birth: 2022-10-16 17:15:15.100035129 -0400
```
(In some runs, the gap was 8ms, probably because the process had to give up the cpu for a bit.)

With a 0.003s sleep, there were ~800, and with a 0.004s sleep or `os.sync()` there were 1000.  I would propose either the sync method or a longer sleep (maybe 0.01s = 10ms?) to be safe.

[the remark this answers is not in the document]
ThomasWaldmann: Interesting, would have expected it to be more fine grained. OK, let's use 0.01s for the sleep so we do not get spurious failures.

About the 4ms: maybe that timer that used to be 100Hz (10ms) is now 250Hz (4ms)?

https://stackoverflow.com/questions/14925568/lowering-linux-kernel-timer-frequency

But still, we want these tests to work in general, so we would prepare for worst case (100Hz).

src/borg/chunker.pyx
```diff
@@ -275,9 +283,9 @@ cdef class Chunker:
             allocation = CH_ALLOC
         else:
             allocation = CH_DATA
+        self.chunking_time += monotonic() - started_chunking
         return Chunk(data, size=got, allocation=allocation)
 
-
```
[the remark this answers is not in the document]
ThomasWaldmann: did you fix this already?

ThomasWaldmann:
@francoo98 thanks for the PR, it's merged now!

francoo98:
That's great 😁!
Are you going to close the issue?

PR #6990, opened 2022-08-23
xattrs / extended stat: improve exception handling (1.2-maint)

ThomasWaldmann:
see #6988

PR #7008, opened 2022-09-07
xattrs / extended stat: improve exception handling (master)

KN4CK3R:
forward port of #6990

The different quotes in both branches made some trouble but should be fine now.

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -1560,7 +1560,7 @@ class ArchiverTestCase(ArchiverTestCaseBase):
             assert xattr.getxattr(b"input/file", b"security.capability") == capabilities
 
     @pytest.mark.skipif(
-        not xattr.XATTR_FAKEROOT, reason="xattr not supported on this system or on this version of" "fakeroot"
+        not xattr.XATTR_FAKEROOT, reason="xattr not supported on this system or on this version of " "fakeroot"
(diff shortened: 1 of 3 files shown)


src/borg/testsuite/archiver.py
```diff
@@ -1560,7 +1560,7 @@ def patched_fchown(fd, uid, gid):
             assert xattr.getxattr(b"input/file", b"security.capability") == capabilities
 
     @pytest.mark.skipif(
-        not xattr.XATTR_FAKEROOT, reason="xattr not supported on this system or on this version of" "fakeroot"
+        not xattr.XATTR_FAKEROOT, reason="xattr not supported on this system or on this version of " "fakeroot"
```
ThomasWaldmann, line 1563: hmm, if that's what black made out of this, guess we better merge the 2 strings into 1.

see also the other 2 places with same string.

ThomasWaldmann (changes requested):

src/borg/testsuite/archiver.py
ThomasWaldmann: @KN4CK3R did you see my feedback?

ThomasWaldmann (approved):
fixed the cosmetics myself so i can merge this soon.

KN4CK3R:
Sorry, was not a home.

PR #7019, opened 2022-09-10
OpenBSD only - statically link OpenSSL (#6474)

bket:
Avoid conflicting with shared libcrypto from the base OS pulled in via dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,13 +154,15 @@ if not on_rtd:
             f"or ensure {lib_pkg_name}.pc is in PKG_CONFIG_PATH."
         )
 
-    crypto_ldflags = []
+    crypto_extra_objects = []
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@bket it is best to run `black .`  in the workdir before committing - black will then fix all the formatting details for you.

ThomasWaldmann:
BTW, i was just working on releasing 2.0.0b2, but maybe we should get this in before b2, so it can get some testing on openbsd.

bket:
> @bket it is best to run `black .` in the workdir before committing - black will then fix all the formatting details for you.

Done, force pushed the fix

ThomasWaldmann:
Thanks!

PR #7024, opened 2022-09-13
do not use version_tuple placeholder in setuptools_scm template

ThomasWaldmann:
that would require setuptools_scm>=5.0.0 but some dists do not have that yet.

also, we do not use the version_tuple from _version.py, so it is not required anyway.


ThomasWaldmann:
@LocutusOfBorg did you test this change meanwhile? merge?

LocutusOfBorg (approved):
Looks working fine on Ubuntu jammy and Sebian sid.

PR #7028, opened 2022-09-16
implement pattern support for --match-archives, fixes #6504

### LINKED ISSUE #6504: `--glob-archive` does not work with regexp in borg list
Not sure, may be not 1.2 version specific, but `--glob-archive` does not working with regexp:

```
borg list user123.mysql --prefix user123-2 | awk '{print $1}'
user123-2022-03-20-065449
...
user123-2022-03-28-031228

borg list user123.mysql -a "re:^user123"
# empty output
```

In my scenario I have to select snapshots to prune with filter based on prefix (exact match) and suffix (regexp), Seems to be not possible now.

From https://borgbackup.readthedocs.io/en/stable/usage/list.html:

> only consider archive names matching the glob. sh: rules apply, see “borg help patterns”.

There is nothing about other patterns should not work in case they are specified, and there is definitely should be the way to select snapshots by prefix AND suffix, and both regexp.

I'm using borg 1.2.0 from compiled github releases, "linuxold" version.

_Originally posted by @knutov in https://github.com/borgbackup/borg/discussions/6337#discussioncomment-2452748_

ThomasWaldmann:
also:
- rename --glob-archives option to --match-archives (short: -a)
- globbing patterns now need sh: prefix
- regex patterns need re: prefix
- "identical" match "pattern" uses an id: prefix
- new default is id: pattern (--glob-archives used sh: glob pattern)
- source code: glob -> match, GLOB -> PATTERN


ThomasWaldmann:
default style: shall we rather default to the simpler `id:` or the more powerful `sh:`?

id: might be a bit more on the safe side (avoiding unintentional use of wildcards or other magic).
sh: might be more on the comfortable side (not needing to prepend `sh:`). we also default to shell style patterns for `--pattern` option.

ThomasWaldmann:
Note: the borg 1.2 pendant (`--glob-archives`) used globbing (== `sh:`).

ThomasWaldmann:
```
15:11  tao$ i think a basic string match should be the default with glob/regex matching a configurable prefix
```

PR #7039, opened 2022-09-19
update automating backups script

Mannuel25:


ThomasWaldmann:
All the backslashes, except the one in the last line, are required. The tell the shell that the command is continued on the next line.

Also the `--glob-archives` need to match the hostname as this is used in the `borg create` command.

You can just copy and paste the script after you changed it and practically try it out to find any issues.

ThomasWaldmann:
Also, `-a` is the short alias of `--glob-archives` - one just decides for one of them. For an example script, using the long form is usually better readable.

Mannuel25:
Okay, thanks for the corrections.

docs/quickstart.rst
```diff
@@ -202,13 +202,13 @@ backed up and that the ``prune`` command is keeping and deleting the correct bac
...
-    borg prune                          \
-        --list                          \
-        --prefix '{hostname}-'          \
-        --show-rc                       \
-        --keep-daily    7               \
-        --keep-weekly   4               \
-        --keep-monthly  6               \
+    borg prune                      \
+    --list                          \
+    --glob-archives '{hostname}-'   \
```
ThomasWaldmann, line 207: needs to be `--glob-archives '{hostname}-*'`.

also maybe rather keep the indentation of the lines with the options, so it is better to see that they belong to the first line with the beginning of the command.

In line 211, remove all spaces before the line end, so there are no trailing spaces.

also, fix the create command in the same way.

Mannuel25: Okay, I'll fix it. But I didn't get you here: `keep the indentation of the lines with the options, so it is better to see that they belong to the first line with the beginning of the command.`

Mannuel25: Okay, I'll fix it. Thanks

ThomasWaldmann (approved):
Looks good now!

Mannuel25:
> Looks good now!

Thanks for helping out 😁

ThomasWaldmann:
Merged, thanks!

Mannuel25:
> Merged, thanks!

Thank you!

PR #7049, opened 2022-09-23
reformulate how to counter a large cache

nbehrnd:
Based on discussion which mentioned a fitting section of the FAQ ([link](https://github.com/borgbackup/borg/discussions/7048)).  If used to use borg *managed by the vorta GUI*, however, the snippet's syntax is a bit unusual.  The edit aims to ease the comprehension.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -715,25 +715,23 @@ which might be unwanted.
 A temporary (but maybe long lived) hack to avoid using lots of disk memory
 for ``chunks.archive.d`` (see :issue:`235` for details) is to replace this
 directory with an empty dummy file of this name.  Note, below,
-``$(borg config id)`` is used as place holder for the alphanumeric string
-specific to your backup repository (REPO_ID) you have to enter instead.
-You identify this string of 64 characters either in your file browsers, or
-by
+``$(borg config id)`` resolves into alphanumeric string specific to your
+backup repository (REPO_ID) you may identify in your file manager, or by
 
 ::
 
   borg config id
 
-With this information, and assuming you work with the same user as in the
-backup, adjust your setup
+With this information, and assuming you work with the same user as for
+the backup, adjust your setup
 
 ::
 
   cd ~/.cache/borg/$(borg config id)
   rm -rf chunks.archive.d ; touch chunks.archive.d
 
-The deletion of the cached archive chunk and presence of the dummy file
-prevents borg to store anything "in" there in future.
+The presence of the dummy file prevents borg to store anything "in" there
+in future.
 
 This has some pros and cons, though:
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
 from the repository). While being faster, the cache needs quite some disk space,
 which might be unwanted.
 
-There is a temporary (but maybe long lived) hack to avoid using lots of disk
-space for chunks.archive.d (see :issue:`235` for details):
+A temporary (but maybe long lived) hack to avoid using lots of disk memory
+for ``chunks.archive.d`` (see :issue:`235` for details) is to replace this
+directory with an empty dummy file of this name.  Note, below,
+``$(borg config id)`` is used as place holder for the alphanumeric string
+specific to your backup repository (REPO_ID) you have to enter instead.
+You identify this string of 64 characters either in your file browsers, or
+by
 
 ::
 
-    # this assumes you are working with the same user as the backup.
-    cd ~/.cache/borg/$(borg config id)
-    rm -rf chunks.archive.d ; touch chunks.archive.d
+  borg config id
 
-This deletes all the cached archive chunk indexes and replaces the directory
-that kept them with a file, so borg won't be able to store anything "in" there
-in future.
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
 from the repository). While being faster, the cache needs quite some disk space,
 which might be unwanted.
 
-There is a temporary (but maybe long lived) hack to avoid using lots of disk
-space for chunks.archive.d (see :issue:`235` for details):
+A temporary (but maybe long lived) hack to avoid using lots of disk memory
```
ThomasWaldmann, line 715: AFAIK, the usual term is **disk space**, not **disk memory**.

docs/faq.rst
```diff
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
...
+for ``chunks.archive.d`` (see :issue:`235` for details) is to replace this
+directory with an empty dummy file of this name.  Note, below,
+``$(borg config id)`` is used as place holder for the alphanumeric string
+specific to your backup repository (REPO_ID) you have to enter instead.
```
ThomasWaldmann, line 719: This is not a placeholder that has to get manually replaced.

The shell will call the command `borg config id` and replace that term by the output of that command.

The command in this short form will only work if `BORG_REPO` env var is set, otherwise one would have to add the repo location to the command.

docs/faq.rst
```diff
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
...
+You identify this string of 64 characters either in your file browsers, or
```
ThomasWaldmann, line 720: The usual term is **file manager**.

docs/faq.rst
ThomasWaldmann: Also, because `borg config` only works for local repos, this does not work for remote repos.

docs/faq.rst
```diff
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
...
 
-    # this assumes you are working with the same user as the backup.
-    cd ~/.cache/borg/$(borg config id)
-    rm -rf chunks.archive.d ; touch chunks.archive.d
+  borg config id
 
-This deletes all the cached archive chunk indexes and replaces the directory
-that kept them with a file, so borg won't be able to store anything "in" there
-in future.
+With this information, and assuming you work with the same user as in the
```
ThomasWaldmann, line 727: "as **for** the backup"?

docs/faq.rst
```diff
@@ -712,18 +712,28 @@ will make the subsequent rebuilds faster (because it needs to transfer less data
...
+backup, adjust your setup
+
+::
+
+  cd ~/.cache/borg/$(borg config id)
+  rm -rf chunks.archive.d ; touch chunks.archive.d
+
+The deletion of the cached archive chunk and presence of the dummy file
+prevents borg to store anything "in" there in future.
```
ThomasWaldmann, line 736: The presence of the dummy file prevents borg from storing anything "in" there in future.

ThomasWaldmann (changes requested):
Thanks for the PR!

Guess it needs some changes in grammar as well as "don't fix what ain't broken".

docs/faq.rst
nbehrnd: I agree on using "disk space" instead of "disk memory".

docs/faq.rst
nbehrnd: So far, I used borg only indirectly with vorta as a GUI.  Based on your explanation I recognize I did not comprehend the intent of ``$(borg config id)``, and retract the suggestion.

docs/faq.rst
nbehrnd: So this was influenced by [Apple advertisement](https://apps.apple.com/us/app/filebrowser-document-manager/id364738545) [English Wikipedia](https://en.wikipedia.org/wiki/File_manager) noting both forms; then sticking to the proper purpose.  Your point is convincing, file manager.

docs/faq.rst
nbehrnd: ``with the same user as the backup`` was reused as found in the same section for commit 60060689 by September 21 (though your commit addressed a different topic).

docs/faq.rst
nbehrnd: It is more succinct.

nbehrnd (commented):
The suggested changes are approved.

ThomasWaldmann:
@nbehrnd can you update your PR with the suggested changes?

docs/faq.rst
```diff
@@ -715,25 +715,23 @@ which might be unwanted.
...
+the backup, adjust your setup
 
 ::
 
   cd ~/.cache/borg/$(borg config id)
   rm -rf chunks.archive.d ; touch chunks.archive.d
 
-The deletion of the cached archive chunk and presence of the dummy file
-prevents borg to store anything "in" there in future.
+The presence of the dummy file prevents borg to store anything "in" there
```
jdchristensen, line 733:
```suggestion
The presence of the dummy file prevents borg from storing anything "in" there
```

PR #7064, opened 2022-09-30
Hashindex header work, fixes #6960

### LINKED ISSUE #6960: HashIndex header size not a multiple of 8
As just discussed in IRC: The current HashIndex (on disk!) data structure has a header whose size is not a multiple of 4 or 8. This makes it a bit cumbersome (and on some platforms also probably slower) to use when mmap'd.

I understand that mmap is not used within borg to access the hashindex-based structures at this point in time, so it may not be worth fixing. Yet it seems like a small enough change to squeeze into Borg 2.0 which may avoid some pain further down the road (if borg ever goes to mmap'ing the indices).

ThomasWaldmann:


src/borg/_hashindex.c
```diff
@@ -39,6 +40,18 @@ typedef struct {
...
+
+BORG_PACKED(
+typedef struct {
+    char magic[MAGIC_LEN];
+    int32_t version;
+    int32_t num_entries;
+    int32_t num_buckets;
+    int32_t num_empty;
+    int32_t key_size;
+    int32_t value_size;
```
horazont, line 53: Is this a good opportunity to get rid of signed values for counts and sizes?

src/borg/_hashindex.c
```diff
@@ -351,23 +359,168 @@ hashindex_read(PyObject *file_py, int permit_compact)
...
+        PyErr_NoMemory();
+        goto fail_decref_header;
+    }
+
+    PyObject_GetBuffer(header_bytes, &header_buffer, PyBUF_SIMPLE);
+    if(PyErr_Occurred()) {
+        goto fail_free_index;
+    }
+
+    header = (HashHeader*) header_buffer.buf;
```
horazont, line 466: Hm, why does this not cause issues with unaligned memory accesses on non-x86 architectures?

I mean it seems like it does not cause such issues, because the old code does the same and I have it running on the same platform I found [the lack of this mpd commit to be an issue](https://github.com/MusicPlayerDaemon/MPD/commit/fb8d8242abdbb0001012c5b91b4e2ca69553b166), but it leaves me curious.

Are we just lucky that `header_buffer.buf` is actually aligned to an 8-byte boundary and the header magic is a multiple of 8?

src/borg/_hashindex.c
```diff
@@ -481,33 +632,34 @@ hashindex_free(HashIndex *index)
...
         .magic = MAGIC,
+        .version = _htole32(2),
         .num_entries = _htole32(index->num_entries),
         .num_buckets = _htole32(index->num_buckets),
-        .key_size = index->key_size,
-        .value_size = index->value_size
+        .num_empty = _htole32(index->num_empty),
+        .key_size = _htole32(index->key_size),
+        .value_size = _htole32(index->value_size)
     };
```
horazont, line 649: My C is a bit rusty, does this zero-initialize the `reserved` space? Would be good to have it always be zero to allow for extensions.

src/borg/_hashindex.c
```diff
@@ -520,9 +672,25 @@ hashindex_write(HashIndex *index, PyObject *file_py)
...
+#ifndef BORG_NO_PYTHON
+static void
+hashindex_write(HashIndex *index, PyObject *file_py, int legacy)
+{
+    PyObject *length_object, *buckets_view;
+    Py_ssize_t length;
+    Py_ssize_t buckets_length = (Py_ssize_t)index->num_buckets * index->bucket_size;
+
+    if (legacy)
+        1 / 0;  // guess we do not write old hashindexes
```
horazont, line 690: What is this about?

src/borg/_hashindex.c
```diff
@@ -39,6 +40,18 @@ typedef struct {
...
+typedef struct {
+    char magic[MAGIC_LEN];
+    int32_t version;
+    int32_t num_entries;
+    int32_t num_buckets;
+    int32_t num_empty;
+    int32_t key_size;
+    int32_t value_size;
+    char reserved[1024 - 32];  // filler to 1024 bytes total
 }) HashHeader;
```
horazont, line 55: Can we `_Static_assert(sizeof(HashHeader) == 1024, "hash header should be exactly 1024 bytes in size")`, or can't we rely on C11 support?

src/borg/_hashindex.c
```diff
@@ -39,6 +40,18 @@ typedef struct {
...
+BORG_PACKED(
+typedef struct {
+    char magic[MAGIC_LEN];
+    int32_t version;
+    int32_t num_entries;
+    int32_t num_buckets;
+    int32_t num_empty;
+    int32_t key_size;
+    int32_t value_size;
+    char reserved[1024 - 32];  // filler to 1024 bytes total
```
horazont, line 54: Do we want to go for 4096, a common memory page size, right away?

I don't see any obvious benefit to it right now, but I wanted to raise the question anyway in case someone else has an idea.

src/borg/_hashindex.c
```diff
@@ -351,23 +359,168 @@ hashindex_read(PyObject *file_py, int permit_compact)
...
+    tmp = PyObject_CallMethod(file_py, "hash_part", "s", "HashHeader");
+    Py_XDECREF(tmp);
+    if(PyErr_Occurred()) {
+        if(PyErr_ExceptionMatches(PyExc_AttributeError)) {
+            /* Be able to work with regular file objects which do not have a hash_part method. */
+            PyErr_Clear();
+        } else {
+            goto fail_decref_header;
+        }
+    }
```
horazont, line 435: I don't know how this works (I took a brief look at `file_integrity.py`, but I'm not awake enough yet to follow it completely), so the obvious question: The hash is not in a place where it creates an offset for the post-header data, right? I.e. this file integrity thing does not break the alignment, does it?

src/borg/_hashindex.c
ThomasWaldmann: could do that, but then the num_empty = -1 would need to get changed to e.g. 0xffffffff (which should be save, because we would make the HT smaller when having that many empty buckets).

src/borg/_hashindex.c
ThomasWaldmann: Well I'ld hope python is clever enough to allocate its stuff in a well-aligned way.

src/borg/_hashindex.c
ThomasWaldmann: I guess the extension mechanism would be to increase the version number.

Then we don't have to care about it being 0, right?

src/borg/_hashindex.c
ThomasWaldmann: old hashindex is only used when reading from borg 1 repos.

borg 2 users might do that once per repo to transfer old archives to new borg 2 repos.
but as the old repo is only read, writing the old hashindex is not used.

in case my assumption were wrong, i just wanted it to blow up there.

src/borg/_hashindex.c
ThomasWaldmann: i rarely do C programming these days, so not sure what we can rely on or what would be best to rely on.

borg 2 is expected to run on rather new systems (it also requires python 3.9), but supports a lot of platforms, thus the code should be portable.

src/borg/_hashindex.c
horazont: Explicitly zeroing it avoids writing random stack contents to disk still, which is probably a good idea.

src/borg/_hashindex.c
ThomasWaldmann: I thought about this and guess it is pointless:

The size of the hashtable is header_size + N * bucket_size and N is pretty arbitrary (currently guess it is usually a prime value due to some unneccessary change in the past).

So, no matter what we choose for header_size, it will use a fraction of a page at the end.
Due to that, it does not matter if we use a fraction of a page at the beginning also.

Besides that, page size on apple silicon (e.g. M1) is 16k.

So I just picked 1k because:
- not too big so it would add significant overhead
- not too small to ever would need to get changed

src/borg/_hashindex.c
ThomasWaldmann: no, the integrity data is in a separate file.

src/borg/_hashindex.c
ThomasWaldmann: Another thing is the on-disk block size for HDDs and SSDs (in case we ever use mmap again), but guess the same consideration applies there also.

src/borg/_hashindex.c
ThomasWaldmann: yeah, indeed.

ThomasWaldmann: not sure what we have as assured default behaviour though.

src/borg/_hashindex.c
horazont: TIL about a relevant architecture with non-4k pages. So I guess it indeed does not matter at all (and I don't think there are actual cases where one might want to map the rest of the hashindex, but not the header).

src/borg/_hashindex.c
horazont: I suggest we add that, and if any compiler complains about lack of `_Static_assert`, fixing that can still be figured out? It's not like it'll cause any weird behaviour or breakage which is not detected ahead of shipping time.

ThomasWaldmann: ok, I'll add it.

src/borg/_hashindex.c
ThomasWaldmann: done: init reserved to zero before writing to disk.

src/borg/_hashindex.c
ThomasWaldmann: i checked the code and the internal in-memory data struct and the rest of the code also uses signed int.

thus, while it could be done, it is not in scope of this PR to refactor this.

ThomasWaldmann:
@horazont ok, guess that's it for now? are you still reviewing / testing?

PR #7093, opened 2022-10-21
tar_filter: recognize .tar.zst as zstd

nyuszika7h:
`.zst` is actually the standard extension as far as I know.

### DIFF
--- a/docs/man/borg-export-tar.1
+++ b/docs/man/borg-export-tar.1
@@ -50,7 +50,7 @@ before writing it to FILE:
 .IP \(bu 2
 \&.tar.xz or .txz: xz
 .IP \(bu 2
-\&.tar.zstd: zstd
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
https://www.rfc-editor.org/rfc/rfc8878.html mentions also `.zst` as extension.

Not sure whether we should keep `.zstd`, but it should be harmless at least.

ThomasWaldmann:
Could you also do a PR against 1.2-maint branch, please?

ThomasWaldmann:
Did you manually edit the docs? These files are generated, see the comments in line 1. The place to edit usually is somewhere in `borg.archiver` python source code.

Usually, we do not update the generated docs before release time (this is part of the release process), because it just makes the PRs harder to review / the diffs bigger.

ThomasWaldmann:
Thanks!

PR #7096, opened 2022-10-22
windows build updates

ThomasWaldmann:
- try to be closer to github CI requirements installation
- remove outdated env vars (no libs source code bundled any more with borg)


ThomasWaldmann:
this stuff is very broken and i won't fix it.

PR #7101, opened 2022-10-26
Fix test_chunking_time

francoo98:
I will fix test_chunking_time in this pr.
Since I can't see the test failing anywhere I don't really know what the error is. My guess is that the execution of the test is too fast, so I gave the test file more info.
I hope the CI here can clarify the issue.

### DIFF
--- a/src/borg/testsuite/archiver/create_cmd.py
+++ b/src/borg/testsuite/archiver/create_cmd.py
@@ -801,7 +801,7 @@ class ArchiverTestCase(ArchiverTestCaseBase):
 
         # Test case set up: create a repository and a file
         self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
-        self.create_regular_file("testfile", contents=randbytes(5000000))
+        self.create_regular_file("testfile", contents=randbytes(5500000))
         # Archive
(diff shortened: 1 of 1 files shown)


src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -801,7 +801,7 @@ def extract_chunking_time(borg_create_output):
 
         # Test case set up: create a repository and a file
         self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
-        self.create_regular_file("testfile", contents=randbytes(5000000))
+        self.create_regular_file("testfile", contents=randbytes(5500000))
```
ThomasWaldmann, line 804: How about just going to 10MB?

I guess increasing it just by 10% would still make it fail almost as frequently as it fails now.

ThomasWaldmann:
> Since I can't see the test failing anywhere I don't really know what the error is.

Well, I told you that the chunking_time test fails and the only assertion you do there is `> 0.0` - so the only way that can fail is if it is `0.0` due to rounding.

> My guess is that the execution of the test is too fast

Exactly.

ThomasWaldmann:
Thanks!

PR #7106, opened 2022-10-27
Use MSYS2 for Windows CI [backport]

### LINKED ISSUE #7105: backport building the windows exe to 1.2-maint branch
related to #7103:

we have basically the same issue in 1.2-maint branch, but there are some differences in borg setup and requirements.

the goal is to remove appveyor also in 1.2-maint and replace it with a similar github workflow as in master branch to build a borg exe.

while doing that, maybe an experiment could be done to separate installing the python/tool environment (via msys2) and installing borg's direct requirements (via pip) - if that works, it would be the preferred way because we then do not have to maintain borg requirements list at multiple places.

bounty: https://app.bountysource.com/issues/113092438-backport-building-the-windows-exe-to-1-2-maint-branch

RayyanAnsari:
Fixes #7105


RayyanAnsari:
@ThomasWaldmann this PR is ready for merge as well

ThomasWaldmann:
Did you try using `pip install -e .` to install borg **with dependencies** (and not install these dependencies via msys2)?

RayyanAnsari:
Yes I have, both inside a `venv` and without one.
With the `venv` it fails during the `pip install -e .` in setuptools.
Without it it seems to break `pip` itself - even running the `pip` command on its own returns errors. Probably there are conflicts between files installed by `pacman` and files installed by `pip`.

ThomasWaldmann:
Thanks!

PR #7108, opened 2022-10-30
Sanitize paths during archive creation and extraction

### LINKED ISSUE #7099: `borg import-tar` does not strip leading `./` from filenames, making mount access impossible

Yes

Bug report

borg 1.2.2 (not sure what version Hetzner uses as I can't execute the `borg` command directly on their storage box)

Debian 11

Source is a `.tar` file, file system should be irrelevant. But just in case, the .tar file was created on ext4, the current system is running btrfs, and the backup was made to a ZFS remote.

About 100 GB at most.

```
borg import-tar --tar-filter='zstd -d' --progress --compression zstd ssh://XXX@XXX.your-storagebox.de:23/./backup/cadoth::2022-07-29 ~/Downloads/backup-20220729.tar.zst
```
```
sudo borg mount -v XXX@XXX.your-storagebox.de:23/./backup/cadoth /mnt/backup
```

When a tar archive was created as `tar -cf NAME.tar .` or similar (emphasis on the `.`), it leads to filenames starting with `./` in the archive. Borg does not strip this either on `import-tar` or `mount` by default, which results in it creating a second `.` directory, but that is "shadowed" by the regular "current directory" `.` entry, making it impossible to access. I even tried checkijng inode numbers but they seem to show up as identical.
```
root@cadoth /mnt/backup/2022-07-30 # ls -ail
total 0
4 drwxr-xr-x 1 root root 0 Oct 22 02:08 .
4 drwxr-xr-x 1 root root 0 Oct 22 02:08 .
1 drwxr-xr-x 1 root root 0 Oct 22 11:11 ..
```

Yes, this is consistently reproducible with a tar archive created using the above syntax.

pgerber:
Paths are not always normalized when creating an archive and, more importantly, never when extracting one. The following example shows how this can be used to attempt to write a file outside the extraction directory:

```
$ echo abcdef | borg create -r ~/borg/a --stdin-name x/../../../../../etc/shadow archive-1 - $ borg list -r ~/borg/a archive-1
-rw-rw---- root   root          7 Sun, 2022-10-23 19:14:27  x/../../../../../etc/shadow
$ mkdir borg/target
$ cd borg/target
$ borg extract -r ~/borg/a archive-1
x/../../../../../etc/shadow: makedirs: [Errno 13] Permission denied: '/home/user/borg/target/x/../../../../../etc'
```

Note that borg tries to extract the file to /etc/shadow and the permission error is a result of the user not having access.

This patch ensures file names are normalized before archiving. As for files extracted from the archive, paths are normalized by making all paths relative, removing './', and removing superfluous slashes (as in '//'). '..' elments, however, are rejected outright. The reasoning here is that it is easy to start a path with './' or insert a '//' by accident (e.g. via --stdin-name or import-tar). '..', however, seem unlikely to be the result of an accident and could indicate a tampered repository.

With paths being normalized as they are being read, this "errors" will be corrected during the `borg transfer` required when upgrading to Borg 2. Hence, the normalization, when reading the archive, can be removed once support for reading v1 repositories is dropped. v2 repository will not contain non-normalized paths.

I recommend treating this as a security issue. I see the following cases where extracting a file outside the extraction path could constitute a security risk:

a) When extraction is done as a different user than archive creation. The user that created the archive may be able to get a file overwritten as a different user.
b) When the archive is created on one host and extracted on another. The user that created the archive may be able to get a file overwritten on another host.
c) When an archive is created and extracted after a OS reinstall. When a host is suspected compromised, it is common to reinstall (or set up a new machine), extract the backups and then evaluate their integrity. A user that manipulates the archive before such a reinstall may be able to get a file overwritten outside the extraction path and may evade integrity checks.

Notably absent is the creation and extraction on the same host as the same user. In such case, an adversary must be assumed to be able to replace any file directly.

This also (partially) fixes #7099.


ThomasWaldmann:
@pgerber good catch, thanks for finding/fixing this.

i agree that there are some potential security issues related to this, but I think they are low severity (compared to tar, zip, etc.) because borg repos/archives usually do not come from unknown/untrusted sources and also borg is not commonly used for data transfer. usually, you either made the repo/archive by yourself or it was made by some fellow admin of yours.

but, as you pointed out, some risks remain.

btw, i noticed quite some typos in your commit comment, can you proofread it / fix them?

ThomasWaldmann:
> With paths being normalized as they are being read, this "errors" will be corrected during
> the borg trasfer required when upgrading to Borg 2. Hence, the normalization, when reading
> the archive, can be removed once support for reading v1 repositories is dropped. v2 repository
> will not contain non-normalized paths.

Guess normalisation could be only removed if we would always trust the archive contents.
But considering the misc. security-related scenarios you pointed out, we should never trust.

ThomasWaldmann:
> With paths being normalized as they are being read, this "errors" will be corrected during
> the borg trasfer required when upgrading to Borg 2. Hence, the normalization, when reading
> the archive, can be removed once support for reading v1 repositories is dropped. v2 repository
> will not contain non-normalized paths.

Guess normalisation could be only removed if we would always trust the archive contents.
But considering the misc. security-related scenarios you pointed out, we should never trust.

src/borg/archiver/create_cmd.py
```diff
@@ -661,6 +662,16 @@ def build_parser_create(self, subparsers, common_parser, mid_common_parser):
...
+        class MakePathSafeAction(argparse.Action):
+            def __call__(self, parser, namespace, path, option_string=None):
+                try:
+                    sanitized_path = make_path_safe(path)
+                except ValueError as e:
+                    raise argparse.ArgumentError(self, e)
+                if sanitized_path == ".":
+                    raise argparse.ArgumentError(self, f"{path!r} is not a valid file name")
+                setattr(namespace, self.dest, sanitized_path)
+
```
ThomasWaldmann, line 674: quite some argument parsing helpers are already in `helpers.parseformat`.

src/borg/helpers/msgpack.py
```diff
@@ -236,3 +237,21 @@ def int_to_timestamp(ns):
...
+    return path
+
+
+def to_normalized_path(path):
+    assert isinstance(path, str)
+    # Older version of Borg still allowed non-normalized paths
+    # to be stored. So, we normalize them when reading.
+    #
+    # Also checks for '..' elements in `path` for reasons of security.
+    return make_path_safe(path)
```
ThomasWaldmann, line 257: why is this in `helpers.msgpack`?

dealing with old borg stuff is usually done in archiver.transfer_cmd, borg.upgrade, borg.item.

also, please speak of **legacy** rather than "old", so we can later find and remove the not needed any more stuff more easily.

src/borg/testsuite/archive.py
```diff
@@ -356,3 +357,9 @@ def test_get_item_uid_gid():
     # because item uid/gid seems valid, do not use the given uid/gid defaults
     assert uid == 9
     assert gid == 10
+
+
+def test_reject_non_normalized_item():
+    for path in rejected_dotdot_paths:
+        with pytest.raises(ValueError, match="nexpected '..' element in path"):
```
ThomasWaldmann, line 364: unexpected?

src/borg/testsuite/helpers.py
```diff
@@ -276,16 +277,36 @@ def test_chunkerparams():
...
-        self.assert_equal(make_path_safe("../../foo/bar"), "foo/bar")
-        self.assert_equal(make_path_safe("/"), ".")
-        self.assert_equal(make_path_safe("/"), ".")
+        self.assert_equal(make_path_safe("."), ".")
+        self.assert_equal(make_path_safe("./"), ".")
+        self.assert_equal(make_path_safe("./foo"), "foo")
+        self.assert_equal(make_path_safe(".//foo"), "foo")
+        self.assert_equal(make_path_safe(".//foo//bar//"), "foo/bar")
+        self.assert_equal(make_path_safe("//foo/bar"), "foo/bar")
+        self.assert_equal(make_path_safe("//foo/./bar"), "foo/bar")
```
ThomasWaldmann, line 300: `foo/../bar` ?

ThomasWaldmann (commented):
some comments

src/borg/helpers/fs.py
```diff
@@ -164,12 +164,39 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
-def make_path_safe(path):
-    """Make path safe by making it relative and local"""
-    return _safe_re.sub("", path) or "."
+    For reasons of security, a ValueError is raised should
+    `path` contain any '..' elements.
+    """
+    path = path.lstrip("/")
+    if path.startswith("../") or "/../" in path or path.endswith("/..") or path == "..":
+        raise ValueError(f"unexpected '..' element in path {path!r}")
+    path = os.path.normpath(path)
```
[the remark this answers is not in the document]
ThomasWaldmann: https://github.com/borgbackup/borg/issues/7120 i filed this because we use normpath quite a lot.

as this has this separate issue now, we can ignore it for the scope of this PR.

src/borg/archive.py
```diff
@@ -829,8 +829,6 @@ def extract_item(
 
         original_path = original_path or item.path
         dest = self.cwd
-        if item.path.startswith(("/", "../")):
-            raise Exception("Path should be relative and local")
```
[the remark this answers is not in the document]
pgerber: This is a path that would be rejected and accordingly needs to be added to `rejected_dotdot_paths`. Will do that.

src/borg/testsuite/helpers.py
pgerber: This is a path that would be rejected and accordingly needs to be added to rejected_dotdot_paths. Will do that.

src/borg/testsuite/archive.py
pgerber: Fixed

src/borg/helpers/msgpack.py
```diff
@@ -84,7 +85,7 @@ def __init__(
         use_single_float=False,
         autoreset=True,
         use_bin_type=USE_BIN_TYPE,
-        strict_types=False
+        strict_types=False,
```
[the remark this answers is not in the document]
pgerber: Changed this after GitHub Actions complained, https://github.com/pgerber/borg/actions/runs/3308624366/jobs/5461145547#step:3:32.

> did you install black via pip install -r requirements.d/codestyle.txt?

Not that I remember.

src/borg/helpers/fs.py
```diff
@@ -164,12 +164,39 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+    return path
+
+
+_dotdot_re = re.compile(r"^(\.\./)*")
+
+
+def remove_dotdot_prefixes(path):
+    """
+    Remove leading '../'s at the begging of `path`. Additionally,
```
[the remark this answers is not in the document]
pgerber: Fixed

src/borg/helpers/fs.py
```diff
@@ -164,12 +164,39 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+    `path` contain any '..' elements.
+    """
+    path = path.lstrip("/")
+    if path.startswith("../") or "/../" in path or path.endswith("/..") or path == "..":
+        raise ValueError(f"unexpected '..' element in path {path!r}")
+    path = os.path.normpath(path)
+    return path
+
+
+_dotdot_re = re.compile(r"^(\.\./)*")
```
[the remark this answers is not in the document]
pgerber: Done

src/borg/helpers/fs.py
```diff
@@ -164,12 +164,39 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
     return tag_names
 
 
-_safe_re = re.compile(r"^((\.\.)?/+)+")
+def make_path_safe(path):
+    """
+    Make path safe by making it relative and normalized
```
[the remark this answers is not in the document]
pgerber: Will do

pgerber:
> > With paths being normalized as they are being read, this "errors" will be corrected during
> > the borg trasfer required when upgrading to Borg 2. Hence, the normalization, when reading
> > the archive, can be removed once support for reading v1 repositories is dropped. v2 repository
> > will not contain non-normalized paths.
> 
> Guess normalisation could be only removed if we would always trust the archive contents. But considering the misc. security-related scenarios you pointed out, we should never trust.

I wasn't very clear. I meant normalization could be removed and replaced by strict checking. In other words, we can switch from `to_normalized_path()` to `assert_normalized_path()`.

src/borg/helpers/msgpack.py
pgerber: > why is this in helpers.msgpack?
>
> dealing with old borg stuff is usually done in archiver.transfer_cmd, borg.upgrade, borg.item.

Yes, so I noticed but this is needed in other code paths too, not just when upgrading. I think we always want to reject '..' and absolute paths and doing that as early as possible seemed safest. ~~Also, there appear to be some places that rely on normalization like #7099 and I think it also messes with pattern matching.~~ (On second though, this only concerns Borg 1. In Borg 2, we'll just normalize all paths during transfer.)

As I understand, Borg 2 only reads v1 repos during `borg transfer`. So, I could move normalization to borg.upgrade but then where should I check for '..' and absolute paths? Still in msgpack?

> also, please speak of legacy rather than "old", so we can later find and remove the not needed any more stuff more easily.

Done

pgerber: > As I understand, Borg 2 only reads v1 repos during borg transfer. So, I could move normalization to borg.upgrade but then where should I check for '..' and absolute paths? Still in msgpack?

Actually, there might be a bit of an issue with absolute paths. During `borg transfer --upgrader=From12To20`, I think we just want to convert absolute paths to relative ones. Since --stdin-name allowed `/a/path`, there might be legitimate archives out there with such paths (and legacy version of Borg may have had other bugs allowing absolute paths). However, we want to reject absolute paths in extract, export-tar, mount and possibly other commands. So, we'd probably want to reject absolute paths early and for all commands (even ones not written yet). Handling this for every command separately will make it more likely that the check is missing somewhere. So, I'd advise against it.

However, this leaves us with one question, where could one put a check for absolute paths that will reject such paths for all commands but `borg transfer --upgrader=From12To20`.

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
pgerber: I could move it to a different commit to keep the patch cleaner. Should I?

pgerber:
> @pgerber good catch, thanks for finding/fixing this.

> i agree that there are some potential security issues related to this, but I think they are low severity (compared to tar, zip, etc.) because borg repos/archives usually do not come from unknown/untrusted sources and also borg is not commonly used for data transfer. usually, you either made the repo/archive by yourself or it was made by some fellow admin of yours.
>
> but, as you pointed out, some risks remain.

Agreed. This is also something that isn't covered by the current [Attack Model](https://borgbackup.readthedocs.io/en/latest/internals/security.html#attack-model). I wonder, though, if it should be covered (for Borg 2). Even if a minor issue, having it in there would remind developers not to blindly trust archive/repository content.

(I'm going ignoring the fact there is a `--encryption {none|authenticated}` and the Attack Model should probably mention that such repositories aren't cover by it.)

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
ThomasWaldmann: Hmm, guess I need to check what version it runs on gh actions. Usually it should not complain about code you did not add or modify.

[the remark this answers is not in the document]
ThomasWaldmann: https://github.com/borgbackup/borg/pull/7133/files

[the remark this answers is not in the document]
ThomasWaldmann: ^ guess that will avoid such surprises.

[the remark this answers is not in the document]
ThomasWaldmann: please:
```
# update your local master from github upstream
# in your branch:
git rebase -i master
pip install -r requirements.d/codestyle.txt
black .
```
guess that should get the style into sync.

src/borg/helpers/msgpack.py
```diff
@@ -236,3 +237,21 @@ def int_to_timestamp(ns):
...
     assert isinstance(ts, Timestamp)
     return ts.to_unix_nano()
+
+
+def assert_normalized_path(path):
+    # `path` should have been normalized earlier. Some features,
+    # like pattern matching relies on a normalized path. As a
+    # precaution we check here again.
+    if to_normalized_path(path) != path:
+        raise AssertionError(f"path {path!r} is not normalized")
```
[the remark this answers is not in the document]
ThomasWaldmann: some of the fs-related changes should be moved from `helpers.msgpack` to `helpers.fs`, just below `make_path_safe` there I guess.

src/borg/helpers/msgpack.py
```diff
@@ -46,6 +46,7 @@
 """
 
 from .datastruct import StableDict
+from . import make_path_safe
```
ThomasWaldmann, line 49: you do not use this in msgpack related code, you only imported it because you placed assert_normalized/to_normalized here (but you do not call them from here either), but they should be in `helpers.fs`also.

src/borg/helpers/msgpack.py
ThomasWaldmann: I think --stdin-name names (paths) should be just made relative in borg transfer (12to20), like borg usually also makes all other paths relative when creating an archive.

That it did not do that for --stdin-name is rather a bug than a feature.

Also, we should make sure it is a valid pathname or there could be issues with FUSE fs or with extract.

ThomasWaldmann: Guess the `borg.item.Item.path` attribute is the right place to convert (borg transfer of old stuff) or simply reject (we could do that at same place in borg 2.1 where no transfer of old archive will be supported any more).

pgerber: `borg.item.Item.path` is where I looked first. When I saw that it was defined like this:

```
path = PropDictProperty(str, 'surrogate-escaped str')
```
I figured it would be cleanest to just add a `encode=` and `decode=`.

How would you suggest I handle this instead. I could make `path` an `@property` but then I'd also have to override `__set__` and `__get__` to ensure it's actually called. That seems rather messy though.

I'd appreciate any suggestions as to how to implement this cleanly.

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
ThomasWaldmann: can you do that?

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
ThomasWaldmann: this could be also fixed quite easily, so it is off the table.

src/borg/helpers/parseformat.py
```diff
@@ -741,7 +741,7 @@ class FakeArchive:
 
         from ..item import Item
 
-        fake_item = Item(mode=0, path="", user="", group="", mtime=0, uid=0, gid=0)
+        fake_item = Item(mode=0, path=".", user="", group="", mtime=0, uid=0, gid=0)
```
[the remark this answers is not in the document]
ThomasWaldmann: also easy fix.

src/borg/testsuite/archiver/tar_cmds.py
```diff
@@ -130,6 +130,31 @@ def test_import_tar(self, tar_format="PAX"):
             self.cmd(f"--repo={self.repository_location}", "extract", "dst")
         self.assert_dirs_equal("input", "output/input", ignore_ns=True, ignore_xattrs=True)
 
+    def test_import_unusual_tar(self, tar_format="PAX"):
+
+        # Contains this, unusual entries:
```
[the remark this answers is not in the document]
ThomasWaldmann: easy fix.

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
pgerber: That didn't actually help. Black still complained but since I'm about to move my changes to `helpers.fs`, this shouldn't matter for this pull request.

src/borg/helpers/fs.py
```diff
@@ -164,12 +164,39 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+    Remove leading '../'s at the begging of `path`. Additionally,
+    the path is made relative.
+
+    `path` is expected to be normalized already (e.g. via `os.path.normpath()`).
+    """
+    path = path.lstrip("/")
+    path = _dotdot_re.sub("", path)
+    if path in ["", ".."]:
+        return "."
+    return path
```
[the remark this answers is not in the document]
pgerber: Which kind of paths should be accepted/rejected? I assume we still want to allow users to specify absolute paths and simple relative paths like a/b/c or ./a/b/c. So, are you suggesting to refuse any ../some/path?

Note that the regex starts with ^ and we only remove prefixes from the path.

pgerber:
@ThomasWaldmann, sorry for ignoring some of your comments. I didn't realize GitHub would collapse unresolved threads. Working on resolving those now.

src/borg/testsuite/archiver/tar_cmds.py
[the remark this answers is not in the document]
pgerber: Done

src/borg/archiver/create_cmd.py
pgerber: Moved

src/borg/helpers/parseformat.py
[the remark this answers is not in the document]
pgerber: Done

src/borg/helpers/msgpack.py
```diff
@@ -236,3 +237,21 @@ def int_to_timestamp(ns):
 def timestamp_to_int(ts):
     assert isinstance(ts, Timestamp)
     return ts.to_unix_nano()
+
+
+def assert_normalized_path(path):
+    # `path` should have been normalized earlier. Some features,
+    # like pattern matching relies on a normalized path. As a
```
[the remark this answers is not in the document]
pgerber: Done

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
pgerber: > How about:
> 
> ```
> assert to_normalized_path(path) == path, f"path {path!r} is not normalized"
> ```

I considered it but as I understand asserts aren't triggered when optimizations are enabled and, for security reasons, we probably want this to always trigger. See https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement.

src/borg/helpers/msgpack.py
pgerber: Import removed

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
ThomasWaldmann: OK, maybe use ValueError then?

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
ThomasWaldmann: Maybe `pip install -Ur ...` or so is needed, so it will change the version of black even if it is already installed in some version. Or uninstall / reinstall black.

src/borg/helpers/msgpack.py
ThomasWaldmann: keep the encode/decode for now.

ThomasWaldmann:
OK, when finished with the minor changes, please update the PR.

src/borg/helpers/msgpack.py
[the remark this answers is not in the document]
pgerber: Done

src/borg/helpers/fs.py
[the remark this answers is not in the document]
ThomasWaldmann: Guess we can approach this from 2 perspectives:

a) **what does your PR change compared to the existing code**

for archiving items the new `remove_dotdot_prefixes` is quite close to what `make_path_safe` did, except that it removes leading slashes (which we must do somewhere) and handles the special cases of "nothing left" ("" and ".."), which is also fine (except that a mere "." as archived path is not really useful for extraction)

b) **what do we really want it to be**

usually borg recurses starting from the recursion roots (which we can normalize first) and then only pretty normal paths will be generated by the recursion. for this, we do not need special path processing per item.

when fed with a paths list via stdin (and not using borg's recursor), borg does not have control over what's coming in from there, but borg is also are not required to accept too crappy stuff and still make great sense from it (the admin or tool feeding that list into borg can be expected to provide reasonable paths), borg instead could skip invalid paths with an error msg.

```
>>> normpath("recursion_root/foo/../bar")
'recursion_root/bar'  # definitely valid, it's inside the recursion root

>>> normpath("recursion_root/foo/../../bar")
'bar'  # slightly crappy, but we could archive this correctly (sibling of recursion root)

>>> normpath("recursion_root/foo/../../../bar")
'../bar'  # we can not archive this as is (../ unwanted due to security) and if we strip the ../ we change the meaning.
```

src/borg/helpers/fs.py
```diff
@@ -161,12 +161,57 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+
+
+def to_normalized_path(path):
+    assert isinstance(path, str)
+    # Legacy versions of Borg still allowed non-normalized paths
+    # to be stored. So, we normalize them when reading.
+    #
+    # Also checks for '..' elements in `path` for reasons of security.
+    return make_path_safe(path)
```
ThomasWaldmann, line 214: btw, as we also use `os.path.normpath` (which implements "normalizing a path"), the names here can be a bit confusing.

pgerber: True, I'll try to come up with a different name. (Suggestions welcome)

src/borg/helpers/fs.py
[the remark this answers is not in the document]
pgerber: > when fed with a paths list via stdin (and not using borg's recursor), borg does not have control over what's coming in from there, but borg is also not required to accept too crappy stuff and still make great sense from it (the admin or tool feeding that list into borg can be expected to provide reasonable paths), borg instead could skip invalid paths with an error msg.

I, personally, wouldn't give this too much consideration. Seems unlikely to me that people will end up providing such crappy paths to Borg. Paths with './' or '//', sure, but something like "root/foo/../../../bar" seems rather unlikely to me.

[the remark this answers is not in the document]
pgerber: Handling of '.' we should perhaps be looked at separately. There are some cases where it isn't handled ideally:

```
user@dev-borg:~$ borg list 85
-rw-r--r-- root   root        905 Sun, 2022-04-10 12:08:45  etc/fstab
user@dev-borg:~$ borg create --paths-from-stdin 86 <<<$'/etc/fstab\n'
user@dev-borg:~$ borg list 86
-rw-r--r-- root   root        905 Sun, 2022-04-10 12:08:45  etc/fstab
drwx------ user   user          0 Sun, 2022-11-13 18:19:19  .
```

Empty line at end is interpreted as '.' somehow.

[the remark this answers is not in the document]
ThomasWaldmann: yeah, likely.

about archiving "." directory making sense: yes, i somehow agree, but how would you ever extract that? borg usually expects nothing being in the way, so it usually rmdirs target and then extracts target. not sure whether that would work with target == ".".

pgerber:
> btw, as we also use os.path.normpath (which implements "normalizing a path"), the names here can be a bit confusing.

I moved from "normalize" to "sanitize". This avoids the conflict in the naming and, I believe, also more clearly expresses what's being done.

Let me know if there is anything else that needs to be done for this to get merged.

src/borg/helpers/fs.py
```diff
@@ -161,12 +161,57 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+    return path
+
+
+def to_sanitized_path(path):
+    assert isinstance(path, str)
+    # Legacy versions of Borg still allowed non-sanitized paths
+    # to be stored. So, we sanitize them when reading.
+    #
+    # Also checks for '..' elements in `path` for reasons of security.
+    return make_path_safe(path)
```
ThomasWaldmann, line 214: So, besides the assert, this is the same as make_path_safe. Why do we have these 2 functions?

When looking at make_path_safe, it can not work with bytes objects for the path, so a str is also required there.

Shall it mean that all calls of `to_sanitized_path` can be dropped after dropping borg1 archive reading support?

src/borg/item.pyx
```diff
@@ -261,7 +262,7 @@ cdef class Item(PropDict):
 
     # properties statically defined, so that IDEs can know their names:
 
-    path = PropDictProperty(str, 'surrogate-escaped str')
+    path = PropDictProperty(str, 'surrogate-escaped str', encode=assert_sanitized_path, decode=to_sanitized_path)
```
ThomasWaldmann, line 265: so the decode here is only to support borg1 archives, while the encode better stays forever?

src/borg/testsuite/archive.py
```diff
@@ -356,3 +357,9 @@ def test_get_item_uid_gid():
     # because item uid/gid seems valid, do not use the given uid/gid defaults
     assert uid == 9
     assert gid == 10
+
+
+def test_reject_non_sanitized_item():
+    for path in rejected_dotdot_paths:
+        with pytest.raises(ValueError, match="unexpected '..' element in path"):
+            Item(path=path, uid=0, gid=0, user="root", group="root")
```
ThomasWaldmann, line 365: is it necessary to giver user and group here or is uid and gid enough?

asking because gid 0 on bsd and macOS is group "wheel".

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -478,7 +505,7 @@ def test_exclude_keep_tagged(self):
         )
         self._assert_test_keep_tagged()
 
-    def test_path_normalization(self):
+    def test_path_sanitation(self):
```
ThomasWaldmann, line 508: interestingly, there is also the word "sanitization", but it has a rather specific meaning, so "sanitation" is correct here.

src/borg/testsuite/__init__.py
```diff
@@ -51,6 +52,21 @@
...
+    "/..",
+    "/../",
+    "/../etc",
+    "/../etc/",
+    "etc/..",
+    "/etc/..",
+    "//etc/..",
+    "etc//..",
+    "etc/..//",
+    "foo/../bar",
```
ThomasWaldmann, line 67: maybe one of the above cases already triggers the same check, but guess we should have something like `../etc/passwd` (or shadow) here, just to catch precisely an attack vector.

src/borg/testsuite/archiver/tar_cmds.py
```diff
@@ -130,6 +130,31 @@ def test_import_tar(self, tar_format="PAX"):
...
+        # /foobar
+        # ./bar
+        # ./foo2/
+        # ./foo//bar
+        # ./
+        tar_archive = os.path.join(os.path.dirname(__file__), "unusual_paths.tar")
+
+        self.cmd(f"--repo={self.repository_location}", "rcreate", "--encryption=none")
+        self.cmd(f"--repo={self.repository_location}", "import-tar", "dst", tar_archive)
+        archives = self.cmd(f"--repo={self.repository_location}", "list", "dst", "--format", "{path}{NL}").splitlines()
```
ThomasWaldmann, line 145: s/archives/files/

src/borg/testsuite/archiver/tar_cmds.py
```diff
@@ -130,6 +130,31 @@ def test_import_tar(self, tar_format="PAX"):
...
+        self.assert_equal(set(archives), {"foobar", "bar", "foo2", "foo/bar", "."})
+
+    def test_import_tar_with_dotdot(self, tar_format="PAX"):
```
ThomasWaldmann, line 148: tar_format is not used.

src/borg/testsuite/archiver/tar_cmds.py
```diff
@@ -130,6 +130,31 @@ def test_import_tar(self, tar_format="PAX"):
             self.cmd(f"--repo={self.repository_location}", "extract", "dst")
         self.assert_dirs_equal("input", "output/input", ignore_ns=True, ignore_xattrs=True)
 
+    def test_import_unusual_tar(self, tar_format="PAX"):
```
ThomasWaldmann, line 133: tar_format is not used.

src/borg/testsuite/helpers.py
```diff
@@ -342,16 +343,36 @@ def test_chunkerparams():
...
+class RemoveDotdotPrefixesTestCase(BaseTestCase):
+    def test(self):
+        self.assert_equal(remove_dotdot_prefixes("."), ".")
+        self.assert_equal(remove_dotdot_prefixes(".."), ".")
+        self.assert_equal(remove_dotdot_prefixes("/"), ".")
+        self.assert_equal(remove_dotdot_prefixes("//"), ".")
+        self.assert_equal(remove_dotdot_prefixes("foo"), "foo")
+        self.assert_equal(remove_dotdot_prefixes("foo/bar"), "foo/bar")
+        self.assert_equal(remove_dotdot_prefixes("/foo/bar"), "foo/bar")
+        self.assert_equal(remove_dotdot_prefixes("../foo/bar"), "foo/bar")
```
ThomasWaldmann, line 355: Stripping "./" and "/" is the usual thing borg does.

When listing (or extracting), we'll **know** it either refers to "/" or to "." (which is already ambiguous, but guess it is what tar does and attic/borg copied that behaviour).

If we just strip arbitrary amounts of "../", we basically have no clue any more about the directory level on which it was.

src/borg/testsuite/helpers.py
ThomasWaldmann: ok.

src/borg/testsuite/helpers.py
```diff
@@ -342,16 +343,36 @@ def test_chunkerparams():
...
+
+
 class MakePathSafeTestCase(BaseTestCase):
     def test(self):
-        self.assert_equal(make_path_safe("/foo/bar"), "foo/bar")
```
ThomasWaldmann, line 347: keep this?

ThomasWaldmann (changes requested):
another review.

btw, please make sure that all places that need to be removed or edited when removing borg1 archive reading or other legacy, mention **legacy** in a comment.

src/borg/testsuite/helpers.py
pgerber: Keeping

src/borg/testsuite/archiver/tar_cmds.py
pgerber: Removed

src/borg/testsuite/helpers.py
pgerber: > When listing (or extracting), we'll know it either refers to "/" or to "." (which is already ambiguous, but guess it is what tar does and attic/borg copied that behaviour).

Talking of tar, GNU tar prints a warning when it strips anything from the path. The same could be considered for borg except that we shouldn't complain when a single / at the beginning of a path is removed.

```
user@dev-borg:~/src/borg$ tar cf tmp.tar ../borg-upstream/
tar: Removing leading `../' from member names
```

src/borg/testsuite/archiver/tar_cmds.py
pgerber: Removed

src/borg/testsuite/archiver/tar_cmds.py
pgerber: Renamed

src/borg/testsuite/__init__.py
pgerber: Certainly won't hurt, added.

src/borg/helpers/fs.py
pgerber: Ah, yeah, I did that because once borg1 archives are no longer supported this should be change to an assertion to verify the path aren't malicious. Adding a comment to that regard …

src/borg/item.pyx
pgerber: In borg2 archives, path will be sanitize but we'll still need a check for malicious paths (i.e.. absolute paths and '..'.

pgerber: In borg2 archives, path will be sanitize but we'll still need a check for malicious paths (i.e.. absolute paths and '..'. Adding comment to that regard …

src/borg/testsuite/helpers.py
ThomasWaldmann: Well, if you assume that the cwd at the time of the borg invocation is known (because it is part of the backup script), we know what `.` referred to.

But if borg gets fed all sorts of `../`*N and it removes them all, then we can't reconstruct the location any more. In case of db dumps it might not matter, but in many other cases it might.

BTW, macOS tar:
```
tw@mba2020 ~ % tar cvf foo.tar ../tw/foo.txt ../../etc/hosts
a ../tw/foo.txt
a ../../etc/hosts
tw@mba2020 ~ % tar tvf foo.tar 
-rw-r--r--  0 tw     staff     361 Jul 21  2021 ../tw/foo.txt
-rw-r--r--  0 root   wheel     264 Dec 26 18:59 ../../etc/hosts
```

pgerber:
> Well, if you assume that the cwd at the time of the borg invocation is known (because it is part of the backup script), we know what . referred to.
> 
> But if borg gets fed all sorts of ../*N (e.g. via stdin) and it removes them all, then we can't reconstruct the location any more. In case of db dumps it might not matter, but in many other cases it might.

True, I did not consider the possibility of multiple ../* paths. So, the following combination would be possible:

```
../../db_backup/ ../vm_snapshots/ ../../documents/
```

We'd end up with db_backup, vm_snapshots and documents in the backup and we can tell their source.

```
../a/etc ../b/*
```
Not ideal as b could contain another etc. But everything else we know where it comes from.

```
../a/etc ../b/etc
```

We'd end up with two etc directories and content from both directories being intermixed and some files might even exist twice.

```
../a/* ../b/*
```

All bets are off.

Still, this examples look very contrived to me. One would have to end up being in a CWD where you'd have to use ../ to reach a backup location and decide not to use an absolute path or simply change the CWD. Then, once you settled for using ../, you'd have to use a wildcard or have two files/directory with the same name. Additionally, you'd probably have to neglect looking at the resulting backup (e.g. via `borg list`). This should be easy to spot. And, finally, you'd likely not try if you can restore the backup

I see how one ends up not doing the last item in the list but everything else I find hardly probable. I still think a simple warning would be sufficient for ../. Not sure if we should do anything about the possibility of a file existing twice in the same archive. Playing around with Borg, the fact that `borg extract` silently overwrites existing files or the fact that a permission denied is only considered a warning are far more real concerns, IMO. In fact, I do ignore all warning as I can't silent some innocuous warning that a files content changed while reading.

```
w@mba2020 ~ % tar cvf foo.tar ../tw/foo.txt ../../etc/hosts
a ../tw/foo.txt
a ../../etc/hosts
tw@mba2020 ~ % tar tvf foo.tar 
-rw-r--r--  0 tw     staff     361 Jul 21  2021 ../tw/foo.txt
-rw-r--r--  0 root   wheel     264 Dec 26 18:59 ../../etc/hosts
```

This is interesting I guess this would mean that importing a tar file could break as well when ../ were to be banned. On GNU tar, you need to specifically request this behavior.

src/borg/testsuite/archive.py
pgerber: Yes, user and group is enough. Changed.

ThomasWaldmann:
To work around `black` pinning issue on github actions, this needs to be rebased on current master branch.

pgerber:
Rebased. I had to disable test_create_read_special_symlink() locally though. Looking at that test there is several ways this test can hang indefinitely. Working on fixing it …

ThomasWaldmann:
@pgerber that test also made troubles on windows already, thanks for trying to fix it.

I had to do another update to master btw to get the black stuff right, not sure if this need another rebase here.

ThomasWaldmann:
I did a rebase on current borg master, fixed a merge conflict in the imports and force-pushed.

This should get rid of some test failures (for windows CI, for timestamps, etc.).

I also removed the special char for "..." in the commit comment and replaced it by 3 normal dots.

ThomasWaldmann:
Rebased on current master, squashed commits, no other changes.

ThomasWaldmann:
rebased on master again, no other changes (yet).

src/borg/helpers/fs.py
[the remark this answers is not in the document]
ThomasWaldmann: Hmm, thought more about this, and guess I have to correct myself:
- "losing information" --> yes, we remove "../../../", so we lose that. **But:** that is not super useful as borg does not track the cwd. The admin might know and it might be reconstructable from scripting etc., but it is not much we lose.
- Also, and maybe more important: by using some "../" to point to parent dirs, the resulting paths usually **get longer**, so they gain information. e.g. if cwd is "/etc", if we refer to "passwd" we just have that. if we refer to "../etc/passwd" and then remove "../" from that, we have "etc/passwd", which is **more**.

src/borg/helpers/fs.py
```diff
@@ -217,12 +217,64 @@ def dir_is_tagged(path, exclude_caches, exclude_if_present):
...
+    assert isinstance(path, str)
+    # `path` should have been sanitized earlier. Some features,
+    # like pattern matching rely on a sanitized path. As a
+    # precaution we check here again.
+    if make_path_safe(path) != path:
+        raise ValueError(f"path {path!r} is not sanitized")
+    return path
+
+
+def to_sanitized_path(path):
```
ThomasWaldmann, line 265: hmm, looks like this is not called from anywhere.

ThomasWaldmann (approved):
Let's merge this, took long enough.

In case we find issues in beta testing, they can be addressed later.

PR #7121, opened 2022-11-03
Adding used storage quota to borg info

### LINKED ISSUE #2870: query quota set / use from repository
Is there any way to query the quota introduced in https://github.com/borgbackup/borg/pull/2565? So that I, as a borg client/borg user, check how much size my remote repo already uses and how much space I still have left, so that I can potentially tweak my prune settings, manually delete large archives, etc.

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/47527011-query-quota-from-remote-repository)

francoo98:
Fixes #2870.


src/borg/archiver/info_cmd.py
```diff
@@ -19,6 +19,18 @@ class InfoMixIn:
...
+        if repository.storage_quota:
+            print(
+                """\
+Storage quota: \
+{used} used out of \
+{quota}
+        """.format(
+                    used=format_file_size(repository.get_storage_quota_use()),
+                    quota=format_file_size(repository.storage_quota),
+                )
```
ThomasWaldmann, line 31: you can use f-strings in master branch.

src/borg/repository.py
```diff
@@ -959,6 +959,14 @@ def replay_segments(self, index_transaction_id, segments_transaction_id):
             self.exclusive = remember_exclusive
             self.rollback()
 
+    def get_storage_quota_use(self):
+        storage_quota_use = 0
+        for i, (segment, filename) in enumerate(self.io.segment_iterator()):
+            objects = list(self.io.iter_objects(segment))
+            for tag, key, offset, size, _ in objects:
+                storage_quota_use += header_size(tag) + size
+        return storage_quota_use
```
ThomasWaldmann, line 968: a borg repo already maintains quota set and quota used, please just search in repository.py.

src/borg/testsuite/archiver/info_cmd.py
```diff
@@ -47,6 +49,13 @@ def test_info_json_of_empty_archive(self):
         info_repo = json.loads(self.cmd(f"--repo={self.repository_location}", "info", "--json", "--last=1"))
         assert info_repo["archives"] == []
 
+    def test_info_on_repository_with_storage_quota(self):
+        self.create_regular_file("file1", size=1024 * 80)
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION, "--storage-quota=1G")
+        self.cmd(f"--repo={self.repository_location}", "create", "test", self.input_path)
+        info_archive = self.cmd(f"--repo={self.repository_location}", "info")
+        assert "Storage quota" in info_archive
+
```
ThomasWaldmann, line 58: what used to be `borg info REPO` for borg < 2.0 is now called `borg rinfo REPO`.

ThomasWaldmann (changes requested):
also: please run the test locally before pushing to github.

you can just invoke pytest or tox in your workdir.

ThomasWaldmann:
The quota info should be persisted in the hints file.

src/borg/repository.py
francoo98: I know, but when running rinfo, storage_quota_use is not loaded from disk, so I still need a function.
I'm gonna change the implementation of `get_storage_quota_use` to read from the hints file.

francoo98:
I'm running the tests on my computer. I have a lot of failing tests because RemoteArchiver doesn't have the attr storage_quota.
In the new implementarion of `get_storage_quota` there is an unhandled exception (raised by msgunpack), it is raised when the hints file cannot be read, in this case should I skip the line or should I add an error?
With an error it could like:
```
Security dir: /home/user/.config/borg/security/78e961e6605e242750c7a76edb72688d81deasdfqwerasdf12431234
Storage quota: Error reading hints file.
Original size: 652.18 MB
```

src/borg/archiver/rinfo_cmd.py
```diff
@@ -48,6 +49,8 @@ def do_rinfo(self, args, repository, manifest, cache):
                     location=repository._location.canonical_path(),
                     version=repository.version,
                     append_only=repository.append_only,
+                    used=format_file_size(repository.get_storage_quota_use()),
+                    storage_quota=format_file_size(repository.storage_quota),
```
ThomasWaldmann, line 53: will that work for a remote repository (ssh://...)?

src/borg/repository.py
```diff
@@ -960,12 +960,11 @@ def replay_segments(self, index_transaction_id, segments_transaction_id):
...
 
     def get_storage_quota_use(self):
-        storage_quota_use = 0
-        for i, (segment, filename) in enumerate(self.io.segment_iterator()):
-            objects = list(self.io.iter_objects(segment))
-            for tag, key, offset, size, _ in objects:
-                storage_quota_use += header_size(tag) + size
-        return storage_quota_use
+        hints_path = os.path.join(self.path, "hints.%d" % self.get_transaction_id())
+        integrity_data = self._read_integrity(self.get_transaction_id(), "hints")
```
ThomasWaldmann, line 964: please only call get_transaction_id once.

src/borg/repository.py
```diff
@@ -960,12 +960,11 @@ def replay_segments(self, index_transaction_id, segments_transaction_id):
...
+        with IntegrityCheckedFile(hints_path, write=False, integrity_data=integrity_data) as fd:
+            hints = msgpack.unpack(fd)
+            return hints["storage_quota_use"]
```
ThomasWaldmann, line 967: also, guess this function is way too specific.

there is a lot of useful data in the hints file, not just the storage_quota_use, so I guess we should either:
- always load (or even better: demand load) the hints file and make it available via repository object attributes (like it is already done when starting a transaction), but make sure it is only saved when it needs to be saved.
- make a function _load_hints that returns the full hints object and use it from everywhere where the hints need loading.

src/borg/archiver/info_cmd.py
```diff
@@ -6,7 +6,7 @@
 from ._common import with_repository
 from ..archive import Archive
 from ..constants import *  # NOQA
-from ..helpers import remove_surrogates, format_timedelta, json_print, basic_json_data
+from ..helpers import remove_surrogates, format_timedelta, json_print, basic_json_data, format_file_size
```
ThomasWaldmann, line 9: guess this is not needed any more - do you look at the diffs you produce?

src/borg/archiver/rinfo_cmd.py
```diff
@@ -40,6 +40,7 @@ def do_rinfo(self, args, repository, manifest, cache):
             {encryption}
             Cache: {cache.path}
             Security dir: {security_dir}
+            Storage quota: {used} used out of {storage_quota}
```
ThomasWaldmann, line 43: line 36..40 all refer to repository properties, but 41..42 is about client-side stuff.

so i'ld move 43 to below 40.

src/borg/testsuite/archiver/rinfo_cmd.py
```diff
@@ -36,6 +37,13 @@ def test_info_json(self):
         assert all(isinstance(o, int) for o in stats.values())
         assert all(key in stats for key in ("total_chunks", "total_size", "total_unique_chunks", "unique_size"))
 
+    def test_info_on_repository_with_storage_quota(self):
+        self.create_regular_file("file1", size=1024 * 80)
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION, "--storage-quota=1G")
+        self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
+        info_repo = self.cmd(f"--repo={self.repository_location}", "rinfo")
+        assert "Storage quota" in info_repo
```
ThomasWaldmann, line 45: would be good if this also tested the 2 values (used/total).

ThomasWaldmann (changes requested):

ThomasWaldmann:
btw, there is already a `repository.info` method that could also return quota infos.

src/borg/repository.py
francoo98: I guess I can add a getter to the variables in the hints that checks if the file has been loaded and load it.
Something like this:
```python3
@property
def shadow_index(self):
    if self._shadow_index = {}:
        hints = self._load_hints()
        # Initialize every variable
    return self.shadow_index()
```
This way, the hints file is only loaded when needed, but that method would basically be repeated for all the variables in the hints.

ThomasWaldmann: All related attributes should be `None` until initialised.

francoo98: I can't change the init values of `shadow_index` and `storage_quota_use` to None. I'm gonna use `info` to load the hints file.
So, getting the used quota is done like this: `storage_quota=format_file_size(repository.info().get("storage_quota_use"))`
And the info methods looks like this:
``` python3
def info(self):
        """return some infos about the repo (must be opened first)"""
        self._load_hints()
        return dict(
            id=self.id, version=self.version, append_only=self.append_only, storage_quota_use=self.storage_quota_use
        )
```

src/borg/testsuite/archiver/rinfo_cmd.py
francoo98: I have a problem with this. In the test I added the used quota is different if I execute the test file or if I execute the test on its own.
The difference is in the tens of byte I think, so I made the file in the test bigger to make it work.

ThomasWaldmann: That sounds strange, considering that the test `rcreate`s a new repository in both cases.

Hmm, due to it using new key material, the chunker might behave differently, cutting different chunks, leading to slightly different compressed chunk sizes or different chunk counts.

Try replacing RK_ENCRYPTION with `"--encryption=none"` to avoid this.

francoo98: It also fails without encryption

src/borg/repository.py
```diff
@@ -505,10 +505,13 @@ def _load_hints(self):
...
         """return some infos about the repo (must be opened first)"""
-        self._load_hints()
-        return dict(
-            id=self.id, version=self.version, append_only=self.append_only, storage_quota_use=self.storage_quota_use
-        )
+        info = dict(id=self.id, version=self.version, append_only=self.append_only)
+        try:
+            self._load_hints()
+        except TypeError:
+            pass
```
ThomasWaldmann, line 512: why is that needed?

src/borg/repository.py
```diff
@@ -493,9 +493,25 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
...
+
     def info(self):
         """return some infos about the repo (must be opened first)"""
-        return dict(id=self.id, version=self.version, append_only=self.append_only)
+        info = dict(id=self.id, version=self.version, append_only=self.append_only)
+        try:
+            self._load_hints()
+        except TypeError:
+            pass
+        info["storage_quota_use"] = self.storage_quota_use
```
ThomasWaldmann, line 513: ok, so a remote repository returns its currently used storage quota via the "info" rpc procedure call (see method in RemoteRepository class).

but it does not return the configured (maximum) quota yet.

src/borg/archiver/rinfo_cmd.py
```diff
@@ -48,9 +46,27 @@ def do_rinfo(self, args, repository, manifest, cache):
                     location=repository._location.canonical_path(),
                     version=repository.version,
                     append_only=repository.append_only,
-                    **info,
+                    encryption=info["encryption"],
```
ThomasWaldmann, line 49: guess `**info` would still work?

src/borg/archiver/rinfo_cmd.py
```diff
@@ -48,9 +46,27 @@ def do_rinfo(self, args, repository, manifest, cache):
...
                 )
             )
+
+            if repository.storage_quota:
```
ThomasWaldmann, line 53: i don't think this works for a remote repo.

src/borg/repository.py
```diff
@@ -493,9 +493,25 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
...
 
+    def _load_hints(self):
+        transaction_id = self.get_transaction_id()
+        hints_path = os.path.join(self.path, "hints.%d" % transaction_id)
+        integrity_data = self._read_integrity(transaction_id, "hints")
+        with IntegrityCheckedFile(hints_path, write=False, integrity_data=integrity_data) as fd:
+            hints = msgpack.unpack(fd)
+        self.version = hints["version"]
+        self.storage_quota_use = hints["storage_quota_use"]
+        self.shadow_index = hints["shadow_index"]
```
ThomasWaldmann, line 504: I think you duplicated that code from somewhere else.

Can you make it so it is not duplicated?

src/borg/archiver/rinfo_cmd.py
francoo98: It does, but that unpacks the entire dictionary and I think that's somewhat pointless if I only need one variable

src/borg/repository.py
francoo98: Because when working with a RemoteRepository, _load_hints raises a TypeError

ThomasWaldmann: yeah, I already noted that it can't work that way. but the solution is not to suppress the error (aka "diaper pattern", catches all the shit :-) ), but to use the .info() method instead of an attribute access.

src/borg/archiver/rinfo_cmd.py
ThomasWaldmann: yeah, it is only one - until we pack more into it.

src/borg/repository.py
ThomasWaldmann: (to correct myself: `try: ... except: ...` is the diaper pattern, but the idea is the same)

src/borg/archiver/rinfo_cmd.py
```diff
@@ -50,10 +50,17 @@ def do_rinfo(self, args, repository, manifest, cache):
...
-            if repository.storage_quota:
-                used = format_file_size(repository.info().get("storage_quota_use"))
-                quota = format_file_size(repository.storage_quota)
-                output += f"\nStorage quota: {used} used out of {quota}\n"
+            try:
+                storage_quota = repository.storage_quota
+                used = repository.info().get("storage_quota_use")
+            except AttributeError:
+                # rpc call
+                pass
```
ThomasWaldmann, line 58: no.

how about:
```
repo_info = repository.info()
quota_used = repo_info[...]
quota_total = repo_info[...]
```

src/borg/repository.py
```diff
@@ -494,11 +494,7 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
...
     def _load_hints(self):
-        transaction_id = self.get_transaction_id()
-        hints_path = os.path.join(self.path, "hints.%d" % transaction_id)
-        integrity_data = self._read_integrity(transaction_id, "hints")
-        with IntegrityCheckedFile(hints_path, write=False, integrity_data=integrity_data) as fd:
-            hints = msgpack.unpack(fd)
+        hints = self._unpack_hints(self.get_transaction_id())
         self.version = hints["version"]
         self.storage_quota_use = hints["storage_quota_use"]
         self.shadow_index = hints["shadow_index"]
```
ThomasWaldmann, line 500: should that be like in line 638+:
```
                assert hints["version"] == 2
                self.segments = hints["segments"]
                self.compact = FreeSpace(hints["compact"])
                self.storage_quota_use = hints.get("storage_quota_use", 0)
                self.shadow_index = hints.get("shadow_index", {})
```

src/borg/archiver/rinfo_cmd.py
```diff
@@ -48,9 +46,31 @@ def do_rinfo(self, args, repository, manifest, cache):
                     location=repository._location.canonical_path(),
                     version=repository.version,
                     append_only=repository.append_only,
-                    **info,
+                    encryption=info["encryption"],
                 )
             )
+
+            response = repository.info()
+            storage_quota = response["storage_quota"] or None
```
ThomasWaldmann, line 54: So you intended to use None instead of 0, but you do not really make a difference in the code below.

ThomasWaldmann: Also, it would be interesting how much space a repo uses, even if there is not a set limit via storage_quota.

src/borg/repository.py
ThomasWaldmann: guess this method is not called any more?

src/borg/testsuite/archiver/rinfo_cmd.py
```diff
@@ -19,6 +20,7 @@ def test_info(self):
         self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
         info_repo = self.cmd(f"--repo={self.repository_location}", "rinfo")
         assert "Original size:" in info_repo
+        assert "Storage quota:" not in info_repo
```
ThomasWaldmann, line 23: maybe do not address the quota stuff within this test, but rather add a similar one as the other, just without a set quota.

src/borg/testsuite/archiver/rinfo_cmd.py
```diff
@@ -36,6 +38,13 @@ def test_info_json(self):
         assert all(isinstance(o, int) for o in stats.values())
         assert all(key in stats for key in ("total_chunks", "total_size", "total_unique_chunks", "unique_size"))
 
+    def test_info_on_repository_with_storage_quota(self):
+        self.create_regular_file("file1", contents=randbytes(1000 * 1000))
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION, "--storage-quota=1G")
+        self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
+        info_repo = self.cmd(f"--repo={self.repository_location}", "rinfo")
+        assert "Storage quota: 1.00 MB used out of 1.00 GB" in info_repo
+
```
ThomasWaldmann, line 47: in the end, you'll need 2 tests: one **with** a quota and one **without**.

ThomasWaldmann (changes requested):

src/borg/repository.py
francoo98: It isn't, I forgot it was there. I'm gonna remove it soon.

src/borg/repository.py
francoo98: I can't use .info() method there because that is info().

ThomasWaldmann: the code in repository.py always uses a local repository (relative from where that code is executed).

in case of a remote repo, the code in remote.py runs on the client and the code in repository.py runs on the repo server.

so, please remove the exception handler and analyse the root cause of whatever issue you hid by it.

src/borg/testsuite/archiver/rinfo_cmd.py
```diff
@@ -45,6 +44,13 @@ def test_info_on_repository_with_storage_quota(self):
         info_repo = self.cmd(f"--repo={self.repository_location}", "rinfo")
         assert "Storage quota: 1.00 MB used out of 1.00 GB" in info_repo
 
+    def test_info_on_repository_without_storage_quota(self):
+        self.create_regular_file("file1", size=1024 * 80)
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
+        info_repo = self.cmd(f"--repo={self.repository_location}", "rinfo")
+        assert "Storage quota:" not in info_repo
```
ThomasWaldmann, line 52: in the end, that should be something like "123 GB used out of unlimited total."

src/borg/archiver/rinfo_cmd.py
```diff
@@ -48,9 +46,31 @@ def do_rinfo(self, args, repository, manifest, cache):
...
-                    **info,
+                    encryption=info["encryption"],
                 )
             )
+
+            response = repository.info()
+            storage_quota = response["storage_quota"]
+            used = response["storage_quota_use"]
+
+            if storage_quota:
```
ThomasWaldmann, line 57: just remove the "if" so it always is output.

of course you need some code to pretty output `storage_quota` for the "unlimited" case.

ThomasWaldmann:
are you still working on this? some stuff still to do, see my comments.

francoo98:
I've been busy the last two weeks. I'm gonna be able to resume next Monday I think.

francoo98:
I can't find the cause of the error when working with a remote repository.
Reading the code I found out that (with remotes repos) the `info` method cannot call `get_transaction_id`, it returns None. I tested this in the master branch too.
Do you know why this happens?

ThomasWaldmann:
Hmm, does the problem with remote repos only happen when testing manually? The tests on github CI do not show a problem.

francoo98:
The tests were passing because of the diaper pattern. I removed it.

ThomasWaldmann:
It would fail for all python versions. That you do not see failure for all is just because the `fail-fast` setting is active on the github CI, so it terminates all jobs after the first one detected a failure.

ThomasWaldmann:
```
TypeError('%d format: a real number is required, not NoneType')
repository.py:570 in _unpack_hints
    hints_path = os.path.join(self.path, "hints.%d" % transaction_id)
```

francoo98:
That was the case, there was no hints file. But this only happens with remote repos, when creating a local one the files hints, integrity, etc are created.

src/borg/repository.py
```diff
@@ -502,8 +502,12 @@ def _load_hints(self):
     def info(self):
         """return some infos about the repo (must be opened first)"""
         info = dict(id=self.id, version=self.version, append_only=self.append_only)
-        self._load_hints()
-        info["storage_quota_use"] = self.storage_quota_use
+        try:
+            self._load_hints()
+            info["storage_quota_use"] = self.storage_quota_use
+        except TypeError:
```
ThomasWaldmann, line 508: "the return of the diaper pattern" :-)

francoo98:
Is there a reason why you haven't seen this?

src/borg/repository.py
```diff
@@ -493,9 +493,23 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
         self.id = unhexlify(self.config.get("repository", "id").strip())
         self.io = LoggedIO(self.path, self.max_segment_size, self.segments_per_dir)
 
+    def _load_hints(self):
+        if not (transaction_id := self.get_transaction_id()):
+            # self is a fresh repo, so transaction_id is None and there is no hints file
```
ThomasWaldmann, line 498: maybe it works like this (not sure if we would ever get a transaction_id of 0), but still, if you want to check for just `None`, please use `... is None` and not `not ...`.

src/borg/repository.py
```diff
@@ -493,9 +493,23 @@ def open(self, path, exclusive, lock_wait=None, lock=True):
...
+            self.storage_quota_use = 0
```
ThomasWaldmann, line 499: from `Repository.__init__`:
```
        self.storage_quota = storage_quota  # from param, defaults to None
        self.storage_quota_use = 0
```
So, looks like it is 0 anyway as long as nothing wrote a different value there.

src/borg/testsuite/archiver/rinfo_cmd.py
ThomasWaldmann: So, this is kind of resolved using the bigger test file now?

ThomasWaldmann:
> Is there a reason why you haven't seen this?

I've seen it, just needed some time to have a deeper look at the code.

Overall, it looks pretty good now, just 2 minor changes.

After that I guess it can be squashed together and merged.

ThomasWaldmann (approved):
LGTM

PR #7123, opened 2022-11-04
Fix test_size_on_disk_accurate on Windows

RayyanAnsari:
Pass the file descriptor instead of the file path


ThomasWaldmann:
Thanks!

PR #7171, opened 2022-11-27
Fix archiver tests on Windows

### LINKED ISSUE #7172: windows: fix archiver tests
master branch: quite some archiver tests are broken on native windows.

bounty: https://app.bountysource.com/issues/114104975-windows-fix-archiver-tests

RayyanAnsari:
`pytest -k "archiver and not binary and not remote"`
Before: `44 failed, 131 passed, 29 skipped, 1378 deselected in 74.48s (0:01:14)`
After: `173 passed, 33 skipped, 1382 deselected in 107.46s (0:01:47)`

Fixes #7172

### DIFF
--- a/src/borg/archiver/create_cmd.py
+++ b/src/borg/archiver/create_cmd.py
@@ -28,6 +28,7 @@ from ..helpers import sig_int, ignore_sigint
 from ..helpers import iter_separated
 from ..manifest import Manifest
 from ..patterns import PatternMatcher
+from ..platform import is_win32
 from ..platform import get_flags
 from ..platform import uid2user, gid2group
 
@@ -68,7 +69,9 @@ class CreateMixIn:
                 if not dry_run:
                     try:
                         try:
-                            proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE, preexec_fn=ignore_sigint)
+                            proc = subprocess.Popen(
+                                args.paths, stdout=subprocess.PIPE, preexec_fn=None if is_win32 else ignore_sigint
+                            )
                         except (FileNotFoundError, PermissionError) as e:
                             self.print_error("Failed to execute command: %s", e)
                             return self.exit_code
@@ -89,7 +92,9 @@ class CreateMixIn:
                 paths_sep = eval_escapes(args.paths_delimiter) if args.paths_delimiter is not None else "\n"
                 if args.paths_from_command:
                     try:
-                        proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE, preexec_fn=ignore_sigint)
+                        proc = subprocess.Popen(
+                            args.paths, stdout=subprocess.PIPE, preexec_fn=None if is_win32 else ignore_sigint
+                        )
                     except (FileNotFoundError, PermissionError) as e:
                         self.print_error("Failed to execute command: %s", e)
                         return self.exit_code
--- a/src/borg/archiver/key_cmds.py
+++ b/src/borg/archiver/key_cmds.py
@@ -101,6 +101,8 @@ class KeysMixIn:
             manager.export_paperkey(args.path)
         else:
             try:
+                if os.path.isdir(args.path):
+                    raise IsADirectoryError
                 if args.qr:
                     manager.export_qr(args.path)
                 else:
--- a/src/borg/platform/windows.pyx
+++ b/src/borg/platform/windows.pyx
@@ -16,22 +16,22 @@ cdef extern from 'windows.h':
 
 @lru_cache(maxsize=None)
 def uid2user(uid, default=None):
-    return default
+    return "root"
 
 
 @lru_cache(maxsize=None)
 def user2uid(user, default=None):
-    return default
+    return 0
 
 
 @lru_cache(maxsize=None)
 def gid2group(gid, default=None):
-    return default
+    return "root"
 
 
 @lru_cache(maxsize=None)
 def group2gid(group, default=None):
-    return default
+    return 0
 
 
 def getosusername():
--- a/src/borg/testsuite/archiver/__init__.py
+++ b/src/borg/testsuite/archiver/__init__.py
@@ -28,6 +28,7 @@ from ...repository import Repository
 from .. import has_lchflags
 from .. import BaseTestCase, changedir, environment_variable
 from .. import are_symlinks_supported, are_hardlinks_supported, are_fifos_supported
+from ..platform import is_win32
 
 RK_ENCRYPTION = "--encryption=repokey-aes-ocb"
 KF_ENCRYPTION = "--encryption=keyfile-chacha20-poly1305"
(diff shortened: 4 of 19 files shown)


src/borg/platform/windows.pyx
```diff
@@ -16,22 +16,22 @@ cdef extern from 'windows.h':
...
+    return "root"
 
 
 @lru_cache(maxsize=None)
 def group2gid(group, default=None):
-    return default
+    return 0
 
 
 def getosusername():
```
ThomasWaldmann, line 37: did you check where this is used and whether it needs faking also?

src/borg/testsuite/archiver/__init__.py
```diff
@@ -23,6 +23,7 @@
 from ...helpers import bin_to_hex
 from ...manifest import Manifest
 from ...logger import setup_logging
+from ..platform import is_win32
```
ThomasWaldmann, line 26: move this to the other `..` imports.

src/borg/archiver/create_cmd.py
```diff
@@ -90,7 +91,7 @@ def create_inner(archive, cache, fso):
                 paths_sep = eval_escapes(args.paths_delimiter) if args.paths_delimiter is not None else "\n"
                 if args.paths_from_command:
                     try:
-                        proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE, preexec_fn=ignore_sigint)
+                        proc = subprocess.Popen(args.paths, stdout=subprocess.PIPE, preexec_fn=None if is_win32 else ignore_sigint)
```
ThomasWaldmann, line 94: do we have some more of these in the tar related commands?

maybe do a global search for `subprocess.Popen`.

ThomasWaldmann:
also, `black` is unhappy. best is to locally run `black .` before each commit.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -500,6 +501,7 @@ def test_repeated_files(self):
         self.cmd(f"--repo={self.repository_location}", "create", "test", "input", "input")
 
     @pytest.mark.skipif("BORG_TESTS_IGNORE_MODES" in os.environ, reason="modes unreliable")
+    @pytest.mark.skipif(is_win32, reason="modes unavailable on Windows")
```
ThomasWaldmann, line 504: guess on win32, we'ld have to look at `st.st_file_attributes` instead of `.st_mode`.

is `.st_mode` completely unused on win32?

src/borg/testsuite/archiver/rlist_cmd.py
```diff
@@ -34,12 +35,12 @@ def test_archives_format(self):
...
         output_1 = self.cmd(f"--repo={self.repository_location}", "rlist", "--short")
-        self.assertEqual(output_1, "test-1\ntest-2\n")
+        self.assertEqual(output_1, "test-1" + os.linesep + "test-2" + os.linesep)
         output_1 = self.cmd(f"--repo={self.repository_location}", "rlist", "--format", "{barchive}/")
         self.assertEqual(output_1, "test-1/test-2/")
         output_3 = self.cmd(f"--repo={self.repository_location}", "rlist", "--format", "{name} {comment}{NL}")
-        self.assert_in("test-1 comment 1\n", output_3)
-        self.assert_in("test-2 comment 2\n", output_3)
+        self.assert_in("test-1 comment 1" + os.linesep, output_3)
+        self.assert_in("test-2 comment 2" + os.linesep, output_3)
```
ThomasWaldmann, line 43: instead of repeating `os.linesep` all over the place, I guess we could consider putting something like this into `borg.constants` from where we usually star-import everything:

```
LS = os.linesep
```

And then use `LS` (or `NL` like "newline") everywhere.

src/borg/platform/windows.pyx
```diff
@@ -16,22 +16,22 @@ cdef extern from 'windows.h':
...
 @lru_cache(maxsize=None)
 def gid2group(gid, default=None):
-    return default
+    return "root"
 
 
 @lru_cache(maxsize=None)
 def group2gid(group, default=None):
-    return default
+    return 0
```
ThomasWaldmann, line 34: guess there should be an open ticket about this temporary hack, which could also be linked from here.

src/borg/testsuite/archiver/create_cmd.py
ThomasWaldmann: i checked: st_mode somehow "works" on win32, too:

usually, files are -rw-rw-rw and dirs are drwxrwxrwx.

so guess the **reason** should say "umask unavailable on Windows".

src/borg/archiver/key_cmds.py
```diff
@@ -104,6 +104,8 @@ def do_key_export(self, args, repository):
                 if args.qr:
                     manager.export_qr(args.path)
                 else:
+                    if os.path.isdir(args.path):
+                        raise IsADirectoryError
```
ThomasWaldmann, line 108: how about line 105, does it have the same issue?

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -243,9 +243,19 @@ def test_create_paths_from_command(self):
         self.create_regular_file("file4", size=1024 * 80)
 
         input_data = "input/file1\ninput/file2\ninput/file3"
-        self.cmd(
-            f"--repo={self.repository_location}", "create", "--paths-from-command", "test", "--", "echo", input_data
-        )
+        if is_win32:
+            with open("filenames.bat", "w") as script:
```
ThomasWaldmann, line 247: .bat works, but is somehow ancient as it stems from DOS.

guess nowadays we rather have .cmd ?

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -243,9 +243,19 @@ def test_create_paths_from_command(self):
...
+                script.write("@echo off\n")
+                for filename in input_data.splitlines():
+                    script.write(f"echo {filename}\n")
```
ThomasWaldmann, line 250: guess you could also use `@echo ...` here and omit line 248.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -243,9 +243,19 @@ def test_create_paths_from_command(self):
...
+            self.cmd(
+                f"--repo={self.repository_location}", "create", "--paths-from-command", "test", "--", "filenames.bat"
+            )
+        else:
+            self.cmd(
+                f"--repo={self.repository_location}", "create", "--paths-from-command", "test", "--", "echo", input_data
+            )
+
```
ThomasWaldmann, line 258: remove filenames.* again - maybe using a try/finally construction?

src/borg/testsuite/archiver/rlist_cmd.py
ThomasWaldmann: can you do a global search whether that is the only place where `{NL}` is used?

src/borg/testsuite/archiver/create_cmd.py
puetzk: st_mode is not real on NTFS (or the windows kernel in general), but the urct (C runtime library) does kind of emulate it based on FILE_ATTRIBUTE_DIRECTORY, FILE_ATTRIBUTE_READONLY, and checking the filename for ".exe", ".cmd", ".bat", and ".com".

Which produces vaguely sensible results

src/borg/archiver/key_cmds.py
ThomasWaldmann: did you see my comment?

src/borg/testsuite/archiver/extract_cmd.py
```diff
@@ -309,12 +309,12 @@ def test_extract_include_exclude_regex_from_file(self):
...
         self.create_regular_file("file4", size=1024 * 80)
         self.create_regular_file("file333", size=1024 * 80)
-        self.create_regular_file("aa:something", size=1024 * 80)
+        self.create_regular_file("aa-something", size=1024 * 80)
 
         # Create while excluding using mixed pattern styles
         with open(self.exclude_file_path, "wb") as fd:
             fd.write(b"re:input/file4$\n")
-            fd.write(b"fm:*aa:*thing\n")
+            fd.write(b"fm:*aa-*thing\n")
```
ThomasWaldmann, line 317: we already talked about this one on irc:

that aa:something filename was just used to test pattern prefix vs. filename confusion and that only can be tested with a 2-letter plus colon prefix. as the colon is invalid on windows, you just removed it. but that makes the whole test pointless, so i'ld suggest removing that file and the match on it.

instead, for testing a fm: pattern, it could also match `fm:file3*` and check if it matches the correct 2 files for that.

src/borg/testsuite/archiver/rlist_cmd.py
ThomasWaldmann: seen my comment?

for consistent, guess it should be `NL`.

src/borg/platform/windows.pyx
ThomasWaldmann: https://github.com/borgbackup/borg/issues/7192

src/borg/testsuite/archiver/rlist_cmd.py
RayyanAnsari: > can you do a global search whether that is the only place where `{NL}` is used?

I've sent some messages regarding this on IRC.

ThomasWaldmann: answered there. also I'll have a look at the code right now myself...

ThomasWaldmann:
Please rebase your PR code onto current master branch to fix the macOS tox issue.

src/borg/testsuite/archiver/transfer_cmd.py
```diff
@@ -249,7 +255,8 @@ def local_to_utc(local_naive, tzoffset, tzinfo):
                         assert item.group in ("root", "wheel")
                         assert "hlid" not in item
                     elif item.path.endswith("bdev_12_34"):
-                        assert stat.S_ISBLK(item.mode)
+                        if not is_win32:
+                            assert stat.S_ISBLK(item.mode)
```
ThomasWaldmann, line 259: just wondering: the bdev needed fixing for windows, but the cdev (line 248) didn't?

ThomasWaldmann:
about the conflict: the barchive placeholder does not exist any more.

src/borg/testsuite/archiver/transfer_cmd.py
RayyanAnsari: https://github.com/borgbackup/borg/pull/7171/commits/cf61e7d58ae3b3e2fcdc7d9261a23833ac6da7a6

ThomasWaldmann: ah, ok.

ThomasWaldmann:
Thanks for working on this!

PR #6149, opened 2022-01-22
remove endianness macro, fixes #6105

ThomasWaldmann:
thanks to everybody who helped with this, esp. @th0ma7 and @methane.

PR #7181, opened 2022-12-02
Fix msgpack runtime on big-endian OpenBSD/sparc64

klemensn:
- Fallback to compiler defines when __BYTE_ORDER is not available
- simplify the endianity handling
- additional cleanup
- Fix build error caused by ntohs, ntohl

Backport https://github.com/msgpack/msgpack-python/pull/513 (as discussed in
https://github.com/borgbackup/borg/pull/6149#issuecomment-1335745572) and https://github.com/msgpack/msgpack-python/pull/514.

Applied with
```
$ gh -R msgpack/msgpack-python pr diff --patch 513 |
	git am --directory src/borg/algorithms/ -
$ gh -R msgpack/msgpack-python pr diff --patch 514 |
	git am --directory src/borg/algorithms/ -
```

Tested on OpenBSD/sparc64 (big-endian) and OpenBSD/amd64 (little-endian).

### DIFF
--- a/src/borg/algorithms/msgpack/_version.py
+++ b/src/borg/algorithms/msgpack/_version.py
@@ -2,5 +2,9 @@
 # Changes:
 # +borg1: drop support for old buffer protocol to be compatible with py310
 #         (backport of commit 9ae43709e42092c7f6a4e990d696d9005fa1623d)
-version = (0, 5, 6, '+borg1')
+# +borg2: Usef __BYTE_ORDER__ instead of __BYTE_ORDER (#513)
+#         (backport of commit 9d45926a596028e39ec59dd909a56eb5e9e8fee7)
+#         Fix build error caused by ntohs, ntohl (#514)
+#         (backport of commit edca770071fc702e0b4c33f87fb0fa3682b486b4)
+version = (0, 5, 6, '+borg2')
 
--- a/src/borg/algorithms/msgpack/sysdep.h
+++ b/src/borg/algorithms/msgpack/sysdep.h
@@ -61,8 +61,6 @@ typedef unsigned int _msgpack_atomic_counter_t;
 #endif
 #endif
 
-#else
-#include <arpa/inet.h>  /* __BYTE_ORDER */
 #endif
 
 #if !defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)
--- a/src/borg/algorithms/msgpack/sysdep.h
+++ b/src/borg/algorithms/msgpack/sysdep.h
@@ -61,6 +61,8 @@ typedef unsigned int _msgpack_atomic_counter_t;
 #endif
 #endif
 
+#else /* _WIN32 */
+#include <arpa/inet.h> /* ntohs, ntohl */
 #endif
 
 #if !defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
@klemensn Hmm, the CI is unhappy, `ntohs` and `ntohl` are not declared?

klemensn:
> @klemensn Hmm, the CI is unhappy, `ntohs` and `ntohl` are not declared?
> 
> https://linux.die.net/man/3/ntohl man page says it needs `arpa/inet.h`.

Indeed.  I guess this header is pulled in by something else already on OpenBSD.

klemensn:
Other big-endian OpenBSD architectures such as [macppc](https://openbsd.org/macppc.html) and [octeon](https://openbsd.org/octeon.html) are NOT effected by this, hence the updated PR title.

This is most likely due to the fact that octeon and macppc use [clang 13.0.0 from base](https://man.openbsd.org/clang-local.1) while sparc64 still uses [GCC 8.4.0 from ports](https://github.com/openbsd/ports/blob/5c472e4a1e0310506f2ec59f0af2e713e5406d7d/lang/gcc/8/Makefile) to build borgbackup and msgpack-python packages.

ThomasWaldmann:
Ah, CI is happy again. So I guess msgpack project should also revert that change.

klemensn:
> Ah, CI is happy again. So I guess msgpack project should also revert that change.

I've pointed that out already in their PR.
Either way, this wouldn't effect borg 1.1.x, I think, so this backport should be good to go now.

ThomasWaldmann:
https://github.com/borgbackup/borg/blob/e79986b2835291431f04683423304a02b039c0c1/src/borg/algorithms/msgpack/_version.py

Can you increase that (...borg1 -> ...borg2) and add a small description there?

Maybe first wait for feedback from msgpack project though.

src/borg/algorithms/msgpack/_version.py
```diff
@@ -2,5 +2,7 @@
 # Changes:
 # +borg1: drop support for old buffer protocol to be compatible with py310
 #         (backport of commit 9ae43709e42092c7f6a4e990d696d9005fa1623d)
-version = (0, 5, 6, '+borg1')
+# +borg2: fix runtime on big-endian OpenBSD/sparc64
+#         (revert of commit 9c514f0124abf03eef79e474299e2e3cf6381ea1)
```
ThomasWaldmann, line 6: as we do not include master here, but 0.5.6, it would be better to point to the changes relative to 0.5.6.
so just give the changesets applied and skip the reverted one.

klemensn:
> Ah, CI is happy again. So I guess msgpack project should also revert that change.

They just did through https://github.com/msgpack/msgpack-python/pull/514.
I'll replace my revert with their commit and rewrap the PR.

src/borg/algorithms/msgpack/_version.py
klemensn: Sure, will do.

ThomasWaldmann:
OK, looks like this is finished (and CI is happy)?

klemensn:
> OK, looks like this is finished (and CI is happy)?

Yup, pretty sure that's it.

ThomasWaldmann:
Thanks for fixing this! (such platform stuff is hard to test)

klemensn:
> Thanks for fixing this! (such platform stuff is hard to test)

Would you be able to add new platforms to your CI?
Big or little endian, LLVM/Clang or GCC, whatever new combination you can test would increase coverage.
I have no idea how GitHub works in this regard.

ThomasWaldmann:
IIRC github actions only has intel/amd CPUs (and linux, macos and windows as OS).

So guess it would need quite some work to do BE testing there, like using kvm to emulate a BE CPU, having some appropriate BE-supporting OS in that VM, ...

Could be also quite slow. For bootstrapping that VM as well as for the borg pytest test run.

PR #7197, opened 2022-12-10
text attributes: validate more strictly, fixes #2290

### LINKED ISSUE #2290: Character set of archive names
Currently archive names are very permissive, they can even be binary. This is problematic for operating systems who don't like binary file names (e.g. Windows, perhaps OSX to some extent?).

Specifically, Windows does dislike these: `<>:"/\|?*`, applications get easily confused if you have these in a file/dir name (you can try this, since ntfs-3g allows these by default). Control characters (ordinals 1 through 31) are not well received as well. There are certain words that may cause trouble for some applications. See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx for reference.

ThomasWaldmann:
- archive names
- archive comments

also: kill the `b*` (bpath, barchive, bcomment) stuff.
- barchive and bcomment is not needed any more (validated plain text, nothing binary / strange inside)
- bpath will need to get solved otherwise later, maybe as base64-* element (for json, which allows only pure text)

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -616,7 +616,7 @@ class BaseFormatter:
         return (
             "- NEWLINE: OS dependent line separator\n"
             "- NL: alias of NEWLINE\n"
-            "- NUL: NUL character for creating print0 / xargs -0 like output, see bpath key below\n"
+            "- NUL: NUL character for creating print0 / xargs -0 like output\n"
             "- SPACE\n"
             "- TAB\n"
             "- CR\n"
@@ -753,8 +753,7 @@ class ItemFormatter(BaseFormatter):
     # shake_* is not provided because it uses an incompatible .digest() method to support variable length.
     hash_algorithms = set(hashlib.algorithms_guaranteed).union({"xxh64"}).difference({"shake_128", "shake_256"})
     KEY_DESCRIPTIONS = {
-        "bpath": "verbatim POSIX path, can contain any character except NUL",
-        "path": "path interpreted as text (might be missing non-text characters, see bpath)",
+        "path": "file path",
         "source": "link target for symlinks (identical to linktarget)",
         "hlid": "hard link identity (same if hardlinking same fs object)",
         "extra": 'prepends {source} with " -> " for soft links and " link to " for hard links',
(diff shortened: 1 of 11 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -540,9 +540,30 @@ def validator(text):
 
 def archivename_validator():
     def validator(text):
+        assert isinstance(text, str)
+        # we make sure that the archive name can be used as directory name (for borg mount)
         text = replace_placeholders(text)
-        if "/" in text or "::" in text or not text:
-            raise argparse.ArgumentTypeError('Invalid archive name: "%s"' % text)
+        if not (0 < len(text) <= 250):  # max path length on windows is rather short
```
RayyanAnsari, line 546: On Windows 10, the max path limit can be removed by adding a registry key - in fact the Python on Windows installer does this.
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry
Something like this may help for checking whether this is enabled: https://stackoverflow.com/a/46546731

ThomasWaldmann: Ah, that's interesting, thanks!

OTOH, it we use the pyinstaller made binary, most windows installations would still have the low limit, right?

src/borg/helpers/parseformat.py
```diff
@@ -540,9 +540,34 @@ def validator(text):
...
+        # we make sure that the archive name can be used as directory name (for borg mount)
         text = replace_placeholders(text)
-        if "/" in text or "::" in text or not text:
-            raise argparse.ArgumentTypeError('Invalid archive name: "%s"' % text)
+        MAX_PATH = 260  # Windows default. Since Win10, there is a registry setting LongPathsEnabled to get more.
+        MAX_DIRNAME = MAX_PATH - len("12345678.123")
+        SAFETY_MARGIN = 48  # borgfs path: mountpoint / archivename / dir / dir / ... / file
+        MAX_ARCHIVENAME = MAX_DIRNAME - SAFETY_MARGIN
+        if not (0 < len(text) <= MAX_ARCHIVENAME):
+            raise argparse.ArgumentTypeError(f'Invalid archive name: "{text}" [0 < length <= {MAX_ARCHIVENAME}]')
```
ThomasWaldmann, line 551: Since we can not rely on LongPathsEnabled and we also need some (48) chars for the mountpoint and path components inside the `archivename` directory, we are now down to 200 chars max for the archivename.

src/borg/helpers/parseformat.py
ThomasWaldmann: see comment in updated PR.

Issue #6606, opened 2022-04-16
--iec parameter ignored in some places


jernejs:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Issue

#### Your borg version (borg -V).

1.2.0

#### Operating system (distribution) and version.

Gentoo Linux amd64

#### Full borg commandline that lead to the problem (leave away excludes and passwords)
````
export BORG_REPO=/mnt/storage/borg
borg --iec create --progress -v -s --compression=zstd ::deepthought-SSD-2022-04-16_15:02:22 / --exclude-from /root/scripts/backup/backup.exclude

borg --iec prune --progress --prefix=deepthought-SSD- -v -s --list -H 24 -d 15 -w 12 -m 24 -y -1
````
## Describe the problem you're observing.

When creating the archive, storage quota is reported in decimal units despite the `--iec` flag. `--stats` are reported in binary units:
````
Creating archive at "/mnt/storage/borg::deepthought-SSD-2022-04-16_15:02:22"
Storage quota: 193.62 GB out of 17.65 TB used.DRASIL/ender/.wine/drive_c/windows/system32/slc.dll
Storage quota: 193.77 GB out of 17.65 TB used.
------------------------------------------------------------------------------
Repository: /mnt/storage/borg
Archive name: deepthought-SSD-2022-04-16_15:02:22
Archive fingerprint: 89f4b63f7deefee5245b190fed80f9ebf2a9009a4a069e5a106f11000be11bb0
Time (start): Sat, 2022-04-16 15:02:23
Time (end):   Sat, 2022-04-16 15:08:45
Duration: 6 minutes 22.57 seconds
Number of files: 953378
Utilization of max. archive size: 0%
------------------------------------------------------------------------------
                       Original size      Compressed size    Deduplicated size
This archive:              72.63 GiB            41.78 GiB           139.09 MiB
All archives:             558.34 GiB           490.62 GiB           180.43 GiB

                       Unique chunks         Total chunks
Chunk index:                  830602              2346427
------------------------------------------------------------------------------
````
When pruning the archives, --stats are reported in decimal units (again, despite having used the --iec flag):
````
Keeping archive (rule: hourly #1):       deepthought-SSD-2022-04-16_01:27:21 Sat, 2022-04-16 01:27:21 [7fc4f37abc15ac65df9d7e98429631b2e36c18d3d543815dd6bec8ca8130bcad]
[...]

PR #7237, opened 2022-12-31
Fix query quota

francoo98:
Added the `--iec` option and also a \n that I forgot to add in my previous pr.
Issue: #6606
Previous pr: #7121

### DIFF
--- a/src/borg/archiver/rinfo_cmd.py
+++ b/src/borg/archiver/rinfo_cmd.py
@@ -54,7 +54,7 @@ class RInfoMixIn:
             storage_quota = response["storage_quota"]
             used = format_file_size(response["storage_quota_use"], iec=True if args.iec else False)
 
-            output += f"Storage quota: {used} used"
+            output += f"\nStorage quota: {used} used"
             if storage_quota:
                 output += f" out of {format_file_size(storage_quota, iec=True if args.iec else False)}"
             output += "\n"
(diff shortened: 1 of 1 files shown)


src/borg/archiver/rinfo_cmd.py
```diff
@@ -52,11 +52,11 @@ def do_rinfo(self, args, repository, manifest, cache):
 
             response = repository.info()
             storage_quota = response["storage_quota"]
-            used = format_file_size(response["storage_quota_use"])
+            used = format_file_size(response["storage_quota_use"], iec=True if args.iec else False)
```
ThomasWaldmann, line 55: if args.iec is a boolean, you can directly give it.

if it is not (didn't check), then give `bool(args.iec)` for the same effect as your code.

src/borg/archiver/rinfo_cmd.py
```diff
@@ -52,11 +52,11 @@ def do_rinfo(self, args, repository, manifest, cache):
...
 
-            output += f"Storage quota: {used} used"
+            output += f"\nStorage quota: {used} used"
```
ThomasWaldmann, line 57: ah, a bug fix :-)

ThomasWaldmann:
Thanks for the PR, that was quick! :-)

ThomasWaldmann:
Thanks!

PR #7251, opened 2023-01-09
Fix test_size_on_disk_accurate for large st_blksize, fixes #7250

### LINKED ISSUE #7250: test_size_on_disk_accurate on 1.2.3 breaks when run on ZFS

Yes

Bug/Regression

NixOS Unstable

ZFS Root

n/a

n/a

Running the test suite on ZFS breaks on `test_size_on_disk_accurate` since https://github.com/borgbackup/borg/commit/8ddfd94f0904449175151faed67f9640710b6454.

ZFS has a commit interval of 5 seconds, so a manual flush might be required for the size to be reflected instantly.

```python
_______________ HashIndexSizeTestCase.test_size_on_disk_accurate _______________
[gw42] linux -- Python 3.10.9 /nix/store/ar76b3831ydkmfhf0r9v3wbma5i61iz2-python3-3.10.9/bin/python3.10
self = <borg.testsuite.hashindex.HashIndexSizeTestCase testMethod=test_size_on_disk_accurate>

    def test_size_on_disk_accurate(self):
        idx = ChunkIndex()
        for i in range(1234):
            idx[H(i)] = i, i**2, i**3
            idx.write(file)
            size = os.path.getsize(file.fileno())
>       assert idx.size() == size
E       assert 90350 == 0
E        +  where 90350 = <built-in method size of borg.hashindex.ChunkIndex object at 0xffffeea51b30>()
E        +    where <built-in method size of borg.hashindex.ChunkIndex object at 0xffffeea51b30> = <borg.hashindex.ChunkIndex object at 0xffffeea51b30>.size

/nix/store/0ppx77a4xk6nij2p1klgsnksmrfnx9lj-borgbackup-1.2.3/lib/python3.10/site-packages/borg/testsuite/hashindex.py:202: AssertionError
```

puetzk:
python's io.BufferedWriter sizes its buffer based on st_blksize

If the write fits in this buffer, then it's possible none (or only some) of the data from idx.write() has been written through to fileno()

Fixes #7250

### DIFF
--- a/src/borg/testsuite/hashindex.py
+++ b/src/borg/testsuite/hashindex.py
@@ -258,9 +258,9 @@ class HashIndexSizeTestCase(BaseTestCase):
         idx = ChunkIndex()
         for i in range(1234):
             idx[H(i)] = i, i**2
-        with tempfile.NamedTemporaryFile() as file:
-            idx.write(file)
-            size = os.path.getsize(file.fileno())
+        with unopened_tempfile() as filepath:
+            idx.write(filepath)
+            size = os.path.getsize(filepath)
         assert idx.size() == size
 
 


puetzk:
Per you comment on https://github.com/borgbackup/borg/pull/7252#discussion_r1065080570, doing the same analysis for `master`

Reviewing all usage of `os.path.getsize()` (I see 8 of them, the `getsize(filename) == 17` --cleanup-commits leftover from borg 1.1 is gone), I don't find anything needing an explicit flush except `test_size_on_disk_accurate` here either.

- Needs explicit `flush()` (added by this PR)
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/hashindex.py#L262-L264

- using the `idx.write(str)`, so flushed by the implicit close following [`with open(path,...):` inside  `IndexBase.write`](https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/hashindex.pyx#L124-L126)
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/hashindex.py#L101-L102
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/hashindex.py#L106-L107
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/hashindex.py#L111-L112

- `getsize` is outside of the [with open(fname, 'wb') as fd:](https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/chunker_pytest.py#L28) block
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/testsuite/chunker_pytest.py#L40

- No preceding open/write, AFAICT checking an pre-existing file unless there their caller is possibly passing a filename they just wrote to and still have open. But it doesn't look like it to me (not very experienced with borg's code, though).
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/repository.py#L1579
  https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/repository.py#L1582
 https://github.com/borgbackup/borg/blob/2cfd5397b1e51d7442435c307234dc4871666eff/src/borg/repository.py#L1629
  Though this last one attempts to open and write into `filename` if (and after) getsize indicates that the existing file is too small. But no race there.

ThomasWaldmann:
Thanks!

PR #7252, opened 2023-01-09
[1.2-maint] Fix test_size_on_disk_accurate for large st_blksize, fixes #7250

### LINKED ISSUE #7250: test_size_on_disk_accurate on 1.2.3 breaks when run on ZFS

puetzk:
python's io.BufferedWriter sizes its buffer based on st_blksize

If the write fits in this buffer, then it's possible none (or only some) of the data from idx.write() has been written through to fileno()

Note:1.2-maint counterpart of #7251.

I see that the contribution guidelines ask for PRs only against master (leaving backports up to you maintainers) but in this case I want to publicly post a commit based on 1.2-maint to this repo, so that @mweinelt can have the nixpkgs derivation cherry-pick that commit for testing. I think we understand what's going on, but the failure is being seen on nixos autobuilders that use ZFS for $TEMP, and I don't personally have anything with that configuration, or direct access to those build machines.

Fixes #7250

### DIFF
--- a/src/borg/testsuite/hashindex.py
+++ b/src/borg/testsuite/hashindex.py
@@ -196,9 +196,9 @@ class HashIndexSizeTestCase(BaseTestCase):
         idx = ChunkIndex()
         for i in range(1234):
             idx[H(i)] = i, i**2, i**3
-        with tempfile.NamedTemporaryFile() as file:
-            idx.write(file)
-            size = os.path.getsize(file.fileno())
+        with unopened_tempfile() as filepath:
+            idx.write(filepath)
+            size = os.path.getsize(filepath)
         assert idx.size() == size
 
 


src/borg/testsuite/hashindex.py
```diff
@@ -198,6 +198,7 @@ def test_size_on_disk_accurate(self):
             idx[H(i)] = i, i**2, i**3
         with tempfile.NamedTemporaryFile() as file:
             idx.write(file)
+            file.flush()
             size = os.path.getsize(file.fileno())
```
ThomasWaldmann, line 202: that pattern (write + getsize) is seen at quite some places in the code. did you only see it failing here?

puetzk: Yeah, this was the only failure @mweinelt said the builders were hitting, and he said they passed with this patch: https://github.com/NixOS/nixpkgs/pull/207631#issuecomment-1375094224

Could be the other writes are just larger though (enough to be bigger than his st_blksize) . Though I think the key is that I don't see any others using getsize.*fileno - the others still seem to be writing to the filepath, and when you pass `str` to IndexBase.write it will be open/write/close/getsize, which will work (close causes a flush). See analysis at I posted in https://github.com/borgbackup/borg/issues/7250#issuecomment-1375100187

puetzk: Reviewing all usage of `os.path.getsize()` (I see 9 of them), I don't find any more that seem problematic in 1.2-maint:

- Needs explicit `flush()` (added by this PR)
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/hashindex.py#L200-L202

- using the `idx.write(str)`, so flushed by the implicit close following [`with open(path,...):` inside  `IndexBase.write`](https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/hashindex.pyx#L117-L119)
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/hashindex.py#L101-L102
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/hashindex.py#L106-L107
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/hashindex.py#L111-L112

- `getsize` is outside of the [with open(fname, 'wb') as fd:](https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/chunker_pytest.py#L50) block
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/testsuite/chunker_pytest.py#L50

- No preceding open/write, AFAICT checking an pre-existing file unless there their caller is possibly passing a filename they just wrote to and still have open. But it doesn't look like it to me (not very experienced with borg's code, though).
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/repository.py#L499
  https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/repository.py#L1482
  https://github.com/borgbackup/borg/blob/1ac19b034bdb2d714a58fb01cbb8a8fda891dd90/src/borg/repository.py#L1485
 https://github.com/borgbackup/borg/blob/1ac19b034bdb2d714a58fb01cbb8a8fda891dd90/src/borg/repository.py#L1533
  Though this last one attempts to open and write into `filename` if (and after) getsize indicates that the existing file is too small. But no race there.

ThomasWaldmann: Thanks for checking these.

I also had another look and found that other tests simply avoid this issue by using `unopened_tempfile` context manager.

So I guess the easiest solution is to also use that CM here and not create troubles by trying to work with the open file.

See e.g. `test_resize` in `testsuite/hashindex.py`.

puetzk: Sure, that certainly avoids the problem (since the file handle has been closed). It changes the coverage of this test (to the [`else:` path of `IndexBase.write`](https://github.com/borgbackup/borg/blob/537a814e53e20013a041faa7192da005f137cf5b/src/borg/hashindex.pyx#L120-L121)), but that's what it was covering before 8ddfd94f0904449175151faed67f9640710b6454. And it just avoids the windows/unix difference in tempfile.NamedTemporaryFile too (that you cited as the reason for that commit).

It seems that passing fileno to os.path.getsize() is undocumented functionality:
> https://docs.python.org/3/library/os.path.html#module-os.path says 
> Changed in version 3.6: Accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object).

but fileno is a bare fd, which is not a [path-like object](https://docs.python.org/3/library/os.html#os.PathLike) (no `__fspath__`).

It clearly works in practice, seemingly because the [cpython implentation defers to `os.stat`](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/genericpath.py#L50), and a  [subset of the function in os.* accept fds and transparently substitute](https://docs.python.org/3/library/os.html#files-and-directories) (e.g. `fstat` in this case): see . `fstat`'s been around basically forever (Version 7 AT&T UNIX, I think?), so `os.stat in os.supports_fd` is true for me, and probably for just about everyone. But still, using it with getsize accepting it was undocumented, and stat accepting it is platform-dependent.

So, long story short, `unopened_tempfile` it sounds good to me :-)

ThomasWaldmann: ok, I'll merge it. thanks for your investigations!

PR #7258, opened 2023-01-11
Add tests to Windows CI

RayyanAnsari:


ThomasWaldmann:
"setuptools >= 60.0 is currently incompatible with MSYS2. You can set export SETUPTOOLS_USE_DISTUTILS=stdlib to work around the issue. We are currently working on restoring compatibility."

can you add that right above the env var setting, so we do not forget to remove it again once they fixed it?

ThomasWaldmann:
maybe a good intermediate goal would be to run `pytest -k "not remote"` (and open a ticket about implementing remote repos for win32 and removing that test restriction again).

until someone implements remote repos (`ssh://...` repos), having a hundred tests failures just because this is not yet implemented is rather annoying than useful.

RayyanAnsari:
Should be ready to merge now.

PR #7262, opened 2023-01-13
support win32 for test_safe_unlink* tests

ThomasWaldmann:


ThomasWaldmann:
@RayyanAnsari ^^^

RayyanAnsari:
Looks good, but the reason why it doesn't work is because admin privileges are needed to create links, not because it isn't supported.
https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

ThomasWaldmann:
No, the reason is an AttributeError, not a permissions issue:
```
__________________________ test_safe_unlink_is_safe ___________________________
tmpdir = local('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_safe_unlink_is_safe0')
    def test_safe_unlink_is_safe(tmpdir):
        contents = b"Hello, world\n"
        victim = tmpdir / "victim"
        victim.write_binary(contents)
        hard_link = tmpdir / "hardlink"
>       hard_link.mklinkto(victim)
E       AttributeError: 'LocalPath' object has no attribute 'mklinkto'
```

RayyanAnsari:
Oh, this uses the `mklinkto` function from pytest (which is [not included](https://github.com/pytest-dev/py/blob/master/py/_path/local.py#L109) on win32) instead of `os.link`.

ThomasWaldmann:
OK, so guess I'll give os.link a try.

Does the win32 test runner have the needed privileges?

PR #7263, opened 2023-01-13
delete left over documentation for borg key change-algorithm

helmutg:
Some remainders of borg key change-algorithm were deleted in 
c51a0626de7695f392f80370f39b5b1060c1909a, but that missed the documentation.


ThomasWaldmann:
Thanks!

PR #7264, opened 2023-01-13
Fix locking (win32, master)

ThomasWaldmann:


ThomasWaldmann:
@RayyanAnsari looks like that fixed it, but it's not pretty. Any ideas for improving this?

src/borg/locking.py
```diff
@@ -191,6 +191,8 @@ def kill_stale_lock(self):
             names = os.listdir(self.path)
         except FileNotFoundError:  # another process did our job in the meantime.
             pass
```
ThomasWaldmann, line 200: hmm, thinking about 193:

guess `pass` is a bit pointless (as it runs into the `os.rmdir(self.path)` then, for a non-existing path).

so we could also just `return False` here.

src/borg/locking.py
```diff
@@ -141,7 +141,7 @@ def acquire(self, timeout=None, sleep=None):
             timer = TimeoutTimer(timeout, sleep).start()
             while True:
                 try:
-                    os.rename(temp_path, self.path)
+                    os.replace(temp_path, self.path)
```
RayyanAnsari, line 144: Not related to this current PR, but we should probably look at all uses of `os.rename` and change them to `os.replace` if possible.

src/borg/locking.py
```diff
@@ -170,15 +170,22 @@ def release(self):
...
-        except OSError as err:
-            if err.errno not in (errno.ENOTEMPTY, errno.EEXIST, errno.ENOENT):
-                # EACCES or EIO or ... = we cannot operate anyway, so re-throw
-                raise err
-            # else:
-            # Directory is not empty or doesn't exist any more.
-            # this means we lost the race to somebody else -- which is ok.
+        retry = 0
+        while retry < 42:
+            retry += 1
```
RayyanAnsari, line 175: Is a `while` loop needed here? Or could `for` be used instead:
```py
for _ in range(42):
```

src/borg/locking.py
ThomasWaldmann: Right. There is some os.rename still in borg.cache and also in the testsuite. I'll have a look at that later.

src/borg/locking.py
ThomasWaldmann: Right. I started with `while True:` but then noticed that this might run into an endless loop in case the condition for breaking out of the loop is never met.

ThomasWaldmann:
@RayyanAnsari anything else for here?

Issue #7062, opened 2022-09-28
borg check --archives-only --last-n-seconds


magma1447:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Feature request

#### Your borg version (borg -V).

1.2.2

#### Operating system (distribution) and version.

Debian 10 / 11

#### Hardware / network configuration, and filesystems used.

Linux software raid + mergerfs + ext4

#### How much data is handled by borg?

Approximately 10 TB

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

``borg check --archives-only --last-n-seconds 864000``

## Describe the problem you're observing.

I have a borg repository to which more than one server sends backups. These servers are VM (kvm/libvirt) servers which handles multiple virtual machines. These are snapshoted and the images are sent to borg.

At times I want to run a check on the repository to make sure that it works as intended. I then run a ``borg check --repository-only``, and want to follow that with a ``borg check --archives-only``, which should include all recent archives. The problem is that borg only supports "last n" backups, and I do not know how many that is.

I would be grateful if a ``--last-n-seconds`` (name not important) was implemented, that could be used to run a check on all archives created the last week for example.

My current workaround is to first make a list, send that to ``jq`` which does some magic and returns to me the number of archives that are new enough. Example:
``borg list --json <repository> |jq '[ .archives[] | select(.start |split(".")[0] |strptime("%Y-%m-%dT%H:%M:%S") |mktime > now - 10*86400) ] |length'``

Then I run ``borg check --archives-only --last <number from above>``.

My workaround is a bit cumbersome and introduces a dependency on ``jq``. There is also the case where another server could start a backup between the list and check commands. That could probably be solved by ``borg with-lock``, but it makes it even more complex. Letting borg handle it itself sounds like a better looking solution.

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes, it's a feature request. :)

#### Include any warning/errors/backtraces from the system logs

No logs neccessary.

Michael-Girma:
@ThomasWaldmann Can I pick up this issue?

Michael-Girma:
@ThomasWaldmann Can I pick up this issue?

ThomasWaldmann:
@Michael-Girma Sure!

Michael-Girma:
@ThomasWaldmann  I see two ways of implementing this:

[...]

PR #7272, opened 2023-01-16
Support for date-based matching during archive listing

Michael-Girma:
Support for date-matching during archive listing (fix for #7062)
---
- Added arguments --older --newer --newest --oldest to argparser
- Added parser to detect relative time markers i.e. 1d/1m
- Added an in-house month-offseter
- Edited manifest to initially filter by date if markers are present
- Added test cases to test date matching arguments

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1692,7 +1692,8 @@ class ArchiveChecker:
         self.error_found = False
         self.possibly_superseded = set()
 
-    def check(self, repository, repair=False, first=0, last=0, sort_by="", match=None, verify_data=False):
+    def check(self, repository, repair=False, first=0, last=0, sort_by="", match=None, older=None, oldest=None,
+              newer=None, newest=None, verify_data=False):
         """Perform a set of checks on 'repository'
 
         :param repair: enable repair mode, write updated or corrected data into repository
@@ -1724,7 +1725,8 @@ class ArchiveChecker:
                 self.error_found = True
                 del self.chunks[Manifest.MANIFEST_ID]
                 self.manifest = self.rebuild_manifest()
-        self.rebuild_refcounts(match=match, first=first, last=last, sort_by=sort_by)
+        self.rebuild_refcounts(match=match, first=first, last=last, sort_by=sort_by, older=older, oldest=oldest,
+                               newer=newer, newest=newest)
         self.orphan_chunks_check()
         self.finish()
         if self.error_found:
@@ -1919,7 +1921,8 @@ class ArchiveChecker:
         logger.info("Manifest rebuild complete.")
         return manifest
 
-    def rebuild_refcounts(self, first=0, last=0, sort_by="", match=None):
+    def rebuild_refcounts(self, first=0, last=0, sort_by="", match=None, older=None, oldest=None, newer=None,
+                          newest=None):
         """Rebuild object reference counts by walking the metadata
 
         Missing and/or incorrect data is repaired when detected
@@ -2113,8 +2116,9 @@ class ArchiveChecker:
                     i += 1
 
         sort_by = sort_by.split(",")
-        if any((first, last, match)):
-            archive_infos = self.manifest.archives.list(sort_by=sort_by, match=match, first=first, last=last)
+        if any((first, last, match, older, newer, newest, oldest)):
+            archive_infos = self.manifest.archives.list(sort_by=sort_by, match=match, first=first, last=last,
+                                                        oldest=oldest, newest=newest, older=older, newer=newer)
             if match and not archive_infos:
                 logger.warning("--match-archives %s does not match any archives", match)
             if first and len(archive_infos) < first:
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1701,8 +1701,8 @@ class ArchiveChecker:
         sort_by="",
         match=None,
         older=None,
-        oldest=None,
         newer=None,
+        oldest=None,
         newest=None,
         verify_data=False,
     ):
@@ -1711,10 +1711,12 @@ class ArchiveChecker:
         :param repair: enable repair mode, write updated or corrected data into repository
         :param first/last/sort_by: only check this number of first/last archives ordered by sort_by
         :param match: only check archives matching this pattern
+        :param older/newer: only check archives older/newer than timedelta from now
+        :param oldest/newest: only check archives older/newer than timedelta from oldest/newest archive timestamp
         :param verify_data: integrity verification of data referenced by archives
         """
         logger.info("Starting archive consistency check...")
-        self.check_all = not any((first, last, match))
+        self.check_all = not any((first, last, match, older, newer, oldest, newest))
         self.repair = repair
         self.repository = repository
         self.init_chunks()
@@ -1935,7 +1937,7 @@ class ArchiveChecker:
         return manifest
 
     def rebuild_refcounts(
-        self, first=0, last=0, sort_by="", match=None, older=None, oldest=None, newer=None, newest=None
+        self, first=0, last=0, sort_by="", match=None, older=None, newer=None, oldest=None, newest=None
     ):
         """Rebuild object reference counts by walking the metadata
 
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1692,8 +1692,20 @@ class ArchiveChecker:
         self.error_found = False
         self.possibly_superseded = set()
 
-    def check(self, repository, repair=False, first=0, last=0, sort_by="", match=None, older=None, oldest=None,
-              newer=None, newest=None, verify_data=False):
+    def check(
+        self,
+        repository,
+        repair=False,
+        first=0,
+        last=0,
+        sort_by="",
+        match=None,
+        older=None,
+        oldest=None,
+        newer=None,
+        newest=None,
+        verify_data=False,
+    ):
         """Perform a set of checks on 'repository'
 
         :param repair: enable repair mode, write updated or corrected data into repository
@@ -1725,8 +1737,9 @@ class ArchiveChecker:
                 self.error_found = True
                 del self.chunks[Manifest.MANIFEST_ID]
                 self.manifest = self.rebuild_manifest()
-        self.rebuild_refcounts(match=match, first=first, last=last, sort_by=sort_by, older=older, oldest=oldest,
-                               newer=newer, newest=newest)
+        self.rebuild_refcounts(
+            match=match, first=first, last=last, sort_by=sort_by, older=older, oldest=oldest, newer=newer, newest=newest
+        )
         self.orphan_chunks_check()
         self.finish()
         if self.error_found:
@@ -1921,8 +1934,9 @@ class ArchiveChecker:
         logger.info("Manifest rebuild complete.")
         return manifest
 
-    def rebuild_refcounts(self, first=0, last=0, sort_by="", match=None, older=None, oldest=None, newer=None,
-                          newest=None):
+    def rebuild_refcounts(
+        self, first=0, last=0, sort_by="", match=None, older=None, oldest=None, newer=None, newest=None
+    ):
         """Rebuild object reference counts by walking the metadata
 
         Missing and/or incorrect data is repaired when detected
@@ -2117,8 +2131,16 @@ class ArchiveChecker:
 
         sort_by = sort_by.split(",")
         if any((first, last, match, older, newer, newest, oldest)):
-            archive_infos = self.manifest.archives.list(sort_by=sort_by, match=match, first=first, last=last,
-                                                        oldest=oldest, newest=newest, older=older, newer=newer)
+            archive_infos = self.manifest.archives.list(
+                sort_by=sort_by,
+                match=match,
+                first=first,
+                last=last,
+                oldest=oldest,
+                newest=newest,
+                older=older,
+                newer=newer,
+            )
             if match and not archive_infos:
                 logger.warning("--match-archives %s does not match any archives", match)
             if first and len(archive_infos) < first:
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -353,7 +353,7 @@ def define_exclusion_group(subparser, **kwargs):
     return exclude_group
 
 
-def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
+def define_archive_filters_group(subparser, *, sort_by=True, first_last=True, oldest_newest=True, older_newer=True):
     filters_group = subparser.add_argument_group(
         "Archive filters", "Archive filters can be applied to repository targets."
     )
@@ -397,6 +397,36 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
             default=0,
             type=positive_int_validator,
             help="consider last N archives after other filters were applied",
+        )
+
+    if oldest_newest:
+        group = filters_group.add_mutually_exclusive_group()
+        group.add_argument(
+            "--oldest",
+            metavar="Nd",
+            dest="oldest",
+            help="consider archives N-days after the oldest archive's timestamp",
+        )
+        group.add_argument(
+            "--newest",
+            metavar="Nd",
+            dest="newest",
+            help="consider archives N-days before the newest archive's timestamp",
+        )
+
+    if older_newer:
+        group = filters_group.add_mutually_exclusive_group()
+        group.add_argument(
+            "--older",
+            metavar="Nd",
+            dest="older",
+            help="consider archives older than N-days ago",
+        )
+        group.add_argument(
+            "--newer",
+            metavar="Nd",
+            dest="newer",
+            help="consider archives between now and N-days ago",
         )
 
     return filters_group
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -406,14 +406,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True, ol
             metavar="Nd",
             type=relative_time_marker_validator,
             dest="oldest",
-            help="consider archives N-days after the oldest archive's timestamp",
+            help="consider archives N [days/months] after the oldest archive's timestamp",
         )
         group.add_argument(
             "--newest",
             metavar="Nd",
             type=relative_time_marker_validator,
             dest="newest",
-            help="consider archives N-days before the newest archive's timestamp",
+            help="consider archives N [days/months] before the newest archive's timestamp",
         )
 
     if older_newer:
@@ -423,14 +423,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True, ol
             metavar="Nd",
             type=relative_time_marker_validator,
             dest="older",
-            help="consider archives older than N-days ago",
+            help="consider archives older than N [days/months] ago",
         )
         group.add_argument(
             "--newer",
             metavar="Nd",
             type=relative_time_marker_validator,
             dest="newer",
-            help="consider archives between now and N-days ago",
+            help="consider archives between now and N [days/months] ago",
         )
 
     return filters_group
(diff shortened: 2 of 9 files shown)


Michael-Girma:
@ThomasWaldmann Made my first pass of changes and added a test case for it. The output seems to reflect the specified archives however I keep getting an "{num} orphaned objects found!". Not sure if this is a problem with how I created the archives in my test. All other test cases seem to work fine. Any thoughts?

src/borg/helpers/time.py
```diff
@@ -109,6 +109,48 @@ def format_timedelta(td):
...
+        return from_date
+
+    day_offset_regex = re.compile(r'\d+d')
+    month_offset_regex = re.compile(r'\d+m')
+
+    day_offset_str = day_offset_regex.search(format_string)
+    if day_offset_str is not None:
+        day_offset = int(day_offset_str.group()[:-1])
+        day_offset *= -1 if earlier else 1
+        return from_date + timedelta(days=day_offset)
```
ThomasWaldmann, line 129: the return value of `.search` is a match object, so using `..._str` is rather misleading here.

Just use `match` and `if match:`?

You could also have the days part of the regex as a named group, then you would not need the `[:-1]` hack, but could directly access it.

src/borg/helpers/time.py
```diff
@@ -109,6 +109,48 @@ def format_timedelta(td):
...
+
+    month_offset_str = month_offset_regex.search(format_string)
+    if month_offset_str is not None:
+        month_offset = int(month_offset_str.group()[:-1])
+        month_offset *= -1 if earlier else 1
+        return offset_n_months(from_date, month_offset)
```
ThomasWaldmann, line 135: same here.

src/borg/helpers/time.py
```diff
@@ -109,6 +109,48 @@ def format_timedelta(td):
...
+
+    return from_date
```
ThomasWaldmann, line 137: looks like giving an invalid format string just makes it silently do ... "something".

guess we rather want to fail in such a case and do nothing.

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -54,6 +54,41 @@ def test_check_usage(self):
         self.assert_not_in("archive1", output)
         self.assert_in("archive2", output)
 
+    def test_date_matching(self):
+        shutil.rmtree(self.repository_path)
```
ThomasWaldmann, line 58: is that needed? usually the test setup/teardown cleans up the fs.

ThomasWaldmann (changes requested):

ThomasWaldmann:
Also:
- run `black .` in the toplevel dir of the repo to get the desired automatic formatting we use.
- run `tox -e flake8` to check if there is anything else

ThomasWaldmann:
About the orphaned objects: interesting, guess I will need to locally run this. Usually borg should not create orphans.

src/borg/testsuite/archiver/check_cmd.py
Michael-Girma: I was trying to get rid of the first 2 archives that are created during test setup so that I can use the archives with specific timestamps only.

Michael-Girma:
@ThomasWaldmann I've gone through the request and resolved them except for the ```shutil.rmtree(self.repository_path)``` in the tests. I've run black and tox to format the code as well. Please review when possible.

src/borg/testsuite/archiver/check_cmd.py
ThomasWaldmann: ah, ok!

src/borg/helpers/__init__.py
```diff
@@ -25,7 +25,14 @@
...
 from .parseformat import format_archive, parse_stringified_list, clean_lines
-from .parseformat import Location, location_validator, archivename_validator, comment_validator, text_validator
+from .parseformat import (
+    Location,
+    location_validator,
+    archivename_validator,
+    comment_validator,
+    text_validator,
+    relative_time_marker_validator,
+)
```
ThomasWaldmann, line 35: I like most of the auto-formatting black is doing, but one thing I don't like: that it quickly tends to "vertically" format stuff, even if there is no real need.

In this case I guess we could easily avoid that, by either adding another `from .parseformat import ...` line or even just moving one imported name (e.g. Location) to an already existing import line (e.g. 27).

src/borg/helpers/parseformat.py
```diff
@@ -536,6 +536,17 @@ def validator(text):
     return validator
 
 
+def relative_time_marker_validator(text: str):
+    day_offset_regex = r"^\d+d$"
+    month_offset_regex = r"^\d+m$"
+    day_match = re.compile(day_offset_regex).search(text)
+    month_match = re.compile(month_offset_regex).search(text)
+    if not day_match and not month_match:
```
ThomasWaldmann, line 544: This could be simplified by using `r"^\d+[md]$"`.

That even has the advantage that we could also add y h m s easily.

src/borg/helpers/time.py
```diff
@@ -110,31 +111,29 @@ def format_timedelta(td):
...
-    month_offset_str = month_offset_regex.search(format_string)
-    if month_offset_str is not None:
-        month_offset = int(month_offset_str.group()[:-1])
-        month_offset *= -1 if earlier else 1
-        return offset_n_months(from_date, month_offset)
+        month_offset_match = month_offset_regex.search(format_string)
+        if month_offset_match:
+            month_offset = int(month_offset_match.group("offset"))
+            month_offset *= -1 if earlier else 1
+            return offset_n_months(from_date, month_offset)
```
ThomasWaldmann, line 134: maybe a separate helper function that parses stuff like 42d and returns a timedelta for that would be generally useful (and maybe we could use it even at other places later).

src/borg/manifest.py
```diff
@@ -36,23 +36,27 @@ class MandatoryFeatureUnsupported(Error):
 
 def filter_archives_by_date(archives, oldest=None, newest=None, older=None, newer=None):
     def get_first_and_last_archive_dates(archives_list):
-        dates = [x.ts.date() for x in archives_list]
+        dates = [x.ts for x in archives_list]
         return min(dates), max(dates)
 
-    today = archive_ts_now().date()
+    today = archive_ts_now()
```
ThomasWaldmann, line 42: the names in this function still remind of being capable to compute only dates (like day granularity).

how about:

- today -> now (or now_ts, if you prefer)
- *_date -> *_ts

src/borg/helpers/time.py
Michael-Girma: Think returning a timedelta might be tricky without a given 'from_date' since timedelta can't account for 'months' independently due to the varying number of days within months. How should I proceed?

ThomasWaldmann: oh, right. and same issue we have with leap years.

src/borg/helpers/time.py
```diff
@@ -109,6 +110,48 @@
...
+def calculate_relative_offset(format_string, from_date, earlier=False):
+    """Calculates offset based on a relative marker. 7d (7 days), 8m (8 months)
+    earlier: whether offset should be calculated to an earlier time.
+    """
+    if from_date is None:
+        from_date = archive_ts_now().date()
+
+    if format_string is not None:
+        day_offset_regex = re.compile(r"(?P<offset>\d+)d")
+        month_offset_regex = re.compile(r"(?P<offset>\d+)+m")
```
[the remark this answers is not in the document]
ThomasWaldmann: guess this is a typo.

Michael-Girma:
Applied changes requested above. Please review when possible.

src/borg/helpers/time.py
```diff
@@ -109,6 +110,48 @@ def format_timedelta(td):
...
+    target_year = (total_months - 1) // 12
+    target_month = (total_months % 12) or 12
+
+    # calculate the max days of the target month by subtracting a day from the next month
+    next_month_month = ((total_months + 1) % 12) or 12
+    next_month_year = (total_months + 1) // 12
+    max_days_in_month = (datetime(next_month_year, next_month_month, 1) - timedelta(1)).day
+
+    return datetime(day=min(from_date.day, max_days_in_month), month=target_month, year=target_year).replace(
+        tzinfo=from_date.tzinfo
```
ThomasWaldmann, line 151: guess we also need _date -> _ts here.

src/borg/helpers/time.py
```diff
@@ -109,6 +110,48 @@ def format_timedelta(td):
...
+def calculate_relative_offset(format_string, from_date, earlier=False):
+    """Calculates offset based on a relative marker. 7d (7 days), 8m (8 months)
+    earlier: whether offset should be calculated to an earlier time.
+    """
+    if from_date is None:
+        from_date = archive_ts_now().date()
+
+    if format_string is not None:
+        day_offset_regex = re.compile(r"(?P<offset>\d+)d")
+        month_offset_regex = re.compile(r"(?P<offset>\d+)m")
```
ThomasWaldmann, line 122: you could use a named group for the unit, like `(?P<unit>[md])")`.

then it is just 1 regex match and the rest if just dispatching based on unit == "m", unit =="d", etc. (safes some code duplication).

src/borg/helpers/time.py
```diff
@@ -109,6 +110,48 @@ def format_timedelta(td):
...
+            return offset_n_months(from_date, month_offset)
+
+    raise ValueError(f"{format_string} is not a recognized relative time marker")
+
+
+def offset_n_months(from_date, n_months):
+    # Calculate target month and year by getting completed total_months until target_month
+    total_months = (from_date.year * 12) + from_date.month + n_months
+    target_year = (total_months - 1) // 12
+    target_month = (total_months % 12) or 12
```
ThomasWaldmann, line 143: guess that is correct, but i also guess easier to understand (div / mod == the pure arithmetic way, instead of the mixed arithmetic and boolean-switch way) would be:
```
target_month = (total_months - 1) % 12 + 1
```

Maybe even better / easier / more correct than that would be to move the "- 1" to the computation of total_months, so that it computes total_months == 0 for the very first month (year 0, month 1) and n_months == 0.

src/borg/helpers/time.py
```diff
@@ -109,6 +110,48 @@ def format_timedelta(td):
...
+
+    # calculate the max days of the target month by subtracting a day from the next month
+    next_month_month = ((total_months + 1) % 12) or 12
+    next_month_year = (total_months + 1) // 12
```
ThomasWaldmann, line 147: similar here.

ThomasWaldmann: btw, some guy on twitter recently suggested humanity should get rid of months and just use day-in-year. :-)

but then we still would have leap-years.

Michael-Girma: Gotcha.

Btw where I'm at, we use a calendar where the days a month is constant (30d) but on leap years the last month has 7 days as opposed to 6. :-)

ThomasWaldmann: Sounds easier - where is that?

src/borg/testsuite/archiver/__init__.py
```diff
@@ -172,8 +172,9 @@ def cmd(self, *args, **kw):
         output = empty.join(line for line in output.splitlines(keepends=True) if pp_msg not in line)
         return output
 
-    def create_src_archive(self, name):
-        self.cmd(f"--repo={self.repository_location}", "create", "--compression=lz4", name, src_dir)
+    def create_src_archive(self, name, ts=None):
+        timestamp_arg = f"--timestamp={ts}" if ts else ""
```
ThomasWaldmann, line 176: this is what causes the current test fails. it is because there is an empty string argument.

src/borg/helpers/time.py
Michael-Girma: Ethiopia, Ethiopian calendar.

Michael-Girma:
Renamed vars and refactored the offset calculation to remove code duplication. Modified the create_src_archive as well. Please review when you can

ThomasWaldmann:
The linters are unhappy, see my previous comments about black/flake8.

src/borg/archive.py
```diff
@@ -1692,7 +1692,20 @@ def __init__(self):
...
+        repository,
+        repair=False,
+        first=0,
+        last=0,
+        sort_by="",
+        match=None,
+        older=None,
+        oldest=None,
+        newer=None,
+        newest=None,
```
ThomasWaldmann, line 1706: guess you need to extend the boolean value computation of `self.check_all` to also consider these new parameters which can also make it not checking all archives, but only some.

src/borg/archiver/_common.py
```diff
@@ -399,6 +399,40 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True):
...
+            help="consider archives older than N-days ago",
+        )
+        group.add_argument(
+            "--newer",
+            metavar="Nd",
+            type=relative_time_marker_validator,
+            dest="newer",
+            help="consider archives between now and N-days ago",
+        )
+
```
ThomasWaldmann, line 435: Is it "N-days" or rather "N days"?

Michael-Girma: Should we also edit it to say "before/after the latest matched archive" to denote that the date filtering happens after match-archives?

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -54,6 +54,44 @@ def test_check_usage(self):
...
+        )
+        self.assert_in("archive3", output)
+        self.assert_not_in("archive1", output)
+        self.assert_not_in("archive2", output)
+        output = self.cmd(
+            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--older=1d", exit_code=0
+        )
+        self.assert_in("archive1", output)
+        self.assert_in("archive2", output)
+        self.assert_not_in("archive3", output)
```
ThomasWaldmann, line 93: As time progresses (always), these tests are problematic and will fail in future.

To avoid this and still test newer/older, you could compute the timedelta in days between now and archive3 datetime.

src/borg/archiver/_common.py
ThomasWaldmann: I wouldn't add too much details here in the options description, but include that in the main docs.

ThomasWaldmann:
@Michael-Girma btw, would you like to earn a bounty for this?

we don't have much funds, but I guess I could put a USD 50 bounty on solving this ticket.

Michael-Girma:
> @Michael-Girma btw, would you like to earn a bounty for this?
> 
> we don't have much funds, but I guess I could put a USD 50 bounty on solving this ticket (we use bountysource.com).

That would be awesome. : ))

Michael-Girma:
Ahh forgot to run the formatters before committing. What are your thoughts on utilizing the pre-commit hook for linter stuff?

src/borg/testsuite/archiver/check_cmd.py
Michael-Girma: Ah forgot to remove the timestamp for archive3. If archive3 were to be created with default archive_ts_now() timestamp, then that should suffice, shouldn't it?

ThomasWaldmann:
@Michael-Girma bounty added (see ticket). about pre-commit: open a ticket about it, so we can discuss there.

src/borg/archiver/_common.py
```diff
@@ -423,14 +423,14 @@ def define_archive_filters_group(subparser, *, sort_by=True, first_last=True, ol
...
-            help="consider archives older than N-days ago",
+            help="consider archives older than N [days/months] ago",
         )
         group.add_argument(
             "--newer",
             metavar="Nd",
             type=relative_time_marker_validator,
             dest="newer",
-            help="consider archives between now and N-days ago",
+            help="consider archives between now and N [days/months] ago",
```
ThomasWaldmann, line 433: maybe the metavar should be "TIMESPAN" or so, not "Nd".

the help text then could refer to TIMESPAN (and maybe 7d and 12m as an example).

src/borg/manifest.py
```diff
@@ -106,18 +106,25 @@ def list(
...
+        then filter by timestamp considering *older* and *newer* followed by a second pass to
+        filter the result considering *oldest* and *newest*
+        then *sort_by*.
+
         Apply *first* and *last* filters, and then possibly *reverse* the list.
 
         *sort_by* is a list of sort keys applied in reverse order.
+        *newer* and *older* are relative time markers that indicate offset from now
+        *newest* and *oldest* are relative time markers that indicate offset from newest/oldest archive's timestamp
+
```
ThomasWaldmann, line 127: add some dots and commas?

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -59,10 +59,9 @@ def test_date_matching(self):
         self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
         earliest_ts = "2022-11-20T23:59:59"
         ts_in_between = "2022-12-18T23:59:59"
-        latest_between = "2023-01-17T23:59:59"
         self.create_src_archive("archive1", ts=earliest_ts)
         self.create_src_archive("archive2", ts=ts_in_between)
-        self.create_src_archive("archive3", ts=latest_between)
+        self.create_src_archive("archive3")
```
ThomasWaldmann, line 64: ah, yeah, that's the easiest way to test that.

src/borg/archive.py
ThomasWaldmann: (done)

ThomasWaldmann:
Test fails:
```
E           TypeError: calculate_relative_offset() got an unexpected keyword argument 'from_date'
```

ThomasWaldmann:
```
        output = self.cmd(
            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--newest=1m", exit_code=0
        )
>       self.assert_in("archive2", output)
E       AssertionError: 'archive2' not found in 'Starting archive consistency check...\nAnalyzing archive archive3 (1/1)\nOrphaned objects check skipped (needs all archives checked).\nArchive consistency check complete, no problems found.\n'

.tox/py39-fuse2/lib/python3.9/site-packages/borg/testsuite/archiver/check_cmd.py:78: AssertionError
```

Michael-Girma:
Keep getting a 
```
.tox/py39/lib/python3.9/site-packages/borg/remote.py:40: in <module>
    BORG_VERSION = parse_version(__version__)
.tox/py39/lib/python3.9/site-packages/borg/version.py:24: in parse_version
    raise ValueError("Invalid version string %s" % version)
E   ValueError: Invalid version string 0.1.dev7507+gf88cc5a
```
when trying to run tests locally.

ThomasWaldmann:
Still draft (anything left to do?) oder "ready for review"?

ThomasWaldmann:
@Michael-Girma before running the tests:
```
# have a full git clone and be in the toplevel project dir
python setup.py clean clean2
pip install -e .
```
(and before that, follow the docs and have all OS level packages installed, like libs and header files.)

ThomasWaldmann (approved):
LGTM.

ThomasWaldmann:
https://github.com/borgbackup/borg/pull/7272#issuecomment-1383784312

how can i reproduce these? did you still get that after fixing the full/partial check flag?

Michael-Girma:
> [#7272 (comment)](https://github.com/borgbackup/borg/pull/7272#issuecomment-1383784312)
> 
> how can i reproduce these? did you still get that after fixing the full/partial check (`self.check_all`) flag?

Resetting my env, will test in a few

ThomasWaldmann:
Working on a more condensed commit comment for the squash commit. How about:

```
check --archives: add --newer/--older/--newest/--oldest, fixes #7062

Options accept a timespan, like Nd for N days or Nm for N months.

Use these to do date-based matching on archives and only check some of them,
like: borg check --archives --newest=7d

Author: Michael Deyaso <mdeyaso@fusioniq.io>
```

ThomasWaldmann:
OK, CI looks good (windows CI failures are expected and not due to this PR).

Michael-Girma:
Tests are all green locally as well. Thanks for the help with setting up my env. The squash-commit comment looks good to me.

ThomasWaldmann:
Merged, thanks! Now claim the bounty...

Issue #5535, opened 2020-12-02
TZ issues within automated tests


ThomasWaldmann:
commet by @amikula from https://github.com/borgbackup/borg/pull/5332#issuecomment-708529283 : 

One thing I don't like about the tests is that they're somewhat **dependent on the developer's time zone**, because command line timestamps are in UTC but prune interprets the timestamps in local time.

I've compensated for that by getting the current timezone and offsetting from UTC accordingly in the tests. Also note that **DST changes are not accounted for**. I've put the test timestamps at midday so there won't be problems with this, but it's a kludge that may bite someone later on down the road.

The alternative of calculating the correct local offset for a particular day doesn't seem to be handled by Python core libraries, so I think the tests would need to **bring in a new library to handle localtime correctly for dates in the past**. Even then, the offset for the tests will be different depending on the environment's timezone, so it's probably best to **bring in a testing library** (eg. freezegun) to **make the tests independent of the test environment's timezone**.

FelixSchwarz:
Just wanted to mention that [freezegun](https://github.com/spulec/freezegun/) did work very well for me (in other projects) so +1 for that.

ThomasWaldmann:
Likely fixed by #7282 in master branch.

PR #7282, opened 2023-01-18
prune: compute pruning in utc timezone

ThomasWaldmann:
guess this fixes issues like #5535.

i tested this with TZ=:
- UTC
- UTC+1 (my current local timezone)
- UTC+12
- UTC-12
- UTC+14

### DIFF
--- a/src/borg/archiver/prune_cmd.py
+++ b/src/borg/archiver/prune_cmd.py
@@ -53,8 +53,10 @@ def prune_split(archives, rule, n, kept_because=None):
 
     a = None
     for a in sorted(archives, key=attrgetter("ts"), reverse=True):
-        # we compute the pruning in local time zone
-        period = a.ts.astimezone().strftime(pattern)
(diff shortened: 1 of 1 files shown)


jdchristensen:
Maybe I'm misunderstanding (I haven't tested this), but I don't think that this will be correct.  "borg help prune" documents that days are interpreted in the local timezone.  For example, `--keep-daily 7` should keep the last backup on each day.  If I have backup1, backup2, and backup3, it could be that backup1 and backup2 are on the same day in the local time zone, while backup3 is on the next day, so that backup2 and backup3 should be kept.  But in a different time zone, backup2 and backup3 could be on the same day, with backup1 on the previous day, so backup1 and backup3 would be kept.  So I suggest reverting this commit.

ThomasWaldmann:
As it is, it would work in the UTC day, not in the localtime day.

I'll try if I can get the tests working using a different approach...

PR #7284, opened 2023-01-18
fix prune tests, prune in localtime

ThomasWaldmann:


ThomasWaldmann:
@jdchristensen ^^^

src/borg/testsuite/helpers.py
```diff
@@ -405,6 +405,17 @@ def __repr__(self):
...
 
+# This is the local timezone of the system running the tests.
+# We need this e.g. to construct archive timestamps for the prune tests,
+# because borg prune operates in the local timezone (it first converts the
+# archive timestamp to the local timezone). So, if we want the y/m/d/h/m/s
+# values which prune uses to be exactly the ones we give [and NOT shift them
+# by tzoffset], we need to give the timestamps in the same local timezone.
+# Please note that the timestamps in a real borg archive or manifest are
+# stored in UTC timezone.
+local_tz = datetime.now().astimezone().tzinfo
```
ThomasWaldmann, line 416: Is this the easiest way to get the local timezone?

ThomasWaldmann: guess it could also be done via time.localtime(), .tm_gmtoff and timezone(timedelta(seconds=tm_gmtoff)), but that's rather more complicated.

jdchristensen: Searching found lots of things.  `datetime.now().astimezone().tzinfo` was one option, but some people prefer `datetime.utcnow().astimezone().tzinfo` for some reason.

jdchristensen (approved):
LGTM.

jdchristensen:
Should the pruning help text say that it is the *client* local timezone that matters, on the system running the prune command (not the system that made the backups)?

src/borg/testsuite/helpers.py
ThomasWaldmann: I'll replace it by `datetime.now(tz=timezone.utc).astimezone(tz=None).tzinfo`.

This avoids default argument values and also naive datetime objects.

Some time ago, I already refactored a lot of other places to use `datetime.now(tz=timezone.utc)`.

ThomasWaldmann:
OK, all done, docs clarified. Thanks for your feedback @jdchristensen!

PR #7285, opened 2023-01-19
use local time / local timezone to output timestamps, fixes #7283

### LINKED ISSUE #7283: To UTC or not to UTC...
#7282 fixed the issue that our tests fail when they are run in some timezones (e.g. `TZ=UTC+12`) by changing the timezone in which borg prunes to UTC. Update: #7284 reverts #7282 and fixes the tests differently.

This corresponds to borg2 currently using UTC for about everything (only exception is when giving `--timestamp=...` on the commandline **without** specifying a tz offset like `+01:00`, then it will use local time).

Some IRC conversation related to this:
```
18:32  ThomasWaldmann$ guess this might give slightly different behaviour from user expectations, but it is more consistent with borg2 having utc archive timestamps.
18:33  ThomasWaldmann$ also, our unit tests failed when running them in some timezones like utc+12 and with that change, they now work in any TZ.
18:35  ThomasWaldmann$ borg2 list now outputs archive listing like:
18:35  ThomasWaldmann$ arch                                 Wed, 2023-01-18 17:34:40 +0000  [592...0d12c]
18:35  ThomasWaldmann$ so that is also utc
18:36  tao$ is it always UTC? or is that just because you set the TZ to UTC
18:36  ThomasWaldmann$ for local users maybe needs a bit getting used to it. for internationally dispersed users with common repos it might be simpler now.
18:36  tao$ hmmm. i'd not like that. i'd rather see local time in every time stamp
18:36  ThomasWaldmann$ no, the archive list is always utc.
18:37  ThomasWaldmann$ yeah, that's what i mean with "getting used to it". :-)
18:38  ThomasWaldmann$ but just imagine you have that international context. having people in different TZs making backups and later looking at them.
18:38  ThomasWaldmann$ then you'ld have all sorts of hard-to-compare timestamps.
18:38  tao$ i don't want to get used to that though. if i make a backup at 19:00 and list the archives i want to see one that says 19:00. setting it to UTC is fine, but should be some way of setting it to localtime too surely?
18:39  ThomasWaldmann$ the ts in the archive is in utc. when running borg list, borg could recompute that to local time.
18:40  tao$ yes. i only mean when i display the list using rlist. internally in the repo metadata i don't care about
18:40  ThomasWaldmann$ but then, if your friend in london made a backup at 18:00 local time there, you'ld get it displayed as being from 19:00 (in germany)
18:41  ThomasWaldmann$ which refers to the same point in time, but might make discussion about it a bit confusing.
18:42  tao$ way i would see it working would be to internally store the timestamp in UTC and then when i do borg rlist it uses the local TZ setting to display it in the local timezone, for UK, Germany, USA whatever
18:42  ThomasWaldmann$ like e.g. if he says "restore the backup of 18:00"
18:43  ThomasWaldmann$ then you might restore the wrong one.
18:43  ThomasWaldmann$ (assuming there is one per hour, for example)
18:43  tao$ i might be wrong about this, but most users i would expect only have one person using that repo in one single timezone. multiple users in multiple timezones seems less likely
18:44  ThomasWaldmann$ that might be, yes. but there is a lot of remote work and there are international companies and organisations.
18:45  tao$ could it be a switch/environment variable at least so you can choose UTC or localtime in operations? or is that too complicated
18:46  ThomasWaldmann$ that's something we could think about and someone would need to write the code.
18:46  ThomasWaldmann$ what i am changing above (url) is not the list behaviour btw (that was done quite a while ago), but how prune looks at dates/times.
18:46  tao$ it's feeling a bit like the --iec thing where iB stats are the only thing that makes sense to me but it's not the default :)
18:47  ThomasWaldmann$ yeah. ISO vs. IEC.
18:47  ThomasWaldmann$ HDD manufacturer like ISO better (because then their HDDs are "bigger")
18:49  ThomasWaldmann$ at least borg2 will be much more precise no matter what ones chooses, as it is always giving the tz offset (like +00:00 for utc tz)
18:50  ThomasWaldmann$ in some contexts (e.g. when looking at log files) that can be quite important
18:51  ThomasWaldmann$ you don't want to get into prison just because somebody confused the datetime when determining who had the IP address in question 
                       at a specific time.
18:53  tao$ yes. i absolutely agree when it's got international use. just for the use case of a one person personal backup, UTC seems wrong
18:54  tao$ it's not so bad for me because i'm in the UK so UTC == GMT, but it will still be an hour out during DST. will be really weird though for people in USA, Japan etc
18:59  ThomasWaldmann$ besides the display issue, there would be also the "pruning behaviour" issue.
```

ThomasWaldmann:
"ls" and also "tar" show timestamps as local time. borg now does the same, but also shows the local tzoffset.

this also affects JSON output.


ThomasWaldmann:
@jdchristensen ^^^

src/borg/testsuite/archiver/transfer_cmd.py
```diff
@@ -85,10 +85,10 @@ def local_to_utc(local_naive, tzoffset, tzinfo):
             del expected_archive["barchive"]
             # timestamps:
             # borg 1.2 transformed to local time and had microseconds = 0, no tzoffset
-            # borg 2 uses an utc timestamp, with microseconds and with tzoffset
+            # borg 2 uses local time, with microseconds and with tzoffset
             for key in "start", "time":
-                # fix expectation: local time meant +01:00, so we convert that to utc +00:00
-                expected_archive[key] = local_to_utc(expected_archive[key], repo12_tzoffset, timezone.utc)
+                # fix expectation: local time meant +01:00, so we convert that to whatever local tz is here.
```
jdchristensen, line 90: I don't understand this comment.  What does "fix expectation" mean?  And why does it say "local time meant +01:00"?

ThomasWaldmann: the json test data i generated back then on my machine with borg 1.2 was in +01:00 timezone (germany, non-DST) - is is contained in a tar file in master branch. borg 1.2 did not include the tzoffset in its json timestamps, so that is something one needs to just know.

then there is what we **got** (from generated borg2 json output) and what we **expected** (from the borg 1.2 json file). as both jsons were generated from the same repository (in case of borg2, it is after "borg transfer" of the archives), both jsons represent the same data, but some differences are expected in the json due to the breaking changes in borg2.

at the end of this test, i basically do `assert got == expected` for the whole deserialized json dict. for that to work, i transform the expected value to be able to precisely match the value we got from borg2 (for timestamps, this is basically: asume +01:00 for the naive ts we got, then transform to a non-naive ts with local tzoffset of the test machine [which is currently also +01:00 on my machine, but could be any other offset]).

ThomasWaldmann:
> I was surprised that the current behaviour shows a mix of timezones. Not sure what is going on there. 

Hmm, guess I need to check whether borg has non-utc timestamps in the manifest (or archive metadata). I thought it is all UTC in there, but I could be wrong.

ThomasWaldmann:
@jdchristensen Hmm, I can't reproduce getting that "mixed" listing with master branch. Did you use an old borg2 for this?

```
Successfully installed borgbackup-2.0.0b5.dev137+gbe1048ab  # == current master branch
% borg -r /tmp/b2master rcreate -e none

# note that my local timezone is +01:00, local time was ~17:01 and master branch (stores and) displays UTC, always. 
% borg -r /tmp/b2master create arch docs
% borg -r /tmp/b2master rlist           
arch                                 Thu, 2023-01-19 16:00:49 +0000 [69e71fb5f47f171faee57d7a27e2ad55c6d171b6996b0b38c41086ae34916b2e]

% TZ=XXX+5 borg -r /tmp/b2master create arch-tz5 docs
% borg -r /tmp/b2master rlist                        
arch                                 Thu, 2023-01-19 16:00:49 +0000 [69e71fb5f47f171faee57d7a27e2ad55c6d171b6996b0b38c41086ae34916b2e]
arch-tz5                             Thu, 2023-01-19 16:01:42 +0000 [7aaa80fd87a35f8d637e0c591110690c85682174e0d5209d7ef911e4f4b28e86]

% TZ=XXX+5 borg -r /tmp/b2master rlist
arch                                 Thu, 2023-01-19 16:00:49 +0000 [69e71fb5f47f171faee57d7a27e2ad55c6d171b6996b0b38c41086ae34916b2e]
arch-tz5                             Thu, 2023-01-19 16:01:42 +0000 [7aaa80fd87a35f8d637e0c591110690c85682174e0d5209d7ef911e4f4b28e86]
```

jdchristensen:
I can't remember exactly what I did, but there's a good chance I used `borg2 2.0.0b4` for the first four archives.  Would that explain it?

ThomasWaldmann:
Yes, likely that was 2.0.0b4 - changeset 15139a80c96cb299a050f6d6f57364dc594a999b was after b4 release.

ThomasWaldmann:
OK, guess this is finished (considering screen and json output)?

jdchristensen:
LGTM.

PR #7289, opened 2023-01-19
ArchiveItem.cmdline list-of-str -> .command_line str, fixes #7246

### LINKED ISSUE #7246: (recreate_)cmdline encoding / binary requirement
borg1: `archive.cmdline` is basically a copy of the `sys.argv` list and when accessing it, borg1 even expects surrogate escapes in each list element. when outputting the cmdline to screen or to json, borg1 removes the surrogate escapes. also, it uses `shlex.quote` on each list element and joins the list elements together with blanks in between. 

borg2: I'ld like to simplify that, just not sure how.

the reason why I want to simplify it is that if we really must have a s-e string, then the corresponding json would need to have `cmdline` and `cmdline_b64` keys (with **lists** as values) to correctly represent this information without data loss. But this seems overkill. And I could not even base64 encode the whole cmdline in one go, but it would have to be a list with b64 encoded elements...

So:
- borg2: do we have to expect surrogate escapes in `sys.argv` anyway?
- are there borg1 archives with surrogate escapes in `archive.cmdline` and if yes, could we just remove or replace them at `borg transfer` time?
- could `cmdline` just be one simple, valid unicode string (without s-e) and not a `sys.argv` list copy?

The same issues apply to `archive.recreate_cmdline` (`borg recreate` uses this).

This is related to #7232 and #6151.

ThomasWaldmann:


docs/internals/data-structures.rst
```diff
@@ -542,7 +542,7 @@ The archive object itself further contains some metadata:
   in the manifest, but leaves the *name* field of the archives as it was.
 * *item_ptrs*, a list of "pointer chunk" IDs.
   Each "pointer chunk" contains a list of chunk IDs of item metadata.
-* *cmdline*, the command line which was used to create the archive
+* *command_line*, the command line which was used to create the archive
```
ThomasWaldmann, line 545: Changed this for 2 reasons:
- it gets the ArchiveItem attribute name in sync with the JSON key
- it simplified dealing with the different types of the value (list-of-str -> str).

PR #7300, opened 2023-01-24
use platformdirs

### LINKED ISSUE #7281: borg2: use platformdirs python package?
@RayyanAnsari suggested to use the `platformdirs` python lib ( https://github.com/platformdirs/platformdirs ) and I had a closer look now:

- the code quality seems decent, from an experienced developer
- the amount of code is not much, but also not trivial. it might be also interesting for us that it supports android also (but not sure it is relevant for termux users).
- some dists have it already packaged, some do not (looked at debian,ubuntu,fedora), so there might be some work for packagers.
- there is 1 open ticket in their tracker with some discussion whether they use some (macOS) directory correctly or not. https://github.com/platformdirs/platformdirs/issues/98
- so, i guess we could add it as a python dependency for borg if we can get rid of some of the corresponding code we have.
- we would need to be careful, so we do not break "borg transfer" when it reads borg1 archives/repos (config/cache access the borg1 way).
- for borg2 config/cache, we could be incompatible as that would be all-new anyway.
- it seems to not yet autocreate not-yet existing directories, but guess the maintainer there is open for a PR with that. https://github.com/platformdirs/platformdirs/issues/120

RayyanAnsari:
fixes #7281

### DIFF
--- a/src/borg/testsuite/helpers.py
+++ b/src/borg/testsuite/helpers.py
@@ -672,9 +672,12 @@ def test_get_keys_dir(monkeypatch):
     monkeypatch.delenv("BORG_BASE_DIR", raising=False)
     if is_win32:
         monkeypatch.delenv("BORG_KEYS_DIR", raising=False)
-        assert get_keys_dir(legacy=False) == os.path.join(
-            os.path.expanduser("~"), "AppData", "Local", "borg", "borg", "keys"
-        )
+        assert get_keys_dir() == os.path.join(os.path.expanduser("~"), "AppData", "Local", "borg", "borg", "keys")
+    elif is_darwin:
+        monkeypatch.delenv("BORG_KEYS_DIR", raising=False)
+        assert get_keys_dir() == os.path.join(os.path.expanduser("~"), "Library", "Preferences", "borg", "keys")
+        monkeypatch.setenv("BORG_KEYS_DIR", "/var/tmp")
+        assert get_keys_dir() == "/var/tmp"
     else:
         monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
         monkeypatch.delenv("BORG_KEYS_DIR", raising=False)
--- a/src/borg/testsuite/helpers.py
+++ b/src/borg/testsuite/helpers.py
@@ -584,13 +584,18 @@ def test_get_base_dir(monkeypatch):
     monkeypatch.delenv("BORG_BASE_DIR", raising=False)
     monkeypatch.delenv("HOME", raising=False)
     monkeypatch.delenv("USER", raising=False)
-    assert get_base_dir() == os.path.expanduser("~")
+    assert get_base_dir(legacy=True) == os.path.expanduser("~")
     monkeypatch.setenv("USER", "root")
-    assert get_base_dir() == os.path.expanduser("~root")
+    assert get_base_dir(legacy=True) == os.path.expanduser("~root")
     monkeypatch.setenv("HOME", "/var/tmp/home")
-    assert get_base_dir() == "/var/tmp/home"
+    assert get_base_dir(legacy=True) == "/var/tmp/home"
     monkeypatch.setenv("BORG_BASE_DIR", "/var/tmp/base")
-    assert get_base_dir() == "/var/tmp/base"
+    assert get_base_dir(legacy=True) == "/var/tmp/base"
+    # non-legacy is much easier:
+    monkeypatch.delenv("BORG_BASE_DIR", raising=False)
+    assert get_base_dir(legacy=False) is None
+    monkeypatch.setenv("BORG_BASE_DIR", "/var/tmp/base")
+    assert get_base_dir(legacy=False) == "/var/tmp/base"
 
 
 def test_get_base_dir_compat(monkeypatch):
--- a/src/borg/testsuite/helpers.py
+++ b/src/borg/testsuite/helpers.py
@@ -702,29 +702,28 @@ def test_get_keys_dir(monkeypatch):
 def test_get_security_dir(monkeypatch):
     """test that get_security_dir respects environment"""
     monkeypatch.delenv("BORG_BASE_DIR", raising=False)
+    home_dir = os.path.expanduser("~")
     if is_win32:
         monkeypatch.delenv("BORG_SECURITY_DIR", raising=False)
-        assert get_security_dir() == os.path.join(
-            os.path.expanduser("~"), "AppData", "Local", "borg", "borg", "security"
-        )
+        assert get_security_dir() == os.path.join(home_dir, "AppData", "Local", "borg", "borg", "security")
         assert get_security_dir(repository_id="1234") == os.path.join(
-            os.path.expanduser("~"), "AppData", "Local", "borg", "borg", "security", "1234"
+            home_dir, "AppData", "Local", "borg", "borg", "security", "1234"
         )
+        monkeypatch.setenv("BORG_SECURITY_DIR", home_dir)
+        assert get_security_dir() == home_dir
     elif is_darwin:
         monkeypatch.delenv("BORG_SECURITY_DIR", raising=False)
-        assert get_security_dir() == os.path.join(os.path.expanduser("~"), "Library", "Preferences", "borg", "security")
+        assert get_security_dir() == os.path.join(home_dir, "Library", "Preferences", "borg", "security")
         assert get_security_dir(repository_id="1234") == os.path.join(
-            os.path.expanduser("~"), "Library", "Preferences", "borg", "security", "1234"
+            home_dir, "Library", "Preferences", "borg", "security", "1234"
         )
         monkeypatch.setenv("BORG_SECURITY_DIR", "/var/tmp")
         assert get_security_dir() == "/var/tmp"
     else:
         monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
         monkeypatch.delenv("BORG_SECURITY_DIR", raising=False)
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
Looks like MYPY cancels it all.

About the tests:

Maybe rather test legacy paths == not-legacy paths and also with/without misc. env vars set, so we can be sure nothing breaks for accessing borg1 stuff.

ThomasWaldmann:
shall I help with some commits so this advances faster?

RayyanAnsari:
> Looks like MYPY cancels it all.

This was because I forgot to add `platformdirs` to `setup.cfg`. It should work now.

> shall I help with some commits so this advances faster?

That would be great - I think all that's left is some small details in the tests. I should be able to get round to this in a few days.

ThomasWaldmann:
Looks like with `legacy=False`, using `BORG_BASE_DIR` will be unsupported.

Guess that would make reading stuff from existing borg1 setups fail (e.g. for `borg transfer`).

Guess more tests are needed, esp. comparing legacy to non-legacy dirs.

ThomasWaldmann:
I'll do a rebase on master / force push soon...

src/borg/testsuite/helpers.py
```diff
@@ -607,6 +607,19 @@ def test_get_config_dir(monkeypatch):
...
 
+def test_get_config_dir_compat(monkeypatch):
+    """test that it works the same for legacy and for non-legacy implementation"""
+    monkeypatch.delenv("BORG_CONFIG_DIR", raising=False)
+    monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
+    assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
+    # TODO fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/Users/tw/.config/borg'
+    monkeypatch.setenv("XDG_CONFIG_HOME", "/var/tmp/.config1")
+    assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
+    # TODO fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/var/tmp/.config1/borg'
```
ThomasWaldmann, line 618: looks like platformdirs does not respect it when XDG_CONFIG_HOME is set.

src/borg/testsuite/helpers.py
```diff
@@ -607,6 +607,19 @@ def test_get_config_dir(monkeypatch):
         assert get_config_dir() == "/var/tmp"
 
 
+def test_get_config_dir_compat(monkeypatch):
+    """test that it works the same for legacy and for non-legacy implementation"""
+    monkeypatch.delenv("BORG_CONFIG_DIR", raising=False)
+    monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
+    assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
+    # TODO fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/Users/tw/.config/borg'
```
ThomasWaldmann, line 615: on macOS, it looks like platformdirs uses a different default: not in `~/.config/<appname>`.

src/borg/testsuite/helpers.py
```diff
@@ -607,6 +607,19 @@ def test_get_config_dir(monkeypatch):
...
+    monkeypatch.setenv("XDG_CONFIG_HOME", "/var/tmp/.config1")
+    assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
+    # TODO fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/var/tmp/.config1/borg'
+    monkeypatch.setenv("BORG_CONFIG_DIR", "/var/tmp/.config2")
+    assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
```
ThomasWaldmann, line 620: on macOS, this worked on.

ThomasWaldmann:
Likely, we'll have the same compatibility issues for the cache dir as seen for the config dir - but let's just first get the config dir sorted.

src/borg/testsuite/helpers.py
ThomasWaldmann: on Linux, this works ok.

src/borg/testsuite/helpers.py
ThomasWaldmann: on Linux, this works ok.

src/borg/testsuite/helpers.py
ThomasWaldmann: on Linux, this works ok.

ThomasWaldmann:
On native win32, guess we can ignore legacy vs. non-legacy compatibility as nobody uses borg < 2 there.

src/borg/testsuite/helpers.py
ThomasWaldmann: on cygwin, this works ok.

https://github.com/platformdirs/platformdirs/issues/58#issuecomment-1416242459

src/borg/testsuite/helpers.py
ThomasWaldmann: on cygwin, this works ok.

src/borg/testsuite/helpers.py
ThomasWaldmann: on cygwin, this works ok.

src/borg/testsuite/helpers.py
ThomasWaldmann: win32 fail: assert 'C:/Users/runneradmin/AppData/Local/borg/borg' == 'D:\\a\\_temp\\msys64\\home\\runneradmin/.config/borg'

RayyanAnsari: Looks like it is using the `HOME` env var, which inside MSYS2 is set to the home directory of the MSYS2 installation.
```py
>>> os.path.expanduser("~")
'C:/Users/Rayyan'
>>> os.environ.get("HOME")
'C:\\msys64\\home\\Rayyan'
```

ThomasWaldmann: ok, i excluded that "legacy compat test" on win32, no need to have it there.

BTW, that `.../Local/borg/borg` path is correct?

ThomasWaldmann:
Test look good.

@RayyanAnsari maybe give the whole PR a review, please.

Are the tests complete, do they make sense? Something missing?

src/borg/testsuite/helpers.py
```diff
@@ -584,60 +584,142 @@ def test_get_base_dir(monkeypatch):
...
-    monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
-    assert get_config_dir() == os.path.join(os.path.expanduser("~"), ".config", "borg")
-    monkeypatch.setenv("XDG_CONFIG_HOME", "/var/tmp/.config")
-    assert get_config_dir() == os.path.join("/var/tmp/.config", "borg")
-    monkeypatch.setenv("BORG_CONFIG_DIR", "/var/tmp")
-    assert get_config_dir() == "/var/tmp"
+    monkeypatch.delenv("BORG_BASE_DIR", raising=False)
+    if is_win32:
+        monkeypatch.delenv("BORG_CONFIG_DIR", raising=False)
+        assert get_config_dir() == os.path.join(os.path.expanduser("~"), "AppData", "Local", "borg", "borg")
```
RayyanAnsari, line 617: perhaps we could also test `BORG_CONFIG_DIR` on win32 as well

src/borg/testsuite/helpers.py
```diff
@@ -584,60 +584,142 @@ def test_get_base_dir(monkeypatch):
...
+    if not is_darwin and not is_win32:
+        monkeypatch.delenv("BORG_CONFIG_DIR", raising=False)
+        monkeypatch.delenv("XDG_CONFIG_HOME", raising=False)
+        # fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/Users/tw/.config/borg'
+        # fails on win2 MSYS2 (but we do not need legacy compat there).
+        assert get_config_dir(legacy=False) == get_config_dir(legacy=True)
+    if not is_darwin and not is_win32:
+        monkeypatch.setenv("XDG_CONFIG_HOME", "/var/tmp/.config1")
+        # fails on macOS: assert '/Users/tw/Library/Preferences/borg' == '/var/tmp/.config1/borg'
+        # fails on win2 MSYS2 (but we do not need legacy compat there).
```
RayyanAnsari, line 645: probably should be `win32` not `win2`

ThomasWaldmann: oh, oops. fixing...

src/borg/testsuite/helpers.py
ThomasWaldmann: right. i fixed that at all places. also some minor refactorings.

src/borg/testsuite/helpers.py
ThomasWaldmann: @RayyanAnsari seen this?

ThomasWaldmann:
Guess we should wait with this until platformdirs 3.0.0 (== after my PRs there are merged), assuming that will happen soon. In case it won't be soon, we should open a ticket about requiring >= 3.0.0 before borg goes RC/release.

Important because we do not want to break directory paths for macOS users **twice**.

ThomasWaldmann:
TODO: consider what happens at `borg transfer` time and what docs or code are needed for that, especially on macOS where the dirs will change in any case. On other OSes, dirs might change under some circumstances.

For keyfile (not: repokey) users this might mean that the borg 1.x key can't be found.

ThomasWaldmann:
@RayyanAnsari merged! thanks for working on this!

PR #7305, opened 2023-01-31
volume based checkpointing, fix item_ptrs orphaned chunks

ThomasWaldmann:


ThomasWaldmann:
We didn't have a test for the checkpointing yet as it was hard to test (only triggered by time or SIGINT).

Had the idea that if we trigger by volume it would be way easier for tests (and also adds `--checkpoint-volume`, maybe even a useful feature).

Interestingly, the repo check fails due to orphan chunks, so maybe this has discovered a bug already.

jdchristensen:
I'm guessing that this fixes the specific issue raised by my test script here: 
 https://github.com/borgbackup/borg/issues/6687#issuecomment-1345625144

(But I think this is different from the main issue discussed in #6687 , which is when `borg compact` creates orphans.)

It might be a couple of days before I get a chance to redo my test, but I'll report back.  Does this issue affect borg1 as well?

ThomasWaldmann:
@jdchristensen when i first noticed the orphans here, i also thought "maybe this is reproducing what we have seen since borg 1.2".

but that is not the case:
- i specifically identified the orphans being item_ptrs chunks by their chunkid in debug logging
- item_ptrs is a new feature in borg 2. borg 1.2 does not have this scalability fix (which vastly increases max. archive metadata stream size).

It might well be what you have seen with your borg2 testing script though.

jdchristensen:
With the latest commit (d0344cb8f) which includes this fix, my test script still shows 1 orphan during the first check.  It's reproducible. The first file in the output, and the second is the script.

[borg2-stress-bugfix.txt](https://github.com/borgbackup/borg/files/10550859/borg2-stress-bugfix.txt)
[borg2-stress-trim.txt](https://github.com/borgbackup/borg/files/10550860/borg2-stress-trim.txt)

ThomasWaldmann:
@jdchristensen oh, interesting. i'll have a look.

ThomasWaldmann:
The first line says it is 2.0.0b4 - usually setuptools_scm should create a more recent version number for the current code.

jdchristensen:
Whoops.  I checked `borg -V` before starting the script, and got the new version, but the script runs `borg2`, which got my installed version.  I'm re-running it now with `borg2 2.0.0b5.dev164+gd0344cb8` and so far there are no errors.

jdchristensen:
I ran a more aggressive version of my script for about 16 hours, and it didn't find any errors.  (This is the same as what happens when I run a similar script using borg 1.2.  I still haven't been able to trigger the orphans that people are finding after `borg compact`.)

Issue #7296, opened 2023-01-23
fix --newer/older/newest/oldest options / support code


ThomasWaldmann:
master branch only:

as `define_archive_filters_group` defaults to adding these options, **all** commands calling this now suddenly "support" these options, according to their help.

but they do not really support these options, because the appropriate code is not there:
- `list_considering` does not yet support them
- check other places

ThomasWaldmann:
@Michael-Girma ^ found this issue after merging.

Michael-Girma:
I did make changes to ```list_considering``` should those be removed?

ThomasWaldmann:
I think you only changed `.list`, but not `.list_considering`.

ThomasWaldmann:
btw, update your local master branch first from borg master.

PR #7306, opened 2023-01-31
Added date-matching support for list_considering

Michael-Girma:
Modified list_considering's call to list to support date based matching for #7296 
---
- Passing date-based matching args to list function
- Created a test function that utilizes rlist to test list_considering's support

### DIFF
--- a/src/borg/manifest.py
+++ b/src/borg/manifest.py
@@ -169,7 +169,7 @@ class Archives(abc.MutableMapping):
             older=args.older,
             newer=args.newer,
             oldest=args.oldest,
-            newest=args.newest
+            newest=args.newest,
         )
 
     def set_raw_dict(self, d):
--- a/src/borg/manifest.py
+++ b/src/borg/manifest.py
@@ -166,6 +166,10 @@ class Archives(abc.MutableMapping):
             match=args.match_archives,
             first=args.first,
             last=args.last,
+            older=args.older,
+            newer=args.newer,
+            oldest=args.oldest,
+            newest=args.newest
         )
 
     def set_raw_dict(self, d):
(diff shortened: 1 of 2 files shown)


src/borg/manifest.py
```diff
@@ -166,6 +166,10 @@ def list_considering(self, args):
             match=args.match_archives,
             first=args.first,
             last=args.last,
+            older=args.older,
+            newer=args.newer,
+            oldest=args.oldest,
+            newest=args.newest,
```
ThomasWaldmann, line 172: did you check all callers of `list_considering` whether they set up these attributes of `args` (== whether they support the corresponding cmd line options in their arg parser)?

ThomasWaldmann: (if the option parser for these options is not added, you'ld get an `AttributeError` here)

ThomasWaldmann (commented):
Looks good, just 1 question.

src/borg/manifest.py
ThomasWaldmann: maybe a less fragile way would be to use `x=getattr(args, "x", correctdefault)`.

i also missed to do that for first and last.

Michael-Girma: Yeah, that seems like a better way to do it.

ThomasWaldmann:
btw, if you mention "fixes #7296" in a commit comment or PR title, it will automatically close that issue if this is merged.

ThomasWaldmann (approved):
lgtm.

ThomasWaldmann:
Thanks!

The linter (black) fail was strange btw:
- seems unrelated to your changes
- uses same black version on github as we use locally
- other PRs do not have this linter fail

PR #7309, opened 2023-01-31
remove part files from final archive

ThomasWaldmann:
checkpoint archives might have a single, incomplete part file as last item.

part files are always a prefix of the full file, growing in size from checkpoint to checkpoint.

we now manage the archive items metadata stream in a special way:
- checkpoint archive A(n) might end with a partial item PI(n)
- checkpoint archive A(n+1) does not contain PI(n)
- checkpoint archive A(n+1) contains a new partial item PI(n+1)
- the final archive does not contain any partial items


ThomasWaldmann:
@RonnyPfannschmidt your approach with "tail management", thanks!

ThomasWaldmann:
OK, guess this is finished.

PR #7310, opened 2023-02-01
avoid orphan content chunks on BackupOSErrors

ThomasWaldmann:


ThomasWaldmann:
Note: this likely affects borg 1.x also. But due to the amount of changes, we better do not backport this (esp. considering that orphaned chunks are a harmless issue).

ThomasWaldmann:
@jdchristensen running your stress test script with code from this branch right now...

jdchristensen:
The version I last uploaded called the archives "::host*" because I was still using borg1 syntax, so it didn't prune correctly.  Here is an updated version, which also includes other features.  I ran this for 16 hours without generating any errors.

[borg2-stress.txt](https://github.com/borgbackup/borg/files/10571747/borg2-stress.txt)

PR #7317, opened 2023-02-01
Fix test hanging reading FIFO when `borg create` failed

pgerber:


src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -784,6 +784,12 @@ def fifo_feeder(fifo_fn, data):
         try:
             self.cmd(f"--repo={self.repository_location}", "create", "--read-special", "test", "input/link_fifo")
         finally:
+            # In case `borg create` failed to open FIFO, read all data to avoid join() hanging.
+            fd = os.open(fifo_fn, os.O_RDONLY | os.O_NONBLOCK)
+            try:
+                os.read(fd, len(data))
+            finally:
+                os.close(fd)
```
ThomasWaldmann, line 792: looks ok. but why would borg create fail to open the FIFO?

pgerber: Well, in my case a import got left off during a rebase and I ended up with a stuck test rather than a usable error message.

ThomasWaldmann: hmm, isn't that unusual somehow?

i mean when the test have start running, after a relatively short time everything borg needs should be imported and reside in memory.

ThomasWaldmann: hmm, maybe it happened with `exec_cmd(..., fork=True)` when it creates a new subprocess.

pgerber: Seems pretty normal to me. In Python, everything is resolved at run time. So, if you forget an import, you'll only notice once it can't find the corresponding module:

```
# forgotten import
# import xyz

# some function being tested
def func():
    # Oops, can't find xyz
    xyz.create()
```

ThomasWaldmann:
Thanks!

PR #7324, opened 2023-02-08
add hashtable stress tests

ThomasWaldmann:
Using NSIndex (repo index) HashIndex, but they all are very similar.

### DIFF
--- a/src/borg/testsuite/hashindex_stress.py
+++ b/src/borg/testsuite/hashindex_stress.py
@@ -8,9 +8,14 @@ from ..hashindex import NSIndex
 
 @pytest.mark.skipif("BORG_TESTS_SLOW" not in os.environ, reason="slow tests not enabled, use BORG_TESTS_SLOW=1")
 def test_hashindex_stress():
-    """checks if the hashtable behaves as expected"""
+    """checks if the hashtable behaves as expected
+
+    This can be used in _hashindex.c before running this test to provoke more collisions (don't forget to compile):
+    #define HASH_MAX_LOAD .99
+    #define HASH_MAX_EFF_LOAD .999
+    """
     ENTRIES = 10000
-    LOOPS = 100
+    LOOPS = 1000
     idx = NSIndex()
     kv = {}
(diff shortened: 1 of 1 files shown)


jdchristensen:
Nice!  Does this code find any issues?

src/borg/testsuite/hashindex_stress.py
```diff
@@ -0,0 +1,37 @@
...
+from ..hashindex import NSIndex
+
+
+@pytest.mark.skipif("BORG_TESTS_SLOW" not in os.environ, reason="slow tests not enabled, use BORG_TESTS_SLOW=1")
+def test_hashindex_stress():
+    """checks if the hashtable behaves as expected
+
+    This can be used in _hashindex.c before running this test to provoke more collisions (don't forget to compile):
+    #define HASH_MAX_LOAD .99
+    #define HASH_MAX_EFF_LOAD .999
```
jdchristensen, line 15: Is it also possible to start with a smaller hash table, to cause resizing to happen more often?

ThomasWaldmann: It didn't find any issues yet. Good idea, I'll make sure that it uses a small ht size.

ThomasWaldmann: The default for capacity is 0, see hashindex.pyx -> IndexBase`, so we cover the worst case already I guess.

ThomasWaldmann: BTW, I ran it >10 times already on my laptop, but I didn't increase params in a way yet so it'll run for days...

ThomasWaldmann: The latter might also need some precautions so the hashtables can't grow too big. It's pretty random now.

jdchristensen: With capacity = 0, it starts with the smallest hash table size, which is 1031.  In the first loop, the test adds 10000 entries, so it will do at least four resizings in the first loop.  So this looks good to me.

Issue #5899, opened 2021-07-05
Segfault with borg 1.1.16 on Alpine Linux


rkunze:
## Have you checked borgbackup docs, FAQ, and open Github issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).
borg 1.1.16

#### Operating system (distribution) and version.
Alpine Linux "edge" (rolling release)

#### Hardware / network configuration, and filesystems used.

* Borgbackup metadata (cache and configuration) is in a btrfs partition on the host machine
* Data to back up (LXC/LXD container metadata and filesystems) is ext4 on Ceph RBD block devices 

#### How much data is handled by borg?
Original data: ~ 500GB (filesystems and metadata of ~20 containers)
After deduplication: ~10GB per backup run in 20 archives (1 archive per container)
Total (all archives): ~3TB

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

I use borgbackup to take daily backups of LXC/LXD containers on a dedicated container host. Each backup run calls 
```
borg create "$ARCHIVE_URL::$CONTAINER-{now:%Y-%m-%dT%H:%M:%S}" . --stats
```
in the toplevel directory of a snapshot of the container file system (snapshots are created outside of borgbackup using Ceph tools). `$ARCHIVE_URL` is the SSH URL of an archive on a remote borgbackup server, and `$CONTAINER` is the container ID.

The first call to `borg create` in this backup run fails with a segfault on about every second run (see below for the backtrace). 

The subsequent calls to back up different containers in the same run have had no errors so far. 

The segfault is not specific to one container: If I change the order of the containers for backup I get the same error for a different container - but always for the first `borg create` call of a backup run.

#### Include any warning/errors/backtraces from the system logs

The crash occurs always in the same line in python code:

```
Fatal Python error: Segmentation fault

Current thread 0x00007fe4c6126b48 (most recent call first):
  File "/usr/lib/python3.9/site-packages/borg/cache.py", line 740 in write_archive_index
  File "/usr/lib/python3.9/site-packages/borg/cache.py", line 736 in fetch_and_build_idx
  File "/usr/lib/python3.9/site-packages/borg/cache.py", line 824 in create_master_idx
  File "/usr/lib/python3.9/site-packages/borg/cache.py", line 870 in sync
  File "/usr/lib/python3.9/site-packages/borg/cache.py", line 474 in __init__
[...]

PR #7325, opened 2023-02-08
hashindex bugfix and refactoring (master)

ThomasWaldmann:
no critical fixes yet, one minor bugfix (guess we want to backport it to 1.2).

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -225,6 +225,8 @@ hashindex_resize(HashIndex *index, int capacity)
             return 0;
         }
     }
+    assert(index->num_entries == new->num_entries);
+
     hashindex_free_buckets(index);
     index->buckets = new->buckets;
     index->num_buckets = new->num_buckets;
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -731,23 +731,23 @@ static int
 hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
 {
     int start_idx;
-    int idx = hashindex_lookup(index, key, &start_idx);
+    int idx = hashindex_lookup(index, key, &start_idx);  /* if idx < 0: start_idx -> EMPTY or DELETED */
     uint8_t *ptr;
     if(idx < 0)
     {
         if(index->num_entries > index->upper_limit) {
+            /* hashtable too full, grow it! */
             if(!hashindex_resize(index, grow_size(index->num_buckets))) {
                 return 0;
             }
-            start_idx = hashindex_index(index, key);
+            /* we have just built a fresh hashtable and removed all tombstones,
+             * so we only have EMPTY or USED buckets, but no DELETED ones any more.
+             */
+            idx = hashindex_lookup(index, key, &start_idx);
+            assert(idx < 0);
+            assert(BUCKET_IS_EMPTY(index, start_idx));
         }
         idx = start_idx;
-        while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
-            idx++;
-            if (idx >= index->num_buckets){
-                idx -= index->num_buckets;
-            }
-        }
         if(BUCKET_IS_EMPTY(index, idx)){
             index->num_empty--;
             if(index->num_empty < index->min_empty) {
@@ -758,14 +758,15 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
                 /* we have just built a fresh hashtable and removed all tombstones,
                  * so we only have EMPTY or USED buckets, but no DELETED ones any more.
                  */
-                idx = start_idx = hashindex_index(index, key);
-                while(!BUCKET_IS_EMPTY(index, idx)) {
-                    idx++;
-                    if (idx >= index->num_buckets){
-                        idx -= index->num_buckets;
-                    }
-                }
+                idx = hashindex_lookup(index, key, &start_idx);
+                assert(idx < 0);
+                assert(BUCKET_IS_EMPTY(index, start_idx));
+                idx = start_idx;
             }
+        } else if(BUCKET_IS_DELETED(index, idx)) {
+            /* as expected, nothing to do */
+        } else {
+            assert(0);  /* bucket is full, must not happen! */
         }
         ptr = BUCKET_ADDR(index, idx);
         memcpy(ptr, key, index->key_size);
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -258,16 +258,10 @@ int get_min_empty(int num_buckets){
 }
 
 int size_idx(int size){
-    /* find the hash_sizes index with entry >= size */
-    int elems = NELEMS(hash_sizes);
-    int entry, i=0;
-    do{
-        entry = hash_sizes[i++];
-    }while((entry < size) && (i < elems));
-    if (i >= elems)
-        return elems - 1;
-    i--;
-    return i;
+    /* find the smallest hash_sizes index with entry >= size */
+    int i = NELEMS(hash_sizes) - 1;
+    while(i >= 0 && hash_sizes[i] >= size) i--;
+    return i + 1;
 }
 
 int fit_size(int current){
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -188,16 +188,14 @@ hashindex_lookup(HashIndex *index, const unsigned char *key, int *start_idx)
         if (idx >= index->num_buckets) {  /* triggers at == already */
             idx = 0;
         }
-        if(idx == start) {
-            /* We have done a full pass over all buckets.
-             * - We did not find a bucket with the key we searched for.
-             * - We did not find an empty bucket either.
-             * So all buckets are either full or deleted/tombstones.
-             * This is an invalid state we never should get into, see
-             * upper_limit and min_empty.
-             */
-            assert(0);  /* should never happen - something is wrong here. */
-        }
+        /* When idx == start, we have done a full pass over all buckets.
+         * - We did not find a bucket with the key we searched for.
+         * - We did not find an empty bucket either.
+         * So all buckets are either full or deleted/tombstones.
+         * This is an invalid state we never should get into, see
+         * upper_limit and min_empty.
+         */
+        assert(idx != start);
     }
     /* we get here if we did not find a bucket with the key we searched for. */
     if (start_idx != NULL) {
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -763,10 +763,9 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
                 assert(BUCKET_IS_EMPTY(index, start_idx));
                 idx = start_idx;
             }
-        } else if(BUCKET_IS_DELETED(index, idx)) {
-            /* as expected, nothing to do */
         } else {
-            assert(0);  /* bucket is full, must not happen! */
+            /* Bucket must be either EMPTY (see above) or DELETED. */
+            assert(BUCKET_IS_DELETED(index, idx));
         }
         ptr = BUCKET_ADDR(index, idx);
         memcpy(ptr, key, index->key_size);
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -749,8 +749,7 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
         }
         idx = start_idx;
         if(BUCKET_IS_EMPTY(index, idx)){
-            index->num_empty--;
-            if(index->num_empty < index->min_empty) {
+            if(index->num_empty <= index->min_empty) {
                 /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
                 if(!hashindex_resize(index, index->num_buckets)) {
                     return 0;
@@ -763,6 +762,7 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
                 assert(BUCKET_IS_EMPTY(index, start_idx));
                 idx = start_idx;
             }
+            index->num_empty--;
         } else {
             /* Bucket must be either EMPTY (see above) or DELETED. */
             assert(BUCKET_IS_DELETED(index, idx));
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
based on #7324, merge after that.

ThomasWaldmann:
BTW, I noticed that one of these loops I removed was in 1.2-maint, but not in 1.1-maint.

src/borg/_hashindex.c
```diff
@@ -731,23 +731,23 @@ static int
...
         idx = start_idx;
-        while(!BUCKET_IS_EMPTY(index, idx) && !BUCKET_IS_DELETED(index, idx)) {
-            idx++;
-            if (idx >= index->num_buckets){
-                idx -= index->num_buckets;
-            }
-        }
         if(BUCKET_IS_EMPTY(index, idx)){
             index->num_empty--;
             if(index->num_empty < index->min_empty) {
```
jdchristensen, line 753: I think it would be logically cleaner to move the line `index->num_empty--` to later in this `if` block, so that we aren't calling `hashindex_resize` and `hashindex_lookup` on a hash table with incorrect metadata.  The line `if(index->num_empty < index->min_empty)` would then change to `<=`.

src/borg/_hashindex.c
```diff
@@ -758,14 +758,15 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
...
-                }
+                idx = hashindex_lookup(index, key, &start_idx);
+                assert(idx < 0);
+                assert(BUCKET_IS_EMPTY(index, start_idx));
+                idx = start_idx;
             }
+        } else if(BUCKET_IS_DELETED(index, idx)) {
+            /* as expected, nothing to do */
+        } else {
+            assert(0);  /* bucket is full, must not happen! */
```
jdchristensen, line 769: How about the simpler code:
```
else {
    assert(BUCKET_IS_DELETED(index, idx));
}
```

jdchristensen (approved):

jdchristensen:
@ThomasWaldmann I started my review before you force-pushed.  What did you change in the force push?

ThomasWaldmann:
@jdchristensen i did rebase it on fresh master (after i merged the ht stresstest to master), no other changes.

jdchristensen:
Ok, LGTM.  I checked the logic fairly carefully.  `hashindex_lookup` could use some additional comments and assertions as well, and maybe some of the other related code too.

jdchristensen:
I didn't notice any bugs in the previous code, but I agree that the new code is cleaner and easier to understand.

jdchristensen (approved):
LGTM.  Despite what I wrote before, I don't see any obvious assertions one could add.

src/borg/_hashindex.c
```diff
@@ -187,10 +189,16 @@ hashindex_lookup(HashIndex *index, const unsigned char *key, int *start_idx)
             idx -= index->num_buckets;
         }
         if(idx == start) {
+            /* we have done a full pass over all buckets. */
             break;
         }
```
ThomasWaldmann, line 194: As we avoid a full HT, this likely never happens.

But if it would happen:
- the code would avoid an endless loop by breaking out
- in case there was no deleted/tombstone bucket, it would return `start` in `*start_idx` and that bucket would be **full**. if the caller checks the bucket state, the caller might notice, but maybe we should fail quicker and just right here?

jdchristensen: Good point.  So if didx == -1, abort the program, since this should never happen?  (IIUC, it should also not happen with didx != -1, so we could log a warning in that case?  Or just abort in either case, since it would indicate something seriously wrong?)

jdchristensen: Could this case happen with one of the `LOAD` variables set to .99 or .999 due to rounding issues?

ThomasWaldmann: ^ made another commit with "fail early".

ThomasWaldmann: had to force push again to fix wording / typos in commit comment, no other changes.

ThomasWaldmann: ok, i've implemented all of this.

src/borg/_hashindex.c
```diff
@@ -189,8 +189,14 @@ hashindex_lookup(HashIndex *index, const unsigned char *key, int *start_idx)
...
-            break;
+            /* We have done a full pass over all buckets.
+             * - We did not find a bucket with the key we searched for.
+             * - We did not find an empty bucket either.
+             * So all buckets are either full or deleted/tombstones.
+             * This is an invalid state we never should get into, see
+             * upper_limit and min_empty.
+             */
+            assert(0);  /* should never happen - something is wrong here. */
         }
```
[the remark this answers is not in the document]
ThomasWaldmann: well, due to 191 that is already known to be false.

jdchristensen:
I also noticed that the `size_idx` function is overly convoluted.  This is shorter and easier to reason about.
```
int size_idx(int size){
    /* find the smallest hash_sizes index with entry >= size */
    int i = NELEMS(hash_sizes) - 1;
    while(i >= 0 and hash_sizes[i] >= size) i--;
    return i+1;
}
```

src/borg/_hashindex.c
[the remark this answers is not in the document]
ThomasWaldmann: ah, you mean removing the "if". yes, indeed, i'll fix it.

jdchristensen:
I've checked over all of the new commits.

src/borg/_hashindex.c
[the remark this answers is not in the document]
ThomasWaldmann: done.

ThomasWaldmann:
`hashindex_compact` might be also good for a review.

src/borg/_hashindex.c
ThomasWaldmann: This actually fixes a bug from the old code.

Old code:
Considering we will use up one empty slot in below code, we first decremented `num_empty`.
Then, hashindex_resize() internally **overwrote** num_empty (recomputing it from `num_buckets - num_entries`). Because `num_entries` is not incremented yet, that resulted in a wrong value (we lost the decrement operation).

jdchristensen: Nice observation!  I knew something was fishy about that code.  I doubt it leads to any data loss, though.

While looking at hashindex_resize, I thought that it would be good to add an assertion:
```
assert(index->num_entries == new->num_entries);
```

ThomasWaldmann:
Another force-push and rebase on master to add "bugfix" and more information to the respective commit.

ThomasWaldmann:
OK, guess this is finished.

src/borg/_hashindex.c
ThomasWaldmann: It only uses the "new" buckets table and points to that from the "old" data structure.
Besides that, it continues to work with the old data structure and just selectively copies some stuff from "new" (but not num_entries).

But yes, we could check if the copying loop results in consistent num_entries values.

ThomasWaldmann: done.

jdchristensen:
I don't fully understand the hashindex_compact code, and I find it very hard to reason about.  There's a much simpler version of the code that accomplishes the same thing and is clearly correct. You search for empty buckets from the start of the hash table and search for filled buckets from the end, and move them to the empty spot.  You stop when the end pointer is < num_entries.  This leaves in place most existing buckets, which means it would usually be faster, even though it only moves one bucket at a time.  If there is one free spot at the front, I think the current code moves every bucket one step forward, one bucket at a time.  But, IIUC, we don't care about the order, so moving from the end can do this in one step. 

Even more, I don't think it's a good idea to shuffle the buckets around in RAM.  In the last eight comments of #5899, we briefly discussed instead just looping over the hash table and writing out the data, which would also be clearly correct (and would be nondestructive).

If hashindex_compact is kept, maybe the new test should call it at the end and then repeat the check that the contents are still correct?

ThomasWaldmann:
I also found the code hard to understand. After a while, I thought it is correct.

But your algorithm sounds much simpler. I'll check what tests we have for that / what we could do there.

jdchristensen:
> BTW, I noticed that one of these loops I removed was in 1.2-maint, but not in 1.1-maint.

I don't see this.  I did
```
git diff 1.1-maint 1.2-maint -- src/borg/_hashindex.c
```
and it shows no changes to `hashindex_set`, except for the argument pointed types.  But maybe my git skills are failing me here.

ThomasWaldmann:
Looks like I used git diff wrong. I had `1.1-maint...1.2-maint` in there.

ThomasWaldmann:
I'll merge this now and might open another PR to the hashindex_compact changes.

jdchristensen:
Here's simpler code for hashindex_compact:
```
#define BUCKET_IS_EMPTY_OR_DELETED(index, idx) (BUCKET_IS_EMPTY(index, idx) || BUCKET_IS_DELETED(index, idx))

/* Move all non-empty/non-deleted entries in the hash table to the beginning. */
static uint64_t
hashindex_compact(HashIndex *index)
{
    int idx = index->num_buckets;
    int tail = 0;
    uint64_t saved_size = (index->num_buckets - index->num_entries) * (uint64_t)index->bucket_size;

    /* idx will point to the last filled spot and tail will point to the first empty spot. */
    for(;;) {
	/* Find the last filled spot >= index->num_entries. */
        while((idx >= index->num_entries) && BUCKET_IS_EMPTY_OR_DELETED(index, idx)) {
            idx--;
        }
	/* If all spots >= index->num_entries are empty, then we must be in a compact state. */
        if(idx < index->num_entries) {
            return saved_size;
        }
	/* Find the first empty spot >= index->num_entries. */
        while((tail < index->num_entries) && !BUCKET_IS_EMPTY_OR_DELETED(index, tail)) {
            tail++;
        }
	assert(tail < index->num_entries);
        memcpy(BUCKET_ADDR(index, tail), BUCKET_ADDR(index, idx), index->bucket_size);
	idx--;
	tail++;
    }
}
```
It builds, but I haven't tested it.  I won't have time for a day or two, so feel free to use it or adapt it.

jdchristensen:
It should include
```
            index->num_buckets = index->num_entries;
	    index->num_empty = 0;
```
before the `return` statement.

jdchristensen:
It looks like the testsuite assumes that the entries are not reordered when compacting...

ThomasWaldmann:
I've opened a new PR with a new problem. :-)

jdchristensen:
Here's a version that preserves the order.  It passes the hashindex testsuite.  It's simpler and probably faster than the existing code, but not as simple and fast as the code posted above which changes the order.  I also put `#undef NDEBUG` as the first line of the file, and the assertion did catch a typo.
```
#define BUCKET_IS_EMPTY_OR_DELETED(index, idx) (BUCKET_IS_EMPTY(index, idx) || BUCKET_IS_DELETED(index, idx))

/* Move all non-empty/non-deleted entries in the hash table to the beginning.
 * The unit tests currently assume that the order is preserved, so we do so.
 */
static uint64_t
hashindex_compact(HashIndex *index)
{
    int tail;
    int start;
    int end;
    int buckets_to_copy;
    uint64_t saved_size = (index->num_buckets - index->num_entries) * (uint64_t)index->bucket_size;

    /* Find first empty/deleted bucket. */
    tail = 0;
    while((tail < index->num_buckets) && !BUCKET_IS_EMPTY_OR_DELETED(index, tail))
	tail++;
    start = tail + 1;
    /* tail points to an empty/deleted bucket, or to one past the end. */
    /* start points to the first place to start looking for used buckets. */

    for(;;) {
	/* Find the first used bucket after tail. */
	while((start < index->num_buckets) && BUCKET_IS_EMPTY_OR_DELETED(index, start))
	    start++;
	/* No used buckets after tail.  Also handles when tail points past the end. */
	if(start >= index->num_buckets)
	    break;

	/* Find the end of the segment of used buckets. */
	end = start + 1;
	while((end < index->num_buckets) && !BUCKET_IS_EMPTY_OR_DELETED(index, end))
	    end++;
	/* end points to one past the end of the segment. */
	buckets_to_copy = end - start;
	/* The regions may overlap, so use memmove. */
	memmove(BUCKET_ADDR(index, tail), BUCKET_ADDR(index, start),
		buckets_to_copy * index->bucket_size);
	tail = tail + buckets_to_copy;
	start = end;
	/* Now tail points to a bucket that is *logically* unused, but may be marked as used.
         * start takes into account the known logically unused buckets, and points to the
         * first bucket after the segment we just moved. */
    }
    assert(tail == index->num_entries);

    index->num_buckets = index->num_entries;
    index->num_empty = 0;
    index->min_empty = 0; /* Do we want this? */
    index->upper_limit = index->num_entries; /* And this? */
    return saved_size;
}
```

jdchristensen:
I'm happy to make a PR if that's easier for you, but I'm not sure which approach is best to use.

ThomasWaldmann:
Well, shouldn't we rather take the simpler algorithm? The changed order might be irrelevant and just a "feature" of the tests?

Usually, when we loaded such a compact hashtable, the first thing that happens is that it will be resized/rebuilt, which does not change the contents, but the layout of the hashtable.

jdchristensen:
I agree that the simpler algorithm is better.  Maybe we can adapt the tests to compare hashtables using something like the code you added recently, which loops over all entries.

jdchristensen:
I can probably make a PR on the weekend, but feel free to beat me to it.  If you do, I'll review it.

jdchristensen:
The first thing I'd like to do is to change the tests so that they don't depend on the order of the entries in the hash table.  What's the quickest way to re-run a specific test after changing it?  Do I need to do a full `pip install -e .` and then run `tox` (limited to that testsuite), or is there a faster way?

ThomasWaldmann:
I usually do:
```
python setup.py clean clean2 ; pip install -e .   # clean / recompile / reinstall everything (after .c / .pyx changes)
pytest -vv --benchmark-skip -k "not remote and sometestname"
```

ThomasWaldmann:
Without the "clean/clean2" it might be a bit faster and hopefully still compiles all necessary stuff.

jdchristensen:
If only .py files are changed (e.g. testsuite/hashindex.py), can I just run the pytest line?

ThomasWaldmann:
@jdchristensen if you used `pip install -e .`:  yes.

PR #7328, opened 2023-02-09
hashindex minor fixes, tweaks, tests

ThomasWaldmann:


src/borg/testsuite/hashindex_pytest.py
```diff
@@ -56,3 +56,13 @@ def test_hashindex_compact():
     for k, v in kv.items():
         assert k in idx and idx[k] == (v, v, v)
     assert len(idx) == len(kv)
+    # now expand the hashtable again. trigger a resize/rebuild by adding an entry.
+    k = b"x"*32 # list(kv)[0]
+    idx[k] = (0, 0, 0)
```
ThomasWaldmann, line 61: hangs here. no output, no boom, no ctrl-c. why?

i searched the code for NDEBUG, but it is used nowhere, so shouldn't the assert already do something?

why doesn't even the null pointer deref terminate the program?

ThomasWaldmann: also:
- there is no empty or deleted bucket in the compacted hashtable
- there is no matching key in the hashtable

thus i am rather sure we should run into the "BOOM!".

src/borg/_hashindex.c
```diff
@@ -196,6 +196,7 @@ hashindex_lookup(HashIndex *index, const unsigned char *key, int *start_idx)
          * upper_limit and min_empty.
          */
         assert(idx != start);
+        if(idx == start) *(char *)NULL;  // BOOM!
```
ThomasWaldmann, line 199: hmm, does the compiler optimize that away?

ThomasWaldmann: also tried: `*(char *)NULL = 0;` - nothing happens, just hangs.

ThomasWaldmann: but still the `assert(false);` does not abort the process.

jdchristensen: But the assertion doesn't cause an error?  That's annoying.

In any case, I didn't expect _lookup or _set to be called on a compacted hash_table.  I guess we'll need to put back the test for idx == start, and break out of the loop if that happens.

src/borg/testsuite/hashindex_pytest.py
ThomasWaldmann: it now segfaults, so it gets there.

ThomasWaldmann:
Hmm, guess i can either:
- make it work with "compact" hashtables by removing the BOOM and return something special (indicating there is no empty/deleted bucket)
- add an empty bucket to the otherwise compact hashtable

src/borg/_hashindex.c
ThomasWaldmann: Yeah, i do return with -2 in that case now.

assert() is documented to log a warning to stderr and call abort() to terminate the process.
the first i maybe don't see due to pytest, but the latter does not happen.

ThomasWaldmann: BTW, i used adding a new element to the compact hashtable only as a way to trigger a resize/rebuild (that API is not available from Python otherwise).

jdchristensen (commented):
The tests look good!

src/borg/_hashindex.c
horazont: Are you building with -DNDEBUG? That would remove assertions.

ThomasWaldmann: @horazont I searched for NDEBUG in the project, but couldn't find anything.

src/borg/_hashindex.c
```diff
@@ -745,16 +748,16 @@ hashindex_set(HashIndex *index, const unsigned char *key, const void *value)
     uint8_t *ptr;
     if(idx < 0)
     {
-        if(index->num_entries > index->upper_limit) {
-            /* hashtable too full, grow it! */
+        if(index->num_entries > index->upper_limit || idx == -2) {
```
jdchristensen, line 751: Since we'll be adding 1 to num_entries below, the condition should be `>=`:
```suggestion
        if(index->num_entries >= index->upper_limit || idx == -2) {
```

jdchristensen:
It turns out that NDEBUG *is* set while _hashindex.c is being compiled.  I put
```
#if defined(NDEBUG)
foo bar fjdsal ;
#endif
```
near the top of the file, and the compiler complains about that line.  If I change it go NDEBUGXX, then it doesn't complain.

Maybe this is relevant?  https://stackoverflow.com/questions/50740518/remove-a-default-compiler-option-when-building-with-setup-py

ThomasWaldmann:
Thanks for finding that, I just put the NDEBUG stuff into #7329.

src/borg/_hashindex.c
ThomasWaldmann: See #7329.

src/borg/_hashindex.c
```diff
@@ -195,7 +195,8 @@ hashindex_lookup(HashIndex *index, const unsigned char *key, int *start_idx)
          * This is an invalid state we never should get into, see
          * upper_limit and min_empty.
          */
-        assert(idx != start);
+        if(idx == start)
+            return -2;
```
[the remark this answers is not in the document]
ThomasWaldmann: sounds good, i pushed some new commits with that.

ThomasWaldmann:
Guess after tests and review for this is finished, I'll squash some of the changesets and merge this.

The NDEBUG stuff and "new hashindex_compact algorithm" should be in separate PRs.

src/borg/_hashindex.c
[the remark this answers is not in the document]
ThomasWaldmann: btw, i thought about removing the `idx == -2` term from the check.

if all goes well, it is not needed, because the first term will already trigger the resize.
but, if that would not trigger due to wrong values, it is still useful to trigger based on -2 return value.

thus, i kept it.

ThomasWaldmann:
rebase onto current master (includes building without NDEBUG), collapsed some commits.

PR #7334, opened 2023-02-10
allow mixing --progress and --list in log-json mode

ThomasWaldmann:


ThomasWaldmann:
@m3nu does vorta use borg with all these options together?

Before this PR, I just disallowed `--list` and `--progress`, because I was getting tired of people mixing them and complaining that it doesn't look pretty (on screen, in their logs).

I tried it with `--log-json` now and looked at the json output, but didn't see any "--progress json", just "--list json" (for `borg create`).

So, for what does vorta use this?

m3nu:
You mean `--progress` is redundant for `borg create`? Seems right. I just commented it out and still got the list of files being processed. We also use `--progress` for `borg check` and `borg compact`. I'm guessing this is the intended usage.

The goal is to show the files being processed. Which `--list` seems to do just as well.

As for why it's there: It was added by an external contributor while improving logging: https://github.com/borgbase/vorta/commit/02625ba4f01bb500b1f580068c1a3e1ce8a4583b#diff-2559f308260b3597027f7b0650bc98f2783fbb1ed513a244a3ccb4756256d11c via https://github.com/borgbase/vorta/pull/611

ThomasWaldmann:
@m3nu guess the other commands need checking, but for `create` it doesn't add more json output.

real-yfprojects:
> I tried it with --log-json now and looked at the json output, but didn't see any "--progress json", just "--list json" (for borg create).
>
> So, for what does vorta use this?


Borg will only emit progress messages if the time needed to backup a file surpasses a certain threshold. Vorta uses the `archive_progress` messages for displaying the the live size and deduplicated size of the archived files while `file_status` is used to show the files processed. Vorta could only use `archive_progress` if it would be emitted for every file and contained an attribute showing whether a file is added or modified.

ThomasWaldmann:
@real-yfprojects ah, yes, guess I didn't consider the timing in my experiment.

But with the code in this PR, I guess vorta's way to call borg should work again and not get rejected.

PR #7335, opened 2023-02-10
Modified Item.pyx to include diffs in ctime and mtime

Michael-Girma:
Modified Item.pyx to include diffs in ctime and mtime
Created flag to switch this feature on/off


Michael-Girma:
@ThomasWaldmann could use some help with the help text for the diff cmd.

src/borg/item.pyx
```diff
@@ -736,6 +739,16 @@ class ItemDiff:
             mode2 = stat.filemode(self._item2.mode)
             return ({"type": "mode", "old_mode": mode1, "new_mode": mode2}, '[{} -> {}]'.format(mode1, mode2))
 
+    def _time_diffs(self):
+        changes = []
+        attrs = ["ctime", "mtime"]
```
ThomasWaldmann, line 744: Sometimes there is also "birthtime", but not sure if we should check it.

Guess we better not check "atime", it changes often and is not that interesting.

src/borg/item.pyx
```diff
@@ -736,6 +739,16 @@ class ItemDiff:
...
+        for attr in attrs:
+            if attr in self._item1 and attr in self._item2 and self._item1.get(attr) != self._item2.get(attr):
```
ThomasWaldmann, line 746: you could use the new walrus operator here.

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -218,10 +219,34 @@ def get_changes(filename, data):
...
+        do_json_asserts(output, True)
+
+    def test_time_diffs(self):
+        self.create_regular_file("test_file", size=10)
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive1", "input")
+        sleep(1)
+        os.unlink("input/test_file")
+        self.create_regular_file("test_file", size=10)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive2", "input", exit_code=0)
```
ThomasWaldmann, line 241: isn't `exit_code=0` the default anyway?

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -218,10 +219,34 @@ def get_changes(filename, data):
...
+        output = self.cmd(
+            f"--repo={self.repository_location}", "diff", "test0", "test1a", "--json-lines", "--content-only"
+        )
+        do_json_asserts(output, True)
+
+    def test_time_diffs(self):
+        self.create_regular_file("test_file", size=10)
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive1", "input")
+        sleep(1)
```
ThomasWaldmann, line 238: Save a second and use `0.01`?

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -218,10 +219,34 @@ def get_changes(filename, data):
...
+        os.unlink("input/test_file")
+        self.create_regular_file("test_file", size=10)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive2", "input", exit_code=0)
+        output = self.cmd(f"--repo={self.repository_location}", "diff", "archive1", "archive2")
+        self.assert_in("mtime", output)
+        self.assert_in("ctime", output)
+        os.chmod("input/test_file", 777)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive3", "input", exit_code=0)
```
ThomasWaldmann, line 246: same.

src/borg/testsuite/archiver/__init__.py
```diff
@@ -195,6 +195,11 @@ def create_regular_file(self, name, size=0, contents=None):
         filename = os.path.join(self.input_path, name)
         if not os.path.exists(os.path.dirname(filename)):
             os.makedirs(os.path.dirname(filename))
+        self.write_to_file(filename, size=size, append_path=False, contents=contents)
+
+    def write_to_file(self, filename, size=0, append_path=True, contents=None):
+        if append_path:
+            filename = os.path.join(self.input_path, filename)
```
ThomasWaldmann, line 202: that's a good idea to save the callers from having to prepend the input path.

but guess that should be:
- refactored globally in a separate PR
- done by default for `create_regular_file`, not adding an extra method

src/borg/item.pyx
```diff
@@ -656,6 +656,9 @@ class ItemDiff:
             changes.append(self._owner_diff())
             changes.append(self._mode_diff())
 
+        if not content_only:
+            changes.extend(self._time_diffs())
+
```
ThomasWaldmann, line 661: Guess this need a clarification about what this feature should do exactly.

The way you implemented it is **without comparing timestamps**.

But the feature name "content-only" (== "data-only") rather suggests **without comparing metadata**.
See the lines above which compare owner and mode...

ThomasWaldmann: After this is decided, the help text needs adjustment.

src/borg/archiver/diff_cmd.py
```diff
@@ -19,7 +20,7 @@ def do_diff(self, args, repository, manifest, archive):
         """Diff contents of two archives"""
 
         def print_json_output(diff, path):
-            print(json.dumps({"path": path, "changes": [j for j, str in diff]}))
+            print(json.dumps({"path": path, "changes": [j for j, str in diff]}, sort_keys=True, cls=BorgJsonEncoder))
```
ThomasWaldmann, line 23: the sort_keys change seems unrelated? separate commit with a good comment?

src/borg/helpers/parseformat.py
```diff
@@ -882,7 +882,7 @@ def __init__(self, archive, format, *, json_lines=False):
         self.used_call_keys = set(self.call_keys) & self.format_keys
 
     def format_item_json(self, item):
-        return json.dumps(self.get_item_data(item), cls=BorgJsonEncoder) + "\n"
+        return json.dumps(self.get_item_data(item), cls=BorgJsonEncoder, sort_keys=True) + "\n"
```
ThomasWaldmann, line 885: same.

src/borg/item.pyx
```diff
@@ -672,7 +675,7 @@ class ItemDiff:
         if self._item1.get('deleted') and self._item2.get('deleted'):
             return True
 
-        attr_list = ['deleted', 'mode', 'target']
+        attr_list = ['deleted', 'mode', 'target', 'ctime', 'mtime']
```
ThomasWaldmann, line 678: this seems to be done unconditionally, not checking `content_only`.

ThomasWaldmann (changes requested):
Thanks for the PR, some stuff I found...

src/borg/testsuite/archiver/__init__.py
ThomasWaldmann: hmm, considering that tests often check like `assert "input/somefile" in output` it is maybe not a good idea to remove this "input" part at the caller location.

ThomasWaldmann:
BTW, some tests fail in a not very informative way. Guess this could be improved in a separate commit.

ThomasWaldmann:
Would like to include this in next borg2 beta **if** it can be finished soon.

@Michael-Girma are you still working on this?

src/borg/helpers/parseformat.py
Michael-Girma: @ThomasWaldmann the enhancement request in the original issue ticket asked that the order of the response be consistent in terms of order (preferably alphabetically ordered). That's why I added those two in.

src/borg/item.pyx
Michael-Girma: I don't think the Cython version we're using supports the use of walrus operators.

ThomasWaldmann: https://github.com/cython/cython/pull/3691 seems to be in master / cython 3.0 alpha only.

src/borg/helpers/parseformat.py
ThomasWaldmann: OK, but please mention this change in the commit comment and also please add ", fixes #XXXX" to refer back to the ticket you are fixing.

Michael-Girma:
@ThomasWaldmann pushed changes. Take a look when possible.

src/borg/testsuite/archiver/__init__.py
```diff
@@ -190,16 +190,20 @@ def open_archive(self, name):
...
         return Repository(self.repository_path, exclusive=True)
 
+
+    def assert_line_exists(self, lines, expected_substrs):
+        exists = False
+        for line in lines:
+            if all(substr in line for substr in expected_substrs):
+                exists = True
+                break
+        assert exists
```
ThomasWaldmann, line 199: if that fails, one likely does not get much information about why it failed.

guess it would only say it is False instead of the expected True.

just a rough idea:
```
assert any(re.match(line, match_re) for line in lines)
```
check if that displays the match_re and also lines if it fails.

src/borg/archiver/diff_cmd.py
ThomasWaldmann, line 23: why did you add `cls=BorgJsonEncoder`?

src/borg/archiver/diff_cmd.py
```diff
@@ -105,6 +108,9 @@ def build_parser_diff(self, subparsers, common_parser, mid_common_parser):
         )
         subparser.add_argument("--sort", dest="sort", action="store_true", help="Sort the output lines by file path.")
         subparser.add_argument("--json-lines", action="store_true", help="Format output as JSON Lines. ")
+        subparser.add_argument(
+            "--content-only", action="store_true", help="Only compare differences in content (Exclude time differences)"
```
ThomasWaldmann, line 112: **exclude** (is in the middle of a sentence, no reason to uppercase that).
**timestamp* (more precise than *time*)

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -127,16 +129,16 @@ def do_asserts(output, can_compare_ids):
...
-            if are_hardlinks_supported():
+            if are_hardlinks_supported() and content_only:
                 assert "input/hardlink_target_removed" not in output
 
             # Another link (marked previously as the source in borg) to the
             # same inode was replaced with a new regular file. This should not
             # change this link at all.
-            if are_hardlinks_supported():
+            if are_hardlinks_supported() and content_only:
                 assert "input/hardlink_target_replaced" not in output
```
ThomasWaldmann, line 140: this looks suspicious, we'll have a look at that if all else is settled...

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -208,20 +213,49 @@ def get_changes(filename, data):
                 assert {"type": "removed", "size": 256} in get_changes("input/hardlink_removed", joutput)
 
             # Another link (marked previously as the source in borg) to the
-            # same inode was removed. This should not change this link at all.
-            if are_hardlinks_supported():
-                assert not any(get_changes("input/hardlink_target_removed", joutput))
+            # same inode was removed. This should not change this link except for its ctime mtime.
+            if content_only:
+                if are_hardlinks_supported():
```
ThomasWaldmann, line 218: suspicious, too.

src/borg/archiver/diff_cmd.py
Michael-Girma: That's more accurate.

src/borg/archiver/diff_cmd.py
ThomasWaldmann: no, guess that's fine as is.

src/borg/helpers/parseformat.py
ThomasWaldmann: the commit comment change is still missing.

maybe you can do a `git rebase -i master` at some time and add that (and also collapse some fixup commits).

ThomasWaldmann:
Windows tests are failing, check why. ctime on windows is creation time, not inode change time as on unix.

Michael-Girma:
@ThomasWaldmann Working with the docs but still struggling to set up a windows dev environment using Cygwin.  Is there another preferred method to set it up or at the very least mimic it for the tests?

ThomasWaldmann:
@Michael-Girma the windows CI works with msys2, not sure if test results are same for cygwin.

Maybe @RayyanAnsari can help better with the Windows stuff than me.

RayyanAnsari:
@Michael-Girma yes, on Windows we use MSYS2 for the build environment.
You can check the windows.yml workflow file to see how it works: https://github.com/borgbackup/borg/blob/master/.github/workflows/windows.yml
In short, you need to:
- Install MSYS2
- Open a UCRT64 terminal
- Define this environment variable `export SETUPTOOLS_USE_DISTUTILS=stdlib`
- Install dependencies (using the script)
- Build (using the script)
- Install the package in-place
- Run pytest

@ThomasWaldmann I should probably add this to the documentation, is it just `installation.rst` that I need to add this to?

ThomasWaldmann:
@RayyanAnsari yes, it would be good to have it in the docs.

installation.rst is the end-user stuff and the development section usually builds on that and adds developer instructions.

we have also that old windows todo file in the repo root dir, guess it is outdated?

Michael-Girma:
@ThomasWaldmann. Been puzzled by an issue regarding the ctime attribute on windows for the past couple of days. One thing I landed on is that there is a different behavior for temporary dirs. When creating a file manually, a difference in ```ctime``` is picked up by borg if a file is recreated with the same filename. However, this behavior is not the same if the parent dir is created with ```mktemp -d```. I've attached a screenshot showing this behavior. In the first two tabs, I created the parent dir with ```mkdir``` and used ```mktemp -d``` for the third one. Any thoughts on what might have caused this?


Screenshot_20230301_014753

ThomasWaldmann:
@Michael-Girma that's strange. like it did not delete the file, but continued with the same one (3rd screenshot).

one difference is maybe the fs type you work on:
- home dir, ext4 maybe?
- /tmp, tmpfs maybe?

maybe you could try the same as in 1st screenshot, but manually use /tmp/something.

of course, /tmp is also mode +t ("sticky"), but i don't see how what you found could be related to that.

Michael-Girma:
@ThomasWaldmann , I thought linux as well. That's why I tried a third archive with the file modified. In that scenario, there would have been a ctime in the last diff output since ctime would change on linux after content manipulation.

Michael-Girma:
@ThomasWaldmann after landing on an [issue](https://github.com/joyent/libuv/issues/901) hinting that ```ctime``` isn't refreshed if the file is recreated within 15 seconds after deletion, I tested this locally and found out why we were having the issue. After adding 15 seconds between the unlink and creation, borg seems to be spitting out the ctime diff and the tests seem to be passing.

src/borg/testsuite/archiver/__init__.py
Michael-Girma: Couldn't get it to spit out the re and the lines so I changed assert_line_exists to honor this approach.

ThomasWaldmann:
@Michael-Girma that 15s thing is interesting, like "wtf did they think when implementing it like that?".

the issue you linked to reads rather confusing and i guess we should focus on python `os.stat(fn) -> stat_result`:

https://docs.python.org/3/library/os.html#os.stat_result.st_ctime

So, is that "creation time" (as documented) under msys2? Or "metadata change time" as on Unix?

Also, whatever you find out should be in a source code comment, so it is easier accessible without having to dig into git / github. There should be a comment explaining it right above that sleep(15).

Also: does this explain the different behaviour you found in home and in /tmp (was it just the timing of your manual commands)?

Michael-Girma:
@ThomasWaldmann I know, crazy idea.
In msys2, it is creation time and metadata change time on UNIX. ```os.stat_result``` doesn't have the ```st_birthtime``` attribute on windows but it is populated when running the same code on linux. 

Yeah, I'm also getting ```ctime``` on tmp dirs. Must have been my timing.

Michael-Girma:
@ThomasWaldmann I've added in the comment above sleep and I've also rebased to clean up the commit history, add the issue number and bring in the latest changes from master.

ThomasWaldmann (changes requested):
some more, sometimes minor, stuff i found.

also the hardlink stuff needs a thorough check as borg is handling these a bit differently in master:
- there is no "source" or "target" any more for hardlinks in the borg item (looks like some comments were wrong before you change already, maybe you could check and fix them)
- hardlinks that originally pointed to same inode have the same "hlid" value in the borg item

src/borg/item.pyx
```diff
@@ -652,9 +653,11 @@ class ItemDiff:
...
             changes.append(self._presence_diff('fifo'))
 
-        if not (self._item1.get('deleted') or self._item2.get('deleted')):
-            changes.append(self._owner_diff())
-            changes.append(self._mode_diff())
+        if not self._content_only:
+            if not (self._item1.get('deleted') or self._item2.get('deleted')):
+                changes.append(self._owner_diff())
+                changes.append(self._mode_diff())
+            changes.extend(self._time_diffs())
```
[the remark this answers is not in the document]
Michael-Girma: Yeah, pondered on that for a while and decided to leave that in. It makes it so that we can get the time of deletion/recreation. Do you think we shouldn't do it if either one is deleted?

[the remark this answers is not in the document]
ThomasWaldmann: well, the deleted item has nothing inside because there was nothing in the filesystem.
so we can't see the time of deletion, only the time of creation IF one compares the right archives or the fs supports creation time.

but one could see the timestamps of the other item.
as long as this is displayed in a clear way we could keep it, but it is somehow similar to owner and mode besides the aspect of timing.

Michael-Girma:
@ThomasWaldmann Checked the hard-link changes. So what the previous test case was trying to do is create 2 hard-links and then remove one of them. It was expected that this shouldn't change the link that hasn't been removed in any way, which was right since we weren't considering ```ctime``` and ```mtime```. The difference now is that we are taking these two fields into account. The decrementation on the hard link count in the inode registers as metadata change thus changing the ```ctime```. This means that it's gonna show up on the diff unless we are checking with ```--content-only```. Will add a comment to the test case to explain it.

ThomasWaldmann:
Ah, good explanation, yes, please add some short comment in the src.

src/borg/item.pyx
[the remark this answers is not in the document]
ThomasWaldmann: isn't 751 False and the function call not generating any output if one of the items is deleted?

if that is the case, 661 could be also indented to not get executed if one item  is deleted.

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -170,14 +172,17 @@ def get_changes(filename, data):
...
+                # TODO: Add mode test flag to assert function
+                if not content_only:
+                    assert any(
+                        chg["type"] == "mode" and chg["new_mode"].startswith("l")
+                        for chg in get_changes("input/dir_replaced_with_link", joutput)
+                    ), get_changes("input/dir_replaced_with_link", joutput)
+                    assert any(
+                        chg["type"] == "mode" and chg["old_mode"].startswith("l")
+                        for chg in get_changes("input/link_replaced_by_file", joutput)
+                    )
```
ThomasWaldmann, line 185: did you forget to add ", get_changes(...)" to the 2nd assert?

src/borg/testsuite/archiver/diff_cmd.py
ThomasWaldmann: How about:
```
            if are_hardlinks_supported() and content_only:
```

src/borg/item.pyx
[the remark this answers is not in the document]
Michael-Girma: Right, in that case it would be better to only call it inside the if clause.

ThomasWaldmann (approved):
LGTM.

src/borg/testsuite/archiver/diff_cmd.py
```diff
@@ -170,14 +172,17 @@ def get_changes(filename, data):
...
-                assert any(
-                    chg["type"] == "mode" and chg["new_mode"].startswith("l")
-                    for chg in get_changes("input/dir_replaced_with_link", joutput)
-                )
-                assert any(
-                    chg["type"] == "mode" and chg["old_mode"].startswith("l")
-                    for chg in get_changes("input/link_replaced_by_file", joutput)
-                )
+
+                # TODO: Add mode test flag to assert function
```
ThomasWaldmann, line 176: What exactly do you mean by that?

src/borg/testsuite/archiver/diff_cmd.py
Michael-Girma: It was a TODO to remove after the second commit. Will remove that.

src/borg/testsuite/archiver/diff_cmd.py
Michael-Girma: This was to be removed before the second commit. Will remove.

Michael-Girma: Amended last commit.

ThomasWaldmann (approved):

ThomasWaldmann:
Thanks!

ThomasWaldmann:
If you like, check how easy/hard this is to backport to 1.2-maint.
We could add it for release in 1.2.4 or it has to wait for borg2 depending on that.

Michael-Girma:
Okay will do!

real-yfprojects:
@ThomasWaldmann The [`All about JSON`](https://borgbackup.readthedocs.io/en/stable/internals/frontends.html#archive-differencing) page needs an update because of this.

PR #7344, opened 2023-02-11
parseformat: remove Windows-specific hack

RayyanAnsari:
This fixes the rest of the non-remote tests.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -472,16 +472,6 @@ class Location:
             p = os.path.normpath(p)
             return ("/." + p) if relative else p
 
-        if is_win32:
-            m = self.win_file_re.match(text)
-            if m:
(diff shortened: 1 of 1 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -472,16 +472,6 @@ def normpath_special(p):
             p = os.path.normpath(p)
             return ("/." + p) if relative else p
 
-        if is_win32:
-            m = self.win_file_re.match(text)
```
ThomasWaldmann, line 476: what's about self.win_file_re now? do we keep it for later?

RayyanAnsari: yes, it will be useful to fix the remote tests.

ThomasWaldmann: ok.

PR #7347, opened 2023-02-12
Make hashindex_compact simpler and probably faster

jdchristensen:
The second commit adjusts the tests so that they no longer assume that `hashindex_compact` preserves the order of the entries.  Other early commits clean up some tests and add additional tests.  The change to hashindex_compact is in the second-last commit.  All tests pass after each commit including the two relevant tests run when `BORG_TESTS_SLOW=1`.  I also ran tests with extreme values of `HASH_MAX_LOAD` and `HASH_MAX_EFF_LOAD`.

Please check the new code carefully before merging.  Check for off-by-one errors.

I didn't test the speed.  I'm not sure if compacting takes enough time for it to be worth checking this.

I have a couple of other questions that I will ask at the relevant spot in the code.

### DIFF
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -570,11 +570,11 @@ hashindex_read(PyObject *file_py, int permit_compact, int legacy)
     }
     index->buckets = index->buckets_buffer.buf;
 
-    if(!permit_compact) {
-        index->min_empty = get_min_empty(index->num_buckets);
-        if (index->num_empty == -1)  // we read a legacy index without num_empty value
-            index->num_empty = count_empty(index);
+    index->min_empty = get_min_empty(index->num_buckets);
+    if (index->num_empty == -1)  // we read a legacy index without num_empty value
+        index->num_empty = count_empty(index);
 
+    if(!permit_compact) {
         if(index->num_empty < index->min_empty) {
             /* too many tombstones here / not enough empty buckets, do a same-size rebuild */
             if(!hashindex_resize(index, index->num_buckets)) {
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -837,7 +837,7 @@ hashindex_compact(HashIndex *index)
     int tail = 0;
     uint64_t saved_size = (index->num_buckets - index->num_entries) * (uint64_t)index->bucket_size;
 
-    /* idx will point to the last filled spot and tail will point to the first empty spot. */
+    /* idx will point to the last filled spot and tail will point to the first empty or deleted spot. */
     for(;;) {
         /* Find the last filled spot >= index->num_entries. */
         while((idx >= index->num_entries) && BUCKET_IS_EMPTY_OR_DELETED(index, idx)) {
--- a/src/borg/_hashindex.c
+++ b/src/borg/_hashindex.c
@@ -113,6 +113,7 @@ static int hash_sizes[] = {
 
 #define BUCKET_IS_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == DELETED)
 #define BUCKET_IS_EMPTY(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) == EMPTY)
+#define BUCKET_IS_EMPTY_OR_DELETED(index, idx) (BUCKET_IS_EMPTY(index, idx) || BUCKET_IS_DELETED(index, idx))
 
 #define BUCKET_MARK_DELETED(index, idx) (*((uint32_t *)(BUCKET_ADDR(index, idx) + index->key_size)) = DELETED)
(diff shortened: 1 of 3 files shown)


src/borg/_hashindex.c
```diff
@@ -569,11 +570,11 @@ hashindex_read(PyObject *file_py, int permit_compact, int legacy)
...
 
-    if(!permit_compact) {
-        index->min_empty = get_min_empty(index->num_buckets);
-        if (index->num_empty == -1)  // we read a legacy index without num_empty value
-            index->num_empty = count_empty(index);
+    index->min_empty = get_min_empty(index->num_buckets);
+    if (index->num_empty == -1)  // we read a legacy index without num_empty value
+        index->num_empty = count_empty(index);
 
+    if(!permit_compact) {
```
jdchristensen, line 577: I moved the setting of `min_empty` and `num_empty` to outside of the if, so they happen even when `permit_compact` is true.  I feel like otherwise they might contain garbage that could in theory cause problems.  This is unrelated to the rest of the PR, so I'll drop it if there is any doubt about it.

src/borg/testsuite/hashindex.py
```diff
@@ -486,87 +477,57 @@ def write_empty(self, key):
     def write_deleted(self, key):
         self.write_entry(key, 0xFFFFFFFE, 0, 0)
 
+    def compare_indexes(self, idx1, idx2):
+        for k, v in idx1.iteritems():
+            assert v == idx2[k]
```
jdchristensen, line 482: If I write `assert idx1[k] == idx2[k]` here I get a key not found error.  Any use of `idx1[k]` gives this error.  Is there some reason I can't look up a key while iterating over the hashtable?  I'm just mentioning it in case it indicates a bug somewhere.

ThomasWaldmann: I don't remember any reason right now, will have a deeper look later.

src/borg/testsuite/hashindex.py
```diff
@@ -482,114 +482,52 @@ def compare_indexes(self, idx1, idx2):
...
+                self.write_entry(H2(k), 3*k+1, 3*k+2, 3*k+3)
+            k += 1
         idx = self.index_from_data()
         cpt = self.index_from_data()
         cpt.compact()
-        assert idx.size() == 1024 + 6 * (32 + 3 * 4)
-        assert cpt.size() == 1024 + 3 * (32 + 3 * 4)
+        assert idx.size() == 1024 + num_buckets * (32 + 3 * 4)
+        assert cpt.size() == 1024 + num_entries * (32 + 3 * 4)
         self.compare_indexes(idx, cpt)
```
ThomasWaldmann, line 509: great, makes this so much better!

src/borg/_hashindex.c
```diff
@@ -828,56 +829,32 @@ hashindex_next_key(HashIndex *index, const unsigned char *key)
...
-            compact_tail_idx++;
-            continue;
-        }
-
-        /* Phase 2: Find some non-empty/non-deleted slots we can move to the compact tail */
-
-        while(empty_slot_count && (idx < index->num_buckets) && !(BUCKET_IS_EMPTY(index, idx) || BUCKET_IS_DELETED(index, idx))) {
-            idx++;
-            empty_slot_count--;
+    /* idx will point to the last filled spot and tail will point to the first empty spot. */
```
ThomasWaldmann, line 840: ... tail will point to the first empty or deleted spot.

src/borg/_hashindex.c
ThomasWaldmann: Will have a deeper look at this later.

jdchristensen: (This is actually slight related to this PR, since I did add `permit_compact=True` to the test suite.)

ThomasWaldmann: LGTM.

I just noticed that `count_empty` could be tweaked for compact hashtables (where we can detect there can be no empty buckets because `num_entries == num_buckets`).

But this doesn't need to get optimized in master, because it now persists the `num_empty` value on disk anyways.

src/borg/testsuite/hashindex.py
ThomasWaldmann: idx1 is an artificially (and not necessarily correctly) constructed HT, idx2 ("cpt") is constructed the same way, but then compacted.

When doing random lookups, the lookup algorithm relies on the HT being set up **correctly** (and gives up searching for correct key once it finds an empty bucket).

This makes lookups in the non-compact table fail, while lookups in the compact table still succeed because there are no empty buckets.

Iteration just goes over everything (skipping empty/deleted), so does not have issues with bad HTs.

Guess this could be copied to a comment here.

jdchristensen: Oh, I figured it out.  It has nothing to do with the iterator, but is simply because the entries aren't located where they should be.  I've added comments explaining this.

ThomasWaldmann: Guess we found it in parallel. :-)

ThomasWaldmann (approved):
LGTM.

jdchristensen:
I'm done, so feel free to merge if you have checked the code.

PR #7351, opened 2023-02-13
create: implement retries for individual fs files

ThomasWaldmann:
Errors handled for backup src files:
- BackupOSError (converted from OSError), e.g. I/O Error
- BackupError (stats race, file changed while we backed it up)

Error Handling:
- retry the same file after some sleep time
- sleep time starts from 1ms, increases exponentially up to 10s
- 10 tries

If retrying does not help:
- BackupOSError: skip the file, log it with "E" status
- BackupError: last try will back it up, log it with "C" status

Works for:
- borg create's normal (builtin) fs recursion
- borg create --paths-from-command
- borg create --paths-from-stdin

Notes:
- update stats.files_stats late (so we don't get wrong
  stats in case of e.g. IOErrors while reading the file).
- _process_any: no changes to the big block, just indented
  for adding the retry loop and the try/except.
- test_create_erroneous_file succeeds because we retry the file.

### DIFF
--- a/src/borg/archiver/create_cmd.py
+++ b/src/borg/archiver/create_cmd.py
@@ -286,59 +286,105 @@ class CreateMixIn:
 
         if dry_run:
             return "+"  # included
-        elif stat.S_ISREG(st.st_mode):
-            return fso.process_file(path=path, parent_fd=parent_fd, name=name, st=st, cache=cache)
-        elif stat.S_ISDIR(st.st_mode):
-            return fso.process_dir(path=path, parent_fd=parent_fd, name=name, st=st)
-        elif stat.S_ISLNK(st.st_mode):
-            if not read_special:
-                return fso.process_symlink(path=path, parent_fd=parent_fd, name=name, st=st)
-            else:
-                try:
-                    st_target = os_stat(path=path, parent_fd=parent_fd, name=name, follow_symlinks=True)
-                except OSError:
-                    special = False
-                else:
-                    special = is_special(st_target.st_mode)
-                if special:
+        MAX_RETRIES = 10  # count includes the initial try (initial try == "retry 0")
+        for retry in range(MAX_RETRIES):
+            last_try = retry == MAX_RETRIES - 1
+            try:
+                if stat.S_ISREG(st.st_mode):
                     return fso.process_file(
-                        path=path, parent_fd=parent_fd, name=name, st=st_target, cache=cache, flags=flags_special_follow
+                        path=path, parent_fd=parent_fd, name=name, st=st, cache=cache, last_try=last_try
+                    )
+                elif stat.S_ISDIR(st.st_mode):
+                    return fso.process_dir(path=path, parent_fd=parent_fd, name=name, st=st)
+                elif stat.S_ISLNK(st.st_mode):
+                    if not read_special:
+                        return fso.process_symlink(path=path, parent_fd=parent_fd, name=name, st=st)
+                    else:
+                        try:
+                            st_target = os_stat(path=path, parent_fd=parent_fd, name=name, follow_symlinks=True)
+                        except OSError:
+                            special = False
+                        else:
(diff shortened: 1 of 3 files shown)


src/borg/archiver/create_cmd.py
```diff
@@ -286,59 +286,84 @@ def _process_any(self, *, path, parent_fd, name, st, fso, cache, read_special, d
...
-            self.print_warning("Unknown file type: %s", path)
-            return
+                    self.print_warning("Unknown file type: %s", path)
+                    return
+            except (BackupError, BackupOSError) as err:
+                # sleep a bit, so temporary problems might go away...
+                sleep_s = 1000.0 / 1e6 * 10 ** (retry / 2)  # retry 0: 1ms, retry 6: 1s, ...
+                time.sleep(sleep_s)
+                if retry < MAX_RETRIES - 1:
+                    logger.warning(f"{err}, slept {sleep_s:.3f}s, next: retry: {retry + 1} of {MAX_RETRIES - 1}...")
```
sophie-h, line 358: Can we make this an `info`? This basically keeps things at a non-zero exit code and alerts users, which is basically what this PR wants to get rid of, right?

src/borg/archiver/create_cmd.py
```diff
@@ -286,59 +286,84 @@ def _process_any(self, *, path, parent_fd, name, st, fso, cache, read_special, d
...
+                else:
+                    logger.error(f"{err}, giving up (skipping this file).")
```
sophie-h, line 360: Wasn't `error` so far more about backups that completely failed?

src/borg/archiver/create_cmd.py
ThomasWaldmann: IIRC, logger.warning (stdlib code) by itself does not set rc 1.

Usually borg calls a wrapper that both log at warning level and tweaks the rc.

src/borg/archiver/create_cmd.py
ThomasWaldmann: rc 2 was sudden death due to severe issues (borg crashing, disk full, ...).

iirc, `borg check also emits log output at ERROR level` when it finds severe issues in a repo.

here we have a severe issue with a unreadable or always-changing input file, which is also quite notable. either it is some crap that should be excluded or your disk has I/O error or the file is super busy in-use and the filesystem better should be snapshotted or some database process stopped.

ThomasWaldmann: BTW, i thought about whether all this output is too spammy and we maybe should only output the last error encountered if all tries failed. But that could lose some error information in case the error is not always the same.

Also, if the 9th retry manages to read your input file, but all previous attempts failed, guess one wants to know that.

src/borg/archiver/create_cmd.py
ThomasWaldmann: Just checked: the wrappers which also set the exit code are `Archiver.print_warning()` and `.print_error()`.

drzraf:
any chance this one could be backported to 1.4.x ?

ThomasWaldmann:
@drzraf Originally, I did not intend to backport this.

Can you tell me your use case, so I maybe can re-evaluate?

sophie-h:
My observation is that file changes happen very frequently on desktop backups. SQLite dbs for messengers, browser databases for bookmarks etc

drzraf:
> My observation is that file changes happen very frequently on desktop backups. SQLite dbs for messengers, browser databases for bookmarks etc

Exactly this.
It creates a lot of noise / false alerts, especially when backup's cron output is consider a significant warning/failure worth to be emailed.

(I've been a very satisfied user of Borg for years, but this particular output/behavioral tweak is what I've awaiting the most :))

ThomasWaldmann:
@drzraf @sophie-h see #9805.

PR #7394, opened 2023-02-28
Make timestamps in manual pages reproducible

helmutg:
Refer to https://reproducible-builds.org/docs/source-date-epoch/ for documentation on SOURCE_DATE_EPOCH.

### DIFF
--- a/setup_docs.py
+++ b/setup_docs.py
@@ -7,6 +7,7 @@ import sys
 import textwrap
 from collections import OrderedDict
 from datetime import datetime
+import time
 
 from setuptools import Command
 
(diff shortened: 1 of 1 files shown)


helmutg:
I forgot to link up the Debian bug https://bugs.debian.org/1029807

setup_docs.py
```diff
@@ -470,7 +471,10 @@ def write_man_header(self, write, title, description):
         self.write_heading(write, description, double_sided=True)
         # man page metadata
         write(":Author: The Borg Collective")
-        write(":Date:", datetime.utcnow().date().isoformat())
+        write(
+            ":Date:",
+            datetime.utcfromtimestamp(int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))).date().isoformat(),
```
ThomasWaldmann, line 476: making this reproducible is fine, but why was the default code (when the env var is not set) changed?

helmutg: I can only second guess here, but I think the idea was to avoid branching. `datetime.utcnow()` and `datetime.utcfromtimestamp(int(time.time()))` shouldn't be that different.

I'll summon @vagrantc for a more detailed answer.

ThomasWaldmann: hmm, ok, result seems to be the same.

ThomasWaldmann: @helmutg no need, i first had the wrong idea about what's in that env var and got the bracketing wrong...

ThomasWaldmann:
@helmutg guess the same change for 1.2-maint would be good.

Issue #7363, opened 2023-02-18
compression test issue on S390 (1.1-maint)


ThomasWaldmann:
From @FelixSchwarz:

Thanks Thomas, I did miss that you had additional commits on 1.1-maint. I cherry-picked some commits which seem relevant but the build still fails:
```
_____________ ArchiverTestCase.test_compression_auto_compressible ______________
self = <borg.testsuite.archiver.ArchiverTestCase testMethod=test_compression_auto_compressible>
    def test_compression_auto_compressible(self):
        size, csize = self._get_sizes('auto,zlib', compressible=True)
        assert csize < size * 0.1
>       assert csize == 35  # same as compression 'zlib'
E       AssertionError: assert 53 == 35
build/lib.linux-s390x-3.6/borg/testsuite/archiver.py:2255: AssertionError
```

Anyway, thank you for maintaining borg the way you do (long branch maintenance, using system libraries, ...) - it really makes distro packaging much easier even though it is a lot of extra work for you.

ThomasWaldmann:
Looks weird, but not like a endian issue.

@FelixSchwarz Can you run `test_compression_zlib_compressible` on S390. Does it fail with the same values in the assert expression? Maybe these tests just assume something that is not true on that system.

ThomasWaldmann:
@FelixSchwarz Looks like the linux and macOS tests are happy, but now needs another test run on S390 with that changeset cherry-picked, see #7364.

FelixSchwarz:
Pushed a new build to s390x, will report about the results.

FelixSchwarz:
better but not there yet
```
_____________ ArchiverTestCase.test_compression_auto_compressible ______________
self = <borg.testsuite.archiver.ArchiverTestCase testMethod=test_compression_auto_compressible>
    def test_compression_auto_compressible(self):
        # this is testing whether the "auto" meta-compressor chooses the "expensive" zlib compression
        # if it detects that the data is very compressible (and not "lz4" nor "none" compression).
        auto_size, auto_csize = self._get_sizes('auto,zlib', compressible=True)
        self.cmd('delete', self.repository_location)
        zlib_size, zlib_csize = self._get_sizes('zlib', compressible=True)
        assert auto_size == zlib_size
        assert auto_csize < auto_size * 0.1  # it did compress!
>       assert auto_csize == zlib_csize  # looking at the result size, it seems to be zlib compressed
E       AssertionError: assert 53 == 100
[...]

PR #7410, opened 2023-03-04
"auto" compressor tests: don't assume a specific size (backport to 1.2-maint)

snsmac:
#7364 
#7372 

I'm not completely sure if it makes sense to backport 8916b5a7dac560b4a54953b53df9aa6bdcee5f38 as it is not directly related with the other commit

### DIFF
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -2551,14 +2551,19 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         assert csize == size + 3  # same as compression 'none'
 
     def test_compression_auto_compressible(self):
-        # this is testing whether the "auto" meta-compressor chooses the "expensive" zlib compression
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Porting #7364 is still a bit early, as we do not have feedback yet about whether it improved the tests on S390.

ThomasWaldmann:
Let's see whether the tests work with the locked requirements.

After that has finished, we'll try again without that commit.

ThomasWaldmann:
OK, it worked with the locked requirements, we no can check without that change.

snsmac:
On my computer it works without the change, so i updated the MR

ThomasWaldmann:
Looks good, just waiting for feedback in #7363.

ThomasWaldmann:
Looks good: https://github.com/borgbackup/borg/issues/7363#issuecomment-1455136193 (2nd patch == #7372)

ThomasWaldmann:
OK, so now only master is left to fix in the same way.

PR #7413, opened 2023-03-06
"auto" compressor tests: do not assume zlib is better than lz4, fixes #7363 (forwardport-master)

### LINKED ISSUE #7363: compression test issue on S390 (1.1-maint)

snsmac:
As the test structure got refactored in master and as far as i can see does not explicitly check against the output size of the different compression algorithms anymore,  i only adapted commit a6cd0fd2bfaa6f912a4930a4329166d9bde71e94

See:
#7372
#7364

### DIFF
--- a/src/borg/testsuite/compress.py
+++ b/src/borg/testsuite/compress.py
@@ -146,7 +146,8 @@ def test_auto():
     ratio = len(compressed_zlib) / len(compressed_lz4)
     assert meta["ctype"] == ZLIB.ID if ratio < 0.99 else LZ4.ID
     assert meta["clevel"] == 9 if ratio < 0.99 else 255
-    assert meta["csize"] == len(compressed_auto_zlib)
+    smallest_csize = min(len(compressed_zlib), len(compressed_lz4))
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Looks good, thanks!

PR #7414, opened 2023-03-07
Show ctime and mtime on borg diff. Fixes #7248 (Backport of #7335)

### LINKED ISSUE #7248: 'borg diff' does not report changed mtime
YES
1 BUG
1 ENHANCEMENT-REQUEST
1.2.1
Debian bullseye x64
AMD x64 processor, 8GB RAM, 
1Gbps ethernet
3 x 8TB drives layered as LUKS, LVM, BTRFS RAID1
Repo1: 22 TB (un-deduplicated), 2TB (deduplicated), 105 archives in repo (most files are hardlinked together).
See bash script below.
1) BUG: 'borg diff' does not report changed mtime for files within a repository, even though 'borg list --json-lines' shows that the file metadata have changed.  (Other metadata than mtime may have same problem...haven't checked).
2) ENHANCEMENT REQUEST: 'borg list --json-lines' sometimes renders attributes (eg size, mtime) in different order; this causes a 'diff' of the directory entry against another instance to to mis-match, even though the attribute values are identical.  Proposed change: please render attributes in a stable order (alphabetic sort order is suggested, but what is important is that it be the same order across all archives within a single repository).
Yes.  See script which follows:
`
\#!/bin/bash

\# This script demonstrates that 'borg diff' (version 1.2.1, at least) does not report a difference in mtime of a file.
\# This issue may also apply to atime and ctime as well, and also to directories, and to other metadata attributes (such as permissions), but isn't tested by this script.

dir=$(mktemp -d)
pushd "$dir" > /dev/null

borg init -e none repo

mkdir work
echo "hello" > work/hello
borg create repo::archive01 work
borg list --json-lines repo::archive01 > list1
sleep 4      
touch work/hello
borg create repo::archive02 work
borg list --json-lines repo::archive02 > list2

\# Enhancement request: please emit attributes (eg, mtime, size) in same order each time, so that a diff of unchanged elements (eg, a directory entry) will match:
\# Note if 'sleep 4' above is change to 'sleep 2', the 'borg list --json-lines' attributes are rendered in same order.  Why?
diff list1 list2

\# Bug: borg diff does not report the change in mtime, even though it is visible in 'borg list'
borg diff repo::archive01 archive02

popd
`

Michael-Girma:
---

- Added assert_line_exists helper in BaseTestCase
- JSON strings in diff output are now sorted alphabetically
- Modified diff test cases to confirm to new output format
- Added a test case to test ctime and mtime inclusion
- Mode, ctime & mtime are now only displayed on diff if --content-only flag is used


ThomasWaldmann:
Did you see the tests are failing?

ThomasWaldmann:
BTW: I likely will do a 1.2.4 release soon.

Michael-Girma:
@ThomasWaldmann was away from my rig. Taking a look now.

ThomasWaldmann:
I restarted the CI, let's see if it happens again.

Michael-Girma:
Ahh, able to recreate after pulling in latest 1.2-maint. Taking a look

Michael-Girma:
Getting two entries for a single file path on both text and JSON outputs. I couldn't get a hint why this happens (specially why it's only on linux). Any idea why this might be @ThomasWaldmann?

Michael-Girma:
@ThomasWaldmann Figured out why we're having double entries. Shouldn't this be pointing to the item's path as opposed to just path? Since it's referring to path it adds a tuple with the path of the last recorded item, leading to two entries for a single item. Changing it to ```item1.path``` fixed the issue and all the tests are passing. Should I push this?

image

ThomasWaldmann:
@Michael-Girma congrats, you found a bug!

I think right above that line, there should be a
```
assert item1.path == item2.path
```

And then it should use `item1.path` instead of path.

Can you check master for that problem and if present there, make a separate bugfix PR just for that?
Then (after successful tests) this could be merged and backported to 1.2-maint.
Then, this PR here could be rebased on updated 1.2-maint branch.

ThomasWaldmann:
I just checked: due to the simplifications in master related to "hardlink masters", that code is not present any more.

So guess it is fine if you just fix it here, in a separate commit please.

ThomasWaldmann (approved):
LGTM!

real-yfprojects:
> * Mode, ctime & mtime are now only displayed on diff if --content-only flag is used

I think these change types are only displayed if the flag is _not_ used.

Issue #7356, opened 2023-02-14
Update Docs for setting up Development Environment?


shalearkane:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?
Yes

## Is this a BUG / ISSUE report or a QUESTION?
Question

I have looked into https://borgbackup.readthedocs.io/en/stable/development.html#building-a-development-environment however, I had to install several packages related to compression algorithms libs. It would be nice for new contributors if these packages were mentioned so that the command `pip install -r requirements.d/development.txt` would run without issues on the first try on Linux.

ThomasWaldmann:
Well, development actually requires the same setup as install borg from source (which is documented in the install section), so if we do not mention that already, that pointer could be added.

abebeos:
Fixed in 

small docs update: https://github.com/borgbackup/borg/pull/7423 + small docs update https://github.com/borgbackup/borg/pull/7429

Issue #7398, opened 2023-03-04
update docs (46)


ThomasWaldmann:
there are quite some docs issues: https://github.com/borgbackup/borg/labels/documentation

grab some doc update issues, create a PR against master (and, after discussion, also to 1.2-maint, if applicable, see here, also linked from PR submission template)

the bounty is for fixing 2 small issues or 1 bigger issue.
docs issues that already have a bounty on them do not qualify.

---

:moneybag: [there is a bounty for this](https://app.bountysource.com/issues/117547819-update-docs-46)

abebeos:
@ThomasWaldmann , Bountysource is usually "one entry per issue", what is the bounty process here?

> the bounty is for fixing 2 small issues or 1 bigger issue.

The issues should have a marker "small", so things are clear and transparent right from start.

abebeos:
> This is just to keep management overhead down.

I assume for you. Because for me, it is more overhead (starting with asking "what is the process?").

What is the process?

ThomasWaldmann:
Update some docs, make a PR.

If you only fix a little, it will only be half of the bounty ticket.

abebeos:
>>Most importantly: Is the amount in escrow, and how/when is it payed?
> You can read more at bountysource.com.

No, I can't, as you don't stick to the bountysource process.

ThomasWaldmann:
Strangely, noone yet had a problem with that. So either try or fix issues elsewhere. :-)

abebeos:
> Strangely, noone yet had a problem with that. 

code-monkeys vs. process-thinker

> So either try or fix issues elsewhere. :-)

I extract: you refuse to clarify what your modified (bountysource-) process is.

abebeos:
> There is no modified process.

Correct. Your process is your (very personal) process.

What is relevant however:

you refuse to clarify what your modified (bountysource-) process is.

=> modified (bountysource-) process

Anyways, backing off here, enough played.

abebeos:
small docs update: https://github.com/borgbackup/borg/pull/7423 + small docs update https://github.com/borgbackup/borg/pull/7429

abebeos:
@ThomasWaldmann , ty for the payment!

PR #7423, opened 2023-03-10
docs: add link to OS dependencies

abebeos:
I've replayed the dev-installation on a fresh ubuntu 22.04 and fell into the same trap as the author of #7356, because my doc entry-point was:

* https://borgbackup.readthedocs.io/en/stable/development.html#
  * https://borgbackup.readthedocs.io/en/stable/development.html#building-a-development-environment

Then..

The "Building a development environment" section links to the "Using git" section. This can result in developers overseeing the os dependencies necessity.

re #7356 (via #7398)

(dev environment setup can be further streamlined, if this is of interest)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -145,6 +145,17 @@ the old version using the same steps as shown above.
 From Source
 -----------
 
+Operating System
+~~~~~~~~~~~~~~~~
+
+* :ref:`debian-ubuntu`
+* :ref:`fedora`
+* :ref:`opensuse-tumbleweed-leap`
+* :ref:`macos`
+* :ref:`freebsd`
+* :ref:`windows-10-s-linux-subsystem`
+* :ref:`cygwin`
+
 .. note::
 
   Some older Linux systems (like RHEL/CentOS 5) and Python interpreter binaries
@@ -327,6 +338,8 @@ Virtualenv_ can be used to build and install Borg without affecting
 the system Python or requiring root access.  Using a virtual environment is
 optional, but recommended except for the most simple use cases.
 
+Ensure to install the dependencies as described within :ref:`source-install`.
+
 .. note::
     If you install into a virtual environment, you need to **activate** it
     first (``source borg-env/bin/activate``), before running ``borg``.
@@ -375,6 +388,8 @@ Using git
 
 This uses latest, unreleased development code from git.
 While we try not to break master, there are no guarantees on anything.
+
+Ensure to install the dependencies as described within :ref:`source-install`.
 
 ::
 


docs/installation.rst
```diff
@@ -193,6 +193,8 @@ ports.
 After you have installed the dependencies, you can proceed with steps outlined
 under :ref:`pip-installation`.
 
+.. _debian-ubuntu:
+
```
ThomasWaldmann, line 197: no need for this.

docs/installation.rst
```diff
@@ -376,6 +378,8 @@ Using git
 This uses latest, unreleased development code from git.
 While we try not to break master, there are no guarantees on anything.
 
+Install the dependencies for your OS, for example :ref:`debian-ubuntu`
+
```
ThomasWaldmann, line 382: this documentation was assuming that people just read it top-down, as usual, but some people directly jump into the middle of it and assume all will be fine...

pip (line 323) as well as git (line 373) are source-based installations of borg and thus have the requirements as documented in the "From source" section (line 143+).

So, if you want to point people there, the correct link target is in line 143 and the pointer needs to be added to the pip section also.

ThomasWaldmann (changes requested):
thanks for the PR, some feedback:

docs/installation.rst
abebeos: > this documentation was assuming that people just read it top-down, as usual, but some people directly jump into the middle of it and assume all will be fine...

The problem of this (and many other) documentation is that it assumes too much.

Naturally, if I want to install a dev-system, I jump right to the dev-system section (and I assume that all be fine, e.g. that authors of the docs managed to have all relevant info there).

> So, if you want to point people there, the correct link target is in line 143 and the pointer needs to be added to the pip section also.

I write documentation (especially for developers) "straight-to-the-point", similar here. Pointing to "143" results in (me, and others, i bet) skipping again fast-forward until they see their OS of interest.

---
my most important note was:

>>(dev environment setup can be further streamlined, if this is of interest)

Are you interested in streamlining the documentation/installation process? Or just in "micromanaging" some doc changes?

docs/installation.rst
abebeos: Would this one be redundant, too? : 

https://github.com/borgbackup/borg/blob/0e711b66728b0c2ff4113eb5df32326a9ce85f9e/docs/installation.rst?plain=1#L23-L26

docs/installation.rst
ThomasWaldmann: > e.g. that authors of the docs managed to have all relevant info there

We won't duplicate the install docs into the developers docs. If we don't have that already, add a pointer from there to here.

> Pointing to "143" results in (me, and others, i bet) skipping again fast-forward until they see their OS of interest.

If you skip from 143 to the OS you use, you should be fine. Just pointing to the debian/ubuntu section is only helpful IF you use these.

docs/installation.rst
```diff
@@ -327,6 +327,9 @@ Virtualenv_ can be used to build and install Borg without affecting
 the system Python or requiring root access.  Using a virtual environment is
 optional, but recommended except for the most simple use cases.
 
+Ensure to install the dependencies as described within :ref:`source-install`.
+(Quick-links: :ref:`debian-ubuntu`, :ref:`fedora`, :ref:`macos`, :ref:`freebsd`, :ref:`cygwin`)
```
ThomasWaldmann, line 331: Please remove the second line, we do not want to maintain all these links.

Making the docs longer and more complex is not helpful. As you have noticed, people don't like to read long docs.

docs/installation.rst
```diff
@@ -376,6 +379,9 @@ Using git
 This uses latest, unreleased development code from git.
 While we try not to break master, there are no guarantees on anything.
 
+Ensure to install the dependencies as described within :ref:`source-install`.
+(Quick-links: :ref:`debian-ubuntu`, :ref:`fedora`, :ref:`macos`, :ref:`freebsd`, :ref:`cygwin`)
```
ThomasWaldmann, line 383: same.

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/7356#issuecomment-1431160264 this would be good so we can close #7356.

docs/installation.rst
abebeos: > we do not want to maintain all these links.

those links are static, in-doc, zero maintenance.

> Making the docs longer and more complex is not helpful. 

The docs are too long  (for target audience: developer) , they need reduction.

> As you have noticed, people don't like to read long docs.

That's why I don't point them to line 143/From Source, forcing them to start reading redundant stuff about "Some older Linux systems...". You can do that (and pay the bounty proudly to yourself). I wont do that.

The solution is:
* point to ubuntu (luckily and logically at the start, as widelly used)
* point to 143/From Source, but provide quicklinks
* alternatively add the quick-links section right at the start of 143/From source

I can assist you to fix docs issues (it's a typical "python-domain" mess), mainly by simplifying things.

(Alternatively you can wait for a "code-monkey" to pickup the bounties. They usually do not care about quality, and love to get directions.)

docs/installation.rst
abebeos: see above

docs/installation.rst
abebeos: > If you skip from 143 to the OS you use, you should be fine.

You should really not write / manage documentation.

abebeos: see below for more.

docs/installation.rst
ThomasWaldmann: no, they are not zero maintenance:

if we add or remove OSes, we would have to update that list. also, in case some more recent release of debian/ubuntu behaves different from an older one: we would have to update that list.

pointing them to the right section of the docs is a good idea, but then they can find their OS themselves there.

abebeos: > if we add or remove OSes, we would have to update that list. also, in case some more recent release of debian/ubuntu behaves different from an older one: we would have to update that list.

This is called "usual work load"

> but then they can find their OS themselves there.

Oh, they can find everything themselves (yelling "who tf has written this manuals"? while at it).

docs/installation.rst
```diff
@@ -145,6 +145,17 @@ the old version using the same steps as shown above.
...
+~~~~~~~~~~~~~~~~
+
+* :ref:`debian-ubuntu`
+* :ref:`fedora`
+* :ref:`opensuse-tumbleweed-leap`
+* :ref:`macos`
+* :ref:`freebsd`
+* :ref:`windows-10-s-linux-subsystem`
+* :ref:`cygwin`
+
```
ThomasWaldmann, line 158: ok, that's better, now it is at least at one place.

abebeos: (can be further refined, but I need at some point a merge to trigger the next revision)

ThomasWaldmann: https://github.com/borgbackup/borg/pull/7423#issuecomment-1463964589 if you could check/add that linking, we could actually close that issue. 1 sentence / 1 link (if we do not have that yet) is enough.

ThomasWaldmann (approved):
LGTM.

ThomasWaldmann:
guess this is finished? merge?

abebeos:
Yes, should be ok to merge.

ThomasWaldmann:
Thanks!

ThomasWaldmann:
If you like: this needs also a backport to 1.2-maint branch (from where 1.2.4 release will be made soon).

abebeos:
K, will do tomorrow.

https://github.com/borgbackup/borg/tree/1.2-maint

ThomasWaldmann:
(you could either register that as a small backport task or a small docs update task)

abebeos:
> (you could either register that as a small backport task or a small docs update task)

I guess: small docs update: #7423 + small docs update #7429 ?

ThomasWaldmann:
Exactly.

PR #7431, opened 2023-03-11
fix bullet-list links

abebeos:
Add refs

(note: your information "no need for this" within https://github.com/borgbackup/borg/pull/7423#discussion_r1132297255 was wrong)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -204,6 +204,8 @@ ports.
 After you have installed the dependencies, you can proceed with steps outlined
 under :ref:`pip-installation`.
 
+.. _debian-ubuntu:
+
 Debian / Ubuntu
 +++++++++++++++
 
@@ -222,6 +224,8 @@ In case you get complaints about permission denied on ``/etc/fuse.conf``: on
 Ubuntu this means your user is not in the ``fuse`` group. Add yourself to that
 group, log out and log in again.
 
+.. _fedora:
+
 Fedora
 ++++++
 
@@ -235,6 +239,8 @@ Install the dependencies with development headers::
     sudo dnf install gcc gcc-c++ redhat-rpm-config
     sudo dnf install fuse-devel fuse         # needed for llfuse
     sudo dnf install fuse3-devel fuse3       # needed for pyfuse3
+
+.. _opensuse-tumbleweed-leap:
 
 openSUSE Tumbleweed / Leap
 ++++++++++++++++++++++++++
@@ -251,6 +257,8 @@ Alternatively, you can enumerate all build dependencies in the command line::
     python3-pytest python3-setuptools python3-setuptools_scm \
     python3-sphinx_rtd_theme gcc gcc-c++
     sudo zypper install python3-llfuse  # llfuse
+
+.. _macos:
 
 macOS
 +++++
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
No, my comment was not wrong. I did not want you to add a lot of links to special cases (like ubuntu), but just link to the main section headline (for all OSes) - and for that one the neeeded label was already present.

Afterwards, you then added a directory of OSes below that headline (which I didn't like much, but found acceptable still as it is just at one close-by place vs. 2 distant places of your previous approach). So you either have to remove that or check whether it works by magic via the headlines.

abebeos:
> No, my comment was not wrong. I did not want you to add a lot of links to special cases (like ubuntu), but just link to the main section headline (for all OSes) - and for that one the neeeded label was already present.

Your comment was wrong, additionally you ignored my clarifying question.

Next time, when you comment, focus on the item you comment. If your comment is in a greater context/scope, you need to mention it. Reviewing is not a guess-work.

> So you either have to remove that or check whether it works by magic via the headlines or add the labels as you did here.

I checked already with the generated site, it does NOT work "by magic".

(this change is necessary, the current merged version has non-links)

abebeos:
> and for that one the neeeded label was already present.

It was not.

abebeos:
> or add the labels as you did here.

As you do not like redundancy and duplication (similar to me) : 

Possible Follow-up:

You can add a new documentation issue that would focus to remove the many `.. _my-manual-anchor:` from the doc sources (the effort is beyond one middle doc bounty - possibly 2 or 3 or even more), something like:

### Automate headline anchor/ref creation

* ensure that every headlines generates as `:ref:` target
* remove redundant `.. _my-manual-anchor:`
* verify the generated anchors, resolve duplicates
* do some tests with the resulting docs

ThomasWaldmann:
OK, it seems like sphinx is magically generating HTML anchors for each headline, but not reST targets (yet)?

Maybe we do not yet have that enabled, check this:

https://stackoverflow.com/questions/18632781/how-to-make-an-internal-link-to-a-heading-in-sphinx-restructuredtext-without-cre

An alternative solution would be to just remove that "directory of OSes" you added. Then you do not need all that manual linking. Each OS has an own headline that is pretty simple to discover without the directory you have added.

abebeos:
> Each OS has an own headline

It has not, and I'll not check any more documents you point me to.

You're becoming rude.

The PRs are functional and finished. You can't expect that this continues endlessly in context of a $30 bounty, just because you can't accept that duplication/redundancy is sometimes a necessity.

I mean you have duplication within docs, doesn't this freak you out?

As said here https://github.com/borgbackup/borg/pull/7431#issuecomment-1464943085, file a new issue to get rid of all the link redundancy already existing in your doc sources (independent of my links).

Just unbelievable all this.

abebeos:
(please note that I'll not further comment on this. Enough is enough, get over your strangeness, or apply it at least consistently over your existent code-base, with the result that you'll go nuts from all the redundancy/duplication)

ThomasWaldmann:
bye!

abebeos:
> bye!

Bye you delusional thief.

abebeos:
> > bye!
> 
> Bye you delusional thief.

Correction: NOT a thief. See https://github.com/borgbackup/borg/issues/7398#issuecomment-1465124740

abebeos:
> Each OS has an own headline that is pretty simple to discover without the directory you have added.

Clarification: You talked about manual discovery.

I talked about the simplified discovery via headline anchors.

Simplified, as things need to be for a good UX.

PR #7435, opened 2023-03-11
xattrs: fix namespace processing on FreeBSD, fixes #6997

### LINKED ISSUE #6997: check xattr namespace behaviour
there was a strange issue in #6988 and the root cause was a strange xattr named `user.security.selinux` (the correct name would be `security.selinux`).

the processing (or rather ignoring) of such invalid stuff was solved by #6990.

but the open question is still: where did that invalid xattr initially come from?

we should make sure it did not come from borg.

ThomasWaldmann:
TODO:
- review by freebsd users / devs
- testing on freebsd

Fixes #6997.

### DIFF
--- a/src/borg/platform/freebsd.pyx
+++ b/src/borg/platform/freebsd.pyx
@@ -50,59 +50,77 @@ cdef extern from "unistd.h":
     int _PC_ACL_NFS4
 
 
-def listxattr(path, *, follow_symlinks=False):
-    ns, prefix = EXTATTR_NAMESPACE_USER, b'user.'
+# On FreeBSD, borg currently only deals with the USER namespace as it is unclear
+# whether (and if so, how exactly) it should deal with the SYSTEM namespace.
+NS_ID_MAP = {b"user": EXTATTR_NAMESPACE_USER, }
 
+
+def split_ns(ns_name, default_ns):
+    # split ns_name (which is in the form of b"namespace.name") into namespace and name.
+    # if there is no namespace given in ns_name, default to default_ns.
+    # note:
+    # borg < 1.1.10 on FreeBSD did not prefix the namespace to the names, see #3952.
+    # we also need to deal with "unexpected" namespaces here, they could come
+    # from borg archives made on other operating systems.
+    ns_name_tuple = ns_name.split(b".", 1)
+    if len(ns_name_tuple) == 2:
+        # we have a namespace prefix in the given name
+        ns, name = ns_name_tuple
+    else:
+        # no namespace given in ns_name (== no dot found), maybe data coming from an old borg archive.
+        ns, name = default_ns, ns_name
+    return ns, name
+
+
+def listxattr(path, *, follow_symlinks=False):
     def func(path, buf, size):
         if isinstance(path, int):
-            return c_extattr_list_fd(path, ns, <char *> buf, size)
+            return c_extattr_list_fd(path, ns_id, <char *> buf, size)
         else:
             if follow_symlinks:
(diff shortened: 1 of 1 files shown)


src/borg/platform/freebsd.pyx
```diff
@@ -50,59 +50,77 @@ cdef extern from "unistd.h":
     int _PC_ACL_NFS4
 
 
-def listxattr(path, *, follow_symlinks=False):
-    ns, prefix = EXTATTR_NAMESPACE_USER, b'user.'
+# On FreeBSD, borg currently only deals with the USER namespace as it is unclear
+# whether (and if so, how exactly) it should deal with the SYSTEM namespace.
```
ThomasWaldmann, line 54: Support for SYSTEM would need a separate PR from an experienced FreeBSD user/developer.

The scope of this PR here is to make the current borg behaviour less broken, especially when restoring archives that were made on other platforms onto a FreeBSD system (they could have xattr names with other namespaces than "user").

src/borg/platform/freebsd.pyx
```diff
@@ -50,59 +50,77 @@ cdef extern from "unistd.h":
...
+                return c_extattr_list_file(path, ns_id, <char *> buf, size)
             else:
-                return c_extattr_list_link(path, ns, <char *> buf, size)
+                return c_extattr_list_link(path, ns_id, <char *> buf, size)
 
+    ns = b"user"
+    ns_id = NS_ID_MAP[ns]
     n, buf = _listxattr_inner(func, path)
-    return [prefix + name for name in split_lstring(buf[:n]) if name]
+    return [ns + b"." + name for name in split_lstring(buf[:n]) if name]
```
ThomasWaldmann, line 88: listxattr currently will always and only yield b"user.something" xattr names.´, if any.

this is e.g. used at `borg create` time, when creating a new backup on freebsd.

the keys returned by listxattr are then used by getxattr to fetch the values.

src/borg/platform/freebsd.pyx
```diff
@@ -50,59 +50,77 @@ cdef extern from "unistd.h":
...
+                return c_extattr_set_file(path, ns_id, name, <char *> value, size)
             else:
-                return c_extattr_set_link(path, ns, name, <char *> value, size)
+                return c_extattr_set_link(path, ns_id, name, <char *> value, size)
 
-    # strip namespace if there, but ignore if not there.
-    # older borg / attic versions did not prefix the namespace to the names.
-    if name.startswith(prefix):
-        name = name[len(prefix):]
-    _setxattr_inner(func, path, name, value)
```
ThomasWaldmann, line 105: this was problematic in case an archive from another platform (e.g. with `name == b"security.selinux"`) was restored on freebsd.

it then used `EXTATTR_NAMESPACE_USER` and `name == b"security.selinux"` in the API call.

also, it doesn't make sense to restore selinux stuff onto freebsd anyway, right?

src/borg/platform/freebsd.pyx
```diff
@@ -50,59 +50,77 @@ cdef extern from "unistd.h":
...
+    ns, name = split_ns(name, b"user")
+    try:
+        ns_id = NS_ID_MAP[ns]  # this will raise a KeyError it the namespace is unsupported
+    except KeyError:
+        pass
+    else:
+        _setxattr_inner(func, path, name, value)
```
ThomasWaldmann, line 123: this will only set USER xattrs and ignore all unsupported stuff.

ThomasWaldmann:
I merged this into 1.2-maint to simplify platform testing.

Review is still welcome!

Issue #7406, opened 2023-03-04
add debug commands for repoobj id_hash/parse/format


ThomasWaldmann:
#7405 fixed `borg debug put-obj / get-obj`:
```
borg debug get-obj id file  # read object id from repo, write it to file
borg debug put-obj id file  # read file, write it to object id in the repo
```

Sometimes, IDs are mentioned in error messages, so `get-obj` can be used to read an object with the known id from the repo and write it to some file. But in borg2, the file format is non-trivial: encrypted/msgpacked chunk metadata plus encrypted/compressed chunk data.

But, `put-obj` alone is not very useful yet in general, because the id must be given, but is usually non-trivial to compute.

Also, to parse or format the repo object format, we need debug wrappers
```
borg debug id-hash file  # read plaintext file, compute the object id from it
borg debug parse-obj objfile  # read object file, parse the object into meta dict and data (decrypting, decompressing)
borg debug format-obj file  # read plaintext file, format it into a compressed, encrypted object with a meta header
```

These functions can all be found in `repoobj.py`.

shalearkane:
I am working on this, issue. I am trying to understand how the id_hash can be calculated. Here's what I understand. 

- id-hash is calculated by the function (which must be implemented by the derived `Key` classes): https://github.com/borgbackup/borg/blob/611e391da907e420631ffbd5dca2c931e54ad619/src/borg/crypto/key.py#L195-L197
- In `1.2maint` you registered a new command, https://github.com/borgbackup/borg/blob/cef5b446acf66cdcdd1d545ecb99ea28e62608c3/src/borg/archiver.py#L3745-L3758
- and you implemented this function https://github.com/borgbackup/borg/blob/cef5b446acf66cdcdd1d545ecb99ea28e62608c3/src/borg/archiver.py#L2210-L2217
- In the above function, the decorator https://github.com/borgbackup/borg/blob/cef5b446acf66cdcdd1d545ecb99ea28e62608c3/src/borg/archiver.py#L2210 returns the derived Key class by parsing the repository.

However in `master` branch, the `with_repository` decorator https://github.com/borgbackup/borg/blob/611e391da907e420631ffbd5dca2c931e54ad619/src/borg/archiver/_common.py#L74-L160 does not pass the `key` in its `kwargs`.

So implementing the same thing in borg2 requires me to implement detecting and passing `key` in `kwargs` in `with_repository`, right?

ThomasWaldmann:
[...]

PR #7416, opened 2023-03-07
[WIP] Add debug commands for repoobj id_hash/parse/format 

shalearkane:
Implements #7406
- [x] id-hash
- [ ] parse-obj
- [ ] format-obj
- [ ] Tests

PR #7436, opened 2023-03-12
debug id-hash: implement file content id-hash computation, see #7406 (forwardport-master)

snsmac:
Forward-port of eb8ccd22d46f60896a66e1d6ce03b204489dba30 #7411

### DIFF
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -255,6 +255,16 @@ class DebugMixIn:
         print("object %s fetched." % hex_id)
         return EXIT_SUCCESS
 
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
hi @snsmac, thanks for the PR!

there is some overlap with #7416, but looks like your forward port is already in the desired state.

just `black` is complaining (see CI `lint` failure) - likely you forgot to locally install it (via `requirements.d/codestyle.txt`) and run `black .` to auto-format the source (we only do this in master).

snsmac:
> just `black` is complaining

I reformatted the code so it should all pass now

ThomasWaldmann:
Thanks!

PR #7438, opened 2023-03-12
Windows CI: only run when certain files are modified

RayyanAnsari:
The Windows CI does not need to run when files such as docs are changed.


ThomasWaldmann:
Thanks!

PR #7443, opened 2023-03-13
Add parse-obj and format-obj debug commands

shalearkane:
Implements #7406 

Format of `parse-obj` and `format-obj` commands are:
```python
borg debug parse-obj --repo REPO_PATH INPUT_ID COMPRESSED_FILE OUTPUT_FILE OUTPUT_JSON 
borg debug format-obj --repo REPO_PATH INPUT_ID INPUT_FILE INPUT_JSON OUTPUT_FILE --compression="xyz,4"
```

The user is expected to use `borg debug get-obj` to get `COMPRESSED_FILE` for `parse-obj`.
Similarly, the user is expected to use `borg debug id-hash` to get `INPUT_ID` for `format-obj`.

### DIFF
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -280,11 +280,11 @@ class DebugMixIn:
             print("object id %s is invalid [%s]." % (hex_id, str(err)))
             return EXIT_ERROR
 
-        with open(args.cdata, "rb") as f:
-            cdata = f.read()
+        with open(args.object_path, "rb") as f:
+            object_file = f.read()
 
         repo_objs = manifest.repo_objs
-        meta, data = repo_objs.parse(id=id, cdata=cdata)
+        meta, data = repo_objs.parse(id=id, cdata=object_file)
 
         with open(args.json_path, "w") as f:
             json.dump(meta, f)
@@ -592,7 +592,7 @@ class DebugMixIn:
         subparser.set_defaults(func=self.do_debug_parse_obj)
         subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
         subparser.add_argument(
-            "cdata", metavar="COMPRESSED_DATA", type=str, help="path of the objectfile to parse data from"
+            "object_path", metavar="OBJECT_PATH", type=str, help="path of the objectfile to parse data from"
         )
         subparser.add_argument(
             "binary_path", metavar="BINARY_PATH", type=str, help="path of the file to write uncompressed data into"
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -268,7 +268,7 @@ class DebugMixIn:
 
     @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
     def do_debug_parse_obj(self, args, repository, manifest):
-        """parse object file into meta dict and data (decrypting, decompressing)"""
+        """parse borg object file into meta dict and data (decrypting, decompressing)"""
 
         # get the object from id
         hex_id = args.id
@@ -281,10 +281,10 @@ class DebugMixIn:
             return EXIT_ERROR
 
         with open(args.object_path, "rb") as f:
-            object_file = f.read()
+            cdata = f.read()
 
         repo_objs = manifest.repo_objs
-        meta, data = repo_objs.parse(id=id, cdata=object_file)
+        meta, data = repo_objs.parse(id=id, cdata=cdata)
 
         with open(args.json_path, "w") as f:
             json.dump(meta, f)
@@ -587,12 +587,12 @@ class DebugMixIn:
             description=self.do_debug_parse_obj.__doc__,
             epilog=debug_parse_obj_epilog,
             formatter_class=argparse.RawDescriptionHelpFormatter,
-            help="parse objectfile into meta dict and data",
+            help="parse borg object file into meta dict and data",
         )
         subparser.set_defaults(func=self.do_debug_parse_obj)
         subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
         subparser.add_argument(
-            "object_path", metavar="OBJECT_PATH", type=str, help="path of the objectfile to parse data from"
+            "object_path", metavar="OBJECT_PATH", type=str, help="path of the object file to parse data from"
         )
         subparser.add_argument(
             "binary_path", metavar="BINARY_PATH", type=str, help="path of the file to write uncompressed data into"
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -3,8 +3,10 @@ from binascii import unhexlify, hexlify
 import functools
 import json
 import textwrap
+import traceback
 
 from ..archive import Archive
+from ..compress import CompressionSpec
 from ..constants import *  # NOQA
 from ..helpers import msgpack
 from ..helpers import sysinfo
@@ -278,28 +280,22 @@ class DebugMixIn:
         except ValueError as err:
             print("object id %s is invalid [%s]." % (hex_id, str(err)))
             return EXIT_ERROR
-        try:
-            cdata = repository.get(id)
-        except Repository.ObjectNotFound:
-            print("object %s not found." % hex_id)
-            return EXIT_ERROR
 
-        output_metadata_path: str = args.json_path
-        output_data_path: str = args.data_path
+        with open(args.cdata, "rb") as f:
+            cdata = f.read()
 
-        key = manifest.key
-        repo_objs = RepoObj(key)
+        repo_objs = manifest.repo_objs
 
         try:
             meta, data = repo_objs.parse(obj_id=id, cdata=cdata, decompress=True)
-        except Exception as e:
-            print(f"encountered exception {e.__str__}")
+        except Exception:
+            traceback.print_exc()
             return EXIT_ERROR
 
-        with open(output_metadata_path, "wb") as f:
-            f.write(json.dump(meta))
+        with open(args.json_path, "wb") as f:
+            json.dump(meta, f)
 
-        with open(output_data_path, "wb") as f:
+        with open(args.data_path, "wb") as f:
             f.write(data)
 
         return EXIT_SUCCESS
@@ -307,23 +303,32 @@ class DebugMixIn:
     @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
     def do_debug_format_obj(self, args, repository, manifest):
         """format file and metadata into borg object file"""
+
+        # get the object from id
+        hex_id = args.id
+        try:
+            id = unhexlify(hex_id)
+            if len(id) != 32:  # 256bit
+                raise ValueError("id must be 256bits or 64 hex digits")
+        except ValueError as err:
+            print("object id %s is invalid [%s]." % (hex_id, str(err)))
+            return EXIT_ERROR
+
         with open(args.data_path, "rb") as f:
             data = f.read()
 
-        with open(args.json_path, "rb") as f:
+        with open(args.json_path, "r") as f:
             meta = json.load(f)
 
-        key = manifest.key
-        repo_objs = RepoObj(key)
+        repo_objs = manifest.repo_objs
 
-        id = repo_objs.id_hash(data)
-        ctype = args.ctype
-        clevel = args.clevel
+        ctype = repo_objs.compressor.ID
+        clevel = repo_objs.compressor.level
 
         try:
-            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
-        except Exception as e:
-            print(f"Encountered exception {e.__str__}")
+            data_encrypted = repo_objs.format(id=hex_id, meta=meta, data=data, ctype=ctype, clevel=clevel)
+        except Exception:
+            traceback.print_exc()
             return EXIT_ERROR
         else:
             with open(args.output_path, "wb") as f:
@@ -599,7 +604,10 @@ class DebugMixIn:
             help="parse objectfile into meta dict and data",
         )
         subparser.set_defaults(func=self.do_debug_parse_obj)
-        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get objectfile from the repo")
+        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
+        subparser.add_argument(
+            "cdata", metavar="COMPRESSED_DATA", type=str, help="path of the objectfile to parse data from"
+        )
         subparser.add_argument(
             "data_path", metavar="DATA_PATH", type=str, help="path of the file to write uncompressed data into"
         )
@@ -610,7 +618,7 @@ class DebugMixIn:
         # format_obj
         debug_format_obj_epilog = process_epilog(
             """
-                This command format the file and metadata into borg objectfile.
+                This command formats the file and metadata into objectfile.
                 """
         )
         subparser = debug_parsers.add_parser(
@@ -623,14 +631,22 @@ class DebugMixIn:
             help="format file and metadata into borg objectfile",
         )
         subparser.set_defaults(func=self.do_debug_format_obj)
+        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
         subparser.add_argument(
             "data_path", metavar="DATA_PATH", type=str, help="path of the file to convert into objectfile"
         )
         subparser.add_argument(
             "json_path", metavar="JSON_PATH", type=str, help="path of the json file to read metadata from"
         )
-        subparser.add_argument("ctype", metavar="CTYPE", type=int, help="compression type 0..255")
-        subparser.add_argument("clevel", metavar="CLEVEL", type=int, help="compression level 0..255")
+        subparser.add_argument(
+            "-C",
+            "--compression",
+            metavar="COMPRESSION",
+            dest="compression",
+            type=CompressionSpec,
+            default=CompressionSpec("lz4"),
+            help="select compression algorithm, see the output of the " '"borg help compression" command for details.',
+        )
         subparser.add_argument(
             "output_path",
             metavar="OUTPUT_PATH",
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -3,7 +3,6 @@ from binascii import unhexlify, hexlify
 import functools
 import json
 import textwrap
-import traceback
 
 from ..archive import Archive
 from ..compress import CompressionSpec
@@ -285,12 +284,7 @@ class DebugMixIn:
             cdata = f.read()
 
         repo_objs = manifest.repo_objs
-
-        try:
-            meta, data = repo_objs.parse(id=id, cdata=cdata, decompress=True)
-        except Exception:
-            traceback.print_exc()
-            return EXIT_ERROR
+        meta, data = repo_objs.parse(id=id, cdata=cdata, decompress=True)
 
         with open(args.json_path, "w") as f:
             json.dump(meta, f)
@@ -321,19 +315,11 @@ class DebugMixIn:
             meta = json.load(f)
 
         repo_objs = manifest.repo_objs
+        data_encrypted = repo_objs.format(id=id, meta=meta, data=data)
 
-        ctype = repo_objs.compressor.ID
-        clevel = repo_objs.compressor.level
-
-        try:
-            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
-        except Exception:
-            traceback.print_exc()
-            return EXIT_ERROR
-        else:
-            with open(args.output_path, "wb") as f:
-                f.write(data_encrypted)
-            return EXIT_SUCCESS
+        with open(args.output_path, "wb") as f:
+            f.write(data_encrypted)
+        return EXIT_SUCCESS
 
     @with_repository(manifest=False, exclusive=True)
     def do_debug_put_obj(self, args, repository):
--- a/src/borg/archiver/debug_cmd.py
+++ b/src/borg/archiver/debug_cmd.py
@@ -265,6 +265,71 @@ class DebugMixIn:
         print(id.hex())
         return EXIT_SUCCESS
 
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
+    def do_debug_parse_obj(self, args, repository, manifest):
+        """parse object file into meta dict and data (decrypting, decompressing)"""
+
+        # get the object from id
+        hex_id = args.id
+        try:
+            id = unhexlify(hex_id)
+            if len(id) != 32:  # 256bit
+                raise ValueError("id must be 256bits or 64 hex digits")
+        except ValueError as err:
+            print("object id %s is invalid [%s]." % (hex_id, str(err)))
+            return EXIT_ERROR
+        try:
+            cdata = repository.get(id)
+        except Repository.ObjectNotFound:
+            print("object %s not found." % hex_id)
+            return EXIT_ERROR
+
+        output_metadata_path: str = args.json_path
+        output_data_path: str = args.data_path
+
+        key = manifest.key
+        repo_objs = RepoObj(key)
+
+        try:
+            meta, data = repo_objs.parse(obj_id=id, cdata=cdata, decompress=True)
+        except Exception as e:
+            print(f"encountered exception {e.__str__}")
+            return EXIT_ERROR
+
+        with open(output_metadata_path, "wb") as f:
+            f.write(json.dump(meta))
+
+        with open(output_data_path, "wb") as f:
+            f.write(data)
+
+        return EXIT_SUCCESS
+
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
+    def do_debug_format_obj(self, args, repository, manifest):
+        """format file and metadata into borg object file"""
+        with open(args.data_path, "rb") as f:
+            data = f.read()
+
+        with open(args.json_path, "rb") as f:
+            meta = json.load(f)
+
+        key = manifest.key
+        repo_objs = RepoObj(key)
+
+        id = repo_objs.id_hash(data)
+        ctype = args.ctype
+        clevel = args.clevel
+
+        try:
+            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
+        except Exception as e:
+            print(f"Encountered exception {e.__str__}")
+            return EXIT_ERROR
+        else:
+            with open(args.output_path, "wb") as f:
+                f.write(data_encrypted)
+            return EXIT_SUCCESS
+
     @with_repository(manifest=False, exclusive=True)
     def do_debug_put_obj(self, args, repository):
         """put file contents into the repository"""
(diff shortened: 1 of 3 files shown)


src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+                raise ValueError("id must be 256bits or 64 hex digits")
+        except ValueError as err:
+            print("object id %s is invalid [%s]." % (hex_id, str(err)))
+            return EXIT_ERROR
+        try:
+            cdata = repository.get(id)
+        except Repository.ObjectNotFound:
+            print("object %s not found." % hex_id)
+            return EXIT_ERROR
+
```
ThomasWaldmann, line 286: guess you duplicated this from `get-obj`?

maybe it is simpler to just use the data from the output file of `get-obj` than having this duplication.

consistency: you also do not put the object to the repo for `debug format-obj`

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        output_metadata_path: str = args.json_path
+        output_data_path: str = args.data_path
+
```
ThomasWaldmann, line 289: simpler: just directly use args.* below?

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        key = manifest.key
+        repo_objs = RepoObj(key)
+
+        try:
+            meta, data = repo_objs.parse(obj_id=id, cdata=cdata, decompress=True)
+        except Exception as e:
+            print(f"encountered exception {e.__str__}")
```
ThomasWaldmann, line 296: would just `{e}` also work? if not, please use `str(e)`.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+            return EXIT_ERROR
+
+        with open(output_metadata_path, "wb") as f:
+            f.write(json.dump(meta))
```
ThomasWaldmann, line 300: Guess you did not test that, that can't work.

```
        with open(output_metadata_path, "w") as f:
            json.dump(meta, f)
```

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+
+        with open(output_data_path, "wb") as f:
+            f.write(data)
```
ThomasWaldmann, line 303: Correct.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+
+        return EXIT_SUCCESS
+
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
+    def do_debug_format_obj(self, args, repository, manifest):
+        """format file and metadata into borg object file"""
+        with open(args.data_path, "rb") as f:
+            data = f.read()
+
+        with open(args.json_path, "rb") as f:
```
ThomasWaldmann, line 313: json is text, thus mode "r".

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+            meta = json.load(f)
+
+        key = manifest.key
+        repo_objs = RepoObj(key)
```
ThomasWaldmann, line 317: That's not correct in any case (old repos would need RepoObj1) and also not needed.

You already have a ready-to-use instance in `manifest.repo_objs` (which will also have the correct compressor, see my other comment below).

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+
+        id = repo_objs.id_hash(data)
```
ThomasWaldmann, line 319: also, this is duplicated from `borg debug id-hash`.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        ctype = args.ctype
+        clevel = args.clevel
+
+        try:
+            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
```
ThomasWaldmann, line 324: see `_common.py:142` where it sets up the compressor for a nicer to use way (this new command will need to have a `--compression` option in the same way as other commands have).

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        except Exception as e:
+            print(f"Encountered exception {e.__str__}")
```
ThomasWaldmann, line 326: see above.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +265,71 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+            return EXIT_ERROR
+        else:
+            with open(args.output_path, "wb") as f:
+                f.write(data_encrypted)
```
ThomasWaldmann, line 330: correct (and the user then could use `put-obj` to write this into the repo - for that he will also need the id, btw).

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +583,61 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+            "parse-obj",
+            parents=[common_parser],
+            add_help=False,
+            description=self.do_debug_parse_obj.__doc__,
+            epilog=debug_parse_obj_epilog,
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            help="parse objectfile into meta dict and data",
+        )
+        subparser.set_defaults(func=self.do_debug_parse_obj)
+        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get objectfile from the repo")
```
ThomasWaldmann, line 602: check other commands how they describe ID and use similar/same wording.

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +583,61 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+            "data_path", metavar="DATA_PATH", type=str, help="path of the file to write uncompressed data into"
+        )
+        subparser.add_argument(
+            "json_path", metavar="JSON_PATH", type=str, help="path of the json file to write metadata into"
+        )
+
+        # format_obj
+        debug_format_obj_epilog = process_epilog(
+            """
+                This command format the file and metadata into borg objectfile.
```
ThomasWaldmann, line 613: format**s**

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +583,61 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+        )
+        subparser.set_defaults(func=self.do_debug_format_obj)
+        subparser.add_argument(
+            "data_path", metavar="DATA_PATH", type=str, help="path of the file to convert into objectfile"
+        )
+        subparser.add_argument(
+            "json_path", metavar="JSON_PATH", type=str, help="path of the json file to read metadata from"
+        )
+        subparser.add_argument("ctype", metavar="CTYPE", type=int, help="compression type 0..255")
+        subparser.add_argument("clevel", metavar="CLEVEL", type=int, help="compression level 0..255")
```
ThomasWaldmann, line 633: see above, needs `--compression` here instead.

src/borg/repoobj.py
```diff
@@ -29,9 +30,9 @@ def format(
         meta: dict,
         data: bytes,
         compress: bool = True,
-        size: int = None,
-        ctype: int = None,
-        clevel: int = None,
+        size: Optional[int] = None,
+        ctype: Optional[int] = None,
+        clevel: Optional[int] = None,
```
ThomasWaldmann, line 35: why did you change this?

ThomasWaldmann (changes requested):
going in the right direction, but needs some changes for correctness and api consistency.

ThomasWaldmann:
BTW, for coverage and also to make sure nothing is broken, it would be good to have tests for this:

there could be a roundtrip test:
- write some known data to the repo (compute its ID, format it, put it)
- get the data back from the repo (get it from that ID, parse it, check if it is expected data and metadata)

src/borg/repoobj.py
shalearkane: This is the correct way of representing types which can be None. So after looking into this code, I thought to correct this. Though in the later part we assert that they are not None based on conditions. Maybe I should put this into a different PR.

src/borg/archiver/debug_cmd.py
shalearkane: Yes. `get-obj` writes to a file, hence I thought not to re-use the code (or otherwise I will have to write another function which gets the data but does not write the file and the modify the code.

ThomasWaldmann: I didn't mean you call the code in get-obj from here.

What I meant is that the user first uses `borg debug get-obj` to write a borg repo object to a file.
Then that file can be transformed via `borg debug parse-obj` to the metadata and data file.

src/borg/repoobj.py
ThomasWaldmann: Yeah, if this is not required to be here, a separate types cleanup PR for repoobj.py would be welcome.

src/borg/archiver/debug_cmd.py
shalearkane: In the case of `id-hash` function, instead of returning the value, we are printing the output. Hence we need to may need to rewrite the function or return multiple values.

src/borg/archiver/debug_cmd.py
shalearkane: Should I output "ID" too then?

src/borg/archiver/debug_cmd.py
shalearkane: Yes. Also should I use `repr(e)` since it may provide more information about the error? ([link](https://stackoverflow.com/questions/37684153/convert-exception-error-to-string)

src/borg/archiver/debug_cmd.py
ThomasWaldmann: i did not mean that you call that function from here.

what i meant is that the user uses `borg debug id-hash` to get the id.
then the user would invoke `borg debug format-obj ID JSONFILE DATAFILE OBJFILE` (or so) to create a raw obj file (which then could be written to the repo using `borg debug put-obj ...`).

src/borg/archiver/debug_cmd.py
ThomasWaldmann: the user already had to supply the ID to this command, so they already have it.

src/borg/archiver/debug_cmd.py
shalearkane: I am using [traceback.print_exc()](https://docs.python.org/3/library/traceback.html) because if this is a debug command, the user may be interested in knowing where the error occurs. {e} also works but it just provides the classname of the error.

shalearkane:
Thanks for helping and reviewing the code :smile: . I have completed the implementation and I have written the test. I have a question about checking if the metadata is same as the one we had as an input to `format-obj`. 
- The `parsed-obj` metadata has `size`, `ctype`, `clevel` and `csize` as extra fields. Should I check for the validity of these fields too?
- Should I add arbitrary `--compression=xyz,4` argument or leave it as it is to `format-obj`?

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -63,6 +63,55 @@ def test_debug_put_get_delete_obj(self):
...
         assert "is invalid" in output
 
+    def test_debug_id_hash_format_put_parse_obj(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        data = b"some data"
+        meta = b'{"some" : "property"}'
+        meta_dict = json.loads(meta)
+
+        self.create_regular_file("file", contents=data)
+        self.create_regular_file("meta.json", contents=meta)
```
ThomasWaldmann, line 75: you could use data.bin and meta.json to make more clear what's in these files.

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -63,6 +63,55 @@ def test_debug_put_get_delete_obj(self):
         output = self.cmd(f"--repo={self.repository_location}", "debug", "delete-obj", "invalid")
         assert "is invalid" in output
 
+    def test_debug_id_hash_format_put_parse_obj(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        data = b"some data"
+        meta = b'{"some" : "property"}'
```
ThomasWaldmann, line 71: guess this is a bit backwards. :-)

instead of "hand writing" json, start with defining the meta_dict and then json.dumps that into meta_json.

that will work easier and makes sure the json is valid, even if the metadata would be a bit more complicated.

src/borg/testsuite/archiver/debug_cmds.py
ThomasWaldmann: could be even plain.bin and (below) object.bin to make the difference between plain (uncompressed, not encrypted, ...) data and a "formatted" repo object more clear in the filenames.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +267,74 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        with open(args.cdata, "rb") as f:
+            cdata = f.read()
+
+        repo_objs = manifest.repo_objs
+
+        try:
+            meta, data = repo_objs.parse(id=id, cdata=cdata, decompress=True)
+        except Exception:
+            traceback.print_exc()
+            return EXIT_ERROR
```
ThomasWaldmann, line 293: iirc, if you do not catch the exception, it is borg's default behaviour to print a traceback and exit with EXIT_ERROR.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +267,74 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        repo_objs = manifest.repo_objs
+
+        ctype = repo_objs.compressor.ID
+        clevel = repo_objs.compressor.level
+
+        try:
+            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
+        except Exception:
+            traceback.print_exc()
+            return EXIT_ERROR
```
ThomasWaldmann, line 332: same here.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +267,74 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        with open(args.json_path, "r") as f:
+            meta = json.load(f)
+
+        repo_objs = manifest.repo_objs
+
+        ctype = repo_objs.compressor.ID
+        clevel = repo_objs.compressor.level
+
+        try:
+            data_encrypted = repo_objs.format(id=id, meta=meta, data=data, ctype=ctype, clevel=clevel)
```
ThomasWaldmann, line 329: That's too much unneeded effort (my fault, guess i did not document `.format` well enough):

```
repo_obj = repo_objs.format(id=id, meta=meta, data=data)
```

Guess that is all you need, compress=True is the default and it will use the compressor that is already set up as `repo_objs.compressor` to do the compression and also put ctype/clevel into the metadata.

The other arguments are for the use case when it is called with compress=False (then it will not compress, e.g. because the caller already has compressed data, but then it will need to know the uncompressed size, the ctype, the clevel, etc.).

ThomasWaldmann (changes requested):
some more stuff...

src/borg/archiver/debug_cmd.py
```diff
@@ -285,12 +284,7 @@ def do_debug_parse_obj(self, args, repository, manifest):
             cdata = f.read()
 
         repo_objs = manifest.repo_objs
-
-        try:
-            meta, data = repo_objs.parse(id=id, cdata=cdata, decompress=True)
-        except Exception:
-            traceback.print_exc()
-            return EXIT_ERROR
+        meta, data = repo_objs.parse(id=id, cdata=cdata, decompress=True)
```
ThomasWaldmann, line 287: is decompress=True the default value?

if so, leave it away (consistency: you also do not give compress=True in parse-obj).

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -113,6 +113,8 @@ def test_debug_id_hash_format_put_get_parse_obj(self):
         for key, value in meta_dict.items():
             assert meta_read.get(key) == value
 
+        assert meta_read.get("size") == len(data_read)
```
ThomasWaldmann, line 116: can you check whether the compression related metadata entries are as expected?

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +574,72 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+            help="format file and metadata into borg objectfile",
+        )
+        subparser.set_defaults(func=self.do_debug_format_obj)
+        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
+        subparser.add_argument(
+            "data_path", metavar="DATA_PATH", type=str, help="path of the file to convert into objectfile"
+        )
+        subparser.add_argument(
+            "json_path", metavar="JSON_PATH", type=str, help="path of the json file to read metadata from"
+        )
```
ThomasWaldmann, line 626: consistency:
- if the first is DATA_PATH, the second needs to be META_PATH
- otherwise you would refer to what's inside for the first and to the specific file format for the second.

(please globally check this)

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +574,72 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+            "--compression",
+            metavar="COMPRESSION",
+            dest="compression",
+            type=CompressionSpec,
+            default=CompressionSpec("lz4"),
+            help="select compression algorithm, see the output of the " '"borg help compression" command for details.',
+        )
+        subparser.add_argument(
+            "output_path",
+            metavar="OUTPUT_PATH",
```
ThomasWaldmann, line 638: consistency: should be object_path / OBJECT_PATH

(please globally check this)

src/borg/archiver/debug_cmd.py
ThomasWaldmann: also, it would be better for intuition / remembering the order if argument order was 1. meta 2. data.

or, as a completely different approach:

just give one path as cli argument and append `.json` for the metadata and `.bin` for the data file.

ThomasWaldmann (changes requested):

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +266,61 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        # get the object from id
+        hex_id = args.id
+        try:
+            id = unhexlify(hex_id)
+            if len(id) != 32:  # 256bit
+                raise ValueError("id must be 256bits or 64 hex digits")
+        except ValueError as err:
+            print("object id %s is invalid [%s]." % (hex_id, str(err)))
+            return EXIT_ERROR
+
```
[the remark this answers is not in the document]
shalearkane: Where can I see the CompressionSpec definition? Is it in `/home/sm/Public/Sock/borg/src/borg/compress.pyx` :sweat_smile:

Also I looked into other validators. Perhaps we can have a `hex_id_validator` like `text_validator` in `/borg/src/borg/helpers/parseformat.py`  if this check is only going to be used in `debug_cmd`s

src/borg/archiver/debug_cmd.py
shalearkane: My logic was that we are providing two different file-type (so `some_kind_of_data` and `json`) instead of two different file-contents (`data` and `meta`) ... so I thought that `json` will make sense.

Also about the second approach, I thought that explicitly mentioning the filepaths will be much better if someone was trying to debug stuff, since they won't have to constantly rename things to be the same for the command to run. Though I guess if `json` is being generated by some piece of code that the user is using, it will be not be a problem.

src/borg/archiver/debug_cmd.py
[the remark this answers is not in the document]
ThomasWaldmann: You should have a global search function in your developer toolbox.

If you didn't try it yet, maybe have a look at PyCharm.

The code missing should be give to the argparse `type=...` and it will parse and validate the argument early, before it even invokes the function that implements the command.

src/borg/archiver/debug_cmd.py
ThomasWaldmann: ah, so you thought as data as a generalisation of binary or text or whatever. i thought data as opposite of meta.

considering all formats can also be interpreted as binary, even if they are text or some other specific format, you could use **binary** for what you meant.

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -63,6 +64,64 @@ def test_debug_put_get_delete_obj(self):
...
+        for key, value in meta_dict.items():
+            assert meta_read.get(key) == value
+
+        assert meta_read.get("size") == len(data_read)
+
+        c = Compressor(name="zstd", level=2)
+        _, data_compressed = c.compress(meta_dict, data=data)
+        assert meta_read.get("csize") == len(data_compressed)
+        assert "ctype" in meta_read
+        assert "clevel" in meta_read
```
ThomasWaldmann, line 123: you can get the type and level from `c` so you can do an assertion of the value, not just of presence.

shalearkane: I had to add some repeated sequence so that the data compresses, otherwise because of the heuristic it is falling back to CNONE(i.e. No compression) and the test is failing. Is this alright or should I make sure the test work with incompressible data?

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -68,7 +68,7 @@ def test_debug_id_hash_format_put_get_parse_obj(self):
         """Test format-obj and parse-obj commands"""
 
         self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
-        data = b"some data"
+        data = b"some data with repeated sequence 00000000000000000000"
```
ThomasWaldmann, line 71: maybe just use `b"some data" * 100`.

that should compress well enough to avoid a fallback to NONE.

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -86,7 +86,7 @@ def test_debug_id_hash_format_put_get_parse_obj(self):
             "input/plain.bin",
             "input/meta.json",
             "output/data.bin",
-            "--compression=zstd,2",
+            "--compression=zstd,3",
```
ThomasWaldmann: No, please keep level at 2.

3 is the default zstd level and we want something special (non-standard) to detect.

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -116,11 +116,11 @@ def test_debug_id_hash_format_put_get_parse_obj(self):
 
         assert meta_read.get("size") == len(data_read)
 
-        c = Compressor(name="zstd", level=2)
+        c = Compressor(name="zstd", level=3)
```
ThomasWaldmann: same here, keep 2.

src/borg/testsuite/archiver/debug_cmds.py
ThomasWaldmann: Yeah, one needs a good amount of compressible data so it does not fall back to NONE compression (because that has smaller overhead).

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +574,72 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+            description=self.do_debug_parse_obj.__doc__,
+            epilog=debug_parse_obj_epilog,
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            help="parse objectfile into meta dict and data",
+        )
+        subparser.set_defaults(func=self.do_debug_parse_obj)
+        subparser.add_argument("id", metavar="ID", type=str, help="hex object ID to get from the repo")
+        subparser.add_argument(
+            "cdata", metavar="COMPRESSED_DATA", type=str, help="path of the objectfile to parse data from"
+        )
```
ThomasWaldmann, line 596: consistency: isn't this the same thing you called object_path / OBJECT_PATH below?

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -63,6 +64,64 @@ def test_debug_put_get_delete_obj(self):
         output = self.cmd(f"--repo={self.repository_location}", "debug", "delete-obj", "invalid")
         assert "is invalid" in output
 
+    def test_debug_id_hash_format_put_get_parse_obj(self):
+        """Test format-obj and parse-obj commands"""
+
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        data = b"some data" * 100
+        meta_dict = {"some": "property"}
+        meta = json.dumps(meta_dict, ensure_ascii=False).encode()
```
ThomasWaldmann, line 73: do you need to give `ensure_ascii=False`?

src/borg/testsuite/archiver/debug_cmds.py
```diff
@@ -63,6 +64,64 @@ def test_debug_put_get_delete_obj(self):
...
+
+        output = self.cmd(
+            f"--repo={self.repository_location}",
+            "debug",
+            "parse-obj",
+            id_hash,
+            "output/compressed.bin",
+            "output/plain.bin",
+            "output/meta.json",
+        )
```
ThomasWaldmann, line 106: compressed.bin -> objectfile.bin?

ThomasWaldmann (commented):
looks good mostly, just some consistency and minor changes.

src/borg/testsuite/archiver/debug_cmds.py
shalearkane: Yes, https://stackoverflow.com/questions/18337407/saving-utf-8-texts-with-json-dumps-as-utf-8-not-as-a-u-escape-sequence

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +266,61 @@ def do_debug_id_hash(self, args, repository, manifest):
...
+        try:
+            id = unhexlify(hex_id)
+            if len(id) != 32:  # 256bit
+                raise ValueError("id must be 256bits or 64 hex digits")
+        except ValueError as err:
+            print("object id %s is invalid [%s]." % (hex_id, str(err)))
+            return EXIT_ERROR
+
+        with open(args.object_path, "rb") as f:
+            object_file = f.read()
```
ThomasWaldmann, line 284: you need to be more careful with names. names are important - if they are appropriate, it makes the code much easier to read and understand.

if something is called `..._file` the first guess would be it is a python file object. a slightly sloppier guess might be that it is a OS-level file handle or a file name or path.

but, what you have in there is just a bytes object (binary data), so `data` or `cdata` (considering 287) or `data_encrypted` (see 318) would be more appropriate.

src/borg/archiver/debug_cmd.py
```diff
@@ -265,6 +266,61 @@ def do_debug_id_hash(self, args, repository, manifest):
         print(id.hex())
         return EXIT_SUCCESS
 
+    @with_repository(compatibility=Manifest.NO_OPERATION_CHECK)
+    def do_debug_parse_obj(self, args, repository, manifest):
+        """parse object file into meta dict and data (decrypting, decompressing)"""
```
ThomasWaldmann, line 271: maybe add "borg" here, like "parse borg object file ..." (see 299).

src/borg/archiver/debug_cmd.py
```diff
@@ -518,6 +574,72 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
...
+                """
+        )
+        subparser = debug_parsers.add_parser(
+            "parse-obj",
+            parents=[common_parser],
+            add_help=False,
+            description=self.do_debug_parse_obj.__doc__,
+            epilog=debug_parse_obj_epilog,
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            help="parse objectfile into meta dict and data",
```
ThomasWaldmann, line 590: also add "borg" here, please, see 617.

ThomasWaldmann (changes requested):
ok, final review with some more consistency/clarity nitpicks.

will merge after fixes (and squash all the commits into one).

ThomasWaldmann:
Thanks!

PR #7452, opened 2023-03-19
Do not retry on permission errors

### LINKED ISSUE #7446: non-readable files in source directory massively stall backup ("borg create")

yes

Issue / Feature Request / bug?

borg-2.0.0b5

Ubuntu Server 22.04

Intel(R) Core(TM) i3-8145U CPU @ 2.10GHz
16 GB RAM
ext4

Not much, just test data. Maybe 5 MB.

`/path/to/borg-2.0.0b5-linuxnew64 create --stats --list --checkpoint-interval=60 --repo /path/to/testrepo1 --verbose archive-{utcnow} /etc`

I am getting my feet wet with the new syntax and related repositories coming in borg 2.0. For that purpose I created some test repos and backed up my /etc directory (as a dummy source directory - a directory with many text files which do not change while running borg). Some files under this directory are not readable for the user borg runs under. This causes borg to log a warning and retry for a total of about 15 seconds - for each file. Example:

```
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.001s, next: retry: 1 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.003s, next: retry: 2 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.010s, next: retry: 3 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.032s, next: retry: 4 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.100s, next: retry: 5 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 0.316s, next: retry: 6 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 1.000s, next: retry: 7 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 3.162s, next: retry: 8 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README', slept 10.000s, next: retry: 9 of 9...
/etc/sudoers.d/README: open: [Errno 13] Permission denied: 'README'
E /etc/sudoers.d/README
```
While waiting for the file(s) to become accessible (which never happens, of course) the backup stalls. This causes the backup to run for about 14 minutes for a measly 4.6 MB of source files. For comparison: Running with the correct permissions (i.e. all files are readable by borg) the backup take WAY less than one second.

Suggestion / feature request:

Either make waiting for "Permission denied" errors non-blocking by using asyncio or threads (--> backup can continue while borg waits) and/or provide an option to skip non-accessible files. 

yes - does not appear to be a bug but rather "works as intended". But maybe not very smart... ;-)

None

snsmac:
Should fix #7446


src/borg/archiver/create_cmd.py
```diff
@@ -370,6 +371,10 @@ def _process_any(self, *, path, parent_fd, name, st, fso, cache, read_special, d
                     self.print_warning("Unknown file type: %s", path)
                     return
             except (BackupError, BackupOSError) as err:
+                if err is BackupOSError:
```
ThomasWaldmann, line 374: I don't think this would work (guess you did not try?) as you are checking identity of a BackupOSError **instance** with the **class** - you need to use `isinstance()`.

src/borg/archiver/create_cmd.py
```diff
@@ -370,6 +371,10 @@ def _process_any(self, *, path, parent_fd, name, st, fso, cache, read_special, d
...
+                    if err.errno == errno.EACCES or err.errno == errno.EPERM:
```
ThomasWaldmann, line 375: in case we might get more such errnos that must not be retried, it also would be better to use something like `err.errno in (errno.EACCESS, errno.EPERM, )`, so we can just add new errnos at the end of that tuple.

finally, did you consider what the exact conditions leading to EACCES and EPERM are and whether both should not be retried? is this the same on all OSes?

src/borg/archiver/create_cmd.py
```diff
@@ -370,6 +371,10 @@ def _process_any(self, *, path, parent_fd, name, st, fso, cache, read_special, d
...
+                        # Do not try again, permission errors can not be fixed by retrying.
```
ThomasWaldmann, line 376: in case we add other errors later, maybe rather say "such errors" instead "permission errors".

ThomasWaldmann (changes requested):
thanks for the PR! some stuff i found.

src/borg/archiver/create_cmd.py
snsmac: According to the [python documentation](https://docs.python.org/3.10/library/errno.html#errno.EACCES), which is a bit lacking in that aspect imo, they corresponded to "linux/include/errno.h", which references t[he posix standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_03) 
This defines EACCES and EPERM the following way:

- [EACCES] Permission denied. An attempt was made to access a file in a way forbidden by its file access permissions.
- [EPERM] Operation not permitted. An attempt was made to perform an operation limited to processes with appropriate privileges or to the owner of a file or other resource.

As I only very rarely use not-linux OSes I don't know exactly to what error conditions these map on other systems

ThomasWaldmann: OK, both sound like permanent issues retrying won't fix.

@RayyanAnsari any comments about that and windows?

ThomasWaldmann: btw, on linux (and other unix), this could be tested by backing up a file with `chmod 000` permissions.

RayyanAnsari: Looks fine, EACCESS and EPERM both map to PermissionError.
A test for this would be good, `icacls` could be used on Windows to deny all permissions.

snsmac: I will try adding tests next weekend, if someone else wants to work on that before I wont mind

ThomasWaldmann: I added a test which works on macOS (and likely also other UNIXoid OSes), but might need more work for windows. @RayyanAnsari

ThomasWaldmann: @RayyanAnsari it executes the `os.chmod(fn, 0o000)` on windows without complaining, but that does not revoke read permissions for owner, so the test fails on win32. can you add correct code for win32?

ThomasWaldmann: Amended, let's see. Thanks for helping!

ThomasWaldmann (approved):
LGTM!

src/borg/archiver/create_cmd.py
ThomasWaldmann: Works.

PR #7440, opened 2023-03-13
Add --format option to `borg diff`, resolve issue #4634

RF-Tar-Railt:
!!! warning
    this pull request may cause conflict with #7414 

- [x] add `--format` in `diff_cmd.py`
- [x] rewrite `ItemDIff` in `item.pyx`
- [x] keep `--content-only` and `--json-lines`
- [x] rewrite part of `BaseFormatter` and subclass
- [x] write `DiffFormatter`
- [x] use `DiffFormatter` in `diff_cmd`  
- [x] test

Issue #7449, opened 2023-03-18
"borg delete --cache-only" does not work as expected


agrommek:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Bug

#### Your borg version (borg -V).

borg-2.0.0b5

#### Operating system (distribution) and version.

Arch Linux, Ubuntu Server 22.04

#### Hardware / network configuration, and filesystems used.

Local Borg repositories for testing, ext4

#### How much data is handled by borg?

A few GB, for testing

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

```
$  borg-2.0.0b5-linuxnew64 delete --repo /path/to/testrepo1  --cache-only --verbose
Enter passphrase for key /mnt/HDD/scratch/testrepos/testrepo1: 
Aborting: if you really want to delete all archives, please use -a 'sh:*' or just delete the whole repository (might be much faster).
```
## Describe the problem you're observing.

Expected behavior: 
From the docs and from the name of the switch I would expect that borg would clean the cache, i.e. delete the files cache and/or delete the chunks cache and/or delete the contents of chunks.archives.d. It should not delete or modify stored archives.

Observed behavior:
Borg refuses to run when no `--match-archives` switch is also given. There is different behavior, depending on the value of this parameter:

1. `--cache-only --match-archives 'id:some-not-existing archive'`: nothing gets deleted (as the archive does not exist), cache files have same size as before. No deletion of cache.
2. `--cache-only --match-archives 'id:some-existing archive'`: archive gets deleted, cache files have same size as before. No deletion of cache.
3. `--cache-only --match-archives 'sh:*`: repo is empty afterwards, chunks cache is small (empty?), files cache remains.

In my opinion this is a bug, as it violates the "principle of least surprise". If an option switch says "cache-only", then it should operate only on the cache. On the other hand, if this switch is supposed to somehow only delete the cache for the archives specified with `--match-archives`, then this behavior should be documented much more extensively.

ThomasWaldmann:
Hmm, that sounds weird. Maybe something got broken due to the misc other changes in borg2. I'll have a look.

ThomasWaldmann:
2 issues here:
- one should use `rdelete` as this is a repo-level (not archive-level) operation.
- borg erroneously accepts `--cache-only` also for `delete`, that is a bug.

ThomasWaldmann:
Fixes by #7455.

PR #7455, opened 2023-03-20
delete: remove --cache-only option, fixes #7440

ThomasWaldmann:
for deleting the cache only, use:

borg rdelete --cache-only


ThomasWaldmann:
Should have closed #7449.

PR #7460, opened 2023-03-21
Add --ignore-zeros flag to import-tar

### LINKED ISSUE #7432: import-tar for multiple or concatenated tar files

Yes

Question

borg 1.2.3

400-500 GB

#### Description

tl;dr: could `borg import-tar` support multiple tar archives, or a file made by simply `cat`-entating several tar files?

Use case / situation: a few hundred GB of exported data is split into multiple tar files by size (50 GB each) rather than content.

Solution A: use `tar --catenate` to combine those into a single tar, then use `borg import-tar`.  Unfortunately GNU tar has a bug that prevents it from concatenating more than one archive at a time, and concatenating one by one is too expensive due to implementation (see https://superuser.com/a/1098273).

Solution B: use `cat 1.tar 2.tar ...` and feed that into `borg import-tar`.  Simply catenating tar archives doesn't produce a completely valid combined archive due to zero-padding.  That padding can be ignored during extraction with `--ignore-zeros` flag of `tar` (https://www.gnu.org/software/tar/manual/html_node/concatenate.html) but that isn't currently supported by `borg import-tar` as far as I can tell from experimenting.  Perhaps this could be implemented similar to libarchive (https://github.com/libarchive/libarchive/issues/457).

What would work is either `borg import-tar` supporting multiple files in a command line, or `--ignore-zeros` inspired option.  I have a preference for the latter for simplicity.

Note the issue is similar to #6807 by title, but that question is about having a directory per tar archive, which is unpreferred in this case.

dotdoom:
Fixes #7432.

### DIFF
--- a/src/borg/archiver/tar_cmds.py
+++ b/src/borg/archiver/tar_cmds.py
@@ -289,7 +289,7 @@ class TarMixIn:
             file_status_printer=self.print_file_status,
         )
 
-        tar = tarfile.open(fileobj=tarstream, mode="r|")
+        tar = tarfile.open(fileobj=tarstream, mode="r|", ignore_zeros=args.ignore_zeros)
 
         while True:
             tarinfo = tar.next()
@@ -487,6 +487,12 @@ class TarMixIn:
             help="only display items with the given status characters",
         )
         subparser.add_argument("--json", action="store_true", help="output stats as JSON (implies --stats)")
+        subparser.add_argument(
+            "--ignore-zeros",
+            dest="ignore_zeros",
+            action="store_true",
+            help="ignore zero-filled blocks in the input tarball",
+        )
 
         archive_group = subparser.add_argument_group("Archive options")
         archive_group.add_argument(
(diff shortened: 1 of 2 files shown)


ThomasWaldmann (approved):
LGTM!

Just for science: did you try same test, but without --ignore-zeros?

dotdoom:
@ThomasWaldmann I haven't.  I think it's a good idea.  Let me give it a shot.  Would you like to also see it as a case in the testsuite, as a negative test?

ThomasWaldmann (approved):

ThomasWaldmann:
Hmm, maybe mention in the command description that this flag is useful for concatenated tar archives?

dotdoom:
Sure, done!  Nothing to add from my side.

ThomasWaldmann (approved):

ThomasWaldmann:
If you like: a backport of this to 1.2-maint branch would be great. I'll release 1.2.4 soon.

dotdoom:
By all means.  #7465

PR #7475, opened 2023-03-25
Ensuring cache presence before starting borg create. Fixes #7450

### LINKED ISSUE #7450: borg crashes with exception when cache files are missing, should handle this gracefully

yes

Bug / Enhancement

borg-2.0.0b5

Arch Linux, Ubuntu Server 22.04

Local repositories for testing, ext4

a few GB, for testing

I messed a bit around in the cache directory in an effort to save some space as this is rather big in my production setup (several GB) and noticed the following behavior: When the whole cache directory is deleted, borg re-sycns the cache from the repository and/or rebuilds the files cache an other data. However, when only part of the cache directory is deleted (only files cache, only chunks cache, only config file, ...), borg exits with an exception:

```
$ borg-2.0.0b5-linuxnew64 create downloads-{utcnow} /home/me/Downloads/ --repo /path/to/testrepo1  --progress --verbose --stats
Enter passphrase for key /path/to/testrepo1: 
Creating archive at "/path/to/testrepo1"
Local Exception
  File "borg/archiver/__init__.py", line 595, in main
  File "borg/archiver/__init__.py", line 523, in run
  File "borg/archiver/_common.py", line 156, in wrapper
  File "borg/archiver/create_cmd.py", line 221, in do_create
  File "borg/cache.py", line 426, in __new__
  File "borg/cache.py", line 411, in local
  File "borg/cache.py", line 525, in __init__
  File "borg/cache.py", line 575, in open
  File "borg/cache.py", line 698, in rollback
  File "borg/cache.py", line 560, in _do_open
  File "borg/crypto/file_integrity.py", line 133, in __init__
FileNotFoundError: [Errno 2] No such file or directory: '/home/me/.cache/borg/74188eaf53161b51a9a6f23140ea342099a0a4d3e84c1d09ff125aca9a0ae1b6/chunks'

Platform: Linux tomateX 6.2.6-arch1-1 #1 SMP PREEMPT_DYNAMIC Mon, 13 Mar 2023 17:02:08 +0000 x86_64
Linux: Unknown Linux  
Borg: 2.0.0b5  Python: CPython 3.11.2 msgpack: 1.0.4 fuse: llfuse 1.4.2 [pyfuse3,llfuse]
PID: 4102  CWD: /home/me/.cache/borg/74188eaf53161b51a9a6f23140ea342099a0a4d3e84c1d09ff125aca9a0ae1b6
sys.argv: ['/home/me/Downloads/borg-2.0.0b5-linuxnew64', 'create', 'downloads-{utcnow}', '/home/me/Downloads/', '--repo', '/path/to/testrepo1', '--progress', '--verbose', '--stats']
SSH_ORIGINAL_COMMAND: None
```
Borg should / could handle this situation more gracefully by doing one of those:

- Exit gracefully and inform the user that something is wrong with their cache directory and that they should go delete it completely.
- Offer / provide a switch, maybe `--repair-cache` or `--rebuild-cache` or so, to make repairing the cache an explicit user action.
- Log a warning, delete the cache directory and start rebuilding it.

Maybe there are other things borg could do. In any case, users should not need to interpret python stack traces when something goes wrong. One could reply that users also shouldn't muck around in the cache directory, but some will do (as I did... ;-) )

Michael-Girma:
@ThomasWaldmann These are the changes for #7450. Right now, it recreates the config-cache if not found, and recreates the chunk-cache and the config-cache(to remove integrity data), if the chunk-cache is not found. Think there is a better/more efficient way to do this?

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -514,6 +514,15 @@ class LocalCache(CacheStatsMixin):
         self.txn_active = False
 
         self.path = cache_dir(self.repository, path)
+        if os.path.exists(self.path) and not self.check_cache_integrity():
+            logger.warning("Rebuilding cache as some cache-files have gone missing")
+            if not os.path.exists(os.path.join(self.path, "chunks")):
+                self.recreate_cache_config(delete_existing=True)  # recreating to remove integrity data stored in config
+                self.chunks = ChunkIndex()
+                self.chunks.write(os.path.join(self.path, "chunks"))
+            elif not os.path.exists(os.path.join(self.path, "config")):
+                self.recreate_cache_config()
+
         self.security_manager = SecurityManager(self.repository)
         self.cache_config = CacheConfig(self.repository, self.path, lock_wait)
 
@@ -900,6 +909,13 @@ class LocalCache(CacheStatsMixin):
             self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
+    def check_cache_integrity(self):
+        cache_files = ["chunks", "config"]
+        for file in cache_files:
+            if not os.path.exists(os.path.join(self.path, file)):
+                return False
+        return True
+
     def check_cache_compatibility(self):
         my_features = Manifest.SUPPORTED_REPO_FEATURES
         if self.cache_config.ignored_features & my_features:
@@ -921,6 +937,15 @@ class LocalCache(CacheStatsMixin):
         self.chunks = ChunkIndex()
         with SaveFile(os.path.join(self.path, files_cache_name()), binary=True):
             pass  # empty file
+        self.recreate_cache_config()
+
+    def recreate_cache_config(self, delete_existing=False):
+        if delete_existing and os.path.exists(os.path.join(self.path, "config")):
+            os.unlink(os.path.join(self.path, "config"))
+        if not os.path.exists(os.path.join(self.path, "config")):
+            self.cache_config = CacheConfig(self.repository, self.path)
+            self.cache_config.create()
+            self.cache_config.load()  # loading config here to initialize property `_config`
         self.cache_config.manifest_id = ""
         self.cache_config._config.set("cache", "manifest", "")
 
--- a/src/borg/testsuite/archiver/create_cmd.py
+++ b/src/borg/testsuite/archiver/create_cmd.py
@@ -13,6 +13,7 @@ import pytest
 
 from ... import platform
 from ...constants import *  # NOQA
+from ...helpers import get_cache_dir
 from ...manifest import Manifest
 from ...platform import is_cygwin, is_win32, is_darwin
 from ...repository import Repository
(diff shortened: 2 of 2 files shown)


src/borg/cache.py
```diff
@@ -900,6 +909,13 @@ def create_master_idx(chunk_idx):
             self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
+    def check_cache_integrity(self):
```
ThomasWaldmann, line 912: this only tests existence, not integrity.

how about:
```
return all(os.path.exists(os.path.join(self.path, file)) for file in "chunks", "config")
```

src/borg/cache.py
```diff
@@ -514,6 +514,15 @@ def __init__(
         self.txn_active = False
 
         self.path = cache_dir(self.repository, path)
+        if os.path.exists(self.path) and not self.check_cache_integrity():
```
ThomasWaldmann, line 517: isn't the first term redundant as the check call would also return False in that case?

src/borg/cache.py
```diff
@@ -514,6 +514,15 @@ def __init__(
...
+            logger.warning("Rebuilding cache as some cache-files have gone missing")
```
ThomasWaldmann, line 518: i guess this should be more precise.

src/borg/cache.py
```diff
@@ -514,6 +514,15 @@ def __init__(
...
+            if not os.path.exists(os.path.join(self.path, "chunks")):
+                self.recreate_cache_config(delete_existing=True)  # recreating to remove integrity data stored in config
+                self.chunks = ChunkIndex()
+                self.chunks.write(os.path.join(self.path, "chunks"))
```
ThomasWaldmann, line 522: this will create an empty (and thus incorrect) chunks index on disk, how is that helpful?

also creation of that file is done in a rather unsafe way.

src/borg/cache.py
ThomasWaldmann: also: only called from one place.

maybe rather do the joins only once and have the rest in-line?

src/borg/cache.py
ThomasWaldmann: also: no "-" in english.

src/borg/cache.py
```diff
@@ -921,6 +937,15 @@ def wipe_cache(self):
         self.chunks = ChunkIndex()
         with SaveFile(os.path.join(self.path, files_cache_name()), binary=True):
             pass  # empty file
+        self.recreate_cache_config()
+
+    def recreate_cache_config(self, delete_existing=False):
```
ThomasWaldmann, line 942: maybe call this `create...` - it only mutates to the more special `recreate_...` in you give `delete_existing=True`.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -659,6 +660,30 @@ def test_file_status_cs_cache_mode(self):
...
+    @pytest.mark.allow_cache_wipe
+    def test_cache_rebuild_on_missing_cache_files(self):
+        """test if borg rebuilds cache if any of the cache files have gone missing"""
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.create_regular_file("file1", size=10)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive1", "input")
+        info = json.loads(self.cmd(f"--repo={self.repository_location}", "rinfo", "--json"))
+        repository = info["repository"]
+        cache_dir = os.path.join(get_cache_dir(), repository["id"])
+        config_cache_path = os.path.join(cache_dir, "config")
```
ThomasWaldmann, line 672: this is the cache config, not the config cache.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -659,6 +660,30 @@ def test_file_status_cs_cache_mode(self):
...
+        chunk_cache_path = os.path.join(cache_dir, "chunks")
```
ThomasWaldmann, line 673: chunks_cache_path

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -659,6 +660,30 @@ def test_file_status_cs_cache_mode(self):
...
 
+    @pytest.mark.allow_cache_wipe
+    def test_cache_rebuild_on_missing_cache_files(self):
+        """test if borg rebuilds cache if any of the cache files have gone missing"""
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.create_regular_file("file1", size=10)
+        self.cmd(f"--repo={self.repository_location}", "create", "archive1", "input")
+        info = json.loads(self.cmd(f"--repo={self.repository_location}", "rinfo", "--json"))
+        repository = info["repository"]
+        cache_dir = os.path.join(get_cache_dir(), repository["id"])
```
ThomasWaldmann, line 671: repo_cache_dir

ThomasWaldmann (changes requested):
thanks for the PR - some stuff i noticed.

src/borg/cache.py
Michael-Girma: Well, I was assuming that creating an empty chunks cache should be enough for ```borg create``` to continue and re-populate it as it normally would. 
1. Should we be entirely rebuilding it here?
2. How else would be creating the file. Is there another piece of code that I can take a look at?

ThomasWaldmann: borg create needs the chunks cache in a valid state (refcounts, sizes for all chunks in the repo) and updates refcounts on that basis.

ThomasWaldmann:
Closing this because there was a recent PR rebuilding missing/corrupt cache files.

PR #7477, opened 2023-03-25
Added pre-commit for linting purposes. Fixes #7476

### LINKED ISSUE #7476: Install pre-commit to avoid un-linted/unformatted code from being committed

Enhancement

As a contributor, I've passed in several commits that haven't been linted/formatted after forgetting to run ```black``` and ```flake8```. This could be avoided by running pre-commit hooks to analyze the code before the commit is made thus circumventing unformatted code from being committed.

Michael-Girma:
Added pre-commit as a requirement and integrated black and flake8 lint scripts to execute before commits. Changed docs to notify fellow contributors to install the pre-commit hooks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+-   repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+    -   id: black
+-   repo: https://github.com/pycqa/flake8
+    rev: 6.0.0
+    hooks:
+    -   id: flake8
+        files: '(src|scripts|conftest.py)'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 22.10.0
+    rev: 23.1.0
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
+-   repo: https://github.com/psf/black
+    rev: 23.1.0
```
ThomasWaldmann, line 3: is it possible to say "23.latest" here somehow?

docs/development.rst
```diff
@@ -179,6 +179,13 @@ virtual env and run::
   pip install -r requirements.d/development.txt
 
 
+This project utilizes pre-commit to format and lint code before it is committed.
+Although pre-commit is installed when running the command above, the pre-commit hooks
+will have to be installed separately. Run this command to install the pre-commit hooks::
+
+  pre-commit install
+
+
```
ThomasWaldmann, line 188: Great! :-)

.pre-commit-config.yaml
ThomasWaldmann: ok, then we'll leave it as is. We'll notice it if things get out of sync.

Issue #7459, opened 2023-03-21
please provide borg-linuxold64 for borg-2.0.0 and beta releases


agrommek:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Question / polite request :-)

#### Your borg version (borg -V).

borg-2.0.0b5

#### Operating system (distribution) and version.

Synology NAS / DSM 7.1.1-42962 Update 4
Linux kernel 3.10.180 (--> old, but cannot be helped)
glibc 2.26, compiled with GCC 8.5.0 (also old, cannot be helped either)

#### Hardware / network configuration, and filesystems used.

ext4, BTRFS

## Describe the problem you're observing.

As the title says: Please continue to provide linux binaries for glibc < 2.28 (i.e. borg-linuxold64) for borg 2.0 and the beta releases.

#### Background

I run borg on a Synology NAS using the precompiled/packaged binaries. Worked great for the last 6 years or so. Right now I am in the process to modify my wrapper scripts for the new borg2 command line syntax. 

However, borg-linux64 does not run on my NAS, the glibc (version 2.26) appears to be too old. AFAIK there is no newer NAS-OS (Synology DSM) with a newer glibc. I cannot build borg-2.0.0 releases on the NAS because there are development headers (OpenSSL, probably others) missing and the pre-installed version of Python is too old anyway. There may be the option to cross-compile borg as a Synology package (instructions see [here](https://help.synology.com/developer-guide/)). But the developer guide appears to be targeted at people who know exactly what they do and are used to cross-compile stuff for different platforms - which does not include myself, unfortunately. Just fetching the binary from the releases page is far easier.

Please continue to provide borg-linuxold64 for borg-2.0 and beta releases.

Thanks for the great work!

ThomasWaldmann:
the "linuxold" binaries for borg 1.2.x were made on a Debian stretch machine.

i tried adding the stretch machine to master branch, but it didn't work for building borg in master branch, so i removed it again. #7010

ThomasWaldmann:
Hmm, guess one option could be to install xxhash header and lib from source. PRs or bounties welcome!

bbx0:
Another option might be [staticx](https://staticx.readthedocs.io/en/latest/introduction.html). I was testing it as an alternative to the distroless images while looking into how to provide borg via docker. I decided against it, to ensure a fully scannable SBOM of the image content. Still, the build seemed to be fine.

[...]

PR #7478, opened 2023-03-26
Add support to build dependencies from source to the Vagrantfile

### LINKED ISSUE #7010: try to revive borg-linuxold64
looks like rhel7 & co does not work with the newer binaries.

so, check if the debian stretch vm can be revived for borg2.

bbx0:
Fix #7010 #7459

This adds support to build the borg dependencies from source and re-adds the stretch64 VM.

* Libraries: LIBATTR=2.5.1, LIBACL=2.3.1, LIBFUSE=3.14.0, LIBLZ4=1.9.4, LIBZSTD=1.5.4, LIBXXHASH=0.8.1, OPENSSL=1.1.1t
* `pyenv` will detect the custom openssl version and use it. This is needed to build python >= 3.10 with openssl >= 1.1.1.
* The stretch64 VM is re-added.

The stretch64 `llfuse` tests seem okay, except 3 issues caused by fakeroot. This is likely unrelated.

The stretch64 `pyfuse3` tests show a few errors for the borg.exe, which is not built with pyfuse3. This seems due to how the FUSE_IMPL is detected.
```
borg mount not available: no FUSE support, BORG_FUSE_IMPL=pyfuse3. 
```
* A borg.exe built with llfuse, will fail the tests when called with `BORG_FUSE_IMPL=pyfuse3` and fall through to `None`.
* A borg.exe built with pyfuse3 will not work with the current `borg.exe.spec`. It excludes the bundled `trio/_ssl.py` import from pyfuse3 causing a silent fall-through to `None`.
  * Replace `fuse_impl.py` line 15 `pass`, with a RuntimeError to show:
 ```
File "borg/fuse_impl.py", line 13, in <module>
File "src/pyfuse3.pyx", line 68, in init pyfuse3
File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
File "trio/__init__.py", line 74, in <module>
File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
File "trio/_ssl.py", line 153, in <module>
ModuleNotFoundError: No module named 'ssl'
```
* A workaround like `mv /vagrant/borg/borg.exe /vagrant/borg/borg.exe~` will skip the binary tests and work.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -418,6 +418,23 @@ Vagrant.configure(2) do |config|
     b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("buster64", ".*none.*")
   end
 
+  config.vm.define "stretch64" do |b|
+    b.vm.box = "debian/stretch64"
+    b.vm.provider :virtualbox do |v|
+      v.memory = 1024 + $wmem
+    end
+    b.vm.provision "fs init", :type => :shell, :inline => fs_init("vagrant")
+    b.vm.provision "packages debianoid", :type => :shell, :inline => packages_debianoid("vagrant")
+    b.vm.provision "install source dependencies", :type => :shell, :privileged => true, :inline => install_source_dependencies("vagrant")
+    b.vm.provision "install pyenv", :type => :shell, :privileged => false, :inline => install_pyenv("stretch64")
+    b.vm.provision "install pythons", :type => :shell, :privileged => false, :inline => install_pythons("stretch64")
+    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("stretch64")
+    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg("llfuse")
+    b.vm.provision "install pyinstaller", :type => :shell, :privileged => false, :inline => install_pyinstaller()
+    b.vm.provision "build binary with pyinstaller", :type => :shell, :privileged => false, :inline => build_binary_with_pyinstaller("stretch64")
+    b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("stretch64", ".*none.*")
+  end
+
   config.vm.define "freebsd64" do |b|
     b.vm.box = "generic/freebsd13"
     b.vm.provider :virtualbox do |v|
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,7 +15,8 @@ def packages_debianoid(user)
     apt-get -y -qq update
     apt-get -y -qq dist-upgrade
     # for building borgbackup and dependencies:
-    apt install -y libssl-dev libacl1-dev libxxhash-dev liblz4-dev libzstd-dev pkg-config
+    apt install -y pkg-config
+    apt install -y libssl-dev libacl1-dev libxxhash-dev liblz4-dev libzstd-dev || true
     apt install -y libfuse-dev fuse || true
     apt install -y libfuse3-dev fuse3 || true
     apt install -y locales || true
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Vagrantfile [hunk 1]
```diff
@@ -15,7 +15,8 @@ def packages_debianoid(user)
     apt-get -y -qq update
     apt-get -y -qq dist-upgrade
     # for building borgbackup and dependencies:
-    apt install -y libssl-dev libacl1-dev libxxhash-dev liblz4-dev libzstd-dev pkg-config
+    apt install -y pkg-config
+    apt install -y libssl-dev libacl1-dev libxxhash-dev liblz4-dev libzstd-dev || true
```
ThomasWaldmann, line 19: guess this will fail to install the remainder of the pkgs, if libxxhash-dev can not be found (like on stretch).

Vagrantfile
```diff
@@ -330,6 +418,23 @@ Vagrant.configure(2) do |config|
...
+    b.vm.box = "debian/stretch64"
+    b.vm.provider :virtualbox do |v|
+      v.memory = 1024 + $wmem
+    end
+    b.vm.provision "fs init", :type => :shell, :inline => fs_init("vagrant")
+    b.vm.provision "packages debianoid", :type => :shell, :inline => packages_debianoid("vagrant")
+    b.vm.provision "install source dependencies", :type => :shell, :privileged => true, :inline => install_source_dependencies("vagrant")
+    b.vm.provision "install pyenv", :type => :shell, :privileged => false, :inline => install_pyenv("stretch64")
+    b.vm.provision "install pythons", :type => :shell, :privileged => false, :inline => install_pythons("stretch64")
+    b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("stretch64")
```
ThomasWaldmann, line 431: hmm, if we first install_pythons, we'll have a recent python even on stretch, which could be used for install_source_dependencies.

ThomasWaldmann (commented):
thanks for the PR!

Vagrantfile
bbx0: Yes, it fails then for all of them. The locally built libs take priority in pkg-config and the DPKG variants would not be used, when `install source dependencies` is active.

Vagrantfile
bbx0: I thought so too and had it working that way before but it makes it a bit awkward with pyfuse3 and openssl. Two functions are needed in this case.

A recent openssl is required to build python and must be installed before pyenv runs.

Pyfuse3 needs root to setuid for fusermount3. The pyenv runs with user vagrant but it does work with sudo ninja install.

ThomasWaldmann: ah, ok, so guess we keep it as is.

Vagrantfile
ThomasWaldmann: ok

ThomasWaldmann:
to solve the pyfuse3 issue, just use llfuse (like we also use otherwise when building the borg.exe)?

bbx0:
Do you remember why `_ssl, ssl` was excluded from the spec file? When I clean the exclude and provide `_pyfuse3` as hiddenimport, the borg.exe can be build with `pyfuse3,llfuse` and then tests are passed for `fuse2` and `fuse3`.

Skipping `fuse3` in the tests is of course also an option, when we just go with `llfuse`.

https://github.com/bbx0/borg/compare/master...bbx0:borg:fix-pyfuse3-tests

```diff
diff --git a/Vagrantfile b/Vagrantfile
index a0a0db4bf8..8a1f83d547 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -429,7 +429,7 @@ Vagrant.configure(2) do |config|
     b.vm.provision "install pyenv", :type => :shell, :privileged => false, :inline => install_pyenv("stretch64")
     b.vm.provision "install pythons", :type => :shell, :privileged => false, :inline => install_pythons("stretch64")
     b.vm.provision "build env", :type => :shell, :privileged => false, :inline => build_pyenv_venv("stretch64")
-    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg("llfuse")
+    b.vm.provision "install borg", :type => :shell, :privileged => false, :inline => install_borg("pyfuse3,llfuse")
     b.vm.provision "install pyinstaller", :type => :shell, :privileged => false, :inline => install_pyinstaller()
     b.vm.provision "build binary with pyinstaller", :type => :shell, :privileged => false, :inline => build_binary_with_pyinstaller("stretch64")
     b.vm.provision "run tests", :type => :shell, :privileged => false, :inline => run_tests("stretch64", ".*none.*")
diff --git a/scripts/borg.exe.spec b/scripts/borg.exe.spec
index 473eb386c7..2aca61390f 100644
--- a/scripts/borg.exe.spec
+++ b/scripts/borg.exe.spec
@@ -12,7 +12,7 @@ basepath = os.path.abspath(os.path.join(here, '..'))
 if is_win32:
     hiddenimports = []
 else:
-    hiddenimports = ['borg.platform.posix', 'pkg_resources.py2_warn', ]
+    hiddenimports = ['borg.platform.posix', 'pkg_resources.py2_warn', '_pyfuse3']
 
 block_cipher = None
 
@@ -25,9 +25,7 @@ a = Analysis([os.path.join(basepath, 'src', 'borg', '__main__.py'), ],
              hiddenimports=hiddenimports,
              hookspath=[],
              runtime_hooks=[],
-             excludes=[
-                '_ssl', 'ssl',
-             ],
+             excludes=[],
              win_no_prefer_redirects=False,
              win_private_assemblies=False,
              cipher=block_cipher)

```

bbx0:
The spec file breaks `pyfuse3`, which would be okay when `llfuse` is used to produce the borg.exe.
 ..but having a borg.exe with only `llfuse` in `PATH` breaks the `fuse3` test of tox, which run against `python -m borg` and `borg.exe`

I see currently theses options to pass the tox tests:
1. Fix the spec file and produce borg.exe with `pyfuse3` and `llfuse` as proposed above.
1. Remove the borg.exe from `PATH` and have tox only run against it's own version of borg. https://github.com/borgbackup/borg/blob/0fa44e4079671bb3a6aa21818661851289cece2a/Vagrantfile#L219 
1. Accept that `fuse3` tests fail against a borg.exe built with `llfuse`, which is reasonable. Maybe there is someway to skip them.
1. Skip the `fuse3` tests and only rely on `llfuse`, which is used for the borg.exe anyway.
1. Add `llfuse` to `BORG_FUSE_IMPL` (-> `fuse3: BORG_FUSE_IMPL=pyfuse3,llfuse`) or just remove the setenv override of `BORG_FUSE_IMPL` for `fuse3`. This should allow the tests to run against the python version with `pyfuse3` and the borg.exe found in `PATH` will fall-back to `llfuse` instead of `None`.

Number 5. is likely most consistent with the supposed FUSE_IMPL detection, when I read `fuse_impl.py` correctly. The spec file fix could be tracked in a diffrent PR, in case the borg binary releases shall switch to `pyfuse3` instead of `llfuse`.

ThomasWaldmann:
The broken fuse3 "binary" (borg.exe) tox tests existed before your PR, so that is not a regression.
Would be good to have a fix, but not in scope of this PR, so just ignore it here.

bbx0:
Alright, thanks! Then the PR is ready to go from my end. (`llfuse` is already used.)

ThomasWaldmann:
OK, thanks for the PR - I'll test this soon within my vagrant setup.

ThomasWaldmann:
Did a first test run, looks as expected (except for the fix #7482 being needed to fix a recently added test).

PR #7488, opened 2023-04-02
Describe performing pull-backups via ssh remote forwarding

centic9:
I tried to use one of the described ways to perform "pull" backups, but all of them either failed or very rather complicated to be used and thus brittle in everyday use.

After some research I came up with simple ssh remote forwarding which for me is much easier to use than any of the currently described approaches.

Let me know if you would describe something differently or if there are security concerns which I did not see (I am just using this in a very small local/home environment, not a big corporate environment).

### DIFF
--- a/docs/deployment/pull-backup.rst
+++ b/docs/deployment/pull-backup.rst
@@ -481,6 +481,17 @@ Security Considerations
 Opening up SSH access this way can pose a security risk as it effectively opens remote access to your
 backup server on the client even if it is located outside of your company network.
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks for the PR! Looks good, but should be combined with a forced command in authorized_keys, so that the client **only** can execute borg serve via ssh and nothing else.

centic9:
> ..., but should be combined with a forced command in authorized_keys, so that the client **only** can execute borg serve via ssh and nothing else.

I added this now.

ThomasWaldmann:
@centic9 As there was no additional feedback, I merged the PR. Thanks!

PR #7512, opened 2023-04-07
prune --list-kept/--list-pruned, fixes #7511

### LINKED ISSUE #7511: prune: only show the archives that are pruned
While discussing the prune output, I'd like the option to only show the archives
          that are pruned, not those that are kept.
          For me, only one or two are pruned each time, but dozens are kept, which clutters the output.

_Originally posted by @jdchristensen in https://github.com/borgbackup/borg/issues/3238#issuecomment-341113742_

ThomasWaldmann:
use this to limit --list output to kept or pruned archives only.

### DIFF
--- a/src/borg/archiver/prune_cmd.py
+++ b/src/borg/archiver/prune_cmd.py
@@ -163,7 +163,11 @@ class PruneMixIn:
                         log_message = "Keeping archive (rule: {rule} #{num}):".format(
                             rule=kept_because[archive.id][0], num=kept_because[archive.id][1]
                         )
-                if args.output_list:
+                if (
+                    args.output_list
+                    or (args.list_pruned and archive in to_delete)
+                    or (args.list_kept and archive not in to_delete)
+                ):
                     list_logger.info(f"{log_message:<40} {formatter.format_item(archive)}")
             pi.finish()
             if sig_int:
@@ -265,6 +269,12 @@ class PruneMixIn:
             "--list", dest="output_list", action="store_true", help="output verbose list of archives it keeps/prunes"
         )
         subparser.add_argument("--short", dest="short", action="store_true", help="use a less wide archive part format")
+        subparser.add_argument(
+            "--list-pruned", dest="list_pruned", action="store_true", help="output verbose list of archives it prunes"
+        )
+        subparser.add_argument(
+            "--list-kept", dest="list_kept", action="store_true", help="output verbose list of archives it keeps"
+        )
         subparser.add_argument(
             "--format",
             metavar="FORMAT",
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
@jdchristensen ^

src/borg/archiver/prune_cmd.py
```diff
@@ -163,7 +163,9 @@ def checkpoint_func():
                         log_message = "Keeping archive (rule: {rule} #{num}):".format(
                             rule=kept_because[archive.id][0], num=kept_because[archive.id][1]
                         )
-                if args.output_list:
+                if args.output_list and (
+                    (args.list_pruned and archive in to_delete) or (args.list_kept and archive not in to_delete)
+                ):
```
jdchristensen, line 168: Would it make sense to change the first `and` to `or`?  Then one would just have to supply `--list-pruned` instead of having to supply both `--list-pruned` and `--list`.
```suggestion
                if args.output_list or (
                    (args.list_pruned and archive in to_delete) or (args.list_kept and archive not in to_delete)
                ):
```

jdchristensen: (The help text would have to change too.)

ThomasWaldmann: I thought the same, but `output_list` also triggers the setup of a special list-logger elsewhere.

jdchristensen: That set-up could be done if any of the three options is set.

In any case, I think the code above is not correct.  If only `--list` is provided, it will print nothing, but the docstrings say that both kept and pruned archives will be listed unless one of the other two options is given.

Currently the docs don't match the code, and I don't think either one specifies the most obvious semantics.  How about `--list` means to show both, `--list-kept` means to show the kept ones, and `--list-pruned` means to show the pruned ones.  If more than one is given, they accumulate, rather than subtracting.

ThomasWaldmann: --list alone works as expected as the default for store_false options is True.
it also works if combined with one of the other options.

but I'll check if i can improve it...

ThomasWaldmann: uh, wow, something broke tests all over the place...

ThomasWaldmann: ok, I special-cased these options in implied logging setup now.

ThomasWaldmann: ok, tests worked. i collapsed the commits into one, guess this is done.

jdchristensen (approved):

src/borg/archiver/prune_cmd.py
jdchristensen: Oh, I think I missed that those were `store_false` options.  In any case, I like the new version.

PR #7517, opened 2023-04-10
Ensure that cli options specified with action=Highlander can only be set once.

### LINKED ISSUE #7500: highlander options "special effects"
borg has some cli options with `action=Highlander`, which raises an error if there is more than one (see also #6269).

E.g. it does not make sense to give `--umask` more than once (especially not with different values).

But, this error is not raised, if the first of these options happens to use the same value as the default value of the option.

E.g.:
```
borg --umask 077 --umask 007 rlist --repo ...  # no error (0o077 is the default)
borg --umask 007 --umask 077 rlist --repo ...  # error "there can be only one"
borg --umask 777 --umask 007 rlist --repo ...  # error "there can be only one"
borg --umask 007 --umask 777 rlist --repo ...  # error "there can be only one"
```

snsmac:
Ensure that cli options specified with action=Highlander can only be set once, even if the set value is a default value. Add tests for action=Highlander. 

Closes #7500

### DIFF
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -249,9 +249,14 @@ def with_archive(method):
 class Highlander(argparse.Action):
     """make sure some option is only given once"""
 
+    def __init__(self, *args, **kwargs):
+        self.__called = False
+        super().__init__(*args, **kwargs)
+
     def __call__(self, parser, namespace, values, option_string=None):
-        if getattr(namespace, self.dest, None) != self.default:
+        if self.__called:
             raise argparse.ArgumentError(self, "There can be only one.")
+        self.__called = True
         setattr(namespace, self.dest, values)
 
 
--- a/src/borg/testsuite/archiver/argparsing.py
+++ b/src/borg/testsuite/archiver/argparsing.py
@@ -11,6 +11,23 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
         self.cmd(f"--repo={self.repository_location}", "delete", "--first", "1", "--last", "1", fork=True, exit_code=2)
 
+    def test_highlander(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "--comment", "comment 1", "test-1", __file__)
+        error_msg = "There can be only one"
+        # Default umask value is 0077
+        # Test that it works with a one time specified default or custom value
+        output_default = self.cmd(f"--repo={self.repository_location}", "--umask", "0077", "rlist")
+        assert error_msg not in output_default
+        output_custom = self.cmd(f"--repo={self.repository_location}", "--umask", "0007", "rlist")
+        assert error_msg not in output_custom
+        # Test that all combinations of custom and default values fail
+        for first, second in [("0007", "0007"), ("0007", "0077"), ("0077", "0007"), ("0077", "0077")]:
+            output_custom = self.cmd(
+                f"--repo={self.repository_location}", "--umask", first, "--umask", second, "rlist", exit_code=2
+            )
+            assert error_msg in output_custom
+
 
 def test_get_args():
     archiver = Archiver()


src/borg/archiver/_common.py
```diff
@@ -249,9 +249,37 @@ def wrapper(self, args, repository, manifest, **kwargs):
...
+        option_strings,
+        dest,
+        nargs=None,
+        const=None,
+        default=None,
+        type=None,
+        choices=None,
+        required=False,
+        help=None,
+        metavar=None,
```
ThomasWaldmann, line 263: is there a need to give these defaults or could one also just use `*args, **kwargs` here?

src/borg/archiver/_common.py
```diff
@@ -249,9 +249,37 @@ def wrapper(self, args, repository, manifest, **kwargs):
...
+            required=required,
+            help=help,
+            metavar=metavar,
+        )
+
     def __call__(self, parser, namespace, values, option_string=None):
-        if getattr(namespace, self.dest, None) != self.default:
+        if self.called:
             raise argparse.ArgumentError(self, "There can be only one.")
+        self.called = True
```
ThomasWaldmann, line 282: ah, good idea! :-)

src/borg/testsuite/archiver/argparsing.py
```diff
@@ -11,6 +12,23 @@ def test_bad_filters(self):
         self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
         self.cmd(f"--repo={self.repository_location}", "delete", "--first", "1", "--last", "1", fork=True, exit_code=2)
 
+    def test_highlander(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "--comment", "comment 1", "test-1", src_dir)
+        error_msg = "There can be only one"
+        # Default permission is 077
```
ThomasWaldmann, line 19: you mean "default value of --umask"?

ThomasWaldmann: also, it is 0077 (according to the docs). the initial 0 means "octal", followed by 3 octal digits for the value.

src/borg/testsuite/archiver/argparsing.py
```diff
@@ -11,6 +12,23 @@ def test_bad_filters(self):
         self.cmd(f"--repo={self.repository_location}", "create", "test", "input")
         self.cmd(f"--repo={self.repository_location}", "delete", "--first", "1", "--last", "1", fork=True, exit_code=2)
 
+    def test_highlander(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(f"--repo={self.repository_location}", "create", "--comment", "comment 1", "test-1", src_dir)
```
ThomasWaldmann, line 17: maybe use something else than "src_dir" - that is a rather expensive thing, archives the complete source directory of borg.

src/borg/testsuite/archiver/argparsing.py
```diff
@@ -11,6 +12,23 @@ def test_bad_filters(self):
...
+        error_msg = "There can be only one"
+        # Default permission is 077
+        # Test that it works with a oen time specified default or custom value
+        output_default = self.cmd(f"--repo={self.repository_location}", "--umask", "077", "rlist")
+        assert error_msg not in output_default
+        output_custom = self.cmd(f"--repo={self.repository_location}", "--umask", "007", "rlist")
+        assert error_msg not in output_custom
+        # Test that all combinations of custom and default values fail
+        for first, second in combinations_with_replacement(["007", "077"], 2):
```
ThomasWaldmann, line 26: guess in such an easy case we rather don't want to use `itertools. combinations_with_replacement`, but just give the 2 tuples directly.

ThomasWaldmann (changes requested):
thanks for the PR! looks good, just some minor nitpicks.

src/borg/testsuite/archiver/argparsing.py
snsmac: I think its only the testsuite dir?

src/borg/archiver/_common.py
```diff
@@ -249,9 +249,14 @@ def wrapper(self, args, repository, manifest, **kwargs):
 class Highlander(argparse.Action):
     """make sure some option is only given once"""
 
+    def __init__(self, *args, **kwargs):
+        self.called = False
+        super(Highlander, self).__init__(*args, **kwargs)
```
ThomasWaldmann, line 254: You can just use `super().__init__(*args, **kwargs)`. The more verbose way was just needed for old pythons.

There is some risk that the super class also uses `.called`, so maybe we rather better use `.__called`?

src/borg/testsuite/archiver/argparsing.py
```diff
@@ -11,6 +13,30 @@ def test_bad_filters(self):
...
+        )
+        error_msg = "There can be only one"
+        # Default umask value is 0077
+        # Test that it works with a one time specified default or custom value
+        output_default = self.cmd(f"--repo={self.repository_location}", "--umask", "0077", "rlist")
+        assert error_msg not in output_default
+        output_custom = self.cmd(f"--repo={self.repository_location}", "--umask", "0007", "rlist")
+        assert error_msg not in output_custom
+        # Test that all combinations of custom and default values fail
+        for first, second in [("0007", "0007"), ("0007", "0077"), ("0077", "0007"), ("0077", "0077")]:
```
ThomasWaldmann, line 34: oh, i had thought that this are just 2 combinations, but sure, if one repeats, one gets 4.

src/borg/testsuite/archiver/argparsing.py
```diff
@@ -11,6 +13,30 @@ def test_bad_filters(self):
...
 
+    def test_highlander(self):
+        self.cmd(f"--repo={self.repository_location}", "rcreate", RK_ENCRYPTION)
+        self.cmd(
+            f"--repo={self.repository_location}",
+            "create",
+            "--comment",
+            "comment 1",
+            "test-1",
+            os.path.dirname(__file__),
```
ThomasWaldmann, line 24: you could just give `__file__`.

ThomasWaldmann (changes requested):

src/borg/testsuite/archiver/argparsing.py
ThomasWaldmann: right. but still too much stuff.

ThomasWaldmann:
@snsmac seen my feedback? I restarted the windows CI, guess that failure wasn't related to your PR.

snsmac:
@ThomasWaldmann Updated

ThomasWaldmann:
src/borg/testsuite/archiver/argparsing.py:2:1: F401 'os.path' imported but unused

snsmac:
Should be (hopefully) all fixed now

ThomasWaldmann (approved):
LGTM! Thanks!

PR #7519, opened 2023-04-11
implement unix domain (ipc) socket support

### LINKED ISSUE #6183: Add UNIX domain socket transport
/kind enhancement

When creating full system backups one must usually run Borg as `root` to avoid permission issues. However, it might be undesirable to store these backups as root, as people often prefer storing backups in known places (like `/var/backups`) owned by dedicated users (like the default `backup` user in Debian-based distributions).

Right now there are only two possible solutions to accomplish this:
- One might `chown` the repo after `borg create` et al. exits
- One might abuse Borg's remote repository feature to connect to `localhost` using SSH as different user (as suggested by @ThomasWaldmann, see references below)

IMO this is a reasonable use case and Borg should support it without taking such rather desperate solutions.

Thus I'd like to suggest adding a `socket://` transport (like `socket:///run/borg/borg.sock`) to Borg. The socket should be created by `borg serve` by accepting an additional `--socket` option with a path (e.g. `borg serve --socket /run/user/1000/borg/borg.sock`, umask 0117). How `borg serve` is being invoked is up to the user. `borg create` et al. can now connect to this socket using the `socket://` transport (e.g. `borg create socket:///run/borg/borg.sock /path/to/backup`).

Even though this is not the goal, it also makes https://github.com/borgbackup/borg/blob/master/docs/deployment/pull-backup.rst#socat easier by eliminating `socat`. It might also enable some more use cases I just can't think of right now.

As I don't know Borg's sources I can't really judge the complexity, but since Borg already supports remote transports with `borg serve`, this could be as easy as replacing `sys.stdin`/`sys.stdout` of `borg serve` with a UNIX domain socket and to let `borg create` et al. communicate with this socket instead of the `ssh` subprocess.

**References**

> For non-ssh repos and running borg as root, but not having the repo owned by root, you can use the ssh://user@localhost/myrepo trick.

\- https://github.com/borgbackup/borg/issues/4082#issuecomment-423679676 

> You can work around that problem by using borg with `user@localhost:/path/to/repo` as repo (in that scenario, borg client can run as root and borg serve would run as user.

\- https://github.com/borgbackup/borg/issues/3587#issuecomment-443694439

ThomasWaldmann:
```
borg serve  --socket=/path/to/repo/socket      # server (listening) side
borg --repo socket:///path/to/repo create ...  # client side
```
fixes #6183.

### DIFF
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -29,7 +29,7 @@ logger = create_logger(__name__)
 
 
 def get_repository(location, *, create, exclusive, lock_wait, lock, append_only, make_parent_dirs, storage_quota, args):
-    if location.proto == "ssh":
+    if location.proto in ("ssh", "socket"):
         repository = RemoteRepository(
             location,
             create=create,
--- a/src/borg/archiver/serve_cmd.py
+++ b/src/borg/archiver/serve_cmd.py
@@ -19,6 +19,7 @@ class ServeMixIn:
             restrict_to_repositories=args.restrict_to_repositories,
             append_only=args.append_only,
             storage_quota=args.storage_quota,
+            socket_path=args.socket_path,
         ).serve()
         return EXIT_SUCCESS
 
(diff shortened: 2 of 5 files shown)


ThomasWaldmann:
TODO / issues:
- server side log output does not go over the socket (socket is bidi, but just 2 channels total. ssh is 3 channels and has separate stderr).
- borg serve terminates after one clientside borg command / connection instead of waiting for next connection

ThomasWaldmann:
see #7615

PR #7544, opened 2023-04-27
do not crash for empty archives list in borg rlist date based matching

### LINKED ISSUE #7522: Borg2: ValueError on date-based matching

Yes

BUG

borg 2.0.0b5

Tested on WSL2 (Debian 11) and Kali 

2 archives adding one new file each (total text 2 files)

```
borg rlist --older 1d --repo XXX
```

I'm getting ValueError on running `borg rlist --older 1d` on a new repo which does not have archives older than a day.

Introduced in #7272 

https://github.com/borgbackup/borg/pull/7272/files#diff-56ce27fae07c0d4ca00050b58f86125f91d88cb4016e38f7c2d5890de9a4e0f7R39-R40

archives_list is empty so min raises ValueError. Perhaps add a check?

```
...
  File "/home/xxx/Projects/borg/env/lib/python3.9/site-packages/borg/manifest.py", line 141, in list
    archives = filter_archives_by_date(archives, oldest=oldest, newest=newest, newer=newer, older=older)
  File "/home/xxx/Projects/borg/env/lib/python3.9/site-packages/borg/manifest.py", line 50, in filter_archives_by_date
    earliest_ts, latest_ts = get_first_and_last_archive_ts(archives)
  File "/home/xxx/Projects/borg/env/lib/python3.9/site-packages/borg/manifest.py", line 41, in get_first_and_last_archive_ts
    return min(timestamps), max(timestamps)
ValueError: min() arg is an empty sequence
```

create a new repo with two archives (both adding one new file). Now run the rlist command with `--older 1d`

jetchirag:
Fixes #7522


src/borg/manifest.py
```diff
@@ -37,6 +37,8 @@ class MandatoryFeatureUnsupported(Error):
 def filter_archives_by_date(archives, older=None, newer=None, oldest=None, newest=None):
     def get_first_and_last_archive_ts(archives_list):
         timestamps = [x.ts for x in archives_list]
+        if not timestamps:
+            return None, None
```
ThomasWaldmann, line 41: how is line 49 working with the None values returned here?

ThomasWaldmann: @jetchirag did you see my feedback?

jetchirag: @ThomasWaldmann I can't remember exactly right now. I'll check it thoroughly and get back. Perhaps I made a mistake.

ThomasWaldmann: I had a look, guess it could be fixed like this:
- add some tests that fail with the current code (see bug report)
- there is one call to `filter_archives_by_date` and before doing that, it checks for `len(archives) > 0` - remove that check.
- in `filter_archives_by_date`, check twice for the archives list being empty:
  - at the very beginning
  - after the archives list has been modified and before calling get_first_and_last_archive_ts
  - if archives list is empty, just `return archives`
- tests should work now

jetchirag: @ThomasWaldmann I just pushed your suggested code. I've appended to the test of original feature PR. Do we need more test?

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -92,6 +92,11 @@ def test_date_matching(self):
         self.assert_in("archive2", output)
         self.assert_not_in("archive3", output)
 
+        # check for output when timespan older than earliest archive is given. Issue #1711
+        output = self.cmd(
+            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--older=9999m", exit_code=0
+        )
+
```
ThomasWaldmann, line 99: ok, that makes sure nothing crashed.

you could also test that it gives the expected result (all archives "not in" the output).

ThomasWaldmann (commented):
thanks for updating this, see the inline comment.

src/borg/manifest.py
ThomasWaldmann: yes, see my other comment. by the main issues are resolved.

src/borg/testsuite/archiver/check_cmd.py
jetchirag: Yes, added.

ThomasWaldmann:
Thanks!

PR #7562, opened 2023-05-13
markup fix + note about max chunk size

elandorr:
as promised

- reST doesn't support alignment in tables so this way is more readable
- technically the use of emphasis here isn't very semantic, but it's readable, and rewriting everything is a bit much for semantics nobody even knows about (the 2 options *should* be subheaders, whatever reST calls it)

I checked v2 and it seems like this part still fits, but you know better.


ThomasWaldmann (changes requested):
Thanks for the PR!

Guess you missed the comment in the first line of these files to **not** edit them (also do not include these in a PR).

https://borgbackup.readthedocs.io/en/stable/development.html#generated-files

(in master branch, it is not `archiver.py`, but `archiver/*_cmd.py`)

elandorr:
I saw the note, but there's a history of you editing the same files, so I thought that was no longer applicable:

https://github.com/borgbackup/borg/commit/3f159ca3bb88f8e2458c655c57b25ee3f376a117
https://github.com/borgbackup/borg/commit/a85b64386626a469415171232ec50851a8e46cd3

I'll check out how to fix it later.

elandorr:
I can only use the gh web UI currently, and it's a nightmare. I'll delete and redo.

PR #7563, opened 2023-05-13
markup fix + note about max chunk size

elandorr:
as promised

- reST doesn't support alignment in tables so this way is more readable
- technically the use of emphasis here isn't very semantic, but it's readable, and rewriting everything is a bit much for semantics nobody even knows about (the 2 options *should* be subheaders, whatever reST calls it)


elandorr:
I can only use the gh web UI currently, and it's a nightmare. I'll delete and redo.

PR #7567, opened 2023-05-14
automated-local.rst: use GPT UUID for consistent udev rule

azrdev:


ThomasWaldmann:
You refer to disks with a GPT partition table. What about MBR partition tables?

azrdev:
@ThomasWaldmann indeed I didn't consider other partition tables because I did not have the hardware to test it at hand (or rather, I only abstracted from my one setup).
I just checked an MBR partitioned USB stick, and while the ID shown by `lsblk -o +PTUUID` is not a UUID but much shorter, I get it again when looking at `udevadm info --path=/sys/block/sdz | grep ID_PART_TABLE_UUID`. I did test using that as documented in an udev rule (with `RUN+="/shell/script/calling/logger"` instead of starting a systemd unit) and it works the same as GPT with a UUID.

So: works with MBR as with GPT (no idea about other partition tables, but I'd guess these work the same)

ThomasWaldmann:
@azrdev so, if the method works for both uusal sorts of partition tables found on USB disks, maybe update the PR to not specifically speak about GPT?

ThomasWaldmann:
@azrdev thanks for the PR! would you like to do a backport to 1.2-maint branch also, so it gets into 1.2.5 release docs?

Issue #5248, opened 2020-07-07
get rid of master/slave terminology


ThomasWaldmann:
master/slave was used a lot in the code that deals with hardlinks:

when we archive multiple hardlinks pointing to same inode (file contents), this happens:

- the first encountered hardlink is treated like any regular file: its file contents are chunked, chunks written to repo, file metadata and chunkid_list is written to the archived item.
- all afterwards discovered hardlinks referring to same inode also create a regular file item, but its `source` attribute points back to the name of the first encountered hardlink. these items do not have a chunkid_list and thus no own content.

so we need new terminology for this now, please suggest.

ThomasWaldmann:
maybe:
- full hardlink vs. reference hardlink
- contentful hardlink vs. contentless/reference hardlinks

don't like:
- primary / secondary (because there might be any amount, not just 2. also, in the filesystem, hardlinks are fully symmetric)

ThomasWaldmann:
related: #5247

strugee:
canonical/alias maybe?

ThomasWaldmann:
@strugee i know that terminology rather from **naming things**. but in our case, hardlink names are all equally "good" in the filesystem (only for symlinks, there is a asymmetry).

So this is rather about how borg stores these items in the archive.

strugee:
Yeah, I see what you're saying. 

>     * contentful hardlink vs. contentless/reference hardlinks

The more I think about it the more I like this pairing. It does a pretty good job communicating the idea right in the name, without needing a ton of further explanation.

ThomasWaldmann:
Not sure why there is a thumbs down on the top post, but maybe it needs some clarification:

Of course renaming some words in an open source project does not change history nor does it improve the situation of slaves that still exist in today's world.

but keeping such negatively loaded terminology without a good reason or opposing to change them would be even worse, so let's just change them.

of course they were never meant in any negative way.

deathtrip:
no need to inject divisive politics into open source software

enkore:
primary/reference, stored/referencing, first/following (<-- this is very accurate, but the terms are possibly too generic to be immediately registered as "jargon"). Contentful/contentless is very accurate and descriptive as well, but imho as far as contrasting pairs go both are pretty similar if you squint your eyes.

> no need to inject divisive politics into open source software

As far as I can tell this is your first comment on a borg repo, perhaps contribute something else before trying to tell people what to do.

enkore:
[...]

PR #7568, opened 2023-05-14
incl./excl. options, path-from-stdin exclusiveness

elandorr:
Should we still bother with 1.2, or is v2 imminent?

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2972,7 +2972,7 @@ class Archiver:
                                 'Paths with fewer elements will be silently skipped.')
 
         def define_exclusion_group(subparser, **kwargs):
-            exclude_group = subparser.add_argument_group('Exclusion options')
+            exclude_group = subparser.add_argument_group('Include/Exclude options')
             define_exclude_and_patterns(exclude_group.add_argument, **kwargs)
             return exclude_group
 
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -3582,8 +3582,8 @@ def define_borg_mount(parser):
                                help='interpret PATH as command and store its stdout. See also section Reading from'
                                     ' stdin below.')
         subparser.add_argument('--paths-from-stdin', action='store_true',
-                               help='read DELIM-separated list of paths to backup from stdin. Will not '
-                                    'recurse into directories.')
+                               help='read DELIM-separated list of paths to backup from stdin. Exclusive option - no recursion'
+                                    ', no excludes, ``-x`` does not apply.')
```
ThomasWaldmann, line 3586: Shall we just say "It will back up all files given, no more, no less."?

elandorr: Sounds good to me. Is the v1 still relevant for more than a few weeks? No point bothering you with 1.2 PRs, if v2 is imminent.

ThomasWaldmann: 1.2.x will be maintained for a longer time, most users have that in production, it is in stable dists and borg2 will still take a while (and even after it is released, people will need some time to adopt/migrate).

ThomasWaldmann: #5248 (and there was no pressure).

ThomasWaldmann:
Thanks!

PR #7589, opened 2023-05-19
relax chunker params validation, tests (1.2-maint)

ThomasWaldmann:


ThomasWaldmann:
forward port: guess we only want to port the new test: `test_buzhash_chunksize_distribution`

PR #7595, opened 2023-05-21
clear empty directories at end of compact process, fixes #6823

### LINKED ISSUE #6823: empty folders in repo_dir/data/ are not automatically removed
Yes

QUESTION

1.2.0

Distribution: Manjaro Linux
Release:	21.3.1
Codename: Ruah

Running a borg Server into a Docker Container with Version 1.2.0

```
All archives:               94.82 GB             71.92 GB             36.81 GB

Chunk index:                  208674               530953
```

borg compact --cleanup-commits 

Having into the Borg Repo Data Folder many empty Folders, overall 11 Folders and 7 Folders are empty

created many Backups over the time and prune the old ones, use Command above and have 7 empty Folders

## Question
Can i delete the empty Folders or do borg neat them in any way?
Can borg delete this empty folders automatically?

nain-F49FF806:
With this commit, empty directories are cleared after segments compacting.
Fixes https://github.com/borgbackup/borg/issues/6823

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -879,6 +879,7 @@ class Repository:
             pi.show()
         pi.finish()
         complete_xfer(intermediate=False)
+        self.io.clear_empty_dirs()
         quota_use_after = self.storage_quota_use
         logger.info('compaction freed about %s repository space.', format_file_size(quota_use_before - quota_use_after))
         logger.debug('compaction completed.')
@@ -1494,6 +1495,22 @@ class LoggedIO:
             safe_unlink(self.segment_filename(segment))
         except FileNotFoundError:
             pass
+
+    def clear_empty_dirs(self):
+        """Delete empty segment dirs, i.e those with no segment files.
+        """
+        data_dir = os.path.join(self.path, 'data')
+        segment_dirs = self.get_segment_dirs(data_dir)
+        for segment_dir in segment_dirs:
+            try:
+                # os.rmdir will only delete the directory if it is empty
+                # so we don't need to explicitly check for emptiness first.
+                os.rmdir(segment_dir)
+            except OSError:
+                # OSError is raised by os.rmdir if directory is not empty. This is expected.
+                # It's subclass FileNotFoundError may be raised if the directory already does not exist. Ignorable.
+                pass
+        sync_dir(data_dir)
 
     def segment_exists(self, segment):
         filename = self.segment_filename(segment)


src/borg/repository.py
```diff
@@ -1475,6 +1476,20 @@ def delete_segment(self, segment):
         except FileNotFoundError:
             pass
 
+    def clear_empty_dirs(self):
+        "Delete empty data dirs, i.e those with no segment files"
+        data_dir = os.path.join(self.path, 'data')
+        segment_dirs = [f.path for f in os.scandir(data_dir) if f.is_dir()]
```
ThomasWaldmann, line 1482: see the `os.listdir` calls that work on the data_dir, they also check the subdir name for isdigit - you should also do that here.

also, instead of creating the full list here, you could just set up a generator.

src/borg/repository.py
```diff
@@ -1475,6 +1476,20 @@ def delete_segment(self, segment):
...
+        for segment_dir in segment_dirs:
+            if len([f for f in os.scandir(segment_dir)]) == 0:
```
ThomasWaldmann, line 1484: simpler: `len(list(os.scandir(segment_dir)))`?

src/borg/repository.py
```diff
@@ -1475,6 +1476,20 @@ def delete_segment(self, segment):
...
+                try:
+                    os.rmdir(segment_dir)
+                    sync_dir(data_dir)
+                except FileNotFoundError:
+                    logger.debug("While trying to clear empty segments dir, it seems it already vanished")
+                except OSError:
+                    logger.debug('While trying to clear empty segments dir, it seems it is not empty after all')
+
```
ThomasWaldmann, line 1492: and even more simple would be to exploit the document behaviour of `os.rmdir`:

https://docs.python.org/3/library/os.html#os.rmdir

- it only removes empty dirs (exactly what we want)
- raises an OSError if directory not found or not empty

we don't need to debug log there, guess this is a valid case to just silently ignore these (expected) error cases.
but, we want a comment in the exception handler about all this.

ThomasWaldmann: nothing is really so important that we need to sync_dir all the time, guess once at the end and we should be good.

src/borg/repository.py
```diff
@@ -1475,6 +1476,20 @@ def delete_segment(self, segment):
         except FileNotFoundError:
             pass
 
+    def clear_empty_dirs(self):
+        "Delete empty data dirs, i.e those with no segment files"
```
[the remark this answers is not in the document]
ThomasWaldmann: also, it needs to say "... empty segment dirs ..." (there is only one data dir and that is the parent of these).

ThomasWaldmann (changes requested):
thanks for the PR, looks pretty good.

i found some minor stuff that could be improved though.

ThomasWaldmann:
related: #7597 (maybe you can address that after this is finished)

src/borg/repository.py
nain-F49FF806: Ignoring the `OSError` as it is expected case. And keeping debug info for `FileNotFoundError` as it is not generally expected. Seems good?

sync_dir moved to end.

src/borg/repository.py
nain-F49FF806: Removing it all together, relying on behaviour of `os.rmdir` as suggested

src/borg/repository.py
```diff
@@ -1475,6 +1476,25 @@ def delete_segment(self, segment):
...
+        segment_dirs = (f.path for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit())
+        for segment_dir in segment_dirs:
+            try:
+                # os.rmdir will only delete the directory if it is empty
+                # so we don't need to explicitly check for emptiness first.
+                os.rmdir(segment_dir)
+            except FileNotFoundError:
+                # This may be indication of another parallel process cleaning up the dir, so log it.
+                logger.debug("While trying to clear empty segments dir, it seems it already vanished. "
+                             "Is another process doing cleanup?.")
```
ThomasWaldmann, line 1492: typo (remove the dot at the end, you already have a question mark).

ThomasWaldmann: also, the repo is locked while borg operates in it, so there can be no parallel process.

so you could just remove 1489..1492 and nothing would be lost. just add a note about "file not found" case in the OSError handler.

ThomasWaldmann (changes requested):

ThomasWaldmann (approved):
LGTM.

src/borg/repository.py
```diff
@@ -1475,6 +1476,22 @@ def delete_segment(self, segment):
         except FileNotFoundError:
             pass
 
+    def clear_empty_dirs(self):
+        """Delete empty segment dirs, i.e those with no segment files.
+        """
+        data_dir = os.path.join(self.path, 'data')
+        segment_dirs = (f.path for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit())
```
ThomasWaldmann, line 1483: guess this should also call the code of #7606.

nain-F49FF806: Yes. Just give me a moment.  Rebasing. :)

nain-F49FF806: Done

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
If you like, you could forward port the 2 PRs to master branch.

nain-F49FF806:
> If you like, you could forward port the 2 PRs to master branch.

Ack. I'll take a look.

PR #7606, opened 2023-05-25
unify scanning and listing of segment dirs, files and apply good practices

### LINKED ISSUE #7597: repository: os.listdir / scandir calls should check for is_dir / is_file and maybe use scandir
in the repository code, some places list all the subdirectories of `data` and then list each of the subdirs to discover the files.

while this works pretty good for the "clean case", it could be a bit more defensive by:
- filtering the segment dirs with `.is_dir()` to make sure we do not try to list a non-directory
- filtering the segment files with `.is_file()` to make sure we do not yield something else than a regular file as segment file

also could use `scandir` instead of `listdir`.

nain-F49FF806:
+ os.scandir instead of os.listdir: Improved speed and added flexibility with attributes (name, path, is_dir(), is_file())
+ use is_dir / is_file to make sure we're reading only dirs / files respectively
+ Filtering to particular start, end index range built in
+ Have separate generators to avoid unnecessary less_than / greater_than checking for every dir/file when not required

Resolves #7597

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1319,41 +1319,37 @@ class LoggedIO:
         safe_fadvise(fd.fileno(), 0, 0, 'DONTNEED')
         fd.close()
 
-    def get_segment_dirs(self, start_index=None, end_index=None):
-        """Returns generator yeilding required segment dirs as `os.DirEntry` objects. Start and end are inclusive.
+    def get_segment_dirs(self, data_dir, start_index=0, end_index=2**30):
+        """Returns generator yielding required segment dirs in data_dir as `os.DirEntry` objects.
+        Start and end are inclusive.
         """
-        data_dir = os.path.join(self.path, 'data')
-        if start_index is None and end_index is None:
-            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit())
-        elif start_index is not None and end_index is None:
-            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name))
-        elif start_index is None and end_index is not None:
-            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and int(f.name) <= end_index)
-        elif start_index is not None and end_index is not None:
-            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name) <= end_index)
+        segment_dirs = (
+            f
+            for f in os.scandir(data_dir)
+            if f.is_dir() and f.name.isdigit() and start_index <= int(f.name) <= end_index
+        )
         return segment_dirs
 
-    def get_segment_files(self, segment_dir, start_index=None, end_index=None):
-        """Returns generator yeilding required segment files in segment_dir as `os.DirEntry` objects. Start and end are inclusive.
+    def get_segment_files(self, segment_dir, start_index=0, end_index=2**32):
+        """Returns generator yielding required segment files in segment_dir as `os.DirEntry` objects.
+        Start and end are inclusive.
         """
-        if start_index is None and end_index is None:
-            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit())
-        elif start_index is not None and end_index is None:
-            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and start_index <= int(f.name))
-        elif start_index is None and end_index is not None:
-            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and int(f.name) <= end_index)
-        elif start_index is not None and end_index is not None:
-            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and start_index <= int(f.name) <= end_index)
+        segment_files = (
+            f
+            for f in os.scandir(segment_dir)
+            if f.is_file() and f.name.isdigit() and start_index <= int(f.name) <= end_index
+        )
         return segment_files
 
     def segment_iterator(self, segment=None, reverse=False):
         if segment is None:
             segment = 0 if not reverse else 2 ** 32 - 1
         start_segment_dir = segment // self.segments_per_dir
+        data_path = os.path.join(self.path, 'data')
         if not reverse:
-            dirs = self.get_segment_dirs(start_index=start_segment_dir)
+            dirs = self.get_segment_dirs(data_path, start_index=start_segment_dir)
         else:
-            dirs = self.get_segment_dirs(end_index=start_segment_dir)
+            dirs = self.get_segment_dirs(data_path, end_index=start_segment_dir)
         dirs = sorted(dirs, key=lambda dir: int(dir.name), reverse=reverse)
         for dir in dirs:
             if not reverse:
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -1319,28 +1319,52 @@ class LoggedIO:
         safe_fadvise(fd.fileno(), 0, 0, 'DONTNEED')
         fd.close()
 
+    def get_segment_dirs(self, start_index=None, end_index=None):
+        """Returns generator yeilding required segment dirs as `os.DirEntry` objects. Start and end are inclusive.
+        """
+        data_dir = os.path.join(self.path, 'data')
+        if start_index is None and end_index is None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit())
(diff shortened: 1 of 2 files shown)


src/borg/repository.py
```diff
@@ -1319,28 +1319,52 @@ def _close_fd(self, ts_fd):
         safe_fadvise(fd.fileno(), 0, 0, 'DONTNEED')
         fd.close()
 
+    def get_segment_dirs(self, start_index=None, end_index=None):
+        """Returns generator yeilding required segment dirs as `os.DirEntry` objects. Start and end are inclusive.
```
ThomasWaldmann, line 1323: typo "yielding"

src/borg/repository.py
```diff
@@ -1319,28 +1319,52 @@ def _close_fd(self, ts_fd):
...
+        """
+        data_dir = os.path.join(self.path, 'data')
+        if start_index is None and end_index is None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit())
+        elif start_index is not None and end_index is None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name))
+        elif start_index is None and end_index is not None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and int(f.name) <= end_index)
+        elif start_index is not None and end_index is not None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name) <= end_index)
```
ThomasWaldmann, line 1333: guess this could be simplified a lot by not using None as the default values for start and end.

you could use start=0 and end=2**30 (the later is a bit arbitrary, but should still fit into a 64bit integer easily and be way more than we'll ever need).

src/borg/repository.py
```diff
@@ -1319,28 +1319,52 @@ def _close_fd(self, ts_fd):
...
+        """Returns generator yeilding required segment files in segment_dir as `os.DirEntry` objects. Start and end are inclusive.
+        """
+        if start_index is None and end_index is None:
+            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit())
+        elif start_index is not None and end_index is None:
+            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and start_index <= int(f.name))
+        elif start_index is None and end_index is not None:
+            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and int(f.name) <= end_index)
+        elif start_index is not None and end_index is not None:
+            segment_files = (f for f in os.scandir(segment_dir) if f.is_file() and f.name.isdigit() and start_index <= int(f.name) <= end_index)
```
ThomasWaldmann, line 1346: same here.

src/borg/repository.py
```diff
@@ -1319,28 +1319,52 @@ def _close_fd(self, ts_fd):
...
+        elif start_index is not None and end_index is None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name))
+        elif start_index is None and end_index is not None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and int(f.name) <= end_index)
+        elif start_index is not None and end_index is not None:
+            segment_dirs = (f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start_index <= int(f.name) <= end_index)
+        return segment_dirs
+
+    def get_segment_files(self, segment_dir, start_index=None, end_index=None):
+        """Returns generator yeilding required segment files in segment_dir as `os.DirEntry` objects. Start and end are inclusive.
```
ThomasWaldmann, line 1337: typo

src/borg/repository.py
```diff
@@ -1319,28 +1319,52 @@ def _close_fd(self, ts_fd):
...
-                filenames = [filename for filename in filenames if filename.isdigit() and int(filename) <= segment]
-            filenames = sorted(filenames, key=int, reverse=reverse)
-            for filename in filenames:
+                files = self.get_segment_files(dir, end_index=segment)
+            files = sorted(files, key=lambda file: int(file.name), reverse=reverse)
+            for file in files:
                 # Note: Do not filter out logically deleted segments  (see "File system interaction" above),
                 # since this is used by cleanup and txn state detection as well.
-                yield int(filename), os.path.join(data_path, dir, filename)
+                yield int(file.name), file.path
```
ThomasWaldmann, line 1367: did you verify that the file.path value starts with data_path?

ThomasWaldmann (changes requested):
Can be quite shorter by a simple "hack".

src/borg/repository.py
nain-F49FF806: oops. Thanks. Fixed.

src/borg/repository.py
nain-F49FF806: Yes. The returned file.path does include the data_path and segment dir. (On Linux). 
I assume the behaviour would be consistent on BSD and Windows.

src/borg/repository.py
nain-F49FF806: Yes. I was actually doing it for performance reasons. 

The comparison `start_index <= int(f.name) <= end_index)` is done on every directory, so for a large number of directories, it could add an *unnecessary* hit when a constrained list is not in fact required (like say for clear_empty_dirs function).

```
# Timing listing from generator with and without index bounds comparision

[tmp]$ python3 -m timeit -s "import os; data_dir = os.path.join('borg-100/', 'data'); start, end = 0, 2**30"  "list((f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit()))" 
10000 loops, best of 5: 22 usec per loop

[tmp]$ python3 -m timeit -s "import os; data_dir = os.path.join('borg-100/', 'data'); start, end = 0, 2**30"  "list((f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start <= int(f.name) <= end))" 
10000 loops, best of 5: 31.9 usec per loop 

# Initial Conclusion: Adding the index bound comparision adds a 50% increase in timing for borg data dir with 100 segment directories.
```

The different cases were added, so we could return the most efficient generator for the specific use case. 

```

# For a 500GiB borg repo we'd have approximately ~ 1000 segment files. 
# So similarly testing for 1000 segment dirs

[tmp]$ python3 -m timeit -s "import os; data_dir = os.path.join('borg-1000/', 'data'); start, end = 0, 2**30"  "list((f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit()))" 
2000 loops, best of 5: 195 usec per loop

[tmp]$ python3 -m timeit -s "import os; data_dir = os.path.join('borg-1000/', 'data'); start, end = 0, 2**30"  "list((f for f in os.scandir(data_dir) if f.is_dir() and f.name.isdigit() and start <= int(f.name) <= end))" 
1000 loops, best of 5: 303 usec per loop

Conclusion: An understandably greater time difference for larger repos.   
But in total only about 0.1 msec per full scan. 
```

In summary, around 0.1 millisecond saving per full scan for larger repos.  So maybe not enough to matter after all?
I'll shorten the code.

ThomasWaldmann: guess time needed for the str to integer conversion and 2 integer comparisons is way less than the usual I/O overhead.

src/borg/repository.py
```diff
@@ -1319,28 +1319,48 @@ def _close_fd(self, ts_fd):
...
+        Start and end are inclusive.
+        """
+        segment_dirs = (
+            f
+            for f in os.scandir(data_dir)
+            if f.is_dir() and f.name.isdigit() and start_index <= int(f.name) <= end_index
+        )
+        return segment_dirs
+
+    def get_segment_files(self, segment_dir, start_index=0, end_index=2**32):
```
ThomasWaldmann, line 1333: let's use `2**32-1`, see line 1346.

also change get_segment_dirs in the same way.

ThomasWaldmann (changes requested):

src/borg/repository.py
ThomasWaldmann: maybe we could have `constants.MIN_SEGMENT = 0` and `.MAX_SEGMENT = 2**32-1` and use them in repository.py everywhere.

nain-F49FF806: Is there a possibility that the starting segment (~0) might change? 
Otherwise, could we just store the upper bounds as constants, since starting at 0 is pretty self-explanatory.

ThomasWaldmann: Well, likely not, but I'ld say if we introduce constants, we do it for both min and max.

nain-F49FF806: Done

ThomasWaldmann (approved):
LGTM!

PR #7608, opened 2023-05-27
unify scanning and listing of segment dirs / segment files and apply good practices

### LINKED ISSUE #7597: repository: os.listdir / scandir calls should check for is_dir / is_file and maybe use scandir

nain-F49FF806:
+ os.scandir instead of os.listdir Improved speed and added flexibility with attributes (name, path, is_dir(), is_file())
+ use is_dir / is_file to make sure  we're reading only dirs / files respectively
+ Filtering to particular start, end index range built in
+ Move value bounds of segment (index) into constants module and use them instead

Resolves #7597

(forward port from commits c9f35a16e9bf9e7073c486553177cef79ff1cb06^..edb5e749f512b7737b6933e13b7e61fefcd17bcb)

### DIFF
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -75,6 +75,12 @@ LIST_SCAN_LIMIT = 100000
 
 FD_MAX_AGE = 4 * 60  # 4 minutes
 
+# Some bounds on segment / segment_dir indexes
+MIN_SEGMENT_INDEX = 0
+MAX_SEGMENT_INDEX = 2**32 - 1
+MIN_SEGMENT_DIR_INDEX = 0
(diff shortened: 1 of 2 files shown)


nain-F49FF806:
related: #7606

Issue #7602, opened 2023-05-23
sh: style patterns: support {alt1,alt2} shell-style alternatives


ThomasWaldmann:
See the TODO comment in `borg.helpers.shellpattern`.

DavidRambo:
Hello, thank you for using the "good first issue" label : )

I'm working on this (along with corresponding tests) and intend to have a pull request soon. My next step is to translate the "stuff" constituting the alternative groups as well as to handle nested groups.

PR #7619, opened 2023-05-31
Add support and tests for shell-style alternatives.

DavidRambo:
This merge would supplement the existing helpers.shellpattern.translate function with support for translating shell-style alternatives into regular expression groups. Tests have been added to confirm what I hope is a robust implementation of this kind of pattern. Nested groups and non-matched braces are handled as would be expected when using brace expansion in the shell. Finally, I have added a remark to the docs where the kinds of shell-style patterns are differentiated from fnmatch.

Of course, I welcome any feedback and would be happy to make changes. For instance, it could use a built-in list in lieu of the LifoQueue to avoid importing the queue module.

Resolves [issue #7602 ](https://github.com/borgbackup/borg/issues/7602)


src/borg/helpers/shellpattern.py
```diff
@@ -18,9 +19,9 @@ def translate(pat, match_end=r"\Z"):
     :copyright: 2001-2016 Python Software Foundation. All rights reserved.
     :license: PSFLv2
 
-    TODO: support {alt1,alt2} shell-style alternatives
-
     """
```
ThomasWaldmann, line 22: nitpick: remove the empty line also.

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
                 elif stuff[0] == "^":
                     stuff = "\\" + stuff
                 res += "[%s]" % stuff
+        elif c == "(" or c == "|" or c == ")":
```
ThomasWaldmann, line 64:
```
elif c in "(!)":
```

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
...
+    # Parse pattern for paired braces.
+    brace_pairs = _parse_braces(pat)
+
+    pat_list = list(pat)  # Convert to list in order to subscript characters.
+
+    # Convert non-escaped commas within groups to pipes.
+    # Passing, e.g. "{a\,b}.txt" to the shell expands to "{a,b}.txt", whereas
+    # "{a\,,b}.txt" expands to "a,.txt" and "b.txt"
+    for key, val in brace_pairs.items():
+        print(f"{key=}, {val=}")
```
ThomasWaldmann, line 124: remove.

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
...
+            if i > 0 and pat[i - 1] != "\\":
+                res += c
         else:
             res += re.escape(c)
 
     return "(?ms)" + res + match_end
+
+
+def _parse_braces(pat):
+    """Returns the index values of paired braces in `pat` as <key, value> pairs, else None.
```
ThomasWaldmann, line 74: `... as <key, value> pairs ...`

That sounds like a list of tuples, but this returns a dict mapping curly brace opening index to corresponding curly brace closing index. But please first see the code feedback below...

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
...
+                # Add to queue
+                curly_q.put(idx)
+        if c == "}" and curly_q.qsize():
+            # If queue is empty, then cannot close pair.
+            if idx > 0 and pat[idx - 1] != "\\":
+                # Closing brace is not escaped.
+                # Pop off the index of the corresponding opening brace, which
+                # provides the key in the dict of pairs, and set its value.
+                pairs[curly_q.get()] = idx
+    return pairs
```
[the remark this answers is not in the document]
ThomasWaldmann: side note: guess having unmatched opening positions in the returned data isn't needed - it does not contain the position of unmatched closing positions either.

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
...
+    """
+    # Parse pattern for paired braces.
+    brace_pairs = _parse_braces(pat)
+
+    pat_list = list(pat)  # Convert to list in order to subscript characters.
+
+    # Convert non-escaped commas within groups to pipes.
+    # Passing, e.g. "{a\,b}.txt" to the shell expands to "{a,b}.txt", whereas
+    # "{a\,,b}.txt" expands to "a,.txt" and "b.txt"
+    for key, val in brace_pairs.items():
```
ThomasWaldmann, line 123: instead of `key, value` guess that `opening, closing` would be easier to read.

ThomasWaldmann: or something like `oidx, cidx`, if you prefer shorter.

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +61,86 @@ def translate(pat, match_end=r"\Z"):
...
+            # If queue is empty, then cannot close pair.
+            if idx > 0 and pat[idx - 1] != "\\":
+                # Closing brace is not escaped.
+                # Pop off the index of the corresponding opening brace, which
+                # provides the key in the dict of pairs, and set its value.
+                pairs[curly_q.get()] = idx
+    return pairs
+
+
+def _translate_groups(pat):
```
ThomasWaldmann, line 110: `_translate_alternatives` or `_translate_alternatives_groups`?

src/borg/testsuite/shellpattern.py
```diff
@@ -66,6 +66,16 @@ def check(path, pattern):
...
+        # Group
+        ("foo1", ["{foo1,foo2}"]),
+        ("foo2", ["foo{1,2}"]),
+        ("foo", ["foo{,1,2}"]),
+        ("foo1", ["{foo{1,2},bar}"]),
+        ("bar", ["{foo{1,2},bar}"]),
+        ("{foo", ["{foo{,bar}"]),
+        ("{foobar", ["{foo{,bar}"]),
+        ("{foo},bar}", ["{foo},bar}"]),
+        ("bar/foobar", ["**/foo{[b]*,[^z]*}"]),
```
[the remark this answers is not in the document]
DavidRambo: My mistake: I was using zsh to come up with patterns, which uses "^" instead of "!" to negate. Incoming squashed commit will fix this.

src/borg/helpers/shellpattern.py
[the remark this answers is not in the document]
DavidRambo: I like this a lot. Initially I was thinking of returning a list of tuples. I'll just change the conditional to `if closing > 0`, since there's no way for a closing brace to be in the zero-index position. I will also change the -1 initiaul value for the closing index to 0 to reflect this.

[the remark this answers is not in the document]
ThomasWaldmann: Well, I liked the -1 better as it is an "impossible index" and does not need any further thinking.

`None` might be an even better value (like "we do not know yet").

DavidRambo (commented):
Thank you so much for your feedback. I have integrated it into a new commit, which I have  squashed with the initial commit introducing the changes. (I apologize if I squashed too soon. I noticed that the contribution guidelines say to commit a fixup that can be squashed "later.")

src/borg/helpers/shellpattern.py
```diff
@@ -60,7 +60,82 @@ def translate(pat, match_end=r"\Z"):
...
+                # Add to queue
+                curly_q.put(idx)
+        if c == "}" and curly_q.qsize():
+            # If queue is empty, then cannot close pair.
+            if idx > 0 and pat[idx - 1] != "\\":
+                # Closing brace is not escaped.
+                # Pop off the index of the corresponding opening brace, which
+                # provides the key in the dict of pairs, and set its value.
+                pairs[curly_q.get()] = idx
+    return [(opening, closing) for opening, closing in pairs.items() if closing]
```
ThomasWaldmann, line 106: Maybe rather use `... if closing is not None` as this is the precise check for "unknown closing position".

ThomasWaldmann (commented):
one cosmetic change and i guess we have it. thanks!

src/borg/helpers/shellpattern.py
ThomasWaldmann: BTW, maybe that sounds nitpicky, but I recently shot myself in the foot with `if repository:` instead of `if repository is not None`. That can behave rather differently than expected if the repository object happens to have a `__len__` method...

DavidRambo: Oh that's right! Does not sound nitpicky to me at all. I've added the change and squashed the last couple commits as fixup commits.

Thank you for shepherding this through. I'm proud to have contributed a little piece to such awesome software : )

docs/usage/help.rst.inc
```diff
@@ -55,7 +55,8 @@ Shell-style patterns, selector ``sh:``
     Like fnmatch patterns these are similar to shell patterns. The difference
     is that the pattern may include ``**/`` for matching zero or more directory
     levels, ``*`` for matching zero or more arbitrary characters with the
-    exception of any path separator. A leading path separator is always removed.
+    exception of any path separator, ``{}`` containing comma-separated alternative
+    patterns. A leading path separator is always removed.
```
ThomasWaldmann, line 59: sorry, we need another change, missed that in my previous reviews:

see the comment in line 1! you need to locate this text in the code, likely in `src/borg/archiver/...` and change it (only) there.

DavidRambo: Ah, very cool. I have moved the text. As I understand the development documentation, I don't need to do anything further, as the maintainers prefer to build the docs on a per-release basis, correct?

PR #7620, opened 2023-05-31
Parametrize compression tests

bigtedde:
The compression tests had a lot of redundant code that was difficult to read. I went through each test using pytest parameterization to reduce the amount of duplicated code and generally clean things up. 

@m3nu @ThomasWaldmann

### DIFF
--- a/src/borg/testsuite/compress.py
+++ b/src/borg/testsuite/compress.py
@@ -2,138 +2,96 @@ import argparse
 import os
 import zlib
 
-try:
-    import lzma
-except ImportError:
-    lzma = None
-
 import pytest
 
 from ..compress import get_compressor, Compressor, CompressionSpec, CNONE, ZLIB, LZ4, LZMA, ZSTD, Auto
 
-
-buffer = bytes(2**16)
-data = b"fooooooooobaaaaaaaar" * 10
+DATA = b"fooooooooobaaaaaaaar" * 10
 params = dict(name="zlib", level=6)
 
 
-def test_get_compressor():
-    c = get_compressor(name="none")
-    assert isinstance(c, CNONE)
-    c = get_compressor(name="lz4")
-    assert isinstance(c, LZ4)
-    c = get_compressor(name="zlib")
-    assert isinstance(c, ZLIB)
-    with pytest.raises(KeyError):
-        get_compressor(name="foobar")
+@pytest.mark.parametrize(
+    "c_type, expected_compressor",
+    [("none", CNONE), ("lz4", LZ4), ("zlib", ZLIB), ("lzma", LZMA), ("zstd", ZSTD), ("foobar", None)],
+)
+def test_get_compressor(c_type, expected_compressor):
+    if expected_compressor is not None:
+        compressor = get_compressor(name=c_type)
+        assert isinstance(compressor, expected_compressor)
+    else:
+        with pytest.raises(KeyError):
+            get_compressor(name=c_type)
 
 
-def test_cnull():
-    c = get_compressor(name="none")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) >= len(data)
-    assert data in cdata  # it's not compressed and just in there 1:1
-    assert data == c.decompress(meta, cdata)[1]
-    assert data == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
-
-
-def test_lz4():
-    c = get_compressor(name="lz4")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) < len(data)
-    assert data == c.decompress(meta, cdata)[1]
-    assert data == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
+@pytest.mark.parametrize("c_type", ["none", "lz4", "zlib", "zstd", "lzma"])
+def test_compression_types(c_type):
+    c = get_compressor(name=c_type)
+    meta, cdata = c.compress({}, DATA)
+    if c_type == "none":
+        assert len(cdata) >= len(DATA)  # it's not compressed and just in there 1:1
+    else:
+        assert len(cdata) < len(DATA)
+    assert DATA == c.decompress(meta, cdata)[1]
+    assert DATA == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
 
 
 def test_lz4_buffer_allocation(monkeypatch):
     # disable fallback to no compression on incompressible data
     monkeypatch.setattr(LZ4, "decide", lambda always_compress: LZ4)
     # test with a rather huge data object to see if buffer allocation / resizing works
-    data = os.urandom(5 * 2**20) * 10  # 50MiB badly compressible data
-    assert len(data) == 50 * 2**20
+    incompressible_data = os.urandom(5 * 2**20) * 10  # 50MiB badly compressible data
     c = Compressor("lz4")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) >= len(data)
-    assert data == c.decompress(meta, cdata)[1]
+    meta, cdata = c.compress({}, incompressible_data)
+    assert len(incompressible_data) == 50 * 2**20
+    assert len(cdata) >= len(incompressible_data)
+    assert incompressible_data == c.decompress(meta, cdata)[1]
 
 
-def test_zlib():
-    c = get_compressor(name="zlib")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) < len(data)
-    assert data == c.decompress(meta, cdata)[1]
-    assert data == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
-
-
-def test_lzma():
-    if lzma is None:
-        pytest.skip("No lzma support found.")
-    c = get_compressor(name="lzma")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) < len(data)
-    assert data == c.decompress(meta, cdata)[1]
-    assert data == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
-
-
-def test_zstd():
-    c = get_compressor(name="zstd")
-    meta, cdata = c.compress({}, data)
-    assert len(cdata) < len(data)
-    assert data == c.decompress(meta, cdata)[1]
-    assert data == Compressor(**params).decompress(meta, cdata)[1]  # autodetect
-
-
-def test_autodetect_invalid():
+@pytest.mark.parametrize("invalid_cdata", [b"\xff\xfftotalcrap", b"\x08\x00notreallyzlib"])
+def test_autodetect_invalid(invalid_cdata):
     with pytest.raises(ValueError):
-        Compressor(**params, legacy_mode=True).decompress({}, b"\xff\xfftotalcrap")
-    with pytest.raises(ValueError):
-        Compressor(**params, legacy_mode=True).decompress({}, b"\x08\x00notreallyzlib")
+        Compressor(**params, legacy_mode=True).decompress({}, invalid_cdata)
 
 
 def test_zlib_legacy_compat():
     # for compatibility reasons, we do not add an extra header for zlib,
-    # nor do we expect one when decompressing / autodetecting
+    # nor do we expect one when decompressing / auto-detecting
     for level in range(10):
         c = get_compressor(name="zlib_legacy", level=level, legacy_mode=True)
-        meta1, cdata1 = c.compress({}, data)
-        cdata2 = zlib.compress(data, level)
+        meta1, cdata1 = c.compress({}, DATA)
+        cdata2 = zlib.compress(DATA, level)
         assert cdata1 == cdata2
         meta2, data2 = c.decompress({}, cdata2)
-        assert data == data2
-        # _, data2 = Compressor(**params).decompress({}, cdata2)
-        # assert data == data2
+        assert DATA == data2
 
 
-def test_compressor():
-    params_list = [
+@pytest.mark.parametrize(
+    "c_params",
+    [
         dict(name="none"),
         dict(name="lz4"),
         dict(name="zstd", level=1),
-        dict(name="zstd", level=3),
-        # avoiding high zstd levels, memory needs unclear
+        dict(name="zstd", level=3),  # avoiding high zstd levels, memory needs unclear
         dict(name="zlib", level=0),
         dict(name="zlib", level=6),
         dict(name="zlib", level=9),
-    ]
-    if lzma:
-        params_list += [
-            dict(name="lzma", level=0),
-            dict(name="lzma", level=6),
-            # we do not test lzma on level 9 because of the huge memory needs
-        ]
-    for params in params_list:
-        c = Compressor(**params)
-        meta_c, data_compressed = c.compress({}, data)
-        assert "ctype" in meta_c
-        assert "clevel" in meta_c
-        assert meta_c["csize"] == len(data_compressed)
-        assert meta_c["size"] == len(data)
-        meta_d, data_decompressed = c.decompress(meta_c, data_compressed)
-        assert data == data_decompressed
-        assert "ctype" in meta_d
-        assert "clevel" in meta_d
-        assert meta_d["csize"] == len(data_compressed)
-        assert meta_d["size"] == len(data)
+        dict(name="lzma", level=0),
+        dict(name="lzma", level=6),  # we do not test lzma on level 9 because of the huge memory needs
+    ],
+)
+def test_compressor(c_params):
+    c = Compressor(**c_params)
+    meta_c, data_compressed = c.compress({}, DATA)
+    assert "ctype" in meta_c
+    assert "clevel" in meta_c
+    assert meta_c["csize"] == len(data_compressed)
+    assert meta_c["size"] == len(DATA)
+    meta_d, data_decompressed = c.decompress(meta_c, data_compressed)
+    assert DATA == data_decompressed
+    assert "ctype" in meta_d
+    assert "clevel" in meta_d
+    assert meta_d["csize"] == len(data_compressed)
+    assert meta_d["size"] == len(DATA)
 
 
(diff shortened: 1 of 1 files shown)


.DS_Store
m3nu, line 1: Please remove those macOS temp files and exclude them in your global Git config.

src/borg/testsuite/compress.py
```diff
@@ -11,36 +11,39 @@
...
+        compressor = get_compressor(name=c_type)
+        assert isinstance(compressor, expected_compressor)
+    else:
+        with pytest.raises(KeyError):
+            get_compressor(name=c_type)
+
+
+@pytest.mark.parametrize(
+    "c_type",
+    ['none', 'lz4', 'zlib', 'zstd'] + (['lzma'] if lzma else [])
```
ThomasWaldmann, line 38: just assume lzma is always there (and also simplify the import).

guess many years ago i wanted the tests to be able to work even with half-broken self-compiled pythons that miss lzma support, but considering lzma is a documented part of the python standard library, i guess we can get rid of that.

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
-
-
-def test_autodetect_invalid():
-    with pytest.raises(ValueError):
-        Compressor(**params, legacy_mode=True).decompress({}, b"\xff\xfftotalcrap")
+@pytest.mark.parametrize("byte_string", [b"\xff\xfftotalcrap", b"\x08\x00notreallyzlib"],)
+def test_autodetect_invalid(byte_string):
     with pytest.raises(ValueError):
-        Compressor(**params, legacy_mode=True).decompress({}, b"\x08\x00notreallyzlib")
+        Compressor(**params, legacy_mode=True).decompress({}, byte_string)
```
ThomasWaldmann, line 66: maybe use `invalid_cdata` rather than `byte_string`?

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
 
 
 def test_zlib_legacy_compat():
     # for compatibility reasons, we do not add an extra header for zlib,
-    # nor do we expect one when decompressing / autodetecting
+    # nor do we expect one when decompressing / auto-detecting
```
ThomasWaldmann, line 71: good catch!

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
-        meta_d, data_decompressed = c.decompress(meta_c, data_compressed)
-        assert data == data_decompressed
-        assert "ctype" in meta_d
-        assert "clevel" in meta_d
-        assert meta_d["csize"] == len(data_compressed)
-        assert meta_d["size"] == len(data)
+    ] + ([
+        dict(name="lzma", level=0),
+        dict(name="lzma", level=6)
+    ] if lzma else []),  # we do not test lzma on level 9 because of the huge memory needs
```
ThomasWaldmann, line 94: also here: just assume lzma is always there.

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
     compressor_lz4 = CompressionSpec("lz4").compressor
     compressor_zlib = CompressionSpec("zlib,9").compressor
-    data = bytes(500)
-    meta, compressed_auto_zlib = compressor_auto_zlib.compress({}, data)
-    _, compressed_lz4 = compressor_lz4.compress({}, data)
-    _, compressed_zlib = compressor_zlib.compress({}, data)
+    data_bytes = bytes(500)
+    meta, compressed_auto_zlib = compressor_auto_zlib.compress({}, data_bytes)
+    _, compressed_lz4 = compressor_lz4.compress({}, data_bytes)
+    _, compressed_zlib = compressor_zlib.compress({}, data_bytes)
```
ThomasWaldmann, line 118: i found `data` and `cdata` an nice and short way to refer to uncompressed and compressed data.

don't see much advantage in using `data_bytes`.

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
-    data = bytes(1100)
-    _, compressed = compressor.compress({}, data)
-    assert 1100 <= len(compressed) <= 1100 + 1024
+    data_bytes = bytes(10000)
+    _, compressed = compressor.compress({}, data_bytes)
+
+    # tests obfuscation with compression
+    if not isinstance(cs.inner.compressor, CNONE):
+        min_compress, max_compress = 0.2, 0.001  # estimate compression factor outer boundaries
+        assert max_compress * len(data_bytes) <= len(compressed) <= min_compress * len(data_bytes) * 1001
```
ThomasWaldmann, line 151: that factor 1001 comes from obfuscation level 2, guess it should be parametrized also.

same for 101 (1) and 10000001 (6).

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
+        assert len({len(compressor.compress({}, data_bytes)[1]) for i in range(100)}) > result_range
+        # compressing 100 times the same data should give multiple different result sizes
+
+    # tests obfuscation without compression
+    elif specs == "obfuscate,1,none":
+        assert len(data_bytes) <= len(compressed) <= len(data_bytes) * 1001
```
ThomasWaldmann, line 157: 101?

src/borg/testsuite/compress.py
```diff
@@ -49,168 +52,122 @@ def test_lz4_buffer_allocation(monkeypatch):
...
+        assert len({len(compressor.compress({}, data_bytes)[1]) for i in range(100)}) > result_range
+    elif specs == "obfuscate,110,none":
+        # should always be in a 1024 byte range, regardless of size.
+        assert 10000 <= len(compressed) <= 10000 + 1024
```
ThomasWaldmann, line 161: adding 1024 comes from obfuscation level 110: `2 ** (110 - 100) == 1024`, should be also parametrized.

src/borg/testsuite/compress.py
```diff
@@ -221,51 +178,40 @@ def test_obfuscate_meta():
...
+def test_compression_specs(c_type, c_name, c_levels):
+    for level in c_levels:
+        # a compression level is specified
+        if isinstance(level, int):
+            cs = CompressionSpec(f"{c_name},{level}").compressor
+            assert isinstance(cs, c_type)
+            assert cs.level == level
+
+        # a compression level is not specified (default)
+        elif not level:
```
ThomasWaldmann, line 205: `elif level == "":`

src/borg/testsuite/compress.py
```diff
@@ -221,51 +178,40 @@ def test_obfuscate_meta():
...
-    assert isinstance(zstd, ZSTD)
-    assert zstd.level == 1
-    zstd = CompressionSpec("zstd,22").compressor
-    assert isinstance(zstd, ZSTD)
-    assert zstd.level == 22
-
-    with pytest.raises(argparse.ArgumentTypeError):
-        CompressionSpec("lzma,9,invalid")
-    with pytest.raises(argparse.ArgumentTypeError):
-        CompressionSpec("invalid")
```
ThomasWaldmann, line 271: guess `CompressionSpec("invalid")` is missing now?

ThomasWaldmann:
guess you want to set up `black` and `pre-commit` tools, see our docs.

src/borg/testsuite/compress.py
bigtedde: I agree with your logic, and only changed to prevent confusion with `data` variable from outer scope. I can revert this change :)

src/borg/testsuite/compress.py
bigtedde: Ah. Thank you for explaining this. I will parameterize 👍

src/borg/testsuite/compress.py
bigtedde: Attempted to replicate this with the `(None, None, ["invalid"]),` parameter on line 193

ThomasWaldmann: Doesn't that test with `"None,invalid"` rather than `"invalid"`.

src/borg/testsuite/compress.py
ThomasWaldmann: yes, please!

if there is a problem with that shadowing, just rename the global to e.g. DATA.

src/borg/testsuite/compress.py
```diff
@@ -157,60 +115,47 @@ def test_auto():
...
+        # compressing 100 times the same data should give multiple different result sizes
+
+    # tests obfuscation without compression
+    elif specs == "obfuscate,1,none":
+        assert len(data) <= len(compressed) <= len(data) * obfuscate_factor
+        assert len({len(compressor.compress({}, data)[1]) for i in range(100)}) > result_range
+    elif specs == "obfuscate,110,none":
+        data = bytes(1000)
+        _, compressed = compressor.compress({}, data)
+        assert 1000 <= len(compressed) <= 1000 + obfuscate_factor
```
ThomasWaldmann, line 149: it's a bit ugly that the additive obfuscation is now tested with a "factor" (which implies multiplication).

rename it?
split into separate tests for additive and multiplicative obfuscation? maybe also separate the none compression and get rid of the if/elif?

src/borg/testsuite/compress.py
```diff
@@ -221,51 +166,41 @@ def test_obfuscate_meta():
...
+            if c_type in (ZLIB, LZMA):
+                assert cs.level == 6
+            elif c_type == ZSTD:
+                assert cs.level == 3
+
+        # a compression level is invalid
+        elif level.endswith("invalid"):
+            invalid_spec = f"{c_name},{level}" if c_name else level
+            with pytest.raises(argparse.ArgumentTypeError):
+                CompressionSpec(invalid_spec)
```
ThomasWaldmann, line 206: hmm, split this into 3 separate tests: for defaults, for given levels, for invalid levels and get rid of the if/elif?

ThomasWaldmann (commented):
maybe check if splitting the tests makes their code easier.

src/borg/testsuite/compress.py
ThomasWaldmann: with separate tests, some of the information in the comments could be also transported via a good test name.

src/borg/testsuite/compress.py
bigtedde: This makes a lot of sense, I will separate these and tests.

However, I am still unclear on why obfuscation level 1 gets an obfuscation 'factor' of 101, while obfuscation level 110 gets an obfuscation 'addition' of 1024?

src/borg/testsuite/compress.py
bigtedde: Absolutely 👍

src/borg/testsuite/compress.py
ThomasWaldmann: it is just defined like that (abusing one levels range for 2 different obfuscators):

https://borgbackup.readthedocs.io/en/2.0.0b5/usage/help.html?highlight=obfuscate#borg-help-compression (you can also read the code)

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
+
+@pytest.mark.parametrize(
+    "specs, c_type, obfuscation_padding",
+    [
+        ("obfuscate,110,none", CNONE, 1024),  # up to 1KiB padding
+        ("obfuscate,120,lz4", LZ4, 1048576),  # up to 1MiB padding
+        ("obfuscate,123,zstd,3", ZSTD, 8388608),  # max, up to 8MiB padding
+    ],
+)
+def test_additive_obfuscation(specs, c_type, obfuscation_padding):
```
bigtedde, line 152: Created additional test cases for additive obfuscation. Before `obfuscate,110,none` was the only example, so I included a couple more instances.

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
+    assert len({len(compressor.compress({}, data)[1]) for i in range(100)}) > result_range
+    # compressing 100 times the same data should give multiple different result sizes
+
+
+@pytest.mark.parametrize(
+    "specs, c_type, obfuscation_padding",
+    [
+        ("obfuscate,110,none", CNONE, 1024),  # up to 1KiB padding
+        ("obfuscate,120,lz4", LZ4, 1048576),  # up to 1MiB padding
+        ("obfuscate,123,zstd,3", ZSTD, 8388608),  # max, up to 8MiB padding
```
ThomasWaldmann, line 149: maybe use `2**10`, `2**20`, `2**23` to make the relation to the obfuscation level more obvious?

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
+    # Testing randomly sized padding, obfuscation spec 110 to 123 inclusive
+    # obfuscate_padding = 2 ** (obfuscation spec - 100)
+    cs = CompressionSpec(specs)
+    assert isinstance(cs.inner.compressor, c_type)
+    compressor = cs.compressor
+    data_list = (bytes(1000), bytes(1100))
+    for data in data_list:
         _, compressed = compressor.compress({}, data)
-        assert 1100 <= len(compressed) <= 1100 + obfuscate_factor
+        if c_type == CNONE:  # no compression
```
ThomasWaldmann, line 161: btw, if you test for object identity (here: class identity), using `is` is more appropriate than `==`, which rather tests for value equality.

src/borg/testsuite/compress.py
ThomasWaldmann: good! :)

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
     assert isinstance(cs.inner.compressor, c_type)
     compressor = cs.compressor
     data = bytes(10000)
     _, compressed = compressor.compress({}, data)
-
-    # tests obfuscation with compression
-    if not isinstance(cs.inner.compressor, CNONE):
+    if c_type == CNONE:  # no compression
+        assert len(data) <= len(compressed) <= len(data) * obfuscation_factor
+    else:  # with compression
```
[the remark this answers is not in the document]
ThomasWaldmann: guess it is ok like it is.

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
+    # obfuscate_factor = 10**(obfuscation spec) +1
     cs = CompressionSpec(specs)
     assert isinstance(cs.inner.compressor, c_type)
     compressor = cs.compressor
     data = bytes(10000)
     _, compressed = compressor.compress({}, data)
-
-    # tests obfuscation with compression
-    if not isinstance(cs.inner.compressor, CNONE):
+    if c_type == CNONE:  # no compression
```
ThomasWaldmann, line 135: see below.

src/borg/testsuite/compress.py
```diff
@@ -116,40 +116,53 @@ def test_auto():
...
 
 @pytest.mark.parametrize(
-    "specs, c_type, result_range, obfuscate_factor",
+    "specs, c_type, result_range, obfuscation_factor",
     [
         ("obfuscate,1,none", CNONE, 50, 101),
-        ("obfuscate,110,none", CNONE, None, 1024),
         ("obfuscate,2,lz4", LZ4, 10, 1001),
         ("obfuscate,6,zstd,3", ZSTD, 90, 10000001),
         ("obfuscate,2,auto,zstd,10", Auto, 10, 1001),
```
ThomasWaldmann: to make the relation more obvious, you could use `10**1`, `10**2`, `10**6`, `10**2` here and use `10*x+1` below.

ThomasWaldmann (changes requested):
much better / simpler!

some minor things, but guess then it is ready to squash/merge.

src/borg/testsuite/compress.py
```diff
@@ -157,60 +115,60 @@ def test_auto():
...
+@pytest.mark.parametrize(
+    "specs, c_type, result_range, obfuscation_factor",
+    [
+        ("obfuscate,1,none", CNONE, 50, 10**1),
+        ("obfuscate,2,lz4", LZ4, 10, 10**2),
+        ("obfuscate,6,zstd,3", ZSTD, 90, 10**6),
+        ("obfuscate,2,auto,zstd,10", Auto, 10, 10**2),
+    ],
+)
+def test_factor_obfuscation(specs, c_type, result_range, obfuscation_factor: int):
```
bigtedde, line 127: Included a type annotation for obfuscation_factor, as it was showing a warning that these were evaluating to tuples without it.

src/borg/testsuite/compress.py
```diff
@@ -157,60 +115,60 @@ def test_auto():
...
-    assert isinstance(cs.inner.compressor, CNONE)
+@pytest.mark.parametrize(
+    "specs, c_type, obfuscation_padding",
+    [
+        ("obfuscate,110,none", CNONE, 2**10),  # up to 1KiB padding
+        ("obfuscate,120,lz4", LZ4, 2**20),  # up to 1MiB padding
+        ("obfuscate,123,zstd,3", ZSTD, 2**23),  # max, up to 8MiB padding
+    ],
+)
+def test_additive_obfuscation(specs, c_type, obfuscation_padding: int):
```
bigtedde, line 152: Same thing here

ThomasWaldmann (approved):
LGTM. Squash/merge?

bigtedde:
> LGTM. Squash/merge?

Squashed and ready for merge. Thank you!

ThomasWaldmann:
Thanks!

PR #7623, opened 2023-06-01
CI: merge Windows workflow with main

RayyanAnsari:
This also enables us to avoid running it when lint fails.

image


ThomasWaldmann:
Thanks!

PR #7624, opened 2023-06-01
CI: try 4 parallel pytest-xdist workers

ThomasWaldmann:


RayyanAnsari:
the runners have only 2 cores available.
would it be better to just run 2 in parallel?

ThomasWaldmann:
Run times of the tests (pytest/tox):

env - before (no parallelism) - after (4 parallel workers):
- linux py39 - 15:11 - 9:01
- linux py310 - 20:32 - 11:11
- linux py311 - 16:54 - 8:40
- macos py311 - 22:48 - 11:29
- windows py310 - 12:22 - 4:38

It's running on 2 or 3 cores and the stuff is usually not cpu bound, but does a lot of I/O - so it needs more workers than cores and will still not get anywhere near 100% cpu load on all cores.

PR #7635, opened 2023-06-09
Improve patterns help

eoli3n:
I had hard time to understand how patterns are working, tried to improve this, making explanations more clear, and adding a useful sample for declarative includes

### DIFF
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -178,6 +178,25 @@ class HelpMixIn:
         exclude stops examination of subdirectories so that potential includes
         will not match - use normal excludes for such use cases.
 
+        Example::
+
+            # Define the recursion root
+            R /
+            # Exclude all iso files in any directory
+            - **/*.iso
+            # Explicitly include all inside etc and root
+            + etc/**
+            + root/**
+            # Exclude a specific directory under each user's home directories
+            - home/*/.cache
+            # Explicitly include everything in /home
+            + home/**
+            # Explicitly exclude some directories without recursing into them
+            ! re:^(dev|proc|run|sys|tmp)
+            # Exclude all other files and directories
+            # that are not specifically included earlier.
+            - **
+
         **Tip: You can easily test your patterns with --dry-run and  --list**::
 
             $ borg create --dry-run --list --patterns-from patterns.txt archive
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -32,11 +32,9 @@ class HelpMixIn:
         If you give ``../../relative`` as root, the paths will be normalized
         as ``relative/``.
 
-        File patterns support five different styles. If followed by a colon ':',
-        the first two characters of a pattern are used as a style selector.
-        Explicit style selection is necessary if a non-default style is desired
-        or when the desired pattern starts with two alphanumeric characters
-        followed by a colon (i.e. ``aa:something/*``).
+        Borg supports different pattern styles. To define a non-default
+        style for a specific pattern, prefix it with two characters followed
+        by a colon ':' (i.e. ``fm:path/*``, ``sh:path/**``).
 
         `Fnmatch <https://docs.python.org/3/library/fnmatch.html>`_, selector ``fm:``
             This is the default style for ``--exclude`` and ``--exclude-from``.
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -28,9 +28,9 @@ class HelpMixIn:
 
         Starting with Borg 1.2, paths that are matched against patterns always
         appear relative. If you give ``/absolute/`` as root, the paths going
-        into the matcher will look relative like ``absolute/.../file.ext``.
-        If you give ``../some/path`` as root, the paths will look like
-        ``some/path/.../file.ext``.
+        into the matcher will start with ``absolute/``.
+        If you give ``../../relative`` as root, the paths will be normalized
+        as ``relative/``.
 
         File patterns support five different styles. If followed by a colon ':',
         the first two characters of a pattern are used as a style selector.


src/borg/archiver/help_cmd.py
```diff
@@ -28,15 +28,13 @@ class HelpMixIn:
 
         Starting with Borg 1.2, paths that are matched against patterns always
         appear relative. If you give ``/absolute/`` as root, the paths going
-        into the matcher will look relative like ``absolute/.../file.ext``.
-        If you give ``../some/path`` as root, the paths will look like
-        ``some/path/.../file.ext``.
+        into the matcher will look relative like ``absolute/``.
+        If you give ``../../relative`` as root, the paths will be normalized
+        as ``relative/``.
```
ThomasWaldmann, line 33: I think the `file.ext` should be kept, so readers will realize this is an example full path going into the matcher.

But as we discussed, if the `...` is confusing, get rid of that.

src/borg/archiver/help_cmd.py
```diff
@@ -28,15 +28,13 @@ class HelpMixIn:
...
 
-        File patterns support five different styles. If followed by a colon ':',
-        the first two characters of a pattern are used as a style selector.
-        Explicit style selection is necessary if a non-default style is desired
-        or when the desired pattern starts with two alphanumeric characters
-        followed by a colon (i.e. ``aa:something/*``).
+        File patterns support five different styles. To define a non-default
```
ThomasWaldmann, line 35: Borg supports different pattern styles. ...

src/borg/archiver/help_cmd.py
```diff
@@ -28,15 +28,13 @@ class HelpMixIn:
...
+        style for a specific pattern, prefix it with two characters followed
+        by a column ';' (i.e. ``fn:path/*``, ``sh:path/**``).
```
ThomasWaldmann, line 37: 3 typos: ... colon ':' ... ``fm:...``

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            # Exclude specific directory under /home
+            - home/*/.cache
+            # Explicilty include all other directories in /home
+            + home/**
+            # Explicitly exclude some directories without recursion
+            ! re:^(dev|proc|run|sys|tmp)
+            # Exclude all other files and directories
+            # that are not specifically included earlier
+            - **
+
```
ThomasWaldmann, line 199: If you add an example, please make sure it does what the comments say.
Also, read it again, typos.

ThomasWaldmann (changes requested):
Thanks for the PR, some stuff I found...

src/borg/archiver/help_cmd.py
eoli3n: As you prefer, I find it more clear like that. Is "realizing this is a full path" important to understand the idea of the leading slash here ? If you confirm, then I'll readd the file.ext without dots.

src/borg/archiver/help_cmd.py
eoli3n: That's the config I use here, and it seems to work. Do you notice something which could not work as described ?

eoli3n: Typo done

src/borg/archiver/help_cmd.py
eoli3n: done

src/borg/archiver/help_cmd.py
ThomasWaldmann: the point is that these are example paths coming from fs recursion, not something from cli args or patterns file.

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
         exclude stops examination of subdirectories so that potential includes
         will not match - use normal excludes for such use cases.
 
+        Example::
+
+            # Define the root path prefix
```
ThomasWaldmann, line 183: maybe rather say **recursion root* (not prefix).

just to make clear that this is not precisely the prefix seen in paths going into the matcher (removal of left slash or other stuff).

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            R /
+            # Exclude all iso files in any directory
+            - **/*.iso
+            # Explicitly include some directories and all their content
+            + etc/**
+            + root/**
```
ThomasWaldmann, line 189: maybe this does **not** back up the `etc` directory item?

ThomasWaldmann: also, this is a bit unusual.

many users would consider these 2 dirs as important and backup worthy rather than wanting to exclude them.

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            # Exclude a specific directory under /home
+            - home/*/.cache
```
ThomasWaldmann, line 191: ... under each user's home directory

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            # Explicitly include all other directories in /home
+            + home/**
```
ThomasWaldmann, line 193: ... everything in /home

note: not just the directories

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            # Explicitly exclude some directories without recursion
```
ThomasWaldmann, line 194: ... without recursing into them.

src/borg/archiver/help_cmd.py
```diff
@@ -180,6 +178,25 @@ class HelpMixIn:
...
+            ! re:^(dev|proc|run|sys|tmp)
+            # Exclude all other files and directories
+            # that are not specifically included earlier
```
ThomasWaldmann, line 197: add a dot at the end of the sentence.

ThomasWaldmann (changes requested):
some more comments

src/borg/archiver/help_cmd.py
eoli3n:
```
root@osz ~# borgmatic mount --repository vps --archive osz-2023-06-08T20:29:47 --mount-point /mnt/m
root@osz ~# cd /mnt/m
root@osz /m/m# ls
etc/  home/  root/
```

> backup worthy rather than wanting to exclude them

I don't understand why you think that it could not backup the /etc dir ?

? I *include* them, not exclude, so I don't get your second comment.

src/borg/archiver/help_cmd.py
eoli3n: discussed in the latest fixup

src/borg/archiver/help_cmd.py
ThomasWaldmann: ehrm, ignore my second comment, guess i was distracted or it was too late already.

maybe "include all inside etc and root" or so would be a good comment.
the 2 directory fs items might be missing, but as they are already present after a fresh install, this is no real issue.

src/borg/archiver/help_cmd.py
ThomasWaldmann: Alternatively, one could say "the paths going into the match will start with ...".

Not "will look like" without the "..." or the example path/filenanem.

ThomasWaldmann (approved):
LGTM.

ThomasWaldmann:
Thanks! Would you like to backport this to 1.2-maint branch also?

PR #7642, opened 2023-06-10
replace `LRUCache` internals with `OrderedDict`

Deric-W:
Hello,
while looking trough the code I noticed that `LRUCache` is using custom logic to keep track of ordering which handles the `list` operations not very efficiently.
Since `OrderedDict` is implemented in C it should provide a speedup and allow for a simpler implementation.

I tried to run the tests on my machine but `conftest.py` fails to parse the version made up by setuptools-scm.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1236,7 +1236,7 @@ class MetadataCollector:
 # (hash_func, chunk_length) -> chunk_hash
 # we play safe and have the hash_func in the mapping key, in case we
 # have different hash_funcs within the same borg run.
-zero_chunk_ids = LRUCache(10, dispose=lambda _: None)
+zero_chunk_ids = LRUCache(10)  # type: ignore[var-annotated]
 
 
 def cached_hash(chunk, id_hash):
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
@Deric-W thanks for the PR. That issue with setuptools-scm is likely because you made a shallow copy of the git repo or a copy without the tags. Just pull the full history / the tags.

ThomasWaldmann:
@Deric-W btw, did you check if you can use dict instead of OrderedDict? Since python 3.7, dicts remember order.

ThomasWaldmann:
@Deric-W `black` is unhappy still, you should install `black` and run it on the source (see developer docs).

Deric-W:
While the [documentation](https://docs.python.org/3/library/collections.html#ordereddict-objects) shows how to use a normal dict for this purpose it mentions that `OrderedDict` is more efficient in this case which a can confirm after testing the snippets with timeit.

ThomasWaldmann (changes requested):
Code looks good, thanks!

The docstring is a bit confusing and its usage of "replacing" could be misunderstood to refer to the "replace" method.

ThomasWaldmann (approved):
LGTM!

PR #7644, opened 2023-06-11
Backport of improvements in patterns help

eoli3n:
See https://github.com/borgbackup/borg/pull/7635

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -2322,16 +2322,9 @@ class Archiver:
         included but not its content. If it does not end with a slash, such as
         `some/path`, both the directory and content will be excluded.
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann (commented):
It seems like this is changing more than #7635?

eoli3n:
Yes because, in the main branch, "pattern prefixes" was already there, but not on that branch.
I then added them see https://github.com/borgbackup/borg/pull/7644/commits/099f23032e86608b235e6b363bc66c7962d4f3d5 and https://github.com/eoli3n/borg/blob/095d5f463bfd5ad9f0978edf6d87d6d49f49b0dc/src/borg/archiver/help_cmd.py#L146-L174

PR #7650, opened 2023-06-12
TAM: always generate and verify

ThomasWaldmann:
remove a lot of complexity from the code that was just there to
support legacy borg versions.

borg < 1.0.9 which did not TAM authenticate the manifest nor the archives.

borg < 1.2.6 did not always (correctly) TAM authenticate the archives.

The upgrade procedure to 1.2.6+ includes enabling TAM authentication for manifest and all archives, including existing archives.


ThomasWaldmann:
@enkore ^

PR #7661, opened 2023-06-22
Archive pytest conversion

bigtedde:
Started with a couple test parameterizations. Will parameterize more tests, trying to ensure I don't impact readability in the process.


src/borg/testsuite/archive.py
```diff
@@ -34,23 +34,22 @@ def test_stats_basic(stats):
...
-    stats.show_progress(stream=out)
-    s = "20 B O 20 B U 1 N "
-    buf = " " * (columns - len(s))
-    assert out.getvalue() == s + buf + "\r"
+    item = Item(path=item_path) if item_path else None
+    s = expected_output
 
-    out = StringIO()
-    stats.update(10**3, unique=False)
-    stats.show_progress(item=Item(path="foo"), final=False, stream=out)
```
bigtedde, line 47: `final` is set to `false` by default in `show_progress`, so removing it should have no effect here.

bigtedde:
I spent a considerable amount of time trying to parameterize `test_get_item_uid_gid`, but found it both wasn't easily readable and didn't work.

src/borg/testsuite/archive.py
```diff
@@ -34,23 +34,22 @@ def test_stats_basic(stats):
...
 
 
-def tests_stats_progress(stats, monkeypatch, columns=80):
+@pytest.mark.parametrize(
+    "item_path, update_size, expected_output",
+    [
+        ("", 0, "20 B O 20 B U 1 N "),  # test unchanged 'stats' fixture
+        ("foo", 10**3, "1.02 kB O 20 B U 1 N foo"),  # test updated original size and set item path
+        ("foo" * 40, 10**3, "1.02 kB O 20 B U 1 N foofoofoofoofoofoofoofoofo...foofoofoofoofoofoofoofoofoofoo"),
+    ],  # test long item path which exceeds 80 characters
```
ThomasWaldmann, line 43: the comment belongs to line 42. so either put it there (if it does not exceed max line length) or above that line.

ThomasWaldmann (changes requested):
Guess we can merge this, just please move the comment to where it belongs.

ThomasWaldmann:
BTW, if you think something is finished, switch it from draft to "ready for review".

ThomasWaldmann (approved):

ThomasWaldmann:
Thanks!

PR #7676, opened 2023-06-25
Properly normalise paths on Windows

### LINKED ISSUE #7666: correctly normalize archived paths (windows)
borg2 / master branch:

`test_archived_paths` is currently skipped on win32 because it fails.

- there are unwanted drive letters in archived paths (`D:`)
- there are unwanted absolute paths in archived paths (`/a/...`)

```
____________________ ArchiverTestCase.test_archived_paths _____________________
[gw0] win32 -- Python 3.10.12 D:/a/_temp/msys64/ucrt64/bin/python.exe

self = <borg.testsuite.archiver.create_cmd.ArchiverTestCase testMethod=test_archived_paths>

    def test_archived_paths(self):
        # As borg comes from the POSIX (Linux, UNIX) world, a lot of stuff assumes path separators
        # to be slashes "/", e.g.: in archived items, for pattern matching.
        # To make our lives easier and to support cross-platform extraction we always use slashes.
        # Similarly, archived paths are expected to be full, but relative (have no leading slash).
        full_path = os.path.abspath(os.path.join(self.input_path, "test"))
        # remove windows drive letter, if any:
        posix_path = full_path[2:] if full_path[1] == ":" else full_path
        # only needed on windows in case there are backslashes:
        posix_path = posix_path.replace("\\", "/")
        # no leading slash in borg archives:
        archived_path = posix_path.lstrip("/")
        self.create_regular_file("test")
        self.cmd(f"--repo={self.repository_location}", "rcreate", "--encryption=none")
        self.cmd(f"--repo={self.repository_location}", "create", "test", "input", full_path)
        # "input" directory is recursed into, "input/test" is discovered and joined by borg's recursion.
        # full_path was directly given as a cli argument and should end up as archive_path in the borg archive.
        expected_paths = sorted(["input", "input/test", archived_path])
        # check path in archived items:
        archive_list = self.cmd(f"--repo={self.repository_location}", "list", "test", "--short")
>       assert expected_paths == sorted([path for path in archive_list.splitlines() if path])
E       AssertionError: assert ['a/_temp/msys64/tmp/tmp72a38lrc/input/test', 'input', 'input/test'] == ['D:/a/_temp/msys64/tmp/tmp72a38lrc/input/test', 'input', 'input/test']
E         At index 0 diff: 'a/_temp/msys64/tmp/tmp72a38lrc/input/test' != 'D:/a/_temp/msys64/tmp/tmp72a38lrc/input/test'
E         Full diff:
E         - ['D:/a/_temp/msys64/tmp/tmp72a38lrc/input/test', 'input', 'input/test']
E         ?   ---
E         + ['a/_temp/msys64/tmp/tmp72a38lrc/input/test', 'input', 'input/test']
```

RayyanAnsari:
Fixes #7666

### DIFF
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -1,7 +1,7 @@
 import errno
 import hashlib
 import os
-import os.path
+import posixpath
 import re
 import stat
 import subprocess
@@ -231,7 +231,7 @@ def make_path_safe(path):
     path = path.lstrip("/")
     if path.startswith("../") or "/../" in path or path.endswith("/..") or path == "..":
         raise ValueError(f"unexpected '..' element in path {path!r}")
-    path = os.path.normpath(path)
+    path = posixpath.normpath(path)
     return path
 
 
@@ -245,6 +245,11 @@ def remove_dotdot_prefixes(path):
 
     `path` is expected to be normalized already (e.g. via `os.path.normpath()`).
     """
+    if is_win32:
+        if len(path) > 1 and path[1] == ":":
+            path = path.replace(":", "", 1)
+        path = path.replace("\\", "/")
+
     path = path.lstrip("/")
     path = _dotdot_re.sub("", path)
     if path in ["", ".."]:
--- a/src/borg/testsuite/archiver/create_cmd.py
+++ b/src/borg/testsuite/archiver/create_cmd.py
@@ -110,8 +110,8 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         # To make our lives easier and to support cross-platform extraction we always use slashes.
         # Similarly, archived paths are expected to be full, but relative (have no leading slash).
         full_path = os.path.abspath(os.path.join(self.input_path, "test"))
-        # remove windows drive letter, if any:
-        posix_path = full_path[2:] if full_path[1] == ":" else full_path
+        # remove colon from windows drive letter, if any:
+        posix_path = full_path.replace(":", "") if full_path[1] == ":" else full_path
         # only needed on windows in case there are backslashes:
         posix_path = posix_path.replace("\\", "/")
         # no leading slash in borg archives:
--- a/src/borg/testsuite/archiver/create_cmd.py
+++ b/src/borg/testsuite/archiver/create_cmd.py
@@ -103,7 +103,6 @@ class ArchiverTestCase(ArchiverTestCaseBase):
         # the interesting parts of info_output2 and info_output should be same
         self.assert_equal(filter(info_output), filter(info_output2))
 
-    @pytest.mark.skipif(is_win32, reason="still broken on windows")
     def test_archived_paths(self):
         # As borg comes from the POSIX (Linux, UNIX) world, a lot of stuff assumes path separators
         # to be slashes "/", e.g.: in archived items, for pattern matching.


src/borg/helpers/fs.py
```diff
@@ -245,6 +245,8 @@ def remove_dotdot_prefixes(path):
 
     `path` is expected to be normalized already (e.g. via `os.path.normpath()`).
     """
+    if is_win32:
+        path = os.path.splitdrive(path)[1]
```
ThomasWaldmann, line 249: what i already wondered: is borg already getting paths with slashes (and not backslashes)?

RayyanAnsari: added some logging, it looks like it is:
```
$ pytest -s -k "test_archived_paths and not remote and not binary"
=========================================================================================== test session starts ============================================================================================
platform win32 -- Python 3.10.9, pytest-7.2.0, pluggy-1.0.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
Tests enabled: root, hardlinks, modes
Tests disabled: BSD flags, fuse2, fuse3, symlinks, atime/mtime
rootdir: C:/msys64/home/Rayyan/borg, configfile: setup.cfg
plugins: benchmark-4.0.0, cov-4.0.0, forked-1.4.0, xdist-3.1.0
collected 1746 items / 1745 deselected / 1 selected

src/borg/testsuite/archiver/create_cmd.py
self.cmd output START
self.cmd output END


self.cmd output START
remove_dotdot_prefixes path =  input
remove_dotdot_prefixes path =  input/test
remove_dotdot_prefixes path =  C:/Users/Rayyan/AppData/Local/Temp/tmpz08pkzkq/input/test
self.cmd output END


self.cmd output START
input
input/test
Users/Rayyan/AppData/Local/Temp/tmpz08pkzkq/input/test
self.cmd output END


self.cmd output START
{"flags": 0, "gid": 0, "group": "root", "healthy": true, "hlid": "", "mode": "drwxrwxrwx", "mtime": "2023-06-25T21:38:16.244628+01:00", "path": "input", "size": 0, "target": "", "type": "d", "uid": 0, "user": "root"}
{"flags": 0, "gid": 0, "group": "root", "healthy": true, "hlid": "", "mode": "-rw-rw-rw-", "mtime": "2023-06-25T21:38:16.244628+01:00", "path": "input/test", "size": 0, "target": "", "type": "-", "uid": 0, "user": "root"}
{"flags": 0, "gid": 0, "group": "root", "healthy": true, "hlid": "", "mode": "-rw-rw-rw-", "mtime": "2023-06-25T21:38:16.244628+01:00", "path": "Users/Rayyan/AppData/Local/Temp/tmpz08pkzkq/input/test", "size": 0, "target": "", "type": "-", "uid": 0, "user": "root"}
self.cmd output END

.

==================================================================================== 1 passed, 1745 deselected in 1.63s ==================================================================================== 
```

ThomasWaldmann: Do you know where exactly the usual backslashes on windows are transformed to slashes?

RayyanAnsari: just tested making an archive manually with borg.exe, looks like the paths are still backslashes...
strange. will look into this more.

RayyanAnsari: appears to work fine now but i don't know if posixpath.normpath is needed in other places as well

ams-tschoening:
There was a discussion on the mailing list about not taking drive letters into account on Windows, which prevents using files of multiple drives in the same archive. That feels a bit strange, especially as some softwares like Cygwin simply use the drive letter character as part of some path and Borg might do the same. The following is an example:

```
C:\windows\system32\something
D:\windows\foobar
```
vs.

```
C/windows/system32/something
D/windows/foobar
```

ThomasWaldmann:
@ams-tschoening makes sense.

Just one thing might complicate this a bit:
guess we don't always have drive letters, so we would only include them if they were part of the recursion root?

E.g.
```
borg create ARCHIVE c:\windows Documents
```
would we back up this like?:
```
c/windows/...
...
Documents/example.doc
...
```

ams-tschoening:
> ```
> c/windows/...
> ...
> Documents/example.doc
> ...
> ```

In my opinion yes and should be in-line with what Borg documents about paths in other OS as well: They are used as is. If one wants relative paths, one needs to CD into parents (`Documents`), otherwise the whole path prefix is used (`C:\Windows`). Borg doesn't calculate some path prefix for other relative paths on other OS as well, so don't need to here. The only question is if to remove `C:\` and make things special this way, which doesn't seem necessary. The only special thing here is that `C:\` should become part of the path using non-Windows syntax as `C/[...]` and making paths behave mostly the same this way across different OS.

```bash
tschoening@tschoening-nb MINGW64 ~/Documents
$ pwd
/c/Users/tschoening/Documents
```

ThomasWaldmann:
Note: lower case "c" on mingw64 (not uppercase).

ams-tschoening:
> Note: lower case "c" on mingw64 (not uppercase).

Should you care about that? Borg doesn't normalize upper-lower case in any other path or does it on Windows? In my opinion things should be forwarded 1:1.

RayyanAnsari:
> There was a discussion on the mailing list about not taking drive letters into account on Windows, which prevents using files of multiple drives in the same archive. That feels a bit strange, especially as some softwares like Cygwin simply use the drive letter character as part of some path and Borg might do the same. The following is an example:

Do you have a link to this discussion?

ams-tschoening:
https://mail.python.org/pipermail/borgbackup/2023q2/002269.html

An answer is missing, because I've sent privately by accident first.

src/borg/helpers/fs.py
```diff
@@ -245,7 +245,10 @@ def remove_dotdot_prefixes(path):
 
     `path` is expected to be normalized already (e.g. via `os.path.normpath()`).
     """
-    path = path.lstrip("/")
+    if is_win32:
+        path = path.replace(":", "").replace("\\", "/")
```
ThomasWaldmann, line 249: guess this would also affect colons used for alternate data streams (not sure how borg does or should deal with these).

so better check for a colon at index 1 and only then replace **one** of them.

src/borg/helpers/fs.py
```diff
@@ -245,7 +245,10 @@ def remove_dotdot_prefixes(path):
...
+    else:
+        path = path.lstrip("/")
```
ThomasWaldmann, line 251: this must be done on windows also.

ThomasWaldmann (changes requested):

ThomasWaldmann (approved):
LGTM.

RayyanAnsari:
anything stopping this from being merged?

ThomasWaldmann:
No. :)

PR #7702, opened 2023-07-03
BORG_WORKAROUNDS=authenticated_no_key, fixes #7700

### LINKED ISSUE #7700: recover data from a repo in 'authenticated' mode without the key/passphrase?

Yes

Question

So I have this borg archive that was created with --encryption=authenticated (or authenticated-blake2, I can't remember) and I have accessed it before, using the passphrase. But since then I've forgotten it. Now I know the data isn't actually encrypted, so how can I get to it? There's only a post on stackexchange asking this question although it did not have any answers.

ThomasWaldmann:


ThomasWaldmann:
Guess the current code only works for a forgotten passphrase, not for a lost or invalid key (which is stored in the repo config in case of an `authenticated` mode).

PR #7722, opened 2023-07-09
Archiver folder - all tests converted from unittest to pytest

bigtedde:
In this PR, I removed the `ArchiverTestCaseBase` and all unittest elements from the 32 test files in the archiver folder of the testsuite, converting everything to the pytest framework. I will continue to work on review and refinement. 

Any and all feedback is greatly appreciated!

@ThomasWaldmann @m3nu


ThomasWaldmann:
Is this additional or alternative to the other PRs?

If this is "ready for review", use that button.

bigtedde:
> Is this additional or alternative to the other PRs?
> 
> If this is "ready for review", use that button.

Hey Thomas, I should have clarified - this is alternative to the other PR's. After talking with Manu we felt combining the entire archiver folder edits into one package that can pass the CI checks would make it easier to get approved.

All checks have passed and the code coverage difference is negligible, so I will mark it ready for review. Thank you!

src/borg/testsuite/archiver/__init__.py
```diff
@@ -120,260 +124,402 @@ def checkts(ts):
...
+    ret, output = exec_cmd(*args, archiver=archiver.archiver, fork=fork, exe=archiver.EXE, **kw)
+    if ret != exit_code:
+        print(output)
+    assert ret == exit_code
+    # if tests are run with the pure-python msgpack, there will be warnings about
+    # this in the output, which would make a lot of tests fail.
+    pp_msg = PURE_PYTHON_MSGPACK_WARNING.encode() if binary_output else PURE_PYTHON_MSGPACK_WARNING
+    empty = b"" if binary_output else ""
+    output = empty.join(line for line in output.splitlines(keepends=True) if pp_msg not in line)
+    return output
```
ThomasWaldmann, line 142: A lot of (all?) code is calling this with something like `f"--repo={archiver.repo_location}"` as first argument.

Considering that we have `archiver` here, that seems a bit too complicated and redundant as we could also create that argument here.

src/borg/testsuite/archiver/__init__.py
```diff
@@ -120,260 +124,402 @@ def checkts(ts):
...
+
+
+def open_repository(archiver):
+    if archiver.prefix == "":
+        return Repository(archiver.repository_path, exclusive=True)
+    elif archiver.prefix == "ssh://__testsuite__":
+        return RemoteRepository(Location(archiver.repository_location))
+    else:
+        raise ValueError(f"Archiver prefix '{archiver.prefix}' is not a valid prefix.")
+
```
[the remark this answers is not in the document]
ThomasWaldmann: Or just call `archiver.open_repository()`.

ThomasWaldmann (changes requested):
Thanks for this PR, that was a lot of work!

Sadly, due to huge diff, it's not very good to review, so I hope you carefully kept the test semantics and you also keep an eye on the test count, so that there are no unexpected differences.

For now, I had mostly a look at the globally used stuff.

src/borg/testsuite/archiver/create_cmd.py
```diff
@@ -169,14 +168,13 @@ def test_create_duplicate_root(archivers, request):
...
-def test_unix_socket(archivers, request):
+def test_unix_socket(archivers, request, tmp_path):
     archiver = request.getfixturevalue(archivers)
     repo_location = archiver.repository_location
     cmd(archiver, f"--repo={repo_location}", "rcreate", RK_ENCRYPTION)
     try:
-        name = tempfile.mktemp(prefix="test_sock", suffix=".sock")  # short path needed for socket.bind
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        sock.bind(name)
+        sock.bind(os.path.join(archiver.input_path, "unix-socket"))
```
ThomasWaldmann: I suspect that this change breaks the tests somewhere. That "short path needed for socket.bind" was there for a reason and the paths created by pytest are rather longer.

src/borg/testsuite/archiver/__init__.py
ThomasWaldmann: Hmm, maybe better to delay that to after this PR is finished/merged - easier to review if separated PR.

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -1,302 +1,347 @@
...
-        # check for output when timespan older than earliest archive is given. Issue #1711
-        output = self.cmd(
-            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--older=9999m", exit_code=0
-        )
-        for archive in ("archive1", "archive2", "archive3"):
-            self.assert_not_in(archive, output)
+def pytest_generate_tests(metafunc):
+    # Generate tests for different scenarios: local repository, remote repository, and using the borg binary.
+    if "archivers" in metafunc.fixturenames:
+        metafunc.parametrize("archivers", ["archiver", "remote_archiver", "binary_archiver"])
```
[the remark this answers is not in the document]
bigtedde: Yes - that would be a much better approach!

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -1,302 +1,347 @@
...
-        output = self.cmd(
-            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--oldest=23e", exit_code=2
-        )
-        output = self.cmd(
-            f"--repo={self.repository_location}", "check", "-v", "--archives-only", "--oldest=1m", exit_code=0
-        )
-        self.assert_in("archive1", output)
-        self.assert_in("archive2", output)
-        self.assert_not_in("archive3", output)
+def check_cmd_setUp(archiver):
```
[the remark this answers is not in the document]
bigtedde: Will do

conftest.py
```diff
@@ -73,3 +75,77 @@ def undo(self):
...
+    yield archiver
+
+
+@pytest.fixture()
+def binary_archiver(archiver):
+    if "binary" not in BORG_EXES:
+        pytest.skip("No borg.exe binary available")
+    archiver.EXE = "borg.exe"
+    archiver.FORK_DEFAULT = True
+    yield archiver
```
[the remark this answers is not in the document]
bigtedde: I will do this and report back.

bigtedde:
@ThomasWaldmann Thank you for the feedback - I am working to implement these changes now. 

> Sadly, due to huge diff, it's not very good to review, so I hope you carefully kept the test semantics and you also keep an eye on the test count, so that there are no unexpected differences.

On test count: yes, there are exactly 712 tests in the archiver folder before and after my changes. Before converting, I ran each test module and recorded the outcome (number of tests, passes, skips, fails, and execution time on my machine). After converting, I recorded these values again and compared for integrity. 

As far as semantics, I am currently in the process of re-reviewing each module and comparing to the original to ensure test integrity. My goal is to provide you confidence in my work, and I want to help make review of this large PR not overly time-consuming for you.

conftest.py
```diff
@@ -88,21 +88,20 @@ def set_env_variables():
...
+        self.tmpdir: str
+        self.prefix = ""
+        self.repository_path: str
+        self.repository_location: str
+        self.input_path: str
+        self.output_path: str
+        self.keys_path: str
+        self.cache_path: str
+        self.exclude_file_path: str
+        self.patterns_file_path: str
```
ThomasWaldmann, line 104: Hmm, so that declares a type, but does not initialize the attribute? But guess we rather want to also have them initialized.

src/borg/testsuite/archiver/checks.py
```diff
@@ -43,7 +43,7 @@ def cmd_raises_unknown_feature(archiver, args):
 
 
 def pytest_generate_tests(metafunc):
-    # Generate tests for different scenarios: local repository, remote repository, and using the borg binary.
+    # Generate tests for local and remote archivers
     if "archivers" in metafunc.fixturenames:
         metafunc.parametrize("archivers", ["archiver", "remote_archiver"])
```
ThomasWaldmann: guess that function could have a `kinds` argument and you could use lambda or partial to generate the functions you need from that.

kinds could be given like "local,remote,binary", so it is easier to type.

conftest.py
bigtedde: Would it be proper to initialize them to an emptystring `: str = ""`?

I attempted the recommendation of `: str = None` but received the warning "expected type 'str', got 'None' instead"

ThomasWaldmann: If one does not yet have a valid str value to initialise to, one must use `None`, so that any usage before the real initialisation happened will blow up early with a TypeError (and that would be noticed as a code bug). Initialising to "" would just lead to strange / unclear (and "late") errors.

ThomasWaldmann: IIRC, type is `Optional[str]` then.

bigtedde: > If one does not yet have a valid str value to initialise to, one must use None, so that any usage before the real initialisation happened will blow up early with a TypeError (and that would be noticed as a code bug). Initialising to "" would just lead to strange / unclear (and "late") errors.

That makes a lot of sense, I appreciate the explanation!

>IIRC, type is Optional[str] then.

Ah, yes, I have used this Optional type before and should have thought about that. Thank you!

src/borg/testsuite/archiver/checks.py
bigtedde: Hmm, I cannot seem to make this work. If I add `kinds` as an argument to `pytest_generate_tests`, then when I import `pytest_generate_tests` into a test module, I cannot call it as `pytest_generate_tests(metafunc, kinds)` because metafunc cannot be given to it. What am I missing here?

conftest.py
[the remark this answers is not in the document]
ThomasWaldmann: ok, so how was the outcome?

src/borg/testsuite/archiver/checks.py
ThomasWaldmann: Guess I'ld rename the function definition to just `generate_archiver_tests`, then import that and then define:
```
pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="...")
```

conftest.py
```diff
@@ -73,3 +76,76 @@ def undo(self):
...
+
+
+class ArchiverSetup:
+    EXE: str = None  # python source based
+    FORK_DEFAULT = False
+    BORG_EXES = []
+
+    def __init__(self):
+        self.archiver = None
+        self.prefix = ""
```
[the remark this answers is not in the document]
bigtedde: A couple tests used it:

https://github.com/borgbackup/borg/blob/33645ad38a312662147a4f72d544fa25e0cff71c/src/borg/testsuite/archiver/rcreate_cmd.py#L26

https://github.com/borgbackup/borg/blob/33645ad38a312662147a4f72d544fa25e0cff71c/src/borg/testsuite/archiver/checks.py#L230

However, I could remove prefix from `archiver`, set it only in `remote_archiver`, and then just replace these instances of `prefix` with a `hasattr(archiver, 'prefix')` check first. Just wasn't sure if this would make the code more difficult to read.

conftest.py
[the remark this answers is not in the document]
bigtedde: @ThomasWaldmann Sorry, I need a little more time with this one. Still learning about Vagrant, and building binaries. I took a look at the docs and into the scripts directory for more information, and was able to build the Borg.exe, however I am still working on getting to a place where I can test it. This will be my focus today.

src/borg/testsuite/archiver/checks.py
```diff
@@ -9,391 +8,449 @@
...
+    os.mkdir(mountpoint)
+    # XXX this might hang if it doesn't raise an error
+    cmd_raises_unknown_feature(archiver, [f"--repo={repo_location}::test", "mount", mountpoint])
+
+
+@pytest.mark.allow_cache_wipe
+def test_unknown_mandatory_feature_in_cache(archivers, request):
+    archiver = request.getfixturevalue(archivers)
+    repo_location, repo_path = archiver.repository_location, archiver.repository_path
+    remote_repo = True if archiver.prefix == "ssh://__testsuite__" else False
```
ThomasWaldmann, line 282: you can get rid of this .prefix usage by looking into archiver.repository_location.

conftest.py
```diff
@@ -73,3 +76,76 @@ def undo(self):
...
+        self.patterns_file_path: Optional[str] = None
+
+
+@pytest.fixture()
+def archiver(tmp_path, set_env_variables):
+    archiver = ArchiverSetup()
+    archiver.archiver = not archiver.FORK_DEFAULT and Archiver() or None
+    archiver.tmpdir = tmp_path
+    archiver.repository_path = os.fspath(tmp_path / "repository")
+    archiver.repository_location = archiver.prefix + archiver.repository_path
```
ThomasWaldmann, line 114: archiver.prefix is always the empty string here?

conftest.py
```diff
@@ -73,3 +76,76 @@ def undo(self):
...
+    old_wd = os.getcwd()
+    os.chdir(archiver.tmpdir)
+    yield archiver
+    os.chdir(old_wd)
+
+
+@pytest.fixture()
+def remote_archiver(archiver):
+    archiver.prefix = "ssh://__testsuite__"
+    archiver.repository_location = archiver.prefix + str(archiver.repository_path)
```
ThomasWaldmann, line 141: archiver.prefix is known here, obviously.

src/borg/testsuite/archiver/rcreate_cmd.py
```diff
@@ -9,79 +8,74 @@
...
+
+    parent_path = os.path.join(archiver.tmpdir, "parent1", "parent2")
+    repository_path = os.path.join(parent_path, "repository")
+    repository_location = archiver.prefix + repository_path
+    with pytest.raises(Repository.ParentPathDoesNotExist):
+        # normal borg rcreate does NOT create missing parent dirs
+        cmd(archiver, f"--repo={repository_location}", "rcreate", "--encryption=none")
+    # but if told so, it does:
+    cmd(archiver, f"--repo={repository_location}", "rcreate", "--encryption=none", "--make-parent-dirs")
+    assert os.path.exists(parent_path)
```
ThomasWaldmann, line 28: guess this might be the trickiest one to solve.

bigtedde: For this one, I was thinking something like:

`repository_location = archiver.prefix + repository_path if hasattr(archiver, 'prefix') else repository_path`

conftest.py
bigtedde: Yes, always. I can change this 👍

conftest.py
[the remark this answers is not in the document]
ThomasWaldmann: Maybe it is cleaner to set `self.prefix = None` here.

conftest.py
ThomasWaldmann: Maybe it is cleaner to set `self.prefix = ""` here.

src/borg/testsuite/archiver/rcreate_cmd.py
ThomasWaldmann: That does not remove usage of .prefix, just makes it more strange.

ThomasWaldmann: Guess it could be solved by adding other kinds of archivers, that return a local / remote repo without existing parent paths.

conftest.py
bigtedde: Makes sense, I will move it here

src/borg/testsuite/archiver/rcreate_cmd.py
bigtedde: Could I just check if `archiver.repository_location.startswith("ssh://__testsuite__")` If it does then I would set the new `repository_location = "ssh://__testsuite__" + repository_path`, else I would set `repository_location = repository_path`

bigtedde:
@ThomasWaldmann I think I was misunderstanding you before. I was under the impression you just wanted `prefix` removed from `ArchiverSetup` and left in the `remote_archiver` setup. If you want `prefix` removed entirely, I think this can be done - everywhere I am currently checking the prefix, I can instead just check if `archiver.repository_location.startswith("ssh://__testsuite__")`

src/borg/testsuite/archiver/checks.py
```diff
@@ -17,7 +17,9 @@
 from .. import llfuse
 from .. import changedir, environment_variable
 from . import cmd, _extract_repository_id, open_repository, check_cache, create_test_files, create_src_archive
-from . import _set_repository_id, create_regular_file, assert_creates_file, RK_ENCRYPTION
+from . import _set_repository_id, create_regular_file, assert_creates_file, generate_archiver_tests, RK_ENCRYPTION
+
+pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,remote")  # NOQA
```
bigtedde, line 22: Received a linting error: `E731 do not assign a lambda expression, use a def`


To avoid the linting error, I could do something like:

```
def pytest_generate_tests(metafunc):
    generate_archiver_tests(metafunc, kinds="local,remote,binary")

```

However, somewhat ironically, this would bring us almost full circle to what we were originally trying to avoid: re-defining `pytest_generate_tests` in each file 😂

conftest.py
```diff
@@ -93,7 +93,6 @@ class ArchiverSetup:
 
     def __init__(self):
         self.archiver = None
-        self.prefix = ""
```
bigtedde: I removed all instances of 'prefix', which I believe is closer to your original desire. Since the prefix was essentially being used to indirectly check if the `archiver` was remote or not, I was considering just adding an `archiver` attribute called `is_remote` that is initialized to false but is set to true during the `remote_archiver` setup. 

Then, in the tests, instead of again indirectly checking if it is remote by looking into the `repository_location` as I have done here, I could instead use `archiver.is_remote` to check directly. Would this be a good approach?

src/borg/testsuite/archiver/checks.py
ThomasWaldmann: Well, both the assignment with lambda as well as the `def` define a function, so it is just 1 LOC difference. So guess just using the `def` is the simplest solution.

conftest.py
ThomasWaldmann: I also thought about something like `is_remote`,  but I think checking the `location` is better.

In future we might have different kinds of locations, not just local vs. ssh:// remote. In fact we already have that now with `socket://`.

src/borg/testsuite/archiver/check_cmd.py
```diff
@@ -9,8 +9,10 @@
 from ...helpers import msgpack
 from ...manifest import Manifest
 from ...repository import Repository
-from . import pytest_generate_tests  # NOQA
-from . import cmd, src_file, create_src_archive, open_archive, RK_ENCRYPTION
+from . import cmd, src_file, create_src_archive, open_archive, generate_archiver_tests, RK_ENCRYPTION
+
+# Tests that include the 'archivers' argument will generate a tests for each kind of archivers specified.
+pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,remote,binary")  # NOQA
```
ThomasWaldmann, line 15: NOQA would be also totally fine here to avoid the linter error.

src/borg/testsuite/archiver/checks.py
ThomasWaldmann: or NOQA as you did it already it also fine.

src/borg/testsuite/archiver/key_cmds.py
```diff
@@ -12,8 +12,10 @@
 from ...repository import Repository
 from .. import environment_variable
 from .. import key
-from . import RK_ENCRYPTION, KF_ENCRYPTION, cmd, _extract_repository_id, _set_repository_id
-from . import pytest_generate_tests  # NOQA
+from . import RK_ENCRYPTION, KF_ENCRYPTION, cmd, _extract_repository_id, _set_repository_id, generate_archiver_tests
+
+# Tests that include the 'archivers' argument will generate a tests for each kind of archivers specified.
```
ThomasWaldmann, line 17: you could move that to the docstring of generate_archiver_tests, so you don't need to duplicate this comment everywhere.

src/borg/testsuite/archiver/checks.py
ThomasWaldmann: NOQA is fine to avoid this here.

conftest.py
bigtedde: What if I added a `get_kind()` method to the `ArchiverSetup` class that returns the `kind` as a string, in a similar way that we name the `kinds` when generating tests. This would be a consistent approach, could lead to more readable code  then checking prefixes, and if we add different `kinds` later we could add them to the method easily.

conftest.py
```diff
@@ -103,6 +103,14 @@ def __init__(self):
...
         self.patterns_file_path: Optional[str] = None
 
+    def get_kind(self) -> str:
+        if self.repository_location.startswith("ssh://__testsuite__"):
+            return "remote"
+        elif self.EXE == "borg.exe":
+            return "binary"
+        else:
+            return "local"
+
```
bigtedde, line 113: We could add different `kinds` down the road if we like.

src/borg/testsuite/archiver/transfer_cmd.py
```diff
@@ -45,7 +45,7 @@ def check_repo(repo_option):
 
 def test_transfer_upgrade(archivers, request):
     archiver = request.getfixturevalue(archivers)
-    if archiver.repository_location.startswith("ssh://__testsuite__") or archiver.EXE == "borg.exe":
+    if archiver.get_kind() in ["remote", "binary"]:
```
bigtedde, line 48: I think this could be easier to read for other contributors.

src/borg/testsuite/archiver/key_cmds.py
bigtedde: Done!

conftest.py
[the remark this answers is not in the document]
bigtedde: Update - I have a call with Manu tomorrow and we are going to discuss Vagrant and testing on a Borg binary.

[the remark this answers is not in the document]
ThomasWaldmann: You don't need to do anything with vagrant. I just pointed you to the Vagrantfile, because you can see all commands there you need to build a binary for local testing.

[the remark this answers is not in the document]
ThomasWaldmann: maybe you could even just use a downloaded 2.0.0beta binary and not build it yourself, see github releases page (assuming the current tests still fit the behaviour of the last beta).

src/borg/testsuite/archiver/rcreate_cmd.py
```diff
@@ -9,79 +8,75 @@
...
+from . import cmd, generate_archiver_tests, RK_ENCRYPTION, KF_ENCRYPTION
+
+pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,remote,binary")  # NOQA
+
+
+def test_rcreate_parent_dirs(archivers, request):
+    archiver = request.getfixturevalue(archivers)
+    if archiver.EXE:
+        pytest.skip("does not raise Exception, but sets rc==2")
+    remote_repo = True if archiver.get_kind() == "remote" else False
```
ThomasWaldmann, line 20: if you code a conditional switch to True/False like this one, always consider that the condition itself is already boolean.

```
remote_repo = archiver.get_kind() == "remote"
```

ThomasWaldmann:
OK, guess this can be merged soon? The one minor thing I found right now plus the borg.exe testing, anything else?

src/borg/testsuite/archiver/rcreate_cmd.py
bigtedde: Yeah, the way I wrote this was a bit redundant, I'll get that changed.

bigtedde:
> OK, guess this can be merged soon? The one minor thing I found right now plus the borg.exe testing, anything else?

I just pushed the `remote_repo` boolean check change. Also, with a bit of assistance from Manu I was able to test on the binary and all tests passed without any unexpected behavior. 

I do believe these changes are good to merge with your approval. There are the same number of tests, they cover the same amount of code, and they all pass as expected.  All the unittest elements have been removed and replaced with pytest, and I believe these changes have left the test suite easier to read and build upon going forward.

ThomasWaldmann (approved):
LGTM, thanks for refactoring this!

ThomasWaldmann:
@bigtedde can you try that now? https://github.com/borgbackup/borg/pull/7722#discussion_r1263712611

bigtedde:
>A lot of (all?) code is calling this with something like f"--repo={archiver.repo_location}" as first argument.

>Considering that we have archiver here, that seems a bit too complicated and redundant as we could also create that argument here.

I searched though the directory and can confirm that all `cmd` calls are with `f"--repo={archiver.repo_location}"` as the first argument. It is redundant, and I can create that argument in the cmd function itself.

bigtedde:
>LGTM, thanks for refactoring this!

Thank you for your help and mentoring along the way Thomas. It means a lot to me, and I am very excited to have this merged in!

src/borg/testsuite/archiver/__init__.py
bigtedde: Addressed this issue [here](https://github.com/borgbackup/borg/pull/7734).

PR #7734, opened 2023-07-24
`cmd` function now creates repo_location argument

bigtedde:
Solves [this comment](https://github.com/borgbackup/borg/pull/7722#issuecomment-1646961750).

Currently, every cmd in `testsuite/archiver/` tests are passing `f"repo={archiver.repository_location}"` as an argument, in addition to the `archiver` itself. @ThomasWaldmann suggested just creating this argument in the function itself, thus removing a lot of redundant and messy code. That is what I have done here.

### DIFF
--- a/src/borg/testsuite/archiver/__init__.py
+++ b/src/borg/testsuite/archiver/__init__.py
@@ -147,7 +147,9 @@ def cmd(archiver, *args, **kw):
     binary_output = kw.get("binary_output", False)
     if fork is None:
         fork = archiver.FORK_DEFAULT
-    ret, output = exec_cmd(*args, archiver=archiver.archiver, fork=fork, exe=archiver.EXE, **kw)
+    ret, output = exec_cmd(
+        f"--repo={archiver.repository_location}", *args, archiver=archiver.archiver, fork=fork, exe=archiver.EXE, **kw
+    )
     if ret != exit_code:
         print(output)
     assert ret == exit_code
@@ -160,11 +162,10 @@ def cmd(archiver, *args, **kw):
 
 
 def create_src_archive(archiver, name, ts=None):
-    repo_location, source_dir = archiver.repository_location, src_dir
     if ts:
-        cmd(archiver, f"--repo={repo_location}", "create", "--compression=lz4", f"--timestamp={ts}", name, source_dir)
+        cmd(archiver, "create", "--compression=lz4", f"--timestamp={ts}", name, src_dir)
     else:
-        cmd(archiver, f"--repo={repo_location}", "create", "--compression=lz4", name, source_dir)
+        cmd(archiver, "create", "--compression=lz4", name, src_dir)
 
 
 def open_archive(repo_path, name):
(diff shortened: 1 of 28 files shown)


src/borg/testsuite/archiver/checks.py
```diff
@@ -46,249 +46,243 @@ def cmd_raises_unknown_feature(archiver, args):
...
         with pytest.raises(Cache.EncryptionMethodMismatch):
-            cmd(archiver, f"--repo={repo_location}", "create", "test.2", "input")
+            cmd(archiver, "create", "test.2", "input")
 
 
 def test_repository_swap_detection2(archivers, request):
     archiver = request.getfixturevalue(archivers)
-    repo_location, repo_path, input_path = archiver.repository_location, archiver.repository_path, archiver.input_path
-    create_test_files(input_path)
-    cmd(archiver, f"--repo={repo_location}_unencrypted", "rcreate", "--encryption=none")
```
bigtedde, line 70: There were only a handful of instances where the first argument was not **exactly**`f"--repo={repo_location}`. This file had most of those instances, so I updated `archiver.repository_location` to mimic the behavior.

src/borg/testsuite/archiver/__init__.py
```diff
@@ -270,29 +271,28 @@ def _set_repository_id(repo_path, id):
 
 
 def _extract_hardlinks_setup(archiver):
-    repo_location, input_path = archiver.repository_location, archiver.input_path
```
ThomasWaldmann, line 273: maybe keep the input_path part, it is used many times and shorter/quicker.

src/borg/testsuite/archiver/__init__.py
```diff
@@ -270,29 +271,28 @@ def _set_repository_id(repo_path, id):
...
-    os.link(os.path.join(input_path, "dir1/source2"), os.path.join(input_path, "dir1/aaaa"))
-
-    cmd(archiver, f"--repo={repo_location}", "rcreate", RK_ENCRYPTION)
-    cmd(archiver, f"--repo={repo_location}", "create", "test", "input")
+    cmd(archiver, "rcreate", RK_ENCRYPTION)
+    cmd(archiver, "create", "test", "input")
 
 
 def _create_test_caches(archiver):
-    repo_location, input_path = archiver.repository_location, archiver.input_path
```
ThomasWaldmann, line 291: same here.

src/borg/testsuite/archiver/rdelete_cmd.py
```diff
@@ -1,25 +1,22 @@
...
-import pytest
-
 from ...constants import *  # NOQA
-from . import create_regular_file, cmd, RK_ENCRYPTION
+from . import create_regular_file, cmd, generate_archiver_tests, RK_ENCRYPTION
+
+pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,remote,binary")  # NOQA
 
 
-@pytest.mark.parametrize("archivers", ["archiver", "remote_archiver", "binary_archiver"])
```
ThomasWaldmann, line 9: did you do a global search for other places you have forgotten to refactor to the new method?

ThomasWaldmann (commented):
much better! nice to have less, better readable test code and to get rid of some of the multi-line `black`ened formats.

src/borg/testsuite/archiver/rdelete_cmd.py
ThomasWaldmann: looks like this and test_rename was all, so this is done.

bigtedde:
I made those fixes you suggested to improve readability. Please let me know if there is anything else you see that needs attention before merging 👍

ThomasWaldmann:
Thanks!

PR #7737, opened 2023-07-25
#6231: Clarify where paths are relative to

jefft:
A documentation patch for #6231 clarifying where include/excludes are relative to. I changed the example to have two directories, to illustrate that the first (not second) directory is used.

### DIFF
--- a/docs/usage/help.rst.inc
+++ b/docs/usage/help.rst.inc
@@ -18,13 +18,13 @@ unwanted ones. Patterns can be used
 - for ``PATH`` arguments that explicitly support them.
 
 Borg always stores all file paths normalized and relative to the
-current recursion root. The recursion root is also named ``PATH`` in
-Borg commands like `borg create` that do a file discovery, so do not
-confuse the root with the ``PATH`` argument of e.g. `borg extract`.
+first directory of the current recursion root. The recursion root is also named
+``PATH`` in Borg commands like `borg create` that do a file discovery, so do
+not confuse the root with the ``PATH`` argument of e.g. `borg extract`.
 
 Starting with Borg 1.2, paths that are matched against patterns always
-appear relative. If you give ``/absolute/`` as root, the paths going
-into the matcher will start with ``absolute/``.
+appear relative. If you give ``/opt/app/`` as root, the paths going
+into the matcher will start with ``opt/``.
 If you give ``../../relative`` as root, the paths will be normalized
 as ``relative/``.
 


docs/usage/help.rst.inc
```diff
@@ -18,13 +18,13 @@ unwanted ones. Patterns can be used
 - for ``PATH`` arguments that explicitly support them.
 
 Borg always stores all file paths normalized and relative to the
-current recursion root. The recursion root is also named ``PATH`` in
-Borg commands like `borg create` that do a file discovery, so do not
-confuse the root with the ``PATH`` argument of e.g. `borg extract`.
+first directory of the current recursion root. The recursion root is also named
+``PATH`` in Borg commands like `borg create` that do a file discovery, so do
+not confuse the root with the ``PATH`` argument of e.g. `borg extract`.
```
ThomasWaldmann, line 23: Yup, that is better!

docs/usage/help.rst.inc
```diff
@@ -18,13 +18,13 @@ unwanted ones. Patterns can be used
...
 
 Starting with Borg 1.2, paths that are matched against patterns always
-appear relative. If you give ``/absolute/`` as root, the paths going
-into the matcher will start with ``absolute/``.
+appear relative. If you give ``/opt/app/`` as root, the paths going
+into the matcher will start with ``opt/``.
```
ThomasWaldmann, line 27: Considering the 2nd example uses `relative`, I'ld rather like to keep `absolute` wording here.

Maybe `/absolute1/absolute2/`?

ThomasWaldmann (changes requested):
Thanks for the PR!

docs/usage/help.rst.inc
jefft: Thanks. I reworded 'absolute' in my subsequent commit.

ThomasWaldmann (approved):
LGTM.

docs/usage/help.rst.inc
```diff
@@ -18,12 +18,12 @@ unwanted ones. Patterns can be used
...
-Borg commands like `borg create` that do a file discovery, so do not
-confuse the root with the ``PATH`` argument of e.g. `borg extract`.
+first directory of the current recursion root. The recursion root is also named
+``PATH`` in Borg commands like `borg create` that do a file discovery, so do
+not confuse the root with the ``PATH`` argument of e.g. `borg extract`.
 
 Starting with Borg 1.2, paths that are matched against patterns always
-appear relative. If you give ``/absolute/`` as root, the paths going
+appear relative. If you give ``/absolute/path/`` as root, the paths going
 into the matcher will start with ``absolute/``.
```
jdchristensen, line 27: I still find the whole discussion of relative paths pretty confusing.  For the root `/absolute/path/`, the first directory is `/absolute`, and if things were "relative" to that first directory, then the paths would *not* start with `absolute`.  I think saying that the paths are "relative" to anything is misleading.  Instead, one could say that the paths that appear *include* the path given as the recursion root, but with any leading slash and leading `../` removed.

docs/usage/help.rst.inc
```diff
@@ -18,12 +18,12 @@ unwanted ones. Patterns can be used
 - for ``PATH`` arguments that explicitly support them.
 
 Borg always stores all file paths normalized and relative to the
-current recursion root. The recursion root is also named ``PATH`` in
-Borg commands like `borg create` that do a file discovery, so do not
-confuse the root with the ``PATH`` argument of e.g. `borg extract`.
+first directory of the current recursion root. The recursion root is also named
+``PATH`` in Borg commands like `borg create` that do a file discovery, so do
+not confuse the root with the ``PATH`` argument of e.g. `borg extract`.
```
jdchristensen, line 23: I find the sentence involving `PATH` confusing.  Why say that because the recursion root is named `PATH` in some commands, you should *not* confuse a `PATH` argument with a recursion root?  Should it be more like "While the recursion root is named PATH in Borg commands like `borg create` that do file discovery, in other commands like `borg extract`, the `PATH` argument should be a normalized path."  ?

docs/usage/help.rst.inc
ThomasWaldmann: Yeah, that would be even better!

docs/usage/help.rst.inc
ThomasWaldmann: Yeah, for create `PATH` argument is a recursion root, for extract it is for matching what to extract.

docs/usage/help.rst.inc
jefft: @jdchristensen, I agree. Does this sound more correct?

```
Borg always stores all file paths normalized, with the first part equal to the
recursion root, but (since 1.2) with any leading / or ../ removed. If you give
``/absolute/path`` as root, the paths going into the matcher will start with
``absolute/path``. If you wanted to exclude e.g.
``/absolute/path/logs/*.log``, the pattern ``logs/*.log`` would not match; you
must use ``absolute/path/logs/*.log`` or ``**/logs/*.log`` (using ``sh``
patterns). Since the leading path separator is stripped in patterns,
``/absolute/path/logs/*.log`` will also match (and is clearest).

The recursion root is also named ``PATH`` in Borg commands like `borg create`
that do a file discovery, so do not confuse the root with the ``PATH``
argument of e.g. `borg extract`.

```

Committed but not yet PR'ed at https://github.com/borgbackup/borg/compare/master...jefft:borg:docfix-6231

jdchristensen: I like the new version.  It's good that it discusses the removal of leading things in the same paragraph and that it gives more examples.  I'm not sure about the sentence that suggests that including the leading / is the clearest, though.

jefft: This whole business of absolute paths pretending to be relative is weird to me. This makes more sense:
```
R /opt/app
- /opt/app/logs/*.log
```
than:
```
R /opt/app
- opt/app/logs/*.log
```
So I think it's at least worth mentioning the first possibility. I've updated the PR again.

docs/usage/help.rst.inc
```diff
@@ -17,16 +17,18 @@ unwanted ones. Patterns can be used
...
+``/absolute/path`` as root, the paths going into the matcher will start with
+``absolute/path``. If you wanted to exclude e.g.
+``/absolute/path/logs/*.log``, the pattern ``logs/*.log`` would not match; you
+must use ``absolute/path/logs/*.log`` or ``**/logs/*.log`` (using ``sh``
+patterns). The pattern ``/absolute/path/logs/*.log`` will also match, since
+the leading path separator is stripped in patterns.
+
+The recursion root is also named ``PATH`` in Borg commands like `borg create`
+that do a file discovery, so do not confuse the root with the ``PATH``
+argument of e.g. `borg extract`.
```
jdchristensen, line 31: I still think this sentence is confusing.  I'd vote for just deleting it, as I can't see how someone would confuse a recursion root with a path they want to extract.  Or maybe just 
```suggestion
The recursion root is named ``PATH`` in Borg commands like `borg create`
that do file discovery.
```

ThomasWaldmann:
Closing this, because:
- it modifies an auto-generated file (sorry for not noticing that back then). the original text is in `src/borg/archiver/help_cmd.py`
- there were meanwhile other changes there clarifying how it works

PR #7736, opened 2023-07-25
remove unused methods of BaseTestCase

ThomasWaldmann:
also: clean up imports / comments

PR #7742, opened 2023-07-26
PR: Remove BaseTestCase from `testsuite/`

bigtedde:
Now that the `ArchiverBaseTestCase` has been removed from `testsuite/archiver/__init__.py`, I would now like to remove `BaseTestCase` from the `testsuite/__init__.py`. After @ThomasWaldmann's recent update to the `BaseTestCase`[here](https://github.com/borgbackup/borg/pull/7736), all that remains are a few assert methods that seem superfluous and could be replaced with the built-in python asserts.

### DIFF
--- a/src/borg/testsuite/helpers.py
+++ b/src/borg/testsuite/helpers.py
@@ -14,39 +14,54 @@ import pytest
 from ..archiver.prune_cmd import prune_within, prune_split
 from .. import platform
 from ..constants import MAX_DATA_SIZE
-from ..helpers import Location
-from ..helpers import Buffer
+from ..helpers import DEFAULTISH, FALSISH, SUPPORT_32BIT_PLATFORMS, TRUISH
 from ..helpers import (
-    partial_format,
-    format_file_size,
-    parse_file_size,
-    format_timedelta,
-    format_line,
+    Buffer,
+    ChunkerParams,
+    ChunkIteratorFileWrapper,
+    Location,
     PlaceholderError,
-    replace_placeholders,
+    ProgressIndicatorPercent,
+    StableDict,
 )
-from ..helpers import remove_dotdot_prefixes, make_path_safe, clean_lines
-from ..helpers import interval
-from ..helpers import get_base_dir, get_cache_dir, get_keys_dir, get_security_dir, get_config_dir, get_runtime_dir
-from ..helpers import is_slow_msgpack
-from ..helpers import msgpack
-from ..helpers import yes, TRUISH, FALSISH, DEFAULTISH
-from ..helpers import StableDict, bin_to_hex
-from ..helpers import parse_timestamp, ChunkIteratorFileWrapper, ChunkerParams
-from ..helpers import archivename_validator, text_validator
-from ..helpers import ProgressIndicatorPercent
-from ..helpers import swidth_slice
-from ..helpers import chunkit
-from ..helpers import safe_ns, safe_s, SUPPORT_32BIT_PLATFORMS
-from ..helpers import popen_with_error_handling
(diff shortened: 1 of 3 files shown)


bigtedde:
@ThomasWaldmann There are four test files remaining in the `testsuite` folder that use `BaseTestCase`. One is `/platform.py`, which uses BaseTestCase and a lot of unittest, so I was considering making this its own PR. The other three are files that include tests used in `selftest`, these are:
1. `/hash index.py`
2. `/crypto.py`
3. `/chunker.py`

I feel that removing the BaseTestCase from these would be straightforward and shouldn't impact the selftest themselves as I would not be introducing any pytest or changing any functionality. However, I wanted to check with you first to see if you are fine with me doing this.

src/borg/testsuite/helpers.py
```diff
@@ -396,37 +395,35 @@ def test_chunkerparams():
...
-        self.assert_equal(make_path_safe(".test"), ".test")
-        self.assert_equal(make_path_safe(".test."), ".test.")
-        self.assert_equal(make_path_safe("..test.."), "..test..")
-        self.assert_equal(make_path_safe("/te..st/foo/bar"), "te..st/foo/bar")
-        self.assert_equal(make_path_safe("/..test../abc//"), "..test../abc")
-
-        for path in rejected_dotdot_paths:
-            with pytest.raises(ValueError, match="unexpected '..' element in path"):
-                make_path_safe(path)
+def test_remove_dot_prefixes():
```
ThomasWaldmann, line 398: the original name had **dotdot**.

ThomasWaldmann (changes requested):
looks good. some more places could use parametrize I guess.

bigtedde:
> looks good. some more places could use parametrize I guess.

I will make that happen. Also, are you okay with me removing BaseTestCase from the 3 files in `selftest`? Being careful to not change any functionality and not introducing pytest.

ThomasWaldmann:
Maybe stay away from the selftest stuff until for now until the other stuff is done.

src/borg/testsuite/helpers.py
ThomasWaldmann: seen this?

bigtedde: Yes, I've got this corrected locally. Adding some additional parameterization to this file and will push soon 🙂

bigtedde:
Cleaned up and parametrized as needed in `helpers.py`

src/borg/testsuite/helpers.py
```diff
@@ -14,39 +14,54 @@
...
+    remove_dotdot_prefixes,
     replace_placeholders,
+    safe_ns,
+    safe_s,
+    safe_unlink,
+    swidth_slice,
+    text_to_json,
+    text_validator,
+    yes,
 )
```
ThomasWaldmann, line 62: That stuff uses to be grouped (more or less what belongs together somehow).

Or did you somehow trigger black doing this?

I rather prefer the horizontal grouped form, less scrolling.

ThomasWaldmann (commented):
Tests look much better with parametrize now.

Just the imports somehow got ugly, IMHO.

src/borg/testsuite/helpers.py
bigtedde: I'll get this changed back to the original 🙂

bigtedde:
@ThomasWaldmann As far as removing BaseTestCase from the `selftest` items - is this something I'd be good to work on now, or should we hold off?

ThomasWaldmann (approved):
LGTM.

PR #7743, opened 2023-07-27
PR: `testsuite/platform.py` conversion to pytest + remove `BaseTestCase`

bigtedde:
This is one of the last remaining files in the test suite to have quite a bit of unittest and `BaseTestCase` usages, so I split it into its own PR. Will convert all unittest to pytest, remove `BaseTestCase` entirely, and rename some tests to follow the naming conventions of the classes they are being pull from.

### DIFF
--- a/src/borg/testsuite/platform.py
+++ b/src/borg/testsuite/platform.py
@@ -5,6 +5,8 @@ import sys
 import tempfile
 import unittest
 
+import pytest
+
 from ..platformflags import is_win32
 from ..platform import acl_get, acl_set, swidth
 from ..platform import get_process_id, process_alive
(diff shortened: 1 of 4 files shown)


bigtedde:
All checks passed, same test coverage before and after. Unittest removed as well as BaseTestCase. Marking as ready for review 👍

src/borg/testsuite/platform.py
```diff
@@ -75,154 +75,162 @@ def are_acls_working():
...
+    acl_get(path, item, os.stat(path), numeric_ids=numeric_ids)
+    return item
+
+
+def set_linux_acl(path, access=None, default=None, numeric_ids=False):
+    item = {"acl_access": access, "acl_default": default}
+    acl_set(path, item, numeric_ids=numeric_ids)
+
+
+@pytest.mark.skipif(not sys.platform.startswith("linux"), reason="linux only test")
```
ThomasWaldmann, line 89: In a later PR `sys.platform.startswith(...)` could be replaced with the already computed `is_someplatform` from `borg.platformflags`.

ThomasWaldmann (commented):
Hmm, now that the test class is gone that functioned as container per platform, you needed to put the platform name in all the function names.

Maybe it would be good to split this into multiple modules for linux, darwin, posix.

bigtedde:
> Maybe it would be good to split this into multiple modules for linux, darwin, posix.

Split into three modules: `platform_linux.py`, `platform_darwin.py`, `platform_posix.py`?

src/borg/testsuite/platform.py
bigtedde: I could go ahead and make those changes to this file in this PR, if you would like.

ThomasWaldmann (commented):
LGTM.

One thing I noticed: we have freebsd platform code, but no freebsd acl tests.
Implementing them would be out of scope of this PR of course, but maybe you could create some sort of dummy platform_freebsd.py test file that has some skipped tests due to "not implemented".

bigtedde:
> One thing I noticed: we have freebsd platform code, but no freebsd acl tests. Implementing them would be out of scope of this PR of course, but maybe you could create some sort of dummy platform_freebsd.py test file that has some skipped tests due to "not implemented".
> 
> After you have created the dummy file, please add a comment to #7745.

Okay, I can do that. I will create a separate PR for that one.

bigtedde:
> LGTM.

Excelent! I have marked this as ready for review. Please let me know if there is anything you would like changed in this PR before merging.

Issue #7745, opened 2023-07-28
tests missing for FreeBSD acl code


ThomasWaldmann:
borg has some platform specific code for ACLs on FreeBSD, but looks like there are no tests for that code (we only have ACL tests for Linux and macOS (darwin)).

bigtedde:
Dummy file created for these tests. Find [here](https://github.com/borgbackup/borg/pull/7748).

ThomasWaldmann:
the freebsd64 vagrant machines needs "acls" option in /etc/fstab.

PR #7748, opened 2023-07-28
`platform_freebsd.py` dummy test file

bigtedde:
Creates a dummy file for ACL testing FreeBSD. Relates to issue #7745.

### DIFF
--- a/src/borg/testsuite/platform_freebsd.py
+++ b/src/borg/testsuite/platform_freebsd.py
@@ -1,4 +1,4 @@
-""" Dummy file for now, will eventually contain FreeBSD ACL tests """
+"""Dummy file for now, will eventually contain FreeBSD ACL tests."""
 import pytest
 
 from .platform import skipif_not_freebsd
--- a/src/borg/testsuite/platform_freebsd.py
+++ b/src/borg/testsuite/platform_freebsd.py
@@ -1,0 +1,25 @@
+""" Dummy file for now, will eventually contain freebsd ACL tests """
+import pytest
+
+from .platform import skipif_not_freebsd
+
+# set module-level skips
+pytestmark = [skipif_not_freebsd]
+
+
+def get_acl():
+    return
+
+
+def get_set_acl():
+    return
+
+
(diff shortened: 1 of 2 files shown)


bigtedde:
@ThomasWaldmann is this similar to what you had in mind for a 'dummy file'?

src/borg/testsuite/platform_freebsd.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.mark.skip(reason="not yet implemented")
+def test_access_acl():
+    pass
+
+
+@pytest.mark.skip(reason="not yet implemented")
+def test_default_acl():
+    pass
```
ThomasWaldmann, line 25: additionally to access and default, freebsd also has `nfs4` acls.

other than that: yeah, that's what I had in mind. like this, it is more obvious that there is something missing and maybe it motivates some developer to add that.

bigtedde: Excellent, I have added that additional test and just committed the changes.

ThomasWaldmann:
OK, looks good. This is finished, right?

bigtedde:
> OK, looks good. This is finished, right?

All set. Marked ready for review 👍

src/borg/testsuite/platform_freebsd.py
```diff
@@ -0,0 +1,30 @@
+""" Dummy file for now, will eventually contain FreeBSD ACL tests """
```
ThomasWaldmann, line 1: Oh, didn't see this initially:
- either `"""bla bla"""
- or:
```
"""
bla
bla
"""

ThomasWaldmann: and add a dot at the end.

ThomasWaldmann: other than that, it's ready for merging.

bigtedde: All set 👍

ThomasWaldmann:
Thanks!

PR #7753, opened 2023-08-01
Fixed exclusion of cache and temp folders in quickstart example.

NoWis9:
The example in the quickstart documentation was missing a leading slash in the excluded folders, resulting in the .cache and temp folders not being excluded.
I am using version 1.1.16, but I found nothing leading me to believe that the path handeling of excluded folders has changed since that version.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -181,8 +181,8 @@ backed up and that the ``prune`` command keeps and deletes the correct backups.
         --show-rc                       \
         --compression lz4               \
         --exclude-caches                \
-        --exclude 'home/*/.cache/*'     \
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
The docs in master branch are for borg2.

Did you notice the version selector on the docs site? You can select 1.1.16 there to get docs for the version you use.

ThomasWaldmann:
For borg >= 1.2 you should use such paths without leading slash.

Previous versions tried to remove leading slashes before matching, but that could only be done easily for simple pattern types, not e.g. for complex regex patterns. Thus borg 1.2+ does not do that any more.

ThomasWaldmann:
@NoWis9 did you see my feedback?

PR #7817, opened 2023-09-11
OpenBSD only: switch to OpenSSL-3.0

bket:
OpenSSL-1.1.1 reached EOL status. Switch building on OpenBSD to OpenSSL-3.0 (LTS version).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -79,7 +79,7 @@ def packages_openbsd
     pkg_add lz4
     pkg_add zstd
     pkg_add git  # no fakeroot
-    pkg_add openssl%1.1
+    pkg_add openssl%3.0
     pkg_add py3-pip
     pkg_add py3-virtualenv
   EOF
--- a/setup.py
+++ b/setup.py
@@ -161,8 +161,8 @@ if not on_rtd:
         # Use openssl (not libressl) because we need AES-OCB via EVP api. Link
         # it statically to avoid conflicting with shared libcrypto from the base
         # OS pulled in via dependencies.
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
@bket did you boot the vagrant machine and did it run the build / tests ok?

bket:
> @bket did you boot the vagrant machine and did it run the build / tests ok?

No, I did not.

The Vagrantfile starts a vm box based on OpenBSD-7.1, which has a package for openssl-3.0.3 in packages-stable. It should(tm) work. Reason for not testing the change to Vagrantfile is that I do not have access to the right tooling.

All OpenSSL related tests have been done locally.

ThomasWaldmann:
@bket ok, I'll give it some practical testing before merging. Sounds good, but one never knows...

ThomasWaldmann:
Something got broken:

When booting the vagrant machine, it somehow stumbles over installing `twine` because there is no rust compiler (which is required to build some requirement of `twine`).

Not sure why this worked before, but guess I could just remove twine from requirements (it is only needed at release time to upload borg to pypi).

ThomasWaldmann:
Removing twine fixed it, merging...

ThomasWaldmann:
Same change for 1.2-maint?

bket:
> Same change for 1.2-maint?

That shouldn't be necessary as LibreSSL provides all the stuff needed by borg-1.2.

borg-2.0 is 'special' as it uses EVP_aes_256_ocb, which LibreSSL does not have.

PR #7830, opened 2023-09-15
replace TAMs by typed repo objects, fixes #7670

### LINKED ISSUE #7670: get rid of the TAM manifest / archive authentication
That stuff was added to borg 1.0.9 (and 1.2.5/1.2.6) to fix a security issue without having to fundamentally change how a borg repo works or how chunks are formatted.

https://borgbackup.readthedocs.io/en/latest/changes_1.x.html#pre-1-0-9-manifest-spoofing-vulnerability-cve-2016-10099

https://borgbackup.readthedocs.io/en/latest/changes_1.x.html#pre-1-2-5-archives-spoofing-vulnerability-cve-2023-36811

https://borgbackup.readthedocs.io/en/latest/internals/security.html#structural-authentication

The root cause of the security issue back then was that borg could not differentiate between:
- **borg metadata** chunks (like the manifest or archives)
- **user file content data** chunks that **look like such borg metadata**

The fix back then was to add this TAM and basically just specially sign the borg metadata by the client.

**In borg2, we could:**

- just make sure borg always knows whether a chunk is user data or borg metadata (and when it wants to read borg metadata of course only accepts borg metadata)
- of course it must not be possible to tamper with this **type information**
- just rely on the AEAD authentication (if the user switched off authentication, like `--encryption none`, they can't expect any authentication or tamper-protection from borg)
- simplify the code a lot and get rid of all the TAM complexity

Actually, this might be a rather small change because a while ago I have already added separately encrypted/authenticated per-chunk metadata (like data size, compressor, compression level, obfuscation infos, etc.).

ThomasWaldmann:
writing: put type into repoobj metadata
reading: check wanted type against type we got

repoobj metadata is encrypted and authenticated.
repoobj data is encrypted and authenticated, also (separately). encryption and decryption of both metadata and data get the same "chunk ID" as AAD, so both are "bound" to that (same) ID.

a repo-side attacker can neither see cleartext metadata/data, nor successfully tamper with it (AEAD decryption would fail).

also, a repo-side attacker could not replace a repoobj A with a differently typed repoobj B without borg noticing:
- the metadata/data is cryptographically bound to its ID. authentication/decryption would fail on mismatch.
- the type check would fail.

thus, the problem (see CVEs in changelog) solved in borg 1 by the manifest and archive TAMs is now already solved by the type check.

### DIFF
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -33,6 +33,14 @@ UMASK_DEFAULT = 0o077
 # forcing to 0o100XXX later
 STDIN_MODE_DEFAULT = 0o660
 
+# RepoObj types
+ROBJ_MANIFEST = "M"  # Manifest (directory of archives, other metadata) object
+ROBJ_ARCHIVE_META = "A"  # main archive metadata object
+ROBJ_ARCHIVE_CHUNKIDS = "C"  # objects with a list of archive metadata stream chunkids
+ROBJ_ARCHIVE_STREAM = "S"  # archive metadata stream chunk (containing items)
+ROBJ_FILE_STREAM = "F"  # file content stream chunk (containing user data)
+ROBJ_DONTCARE = "*"  # used to parse without type assertion (= accept any type)
+
 # in borg < 1.3, this has been defined like this:
 # 20 MiB minus 41 bytes for a PUT header (because the "size" field in the Repository includes
 # the header, and the total size was set to precisely 20 MiB for borg < 1.3).
--- a/src/borg/testsuite/repoobj.py
+++ b/src/borg/testsuite/repoobj.py
@@ -2,6 +2,7 @@ import pytest
 
 from ..constants import ROBJ_FILE_STREAM, ROBJ_MANIFEST, ROBJ_ARCHIVE_META
 from ..crypto.key import PlaintextKey
+from ..helpers.errors import IntegrityError
 from ..repository import Repository
 from ..repoobj import RepoObj, RepoObj1
 from ..compress import LZ4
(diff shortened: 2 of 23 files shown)


src/borg/constants.py
```diff
@@ -33,6 +33,14 @@
...
 STDIN_MODE_DEFAULT = 0o660
 
+# RepoObj types
+ROBJ_MANIFEST = "manifest"
+ROBJ_ARCHIVE_META = "archive-meta"  # main archive metadata object
+ROBJ_ARCHIVE_CHUNKIDS = "archive-chunkids"  # objects with a list of archive metadata stream chunkids
+ROBJ_ARCHIVE_STREAM = "archive-stream"  # archive metadata stream chunk (containing items)
+ROBJ_FILE_STREAM = "file-stream"  # file content stream chunk (containing user data)
+ROBJ_DONTCARE = "dontcare"  # used to parse without type assertion (= accept any type)
+
```
ThomasWaldmann, line 43: Values could be also something shorter to create less overhead.

src/borg/testsuite/repoobj.py
```diff
@@ -80,18 +81,49 @@ def test_borg1_borg2_transition(key):
...
+def test_spoof_archive(key):
+    repo_objs = RepoObj(key)
+    data = b"fake or malicious archive data"  # file content could be provided by attacker.
+    id = repo_objs.id_hash(data)
+    # create a repo object containing user data (file content data).
+    cdata = repo_objs.format(id, {}, data, ro_type=ROBJ_FILE_STREAM)
+    # let's assume an attacker somehow managed to replace an archive with that repo object.
+    # as borg always give the ro_type it wants to read, this should fail:
+    with pytest.raises(AssertionError):
+        repo_objs.parse(id, cdata, ro_type=ROBJ_ARCHIVE_META)
```
ThomasWaldmann, line 129: @enkore Guess that's enough to not need the TAM stuff?

ThomasWaldmann: TODO: replace asserts by raising a custom exception.

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
note: if there are no objections or ongoing reviews, i'll merge this soon.

PR #7836, opened 2023-09-22
Update borg-recreate.1: fix `--target TARGET` description

abdeoliveira:
In `--target TARGET` description it says the new archive will have the  name `ARCHIVE` while the correct is the new name `TARGET`.

### DIFF
--- a/docs/man/borg-recreate.1
+++ b/docs/man/borg-recreate.1
@@ -160,7 +160,7 @@ consider archives older than (now \- TIMESPAN), e.g. 7d oder 12m.
 consider archives newer than (now \- TIMESPAN), e.g. 7d or 12m.
 .TP
 .BI \-\-target \ TARGET
-create a new archive with the name ARCHIVE, do not replace existing archive (only applies for a single archive)
+create a new archive with the name TARGET, do not replace existing archive (only applies for a single archive)
 .TP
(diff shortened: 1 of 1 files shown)


docs/man/borg-recreate.1
```diff
@@ -160,7 +160,7 @@ consider archives older than (now \- TIMESPAN), e.g. 7d oder 12m.
 consider archives newer than (now \- TIMESPAN), e.g. 7d or 12m.
 .TP
 .BI \-\-target \ TARGET
-create a new archive with the name ARCHIVE, do not replace existing archive (only applies for a single archive)
+create a new archive with the name TARGET, do not replace existing archive (only applies for a single archive)
```
ThomasWaldmann, line 163: Ah, good catch! Thanks for finding this.

But, this is not the right place to fix it because these files are automatically generated from the files in `src/borg/archiver/*.py`, so you need to find/fix the string there (master branch).

In 1.2-maint branch, it is in `src/borg/archiver.py`

abdeoliveira: OK! Should I close this PR?

ThomasWaldmann: yes, guess it is easiest to make a new PR.

PR #7837, opened 2023-09-22
Remove recreate --recompress option

ThomasWaldmann:


src/borg/cache.py
```diff
@@ -1115,8 +1113,7 @@ def file_known_and_unchanged(self, hashed_path, path_hash, st):
     def memorize_file(self, hashed_path, path_hash, st, ids):
         pass
 
-    def add_chunk(self, id, meta, data, *, stats, overwrite=False, wait=True, compress=True, size=None):
-        assert not overwrite, "AdHocCache does not permit overwrites — trying to use it for recreate?"
```
ThomasWaldmann, line 1119: @enkore do you remember what the problem was with overwrite=True and AdHocCache (vs. no problem with LocalCache)?

If recreate recompresses a chunk, it calls this with overwrite=True.

enkore: I'd guess because of csize coherency?

ThomasWaldmann: I dropped tracking csize (in archives, in chunks index) a while ago - was just too much pain, standing in the way of other stuff.

Since then borg has gained `borg rcompress` which can efficiently recompress all chunks on repo level without having to touch archive content.

Also, chunks in the repo now have separately encrypted/authenticated metadata, which includes size, csize, ctype, clevel and with PR ... also the **type** of the chunk (as a TAM replacement).

So, guess I could try same/similar `add_chunk` code as in `LocalCache` now?

enkore: The rcompress code is a lot simpler and the I/O pattern it generates should be much better, so it seems totally reasonable to remove the --recompress option.

PR #7846, opened 2023-09-25
files cache with size

ThomasWaldmann:


ThomasWaldmann:
This is trying to prepare a bigger change in how borg works: giving up with precise refcounting and removing chunk size information from the chunks index/cache (so borg could just work with the set of existing chunks with no further information about them in the chunks index/cache).

A first step into that direction is adding the chunk sizes to the files cache (it currently only has the chunk ids).

If borg create detects a file is unchanged, we need to have that information at hand to create the item's chunk list without reading/chunking the file.

PR #7866, opened 2023-10-10
extract_item: do not delete an existing directory if possible

intelfx:
A pre-existing directory might be a Btrfs subvolume that was created by the user ahead of time when restoring several nested subvolumes from a single archive.


intelfx:
I'm also interested in this feature being backported to Borg 1.2. I have a patch that applies on top of `1.2-maint`.

ThomasWaldmann:
Please rebase onto current master to fix the mypy issue.

src/borg/archive.py
```diff
@@ -858,10 +858,13 @@ def same_item(item, st):
...
-            elif stat.S_ISDIR(st.st_mode):
-                os.rmdir(path)
-            else:
+            # remove anything that is not a directory
+            if not stat.S_ISDIR(st.st_mode):
                 os.unlink(path)
+            # only remove a directory if it is conflicting
+            # preserve existing directories because they might be subvolumes
+            elif not stat.S_ISDIR(item.mode):
+                os.rmdir(path)
```
ThomasWaldmann, line 867: It's rather unclear still whether the resulting directory after extraction will be correct (== as in the archive) in all cases:
- simple attrs
- xattrs
- acls

The existing directory could have some metadata set already, but the resulting directory must be exactly what we have in the archived item, not a mix up of fs item and archived item.

src/borg/archive.py
```diff
@@ -858,10 +858,13 @@ def same_item(item, st):
...
             if continue_extraction and same_item(item, st):
                 return  # done! we already have fully extracted this file in a previous run.
-            elif stat.S_ISDIR(st.st_mode):
-                os.rmdir(path)
-            else:
+            # remove anything that is not a directory
+            if not stat.S_ISDIR(st.st_mode):
                 os.unlink(path)
+            # only remove a directory if it is conflicting
+            # preserve existing directories because they might be subvolumes
```
ThomasWaldmann, line 865: is subvolumes a feature implemented by other fs than btrfs?

if not, maybe we better call it "btrfs subvolumes" here, so it is more clear.

src/borg/archive.py
ThomasWaldmann: The code that makes sure about this could be also useful later if we implement extracting into a non-empty directory (for more than the simplest cases, like the already implemented "continue an extraction").

ThomasWaldmann:
A small test with an existing fs dir and then extracting "over" that would be good.

src/borg/archive.py
ThomasWaldmann: hmm, thinking about it...

we currently more or less require extracting into an empty dir (exception: that "continue extraction" feature). we could also require that if there is any pre-existing directory inside the extraction directory, it must not have any special attrs set.

ThomasWaldmann:
@intelfx can you add tests?

ThomasWaldmann:
superceded by #9288.

Issue #6756, opened 2022-06-06
more specific return code / rc / error codes of the borg process


ThomasWaldmann:
Currently, borg has only a few, rather generic return codes: 0 ok, 1 warning, 2 error, >=128 for termination due to signals.

This could be improved (preferably with a breaking release) by moving some stuff to more specific codes. Everything that has no specific code assigned stays at the current code.

First task is to go through the issue tracker and link to all issues talking about (more specific) error codes from here.

Then it needs some planning: what makes sense, what does not, ...

When planning is finished and there is some consensus, implementation can follow.

ThomasWaldmann:
I'll update this list here with whatever we come up with:

```
0 OK
1 WARNING
2 ERROR
128..255 SIGNAL
```

fantasya-pbem:
I'd like to suggest an enhancement of error codes.

We should at least reserve or define a broader range of codes that could be implemented in future versions, for example:

1-63: warning (1 meaning "general warning" as it is now)
64-127: error (64 meaning "general error")

(128+N stays unchanged)

If we don't want to implement more detailed return values yet, at least we would change rc 2 to 64.
Users would have to change their wrappers from rc 2 to rc >= 64 to recognize errors and could keep the logic for warning.

If we want, we could already define some sub-ranges, e.g. "network errors", "disk related", "quota related" and the like.

ThomasWaldmann:
No, rc 0 1 2 need to stay as is. There are enough free codes without changing these.

ThomasWaldmann:
hmm, if we only reserve space for warnings/errors, we only have 0 for success / information. not sure if that is a problem / limitation in practice.

fantasya-pbem:
We could reserve the range from 128+64 onwards, maybe 200+ for possible success codes, if we ever have the need for this. Posix signals / real-time signals have max code 32 / 64.

ThomasWaldmann:
@m3nu @sophie-h ^^^

fantasya-pbem:
**This proposal is a work-in-progress, no need for replying yet.**

**Specific errors**

**3** locked repo (taken from #5158)
*4+ more lock related*
**10** `borg check` issues could not be fixed
*11+ more check related*
**15** `borg init` unsupported encryption mode or key algorithm
*16+ more init or repo related*
**20** `borg create` failed, checkpoint archive left
*21+ more create related*
**40** repository quota reached
**41** backup disk full
*42+ more disk related*
[...]

PR #7867, opened 2023-10-10
Add --lock-rc option that sets a discrete code for locking errors 

intelfx:
Add a new common option such that if `--lock-rc N` is passed, then borg will return `N` instead of `EXIT_FAILURE` if the error condition was due to failure to acquire a lock.

This option can come handy if one is writing a Borg frontend, to distinguish hard errors from cases when the remote repository is simply busy (and as such the error probably does not warrant the user's attention).

---

I'm also interested in this feature (or any equivalent, if a different approach would be suggested by the maintainer) being backported to Borg 1.2. I have a patch that applies on top of `1.2-maint`.


ThomasWaldmann:
Good idea, but I think we should (for compatibility reasons) apply this to 1.2-maint **only**.

For master / borg2, we can just re-define the error codes (see the ticket that is open about that) and not have this cli option.

ThomasWaldmann:
BTW, ignore the mypy failure, I just fixed that.

intelfx:
OK, but #6756 looks like an unknown amount of work with nobody eager to take it on :-) Should I prepare a PR that just picks _some_ exit code, or do you suggest waiting until a decision is made in #6756?

ThomasWaldmann:
Well, guess waiting could take rather long...

OTOH, having a variable or incompatible rc (in 1.2 compared to borg2) does not seem acceptable either.

So guess somebody(tm) needs to at least finish the planning for the **error** codes (which are easier compared to the warning stuff, because there is only 1 error, but there can be multiple warnings).

ThomasWaldmann:
Guess, as of now, this should not be applied at all (neither to master nor to 1.2-maint).

1.4 is intended to supersede 1.2 in the near future and I recently implemented a completely revamped exit code system for errors and warnings (includes locking errors). 1.4.0a1 is ready for alpha-testing this.

1.4-maint was forked from 1.2.7, so besides that bigger change, it should be quite similar as 1.2.

Thanks for the PR nevertheless, this and quite some other tickets talking about more specific exit codes motivated me to implement that on a global scope.

intelfx:
@ThomasWaldmann thanks for working on this! indeed, I would not have had the bandwidth to do it on the global scale anyway, so yeah, thanks for doing it the right way :)

PR #7875, opened 2023-10-15
linux: use pkgconfig to find libacl

steelman:
Using pkgconfig gives a more meaningful error message in case libacl development files are missing.

### DIFF
--- a/docs/usage/general/environment.rst.inc
+++ b/docs/usage/general/environment.rst.inc
@@ -194,6 +194,9 @@ Directories and files:
 Building:
     BORG_OPENSSL_PREFIX
         Adds given OpenSSL header file directory to the default locations (setup.py).
+    BORG_LIBACL_PREFIX
+        Adds given prefix directory to the default locations. If an 'include/acl/libacl.h' is found
+        Borg will be linked against the system libacl instead of a bundled implementation. (setup.py)
     BORG_LIBLZ4_PREFIX
         Adds given prefix directory to the default locations. If a 'include/lz4.h' is found Borg
         will be linked against the system liblz4 instead of a bundled implementation. (setup.py)
--- a/setup.py
+++ b/setup.py
@@ -186,6 +186,17 @@ if not on_rtd:
         dict(extra_compile_args=cflags),
     )
 
+    if sys.platform == "linux":
+        linux_ext_kwargs = members_appended(
+            dict(sources=[platform_linux_source]),
+            lib_ext_kwargs(pc, "BORG_LIBACL_PREFIX", "acl", "libacl", ">=2.3.1"),
+            dict(extra_compile_args=cflags),
+        )
+    else:
+        linux_ext_kwargs = members_appended(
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
Thanks for the PR!

CI indicates that the macOS and windows builds depend on `linux_ext` being defined, which seems strange.

steelman:
Yet another fix is needed.

setup.py
```diff
@@ -186,6 +186,17 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
         dict(extra_compile_args=cflags),
     )
 
+    if sys.platform == "linux":
+        linux_ext_kwargs = members_appended(
+            dict(sources=[platform_linux_source]),
+            lib_ext_kwargs(pc, "BORG_ACL_PREFIX", "acl", "libacl", ">=2.2.53"),
```
ThomasWaldmann, line 192: I guess we need to add BORG_ACL_PREFIX to the docs.

ThomasWaldmann: https://borgbackup.readthedocs.io/en/master/usage/general.html#environment-variables

Also, maybe rename it to `BORG_LIBACL_PREFIX` for better consistency.

ThomasWaldmann:
Note: we can ignore the `Lint / lint` (black) failure, it is a known bug if the latest black release.

setup.py
steelman: > I guess we need to add BORG_ACL_PREFIX to the docs.

Do you mean the man page? If so, the descriptions of other variables refer to "bundled implementation". What does it mean? I can't see any bundled libraries.

ThomasWaldmann: I gave a link, just click on it, then scroll down until you find `BORG_LIBLZ4_PREFIX` - the new env var could be added right above that place, see the corresponding `general.rst` file.

steelman: This file contains exactly the same paragraphs as `docs/man/borg.1` I mentioned (I'll update both files). Are the references to "bundeld libraries" relevant?

ThomasWaldmann: Please only update the rst file, everything else will be automatically updated.

docs/usage/general/environment.rst.inc
```diff
@@ -194,6 +194,9 @@ Directories and files:
 Building:
     BORG_OPENSSL_PREFIX
         Adds given OpenSSL header file directory to the default locations (setup.py).
+    BORG_LIBACL_PREFIX
+        Adds given prefix directory to the default locations. If an 'include/acl/lz4.h' is found
```
ThomasWaldmann, line 198: guess the name is not `lz4.h`. :-)

ThomasWaldmann:
Merged! Thanks for the PR!

ThomasWaldmann:
Just noted a small issue in the old and new docs about BORG_LIB*_PREFIX: borg master branch does not bundle compression code any more (and it never bundled acl lib).

PR #7905, opened 2023-11-05
Python 3.13: include `unistd.h` in `_chunker.c`

FelixSchwarz:
With Python 3.13, Python.h no longer includes the `<unistd.h>` standard header file: https://docs.python.org/3.13/whatsnew/3.13.html#id8

Disclaimer: I only tested this with borg 1.2 as I failed to create a Python 3.13 virtualenv with borg 2. However the fix


ThomasWaldmann:
Thanks!

PR #7928, opened 2023-11-09
optional more specific return codes

### LINKED ISSUE #6756: more specific return code / rc / error codes of the borg process

ThomasWaldmann:
default behaviour is mostly unchanged, but user can enable this via an environment variable.

fixes #6756 and some other issues.

### DIFF
--- a/docs/internals/frontends.rst
+++ b/docs/internals/frontends.rst
@@ -618,7 +618,7 @@ Errors
         Key management not available for unencrypted repositories.
     UnknownKeyType rc: 47 traceback: no
         Key type {0} is unknown.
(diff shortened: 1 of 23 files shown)


ThomasWaldmann:
superseded by #7976.

PR #7944, opened 2023-11-19
Fix typos

Gelma:
Thanks a lot for Borg.

Here's a stupid patch for some typos.


ThomasWaldmann:
Thanks for the PR!

Did you see that:
```
.. IMPORTANT: this file is auto-generated from borg's built-in help, do not edit!
```

This is from the `*.rst.inc` files, but also applies to all below `man/`.

So these files should not be edited / commited.

Usually the correct place to fix texts is in `src/borg/archiver/*.py` - the other files are auto-generated from that (I do that at release time).

ThomasWaldmann:
Oh, "oder" should be "or" (not "older"). These are just 2 examples of a time specification that could be used.

Some german guy's brain seems to have erroneously used the german word for "or" here. :-)

ThomasWaldmann:
@Gelma did you see my feedback?

ThomasWaldmann:
@Gelma I fixed it.

Issue #7802, opened 2023-09-03
Data integrity error: Archive authentication did not verify


sebastianhaberey:
## Is this a BUG / ISSUE report or a QUESTION?

QUESTION, possibly ISSUE

## System information. For client/server mode post info for both machines.

- Client: macOS
- Repository: multiple (local volume & ssh), all same error

#### Your borg version (borg -V).

borg 1.2.6

#### Operating system (distribution) and version.

macOS 13.5.1 (22G90)

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

`borg info /Volumes/BACKUP2/borgbackup`

update: it can also happen with `borg create --stats` as the root cause is in the stats computation code.

## Describe the problem you're observing.

I am getting the output `Data integrity error: Archive authentication did not verify`. This is new (borg has been working fine) and it occured in all three repositories (local volume & ssh) at once.

#### Include any warning/errors/backtraces from the system logs

```
# borg info /Volumes/BACKUP2/borgbackup
Enter passphrase for key /Volumes/BACKUP2/borgbackup:
Repository ID: [redacted]
Location: /Volumes/BACKUP2/borgbackup
Encrypted: Yes (repokey BLAKE2b)
Cache: /Users/user/.cache/borg/[redacted]
Security dir: /Users/user/.config/borg/security/[redacted]
------------------------------------------------------------------------------
                       Original size      Compressed size    Deduplicated size
Data integrity error: Archive authentication did not verify
 ```

ThomasWaldmann:
Did you follow the upgrade hints at top of the changelog?

https://borgbackup.readthedocs.io/en/stable/changes.html#pre-1-2-5-archives-spoofing-vulnerability-cve-2023-36811

sebastianhaberey:
No I didnt - I upgraded borg using brew (and I didn't look at the release notes).

I followed the instructions and everything is well now. Phew. I am quite relieved. Sorry for the hassle.

ThomasWaldmann:
I'll reopen this ticket to make it more visible for a while - guess more people might stumble over this.

sebastianhaberey:
Thank you for your quick reply!

bsdice:
Thoughtful to leave it visible because I found it right away. Arch rolled to 1.2.6 and it would have been nice from packager to print a "WARNING: BREAKING CHANGE" during upgrade.

kylemanna:
[...]

PR #7946, opened 2023-11-22
Remove period that could be interpreted as part of the command

ad8e:
https://github.com/borgbackup/borg/issues/7802#issuecomment-1823651979


ThomasWaldmann (changes requested):
Thanks for the PR!

Can you add a colon after `Run`? Guess that would improve punctuation correctness.

Also, if you like, you could forward port this to master branch (it's in docs/changes-12.rst or so).

ThomasWaldmann (approved):

ThomasWaldmann:
Thanks!

PR #7966, opened 2023-12-07
Change fd_passphrase to read several times

Louson:
Change read to readline so the virtual pipe can send several times the password. It's usefull to pipe the password to a script that runs several borg commands.

### DIFF
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -68,7 +68,7 @@ class Passphrase(str):
         except (ValueError, TypeError):
             return None
         with os.fdopen(fd, mode="r") as f:
-            passphrase = f.read()
+            passphrase = f.readline()
         return cls(passphrase.rstrip("\n"))
 
     @classmethod


ThomasWaldmann:
@Louson Thanks for the PR!

Sounds useful, did you practically try the code?

ThomasWaldmann:
Ah, interesting. Just to avoid that this somehow depends on timing: does that "multi password pipe" also work without that `sleep`?

Louson:
It does not. There's something around buffers and fd.open()

I guess fd.open locks the fd and nothing is written in the pipe meanwhile. I am not sure

ThomasWaldmann:
Hmm, if that means that the password has to be written into the pipe with correct timing, that might be difficult to get right in the general case.

In your example, all commands were rather quick I guess, so each borg command was already waiting for a passphrase when you write it into the pipe.

But what if a borg command takes longer (due to network speed or data volume) - guess it would stop working?

Louson:
I think the issue is only how long you keep open the file descriptor (not very long).

Perhaps there is a better solution, like moving back the seek to the begining of the file

Louson:
The other thing I don't explain is it works when you disable buffering

ThomasWaldmann:
About seeking: is that even possible for such pipes?

Somehow I have doubts whether this change should be applied - if it depends on timing, which may change due to a lot of reasons, it does not seem like a reliable solution.

ThomasWaldmann:
Do you think this can be made working without relying on timing?

If not, I guess we need to abandon this idea because that is too fragile.

Louson:
Well, I have no proper example that does not depends on timing. I understand your point of view, but is there any risk for you to set readline instead read ?

ThomasWaldmann:
Good point.

1. Maybe people could have set a passphrase that includes newline chars. Not very easy to do and also not very likely that somebody does this intentionally, but guess it can be done somehow.

Using readline, we would only read a part of such a multiline passphrase.

The risk would be to break borg completely for previously working setups.

2. In your current implementation, the fd passphrase would behave differently / inconsistently when comparing it to passcommand and env passphrase (where borg always takes the whole output / the whole content, no just one line).

Louson:
Ok, I can understand that

PR #7977, opened 2023-12-23
drop setup_requires, use pip and build

### LINKED ISSUE #6068: setup_requires is deprecated

Yes

ISSUE

tip of  the 1.1-maint branch (c9833911)

Fedora release 33 (Thirty Three)

```
fakeroot -u tox --develop -e py3 borg.testsuite -- -k basic_functionality
```

A warning is printed:
```
/home/user/borg/.tox/py3/lib/python3.9/site-packages/setuptools/__init__.py:148: SetuptoolsDeprecationWarning: setup_requires is deprecated. Supply build dependencies using PEP 517 pyproject.toml build-requires.
```

1. Checkout the git source
2. Install dependencies
2. Run the above command

ThomasWaldmann:


ThomasWaldmann:
@RayyanAnsari can you have a look why it fails only for msys?

I first thought it might be due to the old pyinstaller somehow, but then upgraded to same version as we use in master and it still has same issue here.

Maybe I'll try to move more stuff to `pyproject.tom` here later (similar as in master), not sure if that has an influence. Also (mostly due to the freebsd dir mode issue) I might also try a very recent pyinstaller...

ThomasWaldmann:
trying based on pyinstaller 6.3.0...

ThomasWaldmann:
Ah, adding the `pip install .` helped!

PR #8025, opened 2024-01-03
catch FileNotFoundError in `borg with-lock`, fixes #8022

### LINKED ISSUE #8022: `borg with-lock` raises `FileNotFoundError`
Hey,
I just want to find out what `with-lock` does. I tried some things and I ended up with an exception :)

```bash
kmille@linbox:~ borg --verbose with-lock /home/kmille/tmp/borgrepo list
Local Exception
  File "/usr/lib/python3.11/site-packages/borg/archiver.py", line 5343, in main
                ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/borg/archiver.py", line 5263, in run
    return set_ec(func(args))
                  ^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/borg/archiver.py", line 189, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/borg/archiver.py", line 1853, in do_with_lock
    return subprocess.call([args.command] + args.args, env=env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 389, in call
    with Popen(*popenargs, **kwargs) as p:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'list'

Platform: Linux linbox 6.6.8-hardened1-1-hardened #1 SMP PREEMPT_DYNAMIC Thu, 21 Dec 2023 23:45:24 +0000 x86_64
Linux: Unknown Linux  
Borg: 1.2.7  Python: CPython 3.11.6 msgpack: 1.0.5 fuse: llfuse 1.5.0 [pyfuse3,llfuse]
PID: 186644  CWD: /home/kmille
sys.argv: ['/usr/bin/borg', '--verbose', 'with-lock', '/home/kmille/tmp/borgrepo', 'list']
SSH_ORIGINAL_COMMAND: None
```

I don't understand the [docs](https://borgbackup.readthedocs.io/en/stable/usage/lock.html). After reading, I ask myself:
- What's the purpose of a lock if I can run other borg commands in the meantime
- "It will first try to acquire the lock": Does this just mean: wait until it's free?
- 

Maybe it makes sense to update the docs :)

kmille:
Add exception handling. Prints error message instead of exception if binary run with `with-lock` does not exist:

```bash
(venv) kmille@linbox:borg venv/bin/borg with-lock ~/tmp/borgrepo noo
Could not execute 'noo': [Errno 2] No such file or directory: 'noo'
```

I only handle `FileNotFoundException`. Here are the [subprocess docs about exceptions](https://docs.python.org/3/library/subprocess.html#exceptions). We could also catch [SubprocessError](https://docs.python.org/3/library/subprocess.html#subprocess.SubprocessError), [OSError](https://docs.python.org/3/library/exceptions.html#OSError) and [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) but then I'm still not sure if that's all. I really like python but to find out which exceptions can be raised is... pain..

Let me hear what you think.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1853,7 +1853,7 @@ class Archiver:
             ret = subprocess.call([args.command] + args.args, env=env)
         except (ValueError, FileNotFoundError, OSError) as e:
             self.print_error(f"Could not execute '{args.command}': {e}")
-            ret = -1
+            ret = -2
         finally:
             # we need to commit the "no change" operation we did to the manifest
             # because it created a new segment file in the repository. if we would
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1852,8 +1852,8 @@ class Archiver:
             # we exit with the return code we get from the subprocess
             ret = subprocess.call([args.command] + args.args, env=env)
         except (ValueError, FileNotFoundError, OSError) as e:
-            self.print_error(f"Could not execute '{args.command}': {e}")
-            ret = -2
+            self.print_error(f"Error while trying to run '{args.command}': {e}")
+            ret = EXIT_ERROR
         finally:
             # we need to commit the "no change" operation we did to the manifest
             # because it created a new segment file in the repository. if we would
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1851,7 +1851,7 @@ class Archiver:
         try:
             # we exit with the return code we get from the subprocess
             ret = subprocess.call([args.command] + args.args, env=env)
-        except (ValueError, FileNotFoundError, OSError) as e:
+        except (FileNotFoundError, OSError, ValueError) as e:
             self.print_error(f"Error while trying to run '{args.command}': {e}")
             ret = EXIT_ERROR
         finally:
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1850,13 +1850,10 @@ class Archiver:
         env = prepare_subprocess_env(system=True)
         try:
             # we exit with the return code we get from the subprocess
-            try:
-                ret = subprocess.call([args.command] + args.args, env=env)
-            except FileNotFoundError as e:
-                self.print_error(f"Could not execute '{args.command}': {e}")
-                ret = 1
-            finally:
-                return ret
+            ret = subprocess.call([args.command] + args.args, env=env)
+        except (ValueError, FileNotFoundError, OSError) as e:
+            self.print_error(f"Could not execute '{args.command}': {e}")
+            ret = -1
         finally:
             # we need to commit the "no change" operation we did to the manifest
             # because it created a new segment file in the repository. if we would
@@ -1865,6 +1862,7 @@ class Archiver:
             # any other mechanism relying on existing segment data not changing).
             # see issue #1867.
             repository.commit(compact=False)
+            return ret
 
     @with_repository(manifest=False, exclusive=True)
     def do_compact(self, args, repository):
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1850,7 +1850,13 @@ class Archiver:
         env = prepare_subprocess_env(system=True)
         try:
             # we exit with the return code we get from the subprocess
-            return subprocess.call([args.command] + args.args, env=env)
+            try:
+                ret = subprocess.call([args.command] + args.args, env=env)
+            except FileNotFoundError as e:
+                self.print_error(f"Could not execute '{args.command}': {e}")
+                ret = 1
+            finally:
+                return ret
         finally:
             # we need to commit the "no change" operation we did to the manifest
             # because it created a new segment file in the repository. if we would
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -3188,7 +3188,7 @@ class ArchiverTestCase(ArchiverTestCaseBase):
 
     def test_with_lock_non_existent_command(self):
         self.cmd('init', '--encryption=repokey', self.repository_location)
-        cmd = 'non_existent_command',
+        cmd = ['non_existent_command', ]
         self.cmd('with-lock', self.repository_location, *cmd, fork=True, exit_code=EXIT_ERROR)
 
     def test_recreate_list_output(self):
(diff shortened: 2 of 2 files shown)


kmille:
I removed the try in try. Can you please check if the logic still works (I think so). I changed the return value in case of an error to 2.

src/borg/archiver.py
```diff
@@ -1850,7 +1850,10 @@ def do_with_lock(self, args, repository):
         env = prepare_subprocess_env(system=True)
         try:
             # we exit with the return code we get from the subprocess
-            return subprocess.call([args.command] + args.args, env=env)
+            ret = subprocess.call([args.command] + args.args, env=env)
+        except (ValueError, FileNotFoundError, OSError) as e:
+            self.print_error(f"Could not execute '{args.command}': {e}")
+            ret = -2
```
ThomasWaldmann, line 1856: ret = EXIT_ERROR

src/borg/archiver.py
```diff
@@ -1850,7 +1850,10 @@ def do_with_lock(self, args, repository):
         env = prepare_subprocess_env(system=True)
         try:
             # we exit with the return code we get from the subprocess
-            return subprocess.call([args.command] + args.args, env=env)
+            ret = subprocess.call([args.command] + args.args, env=env)
+        except (ValueError, FileNotFoundError, OSError) as e:
+            self.print_error(f"Could not execute '{args.command}': {e}")
```
ThomasWaldmann, line 1855: It is hard to tell whether this gives a reasonable error message for all the exceptions. We must try to not confuse users.

It is good that you included the original exception message into the error message, so that already helps.

"Could not execute" might be already slightly misleading though (might make people think of x-bit or permissions, but for FileNotFound, it is just not there).

Not sure what conditions lead to ValueError and misc. OSErrors here.

Maybe "Error while trying to run ..."?

ThomasWaldmann (changes requested):
Thanks for the PR!

Some comments below.

ThomasWaldmann:
BTW, in 1.4-maint, dealing with errors is rather different (that was the main reason to increase the minor version number and to do community alpha/beta/rc testing before release).

Maybe it is easier if I do the port to 1.4 and you do the review for that.

kmille:
I applied you changes. Do you want to merge it for 1.2-maint? I can take a look at 1.4-maint and if I'm not too confused I can try to make a PR.

ThomasWaldmann (approved):
LGTM!

src/borg/testsuite/archiver.py
```diff
@@ -3186,6 +3186,11 @@ def test_with_lock(self):
         cmd = 'python3', '-c', 'import os, sys; sys.exit(42 if os.path.exists("%s") else 23)' % lock_path
         self.cmd('with-lock', self.repository_location, *cmd, fork=True, exit_code=42)
 
+    def test_with_lock_non_existent_command(self):
+        self.cmd('init', '--encryption=repokey', self.repository_location)
+        cmd = 'non_existant_command'
```
ThomasWaldmann, line 3191: typo: existent

src/borg/testsuite/archiver.py
```diff
@@ -3186,6 +3186,11 @@ def test_with_lock(self):
...
+        self.cmd('with-lock', self.repository_location, *cmd, fork=True, exit_code=EXIT_ERROR)
```
ThomasWaldmann, line 3192: Guess that *cmd does not quite work as you maybe thought.

Try it in an interactive python interpreter.

ThomasWaldmann (changes requested):

src/borg/testsuite/archiver.py
kmille: I don't see it. Looks [good](https://www.linguee.com/english-german/search?source=auto&query=existent) to me?

src/borg/testsuite/archiver.py
kmille: Also not sure what you mean:
```bash
(venv) kmille@linbox:borg pytest src/borg/testsuite/archiver.py -k with_lock_non -v -s   
...                                                                                                                                                                                          
src/borg/testsuite/archiver.py::ArchiverTestCase::test_with_lock_non_existent_command PASSED
src/borg/testsuite/archiver.py::ArchiverTestCaseBinary::test_with_lock_non_existent_command SKIPPED (no borg.exe available)
src/borg/testsuite/archiver.py::RemoteArchiverTestCase::test_with_lock_non_existent_command PASSED

================================================================================================ 2 passed, 1 skipped, 677 deselected in 2.49s =================================================================================================
(venv) kmille@linbox:borg 
```

src/borg/testsuite/archiver.py
ThomasWaldmann: 3191 has it with an "a".

src/borg/testsuite/archiver.py
ThomasWaldmann:
```
tw@mba2020 ~ % python3
Python 3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin

>>> cmd = 'non_existant_command'

>>> print(*cmd)
n o n _ e x i s t a n t _ c o m m a n d
```

Yes, if something is clear / agreed / resolved in the code, you click that button.

kmille: I thought about that, but the test run without an issue :)
If you like I can add a comma: `cmd = 'non_existent_command', `

src/borg/testsuite/archiver.py
kmille: Silly me...

src/borg/testsuite/archiver.py
ThomasWaldmann: A comma would work, but guess a list would be clearest:
```
cmd = ['non_existent_command']
```

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
Thanks!

I'll port this to 1.4-maint soon...

PR #8098, opened 2024-02-17
optional more specific return codes (master)

ThomasWaldmann:
fwd port of #7976


ThomasWaldmann:
Note: the last few commits may need backporting to 1.4-maint.

PR #8115, opened 2024-02-21
return value fixes (master)

ThomasWaldmann:


ThomasWaldmann:
Note: Already ok in 1.4-maint.

PR #8125, opened 2024-02-24
improve acl_get / acl_set error handling (1.4-maint)

### LINKED ISSUE #4049: error handling in ACL code
currently, the acl get/set functions are called, but their return code is not evaluated.

this should be reviewed, considering:
- is it good? like "doing as much as possible". we also have some other places that try stuff and silently fail (e.g. if a file owned by another user is restored and we are not root: we can not chown to the other user then. we try it and silently fail, so the file is just owned by current user.)
- is it bad? like "hiding severe issues".

Error handling could be added like this (linux):
```
+                rc = acl_set_file(path, ...)
+                if rc < 0:
+                    error_number = errno.errno
+                    raise OSError(error_number, strerror(error_number).decode(), path)
```

As this is platform code, all platforms have to be considered.

### LINKED ISSUE #8139: write ACL tests for FreeBSD
borgbackup has some tests for the platform-specific `acl_get` and `acl_set` calls for Linux and macOS, but not yet for FreeBSD.

See there in `1.4-maint` branch:

https://github.com/borgbackup/borg/blob/1.4.0b1/src/borg/testsuite/platform.py#L74

ThomasWaldmann:
This definitely will need some review and practical testing.

Fixes #4049.
Fixes #8139.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -902,7 +902,11 @@ Utilization of max. archive size: {csize_max:.0%}
                     if not symlink:
                         os.chmod(path, item.mode)
             if not self.noacls:
-                acl_set(path, item, self.numeric_ids, fd=fd)
+                try:
+                    acl_set(path, item, self.numeric_ids, fd=fd)
+                except OSError as e:
+                    if e.errno not in (errno.ENOTSUP, ):
+                        raise
             if not self.noxattrs:
                 # chown removes Linux capabilities, so set the extended attributes at the end, after chown, since they include
                 # the Linux capabilities in the "security.capability" attribute.
@@ -1165,7 +1169,11 @@ class MetadataCollector:
                 attrs['xattrs'] = StableDict(xattrs)
         if not self.noacls:
             with backup_io('extended stat (ACLs)'):
-                acl_get(path, attrs, st, self.numeric_ids, fd=fd)
+                try:
+                    acl_get(path, attrs, st, self.numeric_ids, fd=fd)
+                except OSError as e:
+                    if e.errno not in (errno.ENOTSUP, ):
+                        raise
         return attrs
 
     def stat_attrs(self, st, path, fd=None):
--- a/src/borg/platform/darwin.pyx
+++ b/src/borg/platform/darwin.pyx
@@ -1,6 +1,7 @@
 import os
 
 from libc.stdint cimport uint32_t
+from libc cimport errno
 
 from .posix import user2uid, group2gid
 from ..helpers import safe_decode, safe_encode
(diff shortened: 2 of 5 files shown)


ThomasWaldmann:
macOS (similar API as on *BSD) problem:

It doesn't work:

```
src/borg/testsuite/platform.py:76: in <module>
    class PlatformLinuxTestCase(BaseTestCase):
src/borg/testsuite/platform.py:93: in PlatformLinuxTestCase
    @unittest.skipIf(not are_acls_working(), 'ACLs do not work')
src/borg/testsuite/platform.py:65: in are_acls_working
    acl_get(filepath, read_acl, os.stat(filepath))
src/borg/platform/darwin.pyx:126: in borg.platform.darwin.acl_get
    text = acl_to_text(acl, NULL)
E   FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/61/qtqwzmxd7t1fb19zck5k3q0h0000gn/T/tmpekxlfgg9/file'
```

```
def are_acls_working():
    with unopened_tempfile() as filepath:
        open(filepath, 'w').close()
        try:
            access = b'user::rw-\ngroup::r--\nmask::rw-\nother::---\nuser:root:rw-:9999\ngroup:root:rw-:9999\n'
            acl = {'acl_access': access}
            acl_set(filepath, acl)
            read_acl = {}
            # i checked here: the filepath file definitely EXISTS!
            acl_get(filepath, read_acl, os.stat(filepath))  # BOOM!?
            read_acl_access = read_acl.get('acl_access', None)
            if read_acl_access and b'user::rw-' in read_acl_access:
                return True
        except PermissionError:
            pass
        return False
```

ThomasWaldmann:
Similar code on Linux seems to not have that issue.

ThomasWaldmann:
@m3nu any idea why I get a wrong errno?

m3nu:
`E   FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/h1/8hndypj13nsbj5pn4xsnv1tm0000gn/T/tmpj8gew032/file'`

This error points to a temp file being cleaned too early. We had a similar issue in Vorta once. Solution is to disable cleanup or keeping the reference to the temp file longer.

ThomasWaldmann:
@m3nu I also first suspected that (although the code structure is correct), see the comment in the `are_acls_working` function I posted above - the file IS there.

m3nu:
Then maybe some macOS security feature? Like private temp files in Systemd.

ThomasWaldmann:
Found an issue in borg test helper code: uses "acl_access" (which doesn't exist on macOS, there it is "acl_extended").

But still the error handling is weird on macOS.

ThomasWaldmann:
OK, looks like nobody wants to review this, so we'll test it as is in next beta...

PR #8126, opened 2024-02-25
[1.4-maint] No need to use OpenSSL 3.0 on OpenBSD

bket:
borg-2.0 requires EVP_aes_256_ocb, which is not provided by LibreSSL. As such, OpenSSL-3.0 is needed on OpenBSD. borg-1.4 however does not use EVP_aes_256_ocb thus there is no need to use OpenSSL. Instead rely on LibreSSL.

Undo part of ccb1e92. Tested on OpenBSD -current.


ThomasWaldmann:
Thanks!

PR #8152, opened 2024-03-17
changed insufficiently reserved length for log message 

galqiwi:
During `borg prune` I've encountered a minor issue -- output was not properly aligned.

```
Keeping archive (rule: hourly #3):       2024-02-20-23:13:36                  Tue, 2024-02-20 23:13:38 [53dbb412a3d28b8d788cee0fe189d1ceefd4affd27866d23ffaee7cd45af5971]
Keeping archive (rule: hourly[oldest] #4): 2024-02-20-23:09:22                  Tue, 2024-02-20 23:09:24 [66392999b4cb3a71262f1bdff4ecd05bcb6b74c6dc1ae6a2c9903bcf4a8137b3]
```

This PR fixes this issue. Full command with output can be viewed [here](https://gist.github.com/galqiwi/4402da1a7eb46546c2973f574a261112).

### DIFF
--- a/src/borg/archiver/prune_cmd.py
+++ b/src/borg/archiver/prune_cmd.py
@@ -169,7 +169,7 @@ class PruneMixIn:
                     or (args.list_pruned and archive in to_delete)
                     or (args.list_kept and archive not in to_delete)
                 ):
-                    list_logger.info(f"{log_message:<40} {formatter.format_item(archive, jsonline=False)}")
+                    list_logger.info(f"{log_message:<42} {formatter.format_item(archive, jsonline=False)}")
             pi.finish()
             if sig_int:
                 # Ctrl-C / SIGINT: do not checkpoint (commit) again, we already have a checkpoint in this case.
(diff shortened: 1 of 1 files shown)


galqiwi:
It's not a _proper_ fix, and I don't think that there is a proper fix. We can't know beforehand what length all these messages will have without adding unreasonable amount of complexity

ThomasWaldmann:
@galqiwi Thanks for the PR!

Hmm, did you experience the issue with master branch? Or rather with some borg 1.2.x release (that code is in 1.2-maint branch).

In general, the issue should first get fixed in the branch where it has been encountered and after that, it needs checking whether the other branches are affected also.

About the fix: I guess increasing by 4 instead of by 2 would cover more issues of the same kind.

ThomasWaldmann:
@galqiwi did you see my feedback?

galqiwi:
> @galqiwi did you see my feedback?

Yes, sorry for late response. I've encountered this bug in 1.2.0 version. Will make a PR to the corresponding branch. By looking at the code, I see that 1.3.x and master are affected too.

galqiwi:
> Also, width in this PR might need increasing (e.g. to 44), but you'll see that when practically trying it out.

Yeah, my bad. Fixed it.

ThomasWaldmann:
Thanks!

PR #8195, opened 2024-04-22
vagrant: drop buster VM, fixes #8171

### LINKED ISSUE #8171: borg 1.4: drop or keep the buster-based binary?
The pyinstaller-made fat binaries offered on github releases page are built on debian systems.

They can be usually used on systems with a compatible glibc, thus for compatibility with older systems, an old build system is needed.

But, Debian Buster is running out of LTS support soon: https://wiki.debian.org/LTS

Also, since I removed the bundled xxhash source code from borg 1.4, the build on buster broke, because libxxhash-dev (0.6.5) can not be discovered via `pkg-config`. This feature has been added later there (0.8 or so) and was initially broken, thus needs 0.8.2 or a fixed package.

So there might be no buster-based borg 1.4 binary (glibc 2.28), unless this is fixed / worked around somehow.

Debian Bullseye offers glibc 2.31, Debian Bookworm offers glibc 2.36.

Considering borg 1.4 is quite similar to borg 1.2, users of older systems can also use the borg 1.2 binaries.

ThomasWaldmann:


ThomasWaldmann:
Note: Already removed in master branch.

Issue #8208, opened 2024-05-10
Can't build borg on arm64 (armbian 22.04LTS)


m33m33:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

As best as I could

## Is this a BUG / ISSUE report or a QUESTION?

ISSUE: can't build from sources (either pip or git, this issue is about building from sources using git)

#### Your borg version (borg -V).

Looking forward to build the current version as per git clone

#### Operating system (distribution) and version.

**Armbian**, cat /etc/lsb-release :

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.4 LTS"

#### Hardware / network configuration, and filesystems used.

LePotato (arm64 SBC, mainline kernel support)

## Describe the problem you're observing.

Following instructions from https://borgbackup.readthedocs.io/en/stable/installation.html#git-installation with GIT method.
Everything looks good from here:
```
# get borg from github
git clone https://github.com/borgbackup/borg.git

Cloning into 'borg'...                                                                   
remote: Enumerating objects: 56539, done.
remote: Counting objects: 100% (1747/1747), done.                                                                                                                                 
remote: Compressing objects: 100% (730/730), done.                                                                                                                                remote: Total 56539 (delta 1018), reused 1592 (delta 891), pack-reused 54792                                                                                                      Receiving objects: 100% (56539/56539), 32.14 MiB | 3.56 MiB/s, done.
Resolving deltas: 100% (38240/38240), done.                                                                                                                                       Updating files: 100% (401/401), done.                                                                                                                                             

# create a virtual environment
virtualenv --python=$(which python3) borg-env
source borg-env/bin/activate   # always before using!

created virtual environment CPython3.10.12.final.0-64 in 7290ms                          
  creator CPython3Posix(dest=/.../src/borg-env, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/.../.local/share/virtualenv)
    added seed packages: pip==22.0.2, setuptools==59.6.0, wheel==0.37.1                  
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
(borg-env) user@host> 

# install borg + dependencies into virtualenv
cd borg
pip install -r requirements.d/development.txt
[...]

PR #8209, opened 2024-05-12
On Linux test /tmp exec capacity, cythonize needs it

m33m33:
Proposal for issue #8208

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import platform
 from collections import defaultdict
 from glob import glob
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -27,6 +28,20 @@
...
+        for line in mounts_file:
+            fields = line.split()
+            if len(fields) >= 5 and fields[1] == mount_point:
+                options = fields[3].split(',')
+                if 'noexec' in options:
+                    return True
+    return False
+
+if platform.system() == 'Linux' and is_noexec('/tmp'):
+    print("Warning: /tmp is mounted with noexec capacity, cythonize may fail")
```
ThomasWaldmann, line 43: That is problematic:
- it is rather platform specific and might still fail on other platforms due to similar reasons
- the temp dir could be set to something else than /tmp

m33m33: Agreed (I just don't know any better way to do it)

ThomasWaldmann:
Thanks for discovering this issue and debugging it!

borg hiding the original exception message was a real bug, based on the wrong assumption that only a missing cython install can trigger that import error.

But I guess I rather won't like to have code like in this PR in setup.py (borg supports misc. Linuxes, misc. BSDs, macOS, misc. OpenSolaris dists and even more or less Windows, so having platform specific code in setup.py catching such issues would really make it much bigger and much more complicated).

So, let's rather try to have a helpful error msg, see #8210.

PR #8210, opened 2024-05-12
setup.py: fix import error reporting for cythonize import, see #8208

ThomasWaldmann:
Looks like borg's setup.py has hidden the real cause of a cythonize ImportError.

There are basically 2 cases:
- either there is no Cython installed, then the import fails because the module can not be found, or
- there is some issue within Cython and the import fails due to that.

It's important not to hide the real cause, especially if we run into case 2.

case 1 is kind of expected and frequent, case 2 is rare.


ThomasWaldmann:
The macOS failure is due to:
```
  ld: library 'crypto' not found
```
Maybe some OpenSSL change in the action runner?

PR #8248, opened 2024-06-06
Fixed openSUSE dependencies in installation.rst

dertuxmalwieder:
Installing Borg from `pip` requires a few dependencies. Unfortunately, zypper can't find a few of those listed. Here's the fix.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -236,10 +236,10 @@ Install the dependencies automatically using zypper::
 Alternatively, you can enumerate all build dependencies in the command line::
 
     sudo zypper install python3 python3-devel \
-    libacl-devel openssl-devel libxxhash-devel \
+    libacl-devel openssl-devel xxhash-devel \
     python3-Cython python3-Sphinx python3-msgpack-python python3-pkgconfig pkgconf \
     python3-pytest python3-setuptools python3-setuptools_scm \
-    python3-sphinx_rtd_theme gcc gcc-c++
+    python3-sphinx_rtd_theme gcc gcc-c++ libzstd-devel liblz4-devel
     sudo zypper install python3-llfuse  # llfuse
 
 macOS
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -236,10 +236,10 @@ Install the dependencies automatically using zypper::
 Alternatively, you can enumerate all build dependencies in the command line::
 
     sudo zypper install python3 python3-devel \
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -236,10 +236,10 @@ Install the dependencies automatically using zypper::
 Alternatively, you can enumerate all build dependencies in the command line::
 
     sudo zypper install python3 python3-devel \
-    libacl-devel openssl-devel libxxhash-devel \
+    libacl-devel openssl-devel xxhash-devel \
     python3-Cython python3-Sphinx python3-msgpack-python python3-pkgconfig pkgconf \
     python3-pytest python3-setuptools python3-setuptools_scm \
-    python3-sphinx_rtd_theme gcc gcc-c++
+    python3-sphinx_rtd_theme gcc gcc-c++ libzstd-devel liblz4-devel
```
ThomasWaldmann, line 242: maybe rather move these compression libs/headers directly behind the xxhash-devel. borg directly depends on these, so they belong rather there.

docs/installation.rst
```diff
@@ -236,10 +236,10 @@ Install the dependencies automatically using zypper::
 Alternatively, you can enumerate all build dependencies in the command line::
 
     sudo zypper install python3 python3-devel \
-    libacl-devel openssl-devel libxxhash-devel \
+    libacl-devel openssl-devel xxhash-devel \
```
ThomasWaldmann, line 239: ah, good catch!

docs/installation.rst
dertuxmalwieder: Fixed

ThomasWaldmann: that was quick! :-)

ThomasWaldmann:
Thanks! Can you please do a similar PR also for 1.2-maint and 1.4-maint branches, which are used to create stable releases?

dertuxmalwieder:
Do these versions have other dependencies? I don’t have a second system to test them.

ThomasWaldmann:
1.4-maint should be similar as in master.

1.2-maint has some stuff bundled, but prefers the OS libs, if present and detectable.

dertuxmalwieder:
In that case, I probably can’t, given that the 2.0.0b8 package conflicts with 1.2 and 1.4…

ThomasWaldmann:
You can use pip and install into a virtual env, that's completely isolated and easy to remove again afterwards.

PR #8257, opened 2024-06-29
Automatic rebuild cache on exception, fixes #5213

### LINKED ISSUE #5213: no automatic rebuild of a deleted chunks file
borg does not tolerate a deleted chunks file (deleted because it was corrupted) due to the integrity checks:
```
u0_a151@localhost ~> ./borgbackup.sh

Fri Jun  5 21:11:53 EDT 2020 Starting backup

Killed stale lock localhost@75173968615802.3665-0.
Removed stale exclusive roster lock for host localhost@75173968615802 pid 3665 thread 0.
Removed stale exclusive roster lock for host localhost@75173968615802 pid 9194 thread 0.
Removed stale exclusive roster lock for host localhost@75173968615802 pid 3665 thread 0.
Removed stale exclusive roster lock for host localhost@75173968615802 pid 9194 thread 0.
Creating archive at "aelius@192.168.1.143:/media/borg-monolith::Monolith-2020-06-05T21:11:55"
Local Exception
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/archiver.py", line 4529, in main
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/archiver.py", line 4461, in run
    return set_ec(func(args))
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/archiver.py", line 166, in wrapper
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/archiver.py", line 565, in do_create
    with Cache(repository, key, manifest, do_files=args.cache_files, progress=args.progress,
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 380, in __new__
    return local()
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 372, in local
    return LocalCache(repository=repository, key=key, manifest=manifest, path=path, sync=sync,
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 457, in __init__
    self.open()
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 505, in open
    self.rollback()
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 610, in rollback
    self._do_open()
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/cache.py", line 493, in _do_open
    with IntegrityCheckedFile(path=os.path.join(self.path, 'chunks'), write=False,
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/borgbackup-1.1.11-py3.8-linux-x86_64.egg/borg/crypto/file_integrity.py", line 129, in __init__
    self.file_fd = override_fd or open(path, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/data/data/com.termux/files/home/.cache/borg/6d24295b115c5b4eb74bace9521ab7b09de7f82f096b8b39f8fcd1411fe4b34e/chunks'

Platform: Linux localhost 4.9.197-arter97-r12-g29a968fb09d85 #1 SMP PREEMPT Mon Oct 28 19:22:17 KST 2019 aarch64
Linux: Unknown Linux
Borg: 1.1.11  Python: CPython 3.8.3 msgpack: 0.5.6
PID: 11175  CWD: /data/data/com.termux/files/home
sys.argv: ['/data/data/com.termux/files/usr/bin/borg', 'create', '--verbose', '--filter', 'AME', '--list', '--stats', '--show-rc', '--compression', 'lz4', '--exclude-caches', '--pattern=+/storage/emulated/0/Android/data/org.tasks', '--pattern=+/storage/emulated/0/Android/data/net.sourceforge.opencamera', '--pattern=-/storage/emulated/0/Android', '--pattern=-/storage/emulated/0/DCIM', '--exclude', '*/.sync', '::Monolith-{now}', 'borgbackup.sh', '/storage/emulated/0/']
SSH_ORIGINAL_COMMAND: None

terminating with error status, rc 2
```

Aztorius:
Here is a fix for #5213 by calling wipe_cache() if an exception is raised on LocalCache.open().

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -531,7 +531,12 @@ class LocalCache(CacheStatsMixin):
             self.security_manager.assert_access_unknown(warn_if_unencrypted, manifest, self.key)
             self.create()
 
-        self.open()
+        try:
+            self.open()
+        except:
+            self.wipe_cache()
+            self.open()
+
         try:
             self.security_manager.assert_secure(manifest, self.key, cache_config=self.cache_config)
 
@@ -924,7 +929,7 @@ class LocalCache(CacheStatsMixin):
         return True
 
     def wipe_cache(self):
-        logger.warning("Discarding incompatible cache and forcing a cache rebuild")
+        logger.warning("Discarding incompatible or corrupted cache and forcing a cache rebuild")
         archive_path = os.path.join(self.path, "chunks.archive.d")
         if os.path.isdir(archive_path):
             shutil.rmtree(os.path.join(self.path, "chunks.archive.d"))
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -940,7 +940,7 @@ class LocalCache(CacheStatsMixin):
         self.cache_config.integrity["chunks"] = fd.integrity_data
         with IntegrityCheckedFile(path=os.path.join(self.path, files_cache_name()), write=True) as fd:
             pass  # empty file
-        self.cache_config.integrity["files"] = fd.integrity_data
+        self.cache_config.integrity[files_cache_name()] = fd.integrity_data
         self.cache_config.manifest_id = ""
         self.cache_config._config.set("cache", "manifest", "")
         if not self.cache_config._config.has_section("integrity"):
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -533,7 +533,7 @@ class LocalCache(CacheStatsMixin):
 
         try:
             self.open()
-        except:
+        except (FileNotFoundError, FileIntegrityError):
             self.wipe_cache()
             self.open()
 
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -947,6 +947,7 @@ class LocalCache(CacheStatsMixin):
             self.cache_config._config.add_section("integrity")
         for file, integrity_data in self.cache_config.integrity.items():
             self.cache_config._config.set("integrity", file, integrity_data)
+        # This is needed to pass the integrity check later on inside CacheConfig.load()
         self.cache_config._config.set("integrity", "manifest", "")
 
         self.cache_config.ignored_features = set()
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -943,10 +943,16 @@ class LocalCache(CacheStatsMixin):
         self.cache_config.integrity["files"] = fd.integrity_data
         self.cache_config.manifest_id = ""
         self.cache_config._config.set("cache", "manifest", "")
+        if not self.cache_config._config.has_section("integrity"):
+            self.cache_config._config.add_section("integrity")
+        for file, integrity_data in self.cache_config.integrity.items():
+            self.cache_config._config.set("integrity", file, integrity_data)
+        self.cache_config._config.set("integrity", "manifest", "")
 
         self.cache_config.ignored_features = set()
         self.cache_config.mandatory_features = set()
-        self.cache_config.save(self.manifest, self.key)
+        with SaveFile(self.cache_config.config_path) as fd:
+            self.cache_config._config.write(fd)
 
     def update_compatibility(self):
         operation_to_features_map = self.manifest.get_all_mandatory_features()
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -930,13 +930,18 @@ class LocalCache(CacheStatsMixin):
             shutil.rmtree(os.path.join(self.path, "chunks.archive.d"))
             os.makedirs(os.path.join(self.path, "chunks.archive.d"))
         self.chunks = ChunkIndex()
-        with SaveFile(os.path.join(self.path, files_cache_name()), binary=True):
+        with IntegrityCheckedFile(path=os.path.join(self.path, "chunks"), write=True) as fd:
+            self.chunks.write(fd)
+        self.cache_config.integrity["chunks"] = fd.integrity_data
+        with IntegrityCheckedFile(path=os.path.join(self.path, files_cache_name()), write=True) as fd:
             pass  # empty file
+        self.cache_config.integrity["files"] = fd.integrity_data
         self.cache_config.manifest_id = ""
         self.cache_config._config.set("cache", "manifest", "")
 
         self.cache_config.ignored_features = set()
         self.cache_config.mandatory_features = set()
+        self.cache_config.save(self.manifest, self.key)
 
     def update_compatibility(self):
         operation_to_features_map = self.manifest.get_all_mandatory_features()
--- a/src/borg/testsuite/archiver/corruption.py
+++ b/src/borg/testsuite/archiver/corruption.py
@@ -12,6 +12,7 @@ from . import cmd, create_src_archive, create_test_files, RK_ENCRYPTION
 from ...hashindex import ChunkIndex
 from ...cache import LocalCache
 
+
 def test_check_corrupted_repository(archiver):
     cmd(archiver, "rcreate", RK_ENCRYPTION)
     create_src_archive(archiver, "test")
@@ -52,12 +53,16 @@ def test_cache_chunks(archiver):
     chunks_before_corruption = set(ChunkIndex(path=chunks_path).iteritems())
     corrupt(chunks_path)
 
+    assert not archiver.FORK_DEFAULT  # test does not support forking
+
     chunks_in_memory = None
     sync_chunks = LocalCache.sync
-    def sync_wrapper(*args):
+
+    def sync_wrapper(cache):
         nonlocal chunks_in_memory
-        sync_chunks(*args)
-        chunks_in_memory = set(args[0].chunks.iteritems())
+        sync_chunks(cache)
+        chunks_in_memory = set(cache.chunks.iteritems())
+
     with patch.object(LocalCache, "sync", sync_wrapper):
         out = cmd(archiver, "rinfo")
 
@@ -65,6 +70,7 @@ def test_cache_chunks(archiver):
     assert "forcing a cache rebuild" in out
     chunks_after_repair = set(ChunkIndex(path=chunks_path).iteritems())
     assert chunks_after_repair == chunks_before_corruption
+
 
 def test_cache_files(archiver):
     corrupt_archiver(archiver)
--- a/src/borg/testsuite/archiver/corruption.py
+++ b/src/borg/testsuite/archiver/corruption.py
@@ -8,7 +8,7 @@ import pytest
 from ...constants import *  # NOQA
 from ...helpers import bin_to_hex, Error
 from . import cmd, create_src_archive, create_test_files, RK_ENCRYPTION
-
+from ...hashindex import ChunkIndex
 
 def test_check_corrupted_repository(archiver):
     cmd(archiver, "rcreate", RK_ENCRYPTION)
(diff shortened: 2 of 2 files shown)


src/borg/cache.py
```diff
@@ -531,7 +531,12 @@ def __init__(
             self.security_manager.assert_access_unknown(warn_if_unencrypted, manifest, self.key)
             self.create()
 
-        self.open()
+        try:
+            self.open()
+        except:
```
ThomasWaldmann, line 536: Hmm, would it make sense to make this more specific?

A bare `except:` can often cause problems, especially if it is around a lot of code.

src/borg/cache.py
```diff
@@ -924,19 +929,24 @@ def check_cache_compatibility(self):
...
-        logger.warning("Discarding incompatible cache and forcing a cache rebuild")
+        logger.warning("Discarding incompatible or corrupted cache and forcing a cache rebuild")
         archive_path = os.path.join(self.path, "chunks.archive.d")
         if os.path.isdir(archive_path):
             shutil.rmtree(os.path.join(self.path, "chunks.archive.d"))
             os.makedirs(os.path.join(self.path, "chunks.archive.d"))
         self.chunks = ChunkIndex()
-        with SaveFile(os.path.join(self.path, files_cache_name()), binary=True):
+        with IntegrityCheckedFile(path=os.path.join(self.path, "chunks"), write=True) as fd:
+            self.chunks.write(fd)
```
ThomasWaldmann, line 939: What exactly is it saving to disk here? What's the contents of the chunks hashtable?

src/borg/cache.py
Aztorius: Yes we could only catch FileNotFoundError and FileIntegrityError for now

src/borg/cache.py
Aztorius: The chunks hashtable saved to disk is a valid empty ChunkIndex here thanks to the previous line `self.chunks = ChunkIndex()`
This is similar to the LocalCache.create() function behavior. IntegrityCheckedFile doesn't interfere and just allows us to pass the integrity check when calling open() later.
Does it answer your question ?

src/borg/cache.py
ThomasWaldmann: Yeah, that would be better!

If we find other exceptions fixable by the except block, we can add them later.

src/borg/cache.py
ThomasWaldmann: Can you check if it does a chunks cache resync?

If borg continues and does not exit there, it must do that before working with that (empty) chunks hashtable. If it exits, it needs to do the resync the next time borg is used.

Aztorius: Just checked. The resync was not triggered because of CacheConfig.save() that overrides the manifest_id with the provided manifest.
I just pushed a commit where the config is written as-is with an empty manifest_id. This correctly triggers a call to sync() in  LocalCache.__init__()

src/borg/cache.py
```diff
@@ -943,10 +943,16 @@ def wipe_cache(self):
         self.cache_config.integrity["files"] = fd.integrity_data
         self.cache_config.manifest_id = ""
         self.cache_config._config.set("cache", "manifest", "")
+        if not self.cache_config._config.has_section("integrity"):
+            self.cache_config._config.add_section("integrity")
+        for file, integrity_data in self.cache_config.integrity.items():
+            self.cache_config._config.set("integrity", file, integrity_data)
+        self.cache_config._config.set("integrity", "manifest", "")
```
ThomasWaldmann, line 950: Please add a comment above this line explaining why you do that.

src/borg/cache.py
ThomasWaldmann: So it resyncs the chunks cache in same borg run before using the chunks cache?

src/borg/cache.py
ThomasWaldmann: (so one does not need to search in git history to find that)

src/borg/cache.py
```diff
@@ -924,19 +929,30 @@ def check_cache_compatibility(self):
...
             shutil.rmtree(os.path.join(self.path, "chunks.archive.d"))
             os.makedirs(os.path.join(self.path, "chunks.archive.d"))
         self.chunks = ChunkIndex()
-        with SaveFile(os.path.join(self.path, files_cache_name()), binary=True):
+        with IntegrityCheckedFile(path=os.path.join(self.path, "chunks"), write=True) as fd:
+            self.chunks.write(fd)
+        self.cache_config.integrity["chunks"] = fd.integrity_data
+        with IntegrityCheckedFile(path=os.path.join(self.path, files_cache_name()), write=True) as fd:
             pass  # empty file
+        self.cache_config.integrity["files"] = fd.integrity_data
```
ThomasWaldmann, line 943: hmm, hardcoded "files" might be incorrect, see `files_cache_name()`.

ThomasWaldmann: The other place updating this uses `...integrity[files_cache_name()]`.

Aztorius: Correct

src/borg/cache.py
ThomasWaldmann: I fixed the failing test and pushed the commit. Then I slighlty modified the expected string to make it fail again, so I get to see the output:

E       AssertionError: assert 'xxxxx forcing a cache rebuild' in """
Discarding incompatible or corrupted cache and forcing a cache rebuild
Repository ID: caaff3f127e80db9d19a24762e4f21c7870c3bc7b6066bdb37f9c67771a80f3c
Location: /private/var/folders/61/qtqwzmxd7t1fb19zck5k3q0h0000gn/T/pytest-of-tw/pytest-88/test_cache_chunks0/repository
Repository version: 2
Append only: False
Encrypted: Yes (repokey AES-OCB)
Storage quota: 460 B used
Cache: /private/var/folders/61/qtqwzmxd7t1fb19zck5k3q0h0000gn/T/pytest-of-tw/pytest-88/test_cache_chunks0/cache/caaff3f127e80db9d19a24762e4f21c7870c3bc7b6066bdb37f9c67771a80f3c
Security dir: /private/var/folders/61/qtqwzmxd7t1fb19zck5k3q0h0000gn/T/pytest-of-tw/pytest-88/borg-base-dir0/.local/share/borg/security/caaff3f127e80db9d19a24762e4f21c7870c3bc7b6066bdb37f9c67771a80f3c
Original size: 0 B
Deduplicated size: 0 B
Unique chunks: 0
Total chunks: 0
"""

The last few lines seem to indicate that it **does not** do a chunks cache resync yet, but happily proceeds with the empty chunks cache.

src/borg/cache.py
Aztorius: Added

src/borg/cache.py
Aztorius: Yes on the same run.

Aztorius: Let me check again

Aztorius: It seems that the test do not create an archive so there is nothing in the cache. If we do a create_src_archive(archiver, "test") for example inside the test_cache_chunks() we can clearly see that the cache is not empty and correctly synced after corruption.

ThomasWaldmann: That might be, but borg needs to work correctly for all commands.

Commands must not work with a "fake" chunks cache that is out of sync with repo reality.

Aztorius: Yes I agree. On all my testing it works for any commands (rinfo, create, ...)

If I remove the "chunks" corruption in test_cache_chunks() I get this result : 
```
Repository ID: 610198ddd42c7a3ac2160ae0a965fdbff2b11a199440e717d24a3dd1c8d342b3
Location: /tmp/pytest-of-aztorius/pytest-13/test_cache_chunks0/repository
Repository version: 2
Append only: False
Encrypted: Yes (repokey AES-OCB)
Storage quota: 460 B used
Cache: /tmp/pytest-of-aztorius/pytest-13/test_cache_chunks0/cache/610198ddd42c7a3ac2160ae0a965fdbff2b11a199440e717d24a3dd1c8d342b3
Security dir: /tmp/pytest-of-aztorius/pytest-13/borg-base-dir1/.local/share/borg/security/610198ddd42c7a3ac2160ae0a965fdbff2b11a199440e717d24a3dd1c8d342b3
Original size: 0 B
Deduplicated size: 0 B
Unique chunks: 0
Total chunks: 0
```
So the cache is as expected. I do not see any issue.

ThomasWaldmann: Ah, ok, so that test is not yet suitable to test for the cache sync (as one can't see the difference) - can you modify it so the repo isn't empty when the cache gets corrupted?

Aztorius: I just pushed a commit to create an archive before corruption.
Maybe we should add an assert on the chunks file content to effectively check if everything is in sync ?

ThomasWaldmann: Guess making a "copy" of the chunks hashtable before the corruption and then comparing all keys/value after the chunks cache resync has happened would be best.

We could also check if the resync happens (iirc it tells that on the output).

Aztorius: New commit pushed with a check on chunks hashtable content.

src/borg/testsuite/archiver/corruption.py
```diff
@@ -46,13 +46,16 @@ def corrupt(file, amount=1):
...
+    chunks_before_corruption = set(ChunkIndex(path=chunks_path).iteritems())
+    corrupt(chunks_path)
     if archiver.FORK_DEFAULT:
         out = cmd(archiver, "rinfo")
     else:
         out = cmd(archiver, "rinfo")
     assert "forcing a cache rebuild" in out
-
+    chunks_after_repair = set(ChunkIndex(path=chunks_path).iteritems())
+    assert chunks_after_repair == chunks_before_corruption
```
ThomasWaldmann, line 58: Looks good as far as the on-disk index is concerned.

Not sure about the in-memory index that is actually used in the borg invocation when the rebuild happens.

Aztorius: New commit with the in-memory index checked (inspired by other test files)

src/borg/testsuite/archiver/corruption.py
```diff
@@ -49,10 +51,17 @@ def test_cache_chunks(archiver):
     chunks_path = os.path.join(archiver.cache_path, "chunks")
     chunks_before_corruption = set(ChunkIndex(path=chunks_path).iteritems())
     corrupt(chunks_path)
-    if archiver.FORK_DEFAULT:
-        out = cmd(archiver, "rinfo")
-    else:
```
ThomasWaldmann, line 52: Can you replace this with:
```
assert not archiver.FORK_DEFAULT  # test does not support forking
```

src/borg/testsuite/archiver/corruption.py
```diff
@@ -49,10 +51,17 @@ def test_cache_chunks(archiver):
...
+
+    chunks_in_memory = None
+    sync_chunks = LocalCache.sync
+    def sync_wrapper(*args):
```
ThomasWaldmann, line 57: there is only 1 method arg: `self`

so maybe just call it `cache` here?

Aztorius: ok

src/borg/testsuite/archiver/corruption.py
Aztorius: done

src/borg/cache.py
ThomasWaldmann: looks good.

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
Thanks!

ThomasWaldmann:
Could you do a backport to 1.4-maint so it gets into 1.4.1 release?

Aztorius:
> Could you do a backport to 1.4-maint so it gets into 1.4.1 release?
> 
> BTW, I squashed the commits into a single one, so it is just than one commit to cherry-pick for 1.4-maint.

I will look into that.

PR #8260, opened 2024-07-02
Add repair count to borg check --repair output

supercoder-dev:
Update 'borg check --repair' to include the number of problems repaired in the final output message. This provides users with clearer information about the repairs performed during the consistency check.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1870,9 +1870,14 @@ class ArchiveChecker:
             match=match, first=first, last=last, sort_by=sort_by, older=older, oldest=oldest, newer=newer, newest=newest
         )
         self.orphan_chunks_check()
+        self.repair_count = 0
+        self.orphan_chunks_check()
         self.finish()
         if self.error_found:
-            logger.error("Archive consistency check complete, problems found.")
+            if self.repair_count > 0:
+                logger.error(f"Archive consistency check complete, problems found and {self.repair_count} problem(s) repaired.")
+            else:
+                logger.error("Archive consistency check complete, problems found.")
         else:
             logger.info("Archive consistency check complete, no problems found.")
         return self.repair or not self.error_found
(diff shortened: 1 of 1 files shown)


src/borg/archive.py
```diff
@@ -1870,9 +1870,14 @@ def check(
             match=match, first=first, last=last, sort_by=sort_by, older=older, oldest=oldest, newer=newer, newest=newest
         )
         self.orphan_chunks_check()
+        self.repair_count = 0
+        self.orphan_chunks_check()
```
ThomasWaldmann, line 1874: Looks like you accidentally duplicated the call to orphan_chunks_check.

src/borg/archive.py
```diff
@@ -2338,7 +2343,9 @@ def orphan_chunks_check(self):
                 )
                 for id_ in unused:
                     self.repository.delete(id_)
+                    self.repair_count += 1
```
ThomasWaldmann, line 2346: Guess it is *least* important to count the fixed orphaned objects - while they should not occur in the ideal case, there can be conditions producing them.

But, orphans (chunks that are not referenced by any archive) are only a cosmetic issue and we could also just silently remove them (that's why the log message is on INFO level, not WARNING or ERROR).

It would be more important to count the severe "repaired" issues, e.g. if a referenced chunk is missing from the repo and was replaced by an all-zero replacement chunk. That's data loss and the repair is only bringing back consistency, but it's not bringing back the data.

If such a lost chunk reappears later, borg check can also **heal** it, that's a real repair then.

ThomasWaldmann (changes requested):
Thanks for your PR! See my comments...

ThomasWaldmann:
@supercoder-dev Did you see my feedback?

ThomasWaldmann:
Ping?

ThomasWaldmann:
no reaction...

PR #8272, opened 2024-07-05
docs: fix typo

wzyboy:


src/borg/archiver/config_cmd.py
```diff
@@ -146,7 +146,7 @@ def build_parser_config(self, subparsers, common_parser, mid_common_parser):
         For security reasons, this command only works on local repositories.
 
         To delete a config value entirely, use ``--delete``. To list the values
-        of the configuration file or the default values, use ``--list``.  To get and existing
```
ThomasWaldmann, line 149: Thanks for the PR!

As documented in the first lines of borg-config.1 and config.rst.inc, these 2 files are generated.

Thus, config_cmd.py is the only file that needs to be changed to fix this typo. Can you remove the other files from this PR, please?

wzyboy: Done!

ThomasWaldmann:
Can you please check whether the same typo is in 1.2-maint and 1.4-maint and if so, backport the fix to these branches? Likely, you only need to change the `borg/archiver.py` file there.

ThomasWaldmann:
Thanks!

PR #8281, opened 2024-07-13
Add BORG_USE_CHUNKS_ARCHIVE env var, fixes #8280

### LINKED ISSUE #8280: Disable cache with environment variable/CLI option
This would be easier than the workaround listed in the FAQ till #474 is done.

Aztorius:
Fixes #8280.

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -521,6 +521,7 @@ class LocalCache(CacheStatsMixin):
         self.cache_mode = cache_mode
         self.timestamp = None
         self.txn_active = False
+        self.do_cache = os.environ.get("BORG_USE_CHUNKS_ARCHIVE", "yes").lower() in ["yes", "1", "true"]
 
         self.path = cache_dir(self.repository, path)
         self.security_manager = SecurityManager(self.repository)
@@ -910,10 +911,6 @@ class LocalCache(CacheStatsMixin):
 
         self.begin_txn()
         with cache_if_remote(self.repository, decrypted_cache=self.repo_objs) as decrypted_repository:
-            # TEMPORARY HACK:
-            # to avoid archive index caching, create a FILE named ~/.cache/borg/REPOID/chunks.archive.d -
-            # this is only recommended if you have a fast, low latency connection to your repo (e.g. if repo is local).
-            self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
     def check_cache_compatibility(self):
--- a/src/borg/testsuite/archiver/checks.py
+++ b/src/borg/testsuite/archiver/checks.py
@@ -319,6 +319,23 @@ def test_check_cache(archivers, request):
         check_cache(archiver)
 
 
+def test_env_use_chunks_archive(archivers, request, monkeypatch):
+    archiver = request.getfixturevalue(archivers)
+    create_test_files(archiver.input_path)
+    monkeypatch.setenv("BORG_USE_CHUNKS_ARCHIVE", "no")
+    cmd(archiver, "rcreate", RK_ENCRYPTION)
+    repository_id = bin_to_hex(_extract_repository_id(archiver.repository_path))
+    cache_path = os.path.join(archiver.cache_path, repository_id)
+    cmd(archiver, "create", "test", "input")
+    assert os.path.exists(cache_path)
+    assert os.path.exists(os.path.join(cache_path, "chunks.archive.d"))
+    assert len(os.listdir(os.path.join(cache_path, "chunks.archive.d"))) == 0
+    cmd(archiver, "rdelete", "--cache-only")
+    monkeypatch.setenv("BORG_USE_CHUNKS_ARCHIVE", "yes")
+    cmd(archiver, "create", "test2", "input")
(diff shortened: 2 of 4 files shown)


ThomasWaldmann:
Thanks for working on this!

I think an environment variable like `BORG_USE_CHUNKS_ARCHIVE=yes` would be easier than adding cli options all over the place.

Aztorius:
> I think an environment variable like `BORG_USE_CHUNKS_ARCHIVE=yes` would be easier than adding cli options all over the place.

Ok, I removed the --no-cache and now there is only the environment variable.

src/borg/testsuite/archiver/checks.py
```diff
@@ -319,6 +319,23 @@ def test_check_cache(archivers, request):
         check_cache(archiver)
 
 
+def test_env_use_chunks_archive(archivers, request):
+    archiver = request.getfixturevalue(archivers)
+    create_test_files(archiver.input_path)
+    os.environ["BORG_USE_CHUNKS_ARCHIVE"] = "no"
```
ThomasWaldmann, line 325: Better use this:
```
monkeypatch.setenv("BORG_USE_CHUNKS_ARCHIVE", "no")
```

src/borg/testsuite/archiver/checks.py
```diff
@@ -319,6 +319,23 @@ def test_check_cache(archivers, request):
...
+    cmd(archiver, "rcreate", RK_ENCRYPTION)
+    repository_id = bin_to_hex(_extract_repository_id(archiver.repository_path))
+    cache_path = os.path.join(archiver.cache_path, repository_id)
+    cmd(archiver, "create", "test", "input")
+    assert os.path.exists(cache_path)
+    assert os.path.exists(os.path.join(cache_path, "chunks.archive.d"))
+    assert len(os.listdir(os.path.join(cache_path, "chunks.archive.d"))) == 0
+    cmd(archiver, "rdelete", "--cache-only")
+    os.environ["BORG_USE_CHUNKS_ARCHIVE"] = "yes"
```
ThomasWaldmann, line 334: same here.

src/borg/testsuite/archiver/checks.py
```diff
@@ -319,6 +319,23 @@ def test_check_cache(archivers, request):
...
+    cmd(archiver, "create", "test2", "input")
+    assert len(os.listdir(os.path.join(cache_path, "chunks.archive.d"))) > 0
```
ThomasWaldmann, line 336: Should be `== 1`, right?

ThomasWaldmann (changes requested):
Looks good.

The docs will also need an update.

Search for BORG_ in *.rst* to find the env vars docs.

Search for chunks.archives.d to remove the old hack from the FAQ.

src/borg/testsuite/archiver/checks.py
Aztorius: 2 because of the integrity file. I can replace by an equality. My goal here was just to be sure that something has been written.

src/borg/testsuite/archiver/checks.py
Aztorius: Done

src/borg/testsuite/archiver/checks.py
Aztorius: Done

src/borg/cache.py
```diff
@@ -521,6 +521,10 @@ def __init__(
         self.cache_mode = cache_mode
         self.timestamp = None
         self.txn_active = False
+        self.do_cache = os.environ.get("BORG_USE_CHUNKS_ARCHIVE", "yes").lower() in ["yes", "1", "true"]
+
+        if not self.do_cache:
+            logger.info("Cache for chunks archive is disabled")
```
ThomasWaldmann, line 527: Maybe we better leave that away.
The admin should know well if they set that env var.

Output can sometimes completely disturb some functionality, e.g. if the normal output is some piece of json and suddenly there is that log message in front of it.

Aztorius: I will remove it, it is not really necessary.

Does borg sends json to stdout and logs to stderr ? That way logs should never interfere with data.

docs/man/borg.1
```diff
@@ -460,6 +460,10 @@ When set to a numeric value, this determines the maximum \(dqtime to live\(dq fo
 entries (default: 20). The files cache is used to determine quickly whether a file is unchanged.
 The FAQ explains this more detailed in: \fIalways_chunking\fP
 .TP
+.B BORG_USE_CHUNKS_ARCHIVE
+When set to no (default: yes), the chunks.archive.d folder will not be used. This reduces
+disk space usage but slows down cache resyncs.
+.TP
```
ThomasWaldmann, line 466: You don't need to update that file, it will be generated from the docs when I do a release.

src/borg/cache.py
ThomasWaldmann: I've seen test failing due to such stuff. Guess at least some test capture stderr and stdout together into 1 string.

ThomasWaldmann (approved):
LGTM.

Will merge after tests finish.

Could you please do a backport of this to 1.4-maint branch after that?

ThomasWaldmann:
Thanks!

PR #8282, opened 2024-07-14
Backport BORG_USE_CHUNKS_ARCHIVE env var to 1.4-maint

### LINKED ISSUE #8280: Disable cache with environment variable/CLI option

Aztorius:
This is the backport of the pull request #8281 to fix #8280 on 1.4-maint

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -478,6 +478,7 @@ class LocalCache(CacheStatsMixin):
         self.consider_part_files = consider_part_files
         self.timestamp = None
         self.txn_active = False
+        self.do_cache = os.environ.get("BORG_USE_CHUNKS_ARCHIVE", "yes").lower() in ["yes", "1", "true"]
 
         self.path = cache_dir(repository, path)
         self.security_manager = SecurityManager(repository)
@@ -907,9 +908,6 @@ class LocalCache(CacheStatsMixin):
         self.begin_txn()
         with cache_if_remote(self.repository, decrypted_cache=self.key) as decrypted_repository:
             legacy_cleanup()
-            # TEMPORARY HACK: to avoid archive index caching, create a FILE named ~/.cache/borg/REPOID/chunks.archive.d -
-            # this is only recommended if you have a fast, low latency connection to your repo (e.g. if repo is local disk)
-            self.do_cache = os.path.isdir(archive_path)
             self.chunks = create_master_idx(self.chunks)
 
     def check_cache_compatibility(self):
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Thanks!

src/borg/cache.py
```diff
@@ -907,9 +908,6 @@ def legacy_cleanup():
         self.begin_txn()
         with cache_if_remote(self.repository, decrypted_cache=self.key) as decrypted_repository:
             legacy_cleanup()
-            # TEMPORARY HACK: to avoid archive index caching, create a FILE named ~/.cache/borg/REPOID/chunks.archive.d -
-            # this is only recommended if you have a fast, low latency connection to your repo (e.g. if repo is local disk)
-            self.do_cache = os.path.isdir(archive_path)
```
ThomasWaldmann, line 912: Hmm, I just noted there is a slight problem with removing that:

If users have use the old hack, likely that will break in unpretty ways when they upgrade to the new code.

For master branch, this is not a problem, because it is a breaking release anyway and people will have to migrate to new repos anyway.

For 1.4-maint, this might affect some users upgrading from 1.2.x or 1.4.0 though.

Aztorius: Yes, we could simply check if it is not a directory then remove it and create the directory.

Aztorius: Do you want me to add it in another pull request ?

ThomasWaldmann: Guess we need to check like this in 1.4-maint to not break working setups:

```
# new way: tell whether to use the chunks archive via an environment variable:
do_cache_env = os.environ.get("BORG_USE_CHUNKS_ARCHIVE", "yes").lower() in ["yes", "1", "true"]
# backwards compatibility: a user might have replaced the archive_path directory with
# a regular file (as documented in the borg FAQ) to avoid chunks archive creation:
do_cache_fs = os.path.isdir(archive_path)
self.do_cache = do_cache_env and do_cache_fs
```

ThomasWaldmann: And yeah, a new PR would be welcome!

jdchristensen: What happens in 2.0 if the user has a file instead of a directory there?  I'd suggest that borg exits in that case with a message pointing to the new environment variable, so people know to upgrade their setup.  I don't think silently removing the file and creating a directory is a good idea, nor do I think a random error with a traceback would be good.

ThomasWaldmann: borg2 deals with 2 kinds of repos:
- borg 1.x repos, but only once to transfer archives from (== read archives). I guess we can assume that the chunks index is in sync from prior borg 1.x usage. Not even sure if it uses the chunks index just for reading.
- borg 2 repos are freshly created anyway (different format) and won't support the old "let's put a file there" hack.

PR #8283, opened 2024-07-14
Fix backwards compatibility of old chunks.archive.d file behavior

Aztorius:
Related to #8282 

In case a file named chunks.archive.d, cache would not create a folder with that name to store cache chunks archive. We need to support both the new environment variable and the old "hacky" way of creating a file.


ThomasWaldmann:
Thanks!

PR #8291, opened 2024-07-19
key export fixes (master)

ThomasWaldmann:


ThomasWaldmann:
Note: the windows specific "fix" that caused the crash if no PATH is given was only done in master branch.

PR #8314, opened 2024-08-01
docs: align /etc/backups path references in automated backups deployment guide

ruben-rodriguez:
Hey there!

This small documentation fix aligns all `/etc/backups` path references in the Automated Backups to a Local Hard Drive deployment guide. Previously, there were some references to `/etc/backup` instead of `/etc/backups`, which caused confusion.


ThomasWaldmann:
Good catch, thanks for the PR!

ThomasWaldmann:
@ruben-rodriguez can you please check if 1.4-maint needs the same fix and if so, backport it?

ruben-rodriguez:
Fixed 1.4-maint in #8321 !

PR #8321, opened 2024-08-04
docs: align /etc/backups path references in automated backups deployment guide (1.4-maint)

ruben-rodriguez:
Backport (cherrypick) from master to 1.4-maint of #8314 

This small documentation fix aligns all /etc/backups path references in the Automated Backups to a Local Hard Drive deployment guide. Previously, there were some references to /etc/backup instead of /etc/backups, which caused confusion.


ThomasWaldmann:
Thanks!

Issue #8318, opened 2024-08-02
include stricter shell as recommended ssh configuration


qyanu:
Hello!

## Context

i currently use borg 1.x on a debian bookworm system, where i (also) want the server to not trust the clients that much and said clients connect to the server (and send data) over the network via ssh. This is as described in (one of) the configurations recommended by borg in its documentation.

Documentation reference: https://github.com/borgbackup/borg/blob/6039349932a874fbb6d053b1dd5d7d400780378b/docs/usage/serve.rst?plain=1#L47

I create this issue against the 2.x branch, but it both concerns the 1.x and 2.x documentation. I couldn't find a previous issue discussing (and possibly rejecting) my proposal through the github issue search, but i am not very confident that i didn't miss a related issue.

This issue proposes an improvement to the documentation, described just now. Further down, a longer rationale is given for the change.

## Proposed Improvement

In the documentation, in the section `Usage > borg serve > Example` and also the following section `SSH Configuration`,
the text should also mention that setting a shell with smaller footprint for the user under which `borg serve` is started, will reduce attack surface.

* recommend a `backup` user
* recommend `chsh -s /bin/sh backup`

## Rationale

I suppose that many users expect their backup-strategy to also hold (i.e. provide a means of disaster-recovery) with a reasonable certainty in the face of a "hacked" borg-client (e.g. a user's daily workstation); this is certainly a goal with my backup-strategy. When borg  is chosen as a building block for this backup-strategy, and borg is setup on a separate backup-server according to the description presented in the documentation section "Usage > borg serve", then i think this setup is needlessly in risk of shell-bugs similar to shellshock (CVE-2014-6271 and others, see: https://en.wikipedia.org/wiki/Shellshock_(software_bug) ).

<details><summary>Note</summary>
It must be noted, that all this discussion is strictly speaking outside the scope of borg. However, since borg elects to giving ssh-related configuration directions in it's documentation, it takes on (some) responsibility of recommending a secure-by-default configuration.
</details> 

In the recommended setup, the forced command (specified in the authorized_keys file) is executed by sshd by using `$SHELL -c $COMMAND`, where $SHELL is the user's login shell, and $COMMAND is the forced command (ref: https://manpages.debian.org/bookworm/openssh-server/sshd_config.5.en.html#ForceCommand ).
[...]

PR #8331, opened 2024-08-11
docs: borg-serve: simplify example of env in authorized_keys

qyanu:
see #8318

so long as it can be assumed that the user has configured a POSIX compliant login shell, using a simple command [1] looks cleaner, as no ``export`` or ``;`` are used.

[1] Section "2.9.1 Simple Commands" in volume "Shell & Utilities" of POSIX.1-2024

### DIFF
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -29,10 +29,12 @@ locations like ``/etc/environment`` or in the forced command itself (example bel
 
     # Set a BORG_XXX environment variable on the "borg serve" side
     $ cat ~/.ssh/authorized_keys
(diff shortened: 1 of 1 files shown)


qyanu:
While thinking about issue https://github.com/borgbackup/borg/issues/8318 i noticed, that the recommended authorized_keys command to set a BORG_XXX environment variable can do without the ";" and "export". so i just quickly created this pull request.

ThomasWaldmann:
Looks good IMHO/AFAIK.

I guess we'll keep this open for a while in case somebody has objections.

ThomasWaldmann:
Thanks!

Issue #5324, opened 2020-09-04
rclone Integration


buengese:
Hi,
I'm one of developers of [rclone](https://github.com/rclone/rclone) a command line tool written in golang to interact with various cloud storage providers. I'm wondering if the team here is interested in some sort of integration with rclone. 
We have a number of users that users that use borg with a cloud storage provider mounted via rclone with varying levels of success. Even though rclone's vfs has gotten much mature over time there is still significant overhead associated with it so a more direct would be beneficial. Currently rclone already has integration with restic that is achieved by a custom http2 api we are running on stdio. If only had a quick look at borg's remote repository protocol and not yet sure if it would be viable to implement something based on this on rclone's side.

Given the rather obvious absence of any cloud storage integration on borg's side, which may have been a conscious decision, I'd like to know if this something you would be in principle interested in before looking into this any further.
buengese

ThomasWaldmann:
I guess it would be interesting if it could be done without much changes to the existing `Repository` and `RemoteRepository` classes, e.g. by adding yet another one and triggering its use via repo url.

I can help if there are questions about borg internals (that are not already answered in the internals docs), but I don't use cloud storage myself.

buengese:
Thanks for the info. I'm going to look into how far this possible with minimal changes to the current `RemoteRepository` setup.

dragetd:
Some thoughts I want to drop here, without being deep in the code base on either project. Please excuse if I misunderstanding concepts - if this is the case, I will delete my post. :) Background: Long time borg an rclone user, currently using it via rclone fuse/mount, including all the pain. :P


borg has a `serve` mode where it basically takes commands via SSH and acts as the 'storage engine'. Having code at the receiving end allows to do some operations on the server-side without roundtrips to the client. This of course is not possible with a typical cloud storage.

So one would not be able to use this architecture and would need to hook only into the storage layers. Which prevents some optimizations, but that is a logical consequence.

As far as I know, there is no big abstraction in the storage/file-io code in general in borg. Given the huge amount of possible options rclone has, I think a tight integration might be difficult here. But at the same time, rclone already has an architecture for providing backends through other protocols (e.g. SFTP). One idea I had was, that we decouple things by rclone running as its own process and providing a prococol (backend) that we integrate into RemoteRepository on the Borg side or add a different kind of RemoteRepository.
Borg on the client side takes care of deduping, hashing etc. and the `borg serve` side 'only' needs to support some repository operations.

[...]

PR #8332, opened 2024-08-14
use borgstore and other big changes

### LINKED ISSUE #474: borgception
i thought a bit about how to optimize the chunks cache and just wanted to document one weird idea.

the issue with the chunks cache is that it needs to match the overall repository state (== have up-to-date information about all chunks in all archives, including refcount, size, csize). when backing up multiple machines into same repo, creating an archive of one machine invalidates all chunk caches on the other machines and they need to resync their chunks cache with the repo, which is expensive.

so, there is the idea to store the chunk index into the repo also, so all out-of-sync clients can just fetch the index from the repo.

But:
- index can be large (way larger than segment size)
- when using raw hashtable it could have up to 75% unused bucket space
- index has additional information about chunks, so we should not store it unencrypted if the repo is encrypted
- the index should match the chunks in the repo, so it should not create own chunks in the repo.

So we need:
- chunking of the index into smaller pieces
- compression (unused bucket space is mostly binary zeros AFAIK)
- encryption

This pretty much sounds like we should just backup the index of repo A into a related, but separate borg repository A'. :-)

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/28925696-borgception)

### LINKED ISSUE #768: Please allow "borg create" while repo/archive is mounted via "borg mount"
Using `borg mount` to access an archive or even the whole repository is incredibly powerful and a very nice feature.

Unfortunately, this also locks the repository so that no new archives can be created while an archive from the same repository (or even the repository itself) is mounted.

My understanding is that a mounted archive/repository should be read-only anyways, so I do not see a good reason why a simultaneous `borg create` should not be allowed. I could even live with new archives added after the mount no being visible.

Background: If you run `borg create` quite often (like every 15 minutes) and your backed-up sources are quite large, the "restore" window between the end of one `borg create` and the beginning of the next one shrinks or mere minutes. Inspection of archives, non-standard or long-running restores become unfeasible.

Yes, you could store the repository on btrfs or zfs, create a snapshot of the filesystem containing the repository, mount that btrfs/zfs snapshot on /mnt/repository and _then_ run `borg mount` off of that - but this seems a bit over the top, plus there are those of us who still use ext4 or similar, non-snapshotting file systems.

Changing the code so that `borg mount` does not lock the repository seems like a huge improvement in usability.

### LINKED ISSUE #916: Many servers without cache rebuilding
Is it reasonable to use Borg at scale, backing up many servers to one repository? I'm afraid the need to rebuild the local cache is a blocker for this use case.

I may be way off base with this suggestion, but could the local cache be configured to store only the checksums from the local server? Sure, the servers may send some files to the repository that have already been hashed elsewhere, but no rebuilding would be required (not to mention lower space requirements for the local cache).

---

:moneybag: [there is a bounty for this](https://www.bountysource.com/issues/33011952-many-servers-without-cache-rebuilding)

### LINKED ISSUE #1244: Doom transactions on critical errors in Repository as fail-safe
Doom(able|ed) transactions are a concept a few databases have to guard against wrong error handling but also allow easy transaction control / abortion. We would only use it as a fail-safe against incorrect error handling and also "strange" issues with networking should they occur.

It could look like this:

``` py

# Sketch code, not fleshed out, certainly not tested.

class Repository:
    @staticmethod
    def exception_dooms(whitelist=(,)):
        """All exceptions not on *whitelist* raised in the decorated method will doom the current transaction."""
        def decorator(method):
            def meth_wrapper(self, *args, **kwargs):
                try:
                    return method(self, ...)
                except Exception as exc:
                    if type(exc) not in whitelist:
                        self.doom_transaction(exc)
                    raise
            return meth_wrapper
        return decorator

    def doom_transaction(self, exception=None):
        self.transaction_doomed = True
        self.transaction_doomed_by = exception

    def begin_transaction(self):
        ...
        self.transaction_doomed = False

    @exception_dooms()
    def write_stuff(self, ...):
        # do something
        if bad:
            raise OSError

    @exception_dooms(whitelist=(ObjectNotFound,))
    def get(self, id):
        ...

    def commit(self):
        if self.transaction_doomed:
            log_exception(self.transaction_doomed_by)
            assert not self.transaction_doomed, 'Tried to commit doomed transaction'
         ...
```

### LINKED ISSUE #1293: Maybe use per-archive cache for delete / prune

### LINKED ISSUE #1766: reduce cache size needs (txn.active + files)
Related to "significantly reduce cache space needs" #235. But here it isn´t large chunks.archive.d, but rather txn.active + files. This is a cache for /home which has an excessive number of files, but its not that large:

merkaba:~> du -sh /home
164G    /home

merkaba:~> find /home | wc -l
2865746
merkaba:~> find /home -type f | wc -l
2797574
merkaba:~> find /home -type d | wc -l
64319
merkaba:~> find /home/martin/.local/share/local-mail | wc -l
2128451
merkaba:~> find /home/martin/.local/share/local-mail -type f | wc -l
2127093
merkaba:~> find /home/martin/.local/share/local-mail -type d | wc -l
1358

Today a borgbackup run into exception due to insufficient free space (see issue "limit cache space needs to avoid out of space exceptions" #1765)

And really:

merkaba:~/.cache/borg> du -sch \* | sort -rh | head -4
5,4G    insgesamt
4,9G    673e17ea929[…]
420M    809d8850c0bd[…]
44M     723a580c358b[…]

merkaba:~/.cache/borg/673e17ea92925012e32db8d4d92c6ec18a2a08f490d431b068fe6bdaae073737> LANG=C du -sch \* | sort -rh | head -5
4.9G    total
2.3G    txn.active
2.1G    files
296M    chunks.archive.d
253M    chunks

### LINKED ISSUE #2398: check --repair commit prompt
After #2397: Ability to prompt before committing - avoids having to run check twice (once to see what it finds, another time to apply repairs, with possibly different results).

### LINKED ISSUE #2444: review IntegrityErrors, especially when they kill

### LINKED ISSUE #2454: Review check operation if no commit exists
Cf. https://github.com/borgbackup/borg/pull/2432#discussion_r113882464

### LINKED ISSUE #2571: Allow disabling chunks.archive.d before first cache sync is done.
We currently recommend creating `chunks.archive.d` in the cache directory as file to disable the chunks archive.
At work we want to do this before the repository is accessed the first time from a given system.
So my first attempt was to create the cache directory and place create a file `chunks.archive.d` there. But borg assumes that when the directory exists that it must contain a cache with a valid configuration file. This of course was not the case.
Design space:
* Change the detection for a existing cache to use e.g. the existence of the config file as indicator (this would solve the problem for expert user who can actually created the needed directory from the repository id, but this doesn't sound to nice)
* Add a (global) command line option to disable the chunks archive (that maybe persists this setting?)
* Add a more hidden way to enable this (environment variable, etc)

### LINKED ISSUE #2681: Permit partial cache sync on archive level
Refinement of #2659

-    Skip version!=1 archives during cache sync (or use per-archive flags in conjunction with version=2)
-   When switching to version=2 archive format, add a feature flag to read/check/delete (but not write)

### LINKED ISSUE #3036: apply transactions within time range
I'm using `borg serve` with the `--append-only` option set to have a server backup itself to a remote repository [without being able to tamper with the backup data](https://borgbackup.readthedocs.io/en/stable/faq.html#how-can-i-protect-against-a-hacked-backup-client).

That's fine, but prune operations with the append-only mode activated of course doesn't free space. That's why I'm struggling how to handle this properly. I could run a local borg command which applies all transactions, but I would then have to trust there are no bad transactions already in the repository.

One may be aware of an attack after some time has already passed. **That's why would like to have a borg command to "apply" all transactions within a certain period of time**, e.g. older than 30 days, where I am (quite) sure the backup client hasn't been corrupted in this time.

### LINKED ISSUE #3128: use chunks.archive.d for borg info repo::archive?
check if we can speed up using the single-archive hashindex cached there.

note: for my own ~500GB repo and a ~130GB archive, `borg info repo::archive` took ~6 minutes.

### LINKED ISSUE #3196: chunks cache sync log output
```
$ borg-1.0.11 create -v tw@localhost:repo::test input/
Synchronizing chunks cache...
Archives: 9, w/ cached Idx: 0, w/ outdated Idx: 0, w/o cached Idx: 9.
Fetching and building archive index for test2 ...
Merging into master chunks index ...
...
Fetching and building archive index for test5 ...
Merging into master chunks index ...
Done.
```
In #2866 the question came up whether that is too verbose and should not be shown without --progress given.

1.1.0 output is the same.

### LINKED ISSUE #4004: use fine-grained locks
it would be nice to evaluate all locks and fine-tune them.
maybe different actions hold different locks resulting in actions being able to be done at the same time.
the most basic approach would be read/write which may allow for ```list``` or ```info``` actions during a running ```create```. this would also create the need to handle incomplete data. e.g. the running backup should be ignored. it would be a good idea to state that one is currently running, though.

maybe it's possible to make long-runnig tasks like a ```check``` non-blocking.

### LINKED ISSUE #4428: config: separate into static and dynamic parts
see #4422 discussion.

in the repo and also in the cache, we have a ini-like `config` file that stores some values.

some of them are mostly-static:
- repo in repokey mode: the encrypted key material (only seldomly changed manually) - very important, must not be lost.
- repo parameters (like segment size, segment count per directory) - also quite static, not super important though (can be determined by looking at data/ contents)

some are more dynamic:
- repo last_segment_checked - frequently updated when incremental repo checks are used.
- nonce? check how we store that.

while we have some rather save config-updating code, we want to avoid by all means that some of the rather dynamic updates somehow kills the important config entries, so splitting the config into config and state would be a good thing.

### LINKED ISSUE #4438: wishlist: Actual (not necessarily fast) stats on dry-run delete and prune
The delete (and prune) operations on --dry-run currently report 0B of "Deleted data"; that borg accepts those arguments together is considered an error and being worked on at #4373 on grounds that it could not run fast.

I suggest that --dry-run and delete/prune should take its time to work through what would be deleted, which might take just as long as the full thing but be non-destructive. (Possibly gated by a flag that allows slow operations, or only gated by the --stats flag given that whoever requested the stats probably just wants what this is about).

The use case for this is to judge the effectiveness of changes to the pruning policy – say approaching disk space limits and want to know whether it'll do me any good to decrease the number of daily backups or whether I need to delete old monthlies, or on which repositories that would actually do any good.

Judging from the discussion in #4373, implementing this would not only mean actually calculating the to-be-freed parts, but also to persist the information of what-would-already-be-gone between the distinct invocations of delete that result from a prune.

Yes

It's an issue, but out of the scope of currently documented behavior, thus classified as wishlist item.

## System information.

`borg 1.1.9`

`borg delete -v --stats --dry-run chrysn@prometheus:hephaistos.borg::chrysn-full-2016-11-29-18:26`

### LINKED ISSUE #4827: Measure how much the chunk cache changes in big real life repositories.
While discussing ways to avoid cache sync on irc @ThomasWaldmann came up with a simple idea how to use current borg to measure how much transfers a simple chunk based "store cache into the repository" approach would take.

The basic idea is to see if using mostly existing code to store the chunk cache into the repository can produce good enough results to further investigate.

The basic idea is to take a repository with a fairly large chunk cache that is only accessed from a single location and to use borg to backup just the chunk cache of that repository to a local dummy repository. Using --stats gives an indication of the amount of traffic that saving the cache would add. And also how much data would be needed to transfer for a different location to catch up to the current cache. (catching up with multiple backups in one go might or might not need less data transfer)

It would be interesting to see this with some different chunker parameters. I would expect a fixed size chunking setting to be the prime target, but of course all chunker setting can be tested.

### LINKED ISSUE #5050: Easy way to retrieve hints data from encrypted repository without key
Is there an easy way to retrieve hints (and other) data from an encrypted repository without having the key? I would like to parse the currently used quota on my backup server to send quota warnings to users without needing to enter a passphrase. I also don't want to spend time scrubbing the disk to get an estimated quota that is even slightly wrong.

My quick and dirty way:
```
def get_borg_repo_hints(repo_path):
    import os
    from borg.repository import Repository
    from borg.helpers import msgpack

    repo = Repository(repo_path)
    transaction_id = repo.get_index_transaction_id()
    hints_path = os.path.join(repo_path, 'hints.%d' % transaction_id)
    with open(hints_path, 'rb') as fd:
        hints = msgpack.unpack(fd)
    return hints
```

This is of course not very safe, doesn't use locking and doesn't verify the hints file integrity. I could expand this of course, by following the implementation from `Repository.prepare_txn()` but I don't really want to and don't think it is a good idea to re-implement borg repository handling outside of borg itself.

Is there a better way without digging even deeper and reimplementing even more? (I would prefer to just run `borg info --json` but this requires a passphrase which I do not have and do not want.)

If not: Would it be feasible to add an option (like `--only-unencrypted`) to `borg info` or `borg debug` that would make it possible to retrieve all information that is not encrypted?

(I know I am working against the borg attack model here. But some data will never be encrypted because it is impossible to combine server enforced repository quotas without the server being able to read/write the used quota value. If it is not encrypted it should not be hard to make use of it.)

### LINKED ISSUE #5261: try parallel ops using --bypass-lock
borg usually does not let you do multiple operations in parallel. it blocks that using an exclusive lock on the repository.

there is a (potentially dangerous) hack with `--bypass-lock`.

for some situations, it might actually work ok without leading to problems.

practically try:

- `borg create repo::archive lotsofdata`  plus `borg mount repo`
- `borg mount repo` plus `borg create repo::archive lotsofdata`
- (and other read-only operations like borg list)

Do not try multiple repo-writing ops in parallel!
Do not use a production repo for experiments, just create a temporary one!

### LINKED ISSUE #5514: investigate locking system
borg operations always use an exclusive lock on the repository.

this goes back to a change when a deadlock issue related to lock-upgrade (from shared to exclusive lock) was resolved by just always doing exclusive locks.

always-exclusive locks can be a problem if one has long-running (or frequent) backup jobs (or in general: repo-writing operations) and in parallel one wants to do other borg operations:

- "read-only" operations, like `borg mount / list / info / extract`
- write operations (this is definitely the more critical / complicated case and could be omitted to simplify)

topics to research in this context could be:
- analyze / document repo locking requirements on a rather detailed level
- analyze / document other (cache? clientside indexes?) locking requirements on a rather detailed level
- can we have a deadlock-free locking without using always-exclusive locking?

### LINKED ISSUE #5654: improve stale lock handling
taken from #955 where this was suggested:

by @jdchristensen : However, if some kind of automated lock removal was desired, one idea would be for borg to update the timestamp on the lock every 5 minutes. Then a future borg process could break a lock if it found it older than, say, 10 minutes.

by @poelzi : The correct way to handle such locks is with a pid+machine id inside the lock and and timestamp couter that gets updated every minute. If there is no such process on the same machine: stale lock, remove. If the machineid is a foreign machine and the timestamp is not updated some minutes, remove lock. If the lock is updated, wait until cleared.

### LINKED ISSUE #6057: Aborting (Ctrl + C) borg check leaves a lock

Yes

BUG / ISSUE report 

Not so easy, since borg serve has no support for this?

```
$ borg --version
borg 1.1.17
```

```
$ uname -a
Linux desktop 5.14.0-4-amd64 #1 SMP Debian 5.14.16-1 (2021-11-03) x86_64 GNU/Linux
```

`borg check`

Execute `borg check` and press `Ctrl` + `C`. Next operation on the repo yields 

> Failed to create/acquire the lock <repo>/lock.exclusive (timeout).

See above.

### LINKED ISSUE #6094: Stale Locks Remain Stale Until you Manually Unlock Them (not good for automation)
I find it ironic that this issue is "locked" (given that it pertains to stale "locks"). The issue is marked as resolved, but it doesn't seem to be in resolved in borg 1.1.16:
https://github.com/borgbackup/borg/issues/3191

A power outage is what caused my stale lock. I can unlock it, but what concerns me is Borg's inability to auto-notice that a lock is stale. I ultimately intend to automate these borg backups, and if it cannot recover from a stale lock, I can see potential for recurring failures going unnoticed. I will certainly take great care in my scripting to be notified of such failures, because I read [here](https://github.com/borgbackup/borg/issues/3191) of a person who went 3 months without noticing that a stale lock was preventing all his automated backups!

Rather than just complaining about this, maybe I can suggest some solutions. Hmm.

Certainly borg logs activity on a repo with timestamps, no? If so, it seems that a stale lock rejection should trigger a log evaluation to occur, if there's been no log activity in the last 10 minutes (for example), then instead of continuing to cry about a lock, maybe borg should simply make a checkpoint out of the abandoned session and auto-unlock the repo. That's the first thing that comes to my mind; hopefully someone else can top it.

### LINKED ISSUE #6121: Fetching archive indexes is extremely slow - is network usage inefficient?
For reasons outlined in the FAQs, having a client side chunk cache may be inconvenient because it can end up taking a huge amount of space wrt the client storage abilities that may be limited. Furthermore, in many situations it only rarely occurs that the local “chunks” index needs to be rebuilt.  On the other hand, in these rare occasions, not having a cache means that one spends a huge amount of time fetching archive indexes.

Now, I understand that there is a space-time trade off here and that borgception is meant to relax it one day.  However, I wonder if there is something that I could try until borgception arrives. Specifically, I would like to understand if the huge amount of time needed by the index fetching may be due to some unfortunate setup in my systems and could be reduced by modifying such configuration.

What strikes me is that during this fetching:
- the client (the machine being backed up) appears to be mostly unloaded;
- the server (the machine storing the back up) appears to be mostly unloaded too;
- the network appears quite loaded, but so to say in an inefficient way: differently from what would happen with a large file transfer on the net, the network graphs do not go up and stay up, but appear to be made of quick bursts, so that the average download rate on the client stays relatively low (~ 250-500 kiB/s on a 5 GHz wifi network where this is substantially the only activity and the throughput could be much larger). At the same time, I see a non-negligible (some tens kB/s) upload rate from the client.

Is this expected? Is this the sign that there is something unfavourable on my setup? Is it the sign that the protocol involves lots of small transfers and bidirectional exchanges? Could something (e.g. in the protocol) be improved to better exploit the network throughput (e.g., by putting more CPU at use either on the client or the server or both in this procedure)?

Yes

Question possibly revealing an issue/inefficiency.

1.1.17 (both client and server)

Ubuntu 20.04 (both client and server)

Client is a laptop. Server is a mini desktop. Both intel (client is way more recent and powerful, but it is my understanding that requirements on server are low).  Client and server communicate via wifi (it is unfortunate that the server is temporarily on wifi too) via a wifi router/access point.
Wifi "published" connection speed is 780 Mbit/s for the client and 390 Mbit/s for the server.
Filesystem: ext 4. Client on SSD, server on rotating disk with disk on USB3 attachment.

### LINKED ISSUE #6288: borg init: add a --additional-free-space=X option? default?
On a remote backup the disk quota was reached and borg failed to connect because ``borg serve`` just dies and locally I get:
``Connection closed by remote host. Is borg working on the server?``

I couldn't find a way to get it working again, so I ended up having to delete the full backup and loose all history.

I found the ``storage_quota`` and ``additional_free_space`` options, but no way to set them on an existing backup except by manually using sftp. It seems only ``--storage-quota`` is exposed via the ``init`` command, but not the other option (which from what I read seems more useful).

Locally I have borg 1.1.17, not sure what version Hetzner has on its storage boxes/backup spaces.

Shouldn't ``additional_free_space`` be set by default to prevent unrecoverable errors like this?

### LINKED ISSUE #6289: segments with mostly DEL are "compacted" again and again
Noticed while doing #5679:

1.2.0rc1 (likely same in 1.1.x)

borg compact REPO

If one heavily deleted archives, there are likely some segments mostly consisting of DEL tags.
In `hints -> compact` data structure, a DEL is accounted as a freeable object of 41 bytes (9 header + 32 chunk id).
Thus, such a segment is about 100% freeable and a candidate for compaction.

Not nice: if the DELs are still needed because the segments with the respective PUTs were not covered by the compaction run (e.g. because they are below compaction threshold of 10%), it will just shuffle all these DELs to a new segment, without freeing anything.

And as that new segment again is 100% freeable, this will repeat in future compaction runs.

### LINKED ISSUE #6291: force total compaction?
Related to #6289:

When investigating that, I also ran `borg compact --threshold=0 REPO` and it still kept the segments full of DELs.

So, my current guess about that would be that `hints -> compact` is incomplete and does not hint to all segments where stuff could be compacted (== where PUTs exist that are DELeted in later segments).

Possible solutions:
- `borg check --repair` + `borg compact` maybe? check whether this completely recomputes the hints file contents.
- modify `borg compact` so it unconditionally runs over all segments (that would be a quite I/O heavy operation, rewriting all not-deleted segment content to new segments)

### LINKED ISSUE #6331: FileNotFoundError thrown on borg create: hints-file missing

Yes

ISSUE / QUESTION

borg 1.1.16

Debian 11.2 "Bullseye"

Asus PN41 Mini-PC with a Samsung 970 EVO Plus 2TB, local backup, no network involved
ext4 with LVM (encrypted)

300GB

```
/usr/bin/borg create --compression zstd --exclude-caches --exclude '*/.Trash-*' '/var/borg/local::{hostname}-{now:%Y-%m-%d-%H%M%S}' '/var/nas' '/var/www'
```

I use borg to create hourly backups on my NAS-server. Some times `borg create` throws an Exception, because a hints-file (like hints.2763.tmp) was not found inside the repository. On the next run, I get some warnings that segments are not found, but listed in compaction data (see logs below). The following runs never return any errors.

The data on the NAS rarly changes and the backuped files and the repository are stored on the same fast NVMe (in case you are wondering: I use the backup to be able to undo changes, e.g. files overwritten by mistake). The `borg create` command runs only for seconds, but it is possible that files are changed while the backup runs.

Are these Exceptions critical? What can I do to avoid them?

The same server also uses borg to backup the data to an external cloud-storage. This backup only runs once a day and never returns any errors.

Thanks for your advice!

No, the Exception occurs only sporadic.

I run `borg check --verify-data` on the repository and it did not found any errors. I can also mount the repository and restore data from it.

Run with error:

```
Local Exception
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4690, in main
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 4622, in run
    return set_ec(func(args))
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 177, in wrapper
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 595, in do_create
    create_inner(archive, cache)
  File "/usr/lib/python3/dist-packages/borg/archiver.py", line 560, in create_inner
    archive.save(comment=args.comment, timestamp=args.timestamp)
  File "/usr/lib/python3/dist-packages/borg/archive.py", line 530, in save
    self.repository.commit()
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 476, in commit
    self.write_index()
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 637, in write_index
    rename_tmp(hints_file)
  File "/usr/lib/python3/dist-packages/borg/repository.py", line 587, in rename_tmp
    os.rename(file + '.tmp', file)
FileNotFoundError: [Errno 2] No such file or directory: '/var/borg/local/hints.2763.tmp' -> '/var/borg/local/hints.2763'

Platform: Linux server 5.10.0-11-amd64 #1 SMP Debian 5.10.92-1 (2022-01-18) x86_64
Linux: Unknown Linux  
Borg: 1.1.16  Python: CPython 3.9.2 msgpack: 0.5.6.+borg1
PID: 519872  CWD: /root
sys.argv: ['/usr/bin/borg', 'create', '--compression', 'zstd', '--exclude-caches', '--exclude', '*/.Trash-*', '/var/borg/local::{hostname}-{now:%Y-%m-%d-%H%M%S}', '/var/nas', '/var/www']
SSH_ORIGINAL_COMMAND: None
```

The following run:

```
segment 2758 not found, but listed in compaction data
segment 2760 not found, but listed in compaction data
```

### LINKED ISSUE #6567: optimise calls to iter_objects (repo v2)
if a repo only has PUT2 (and no PUT any more), e.g. `replay_segments` should call `iter_objects(..., read_data=False)`.

with PUT tags, a CRC32 check was only possible when reading the data (even if we didn't really need the data).
with PUT2, the CRC32 check is now possible without reading the data, so we can get much faster.

`replay_segments` is called e.g. if there is:
- an index_transaction N
- a segments_transaction M
- N > M
- that means: some segment files at the end were deleted (I just did that for a repo on a disk with I/O errors - in that case it will completely read ALL the segment files..................................).

besides `replay_segments` also check **all other callers** of `iter_objects` which do not already give `read_data=False` - maybe they could!

### LINKED ISSUE #6899: interrupted borg compact / compact_segments
### Discussed in https://github.com/borgbackup/borg/discussions/6283

<div type='discussions-op-text'>

<sup>Originally posted by **sophie-h** February 11, 2022</sup>

Will borg compact in borg 1.2 be able to continue where it was left off?

I'm currently trying to prune hundreds of archives but there does not seem to be an easy way to complete this.
</div>

### LINKED ISSUE #6983: repository: add index query function
There are some infos in the repo index which the client might need:
- csize, the encrypted/compressed chunk size
- maybe flags

Scenarios:
- ~~chunks index gets lost on the client side (has id -> refcount, size mapping)~~ (we can neither reconstruct the refcount from the index nor the uncompressed plaintext size, since borg2 we do not have csize in the chunks index any more)
- ~~chunks index missing size info for some chunks~~ (same reason)
- we could get csize for some statistics

### LINKED ISSUE #7154: OSError: [Errno 39] Directory not empty: '/.../lock.exclusive' (borg 1.1.x)
Hi,

I am seeing borg unable to delete lock.exclusive directory even though the directory is empty. This started happening since 10days ago and prior to that it was running without any issues. 

Setup:
- borg is running in a docker container - alpine:3.11.3
- mounting S3 storage as filesystem using s3fs
- backup is taken on the mounted file-system

was running fine until 10days ago and also is not happening every time. some days, backup is running just fine. 

```

Exception ignored in: <function Repository.__del__ at 0x7f0087f1e8b0>
  File "/usr/lib/python3.8/site-packages/borg/repository.py", line 178, in __del__
    self.close()
  File "/usr/lib/python3.8/site-packages/borg/repository.py", line 427, in close
    self.lock.release()
  File "/usr/lib/python3.8/site-packages/borg/locking.py", line 383, in release
  File "/usr/lib/python3.8/site-packages/borg/locking.py", line 282, in modify
    elements.remove(self.id)
KeyError: ('runner-6e2b1c97-project-456-concurrent-0@24853
Local Exception
  File "/usr/lib/python3.8/site-packages/borg/archiver.py", line 4501, in main
  File "/usr/lib/python3.8/site-packages/borg/archiver.py", line 4433, in run
    return set_ec(func(args))
  File "/usr/lib/python3.8/site-packages/borg/archiver.py", line 166, in wrapper
  File "/usr/lib/python3.8/site-packages/borg/repository.py", line 203, in __exit__
    self.close()
  File "/usr/lib/python3.8/site-packages/borg/repository.py", line 427, in close
    self.lock.release()
  File "/usr/lib/python3.8/site-packages/borg/locking.py", line 3, in release
    self._lock.release()
  File "/usr/lib/python3.8/site-packages/borg/locking.py", line 150, in release
    os.rmdir(self.path)
OSError: [Errno 39] Directory not empty: '/mnt/prod/lock.exclusive'
```

### LINKED ISSUE #7278: Unnecessary chunks cache sync?

Yes.  There's a chance this is related to #7274 , but I think it is probably different.

BUG

Ubuntu 22.04

borg 1.2.0.  Also happens with master branch borg2.

Ubuntu 22.04

N/A

N/A

See below.

When a host accesses a borg repo after another host has accessed it, sometimes some archives have their chunks cache synced even when they should already be in the cache.  I first noticed it in #7277 (with borg2), where the line
```
Fetching and building archive index for host1test19
```
shouldn't be there, since that archive was made with the same host.  Then I reproduced it with borg 1.2.0, using the steps shown below.

```
borg-test-sync$ mkdir host1 host2 test-data
borg-test-sync$ echo foo > test-data/foo
borg-test-sync$ export BORG_BASE_DIR=host1
borg-test-sync$ export BORG_REPO=test.repo
borg-test-sync$ borg init -e none
borg-test-sync$ borg create -v ::${BORG_BASE_DIR}-test1 ./test-data
Creating archive at "test.repo::host1-test1"
borg-test-sync$ borg create -v ::${BORG_BASE_DIR}-test2 ./test-data
Creating archive at "test.repo::host1-test2"
borg-test-sync$ export BORG_BASE_DIR=host2
borg-test-sync$ borg create -v ::${BORG_BASE_DIR}-test3 ./test-data
Warning: Attempting to access a previously unknown unencrypted repository!
Do you want to continue? [yN] y
Creating archive at "test.repo::host2-test3"
Synchronizing chunks cache...
Archives: 2, w/ cached Idx: 0, w/ outdated Idx: 0, w/o cached Idx: 2.
Fetching and building archive index for host1-test1 ...
Merging into master chunks index ...
Fetching and building archive index for host1-test2 ...
Merging into master chunks index ...
Done.
borg-test-sync$ borg create -v ::${BORG_BASE_DIR}-test4 ./test-data
Creating archive at "test.repo::host2-test4"
borg-test-sync$ export BORG_BASE_DIR=host1
borg-test-sync$ borg create -v ::${BORG_BASE_DIR}-test5 ./test-data
Creating archive at "test.repo::host1-test5"
Synchronizing chunks cache...
Archives: 4, w/ cached Idx: 0, w/ outdated Idx: 0, w/o cached Idx: 4.  <<<<<<
Fetching and building archive index for host1-test1 ...  *************
Merging into master chunks index ...
Fetching and building archive index for host1-test2 ...  *************
Merging into master chunks index ...
Fetching and building archive index for host2-test3 ...
Merging into master chunks index ...
Fetching and building archive index for host2-test4 ...
Merging into master chunks index ...
Done.
```
In this case, both archives made on "host1" have their data fetched, even though we are currently on host1.  See the two lines marked with "******".  Also, the line marked with "<<<<<<" shows that borg thinks there is nothing locally cached.  In #7277, only one gets synced that shouldn't, so it is less extreme there.  Not sure what is going on.

### LINKED ISSUE #7377: repository - going away from transactions, log, refcounts?
## current state (current master branch, borg 1.x, borg 0.x, attic)

A borg repository is primarily a key/value store (with some aux functions).

The key is the chunk id (== MAC(plaintext)), the value is the compressed/encrypted/authenticated data.

borg uses transactions and a LOG when writing to the repo:
- start of transaction (usually triggered by PUT/DEL)
- writes more objects by appending PUT entries to the log
- deletes objects by appending DEL entries to the log
- commits (appends a COMMIT entry to the log)
- end of transaction (S: saves repo index and hints, C: saves chunks index and files cache)

LOG means that new stuff is always appended at the end of the last/current segment file. In general, old segment files are never modified in place.

`borg compact` defrags non-compact segment files:
- a segment file contains PUTs, DELs, COMMITs
- if a PUT(id) is later deleted by a DEL(id), it creates a logical hole in a segment file (that object is not used any more), making it non-compact
- compaction / defragging works by reading all still-needed objects from an old segment file and appending them to a new segment file. after that is finished, the old segment file is deleted (and that frees disk space because the new segment file is smaller).

## advantages of this approach
- transactions and append-only log are a very safe approaches (even if stuff crashes it usually can roll back to previous state and be fine again)
- segment files are medium size files: not too large, not too small, not too many
  - works well even with not very scalable filesystems
  - has little overhead due to fs block / cluster size
  - can be copied or deleted rather quickly (not many fs objects)

## disadvantages of this approach
- borg compact can cause lots of I/O when shuffling objects from old non-compact segments to new compact segments
- borg compact needs some space on the fs to be able to work. bad if your fs is 100% full...
- compaction code is rather complex, same for transaction management
- to quickly access objects, the repository needs an index mapping `id -> (segment, offset, flags)`
- borg currently loads the repo index (hashtable) into memory. RAM usage is about 44b * object_count + free space in hashtable. if you have a lot of files and/or a lot of data volume, repo index can need GBs of RAM.
- to implement this, some special borg code is needed with access to the repo filesystem
- hard to work like this without locking the repository against concurrent access.

### LINKED ISSUE #7379: get rid of refcounts?
borg currently is usually precisely refcounting objects. this ticket is split off from #7377 to focus on refcounting here only:
- precise refcounting 0..N (like currently done in borg)
- is `object_still_needed = bool(refcount > 0)` all we really need?

### LINKED ISSUE #8325: Borg taking 2 days to complete backup as it is Replaying segments

Yes

BUG / ISSUE 

borg 1.4.0

Ubuntu Server 24.04 LTS

ext4
rclone v.167
fuse

1226579 files

```
#!/bin/bash

SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
CONFIG="$SCRIPT_DIR/BackupRestore-Nextcloud-rem.conf"

# Check if config file exists
if [ ! -f "$CONFIG" ]; then
    echo "ERROR: Configuration file $CONFIG cannot be found!"
    echo "Please make sure that a configuration file '$CONFIG' is present in the main directory of the scripts."
    echo "This file can be created automatically using the setup.sh script."
    exit 1
fi

source "$CONFIG"

# Create a log file to record command outputs
touch "$LogFile"
exec > >(tee -a "$LogFile")
exec 2>&1

# Function for error messages
errorecho() {
    cat <<< "$@" 1>&2
}

## ---------------------------------- TESTS ------------------------------ #
# Check if the script is being executed by root or with sudo
if [ $EUID -ne 0 ]; then
   echo "========== This script needs to be executed as root or with sudo. ==========" 
   exit 1
fi

# -------------------------------FUNCTIONS----------------------------------------- #
BORG_OPTS="--verbose --filter AME --list --progress --stats --show-rc --compression lz4 --exclude-caches"

# Start Rclone Mount
systemctl start borgbackup.service

borg break-lock $BORG_REPO

# Function to Nextcloud Maintenance Mode
nextcloud_enable() {
    # Enabling Maintenance Mode
	sudo -u www-data php $NextcloudConfig/occ maintenance:mode --on
}

nextcloud_disable() {
    # Disabling Nextcloud Maintenance Mode
	sudo -u www-data php $NextcloudConfig/occ maintenance:mode --off
}

# Function to WebServer Stop Start
stop_webserver() {
    # Stop Web Server
	systemctl stop $webserverServiceName
}

start_webserver() {
    # Stop Web Server
	systemctl start $webserverServiceName
}

prune() {
    info "Pruning repository"

    # Use the subcoming `prune` to keep 7 days, 4 per week and 6 per month
    # files of this machine.The prefix '{hostname}-' is very important for
    # limits PLA's operation to files in this machine and does not apply to
    # Files of other machines too:

    borg prune --list --progress --show-rc --keep-daily 7 --keep-weekly 4 --keep-monthly 6

}

# Function to backup Nextcloud settings
nextcloud_settings() {
    echo "========== Backing up Nextcloud settings $( date )... =========="
    echo ""

    nextcloud_enable
    stop_webserver

   	# Export the database.
	mysqldump --quick -n --host=localhost $NextcloudDatabase --user=$DBUser --password=$DBPassword > "$NextcloudConfig/nextclouddb.sql"

    # Backup
    borg create $BORG_OPTS ::'NextcloudConfigs-{now:%Y%m%d-%H%M}' $NextcloudConfig --exclude $NextcloudDataDir

    # Remove the database 
    rm "$NextcloudConfig/nextclouddb.sql"

    start_webserver
    nextcloud_disable
}

# Function to backup Nextcloud DATA folder
nextcloud_data() {
    # Filters for Inclusion Exclusion Borg
    BorgFilters="./nc-patterns.lst"

    # Create a file with the delete standards Borg Inclusion
    tee -a "$BorgFilters" <<EOF > /dev/null 2>&1
P sh
R /

# DO NOT LOOK IN THESE FOLDERS
! proc

# DIRECTORIES TO BE EXCLUDED FROM BACKUP  
- $NextcloudDataDir/*/files_trashbin

# DIRECTORIES FOR BACKUP
+ $NextcloudDataDir/

# DO NOT INCLUDE ANY MORE FILES
- **
EOF

    echo "========== Backing up Nextcloud DATA folder $( date )...=========="
    echo ""

    nextcloud_enable

    borg create $BORG_OPTS --patterns-from "$BorgFilters" ::'NextcloudData-{now:%Y%m%d-%H%M}'

    rm "$BorgFilters"

    nextcloud_disable
}

# Function to perform a complete Nextcloud backup
nextcloud_complete() {
    echo "========== Backing up Nextcloud $( date )... =========="
    echo ""

    nextcloud_enable
    stop_webserver

   	# Export the database.
	mysqldump --quick -n --host=localhost $NextcloudDatabase --user=$DBUser --password=$DBPassword > "$NextcloudConfig/nextclouddb.sql"

    # Backup
    borg create $BORG_OPTS ::'NextcloudFull-{now:%Y%m%d-%H%M}' $NextcloudConfig $NextcloudDataDir --exclude "$NextcloudDataDir/*/files_trashbin"

    # Remove the database
    rm "$NextcloudConfig/nextclouddb.sql"

    start_webserver
    nextcloud_disable
}

# Check if an option was passed as an argument
if [[ ! -z ${1:-""} ]]; then    # Execute the corresponding Backup option
    case $1 in
        1)
            nextcloud_settings
            ;;
        2)
            nextcloud_data
            ;;
        3)
            nextcloud_complete
            ;;
        *)
            echo "Invalid option!"
            ;;
    esac
else
    # Display the menu to choose the Backup option
    echo "Choose a Backup option:"
    echo "1. Backup Nextcloud configurations and database."
    echo "2. Backup only the Nextcloud data folder. Useful if the folder is stored elsewhere."
    echo "3. Backup Nextcloud configurations, database, and data folder."
    echo "4. To go out."

    # Read the option entered by the user
    read option

    # Execute the corresponding Backup option
    case $option in
        1)
            nextcloud_settings
            ;;
        2)
            nextcloud_data
            ;;
        3)
            nextcloud_complete
            ;;
        4)
            echo "Leaving the script."
            exit 0
            ;;
        *)
            echo "Invalid option!"
            ;;
    esac
fi

    # Sleep for 90 minutes before unmounting the drive
sleep 5400

    # Stop Rclone Mount
systemctl stop borgbackup.service
```

After a broken update of a program, I had to reinstall the server again and, to my surprise, since 01/06, my backups have become extremely slow, starting to take about 3 to 4 hours to complete the backup, when they complete at all.

For about a month now, things have only gotten worse, because instead of 3 to 4 hours on a given day, it gives an error that the mount is not available and ends the backup halfway through and the next day it starts the backup normally but it only finishes the next day or takes up to 2 days and within the log the information contained is that it is recreating the cache and then it starts the backup then stops halfway and starts to recreate the cache again.

I have already made several attempts, such as changing the cache storage, disabling the cache, some changes in rclone and even the cloud where the backup is stored. I have changed and nothing seems to have any effect.

Before the problem with the server, I was running Ubuntu 22.04 LTS and borg. 1.2.8

The problem occurs when running the Borg Create command where it takes a long time in an rclone assembly, taking up to 2 to 3 days to perform a backup of just a few MB's.

```
Log File Exceeds
```

### LINKED ISSUE #8330: Cancelled `borg info` invalidates cache, causing expensive sync on next operation
**Have you checked borgbackup docs, FAQ, and open GitHub issues?** Yes

**Is this a BUG / ISSUE report or a QUESTION?** Question

**Borg Version** `borg 1.1.15`

**Operating system (distribution) and version.** Ubuntu 20.04

**Hardware / network configuration, and filesystems used.** Btrfs

**How much data is handled by borg?** 750PB (15000 Snapshots of 50GB, 120GB Deduplicated)

Everything is on the same host and disk (a directory is backed up to a borg repository on the same device, no other devices accessing the same repo)

Cancelling (Ctrl-C'ing) a `borg info` command apparently invalidates the chunks cache, causing an expensive re-build on the next operation (e.g. create)

I don't know what happens under the hood, but `borg info` feels like it should be a read-only operation, and therefore should not be able to invalidate the cache.

Is this a bug, or expected behaviour?

I cancelled the `borg info` call because it was taking a long-ish time (~15 minutes)

Many thanks!

ThomasWaldmann:
Fixes #8330 (no cache sync anymore).

Fixes #8325 (no segments, replaying segments anymore).

Fixes #7377, fixes #7379 (no transactions, no refcounting anymore).

Fixes #7154 (new repo locking code).

Fixes #6983 (no repo index any more).

Fixes #6899 (no compact segments any more).

Fixes #6121, fixes #7278 (no cache sync any more, no archives fetching).

Fixes #6567 (not needed any more).

Fixes #6331 (hints file not used anymore).

Fixes #6291, fixes #6289 (no segment files any more, no DEL tags).

Fixes #6288 in a different way (borg rspace command to reserve disk space that can be freed in case of emergencies).

Fixes #5654, fixes #6057, fixes #6094, fixes #7154 (new lock implementation for borgstore).

Fixes #5514 (new locking system allows shared locks for most).

Fixes #5261 (bypass-lock was removed).

Fixes #5050 (no hints any more).

Fixes #4827 (no persistent chunks cache, no cache sync anymore).

Fixes #4438 (not possible, delete/prune just kill the root reference now and don't look at anything else).

Fixes #4428 (separate objects in config/*).

Fixes #4004 - most borg commands now can use a shared repository lock (exceptions: borg check and compact).

Fixes #3128. Fixes #3196. (no cache sync anymore)

Fixes #2454, fixes #2398, fixes #3036 (no commits, no transactions any more, no log-like/append-only segments, new check implementation).

Fixes #2681, fixes #2571 (no cache sync, no chunks.archive.d).

Fixes #2454 (no commits, no transactions anymore).

Fixes #2444 (new borg check implementation, no LoggedIO code used any more).

Fixes #1293 (solved in a different way, delete/prune are super fast now).

Fixes #1244 (no transactions anymore).

Fixes #916, fixes #474, fixes #1766 (no LocalCache any more, no cache transaction any more).

Fixes #768 - most borg commands now can use a shared repository lock (exceptions: borg check and compact).

Maybe builds the foundation to solve / work on:
- #102 (S3 and other clouds, without FUSE)
- #1070 (sftp: already works)
- #5324

## new repository based on borgstore project

stores chunks into separate files (not: segment/pack files, at least for now).

borgstore has a very simple api that makes implementing backends easy.

in borg 2.0, this will primarily use the "file:" backend from borgstore to implement `file:` and `ssh:` repositories, but long term we might go away from the borg.remote code (RPC api via ssh) and just use a "remote" borgstore.

there is also a `sftp:` repository now implemented via the respective borgstore backend. more might be coming, even cloud stuff should be easily possible with that (PRs welcome!).

## repository: convergence rather than transactions

borg 1.x used the segment files in a log-like way (only appending new stuff at the end) and implemented transactions via a COMMIT tag - if the transaction was not completed (no COMMIT at the end), it rolled back the incomplete transaction to the last commit.

the code implementing transactions was rather complex and required an exclusive lock on the repo for correct operations.

borg2 now just adds repo objects in the right order, first pushing referenced objects, then the references to them. even if an operation is interrupted, nothing bad happens.
there might be some unreferenced objects for a while, but they will get referenced if the operation is retried later and completes. borg compact will deal with anything not needed.

## no checkpoint archives, no .borg_part files anymore

saving them was only needed due to the transactional/rollback behavior of borg1.

borg2 does not do that rollback any more, so the checkpoints are not needed.
the user can just re-run the interrupted command and it will notice that some stuff is already present in the repo and only transfer new stuff.

## new borg compact doing garbage collection

borg compact is still needed to free space in the repo, but it doesn't really need to "compact segments" as there are no segment files anymore. so it will do less I/O to move stuff around in the repo.

but maybe some sort of segment/pack files will come back later, so we'll just keep the command name.

borg compact is now doing more work that was previously done by borg delete and borg check: it determines which chunks are not used anymore (and removes them). because it needs to read the archives for that, borg compact now needs the borg key.

## super fast borg delete and borg prune

as borg does not do precise refcounting any more, delete now just kills the archive from `archives/*` (removing the reference to its root) and lets borg compact clean up all the now unreferenced chunks.

## new repo locking code

repository locking code using borgstore locks/*.

locks auto-expire and get deleted if they don't get refreshed regularly (this is good to clean up stale locks of now-dead remote borg processes).

locks also get deleted if their owner process is dead (this is good to clean up stale locks of now-dead local borg processes).

most borg commands now can use a shared repository lock (exceptions: borg check and compact, which must use an exclusive lock).

## new repo config / repo key storage

Config is now stored into separate files config/* - less risk (e.g. for the repokey) if other config items need updating.

Repokey now stored into keys/* (only 1 key for now).

## new manifest storage

the manifest chunk that also had the list of archives inside was split into config/manifest and separate files archives/*.

some features might come back later (stats, quota, append-only, ...)


ThomasWaldmann:
mypy still unhappy: src/borg/testsuite/locking3.py:5: error: Skipping analyzing "borgstore.store": module is installed, but missing library stubs or py.typed marker  [import-untyped]

src/borg/helpers/misc.py
```diff
@@ -2,7 +2,7 @@
 import io
 import os
 import os.path
-import platform
+import platform  # python stdlib import - if this fails, check that pwd != src/borg/
```
ThomasWaldmann, line 5: cwd not pwd

src/borg/testsuite/archiver/transfer_cmd.py
```diff
@@ -35,6 +35,7 @@ def check_repo():
     check_repo()
 
     archiver.repository_location = original_location + "2"
+    # XXX broken: --other-repo only accepts "Repository", but not "Repository3" repos:
```
ThomasWaldmann, line 38: TODO?

src/borg/archiver/__init__.py
```diff
@@ -36,7 +36,7 @@
     from ..helpers import ErrorIgnoringTextIOWrapper
     from ..helpers import msgpack
     from ..helpers import sig_int
-    from ..remote import RemoteRepository
+    from ..remote import RemoteRepository  # TODO 3?
```
ThomasWaldmann, line 39: TODO?

ThomasWaldmann:
Guess this needs some review. Anybody?

m3nu:
I'm about 25% through and this looks like a great change to decouple storage logic and simplify more things. Still want to run this branch locally to see how it works. Anyways, solid job @ThomasWaldmann!

ThomasWaldmann:
@buengese I suspect it could be quite similar to the rclone REST server for restic, complemented by a REST client/backend in borgstore.

m3nu:
Did some more testing of this PR and didn't encounter any errors. Tested those common cases:

- Create local repo, add some archives and files, extract single file
- Create remote repo over SSH, add some archives and files, extract single file
- Resume interrupted backup to remote repo

Anything you would want to see tested more specifically, @ThomasWaldmann ?

jdchristensen:
I'd be interested in seeing some timing comparisons between this and the previous repo format for various operations under different workloads, including with larger repos with many archives, and also with both local and remote repos.  (I'd volunteer to try this myself, but I'm swamped right now.)

ThomasWaldmann:
@m3nu if you have time:

- try big backups, scalability?
- try sftp://user@host:port/abs/olute/path/to/repo with a big backup, scalability?
- performance comparison with borg 1.x?
- try accessing a repo in parallel (currently, repo locking should behave similar as in borg 1.x, but is a new implementation)

ThomasWaldmann:
@m3nu quota, storage use info, append-only are all not yet implemented.

when running `borg compact -v`, you get some infos, but even that can be further improved.

ThomasWaldmann:
`borg compact -v` now gives some more and more useful infos:
- repo size
- compression and dedup factor
- how much space it freed (if any)

ThomasWaldmann:
@m3nu Yes. The idea is "if you can do it at the source, you can also do it at the repo". At least, due to dedup, it doesn't go up proportional to the archives count. Due to non-deduped files and due to long files being split into multiple chunks, the repo files count will be higher than the src files count.

Of course this has pros and cons, but that was also the case for segment files and segment compaction.

When doing performance comparisons, we have to consider that some bookkeeping work shifted its place and some stuff works rather differently now.

E.g. borg1 delete/prune could take rather long, because it was precisely counting down the refcounts and writing DEL tags to the repo when they reached zero. borg2 (this PR) now just kills the archive root reference from the archives list and leaves the rest of the work to borg compact (which has to determine "is a chunk still needed" for all chunks anyway).

borg1 check also used to do this (finding "orphan chunks"), but as borg2 compact does this now, I have removed it from borg2 check. borg2 compact also notices if chunks needed for healing have appeared and then advises to run borg check --repair. same if borg2 compact notices that you have lost chunks that are in use by archives.

and, of course, borg2 compact does NOT copy around huge amounts of data any more to free space as borg1 did with the segment files to remove their "holes".

borg2 create could be a bit faster due to not checkpointing any more (which IIRC also saved the indexes/caches), which borg1 did every 30 mins by default and more often if people modified it for unstable connections. for borg2, it behaves like "always checkpointing" (because there are no repo transaction rollbacks any more), but there is no overhead at all. 

OTOH, borg2 create could also be a bit slower, because more files are opened/closed and it dynamically loads the existing chunk ids list from the repo at every invocation.

so there likely is quite some optimization potential still, but I consider this way of doing it also as a door-opener for the future, with more parallelism / less locking and more repo backend options. so even if borg2 create takes longer, it could get overall faster in the future, because we hopefully will be able to run multiple backups to same repo at the same time.

m3nu:
Right. It's a trade-off at multiple levels. I believe Restic has a related [setting](https://restic.readthedocs.io/en/latest/047_tuning_backup_parameters.html#pack-size), pack size, to control if one wants more or fewer data segments. To get around limitations in the number of files on the backend.

ThomasWaldmann:
guess I am mostly done here for now.

some stuff needs server-side support (e.g. append-only and enforcing quotas by borg), so that should be delayed until that is decided/reimplemented (e.g. by a borg rest server / borgstore rest client).

ThomasWaldmann:
I did another pass through the issue tracker and added "fixes xxxx" (meaning: fixed in master branch, after this gets merged) to the topmost comment, also roughly sorted these.

We'll get rid of a lot of issues, but due to quite some new code, maybe also gain some new ones.

ThomasWaldmann:
Hmm, I either need some feedback to do more work here or this should get merged ASAP due to the widespread changes causing conflicts for other PRs.

ThomasWaldmann:
TODO: encrypt `archives/*` objects (contain archive name, ID, timestamp).

witten:
Initial feedback... Running the following on a non-existent repo directory (which borgmatic does to probe whether a configured repo already exists prior to `rcreate`):

```bash
$ ~/Downloads/borg-env/bin/borg rinfo --json --repo ~/Downloads/test.borg 
```

Results in this (exit code 2):

```
Local Exception

Error:

BackendDoesNotExist: posixfs storage base path does not exist or is not a directory: /home/witten/Downloads/test.borg

If reporting bugs, please include the following:

  File "/home/witten/Downloads/borg/src/borg/archiver/__init__.py", line 622, in main
                ^^^^^^^^^^^^^^^^^^
  File "/home/witten/Downloads/borg/src/borg/archiver/__init__.py", line 516, in run
    rc = func(args)
         ^^^^^^^^^^
  File "/home/witten/Downloads/borg/src/borg/archiver/_common.py", line 151, in wrapper
    with repository:
  File "/home/witten/Downloads/borg/src/borg/repository.py", line 133, in __enter__
    self.open(exclusive=bool(self.exclusive), lock_wait=self.lock_wait, lock=self.do_lock)
  File "/home/witten/Downloads/borg/src/borg/repository.py", line 180, in open
    self.store.open()
  File "/home/witten/Downloads/borg-env/lib/python3.12/site-packages/borgstore/store.py", line 65, in open
    self.backend.open()
  File "/home/witten/Downloads/borg-env/lib/python3.12/site-packages/borgstore/backends/posixfs.py", line 54, in open
    raise BackendDoesNotExist(
borgstore.backends.errors.BackendDoesNotExist: posixfs storage base path does not exist or is not a directory: /home/witten/Downloads/test.borg

Platform: Linux flux 6.1.99-1-MANJARO #1 SMP PREEMPT_DYNAMIC Mon Jul 15 21:35:02 UTC 2024 x86_64
Linux: Unknown Linux  
Borg: 2.0.0b10.dev80+g74a67749  Python: CPython 3.12.4 msgpack: 1.0.8 fuse: None [pyfuse3,llfuse]
PID: 2588631  CWD: /home/witten/Downloads
sys.argv: ['/home/witten/Downloads/borg-env/bin/borg', 'rinfo', '--json', '--repo', '/home/witten/Downloads/test.borg']
SSH_ORIGINAL_COMMAND: None
```

witten:
More very basic testing with borgmatic:

 * [x] interrupting a `borgmatic create` and then re-running it
 * [x] `borgmatic rlist`
 * [x] `borgmatic list`
 * [x] `borgmatic prune`
 * [x] `borgmatic delete`
 * [x] `borgmatic compact`
 * [x] `borgmatic break-lock`
 * [x] `borgmatic key export`

ThomasWaldmann:
maybe i can catch that exception and re-raise a borg exception as borg1 did. Von meinem iPhone gesendetAm 01.09.2024 um 17:00 schrieb Dan Helfman ***@***.***>:﻿
Initial feedback... Running the following on a non-existent repo directory (which borgmatic does to probe whether a configured repo already exists prior to rcreate):
$ ~/Downloads/borg-env/bin/borg rinfo --json --repo ~/Downloads/test.borg 
Results in this (exit code 2):
Local Exception

Error:

BackendDoesNotExist: posixfs storage base path does not exist or is not a directory: /home/witten/Downloads/test.borg

If reporting bugs, please include the following:

  File "/home/witten/Downloads/borg/src/borg/archiver/__init__.py", line 622, in main
                ^^^^^^^^^^^^^^^^^^
  File "/home/witten/Downloads/borg/src/borg/archiver/__init__.py", line 516, in run
    rc = func(args)
         ^^^^^^^^^^
  File "/home/witten/Downloads/borg/src/borg/archiver/_common.py", line 151, in wrapper
    with repository:
  File "/home/witten/Downloads/borg/src/borg/repository.py", line 133, in __enter__
    self.open(exclusive=bool(self.exclusive), lock_wait=self.lock_wait, lock=self.do_lock)
  File "/home/witten/Downloads/borg/src/borg/repository.py", line 180, in open
    self.store.open()
  File "/home/witten/Downloads/borg-env/lib/python3.12/site-packages/borgstore/store.py", line 65, in open
    self.backend.open()
  File "/home/witten/Downloads/borg-env/lib/python3.12/site-packages/borgstore/backends/posixfs.py", line 54, in open
    raise BackendDoesNotExist(
borgstore.backends.errors.BackendDoesNotExist: posixfs storage base path does not exist or is not a directory: /home/witten/Downloads/test.borg

Platform: Linux flux 6.1.99-1-MANJARO #1 SMP PREEMPT_DYNAMIC Mon Jul 15 21:35:02 UTC 2024 x86_64
Linux: Unknown Linux  
Borg: 2.0.0b10.dev80+g74a67749  Python: CPython 3.12.4 msgpack: 1.0.8 fuse: None [pyfuse3,llfuse]
PID: 2588631  CWD: /home/witten/Downloads
sys.argv: ['/home/witten/Downloads/borg-env/bin/borg', 'rinfo', '--json', '--repo', '/home/witten/Downloads/test.borg']
SSH_ORIGINAL_COMMAND: None

witten:
> maybe i can catch that exception and re-raise a borg exception as borg1 did

That'd be great. To be clear, borgmatic does "catch" the Borg error and proceed with the `rcreate`. It's just that the displayed traceback doesn't make for the nicest user experience.

ThomasWaldmann:
I am currently working on manifest.archives.

The archives list in borg 1.x used to be contained in the manifest and that had some good and some bad effects:
- borg locked the repo, loaded manifest (and archives list), created an archive, modified manifest (and archives list), saved manifest, unlocked repo
- bad: has an exclusive lock all the time
- bad: even in this PR could not yet get rid of the lock due to that long read-modify-write operation, although storing chunks to the repo does not need the lock anymore.
- good: loaded a "signed list of archives", modified it on the client (e.g. adding or deleting an archive), saved a "signed list of archives"

This PR:
- archives list is `archives/*` directory in the borgstore, each directory entry is a separate object and points to an archive metadata chunk
- good: no problem adding another object there or deleting an object, no r-m-w needed, no lock needed (todo)
- bad: no signed list of archives anymore (but each archive metadata chunk is signed as usual). removing an archive entry is possible on the repo side without the key. but otoh, an attacker with repo r/w access can do arbitrary damage to the repo anyway, including removing it completely.

ThomasWaldmann:
Yay, just had multiple borg (all running on same machine / same user) share the same repo at the same time (for borg create, mount, list, ... - excluding borg check and compact). :-)

Only worked when I set `BORG_CACHE_IMPL=adhoc` so it doesn't lock the cache (otherwise the cache lock prevents parallel access). Due to the missing files cache in that mode, it was a bit slow, because all files were `A(dded)`.

ThomasWaldmann:
@m3nu did you find something in your experiments with this code?

ThomasWaldmann:
Bug or feature?

borg delete and prune now work in a very "quick and dirty" way: they just remove the deleted archive(s) from the `archives/*` directory in the borgstore.

If one runs `borg compact` afterwards, it notices a lot of unused objects and actually deletes them all (this includes the main archive metadata object, the archived items metadata stream and also the items content chunks, if not otherwise used).

Bug? If one runs `borg check --repair` before `borg compact` AND the manifest object is missing or corrupted, it will rebuild the manifest object and also checks for missing entries in the archives directory. It does that by iterating over all objects in the repo and if it finds these main archive metadata objects, it creates an archives directory entry for them. That means it **UNDELETES** previously deleted archives.

Feature? Guess if you accidentally deleted archives and you want them back, this could be seen as a feature.

witten:
This may be a pain, but this "feature" might be nice to have—if the user explicitly asks for it. So like a `borg check --repair --undelete` flag or something. I have no idea how feasible that is, but I think it'd be a little unexpected if I'm merely `borg check --repair`ing a broken repository and my deleted archives come back from the dead without me asking for that.

ThomasWaldmann:
Rebuilding a missing or corrupted manifest is now much faster, as it won't scan the whole repo for archives anymore.

It only does that scan for `borg check --repair --undelete-archives`.

ThomasWaldmann:
Guess 2024-09-09 would be a nice b10 release date.

If somebody finds some major issue, hold me back. :-)

ThomasWaldmann:
I'll merge this now. b10 tomorrow and I also might work on some other stuff.

callegar:
Some questions about absence of packs/segments.

The restic documentation hints at the importance of packages and on the tunability of the package size where it says (emphasis is mine):
> In certain instances, such as very large repositories (in the TiB range) or very fast upload connections, it is desirable to use larger pack sizes to reduce the number of files in the repository and improve upload performance. **Notable examples are OpenStack Swift and some Google Drive Team accounts, where there are hard limits on the total number of files.**

- With borg moving to an arrangement where the borgstore can be managed by multiple backend and reside on different "providers" (local/cloud) can this become an issue?
- What happens when borg runs out of storage space, but not in terms of storage itself, but of available files?
- Is there provision to bring back packs and if so where would the "packaging" be implemented? In the borgstore backend or "before" the borgstore?
- Would packages let one use a more aggressive (smaller) chunk size? In restic there is a very interesting bug report, where they compare deltas sized by chunk size (with tests based on zpaq).
    - Apparently, the delta size grows rapidly with the chunk size: at 8kB chunks (bup'd default), the deltas are 115% than at 4kB; at 64kB (zpaq's default) they are almost twice as large than at 4kBchunks; at 1MB (restic's default) the deltas are more than 3 times as large than at 4kB. The borg default, at 2MB chunks, makes the deltas 320% the size as with 4kB chunks.
    - Clearly this depends very much on the corpus that is backed up. However, there is a very clear trend: the increase in the delta size slows down for very large chunk sizes, because at a certain point file based deduplication becomes prevalent. Yet, when this happens, chunking gets effectively "switched off" for most files.
    - As the restic developers make clear, the advantage of small chunks largely depends on the efficiency of the indexing approach, because for every chunk you end up paying a price in the indexes which may well hinder the gain on the delta sizes. Specifically, the restic developers at this point clarify that they simply cannot have chunks smaller than 1MB with their current indexing approach. Even the bup's developers say that they might need to move to something somehow larger than 8kB. What is the current situation with the borg2 indexing efficiency?

ThomasWaldmann:
@callegar Please read #8572 and ask all remaining questions there.

PR #8334, opened 2024-08-16
Suppress modified changes for files which weren't actually modified in JSON output.

cr1901:
Right now, if `borg diff` is run with the `--json-lines` parameter (and without `--content-only`, files whose content have _not_ been modified will still be given a `modified` entry of `{"added": 0, "removed": 0, "type": "modified"}`, e.g.:

```
 {"changes": [{"added": 0, "removed": 0, "type": "modified"}, {"new_ctime": "2024-07-22T20:13:59.352959", "old_ctime": "2020-01-06T02:21:13.955255", "type": "ctime"}], "path": ".areca/workspace/.log/areca.12-06-24.log"}
```

Per IRC discussion, this is considered a bug, as the file's content was not actually _changed_, and thus shouldn't be in the `changes` JSON array. This PR fixes the behavior so that JSON entries no longer get a `"type": "modified"` object for files whose contents weren't changed, but `ctime`, `mtime`, etc _were_ changed. E.g.:

```
{"changes": [{"new_ctime": "2024-07-22T20:13:59.352959", "old_ctime": "2020-01-06T02:21:13.955255", "type": "ctime"}], "path": ".areca/workspace/.log/areca.12-06-24.log"}
```

_Text output (without `--json-lines`) is unaffected_; I feel that the  `      0 B       0 B` is more useful for humans looking at output, compared to JSON processors :):

```
      0 B       0 B [ctime: Mon, 2020-01-06 02:21:13 -> Mon, 2024-07-22 20:13:59] .areca/workspace/.log/areca.12-06-24.log
```

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1122,7 +1122,16 @@ class Archiver:
         """Diff contents of two archives"""
 
         def print_json_output(diff, path):
-            print(json.dumps({"path": path, "changes": [j for j, str in diff]}, sort_keys=True, cls=BorgJsonEncoder))
+            def actual_change(j):
+                if j["type"] == "modified":
+                    # It's useful to show 0 added and 0 removed for text output
+                    # but for JSON this is essentially noise. Additionally, the
+                    # JSON key is "changes", and this is not actually a change.
+                    return not (j["added"] == 0 and j["removed"] == 0)
+                else:
+                    return True
+
+            print(json.dumps({"path": path, "changes": [j for j, str in diff if actual_change(j)]}, sort_keys=True, cls=BorgJsonEncoder))
 
         def print_text_output(diff, path):
             print("{:<19} {}".format(' '.join([str for j, str in diff]), path))
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1121,20 +1121,22 @@ class Archiver:
     def do_diff(self, args, repository, manifest, key, archive):
         """Diff contents of two archives"""
 
-        def print_json_output(diff, path):
-            def actual_change(j):
-                if j["type"] == "modified":
-                    # It's useful to show 0 added and 0 removed for text output
-                    # but for JSON this is essentially noise. Additionally, the
-                    # JSON key is "changes", and this is not actually a change.
-                    return not (j["added"] == 0 and j["removed"] == 0)
-                else:
-                    return True
+        def actual_change(j):
+            if j["type"] == "modified":
+                # Added/removed keys will not exist if chunker params differ
+                # between the two archives. Err on the side of caution and assume
+                # a real modification in this case (short-circuiting retrieving
+                # non-existent keys).
+                return not {"added", "removed"} <= j.keys() or not (j["added"] == 0 and j["removed"] == 0)
+            else:
+                # All other change types are indeed changes.
+                return True
 
+        def print_json_output(diff, path):
             print(json.dumps({"path": path, "changes": [j for j, str in diff if actual_change(j)]}, sort_keys=True, cls=BorgJsonEncoder))
 
         def print_text_output(diff, path):
-            print("{:<19} {}".format(' '.join([str for j, str in diff]), path))
+            print("{:<19} {}".format(' '.join([str for j, str in diff if actual_change(j)]), path))
 
         print_output = print_json_output if args.json_lines else print_text_output
 
--- a/src/borg/testsuite/archiver.py
+++ b/src/borg/testsuite/archiver.py
@@ -23,6 +23,7 @@ from datetime import timedelta
 from hashlib import sha256
 from io import BytesIO, StringIO
 from unittest.mock import patch
+from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


src/borg/archiver.py
```diff
@@ -1122,7 +1122,16 @@ def do_diff(self, args, repository, manifest, key, archive):
...
         def print_json_output(diff, path):
-            print(json.dumps({"path": path, "changes": [j for j, str in diff]}, sort_keys=True, cls=BorgJsonEncoder))
+            def actual_change(j):
+                if j["type"] == "modified":
+                    # It's useful to show 0 added and 0 removed for text output
+                    # but for JSON this is essentially noise. Additionally, the
+                    # JSON key is "changes", and this is not actually a change.
+                    return not (j["added"] == 0 and j["removed"] == 0)
+                else:
+                    return True
```
ThomasWaldmann, line 1132: This looks like it could be done with a single boolean expression.

Not sure that text should behave differently than json btw. In the end, it is borg **diff**, so it should show changes, not not-changes.

cr1901: You mean `(j["added"] == j["removed"] == 0)`?

Alright, I'll fix it for text-based output.

ThomasWaldmann: no, i mean 1126..1132.

ThomasWaldmann: Guess this would behave the same:
```
return  j["type"] != "modified" or not (j["added"] == 0 and j["removed"] == 0)
```

cr1901: The exact line I came up with locally :P. Still need to fix the tests...

cr1901: > Guess this would behave the same:
> 
> ```
> return  j["type"] != "modified" or not (j["added"] == 0 and j["removed"] == 0)
> ```

This one-liner ended up not being possible; if comparing two archives with different chunking params, the `added` and `removed` keys won't exist. I went with:

```py
if j["type"] != "modified":
    return True
else:
    # Added/removed keys will not exist if chunker params differ
    # between the two archives. Err on the side of caution and assume
    # a real modification in this case (short-circuiting retrieving
    # non-existent keys).
    return not {"added", "removed"} <= j.keys() or not (j["added"] == 0 and j["removed"] == 0)
```

Not clear to me it's worth the one-liner, but if you can massage the logic a bit, happy to make changes.

src/borg/archiver.py
```diff
@@ -1121,11 +1121,22 @@ def item_to_tarinfo(item, original_path):
...
+                # non-existent keys).
+                return not {"added", "removed"} <= j.keys() or not (j["added"] == 0 and j["removed"] == 0)
+
         def print_json_output(diff, path):
-            print(json.dumps({"path": path, "changes": [j for j, str in diff]}, sort_keys=True, cls=BorgJsonEncoder))
+            print(json.dumps({"path": path, "changes": [j for j, str in diff if actual_change(j)]}, sort_keys=True, cls=BorgJsonEncoder))
 
         def print_text_output(diff, path):
-            print("{:<19} {}".format(' '.join([str for j, str in diff]), path))
+            print(diff)
```
ThomasWaldmann, line 1138: Was that for debugging while developing this?

src/borg/archiver.py
```diff
@@ -1121,11 +1121,22 @@ def item_to_tarinfo(item, original_path):
     def do_diff(self, args, repository, manifest, key, archive):
         """Diff contents of two archives"""
 
+        def actual_change(j):
+            if j["type"] != "modified":
+                return True
```
ThomasWaldmann, line 1126: Maybe add a comment here, guess without one this is not clear enough and may rather seem non-sensical:

"not modified" -> "yes, actual change!"

src/borg/testsuite/archiver.py
```diff
@@ -4562,6 +4565,15 @@ def do_asserts(output, can_compare_ids, content_only=False):
             change = '0 B' if can_compare_ids else '{:<19}'.format('modified')
             self.assert_line_exists(lines, f"{change}.*input/empty")
 
+            # Do not show a 0 byte change for a file whose contents weren't
```
ThomasWaldmann, line 4568: Not a native speaker: is "whose" correct here or would that only be used for persons?

src/borg/testsuite/archiver.py
```diff
@@ -4610,6 +4622,16 @@ def get_changes(filename, data):
             # File unchanged
             assert not any(get_changes('input/file_unchanged', joutput))
 
+            # Do not show a 0 byte change for a file whose contents weren't
```
ThomasWaldmann, line 4625: same here.

src/borg/archiver.py
ThomasWaldmann: Yeah, keep it like that, gets too long/complicated as a one-liner.

src/borg/testsuite/archiver.py
cr1901: Interesting question. As a native speaker, I never thought of it.

According to Wiki, it's [valid](https://en.wikipedia.org/wiki/Inanimate_whose), but ppl don't like it. Two alternate constructions:

* Do not show a 0 byte change for a file of which the contents weren't...
* Do not show a 0 byte change for a file the contents of which weren't...

(I prefer inanimate whose, but if forced to choose between the above two, I choose the former)

src/borg/archiver.py
cr1901: I can reverse the `if`-`else` branches as well to make things more obvious, which in retrospect probably makes more sense.

src/borg/archiver.py
cr1901: Indeed... oops :D.

ThomasWaldmann:
@cr1901 OK, guess this is ready to merge?

ThomasWaldmann:
@cr1901 Thanks!

ThomasWaldmann:
Guess a forward port to master branch would be good.

1.2-maint will be phased out in a while, so a backport to there is optional.

ThomasWaldmann:
Will do the port to master now.

PR #8337, opened 2024-08-16
Add quarterly pruning strategy.

### LINKED ISSUE #4750: prune option keep quarterly

Yes

Question / Enhancement

1.1.9

Debian stretch

When using prune command, I think would be useful a retention option (`--keep-x`) between monthly and yearly, so I tought to something like `--keep-quarterly`.
Furthermore I would suggest, if it's not too complicated, a way that could make the retention policy even more customizable: use a flexible syntax like `--keep-n# number` where n is a number and # is a character in the current set (h,d,m,y). I'll make a few examples that are worth 10 explanations.
`--keep-1w` is the same as `--keep-weekly`.
`--keep-2w 3` means keep one backup every 14 days (two weeks) for a max of 3
`--keep-4m` would mean `--keep-quarterly`
`--keep-2d 5` means keep one backup every two days for a maximum of 5.
Basically the syntax is the same of the argument `--keep-within`.
Here is a simple use case. Let's say I want to backup my music folder monthly, but I don't need all the monthly backups, for example because I want to save only the big changes in my musical attitudes (switching from blues to rap let's say), then I would use `--keep-quarterly`.

Probably I wrote too much. The point is that if we use the same simple syntax used in `--keep-within` command, we could achieve a great flexibility in the retention policy.

cr1901:
This implements and would close #4750 as described. It works for my use case, but is not ready to be merged due to no tests and docs, and there are subtleties to be addressed wrt [what a quarter year actually means](https://en.wikipedia.org/wiki/Calendar_year#Quarter_year).

To do:
* [x] Split into `--keep-3monthly` (using the same logic `--keep-monthly` does) and `--keep-13weekly` (using the ISO 8601 week-based year), maybe get rid of the latter entirely.
* [x] Tests
* [x] Docs

### DIFF
--- a/docs/misc/prune-example.txt
+++ b/docs/misc/prune-example.txt
@@ -100,3 +100,27 @@ example simple. They all work in basically the same way.
 
 The weekly rule is easy to understand roughly, but hard to understand in all
 details. If interested, read "ISO 8601:2000 standard week-based year".
+
+The 13weekly and 3monthly rules are two different strategies for keeping one
+every quarter of a year. There are `multiple ways` to define a quarter-year;
+borg prune recognizes two:
+
+* --keep-13weekly keeps one backup every 13 weeks using ISO 8601:2000's
+  definition of the week-based year. January 4th is always included in the
+  first week of a year, and January 1st to 3rd may be in week 52 or 53 of the
+  previous year. Week 53 is also in the fourth quarter of the year.
+* --keep-3monthly keeps one backup every 3 months. January 1st to
+  March 31, April 1st to June 30th, July 1st to September 30th, and October 1st
+  to December 31st form the quarters.
+
+If the subtleties of the definition of a quarter year don't matter to you, a
+short summary of behavior is:
+
+* --keep-13weekly favors keeping backups at the beginning of Jan, Apr, July,
+  and Oct.
+* --keep-3monthly favors keeping backups at the end of Dec, Mar, Jun, and Sept.
+* Both strategies will have some overlap in which backups are kept.
+* The differences are negligible unless backups considered for deletion were
+  created weekly or more frequently.
+
+.. _multiple ways: https://en.wikipedia.org/wiki/Calendar_year#Quarter_year
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -4675,9 +4675,13 @@ class Archiver:
         the local timezone, and weeks go from Monday to Sunday. Specifying a
         negative number of archives to keep means that there is no limit. As of borg
         1.2.0, borg will retain the oldest archive if any of the secondly, minutely,
-        hourly, daily, weekly, monthly, or yearly rules was not otherwise able to meet
-        its retention target. This enables the first chronological archive to continue
-        aging until it is replaced by a newer archive that meets the retention criteria.
+        hourly, daily, weekly, monthly, quarterly, or yearly rules was not otherwise
+        able to meet its retention target. This enables the first chronological archive
+        to continue aging until it is replaced by a newer archive that meets the
+        retention criteria.
+
+        The ``--keep-13weekly`` and ``--keep-3monthly`` rules are two different
+        strategies for keeping archives every quarter year.
 
         The ``--keep-last N`` option is doing the same as ``--keep-secondly N`` (and it will
         keep the last N archives under the assumption that you do not create more than one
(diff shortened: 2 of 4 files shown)


src/borg/archiver.py
```diff
@@ -4715,6 +4715,8 @@ def define_borg_mount(parser):
                                help='number of weekly archives to keep')
         subparser.add_argument('-m', '--keep-monthly', dest='monthly', type=int, default=0,
                                help='number of monthly archives to keep')
+        subparser.add_argument('-q', '--keep-quarterly', dest='quarterly', type=int, default=0,
```
ThomasWaldmann, line 4718: Not sure if we want to "waste" "-q" for this, especially considering we'll need 2 different quarterly options and not both can have "-q".

ThomasWaldmann: The long option name should be something making clear what one gets, considering there are different "quarters" according to WP.

ThomasWaldmann (commented):
Extending the tests would be good, so this code gets executed and tested.

src/borg/archiver.py
cr1901: Should `--keep-13weekly` and `--keep-3monthly` be mutually-exclusive arguments? I don't think it makes sense to supply both args to one invocation, but I want a second opinion.

ThomasWaldmann: Yes, mutually exclusive.

src/borg/helpers/misc.py
```diff
@@ -59,14 +59,31 @@ def quarterly_period_func(a):
         return (year, 4)
 
 
+def quarterly_3monthly_period_func(a):
+    lt = to_localtime(a.ts)
+    if lt.month <= 3:
+        # 1-1 to 3-31
+        return (lt.year, 1)
+    elif lt.month > 3 and lt.month <= 6:
```
ThomasWaldmann, line 67: 4 <= lt.month <= 6

also fix the other places in the same style.

docs/usage/prune.rst.inc
```diff
@@ -17,7 +17,7 @@ borg prune
...
     |                                                                             | ``REPOSITORY``                                    | repository to prune                                                                                                               |
     +-----------------------------------------------------------------------------+---------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------+
-    | **optional arguments**                                                                                                                                                                                                                                              |
+    | **options**                                                                                                                                                                                                                                                         |
```
ThomasWaldmann, line 20: you don't need to include the generated file(s) into the PR.

i'll generate them as part of the release process.

cr1901:
Feedback addressed, tests are coming along and should be done soon (30 mins?).

docs/misc/prune-example.txt
```diff
@@ -100,3 +100,27 @@ example simple. They all work in basically the same way.
...
+* --keep-3monthly keeps one backup every 3 months. January 1st to
+  March 31, April 1st to June 30th, July 1st to September 30th, and October 1st
+  to December 31st form the quarters.
+
+If the subtleties of the definition of a quarter year doesn't matter to you,
+a short summary of behavior is:
+
+* --keep-13weekly favors keeping backups at the beginning of Jan, Apr, July,
+  and Oct.
+* --keep-3montly favors keeping backups at the end of Dec, Mar, Jun, and Sept.
```
ThomasWaldmann, line 121: typo: 3montly

cr1901: Never fails... will correct/rebase when I push tests.

docs/misc/prune-example.txt
```diff
@@ -100,3 +100,27 @@ example simple. They all work in basically the same way.
...
+
+* --keep-13weekly keeps one backup every 13 weeks using ISO 8601:2000's
+  definition of the week-based year. January 4th is always included in the
+  first week of a year, and January 1st to 3rd may be in week 52 or 53 of the
+  previous year. Week 53 is also in the fourth quarter of the year.
+* --keep-3monthly keeps one backup every 3 months. January 1st to
+  March 31, April 1st to June 30th, July 1st to September 30th, and October 1st
+  to December 31st form the quarters.
+
+If the subtleties of the definition of a quarter year doesn't matter to you,
```
cr1901, line 116: Should be "If the subtleties... **don't** matter". I'll push when I wake up. Hopefully won't require running CI :P.

cr1901:
@ThomasWaldmann Okay to merge if you don't have additional changes requested.

ThomasWaldmann:
Same here: port to master required after merge, port to 1.2-maint is very optional, considering this is new functionality and 1.2-maint will be phased out in a while.

ThomasWaldmann:
@cr1901 Thanks!

ThomasWaldmann:
I'll do the master port now.

Issue #8361, opened 2024-09-08
Update repository URL in docs to use new syntax


AfroThundr3007730:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

Issue (docs)

#### Your borg version (borg -V).

borg 1.2.4

## Describe the problem you're observing.

In the docs page [*Central repository server with Ansible or Salt*](https://borgbackup.readthedocs.io/en/stable/deployment/central-backup-server.html), it's mentioned how to use `cd` in the `authorized_keys`, allowing relative paths in the repository URLs and hiding the actual location details from the connecting users.

That page shows the SCP syntax for specifying a relative repository path, but newer borg versions like 2.0 would need the URL syntax. Tucked away in the [page on repository URLs](https://borgbackup.readthedocs.io/en/stable/usage/general.html#repository-urls) I found we need to use `ssh://user@host/./repo` to use a relative path.

Can we update the other docs to show this new syntax? I'm thinking if the URL syntax is the preferred method, all the docs should just be updated to use it everywhere, and kind of pretend the old syntax no longer exists.

ThomasWaldmann:
Can you make a PR? It is the .rst file(s) that need updating.

AfroThundr3007730:
Found a few spots using the SCP syntax. I'm not sure if you want it removed completely from the repository URLs page or to keep it for reference. I just adjusted that one to note the deprecation.

ThomasWaldmann:
PR is fine.

For updating the 1.2-maint branch, the changset(s) need to be cherry picked and a new PR made (after the first PR is finished/merged).

AfroThundr3007730:
Right, master still needs fixing.

PR #8371, opened 2024-09-10
Update repository URLs in docs to use new syntax

AfroThundr3007730:
Fixes #8361

PR #8377, opened 2024-09-12
Update repository URLs in docs to use new syntax

AfroThundr3007730:
Fixes: #8361

See Also: #8371


ThomasWaldmann (approved):
looks good.

guess the other changes in 1.x-maint weren't needed here as I already removed the scp-style URL syntax docs.

ThomasWaldmann:
Thanks!

Issue #8381, opened 2024-09-15
Windows build of master currently failing


sothix:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).

Master

## Describe the problem you're observing.

During the build it fails to compile:
```
Collecting cryptography<43.0.0 (from borgstore~=0.0.1->borgbackup==2.0.0b11.dev3+gc278a8e8)
Using cached cryptography-42.0.8.tar.gz (671 kB)
Installing build dependencies ... error
...
        gcc -DNDEBUG -g -fwrapv -O3 -Wall -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -O3 -march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-st
rong -O3 -DFFI_BUILDING=1 -IC:/msys64/ucrt64/include/python3.11 -c src/c/_cffi_backend.c -o build/temp.mingw_x86_64_ucrt-cpython-311/src/c/_cffi_backend.o
        src/c/_cffi_backend.c: In function 'b_do_dlopen':
        src/c/_cffi_backend.c:4568:40: error: passing argument 1 of 'PyUnicode_AsWideChar' from incompatible pointer type [-Wincompatible-pointer-types]
         4568 |             sz1 = PyUnicode_AsWideChar((PyUnicodeObject *)filename_unicode,
              |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
              |                                        |
              |                                        PyUnicodeObject *
        In file included from C:/msys64/ucrt64/include/python3.11/Python.h:51,
                         from src/c/_cffi_backend.c:2:
        C:/msys64/ucrt64/include/python3.11/unicodeobject.h:299:15: note: expected 'PyObject *' {aka 'struct _object *'} but argument is of type 'PyUnicodeObject *'
          299 |     PyObject *unicode,          /* Unicode object */
              |     ~~~~~~~~~~^~~~~~~
        error: command 'C:\\msys64\\ucrt64\\bin/gcc.EXE' failed with exit code 1
        [end of output]
```

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

Yes 
new install of MSYS2.
```
[...]

Issue #8264, opened 2024-07-02
windows: github CI broken 


ThomasWaldmann:
https://github.com/borgbackup/borg/actions/runs/9766803308/job/26960577084?pr=8263

ThomasWaldmann:
```
Run ./scripts/msys2-build
Obtaining file:///D://a/borg/borg
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Checking if build backend supports build_editable: started
  Checking if build backend supports build_editable: finished with status 'done'
  Getting requirements to build editable: started
  Getting requirements to build editable: finished with status 'done'
  Preparing editable metadata (pyproject.toml): started
  Preparing editable metadata (pyproject.toml): finished with status 'done'
Requirement already satisfied: msgpack<=1.1.0,>=1.0.3 in d:/a/_temp/msys64/ucrt64/lib/python3.11/site-packages (from borgbackup==1.4.0rc2.dev13+g76be26d4.d20240702) (1.0.6)
Requirement already satisfied: packaging in d:/a/_temp/msys64/ucrt64/lib/python3.11/site-packages (from borgbackup==1.4.0rc2.dev13+g76be26d4.d20240702) (24.1)
Building wheels for collected packages: borgbackup
  Building editable for borgbackup (pyproject.toml): started
  Building editable for borgbackup (pyproject.toml): still running...
  Building editable for borgbackup (pyproject.toml): finished with status 'done'
  Created wheel for borgbackup: filename=borgbackup-1.4.0rc2.dev13+g76be26d4.d20240702-0.editable-cp311-cp311-mingw_x86_64_ucrt.whl size=7625 sha256=d72c0ff4aa6b9bf80bc940bdf50da0a4b92bc62081ca0697e91f48b5774a3cc8
  Stored in directory: D:/a/_temp/msys64/tmp/pip-ephem-wheel-cache-mczy8ud2/wheels/bb/6d/6e/fdef5aca9f58a65a983afbc55f7dae75316edcc2fbf171af32
Successfully built borgbackup
Installing collected packages: borgbackup
Successfully installed borgbackup-1.4.0rc2.dev13+g76be26d4.d20240702
171 INFO: PyInstaller: 6.3.0
171 INFO: Python: 3.11.9
171 INFO: Platform: Windows-10-10.0.20348-SP0
171 INFO: Extending PYTHONPATH with paths
['D:/a/borg/borg/src', 'D:/a/borg/borg']
[...]

PR #8382, opened 2024-09-15
Fix for #8264 Windows CI broken and #8381 Windows build failing

sothix:
Will currently upgrade PIP from the package 24.1.2 to 24.2 and solves #8264 
python-paramiko solves #8381

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,7 +196,7 @@ jobs:
 
   windows:
 
-    if: false  # build is broken, thus disabled, see #8264
+    if: true  # build enabled
     runs-on: windows-latest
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Thanks for the PR!

I remember disabling the windows CI because it was broken, so that has to be re-enabled.

Maybe that will be enough to trigger the important parts of the CI, otherwise guess you need to do a tmp modification to some other file (e.g. setup.py).

ThomasWaldmann:
OK, it almost works, just 1 test failing that doesn't fail on linux / macOS.

ThomasWaldmann:
@sothix thanks for the PR, we have working windows CI again! \o/

PR #8403, opened 2024-09-21
chunks index caching

ThomasWaldmann:
borg compact uses ChunkIndex (a specialized, memory-efficient data structure), so it needs less memory now. Also, it saves that chunks index to cache/chunks in the repository.

When the chunks index is needed, it is first tried to get it from cache/chunks and only fall back to building the chunks index via repository.list() (which can be rather slow).

borg check --repair currently just invalidates the chunks cache.

borg create updates the chunks cache.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -50,7 +50,7 @@ from .patterns import PathPrefixPattern, FnmatchPattern, IECommand
 from .item import Item, ArchiveItem, ItemDiff
 from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
 from .remote import RemoteRepository, cache_if_remote
-from .repository import Repository, NoManifestError
+from .repository import Repository, NoManifestError, StoreObjectNotFound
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
Code is a bit less pretty now, but more efficient. Also less stats.

mirko:
Is this (also) addressing the issue of `listdir()` being called for every (sub-)directory in `data/` for every `borg create`-run?

ThomasWaldmann:
yesVon meinem iPhone gesendetAm 24.09.2024 um 01:13 schrieb Mirko Vogt ***@***.***>:﻿
Is this (also) addressing the issue of listdir() being called for every (sub-)directory in data/ for every borg create-run?

ThomasWaldmann:
@mirko merged this.

also found another issues that it was doing one full repo.list too much, PR incoming soon.

so, master branch should be quite a bit faster now.

only check and compact are expected to always do the repository.list(), just to be on the safe side and not rely on caches.

PR #8408, opened 2024-09-23
FAQ: Why is backing up an unmodified FAT filesystem slow on Linux?

helmutg:


docs/faq.rst
```diff
@@ -357,6 +357,16 @@ to change them.
...
 
+Why is backing up an unmodified FAT filesystem slow on Linux?
+-------------------------------------------------------------
+
+By default, the files cache used by BorgBackup considers the inode of files.
+When an inode changes compared to the last backup, it hashes the file again.
+The ``vfat`` kernel driver does not produce stable inodes by default. One way
+to achieve stable inodes is mounting the filesystem using ``nfs=nostale_ro``.
+Doing so implies mounting the filesystem read-only. Another option is to not
+consider inodes in the files cache by passing ``--files-cache=ctime,size``.
```
ThomasWaldmann, line 368: Thanks for this FAQ update!

I have only one small nitpick: maybe rather say "inode number".

Changing the number / having an unstable number is something different than changing the inode (contents).

ThomasWaldmann:
Thanks!

Will need also a backport to 1.4-maint (and optionally to 1.2-maint) to get into next stable releases.

PR #8460, opened 2024-10-05
Suppress modified changes for files which weren't actually modified in JSON output.

ThomasWaldmann:
Port of #8334.

### DIFF
--- a/src/borg/testsuite/archiver/diff_cmd_test.py
+++ b/src/borg/testsuite/archiver/diff_cmd_test.py
@@ -173,7 +173,9 @@ def test_basic_functionality(archivers, request):
         unexpected = {"type": "modified", "added": 0, "removed": 0}
         assert unexpected not in get_changes("input/file_touched", joutput)
         if not content_only:
-            assert {"ctime", "mtime"}.issubset({c["type"] for c in get_changes("input/file_touched", joutput)})
+            # on win32, ctime is the file creation time and does not change.
+            expected = {"mtime"} if is_win32 else {"mtime", "ctime"}
+            assert expected.issubset({c["type"] for c in get_changes("input/file_touched", joutput)})
         else:
             # And if we're doing content-only, don't show the file at all.
             assert not any(get_changes("input/file_touched", joutput))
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
@cr1901 Due to changes in master, these changesets didn't apply well and I had to make some changes, please review.

src/borg/testsuite/archiver/diff_cmd_test.py
```diff
@@ -150,6 +169,15 @@ def get_changes(filename, data):
         # File unchanged
         assert not any(get_changes("input/file_unchanged", joutput))
 
+        # Do not show a 0 byte change for a file whose contents weren't modified.
+        unexpected = {"type": "modified", "added": 0, "removed": 0}
+        assert unexpected not in get_changes("input/file_touched", joutput)
+        if not content_only:
+            assert {"mtime"}.issubset({c["type"] for c in get_changes("input/file_touched", joutput)})
```
cr1901, line 176: If `mtime` changes due to `touch`, won't `ctime` change as well? This isn't wrong, but I was checking `ctime` as well because I expected it to change.

ThomasWaldmann: ctime on windows is file creation time. guess it won't change as long as the file exists.

on posix it is inode change time and it changes for all metadata or data changes.

cr1901:
>these changesets didn't apply well and I had to make some changes

That's fine, because it looks like all but one of the test code changes applied correctly. I marked the single difference in a comment asking for clarification.

All the non-test code looks fine too.  Since CI is green for (almost) identical tests, this LGTM pending clarification.

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: How do you feel about this?

```python
expected_changes = "mtime"

if os.name in ('posix',):
    expected_changes += "ctime"
```

ThomasWaldmann: good idea. we have some platformflags, yeah, guess i could check for "not windows".

ThomasWaldmann: done.

ThomasWaldmann: CI says it worked. \o/

PR #8483, opened 2024-10-21
docs/FAQ: Clean up entries regarding SSH settings

mrichtarsky:
Link to recommended ones

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -602,31 +602,25 @@ You will also get this error if you try to access a repository with a key that u
 We recommend upgrading to the latest stable version and trying again. We are sorry. We should have thought about forward
 compatibility and implemented a more helpful error message.
 
-Why does Borg extract hang after some time?
--------------------------------------------
+Why am I seeing idle borg serve processes on the repo server?
+-------------------------------------------------------------
 
-When I do a ``borg extract``, after a while all activity stops, no cpu usage,
-no downloads.
+Please see the next question.
 
-This may happen when the SSH connection is stuck on server side. You can
-configure SSH on client side to prevent this by sending keep-alive requests,
-for example in ~/.ssh/config:
+Why does Borg disconnect or hang when backing up to a remote server?
+--------------------------------------------------------------------
 
-::
+Communication with the remote server (using an ssh: repo URL) happens via an SSH
+connection. This can lead to some issues that would not occur during a local backup:
 
-    Host borg.example.com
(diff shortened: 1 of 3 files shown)


docs/faq.rst
```diff
@@ -602,31 +602,22 @@ You will also get this error if you try to access a repository with a key that u
...
-        # Client kills connection after 3*30 seconds without server response:
-        ServerAliveInterval 30
-        ServerAliveCountMax 3
+Why does Borg disconnect or hang when backing up to a remote server?
+--------------------------------------------------------------------
 
-You can also do the opposite and configure SSH on server side in
-/etc/ssh/sshd_config, to make the server send keep-alive requests to the client:
+Communication with the remote server (using an ssh: repo URL) happens via the SSH
+protocol. This can lead to some issues that would not occur during a local backup:
```
ThomasWaldmann, line 611: ... via a ssh connection.

(it's not like borg is talking ssh protocol somehow, there is a pair of ssh-sshd processes doing that)

docs/faq.rst
```diff
@@ -602,31 +602,22 @@ You will also get this error if you try to access a repository with a key that u
...
 
-::
+- Since Borg does not send data all the time, the connection may get closed, leading
+  to errors like "connection closed by remote".
```
ThomasWaldmann, line 614: Not sure if that issue matches that error msg.

mrichtarsky: That was the message from the original FAQ entry, that's why I kept it. In my setup, there was a different one but I don't have it anymore (also, I used the `socat` pull mode so not sure if my error message would have been representative anyway)

mrichtarsky:
@ThomasWaldmann  Could you please merge so I can downport?

docs/faq.rst
```diff
@@ -642,32 +633,10 @@ could try to work around:
...
-    # /etc/ssh/sshd_config on borg repo server - kill connection to client
-    # after ClientAliveCountMax * ClientAliveInterval seconds with no response
-    ClientAliveInterval 20
-    ClientAliveCountMax 3
-
-If you have multiple borg create ... ; borg create ... commands in a already
-serialized way in a single script, you need to give them ``--lock-wait N`` (with N
-being a bit more than the time the server needs to terminate broken down
-connections and release the lock).
+Please see :ref:`disconnect_hang`
```
ThomasWaldmann, line 639: can you move this to right above the disconnect_hang label?

maybe one could just say "see below" then

PR #8510, opened 2024-11-01
docs: Fix grammar in borg mount docs

fantasya-pbem:
"clean" as adverb does not have suffix -ly.

Source: https://dict.leo.org/grammatik/englisch/adjv_usage_flat.php?lang=de#id=5.3.1.b

### DIFF
--- a/src/borg/archiver/mount_cmds.py
+++ b/src/borg/archiver/mount_cmds.py
@@ -120,7 +120,7 @@ class MountMixIn:
         Unmounting in these cases could cause an active rsync or similar process
         to delete data unintentionally.
 
-        When running in the foreground ^C/SIGINT unmounts cleanly, but other
+        When running in the foreground ^C/SIGINT unmounts clean, but other
         signals or crashes do not.
         """
         )


ThomasWaldmann:
Thanks!

jdchristensen:
"cleanly" is correct in this situation, so this should be reverted.  For example, there are lots of examples [here](https://dictionary.cambridge.org/dictionary/english/cleanly).

jdchristensen:
The link provided in the original description also gives a similar example:  "The hull of the boat cut cleanly through the placid water."

fantasya-pbem:
I did not say that "cleanly" is not a correct adverbial form. But the adjective "clean" does have an adverbial form"clean" that is, in my opinion, the correct adverb here. "Cut cleanly" is not the same clean as in "wipe clean".

jdchristensen:
I agree that "wipe clean" is correct grammar, but it doesn't match the situation in "unmount cleanly".  The change in this patch is definitely wrong.  Putting it another way, one would never say that we "clean unmount a filesystem".  We say that we "cleanly unmount a filesystem", or in other words, that it "unmounts cleanly".

But the original text is slightly awkward, so I would propose:  "When running in the foreground, ^C/SIGINT cleanly unmounts the filesystem, but other signals or crashes do not."

(Not that this guarantees anything, but I'm a native speaker.)

fantasya-pbem:
Native speaker wins!

ThomasWaldmann:
Fixed by #8519.

PR #8515, opened 2024-11-02
soft delete / undelete

ThomasWaldmann:


src/borg/archive.py
```diff
@@ -499,8 +500,9 @@ def __init__(
             self.tags = set()
         else:
             if name_is_id:
-                # we also go over the manifest here to avoid quick&dirty deleted archives
-                info = self.manifest.archives.get_by_id(name)
+                # we also go over the manifest here to avoid quick&dirty deleted archives,
```
ThomasWaldmann, line 503: s/q&d/soft-/

src/borg/archiver/undelete_cmd.py
```diff
@@ -0,0 +1,90 @@
...
+
+        undelete_epilog = process_epilog(
+            """
+        This command undeletes archives in the repository.
+
+        Important: Undeleting archives is only possible while they are in the
+        "soft-deleted" state. Once ``borg compact`` has run, they will be
+        completely gone and undelete is not possible anymore.
+
+        When in doubt, use ``--dry-run --list`` to see what would be deleted.
```
ThomasWaldmann, line 66: undeleted

src/borg/archiver/check_cmd.py
```diff
@@ -182,9 +180,11 @@ def build_parser_check(self, subparsers, common_parser, mid_common_parser):
 
         If ``--repair --undelete-archives`` is given, Borg will scan the repository
```
ThomasWaldmann, line 181: guess this should be renamed to --find-lost-archives.

if users delete archives, they can easily `borg undelete` these.
check does something different: with this option it finds archives for which we have lost the entry in archives/ directory.

src/borg/archive.py
ThomasWaldmann: fixed

src/borg/archiver/undelete_cmd.py
ThomasWaldmann: fixed

src/borg/archiver/check_cmd.py
ThomasWaldmann: fixed

src/borg/archiver/check_cmd.py
```diff
@@ -37,8 +37,6 @@ def do_check(self, args, repository):
             )
         if args.repair and args.max_duration:
             raise CommandError("--repair does not allow --max-duration argument.")
-        if args.undelete_archives and not args.repair:
```
PhrozenByte, line 40: I **assume** that we also need the following here, right? If not, also remove the matching sentence in my docs suggestion below.

```py
if args.repo_only and args.find_lost_archives:
    raise CommandError("--repository-only contradicts the --find-lost-archives option.")
```

src/borg/archiver/prune_cmd.py
```diff
@@ -215,7 +215,11 @@ def build_parser_prune(self, subparsers, common_parser, mid_common_parser):
         The prune command prunes a repository by deleting all archives not matching
         any of the specified retention options.
 
-        Important: Repository disk space is **not** freed until you run ``borg compact``.
+        Important:
+
+        - Repository disk space is **not** freed until you run ``borg compact``.
```
PhrozenByte, line 220:
```suggestion
        - The prune command will only mark archives for deletion
          ("soft-deleted archives"), repository disk space is **not**
          freed until you run ``borg compact``.
```

src/borg/archiver/delete_cmd.py
```diff
@@ -64,8 +64,11 @@ def build_parser_delete(self, subparsers, common_parser, mid_common_parser):
             """
         This command deletes archives from the repository.
 
-        Important: When deleting archives, repository disk space is **not** freed until
-        you run ``borg compact``.
+        Important:
+
+        - Repository disk space is **not** freed until you run ``borg compact``.
```
PhrozenByte, line 69:
```suggestion
        - The delete command will only mark archives for deletion
          ("soft-deleted archives"), repository disk space is **not**
          freed until you run ``borg compact``.
```

src/borg/archiver/delete_cmd.py
```diff
@@ -64,8 +64,11 @@ def build_parser_delete(self, subparsers, common_parser, mid_common_parser):
...
+        - You can use ``borg undelete`` to undelete archives, but only until
+          you run ``borg compact``.
```
PhrozenByte, line 71:
```suggestion
        - You can use ``borg undelete`` to undelete soft-deleted archives,
          but only until you run ``borg compact``.
```

src/borg/archiver/prune_cmd.py
```diff
@@ -215,7 +215,11 @@ def build_parser_prune(self, subparsers, common_parser, mid_common_parser):
...
+        - You can use ``borg undelete`` to undelete archives, but only until
+          you run ``borg compact``.
```
PhrozenByte, line 222:
```suggestion
        - You can use ``borg undelete`` to undelete soft-deleted archives,
          but only until you run ``borg compact``.
```

src/borg/archiver/check_cmd.py
```diff
@@ -180,11 +178,13 @@ def build_parser_check(self, subparsers, common_parser, mid_common_parser):
...
+        If ``--repair --find-lost-archives`` is given, Borg will scan the repository
         for archive metadata and if it finds some where no corresponding archives
-        directory entry exists, it will create the entries. This is basically undoing
-        ``borg delete archive`` or ``borg prune ...`` commands and only possible before
-        ``borg compact`` would remove the archives' data completely.
+        directory entry exists (neither a normal not-deleted archive entry nor a
+        soft-deleted archive entry), it will create the entries as not-deleted.
+        This will make archives reappear for which the directory entry was lost.
+        This is only possible before ``borg compact`` would remove the archives'
+        data completely.
```
PhrozenByte, line 187: Since `--find-lost-archives` is no longer limited to `--repair`, this should be explained in the general (non-repair mode) section.

How about adding this to the third paragraph (no. 2 in the list of `check`' steps) after the third sentence (i.e. after "This requires reading archive and file metadata, but not data."), also updating the following sentence about `--verify-data`:

```
To scan for archives whose entries were lost from the archive directory, pass
`--find-lost-archives``. It requires reading all data and is hence very time
consuming. To additionally cryptographically verify the file (content) data
integrity pass ``--verify-data``, which is hence even more time consuming.
```

Then how about adding the following as last paragraph of the general section (i.e. after the paragraph about `--verify-data`):

```
Similar to ``--verify-data``, the ``--find-lost-archives`` option will also
scan the whole repository, but tells Borg to search for lost archive metadata.
If Borg encounters any archive metadata that doesn't match with an
existing archive entry (both normal/not-deleted archives, and soft-deleted
archives), it means that an archive entry was lost from the directory. Unless
``borg compact`` is called, these archives can be fully restored with
``--repair``. Please note that ``--find-lost-archives`` must read a lot of
data from the repository and is thus very time consuming. You can not use
``--find-lost-archives`` with ``--repository-only``.
```

The paragraph in the repair mode section might then be shortened to the following:

```suggestion
       As explained earlier, ``--find-lost-archives`` causes Borg to scan the
       repository for dangling archive metadata. Together with ``--repair``
       Borg will use that archive metadata to re-create entries in the
       directory. This will make previously lost archives reappear as
       normal/not-deleted archive. Note that this is only possible as long as
       you didn't run ``borg compact``, because ``borg compact`` would remove
       the archives' data completely.
```

PhrozenByte (commented):
Looks very good :+1:

I have some comments and suggestions about the docs, e.g. I recommend being very consistent in always calling them "soft-deleted archives".

Important: I neither did a code review, nor did I test any of the changes.

src/borg/archiver/check_cmd.py
ThomasWaldmann: ah, right! good catch!

src/borg/archiver/prune_cmd.py
ThomasWaldmann: From a user perspective, this phrasing might be confusing. Sounds like there was some tag "delete-this" or some other visible mark, but the archive still being normally visible.

But from a user perspective it is gone (except when used special stuff like the ``--deleted`` flag or a undelete dry-run).

src/borg/archiver/check_cmd.py
ThomasWaldmann: Hmm, applying this many/big changes from a suggestion via the github UI does not work.

I'll fetch the stuff to local repo and will do some updates...

src/borg/archiver/prune_cmd.py
PhrozenByte: Good point. Then let's explain it better. How about the following:

```
**Important:** Borg doesn't delete archives right away with the prune and
delete commands, but rather internally marks them for deletion. Borg generally
assumes these "soft-deleted archives" to not exist and thus also doesn't show
them with the list command (unless the ``--deleted`` option is given). However,
"soft-deleted archives" still take up disk space and can be restored with the
undelete command. To free the disk space, run ``borg compact``; this will
actually remove the archive and all its data from the repository. Consequently,
``borg undelete`` can't undelete archives after ``borg compact`` has been
executed.
```

The exact same explanation (I intentionally mentioned both `delete` and `prune` to emphasize that they work the same) can be used for `delete`.

It's quite verbose, I thus also thought about moving this to `compact`. But I feel like it's more important to explain this here.

PhrozenByte:
What do you think about also updating the `compact` epilog to mention all the things users should know?

```
            Free repository space by deleting unused chunks.

            ``borg compact`` iterates over all chunks of data the repo consists
            of and checks whether they are used by any archive in the archives
            directory. All chunks that aren't referenced by any archive are
            then deleted from the filesystem to free disk space. This most
            notably includes all archives that were previously internally
            marked for deletion ("soft-deleted archives") by the delete and
            prune commands. Running the compact command consequently also
            impedes ``borg undelete`` from undeleting any soft-deleted
            archives. ``borg compact`` also removes not needed junk data from
            previous backups, e.g. data of aborted backups, files that had to
            be skipped due to I/O errors and other issues that arised while
            creating previous backups.

            You usually don't want to run ``borg compact`` after every write
            operation on the repository, but either regularly (e.g. once a
            month, possibly together with ``borg check``) or when disk space
            needs to be freed.

            **Important:** ``borg compact`` doesn't just remove data from
            intentionally deleted archives and junk data, but might in rare
            cases also delete data from archives that weren't intentionally
            deleted, but "lost" due to data corruption on a filesystem or
            hardware level. Such archives could potentially be restored with
            ``borg check --find-lost-archives [--repair]``, but note that
            finding such lost archives is a very time consuming task one
            usually doesn't want to take unless there are signs of lost
            archives (e.g. when seeing fatal errors when creating backups or
            when archives are missing in ``borg list``).

            Differently than borg 1.x, borg2's compact needs the borg key if the repo is
            encrypted.
```

ThomasWaldmann:
I thought about it:

I will remove all user docs of "soft-deleted" again from borg and only explain what's happening in `borg undelete` and/or `borg compact` command, where it is somehow relevant.

Mentioning it all over the place isn't necessarily helpful, explaining it everywhere makes the docs longer than necessary.

src/borg/archiver/check_cmd.py
ThomasWaldmann: fixed.

src/borg/archiver/prune_cmd.py
ThomasWaldmann: Solved differently. See undelete/compact help.

src/borg/archiver/delete_cmd.py
ThomasWaldmann: Solved differently. See undelete/compact help.

src/borg/archiver/prune_cmd.py
ThomasWaldmann: Solved differently. See undelete/compact help.

src/borg/archiver/check_cmd.py
ThomasWaldmann: fixed.

ThomasWaldmann:
@PhrozenByte I did it a bit differently, but hopefully giving all the information that is useful to the user (and not just an implementation detail).

src/borg/archiver/undelete_cmd.py
```diff
@@ -0,0 +1,90 @@
...
+    def build_parser_undelete(self, subparsers, common_parser, mid_common_parser):
+        from ._common import process_epilog, define_archive_filters_group
+
+        undelete_epilog = process_epilog(
+            """
+        This command undeletes archives in the repository.
+
+        Important: Undeleting archives is only possible before compacting.
+        Once ``borg compact`` has run, all disk space occupied only by the
+        deleted archives will be freed and undelete is not possible anymore.
```
PhrozenByte, line 64:
```suggestion
        soft-deleted archives will be freed and undelete is not possible
        anymore.
```

See main review comment.

PhrozenByte (commented):
> I will remove all user docs of "soft-deleted" again from borg and only explain what's happening in borg undelete and/or borg compact command, where it is somehow relevant.
> 
> Mentioning it all over the place isn't necessarily helpful, explaining it everywhere makes the docs longer than necessary.

I understand that you want to shorten the docs and remove unnecessary detail, not that you object documenting this distinction at all, right?

I totally agree with shortening the docs: reading over my suggestions again they indeed were a bit excessive (I have that tendency :laughing:). However, there's a big issue with trying not to mention the distinction at all: At some occurrences Borg must talk about them, because Borg is specifically dealing with them. Therefore the docs must deal with the distinction and the current docs try solving it in two ways: By either calling them "deleted archives", or by not mentioning them at all. Both IMHO is pretty problematic:

- Borg shouldn't call them "deleted archives" - because they simply weren't deleted yet. This IMHO creates way more confusion, because, as a user, I know the common term "deleted" to mean "gone for good". This was fine before, because deletion actually meant deletion (`check --repair` was and is data recovery), but with `undelete` it no longer does.

   Take email software like Thunderbird with default `.mbox` storage in comparison: Completely removing an email is called "Delete" in the UI, even though the email isn't actually deleted from the `.mbox` file, but only after the `.mbox` file is compacted separately. Thunderbird calls that action "Delete" nevertheless, because Thunderbird doesn't allow restoring that email. However, alternatively one can also "Trash" an email, meaning that the email isn't actually deleted, but just moved out of the user's sight. `borg delete` previously only did the first variant, but with `undelete` it rather does the latter - and therefore the term must change as well.

    People know what "delete" means, and if it unexpectedly means something else, it creates confusion - not for you and me, but new users, not knowing yet what Borg is doing there. Thus, *instead of* (i.e. not additionally, just instead of) calling them "deleted archives", better call them "soft-deleted archives": it's more precise, doesn't cause such confusion and gives a nice "heads up" that there must be an easy way to get them back (i.e. `undelete`).

- I agree that the docs don't have to explain what "soft-deleted" means all over the place, especially if nothing special is happening. From an users perspective one expects "archive" (i.e. without an adjective) to mean a normal/not-deleted archive, i.e. the archives one sees with e.g. `list`. Calling them "normal/not-deleted" implies speciality that isn't there, so I agree that it causes confusion to call them anything but just "archive". However, if Borg (also) considers soft-deleted archives, it should mention this - and only then. Otherwise one just can't properly assess (both "not understanding" and "understanding wrong") what Borg will do from the docs.

   For example, in the `check` docs, "archive" meant both normal and soft-deleted archives in the main section, but just normal archive in the repair section (see line comment there). This sends the user on the wrong track within the main section, and prevents the user from knowing what will happen in the repair section, thus causing confusion. A very short and simple clarification is sufficient to completely fix that.

    If you just don't like the term "soft-deletion" I'd like to emphasize that there are alternatives. For example, many UI's use "trash" (`delete`) and "untrash" (`undelete`).

This naturally also means that one must explain what "soft-deleted archive" means at some point - and instead of excessively explaining it all over the place like I did in my previous suggestion, it's sufficient to just use the term "soft-delete" in `prune` and `delete` once respectively, together with the already existing mention of `undelete`. This makes it absolutely clear what "soft-deletion" means and allows users to understand it anywhere else, too. Simple and precise.

PR #8525, opened 2024-11-06
Update docs to better incorporate new undelete command

PhrozenByte:
Follow-up to #8515

Refer to https://github.com/borgbackup/borg/pull/8515#pullrequestreview-2417746853 for details.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -97,6 +97,8 @@ Compatibility notes:
   - removed --save-space option (does not change behaviour)
 - removed --bypass-lock option
 - removed borg config command (only worked locally anyway)
+- compact command now requires access to the borg key if the repo is encrypted
+  or authenticated
 - using --list together with --progress is now disallowed (except with --log-json), #7219
 - the --glob-archives option was renamed to --match-archives (the short option
   name -a is unchanged) and extended to support different pattern styles:
--- a/src/borg/archiver/compact_cmd.py
+++ b/src/borg/archiver/compact_cmd.py
@@ -127,7 +127,7 @@ class ArchiveGarbageCollector:
             logger.warning(f"{len(self.reappeared_chunks)} previously missing objects re-appeared!" + run_repair)
             set_ec(EXIT_WARNING)
 
-        logger.info("Cleaning archives directory from deleted archives...")
+        logger.info("Cleaning archives directory from soft-deleted archives...")
         archive_infos = self.manifest.archives.list(sort_by=["ts"], deleted=True)
         for archive_info in archive_infos:
             name, id, hex_id = archive_info.name, archive_info.id, bin_to_hex(archive_info.id)
@@ -176,16 +176,17 @@ class CompactMixIn:
             Free repository space by deleting unused chunks.
 
             borg compact analyzes all existing archives to find out which repository
-            objects are actually used (referenced). It then removes all unused objects
-            to free repository space.
+            objects are actually used (referenced). It then deletes all unused objects
+            from the repository to free space.
 
             Unused objects may result from:
 
             - borg delete or prune usage
-            - interrupted backups (maybe retry the backup first before running compact!)
+            - interrupted backups (maybe retry the backup first before running compact)
             - backup of source files that had an I/O error in the middle of their contents
-              and that were skipped due to this.
-            - corruption of the repository (e.g. the archives directory having lost entries)
+              and that were skipped due to this
+            - corruption of the repository (e.g. the archives directory having lost
+              entries, see notes below)
 
             You usually don't want to run ``borg compact`` after every write operation, but
             either regularly (e.g. once a month, possibly together with ``borg check``) or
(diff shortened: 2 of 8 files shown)


docs/changes.rst
```diff
@@ -97,6 +97,8 @@ Compatibility notes:
   - removed --save-space option (does not change behaviour)
 - removed --bypass-lock option
 - removed borg config command (only worked locally anyway)
+- compact command now requires access to the borg key if the repo is encrypted,
+  thus requiring a client to initiate compaction
```
ThomasWaldmann, line 101: ah, good catch!

... is encrypted or authenticated.

and leave away the "client" part, people could also have the key on a server. "access to the borg key" is the important point.

src/borg/archiver/compact_cmd.py
```diff
@@ -176,32 +176,33 @@ def build_parser_compact(self, subparsers, common_parser, mid_common_parser):
             Free repository space by deleting unused chunks.
 
             borg compact analyzes all existing archives to find out which repository
-            objects are actually used (referenced). It then removes all unused objects
-            to free repository space.
+            objects are actually used (referenced). It then deletes all unused objects
+            from the filesystem to free space.
```
ThomasWaldmann, line 180: no "from the filesystem". borg supports all kinds of storages now (including cloud storage) and we can't know whether they use a filesystem.

src/borg/archiver/compact_cmd.py
```diff
@@ -176,32 +176,33 @@ def build_parser_compact(self, subparsers, common_parser, mid_common_parser):
...
             You usually don't want to run ``borg compact`` after every write operation, but
             either regularly (e.g. once a month, possibly together with ``borg check``) or
             when disk space needs to be freed.
 
             **Important:**
 
-            After compacting it is not possible anymore to use ``borg undelete`` to recover
-            previously deleted archives.
+            After compacting it is no longer possible to use ``borg undelete`` to recover
+            soft-deleted archives.
```
ThomasWaldmann, line 198: except the soft-deleted archives you deleted after that compact. :-) that's why I added "previously".

src/borg/archiver/delete_cmd.py
```diff
@@ -62,7 +62,7 @@ def build_parser_delete(self, subparsers, common_parser, mid_common_parser):
 
         delete_epilog = process_epilog(
             """
-        This command deletes archives from the repository.
+        This command soft-deletes archives from the repository.
```
ThomasWaldmann, line 65: do we have a glossary somewhere? to avoid having it to explain everywhere, but at least explain it at some defined place.

docs/changes.rst
PhrozenByte: Done.

I just forgot to also delete that part from `compact`, the reason why I even suggested that :smile: Done too now.

src/borg/archiver/compact_cmd.py
PhrozenByte: Hmmm... :thinking: Being pedantic, that's indeed also true for any cloud storage, it's just a matter of how many other layers are in between :laughing: Anyway, you're right, it's bad wording. I don't like "free repository space" very much though, changed it to "from the repository to free space".

src/borg/archiver/compact_cmd.py
PhrozenByte: Hmmm... :thinking: Yeah, that's right. The reason I wanna change that is because it kinda implies that it's somehow possible to recover archives that were "truly deleted" before. Let's just leave that "deleted" part out completely, because after `compact` it can't recover anything, so I just changed it to "to recover archives".

src/borg/archiver/delete_cmd.py
PhrozenByte:
```sh
$ grep -R glossary
docs/internals/data-structures.rst:.. todo:: Clarify terms, perhaps create a glossary.
```

:laughing: 

To be honest, I'm no big fan of glossaries in user docs. They might help for dev docs, but in user docs I'd (naturally I can only speak for myself here...) always rather just google for the word - and if this doesn't bring up meaningful stuff, it needs to be explained "inline". The first result for "soft delete" on Google is perfectly accurate btw:

> Soft delete is a widely used pattern […]. It allows you to mark some records as deleted without actually deleting them […]

---

About "explain it at some defined place": IMHO the best place for that is `delete` and `prune`. We could choose `delete` and just reference it in `prune`, but in the end it's not that big of an explanation. To keep it simple I'd recommend https://github.com/borgbackup/borg/pull/8515#discussion_r1827014820 again. I just did that. What do you think?

src/borg/archiver/compact_cmd.py
```diff
@@ -176,35 +176,33 @@ def build_parser_compact(self, subparsers, common_parser, mid_common_parser):
...
-            maybe usually don't want to do that unless there are signs of lost archives
-            (e.g. when seeing fatal errors when creating backups or when archives are
-            missing in ``borg list``).
-
-            Differently than borg 1.x, borg2's compact needs the borg key if the repo is
-            encrypted.
+            ``borg check --find-lost-archives [--repair]``, which is slow. You therefore
+            might not want to do that unless there are signs of lost archives (e.g. when
+            seeing fatal errors when creating backups or when archives are missing in
+            ``borg list``).
```
ThomasWaldmann, line 205: borg repo-list

bug existed before, but we can fix it now.

src/borg/archiver/delete_cmd.py
ThomasWaldmann: ok.

src/borg/archiver/compact_cmd.py
ThomasWaldmann: What I mean is, consider this sequence:
```
D D D C D D ...
```
D is a soft-deletion of an archive, C is compact.

You are saying the after C (to the right of C) it is not possible to recover soft-deleted archives, but it is, but only the one to the right of C. That's why I wanted that "previously".

PhrozenByte: Ah! Sorry, I didn't understand what you meant. That's indeed a valid point, I didn't even think about that users could assume that compaction has an ongoing effect, but nothing really contradicts that. So, "recover previously soft-deleted archives" it is I guess? :smiley:

src/borg/archiver/compact_cmd.py
PhrozenByte: Good catch! Done :+1:

src/borg/archiver/compact_cmd.py
ThomasWaldmann: You wouldn't believe what users can "read between the lines". :-)

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
Thanks!

Issue #8553, opened 2024-11-19
borg2: `setup.py` does not support OpenBSD's current versions


dertuxmalwieder:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yep.

## Is this a BUG / ISSUE report or a QUESTION?

Bug, basically.

#### Your borg version (borg -V).

None (can't).

#### Operating system (distribution) and version.

OpenBSD 7.6.

## Describe the problem you're observing.

It is currently not possible to run `pipx install borgbackup==2.0.0b14` on OpenBSD:

```
src/borg/crypto/low_level.c:16241:37: error: use of undeclared identifier 'EVP_aes_256_ocb'
```

As OpenBSD defaults to LibreSSL, Borg's `setup.py` contains a workaround; alas, that's outdated:

```python
elif is_openbsd:
        # Use openssl (not libressl) because we need AES-OCB via EVP api. Link
        # it statically to avoid conflicting with shared libcrypto from the base
        # OS pulled in via dependencies.
        crypto_ext_lib = {"include_dirs": ["/usr/local/include/eopenssl30"]}
        crypto_extra_objects += ["/usr/local/lib/eopenssl30/libcrypto.a"]
```

OpenBSD [does not have](https://openports.pl/cat/security) OpenSSL 3.0 anymore (currently, 1.1, 3.2 and 3.3 are there), so trying to use this hard-coded path *will* fail.

## Solution (idea)

Make the crypto library path(s) configurable (or, at least, the `eopenssl30` part), so when OpenBSD drops 3.2 and 3.3, the port maintainer(s) only need to change an environment variable.

bket:
@dertuxmalwieder, please have a look at #8557. Does this fit your need?

dertuxmalwieder:
@bket Looks mostly good to me, although I would probably set the default to `eopenssl32` now as that's the widest supported version...?

bket:
@dertuxmalwieder, you are right. However, I set it to `eopenssl30` as this fits the Vagrant box @ThomasWaldmann is using currently, which is running OpenBSD-7.4 and has no port for OpenSSL-3.2.

Nonetheless, default can be changed easily. I leave it up to @ThomasWaldmann to decide.

ThomasWaldmann:
See my comments in the PR.

ThomasWaldmann:
Fixed by #8557.

dertuxmalwieder:
FYI, it's 35 on OpenBSD 7.8. :)

PR #8557, opened 2024-11-23
borg2 - Support other OpenSSL versions on OpenBSD (#8553)

bket:
`setup.py` hardcoded crypto library paths for OpenBSD, causing build issue when OpenBSD drops specific OpenSSL version. Solution is to make paths configurable.

Addresses #8553.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,23 +132,23 @@ if not on_rtd:
             f"or ensure {lib_pkg_name}.pc is in PKG_CONFIG_PATH."
         )
 
-    crypto_extra_objects = []
     if is_win32:
         crypto_ext_lib = lib_ext_kwargs(pc, "BORG_OPENSSL_PREFIX", "libcrypto", "libcrypto", ">=1.1.1", lib_subdir="")
     elif is_openbsd:
         # Use openssl (not libressl) because we need AES-OCB via EVP api. Link
         # it statically to avoid conflicting with shared libcrypto from the base
         # OS pulled in via dependencies.
-        crypto_ext_lib = {"include_dirs": ["/usr/local/include/eopenssl30"]}
-        crypto_extra_objects += ["/usr/local/lib/eopenssl30/libcrypto.a"]
+        openssl_prefix = os.environ.get("BORG_OPENSSL_PREFIX", "/usr/local")
+        openssl_name = os.environ.get("BORG_OPENSSL_NAME", "eopenssl33")
+        crypto_ext_lib = dict(
+            include_dirs=[os.path.join(openssl_prefix, "include", openssl_name)],
+            extra_objects=[os.path.join(openssl_prefix, "lib", openssl_name, "libcrypto.a")],
+        )
     else:
         crypto_ext_lib = lib_ext_kwargs(pc, "BORG_OPENSSL_PREFIX", "crypto", "libcrypto", ">=1.1.1")
 
     crypto_ext_kwargs = members_appended(
-        dict(sources=[crypto_ll_source]),
-        crypto_ext_lib,
-        dict(extra_compile_args=cflags),
-        dict(extra_objects=crypto_extra_objects),
+        dict(sources=[crypto_ll_source]), crypto_ext_lib, dict(extra_compile_args=cflags)
     )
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -132,23 +132,20 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
...
     elif is_openbsd:
         # Use openssl (not libressl) because we need AES-OCB via EVP api. Link
         # it statically to avoid conflicting with shared libcrypto from the base
-        # OS pulled in via dependencies.
-        crypto_ext_lib = {"include_dirs": ["/usr/local/include/eopenssl30"]}
-        crypto_extra_objects += ["/usr/local/lib/eopenssl30/libcrypto.a"]
+        # OS pulled in via dependencies. Make crypto paths configurable.
+        include_dirs = [os.environ.get("BORG_OPENSSL_INC", "/usr/local/include/eopenssl30")]
+        extra_objects = [os.path.join(os.environ.get("BORG_OPENSSL_LIB", "/usr/local/lib/eopenssl30"), "libcrypto.a")]
+        crypto_ext_lib = {"include_dirs": include_dirs, "extra_objects": extra_objects}
```
ThomasWaldmann, line 143: I recently had a look at this, whether the code used for the other platforms could do this - and the main problem I found is that it needs that libcrypto.a here in extra_objects.

But I thought:
- it could use BORG_OPENSSL_PREFIX=/usr/local
- a new BORG_OPENSSL_NAME=eopenssl30 to be able to vary that part of the path.
- guess everything else is fixed?

ThomasWaldmann: @bket seen this ^ ?

setup.py
```diff
@@ -132,23 +132,20 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
...
     if is_win32:
         crypto_ext_lib = lib_ext_kwargs(pc, "BORG_OPENSSL_PREFIX", "libcrypto", "libcrypto", ">=1.1.1", lib_subdir="")
     elif is_openbsd:
         # Use openssl (not libressl) because we need AES-OCB via EVP api. Link
         # it statically to avoid conflicting with shared libcrypto from the base
-        # OS pulled in via dependencies.
-        crypto_ext_lib = {"include_dirs": ["/usr/local/include/eopenssl30"]}
-        crypto_extra_objects += ["/usr/local/lib/eopenssl30/libcrypto.a"]
+        # OS pulled in via dependencies. Make crypto paths configurable.
+        include_dirs = [os.environ.get("BORG_OPENSSL_INC", "/usr/local/include/eopenssl30")]
```
ThomasWaldmann, line 141: about the default:

what's beneficial here to most users?

i can adapt the vagrantfile so it'll use 3.0 for there.

ThomasWaldmann:
macos CI failure: looks unrelated, somebody fscked with pkg-config in brew.

setup.py
dertuxmalwieder: Most OpenBSD users use a version of OpenBSD that doesn't have OpenSSL 3.0.

setup.py
bket: @ThomasWaldmann, yes, I saw your comments, and updated the PR accordingly.

setup.py
ThomasWaldmann: I'll run tests on my vagrant box now.

What I meant with beneficial for most users is that there could be a default of eopenssl32 (or whatever), so that it works for most users and I would just use the env vars to make it fit the old vagrant box. of course, if there is a different openssl version used for every openbsd version, there might be no clear choice.

ThomasWaldmann: due to the empty defaults, guess it currently works for nobody "out of the box" without setting the env vars.

ThomasWaldmann: it worked on my vagrant box.

bket: > due to the empty defaults, guess it currently works for nobody "out of the box" without setting the env vars.


I set some sane defaults so that users, who run OpenBSD current, can do without setting env vars.

ThomasWaldmann:
Thanks!

PR #8563, opened 2024-11-24
Clarify  excluded and included flags for dry-run in borg the docs

goebbe:
Rewording of the "flags" description in the docs of borg 1.4. https://github.com/borgbackup/borg/issues/8556

Issue #8556, opened 2024-11-22
docs: Clarify the meaning of flags when using borg recreate --dry-run 


goebbe:
Using borg 1.4.0 I want to delete a huge folder (miniforge) from my old archives.

I understand that `borg recreate` is the correct command for this task and as suggested in the docs I first 
did a `--dry-run --verbose --list` to check the consequences of borg recreate: 

Simplified example: `borg recreate repo  --exclude home/user/miniforge3 --dry-run --verbose --list`  

The result is (as expected) a huge list of items -  with flags (x or -): 
```
x home/user/miniforge3
- home/user/Documents
```

To understand the exact meaning of these flags (in the context of borg recreate), I had a look at the docs: 
https://borgbackup.readthedocs.io/en/stable/usage/create.html#item-flags
and found the following description: 

Other flags used include:
    ‘-’ = dry run, item was not backed up
    ‘x’ = excluded, item was not backed up

From this description, it seems that all items in my /home/user/* are  "not backed up" for different reasons. I believe this is confusing, since in the context of `borg recreate`, the files are already "backed up" in the (old) archives. 

If I interpret the description correctly, then: 
    ‘-’ = dry run, item was not backed up     <- in the context of borg recreate: item will remain in the recreated archive(s)
    ‘x’ = excluded, item was not backed up  <- in the context of borg recreate: item will be excluded (i.e. removed) from the recreated archive(s)

Would it be possible to **add the following additional description to the docs?** 

In the context of `borg recreate`: 
    ‘-’ = dry run, item will remain in the recreated archive(s)
    ‘x’ = excluded, item will be excluded (i.e. removed) from the recreated archive(s)

ThomasWaldmann:
Yes, the docs could be improved and your suggestion sounds good. Can you make a pull request with that?

goebbe:
The docs for the "flags" are currently in the (docs) section "borg create". 
This kind of explains why the docs (1.4) describe the meaning of "flags" in the context of `borg create`.  

Only now, I realize, that the docs for "flags" look **different in** the **latest borg 2.0**: 
https://borgbackup.readthedocs.io/en/**latest**/usage/create.html#item-flags
Other flags used include:
    ‘+’ = included, item would be backed up (if not in dry-run mode)
    ‘-’ = excluded, item would not be / was not backed up
    ‘i’ = backup data was read from standard input (stdin)
    ‘?’ = missing status code (if you see this, please file a bug report!)

[...]

PR #8566, opened 2024-11-24
 Clarify excluded and included flags for dry-run in borg the docs #8563

goebbe:
Rewording of the "flags" description in the docs of borg 1.2. #8556


ThomasWaldmann:
Thanks!

PR #8575, opened 2024-11-29
Add flags to borg extract Command (Fixes #8564)

### LINKED ISSUE #8564: No dry-run flags when using borg extract
Using borg 1.4, I checked the use of "flags" when doing a dry-run for different borg commands. 

The command `borg extract --dry-run --list` does not print the flags (that are used with e.g. borg create / borg recreate
Instead, `borg extract` prints only the list of items that would be included in the extract without the --dry-run. 

This was the command, I used: 
`borg extract REPO::ARCHIVE /home/user/test Admin -e 'home/user/test/*.bak'  --dry-run --verbose --list `

In the name of consistency, dry-run flags could be added to borg extract.

alighazi288:
## Summary 
This pull request addresses the issue with the `borg extract` command not displaying the flags as seen in other commands eg. `borg create` and `borg recreate`.


src/borg/archiver/extract_cmd.py
```diff
@@ -71,6 +71,7 @@ def do_extract(self, args, repository, manifest, archive):
                 logging.getLogger("borg.output.list").info(remove_surrogates(item.path))
             try:
                 if dry_run:
+                    print(f"- {item.path}")
```
ThomasWaldmann, line 74: check the meaning of the flags (see our docs) and try running borg create with some excludes. borg extract should flag similarly to create.

ThomasWaldmann: also be careful with print, print outputs to stdout and borg extract also has the option to write extracted file contents to stdout - these 2 shall not be mixed up.

alighazi288: Apologies for the oversight. Trying a new fix and uploading soon for review.

ThomasWaldmann:
BTW, don't add `coverage.*` files to the repo, these are temporary files.

alighazi288:
Got it, I’ll ensure temporary files are omitted moving forward. Could you please share more details or context about the errors above? This would help me better understand and address the issue.

ThomasWaldmann (commented):
some more feedback.

ThomasWaldmann:
The failing `test_return_codes` test is because `borg extract` is used there and expected to finish with rc 0, but it crashes and returns rc 2.

alighazi288:
Changed the code and now everything seems to be running as expected at my end and **borg extract** finishes with rc 0. Just wanted to ask if the errors could be because of me working on MacOS and not having pyfuse3?

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,42 @@ def do_extract(self, args, repository, manifest, archive):
...
                     dir_item = dirs.pop(-1)
                     try:
                         archive.extract_item(dir_item, stdout=stdout)
                     except BackupError as e:
                         self.print_warning_instance(BackupWarning(remove_surrogates(dir_item.path), e))
-            if output_list:
-                logging.getLogger("borg.output.list").info(remove_surrogates(item.path))
+
+                if output_list and matcher.match(item.path):
+                    logging.getLogger("borg.output.list").info(remove_surrogates(item.path))
```
ThomasWaldmann, line 75: this log output is different from the other one, no status char.

do you practically try your code after making changes?

ThomasWaldmann: you also can run the unit tests locally, just invoke pytest or tox.

ThomasWaldmann:
The errors are unrelated to macOS and not having pyfuse3 - I also work on a macbook.

If one want to test or work on the mount / FUSE code in borg, one would need macFUSE installed and use llfuse (not pyfuse3). But this is not required here.

src/borg/archiver/extract_cmd.py
alighazi288: @ThomasWaldmann ?

alighazi288:
@ThomasWaldmann Please review.

src/borg/archiver/extract_cmd.py
```diff
@@ -71,29 +75,34 @@ def do_extract(self, args, repository, manifest, archive):
...
+                            )
             except BackupError as e:
                 self.print_warning_instance(BackupWarning(remove_surrogates(orig_path), e))
-
         if pi:
             pi.finish()
+        # Display excluded items
+        if excluded_items:
+            for excluded_item in excluded_items:
+                logging.getLogger("borg.output.list").info(f"- {remove_surrogates(excluded_item)}")
```
ThomasWaldmann, line 105: It would be better to display included and excluded items both together in the order they appear in the archive.

alighazi288: @ThomasWaldmann done.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,53 @@ def do_extract(self, args, repository, manifest, archive):
         else:
             pi = None
 
-        for item in archive.iter_items(filter, preload=True):
+        for item in archive.iter_items(preload=True):
```
ThomasWaldmann, line 59: Hmm, this develops to be a bit more tricky than I initially thought it is.

If you leave away the filter here (which at first looks like a good idea to get ALL items yielded), it will preload the file content chunks of ALL yielded items (not just the ones we might extract).

That's not necessary and it is even harmful, leading to a memory leak if one does not retrieve these chunks, see comment in `unpack_many`.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,53 @@ def do_extract(self, args, repository, manifest, archive):
...
-                    archive.extract_item(item, dry_run=True, hlm=hlm, pi=pi)
-                else:
-                    if stat.S_ISDIR(item.mode):
-                        dirs.append(item)
-                        archive.extract_item(item, stdout=stdout, restore_attrs=False)
+            components = orig_path.split(os.sep)
+            if strip_components >= len(components):
+                sanitized_path = ""
+            else:
+                sanitized_path = os.sep.join(components[strip_components:])
```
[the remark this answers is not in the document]
ThomasWaldmann: also `stripped_path` would be a better name, there is no sanitation done here.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,53 @@ def do_extract(self, args, repository, manifest, archive):
...
+
+            item.path = sanitized_path
+
+            is_matched = matcher.match(item.path)
+            log_prefix = "+" if is_matched else "-"
+            logging.getLogger("borg.output.list").info(f"{log_prefix} {remove_surrogates(item.path)}")
+
+            if is_matched:
+
+                if not args.dry_run:
```
ThomasWaldmann, line 78: use dry_run.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,53 @@ def do_extract(self, args, repository, manifest, archive):
...
+                sanitized_path = ""
+            else:
+                sanitized_path = os.sep.join(components[strip_components:])
+
+            if not sanitized_path:
+                continue
+
+            item.path = sanitized_path
+
+            is_matched = matcher.match(item.path)
```
ThomasWaldmann, line 72: hmm, doesn't the original code match using the full (non-stripped) path?

src/borg/archiver/extract_cmd.py
```diff
@@ -56,32 +56,53 @@ def do_extract(self, args, repository, manifest, archive):
...
+                try:
+                    if dry_run:
+                        archive.extract_item(item, dry_run=True, hlm=hlm, pi=pi)
                     else:
-                        archive.extract_item(
-                            item, stdout=stdout, sparse=sparse, hlm=hlm, pi=pi, continue_extraction=continue_extraction
-                        )
-            except BackupError as e:
-                self.print_warning_instance(BackupWarning(remove_surrogates(orig_path), e))
+                        if is_matched:
```
ThomasWaldmann, line 90: can it ever be False here?

ThomasWaldmann (changes requested):
This is getting tricky.

src/borg/archiver/extract_cmd.py
alighazi288: When you use the filter, only items that match the filter criteria are processed which prevents us from logging any excluded items, which is why I proposed the "two pass" solution but that would display excludes either at the beginning or the end.

ThomasWaldmann: Sure. So there is no good solution yet and it is not easy due to the preloading requirements.

ThomasWaldmann: I am working right now on decoupling preloading and item iteration to help you with that.

After that is merged, you can rebase your work on these changes.

alighazi288: Thank you so much. I really appreciate it!

ThomasWaldmann: #8592 was just merged into master.

Please update your local master branch and rebase your PR branch onto that.

As you can see there, `archive.iter_items()` does not support preloading anymore.
At the places where preloading is needed, I inserted a new call that does the preloading.

Be careful: the preload call must only be done if the item chunks will be fetched later (e.g. if the item is extracted, because it is included / not excluded). The `borg extract` code fetches the content chunks even if it is in dry-run mode, because it is documented to do all steps of the extraction except writing content to disk files (so people can use it as a practical test whether the extraction would work without needing the space to do the extraction).

ThomasWaldmann:
Can you rather rebase on master please than merge master in here?

src/borg/archiver/extract_cmd.py
alighazi288: Thank you so much once again for the clarification and your help! Working on it.

src/borg/archiver/extract_cmd.py
```diff
@@ -61,6 +61,34 @@ def do_extract(self, args, repository, manifest, archive):
             orig_path = item.path
             if strip_components:
                 item.path = os.sep.join(orig_path.split(os.sep)[strip_components:])
+            status = None
+
+            if matcher.match(item.path):
```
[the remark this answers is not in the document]
ThomasWaldmann: also, does it have a special reason why you moved the extraction to here and not kept it at the original place? the code in 91+ extracts needed dirs and that should be done before files are extracted that have these as parent dirs.

alighazi288:
I wanted to let you know that there appears to be a large number of changes in this pull request because I encountered some issues while attempting to rebase on master. I apologize for any confusion this may have caused.

src/borg/archiver/extract_cmd.py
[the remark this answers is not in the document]
ThomasWaldmann: Maybe it makes sense to keep the test and start from scratch with the extraction code changes.

Now that the preloading is out of the way, it should be easy to get a list of ALL items, do the matching here, extract and "+" if it matches, "-" if it doesn't.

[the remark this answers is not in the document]
ThomasWaldmann: My comment above was written before the most recent changes, so let's close this.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,33 +55,50 @@ def do_extract(self, args, repository, manifest, archive):
         else:
             pi = None
 
-        for item in archive.iter_items(filter):
-            archive.preload_item_chunks(item, optimize_hardlinks=True)
+        for item in archive.iter_items(None):
```
ThomasWaldmann, line 58: Isn't "None" the default anyway?

src/borg/archiver/extract_cmd.py
```diff
@@ -56,33 +55,50 @@ def do_extract(self, args, repository, manifest, archive):
...
-            if strip_components:
-                item.path = os.sep.join(orig_path.split(os.sep)[strip_components:])
-            if not args.dry_run:
-                while dirs and not item.path.startswith(dirs[-1].path):
-                    dir_item = dirs.pop(-1)
-                    try:
-                        archive.extract_item(dir_item, stdout=stdout)
-                    except BackupError as e:
-                        self.print_warning_instance(BackupWarning(remove_surrogates(dir_item.path), e))
-            if output_list:
```
ThomasWaldmann, line 71: Why did you remove checking output_list?

ThomasWaldmann (changes requested):
That looks quite ok now, some minor things todo!

src/borg/archiver/extract_cmd.py
alighazi288: I thought it simplified the logging logic. In the updated code, we always log the paths of the items with a prefix indicating whether they are matched or not. If it is necessary to conditionally control the logging based on the output_list flag, I can reintroduce that check.

src/borg/archiver/extract_cmd.py
alighazi288: Yes, you're right. I thought passing None would make the code's behavior more explicit and clear. I can remove it to simplify the code.

src/borg/archiver/extract_cmd.py
alighazi288: @ThomasWaldmann Yes it is. However, I have still added the output_list as per your request. Please let me know what else is required to finally close this?

src/borg/archiver/extract_cmd.py
```diff
@@ -66,7 +67,10 @@ def do_extract(self, args, repository, manifest, archive):
 
             is_matched = matcher.match(orig_path)
             log_prefix = "+" if is_matched else "-"
-            logging.getLogger("borg.output.list").info(f"{log_prefix} {remove_surrogates(item.path)}")
+
+            # Conditionally log based on the output_list flag
```
ThomasWaldmann, line 71: please avoid such comments. it does not add any information that is not obvious from the next 2 lines anway.

alighazi288: Understood. Would you like me to remove it and commit again?

ThomasWaldmann: Yes, please.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,33 +56,51 @@ def do_extract(self, args, repository, manifest, archive):
...
-                    try:
-                        archive.extract_item(dir_item, stdout=stdout)
-                    except BackupError as e:
-                        self.print_warning_instance(BackupWarning(remove_surrogates(dir_item.path), e))
+            components = orig_path.split(os.sep)
+            stripped_path = os.sep.join(components[strip_components:])
+
+            if not stripped_path:
+                continue
+            item.path = stripped_path
```
ThomasWaldmann, line 66: the original code only did that "if strip_components".

src/borg/archiver/extract_cmd.py
```diff
@@ -38,8 +38,8 @@ def do_extract(self, args, repository, manifest, archive):
         matcher = build_matcher(args.patterns, args.paths)
 
         progress = args.progress
-        output_list = args.output_list
         dry_run = args.dry_run
+        output_list = args.output_list
```
ThomasWaldmann, line 42: as you are back to the original code here, move it 1 line up, so it disappears from the diff.

also, in case you don't do that already, I'ld suggest you read the diff, so you can find such things by yourself.

src/borg/archiver/extract_cmd.py
ThomasWaldmann: also, you could stay closer to the original code (like: "don't fix what ain't broken") and just add the continue if item.path ends up being empty.

src/borg/archiver/extract_cmd.py
```diff
@@ -56,33 +56,51 @@ def do_extract(self, args, repository, manifest, archive):
...
-                        self.print_warning_instance(BackupWarning(remove_surrogates(dir_item.path), e))
+            components = orig_path.split(os.sep)
+            stripped_path = os.sep.join(components[strip_components:])
+
+            if not stripped_path:
+                continue
+            item.path = stripped_path
+
+            is_matched = matcher.match(orig_path)
+            log_prefix = "+" if is_matched else "-"
```
ThomasWaldmann, line 69: if you don't use log_prefix except in line 72, move 69 directly above that line.

ThomasWaldmann (commented):
some minor nitpicks, please fix.

it's mostly about staying closer to the original src and not doing unneeded changes.

guess it can be merged after that (I will squash all the commits, so history will be cleaner). thanks for working on this!

ThomasWaldmann:
There is still that ".coverage 2" file in your PR, you need to remove it, it is a temporary file and must not be committed.

src/borg/archiver/extract_cmd.py
alighazi288: Got it. The only issue is, that just adding continue resulted in test failures due to a `ValueError` being raised. `ValueError: path '' is not sanitized`. To resolve this, I updated the code to ensure that `item.path` is only set if the resulting path is not empty.

ThomasWaldmann (approved):
LGTM.

Issue #8589, opened 2024-12-20
Wildcard extraction does not work


Atemu:
## Have you checked borgbackup docs, FAQ, and open GitHub issues?

Yes

## Is this a BUG / ISSUE report or a QUESTION?

BUG

#### Your borg version (borg -V).

borg 1.4.0

#### Operating system (distribution) and version.

https://github.com/NixOS/nixpkgs/commit/5d67ea6b4b63378b9c13be21e2ec9d1afc921713

#### Hardware / network configuration, and filesystems used.

tmpfs

#### How much data is handled by borg?

kilobytes

#### Full borg commandline that lead to the problem (leave away excludes and passwords)

    borg extract --dry-run --list /var/tmp/borg/::foo 'a*'

## Describe the problem you're observing.

I tried to extract multiple files that have the same prefix. The simplest method to achieve this in my mind would be to use a wildcard. According to https://github.com/borgbackup/borg/issues/361, wildcards should be supported by borg but there is no documentation on it (:/) and they evidently do not work.

Am I doing something wrong? Are users expected to understand and create borg-specific patterns instead to achieve this sort of thing?

#### Can you reproduce the problem? If so, describe how. If not, describe troubleshooting steps you took before opening the issue.

1. `touch a aa aaa`
2. `borg create /var/tmp/borg/::foo .`
3. `borg extract --dry-run --list /var/tmp/borg/::foo 'a*'`

#### Include any warning/errors/backtraces from the system logs

```
Include pattern 'a*' never matched.
```

Atemu:
`borg extract --dry-run --list /var/tmp/borg/::foo --pattern '+ a*' --pattern '- *'` appears to work but that is not at all ergonomical.

infectormp:
You should use `--pattern` to define what you want to include/exclude. IMHO it is clearly documented. If not feel free to do PR to improve documentation.

ThomasWaldmann:
Maybe the docs need to be improved there. I had a look at the 1.4.0 source and this is what it does:

The PATHS list from the command line is processed as "include paths" and the **default** pattern style used is `PathPrefixPattern`.

If one wants another style, e.g. `sh:` or `re:` or `fm:`, that needs to be explicitly prefixed.

This works:
```
borg extract --list ::archive 'sh:a*'    
```

Atemu:
Indeed that works.

I'd personally advocate for that to be made the default pattern interpretation because that is the least surprising IMHO.

[...]

PR #8590, opened 2024-12-20
extract: document lack of wildcard support in default pattern style

Atemu:
Fixes https://github.com/borgbackup/borg/issues/8589

### DIFF
--- a/docs/usage/extract.rst
+++ b/docs/usage/extract.rst
@@ -19,6 +19,9 @@ Examples
     # Extract the "src" directory but exclude object files
     $ borg extract my-files home/USERNAME/src --exclude '*.o'
 
+    # Extract only the object files
+    $ borg extract my-files 'sh:home/USERNAME/src/*.o'
+
     # Restore a raw device (must not be active/in use/mounted at that time)
     $ borg extract --stdout my-sdx | dd of=/dev/sdx bs=10M
 
--- a/src/borg/archiver/extract_cmd.py
+++ b/src/borg/archiver/extract_cmd.py
@@ -130,6 +130,10 @@ class ExtractMixIn:
             When parent directories are not extracted (because of using file/directory selection
             or any other reason), borg can not restore parent directories' metadata, e.g. owner,
             group, permission, etc.
+
+            The default interpretation for the paths to extract is `pp:` which is a literal
+            path-prefix match. If you want to use e.g. a wildcard, you must select a different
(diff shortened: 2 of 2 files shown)


docs/usage/extract.rst
```diff
@@ -19,6 +19,9 @@ Examples
     # Extract the "src" directory but exclude object files
     $ borg extract my-files home/USERNAME/src --exclude '*.o'
 
+    # Extract only the object files
+    $ borg extract my-files 'sh:home/USERNAME/src/*.o'
+
     # Restore a raw device (must not be active/in use/mounted at that time)
```
ThomasWaldmann, line 25: Good to add an example, but please use something more familiar to users (.pdf rather than .o). Not everybody is a developer.

src/borg/archiver/extract_cmd.py
```diff
@@ -130,6 +130,10 @@ def build_parser_extract(self, subparsers, common_parser, mid_common_parser):
             When parent directories are not extracted (because of using file/directory selection
             or any other reason), borg can not restore parent directories' metadata, e.g. owner,
             group, permission, etc.
+
+            The default interpretation for the paths to extract is `pp:` which is a literal
+            path-prefix match. If you want to use e.g. a wildcard, you must select a different
+            pattern style such as `sh:` or `fm:`. See :ref:`borg_patterns` for more information.
```
ThomasWaldmann, line 136: PATHs is already explained in 113. Please rephrase and move this there.

ThomasWaldmann (commented):
Thanks for the PR!

ThomasWaldmann:
@Atemu did you see my feedback, can you fix it?

Atemu:
I did see your feedback, I will get to this eventually.

ThomasWaldmann:
I'll release 1.4.1 soon, would be good if we could get this merged.

docs/usage/extract.rst
ThomasWaldmann: hmm, that problem existed previously to this PR, as other examples also used object files.

src/borg/archiver/extract_cmd.py
ThomasWaldmann: fixed

ThomasWaldmann:
Fixed it myself.

ThomasWaldmann:
Hmm, is there an issue in #9026 or why does it not backport?

Update: it triggered the action, but then just said "Skipped"!?

PR #8593, opened 2024-12-24
docs: update install docs, nothing bundled anymore! fixes #8342.

### LINKED ISSUE #8342: install docs need update about not bundled libs
From @kubedzero : 

The documentation https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies still says the libraries are bundled, should that be removed?

ThomasWaldmann:
borg 1.2.x used to bundle some source code for lz4, zstd and xxhash, but borg 1.4.x got rid of that - all that is meanwhile available as a library in all OSes we support.

also, the python package pkgconfig does not need to be "preinstalled" anymore, because our pyproject.toml cares for that. otoh, the cli tool pkg-config must be preinstalled so that libs and headers can be found automagically.

Also be a bit more clear about the FUSE stuff.

### DIFF
--- a/docs/global.rst.inc
+++ b/docs/global.rst.inc
@@ -13,6 +13,7 @@
 .. _ACL: https://en.wikipedia.org/wiki/Access_control_list
 .. _libacl: https://savannah.nongnu.org/projects/acl/
 .. _libattr: https://savannah.nongnu.org/projects/attr/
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
No backport to 1.2-maint needed, the stuff was still bundled there and also we didn't have pyproject.toml yet.

master: needs checking what needs to be ported to there.

PR #8605, opened 2024-12-26
Enhance passphrase handling (Fixes #8496)

### LINKED ISSUE #8496: add passphrase debugging option?
Sometimes people have issues when trying to access a borg repo from a new or newly installed host with borg telling them that the passphrase is incorrect.

Besides the obvious root cause of the passphrase really being incorrect, there are also:
- keyboard issues (kbd mapping, deadkey vs nodeadkey with accents in passphrases, ...)
- differences in unicode normalisation between platforms
- troubles correctly setting `BORG_PASSPHRASE`, `BORG_PASSCOMMAND`, ... env vars

`borg init` (borg 1.x)  and `borg repo-create` (borg 2) show the utf-8 hex bytes sequence of the encoded passphrase **IF** the passphrase is non-ascii.

We could:
- init: display the hex bytes in any case (the deadkey issue can also happen within ascii, e.g. if the original passphrase has `^a` and the wrong one happens to have `â`).
- add an option to display the password and/or hex bytes when the password is **USED** and is determined to be incorrect.
- show which env vars were used

alighazi288:
## Description

This pull request improves the handling of passphrases and adds comprehensive debug information. The main changes include:

- Displaying hex bytes for passphrases in all cases, including ASCII and non-ASCII characters.
- Adding an option to display the hex bytes when the password is used and found incorrect.
- Showing which environment variables were used during passphrase operations.
- Preventing sensitive passphrase information from being logged unintentionally by directing it to sys.stderr.
- Adding tests to verify:
  - Handling of incorrect passphrases.
  - Passphrase verification logic.
  - Debug information display for passphrases.

### DIFF
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -364,7 +364,6 @@ class FlexiKey:
         target = key.find_key()
         prompt = "Enter passphrase for key %s: " % target
         passphrase = Passphrase.env_passphrase()
-        debug_enabled = os.environ.get("BORG_DEBUG_PASSPHRASE") == "YES"
         if passphrase is None:
             passphrase = Passphrase()
             if not key.load(target, passphrase):
@@ -372,14 +371,12 @@ class FlexiKey:
                     passphrase = Passphrase.getpass(prompt)
                     if key.load(target, passphrase):
                         break
-                    elif debug_enabled:
-                        Passphrase.display_debug_info(passphrase)
+                    Passphrase.display_debug_info(passphrase)
                 else:
                     raise PasswordRetriesExceeded
         else:
             if not key.load(target, passphrase):
-                if debug_enabled:
-                    Passphrase.display_debug_info(passphrase)
+                Passphrase.display_debug_info(passphrase)
                 raise PassphraseWrong
         key.init_ciphers(manifest_data)
         key._passphrase = passphrase
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -375,7 +375,7 @@ class FlexiKey:
                     raise PasswordRetriesExceeded
         else:
             if not key.load(target, passphrase):
-                raise PassphraseWrong
+                raise PassphraseWrong(passphrase)
         key.init_ciphers(manifest_data)
         key._passphrase = passphrase
         return key
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -371,6 +371,8 @@ class FlexiKey:
                     passphrase = Passphrase.getpass(prompt)
                     if key.load(target, passphrase):
                         break
+                    else:
+                        Passphrase.display_debug_info(passphrase)
                 else:
                     raise PasswordRetriesExceeded
         else:
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -364,6 +364,7 @@ class FlexiKey:
         target = key.find_key()
         prompt = "Enter passphrase for key %s: " % target
         passphrase = Passphrase.env_passphrase()
+        debug_enabled = os.environ.get("BORG_DEBUG_PASSPHRASE") == "YES"
         if passphrase is None:
             passphrase = Passphrase()
             if not key.load(target, passphrase):
@@ -371,13 +372,15 @@ class FlexiKey:
                     passphrase = Passphrase.getpass(prompt)
                     if key.load(target, passphrase):
                         break
-                    else:
+                    elif debug_enabled:
                         Passphrase.display_debug_info(passphrase)
                 else:
                     raise PasswordRetriesExceeded
         else:
             if not key.load(target, passphrase):
-                raise PassphraseWrong(passphrase)
+                if debug_enabled:
+                    Passphrase.display_debug_info(passphrase)
+                raise PassphraseWrong
         key.init_ciphers(manifest_data)
         key._passphrase = passphrase
         return key
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -121,16 +121,21 @@ class Passphrase(str):
     @staticmethod
     def display_debug_info(passphrase):
         if os.environ.get("BORG_DEBUG_PASSPHRASE") == "YES":
-            print("Incorrect passphrase!", file=sys.stderr)
-            print(f'Passphrase used (between double-quotes): "{passphrase}"', file=sys.stderr)
-            print(f'Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode("utf-8"))}', file=sys.stderr)
-            print("Relevant Environment Variables:", file=sys.stderr)
+            env_vars_str = ""
             for env_var in ["BORG_PASSPHRASE", "BORG_PASSCOMMAND", "BORG_PASSPHRASE_FD"]:
                 env_var_value = os.environ.get(env_var)
                 if env_var_value is not None:
-                    print(f'{env_var} = "{env_var_value}"', file=sys.stderr)
+                    env_vars_str += f'{env_var} = "{env_var_value}"\n'
                 else:
-                    print(f"# {env_var} is not set", file=sys.stderr)
+                    env_vars_str += f"# {env_var} is not set\n"
+            passphrase_info = (
+                f"Incorrect passphrase!\n"
+                f'Passphrase used (between double-quotes): "{passphrase}"\n'
+                f"Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode('utf-8'))}\n"
+                f"Relevant Environment Variables:\n"
+                f"{env_vars_str}"
+            )
+            print(passphrase_info, file=sys.stderr)
 
     @classmethod
     def new(cls, allow_empty=False):
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -31,6 +31,10 @@ class PassphraseWrong(Error):
 
     exit_mcode = 52
 
+    def __init__(self, passphrase):
+        super().__init__(self)
+        Passphrase.display_debug_info(passphrase)
+
 
 class PasswordRetriesExceeded(Error):
     """exceeded the maximum password retries"""
@@ -111,18 +115,26 @@ class Passphrase(str):
         ):
             print('Your passphrase (between double-quotes): "%s"' % passphrase, file=sys.stderr)
             print("Make sure the passphrase displayed above is exactly what you wanted.", file=sys.stderr)
+            print(
+                "Your passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
+            )
             try:
                 passphrase.encode("ascii")
             except UnicodeEncodeError:
-                print(
-                    "Your passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")),
-                    file=sys.stderr,
-                )
                 print(
                     "As you have a non-ASCII passphrase, it is recommended to keep the "
                     "UTF-8 encoding in hex together with the passphrase at a safe place.",
                     file=sys.stderr,
                 )
+
+    @staticmethod
+    def display_debug_info(passphrase):
+        print(
+            "Incorrect passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
+        )
+        for env_var in ["BORG_PASSPHRASE", "BORG_PASSCOMMAND", "BORG_PASSPHRASE_FD"]:
+            env_var_value = os.environ.get(env_var)
+            print(f"{env_var} = {env_var_value}", file=sys.stderr)
 
     @classmethod
     def new(cls, allow_empty=False):
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -118,6 +118,11 @@ class Passphrase(str):
             print(
                 "Your passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
             )
+            print(
+                "It is recommended to keep the UTF-8 encoding in hex together with the passphrase at a safe place. "
+                "In case you should ever run into passphrase issues, it could sometimes help debugging them.\n",
+                file=sys.stderr,
+            )
             try:
                 passphrase.encode("ascii")
             except UnicodeEncodeError:
@@ -129,12 +134,16 @@ class Passphrase(str):
 
     @staticmethod
     def display_debug_info(passphrase):
-        print(
-            "Incorrect passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
-        )
+        print("Incorrect passphrase!", file=sys.stderr)
+        print(f'Passphrase used (between double-quotes): "{passphrase}"', file=sys.stderr)
+        print(f'Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode("utf-8"))}', file=sys.stderr)
+        print("Relevant Environment Variables:", file=sys.stderr)
         for env_var in ["BORG_PASSPHRASE", "BORG_PASSCOMMAND", "BORG_PASSPHRASE_FD"]:
             env_var_value = os.environ.get(env_var)
-            print(f"{env_var} = {env_var_value}", file=sys.stderr)
+            if env_var_value is not None:
+                print(f'{env_var} = "{env_var_value}"', file=sys.stderr)
+            else:
+                print(f"# {env_var} is not set", file=sys.stderr)
 
     @classmethod
     def new(cls, allow_empty=False):
--- a/src/borg/helpers/passphrase.py
+++ b/src/borg/helpers/passphrase.py
@@ -31,10 +31,6 @@ class PassphraseWrong(Error):
 
     exit_mcode = 52
 
-    def __init__(self, passphrase):
-        super().__init__(self)
-        Passphrase.display_debug_info(passphrase)
-
 
 class PasswordRetriesExceeded(Error):
     """exceeded the maximum password retries"""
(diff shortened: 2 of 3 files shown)


src/borg/crypto/key.py
```diff
@@ -375,7 +375,7 @@ def detect(cls, repository, manifest_data):
                     raise PasswordRetriesExceeded
         else:
             if not key.load(target, passphrase):
-                raise PassphraseWrong
+                raise PassphraseWrong(passphrase)
```
ThomasWaldmann, line 378: please read the code above, you only covered env passphrase.

src/borg/helpers/passphrase.py
```diff
@@ -111,19 +115,27 @@ def verification(cls, passphrase):
...
             except UnicodeEncodeError:
-                print(
-                    "Your passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")),
-                    file=sys.stderr,
-                )
                 print(
                     "As you have a non-ASCII passphrase, it is recommended to keep the "
                     "UTF-8 encoding in hex together with the passphrase at a safe place.",
                     file=sys.stderr,
                 )
```
ThomasWaldmann, line 128: guess we can always do that print. as i explained in the ticket, there can be issues even with ascii passphrases.

"It is recommended to keep the UTF-8 encoding in hex together with the passphrase at a safe place. In case you should ever run into passphrase issues, it can sometimes help debugging them."

src/borg/helpers/passphrase.py
```diff
@@ -111,19 +115,27 @@ def verification(cls, passphrase):
...
 
+    @staticmethod
+    def display_debug_info(passphrase):
+        print(
+            "Incorrect passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
+        )
+        for env_var in ["BORG_PASSPHRASE", "BORG_PASSCOMMAND", "BORG_PASSPHRASE_FD"]:
+            env_var_value = os.environ.get(env_var)
+            print(f"{env_var} = {env_var_value}", file=sys.stderr)
```
ThomasWaldmann, line 137: if would be good to output the env variables in a shell-like syntax (bash), including quotes around the value, so that leading/trailing blanks can be seen.

also guess you need to specialcase not set env vars, outputting "None" would be misleading.

src/borg/helpers/passphrase.py
```diff
@@ -111,19 +115,27 @@ def verification(cls, passphrase):
...
                     "As you have a non-ASCII passphrase, it is recommended to keep the "
                     "UTF-8 encoding in hex together with the passphrase at a safe place.",
                     file=sys.stderr,
                 )
 
+    @staticmethod
+    def display_debug_info(passphrase):
+        print(
+            "Incorrect passphrase (UTF-8 encoding in hex): %s" % bin_to_hex(passphrase.encode("utf-8")), file=sys.stderr
+        )
```
ThomasWaldmann, line 134: Guess we also want to output the passphrase in clear text here.

src/borg/helpers/passphrase.py
```diff
@@ -111,19 +115,27 @@ def verification(cls, passphrase):
...
-                    file=sys.stderr,
-                )
                 print(
                     "As you have a non-ASCII passphrase, it is recommended to keep the "
                     "UTF-8 encoding in hex together with the passphrase at a safe place.",
                     file=sys.stderr,
                 )
 
+    @staticmethod
+    def display_debug_info(passphrase):
```
ThomasWaldmann, line 131: Maybe use a more structured output here, like:

Incorrect passphrase!

Passphrase used (without the quotes): "..."
Same, UTF-8 encoded, in hex: ...

Relevant Environment variables:
BORG_PASSPHRASE = "..."
...

ThomasWaldmann (changes requested):
Thanks for the PR! Some feedback.

alighazi288:
Thank you so much for the feedback! Will try to fix all this by tomorrow!

src/borg/crypto/key.py
alighazi288: I see that `Passphrase.env_passphrase()` handles `BORG_PASSPHRASE`, `BORG_PASSCOMMAND`, and `BORG_PASSPHRASE_FD`. Are you suggesting I need to adjust how `PassphraseWrong` is raised in these cases, or is there another specific issue you had in mind?

ThomasWaldmann: The goal of this is to be able to be helpful with "passphrase wrong" issues, no matter how the pp is entered.

If somebody enters the passphrase on the console and does not manage to enter a correct one, it will raise PasswordRetriesExceeded - but that is just a special case of PassphraseWrong.

alighazi288: Thank you for the clarification! I've updated the code to ensure that debugging information is displayed after every failed attempt, no matter how the passphrase is entered.

If you'd like, I can further streamline the debugging output for interactive retries by skipping environment variable details `(e.g., BORG_PASSPHRASE, BORG_PASSCOMMAND, etc.)`, as those may not always be relevant in an interactive context.

ThomasWaldmann (commented):
some feedback.

beside addressing the feedback and writing a unit test, i also recommend that you practically try this code for some scenarios.

E.g.:

Use `borg repo-create -e repokey ...` - this will ask you to define a passphrase. Answer y or answer n when it asks to display the passphrase.

Use `borg repo-list ...` to access the repo. It will ask for a passphrase and under normal circumstances, it MUST NOT display what you entered. But there should be a way for optional "passphrase debugging" and it displaying the passphrase and its hex encoding.

alighazi288:
This is exactly how I've been testing. I was just uncertain whether the requirement was to always show debug information when a wrong passphrase is entered, or if it should be shown "optionally" based on the user's input. I chose to conditionally display it based on the new environment variable `BORG_DEBUG_PASSPHRASE`.

Now, debugging information (including passphrase and environment variable details) is only shown when the BORG_DEBUG_PASSPHRASE environment variable is set to "YES".

I hope this matches the intended behavior. Looking forward to your thoughts!

src/borg/helpers/passphrase.py
```diff
@@ -113,37 +109,28 @@ def verification(cls, passphrase):
...
-            env_var_value = os.environ.get(env_var)
-            if env_var_value is not None:
-                print(f'{env_var} = "{env_var_value}"', file=sys.stderr)
-            else:
-                print(f"# {env_var} is not set", file=sys.stderr)
+        if os.environ.get("BORG_DEBUG_PASSPHRASE") == "YES":
+            print("Incorrect passphrase!", file=sys.stderr)
+            print(f'Passphrase used (between double-quotes): "{passphrase}"', file=sys.stderr)
+            print(f'Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode("utf-8"))}', file=sys.stderr)
+            print("Relevant Environment Variables:", file=sys.stderr)
```
ThomasWaldmann, line 127: you could use a multi-line f-string and a single print here also.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,12 +1408,14 @@ def test_passphrase_new_retries(self, monkeypatch):
...
-    def test_passphrase_wrong(self, capsys, monkeypatch):
+    def test_passphrase_wrong_debug_enabled(self, capsys, monkeypatch):
+        monkeypatch.setenv("BORG_DEBUG_PASSPHRASE", "YES")
         monkeypatch.setenv("BORG_PASSPHRASE", "wrong_passphrase")
         monkeypatch.setenv("BORG_PASSCOMMAND", "echo wrong_passphrase")
         monkeypatch.setenv("BORG_PASSPHRASE_FD", "123")
 
         with pytest.raises(PassphraseWrong) as exc_info:
+            Passphrase.display_debug_info("wrong_passphrase")
             raise PassphraseWrong("wrong_passphrase")
```
ThomasWaldmann, line 1419: guess you just had removed the passphrase argument.

src/borg/testsuite/helpers_test.py
```diff
@@ -1428,6 +1430,17 @@ def test_passphrase_wrong(self, capsys, monkeypatch):
             "passphrase supplied in BORG_PASSPHRASE, by BORG_PASSCOMMAND or via BORG_PASSPHRASE_FD is incorrect."
         )
 
+    def test_passphrase_wrong_debug_disabled(self, capsys, monkeypatch):
+        monkeypatch.delenv("BORG_DEBUG_PASSPHRASE", raising=False)
+
+        with pytest.raises(PassphraseWrong):
+            Passphrase.display_debug_info("wrong_passphrase")
+            raise PassphraseWrong("wrong_passphrase")
```
ThomasWaldmann, line 1438: guess you just had removed the passphrase argument.

ThomasWaldmann (commented):
good to require that env var, so the stuff isn't printed by default.

src/borg/testsuite/helpers_test.py
alighazi288: I accidently overlooked that, thanks for pointing it out!

alighazi288:
@ThomasWaldmann Any suggestions for another issue I could work on next?

src/borg/helpers/passphrase.py
```diff
@@ -109,20 +109,33 @@ def verification(cls, passphrase):
...
-                    "UTF-8 encoding in hex together with the passphrase at a safe place.",
-                    file=sys.stderr,
-                )
+            pw_msg = f"""
+            Your passphrase (between double-quotes): "{passphrase}"
+            Make sure the passphrase displayed above is exactly what you wanted.
+            Your passphrase (UTF-8 encoding in hex): {bin_to_hex(passphrase.encode("utf-8"))}
+            It is recommended to keep the UTF-8 encoding in hex together with the passphrase at a safe place.
+            In case you should ever run into passphrase issues, it could sometimes help debugging them.
+            """
```
ThomasWaldmann, line 118: here you use a multi-line f-string (guess it needs a `dedent` wrapper to remove the leading spaces).

Note that you can use double-quoted strings inside that f-string.

src/borg/helpers/passphrase.py
```diff
@@ -109,20 +109,33 @@ def verification(cls, passphrase):
...
+                if env_var_value is not None:
+                    env_vars_str += f'{env_var} = "{env_var_value}"\n'
+                else:
+                    env_vars_str += f"# {env_var} is not set\n"
+            passphrase_info = (
+                f"Incorrect passphrase!\n"
+                f'Passphrase used (between double-quotes): "{passphrase}"\n'
+                f"Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode('utf-8'))}\n"
+                f"Relevant Environment Variables:\n"
+                f"{env_vars_str}"
```
ThomasWaldmann, line 136: You could use a multi-line f-string here also. And you can use double-quotes inside that without a problem. Also dedent.

src/borg/helpers/passphrase.py
```diff
@@ -109,20 +109,33 @@ def verification(cls, passphrase):
...
+            Make sure the passphrase displayed above is exactly what you wanted.
+            Your passphrase (UTF-8 encoding in hex): {bin_to_hex(passphrase.encode("utf-8"))}
+            It is recommended to keep the UTF-8 encoding in hex together with the passphrase at a safe place.
+            In case you should ever run into passphrase issues, it could sometimes help debugging them.
+            """
+            print(pw_msg, file=sys.stderr)
+
+    @staticmethod
+    def display_debug_info(passphrase):
+        if os.environ.get("BORG_DEBUG_PASSPHRASE") == "YES":
```
ThomasWaldmann, line 123: You only print something if that env var is set.

So, you could just call this function unconditionally without using a `debug_enabled` flag, see the 2 callers of this. The check is duplicated there.

ThomasWaldmann (changes requested):
consistency / simplification suggestions.

src/borg/helpers/passphrase.py
alighazi288: I initially tried using `textwrap.dedent`, but it caused alignment issues when displaying the environment variables on line 136. To fix this, I switched to explicit string construction, which ensured consistent formatting. If using an f-string is crucial, I can explore alternative implementations to fetch and display the environment variables correctly.

src/borg/helpers/passphrase.py
```diff
@@ -109,13 +110,15 @@ def verification(cls, passphrase):
             retry=True,
             env_var_override="BORG_DISPLAY_PASSPHRASE",
         ):
-            pw_msg = f"""
+            pw_msg = textwrap.dedent(
+                f"""
```
ThomasWaldmann, line 114: BTW, this introduces a newline at the start of the string.

Sometimes that is useful, but if it is not, then rather use:
```
f"""\
```

src/borg/helpers/passphrase.py
ThomasWaldmann: Ah, guess that is because the env var lines are differently indented as the f-string contents.

We can fix that easily and it also makes it a bit more pretty:

To format a single env var, use a local function like:
```
def fmt(env_var):
    value = os.environ.get(env_var)
    if value is not None:
        return f'{env_var} = "{value}"'
    else:
        return f"# {env_var} is not set"
```

Then call that from inside the main f-string, like:
```
passphrase_info = textwrap.dedent(
    f"""\
    Incorrect passphrase!
    Passphrase used (between double-quotes): "{passphrase}"
    Same, UTF-8 encoded, in hex: {bin_to_hex(passphrase.encode('utf-8'))}
    Relevant Environment Variables:
    {fmt(BORG_PASSPHRASE)}
    {fmt(BORG_PASSCOMMAND)}
    {fmt(BORG_PASSPHRASE_FD)}
    """)
```

ThomasWaldmann (changes requested):

src/borg/helpers/passphrase.py
alighazi288: Thank you so much for this information!

ThomasWaldmann:
The docs CI failure seems unrelated to your PR changes.

Maybe some stuff in sphinx changed, so it is triggering new warnings.

Guess I need to change so that code unit tests do not depend on docs test.

alighazi288:
@ThomasWaldmann  if this is ready to close, do you have any suggestions for another issue I could work on next?

ThomasWaldmann:
I've fixed the docs issue in master.

Please update your local master branch, then rebase the PR branch onto master, then force push to github.

In the issue tracker, some stuff is labelled as easy, good first issue or help wanted - choose from these if you feel it is a good fit for you.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+        out, err = capsys.readouterr()
+        assert 'Your passphrase (between double-quotes): "test_passphrase"' in err
+        assert "UTF-8 encoding in hex" in err
+
+        # Case 3: Non-ASCII passphrase
+        passphrase = "tëst_päṩṩ"
+        hex_value = passphrase.encode("utf-8").hex()
+        Passphrase.verification(passphrase)
+        out, err = capsys.readouterr()
+        assert hex_value in err
```
ThomasWaldmann, line 1463: We removed the special case for non-ascii, so this test is a bit off. Guess it should test with a ascii passphrase.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+        assert str(exc_info.value) == (
+            "passphrase supplied in BORG_PASSPHRASE, by BORG_PASSCOMMAND or via BORG_PASSPHRASE_FD is incorrect."
+        )
+
+    def test_passphrase_wrong_debug_disabled(self, capsys, monkeypatch):
+        monkeypatch.delenv("BORG_DEBUG_PASSPHRASE", raising=False)
+
+        with pytest.raises(PassphraseWrong):
+            Passphrase.display_debug_info("wrong_passphrase")
+            raise PassphraseWrong()
```
ThomasWaldmann, line 1438: also strange.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+
+        out, err = capsys.readouterr()
+        assert "Incorrect passphrase!" not in err
+        assert 'Passphrase used (between double-quotes): "wrong_passphrase"' not in err
```
ThomasWaldmann, line 1442: would be easier/safer to just assert "wrong_passphrase" not in err.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+
+    def test_verification(self, capsys, monkeypatch):
+        passphrase = "test_passphrase"
+
+        monkeypatch.setenv("BORG_DISPLAY_PASSPHRASE", "no")
+        Passphrase.verification(passphrase)
+        out, err = capsys.readouterr()
+        assert "Your passphrase (between double-quotes)" not in err
```
ThomasWaldmann, line 1450: assert passphrase not in err

that's the main thing we want to check without depending on the specific phrasing.

src/borg/testsuite/helpers_test.py
ThomasWaldmann: or even: assert passphrase not in err (using a variable, like you did below).

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+
+        monkeypatch.setenv("BORG_DISPLAY_PASSPHRASE", "yes")
+        Passphrase.verification(passphrase)
+        out, err = capsys.readouterr()
+        assert 'Your passphrase (between double-quotes): "test_passphrase"' in err
+        assert "UTF-8 encoding in hex" in err
```
ThomasWaldmann, line 1456: assert passphrase in err
assert passphrase_hex in err

(do not depend on phrasing, just check the important stuff)

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,85 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+
+        # Case 3: Non-ASCII passphrase
+        passphrase = "tëst_päṩṩ"
+        hex_value = passphrase.encode("utf-8").hex()
+        Passphrase.verification(passphrase)
+        out, err = capsys.readouterr()
+        assert hex_value in err
+
+    def test_display_debug_info(self, capsys, monkeypatch):
```
ThomasWaldmann, line 1465: looks like this has some overlap with `test_passphrase_wrong_debug_*` test methods, please review / consolidate.

ThomasWaldmann (changes requested):
found quite some more todo in the tests.

alighazi288:
I realized I completely forgot to update the tests after changing my implementation—apologies for that oversight! I've reviewed and adjusted everything now, so the tests should align correctly with the updated implementation. Let me know if there's anything else you'd like me to refine.

src/borg/testsuite/helpers_test.py
```diff
@@ -1408,6 +1408,45 @@ def test_passphrase_new_retries(self, monkeypatch):
...
+        out, err = capsys.readouterr()
+
+        assert "Incorrect passphrase!" not in err
+        assert passphrase not in err
+
+    def test_verification(self, capsys, monkeypatch):
+        passphrase = "test_passphrase"
+        hex_value = passphrase.encode("utf-8").hex()
+
+        monkeypatch.setenv("BORG_DISPLAY_PASSPHRASE", "no")
```
ThomasWaldmann, line 1439: maybe use delenv here. users will rarely set this to no, but we need to check that the default is like no.

alighazi288: I encountered an issue when implementing this change. Using `delenv` causes the test to wait for user input, creating an infinite loop.

Do you think mocking the yes function in `verification` to return False by default is the way to fix it?

ThomasWaldmann: Ehrm, sorry, did not consider that it is asking interactively in that case.

So guess just keeping the "no" in the env var is fine.

alighazi288: Reverted back to the previous commit.

ThomasWaldmann (approved):
LGTM!

PR #8620, opened 2025-01-04
docs: borg serve: recommend using a simple shell (closes #3818)

### LINKED ISSUE #3818: __builtin_bswap32 referenced symbol not found
The latest borg git repo refers to the __builtin_bswap32 symbol when built on Big Endian architectures without using gcc.  (Solaris 10, Sun's compiler).
```
% borg
  File "...python/borg-env/bin/borg", line 7, in <module>
    from borg.archiver import main
  File "...python/borg-env/lib/python3.6/site-packages/borg/archiver.py", line 36, in <module>
    from . import helpers
  File ".../borg-env/lib/python3.6/site-packages/borg/helpers.py", line 44, in <module>
    from . import hashindex
ImportError: ld.so.1: python3.6: fatal: relocation error:
    file .../python/borg-env/lib/python3.6/site-packages/borg/hashindex.cpython-36m.so:
    symbol __builtin_bswap32: referenced symbol not found
```

The macros in src/borg/_endian.h are to blame.

qyanu:
see https://github.com/borgbackup/borg/issues/8318

As a short summary of issue 8318:
openssh executes the user's configured shell with the whole command (either remote-command or forced-command) as a single argument like so: `"$SHELL" -c "$COMMAND"`.
This entails a larger attack surface, when a complex and feature-rich shell is configured for the user.

Because in the example configuration of a push backup server a complex and feature-rich shell is not needed, recommending the usage of a simple shell like /bin/sh will reduce the opportunity of yet-undiscovered security vulnerabilities to arise in the future.

### DIFF
--- a/docs/usage/serve.rst
+++ b/docs/usage/serve.rst
@@ -48,6 +48,7 @@ Details about sshd usage: `sshd(8) <https://www.openbsd.org/cgi-bin/man.cgi/Open
 
 SSH Configuration
 ~~~~~~~~~~~~~~~~~
+
 ``borg serve``'s pipes (``stdin``/``stdout``/``stderr``) are connected to the ``sshd`` process on the server side. In the event that the SSH connection between ``borg serve`` and the client is disconnected or stuck abnormally (for example, due to a network outage), it can take a long time for ``sshd`` to notice the client is disconnected. In the meantime, ``sshd`` continues running, and as a result so does the ``borg serve`` process holding the lock on the repository. This can cause subsequent ``borg`` operations on the remote repository to fail with the error: ``Failed to create/acquire the lock``.
 
 In order to avoid this, it is recommended to perform the following additional SSH configuration:
@@ -74,3 +75,16 @@ This will cause the server to send a keep alive to the client every 10 seconds. 
 If you then run borg commands with ``--lock-wait 600``, this gives sufficient time for the borg serve processes to terminate after the SSH connection is torn down after the 300 second wait for the keepalives to fail.
 
 You may, of course, modify the timeout values demonstrated above to values that suit your environment and use case.
+
+When the client is untrusted, it is a good idea to set the backup
+user's shell to a simple implementation (``/bin/sh`` is only an example and may or may
+not be such a simple implementation)::
+
+  chsh -s /bin/sh BORGUSER
+
+Because the configured shell is used by `openssh <https://www.openssh.com/>`_
+to execute the command configured through the ``authorized_keys`` file
+using ``"$SHELL" -c "$COMMAND"``,
+setting a minimal shell implementation reduces the attack surface
+compared to when a feature-rich and complex shell implementation is
+used.


docs/usage/serve.rst
```diff
@@ -74,3 +75,14 @@ This will cause the server to send a keep alive to the client every 10 seconds.
 If you then run borg commands with ``--lock-wait 600``, this gives sufficient time for the borg serve processes to terminate after the SSH connection is torn down after the 300 second wait for the keepalives to fail.
 
 You may, of course, modify the timeout values demonstrated above to values that suit your environment and use case.
+
+On a dedicated borg backup server, it is recommended to set the
```
ThomasWaldmann, line 79: Guess the **server** does not need to be "dedicated" for that, but the **user** should be dedicated to borg backups.

docs/usage/serve.rst
```diff
@@ -74,3 +75,14 @@ This will cause the server to send a keep alive to the client every 10 seconds.
...
+backup user's shell to a simple implementation::
+
+  chsh -s /bin/sh BORGUSER
```
ThomasWaldmann, line 82: On debian, this links to dash.

Is `/bin/sh` pointing to a "simple shell" on all linux dists?

docs/usage/serve.rst
```diff
@@ -74,3 +75,14 @@ This will cause the server to send a keep alive to the client every 10 seconds.
...
+
+Because the configured shell is used by `openssh <https://www.openssh.com/>`_
+to execute the command configured through the ``authorized_keys`` file
+using ``"$SHELL" -c "$COMMAND"``,
+setting a minimal shell implementation reduces the attack surface
+associated with a feature-rich complex code base.
```
ThomasWaldmann, line 88: It's not clear here how there can be an attack, considering `authorized_keys` is controlled by the server admin / hosting company, if users shall be forced to only run "borg serve".

In case of shellshock it worked via an environment variable content coming from the client side.

ThomasWaldmann (commented):
Thanks for the PR!

docs/usage/serve.rst
qyanu: Yes. I have reformulated the sentence. Now, the text does not mention a dedicated backup server, or a dedicated backup user. Rather, the text now refers to the consideration that the client is untrusted.

In this case, that the client is untrusted, using a simple shell implementation reduces the attack surface.

{As a personal note: i would recommend a dedicated backup server, and dedicated backup user, and always treat all clients as untrusted as much as possible. But all this is probably not a good fit for the borg documentation}

docs/usage/serve.rst
qyanu: I honestly do not have good expertise on this question, there are so many different distributions! I have reformulated the sentence such, that /bin/sh may not actually be such a simple implementation, even though it's used as an example.

For debian, i am confident when i say that dash is a simpler implementation than bash. And I am also confident to say, that other considerations being equal, a simpler software has the tendency to contain less security vulnerabilities than a more complex software.
However, in a specific comparison of two implementations, it is not really possible to guarantee that the simple implementation is a more secure choice than the complex impementation; Even a complex software can be maintained with rigour, and likewise a simple software can be neglected.

But, all things considered and even though i know of no actual security issue with using bash, given the choice, i would still give the general recommendation using dash over bash when configuring a dedicated backup user.

docs/usage/serve.rst
qyanu: So the theory goes like this:

openssh's sshd does all the checks, and then prepares to execute the command using the user's shell. This command is forced to a certain value through authorized_key's command= option. The ssh client still has certain powers over the environment variables.

As the user shell is a full program, it is theoretically possible that some feature or behaviour exists in the user shell that gives the client more control over what actions are performed on the backup server than was intended by the server admin.

While bash's shellshock vulnerability is indeed what i have in mind, i really did not attempt to produce any proof-of-concept that these earlier bash versions were actually exploitable in the context of a dedicated borg user with `borg serve` as forced ssh command. In any case, as fas as i know, current bash versions are not vulnerable to any environment manipulations that are possible by a ssh client in the scenario of a `command="borg serve",restrict ssh-rsa AAAAB3[..]` setup.

However, even though there is no actually exploitable current vulnerability known, i would still recommend to anyone who asks, that certain hardening is performed just-in-case.

In this regard, i think that the borg plays a role as i would assume that the documentation is used as a copy&paste source of server-configuration.
  If it is possible to present a copy&paste-able configuration that is preferential in some hypothetical cases but still does not impede all the features, then i would advocate for recommending it even though the scenario might never actually materialise.

qyanu: > It's not clear here how there can be an attack, considering `authorized_keys` is controlled by the server admin / hosting company, if users shall be forced to only run "borg serve".

PS: the technical nitpicking being, that the user is not forced to run "borg serve", but is forced to run the user shell with a specific forced argument. The user shell then decides whether (only) "borg serve" is run and no unintended side-effects happen.

docs/usage/serve.rst
ThomasWaldmann: Good change, "untrusted client" is on-point.

docs/usage/serve.rst
```diff
@@ -74,3 +75,16 @@ This will cause the server to send a keep alive to the client every 10 seconds.
 If you then run borg commands with ``--lock-wait 600``, this gives sufficient time for the borg serve processes to terminate after the SSH connection is torn down after the 300 second wait for the keepalives to fail.
 
 You may, of course, modify the timeout values demonstrated above to values that suit your environment and use case.
+
+When the client is untrusted, it is a good idea to set the backup
+user's shell to a simple implementation (The example /bin/sh may or may
```
ThomasWaldmann, line 80:
```suggestion
user's shell to a simple implementation (``/bin/sh`` is only an example and may or may
```

ThomasWaldmann (commented):
Added a small phrasing suggestion (also removes a typo), but other than that I think it's good for merging now.

ThomasWaldmann (approved):
LGTM, thanks!

Issue #8624, opened 2025-01-05
Allow timespan to be specified with common time units


KenKundert:
Currently the timespan specifications (--older, --newer, --oldest, --newest) are limited to be given in whole days or months.  This seams reasonable for backup scenarios where the backups occur on the order of once per day.  However, other scenarios are possible.  For example, one may want to take snapshots every few minutes to make it easy to recover from mistakes made during the day.  In these cases, a maximum resolution of 1 day seems too coarse.  Furthermore, supporting only two of many possible time units forces the user to remember those few that are supported.

I suggest that you support all common time units, including years (y), months (M), weeks (w), days (d), hours (h), minutes (m), and seconds (s).

As an example of the utility of this enhancement request, I run backups every 10 minutes and recently recovered a file from an archive that was 30 minutes old.

In addition, I am writing a program that runs borg for me, and from a convenience perspective, I would just like to specify the timespan to seconds so as to avoid all issues with resolution and fractional time units.

ThomasWaldmann:
Note: for consistency, units need to be Y/m/d and H/M/S (time library and other borg options use that convention).

KenKundert:
Other that format strings (ex: '%Y-%m-%dT%H:%M:%S.%f') I cannot find where the convention you mention is used.  I was looking because you left out weeks, and I was trying to determine if that was intentional or not.  Am I missing something, or are you trying to be compatible with the old strftime format strings?

ThomasWaldmann:
My initial idea considered strftime formatting convention, but as @jdchristensen pointed out that borg uses lowercase "y" elsewhere already [not Y like in strftime].

borg cli has some short options named after that convention (not for all, but for the frequently used stuff). Adding weeks as unit is fine.

For prune, we even have 2 ideas of what a "quarter" is since recently: 13weekly or 3monthly.

KenKundert:
I can add quarters if you like.

I have noticed a slight inconsistency between --newer, --newest, --older, --oldest and prune's --keep-within.  With the former, month and year are defined literally, so if you specify 1m you get the same day on the previous month even if the month does not have 31 days, whereas with --keep-within a month is defined as 31 days. Same with years on leap years.

This inconsistency seems natural and I don't think it is worth fretting over.  I bring it up because it might affect how quarters is implemented.  To be consistent with the current approach, it feels like a quarter should be defined as 3 months for --newer, ..., and it should be defined as 13 weeks for --keep-within.

[...]

PR #8626, opened 2025-01-06
Allow timespan to be specified with common time units

KenKundert:
https://github.com/borgbackup/borg/issues/8624

### DIFF
--- a/docs/usage/general/date-time.rst.inc
+++ b/docs/usage/general/date-time.rst.inc
@@ -13,5 +13,6 @@ Internally, we store and process date and time as UTC.
 .. rubric:: TIMESPAN
 
 Some options accept a TIMESPAN parameter, which can be given as a
-number of days (e.g. ``7d``) or months (e.g. ``12m``).
+number of years (e.g. ``2Y``), months (e.g. ``12m``), days (e.g. ``7d``),
+hours (e.g. ``8H``), minutes (e.g. ``30M``), or seconds (e.g. ``84600s``).
 
--- a/docs/usage/general/date-time.rst.inc
+++ b/docs/usage/general/date-time.rst.inc
@@ -12,7 +12,7 @@ Internally, we store and process date and time as UTC.
 
 .. rubric:: TIMESPAN
 
-Some options accept a TIMESPAN parameter, which can be given as a
-number of years (e.g. ``2Y``), months (e.g. ``12m``), days (e.g. ``7d``),
-hours (e.g. ``8H``), minutes (e.g. ``30M``), or seconds (e.g. ``84600s``).
-
+Some options accept a TIMESPAN parameter, which can be given as a number of
+years (e.g. ``2y``), months (e.g. ``12m``), weeks (e.g. ``2w``),
+days (e.g.  ``7d``), hours (e.g. ``8H``), minutes (e.g. ``30M``),
+or seconds (e.g. ``150S``).
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -126,26 +126,40 @@ def positive_int_validator(value):
 
 
 def interval(s):
-    """Convert a string representing a valid interval to a number of hours."""
-    multiplier = {"H": 1, "d": 24, "w": 24 * 7, "m": 24 * 31, "y": 24 * 365}
+    """Convert a string representing a valid interval to a number of seconds."""
+    seconds_in_a_minute = 60
+    seconds_in_an_hour = 60 * seconds_in_a_minute
+    seconds_in_a_day = 24 * seconds_in_an_hour
+    seconds_in_a_week = 7 * seconds_in_a_day
+    seconds_in_a_month = 31 * seconds_in_a_day
+    seconds_in_a_year = 365 * seconds_in_a_day
+    multiplier = dict(
+        S = 1,
+        M = seconds_in_a_minute,
+        H = seconds_in_an_hour,
+        d = seconds_in_a_day,
+        w = seconds_in_a_week,
+        m = seconds_in_a_month,
+        y = seconds_in_a_year,
+    )
 
     if s.endswith(tuple(multiplier.keys())):
         number = s[:-1]
         suffix = s[-1]
     else:
-        # range suffixes in ascending multiplier order
-        ranges = [k for k, v in sorted(multiplier.items(), key=lambda t: t[1])]
-        raise argparse.ArgumentTypeError(f'Unexpected interval time unit "{s[-1]}": expected one of {ranges!r}')
+        raise argparse.ArgumentTypeError(
+            f'Unexpected time unit "{s[-1]}": choose from {", ".join(multiplier)}'
+        )
 
     try:
-        hours = int(number) * multiplier[suffix]
+        seconds = int(number) * multiplier[suffix]
     except ValueError:
-        hours = -1
+        seconds = -1
 
-    if hours <= 0:
-        raise argparse.ArgumentTypeError('Unexpected interval number "%s": expected an integer greater than 0' % number)
+    if seconds <= 0:
+        raise argparse.ArgumentTypeError(f'Invalid number "{number}": expected positive integer')
 
-    return hours
+    return seconds
 
 
 def ChunkerParams(s):
@@ -565,7 +579,10 @@ class Location:
 def location_validator(proto=None, other=False):
     def validator(text):
         try:
-            loc = Location(text, other=other)
+            try:
+                loc = Location(text, other=other)
+            except Exception as e:
+                assert False
         except ValueError as err:
             raise argparse.ArgumentTypeError(str(err)) from None
         if proto is not None and loc.proto != proto:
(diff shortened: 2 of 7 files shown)


KenKundert:
This version supports Y/m/d and H/M/S as you specified.  Please let me know if you want to support weeks as well.  Also, let me know you want make all units other than m and M case insensitive.

jdchristensen:
I think we should match how `borg prune --keep-within` works, which allows weeks.  It's probably case sensitive, but someone could check the code.  If we support minutes and seconds here, we should probably do the same with `--keep-within` as well.
```
The --keep-within option takes an argument of the form “<int><char>”, where char is “H”, “d”, “w”, “m”, “y”. For example, --keep-within 2d means to keep all archives that were created within the past 48 hours. “1m” is taken to mean “31d”. 
```

ThomasWaldmann:
Yeah, what @jdchristensen says. Consistency within borg is of course more important than consistency to strftime formats (which most users are not aware of anyway).

So, ymwd lowercase, HMS uppercase.

docs/usage/general/date-time.rst.inc
```diff
@@ -13,5 +13,6 @@ Internally, we store and process date and time as UTC.
 .. rubric:: TIMESPAN
 
 Some options accept a TIMESPAN parameter, which can be given as a
-number of days (e.g. ``7d``) or months (e.g. ``12m``).
+number of years (e.g. ``2Y``), months (e.g. ``12m``), days (e.g. ``7d``),
+hours (e.g. ``8H``), minutes (e.g. ``30M``), or seconds (e.g. ``84600s``).
```
ThomasWaldmann, line 17: S must be uppercase to comply with our convention.

src/borg/helpers/parseformat.py
```diff
@@ -579,10 +579,10 @@ def validator(text):
 
 
 def relative_time_marker_validator(text: str):
-    time_marker_regex = r"^\d+[md]$"
+    time_marker_regex = r"^\d+[YmdHMS]$"
```
ThomasWaldmann, line 582: like @jdchristensen said: rather ymd HMS and also w for week.

src/borg/helpers/time.py
```diff
@@ -131,6 +131,14 @@ def calculate_relative_offset(format_string, from_ts, earlier=False):
                 return from_ts + timedelta(days=offset)
             elif unit == "m":
                 return offset_n_months(from_ts, offset)
+            elif unit == "Y":
+                return from_ts.replace(year=from_ts.year + offset)
+            elif unit == "H":
+                return from_ts + timedelta(seconds=offset * 60 * 60)
+            elif unit == "M":
+                return from_ts + timedelta(seconds=offset * 60)
+            elif unit == "S":
```
ThomasWaldmann, line 140: correct.

src/borg/helpers/time.py
```diff
@@ -131,6 +131,14 @@ def calculate_relative_offset(format_string, from_ts, earlier=False):
                 return from_ts + timedelta(days=offset)
             elif unit == "m":
                 return offset_n_months(from_ts, offset)
+            elif unit == "Y":
```
ThomasWaldmann, line 134: y

ThomasWaldmann (changes requested):
consistency feedback. thanks for the PR!

src/borg/helpers/time.py
```diff
@@ -131,6 +131,16 @@ def calculate_relative_offset(format_string, from_ts, earlier=False):
...
+            elif unit == "y":
+                return from_ts.replace(year=from_ts.year + offset)
+            elif unit == "w":
+                return from_ts + timedelta(days=offset * 7)
+            elif unit == "H":
+                return from_ts + timedelta(seconds=offset * 60 * 60)
+            elif unit == "M":
+                return from_ts + timedelta(seconds=offset * 60)
+            elif unit == "S":
+                return from_ts + timedelta(seconds=offset)
```
ThomasWaldmann, line 143: We could have this if/elif block ordered from big to small time units.

ThomasWaldmann:
BTW, I just saw that you are working in your local master branch. This can often be a problem (e.g. if you need to rebase onto current upstream master for some reason), so try to remember to always checkout a feature/fix branch first (like `git checkout -b myfeature`).

It looks like we can quickly merge this PR, so no need to change branches here.

KenKundert:
I have enhanced 'prune --keep-within' to support seconds and minutes and checked in the changes.  I believe I have also made all requested changes.

I cannot run tests locally, there seems to be some issue with the version, the tests all end in a failure complaining about an invalid version number when I run tests locally.  So I am just waiting for you to approve the workflows so I can see how many mistakes I made.

Thanks for the suggestion to create and operate on a feature branch.  I have limited experience contributing to other projects.

src/borg/helpers/parseformat.py
```diff
@@ -126,26 +126,40 @@ def positive_int_validator(value):
...
 
     if s.endswith(tuple(multiplier.keys())):
         number = s[:-1]
         suffix = s[-1]
     else:
-        # range suffixes in ascending multiplier order
-        ranges = [k for k, v in sorted(multiplier.items(), key=lambda t: t[1])]
-        raise argparse.ArgumentTypeError(f'Unexpected interval time unit "{s[-1]}": expected one of {ranges!r}')
+        raise argparse.ArgumentTypeError(
+            f'Unexpected time unit "{s[-1]}": choose from {", ".join(multiplier)}'
```
ThomasWaldmann, line 151: are the units still sorted in the message?

ThomasWaldmann: and considering iso-8601 (and also some other popular date/time conventions), sorting like ymwdHMS is nicer than SMHdwmy.

src/borg/helpers/parseformat.py
```diff
@@ -565,7 +579,10 @@ def with_timestamp(self, timestamp):
 def location_validator(proto=None, other=False):
     def validator(text):
         try:
-            loc = Location(text, other=other)
+            try:
+                loc = Location(text, other=other)
+            except Exception as e:
+                assert False
```
ThomasWaldmann, line 585: Hmm?

src/borg/testsuite/helpers_test.py
```diff
@@ -553,17 +553,25 @@ def test_prune_split_no_archives():
...
-def test_interval(timeframe, num_hours):
-    assert interval(timeframe) == num_hours
+@pytest.mark.parametrize("timeframe, num_hours", [
+    ("5S", 5),
+    ("2M", 120),
+    ("1H", 3600),
+    ("1d", 86_400),
+    ("1w", 604_800),
+    ("1m", 2_678_400),
+    ("1y", 31_536_000)
```
ThomasWaldmann, line 563: I'ld usually just put "2M": 2 * 60 here instead of pre-calculating that number.

src/borg/testsuite/helpers_test.py
```diff
@@ -553,17 +553,25 @@ def test_prune_split_no_archives():
...
 
 @pytest.mark.parametrize(
     "invalid_interval, error_tuple",
     [
-        ("H", ('Unexpected interval number "": expected an integer greater than 0',)),
-        ("-1d", ('Unexpected interval number "-1": expected an integer greater than 0',)),
-        ("food", ('Unexpected interval number "foo": expected an integer greater than 0',)),
+        ("H", ('Unexpected number "": expected positive integer',)),
+        ("-1d", ('Unexpected number "-1": expected positive integer',)),
+        ("food", ('Unexpectedinterval number "foo": expected positive integer',)),
```
ThomasWaldmann, line 574: needs fixing

ThomasWaldmann (changes requested):
Some minor stuff I found.

src/borg/testsuite/helpers_test.py
```diff
@@ -553,15 +553,18 @@ def test_prune_split_no_archives():
...
+@pytest.mark.parametrize(
+    "timeframe, num_hours",
+    [
+        ("5S", 5),
+        ("2M", 2 * 60),
+        ("1H", 60 * 60),
+        ("1d", 24 * 60 * 60),
+        ("1w", 7 * 24 * 68 * 60),
+        ("1m", 31 * 24 * 68 * 60),
+        ("1y", 365 * 24 * 68 * 60),
```
ThomasWaldmann, line 565: 1w 1m 1y have a typo: 68 instead of 60

ThomasWaldmann (changes requested):
found a typo

src/borg/testsuite/helpers_test.py
```diff
@@ -553,17 +553,28 @@ def test_prune_split_no_archives():
     assert kept_because == {}
 
 
-@pytest.mark.parametrize("timeframe, num_hours", [("1H", 1), ("1d", 24), ("1w", 168), ("1m", 744), ("1y", 8760)])
-def test_interval(timeframe, num_hours):
-    assert interval(timeframe) == num_hours
+@pytest.mark.parametrize(
+    "timeframe, num_hours",
```
ThomasWaldmann, line 557: -> num_secs

KenKundert:
I finally was able to fetch the tags and the tests now are working locally.  So going forward there should be many fewer of my stupid mistakes to review.  I apologize for the churn.

PR #8630, opened 2025-01-10
update prune documentation for new --keep-within intervals

KenKundert:
Update the prune doc to describe the new --keep-within intervals.


ThomasWaldmann:
Thanks!

Issue #5638, opened 2021-01-20
restore big files/block devices with considering already present data


Lanozavr:
I use borg to make backups of my LVM volumes. It works fine at backup creation time but not as good when I do 'borg extract'. It pulls many gigabytes of data from remote borg repo server although I still have an almost unchanged device in place.

It would be nice to have an option to not just extract data from a borg repo but to also check against an existing file/block device, if it already has the required data and only retrieve changed data from the remote repo.

ThomasWaldmann:
Yeah, that would be good.

Notable:
- when writing to a block device, one likely would use `borg extract --stdout ... > /dev/blkxxx` to directly extract to the block device and when doing that, borg does not even know where the data is going / what it could compare against
- for similar reasons `borg create --stdin ...` might be used at backup time - in this case we might not have the blockdevice name inside the archive
- this is the simpler part of a more complex "missing feature": that of bringing a local data set into an archived state (here, simple: 1 file == a block device, more complex case: a filesystem with existing directories and existing files)
- borg create already has `--read-special` to indicate that special files should actually have their content read, not just the device file archived. in that case, we have blockdevice names (relative though) in the archive, but at extraction time, desired device names could be different

ThomasWaldmann:
If A is archived data stream and F is the fs data stream:

- iterate over A chunks list to get chunkid_A, size
- read size bytes from F, compute chunkid_F
- if chunkid_A == chunkid_F, we already have the correct data, next...
- if not, read <size> chunk from A, write to F (seek to correct pos first), next...
- at the end of A, truncate F to the same size.

ThomasWaldmann:
@alighazi288 sure, feel free to work on this in master branch.

Don't use block devices, but just normal files with multiple chunks. To test, use chunker settings that create rather small chunks (e.g. 512 bytes), so that we don't need very big files to test this.

Maybe it is simplest to even use the "fixed" chunker when starting to work on tests for this. As the block size is fixed, it is easy to produce a file with specific block content and then do changes in known blocks and check how it behaves.

alighazi288:
Thank you so much for the information! Also, any particular reason for working directly in master instead of a feature branch?

alighazi288:
Hey @ThomasWaldmann. Opened up a PR (#9703). Would love a review when you get the chance.

PR #8632, opened 2025-01-12
Implement chunk comparison and selective extraction for borg extract (#5638)

alighazi288:
## Archive File Chunk Comparison and Extraction

This implementation provides efficient file restoration from archives by comparing and extracting chunks. Instead of blindly extracting entire files, it:

1. Compares existing file content with archived chunks
2. Only fetches and updates chunks that differ
3. Handles various edge cases:
   - Partial chunks at end of files
   - Files longer/shorter than archive version
   - Empty files
   - Cross-chunk boundary changes

### DIFF
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -127,9 +127,9 @@ def with_repository(
             )
 
             with repository:
-                if repository.version not in (3,):
+                if repository.version not in (4,):
                     raise Error(
-                        f"This borg version only accepts version 3 repos for -r/--repo, "
+                        f"This borg version only accepts version 4 repos for -r/--repo, "
                         f"but not version {repository.version}. "
                         f"You can use 'borg transfer' to copy archives from old to new repos."
                     )
@@ -194,10 +194,10 @@ def with_other_repository(manifest=False, cache=False, compatibility=None):
             )
 
             with repository:
-                acceptable_versions = (1,) if v1_legacy else (3,)
+                acceptable_versions = (1,) if v1_legacy else (4,)
                 if repository.version not in acceptable_versions:
                     raise Error(
-                        f"This borg version only accepts version {' or '.join(acceptable_versions)} "
+                        f"This borg version only accepts version {' or '.join(str(v) for v in acceptable_versions)} "
                         f"repos for --other-repo."
                     )
                 kwargs["other_repository"] = repository
--- a/src/borg/repoobj.py
+++ b/src/borg/repoobj.py
@@ -13,11 +13,14 @@ AUTHENTICATED_NO_KEY = "authenticated_no_key" in workarounds
 OBJ_MAGIC = b"BORG_OBJ"
 OBJ_VERSION = 0x01
 
+# Fixed header size per blob: OBJ_MAGIC(8) + version(1) + chunk_id(32) + meta_size(4) + data_size(4)
+REPOOBJ_HEADER_SIZE = 49
+
 
 class RepoObj:
-    # Object header: magic (8b), format version (1b), meta size (4b), data size (4b).
-    obj_header = Struct("<8sBII")
-    ObjHeader = namedtuple("ObjHeader", "magic version meta_size data_size")
+    # Object header: magic (8b), format version (1b), chunk_id (32b), meta size (4b), data size (4b).
+    obj_header = Struct("<8sB32sII")
+    ObjHeader = namedtuple("ObjHeader", "magic version chunk_id meta_size data_size")
 
     @classmethod
     def extract_crypted_data(cls, data: bytes) -> bytes:
@@ -72,7 +75,7 @@ class RepoObj:
         data_encrypted = self.key.encrypt(id, data_compressed)
         meta_packed = msgpack.packb(meta)
         meta_encrypted = self.key.encrypt(id, meta_packed)
-        hdr = self.ObjHeader(OBJ_MAGIC, OBJ_VERSION, len(meta_encrypted), len(data_encrypted))
+        hdr = self.ObjHeader(OBJ_MAGIC, OBJ_VERSION, id, len(meta_encrypted), len(data_encrypted))
         hdr_packed = self.obj_header.pack(*hdr)
         return hdr_packed + meta_encrypted + data_encrypted
 
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -115,15 +115,13 @@ class Repository:
             location = Location(url)
         self._location = location
         self.url = url
-        # lots of stuff in data: use 2 levels by default (data/00/00/ .. data/ff/ff/ dirs)!
-        data_levels = int(os.environ.get("BORG_STORE_DATA_LEVELS", "2"))
         ns_config = {
             "archives/": {"levels": [0]},
             "cache/": {"levels": [0]},
             "config/": {"levels": [0]},
-            "data/": {"levels": [data_levels]},
             "keys/": {"levels": [0]},
             "locks/": {"levels": [0]},
+            "packs/": {"levels": [1]},
         }
         # Get permissions from parameter or environment variable
         permissions = permissions if permissions is not None else os.environ.get("BORG_REPO_PERMISSIONS", "all")
@@ -136,9 +134,9 @@ class Repository:
                 "archives": "lrw",
                 "cache": "lrwWD",  # WD for chunks.<HASH>, last-key-checked, ...
                 "config": "lrW",  # W for manifest
-                "data": "lrw",
                 "keys": "lr",
                 "locks": "lrwD",  # borg needs to create/delete a shared lock here
+                "packs": "lrw",
             }
         elif permissions == "write-only":  # mostly no reading
             permissions = {
@@ -146,9 +144,9 @@ class Repository:
                 "archives": "lw",
                 "cache": "lrwWD",  # read allowed, e.g. for chunks.<HASH> cache
                 "config": "lrW",  # W for manifest
-                "data": "lw",  # no r!
                 "keys": "lr",
                 "locks": "lrwD",  # borg needs to create/delete a shared lock here
+                "packs": "lw",  # no r!
             }
         elif permissions == "read-only":  # mostly r/o
             permissions = {"": "lr", "locks": "lrwD"}
@@ -171,7 +169,7 @@ class Repository:
         self._send_log = send_log_cb or (lambda: None)
         self.do_create = create
         self.created = False
-        self.acceptable_repo_versions = (3,)
+        self.acceptable_repo_versions = (4,)
         self.opened = False
         self.lock = None
         self.do_lock = lock
@@ -209,10 +207,10 @@ class Repository:
         self.store.open()
         try:
             self.store.store("config/readme", REPOSITORY_README.encode())
-            self.version = 3
+            self.version = 4
             self.store.store("config/version", str(self.version).encode())
             self.store.store("config/id", bin_to_hex(os.urandom(32)).encode())
-            # we know repo/data/ still does not have any chunks stored in it,
+            # we know repo/packs/ still does not have any chunks stored in it,
             # but for some stores, there might be a lot of empty directories and
             # listing them all might be rather slow, so we better cache an empty
             # ChunkIndex from here so that the first repo operation does not have
@@ -327,22 +325,21 @@ class Repository:
         def check_object(obj):
             """Check if obj looks valid."""
             hdr_size = RepoObj.obj_header.size
-            obj_size = len(obj)
-            if obj_size >= hdr_size:
-                hdr = RepoObj.ObjHeader(*RepoObj.obj_header.unpack(obj[:hdr_size]))
-                if hdr.magic != OBJ_MAGIC:
-                    log_error("invalid object magic.")
-                elif hdr.version != OBJ_VERSION:
-                    log_error(f"unsupported object version: {hdr.version}.")
-                else:
-                    meta = obj[hdr_size : hdr_size + hdr.meta_size]
-                    if hdr.meta_size != len(meta):
-                        log_error("metadata size mismatch.")
-                    data = obj[hdr_size + hdr.meta_size : hdr_size + hdr.meta_size + hdr.data_size]
-                    if hdr.data_size != len(data):
-                        log_error("data size mismatch.")
-            else:
+            if len(obj) < hdr_size:
                 log_error("too small.")
+                return
+            hdr = RepoObj.ObjHeader(*RepoObj.obj_header.unpack(obj[:hdr_size]))
+            if hdr.magic != OBJ_MAGIC:
+                log_error("invalid object magic.")
+            elif hdr.version != OBJ_VERSION:
+                log_error(f"unsupported object version: {hdr.version}.")
+            else:
+                meta = obj[hdr_size : hdr_size + hdr.meta_size]
+                if hdr.meta_size != len(meta):
+                    log_error("metadata size mismatch.")
+                data = obj[hdr_size + hdr.meta_size : hdr_size + hdr.meta_size + hdr.data_size]
+                if hdr.data_size != len(data):
+                    log_error("data size mismatch.")
 
         # TODO: progress indicator, ...
         partial = bool(max_duration)
@@ -376,11 +373,11 @@ class Repository:
         # As we don't do garbage collection here, this is not a problem.
         # We also don't know the plaintext size, so we set it to 0.
         init_entry = ChunkIndexEntry(flags=ChunkIndex.F_USED, size=0)
-        infos = self.store.list("data")
+        infos = self.store.list("packs")
         try:
             for info in infos:
                 self._lock_refresh()
-                key = "data/%s" % info.name
+                key = "packs/%s" % info.name
                 if key <= last_key_checked:  # needs sorted keys
                     continue
                 try:
@@ -412,8 +409,9 @@ class Repository:
                     # add all existing objects to the index.
                     # borg check: the index may have corrupted objects (we did not delete them)
                     # borg check --repair: the index will only have non-corrupted objects.
-                    id = hex_to_bin(info.name)
-                    chunks[id] = init_entry
+                    pack_id = hex_to_bin(info.name)
+                    chunk_id = pack_id  # N=1: chunk_id == pack_id
+                    chunks[chunk_id] = init_entry
                 now = time.monotonic()
                 if now > t_last_checkpoint + 300:  # checkpoint every 5 mins
                     t_last_checkpoint = now
@@ -437,7 +435,7 @@ class Repository:
                         self, chunks, incremental=False, clear=True, force_write=True, delete_other=True
                     )
         except StoreObjectNotFound:
-            # it can be that there is no "data/" at all, then it crashes when iterating infos.
+            # it can be that there is no "packs/" at all, then it crashes when iterating infos.
             pass
         logger.info(f"Checked {objs_checked} repository objects, {objs_errors} errors.")
         if objs_errors == 0:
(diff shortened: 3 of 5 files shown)


alighazi288:
@ThomasWaldmann Since this is a significant change, I wanted to open this PR as a draft to get your feedback before proceeding further.

Could you please review the current implementation and provide any suggestions or improvements?

ThomasWaldmann (commented):
can you add some tests, so that codecov does not spam the whole diff view?

alighazi288:
Done.

src/borg/testsuite/archive_test.py
```diff
@@ -402,3 +407,134 @@ def test_reject_non_sanitized_item():
     for path in rejected_dotdot_paths:
         with pytest.raises(ValueError, match="unexpected '..' element in path"):
             Item(path=path, user="root", group="root")
+
+
+def test_compare_and_extract_chunks(tmpdir, monkeypatch):
```
ThomasWaldmann, line 412: maybe use less c&p for the tests. the code in these is very similar, so you could replace it by generic code and parametrize that.

ThomasWaldmann (changes requested):
some first feedback

src/borg/archive.py
```diff
@@ -719,6 +719,78 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
                 # In this case, we *want* to extract twice, because there is no other way.
                 pass
 
+    def compare_and_extract_chunks(self, item, fs_path):
+        print(f"Initial fs_path: {fs_path}")
+        print(f"self.cwd: {self.cwd}")
```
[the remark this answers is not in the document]
alighazi288: The prints were for me (testing). I guess we don't need them in the final code?

alighazi288:
@ThomasWaldmann  I'm having trouble with the `test_extract_continue` test in `extract_cmd_test.py`. The test expects a new inode when extracting a truncated hardlinked file. I've tried adding a check for hardlinks in `compare_and_extract_chunks` to skip in-place modification but I feel like that defeats the purpose of the enhancement.

Looking at extract_item, I'm unsure about the right approach (the command seems to be working fine in practical testing). I have pushed my current implementation if you'd like to take a look.

ThomasWaldmann (commented):
some feedback

src/borg/archive.py
```diff
@@ -719,6 +719,40 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
                 # In this case, we *want* to extract twice, because there is no other way.
                 pass
 
+    def compare_and_extract_chunks(self, item, fs_path):
+        """Compare file chunks and patch if needed. Returns True if patching succeeded."""
+        try:
+            st = os.stat(fs_path, follow_symlinks=False)
```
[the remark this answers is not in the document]
alighazi288: I'm sorry but I'm still a bit confused about this. Are you referring to:
- Size tracking and verification
- Damaged chunks detection
- Progress reporting
- Sparse file support
- Proper attribute restoration
I can see the loophole now that if `compare_and_extract_chunks` returns `True`, it does not take into account the things mentioned above whereas they are present in a "normal" extraction.

src/borg/archive.py
```diff
@@ -719,6 +719,40 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+                    fs_file.seek(chunk_offset)
+                    data_F = fs_file.read(size)
+
+                    needs_update = True
+                    if len(data_F) == size:
+                        chunkid_F = self.key.id_hash(data_F)
+                        needs_update = chunkid_A != chunkid_F
+
+                    if needs_update:
+                        chunk_data = b"".join(self.pipeline.fetch_many([chunkid_A], ro_type=ROBJ_FILE_STREAM))
```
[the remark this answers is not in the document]
alighazi288: You're absolutely right. Didn't really see it that way before. Thank you for the feedback!

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
-                return True
+            # Second pass: Update file and consume EVERY chunk from the iterator
+            offset = 0
+            item_chunk_size = 0
+            with backup_io("open"):
+                fs_file = open(fs_path, "rb+")
+            with fs_file:
+                for fs_chunk, item_chunk in zip(fs_chunks, item.chunks):
+                    with backup_io("seek"):
+                        fs_file.seek(offset)
```
ThomasWaldmann, line 767: I think it should seek in the if-block and write in the else-block.

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+                    if fs_chunk is not None and fs_chunk.id == item_chunk.id:
+                        offset += item_chunk.size
+                        item_chunk_size += item_chunk.size
+                    else:
+                        chunk_data = next(chunk_data_iter)
+                        if pi:
+                            pi.show(increase=len(chunk_data), info=[remove_surrogates(item.path)])
```
ThomasWaldmann, line 774: guess progress is made in the if-block also, so maybe do that after the if/else.

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+                        with backup_io("write"):
+                            if sparse and not chunk_data.strip(b"\0"):
+                                fs_file.seek(len(chunk_data), 1)  # Seek over sparse section
+                                offset += len(chunk_data)
```
ThomasWaldmann, line 778: Guess we can only do it that way when writing to a new file.
But as you are writing to an existing file, you'll need to overwrite existing content with the zeros.

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+            # Fetch all needed chunks and iterate through ALL of them
+            chunk_data_iter = self.pipeline.fetch_many(
+                [chunk.id for chunk in needed_chunks], is_preloaded=True, ro_type=ROBJ_FILE_STREAM
+            )
 
-                fs_file.truncate(item.size)
-                return True
+            # Second pass: Update file and consume EVERY chunk from the iterator
+            offset = 0
+            item_chunk_size = 0
```
ThomasWaldmann, line 761: do they always have the same value?

ThomasWaldmann: also: better name instead of item_chunk_size?

ThomasWaldmann (changes requested):

src/borg/archive.py
alighazi288: You are right that both variables track the same value. I initially thought we needed separate variables because `offset` is for file positioning while `item_chunk_size` is for tracking total processed data, but since they increment the same way, we could use just `offset`.

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+        try:
+            # First pass: Build fs chunks list
+            fs_chunks = []
+            offset = 0
+            with backup_io("open"):
+                fs_file = open(fs_path, "rb")
+            with fs_file:
+                for chunk in item.chunks:
+                    with backup_io("seek"):
+                        fs_file.seek(offset)
```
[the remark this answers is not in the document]
alighazi288: I didn't consider that we're just reading chunks and read() automatically advances the file pointer. I will remove it.

src/borg/archive.py
```diff
@@ -719,38 +719,82 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
-                        chunk_data = b"".join(self.pipeline.fetch_many([chunkid_A], ro_type=ROBJ_FILE_STREAM))
-                        fs_file.seek(chunk_offset)
-                        fs_file.write(chunk_data)
+            if not needed_chunks:
+                return True
 
-                    chunk_offset += size
+            # Fetch all needed chunks and iterate through ALL of them
+            chunk_data_iter = self.pipeline.fetch_many(
+                [chunk.id for chunk in needed_chunks], is_preloaded=True, ro_type=ROBJ_FILE_STREAM
```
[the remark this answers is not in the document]
alighazi288: I thought I was fetching all chunks but I was wrong. I see now that I only fetch the mismatched chunks.

src/borg/archive.py
alighazi288: I tried your suggestion but without an explicit seek before write, the file contents got corrupted (wrote chunks in wrong positions). Adding the seek fixed it.

ThomasWaldmann (changes requested):

src/borg/archive.py
```diff
@@ -719,11 +719,10 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
                 # In this case, we *want* to extract twice, because there is no other way.
                 pass
 
-    def compare_and_extract_chunks(self, item, fs_path, st=None, *, pi=None, sparse=False):
+    def compare_and_extract_chunks(self, item, fs_path, *, pi=None, sparse=False):
         """Compare file chunks and patch if needed. Returns True if patching succeeded."""
-        if st is None or not stat.S_ISREG(st.st_mode):
+        if not os.path.exists(fs_path):
             return False
```
ThomasWaldmann, line 725: That's worse because it does another OS call (guess os.path.exists internally just does a stat call) for something it already knows (it already did a stat call as I pointed out in last review).

src/borg/archive.py
```diff
@@ -742,49 +741,38 @@ def compare_and_extract_chunks(self, item, fs_path, st=None, *, pi=None, sparse=
...
-                        offset += item_chunk.size
+                            fs_file.seek(item_chunk.size, 1)
                     else:
                         chunk_data = next(chunk_data_iter)
-                        # Need explicit seek before write in rb+ mode to ensure correct file position Without an
-                        # explicit seek before write, the file contents got corrupted (wrote chunks in wrong positions)
+
                         with backup_io("seek"):
-                            fs_file.seek(offset)
+                            fs_file.seek(fs_file.tell())
```
ThomasWaldmann, line 759: Why is that seek needed?

src/borg/archive.py
```diff
@@ -742,49 +741,38 @@ def compare_and_extract_chunks(self, item, fs_path, st=None, *, pi=None, sparse=
...
 
                         with backup_io("write"):
-                            if sparse and not chunk_data.strip(b"\0"):
-                                fs_file.write(b"\0" * len(chunk_data))  # Write zeros to overwrite old content
-                                offset += len(chunk_data)
-                            else:
-                                fs_file.write(chunk_data)
-                                offset += len(chunk_data)
-                    if pi:  # <-- Moved outside to track progress in both cases
+                            data = b"\0" * len(chunk_data) if sparse and not chunk_data.strip(b"\0") else chunk_data
```
ThomasWaldmann, line 762: isn't the if-block value always the same as the value of chunk_data?

ThomasWaldmann (changes requested):

src/borg/archive.py
alighazi288: @ThomasWaldmann The issue is  if I try to use `st` directly, I get an `UnboundLocalError` since `st` is only set inside the try block.

If I try to handle the stat info differently (always pass `st` from extract_item), I run into other test failures (symlink tests, directory timestamp tests, etc). Hence the additional OS call.

Could you clarify how I can properly access and use the `st` value as you suggested?

ThomasWaldmann: line 870 catches all OSErrors and just ignores them and leaves st undefined.

I guess we need to improve that:
```
st = None  # None means either there is no such file or we could not access it
```

ThomasWaldmann: For the case when it rmdirs an existing directory (so it can freshly extract the one from the archive), you can also add a `st = None` to that block, because after the rmdir it is the same fs state as if that dir never existed.

alighazi288: Great! Thanks! Had the same idea but wanted to double check before I touched existing code.

src/borg/archive.py
```diff
@@ -719,6 +719,63 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+            chunk_data_iter = self.pipeline.fetch_many([chunk.id for chunk in needed_chunks], ro_type=ROBJ_FILE_STREAM)
+
+            # Second pass: Update file
+            with backup_io("open"):
+                fs_file = open(fs_path, "rb+")
+            with fs_file:
+                for fs_chunk, item_chunk in zip(fs_chunks, item.chunks):
+                    if fs_chunk == item_chunk:
+                        with backup_io("seek"):
+                            fs_file.seek(item_chunk.size, 1)
```
ThomasWaldmann, line 754: use `os.SEEK_CUR` instead of 1.

src/borg/archive.py
```diff
@@ -719,6 +719,63 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+                    else:
+                        chunk_data = next(chunk_data_iter)
+
+                        with backup_io("write"):
+                            fs_file.write(chunk_data)
+                    if pi:
+                        pi.show(increase=len(chunk_data), info=[remove_surrogates(item.path)])
```
ThomasWaldmann, line 761: that's incorrect, you need to consider all cases.

src/borg/archive.py
```diff
@@ -719,6 +719,63 @@ def extract_helper(self, item, path, hlm, *, dry_run=False):
...
+
+                final_size = fs_file.tell()
+                with backup_io("truncate_and_attrs"):
+                    fs_file.truncate(item.size)
+                    fs_file.flush()
+                    self.restore_attrs(fs_path, item, fd=fs_file.fileno())
```
ThomasWaldmann, line 767: use the same code as you see in 899+.

src/borg/archive.py
```diff
@@ -802,12 +859,14 @@ def same_item(item, st):
                 return  # done! we already have fully extracted this file in a previous run.
             elif stat.S_ISDIR(st.st_mode):
                 os.rmdir(path)
+                st = None
             else:
                 os.unlink(path)
+                st = None
```
ThomasWaldmann, line 865: did you think about this?

ThomasWaldmann (changes requested):

src/borg/archive.py
alighazi288: @ThomasWaldmann I didn't realize that this would break the `continue_extraction` functionality. The issue is `compare_and_extract_chunks` still tries to use stale st info after the file is unlinked.

I have tried:
- Tracking unlink state with flags
- Checking inode/link count
- Modifying comparison logic

The only fix in my mind is the extra OS call to check the file's existence. Maybe I'm missing something?

ThomasWaldmann: If the fs file is a normal file, your code requires it to be there, so it can be "updated" - thus it must not be removed.

ThomasWaldmann: And we also need to think about what to do with existing metadata, like acls, xattrs, fs flags, ... - the current code assumes that there is no existing metadata and just adds the stuff from the archive item.

alighazi288: @ThomasWaldmann another solution I can think of is:
```python
try:
    st = os.stat(path, follow_symlinks=False)
    if continue_extraction and same_item(item, st):
        return  # done! we already have fully extracted this file in a previous run.
    if stat.S_ISREG(st.st_mode) and not continue_extraction:
        if self.compare_and_extract_chunks(item, path, st=st, pi=pi):
            return
    elif stat.S_ISDIR(st.st_mode):
        os.rmdir(path)
    else:
        os.unlink(path)
```
This way I'm only attempting in-place updates when it's a regular file and not in continue_extraction mode. The function returns True if the update succeeded, in which case we're done. Otherwise (returns False), we fall back to the original remove/recreate behavior.

Also, since I'm using `restore_attrs()` just like the existing code and not handling metadata directly at all, shouldn't it be consistent with how Borg already works?

ThomasWaldmann: As I already said: the restore_attrs code expects a fresh state of the file (like newly created, no acls, no xattrs) and just adds the stuff from the archived item.

But if you are updating an existing file, there can be already acls or xattrs that do not match what's in the archive (and what shall be the final state).

alighazi288: @ThomasWaldmann This is getting a bit too advanced for my understanding, but I've still tried to implement and verify the attribute restoration. I'm still stuck on the `recreate_cmd` test failures - would appreciate some guidance there.

alighazi288:
@ThomasWaldmann This is getting a bit too advanced for my understanding, but I've still tried to implement and verify the attribute restoration. I'm still stuck on the recreate_cmd test failures - would appreciate some guidance there.

alighazi288:
@ThomasWaldmann any advice?

src/borg/archive.py
```diff
@@ -747,29 +747,60 @@ def compare_and_extract_chunks(self, item, fs_path, *, st, pi=None):
...
                 with backup_io("truncate_and_attrs"):
-                    fs_file.truncate(item.size)
+                    pos = item_chunks_size = fs_file.tell()
+                    fs_file.truncate(pos)
                     fs_file.flush()
+
+                    if not self.noacls:
+                        try:
+                            # Clear ACLs by setting empty ACL entries
+                            acl_set(fs_path, {"acl_entries": []}, self.numeric_ids, fd=fs_file.fileno())
```
ThomasWaldmann, line 772: you didn't read the acl_set code, did you?

src/borg/archive.py
```diff
@@ -747,29 +747,60 @@ def compare_and_extract_chunks(self, item, fs_path, *, st, pi=None):
...
+                            set_flags(fs_path, 0, fd=fs_file.fileno())
+                        except OSError:
+                            pass
+                    if not self.noxattrs:
+                        try:
+                            # Clear ALL xattrs
+                            attrs = xattr.listxattr(fs_file.fileno(), follow_symlinks=False)
+                            for attr in attrs:
+                                try:
+                                    xattr.setxattr(fs_file.fileno(), attr, b"", follow_symlinks=False)
```
ThomasWaldmann, line 788: that does not do what you intended. setting an xattr to an empty bytestring  is not the same as removing an xattr.

src/borg/archive.py
```diff
@@ -747,29 +747,60 @@ def compare_and_extract_chunks(self, item, fs_path, *, st, pi=None):
...
+                                except OSError as e:
+                                    if e.errno != xattr.ENOATTR:  # Ignore if already removed
+                                        raise
+                        except OSError as e:
+                            if e.errno != errno.ENOTSUP:  # Ignore if xattrs not supported
+                                raise
                     self.restore_attrs(fs_path, item, fd=fs_file.fileno())
```
ThomasWaldmann, line 795: you could create a method clear_attrs close to the definition of restore_attrs and move that code you added above to there.

src/borg/archive.py
```diff
@@ -747,29 +747,60 @@ def compare_and_extract_chunks(self, item, fs_path, *, st, pi=None):
             # Second pass: Update file
             with backup_io("open"):
                 fs_file = open(fs_path, "rb+")
-            with fs_file:
```
ThomasWaldmann: you removed the usage of the fs_file's context manager (removed the "with"). that means that it does not close the fs_file anymore (as it did when leaving the with-block), but I guess it should close the file.

src/borg/archive.py
ThomasWaldmann: you are calling `fs_file.fileno()` way too often.

ThomasWaldmann (changes requested):
some stuff i found.

ThomasWaldmann:
BTW, an alternative way to solve the "we need default / cleared metadata" issue would be to NOT reuse the complete local file (metadata + data), but create a new temp file and then either:
- copy some content from the existing local file
- copy some content from the repo

After that, apply the metadata to the temp file and finally, replace the existing local file by the temp file.

ThomasWaldmann:

PR #8640, opened 2025-01-18
docs: fix udev rule priority in automated-local.rst

### LINKED ISSUE #8639: Automated backups as explained in documentation do not work on NixOS (and possibly other distributions)
After following the instructions in the [documentation](https://borgbackup.readthedocs.io/en/stable/deployment/automated-local.html) for setting up automated backup as soon as the USB storage is connected, the udev rule is never triggered on NixOS 24.11 and thus the backup is not started.

The reason is that `/etc/backups/40-backup.rules` has too high priority, before `50-udev-default.rules` and `60-persistent-storage.rules`. I suspect that some parts of the environment that are matched against are not populated when the rule is applied.

The fix is to change the priority by renaming the file to `80-backup.rules`.

httpdev:
fixes #8639

The priority of 40 for the udev rules as stated in to documentation applies the rule too early on some systems, which prevents the rule from matching. This commit changes the priority to 80.

### DIFF
--- a/docs/deployment/automated-local.rst
+++ b/docs/deployment/automated-local.rst
@@ -32,7 +32,7 @@ All configuration goes into this directory.
 Find out the ID of the partition table of your backup disk (here assumed to be /dev/sdz):
     lsblk --fs -o +PTUUID /dev/sdz
 
-Then, create ``/etc/backups/40-backup.rules`` with the following content (all on one line)::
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Thanks! Can you please also do a backport to 1.4-maint branch?

PR #8646, opened 2025-01-19
CI: ARM Linux (master)

ThomasWaldmann:


ThomasWaldmann:
Guess we wait a bit before merging this, the ARM64 VM (running on a MS Cobalt 100 128-Core CPU) seems a bit slower than the others, maybe because this is a preview still with limited resources.

ThomasWaldmann:
As of 2025-03-01, the arm runner is still way slower than the x64-64 runner. 👎

ThomasWaldmann:
Guess we need to give up here. Performance on ARM is still 4x slower than on the intel/amd machine. :-(

Nobody wants to wait 45 minutes until the ARM stuff completes if the other stuff is ready after 11 mins.

PR #8673, opened 2025-02-19
Fix #8522: Add {unixtime} placeholders

### LINKED ISSUE #8522: Please add {unixtime} to placeholders

Yes

ISSUE

What kind of information?

Borg 1.4.0, and I've searched the master branch for this functionality.

Debian 12 (Bookworm)

It varies, btrfs for source and ext4 for target (borg repository).

Between eight and nine years of backups, which is the longest I've ever stuck with the same backup system.  I just rant a "compact" operation and the current stats are:
```
All archives:                6.00 TB              4.86 TB              1.63 TB
```

NA

Well, I'm surprised borg doesn't have this feature!  As for the practical use, I find it easier to read narrow lines, which is allegedly the case for most people, and `{unixtime}` is the shortest recognisable way that I know of to insure unique archive names.  It also actually means something when it's reversed, unlike pushing something unique through a hash function and then truncating the hash...which would also probably solve the unique archive name problem, although I'd worry about collisions.

Yes

#### Solution

```
int(time.time())
````

I hope you agree this sounds useful and fun to implement.

### PR DESCRIPTION
**Pull request by abhijeetsharma200**

### CODE REVIEW

File: src/borg/helpers/parseformat.py [hunk 1]
```diff
@@ -280,6 +281,7 @@ def _replace_placeholders(text, overrides={}):
         "hostname": hostname,
         "now": DatetimeWrapper(current_time.astimezone()),
         "utcnow": DatetimeWrapper(current_time),
+        "unixtime": int(time.time()),
```
ThomasWaldmann, line 284: Looks ok, but we maybe could save some lib/os calls by deriving the timestamp from current_time (see line 276).

Also better consistency in (exotic) cases when someone uses multiple time-related placeholders.

ThomasWaldmann: https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

abhijeetsharma200: Thanks! I have made the changes as suggested.

ThomasWaldmann: Thanks, looks good!

Did you manually try the new placeholder using the latest code?

abhijeetsharma200: Yup, it worked!

ThomasWaldmann:
Thanks! Can you please do a backport to 1.4-maint branch?

RanaFahad01:
> Thanks! Can you please do a backport to 1.4-maint branch?

Sure, I have never done a backport before so just to confirm I branch of 1.4 and make the same changes right?

abhijeetsharma200:
> Thanks! Can you please do a backport to 1.4-maint branch?

Sure! I’ve never done a backport before. Just to confirm, I should branch off from 1.4, apply the same changes, and create a PR targeting 1.4. Is that correct?

ThomasWaldmann:
one usually would:
- git checkout 1.4-maint
- git checkout -b feature_8522_add_unixtime_to_placeholders-1.4
- git cherry-pick (add hash here for the changeset in master branch)

PR #8675, opened 2025-02-25
Fix #8522: Add {unixtime} placeholders (1.4-maint)

### LINKED ISSUE #8522: Please add {unixtime} to placeholders

abhijeetsharma200:


src/borg/helpers/parseformat.py
```diff
@@ -212,18 +212,19 @@ def replace_placeholders(text, overrides={}):
...
+        "hostname": hostname,
+        "now": DatetimeWrapper(current_time.astimezone()),
+        "utcnow": DatetimeWrapper(current_time),
+        "unixtime": int(current_time.timestamp()),
+        "user": getosusername(),
+        "uuid4": str(uuid.uuid4()),
+        "borgversion": borg_version,
+        "borgmajor": "%d" % borg_version_tuple[:1],
+        "borgminor": "%d.%d" % borg_version_tuple[:2],
+        "borgpatch": "%d.%d.%d" % borg_version_tuple[:3],
```
ThomasWaldmann, line 227: borg master branch uses black for code style, but 1.4 branch does not.

can you please only do the minimal change necessary?

abhijeetsharma200: Got rid of the code styles. Should now only be a single line change.

ThomasWaldmann:
Thanks!

Issue #8649, opened 2025-01-21
review "assert" usage


ThomasWaldmann:
See there for the problem description:

https://community.sonarsource.com/t/feature-python-assert-should-be-consider-harmful/38501

`assert` should be only used for:
- test suite
- at places in production code where it doesn't really matter if the assert is executed or not. it must never be used if not executing the assert would influence correctness or security.
- if it matters, the assert should be replaced by `if <condition>: raise SomeException`

So, the task for borg master branch is to do a systematic review and fix all problematic asserts.

In case we find a lot of places to fix, a quick workaround for 1.4-maint branch could be to disallow running borg with assertions switched off, something like:

```
+try:
+    assert False
+except AssertionError:
+    pass  # OK
+else:
+    print("Borg requires working assertions. Please run Python without -O and/or unset PYTHONOPTIMIZE")
+    sys.exit(2)
```

ThomasWaldmann:
The 2 places you found are to catch programming errors and are executed rather frequently. 

Thus, such a programming error likely would be catched before a release is made.

Guess `archiver.py` is the best place to put a quick fix into 1.4.

120EE0980:
@ThomasWaldmann I am working on replacing assert ..will soon open a PR.Do you also suggest that I make the quick fix into 1.4 in another PR right now or wait?

ThomasWaldmann:
Yeah, you could do the quick fix for 1.4 first, then work on master, fixing the asserts.

120EE0980:
@ThomasWaldmann After doing quick fix, when I ran `tox -e py39`, 1 test is failing..I am unable to debug. Please help.

<img width="1191" alt="Image" src="https://github.com/user-attachments/assets/c624332b-35ed-42bb-8f8a-42e4b953f82a" />

<img width="1196" alt="Image" src="https://github.com/user-attachments/assets/713e0b29-0340-4207-a870-ff508d682f99" />

ThomasWaldmann:
You need to run pytest with -vv to see the full diff.

120EE0980:
> You need to run pytest with -vv to see the full diff.

<img width="1197" alt="Image" src="https://github.com/user-attachments/assets/ed0a952c-2fea-4bf4-8bb7-9ff60c973d68" />

ThomasWaldmann:
Hmm, that does not look like it is related to the quick fix. Can you:
- reproduce the issue?
- is it only with py39?
- can you reproduce the issue without doing any own changes?

[...]

PR #8682, opened 2025-03-03
bugfix : review assert and replace by if condition

120EE0980:
Related to #8649 
This PR only incorporates very small part of the solution to the issue.(i.e only replaces assert in 2 files)
This is just a draft to confirm whether right path is being followed.(will push other changes after I get an OK from reviewer)


ThomasWaldmann:

ThomasWaldmann:
Updated and force-pushed:
- rebased on current master, fixed merge conflicts
- fixed the failing test (wrong exception)

Quite some tests broken due to issues with our requirements, this is unrelated to this PR.

PR #8685, opened 2025-03-08
quickfix :  borg exits when assertions are disabled with Python optimizations

120EE0980:
Quick-workaround for #8649

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3,6 +3,15 @@
 import sys
 import traceback
 
+# quickfix to disallow running borg with assertions switched off
+try:
+    assert False
+except AssertionError:
+    pass  # OK
+else:
+    print("Borg requires working assertions. Please run Python without -O and/or unset PYTHONOPTIMIZE.", file=sys.stderr)
+    sys.exit(2)  # == EXIT_ERROR
+
 try:
     import argparse
     import collections


src/borg/archiver.py
```diff
@@ -3,6 +3,15 @@
 import sys
 import traceback
 
+# quickfix to disallow running borg with assertions switched off
+try:
+    assert False
+except AssertionError:
+    pass   #OK
```
ThomasWaldmann, line 10: 2 blanks to the left of #, 1 blank to the right.

120EE0980: Done

120EE0980:
Should I add tests for this fix?(Is that why Codecov is showing missing coverage )

src/borg/archiver.py
```diff
@@ -3,6 +3,15 @@
 import sys
 import traceback
 
+# quickfix to disallow running borg with assertions switched off
+try:
+    assert False
+except AssertionError:
+    pass  # OK
+else:
+    print("Borg requires working assertions. Please run Python without -O and/or unset PYTHONOPTIMIZE")
```
ThomasWaldmann, line 12: Maybe it is better to output this on stderr.

ThomasWaldmann:
No, I don't think we need a test for this.

ThomasWaldmann:
Thanks!

Issue #8691, opened 2025-03-21
Borg 1.4 Fix for python3.13: ImportError: cannot import name 'pbkdf2_hmac'


dagelf:
Installing borg 1.4 on a machine with Python 3.13 gives this:

$ borg -V
Traceback (most recent call last):
  File "/home/coenraad/.venv/lib/python3.13/site-packages/borg/archiver.py", line 38, in <module>
    from .archive import Archive, ArchiveChecker, ArchiveRecreater, Statistics, is_special
  File "/home/coenraad/.venv/lib/python3.13/site-packages/borg/archive.py", line 24, in <module>
    from .cache import ChunkListEntry
  File "/home/coenraad/.venv/lib/python3.13/site-packages/borg/cache.py", line 31, in <module>
    from .crypto.key import PlaintextKey
  File "/home/coenraad/.venv/lib/python3.13/site-packages/borg/crypto/key.py", line 10, in <module>
    from hashlib import sha256, sha512, pbkdf2_hmac
ImportError: cannot import name 'pbkdf2_hmac' from 'hashlib' (/usr/lib/python3.13/hashlib.py)

$ sudo pacman -Ql borg|grep crypto/key.py
borg /usr/lib/python3.13/site-packages/borg/crypto/key.py
$ grep pbkdf /usr/lib/python3.13/site-packages/borg/crypto/key.py
from _hashlib import pbkdf2_hmac

The fix is to move the `pbkdf2_hmac` to a new line importing from _hashlib. 

```
import sys

if sys.version_info >= (3, 13):
    from _hashlib import pbkdf2_hmac
else:
    from hashlib import pbkdf2_hmac
```

dagelf:
Sorry about posting this here but I can't find the old/a borgbackup repo, without spending an hour or more now looking at where the various distros pull it from.

ThomasWaldmann:
This is the correct repo and issue tracker, the 1.4 code is in 1.4-maint branch.

But that "fix" looks very suspicious.

Considering that this function should be in hashlib (as documented there: https://docs.python.org/3/library/hashlib.html#hashlib.pbkdf2_hmac ), I would suspect that your Python maybe was not built correctly with OpenSSL?

ThomasWaldmann:
@dagelf ping?

dagelf:
Apparently the function moved to a Python internal in 3.13 - the following
is listed in the Changelog:
[...]

PR #8692, opened 2025-03-21
Update Borg 1.4 key.py for compatibility with Python 3.13

dagelf:
Fix for #8691

Also FYI seems https://borgbackup.readthedocs.io/en/latest/development.html#contributions is out of date, as it states:
```
Contributions

… are welcome!

Some guidance for contributors:

    Discuss changes on the GitHub issue tracker, on IRC or on the mailing list.

    Make your PRs on the master branch (see [Branching Model](https://borgbackup.readthedocs.io/en/latest/development.html#branching-model) for details).
```

But this is slated for removal in Borg 2 according to other issues. I will make a separate PR for clearer language above.

### DIFF
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -7,7 +7,12 @@ import shlex
 import sys
 import textwrap
 import subprocess
-from hashlib import sha256, sha512, pbkdf2_hmac
+from hashlib import sha256, sha512
+
+if sys.version_info >= (3, 13):
+    from _hashlib import pbkdf2_hmac
+else:
+    from hashlib import pbkdf2_hmac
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
See discussion in #8691.

dagelf:
Interesting. it failed for me on Archlinux the day I posted this, and I also tested with Python 3.13 on an old Ubuntu using pyenv.

dagelf:
Looks like the import failed due to an incompatible OpenSSL library dependency - so the right solution is to update OpenSSL (which probably happened in between me making this change making me falsely believe that this actually fixed it, My demands that I recreate the situation to confirm, will report on the linked ticket.)

PR #8718, opened 2025-04-04
implement padme chunk size obfuscation (SPEC 250), fixes #8705

div-dev123:
Added Padmé padding as a new obfuscation option with SPEC=250.

PR #8722, opened 2025-04-06
tests: ignore 'com.apple.provenance' xattr (macOS specific)

div-dev123:
Reference: #8718 

com.apple.provenance is an extended attribute that macOS assigns to files downloaded from the internet, mainly to track their origin.

### DIFF
--- a/src/borg/testsuite/xattr_test.py
+++ b/src/borg/testsuite/xattr_test.py
@@ -19,7 +19,7 @@ def tempfile_symlink(tmp_path):
 
 def assert_equal_se(is_x, want_x):
     # check 2 xattr lists for equality, but ignore security.selinux attr
-    is_x = set(is_x) - {b"security.selinux"}
+    is_x = set(is_x) - {b"security.selinux", b"com.apple.provenance"}
     want_x = set(want_x)
     assert is_x == want_x
 


div-dev123:
@ThomasWaldmann I have create a new PR just like you asked for the attribute fix

ThomasWaldmann:
The commit comment is wrong, was that your spellchecker or some AI?

Besides the wrong xattr name it mentions, you also never need to mention the names of the files you edited or repeat how you changed specific lines in the code - if somebody is interested in that, they can use `git diff`.

Most important in a commit comment is that it briefly describes what has changed and also why it was changed (if that isn't totally obvious). The borg changelog gets made from these commit comments, so commit comments directly usable for that are best.

div-dev123:
I think this is the same commit message you told me to use when I committed in the previous PR , if it needs to be changed I will do so

ThomasWaldmann:
Ehrm. The commit comment is ok, what I meant was the PR text (which is usually identical, but here it wasn't), which you edited now.

ThomasWaldmann:
Thanks for fixing this!

div-dev123:
> Ehrm. The commit comment is ok, what I meant was the PR text (which is usually identical, but here it wasn't), which you edited now.

Yess understood! Thank you!

Issue #8724, opened 2025-04-07
The `birthtime` metadata on macOS is imprecise


songxiaocheng:
Python's `st_birthtime_ns` attribute remains unavailable on macOS, despite its introduction in Python 3.12.

The current implementation approximates this value via floating-point arithmetic, which introduces precision loss: https://github.com/borgbackup/borg/blob/89daeabd7bd07af2dac201b42fc2b055a1c8440a/src/borg/archive.py#L1075-L1076

The exact birth time in nanosecond precision can instead be retrieved directly via macOS's `stat64` system call.

ThomasWaldmann:
Thanks for finding this and working on it!

I have looked at the PR code, but I think we should first discuss a bit about it:

## What is the use case for this?

Guess one approach is to "archive as much and as good as possible".

But besides that, what do we do with the archived birthtime?
- we could maybe LOOK at the birthtime via borg list or borg mount (not sure, needs checking whether that is possible)
- but we can NOT restore the birthtime with borg extract (os.utime just lets us set mtime and atime)

## What are the impacts of this?

- the current PR code does an additional `stat64` call on macOS, slowing down the performance (especially for unchanged files as the metadata is always read from the fs)
- the current PR code only works with the file NAME and not with the file FD, potentially re-introducing race conditions already fixed in other parts of the code, which work with the FD of the opened file (this could be fixed by adding code for FD)

## Other approaches?
- as ns birthtime is also available with py312 stat_result, we could have a long-term solution if we just check for that new attribute (no addtl. cost, as we do the normal `os.stat` call in any case). users of older pythons would have the worse resolution.
- we could replace our `os.stat` call by an own `os_stat` call into an improved implementation (but that's likely lots of work, duplicating the work done within CPython).

PR #8725, opened 2025-04-07
fix: retrieve `birthtime` in nanosecond precision via system call

songxiaocheng:
Related to #8724.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -15,7 +15,7 @@ from itertools import groupby, zip_longest
 from typing import Iterator
 from shutil import get_terminal_size
 
-from .platformflags import is_win32, is_darwin
+from .platformflags import is_win32
 from .logger import create_logger
 
 logger = create_logger()
@@ -33,7 +33,7 @@ from .hashindex import ChunkIndex, ChunkIndexEntry
 from .helpers import HardLinkManager
 from .helpers import ChunkIteratorFileWrapper, open_item
 from .helpers import Error, IntegrityError, set_ec
-from .platform import uid2user, user2uid, gid2group, group2gid
+from .platform import uid2user, user2uid, gid2group, group2gid, get_birthtime_ns
 from .helpers import parse_timestamp, archive_ts_now
 from .helpers import OutputTimestamp, format_timedelta, format_file_size, file_status, FileSize
 from .helpers import safe_encode, make_path_safe, remove_surrogates, text_to_json, join_cmd, remove_dotdot_prefixes
@@ -52,9 +52,6 @@ from .platform import acl_get, acl_set, set_flags, get_flags, swidth, hostname
 from .remote import RemoteRepository, cache_if_remote
 from .repository import Repository, NoManifestError
 from .repoobj import RepoObj
-
-if is_darwin:
-    from .platform import is_darwin_feature_64_bit_inode, get_birthtime_ns
 
 has_link = hasattr(os, "link")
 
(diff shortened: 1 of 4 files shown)


src/borg/archive.py
```diff
@@ -1072,8 +1075,11 @@ def stat_simple_attrs(self, st):
         if not self.noctime:
             attrs["ctime"] = safe_ns(st.st_ctime_ns)
         if not self.nobirthtime and hasattr(st, "st_birthtime"):
-            # sadly, there's no stat_result.st_birthtime_ns
-            attrs["birthtime"] = safe_ns(int(st.st_birthtime * 10**9))
+            if is_darwin:
+                attrs["birthtime"] = safe_ns(get_birthtime_ns(fd or path, follow_symlinks=False))
+            else:
+                # sadly, there's no stat_result.st_birthtime_ns
+                attrs["birthtime"] = safe_ns(int(st.st_birthtime * 10**9))
```
ThomasWaldmann, line 1082: guess the decision logic needs to be done a bit differently.

```
if not self.nobirthtime:
    if hasattr(st, "st_birthtime_ns"):  # py312+ ns resolution
        ...
    elif is_darwin:  # use our platform code, ns resolution
        ...
    elif hasattr(st, "st_birthtime"):  # s resolution
        ...
```

ThomasWaldmann (changes requested):

src/borg/archive.py
```diff
@@ -1074,10 +1074,12 @@ def stat_simple_attrs(self, st, path, fd=None):
...
-        if not self.nobirthtime and hasattr(st, "st_birthtime"):
-            if is_darwin:
+        if not self.nobirthtime:
+            if hasattr(st, "st_birthtime_ns"):
+                attrs["birthtime"] = safe_ns(st.st_birthtime_ns)
+            elif is_darwin and is_darwin_feature_64_bit_inode:
                 attrs["birthtime"] = safe_ns(get_birthtime_ns(fd or path, follow_symlinks=False))
-            else:
+            elif hasattr(st, "st_birthtime"):
                 # sadly, there's no stat_result.st_birthtime_ns
```
ThomasWaldmann, line 1083: guess this comment needs to be removed.

you could add a comment to 1078 noting that this was added in Python 3.12.

src/borg/platform/darwin.pyx
```diff
@@ -10,6 +10,18 @@ from .xattr import _listxattr_inner, _getxattr_inner, _setxattr_inner, split_str
 
 API_VERSION = '1.2_05'
 
+cdef extern from *:
```
ThomasWaldmann, line 13: interesting, didn't know that construction.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -579,20 +579,32 @@ def patched_setxattr_EACCES(*args, **kwargs):
...
+    def darwin_get_birthtime_ns(path):
+        st = os.stat(input_path)
+        if hasattr(st, "st_birthtime_ns"):
+            return st.st_birthtime_ns
+        elif is_darwin_feature_64_bit_inode:
+            return get_birthtime_ns(path, follow_symlinks=False)
+        elif hasattr(st, "st_birthtime"):
+            return int(st.st_birthtime * 10**9)
+        else:
+            return None
```
ThomasWaldmann, line 592: Instead of duplicating this, you could add it to `borg.helpers.fs` as `get_birthtime_ns` and call it from all places needing it.

The os.stat call needs to be removed for that and st, path, fd should be given as parameters.

Side note: we'll need another such helper in the future for windows as they are deprecating the usage of ctime for getting the file creation time on windows (in future one shall use birthtime there, too).

ThomasWaldmann (approved):
LGTM!

ThomasWaldmann:
Guess this is finished, right?

If you like, you could also work on a similar issue for windows ctime / birthtime, which could be solved by a similar wrapper as you did it here. See #8730.

ThomasWaldmann:
Thanks! This could get backported to 1.4-maint, so it'll get into a stable release soon (master branch getting released as borg2 will still take quite a while).

PR #8746, opened 2025-04-12
[DOCS] Fix minor typo (#8745)

### LINKED ISSUE #8745: Minor typo in the documentation for `borg config`
The description section of the `borg config` documentation mentions how to get the value of an existing key:
> To get and existing key, pass only the key name.

I assume the "and" should be replaced with "an" here

destructor64:
Simple PR to fix #8745 

Could probably be directly applied to older branches as well

### DIFF
--- a/docs/man/borg-config.1
+++ b/docs/man/borg-config.1
@@ -39,7 +39,7 @@ This command gets and sets options in a local repository or cache config file.
 For security reasons, this command only works on local repositories.
 .sp
 To delete a config value entirely, use \fB\-\-delete\fP\&. To list the values
-of the configuration file or the default values, use \fB\-\-list\fP\&.  To get and existing
+of the configuration file or the default values, use \fB\-\-list\fP\&.  To get an existing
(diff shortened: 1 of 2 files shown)


ThomasWaldmann (changes requested):
Thanks for the PR!

You didn't fix the authoritative place for that text though (find it in the .py files, usually in src/borg/archiver...).

The 2 files you edited are both generated files, you don't need to change / commit there. When I make a release, these 2 files will be generated from .py files.

destructor64:
Oh, didn't know that. However, there it is already fixed. See commit b379202beb4437050aabc46aa5481855f607a47a

destructor64:
That also explains why my search for further occurences in the whole source tree didn't yield any more hits

ThomasWaldmann:
Oh, sorry for the double work. Guess I also should re-generate these files now and then between releases.

ThomasWaldmann:
Already fixed.

PR #8747, opened 2025-04-12
Fix paper key import error message

destructor64:
I just experimented a bit with printable key ex/import and noticed the error message checking the number of slashes in the `id`  line being wrong.

This also applies to at least the 1.4 and 1.2 versions


ThomasWaldmann:
Good catch, thanks!

Yeah, it also applies to 1.2-maint and 1.4-maint, can you please also do the backport PRs?

ThomasWaldmann:
Thanks!

Issue #8771, opened 2025-04-19
SetuptoolsDeprecationWarning: `project.license`, `project.license-files`, "License classifiers"


FelixSchwarz:
When building the project in Fedora, I get these warnings:
```
[4/6] Cythonizing src/borg/crypto/low_level.pyx
/usr/lib/python3.13/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!

        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  corresp(dist, value, root_dir)
/usr/lib/python3.13/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: BSD License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  dist._finalize_license_expression()
/usr/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: BSD License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  self._finalize_license_expression()
```

ThomasWaldmann:
Yes, I am aware of these.

Should be fixed in master, but I guess not yet in 1.4-maint or 1.2-maint as project.license and project.license-files require a very recent setuptools (which might be not yet available everywhere).

ThomasWaldmann:
@FelixSchwarz unsure about the consequences of backporting this to 1.4-maint:

If a linux dist has borg 1.4.x and I raise the minimum requirement for setuptools, will that block adoption for that dist (because it provides only an older setuptools)?

FelixSchwarz:
[...]

PR #8774, opened 2025-04-19
pyproject.toml: SPDX expression for license, add license-files, see #8771

ThomasWaldmann:
Also: raise setuptools version requirement appropriately.


ThomasWaldmann:
Note: before backporting to 1.4-maint, consider whether the setuptools requirement creates issues.

Issue #8715, opened 2025-03-31
borg2: Add `date:` archive pattern


PhrozenByte:
**Rationale**

Borg 2 introduces the concept of archive series and discourages unique archive names. To work on a single archive the docs currently tell users to use `aid:` patterns. IMHO this feels a bit cumbersome.

The probably most common variant of creating unique archive names with Borg 1 are date patterns. I feel like that this still is the most practical way of identifying single archives. However, AFAIK there's currently no simple way of matching single archives in Borg 2 using date patterns. Options like `--oldest` et al. are designed to match multiple archives within time spans only. Tags lack the support of wildcards.

**Suggestion**

I thus like to suggest adding a `date:` archive pattern. This pattern matches a given ISO 8601 date time pattern against an archive's creation time. Since backups usually take longer than a second, a full ISO 8601 date time (e.g. `date:2025-03-31T14:24:27`) is practically guaranteed to match just a single archive. Whether a date-only pattern (e.g. `date:2025-03-31`) matches just a single archive depends on the user's usage pattern.

ISO 8601 date time strings are more powerful than just specifying a single moment of time: They can also be used to specify periods of time (both as explicit period, and implicitly as duration or time intervals). The `date:` archive pattern should support these as well (e.g. `date:2025-01-01T00:00:00/P3M` matching all archives created in Jan to Mar 2025). This also allows Borg 2 to remove the `--oldest` et al. options and use the new and more powerful `date:` archive pattern instead.

In regards to priorities: Since this wouldn't mean a breaking change (the `--oldest` et al. options could just be deprecated instead of being removed), it's not necessary to have this with Borg 2.0, if considered to be useful at all. It could also be implemented in steps: Supporting arbitrary wildcards in date time strings is harder than supporting incomplete date time strings. Supporting time periods explicitly is harder than supporting wildcards in date time strings. Adding support for the custom features necessary to replace the `--oldest` et al. options is even harder. It's not necessary to have everything from the start.

**Implementation details**

The `date:` archive pattern should support wildcards (e.g. `date:2025-*-01` should match all archives created at any month's first day in 2025). Incomplete ISO 8601 date times are treated as if any missing field was given as a wildcard, i.e. `date:2025` is equivalent to `date:2025-*-*T*:*:*`).

[...]

PR #8775, opened 2025-04-19
Prune by either int or interval for all retention policies

Goddesen:
This PR adds optional interval handling for all retention filter flags of the `prune` command, previously only available on `--keep-within`. E.g. `prune --keep-hourly=7d` will keep hourly archives for the last 7 days regardless of count. Adds `--keep` which acts as a combination of (and supersedes) `--keep-last` and `--keep-within`.

Implements #8637.

------

This has been a pet peeve of mine in the pruning command for a long time. In my mind the most clean backup regime keeps all backups for a short time (allows catching small errors quickly), then hourly backups for a reasonable time (say, 7 days), then daily backups for a little longer and then finally weekly/monthly as storage permits. This was not previously possible, requiring for example `--keep-within 7d --keep-hourly 168 --keep-daily 14 --keep-weekly -1` for an approximation. But keeping around 168 archives for a machine that's only running a few hours a day seems mighty excessive. So here we are :)
With this implemented my ideal retention for my primary laptop with archives every 15m looks something like `--keep 3d --keep-hourly 7d --keep-daily 30d --keep-weekly -1`.

PR #8776, opened 2025-04-19
Add preliminary support for ISO-8601 timestamps via date: archive match pattern (#8715)

c-herz:
This PR adds preliminary support for matching ISO 8601 timestamps with the `date:` archive filter, and intends to begin addressing the requirements of #8715.

~~Timestamps (except for Unix epoch forms) are currently assumed to be in the user's local timezone and converted to UTC internally. The following formats are currently supported:

1. `YYYY`
2. `YYYY-MM` 
3. `YYYY-MM-DD`  
4. `YYYY-MM-DDTHH`         -> matches a 1-hour interval  
5. `YYYY-MM-DDTHH:MM`      -> matches a 1-minute interval  
6. `YYYY-MM-DDTHH:MM:SS`   -> matches a 1-second interval  
7. `YYYY-MM-DDTHH:MM:SS.ffff` -> matches an *exact* timestamp, including up to microseconds
8. `@123456789`            -> Unix epoch (interpreted as UTC)

Edit: Scoped down significantly from original PR; future improvements can be made in future PRs.

### DIFF
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -330,26 +330,41 @@ class HelpMixIn:
             This is very powerful, but can also get rather complicated.
 
         Date patterns, selector ``date:``
-            Match archives by creation timestamp. Supported forms are:
+            Match archives by creation timestamp. You can either match a single archive by
+            passing its exact creation time, or all archives created within a given time
+            interval.
 
-            - ``YYYY``: 1 year
-            - ``YYYY-MM``: 1 month
-            - ``YYYY-MM-DD``: 1 day
-            - ``YYYY-MM-DDTHH``: 1 hour
-            - ``YYYY-MM-DDTHH:MM``: 1 minute
-            - ``YYYY-MM-DDTHH:MM:SS``: 1 second
-            - ``@1735732800``: 1 second
-            - ``YYYY-MM-DDTHH:MM:SS.ffffff``: exact timestamp
-            - ``@1735732800.123456``: exact timestamp
+            To match a single archive by its exact creation time, use the forms:
+
+            - ``YYYY-MM-DDTHH:MM:SS.ffffff``: ISO-8601-like date-time string
+            - ``@1735732800.123456``: UNIX timestamp
+
+            To match a single archive, the pattern must specify the archive's complete
+            creation timestamp, including any fractional seconds. Fractional-second
+            patterns accept 1 to 6 digits.
+
+            To match all archives created within a given time interval, use the forms:
+
+            - ``YYYY``: match all archives created within the given year
+            - ``YYYY-MM``: within the given month
+            - ``YYYY-MM-DD``: on the given day
+            - ``YYYY-MM-DDTHH``: in the given hour
+            - ``YYYY-MM-DDTHH:MM``: in the given minute
+            - ``YYYY-MM-DDTHH:MM:SS``: in the given second
+            - ``@1735732800``: within the 1 second interval from the given UNIX timestamp
+
+            The ``T`` date-time separator may also be written as a space, e.g.
+            ``date:2025-01-01 14:30``.
 
             Date and time patterns match the interval implied by their precision, including
-            the start and excluding the end. Fractional-second patterns accept 1 to 6
-            digits and match exactly.
+            the start and excluding the end. For example, ``date:2026-06`` matches archives
+            created on or after ``2026-06-01T00:00:00`` and before ``2026-07-01T00:00:00``.
 
-            Date and time patterns may include a timezone suffix: ``Z``, ``+HH:MM``,
+            Date and time patterns may include a timezone suffix: ``Z`` (UTC), ``+HH:MM``,
             ``-HH:MM``, or ``[Region/City]``. Patterns without a timezone are interpreted
-            in the local timezone. Unix timestamp patterns are UTC and do not accept a
-            timezone suffix.
+            in the local timezone. Be wary of Daylight Saving Time (DST) transitions, as
+            they can make time intervals ambiguous or nonexistent. Use UTC to avoid such
+            issues. Unix timestamps are always UTC and do not accept a timezone suffix.
 
         Examples::
 
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -362,9 +362,11 @@ class HelpMixIn:
 
             Date and time patterns may include a timezone suffix: ``Z`` (UTC), ``+HH:MM``,
             ``-HH:MM``, or ``[Region/City]``. Patterns without a timezone are interpreted
-            in the local timezone. Be wary of Daylight Saving Time (DST) transitions, as
-            they can make time intervals ambiguous or nonexistent. Use UTC to avoid such
-            issues. Unix timestamps are always UTC and do not accept a timezone suffix.
+            in the local timezone. Unix timestamps are always UTC and do not accept a timezone suffix.
+
+            Be wary of Daylight Saving Time (DST) transitions, as they can make time intervals
+            ambiguous or nonexistent. For example, named zones such as ``[Europe/Berlin]`` track DST,
+            but the equivalent (winter) UTC offset of ``+01:00`` does not. Use UTC to avoid such issues.
 
         Examples::
 
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -314,6 +314,7 @@ class HelpMixIn:
         - user: exact match on the username who created the archive
         - host: exact match on the hostname where the archive was created
         - tags: match on the archive tags
+        - date: match on the archive creation timestamp
 
         In case of a name pattern match,
         it uses pattern styles similar to the ones described by ``borg help patterns``:
@@ -327,6 +328,28 @@ class HelpMixIn:
         `Regular expressions <https://docs.python.org/3/library/re.html>`_, selector ``re:``
             Full regular expression support.
             This is very powerful, but can also get rather complicated.
+
+        Date patterns, selector ``date:``
+            Match archives by creation timestamp. Supported forms are:
+
+            - ``YYYY``: 1 year
+            - ``YYYY-MM``: 1 month
+            - ``YYYY-MM-DD``: 1 day
+            - ``YYYY-MM-DDTHH``: 1 hour
+            - ``YYYY-MM-DDTHH:MM``: 1 minute
+            - ``YYYY-MM-DDTHH:MM:SS``: 1 second
+            - ``@1735732800``: 1 second
+            - ``YYYY-MM-DDTHH:MM:SS.ffffff``: exact timestamp
+            - ``@1735732800.123456``: exact timestamp
+
+            Date and time patterns match the interval implied by their precision, including
+            the start and excluding the end. Fractional-second patterns accept 1 to 6
+            digits and match exactly.
+
+            Date and time patterns may include a timezone suffix: ``Z``, ``+HH:MM``,
+            ``-HH:MM``, or ``[Region/City]``. Patterns without a timezone are interpreted
+            in the local timezone. Unix timestamp patterns are UTC and do not accept a
+            timezone suffix.
 
         Examples::
 
@@ -349,7 +372,12 @@ class HelpMixIn:
             borg delete -a 'host:kenny-pc'
 
             # tags match
-            borg delete -a 'tags:TAG1' -a 'tags:TAG2'\n\n"""
+            borg delete -a 'tags:TAG1' -a 'tags:TAG2'
+
+            # archive creation date match
+            borg delete -a 'date:2025-01'
+            borg delete -a 'date:2025-01-01T14:30Z'
+            borg delete -a 'date:2025-01-01T09:30[America/New_York]'\n\n"""
     )
     helptext["placeholders"] = textwrap.dedent(
         """
--- a/src/borg/helpers/time.py
+++ b/src/borg/helpers/time.py
@@ -255,7 +255,7 @@ DATE_PATTERN_RE = r"""
          (?:
              -(?P<day>\d{2})
              (?:
-                 T(?P<hour>\d{2})
+                 [ T](?P<hour>\d{2})
                  (?:
                      :(?P<minute>\d{2})
                      (?:
@@ -267,7 +267,7 @@ DATE_PATTERN_RE = r"""
      )?
   )
   (?P<tz>Z|[+\-]\d\d:\d\d|\[[^]]+\])?
-  $
+  \Z
 """
 
 
@@ -341,7 +341,6 @@ def compile_date_pattern(expr: str):
     Supported expressions are static calendar timestamps from year to fractional-second precision,
     optional timezone suffixes (Z, +/-HH:MM, or [Region/City]), and Unix epoch timestamps prefixed with @.
     """
-    expr = expr.strip()
     start, end = parse_date_pattern_interval(expr)
     if start == end:
         return date_match_exact(start)
--- a/src/borg/helpers/time.py
+++ b/src/borg/helpers/time.py
@@ -1,6 +1,7 @@
 import os
 import re
-from datetime import UTC, datetime, timedelta
+from datetime import UTC, datetime, timedelta, timezone
+from zoneinfo import ZoneInfo
 
 
 def parse_timestamp(timestamp, tzinfo=UTC):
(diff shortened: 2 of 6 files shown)


PhrozenByte:
Thanks for picking this up! :heart:

> 7. `YYYY-MM-DDTHH:MM:SS.ffff` -> matches an exact timestamp, including fractional seconds
> 8. `@123456789` -> Unix epoch (interpreted as UTC)
>
> (Is the fractional-second exact match useful in practice? Feedback welcome on this.)

What's the precision of an archive's creation time? From the code I assume it's with fractional seconds, right? I absolutely agree with you then: There should be a variant that is guaranteed to match a single archive. I feel like that Unix timestamps should also optionally support fractional seconds then.

> 4. `YYYY-MM-DDTHH` -> matches a 1-hour interval

From reading the code I derive that the 1-hour interval is matched *exclusively*, i.e. it's actually matching any archive within 00:59:59.9999… hours, correct? Perfect 👍

> ```python
> # Year/Year-month/Year-month-day
> parts = expr.split("-")
> try:
>     if len(parts) == 1:                    # YYYY
>         year = int(parts[0])
> ```

Even though I like the simplicity, I feel like that Borg should be pretty strict about the format, because being less strict easily leads to ambiguity. For example, is `date:0` supposed to match any archive created in year 0? Probably, but it gets way less clear with (now deprecated) truncated ISO8601 dates: What does the pattern `date:25-1` describe? January of the year 25, January 1925, or January 2025?

If there's no library that can be used, I always imagined that the code would basically revolve around a single, rather strict regex with optional groups for month, day, hour, minutes, seconds, and fractal seconds, or `*` as wildcard, supplemented by another regex to match periods, and simple matchers for Unix timestamps and keywords. I'm not saying that this is the best approach, that's just what I imagined while writing #8715.

In general I like to encourage creating extensive unit tests as early as possible. It's elegant and simple code now (🚀 👍), but complexity will increase greatly when adding more and more features.

Note: I can read the code, but can't do an actual code review - for that I just don't known enough of Borg's code.

src/borg/helpers/time.py
```diff
@@ -185,3 +185,100 @@ def isoformat(self):
...
+    start_utc = local(start).astimezone(timezone.utc)
+    end_utc   = local(end).astimezone(timezone.utc)
+    return lambda ts: start_utc <= ts < end_utc
+
+
+def compile_date_pattern(expr: str):
+    """
+    Turn a date: expression into a predicate ts->bool.
+    Supports:
+      1) Full ISO‑8601 timestamps with minute (and optional seconds/fraction)
```
ThomasWaldmann, line 216: So HH:MM:SS would be an interval of 1s?

src/borg/helpers/time.py
```diff
@@ -185,3 +185,100 @@ def isoformat(self):
...
+      2) Hour-only:   YYYY‑MM‑DDTHH      -> interval of 1 hour
+      3) Minute-only: YYYY‑MM‑DDTHH:MM   -> interval of 1 minute
+      4) YYYY, YYYY‑MM, YYYY‑MM‑DD       -> day/month/year intervals
```
ThomasWaldmann, line 219: If you want to list all patterns, please keep the list sorted by length of interval and use 1 pattern/interval per line.

src/borg/helpers/time.py
```diff
@@ -185,3 +185,100 @@ def isoformat(self):
...
+    minute_re = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}$")
+    if minute_re.match(expr):
+        start = parse_local_timestamp(expr + ":00", tzinfo=timezone.utc)
+        return interval_predicate(start, start + timedelta(minutes=1))
+
+    # 4) Hour-only
+    hour_re = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}$")
+    if hour_re.match(expr):
+        start = parse_local_timestamp(expr + ":00:00", tzinfo=timezone.utc)
+        return interval_predicate(start, start + timedelta(hours=1))
```
ThomasWaldmann, line 248: Maybe just use 1 regex with group names (`(?P<name>...)` that covers 1) .. 4) and also YYYY, YYYY-MM, YYYY-MM-DD cases from below.

After a single `m = re.match(regex, expr)`, you can check `m.groupdict()` in the right order (fraction, S, M, H, d, m, y) to determine which case you have.

ThomasWaldmann: Maybe use re.VERBOSE so you can have a multi-line, commented regex for this.

ThomasWaldmann (changes requested):
Thanks for the PR!

Some minor stuff I found...

ThomasWaldmann:
BTW, if you install the pre-commit hook, you can have your commits automatically formatted.

src/borg/helpers/time.py
```diff
@@ -191,96 +192,136 @@ class DatePatternError(ValueError):
...
+    pattern = r"""
+        ^
+        (?:
+            (?P<fraction>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+)   # full timestamp with fraction
+          | (?P<second>  \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})      # no fraction
+          | (?P<minute>  \d{4}-\d{2}-\d{2}T\d{2}:\d{2})            # minute precision
+          | (?P<hour>    \d{4}-\d{2}-\d{2}T\d{2})                 # hour precision
+          | (?P<day>     \d{4}-\d{2}-\d{2})                       # day precision
+          | (?P<month>   \d{4}-\d{2})                             # month precision
+          | (?P<year>    \d{4})                                   # year precision
```
ThomasWaldmann, line 260: Interesting approach.

What I meant was rather something like (simplified to cover only YYYY and YYYY-MM here as an example):
```
(?P<year>\d{4})
(-
 (?P<month>\d{2})
)?
```

c-herz: Ah yes this would've been much simpler. Will try to work on refactoring it to this approach tomorrow.

src/borg/helpers/time.py
```diff
@@ -185,3 +193,290 @@ def isoformat(self):
...
+
+DURATION_RE = re.compile(
+    r"^D"
+    r"(?:(?P<years>\d+)Y)?"
+    r"(?:(?P<months>\d+)M)?"
+    r"(?:(?P<weeks>\d+)W)?"
+    r"(?:(?P<days>\d+)D)?"
+    r"(?:(?P<hours>\d+)h)?"
+    r"(?:(?P<minutes>\d+)m)?"
+    r"(?:(?P<seconds>\d+)s)?"
```
ThomasWaldmann, line 312: didn't we use to have lowercase ymwd and uppercase HMS?

PhrozenByte: @c-herz, is that a custom format, or did I miss an ISO 8601 or RFC update? I know ISO-8601's `P3Y6M4DT12H30M5S` (`P` designator, 3 years, 6 months, 4 days, `T` time separator, 12 hours, 30 minutes, 5 seconds, all being optional) format. I honestly don't like that "official" format very much (it's so cumbersome…), especially in regards to using `M` for both months and minutes (it's still unambiguous due to the `T` separator), but I feel like we should support it, because it's an official part of ISO 8601.

However, I kinda like the idea of adding `W` for weeks. I also like the idea to simplify things and *additionally* support our own format. Like, why not support `12:34:56` (or similar, just a quick idea) to specify a 12 hours, 34 minutes, 56 seconds duration? Not sure whether we should put things into the "official" `P` designator, or add our own (like the suggested `D`). WDYT?

src/borg/helpers/time.py
```diff
@@ -185,3 +193,298 @@ def isoformat(self):
...
+
+# Regex for ISO-8601 timestamps:
+# Accepts both 'T' and space as separators between date and time per RFC-3339/IXDTF.
+MAIN_RE = r"""
+  ^
+  (?:
+     # ISO week date: YYYY-Www or YYYY-Www-D
+     (?P<isoweek_year>\d{4})-W(?P<isoweek_week>\d{2})(?:-(?P<isoweek_day>\d))?
+   | # Ordinal date: YYYY-DDD
+     (?P<ordinal_year>\d{4})-(?P<ordinal_day>\d{3})
```
PhrozenByte, line 279: This might be a bit silly, but it just got me thinking: Could we support ordinal days with wildcard years? :thinking: Same for ISO weeks, possibly even `*-W*-5` (wow, that looks crazy :laughing:) to match all archives created on a Friday? Not sure whether users would use it, but if it's possible? WDYT?

src/borg/helpers/time.py
```diff
@@ -185,3 +193,298 @@ def isoformat(self):
...
+   | # Unix epoch
+     @(?P<epoch>\d+)
```
PhrozenByte, line 281: Also supporting fractal seconds here would be amazing! :heart:

ThomasWaldmann:
before doing further changes, please rebase on current master branch to get the cython workaround. otherwise, builds will fail.

ThomasWaldmann:
Guess I'ld like to merge this for next beta, can we finish this until then?

ThomasWaldmann:
ping?

c-herz:
My apologies, I have been caught up with finals and starting a new internship. I should be able to look into @PhrozenByte's suggestions by this weekend. Apologies for the delays and the rather poor communication!

ThomasWaldmann:
ping? (no hurry, but if you find some time, it would be nice to finish this)

src/borg/helpers/time.py
ThomasWaldmann: fractional :-)

ThomasWaldmann:
Note: this PR can't even run the tests anymore, but it's not the fault of the code here, but current Cython requires some fixes not yet present here, but fixed in current master branch.

Please rebase on current master.

ThomasWaldmann:
Also, I would appreciate if we could get this into a mergable state ASAP.

We will soon have bigger changes and movements in the code, so there could be lots of merge conflicts coming.

ThomasWaldmann:
ping

PhrozenByte:
What do you think about reducing the scope of this PR? It might help move things forward and get the core functionality merged sooner.

A recent example is #8775, which was opened just one day before this PR and was successfully merged yesterday after its scope was reduced.

In particular, support for durations, from/to matching, and special keywords (e.g. "oldest") seems to be more involved than originally anticipated. Would it make sense to focus on date support first, perhaps with only very basic or no wildcard support, and leave the more complex parts for follow-up PRs?

c-herz:
Re: @PhrozenByte's suggestion--that definitely would make it a bit easier for me to get things up to speed after a year of neglect. 

My sincere apologies for the radio silence on my end. Over the past year life really got in the way for me and I just didn't have time to work on hobby projects. However, I certainly should've communicated that better. 

In any case, I should have some time this weekend and possibly here and there throughout the week to clean things up and at least get the basic feature I worked on last year rebased onto master and integrated with the major changes since then (which to be honest, I need to catch up on). I can shelve any nice-to-haves for now and focus on core functionality, and certainly keep everything as a reference if these are desired in the future and/or I have time to extend. Please let me know what should be the highest priority moving forward!

PhrozenByte:
There's really no need to apologize. We're all contributing in our spare time, and there are certainly more important things in life than getting a PR merged quickly 🙂

In my opinion, it should ultimately be up to you to decide which features you want to work on and where you want to draw the line for the initial implementation.

That said, since I originally suggested this in #8715, my personal motivation was simply being able to match archives by static dates. I create daily backups, and while the date used to be part of the archive name in Borg 1.x, archive series in Borg 2.0 work differently. Because of that, something as simple as a static match pattern like `date:2026-06-22` would already cover my personal use case. Everything else suggested in #8715 mostly came from considering how the feature could be extended in useful ways. For example, the idea of even more coarse-grained dates such as `date:2026-06` or `date:2026-W26` came from considering users that create monthly or weekly backups.

But again, you're the one doing the actual work here, so feel free to draw the line wherever it makes sense to you, both in terms of implementation effort and what you actually feel like working on.

ThomasWaldmann:
_Review done by Claude Opus 4.8_

## Overall

The implementation is clean and the integration into `manifest.py` mirrors the existing `host:`/`user:`/`tags:` dispatch correctly. Interval semantics (start-inclusive, end-exclusive) are right, overflow/invalid inputs are funneled into a clean `CommandError` rather than tracebacks, and test coverage is genuinely good (precision levels, tz equivalence, local-tz fallback, overflow, rejection cases). The earlier strictness concern is addressed — 4-digit years only, no truncated/ambiguous dates. The regex-with-named-groups design is exactly what was suggested in the earlier review rounds.

## Blocking / process

- **Won't build or test as-is.** The branch is ~584 commits behind master and predates the Cython workaround. It needs a rebase before CI can run. Reviewing against current master, the `manifest.py`/`time.py` integration points still line up cleanly, so the rebase should be mostly mechanical.

## Substantive findings

1. **`[Region/City]` timezones depend on `tzdata`, which isn't a declared dependency.** On Linux/macOS `zoneinfo` reads the system tz database, but on Windows there is none — `ZoneInfo("America/Los_Angeles")` raises `ZoneInfoNotFoundError`, which the code maps to `"invalid timezone format"`. So named-zone patterns silently don't work for Windows users, and `test_match_la_equivalents` / `test_match_utc_equivalents` would fail on Windows binary builds. Fix: add `tzdata` to dependencies (e.g. `; sys_platform == 'win32'`), or document the limitation.

2. **`$` in the regex permits a trailing newline.** `date:2025\n` parses as year 2025 (Python `$` matches before a final `\n`). Using `\Z` instead of `$` in `DATE_PATTERN_RE` hardens this. Low severity, but trivial to fix.

## Minor / polish

- **No changelog entry** in `docs/changes.rst` for this user-facing feature.
- Out-of-range components (`month=13`, `day=30` in Feb, `hour=24`) are accepted by the regex and only rejected later when `datetime()` raises `ValueError`. This is fine and caught cleanly — noted as by-design, not a hole.
- `compile_date_pattern` calls `expr.strip()` although the input is already a stripped match suffix — harmless redundancy.
- The help-text list mixes the `@epoch` entries between the calendar precisions, breaking the ascending-interval grouping.
- For named-zone (`ZoneInfo`) patterns at day/month/year precision, interval-end arithmetic via `timedelta` is wall-clock-naive across DST. Irrelevant for sub-day precisions and an extreme edge case for the coarse ones; not worth fixing now.

## Bottom line

Solid, mergeable-quality core once rebased. Suggested merge gates: (1) rebase onto current master, (2) the `tzdata`/Windows decision, (3) a changelog line. The `\Z` fix and help-text ordering are nice-to-haves.

ThomasWaldmann:
_Follow-up by Claude Opus 4.8_

## The CI prune failures are caused by this PR's new test leaking `TZ`

The three failing tests are all in `prune_cmd_test.py` (`test_prune_repository_example`, `test_prune_int_rolling_schedule_oldest_retention`, `test_prune_interval_rolling_schedule_oldest_retention`) and look unrelated to date matching — but they are a side effect of the new test:

```python
@pytest.mark.skipif(is_win32, reason="time.tzset() is not available on Windows")
def test_match_bare_pattern_uses_local_timezone(archivers, request, monkeypatch):
    monkeypatch.setenv("TZ", "America/New_York")
    time.tzset()   # pushes Eastern into libc
    ...
```

`monkeypatch.setenv` restores the `TZ` *environment variable* at teardown, but it never calls `time.tzset()` again, so libc keeps `America/New_York` active for every subsequent test in that worker process. This is the only place in `src/borg/testsuite/` that touches the process timezone, so nothing reinstates the runner's default zone afterwards.

Why it breaks prune:
- `match_archives_date_test.py` sorts before `prune_cmd_test.py`, so the leak is active when prune runs.
- `prune()` buckets archives by **local-time** periods. With the leaked Eastern zone, a UTC `2024-01-01T00:00` archive becomes local `2023-12-31T19:00` and shifts into the previous month/year bucket, which throws off the oldest-retention boundary math — hence `01-01`/`02-01` instead of `01-31`.
- The tell is in the failure output: every prune timestamp prints `-0500`/`-0400` (Eastern). On the runner's normal zone these tests pass.

So this is a test-isolation bug introduced here, not a prune regression.

## Suggested fix

`monkeypatch` alone can't fix this cleanly (its env-restore finalizer races with any `addfinalizer(time.tzset)` ordering). Restore the zone explicitly so `tzset()` runs *after* `TZ` is put back:

```python
def test_match_bare_pattern_uses_local_timezone(archivers, request):
    old_tz = os.environ.get("TZ")
    def restore_tz():
        if old_tz is None:
            os.environ.pop("TZ", None)
        else:
            os.environ["TZ"] = old_tz
        time.tzset()
    request.addfinalizer(restore_tz)

    os.environ["TZ"] = "America/New_York"
    time.tzset()
    ...
```

`os` and `time` are already imported in the test file. Factoring this into a small `set_timezone` fixture would be even nicer, since a tz-leak guard is generally reusable.

src/borg/archiver/help_cmd.py
```diff
@@ -328,6 +329,28 @@ class HelpMixIn:
             Full regular expression support.
             This is very powerful, but can also get rather complicated.
 
+        Date patterns, selector ``date:``
```
PhrozenByte, line 332: At first glance, I found this a little hard to follow. Here's a suggestion for the help text:

```
        Date patterns, selector ``date:``
            Match archives by creation timestamp. You can either match a single archive by
            passing its exact creation time, or all archives created within a given time
            interval.

            To match a single archive by its exact creation time, use the forms:

            - ``YYYY-MM-DDTHH:MM:SS.ffffff``: ISO-8601-like date-time string
            - ``@1735732800.123456``: UNIX timestamp

            To match a single archive, the pattern must specify the archive's complete
            creation timestamp, including any fractional seconds. Fractional-second
            patterns accept 1 to 6 digits.

            To match all archives created within a given time interval, use the forms:

            - ``YYYY``: match all archives created within the given year
            - ``YYYY-MM``: within the given month
            - ``YYYY-MM-DD``: on the given day
            - ``YYYY-MM-DDTHH``: in the given hour
            - ``YYYY-MM-DDTHH:MM``: in the given minute
            - ``YYYY-MM-DDTHH:MM:SS``: in the given second
            - ``@1735732800``: within the 1 second interval from the given UNIX timestamp

            Date and time patterns match the interval implied by their precision, including
            the start and excluding the end. For example, ``date:2026-06`` matches archives
            created on or after ``2026-06-01T00:00:00`` and before ``2026-07-01T00:00:00``.

            Date and time patterns may include a timezone suffix: ``Z`` (UTC), ``+HH:MM``,
            ``-HH:MM``, or ``[Region/City]``. Patterns without a timezone are interpreted
            in the local timezone. Be wary of Daylight Saving Time (DST) transitions, as
            they can make time intervals ambiguous or nonexistent. Use UTC to avoid such
            issues. Unix timestamps are always UTC and do not accept a timezone suffix.
```

It separates exact timestamp matches from interval matches, explains why exact matches include fractional seconds, is a bit more explicit about how interval matching works, and adds a note about timezones and DST.

PhrozenByte (commented):
From an user's perspective: LGTM, great work! :heart:

Just three rather minor suggestions:
1. WDYT about also allowing ` ` (space) as a date-time separator (e.g. `date:2025-01-01 14:30`)?
2. It might be a good idea to add a test for Daylight Saving Time (DST) specifically. For example, `date:2026-03-29T03[Europe/Berlin]` will behave a little weird and different from `date:2026-03-29T03+02:00`. IMHO there's no need to do anything about it, I just think it might be a good idea to make the behaviour reproducible with a unit test and note to users that timezone quirks exist (IMHO it's not necessary to go into detail, just that timezones can be weird and that one might want to force UTC instead).
3. I have a few suggestions concerning the help text. See comment below.

As always, I didn't do an actual code review, I'm reviewing this purely from an user's perspective by reading the code changes, unit tests, and docs.

I can do an actual test run once this has been rebased on latest `master`, but due to the pretty extensive test suite I don't think that I'll catch anything by that.

ThomasWaldmann:
Beta22 soon (~ 1week?), would be nice to get this in there.

ThomasWaldmann:
windows tests are failing, change log has a conflict.

master branch had quite a lot of changes recently, maybe do a rebase on fresh master.

c-herz:
Should be up to date now. I'll take a look into the windows tests failing.

ThomasWaldmann:
_Follow-up review by Claude Fable 5_

The updated branch addresses all earlier findings (rebase, `tzdata` on Windows, `\Z`, changelog, the `TZ` leak via the `set_timezone` fixture) and CI is green. I re-checked the new tests assertion-by-assertion — the boundary arithmetic, the timezone conversions (including that 2026-03-29 really is the EU spring-forward date and the DST test only uses wall times that exist uniquely), the fork-mode error handling, and the name assertions for substring collisions. All sound. Two remaining points, both about test cost rather than correctness:

## 1. Please drop `remote` from the test kinds

`match_archives_date_test.py` uses:

```python
pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,remote,binary")  # NOQA
```

The kind-picking policy in `src/borg/testsuite/archiver/__init__.py` (from the #9324 testsuite speedup) says `remote` should only be added where a backend genuinely branches per transport, and explicitly puts pure command/formatting logic in the `local,binary` bucket. `date:` matching is client-side manifest filtering — nothing transport-specific. With parametrize expansion this file is ~33 test instances, each creating a repo plus archives, so the remote leg spawns ~33 REST-server sessions for no extra coverage. Suggest:

```python
pytest_generate_tests = lambda metafunc: generate_archiver_tests(metafunc, kinds="local,binary")  # NOQA
```

## 2. Even better: test the parser directly

Nearly all of the boundary and rejection coverage exercises `compile_date_pattern` / `parse_date_pattern_interval`, which are pure functions. Those cases could move to plain unit tests (no repo, no archives, no subprocess) — e.g. asserting the `(start, end)` interval returned by `parse_date_pattern_interval` and `pytest.raises(DatePatternError)` for the invalid inputs. That would cut the runtime of this file to a fraction while keeping a handful of end-to-end tests for what genuinely needs the archiver: the `manifest.py` wiring, the bare-pattern local-timezone behavior, and the fork/binary error path.

Neither point blocks the feature — the implementation and the test logic itself look good.

PhrozenByte (commented):
LGTM :heart:

Thanks for adding a unit test for DST! That's definitely a useful test.

However, I actually had a different case in mind, and I just realized I described it wrong before :neutral_face: Instead of `date:2026-03-29T03[Europe/Berlin]`, I meant `date:2026-03-29T02[Europe/Berlin]`. This hours-interval should conceptually match all archives between `>= 02:00:00` and `< 03:00:00`, but that hour doesn't exist due to the DST transition. `start = datetime(2026, 3, 29, 2, 0, 0, tzinfo=ZoneInfo("Europe/Berlin"))` is the same instant as `end = start + timedelta(hours=1)`. However, the `start == end` comparison in `compile_date_pattern()` still evaluates to `False`, which causes Borg to use `date_match_interval()` to match archives. In `date_match_interval()`, both `datetime` instances are converted to UTC, where `start_utc == end_utc` now evaluates to `True` - which is why `start_utc <= ts.astimezone(UTC) < end_utc` can never match an archive. I think that's fine behavior-wise (this hour simply doesn't exist, so it's sensible that it can't match an archive), but I believe we should make this behavior explicit with a unit test. Sorry for describing it wrong first.

That said, this is really just a nit. If you'd rather merge this PR as-is, it LGTM nevertheless.

Just as a reminder, I'm not really doing a code review here (even though I'm obviously reading the code…), but rather reviewing it from a user's perspective.

ThomasWaldmann:
Merged this "as is" to get it into next beta. Further changes can be done in future PRs.

PR #8798, opened 2025-04-28
remove remainders of append-only support

ThomasWaldmann:
Some features like append-only repositories rely on a server-side component that enforces them (because that shall only be controllable server-side, not client-side).

So, that can only work, if such a server-side component exists, which is the case for borg 1.x ssh: repositories (but not for borg 1.x non-ssh: repositories).

For borg2, we currently have:
- fs repos
- sftp: repos
- rclone: repos (enabling many different cloud providers)
- s3/b3: repos
- ssh: repos using client/server rpc code similar as in borg 1.x

So, only for the last method we have a borg server-side process that could enforce some features, but not for any of the other repo types.

For append-only the current idea is that this should not be done within borg, but solved by a missing repo object delete permission enforced by the storage.

borg create could then use credentials that miss permission to delete, while borg compact would use credentials that include permission to delete.


Wqrld:
> For append-only the current idea is that this should not be done within borg, but solved by a missing repo object delete permission enforced by the storage.

Would this work? I assume borg would need file edit permissions even to create backups, in which case you could just write garbage to the files or encrypt them á la ransomware.

I think some form of append-only is very important to have, but i understand if the borg 1.0 system was too hacky to keep.

ThomasWaldmann:
@Wqrld you can test it in master branch. The "no-delete" permission disallows deleting objects as well as overwriting existing objects.

Wqrld:
Oh, amazing! Thanks for the great software!

ThomasWaldmann:
See also my comment there:

https://news.ycombinator.com/item?id=44216394

PR #8872, opened 2025-05-22
ChunkerParams: reject even window size for buzhash, fixes #8868

### LINKED ISSUE #8868: Buzhash xor-seed requires odd hash-window
The **buzhash chunker** allows the user to specify a custom `HASH_WINDOW_SIZE`, including even-sized windows.

The problem is that, for even window sizes, the seed is partially or fully canceled out, rendering it either less effective or completely ineffective. For a given seed `S`, the buzhash output is effectively XORed with:

$$
\bigoplus_{i=0}^{\text{Window}-1} \text{Rot}(S, i)
$$

For example, any 32-bit seed is completely canceled when the window size is 64 bytes, or reduced to one of two values — `0` or `0xffffffff` — for a 32-byte window, depending on the `popcount(S)`.

As far as I understand, this issue affects all even-sized windows. Circular rotation can be represented as multiplication of a bit vector by a permutation matrix `P`. Therefore, the XOR of rotated seeds becomes:

$$
S \cdot \left( \bigoplus_{i=0}^{\text{Window}-1} P^i \right) = S \cdot O
$$

If I did numpy math correctly, the resulting matrix `O` has non-full rank for every even window size, which implies that `O` is not invertible. As a result, multiplication by `O` is not a surjective function — some bits are inevitably lost.

I haven’t done formal algebra in over a decade, but at the very least, the case where `Window mod 64 = 0` is trivial and clearly demonstrates the issue. 🙂

To the best of my understanding, #5070 is also affected by this issue. Also, this isn’t related to the issue documented in [CDC-issues-reported-2025](https://github.com/borgbackup/borg/wiki/CDC-issues-reported-2025) — it's Buzhash-specific problem.

I think 1.x should, at most, emit a warning to the user about the potentially undesirable consequences of using an even HASH_WINDOW_SIZE with the buzhash chunker. It might be completely omitted in case of disabled encryption.

For 2.x, it might be worth considering a per-repository buzhash S-box instead of relying on a bare 32-bit seed. That said, I'm aware of #8841 and #2128.

ThomasWaldmann:


ThomasWaldmann:
1.4-maint backport: just emit a warning.

PR #8887, opened 2025-05-28
blake2b_256: fix key passing, fixes #8867

### LINKED ISSUE #8867: unusual blake2b usage
This is from master branch:
```
def blake2b_256(key, data):
    return hashlib.blake2b(key+data, digest_size=32).digest()
```

It is one of the chunkid hashes borg uses.

The usage there is unusual, the python docs rather suggest:
```
def blake2b_256(key, data):
    return hashlib.blake2b(data, key=key, digest_size=32).digest()
```

ThomasWaldmann:
borg 1.x involved the key a bit unusually in the blake2b hash calculation (padding 64B of key material with another 64B of zeros to a total of 128B and then just prepending it to the data).

since a while, we use blake2b from python stdlib and they support passing the key as a separate argument. up to 64B keys are allowed here.

besides the way how the 64B of key material are passed, there are also other differences in how the key gets involved in the digest computation, thus the legacy blake2b hashes from borg 1.x are not compatible with the ones borg 2 will use. The hash is not the same for same key and data values.

### DIFF
--- a/src/borg/archiver/benchmark_cmd.py
+++ b/src/borg/archiver/benchmark_cmd.py
@@ -248,7 +248,7 @@ class BenchmarkMixIn:
             else:
                 print(f"{spec:<24} {format_file_size(size):<10} {dt:.3f}s")
 
-        from ..crypto.low_level import AES256_CTR_BLAKE2b, AES256_CTR_HMAC_SHA256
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
Abandoning this.

We will:
- keep blake2b "as is", but only to read from borg 1.x repos
- not offer blake2b for borg2 repos
- add blake3 for borg2

PR #8899, opened 2025-06-03
fish: fix archive name completion

ThomasWaldmann:


ThomasWaldmann:
I made this with Junie AI help (considering that I don't use fish, it likely has more clue about it than I have).

It works, but I am wondering if this can be done in a simpler way.

The main points of this are:
- to show a list of "aid:XXXXXXXX" archive IDs together with archive (series) names and timestamp, so users can select from that list.
- NOT show filenames or paths from cwd as completion for positional arguments that want an archive name or ID
- remove all borg 1.x repo::archive stuff, this pattern is not used anymore in borg2. the repo is either given by --repo R or via the BORG_REPO env var.

PR #8904, opened 2025-06-06
Allow msgpack 1.1.1(rc1)

ThomasWaldmann:


ThomasWaldmann:
@methane While the CI tests here only pull released versions (not RC), I did a local test on MBP M3 using msgpack 1.1.1rc1 and it worked as expected.

PR #8920, opened 2025-06-13
buzhash64e: with rolling hash encryption

ThomasWaldmann:
Rather simple and very slow encryption code, needs to be tuned later IF we find this worthwhile doing.

Note:
- the encryption of a buzhash64 current sum value must only depend on the sum value and the encryption key, NOT on the position in the current stream (thus: no AES-CTR or any other cipher that does not independently encrypt blocks, this is why ECB is used here) - otherwise it would cut the chunks not only based on the content but also based on position in the stream and that would cut differently and would destroy deduplication.
- instead of encrypting just 64bit using ECB, we could collect e.g. 1024 sums, encrypt them all on in one `.encrypt_many()` call and then do another loop over the esums to decide where to cut. would be much less overhead than now, but needs maintaining yet another buffer efficiently.

### DIFF
--- a/scripts/chunker_comparison.py
+++ b/scripts/chunker_comparison.py
@@ -1,0 +1,512 @@
+"""
+Chunker Comparison Tool for Borg Backup
+
+This script analyzes and compares the statistical properties of different chunking algorithms
+used in Borg Backup (BuzHash and BuzHash64). It helps evaluate how data is split into chunks
+by each algorithm, which is crucial for deduplication efficiency.
+
+Usage:
+    python scripts/chunker_comparison.py [options]
+
+Options:
+    -g, --graphical       Enable graphical output (requires matplotlib)
+    -o, --output PATH     Output file prefix for saving plots (implies --graphical)
+    -d, --directory PATH  Path to directory containing files to analyze (instead of random data)
+    -s, --size SIZE       Size of random data in MB (default: 100MB, only used when not using --directory)
+
+Examples:
+    # Analyze with 100MB of random data
+    python scripts/chunker_comparison.py
+
+    # Analyze with 500MB of random data
+    python scripts/chunker_comparison.py --size 500
+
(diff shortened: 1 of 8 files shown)


darkk:
(As a sidenote on speed). I've found it quite interesting that fast [PCG RNG](https://www.pcg-random.org/) (numpy's default prng) is predictable, but [seed reconstruction out of 64 consecutive values](https://tosc.iacr.org/index.php/ToSC/article/view/8700) is a non-trivial operation — 2¼ CPU-years of computation. However, it's still just a €108, if we assume a CPU core to cost 4 €/month.

ThomasWaldmann:
rebased on current master, extended chunker comparison script to fastcdc, force pushed.

ThomasWaldmann:
Superseded by #9987, which realizes the same hash-then-encrypt idea with sound foundations and real performance:

- The universal-hash layer matters: buzhash64 over a 4095-byte window is not an ε-almost-universal family (equal bytes 64 positions apart cancel for every table, so colliding window pairs exist unconditionally), which is why encrypting its output cannot give the provable no-leakage guarantee. #9987's chunkers (rabin-aes, toeplitz-aes, goldilocks-aes) use 64-byte-window UHFs with proven collision bounds (down to exactly 2^-64 for toeplitz-aes) behind the same AES-128 idea.
- The "collect many sums, encrypt in one call" tuning note from this PR's description is implemented there: the C kernels batch 512 blocks through OpenSSL EVP on the portable path and interleave groups of 8 AES blocks with the rolling hash on arm64/AES-NI, reaching ~700 MB/s.
- The ECB/position-independence observation from this PR carries over unchanged and is documented in the new docs/internals/chunker.rst.

Closing this draft; it was the right idea and #9987 is its grown-up form.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8924, opened 2025-06-15
buzhash64: deterministically create a balanced bh table

ThomasWaldmann:
the previous approach had cryptographic strength randomness, but a precise 50:50 0/1 bit distribution per bit position in the table was not assured.

now this is always the case due to the way how the table is constructed.


ThomasWaldmann:
Hmm, is `random.Random` (Mersenne Twister) seeded with 256bit key derived from 256bit secret good enough for this or do we need a CSPRNG?

Considering that the `secrets` stdlib module does not offer seeding we would need to use AES-CTR or so for this.

**Update:** created own CSPRNG based on AES256-CTR. Not only because of it being cryptographically strong, but also because the seeding for `random.Random` sounded a bit like it could change depending on the Python version (although they would "offer" a compatible seeding, but guess that might require code changes on our side then!?). For our chunker, we need something long-term stable and really deterministic.

PR #8944, opened 2025-06-27
import-tar: normalize and sanitize tarinfo.linkname (master)

ThomasWaldmann:


src/borg/archive.py
```diff
@@ -1569,13 +1569,17 @@ def process_dev(self, *, tarinfo, status, type):
 
     def process_symlink(self, *, tarinfo, status, type):
         with self.create_helper(tarinfo, status, type) as (item, status):
-            item.target = tarinfo.linkname
+            normalized_path = os.path.normpath(tarinfo.linkname)
+            safe_path = make_path_safe(normalized_path)
+            item.target = safe_path
```
ThomasWaldmann: Hmm, after some more thoughts about this, I believe this would not be correct.

The symlink target (tarinfo.linkname) has to be seen like the contents of a user file and we must not change/edit that.

If a symlink points to `../../foo`, that must stay that way.

It's also not a safety issue, because extracting a symlink with such a target does no harm.

src/borg/archive.py
```diff
@@ -1569,13 +1569,17 @@ def process_dev(self, *, tarinfo, status, type):
...
             return status
 
     def process_hardlink(self, *, tarinfo, status, type):
         with self.create_helper(tarinfo, status, type) as (item, status):
             # create a not hardlinked borg item, reusing the chunks, see HardLinkManager.__doc__
-            chunks = self.hlm.retrieve(tarinfo.linkname)
+            normalized_path = os.path.normpath(tarinfo.linkname)
+            safe_path = make_path_safe(normalized_path)
+            chunks = self.hlm.retrieve(safe_path)
```
ThomasWaldmann, line 1580: Here we have a different situation: we want to fetch the correct entry from our hard link manager's key/value store.

See line 1600: it uses item.path as key (id) and item.path has been normalized/made safe by create_helper.

So, to access the kv store correctly, we must normalize/make safe before using `self.hlm.retrieve`.

src/borg/archive.py
```diff
@@ -1593,7 +1597,7 @@ def process_file(self, *, tarinfo, status, type, tar):
             item.get_size(memorize=True, from_chunks=True)
             self.stats.nfiles += 1
             # we need to remember ALL files, see HardLinkManager.__doc__
-            self.hlm.remember(id=tarinfo.name, info=item.chunks)
+            self.hlm.remember(id=item.path, info=item.chunks)
```
ThomasWaldmann, line 1598: This uses the alreaady normalized/made safe item.path as key (id) for the hlm kv store.

src/borg/archive.py
ThomasWaldmann: I reverted that change.

PR #8964, opened 2025-07-28
fixing reader bug, fixes #8963

### LINKED ISSUE #8963: file corruption with borg-2.0.0b18
My notebook ssd was too small and i used borg-2.0.0b18 to do a full restore to the new ssd.
Unfortunately i found one file in my home directory  which has changed. Size is matching but sha256 differs.

The new ssd is not much testet but all filesystems are ZFS and there is no checksum failure.
The notebook is running debian bookworm with openzfs and the backup was done to a local 2TB USB SSD.
I used the linux binary from github. 

Any idea how  to track this down? I restored 285GB to /home and it seems that only one file is broken.

ThomasWaldmann:


ThomasWaldmann:
Note: issue was introduced into master branch only after 2.0.0b17, no need for backports.

PR #8986, opened 2025-08-09
FUSE testing config (1.4-maint)

ThomasWaldmann:
Looks like the toml config format of tox does not yet support an easier way for generative test environments.


ThomasWaldmann:
Note: no backport to 1.2-maint needed, it still uses the tox.ini, not pyproject.toml.

Issue #4549, opened 2019-05-10
using borg with smartcards, crypto sticks, etc.


ThomasWaldmann:
borg has no special interface / does not use some special api for smartcards or crypto usb sticks.

But there is this:

## environment vars for the passphrase

https://borgbackup.readthedocs.io/en/stable/usage/general.html#environment-variables

- `BORG_PASSPHRASE` - directly give a passphrase to borg

- `BORG_PASSCOMMAND` - execute a shell command which outputs the passphrase

- `BORG_PASSPHRASE_FD` - read the passphrase from an open file descriptor

With these, some tool that might come with your hardware and some shell scripting, you can "glue together" the hardware and borg (no internal change in borg is needed for that).

You can use `repokey` as well as `keyfile` mode of borg for this, the passphrase that would unlock (decrypt) the key would from / via your special hardware, borg's encrypted key would be stored as usual.

## environment vars for the key directory

Alternatively, for usb sticks with encrypted mass storage, the `BORG_KEYS_DIR` (or `BORG_KEY_FILE`) environment variable could be used to let borg read the keyfile directly from the storage on the stick (keyfile mode).

The storage needs to be mounted at some place and `BORG_KEYS_DIR` (or `BORG_KEY_FILE`) needs to point into there (doing that is also scripting and outside the scope of what borg does internally).

ThomasWaldmann:
I personally do not use smartcards / crypto sticks (yet?), so I can not help much more than this.

So if anybody already has a setup for this, please add scripts and setup tips below.

jans23:
if you want a free Nitrokey device to develop and test such feature, please drop us an email (see https://www.nitrokey.com/contact ).

klausenbusk:
> So if anybody already has a setup for this, please add scripts and setup tips below.

Something like this should do it (assuming the key is on a smartcard):
```
head -c 32 /dev/urandom | base64 | gpg --encrypt -r <key on smartcard ID> > ~/.borg-passphrase.gpg
export BORG_PASSCOMMAND="gpg --decrypt $HOME/.borg-passphrase.gpg"
```

ThomasWaldmann:
It would be useful if a nitrokey (yubikey, ...) user could confirm that the solution @klausenbusk was suggesting is actually working.

If you test it, report back whether it is working and give the hardware / key setup you are using.

If it can be made working that simply, guess there is no need for hardware key specific development work (but we could add it to the docs / faq).

klausenbusk:
[...]

PR #8995, opened 2025-08-15
[RFC] Introduce support for FIDO2 to protect keys

steelman:
Use FIDO2 devices with [hmac-secret](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-hmac-secret-extension) extension to generate key encryption key (KEK) instead of passphrase processed with argon2 algorithm.

To use such device a user needs to point to one with `--fido2-device` option for either `repo-create` or `key change-passphrase`. For reading, borg tries to find which of plugged in devices can be used to decrypt KEK (see `fido2.Fido2Operations.find_device()`).
  
1. This is not a final version by any means. It is a working one, however. Please review it and report any improvement you think this patch may use.
2. I've introduced `fido2` module that wraps around [Yubico's module](https://github.com/Yubico/python-fido2/). I haven't updated the requirements because I believe this should remain optional. I've made the wrapper to be loadable without Yubico's module and defer reporting errors until an attempt to use it.
3. I've added help messages for command-line options I've introduced, but nothing more. Please tell, where to add a broader description of how FIDO2 device works with borg.
4. This code is somewhat inspired by how systemd uses FIDO2 (see [libfido2-util.c](https://github.com/systemd/systemd/blob/main/src/shared/libfido2-util.c)).
5. I've used borg for quite some time now, but in rather simplistic setups and I am not very familiar with the whole code base, so I probably missed some usage scenarios. 
6. I haven't really tested the `find_device()` function with more devices because I have just one.
7. I haven't added tests and I am not quite sure how to do it when a piece of hardware is involved and by definition not a reproducible one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
 [project.optional-dependencies]
 llfuse = ["llfuse >= 1.3.8"]
 pyfuse3 = ["pyfuse3 >= 3.1.1"]
+fido2 = ["fido2 >= 0.9.1"]
 nofuse = []
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ s3 = ["borgstore[rest,blake3,s3] ~= 0.6.1"]
 sftp = ["borgstore[rest,blake3,sftp] ~= 0.6.1"]
 rclone = ["borgstore[rest,blake3,rclone] ~= 0.6.1"]
 cockpit = ["textual>=6.8.0"]  # might also work with older versions, untested
+fido2 = ["fido2 >= 0.9.1"]
 
 [project.urls]
 "Homepage" = "https://borgbackup.org/"
--- a/scripts/shell_completions/bash/borg
+++ b/scripts/shell_completions/bash/borg
@@ -85,7 +85,7 @@ _borg()
                 local opts="-a --match-archives ${archive_filter_opts} ${common_opts}"
                 ;;
             *' repo-create '*)
-                local opts="--other-repo --from-borg1 -e --encryption --copy-crypt-key ${common_opts}"
+                local opts="--other-repo --from-borg1 -e --encryption --copy-crypt-key --fido2-device ${common_opts}"
                 ;;
             *' repo-list '*)
                 local opts="--short --format --json ${common_opts} -a --match-archives ${archive_filter_opts} --deleted"
@@ -136,8 +136,9 @@ _borg()
                 ;;
             # umount
             #   no specific options
-            # key change-passphrase
-            #   no specific options
+            *' change-passphrase '*)
+                local opts="${common_opts} --fido2-device"
+                ;;
             *' change-location '*)
                 local opts="${common_opts} keyfile repokey --keep"
                 ;;
--- a/scripts/shell_completions/zsh/_borg
+++ b/scripts/shell_completions/zsh/_borg
@@ -412,6 +412,7 @@ _borg-key() {
     case $line[1] in
     (change-passphrase)
       _arguments -s -w -S : \
+        '--fido2-device=[select a FIDO2 device, use "fido2-token -L" to list available devices]:devpath:_files -P /dev/ -W /dev' \
         $common_options
       ;;
     (export)
@@ -556,6 +557,7 @@ _borg-repo-create() {
 
   _arguments -s -w -S : \
     '(-e --encryption)'{-e,--encryption}'=[select encryption key mode (required)]:MODE:(none authenticated authenticated-blake2 keyfile-aes-ocb repokey-aes-ocb keyfile-chacha20-poly1305 repokey-chacha20-poly1305 keyfile-blake2-aes-ocb repokey-blake2-aes-ocb keyfile-blake2-chacha20-poly1305 repokey-blake2-chacha20-poly1305)' \
+    '--fido2-device=[select a FIDO2 device, use "fido2-token -L" to list available devices]:devpath:_files -P /dev/ -W /dev' \
     $common_repo_options \
     '--make-parent-dirs[create parent directories]'
 }
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -123,7 +123,7 @@ def with_repository(
                         f"You can use 'borg transfer' to copy archives from old to new repos."
                     )
                 if manifest or cache:
-                    manifest_ = Manifest.load(repository, compatibility, other=False)
+                    manifest_ = Manifest.load(repository, compatibility, args, other=False)
                     kwargs["manifest"] = manifest_
                     if "compression" in args:
                         manifest_.repo_objs.compressor = args.compression.compressor
@@ -569,6 +569,13 @@ def define_common_options(add_common_option):
         default=Location(other=False),
         action=Highlander,
         help="repository to use",
+    )
+    add_common_option(
+        "--fido2-device",
+        metavar="DEVICE",
+        dest="fido2_device",
+        default=os.environ.get("BORG_FIDO2_DEVICE", "none"),
+        help="select fido2 device to protect the repository key, use ``fido2-token -L`` to list available devices.",
     )
 
 
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -154,7 +154,7 @@ def with_repository(
                         from ..legacy.repoobj import RepoObj1
 
                         ro_cls = RepoObj1
-                    manifest_ = Manifest.load(repository, compatibility, other=False, ro_cls=ro_cls)
+                    manifest_ = Manifest.load(repository, compatibility, args, other=False, ro_cls=ro_cls)
                     kwargs["manifest"] = manifest_
                     if "compression" in args:
                         manifest_.repo_objs.compressor = args.compression.compressor
@@ -561,6 +561,13 @@ def define_common_options(add_common_option):
         default=Location(other=False),
         action=Highlander,
         help="repository to use",
+    )
+    add_common_option(
+        "--fido2-device",
+        metavar="DEVICE",
+        dest="fido2_device",
+        default=os.environ.get("BORG_FIDO2_DEVICE", "none"),
+        help="select fido2 device to protect the repository key, use ``fido2-token -L`` to list available devices.",
     )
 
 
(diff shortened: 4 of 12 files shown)


ThomasWaldmann:
Thanks for the PR!

Yes, guess this would be a good new feature, we even have a related ticket #4549.

I only had a quick glance at the code:
- maybe move `fido2` module into to the `borg.crypto` package
- i feel a bit uncomfortable about talking to fido2 hw without the user requesting it (like with `find_device()`). when talking to hw, things can go wrong (permissions, driver issues, malfunctioning hw, ...) and it would be bad if borg were just hanging or crashing in there even though the user maybe does not even use the fido2 hw with the current repository.
- we can think about whether a commandline option or an environment variable is better to specify the device that shall be used.
- yes, related package dependencies must be optional. borg can support installing them, see `pyproject.toml` -> `project.optional-dependencies`.
- ruff seems to be unhappy: https://github.com/borgbackup/borg/actions/runs/16999373707/job/48197734819?pr=8995
- black also seems to be unhappy: https://github.com/borgbackup/borg/actions/runs/16999373689/job/48197734800?pr=8995 - maybe install the pre-commit hook, see our developer docs.

steelman:
> Thanks for the PR!
> 
> Yes, guess this would be a good new feature, we even have a related ticket #4549.
> 
> I only had a quick glance at the code:
> 
>     * maybe move `fido2` module into to the `borg.crypto` package

Done. 

>     * i feel a bit uncomfortable about talking to fido2 hw without the user requesting it (like with `find_device()`). when talking to hw, things can go wrong (permissions, driver issues, malfunctioning hw, ...) and it would be bad if borg were just hanging or crashing in there even though the user maybe does not even use the fido2 hw with the current repository.

TL;DR — only device that created `credential_id` would accept it and return an assertion. Feeding `credential_id` and `salt` to a device different that the one used to create `credantial_id` is harmless.

I believe I can explain a little bit more how FIDO2 hmac-secret and it may ease your mind. Upon the creation of FIDO2 backed KEK [three values are created](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/fido2.py#L179) 

* `credential_id` — derived from values passed to the `make_credential()` method and device's secret key
* `salt` — a random value, as usual
* `secret` — the KEK

We [store the first two](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/crypto/key.py#L558) as an "encrypted key". They, together with the particular device, are needed to retrieve the secret in the future. Only the device used to create the `credential_id` we have, would accept it and return a secret (to get _the_ secret we need the right `salt` value). Considering that FIDO2 devices are meant to be indistinguishable from one another for privacy reasons (there is a specification for _enterprise_ tokens which are identifiable) the automatic process of matching `credential_id` from an encrypted key against all available devices during **decryption** is more convenient for users than asking them (in a rear event when they have more than one device attached) to identify the token manually.

Also — in the current setup (at least mine and this may need more investigation) a device asks for _user presence_ confirmation (_up_ in FIDO2 docs) which means pushing/touching a button on the device for it to return the `secret` value. Indeed, the requirement of _up_ is [disabled](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/fido2.py#L43) in the process of finding the matching device, but this is a _pre-flight_ phase as described in the FIDO2 specifications linked in the comments in `find_device()`.

>     * we can think about whether a commandline option or an environment variable is better to specify the device that shall be used.

OK, I don't have strong opinions about the UI (personally I like it to be as user controllable as possible).

However, do keep in mind that as I explained above it is crucial to manually specify the device only when calling [encrypt_key_file()](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/crypto/key.py#L522) (and thus [encrypt_key_file_fido2()](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/crypto/key.py#L554)). And even in that case the device would perform _user-presence_ check i.e. [wait for the user to touch the device](https://github.com/borgbackup/borg/blob/196d95c3e5fb40f51c644521be6b719323399eb7/src/borg/fido2.py#L119). Therefore, with most users having a single FIDO2 token using the first one returned by `CtapHidDevice.list_devices()` and waiting for user to confirm it is the right one by touching its button seems quite reasonable (see systemd's [fido2_find_device_auto()](https://github.com/systemd/systemd/blob/fb4aabf4432d523b97376099ce4353b5c268ae82/src/shared/libfido2-util.c#L1237) function)

>     * yes, related package dependencies must be optional. borg can support installing them, see `pyproject.toml` -> `project.optional-dependencies`.

Done.
 
>     * ruff […]
> 
>     * black […]

Done.

I also extended the changes to `passphrase.py` and moved them to a separate commit.

EDIT: One thing I haven't touched is key export/backup.

Unlike in case of passphrase based keys where only the encrypted key needs to be backed up and the passphrase is supposed to be remembered, the recovery scenario for hardware backed encryption needs to consider hardware failure. In [systemd-cryptenroll(1)](https://www.freedesktop.org/software/systemd/man/latest/systemd-cryptenroll.html) there is an option to enroll a _recovery key_. Technically it is just a very long random passphrase which in case of LUKS is enough because LUKS works like borg repokey that is an encrypted key is stored in volume's header. Unlike in borg (AFAIK), there may be up to 8 key slots in which a volume key may be encrypted with different passphrases (systemd-cryptenroll encodes the output of FIDO2 hmac-secret with base64 and passes it as a passphrase to cryptsetup).

In borg (AFAIK) there is no way to have a master repository key encrypted with several different KEKs except of manual (as of now) fiddling with keyfiles. In this case the easiest way to enable a user to recover from a hardware failure is to show them a secret value as returned by FIDO2 device and later accept it typed in should user's token fail (base64 or otherwise encoded for convenience). This method might also complement current paperkey export of passphrase based KEKs.

ThomasWaldmann:
Quite a lot of tests are failing. This is because fido2 related code gets used even if it was not explicitly requested (and there usually is no fido2 device while testing anyway).

While this could be fixed/adapted, it also proves my point about that things can go wrong and that fido2 functionality should not be used if not explicitly requested. My point is mainly about hw, driver and OS-level malfunctioning (which we can't fix nor foresee).

Just as an unrelated example: I use a yubikey for misc. purposes and sometimes I have to physically disconnect/reconnect it to resolve it being unresponsive in some application (while it worked flawlessly shortly before in another application). We must avoid that such malfunctions block borg from working.

There could be an env var BORG_FIDO2_DEVICE (or so), default value "none". The env variable could be queried at the relevant places and if it is "none", no fido2 code gets executed (behaves as before this PR). Other values could be a device spec or "auto" to automatically find a device.

steelman:
> Quite a lot of tests are failing. This is because fido2 related code gets used even if it was not explicitly requested (and there usually is no fido2 device while testing anyway).

I'll am back at work so I will fix it not as soon as I would like to, but I will.

> While this could be fixed/adapted, it also proves my point about that things can go wrong and that fido2 functionality should not be used if not explicitly requested. My point is mainly about hw, driver and OS-level malfunctioning (which we can't fix nor foresee).

IMHO requesting to work with a repository encrypted with fido2 backed key qualifies as an explicit request to use fido2 functionality. The only exception from such notion would be a request to manually enter a KEK in case of disaster recovery (broken/lost device).

> There could be an env var BORG_FIDO2_DEVICE (or so), default value "none". The env variable could be queried at the relevant places and if it is "none", no fido2 code gets executed (behaves as before this PR). Other values could be a device spec or "auto" to automatically find a device.

That sounds like a good idea. How about supporting also "auto" value triggering `find_device()`?

ThomasWaldmann:
Still on your radar @steelman?

steelman:
yes, but somewhat waiting for another holiday (-; still using borg

ThomasWaldmann (commented):
review done by Gemini 3.1 Pro (High):

Thanks for working on this! This is a great addition to Borg. I've reviewed the code and have a few suggestions and observations:

1. **Bug in string formatting**: In `src/borg/crypto/key.py`, inside the `create` method:
```python
passphrase = Passphrase.new(pin_prompt="Enter PIN for {args.fido2_device}: ")
```
This is missing an `f` prefix to be an f-string, so it will print literally `{args.fido2_device}`. It should be:
```python
passphrase = Passphrase.new(pin_prompt=f"Enter PIN for {args.fido2_device}: ")
```

2. **Passing `args` deeply through the call stack**: You are passing the raw `args` object from argparse through many layers (`Manifest.load`, `key_factory`, `Key.load`, `Key.save`, `decrypt_key_file`, etc.) just to access `args.fido2_device`. It's generally better practice to extract the specific values needed from `args` early on and pass only those specific values (e.g., `fido2_device=args.fido2_device`), or use an intermediate configuration object. This avoids tightly coupling lower-level crypto functions to the CLI `args` namespace.

3. **TODOs in the code**: There are several `# TODO` comments left in the code (e.g., `# TODO: verify that the device supports hmac-secret`, `# TODO: decide whether to use or not credentialProtectionPolicy`, `# TODO: replace with…`, and `# TODO: try to be more descriptive about the device`). It would be good to address these or turn them into tracked issues if they are meant for future PRs.

4. **Testing**: You mentioned you weren't sure how to test this since it involves hardware. Providing a mock implementation of `Fido2Operations` for unit tests that verifies the key loading/saving logic works correctly without a physical device would be very beneficial. For functional tests, you could look into using a software/virtual FIDO2 token (like `python-fido2`'s own test utilities do).

5. **Documentation**: To answer your question about where to add broader descriptions: `docs/usage/` would be a good place. You could add a new section in `docs/usage/key.rst` or a dedicated entry in `docs/faq.rst` explaining how to set up and use a FIDO2 device with Borg.

Overall, it's a solid working prototype. Addressing the above points will help make it ready for merging!

ThomasWaldmann:
About item 2 in the Gemini review: sometimes it is better NOT to add a cli option, but just use an env var:
- avoids having to pass through stuff through many layers
- does not clutter borg's help (especially true for common options: they appear in every borg command's help output, man page, online docs)

ThomasWaldmann:
@steelman Shall I rebase this onto current master?

steelman:
Done (-; Tests are failing though.

EDIT: The review not considered yet.

ThomasWaldmann:
@steelman See #10399.

ThomasWaldmann:
Closing in favour of #10399 (but considering that user presence is always required with this approach, maybe we won't take that one "as is" either).

PR #9005, opened 2025-09-07
diff: --sort-by=field[,field,...], fixes #8998

### LINKED ISSUE #8998: Add `borg diff --sort-by=field`
/kind enhancement

`borg diff` supports the `--sort` option since ages, but is limited to sorting by path right now. I'd like to suggest adding support for sorting by other fields as well. For example, `--sort=size_added` would sort paths by bytes added to the archive. Other reasonable fields to sort by could be `size_removed`, `size` (which could be `size_added` - `size_removed`), `mode` (octal), `user` and `uid`, `group` and `gid`, `ctime` and `mtime`, `path` (of course), and maybe more. Passing `--sort` without a value should default to sorting by path in ascending order (as before).

It should be possible to add a `>` or `<` prefix to field names that controls order: `>` means descending order, `<` ascending order. Default is ascending order.

Furthermore, it should be possible to pass the `--sort` option multiple times. So, `--sort=">size_added" --sort=">size_removed" --sort` would first sort by bytes added to the archive (descending order), then by bytes removed (descending order), and lastly by path (ascending order).

**Reasoning:**

My backup script always compares the deduplicated size of newly created archives against a configured treshold and warns me in case this treshold is exceeded. I added this check because I sometimes forget to remove temporary files from paths that are backed up by Borg. I usually remember what paths are at fault after receiving that warning, so I rectify the situation, restart the backup and remove the old archive, but sometimes I don't remember what paths are at fault, or e.g. some log file went rouge - and then it's quite hard to figure out what caused the archive to grow beyond the treshold. That's why I've created a wrapper for `borg diff` that allows me to sort by size.

<details>
<summary><code>borg-diff-sorted</code> shell function (click to expand; not really relevant for this issue, just in case someone is interested…)</summary>

```sh
borg-diff-sorted() {
    borg diff --content-only --json-lines "$@" \
        | jq -r '
            . as $obj |
            {
              path: .path,
              added: (
                [ .changes[] |
                  if .type == "modified" then .added
                  elif .type == "added" then .size
                  elif (.type | startswith("added ")) then "added"
                  else null
                  end
                ] |
                reduce .[] as $item (
                  null;
                  if type == "string" then .
                  elif $item | type == "string" then $item
                  else . + $item
                  end
                )
              ),
              removed: (
                [ .changes[] |
                  if .type == "modified" then .removed
                  elif .type == "removed" then .size
                  elif (.type | startswith("removed ")) then "removed"
                  else null
                  end
                ] |
                reduce .[] as $item (
                  null;
                  if type == "string" then .
                  elif $item | type == "string" then $item
                  else . + $item
                  end
                )
              )
            } |
            "\(.added // "")\t\(.removed // "")\t\(.path)"
        ' \
        | sort -t $'\t' -k1,1nr -k2,2nr -k3,3 \
        | awk -F'\t' '
            function human(x, p) {
              if (x+0 != x) return x
              if (x < 100000) return sprintf("%s%d B ", p, x)
              split("B kB MB GB TB PB", units, " ")
              for (i = 1; x >= 10000 && i < length(units); i++) x /= 1000
              return sprintf("%s%.2f %s", p, x, units[i])
            }
            { printf "%12s  %12s  %s\n", human($1, "+"), human($2, "-"), $3 }
        '
}
```

</details>

I'm aware of Borg 2 adding the `borg analyze` command, so for my particular use case I'd definitely switch to `analyze` instead - it simply is more powerful and better matches what I need. However, being able to sort by arbitrary fields might still be beneficial. It's a nice-to-have feature, but definitely nothing that should block Borg 2.0…

ThomasWaldmann:


src/borg/archiver.py
```diff
@@ -1172,15 +1172,69 @@ def print_text_output(diff, path):
...
+                'uid': -1,
+                'gid': -1,
+                'ctime': 0,
+                'mtime': 0,
+            }
+            if field in attr_defaults:
+                default = attr_defaults[field]
+                return getattr(it, field, default)
+            # default unknown fields: path
+            return remove_surrogates(path)
```
PhrozenByte: IMHO Borg should rather fail if one passes an unknown field (but really just an IMHO, no strong opinion about this).

src/borg/archiver.py
```diff
@@ -4096,6 +4150,48 @@ def define_borg_mount(parser):
...
+            Sorting
+            ++++++++
+            Use ``--sort FIELD`` to sort the output. Multiple ``--sort`` options apply stable sorts from
+            last to first. Prepend ">" for descending, "<" (or no prefix) for ascending, for example
+            ``--sort=">size_added" --sort=path``. Supported fields include:
+
+            - path: the item path
+            - size_added: total bytes added for the item content
+            - size_removed: total bytes removed for the item content
+            - size: shorthand for (size_added - size_removed)
```
PhrozenByte:
```suggestion
            - size_diff: result of (size_added - size_removed)
```

See below for reasoning about renaming the field to `size_diff`.

src/borg/archiver.py
```diff
@@ -4096,6 +4150,48 @@ def define_borg_mount(parser):
...
+            - mode, user, group, uid, gid, ctime, mtime: taken from the item state in ARCHIVE2 when present
```
PhrozenByte: I initially didn't really think about sorting by the literal `mtime` of a file, but rather by the `mtime` difference. My initial idea was that `--sort=">mtime"` would list files with the biggest time gap between modifications first. However, thinking about this again I'm not so sure any more whether this makes much sense… First, this obviously doesn't work with `user` and `group` (no idea why I listed them). For `uid` and `gid` one could indeed calculate a difference, but the significance is limited - to say the least. `mode` makes more sense (changing from 644 to 755 would yield a sorting value of 111). This approach originates from my `borg-diff-sorted` Shell function in #8998: I only had the information the JSON output provided. Given, sorting by `mtime`/`ctime`, `uid`/`gid`, and `mode` differences "feels" a bit weird.

Using the literal `mode`, `user`, `group`, `uid`, `gid`, `ctime`, and `mtime` of an item from ARCHIVE2 as sort value instead is more straight-forward, but I'm struggling with the use case a bit. I see an use case for sorting by `user`/`uid` and `group`/`gid`, because sorting by user/group is somewhat related to grouping by user/group - and I see benefit for that. `ctime`/`mtime` works very similar to my initial idea of calculating time differences, often the result might even be identical. But I can't think of an use case for `mode`. We must keep in mind that this is `borg diff`, not `borg list`, i.e. we only list items that were modified, so the listing is inherently not including everything.

I'm really struggling with what approach is more reasonable… WDYT?

If we go with the literal `mode` et. al., I'd rename the computed `size_added` - `size_removed` field to `size_diff` and use `size` for the literal uncompressed file size of the item instead. Are there any other reasonable literal fields? A `_diff` suffix would, in theory, also allow for implementing both, like using `mtime` for the literal value, and `mtime_diff` for the difference.

Two more notes:
1. I was thinking about whether there's any benefit in allowing users to choose whether the state is taken from ARCHIVE1, or ARCHIVE2 (like `old_mode` and `new_mode` in the JSON output; I'd keep `mode` as alias though)… Not sure yet whether this makes sense though.
2. Borg currently doesn't support sorting the output of `list`. Even though that's definitely out-of-scope here, we should at least keep it in mind to make it consistent. I'd say that's a strong argument for using literal `mode` et. al. instead of my original idea (or at least renaming `size` to `size_diff`).

src/borg/archiver.py
```diff
@@ -4119,8 +4215,8 @@ def define_borg_mount(parser):
                                help='only consider numeric user and group identifiers')
         subparser.add_argument('--same-chunker-params', dest='same_chunker_params', action='store_true',
                                help='Override check of chunker parameters.')
-        subparser.add_argument('--sort', dest='sort', action='store_true',
-                               help='Sort the output lines by file path.')
+        subparser.add_argument('--sort', dest='sort', action='append', nargs='?', const='path', metavar='FIELD',
+                               help='Sort the output lines by one or more fields. Example: --sort=">size_added" --sort=path. Use multiple --sort options for tie-breakers. Default (when omitted or without value) is path ascending.')
```
PhrozenByte: Question: What's the current behaviour without `--sort`?

Even though it likely impacts nothing, I'd rather keep it unsorted without `--sort`. However, like with unknown fields above, really just an IMHO, no strong opinion about this.

PhrozenByte (commented):
Thanks Thomas, great work, as always! :rocket: :heart: 

I didn't actually test this yet (but will do in the next couple of days), but I got some comments and open questions from the code (not an actual code review though, just from what I understand how it would work).

src/borg/archiver.py
ThomasWaldmann: Yes, you are correct. Guess we can enforce a set of valid options on the argparse level and just raise an exception here for the (then impossible) case.

src/borg/archiver.py
ThomasWaldmann: Without --sort I guess it just outputs the different items in the order as they are encountered in the archive.

src/borg/archiver.py
ThomasWaldmann: ## *_diff (archive2 - archive1 values)

sure, size difference is interesting.

guess sorting by mtime difference makes sense to find frequently changing files (including stuff that should not get backed up).

don't see much sense in uid/gid and mode differences. also, these might be rare anyway.

## * (archive2 values)

sorting by path, user, uid, group, gid, size, mtime/ctime for sure.

don't see much sense in sorting by mode.

I'ld say the newer archiver is usually more interesting for sorting. If not, the user could swap archive1 and archive2 arguments.

## borg list sorting

could be done, but might eat a lot of memory as it would need to buffer a lot of metadata (for all items, not only for different ones).

src/borg/archiver.py
ThomasWaldmann: ok

src/borg/archiver.py
PhrozenByte: Sounds reasonable :+1: If one actually needs more one will likely open an issue anyway, so no need to make things more complicated than this for now.

src/borg/archiver.py
ThomasWaldmann: done

src/borg/archiver.py
ThomasWaldmann: done

src/borg/archiver.py
ThomasWaldmann: done, please review.

src/borg/archiver.py
```diff
@@ -4119,8 +4254,8 @@ def define_borg_mount(parser):
                                help='only consider numeric user and group identifiers')
         subparser.add_argument('--same-chunker-params', dest='same_chunker_params', action='store_true',
                                help='Override check of chunker parameters.')
-        subparser.add_argument('--sort', dest='sort', action='store_true',
-                               help='Sort the output lines by file path.')
+        subparser.add_argument('--sort', dest='sort', action='append', nargs='?', const='path', metavar='FIELD', type=diff_sort_spec_validator,
```
PhrozenByte: `nargs='?'` apparently doesn't work with options and arguments: Python's `argparse` always treats the next item as option value, i.e. `borg diff --sort /path/to/repo::archive1 archive2` causes `"/path/to/repo::archive1"` to be interpreted as value for `--sort` (yielding a `borg diff: error: argument --sort: unsupported sort field: /path/to/repo::archive1` error)…

That's a problem considering the necessary backwards compatibility… :see_no_evil: With Borg 2 we could simply switch to `nargs=1` and require an option value, but with Borg 1.4 we need `--sort` to work without an option value, too. One can circumvent this issue by passing `--sort` last (like you did in the test suite), but it would be a breaking change nevertheless.

I can think of three alternatives:
1. Defer this feature until Borg 2 and use `nargs=1` then.
2. Deprecate `--sort` with Borg 1.4 and add a new `--sort-by` (option name matches `borg1 list` resp. `borg2 repo-list`, so it's a reasonable choice) with the new functionality and `nargs=1` now instead. We could then either remove `--sort` with Borg 2 and keep `--sort-by` unchanged, or rename `--sort-by` to `--sort` with Borg 2.
3. Do some magic :magic_wand: I'm not entirely sure whether this is even possible, but in theory `argparse` could decide whether the next item is the option value or an argument by comparing it against a list of valid option values (maybe `choices` does the trick? idk…). However, as I said, I don't know whether this is even possible with `argparse` and even if it is, I'm no big fan of such magic…

WDYT?

docs/usage/diff.rst
```diff
@@ -41,4 +41,12 @@ Examples
     {"path": "file1", "changes": [{"type": "modified", "added": 17, "removed": 5}, {"type": "mode", "old_mode": "-rw-r--r--", "new_mode": "-rwxr-xr-x"}]}
     {"path": "file2", "changes": [{"type": "modified", "added": 135, "removed": 252}]}
     {"path": "file4", "changes": [{"type": "added", "size": 0}]}
-    {"path": "file3", "changes": [{"type": "removed", "size": 0}]}
\ No newline at end of file
+    {"path": "file3", "changes": [{"type": "removed", "size": 0}]}
+
+    # Multiple --sort options apply stable sorts from last to first.
+    # Here: primary by net size change descending, tie-breaker by path ascending
+    $ borg diff --sort=">size" --sort="<path" testrepo::archive1 archive3
```
PhrozenByte:
```suggestion
    $ borg diff --sort=">size_diff" --sort="<path" testrepo::archive1 archive3
```

src/borg/archiver.py
```diff
@@ -1172,15 +1172,83 @@ def print_text_output(diff, path):
...
+                'mtime': 0,
+            }
+            if field in attr_defaults:
+                default = attr_defaults[field]
+                return getattr(it, field, default)
+            raise ValueError('Invalid field name: %s' % field)
+
+        # Apply stable sorts from last to first spec
+        sort_specs = args.sort or ['path']
+        for spec in reversed(sort_specs):
```
PhrozenByte:
```suggestion
        for spec in reversed(args.sort or []):
```

I did some testing with my production repo (see main review comment; it's a 34.87 TB / 29.98 TB compressed / 703.12 GB deduplicated repo with recently usually around 320 GB / 280 GB compressed / 500-1000 MB deduplicated archives, the diff in question has seen 332,602 changes) and sorting by default has a noticeable performance impact (from 2m26s to 3m8s; the exact difference changes from run to run, but it's always noticeably slower). I'd thus vote for not sorting by default. Still not the strongest opinion on the matter, but stronger than before and I'd rather leave it unsorted (like before).

src/borg/archiver.py
```diff
@@ -4089,13 +4157,80 @@ def define_borg_mount(parser):
...
+
+            Output formats
+            ++++++++++++++
+            The default (text) output shows one line per changed path, e.g.::
+
+                +135 B    -252 B [ -rw-r--r-- -> -rwxr-xr-x ] path/to/file
+
+            JSON Lines output (``--json-lines``) prints one JSON object per changed path. The schema is::
+
+                {"path": "PATH", "changes": [
```
PhrozenByte, line 4220: I like the idea of documenting the format. WDYT about a complete list of all possible types?

So, for example, there are more `type`s for other attribute changes, like `ctime` and `mtime`, but also user and group changes (not sure how their `type` is called). Also, AFAIK `"type": "removed"` is only for regular files, other item types (like directories and symlinks) use something like `"type": "removed <type>"`.

Unfortunately I don't have a full list either, just the things I learned from writing my `borg-diff-sorted` Bash function.

PhrozenByte (commented):
Thanks Thomas! :+1:

I just did another review (as usual no code review though) and installed your branch locally to test with my production repo (with a Podman overlay volume, i.e. the containerized Borg just had r/o access to the repo, but could still write to an overlayfs layer, allowing me to use production data but still being safe from possible corruption; just fyi, might be useful).

Sorting works like a charm in general. However, I found one issue and have some more notes. See below.

src/borg/archiver.py
ThomasWaldmann: Sure, default should be unsorted.

src/borg/archiver.py
ThomasWaldmann: Guess I'll rather remove the schema description from this PR, it is out of scope.

ThomasWaldmann: I kept the json, but as an example. it is not intended to be a schema specification, at least not within the scope of this PR.

src/borg/archiver.py
ThomasWaldmann: option 2 sounds good, i'll work on that...

ThomasWaldmann: fixed

ThomasWaldmann:
OK, guess I am finished here? @PhrozenByte can you give it another test?

TODO: collapse all commits into one, except the preprocess_args fix.

PhrozenByte:
Works like a charm! :rocket:

I tested with the latest commit (with `--sort-by ">size_diff,>mtime_diff,>ctime_diff,path"`) and went through the change set another time (still no code review though) and didn't find anything, so LGTM :+1:

Some notes about performance: Previously `borg diff` didn't buffer its output when no sorting was enabled, now it does buffer the output (i.e. it now prints everything in bulk right before the command finishes). Execution time is more or less the same (increase from 2:14.09 to 2:18.43, which is within normal fluctuation), but peak memory usage increased by 42% (max resident set size according to GNU time went up from 1,388,180 kB to 1,969,272 kB). Enabling sorting increased execution time as expected, but peak memory usage stayed at the mentioned 1.9 GB. So I presume there's still some buffering happening even without sorting (I suspect moving `diff.changes()`). Might be worth a look, however, personally I don't consider it a critical issue. If you want to address it later let me know, I'll open a separate issue then.

src/borg/archiver.py
```diff
@@ -1172,15 +1172,93 @@ def print_text_output(diff, path):
...
-        diffs = ((path, diff.changes()) for path, diff in diffs if not diff.equal)
-
-        if args.sort:
-            diffs = sorted(diffs)
-
-        for path, diff in diffs:
-            print_output(diff, remove_surrogates(path))
+        diffs_iter = Archive.compare_archives_iter(self.print_warning, archive1, archive2, matcher, can_compare_chunk_ids=can_compare_chunk_ids, content_only=args.content_only)
+        # Keep ItemDiff object for sorting on various fields
+        diffs_list = [(path, diff) for path, diff in diffs_iter if not diff.equal]
```
ThomasWaldmann: this builds a big list in memory, even if we do not need it, because we don't sort.

ThomasWaldmann: fixed

PhrozenByte:
Tested again with latest, still works like a charm and peak memory usage without sorting is down to 1,395,016 kB again (fluctuations by a few MB are normal). Bulk output is gone, too.

Note: At first I saw a peak memory usage of about 1.59 GB, but after some debugging I found that this was due to my test setup. I run native Arch Linux with Python 3.13.7, but tested within a Ubuntu 24.04 container running Python 3.10.12. After switching to an Arch Linux container for testing I saw the same peak memory usages around 1.39 GB again. So, my previous comparisons were wrong: The peak memory usage wasn't up by +41%, but "just" by +23.5%. Sorry about that unintentional exaggeration, I will do proper comparisons in the future.

So, LGTM! Great work, thank you Thomas! :+1:

PR #9006, opened 2025-09-07
feat(backend): enable s3 support using borgstore

mikemrm:
Borg2 documentation mentions the support for the s3 backend however, borg was missing the parsing bits for an s3 repo.

This updates the Location parser to parse the s3 url using the same logic as borgstore.

Note: borgstore should be installed with the s3 dependencies in order for the s3 backend to work.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
 llfuse = ["llfuse >= 1.3.8"]
 pyfuse3 = ["pyfuse3 >= 3.1.1"]
 nofuse = []
+s3 = ["borgstore[s3] ~= 0.3.0"]
 
 [project.urls]
 "Homepage" = "https://borgbackup.org/"
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -472,6 +472,27 @@ class Location:
         re.VERBOSE,
     )
 
+    # (s3|b2):[profile|(access_key_id:access_key_secret)@][schema://hostname[:port]]/bucket/path
+    s3_re = re.compile(
+        r"""
+        (?P<s3type>(s3|b2)):
+        ((
+            (?P<profile>[^@:]+)  # profile (no colons allowed)
+            |
+            (?P<access_key_id>[^:@]+):(?P<access_key_secret>[^@]+)  # access key and secret
+        )@)?  # optional authentication
+        (
+            [^:/]+://  # scheme
+            (?P<hostname>[^:/]+)
+            (:(?P<port>\d+))?
+        )?  # optional endpoint
+        /
+        (?P<bucket>[^/]+)/  # bucket name
+        (?P<path>.+)  # path
+    """,
+        re.VERBOSE,
+    )
+
     rclone_re = re.compile(r"(?P<proto>rclone):(?P<path>(.*))", re.VERBOSE)
 
     file_or_socket_re = re.compile(r"(?P<proto>(file|socket))://" + abs_path_re, re.VERBOSE)
@@ -483,6 +504,7 @@ class Location:
         self.valid = False
         self.proto = None
         self.user = None
+        self._pass = None
         self._host = None
         self.port = None
         self.path = None
@@ -524,6 +546,15 @@ class Location:
             self.proto = m.group("proto")
             self.path = os.path.normpath(m.group("path"))
             return True
+        m = self.s3_re.match(text)
+        if m:
+            self.proto = m.group("s3type")
+            self.user = m.group("profile") if m.group("profile") else m.group("access_key_id")
+            self._pass = True if m.group("access_key_secret") else False
+            self._host = m.group("hostname")
+            self.port = m.group("port") and int(m.group("port")) or None
+            self.path = m.group("bucket") + "/" + m.group("path")
+            return True
         m = self.local_re.match(text)
         if m:
             self.proto = "file"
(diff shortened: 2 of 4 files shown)


src/borg/helpers/parseformat.py
```diff
@@ -564,7 +596,7 @@ def host(self):
     def canonical_path(self):
         if self.proto in ("file", "socket"):
             return self.path
-        if self.proto == "rclone":
+        if self.proto in ("rclone", "s3"):
             return f"{self.proto}:{self.path}"
```
ThomasWaldmann, line 600: the "ssh" case includes much more stuff into the returned value. guess s3 is more similar to that.

src/borg/helpers/parseformat.py
```diff
@@ -472,6 +472,27 @@ class Location:
         re.VERBOSE,
     )
 
+    # (s3|b2):[profile|(access_key_id:access_key_secret)@][schema://hostname[:port]]/bucket/path
+    s3_re = re.compile(
+        r"""
+        (?P<s3type>(s3|b2)):
```
ThomasWaldmann, line 478: hmm, where is s3type used?

src/borg/helpers/parseformat.py
mikemrm: Sounds good, updated!

src/borg/helpers/parseformat.py
mikemrm: s3type isn't used here, this was [copied from borgstore](https://github.com/borgbackup/borgstore/blob/master/src/borgstore/backends/s3.py#L29-L46).

I went ahead and removed the unused groups.

src/borg/testsuite/helpers/parseformat_test.py
```diff
@@ -86,75 +86,94 @@ def test_ssh(self, monkeypatch, keys_dir):
...
+            repr(Location("s3:profile@http://172.28.52.116:9000/test/path"))
+            == "Location(proto='s3', user='profile', pass=None, host='172.28.52.116', port=9000, path='test/path')"  # noqa: E501
+        )
+        assert (
+            repr(Location("s3:user:pass@http://172.28.52.116:9000/test/path"))
+            == "Location(proto='s3', user='user', pass='REDACTED', host='172.28.52.116', port=9000, path='test/path')"  # noqa: E501
+        )
+        assert (
+            repr(Location("b2:user:pass@https://s3.us-east-005.backblazeb2.com/test/path"))
+            == "Location(proto='s3', user='user', pass='REDACTED', host='s3.us-east-005.backblazeb2.com', port=None, path='test/path')"  # noqa: E501
```
ThomasWaldmann: doesn't this lose the information that the user wanted b2 not s3? what does it give to borgstore?

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
 llfuse = ["llfuse >= 1.3.8"]
 pyfuse3 = ["pyfuse3 >= 3.1.1"]
 nofuse = []
+s3 = ["borgstore[s3]"]
```
ThomasWaldmann: did you try if that works? borgstore is already specified in line 34.

also, the version requirement is missing here.

src/borg/helpers/parseformat.py
```diff
@@ -524,6 +546,15 @@ def _parse(self, text):
             self.proto = m.group("proto")
             self.path = os.path.normpath(m.group("path"))
             return True
+        m = self.s3_re.match(text)
+        if m:
+            self.proto = "s3"
+            self.user = m.group("profile") if m.group("profile") else m.group("access_key_id")
+            self._pass = True if m.group("access_key_secret") else False
```
ThomasWaldmann, line 553: How does it give the secret to borgstore?

ThomasWaldmann (commented):
some stuff i noticed.

src/borg/testsuite/helpers/parseformat_test.py
mikemrm: b2 just adds some connection/session changes to work better with b2, but it still uses the s3 backend store. But I can understand the user might not expect the s3 output, so I've updated it to report the correct s3 type.

pyproject.toml
mikemrm: It does work as it's installing optional its optional dependencies. I had assumed the main dependencies already had the version restraint, but I will define the version here as well.

src/borg/helpers/parseformat.py
mikemrm: borgstore gets the original, repo url (with some processing). The parsing for Location is just so that we can detect that it's for s3 so we can pass the url off to borgstore.

References `Location.processed`: https://github.com/borgbackup/borg/blob/332ef28c7e3d73bf884206c98e6ca11b497acfa9/src/borg/repository.py#L110
`Location.processed` is the raw url with placeholders replaced: https://github.com/borgbackup/borg/blob/332ef28c7e3d73bf884206c98e6ca11b497acfa9/src/borg/helpers/parseformat.py#L501

src/borg/helpers/parseformat.py
mikemrm: Added back to be used in the proto field.

mikemrm:
Not sure why codecov is failing. It seems the previous codecov is 4 commits behind on master so codecov looks like a lot more has changed, but it hasn't. Thoughts?

ThomasWaldmann:
Guess codecov wants a bit more coverage. Check if it makes sense, if not, ignore it.

BTW, I did some recent CI testing work in borgstore project and added local sftp and s3 servers on the CI host that can be used for testing. Got it working there and now copied that to borg project, so we can also do s3 and sftp testing here: #9011 

Maybe you can rebase this PR onto master after #9011 is merged and add some s3-based tests for borg?

ThomasWaldmann:
@mikemrm #9013

mikemrm:
Thanks @ThomasWaldmann!

PR #9018, opened 2025-09-19
Add Arch Linux to installing from source docs

PhrozenByte:
Note that Arch Linux usually has no separate `-dev` packages, but ships development headers with the normal package.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -191,6 +191,19 @@ instructions, try to use your package manager to install the dependencies.
 After you have installed the dependencies, you can proceed with steps outlined
 under :ref:`pip-installation`.
 
+Arch Linux
+++++++++++
+
+Install the runtime and build dependencies::
+
+    pacman -S python python-pip python-virtualenv openssl acl xxhash lz4 zstd base-devel
+    pacman -S fuse2     # needed for llfuse
+    pacman -S fuse3     # needed for pyfuse3
+
+Note that Arch Linux specifically doesn't support
+`partial upgrades <https://wiki.archlinux.org/title/Partial_upgrade>`__,
+so in case some packages cannot be retrieved from the repo, run with ``pacman -Syu``.
+
 Debian / Ubuntu
 +++++++++++++++
 


docs/installation.rst
```diff
@@ -191,6 +191,17 @@ instructions, try to use your package manager to install the dependencies.
 After you have installed the dependencies, you can proceed with steps outlined
 under :ref:`pip-installation`.
 
+Arch Linux
+++++++++++
+
+Install the runtime and build dependencies::
+
+    pacman -S python python-pip python-virtualenv openssl acl xxhash lz4 zstd base-devel
```
ThomasWaldmann, line 199: the fuse stuff is installed by default?

ThomasWaldmann: also: pkg-conf(ig) and libffi?

PhrozenByte: I had already added FUSE a few hours back, but hadn't pushed it yet because testing it inside an unprivileged Podman container was a bit difficult and took way longer than anticipated. Sorry for the delay, got it running now.

`libffi` is already pulled in by the `python` package. Shall I add it explicitly?

Same for `pkg-config`: It's part of the `pkgconf` package, which is already pulled in by `base-devel` (like the other stuff that is actually required, but not listed explicitly).

I didn't include `git` because it's also missing for e.g. Debian / Ubuntu. But I guess that's not really a strict requirement due to the `pip` install method.

Partial Off Topic: The `test_readonly_{check,diff,export_tar,extract,info,list}` tests fail no matter what I do (I'm testing with the `1.4-maint` branch; `master` branch tests successfully). Is this expected or is something wrong?

Another partial Off Topic: I was thinking about adding a section for building Borg from source using a Podman container (likely also works with Docker, but I don't use Docker). Just to be clear, I don't mean creating a Borg container for others to use as-is by pulling an "official" Borg image from e.g. Docker Hub (i.e. *not* #4364 / #4372). I just mean adding instructions to the install docs about how to create a container in which one can install Borg from source and run it. This would then work with basically any Linux distribution, not just those currently listed in the docs. WDYT?

ThomasWaldmann: I don't want to add container-specific instructions:
- the borg part should be the same
- the container specific part depends on the container sw

About the test failures: please file a bug.

ThomasWaldmann (approved):
LGTM! Thanks for the PR.

ThomasWaldmann:
Can you backport this to 1.4-maint please?

PhrozenByte:
Thanks, sure, see #9019 :+1:

PR #9031, opened 2025-09-30
Add tests for diff output of archives with hard links.

cr1901:
This is a follow-up to #8344. After mulling it over, I generally think `borg` is doing the right thing when diffing with hard links, as the changes reported [match POSIX semantics](https://pubs.opengroup.org/onlinepubs/9799919799/functions/unlink.html) for unlinking:

```
Upon successful completion, unlink() shall mark for update the last data modification and last file status change timestamps of the parent directory. Also, if the file's link count is not 0, the last file status change timestamp of the file shall be marked for update.
```

An `mtime` and `ctime` change with no content modification or permission change between the same file in two repos is probably a reasonable hint that a hard link was broken. Similarly, a `ctime` change with no permission change is probably a reasonable hint that the hard link count changed. _This is the same info `borg` v1 currently reports for a hard link that was broken/hard link count that changed._ Additionally, excluding one hard link but not another from a diff seems to work just fine as well, completely sidestepping the need for a fix like #8344 :) in v1.

It may be interesting to extend the `diff` output to be hard-link aware, but not clear to me what this should look like, short of the "primary/secondary" mechanism that required #8344 in the first place :). For now users interested in actual hard link info can supply a `--format` like `'{hlid} {path}'` to `borg-2 list` and `sort`:

```
william@xubuntu-dtrain:/tmp/borg-hl-test$ borg-2 list --format '{path} {hlid}{NL}' -r ./repo-2 test
.
a 72b133632359d572aeb98bebb9654a73912e922f5d714f24098dd0cdc11a4909
b 72b133632359d572aeb98bebb9654a73912e922f5d714f24098dd0cdc11a4909
```

As long as `borg` keeps track of the info, there shouldn't be much need for end users to worry about hard link diffs (and the `mtime`/`ctime` hints above are good markers).

### DIFF
--- a/src/borg/testsuite/archiver/diff_cmd_test.py
+++ b/src/borg/testsuite/archiver/diff_cmd_test.py
@@ -325,6 +325,9 @@ def test_sort_option(archivers, request):
 def test_hard_link_deletion_and_replacement(archivers, request):
     archiver = request.getfixturevalue(archivers)
 
+    # repo-create changes umask, so create the repo first to avoid any
+    # unexpected permission changes.
+    cmd(archiver, "repo-create", RK_ENCRYPTION)
     path_a = os.path.join(archiver.input_path, "a")
     path_b = os.path.join(archiver.input_path, "b")
     os.mkdir(path_a)
@@ -334,9 +337,8 @@ def test_hard_link_deletion_and_replacement(archivers, request):
     create_regular_file(archiver.input_path, hl_a, contents=b"123456")
     os.link(hl_a, hl_b)
 
-    cmd(archiver, "repo-create", RK_ENCRYPTION)
     cmd(archiver, "create", "test0", "input")
-    os.unlink(hl_a)  # Don't duplicate warning message- one is enough.
+    os.unlink(hl_a)  # Don't duplicate warning message - one is enough.
     cmd(archiver, "create", "test1", "input")
 
     # Moral equivalent of test_multiple_link_exclusion in borg v1.x... see #8344
@@ -384,11 +386,14 @@ def test_hard_link_deletion_and_replacement(archivers, request):
     # modifications should be a hint that something hard-link related is going on.
     assert_line_exists(lines, "[cm]time:.*[cm]time:.*input/a/hardlink")
     assert_line_not_exists(lines, "modified.*B.*input/a/hardlink")
-    assert_line_not_exists(lines, "[.* -> .*].*input/dir_replaced_with_file")
-    # The hard-link count went down, but file content isn't modified.
+    assert_line_not_exists(lines, "-[r-][w-][x-].*input/a/hardlink")
+    # ctime changed because the hard-link count went down. But no mtime changes
+    # because file content isn't modified. No permissions changes either.
+    # This is another hint that something hard-link related changed.
     assert_line_exists(lines, "ctime:.*input/b/hardlink")
     assert_line_not_exists(lines, ".*mtime:.*input/b/hardlink")
     assert_line_not_exists(lines, "modified.*B.*input/b/hardlink")
+    assert_line_not_exists(lines, "-[r-][w-][x-].*input/b/hardlink")
 
     # Finally, compare test1 and test2.
     output = cmd(archiver, "diff", "test1", "test2", exit_code=EXIT_SUCCESS)
--- a/src/borg/testsuite/archiver/diff_cmd_test.py
+++ b/src/borg/testsuite/archiver/diff_cmd_test.py
@@ -3,6 +3,7 @@ import os
 from pathlib import Path
 import stat
 import time
+import pytest
 
 from ...constants import *  # NOQA
 from .. import are_symlinks_supported, are_hardlinks_supported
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@cr1901 can you install pre-commit, so black runs automatically?

https://borgbackup.readthedocs.io/en/latest/development.html#building-a-development-environment

cr1901:
Okay, looks like I properly installed the pre-commits and CI is passing now.

src/borg/testsuite/archiver/diff_cmd_test.py
```diff
@@ -318,3 +319,83 @@ def test_sort_option(archivers, request):
...
+    os.mkdir(path_a)
+    os.mkdir(path_b)
+    hl_a = os.path.join(path_a, "hardlink")
+    hl_b = os.path.join(path_b, "hardlink")
+    create_regular_file(archiver.input_path, hl_a, contents=b"123456")
+    os.link(hl_a, hl_b)
+
+    cmd(archiver, "repo-create", RK_ENCRYPTION)
+    cmd(archiver, "create", "test0", "input")
+    os.unlink(hl_a)  # Don't duplicate warning message- one is enough.
```
ThomasWaldmann: nitpick: add a blank before the dash

src/borg/testsuite/archiver/diff_cmd_test.py
```diff
@@ -318,3 +319,83 @@ def test_sort_option(archivers, request):
...
+    # Compare test0 and test2.
+    output = cmd(archiver, "diff", "test0", "test2", exit_code=EXIT_SUCCESS)
+    lines = output.splitlines()
+    # Adding a file changes c/mtime.
+    assert_line_exists(lines, "[cm]time:.*[cm]time:.*input/a$")
+    # Different c/mtime but no apparent changes (i.e. perms) or content
+    # modifications should be a hint that something hard-link related is going on.
+    assert_line_exists(lines, "[cm]time:.*[cm]time:.*input/a/hardlink")
+    assert_line_not_exists(lines, "modified.*B.*input/a/hardlink")
+    assert_line_not_exists(lines, "[.* -> .*].*input/dir_replaced_with_file")
```
ThomasWaldmann: c&p error?

src/borg/testsuite/archiver/diff_cmd_test.py
```diff
@@ -318,3 +319,83 @@ def test_sort_option(archivers, request):
...
+    # The hard-link count went down, but file content isn't modified.
```
ThomasWaldmann: It doesn't output info about hl count. But we see the ctime change because of the hl count change.

ThomasWaldmann (changes requested):
Thanks for the PR. Some minor things I found...

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: Is that how hyphens are written down in Germany? I've always put the hyphen right after the last letter of the preceding word- like in this sentence here :).

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: Yes, should be `input/a/hardlink`.

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: Yes, this is intentional; there's no "hl count changed" diff field; it is incorporated into the fact that `ctime` changed along with nothing else. I'll fix the comment to make that more clear. In fact, I need to add:

```
assert_line_not_exists(lines, "[.* -> .*].*input/b/hardlink")
```

as well.

src/borg/testsuite/archiver/diff_cmd_test.py
ThomasWaldmann: Interesting. I have always seen it like an operator and put blanks on each side.

Claude.AI says:

In English, there are different types of dashes, and the spacing depends on which one you use:

Em Dash (—)

This is the long dash

American English: No spaces

Example: "She was determined—more than ever—to succeed."

British English: Spaces before and after

Example: "She was determined — more than ever — to succeed."


En Dash (–)

This is the medium-length dash

Mainly used for ranges (e.g., "pages 10–15")
No spaces when used for ranges

When used as a dash in text: spaces on both sides

In practice, American English most commonly uses the em dash without spaces, while British English uses both versions (with and without spaces).

For comparison: In German, you typically use the en dash with spaces ("Sie war entschlossen – mehr denn je – erfolgreich zu sein").

cr1901: Okay, so it appears that I've been using em/en-dashes incorrectly for the better half of my life. I'll change it to have a space. An spaced e*n*-dash does seem to be [acceptable](https://en.wikipedia.org/wiki/Dash#cite_ref-Bringhurst_9-2) for this comment.

Now if you'll excuse me, I have 10+ years of writing online that I'll have to correct :).

ThomasWaldmann: I didn't verify claude's output, did you? :-)

ThomasWaldmann: https://cmosshoptalk.com/2024/01/23/hyphens-and-dashes-a-refresher/

"More commonly, dash refers to the punctuation that’s used to set something off from the surrounding text (much like parentheses). In Chicago style, such dashes consist of em dashes—like this—with no space before or after. In British style, spaced en dashes – like this – are more common."

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: This actually was incorrect; for local repos, `repo-create` changes the `umask`, so the permissions for `input/a/hardlink` **will** have changed.

`[.* -> .*]` is just a plain bad regex from what I intended (search for permission changes), so I made the regex better for both this and [`input/b/hardlink`](https://github.com/borgbackup/borg/pull/9031#discussion_r2392654220) as well.

src/borg/testsuite/archiver/diff_cmd_test.py
cr1901: > I didn't verify claude's output, did you? :-)

For better or worse, I consider Wiki to be a reliable source in 2025. Gone are the high school days where you couldn't site it.

src/borg/testsuite/archiver/diff_cmd_test.py
```diff
@@ -384,11 +384,16 @@ def test_hard_link_deletion_and_replacement(archivers, request):
     # modifications should be a hint that something hard-link related is going on.
     assert_line_exists(lines, "[cm]time:.*[cm]time:.*input/a/hardlink")
     assert_line_not_exists(lines, "modified.*B.*input/a/hardlink")
-    assert_line_not_exists(lines, "[.* -> .*].*input/dir_replaced_with_file")
-    # The hard-link count went down, but file content isn't modified.
+    # local repo-create changed the umask of newly-created files.
+    if archiver in ("archiver",):
+        assert_line_exists(lines, "\-[r-][w-][x-].*input/a/hardlink")
```
ThomasWaldmann: maybe just move the repo-create up higher, so borg's umask is set before you create any dirs / files?

cr1901: Fixed. Also cleaned up the perms regex to avoid a `DeprecationWarning` for escaping `-` (sometimes it's required, sometimes not).

cr1901:
Yes, I am finished. Feel free to merge.

ThomasWaldmann:
Thanks for helping!

PR #9089, opened 2025-10-16
extract: misc. fixes for restoring linux fs flags

ThomasWaldmann:


src/borg/platform/linux.pyx
```diff
@@ -136,17 +138,31 @@ def set_flags(path, bsd_flags, fd=None):
...
-            flags |= linux_flag
+            new_flags |= linux_flag
+
     open_fd = fd is None
     if open_fd:
         fd = os.open(path, os.O_RDONLY|os.O_NONBLOCK|os.O_NOFOLLOW)
     try:
+        # Get current flags. If this fails, fall back to 0 so we can still attempt to set.
+        if ioctl(fd, FS_IOC_GETFLAGS, &flags) == -1:
+            flags = 0
```
ThomasWaldmann: Discuss: shall we use flags=0 and try to continue or shall we rather just give up and not try to set the flags if we can't determine the current flags from the fs?

ThomasWaldmann: changed to just give up, see comment in src.

frispete:
@ThomasWaldmann This approach looks good to me, @jirislaby what do you think?

PR #9096, opened 2025-10-21
json: include archive keys in JSON lines when requested via --format, fixes #9095

### LINKED ISSUE #9095: `borg list --json-lines` does not add  archivename or archiveid if specified in `format`
Borg version: 1.4.1 (though it looks like this applies to the 2.x betas as well, I havenʼt tested there)

According to the documentation, if you run `borg list --json-lines <archive> --format="<format>"`, it will include all fields specified in `<format>` in the output.  This appears to be true for all fields *except* the `archivename` or `archiveid`.  I can see in the code that those are handled separately (they are put in [`static_keys`](https://github.com/borgbackup/borg/blob/1.4-maint/src/borg/helpers/parseformat.py#L796), which does not get added to `item_data` if using `--json-lines`), but it can be useful to have some details of the archive in lines about the contents of the archive.

I can think of various workarounds, including just `sed`ding in the appropriate fields of the output, but having them included correctly by Borg would simplify the process if I want to get details of multiple archives in a repository.

ThomasWaldmann:


dhouck:
Oops I accidentally commented on the issue not here.

This looks good to me and works well in testing I have a couple thoughts:
1. I wonder if line 807 should be `for k in static_keys:` or something.  Probably not; that includes things like `NL` that you might include in a format but definitely not want in the json.
2. I notice `extra` also isnʼt ever in the json and this isnʼt documented.  It doesnʼt make sense to include it in the json, probably, and Iʼm not sure if itʼs even worth documenting.  Not including `bpath` in the json is documented, and thatʼs the only other difference I could find besides `healthy` being a bool vs. a string.

ThomasWaldmann:
1. -> exactly!
2. extra is hardlink/symlink related pre-formatted text, like "-> linktarget" as a convenience placeholder for normal console output, doesn't make sense like that in json.

Issue #8474, opened 2024-10-15
fix windows "file:" URLs, re-enable windows CI


ThomasWaldmann:
See there: https://github.com/borgbackup/borgstore/issues/82

Re-enable windows CI after fixing this.

ThomasWaldmann:
This was fixed in borgstore recently.

Windows: `file:///` + full windows path including drive letter (using fwd slashes)

Everything else: `file://` + full absolute path, starting with a slash

valtron:
There are two issues with [this fix for windows](https://github.com/borgbackup/borgstore/pull/100/files#diff-768543a009d0e9d606898d686d69cbd4632358acbe2efa765a999b8487d436e6R33):
- it expects the first slash (after the drive letter) to be `/`; because of [`abspath`](https://github.com/borgbackup/borg/blob/d0dabe69a84e0d1ec4997fcdecddd1bdba8c8046/src/borg/helpers/parseformat.py#L561), this will always be `\`
- it expects `file:///` (three slashes) but because of [this](https://github.com/borgbackup/borg/blob/d0dabe69a84e0d1ec4997fcdecddd1bdba8c8046/src/borg/repository.py#L108), paths will start with the drive letter; basically, the file URI scheme is better thought as `file://<host><path>`, where `<path>` must always start with `/` (true for absolute POSIX paths, not for windows paths!)

Changing
```
windows_file_regex = r"""
    file://  # only empty host part is supported.
    (?P<drive_and_path>([a-zA-Z]:[/\\].*))  # path must be an absolute path.
"""
```
makes it work. (`[a-zA-Z]:[/\\]` is roughly the windows equivalent of `/`, i.e. is what determines whether a path is absolute.)

I can make a PR to borgstore if you'd like.

ThomasWaldmann:
Hi Valtron, thanks for helping with the windows stuff.

Yes, improvements for borgstore (first) and borg (second) so that it works on windows would be great.

I just added Windows CI testing for borgstore and there is a failing test now visible in CI.
Maybe that is related and should be fixed first.

In general, I think in URLs, there are always forward slashes (including on Windows), see https://en.wikipedia.org/wiki/File_URI_scheme .

Also, guess quite some code in borg expects forward slashes in paths and URLs also.

Please try to make a minimal good fix in borgstore first, thanks!

ThomasWaldmann:
@valtron can we close this, did you test the recent borg/borgstore code?

valtron:
[...]

PR #8727, opened 2025-04-08
Windows fixes

ThomasWaldmann:

PR #9097, opened 2025-10-24
Fix windows file URLs

valtron:
Ensure that file URLs created from windows paths are valid, i.e. look like `file:///C:/foo/bar`. Additionally I saw a few places where `preexec_fn` was being used on windows, so I added guards there.

Test results:
- Before: 554 failed, 889 passed, 395 skipped, 6 warnings, 107 errors
- After: 379 failed, 1080 passed, 395 skipped, 8 warnings, 91 errors

Related: https://github.com/borgbackup/borg/issues/8474

### DIFF
--- a/src/borg/legacyremote.py
+++ b/src/borg/legacyremote.py
@@ -277,7 +277,13 @@ class LegacyRemoteRepository:
             logger.debug("SSH command line: %s", borg_cmd)
             # we do not want the ssh getting killed by Ctrl-C/SIGINT because it is needed for clean shutdown of borg.
             self.p = Popen(
-                borg_cmd, bufsize=0, stdin=PIPE, stdout=PIPE, stderr=PIPE, env=env, preexec_fn=None if is_win32 else ignore_sigint
+                borg_cmd,
+                bufsize=0,
+                stdin=PIPE,
+                stdout=PIPE,
+                stderr=PIPE,
+                env=env,
+                preexec_fn=None if is_win32 else ignore_sigint,
             )  # nosec B603
             self.stdin_fd = self.p.stdin.fileno()
             self.stdout_fd = self.p.stdout.fileno()
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -578,7 +578,13 @@ class RemoteRepository:
             logger.debug("SSH command line: %s", borg_cmd)
             # we do not want the ssh getting killed by Ctrl-C/SIGINT because it is needed for clean shutdown of borg.
             self.p = Popen(
-                borg_cmd, bufsize=0, stdin=PIPE, stdout=PIPE, stderr=PIPE, env=env, preexec_fn=None if is_win32 else ignore_sigint
+                borg_cmd,
+                bufsize=0,
+                stdin=PIPE,
+                stdout=PIPE,
+                stderr=PIPE,
+                env=env,
+                preexec_fn=None if is_win32 else ignore_sigint,
             )  # nosec B603
             self.stdin_fd = self.p.stdin.fileno()
             self.stdout_fd = self.p.stdout.fileno()


src/borg/repository.py
```diff
@@ -105,11 +106,11 @@ def __init__(
         if isinstance(path_or_location, Location):
             location = path_or_location
             if location.proto == "file":
-                url = f"file://{location.path}"  # frequently users give without file:// prefix
+                url = _local_abspath_to_file_url(location.path)  # frequently users give without file:// prefix
```
ThomasWaldmann, line 109: how would a windows file:// url look like to need this?
what would be in location.path?

src/borg/repository.py
```diff
@@ -105,11 +106,11 @@ def __init__(
...
             else:
                 url = location.processed  # location as given by user, processed placeholders
         else:
-            url = "file://%s" % os.path.abspath(path_or_location)
+            url = _local_abspath_to_file_url(os.path.abspath(path_or_location))
```
ThomasWaldmann, line 113: so abspath would return something like \Windows\System32 ? 

Or C:\Windows\System32 ?

src/borg/repository.py
```diff
@@ -565,3 +566,8 @@ def store_delete(self, name, *, deleted=False):
     def store_move(self, name, new_name=None, *, delete=False, undelete=False, deleted=False):
         self._lock_refresh()
         return self.store.move(name, new_name, delete=delete, undelete=undelete, deleted=deleted)
+
+def _local_abspath_to_file_url(path: str) -> str:
+    if sys.platform in ("win32", "msys", "cygwin"):
+        path = "/" + path.replace("\\", "/")
```
ThomasWaldmann, line 579: guess prepending the slash is for cases when path starts with a drive letter.

maybe add a comment about what path arg looks like on Windows.

would it result in a valid url if path == "\Windows\System32"?

src/borg/repository.py
valtron: `_local_abspath_to_file_url` expects its argument to be a string representing an absolute path on the local system. In windows these look like `c:\foo\bar`.

src/borg/repository.py
valtron: The latter.

src/borg/repository.py
valtron: See https://github.com/borgbackup/borg/pull/9097#discussion_r2462233737

Analogous issues exist for POSIX. If a caller calls it with an invalid POSIX path, like `///foo`, it'll output an "invalid" file url `file://///foo`.

ThomasWaldmann: well, but \Windows\System32 is a valid windows path.

src/borg/repository.py
ThomasWaldmann: ok, so it would fail if a user just gives \myfiles as path?

src/borg/repository.py
valtron: I don't know of a normative reference that defines which strings represent valid paths on windows, so the best I can do is see how windows itself interprets these weird edge cases. `\Windows\System32` gets interpreted as a drive-relative path:
```
C:\Users\mti>cd \Windows\System32

C:\Windows\System32>d:

D:\>cd \Windows\System32
The system cannot find the path specified.
```
so it's not an absolute path.

src/borg/repository.py
valtron: `location.path` isn't the raw user input here, but after it's been [put through `abspath`](https://github.com/borgbackup/borg/blob/ca706652d48427d906ec690163e0e0c415eff6fb/src/borg/helpers/parseformat.py#L561), so it will always be an absolute path and start with a drive letter. (Or so I assume, I didn't look at how `os.path.abspath` is implemented.)

src/borg/repository.py
ThomasWaldmann: yeah, there are 2 kinds of "relative" on windows, interpreted with current drive and current path as context.

ThomasWaldmann: So, maybe put a comment below that "if", that the code expects a full path including drive letter.

valtron: Added docstring

ThomasWaldmann:
Linter is unhappy.

Maybe do that to avoid this automatically: https://borgbackup.readthedocs.io/en/master/development.html#building-a-development-environment

ThomasWaldmann:
Thanks!

Are changes in borgstore needed also?

ThomasWaldmann:
Hmm, looks like I overlooked another linter fail. I fixed it in #9108.

valtron:
> Are changes in borgstore needed also?

I didn't have any, and haven't run into any other issues so far.

ThomasWaldmann:
@valtron can you have a look at #8727?

I think there is a small overlap between what was fixed here and the fixes there, but it could be that we still need some fixes from there.

PR #9101, opened 2025-10-25
Fixes in `FilesCacheMixin`

valtron:
This PR includes two fixes in `FilesCacheMixin`
- fix usage of `discard_after` in `_write_files_cache` (should be `<= discard_after` instead of `< discard_after`); without this it always discards the file with the latest cmtime; this is also implied by the naming of `discard_after` (if `<` was correct it should be called `keep_before`)
- update `_newest_cmtime` in `_read_files_cache`; `_read_files_cache`, unlike `_build_files_cache`, wasn't maintaining `_newest_cmtime`; I wasn't sure about maintaining `_newest_path_hashes` though, so I didn't include that

### DIFF
--- a/src/borg/cache.py
+++ b/src/borg/cache.py
@@ -559,6 +559,8 @@ class FilesCacheMixin:
                             except KeyError:
                                 # repo is missing a chunk referenced from entry
                                 logger.debug(f"compress_entry failed for {entry}, skipping.")
+                            else:
+                                self._newest_cmtime = max(self._newest_cmtime, timestamp_to_int(entry.ctime), timestamp_to_int(entry.mtime))
                     except (TypeError, ValueError) as exc:
                         msg = "The files cache seems invalid. [%s]" % str(exc)
                         break


ThomasWaldmann:
Not sure these are fixes, let's see:

1. "update _newest_cmtime in _read_files_cache; _read_files_cache, unlike _build_files_cache, wasn't maintaining _newest_cmtime; I wasn't sure about maintaining _newest_path_hashes though, so I didn't include that"

The files cache read was previously written by _write_files_cache. That method did not write any file that has cmtime >= discard_after, so these are not present anymore in the on-disk cache files. Thus, there is no need to compute _newest_cmtime for the remaining files (or to drop any files, they already were dropped).

If borg create then adds new files to the files cache _newest_cmtime is updated on that basis and, when _write_files_cache is executed after that, again the files with the newest cmtimes are not written to the on-disk files cache.

Maybe I am missing something, but I don't see a bug needing a fix there.

ThomasWaldmann:
2. "fix usage of discard_after in _write_files_cache (should be <= discard_after instead of < discard_after); without this it always discards the file with the latest cmtime; this is also implied by the naming of discard_after (if < was correct it should be called keep_before)"

The intention of the code is to keep all files with older timestamps, but to drop all files with the latest timestamp. These are dropped because they could be affected by the issue described in the FAQ relating to snapshots and timestamp granularity.

Maybe I am missing something, but I don't see a bug needing a fix there.

Naming: yes, `keep_before` is a better variable name. looks like that in technical docs, before and after never include equality, but strictly mean "<" and ">".

valtron:
Here is the issue I ran into. With this script:
```
set -e
BORG="path to borg"
rm -rf mydata myrepo
mkdir -p mydata
touch mydata/foo
sleep 1
touch mydata/bar
$BORG -r myrepo repo-create -e none 2> /dev/null

echo "#1"
$BORG -r myrepo create myarchive mydata --stats 2>&1 | grep ' files:'
echo "#2"
$BORG -r myrepo create myarchive mydata --stats 2>&1 | grep ' files:'
echo "#3"
$BORG -r myrepo create myarchive mydata --stats 2>&1 | grep ' files:'
```

Output before this PR was (after removing irrelevant "number", "modified", and "error" lines to save space):
```
#1
Added files: 2
Unchanged files: 0
#2
Added files: 1
Unchanged files: 1
#3
Added files: 1
Unchanged files: 1
```
This is because it's always detecting the file with the latest cmtime (`mydata/bar`) as changed (added).

Output after the first commit in the PR:
```
#1
Added files: 2
Unchanged files: 0
#2
Added files: 0
Unchanged files: 2
#3
Added files: 2
Unchanged files: 0
```
This is because in every odd-numbered run, `_newest_cmtime == 0` so `discard_after == 0`.

Output after both commits:
```
#1
Added files: 2
Unchanged files: 0
#2
Added files: 0
Unchanged files: 2
#3
Added files: 0
Unchanged files: 2
```
This is the expected behaviour.

ThomasWaldmann:
"This is because it's always detecting the file with the latest cmtime (mydata/bar) as changed (added)."

Yes, this is expected, see the faq.

It is because it intentionally drops these files from the files cache when writing the cache to disk.

Issue #9072, opened 2025-10-14
Show warning on every access to an unencrypted repo?


bitwombat:
Due to being a dumb dumb, I've had an unencrypted repo for YEARS because my key had an empty password.

I only found out about it because I was restoring from a new machine and this message came up:

> Warning: Attempting to access a previously unknown unencrypted repository!

Two problems:

1) This message only shows once on a given machine.
2) It doesn't say which of two reasons it's "unencrypted". I did a `borg info` and saw that my repo was encrypted, so I almost shrugged it off as a bug it until a redditor mentioned I might have an empty passphrase.

Better would be for this sort of behavior, every access.
```
$ borg info server:repo
Warning: Repo is unencrypted (encryption=none set)
```

or

```
$ borg info server:repo
Warning: Repo is unencrypted (empty key passphrase)
```

People with a known unencrypted repo can set an env var to shut it up.

I know it's not terribly impressive that I let this go unnoticed, but it was silent and there will be other people hanging out with me on this part of the bell curve :)

ThomasWaldmann:
Technically, an empty passphrase on an encrypted repo does not mean the repo is unencrypted - it still is fully AES encrypted with a random encryption key.

But, if an attacker can access the borg key, the AES key won't be protected if there is an empty passphrase on it (similar also for a trivial-to-guess passphrase), so that would make it somehow equivalent to unencrypted from a security perspective. IIRC borg has some places where it considers it from that perspective.

Access to the borg key:
- repokey mode: the key is stored in the repo config, inside the repository. that could be e.g. on a remote server or on a USB disk or whereever you have your repository.
- keyfile mode: the key is stored on the client, in home directory

About `borg info`: will have to check the code if that can be changed easily.

borg could also in general emit a warning if the passphrase is empty. But, if intentional, that could be also annoying, especially if the warning also sets the return code to warning (rc != 0).

PhrozenByte:
> borg could also in general emit a warning if the passphrase is empty. But, if intentional, that could be also annoying, especially if the warning also sets the return code to warning (rc != 0).

[...]

PR #9103, opened 2025-10-26
Docs: Rewrite `borg init --encryption` docs

PhrozenByte:
This was one of these "okay, let's just add this one thing real quick" tasks after which one ends up with more or less a rewrite… :sweat_smile: The initial idea was to do https://github.com/borgbackup/borg/issues/9072#issuecomment-3448588042, but since I didn't find a place for it and since I always felt like that the `init --encryption` docs are a little too hard to understand for this crucial decision (taken by absolute beginners, because that's literally the first thing a user needs to do), I ended up more or less rewriting the `init --encryption` docs.

GitHub's diff is basically useless here. The diff looks a little better locally (at least with `histogram`), but still not really usable. I recommend reading it as if it's new.

I incorporated all previous content (often it's just moved around and rewritten slightly), nothing is lost. Yes, the docs are longer now, but I really feel like that the higher level of detail, the real-world examples, and some repetitions allow users to better understand what this whole thing is about. I also extended e.g. the hardware acceleration docs.

I made a few opinionated decisions in the text. For example, I've pushed Blake2 into the background a bit since SHA256 hardware acceleration is rather widespread now. About a few other things I'm fully aware that one can argue they aren't *always* true. I aimed for "95% of cases" (but whether that's actually reached is open for debate of course). Please read through it and if you feel like I'm stretching things too far, let me know, I'm absolutely fine with changing things accordingly. Also, there's always the chance that I understood things wrong, so let me know if I did.

### DIFF
--- a/docs/borg_theme/css/borg.css
+++ b/docs/borg_theme/css/borg.css
@@ -131,6 +131,17 @@ table.docutils.borg-options-table tr td:first-child:not([colspan="3"]) {
     text-align: center;
 }
 
+.borg-encryption-table th,
+.borg-encryption-table td {
+    vertical-align: top;
+}
+
+.borg-encryption-table .literal {
+    display: block;
+    margin: 0;
+    white-space: normal;
+}
+
 kbd, /* used in usage pages for options */
 code,
 .rst-content tt.literal,


ThomasWaldmann:
Tell me when this is ready for review.

ThomasWaldmann:
@PhrozenByte Junie and I did some changes, can you double-check them?

PhrozenByte:
Will do, also many thanks for fixing the ref issues! :+1:

I'm not quite sure whether and if yes how to include some of the clarifications in #9104. I want to do some benchmarking to verify the theory, but need to setup a device without hardware acceleration first. Will see then.

I'll give you a heads-up in the next couple of days when this is ready to merge :+1:

ThomasWaldmann:
BTW, I want to release 1.4.2 on halloween, so maybe we get this merged asap and do more changes later.

PhrozenByte:
I see, thanks for the info :+1:

I just commited another CSS fix for the encryption table for certain display / window widths. I had already checked your and Junie's corrections beforehand, everything fits, i.e. the PR can be merged :+1:

ThomasWaldmann:
master branch port: guess we may want to port SOME of the changes, but be careful, master branch crypto is different!

PhrozenByte:
Thanks for the heads-up! I will look into the master branch port after a possible follow-up for `1.4-maint` (if any). I already have a few suggestions for `repo-create` with Borg 2 in mind, not purely docs improvements, but I have to look a little deeper into the topic and will open a separate suggestion issue then.

PR #9122, opened 2025-11-01
drop Python 3.9 (EOL'ed by python.org) (1.4-maint)

### LINKED ISSUE #9056: drop official python 3.9 support for 1.4.x
As of 2025-10, Python 3.9 is end of life and does not get upstream support from python.org anymore.

But it might still be in use here and there:
- debian bullseye has py39
- ubuntu 20.04 "focal" has py38 (and py39?)
- other older stable dists

My plan is to remove official py39 support in 1.4-maint branch after borg 1.4.2 release.

I don't think this will cause major issue, because:
- older stable dists usually don't jump onto new borg releases anyway, but rather patch their old versions
- older borg versions, like 1.2.x and 1.4.y (y <= 2) will still support py39.
- even for > 1.4.2, it might be easy to get it working on py39 (just "as is", if you are lucky, or with minor python syntax patches, in case we use something that needed >3.9) - we just won't do testing on py39 anymore.

1.2-maint won't see any new 1.2.x releases.

For master branch (borg2), we already dropped py39 (but borg2 has no non-beta releases yet).

ThomasWaldmann:
Fixes #9056.

borg 1.4.2 and 1.2.9 were the last borg releases with official py39 support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,13 +50,12 @@ jobs:
         ${{ fromJSON(
           github.event_name == 'pull_request' && '{
             "include": [
-              {"os": "ubuntu-22.04", "python-version": "3.9", "toxenv": "py39-fuse2"},
+              {"os": "ubuntu-22.04", "python-version": "3.10", "toxenv": "py310-fuse2"},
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
About the master port:
- IIRC we already dropped py39 there
- some other changes might still need porting, e.g. the openindiana changes?

PR #9133, opened 2025-11-02
Bugfix: local path regex should not match s3 and b2 urls

f-froehlich:
There was an issue for matching the s3 and b2 urls because of an not updated regex for local repository paths. The local path was matching the s3 and b2 urls so the repository was created locally and not on s3/b2. Now it is possible to create and use s3 and b2 repositories.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -451,7 +451,7 @@ class Location:
     # path may contain any chars. to avoid ambiguities with other regexes,
     # it must not start with "//" nor with "scheme://" nor with "rclone:".
     local_path_re = r"""
-        (?!(//|(ssh|socket|sftp|file)://|rclone:))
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Hmm, we have `test_s3_repo_basics` and `BORG_TEST_S3_REPO` set on github actions, guess this needs checking why it did not fail.

Ah, haha, guess it just happily worked with the local repo it created then.

f-froehlich:
> Hmm, we have `test_s3_repo_basics` and `BORG_TEST_S3_REPO` set on github actions, guess this needs checking why it did not fail.
> 
> Ah, haha, guess it just happily worked with the local repo it created then.

yes exactly. It is happy while creating the local repository instead of the s3 one

ThomasWaldmann:
> [gw1] [ 52%] PASSED src/borg/testsuite/archiver/remote_repo_test.py::test_s3_repo_basics 

Now for real.

ThomasWaldmann:
Thanks!

PR #9135, opened 2025-11-02
CI: add GitHub artifact attestations for release binaries, fixes #9134 (master)

### LINKED ISSUE #9134: use artifact attestations for borg binaries built on github
Guess that might be a bit better than just the sha256:

https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations

ThomasWaldmann:
- grant id-token and attestations permissions to posix_tests job
- add actions/attest-build-provenance@v3 step for built artifacts

This publishes SLSA-style provenance for our tag builds (only when binaries are produced) so users can verify the origin of downloaded borg binaries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,10 @@ jobs:
   posix_tests:
 
     needs: [lint, security]
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Needed to merge this to really test it...

PR #9180, opened 2025-11-17
archive: preserve pre-open original atime, fixes #6194

### LINKED ISSUE #6194: atime touched inside backup archive if there is no NOATIME support
master branch: https://github.com/borgbackup/borg/blob/1.2.0b4/src/borg/testsuite/archiver.py#L534

this fails on cygwin. `sto.st_atime_ns` is newer than expected.

maybe this has to do with borg working fd-based now, so we open the files early and then do all file ops with the fd.
if opening already touches the atime before we do the fstat(), this could be the consequence.

https://github.com/borgbackup/borg/blob/1.2.0b4/src/borg/archive.py#L1334

not super important maybe as borg 1.2 by default does not store the atime into an archive and other platforms do have NOATIME support.

ThomasWaldmann:
When creating an archive with --atime on platforms without O_NOATIME, opening a file for reading could update atime before we recorded it (thus we archived the updated atime, not the original one).

Capture pre-open path-based-stat atime and use it
if it pre-dates the atime we got from the fd AND
if it refers to same fs object (avoid race condition).

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1088,7 +1088,7 @@ class MetadataCollector:
         self.noxattrs = noxattrs
         self.nobirthtime = nobirthtime
 
-    def stat_simple_attrs(self, st, path, fd=None):
+    def stat_simple_attrs(self, st, path, fd=None, *, st_before_open=None):
         attrs = {}
         attrs["mode"] = st.st_mode
         # borg can work with archives only having mtime (very old borg archives do not have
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
difficult to test in master: windows is too broken, lots of fail in cygwin.

ThomasWaldmann:
check in master again, on cygwin and didn't see an improvement.

PR #9192, opened 2025-11-21
CI: backport vm_tests (FreeBSD/NetBSD/OpenBSD/Haiku) from master branch

ThomasWaldmann:


ThomasWaldmann:
netbsd/freebsd: hostname seems to be empty - why?

does master have other code or hostname or why does it not fail there?

update: added setting the hostname as a workaround, also filed a bug at c-p-a project.

ThomasWaldmann:
the netbsd failures are not due to borg code. the workers sometimes crash, this has been an unresolved netbsd issue since long, i have also seen that in virtualbox VMs.

PR #9193, opened 2025-11-22
Fix ChunkerFixed sparse handling and update tests

ThomasWaldmann:


src/borg/testsuite/chunkers/fixed_test.py
```diff
@@ -40,7 +40,40 @@ def get_chunks(fname, sparse, header_size):
 
     fn = str(tmpdir / fname)
     make_sparsefile(fn, sparse_map, header_size=header_size)
-    get_chunks(fn, sparse=sparse, header_size=header_size) == make_content(sparse_map, header_size=header_size)
```
ThomasWaldmann, line 43: an assert was missing here.

ThomasWaldmann:
see #9589.

PR #9203, opened 2025-12-01
compact: fix dealing with mismatching hints (1.4-maint)

ThomasWaldmann:


ThomasWaldmann:
TODO: Check if some bits of this should be ported to master. The borg2 repo code is totally different, but there is legacyrepository code there (mainly for reading though).

PR #9223, opened 2025-12-15
borg --cockpit: show TUI based on Textual

ThomasWaldmann:


ThomasWaldmann:
This is a rather hackish first implementation of a console-based status display.

This is **not** meant to grow into a fully interactive borg TUI with functionality like offered by borg GUIs (like Vorta, Pika, etc.), but should focus on status and progress display (for create, extract, check, ...).

It's currently based on the normal text log output of borg. In a better, future implementation it would use borg's json output to implement a lot more functionality (and likely, borg's json api will need some extensions as well to do that even better).

Issue #9254, opened 2026-01-09
[Usability Improvement] Improved handling of symlinks


MartinKurtz:
As about a week ago i and @ThomasWaldmann both were dumbfounded and wasted several hours on issue #9245 i have thought long and hard about how the handling of symlinks could be improved significantly, and here is what i have come up with.

In the following, “mount” or "mount command" refers to `borg mount` of an archive (or set of archives).

# The Problem
suppose you are backing up your home folder on a linux system

## What causes the problem
said home folder contains a path: `/home/martin/annoyingly/long/path/which/you/dont/wanna/type/out/every/time` along with a host of other files, so you create a symlink `/home/martin/annoyingpath` which points towards this path and often enough (especially when automatically created by some program), this symlink will point to the absolute path and borg will back it up exactly like that.

## What does it do

well after you've created your borg repo and backed up all your data for months, maybe one day a file in this very annoying path gets corrupted, overridden, whatever. you have decided you want to restore the file. as such you mount the borg repository and go to yesterdays archive you have mounted perhaps in `/mnt/borg/repo` and go look at `/mnt/borg/repo/yesterday/home/martin/annoyingpath`. 

and what do you see?

## but why?

well because borg has backed up the absolute-path-symlink exactly "as is", and of course it will point to the absolute path, but in the cli or the file explorer program that may not be immediately clear, and it may also not be always as easy to find, as with #9245 for example it was an absolute symlink which was hidden in a more than 10 folders long path.

additionally because accessing this folder may have become muscle memory one might not immediately think this is an issue, it is what has happened to me anyway.

# How to fix it?
well there are some easy and some harder ways to fix this.

## The easiest, most non-invasive and allmost trivial option
Borg upon calling `borg mount` could and should warn the user, that symlinks can and will point outside of the borg repository. a warning could look like this:

```
$borg mount <repo> <mountpoint>

WARNING: this repository may contain symlinks. Symlinks can and will point outside of the backed up folder structure. When following symlinks you may see files in the repository which are not contained in the repository, or are different versions of the files in the repository.
```
or perhaps alternatively

```
$borg mount <repo> <mountpoint>

WARNING: the mounted archive may contain symlinks that point outside the archive tree.
When following symlinks you may see files that are not part of the archive, or different
versions of files than those stored in the archive.
```
this or a similar warning does not prohibit other options from being used/implemented

This alone would have prevented the confusion in #9245, costs almost nothing to implement, and does not alter existing semantics in any way, shape or form.

[...]

PR #9255, opened 2026-01-10
Small change to warn about symlinks pointing outside of a backup

MartinKurtz:
Hi, this is related to https://github.com/borgbackup/borg/issues/9254, and should implement the most basic version of "the easy option" outlined there.

It implements a warning about symlinks pointing outside of the backup mountpoint when using borg mount and adds it to the `borg mount` documentation as well as logs such a warning on a mount call.

I will mark this as a draft as i see this more as a preliminary suggestion.

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -567,7 +567,7 @@ class FuseOperations(llfuse.Operations, FuseBackend):
                                 user=dir_user, group=dir_group, uid=dir_uid, gid=dir_gid)
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
-        logger.warning('ATTENTION: The mounted archive may contain symlinks that point outside the archive tree. \nWhen following such symlinks you may see files and directories within the mountpoint that do not reflect the archive content.')
+        logger.warning('Warning: The mounted archive is capable of containing symlinks that point outside the archive tree. When following such symlinks you may see files and directories within the mountpoint that do not reflect the archive content.')
         if not foreground:
             if isinstance(self.repository_uncached, RemoteRepository):
                 daemonize()


docs/man/borg-mount.1
```diff
@@ -45,6 +45,11 @@ When mounting a repository, the top directories will be named like the
 archives and the directory structure below these will be loaded on\-demand from
 the repository when entering these directories, so expect some delay.
 .sp
+Care should be taken, as borg will backup symlinks as is,
+and when having mounted an archive or repository,
+it is possible to "jump" out of the mountpoint by following a symlink.
+As of version 1.4.4 borg will print a warning when mounting for this reason.
```
ThomasWaldmann: Don't edit the manpage directly, it is auto-generated.

The original text is in archiver.py, only edit there.

ThomasWaldmann: You do not need to mention 1.4.4 and that it outputs a warning here, as the docs are versioned, so everybody can read the docs matching their borg version.

src/borg/fuse.py
```diff
@@ -567,6 +567,7 @@ def pop_option(options, key, present, not_present, wanted_type, int_base=0):
                                 user=dir_user, group=dir_group, uid=dir_uid, gid=dir_gid)
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
+        logger.warning('ATTENTION: The mounted archive may contain symlinks that point outside the archive tree. \nWhen following such symlinks you may see files within the mountpoint that are not part of the archive, \nor different versions of files than those stored in the archive.')
```
ThomasWaldmann: Isn't the last part of the sentence redundant?

MartinKurtz: i have altered it to address that

src/borg/fuse.py
```diff
@@ -567,6 +567,7 @@ def pop_option(options, key, present, not_present, wanted_type, int_base=0):
                                 user=dir_user, group=dir_group, uid=dir_uid, gid=dir_gid)
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
+        logger.warning('ATTENTION: The mounted archive may contain symlinks that point outside the archive tree. \nWhen following such symlinks you may see files and directories within the mountpoint that do not reflect the archive content.')
```
MartinKurtz: One change i am still thinking about making is to reformulate the first part of this to be less alarmist, somethink along the likes of "Note: Symlinks within the archive are capable of pointing outside the archive tree"

what do you think

ThomasWaldmann: Well, they both transport the same meaning. Whatever you think is better.

Is there a special reason to use " \n"? Maybe avoid having trailing blanks.

src/borg/archiver.py
```diff
@@ -3227,6 +3227,13 @@ def define_borg_mount(parser):
         archives and the directory structure below these will be loaded on-demand from
         the repository when entering these directories, so expect some delay.
 
+        Care should be taken, as Borg backs up symlinks as-is. When an archive 
+        or repository is mounted, it is possible to “jump” outside the mount point 
+        by following a symlink. If this happens, files or directories (or versions of them)
+        that are not part of the archive or repository may appear to be within the mount point.
+        To avoid this behavior, it is recommended to use ``borg extract`` 
+        to retrieve files from backups.
```
ThomasWaldmann: The last sentence is misleading IMHO.

In both cases, symlinks may point to unexpected places (except if the extraction is to same directory where the files were originally located).

ThomasWaldmann: this is not a borg-specific problem btw, if you just copy a tree to a different location, you have the same problem already.

src/borg/fuse.py
MartinKurtz: yeah youre right, "\n" doesnt really matter here, its gonna wrap anyway by itself if needed, ill remove that.

src/borg/archiver.py
MartinKurtz: ive removed that last sentence

src/borg/fuse.py
MartinKurtz: i have settled on a wording i like

MartinKurtz:
I think i am happy with the wordings now

ThomasWaldmann:
Thanks!

Can you do a port to master branch?

MartinKurtz:
does that mean a second PR to the master branch with the same change? or is that some special git feature

ThomasWaldmann:
Yes, that means another PR to master branch.

Considering the differences between 1.4-maint and master, it is easiest to do manually.

archiver.py was split into many files inside archiver/ in master.

PR #9260, opened 2026-01-19
 Small change to warn about symlinks pointing outside of a backup

MartinKurtz:
this is a port of #9255 to master, i will mark this as draft until i have installed and tested this, as small as the changes are

### DIFF
--- a/src/borg/fuse.py
+++ b/src/borg/fuse.py
@@ -566,8 +566,8 @@ class FuseOperations(llfuse.Operations, FuseBackend):
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
         logger.warning(
-            "Warning: The mounted archive is capable of containing symlinks that point outside the archive tree."
-            "When following such symlinks you may see files and directories within the mountpoint"
+            "Warning: The mounted archive is capable of containing symlinks that point outside the archive tree. "
+            "When following such symlinks you may see files and directories within the mountpoint "
             "that do not reflect the archive content."
         )
         if not foreground:


MartinKurtz:
functions for me now

src/borg/fuse.py
```diff
@@ -565,6 +565,11 @@ def pop_option(options, key, present, not_present, wanted_type, int_base=0):
         )
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
+        logger.warning(
+            "Warning: The mounted archive is capable of containing symlinks that point outside the archive tree. \
+When following such symlinks you may see files and directories within the mountpoint \
+that do not reflect the archive content."
+        )
```
ThomasWaldmann, line 572: Instead of using backslash to escape the line endings, one can also do it like that:

```
logger.warning(
    "..."
    "..."
)
```

MartinKurtz: yes, i know. i can change it of course, but i prefer this.

ThomasWaldmann: Yes, guess we should change it. I have a slight preference for that method and it seems I am not alone:

https://gemini.google.com/share/45197bda6024

MartinKurtz: ive changed it, will need to test again for this though

MartinKurtz: seems to work

src/borg/fuse.py
```diff
@@ -566,9 +566,9 @@ def pop_option(options, key, present, not_present, wanted_type, int_base=0):
         self._create_filesystem()
         llfuse.init(self, mountpoint, options)
         logger.warning(
-            "Warning: The mounted archive is capable of containing symlinks that point outside the archive tree. \
-When following such symlinks you may see files and directories within the mountpoint \
-that do not reflect the archive content."
+            "Warning: The mounted archive is capable of containing symlinks that point outside the archive tree."
```
ThomasWaldmann: Add a blank or linefeed or it'll look bad when the next line is appended.

MartinKurtz: ah ffs, yeah ive already got a commit up there for the exact same thing

MartinKurtz: done

ThomasWaldmann:
Thanks for helping!

PR #9171, opened 2025-11-14
file url fixes

valtron:
Uses `Path.as_uri` to create file URIs from in all the places that I found.

Issue #7294, opened 2023-01-22
mapchars: map invalid chars to remap range (win32)


ThomasWaldmann:
This might be useful if we encounter strange filenames with borg on win32:

https://www.kernel.org/doc/html/latest/admin-guide/cifs/usage.html

**mapchars**

Translate six of the seven reserved characters (not backslash): `*?<>|:` to the remap range (above 0xF000), which also allows the CIFS client to recognize files created with such characters by Windows’s POSIX emulation. This can also be useful when mounting to most versions of Samba (which also forbids creating and opening files whose names contain any of these seven characters). This has no effect if the server does not support Unicode on the wire.

ThomasWaldmann:
Fixed in master by #9264.

PR #9264, opened 2026-01-22
path sep and windows related changes

### LINKED ISSUE #4731: windows: normalize backslashes to slashes
with borg's native windows build, users might use backslashes for pathes when invoking borg.

as borg is platform-independent, it should be possible to extract data archived on windows on e.g. linux, so i guess we do not want to have backslashes in archived item metadata, but only forward slashes.

there might be also some code in borg, that only works correctly for slashes, not sure.

so the idea is to normalize to forward slashes:

`borg create repo::archive \some\path1` -> `borg create repo::archive /some/path`

In the archive (`borg list`), this needs to look like `some/path` later as borg only stores relative pathes.

Guess this can be done by using a special argparse validator for processing path-like cli arguments that does the transformation.

Guess it is wise to completely ignore the existence of drive letters to limit the scope of the fix to just path separators working as they should. We can just assume that we only archive data from the current drive and a drive letter is not used within cli arguments.

Later, fix drive letters (only) in a separate PR.

### LINKED ISSUE #7120: os.path.normpath converts to backslashes on windows
Just stumbled over this:

https://docs.python.org/3/library/os.path.html#os.path.normpath

esp.: **On Windows, it converts forward slashes to backward slashes.**

For usage in the filesystem that might be fine/expected, but in borg we also use normpath for matching patterns and paths in borg archives - for both we **always** (on linux anyway, but guess also on windows) use the normal slashes, not backslashes.

So guess this will cause quite some breakage on windows currently.

### LINKED ISSUE #8474: fix windows "file:" URLs, re-enable windows CI

ThomasWaldmann:
Fixes #7120. Fixes #4731. Fixes #8474.

Also incorporated the changes from #9171, thanks to @valtron!

### DIFF
--- a/src/borg/archiver/help_cmd.py
+++ b/src/borg/archiver/help_cmd.py
@@ -109,9 +109,14 @@ class HelpMixIn:
             **Windows path handling**: All paths in Borg archives use forward slashes (``/``)
             as path separators, regardless of the platform. When creating archives on Windows,
             backslashes from filesystem paths are automatically converted to forward slashes.
-            When extracting archives created on POSIX systems that contain literal backslashes
-            in filenames (which is rare, but possible), the backslash character is replaced
-            with ``%`` on Windows to prevent misinterpretation as a path separator.
+
+        .. note::
+
+            **Windows reserved characters**: On Windows, when extracting archives created on
+            POSIX systems, paths may contain characters that are reserved from being used in
+            file or directory names (like: ``< > : " \\ | ? *``).
+            These are replaced by characters in the unicode private use area (``U+F0xx``) like
+            the CIFS mapchars feature also does it. It won't be pretty, but at least it works.
 
         Exclusions can be passed via the command line option ``--exclude``. When used
         from within a shell, the patterns should be quoted to protect them from
--- a/src/borg/helpers/fs.py
+++ b/src/borg/helpers/fs.py
@@ -252,7 +252,7 @@ def make_path_safe(path):
     if "\\.." in path or "..\\" in path:
         raise ValueError(f"unexpected '..' element in path {path!r}")
 
-    path = percentify(path)
+    path = map_chars(path)
 
     path = path.lstrip("/")
     if path.startswith("../") or "/../" in path or path.endswith("/..") or path == "..":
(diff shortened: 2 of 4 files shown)


ThomasWaldmann:
@jedenastka do you have feedback about the contents of the plan?

jedenastka:
My concern is that it is AI generated. To me, this signals laziness and lack of thought, which is not something I want to see in a backup tool.

A *plan* means you're actually *planning* something. If AI does it, it is not planning, because AI cannot think.

Feel free to disagree. I do not plan to engage in further conversation.

ThomasWaldmann:
@jedenastka Well, I understand your concerns. I also have quite some concerns about AI usage, but (as so often), it depends on how you use it.

The planning one can see here is the result of an hours long "discussion" between me and the AI, so maybe that was not what you meant with "lazy" and "lack of thought". It helped not to overlook aspects or places of the needed changes.

Just wanted to clarify that. You don't need to reply if you do not have more specific feedback about the plan contents.

RonnyPfannschmidt:
wrt drive letters and extraction - how about storing the "extraction root" in the archive in a way that all items in a archive are relative paths to the extraction root

this would come with some perks like backing up the same file tree in a different location would generate the same archive content including path names for all but the archive metadata


wrt forbidden path names on windows - would using unc full paths work - i dislike schemes that may trigger a overlap - and statefull overlap prevention is unreasonably expensive in terms of implementation complexity

ThomasWaldmann:
@RonnyPfannschmidt one root would only be the simplest use case, but would not be enough for the general case. You can give multiple paths to borg as cli params or via the patterns file. You can even feed a list of arbitrary paths to borg.

trxvorr:
@ThomasWaldmann I tested on Windows 11 (Native) using the MSYS2 UCRT64 environment.

Works as expected. The path normalization logic (`slashify`, `percentify`) correctly handles Windows paths.

### Environment
* OS: Windows 11
* Build Env: MSYS2 UCRT64 (GCC/MinGW)
* Python: 3.12.12
* Dependencies: Installed via `scripts/msys2-install-deps development`

### 1. Build Configuration
* Built inside a virtual environment with `--system-site-packages` to inherit MSYS2's Python packages.
* Note for other testers: If you get a version parsing error like `Invalid version string 0.1.dev...`, ensure you fetch upstream tags first (`git fetch upstream --tags`). Alternatively, `export SETUPTOOLS_SCM_PRETEND_VERSION=2.0.0` works as a workaround.

### 2. Automated Tests (`fs_test.py`)
* **Command:** `pytest --pyargs borg.testsuite.helpers.fs_test -rs --benchmark-skip`
* **Result:** 47 passed, 1 failed.
* **Failure Analysis:** The single failure in `test_get_base_dir` is **unrelated to this PR**. The test asserts POSIX-style paths (e.g., `/var/tmp/home`) which don't resolve the same way on Windows.

### 3. Manual Function Verification
I wrote a script to specifically test the functions modified in `src/borg/helpers/fs.py`:

| Function | Result |
|----------|--------|
| `slashify()` | Correctly converts `C:\Users\test` → `C:/Users/test` |
| `percentify()` | Correctly replaces backslashes with `%` |
| `make_path_safe()` | Validates paths and rejects `..` traversal (both `/` and `\` variants) |

The path handling changes in `src/borg/archive.py` and `create_cmd.py` integrate these helpers correctly, enabling consistent archive creation and extraction on Windows.

I think its ready to merge. The core path normalization logic works correctly.

ThomasWaldmann:
@trxvorr Thanks for testing/reviewing.

Guess `percentify` is not good enough yet, there are more "special" chars in windows paths that can't be just used as normal chars, see #7294.

ThomasWaldmann:
@trxvorr Improved that, `percentify` is now `map_chars`.

ThomasWaldmann:
@trxvorr Merged it.

trxvorr:
@ThomasWaldmann awesome glad I could help verify the Windows behavior

didn't get time to test map_chars will do it soon and then work on #9279

PR #9271, opened 2026-01-25
Added info output to prune command

### LINKED ISSUE #9262: `borg --verbose prune` could print archive counts
`borg --verbose prune` currently doesn't output anything useful during a normal run. It could print some archive counts:
- number of archives (in the repository)
- number of glob-matched archives (if `--glob-archives` is used)
- number of pruned archives (in a separate line at the end if unknown at the beginning)

Konmous04:
Fixes #9262.

-v / --info displays archive counts (total, kept, pruned).

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -1658,7 +1658,6 @@ class Archiver:
                 keep += prune_split(archives, rule, num, kept_because)
 
         to_delete = (set(archives) | checkpoints) - (set(keep) | set(keep_checkpoints))
-        logger = logging.getLogger('borg.output.prune')
         pruned_checkpoints_len = len(set(checkpoints) - set(keep_checkpoints))
         pruned_archives_len = len(to_delete) - pruned_checkpoints_len
         logger.info('Found %d normal archives and %d checkpoint archives.',


src/borg/archiver.py
```diff
@@ -1705,12 +1705,23 @@ def checkpoint_func():
             elif uncommitted_deletes > 0:
                 checkpoint_func()
             if args.stats:
+                stats_logger = logging.getLogger('borg.output.stats')
+                stats_logger.info('Archive counts:')
+                stats_logger.info('Number of archives:                %d', len(manifest.archives))
```
ThomasWaldmann: that way, the numbers are left-aligned.

guess I'ld prefer either:
- no alignment (simplest)
- right aligned

src/borg/archiver.py
```diff
@@ -1705,12 +1705,23 @@ def checkpoint_func():
...
+                
+                if args.glob_archives:
+                    stats_logger.info('Number of archives matching pattern: %d', len(archives_checkpoints))
```
ThomasWaldmann: maybe say "globbing pattern" to be more specific.

src/borg/archiver.py
```diff
@@ -1705,12 +1705,23 @@ def checkpoint_func():
...
+                stats_logger.info('Number of archives pruned:         %d', len(to_delete))
+                stats_logger.info(' ')  
+
                 log_multi(DASHES,
                           STATS_HEADER,
                           stats.summary.format(label='Deleted data:', stats=stats),
                           str(cache),
                           DASHES, logger=logging.getLogger('borg.output.stats'))
-
+
```
ThomasWaldmann: remove

Konmous04:
I updated the code according to your suggestions. The stats are now displayed via logger info before the deletion loop.

src/borg/archiver.py
```diff
@@ -1704,13 +1712,13 @@ def checkpoint_func():
...
                 checkpoint_func()
-            if args.stats:
+
                 log_multi(DASHES,
                           STATS_HEADER,
                           stats.summary.format(label='Deleted data:', stats=stats),
                           str(cache),
                           DASHES, logger=logging.getLogger('borg.output.stats'))
-
+
```
ThomasWaldmann: remove

src/borg/archiver.py
```diff
@@ -1658,6 +1658,14 @@ def do_prune(self, args, repository, manifest, key):
                 keep += prune_split(archives, rule, num, kept_because)
 
         to_delete = (set(archives) | checkpoints) - (set(keep) | set(keep_checkpoints))
+        stats_logger = logging.getLogger('borg.output.prune')
```
ThomasWaldmann: the new output is not stats, just use the normal logger.

src/borg/archiver.py
```diff
@@ -1658,6 +1658,14 @@ def do_prune(self, args, repository, manifest, key):
...
+        pruned_checkpoints_len = len([a for a in to_delete if a in checkpoints])
```
ThomasWaldmann: avoid loops if you can use set operations: len(set(checkpoints) - set(keep_checkpoints))

src/borg/archiver.py
```diff
@@ -1704,13 +1712,13 @@ def checkpoint_func():
                 raise Error("Got Ctrl-C / SIGINT.")
             elif uncommitted_deletes > 0:
                 checkpoint_func()
-            if args.stats:
```
ThomasWaldmann, line 1707: don't change the existing code, this is unrelated to your change.

ThomasWaldmann:
BTW, did you practically test it how the output looks like when mixing -v / --list / --stats / --progress in different combinations?

--list and --progress are expected to not look good when used together, but the other combinations should look ok.

Konmous04:
I have updated the code according to your feedback:

Replaced list comprehensions with set operations (len(set(...) - set(...))) for better performance.

Restored the original if args.stats: block at the end that I had accidentally removed.

Verified that the output looks correct when combining -v with --list, --stats, and --progress.

src/borg/archiver.py
```diff
@@ -1658,6 +1658,13 @@ def do_prune(self, args, repository, manifest, key):
                 keep += prune_split(archives, rule, num, kept_because)
 
         to_delete = (set(archives) | checkpoints) - (set(keep) | set(keep_checkpoints))
+        logger = logging.getLogger('borg.output.prune')
```
ThomasWaldmann: you already have the normal logger, you do not need this line.

just use logger.info("...")

ThomasWaldmann (changes requested):

Konmous04:
Removed the redundant logger line

ThomasWaldmann:
Thanks @Konmous04!

PR #9276, opened 2026-01-28
binary build: install cockpit,s3,sftp extras

ThomasWaldmann:


ThomasWaldmann:
In the binary build, the cockpit is still crashing due to unicode stuff missing in `rich`.

PR #9287, opened 2026-02-07
docs: consolidate key backup info in borg key export, fixes #6204

### LINKED ISSUE #6204: Documentation contradiction: required backup of keyfile

ThomasWaldmann:
Consolidate key backup documentation into `borg key export` and reference it from Quickstart and FAQ to avoid duplication and inconsistency.

Clarify that while `repokey` or `authenticated` mode stores the key in the repo, a separate backup is still recommended to protect against repository corruption or data loss.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -350,20 +350,9 @@ For automated backups the passphrase can be specified using the
 .. warning:: The repository data is totally inaccessible without the key
     and the key passphrase.
 
-    Make a backup copy of the key file (``keyfile`` mode) or repo config
-    file (``repokey`` mode) and keep it at a safe place, so you still have
-    the key in case it gets corrupted or lost. Also keep your passphrase
-    at a safe place. You can make backups using :ref:`borg_key_export`
-    subcommand.
+    In any case, make a backup of the borg key, see :ref:`borg_key_export` for
+    more details.
 
-    If you want to print a backup of your key to paper use the ``--paper``
-    option of this command and print the result, or print this `template`_
-    if you need a version with QR-Code.
-
-    A backup inside of the backup that is encrypted with that key/passphrase
-    won't help you with that, of course.
-
-.. _template: paperkey.html
 
 .. _remote_repos:
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -381,11 +381,10 @@ How important is the $HOME/.config/borg directory?
 The Borg config directory has content that you should take care of:
 
 ``keys`` subdirectory
-  All your borg keyfile keys are stored in this directory. Please note that
-  borg repokey keys are stored inside the repository. You MUST make sure to have an
-  independent backup of these keyfiles, otherwise you cannot access your backups anymore if you lose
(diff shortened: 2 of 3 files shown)


docs/quickstart.rst
```diff
@@ -350,18 +350,8 @@ For automated backups the passphrase can be specified using the
...
-    If you want to print a backup of your key to paper use the ``--paper``
-    option of this command and print the result, or print this `template`_
-    if you need a version with QR-Code.
-
-    A backup inside of the backup that is encrypted with that key/passphrase
-    won't help you with that, of course.
+    In any case, make a backup of the borg key, see :ref:`borg_key_export` for
+    more details.
 
 .. _template: paperkey.html
```
ThomasWaldmann: remove

PR #9288, opened 2026-02-08
extract: do not delete existing directory if possible, fixes #4233

### LINKED ISSUE #4233: "borg extract" removes empty btrfs subvolumes prior to extraction

Yes

Bug

btrfs destination, xfs source, local usage

1.1.8

Gentoo Linux 64-bit

```
borg extract -v --list path/to/repo::archive -e something/i/dont/want
```

Prior to extracting the backup to an empty btrfs, I recreate my subvolume structure including nocow and compression flags (chattr) from a script generated during backup. This step is needed because borg doesn't restore btrfs subvolumes as subvolumes but only as directories. Since some kernel version, it is allowed to remove empty subvolumes with just "rmdir". Borg seems to do exactly that: It removes empty directories before restoring content into it. I dind't expect that behavior and there's no command line option to work around this. The result is a btrfs with a lot of subvolumes killed (except they contain nested subvolumes).

I'm currently letting my script use `touch .keep-subvolume` for each subvolume it creates. This seems to work around the issue: Borg seems to no longer kill the subvolumes because it only removes empty directories.

It would be good if borg wouldn't kill empty subvolumes. Even better would be if it restored subvolmes as subvolumes instead of plain directories if restoring to a btrfs volume. Maybe it could even store the nocow/compress attribute (lsattr) for subvolumes, tho it could make sense to store these even per file (but I don't use that per file).

ThomasWaldmann:
A pre-existing directory might be a btrfs subvolume that was created by the user ahead of time when restoring several nested subvolumes from a single archive.

If the archive item to be extracted is a directory and there is already a directory at the destination path, do not remove (and recreate) it, but just use it.

That way, btrfs subvolumes (which look like directories) are not deleted.

Fix originally contributed by @intelfx in #7866, but needed more work, so I thought more about the implications and added a test.

Note:

In the past, we first removed (empty) directories, then created a fresh one, then called restore_attrs for that. That produced correct metadata, but only for the case of an EMPTY exisiting directory. If the existing directory was not empty, the simply os.rmdir we tried did not work anyway and did not remove the existing directory.

Usually we extract to an empty base directory, thus encountering this edge case is mostly limited to continuing a previous extraction. In that case, calling restore_attrs again on a directory that already has existing attrs should be harmless, because they are identical.

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -782,12 +782,17 @@ Duration: {0.duration}
 
         def same_item(item, st):
             """Is the archived item the same as the filesystem item at the same path with stat st?"""
-            if not stat.S_ISREG(st.st_mode):
-                # we only "optimize" for regular files.
+            is_file = stat.S_ISREG(st.st_mode)
+            is_dir = stat.S_ISDIR(st.st_mode)
+            if not (is_file or is_dir):
+                # we only "optimize" for regular files and directories.
                 # other file types are less frequent and have no content extraction we could "optimize away".
                 return False
-            if item.mode != st.st_mode or item.size != st.st_size:
-                # the size check catches incomplete previous file extraction
+            if item.mode != st.st_mode:
+                # we want to extract a different type of file than what is present in the filesystem.
+                return False
+            if is_file and item.size != st.st_size:
+                # the size check catches incomplete previous regular file extraction
                 return False
             if item.get("mtime") != st.st_mtime_ns:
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
TODO: Maybe `same_item` could be extended to consider directories.

That would avoid any potential issues with calling restore_attrs on directories that already have metadata.

Also it would speed up continuing an extraction.

PR #9292, opened 2026-02-10
list --format: add fingerprint placeholder

ThomasWaldmann:
This allows users to compare file content efficiently without reading the full file data, by exposing a hash of the chunk IDs and the relevant conditions for valid comparisons, like chunker params, chunker seed/key, id key, key type, etc.
    
This is based on PR #5167 by @hrehfeld, code + discussion, with some changes:

- the conditions hash now includes more relevant input params
- returning a single value that is composed of 2 parts
- tests (including new buzhash64)
    
Example output (different files in same archive):

```
1e88bfb02d0a5320-a539587200c33b857f9827d01fcb7dabacf30501c83929e7308668d43f4a6302 file1
1e88bfb02d0a5320-9ed78a4c14d0506d9ae75d914cca90db64655ddea22647dd1c89f19e2fc080ae file2
```

The fingerprint has 2 parts:
    
First part: same hash, indicates same chunking / chunk id generation params, meaning that the second part is valid to be compared.
    
Second part: different hash, because file content is different. same hash here would mean same content.

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -15,6 +15,7 @@ from typing import ClassVar, Any, TYPE_CHECKING, Literal
 from collections import OrderedDict
 from datetime import datetime, timezone
 from functools import partial
+from hashlib import sha256
 from string import Formatter
 
 from ..logger import create_logger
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Example output (different files in same archive):

```
1e88bfb02d0a5320-a539587200c33b857f9827d01fcb7dabacf30501c83929e7308668d43f4a6302 file1
1e88bfb02d0a5320-9ed78a4c14d0506d9ae75d914cca90db64655ddea22647dd1c89f19e2fc080ae file2
```

First part: same hash, indicates same chunking / chunk id generation params, meaning that the second part is valid to be compared.
Second part: different hash, because file content is different. same hash here would mean same content.

PR #9294, opened 2026-02-11
testsuite: remove deprecated manual cleanup in create_cmd_test

ai-man-codes:
Manual cleanup is not needed anymore because of #9185 already being implemented.
So we can remove the #8980 code.

I have checked everything with the /tmp folder, the generated files are cleaned up safely after a test.


ThomasWaldmann:
Thanks!

PR #9314, opened 2026-02-13
CI: add omniOS

### LINKED ISSUE #9118: vagrant: openindiana VM hard to maintain
I have a openindiana vagrant VM (virtualbox) to test borg on it, but keeping it running is a pain:

Since a while, it does not install all the packages as given in the Vagrantfile (esp. not gcc-13 and git) giving a weird error msg (like that this is not allowed).

When I run `pkg update` (which is super slow and downloads a ton of stuff) **and** reboot, then I can install the required packages.

I tried to automate the reboot via vagrant, but:

```
Vagrant attempted to execute the capability 'reboot'
on the detect guest OS 'solaris11', but the guest doesn't
support that capability. This capability is required for your
configuration of Vagrant. Please either reconfigure Vagrant to
avoid this capability or fix the issue by creating the capability.
```

What I need is that it continues provisioning after the reboot, installing packages, running the tests, etc.

Please help.

ThomasWaldmann:
fixes #9118.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,6 +396,11 @@ jobs:
             display_name: OpenBSD
             do_binaries: false
 
+          - os: omnios
+            version: 'r151056'
+            display_name: OmniOS
+            do_binaries: false
+
           - os: haiku
             version: 'r1beta5'
             display_name: Haiku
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
libxxhash: https://github.com/omniosorg/illumos-omnios/issues/1647

PR #9336, opened 2026-02-15
Use jsonargparse

ThomasWaldmann:
WIP

### DIFF
--- a/src/borg/helpers/parseformat.py
+++ b/src/borg/helpers/parseformat.py
@@ -361,8 +361,6 @@ def SortBySpec(text):
     from ..manifest import AI_HUMAN_SORT_KEYS
 
     for token in text.split(","):
-        if token == "ts":
-            continue
         if token not in AI_HUMAN_SORT_KEYS:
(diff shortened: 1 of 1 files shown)


src/borg/archiver/__init__.py
ThomasWaldmann, line 270: somehow it does fail here on windows.

mauvilsa:
FIY, with pull request https://github.com/omni-us/jsonargparse/pull/845 now it is possible to add arguments with type and action. Not yet in a release, but the dependency could be set to install directly from the github main branch. I haven't yet looked at the pull request in detail.

ThomasWaldmann:
@mauvilsa Ah, that's great, thanks for the info.

That limitation was one of the major pain points in switching to jsonargparse.

ThomasWaldmann:
superceded by #9413

ThomasWaldmann:
@mauvilsa that type+action change was very helpful for borg, thanks.

doing #9413 was quite a bit easier than this first try.

PR #9341, opened 2026-02-15
windows: implement st_birthtime restoration, fixes #8730

### LINKED ISSUE #8730: windows: birthtime and ctime
From: https://docs.python.org/3.12/library/os.html

"""
Changed in version 3.12: st_birthtime[_ns] is now available on Windows.
"""

"""
Changed in version 3.12: The st_ctime attribute of a stat result is deprecated on Windows. The file creation time is properly available as st_birthtime, and in the future st_ctime may be changed to return zero or the metadata change time, if available.
"""

"""
Changed in version 3.12: st_ctime_ns is deprecated on Windows. Use st_birthtime_ns for the file creation time. In the future, st_ctime will contain the time of the most recent metadata change, as for other platforms.
"""

"""
Changed in version 3.12: On Windows, st_ctime is deprecated. Eventually, it will contain the last metadata change time, for consistency with other platforms, but for now still contains creation time. Use st_birthtime for the creation time.
"""

These are nice consistency changes with the goal of making windows timestamp semantics closer to posix OSes like linux.

trxvorr:
This PR implements the restoration of file creation time (`st_birthtime`) on Windows, addressing Issue #8730.

## Changes
*   **Platform Support**: Added [set_birthtime](cci:1://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/platform/__init__.py:81:0-125:83) in [src/borg/platform/__init__.py](cci:7://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/platform/__init__.py:0:0-0:0) using `ctypes` and `SetFileTime` to precisely set creation time on NTFS.
*   **Archiver**: Updated `Archive.restore_attrs` to call [set_birthtime](cci:1://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/platform/__init__.py:81:0-125:83) on Windows when birthtime metadata is present.
*   **Tests**:
    *   Added unit (`win32_birthtime_test.py`) and E2E (`win32_birthtime_e2e_test.py`) tests verifying birthtime preservation.
    *   Updated [is_birthtime_fully_supported](cci:1://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/testsuite/__init__.py:210:0-233:20) and fixed [test_nobirthtime](cci:1://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/testsuite/archiver/create_cmd_test.py:209:0-229:48) to validly test birthtime behavior on Windows (handling `st_birthtime` != `st_mtime` semantics).

## Verification
*   Verified on Windows with Python 3.12+.
*   New unit and E2E tests pass.
*   Existing [test_nobirthtime](cci:1://file:///C:/Users/trevo/Desktop/Projects/borg/src/borg/testsuite/archiver/create_cmd_test.py:209:0-229:48) now passes locally on Windows.

### DIFF
--- a/src/borg/platform/__init__.py
+++ b/src/borg/platform/__init__.py
@@ -21,6 +21,7 @@ if is_linux:  # pragma: linux only
     from .linux import listxattr, getxattr, setxattr
     from .linux import acl_get, acl_set
     from .linux import set_flags, get_flags
+    from .base import set_birthtime, set_timestamps
     from .linux import SyncFile
     from .posix import process_alive, local_pid_alive
     from .posix import get_errno
@@ -31,6 +32,7 @@ elif is_freebsd:  # pragma: freebsd only
     from .freebsd import acl_get, acl_set
     from .freebsd import set_flags
     from .base import get_flags
+    from .base import set_birthtime, set_timestamps
     from .base import SyncFile
     from .posix import process_alive, local_pid_alive
     from .posix import get_errno
@@ -40,6 +42,7 @@ elif is_netbsd:  # pragma: netbsd only
     from .netbsd import listxattr, getxattr, setxattr
     from .base import acl_get, acl_set
     from .base import set_flags, get_flags
+    from .base import set_birthtime, set_timestamps
     from .base import SyncFile
     from .posix import process_alive, local_pid_alive
     from .posix import get_errno
@@ -52,6 +55,7 @@ elif is_darwin:  # pragma: darwin only
     from .darwin import set_flags
     from .darwin import fdatasync, sync_dir  # type: ignore[no-redef]
     from .base import get_flags
+    from .base import set_birthtime, set_timestamps
     from .base import SyncFile
     from .posix import process_alive, local_pid_alive
     from .posix import get_errno
@@ -62,6 +66,7 @@ elif not is_win32:  # pragma: posix only
     from .base import listxattr, getxattr, setxattr
     from .base import acl_get, acl_set
     from .base import set_flags, get_flags
+    from .base import set_birthtime, set_timestamps
     from .base import SyncFile
     from .posix import process_alive, local_pid_alive
     from .posix import get_errno
@@ -72,6 +77,7 @@ else:  # pragma: win32 only
     from .base import listxattr, getxattr, setxattr
     from .base import acl_get, acl_set
     from .base import set_flags, get_flags
+    from .windows import set_birthtime, set_timestamps  # type: ignore[no-redef]
     from .base import SyncFile
     from .windows import process_alive, local_pid_alive
     from .windows import getosusername
--- a/src/borg/platform/base.py
+++ b/src/borg/platform/base.py
@@ -69,6 +69,38 @@ def setxattr(path, name, value, *, follow_symlinks=False):
     """
 
 
+def set_birthtime(path, birthtime_ns, *, fd=None):
+    """
+    Set creation time (birthtime) on *path* to *birthtime_ns*.
+    """
+    raise NotImplementedError("set_birthtime is not supported on this platform")
+
+
+def set_timestamps(path, item, fd=None, follow_symlinks=False):
+    """Set timestamps (mtime, atime, birthtime) from *item* on *path* (*fd*)."""
+    mtime = item.mtime
+    atime = item.atime if "atime" in item else mtime
+    if "birthtime" in item:
+        # This implementation uses the "utime trick" to set birthtime on BSDs and macOS.
+        # It sets mtime to the birthtime first, which pulls back birthtime, and then
+        # sets it to the final value. On other POSIX platforms, it's harmless.
+        birthtime = item.birthtime
+        try:
+            if fd:
+                os.utime(fd, ns=(atime, birthtime))
+            else:
+                os.utime(path, ns=(atime, birthtime), follow_symlinks=follow_symlinks)
+        except OSError:
+            pass
+    try:
+        if fd:
+            os.utime(fd, ns=(atime, mtime))
+        else:
+            os.utime(path, ns=(atime, mtime), follow_symlinks=follow_symlinks)
+    except OSError:
+        pass
+
+
 def acl_get(path, item, st, numeric_ids=False, fd=None):
     """
     Save ACL entries.
--- a/src/borg/platform/windows.pyx
+++ b/src/borg/platform/windows.pyx
@@ -1,3 +1,7 @@
+import ctypes
+from ctypes import wintypes
+import msvcrt
+
 import os
 import platform
 
@@ -11,6 +15,19 @@ cdef extern from 'windows.h':
     HANDLE OpenProcess(DWORD dwDesiredAccess, BOOL bInheritHandle, DWORD dbProcessId)
 
     cdef extern int PROCESS_QUERY_INFORMATION
+
+
+# Windows API Constants
+FILE_WRITE_ATTRIBUTES = 0x0100
+FILE_SHARE_READ = 0x00000001
+FILE_SHARE_WRITE = 0x00000002
+FILE_SHARE_DELETE = 0x00000004
+OPEN_EXISTING = 3
+FILE_FLAG_BACKUP_SEMANTICS = 0x02000000
+
+
+class FILETIME(ctypes.Structure):
+    _fields_ = [("dwLowDateTime", wintypes.DWORD), ("dwHighDateTime", wintypes.DWORD)]
 
 
 def getosusername():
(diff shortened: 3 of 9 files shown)


src/borg/platform/__init__.py
```diff
@@ -4,6 +4,8 @@
 Public APIs are documented in platform.base.
 """
 
+import ctypes
+from ctypes import wintypes
```
ThomasWaldmann: we have a platform_windows.py(x) - in the end, windows specific code should be only there.

src/borg/testsuite/archiver/create_cmd_test.py
```diff
@@ -211,19 +212,22 @@ def test_unix_socket(archivers, request, monkeypatch):
 def test_nobirthtime(archivers, request):
     archiver = request.getfixturevalue(archivers)
     create_test_files(archiver.input_path)
-    birthtime, mtime, atime = 946598400, 946684800, 946771200
-    os.utime("input/file1", (atime, birthtime))
-    os.utime("input/file1", (atime, mtime))
+    birthtime_ns, mtime_ns, atime_ns = 946598400 * 10**9, 946684800 * 10**9, 946771200 * 10**9
+    platform.set_birthtime("input/file1", birthtime_ns)
```
ThomasWaldmann: guess this is currently broken except on windows.

src/borg/platform/base.py
```diff
@@ -68,6 +69,12 @@ def setxattr(path, name, value, *, follow_symlinks=False):
     """
 
 
+def set_birthtime(path, birthtime_ns):
+    """
+    Set creation time (birthtime) on *path* to *birthtime_ns*.
+    """
```
ThomasWaldmann: this is a potential footgun.

it is available on all platforms, but does nothing except on windows.

maybe it should raise NotImplementedError.

trxvorr: agreed. 

i will update the base implementation to raise that error instead of passing silently.

src/borg/platform/windows.pyx
```diff
@@ -37,3 +37,51 @@ def process_alive(host, pid, thread):
 def local_pid_alive(pid):
     """Return whether *pid* is alive."""
     raise NotImplementedError
+
+
+def set_birthtime(path, birthtime_ns):
```
ThomasWaldmann: is there also a way to do this with an open file descriptor?

ThomasWaldmann: a lot of borg file functions take a file descriptor and a path. if the (open) file descriptor is provided, the function uses that and if not, it falls back to using the path.

src/borg/testsuite/archiver/create_cmd_test.py
```diff
@@ -211,19 +212,23 @@ def test_unix_socket(archivers, request, monkeypatch):
 def test_nobirthtime(archivers, request):
     archiver = request.getfixturevalue(archivers)
     create_test_files(archiver.input_path)
-    birthtime, mtime, atime = 946598400, 946684800, 946771200
-    os.utime("input/file1", (atime, birthtime))
-    os.utime("input/file1", (atime, mtime))
+    birthtime_ns, mtime_ns, atime_ns = 946598400 * 10**9, 946684800 * 10**9, 946771200 * 10**9
+    if sys.platform == "win32":
+        platform.set_birthtime("input/file1", birthtime_ns)
```
ThomasWaldmann: is_win32

also: don't kill it for other platforms.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -168,7 +168,7 @@ def test_birthtime(archivers, request):
     archiver = request.getfixturevalue(archivers)
     create_test_files(archiver.input_path)
     birthtime, mtime, atime = 946598400, 946684800, 946771200
-    os.utime("input/file1", (atime, birthtime))
+    set_birthtime("input/file1", birthtime * 1_000_000_000)  # noqa: F821
```
ThomasWaldmann: don't kill it for other platforms.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -177,7 +177,11 @@ def test_birthtime(archivers, request):
     sti = os.stat("input/file1")
     sto = os.stat("output/input/file1")
     assert same_ts_ns(sti.st_birthtime * 1e9, sto.st_birthtime * 1e9)
-    assert same_ts_ns(sto.st_birthtime * 1e9, birthtime * 1e9)
+    if is_win32:
+        # allow for small differences (e.g. 10ms)
+        assert abs(sto.st_birthtime * 1e9 - birthtime * 1e9) < 10_000_000
```
ThomasWaldmann: if we generally have to expect up to 10ms variance on windows, this could be moved into same_ts_ns.

src/borg/testsuite/archiver/win32_birthtime_e2e_test.py
```diff
@@ -0,0 +1,52 @@
...
+    # Verify it was set correctly initially
+    st_verify = os.stat(input_file)
+    assert get_birthtime_ns(st_verify, input_file) == old_birthtime_ns
+
+    # Archive it
+    cmd(archiver, "create", "test", "input")
+
+    # Extract it to a different location
+    if not os.path.exists("output"):
+        os.makedirs("output")
```
ThomasWaldmann: don't we always have the output dir?

also: check makedirs parameters.

src/borg/testsuite/archiver/win32_birthtime_test.py
```diff
@@ -0,0 +1,52 @@
+import os
```
ThomasWaldmann: in the end, this should be in the tests of the extract command.

src/borg/testsuite/__init__.py
```diff
@@ -21,7 +21,7 @@
     raises = None
 
 from ..fuse_impl import llfuse, has_any_fuse, has_llfuse, has_pyfuse3, has_mfusepy, ENOATTR  # NOQA
-from .. import platform
+from borg import platform as borg_platform
```
ThomasWaldmann: don't do absolute imports.

src/borg/testsuite/__init__.py
```diff
@@ -185,42 +185,51 @@ def are_fifos_supported():
...
         else:
             open(filepath, "w").close()
         try:
-            birthtime, mtime, atime = 946598400, 946684800, 946771200
-            os.utime(filepath, (atime, birthtime), follow_symlinks=False)
-            os.utime(filepath, (atime, mtime), follow_symlinks=False)
-            new_stats = os.stat(filepath, follow_symlinks=False)
-            if new_stats.st_birthtime == birthtime and new_stats.st_mtime == mtime and new_stats.st_atime == atime:
+            birthtime_ns, mtime_ns, atime_ns = 946598400 * 10**9, 946684800 * 10**9, 946771200 * 10**9
+            borg_platform.set_birthtime(filepath, birthtime_ns)
```
ThomasWaldmann: don't kill it for other platforms.

ThomasWaldmann:
Please update your local master branch and then rebase this branch onto that master and force push.

That should solve the CI test failures that are unrelated to your work.

src/borg/platform/windows.pyx
```diff
@@ -37,3 +37,55 @@ def process_alive(host, pid, thread):
...
+
+
+def set_birthtime(path, birthtime_ns, *, fd=None):
+    """
+    Set creation time (birthtime) on *path* (or *fd*) to *birthtime_ns*.
+    """
+    import ctypes
+    from ctypes import wintypes
+    import msvcrt
+
```
ThomasWaldmann: hmm, guess these are frequently needed on windows for all sorts of stuff - can we import them on module level?

src/borg/platform/windows.pyx
```diff
@@ -37,3 +37,55 @@ def process_alive(host, pid, thread):
...
+            str(path),
+            FILE_WRITE_ATTRIBUTES,
+            FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+            None,
+            OPEN_EXISTING,
+            FILE_FLAG_BACKUP_SEMANTICS,
+            None,
+        )
+        if handle == -1:
+            return
```
ThomasWaldmann: Looks like it silently fails here. Maybe we can rather raise an exception here - we can still decide on a higher level if we want to ignore it.

src/borg/platform/windows.pyx
```diff
@@ -37,3 +37,55 @@ def process_alive(host, pid, thread):
...
+
+    # Convert ns to Windows FILETIME
+    unix_epoch_in_100ns = 116444736000000000
+    intervals = (birthtime_ns // 100) + unix_epoch_in_100ns
+
+    ft = FILETIME()
+    ft.dwLowDateTime = intervals & 0xFFFFFFFF
+    ft.dwHighDateTime = intervals >> 32
+
+    close_handle = False
```
ThomasWaldmann: move this line to if-block. then you will set it in if-block as well as in else-block.

src/borg/testsuite/archiver/create_cmd_test.py
ThomasWaldmann: i guess this still only works on windows. set_birthtime is not implemented on other platforms and the first os.utime call that worked on other platforms is removed (doing that os.utime call twice is weird, but it seems to be the way it should work on posix).

src/borg/testsuite/archiver/create_cmd_test.py
```diff
@@ -211,19 +211,22 @@ def test_unix_socket(archivers, request, monkeypatch):
...
         cmd(archiver, "extract", "test")
     sti = os.stat("input/file1")
     sto = os.stat("output/input/file1")
-    assert same_ts_ns(sti.st_birthtime * 1e9, birthtime * 1e9)
-    assert same_ts_ns(sto.st_birthtime * 1e9, mtime * 1e9)
+    assert same_ts_ns(sti.st_birthtime * 1e9, birthtime_ns)
+    if is_win32:
+        assert not same_ts_ns(sto.st_birthtime * 1e9, birthtime_ns)
+    else:
+        assert same_ts_ns(sto.st_birthtime * 1e9, mtime_ns)
```
ThomasWaldmann: that matches the current reality of the code, but not how it should be, see previous comment.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -39,7 +38,7 @@
 def test_symlink_extract(archivers, request):
     archiver = request.getfixturevalue(archivers)
     create_test_files(archiver.input_path)
-    cmd(archiver, "repo-create", RK_ENCRYPTION)
```
ThomasWaldmann, line 42: i see the temptation to change this if it seems unneeded, but one should resist it and stay better focussed on the PR topic. rather open an issue to do a separate cleanup PR for this later.

src/borg/testsuite/archiver/extract_cmd_test.py
ThomasWaldmann: still a problem.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -825,3 +824,79 @@ def test_extract_existing_directory(archivers, request):
         cmd(archiver, "extract", "test")
         st2 = os.stat("input/dir")
     assert st1.st_ino == st2.st_ino
+
+
+@pytest.mark.skipif(not is_win32, reason="Windows only test")
+def test_set_birthtime(tmpdir):
```
ThomasWaldmann: maybe this and the next test could rather live in some platform testing module, it is not an "extract cmd" test.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -825,3 +824,79 @@ def test_extract_existing_directory(archivers, request):
...
+
+    set_birthtime(test_file, new_birthtime_ns)
+
+    st_new = os.stat(test_file)
+    restored_birthtime = get_birthtime_ns(st_new, test_file)
+    assert same_ts_ns(restored_birthtime, new_birthtime_ns)
+
+
+@pytest.mark.skipif(not is_win32, reason="Windows only test")
+def test_set_birthtime_fd(tmpdir):
```
ThomasWaldmann: btw, pytest also has a pathlib based tmp_path fixture, that gives you a Path object that is more conventient to work with.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -825,3 +824,79 @@ def test_extract_existing_directory(archivers, request):
...
+
+    with open(test_file, "r+") as f:
+        set_birthtime(test_file, new_birthtime_ns, fd=f.fileno())
+
+    st_new = os.stat(test_file)
+    restored_birthtime = get_birthtime_ns(st_new, test_file)
+    assert same_ts_ns(restored_birthtime, new_birthtime_ns)
+
+
+@pytest.mark.skipif(not is_birthtime_fully_supported() or not is_win32, reason="birthtime not supported or not windows")
```
ThomasWaldmann: if birthtime is fully supported on win32, should is_birthtime_fully_supported just return True there?

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -825,3 +824,79 @@ def test_extract_existing_directory(archivers, request):
...
+def test_birthtime_restore(archivers, request):
+    archiver = request.getfixturevalue(archivers)
+    cmd(archiver, "repo-create", RK_ENCRYPTION)
+    # Create a file in input directory
+    input_file = os.path.join(archiver.input_path, "test_file")
+    with open(input_file, "w") as f:
+        f.write("data")
+
```
ThomasWaldmann: iirc we have a convenience function to create an input file with some contents, check the existing tests.

src/borg/testsuite/archiver/extract_cmd_test.py
```diff
@@ -825,3 +824,79 @@ def test_extract_existing_directory(archivers, request):
...
+    st = os.stat(input_file)
+    original_birthtime = get_birthtime_ns(st, input_file)
+
+    # Set an old birthtime (10 years ago)
+    old_birthtime_ns = original_birthtime - 10 * 365 * 24 * 3600 * 10**9
+    # Ensure it's 100ns aligned (Windows precision)
+    old_birthtime_ns = (old_birthtime_ns // 100) * 100
```
ThomasWaldmann: hmm, if you query it first on windows, then only modify by a multiple of 100ns, shouldn't it always be aligned?

src/borg/testsuite/__init__.py
```diff
@@ -21,7 +21,7 @@
     raises = None
 
 from ..fuse_impl import llfuse, has_any_fuse, has_llfuse, has_pyfuse3, has_mfusepy, ENOATTR  # NOQA
-from .. import platform
+from .. import platform as borg_platform
```
ThomasWaldmann: maybe avoid renaming stuff if not absolutely required.

src/borg/testsuite/__init__.py
```diff
@@ -52,6 +52,9 @@ def same_ts_ns(ts_ns1, ts_ns2):
     """Compare two timestamps (both in nanoseconds) to determine whether they are (roughly) equal."""
     diff_ts = int(abs(ts_ns1 - ts_ns2))
     diff_max = 10 ** (-st_mtime_ns_round)
+    # On Windows, we generally have to expect up to 10ms variance.
+    if is_win32:
+        diff_max = max(diff_max, 10 * 1000 * 1000)
```
ThomasWaldmann: looks like that st_mtime_ns_round is some constant. check where it is defined/used and if it makes sense maybe just define the constant differently on win32.

src/borg/testsuite/archiver/create_cmd_test.py
trxvorr: @ThomasWaldmann we could potentially implement this using platform specific APIs (like setattrlist on Darwin) for mac and freebsd

ThomasWaldmann (commented):
some feedback

src/borg/testsuite/archiver/create_cmd_test.py
ThomasWaldmann: maybe a platform function that just gets an item, a path and a fd and sets all the timestamps it finds in there? check first whether that is a good way before starting to implement, look at the existing posix and win32 code.

ThomasWaldmann:
Check the CI, there are quite some failures.

ThomasWaldmann:
@trxvorr Before I review, please:
- rebase on current master branch and fix the conflicts
- try to fix the CI failures

ThomasWaldmann:
Same here, please rebase on current master, fix the conflicts, try to fix the CI failures.

ThomasWaldmann:
Superseded by #10121 and #10177.

PR #9348, opened 2026-02-15
archiver: warn about MSYS2 path translation, fixes #9339

### LINKED ISSUE #9339: MSYS2: /tmp paths resolve to C:/msys64/tmp/ in archives
in an MSYS2 environment, passing a POSIX-style path like `/tmp` as a source argument results in the path being expanded to its Windows equivalent (e.g., `C:/msys64/tmp/...`) before reaching Borg. This causes these absolute host-specific paths to be stored in the archive, which might be undesirable for cross-platform portability

trxvorr:
This PR addresses issue #9339 by adding a runtime warning when Borg detects it is running under an MSYS2/Git Bash environment without the necessary configuration to prevent path translation.

In MSYS2 environments, POSIX-style paths (e.g., /tmp) are automatically translated to Windows paths (e.g., C:/msys64/tmp) by the shell before being passed to the application. This causes unintended absolute Windows paths to be stored in archives, reducing portability.

Validated that the warning triggers when running in an unmodified MSYS2 shell.
Validated that setting the environment variables suppresses the warning.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -326,6 +326,29 @@ Install the dependencies with the provided script::
 
     ./scripts/msys2-install-deps
 
+.. _msys2_path_translation:
+
+MSYS2 Path Translation
+++++++++++++++++++++++
+
+When running Borg within an MSYS2 environment, the shell
+automatically translates POSIX-style paths (like ``/tmp`` or ``/C/Users``) to
+Windows paths (like ``C:\msys64\tmp`` or ``C:\Users``) before they reach the
+Borg process.
+
+This behavior can result in absolute Windows paths being stored in your backups,
+which may not be what you intended if you use POSIX paths for portability.
+
+To disable this automatic translation for Borg, you can use environment variables
+to exclude everything from conversion. Similarly, MSYS2 also translates
+environment variables that look like paths. To disable this generally for Borg,
+set both variables::
+
+    export MSYS2_ARG_CONV_EXCL="*"
+    export MSYS2_ENV_CONV_EXCL="*"
+
+For more details, see the `MSYS2 documentation on filesystem paths <https://www.msys2.org/docs/filesystem-paths/>`__.
+
 Windows 10's Linux Subsystem
 ++++++++++++++++++++++++++++
 
--- a/src/borg/archiver/__init__.py
+++ b/src/borg/archiver/__init__.py
@@ -46,6 +46,7 @@ try:
     from ..helpers import msgpack
     from ..helpers import sig_int
     from ..helpers import get_config_dir
+    from ..platformflags import is_msystem
     from ..remote import RemoteRepository
     from ..selftest import selftest
 except BaseException:
@@ -397,7 +398,16 @@ class Archiver(
         return functools.partial(self.do_maincommand_help, parser)
 
     def prerun_checks(self, logger, is_serve):
-
+        if (
+            not is_serve
+            and is_msystem
+            and ("MSYS2_ARG_CONV_EXCL" not in os.environ or "MSYS2_ENV_CONV_EXCL" not in os.environ)
+        ):
+            logger.warning(
+                "MSYS2 path translation is active. This can cause POSIX paths to be mangled into "
+                "Windows paths in archives. Consider setting MSYS2_ARG_CONV_EXCL='*' and "
+                "MSYS2_ENV_CONV_EXCL='*'. See https://www.msys2.org/docs/filesystem-paths/ for details."
+            )
         selftest(logger)
 
     def _setup_implied_logging(self, args):
--- a/src/borg/platformflags.py
+++ b/src/borg/platformflags.py
@@ -4,6 +4,7 @@ Flags for platform-specific APIs.
 Use these flags instead of sys.platform.startswith('<os>') or try/except.
 """
 
+import os
 import sys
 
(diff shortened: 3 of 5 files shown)


src/borg/archiver/__init__.py
```diff
@@ -455,7 +455,14 @@ def parse_args(self, args=None):
         return args
 
     def prerun_checks(self, logger, is_serve):
-
+        if not is_serve:
+            if sys.platform == "win32" and "MSYSTEM" in os.environ:
```
ThomasWaldmann: please use is_win32 for the first part.

maybe is makes sense to also add a is_msystem for the second part.

docs/installation.rst
```diff
@@ -328,6 +328,29 @@ Install the dependencies with the provided script::
 
     ./scripts/msys2-install-deps
 
+.. _msys2_path_translation:
+
+MSYS2 Path Translation
+++++++++++++++++++++++
+
+When running Borg within an MSYS2 (including Git Bash) environment, the shell
```
ThomasWaldmann: not sure why "Git Bash" is mentioned here?

docs/installation.rst
```diff
@@ -328,6 +328,29 @@ Install the dependencies with the provided script::
...
+automatically translates POSIX-style paths (like ``/tmp`` or ``/c/Users``) to
```
ThomasWaldmann: is msys2 normalizing to always lowercase "/c" ?

src/borg/platformflags.py
```diff
@@ -15,3 +16,6 @@
 is_openbsd = sys.platform.startswith("openbsd")
 is_darwin = sys.platform.startswith("darwin")
 is_haiku = sys.platform.startswith("haiku")
+
+# MSYS2/Git Bash (on Windows)
+is_msystem = "MSYSTEM" in os.environ
```
ThomasWaldmann: hmm, shall we add "is_win32 and ..." here?

ThomasWaldmann:
Please update your local master branch and then rebase this branch onto that master and force push.

That should solve the CI test failures that are unrelated to your work.

ThomasWaldmann:
btw, if you address feedback by changing the code, maybe give a short answer to the feedback and/or click on "resolve" if you think it is done.

src/borg/archiver/__init__.py
```diff
@@ -455,7 +456,14 @@ def parse_args(self, args=None):
         return args
 
     def prerun_checks(self, logger, is_serve):
-
+        if not is_serve:
+            if is_win32 and is_msystem:
+                if "MSYS2_ARG_CONV_EXCL" not in os.environ or "MSYS2_ENV_CONV_EXCL" not in os.environ:
```
ThomasWaldmann: as there is no "else" here, you could join the 3 if-conditions using "and".

ThomasWaldmann: also, since you are checking is_win32 already when is_msystem is assigned, you don't need it here.

ThomasWaldmann: @trxvorr did you see my feedback? would be good if this could be merged soon.

trxvorr: hey @ThomasWaldmann yes was planning on working on it today or tomorrow 

will push it soon, sorry for the delay

src/borg/testsuite/archiver/create_cmd_test.py
```diff
@@ -150,6 +151,28 @@ def test_archived_paths(archivers, request):
...
+    cmd(archiver, "create", "test", full_path)
+
+    archive_list = cmd(archiver, "list", "test", "--json-lines")
+    paths = [json.loads(line)["path"] for line in archive_list.splitlines() if line]
+
+    # Assert no path contains a Windows drive letter or backslashes
+    for p in paths:
+        assert ":" not in p
+        assert "\\" not in p
+
```
ThomasWaldmann: somehow this tests for something different than the effect described in #9339.

the main issue was stuff like C:/msys64/tmp/... showing up in archives.

some code recently added in borg might already translate this to c/msys64/tmp/... but it is still wrong if the user just gave /tmp as recursion root.

ThomasWaldmann (changes requested):

src/borg/testsuite/archiver/create_cmd_test.py
```diff
@@ -150,6 +151,25 @@ def test_archived_paths(archivers, request):
     assert expected_paths == sorted([json.loads(line)["path"] for line in archive_list.splitlines() if line])
 
 
+@pytest.mark.skipif(not is_msystem, reason="only for msystem")
```
ThomasWaldmann, line 154: ok, that just tests for the warning.

but what is missing is a test whether the feature actually works how it was intended.

in the docs you write that the problem is that one gives /tmp as recursion root to borg and then ends up with archived paths similar to C:\msys64\tmp in the archive.

there are some transformations in recent borg that might already get rid of the colon and transform the backslash, but you should check that the archived paths do not contain "msys64" and really start with "tmp/".

ThomasWaldmann (approved):
LGTM!

trxvorr:
@ThomasWaldmann are the codecov failures fine?

ThomasWaldmann:
Yes.

ThomasWaldmann:
Thanks for working on this!

PR #9359, opened 2026-02-17
improve dependabot configuration: more stability

ThomasWaldmann:


ThomasWaldmann:
@mr-raj12 see diff.

mr-raj12:
Looks better with these improvements. The cooldown settings make a lot of sense as no need to chase every new release immediately. And agreed on ignoring black, keeping that pinned avoids formatter inconsistencies across contributors.

PR #9366, opened 2026-02-17
benchmark crud: suppress compact warnings during benchmark runs, fixes #9365  

### LINKED ISSUE #9365: benchmark crud: suppress "borg compact" warnings during benchmark runs
During `borg benchmark crud`, each `do_delete()` call prints:

     Done. Run "borg compact" to free space.

 Since `measurement_run()` calls `do_delete()` three times per test sample, these warnings appear interleaved with the benchmark results, cluttering the output. For `--json-lines` mode, non-JSON warning lines get mixed into the JSON stream.

 Noted by @ThomasWaldmann in #9354:
 > Those lines with borg compact output are a bit unpretty, but fixing that is out of scope of this PR.

 **Affected code:**
 - `src/borg/archiver/benchmark_cmd.py` — `measurement_run()` calls `self.do_delete()` at lines 60, 73, 87
 - `src/borg/archiver/delete_cmd.py:55` — the warning source

mr-raj12:
## Summary
  - Suppress repeated `Done. Run "borg compact" to free space.` warnings that clutter `borg benchmark crud` output
  - `measurement_run()` calls `do_delete()` 3 times per test sample, each emitting the warning — 18 extra lines with the default 6 samples
  - For `--json-lines` mode, non-JSON warnings contaminate the JSON stream
  - Fix: temporarily raise the `borg.archiver` logger level to ERROR during measurement runs, restoring it afterward via try/finally
  - Single-file change in `benchmark_cmd.py`, no interface changes

  Closes #9365

  ## Checklist
  - [x] PR is against `master`
  - [x] Tests pass
  - [x] Locally verified: `pytest benchmark_cmd_test.py` (4/4 passed) and `borg benchmark crud --json-lines` produces clean JSON-only output

### DIFF
--- a/src/borg/archiver/benchmark_cmd.py
+++ b/src/borg/archiver/benchmark_cmd.py
@@ -2,6 +2,7 @@ import argparse
 from contextlib import contextmanager
 import functools
 import json
+import logging
 import os
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


src/borg/archiver/benchmark_cmd.py
```diff
@@ -29,67 +30,77 @@ def parse_args(args, cmd):
...
-                        args,
-                        [
-                            f"--repo={repo}",
-                            "create",
-                            compression,
-                            "--files-cache=disabled",
-                            "borg-benchmark-crud1",
-                            path,
-                        ],
+            # Suppress "Done. Run borg compact..." warnings from internal do_delete() calls —
```
ThomasWaldmann, line 33: Just wondered whether this is a warning or an info.

Borg indeed emits this as warning.

Guess we made it that way for better visibility, because some people forgot to compact.

ThomasWaldmann:
Can you please check whether we have the same issue in 1.4-maint?

src/borg/archiver/benchmark_cmd.py
mr-raj12: Yes, it's emitted as a warning via self.print_warning() in delete_cmd.py:55 , makes sense for normal usage to remind users to compact. For the benchmark though, the repo is temporary and all return codes are already asserted, so suppressing warnings during measurement_run() seemed like the cleanest approach without changing do_delete's interface.

mr-raj12:
Checked 1.4-maint, this issue doesn't exist there. On 1.4-maint, _delete_archives() only emits a warning ("Run borg check --repair") in the forced==2 path. The standard delete path (which benchmark crud uses, since it doesn't pass --force) has no such warning at all. Also, 1.4-maint doesn't have --json-lines for benchmark crud, so the JSON contamination angle doesn't apply either.

PR #9374, opened 2026-02-19
21 feat create diy for check

gabo26126:
## Checklist

- [ ] PR is against `master` (or maintenance branch if only applicable there)


ThomasWaldmann:
@gabo26126 Hmm, why would we want that "diy" coverage testing if we already have the pytest-cov / codecov based testing?

It seems rather intrusive to have to manage that everywhere.

ThomasWaldmann:
ping?

PR #9381, opened 2026-02-20
remote: fix StoreObjectNotFound exception lost over RPC, fixes #9380

### LINKED ISSUE #9380: remote: RemoteRepository raises Repository.ObjectNotFound instead of StoreObjectNotFound
When `store_load()` / `store_delete()` are called on a local `Repository`, they propagate `borgstore.store.ObjectNotFound` (aliased as `StoreObjectNotFound`) directly. But when called via `RemoteRepository`, the RPC layer serializes the exception using `e.__class__.__name__` which is `"ObjectNotFound"` for both `StoreObjectNotFound` and `Repository.ObjectNotFound`.

  The client-side `handle_error` sees `"ObjectNotFound"` and always reconstructs `Repository.ObjectNotFound` — the wrong exception type.

  This is already acknowledged in `cache.py` with TODO comments at 3 call sites:
  ```python
  except (Repository.ObjectNotFound, StoreObjectNotFound):
      # TODO: ^ seem like RemoteRepository raises Repository.ONF instead of StoreONF
```

  Root cause:

  - Server side (remote.py:276): e.__class__.__name__ is "ObjectNotFound" for both borgstore.store.ObjectNotFound and Repository.ObjectNotFound — they share the same __name__ despite being unrelated classes.
  - Client side (remote.py:791): handle_error maps "ObjectNotFound" → Repository.ObjectNotFound unconditionally.

  Proposed fix:

  - Server side: detect StoreObjectNotFound with isinstance() and serialize it as "StoreObjectNotFound" instead of the ambiguous "ObjectNotFound".
  - Client side: add a "StoreObjectNotFound" branch in handle_error to reconstruct it correctly.
  - cache.py: simplify the 3 except clauses to just except StoreObjectNotFound: and remove the TODO comments.

  Affected code:
  - src/borg/remote.py:38 — import
  - src/borg/remote.py:276,288 — server-side serialization
  - src/borg/remote.py:791 — client-side reconstruction
  - src/borg/cache.py:715,772,785 — workaround catch clauses with TODO comments

### PR DESCRIPTION
**Pull request by mr-raj12**
## Summary
  - `RemoteRepository` RPC layer serializes exceptions using `e.__class__.__name__`, but both `borgstore.store.ObjectNotFound` and `Repository.ObjectNotFound` share `__name__ == "ObjectNotFound"`
  - Client-side `handle_error` always reconstructs `Repository.ObjectNotFound`, even when the server actually raised `StoreObjectNotFound` from a `store_load()`/`store_delete()` call
  - This was already acknowledged with TODO comments at 3 sites in `cache.py`: `# TODO: ^ seem like RemoteRepository raises Repository.ONF instead of StoreONF`
  - Fix: distinguish `StoreObjectNotFound` during server-side serialization and reconstruct it correctly on the client side

  Fixes #9380

  ## Changes

  | File | Change |
  |------|--------|
  | `src/borg/remote.py` | Import `StoreObjectNotFound`; serialize it as `"StoreObjectNotFound"` (not ambiguous `"ObjectNotFound"`) on the server side; add client-side `handle_error` branch to reconstruct it; add `inject_exception` test hook |
  | `src/borg/cache.py` | Simplify 3 `except` clauses from `(Repository.ObjectNotFound, StoreObjectNotFound)` to `StoreObjectNotFound`; remove TODO comments |  
  | `src/borg/testsuite/repository_test.py` | Add `StoreObjectNotFound` round-trip test in `test_remote_rpc_exception_transport` |

  ## Backward compatibility note

  This changes the RPC wire format for `StoreObjectNotFound`:
  - **Old server → new client**: old server sends `"ObjectNotFound"`, new client raises `Repository.ObjectNotFound` - `cache.py` callers would miss it. In practice, `store_*` methods are borg2-only and borg2 is in beta, so mixed-version deployments are not a supported scenario.
  - **New server → old client**: old client falls through to `RPCError` fallback.

  ## Checklist
  - [x] PR is against `master`
  - [x] New code has tests
  - [x] Tests pass
  - [x] Code formatted with black

### DIFF
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -35,7 +35,7 @@ from .logger import create_logger, borg_serve_log_queue
 from .manifest import NoManifestError
 from .helpers import msgpack
 from .legacyrepository import LegacyRepository
-from .repository import Repository
+from .repository import Repository, StoreObjectNotFound
 from .version import parse_version, format_version
 from .checksums import xxh64
 from .helpers.datastruct import EfficientCollectionQueue
@@ -269,11 +269,17 @@ class RepositoryServer:  # pragma: no cover
                                 logging.debug("\n".join(ex_full))
 
                             sys_info = sysinfo()
+                            # StoreObjectNotFound and Repository.ObjectNotFound both have
+                            # __name__ == "ObjectNotFound", so we need to distinguish them
+                            # explicitly for correct client-side reconstruction.
+                            exc_cls_name = (
+                                "StoreObjectNotFound" if isinstance(e, StoreObjectNotFound) else e.__class__.__name__
+                            )
                             try:
                                 msg = msgpack.packb(
                                     {
                                         MSGID: msgid,
-                                        "exception_class": e.__class__.__name__,
+                                        "exception_class": exc_cls_name,
                                         "exception_args": e.args,
                                         "exception_full": ex_full,
                                         "exception_short": ex_short,
@@ -285,7 +291,7 @@ class RepositoryServer:  # pragma: no cover
                                 msg = msgpack.packb(
                                     {
                                         MSGID: msgid,
-                                        "exception_class": e.__class__.__name__,
+                                        "exception_class": exc_cls_name,
                                         "exception_args": [
                                             x if isinstance(x, (str, bytes, int)) else None for x in e.args
                                         ],
@@ -403,6 +409,8 @@ class RepositoryServer:  # pragma: no cover
             raise PathNotAllowed("foo")
         elif kind == "ObjectNotFound":
             raise self.RepoCls.ObjectNotFound(s1, s2)
+        elif kind == "StoreObjectNotFound":
+            raise StoreObjectNotFound(s1)
         elif kind == "InvalidRPCMethod":
             raise InvalidRPCMethod(s1)
         elif kind == "divide":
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/borg/remote.py [hunk 1]
```diff
@@ -269,11 +269,17 @@ def inner_serve():
                                 logging.debug("\n".join(ex_full))
 
                             sys_info = sysinfo()
+                            # StoreObjectNotFound and Repository.ObjectNotFound both have
+                            # __name__ == "ObjectNotFound", so we need to distinguish them
+                            # explicitly for correct client-side reconstruction.
```
ThomasWaldmann, line 274: Good analysis.

src/borg/remote.py
```diff
@@ -784,6 +792,8 @@ def handle_error(unpacked):
                 raise Repository.ParentPathDoesNotExist(args[0])
             elif error == "ObjectNotFound":
                 raise Repository.ObjectNotFound(args[0], self.location.processed)
+            elif error == "StoreObjectNotFound":
+                raise StoreObjectNotFound(*args)
```
ThomasWaldmann: maybe double check how many args this exception class shall get.

src/borg/remote.py
mr-raj12: Thank you

src/borg/remote.py
mr-raj12: I found that borgstore's ObjectNotFound inherits from Exception with no custom __init__, and it's always raised with a single arg (the key name). So, changed to StoreObjectNotFound(args[0]) to match how all other exceptions in handle_error are reconstructed.

ThomasWaldmann:
Thanks!

PR #9408, opened 2026-02-23
test: repository path

SebastianSpeitel:
## Description

I found a bug, where a repository with a path containing `@` can't be used.
This adds a test which reproduces the issue.

## Checklist

### DIFF
--- a/src/borg/testsuite/repository_test.py
+++ b/src/borg/testsuite/repository_test.py
@@ -95,6 +95,12 @@ def test_basic_operations(repo_fixtures, request):
                 continue
             assert pdchunk(repository.get(H(x))) == b"SOMEDATA"
 
+def test_repository_path(tmp_path):
+    location = os.fspath(tmp_path / "@repo")
+    repository = Repository(location)
+    assert repository.url.endswith("@repo")
+    
+    repository.open(exclusive=False)
 
 def test_read_data(repo_fixtures, request):
     with get_repository_from_fixture(repo_fixtures, request) as repository:


ThomasWaldmann:
In the failing test, the repo url looks like:
```
file:///tmp/pytest-of-runner/pytest-0/test_repository_path0/%40repo
```

We have an issue discussing about urlquoting: borgbackup/borgstore#132 

From Gemini:

Handling the @ Character

According to RFC 3986, the @ character is a "sub-delimiter" and is technically allowed to appear unencoded within a path segment.

However, in many systems (especially in Python or when passing paths to CLI tools), you can safely use the percent-encoded version to avoid any ambiguity with the "userinfo" part of a URL (e.g., user@host).

SebastianSpeitel:
Even if the @ is technically allowed, there are other characters which would be encoded.
In my usecase the path is no URL at all (local repository), so there is no reason to think an @ would be problematic since it's a valid character in paths.

The two solutions I see:
- staying with standard compliant URLs but making sure everything gets correctly encoded *and* decided. But what happens, when the user passes a already encoded location intentionally?
- using a custom format or structure to represent locations. (I saw there is a `Location` type already)

ThomasWaldmann:
https://github.com/borgbackup/borgstore/pull/131

borg always uses `Path(p).absolute().as_uri()` if the user gave a fs path p.

If the user gives a URL, it is just used "as is".

ThomasWaldmann:
I am closing this, the test is wrong for how borg and borgstore work together (via an URL, that contains the path URL-encoded). See the correct fix in my previous post.

PR #9413, opened 2026-02-25
use jsonargparse

### LINKED ISSUE #6551: argparse++
Have a look at:

- https://jsonargparse.readthedocs.io/en/stable/
- https://github.com/bw2/ConfigArgParse
- https://pypi.org/project/confuse/
- https://github.com/tiangolo/typer and https://palletsprojects.com/p/click/
  - see also #251
- https://github.com/omry/omegaconf
- https://github.com/google/python-fire

ThomasWaldmann:
## Description

Adopt jsonargparse, fixes #6551.

## Checklist

- [X] New code has tests and docs where appropriate

### DIFF
--- a/docs/usage/create.rst
+++ b/docs/usage/create.rst
@@ -89,6 +89,9 @@ Examples
     $ find ~ -size -1000k | borg create --paths-from-stdin small-files-only
     # Use --paths-from-command with find to back up files from only a given user
     $ borg create --paths-from-command joes-files -- find /srv/samba/shared -user joe
+    # Use --paths-from-shell-command with find to back up a few files from only a given user -
+    # BE VERY CAREFUL AND ONLY USE TRUSTED INPUT FOR THE SHELL COMMAND!
+    $ borg create --paths-from-shell-command some-of-joes-files -- "find /srv/samba/shared -user joe | head"
     # Use --paths-from-stdin with --paths-delimiter (for example, for filenames with newlines in them)
     $ find ~ -size -1000k -print0 | borg create \
         --paths-from-stdin \
--- a/src/borg/archiver/create_cmd.py
+++ b/src/borg/archiver/create_cmd.py
@@ -91,13 +91,24 @@ class CreateMixIn:
                 else:
                     status = "+"  # included
                 self.print_file_status(status, path)
-            elif args.paths_from_command or args.paths_from_stdin:
+            elif args.paths_from_command or args.paths_from_shell_command or args.paths_from_stdin:
                 paths_sep = eval_escapes(args.paths_delimiter) if args.paths_delimiter is not None else "\n"
-                if args.paths_from_command:
+                if args.paths_from_command or args.paths_from_shell_command:
                     try:
                         env = prepare_subprocess_env(system=True)
-                        proc = subprocess.Popen(  # nosec B603
-                            args.paths, stdout=subprocess.PIPE, env=env, preexec_fn=None if is_win32 else ignore_sigint
+                        if args.paths_from_shell_command:
+                            # Use shell=True to support pipes, redirection, etc.
+                            shell = True
+                            cmd = " ".join(args.paths)
+                        else:
+                            shell = False
+                            cmd = args.paths
+                        proc = subprocess.Popen(
+                            cmd,
+                            stdout=subprocess.PIPE,
+                            env=env,
+                            shell=shell,  # nosec B602
+                            preexec_fn=None if is_win32 else ignore_sigint,
                         )
                     except (FileNotFoundError, PermissionError) as e:
                         raise CommandError(f"Failed to execute command: {e}")
@@ -131,7 +142,7 @@ class CreateMixIn:
                     self.print_file_status(status, path)
                     if not dry_run and status is not None:
                         fso.stats.files_stats[status] += 1
-                if args.paths_from_command:
+                if args.paths_from_command or args.paths_from_shell_command:
                     rc = proc.wait()
                     if rc != 0:
                         raise CommandError(f"Command {args.paths[0]!r} exited with status {rc}")
@@ -762,8 +773,8 @@ class CreateMixIn:
 
         If you need more control and you want to give every single fs object path
         to borg (maybe implementing your own recursion or your own rules), you can use
-        ``--paths-from-stdin`` or ``--paths-from-command`` (with the latter, borg will
-        fail to create an archive should the command fail).
+        ``--paths-from-stdin``, ``--paths-from-command`` or ``--paths-from-shell-command``
+        (with the latter two, borg will fail to create an archive should the command fail).
 
         Borg supports paths with the slashdot hack to strip path prefixes here also.
         So, be careful not to unintentionally trigger that.
@@ -841,6 +852,11 @@ class CreateMixIn:
             "--paths-from-command",
             action="store_true",
             help="interpret PATH as command and treat its output as ``--paths-from-stdin``",
+        )
+        subparser.add_argument(
+            "--paths-from-shell-command",
+            action="store_true",
+            help="interpret PATH as shell command and treat its output as ``--paths-from-stdin``",
         )
         subparser.add_argument(
             "--paths-delimiter",
--- a/src/borg/testsuite/archiver/create_cmd_test.py
+++ b/src/borg/testsuite/archiver/create_cmd_test.py
@@ -418,6 +418,21 @@ def test_create_paths_from_command_missing_command(archivers, request):
     assert output.endswith("No command given." + os.linesep)
 
 
+@pytest.mark.skipif(is_win32, reason="shell patterns not supported on Windows")
+def test_create_paths_from_shell_command(archivers, request):
+    archiver = request.getfixturevalue(archivers)
+    cmd(archiver, "repo-create", RK_ENCRYPTION)
+    create_regular_file(archiver.input_path, "file1", size=1024 * 80)
+    create_regular_file(archiver.input_path, "file2", size=1024 * 80)
+    create_regular_file(archiver.input_path, "file3", size=1024 * 80)
+    input_data = "input/file1\ninput/file2\ninput/file3"
+    # Use a shell pipe to test that shell=True works correctly.
+    cmd(archiver, "create", "--paths-from-shell-command", "test", "--", f"echo '{input_data}' | head -n 2")
+    archive_list = cmd(archiver, "list", "test", "--json-lines")
+    paths = [json.loads(line)["path"] for line in archive_list.split("\n") if line]
+    assert paths == ["input/file1", "input/file2"]
+
+
 def test_create_without_root(archivers, request):
     """test create without a root"""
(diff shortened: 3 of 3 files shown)


src/borg/archiver/_common.py
```diff
@@ -268,13 +269,16 @@ def process_epilog(epilog):
 
 
 def define_exclude_and_patterns(add_option, *, tag_files=False, strip_components=False):
+    add_option("--pattern-roots-internal", dest="pattern_roots", action="append", default=[], help=argparse.SUPPRESS)
+    add_option("--patterns-internal", dest="patterns", action="append", default=[], help=argparse.SUPPRESS)
```
ThomasWaldmann: that's a bit annoying to have, it spams our usage docs option table.

ThomasWaldmann: i modified build_usage/build_man, so they exclude suppressed options.

ThomasWaldmann:
@mauvilsa can you have a look, please?

maybe look at the individual commits.

is that `flatten_namespace` code in `borg.helpers.argparsing` interesting for other `jsonargparse` users?

had to import some internal stuff to implement `borg completion` command, guess there is currently no better way?

guess this can soon be merged - it does not use most features of jsonargparse, but that should be in future PRs.

scripts/make.py
```diff
@@ -46,7 +47,7 @@ def generate_level(self, prefix, parser, Archiver, extra_choices=None):
         is_subcommand = False
         choices = {}
         for action in parser._actions:
-            if action.choices is not None and "SubParsersAction" in str(action.__class__):
+            if action.choices is not None and "SubCommands" in str(action.__class__):
```
mauvilsa, line 50: Since this script should be argparse, isn't this and other changes in this file wrong?

src/borg/archiver/debug_cmd.py
```diff
@@ -339,49 +335,43 @@ def build_parser_debug(self, subparsers, common_parser, mid_common_parser):
         already appended at the end of the traceback.
         """
         )
-        subparser = debug_parsers.add_parser(
-            "info",
-            parents=[common_parser],
+        subparser = ArgumentParser(
+            parents=[mid_common_parser],
```
mauvilsa: I think there aren't tests in jsonargparse related to parent parsers. Could be that it works just fine. But I should take note of this to have a closer look.

src/borg/helpers/argparsing.py
```diff
@@ -0,0 +1,44 @@
...
+from argparse import Action, ArgumentError, ArgumentTypeError, RawDescriptionHelpFormatter  # noqa: F401
+from jsonargparse import ArgumentParser, Namespace, SUPPRESS, REMAINDER  # noqa: F401
+
+# borg completion uses these private symbols, so we need to import them:
+from jsonargparse._actions import _ActionSubCommands  # noqa: F401
+from jsonargparse._completions import prepare_actions_context, shtab_prepare_actions  # noqa: F401
+from jsonargparse._completions import bash_compgen_typehint  # noqa: F401
+
+
+def flatten_namespace(ns: Any) -> Namespace:
```
mauvilsa, line 113: jsonargparse namespaces have [as_flat()](https://jsonargparse.readthedocs.io/en/v4.46.0/#jsonargparse.Namespace.as_flat). This function could become very short using that. Or maybe removed if there is refactoring elsewhere?

src/borg/helpers/argparsing.py
```diff
@@ -0,0 +1,44 @@
...
+# here are the only imports from argparse and jsonargparse,
+# all other imports of these names import them from here:
+from argparse import Action, ArgumentError, ArgumentTypeError, RawDescriptionHelpFormatter  # noqa: F401
+from jsonargparse import ArgumentParser, Namespace, SUPPRESS, REMAINDER  # noqa: F401
+
+# borg completion uses these private symbols, so we need to import them:
+from jsonargparse._actions import _ActionSubCommands  # noqa: F401
+from jsonargparse._completions import prepare_actions_context, shtab_prepare_actions  # noqa: F401
+from jsonargparse._completions import bash_compgen_typehint  # noqa: F401
+
```
mauvilsa, line 105: I see `formatter_class=RawDescriptionHelpFormatter` in many places. Maybe better to create a subclass here to avoid having to set the formatter every time? Something like:

```suggestion

class RawDescriptionArgumentParser(ArgumentParser):
    def __init__(*args, formatter_class=RawDescriptionHelpFormatter, **kwargs):
        super().__init__(*args, formatter_class=formatter_class, **kwargs)
```

src/borg/helpers/argparsing.py
```diff
@@ -0,0 +1,44 @@
...
+# here are the only imports from argparse and jsonargparse,
+# all other imports of these names import them from here:
+from argparse import Action, ArgumentError, ArgumentTypeError, RawDescriptionHelpFormatter  # noqa: F401
+from jsonargparse import ArgumentParser, Namespace, SUPPRESS, REMAINDER  # noqa: F401
+
+# borg completion uses these private symbols, so we need to import them:
+from jsonargparse._actions import _ActionSubCommands  # noqa: F401
+from jsonargparse._completions import prepare_actions_context, shtab_prepare_actions  # noqa: F401
+from jsonargparse._completions import bash_compgen_typehint  # noqa: F401
```
mauvilsa: I will look at this in more detail. Eventually there shouldn't be private imports. But I need to see if there is already an alternative, or a modification in jsonargparse is required.

scripts/make.py
ThomasWaldmann: The make script itself imports from argparse, so it does not fail with a fresh borg checkout and somebody running "make.py clean ; pip install -e .".

But for the make subcommands that introspect into the borg argparser (e.g. build_man, build_usage), the borg argparsing code is called and that uses jsonargparse now.

src/borg/archiver/debug_cmd.py
ThomasWaldmann: it seems to work, at least neither our automated tests nor me practically trying it noticed an issue yet.

src/borg/helpers/argparsing.py
ThomasWaldmann: good idea, will do that!

src/borg/helpers/argparsing.py
ThomasWaldmann: yeah, much easier with that, will update the PR soon...

ThomasWaldmann:
(rebased on current master, added auto env var support)

ThomasWaldmann:
Hmm, I can't get "store_true" stuff working correctly:
```
subparser.add_argument("-s", "--stats", dest="stats", action="store_true", help="print stats")
```
With that and without the value being set in the default config, overriding via `BORG_CREATE__STATS=true` crashes the code, because it puts a str into args.stat instead of the expected bool.

If I set the value in the default config, it works, but I can not overwrite it with the env var.

A hack with ActionYesNo also didn't work, because it does not like short options like `-s`. https://github.com/omni-us/jsonargparse/issues/355#issuecomment-2600901284

Another weird hack: https://github.com/omni-us/jsonargparse/issues/497 - is that the intended way?

mauvilsa:
> Hmm, I can't get "store_true" stuff working correctly:

This I would consider this a bug. I will fix it.

src/borg/archiver/completion_cmd.py
```diff
@@ -62,12 +60,15 @@
     FilesCacheMode,
     PathSpec,
     ChunkerParams,
+    CompressionSpec,
     tag_validator,
     relative_time_marker_validator,
     parse_file_size,
 )
+from ..helpers.argparsing import ArgumentParser
+from ..helpers.argparsing import _ActionSubCommands
```
mauvilsa: Some time ago I noticed that `add_subcommand()` does not show up in the [API reference](https://jsonargparse.readthedocs.io/en/v4.46.0/#api-reference). I will make this class public to fix this and so that it can be used like here.

src/borg/archiver/completion_cmd.py
```diff
@@ -732,8 +731,20 @@ def do_completion(self, args):
...
+        if args.shell == "bash":
+            preambles.append(bash_compgen_typehint.strip().replace("%s", prog))
+            preambles.append(bash_preamble)
+        elif args.shell == "zsh":
+            preambles.append(zsh_preamble)
+
+        with prepare_actions_context(args.shell, prog, preambles):
+            shtab_prepare_actions(parser)
+
+        script = shtab.complete(parser, shell=args.shell, preamble="\n".join(preambles))  # nosec B604
```
mauvilsa: For this I am thinking that a method is needed to programmatically get a completion script. The code would change as follows:

```suggestion
        preambles = []
        if args.shell == "bash":
            preambles.append(bash_preamble)
        elif args.shell == "zsh":
            preambles.append(zsh_preamble)

        script = parser.get_completions_script(f"shtab-{args.shell}", preambles=preambles)
```

The method would not be shtab specific so that it can be used in potential future extensions, e.g. https://github.com/omni-us/jsonargparse/issues/618#issuecomment-3239836531

mauvilsa: @ThomasWaldmann please have a look at https://github.com/omni-us/jsonargparse/pull/859

ThomasWaldmann: Left a comment there, thanks!

ThomasWaldmann: oh, i overlooked the preambles argument, need to try that...

ThomasWaldmann: much better with that, "just works".

mauvilsa: This is public API, so I will not rush to merge just yet. I will leave some of days in case we notice that it makes sense to change in some way.

src/borg/helpers/argparsing.py
```diff
@@ -0,0 +1,156 @@
...
+# here are the only imports from argparse and jsonargparse,
+# all other imports of these names import them from here:
+from argparse import Action, ArgumentError, ArgumentTypeError, RawDescriptionHelpFormatter  # noqa: F401
+from jsonargparse import ArgumentParser as _ArgumentParser  # we subclass that to add custom behavior
+from jsonargparse import Namespace, SUPPRESS, REMAINDER  # noqa: F401
+from jsonargparse.typing import register_type  # noqa: F401
+from jsonargparse.typing import PositiveInt  # noqa: F401
+
+# borg completion uses these private symbols, so we need to import them:
+from jsonargparse._actions import _ActionSubCommands  # noqa: F401
```
mauvilsa: With https://github.com/omni-us/jsonargparse/pull/866 now you can do `from jsonargparse import ActionSubCommands`.

ThomasWaldmann: Thanks!

ThomasWaldmann:
TODO before merging this:
- Wait until that PR is merged into main branch there: https://github.com/omni-us/jsonargparse/pull/859
- Change pyproject.toml to pull main branch (and later from pypi).
- rebase on current master, collapse some commits

PR #9419, opened 2026-02-28
fix: add --cov flags to Windows CI pytest command, fixes #9418

### LINKED ISSUE #9418: ci: Windows CI does not collect coverage, breaks Codecov patch check for Windows-only code
## Problem
The Windows CI job in `.github/workflows/ci.yml` ([line 662](https://github.com/borgbackup/borg/blob/master/.github/workflows/ci.yml#L662)) runs pytest without `--cov`:

      python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --junitxml=test-results.xml

  The Linux/macOS jobs use `tox`, which includes `--cov=borg --cov-config=pyproject.toml`, so coverage is generated and uploaded to Codecov. The Windows job skips coverage entirely, even though `pytest-cov` is already installed via `scripts/msys2-install-deps`.

  This means any PR adding Windows-only source code (e.g., `platform/windows.pyx`) will always fail the `codecov/patch` check with 0% coverage, regardless of test quality. PR #9389 is currently affected with all CI checks green, tests passing, but Codecov reports "0.00% of diff hit (target 76.47%)".

   | CI job | Test runner | Coverage collected? |
   |--------|------------|-------------------|
   | Linux/macOS (`native_tests`) | `tox` with `--cov=borg` | Yes |
   | Windows (`windows_tests`) | `pytest` directly | **No** |

  ## Proposed fix

  Add `--cov=borg --cov-config=pyproject.toml` to the Windows pytest command, matching what tox uses for Linux/macOS:    

  ```yaml
  # before
  python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --junitxml=test-results.xml

  # after
  python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --junitxml=test-results.xml
```
  - pytest-cov is already available — scripts/msys2-install-deps installs mingw-w64-ucrt-x86_64-python-pytest-cov        
  - The coverage config in pyproject.toml ([tool.coverage.run]) is platform-agnostic and works as-is
  - One-line change in .github/workflows/ci.yml

### PR DESCRIPTION
**Pull request by mr-raj12**
## Description

  The Windows CI job (`windows_tests`) was running pytest without `--cov`, so Codecov received no coverage data from Windows. Any PR touching Windows-only code would fail `codecov/patch` with 0%.

  Added `--cov=borg --cov-config=pyproject.toml` to the Windows pytest command, matching the flags already used by the Linux/tox-based CI jobs.

  Fixes #9418

  ## Checklist

  - [X] PR is against `master` 
  - [X] Tests pass 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -659,7 +659,7 @@ jobs:
           export PATH="$GITHUB_WORKSPACE/dist/binary/borg-dir:$PATH"
           borg.exe -V
           . env/bin/activate
-          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --cov-report=xml:coverage.xml --junitxml=test-results.xml
+          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --junitxml=test-results.xml
 
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
@@ -682,6 +682,4 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           report_type: coverage
-          files: coverage.xml
-          flags: windows
           env_vars: OS,python

### CODE REVIEW

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -659,7 +659,7 @@ jobs:
           export PATH="$GITHUB_WORKSPACE/dist/binary/borg-dir:$PATH"
           borg.exe -V
           . env/bin/activate
-          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --junitxml=test-results.xml
+          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --junitxml=test-results.xml
```
ThomasWaldmann: good catch.

but why does it say the test coverage of the project did change by 0%? shouldn't we see an increase due to this change?

ThomasWaldmann: https://app.codecov.io/gh/borgbackup/borg/pull/9419/blob/src/borg/platform/__init__.py?dropdown=coverage

does not show coverage for win32. also not for some others.

mr-raj12: trying few tweaks before final squash and force push

ThomasWaldmann: now it looks better for win32.

.github/workflows/ci.yml
```diff
@@ -674,11 +674,14 @@ jobs:
...
+        if: ${{ !cancelled() }}
         uses: codecov/codecov-action@v5
         env:
           OS: ${{ runner.os }}
           python: '3.11'
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           report_type: coverage
+          files: coverage.xml
+          flags: windows
```
ThomasWaldmann: why are these needed? the corresponding linux action does not have these lines.

.github/workflows/ci.yml
```diff
@@ -659,7 +659,7 @@ jobs:
           export PATH="$GITHUB_WORKSPACE/dist/binary/borg-dir:$PATH"
           borg.exe -V
           . env/bin/activate
-          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --junitxml=test-results.xml
+          python -m pytest -n4 --benchmark-skip -vv -rs -k "not remote" --cov=borg --cov-config=pyproject.toml --cov-report=xml:coverage.xml --junitxml=test-results.xml
```
ThomasWaldmann: is --cov-report needed? the linux test run does not have this.

.github/workflows/ci.yml
mr-raj12: removed them in latest push and it just matches the Linux setup ,  waiting on CI to confirm codecov picks it up

.github/workflows/ci.yml
ThomasWaldmann: hmm, latest run does not have windows coverage.

ThomasWaldmann: I will merge this now. It makes windows consistent with linux regarding coverage data production and processing.

We can look at the remaining coverage related issues in a future PR.

PR #9438, opened 2026-03-05
Improve coverage (master)

ThomasWaldmann:


ThomasWaldmann:
Now looks better: https://app.codecov.io/gh/borgbackup/borg/pull/9438/blob/src/borg/platform/__init__.py?dropdown=coverage

Just an annoyance that the macOS runner is so slow / overloaded.

Issue #6535, opened 2022-04-04
use xxhash from pypi?


ThomasWaldmann:
we have our own binding to (lib)xxhash in `borg.checksums`, using cython.

it is not much code (because we only have bindings for the stuff we actually use), but we could maybe easily outsource it to that project and also have easier access to xxh128 or xxh3 than writing that ourselves.

https://pypi.org/project/xxhash/

ThomasWaldmann:
also, considering that we only have xxh64 and crc32 in `borg.checksums`, we could completely eliminate that module if we use pypi xxhash and after we do #6519.

PR #9440, opened 2026-03-05
use xxhash from pypi, for issue #6535

defnvary:
### Description

Use xxhash from pypi, instead of the custom cython binding.

### Issue
#6535

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
   "argon2-cffi",
   "shtab>=1.8.0",
   "backports-zstd; python_version < '3.14'", # for python < 3.14.
-  "xxhash",
+  "xxhash>=2.0.0",
 ]
 
 [project.optional-dependencies]


src/borg/testsuite/archiver/check_cmd_test.py
```diff
@@ -367,7 +367,14 @@ def test_verify_data(archivers, request, init_args):
...
     def fake_xxh64(data, seed=0):
-        return b"fakefake"
+        return FakeDigest()
+
+    # probably a hack
+    # returned object needs to have a digest method
+    # because xxhash.xxh64.digest() returns -> bytes
+    class FakeDigest:
+        def digest(self):
+            return b"fakefake"
```
ThomasWaldmann: if you only need it there, maybe move the class definition into the fake_xxh64 function body.

src/borg/cache.py
```diff
@@ -7,7 +7,7 @@
 from datetime import datetime, timezone, timedelta
 from pathlib import Path
 from time import perf_counter
-
+from xxhash import xxh64
 from borgstore.backends.errors import PermissionDenied
```
ThomasWaldmann, line 11: please do imports like this:

imports from stdlib
empty line
imports from 3rd party
empty line
imports from borg

also try to keep it alpha-sorted.

src/borg/checksums.pyi
```diff
@@ -1,8 +1 @@
-def crc32(data: bytes, value: int = 0) -> int: ...
-def xxh64(data: bytes, seed: int = 0) -> bytes: ...
-
-class StreamingXXH64:
-    def __init__(self, seed: int = 0) -> None: ...
-    def update(self, data: bytes) -> None: ...
-    def digest(self) -> bytes: ...
-    def hexdigest(self) -> str: ...
+def crc32(data: bytes, value: int = 0) -> int: ...
```
ThomasWaldmann: please keep the linefeed at the end of the file, otherwise diff complains.

src/borg/checksums.pyx
```diff
@@ -1,88 +1,6 @@
...
-    XXH64_hash_t XXH64_digest(const XXH64_state_t* statePtr)
-
-    void XXH64_canonicalFromHash(XXH64_canonical_t* dst, XXH64_hash_t hash)
-    XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src)
-
-
-cdef Py_buffer ro_buffer(object data) except *:
-    cdef Py_buffer view
-    PyObject_GetBuffer(data, &view, PyBUF_SIMPLE)
-    return view
```
ThomasWaldmann, line 39: guess this is still used by the code?

defnvary: this was called in StreamingXXH64 which is no longer needed. That portion is handled by xxhash.xxh64() now?

ThomasWaldmann: ah, right. i confused it with another definition that uses the same name.

defnvary:
Also there needs to some doc changes for this commit. should i include them in this pr or open a separate one. And I think we can remove streaming_xxh64 test from checksums_test.py. Also in test_xxh64 there is unnecessary conversion from bin_to_hex and from hex_to_bin in the third assert statement. Should i include test changes in this pr too or open a new one?

ThomasWaldmann:
> Also there needs to some doc changes for this commit. should i include them in this pr or open a separate one.

Include docs changes here. Be careful, do not edit generated docs (see comment in 1st line).

> And I think we can remove streaming_xxh64 test from checksums_test.py.

Sure.

> Also in test_xxh64 there is unnecessary conversion from bin_to_hex and from hex_to_bin in the third assert statement.
> Should i include test changes in this pr too or open a new one?

Yes, change the conversions, but keep the test data and hashes the same values.

ThomasWaldmann:
(ignore the macOS CI failure, I already fixed that in master)

defnvary:
Sorry for this many spammy pushes. I had messed up the commit history.

setup.py
```diff
@@ -163,11 +163,7 @@ def lib_ext_kwargs(pc, prefix_env_var, lib_name, lib_pkg_name, pc_version, lib_s
         dict(extra_compile_args=cflags),
     )
 
-    checksums_ext_kwargs = members_appended(
-        dict(sources=[checksums_source]),
-        lib_ext_kwargs(pc, "BORG_LIBXXHASH_PREFIX", "xxhash", "libxxhash", ">= 0.7.3"),
```
ThomasWaldmann, line 168: guess for this we need to require xxhash >= 2.0.0.

ThomasWaldmann: also somehow the CI testing did not run, let's see what it does with next push.

update: no, the tests are running, just not listed in that last post where they usually show up. the bsd failure is unrelated to your change.

ThomasWaldmann: looks like BORG_LIBXXHASH_PREFIX was never in our environment docs, thus no change needed there.

defnvary: > guess for this we need to require xxhash >= 2.0.0.

xxhash is bundled in python-xxhash, we dont need to build it at all. If i understand correctly this line searches xxhash or libxxhash on system (os specific packages) to compile xxhash?

ThomasWaldmann: i meant the python xxhash package, versions older than 2.0.0 have some (C) xxhash that is less than our original requirement.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "argon2-cffi",
   "shtab>=1.8.0",
   "backports-zstd; python_version < '3.14'", # for python < 3.14.
+  "xxhash",
```
ThomasWaldmann:
```
>= 2.0.0
```

defnvary: ok, got it.

ThomasWaldmann:
Thanks!

Issue #9463, opened 2026-03-11
is openbsd slower now somehow?


ThomasWaldmann:
I noticed that the OpenBSD CI test runner recently got cancelled frequently after a timeout of 90 minutes.

The other BSDs usually finish in less than half of that time.

Root cause?

false200:
Hi @ThomasWaldmann, I looked through the recent CI runs and it seems this slowdown may be related to the coverage config change rather than OpenBSD itself.

In commit 6df608ab (merged Mar 7, 2026), these lines were removed from pyproject.toml:

[tool.coverage.run]
parallel = true
sigterm = true

Timing evidence from OpenBSD CI runs:

Before the change:
run #5785 – https://github.com/borgbackup/borg/actions/runs/22803766748/job/66149469661 – ~21.0 min
run #5790 – https://github.com/borgbackup/borg/actions/runs/22805478221/job/66191681856 – ~21.4 min

After the change:
run #5792 – https://github.com/borgbackup/borg/actions/runs/22806558564/job/66158750167 – ~79.9 min
run #5795 – https://github.com/borgbackup/borg/actions/runs/22808518795/job/66171470526 – ~78.2 min
run #5800 – https://github.com/borgbackup/borg/actions/runs/22870304200/job/66348003545 – ~79.9 min

Since the tests run with `pytest -n auto --cov`, removing `parallel = true` could cause contention when multiple xdist workers write coverage data.

So i guess, it might be worth restoring `parallel = true` and `sigterm = true` under `[tool.coverage.run]` and re-running the OpenBSD CI to see if the runtime returns to normal.

ThomasWaldmann:
Strange. "parallel" was implied by another setting according to coverage docs, IIRC.

So, guess you could try adding sigterm back first and if that doesn't already help, parallel next.

false200:
> Strange. "parallel" was implied by another setting according to coverage docs, IIRC.
> 
> So, guess you could try adding sigterm back first and if that doesn't already help, parallel next.

Got it, yeahh that makes sense.

Would you like me to open a small PR restoring `sigterm = true` first so we can check how it affects the OpenBSD CI runtime?

ThomasWaldmann:
Yes, please do so.

false200:
> Yes, please do so.

I've opened a small PR restoring `sigterm = true` as discussed.

ThomasWaldmann:
Result of the experiments: coverage patch for "subprocess" made openbsd testing much slower.

[...]

PR #9465, opened 2026-03-11
tests: restore coverage sigterm handling to investigate OpenBSD CI slowdown (#9463)

false200:
## Description

Restore `sigterm = true` under `[tool.coverage.run]` in `pyproject.toml`.

This follows the discussion in #9463 to test whether restoring the `sigterm` setting affects the OpenBSD CI runtime.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -284,7 +284,7 @@ commands = [["bandit", "-r", "src/borg", "-c", "pyproject.toml"]]
 [tool.coverage.run]
 branch = true
 disable_warnings = ["module-not-measured", "no-ctracer"]
-patch = ["subprocess"]
+patch = []
 parallel = true
 sigterm = true
 source = ["src/borg"]


ThomasWaldmann:
We have to also consider that there is quite some variance due to github runner load, see for example how long the ubuntu runs took.

false200:
> We have to also consider that there is quite some variance due to github runner load, see for example how long the ubuntu runs took.

In that case, should we compare a few more OpenBSD runs to see if the ~80–90 min pattern still holds?

ThomasWaldmann:
Well, we can't be sure, but it still looks slow. Try "parallel" next.

ThomasWaldmann:
Another aspect is that these options should not be OS-specific (AFAIK). But we'll see...

false200:
> Well, we can't be sure, but it still looks slow. Try "parallel" next.

Yeah, I’ll try that out.

false200:
> Try "parallel" next.

Should I remove the sigterm = true change and instead try adding the parallel approach?

ThomasWaldmann:
hmm, still slow.

false200:
> hmm, still slow.

Right.

ThomasWaldmann:
still slow.

i suspect that coverage enabled is just slow. and we now have enabled it for subprocess, so it is enabled for more code.

final experiment: remove subprocess. just for science.

ThomasWaldmann:
OK, now openbsd is faster again, 1st run 33mins, 2nd run: 35mins.

The subprocess patch made it slower. But as we need subprocess coverage, guess there is no choice.

false200:
> OK, now openbsd is faster again, 33mins (I will restart that test run to get another measurement).
> 
> The subprocess patch made it slower. But as we need subprocess coverage, guess there is no choice.

Right.

ThomasWaldmann:
I filed an issue about coverage speed, but guess we can't solve it right now in this PR.

Thanks for working on this!

Issue #9433, opened 2026-03-03
remove chardet < 6 for requests >= 2.32.6


ThomasWaldmann:
requirements.d/*.txt - that chardet requirement was added because requests <= 2.32.5 spills out warnings with chardet 6+ that break our tests.

when requests releases a new, fixed version, we can remove that chardet requirement (we do not directly use it).

sagar-h007:
Hi @ThomasWaldmann , I'd like to work on this.
From the requests releases, versions >=2.32.6 include the fix, so the chardet < 6 constraint in requirements.d/*.txt may no longer be necessary. My plan is to remove that constraint, run the test suite to confirm there are no warnings or failures, and update dependencies accordingly.

Let me know if this approach sounds good.

ThomasWaldmann:
It's not released yet. But maybe we could just disable that warning, try that.

ThomasWaldmann:
requests has a new release!

PR #9467, opened 2026-03-11
tests: suppress RequestsDependencyWarning from requests/chardet

sagar-h007:
requests <= 2.32.5 emits RequestsDependencyWarning when chardet >= 6 is installed. The upstream fix is not released yet, so the chardet < 6 pin remains.

Suppress the warning in pytest configuration to avoid noisy test output.

fix: #9433

## Description

This PR suppresses the RequestsDependencyWarning emitted by requests when chardet >= 6 is installed.

Currently, requests <= 2.32.5 raises this warning because it expects chardet < 6. The upstream fix has not been released yet, so the existing chardet < 6 constraint in requirements.d/development.txt remains unchanged.

To avoid noisy test output, this change adds a filterwarnings rule in the pytest configuration (pyproject.toml) to ignore requests.exceptions.RequestsDependencyWarning during tests.

No dependency constraints were modified.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,12 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
 [tool.pytest.ini_options]
 markers = []
+filterwarnings = [
+    # requests <= 2.32.5 emits RequestsDependencyWarning when chardet >= 6 is installed.
+    # Keep the chardet < 6 pin in requirements.d/development.txt until the upstream fix
+    # in requests is released; suppress the noisy warning during tests in the meantime.
+    "ignore::requests.exceptions.RequestsDependencyWarning",
+]
 
 [tool.mypy]
 python_version = "3.10"


pyproject.toml
```diff
@@ -150,6 +150,12 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
 [tool.pytest.ini_options]
 markers = []
+filterwarnings = [
+    # requests <= 2.32.5 emits RequestsDependencyWarning when chardet >= 6 is installed.
+    # Keep the chardet < 6 pin in requirements.d/development.txt until the upstream fix
```
ThomasWaldmann, line 155: if we keep the pin, we don't need to filter the warning.

pyproject.toml
```diff
@@ -150,6 +150,12 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
...
+    # in requests is released; suppress the noisy warning during tests in the meantime.
```
ThomasWaldmann, line 156: i would suspect that if they release a "fixed" requests, there won't be a warning anymore.

sagar-h007:
@ThomasWaldmann You’re right — since chardet < 6 is pinned, the RequestsDependencyWarning shouldn’t occur in the test environment, so the filter isn’t necessary.

I’ll close this PR.

PR #9476, opened 2026-03-13
feat: add progress indicator to repository check loop

### LINKED ISSUE #9443: Add progress indicator to repository check loop
The `Repository.check()` method in `src/borg/repository.py` (line 319) has an existing TODO comment for a progress indicator, but it has not been implemented.

  For large repositories, the repository check loop iterates over all data objects, loading and verifying checksums for each one. This can take many hours with no user-visible progress output - only a raw key checkpoint is logged every 5 minutes (line 390-392).

  `ProgressIndicatorPercent` is already used in several other places in the codebase (`compact_cmd.py`, `archive.py`, `analyze_cmd.py`), so the pattern is well established.

  ### Proposed change

  Add a `ProgressIndicatorPercent` to the `check()` loop in `Repository`, showing the user a percentage/count as objects are verified. The total object count could be obtained via `store.list()` or a pre-count step.

  ### Related

  - `# TODO: progress indicator, ...` at `src/borg/repository.py:319`
  - #8996 - user unable to distinguish a working check from a frozen process
  - #8009 - similar request for segment check progress (borg 1.x)

sagar-h007:
### Description

This PR adds a simple progress indicator to the Repository.check() loop.

Running borg check on large repositories can take a long time while objects are verified, but currently there is no visible progress apart from periodic key checkpoint logs. This can make it difficult to tell whether the process is still working or has stalled.

This change adds a lightweight progress indicator that reports ongoing activity as objects are processed.

Following the maintainer’s suggestion, this does not compute totals using store.list(), since that could be slow for large or remote repositories. Instead, it only shows ongoing progress without percentage or ETA.

Fixes #9443

## Checklist

### DIFF
--- a/src/borg/helpers/progress.py
+++ b/src/borg/helpers/progress.py
@@ -9,7 +9,7 @@ logger = create_logger()
 
 class ProgressIndicatorBase:
     LOGGER = "borg.output.progress"
-    JSON_TYPE: str = None
+    JSON_TYPE: str = None  # type: ignore[assignment]
 
     operation_id_counter = 0
 
@@ -38,6 +38,38 @@ class ProgressIndicatorMessage(ProgressIndicatorBase):
 
     def output(self, msg):
         j = self.make_json(message=msg)
+        self.logger.info(j)
+
+
+class ProgressIndicatorObjectCounter(ProgressIndicatorBase):
+    JSON_TYPE = "progress_message"
+
+    def __init__(self, step=1000, msg="%d objects", msgid=None):
+        """
+        Activity-based progress indicator, simply tracking a changing count.
+
+        :param step: step size
+        :param msg: output message; must contain one %d placeholder for the count.
+        """
+        self.step = step
+        self.msg = msg
+        self.trigger_at = step
+        super().__init__(msgid=msgid)
+
+    def show(self, current=None):
+        """
+        Show and output the progress message if the step condition is met.
+
+        :param current: Set the current counter value.
+        """
+        if current is not None and current >= self.trigger_at:
+            # adjust trigger_at to the next step threshold
+            while self.trigger_at <= current:
+                self.trigger_at += self.step
+            return self.output(self.msg % current)
+
+    def output(self, message):
+        j = self.make_json(message=message)
         self.logger.info(j)
 
 
--- a/src/borg/helpers/progress.py
+++ b/src/borg/helpers/progress.py
@@ -41,8 +41,8 @@ class ProgressIndicatorMessage(ProgressIndicatorBase):
         self.logger.info(j)
 
 
-class ProgressIndicatorObjectCounter(ProgressIndicatorBase):
-    JSON_TYPE = "progress_message"
+class ProgressIndicatorCounter(ProgressIndicatorBase):
+    JSON_TYPE = "progress_counter"
 
     def __init__(self, step=1000, msg="%d objects", msgid=None):
         """
--- a/src/borg/helpers/progress.py
+++ b/src/borg/helpers/progress.py
@@ -9,7 +9,7 @@ logger = create_logger()
 
 class ProgressIndicatorBase:
     LOGGER = "borg.output.progress"
-    JSON_TYPE: str = None  # type: ignore[assignment]
+    JSON_TYPE: str | None = None
 
     operation_id_counter = 0
 
@@ -44,7 +44,7 @@ class ProgressIndicatorMessage(ProgressIndicatorBase):
 class ProgressIndicatorCounter(ProgressIndicatorBase):
     JSON_TYPE = "progress_counter"
 
-    def __init__(self, step=1000, msg="%d objects", msgid=None):
+    def __init__(self, step=1, msg="%d objects", msgid=None):
         """
         Activity-based progress indicator, simply tracking a changing count.
 
@@ -56,12 +56,18 @@ class ProgressIndicatorCounter(ProgressIndicatorBase):
         self.trigger_at = step
         super().__init__(msgid=msgid)
 
-    def show(self, current=None):
+    def show(self, current=None, increase=1, info=None):
         """
         Show and output the progress message if the step condition is met.
 
         :param current: Set the current counter value.
+        :param increase: Increase the current counter value.
         """
+        if current is not None:
+            self.counter = current
+        if hasattr(self, 'counter'):
+            self.counter += increase
+            current = self.counter
         if current is not None and current >= self.trigger_at:
             # adjust trigger_at to the next step threshold
             while self.trigger_at <= current:
--- a/src/borg/remote.py
+++ b/src/borg/remote.py
@@ -957,7 +957,8 @@ class RemoteRepository:
     @api(
         since=parse_version("1.0.0"),
         max_duration={"since": parse_version("1.2.0a4"), "previously": 0},
-        progress={"since": parse_version("never"), "dontcare": True},
+        # NOTE: update version when next beta is released
+        progress={"since": parse_version("2.0.0b21"), "dontcare": True},
     )
     def check(self, repair=False, max_duration=0, progress=False):
         """actual remoting is done via self.call in the @api decorator"""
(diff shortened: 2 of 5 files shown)


src/borg/helpers/progress.py
```diff
@@ -41,6 +41,38 @@ def output(self, msg):
         self.logger.info(j)
 
 
+class ProgressIndicatorObjectCounter(ProgressIndicatorBase):
+    JSON_TYPE = "progress_message"
```
ThomasWaldmann: ProgressIndicatorCounter
"progress_counter"

src/borg/repository.py
```diff
@@ -317,7 +317,7 @@ def check_object(obj):
             else:
                 log_error("too small.")
 
-        # TODO: progress indicator, ...
+        pi = ProgressIndicatorObjectCounter(step=1000, msg="Checking objects: %d", msgid="repository.check")
```
ThomasWaldmann: "Checked objects: %d" (it refers to what it already has done)

src/borg/repository.py
```diff
@@ -397,6 +398,7 @@ def check_object(obj):
                     self.store.store(LAST_KEY_CHECKED, key.encode())
                     break
             else:
+                pi.finish()
```
ThomasWaldmann: Guess you need another pi.finish() above the break in line 399 (guess before the log output in line 397), because it also leaves the loop there.

src/borg/repository.py
```diff
@@ -364,6 +364,7 @@ def check_object(obj):
                 obj_corrupted = False
                 check_object(obj)
                 objs_checked += 1
+                pi.show(objs_checked)
```
ThomasWaldmann: it unconditionally outputs the progress there. check how other borg commands deal with progress, guess most only enable it when `--progress` is given.

ThomasWaldmann (changes requested):
Thanks for the PR, some feedback.

ThomasWaldmann:
@sagar-h007 did you see my feedback?

src/borg/remote.py
```diff
@@ -954,8 +954,12 @@ def open(self, path, create=False, lock_wait=None, lock=True, exclusive=False, v
     def info(self):
         """actual remoting is done via self.call in the @api decorator"""
 
-    @api(since=parse_version("1.0.0"), max_duration={"since": parse_version("1.2.0a4"), "previously": 0})
-    def check(self, repair=False, max_duration=0):
+    @api(
+        since=parse_version("1.0.0"),
+        max_duration={"since": parse_version("1.2.0a4"), "previously": 0},
+        progress={"since": parse_version("never"), "dontcare": True},
```
ThomasWaldmann: for now, you can use a recent beta version "2.0.0b21" (or b20) here and put a comment above it that it will need updating to next beta version when it is released, so it is correct in the end.

ThomasWaldmann (changes requested):
A hint how to deal with api versioning.

src/borg/remote.py
sagar-h007: Got it, thanks for the hint

ThomasWaldmann: that "never" never worked anyway, see the failing CI:
```
  File "/home/runner/work/borg/borg/src/borg/remote.py", line 960, in RemoteRepository
    progress={"since": parse_version("never"), "dontcare": True},
                       ~~~~~~~~~~~~~^^^^^^^^^
  File "/home/runner/work/borg/borg/src/borg/version.py", line 24, in parse_version
    raise ValueError("Invalid version string %s" % version)
```

src/borg/repository.py
```diff
@@ -317,15 +317,12 @@ def check_object(obj):
...
-            ProgressIndicatorCounter(
-                step=1000,
-                msg="Checked objects: %d",
-                msgid="repository.check",
-            )
-            if progress
-            else None
-        )
+        if progress:
+            from .helpers.progress import ProgressIndicatorCounter
```
ThomasWaldmann: if possible, try to avoid doing imports inside loops, rather to them on module level.

check the other code if it also initializes the Progress Indicators only if --progress is given. guess initializing and ending them is no problem, but `.show()` needs to depend on --progress, of course.

sagar-h007: Moved the import to module level and removed the conditional import.
Now the progress indicator is always initialized, but `.show()` and `.finish()` only run when `--progress` is enabled, so no extra output without the flag.
Let me know if this looks right.

src/borg/helpers/progress.py
```diff
@@ -9,7 +9,7 @@
 
 class ProgressIndicatorBase:
     LOGGER = "borg.output.progress"
-    JSON_TYPE: str = None
+    JSON_TYPE: str = None  # type: ignore[assignment]
```
ThomasWaldmann: did mypy complain about this?

if so, maybe rather fix the type.

src/borg/helpers/progress.py
```diff
@@ -41,6 +41,38 @@ def output(self, msg):
         self.logger.info(j)
 
 
+class ProgressIndicatorCounter(ProgressIndicatorBase):
+    JSON_TYPE = "progress_counter"
+
+    def __init__(self, step=1000, msg="%d objects", msgid=None):
```
ThomasWaldmann: 1000 might be a good fit for your current usecase, but in general I think it might be a problematic default value. if not overriden and we have fewer, but larger objects to process, it might not display anything for a long time.

src/borg/repository.py
```diff
@@ -411,6 +417,8 @@ def check_object(obj):
                     )
         except StoreObjectNotFound:
             # it can be that there is no "data/" at all, then it crashes when iterating infos.
+            if progress:
+                pi.finish()
             pass
```
ThomasWaldmann: remove the "pass"

src/borg/helpers/progress.py
ThomasWaldmann: So, default to 1?

src/borg/helpers/progress.py
```diff
@@ -41,6 +41,38 @@ def output(self, msg):
...
+
+        :param step: step size
+        :param msg: output message; must contain one %d placeholder for the count.
+        """
+        self.step = step
+        self.msg = msg
+        self.trigger_at = step
+        super().__init__(msgid=msgid)
+
+    def show(self, current=None):
```
ThomasWaldmann: Implement the same signature as in the percent pi?
```
def show(self, current=None, increase=1, info=None)
```

src/borg/repository.py
```diff
@@ -364,6 +364,8 @@ def check_object(obj):
                 obj_corrupted = False
                 check_object(obj)
                 objs_checked += 1
+                if progress:
+                    pi.show(objs_checked)
```
ThomasWaldmann, line 369: somehow there is no consistency in master branch about this pattern.

ThomasWaldmann (changes requested):
some stuff i found

src/borg/helpers/progress.py
sagar-h007: Makes sense, defaulting to 1 is safer. I’ll update it.

src/borg/helpers/progress.py
sagar-h007: Got it

src/borg/helpers/progress.py
```diff
@@ -56,12 +56,18 @@ def __init__(self, step=1000, msg="%d objects", msgid=None):
...
 
         :param current: Set the current counter value.
+        :param increase: Increase the current counter value.
         """
+        if current is not None:
+            self.counter = current
+        if hasattr(self, 'counter'):
+            self.counter += increase
+            current = self.counter
         if current is not None and current >= self.trigger_at:
```
ThomasWaldmann, line 71: initialize self.counter = 0 in init method, so you don't need that hasattr.

also, move the code that is not related to producing output to a `progress` method as you see in the percent pi.

ThomasWaldmann: @sagar-h007 seen this? ^

ThomasWaldmann:
Closing this:
- no reaction by submitter
- meanwhile outdated due to the packs work by @mr-raj12 and borgstore related changes by @ThomasWaldmann (e.g. remote.py is not used any more for non-legacy repos)

PR #9479, opened 2026-03-15
Fix recreate function to handle archive None case

ANKITAPRAMODPATIL:
Added a check to ensure the archive is opened correctly before continuing the recreate process.  This prevents errors when the archive cannot be opened.

## Description
Added a validation check in the recreate function to ensure that the archive is opened successfully before processing.

This prevents potential errors if the archive cannot be opened and improves the robustness of the recreate function.

## Checklist

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -2246,15 +2246,22 @@ class ArchiveRecreater:
         self.stats = stats
         self.progress = progress
         self.print_file_status = file_status_printer or (lambda *args: None)
-
+        
     def recreate(self, archive_id, target_name, delete_original, comment=None):
         archive = self.open_archive(archive_id)
+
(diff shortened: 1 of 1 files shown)


src/borg/archive.py
```diff
@@ -2246,15 +2246,22 @@ def __init__(
         self.stats = stats
         self.progress = progress
         self.print_file_status = file_status_printer or (lambda *args: None)
-
+        
     def recreate(self, archive_id, target_name, delete_original, comment=None):
         archive = self.open_archive(archive_id)
+
+        if archive is None:
```
ThomasWaldmann, line 2253: there is no way open_archive can return None.

ThomasWaldmann:
No useful content.

PR #9501, opened 2026-03-17
Fixes #9495 preserve cwd archive metadata

### LINKED ISSUE #9495: borg2: transfer "cwd" archive metadata
TODO: check if it transfers this correctly (new since borg 1.4.4).

ebuzerdrmz44:
## Description                                                                                                                                                                       
   
  `borg transfer` was not preserving the `cwd`archive metadata from the source archive. Instead, the transferred archive would end up with the cwd of the machine running the transfer command.
                                                                                                                                                                                       
  This adds `cwd` to the kept attributes in both `UpgraderNoOp` and `UpgraderFrom12To20`, so the original value is carried over. Also added a check for it `test_transfer_archive_metadata`.
                                                                                                                                                                                       
  Fixes #9495                                                                                                                                                                          
  
  ## Checklist                                                                                                                                                                         
                  
### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -712,7 +712,7 @@ Duration: {0.duration}
         metadata |= {"size": stats.osize, "nfiles": stats.nfiles}
         metadata |= additional_metadata or {}
         if metadata.get("cwd") is None:
-            del metadata["cwd"] 
+            del metadata["cwd"]
         metadata = ArchiveItem(metadata)
         data = self.key.pack_metadata(metadata.as_dict())
         self.id = self.repo_objs.id_hash(data)


src/borg/upgrade.py
```diff
@@ -149,7 +150,7 @@ def upgrade_archive_metadata(self, *, metadata):
         new_metadata = {}
         # keep all metadata except archive version and stats. also do not keep
         # recreate_source_id, recreate_args, recreate_partial_chunks which were used only in 1.1.0b1 .. b2.
-        for attr in ("hostname", "username", "comment", "chunker_params"):
+        for attr in ("hostname", "username", "comment", "chunker_params", "cwd"):
             if hasattr(metadata, attr):
                 new_metadata[attr] = getattr(metadata, attr)
```
ThomasWaldmann, line 155: ok, guess that solves it for borg 1.4 archives that do have the cwd metadata (borg 1.4.4+).

but what happens if an older archive gets transferred that does not have the cwd metadata, what will be and what should be in the resulting archive?

ebuzerdrmz44: thanks for the review. I guess we shouldn't check cwd inside the for loop. After the loop, could we check it specifically, like` getattr(metadata, "cwd", "")`?

ebuzerdrmz44: But I'm not sure if "" is the best choice here . would you prefer leaving cwd unset or is that fine

ThomasWaldmann: if we can't know, having it unset might be the best choice. check if the code can cope with that.

ebuzerdrmz44: Turns out omitting cwd from additional_metadata wasn't enough . save() always falls back to self.cwd. Used None as a sentinel instead: upgraders pass cwd=None for archives without it, and save() drops the key when it sees None.

ThomasWaldmann:
Thanks!

Issue #241, opened 2015-10-03
possible loss of backup archives when following example in documentations with prune command


bedros:
in the automatic backup example and at the bottom there's a prune command as follows 

borg prune -v $REPOSITORY --keep-daily=7 --keep-weekly=4 --keep-monthly=6

however, with backup of multiple machines, and one machine has been disconnected for more than 6 months, such as a dead hard drive, the user will lose archives from all machines that are not backed up in last 6 months when issuing the above prune command from any machine, and the user will ultimately use the dead machine backup permanently; I've done dummy testing and verified that older archives from older machines are removed.

after looking into the documentation, I found the perfect solution which is to add --prefix to the prune command as follows 

borg prune -v $REPOSITORY --keep-daily=7 --keep-weekly=4 --keep-monthly=6 --prefix=`hostname`

such a prune command will only delete old archives of the current machine, and will leave archives of other machines untouched

please update the example by adding the --prefix=`hostname`

Thanks,

Bedros

ThomasWaldmann:
Thanks for the hint and sorry for any inconvenience this may have caused.

I had added a warning to borg prune --help a while ago, but I missed the example in the quickstart section. I also added a warning to the usage section, including a warning about choice of prefix which applies if you only use hostname as prefix - if you have hosts foo and foobar, this may not do what you want, so better include the "-" also if you have "hostname-date" as archive names.

Fixed by https://github.com/borgbackup/borg/commit/bf3cf21ac5d128785dacbc0282370524763099c6 .

Issue #213, opened 2015-09-20
use pyinstaller to create standalone executables


ThomasWaldmann:
pyinstaller has a 3.0.dev2 pre-release now on PyPi (for some platforms git repo code is needed).

```
 . env/bin/activate
cd borg
pip install -e .
pip install 'pyinstaller==3.0.dev2'
pyinstaller -F -n borg-PLATFORM --hidden-import=logging.config borg/__main__.py
```
See issue #214 for latest binaries.

ThomasWaldmann:
linux: worked for 32 and 64bit
mac os x: currently broken build, no python dynlib there
freebsd: currently broken build, bootloader does not compile, python dynlib is not found
netbsd/openbsd: no support in pyinstaller yet

level323:
@ThomasWaldmann Creating a standalone binary (or, more generally, redistributable application files) has great value to me (and, I suspect, to others). 

I had actually just begun hacking the Vagrantfile from the borg code (#196) with a plan of using Vagrant to build a redistributable. My initial plan was to automate a process I had already documented to build python from source in a borg-specific directory under /usr/local/, build borg in that dedicated python environment, ensure it passes the tests, and deliver a tarball of the /usr/local/ borg install.

There is obvious overlap with what you're trying with pyinstaller. pyinstaller will produce nice tidy single-file executables which are also much smaller (~6MB for pyinstaller vs ~60MB for my method). However, pyinstaller seems quite bleeding edge (at least in terms of python 3 support) and, according to your comments, is only working on linux presently.

If there is anyone interested in my approach I'm happy to keep you posted.

Naturally, I'm also very interested in the pyinstaller approach in the longer term. @ThomasWaldmann do you have a script or any special setup for building the borg executable with pyinstaller or is it just straightforward?

ThomasWaldmann:
> @ThomasWaldmann <https://github.com/ThomasWaldmann> Creating a
> standalone binary (or, more generally, redistributable application
> files) has great value to me (and, I suspect, to others).

Ah, good that it is useful.

I thought so, that not everybody likes the sometimes painful way to
install python stuff.

> I had actually just begun hacking the Vagrantfile from the borg code
> (#196 <https://github.com/borgbackup/borg/issues/196>) with a plan of
> using Vagrant to build a redistributable. My initial plan was to
> automate a process I had already documented to build python from source
> in a borg-specific directory under /usr/local/, build borg in that
> dedicated python environment, ensure it passes the tests, and deliver a
[...]

PR #9503, opened 2026-03-18
add tcsh completion support

ThomasWaldmann:
`borg completion tcsh` now generates a usable completion script.

Since the original version of this PR, all the generator fixes it used to carry as a local
monkeypatch of `shtab.complete_tcsh` have landed upstream in [shtab
1.9.3](https://github.com/tqdm/shtab/pull/213), which we already require — positional completion
under subcommands at any depth, no out-of-range `$cmd` indexing, custom `.complete` patterns in
multi-requirement rules, `--opt=` completion and rule deduplication. That monkeypatch (132 lines)
is therefore gone: with shtab 1.9.3 it produced exactly the same script as stock shtab.

What is left here:

- a tcsh preamble providing the dynamic completion helpers as aliases (tcsh has no functions),
  and the tcsh patterns for sort keys, files-cache mode, compression specs, chunker params,
  relative times, timestamps, file sizes and help topics
- completion of archive names, archive IDs (for tokens starting with `aid:`) and tags. tcsh can
  neither define functions nor use backquotes in these helpers (the completion rule invoking them
  is backquoted already), so both run one POSIX sh script that parses `$COMMAND_LINE` for
  `--repo`/`-r` and queries `borg repo-list`
- `_tcsh_anchor_positional_patterns()`, a temporary ~30-line backport of
  [tqdm/shtab#241](https://github.com/tqdm/shtab/pull/241) (merged upstream, not released yet)

### Why the backport is needed

A completion pattern (`f`, `d`, ...) for a positional of a subcommand ends up inside that
positional's `p@N@` rule. tcsh runs the clauses in such a rule as *commands* and completes from
their output, so a pattern there does nothing at all: without this, `borg umount <TAB>` offers no
mountpoint and `borg key export <TAB>` no path. The fix is to express them as `n/<subcommand>/`
rules instead, which tcsh does apply as patterns.

The rewrite produces byte-identical output to what a shtab containing #241 generates on its own,
and it is a no-op with such a shtab — so it can simply be deleted once we require the release that
contains it. That is noted as a TODO in its docstring.

### Not covered

- tcsh has no completion descriptions, so archive completion is a plain candidate list there,
  unlike in zsh and fish
- tcsh matches completions for positional arguments by word position, so options preceding an
  archive name shift it out of place and it is not completed - the command's epilog points tcsh
  users at BORG_REPO for this. [tqdm/shtab#247](https://github.com/tqdm/shtab/pull/247) fixes
  this for options before the subcommand (`borg -r REPO list <TAB>`)
- options are mingled across subcommands in shtab's tcsh backend, so e.g. `--compression` is
  offered for every subcommand

### Testing

Tests pass against both shtab 1.9.3 and current shtab main (21 passed, 1 skipped), and the
generated script was verified by driving a real interactive tcsh 6.21.00: `borg <TAB>`,
`borg key <TAB>`, `borg key change-location <TAB>`, `borg help <TAB>`, `borg umount <TAB>`,
`borg key export <TAB>`, `borg create --compression=z<TAB>` and `borg list --sort-by=<TAB>` all
complete as expected. Against a real repository (two archives, two tags), `borg list <TAB>` gives
the archive names, `borg list aid:<TAB>` gives `aid:` + 8 hex digits per archive, and
`borg tag --add=<TAB>` gives the tags.

## Checklist

### DIFF
--- a/src/borg/archiver/completion_cmd.py
+++ b/src/borg/archiver/completion_cmd.py
@@ -16,6 +16,7 @@ The following argument types have intelligent, context-aware completion:
    - Completes archive names by default (e.g., "my-backup-2024")
    - Completes archive IDs when prefixed with "aid:" (e.g., "aid:12345678")
    - In zsh and fish, shows archive metadata (name, timestamp, user@host) as descriptions
+     (tcsh has no completion descriptions)
    - Respects --repo/-r flags to query the correct repository
 
 2. Sort keys (SortBySpec):
@@ -53,6 +54,8 @@ The following argument types have intelligent, context-aware completion:
 11. File sizes (parse_file_size):
    - Suggests common file size values (500M, 1G, 10G, 100G, 1T, etc.)
 """
+
+import re
 
 import shtab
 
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
This was created with AI assistance (Gemini Flash/Pro) and likely needs more work:
- that monkeypatching of shtab isn't pretty. IF there are bugs / missing features in shtab tcsh support, they should be fixed there.
- archivename / aid completion is missing. The AI tried to write that, but ran into severe troubles. To at least have something simpler to begin with, I decided to skip that for the initial commit.

ThomasWaldmann:
@simaoafonso-pwt Could you have a look at that monkeypatch please?

simaoafonso-pwt:
> @simaoafonso-pwt Could you have a look at that monkeypatch please?

Thanks for tagging me, I know that tcsh support is not fully featured, but it was the best I could do.

I see that monkey patching is changing a single function, can you please create a diff between that new function, and the one on `shtab` itself, so that it's easier to review the changes?

ThomasWaldmann:
@simaoafonso-pwt I added a draft PR there:

https://github.com/iterative/shtab/pull/213

simaoafonso-pwt:
> @simaoafonso-pwt I added a draft PR there:
> 
> [iterative/shtab#213](https://github.com/iterative/shtab/pull/213)

Thanks, I'll review that.

ThomasWaldmann:
Rebased onto current master and reworked.

- shtab [#213](https://github.com/tqdm/shtab/pull/213) is merged and released in **shtab 1.9.3**,
  which borg already requires, so the local monkeypatch of `shtab.complete_tcsh` (132 lines) is
  removed: its generated script was identical to stock 1.9.3's apart from the header comment.
- What remains of the generator work is `_tcsh_anchor_positional_patterns()`, a temporary backport
  of [tqdm/shtab#241](https://github.com/tqdm/shtab/pull/241) (merged, unreleased) so that
  `borg umount <TAB>` / `borg key export <TAB>` complete a mountpoint / path at all. Its output is
  byte-identical to what a shtab with #241 generates, and it is a verified no-op with such a
  shtab, so it can be dropped once we require that release.
- Rebasing over the fish-completion work meant adopting master's `for_all_shells()` helper; tcsh is
  now folded into it (`tcsh=False` for the archive and tags helpers, which have no tcsh
  implementation).
- New test asserts the anchored rules exist and, more generally, that no `p@N@` clause action is a
  bare completion pattern.

Diff is down from +192/−8 to +122/−7. Title and description updated to match.

ThomasWaldmann:
Added archive name, `aid:` and tag completion for tcsh, which this PR previously left out.

tcsh has no functions, and these helpers cannot use backquotes because the completion rule
invoking them is backquoted already. So both aliases run one POSIX sh script (kept in a csh
variable) that parses `$COMMAND_LINE` for `--repo`/`-r` and queries `borg repo-list`. Verified
against a real repository by driving an interactive tcsh:

```
borg list <TAB>        ->  arch-one arch-two
borg list aid:<TAB>    ->  aid:2f4b2c84 aid:a428b2b4
borg tag --add=<TAB>   ->  important nightly
```

Caveat, now noted in the command's epilog: tcsh matches positional completions by word position,
so `borg list --repo REPO <TAB>` shifts the archive out of place and completes nothing — with
BORG_REPO set it works. [tqdm/shtab#247](https://github.com/tqdm/shtab/pull/247) fixes the
options-before-the-subcommand half of that upstream (`borg -r REPO list <TAB>`); with that shtab,
`borg --repo REPO info <TAB>` completes archive names here.

Tests are up to 21 passed / 1 skipped against both shtab 1.9.3 and shtab main.

PR #9530, opened 2026-03-22
Use mtime-based files cache default on Windows

### LINKED ISSUE #7193: borg2: ctime (windows)
borg usually uses ctime, inode, size for `--files-cache`.

on linux, this is the best way to determine a file has not change since we cache information about it into the files cache.

on windows, ctime attribute is the **file creation time**, thus pretty useless for this purpose.

because of that, `test_file_status_cs_cache_mode` is currently skipped on windows, but this can't be the final solution.

borg rather should use `--files-cache=mtime,size` on windows (not sure about inode).

hiijoshi:
## Summary

Fix default files-cache mode on Windows.

On Windows, `ctime` represents file creation time rather than inode change time. Using `ctime` in the default files-cache mode can therefore fail to detect modified files correctly.

## Changes

- Use an `mtime`-based files cache mode on Windows
- Keep existing `ctime,size,inode` default behavior unchanged on POSIX systems

## Reasoning

This aligns with the issue discussion and ensures more reliable change detection on Windows systems.

## Testing

- Verified code compiles and integrates with existing logic
- No changes to non-Windows behavior
## Additional Improvement

Also added a warning when using ctime-based files cache mode on Windows, since ctime represents file creation time and may not detect file changes correctly.

Fixes #7193

### DIFF
--- a/src/borg/archiver/create_cmd.py
+++ b/src/borg/archiver/create_cmd.py
@@ -919,8 +919,10 @@ class CreateMixIn:
             dest="files_cache_mode",
             action=Highlander,
             type=FilesCacheMode,
-            default=FILES_CACHE_MODE_UI_DEFAULT,
-            help="operate files cache in MODE. default: %s" % FILES_CACHE_MODE_UI_DEFAULT,
+            default=FILES_CACHE_MODE_UI_DEFAULT_WIN32 if is_win32 else FILES_CACHE_MODE_UI_DEFAULT,
+            help=("operate files cache in MODE. default: %s"
+             % (FILES_CACHE_MODE_UI_DEFAULT_WIN32 if is_win32 else FILES_CACHE_MODE_UI_DEFAULT)
+),
         )
         fs_group.add_argument(
             "--files-changed",
--- a/src/borg/constants.py
+++ b/src/borg/constants.py
@@ -118,6 +118,7 @@ CH_DATA, CH_ALLOC, CH_HOLE = 0, 1, 2
 
 # operating mode of the files cache (for fast skipping of unchanged files)
 FILES_CACHE_MODE_UI_DEFAULT = "ctime,size,inode"  # default for "borg create" command (CLI UI)
+FILES_CACHE_MODE_UI_DEFAULT_WIN32 = "mtime,size,inode" # default for "borg create" on win32
 FILES_CACHE_MODE_DISABLED = "d"  # most borg commands do not use the files cache at all (disable)
 
 # account for clocks being slightly out-of-sync, timestamps granularity.


src/borg/archiver/create_cmd.py
```diff
@@ -226,6 +226,11 @@ def create_inner(archive, cache, fso):
         self.noacls = args.noacls
         self.noxattrs = args.noxattrs
         dry_run = args.dry_run
+        if is_win32 and "ctime" in str(args.files_cache_mode):
+            logger.warning(
+        "Using ctime on Windows may not detect file changes correctly. "
+        "Consider using mtime-based cache mode."
+    )
```
ThomasWaldmann: If you do it like that, it will emit a warning, but then it will continue and use ctime, although we know that this won't work.

hiijoshi: Thanks, that makes sense. A warning alone would still continue with an unreliable ctime-based mode on Windows. I’ve removed that part and kept this PR focused on switching the default files-cache mode on Windows to an mtime-based mode, while leaving non-Windows behavior unchanged.

src/borg/archiver/create_cmd.py
```diff
@@ -919,7 +919,7 @@ def build_parser_create(self, subparsers, common_parser, mid_common_parser):
             dest="files_cache_mode",
             action=Highlander,
             type=FilesCacheMode,
-            default=FILES_CACHE_MODE_UI_DEFAULT,
+            default=FilesCacheMode("mtime,size") if is_win32 else FILES_CACHE_MODE_UI_DEFAULT,
```
ThomasWaldmann: No, the return value of that call is a different kind of string than FILES_CACHE_MODE_UI_DEFAULT. Also, we need to keep using inode.

So, add another constant for win32 and use that for win32 systems.

FilesCacheMode is a validator as well as a transformer. The validation part needs to be changed when running on win32.

hiijoshi:
Thanks, I reworked this to use a dedicated Windows default constant and kept inode in the mode.
The files-cache related tests pass locally. The remaining failure is in a remote_archiver test due to a remote connection issue, which appears unrelated to this change.

hiijoshi:
Thanks, I reworked this to use a dedicated Windows default constant and kept `inode` in the mode as suggested.

The files-cache related tests pass locally. The remaining failure is in a `remote_archiver` test due to a remote connection issue, which appears unrelated to this change.

ThomasWaldmann:
That validator is still to do, see my previous comment.

ThomasWaldmann:
https://github.com/borgbackup/borg/actions/runs/23470453669/job/68345226260?pr=9530

Run black locally, ideally by using pre-commit (see our developer docs).

PR #9535, opened 2026-03-23
ci: disable MSYS2 path conversion in windows tests, fixes #9513

### LINKED ISSUE #9513: CI: tests failing on windows
Seen in the canary tests, the windows CI (PR/Merge) should also be checked:
```
E       ValueError: invalid literal for int() with base 10: "MSYS2 path translation is active. This can cause POSIX paths to be mangled into Windows paths in archives. Consider setting MSYS2_ARG_CONV_EXCL='*' and MSYS2_ENV_CONV_EXCL='*'. See https://www.msys2./
```

hiijoshi:


ThomasWaldmann:
@hiijoshi Looks like the fix is correct.

But why is there a non-sensical PR comment?

Also, the commit comment should have a trailing ", fixes #XXXX" so it auto-closes the respective issue.

hiijoshi:
@ThomasWaldmann 

Thanks, I removed the unnecessary PR comment and updated the commit message to include fixes #9513.

PR #9542, opened 2026-03-24
patterns: clean up TODOs, move is_include_cmd to IECommand property, fixes #9442

### LINKED ISSUE #9442: Clean up TODOs in patterns.py: recurse_dir_default config, is_include_cmd placement, validation
Description:                                                                                                                                                                                                                                                                                                                  
  There are several small TODO items in `src/borg/patterns.py` that could be addressed as code quality improvements:                                                                                                                                                                                                            
  ### 1. `recurse_dir_default` as config option (line 93-94)

  Currently `recurse_dir_default` is hardcoded to `True` in `PatternMatcher.__init__`. When no pattern matches a path, `match()` falls back to this value (line 162), always recursing into unmatched directories.

  Should this be made configurable so users can control whether unmatched directories are recursed into by default?

  ### 2. Move `is_include_cmd` mapping (line 98-99)

  The `IECommand -> bool` mapping (`{Exclude: False, ExcludeNoRecurse: False, Include: True}`) is defined inside `PatternMatcher.__init__` but logically belongs closer to the `IECommand` enum itself or in `parse_inclexcl_command`.

  It could be refactored as a class attribute on `IECommand` or a standalone helper, improving readability and reducing coupling.

  ### 3. Validate `RootPath` string (line 372)

  In `parse_inclexcl_command`, when the command is `IECommand.RootPath`, the remainder string is assigned directly without any validation. Adding basic validation (e.g., checking it is a valid path) would make error reporting better.

  ### 4. Handle unexpected cmds in `command_recurses_dir` (line 320)

  `command_recurses_dir()` currently does not handle unexpected `cmd` values like `RootPath` or `PatternStyle`. It should either raise an error or return `None` for those cases instead of silently returning `True`.

mr-raj12:
## Description

Address the TODO items in `src/borg/patterns.py` as discussed in #9442:

1. **`recurse_dir_default`**: Replaced the misleading TODO with a comment explaining why it must stay `True` as include patterns inside excluded directories (`+` inside `-`) depend on it.

2. **`is_include_cmd`**: Moved from a dict on `PatternMatcher` to an `is_include` property on `IECommand` and now the matcher calls `cmd.is_include` instead of doing a dict lookup.

3. **`RootPath` validation**: Added a warning when the root path doesn't start with `/`, so backups don't fail due to a typo.

4. **`command_recurses_dir`**: Replaced `cmd not in [IECommand.ExcludeNoRecurse]` with explicit `if`/`raise` branches using `is` comparisons. Now raises `ValueError` for unexpected commands like `RootPath` or `PatternStyle`.

Tests updated to use real `IECommand` values and cover the new code paths.

## Checklist

- [x] PR is against `master` 
- [x] Tests pass

### DIFF
--- a/src/borg/patterns.py
+++ b/src/borg/patterns.py
@@ -3,8 +3,10 @@ import posixpath
 import re
 import sys
 import unicodedata
+import warnings
 from collections import namedtuple
 from enum import Enum
+from pathlib import Path
 
 from .helpers import clean_lines, shellpattern
 from .helpers.argparsing import Action, ArgumentTypeError
@@ -89,14 +91,12 @@ class PatternMatcher:
         # False when calling match().
         self.recurse_dir = None
 
-        # whether to recurse into directories when no match is found
-        # TODO: allow modification as a config option?
+        # Whether to recurse into directories when no match is found.
+        # This must be True so that include patterns inside excluded directories
+        # work correctly (e.g. "+ /excluded_dir/important" inside "- /excluded_dir").
         self.recurse_dir_default = True
 
         self.include_patterns = []
-
-        # TODO: move this info to parse_inclexcl_command and store in PatternBase subclass?
-        self.is_include_cmd = {IECommand.Exclude: False, IECommand.ExcludeNoRecurse: False, IECommand.Include: True}
 
     def empty(self):
         return not len(self._items) and not len(self._path_full_patterns)
@@ -150,13 +150,13 @@ class PatternMatcher:
         if value is not non_existent:
             # we have a full path match!
             self.recurse_dir = command_recurses_dir(value)
-            return self.is_include_cmd[value]
+            return value.is_include
 
         # this is the slow way, if we have many patterns in self._items:
         for pattern, cmd in self._items:
             if pattern.match(path, normalize=False):
                 self.recurse_dir = pattern.recurse_dir
-                return self.is_include_cmd[cmd]
+                return cmd.is_include
 
         # by default we will recurse if there is no match
         self.recurse_dir = self.recurse_dir_default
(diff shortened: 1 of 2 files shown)


src/borg/patterns.py
```diff
@@ -368,7 +376,8 @@ def parse_inclexcl_command(cmd_line_str, fallback=ShellPattern):
         raise ArgumentTypeError("A pattern/command must have a value part.")
 
     if cmd is IECommand.RootPath:
-        # TODO: validate string?
+        if not remainder_str.startswith("/"):
```
ThomasWaldmann: the roots in a pattern file are somehow equivalent to the PATHs given as cli arguments for borg create.

we do not require the PATHs to be absolute paths, so guess we should not require the roots to be abs paths either.

in both cases it might be clearer and less risky to use abs paths, but the user can use rel paths on their own risk.

validation: check if the path exists and if not, output a warning.

src/borg/patterns.py
```diff
@@ -376,8 +377,18 @@ def parse_inclexcl_command(cmd_line_str, fallback=ShellPattern):
         raise ArgumentTypeError("A pattern/command must have a value part.")
 
     if cmd is IECommand.RootPath:
-        if not remainder_str.startswith("/"):
-            logger.warning("Root path %r does not look like an absolute path", remainder_str)
+        # Check if path is absolute
+        is_absolute = remainder_str.startswith("/")
```
ThomasWaldmann: that does not work on win32.

`Path(p).is_absolute()` or so.

src/borg/patterns.py
```diff
@@ -376,8 +377,18 @@ def parse_inclexcl_command(cmd_line_str, fallback=ShellPattern):
...
+        # Check if path exists
+        path_exists = os.path.exists(remainder_str)
+
+        # Warn about relative paths
+        if not is_absolute:
+            logger.warning("Root path %r is relative, recommended to use absolute path", remainder_str)
```
ThomasWaldmann: ..., it is recommended to use an ...

PR #9544, opened 2026-03-25
fix: chunk store_load over RPC to avoid msgpack BufferFull.

hiijoshi:
Fixes a crash when loading large stores via RPC


ThomasWaldmann:
Your PRs are a mess and I am tired of reviewing them.

PR #9547, opened 2026-03-28
refactor: replace logging with structlog in helpers/yes_no.py (issue #9529)

### LINKED ISSUE #9529: structlog?
borg uses human readable log output as well as json log output.

Maybe we could use Hynek's `structlog` and refactor log output.

https://www.structlog.org/en/stable/index.html

Note: some tools wrap borg and use cli options, json api and rc code to interface to a borg subprocess. Try to stay compatible.

Could maybe be also helpful for #9454.

hiijoshi:
Closes #9529

Refactored log output in `helpers/` to use `structlog` instead of Python's built-in `logging` module.

**Changes:**
- `helpers/yes_no.py`: replaced `logging.getLogger` with `structlog.get_logger`
- `helpers/misc.py`: removed unused `logging` import, replaced `logging.INFO` with plain integer `20`

**Skipped for now:**
- `helpers/progress.py`: uses `setLevel()` which structlog doesn't support directly — needs discussion
- `helpers/__init__.py`: uses `logging.getLogger("borg.output.show-rc")` — needs discussion

All existing tests pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,8 @@ dependencies = [
   "backports-zstd; python_version < '3.14'", # for python < 3.14.
   "xxhash>=2.0.0",
   "jsonargparse>=4.47.0",
-  "PyYAML>=6.0.2",  # we need to register our types with yaml, jsonargparse uses yaml for config files
+  "PyYAML>=6.0.2",
+  "structlog",
 ]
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
not enough to be useful.

PR #9558, opened 2026-04-03
Fixes tests/docs assuming XDG_* vars not used on macOS

Goddesen:
## Description

I use `XDG_CONFIG_HOME` in my setup on macOS and noticed that some tests would fail on `master`. This PR changes the failing tests to fix the assumption that `XDG_*` vars wouldn't be respected on macOS using `platformdirs`. I also fixed any references I found in relevant documentation.

I am not entirely sure about the legacy config dir tests that had branches for `not is_darwin and not is_win32`, someone should double check that. :)

This may have been a very recent change in `platformdirs` and may also require pinning its version to an appropriate range:
- https://platformdirs.readthedocs.io/en/latest/changelog.html#id8
- https://platformdirs.readthedocs.io/en/latest/parameters.html#xdg-environment-variables

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -124,7 +124,7 @@ Compatibility notes:
 
 - use platformdirs 3.x.x instead of home-grown code. Due to that:
 
-  - XDG_*_HOME is not honoured on macOS and on Windows.
+  - XDG_*_HOME is not honoured on Windows.
   - BORG_BASE_DIR can still be used to enforce some base dir + .config/ or .cache/.
   - on macOS, the default directories move to native locations:
     config/data: ``~/Library/Application Support/borg/``,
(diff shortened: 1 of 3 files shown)


ThomasWaldmann (approved):
LGTM! Thanks for the PR!

Goddesen:
Cool!

@ThomasWaldmann Do keep in mind this means `platformdirs` may need to be locked to `>= 4.6.0` on macOS, I didn't notice that exactly until after.

ThomasWaldmann:
Yes, if one wants the tests to work. pip usually installs the latest package, so that would be the case.

If somebody wants to just use borg with an older platformdirs, that would work, too, just not the tests.

Issue #9556, opened 2026-04-02
Separate Borg 1.x legacy code into borg.legacy package


mr-raj12:
### Problem

Borg 2.x code is tangled with 1.x legacy code across ~15 files. `manifest.py` has 15+ `if not self.legacy:` branches in the `Archives` class alone. `legacyrepository.py` and `legacyremote.py` already have "legacy" in the name but sit at the package root. `src/borg/legacy/` exists but is empty.

### What this does

Move the bulk of 1.x code into `src/borg/legacy/`. The goal is to eliminate almost every last `# legacy` comment , some one-liner branches are worse to extract than to keep. But the major classes and branch-heavy code (LegacyRepository, LegacyRemoteRepository, RepoObj1, the 4 legacy key types, the entire Archives legacy path) all move out.

After this:
- v2 core modules are mostly free of legacy code
- `borg transfer --from-borg1` keeps working via `borg.legacy`
- Dropping v1 support later = delete `borg.legacy` + remove a handful of scattered one-liners

### Plan (6 phases)

Each phase is a separate PR.

| Phase | What happens |
|---|---|
| 1 | `git mv` of `legacyrepository.py` + `legacyremote.py` into `legacy/` |
| 2 | Extract `RepoObj1`, move legacy key types (`KeyfileKey`, `RepoKey`, etc.) into `legacy/` with duplicated base classes |
| 3 | Extract `LegacyArchives` from `Archives`, remove all `self.legacy` branches in `manifest.py` |
| 4 | Switch `remote.py` to lazy-import `LegacyRepository` only inside `RepositoryServer.open()` |
| 5 | `git mv` of `upgrade.py` into `legacy/` |
| 6 | Inventory remaining scattered legacy markers, split what can be split, document what stays |

Phase 1 + 2.1 (RepoObj1) done in #9559.

### Target structure

```
src/borg/legacy/
├── __init__.py
├── repository.py      ← LegacyRepository
├── remote.py          ← LegacyRemoteRepository
├── repoobj.py         ← RepoObj1
├── manifest.py        ← LegacyArchives
├── upgrade.py         ← UpgraderFrom12To20, UpgraderNoOp
└── crypto/
    ├── __init__.py
    └── key.py         ← Legacy AES-CTR key types (self-contained, duplicated bases)
```

[...]

PR #9559, opened 2026-04-03
legacy: move LegacyRepository, LegacyRemoteRepository, RepoObj1 into borg.legacy package, refs #9556

mr-raj12:
## Description

Moves Borg 1.x legacy code into `borg.legacy`, refs #9556.

- `legacyrepository.py` → `legacy/repository.py` (git mv)
- `legacyremote.py` → `legacy/remote.py` (git mv)
- `RepoObj1` extracted from `repoobj.py` → `legacy/repoobj.py`
- All consumer imports updated to `borg.legacy.*` paths
- `repoobj.py` keeps a re-export for `RepoObj1` since the file still exists for `RepoObj`

No logic changes. Just file moves, import path updates (`.foo` → `..foo`), and one logger name fix in the tests.

No stubs at old paths for the moved files, they weren't public API.

Refs #9556

## Checklist

- [X] PR is against master
- [ ] New code has tests
- [X] Tests pass


ThomasWaldmann:
This is hard to review due to the code moves.

Can you use an approach using git mv (when possible), commit, modify?

mr-raj12:
Sure

mr-raj12:
fixed with git mv, now it shows in a proper way

ThomasWaldmann:
RepoObj1 is identical, just moved?

mr-raj12:
> RepoObj1 is identical, just moved?

yes, the same RepoObj1 with updated relative imports

mr-raj12:
I have updated the description to match what actually landed after the git mv rework.

Two commits:
- 4db927c6: git mv of legacyrepository.py and legacyremote.py into legacy/, import paths adjusted
- 15bec44c: extract RepoObj1 into legacy/repoobj.py, update consumer imports

RepoObj1 is identical, only the imports changed (. to  ..) as it's one level deeper now. Couldn't git mv a class out of a file that still has other code, so that one was a manual extract only.

There's still a `from .legacy.repoobj import RepoObj1` re-export at the bottom of repoobj.py, all internal imports already use the new path though. Happy to drop it if you'd rather not keep it.

Follow-up PRs I have planned (each separate, building on this):
- Move the 4 legacy key classes (KeyfileKey, RepoKey, Blake2KeyfileKey, Blake2RepoKey) into borg.legacy.crypto.key, with duplicated base classes so legacy crypto is fully independent of borg.crypto.key
- Extract LegacyArchives from manifest.py to eliminate the 15+ `if not self.legacy:` branches in the Archives class
- Switch remote.py to lazy-import LegacyRepository only inside RepositoryServer.open() when v1_or_v2 is set
- Move upgrade.py → legacy/upgrade.py (only used by transfer_cmd.py)

After all of this merges, removing borg.legacy is a single directory delete with no impact on v2 code.

ThomasWaldmann:
Thanks! Merged.

PR #9560, opened 2026-04-04
Fixes cleanup of append-only test tempfiles on macOS/BSD

Goddesen:
## Description

The `test_extract_restores_append_flag` test leaves append-only tempfiles around on macOS and FreeBSD that cannot be removed cleanly. This was previously just ignored by the cleanup func but those files occasionally caused lots of warning output on subsequent test runs. Fixed by attempting to clear flags and retry whenever the cleanup function fails.

I am not entirely sure what triggers the warnings, I can usually do a couple of test runs before the warnings start showing even though the files would have been left around after the first test run. Cleaning them up properly is the right thing to do anyway.

## Checklist

### DIFF
--- a/src/borg/conftest.py
+++ b/src/borg/conftest.py
@@ -133,7 +133,21 @@ def archiver(tmp_path, set_env_variables):
     os.chdir(archiver.tmpdir)
     yield archiver
     os.chdir(old_wd)
-    shutil.rmtree(archiver.tmpdir, ignore_errors=True)  # clean up
+
+    def maybe_clear_flags_and_retry(func, path, _exc_info):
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Ah, good catch!

I already wondered why it does not clean up everything, but didn't debug it yet.

ThomasWaldmann:
BTW, the haiku test failure comes from setuptools_scm + its dependencies, just ignore it for now.

Goddesen:
Codecov triggers on conftest.py? 🤔

Maybe that should be ignored or moved into the testsuite.

ThomasWaldmann:
Thanks!

PR #9576, opened 2026-04-20
_chunker.c: fix some bugs

ThomasWaldmann:
- better check return value of fd.read(n) and reject if it returns more bytes than requested.
- avoid giving len<=0 to posix_fadvise(), which could drop the rest of the file from cache.
- chunker_fill: correctly Py_DECREF in case of an error in PyBytes_Size
- buzhash: check for len == 0 edge case


ThomasWaldmann:
superceded by #9586

PR #9587, opened 2026-05-04
LRUCache fixes 1.4

ThomasWaldmann:


ThomasWaldmann:
Note: no port to master needed, the implementation there was changed to use an OrderedDict, so this patch does not apply there.

PR #9385, opened 2026-02-20
platform: use F_FULLSYNC on macOS for SyncFile data durability, fixes #9383

mr-raj12:
- `SyncFile.sync()` calls `platform.fdatasync(self.fd)`, which resolves to `os.fdatasync` or `os.fsync` - On macOS, `fsync()` only flushes data to the drive's write cache, not to persistent storage                                                         
  - True durability requires `fcntl(fd, F_FULLFSYNC)` — without it, a power loss could lose data that `SyncFile` believed was safely persisted
  - This was acknowledged with a TODO at `src/borg/platform/base.py:154`: `TODO: Use F_FULLSYNC on macOS.`
  - Implements a platform-specific sync wrapper as suggested in #9383

  Fixes #9383

  ## Changes

  | File | Change |
  |------|--------|
  | `src/borg/platform/darwin.pyx` | Add `fdatasync()` using `fcntl F_FULLFSYNC` with `os.fsync()` fallback; add `sync_dir()` using the same | 
  | `src/borg/platform/__init__.py` | Import `fdatasync`, `sync_dir` from `.darwin` in the darwin block |
  | `src/borg/platform/base.py` | Remove resolved TODO comment |
  | `src/borg/testsuite/platform/darwin_test.py` | Add 4 tests: F_FULLFSYNC usage, fsync fallback, fdatasync integration, sync_dir integration |       

  ## Checklist
  - [x] PR is against `master`
  - [x] New code has tests
  - [x] Tests pass
  - [x] Commit message references related issue
  - [x] Code formatted with black

PR #9592, opened 2026-05-04
platform: use F_FULLFSYNC on macOS for SyncFile data durability, fixes #9383

### LINKED ISSUE #9383: platform: use F_FULLSYNC on macOS for SyncFile data durability
## Summary
  - `SyncFile.sync()` calls `platform.fdatasync(self.fd)`, which resolves to `os.fdatasync` or `os.fsync`
  - On macOS, `fsync()` only flushes data to the drive's write cache, not to persistent storage
  - True durability on macOS requires `fcntl(fd, F_FULLSYNC)`
  - This was already acknowledged with a TODO comment at `src/borg/platform/base.py:154`: `TODO: Use F_FULLSYNC on macOS.`
  - Without `F_FULLSYNC`, Borg's data durability guarantees on macOS are weaker than intended - a power loss could result in data that `SyncFile` believed was safely persisted being lost

  ## Proposed fix

  Detect macOS (`sys.platform == "darwin"`) and use `fcntl.fcntl(fd, fcntl.F_FULLSYNC)` in the fdatasync path:

  ```python
  import sys

  if sys.platform == "darwin":
      import fcntl
      fcntl.fcntl(fd, fcntl.F_FULLSYNC)
  else:
      os.fsync(fd)
  ```

  This could either live in `platform/darwin.py` as a platform-specific override of `fdatasync`, or inline in `base.py` with a platform check.

  | File | Change |
  |------|--------|
  | `src/borg/platform/base.py` or `src/borg/platform/darwin.py` | Use `fcntl.F_FULLSYNC` on macOS instead of `os.fsync`;|

mr-raj12:
Backport of #9385 to `1.4-maint`.

`SyncFile.sync()` calls `platform.fdatasync(self.fd)`, which on macOS resolves to `os.fsync()`. `fsync()` only flushes to the drive's write cache as it does **not** guarantee data reaches persistent storage. A power loss could silently lose data
that `SyncFile` believed was safely persisted.

The fix was acknowledged with a TODO in `src/borg/platform/base.py`: `TODO: Use F_FULLSYNC on macOS.`

This backport implements the platform-specific override suggested in #9383: add `fdatasync()` and `sync_dir()` to `platform/darwin.pyx` using `fcntl(fd, F_FULLFSYNC)` with an `os.fsync()` fallback for network filesystems that do not support `F_FULLFSYNC`.

## Changes

| File | Change |
|------|--------|
| `src/borg/platform/darwin.pyx` | Add `fdatasync()` using `fcntl F_FULLFSYNC` with `os.fsync()` fallback; add `sync_dir()` using the same |
| `src/borg/platform/__init__.py` | Import `fdatasync`, `sync_dir` from `.darwin` in the darwin block |
| `src/borg/platform/base.py` | Remove resolved TODO comment |
| `src/borg/testsuite/platform.py` | Add 4 tests: F_FULLFSYNC usage, fsync fallback, fdatasync integration, sync_dir integration |
| `docs/changes.rst` | Add changelog entry for 1.4.x |

## Checklist

- [x] PR is against `1.4-maint` (maintenance branch as change is only applicable here as master is already fixed via #9385)
- [x] Tests pass

### DIFF
--- a/src/borg/platform/__init__.py
+++ b/src/borg/platform/__init__.py
@@ -43,6 +43,7 @@ elif is_darwin:  # pragma: darwin only
     from .darwin import acl_get, acl_set
     from .darwin import is_darwin_feature_64_bit_inode, _get_birthtime_ns
     from .darwin import set_flags
+    from .darwin import fdatasync, sync_dir  # type: ignore[no-redef]
 
 
 def get_birthtime_ns(st, path, fd=None):
--- a/src/borg/platform/darwin.pyx
+++ b/src/borg/platform/darwin.pyx
@@ -1,3 +1,4 @@
+import fcntl
 import os
 
 from libc.stdint cimport uint32_t
@@ -259,3 +260,30 @@ def set_flags(path, bsd_flags, fd=None):
         path_bytes = os.fsencode(path)
         if lchflags(path_bytes, c_flags) == -1:
             raise OSError(errno.errno, os.strerror(errno.errno), os.fsdecode(path_bytes))
+
+
+def fdatasync(fd):
+    """macOS fdatasync using F_FULLFSYNC for true data durability.
+
+    os.fsync() is an OS-level flush (kernel page cache -> drive write buffer).
+    F_FULLFSYNC additionally issues a HW-level flush (drive write buffer -> persistent storage).
+    Falls back to os.fsync() if F_FULLFSYNC is not supported (e.g. network fs).
+    """
+    try:
+        fcntl.fcntl(fd, fcntl.F_FULLFSYNC)
+    except OSError:
+        os.fsync(fd)
+
+
+def sync_dir(path):
+    """Sync a directory to persistent storage on macOS using F_FULLFSYNC."""
+    if isinstance(path, str):
+        path = os.fsencode(path)
+    fd = os.open(path, os.O_RDONLY)
+    try:
+        fdatasync(fd)
+    except OSError as os_error:
+        if os_error.errno != errno.EINVAL:
+            raise
+    finally:
+        os.close(fd)
--- a/src/borg/platform/base.py
+++ b/src/borg/platform/base.py
@@ -143,7 +143,6 @@ class SyncFile:
 
     Calling SyncFile(path) for an existing path will raise FileExistsError, see comment in __init__.
 
-    TODO: Use F_FULLSYNC on macOS.
     TODO: A Windows implementation should use CreateFile with FILE_FLAG_WRITE_THROUGH.
     """
 
--- a/src/borg/testsuite/platform.py
+++ b/src/borg/testsuite/platform.py
@@ -9,6 +9,7 @@ import unittest
 from ..platformflags import is_win32, is_linux, is_freebsd, is_netbsd, is_darwin, is_haiku
 from ..platform import acl_get, acl_set, swidth
 from ..platform import get_process_id, process_alive
+from ..platform import fdatasync, sync_dir
 from . import BaseTestCase, unopened_tempfile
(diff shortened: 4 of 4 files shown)


docs/changes.rst
```diff
@@ -414,6 +414,14 @@ Compatibility notes:
...
 ==========
 
+Version 1.4.5 (not yet released)
+---------------------------------
+
+Fixes:
+
+- platform: use F_FULLFSYNC on macOS for true data durability in SyncFile, #9383.
+
+
```
ThomasWaldmann: I usually create the change log at release time (or periodically).

Having updates in PRs regularly would just cause a lot of merge conflicts.

mr-raj12: > I usually create the change log at release time (or periodically).
> 
> Having updates in PRs regularly would just cause a lot of merge conflicts.

Removing this as of now

src/borg/platform/__init__.py
```diff
@@ -4,7 +4,14 @@
...
 
-from ..platformflags import is_win32, is_linux, is_freebsd, is_darwin, is_cygwin, is_haiku
+from ..platformflags import (
+    is_win32,
+    is_linux,
+    is_freebsd,
+    is_darwin,
+    is_cygwin,
+    is_haiku,
+)
```
ThomasWaldmann: there is still a black change.

ThomasWaldmann:
I ran the tests on macOS (ARM), they passed.

src/borg/platform/darwin.pyx
```diff
@@ -259,3 +260,30 @@ def set_flags(path, bsd_flags, fd=None):
         path_bytes = os.fsencode(path)
         if lchflags(path_bytes, c_flags) == -1:
             raise OSError(errno.errno, os.strerror(errno.errno), os.fsdecode(path_bytes))
+
+
+def fdatasync(fd):
+    """macOS fdatasync using F_FULLFSYNC for true data durability.
+
+    On macOS, os.fsync() only flushes to the drive's write cache.
+    fcntl F_FULLFSYNC flushes to persistent storage.
```
ThomasWaldmann: The first sentence is a bit weird. It "only" flushes the drive's write cache. What else is there to flush that FULLSYNC does?

ThomasWaldmann: fsync() on macOS

fsync() (and Python's os.fsync()) asks the OS kernel to flush dirty data from its page cache to the drive. On macOS, however, fsync() does not guarantee that data has been committed to stable storage on the physical device. It only ensures the data has been handed off to the drive's own write buffer/cache.

If the drive has a volatile write cache (most HDDs and many SSDs do), a power loss after fsync() returns can still result in data loss or corruption, because the drive firmware hasn't necessarily flushed its internal cache to the platters/NAND.

F_FULLFSYNC (macOS-specific)

F_FULLFSYNC is a macOS-specific fcntl command that goes one step further: it issues a hardware flush command (like FLUSH CACHE in ATA or SYNCHRONIZE CACHE in SCSI/NVMe) to the drive, forcing it to commit all buffered writes from the drive's own cache to persistent storage.

ThomasWaldmann: so, os.fsync is only a OS-level flush, while fcntl F_FULLSYNC does a hw-level write buffer flush additionally.

mr-raj12: Have updated the docstring to reflect the two-stage distinction

ThomasWaldmann:
@mr-raj12 Can you apply that docs change to master branch also, please?

mr-raj12:
Done, opened #9597.

PR #9600, opened 2026-05-11
legacy: move legacy AES-CTR key classes into borg.legacy.crypto, refs #9556

mr-raj12:
## Description

Moves the four Borg 1.x AES-CTR key classes into `borg.legacy.crypto`, refs #9556.

- `KeyfileKey`, `RepoKey`, `Blake2KeyfileKey`, `Blake2RepoKey` → `legacy/crypto/key.py`
- `LEGACY_KEY_TYPES` moved with them
- `borg.crypto.key` re-exports all four classes and `LEGACY_KEY_TYPES` for backwards compatibility
- Shared base classes (`FlexiKey`, `AESKeyBase`, `ID_HMAC_SHA_256`, `ID_BLAKE2b_256`) remain in `borg.crypto.key` and are imported by the legacy module (no duplication)

No logic changes. The legacy import in `borg.crypto.key` is placed after `FlexiKey` is fully defined so the partial module already has all needed names during the circular load.

Refs #9556

## Checklist

- [X] PR is against master
- [ ] New code has tests
- [X] Tests pass

### DIFF
--- a/src/borg/crypto/key.py
+++ b/src/borg/crypto/key.py
@@ -29,7 +29,7 @@ from ..repoobj import RepoObj
 
 
 from .low_level import AES, bytes_to_int, num_cipher_blocks, hmac_sha256, blake2b_256
-from .low_level import AES256_CTR_HMAC_SHA256, AES256_CTR_BLAKE2b, AES256_OCB, CHACHA20_POLY1305
+from .low_level import AES256_OCB, CHACHA20_POLY1305
 from . import low_level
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
LGTM! This was about the repository data encryption.

There is also some other legacy crypto stuff in low_level.pyx, but separating that would be a bit more effort as it affects also the build process, so guess we can keep that as is for now.

PR #9602, opened 2026-05-11
Argon2 via OpenSSL >= 3.2

ThomasWaldmann:


ThomasWaldmann:
Guess we need to wait for:
- [X] **Ubuntu 26.04 github actions runner image**
  - https://github.com/actions/runner-images#available-images
  - https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2604-Readme.md
  - https://github.com/actions/runner-images/issues/14226
  - [ ] do they finally offer pre-cached other Python versions? https://github.com/borgbackup/borg/pull/10097 not yet, but it also works without them.

- [x] NetBSD 11 via cross-platform-actions, see #10071

ThomasWaldmann:
Note: Delay merge after 2.0.0b23 release (give beta testers a bit more time in case they still run ubuntu 24.04 or some other dist with too old openssl).

ThomasWaldmann:
@mr-raj12 do you have Ubuntu 26.04 already (if you use Ubuntu as dev system)?

mr-raj12:
> @mr-raj12 do you have Ubuntu 26.04 already (if you use Ubuntu as dev system)?

Yes

ThomasWaldmann:
## argon2 equivalence check: OpenSSL vs argon2-cffi

I compared `borg.crypto.low_level.argon2_hash` (OpenSSL `EVP_KDF`) against `argon2-cffi` over a
broad parameter sweep. **642 parameter sets compared, 0 mismatches.**

Environment: macOS/arm64, OpenSSL 3.6.3, argon2-cffi 25.1.0 (libargon2 version `0x13`),
`os.cpu_count() == 12`, branch head `afe7fb8ef`.

### 1. Structured sweep — 137 combinations, all match

Each dimension swept while the rest are held at borg's defaults, plus an exhaustive small
cross product:

| Dimension | Values covered |
|---|---|
| secret | `b""`, 1, 4, 22, 200 bytes, plus non-ASCII UTF-8 (umlauts + emoji) |
| salt | 8, 16, 32 bytes and a 24-byte random salt |
| time_cost | 1, 2, 3, 4, 5, 8 |
| (memory_cost, parallelism) | (8,1) (16,2) (64,8) (256,1) (1024,4) (65536,1/4/8) (131072,2) |
| hash_len | 4, 5, 16, 31, 32, 33, 48, 64, 100, 128 |
| type | `i`, `d`, `id` |

This includes both configurations borg actually uses: the default
`t=3, m=64MiB, p=4, len=32, id` and the `BORG_TESTONLY_WEAKEN_KDF` one (`t=1, m=8, p=1`).
The returned length always equalled the requested `hash_len`, and the 60 distinct 32-byte
results were in fact all distinct (no parameters accidentally collapsing onto the same output).

### 2. Randomized fuzz — 500 seeded-random sets, all match

Random secrets (0..255 bytes), salts (8..64 bytes), `t in {1,2,3,4,8}`, `p in {1..24}`,
`m` at and just above the `8*p` minimum plus larger values, `hash_len in {4..100}`, all three
types. `p = 16` and `p = 24` are deliberately above `cpu_count = 12`, so the
`threads = min(parallelism, os.cpu_count())` clamp is exercised — results stayed identical
to argon2-cffi, which confirms the output is independent of the thread count actually used.

### 3. High-memory cases — all match

```
t=3 m=65536KiB   p=4  hl=32 id  MATCH   (borg default)
t=4 m=262144KiB  p=8  hl=32 id  MATCH
t=2 m=1048576KiB p=8  hl=64 id  MATCH   (1 GiB)
t=1 m=1048576KiB p=12 hl=32 d   MATCH   (lanes == cpu_count)
t=3 m=524288KiB  p=16 hl=32 i   MATCH   (lanes > cpu_count)
```

### 4. Invalid inputs

Both implementations reject the same inputs — salt shorter than 8 bytes, `hash_len < 4`,
`memory_cost < 8*p`, `time_cost = 0`, `parallelism = 0`. Only the diagnostic differs: OpenSSL
gives a generic `CryptoError: EVP_KDF_derive failed` where argon2-cffi names the specific
problem ("Salt is too short", "Memory cost is too small", "Too few lanes", ...). Same
accept/reject decision in every case tried, just a coarser message. An unknown type string
still raises `ValueError: Invalid argon2 type` before OpenSSL is touched.

### Performance

At borg's defaults the two are a wash: min 31.0 ms (OpenSSL) vs 30.0 ms (argon2-cffi) over
7 runs each. The ~2.8x speedup from the multi-threaded lane computation is relative to the
single-threaded OpenSSL path, not relative to argon2-cffi.

PR #9618, opened 2026-05-13
docs: document error handling in borg create (fixes #4912)

### LINKED ISSUE #4912: What happens when Borg can't read a file/folder?
This is a question, a request to update the documentation, and a feature request all in one.

**Question:** What happens when `borg create` can't read a file? Does it act as if the file didn't exist?

**Feature request:** Assuming that is indeed what happens, could we have an option that makes borg treat the broken file/folder as if it wasn't modified since the last backup?

**Documentation:** This doesn't seem to be documented anywhere, so the docs should be updated with this information.

ThomasWaldmann:
This PR adds a new 'Error handling' section to the 'borg create' help text. It explicitly documents that when Borg encounters unreadable files or directories (e.g., due to permissions or locks), it skips them with a warning, continues with the rest of the backup, and exits with return code 1. This fixes #4912.

### DIFF
--- a/src/borg/archiver.py
+++ b/src/borg/archiver.py
@@ -3748,6 +3748,21 @@ class Archiver:
         Therefore, when using ``--one-file-system``, you should double-check that the backup works as intended.
 
 
+        Error handling
+        ++++++++++++++
+ 
+        When Borg encounters a file or directory it cannot read (e.g., due to permission
+        denied, or another process holding an exclusive lock on Windows), it will:
+ 
+        - Emit a warning message to stderr.
+        - Skip that file or directory and continue backing up the remaining files.
+        - Exit with return code 1 (Warning) at the end of the operation, instead of 0
+          (Success).
+ 
+        This ensures that a single problematic file does not abort your entire backup.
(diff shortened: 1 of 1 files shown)


src/borg/archiver.py
```diff
@@ -3748,6 +3748,21 @@ def define_borg_mount(parser):
         Therefore, when using ``--one-file-system``, you should double-check that the backup works as intended.
 
 
+        Error handling
+        ++++++++++++++
+ 
+        When Borg encounters a file or directory it cannot read (e.g., due to permission
+        denied, or another process holding an exclusive lock on Windows), it will:
+ 
+        - Emit a warning message to stderr.
```
ThomasWaldmann: "Log a warning message."

src/borg/archiver.py
```diff
@@ -3748,6 +3748,21 @@ def define_borg_mount(parser):
...
+        - Skip that file or directory and continue backing up the remaining files.
+        - Exit with return code 1 (Warning) at the end of the operation, instead of 0
```
ThomasWaldmann: "Exit with a Warning RC at the end of the operation."

PR #9639, opened 2026-05-13
legacy: extract LegacyArchives into legacy/archives.py, add ArchivesInterface, refs #9556

mr-raj12:
Moves the Borg 1.x archive list management out of `Archives` and into `LegacyArchives`, refs #9556.

- `LegacyArchives` extracted from `Archives` into `legacy/archives.py`
- `Archives` in `manifest.py` is now 2.x only as all `self.legacy` branches removed
- `ArchivesInterface` added to `manifest.py` as a `@runtime_checkable` Protocol as mypy/pyright will catch drift between the two implementations
- `Manifest.__init__` picks `LegacyArchives` or `Archives` based on repository type via lazy imports

No logic changes. `borg transfer --from-borg1` keeps working through `borg.legacy.archives`. `borg serve` for v1 clients is unaffected — it only opens `LegacyRepository` for raw RPC byte-forwarding and never instantiates `LegacyArchives`.

Refs #9556

Checklist

- [x] PR is against master
- [x] New code has tests
- [x] Tests pass

### DIFF
--- a/src/borg/legacy/archives.py
+++ b/src/borg/legacy/archives.py
@@ -1,0 +1,275 @@
+"""Legacy archive list management for Borg 1.x repositories.
+
+In Borg 1.x the list of archives is embedded directly in the manifest blob
+as a dict {name: {"id": bytes, "time": str}}.  This class manages that dict.
+Used by ``borg transfer --from-borg1`` and ``borg serve`` (serving v1 clients).
+
+This module can be removed entirely when Borg 1.x support is dropped.
+"""
+
+import re
+from datetime import datetime
+from operator import attrgetter
+
+from ..constants import *  # NOQA
+from ..helpers.datastruct import StableDict
+from ..helpers.errors import CommandError, Error
+from ..helpers.parseformat import bin_to_hex
+from ..helpers.time import parse_timestamp
+from ..item import ArchiveItem
+from ..patterns import get_regex_from_pattern
+
+# ArchiveInfo and filter_archives_by_date are imported from ..manifest below.
+# These module-level imports are safe because legacy/archives.py is only ever
+# imported from inside Manifest.__init__ — by that point manifest.py is fully
+# loaded and present in sys.modules.
+from ..manifest import ArchiveInfo, filter_archives_by_date
+
+
+class LegacyArchives:
+    """
+    Manage the list of archives for a Borg 1.x repository.
+
+    The archive registry lives inside the manifest blob itself:
+        {name: {"id": <bytes>, "time": <iso-timestamp-str>}}
+
+    Manifest.__init__ chooses this class over Archives when the repository is a
+    LegacyRepository.  It can be deleted entirely when Borg 1.x support is dropped.
+    """
+
+    def __init__(self, repository, manifest):
+        self.repository = repository
+        self.manifest = manifest
+        # key: str archive name, value: dict("id": bytes_id, "time": str_iso_ts)
+        self._archives = {}
+
+    def prepare(self, manifest, m):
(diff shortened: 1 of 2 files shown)


src/borg/legacy/archives.py
```diff
@@ -0,0 +1,275 @@
+"""Legacy archive list management for Borg 1.x repositories.
+
+In Borg 1.x the list of archives is embedded directly in the manifest blob
+as a dict {name: {"id": bytes, "time": str}}.  This class manages that dict.
+Used by ``borg transfer --from-borg1`` and ``borg serve`` (serving v1 clients).
```
ThomasWaldmann, line 5: I don't think `borg serve` uses `LegacyArchives` (or `Archives`).

Think of an encrypted repo - only the client has the key, the server can't even look into the blobs.

src/borg/manifest.py
```diff
@@ -69,65 +70,92 @@ def get_first_and_last_archive_ts(archives_list):
...
 class Archives:
     """
-    Manage the list of archives.
+    Manage the list of archives for a Borg 2.x repository.
 
-    We still need to support the borg 1.x manifest-with-list-of-archives,
-    so borg transfer can work.
-    borg2 has separate items archives/* in the borgstore.
+    Each archive has a separate entry in borgstore at archives/<hex-id>.
+    The manifest blob itself carries an empty archives dict.
```
ThomasWaldmann, line 125: Guess that is rather a leftover and should not be a documented feature. :-)

src/borg/manifest.py
```diff
@@ -272,82 +286,55 @@ def _lookup_name(self, name, raw=False):
...
+                        name=archive_info["name"],
+                        id=archive_info["id"],
+                        ts=ts,
+                        tags=archive_info["tags"],
+                        user=archive_info["username"],
+                        host=archive_info["hostname"],
+                    )
+                return archive_info
+        return None  # id not in store, or archive metadata blob missing from repo
+        # TODO: add a test that calls get_by_id() with a non-existent id and asserts None is returned
```
ThomasWaldmann, line 313: seen that?

src/borg/manifest.py
```diff
@@ -272,82 +286,55 @@ def _lookup_name(self, name, raw=False):
...
 
     def create(self, name, id, ts, *, overwrite=False):
         assert isinstance(name, str)
         assert isinstance(id, bytes)
         if isinstance(ts, datetime):
             ts = ts.isoformat(timespec="microseconds")
         assert isinstance(ts, str)
-        if not self.legacy:
-            # we only create a directory entry, its name points to the archive item:
```
ThomasWaldmann, line 328: looks like it killed my comment and replaced it by another.

ThomasWaldmann:
This is a good first step for the archives list part of the Manifest.

In a further step (not this PR) we could try whether we still need other parts of the Manifest or whether we could get rid of them.

The main reason why this thing was called Manifest in borg 1.x (the manifest as a signed list of archives) is not in the manifest blob anymore in borg2.

src/borg/legacy/archives.py
mr-raj12: thanks for correcting, removed it

src/borg/manifest.py
mr-raj12: added the test

src/borg/manifest.py
mr-raj12: restored your comment, apologies it missed my local review there

src/borg/manifest.py
mr-raj12: true

PR #9643, opened 2026-05-14
add blake3, blake2b is legacy

ThomasWaldmann:
See #8867.


ThomasWaldmann:
Either wait or try setting that env var or downgrade testing to 3.14:
```
   error: the configured Python interpreter version (3.15) is newer than PyO3's maximum supported version (3.14)
    = help: please check if an updated version of PyO3 is available. Current version: 0.26.0
    = help: set PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 to suppress this check and build anyway using the stable ABI
```

PR #9650, opened 2026-05-16
legacy: lazy-import legacy classes inside v1 branches only, refs #9556

mr-raj12:
Removes top-level `LegacyRepository`, `LegacyRemoteRepository`, and `RepoObj1` imports from three files, moving each to the branch that uses it, refs #9556.

- `borg.remote`: `LegacyRepository` imported inside `RepositoryServer.open()`, guarded by `v1_or_v2`
- `borg.archiver._common`: `LegacyRemoteRepository` and `LegacyRepository` imported inside `get_repository()` under the same guard; `RepoObj1` imported inside `with_other_repository()` under `repository.version == 1`
- `borg.archiver.transfer_cmd`: `LegacyRepository` imported at the top of `transfer_chunks()`, where it appears in the `except` tuple

No logic changes. None of these modules pull in `borg.legacy` at import time anymore, so a future phase can drop the package without them breaking at startup.

Unit tests added in `testsuite/archiver/_common_test.py` covering the `get_repository()` v1_or_v2 branches directly (SSH and local path), using mocks to avoid requiring a real legacy repository.

Refs #9556

Checklist

- [x] PR is against master
- [x] New code has tests
- [x] Tests pass

### DIFF
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -13,12 +13,9 @@ from ..helpers.argparsing import SUPPRESS, PositiveInt
 from ..helpers.nanorst import rst_to_terminal
 from ..manifest import Manifest, AI_HUMAN_SORT_KEYS
 from ..patterns import PatternMatcher
-from ..legacy.remote import LegacyRemoteRepository
 from ..remote import RemoteRepository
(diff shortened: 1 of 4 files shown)


ThomasWaldmann (approved):
LGTM

PR #9655, opened 2026-05-18
legacy: move upgrade.py into borg.legacy, refs #9556

mr-raj12:
## Description

Moves `UpgraderNoOp` and `UpgraderFrom12To20` into `borg.legacy`, refs #9556.

- `src/borg/upgrade.py` → `src/borg/legacy/upgrade.py`
- Relative imports updated from `.X` to `..X`
- `transfer_cmd.py`: `from .. import upgrade as upgrade_mod` → `from ..legacy import upgrade as upgrade_mod`

No logic changes. Both classes are Borg 1.x-only. Leaving them at the package root means an extra file to track down when v1 support is dropped; in `borg.legacy` it goes away with the rest.

Refs #9556

## Checklist

- [X] PR is against master
- [ ] New code has tests
- [X] Tests pass


ThomasWaldmann:
UpgraderFrom12To20 - that is for borg 1.x (at that time it was 1.2, meanwhile it is 1.4).

The NoOp upgrader is not legacy, it is used when transferring archives from borg2 -> borg2 repos. Not really an upgrade, rather moving stuff.

ThomasWaldmann (approved):
LGTM (I assume you just moved the 12to20 code and you didn't change it).

PR #9664, opened 2026-05-23
legacy: move hardlink helpers, NSIndex1, AES, and PBKDF2 key file methods to borg.legacy, refs #9556

mr-raj12:
## Description

Moves four borg 1.x components into `borg.legacy`, refs #9556.

- `helpers/fs.py` → `legacy/helpers.py`: hardlink helper functions
- `hashindex.pyx` → `legacy/hashindex.py`: `NSIndex1` (borg 1.x hash index format)
- `crypto/low_level.pyx` → `legacy/crypto/low_level.pyx`: `cdef class AES` (AES-256-CTR for borg 1.x key file encryption)
- `crypto/key.py` → `legacy/crypto/key.py`: PBKDF2 key file methods extracted into `Pbkdf2FileMixin`

`AES` needs its own `.pyx` , `cdef` types can't live in a `.py` file and can't be shared across binary extensions. `ro_buffer` (a cdef helper) is duplicated rather than imported.

`setup.py` and `pyproject.toml` updated to build and type-check `borg.legacy.crypto.low_level`. All consumer imports updated. No logic changes.

Refs #9556

## Checklist

- [X] PR is against master
- [ ] New code has tests
- [X] Tests pass
- [X] Commit message references related issue

PR #9665, opened 2026-05-24
benchmark cpu: drop KDF section

mr-raj12:
Drops the KDF section from `do_benchmark_cpu`.

KDFs are supposed to be slow, that's the whole point. Timing them gives no useful signal, and the benchmark output already said so: `"KDFs (slow is GOOD, use argon2!)"`.

Suggested in https://github.com/borgbackup/borg/pull/9664#pullrequestreview-4350990208.

- Removed the KDF benchmark block (print header + for loop)
- Removed `number_kdf` and the `FlexiKey` import, both only used by that block

No logic changes.

## Checklist

- [X] PR is against master
- [ ] New code has tests
- [X] Tests pass
- [ ] Commit message references related issue

### DIFF
--- a/src/borg/archiver/benchmark_cmd.py
+++ b/src/borg/archiver/benchmark_cmd.py
@@ -6,7 +6,6 @@ import tempfile
 import time
 
 from ..constants import *  # NOQA
-from ..crypto.key import FlexiKey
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Thanks!

PR #9666, opened 2026-05-24
fix: xattr xdg backup exclusion should be on 'false'

Flachz:
## Description

According to the [freedesktop proposal](https://www.freedesktop.org/wiki/CommonExtendedAttributes/) the xattrs attribute `user.xdg.robots.backup` should be set to `false` to exclude files from backups.

The current implementation has it backwards, as files are excluded if `user.xdg.robots.backup` is set to `true`.

## Checklist

### DIFF
--- a/src/borg/archive.py
+++ b/src/borg/archive.py
@@ -1231,7 +1231,7 @@ def maybe_exclude_by_attr(item):
     if xattrs := item.get("xattrs"):
         apple_excluded = xattrs.get(b"com.apple.metadata:com_apple_backup_excludeItem")
         linux_excluded = xattrs.get(b"user.xdg.robots.backup")
-        if apple_excluded is not None or linux_excluded == b"true":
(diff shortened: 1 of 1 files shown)


src/borg/archive.py
```diff
@@ -1231,7 +1231,7 @@ def maybe_exclude_by_attr(item):
     if xattrs := item.get("xattrs"):
         apple_excluded = xattrs.get(b"com.apple.metadata:com_apple_backup_excludeItem")
         linux_excluded = xattrs.get(b"user.xdg.robots.backup")
-        if apple_excluded is not None or linux_excluded == b"true":
+        if apple_excluded is not None or linux_excluded == b"false":
             raise BackupItemExcluded
```
ThomasWaldmann, line 1235: Good catch! Thanks for finding this!

But please rename linux_excluded, in this case it should be linux_included.

ThomasWaldmann:
Ah, crap, clicked too fast...

#9667

PR #9669, opened 2026-05-27
docs: add pack file format design and internals documentation, refs #8572

mr-raj12:
### pack format design and documentation

Adds `docs/internals/packs.rst` covering the pack file format design: binary layout, pack ID, Phase 1 (N=1), write order/crash safety, index namespace, recovery path, and repository version requirements.
Includes three diagrams (ObjHeader structure, binary layout, write order).

refs #8572.


docs/internals/pack-layout.png
ThomasWaldmann, line 1: This is quite close to what we have currently, so it might be good for a little-change first step.

But I have thought about these changes: 

- blob_len is somehow redundant, can be computed from meta and data len
- if there is corruption in the length values, all entries afterwards are undiscoverable, because we can't determine their offsets. i thought about repeating the "BORGPACK" magic (and version) before EACH of the blob entries. then we could scan for that in case length is corrupted (and also we could easily notice THAT length is corrupted). this would make all pack entries look the same and remove the special case for the magic header.

docs/internals/pack-objheader.png
ThomasWaldmann, line 1: - we can check full pack integrity server-side by sha256 now (that was not the case yet), what would xxh64(meta) and xxh64(data) give us additionally? if borg check runs on the client and has detected that a pack is damaged (by server-side sha256), it could read the pack from the client. the aead decryption would detect any problem (and just fail) in the meta and data blobs (each of these are encrypted/authenticated separately).
- the magic and version could just be part of this header structure

docs/internals/pack-write-order.png
ThomasWaldmann, line 1: As I mentioned, there is no real "manifest" (meaning a signed list of archives) in borg2 now. Each single archive is a trust root now. The manifest is rather an artifact from the past which I would like to get rid of.

So, "write archive" + "write archive pointer" instead of "write manifest"?

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+
+.. _packs:
+
+Pack files
+==========
+
+Borg currently stores each repository object (chunk) as a separate object in the
+borgstore.  For large repositories this means millions of individual objects, each
+requiring its own I/O round trip to read or write.  On high-latency backends (SFTP,
+cloud object storage) this overhead dominates backup and restore times.
```
ThomasWaldmann: As the docs describe the current state, we will have to remove this later.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+    0       4     uint32le meta_size
+    4       4     uint32le data_size
+    8       8     bytes    xxh64(encrypted_meta)
+    16      8     bytes    xxh64(encrypted_data)
+
+The ObjHeader is stored unmodified inside the pack blob; the pack layer treats blobs
+as opaque bytes and does not rewrite the header.  The SHA256 content-addressed
+``pack_id`` handles pack-level integrity; the xxh64 fields come from the existing
+RepoObj wire format and are left as-is.  They remain useful for the keyless recovery
+scan (see :ref:`pack-recovery`) where AEAD decryption is not available.
```
ThomasWaldmann: In the encryption mode docs, we recommend that everybody is at least using an "authenticated" mode (and not "none"). So we could use the crypto authentication usually.

Those people who insist on using the least safe mode would still have the full pack sha256, but would lose a full pack if something goes wrong.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+    :figwidth: 100%
+    :width: 100%
+    :figclass: figure-padded
+
+    Pack file binary layout: 9-byte file header followed by contiguous
+    length-prefixed blobs, each containing an ObjHeader and the encrypted payload.
+
+There is no trailing table of contents.  The ``index/`` namespace (see
+:ref:`pack-index-namespace`) is the sole authoritative source of chunk-to-location
+mappings for normal operation.
```
ThomasWaldmann: Do we need to document things that do not and shall not exist?

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+    pack_id = SHA256(pack_file_bytes)
+
+This makes pack files content-addressed: the stored filename is a commitment to the
+content.  ``borg check`` can detect silent corruption by recomputing the digest and
+comparing it to the filename without decrypting any blob.
+
+Namespace
+~~~~~~~~~
+
+Pack files are stored under the ``packs/`` namespace in borgstore, using a two-level
```
ThomasWaldmann: Guess we could change to single level for almost everybody.

256 dirs * 1000 files per dir * 1GB per file = 256TB already.

And we could use even more files per dir and even a bit larger files up to 4GB.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+The nesting depth is controlled by the ``packs/`` entry in the repository's
+``levels_config``, the same mechanism used by the ``data/`` namespace.
+
+
+.. _pack-phase1:
+
+Phase 1 Implementation (N=1)
+-----------------------------
+
+The initial implementation puts one blob per pack file.  Assembly is simpler: no
```
ThomasWaldmann: Double space after dot is for old people I heard. :-)

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+.. _pack-write-order:
+
+Write Order and Crash Safety
+-----------------------------
+
+Pack data must reach stable storage before any index or manifest entry references
+it.  The required write order is:
+
+1. Write the pack file to ``packs/<pack_id>``.
+2. ``fsync`` the pack file and its containing directory.
```
ThomasWaldmann: As we use borgstore, the low level stuff is implemented there and borg just does `Store.store(name, value)`.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+3. Write the index piece file to ``index/<index_id>`` (see :ref:`pack-index-namespace`).
+4. ``fsync`` the index piece file.
```
ThomasWaldmann: same here

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+
+A crash between steps 1 and 3 leaves orphan pack files in ``packs/``.  No archive
+references these chunks; ``borg compact`` removes them on the next run.
+
+A crash between steps 3 and 5 leaves an index piece covering packs not yet committed
+to any archive.  The extra index entries point to valid, fully-written pack data; they
+are harmless and will be cleaned up by the next ``borg compact``.
+
+A crash after step 5 cannot leave the repository in an inconsistent state.  The
+manifest is the commit point: data that the manifest does not reference is unreachable
```
ThomasWaldmann: data that is not referenced by an archive pointer ...

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+and treated as garbage by ``borg compact``.
+
+Deletion is soft: ``repository.delete()`` does not remove the pack file.  The pack
+stays on disk until ``borg compact`` confirms via mark-and-sweep that none of its
+blobs appear in any archive, then removes the whole file.  The same approach carries to N>1: there is no way to remove one blob from a pack
+without rewriting the whole file, so soft-delete is the only option.
```
ThomasWaldmann: the only parts of the code that would do any sort of chunk or pack deletion are borg check and borg compact. Everything else is just adding stuff if it is not there already.

if borg compact decides it is finished with a pack file (because it either had no used chunk left anyway or because it got compacted and all used chunks have already been written to a new pack and new index), why do we need soft deletion of the pack? guess we can really get rid of it using the real deletion.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+
+
+.. _pack-index-namespace:
+
+Index Namespace
+---------------
+
+Borg does not embed a table of contents inside each pack file.  Chunk-to-location
```
ThomasWaldmann: Don't document what doesn't exist nor is planned.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+mappings are stored as a separate set of encrypted piece files under the ``index/``
+namespace.
+
+Each piece file covers the packs written in one backup session.  Its name is the
```
ThomasWaldmann: maybe "partial index file" is a better term?

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+
+Piece files are write-once.  A session appends new piece files; existing files are
+never modified.  On repository open, the client downloads all files under ``index/``,
+decrypts them, and merges the results into the in-memory PackIndex (a ``borghash``
+``HashTableNT`` keyed on ``chunk_id``).  The merge is commutative and idempotent;
+piece file order does not matter.
+
+``borg compact`` consolidates all existing piece files into a single replacement file
+that covers only live chunks, writes it to ``index/``, and removes the files it
+supersedes.  This keeps the namespace small and open-time merge cost bounded.
```
ThomasWaldmann: In the end, we rather want medium-sized partial indexes and avoid creation of all-in-one indexes.

If we create big single index files, we will cause a lot of traffic due to cache invalidation on other clients.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+.. _pack-recovery:
+
+Recovery Path
+-------------
+
+When ``borg check --repair`` detects a missing or incomplete PackIndex it rebuilds
+it by forward-scanning all pack files in ``packs/``.
+
+The 4-byte ``blob_len`` prefix before each blob makes the scan self-contained: no
+prior knowledge of blob sizes or count is required.  The algorithm for one pack file::
```
ThomasWaldmann: Maybe don't put detailled algorithms into the docs, they will change in real code and get outdated anytime soon.

docs/internals/packs.rst
```diff
@@ -0,0 +1,318 @@
...
+                           pack_id = pack_id,
+                           offset  = pos + 4,
+                           length  = blob_len)
+        pos += 4 + blob_len
+
+The ``offset`` recorded in the rebuilt index points past the ``blob_len`` prefix,
+directly at the ObjHeader, consistent with normal PackIndex entries.
+
+Reconstructing ``chunk_id`` values requires the repository key because the chunk ID
+is a keyed MAC of the plaintext data (``id_hash(plaintext_data)``).  Without the key,
```
ThomasWaldmann: I guess we have the chunk_id inside that "meta" part, so we do not have to read all data just to recomputed that when rebuilding the index.

ThomasWaldmann (changes requested):
Here is some feedback. If there are questions or other ideas, we can either discuss here or on IRC.

docs/internals/pack-layout.png
mr-raj12: Removed blob_len and added BORGPACK magic at the start of every blob as discussed

docs/internals/pack-objheader.png
mr-raj12: xxh64 is gone now

docs/internals/pack-write-order.png
mr-raj12: Fixed, now using "archive pointer" throughout

docs/internals/packs.rst
mr-raj12: have tried a new intro this time

docs/internals/packs.rst
mr-raj12: changed as AEAD authenticates each blob and sha256 of pack content covers the pack

docs/internals/packs.rst
mr-raj12: removed

docs/internals/packs.rst
mr-raj12: changed to single level

docs/internals/packs.rst
mr-raj12: fixed :-)

docs/internals/packs.rst
mr-raj12: Changed to "via borgstore"

docs/internals/packs.rst
mr-raj12: fixed

docs/internals/packs.rst
mr-raj12: corrected

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
-(always 24 bytes) plus ``encrypted_meta`` plus ``encrypted_data``.  It satisfies::
+``chunk_id`` is the keyed MAC of the plaintext data (``id_hash(plaintext_data)``).
+Storing it in the unencrypted header lets a scanner rebuild the
+``chunk_id → location`` index without decrypting any blob.
 
-    blob_len == 24 + meta_size + data_size
+A reader locates the next blob by advancing::
 
-The ObjHeader is the existing 24-byte structure shared by all Borg repo objects::
+    next_blob_offset = current_blob_offset + 49 + meta_size + data_size
```
ThomasWaldmann: you could give that 49bytes part a name and then say xxx_size here.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
-Each blob is stored as a 4-byte length prefix followed by the full repo object::
+Each blob is a self-contained unit::
 
     Offset (relative to blob start)  Size        Type     Field
     --------------------------------  ----------  -------  -----
-    0                                 4           uint32le blob_len
-    4                                 24          bytes    ObjHeader
-    4 + 24                            meta_size   bytes    encrypted_meta
-    4 + 24 + meta_size                data_size   bytes    encrypted_data
+    0                                 8           bytes    Magic: ASCII b"BORGPACK"
```
ThomasWaldmann: Considering it is in each entry maybe we shouldn't call the magic ...PACK.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
+    8                                 1           uint8    Format version: 0x01
+    9                                 32          bytes    chunk_id
```
ThomasWaldmann: Stuff like chunk_id (or "id") should be in meta.

Usually, we will have it in the index also and in the rare case we need to rebuild the index, we can also read/decrypt the meta part (it's tiny).

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
+    41                                4           uint32le meta_size
+    45                                4           uint32le data_size
+    49                                meta_size   bytes    encrypted_meta
+    49 + meta_size                    data_size   bytes    encrypted_data
 
-``blob_len`` is the total byte count of the repo object that follows: ObjHeader
-(always 24 bytes) plus ``encrypted_meta`` plus ``encrypted_data``.  It satisfies::
+``chunk_id`` is the keyed MAC of the plaintext data (``id_hash(plaintext_data)``).
```
ThomasWaldmann: maybe rather call it ID hash always.

if the repo is not encrypted, iirc it is sha256(content).

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
 
-    Offset  Size  Type     Field
-    ------  ----  -------  -----
-    0       4     uint32le meta_size
-    4       4     uint32le data_size
-    8       8     bytes    xxh64(encrypted_meta)
-    16      8     bytes    xxh64(encrypted_data)
+The per-blob magic limits corruption blast radius: a damaged blob causes the
+scanner to lose at most one chunk. Once it finds the next ``BORGPACK`` sequence
+it resumes.
```
ThomasWaldmann: It is specifically about damaged length informations. If the corruption is elsewhere, we don't need to scan for the magic.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
+The pack ID equals the ``chunk_id`` of the blob it contains::
 
-    pack_id = SHA256(pack_file_bytes)
+    pack_id = chunk_id
 
-This makes pack files content-addressed: the stored filename is a commitment to the
-content.  ``borg check`` can detect silent corruption by recomputing the digest and
-comparing it to the filename without decrypting any blob.
+Since ``chunk_id`` is a keyed MAC of the plaintext, the filename commits to the
+content. ``borg check`` can detect silent corruption without decrypting any blob.
```
ThomasWaldmann: No.

The intention is that the low-level part of borg check (the "repository check", not the "archives check") checks the pack id (not: chunk id) against the pack name, both must match.

ThomasWaldmann: At N=1, we'll use chunkid == packid, but in the general case, it must use the packid for that check, not the chunkid.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
+single directory level keyed on the first byte of the pack ID (hex-encoded)::
 
     packs/
       00/ .. ff/
-        00/ .. ff/
-          <pack_id_hex>
+        <pack_id_hex>
 
 The nesting depth is controlled by the ``packs/`` entry in the repository's
 ``levels_config``, the same mechanism used by the ``data/`` namespace.
```
ThomasWaldmann: that was renamed to "config" yesterday.

strictly taken, you don't even need to talk about nesting details here, that will be handled by borgstore.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
-
-    pack_id = chunk_id   # Phase 1 only
+Pack Index Entry
+----------------
 
-Computing SHA-256 over the pack content is therefore unnecessary.  The pack for a
-given chunk is always at::
+Each pack contains one blob. The pack for a given chunk is always at::
 
     packs/<hex(chunk_id)>
```
ThomasWaldmann: pack_id (then you don't need to change it later)

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
 
-where the chunk ID is the same keyed MAC (``id_hash(plaintext_data)``) used today
-for objects in ``data/``.
+A PackIndex entry records the pack and byte range needed to read a blob::
```
ThomasWaldmann: IIRC it is called ChunkIndex right now.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
-                  offset  = 13,        # 9-byte file header + 4-byte blob_len
-                  length  = blob_len)  # value read from bytes 9-12 of the pack file
-
-.. note::
-
-    The N value is configurable at ``borg repo-create`` time.  Expanding to N>1 in
-    a follow-on change requires no modification to the pack file format: the file
-    header and per-blob layout are identical.  Only the pack assembly, PackIndex
-    update, and pack ID computation logic changes.
+    chunk_id  →  (pack_id = chunk_id, offset, length)
```
ThomasWaldmann: For N=1 you do not need that yet. Always same offset and just the whole object.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
 
+``offset`` is the position of the blob's size fields within the pack file.
+``length`` covers those size fields plus the encrypted meta and data payloads.
```
ThomasWaldmann: Hmm, just point to the magic and always read the whole header also? Could be used for an addtl. sanity check.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
 
-Borg does not embed a table of contents inside each pack file.  Chunk-to-location
-mappings are stored as a separate set of encrypted piece files under the ``index/``
-namespace.
+Chunk-to-location mappings are stored as a separate set of encrypted partial index
+files under the ``index/`` namespace.
 
-Each piece file covers the packs written in one backup session.  Its name is the
-SHA-256 digest of its own content::
+Each partial index file covers the packs written in one backup session. Its name is
```
ThomasWaldmann, line 146: Can be done like this initially. But in the worst case (== every first backup of a new data set), that index file might be rather big.

docs/internals/packs.rst
```diff
@@ -6,234 +6,158 @@
...
 
-``borg compact`` consolidates all existing piece files into a single replacement file
-that covers only live chunks, writes it to ``index/``, and removes the files it
-supersedes.  This keeps the namespace small and open-time merge cost bounded.
+``borg compact`` rewrites the ``index/`` namespace: it identifies live chunks via
+mark-and-sweep, consolidates the surviving mappings into medium-sized replacement
+files (targeting roughly 10–100 packs per file), and removes the files it supersedes.
+Medium-sized files keep the open-time merge cost bounded while avoiding the
+cache-invalidation traffic on other clients that a single all-in-one index would
+cause.
```
ThomasWaldmann, line 171: Maybe the same code that produces these medium sized index files here can also be used for borg create.

docs/internals/packs.rst
```diff
@@ -247,55 +171,22 @@ Recovery Path
...
-directly at the ObjHeader, consistent with normal PackIndex entries.
-
-Reconstructing ``chunk_id`` values requires the repository key because the chunk ID
-is a keyed MAC of the plaintext data (``id_hash(plaintext_data)``).  Without the key,
-a structural scan can still verify magic bytes, version, blob boundaries, and
-ObjHeader xxh64 hashes, but cannot produce a usable ``chunk_id → location`` mapping.
+Each blob's unencrypted header supplies the ``BORGPACK`` magic (for re-sync after
+corruption), the ``chunk_id``, and the size fields needed to locate the next blob.
+The scan produces a complete ``chunk_id → (pack_id, offset, length)`` mapping
+without decrypting any blob and without the repository key.
```
ThomasWaldmann: Losing the index and having to rebuild it should be a rare event. Thus, we do not optimize for it (avoiding decryption) and having the chunk_id external of meta is not a design requirement.

ThomasWaldmann (changes requested):
Good overall, some feedback.

The question is always: document the end goal or the target for the next step?

docs/internals/packs.rst
ThomasWaldmann: Hmm, it is really not in meta (yet).

Guess because the object name was the chunk id, so there was no need yet.

ThomasWaldmann: The reason I prefer meta over header: meta is a variable size object, changing its content does not require changing the repository format.

docs/internals/packs.rst
mr-raj12: BLOB_HEADER_SIZE = 8 + 1 + 32 + 4 + 4 = 49
next_blob_offset = current_blob_offset + BLOB_HEADER_SIZE + meta_size + data_size
would be fine?

docs/internals/packs.rst
mr-raj12: maybe  BORGOBJ since these are repo objects, or anything else as name?

docs/internals/packs.rst
mr-raj12: sure

docs/internals/packs.rst
mr-raj12: okay, will go as re-sync via magic is only needed when a length field is corrupted

docs/internals/packs.rst
mr-raj12: okay, just chunk_id → pack_id as of now

docs/internals/packs.rst
mr-raj12: header only:
[BORGPACK: 8B][version: 1B][chunk_id: 32B][meta_size: 4B][data_size: 4B][enc_meta][enc_data]


if we only add in meta then, there's a circular dependency in the encryption: `key.py` uses chunk_id as the AEAD AAD for both meta and data (cipher.encrypt(..., aad=id)). Decrypting meta requires chunk_id as AAD. If chunk_id is inside the meta ciphertext, we will need it to decrypt but can't read it without decrypting first.


maybe keeping the chunk_id in the unencrypted header and also write it into meta (redundant but no format change needed later)

header + meta:
[BORGPACK: 8B][version: 1B][chunk_id: 32B][meta_size: 4B][data_size: 4B][enc_meta][enc_data]
                                                                            ^ meta dict also contains chunk_id



or change pack blob encryption to use something other than chunk_id as AEAD AAD (with pack-level nonce or sequence number), so meta is decryptable without it.

docs/internals/packs.rst
ThomasWaldmann: Yes, BORGOBJ ~ RepoObj.

docs/internals/packs.rst
ThomasWaldmann: If the thing is a BORGOBJ / RepoObj, then that is the REPOOBJ_HEADER?

mr-raj12: true

docs/internals/packs.rst
ThomasWaldmann: will have to look at that tomorrow. for now, duplicate the id hash to the header and in meta.

mr-raj12: okay

docs/internals/packs.rst
```diff
@@ -22,7 +22,7 @@ one-per-pack.
 Pack File Format
 ----------------
 
-There is no separate file header. Each blob starts with an 8-byte ``BORGPACK``
+There is no separate file header. Each blob starts with an 8-byte ``BORGOBJ``
```
ThomasWaldmann: That is 7 bytes. Just leave away, everybody can count.

ThomasWaldmann: Or make it BORG_OBJ

docs/internals/packs.rst
```diff
@@ -33,30 +33,39 @@ Each blob is a self-contained unit::
...
 
-    next_blob_offset = current_blob_offset + 49 + meta_size + data_size
+    next_blob_offset = current_blob_offset + REPOOBJ_HEADER_SIZE + meta_size + data_size
 
-The per-blob magic limits corruption blast radius: a damaged blob causes the
-scanner to lose at most one chunk. Once it finds the next ``BORGPACK`` sequence
-it resumes.
+The per-blob magic limits the blast radius of corrupted length fields: if
+``meta_size`` or ``data_size`` is damaged, the scanner loses at most one blob.
+Once it finds the next ``BORGOBJ`` sequence it resumes. Other corruption
```
ThomasWaldmann: Try not to repeat immediate values so often.

Like 8 or "BORG_OBJ". Rather give them a name (obj_magic_len, obj_magic) and refer to them by name, so we do not have to edit N places if we change something.

docs/internals/packs.rst
```diff
@@ -33,30 +33,39 @@ Each blob is a self-contained unit::
...
+(payload bit flips) is caught by AEAD on that blob without losing position.
 
 Blobs follow one another contiguously with no padding::
 
-    [BORGPACK_0: 8B][0x01: 1B][chunk_id_0: 32B][meta_size_0: 4B][data_size_0: 4B][encrypted_meta_0][encrypted_data_0]
-    [BORGPACK_1: 8B][0x01: 1B][chunk_id_1: 32B][meta_size_1: 4B][data_size_1: 4B][encrypted_meta_1][encrypted_data_1]
+    [BORGOBJ_0: 8B][0x01: 1B][chunk_id_0: 32B][meta_size_0: 4B][data_size_0: 4B][encrypted_meta_0][encrypted_data_0]
```
ThomasWaldmann: See previous comment.

OBJ_MAGIC | version = 0x01 | id_hash = ... | meta_size = ... | data_size = ... | encrypted_meta | encrypted_data

docs/internals/packs.rst
```diff
@@ -66,7 +75,7 @@ The pack ID equals the ``chunk_id`` of the blob it contains::
 
     pack_id = chunk_id
 
-Since ``chunk_id`` is a keyed MAC of the plaintext, the filename commits to the
+Since ``chunk_id`` is the ID hash of the plaintext, the filename commits to the
```
ThomasWaldmann, line 78: Using id_hash (or obj_id?) instead of chunk_id would be even better:

if a file is cut into pieces, we get chunks. if a archive metadata stream is cut into pieces, we also get chunks.

but the main archive metadata object is not a chunk, it's the whole thing, so chunk_id is misleading for that.

ThomasWaldmann: We can also rename that later. Small steps...

docs/internals/packs.rst
mr-raj12: renamed to BORG_OBJ

docs/internals/packs.rst
mr-raj12: now used OBJ_MAGIC throughout

docs/internals/packs.rst
mr-raj12: Updated  to the pipe-separated notation using OBJ_MAGIC as the field name

docs/internals/packs.rst
mr-raj12: leaving that for a follow up as of now

ThomasWaldmann (approved):
LGTM

ThomasWaldmann:
Merge or do you have anything else for this?

mr-raj12:
> Merge or do you have anything else for this?

Can merge this , I will add the graphics later on

PR #9675, opened 2026-05-29
testsuite: add UpgraderFrom12To20 unit tests, refs #9556

mr-raj12:
## Description

Unit tests for `UpgraderFrom12To20`, refs #9556.

`UpgraderFrom12To20` (`legacy/upgrade.py`) shipped in earlier phases without any test coverage. This PR adds `src/borg/testsuite/legacy_upgrade_test.py` covering all three public methods.

`src/borg/testsuite/legacy_upgrade_test.py` -- 20 tests for `UpgraderFrom12To20`:

`upgrade_item` (8 tests):
- regular file passes through unchanged
- whitelist strips legacy-only keys (`acl`, `chunks_healthy`, `hardlink_master`)
- `user`/`group` set to `None` are dropped
- symlink `source` renamed to `target`
- hardlink master gets `hlid` assigned and `hardlink_master` key removed
- hardlink slave resolves `hlid` and reuses chunks via `cache.reuse_chunk`
- master and slave share the same `hlid`
- `REQUIRED_ITEM_KEYS` always present in output

`upgrade_archive_metadata` (9 tests):
- `cmdline` list becomes `command_line` string
- `recreate_cmdline` list becomes `recreate_command_line` string
- `time` gets `+00:00` UTC offset appended
- 4-tuple `chunker_params` gets `buzhash` prefix
- 5-tuple `chunker_params` passes through unchanged
- rechunking override replaces stored `chunker_params`
- recreate fields (`recreate_source_id`, `recreate_args`, `recreate_partial_chunks`) dropped
- `tags` always set to `[]`
- missing optional attrs absent from result

`upgrade_compressed_chunk` (3 tests):
- ZLIB-legacy data detected and `ctype` promoted to `ZLIB.ID`
- non-ZLIB chunks have two prefix bytes stripped
- `ObfuscateSize` big-endian `csize` header re-parsed correctly; inner payload and padding preserved

Also fixes `src/borg/compress.pyi`: `CompressorBase.ID` was typed as `bytes` but is `int` at runtime in every subclass.

No production code changes.

Refs #9556

## Checklist

- [X] New code has tests and docs where appropriate


src/borg/compress.pyi
```diff
@@ -10,7 +10,7 @@ class Compressor:
     def detect(data: bytes) -> Any: ...
 
 class CompressorBase:
-    ID: bytes = ...
+    ID: int = ...
```
ThomasWaldmann, line 13: Good catch!

Yeah, it's an int, but stored into a single byte. Guess that's where the confusion came from.

src/borg/testsuite/legacy_upgrade_test.py
```diff
@@ -0,0 +1,233 @@
...
+
+def test_master_and_slave_share_the_same_hlid():
+    u, _, _ = _upgrader()
+    master = _item(hardlink_master=True, chunks=[[CHUNK_ID, CHUNK_SIZE]])
+    master_result = u.upgrade_item(item=master)
+
+    slave = _item(path="dir/link2", source="dir/file")
+    slave_result = u.upgrade_item(item=slave)
+
+    assert master_result.hlid == slave_result.hlid
```
ThomasWaldmann: these 2 tests could be merged into 1 test.

src/borg/testsuite/legacy_upgrade_test.py
```diff
@@ -0,0 +1,233 @@
...
+    u, _, _ = _upgrader()
+    meta, out = u.upgrade_compressed_chunk({}, data)
+    assert meta["ctype"] == 0x01
+    assert meta["clevel"] == 0xFF
+    assert out == payload
+    assert meta["csize"] == len(payload)
+
+
+def test_obfuscate_old_big_endian_csize_is_upgraded():
+    # Borg 1.x ObfuscateSize used big-endian csize; borg 2 uses little-endian.
```
ThomasWaldmann: That's not quite the point.

borg 1.x used a fixed format (and big endian, curiously).

borg2 has the payload size and compressed size in the metadata (a msgpacked dict, we do not need to care for endianness).

ThomasWaldmann (changes requested):
Thanks for adding these.

BTW, it's not correct there were no test for the upgrader, there is a quite big and quite practical one inside the `transfer_cmd` tests, do a real repository upgrade from 1.2 to 2.

src/borg/testsuite/legacy_upgrade_test.py
mr-raj12: fixed

src/borg/testsuite/legacy_upgrade_test.py
mr-raj12: merged into single test_hardlink_slave_gets_master_hlid_and_chunks

ThomasWaldmann (approved):
LGTM.

PR #9676, opened 2026-05-30
Support rest http stdio

ThomasWaldmann:


ThomasWaldmann:
@mr-raj12 ^ I just switched the remote archiver tests to use `rest:///` URLs (talking http over stdio to a borgstore-server-rest process, locally).

In practice the same can be done via ssh to a remote server, replacing the legacy "RPC via ssh" remoting (in `remote.py`) by `borgstore`.

PR #9677, opened 2026-05-30
repoobj: remove xxh64 checksums from blob header

mr-raj12:
## Description

The `RepoObj` blob header previously had two xxh64 checksums (meta and data). `check_object()` used them to detect byte-level corruption server-side without needing the borg key.

With SHA-256-named pack files, the server verifies a whole pack via its filename hash. Client-side AEAD decryption catches byte corruption too. The xxh64 checks are redundant given either of those.

Header shrinks from 24 bytes (`<II8s8s`) to 8 bytes (`<II`) per object.

Changes:
- `repoobj.py`: drop `xxhash` import, remove `meta_hash`/`data_hash` from struct and namedtuple, simplify `format()`.
- `repository.py`: drop `xxhash` import, remove the two hash-check branches from `check_object()`. Size/truncation checks remain.
- `check_cmd_test.py`: `test_verify_data` no longer needs the `fake_xxh64` workaround. `test_corrupted_file_chunk` now expects `--repository-only` to pass and `--verify-data` to catch corruption.
- `repository_test.py`, `legacyrepository_test.py`: drop `xxhash` import, update `fchunk()` to pack only sizes.

refs #8572

## Checklist

### DIFF
--- a/src/borg/testsuite/archiver/check_cmd_test.py
+++ b/src/borg/testsuite/archiver/check_cmd_test.py
@@ -362,50 +362,34 @@ def test_verify_data(archivers, request, init_args):
     if archiver.get_kind() != "local":
         pytest.skip("only works locally, patches objects")
 
-    # it's tricky to test the cryptographic data verification, because usually already the
-    # repository-level xxh64 hash fails to verify. So we use a fake one that doesn't.
-    # note: it only works like tested here for a highly engineered data corruption attack,
-    # because with accidental corruption, usually already the xxh64 low-level check fails.
-    def fake_xxh64(data, seed=0):
-        # xxhash.xxh64.digest() returns -> bytes
-        class FakeDigest:
-            def digest(self):
-                return b"fakefake"
+    check_cmd_setup(archiver)
+    shutil.rmtree(archiver.repository_path)
+    cmd(archiver, "repo-create", *init_args)
+    create_src_archive(archiver, "archive1")
+    archive, repository = open_archive(archiver.repository_path, "archive1")
+    with repository:
+        for item in archive.iter_items():
+            if item.path.endswith(src_file):
+                chunk = item.chunks[-1]
+                data = repository.get(chunk.id)
+                data = data[0:123] + b"x" + data[123:]
+                repository.put(chunk.id, data)
+                break
 
-        return FakeDigest()
+    # the normal archives check does not read file content data.
+    cmd(archiver, "check", "--archives-only", exit_code=0)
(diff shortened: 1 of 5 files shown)


src/borg/testsuite/archiver/check_cmd_test.py
```diff
@@ -427,13 +411,16 @@ def test_corrupted_file_chunk(archivers, request, init_args):
                 repository.put(chunk.id, data)
                 break
 
-    # the normal check checks all repository objects and the xxh64 checksum fails.
-    output = cmd(archiver, "check", "--repository-only", exit_code=1)
-    assert f"{bin_to_hex(chunk.id)} is corrupted: data does not match checksum." in output
+    # repository-level check catches size/truncation errors but not byte-level corruption.
+    cmd(archiver, "check", "--repository-only", exit_code=0)
```
ThomasWaldmann: Guess this comment and call should be rather removed.

We WILL be able to detect corruption with the sha256-named packs, but we can't test that like done here.

src/borg/testsuite/legacyrepository_test.py
```diff
@@ -75,7 +73,7 @@ def get_path(repository):
 
 def fchunk(data, meta=b""):
     # Create a raw chunk that has a valid RepoObj layout but does not use encryption or compression.
-    hdr = RepoObj.obj_header.pack(len(meta), len(data), xxh64(meta).digest(), xxh64(data).digest())
```
ThomasWaldmann, line 78: Is this the early borg2 "v2" repo format I just removed recently?

I also already wondered why we have xxh64 in legacyrepository - the borg 1.x repository format does not use xxh64, just crc32.

ThomasWaldmann: So, the borg 1.x stuff should use RepoObj1 rather than RepoObj.

ThomasWaldmann (changes requested):

src/borg/testsuite/legacyrepository_test.py
ThomasWaldmann: Let me fix that after this PR, do not fix it here (this was a pre-existing issue and not really in scope here).

src/borg/testsuite/archiver/check_cmd_test.py
mr-raj12: fixed

ThomasWaldmann (approved):
LGTM.

PR #9684, opened 2026-05-30
repoobj: add OBJ_MAGIC blob header for pack file format

mr-raj12:
## Description

`RepoObj` blob header grows from 8 bytes (`<II`) to 17 bytes (`<8sBII`): 8-byte magic (`BORG_OBJ`), 1-byte version, then the existing meta/data sizes.

The magic is a re-sync point. If a pack is corrupted, borg scans forward for the next `BORG_OBJ` to find the next valid blob. The version byte lets borg reject unknown formats rather than misreading them.

This is the N=1 pack format (one chunk per pack file). No chunk id in the header yet -- for N=1 the pack filename is the chunk id (SHA-256 hex), so it is implicit. Chunk id in the header comes later when chunk id != pack id.

Changes:
- `repoobj.py`: add `OBJ_MAGIC`/`OBJ_VERSION` constants, extend struct to `<8sBII`, validate magic and version in `format()`, `parse()`, `parse_meta()`, `extract_crypted_data()`.
- `repository.py`: import `OBJ_MAGIC`/`OBJ_VERSION`, update `check_object()` to check magic/version before sizes, fix `get()` partial-read path for 17-byte header.
- `repository_test.py`: import `OBJ_MAGIC`/`OBJ_VERSION`, update `fchunk()` to include magic and version, fix `pchunk()` slice to `[2:4]`, update `test_read_data()`.

Legacy repo tests (`legacyrepository_test.py`) were decoupled from `RepoObj` in #9687 and #9689 and are not touched here.

refs #8572

## Checklist

### DIFF
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -16,7 +16,7 @@ from .helpers import bin_to_hex, hex_to_bin
 from .storelocking import Lock
 from .logger import create_logger
 from .manifest import NoManifestError
-from .repoobj import RepoObj
+from .repoobj import RepoObj, OBJ_MAGIC, OBJ_VERSION
 
 logger = create_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


ThomasWaldmann (commented):
Looks good.

Next after this, I guess you need to decouple the legacyrepository_test yourself (can't do that if you keep changing it although that test should only test legacy format repos).

The legacy repo (borg 1.x) shall use RepoObj1 format, not RepoObj. It can still use fchunk ("format a chunk") and p(d?)chunk ("parse a chunk"), but needs to be adapted so it works with the old format.

mr-raj12:
Thanks! Understood on legacyrepository_test, I'll fix fchunk/pchunk there to use RepoObj1.obj_header in the next PR.
Can this one be merged as-is, or should the fix come first?

ThomasWaldmann:
If you decouple the legacy tests first, you can remove the changes to the legacy test file from this one.

src/borg/repository.py
```diff
@@ -302,12 +302,17 @@ def check_object(obj):
...
-                if hdr.data_size != len(data):
-                    log_error("data size incorrect.")
+                if hdr.magic != OBJ_MAGIC:
+                    log_error("invalid object magic.")
+                elif hdr.version != OBJ_VERSION:
+                    log_error(f"unsupported object version: {hdr.version}.")
+                else:
+                    meta = obj[hdr_size : hdr_size + hdr.meta_size]
+                    if hdr.meta_size != len(meta):
+                        log_error("metadata size incorrect.")
```
ThomasWaldmann: maybe rather say "... size mismatch".

in such a case, we do not know whether meta_size is wrong (corrupted) or meta is wrong (e.g. truncated).

src/borg/repository.py
```diff
@@ -302,12 +302,17 @@ def check_object(obj):
...
+                    data = obj[hdr_size + hdr.meta_size : hdr_size + hdr.meta_size + hdr.data_size]
+                    if hdr.data_size != len(data):
+                        log_error("data size incorrect.")
```
ThomasWaldmann: similar here.

ThomasWaldmann (changes requested):

src/borg/repository.py
mr-raj12: okay

ThomasWaldmann:
LGTM!

Issue #8036, opened 2024-01-08
add another hmac to the borg key?


ThomasWaldmann:
When looking at #8035, i noticed borg does an hmac check on the decrypted key.

But to get that, it uses the key derived from the user's passphrase *and* the encrypted key material from the borg key file.

If the hmac check fails:
- it could be a wrong passphrase fed into the kdf leading to a wrong key decryption key
- it could be wrong encrypted key material from the keyfile (in #8035, the user typed that in from a raw printout of the keyfile contents)

So, that is not super helpful to fix whatever needs fixing, passphrase or key.

Maybe we should add another hmac computed over the encrypted key in borg2, so we can detect if there is an error in the keyfile? also: if we know the keyfile is correct, we could also be sure that the passphrase is wrong if the other hmac doesn't match.

ThomasWaldmann:
Ehrm, guess we need to use a hash for that, not an hmac.

Because for the hmac, we need a key. The key would be derived from passphrase. And due to that, we run into same problem: we won't know if passphrase is wrong or key is corrupted if the check fails.

ThomasWaldmann:
Related idea: use AE / AEAD for the borg key encryption?

ThomasWaldmann:
# Analysis by Claude Opus 4.6 (Thinking):

## Problem Statement

When borg decrypts a key file, it can detect that *something is wrong* but cannot tell the user **what** is wrong:

1. **Wrong passphrase** → KDF derives the wrong decryption key → HMAC/AEAD check fails
2. **Corrupted key material** → correct passphrase, but the ciphertext is mangled → same check fails

The user gets the same generic failure in both cases. This was directly observed in #8035, where a user manually retyped a keyfile from a printout and couldn't determine whether the passphrase or the keyfile content was the problem.

## Current Key Decryption Flow

Two code paths exist in `src/borg/crypto/key.py`:

### Legacy PBKDF2 path (`decrypt_key_file_pbkdf2`)

```python
key = self.pbkdf2(passphrase, encrypted_key.salt, encrypted_key.iterations, 32)
data = AES(key, b"\0" * 16).decrypt(encrypted_key.data)
if hmac.compare_digest(hmac_sha256(key, data), encrypted_key.hash):
    return data
return None
```

The HMAC is computed over the **decrypted** plaintext using the **passphrase-derived** key. Both a wrong passphrase and corrupted ciphertext produce indistinguishable failures.

[...]

PR #9686, opened 2026-05-30
testsuite: add Pbkdf2FileMixin unit tests, refs #9556

mr-raj12:
## Description

Unit tests for `Pbkdf2FileMixin`, refs #9556.

`Pbkdf2FileMixin` (`legacy/crypto/key.py`) is the PBKDF2 + AES-256-CTR layer borg 1.x uses to protect key files on disk. It shipped in earlier phases with no unit tests. This PR adds `src/borg/testsuite/crypto/legacy_key_test.py` with 4 tests covering `encrypt_key_file` and `decrypt_key_file`.

`src/borg/testsuite/crypto/legacy_key_test.py` -- 4 tests for `Pbkdf2FileMixin`:

- encrypt/decrypt round-trip: `encrypt_key_file` -> `decrypt_key_file` recovers the original plaintext
- wrong passphrase returns `None`: PBKDF2 derives a different key, HMAC check fails, `None` returned (not raised)
- tampered HMAC returns `None`: correct passphrase and ciphertext but zeroed-out hash field fails `hmac.compare_digest`
- unsupported version raises `UnsupportedKeyFormatError`: version != 1 is not part of the borg 1.x format

`BORG_TESTONLY_WEAKEN_KDF` is set globally by `conftest.py`, so PBKDF2 runs at 1 iteration and tests are fast.

No production code changes.

Refs #9556

## Checklist

- [X] New code has tests (this PR *is* the tests)

### DIFF
--- a/src/borg/testsuite/crypto/legacy_key_test.py
+++ b/src/borg/testsuite/crypto/legacy_key_test.py
@@ -1,0 +1,34 @@
+"""Tests for borg.legacy.crypto.key (Pbkdf2FileMixin)."""
+
+import pytest
+
+from ...crypto.key import UnsupportedKeyFormatError
+from ...helpers import msgpack
+from ...legacy.crypto.key import KeyfileKey as LegacyKeyfileKey
+
+
+# ── Pbkdf2FileMixin ───────────────────────────────────────────────────────────
+
+
+def test_pbkdf2_encrypt_decrypt_roundtrip():
+    # encrypt_key_file dispatches to encrypt_key_file_pbkdf2; decrypt_key_file
+    # dispatches back — the round-trip must recover the original plaintext
+    key_obj = LegacyKeyfileKey(None)
+    plaintext = b"secret key material"
+    blob = key_obj.encrypt_key_file(plaintext, "correct passphrase", "sha256")
+    assert key_obj.decrypt_key_file(blob, "correct passphrase") == plaintext
+
+
+def test_pbkdf2_wrong_passphrase_returns_none():
+    # a wrong passphrase derives a different key, so the HMAC check fails;
+    # decrypt_key_file signals this by returning None, not by raising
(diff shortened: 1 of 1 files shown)


src/borg/testsuite/crypto/legacy_key_test.py
```diff
@@ -0,0 +1,49 @@
...
+    plain = b"secret key material"
+    salt = b"salt" * 4
+    passphrase = "correct passphrase"
+    key_bytes = LegacyKeyfileKey.pbkdf2(passphrase, salt, 1, 32)
+    data = AES(key_bytes, b"\0" * 16).encrypt(plain)
+    blob = msgpack.packb(
+        {"version": 1, "algorithm": "sha256", "iterations": 1, "salt": salt, "data": data, "hash": b"\x00" * 32}
+    )
+    assert LegacyKeyfileKey(None).decrypt_key_file(blob, passphrase) is None
+
```
ThomasWaldmann: Is this testing an important property of a borg key or rather showing a problematic design?

#8036 ?

mr-raj12: Agreed as that test documents the design limitation in #8036, not an important property. Dropping it and keeping only the round-trip and wrong-passphrase tests.

ThomasWaldmann (approved):
LGTM.

ThomasWaldmann:
Do we already have similar tests for the argon2 kdf?

mr-raj12:
> Do we already have similar tests for the argon2 kdf?

`test_key_file_roundtrip` in `crypto/key_test.py` covers the Argon2 roundtrip. Wrong-passphrase for Argon2 is not tested yet. Should I add it here?

ThomasWaldmann:
yes, please, would be good to have similar tests for both.

PR #9687, opened 2026-05-30
testsuite: decouple legacyrepository_test from RepoObj, refs #8572

mr-raj12:
## Description

`legacyrepository_test` was importing `RepoObj` (borg 2) to build test blobs in `fchunk`/`pchunk`. That coupling was wrong: borg 1.x blobs have a 1-byte compressor type header followed by the payload, not the `Struct("<II")` meta/data header from borg 2.

Remove the `RepoObj` import. `fchunk` now prepends a 1-byte `CNONE` type header (`0x00`) to simulate the generic borg 1.x blob format without real compression. `pchunk` strips that byte. `test_max_data_size` updated to account for the 1-byte overhead.

This way, changes to the borg 2 blob format (e.g. `OBJ_MAGIC` in #9684) no longer touch the legacy test helpers.

refs #8572

## Checklist

### DIFF
--- a/src/borg/testsuite/legacyrepository_test.py
+++ b/src/borg/testsuite/legacyrepository_test.py
@@ -14,7 +14,6 @@ from ..platformflags import is_win32
 from ..legacy.remote import LegacyRemoteRepository, InvalidRPCMethod, PathNotAllowed
 from ..legacy.repository import LegacyRepository, LoggedIO
 from ..legacy.repository import MAGIC, MAX_DATA_SIZE, TAG_DELETE, TAG_PUT, TAG_COMMIT
-from ..repoobj import RepoObj
 from .hashindex_test import H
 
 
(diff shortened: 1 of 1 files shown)


src/borg/testsuite/legacyrepository_test.py
```diff
@@ -72,21 +71,15 @@ def get_path(repository):
 
 
 def fchunk(data, meta=b""):
-    # Create a raw chunk that has a valid RepoObj layout but does not use encryption or compression.
-    hdr = RepoObj.obj_header.pack(len(meta), len(data))
+    # RepoObj1 blobs are raw bytes with no header; meta is unused in the borg 1.x format.
```
ThomasWaldmann: IIRC, not quite true: there is at least a 1-byte type header (except zlib_legacy).

So maybe don't pick the special case "no header" here, but the more generic case "there is some header".

src/borg/testsuite/legacyrepository_test.py
```diff
@@ -72,21 +71,15 @@ def get_path(repository):
...
 
 def pchunk(chunk):
-    # Parse data and meta from a raw chunk made by fchunk.
-    hdr_size = RepoObj.obj_header.size
-    hdr = chunk[:hdr_size]
-    meta_size, data_size = RepoObj.obj_header.unpack(hdr)[0:2]
-    meta = chunk[hdr_size : hdr_size + meta_size]
-    data = chunk[hdr_size + meta_size : hdr_size + meta_size + data_size]
-    return data, meta
+    # RepoObj1 blobs are raw bytes; meta is always empty in the borg 1.x format.
```
ThomasWaldmann: Here, just strip that 1-byte header.

ThomasWaldmann (changes requested):
The idea is:
- fchunk(data, meta) -> chunk_content -> repo put
- repo get -> pchunk -> data, meta

That fchunk/pchunk is just a simplified way to do something close to what borg usually does, but without having to set up compression and encryption.

src/borg/testsuite/legacyrepository_test.py
mr-raj12: fixed using CNONE (ID 0x00) as the 1-byte ctype header now

src/borg/testsuite/legacyrepository_test.py
mr-raj12: done, chunk[1:] now!

ThomasWaldmann (approved):
LGTM.

PR #9689, opened 2026-05-31
legacy repo tests: always format and parse chunk

ThomasWaldmann:
This was missing in #9687 and overlooked in review.


ThomasWaldmann:
@mr-raj12 ^^^

PR #9692, opened 2026-05-31
repoobj, repository: add chunk_id to header, introduce packs/ namespace

mr-raj12:
## Description

`RepoObj` blob header grows from 17 bytes (`<8sBII`) to 49 bytes (`<8sB32sII`): a 32-byte `chunk_id` field between the version byte and the size fields. `REPOOBJ_HEADER_SIZE = 49` names the fixed part.

`chunk_id` is the ID hash of the plaintext data. Putting it in the unencrypted header lets a forward scanner rebuild the `chunk_id -> location` index without touching the key. It is also the additional authenticated data for AEAD, so a plaintext copy in the header is unavoidable -- recovering it from ciphertext requires the key.

Chunks move from `data/` to `packs/`. The new namespace uses one directory level keyed on the first byte of the pack ID (256 subdirs `00/`..`ff/`), versus two levels for `data/`. For the expected object count per pack, one level is enough.

For N=1 packs (one chunk per pack file), `pack_id == chunk_id`. The pack filename is the chunk ID (`id_hash` of the plaintext chunk content).

Changes:
- `repoobj.py`: extend struct from `<8sBII` to `<8sB32sII`, add `chunk_id` to `ObjHeader`, add `REPOOBJ_HEADER_SIZE = 49`, pass `id` into `ObjHeader` in `format()`.
- `repository.py`: add `packs/` to `ns_config` with `{"levels": [1]}`, add packs to permission maps, replace all `data/` keys with `packs/` in `get()`/`put()`/`delete()`/`list()`/`check()`, introduce `pack_id = id` (N=1 invariant), wrap stored blobs in a BORGPACK header (13 bytes: magic + version + blob_len), bump repo version to 4.
- `repository_test.py`: add `chunk_id` parameter to `fchunk()`, fix `pchunk()` slice to `[3:5]`, update `test_read_data()` to pass `H(0)` as `chunk_id`.

refs #8572

## Checklist

### DIFF
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -127,9 +127,9 @@ def with_repository(
             )
 
             with repository:
-                if repository.version not in (3, 4):
+                if repository.version not in (4,):
                     raise Error(
-                        f"This borg version only accepts version 3 or 4 repos for -r/--repo, "
+                        f"This borg version only accepts version 4 repos for -r/--repo, "
                         f"but not version {repository.version}. "
                         f"You can use 'borg transfer' to copy archives from old to new repos."
                     )
@@ -194,10 +194,10 @@ def with_other_repository(manifest=False, cache=False, compatibility=None):
             )
 
             with repository:
-                acceptable_versions = (1,) if v1_legacy else (3,)
+                acceptable_versions = (1,) if v1_legacy else (3, 4)
                 if repository.version not in acceptable_versions:
                     raise Error(
-                        f"This borg version only accepts version {' or '.join(acceptable_versions)} "
+                        f"This borg version only accepts version {' or '.join(str(v) for v in acceptable_versions)} "
                         f"repos for --other-repo."
                     )
                 kwargs["other_repository"] = repository
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -127,9 +127,9 @@ def with_repository(
             )
 
             with repository:
-                if repository.version not in (3,):
+                if repository.version not in (3, 4):
                     raise Error(
-                        f"This borg version only accepts version 3 repos for -r/--repo, "
+                        f"This borg version only accepts version 3 or 4 repos for -r/--repo, "
                         f"but not version {repository.version}. "
                         f"You can use 'borg transfer' to copy archives from old to new repos."
                     )
--- a/src/borg/archiver/_common.py
+++ b/src/borg/archiver/_common.py
@@ -194,7 +194,7 @@ def with_other_repository(manifest=False, cache=False, compatibility=None):
             )
 
             with repository:
-                acceptable_versions = (1,) if v1_legacy else (3, 4)
+                acceptable_versions = (1,) if v1_legacy else (4,)
                 if repository.version not in acceptable_versions:
                     raise Error(
                         f"This borg version only accepts version {' or '.join(str(v) for v in acceptable_versions)} "
--- a/src/borg/repository.py
+++ b/src/borg/repository.py
@@ -17,8 +17,15 @@ from .helpers import bin_to_hex, hex_to_bin
 from .storelocking import Lock
 from .logger import create_logger
 from .manifest import NoManifestError
+from struct import Struct
+
 from .repoobj import RepoObj, OBJ_MAGIC, OBJ_VERSION
 from .crypto.key import is_keyfile
+
+PACK_MAGIC = b"BORGPACK"
+PACK_VERSION = 0x01
+_pack_header = Struct("<8sBI")  # magic(8) + version(1) + blob_len(4)
+PACK_HEADER_SIZE = _pack_header.size  # 13 bytes
 
 logger = create_logger(__name__)
 
@@ -174,7 +181,7 @@ class Repository:
         self._send_log = send_log_cb or (lambda: None)
         self.do_create = create
         self.created = False
-        self.acceptable_repo_versions = (3,)
+        self.acceptable_repo_versions = (4,)
         self.opened = False
         self.lock = None
         self.do_lock = lock
@@ -212,10 +219,10 @@ class Repository:
         self.store.open()
         try:
             self.store.store("config/readme", REPOSITORY_README.encode())
-            self.version = 3
+            self.version = 4
             self.store.store("config/version", str(self.version).encode())
             self.store.store("config/id", bin_to_hex(os.urandom(32)).encode())
-            # we know repo/data/ still does not have any chunks stored in it,
+            # we know repo/packs/ still does not have any chunks stored in it,
             # but for some stores, there might be a lot of empty directories and
             # listing them all might be rather slow, so we better cache an empty
             # ChunkIndex from here so that the first repo operation does not have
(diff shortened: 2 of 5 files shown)


src/borg/repository.py
```diff
@@ -122,6 +122,7 @@ def __init__(
             "data/": {"levels": [data_levels]},
```
ThomasWaldmann: does it still use data/ ?

src/borg/repository.py
```diff
@@ -137,6 +138,7 @@ def __init__(
                 "data": "lrw",
```
ThomasWaldmann: same.

src/borg/repository.py
```diff
@@ -306,6 +309,8 @@ def check_object(obj):
                     log_error("invalid object magic.")
                 elif hdr.version != OBJ_VERSION:
                     log_error(f"unsupported object version: {hdr.version}.")
+                elif hdr.chunk_id != hex_to_bin(info.name):
+                    log_error("chunk_id mismatch in header.")
```
ThomasWaldmann: info.name is the pack id, hdr.chunk_id is the chunk id.

So this is wrong in general and only true now in this simplified N=1 case.

Maybe rather never add code that is wrong in the general case.

src/borg/repository.py
```diff
@@ -384,8 +389,8 @@ def check_object(obj):
                     # add all existing objects to the index.
                     # borg check: the index may have corrupted objects (we did not delete them)
                     # borg check --repair: the index will only have non-corrupted objects.
-                    id = hex_to_bin(info.name)
-                    chunks[id] = init_entry
+                    pack_id = hex_to_bin(info.name)  # N=1: pack_id == chunk_id
+                    chunks[pack_id] = init_entry
```
ThomasWaldmann: Wrong in general, the index into the chunks index is the chunk id.

src/borg/repository.py
```diff
@@ -428,30 +433,31 @@ def list(self, limit=None, marker=None):
...
         result = []
-        infos = self.store.list("data")  # generator yielding ItemInfos
+        infos = self.store.list("packs")  # generator yielding ItemInfos
         while True:
             self._lock_refresh()
             try:
                 info = next(infos)
             except StoreObjectNotFound:
-                break  # can happen e.g. if "data" does not exist, pointless to continue in that case
+                break  # can happen e.g. if "packs" does not exist, pointless to continue in that case
```
ThomasWaldmann, line 463: this is because store.list() only returns an iterator, but does not iterate yet. so if there is no packs/, you'll only notice it after starting to use the iterator.

as there might be lots of objects, there is no better way.

src/borg/repository.py
```diff
@@ -428,30 +433,31 @@ def list(self, limit=None, marker=None):
...
             except StopIteration:
                 break
             else:
-                id = hex_to_bin(info.name)
+                pack_id = hex_to_bin(info.name)  # N=1: pack_id == chunk_id
```
ThomasWaldmann: rather add a line `chunk_id = pack_id` with this comment.

src/borg/repository.py
```diff
@@ -428,30 +433,31 @@ def list(self, limit=None, marker=None):
...
                 if collect:
-                    result.append((id, info.size))
+                    result.append((pack_id, info.size))
```
ThomasWaldmann: guess the idea here is to collect the chunk_ids and the chunk data sizes.

ThomasWaldmann (changes requested):

src/borg/repository.py
mr-raj12: removed

ThomasWaldmann:
"The pack filename is the chunk's SHA-256 hex ID."

That is never true.

For N=1 the pack filename (pack_id) is the chunk_id (id_hash(chunk content), usually not sha256).

For N>1 the pack filename (pack_id) shall be sha256(pack content).

src/borg/repository.py
mr-raj12: removed this check entirely

src/borg/repository.py
mr-raj12: now using pack_id = hex_to_bin(info.name) then chunk_id = pack_id  # N=1: chunk_id == pack_id, and the index is keyed by chunk_id.

src/borg/repository.py
mr-raj12: updated

src/borg/archiver/_common.py
```diff
@@ -127,9 +127,9 @@ def wrapper(self, args, **kwargs):
             )
 
             with repository:
-                if repository.version not in (3,):
+                if repository.version not in (3, 4):
```
ThomasWaldmann: i don't think we'll continue supporting v3 repos?

mr-raj12: okay, updated!

src/borg/archiver/_common.py
```diff
@@ -194,10 +194,10 @@ def wrapper(self, args, **kwargs):
             )
 
             with repository:
-                acceptable_versions = (1,) if v1_legacy else (3,)
+                acceptable_versions = (1,) if v1_legacy else (3, 4)
```
ThomasWaldmann: only 4?

src/borg/repository.py
```diff
@@ -16,8 +16,15 @@
...
 from .logger import create_logger
 from .manifest import NoManifestError
+from struct import Struct
+
 from .repoobj import RepoObj, OBJ_MAGIC, OBJ_VERSION
 
+PACK_MAGIC = b"BORGPACK"
+PACK_VERSION = 0x01
+_pack_header = Struct("<8sBI")  # magic(8) + version(1) + blob_len(4)
+PACK_HEADER_SIZE = _pack_header.size  # 13 bytes
```
ThomasWaldmann: Why did you add this?

src/borg/repository.py
```diff
@@ -298,23 +303,36 @@ def log_error(msg):
...
+            if magic != PACK_MAGIC:
+                log_error("invalid pack magic.")
+                return
+            if version != PACK_VERSION:
+                log_error(f"unsupported pack version: {version}.")
+                return
+            blob = obj[PACK_HEADER_SIZE:]
+            if len(blob) != blob_len:
+                log_error(f"pack blob_len mismatch: header says {blob_len}, actual {len(blob)}.")
+                return
```
ThomasWaldmann: Didn't we want to have the pack just as a sum of the objects? Each object has its own magic and length infos, so it has all the infos we need.

borgstore.defrag can't recompute lengths, just copy blocks or not.

src/borg/archiver/check_cmd.py
```diff
@@ -49,8 +49,8 @@ def do_check(self, args, repository):
             archive_checker = ArchiveChecker()
             try:
                 archive_checker.key = archive_checker.make_key(repository, manifest_only=True)
-            except IntegrityError:
-                pass  # will try to make key later again
+            except Exception as e:
+                logger.debug("early make_key failed, will retry later: %s", e)
```
ThomasWaldmann: Was this change needed? It is somehow out of scope of this PR as you only work on the lower levels.

Also: catching ALL exceptions and outputting them on (not normally visible) debug level is problematic.

src/borg/repository.py
mr-raj12: PACK_HEADER removed

src/borg/archiver/check_cmd.py
mr-raj12: added to fix a pre-existing CI failure on corrupted-manifest test cases although unrelated to pack files.
reverting it from this PR

src/borg/repository.py
```diff
@@ -456,33 +454,34 @@ def list(self, limit=None, marker=None):
...
+                break  # can happen e.g. if "packs" does not exist, pointless to continue in that case
             except StopIteration:
                 break
             else:
-                id = hex_to_bin(info.name)
+                pack_id = hex_to_bin(info.name)
+                chunk_id = pack_id  # N=1: chunk_id == pack_id
                 if collect:
-                    result.append((id, info.size))
+                    result.append((chunk_id, info.size))
```
ThomasWaldmann:
```
chunk_size = info.size  # only correct for N=1
result.append(chunk_id, chunk_size)
```

ThomasWaldmann:
please rebase on current master again.

src/borg/repository.py
mr-raj12: updated

ThomasWaldmann (approved):
LGTM!

PR #9723, opened 2026-06-05
repository: add PackWriter and two-phase chunk index update

mr-raj12:
## Description

`PackWriter` buffers `(chunk_id, cdata)` pairs and flushes them as a pack file via borgstore once `max_count` chunks accumulate. At N=1 (`max_count=1`), `pack_id == chunk_id` and pack files land at `packs/{chunk_id_hex}`. No changes needed to `get()` or `delete()`. `UNKNOWN_INT32 = 0xFFFFFFFF` is the sentinel for pack location fields that are not yet written. 0xFFFFFFFF is above MAX_DATA_SIZE (~20 MB), so it can never collide with a real `obj_offset`. `chunks.add()` writes the placeholder; `update_pack_info()` fills in real values after `flush()`.

`flush()` clears `_pieces` in a `try/finally`. Without this, a store failure would leave the chunk in the buffer and it would get re-bundled with the next chunk, pushing the N>1 code path and writing under a hash-derived key instead of the chunk's own id.

Changes:
- `repository.py`: add `PackWriter`; `put()` delegates to
  `_pack_writer.add()` and returns pack results.
- `constants.py`: add `UNKNOWN_INT32 = 0xFFFFFFFF`.
- `hashindex.pyx`: `add()` uses `UNKNOWN_INT32` placeholders; new
  `update_pack_info()`.
- `hashindex.pyi`: type stub for `update_pack_info()`.
- `cache.py`: `add_chunk()` calls `update_pack_info()` from pack results.
- `archive.py`: `add_reference()` in `rebuild_archives()` does the same.

refs #8572

## Checklist

- [x] PR is against `master`
- [x] Tests pass


src/borg/hashindex.pyx
```diff
@@ -105,6 +109,11 @@ class ChunkIndex(HTProxyMixin, MutableMapping):
         user_flags = value.flags & self.M_USER
         self.ht[key] = value._replace(flags=system_flags | user_flags)
 
+    def update_pack_info(self, key, pack_id, obj_offset, obj_size):
+        """Update the on-disk location fields for an existing entry without changing flags or size."""
+        existing = self[key]
+        self[key] = existing._replace(pack_id=pack_id, obj_offset=obj_offset, obj_size=obj_size)
+
```
ThomasWaldmann: you could give the whole list to this method, so it does the iteration here.

src/borg/repository.py
```diff
@@ -34,6 +34,71 @@ def repo_lister(repository, *, limit=None):
...
+        self.store = store
+        self.max_count = max_count
+        self._pieces = []  # list of (chunk_id, cdata)
+
+    def add(self, chunk_id, cdata):
+        """Buffer a chunk.  Returns flush results if the pack is now full, else []."""
+        self._pieces.append((chunk_id, cdata))
+        if len(self._pieces) >= self.max_count:
+            return self.flush()
+        return []
```
ThomasWaldmann: If we later go to reasonable pack sizes, it will frequently return this empty list (which needs to be garbage collected soon afterwards). Returning None here would be cheaper.

src/borg/repository.py
```diff
@@ -34,6 +34,71 @@ def repo_lister(repository, *, limit=None):
...
+
+    def flush(self):
+        """Write the current pack to the store.
+
+        Returns a list of (chunk_id, pack_id, obj_offset, obj_size) tuples —
+        one entry per chunk that was written.  Returns [] if there was nothing
+        to flush.
+        """
+        if not self._pieces:
+            return []
```
ThomasWaldmann: None here also.

ThomasWaldmann (changes requested):
Looks good overall. Some small optimizations could be done.

Later, when introducing a size limit, it will get a bit more complicated, if we want to absolutely obey that limit and not possibly exceed it by a maximum chunksize in the worst case.

Update: thinking about it, we could also just accept that it is no strict limit. Simpler code. E.g. when setting 50MB as limit, it could be also 70MB.

src/borg/repository.py
ThomasWaldmann: None is a singleton, there is only one of it.

ThomasWaldmann:
range-load using the pack values from index within this PR or in next one?

src/borg/archive.py
```diff
@@ -2015,10 +2015,16 @@ def add_reference(id_, size, cdata):
                 assert cdata is not None
                 pack_id = id_  # only correct for N=1
                 self.chunks[id_] = ChunkIndexEntry(
-                    flags=ChunkIndex.F_USED, size=size, pack_id=pack_id, obj_offset=0, obj_size=0
+                    flags=ChunkIndex.F_USED,
+                    size=size,
+                    pack_id=pack_id,
```
ThomasWaldmann: guess you could use UNKNOWN_256BIT now here, it will be later updated anyway.

ThomasWaldmann: Or UNKNOWN_BYTES32.

src/borg/hashindex.pyx
```diff
@@ -80,7 +82,9 @@ class ChunkIndex(HTProxyMixin, MutableMapping):
             flags = v.flags | self.F_USED
             assert v.size == 0 or v.size == size
         pack_id = key  # N=1: chunk_id == pack_id
-        self[key] = ChunkIndexEntry(flags=flags, size=size, pack_id=pack_id, obj_offset=0, obj_size=0)
+        self[key] = ChunkIndexEntry(
+            flags=flags, size=size, pack_id=pack_id, obj_offset=UNKNOWN_INT32, obj_size=UNKNOWN_INT32
```
ThomasWaldmann: UNKNOWN_BYTES32

mr-raj12:
> range-load using the pack values from index within this PR or in next one?

in the next one

ThomasWaldmann:
OK, so please finish this one.

For the next one:

Guess that will be a small and simple PR, just use the obj_offset and obj_size from the index to do a range-load. As that is always 0 and filesize right now, nothing should break.

Idea for the next one after that:

As you now update the index after you can know the sha256 pack_id, add a env var (SHA256_PACK_ID=1 or so) to disable the pack_id == chunk_id hack and use the real sha256 pack_id from the index. Still stay at max_count=1.

Likely, that will show a lot of problems in the existing code, pointing to all the places that now need to use the index to get the pack_id / work based on packs. Make a priority list of what needs fixing.

The CI could get 1 informative but otherwise ignored job that runs the tests with that env var, so we can see how less and less tests fail while you fix more and more stuff.

src/borg/hashindex.pyx
mr-raj12: done

src/borg/repository.py
mr-raj12: updated

src/borg/archive.py
mr-raj12: updated

src/borg/repository.py
```diff
@@ -296,6 +362,7 @@ def open(self, *, exclusive, lock_wait=None, lock=True):
         # important: lock *after* making sure that there actually is an existing, supported repository.
         if lock:
             self.lock = Lock(self.store, exclusive, timeout=lock_wait).acquire()
+        self._pack_writer = PackWriter(self.store, max_count=1)
         self.opened = True
 
     def close(self):
```
ThomasWaldmann, line 423: you need to add a call to `self._pack_writer.flush()` here to make it work for max_count > 1.

ThomasWaldmann: and also update the index.

mr-raj12: fixed

src/borg/repository.py
```diff
@@ -365,7 +365,20 @@ def open(self, *, exclusive, lock_wait=None, lock=True):
...
+        chunks.update_pack_info() before closing, so index entries for the last
+        batch of chunks get real pack location values instead of UNKNOWN_*.
+        """
+        if self._pack_writer is not None:
+            return self._pack_writer.flush()
+        return None
+
     def close(self):
+        if self._pack_writer is not None:
+            self._pack_writer.flush()  # safety drain; caller must have called flush() to update index
```
ThomasWaldmann: That will flush remaining pieces to a pack, but it will discard the index updates returned by that call.

ThomasWaldmann: So, if that would only do half of the job and the caller really needs to rather call the flush method here: maybe better replace this by an assertion that the _pack_writer queue is actually empty here?

mr-raj12: replaced with an assert

ThomasWaldmann (approved):
LGTM.

Guess the real test will come when actually using the indexed pack-related values.

src/borg/repository.py
```diff
@@ -34,6 +34,71 @@ def repo_lister(repository, *, limit=None):
...
+        # Build the pack bytes once by joining all pieces (avoids O(n^2) copies
+        # that incremental string concatenation would cause in Python).
+        pack_data = b"".join(cdata for _, cdata in self._pieces)
+
+        # Determine pack_id.
+        # N=1: the pack contains exactly one chunk, so we keep pack_id == chunk_id
+        #      (backward-compatible file naming: packs/{chunk_id_hex}).
+        # N>1: the pack contains multiple chunks; use SHA256(pack_bytes) so the
+        #      file is content-addressed and borgstore can verify/cache it.
+        if len(self._pieces) == 1:
```
ThomasWaldmann: the condition is wrong, what you wanted is "if self.max_count == 1".

if self.max_count is a bigger value and len(self._pieces) happens to be be 1 when flushing (e.g. the final pack), you don't want to use the N=1 hack.

mr-raj12: fixed

PR #9725, opened 2026-06-06
allow msgpack 1.2.0

ThomasWaldmann:


ThomasWaldmann:
Did a local test with msgpack 1.2.0rc1 and it looked ok. master branch is not for production, so guess we can just allow and test it here now.

